PR #207, opened 2026-02-18
Fix duplicate migration 057 numbering conflict

brandonrc:
## Summary

- Renumber incus format migrations to resolve duplicate migration 057 conflict
  - `057_incus_format.sql` -> `058_incus_format.sql`
  - `058_incus_scan_type.sql` -> `059_incus_scan_type.sql`
- Regenerate SQLx offline cache for the corrected migration sequence

PR #205 (service accounts) and PR #206 (incus format) both created migration 057. When both merged to main, running migrations fails with a unique constraint violation on `_sqlx_migrations_pkey`.

## Test plan

- [ ] `cargo sqlx prepare --workspace` succeeds
- [ ] `cargo test --workspace --lib` passes (4865 tests)
- [ ] `cargo clippy --workspace` clean
- [ ] Fresh database applies all 59 migrations without error
- [ ] Container image builds and starts cleanly


brandonrc:
Fixing this in 208

PR #334, opened 2026-03-02
feat: add GCS with workload identity support

### LINKED ISSUE #340: feat: GCS Application Default Credentials / Workload Identity support
## Problem

The GCS storage backend requires an RSA private key (`GCS_PRIVATE_KEY` or `GCS_PRIVATE_KEY_PATH`) for all operations. This means deploying on GKE with Workload Identity, or anywhere that provides Application Default Credentials (ADC), requires exporting and managing a service account key file. Key files are a security liability (they don't rotate automatically, can be leaked, and bypass audit logging that Workload Identity provides).

## Proposed solution

Add a second authentication path for the GCS backend that uses GCP Application Default Credentials via the metadata server. When neither `GCS_PRIVATE_KEY` nor `GCS_PRIVATE_KEY_PATH` is set, the backend should operate in ADC mode:

- All storage operations use an `Authorization: Bearer` header with a token fetched from the GCE metadata server (`http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token`)
- Tokens are cached with proactive refresh before expiry
- Only `GCS_BUCKET` is required in ADC mode (`GCS_PROJECT_ID` and `GCS_SERVICE_ACCOUNT_EMAIL` become optional)
- `supports_redirect()` returns `false` in ADC mode (V4 signed URLs require an RSA private key)
- `get_presigned_url()` returns `Ok(None)` in ADC mode

The service account key path (Mode A) continues to work as before, with the addition of using OAuth2 bearer tokens for all API calls (not just reads).

## Use cases

- GKE deployments with Workload Identity (no key file needed, just annotate the KSA)
- Cloud Run, GCE, and other GCP services that provide ADC automatically
- Reduced attack surface by not storing long-lived service account keys

## Implementation

PR #334 by @todpunk (tested on a live GKE cluster with Workload Identity).

todpunk:
## Summary

Add a second authentication path for the GCS storage backend that uses GCP Application Default Credentials via the metadata server instead of an RSA private key. When neither GCS_PRIVATE_KEY nor GCS_PRIVATE_KEY_PATH is set, the backend operates in ADC mode: all storage operations send an Authorization: Bearer header with a token fetched from the GCP metadata server. Tokens are cached with proactive refresh 60 seconds before expiry using a tokio RwLock.

GcsConfig.project_id and .service_account_email are now Option<String>; they are still required (and validated) when a private key is present (Mode A), but are optional in ADC mode (Mode B). GCS_BUCKET remains required in both modes.

In ADC mode supports_redirect() returns false and get_presigned_url() returns Ok(None); signed URL generation is only available in Mode A.

Closes #340

## Test Checklist
- [X] Integration tests added/updated (if applicable)
- [X] E2E tests added/updated (if applicable)
- [X] Manually tested locally (specifically in a GKE cluster)

## API Changes
- [X] N/A - no API changes

### DIFF
--- a/backend/src/services/storage_service.rs
+++ b/backend/src/services/storage_service.rs
@@ -747,10 +747,12 @@ mod tests {
     // -----------------------------------------------------------------------
 
     // Serialize env-var tests to avoid cross-test interference.
-    static GCS_ENV_LOCK: std::sync::OnceLock<std::sync::Mutex<()>> = std::sync::OnceLock::new();
+    // Uses tokio::sync::Mutex so the guard can be held across .await points
+    // without triggering clippy::await_holding_lock.
+    static GCS_ENV_LOCK: std::sync::OnceLock<tokio::sync::Mutex<()>> = std::sync::OnceLock::new();
 
-    fn gcs_env_lock() -> &'static std::sync::Mutex<()> {
-        GCS_ENV_LOCK.get_or_init(|| std::sync::Mutex::new(()))
+    fn gcs_env_lock() -> &'static tokio::sync::Mutex<()> {
+        GCS_ENV_LOCK.get_or_init(|| tokio::sync::Mutex::new(()))
     }
 
     fn minimal_config(storage_backend: &str) -> crate::config::Config {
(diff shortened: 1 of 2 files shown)


todpunk:
I have also done the polish for the Dashboard storage health check and the StorageService support of gcs. API only validation worked fine, but testing the rest of my workflows revealed the missing pieces (and graceful error messages, thank you!)

PR #336, opened 2026-03-03
fix: strip index/ prefix from cargo sparse index upstream proxy path

todpunk:
## Summary
   The cargo_sparse_index_path() function included an 'index/' prefix that
   is correct for artifact-keeper's own routing but wrong for proxying to
   upstream registries. The Cargo sparse registry protocol stores index files
   at the path root (e.g. https://index.crates.io/se/rd/serde), not under
   an index/ subdirectory.

   - Add cargo_sparse_index_path_upstream() without the index/ prefix for upstream proxy URL construction
   - Update try_remote_index() to use the new function; also uses a new index_upstream_url field on RepoInfo to support registries like crates.io that split the sparse index (index.crates.io) and download API (crates.io) across two hosts
   - Rewrite try_virtual_index() to manually iterate members, using cargo_sparse_index_path_upstream() for proxy paths and honouring each member's index_upstream_url from repository_config
   - Load index_upstream_url from repository_config in resolve_cargo_repo()
   - Add index_upstream_url to CreateRepositoryRequest and UpdateRepositoryRequest (persisted to repository_config on create/update when provided)

## Test Checklist
- [X] Integration tests added/updated (if applicable)
- [X] E2E tests added/updated (if applicable)
- [X] Manually tested locally

## API Changes
- [X] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -47,12 +47,22 @@ pub fn router() -> Router<SharedState> {
             "/:repo_key/api/v1/crates/:name/:version/download",
             get(download),
         )
-        // Sparse index — various path layouts
+        // Sparse index — index/ prefixed paths (legacy / internal)
         .route("/:repo_key/index/1/:name", get(sparse_index_1))
         .route("/:repo_key/index/2/:name", get(sparse_index_2))
         .route("/:repo_key/index/3/:prefix/:name", get(sparse_index_3))
         .route(
             "/:repo_key/index/:prefix1/:prefix2/:name",
+            get(sparse_index_4plus),
+        )
+        // Sparse index — root-level paths (Cargo sparse registry protocol)
+        // Cargo clients expect index files at the registry root, not under index/.
+        // Axum prioritises static segments first so api/v1/crates etc. still win.
+        .route("/:repo_key/1/:name", get(sparse_index_1))
+        .route("/:repo_key/2/:name", get(sparse_index_2))
+        .route("/:repo_key/3/:prefix/:name", get(sparse_index_3))
+        .route(
+            "/:repo_key/:prefix1/:prefix2/:name",
             get(sparse_index_4plus),
         )
         .layer(DefaultBodyLimit::max(512 * 1024 * 1024)) // 512 MB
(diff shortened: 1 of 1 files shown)


brandonrc:
@todpunk I've rebased your branch on latest main and pushed some improvements on top. Here's what changed:

**Rebase on main** - Your branch was forked before PR #338 (the auth middleware unification) landed. The rebase picks that up, so the `publish` handler now uses the shared middleware auth path (`require_auth_with_bearer_fallback`) instead of the local `authenticate()` function. This means JWT, scoped API tokens, X-API-Key headers, and cookie auth all work correctly for `cargo publish`.

**Your core fix is preserved as-is:**
- `cargo_sparse_index_path_upstream()` (no `index/` prefix for upstream proxy)
- Root-level sparse index routes for standard Cargo clients
- `index_upstream_url` field for registries that split index and download hosts
- The rewritten `try_virtual_index` and `try_remote_index`

**Additional cleanup on top:**
- Replaced the N+1 per-member `repository_config` query in `try_virtual_index` with a single batch query using `ANY($1)`
- Added `tracing::warn!` for DB errors that were previously silently swallowed (both the `index_url_overrides` fetch and the artifact query)
- Extracted the duplicate `index_upstream_url` upsert SQL into a shared `upsert_index_upstream_url()` helper in `repositories.rs`
- Simplified `cargo_sparse_index_path` to delegate to `cargo_sparse_index_path_upstream` (eliminates duplicated match arms)
- Switched `resolve_cargo_repo` to use `query_scalar` instead of `query_as` with a tuple pattern

**One thing to note for a follow-up:** there's currently no way to *clear* an `index_upstream_url` once set via the API. An empty string would get upserted rather than deleting the row. Not blocking for this PR, but worth a small follow-up to handle that case.

All 6,264 unit tests pass, clippy clean, fmt clean.

brandonrc:
@todpunk Since this PR is from your fork, I wasn't able to push the rebased/improved version here directly. I've opened #342 with your work rebased on main plus the improvements described above.

Closing this one in favor of #342. Your commits are preserved as the first two in the history, with attribution intact. Thank you for catching this bug and for the thorough fix.

PR #377, opened 2026-03-05
fix: cargo with token, now with more performance

todpunk:
## Summary
When testing in a production environment (my k8s cluster) against real tools, cargo did not accept the token generated for a user on a private virtual repository. I got that working with some minor changes, but it was also over 4 seconds for pulling a package + deps when crates.io was 0.37 seconds, which was not an acceptable difference to me. Performance improvements to get that to a state of "working" are included. It is still just under 1 second for the pull, sadly.

I do not know if this is the approach, but it is at least illustrative of the issues (token and performance).

Claude Code's summary of the difference:

```
  Problem

  Artifact Keeper was ~4× slower than crates.io for a warm cargo build. The bottleneck was the index request path: for every GET /cargo/{repo}/se/rd/serde request, even on a fully-warm cache, the server was doing:

  - 2 storage reads per package: one for __cache_meta__.json (TTL/ETag check), one for the content itself
  - 1 SHA-256 hash of the full content bytes to verify integrity

  With 50 transitive dependencies, a warm cargo fetch triggered 100 storage reads and 50 hash computations before returning a single byte to cargo. This happens on every build because cargo re-validates its index on each run.

  crates.io avoids this entirely — it serves pre-generated static files from disk. Serving an index entry is a single filesystem read; the file on disk is the cache.

  Fix

  Added an in-process index entry cache to AppState:

  pub type IndexCache = Arc<RwLock<HashMap<String, (Bytes, Instant)>>>;

  Cache key: "{repo_key}:{crate_name_lowercase}". On a warm hit, serving an index entry is now one RwLock::read() + one HashMap::get() + one TTL comparison — no storage I/O, no DB query, no SHA-256.

  Cache population happens on the first request that misses — the result is stored immediately before returning the response. This covers all three repo types:
  - Remote repos (try_remote_index): bytes from the upstream proxy fetch are cached before building the response
  - Virtual repos (try_virtual_index): result from whichever member wins (local DB or upstream proxy) is cached under the virtual repo's key
  - Hosted repos (serve_index): DB query result is cached

  Cache invalidation on publish: after a successful cargo publish, the hosted repo's cache entry is dropped, and a single query finds any virtual repos containing this hosted repo and drops their entries too. The next request after publish rebuilds from DB and
  re-caches.

  TTL is 300 seconds (matching the HTTP max-age=300 already set on index responses). The storage-layer proxy cache (proxy-cache/ prefix) is unchanged — the in-process cache sits in front of it as a faster layer. A server restart clears the in-process cache; the
  first post-restart request hits storage (warm) rather than upstream (cold).
```

## Test Checklist
- [X] Manually tested locally

## API Changes
- [X] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -232,6 +232,14 @@ async fn config_json(
 ) -> Result<Response, Response> {
     let _repo = resolve_cargo_repo(&state.db, &repo_key, &state.repo_cache).await?;
 
+    // Check repo visibility from the cache (populated by resolve_cargo_repo).
+    let is_private = !state
+        .repo_cache
+        .read()
+        .ok()
+        .and_then(|c| c.get(&repo_key).map(|(r, _)| r.is_public))
+        .unwrap_or(true);
+
     // Determine the host from the request headers or fall back to config
     let host = headers
         .get("host")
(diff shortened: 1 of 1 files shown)


brandonrc:
> I guess the quality gate doesn't actually run on my commits. My apologies. All my commits are in though, and until I hear a request to the contrary this cook is out of the kitchen on this one

Appreciate the work! Think I fixed this issue with the pipeline, doing one more review and get this bad boy merged.

PR #419, opened 2026-03-11
fix: rename version_req to req in cargo sparse index dependency entries

inspired-geek:
## Summary
- Cargo's publish API sends dependency metadata with a `version_req` field, but the [sparse index format](https://doc.rust-lang.org/cargo/reference/registry-index.html) requires this field to be named `req`
- The `extract_index_fields` function now renames `version_req` → `req` on the fly when building index entries, so cargo clients can correctly parse dependency version requirements
- If `req` is already present (e.g. from a proxied upstream index), the existing value is preserved

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

## Test plan
- [x] `test_build_index_entry_renames_version_req_to_req` — verifies `version_req` is renamed to `req` and `version_req` is removed
- [x] `test_build_index_entry_keeps_req_if_already_correct` — verifies passthrough when metadata already uses `req`
- [x] All existing cargo handler tests pass
- [x] `cargo fmt --check` clean
- [x] `cargo clippy --workspace` clean
- [x] Coverage: 88% of new lines covered (2 uncovered lines are unreachable defensive fallbacks)

### DIFF
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -1707,41 +1707,26 @@ mod tests {
         assert_eq!(entry["yanked"], false);
     }
 
-    /// Build an index entry from metadata containing the given dep object,
-    /// then return the first dependency from the parsed index JSON.
-    fn index_dep_from_meta(dep_obj: serde_json::Value) -> serde_json::Value {
-        let meta = serde_json::json!({ "deps": [dep_obj], "features": {} });
-        let entry_str = build_index_entry("my-crate", "0.1.0", "abc123", Some(&meta));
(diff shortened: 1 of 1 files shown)


brandonrc:
> @brandonrc It looks like some kind of bug in SonarCloud. I see duplications in the report, but only in [backend/src/api/handlers/maven.rs](https://sonarcloud.io/component_measures?id=artifact-keeper_artifact-keeper&metric=new_duplicated_lines_density&pullRequest=419&selected=artifact-keeper_artifact-keeper%3Abackend%2Fsrc%2Fapi%2Fhandlers%2Fmaven.rs) which is irrelevant to my changes. I've tried several tricks to pass the check, but no luck so far.

Yea that was a weird one, just merging it in

PR #435, opened 2026-03-15
fix: paginated Meilisearch reindex to avoid OOM on large registries

inspired-geek:
## Summary

- Replace load-all-into-memory reindex with cursor-based pagination (`WHERE id > $1 LIMIT $2`) to prevent OOM on large registries
- Separate download count aggregation into per-batch `WHERE artifact_id = ANY($1)` query, eliminating full-table LEFT JOIN subquery
- Fix `MeiliService::new` to return `Result` instead of panicking, with graceful degradation (server continues without search)
- Return actual indexed counts from reindex functions; admin endpoint now reports real numbers

## Test plan

- [x] `cargo test --lib meili_service::tests` — 30 unit tests (source-introspection guards for cursor pagination, no OFFSET, batch-scoped download counts, progress logging, error context, return types)
- [x] `DATABASE_URL=... cargo test --test meili_reindex_tests -- --ignored` — 6 integration tests verifying multi-batch pagination, exact batch boundaries, single-row edge case, repository pagination, and batch-scoped download counts against real PostgreSQL
- [x] `cargo clippy --workspace` — clean
- [x] `cargo fmt --check` — clean

### DIFF
--- a/backend/src/services/meili_service.rs
+++ b/backend/src/services/meili_service.rs
@@ -343,15 +343,7 @@ impl MeiliService {
                 .collect();
 
             last_id = rows.last().map(|row| row.id);
-            let documents: Vec<ArtifactDocument> = rows
-                .into_iter()
-                .map(|row| {
-                    // Artifacts with no download records default to 0,
-                    // matching the previous COALESCE(ds.download_count, 0) behavior.
-                    let download_count = download_counts.get(&row.id).copied().unwrap_or_default();
-                    artifact_document_from_row(row, download_count)
-                })
-                .collect();
+            let documents = build_artifact_batch(rows, &download_counts);
 
             total += documents.len();
             index
@@ -422,8 +414,7 @@ impl MeiliService {
             }
 
             last_id = rows.last().map(|row| row.id);
-            let documents: Vec<RepositoryDocument> =
-                rows.into_iter().map(repository_document_from_row).collect();
+            let documents = build_repository_batch(rows);
 
             total += documents.len();
             index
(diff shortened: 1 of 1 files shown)


inspired-geek:
@brandonrc

**Coverage gap on new code (65% vs 70% threshold)**

The new `full_reindex_artifacts()` and `full_reindex_repositories()` methods contain ~170 lines of loop logic that depends on both PgPool and Meilisearch SDK at the same time. Unit tests (`cargo test --lib`) can't reach this code without mocking both services, so SonarCloud reports 65% coverage on new code (threshold: 70%).

The cursor pagination itself is already validated by integration tests in `meili_reindex_tests.rs` (require PostgreSQL, run with `--ignored`), and the batch-building logic is covered by unit tests. The gap is specifically the orchestration code inside the loops.

**Options:**

1. **Trait abstraction** — introduce `SearchIndexer` + `DataSource` traits (similar to existing `StorageBackend`), mock both in unit tests. Covers the loop logic fully, but adds ~200 lines of refactoring and widens the scope of this bugfix PR.

2. **wiremock for Meilisearch HTTP** — mock only the Meilisearch side; DB queries stay uncovered. Simpler, partial improvement, may or may not cross 70%.

3. **Defer to a follow-up** — merge as-is, open a separate issue for the trait refactoring so the bugfix isn't blocked by a coverage gate on inherently integration-level code.

What's your preference?

brandonrc:
Thanks for this contribution! The cursor-based pagination and download count decomposition are well-engineered.

I've incorporated your changes into #440 with one fix: moved the `total` counter increment to after the `add_documents` await. The original code incremented before the fallible call, so error messages would report inflated counts (including the failed batch). Now `total` only reflects successfully indexed documents.

All 6,569 unit tests pass with the change. Closing this in favor of #440.

PR #445, opened 2026-03-17
Ensured caddy ignores proxy environment variables

msegura501:
## Summary
FIx  https://github.com/orgs/artifact-keeper/discussions/444

Should ease the [quickstart](https://artifactkeeper.com/docs/getting-started/quickstart/#quick-start)

(Apologies in advance for my own clumsiness I am more stupid than evil, I meant only to provide a very small help)

## Test Checklist
- [ X ] No regressions in existing tests

## API Changes
- [ X ] N/A - no API changes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -274,6 +274,12 @@ services:
         condition: service_started
     environment:
       SITE_ADDRESS: ${SITE_ADDRESS:-localhost}
+      HTTP_PROXY: ""
+      HTTPS_PROXY: ""
+      http_proxy: ""
+      https_proxy: ""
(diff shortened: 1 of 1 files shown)


todpunk:
(I should also emphasize this isn't speaking with any authority, just my own two cents, and I appreciate your explanation. It's an interesting use case.)

msegura501:
I will test but yes I think it can be worked around with an environment file :)
I do not mean to impose, I can close the issue or leave it open if this is interesting. 
To be honest, I missed the idea of using an environment file in the first place, so no problem at least at our side.

PR #484, opened 2026-03-20
Fix large generic artifact uploads by removing the 512MB body limit

TechEnchante:
The generic repository artifact upload routes in backend/src/api/handlers/repositories.rs are hard-limited to 512 MB via DefaultBodyLimit::max(512 * 1024 * 1024). This causes large uploads, such as 8 GB ISOs, to fail during multipart parsing with a 400 validation error. The top-level API router already disables Axum’s default body limit for large artifact uploads, so the repository router should not reintroduce a restrictive 512 MB cap for generic artifact uploads.

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -134,7 +134,7 @@ pub fn router() -> Router<SharedState> {
         // Label routes nested under repository
         .merge(super::repository_labels::repo_labels_router())
         // Allow up to 512MB uploads (matches format-specific handlers)
-        .layer(DefaultBodyLimit::max(512 * 1024 * 1024))
+        .layer(DefaultBodyLimit::disable())
 }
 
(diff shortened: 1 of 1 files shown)


brandonrc:
@TechEnchante Thanks for identifying this limitation. You're right that 512MB is too restrictive for large artifacts like ISOs.

However, disabling the body limit entirely creates a memory exhaustion risk since Axum buffers the full request body before passing it to handlers. A single large upload could consume all available memory.

The industry standard (Artifactory, Nexus, GitLab) is to make this configurable rather than removing it. I've created two issues to address this properly:

- **Backend:** #487 - Add a `MAX_UPLOAD_SIZE` env var (default 10GB) that controls the body limit across all upload routes
- **Web:** artifact-keeper/artifact-keeper-web#189 - Surface the limit in the UI and show helpful error messages when exceeded

This gives operators control over the limit for their environment while keeping a safety cap in place. A PR implementing #487 is welcome if you'd like to contribute the configurable approach.

PR #519, opened 2026-03-22
fix: return raw binary content from artifact download endpoint

brandonrc:
## Summary

`GET /:key/artifacts/*path` was mapped to `get_artifact_metadata` which returns JSON metadata. Clients uploading binary content and downloading from the same path got a JSON envelope instead of the raw file, causing checksum mismatches.

Remapped the route to `download_artifact` which returns raw bytes with correct Content-Type, Content-Length, and x-checksum-sha256 headers.

## Test Checklist
- [ ] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -121,7 +121,7 @@ pub fn router() -> Router<SharedState> {
         )
         .route(
             "/:key/artifacts/*path",
-            get(download_artifact)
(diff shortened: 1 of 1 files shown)


brandonrc:
Closing: the route structure is intentional (metadata vs download). The integrity test was hitting the metadata endpoint instead of the download endpoint. Fixed in artifact-keeper-test PR #13 instead.

PR #650, opened 2026-04-04
feat: quickstart installer script (curl | bash)

brandonrc:
## Summary

Adds a one-command installer for evaluating Artifact Keeper:

```
curl -fsSL https://get.artifactkeeper.com | bash
```

The script:
- Detects Docker (Compose V2/V1) or Podman (podman-compose)
- Generates secure random credentials (admin password, JWT secret, Meilisearch key)
- Downloads `docker-compose.yml` and `docker/` configs from the repo (no duplication, always current)
- Writes a `.env` with all generated secrets
- Pulls images, starts the stack, waits for health, prints URL + credentials
- Opens browser on macOS/Linux

No versions are hardcoded. `AK_VERSION` defaults to `latest` and controls the Docker image tag. All configuration is via environment variables.

Options:
- `AK_MINIMAL=1` skips scanners (Trivy, OpenSCAP, Dependency-Track) for lighter resource usage
- `AK_SKIP_START=1` generates config files without starting
- `AK_HTTP_PORT` / `AK_HTTPS_PORT` for custom ports
- `AK_ADMIN_PASSWORD` to set a specific password

Still needs:
- [ ] Set up `get.artifactkeeper.com` to serve the script (CNAME or CDN)
- [ ] Windows PowerShell variant (`irm https://get.artifactkeeper.com/windows | iex`)
- [ ] Native binary fallback for systems without Docker
- [ ] Test on various Linux distros, macOS Intel/ARM, WSL2

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/scripts/install.sh
+++ b/scripts/install.sh
@@ -1,0 +1,307 @@
+#!/usr/bin/env bash
+# =============================================================================
+# Artifact Keeper - Quickstart Installer
+# =============================================================================
+#
+# One-command setup for evaluating Artifact Keeper locally.
+# Detects Docker or Podman, generates secure credentials, downloads the
+# compose file, and starts the full stack.
+#
+# Usage:
+#   curl -fsSL https://get.artifactkeeper.com | bash
+#
+# Or run directly:
+#   bash scripts/install.sh
+#
+# Options (environment variables):
+#   INSTALL_DIR       - Where to create the project directory (default: ./artifact-keeper)
+#   AK_VERSION        - Version to install (default: latest)
+#   AK_HTTP_PORT      - HTTP port (default: 80)
+#   AK_HTTPS_PORT     - HTTPS port (default: 443)
+#   AK_ADMIN_PASSWORD - Admin password (default: auto-generated)
+#   AK_SKIP_START     - Set to 1 to generate files without starting (default: 0)
+#   AK_MINIMAL        - Set to 1 to skip scanners (Trivy, OpenSCAP, Dependency-Track)
+#
+# Requirements:
+#   - Docker 20.10+ with Compose V2, or Podman 4+ with podman-compose
+#   - 2 GB RAM minimum (4 GB recommended with scanners enabled)
+#   - Ports 80 and 443 available (configurable)
+#
+# =============================================================================
+set -euo pipefail
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+INSTALL_DIR="${INSTALL_DIR:-./artifact-keeper}"
+AK_VERSION="${AK_VERSION:-latest}"
+AK_HTTP_PORT="${AK_HTTP_PORT:-80}"
+AK_HTTPS_PORT="${AK_HTTPS_PORT:-443}"
+AK_ADMIN_PASSWORD="${AK_ADMIN_PASSWORD:-}"
+AK_SKIP_START="${AK_SKIP_START:-0}"
+AK_MINIMAL="${AK_MINIMAL:-0}"
+
+REPO_BASE="https://raw.githubusercontent.com/artifact-keeper/artifact-keeper/main"
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+info()  { printf "\033[1;34m==>\033[0m %s\n" "$*"; }
+ok()    { printf "\033[1;32m==>\033[0m %s\n" "$*"; }
+warn()  { printf "\033[1;33m==>\033[0m %s\n" "$*" >&2; }
+die()   { printf "\033[1;31mError:\033[0m %s\n" "$*" >&2; exit 1; }
+
+generate_password() {
+    # 24-char alphanumeric password, works on Linux and macOS
+    if command -v openssl >/dev/null 2>&1; then
+        openssl rand -base64 18 | tr -d '/+=' | head -c 24
+    elif [ -f /dev/urandom ]; then
+        LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 24
+    else
+        # Last resort: use date + PID as entropy source
+        echo "AK$(date +%s%N | sha256sum | head -c 22)" 2>/dev/null || \
+        echo "AK$(date +%s)$$$(od -An -N8 -tx1 /dev/random 2>/dev/null | tr -d ' ')" | head -c 24
+    fi
+}
+
+# ---------------------------------------------------------------------------
+# Pre-flight checks
+# ---------------------------------------------------------------------------
+
+info "Artifact Keeper Quickstart Installer"
+echo ""
+
+# Detect container runtime
+COMPOSE_CMD=""
+if command -v docker >/dev/null 2>&1 && docker compose version >/dev/null 2>&1; then
+    COMPOSE_CMD="docker compose"
+    info "Found Docker with Compose V2"
+elif command -v docker >/dev/null 2>&1 && command -v docker-compose >/dev/null 2>&1; then
+    COMPOSE_CMD="docker-compose"
+    info "Found Docker with docker-compose"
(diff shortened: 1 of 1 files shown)


brandonrc (commented):
## Code Review: Quickstart Installer Script

Overall this is a well-structured installer with good UX (colored output, browser opening, clear credential display). The header docs are thorough and the container runtime detection is solid. Below are findings grouped by severity.

---

### MUST FIX

**1. `sed -i` is not portable between macOS and Linux (line 197)**

```
sed -i.bak 's/DEPENDENCY_TRACK_ENABLED=true/DEPENDENCY_TRACK_ENABLED=false/' .env
rm -f .env.bak
```

GNU `sed -i` and BSD (macOS) `sed -i` have different semantics. On macOS, `sed -i ''` requires an empty string argument; `sed -i.bak` works on both but only because you explicitly pass the backup extension. This particular form actually does work cross-platform, but the approach is fragile. Since you are already generating the `.env` with `cat > .env`, just conditionally set the value in the heredoc instead of writing it and then patching it. For example:

```bash
DTRACK_ENABLED="true"
if [ "$AK_MINIMAL" = "1" ]; then
    DTRACK_ENABLED="false"
fi
```

Then use `DEPENDENCY_TRACK_ENABLED=${DTRACK_ENABLED}` in the heredoc.

**2. Version variable mismatch: `AK_VERSION` has no effect on most images**

The installer writes `ARTIFACT_KEEPER_VERSION=${AK_VERSION}` to `.env`, but the compose file only references `${ARTIFACT_KEEPER_VERSION}` for the `dtrack-init` service (line 107). The `backend`, `web`, and `openscap` images are hardcoded to `:latest`. This means `AK_VERSION=1.1.0` will not actually pin those images.

Either fix the compose file to use `${ARTIFACT_KEEPER_VERSION:-latest}` on all AK-owned images, or document that version pinning only works for the init container. This is a correctness issue since the PR description claims "AK_VERSION defaults to latest and controls the Docker image tag."

**3. Port default mismatch between installer and compose file**

The installer defaults `AK_HTTP_PORT` to 80 and `AK_HTTPS_PORT` to 443, writing them as `HTTP_PORT` and `HTTPS_PORT` to `.env`. But the compose file defaults to `${HTTP_PORT:-30080}` and `${HTTPS_PORT:-30443}`. The `.env` values will override the compose defaults, so the behavior is correct when using the installer. However, if a user runs the compose file without the installer (no `.env`), they get ports 30080/30443, while the installer gives them 80/443. This inconsistency should be intentional and documented, or the defaults should be aligned.

**4. No cleanup on failure**

If the script fails partway through (e.g., a `curl` download fails after creating the directory and writing partial files), the user is left with a broken `$INSTALL_DIR`. Add a trap to clean up on error:

```bash
cleanup() {
    if [ "${INSTALL_COMPLETE:-0}" = "0" ] && [ -d "$INSTALL_DIR" ]; then
        warn "Installation failed. Cleaning up $INSTALL_DIR..."
        rm -rf "$INSTALL_DIR"
    fi
}
trap cleanup EXIT
# ... at the very end, before the success message:
INSTALL_COMPLETE=1
```

**5. No integrity verification for downloaded files**

The script downloads 5 files from GitHub over HTTPS with no checksum verification. If GitHub's CDN serves a cached stale version or a truncated response (which `curl -f` alone does not fully protect against for non-zero-length partial content), the user gets a broken install. At minimum, verify the downloaded compose file is valid YAML and non-empty:

```bash
if [ ! -s docker-compose.yml ]; then
    die "Failed to download docker-compose.yml (empty file)"
fi
```

---

### SHOULD FIX

**6. Health check URL may not go through Caddy on port 80**

The health check polls `http://localhost:${AK_HTTP_PORT}/health` and `/livez`. Caddy is the service bound to `$AK_HTTP_PORT`, so the health request goes through the reverse proxy. If Caddy is healthy but the backend is still starting, the check may get a 502 from Caddy rather than a connection refused, and `curl -sf` treats a 502 as a failure. This should work, but consider checking the backend directly on port 8080 (which is exposed in compose via Caddy, not directly). Actually, looking again, port 8080 is NOT exposed to the host. The health check through Caddy is the right approach but the 502 issue means it may report "still starting" even when things are fine. Worth a comment in the script explaining this.

**7. `generate_password` last-resort fallback is weak**

The third branch of `generate_password` (no openssl, no /dev/urandom) uses `date +%s%N | sha256sum`. The `%N` nanoseconds format is not supported on macOS (`date` is BSD). The fallback to `od /dev/random` will never be reached on macOS because `/dev/urandom` always exists there. But on minimal containers or CI environments where `/dev/urandom` might not exist and `openssl` is not installed, this path produces predictable output. Consider documenting that openssl is effectively required or checking for it explicitly.

**8. `$COMPOSE_CMD pull` output suppressed too aggressively**

```bash
$COMPOSE_CMD pull 2>&1 | tail -1
```

This hides all pull output except the last line. If a pull fails for a specific image (auth error, rate limit), the user only sees the last line of output, which may not be the error. Show stderr separately or at minimum show the last few lines:

```bash
$COMPOSE_CMD pull 2>&1 | tail -5
```

Or better, do not suppress output at all since image pulls can take minutes and progress feedback is valuable.

**9. Minimal mode starts all services anyway**

When `AK_MINIMAL=1`, the script selectively pulls only `postgres meilisearch backend web caddy`, but `$COMPOSE_CMD up -d` starts ALL services defined in the compose file, including trivy, openscap, and dependency-track. The pull was selective, but `up -d` is not. This means scanner containers will start but may fail because their images were not pulled (or will pull at start time, defeating the purpose of minimal mode).

Fix by passing the same service list to `up -d`:

```bash
if [ "$AK_MINIMAL" = "1" ]; then
    $COMPOSE_CMD up -d postgres meilisearch backend web caddy
else
    $COMPOSE_CMD up -d
fi
```

**10. `ADMIN_PW_GENERATED` referenced without `set -u` safety**

The variable `ADMIN_PW_GENERATED` is set inside a conditional block (line 131-135), but if that code path somehow doesn't execute (impossible with current logic, but defensive programming), line 276 would fail with `set -u`. Minor, but since `set -u` is enabled, initialize `ADMIN_PW_GENERATED=0` earlier with the other defaults.

---

### CONSIDER

**11. The "quick start" curl example in the output leaks the admin password to terminal history.**

```
curl -u admin:${AK_ADMIN_PASSWORD} http://localhost:${AK_HTTP_PORT}/api/v1/repositories ...
```

This puts the password in the terminal scrollback. Consider using `-u admin` (which prompts for the password) or at least note that the password is visible.

**12. Support `--help` flag**

The script has no argument parsing. Adding `if [ "${1:-}" = "--help" ]; then ...` would make it friendlier for users who download and run it directly rather than via curl-pipe.

**13. `INSTALL_DIR` defaults to a relative path**

`INSTALL_DIR` defaults to `./artifact-keeper`, which is resolved to an absolute path via `cd` + `pwd` on line 157. This works, but if the user sets `INSTALL_DIR=/opt/artifact-keeper`, the `cd` + `pwd` approach still works. Good.

**14. Quoting in the quick-start curl example**

The JSON body in the example uses single quotes inside double quotes, which is correct for bash, but may confuse users copying the example. Minor.

---

### LOOKS GOOD

- `set -euo pipefail` at the top -- good.
- HTTPS for all downloads from `raw.githubusercontent.com`.
- Container runtime detection covers Docker Compose V2, V1, Podman with podman-compose, and Podman Compose V2 -- thorough.
- Port availability check using `lsof` with graceful fallback.
- Credential generation prefers `openssl rand` with sensible fallbacks.
- Existing installation guard prevents accidental overwrites.
- `AK_SKIP_START=1` for config-only mode is a nice touch.
- Browser opening is best-effort with `|| true`.
- The colored output helpers are clean and consistent.

Summary: The main issues to address before merge are the version variable mismatch (#2), the minimal mode not actually being minimal on `up -d` (#9), and the lack of failure cleanup (#4). The rest are hardening improvements.

PR #695, opened 2026-04-12
fix: proxy APT repo metadata from upstream for remote repos (#674)

### LINKED ISSUE #674: APT remote type does not proxy the `InRelease` and `Release` files
### Description

Currently, the APT source only seems to generate `InRelease` and `Release` files from private repositories, so it lacks pull-through capability.

Additionally, since `InRelease` and `Release` store repository metadata, their cache timeout should be shorter than that of other files. However, it seems that `artifact-keeper` currently lacks this capability?

### Environment Information

- Deployment type: docker-compose
- Operating System: 
- Architecture: 
- Database: 
- Reverse Proxy: None
- Network setup: 

### Backend Version

1.1.2

### Web UI Version

_No response_

### Package Format

None

### Expected Behavior

APT pull-through works

### Observed Behavior

`apt update` fails to fetch the `Release` files.

### Steps to Reproduce

repo info:
```shell
# ak repo show debian
Key:          debian
Name:         Debian
Format:       debian
Type:         remote
Upstream URL: http://deb.debian.org/debian
Public:       yes
Description:
Storage Used: 0 B
Quota:        unlimited
Created:      2026-04-09 09:38:10 UTC
Updated:      2026-04-09 09:38:10 UTC
```

1. modify `/etc/apt/sources.list.d/debian.sources` for mirror url
2. run `apt update`

### Logs

```shell
$ apt update
Get:1 http://ip:10000/debian/debian trixie InRelease [356 B]
Ign:1 http://ip:10000/debian/debian trixie InRelease
Get:2 http://ip:10000/debian/debian trixie-updates InRelease [372 B]
Ign:2 http://ip:10000/debian/debian trixie-updates InRelease
Get:3 http://ip:10000/debian/debian trixie Release [356 B]
Get:4 http://ip:10000/debian/debian trixie-updates Release [372 B]
Ign:5 http://ip:10000/debian/debian trixie Release.gpg
Ign:6 http://ip:10000/debian/debian trixie-updates Release.gpg
Get:7 http://deb.debian.org/debian-security trixie-security InRelease [43.4 kB]
Get:8 http://deb.debian.org/debian-security trixie-security/main amd64 Packages [121 kB]
Fetched 165 kB in 1s (297 kB/s)
All packages are up to date.
Notice: Missing Signed-By in the sources.list(5) entry for 'http://ip:10000/debian/debian'
Notice: Missing Signed-By in the sources.list(5) entry for 'http://ip:10000/debian/debian'
```

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

Fixes #674: APT remote proxy repos returned tiny generated Release files (~356 bytes) instead of proxying the real upstream metadata, breaking \`apt-get update\` for all Debian/Ubuntu pull-through scenarios.

### Root cause

The \`release_file\`, \`in_release_file\`, \`release_gpg\`, \`packages_index\`, and \`packages_index_gz\` handlers in \`backend/src/api/handlers/debian.rs\` unconditionally generated Release/Packages content from the local artifacts table, regardless of whether the repo was type \`Remote\`. The \`pool_download\` handler already had proxy support for \`.deb\` files, but the metadata paths were missed, so apt would fetch the signed metadata from our tiny generated file, fail signature verification against the upstream mirror's key, and then have no way to discover packages.

From the reporter's log:
\`\`\`
Get:1 http://ip:10000/debian/debian trixie InRelease [356 B]
Ign:1 http://ip:10000/debian/debian trixie InRelease
\`\`\`

A real \`InRelease\` from deb.debian.org is ~80KB, not 356 bytes.

### Fix

Added a single \`try_proxy_debian_metadata\` helper that fetches the exact upstream path through the existing \`proxy_service\` when the repo is \`RepositoryType::Remote\`. The upstream bytes are returned unchanged so apt can verify the upstream mirror's GPG signature against its own key ring. Local and Staging repos still take the existing generated-content path.

Applied to all five metadata endpoints:
- \`dists/{dist}/Release\`
- \`dists/{dist}/InRelease\`
- \`dists/{dist}/Release.gpg\`
- \`dists/{dist}/{component}/binary-{arch}/Packages\`
- \`dists/{dist}/{component}/binary-{arch}/Packages.gz\`

The existing proxy cache (TTL 24h by default) handles cache timing. The reporter also asked for a shorter TTL for Release files specifically; that's deferred to a follow-up since it requires format-aware cache TTL which is a bigger change.

### Tests

Added 7 unit tests covering upstream path construction for all five endpoints, plus an end-to-end URL assembly test that pins the exact URL apt will hit against \`http://deb.debian.org/debian\` for the trixie InRelease case.

## Test Checklist
- [x] Manually tested locally (cargo test --workspace --lib: 7295 passed)

## API Changes
- [x] N/A - no API changes

Closes #674

### DIFF
--- a/backend/src/api/handlers/debian.rs
+++ b/backend/src/api/handlers/debian.rs
@@ -1315,101 +1315,54 @@ mod tests {
     // Upstream path construction for APT remote proxy (#674)
     // -----------------------------------------------------------------------
 
-    fn build_release_upstream_path(distribution: &str) -> String {
-        format!("dists/{}/Release", distribution)
-    }
-
-    fn build_inrelease_upstream_path(distribution: &str) -> String {
-        format!("dists/{}/InRelease", distribution)
-    }
-
-    fn build_release_gpg_upstream_path(distribution: &str) -> String {
-        format!("dists/{}/Release.gpg", distribution)
-    }
-
-    fn build_packages_upstream_path(
-        distribution: &str,
-        component: &str,
-        binary_arch: &str,
-    ) -> String {
-        format!(
-            "dists/{}/{}/{}/Packages",
-            distribution, component, binary_arch
-        )
-    }
-
-    fn build_packages_gz_upstream_path(
-        distribution: &str,
-        component: &str,
-        binary_arch: &str,
-    ) -> String {
-        format!(
-            "dists/{}/{}/{}/Packages.gz",
-            distribution, component, binary_arch
-        )
+    #[test]
+    fn test_upstream_dists_paths_match_debian_mirror_layout() {
+        // All five metadata endpoints build upstream paths via
+        // try_proxy_dists_file(state, repo, key, dist, suffix, ct).
+        // The path is always "dists/{dist}/{suffix}". Verify the
+        // expected paths match the real Debian/Ubuntu mirror layout.
+        let cases = vec![
+            ("trixie", "Release", "dists/trixie/Release"),
+            ("trixie-updates", "Release", "dists/trixie-updates/Release"),
+            ("bookworm", "InRelease", "dists/bookworm/InRelease"),
+            (
+                "bookworm-security",
+                "InRelease",
+                "dists/bookworm-security/InRelease",
+            ),
+            ("trixie", "Release.gpg", "dists/trixie/Release.gpg"),
+            (
+                "trixie",
+                "main/binary-amd64/Packages",
+                "dists/trixie/main/binary-amd64/Packages",
+            ),
+            (
+                "trixie",
+                "non-free/binary-arm64/Packages",
+                "dists/trixie/non-free/binary-arm64/Packages",
+            ),
+            (
+                "trixie",
+                "main/binary-amd64/Packages.gz",
+                "dists/trixie/main/binary-amd64/Packages.gz",
+            ),
+        ];
+        for (dist, suffix, expected) in &cases {
+            let path = format!("dists/{}/{}", dist, suffix);
+            assert_eq!(
+                &path, expected,
+                "path mismatch for dist={}, suffix={}",
+                dist, suffix
+            );
+        }
     }
 
     #[test]
-    fn test_upstream_path_release_matches_debian_mirror_layout() {
-        assert_eq!(
-            build_release_upstream_path("trixie"),
-            "dists/trixie/Release"
-        );
-        assert_eq!(
-            build_release_upstream_path("trixie-updates"),
-            "dists/trixie-updates/Release"
-        );
-    }
-
-    #[test]
-    fn test_upstream_path_inrelease_matches_debian_mirror_layout() {
-        assert_eq!(
-            build_inrelease_upstream_path("bookworm"),
-            "dists/bookworm/InRelease"
-        );
-        assert_eq!(
-            build_inrelease_upstream_path("bookworm-security"),
-            "dists/bookworm-security/InRelease"
-        );
-    }
-
-    #[test]
-    fn test_upstream_path_release_gpg_matches_debian_mirror_layout() {
-        assert_eq!(
-            build_release_gpg_upstream_path("trixie"),
-            "dists/trixie/Release.gpg"
-        );
-    }
-
-    #[test]
-    fn test_upstream_path_packages_matches_debian_mirror_layout() {
-        assert_eq!(
-            build_packages_upstream_path("trixie", "main", "binary-amd64"),
-            "dists/trixie/main/binary-amd64/Packages"
-        );
-        assert_eq!(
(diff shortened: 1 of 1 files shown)


brandonrc (commented):
## Code Review: fix: proxy APT repo metadata from upstream for remote repos (#674)

The fix correctly identifies and addresses the root cause. All five metadata endpoints now proxy from upstream when the repo type is Remote, and local/staging repos continue to use the generated-content path. The existing proxy cache and SSRF validation (via `validate_outbound_url` at repo creation time) apply automatically. Good work overall.

---

### SHOULD FIX: Double database query on the local/staging path

For non-remote repos (local and staging), every request hits the database twice:

1. `DebianProxy::resolve()` calls `resolve_debian_repo()` (first query)
2. `proxy.dists()` returns `Ok(())` because `repo_type != Remote`
3. `local_release_content()` calls `resolve_debian_repo()` again (second query)

The `repo` from `DebianProxy::resolve` is already available but unused for the local path. `local_release_content` should accept a `RepoInfo` instead of re-querying, or the handlers should reuse the repo from `DebianProxy::resolve`. For example, `release_file` could pass `repo.id` directly to `generate_release_content`:

```rust
let (proxy, repo) = DebianProxy::resolve(&state, &repo_key, &distribution).await?;
proxy.dists("Release", "text/plain; charset=utf-8", &repo).await?;
let release = generate_release_content(&state, repo.id, &distribution).await?;
```

This eliminates `local_release_content` entirely and removes one DB round-trip per request.

---

### SHOULD FIX: `Err(Response)` for successful proxy responses is misleading

`DebianProxy::dists()` returns `Err(response_with_200_OK)` to short-circuit the handler when upstream returns successfully. This is the only place in the codebase that uses `Err(Response::builder().status(StatusCode::OK)...)` for a success path. Every other handler (Alpine, npm, Maven, etc.) uses `return Ok(response)` for successful proxy results.

The `Result<(), Response>` signature forces this because `Ok(())` means "fall through to local" and `Err(response)` means "done, return this." The type is technically sound since Axum's `Result<Response, Response>` treats both sides as HTTP responses, but the semantics are inverted and will confuse future readers.

A clearer approach: return `Result<Option<Response>, Response>` where `Ok(Some(response))` means proxied, `Ok(None)` means fall through, and `Err` means actual error. Or return an enum. The handlers would then be:

```rust
if let Some(response) = proxy.dists("Release", "text/plain; charset=utf-8", &repo).await? {
    return Ok(response);
}
```

This matches how the pool download handler and Alpine handler both work (check, return early with `Ok` on hit, fall through on miss).

---

### CONSIDER: Test comment references non-existent function name

The test comment at line 1279 mentions `try_proxy_dists_file(state, repo, key, dist, suffix, ct)`, but this function does not exist in the code. The actual function is `DebianProxy::dists()`. Minor, but could confuse someone searching for the function by name.

---

### CONSIDER: Tests only validate string formatting, not actual proxy behavior

The two new tests verify that `format!("dists/{}/{}", dist, suffix)` produces the expected string. This is testing `std::fmt` more than the application logic. The path construction itself is just a single `format!` call inside `DebianProxy::dists()`, so there is no complex logic to get wrong.

What is not tested (and harder to test without integration setup):
- That `proxy.dists()` is actually called before the local fallback
- That upstream errors (502, timeout) result in the right HTTP status
- That the content-type fallback logic works (upstream CT vs. default CT)

Not blocking, since integration tests cover this at a higher level, but worth noting that the unit test coverage here is shallow.

---

### CONSIDER: SonarCloud duplication (15.7%)

The duplication likely comes from the repeated pattern in each handler:

```rust
let (proxy, repo) = DebianProxy::resolve(&state, &repo_key, &distribution).await?;
proxy.dists("...", "...", &repo).await?;
let (release, repo) = local_release_content(&state, &repo_key, &distribution).await?;
```

If the double-query fix above is applied (removing `local_release_content` and inlining the `generate_release_content` call), that should reduce the duplication. Beyond that, the remaining duplication is structural (each handler needs its own Axum signature and response building), which is not worth abstracting further.

---

### LOOKS GOOD

- All 5 metadata endpoints covered: Release, InRelease, Release.gpg, Packages, Packages.gz
- Upstream content is passed through byte-for-byte, preserving GPG signatures for `apt` verification
- Content-type correctly prefers the upstream header with a sensible fallback
- SSRF protection is inherited: `upstream_url` is validated via `validate_outbound_url` at repo creation time, and `ProxyService` validates OCI bearer token realm URLs
- Proxy cache (24h default TTL) applies automatically via `ProxyService::fetch_artifact`
- The `_distribution` to `distribution` rename in `packages_index` and `packages_index_gz` is correct; the old code silently ignored the distribution parameter
- The `DebianProxy` struct is a reasonable abstraction that reduces per-handler boilerplate

Issue #688, opened 2026-04-10
Prevent Password Reuse by Tracking Password History


thimomulder:
## Summary
Prevent local users from reusing recent passwords by tracking password history securely.
***
## Motivation
Allowing password reuse undermines the effectiveness of password expiration and resets. Tracking password history reduces this risk significantly.

### Proposed Solution

Store hashes of the previous _N_ passwords per user and disallow reuse should be configured within the Web UI/REST-API

Key requirements:
- Only password hashes are stored
- Hashes use the same strong hashing mechanism as current passwords
- Comparison is done securely

### Component

None

## Benefits
- Stronger long-term credential hygiene
- Compatibility with expiration and forced-change features

### Pre-submission Checklist

- [x] I have searched existing issues and feature requests to ensure this isn't a duplicate
- [x] I have reviewed the documentation to confirm this feature doesn't already exist

Issue #687, opened 2026-04-10
Password Expiration for Local Users


thimomulder:
## Summary
Allow administrators to configure password expiration for local users after a defined number of days.
***
## Motivation
Password expiration policies are often required by security standards and internal policies, especially in regulated environments.

### Proposed Solution

Introduce a configurable password lifetime (in days) configurable within the Web UI/REST-API.

Behavior:
- Passwords expire automatically after the configured interval
- Expired passwords require immediate change at login

### Component

None

### Pre-submission Checklist

- [x] I have searched existing issues and feature requests to ensure this isn't a duplicate
- [x] I have reviewed the documentation to confirm this feature doesn't already exist

Issue #683, opened 2026-04-10
Force Password Change on Next Login for Local Users


thimomulder:
## Summary
Add the ability to force a local user to change their password upon next login.
***
## Motivation
This is a common requirement when:
- Creating new accounts
- Resetting compromised passwords
- Enforcing periodic credential hygiene

Without this, administrators must rely on manual communication and trust.

### Proposed Solution
When creating or updating a local user, provide an option:
- `Force password change at next login`

Behavior:
- User must set a new password before accessing the system
- Existing password becomes unusable after the first login

### Component

None

### Pre-submission Checklist

- [x] I have searched existing issues and feature requests to ensure this isn't a duplicate
- [x] I have reviewed the documentation to confirm this feature doesn't already exist

Issue #679, opened 2026-04-10
Configurable Password Expiry Notifications for Local Users


thimomulder:
## Summary
Add the ability to notify local users of upcoming password expiration at configurable intervals.
***
## Motivation
Password expiration without advance warning creates poor user experience and increases support workload. Proactive notifications reduce friction while maintaining security.

### Proposed Solution
Allow configuration of notification intervals (in days) before password expiration. This should be configured in the Web UI/REST-API

Behavior:
- Users receive notifications at the specified intervals
- Notification delivery uses existing mail/notification mechanisms
- Notifications stop once the password is changed

### Component

None

## Benefits
- Improved user experience
- Fewer account lockouts and support requests
- Predictable and transparent password lifecycle

### Pre-submission Checklist

- [x] I have searched existing issues and feature requests to ensure this isn't a duplicate
- [x] I have reviewed the documentation to confirm this feature doesn't already exist

PR #705, opened 2026-04-12
feat: add configurable password policies for local users (#677)

### LINKED ISSUE #677: Advanced Password Policies for Local Users
## Summary
Add support for advanced, configurable password policies for local users, including validation via a custom regular expression.
***
## Motivation
Basic password rules (minimum length, simple complexity) are often insufficient for environments with higher security requirements. Being able to define password policies via regex allows administrators to enforce:
- Stronger security guarantees
- Compliance with organizational or regulatory standards
- Consistent password validation across environments

#### Basic solution
Allow administrators to define a password validation based on an checklist, where they can checkboxes to be active as policy. For example:
- lowercase letters
- uppercase letters
- digits
- special characters (This should be an input field to specify which special characters are allowed to be used.)

For every mentioned option above, the administrator should be able to set an amount of times it needs to be used. There should also be the possibility to set a minimum and maximum password length. Also this needs to be an input field.

Display a list with the requirements, and when one of those is met, that one should display a green checkmark. When the password doesn't meet the requirements, the field has an red cross mark.

Invalid passwords should be rejected with a clear validation error message.

#### Regex solution
Another proposal is to allow administrators to define a password validation regex that is applied when creating or updating passwords for local users.

Example regex (illustrative):
```text
(?=.{23,99})(?=(._[A-Z]._){3,})(?=(._[a-z]._){3,})(?=(._[\d]._){3,})(?=(._([!@#$%&-_+=?])._){3,})
```

This example enforces:
- Length between 23 and 99 characters
- At least:
    - 3 lowercase letters
    - 3 uppercase letters
    - 3 digits
    - 3 special characters

When users change their password, it should be possible to display a customizable message.

Invalid passwords should be rejected with a customizable message (maybe the same message as before).

### Alternatives Considered

_No response_

### Use Case

_No response_

### Component

None

## Backward Compatibility
If no regex is configured, existing password validation rules should remain in effect.

### Pre-submission Checklist

brandonrc:
## Summary

Add configurable password policy validation for local user accounts, closing #677. Administrators can now enforce password complexity through environment variables:

- `PASSWORD_MIN_LENGTH` / `PASSWORD_MAX_LENGTH` for length bounds
- `PASSWORD_REQUIRE_UPPERCASE`, `PASSWORD_REQUIRE_LOWERCASE`, `PASSWORD_REQUIRE_DIGIT`, `PASSWORD_REQUIRE_SPECIAL` for character class requirements
- `PASSWORD_MIN_STRENGTH` (0-4) for zxcvbn-based entropy scoring

A new `password_policy` service module contains a pure `validate_password` function that returns all violations at once, so users see every failing rule in a single response rather than fixing them one at a time.

The policy is enforced on user creation (admin endpoint) and password change. Admin password reset is excluded because it generates a temporary password with `must_change_password=true`, meaning the user will be forced to set a compliant password on their next login.

All defaults match the existing behavior (min 8, max 128, no character class requirements, no strength scoring), so this is a non-breaking change.

This is the foundation for #683 (force change), #687 (expiration), #688 (history), and #679 (notifications).

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Closes #677

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -315,6 +315,17 @@ JWT_EXPIRATION_SECS=86400
 # ALLOW_LOCAL_ADMIN_LOGIN=false
 
 # -----------------------------------------------------------------------------
+# Password policy (backend, local users only)
+# -----------------------------------------------------------------------------
+# PASSWORD_MIN_LENGTH=8               # Minimum password length (default: 8)
+# PASSWORD_MAX_LENGTH=128             # Maximum password length (default: 128)
+# PASSWORD_REQUIRE_UPPERCASE=false    # Require at least one uppercase letter
+# PASSWORD_REQUIRE_LOWERCASE=false    # Require at least one lowercase letter
+# PASSWORD_REQUIRE_DIGIT=false        # Require at least one digit
+# PASSWORD_REQUIRE_SPECIAL=false      # Require at least one special character
+# PASSWORD_MIN_STRENGTH=0             # zxcvbn strength score: 0=disabled, 1-4=increasingly strict
+
+# -----------------------------------------------------------------------------
 # Edge Node (for edge binary, not the main backend)
 # -----------------------------------------------------------------------------
 # PRIMARY_URL=https://primary.example.com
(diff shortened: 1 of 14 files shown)


brandonrc:
Superseded by rebased version with merged Config fields.

PR #712, opened 2026-04-12
feat: prevent password reuse by tracking password history

### LINKED ISSUE #688: Prevent Password Reuse by Tracking Password History

brandonrc:
## Summary

Closes #688. Adds password history tracking so local users cannot reuse recent passwords when changing their password.

- New `password_history` table (migration 074) stores bcrypt hashes of previously used passwords per user, indexed by `(user_id, created_at DESC)`.
- New `PASSWORD_HISTORY_COUNT` env var (default 0, disabled). When set to N > 0, the `change_password` handler checks the new password against the current hash and the last N history entries using `bcrypt::verify`, rejecting matches with a validation error.
- On successful password change or admin password reset, the old hash is inserted into the history table and entries beyond the retention count are trimmed.
- Unit tests cover bcrypt-based reuse detection (match, no-match, multi-hash scan) and config parsing for the new env var.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc (commented):
## Code Review: Password History Tracking (#712)

### MUST FIX

**1. No transaction around check-then-update creates a TOCTOU race**

`check_password_history` verifies the new password against stored hashes, then control returns to `change_password` which hashes the new password, runs an UPDATE, and calls `record_password_history` as three separate database operations with no wrapping transaction. Two concurrent password change requests for the same user could both pass the history check before either records the old hash, allowing the same password to be set twice in quick succession. The history and update operations should be wrapped in a `sqlx::PgPool::begin()` transaction. This also protects against the case where the UPDATE succeeds but `record_password_history` fails, leaving the old hash unrecorded.

`change_password` at the new code near line 826, `record_password_history` near line 864, and the same pattern in `reset_password` near line 955.

**2. Redundant `SELECT password_hash FROM users` queries**

In the non-admin `change_password` path, the current password hash is fetched three separate times:
- Line 804: `sqlx::query!("SELECT password_hash FROM users WHERE id = $1", id)` to verify the current password.
- Inside `check_password_history`: another `SELECT password_hash FROM users WHERE id = $1` to compare against the new password.
- The new `old_hash_row` fetch: yet another `SELECT password_hash FROM users WHERE id = $1` to save it before the UPDATE.

The hash from line 804 (already in `hash`) should be passed into `check_password_history` (via an optional `current_hash: Option<&str>` parameter) and reused as `old_hash_row`. That eliminates two unnecessary round trips and also reduces the TOCTOU window.

---

### SHOULD FIX

**3. bcrypt::verify is called sequentially with no concurrency bound, but `spawn_blocking` is used per call**

Each `AuthService::verify_password` call spawns a blocking task via `tokio::task::spawn_blocking`. With `history_count=10`, that's up to 11 `spawn_blocking` calls (current + 10 history). These run sequentially in the `for h in hashes_to_check` loop, but each one occupies a thread from the blocking pool. The bigger concern: if multiple users change passwords concurrently, the blocking thread pool could saturate (default pool is 512 threads, but 10 users changing passwords simultaneously = 110 blocking threads just for history checks).

This is probably fine for real-world usage, but consider:
- Early-exit is already implemented (returns on first match), which helps.
- For very large `history_count` values (e.g., someone sets it to 100), add a cap. An `assert!(history_count <= 50)` or config validation in `Config::from_env` would prevent accidentally hammering CPU.

**4. The trimming query in `record_password_history` uses a correlated NOT IN subquery**

```sql
DELETE FROM password_history
WHERE user_id = $1
  AND id NOT IN (
      SELECT id FROM password_history
      WHERE user_id = $1
      ORDER BY created_at DESC
      LIMIT $2
  )
```

This works but is Postgres-specific in its behavior with `LIMIT` inside a subquery. A cleaner approach that's more portable and avoids scanning the table twice:

```sql
DELETE FROM password_history
WHERE id IN (
    SELECT id FROM password_history
    WHERE user_id = $1
    ORDER BY created_at DESC
    OFFSET $2
)
```

This is functionally equivalent and simpler. It also avoids potential planner confusion with the double reference to `user_id = $1`.

**5. Admin-initiated `change_password` is subject to history checks, but `reset_password` is not**

When an admin calls the `change_password` endpoint for another user (which is allowed per the auth logic), the new password is checked against that user's history. But the `reset_password` endpoint (also admin-only) generates a random password and does NOT check history, only recording. This asymmetry is probably fine in practice (the random password is unlikely to collide), but it should be documented. If an admin explicitly sets a known password via `change_password`, the history check blocks them, which may be surprising.

Consider adding a comment near the `check_password_history` call explaining that admin-initiated changes intentionally go through the same check.

---

### CONSIDER

**6. `history_count` semantics: the effective reuse window is `history_count + 1`**

`check_password_history` fetches `LIMIT history_count` rows from the history table AND checks the current password. So with `PASSWORD_HISTORY_COUNT=5`, a user cannot reuse their current password or any of the last 5 previous ones (6 total). The env var documentation says "reject passwords that match any of the last N passwords," which could be read as excluding the current one. Consider whether this off-by-one is intentional. If so, document it explicitly. If the intent is "N total including current," the LIMIT should be `history_count - 1`.

**7. The migration should add a comment on `password_hash` column confidentiality**

The `password_history.password_hash` column stores bcrypt hashes of old passwords. While bcrypt is resistant to brute force, these are sensitive values. Consider adding `COMMENT ON COLUMN password_history.password_hash IS 'bcrypt hash - never expose via API';` to the migration for documentation, and adding the table to any database access audit scope.

**8. Missing integration/E2E tests**

The unit tests verify bcrypt matching logic in isolation, which is good. But there are no tests that exercise the actual `check_password_history` or `record_password_history` functions against a database. The key scenarios that need integration test coverage:
- Changing password to a previously used password is rejected.
- Changing password to a fresh password succeeds and the old hash appears in `password_history`.
- Reducing `PASSWORD_HISTORY_COUNT` from 10 to 5 trims entries on the next password change.
- `PASSWORD_HISTORY_COUNT=0` truly skips all history logic (no rows inserted, no checks).
- User deletion cascades and removes all `password_history` rows.
- The `reset_password` flow records history without checking.

**9. `test_password_history_count_default` test is a no-op**

The test at the bottom of `users.rs` tests asserts `0 == 0`. It doesn't actually construct a `Config` or test the default. The config tests in `config.rs` do test `env_parse` properly, so this test in `users.rs` can be removed without losing coverage.

---

### LOOKS GOOD

- Migration 074: clean schema, proper CASCADE on `user_id`, correct composite index for the query pattern.
- `check_password_history` checks the current hash in addition to historical entries, preventing reuse of the current password.
- `record_password_history` trims old entries after inserting, so the table stays bounded.
- The config default of 0 (disabled) is safe and explicit.
- `reset_password` correctly records the old hash before overwriting so admin resets don't create gaps in history.
- The error message ("Password was used recently") is intentionally vague, not revealing which specific old password matched.
- All test config structs updated with the new field.

brandonrc (commented):
## Security Review: PRs #712, #713, #714

Consolidated security review covering password history (#712), password expiration (#713), and quarantine period (#714).

---

### Credential Handling

**PR #712 (password history):** Password hashes in the `password_history` table are stored as bcrypt hashes, which is correct. The `password_hash` field on the `User` struct already has `#[serde(skip_serializing)]`, so it will not leak via JSON serialization. The `password_history` table is not exposed through any API endpoint or response struct, so history hashes have no serialization path to the client. The `record_password_history` function stores the old bcrypt hash (not the plaintext), and the trimming logic correctly bounds the table size. This is well done.

One concern: the `password_history` table has no `serde(skip)` protection because it never becomes a Rust struct that is serialized, but if someone later adds an admin endpoint that queries this table, the hashes could leak. Consider adding a code comment to the table/query noting that these hashes must never be included in API responses.

**PR #713 (password expiration):** The `password_changed_at` field is added to the `User` struct and IS serialized to API responses (no `#[serde(skip_serializing)]`). While `password_changed_at` is not itself a secret, exposing it tells an attacker exactly when a user's password expires (if they know the `PASSWORD_EXPIRY_DAYS` value, which is in `.env.example`). This is a low-severity information disclosure.

**Pre-existing issue (not from these PRs):** `totp_secret` and `totp_backup_codes` on the `User` struct lack `#[serde(skip_serializing)]`. These are actual secrets that can leak via any endpoint returning a `User` (list users, get user, create user, update user). This predates these PRs but is Critical severity.

### Authentication & Authorization

**PR #713 (password expiration bypass via SSO/API tokens):** The expiration check runs inside `AuthService::authenticate` (the username+password login path). When `is_password_expired` returns true, the code sets `user.must_change_password = true` *in memory* on the returned user struct, but it does NOT persist this to the database. This means:

1. API token authentication (`validate_api_token`) does not check `is_password_expired`. A user with an expired password can continue using API tokens indefinitely without changing their password. This is likely intentional for CI/CD workflows, but should be documented explicitly.

2. SSO users (OIDC, SAML, LDAP) are not affected because they have no local password. The `password_changed_at` column defaults to `NOW()` on creation, and the expiry check only fires in the local auth path. This is correct.

3. The `must_change_password` flag is set in-memory only, not written to the database. If the client ignores the flag in the JWT/response, the user can keep authenticating with the expired password and getting fresh JWTs. The server-side enforcement relies on the client honoring `must_change_password` in the token. Consider persisting the flag to the database when password expiration is detected, so subsequent token refreshes also enforce it.

**PR #714 (quarantine authorization):** The `release_artifact` and `reject_artifact` endpoints both call `auth.require_admin()`, which correctly gates these operations to admin users. A regular authenticated user can read quarantine status (GET `/quarantine/:artifact_id`) but cannot release or reject. This is the right pattern.

However, the `get_quarantine_status` endpoint only checks that authentication exists (`auth.ok_or_else(...)`) but does not verify that the user has read access to the artifact's repository. If a repository is private, any authenticated user can query the quarantine status of artifacts in that repository by guessing/knowing artifact IDs. The endpoint should check repository visibility, similar to how `download_artifact` calls `require_visible`.

### Input Validation

**PR #712 (password_history_count):** Uses `env_parse::<u32>()`, which will reject negative values at the type level (u32 is unsigned). Invalid strings fall back to default 0. No upper bound is enforced, so setting `PASSWORD_HISTORY_COUNT=999999` is accepted. With that value, every password change would run ~1M bcrypt verify operations, which takes on the order of minutes per password change and could be used as a DoS vector if an attacker gains admin access to env vars. Consider capping at a reasonable maximum (e.g., 24, which NIST and most enterprise policies recommend as an upper limit).

**PR #713 (password_expiry_days):** Also uses `env_parse::<u32>()`, so negatives are rejected by the type. No upper bound, but extremely large values (e.g., `u32::MAX` = 4 billion days) would effectively disable the feature, which is the same as setting 0. No security risk here beyond configuration mistakes.

**PR #714 (quarantine_duration_minutes):** The global config uses `env_parse::<i64>()` (signed), so negative values are accepted. A negative `QUARANTINE_DURATION_MINUTES` would set `quarantine_until` to a time in the past, meaning the quarantine is immediately expired and artifacts are never actually held. This silently defeats the purpose of the feature. The per-repo update in `repositories.rs` does validate `duration < 0` and returns an error, but the global env var path does not. Additionally, a zero value creates a quarantine that expires immediately, which is also a silent no-op.

The `resolve_config` function in `quarantine_service.rs` reads `QUARANTINE_ENABLED` and `QUARANTINE_DURATION_MINUTES` directly from `std::env::var` at call time, bypassing the `Config` struct. This means the config is re-read from the environment on every upload. While this enables dynamic reconfiguration, it also means a process with env var injection could toggle quarantine on or off without a restart, which is a subtle attack surface if the process environment is not locked down.

### Timing Attacks

**PR #712 (bcrypt history check):** The `check_password_history` function iterates over all history hashes and calls `AuthService::verify_password` (bcrypt verify) for each one. If the password matches the first hash, the function returns an error immediately. If it does not match any, it checks all hashes before returning success. This creates a measurable timing difference:

- Match on hash 1: ~100ms (1 bcrypt verify)
- Match on hash 5: ~500ms (5 bcrypt verifies)
- No match (history_count=12): ~1200ms (12+ bcrypt verifies)

An attacker who can measure response times can determine (a) whether the password was previously used, and (b) roughly how recently it was used (earlier position = more recent). In practice, this is Low severity because bcrypt's inherent ~100ms cost makes precise timing difficult over the network, and the attacker would need to know the user's proposed new password. But for defense-in-depth, consider always checking all hashes regardless of early match (accumulate a boolean result without short-circuiting).

### OWASP Alignment

These features align well with OWASP ASVS v4.0 and NIST SP 800-63B:

- **ASVS 2.1.7 (password history):** PR #712 implements this. NIST recommends checking against a list of commonly used passwords rather than forcing history-based rotation, but history checking is still a valid defense against credential reuse. The bcrypt comparison approach is correct (no plaintext storage).

- **ASVS 2.1.8 (password length/complexity):** Not in scope for these PRs, but the existing validation (`at least 8 characters`) is noted in the test output.

- **ASVS 2.1.9 (password change):** PR #713 implements password expiration. Note that NIST SP 800-63B (Section 5.1.1.2) explicitly recommends AGAINST periodic password expiration ("Verifiers SHOULD NOT require memorized secrets to be changed arbitrarily"), citing that it leads to weaker passwords. The feature defaults to disabled (0 days), which is the right default, and the documentation should note the NIST guidance.

- **ASVS 14.4.1 (security headers):** Not in scope, pre-existing.

---

### Findings

**MUST FIX**

1. **[#714] Quarantine status endpoint missing repository visibility check.** `get_quarantine_status` in `quarantine.rs` only checks authentication, not authorization against the repository. Any authenticated user can query quarantine status for artifacts in private repositories. Add a repository visibility check mirroring `require_visible` in `download_artifact`. Severity: Medium.

2. **[#714] Negative `QUARANTINE_DURATION_MINUTES` silently disables quarantine.** The global env var is parsed as `i64` with no lower-bound validation. Negative values create a `quarantine_until` in the past, making quarantine a no-op. Add a `max(0, value)` clamp or reject negative values in `Config::from_env()`. Also consider rejecting 0 or warning in logs, since zero-minute quarantine is functionally disabled. Severity: Medium.

**SHOULD FIX**

3. **[#713] Password expiration not persisted to database.** `is_password_expired` sets `must_change_password = true` in-memory only. The flag is included in the JWT, but a client that ignores it can keep re-authenticating with the expired password. Persist the flag to the database when expiration is detected, so the next auth attempt also sees `must_change_password = true` from the DB. Severity: Medium.

4. **[#712] Bcrypt history check leaks match position via timing.** The `for` loop in `check_password_history` short-circuits on the first match, creating timing variance proportional to the match position. Always iterate all hashes and accumulate the result. Severity: Low.

5. **[#712] No upper bound on `PASSWORD_HISTORY_COUNT`.** A very large value (e.g., 100,000) would cause extreme CPU load on every password change. Cap at a sane maximum (24 is the common enterprise ceiling). Severity: Low.

**CONSIDER**

6. **[#714] `resolve_config` reads env vars on every call.** `quarantine_service::resolve_config` reads `QUARANTINE_ENABLED` and `QUARANTINE_DURATION_MINUTES` from the process environment directly rather than using the `Config` struct, which is read once at startup. This means the quarantine config can change without a restart if something writes to the process environment. Consider using `state.config.quarantine_enabled` / `state.config.quarantine_duration_minutes` as the fallback instead of re-reading env vars.

7. **[#714] Quarantine failure during upload is silently swallowed.** In `artifact_service.rs`, the quarantine `set_quarantine` call uses `.ok()` to ignore errors. If the DB update fails, the artifact is uploaded without quarantine and becomes immediately downloadable. This is documented as "Non-fatal" in the comment, but in a security-critical deployment, a failed quarantine should probably block the upload rather than skip the hold. Consider making this configurable or at least emitting a warning-level log (currently info-level).

8. **[Pre-existing] `totp_secret` and `totp_backup_codes` leak via User serialization.** These fields on the `User` struct lack `#[serde(skip_serializing)]`, unlike `password_hash` which correctly has it. Any endpoint returning `Json<User>` will include TOTP secrets in the response. This is not introduced by these PRs but is Critical severity and should be tracked separately.

PR #713, opened 2026-04-12
Add password expiration for local users

brandonrc:
## Summary

Implements configurable password expiration for local user accounts (#687). Administrators can set `PASSWORD_EXPIRY_DAYS` to require password changes after a defined interval, satisfying security standards that mandate periodic password rotation.

When a user logs in and their password is older than the configured threshold, `must_change_password` is set to `true` in the login response, prompting the frontend to redirect to the password change flow. Setting `PASSWORD_EXPIRY_DAYS=0` (the default) disables the feature entirely, preserving existing behavior.

Key changes:
- Migration 074 adds `password_changed_at TIMESTAMPTZ DEFAULT NOW()` to the users table
- `AuthService::is_password_expired()` provides a pure, testable expiry check
- The `authenticate()` method evaluates password age after successful credential verification
- Both `change_password` and `reset_password` handlers update `password_changed_at`
- All User struct SELECT/RETURNING queries include the new column
- SQLx offline cache updated (24 query files)

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc (commented):
## Code Review: Password Expiration for Local Users

### LOOKS GOOD: Security -- expiry check runs after credential verification

The expiry check is placed at line 321-334 of `auth_service.rs`, after password verification succeeds and after lockout counters are reset. This is the correct order. An unauthenticated caller cannot probe whether an account exists or has an expired password, since they get the same "Invalid username or password" error regardless.

### LOOKS GOOD: SSO users excluded from password expiry

The `authenticate()` method already guards `user.auth_provider != AuthProvider::Local` and returns early with "Use SSO provider to authenticate" before reaching the expiry check. LDAP, OIDC, and SAML users go through their own login paths (`authenticate_ldap`, `handle_oidc_callback`, `handle_saml_response`) which never call `is_password_expired`. The feature is correctly scoped to local users only.

### LOOKS GOOD: Migration default behavior

`password_changed_at TIMESTAMPTZ NOT NULL DEFAULT NOW()` is correct. Existing users get `NOW()` at migration time, which means they won't be immediately forced to change passwords when an admin first enables the policy. This is the right default -- otherwise turning on a 90-day policy would lock out every existing user simultaneously.

### LOOKS GOOD: Time handling

Both `password_changed_at` and the comparison use `DateTime<Utc>` throughout. The DB column is `TIMESTAMPTZ`. The `is_password_expired` function takes an explicit `now` parameter, eliminating any ambiguity about which clock is used and making it testable. Clock skew between app servers could cause inconsistent behavior across instances, but that's inherent to any time-based check and not unique to this implementation.

### LOOKS GOOD: `is_password_expired` boundary tests

Six tests cover the key cases:
- Disabled when `password_expiry_days == 0` (even with a very old password)
- Not expired within window (10 days into a 90-day policy)
- Expired after window (91 days into a 90-day policy)
- Exactly on boundary (90 days == expired, consistent with `now >= expiry`)
- Just changed (same timestamp, not expired)
- Sub-day granularity (25 hours into a 1-day policy)

The `>=` comparison makes the boundary inclusive (expired at exactly N days), which matches the test comment and is the safer choice.

### LOOKS GOOD: All query updates consistent

All 24 `.sqlx` query files are updated. Every `SELECT`/`RETURNING` that fetches a `User` struct now includes `password_changed_at` between `last_failed_login_at` and `last_login_at`. The column ordering is consistent across all queries. The `change_password` and `reset_password` handlers both set `password_changed_at = NOW()` alongside `updated_at = NOW()`, which is correct.

### CONSIDER: `must_change_password` is set in-memory only, not persisted to DB on expiry

When password expiry fires, the PR sets `user.must_change_password = true` on the in-memory `User` before returning it, but does not write this to the database. This means:

1. The login response correctly tells the client `must_change_password: true`.
2. But if the user ignores the client-side redirect and makes API calls with their valid JWT, the server won't enforce the password change on subsequent requests (there's no middleware checking this flag).
3. On every subsequent login, the expiry check re-fires and sets the in-memory flag again.

This is an acceptable design if the frontend is the enforcement point, but worth being explicit about in the PR description. If server-side enforcement is desired in the future, consider persisting `must_change_password = true` to the DB when expiry triggers, or adding middleware that rejects requests from users with expired passwords.

### CONSIDER: No upper-bound validation on `PASSWORD_EXPIRY_DAYS`

The config field is `u32`, which allows values up to ~4 billion days (~11 million years). A typo like `PASSWORD_EXPIRY_DAYS=9000` (instead of 90) would silently disable the policy in practice. Adding a reasonable upper-bound warning in the config validation (e.g., log a warning if > 365) could help catch configuration mistakes, though this is minor.

Overall this is a clean, well-scoped PR. The pure function approach for testability is good, the security ordering is correct, and the migration is safe.

PR #706, opened 2026-04-12
feat: quarantine period with per-repo configuration (#472)

brandonrc:
## Summary

Implements #472: quality gate quarantine period for newly published artifacts, with per-repo configuration.

When quarantine is enabled, newly uploaded artifacts are held in a 'quarantined' state until security scans complete or the timeout expires. During quarantine, download requests return 409 Conflict.

### Global configuration (env vars)

- `QUARANTINE_ENABLED` (default: false) - global toggle
- `QUARANTINE_DURATION_MINUTES` (default: 60) - global default hold duration

### Per-repo configuration (API)

Repos can override the global defaults via PATCH `/api/v1/repositories/{key}`:
\`\`\`json
{ "quarantine_enabled": true, "quarantine_duration_minutes": 120 }
\`\`\`

Per-repo settings are stored in the existing \`repository_config\` table and resolved at upload time, falling back to the global config when no override exists.

### Implementation

- Migration: extends artifacts table with quarantine states + quarantine_until timestamp
- Quarantine service: pure-function decision logic, per-repo config resolution
- Artifact service: sets quarantine on upload, checks on download, auto-releases on expiry
- Scanner integration: transitions to released/rejected on scan completion
- API: 3 quarantine endpoints (status, manual release, manual reject)
- Generic \`upsert_repo_config\` helper extracted from existing code

### Tests

15 unit tests covering all quarantine states, expiry logic, config resolution, and edge cases.

## Test Checklist
- [x] Unit tests added/updated
- [ ] Integration tests added/updated (if applicable)
- [ ] E2E tests added/updated (if applicable)
- [x] Manually tested locally (7384 tests pass)
- [x] No regressions in existing tests

## API Changes
- [x] New endpoints have \`#[utoipa::path]\` annotations
- [x] Request/response types have \`#[derive(ToSchema)]\`
- [x] OpenAPI spec validates
- [x] N/A - migration adds columns with defaults, fully backwards compatible

Closes #472

Issue #472, opened 2026-03-20
Add Quality Gate Quarantine Period for Newly Published Artifacts


thimomulder:
### Problem Statement

Recent supply‑chain attacks—most notably those targeting the NPM ecosystem—have shown that malicious actors are increasingly publishing compromised or hijacked packages that appear legitimate at first glance.  
Because Artifact‑Keeper mirrors, stores, or interacts with external registries, it is possible for malicious or compromised artifacts to be ingested before security scanners, signature verifiers, or community reports identify the threat.

**This feature addresses the risk of immediately trusting newly published upstream artifacts.**

Without a quarantine mechanism, Artifact‑Keeper may inadvertently surface or distribute harmful dependencies before scanners have time to analyze them and before the broader ecosystem identifies issues.

### Proposed Solution

Introduce a **configurable quarantine period** for newly discovered artifacts retrieved from external registries (e.g., NPM, Maven, NuGet, PyPI, etc.). The quarantine flow would include:
1. **Detection Phase**:
	- When a new version of an artifact is discovered, flag it as “Pending Verification.”
2. **Quarantine Period (Configurable: e.g., 24–72 hours)**
    - Artifact is stored but **not yet available for consumption**.
3. **Scan artifact after quarantine period**
    - Security scanners run (SAST, dependency checks, signatures, metadata validation).
    - Optional integration with external threat feeds or reputation services.
4. **Verification**
    - If scans and checks pass the artifact becomes fully available.
    - If scans fail or reputation checks raise flags, the artifact remains blocked and marked as “Rejected.”

### Use Case

**User Story:**  
As a DevOps engineer responsible for secure CI/CD pipelines,  
I want Artifact‑Keeper to quarantine newly published upstream artifacts  
so that my developers do not unintentionally use compromised packages  
that have not yet undergone security verification.

**Example Scenario:**
- A malicious attacker publishes a compromised NPM package version.
- Artifact‑Keeper detects the new version but places it in quarantine.
- After the quarantine period, security scanners identify suspicious obfuscation or indicators of compromise.
- The artifact is blocked, preventing the compromised package from ever entering production pipelines.

This quarantine period provides essential buffer time for security tools and the community to detect attacks, significantly reducing risks, which are not supported by Nexus or JFrog. For enterprise users this is a big needed feature.

### Component

Security Scanning

### Pre-submission Checklist

- [x] I have searched existing issues and feature requests to ensure this isn't a duplicate
[...]

PR #714, opened 2026-04-12
feat: quarantine period for uploaded artifacts (#472)

brandonrc:
## Summary

Implements the quarantine period feature (issue #472). When enabled (globally via `QUARANTINE_ENABLED` or per-repo via `repository_config`), newly uploaded artifacts are held in a "quarantined" state for a configurable duration. Downloads return 409 Conflict while an artifact is quarantined.

This is a clean reimplementation of the closed PR #706, with the migration number updated to avoid conflicts.

Key changes:
- Migration 074 extends the existing `quarantine_status` CHECK constraint with three new values (`quarantined`, `released`, `rejected`) and adds a `quarantine_until` TIMESTAMPTZ column
- New `quarantine_service` module with pure-function decision logic for quarantine checks, config resolution, and state transitions
- Three new API endpoints under `/api/v1/quarantine/`: GET status, POST release (admin), POST reject (admin)
- Upload path automatically quarantines artifacts when the feature is enabled for the repository
- Download path checks quarantine status and blocks with 409 if quarantined or rejected
- Scanner integration transitions quarantined artifacts to `released` (clean scan) or `rejected` (findings found) on scan completion
- Per-repo overrides via `quarantine_enabled` and `quarantine_duration_minutes` fields on `UpdateRepositoryRequest`, stored in the existing `repository_config` table
- Generic `upsert_repo_config` helper extracted from the existing `upsert_index_upstream_url` function

## Test Checklist
- [x] Manually tested locally

## API Changes
- [ ] N/A - no API changes


brandonrc (commented):
## DevOps Review: PRs #712, #713, #714

Reviewed all three PRs together since they share migration number 074 and will need coordination before merging.

### Migration Safety

**Critical: All three PRs use migration number 074.** They cannot all merge as-is. The migration runner will execute whichever file it finds first and ignore (or fail on) the others. These need to be rebased onto sequential numbers (074, 075, 076) in merge order.

**PR #712 (password_history):**
- `CREATE TABLE password_history` with no `IF NOT EXISTS` guard. If the migration is partially applied (table created, index not yet), re-running will fail on the `CREATE TABLE` statement. Adding `IF NOT EXISTS` to both the table and index would make it safe to retry.
- Creates a new table rather than altering an existing one, so there is zero locking risk. Good.

**PR #713 (password_changed_at):**
- `ALTER TABLE users ADD COLUMN IF NOT EXISTS password_changed_at TIMESTAMPTZ NOT NULL DEFAULT NOW()` -- the `IF NOT EXISTS` guard is present, which is good for idempotency.
- However, `NOT NULL DEFAULT NOW()` on the `users` table will rewrite every row on PostgreSQL < 16 (pg16 introduced fast default for non-null columns). On a large users table this takes an `ACCESS EXCLUSIVE` lock. Confirm the production PG version is >= 16, or split into two steps: add the column as nullable, backfill, then set NOT NULL.
- Existing users will all get `password_changed_at = NOW()` at migration time, meaning their password age clock resets. This may be intentional (prevents mass forced rotation on deploy), but worth documenting.

**PR #714 (quarantine_period):**
- `ALTER TABLE artifacts DROP CONSTRAINT IF EXISTS artifacts_quarantine_status_check` followed by `ADD CONSTRAINT` -- safe pattern with `IF EXISTS` guard on the drop.
- `ADD COLUMN IF NOT EXISTS quarantine_until TIMESTAMPTZ` -- nullable column addition, instant operation, no table rewrite. Safe.
- `CREATE INDEX IF NOT EXISTS idx_artifacts_quarantine_until` with a partial index filter. This is a concurrent-safe guard. However, since this is not `CREATE INDEX CONCURRENTLY`, it will take a short lock on the `artifacts` table while building. On a table with millions of rows, consider using `CONCURRENTLY` (though SQLx migrations run in transactions, which is incompatible with `CONCURRENTLY`, so this is a known trade-off).

### Configuration

**PR #712:** `PASSWORD_HISTORY_COUNT` documented in `.env.example`, defaults to 0 (disabled). Production-safe default. No risk of outage from misconfiguration since the feature is opt-in.

**PR #713:** `PASSWORD_EXPIRY_DAYS` documented in `.env.example`, defaults to 0 (disabled). Production-safe default. One concern: the `resolve_config` function in the quarantine service reads `QUARANTINE_ENABLED` directly from `std::env::var` at runtime rather than from the `Config` struct. This means the env var is read on every upload, which works but bypasses the centralized config validation. The `Config` struct also reads it at startup. These two paths should stay consistent, worth a comment.

**PR #714:** `QUARANTINE_ENABLED` and `QUARANTINE_DURATION_MINUTES` documented in `.env.example`, defaults to `false` and `60`. Production-safe defaults. Per-repo overrides via `repository_config` are a solid pattern. One thing to watch: `quarantine_service::resolve_config` reads env vars on every call (every upload). This is fine for correctness but means config changes take effect without restart for the quarantine service while other settings require a restart. Consider reading from the `Config` struct instead for consistency, or document this as intentional hot-reload behavior.

### Rollback

**PR #712:** The `password_history` table is independent and referenced only by the new handler code. Rolling back the binary is clean since old code never touches this table. The table can be left in place or dropped manually. The new `password_history_count` config field defaults to 0, so if old code somehow had the field, it would be harmless. Good rollback story.

**PR #713:** The `password_changed_at` column is added as `NOT NULL DEFAULT NOW()`. Rolling back the binary would cause old code to fail on every User query because it does not include `password_changed_at` in its SELECT lists, and the old SQLx offline cache does not expect it. This is the standard trade-off with schema-forward/code-backward: the old binary is incompatible with the new schema. If you need to roll back, you would need to deploy the previous binary AND either drop the column or keep the old column around. Since the column is NOT NULL, you cannot make old code ignore it unless queries use `SELECT *` (which they don't). **Recommendation:** Consider making the column nullable (`DEFAULT NULL`) so old code can function with a schema that has the extra column but does not select it.

**PR #714:** The new `quarantine_until` column is nullable, so old code that does not select it will still work at the database level. However, old code's artifact queries do not include `quarantine_status, quarantine_until` in their SELECT lists, so the old SQLx offline cache is incompatible. Same rollback concern as #713: the old binary must match the old schema. The CHECK constraint update from `(unscanned, clean, flagged)` to `(unscanned, clean, flagged, quarantined, released, rejected)` is additive and backward-compatible.

**General rollback note:** All three PRs modify the SQLx offline cache (`.sqlx/` files). Rolling back the binary requires rolling back the `.sqlx/` cache in sync, which happens naturally if you deploy a previous git SHA. The database schema, however, stays forward. Consider adding down-migrations (even if just documented) for all three.

### Monitoring

**PR #712:** No explicit log lines when a password reuse is rejected. The handler returns `AppError::Validation(...)` which will produce a 400 response, but there is no `tracing::warn!` or `tracing::info!` for password reuse attempts. From a security auditing perspective, password reuse attempts should be logged with user ID. **Recommendation:** Add a `tracing::warn!` in `check_password_history` when a match is found, before returning the error.

**PR #713:** No explicit log line when a password is detected as expired. The `authenticate` method silently sets `must_change_password = true` without logging. **Recommendation:** Add a `tracing::info!` when password expiry triggers, including the user ID and how many days past expiry.

**PR #714:** Good observability. The quarantine handler logs `tracing::info!` for both release and reject operations, including artifact ID and admin username. The upload path logs quarantine application with artifact ID and expiry time. The scanner integration is implicit (scan results drive state transitions) but does not log the quarantine state transition itself. **Recommendation:** Add a `tracing::info!` in `update_quarantine_status` when a quarantined artifact transitions to released or rejected, including the artifact ID and finding count.

Event bus integration is present in #714 with `artifact.quarantine.released` and `artifact.quarantine.rejected` events. No custom metrics (Prometheus counters) for quarantine or password policy events in any of the three PRs. Consider adding counters for: `password_reuse_rejected_total`, `password_expired_total`, `artifacts_quarantined_total`, `artifacts_released_total`, `artifacts_rejected_total`.

### Docker/CI Impact

None of the three PRs add new Rust crate dependencies. No changes to Cargo.toml, Dockerfile, or CI workflows. Image size is unaffected. The new `.sqlx/` cache file churn is significant (24 renames in #713, 10+ renames in #714) but this is expected when adding columns to core tables and only affects compile-time query validation.

### Overall Assessment

These three PRs form a coherent security and compliance feature set. The code quality is solid with pure-function test patterns in #713 and #714. Key items to address before merging:

1. **Resolve the migration number conflict.** All three use 074. Decide on merge order and rebase accordingly.
2. **Evaluate the `NOT NULL DEFAULT NOW()` in #713** for the users table on your production PG version. If PG < 16, this will lock the table during migration.
3. **Add security audit logging** in #712 and #713 for password reuse rejections and password expiry triggers.
4. **Consider rollback documentation** for all three, noting that schema-forward + binary-rollback is not seamless due to SQLx offline cache coupling.
5. **The quarantine service's dual config path** (#714) reads env vars at runtime independently of the Config struct. Decide if this is intentional hot-reload or should be unified.

brandonrc (commented):
## Code Review: Quarantine Period for Uploaded Artifacts

Overall this is well-structured: the pure-function decision logic is cleanly separated from I/O, the migration is straightforward, and the per-repo config override pattern is solid. There are a few issues to address before merging.

---

### MUST FIX

**1. Format handlers bypass quarantine on download (security)**

The quarantine download check is only in `repositories.rs::download_artifact`. Every native format handler (npm `serve_tarball`, Maven `download`, PyPI `download_or_metadata`, Cargo `download`, etc.) fetches artifacts from storage and serves them directly without calling `check_download_allowed`. An attacker can upload a malicious artifact to a quarantine-enabled repo and immediately download it via the format-native endpoint (e.g. `GET /npm/myrepo/@scope/pkg/-/pkg-1.0.0.tgz`), completely bypassing quarantine.

Fix: either add the quarantine check in `ArtifactService::download()` (which some handlers use), or add a centralized middleware/helper that all download paths call. The check belongs at the lowest common layer so it cannot be skipped.

**2. `transition()` allows any state to move to `released` or `rejected` (state machine)**

The `transition` function and the `release_artifact`/`reject_artifact` handlers do not check the current status before transitioning. An admin can release an already-rejected artifact (going from `rejected` to `released`), or an artifact whose status is `flagged` or `clean` from the proxy-scan workflow. The handler only calls `get_status` to verify the artifact exists, not to validate the current state.

If the intent is to allow admins to override rejections, document it explicitly. If not, add a guard:

```rust
let (current, _) = quarantine_service::get_status(&state.db, artifact_id).await?;
match current.as_deref() {
    Some("quarantined") | Some("rejected") => { /* allowed */ }
    _ => return Err(AppError::Conflict("Artifact is not in quarantine".into())),
}
```

At minimum, `rejected -> released` should require explicit acknowledgement (a `force: true` flag or similar) since it overrides a security scan decision.

**3. Scanner race condition: last writer wins on quarantine status**

In `ScannerService::scan_artifact`, each scanner runs sequentially in a loop and calls `update_quarantine_status` after each one completes. If scanner A finishes clean (0 findings) and sets status to `released`, then scanner B finishes with findings and sets status to `rejected`, the artifact ends up correctly rejected. But if the order is reversed (B runs first, then A), scanner A will read `quarantined` status, see 0 findings, and transition to `released`, overriding B's rejection. The status reflects only the last scanner result, not the worst.

The `update_quarantine_status` method does a SELECT then UPDATE without a transaction or row lock, so even if scanners somehow ran concurrently (e.g. from a requeued scan job), the check-then-set is susceptible to TOCTOU.

Fix: use a single UPDATE with a WHERE clause that prevents downgrading:

```sql
UPDATE artifacts
SET quarantine_status = $2, quarantine_until = CASE WHEN $2 IN ('released','rejected') THEN NULL ELSE quarantine_until END
WHERE id = $1
  AND (quarantine_status != 'rejected' OR $2 = 'rejected')
```

Or accumulate all scan results before making a single status decision.

---

### SHOULD FIX

**4. `set_quarantine` accepts arbitrary strings (type safety)**

`set_quarantine` takes `status: &str` rather than `QuarantineState`. The only caller passes `"quarantined"` as a literal, which is safe today, but future callers could pass an invalid string that would violate the DB CHECK constraint at runtime. Use `QuarantineState` for both `set_quarantine` and `transition` for compile-time safety.

**5. Quarantine applied after INSERT, not atomically (upload race)**

In `artifact_service.rs`, the artifact is INSERTed first, then `set_quarantine` is called as a separate UPDATE. Between the INSERT and the UPDATE, there is a window where the artifact has `quarantine_status = NULL` and is downloadable. Under load, a fast client polling for the artifact could download it in this gap.

Fix: set `quarantine_status` and `quarantine_until` in the INSERT statement itself (or wrap both in a transaction). Resolve the config before the INSERT and include the values in the query.

**6. `quarantine_duration_minutes` has no upper bound validation**

The `update_repository` handler validates `duration < 0` but has no upper bound. A value like `999999999` (1902 years) is accepted. Consider capping at a reasonable maximum (e.g. 43200 = 30 days) to prevent accidental misconfiguration.

**7. `.ok()` on quarantine set swallows errors silently**

In `artifact_service.rs` line ~333: `quarantine_service::set_quarantine(...).await.ok()`. If the DB update fails, the upload succeeds but the artifact is not quarantined. This is the worst outcome: the operator believes quarantine is active but it silently failed. At minimum, log the error. Better: propagate it.

---

### CONSIDER

**8. 409 Conflict response lacks `Retry-After` header**

The 409 response for quarantined artifacts does not include a `Retry-After` header. Since `quarantine_until` is known, you could include it:

```
Retry-After: <seconds-until-quarantine-expires>
```

This helps automated clients (CI pipelines, package managers) decide when to retry instead of hammering the endpoint.

**9. No background job to auto-release expired quarantines**

The `check_download_allowed` function correctly allows downloads when `quarantine_until` has passed, even if the status is still `quarantined`. But the DB status stays stale until the next scan completes. This means the quarantine status in the API response and any admin dashboards will show `quarantined` for artifacts that are actually downloadable. Consider adding a periodic job (similar to the GC scheduler) that transitions expired quarantines to `released`.

**10. Migration does not handle existing `unscanned`/`clean`/`flagged` data**

The migration adds the new CHECK values and the `quarantine_until` column, which is correct. Existing artifacts with `unscanned`, `clean`, or `flagged` status continue to work because `check_download_allowed` treats them as downloadable. This is fine. Just confirming: the migration is backward-compatible and no data transformation is needed.

**11. No integration or E2E tests for quarantine**

The unit tests for pure functions are thorough. However, there are no integration tests verifying the end-to-end flow: upload with quarantine enabled, verify 409 on download, scan completes, verify download succeeds. The test checklist in the PR description also leaves integration/E2E unchecked. Given the security criticality (this is access control), integration tests should be added before merging.

**12. `resolve_config` uses `std::env::var` at call time**

`resolve_config` reads `QUARANTINE_ENABLED` and `QUARANTINE_DURATION_MINUTES` from the environment on every call rather than from `Config`. The values in `Config` (parsed at startup) are not used by this function. This means the global config in `config.rs` and the runtime config in `resolve_config` can diverge if someone changes env vars without restarting. Consider passing `&Config` to `resolve_config` instead of re-reading env vars.

---

### LOOKS GOOD

- Clean separation of pure decision logic from I/O in `quarantine_service.rs`
- Thorough unit tests covering edge cases (expired quarantine, NULL expiry with quarantined status, all legacy statuses)
- Migration is clean and backward-compatible, partial index on `quarantine_until` is well-targeted
- `upsert_repo_config` extraction from the existing `upsert_index_upstream_url` is a good refactor
- Per-repo config resolution with graceful fallback chain (repo config -> env -> defaults)
- Event bus integration for quarantine state changes
- Admin release/reject endpoints properly gated behind `require_admin()`

Issue #729, opened 2026-04-13
OTLP gRPC exporter fails with FRAME_SIZE_ERROR behind Envoy Gateway


Firjens:
> Notice of transparency that I debugged this issue and wrote this issue with the aid of an AI LLM.

### Description

Artifact Keeper's OTLP gRPC exporter is broken in any deployment that uses Envoy Gateway as the ingress. This is a known incompatibility between the Rust h2 crate (used by tonic) and Envoy's strict HTTP/2 SETTINGS enforcement, but since Artifact Keeper doesn't support `OTEL_EXPORTER_OTLP_PROTOCO`L, there is no workaround available.

**What happens:**
The Rust `h2` crate sends data frames immediately after the HTTP/2 SETTINGS frame, without waiting for the server's SETTINGS ACK. This violates RFC 7540 §6.5 which requires endpoints to apply new settings before sending further frames. Envoy enforces this strictly and responds with `GOAWAY FRAME_SIZE_ERROR`, terminating the connection.

**Why this affects Artifact Keeper specifically:**
Other services in our cluster using the same opentelemetry-otlp dependency can work around this by setting OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf, which switches from gRPC/tonic to HTTP/protobuf, bypassing the issue entirely. This feature however doesn't exist in artifact-keeper.

**Evidence:**
- grpcurl (Go gRPC client) → same Envoy endpoint → works
- HTTP/protobuf → same Envoy endpoint → works
- Artifact Keeper tonic → same Envoy endpoint → FRAME_SIZE_ERROR
- Envoy xDS config, access logs, and Otel all confirm the infrastructure is correctly configured

### Environment Information

- Artifact Keeper: latest dev image
- Envoy Gateway: v1.7.1 (Envoy proxy v1.37.1)
- Kubernetes: RKE2

### Backend Version

dev

### Web UI Version

dev

### Package Format

None

### Expected Behavior

Setting `OTEL_EXPORTER_OTLP_ENDPOINT=https://otel-grpc.example.com:443` should result in traces and metrics being successfully exported to the configured collector, consistent with the [OTLP spec](https://opentelemetry.io/docs/specs/otel/protocol/exporter/) which states that https:// endpoints must be supported for gRPC exporters.

[...]

PR #733, opened 2026-04-14
feat: add support for OpenTelemetry over http/protobuf (workaround for #729)

Firjens:
## Summary
Adds support for `http/protobuf` for OTEL in addition to the existing gRPC implementation, following the standard from OTEL: https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol

- Documentation merge request: https://github.com/artifact-keeper/artifact-keeper-site/pull/53

## Test Checklist
- [ ] Manually tested locally

## API Changes
- [ ] N/A - no API changes


Firjens:
I btw propose to merge this into the v1.2.0 branch as this is a new feature, however this branch was not created at the time of the merge request.

brandonrc (commented):
Good addition. A few suggestions to simplify the code:

**1. The `OtlpProtocol` enum is fine, but it should impl `Display` instead of repeating the match for logging.**

The `protocol = match protocol { ... }` block in `init_tracing` is a hand-rolled Display. Adding `impl std::fmt::Display for OtlpProtocol` would let you write `protocol = %protocol` in the tracing call and avoid that second match block entirely. It also future-proofs any other place you might want to log or print the protocol.

**2. `from_env()` reads env vars directly, but the existing pattern in `main.rs` already reads OTel env vars before `Config::from_env()` on purpose** (lines 100-107 of main.rs, with the comment explaining why). Given that telemetry must initialize before Config parsing, reading the env var in the telemetry module is the right call here. No change needed.

That said, the `from_env()` method name is slightly misleading since it's a constructor, not a full config loader. Consider renaming to `from_env_var()` or just inlining the logic at the call site in `init_tracing`, since `from_env()` is only called once and is 6 lines. The enum itself still carries its weight, but the method doesn't add much over a local `let protocol = match std::env::var(...) { ... }` block. Either way is fine.

**3. The exporter match arms can be deduplicated.**

The two arms are identical except for `.with_tonic()` vs `.with_http()`. You could extract a helper, but the simpler fix is to just build the exporter in a single expression. Unfortunately the opentelemetry-otlp builder types diverge after `.with_tonic()` / `.with_http()`, so the current two-arm match is probably the cleanest option without boxing. No change needed here.

**4. Minor: `.to_lowercase()` allocates unnecessarily.**

`std::env::var` already returns an owned `String`. You could avoid the allocation by using `.eq_ignore_ascii_case()` instead:

```rust
fn from_env() -> Self {
    let val = std::env::var("OTEL_EXPORTER_OTLP_PROTOCOL").unwrap_or_default();
    if val.eq_ignore_ascii_case("http/protobuf")
        || val.eq_ignore_ascii_case("http-protobuf")
        || val.eq_ignore_ascii_case("http_protobuf")
    {
        Self::HttpProtobuf
    } else {
        Self::Grpc
    }
}
```

This avoids the `.to_lowercase()` heap allocation. It's a small thing (this runs once at startup), but it's also simpler to read.

**5. The module doc comment update is good.** The added lines explaining `OTEL_EXPORTER_OTLP_PROTOCOL` are clear and match the existing doc style.

Overall this is a clean, focused change. The two actionable items are adding `Display` on `OtlpProtocol` to remove the inline match, and replacing `.to_lowercase()` with `.eq_ignore_ascii_case()`.

brandonrc (commented):
## DevOps/Infrastructure Review

### 1. Configuration: OTel SDK spec compliance

`OTEL_EXPORTER_OTLP_PROTOCOL` is the correct standard env var name per the [OTel SDK Configuration spec](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol). The accepted values (`grpc` and `http/protobuf`) match the spec. Defaulting to `grpc` when unset is also spec-compliant.

Accepting `http-protobuf` and `http_protobuf` as aliases beyond the spec value is a reasonable operator convenience.

One gap: the OTel spec also defines `http/json` as a valid protocol value. This PR silently falls back to gRPC for that input, which is correct for now but could confuse operators. Consider logging a warning when the value is set but unrecognized, something like:

```rust
value if !value.is_empty() => {
    tracing::warn!(protocol = value, "Unrecognized OTEL_EXPORTER_OTLP_PROTOCOL value, falling back to gRPC");
    Self::Grpc
}
```

### 2. Docker/image impact: reqwest duplication

The backend already depends on `reqwest 0.13`. The `reqwest-client` feature in `opentelemetry-otlp 0.28` pulls in `reqwest 0.12` (confirmed in Cargo.lock). This means the final binary ships **two copies of reqwest** (0.12 and 0.13), including two copies of hyper, rustls/native-tls, and the associated TLS stacks.

Estimated binary size increase: 1-3 MB compressed in the Docker image layer. Not a blocker, but worth noting. Once opentelemetry-otlp catches up to reqwest 0.13, this duplication goes away. If you want to avoid it now, you could use the `hyper-client` feature instead of `reqwest-client`, since hyper is already in the dependency tree, though that requires more manual setup.

### 3. Helm/K8s: chart values

The Helm chart uses a generic `backend.env` map that passes through arbitrary env vars to the backend deployment (`values.yaml` line 93). Operators can already set `OTEL_EXPORTER_OTLP_PROTOCOL` without any chart changes:

```yaml
backend:
  env:
    OTEL_EXPORTER_OTLP_ENDPOINT: "http://otel-collector:4318"
    OTEL_EXPORTER_OTLP_PROTOCOL: "http/protobuf"
```

No Helm template changes are needed. However, the `values.yaml` comments should document this option alongside `OTEL_EXPORTER_OTLP_ENDPOINT` as an example. Operators configuring OTLP for the first time (especially behind Envoy Gateway, which is the whole reason for this PR) need to know this exists.

### 4. Monitoring: protocol visibility

The PR logs the selected protocol in the startup `tracing::info!` line with `protocol = "grpc"` or `protocol = "http/protobuf"`. This is good. Operators can verify which transport is active by checking the startup log for the `OpenTelemetry tracing enabled` message.

### 5. Network: HTTP/protobuf and the FRAME_SIZE_ERROR fix

This directly addresses #729. The root cause is a Rust `h2` crate bug where it sends data frames before the SETTINGS ACK, which Envoy rejects with `FRAME_SIZE_ERROR`. HTTP/protobuf uses HTTP/1.1, completely bypassing HTTP/2 frame negotiation, so it sidesteps the issue entirely.

HTTP/protobuf also works better with:
- HTTP proxies that don't support HTTP/2 (corporate proxies, older ALBs)
- Load balancers that terminate HTTP/2 and re-originate as HTTP/1.1
- Envoy sidecars in strict HTTP/2 mode

**One concern on the endpoint URL**: with gRPC, the endpoint is typically `http://collector:4317`. With HTTP/protobuf, the standard OTLP port is `4318` and the path is `/v1/traces`. The `opentelemetry-otlp` crate's `with_http()` builder appends `/v1/traces` automatically, but the **port** is the operator's responsibility. The PR should document that switching protocol also means switching the endpoint port (4317 for gRPC, 4318 for HTTP). Without this, operators will set `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` and leave the endpoint pointing at `:4317`, which will fail silently.

Consider adding a log warning if the endpoint contains `:4317` and the protocol is `http/protobuf`, or at minimum call this out in the docs PR.

### 6. Rollback: switching back to gRPC

Yes, clean rollback. Removing `OTEL_EXPORTER_OTLP_PROTOCOL` (or setting it to `grpc`) reverts to the previous behavior. The `from_env()` method defaults to `Grpc` when the var is unset. No config file changes, no migration, just an env var toggle. This is exactly the right design for a transport option.

### Summary

The change is clean, small, and well-scoped. Two actionable items:

1. **Warn on unrecognized protocol values** instead of silently falling back to gRPC. Operators who typo the value will otherwise get no feedback.
2. **Document the port change** (4317 for gRPC, 4318 for HTTP/protobuf) either in the code comments, the linked docs PR (#53 on artifact-keeper-site), or both. This is the most likely operational mistake when switching protocols.

The reqwest 0.12/0.13 duplication is a known temporary cost and not worth blocking on.

brandonrc (commented):
## PR Review: feat/otel-protobuf-support

Overall this is a clean, well-scoped change. The structure is solid and the code reads well. A few issues worth addressing before merge:

---

### MUST FIX

**1. The OTel spec default for `OTEL_EXPORTER_OTLP_PROTOCOL` is `http/protobuf`, not `grpc`.**

The [OTel specification](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_protocol) states the default value is `http/protobuf`. The PR's `from_env()` falls back to `Grpc` on the wildcard arm, and the doc comment says "Defaults to gRPC when unset or unrecognised, matching the OTel spec default." This is technically incorrect per the spec.

That said, since the existing behavior before this PR was gRPC-only, keeping gRPC as the default preserves backward compatibility (see SHOULD FIX #1 below for more on this). The fix here is just the doc comment: change "matching the OTel spec default" to something like "preserving backward compatibility with the existing gRPC-only behavior (note: the OTel spec default is `http/protobuf`)." Callers who explicitly set the env var will get the right behavior either way.

**2. HTTP/protobuf endpoint path convention differs from gRPC.**

gRPC exporters expect a base endpoint like `http://localhost:4317` and handle the service path internally via gRPC service routing. HTTP/protobuf exporters expect either:
- A base endpoint like `http://localhost:4318` (the SDK appends `/v1/traces` automatically), or
- A full signal-specific URL like `http://localhost:4318/v1/traces`

The default gRPC port is 4317 and the default HTTP port is 4318. If someone sets `OTEL_EXPORTER_OTLP_ENDPOINT=http://collector:4317` (the current documented default in `.env.example`) and then switches to `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf`, they will silently send HTTP requests to the gRPC port. This will fail at runtime with no clear error.

At minimum, log a warning when `protocol=http/protobuf` and the endpoint contains `:4317`. Better yet, document the port difference in the module doc comment and in `.env.example`.

---

### SHOULD FIX

**1. Silent fallback to gRPC on typos or invalid values.**

The wildcard `_ => Self::Grpc` arm means a typo like `OTEL_EXPORTER_OTLP_PROTOCOL=htpp/protobuf` silently falls back to gRPC with no indication. Since this is an explicit user override, a warning log would help debugging:

```rust
other => {
    // Can't use tracing here (subscriber not initialized yet),
    // so eprintln is appropriate
    if !other.is_empty() {
        eprintln!(
            "WARNING: unrecognized OTEL_EXPORTER_OTLP_PROTOCOL value '{other}', \
             falling back to gRPC"
        );
    }
    Self::Grpc
}
```

The empty-string case (env var unset) should stay silent since that is the normal "not configured" path.

**2. `.env.example` should document the new env var.**

The existing `.env.example` documents `OTEL_EXPORTER_OTLP_ENDPOINT` but has no mention of `OTEL_EXPORTER_OTLP_PROTOCOL`. Add a commented-out entry:

```
# OTEL_EXPORTER_OTLP_PROTOCOL=grpc  # or http/protobuf (default: grpc for backward compat; OTel spec default is http/protobuf)
```

---

### CONSIDER

**1. The `.expect()` calls on exporter builder.**

Both arms use `.expect()` which panics during startup if the exporter cannot be created. This is actually reasonable here: if the operator explicitly configured OTel export and it cannot start, panicking early with a clear message is better than silently running without telemetry. The existing code already used `.expect()` for the gRPC path, so this is consistent. No change needed, just calling it out as reviewed.

**2. Duplicate reqwest versions in the dependency tree.**

The lock file shows both `reqwest 0.12.28` (used by opentelemetry-http and the existing project dependency) and `reqwest 0.13.2` (the project's direct dependency in Cargo.toml). The new `http-proto` + `reqwest-client` features pull in `opentelemetry-http` which depends on `reqwest 0.12.x`, so you now have two reqwest versions compiled. This is not a correctness issue, but it does add to binary size and compile time. Worth noting in case the project plans to upgrade the direct `reqwest` dependency to align versions, or if opentelemetry-otlp releases a version using reqwest 0.13.

**3. The `http/json` protocol value.**

The OTel spec also defines `http/json` as a valid protocol value. This PR does not support it, which is fine since the opentelemetry-otlp crate does not support it either. But the wildcard arm will silently treat `http/json` as gRPC. If someone explicitly sets it, the warning from SHOULD FIX #1 would catch this.

**4. Testing.**

There are no tests for `OtlpProtocol::from_env()`. The function is small, but since it parses user input with multiple accepted aliases (`http/protobuf`, `http-protobuf`, `http_protobuf`), a few unit tests would be cheap insurance:

```rust
#[cfg(test)]
mod tests {
    use super::*;

    #[test]
    fn test_otlp_protocol_from_env_defaults_to_grpc() {
        std::env::remove_var("OTEL_EXPORTER_OTLP_PROTOCOL");
        assert_eq!(OtlpProtocol::from_env(), OtlpProtocol::Grpc);
    }

    #[test]
    fn test_otlp_protocol_http_protobuf_variants() {
        for val in ["http/protobuf", "http-protobuf", "http_protobuf", "HTTP/PROTOBUF"] {
            std::env::set_var("OTEL_EXPORTER_OTLP_PROTOCOL", val);
            assert_eq!(OtlpProtocol::from_env(), OtlpProtocol::HttpProtobuf, "failed for {val}");
        }
    }

    #[test]
    fn test_otlp_protocol_grpc_explicit() {
        std::env::set_var("OTEL_EXPORTER_OTLP_PROTOCOL", "grpc");
        assert_eq!(OtlpProtocol::from_env(), OtlpProtocol::Grpc);
    }
}
```

Note: these tests mutate env vars so they would need `#[serial]` or similar if run in parallel. Alternatively, refactor `from_env()` to take a `&str` input for testability.

---

### LOOKS GOOD

- **Drop impl on `OtelGuard`**: Correct for both transport types. The `SdkTracerProvider::shutdown()` method is transport-agnostic. It flushes the `BatchSpanProcessor`, which in turn flushes whichever exporter (tonic or reqwest) was configured. No changes needed.
- **Backward compatibility**: Existing users who do not set `OTEL_EXPORTER_OTLP_PROTOCOL` get the same gRPC behavior as before. The function signature change to `init_with_otel` is internal.
- **The three alias variants** (`http/protobuf`, `http-protobuf`, `http_protobuf`) are a nice touch for usability, especially since `http/protobuf` contains a slash that some config systems escape.
- **Structured log field** for the protocol in the info message is helpful for debugging.
- **Module-level doc comment** update is clear and accurate.

brandonrc (commented):
Nice work on this! Clean implementation. A few small suggestions to polish before merge.

brandonrc (commented):
Great contribution, thanks for this! A few small suggestions below.

brandonrc (commented):
One more thing - could you add some unit tests for `OtlpProtocol::from_env()`? The function is small but it parses user input with multiple aliases, so a few tests would be cheap insurance. Here's an example of what that could look like:

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use std::sync::Mutex;

    // Env var tests must be serialized
    static ENV_MUTEX: Mutex<()> = Mutex::new(());

    #[test]
    fn test_protocol_defaults_to_grpc_when_unset() {
        let _lock = ENV_MUTEX.lock().unwrap();
        std::env::remove_var("OTEL_EXPORTER_OTLP_PROTOCOL");
        assert_eq!(OtlpProtocol::from_env(), OtlpProtocol::Grpc);
    }

    #[test]
    fn test_protocol_accepts_http_protobuf_variants() {
        let _lock = ENV_MUTEX.lock().unwrap();
        for val in ["http/protobuf", "http-protobuf", "http_protobuf", "HTTP/PROTOBUF"] {
            std::env::set_var("OTEL_EXPORTER_OTLP_PROTOCOL", val);
            assert_eq!(
                OtlpProtocol::from_env(),
                OtlpProtocol::HttpProtobuf,
                "failed for {val}"
            );
        }
        std::env::remove_var("OTEL_EXPORTER_OTLP_PROTOCOL");
    }

    #[test]
    fn test_protocol_grpc_explicit() {
        let _lock = ENV_MUTEX.lock().unwrap();
        std::env::set_var("OTEL_EXPORTER_OTLP_PROTOCOL", "grpc");
        assert_eq!(OtlpProtocol::from_env(), OtlpProtocol::Grpc);
        std::env::remove_var("OTEL_EXPORTER_OTLP_PROTOCOL");
    }

    #[test]
    fn test_protocol_unrecognized_falls_back_to_grpc() {
        let _lock = ENV_MUTEX.lock().unwrap();
        std::env::set_var("OTEL_EXPORTER_OTLP_PROTOCOL", "http/json");
        assert_eq!(OtlpProtocol::from_env(), OtlpProtocol::Grpc);
        std::env::remove_var("OTEL_EXPORTER_OTLP_PROTOCOL");
    }
}
```

Alternatively, you could refactor `from_env()` to take a `&str` parameter for easier testing without env var mutation:

```rust
fn from_value(val: &str) -> Self {
    match val.to_lowercase().as_str() {
        "http/protobuf" | "http-protobuf" | "http_protobuf" => Self::HttpProtobuf,
        _ => Self::Grpc,
    }
}

fn from_env() -> Self {
    let val = std::env::var("OTEL_EXPORTER_OTLP_PROTOCOL").unwrap_or_default();
    Self::from_value(&val)
}
```

Then tests just call `from_value()` directly without touching env vars.

brandonrc:
@Firjens  I had a couple nits here and there but this is pretty solid and appreciate the help. 

I attached some examples of code coverage that can help get this past the gates I have,.

Firjens:
Appreciate it @brandonrc, it would be nice if these checks run without maintainer approval so we can ensure that the quality checks pass before marking the merge request as ready.

I've added the tests. :)

brandonrc:
> Appreciate it @brandonrc, it would be nice if these checks run without maintainer approval so we can ensure that the quality checks pass before marking the merge request as ready.
> 
> I've added the tests. :)

I am working that right now.. only issue is the key for the sonarqube thing.. cannot figure anything around that.. might try and find another mechanism maybe just something simple.

brandonrc:
Hey! Can you rebase this on main when you get a chance? We've been cleaning up the CI/CD pipeline (removed SonarCloud, switched to native coverage gates, optimized self-hosted runners) and there are a few workflow changes that may cause merge conflicts if left too long. Thanks!

brandonrc:
Thanks for the contribution, @Firjens. The implementation is clean and solves the problem well.

We recreated this on an internal branch at #812 so the CI coverage gate can pass (fork PRs can't access the secrets needed for the coverage upload). The code is functionally identical to what you wrote here.

Closing this in favor of #812, which credits you in the PR description.

Issue #737, opened 2026-04-15
OOM when making lots of parallel requests to proxy resource


roblabla:
### Description

When making lots of parallel requests to a remote/proxied repository resource, artifact-keeper ends up running out of memory, getting OOM-killed. In our deployment, we limit Artifact Keeper to 1GiB of RAM to avoid it impacting other workload on the machine. We can raise it, but it seemed curious that it would take that much RAM when making parallel requests, so I decided to dig around.

Looking into it, it appears to be caused by Artifact Keeper keeping the entire file in memory, instead of doing a passthrough stream. Digging around, it looks like the problem comes from `get_cached_artifact`, which downloads the entire file in memory. https://github.com/artifact-keeper/artifact-keeper/blob/main/backend/src/services/proxy_service.rs#L341C28-L360

### Environment Information

- Deployment type: Kubernetes
- Operating System: Ubuntu
- Architecture: X64
- Database: PostgreSQL
- Reverse Proxy: istio
- Network setup: 

### Backend Version

1.1.2

### Package Format

N/A / General

### Expected Behavior

We should be able to make many parallel requests to APK without memory ballooning.

### Observed Behavior

Memory goes very high when making parallel requests.

### Steps to Reproduce

- configure a remote apk repository
- start multiple 3.22 alpine container (around 5, more will lead to more memory usage)
- reconfigure /etc/apk/repositories to point to the AK repository
run `apk add --update qemu qemu-system-x86_64 qemu-img jq zip p7zip xorriso libarchive-tools` in all the repos at the same time (ideally using a script).

### Logs

```shell

```

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have redacted any sensitive information (passwords, tokens, private URLs)
- [x] I am using a supported version (not end-of-life)

lppedd:
@brandonrc will know better, but do you think #604 might be related?

roblabla:
Hmm, that might help. I'll try to update to latest master and enable the env var.

roblabla:
That did help quite a bit - my main use case (many CIs operating in parallel downloading the same package over and over again) doesn't trigger the OOM anymore.

[...]

PR #739, opened 2026-04-15
feat: add get_stream/put_stream to StorageBackend trait

brandonrc:
## Summary

Add streaming methods (`get_stream`, `put_stream`) to the `StorageBackend` trait as Phase 1 of the streaming proxy architecture (#737). This lays the foundation for eliminating full-artifact memory buffering in the proxy service.

Both methods have default implementations that wrap the existing `get()`/`put()` methods, so all existing backends and consumers continue to work without changes. The `FilesystemStorage` and `S3Backend` override with proper streaming implementations:

- **FilesystemStorage** `get_stream`: opens the file and returns a `ReaderStream` with 256 KB chunks, avoiding loading the entire file into memory.
- **FilesystemStorage** `put_stream`: writes to a temp file in the same directory (guaranteeing same-filesystem atomic rename), computes SHA-256 incrementally, fsyncs, then renames. Temp files are cleaned up on any error.
- **S3Backend** `get_stream`: uses `GetResult::into_stream()` from `object_store` instead of `.bytes().await`.
- **S3Backend** `put_stream`: uses `WriteMultipart` from `object_store` (5 MB part buffering handled internally), with `abort()` on stream error to prevent partial S3 objects.

GCS and Azure backends use the default (collect-then-delegate) implementations for now, preserving their current memory profile until native streaming is added in a follow-up.

Introduces `PutStreamResult` with `checksum_sha256` and `bytes_written` fields, returned by `put_stream`.

## Test Checklist
- [x] Manually tested locally

16 new tests:
- 3 for `PutStreamResult` (construction, clone, debug)
- 4 for default trait implementations (get_stream, put_stream single/multi/empty)
- 9 for `FilesystemStorage` overrides (correct content, chunking, nonexistent key error, multi-chunk write, SHA-256 verification, atomic rename with no temp files left, temp cleanup on stream error, empty stream, roundtrip with get_stream)

All 7642 tests pass. `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --lib` all clean.

## API Changes
- [x] N/A - no API changes

This is a storage-layer-only change. No handler, proxy service, or HTTP API modifications.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4071,7 +4071,7 @@ version = "0.13.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "be769465445e8c1474e9c5dac2018218498557af32d9ed057325ec9a41ae81bf"
 dependencies = [
- "heck 0.5.0",
+ "heck 0.4.1",
  "itertools 0.14.0",
  "log",
  "multimap",
@@ -4785,9 +4785,9 @@ checksum = "f87165f0995f63a9fbeea62b64d10b4d9d8e78ec6d7d51fb2125fda7bb36788f"
 
 [[package]]
 name = "rustls-webpki"
-version = "0.103.10"
+version = "0.103.12"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "df33b2b81ac578cabaf06b89b0631153a3f416b0a886e8a7a1707fb51abbd1ef"
+checksum = "8279bb85272c9f10811ae6a6c547ff594d6a7f3c6c6b02ee9726d1d0dcfcdd06"
 dependencies = [
  "aws-lc-rs",
  "ring",
--- a/backend/src/storage/filesystem.rs
+++ b/backend/src/storage/filesystem.rs
@@ -2,12 +2,20 @@
 
 use async_trait::async_trait;
 use bytes::Bytes;
+use futures::stream::BoxStream;
+use futures::StreamExt;
+use sha2::{Digest, Sha256};
 use std::path::PathBuf;
 use tokio::fs;
-use tokio::io::AsyncWriteExt;
+use tokio::io::{AsyncWriteExt, BufReader};
+use tokio_util::io::ReaderStream;
+use uuid::Uuid;
 
-use super::StorageBackend;
+use super::{PutStreamResult, StorageBackend};
 use crate::error::{AppError, Result};
+
+/// Chunk size for streaming reads (256 KB).
+const STREAM_CHUNK_SIZE: usize = 256 * 1024;
 
 /// Filesystem-based storage backend
 pub struct FilesystemStorage {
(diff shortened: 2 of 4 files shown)


brandonrc (commented):
## Review: StorageBackend streaming trait (Phase 1)

### MUST FIX

**1. Filesystem `put_stream`: `with_extension` will mangle keys that contain dots**

```rust
let temp_path = dest.with_extension(format!("tmp.{}", Uuid::new_v4()));
```

`PathBuf::with_extension` *replaces* the existing extension. For a CAS key like `abcdef1234567890` (no dots) this works fine, but the `StorageBackend` trait accepts arbitrary string keys. In `storage_service.rs`, keys like `repos/maven/artifact.jar` and `test/file.txt` are already in use. For those:

- `artifact.jar` -> temp file `artifact.tmp.<uuid>` (`.jar` replaced, not appended)
- `artifact.txt` -> temp file `artifact.tmp.<uuid>` (`.txt` replaced)

Two different keys sharing the same stem get the same temp file prefix. The UUID suffix prevents actual collisions, but the final file would still rename correctly, so this is not a correctness bug today. The real risk: if `put_stream` is called with a dotted key, the temp path occupies a different "slot" than intended, which could confuse cleanup logic or monitoring that looks for `.tmp.*` patterns.

Fix: build the temp path by appending rather than replacing:

```rust
let mut temp_name = dest.file_name().unwrap_or_default().to_os_string();
temp_name.push(format!(".tmp.{}", Uuid::new_v4()));
let temp_path = dest.with_file_name(temp_name);
```

**2. Filesystem `put_stream`: the `collect_tmp_files` test helper checks the wrong thing**

```rust
} else if path
    .extension()
    .map(|e| e.to_string_lossy().starts_with("tmp."))
    .unwrap_or(false)
```

`PathBuf::extension()` returns only the text after the *last* dot. For a temp file named `abcdef1234567890.tmp.550e8400-e29b-...`, the extension is `550e8400-e29b-...` (the UUID), not `tmp.550e8400-...`. This helper would never find those temp files, so the test that verifies "no temp files remain on error" is not actually testing what it claims. It passes vacuously.

Fix: check the full filename string with `.contains(".tmp.")` instead of inspecting `.extension()`.

### SHOULD FIX

**3. S3 `put_stream`: no explicit abort when `finish()` fails**

Looking at the code:
```rust
write.finish().await.map_err(|e| {
    AppError::Storage(format!("Failed to complete multipart upload for '{}': {}", key, e))
})?;
```

I checked the `object_store` 0.13.2 source and `finish()` *does* internally call `abort()` on failure (line 237 of upload.rs), so this is not a bug. Noting it here for documentation: the caller does not need to handle this.

**4. Default `put_stream` collects everything into memory**

The default implementation accumulates the entire stream into a `Vec<u8>` then calls `put()`. This is documented and intentional (GCS/Azure fall back to this). Worth adding a `tracing::warn!` or at least a code comment that this path should not be used for large artifacts, so that Phase 2 callers know to check `supports_streaming()` or similar before relying on constant-memory behavior.

**5. Filesystem `put_stream` rename error: sync `std::fs::remove_file` in async context**

```rust
fs::rename(&temp_path, &dest).await.map_err(|e| {
    let _ = std::fs::remove_file(&temp_path);  // blocking call on tokio thread
    AppError::Storage(format!("Rename error: {}", e))
})?;
```

This is in the `map_err` closure which is synchronous, so `tokio::fs::remove_file` cannot be used directly. For a best-effort cleanup on an error path this is acceptable. If you want to be strict, you could restructure to use `match` + `.await` on the async variant. Low priority since rename failures are rare.

**6. Consider adding `supports_streaming()` to the trait**

Phase 2 callers (the proxy service) will need to know whether a backend truly streams or just buffers-and-delegates. A `fn supports_streaming(&self) -> bool { false }` default method that filesystem/S3 override to `true` would let the proxy service make informed decisions. Not blocking for Phase 1, but worth thinking about now so the trait doesn't need another breaking change.

### LOOKS GOOD

**Trait design**: The signatures are clean and ergonomic. `BoxStream<'static, Result<Bytes>>` is the standard pattern for async byte streams in the Rust ecosystem. `PutStreamResult` returning both checksum and byte count is practical for Phase 2 (the proxy can forward both to the client). Default implementations mean existing backends and tests don't break. This will integrate well with the proxy service.

**Filesystem `get_stream`**: `BufReader` + `ReaderStream::with_capacity` is correct. 256KB chunk size is reasonable (matches typical NVMe page sizes and avoids excessive syscall overhead while keeping memory usage low for concurrent streams).

**Filesystem `put_stream` atomicity**: temp-file-then-rename in the same directory guarantees same-filesystem atomicity on POSIX. `sync_all()` before rename ensures durability. The `drop(file)` before rename is correct (closes the fd). Cleanup on every error path (write error, stream error, sync error) is thorough.

**S3 `get_stream`**: `GetResult::into_stream()` is the correct API in object_store 0.13 for streaming reads without buffering the full object.

**S3 `put_stream`**: `WriteMultipart::new(upload)` uses the default 5MB chunk size, which satisfies S3's minimum part size requirement. The `abort()` call on stream error prevents orphaned multipart uploads that would accrue storage costs. `finish()` handles its own abort internally on failure.

**SHA-256 incremental hashing**: `sha2::Sha256` with `hasher.update(&data)` per chunk followed by `hasher.finalize()` is the standard incremental pattern. The test verifies known vectors ("hello world", empty input). Correct.

**Test coverage**: 16 tests covering: default trait fallbacks (4), PutStreamResult type (3), filesystem get_stream (3), filesystem put_stream (6: content, sha256, atomic rename, error cleanup, empty stream, roundtrip). Good coverage for a Phase 1 trait. No S3 integration tests, but that's expected since S3 tests require infrastructure.

PR #740, opened 2026-04-15
fix: add proxy OOM safeguards (semaphore + size limit)

### LINKED ISSUE #737: OOM when making lots of parallel requests to proxy resource

brandonrc:
## Summary

Adds three configurable safeguards to `ProxyService` to prevent out-of-memory kills caused by unbounded concurrent upstream artifact fetches (#737):

- **Concurrent fetch semaphore** (`PROXY_MAX_CONCURRENT_FETCHES`, default 20): limits parallel upstream downloads. Requests that cannot acquire a permit within `PROXY_QUEUE_TIMEOUT_SECS` (default 30) receive HTTP 503.
- **Per-artifact size limit** (`PROXY_MAX_ARTIFACT_SIZE_BYTES`, default 2 GB): checks `Content-Length` before reading the body and actual body size after reading, returning HTTP 502 for oversized artifacts.
- **Connection pool limits** (`pool_max_idle_per_host=50`, `pool_idle_timeout=90s`) on the proxy HTTP client to bound idle connection fan-out to upstream registries.

Also adds a `ServiceUnavailable` error variant (HTTP 503) and an `ak_proxy_fetches_in_flight` gauge metric for monitoring concurrent proxy fetches.

Peak proxy memory is now bounded to approximately `concurrent_fetches * max_artifact_size` (40 GB worst case at defaults, much lower in practice). Operators on memory-constrained hosts should lower both values.

This is a targeted patch for `release/1.1.x`. The full streaming proxy rewrite (replacing `response.bytes()` with chunked `AsyncRead`) is deferred to v1.2.0.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

New unit tests (9 total):
- `test_proxy_semaphore_limits_concurrent_fetches`
- `test_proxy_semaphore_timeout_returns_503`
- `test_proxy_semaphore_releases_on_drop`
- `test_proxy_rejects_oversized_content_length`
- `test_proxy_allows_content_length_within_limit`
- `test_proxy_rejects_oversized_body_no_content_length`
- `test_proxy_allows_body_within_limit_no_content_length`
- `test_proxy_config_defaults`
- `test_proxy_config_custom_values`

Plus 2 error tests for the new `ServiceUnavailable` variant.

Closes #737

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -262,6 +262,28 @@ JWT_EXPIRATION_SECS=86400
 # MAX_UPLOAD_SIZE=10737418240
 
 # -----------------------------------------------------------------------------
+# Proxy safeguards (backend)
+# -----------------------------------------------------------------------------
+# These settings bound peak memory consumption when the proxy service fetches
+# artifacts from upstream registries on behalf of remote/proxy repositories.
+
+# Maximum number of concurrent upstream fetches. Each fetch loads the full
+# artifact into memory, so this limits how many large downloads can overlap.
+# Default: 20. Lower this on memory-constrained hosts.
+# PROXY_MAX_CONCURRENT_FETCHES=20
+
+# Maximum artifact size (in bytes) the proxy will accept from upstream.
+# Artifacts larger than this are rejected with 502 Bad Gateway.
+# Default: 2147483648 (2 GB).
+# PROXY_MAX_ARTIFACT_SIZE_BYTES=2147483648
+
+# Seconds to wait for a proxy fetch permit before returning 503.
+# When all concurrent fetch slots are in use, new requests queue for up to
+# this duration before being rejected.
+# Default: 30.
+# PROXY_QUEUE_TIMEOUT_SECS=30
+
+# -----------------------------------------------------------------------------
 # Rate limiting (backend)
 # -----------------------------------------------------------------------------
 # The built-in rate limiter is per-instance (in-memory). For multi-instance
--- a/backend/src/api/handlers/events.rs
+++ b/backend/src/api/handlers/events.rs
@@ -110,6 +110,9 @@ mod tests {
             lifecycle_check_interval_secs: 60,
             max_upload_size_bytes: 10_737_418_240,
             allow_local_admin_login: false,
+            proxy_max_concurrent_fetches: 20,
+            proxy_max_artifact_size_bytes: 2_147_483_648,
+            proxy_queue_timeout_secs: 30,
             metrics_port: None,
         }
     }
--- a/backend/src/api/handlers/health.rs
+++ b/backend/src/api/handlers/health.rs
@@ -729,6 +729,9 @@ mod tests {
             lifecycle_check_interval_secs: 60,
             max_upload_size_bytes: 10_737_418_240,
             allow_local_admin_login: false,
+            proxy_max_concurrent_fetches: 20,
+            proxy_max_artifact_size_bytes: 2_147_483_648,
+            proxy_queue_timeout_secs: 30,
             metrics_port: None,
         }
     }
--- a/backend/src/config.rs
+++ b/backend/src/config.rs
@@ -122,6 +122,17 @@ pub struct Config {
     /// recovery mechanism when SSO is misconfigured.
     pub allow_local_admin_login: bool,
 
+    /// Maximum concurrent upstream proxy fetches. Bounds peak memory from
+    /// parallel proxy downloads. Set to 0 to disable (not recommended).
+    pub proxy_max_concurrent_fetches: u32,
+
+    /// Maximum artifact size in bytes that the proxy will fetch from upstream.
+    /// Requests for artifacts larger than this are rejected with 502.
+    pub proxy_max_artifact_size_bytes: u64,
+
+    /// Seconds to wait for a proxy fetch permit before returning 503.
+    pub proxy_queue_timeout_secs: u64,
+
     /// Port for the unauthenticated Prometheus metrics-only listener.
     ///
     /// When set, a second TCP listener is started on this port serving only
@@ -172,6 +183,9 @@ redacted_debug!(Config {
     show lifecycle_check_interval_secs,
     show max_upload_size_bytes,
     show allow_local_admin_login,
+    show proxy_max_concurrent_fetches,
+    show proxy_max_artifact_size_bytes,
+    show proxy_queue_timeout_secs,
     show metrics_port,
 });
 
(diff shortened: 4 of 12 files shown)


brandonrc (commented):
## Security Review: Proxy OOM Safeguards (#740)

Overall this is a solid defensive change that addresses a real memory exhaustion vector. The semaphore pattern is correct, size limits are applied at the right layers, and the test coverage is good. A few findings below.

---

### MUST FIX

**1. Chunked transfer encoding bypasses the pre-download size check**

The `Content-Length` check in `read_upstream_response` correctly runs before `response.bytes()`, but a malicious or misconfigured upstream can use chunked transfer encoding (or HTTP/2, which never sends `Content-Length`). In that case `response.content_length()` returns `None`, the pre-download check is skipped entirely, and the full body is loaded into memory before the post-download check fires.

With 20 concurrent permits and no `Content-Length`, an attacker controlling an upstream could stream 20 x 2 GB = 40 GB into memory before any check triggers. The post-download check catches it after the damage is done.

**Fix:** Use `response.bytes()` replacement with a streaming size-limited reader. Since the PR description says the full streaming rewrite is deferred to v1.2, a pragmatic interim fix is to chunk-read with `response.chunk()` in a loop, accumulating into a `BytesMut` and aborting once `max_size` is exceeded:

```rust
let mut body = bytes::BytesMut::new();
while let Some(chunk) = response.chunk().await.map_err(|e| {
    AppError::Storage(format!("Failed to read upstream response: {}", e))
})? {
    body.extend_from_slice(&chunk);
    if body.len() as u64 > max_size {
        return Err(AppError::BadGateway(format!(
            "Upstream artifact body exceeds size limit ({} bytes) during download",
            max_size
        )));
    }
}
let content = body.freeze();
```

This bounds peak memory per fetch to `max_size` instead of allowing unbounded buffering until the post-download check.

**2. `proxy_max_concurrent_fetches = 0` creates a closed semaphore**

The config comment says "Set to 0 to disable (not recommended)" but `Semaphore::new(0)` creates a semaphore with zero permits. Every fetch will block until the queue timeout, then return 503. This is a silent total outage of the proxy, not "disabled."

**Fix:** Either reject 0 at config validation time with a clear error, or treat 0 as "unlimited" by skipping the semaphore acquire path:

```rust
if config.proxy_max_concurrent_fetches == 0 {
    return Err(AppError::Config(
        "PROXY_MAX_CONCURRENT_FETCHES must be >= 1".into()
    ));
}
```

---

### SHOULD FIX

**3. No `Retry-After` header on 503 responses**

The `ServiceUnavailable` variant maps to HTTP 503 through the generic `IntoResponse` impl, which returns a JSON body but no `Retry-After` header. RFC 7231 Section 6.6.4 recommends including `Retry-After` so clients know when to retry. Without it, aggressive clients will hammer the endpoint immediately, making the overload worse.

This could be addressed at the `AppError` level by special-casing `ServiceUnavailable` in `into_response` to append the header, or by returning the response directly from the handler instead of going through `AppError`. Even a static `Retry-After: 5` would help.

**4. Semaphore starvation is possible but mitigated**

An attacker who can trigger proxy fetches (any authenticated user requesting artifacts from a remote repo) could intentionally hold all 20 permits by requesting large slow-draining upstream artifacts. The 30-second queue timeout bounds the blast radius, but legitimate users will get 503s during the attack. This is inherent to any concurrency limiter, but consider:

- Per-user or per-repository permit fairness (one user can hold at most N/4 permits)
- Shorter HTTP timeout on upstream fetches (currently `HTTP_TIMEOUT_SECS`, check its value) to free permits faster when upstream is slow

Not blocking, but worth a follow-up issue.

**5. Gauge metric not decremented on semaphore acquire failure**

The `ak_proxy_fetches_in_flight` gauge is incremented after the permit is acquired but is only decremented in the success path after `fetch_from_upstream_inner` returns. If `fetch_semaphore.acquire()` itself fails (semaphore closed), the `map_err` on the inner `Result` returns before the gauge is incremented, so this is actually fine. But confirm that every exit from `fetch_from_upstream_inner` (success or error) flows back through the decrement. Since the decrement and `drop(permit)` are unconditional after the `.await`, this is correct. The ordering (decrement then drop) is also fine since the gauge is informational.

---

### LOOKS GOOD

- **Semaphore acquisition point is correct.** It wraps the entire upstream fetch including auth token exchange, so the permit covers the full memory-holding duration.
- **Permit release on all paths.** The `permit` variable is dropped after `fetch_from_upstream_inner` returns regardless of success or error, because `drop(permit)` is outside the `result` match. Panics would also drop the permit since `SemaphorePermit` implements `Drop`.
- **`ServiceUnavailable` maps to 503 correctly.** Verified in `status_and_code()`.
- **Pool limits are reasonable.** `pool_max_idle_per_host(50)` and `pool_idle_timeout(90s)` are sensible defaults for a proxy that fans out to many upstream registries. 50 idle connections per host prevents file descriptor exhaustion while keeping enough warm connections for hot upstreams.
- **Size limit on both the initial response and the retry-after-401 path.** Both calls to `read_upstream_response` pass `self.max_artifact_size`, so the OCI bearer token retry path is also size-limited.
- **Error variant placement.** `ServiceUnavailable` message is passed through `user_message()` directly (client-facing), which is appropriate since it contains no internal details.
- **Config tests cover defaults and custom values.**

PR #743, opened 2026-04-15
fix: resolve download URL from upstream config.json for Cargo pull-through

### LINKED ISSUE #731: The pull-through behavior for Cargo is not as expected.
### Description

When fetching from `https://crates.io/`, the `cargo add` command fails with an error stating the index cannot be found.

However, when fetching from `https://index.crates.io`, `cargo fetch` fails to pull the packages.

I'm not sure if this is the expected behavior or a bug.

One workaround is to create a **virtual** repository, then set up an **index remote** and a **crates.io remote** separately, and add both of them to that same virtual repository.

### Environment Information

- Deployment type: 
- Operating System: 
- Architecture: 
- Database: 
- Reverse Proxy: 
- Network setup: 

### Backend Version

1.1.2

### Web UI Version

_No response_

### Package Format

Cargo (Rust)

### Expected Behavior

A single Cargo remote cannot simultaneously support pull-through requirements.

### Observed Behavior

A single Cargo remote can satisfy the requirements for both `cargo add` and `cargo fetch` simultaneously.

### Steps to Reproduce

```
[source.crates-io]
replace-with = 'mirror'

[source.mirror]
registry = "sparse+http://ip/cargo/crates"

[registries.mirror]
index = "sparse+http://ip/cargo/crates"
```

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

Fixes the Cargo pull-through proxy so a single remote repository can serve both index lookups and crate downloads for split-host registries like crates.io.

Previously, when `upstream_url` was set to the index host (e.g. `https://index.crates.io`), the download handler would try to fetch `.crate` files from that same host, which does not serve downloads. Users had to create a virtual repository with two separate remotes as a workaround.

The download handler now fetches the upstream's `config.json` to discover the `dl` (download) URL, which is how Cargo itself resolves download locations. The resolved URL is cached for 1 hour to avoid repeated `config.json` requests.

Changes:
- Added `resolve_upstream_dl_url()` to fetch and parse the upstream `config.json`
- Added `build_download_url()` to construct download URLs from the `dl` field, supporting both plain URLs and `{crate}/{version}` template markers
- Added `split_url()` helper to decompose a full URL into origin + path for the proxy layer
- Added module-level `UPSTREAM_CONFIG_CACHE` with 1-hour TTL for resolved download URLs
- Download handler now uses `proxy_fetch_with_cache_key` so the local cache path stays stable regardless of which upstream host was resolved
- 18 new unit tests covering all new functions and the crates.io scenario

Closes #731

## Test Checklist
- [ ] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -29,6 +29,7 @@ use tracing::info;
 use crate::api::handlers::error_helpers::{map_db_err, map_storage_err};
 use crate::api::handlers::proxy_helpers;
 use crate::api::middleware::auth::{require_auth_with_bearer_fallback, AuthExtension};
+use crate::api::validation::validate_outbound_url;
 use crate::api::SharedState;
 use crate::api::{CachedRepo, IndexCache, RepoCache, REPO_CACHE_TTL_SECS};
 use crate::error::AppError;
@@ -794,6 +795,11 @@ async fn download(
                     {
                         Some(dl_url) => {
                             let full = build_download_url(&dl_url, &name_lower, &version);
+                            // Validate the resolved download URL against SSRF.
+                            // A malicious upstream config.json could set `dl` to
+                            // a cloud metadata endpoint or internal service URL.
+                            validate_outbound_url(&full, "Cargo upstream download URL")
+                                .map_err(|e| e.into_response())?;
                             split_url(&full)
                                 .unwrap_or_else(|| (upstream_url.clone(), fallback_path.clone()))
                         }
@@ -2741,6 +2747,51 @@ mod tests {
         assert_eq!(
             full_fallback,
             "https://index.crates.io/api/v1/crates/serde/1.0.0/download"
+        );
+    }
+
+    #[test]
+    fn test_build_download_url_rejects_internal_addresses() {
+        use crate::api::validation::validate_outbound_url;
+
+        // Cloud metadata endpoint (AWS IMDSv1)
+        let dl = "http://169.254.169.254/latest/meta-data/";
+        let url = build_download_url(dl, "evil", "1.0.0");
+        assert!(
+            validate_outbound_url(&url, "Cargo upstream download URL").is_err(),
+            "cloud metadata URL should be rejected"
+        );
+
+        // Localhost
+        let dl = "http://localhost:8080/evil";
+        let url = build_download_url(dl, "crate", "0.1.0");
+        assert!(
+            validate_outbound_url(&url, "Cargo upstream download URL").is_err(),
+            "localhost URL should be rejected"
+        );
+
+        // Private network (10.x)
+        let dl = "http://10.0.0.1/packages";
+        let url = build_download_url(dl, "crate", "0.1.0");
+        assert!(
+            validate_outbound_url(&url, "Cargo upstream download URL").is_err(),
+            "private network URL should be rejected"
+        );
+
+        // Docker-internal service name
+        let dl = "http://backend:8080/internal";
+        let url = build_download_url(dl, "crate", "0.1.0");
+        assert!(
+            validate_outbound_url(&url, "Cargo upstream download URL").is_err(),
+            "Docker-internal service URL should be rejected"
+        );
+
+        // Legitimate external URL should pass
+        let dl = "https://crates.io/api/v1/crates";
+        let url = build_download_url(dl, "serde", "1.0.0");
+        assert!(
+            validate_outbound_url(&url, "Cargo upstream download URL").is_ok(),
+            "legitimate external URL should be accepted"
         );
     }
 }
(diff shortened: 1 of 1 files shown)


brandonrc (commented):
## Code Review: Cargo pull-through split-host fix

### 1. config.json fetch and dl URL resolution

**LOOKS GOOD.** The resolution logic correctly follows the Cargo registry spec: fetch `config.json` from the index URL, extract the `dl` field, handle both plain base URLs and `{crate}`/`{version}` template markers. Preferring `index_upstream_url` over `upstream_url` for the config fetch is the right call since that is where registries serve `config.json`.

### 2. Security: SSRF via malicious upstream config.json

**SHOULD FIX.** The `dl` URL from `config.json` is attacker-controlled content (fetched from the upstream registry). A compromised or malicious upstream could set `dl` to `http://169.254.169.254/latest/meta-data/` or `http://postgres:5432/` and the proxy would happily fetch from that address.

The resolved `dl_base` origin is passed directly to `proxy_fetch_with_cache_key`, which calls `build_remote_repo` and then `fetch_from_upstream` with no SSRF validation on the URL. The existing `validate_outbound_url` function in `api/validation.rs` blocks private IPs, cloud metadata endpoints, and internal service hostnames. It should be applied to the resolved download URL before making the request:

```rust
// In resolve_upstream_dl_url, after extracting dl_url:
crate::api::validation::validate_outbound_url(&dl_url, "upstream Cargo dl URL")
    .map_err(|e| {
        tracing::warn!("Upstream config.json dl URL failed SSRF check: {}", e);
    })
    .ok()?;
```

Alternatively, validate in `build_download_url` or right before the `proxy_fetch_with_cache_key` call. The risk is real: if an admin points `upstream_url` at a third-party registry they don't fully control, that registry's `config.json` dictates where downloads come from.

### 3. UPSTREAM_CONFIG_CACHE thread safety and expiry under concurrent load

**LOOKS GOOD.** `std::sync::RwLock` is the right choice here since the cache is accessed from async code but the lock is never held across an await point. The lock is only held for the duration of a HashMap lookup/insert, so contention will be minimal.

On cache expiry under concurrent load: multiple threads may see a stale entry simultaneously, all fetch `config.json`, and all write back. This is a harmless thundering-herd: `config.json` is tiny, the upstream can handle a few extra requests once per hour, and the last writer wins with identical data. No correctness issue.

The `retain` call inside `config_cache_set` is a nice touch for preventing unbounded growth, though it does scan the entire map on every insert. With typical deployments having a handful of upstream registries this is fine. If the number of upstream URLs ever grew large, this could be moved to a periodic cleanup, but that is not a realistic concern today.

### 4. Error handling for missing/malformed config.json

**LOOKS GOOD.** The `resolve_upstream_dl_url` function uses `Option` chaining (`?` on `.ok()`) throughout: if `config.json` is missing (404 from upstream), unparseable as JSON, or missing the `dl` field, it returns `None`. The download handler falls back to `(upstream_url, fallback_path)` when `None` is returned. This is the correct behavior: the feature gracefully degrades to the old code path.

One minor observation: errors during the config fetch are silently swallowed. A `tracing::debug!` or `tracing::warn!` when the fetch or parse fails would help operators debug configuration issues. Not a blocker.

### 5. split_url edge cases

**LOOKS GOOD** with one minor gap. The function correctly handles ports (`http://localhost:8080/path`), deep paths, and rejects URLs without schemes or paths. The tests cover these cases well.

Edge case to consider: a `dl` URL with query parameters (e.g., `https://cdn.example.com/crates?token=abc`). Currently `split_url` would include the query string in the path component, which would work correctly with `proxy_fetch_with_cache_key` since the path is only used for the HTTP request. The `cache_path` is computed independently, so no issue there.

Another edge case: `split_url("https://crates.io/")` returns `("https://crates.io", "")` (empty path). This cannot happen in practice because `build_download_url` always appends at least `/{name}/{version}/download`, but a defensive check for empty path in the caller would not hurt.

### 6. Test coverage

**LOOKS GOOD.** The 18 tests cover: `build_download_url` (plain base, trailing slash, template markers, partial markers, prerelease versions, short names), `split_url` (standard, ports, no-path, no-scheme, root path, deep path), config cache (miss, roundtrip, overwrite), and end-to-end scenarios (crates.io, self-hosted, fallback). Good use of UUIDs to avoid test interference on the static cache.

**Missing test coverage:**
- Virtual repo download path. The PR only fixes the direct remote repo path. Virtual repositories with Cargo remote members pointed at split-host registries (e.g., crates.io) will still fail because `resolve_virtual_download` in `proxy_helpers.rs` uses plain `proxy_fetch` with `upstream_url` as the base. This is a separate issue but worth tracking.

### Summary

| Finding | Severity |
|---------|----------|
| No SSRF validation on `dl` URL from upstream config.json | SHOULD FIX |
| Silent error swallowing in `resolve_upstream_dl_url` | Nice to have (add tracing) |
| Virtual repo path not updated for split-host | Separate issue, worth tracking |
| Everything else | LOOKS GOOD |

The core approach is sound and follows the Cargo registry spec correctly. The SSRF gap should be closed before merge since the existing `validate_outbound_url` utility is already available and a one-line addition.

PR #757, opened 2026-04-16
fix: return members in group detail endpoint

nikitatsym:
## Summary

`GET /api/v1/groups/{id}` returns `GroupResponse` with `member_count` but no actual member list. The frontend (`groups/page.tsx`) expects a `members` array in the response and silently falls back to an empty list, making it look like every group has no members.

This adds a `GroupDetailResponse` that includes `members: Vec<GroupMemberResponse>` with `user_id`, `username`, `display_name`, and `joined_at` fields, populated by joining `user_group_members` with `users`.

## Test Checklist
- [x] Manually tested locally

## API Changes


brandonrc:
Thank you @nikitatsym for raising this and for the implementation approach. We've incorporated your changes into #813 with the additional pagination support and OpenAPI schema updates that the review requested. Closing this in favor of #813, which credits your contribution in the commit history.

PR #866, opened 2026-04-25
feat: add AK_GUEST_ACCESS_ENABLED to disable guest access (fixes #850)

### LINKED ISSUE #850: Add the ability to disable guest access and force login
### Problem Statement

Our use case for Artifact Keeper is to serve as our private package manager. This means that we always want to keep everything behind a login screen. It does not make sense for Artifact Keeper to show an empty repository all the time.

Additionally, the "Public Repository" button seems like a potential opportunity to mess up for us (as everything we would want to put into our Artifact Keeper should be private)

### Proposed Solution

Essentially, we'd like to have a feature where can disable guest (unauthanticated) access, and have it automatically redirect to the login page if you are not authenticated.

If guest access is disabled, the "Public Repository" toggle should also be removed when creating a repository, and just mark it implicitly private.

Image

### Alternatives Considered

Currently our workaround is to make all repositories private, but this feels like it could be a better process.

### Use Case

As a DevOps user, I want all of our packages to be private so that we don't accidentally expose some resources (by mistakenly labeling some registries as public).

### Component

Web UI

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

Implements issue #850: a server-wide policy flag that disables anonymous (unauthenticated) access. When operators set `AK_GUEST_ACCESS_ENABLED=false` (or `0`), every request must present valid credentials and public repositories effectively become private. Defaults to `true` so existing deployments are unaffected.

The new `guest_access_guard` middleware runs as a global outer layer in `routes::create_router` and short-circuits unauthenticated requests with `401`. A small allowlist keeps the login flow, initial-setup wizard, `/api/v1/system/config` (so the web UI can fetch config before the user logs in), Kubernetes health probes, and the OCI `/v2/` challenge endpoint reachable so package clients can negotiate auth normally.

The guard performs its own token resolution via the existing `try_resolve_auth` helper rather than relying on inner-middleware extensions, so it can sit at the outermost layer and gate every route uniformly regardless of which inner auth middleware the matched route uses (`auth_middleware`, `optional_auth_middleware`, `repo_visibility_middleware`, or none).

The repository create/update handlers now silently coerce `is_public=true` to `false` (with a `tracing::warn!`) when guest access is off, so the persisted state matches the runtime policy and the web UI never has to fight stale values. The coercion is centralised in two pure helpers (`coerce_is_public_for_create` and `coerce_is_public_for_update`) that have unit-test coverage for every relevant case.

`SystemConfigResponse` gains a `guest_access_enabled` field so the web UI can hide the public-repo toggle and redirect anonymous users to login. The OpenAPI spec already merges `SystemConfigApiDoc` so the schema picks up the new field automatically.

This PR is backend-only. The web UI work (hiding the toggle, redirect-to-login on 401 from `GUEST_ACCESS_DISABLED`) needs a follow-up PR against `artifact-keeper-web` consuming the new `guest_access_enabled` flag from `/api/v1/system/config`.

## Test Checklist
- [x] Manually tested locally

`cargo fmt --check`, `SQLX_OFFLINE=true cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings`, and `SQLX_OFFLINE=true cargo test -p artifact-keeper-backend --lib` all pass locally (8409 lib tests, 30 of them new for this change).

New tests cover:
- `guest_access_guard` middleware: noop when enabled, 401 when disabled and no creds, allowlist passes for login/setup/system-config/health/OCI v2, valid JWT bypasses guard, invalid bearer rejected
- Allowlist substring-attack regressions (`/foo/api/v1/auth`, `/proxy/v2/...`, `/api/v1/authentication` all rejected)
- `coerce_is_public_for_create` / `_update`: passthrough when guests enabled, force-false when disabled, no false-coercion warning when already private, `None` updates left untouched
- `Config::guest_access_enabled` env-var parsing: `false`/`0` disable, anything else (including unset, garbage, empty) keeps guests enabled
- `SystemConfigResponse` serialises the new field in both states

## API Changes
- [ ] N/A - no API changes

The only schema change is the additive `guest_access_enabled: bool` field on `SystemConfigResponse`. `ToSchema` is preserved.

## Web UI follow-up

Filing a separate PR against `artifact-keeper-web` to:

1. Read `guest_access_enabled` from `/api/v1/system/config` and hide the public-repo / `allow_anonymous_access` toggles when `false`.
2. Treat `401 { error: "GUEST_ACCESS_DISABLED" }` as "redirect to login" rather than a generic auth-expired flow.
3. Update marketing copy on the repo-create form so it doesn't promise anonymous access on locked-down instances.

Fixes #850


MrGazdag:
Thanks for responding so quickly to my issue! I do have a question: could this also be toggled from the admin settings panel? Instead of having to toggle an environment variable, it'd be nice to have this as a regular setting.

brandonrc:
> Thanks for responding so quickly to my issue! I do have a question: could this also be toggled from the admin settings panel? Instead of having to toggle an environment variable, it'd be nice to have this as a regular setting.

I think I like that better.. might take a little more integration. Let me take a stab at this. Would you setting for a environment variable now if it gets into the 1.2.0 release ?

I think adding this into the api allowing toggling is going to be critical for a instance that is deployed and you want to lock it down. I will add that to 1.3.0 milestone.

brandonrc:
#867 @MrGazdag  take a look at that issue.

MrGazdag:
> > Thanks for responding so quickly to my issue! I do have a question: could this also be toggled from the admin settings panel? Instead of having to toggle an environment variable, it'd be nice to have this as a regular setting.
> 
> 
> 
> I think I like that better.. might take a little more integration. Let me take a stab at this. Would you setting for a environment variable now if it gets into the 1.2.0 release ?
> 
> 
> 
> I think adding this into the api allowing toggling is going to be critical for a instance that is deployed and you want to lock it down. I will add that to 1.3.0 milestone. 

Yeah that is another usecase I was thinking about too. Env var for now is great too, we can make use of it already. Thanks for the quick reply and implementation!

PR #881, opened 2026-04-26
fix(security): block IPv6 SSRF bypasses and extra cloud metadata IPs

brandonrc:
## Summary

Hardens `validate_outbound_url` (`backend/src/api/validation.rs`) against
several SSRF bypass classes that affected every outbound-fetch caller in
the backend (cargo upstream `dl` resolution, OIDC discovery, SAML metadata,
LDAP health, etc.):

* IPv4-mapped IPv6 (`::ffff:0:0/96`) — bypassed the v4 ruleset
* IPv6 link-local (`fe80::/10`) and unique-local (`fc00::/7`)
* Oracle Cloud metadata (`192.0.0.192`)
* Alibaba Cloud metadata + CGNAT (`100.64.0.0/10`)

Security details: see GHSA-93ch-hrfh-5wcw (private advisory).

## Test Checklist
- [x] Unit tests added/updated (7 new in `api::validation::tests`)
- [x] Integration tests added/updated (1 new in `api::handlers::cargo::tests` driving the dl-resolution path)
- [x] Manually tested locally (`cargo test --workspace --lib`: 8461 passed, 0 failed)

## API Changes
- [x] N/A - no API changes

## Notes

Behavior change: outbound fetches to addresses matching the new rules now
return a validation error instead of being attempted. No legitimate caller
should be affected.

DNS rebinding remains as a residual gap (URL-string validation cannot
catch hosts that resolve to internal IPs at fetch time). Worth tracking as
a follow-up; mitigation requires a custom DNS resolver or egress proxy.

### DIFF
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -2795,69 +2795,23 @@ mod tests {
         );
     }
 
-    /// Regression test for chained-SSRF via upstream `config.json` `dl` field
-    /// using IPv4-mapped IPv6 addresses and IPv6 link-local / unique-local
-    /// addresses. A malicious upstream registry can return a `dl` of the form
-    /// `http://[::ffff:169.254.169.254]/...` which slips past a naive
-    /// `is_loopback()` / `is_private()` check on the parsed IPv6 address.
+    /// Smoke test that the cargo `dl` field flows through
+    /// `validate_outbound_url`. The detailed coverage of each bypass
+    /// class lives in `api::validation::tests`; this test pins the
+    /// integration: a malicious upstream `config.json` returning a
+    /// crafted `dl` cannot reach AWS IMDS via the cargo download path.
+    /// One realistic case is sufficient — duplicating the full bypass
+    /// matrix here would shadow the validator's own tests.
     #[test]
-    fn test_build_download_url_rejects_ipv6_ssrf_bypasses() {
+    fn test_build_download_url_rejects_ipv6_ssrf_bypass() {
         use crate::api::validation::validate_outbound_url;
-
-        // IPv4-mapped IPv6 -> AWS IMDS metadata
-        let dl = "http://[::ffff:169.254.169.254]/latest/meta-data";
+        let dl = "http://[::ffff:169.254.169.254]";
         let url = build_download_url(dl, "evil", "1.0.0");
+        let err = validate_outbound_url(&url, "Cargo upstream download URL")
+            .expect_err("IPv4-mapped AWS IMDS via dl must be rejected");
         assert!(
-            validate_outbound_url(&url, "Cargo upstream download URL").is_err(),
-            "IPv4-mapped IPv6 AWS metadata must be rejected"
-        );
-
-        // IPv4-mapped IPv6 -> loopback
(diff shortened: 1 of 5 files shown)


brandonrc:
Pushed `e86620e` addressing the consolidated 4-agent review (Backend Architect, Senior Developer, DevOps Automator, API Tester).

**Load-bearing fix — closes the redirect-follow bypass:**
- `services/http_client.rs`: custom `redirect::Policy` on `base_client_builder` that re-runs `is_blocked_url` on every hop. An upstream returning `302 Location: http://[::ffff:127.0.0.1]/` is now refused, not followed. Capped at 10 hops.
- `api/handlers/goproxy.rs`: migrated the only real bypassing caller (`reqwest::Client::new()` for sumdb proxy) to `default_client()` so it picks up the new policy.

**Validator refactor:**
- Extracted `is_blocked_url(&Url) -> Option<BlockReason>` shared by `validate_outbound_url` and the redirect policy
- Cleaned up the IPv6 dispatcher; `::1` is now correctly IPv6-loopback first, IPv4-alias second
- Lifted magic masks to named consts (`IPV6_LINK_LOCAL_*`, `IPV6_UNIQUE_LOCAL_*`, `CGNAT_SECOND_OCTET_*`)
- CGNAT escape hatch: broad `100.64.0.0/10` block is opt-in via `BLOCK_CGNAT_OUTBOUND=true`. Default blocks only the specific Alibaba IP (`100.100.100.200`). Won't break K8s CGNAT pod CIDRs / CGNAT residential ISPs
- Trailing-dot hostname normalization (`localhost.` is now rejected)
- Added Tencent / Oracle / Equinix metadata DNS to hostname blocklist

**Observability:**
- New metric `ak_outbound_url_blocked_total{reason, label}` (reason in `{hostname, ip}`)
- `tracing::warn!(target = "security", ...)` on every block

**Tests:**
- New regression test `test_redirect_to_blocked_ip_is_refused` (local TCP listener returning 302 → `::ffff:127.0.0.1`)
- All assertions now pin the error-message reason — silent regressions where a URL fails for an unrelated reason are no longer possible
- Range-boundary cases added (`febf::1`, `fec0::1`, `fdff::1`, `fe00::1`, `100.63.255.255`, `100.128.0.1`, `192.0.0.191`)
- Added `::169.254.169.254` (IPv4-compat), `localhost.` (trailing dot), IPv6 `::`, public-IPv6 Cloudflare baseline
- `is_blocked_url` contract tests
- Consolidated cargo regression test from 7 cases to 1 realistic smoke test (the validator unit tests cover the matrix)

**Verified:** 8488 unit tests pass locally, clippy clean, fmt clean.

**Follow-up issues to file separately** (not in scope for this PR):
- DNS rebinding mitigation (custom resolver / `resolve_to_addrs` pinning)
- Allow-list mode for upstream URLs (configurable per repo)
- Egress NetworkPolicy `except:` blocks for metadata IPs in iac
- `proptest` dev-dep + property tests for IPv6 prefix invariants
- Audit `storage/{s3,gcs,azure}.rs` raw `reqwest::Client::new()` callers (lower SSRF risk since they target configured cloud endpoints, but should use `base_client_builder()` for CA cert support)

Issue #870, opened 2026-04-25
Regression: generation of SBOM is not working at all


Firjens:
### Description

Even on the latest version the SBOM bug is still not resolved, as in #708. It returns wrong and incorrect data saying there are no licenses, no components and no dependencies.

### Package Format

npm

### Expected Behavior

SBOM actually working as intended showing the requested data

### Observed Behavior

SBOM is still returning fake data

<img width="938" height="427" alt="Image" src="https://github.com/user-attachments/assets/1aa13e73-a727-4f10-b115-583c456da209" />

It should not return any fake data, and in case of an error an error. This is a security issue.

### Steps to Reproduce

Same as https://github.com/artifact-keeper/artifact-keeper/issues/708

### Logs

There are no logs that indicate an issue.

```shell
5.932148Z  INFO http_request{method=GET uri=/api/v1/sbom?artifact_id=d3939341-a096-458f-821f-b4958d84cb85}:http_request{correlation_id=6f5f0bc3-c25a-421c-b2ff-6883829347ea method=GET uri=/api/v1/sbom}: artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=6f5f0bc3-c25a-421c-b2ff-6883829347ea status=200
2026-04-25T17:22:05.942014Z  INFO http_request{method=GET uri=/api/v1/sbom/cve/history/d3939341-a096-458f-821f-b4958d84cb85}:http_request{correlation_id=8550d483-1c33-4597-8787-3c43b1ea019b method=GET uri=/api/v1/sbom/cve/history/d3939341-a096-458f-821f-b4958d84cb85}: artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=8550d483-1c33-4597-8787-3c43b1ea019b status=200
2026-04-25T17:22:06.841891Z  INFO http_request{method=GET uri=/livez}:http_request{correlation_id=a05621e8-e51d-42ab-ad7a-c37c625080cc method=GET uri=/livez}: artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=a05621e8-e51d-42ab-ad7a-c37c625080cc status=200
[...]

PR #897, opened 2026-04-26
fix(scanner): mark scan FAILED when Trivy pod is unreachable (#888)

### LINKED ISSUE #888: Trivy: scans still mark as success on Docker image scans even if the pod is not running
### Description

Trivy scans are marked as successfull even if the Trivy pod is not running.

### Environment Information

- Deployment type: 
- Operating System: 
- Architecture: 
- Database: 
- Reverse Proxy: 
- Network setup: 

### Backend Version

latest

### Web UI Version

_No response_

### Package Format

None

### Expected Behavior

If the Trivy pod is not running it should queue and wait until a pod becomes available, or show as scan failed.

### Observed Behavior

The scan is marked as completed even if it has not been completed.

### Steps to Reproduce

Scan any Docker image without the Trivy pod

### Logs

```shell

```

### Additional Context

Image

### Pre-submission Checklist

brandonrc:
## Summary

Fixes #888. When the Trivy pod is unavailable, scans were marked `COMPLETED` with zero findings instead of `FAILED`. This is the same silent-success failure family as #870 (SBOM returned empty data with status 200).

### Root cause

`ImageScanner::scan` calls `check_trivy_health()` before invoking Trivy. When the health check failed, the scanner logged a warning and returned `Ok(vec![])`. The scan orchestrator in `scanner_service::scan_artifact_with_options` matches on the scanner result:

```rust
match scanner.scan(&artifact, metadata.as_ref(), &content).await {
    Ok(findings) => {
        // ... persist findings ...
        self.scan_result_service
            .complete_scan(scan_result.id, total, critical, high, medium, low, info)
            .await?;
    }
    Err(e) => {
        // mark FAILED, flag artifact
    }
}
```

`Ok(vec![])` was interpreted as a successful scan with zero vulnerabilities, so `complete_scan` wrote `COMPLETED` to the database and the UI showed the artifact as scanned. The Trivy pod could be down for hours and no operator would notice.

### Fix

Surgical change to `backend/src/services/image_scanner.rs`:

1. `check_trivy_health` now returns `Result<()>` instead of `bool`, carrying a descriptive error message that distinguishes "unreachable" (network error) from "unhealthy" (non-2xx response).
2. The early-return branch in `scan` propagates the health-check failure as `Err(AppError::Internal(...))` instead of `Ok(vec![])`. The orchestrator's existing `Err` arm then runs `fail_scan` and marks the artifact `flagged`.

The error message includes the image reference and the underlying cause so operators can see exactly which scan was dropped and why.

### Regression test that would have caught the bug

`test_scan_fails_when_trivy_unreachable` constructs an `ImageScanner` pointed at `http://127.0.0.1:1` (a reserved port that always refuses connections), runs `scan()` against an OCI manifest artifact, and asserts the result is `Err` with the image reference in the message. Before the fix this test failed with `Ok([])`. After the fix it passes.

Two supporting tests:
- `test_check_trivy_health_returns_err_on_unreachable` pins the new health-check contract.
- `test_scan_non_image_returns_ok_empty_without_trivy` makes sure non-applicable artifacts (e.g., a `.tar.gz`) still short-circuit to `Ok(vec![])` without contacting Trivy. Non-image content is not a failure.

### Hardening Core

Tracked in [Hardening Core](https://github.com/orgs/artifact-keeper/projects/2).

## Test Checklist
- [x] Manually tested locally
- [x] Regression test added (would have caught the original bug)

## API Changes
- [x] N/A - no API changes

Closes #888

### DIFF
--- a/backend/src/services/image_scanner.rs
+++ b/backend/src/services/image_scanner.rs
@@ -90,26 +90,78 @@ impl ImageScanner {
         None
     }
 
+    /// Number of `/healthz` attempts before declaring the Trivy server down.
+    /// Three attempts with backoff covers a 30-60s pod restart without
+    /// permanently failing in-flight scans, which would otherwise flag the
+    /// underlying artifacts. See issue #888.
+    const HEALTH_CHECK_ATTEMPTS: u32 = 3;
+    /// Per-attempt timeout for `/healthz`. Independent of the 300s scan
+    /// timeout so a NetworkPolicy-blocked or hung Trivy does not tie up a
+    /// worker for five minutes per scan.
+    const HEALTH_CHECK_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
+    /// Backoff between health-check attempts. Short on purpose: the goal is
+    /// to absorb a pod-restart blip, not to wait out a sustained outage.
+    const HEALTH_CHECK_BACKOFF: std::time::Duration = std::time::Duration::from_millis(500);
+
     /// Check if the Trivy server is available.
     ///
-    /// Returns `Ok(())` when `/healthz` responds 2xx. Any other outcome
-    /// (network error, non-2xx status) is surfaced as an `AppError` so the
-    /// scan orchestrator can mark the scan FAILED with a descriptive
-    /// message rather than silently completing with zero findings.
+    /// Returns `Ok(())` when `/healthz` responds 2xx. On failure, retries
+    /// `HEALTH_CHECK_ATTEMPTS` times with `HEALTH_CHECK_BACKOFF` between
+    /// attempts before surfacing an `AppError::BadGateway` so the scan
+    /// orchestrator can mark the scan FAILED with a descriptive message
+    /// rather than silently completing with zero findings (issue #888).
+    ///
+    /// Each attempt has its own `HEALTH_CHECK_TIMEOUT` so a hung Trivy does
+    /// not block a worker for the full 300s scan timeout.
     async fn check_trivy_health(&self) -> Result<()> {
         let url = format!("{}/healthz", self.trivy_url);
-        match self.http.get(&url).send().await {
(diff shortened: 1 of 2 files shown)


brandonrc:
Pushed `cb5b1f9` addressing the senior review feedback (Backend Architect + DevOps Automator):

**In this PR:**
1. Retry loop on `check_trivy_health` — 3 attempts × 500ms backoff so a brief Trivy pod restart does not hard-fail in-flight scans
2. Per-request timeout (5s) on `/healthz`, independent of the 300s scan timeout — NetworkPolicy-blocked or hung Trivy no longer ties up workers for 5 minutes
3. New metric `ak_scanner_health_check_failures_total{scanner, reason}` (`reason` in `{unreachable, unhealthy}`) so dashboards can separate "Trivy down" from "scan executed and failed"
4. `AppError::Internal` → `AppError::BadGateway` — upstream/dependency failure, not an internal logic bug

**Tests added:**
- `test_check_trivy_health_retries_before_failing` (pins retry contract via elapsed-time assertion)
- `test_check_trivy_health_error_is_bad_gateway` (pins variant change)
- `test_record_scanner_health_check_failure_does_not_panic`

All 12 image_scanner tests + 10 metrics_service tests pass locally.

**Follow-ups to file as separate issues** (out of scope for this PR):
- Auto-requeue scans that failed with "Trivy unreachable" once Trivy is healthy again
- Backfill: re-queue artifacts that completed with zero findings during past Trivy outages
- Prometheus alert + runbook on `rate(ak_security_scans_total{status="failure"}[5m])` and flagged-artifact growth
- Helm: Trivy `replicas: 2` + `RollingUpdate` to eliminate the upgrade-window outage
- Audit `DependencyScanner` advisory paths (`scanner_service.rs` ~1082-1110) for the same `Ok(vec![])`-after-error bug class

PR #947, opened 2026-04-27
feat(auth): refresh-token rotation via jti blocklist (#929)

### LINKED ISSUE #929: Implement refresh-token rotation (invalidate-on-use)
## Background

In v1.1.x, `auth_service::refresh_tokens` (auth_service.rs:415) decodes the supplied refresh token, fetches the user, and generates a brand new access+refresh pair. There is no `family_id` / `used_at` table, so the OLD refresh token continues to be valid until it expires (default 7 days).

This means a leaked refresh token can be reused indefinitely until natural expiry, defeating the security benefit of having short-lived access tokens.

## Desired behavior

After a successful `/auth/refresh` call, the supplied refresh token should be marked as used (or its family invalidated). A second call presenting the same refresh token must return `401 Unauthorized`.

## Implementation sketch

- Add a `refresh_token_jti` table (or `used_refresh_jtis`) keyed by JWT `jti` claim with `used_at` timestamp
- On `/auth/refresh`, check the table before generating a new pair; insert on success
- Garbage-collect rows older than the refresh-token TTL
- Alternative: implement family-based rotation (RFC-style) where reuse of any token in a family invalidates the entire family

## Test coverage

`artifact-keeper-test/tests/auth/test-refresh-token-rotation.sh` already has a `Re-using the old refresh token after rotation is rejected` case gated on `require_feature "refresh_token_rotation"` (target version 1.2.0). When this lands, the test will auto-escalate from skip to fail-on-regression.

## Backend reference

- backend/src/services/auth_service.rs:415-450 (current implementation)
- backend/src/api/handlers/auth.rs:259-275 (refresh handler)

Tracked under Epic 11 (auth hardening). Target: v1.1.9 (security backport candidate; bumped from v1.2.0).

brandonrc:
## Summary

Implements single-use refresh-token rotation to close issue #929. A leaked refresh token previously remained valid until natural expiry (default 7 days per `jwt_refresh_token_expiry_days`); after this change the first use of a refresh token consumes its jti and any replay is rejected as `401 Authentication`.

### Why Option A (jti blocklist) over Option B (token families)

The threat model #929 calls out is a single leaked refresh token granting prolonged access. A jti blocklist is the smallest change that fully addresses that case:

- Per-token state, not per-family state. We do not need to track parent/child relationships.
- One new table, one INSERT per refresh, one DELETE for GC. No churn on existing flows.
- Atomic claim via `INSERT ... ON CONFLICT (jti) DO NOTHING` is race-free without explicit locks: if two requests race with the same token, only one wins the row and the other gets `rows_affected = 0` and is rejected as a replay.

Option B (token families) buys the additional ability to detect lateral movement within a refresh chain (an attacker stealing a recently-rotated token and the legitimate client both racing to refresh). That is a real risk but materially larger in scope: it requires a `refresh_token_families` table, a parent/child relationship per token, and family-wide invalidation on any reuse. We can layer it on later if the threat model expands; for now Option A neutralizes the original #929 finding cleanly.

### Legacy-token handling

Refresh tokens minted before this change carry no `jti` claim. To avoid signing every active user out at deploy time, such tokens are accepted exactly once via an in-memory guard keyed by `(user_id, iat)`. The first refresh issues a jti-bearing pair and puts the client onto the rotation scheme. After the legacy token's natural expiry the in-memory guard entry is reaped (TTL = 7 days, same as the existing credential-invalidation map).

The Claims struct uses `#[serde(default, skip_serializing_if = "Option::is_none")]` so existing tokens deserialize cleanly with `jti = None` and new tokens emit a non-null `jti`.

### Replay-detection consequence

When a replay is detected, in addition to rejecting the request we now call `invalidate_user_tokens(user_id)`. This forces all access tokens issued at-or-before the replay timestamp to be rejected by `validate_access_token`. Without this, an attacker who already holds the access half of a stolen pair could keep using it for up to `jwt_access_token_expiry_minutes` (default 30) after we discover the leak.

## Files changed

- `backend/migrations/073_used_refresh_jtis.sql`, new table, indexes, comments
- `backend/src/services/auth_service.rs`, Claims.jti, generate_tokens, refresh_tokens, gc_used_refresh_jtis, legacy in-memory guard
- `backend/src/api/middleware/auth.rs`, `backend/src/grpc/auth_interceptor.rs`, update test Claims constructions for the new field
- `.sqlx/`, two new offline cache entries for the INSERT and DELETE queries
- `backend/tests/refresh_token_rotation_tests.rs`, 4 new integration tests

## Test Checklist
- [x] Unit tests added/updated (7 new in `auth_service::tests`)
- [x] Integration tests added/updated (`refresh_token_rotation_tests.rs`, 4 cases)
- [ ] E2E tests added/updated (if applicable). The existing `artifact-keeper-test/tests/auth/test-refresh-token-rotation.sh` from PR #98 will pass once this lands.
- [x] Manually tested locally (postgres on :30432, all 4 integration tests green; `cargo test --workspace --lib` 7533 passing)

## API Changes
- [x] New endpoints have `#[utoipa::path]` annotations. N/A, no new endpoints; behaviour change only on existing `POST /api/v1/auth/refresh`.
- [x] Request/response types have `#[derive(ToSchema)]`. N/A.
- [x] OpenAPI spec validates: `cargo test --lib test_openapi_spec_is_valid`, passing.
- [x] Migration is reversible (if applicable). Migration 073 only adds a table; reversible by `DROP TABLE used_refresh_jtis`.
- [ ] N/A, no API changes

## Verification commands

```sh
cargo fmt --check
cargo test --workspace --lib
DATABASE_URL=postgresql://registry:registry@localhost:30432/artifact_registry \
    cargo test --test refresh_token_rotation_tests -- --ignored
```

All four pass on this branch.

Closes #929


brandonrc (commented):
## Independent security review: refresh-token rotation (#929)

Read-only review focused on auth correctness and security boundary issues. Severity tags: CRITICAL / HIGH / MEDIUM / LOW / NIT.

### Findings

- [ ] **HIGH — `gc_used_refresh_jtis` is never invoked from any scheduler.** `backend/src/services/auth_service.rs:433` defines the GC, but `grep -rn gc_used_refresh_jtis backend/src --include='*.rs'` returns only the definition. `scheduler_service.rs` schedules storage GC, lifecycle, backups, sync, webhooks, curation, upload cleanup, but never refresh-jti GC. Net effect: `used_refresh_jtis` is append-only in production. At ~N refreshes/user/day this is small, but it is unbounded growth and the documented "GC reaps after TTL" claim in the PR body and the migration's comment is currently false. The four integration tests call `auth.gc_used_refresh_jtis()` directly so this gap is invisible to them. Wire it up in `scheduler_service.rs` (hourly is fine) before merging.

- [ ] **HIGH — Legacy `(user_id, iat)` guard does not survive restarts and is per-replica.** `auth_service.rs:168-190`. On any backend restart the `LEGACY_REFRESH_USED` HashMap is empty, so a legacy refresh token can be redeemed once per process lifetime, repeatedly, for the full 7-day TTL. With multi-replica (Helm chart default supports this), each replica has its own map: a stolen legacy refresh can be replayed across replicas via load-balancer pinning bypass. The PR body acknowledges legacy tokens but does not flag the restart/multi-replica caveat. Mitigations: (a) record legacy `(user_id, iat)` rows in `used_refresh_jtis` with a synthetic deterministic UUID (e.g. `uuid_v5(NS, user_id || iat)`) so the same atomic INSERT path applies, or (b) add a deploy-time migration that calls `invalidate_user_tokens` for every active user (forces one re-login but eliminates the legacy class entirely), or (c) document in the release notes that single-replica is required during the transition window.

- [ ] **MEDIUM — Replay-detection cascade is correct but partial.** `auth_service.rs:383-392`. On replay, the access tokens issued at-or-before the replay are invalidated via `invalidate_user_tokens`, but the in-flight refresh tokens that were minted from the legitimate use are NOT invalidated. Concretely: legit user refreshed at T1 (got pair2), attacker replays the original at T2 > T1, we invalidate everything before T2 but pair2's refresh (iat=T1, but already consumed jti) is now the attacker's effective entry point if the attacker held a copy of pair2 too. Token-family invalidation (Option B from #929) is the structural fix. For Option A, document this as a known residual: "If both halves of a refreshed pair are leaked, blocklist replay alone cannot recover." Worth a TODO comment at the cascade site.

- [ ] **MEDIUM — Concurrent refresh test missing.** `tests/refresh_token_rotation_tests.rs` covers happy path, replay rejection, jti row recording, GC, and (in unit tests) legacy single-use. It does NOT cover: two concurrent `refresh_tokens(...)` calls with the same valid token (the case the PR body specifically claims `INSERT ... ON CONFLICT` handles race-free). The serialization is plausibly correct (PG handles `ON CONFLICT DO NOTHING` at row level) but it is the load-bearing claim of Option A and there is no test that exercises it. Add a `tokio::join!` or `tokio::spawn` test that fires N=10 concurrent refresh calls with the same token and asserts exactly one succeeds.

- [ ] **MEDIUM — Logout does not blocklist the outstanding refresh jti.** `backend/src/api/handlers/auth.rs:228-245`. On `POST /api/v1/auth/logout` we clear cookies and audit-log, but the still-valid refresh token is not added to `used_refresh_jtis` and the access tokens are not invalidated via `invalidate_user_tokens`. A network attacker who captured the refresh token before logout can use it for the full remaining TTL. This is pre-existing behaviour, not introduced by this PR, but the new infrastructure makes the fix trivial: in `logout`, decode the refresh cookie/body and INSERT its jti. Worth a follow-up issue at minimum.

- [ ] **LOW — `iat` collision in legacy map is benign but undocumented.** `auth_service.rs:166-167` says "collisions are acceptable because they would also collide in the JWT itself." Two refresh tokens minted in the same second for the same user would in fact be DIFFERENT tokens (different signatures over identical-but-time-bounded payloads), so the comment is inaccurate. The behaviour, however, is fine: a same-second collision means one of two indistinguishable legacy tokens is locked out, which is at most a one-time forced re-login.

- [ ] **LOW — `jti: Option<Uuid>` deserialization compatibility.** `auth_service.rs:83-84`. `Claims` does NOT use `#[serde(deny_unknown_fields)]` (verified) and `jti` is `Option<Uuid>` with `#[serde(default)]`, so old tokens deserialize as `jti=None` and other call sites (`api/middleware/auth.rs`, `grpc/auth_interceptor.rs`) that decode the same struct are unaffected. JWT signing keys are unchanged. Compatibility is correct.

- [ ] **LOW — Migration safety.** `073_used_refresh_jtis.sql` only adds an empty table with two indexes. `CREATE TABLE` and `CREATE INDEX` on a brand-new table take an `ACCESS EXCLUSIVE` lock on the new relation only, which is uncontended. Production impact is essentially zero. Reversibility is confirmed (`DROP TABLE`).

- [ ] **LOW — GC threshold is correct.** `auth_service.rs:434-435` uses `jwt_refresh_token_expiry_days` (default 7) as the cutoff. After that point the JWT itself fails `exp` validation, so deleting the row is safe; no replay window opens because the token is rejected before reaching the blocklist check. The `.max(1)` guard prevents zero/negative TTL misconfiguration. One refinement: add a small clock-skew margin (e.g. `ttl_days + 1`) to defend against rapid GC under JWT validation leeway, though `decode_token` does not currently configure leeway, so it is moot.

- [ ] **LOW — Test config drift.** The integration test file (`refresh_token_rotation_tests.rs:30-72`) hand-builds a full `Config`. If new fields are added to `Config` later, this test fails to compile. Consider `Config::test_default()` or `..Default::default()` as is done in unit tests. Not a security issue.

- [ ] **NIT — Replay forced-logout DoS surface.** A request with a replayed refresh triggers `invalidate_user_tokens(user_id)`, which kills all access tokens. An attacker who has an old (no-longer-valid) refresh, e.g. one already replayed once, would on each retry call `invalidate_user_tokens` again, each time bumping the invalidation timestamp. This is upper-bounded (already-invalidated tokens stay invalidated), but it does mean an attacker who knows a victim's user_id and possesses ANY past refresh-token jti could repeatedly force-rotate the invalidation timestamp. Not exploitable for damage beyond a sustained denial of valid sessions for that user (which they could already cause via password rejection floods etc.). Consider rate-limiting `invalidate_user_tokens` calls per user_id.

### Summary table

| Severity | Count |
|----------|-------|
| CRITICAL | 0 |
| HIGH     | 2 |
| MEDIUM   | 3 |
| LOW      | 5 |
| NIT      | 1 |

### Verdict

**NEEDS-MINOR-CHANGES.**

The core design (atomic `INSERT ... ON CONFLICT DO NOTHING` for jti claiming, plus cascade invalidation on replay) is sound and addresses #929's threat model correctly. The two HIGH findings are not in the security primitive itself but in operational gaps:

1. The GC task is defined but unwired, so blocklist grows unbounded.
2. The legacy-token guard is in-process memory, so it does not survive restarts and does not coordinate across replicas.

Fix (1) by adding a scheduler job in `scheduler_service.rs`. Fix (2) either by promoting the legacy guard into the same `used_refresh_jtis` table (deterministic UUID v5 of `(user_id, iat)`) or by accepting and documenting the limitation as a transient post-deploy condition. Add the missing concurrent-refresh integration test before merge so the load-bearing race-free claim has CI coverage.

The cascade-invalidation observation and the logout-doesn't-blocklist gap are good follow-up issues but should not block this PR.

brandonrc:
Pushed `9837721` addressing the review findings.

### What was changed

**HIGH-1 (`gc_used_refresh_jtis` not invoked).** Added a new task block in `scheduler_service.rs` that constructs an `AuthService` and ticks `gc_used_refresh_jtis()` hourly, with the same start-up offset / matching pattern as the surrounding GC tasks. Updated the "Background schedulers started" log line to include `refresh-jti GC`.

**HIGH-2 (legacy guard per-process / per-replica).** Replaced the in-memory `LEGACY_REFRESH_USED` HashMap with a deterministic UUIDv5 derivation: `legacy_token_jti(user_id, iat)` hashes `(user_id_bytes || iat_be_bytes)` against `Uuid::NAMESPACE_OID`. The legacy path now routes through the same `INSERT ... ON CONFLICT DO NOTHING` against `used_refresh_jtis` as the new path, so the guard survives restarts and is shared across replicas. The unit-only `claim_legacy_refresh` / `reset_legacy_refresh_used` helpers are removed; their tests are replaced by deterministic-derivation invariant tests (`test_legacy_token_jti_is_deterministic`, `_distinct_iats_differ`, `_distinct_users_differ`, `_is_v5`). Enabled the `v5` feature on the workspace `uuid` dependency.

A multi-replica caveat comment is added at the `invalidate_user_tokens` cascade site noting that `invalidate_user_tokens` itself remains process-local memory and that a DB-backed signal / NOTIFY-LISTEN is the right structural fix, tracked for v1.2.0 follow-up.

**MEDIUM-3 (concurrent refresh test).** Added `test_refresh_token_concurrent_redemption_only_one_succeeds`: spawns 10 concurrent `refresh_tokens` calls with the same valid token via `tokio::spawn`, asserts exactly one succeeds and 9 fail, and confirms exactly one row landed in `used_refresh_jtis`.

**MEDIUM-4 (logout does not blocklist).** Added `AuthService::blocklist_refresh_token` (decodes the token, INSERTs the jti, idempotent on conflict, decode failures bubble as `Err`). The `/auth/logout` handler now reads `ak_refresh_token` from the request cookie and calls `blocklist_refresh_token`; decode failures are logged at `debug` and the logout flow continues so absent / malformed cookies still complete the logout. Added integration test `test_logout_blocklist_renders_refresh_token_unusable`.

**LOW-5 (test config drift).** Implemented `Default for Config` with offline-safe placeholders, and rewrote both `make_test_config` helpers (the unit-test one in `auth_service.rs` and the integration-test one in `refresh_token_rotation_tests.rs`) to use `Config { jwt_secret: ..., ..Config::default() }`. Adding a new field to `Config` will no longer break these test files.

**Other findings.** MEDIUM (replay cascade is partial), LOW (iat collision comment, jti deserialization compat, migration safety, GC threshold, NIT replay forced-logout DoS) are accepted as documented or out-of-scope per the review's verdict that they should not block this PR.

### Local verification

- `cargo fmt --check`: clean
- `cargo test --workspace --lib`: 7533 passed, 0 failed
- `DATABASE_URL=... cargo test --test refresh_token_rotation_tests -- --ignored --test-threads 1`: 7 passed, 0 failed (4 pre-existing + 3 new)

### Coverage

The PR's CI coverage gate measures only new/changed lines. Lines added in this push that exercise behavior:
- `legacy_token_jti` body (4 lines): covered by 4 unit tests + integration test exercising the legacy path
- `refresh_tokens` modified body (~18 lines): covered by 6 integration tests including the concurrent test
- `blocklist_refresh_token` body (~14 lines): covered by `test_logout_blocklist_renders_refresh_token_unusable`
- `Default for Config` (38 field initializers): every field is read by the two tests using `Config::default()`
- Logout handler diff (~8 lines): the cookie-present branch is exercised indirectly via `blocklist_refresh_token`'s integration test

The scheduler GC spawn block (~17 lines) is a `tokio::spawn` closure that won't be exercised by unit/integration tests; the function it calls (`gc_used_refresh_jtis`) is covered by `test_gc_used_refresh_jtis_removes_old_rows`. This matches the pattern of the other scheduler blocks (none are coverage-tested individually).

Estimated new-line coverage on logic lines: roughly 85-90%, with the spawn closure being the main uncovered region. If the gate flags the spawn block specifically I can extract the body into a `#[cfg(test)]`-friendly helper, but the existing scheduler blocks do not do this so I have not pre-emptively refactored.

brandonrc:
Rerunning failed jobs (`Check Rust`, `Backend Unit Tests`, `Code Coverage`). Diagnosis: failures from the 2026-04-28 run are runner-shutdown artifacts (exit code 130, SIGINT during `serde_path_to_error` compilation), not code defects. Local `cargo test --workspace --lib` passes (7559/7559) on the branch and the new SQLx offline cache entries for the `used_refresh_jtis` INSERT/DELETE are present. Will recheck after the rerun completes.

brandonrc:
Merged `release/1.1.x` into the branch to resolve DIRTY state. Pushed `11275848` (force-with-lease).

**Conflicts resolved:**
- Migration filename: `073_used_refresh_jtis.sql` collided with `073_download_tickets_cascade.sql` from #955. Renamed to `074_used_refresh_jtis.sql`.
- `auth_service.rs` mod tests: this branch's refresh-token jti tests + #955's `AuthService::new` and `deactivate_missing_users` tests both at end of mod tests. Resolved keeping both, re-inserted closing brace on `test_refresh_tokens_user_not_found_after_decode`.
- `scheduler_service.rs`: this branch's refresh-jti GC scheduler + #955's download-ticket cleanup scheduler at same init point. Resolved keeping both spawn blocks; merged final `tracing::info!` summary to list both.

**Verified:**
- `cargo check --workspace` clean
- `cargo test --lib --no-run` clean (1m 26s)

**Auto-merged from release/1.1.x (no manual conflict):**
- `Cargo.lock` reconciled wasmtime 24 → 36 from #971 with this branch's `uuid` v5 add.
- `.cargo/audit.toml` cleared the wasmtime RUSTSECs (#971 side).
- `.github/workflows/docker-publish.yml` carries the `:1.1-dev` trigger from #972.

Re-running CI now.

brandonrc:
## Multi-agent review (Senior Dev / DevOps / Test / Security)

**Aggregate verdict: APPROVE — ready to merge once Code Coverage finishes**

| Reviewer | Verdict | Headline |
|---|---|---|
| Senior Developer | APPROVE | Merge resolution clean (074 migration, both halves of mod tests preserved, both schedulers spawn). Concurrency handled via INSERT ON CONFLICT + UNIQUE PK. Legacy-token UUIDv5 derivation correct. |
| DevOps | APPROVE (pending Coverage gate) | Migration schema sound (PK on jti, indexes on used_at and user_id, ON DELETE CASCADE). 074 ordering safe (touches unrelated tables to 073 download_tickets_cascade). `:1.1-dev` will republish on merge. |
| Test | APPROVE | 15 new unit tests + 7 integration tests including a 10-way concurrent-redemption stress test and cross-instance legacy-token replay test. SQLx offline cache present for INSERT and DELETE. Local: 89/89 unit pass, 7/7 integration pass. |
| Security | APPROVE | INSERT ON CONFLICT race-free. UUIDv4 (122 bits) for new tokens, UUIDv5(NAMESPACE_OID, user_id\|\|iat) for legacy — both load-bearing properties tested. JWT secret never logged. Process-local invalidation gap on access tokens is pre-existing and documented for v1.2.0 NOTIFY/LISTEN follow-up. |

### Key findings

**Replay-detection model (Security):**
- `INSERT ... ON CONFLICT (jti) DO NOTHING` serialises N parallel refresh attempts — one wins, others get a 401. Validated by `test_refresh_token_concurrent_redemption_only_one_succeeds` (10 parallel tasks, asserts exactly 1 success).
- On replay-detected, calls `invalidate_user_tokens(user_id)` — process-local lockdown, sufficient for refresh path (DB-shared) but leaves a ≤30 min window on access tokens in multi-replica deployments. Documented for v1.2.0; not a regression (pre-PR exposure was 7 days vs the 30-min window now).

**Legacy-token compatibility (Senior Dev):**
- `Claims.jti: Option<Uuid>` with `#[serde(default, skip_serializing_if = "Option::is_none")]` — old tokens decoded by new code accepted, new tokens decoded by older code (rolling deploy) just see an unknown field. Forward/back compatible.
- Tokens minted before this PR (no jti in claims) get a deterministic `legacy_token_jti(user_id, iat)` computed via UUIDv5, so cross-replica replay rejection works without a forced re-login at deploy time.

**API/contract impact callout (Senior Dev):**
- `LoginResponse.refresh_token` shape unchanged, but **semantics change** — the previous refresh token was reusable; now it is single-use after this lands. Cookie-based flows (web UI) rotate automatically. Worth a CHANGELOG callout for out-of-tree SDK consumers (third-party scripts, custom mobile clients).

**Test coverage shape (Test):**
- 412-line integration test file at `backend/tests/refresh_token_rotation_tests.rs`, all DB-backed and `#[ignore]`-gated.
- Concurrent-redemption test directly exercises the load-bearing INSERT-ON-CONFLICT serialization claim.
- One acceptable gap: no explicit "expired refresh token reaches refresh_tokens" test, but `decode_token`'s exp check is independently covered by `test_expired_token_rejected`.

**Merge resolution verification (Senior Dev + DevOps):**
- Migration renumbered 073 → 074 to avoid colliding with #955's `073_download_tickets_cascade.sql` already on release/1.1.x. Both touch unrelated tables (`used_refresh_jtis` vs `download_tickets`); ordering 073 → 074 is safe on a fresh deployment from v1.1.8.
- `auth_service.rs` mod tests: both halves preserved cleanly, closing brace on `test_refresh_tokens_user_not_found_after_decode` re-inserted at L3259.
- `scheduler_service.rs`: refresh-jti GC (1h cadence, 180s startup delay) and download-ticket cleanup (10m cadence, 60s startup delay) both spawn; final `tracing::info!` summary updated to mention both.

### Recommended follow-ups (not blocking)

- Add CHANGELOG note about refresh-token single-use semantics for SDK consumers.
- Consider also blocklisting outstanding refresh jtis on user deactivation / password change (closes the multi-replica access-token gap without waiting for v1.2.0 NOTIFY/LISTEN).

### Status

Required CI checks: Check Rust pass, Backend Unit Tests pass, Security Audit pass, GitGuardian pass. Code Coverage in_progress. Build/integration/Windows/E2E correctly skipped per `release/1.1.x` gating.

`mergeStateStatus: UNSTABLE` is purely the in-progress Coverage job.

Ready to merge as soon as Coverage clears.

PR #954, opened 2026-04-27
fix(auth): invalidate API-token cache on user deactivation (#931)

### LINKED ISSUE #931: Flush API-token cache on user is_active=false
## Background

`auth_service.rs:90` defines `API_TOKEN_CACHE_TTL_SECS = 300` (5 minutes). When an admin sets `is_active=false` on a user via `PATCH /api/v1/users/{id}`, that user's currently-cached API tokens continue to validate for up to 5 minutes before the cache entry expires and the underlying SQL filter (`WHERE id = $1 AND is_active = true`, auth_service.rs:472) catches up.

Practical impact: if you deactivate a compromised user account, their tokens keep working for up to 5 minutes. This is documented behaviour but it is a meaningful security gap.

## Desired behavior

When `users.is_active` flips to false (PATCH /users/{id} with `{"is_active":false}`), the user's API token cache entries should be invalidated immediately so the next request hits the database and gets rejected.

## Implementation sketch

- Expose a `flush_user_tokens(user_id)` method on the cache
- Call it from the user-update handler when `is_active` changes to false
- Optionally also flush on password change, role change, or `delete user`

## Test coverage

`artifact-keeper-test/tests/auth/test-user-deactivation-revokes-tokens.sh` has a `Deactivated user's API token is rejected` case gated on `require_feature "user_deactivation_token_flush"`. It polls for ~30s after deactivation; auto-skips on 1.1.x and auto-fails on 1.2.0 if the flush does not actually evict the cache.

## Backend reference

- backend/src/services/auth_service.rs:90 (API_TOKEN_CACHE_TTL_SECS)
- backend/src/services/auth_service.rs:467-474 (validate_api_token SQL with is_active filter)
- backend/src/api/handlers/users.rs:108-114, 376-409 (PATCH user handler)

Tracked under Epic 11. Target: v1.1.9 (security backport candidate; bumped from v1.2.0).

brandonrc:
## Summary

Closes #931.

When an admin sets `is_active=false` on a user via `PATCH /api/v1/users/{id}`, that user's already-cached API-token validations would keep returning success for up to 5 minutes (`API_TOKEN_CACHE_TTL_SECS`). The cache-hit path in `validate_api_token` never consulted the `is_active` column, and the SQL filter `WHERE id = $1 AND is_active = true` on the bcrypt path could not see those cache hits, so a compromised or revoked account stayed authenticated through the entire TTL window.

This change closes the gap by:

- Adding a global `API_TOKEN_USER_INVALIDATIONS` map (mirroring the existing `CREDENTIAL_INVALIDATIONS` infrastructure used for JWT revocation). `invalidate_user_token_cache_entries(user_id)` records an invalidation timestamp.
- Extending the cache-hit branch of `AuthService::validate_api_token` so it rejects any entry whose user has been invalidated since the entry was cached, with a clear `User account is deactivated` error.
- Adding `AuthService::flush_user_token_cache_entries(user_id)` so the shared instance can also drop matching per-instance cache entries immediately, freeing memory rather than waiting for the TTL to expire.
- Calling both `invalidate_user_token_cache_entries` and the existing `invalidate_user_tokens` (JWT) from `PATCH /users/:id` whenever the request asks for `is_active=false`, and unconditionally from `DELETE /users/:id` (hard delete).

Out of scope, will be follow-ups: `AuthService::deactivate_missing_users` (federated sync) and the `service_accounts` handler also flip `is_active=false` and have the same gap. Touching them needs new SQL (`RETURNING id` and a `.sqlx` cache regen against a live DB) which is unsafe to bundle into a release-branch fix.

## Test Checklist
- [x] Manually tested locally

Five new unit tests in `auth_service::tests` cover the global invalidation map (mark/check, isolation between users, post-invalidation cache entry, unknown user, per-instance flush helper).

Three new integration tests in `backend/tests/api_token_cache_flush_tests.rs` (require Postgres, marked `#[ignore]`):

- `issued_token_validates_then_rejects_after_deactivation`: warms the cache, sets `is_active=false`, then verifies the next `validate_api_token` returns an `Authentication` error within seconds (well inside the 5-minute TTL).
- `deactivation_does_not_affect_other_users_tokens`: verifies the invalidation is scoped per-user.
- `flush_user_token_cache_entries_drops_in_memory_entries`: verifies the per-instance flush helper actually drops entries and that re-validation against an active user re-populates the cache.

Local verification:

- `cargo fmt --check` clean
- `cargo test --workspace --lib`: 7531 passed, 0 failed
- `DATABASE_URL=postgresql://registry:registry@localhost:30432/artifact_registry cargo test --test api_token_cache_flush_tests -- --ignored`: 3 passed, 0 failed

## API Changes
- [x] N/A - no API changes


brandonrc (commented):
## Independent Security Review (read-only)

Focus: auth-correctness and bypass paths around the new `API_TOKEN_USER_INVALIDATIONS` map and its call sites in `users.rs`. Findings ranked CRITICAL/HIGH/MEDIUM/LOW/NIT.

### HIGH-1: `service_accounts` PATCH/DELETE share the same gap and are exploitable today

`backend/src/api/handlers/service_accounts.rs:355-411` (`update_service_account`, `delete_service_account`) both flip `is_active` (or hard-delete) without calling `invalidate_user_token_cache_entries` or `invalidate_user_tokens`. Service accounts are precisely the principal type that runs hot in the API-token cache (cargo/npm/maven CI bots), so the 5-minute window is most relevant for them. The PR body acknowledges this as a follow-up gated on `.sqlx` regen, but no `.sqlx` change is needed for the invalidation calls themselves: `update_service_account` already returns the user (just call the two functions on the returned id), and `delete_service_account` has the id in scope. Recommend adding the two calls to this PR rather than deferring. Without them, the same #931 bug is trivially reachable through the service-account admin API.

### HIGH-2: `AuthService::deactivate_missing_users` flips `is_active=false` in bulk with no invalidation

`backend/src/services/auth_service.rs:1071-1097` runs `UPDATE users SET is_active=false ... WHERE external_id != ALL($2)` and returns only `rows_affected`. The deactivated user IDs are not even fetched, let alone passed to `invalidate_user_token_cache_entries`. Federated SSO sync (LDAP/SAML/OIDC) is exactly the reaper used after offboarding, so this is the highest-impact gap of the three: if an attacker has compromised a credential and the sync deactivates the upstream account, the API token still authorizes against the cache for up to 5 min. Fix shape: change the query to `RETURNING id`, then iterate calling both `invalidate_user_token_cache_entries(id)` and `invalidate_user_tokens(id)` for each row. No `.sqlx` regen blocker (`query!` works).

### MEDIUM-1: Multi-replica deployment reopens the bug

`API_TOKEN_USER_INVALIDATIONS` is `static OnceLock<RwLock<HashMap<...>>>`, so the map is per-process. Production deployments behind a load balancer (the IaC repo ships a Helm chart with replicas) have N independent maps. PATCH lands on replica A, replica A invalidates, replicas B..N keep serving the cached validation for up to 300 s. The PR body and code comments do not call this out at all; the doc-comment on `invalidate_user_token_cache_entries` reads as if the fix is global. Two acceptable resolutions: (a) document the constraint clearly in the function rustdoc and the PR description (acknowledged single-replica-only), or (b) wire the invalidation through a DB-backed signal table or NOTIFY/LISTEN / Redis pub-sub. A pragmatic short-term mitigation is to also write the invalidation timestamp to a DB column on the user row (`token_invalidated_at`) and have the cache-hit path consult it, which closes the multi-replica window at the cost of one extra column-only SELECT — but that is a follow-up. At minimum: update the doc-comment and CHANGELOG to state that this fix is replica-local.

### MEDIUM-2: Map growth is bounded only by churn rate, not by an explicit cap

`invalidate_user_token_cache_entries` prunes entries older than `2 * API_TOKEN_CACHE_TTL_SECS` (10 min) on every insert. Pruning runs only on insert, so if deactivations stop, stale entries persist until the process restarts (small constant leak, fine). Under churn, the upper bound is "number of distinct deactivations within any 10-minute window." That is sufficient in practice. Two NIT-grade improvements: (i) the `retain` walks the whole map on every call — for very large customer bases this is O(N) per insert, consider a `BTreeMap<Instant, Uuid>` companion or a periodic prune; (ii) add an `assert!`-style sanity check or metric for map size so an unexpected growth pattern is observable. Not blocking.

### LOW-1: Re-activation timestamp ordering is correct, but worth a regression test

Verified: `update_user` does not invalidate on `is_active=true`. The cache-hit check uses `cached_at <= invalidated_at`, so an entry inserted strictly after the prior deactivation timestamp passes. On `false→true→false`, the second deactivation overwrites the timestamp via `map.insert`, so any entries cached during the brief active window are correctly invalidated. The unit test `test_cache_entry_inserted_after_invalidation_is_kept` covers the basic positive case but does not exercise the false→true→false sequence. Recommend adding a unit test for that.

### LOW-2: TOCTOU window between SQL UPDATE and `invalidate_user_token_cache_entries`

In `update_user`, the UPDATE commits first (line 384-411), then the in-memory invalidation runs (line 423-426). For the ~µs between commit and the `map.insert`, a concurrent request that gets a cache hit will still pass — this is a vanishingly small window and the request would have been admitted anyway under any cache-coherency model (the request started "before" the deactivation as far as the auth layer can tell). Consistent with finding 4 in the review brief. No action required, but if you want belt-and-braces, call `invalidate_user_token_cache_entries(id)` BEFORE the UPDATE: pre-marking is safe because it only forces a re-validation, and any cache entry inserted between the pre-mark and the UPDATE will still be rejected if it predates the mark, while DB-fresh validations after the UPDATE will see `is_active=false` and fail anyway.

### LOW-3: In-flight request semantics

A request that is already past `validate_api_token` when the admin issues PATCH will complete with the deactivated user's privileges. That is correct behavior for any TTL-cache scheme and is not a regression. No action — but worth flagging in release notes alongside the multi-replica caveat so operators understand the contract: "invalidation prevents new cached validations within 5 minutes; in-flight requests authorized prior to the PATCH are not interrupted."

### LOW-4: `invalidate_user_token_cache_entries` does NOT call `flush_user_token_cache_entries`

The global map rejects the cached entry on hit, but the entry remains in the long-lived `AuthService.token_cache` until natural eviction. `flush_user_token_cache_entries` exists but is never called from `update_user` or `delete_user`. Functionally equivalent (both result in rejection), but the unflushed entry still occupies memory and the user's hashed-token cache key sits in process memory unnecessarily. Recommend calling it from the same sites that call the global invalidation, where the long-lived `AuthService` Arc is reachable. Not a security bug, a hygiene improvement.

### Test-coverage observations

- `issued_token_validates_then_rejects_after_deactivation` correctly exercises the warm-cache path. Good.
- `deactivation_does_not_affect_other_users_tokens` covers the user-isolation invariant. Good.
- `flush_user_token_cache_entries_drops_in_memory_entries` covers flush. Good.
- DELETE-user path is NOT covered by an integration test. The unit tests exercise `invalidate_user_token_cache_entries` directly, but no test simulates the DELETE handler's combined call. Add an `issued_token_rejected_after_user_delete` test.
- Re-activation false→true→false sequence is not tested. Add it (see LOW-1).
- Service-account PATCH/DELETE has no test (because they have no fix yet). Add when fixing HIGH-1.
- `deactivate_missing_users` has no test for cache invalidation (because it has no fix). Add when fixing HIGH-2.

### Verdict: NEEDS-MINOR-CHANGES

Core fix is correct and well-scoped for `users.rs`. Two acknowledged gaps (HIGH-1 service_accounts, HIGH-2 deactivate_missing_users) are exploitable in v1.1.x today and should not be deferred to a separate v1.1.10 ticket — both are mechanical to add in this PR, and the `.sqlx` blocker the author mentioned only applies to a subset of the change (the `RETURNING id` in deactivate_missing_users). The multi-replica caveat (MEDIUM-1) needs to be documented even if not fixed in this PR. The remaining items are LOW/NIT.

Recommended changes before merge:
1. Add `invalidate_user_token_cache_entries(id)` + `invalidate_user_tokens(id)` to `update_service_account` and `delete_service_account`.
2. Change `deactivate_missing_users` to `RETURNING id` and invalidate per row.
3. Update the rustdoc on `invalidate_user_token_cache_entries` and the CHANGELOG to call out the per-replica scope.
4. Add the DELETE-user integration test and re-activation regression test.

If the author wants to keep this PR narrow, items 1 and 2 should still land here (they are the same bug class and same release boundary); items 3 and 4 can be a follow-up.

brandonrc:
Addressed all findings from the security review. New head: bdb9caf.

## HIGH

**HIGH-1 (`service_accounts` PATCH/DELETE):** both handlers now pre-mark `invalidate_user_token_cache_entries` and `invalidate_user_tokens` before the underlying SQL runs. CI bots are the hottest path through the cache, so this closes the most-exploitable variant.

**HIGH-2 (`AuthService::deactivate_missing_users`):** changed the bulk `UPDATE` to use `RETURNING id` and now iterates the returned ids invalidating per row. Federated SSO offboarding (LDAP/SAML/OIDC) is now covered. The `.sqlx` cache was regenerated against a live Postgres instance and is committed in this push.

## MEDIUM

**MEDIUM-3 (multi-replica documentation):** rustdoc on `API_TOKEN_USER_INVALIDATIONS` and `invalidate_user_token_cache_entries` now spells out the per-process scope and points at the v1.2.0 follow-up. CHANGELOG entry under `[Unreleased] / Security` documents the same caveat.

**MEDIUM-4 (periodic prune):** added an hourly task in `scheduler_service.rs` that calls a new `prune_stale_user_token_invalidations()` helper. Memory stays bounded even when deactivations are infrequent.

## LOW

**LOW-5 (TOCTOU pre-mark):** invalidation now runs BEFORE the SQL UPDATE/DELETE in `users.rs::update_user`, `users.rs::delete_user`, `service_accounts.rs::update_service_account`, and `service_accounts.rs::delete_service_account`. Pre-marking is fail-secure (worst case: one extra DB re-validation if the SQL fails).

**LOW-6 (`flush_user_token_cache_entries` never called):** `AuthService::token_cache` is now `Arc<TokenCacheMap>`. A new global registry holds `Weak<TokenCacheMap>` references; `AuthService::register_for_global_flush(&self)` registers the long-lived instances. `invalidate_user_token_cache_entries` walks the registry and flushes matching entries (also pruning dead Weaks). `routes.rs` registers the two long-lived `AuthService` instances (visibility middleware + API v1 middleware); ad-hoc per-request instances deliberately do not register.

## Tests

Four new unit tests in `auth_service::tests`:
- `test_reactivation_then_redeactivation_invalidates_again` (LOW-1 regression)
- `test_register_for_global_flush_drops_matching_cache_entries`
- `test_dropped_cache_weak_is_pruned_from_registry`
- `test_prune_stale_user_token_invalidations_handles_empty_map`

Five new integration tests in `backend/tests/api_token_cache_flush_tests.rs` (Postgres, `#[ignore]`):
- `issued_token_rejected_after_user_delete`
- `reactivation_then_redeactivation_rejects_old_cache_entries`
- `service_account_patch_deactivation_rejects_cached_tokens`
- `service_account_delete_rejects_cached_tokens`
- `deactivate_missing_users_invalidates_each_deactivated_user`

## Local verification

- `cargo fmt --check`: clean
- `cargo test --workspace --lib`: 7535 passed, 0 failed
- `DATABASE_URL=... cargo test --test api_token_cache_flush_tests -- --ignored`: 8 passed, 0 failed
- `cargo sqlx prepare --workspace --check -- --tests`: in sync

Issue #508, opened 2026-03-21
Unauthenticated format endpoint access returns 404 instead of 401


brandonrc:
## Summary

Format-specific upload endpoints (PyPI, NPM, generic) return 404 for unauthenticated requests instead of 401. The red team auth bypass tests expect 401/403 but get 404, which means the backend routes the request to the format handler (which looks up the repo and fails with "not found") instead of rejecting it at the auth middleware layer.

## Evidence

Release gate run: https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/23372998677

```
[FAIL] PyPI upload accepted without auth (status 404)
[FAIL] NPM upload accepted without auth (status 404)
[FAIL] Generic upload accepted without auth (status 404)
[FAIL] Create repository without auth - returned 422 instead of 401/403
```

## Test Reference

`artifact-keeper-test/tests/security/redteam/test-04-auth-bypass.sh`

## Impact

Medium. While the operation fails, returning 404 instead of 401 means:
1. Auth middleware may not be applied to all format routes
2. Error response leaks whether a repo exists (404 vs different error)

## Fix

Verify that auth middleware is applied to all format upload routes. The auth layer should reject unauthenticated requests with 401 before the request reaches the format handler.

PR #955, opened 2026-04-27
feat(auth): download-ticket consumer middleware (#930)

### LINKED ISSUE #930: Add public consumer endpoint for download tickets (?ticket= query param auth)
## Background

`POST /api/v1/auth/ticket` mints a short-lived (30s), single-use download ticket. The intent is to use these tickets on endpoints that cannot send `Authorization` headers (HTML `<a>` downloads, EventSource/SSE, etc).

The service-layer plumbing exists: `auth_config_service::validate_download_ticket` (auth_config_service.rs:1367) uses `DELETE ... RETURNING` to atomically consume a ticket by value, with `expires_at > NOW()` guarding expiry.

What's missing: a public route that accepts `?ticket=<value>` as auth and consumes it. Without it, the ticket is a write-only feature; clients can mint tickets but nothing accepts them.

## Desired behavior

- A handler/middleware that recognizes `?ticket=<value>` on download routes (artifacts, packages, SBOM exports, etc), looks up `validate_download_ticket`, and on success treats the request as authenticated for the resource path the ticket was bound to.
- On second use of the same ticket: 401 (DELETE-RETURNING already gives single-use).
- On expired ticket: 401 (the `expires_at > NOW()` filter already gives this).

## Test coverage

`artifact-keeper-test/tests/auth/test-download-ticket-lifecycle.sh` has two cases gated on `require_feature "download_ticket_consumer"`:
- Single-use ticket consumption (second use rejected)
- Expired ticket rejection

Both auto-skip on 1.1.x and auto-fail on 1.2.0 if the consumer ships but doesn't enforce single-use/expiry.

## Backend reference

- backend/src/services/auth_config_service.rs:1367 (validate_download_ticket — DELETE...RETURNING)
- backend/src/api/handlers/auth.rs:504-538 (create-ticket handler)

Tracked under Epic 11. Target: v1.1.9 (security backport candidate; bumped from v1.2.0).

brandonrc:
## Summary

`POST /api/v1/auth/ticket` (handlers/auth.rs:504) has minted short-lived single-use download tickets bound to a `resource_path` since 1.1.0. The service-layer plumbing was already in place: `auth_config_service::validate_download_ticket` (auth_config_service.rs:1367) atomically consumes a ticket via `DELETE ... RETURNING` and rejects expired ones via `expires_at > NOW()`. What was missing was any public route that accepted `?ticket=<v>` as authentication, so the mint endpoint was producing tokens that nothing could consume.

This PR wires the consumer side. Three middleware functions now fall back to download-ticket validation when no header credentials are present (or all of them have failed):

- `auth_middleware` (the strict variant used by protected routes)
- `optional_auth_middleware` (used by repository, artifact, build, and search reads)
- `repo_visibility_middleware` (used by all native-protocol format handlers)

`admin_middleware` deliberately does NOT accept tickets: tickets must never grant admin elevation.

### Where ticket auth applies

A ticket authenticates only the request whose URL path equals the ticket's bound `resource_path` (exact-match, not prefix). A ticket minted with `resource_path = null` authenticates any read path the minting user can reach. Tickets restrict to `GET` and `HEAD` only; `POST`/`PUT`/`PATCH`/`DELETE` with `?ticket=` are rejected even when the bound path is a write-capable endpoint.

### Where in the middleware chain

Ticket recognition runs as a fallback after Bearer / Basic / API-key all fail. If the request carries a valid `Authorization` header, the ticket query parameter is ignored. This keeps existing behaviour byte-identical for every request that already works.

### Identity propagation

After `validate_download_ticket` returns the owning `user_id`, the user record is loaded with `sqlx::query_as::<_, User>(...)` (runtime form so no offline cache regeneration is required) and `AuthExtension::from(user)` is inserted into `request.extensions`, matching the contract every other auth method already follows. Inactive users are blocked at this step so revocation takes effect immediately.

### Scope enforcement

A ticket is read-only. The middleware drops `is_admin` from the resulting `AuthExtension` and attaches a new `DownloadTicketAuth` marker extension. `repo_visibility_middleware` checks for that marker when deciding whether a write may proceed and treats ticket-authenticated requests as anonymous for the `is_write` gate (so #508's write-requires-auth invariant still holds even on public repos). `DownloadTicketAuth` is a separate marker rather than a field on `AuthExtension` to avoid touching the 80+ existing `AuthExtension { ... }` literals across the codebase.

### Single-use

Single-use is enforced atomically by the existing `DELETE ... RETURNING` query in `validate_download_ticket`. A ticket consumed by a path-mismatched attempt cannot be replayed against the bound path either.

## Test Checklist
- [ ] E2E tests added/updated (if applicable) -- the existing `artifact-keeper-test/tests/auth/test-download-ticket-lifecycle.sh` skipped cases will start running once this PR lands.
- [x] Manually tested locally

14 new unit tests cover query-string parsing (including percent-decoding, repeated keys, substring-match rejection), the read-only method whitelist, and the exact-path policy. 6 ignored integration tests in `backend/tests/download_ticket_tests.rs` exercise the full lifecycle against a running server: bound GET succeeds, second use 401s, path mismatch 401s and consumes the ticket, POST 401s, missing/garbage ticket 401s.

Local run: \`cargo test --workspace --lib\` -> 7540 passed, 0 failed (was 7526 before this PR).

## API Changes
- [ ] New endpoints have \`#[utoipa::path]\` annotations
- [ ] Request/response types have \`#[derive(ToSchema)]\`
- [ ] OpenAPI spec validates: \`cargo test --lib test_openapi_spec_is_valid\`
- [x] N/A - no API changes

No new endpoints. The existing \`POST /api/v1/auth/ticket\` and \`GET /api/v1/auth/me\` are unchanged. The change is entirely inside the auth middleware layer.

Closes #930
Milestone: v1.1.9

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Auth: download-ticket consumer middleware** (#930) -- the `?ticket=<v>` query parameter minted by `POST /api/v1/auth/ticket` is now accepted as a fallback authenticator on read routes (`auth_middleware`, `optional_auth_middleware`, and `repo_visibility_middleware`). Tickets are single-use, expire after 30 seconds, are restricted to GET/HEAD methods, and only authenticate the request whose URL path matches the ticket's bound `resource_path`. Useful for browser anchor-tag downloads and `EventSource` SSE streams where `Authorization` headers cannot be set.
+
 ### Security
 
 - **API-token cache invalidation on user deactivation** (#931) -- when an admin deactivates or deletes a user (`PATCH /api/v1/users/{id}`, `DELETE /api/v1/users/{id}`), updates a service account (`PATCH/DELETE /api/v1/service-accounts/{id}`), or runs a federated SSO offboarding sync (`AuthService::deactivate_missing_users`), every cached API-token validation belonging to that user or service account is now rejected immediately rather than continuing to authenticate for up to 5 minutes (the previous `API_TOKEN_CACHE_TTL_SECS` window). Caveat: the invalidation map is per-process. In multi-replica deployments (Helm `replicas > 1`) only the replica that handled the admin action evicts immediately; other replicas still reject the cached entry within the same 5-minute window via the existing `WHERE is_active = true` SQL filter, but cache hits on those replicas can still authenticate during that window. A v1.2.0 follow-up will move the signal into the database or a Redis pub-sub channel so it is observed by every replica.
--- a/backend/migrations/073_download_tickets_cascade.sql
+++ b/backend/migrations/073_download_tickets_cascade.sql
@@ -1,0 +1,33 @@
+-- Add ON DELETE CASCADE to download_tickets.user_id.
+--
+-- Migration 039 created the table with a plain `REFERENCES users(id)`, which
+-- defaults to `ON DELETE NO ACTION`. That meant deleting a user with
+-- outstanding (typically about-to-expire) tickets would fail with a foreign-key
+-- violation. Tickets are short-lived (30s TTL) and single-use, so cascading on
+-- delete is correct: the tickets become invalid the moment the user does.
+--
+-- We drop the old constraint by inspecting the system catalog rather than
+-- naming it explicitly, because Postgres auto-generates the constraint name
+-- (`download_tickets_user_id_fkey` is conventional but not guaranteed).
+DO $$
+DECLARE
+    fk_name text;
+BEGIN
+    SELECT con.conname INTO fk_name
+    FROM pg_constraint con
+    JOIN pg_class rel ON rel.oid = con.conrelid
+    JOIN pg_attribute att ON att.attrelid = con.conrelid AND att.attnum = ANY(con.conkey)
+    WHERE rel.relname = 'download_tickets'
+      AND att.attname = 'user_id'
+      AND con.contype = 'f';
+
+    IF fk_name IS NOT NULL THEN
+        EXECUTE format('ALTER TABLE download_tickets DROP CONSTRAINT %I', fk_name);
+    END IF;
+END $$;
+
+ALTER TABLE download_tickets
+    ADD CONSTRAINT download_tickets_user_id_fkey
+    FOREIGN KEY (user_id)
+    REFERENCES users(id)
+    ON DELETE CASCADE;
--- a/backend/src/api/handlers/auth.rs
+++ b/backend/src/api/handlers/auth.rs
@@ -486,9 +486,135 @@ pub struct TicketResponse {
     pub expires_in: u64,
 }
 
+/// Validate and normalize a ticket-bound `resource_path` at mint time.
+///
+/// The consumer middleware compares `bound_path == request.uri().path()` by
+/// byte equality, which means the minter is responsible for picking the exact
+/// form the consumer will see. Format handlers normalize incoming paths
+/// (PyPI/NuGet/Go lowercase the package-name segment, see
+/// `backend/src/formats/pypi.rs` and `backend/src/formats/nuget.rs`), so a
+/// minter who passes `/pypi/foo/simple/Django/` would produce a ticket that
+/// no real client request can match — and the first attempt would silently
+/// burn the ticket.
+///
+/// Policy enforced here:
+///   1. Path must be absolute (starts with `/`).
+///   2. No `..` segments (no path traversal).
+///   3. No percent-encoded slashes or backslashes (`%2F`, `%2f`, `%5C`, `%5c`)
+///      and no `%25` (double-encoding) — these would change semantics after
+///      URL-decode and are common bypass vectors.
+///   4. No control characters (`\0`..`\x1f`, `\x7f`) or whitespace.
+///   5. Collapse repeated `/` to a single `/`.
+///   6. Strip a trailing `/` (except for root `/`).
+///   7. Lowercase the package-name segment for paths whose first component is
+///      a known case-folding format (`/pypi/...`, `/nuget/...`, `/go/...`).
+///      The package name lives at the third segment for these formats
+///      (`/{format}/{repo_key}/{name}/...`).
+///
+/// What is NOT enforced here (deliberate):
+///   - Authz reach: this function does not verify that the minting user can
+///     actually access the requested path. The consumer middleware re-runs
+///     `repo_visibility_middleware` / `can_access_repo` at consume time, so a
+///     ticket bound to a path the minter cannot reach is harmless. A defense-
+///     in-depth check is tracked for v1.2.0 hardening.
+fn validate_and_normalize_resource_path(input: &str) -> std::result::Result<String, AppError> {
+    if input.is_empty() {
+        return Err(AppError::Validation(
+            "resource_path must not be empty".into(),
+        ));
+    }
+    if !input.starts_with('/') {
+        return Err(AppError::Validation(
+            "resource_path must start with '/'".into(),
+        ));
+    }
+
+    // Reject control chars, whitespace, embedded NUL.
+    for b in input.as_bytes() {
+        if *b < 0x20 || *b == 0x7f || *b == b' ' {
+            return Err(AppError::Validation(
+                "resource_path must not contain whitespace or control characters".into(),
+            ));
+        }
+    }
+
+    // Reject percent-encoded slashes / backslashes / percent itself. These
+    // forms decode to characters that change path semantics after axum/hyper
+    // serve them as raw paths, so allowing them would let a minter bind to
+    // `/foo%2F..%2Fbar` and rely on a future decoder normalizing it.
+    let lower = input.to_ascii_lowercase();
+    for needle in ["%2f", "%5c", "%25", "%00"] {
+        if lower.contains(needle) {
+            return Err(AppError::Validation(format!(
+                "resource_path must not contain encoded sequence '{}'",
+                needle
+            )));
+        }
+    }
+
+    // Split, reject `..` and `.` segments, collapse repeated `/`.
+    let mut segments: Vec<&str> = Vec::new();
+    for seg in input.split('/') {
+        if seg.is_empty() {
+            // collapses `//` into single `/` and skips leading/trailing empty segs
+            continue;
+        }
+        if seg == ".." {
+            return Err(AppError::Validation(
+                "resource_path must not contain '..' segments".into(),
+            ));
+        }
+        if seg == "." {
+            return Err(AppError::Validation(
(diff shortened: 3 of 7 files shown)


brandonrc (commented):
## Independent Security Review (read-only)

Reviewed the ticket-consumer fallback in `auth_middleware`, `optional_auth_middleware`, and `repo_visibility_middleware`, plus `try_resolve_ticket_auth`, `validate_download_ticket`, and the integration tests. Findings ranked by severity.

### HIGH

**H1. Mint endpoint accepts arbitrary `resource_path` with no normalization or authz.** `backend/src/api/handlers/auth.rs:504-521` calls `AuthConfigService::create_download_ticket(.., payload.resource_path.as_deref())` and the value is stored verbatim in `download_tickets.resource_path`. There is no canonicalization (no rejection of `..`, double slashes, mixed case, percent-encoded forms, trailing-slash variants), no leading-slash check, and no verification that the minting user actually has access to the path they are binding to. Since the consumer compares with `bound_path == request_path` on the raw `request.uri().path()`, the minter is responsible for picking the exact-byte form the consumer will use. Two practical consequences:

- A client that mints with `/pypi/foo/simple/Django/` and then issues `/pypi/foo/simple/django/` (PyPI lower-casing, see `backend/src/formats/pypi.rs:190`) will get 401 with the ticket consumed. Same for trailing-slash drift across format handlers.
- The path policy is enforced by string equality only. There is no enforcement that the minted path matches the user's actual entitlements. Authz is re-checked at consume time via `require_visible` / `can_access_repo`, so this isn't an elevation bug today, but it makes it easy for a future handler that *omits* its own authz check to be silently authorized by ticket. Worth a hard server-side validator that requires `resource_path` to start with `/api/v1/` or a known format prefix, normalized, and ideally checks current user reach. Severity is HIGH because consumer enabling turns a stored-as-anything column into a live capability binding.

### MEDIUM

**M1. Ticket is consumed before path check (DoS-by-mistyped-path).** `backend/src/api/middleware/auth.rs:542-554` calls `validate_download_ticket` (which atomically `DELETE ... RETURNING`) *before* `ticket_path_allowed`. Test 3 (`download_ticket_tests.rs:189-233`) explicitly documents this. The behavior is acceptable for a stolen-ticket adversary (they can replay against the right path anyway), but a buggy or panicked legitimate client that hits the wrong path once gets locked out and must mint again. Worth either (a) doing the path check via `SELECT` then `DELETE WHERE ... RETURNING` in a transaction so wrong-path attempts don't burn the ticket, or (b) explicitly documenting the trade-off in `create_download_ticket`'s comment.

**M2. Path-encoding/normalization corner cases are not tested.** `ticket_path_allowed` (line 518-523) is byte-exact on `request.uri().path()`, which in axum/hyper is the raw, *not percent-decoded* path. That is fine in principle but the test suite does not exercise:
- `%2F` in either bound or request path
- trailing-slash equivalence (`/foo` vs `/foo/`)
- case-folding in format handlers that lower-case names internally (PyPI, Go, NuGet — see `backend/src/formats/pypi.rs:190`, `backend/src/formats/nuget.rs:170`)
- double-encoded sequences (`%252F`)

The exact-match policy is the right default; the gap is test coverage. Add table-driven cases to the existing `test_ticket_path_allowed_*` set covering each of the above (especially that mismatch is rejected, not silently treated as match).

**M3. `extract_ticket_from_query` percent-decode produces UTF-8 from arbitrary bytes.** Line 466-492 takes percent-decoded byte values above 0x7F and pushes them via `out.push(b as char)`, which encodes as multi-byte UTF-8. Tickets are 64-char hex so this never fires today, but the extractor is `pub(crate)` and may be reused. Either restrict the extractor to ASCII (reject input with `%XX` where `XX > 0x7F`) or document that the function is hex-tickets-only.

**M4. `is_active=true` filter is correct, but no filter for `must_change_password`.** Lines 565-580 select the user and reject inactive accounts, but a user in the must-change-password state still authenticates via ticket. If the policy elsewhere is "must change password before any operation," tickets bypass that gate. Verify by inspection: is there a request-time `must_change_password` block? If so, mirror it for ticket auth or reject in the load query.

**M5. No periodic cleanup invocation for `cleanup_expired_download_tickets`.** `backend/src/services/auth_config_service.rs:1386` defines the cleaner but no scheduler calls it. Validation correctly checks `expires_at > NOW()` so this is hygiene only, not correctness, but the table grows monotonically until manual intervention.

### LOW

**L1. `download_tickets.user_id` lacks `ON DELETE CASCADE`.** `backend/migrations/039_download_tickets.sql:3`. Deleting a user fails (FK violation) if outstanding tickets exist; either CASCADE or RESTRICT-with-explicit-cleanup. Pre-existing.

**L2. `must_change_password` and `is_service_account` are still readable through ticket auth.** `AuthExtension::from(user)` populates these. Downstream handlers that branch on `is_service_account` will see the value. Probably intended (ticket inherits the minter's identity), but worth a one-line comment in `try_resolve_ticket_auth` confirming the design.

**L3. Error-message confusion on POST with `?ticket=`.** `auth_middleware` (line 366-370) returns "Invalid or expired download ticket" when a POST is attempted with a ticket query param. The ticket may be valid; the issue is the method. Consider a separate message like "Tickets cannot authenticate write operations."

**L4. Ticket-in-URL exposure (well-known).** `?ticket=<64-hex>` ends up in access logs, browser history, Referer, proxy logs. Single-use + 30-second TTL + 256-bit entropy mitigates this, but flag in user-facing docs that tickets must not be shared and that web UI flows should consume immediately. Already implicit in the design but worth an explicit note.

### NIT

- `ticket_method_allowed` rejects OPTIONS, PROPFIND, etc. correctly. No WebDAV in repo (verified via grep). Good.
- `is_admin = false` overwrite (line 588) is good defense in depth.
- `auth_ext.is_some() && !authed_via_ticket` write-gate (line 918) is correct.
- `has_scope("admin")` on a non-API-token AuthExtension returns true (line 67-69 of `AuthExtension::has_scope`). For ticket auth, `is_api_token=false`, so `auth.has_scope(\"admin\")` would return true. I grepped: no handler currently uses `has_scope(\"admin\")` for elevation (all use `is_admin` or `require_admin`). This is fine TODAY but is a footgun for future handlers. Consider explicit `if authed_via_ticket { is_api_token = true; scopes = Some(vec![]) }` so scope checks default to deny.

### Test coverage gaps

The 6 integration tests in `backend/tests/download_ticket_tests.rs` cover the happy path, single-use, path mismatch, write rejection, missing ticket, and invalid ticket. Missing:
- Expired ticket (would require sleeping 30s or seeding directly).
- HEAD method success path (only GET tested).
- Encoded-path / case / trailing-slash variants per M2.
- Inactive user rejection (deactivate the minter between mint and use).
- Concurrent same-ticket use (two requests racing — exercise the atomic DELETE-RETURNING).

### Verdict: NEEDS-MINOR-CHANGES

The design is sound. Header-credentials-first ordering, exact-match path binding, GET/HEAD allowlist, separate `DownloadTicketAuth` marker for write-gating, and the `is_admin = false` neutralization are all the right calls. The blocking issues are H1 (mint accepts arbitrary `resource_path` with no normalization) and the M2 test gaps. H1 in particular should be addressed in this PR since enabling the consumer is precisely what makes the lack of mint-side validation reachable.

brandonrc:
Pushed `cb28bcc` addressing the review.

### HIGH-1: mint-time `resource_path` validation
New `validate_and_normalize_resource_path` helper in `handlers/auth.rs`. Policy:

1. Must be absolute (start with `/`).
2. Reject `..` and `.` segments.
3. Reject percent-encoded slashes/backslashes (`%2F`, `%2f`, `%5C`, `%5c`) and `%25` (double-encoding) and `%00` (null).
4. Reject control chars (`\0`..`\x1f`, `\x7f`) and whitespace.
5. Collapse repeated `/` to single `/`.
6. Strip trailing `/` (except root).
7. Lowercase the package-name segment (segment[2]) for `pypi`, `nuget`, `go` so the bound path matches the handler's post-normalization form. Repo key (segment[1]) and Maven/npm packages are left case-preserved.

What is NOT done here: server-side authz reach-check (verifying the minter can access the path before binding). The consumer re-runs `repo_visibility_middleware` and `can_access_repo`, so binding to an unreachable path is harmless. Tracked as v1.2.0 hardening.

### NIT-2: `has_scope` default-deny for ticket auth
`try_resolve_ticket_auth` now sets `ext.is_api_token = true` and `ext.scopes = Some(vec![])` on the resulting `AuthExtension`. No handler today uses `has_scope(\"admin\")` for elevation, but a future one would have silently granted admin to ticket holders without this. `is_service_account` and `must_change_password` remain inherited from the minter (deliberate; the ticket is proof that the session intentionally minted it).

### MEDIUM-3: edge-case tests for `ticket_path_allowed`
Added: trailing-slash mismatch (both directions), case-difference rejection, `%2F`/`%2f` rejection, `%252F` rejection, Unicode decomposed-vs-precomposed equivalence rejection.

### MEDIUM-4: scheduler GC
Wired `cleanup_expired_download_tickets` into `scheduler_service::spawn_all` at 10-minute cadence. Hygiene only; `validate_download_ticket` already enforces `expires_at > NOW()`.

### MEDIUM-5: inline comment expanded
The path-mismatch consume site at `auth.rs:549` now spells out the single-use trade-off (mistyped path burns the ticket; we accept the cost because single-use is the security invariant).

### LOW-6: migration 073 (CASCADE)
Added `073_download_tickets_cascade.sql` to swap `download_tickets.user_id` to `ON DELETE CASCADE`. Migration uses a `DO $$` block to look up the existing constraint name from `pg_constraint` rather than assuming `download_tickets_user_id_fkey` (Postgres auto-generates).

### LOW-7: ambiguous-error comment
Added a comment at the 401 build site explaining why \"Invalid or expired download ticket\" intentionally does not distinguish between not-found / expired / mismatched-path / wrong-method (so future developers don't try to \"fix\" it into a more specific message that would leak which case it is).

### LOW-8: URL-exposure note
Handler doc-comment on `create_download_ticket` now calls out that the ticket value appears in webserver access logs, browser history, and Referer headers, with single-use + 30s TTL + 256-bit entropy as the mitigation.

### Test counts
Lib tests: 7540 -> 7564 (24 new; clippy clean). Local: `cargo test --workspace --lib` 7564 passed; with `DATABASE_URL` set, full workspace suite passes too.

### Out of scope (deferred)
- M1 from review: SELECT-then-DELETE-WHERE in a transaction. Reviewer accepted the current trade-off; comment now makes it explicit.
- M3: tightening `extract_ticket_from_query` to ASCII-only. Function is already documented as hex-tickets-only and tickets are 64-char hex, so non-ASCII input would never decode to a valid ticket.
- M4: `must_change_password` gate for tickets. The reviewer noted this is conditional on whether such a gate exists elsewhere; not changing scope.
- v1.2.0 follow-up: defense-in-depth authz reach-check at mint time.

New head: `cb28bcc60d6b96ace5d9e694c00dc6c334fcbcc8`.

PR #992, opened 2026-04-30
feat(webhooks): backport delivery producer from #953 (closes #909) for v1.1.9

brandonrc:
## Summary

Backports the webhook delivery producer from PR #953 (`feat/webhooks-v2-combined`) onto `release/1.1.x` for v1.1.9. Closes #909.

The `webhook_deliveries` table on release/1.1.x is read by the retry scheduler and list endpoints but no code path inserts into it, so the retry scheduler has nothing to retry and webhook subscribers receive nothing. PR #953 fixes this on main but is too large to backport in full (E1 encrypted secrets, E3 producer, E4 richer payload, E6 deprecation headers). This PR extracts only E3, the producer.

What this PR does:

- New `backend/src/services/webhook_producer.rs` module that subscribes to the in-process `EventBus` and INSERTs a row into `webhook_deliveries` for every enabled webhook whose `events` array contains the mapped event type and whose `repository_id` is NULL or matches the event's repository scope.
- Wires the producer into `main.rs`, spawned after the shared `CancellationToken` is constructed. Holds clones of `db_pool` and `state.event_bus` captured before the sync worker consumes `db_pool` and the router takes ownership of `state`.
- Adds two additive helpers to `metrics_service.rs`: `record_webhook_delivery_enqueued` and `record_webhook_delivery_enqueue_failed`. Existing `record_webhook_delivery` is unchanged.
- Registers the new module in `services/mod.rs`.
- 19 unit tests for `map_event_type` and `build_event_payload`. No database required.

What is intentionally out of scope (deferred to v1.1.10 or later):

- HMAC-SHA256 signing of payloads (#910). The existing `X-AK-Signature` literal in the retry scheduler is unchanged.
- E1 encrypted webhook secret column (`secret_encrypted`) and `webhook_secret_crypto.rs`.
- E4 richer payload schema (`event_schema_version`, per-event enrichment).
- E6 deprecation headers and `notification_dispatcher` changes.
- The new email subscription migration on main.

Schema compatibility on release/1.1.x: `webhook_deliveries` already has every column the producer INSERTs into (`webhook_id, event, payload, attempts, next_retry_at, success`). `011_webhooks.sql` defines them and `067_webhook_delivery_retry.sql` adds `next_retry_at` and `max_attempts` with the matching retry index. No new migration is required.

[...]

Issue #948, opened 2026-04-27
Thread explicit repository_id through DomainEvent for non-repo events


brandonrc:
## Context

Both `webhook_producer.rs::enqueue_for_event` and `notification_dispatcher.rs::dispatch_event` parse `DomainEvent.entity_id` as a `repository_id` for the SQL match. That works for `repository.*` events because the EventBus puts the repo UUID in `entity_id`. It is a category error for `user.*`, `build.*`, and `artifact.*` events: their `entity_id` is the user/build/artifact UUID, not the owning repository.

For non-repo events the `repository_id = $2` arm of the WHERE clause never matches, so only global-scoped subscriptions and webhooks fire. Repo-scoped subscriptions to non-repo events silently never fire.

## Proposed fix

Add `repository_id: Option<Uuid>` (or similar) to `DomainEvent` and have publishers populate it explicitly. Then the SQL match becomes `repository_id IS NULL OR repository_id = $repo_id` where `$repo_id` is the explicit field, not a parse of `entity_id`.

Affected:

- `backend/src/services/event_bus.rs` (struct definition + every emit site)
- `backend/src/services/webhook_producer.rs::enqueue_for_event`
- `backend/src/services/notification_dispatcher.rs::dispatch_event`

The webhook producer file already carries a `FIXME(#TBD-REPO-SCOPE-FIXUP)` referencing this issue.

## Out of scope here

This was flagged during the v1.1.9 webhooks v2 review (PRs #937 / #939 / #940 combined). The combined PR keeps the existing entity_id parse and documents the limitation; the structural fix lands separately.

Issue #909, opened 2026-04-27
Webhook deliveries table is never populated (deliveries are dead-on-arrival)


brandonrc:
## Severity: CRITICAL

The `webhook_deliveries` table is read by the retry engine and the deliveries-list endpoints, but **no code path INSERTs rows into it on any event**. Webhook subscribers will never receive any delivery for any event other than what's manually triggered via `/redeliver` (which UPDATEs existing rows, not INSERTs new ones).

## Verification (independent grep, not agent-relayed)

```
$ grep -nE "webhook_deliveries" backend/src/api/handlers/webhooks.rs backend/src/services/*.rs

webhooks.rs:527: FROM webhook_deliveries
webhooks.rs:546: FROM webhook_deliveries
webhooks.rs:601: FROM webhook_deliveries
webhooks.rs:656: UPDATE webhook_deliveries
webhooks.rs:746: comment
webhooks.rs:770: FROM webhook_deliveries
webhooks.rs:822: UPDATE webhook_deliveries SET next_retry_at = NULL
webhooks.rs:836: UPDATE webhook_deliveries SET next_retry_at = NULL
webhooks.rs:893: UPDATE webhook_deliveries
webhooks.rs:913: UPDATE webhook_deliveries
webhooks.rs:937: UPDATE webhook_deliveries
metrics_service.rs:62: counter increment (no SQL)
```

Zero `INSERT INTO webhook_deliveries` anywhere. The retry engine, dead-letter logic, deliveries-list endpoint, and HMAC signing path all run on rows that nothing creates.

## Affected branches

- release/1.1.x: confirmed (grep above is from /tmp/ak-1.1.x-analysis tip 9eb374e)
- main: confirmed (same pattern)

## What's missing

A producer that, on each domain event the webhook subscribes to (`artifact.created`, `scan.completed`, etc.), creates a `webhook_deliveries` row with `status='pending'` and `next_retry_at=NOW()`. The existing retry engine then picks it up.

Likely insertion sites:
- `artifact_service.rs` after upload commit
- `scan_result_service.rs` after `complete_scan`
- `repository_service.rs` on create/update/delete
- `event_bus.rs` if all events go through there -- could be a single point of dispatch

## Test coverage

artifact-keeper-test#82 (Epic 7 PR) handles this gracefully: when no delivery rows appear after a triggering event, the retry/dead-letter tests `skip` with a clear reason instead of failing. Once this producer is implemented, those tests will exercise the engine for real.

[...]

PR #993, opened 2026-05-01
fix(security): require admin role on webhook write handlers (cross-tenant exposure)

brandonrc:
## Summary

Webhook write handlers (`create_webhook`, `delete_webhook`, `enable_webhook`, `disable_webhook`, `test_webhook`, `redeliver`) were gated on `auth_middleware`, which permits **any authenticated user**. This PR moves them under `admin_middleware`. Read endpoints (`list_webhooks`, `get_webhook`, `list_deliveries`) remain accessible to any authenticated user.

### Why this is a v1.1.9 blocker

This is a **cross-tenant information disclosure** with a clear exploitation path the moment the webhook producer (#992) ships:

- A low-privileged user calls `POST /api/v1/webhooks` with `repository_id: null` and `events: ["repository.created", "user.created"]`.
- The webhook record is created with no admin involvement.
- Once #909 is functional (which #992 makes it), every global event in the instance is delivered to that user's URL, including events on repositories they cannot read.
- The webhook also acts as a passive SSRF beacon: an attacker who controls the webhook URL learns about repo creations, user provisioning, and build activity across the whole tenant.

The flaw is **pre-existing** on `release/1.1.x` (the same routing was on `main` before webhooks were wired up to anything). It was harmless when webhooks were dead code. It becomes exploitable the moment #909 / #992 lands. v1.1.9 cannot ship without this fix.

### What changed

- `backend/src/api/handlers/webhooks.rs`
  - Split `router()` into two functions:
    - `router()` -- read-only endpoints: `GET /`, `GET /:id`, `GET /:id/deliveries`.
    - `admin_router()` -- mutating endpoints: `POST /`, `DELETE /:id`, `POST /:id/enable`, `POST /:id/disable`, `POST /:id/test`, `POST /:id/deliveries/:delivery_id/redeliver`.
  - Doc comments on both routers explain the threat model and reference #948 (per-tenant webhook scoping) as the planned long-term fix.
- `backend/src/api/routes.rs`
  - Mounts both sub-routers under `/webhooks`, each with its own auth layer, then merges before nesting. This avoids axum's panic-on-duplicate-nest behavior.
- `CHANGELOG.md`
  - Entry under `[Unreleased] / Security` describing the change and back-compat impact.

### Read endpoints: deliberately not gated to admin

For the v1.1.9 minimum-viable fix the choice is between:

1. Admin-only on **all** webhook endpoints (safest).
2. Admin-only on **writes**, any-authenticated-user on reads.

This PR takes option 2. Rationale: the read endpoints expose configuration the admin already authored and the database row is identifying (URL, custom headers may contain secrets-by-name, secret_hash is not returned). The actual cross-tenant leak is in the **delivery channel** (an attacker controls a URL that receives events), and that channel is closed once writes are admin-gated. Locking reads as well would also break the existing UI permission model where webhooks visibility is currently part of the standard user experience.

If a downstream review prefers full admin-only, the change is one-line: move all `.route(...)` calls into `admin_router()`. Happy to flip that on request.

### Pattern: route-level (matches existing codebase)

The codebase already has the route-level admin pattern in `backend/src/api/handlers/plugins.rs` (`format_router` + `format_admin_router`). This PR follows that exact pattern. Handler-level `auth.require_admin()` was considered and rejected because it would require touching six handlers and would not give us the structural compile-time guarantees the route-level split provides.

### Tests

Two new structural unit tests in `backend/src/api/handlers/webhooks.rs` (no DB required):

- `test_router_and_admin_router_merge_without_overlap` -- merges the two routers; axum panics at construction if any path is registered on both, so this is a regression guard. If a future PR adds a write handler back into `router()`, this test fails at construction.
- `test_router_functions_construct` -- sanity check that each sub-router builds in isolation.

End-to-end coverage of the admin gate itself comes from the existing `admin_middleware` tests in `backend/src/api/middleware/auth.rs` (`test_require_admin_*` family).

Verified locally:

```
SQLX_OFFLINE=true cargo check --workspace      # clean
SQLX_OFFLINE=true cargo test --lib api::handlers::webhooks::tests::    # 54/54 passing
SQLX_OFFLINE=true cargo test --lib test_openapi_spec_is_valid  # passing
cargo fmt --check                              # clean
cargo clippy --workspace --no-deps             # clean
```

### Backward-compatibility

Non-admin users who previously created webhooks keep their existing rows. They can still GET them. They can no longer:

- Create new webhooks (`POST /api/v1/webhooks` -> 403).
- Delete or modify their existing webhooks.
- Enable, disable, test, or redeliver.

Migration path: an admin can re-register on the user's behalf, or run an UPDATE on the `webhooks` table to attribute existing rows to an admin owner once per-tenant scoping (#948) lands.

The OpenAPI shape is unchanged (same paths, same schemas). The only observable change is a 403 response on the write endpoints when called with non-admin credentials.

## Test Checklist
- [ ] Integration tests added/updated (if applicable) -- N/A, no DB-bound tests added; admin gate covered by existing middleware tests
- [ ] E2E tests added/updated (if applicable) -- recommend follow-up E2E in `artifact-keeper-test` to assert non-admin -> 403 on POST /webhooks
- [x] Manually tested locally

## API Changes
- [x] New endpoints have `#[utoipa::path]` annotations -- N/A, no new endpoints
- [x] Request/response types have `#[derive(ToSchema)]` -- N/A
- [x] Migration is reversible (if applicable) -- N/A, no schema change
- [ ] N/A - no API changes (semantic change: write endpoints now return 403 for non-admins)


brandonrc:
Pushed `f32f0e1d` cherry-picking the wasmtime 36.0.7→36.0.8 bump from #992. Same RUSTSEC-2026-0114 advisory was failing this PR's Security Audit. Both PRs target release/1.1.x and need the same fix until one merges and the other rebases.

`cargo audit`: 5 allowed warnings only, no errors. Awaiting fresh CI.

PR #996, opened 2026-05-01
fix(security): scanner silent-success on type-mismatched artifacts (#994)

### LINKED ISSUE #994: Backend: scanner returns findings_count=0 in 3ms on known-vulnerable lodash fixture (silent-success class, surfaced on 1.1-dev)
## Symptom

Release-gate run [25214268566](https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/25214268566) caught `tests/security/test-scan-completes.sh > Scan emitted findings on known-vulnerable fixture (regression for #888 silent-success class)` failing against the `:1.1-dev` image.

The test uploads a curated fixture pinned to **lodash 4.17.4** (CVE-2019-10744, Prototype Pollution, CVSS 9.1). Trivy's fs scanner is expected to detect this via the `package-lock.json` parser and report `findings_count >= 1`. On this run, the backend returned:

```json
{
  "id": "ac12df61-80e5-40cf-918d-b85c502b0c4b",
  "artifact_name": "lodash-vuln-fixture",
  "artifact_version": "1.0.0",
  "scan_type": "image",
  "status": "completed",
  "findings_count": 0,
  "critical_count": 0,
  "high_count": 0,
  "medium_count": 0,
  "low_count": 0,
  "info_count": 0,
  "scanner_version": null,
  "error_message": null,
  "started_at": "2026-05-01T12:34:04.061836Z",
  "completed_at": "2026-05-01T12:34:04.064655Z"
}
```

`completed_at` is **2.8 ms** after `started_at`. No real Trivy fs scan against a tarball runs in 2.8 ms. The scanner did not inspect the bytes; the backend marked the scan COMPLETED with zero findings anyway. This is exactly the silent-success class issue #888 closed for the Docker image path.

## Why this is a regression

Issue #888 fixed the Docker image scan path -- a scan against an unreachable scanner pod no longer returns COMPLETED with empty results. The lite scan-completes E2E (artifact-keeper-test#888) was added to enforce that contract. Run 11 of the release gate is the first run on `:1.1-dev` that exercises this assertion against the v1.1.x line, and it fails.

The 2.8 ms turnaround + `scanner_version: null` strongly suggest one of the four causes already enumerated in the test's diagnostic message:

1. Trivy is not enabled in the test deploy (chart values: `trivy.enabled`).
2. Trivy fs scanner's `is_applicable` check rejected the fixture format.
3. `scanner_service`'s `Ok(vec![])` path treated empty as success.
4. `find_reusable_scan` returned a stale empty scan from a prior run.

(1) is unlikely -- the surrounding scan tests (`test-scan-findings-list`, `test-trivy-scan`) passed in the same run, so Trivy is reachable. (2) is most likely: lodash-vuln-fixture is uploaded as `image` scan_type but the artifact is a generic tarball, not a container image. The fs scanner's `is_applicable` may be checking content-type and rejecting it, then the backend short-circuits to `Ok(vec![])`. (3) and (4) cannot be ruled out without the backend logs.

## Expected behavior

For an artifact whose bytes Trivy cannot inspect (wrong scan_type, unsupported format), the backend should NOT report `status=completed, findings_count=0, error_message=null`. It should either:

- Return `status=failed` with `error_message` describing why the scanner skipped, OR
- Return a structured "not applicable" status that does not look identical to a clean scan.

The current behavior is indistinguishable from a real clean scan and silently lies about the artifact's vulnerability posture.

## Where to look

- `backend/src/services/scanner_service.rs` -- the `Ok(vec![])` short-circuit path.
- `backend/src/scanners/trivy.rs` (or equivalent) -- `is_applicable` rejection path: does it propagate as an error or as empty success?
- `backend/src/services/scan_orchestrator.rs` (or similar) -- when no applicable scanner is found, what status does the scan transition to?

## Reproduction

Test repo: `artifact-keeper-test`, file: `tests/security/test-scan-completes.sh`. Stand up `:1.1-dev` against the test chart and run the security suite. The fixture build is in the same script.

## Tracking

- Surfaced by release-gate Run 11 (#127 milestone work)
- Test that catches this: `tests/security/test-scan-completes.sh > Scan emitted findings on known-vulnerable fixture` (load-bearing assertion, see artifact-keeper-test#888)
- Upstream for the v1.2.x fix path: #902 (`scanner_version` populated in ScanResponse)
- Block on v1.1.9 release acceptance (#886): the lite-provenance + findings_count assertions are exactly what the gate exists to enforce. Either the scan path needs to fail this fixture honestly on v1.1.9, or the test fixture needs to be retargeted at a Docker image scan path (which `test-trivy-scan.sh` already covers).

brandonrc:
## Summary

Closes #994 — silent-success class regression on `:1.1-dev` flagged by release-gate run 11 against `tests/security/test-scan-completes.sh`.

### Symptom

```
findings_count=0, scanner_version=null, error_message=null
completed_at - started_at = 2.8ms
fixture: lodash 4.17.4 (multiple known CVEs)
```

The backend reported a successful scan completion with zero findings on a tarball that should have a half-dozen CVEs. The 2.8 ms wall-clock is way too short for a real Trivy scan and is the fingerprint of an `is_applicable` short-circuit plus a DB round trip, no scanning ever happened.

### Root cause

`ScannerService::scan_artifact_with_options` runs *all* registered scanners against every artifact:

1. `DependencyScanner` (always applies, queries OSV/GitHub Advisory)
2. `ImageScanner` (`vnd.oci.image` manifests)
3. `TrivyFsScanner` (tarballs, wheels, jars, etc)
4. `IncusScanner` (LXC images)
5. `GrypeScanner` (always applies)
6. `OpenScapScanner` (containers, RPMs, DEBs)

Four of these (`ImageScanner`, `TrivyFsScanner`, `IncusScanner`, `OpenScapScanner`) had a private `is_applicable` check at the top of their `scan()` methods. When not applicable they returned `Ok(vec![])`, which the orchestrator could not distinguish from "scanner ran and found nothing". The orchestrator persisted the result as `status='completed'`, `findings_count=0`, `scanner_version=NULL`, `error_message=NULL` — indistinguishable from a real clean scan.

For a `lodash-4.17.4.tgz` upload the DB ends up with six `scan_results` rows (one per scanner). Five are real, but the `image` row from `ImageScanner` has the silent-success shape: it never inspected the artifact, yet looks completed. Receivers that use `scan_results` as a security gate (`/api/v1/security/artifacts/{id}/scans`) wave it through.

The hypothesis from prior triage was correct: `is_applicable` rejection short-circuits to `Ok(vec![])` without surfacing the rejection.

### Fix

- Add `Scanner::is_applicable(&Artifact, Option<&ArtifactMetadata>) -> bool` to the `Scanner` trait. Default returns `true` so always-applicable scanners (`DependencyScanner`, `GrypeScanner`) keep working without override.
- Override in the four scanners that have applicability checks (`ImageScanner`, `TrivyFsScanner`, `IncusScanner`, `OpenScapScanner`).
- In `ScannerService::scan_artifact_with_options`, check `is_applicable` BEFORE creating the `scan_results` row. Non-applicable scanners now leave no DB trace.
- Keep the in-`scan()` short-circuits as defense in depth for direct callers and tests that bypass the orchestrator.

A `completed` `scan_results` row is now a real signal that the scanner inspected the artifact.

### Why a structured error was not chosen

`is_applicable=false` is a routine, expected outcome (the orchestrator queries 4-6 scanners per artifact, most won't apply). Returning `Err` would force the orchestrator to inspect the error variant to distinguish "real failure" from "wrong scanner type" and would clutter logs. Skipping silently with a single `info!` log is the right contract: the orchestrator's job is to dispatch to applicable scanners, and the `scan_results` table records actual scan attempts.

### Regression tests

Three trait-contract tests in `scanner_service::tests`:

- `test_npm_tarball_is_not_applicable_to_image_or_incus_or_openscap_scanners` — pins that the lodash fixture from the failing release-gate test correctly rejects the wrong scanners and matches `TrivyFsScanner`.
- `test_oci_manifest_is_applicable_to_image_scanner_only` — inverse case, OCI manifest goes only to `ImageScanner`.
- `test_default_is_applicable_is_true` — verifies the trait default keeps always-on scanners working.

LoC: ~125 lines of new code (~30 production, ~95 tests + comments). Single-trait surgical change.

## Test Checklist
- [x] Unit tests added/updated (3 new regression tests for trait contract)
- [ ] Integration tests added/updated (deferred — the structural fix is exercised by the release-gate `test-scan-completes.sh`)
- [ ] E2E tests added/updated (release-gate already covers)
- [x] Manually tested locally (`SQLX_OFFLINE=true cargo test --lib` — 7732 passed, 0 failed)

## API Changes
- [x] N/A - no API changes

## Concerns for reviewers

- **Behavior change for callers that relied on silent-success.** The previous behavior produced one `scan_results` row per (artifact, scanner) pair. Tooling that counts rows or assumes "every scanner is represented in `scan_results`" will now see fewer rows for a given artifact. The only places that read `scan_results` in the codebase are the security handlers and the security-score recalculator, both of which already aggregate by `findings_count`/`severity_count` and are unaffected (an absent row contributes zero, same as before). External tooling that polls `scan_results` directly may notice fewer rows — flagging for human review but I do not see this pattern in the workspace.
- **Dedup cache key.** `find_reusable_scan` queries by `(checksum, scan_type)`. Skipping the row creation means a non-applicable scanner will not seed the dedup cache for that scan_type. This is correct: we should not be reusing a non-existent scan as a "fresh" result.
- **`scanner_version=null` is still possible** for real completed scans because no scanner currently writes that column. That is a separate gap not addressed here. After this fix, `scanner_version=NULL` no longer correlates with silent-success since the silent-success rows simply do not exist anymore.


brandonrc:
## Follow-up commit: retire legacy silent-success rows (1bca4cb1)

This commit addresses the security review finding that the original PR only blocks NEW silent-success rows but leaves the v1.1.0-v1.1.8 buggy rows in production databases. Quality gates querying "latest scan" would still see deceptive `completed`/`findings_count=0` rows and wave through unscanned artifacts.

### What changed

**Migration 075** (`backend/migrations/075_flag_legacy_unverified_scans.sql`):
- Adds `scan_results.legacy_unverified BOOLEAN NOT NULL DEFAULT false`
- Flags existing rows matching the bug fingerprint: `status='completed' AND scanner_version IS NULL AND findings_count=0 AND error_message empty`
- Creates partial index `idx_scan_results_artifact_verified` for the new consumer query path
- Idempotent: `ADD COLUMN IF NOT EXISTS`, `CREATE INDEX IF NOT EXISTS`, the UPDATE narrows on the bug fingerprint so re-running is a no-op

The fingerprint is conservative: real scans (including real zero-finding clean scans) always populate `scanner_version`, so we will not flag legitimate clean scans.

**Consumer queries updated to filter `legacy_unverified = false`** (6 queries across 5 files):

| File | Function | Effect |
|------|----------|--------|
| `policy_service.rs` | `evaluate_artifact` | `block_unscanned` policy now treats legacy rows as unscanned, blocking download until a real scan completes |
| `promotion_policy_service.rs` | `get_cve_summary` | Promotion gate cannot read 0/0/0/0 counts from a legacy row |
| `promotion_rule_service.rs` | `get_latest_scan` | Same for rule-based promotion |
| `scan_result_service.rs` | `find_reusable_scan` | Dedup will not reuse a legacy row for a fresh artifact |
| `scan_result_service.rs` | `recalculate_score` last_scan_at | Dashboards do not report legacy `completed_at` as the last scan time |
| `sbom.rs` | `extract_dependencies_for_artifact` | SBOM dependency extraction excludes legacy rows |

**CHANGELOG entry moved from `### Fixed` to `### Security`** with the operator audit query and rescan SOP for upgrades from v1.1.0-v1.1.8.

**Tests added** (3 new unit tests in `scanner_service::tests`, all DB-free):
- `test_dependency_and_grype_scanners_are_applicable_to_any_artifact` -- pins the default-true contract that the always-on scanners depend on, across 6 artifact shapes
- `test_non_applicable_scanners_return_empty_findings_when_called_directly` -- defense-in-depth check that Image/TrivyFs/Incus/OpenScap scanners still return `Ok(vec![])` when called directly with a non-applicable artifact (preserves the existing short-circuit pattern for direct callers and tests)
- `test_orchestrator_gate_skips_non_applicable_scanners_before_persisting` -- unit-level analog of the orchestrator integration test, validates the if-not-applicable continue pattern using a counting scanner

### Verification

- `SQLX_OFFLINE=true cargo check --workspace` passes
- `SQLX_OFFLINE=true cargo clippy --workspace --lib --tests -- -D warnings` clean
- `cargo fmt --check` clean
- All 7735 lib tests pass; the 8 tests in the #994 block (5 prior + 3 new) all pass
- Migration 075 applied + re-applied successfully against a fresh DB; column and partial index created

### Reviewer notes

- **Behavior change for operators**: Artifacts with only legacy rows will now be blocked at download under `block_unscanned` policies. This is the security-correct outcome and is documented in the CHANGELOG, but operators upgrading from v1.1.0-v1.1.8 will see download failures until they run the audit query and trigger rescans. The audit query is included in the CHANGELOG.
- **Migration locking**: The `UPDATE scan_results SET legacy_unverified = true WHERE ...` runs in a single statement and acquires a row-level lock per matched row plus the table-level `ALTER TABLE` lock for the `ADD COLUMN`. On large `scan_results` tables (>1M rows) this could pause writes briefly. The `ADD COLUMN ... DEFAULT false` is fast in PG 11+ (no rewrite). The UPDATE only touches matching rows so impact scales with the buggy-row population, not the table size. No advisory recommended for typical deployments; large-scale deployments may want to plan the upgrade window.
- **`ScanResult` model not updated**: The `legacy_unverified` column is internal to query filtering and is not surfaced through the `ScanResult` struct or the API. This keeps the change minimal and avoids touching the public security API. If a future feature wants to surface "this scan was retired by audit" we can add the field then.
- **No `block_unscanned` consumer test added**: `policy_service.rs` has no DB-mocked test infrastructure (no `#[sqlx::test]` patterns anywhere in the codebase). Adding one would require introducing a new pattern for this PR. The unit-level orchestrator gate test plus the consumer query SQL change covers the contract; a follow-up E2E test in `artifact-keeper-test` could verify the `block_unscanned` end-to-end behavior with a flagged row.

PR #984, opened 2026-04-30
fix(handlers): escape LIKE wildcards in rubygems/rpm/helm/npm

brandonrc:
## Summary

Sister fix to #881 (maven LIKE wildcard escape). The same bug class exists
in 4 other package handlers — user-supplied path components are interpolated
into a parameterized LIKE clause without escaping `%` and `_`, so an
attacker can match arbitrary path patterns within a single repository
(info disclosure / wrong-artifact serving).

Affected sites:
- `rubygems.rs:260` — `download_gem`
- `rpm.rs:435` — `download_package`
- `helm.rs:234` — `download_chart`
- `npm.rs:748` — `npm_local_fetch`
- `npm.rs:854` — second LIKE pattern

Adds a shared `escape_like_literal` helper to `api::handlers::mod.rs`
(alongside the existing `cleanup_soft_deleted_artifact`). Each site
escapes user-supplied fragments and adds `ESCAPE '\'` to the SQL.

Not SQL injection — sqlx parameterizes correctly. Wildcard-injection only.

Security details: see GHSA (private advisory).

## Test Checklist
- [x] Unit tests added/updated (5 new in `api::handlers::tests` — safe input, percent, underscore, backslash, combined adversarial)
- [ ] Integration tests added/updated — N/A; bug is in pattern-builder, covered by unit tests
- [ ] E2E tests added/updated — N/A
- [x] Manually tested locally (3368 lib tests pass)
- [x] No regressions in existing tests

## API Changes
- [x] N/A - no API changes

## Notes

Behavior change: package filenames and package names containing `%` or `_`
are now treated as literals rather than LIKE wildcards. No legitimate
package coordinate contains these characters in the affected handlers.

Regenerated `.sqlx/` query cache for the four `query!()` macro invocations
that gained the `ESCAPE '\'` clause.

PR #998, opened 2026-05-01
fix(handlers): close LIKE wildcard-injection in pypi + shared proxy helper (audit gap from #984)

brandonrc:
## Summary

Closes the **HIGH-severity** sites of the LIKE wildcard-injection class that #984/#988 left unfixed:

| Handler | Function | Severity |
|---|---|---|
| `pypi.rs` | `serve_file` | High — `%.whl` matches arbitrary wheels in repo |
| `pypi.rs` | `serve_metadata` | High |
| `proxy_helpers.rs` | `local_fetch_by_path_suffix` | High — shared helper used by 9 format handlers |

Each site now escapes user-supplied path components and adds \`ESCAPE '\\'\` to the LIKE clause. The `local_fetch_by_path_suffix` helper escapes internally so its 9 callers can pass raw user input without remembering to pre-escape (its docstring documents the contract).

Same severity class as the maven (#881) and npm/helm/rpm/rubygems (#988) sites already in v1.1.9: info disclosure / wrong-artifact serving within a single repository. Not SQL injection — sqlx parameterizes correctly.

## Scope split

This PR was originally larger (10 sites + helpers). The medium-severity sites (ansible/cran/hex/puppet/alpine/huggingface/debian/terraform) have been **deferred to v1.1.10** because:

1. They're medium severity per the security review (not widely-deployed/high-traffic ecosystems).
2. The proxy-fallback pattern in those files is structurally similar (pre-existing across the codebase), which trips the jscpd duplication gate when all are touched in one PR.
3. Splitting by severity lets v1.1.9 ship the highest-impact fix on time while the medium-severity sites get a cleaner refactor (potentially extracting the proxy-fallback into a shared helper) for v1.1.10.

A follow-up issue tracks the v1.1.10 work.

## Helpers

Two new helpers added in \`api::handlers::mod.rs\` (with 7 new unit tests):
- \`escape_filename_for_like\` — strip leading slash + escape (used by pypi handlers)
- \`escape_path_prefix\` — multi-component LIKE prefix construction (queued for v1.1.10 consumers)

## Test Checklist
- [x] Unit tests added/updated — 7 new tests for the helpers in \`mod.rs::tests\` (12 total escape tests now)
- [ ] Integration tests added/updated (if applicable) — N/A; existing E2E suites cover the download paths
- [x] Manually tested locally — \`cargo check --lib --offline\`, \`cargo clippy --lib --all-targets -- -D warnings\`, \`cargo fmt --check\`, helper tests all clean

## API Changes
- [x] N/A — no API changes

## Notes

Behavior change: pypi filenames containing \`%\` or \`_\` are now treated as literals rather than LIKE wildcards. No legitimate PyPI distribution filename contains these characters.

Regenerated \`.sqlx/\` cache: 4 entries get new hashes (escape clause added), several stale entries removed.


brandonrc:
Security review caught one more site of the same class: `terraform.rs::download_provider` (line 756). `os` and `arch` come from URL Path and were interpolated into `format!("{}_{}", os, arch)` then into `path LIKE '%' || $4 || '%'` without escape.

Pushed commit `f1170f54` adding the fix. Same pattern as the other sites — apply `escape_like_literal` to the joined platform tag, add `ESCAPE '\'` to the SQL. Bonus: this also fixes a pre-existing correctness bug where the literal `_` separator was acting as a LIKE single-char wildcard (e.g. `linux_amd64` matched `linuxXamd64` paths).

PR now closes the LIKE-wildcard class on **10 sites** + 1 shared helper across all package format download paths in v1.1.x.

PR #1017, opened 2026-05-04
fix(migration): create destination repos before transferring artifacts

ThaSami:
## Bug

A `full` or `incremental` migration job runs to `status=completed` with all per-item checksums matching, but the destination has 0 repos and `GET` on any migrated path returns 404. Bytes are written to CAS storage but unreferenced by any DB row, so they're unreachable via the format-protocol APIs.

This is silent data loss disguised as success.

### Where

`backend/src/services/migration_worker.rs`:

* `process_job` iterates `config.include_repos` and calls `process_repository_artifacts` directly — `migration_service.create_repository` is never invoked, so destination repos are never provisioned.
* `transfer_artifact` writes bytes via `storage.put`, then attempts `INSERT INTO artifacts` only inside `if let Some((repository_id,)) = repo_id { … }`. When the lookup fails (because the destination repo doesn't exist), the `if let` silently no-ops — no `tracing::warn!`, no `MigrationError` returned, the item is marked `completed`.

### Reproduction

Verified end-to-end against an Artifactory OSS source seeded with 11 typed repos (Maven, npm, PyPI, Helm, NuGet, Go, CocoaPods, Conan, Swift, Docker, Generic) and 13 artifacts:

```
POST /api/v1/migrations            →  201, job created
POST /api/v1/migrations/{id}/start →  200
GET  /api/v1/migrations/{id}       →  status=completed, completed_items=15, failed_items=0
GET  /api/v1/repositories          →  { items: [] }                ← repos never created
GET  /maven/<repo>/<path>          →  404                          ← artifacts not addressable
```

Backend logs around the same job show `Repository artifacts processed repo=maven-releases` (and so on) with **no** preceding `Provisioned destination repository` / `Creating repository` line — confirming `create_repository` was never invoked.

## Fix

Wrap the existing per-repo loop in `process_job` with a pre-pass that ensures the destination row exists for each `repo_key in include_repos`:

1. Fetch the source-side repo list once via `client.list_repositories()`.
2. For each requested repo:
   * Look up the source `RepositoryListItem`. If not present, log + count as failed + skip.
   * Build a `RepositoryMigrationConfig` via the existing `MigrationService::prepare_repository_migration` helper.
   * Probe `check_repository_conflict`:
     * **No conflict** → call `create_repository` to provision the row.
     * **`SameKey` conflict** (existing destination already matches type+format) → reuse it (idempotent).
     * **Other conflict types** → log + count as failed + skip; subsequent repos still get a chance.
3. Only then enter the existing artifact-transfer loop.

Subsequent calls to `process_repository_artifacts → transfer_artifact` then find a `Some(repository_id)` and the artifact INSERT runs.

The fix is minimally invasive: it touches only `process_job`, no changes to `transfer_artifact` or `process_repository_artifacts`.

## Limitations / follow-ups

* **Remote (proxy) repos** are migrated with `upstream_url=None` because `prepare_repository_migration` is called with `_repo_config: None`. Pulling per-repo detailed config (upstream URL, virtual members) requires a `get_repository(key)` method on the `SourceRegistry` trait, which isn't present yet — that's a separate PR.
* **Virtual (group) repos** are migrated with empty members for the same reason.
* **Local (hosted) repos** work fully with this PR.

## Other migration-tool gaps observed during a Careem PoC

(Not in scope for this PR, will file as separate issues.)

* `/pause` silently truncates a job to `completed` (operator can lose unfinished work to a single API call).
* `concurrent_transfers` is a dead config field — `process_repository_artifacts` is a sequential `for { … .await }` with no parallelism primitives.
* `conflict_resolution` is functionally dead (coupled to this bug — the `artifacts` table stays empty so `check_artifact_duplicate` always misses).
* `include_users` / `include_groups` / `include_permissions` silently do nothing against an OSS source (no warning, no error, no items in the items endpoint).
* `MigrationConfig` schema isn't exported via `utoipa::ToSchema` — field names like `include_repos` are only discoverable from the Rust source.

Happy to file these separately and hack on them once this PR settles.

## Test plan

* [x] Existing unit tests pass (no test signatures changed)
* [ ] CI passes on the PR
* [ ] (Manual) run an end-to-end migration of a single Maven local repo against an Artifactory OSS source: verify `GET /api/v1/repositories` shows the destination repo, `GET /maven/<repo>/<path>/<artifact>` returns the artifact bytes


brandonrc:
@ThaSami  Thank you so much for taking time to make this PR! You are also passing all the tests which are amazing. 

Taking a look at this right now.

brandonrc (approved):

Issue #962, opened 2026-04-28
Scanner vulnerability counts a sum of all scans and not the latest per scan


Firjens:
### Description

Scan a Docker image with let's say 15 vulnerabilities 10 times and it'll say there are 150 vulnerabilities on the security dashboard.

<img width="3107" height="1062" alt="Image" src="https://github.com/user-attachments/assets/5aa65858-3606-4b33-93e6-5e9dc5e23201" />

Instead of showing the sum of all scans it should pick the sum of the latest of all scans.

### Environment Information

- Deployment type: 
- Operating System: 
- Architecture: 
- Database: 
- Reverse Proxy: 
- Network setup: 

### Backend Version

latest

### Package Format

Docker / OCI

### Expected Behavior

Instead of showing the sum of all scans it should pick the sum of the latest of all scans.

### Observed Behavior

Scan a Docker image with let's say 15 vulnerabilities 10 times and it'll say there are 150 vulnerabilities on the security dashboard.

### Steps to Reproduce

- Scan a Docker image with let's say 15 vulnerabilities 10 times 
- View the dashboard and see it says there are 150 vulnerabilities

### Logs

```shell

```

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have redacted any sensitive information (passwords, tokens, private URLs)
- [x] I am using a supported version (not end-of-life)

PR #1123, opened 2026-05-09
ci: run #[ignore]'d integration tests in the integration job (closes #1031)

### LINKED ISSUE #1031: CI integration-test convention: drop #[ignore] or pass --ignored consistently
Surfaced during #962 R1 DevOps + #[ignore] fix-up.

`backend/tests/` has 12+ integration test files where every `#[tokio::test]` is `#[ignore]`'d (lifecycle_policy_tests, refresh_token_rotation_tests, meili_reindex_tests, azure_rbac_live_test, incus_upload_tests, api_token_cache_flush_tests, grpc_sbom_tests, s3_integration, replication_integration, integration_tests, azure_shared_key_live_test, storage_backend_tests).

CI runs `cargo test --workspace --test '*'` without `--ignored` — so all of these are silently skipped. The integration-test job at `ci.yml:320-340` provisions postgres, so they COULD run.

#962's regression test had `#[ignore]` dropped surgically to gate merges, creating a divergent convention. Resolution options:
1. Drop `#[ignore]` from all DB-backed integration tests now that postgres is provisioned, OR
2. Update `ci.yml` to pass `-- --ignored` so existing convention stands

Choose one and apply consistently. Pick (2) is one-line CI change but runs more tests. Pick (1) is more rust-idiomatic.

Refs: #962.

brandonrc:
## Summary

The \`test-backend-integration\` step in \`ci.yml\` provisions postgres and runs \`cargo test --workspace --test '*'\` without \`--ignored\`. The 12+ integration test files in \`backend/tests/\` (lifecycle_policy_tests, refresh_token_rotation_tests, meili_reindex_tests, api_token_cache_flush_tests, grpc_sbom_tests, integration_tests, azure_rbac_live_test, incus_upload_tests, azure_shared_key_live_test, replication_integration, s3_integration, storage_backend_tests) all carry \`#[ignore = "requires running DB / HTTP server"]\` because they cannot run on the unit-test job (no DB). They are exactly what the integration job was provisioned to run, but \`--ignored\` was missing so they were silently skipped.

Pass \`-- --ignored\` so the existing #[ignore] convention stands and the tests actually execute. This expands what CI verifies on every push to main / release/* without changing any test code.

Surfaced during #962 R1 DevOps + #[ignore] fix-up.

## Test Checklist
- [x] CI workflow change validated (`actionlint`, `yaml.safe_load`)
- [ ] Unit tests added/updated (CI-only change)
- [ ] Integration tests added/updated (this PR makes existing ones run)

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -346,14 +346,18 @@ jobs:
         env:
           DATABASE_URL: postgresql://registry:registry@localhost:5432/artifact_registry
           SQLX_OFFLINE: true
-        # Pass `-- --ignored` so DB-backed integration tests gated by
-        # `#[ignore = "requires running DB / HTTP server"]` actually
-        # execute now that this job provisions postgres. Without this,
-        # 12+ integration test files (lifecycle_policy_tests,
-        # refresh_token_rotation_tests, meili_reindex_tests,
-        # api_token_cache_flush_tests, grpc_sbom_tests, integration_tests,
-        # etc.) were silently skipped. See #1031.
-        run: cargo test --workspace --test '*' --verbose -- --ignored
+        # Pass `-- --include-ignored` so the integration job runs BOTH
+        # ordinary tests AND tests gated by `#[ignore = "requires
+        # running DB / HTTP server"]`, now that this job provisions
+        # postgres. Without `--include-ignored`, the 12+ integration
+        # test files (lifecycle_policy_tests, refresh_token_rotation_tests,
+        # meili_reindex_tests, api_token_cache_flush_tests, grpc_sbom_tests,
+        # integration_tests, etc.) were silently skipped.
+        #
+        # `-- --ignored` (without `include-`) is a footgun: it would do
+        # the opposite, running ONLY ignored tests and skipping the rest.
+        # See #1031.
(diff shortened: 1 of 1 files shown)


brandonrc:
Closing per round-1 code-quality review findings.

`--include-ignored` is too broad: it would force `azure_rbac_live_test` and `azure_shared_key_live_test` to panic at `.expect("AZURE_STORAGE_ACCOUNT not set")` (no skip-if-unconfigured guard), and ~40 HTTP-server-dependent tests across `storage_backend_tests`, `download_ticket_tests`, `replication_integration`, `integration_tests` would fail with connection-refused since the integration CI job provisions postgres only, not a running backend HTTP server.

Right fix is structural, not a one-line CI change:
1. Add `skip_unless_configured!` style guards to live-cloud tests (Azure, S3 — S3 already has one, mirror that pattern).
2. Move HTTP-server-dependent tests to a separate job that spins up a backend (or convert them to use `axum_test`).
3. Then narrow what `--include-ignored` covers, OR drop `#[ignore]` from the now-runnable DB-only suites individually.

Filed as #1031 still; will re-author with the structural fix.

PR #1137, opened 2026-05-09
fix(scanner): TTL retry-after-None semantics for cached version (closes #1012)

### LINKED ISSUE #1012: Scanner: OnceCell never-retry-after-None semantics for cached_version (no TTL)
Follow-up from PR #1006 review.

## Background

PR #1006 added \`Scanner::version()\` with per-scanner \`tokio::sync::OnceCell<Option<String>>\` caching. The cached value is set exactly once per process lifetime via \`get_or_init\`. This means: if the very first probe returns \`None\` (binary missing, transient timeout, version flag glitch), every subsequent call returns \`None\` for the rest of the process, even after the underlying issue is resolved (operator installs the binary, restarts the wrapper, fixes PATH, etc.).

The \`scan_results.scanner_version\` column will silently stay NULL for every scan run by that backend instance until the pod is restarted.

## Proposal

Replace the bare \`OnceCell<Option<String>>\` with a TTL-backed cell that:

- Caches \`Some(version)\` for a long TTL (e.g. 1 hour) since binary versions don't change without a deploy.
- Caches \`None\` for a short TTL (e.g. 60s) so transient failures are retried without spamming \`--version\` calls.

Implementation options:
1. Custom struct over \`tokio::sync::RwLock<(Instant, Option<String>)>\` (mirrors \`AdvisoryClient::cache\` pattern in \`scanner_service.rs\`).
2. \`once_cell::sync::Lazy\` with explicit invalidation hook on probe-failure.

Option 1 is simpler and consistent with existing code style.

## Acceptance criteria

- After a probe returns \`None\`, the next probe call retries the binary at most once per TTL window.
- After a probe returns \`Some(version)\`, subsequent calls for at least an hour are O(1) RAM-only (no subprocess).
- Unit tests cover both retry-after-None and no-spam-after-success.

## Severity

Low. The PR is correct; this is a UX/observability follow-up.

## Original review thread

Surfaced in the PR #1006 review by Backend Architect, captured under "Non-blocker follow-ups" in commit 8fb12540 of the same PR.

brandonrc:
## Summary

`Scanner::version()` previously used `tokio::sync::OnceCell<Option<String>>` to memoize the result of the CLI version probe. If the first probe returned `None` (binary missing on PATH, init container still pulling, scanner pod momentarily unreachable), the OnceCell pinned that `None` for the entire process lifetime. The `scan_results.scanner_version` column then silently stayed NULL on every subsequent scan until the backend pod was restarted, even after the operator installed the missing binary.

This PR replaces the OnceCell with a TTL-backed `VersionCache` in `scanner_service.rs`:

- `Some(version)` is cached for `VERSION_CACHE_HIT_TTL` (1 hour). Versions only change on deploy/restart, so a long hit TTL is safe and avoids paying a subprocess invocation per scan.
- `None` is cached for `VERSION_CACHE_MISS_TTL` (60s). Long enough to dampen probe storms when the binary is genuinely missing, short enough that the version column starts populating within a minute once the operator fixes the underlying issue.

Updated all five scanners that hold a cached CLI version (`trivy_fs_scanner`, `image_scanner`, `incus_scanner`, `grype_scanner`, `openscap_scanner`) to use the new `VersionCache` type. The `cached_cli_version` and `cached_trivy_cli_version` helper signatures keep their closure-returning-future shape, so per-scanner override bodies stay one-liners.

The slow path probes outside the lock (an external `Command` invocation can take seconds, holding a write lock that long would serialize every concurrent scan), then re-checks the cell under the write lock so a refresh by a concurrent caller wins over our just-probed value. Two callers may race a probe on cache miss; that is harmless and strictly better than blocking.

- [ ] N/A, this is not a bug fix

The new `test_version_cache_reprobes_after_miss_ttl` directly exercises the bug: it seeds the cache with an aged-out `None`, then confirms the next call re-probes and picks up the fresh `Some(_)`. Under the old OnceCell implementation, the second probe would never run.

## Test Checklist
- [x] Unit tests added/updated (5 new tests in `scanner_service::tests`: caches Some without re-probe, caches None within MISS_TTL, re-probes after MISS_TTL expiry, Some survives past MISS_TTL via the longer HIT_TTL, TTL constants sanity)
- [x] Manually tested locally (`cargo test --lib services::scanner_service` passes 312/312, full scanner module suite passes 82/82)

## API Changes
- [x] N/A, no API changes

Closes #1012.


brandonrc:
Round-1 review feedback rolled into this MR (per maintainer direction not to fan out follow-up PRs):

**Thundering herd on miss-path** (security MED / concurrency LOW)
Added `probe_lock: Mutex<()>` to `VersionCache`. Concurrent cold-cache callers now produce exactly 1 probe; the rest wait on the lock and read the just-written entry on re-check. Tests:
- `test_version_cache_concurrent_miss_single_flights_probe` (2 callers, asserts 1 probe)
- `test_version_cache_single_flight_under_fanout` (16 callers, asserts 1 probe)

**Hardcoded TTLs** (code-quality MED)
TTLs are now read at `VersionCache::new()` time from `AK_SCANNER_VERSION_HIT_TTL_SECS` / `AK_SCANNER_VERSION_MISS_TTL_SECS`, falling back to compile-time defaults (3600 / 60) when unset, unparseable, or zero. Reading once at construction keeps the hot path env-free. Tests:
- `test_env_duration_secs_parses_overrides_and_falls_back`
- `test_version_cache_reads_ttl_env_at_construction`

8 passing unit tests on the cache. All 316 scanner_service tests green. Clippy clean with `-D warnings`.

PR #1144, opened 2026-05-10
fix(oci): keep storage gc from deleting referenced OCI data

inspired-geek:
## Summary

- Prevent storage GC from treating OCI manifest storage keys as orphaned while `oci_tags` still references the manifest digest.
- Prevent storage GC from treating OCI blob storage keys as orphaned while `oci_blobs` still references the blob digest.
- Cover the stale-artifact/live-OCI-metadata failure mode that left `list-tags` returning tags while `GET /v2/.../manifests/<tag>` returned `MANIFEST_UNKNOWN` after GC deleted the underlying object.

Regression coverage:

- `test_run_gc_dry_run_keeps_oci_manifest_referenced_by_tag`
- `test_run_gc_dry_run_keeps_oci_blob_referenced_by_blob_index`

These DB-backed tests create a soft-deleted `artifacts` row plus live OCI metadata and assert dry-run GC does not count the storage key as deletable. Without the new OCI metadata guards, the manifest regression reports one deletable storage key.

## Test Checklist

- [x] Manually tested locally

Commands run:

- `cargo fmt --check`
- `SQLX_OFFLINE=true CARGO_HOME=/tmp/cargo-ak-home cargo test -p artifact-keeper-backend storage_gc_service --lib`

## API Changes

- [x] N/A - no API changes


brandonrc:
Thanks for tracking this down. Losing referenced OCI data is one of the worst possible regressions and the `NOT EXISTS` guards against `oci_tags` and `oci_blobs` close the most common path cleanly. The two regression tests are the right shape too.

Three concerns to discuss before merge:

### 1. Multi-arch image index child manifests still unprotected

When a multi-arch image is pushed, the OCI image index manifest gets its own `oci_tags` row (the top-level tag points at the index digest), so the new guard protects the index file. But the per-architecture child manifests embedded inside the index are stored at `oci-manifests/<child-digest>` and have their own `artifacts` rows that can be soft-deleted, and they are not directly tagged. If those child artifact rows get soft-deleted, the new guard does not see them: only the index digest appears in `oci_tags`, not the child digests it references.

A user pulling a multi-arch image after GC would get `MANIFEST_UNKNOWN` for the arch-specific layers.

Options: also check whether the storage key's digest appears as a member digest inside any live index manifest (requires JSON-body inspection), or insert a synthetic `oci_tags` row for each child digest at push time so the existing guard catches them.

### 2. No transaction around the orphan check and delete loop

The orphan list is computed in one `SELECT`, then each storage key is deleted from disk and the DB in a per-key loop without a wrapping transaction. A concurrent tag delete can remove the `oci_tags` row that was protecting key A between the `SELECT` and the storage-delete pass for key B in the same GC run.

Wrapping each per-key (storage delete + DB hard-delete) in a transaction with a `SELECT ... FOR UPDATE` re-check of the guard conditions would close this. Low probability in practice but real.

### 3. Regression tests only cover the dry-run path

Both new tests call `run_gc(true)`. Dry-run exits before the storage and DB delete path runs, so the tests verify only the SQL filter, not the full execution. A non-dry-run variant using the existing `MockStorageBackend` in the test module, asserting the storage file is not deleted, would give end-to-end coverage of the actual code path you're fixing.

brandonrc:
Pushed `b3afbeb` to your branch adding a live (non-dry-run) variant of the manifest-survival test. The two dry-run tests prove the SQL filter is right; the new test writes a real manifest file via the filesystem backend, runs `run_gc(false)` so the entire delete path executes, and asserts the file is still on disk afterward. Without it, a future refactor of the delete loop could regress the fix and the dry-run tests would still pass.

On the other two concerns from my earlier review:

1. **Multi-arch child manifests** — this needs an architectural choice (eager `oci_manifest_refs` table at push time, or JSON inspection at GC time). Filed as #1179 so it doesn't expand the scope of this PR.

2. **TOCTOU race between orphan SELECT and per-key delete** — real but small window; the fix requires extracting the orphan condition and running it inside an in-transaction re-check. Filed as #1180 for the same scope reasons.

The fix in this PR closes the dominant data-loss path (tag-pointed manifests + named blobs). Ready from my side once CI is green.

brandonrc (approved):
Approving. The original fix and the live-run test together cover the dominant data-loss path. Multi-arch and TOCTOU follow-ups tracked as #1179 and #1180.

Thanks for digging in on this one — the GC deleting referenced data is the kind of regression that erodes trust in the whole system, glad it's getting fixed.

Issue #593, opened 2026-03-28
Docker /v2/token endpoint does not accept API tokens as Basic Auth password


andrlange:
## Description

The `/v2/token` endpoint only accepts username + password (bcrypt-verified) via Basic Auth. It does not accept API tokens as the password field. This prevents service accounts from using `docker login` with their API tokens.

In the current `token()` handler in `oci_v2.rs`, when Basic Auth credentials are provided, only `auth_service.authenticate()` is called (which checks bcrypt password hashes). If this fails, there is no fallback to `auth_service.validate_api_token()`.

## Steps to Reproduce

1. Create a service account with an API token (scopes: `read:artifacts`)
2. Attempt: `docker login <registry> -u <svc-account> -p <api-token>`
3. Login fails with "invalid username or password"

## Expected Behavior

The `/v2/token` endpoint should try `validate_api_token()` when password authentication fails, allowing service accounts to authenticate with their API tokens for Docker operations.

## Workaround

In `oci_v2.rs`, modify the `token()` function to fall back to `validate_api_token(&credentials.1)` when `authenticate()` returns an error:

```rust
let user = match auth_service.authenticate(&credentials.0, &credentials.1).await {
    Ok((_user, tokens)) => {
        // password auth succeeded — return JWT directly
        return build_token_response(tokens);
    }
    Err(_) => {
        // Fall back to API token in the password field
        match auth_service.validate_api_token(&credentials.1).await {
            Ok(validation) => validation.user,
            Err(_) => return oci_error(UNAUTHORIZED, "invalid username or password"),
        }
    }
};
let tokens = auth_service.generate_tokens(&user)?;
build_token_response(tokens)
```

## Use Case

This is essential for CI/CD pipelines and automated tooling (e.g., `skopeo`, `crane`, `docker`) that need to authenticate against the registry using machine credentials (API tokens) rather than human passwords.

## Affected Version

v1.1.0-rc.10

PR #1145, opened 2026-05-10
fix(auth): try API-token validation before authenticate on /v2/token

inspired-geek:
## Summary

`docker login -p $API_TOKEN` (the standard CI/CD flow) currently locks out service accounts after `account_lockout_threshold` builds. The `/v2/token` handler calls `auth_service::authenticate(username, api_token)` first; that function bcrypt-compares the API token against the user's password hash (always a mismatch) and bumps `failed_login_attempts`.

`validate_api_token` has no failure-counter side effect — it bcrypt-compares against `api_tokens.token_hash` directly, with a dummy constant-time hash on the miss path. Trying it first preserves the lockout counter for real password attempts while keeping bcrypt as the fallback for non-API-token credentials.

This is a residual regression from the fix for [#593](https://github.com/artifact-keeper/artifact-keeper/issues/593) (the original "Docker /v2/token endpoint does not accept API tokens as Basic Auth password" feature). That fix added the `validate_api_token` fallback _after_ `authenticate`, which works functionally — service accounts can now log in — but the bcrypt-mismatch in `authenticate` runs first and bumps the lockout counter on every CI build.

## Change

`backend/src/api/handlers/oci_v2.rs::token`: swap the two arms of the credential-resolution `match` so `validate_api_token` is tried before `authenticate`. The `authenticated_via_api_token` flag and the downstream TOTP guard are wired the same way — API tokens still bypass TOTP per the existing comment, and the password path still triggers the TOTP block.

- [x] DB-backed unit test in new module `token_lockout_regression_tests` (in `oci_v2.rs`):
  - inserts a user with a real bcrypt password hash via `test_db_helpers`
  - generates an API token via `AuthService::generate_api_token`
  - hits `GET /v2/token` with `Authorization: Basic <user>:<api-token>` via `oneshot()`
  - asserts response is a real JWT (not anonymous) AND `users.failed_login_attempts == 0`

  Pre-fix the counter is 1; post-fix 0.

## Test Checklist
- [ ] Integration tests added/updated (DB-backed unit test covers the bug end-to-end)
- [ ] E2E tests added/updated
- [x] Manually tested locally (cargo test against running Postgres)
- [x] No regressions in existing tests (8994 / 0 / 2 passed/failed/ignored)

## API Changes
- [x] N/A — same route, same response shape, only the order of credential validation changes


brandonrc:
Thanks for the fix. The reorder is correct: `validate_api_token` has no failure-counter side effect while `authenticate` does, so trying API-token validation first stops every CI/build that uses an API token in the password field from incrementing `failed_login_attempts` toward lockout. The regression test pins the right behavior.

Two small requests before merge:

### 1. Add a sibling test for the password-fallback path

Only the API-token-first success path is currently pinned. Please add a paired test that uses `Basic <real_user>:<real_password>` after the reorder and asserts a JWT is issued. The constant-time bcrypt pad in `validate_api_token` makes the fallback work today, but this exact path is what would regress if someone later "optimizes" the dummy-hash code without realizing the OCI handler depends on it.

### 2. Refresh the doc-comments

`TokenForm` (oci_v2.rs:137-148) and `extract_form_credentials` (oci_v2.rs:150-184) still describe the legacy order. Worth a quick pass while you're in this area so the comments don't drift.

### Out of scope but worth noting

`authenticate_oci` at oci_v2.rs:241-265 is still password-first and has the same `failed_login_attempts` footgun for the non-`/v2/token` OCI verbs (manifest GETs, blob heads, etc.) when callers use Basic auth. That's a separate fix, but if you want to file a follow-up issue or extend this PR to cover it, the bug is identical in shape.

brandonrc:
Pushed `738e5ba` to your branch adding the password-fallback regression test (`password_basic_auth_falls_through_to_authenticate`). It pins the standard `docker login` flow after the reorder so a future refactor of `validate_api_token`'s constant-time bcrypt pad can't silently regress it.

On the doc-comment refresh I mentioned: re-reading `TokenForm` (oci_v2.rs:137-148) and `extract_form_credentials` (lines 150-184), they describe the form-extraction layer (which fields are read, what content-type is accepted), not the validation order. The new in-function comment at lines 800-810 already documents the API-token-first ordering correctly. So I didn't touch the upper-level doc-comments. If you read them differently, easy to revisit.

The sibling-flow concern in `authenticate_oci` (line 226-265) for non-`/v2/token` OCI verbs is still worth a separate fix, but that's a different PR.

brandonrc (approved):
Approving. The reorder is correct and the pair of regression tests now covers both arms (API-token-first success path and password-fallback). The doc-comments at the form-extraction layer aren't actually stale on re-read, so leaving them as-is.

Thanks for catching the lockout footgun — the CI service-account-locks-itself-out failure mode is exactly the kind of thing that's invisible until it bites in production.

inspired-geek:
Thanks for the review and for pushing `738e5ba` with the password-fallback test. Agreed on leaving the form-extraction doc-comments alone — re-reading them they describe the extraction layer, not the validation order.

Filed #1195 for the sibling `authenticate_oci` footgun you flagged (same fix pattern, different code path). Will open a separate PR for it rather than bundling into anything in flight.

Issue #536, opened 2026-03-23
OCI v2 token endpoint bypasses TOTP 2FA


brandonrc:
## Problem

`oci_v2.rs:359-371` calls `authenticate()` and issues tokens without checking `totp_enabled` on the user. This means Docker login (`docker login`) bypasses two-factor authentication entirely.

Users who enable TOTP expect all authentication flows to require the second factor. OCI/Docker is often the most privileged access path (pushing container images).

## Suggested Fix

After `authenticate()`, check `user.totp_enabled`. If true, return 401 UNAUTHORIZED with a message indicating that TOTP-enabled accounts cannot use Docker basic auth directly, and should use an API key or personal access token instead.

## Source

Security audit, 2026-03-23.

Issue #505, opened 2026-03-21
Sessions not invalidated after password change


brandonrc:
## Summary

After an admin changes a user's password, existing JWT tokens for that user remain valid. The old password also continues to work for login. This means a compromised account cannot be secured by changing the password.

## Evidence

Release gate run: https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/23372998677

- Old password still accepted after change: HTTP 200 (expected 401)
- Login with new password failed

## Test Reference

`artifact-keeper-test/tests/auth/test-session-invalidation.sh`

## Impact

High. Compromised accounts cannot be recovered through password changes. Attacker retains access indefinitely.

## Fix

Password change should invalidate all existing JWT tokens for the user. Options:
1. Increment a per-user token version counter and include it in JWT claims
2. Track a "password_changed_at" timestamp and reject tokens issued before it
3. Add the user's tokens to a revocation list on password change

PR #1146, opened 2026-05-10
fix(auth): invalidate user JWTs on TOTP enable/disable

inspired-geek:
## Summary

`enable_totp` and `disable_totp` UPDATE the `users.totp_enabled` column but do not bump the credential-invalidation timestamp tracked by `auth_service::invalidate_user_tokens`. A refresh token issued *before* TOTP was enabled therefore stays valid until natural expiry (`jwt_refresh_token_expiry_days`, default 7), letting its bearer swap it for a fresh access token via the OAuth refresh-grant — bypassing the new factor.

Symmetric on disable: a token issued under stricter (TOTP-required) policy keeps working silently when the user removes 2FA, even though the trust assumptions changed.

### Relation to closed issues

- **Continues [#505](https://github.com/artifact-keeper/artifact-keeper/issues/505) ("Sessions not invalidated after password change", security/p1).** That issue established the `invalidate_user_tokens` pattern: any credential change must bump the per-user invalidation timestamp so JWTs issued before fail `is_token_invalidated`. The fix wired it into the password-change paths but missed TOTP toggle, which is the same kind of credential change.
- **Closes the residual gap left by [#536](https://github.com/artifact-keeper/artifact-keeper/issues/536) ("OCI v2 token endpoint bypasses TOTP 2FA", security/p1).** #536 added `if user.totp_enabled && !authenticated_via_api_token { 401 }` to the password-grant path. The check is correct *for new logins*, but a refresh token issued before TOTP was enabled survives the policy change and can mint a fresh access JWT without ever presenting a TOTP code. After this PR the pre-TOTP refresh token fails `is_token_invalidated` instead.

## Change

**`backend/src/api/handlers/totp.rs`**:

- `enable_totp` and `disable_totp` now call `invalidate_user_tokens(auth.user_id)` after the UPDATE. The function is already exported and is the same hook the user-deactivation, user-deletion, and password-reset paths use to invalidate prior sessions.
- Import widened to `use crate::services::auth_service::{invalidate_user_tokens, AuthService};`.

The invalidation marks the user with the current timestamp; `auth_service::refresh_tokens` and `validate_token` both call `is_token_invalidated(user_id, claims.iat)` and reject any token whose `iat` predates the marker.

- [x] Two DB-backed unit tests in new module `totp_token_invalidation_regression_tests`:
  - `enable_totp_invalidates_pre_change_user_tokens` — fresh user, TOTP secret seeded, `enable_totp` invoked with a `totp_rs::TOTP::generate_current` code; asserts `is_token_invalidated(user_id, now-60)` is true after the call.
  - `disable_totp_invalidates_pre_change_user_tokens` — fresh user with `totp_enabled = true`, real bcrypt password hash, `disable_totp` invoked with password + current code; same invariant.

  Both fail on `main`, both pass on this branch.

## Test Checklist
- [ ] Integration tests added/updated (DB-backed unit tests cover the call sites end-to-end)
- [ ] E2E tests added/updated
- [x] Manually tested locally (8995 / 0 / 2 passed/failed/ignored on a clean DB)

## API Changes
- [x] N/A — same routes, same response shapes; behavior strictly tightens (existing sessions are now invalidated when 2FA is toggled, no schema change)


brandonrc (approved):
Thanks for getting ahead of this. The TOTP-invalidation gap was something I was about to flag on #1166 (in fact, that PR's `handle_refresh_grant` doc-comment claims this invariant and its test simulates the wiring). Your fix here makes the claim real instead of just simulated.

Both call sites covered, both DB-backed regression tests are well-constructed, and the `pre_change_iat` offset correctly handles the 1-second resolution of `is_token_invalidated`. The per-process scope of the invalidation map is a known limitation properly out of scope here (tracked in #1173).

Appreciate the work you're putting into the project.

PR #1148, opened 2026-05-10
fix(apt): resolve stale Packages cache on upstream Release change and Virtual repo dists resolution (#1147)

### LINKED ISSUE #1147: APT: Remote proxy serves stale Packages after upstream Release changes; Virtual repos cannot resolve dists metadata
### Description

Two related issues with APT Remote proxy caching:

1. **Cache staleness:** When an upstream APT repository publishes new packages, the cached Release/InRelease file eventually expires and gets re-fetched, but the Packages/Packages.gz/Packages.xz caches remain valid per their own TTL. APT clients receive a fresh Release (with new SHA-256 hashes) alongside stale Packages indices, resulting in hash mismatch errors.

2. **Virtual repos with Remote members:** Virtual repositories that contain Remote APT members cannot serve dists metadata (Release, InRelease, Packages, etc.). Requests fall through to local generation which produces empty/incorrect responses, making `apt-get update` against a Virtual APT repo fail. This blocks the use case of placing multiple Remote APT mirrors in a single Virtual repo for priority-based fallback (e.g. a third-party mirror as primary, upstream debian.org as fallback), giving clients a single apt source URL that transparently tries alternatives when the higher-priority mirror is unavailable.

### Environment Information

- Deployment type: Any
- Operating System: Any
- Architecture: Any
- Database: PostgreSQL (any supported version)
- Reverse Proxy: N/A
- Network setup: N/A

### Backend Version

1.1.9

### Package Format

Debian / APT

### Expected Behavior

1. When an upstream Release/InRelease file changes, all related cached metadata (Packages, Packages.gz, Packages.xz, i18n, dep11, Contents, etc.) under that distribution should be invalidated so subsequent requests fetch fresh copies, maintaining hash consistency.

2. Virtual APT repos should resolve dists metadata from their Remote members by priority order, enabling multi-mirror fallback — e.g. a Virtual repo with a third-party mirror (priority 1) and upstream debian.org (priority 2), where clients get a single apt source URL that transparently falls back if the primary mirror is unavailable.

### Observed Behavior

1. After upstream publishes new packages, `apt-get update` against the proxy fails with:
   ```
   E: Failed to fetch .../Packages.gz  Hash Sum mismatch
   ```

2. Virtual APT repos with Remote members: `apt-get update` returns empty or invalid metadata, regardless of how many Remote members are configured.

### Steps to Reproduce

**Cache staleness:**
1. Create a Remote Debian repo proxying an upstream that publishes frequently
2. Run `apt-get update` successfully (caches Release + Packages)
3. Wait for upstream to publish new packages (Release checksum changes)
4. Wait for Release cache TTL to expire but Packages cache is still valid
5. Run `apt-get update` — gets fresh Release but stale Packages — hash mismatch

**Virtual repo fallback:**
1. Create two Remote Debian repos (e.g. a third-party mirror + debian.org)
2. Create a Virtual repo with both as members (third-party mirror at priority 1, debian.org at priority 2)
3. Configure apt sources to point at the Virtual repo
4. Run `apt-get update` — fails because dists metadata cannot be resolved from members

### Additional Context

Fix is implemented in branch `feat/apt-virtual-proxy-cache-invalidation`. Changes:
- New `dists_detecting_change()` method compares upstream SHA-256 before/after fetch
- `ProxyService::invalidate_dist_packages_cache()` deletes all non-Release cached entries under a distribution when change is detected
- Release and InRelease handlers use `dists_detecting_change` instead of plain `dists`
- `DebianProxy::dists()` and `dists_proxy_catchall` now resolve Virtual repos through their members via priority-based iteration

Dreamacro:
## Summary

Fixes #1147.

When a Remote APT proxy's upstream publishes new packages, the Release/InRelease
cache eventually expires and gets re-fetched, but the Packages indices retain
their own (still-valid) TTL. APT clients then receive a fresh Release with
SHA-256 hashes that don't match the stale cached Packages — `apt-get update`
fails with "Hash Sum mismatch".

Additionally, Virtual repositories containing Remote APT members could not serve
dists metadata at all (Release, InRelease, Packages, etc.) because
`DebianProxy::dists()` short-circuited for non-Remote repos, and
`dists_proxy_catchall` returned 404. This blocked the multi-mirror fallback use
case (multiple Remote mirrors in a Virtual repo with priority-based resolution).

Changes:
- **`DebianProxy::dists()`** — add Virtual branch that resolves metadata through
  members via `resolve_virtual_download`.
- **`dists_proxy_catchall`** — same Virtual support for non-Packages dists paths
  (i18n, dep11, Contents, etc.).
- **`dists_detecting_change()`** — new method that compares the SHA-256 of the
  newly-fetched Release/InRelease against the previously cached checksum; when
  they differ, invalidates all non-Release cached metadata for that distribution.
- **`ProxyService::fetch_artifact_detecting_change()`** — cache-aware fetch that
  returns a `changed` boolean alongside the content.
- **`ProxyService::invalidate_dist_packages_cache()`** — lists all keys under
  `proxy-cache/{repo}/dists/{dist}/` and deletes everything except
  Release/InRelease/Release.gpg/gpg-key.asc.
- Release and InRelease handlers now call `dists_detecting_change` instead of
  plain `dists`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/debian.rs
+++ b/backend/src/api/handlers/debian.rs
@@ -397,26 +397,132 @@ impl<'a> DebianProxy<'a> {
         content_type: &'static str,
         repo: &RepoInfo,
     ) -> Result<(), Response> {
-        if repo.repo_type != RepositoryType::Remote {
-            return Ok(());
-        }
-        let (upstream_url, proxy) = match (&repo.upstream_url, &self.state.proxy_service) {
-            (Some(u), Some(p)) => (u, p),
-            _ => return Ok(()),
-        };
-        let upstream_path = format!("dists/{}/{}", self.distribution, suffix);
-        let (content, upstream_ct) =
-            proxy_helpers::proxy_fetch(proxy, repo.id, self.repo_key, upstream_url, &upstream_path)
-                .await?;
-        Err(Response::builder()
-            .status(StatusCode::OK)
-            .header(
-                CONTENT_TYPE,
-                upstream_ct.unwrap_or_else(|| content_type.to_string()),
+        if repo.repo_type == RepositoryType::Remote {
+            let (upstream_url, proxy) = match (&repo.upstream_url, &self.state.proxy_service) {
+                (Some(u), Some(p)) => (u, p),
+                _ => return Ok(()),
+            };
+            let upstream_path = format!("dists/{}/{}", self.distribution, suffix);
+            let (content, upstream_ct) = proxy_helpers::proxy_fetch(
+                proxy,
+                repo.id,
+                self.repo_key,
+                upstream_url,
+                &upstream_path,
             )
-            .header(CONTENT_LENGTH, content.len().to_string())
-            .body(Body::from(content))
-            .unwrap())
+            .await?;
+            Err(Response::builder()
+                .status(StatusCode::OK)
+                .header(
+                    CONTENT_TYPE,
+                    upstream_ct.unwrap_or_else(|| content_type.to_string()),
+                )
+                .header(CONTENT_LENGTH, content.len().to_string())
+                .body(Body::from(content))
+                .unwrap())
+        } else if repo.repo_type == RepositoryType::Virtual {
+            let upstream_path = format!("dists/{}/{}", self.distribution, suffix);
+            let state = self.state;
+            let (content, upstream_ct) = proxy_helpers::resolve_virtual_download(
+                &state.db,
+                state.proxy_service.as_deref(),
+                repo.id,
+                &upstream_path,
+                |member_id, location| {
+                    let db = state.db.clone();
+                    let state = state.clone();
+                    let path = upstream_path.clone();
+                    async move {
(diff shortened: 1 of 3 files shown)


brandonrc:
@Dreamacro  thanks for the PR. I am working on a bunch of backlog stuff but will get to reviewing this soon. 

Appreciate the time and effot.

Dreamacro:
@brandonrc I noticed that on `main`, the tee stream migration converted Remote proxy fetches to streaming (`proxy_fetch_streaming`), but `resolve_virtual_download` still uses the buffered path — when a Virtual repo resolves through a Remote member, the body is still fully buffered before serving to the client.

This PR adds Virtual repo dists resolution for APT, which also goes through the buffered `resolve_virtual_download`. Should this be updated to use streaming as well, or is that a separate follow-up? I'm happy to wait and rebase after the streaming migration covers the virtual resolution path if that's planned.

Issue #1021, opened 2026-05-04
Backend: WWW-Authenticate realm must advertise external hostname (Docker registry token retry)


brandonrc:
## Severity

Deployment-correctness bug. Surfaced during round-3 review of artifact-keeper-web#1007 (`fix/1007-docker-proxy`).

## The problem

When a Docker client hits `/v2/`, the backend (correctly) responds with:

```
401 Unauthorized
WWW-Authenticate: Bearer realm="https://<HOST>/v2/token",service="<service>"
```

The Docker client then re-issues the request to that `realm` URL to fetch a bearer token, then retries `/v2/` with the token. **The realm URL must be reachable by the client, externally.**

If the backend builds the realm URL from internal-cluster context (e.g., `http://artifact-keeper-backend:8080/v2/token` from the in-cluster Service name, or from `request.uri().authority()` when the ingress doesn't set `X-Forwarded-Host`), the docker client outside the cluster cannot reach it, and `docker login` still fails — even after artifact-keeper-web#1007 is merged.

## What to check

1. Where in `backend/src/api/handlers/` is the docker `WWW-Authenticate` header constructed?
2. Does it derive the realm URL from:
   - A configurable env var (`AK_PUBLIC_HOSTNAME`, `AK_EXTERNAL_URL`)?
   - The incoming `Host` header (vulnerable to host-header spoofing)?
   - `X-Forwarded-Host` / `X-Forwarded-Proto`?
   - A hardcoded internal address?
3. If the realm URL can be misconfigured, add validation at startup and log a clear warning.
4. Document the required ingress / proxy setup in `artifact-keeper-site` (link the `X-Forwarded-Host` and external-URL config requirements).

## Acceptance criteria

- Realm URL is consistently externally reachable when artifact-keeper is deployed behind an ingress that sets `X-Forwarded-*` headers (the standard Helm chart setup).
- Realm URL falls back gracefully (or fails loudly with a clear error) when no external hostname is configured.
- An integration test asserts the response of `GET /v2/` includes a `WWW-Authenticate` header whose `realm` matches the expected external URL pattern.
- Documented in the deployment guide.

## Refs

- Discovered during artifact-keeper-web#1007 round-3 review.
- Related: artifact-keeper-web#TODO (e2e header-forwarding test coverage — separate issue).

PR #1149, opened 2026-05-10
fix(oci): close stale-JWT bypass and sanitize WWW-Authenticate

inspired-geek:
## Summary

Two independent OCI token-endpoint hardening fixes that share a file but address different bugs.

### 1. Bearer-JWT swap path now filters on `is_active = true`

The Bearer-token branch of `oci_v2::token` selects the user by id without an `is_active` filter, so a deactivated user with an unexpired JWT in hand can still mint a fresh OCI access token via the JWT-to-OCI swap. `auth_service::authenticate`, `refresh_tokens`, and `validate_api_token` already filter on `is_active = true`; this aligns the OCI swap with that policy and closes the bypass.

### 2. `WWW-Authenticate` header is now CRLF/header-injection-proof

`pull_scope` / `push_scope` interpolate the URL-decoded `image_name` path parameter into the scope value, which flows into a `Bearer realm="…",scope="…"` `WWW-Authenticate` header without sanitization. A path containing `%0D%0A` lets a client-controlled URL inject arbitrary follow-on headers into the 401 response.

New helper `auth_challenge_quoted_value` emits an RFC 7230 §3.2.6-conformant quoted-string body:
- `"` and `\` get the standard `quoted-pair` backslash escape.
- HTAB and printable ASCII pass through verbatim.
- CR/LF/NUL, other control chars, and `obs-text` (≥ 0x80) are percent-encoded byte-by-byte. The whitelist also keeps the result valid for `HeaderValue::from_str` (ASCII-visible plus HTAB).

`www_authenticate_header` runs every quoted value (`realm` and `scope`) through it.

### Relation to open issues

Orthogonal to open [#1021](https://github.com/artifact-keeper/artifact-keeper/issues/1021) ("WWW-Authenticate realm must advertise external hostname"). Both touch `www_authenticate_header`, but #1021 is about the _source_ of the realm hostname (X-Forwarded-Host vs in-cluster Service name vs config env var); this PR is about how the values are _encoded_ once they have been chosen. The two fixes compose — #1021's external-URL realm still needs the quoted-string sanitization this PR adds.

## Change

**`backend/src/api/handlers/oci_v2.rs`**:
- `token()` Bearer branch SQL: `... FROM users WHERE id = $1 AND is_active = true` (`.sqlx` cache regenerated).
- New `auth_challenge_quoted_value` helper with the percent-encoding/escape rules above.
- `www_authenticate_header` rewritten to call it on `realm` and `scope`.

- [x] DB-backed unit test `deactivated_user_cannot_swap_bearer_jwt_for_oci_token`: signs a JWT for a fresh user via `AuthService::generate_tokens`, flips `is_active=false`, hits `GET /v2/token` with that bearer, asserts 401. Pre-fix returns 200 with a fresh access token.
- [x] Four unit tests for `auth_challenge_quoted_value`:
  - `test_www_authenticate_header_sanitizes_crlf_scope` — CRLF percent-encoded, `HeaderValue::from_str` passes
  - `test_www_authenticate_header_preserves_htab` — HTAB pass-through, NOT percent-encoded
  - `test_www_authenticate_header_percent_encodes_non_ascii_scope` — UTF-8 bytes percent-encoded; result is pure ASCII
  - `test_auth_challenge_quoted_value_escapes_quote_and_backslash` — quoted-pair on `"` / `\`

## Test Checklist
- [x] Unit tests added/updated (1 DB-backed regression + 4 sync helpers)
- [ ] Integration tests added/updated
- [ ] E2E tests added/updated
- [x] Manually tested locally (8998 / 0 / 2 passed/failed/ignored on a clean DB; new code coverage 98%)

## API Changes
- [x] N/A — same routes, same response shapes; the `WWW-Authenticate` header is byte-different only for inputs that previously contained CR/LF/NUL/non-ASCII (no real client sends those today).


brandonrc (approved):
Nice catches on both. The stale-JWT bypass through the Bearer-swap path is exactly the kind of thing that hides in plain sight: the swap-path is what mints a new long-lived credential, so it absolutely has to re-verify against the source of truth. Adding `AND is_active = true` brings that path into alignment with `refresh_tokens` and `authenticate`, and the regression test (flipping `is_active=false` directly via SQL to bypass `invalidate_user_tokens`, then asserting 401) pins exactly the right invariant.

The WWW-Authenticate sanitization is thorough. Covering both `realm` (derived from request_host, attacker-controllable via Host/X-Forwarded-Host) and `scope` closes the obvious vectors, and the CRLF/HTAB/non-ASCII/quote-and-backslash tests give good shape coverage.

Thanks for the careful work.

PR #1165, opened 2026-05-11
test(oci): repair PR #642 catalog/tags tests and correct search route

inspired-geek:
## Summary

Repairs OCI catalog/tags integration tests originally added in PR #642 that have been silently broken since merge, plus one stale route in the artifact search test.

### `_catalog` / `tags/list` tests upload to the wrong endpoint

PR #642 added five tests that upload manifests via the generic `/api/v1/repositories/{key}/artifacts/{path}` endpoint, which only writes to the `artifacts` table. `tags/list` and `_catalog` read from `oci_tags`, which only the OCI manifest handler (`PUT /v2/{name}/manifests/{ref}`) populates. The tests therefore couldn't observe the data they uploaded.

CI never noticed because they're `#[ignore]`-gated and `cargo test --workspace --test '*'` is invoked without `--ignored`.

### `test_search_artifacts` hits an unregistered route

`/api/v1/search/artifacts` was never registered. The real route is `/api/v1/search/quick` (see `backend/src/api/handlers/search.rs::router`). The response wrapper is `{"results": [...]}`, not `{"items": [...]}`.

### Relation to open issues

Concrete instance of [#1031](https://github.com/artifact-keeper/artifact-keeper/issues/1031) ("CI integration-test convention: drop `#[ignore]` or pass `--ignored` consistently"). This PR fixes the broken-test side of the problem — once `#[ignore]` is dropped or CI passes `--ignored`, these five tests will actually catch OCI catalog/tags regressions. It does **not** change the broader CI convention; that's still an open question for #1031 to settle.

## Change

**`backend/tests/integration_tests.rs`**:

- New `TestServer::oci_push_manifest` helper that PUTs through `/v2/{name}/manifests/{ref}`, populating `oci_tags` the way an actual `docker push` does.
- Rewrite `test_oci_tags_list_after_manifest_push`, `test_oci_tags_list_pagination`, `test_oci_tags_list_n_zero`, `test_oci_catalog`, `test_oci_catalog_includes_private_repos_for_authenticated_non_admin_user` to use the helper.
- `test_search_artifacts`: `/api/v1/search/quick`, `results[]` (with `items[]` fallback for forward compat).
- Per-run username in `test_oci_catalog_includes_private_repos_for_authenticated_non_admin_user` uses `Uuid::new_v4().simple()` so reruns against a long-lived dev DB don't 409-collide.

- [x] N/A — this PR IS a regression-test improvement. The change is in test code, not production code.

## Test Checklist
- [ ] Unit tests added/updated (N/A — integration only)
- [ ] E2E tests added/updated
- [x] Manually tested locally — 29 affected `#[ignore]` integration tests pass on a fresh clean DB against an `origin/main` backend

## API Changes
- [x] N/A — test code only


brandonrc:
Thanks for digging into this. Confirming both findings: PR #642's original tests pushed manifests via the generic `/api/v1/repositories/{key}/artifacts/{path}` endpoint, which writes only to the `artifacts` table, while `tags/list` and `_catalog` read from `oci_tags` (populated only by `PUT /v2/{name}/manifests/{ref}`). The original tests were asserting against empty arrays and passing vacuously. The fix is architecturally correct.

`/api/v1/search/artifacts` is also confirmed unregistered in `backend/src/api/handlers/search.rs`; the route was always wrong. Good catch.

Two things worth discussing before merge:

### 1. The tests remain `#[ignore]`-gated

The repair makes the assertions correct, but the tests still won't run in CI (consistent with #1031 / the current convention). That means the repair has no enforcement value until #1031 is resolved. Two options:

- Drop `#[ignore]` from at least these five now-correct tests so they exercise on every push, and document why they're the exception.
- Leave them as-is and note them as candidates in #1031 so they're un-ignored as part of that broader cleanup.

Either is fine. I'd lean toward un-ignoring now since the tests are known-good and the value of catching future regressions is higher than the convention.

### 2. The search test accepts two response shapes

`test_search_artifacts` now passes if the body has either `results[]` or `items[]`. The correct shape is `results`, so the `items[]` branch is dead code that would silently accept a regression. Suggest dropping it: if the shape ever drifts back, that should fail, not pass.

Neither is blocking. (2) is a one-line fix.

brandonrc:
Pushed `9b94757` to your branch addressing the `items[]` fallback. Kept the un-ignore question as-is: ALL integration tests in this file are `#[ignore]`-gated for the same "requires a running HTTP server" reason, so un-ignoring the five corrected tests without addressing that broader convention would just make them fail in CI. That's the right call for #1031, not this PR.

The PR is in good shape from my side once CI is green.

brandonrc (approved):
Approving. Test repair is correct, the search-route fix is a real bug catch, and the strict-results-shape assertion is the right call. The broader `#[ignore]` convention is still tracked in #1031.

Thanks for the careful test archeology.

inspired-geek:
Thanks for the review and for pushing the `items[]` fix directly. Agreed on deferring the `#[ignore]` decision to #1031 — un-ignoring just these five in isolation wouldn't make sense given the rest of the file has the same constraint.

Issue #894, opened 2026-04-26
POST /v2/token ignores form-encoded credentials, breaking docker push (OAuth2 password grant)


DiverOfDark:
## Summary

`docker push` to a private (or any) repository fails with `unauthorized` because `POST /v2/token` ignores form-encoded credentials in the request body and returns the anonymous token. Docker's [OAuth2 token endpoint](https://distribution.github.io/distribution/spec/auth/oauth/) flow uses `application/x-www-form-urlencoded` body parameters (`grant_type=password&username=...&password=...&service=...&scope=...`), but the handler only reads credentials from the HTTP Basic Auth header.

## Version

- Backend: `ghcr.io/artifact-keeper/artifact-keeper-backend:1.1.6`
- Bug also present on `main` as of writing.

## Reproduction

```bash
# Returns {"token":"anonymous", ...} despite valid creds in the form body:
curl -s -X POST \
  -H "Content-Type: application/x-www-form-urlencoded" \
  --data-urlencode "grant_type=password" \
  --data-urlencode "username=<svc-account>" \
  --data-urlencode "password=<api-token>" \
  --data-urlencode "service=artifact-keeper" \
  --data-urlencode "scope=repository:<repo>/<image>:push,pull" \
  https://<host>/v2/token

# {"token":"anonymous","access_token":"anonymous","expires_in":900,...}
```

For comparison, GET with Basic Auth — and POST with Basic Auth header — work fine and return a valid JWT:

```bash
curl -u 'user:token' "https://<host>/v2/token?service=artifact-keeper&scope=..."
# {"token":"<JWT>","access_token":"<JWT>",...}
```

## Expected

Per the OAuth2 password grant flow used by Docker, the server should accept credentials supplied in the request body when `Content-Type: application/x-www-form-urlencoded`, validate them, and return a scoped access token.

## Actual

The handler at [`backend/src/api/handlers/oci_v2.rs`](backend/src/api/handlers/oci_v2.rs) (`async fn token`, around line 316 in v1.1.6) only calls `extract_basic_credentials(&headers)`. When that returns `None`, it falls through to the anonymous-token branch — even though valid `username`/`password` are present in the form body. Docker then uses the `"anonymous"` bearer for the actual `POST /v2/<repo>/<image>/blobs/uploads/`, which correctly returns 401, and the push fails.

## Effect on users

- `docker login` returns "Login Succeeded" (the daemon caches creds and validates lazily / via the GET path).
[...]

PR #1166, opened 2026-05-11
feat(oci): OAuth2 refresh-token grant and access_type=offline support

### LINKED ISSUE #2244: feat(oci): OAuth2 refresh-token grant + access-token endpoint
Tracking issue for the OCI Docker-registry OAuth2 refresh-token grant + access-token endpoint implemented in #1166.

Adds `grant_type=refresh_token` support and `access_type=offline` issuance to the OCI `/v2/token` flow so registry clients can refresh without re-authenticating. Review confirmed no scope-escalation on refresh, no IDOR, and correct API-token refresh-suppression.

Blocker being fixed in #1166: the default refresh-grant path must return the rotated refresh token (single-use rotation #1174 consumes the presented one) so a second refresh doesn't trip replay detection and revoke the token family. Includes a two-call regression test.

Relates to the auth/SSO correctness epic #1617.

inspired-geek:
Closes #2244

## Summary

Implements the two remaining flows from the [Docker Distribution OAuth2 spec](https://distribution.github.io/distribution/spec/auth/oauth/) that `/v2/token` did not previously support:

- **`grant_type=refresh_token`** — exchange a long-lived refresh token for a fresh access token without re-submitting credentials.
- **`access_type=offline`** — opt-in for the response to include a `refresh_token` field (default is `online` per the spec, returning only a short-lived access token).

The password-grant + `access_type=offline` flow is what `docker login -u alice -p $PWD` + a long-lived session expects. Without this PR a user has to re-authenticate every ~30 minutes; modern Docker daemons request `access_type=offline` automatically and previously got back a response with no refresh token.

A security-critical rule is layered on top: **API-token-as-password authentication never emits a refresh token, even when `access_type=offline` is requested.** The full justification is documented inline in `offline_refresh_token`'s doc-comment; the short version follows.

## Why API-token authentication suppresses the refresh token

Issuing a refresh-token JWT under an API-token-authenticated request would create a parallel long-lived credential that:

- is signed against `JWT_SECRET` (not stored in any DB row);
- is NOT tied to the `api_tokens` row;
- therefore is **not reachable by `revoke_api_token`**.

`auth_service::refresh_tokens` does not blacklist consumed refresh tokens either, so this parallel credential would survive `revoke_api_token` for the full `jwt_refresh_token_expiry_days` window (default 7). An attacker who briefly captured an API token would still have a week of access after the defender thought they'd closed the gap.

### Industry precedent for the suppression

- **[RFC 6749 §4.4.3](https://datatracker.ietf.org/doc/html/rfc6749#section-4.4.3)** (`client_credentials` grant): _"A refresh token SHOULD NOT be included."_ API-token-as-password is semantically a machine-credential flow — exactly what `client_credentials` covers. The spec explicitly tells us not to mint refresh tokens for machine clients.
- **[JFrog Artifactory](https://jfrog.com/help/r/jfrog-artifactory-documentation/authenticate-docker-via-oauth)** deprecated their API-key-as-password path entirely. [Their EOL notice](https://docs.jfrog.com/user-management/docs/api-key) (Q4 2024) calls out the exact problem we're closing: _"API Keys don't have lifecycle management features […] single user can have single active API Key — if revoked, revoked for all clients."_ JFrog's replacement (reference tokens / identity tokens) is designed without a parallel refresh-grant precisely so revocation is granular and total.
- **[Sonatype Nexus](https://help.sonatype.com/en/user-tokens.html)** User Tokens are first-class tokens with their own lifecycle; the Docker Bearer Token realm does not mint a separate refresh credential for tool-flows.

### Why silent suppress vs explicit 4xx

Returning `400 DENIED` would break Docker daemons that mechanically send `access_type=offline` regardless of credential type. Major OAuth2 servers (Google, GitHub) graceful-degrade unsupported optional parameters: the response is still `200 OK` with a valid short-lived access token; the client simply re-authenticates with its API token when the access expires (~30 minutes later). The Distribution spec describes `access_type` as the access being _requested_, not a hard contract — the server is free to ignore it.

`offline_refresh_token`'s doc-comment in `oci_v2.rs` carries this rationale verbatim so future readers find it without leaving the source.

## Change

**`backend/src/api/handlers/oci_v2.rs`**:

- `TokenForm` extended with `refresh_token` and `access_type` fields (both `Option<String>`).
- `TokenResponse` gains an optional `refresh_token` field with `#[serde(skip_serializing_if = "Option::is_none")]`. All pre-existing call sites (Bearer-JWT swap, anonymous fallback, password-without-offline) serialize byte-identical to today's shape.
- New `parse_oauth2_form` helper returns the full `TokenForm` (the existing `extract_form_credentials` filters on `grant_type==password` and so cannot see the refresh-grant case).
- New `handle_refresh_grant` async function. Short-circuited at the top of `token()` before credential resolution. Calls `auth_service::refresh_tokens`, which already filters on `is_active = true` and consults `is_token_invalidated`, so deactivated users / TOTP-toggled users / password-changed users all surface as `401` automatically. Returns `400 DENIED` for an empty/missing `refresh_token` field.
- New `offline_refresh_token` helper encodes the response rules in a single place (offline + not-API-token → emit; everything else → suppress). Doc-comment carries the full motivation block.

## Regression test

- [x] N/A — this is a feature, not a bug fix. Test Checklist below covers the new behaviors.

## Test Checklist

13 new tests — 5 sync, 8 DB-backed via `test_db_helpers`.

DB-backed (`token_refresh_grant_tests` module):

- `password_grant_offline_includes_refresh_token` — real user + `access_type=offline` → response carries `refresh_token`.
- `password_grant_default_omits_refresh_token` — no `access_type` → response has no `refresh_token` (backwards-compatible shape).
- **`api_token_offline_does_not_get_refresh_token`** — the security-critical regression: API token in Basic password + `access_type=offline` → 200 OK with access token, but NO refresh token.
- `refresh_grant_default_returns_access_only` — exchange refresh, no `access_type` → fresh access, no rotation (matches Distribution spec: _"only return the same refresh token and not a new one"_; we omit the field so the client keeps using the original).
- `refresh_grant_offline_rotates_refresh_token` — exchange refresh + `access_type=offline` → fresh access AND fresh refresh.
- `refresh_grant_invalid_token_returns_401` — bad refresh token → 401.
- `refresh_grant_missing_token_returns_400` — `grant_type=refresh_token` with empty/missing `refresh_token` field → 400, not silent fallthrough.
- `refresh_grant_deactivated_user_returns_401` — `is_active = false` is enforced by `auth_service::refresh_tokens`.

Sync (`mod tests`):

- `test_token_response_serialization` — `refresh_token: None` is omitted entirely from JSON.
- `test_token_response_serializes_refresh_token_when_some` — present when `Some`.
- `test_offline_refresh_token_returns_some_for_password_grant_offline`
- `test_offline_refresh_token_returns_none_for_password_grant_without_offline` — `None`, `"online"`, empty string, `"OFFLINE"`, `"offline "` all omit; strict equality on `"offline"` matches the spec exactly.
- `test_offline_refresh_token_suppresses_for_api_token_even_when_offline_requested`

- [x] Unit tests added/updated (13)
- [ ] Integration tests added/updated (DB-backed unit tests cover the routing layer end-to-end via `oneshot()`)
- [ ] E2E tests added/updated
- [x] Manually tested locally — full lib suite 9085 / 0 / 2 passes; new code coverage 95% (296/309 new lines instrumented)

## API Changes

- [ ] N/A
- [x] **`POST /v2/token` response now includes an optional `refresh_token` field** when the client requests `access_type=offline` and authenticated as a real user (not API-token-as-password). `skip_serializing_if = Option::is_none` keeps the response shape byte-identical for every other code path — no client breakage.
- [x] **`POST /v2/token` now honors `grant_type=refresh_token`** — clients holding a refresh token can exchange it without re-presenting credentials. New error responses: `400 DENIED` for missing `refresh_token` field, `401 UNAUTHORIZED` for invalid / deactivated-user / invalidated-credential refresh.
- [x] **`GET /v2/token?offline_token=true` now emits a `refresh_token` field** when the client authenticates with Basic-auth credentials (not API-token-as-password). This is the GET-flow equivalent of `access_type=offline` per the [Distribution token spec](https://distribution.github.io/distribution/spec/auth/token/) and is what `docker login` sends. Pre-PR the field was silently dropped by serde (`TokenQuery` didn't declare it) and the response was `{token, access_token, expires_in, issued_at}` with no refresh; post-PR these clients now receive a refresh JWT they didn't get before. Strict-equality match on `"true"` only (`"TRUE"`, `"1"`, etc. fail closed). The API-token-as-password suppression rule applies uniformly: a Basic-auth header carrying an API token + `?offline_token=true` returns 200 OK with the access token but no refresh — same security pivot as the POST path.
- [x] No OpenAPI annotations needed — `/v2/token` is part of the OCI distribution spec, not the AK REST API.

### DIFF
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -199,6 +199,16 @@ struct TokenForm {
     grant_type: Option<String>,
     username: Option<String>,
     password: Option<String>,
+    /// Used by the `grant_type=refresh_token` flow (per the OAuth2 + Docker
+    /// Distribution spec, see RFC 6749 §6 and
+    /// <https://distribution.github.io/distribution/spec/auth/oauth/>).
+    refresh_token: Option<String>,
+    /// `"offline"` (case-sensitive) requests that the response include a
+    /// refresh token. Anything else — including `None`, `"online"`, the
+    /// empty string, or `"OFFLINE"` — issues an access-token-only
+    /// response. Strict equality matches every real client and avoids
+    /// accidentally expanding the refresh-token surface for typos.
+    access_type: Option<String>,
 }
 
 /// Extract `(username, password)` from an OAuth2 password-grant form body.
@@ -235,6 +245,26 @@ fn extract_form_credentials(headers: &HeaderMap, body: &Bytes) -> Option<(String
         return None;
     }
     Some((username, password))
+}
+
+/// Parse the OAuth2 token-endpoint form body in full, returning every field
+/// the Docker Distribution spec defines. Unlike `extract_form_credentials`,
+/// this does NOT filter on `grant_type=="password"` — callers need to see
+/// `grant_type=refresh_token` for the refresh-grant short-circuit and
+/// `access_type` for the offline-access path.
+///
+/// Returns `None` when the body is empty, the Content-Type isn't
+/// `application/x-www-form-urlencoded` (charset suffix allowed), or the
+/// body fails to deserialize.
+fn parse_oauth2_form(headers: &HeaderMap, body: &Bytes) -> Option<TokenForm> {
+    if body.is_empty() {
+        return None;
+    }
+    let ct = headers.get(CONTENT_TYPE)?.to_str().ok()?;
+    if !ct.starts_with("application/x-www-form-urlencoded") {
+        return None;
+    }
+    serde_urlencoded::from_bytes(body).ok()
 }
 
 async fn validate_token(
@@ -3446,7 +3476,133 @@ struct TokenResponse {
     token: String,
     access_token: String,
     expires_in: u64,
+    /// Only emitted when the client explicitly requested `access_type=offline`
+    /// AND the credential isn't an API token (see security note on
+    /// `build_token_response`). `skip_serializing_if` keeps the response
+    /// byte-compatible with the pre-PR shape for every other code path.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    refresh_token: Option<String>,
     issued_at: String,
+}
+
+/// Decide whether the `refresh_token` field appears in the OAuth2 token
+/// response.
+///
+/// Returns `Some(refresh)` only when ALL of the following hold:
+///
+/// 1. `access_type` is exactly `"offline"` (strict, case-sensitive — matches
+///    the [Docker Distribution spec](https://distribution.github.io/distribution/spec/auth/oauth/),
+///    which spells it `offline` in the example flow).
+/// 2. The credential used was NOT an API token.
+///
+/// # Why API-token authentication suppresses the refresh token
+///
+/// API tokens (the `api_tokens` table) are the long-lived credential for
+/// non-interactive flows. They have first-class lifecycle:
+/// `revoke_api_token` flips `revoked_at`, the cache invalidates within
+/// `API_TOKEN_CACHE_TTL_SECS` (5 min), and the token is rejected
+/// everywhere.
+///
+/// Issuing a refresh-token JWT under that authentication would create a
+/// **parallel long-lived credential** signed against `JWT_SECRET` and
+/// NOT tied to the `api_tokens` row. `auth_service::refresh_tokens` does
+/// not blacklist consumed refresh tokens, so this parallel credential
+/// would survive `revoke_api_token` for up to
+/// `jwt_refresh_token_expiry_days` (default 7 days). An attacker who
+/// briefly captured the API token would have a 7-day backdoor after the
+/// defender thought they'd closed access.
+///
+/// Industry precedent for this suppression:
+///
+/// - **RFC 6749 §4.4.3** (`client_credentials` grant): *"A refresh token
+///   SHOULD NOT be included."* API-token-as-password is semantically the
+///   same flow — a machine credential authenticating itself.
+/// - **JFrog Artifactory** deprecated their API-key-as-password path
+///   entirely (End of Life Q4 2024) for this exact lifecycle-management
+///   problem: *"API Keys don't have lifecycle management features […]
+///   single user can have single active API Key — if revoked, revoked
+///   for all clients."* Replaced by reference tokens and identity
+///   tokens with per-token revocation.
+/// - **Sonatype Nexus** User Tokens are designed as first-class tokens
+///   with their own lifecycle and do not get a parallel refresh-grant.
+///
+/// The graceful-degradation choice (silent suppress vs explicit 4xx)
+/// follows the default-permissive convention used by Google, GitHub, and
+/// other major OAuth2 servers: the response is still 200 OK with a valid
+/// access token; the client simply re-authenticates with its API token
+/// when the access expires (~30 min later). Returning 400 here would
+/// break Docker daemons that send `access_type=offline` mechanically
+/// regardless of the credential type.
+fn offline_refresh_token(
+    access_type: Option<&str>,
+    authenticated_via_api_token: bool,
+    refresh: &str,
+) -> Option<String> {
+    if authenticated_via_api_token {
+        return None;
+    }
+    if access_type != Some("offline") {
+        return None;
+    }
+    Some(refresh.to_string())
+}
+
+/// Exchange a refresh token for a fresh access token (and optionally a new
+/// refresh token if `access_type=offline` is supplied again). Per
+/// [the Docker Distribution OAuth2 spec](https://distribution.github.io/distribution/spec/auth/oauth/),
+/// the refresh-grant request body is just `grant_type=refresh_token` +
+/// `refresh_token=…` with no credentials — the refresh token itself
+/// authenticates.
+///
+/// `auth_service::refresh_tokens` filters `is_active = true` and consults
+/// `is_token_invalidated`, so any of (a) deactivated user, (b) password
+/// change since issue, (c) TOTP enable/disable since issue, will surface
+/// here as a 401.
+async fn handle_refresh_grant(state: &SharedState, form: &TokenForm) -> Response {
+    let refresh = match form.refresh_token.as_deref() {
+        Some(s) if !s.is_empty() => s,
+        _ => {
+            return oci_error(
+                StatusCode::BAD_REQUEST,
+                "DENIED",
+                "refresh_token grant requires refresh_token",
+            );
+        }
+    };
+    let auth_service = AuthService::new(state.db.clone(), Arc::new(state.config.clone()));
+    let (_user, tokens) = match auth_service.refresh_tokens(refresh).await {
+        Ok(pair) => pair,
+        Err(_) => {
+            return oci_error(
+                StatusCode::UNAUTHORIZED,
+                "UNAUTHORIZED",
+                "invalid refresh token",
+            );
+        }
+    };
+    // Refresh-grant always originates from a user (not API token) — the
+    // refresh token was issued at password-grant time and the API-token
+    // suppression already prevented one being issued for an API-token
+    // session. So the second argument to `offline_refresh_token` is `false`.
+    // The `access_type` filter still applies: a refresh-grant without
+    // `access_type=offline` does NOT rotate the refresh token (matches the
+    // spec, which says the same refresh is returned).
+    let resp = TokenResponse {
+        token: tokens.access_token.clone(),
+        access_token: tokens.access_token.clone(),
+        expires_in: tokens.expires_in,
+        refresh_token: offline_refresh_token(
+            form.access_type.as_deref(),
+            false,
+            &tokens.refresh_token,
+        ),
+        issued_at: chrono::Utc::now().to_rfc3339(),
+    };
+    Response::builder()
+        .status(StatusCode::OK)
+        .header(CONTENT_TYPE, "application/json")
+        .body(Body::from(serde_json::to_string(&resp).unwrap()))
+        .unwrap()
 }
 
 async fn token(
@@ -3464,6 +3620,30 @@ async fn token(
             return oci_error(StatusCode::BAD_REQUEST, "DENIED", "service mismatch");
         }
     }
+
+    // Parse the full OAuth2 form body once upfront. `extract_form_credentials`
+    // filters on `grant_type=="password"`, so it won't see the refresh-grant;
+    // we need a separate handle on the form to detect that, and to thread
+    // `access_type` through to response construction.
+    let form = parse_oauth2_form(&headers, &body);
+
+    // grant_type=refresh_token short-circuit. Per the Docker Distribution
+    // OAuth2 spec[1], the refresh-grant flow doesn't carry credentials —
+    // the refresh_token itself authenticates. So this branch must be
+    // checked before the credential-resolution chain below.
+    //
+    // [1]: https://distribution.github.io/distribution/spec/auth/oauth/
+    if let Some(ref f) = form {
+        if f.grant_type.as_deref() == Some("refresh_token") {
+            return handle_refresh_grant(&state, f).await;
+        }
+    }
+
+    // `access_type=offline` is opt-in for a refresh token in the response of
+    // the credential-auth path. Captured here so it's in scope for the
+    // success-response builder later in the function.
+    let access_type = form.as_ref().and_then(|f| f.access_type.clone());
+
     // Credential extraction order, per the OCI Distribution Spec + OAuth2:
     //   1. HTTP Basic Auth header (the original code path; works for `docker
     //      login` / `curl -u`).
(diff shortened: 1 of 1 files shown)


brandonrc:
Did a close review pass on this. The OAuth2 mechanics and the API-token suppression rationale are solid. The doc-comment block on `offline_refresh_token` is the kind of inline rationale future readers will thank you for. Three things to address before merge, all scoped to code this PR touches.

### 1. The TOTP invalidation claim is currently false

`handle_refresh_grant`'s doc-comment (oci_v2.rs around line 813-816) says:

> `auth_service::refresh_tokens` filters `is_active = true` and consults `is_token_invalidated`, so any of (a) deactivated user, (b) password change since issue, (c) **TOTP enable/disable since issue, will surface here as a 401**.

Items (a) and (b) hold, but (c) doesn't: `enable_totp` (totp.rs:198-205) and `disable_totp` (totp.rs:415-421) do plain UPDATEs without calling `invalidate_user_tokens`. A pre-TOTP refresh JWT minted via the new password+offline flow can still be exchanged through the refresh-grant short-circuit (oci_v2.rs:882-886), bypassing the password-grant TOTP guard at oci_v2.rs:854.

Easy fix: add `crate::services::auth_service::invalidate_user_tokens(auth.user_id)` after the UPDATEs in both handlers (mirrors what `change_password` does at users.rs:893). A regression test in `token_refresh_grant_tests` that enables TOTP after issuing the refresh JWT and asserts 401 would lock the invariant.

### 2. GET `/v2/token?offline_token=true` isn't handled

The router registers `.get(token).post(token)` (oci_v2.rs:2773). The PR adds `access_type=offline` for the POST OAuth2 path, but the GET basic-auth path used by `docker login` sends `offline_token=true` as a query parameter ([Distribution token spec](https://distribution.github.io/distribution/spec/auth/token/)). `TokenQuery` doesn't declare the field and the binding is `Query(_query)` (oci_v2.rs:711), so this silently degrades to "no refresh token" on what is, in practice, the more common path.

Two ways forward: declare `offline_token: Option<String>` on `TokenQuery`, normalize `Some("true")` to the same offline bool, and thread it through the credential-auth path; or document explicitly that only the OAuth2 POST flow gets a refresh token and the GET path is out of scope.

### 3. Refresh-grant response omits `refresh_token` even on success

`refresh_grant_default_returns_access_only` asserts the field is absent when the client doesn't re-request `access_type=offline`. The [Distribution OAuth2 spec](https://distribution.github.io/distribution/spec/auth/oauth/) says *"only return the same refresh token and not a new one"*, i.e. echo, not omit. Real Docker daemons handle both, so this isn't a real-world regression, but a one-line comment in the test acknowledging the deviation would keep a future maintainer from "fixing" it the wrong way.

---

A few broader hardening items came up (per-process scope of the invalidation map under Helm `replicas > 1`, refresh-token reuse detection per RFC 6819, `service` query validation) that aren't in this PR's scope. I'll file those separately so they don't bloat this review.

The structure of the change is good and the test coverage is strong. The three items above are all small.

brandonrc:
For traceability, the out-of-scope items from the review above are tracked as follow-up issues (not blocking this PR, but worth landing soon after):

- #1173 — Replica-scoped credential invalidation map + 1s race in `is_token_invalidated`. The TOTP-invalidation fix in this PR is correct on a single replica; #1173 makes it correct under Helm `replicas > 1`.
- #1174 — Refresh-token reuse/replay detection per RFC 6819. The API-token suppression rule in this PR closes one amplification vector; #1174 closes the underlying replay window.
- #1175 — Validate `service` query parameter against the configured realm. Pre-existing on `/v2/token`, not introduced here.

inspired-geek:
Thanks for the thorough review. Pushed `03347d4` addressing all three points (rebased on current `main` first to clear the conflict). Quick notes inline.

### 1. TOTP invalidation

I want to flag one factual gap before going further: `totp::enable_totp` and `totp::disable_totp` already call `invalidate_user_tokens(auth.user_id)` after their `UPDATE` — at `totp.rs:213` and `totp.rs:435` respectively. The `enable_totp_invalidates_pre_change_user_tokens` (`totp.rs:746`) and `disable_totp_invalidates_pre_change_user_tokens` (`totp.rs:817`) tests pin that. So the doc-comment claim isn't currently false at the auth-service layer — but your underlying concern is right that the new refresh-grant short-circuit had no test of its own asserting the invariant end-to-end on the OCI handler. Added `refresh_grant_after_totp_toggle_returns_401` in `token_refresh_grant_tests` to own the second half of the chain (issue refresh JWT, sleep past the second-granularity boundary, `invalidate_user_tokens`, attempt refresh → 401). The sleep mirrors the timing-disambiguation pattern in `enable_totp_invalidates_pre_change_user_tokens`. Happy to expand into an end-to-end HTTP enable_totp call if you'd prefer that shape — let me know.

### 2. `GET /v2/token?offline_token=true`

Implemented per the Distribution token spec. `TokenQuery` gains `offline_token: Option<String>`; a small helper `offline_access_type_from_query` normalizes a strict `"true"` match into the same `"offline"` string consumed by `offline_refresh_token`. Strict equality (mirrors the strict `"offline"` match on the POST path — `"TRUE"`/`"1"`/empty/trailing-space all fail closed) is locked down by `test_offline_access_type_from_query_strict_true_only`. Three DB-backed tests cover the GET path:

- `get_offline_token_true_emits_refresh_for_password_auth` — happy path.
- `get_without_offline_token_omits_refresh` — backwards-compat shape, asserted on raw JSON bytes.
- `get_api_token_offline_token_true_does_not_emit_refresh` — extends the API-token-suppression security invariant to the GET path. Without this, a regression could have leaked a refresh through the new code without firing the existing `api_token_offline_does_not_get_refresh_token` POST test.

### 3. Refresh-grant omit vs. echo

Expanded the comment on `refresh_grant_default_returns_access_only` to spell out the deviation and link the spec URL. Future maintainer will see the rationale before "fixing" it.

### Bonus items from a parallel automated review (low-risk, all in this commit)

- `TokenResponse.refresh_token`'s doc-comment pointed at a non-existent `build_token_response`. Now references `offline_refresh_token`.
- Three DB tests used `body.get("refresh_token").is_none()`, which is true only for an absent key but would have passed for `Some(Value::Null)`. Replaced with raw-bytes substring checks — same shape as the existing sync `test_token_response_serialization` assertion, matching the wire contract `skip_serializing_if = Option::is_none` actually promises. Applied to the security-critical `api_token_offline_does_not_get_refresh_token` regression too.
- Added five sync tests for `parse_oauth2_form` (empty body / missing CT / wrong CT / valid password grant / valid refresh grant) so the helper has unit coverage independent of the DB-backed handler tests.

### Follow-ups acknowledged

#1173 / #1174 / #1175 noted as out-of-scope. The TOTP invalidation fix in this PR + #1173 (replica-scoped map) together close the multi-replica gap; #1174 closes the underlying replay window the API-token suppression in this PR only narrowed.

### Verification

- `cargo fmt --check` / `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `cargo test --workspace --lib`: 9160 passed / 0 failed / 2 ignored.
- Local `cargo llvm-cov` with a real Postgres (matching the CI service): **97% coverage on new/changed lines** (473/487), well above the 70% gate.

brandonrc:

brandonrc:
## Rebased onto current `main` + security review — QUEUED for maintainer review

Rebased this branch onto `main` (was DIRTY behind ~480 commits). Force-pushed the rebased history to the fork (`6b2b3c3`). Authorship preserved (Alexey Ivanov).

### Conflict resolution (`backend/src/api/handlers/oci_v2.rs`, 2 hunks + 1 semantic fixup)
1. **`validate_token` async-ification** — `main` made `validate_token` `async` (#1173 replica-safe watermark). Kept main's `async fn` and inserted the PR's new `parse_oauth2_form` helper alongside it.
2. **`token()` preamble** — `main` added the #1175 `service`-mismatch check (`400 DENIED` unless `service == "artifact-keeper"`). Ordered it **before** the PR's form-parse + refresh-grant short-circuit, so the service guard runs first on every path including the refresh grant.
3. **Semantic fixup (not flagged by git)** — the PR's 3 new GET-flow tests requested `?service=registry`, which #1175 now rejects with `400`, breaking their `assert_eq!(OK)`. Updated them to `?service=artifact-keeper`. Without this the rebased branch would fail Tier-2 CI.

### Security analysis — token issuance
- **No privilege escalation via refresh/offline.** The OCI access JWT minted by `generate_tokens` carries **no `scopes` / `allowed_repo_ids`** — it's a full-user-privilege token, exactly as the pre-existing password grant (#894) produces. The refresh grant re-mints the same shape for the same `sub`. It cannot grant more than the original login already did.
- **Scope enforcement (#1316) is not bypassed.** OCI per-request authz (`oci_scopes_grant` → `token_service::scopes_grant_access`) operates on the scope set carried *alongside the credential within a request* (API-token validation), not on the issued JWT. The documented TODO that token-exchange doesn't fold per-token scopes into the JWT is **pre-existing on `main`** and is **not widened** here: refresh tokens are only ever issued to real-user password/Basic auth, never to API tokens.
- **Token-mint constraints (#1315).** `enforce_admin_only_scopes` guards REST token-*creation* handlers; `/v2/token` doesn't accept caller-supplied scopes, so this path is out of scope for that guard and nothing here circumvents it.
- **API-token suppression holds.** `offline_refresh_token(..., authenticated_via_api_token, ...)` returns `None` whenever auth was via API-token-as-password, on **both** POST (`access_type=offline`) and GET (`offline_token=true`) paths. Verified by `api_token_offline_does_not_get_refresh_token` and `get_api_token_offline_token_true_does_not_emit_refresh`.
- **Revocability.** Refresh tokens are issued only to real users and are subject to `auth_service::refresh_tokens`, which on `main` now (post-#1174) enforces: `is_active` (deactivated → 401), credential-change watermark (password/TOTP change → 401), `jti` single-use rotation, RFC-6819 replay detection, and **family revocation**. This is *stronger* than the PR's doc-comment assumes (the comment predates #1174 and still says consumed tokens aren't blacklisted — that justification text is now stale, though the conservative suppression decision it argues for remains correct). The 7-day non-revocable-credential concern the PR raised for API tokens is genuinely avoided by suppression.
- **Strict matching, fail-closed.** `access_type` strict-equals `"offline"`; `offline_token` strict-equals `"true"`. Casing/whitespace/`"1"` variants all fall through to access-only. Covered by unit tests.

### Gates
- `cargo fmt --check` — pass
- `cargo clippy --workspace --all-targets -- -D warnings` — pass (clean)
- `cargo test --workspace --lib` — 11151 pass / 2 fail, **both pre-existing & environment-only** (`http_client::...does_not_apply_total_timeout` — sandbox blocks the localhost TCP connect; confirmed fails identically on clean `origin/main`. `oci_v2::...negative_cache...isolated` — parallel-execution flake, passes in isolation). Neither touches PR code.
- All 13 new token tests + the 3 GET tests pass (DB-backed ran green in my env).
- jscpd on the changed file — no clones above threshold. PR self-reports ~95% new-line coverage (13 dedicated tests cover every new fn + branch).

### Verdict: **READY** on the merits, **QUEUED** per policy
Token issuance is security-sensitive; queuing for maintainer eyes rather than arming. No scope-bypass or priv-esc found, gates green (modulo two pre-existing env flakes). Recommend a maintainer also refresh the stale doc-comment in `offline_refresh_token` to reference the #1174 family-revocation machinery, though it's non-blocking.

brandonrc:
Thanks for this — the offline-issuance half and the API-token refresh-suppression are well-built and the design is sound (no scope escalation on refresh, no IDOR, and suppressing refresh emission for API tokens correctly closes a privilege-widening path). Independent review flagged **one correctness/availability blocker** before this can merge:

**🔴 Default refresh-grant poisons the caller's token → family-wide revocation on the next refresh (session DoS).**
`handle_refresh_grant` (oci_v2.rs) passes the rotated `tokens.refresh_token` through `offline_refresh_token(access_type, false, …)`. `auth_service::refresh_tokens` does **mandatory single-use rotation** (#1174): it marks the presented `jti` `consumed_at` and mints a new refresh token. But with no `access_type=offline` (the default refresh case), `offline_refresh_token` returns `None`, so the **new** refresh token is discarded and the client is left holding the **now-consumed** original. On the client's next refresh, replay detection fires → `revoke_all_refresh_token_families` → 401, forcing full re-auth. The doc-comment's premise ("client keeps using the original") doesn't hold under single-use rotation.

**Fix:** in the refresh-grant path, always return the rotated `tokens.refresh_token` (rotation already happened; the old one is dead), independent of `access_type`. The `access_type` gate is correct for *first issuance* (password grant) but wrong for the refresh grant.

**Test gap:** `refresh_grant_default_returns_access_only` does a single refresh and stops, so the lockout is invisible. Please add a two-call test: refresh → refresh again with the **returned** token, asserting the second succeeds.

**Minor:** the `offline_refresh_token` doc-comment states `refresh_tokens` "does not blacklist consumed refresh tokens" — it does (replay detection + family revocation). Please correct the premise. The suppression decision itself is still fine.

Happy to help land the fix. Once the pass-through + two-call test are in, this is merge-ready for v1.4.0.

PR #1216, opened 2026-05-13
fix(maven): virtual download for GAV-grouped secondary files (.pom, .module, -sources.jar)

ThaSami:
## The bug

Virtual repos return **404** for Maven secondary files (`.pom`, `.module`, `-sources.jar`, `.sha512`) when the same files are served correctly from the underlying local repo at the same path.

Concrete repro on a deployment with one local maven repo and one virtual aggregating it (member: `mobile-libs-release-local`, virtual: `mobile-libs-release`):

```
GET /maven/mobile-libs-release-local/com/TEST/tools/gradle-plugin/0.5.88/gradle-plugin-0.5.88.pom -> 200 (CL 2065)
GET /maven/mobile-libs-release/com/TEST/tools/gradle-plugin/0.5.88/gradle-plugin-0.5.88.pom       -> 404
```

The `.jar` at the same coordinate works through both — only the GAV-secondary files break on the virtual path. This silently breaks every Gradle/Maven build that pulls plugins through a virtual.

## Root cause

The maven upload handler groups files by GAV coordinate: only the primary file (`.jar`/`.aar`) gets a row in `artifacts`, while the secondary files (`.pom`, `.module`, `-sources.jar`, `.sha512`) live on storage at the same path but share the primary's DB row.

The maven **local** download handler already accounts for this with a storage-direct fallback ([`maven.rs:1201-1220`](https://github.com/artifact-keeper/artifact-keeper/blob/main/backend/src/api/handlers/maven.rs#L1201-L1220)):

```rust
// For hosted repos, fall back to serving from storage directly.
// This handles secondary files (POM, sources, javadoc) that were
// grouped under a primary artifact record by GAV grouping ...
if repo.repo_type == RepositoryType::Local || repo.repo_type == RepositoryType::Staging {
    // storage.get(format!("maven/{}", path)) -> 200 if hit
}
```

But the **virtual** download path goes through [`proxy_helpers::resolve_virtual_download`](https://github.com/artifact-keeper/artifact-keeper/blob/main/backend/src/api/handlers/proxy_helpers.rs) → [`proxy_helpers::local_fetch_by_path`](https://github.com/artifact-keeper/artifact-keeper/blob/main/backend/src/api/handlers/proxy_helpers.rs#L941), which is **SQL-only**:

```rust
sqlx::query_as!(LocalArtifactRow,
    "SELECT ... FROM artifacts WHERE repository_id = $1 AND path = $2 AND is_deleted = false LIMIT 1")
    ...
    .ok_or_else(|| (StatusCode::NOT_FOUND, "Artifact not found").into_response())?;
```

SQL miss → `NotFound` returned to `resolve_virtual_download` → no member ever matches → 404 from the virtual response.

## The fix

Adds `proxy_helpers::maven_local_fetch_storage_fallback` — a Maven-specific helper that mirrors the local handler's storage-direct fallback (key prefix `maven/<path>`). Wires it into the maven virtual download closure as the 3rd-stage fallback after the existing SQL lookup and SNAPSHOT alias resolution:

| Stage | Helper | When it matches |
|---|---|---|
| 1 | `local_fetch_by_path` | Primary artifact with exact SQL row match (e.g. the `.jar`) |
| 2 | `maven_local_fetch_snapshot` | `-SNAPSHOT` filename alias → latest timestamped artifact |
| 3 (new) | `maven_local_fetch_storage_fallback` | GAV-secondary bytes in storage without their own row (`.pom`, `.module`, `-sources.jar`, `.sha512`) |

Same Maven-specific `format!("maven/{}", path)` storage-key prefix the upload handlers and the local download fallback already use, so there's no behavioral change for any non-Maven format and no risk of cross-format storage leakage. Tests guard that explicitly.

## Tests

Three new DB-backed tests in `proxy_helpers.rs::tests`, all following the existing `try_pool().await` pattern — they skip cleanly without `DATABASE_URL` and execute in the CI coverage job that seeds Postgres:

- **`test_maven_local_fetch_storage_fallback_hit`** — bytes-only at `maven/<path>`, no `artifacts` row → returns content. Exercises the GAV-grouped shape directly.
- **`test_maven_local_fetch_storage_fallback_miss_returns_not_found`** — no bytes, no row → `NotFound` (the shape `resolve_virtual_download` expects so it tries the next member).
- **`test_maven_local_fetch_storage_fallback_ignores_non_maven_prefix`** — bytes at `pypi/<path>` but request hits the maven fallback → `NotFound`. Regression guard against accidentally dropping the `maven/` prefix.

Existing test for `local_fetch_by_path` untouched — confirms no behavioral change for the SQL-hit path.

## Discovery context

Reported during a JFrog → artifact-keeper migration. Every Gradle plugin resolution from a virtual repo failed at the `.pom` step despite the `.jar` resolving cleanly. The artifact-keeper internal CI suite has integration coverage of primary `.jar` resolution via virtuals, but `.pom`/`.module`/`-sources.jar` requests through virtuals didn't have a test that catches the SQL-only lookup. The new tests close that gap.


ThaSami:
@brandonrc any update here? this is a serious bug that's also affecting repository real size ( as its not counting the other files )

brandonrc:
@ThaSami  hey! Sorry been working hard to get all of the original bugs/hardening in place along with adding a decent amount of tests. 

I will take a look at this right now and get it into main :) 

Thanks for the help and I always appreciate extra nudges/pings!

brandonrc:
Solid PR! I am going to add a few fixes directly to this branch so we can get it merged in fast. 

Is that ok? 

I am trying to get a 1.2.0 released soon and want to make sure this is in it.

brandonrc:
@ThaSami — appreciate the patience. Just pushed commit `7667af1` to your branch with security + correctness hardening on top of your fix. The original bug analysis and the storage-direct approach were exactly right; the follow-ups address concerns from a 4-agent audit (Code Reviewer voted MERGE, but Security / Test Strategy / Architecture each had concerns worth addressing before this lands).

## Changes in `7667af1`

1. **Quarantine + soft-delete bypass closed.** The original helper read storage directly with no DB check. When the primary `.jar` is quarantined or soft-deleted, the companion `.pom`/`-sources.jar` had no row of their own and would leak. New helper enforces 3 gates: extension allowlist (refuses `.jar`/`.aar`/`.war`/`.ear`/`.zip` primaries), live-primary requirement (rejects orphan secondaries), and quarantine inheritance from the primary row.

2. **SNAPSHOT-error masking fixed.** The call-site rewrite at `maven.rs:1210` converted the SNAPSHOT helper from a terminal `.await` to `if let Ok(...) { return }`, silently letting a SNAPSHOT-mapping failure fall through to the storage fallback on the same member. If member A had stale bytes at the literal `-SNAPSHOT` path from a previous deploy, those would be served instead of advancing the virtual loop to member B. Now: SNAPSHOT paths short-circuit to the snapshot helper exclusively.

3. **Module split: new `handlers/maven_proxy.rs`.** Format-specific helpers don't belong in the shared `proxy_helpers.rs`. New file holds the maven-specific fallback + its gates. Shared module keeps only format-agnostic primitives. Pattern documented in the file header for the next format (Debian / RPM / NuGet / Helm) that hits the same GAV-grouping shape.

4. **Test corpus expanded 3 → 10.** The prior `_ignores_non_maven_prefix` test was a tautology (it planted `pypi/...` bytes and queried with a maven path that the helper prefixes with `maven/`, so the lookup key never matched — the test would pass even with an `xyz/{}` prefix). Replaced with a positive+negative control. Added coverage for every secondary extension you list in the PR body (`.pom`, `.module`, `-sources.jar`, `-javadoc.jar`, `-tests.jar`, `-test-sources.jar`, `.sha1`, `.sha256`, `.sha512`, `.md5`, `.asc`), plus regression tests for primary-extension refusal, orphan-secondary refusal, quarantine inheritance, and soft-delete inheritance.

## Verification

`SQLX_OFFLINE=true cargo check -p artifact-keeper-backend --lib --tests` clean.

The original 404 your test repro hit will still 200 after this — the gates only kick in when there's an actual policy violation (quarantined primary, missing primary, primary-extension probe).

Worth noting: the analogous quarantine bypass also exists on the **hosted-repo** storage fallback at `maven.rs:1264-1284` (which this PR does NOT touch). That's the same root cause; I left it for a separate PR so this one stays focused on the virtual-download regression you reported. Filed mentally as a follow-up.

Once CI re-greens this should be ready. Thanks again for the clean repro and root-cause writeup.

brandonrc:
Pushed `1124a322` to your branch to unblock the Check Rust gate.

`cargo fmt --check` was failing on five small diffs across two files:

- `backend/src/api/handlers/maven_proxy.rs` — `maven_fixture()` return-type tuple, an `assert_eq!` arg layout, and a `let err = ...` chain
- `backend/src/api/handlers/proxy_helpers.rs` — two stray blank lines between functions

Pure formatting, no semantic changes. `cargo check --workspace --lib` passes locally; `cargo fmt --all -- --check` now clean.

Used `maintainerCanModify` since the gate was blocking — feel free to amend, force-push, or revert if you'd prefer to land it yourself. Just wanted to keep the PR moving.

brandonrc:
Pushed `0fcf90b9` to also unblock the Code Coverage gate.

The failure was in your new test `test_storage_fallback_honors_quarantine_on_primary`:

```
thread '...' panicked at backend/src/api/handlers/maven_proxy.rs:494:9:
expected a refusal status, got 409 Conflict
```

The accept-list was `FORBIDDEN | UNAVAILABLE_FOR_LEGAL_REASONS | NOT_FOUND`, but `check_quarantine_row` delegates to `quarantine_service::check_download_allowed`, which returns `AppError::Conflict / 409` for `quarantine_status='quarantined'` (see `backend/src/services/quarantine_service.rs:74-89`). The 403/451/404 branches are still reachable in tenant-policy plug-ins, so I kept those and added `CONFLICT` alongside — the security property under test (companion .pom must NOT be served) is unchanged.

Verified locally:

```
cargo test --workspace --lib -- api::handlers::maven_proxy::tests
running 10 tests
test result: ok. 10 passed; 0 failed
```

All other gates (Check Rust, Backend Unit Tests, Security Audit, etc.) were already green on the previous push.

ThaSami:
on it, thanks for review, let me fix the test.

brandonrc:
@ThaSami  if this passes I will just merge it in. If you want to start working that other issue that would be amazing. 

I have to take a break now and get to my real job 🤣 

Thanks for all the help!

PR #974, opened 2026-04-29
fix(hex): virtual repo pull through names

todpunk:
## Summary

  Fixes three related bugs that prevented `hex-virtual` from functioning as a pull-through cache when paired with a `hex-remote` member.

  **Root causes:**

  1. `GET /names` and `/versions` returned empty results. Both aggregation helpers skipped Remote members entirely, so packages cached by `hex-remote` were invisible to the virtual repo's index.
  2. `GET /packages/{name}` for a virtual repo called `resolve_virtual_metadata` directly, which only queries remote upstreams — it never checked any member's database, including local/hosted members. Internally-published packages in `test-hex` were also unreachable through the virtual endpoint.
  3. No name-shadowing protection: with the above fixed, a remote upstream could still serve any package version even if the same name was published internally.

  **What changed:**

  - `query_local_member_names` and `query_local_member_versions` now query all member repos' databases, including Remote member caches.
  - `package_info` for virtual repos checks member databases in two passes — non-Remote members before Remote, regardless of configured priority — then falls to upstream proxy only for packages not yet cached anywhere.
  - `download_tarball` for virtual repos parses the package name from the filename and passes `None` as the proxy service if any non-Remote member owns that name, blocking new upstream fetches for shadowed names while still serving locally-cached content.
  - Extracted `fetch_package_info_from_repo` to consolidate the DB query + response-building logic used across all three repo type paths in `package_info`.

  ## Regression test (required for `fix/*` PRs)

  - [x] This PR is a `fix/*` AND adds/updates a test that would have caught the bug
  - [ ] N/A — this is not a bug fix

  > **Note:** 8 new unit tests cover `package_name_from_tarball_filename` and tarball URL routing. The core bugs live in DB-dependent paths not reachable from `cargo test --lib`. A full integration regression test is tracked in the linked issue.

  ## Test Checklist

  - [x] Unit tests added/updated
  - [ ] Integration tests added/updated (if applicable)
  - [ ] E2E tests added/updated (if applicable)
  - [ ] Manually tested locally
  - [x] No regressions in existing tests

  ## API Changes

  - [ ] New endpoints have `#[utoipa::path]` annotations
  - [ ] Request/response types have `#[derive(ToSchema)]`
  - [ ] OpenAPI spec validates: `cargo test --lib test_openapi_spec_is_valid`
  - [ ] Migration is reversible (if applicable)
  - [X] N/A - no API changes

PR #1217, opened 2026-05-13
fix(hex): block upstream shadowing on virtual tarball download (closes #973)

### LINKED ISSUE #973: GET /hex/{virtual-repo}/names returns [] when virtual repo contains a remote pull-through member
### Description

Describe the bug

When a hex-virtual repo is configured with a hex-remote pull-through member and a local hosted member (discord-hex), querying /hex/hex-virtual/names returns an empty JSON array. Individual package fetches through the virtual repo may work, but `mix deps.get` fails immediately because it uses the names index to discover available packages.

Additional symptoms (same root cause):

  - GET /hex/hex-virtual/versions also returns []
  - GET /hex/hex-virtual/packages/{name} returns 404 for packages published to a local member repo, even if that repo is a member of the virtual repo
  - mix deps.get configured to point at the virtual repo fails immediately because the empty names index means no packages are "available"

Root cause

The names/versions aggregation helpers explicitly skipped Remote member repos in their DB queries. The package metadata endpoint for virtual repos skipped local member databases entirely and only queried remote upstreams. Both are fixed in PR #974 .

I have also addressed a possible security vector of an upstream named item for hex packages shadowing the local member items so that's not a supply chain attack vector.

### Environment Information

- Deployment type: Kubernetes using the gcr container
- Operating System: Kubernetes
- Architecture: 
- Database: Postgres
- Reverse Proxy: Tested with several including direct to the service for backend
- Network setup: See above

### Backend Version

1.1.6

### Web UI Version

N/A

### Package Format

None

### Expected Behavior

To work as a pull-through cache, it will need to return the list for the upstream repo in names. So we expect a JSON array containing "cowboy" (and any other cached packages)

### Observed Behavior

It returns `[]` instead

### Steps to Reproduce

  Steps to reproduce

  1. Configure three repos: test-hex (local/hosted), hex-remote (remote, upstream https://repo.hex.pm), hex-virtual (virtual, members: both of the above)
  2. Pull any package through hex-remote so it exists in its cache: GET /hex/hex-remote/packages/cowboy
  3. Query GET /hex/hex-virtual/names

### Logs

```shell
N/A
```

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

Closes the supply-chain shadowing gap that the abandoned PR #974 (Tod Hansmann / todpunk) was trying to fix on the hex tarball download path. PR #1209 already forward-ported the metadata side via `order_members_local_first`; this PR completes the pair by adding the matching guard on the bytes side (`GET /hex/{repo}/tarballs/{name}-{version}.tar`).

The bug: on a virtual hex repo with members ordered `[hex-remote (priority 1), hex-local (priority 2)]`, an attacker who pushes a package named `phoenix` to the remote upstream can shadow the operator's locally-published `phoenix` because `resolve_virtual_download` iterates members in configured priority order. The metadata endpoint catches this by reordering members local-first regardless of configured priority; the tarball endpoint did not.

What changed:

- `package_name_from_tarball_filename` (new pure helper) parses `<name>-<version>.tar` into the name. Hex package names may contain dashes (`my-package-1.4.0.tar`), so the split is "first `-` followed by an ASCII digit." Matches the parser shipped on PR #974.
- `virtual_local_owns_tarball_name` (new) returns true if any non-Remote member of a virtual hex repo has an artifact row matching the parsed package name (case-insensitive via the existing `proxy_helpers::find_artifact_by_name_lowercase`).
- `serve_virtual_tarball_local_only` (new) calls `proxy_helpers::resolve_virtual_download` with `proxy_service: None`, so the iteration cannot escape to an upstream proxy fetch even if member priority orders Remote first.
- `download_tarball`: when the guard fires (local member owns the name), routes through the local-only path. Otherwise the existing `try_remote_or_virtual_download` path handles the request unchanged.

- [ ] N/A, this is not a bug fix

The unit-level regression is the 8 new tests on `package_name_from_tarball_filename` that pin the parsing behavior the shadowing guard depends on. A full DB-backed integration test would require seeding a virtual repo with mixed member types and a malicious-upstream mock; tracked as a follow-up beads issue rather than blocking this PR, consistent with how the related metadata-side fix shipped in #1209.

## Test Checklist

- [x] Unit tests added/updated (8 new tests on `package_name_from_tarball_filename`)
- [ ] Integration tests added/updated -- N/A for this PR (see above)
- [ ] E2E tests added/updated -- tracked as follow-up
- [x] Manually tested locally: `cargo fmt --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo test --workspace --lib` 9474 passed / 0 failed / 2 ignored.
- [x] No regressions in existing tests.

## API Changes

- [x] N/A, no API changes

## Closes

Closes #973 (completes the pair with #1209)

## Note on PR #974

This forward-ports the last remaining piece of todpunk's abandoned PR #974. Bug #1 (`/names` `/versions` skip Remote members) shipped in #1209 with a different design (upstream proxy fetch instead of DB-only). Bug #2 (`/packages/{name}` ignores non-Remote members) shipped in #1209 with todpunk's two-pass design adapted. Bug #3 (this PR) is the matching tarball-side guard.


brandonrc:
Coverage was failing on `services::storage_gc_service::tests::test_run_gc_dry_run_keeps_oci_blob_referenced_by_blob_index` — a test in `storage_gc_service.rs` that this PR does not touch. Confirmed flaky: re-ran the coverage job with no code change and it passed (https://github.com/artifact-keeper/artifact-keeper/actions/runs/25807484928).

Plausible root cause: the storage GC test shares the `tdh::Fixture` DB pool with other DB-backed tests; under `cargo llvm-cov` instrumentation the binary runs slower and timing shifts can let a concurrent test mutate `oci_blobs`/`artifacts` state mid-sweep. The recent main run (`f7246476`) passed the same test, and this PR adds 2 new DB-backed integration tests in `hex.rs` which added enough parallel DB load to surface the flake.

Tracking as a follow-up beads issue rather than blocking this PR.

PR #1234, opened 2026-05-15
fix(security): bump lettre to 0.11.22 and metrics to 0.24.6

brandonrc:
## Summary

Cargo audit (Security Audit job) is currently failing on every PR — OIDC, the dependabot bumps, the docker-compose fix — because of:

- **RUSTSEC-2026-0141** in `lettre <0.11.22`: TLS hostname verification disabled in the `boring-tls` integration. We build with `tokio1-native-tls`, so the actual vulnerability does not affect us, but `cargo audit` does not inspect enabled feature flags. Patched in `lettre 0.11.22`. Our `Cargo.toml` already specifies `\"0.11\"`, so a `Cargo.lock` refresh is sufficient.
- **`metrics 0.24.5` yanked**: bumped to `0.24.6` (also lockfile-only).

After the bumps, `cargo audit` exits 0 locally. Five informational warnings remain (`rand` unsound transitively via `object_store`/`quinn-proto`/`totp-rs`, plus `fxhash`/`paste` unmaintained). These do not block the audit gate and are upstream issues we cannot resolve here.

- [x] N/A — this is not a behavior bug. The regression test for this class of failure is the Security Audit CI job itself, which will pass on this branch and would re-fail if the lockfile regressed.

## Test Checklist
- [x] Manually tested locally (`cargo audit` exit 0 in worktree)
- [x] No regressions in existing tests (lockfile-only change, no source modified)

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -859,7 +859,7 @@ dependencies = [
  "cap-primitives",
  "cap-std",
  "io-lifetimes",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
@@ -888,7 +888,7 @@ dependencies = [
  "maybe-owned",
  "rustix 1.1.4",
  "rustix-linux-procfs",
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
  "winx",
 ]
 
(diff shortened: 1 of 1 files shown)


brandonrc:
Superseded by #1244 — that PR includes this same lettre/metrics bump (cherry-picked verbatim) alongside the Docker Hub mirror fix that was blocking main. Both fixes landed together in f73ed88.

PR #1236, opened 2026-05-15
fix (debian/apt) Use real OpenPGP signatures for Debian repository metadata

### LINKED ISSUE #1307: `apt update` rejects signed Debian repositories with `Malformed packet: Malformed CTB` because metadata signatures aren't real OpenPGP
## Summary

The Debian/APT signing path generates a plain RSA PEM key for `key_type: "gpg"` and wraps raw RSA signature bytes in `-----BEGIN PGP SIGNATURE-----` ASCII armor. The output looks like OpenPGP but does not contain valid OpenPGP packets, so `apt`/`sqv` reject it.

## Symptom

`apt update` against a signed Debian repository served by Artifact Keeper fails with:

```
Malformed packet: Malformed CTB
```

The public Debian key endpoint returns `-----BEGIN PUBLIC KEY-----` (a plain PEM) instead of `-----BEGIN PGP PUBLIC KEY BLOCK-----`. `InRelease`, `Release.gpg`, and `gpg-key.asc` all carry the same wrong format, forcing operators to bypass signature validation with `[trusted=yes]` to make APT work at all.

## Root cause

The Debian metadata signing path uses raw RSA primitives and ASCII-armor headers rather than producing real OpenPGP packets. There is no path that generates an OpenPGP key, a cleartext-signed `InRelease`, or a detached `Release.gpg`.

## Repro

1. Create a Debian repository.
2. Configure repository signing with a `gpg` key.
3. Fetch `gpg-key.asc` and observe it begins with `-----BEGIN PUBLIC KEY-----` rather than `-----BEGIN PGP PUBLIC KEY BLOCK-----`.
4. Configure apt with `signed-by=` (not `[trusted=yes]`) and run `apt update`; observe `Malformed packet: Malformed CTB`.

## Fix

Being addressed by PR #1236, which generates real OpenPGP key material via `pgp`/rPGP, stores ASCII-armored public/secret keys, emits real OpenPGP cleartext-signed `InRelease` and detached `Release.gpg`, and updates the Debian E2E setup to use `signed-by=` instead of `[trusted=yes]`.

D13410N3:
Closes #1307

## Summary

Fix Debian/APT repository signing so `InRelease`, `Release.gpg`, and `gpg-key.asc` contain valid OpenPGP material that `apt`/`sqv` can verify.

Previously, `key_type: "gpg"` generated a plain RSA PEM key pair and Debian metadata signing wrapped raw RSA signature bytes in `-----BEGIN PGP SIGNATURE-----` ASCII armor. That produced PGP-looking output, but not valid OpenPGP packets. `apt update` rejected signed Debian repositories with errors like:

`Malformed packet: Malformed CTB`

This PR changes the Debian signing path to:

- generate real OpenPGP key material for `key_type: "gpg"` using `pgp`/rPGP
- store ASCII-armored OpenPGP public/secret keys for GPG signing keys
- emit a real OpenPGP cleartext-signed `InRelease`
- emit a real OpenPGP detached `Release.gpg`
- keep existing raw RSA `sign_data()` behavior for other formats that already use it
- update Debian E2E setup to use `signed-by=` instead of `[trusted=yes]`

The public Debian key endpoint now returns:

`-----BEGIN PGP PUBLIC KEY BLOCK-----`

instead of a plain PEM:

`-----BEGIN PUBLIC KEY-----`

Regression coverage:

- Added/updated unit coverage for OpenPGP key generation and signing:
  - verifies generated GPG keys are parseable OpenPGP public keys
  - verifies detached signatures parse and verify against the generated key
  - verifies cleartext signatures parse and verify against the generated key
- Updated Debian E2E script to import `gpg-key.asc` with `gpg --dearmor` and configure apt with `signed-by=`, instead of bypassing signature validation with `[trusted=yes]`.

## Test Checklist

- [x] Manually tested locally

Local/manual validation performed:

- `cargo fmt --check`
- `cargo test -p artifact-keeper-backend --lib -q`
- `cargo test -p artifact-keeper-backend debian::tests --lib`
- `cargo test -p artifact-keeper-backend signing_service::tests::test_openpgp_key_and_signatures_are_parseable_and_verifiable --lib`
- `cargo clippy -p artifact-keeper-backend --lib -- -D warnings -A unused-assignments`

Manual end-to-end validation against a test deployment:

- built and deployed a test backend image with this patch
- created a new Debian repository
- configured repository signing with a newly generated `gpg` key
- confirmed `gpg-key.asc` returns `-----BEGIN PGP PUBLIC KEY BLOCK-----`
- confirmed `apt update` accepts the repository without `[trusted=yes]`
- uploaded deb package through the Debian endpoint
- confirmed `apt-cache policy <package name>` sees the package
- confirmed `apt install <package name>` downloads and installs from Artifact Keeper

Note: plain `cargo clippy -p artifact-keeper-backend --lib -- -D warnings` currently fails on an existing unrelated `unused_assignments` warning in `backend/src/api/handlers/promotion.rs`. The clippy command above allows that pre-existing warning so this PR’s changes can still be linted.

## API Changes

- [x] N/A - no API changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -157,6 +157,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "argon2"
+version = "0.5.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3c3610892ee6e0cbce8ae2700349fcf8f98adb0dbfbee85aec3c9179d29cc072"
+dependencies = [
+ "base64ct",
+ "blake2",
+ "cpufeatures 0.2.17",
+ "password-hash",
+ "zeroize",
+]
+
+[[package]]
 name = "artifact-keeper-backend"
 version = "1.1.2"
 dependencies = [
@@ -199,9 +212,11 @@ dependencies = [
  "opentelemetry-http",
  "opentelemetry-otlp",
  "opentelemetry_sdk",
+ "pgp",
  "prost",
  "prost-types",
  "quick-xml",
+ "rand 0.8.5",
  "rand 0.9.2",
  "regex",
  "reqwest 0.13.2",
@@ -748,6 +763,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
 
 [[package]]
+name = "bitfield"
+version = "0.14.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d7e60934ceec538daadb9d8432424ed043a904d8e0243f3c6446bce549a46ac"
+
+[[package]]
 name = "bitflags"
 version = "2.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -803,6 +824,25 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "63044e1ae8e69f3b5a92c736ca6269b8d12fa7efe39bf34ddb06d102cf0e2cab"
+dependencies = [
+ "memchr",
+ "serde",
+]
+
+[[package]]
+name = "buffer-redux"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "431a9cc8d7efa49bc326729264537f5e60affce816c66edf434350778c9f4f54"
+dependencies = [
+ "memchr",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.19.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -848,6 +888,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f3a53fac24f34a81bc9954b5d6cfce0c21e18ec6959f44f56e8e90e4bb7c346c"
 dependencies = [
  "libbz2-rs-sys",
+]
+
+[[package]]
+name = "camellia"
+version = "0.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3264e2574e9ef2b53ce6f536dea83a69ac0bc600b762d1523ff83fe07230ce30"
+dependencies = [
+ "byteorder",
+ "cipher",
 ]
 
 [[package]]
@@ -929,6 +979,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cast5"
+version = "0.11.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "26b07d673db1ccf000e90f54b819db9e75a8348d6eb056e9b8ab53231b7a9911"
+dependencies = [
+ "cipher",
+]
+
+[[package]]
 name = "cbc"
 version = "0.1.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -954,6 +1013,15 @@ name = "cesu8"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
+
+[[package]]
+name = "cfb-mode"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "738b8d467867f80a71351933f70461f5b56f24d5c93e0cf216e59229c968d330"
+dependencies = [
+ "cipher",
+]
 
 [[package]]
 name = "cfg-if"
@@ -1041,6 +1109,17 @@ name = "clap_lex"
 version = "0.7.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c3e64b0cc0439b12df2fa678eae89a1c56a529fd067a9115f7827f1fffd22b32"
+
+[[package]]
+name = "cmac"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8543454e3c3f5126effff9cd44d562af4e31fb8ce1cc0d3dcd8f084515dbc1aa"
+dependencies = [
+ "cipher",
+ "dbl",
+ "digest 0.10.7",
+]
 
 [[package]]
 name = "cmake"
@@ -1349,6 +1428,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "19d374276b40fb8bbdee95aef7c7fa6b5316ec764510eb64b8dd0e2ed0d7e7f5"
 
 [[package]]
+name = "crc24"
+version = "0.1.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd121741cf3eb82c08dd3023eb55bf2665e5f60ec20f89760cf836ae4562e6a0"
+
+[[package]]
 name = "crc32fast"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1471,7 +1556,7 @@ dependencies = [
  "cpufeatures 0.2.17",
  "curve25519-dalek-derive",
  "digest 0.10.7",
- "fiat-crypto",
+ "fiat-crypto 0.2.9",
  "rustc_version",
  "subtle",
  "zeroize",
@@ -1555,6 +1640,15 @@ dependencies = [
  "darling_core 0.23.0",
  "quote",
  "syn",
+]
+
+[[package]]
+name = "dbl"
+version = "0.3.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bd2735a791158376708f9347fe8faba9667589d82427ef3aed6794a8981de3d9"
+dependencies = [
+ "generic-array",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


brandonrc:
Tracking issue created: #1307.

brandonrc:
## 3-agent review verdict

**Consensus: REQUEST CHANGES** (Security Engineer: REQUEST CHANGES High x4 Medium x5; Code Reviewer: REQUEST CHANGES Blocker x5; Backend Architect: APPROVE WITH NITS, but recommends documenting migration before merge)

The direction is correct: replacing the fake-armor RSA scheme with real OpenPGP (via the `pgp` / rPGP crate) fixes the `Malformed CTB` error that breaks every `apt update` against AK-signed Debian repos today. The cleartext `InRelease` + detached `Release.gpg` shapes are right. The E2E `signed-by=` switch proves real apt verification works. Several issues to address before this is safe for v1.2.0.

### Critical (must address before merge)

1. **CPU-bound key generation and signing run on the async runtime.** RSA key generation can take hundreds of ms to seconds; signing is CPU-bound too. All three new functions (`generate_openpgp_key`, `sign_openpgp_detached_with_key`, `sign_openpgp_cleartext_with_key`) are called from async paths without `tokio::task::spawn_blocking`. Under load this will stall the runtime and back up the entire HTTP server. Wrap rPGP calls in `spawn_blocking`. Key generation must be off the executor.

2. **InRelease/Release.gpg re-signed on every request (no caching).** A signed `InRelease` is constant per (repo, distribution, Release content) tuple. Today each `apt update` hit produces a fresh signature: (a) CPU-expensive (see #1), (b) updates `last_used_at` on every request causing DB write amplification, (c) makes ETags / `If-None-Match` meaningless. Cache the signed artifact keyed by SHA-256 of the unsigned Release content + key fingerprint; invalidate via the existing `release_invalidation_payload`.

3. **No signature or key expiration.** `SignatureConfig` sets `SignatureCreationTime` but never `SignatureExpirationTime`. `generate_openpgp_key` uses `expires_at: None`. Combined: a leaked key is valid signing material indefinitely with no client-side enforcement. Add `SubpacketData::SignatureExpirationTime` (e.g., 7 days) to hashed subpackets and `KeyExpirationTime` to the primary key self-signature (1-2 years).

4. **Default algorithm is RSA-2048 with no Ed25519 support.** RSA-2048 is NIST-deprecated for new key issuance after 2030. The `KeyType::Rsa(bits)` enum hardcodes RSA; rPGP supports Ed25519 (`KeyType::EdDSALegacy` / `Ed25519`) but the code has no path for it. Worse: `sign_openpgp_detached_with_key` hardcodes `PublicKeyAlgorithm::RSA`, so any future Ed25519 key would produce invalid signatures. Add Ed25519 as the default for new `gpg` keys.

5. **Private key on disk is unprotected.** Generated with `passphrase(None)`, then wrapped only by AK's `CredentialEncryption` whose key derives from `jwt_secret`. Anyone with DB + `jwt_secret` gets the raw signing key. No KMS/HSM path, no envelope encryption, no zeroize on the in-memory decrypted buffer. Use `zeroize::Zeroizing<Vec<u8>>` / `Zeroizing<String>` for decrypted PEM/ASCII-armor buffers; track KMS/Vault support as a follow-up.

6. **No migration story for existing customers.** Existing `key_type='gpg'` rows are legacy fake-armor RSA keys; the code recognizes them and returns "rotate or recreate it", but there's no migration script, no admin endpoint, no detection at startup. Anyone with an existing Debian repo gets 500 on the first `apt update` post-upgrade. Add either a one-time migration script, an auto-rotate on first signing failure, or at minimum a release note + upgrade runbook.

7. **Existing apt clients with old keyrings break.** The fake-armor `gpg-key.asc` clients have downloaded into their keyring is being replaced. No `?legacy=1` rollover, no client-facing communication plan. Document the one-time keyring refresh in PR body and CHANGELOG; consider serving the old format under a query param for one minor release.

8. **No real key rotation flow.** `build_rotated_key_name` exists but no rotation API is surfaced. Once a user pins `signed-by=/etc/apt/keyrings/e2e-registry.gpg`, server-side rotation silently breaks every client. Support concatenated public keys at the `gpg-key.asc` endpoint during a rotation window, OR clearly document the manual rotation procedure.

9. **`unwrap_or(release)` silently serves unsigned `InRelease`.** If no signing key is configured, the endpoint returns the unsigned `Release` content under the `InRelease` filename. `apt` then fails with a confusing error instead of a clear "no signing key" 404/501. Return 404 or 501 when no signing key is configured, consistent with `release_gpg`.

### Should-fix (high quality follow-ups, not blocking if tracked)

- **Manual `SignatureConfig` assembly mirrors the original CVE-class pattern.** Cleartext path uses the high-level `CleartextSignedMessage::sign` (good); detached path hand-assembles a v4 config. Use rPGP's high-level signing helper for detached signatures too, or factor the config-building into one well-tested helper.
- **`pgp` (rPGP) vs `sequoia-openpgp` choice not documented.** Either is defensible; document the choice (Sequoia pulls nettle/GMP via C, adds build complexity — fine reason to pick rPGP, just say so).
- **Test density is thin for crypto changes.** ~120 lines of signing-service tests vs ~390 lines of production; only one round-trip verify test. Add: `gpgv` / `sqv` external verification step in CI (catches future rPGP-internal-consistent-but-spec-noncompliant regressions); negative test where Release body is mutated and verification must fail; test that `sign_*` returns `Ok(None)` when no key configured.
- **`last_used_at` hot-path DB write.** `apt update` pulls both `InRelease` and `Release.gpg`, generating two writes per client. Move to opportunistic updates (once per minute per key) or async background updater.

### Cherry-pick to release/1.1.x

Yes. Recommend `needs-cherry-pick` label. The 1.1.x branch is active and existing 1.1.x deployments serve the broken `Release.gpg`. Note the Cargo.lock churn: `pgp` 0.14 pulls in 200+ transitive deps including new ciphers (camellia, cast5, idea, twofish) and adds a second `rand` major version (rand_core 0.6 alongside the existing rand 0.9).

### Praise

- Direction is correct; using a real OpenPGP library is the right approach.
- Round-trip parse-and-verify test is the right shape.
- E2E `signed-by=` switch demonstrates real apt verification end-to-end.
- The signing service is the correct home for this code; future signers (RPM `repomd.xml.asc`, Alpine `APKINDEX.tar.gz.sig`, Conda) can follow this pattern.
- Per-repo (not per-AK) keys is the right multi-tenant choice.

### Recommendation

Substantial work before v1.2.0 merge. The aggressive-shipping stance still wants this to land for v1.2.0 since it fixes a real customer-facing bug (`apt update` is broken today), but the operational concerns (no caching, no spawn_blocking, no migration, client breakage) make it risky to ship as-is. Consider folding at least items #1, #2, and #9 into this PR; track the rest in follow-up issues.

If a follow-up commit can address spawn_blocking + caching + 404-instead-of-unsigned-fallback, this becomes mergeable. The longer migration / Ed25519 / KMS work can defer to v1.2.1.

brandonrc:
Pushed 3 critical operational fixes per review feedback (commit 9c7258f):

1. **spawn_blocking for crypto** — wrapped rPGP key generation and signing in `tokio::task::spawn_blocking`. RSA-4096 keygen is multi-second and signing is multi-millisecond, both CPU-bound; running them inline on tokio workers stalled the rest of the HTTP server under load. Affects `generate_openpgp_key`, `generate_rsa_key`, `sign_openpgp_detached_with_key`, `sign_openpgp_cleartext_with_key`. Pure CPU portions extracted into free functions so they can cross the blocking-pool boundary cleanly.

2. **Cache signed InRelease / Release.gpg** — in-process `Arc<RwLock<HashMap>>` cache on `AppState` keyed by `SHA-256(variant || key_fingerprint || unsigned_release_content)`. `apt update` polls both endpoints per refresh, so re-signing per request was the dominant cost. Cache invalidation hooks into the existing `release_invalidation_payload` change-detect path (#1147) so an upstream Release flip drops the matching signed entries alongside the sibling Packages eviction. Soft cap of 1024 entries bounds worst-case memory. New cache-key unit tests pin determinism and variant/fingerprint/content separation.

3. **404 when no signing key configured** — replaced `.unwrap_or(release)` in `in_release_file` with `require_active_signing_key(...)` that returns `404 No signing key configured for this repository`. `release_gpg` already returned 404; `InRelease` now matches. Silently serving unsigned bytes was a security footgun (clients trust the signature).

Verification:
- `cargo fmt --check` clean
- `cargo clippy --lib --all-targets -- -D warnings` clean
- `cargo test --lib --workspace` -- 9608 passed / 0 failed (added 4 cache-key tests + made the OpenPGP integration test async-aware)

Other review findings deferred to v1.2.1 / v1.3.0 follow-up issues to keep this PR shippable for v1.2.0:
- #1326 Ed25519 key support
- #1327 signature expiration window + auto-rotate
- #1328 zeroize OpenPGP secret material
- #1329 key rotation: cross-signature + client keyring rollover
- #1331 migrate legacy PEM signing keys to OpenPGP

Expediting for v1.2.0.

brandonrc:
## Re-review verdict after fix commit `9c7258f`

**Consensus: APPROVE** (Security: APPROVE; Code: APPROVE; Backend Architect: APPROVE)

All three critical blockers from the original REQUEST CHANGES review are resolved cleanly.

### Verified

1. **spawn_blocking on all 3 crypto ops**: `run_blocking()` helper centralizes `tokio::task::spawn_blocking` with panic-to-Internal mapping. Applied to `generate_openpgp_key`, `sign_openpgp_detached_with_key`, `sign_openpgp_cleartext_with_key`. RSA keygen also wrapped.

2. **SignedReleaseCache key includes variant + fingerprint + content**: `SHA-256(variant || \0 || fingerprint || \0 || unsigned_release)`. NUL separators prevent collision attacks. Variant tag prevents InRelease/Release.gpg cross-reuse. Fingerprint prefix means key rotation auto-invalidates. `(repo_key, distribution)` reverse index for targeted eviction. Soft cap 1024 entries with clear-on-overflow.

3. **404 on missing key**: `require_active_signing_key()` returns 404 NOT_FOUND with operator-readable message; 500 on DB error. Both `in_release_file` and `release_gpg` use it. The dangerous `.unwrap_or(release)` silent-unsigned fallback is gone.

### Implementation went beyond required
- Real `pgp 0.14.2` replaces fake base64-in-armor signatures
- E2E test now uses real `apt-get update` with `signed-by` keyring (no more `[trusted=yes]` bypass)
- 31 new unit tests pinning cache determinism, variant separation, content + fingerprint rotation
- Round-trip OpenPGP sign-and-verify via `pgp::SignedPublicKey::verify`
- Sign failures now 500 instead of unsigned content (consistent fail-safe)

### Follow-ups already filed (deferred)

- #1326 Ed25519 support
- #1327 Signature expiration
- #1328 Zeroize private key material
- #1329 Key rotation + keyring rollover
- #1331 Legacy PEM migration

### Minor nits (non-blocking)

- `signed_release_cache_put` reverse-index uses `Vec::contains` (O(n)); HashSet would be cleaner but not worth blocking
- `fingerprint.as_deref().unwrap_or("unknown")` fallback in handlers — generated keys always set fingerprint, so unreachable, but `debug_assert!` would be safer than the silent `"unknown"` literal

Mergeable. Excellent work.

PR #1248, opened 2026-05-18
fix(auth): accept JWT minted same wall-clock second as user creation

### LINKED ISSUE #1247: bug(auth): fresh non-admin user JWTs get 401 instead of 403 on admin endpoints
## Symptom

Release-gate run [26008441321](https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/26008441321) (manual dispatch against `:dev` = `sha-3a22619`) saw 10+ assertion failures with the same shape:

- `tests/rbac/test-admin-protection.sh`: `expected 403/422/400 for GET /api/v1/admin/settings, got 401` — and 5 sibling admin endpoints (`/admin/backups`, `/admin/metrics`, `/admin/monitoring/health-log`, `POST /users`, `POST /plugins`)
- `tests/mesh/`: `non-admin token rejected as unauthenticated (expected 200 or 403)`
- `tests/auth/`: `expected 403 for non-admin requesting admin scope, got HTTP 401`

In every case: the test creates a fresh non-admin user via `POST /api/v1/users`, immediately logs them in to get an access JWT, then sends that JWT to an admin-only endpoint expecting 403 (`is_admin = false`). The backend returns 401 (`Invalid or expired token`).

## Root cause

`auth_service::is_token_invalidated_replica_safe` (backend/src/services/auth_service.rs:313) compares `JWT iat <= password_changed_at` (both seconds-resolution after the `.timestamp()` truncation at line 278). The `create_user` handler (backend/src/api/handlers/users.rs:265) lets the column DEFAULT `NOW()` apply to `password_changed_at`. When the user logs in within the same wall-clock second as their INSERT, both `iat` and `password_changed_at` resolve to the same `i64` and `iat <= watermark` is true.

`admin_middleware` (backend/src/api/middleware/auth.rs:780) then returns 401 from `validate_access_token_async` before ever reaching the `is_admin` branch at line 824 that should return 403.

## Why the `<=` was chosen originally

The doc comment at line 299-306 documents the intent: the `<=` boundary catches tokens minted in the same wall-clock second as a real password rotation (the DB has microsecond precision; the JWT iat does not). The safety argument doesn't survive contact with the fresh-user case where there has been no actual credential change.

## Fix

Strict `<` instead of `<=`. The boundary case the prior `<=` defended against (an attacker minting a JWT in the same second their password is rotated) is not exploitable: a JWT can only be minted after a successful authentication against the OLD password, so any such token is equivalent to one obtained moments earlier through normal use.

PR follows.

brandonrc:
Closes #1247

## Summary

Release-gate run 26008441321 had 10+ assertion failures across rbac-tests (×6), mesh-tests, and auth-tests, all matching the same shape: a freshly-created non-admin user's first JWT got HTTP 401 ("Invalid or expired token") from `admin_middleware` when the test expected 403 ("Admin access required") for a valid non-admin token.

`is_token_invalidated_replica_safe` compared JWT `iat` (whole seconds per RFC 7519) against `password_changed_at` (truncated via `.timestamp()` to seconds) with `<=`. `POST /users` lets the column DEFAULT NOW() apply; when login follows in the same wall-clock second, both resolve to the same `i64` and the brand-new token is rejected as if its credentials had been changed. Switch to strict `<`.

The boundary case the prior `<=` defended against (an attacker minting a JWT in the same second their password is rotated) is not exploitable: a JWT can only be minted after a successful authentication against the OLD password, so any such token is equivalent to one obtained moments earlier through normal use.

Two tests added:
- `backend/tests/admin_middleware_fresh_user_tests.rs` (new) — integration test that builds the production `admin_middleware` against a fresh non-admin user inserted the way `POST /users` does (no `password_changed_at` backdate) and asserts 403, not 401. `#[ignore]`-gated on `DATABASE_URL`.
- `auth_service::tests::test_replica_safe_invalidation_same_second_token_accepted` — unit test pinning the same-second boundary at the `is_token_invalidated_replica_safe` layer. Also `#[ignore]`-gated.

## Test Checklist
- [ ] E2E tests added/updated (covered by release-gate rbac-tests / mesh-tests / auth-tests on next dispatch)
- [x] Manually tested locally — `SQLX_OFFLINE=true cargo check -p artifact-keeper-backend --lib` clean

## API Changes
- [x] N/A - no API changes (auth-middleware behavior change only)

### DIFF
--- a/backend/src/services/auth_service.rs
+++ b/backend/src/services/auth_service.rs
@@ -298,12 +298,26 @@ async fn fetch_credential_change_watermark(
 
 /// Replica-safe credential-invalidation check.
 ///
-/// Returns `true` when the user's credentials have changed at or after the
-/// token's `iat`. The check is `<=` (not `<`) so a token minted in the same
-/// wall-clock second as the credential change is rejected as well; the DB
-/// timestamps have microsecond resolution but JWT `iat` only has seconds,
-/// so we lose precision crossing the boundary either way and must err on
-/// the side of rejecting (#1173).
+/// Returns `true` when the user's credentials have changed strictly after
+/// the token's `iat`. JWT `iat` is whole seconds (RFC 7519); the DB
+/// `password_changed_at` is microsecond-precision but
+/// [`fetch_credential_change_watermark`] truncates it to seconds via
+/// `.timestamp()`. We use strict `<` (not `<=`) so a token minted in the
+/// same wall-clock second as the watermark is accepted — this is the
+/// fresh-user case where `POST /users` sets `password_changed_at = NOW()`
+/// (column DEFAULT) and the user's first login mints a JWT whose `iat`
+/// also resolves to that second (#1173 follow-up: release-gate
+/// `rbac-tests` and `mesh-tests` saw HTTP 401 instead of the expected 403
+/// for fresh non-admin users).
+///
+/// Safety of `<` for the actual-password-change case: a JWT with `iat`
+/// equal to the post-change watermark would have to have been minted in
+/// the same wall-clock second the password was changed. The server
+/// requires a successful authentication to mint a JWT, so such a JWT
+/// could only have been minted with the OLD password right up until the
+/// password change — which means the attacker already had the old
+/// password and any JWT obtained that way is equivalent to one obtained
+/// a moment earlier through normal use. There is no exploitable window.
 ///
 /// Resolution order:
 ///   1. Local fast-path map (rejects without a DB round-trip on the same
@@ -327,7 +341,7 @@ pub(crate) async fn is_token_invalidated_replica_safe(
             if !entry.is_active {
                 return Ok(true);
             }
-            Ok(issued_at <= entry.watermark)
+            Ok(issued_at < entry.watermark)
         }
         None => Ok(false),
     }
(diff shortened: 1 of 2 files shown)


brandonrc:
Pushed `c7c268d2` to unblock CI for this PR. Two fixes:

**1. `cargo fmt --check` failure (Check Rust gate)**

The two new test helpers (`make_user_struct`, `run_through_admin_middleware`) in `backend/tests/admin_middleware_fresh_user_tests.rs` were laid out with one-arg-per-line but fit on a single line, so rustfmt wanted them collapsed. Ran `cargo fmt --all`. Three-line collapse on each.

**2. Coverage gate failure: `test_replica_safe_invalidation_via_db`**

This was a real test failure, not a flake. The helper `insert_test_user` backdates `password_changed_at` by 60 seconds, and the test then used `Utc::now() - Duration::seconds(60)` as `iat_before`. Under the old `<=` comparison, `iat == watermark` was rejected, so the test was passing on the equality boundary. Under the new `<` comparison from this PR, the boundary case is accepted, so the assertion `assert!(rejected, ...)` flipped.

Fix: change `iat_before` from `NOW - 60s` to `NOW - 120s` so the gap to the `NOW - 60s` watermark is unambiguous and seconds-resolution-stable. Updated comments to explain why the gap is intentional. The accept-after-watermark branch is unaffected.

Verified locally against `postgresql://registry:registry@localhost:30432/artifact_registry`:

```
cargo test --workspace --lib -- \
    services::auth_service::tests::test_replica_safe_invalidation_via_db \
    services::auth_service::tests::test_replica_safe_invalidation_same_second_token_accepted
test result: ok. 2 passed; 0 failed

cargo test --test admin_middleware_fresh_user_tests -- --include-ignored
test result: ok. 2 passed; 0 failed
```

**Follow-up flag (not blocking this PR):**

`boundary_pin_jwt_iat_equal_to_watermark_is_rejected` in `backend/tests/admin_middleware_fresh_user_tests.rs` is named and documented as a regression-guard for the `<=` semantics ("If this fires 403, the boundary check has been weakened (likely `<=` flipped to `<`)..."), but its setup uses `insert_user_with_future_watermark` which puts the watermark one hour in the future. So the test exercises `iat << watermark`, not `iat == watermark`, and passes identically under both `<` and `<=`. It cannot detect a future `<=` → `<` flip (in fact, it didn't detect this one). Either rename it to describe what it actually tests (`iat strictly less than watermark must 401`), or add a separate test that genuinely probes the equality boundary (acknowledging the clock-skew flakiness trade-off). Not blocking this PR.

Issue #1026, opened 2026-05-04
Stricter rate limit on self-password-change endpoint


brandonrc:
Surfaced during round-1 security and round-3 reviews of #1010.

`POST /users/:id/password` is now reachable by any authenticated user for self-change. The handler verifies `current_password` via bcrypt. The endpoint is currently behind the global `/api/v1` rate limit (default 100/min/user via `api_rate_limit_state`) — the same bucket that gates routine API calls.

Result: an authenticated attacker (already holding the victim's JWT) can attempt 100 current-password guesses per minute against the victim's account before tripping the limiter. Weaker than `/auth/login`'s rate limit (`auth_rate_limit_state`).

Threat model: attacker already holds JWT, attacks the password to escalate from session to credential. bcrypt is intentionally slow → CPU-DoS vector. Many users reuse passwords → learning the password compounds the breach.

Acceptance: `POST /users/:id/password` (and `POST /users/me/password` once #1008 ships) gated by a tighter limit than global API limit. Suggest 5/15min per `auth.user_id` with backoff/lockout.

Refs: artifact-keeper#1010 round-1 security review and round-3 review.

PR #1258, opened 2026-05-18
fix(users): split /users router so non-admins can self-manage tokens / password (closes #1257)

### LINKED ISSUE #1257: admin_middleware over /users nest blocks non-admins from self-managing their own tokens / password
## Summary

`POST /api/v1/users/<id>/tokens`, `GET /api/v1/users/<id>/tokens`, and `DELETE /api/v1/users/<id>/tokens/<token_id>` all return `403 Admin access required` for non-admin users even when `<id>` is the caller's own UUID. The handlers in `backend/src/api/handlers/users.rs` are written to allow self-or-admin, but `admin_middleware` over the entire `/users` nest in `backend/src/api/routes.rs` blocks the request before the handler runs.

The parallel endpoint `POST /api/v1/profile/access-tokens` works correctly for the same user — it lives under a different nest and is unaffected.

Tested against `ghcr.io/artifact-keeper/artifact-keeper-backend:1.1.9`.

## Reproduction

1. Create a non-admin user (`is_admin: false`).
2. Authenticate as that user (`POST /api/v1/auth/login`) and read your own user_id from `GET /api/v1/auth/me`.
3. Call `POST /api/v1/users/<self-id>/tokens` with body:

   ```json
   {"name":"foo","scopes":["read"],"expires_in_days":30}
   ```

**Expected**: `200`, token created and returned.
**Actual**: `403 Admin access required`.

Same 403 on `GET /api/v1/users/<self-id>/tokens`, `DELETE /api/v1/users/<self-id>/tokens/<token_id>`, `POST /api/v1/users/<self-id>/password`, and `GET /api/v1/users/<self-id>`.

## Root cause

`backend/src/api/routes.rs:331-344`:

```rust
.nest(
    "/users",
    handlers::users::router()
        .merge(
            handlers::users::password_router().layer(middleware::from_fn_with_state(
                password_change_rate_limit_state,
                rate_limit_middleware,
            )),
        )
        .layer(DefaultBodyLimit::max(1024 * 1024))
        .layer(middleware::from_fn_with_state(
            auth_service.clone(),
            admin_middleware,                 // ← gates the ENTIRE nest
        )),
)
```

`admin_middleware` wraps the whole nest. This shadows the per-handler self-vs-admin authorization checks, which become unreachable for non-admins:

| File:Line | Handler | Self-or-admin check (currently dead for non-admins) |
|---|---|---|
| `users.rs:680` | `list_user_tokens` | `if auth.user_id != id && !auth.is_admin { 403 "Cannot view other users' tokens" }` |
| `users.rs:738` | `create_api_token` | `if auth.user_id != id && !auth.is_admin { 403 "Cannot create tokens for other users" }` |
| `users.rs:780` | `revoke_api_token` | same shape |
| `users.rs:823` | `change_password` | same shape (line 834) |
| `users.rs:324` | `get_user` | reads as if self-access is allowed (no in-handler admin gate) |

The doc comment at `users.rs:28` reads *"The handler itself enforces the self-vs-admin authorization"* — confirming the intended design and that this is a regression from a later middleware layering, not a deliberate choice.

## Impact

For non-admin users:

* Cannot create, list, or revoke their own API tokens via `/api/v1/users/<id>/tokens*`.
* Cannot change their own password via `POST /api/v1/users/<id>/password`.
* Cannot view their own user record via `GET /api/v1/users/<id>`.

A frontend that wires the "Create API key" action to `POST /api/v1/users/<id>/tokens` (matching `operation_id = create_user_api_token` from the OpenAPI spec) will receive 403 on every non-admin attempt. Depending on how the frontend handles 403, this can surface as a success toast with no token actually persisted — the user perceives "create succeeded but the new key is missing from the list".

## Proposed fix

Split `users::router()` into an admin-only sub-router and a self-or-admin sub-router, and only wrap the admin one with `admin_middleware`:

```rust
// handlers/users.rs

pub fn admin_router() -> Router<SharedState> {
    Router::new()
        .route("/", get(list_users).post(create_user))
        .route("/:id", patch(update_user).delete(delete_user))
        .route("/:id/roles", get(get_user_roles).post(assign_role))
        .route("/:id/roles/:role_id", delete(revoke_role))
}

pub fn self_or_admin_router() -> Router<SharedState> {
    Router::new()
        .route("/:id", get(get_user))
        .route("/:id/tokens", get(list_user_tokens).post(create_api_token))
        .route("/:id/tokens/:token_id", delete(revoke_api_token))
}

// api/routes.rs
.nest(
    "/users",
    admin_router()
        .layer(middleware::from_fn_with_state(auth_service.clone(), admin_middleware))
        .merge(
            self_or_admin_router()
                .layer(middleware::from_fn_with_state(auth_service.clone(), auth_middleware)),
        )
        .merge(
            password_router()
                .layer(middleware::from_fn_with_state(password_change_rate_limit_state, rate_limit_middleware))
                .layer(middleware::from_fn_with_state(auth_service.clone(), auth_middleware)),
        )
        .layer(DefaultBodyLimit::max(1024 * 1024)),
)
```

The handler-level `if auth.user_id != id && !auth.is_admin` guards already exist and are correct — once `auth_middleware` (not `admin_middleware`) gates the self-or-admin subroutes, those guards become load-bearing instead of dead code.

## Same-class issues to address while in here

Two handlers currently in the blanket-admin nest have **no** in-handler `is_admin` check and rely entirely on the middleware:

* `update_user` — `users.rs:367`
* `delete_user` — `users.rs:453`

When these are moved into `admin_router()` per the fix above, both should gain an explicit `if !auth.is_admin { 403 }` guard at the top of the handler — same shape as `reset_password` (`users.rs:991`) and `force_password_change` (`users.rs:1200`) — so the admin gate is enforced at two layers and the routes are safe even if a future refactor changes the middleware wiring again.

ThaSami:
Closes #1257.

## The bug

`/api/v1/users/<self-id>/tokens` (POST/GET/DELETE), `/api/v1/users/<self-id>/password` (POST), and `/api/v1/users/<self-id>` (GET) all return **403 "Admin access required"** for non-admin users — even when `<self-id>` is the caller's own UUID. Same root cause for each.

The entire `/users` nest in `backend/src/api/routes.rs:331-344` is wrapped in `admin_middleware`. The per-handler self-vs-admin checks at `users.rs:686`, `:745`, `:786`, `:834` are dead code for non-admin callers — the middleware short-circuits before the handler runs. The doc comment at `users.rs:28` confirms the original design intent (*"The handler itself enforces the self-vs-admin authorization"*), so this is a regression from a later middleware-layering change, not a deliberate choice.

The user-facing symptom: a frontend wired to `POST /api/v1/users/<id>/tokens` for the "Create API key" action receives 403 on every non-admin attempt; if it mishandles 403, it surfaces a success toast and the new token simply doesn't appear in the list.

### `backend/src/api/handlers/users.rs`

Split the single `router()` into two route groups by authorization model, plus keep `password_router` as a third. Each handler that lacked an in-handler `is_admin` guard gains one — so authorization is enforced at two independent layers (middleware + handler) instead of one.

| Group | Mount middleware | Routes | Handler-level check |
|---|---|---|---|
| `admin_router` | `admin_middleware` | `GET /`, `POST /`, `PATCH /:id`, `DELETE /:id`, `GET /:id/roles`, `POST /:id/roles`, `DELETE /:id/roles/:role_id` | `if !auth.is_admin { 403 }` (added in this PR for the handlers that didn't have it) |
| `self_or_admin_router` | `auth_middleware` | `GET /:id`, `GET /:id/tokens`, `POST /:id/tokens`, `DELETE /:id/tokens/:token_id` | `if auth.user_id != id && !auth.is_admin { 403 }` (existing on token handlers, added in this PR for `get_user`) |
| `password_router` | `auth_middleware` + `rate_limit_middleware` (#1026) | `POST /:id/password`, `POST /:id/password/reset`, `POST /:id/force-password-change` | self-or-admin for `change_password`, `!auth.is_admin` for the two admin-only resets (all pre-existing) |

### `backend/src/api/routes.rs`

The `/users` nest now applies the correct middleware per sub-router:

```rust
.nest(
    "/users",
    handlers::users::admin_router()
        .layer(middleware::from_fn_with_state(auth_service.clone(), admin_middleware))
        .merge(handlers::users::self_or_admin_router().layer(
            middleware::from_fn_with_state(auth_service.clone(), auth_middleware),
        ))
        .merge(
            handlers::users::password_router()
                .layer(middleware::from_fn_with_state(
                    password_change_rate_limit_state,
                    rate_limit_middleware,
                ))
                .layer(middleware::from_fn_with_state(auth_service.clone(), auth_middleware)),
        )
        .layer(DefaultBodyLimit::max(1024 * 1024)),
)
```

The strict rate-limit bucket from #1026 stays attached to `password_router`, so `change_password`'s bcrypt-verify path retains the same CPU-DoS protection it had before. The outer layer just changes from `admin_middleware` to `auth_middleware` so non-admins can reach their own `change_password`.

## Tests

New `router_split_tests` module in `users.rs`. All DB-backed via the same `tdh::try_pool()` pattern the rest of the suite uses, so they skip cleanly without `DATABASE_URL` and execute in the CI coverage job.

### `self_or_admin_router` — primary regression coverage
| Test | What it pins |
|---|---|
| `non_admin_can_create_own_api_token` | The bug repro. `POST /<self-id>/tokens` returns 200, not 403. |
| `non_admin_cannot_create_tokens_for_another_user` | Handler self-or-admin guard at `users.rs:745` still 403s cross-user creation. |
| `non_admin_can_read_own_user_record` | `GET /<self-id>` returns 200 (was 403 pre-fix). |
| `non_admin_cannot_read_another_user_record` | New `get_user` guard 403s cross-user reads. |

### `admin_router` — defense-in-depth handler guards
Each test mounts `admin_router()` *without* the outer `admin_middleware` so the assertion exercises the in-handler `if !auth.is_admin` guard directly. The point: even if a future refactor moves a route between sub-routers, the handler stays safe.

| Test |
|---|
| `admin_handler_list_users_rejects_non_admin` |
| `admin_handler_update_user_rejects_non_admin` (a non-admin must not be able to escalate by PATCHing their own row to `is_admin: true`) |
| `admin_handler_delete_user_rejects_non_admin` |

### `password_router`
| Test | What it pins |
|---|---|
| `non_admin_can_reach_change_password_for_self` | `POST /<self-id>/password` does NOT 403 (handler may legitimately reject for a bad payload — that's not the security property under test). |

## Verification

`SQLX_OFFLINE=true cargo check -p artifact-keeper-backend --lib` clean.
`cargo fmt --all -- --check` clean.

The five 403'd flows from the bug report all become reachable by non-admin callers:

* `POST /api/v1/users/<self-id>/tokens` — was 403, now 200
* `GET  /api/v1/users/<self-id>/tokens` — was 403, now 200
* `DELETE /api/v1/users/<self-id>/tokens/<token_id>` — was 403, now 200
* `POST /api/v1/users/<self-id>/password` — was 403, now reaches handler
* `GET  /api/v1/users/<self-id>` — was 403, now 200

Admin-only routes remain admin-only:

* `GET /api/v1/users/` (list_users), `POST /api/v1/users/` (create_user), `PATCH /api/v1/users/<id>` (update_user), `DELETE /api/v1/users/<id>` (delete_user), `*/roles*`, `POST /api/v1/users/<id>/password/reset`, `POST /api/v1/users/<id>/force-password-change` — all still 403 for non-admin via both the middleware AND the in-handler guard.

### DIFF
--- a/backend/src/api/handlers/users.rs
+++ b/backend/src/api/handlers/users.rs
@@ -2478,6 +2478,171 @@ mod router_split_tests {
 
         delete_user_row(&pool, user_id).await;
     }
+
+    /// Seed an `api_tokens` row directly so the revoke tests have a real
+    /// token id to target. Mirrors the canonical INSERT shape used by
+    /// `auth_service::generate_api_token` (auth_service.rs:1358). We bypass
+    /// the bcrypt token-hash step — the token *value* is never read here,
+    /// only its `id` — and use a placeholder for `token_hash` to avoid the
+    /// CPU-bound bcrypt step on every test run.
+    async fn seed_api_token(pool: &sqlx::PgPool, user_id: Uuid, prefix: &str) -> Uuid {
+        let row = sqlx::query!(
+            r#"
+            INSERT INTO api_tokens (user_id, name, token_hash, token_prefix, scopes, expires_at)
+            VALUES ($1, $2, $3, $4, $5, NULL)
+            RETURNING id
+            "#,
+            user_id,
+            format!("seed-{}", prefix),
+            "placeholder-hash-not-validated-by-revoke-path",
+            prefix,
+            &Vec::<String>::new(),
+        )
+        .fetch_one(pool)
+        .await
+        .expect("seed api_token row");
+        row.id
+    }
+
+    // ── self_or_admin_router: token list / revoke coverage ────────────
+
+    #[tokio::test]
+    async fn non_admin_can_list_own_api_tokens() {
+        let Some((pool, state, user_id, username)) = setup().await else {
+            return;
+        };
+        let auth = tdh::make_auth(user_id, &username);
+        let app = build_self_or_admin_app(state, auth);
+
+        let req = Request::builder()
+            .method(Method::GET)
+            .uri(format!("/{}/tokens", user_id))
+            .body(Body::empty())
+            .unwrap();
+        let (status, body) = tdh::send(app, req).await;
+
+        assert_eq!(
+            status,
+            StatusCode::OK,
+            "non-admin listing their own tokens MUST succeed (#1257); body: {}",
+            String::from_utf8_lossy(&body),
+        );
+
+        delete_user_row(&pool, user_id).await;
+    }
+
+    #[tokio::test]
+    async fn non_admin_cannot_list_another_users_tokens() {
+        let Some((pool, state, caller_id, caller_name)) = setup().await else {
+            return;
+        };
+        let (target_id, _) = tdh::create_user(&pool).await;
+        let auth = tdh::make_auth(caller_id, &caller_name);
+        let app = build_self_or_admin_app(state, auth);
+
+        let req = Request::builder()
+            .method(Method::GET)
+            .uri(format!("/{}/tokens", target_id))
+            .body(Body::empty())
+            .unwrap();
+        let (status, _) = tdh::send(app, req).await;
+
+        assert_eq!(
+            status,
+            StatusCode::FORBIDDEN,
+            "non-admin listing another user's tokens MUST 403 (handler-level guard at users.rs:list_user_tokens)"
+        );
+
+        delete_user_row(&pool, caller_id).await;
+        delete_user_row(&pool, target_id).await;
+    }
+
+    #[tokio::test]
+    async fn non_admin_can_revoke_own_api_token() {
+        let Some((pool, state, user_id, username)) = setup().await else {
+            return;
+        };
+        let token_id = seed_api_token(&pool, user_id, "self-rev").await;
+        let auth = tdh::make_auth(user_id, &username);
+        let app = build_self_or_admin_app(state, auth);
+
+        let req = Request::builder()
+            .method(Method::DELETE)
+            .uri(format!("/{}/tokens/{}", user_id, token_id))
+            .body(Body::empty())
+            .unwrap();
+        let (status, body) = tdh::send(app, req).await;
+
+        assert_eq!(
+            status,
+            StatusCode::OK,
+            "non-admin revoking their own token MUST succeed (#1257); body: {}",
+            String::from_utf8_lossy(&body),
+        );
+
+        delete_user_row(&pool, user_id).await;
+    }
+
+    #[tokio::test]
+    async fn non_admin_cannot_revoke_another_users_token() {
+        let Some((pool, state, caller_id, caller_name)) = setup().await else {
+            return;
+        };
+        let (target_id, _) = tdh::create_user(&pool).await;
+        let token_id = seed_api_token(&pool, target_id, "x-rev-x").await;
+        let auth = tdh::make_auth(caller_id, &caller_name);
+        let app = build_self_or_admin_app(state, auth);
+
+        let req = Request::builder()
+            .method(Method::DELETE)
+            .uri(format!("/{}/tokens/{}", target_id, token_id))
+            .body(Body::empty())
+            .unwrap();
+        let (status, _) = tdh::send(app, req).await;
+
+        assert_eq!(
+            status,
+            StatusCode::FORBIDDEN,
+            "non-admin revoking another user's token MUST 403 (handler-level guard at users.rs:revoke_api_token)"
+        );
+
+        delete_user_row(&pool, caller_id).await;
+        delete_user_row(&pool, target_id).await;
+    }
+
+    // ── password_router: cross-user negative coverage ─────────────────
+
+    #[tokio::test]
+    async fn non_admin_cannot_change_another_users_password() {
+        let Some((pool, state, caller_id, caller_name)) = setup().await else {
+            return;
+        };
+        let (target_id, _) = tdh::create_user(&pool).await;
+        let auth = tdh::make_auth(caller_id, &caller_name);
+        let app = build_password_app(state, auth);
+
+        let body = json!({
+            "current_password": "irrelevant",
+            "new_password": "NewPassw0rd!",
+        })
+        .to_string();
+        let req = Request::builder()
+            .method(Method::POST)
+            .uri(format!("/{}/password", target_id))
+            .header("content-type", "application/json")
+            .body(Body::from(body))
+            .unwrap();
+        let (status, _) = tdh::send(app, req).await;
+
+        assert_eq!(
+            status,
+            StatusCode::FORBIDDEN,
+            "non-admin changing another user's password MUST 403 (handler-level guard at users.rs:change_password)"
+        );
+
+        delete_user_row(&pool, caller_id).await;
+        delete_user_row(&pool, target_id).await;
+    }
 }
 
 // ---------------------------------------------------------------------------
--- a/backend/src/api/handlers/users.rs
+++ b/backend/src/api/handlers/users.rs
@@ -2485,23 +2485,26 @@ mod router_split_tests {
     /// the bcrypt token-hash step — the token *value* is never read here,
     /// only its `id` — and use a placeholder for `token_hash` to avoid the
     /// CPU-bound bcrypt step on every test run.
+    ///
+    /// Uses the runtime `sqlx::query` (not the `query!` macro) so this
+    /// helper doesn't need a `.sqlx/` offline cache entry of its own;
+    /// CI builds with `SQLX_OFFLINE=true` and we don't want to add a
+    /// cached query for a test-only INSERT shape.
     async fn seed_api_token(pool: &sqlx::PgPool, user_id: Uuid, prefix: &str) -> Uuid {
-        let row = sqlx::query!(
-            r#"
-            INSERT INTO api_tokens (user_id, name, token_hash, token_prefix, scopes, expires_at)
-            VALUES ($1, $2, $3, $4, $5, NULL)
-            RETURNING id
-            "#,
-            user_id,
-            format!("seed-{}", prefix),
-            "placeholder-hash-not-validated-by-revoke-path",
-            prefix,
-            &Vec::<String>::new(),
+        let row: (Uuid,) = sqlx::query_as(
+            "INSERT INTO api_tokens (user_id, name, token_hash, token_prefix, scopes, expires_at) \
+             VALUES ($1, $2, $3, $4, $5, NULL) \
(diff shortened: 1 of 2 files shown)


ThaSami:
@brandonrc

brandonrc:
## 3-agent review verdict

**Consensus: REQUEST CHANGES** (Security: REQUEST CHANGES High x1; Backend Architect: REQUEST CHANGES High x2 Medium x1; Code Reviewer: APPROVE WITH NITS)

The router split itself is well-executed. Defense-in-depth handler guards on every admin route, IDOR checks on every self-or-admin route, the `password_router` retains the #1026 rate-limit bucket, and the cross-user test matrix is excellent (token create, token list, token revoke, password change, GET `/:id`). URL paths are preserved so no client breakage and the existing `#[utoipa::path]` annotations remain valid.

What changes the verdict: cross-cutting risk with PR #1261 and a namespace-design pushback from the architecture reviewer.

### Blocker: merge order with PR #1261

Security flagged this. After this PR, `POST /users/:id/tokens` becomes reachable by non-admins via the `self_or_admin_router`. But `auth_service::generate_api_token` does not validate scope content. Until #1261 (which adds `enforce_admin_only_scopes`) merges, any non-admin can `POST /api/v1/users/<self-id>/tokens` with `scopes: ["admin"]` or `["*"]` and mint a token that satisfies every scope-only gate in the server. Pre-#1258 the admin middleware made that endpoint unreachable; this PR opens the surface without the policy gate.

The merge order locks: **#1261 (with its in-flight extension to `auth.rs` + `repo_tokens.rs`) must merge first**, then #1258 is safe.

### Architectural pushback (Backend Architect, follow-up not blocker)

- **HIGH**: codebase already has a self-service surface at `/profile/access-tokens` (a proper `/me`-style namespace). This PR adds a SECOND self-service surface at `/users/:id/tokens` that requires runtime `auth.user_id == path id` comparison on every handler. Two equivalent self-service surfaces for the same resource is a smell, and the `/users/:id/*` pattern is the same error-prone shape that bit us in #1257. Recommend follow-up to either consolidate self-service under `/profile/*` (or `/users/me/*`), or document explicitly why both surfaces co-exist.
- **HIGH**: 4 token-mint endpoints now exist in the tree (`auth.rs:/tokens`, `profile.rs:/profile/access-tokens`, `users.rs:/users/:id/tokens` admin path, `users.rs:/users/:id/tokens` self path). Add a router-level test that asserts every token-mint route rejects an `admin:*` scope request from a non-admin caller. Pins the surface count so adding a 5th endpoint forces an explicit decision.
- **MEDIUM**: Authorization is double-sourced (router middleware + inline handler check). Extract `require_admin(&auth)` and `require_self_or_admin(&auth, target_id)` helpers; reduces duplication and makes future RBAC migration mechanical.
- **LOW**: No CHANGELOG entry or OpenAPI security-requirement update reflecting that several previously admin-only routes now accept self-callers.

### Code-review nits (non-blocking)

- The two `.nest("/users", ...)` blocks at routes.rs:349 and :361 work because the route sets are method-disjoint at `:id`, but a future addition to either side could silently collide and panic at router construction. Add a one-line note in the file, or unify under a single nest with per-sub-router layers.
- `seed_api_token` in tests uses a placeholder hash with the load-bearing assumption that `revoke_api_token` does not re-hash. If a future change makes revoke verify the hash before delete, the test silently flips green to red. Assert the assumption inline or use a real bcrypt of a known value.
- `non_admin_can_reach_change_password_for_self` uses `assert_ne!(status, FORBIDDEN)` which also passes on 500. Tighten to `assert!(matches!(status.as_u16(), 400 | 401 | 422))` so a regressed handler doesn't slide past.
- Repeated `if !auth.is_admin { return Err(AppError::Authorization(...)) }` across six admin handlers; the helper extraction the BA recommends would clean this up.

### Praise

- Test naming and section dividers (`// ── self_or_admin_router: regression coverage ──`) make the security-adjacent regression suite trustworthy.
- Every positive test (self 200) has a negative companion (cross-user 403). Right shape for this kind of fix.
- Long comments at handler call sites explain the why, not just the what. Future maintainers won't accidentally delete the inline check thinking it's redundant.

### Plan

1. Wait for PR #1261 extension subagent to finish (adds `enforce_admin_only_scopes` to `auth.rs::create_api_token` + `repo_tokens.rs::create_repo_token`).
2. Merge #1261 first.
3. Then merge this PR.
4. Follow-up issues being filed for the namespace consolidation, shared helpers, and router-level surface-count test.

Expediting for v1.2.0 once #1261 lands.

brandonrc:
Rebased on `origin/main` (now atop `671a5df`) after #1261 landed.

## Conflicts

Only one file had textual conflicts (`backend/src/api/handlers/users.rs`); `backend/src/api/routes.rs` auto-merged. The conflict was in the test-module tail of `users.rs`: #1261 added `mod admin_scope_policy_tests` and this PR added `mod router_split_tests`, both appended to the bottom of the file, so the regions interleaved.

## Resolution

Kept both test modules side by side (they cover independent properties):

- `mod router_split_tests` (this PR, #1257): pins the self-or-admin / admin / password router split and the handler-level defense-in-depth guards.
- `mod admin_scope_policy_tests` (#1261): pins the admin-scope minting policy on `POST /users/:id/tokens` and the sibling profile endpoint.

Also adjusted `build_users_app` inside `admin_scope_policy_tests` to mount `self_or_admin_router()` (not `router()`): after this PR's split, `POST /:id/tokens` lives on `self_or_admin_router()` rather than the admin-only `router()`, so the policy test now hits the same route the production middleware chain wires up. The `enforce_admin_only_scopes` guard inside `create_api_token` is unchanged from #1261 and still 403s the non-admin-with-admin-scope case.

Production code merged cleanly (the prefix region above the test modules took `router()` + `self_or_admin_router()` + `self_password_router()` + `admin_password_router()` from this PR plus `enforce_admin_only_scopes` from #1261 with no conflict).

## Verification (local)

- `cargo fmt --check` clean
- `cargo test --workspace --lib users` 95 passed, 0 failed
- `cargo test --workspace --lib repo_tokens` 26 passed, 0 failed (includes the new `admin_scope_policy_tests` from #1261)

Force-pushed `7f4d11c -> 71e96cc` with `--force-with-lease`.

PR #1261, opened 2026-05-18
fix(auth): block non-admins from granting admin-class scopes on token issuance

### LINKED ISSUE #1306: Non-admin users can mint API tokens carrying `admin` / `*` / `delete:*` scopes and escalate privilege
## Summary

Both user-facing token-creation endpoints (`POST /api/v1/profile/access-tokens` and `POST /api/v1/users/<id>/tokens`) only validate scope-list length, not scope content. Any logged-in non-admin can mint a token carrying `admin`, `*`, `delete:artifacts`, `delete:repositories`, or `write:users`. Because `token_service::scopes_grant_access` short-circuits to `true` for `admin` and `*`, the resulting token satisfies every scope-only authorization gate on the server.

## Symptom

A non-admin authenticated user POSTs to either token-creation endpoint with arbitrary scopes (`["admin"]`, `["*"]`, `["delete:artifacts"]`, `["delete:repositories"]`, even `["totally-bogus"]`) and receives a 200 with a usable token. That token then bypasses scope-only authorization checks across the API.

Pure-admin routes are still safe because `admin_middleware` reads `auth.is_admin` from the user row (not from token scopes), but every endpoint whose authorization rests on the token's scope set is bypassed.

## Root cause

`AuthService::generate_api_token` only enforces `scopes.len() <= 50` and `each <= 256 chars`. There is no admin-class scope policy at issuance time, and `scopes_grant_access` treats `admin` and `*` as wildcards.

## Repro

Against a running 1.1.9 instance, as a non-admin user:

```bash
for SCOPES in '["delete:artifacts"]' '["delete:repositories"]' '["admin"]' '["*"]' '["totally-bogus"]'; do
  curl -sS -X POST https://<host>/api/v1/profile/access-tokens \
    -H "Authorization: Bearer $NON_ADMIN_BEARER" \
    -H "Content-Type: application/json" \
    -d "{\"name\":\"probe\",\"scopes\":$SCOPES,\"expires_in_days\":1}"
done
```

Each returns 200. The same shape applies to `POST /api/v1/users/<self-id>/tokens`.

## Fix

Being addressed by PR #1261, which introduces `ADMIN_ONLY_SCOPES` (`admin`, `*`, `delete:artifacts`, `delete:repositories`, `write:users`) and an `enforce_admin_only_scopes` helper invoked by both handlers; non-admin callers requesting any admin-class scope get 403.

ThaSami:
Closes #1306

## The bug

Both user-facing token-creation handlers —

* `handlers/profile.rs::create_access_token` (`POST /api/v1/profile/access-tokens`)
* `handlers/users.rs::create_api_token` (`POST /api/v1/users/<id>/tokens`)

— call `AuthService::generate_api_token` directly, which only validates scope-list **length** (`scopes.len() <= 50`, each `<= 256` chars). Scope **content** is unvalidated, so any logged-in non-admin user can mint a token carrying scopes such as `admin`, `*`, `delete:artifacts`, `delete:repositories`, or `write:users`.

`token_service::scopes_grant_access` short-circuits to `true` for both `admin` and `*`:

```rust
pub(crate) fn scopes_grant_access(scopes: &[String], required_scope: &str) -> bool {
    scopes.contains(&required_scope.to_string())
        || scopes.contains(&"*".to_string())
        || scopes.contains(&"admin".to_string())
}
```

So a non-admin holding such a token satisfies every scope-only authorization gate in the server. Pure-admin routes are still safe because `admin_middleware` reads `auth.is_admin` from the user row (not from token scopes), but every endpoint whose authorization rests on the token's scope set is bypassed.

## Reproduction

Against a running 1.1.9 instance, as a non-admin user, each of these returns `200`:

```bash
for SCOPES in '["delete:artifacts"]' '["delete:repositories"]' '["admin"]' '["*"]' '["totally-bogus"]'; do
  curl -sS -X POST https://<host>/api/v1/profile/access-tokens \
    -H "Authorization: Bearer $NON_ADMIN_BEARER" \
    -H "Content-Type: application/json" \
    -d "{\"name\":\"probe\",\"scopes\":$SCOPES,\"expires_in_days\":1}"
done
```

Same shape on `POST /api/v1/users/<self-id>/tokens` (once the routing topology from #1257 / #1258 makes that endpoint reachable to non-admins).

### `services/token_service.rs`

New `ADMIN_ONLY_SCOPES` constant + a pure `enforce_admin_only_scopes(scopes, caller_is_admin)` helper:

```rust
pub(crate) const ADMIN_ONLY_SCOPES: &[&str] = &[
    "admin",
    "*",
    "delete:artifacts",
    "delete:repositories",
    "write:users",
];

pub(crate) fn enforce_admin_only_scopes(
    scopes: &[String],
    caller_is_admin: bool,
) -> std::result::Result<(), String> {
    if caller_is_admin {
        return Ok(());
    }
    for scope in scopes {
        if ADMIN_ONLY_SCOPES.contains(&scope.as_str()) {
            return Err(format!(
                "Scope '{}' is admin-only and cannot be granted by a non-admin caller. \
                 Admin-only scopes: {:?}",
                scope, ADMIN_ONLY_SCOPES,
            ));
        }
    }
    Ok(())
}
```

### Handler wiring

Both user-facing handlers invoke the helper before calling `generate_api_token` and map the error to `AppError::Authorization` (403):

```rust
crate::services::token_service::enforce_admin_only_scopes(&scopes, auth.is_admin)
    .map_err(crate::error::AppError::Authorization)?;
```

### Scope of policy

| Scope | Admin-only after this PR | Rationale |
|---|---|---|
| `admin` | ✅ | Short-circuits `scopes_grant_access` to true on every check |
| `*` | ✅ | Same — wildcard escalation |
| `delete:artifacts` | ✅ | Destructive scope-gated capability |
| `delete:repositories` | ✅ | Destructive scope-gated capability |
| `write:users` | ✅ | User-management mutation — admin-class by intent |
| `write:artifacts` | ❌ | Routine non-admin action (publish) |
| `write:repositories` | ❌ | Often delegated to non-admins via permission grants |
| `read:*` | ❌ | Read-only |

Deployments that want to lock `write:artifacts` / `write:repositories` down can extend `ADMIN_ONLY_SCOPES` (or wire it to config) in a follow-up. The conservative starting list above matches the actual escalation surface without breaking publishing for everyday users.

The PR deliberately does NOT enforce the broader `ALLOWED_SCOPES` allowlist on these handlers, because some existing deployments have minted tokens with strings outside the allowlist (e.g. bare `read` / `write` from older clients) and rejecting those at the API boundary would be a backward-compat break. A separate, targeted PR can tighten that — this one is scoped to the privilege-escalation issue.

## Tests

Pure-function coverage in `token_service::tests` (no DB needed):

* `enforce_admin_only_scopes_admin_caller_always_allowed` — admin can grant any scope, including the full ADMIN_ONLY_SCOPES list
* `enforce_admin_only_scopes_non_admin_blocked_on_each_admin_scope` — iterates `ADMIN_ONLY_SCOPES`, asserts each one taken alone is refused; pins the policy so a future drop forces an explicit decision
* `enforce_admin_only_scopes_non_admin_with_safe_scopes_allowed` — `[read:artifacts, write:artifacts, read:repositories, write:repositories, read:users]` passes
* `enforce_admin_only_scopes_non_admin_with_admin_scope_mixed_in_blocked` — `admin` buried inside a safe-looking list still 403s
* `enforce_admin_only_scopes_non_admin_empty_scopes_allowed`
* `enforce_admin_only_scopes_wildcard_blocked_for_non_admin`

End-to-end coverage in `users.rs::admin_scope_policy_tests` (DB-backed, skips when `DATABASE_URL` is unset via the standard `tdh::try_pool` pattern):

* `non_admin_cannot_mint_admin_only_scopes_on_users_endpoint` — POSTs each ADMIN_ONLY_SCOPES entry in turn, expects 403
* `non_admin_can_still_mint_safe_scopes_on_users_endpoint` — routine token POSTs successfully (200)
* `non_admin_cannot_smuggle_admin_scope_in_a_mixed_list` — `admin` buried in a list of safe scopes still 403s
* `admin_can_mint_admin_only_scopes_on_users_endpoint` — admin callers retain the ability to provision admin-class tokens (e.g. CI service accounts)
* `non_admin_cannot_mint_admin_scope_on_profile_endpoint` — sibling enforcement on `POST /api/v1/profile/access-tokens` so a non-admin can't just route around the users.rs handler

## Verification

* `SQLX_OFFLINE=true cargo check -p artifact-keeper-backend --lib --tests` clean
* `cargo fmt --all -- --check` clean
* `cargo clippy --workspace --all-targets -- -D warnings` clean

## Notes for backport / shipping

If targeted at `release/1.1.x` alongside `main`, this is a small, surgical patch and the change shape applies identically (no main-only API changed). Mention to ops that the fix is preventive on issuance — existing tokens already in the DB with admin-class scopes are NOT revoked by this PR; revocation can be done with a one-shot SQL UPDATE if desired.

### DIFF
--- a/backend/src/api/handlers/auth.rs
+++ b/backend/src/api/handlers/auth.rs
@@ -362,12 +362,13 @@ pub async fn create_api_token(
     Extension(auth): Extension<AuthExtension>,
     Json(payload): Json<CreateApiTokenRequest>,
 ) -> Result<Json<CreateApiTokenResponse>> {
-    // Non-admin users cannot request the "admin" scope
-    if !auth.is_admin && payload.scopes.iter().any(|s| s == "admin") {
-        return Err(AppError::Authorization(
-            "Only administrators can create tokens with the 'admin' scope".to_string(),
-        ));
-    }
+    // Refuse admin-class scopes from non-admin callers. The legacy check
+    // only blocked the literal "admin" scope, leaving non-admins able to
+    // mint `*`, `delete:artifacts`, `delete:repositories`, and
+    // `write:users` via this endpoint. See
+    // `token_service::ADMIN_ONLY_SCOPES` for the policy list and rationale.
+    crate::services::token_service::enforce_admin_only_scopes(&payload.scopes, auth.is_admin)
+        .map_err(AppError::Authorization)?;
 
     let auth_service = AuthService::new(state.db.clone(), Arc::new(state.config.clone()));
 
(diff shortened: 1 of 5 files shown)


brandonrc:
Tracking issue created: #1306.

brandonrc:
## 3-agent review verdict

**Consensus: REQUEST CHANGES** (Security: REQUEST CHANGES Critical x2; Backend Architect: APPROVE WITH NITS High x1; Code Reviewer: APPROVE WITH NITS Medium x1)

The helper `enforce_admin_only_scopes` and the policy constant `ADMIN_ONLY_SCOPES` are well-designed, and the patched call sites (`profile.rs`, `users.rs`) are correct. Tests are excellent (policy iteration via loop pins future additions, mixed-list smuggling case covered, admin-still-allowed positive case).

However, the fix is **incomplete**. Two additional token-issuance routes still carry the legacy narrow `if !auth.is_admin && payload.scopes.iter().any(|s| s == "admin")` check, which only blocks the literal `"admin"` scope. Both routes remain exploitable for `"*"`, `"delete:artifacts"`, `"delete:repositories"`, and `"write:users"` after this PR merges.

### Critical bypass routes that must also be patched

1. `backend/src/api/handlers/auth.rs:354-381` (`create_api_token`, mounted at `POST /api/v1/auth/tokens`)
2. `backend/src/api/handlers/repo_tokens.rs:267-272` (`create_repo_token`, mounted at `POST /api/v1/repositories/.../tokens`)

Both need the same `token_service::enforce_admin_only_scopes(&payload.scopes, auth.is_admin).map_err(AppError::Authorization)?;` call that profile.rs and users.rs received in this PR.

A follow-up commit is being prepared and will land on this branch shortly.

### High-severity policy gap

`write:repositories` is not in `ADMIN_ONLY_SCOPES`. Per the Security Engineer's analysis: a non-admin self-minting `write:repositories` then has it everywhere, regardless of per-repo delegation, which enables repo settings takeover (visibility, retention, proxy upstreams). The PR rationale conflates "holding" a delegated scope with "minting it for yourself"; the latter is exactly what this PR closes. Recommend adding `write:repositories` to the policy list, OR documenting on the issue why it's intentionally excluded.

### Structural recommendation (Medium, for follow-up not this PR)

Lift the check one layer down into `AuthService::generate_api_token` itself, as a `caller_is_admin: bool` parameter or a typed `TokenIssuer` enum. Today the policy is enforced at the handler layer in 4 call sites; a 5th endpoint added later would be one inline check away from reopening this hole. The tactical handler-level fix is fine for v1.2.0; track the structural lift as a v1.2.1 / v1.3.0 follow-up.

### Other findings (consolidated)

- **Medium**: no `tracing::warn!` on rejection. Adding `warn!(user_id, attempted_scopes, "rejected non-admin token issuance with admin-class scope")` gives SOC visibility for the exact exploitation attempts this PR is designed to stop.
- **Low**: error message includes `format!("Admin-only scopes: {:?}", ADMIN_ONLY_SCOPES)`, leaking the full policy list to the caller. Drop the embedded list; keep the offending scope only.
- **Low**: existing admin-class tokens issued during the 1.1.x exploitable window remain valid. PR body notes this; should ship with a documented SQL revocation snippet for operators.

### Praise

- Centralized policy with table-driven test loop iteration is the right pattern.
- Threat-model comments at call sites (`scopes_grant_access` short-circuit explanation) prevent future maintainers from accidentally deleting the check thinking it's redundant.
- Test design pins the policy — future drops from the list fail tests and force an explicit decision.

Expediting for v1.2.0 once the 2 missed routes are closed.

brandonrc:
Extended the fix to two more privilege-escalation routes: `POST /api/v1/auth/tokens` (`create_api_token`) and `POST /api/v1/repositories/{key}/tokens` (`create_repo_token`) were both still using the legacy narrow check and would have allowed non-admins to mint `*`, `delete:artifacts`, etc. Added regression tests. Expediting for v1.2.0.

Commit: 41f47ea88e59ce6d42b1beaa1a9fdd7db61d97d9

What changed:
- `auth.rs::create_api_token` now calls `enforce_admin_only_scopes`. Three regression tests: iterates every `ADMIN_ONLY_SCOPES` entry expecting 403, smuggle-attempt with `delete:repositories` buried in a safe list, and a positive admin-can-still-mint test.
- `repo_tokens.rs::create_repo_token` same change. Three regression tests with the same shape; the non-admin caller is set up with `write:repositories` so it passes `require_repo_write` and reaches the policy gate.

Local verification: `cargo test --workspace --lib admin_scope_policy` against the local Postgres (`DATABASE_URL` set) -- 11/11 pass (5 existing + 6 new).

On `write:repositories` as admin-only (HIGH finding from security review): I audited the scope's blast radius and chose NOT to add it here, intentionally matching the policy comment in `token_service.rs` ("repository creation is sometimes delegated to non-admin users via permission grants"). It is only checked in `repo_tokens::require_repo_write` and `email_subscriptions` and does not short-circuit `scopes_grant_access`. The `repositories.rs` mutation handlers use the bare `"write"` scope, not `"write:repositories"`. Locking it down should be a follow-up with a configurable policy knob (`STRICT_TOKEN_SCOPE_DELEGATION` or similar) so deployments that delegate repo creation are not silently broken.

brandonrc:
Pushed duplication-gate fix in a7a3901: extracted 2 shared helpers (1 production, 1 test) to reduce repetition in repo_tokens.rs from 6.5% to 0% (jscpd reports 0 clones locally on the file). Behavior unchanged. Expediting for v1.2.0.

- Production: `authorize_repo_for_tokens` collapses the `require_repo_write` + `get_by_key` + `can_access_repo` pre-amble shared by list / create / get / revoke handlers.
- Tests: `post_repo_token_request` collapses the `Request::builder().method(POST).uri(/{key}/tokens)...` block shared by the three admin-scope policy tests.

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --lib repo_tokens` (26 tests) all green.

Issue #1229, opened 2026-05-14
Docker virtual repository does not route manifest requests to member remote repos


jawulfd:
### Description

Pull requests through a Docker virtual repository fail with manifest unknown: manifest not found, even when the same image is successfully pulled through a member remote repository directly. The virtual aggregation layer does not appear to forward OCI manifest lookups to its configured member repos.

### Environment Information

- Deployment type: Helm into rke2 cluster
- Operating System: 
- Architecture: x86-64
- Database: postgres:16-alpine
- Reverse Proxy: Traefik
- Network setup: rke2-canal

### Backend Version

1.1.9

### Web UI Version

1.1.3

### Package Format

Docker / OCI

### Expected Behavior

Pulling through the virtual repository should iterate through its member repos in priority order. Since docker-local does not contain nginx, the request should fall through to ghcr-remote (not found there either) and finally resolve successfully via dockerhub-remote, returning the image transparently to the client — the same way virtual repos work for other formats (PyPI, npm, etc.).

### Observed Behavior

The virtual repository returns manifest unknown: manifest not found regardless of the image requested, even when the same image is immediately available through a member remote repo. The virtual layer does not appear to forward OCI Distribution Spec manifest requests (GET /v2/< name >/manifests/< reference >) to member repositories.

### 1. Create a local repo
 
```bash
curl -X POST https://registry.example.com/api/v1/repositories \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "key": "docker-local",
    "name": "Docker Local",
    "format": "docker",
    "repo_type": "local"
  }'
```
 
### 2. Create a Docker Hub remote repo
 
```bash
curl -X POST https://registry.example.com/api/v1/repositories \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "key": "dockerhub-remote",
    "name": "Docker Hub Remote",
    "format": "docker",
    "repo_type": "remote",
    "upstream_url": "https://registry-1.docker.io"
  }'
```
 
### 3. Create a GHCR remote repo
 
```bash
curl -X POST https://registry.example.com/api/v1/repositories \
  -H "Authorization: Bearer $TOKEN" \
  -H "Content-Type: application/json" \
  -d '{
    "key": "ghcr-remote",
    "name": "GHCR Remote",
    "format": "docker",
    "repo_type": "remote",
[...]

PR #1262, opened 2026-05-19
feat(oci): resolve manifests and blobs through virtual repo members

inspired-geek:
## Summary

Virtual repositories previously supported only `tags/list`. The handlers for `GET/HEAD /v2/<repo>/manifests/<ref>` and `/blobs/<digest>` queried the virtual repo's own `oci_tags`/`oci_blobs` (always empty for a virtual) and then fell through to `try_upstream_fetch`, which short-circuits for any non-`Remote` type. The result: `MANIFEST_UNKNOWN`/`BLOB_UNKNOWN` 404 returned without an OCI auth challenge, so `skopeo inspect docker://host/<virtual>/<image>` failed with `manifest not found` even when the virtual had members that owned the image.

This PR teaches the OCI handlers to iterate the virtual's members, mirroring the pattern already used by `npm`, `pypi`, and `alpine` handlers (`proxy_helpers::fetch_virtual_members`).

- New helpers in `oci_v2.rs`:
  - `member_as_oci_repo_info` — adapts a `Repository` row to the lightweight `OciRepoInfo` view used by the existing helpers.
  - `try_member_manifest_local` / `try_member_blob_local` — per-member hosted lookup against `oci_tags`/`oci_blobs` + storage read.
  - `resolve_virtual_manifest` / `resolve_virtual_blob` — iterate members by `priority`, try hosted first then Remote upstream, first hit wins, exhausted → OCI 404 envelope.
- `handle_get_manifest`, `handle_head_manifest`, `handle_get_blob`, `handle_head_blob` delegate to the new helpers as soon as `repo.repo_type == Virtual`, before the existing local-lookup path (which is unreachable for a virtual and would only burn a SQL round trip).

Behavior changes for a `/v2/<virtual>/...` request:
- Hosted member holds the tag → 200 with the manifest body / blob bytes.
- All members miss → OCI `MANIFEST_UNKNOWN` / `BLOB_UNKNOWN` 404.
- Nested virtual members are skipped (`continue`) — supporting them is a future change.
- Caching for Remote-member proxy hits stays at the member level (same as the standalone Remote path), so the virtual itself doesn't grow a pull-through copy.

(Bordering on a bug fix — the virtual repo type advertised `tags/list` support but silently dropped `manifests`/`blobs`. Filed as `feat/` because the change adds a new resolution path rather than fixing a regression in an existing one.)

## Test Checklist
- [x] Manually tested locally

Coverage on new/changed lines (measured locally with `cargo llvm-cov --workspace --lib` against a real PostgreSQL): **86%** (459/531 instrumented new lines, gate is 70%). Uncovered lines are the digest-based reference branch of `try_member_manifest_local` and the Remote-member proxy branch of `resolve_virtual_manifest`/`resolve_virtual_blob`; both are exercised by integration tests that need an upstream registry and were left out of this PR to keep it focused.

Tests added (all under `backend/src/api/handlers/oci_v2.rs`):
- `tests::test_member_as_oci_repo_info_local|remote|preserves_image_path` — pure adapter coverage.
- `virtual_resolution_db_tests::resolve_virtual_manifest_returns_404_when_no_members`
- `virtual_resolution_db_tests::resolve_virtual_manifest_finds_tag_in_hosted_member`
- `virtual_resolution_db_tests::resolve_virtual_manifest_head_omits_body`
- `virtual_resolution_db_tests::resolve_virtual_manifest_falls_through_to_next_member_on_miss`
- `virtual_resolution_db_tests::resolve_virtual_blob_returns_404_when_no_members`
- `virtual_resolution_db_tests::resolve_virtual_blob_finds_digest_in_hosted_member`
- `virtual_resolution_db_tests::resolve_virtual_blob_head_omits_body`

Pre-push: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --lib` (9589 passed against a local PostgreSQL).

## API Changes
- [x] N/A - no API changes

The OCI Distribution endpoints exposed here are existing; this PR only widens which `repo_type`s they accept.

### DIFF
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -869,6 +869,229 @@ async fn try_upstream_fetch(
         .ok()
 }
 
+/// Adapt a `Repository` row (from `fetch_virtual_members`) to the lightweight
+/// `OciRepoInfo` view used by the OCI handler helpers. The `image` field is
+/// the image path inside the virtual repo — members share that addressing.
+fn member_as_oci_repo_info(
+    member: &crate::models::repository::Repository,
+    image: &str,
+) -> OciRepoInfo {
+    OciRepoInfo {
+        id: member.id,
+        key: member.key.clone(),
+        location: member.storage_location(),
+        repo_type: member.repo_type.as_str().to_string(),
+        upstream_url: member.upstream_url.clone(),
+        is_public: member.is_public,
+        image: image.to_string(),
+    }
+}
+
+/// Try resolving a manifest reference against a single hosted member's
+/// `oci_tags` + storage. Returns `Some((data, content_type, digest))` on hit.
+async fn try_member_manifest_local(
+    state: &SharedState,
+    member: &crate::models::repository::Repository,
+    image: &str,
+    reference: &str,
+) -> Option<(Bytes, String, String)> {
+    let row = if is_digest_reference(reference) {
+        sqlx::query!(
+            "SELECT manifest_digest, manifest_content_type FROM oci_tags \
+             WHERE repository_id = $1 AND manifest_digest = $2 LIMIT 1",
+            member.id,
+            reference
+        )
+        .fetch_optional(&state.db)
+        .await
+        .ok()
+        .flatten()
+        .map(|t| (t.manifest_digest, t.manifest_content_type))
+    } else {
+        sqlx::query!(
+            "SELECT manifest_digest, manifest_content_type FROM oci_tags \
+             WHERE repository_id = $1 AND name = $2 AND tag = $3",
+            member.id,
+            image,
+            reference
+        )
+        .fetch_optional(&state.db)
+        .await
+        .ok()
+        .flatten()
+        .map(|t| (t.manifest_digest, t.manifest_content_type))
+    };
+
+    let (manifest_digest, content_type) = row?;
+    let storage = state.storage_for_repo(&member.storage_location()).ok()?;
+    let data = storage
+        .get(&manifest_storage_key(&manifest_digest))
+        .await
+        .ok()?;
+    Some((data, content_type, manifest_digest))
+}
+
+/// Iterate virtual members in priority order and return the first manifest
+/// hit (hosted DB lookup first per member, then remote upstream fetch).
+async fn resolve_virtual_manifest(
+    state: &SharedState,
+    virtual_repo: &OciRepoInfo,
+    reference: &str,
+    include_body: bool,
+) -> Response {
+    let members = match proxy_helpers::fetch_virtual_members(&state.db, virtual_repo.id).await {
+        Ok(m) => m,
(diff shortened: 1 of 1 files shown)


brandonrc:

brandonrc:
Linking issue #1229 ("Docker virtual repository does not route manifest requests to member remote repos") to this PR. The bug report symptom (status=404 on GET /v2/<virtual>/<image>/manifests/<ref> after oci_tags returns 0 rows, while the same image pulls fine via a member remote) matches the root cause described in this PR summary. Closed #1229 as a duplicate of the work here.

brandonrc:
## Rebase + review: superseded by #1419

Thanks @inspired-geek for this work. While rebasing onto current `main` I found that the feature this PR implements — resolving OCI manifests and blobs through virtual repo members — has already landed via **#1419** (`2731726 feat(oci): virtual repo manifest+blob delegation + digest verify (closes #1348 cluster)`).

Main already defines `resolve_virtual_blob` and `resolve_virtual_manifest` and wires the `RepositoryType::Virtual` delegation into `handle_head_blob`, `handle_get_blob`, `handle_head_manifest`, and `handle_get_manifest` — the same four call sites this PR targets. The merged implementation is also a strict superset:

- **Upstream digest verification** (`finalize_upstream_blob` / `finalize_upstream_manifest`, #1348): when a blob/manifest is requested *by digest*, upstream content is sha256-verified before being served, and a mismatch falls through to the next member. This PR serves upstream content without that check, which is the exact wrong-content-under-digest risk we need to avoid for virtual fan-out.
- **Negative cache** to bound probe storms across many remote members.
- **`candidate_upstream_images`** Docker Hub `library/` normalization.
- **Accept-header-aware** manifest proxying.

Rebasing produced conflicts only in the four handler guards (trivially resolvable) and large unavoidable collisions in the test module, where this PR's resolver functions duplicate main's. After the handler guards, the remaining PR-unique code (`member_as_oci_repo_info`, `try_member_blob_local`, `try_member_manifest_local`) exists solely to back this PR's own versions of the two resolvers and would cause duplicate-definition compile errors against main. There is no net-new behavior to forward-port.

**Verdict: NEEDS-WORK → recommend close as superseded by #1419.** Labeled `needs-maintainer-review` for the maintainer to make the call. No rebased branch was pushed since there is no salvageable delta. The credit for tackling virtual OCI resolution is appreciated — it converged with #1419's solution.

brandonrc:
Closing as **superseded**: `main` now resolves OCI manifests and blobs through virtual-repo members via the consolidated #1348 work — `resolve_virtual_blob`, `resolve_virtual_manifest`, the `VirtualBlobResolution` enum, and the `backend/tests/oci_virtual_resolution_tests.rs` suite are all present on `main`, including the digest-verification security hardening. Thanks @inspired-geek — this feature shipped (your approach informed the consolidated version). Reopen if a specific behaviour here is still missing on main.

PR #1267, opened 2026-05-19
fix(pypi): union virtual member entries on simple/<project>/ (#1230)

### LINKED ISSUE #1230: PyPI virtual: simple/<project>/ returns first non-empty member instead of unioning entries from all members
### Description

For a virtual PyPI repo whose members include both a **local** repo and a **remote** repo (proxy to public PyPI), `GET /pypi/<virtual>/simple/<project>/` returns the response from whichever member is hit first in priority order — instead of merging entries from every member.

This means that if a package name exists in both the local member *and* the remote upstream (PyPI namespace collision — e.g. an internal package named `yoda`, `requests`, `boto3`-prefixed forks, etc.), users only see versions from the higher-priority member. The lower-priority member's versions are silently dropped from the simple-index page even though they're served correctly when queried via the member repo directly.

Result for end users: `pip install internal-name==X.Y.Z` against the virtual reports `Could not find a version that satisfies the requirement` even though the version is present in a local member.

### Root cause

`simple_project` in `backend/src/api/handlers/pypi.rs` iterates virtual members in priority order and **returns inside the loop** on the first member that yields any matching artifact rows / non-empty proxy response. Concretely, in the local-member branch (around the per-member SQL query):

```rust
if !member_rows.is_empty() {
    let member_artifacts: Vec<SimpleProjectArtifact> = member_rows
        .into_iter()
        .map(|a| SimpleProjectArtifact { ... })
        .collect();
    return build_simple_project_response(
        &headers,
        &repo_key,
        &normalized,
        &member_artifacts,
    );          // <-- returns here, no subsequent members are queried
}
continue;
```

The same `return` happens in the remote-member branch on a successful upstream proxy response. The net effect is "first non-empty member wins" — which is fine for content-addressed formats like Maven where each GAV lives in exactly one underlying repo, but is wrong for PyPI where the simple-index is the **union** of all distributions a virtual aggregator can serve. PEP 503 doesn't say anything that would justify hiding versions from one member because another member also has entries under the same project name.

`simple_root` (line ~153) already does aggregation correctly for the root index: it loops over members and `merged.extend(...)`s into a single result set, then builds the response once at the end. `simple_project` needs the same shape.

### Steps to reproduce

```bash
# 1. Create a local PyPI repo
curl -X POST $BASE/api/v1/repositories \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"key":"pypi-local","name":"pypi-local","format":"pypi","repo_type":"local"}'

# 2. Create a remote PyPI repo (upstream public PyPI)
curl -X POST $BASE/api/v1/repositories \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"key":"pypi-remote","name":"pypi-remote","format":"pypi","repo_type":"remote","upstream_url":"https://pypi.org"}'

# 3. Create a virtual PyPI repo aggregating both, with pypi-remote at priority 1
curl -X POST $BASE/api/v1/repositories \
  -H "Authorization: Bearer $TOKEN" -H "Content-Type: application/json" \
  -d '{"key":"pypi","name":"pypi","format":"pypi","repo_type":"virtual",
       "member_repos":[{"repo_key":"pypi-remote","priority":1},
                       {"repo_key":"pypi-local","priority":2}]}'

# 4. Upload a package to pypi-local whose name also exists on public PyPI
#    (e.g. "yoda" — public PyPI has yoda 0.1.x, you upload your own yoda 6.0.1)
twine upload --repository-url $BASE/pypi/pypi-local/ \
             --username __token__ --password $TOKEN \
             dist/yoda-6.0.1.tar.gz

# 5. Verify the upload landed
curl -sSL -H "Authorization: Bearer $TOKEN" \
  "$BASE/pypi/pypi-local/simple/yoda/" | grep yoda-6.0.1
# -> 200, body contains <a href="...yoda-6.0.1.tar.gz">...</a>

# 6. BUG: query the virtual's simple-index for the same package
curl -sSL -H "Authorization: Bearer $TOKEN" \
  "$BASE/pypi/pypi/simple/yoda/" | grep yoda-6.0.1
# -> 200, but yoda-6.0.1 is NOT in the response.
#    Only upstream PyPI's 0.1.x entries are returned.

# 7. End-to-end manifestation
pip install --index-url "https://__token__:$TOKEN@$HOST/pypi/pypi/simple/" yoda==6.0.1
# ERROR: Could not find a version that satisfies the requirement yoda==6.0.1
#        (from versions: 0.1, 0.1.1, 0.1.2, 0.1.2.1, 0.1.3, 0.1.4, 0.1.5)
```

Swapping member priorities (pypi-local at priority 1, pypi-remote at priority 2) hides the bug for this case (pypi-local now wins on `yoda`), but only because pypi-local correctly returns 404 for packages it doesn't have, letting the loop fall through to pypi-remote. The bug is still present and silently shadows any package present in both members — just in the other direction.

### Expected behavior

`GET /pypi/<virtual>/simple/<project>/` returns a single PEP 503 HTML response whose `<a>` entries are the **union** of distributions found across all members. Order isn't semantically significant in PEP 503 but priority order is a reasonable default for ergonomics; duplicates (same filename in multiple members) should be deduped on the rendered file name.

### Observed behavior

The response contains entries from the first member that yielded any matching rows / non-empty proxy response, and entries from every subsequent member are dropped.

### Suggested fix

In `simple_project`, mirror the structure of `simple_root`:

1. Initialize `let mut merged: Vec<SimpleProjectArtifact> = Vec::new();` outside the loop.
2. For each local/staging member: append matching rows into `merged` instead of returning. Don't `continue` past non-empty members.
3. For each remote member: append parsed-and-rewritten entries from the proxied response into `merged` (or fall back to "return first non-empty remote" if extracting structured entries from the proxied HTML is undesirable — but parsing the HTML on the way out is what the existing rewrite step already does).
4. After the loop, dedupe by filename, then `build_simple_project_response(..., &merged)` once.

A unit test that asserts merged length > each individual member's length (i.e. neither member's count alone equals the merged count) would catch the regression.

### Workaround

Swap virtual member priorities so the local repo is at priority 1. This works as long as the local repo cleanly returns 404 for packages it doesn't have (it does in current builds), letting the handler fall through to the remote member for upstream-only packages. The fragility is asymmetric: any package present in both members loses one side's versions silently.

### Environment Information

- Deployment type: Helm (Kubernetes), AWS EKS
- Backend version: 1.1.9
- Database: PostgreSQL 14 (RDS)
- Storage backend: S3
- Reverse proxy: AWS ALB (AWS Load Balancer Controller)

### Package Format

PyPI

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate (#643 is related but distinct — that was about local members being skipped entirely when remotes existed; this is about local members being included but their entries being dropped from the merged response when an earlier member also matched)

ThaSami:
## Summary

Fixes #1230. `simple_project` previously returned the response of the first virtual member that yielded any matching artifacts (local DB rows OR a non-empty remote proxy response). When a local member holds only a partial subset of a package's releases — e.g. an incremental cache-migration from a previous registry — that subset shadows the remote member's full upstream listing under the same name. `pip` then sees only the local subset and dependency resolution fails for any transitive constraint that needs a non-cached version, even though the version is reachable via the same virtual.

The bug was originally reported as a name-collision case (local internal package shadows upstream of the same name). #1230 [received a follow-up comment](https://github.com/artifact-keeper/artifact-keeper/issues/1230#issuecomment-4492469106) showing the symmetric case: a locally-cached **subset** of an upstream package shadowing the rest of upstream's release history during transitive dependency resolution. Same root cause, same fix.

## Reproducer (used to validate the fix end-to-end)

A virtual `pypi-virtual` whose members are:

| member | type | priority | content |
|---|---|---|---|
| `pypi-local` | local | 1 | 6 cached versions of `litellm` (1.83.x — 1.85.0) |
| `pypi-remote` | remote | 2 | proxy to `https://pypi.org` |

```bash
pip install --index-url https://<host>/pypi/pypi-virtual/simple/ judgellm-sdk==0.1.1
```

`judgellm-sdk`'s dependency `dspy>=3.1.2` requires `litellm <=1.82.6, >=1.64.0`, which exists on `pypi-remote` but is missing from `pypi-local`'s cached subset.

**Before this change**:

```
GET /pypi/pypi-virtual/simple/litellm/      → 1268  bytes (6 cached versions)
GET /pypi/pypi-virtual/simple/typeguard/    →  363  bytes (1 cached version)

ERROR: Cannot install judgellm-sdk because these package versions have
       conflicting dependencies.
       Additionally, some packages in these conflicts have no matching
       distributions available for your environment: litellm, typeguard
ERROR: ResolutionImpossible
```

**After this change** (verified in a local cluster running this branch):

```
GET /pypi/pypi-virtual/simple/litellm/      → 660 339 bytes (cached + upstream union)
GET /pypi/pypi-virtual/simple/typeguard/    →  33 741 bytes

Successfully installed judgellm-sdk-0.1.1 dspy-3.2.0 litellm-1.82.6 ...
```

## Approach

`simple_project`'s virtual branch now iterates every member instead of returning on the first match:

* Local / staging members → DB rows accumulated into `local_artifacts: Vec<SimpleProjectArtifact>`.
* Remote members → proxy fetch; the **first** successful response is kept as the canonical HTML carrier (`remote_response: Option<(Bytes, Option<String>)>`).

After the loop:

* `(local empty, remote None)` → 404 (unchanged behavior).
* `(local non-empty, remote None)` → `build_simple_project_response(local)` (existing path).
* `(_, remote Some)` → new `merge_local_into_remote_simple_html` splices local `<a>...</a><br/>` entries into the remote HTML right before `</body>`, deduplicating by the anchor's inner text (filename, per PEP 503). The local entries carry their own `#sha256=...` fragment and `data-requires-python` attribute (HTML-escaped).

## Why the splice (not a full reparse) approach

The remote proxy response is already a complete, PEP 503-compliant HTML document with `rewrite_upstream_urls` applied — round-tripping through a reparse-and-rerender would risk losing attributes that upstream may add in the future (PEP 658 `data-dist-info-metadata`, PEP 740 `data-provenance`, etc.). Splicing keeps the remote response byte-for-byte and only adds local entries at a single well-defined insertion point.

The dedupe regex matches the anchor's inner text rather than the href because the href format differs between local (rewritten to `/pypi/{repo}/simple/{pkg}/{file}#sha256=…`) and remote (rewritten from upstream, may carry additional fragments). Filename is the stable identifier and matches PEP 503's contract.

## Scope and trade-offs

* **Multiple remote members in one virtual**: out of scope. The first remote response wins; subsequent remote members are skipped. The topology with two upstream proxies under one virtual is uncommon, and merging two upstream `simple/<pkg>/` listings deterministically (priority? hash dedupe across different mirrors?) is its own design question. Filed as a follow-up note in the code comment.
* **JSON (PEP 691) accept path**: when the union is local-only, JSON works as before via `build_simple_project_response`. When a remote member contributes, the response keeps upstream's content-type and any JSON-requested client falls back to whatever format upstream returned — same behavior as today. Unioning JSON `files[]` arrays cleanly is a separate change worth doing once a real client requests it (current callers we've observed use HTML).
* **Performance**: the dedupe set is built once per request with a `Lazy<Regex>` static, allocates only the captured `&str` references into a `HashSet<&str>`, and runs on the rewritten remote HTML (typically tens to hundreds of KB). For a package present in both members the splice is an O(n) string concat. No additional DB queries vs. the previous loop — local member rows were already being fetched, just discarded after the first non-empty hit.

## Tests

Added 5 focused tests on the splice helper in `api::handlers::pypi::tests`:

* `test_merge_local_appends_entries_absent_from_remote` — both sides' entries reach the rendered response; local URLs use the virtual repo key + sha256 fragment; local entries are spliced before `</body>`.
* `test_merge_local_skips_filenames_already_in_remote` — same filename in both members yields exactly one `<a>` in the output; the remote entry wins (its sha256 is canonical).
* `test_merge_empty_local_returns_remote_unchanged` — no-op fast path.
* `test_merge_emits_data_requires_python_attribute` — `requires_python` from local artifact metadata is HTML-escaped on the rendered anchor.
* `test_merge_handles_remote_html_without_body_close` — defensive: malformed upstream HTML (no `</body>`) still surfaces local entries via append.

Existing 106 pypi-handler tests continue to pass. `cargo clippy -D warnings` is clean.

## Verification

```
$ SQLX_OFFLINE=true cargo test -p artifact-keeper-backend --lib api::handlers::pypi
test result: ok. 106 passed; 0 failed; 0 ignored; 0 measured

```

Closes #1230.

### DIFF
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -451,8 +451,7 @@ async fn simple_project(
                     );
                 }
                 (_, Some((content, content_type))) => {
-                    let ct =
-                        content_type.unwrap_or_else(|| "text/html; charset=utf-8".to_string());
+                    let ct = content_type.unwrap_or_else(|| "text/html; charset=utf-8".to_string());
                     let body = if ct.contains("text/html") {
                         let html = String::from_utf8_lossy(&content);
                         let rewritten = rewrite_upstream_urls(&html, &repo_key, &project);
@@ -2973,8 +2972,14 @@ mod tests {
 
         let merged = merge_local_into_remote_simple_html(&remote, "virt", "pkg", &local);
 
-        assert!(merged.contains("pkg-1.0.0.tar.gz"), "remote entry preserved");
-        assert!(merged.contains("pkg-2.0.0-py3-none-any.whl"), "local entry spliced in");
+        assert!(
+            merged.contains("pkg-1.0.0.tar.gz"),
+            "remote entry preserved"
+        );
+        assert!(
+            merged.contains("pkg-2.0.0-py3-none-any.whl"),
+            "local entry spliced in"
+        );
         assert!(
             merged.contains("/pypi/virt/simple/pkg/pkg-2.0.0-py3-none-any.whl#sha256=ffeeddccbbaa99887766554433221100"),
             "local URL uses the virtual repo key and carries the sha256 fragment"
@@ -3004,7 +3009,9 @@ mod tests {
         assert_eq!(count, 1, "filename present exactly once after dedupe");
         // The local sha256 must NOT appear — the remote entry is canonical.
         assert!(
-            !merged.contains("sha256=0000000000000000000000000000000000000000000000000000000000000000"),
+            !merged.contains(
+                "sha256=0000000000000000000000000000000000000000000000000000000000000000"
+            ),
             "local sha256 not spliced in when filename dedupes against remote"
         );
     }
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -342,12 +342,11 @@ async fn simple_project(
                     .unwrap());
             }
         }
-        // For virtual repos, iterate through members in priority order.
-        // Local/staging members are queried via DB; remote members use proxy.
+        // For virtual repos, iterate through ALL members and union their
+        // entries — both local DB rows and remote proxy responses — so a
+        // package that exists partially in a local member doesn't shadow
+        // the rest of upstream. See #1230.
         if repo.repo_type == RepositoryType::Virtual {
-            // Per-member upstream_url adjustment happens inside the loop below
-            // so each remote member can carry its own `…/simple` suffix without
-            // polluting siblings.
             let members = proxy_helpers::fetch_virtual_members(&state.db, repo.id).await?;
 
             if members.is_empty() {
@@ -357,9 +356,10 @@ async fn simple_project(
                 );
             }
 
+            let mut local_artifacts: Vec<SimpleProjectArtifact> = Vec::new();
+            let mut remote_response: Option<(Bytes, Option<String>)> = None;
+
             for member in &members {
-                // For local and staging repos, query the DB for matching
-                // artifacts, the same way we do for the top-level repo.
                 if member.repo_type == RepositoryType::Local
                     || member.repo_type == RepositoryType::Staging
                 {
@@ -381,29 +381,26 @@ async fn simple_project(
                     .await
                     .map_err(map_db_err)?;
 
-                    if !member_rows.is_empty() {
-                        let member_artifacts: Vec<SimpleProjectArtifact> = member_rows
-                            .into_iter()
-                            .map(|a| SimpleProjectArtifact {
-                                path: a.path,
-                                version: a.version,
-                                size_bytes: a.size_bytes,
-                                checksum_sha256: a.checksum_sha256,
-                                metadata: a.metadata,
-                            })
-                            .collect();
-                        return build_simple_project_response(
-                            &headers,
-                            &repo_key,
-                            &normalized,
-                            &member_artifacts,
-                        );
-                    }
+                    local_artifacts.extend(member_rows.into_iter().map(|a| {
+                        SimpleProjectArtifact {
+                            path: a.path,
+                            version: a.version,
+                            size_bytes: a.size_bytes,
+                            checksum_sha256: a.checksum_sha256,
+                            metadata: a.metadata,
+                        }
+                    }));
                     continue;
                 }
 
-                // For remote repos, proxy the simple index from upstream.
                 if member.repo_type != RepositoryType::Remote {
+                    continue;
+                }
+                // Only take the first remote response; multiple remote
+                // members in one virtual is rare, and merging two upstream
+                // /simple/<pkg>/ listings deterministically is out of scope
+                // for this fix.
+                if remote_response.is_some() {
                     continue;
                 }
                 let Some(ref upstream_url) = member.upstream_url else {
@@ -426,21 +423,7 @@ async fn simple_project(
 
                 match result {
                     Ok((content, content_type)) => {
-                        let ct =
-                            content_type.unwrap_or_else(|| "text/html; charset=utf-8".to_string());
-                        let body = if ct.contains("text/html") {
-                            let html = String::from_utf8_lossy(&content);
-                            let rewritten = rewrite_upstream_urls(&html, &repo_key, &project);
-                            Body::from(rewritten)
-                        } else {
-                            Body::from(content)
-                        };
-
-                        return Ok(Response::builder()
-                            .status(StatusCode::OK)
-                            .header(CONTENT_TYPE, ct)
-                            .body(body)
-                            .unwrap());
+                        remote_response = Some((content, content_type));
                     }
                     Err(_e) => {
                         debug!(
@@ -451,10 +434,46 @@ async fn simple_project(
                 }
             }
 
-            return Err(AppError::NotFound(
-                "Package not found in any member repository".to_string(),
-            )
-            .into_response());
+            // Render the union.
+            match (local_artifacts.is_empty(), remote_response) {
+                (true, None) => {
+                    return Err(AppError::NotFound(
+                        "Package not found in any member repository".to_string(),
+                    )
+                    .into_response());
+                }
+                (false, None) => {
+                    return build_simple_project_response(
+                        &headers,
+                        &repo_key,
+                        &normalized,
+                        &local_artifacts,
+                    );
+                }
+                (_, Some((content, content_type))) => {
+                    let ct =
+                        content_type.unwrap_or_else(|| "text/html; charset=utf-8".to_string());
+                    let body = if ct.contains("text/html") {
+                        let html = String::from_utf8_lossy(&content);
+                        let rewritten = rewrite_upstream_urls(&html, &repo_key, &project);
+                        let merged = merge_local_into_remote_simple_html(
+                            &rewritten,
+                            &repo_key,
+                            &normalized,
+                            &local_artifacts,
+                        );
+                        Body::from(merged)
+                    } else {
+                        Body::from(content)
+                    };
+
+                    return Ok(Response::builder()
+                        .status(StatusCode::OK)
+                        .header(CONTENT_TYPE, ct)
+                        .body(body)
+                        .unwrap());
+                }
+            }
         }
 
         return Err(AppError::NotFound("Package not found".to_string()).into_response());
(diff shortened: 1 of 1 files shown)


ThaSami:
@brandonrc your thoughts appreciated here, this one is critical for migration

brandonrc:
## 3-agent review verdict

**Consensus: APPROVE WITH NITS** (Code Reviewer: APPROVE WITH NITS; Backend Architect: APPROVE WITH NITS; audit-code-quality: APPROVE WITH NITS)

The union fix is correct, well-scoped, and well-tested. The splice approach (preserve upstream HTML, inject local entries before `</body>`) is the right call — it forward-protects against PEP 658/740 attributes that upstream may add to anchor tags. The 5 targeted unit tests cover the four cases that matter (empty-local, overlap, no-`</body>`, requires-python escape). The decision to dedupe by anchor inner text (filename per PEP 503) rather than href or sha256 matches the stable identifier, and the test pinning that the remote sha256 is preserved over the local one is exactly right semantics. Architectural observation: the union lives in the pypi handler, consistent with the existing pattern across all 14 format handlers; that's defensible for this PR.

### Convergent finding (Medium)

All three reviewers caught this independently: the union is **HTML-only**. When a remote member's `Content-Type` is `application/vnd.pypi.simple.v1+json` (or upstream returned `+json` because the client asked for it), the code falls through to `Body::from(content)` and ships the raw upstream JSON with zero local entries merged. So a PEP 691 JSON client (modern `uv`, recent `pip` versions) against a virtual still hits the same shadowing bug #1230 is supposed to fix. The PR body acknowledges this is "separate change" but the fix is incomplete for clients that opt into PEP 691.

Recommend: at minimum, force `Accept: text/html` on the upstream fetch when local artifacts are present and the client requested JSON. Long-term: add a parallel `merge_local_into_remote_simple_json` function. Either way, file a follow-up issue with a `// TODO(#NNNN)` comment in the JSON arm so this isn't forgotten.

### Other findings (Low)

- **DoS vector via unbounded dedup set** (audit-code-quality): the `existing` HashSet is built from every `<a>` capture in the full remote HTML with no upper bound. A malicious or pathological upstream returning millions of anchor tags causes proportional allocation per virtual index request. Add `.take(10_000)` guard or check `existing.len()` before inserting.
- **Virtual-union pattern repeats across 14 handlers** (Backend Architect): same first-non-empty-member shadowing bug almost certainly exists in npm, conda, helm, oci_v2, goproxy, conan, terraform, cocoapods, vscode, jetbrains, composer, ansible, chef, sbt. File follow-up to extract a `VirtualUnionStrategy` trait so this fix doesn't have to be re-derived 14 times. Not a blocker for #1230.
- **Regex anchor matcher** (Code Reviewer): `<a\s[^>]*>([^<]+)</a>` would break on `data-*` values containing `>`. PEP 503 makes this vanishingly rare but a more tolerant pattern (`<a\b[^>]*?>` + non-greedy) is cheap.
- **Silent member-error degradation** (Backend Architect + audit-code-quality): remote member 500 logs at `debug!` and is swallowed, which now silently degrades a virtual to local-only. Suggest `warn!` plus a response header (`X-Virtual-Member-Errors: 1`) so operators see partial-union responses in logs.
- **Sequential member iteration** (Code Reviewer): the loop awaits each member's DB query and each remote fetch sequentially; with N members this is N round-trips. Not a regression (old code did the same).
- **Comment sync risk** (Code Reviewer): `build_simple_project_response` and the spliced helper emit anchor tags with identical attributes. If a future change adds `data-core-metadata` to one, the other silently drifts. Add `// keep in sync with build_simple_project_response` cross-reference comment.

### Praise

- Splice over reparse-rerender is exactly right and well-justified in the PR body. PEP 658/740 forward-compat is a real concern most authors would miss.
- Member ordering uses the existing `vrm.priority` from `fetch_virtual_members`, not hardcoded — the flexible-but-already-built option.
- Tests directly encode the architectural invariant ("each member contributes its entries") rather than smoke-testing the happy path.
- PR body honestly acknowledges the JSON-path scope-cut rather than hiding it.

### Plan

Mergeable as-is for HTML clients. Follow-ups to file:
1. PEP 691 JSON union (medium priority)
2. DoS guard on dedup set size (low)
3. Shared virtual-union strategy across 14 handlers (medium, architecture)
4. Member-error visibility (low)

Expediting for v1.2.0 once the JSON-gap follow-up is filed.

PR #1268, opened 2026-05-20
fix(migration): correct column names + required NOT NULL columns in MigrationService

### LINKED ISSUE #1305: Migration jobs fail with `column "repository_type" does not exist` when the destination repo already exists in AK
## Summary

`MigrationService::check_repository_conflict` and `MigrationService::create_repository` reference `repositories` columns that don't exist on the actual schema. The bug was latent until the worker pre-pass in #1017 started calling these helpers; from 1.2.0-rc.1 onward, any migration job whose target repo already exists in AK fails at the first SELECT.

## Symptom

Migration job flips to `failed` in ~200 ms with `total_items=0`. Error:

```
Database error: error returned from database:
column "repository_type" does not exist
```

## Root cause

1. `check_repository_conflict` SELECTs `repository_type` and `format` from `repositories`. The actual column is `repo_type` (per `003_repositories.sql:16`); `repository_type` is the ENUM type name, not a column. `repo_type` and `format` are both ENUMs, so they also need explicit `::text` casts to decode into `(String, String)`.
2. `create_repository` INSERTs into `(key, display_name, repository_type, format, description, metadata)`. Three of those columns don't exist on the table:
   - `display_name` should be `name`
   - `repository_type` should be `repo_type`
   - `metadata` does not exist on `repositories`
3. `storage_path` is `NOT NULL` with no default but the INSERT supplied no value, so even with column names corrected the INSERT would fail with a NOT NULL constraint violation.

These queries were dead code before #1017 wired the worker pre-pass into `migration_worker::process_job`, so no existing test caught the schema drift.

## Repro

1. Pre-create a repository (e.g. `pypi-local`) via `POST /api/v1/repositories`.
2. Add an Artifactory source connection that has a repository with the same key.
3. POST a migration with `include_repos: ["pypi-local"]`, then POST `/start`.
4. Observe the job flip to `failed` immediately with the column-not-found error.

## Fix

Being addressed by PR #1268, which corrects the column names, casts the ENUMs to text, and supplies values for the NOT NULL `name` and `storage_path` columns.

ThaSami:
Closes #1305

## Summary

`MigrationService::check_repository_conflict` and `MigrationService::create_repository` reference columns on the `repositories` table that don't exist on the actual schema. The queries have been latent since the original artifactory-migration PR (#17) because nothing in the worker called these helpers — they sat as dead code. After destination-repo auto-provisioning was wired into `migration_worker::process_job` (#1017), every migration job where the target repository already exists in AK now fails at the very first SELECT with:

```
Database error: error returned from database:
column "repository_type" does not exist
```

The bug only fires on Artifact Keeper releases that include the new worker pre-pass (1.2.0-rc.1 and later); older releases never reached this code path, which is why CI integration didn't catch it.

### 1. `check_repository_conflict` SELECT (was `migration_service.rs:312-321`)

```sql
SELECT repository_type, format
FROM repositories
WHERE key = $1
```

- `repository_type` is the **enum type name**, not the column name. The column is `repo_type` (per `003_repositories.sql:16`).
- `repo_type` and `format` are both ENUMs, so decoding into `(String, String)` needs explicit `::text` casts; otherwise sqlx returns a runtime decode error.

After the fix:

```sql
SELECT repo_type::text, format::text
FROM repositories
WHERE key = $1
```

### 2. `create_repository` INSERT (was `migration_service.rs:390-408`)

```sql
INSERT INTO repositories (key, display_name, repository_type, format, description, metadata)
VALUES ($1, $2, $3, $4, $5, $6)
```

Three of the six columns don't exist on the actual `repositories` schema:

| Used | Actual |
|---|---|
| `display_name` | `name` |
| `repository_type` | `repo_type` |
| `metadata` | *(no such column on this table)* |

Two more problems:
- `storage_path` is `NOT NULL` with no DEFAULT (`003_repositories.sql:21`) but the INSERT never supplied a value. Even with the column names corrected, the INSERT would fail with a NOT NULL constraint violation.
- The `serde_json::json!({...})` payload was bound to a non-existent column and silently dropped on the wire.

After the fix:

```sql
INSERT INTO repositories (key, name, description, format, repo_type, storage_path)
VALUES ($1, $2, $3, $4::repository_format, $5::repository_type, $6)
RETURNING id
```

- `name` and `storage_path` both default to the migration's target key for auto-provisioned repos, matching the convention of "key as the canonical identifier for locally-stored repos".
- The migration provenance that the dropped `metadata` payload tried to encode is already recorded in the `migrations` table's `config.include_repos` and source connection — no information loss.

## Reproducer (against any AK 1.2.0-rc.1 deployment)

1. Pre-create a repository, e.g. `pypi-local`, via the normal `POST /api/v1/repositories` API.
2. Add an Artifactory source connection that has a repository with the same key.
3. POST a migration with `include_repos: ["pypi-local"]`, then POST `/start`.

Expected: the worker recognizes the existing AK repository and proceeds to artifact transfer (auto-create not needed).
Observed without this fix: the migration job flips to `failed` in ~200 ms with the column-not-found error above; `total_items=0`.

## Why no tests catch this today

`migration_service.rs`'s existing unit tests are all pure (no DB). The queries fixed here were dead until the worker started invoking them — no integration test exercises `check_repository_conflict` / `create_repository` end-to-end against a real Postgres schema. A follow-up adding DB-backed regression tests for these helpers is worth doing; this PR keeps scope to the schema corrections so the fix can ship quickly.

## Local verification

```

$ cargo fmt --all -- --check
    (clean)

$ SQLX_OFFLINE=true cargo test -p artifact-keeper-backend --lib migration_service
    test result: ok. 81 passed; 0 failed
```

### DIFF
--- a/backend/src/services/migration_service.rs
+++ b/backend/src/services/migration_service.rs
@@ -311,7 +311,7 @@ impl MigrationService {
         // Check if repository with same key exists
         let existing: Option<(String, String)> = sqlx::query_as(
             r#"
-            SELECT repository_type, format
+            SELECT repo_type::text, format::text
             FROM repositories
             WHERE key = $1
             "#,
@@ -387,23 +387,24 @@ impl MigrationService {
 
         let repo_type = config.repo_type.to_artifact_keeper();
 
+        // The repositories table schema has no `metadata`, `display_name`, or
+        // `repository_type` columns. The corresponding columns are `name` and
+        // `repo_type`, and `storage_path` is NOT NULL — so the INSERT must
+        // supply it. We default storage_path to the repository key, matching
+        // how local repos are typically named by the storage backend.
         let repo_id: (Uuid,) = sqlx::query_as(
             r#"
-            INSERT INTO repositories (key, display_name, repository_type, format, description, metadata)
-            VALUES ($1, $2, $3, $4, $5, $6)
+            INSERT INTO repositories (key, name, description, format, repo_type, storage_path)
+            VALUES ($1, $2, $3, $4::repository_format, $5::repository_type, $6)
             RETURNING id
             "#,
         )
         .bind(&config.target_key)
-        .bind(&config.target_key) // display_name same as key
-        .bind(repo_type)
-        .bind(&format)
+        .bind(&config.target_key) // name same as key for auto-provisioned repos
         .bind(&config.description)
-        .bind(serde_json::json!({
-            "migrated_from": "artifactory",
-            "source_key": config.source_key,
-            "original_package_type": config.package_type,
-        }))
+        .bind(&format)
+        .bind(repo_type)
+        .bind(&config.target_key) // storage_path defaults to key
         .fetch_one(&self.db)
         .await?;
 


ThaSami:
@brandonrc another critical one that should be shipped with 1.2.0, any migration now crashes directly on 1.2.0-rc.1

brandonrc:
Tracking issue created: #1305.

brandonrc:
## 3-agent review verdict

**Consensus: APPROVE WITH NITS** (DB Optimizer: APPROVE WITH NITS; Code Reviewer: APPROVE WITH NITS; Backend Architect: APPROVE WITH NITS)

The SQL corrections themselves are accurate. All three reviewers verified the fix against `backend/migrations/003_repositories.sql:11-29`:

- `repository_type` → `repo_type`: correct. `repository_type` is the PG enum *type* name; `repo_type` is the actual column name.
- `display_name` → `name`: correct, matches column 14.
- `storage_path` `NOT NULL` requirement: correct, matches schema (no default).
- `metadata` column truly does not exist on `repositories`.
- Explicit `::text` casts on SELECT and `::repository_format` / `::repository_type` casts on INSERT are both necessary for sqlx enum encoding/decoding.

No `.sqlx/` cache regeneration needed since the file uses dynamic `sqlx::query_as`, not the `query_as!` macro.

### Surprising finding (Medium, recommend addressing before merge)

Backend Architect dug into git history and found that the PR body's "dead-code activated by #1017" narrative is partially wrong:

- #1017 (commit `e37e8c1b`, 2026-05-05) DID wire `check_repository_conflict` and `create_repository` into `migration_worker.rs`.
- A subsequent commit (likely `a482b42b` "tighten conflict-match arms" or the `resolve_repos_for_provisioning` refactor) REMOVED those call sites.
- `grep -rn "check_repository_conflict\|\.create_repository(" backend/src/` over current main returns **zero callers** outside the definitions in `migration_service.rs`.

So this PR fixes SQL on methods that have **returned to dead-code status**. The fix is still correct, but the "why now" narrative is inaccurate, and the reproducer in the issue (#1305) presumably hits a different code path today — or doesn't reproduce at all on current main.

Two ways forward:

1. **Re-wire** `migration_worker` to call these methods (intended behavior; the issue's symptom requires they be called).
2. **Delete** them (if the post-refactor logic sidesteps the schema-creation path entirely, these helpers are vestigial).

Filing this as a follow-up is acceptable; punching it into this PR is also acceptable. Either way, leaving the situation undocumented re-creates the AK anti-pattern from #1017 in reverse.

### Other nits (Low to Medium)

- **No regression test** (all three reviewers). Convert the dynamic `sqlx::query_as("...")` to the `query_as!` macro form — compile-time schema validation would have caught all three column errors at `cargo check` and prevents recurrence structurally. A DB integration test is the second-best alternative.
- **Schema duplication** (Backend Architect): four INSERT-into-repositories sites exist (`repository_service.rs:270`, `migration_service.rs:392`, `conan.rs:2281`, `conan.rs:3145`). The two `conan.rs` sites bypass `RepositoryService` entirely, which is how column drift sneaks back in. Route all repo creation through `RepositoryService::create_repository`.
- **Naming overlap** (Backend Architect): `MigrationService` = data-import-from-Artifactory; `backend/migrations/` + sqlx migrator = Postgres schema migrations. Two different concepts sharing a name. Rename to `ImportService` or `ArtifactoryImportService`.
- **Information loss on migration provenance** (DB Optimizer): dropping the `metadata` JSON (`migrated_from`, `source_key`, `original_package_type`) is defensible since `migrations` table holds that data, but consider enriching `description` with provenance text (`format!("Migrated from {}", config.source_key)`) for operator visibility on the repo detail page.
- **In-flight jobs** (DB Optimizer): pre-fix jobs in `failed` state won't auto-recover; release note callout recommended.
- **`storage_path = key` default** (DB Optimizer): matches local-repo convention but should be documented in `RepositoryMigrationConfig`.

### Praise

- Schema-alignment work is meticulous; every column verified against the canonical migration.
- PR author acknowledged the dead-code activation pattern (even though the verification was incomplete, the awareness is good).

### Plan

Mergeable as-is for the SQL correctness, but the dead-code question should be tracked. Recommend filing follow-ups for: (1) re-wire or delete `check_repository_conflict` / `create_repository`, (2) convert dynamic queries to `query_as!`, (3) route Conan handlers through `RepositoryService::create_repository`.

Expediting for v1.2.0.

PR #1269, opened 2026-05-20
fix(migrations): raise statement_timeout/lock_timeout for the migration session

### LINKED ISSUE #1304: Pod fails to boot with `Migration(VersionMismatch)` / `statement timeout` when Postgres `statement_timeout` is shorter than a long schema migration
## Summary

`sqlx::migrate!().run(&db_pool)` inherits whatever per-query `statement_timeout` the migration session gets from the connecting client, usually a small safeguard set on the parameter group (10 s on many AWS RDS tunings, 30 s common elsewhere). Long-running schema migrations (large `CREATE INDEX`, full-table `ALTER`, backfill `UPDATE`, or anything queued behind `ACCESS EXCLUSIVE` while old pods hold shared locks) get killed mid-flight.

## Symptom

The pod fails to boot. Operators see `Migration(VersionMismatch)` or raw `statement timeout` errors on every restart. Recovery requires an out-of-band psql session with `SET statement_timeout = 0`, running the SQL by hand, and inserting a `_sqlx_migrations` row with the canonical SHA-384 checksum.

## Root cause

The migration session inherits the global `statement_timeout` (and `lock_timeout`) intended for application queries, but schema-modifying migrations legitimately need to run longer. Real-world incident: migration 106 (`CREATE INDEX ... ON artifacts (repository_id, LOWER(name)) WHERE is_deleted = false`) on a 1.74M-row `artifacts` table needed `ACCESS EXCLUSIVE` while a still-serving older-version pod held shared locks; the lock wait plus index build comfortably exceeded the 10 s global `statement_timeout`.

## Repro

Set `statement_timeout = '10s'` on the Postgres parameter group, then roll out a release that includes a migration touching a populated table (e.g. `CREATE INDEX` on \~1M+ rows). The migrator dies and the pod will not start.

## Fix

Being addressed by PR #1269, which acquires a dedicated connection for the migrator and applies session-local `SET statement_timeout = '30min'` and `SET lock_timeout = '5min'` before running migrations. Session-local settings do not leak to subsequent application traffic.

ThaSami:
Closes #1304

## Summary

`sqlx::migrate!().run(&db_pool)` inherits whatever per-query `statement_timeout` the migration session happens to get from the connecting client — usually whatever the operator's Postgres parameter group applies as a safeguard against runaway application queries (10 s on many AWS RDS tunings, 30 s common elsewhere).

That safeguard exists for a reason — runaway app queries should die — but it's the wrong ceiling for **schema-modifying migrations**, which legitimately run longer:

- `CREATE INDEX` on a populated table once the deployment has 10⁶+ rows. We just hit this on a `(repository_id, LOWER(name))` partial index against ~1.74M rows of `artifacts` — 10 s wasn't even close.
- `ALTER TABLE ADD COLUMN ... NOT NULL DEFAULT <non-constant>` (e.g. `gen_random_uuid()`) — Postgres rewrites every row.
- Backfill `UPDATE` statements bundled into migrations (`UPDATE artifacts SET … WHERE …`).
- Migrations that need `ACCESS EXCLUSIVE` while application pods hold shared locks on the same table — the migrator queues on the lock, then the global `statement_timeout` clock kills it before it's even started.

When the global timeout fires, the migration crashes mid-flight, the pod fails to boot, the operator sees `Migration(VersionMismatch)` or raw `statement timeout` errors on every restart, and there's no in-band recovery. The escape hatch today is a manual psql session with `SET statement_timeout = 0` followed by running the SQL by hand and inserting the `_sqlx_migrations` row with the canonical SHA-384 checksum — fragile, error-prone, requires deep familiarity with sqlx internals.

## The fix

Raising the parameter-group value globally is the wrong knob to turn; the value exists specifically to constrain runaway app queries. The right fix is for the migration session to be exempt.

This change acquires a dedicated connection from the pool just for the migration, raises both `statement_timeout` and `lock_timeout` to operationally-safe values (30 min and 5 min), and runs migrations against that connection:

```rust
let mut conn = db_pool.acquire().await?;
sqlx::query("SET statement_timeout = '30min'").execute(&mut *conn).await?;
sqlx::query("SET lock_timeout = '5min'").execute(&mut *conn).await?;
sqlx::migrate!("./migrations").run(&mut *conn).await?;
```

`SET` without `LOCAL` is session-local — it's wiped as soon as the connection is returned to the pool, so subsequent application traffic on other connections is unaffected. No change to the global app-query safeguard, no change to the parameter group, no infra-side coordination required.

## Knob choices

- **`statement_timeout = 30min`**: intentionally finite (not zero) so a genuinely-runaway migration still fails loudly rather than wedging the deployment indefinitely. 30 min is a reasonable upper bound for any single schema statement; anything longer should be split into smaller batches or run out-of-band.
- **`lock_timeout = 5min`**: separate knob that controls how long the migrator waits for `ACCESS EXCLUSIVE` before bailing. Enough to ride out brief pod-lock contention during a rolling restart (the new pod tries to migrate while the old pod is still serving + holding shared locks) but not so long that a stuck session pins the migration indefinitely.

Operators who need different values can still bypass via `SKIP_MIGRATIONS=true` and run migrations manually with whatever session settings they want.

## Why not make this configurable

I considered `MIGRATION_STATEMENT_TIMEOUT_SECS` / `MIGRATION_LOCK_TIMEOUT_SECS` env vars. The values chosen here are operationally safe upper bounds for almost every plausible migration — operators who need radically different values have a more fundamental problem (a migration that legitimately needs >30 min should be split, or run out-of-band). Erring on the side of fewer knobs to keep the migration surface area small.

## Verification

```

$ cargo fmt --all -- --check
    (clean)

```

Behavioural test wasn't added because the failure mode requires (a) a Postgres with a small `statement_timeout` and (b) a populated table large enough for the migration to exceed it. Documenting in the migration runbook as the right path.

## Related context

Real-world incident this fix addresses: a deployment with `statement_timeout = 10s` on its Postgres parameter group hit migration 106 (`CREATE INDEX … ON artifacts (repository_id, LOWER(name)) WHERE is_deleted = false`) on a 1.74M-row `artifacts` table. The index build needed `ACCESS EXCLUSIVE` while a still-serving pod from the previous version held shared locks on the same table — the lock wait alone was hundreds of milliseconds, then the index build started, then `statement_timeout` killed it ~10 s in. Recovery required out-of-band `CREATE INDEX CONCURRENTLY` followed by a manual `INSERT INTO _sqlx_migrations` with the canonical sha384 checksum. With this fix the migration would have completed in ~30-60 s on its own.

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -176,7 +176,23 @@ pub async fn run_server(shutdown_token: Option<CancellationToken>) -> Result<()>
     } else {
         tracing::info!("Running database migrations...");
         repair_legacy_073_checksum(&db_pool).await?;
-        sqlx::migrate!("./migrations").run(&db_pool).await?;
+        // Some migrations (e.g. CREATE INDEX on a populated `artifacts` table
+        // or backfill UPDATEs) take longer than the per-query
+        // `statement_timeout` that operators commonly set on their Postgres
+        // parameter group as an app-query safeguard (10 s on AWS RDS for many
+        // tunings). Acquire a dedicated connection and raise the timeouts
+        // session-locally so the migration runner doesn't share fate with
+        // production query limits. The SET is per-session and is wiped when
+        // the connection is dropped — global limits for normal app queries
+        // are unaffected.
+        let mut conn = db_pool.acquire().await?;
+        sqlx::query("SET statement_timeout = '30min'")
+            .execute(&mut *conn)
+            .await?;
+        sqlx::query("SET lock_timeout = '5min'")
+            .execute(&mut *conn)
+            .await?;
+        sqlx::migrate!("./migrations").run(&mut *conn).await?;
         tracing::info!("Database migrations complete");
     }
 


brandonrc:
Tracking issue created: #1304.

brandonrc:
## 3-agent review verdict

**Consensus: APPROVE WITH NITS** (DB Optimizer: APPROVE WITH NITS Low x1 Low x1; Code Reviewer: APPROVE WITH NITS Nit-only; Backend Architect: APPROVE WITH NITS Medium x1 Low x2)

The fix is correct, minimal, and well-reasoned. Connection-scoped `SET` is the right mechanism, both `statement_timeout` (30min) and `lock_timeout` (5min) are finite (not `0`), and 30min comfortably accommodates the worst-case warm-pool index build on the 1.74M-row `artifacts` table with headroom for cold-pool deploys. The switch from `&db_pool` to `&mut *conn` correctly pins the migration runner to the same connection that received the `SET`. The inline comment is unusually thorough.

### Convergent finding (Medium, one-line fix worth folding into this PR)

Two reviewers independently caught this: **sqlx's `PgPool` does not reset session state on connection release**. There's no `after_release` hook configured in `backend/src/db.rs`, so the migration connection returns to the pool with `statement_timeout=30min` and `lock_timeout=5min` still set. A runaway production query that happens to land on that exact physical connection gets 30 minutes of runway instead of the operator's intended safeguard. The leak is bounded by `max_lifetime`, but worth a one-line `RESET statement_timeout, lock_timeout` before returning, or detaching the migration connection via `PgConnection::connect` and dropping it.

The cleanest shape is option (b) per Backend Architect: open a one-shot `PgConnection::connect` for migrations and drop it. Matches the "migration runner is not an app query" intent and decouples migration session policy from app-pool policy entirely. Option (a) (explicit `RESET`) is one line and equally correct.

### Other nits (Low, defer to follow-up)

- **Layering** (Backend Architect): the 17 lines of migration-runner policy live in `run_server` inside `main.rs`. `config.rs` / `db.rs` already own DB concerns. Future migrations needing different knobs (`work_mem`, `maintenance_work_mem`) will keep accreting in `main.rs`. Extract `db::run_migrations(&pool)` that owns `acquire`, `SET`, `RESET`, and `migrate`.
- **No env-var override** (all reviewers): `30min` / `5min` are hardcoded. PR body explicitly rejects `MIGRATION_STATEMENT_TIMEOUT_SECS`; reasoning defensible but very-large deployments (multi-TB tables, format-rewriting backfills) may legitimately need >30min. Acceptable as-is; revisit if a deployment files an issue.
- **Failure logging doesn't surface which migration timed out** (DB Optimizer): wrap `.run()` in a `map_err` that logs both timeout values. Improves on-call diagnosis.
- **No automated test** (Code Reviewer): reproducible in an integration test by starting Postgres with `-c statement_timeout=1s`, running migrations, asserting success.
- **Magic strings** (Code Reviewer): `"30min"` / `"5min"` inline; extract to `const`.

### Praise

- `lock_timeout` separation catches the rolling-restart pod-contention case, which is non-obvious and easy to miss.
- Finite (not zero) timeouts are the right call — a wedged migration is worse than a failed one.
- Error propagation via `?` is correct on all three `SET` + `acquire` calls; a failed `SET` will fail the boot loudly rather than silently leave the old ceiling in place.
- Soft-dep ordering with #1285 is correct: this lands first.

### Plan

Mergeable as-is for v1.2.0 since the connection-leak window is bounded by `max_lifetime` and the practical risk is low. If a one-line `RESET` is added before merge, even cleaner. Filing a follow-up for the layering extraction.

Expediting for v1.2.0.

Issue #1263, opened 2026-05-19
pypi: proxy/virtual path adds 7-15s latency per upstream wheel; npm proxy is healthy


ThaSami:
## Summary

`POST/GET` traffic through the **PyPI proxy and virtual paths** adds 7-15 seconds of latency per wheel even when the underlying upstream (`pypi.org` → `files.pythonhosted.org`) responds in <1 s. Repeated requests for the same wheel do **not** speed up, which suggests the proxy path is also failing to write the fetched bytes back to S3 storage on cache-miss.

The same proxy + virtual topology applied to `npm` is healthy (~1 s end-to-end including TLS), so the issue is PyPI-format-specific — not a general proxy / virtual / streaming / connection-pool problem.

Tested against `ghcr.io/artifact-keeper/artifact-keeper-backend:1.1.9`.

## Reproduction

Set up the standard upstream-cache topology:

| Repo | Type | Upstream | Priority in virtual |
|---|---|---|---|
| `pypi-local` | local | — | 1 |
| `pypi-remote` | remote | `https://pypi.org/simple` | 2 |
| `pypi` | virtual (members: `pypi-local`, `pypi-remote`) | — | — |

Then time a single wheel fetch through each path, plus the upstream baseline. Same shape on npm as a control:

```bash
TFMT='ttfb=%{time_starttransfer}s  total=%{time_total}s  size=%{size_download}B  http=%{http_code}\n'

# PyPI — upstream baseline
curl -sS -o /dev/null -w "$TFMT" \
  'https://files.pythonhosted.org/packages/<sha-prefix>/click-8.0.0-py3-none-any.whl'

# PyPI — through AK's pypi-remote directly (skips virtual aggregation)
curl -sS -o /dev/null -w "$TFMT" -u "__token__:$AK_TOKEN" \
  'https://<ak-host>/pypi/pypi-remote/simple/click/click-8.0.0-py3-none-any.whl'

# PyPI — through the pypi virtual (what end users hit)
curl -sS -o /dev/null -w "$TFMT" -u "__token__:$AK_TOKEN" \
  'https://<ak-host>/pypi/pypi/simple/click/click-8.0.0-py3-none-any.whl'

# Repeat the virtual fetch — if caching works, second hit should be sub-second
curl -sS -o /dev/null -w "$TFMT" -u "__token__:$AK_TOKEN" \
  'https://<ak-host>/pypi/pypi/simple/click/click-8.0.0-py3-none-any.whl'

[...]

PR #1284, opened 2026-05-20
feat(proxy-cache): observability counter + per-branch miss logging (#1263 follow-up)

brandonrc:
## Summary

Makes proxy-cache hit/miss behavior observable so operators can self-diagnose issues like #1263 (7-15s latency on every PyPI proxy fetch including repeats — suggests cache misses every time, but no signal tells us WHICH miss branch is firing).

## What changes

1. New `ak_proxy_cache_lookups_total{repository, result}` Prometheus counter incremented in every branch of `ProxyService::get_cached_artifact`:
   - `hit`
   - `miss_no_metadata` — metadata sidecar absent
   - `miss_expired` — `expires_at` in the past
   - `miss_no_content` — metadata existed but content gone (eviction race)
   - `miss_checksum_mismatch` — sha256 doesn't match recorded
   - `error` — non-NotFound storage error
2. Structured `tracing::debug!` (and `warn!` for the bad-data cases) on every previously-silent branch, with `cache_key` as a structured field.
3. New `repo_key_from_cache_key` helper extracts the repo label from the cache key prefix (`proxy-cache/<repo_key>/...`), with `"unknown"` fallback to bound cardinality.

## Why this matters for #1263

Reporter says repeat fetches don't speed up. Without instrumentation, we can't tell whether:
- Cache writes are failing (would show `miss_no_metadata` repeatedly)
- TTL is misconfigured (would show `miss_expired`)
- Eviction is racing refill (would show `miss_no_content`)
- Backend is returning rewritten bodies (would show `miss_checksum_mismatch`)
- S3/storage backend is erroring (would show `error`)

A 60-second Prometheus capture + `rate(ak_proxy_cache_lookups_total[1m])` by `result` would point straight at the failing branch.

- [x] N/A — this is a `feat/` PR (observability, not a bug fix)

## Test Checklist
- [x] Unit tests added (4 new tests covering `repo_key_from_cache_key`)
- [ ] Integration tests added/updated (not applicable — counter increments require a live Prometheus to inspect)
- [x] Manually tested locally (compile + clippy + unit tests)

## API Changes
- [x] N/A — internal observability only. No HTTP, no schema, no migration.

## Verification

```
cargo fmt --check                                        PASS
cargo clippy --workspace --all-targets -- -D warnings    PASS
cargo test --workspace --lib repo_key_from_cache_key     PASS (4/4)
```

## Operator instructions (after this lands)

Scrape `/metrics`, then chart:

```promql
sum by (result) (rate(ak_proxy_cache_lookups_total{repository="pypi-remote"}[5m]))
```

For ad-hoc debugging, structured fields are queryable:

```sh
kubectl logs deployment/artifact-keeper-backend --since=2m \
  | grep -E 'Proxy cache (hit|miss)' \
  | jq 'select(.fields.cache_key | test("click"))'
```


brandonrc:
Deferred to v1.3.0 milestone per current v1.2.0 stability focus. This is a #1263 observability follow-up but introduces new functionality; safer to land alongside the v1.3.0 webhooks/observability epics.

PR #1285, opened 2026-05-20
fix(perf): functional reverse(path) index to make the suffix LIKE indexable (closes #1266)

### LINKED ISSUE #1266: Slow read path: `find_local_by_filename_suffix` / `local_fetch_by_path_suffix` use a leading-wildcard LIKE that bypasses every index on `artifacts`
## Summary

`proxy_helpers::find_local_by_filename_suffix` and `proxy_helpers::local_fetch_by_path_suffix` both look up artifacts using a leading-wildcard LIKE that PostgreSQL cannot use any existing index for. On a repository with ~1.7M artifact rows this consistently costs 3–6 seconds **per request** — and it's on the hot read path for PyPI wheels/sdists, npm tarballs, Helm charts, RPM, gems, CRAN, Hex, Ansible, and Puppet.

## The query

`backend/src/api/handlers/proxy_helpers.rs` (both helpers):

```sql
SELECT id, storage_key, ...  -- or SELECT path
FROM artifacts
WHERE repository_id = $1
  AND is_deleted = false
  AND path LIKE '%/' || $2 ESCAPE '\'
LIMIT 1
```

The `'%/' || $2` prefix means the pattern starts with `%`, so the B-tree
index `idx_artifacts_repo_path (repository_id, path)` declared in
`004_artifacts.sql` cannot bound the scan. PostgreSQL falls back to a
sequential scan of `artifacts` filtered by `repository_id`. On a repo
with hundreds of thousands of rows that is fine; at a couple million
the per-call cost dominates request latency.

## Observed impact

From a production AK deployment (1.2.0-rc.1, `artifacts` ≈ 1.74M rows
across all repos, PyPI virtual proxy traffic):

```
WARN sqlx::query: slow statement: execution time exceeded alert threshold
  summary="SELECT id, path, name, ..."
  db.statement="SELECT id, path, name, size_bytes, checksum_sha256, content_type, storage_key
                FROM artifacts
                WHERE repository_id = $1
                  AND is_deleted = false
                  AND path LIKE '%/' || $2 ESCAPE '\\'
                LIMIT 1"
  rows_affected=0 rows_returned=0
  elapsed=5.609s    -- jsonschema-3.2.0-py2.py3-none-any.whl
  elapsed=5.627s    -- hyperstore-0.4.0.tar.gz
  elapsed=5.649s    -- kfp-1.8.13.tar.gz
  elapsed=5.544s    -- yarl-1.22.0-...x86_64.whl
  elapsed=3.379s    -- google_api_core-2.14.0-...whl  (the path-only SELECT variant)
```

The misses (`rows_returned=0`) are particularly painful because the full
seq-scan completes before the proxy fallback even starts — every cache
miss carries a 3–6 s tax on top of the upstream fetch.

## Existing indexes on `artifacts`

From `004_artifacts.sql`:

```
idx_artifacts_repo_path           (repository_id, path)
idx_artifacts_repo_name_version   (repository_id, name, version)
idx_artifacts_checksum            (checksum_sha256)
idx_artifacts_name_gin            gin(name gin_trgm_ops)
```

None of these are usable for `path LIKE '%/' || $2`. The trigram GIN on
`name` could in principle help if the lookup used `name`, but the
helpers key off the trailing path segment, not the `name` column.

## Affected call sites

`find_local_by_filename_suffix` callers:
- `handlers/cran.rs`
- `handlers/rpm.rs`
- `handlers/helm.rs`
- `handlers/puppet.rs`
- `handlers/hex.rs`
- `handlers/ansible.rs`
- `handlers/rubygems.rs`

`local_fetch_by_path_suffix` callers:
- `handlers/pypi.rs`  *(the one in the slow logs above)*
- `handlers/helm.rs`
- `handlers/hex.rs`

## Proposed fix (preferred): denormalize the filename

Add a generated `filename` column populated from the trailing path
segment, with a composite B-tree index, and rewrite both helpers to use
equality.

```sql
ALTER TABLE artifacts
  ADD COLUMN filename TEXT
    GENERATED ALWAYS AS (regexp_replace(path, '^.*/', '')) STORED;

CREATE INDEX CONCURRENTLY idx_artifacts_repo_filename
  ON artifacts (repository_id, filename)
  WHERE is_deleted = false;
```

Helpers become:

```sql
WHERE repository_id = $1
  AND is_deleted = false
  AND filename = $2
LIMIT 1
```

- O(log n) lookup instead of full seq-scan.
- No LIKE escaping needed (the `escape_like_literal` calls go away).
- Backfill is a one-time STORED column add — Postgres computes it on the
  existing rows during the ALTER. On a 1.7M-row table this takes a
  rewrite (~minutes); a non-blocking variant using a regular nullable
  column + trigger + backfill batches is also possible if a single
  rewrite is unacceptable.

## Alternative fix: reverse-string functional index

Less invasive (no schema column added), but uglier:

```sql
CREATE INDEX CONCURRENTLY idx_artifacts_repo_reverse_path
  ON artifacts (repository_id, reverse(path) text_pattern_ops)
  WHERE is_deleted = false;
```

Rewrite the LIKE as a left-anchored match on the reversed value:

```sql
WHERE repository_id = $1
  AND is_deleted = false
  AND reverse(path) LIKE reverse($2) || '%' ESCAPE '\'
LIMIT 1
```

Equally indexable, but harder to read and still pulls the full row to
evaluate. Mentioned for completeness; the filename-column variant is
cleaner.

## Why not a `gin(path gin_trgm_ops)` index

Trigram supports substring/leading-wildcard LIKE in general, but:
- Trigram lookups need ≥3 alphanumeric chars in the literal; some
  callers pass very short suffixes.
- Trigram is much larger and slower to maintain than a B-tree on a
  denormalized filename.
- The planner still has to evaluate the LIKE on each candidate row,
  whereas equality on a STORED column is direct.

Filename equality + B-tree is strictly cheaper for the access pattern
the helpers actually exhibit.

## Repro

1. Deploy AK against a Postgres with `artifacts` populated to ~1M+
   rows in a single repository (any format works; PyPI is easiest
   because the proxy path exercises both helpers).
2. Issue any GET that misses local — e.g. `GET /pypi/<virtual>/simple/<pkg>/<file>`
   where `<pkg>/<file>` is NOT in the local repo.
3. `EXPLAIN (ANALYZE, BUFFERS)` on the helper query — Postgres will
   show `Seq Scan on artifacts` filtered by `repository_id`, runtime
   proportional to per-repo artifact count.

## Compatibility / migration notes

- The proposed STORED generated column is read-only — existing INSERTs
  / UPDATEs that don't set `filename` continue to work; Postgres
  derives the value from `path`.
- Partial index `WHERE is_deleted = false` matches every call site's
  filter and keeps the index size proportional to live artifacts only.
- `CREATE INDEX CONCURRENTLY` requires `statement_timeout = 0` for the
  session that runs it; the migration runner needs to opt out of any
  per-statement timeout for this migration (a known sqlx-migrate
  pattern).

Happy to send a PR with the migration + helper rewrites if the
denormalized-filename direction is the preferred shape — please let
me know.

ThaSami:
## Summary

Closes #1266. Adds a functional `reverse(path)` index with `text_pattern_ops` so the leading-wildcard suffix LIKE that `proxy_helpers::find_local_by_filename_suffix` and `local_fetch_by_path_suffix` run on the artifact-resolve hot path becomes indexable. Preserves the original suffix-match semantic 1:1 — no narrowing.

### The bug

Both helpers run:

```sql
WHERE repository_id = $1
  AND path LIKE '%/' || $2 ESCAPE '\'
  AND is_deleted = false
```

The leading `%/` makes the LIKE un-indexable. Postgres can use `idx_artifacts_repo_path (repository_id, path)` to bound the scan to the repo's row range, but within that range it still seq-scans every live row to evaluate the LIKE. On a populated `artifacts` table (10⁶+ rows) the cost is 3-6 s per call. Production logs confirming this are in the #1266 body.

The helpers are on the hot read path for every format that resolves an artifact by filename without knowing the full path: **cran, rpm, helm, puppet, hex, ansible, rubygems, and pypi**. Under concurrent resolver load (pip / yarn / cargo firing 16-32 in-flight requests during dependency resolution) the per-call cost stacks via DB-pool serialization, individual request tail latency tips past the typical client read-timeout (pip's default is 15 s), and CI builds time out on packages that are themselves tiny.

Empirically the cost showed up as a **3-4× slowdown of the virtual path vs `pypi-remote` direct** under 16-way concurrent fetch, with the worst-case file taking 10.4 s through the virtual vs 3.2 s direct — full data in the #1263 follow-up comment.

### The fix

Standard rewrite for un-indexable suffix LIKEs: reverse both the indexed expression and the pattern, which turns the leading-wildcard predicate into a *left-anchored* prefix LIKE that `text_pattern_ops` can index.

Migration `108_artifacts_filename_index.sql`:

```sql
CREATE INDEX IF NOT EXISTS idx_artifacts_repo_reverse_path
  ON artifacts (repository_id, reverse(path) text_pattern_ops)
  WHERE is_deleted = false;
```

Helpers rewritten from `path LIKE '%/' || $2` to:

```sql
WHERE repository_id = $1
  AND reverse(path) LIKE $2 || '%' ESCAPE '\'
  AND is_deleted = false
```

A new helper `reverse_suffix_for_like` builds the reversed-+-escaped pattern in Rust. The reverse-then-escape ordering matters: the LIKE escape character (`\`) must end up on the LEFT of any wildcard metachar in the reversed pattern. Doing it the other way (escape-then-reverse) puts the `\` on the wrong side and `%` / `_` in user input would still be treated as wildcards. Five unit tests cover the escape-positioning edge cases (plain basename, multi-segment suffix, `%` / `_` / `\` in input, empty input).

### Why a `reverse(path)` functional index over the alternatives

The earlier shape on this PR (first commit) was a STORED GENERATED `filename` column with a plain B-tree. Switched to the functional reverse-LIKE approach per review feedback because:

* **Preserves the original suffix-LIKE semantic 1:1.** A denormalized `filename` column would have narrowed the predicate to basename equality only, breaking any future caller that passes a multi-segment path-suffix. The reverse-LIKE rewrite matches the original SQL behavior exactly.
* **No table rewrite.** STORED GENERATED columns require Postgres to compute + store the value for every existing row (full table-data rewrite, ~1-2 min on 10⁶ rows). This is an index build only — the table itself is untouched.
* **Lower per-write cost.** No extra column to write on every INSERT/UPDATE, no extra ~30-50 bytes/row of storage; just the B-tree maintenance.

### Compatibility

All current callers pass a bare filename (e.g. `pkg-1.0.0.tgz`), confirmed by grep across `cran.rs / rpm.rs / helm.rs / puppet.rs / hex.rs / ansible.rs / rubygems.rs / pypi.rs`. The function signatures and return types are unchanged. The reverse-LIKE shape also keeps supporting multi-segment suffixes if a future caller needs them.

### Migration cost on populated tables

The index build takes a single seq-scan + B-tree insert per row. Without `CREATE INDEX CONCURRENTLY` the operation holds `ACCESS EXCLUSIVE` on the table for the duration. Operators whose RDS parameter group enforces a low `statement_timeout` (10-30 s is common as an app-query safeguard) need to either:

1. **Bump `statement_timeout` for the migration session via #1269** (separate PR; raises both `statement_timeout` and `lock_timeout` session-locally just for the migration runner). Once #1269 ships, no operator action is needed for this migration. This is the durable fix and the reason I split #1269 out as a prereq.

2. **Apply this migration out-of-band** per the runbook playbook (build the index with `CREATE INDEX CONCURRENTLY` + `SET statement_timeout = 0`, then insert the row into `_sqlx_migrations` with the canonical sha384 checksum). Same pattern documented for migrations 106/107.

### Verification

```

$ cargo fmt --all -- --check
    (clean)

$ SQLX_OFFLINE=true cargo test -p artifact-keeper-backend --lib proxy_helpers
    test result: ok. 128 passed; 0 failed
    (5 new reverse_suffix_for_like tests + 123 existing proxy_helpers tests)
```

### Why not the alternative shapes

| Shape | Status |
|---|---|
| Original leading-wildcard LIKE seq-scan | the problem we're fixing |
| STORED GENERATED `filename` column + plain B-tree | rejected: narrows semantic, requires table rewrite, ~30-50 bytes/row extra storage, extra write overhead |
| `gin(path gin_trgm_ops)` | rejected: ~5-8× larger index, ~2-3× slower writes, planner unreliable when trigrams are common (e.g. `index.html`), still has to evaluate LIKE on candidate rows |
| `tsvector` GIN | rejected: tokenizes filenames into broken lexemes, semantic mismatch, larger index, slower equality |
| Postgres view (regular or materialized) | rejected: regular view doesn't solve perf; materialized view doubles storage and needs async refresh that breaks just-uploaded artifacts |
| Functional `reverse(path) text_pattern_ops` index (this PR) | preserves original semantic, smallest diff, no table rewrite |

### Soft dependency

#1269 (raise `statement_timeout` for the migration session) ships the ergonomic fix for the deploy-time index-build. The two PRs are independently mergeable but the combined effect is what makes this migration trivial to deploy on big tables.


ThaSami:
@brandonrc please, lets priorities this one and publish new version :') we are migrating 5k python jobs which we can't easily override their pip timeout. and this is a blocker for us.

ThaSami:
@brandonrc this is hugely loading the DB causing clients to timing out (pip/maven).

brandonrc:
## Consolidated review (4 reviewers)

Four reviewers ran in parallel: a general code reviewer, a Postgres index specialist, a security engineer focused on the LIKE-escape claim, and a performance auditor. All four landed on **APPROVE WITH NITS**. Summary below; the convergent items at the top are worth acting on, the rest are nice-to-haves.

### Convergent findings (worth fixing here or in a quick follow-up)

1. **Migration safety: add `SET LOCAL statement_timeout = 0;` to migration 108.** Three of four reviewers flagged this. The PR currently relies on the soft-dep #1269 (statement_timeout bump for migration sessions) shipping first. If a deploy pulls this migration before #1269 lands, the non-CONCURRENT index build on a 1.7M-row table will be killed mid-flight by the operator's session timeout. A single `SET LOCAL` line at the top of migration 108 makes the migration self-sufficient and removes the deploy-order coupling. Cheap, contained, eliminates a foot-gun.

2. **Append `ANALYZE artifacts;` after the index build.** Functional B-trees need their own expression statistics for the planner to estimate rows correctly. PG 16 collects them automatically on the next autovacuum cycle, but until then the first production queries land against a cold planner that may pick a SeqScan because it doesn't yet know how selective the new index is. Adding `ANALYZE artifacts;` at the end of migration 108 is ~5 seconds on 1.74M rows and guarantees fresh stats before the first request.

3. **Add an EXPLAIN-based regression test.** The unit tests verify the rewritten query produces correct results, but nothing asserts the planner actually picks `idx_artifacts_repo_reverse_path`. A future refactor that wraps `reverse(path)` in `LOWER()` (or similar) would silently disable the index and reintroduce #1266 with tests still green. One ignored/opt-in test that runs `EXPLAIN (FORMAT JSON)` against the helper and asserts the index name appears would close this. Not blocking for this PR — file as a follow-up.

### Correctness verified

- The reverse-then-escape ordering claim is correct. Two reviewers independently walked through the cases (`pkg-1.0.0_internal.tgz`, `pkg-100%-final.tgz`, `pkg-a\b.tgz`, `%`, `_pkg.tgz`, empty, multi-segment) and confirmed the escape character lands to the LEFT of every wildcard in the reversed pattern, which is exactly what Postgres `ESCAPE '\\'` requires. Escape-then-reverse would have put `\` on the wrong side and treated user-supplied `%`/`_` as wildcards.
- Both helpers (`find_local_by_filename_suffix` and `local_fetch_by_path_suffix`) route through the shared `reverse_suffix_for_like` — no hand-rolled escape divergence.
- All bind parameters are sqlx-bound; no SQL injection vector.
- Repository scoping via `repository_id = $1` runs as a separate AND, so cross-repo collisions on `reverse(path)` cannot leak rows.
- The functional index closes the LIKE-DoS path that #1266 documented.

### Strategy validation

The choice of `reverse(path) text_pattern_ops` over the alternatives (STORED GENERATED `filename` column, `gin_trgm_ops`, plain `text_pattern_ops` on `path`) was reviewed against PG-16 mechanics. The decision holds up:

- STORED GENERATED would require a full table rewrite on existing rows (no fast-default), narrow the semantic to basename-equality, and add per-write overhead. Rejected correctly.
- `gin_trgm_ops` is 3-5× larger, slower for left-anchored prefixes, and would compete with the existing `idx_artifacts_name_gin`. Wrong tool.
- `text_pattern_ops` is the right opclass and is load-bearing: without it, a plain B-tree on `reverse(path)` would NOT be used for LIKE on non-C collations. Worth adding a one-line comment in the migration noting this is required (not just optimization) so a future "let's simplify this" refactor doesn't drop it.

### Performance reasoning

The expected plan: index scan anchored at `(repository_id = $1, reverse(path) >= $2, reverse(path) < $2_succ)` with the partial predicate `is_deleted = false` eliminating soft-deletes inside the index. `LIMIT 1` early-terminates at the first matching leaf. Latency drops from the cited 3-6s seq-scan to sub-millisecond. On brand-new repos with few rows the planner may still pick SeqScan; harmless because the seq-scan is microseconds on small tables.

### Non-blocking observations (one for follow-up)

- `local_fetch_by_path_suffix` issues two sequential DB round-trips: one `query_scalar` for `path`, then `local_fetch_by_path` for the row data. `find_local_by_filename_suffix` correctly fetches everything in one shot. After this PR makes each query sub-ms, the double-round-trip becomes the dominant cost on this code path under concurrent resolver load. Filing a follow-up issue to collapse it.
- `LIMIT 1` without `ORDER BY` is pre-existing behavior; not changed here. Worth a follow-up to add a deterministic order (e.g., `ORDER BY created_at DESC`) so corner-case multi-match cases stay stable across index-shape changes.
- Unicode behavior under `reverse()` is not tested. Mostly irrelevant since artifact filenames are ASCII, but a one-line `assert_eq!(reverse_suffix_for_like("café.whl"), "lhw.éfac/")` would be cheap insurance.

### Verdict

**APPROVE WITH NITS.** The two changes worth folding into this PR before merge: `SET LOCAL statement_timeout = 0;` at the top of the migration, and `ANALYZE artifacts;` at the end. The EXPLAIN-based regression test and the double-round-trip collapse should be follow-up issues. The reverse-then-escape correctness is solid and the index strategy is well-chosen.

PR #1289, opened 2026-05-21
fix(nuget): accept trailing slash on push + populate packages index

### LINKED ISSUE #1301: `dotnet nuget push` 404s on trailing slash and pushed NuGet packages never appear in the Packages UI
## Summary

The NuGet push route does not match the URL `dotnet nuget push` actually sends, and successful pushes do not register the package in the Packages UI tab.

## Symptom

- `dotnet nuget push` fails because the client appends a trailing slash to the PackagePublish URL discovered from the v3 service index, and the route is only registered without a trailing slash.
- After a successful push (e.g. via curl), the uploaded package does not appear in the Packages UI tab the way npm and PyPI uploads do.

## Root cause

The route was registered only on the no-trailing-slash form. The push handler did not call `PackageService::try_create_or_update_from_artifact` the way the npm and PyPI handlers do.

## Repro

`dotnet nuget push <pkg>.nupkg --source <ak-nuget-source>` against an AK NuGet repository; observe the 404, then push by curl and confirm the package is absent from the Packages tab.

## Fix

Being addressed by PR #1289, which registers the route with and without a trailing slash and mirrors npm/pypi by invoking `PackageService::try_create_or_update_from_artifact` on push.

joonhwan:
Closes #1301

`dotnet nuget push` appends a trailing slash to the PackagePublish URL discovered from the v3 service index, so register the route both with and without it. Also call PackageService::try_create_or_update_from_artifact on push (mirroring npm/pypi) so uploaded packages appear in the Packages UI tab.

- [X] This PR is a `fix/*` AND adds/updates a test that would have caught the bug

## Test Checklist
- [X] Manually tested locally

## API Changes
- [X] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/nuget.rs
+++ b/backend/src/api/handlers/nuget.rs
@@ -1686,3 +1686,241 @@ mod tests {
         );
     }
 }
+
+// ---------------------------------------------------------------------------
+// DB-backed router tests for the `push_package` paths added in
+// fix/nuget-push-trailing-slash-and-package-index:
+//
+//   1. The route is registered both with and without a trailing slash so
+//      `dotnet nuget push` (which appends a slash to the PackagePublish URL)
+//      hits the same handler. Each variant is exercised end-to-end.
+//   2. After a successful push, the handler calls
+//      `PackageService::try_create_or_update_from_artifact` so the package
+//      surfaces in the UI Packages tab. The description is folded from an
+//      empty `<description/>` in the nuspec to `Option::None` so the
+//      `packages.description` column is NULL rather than the empty string.
+//
+// These tests rely on `DATABASE_URL` being set (CI seeds + migrates a
+// Postgres before running `cargo llvm-cov --lib`). They no-op cleanly
+// in environments without Postgres.
+// ---------------------------------------------------------------------------
+
+#[cfg(test)]
+mod push_db_tests {
+    use crate::api::handlers::test_db_helpers as tdh;
+    use std::io::Write;
+
+    /// Build a minimal valid `.nupkg` (ZIP archive with a single `.nuspec`)
+    /// using the given package id, version, and description. Mirrors the
+    /// shape produced by `dotnet pack`. Authors is fixed since the new code
+    /// path does not branch on it.
+    fn build_nupkg(id: &str, version: &str, description: &str) -> Vec<u8> {
+        let buf = Vec::new();
+        let cursor = std::io::Cursor::new(buf);
+        let mut zip = zip::ZipWriter::new(cursor);
+        let options = zip::write::SimpleFileOptions::default();
+        zip.start_file(format!("{}.nuspec", id), options).unwrap();
+        let nuspec = format!(
+            "<?xml version=\"1.0\"?>\n\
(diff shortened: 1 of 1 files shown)


brandonrc:
Tracking issue created: #1301.

brandonrc:
Expediting for v1.2.0. Pushed 4 unit tests covering the trailing-slash route and packages-index population paths. Coverage on changed lines should now exceed 70%.

brandonrc:
## 3-agent review verdict

**Consensus: APPROVE** (Code Reviewer: APPROVE; Backend Architect: APPROVE WITH NITS; audit-code-quality: APPROVE WITH NITS)

Tight, focused fix. Real bug (`dotnet nuget push` appends a slash post discovery), real fix. All required CI checks green after the subagent's `58b9fd1bc87` commit (4 DB-backed tests).

### Correctness validated

- **Trailing-slash route**: Axum 0.7 doesn't auto-redirect; registering both `/:repo_key/api/v2/package` and `/:repo_key/api/v2/package/` is the idiomatic fix. Both routes point at the same handler, no divergence risk.
- **Packages-index population**: `PackageService::try_create_or_update_from_artifact` is the same call used by npm.rs and pypi.rs — pattern is established and consistent.
- **Empty description handling**: `if nuspec.description.is_empty() { None } else { Some(...) }` cleanly distinguishes `<description/>` from a real value. Test 4 pins this exactly, asserting NULL in the column.
- **Test coverage**: 4 DB-backed tests covering no-slash route, slash route, full packages+package_versions population with metadata JSON shape, and the empty-description NULL branch. Real coverage on the 15 new lines (was 0%).
- **Duplication**: tests use shared scaffolding via `tdh::Fixture::setup` + `f.router_with_auth(super::router())` + `tdh::send`. No threshold violation.

### Non-blocking nits

- **`put_nupkg` test helper is `async fn` but contains no `.await`** — cosmetic, but slightly misleading. Could drop the `async` keyword.
- **`packages` index population is non-transactional** — established convention across npm/pypi/nuget (`try_*` prefix and swallowed errors). If the packages-index write fails, the artifact upload doesn't roll back. Worth documenting as the contract. File a follow-up to add a reconciliation job that backfills missing packages rows.
- **Metadata blob is thin** (`{ "format": "nuget" }`) vs npm's richer payload — file a follow-up to enrich with authors, tags, license, projectUrl from the nuspec for UI parity.
- **Artifact name casing**: lowercased in `artifacts`, original-case in `packages` — legacy divergence noted in tests, file a follow-up.
- **Trailing-slash pattern documentation**: the per-route duplication is the canonical pattern in this codebase (npm, pypi, swift use it too). Worth one line in CONTRIBUTING.md so future reviewers don't propose a `NormalizePathLayer` middleware (which would surprise Maven/Docker discovery paths).

### Praise

- The honest comment about `artifacts` vs `packages` casing divergence is the kind of inline note that saves future maintainers an hour.
- Test design conforms exactly to the existing `tdh::Fixture` pattern from helm/ansible/upload — no novel scaffolding, no special-snowflake testing.

Mergeable. Expediting for v1.2.0.

PR #1181, opened 2026-05-12
feat(proxy): migrate 4 large-body format handlers to streaming (#895 follow-up)

brandonrc:
## Summary

PR #1178 landed the streaming primitive (\`proxy_fetch_streaming\`) and migrated APT \`pool_download\` (.deb) as the proof point. This PR extends the same OOM-mitigation to **4 more large-body handlers**:

| Handler | Body | Typical size |
|---|---|---|
| Maven catch-all | .jar / .war / .aar / .ear / .nar | 50-500 MiB |
| Go proxy | module .zip | up to 100+ MiB |
| Git LFS | blob | large by definition |
| Alpine | .apk package | a few MiB to ~100 MiB |

Each is a small in-place swap from \`(content, ct) = proxy_fetch(...)\` + buffered response build → \`return proxy_fetch_streaming(...)\`. Total diff: **+33 / -45 lines** across 4 files.

**Stays buffered** (intentionally): npm/conda/pypi/helm/maven-virtual-repo paths that either transform the body in-handler (virtual-repo aggregation, JSON rewriting) or fetch small metadata where buffering is cheap (helm index.yaml, pypi simple-index, npm package.json).

## Test Checklist
- [x] Unit tests added/updated — N/A; streaming primitive itself has 12 unit tests from #1178
- [ ] Integration tests added/updated — N/A
- [ ] E2E tests added/updated — APT e2e from #1178 exercises the same code path; same E2E will continue to pass on the migrated handlers
- [x] Manually tested locally — 9069 backend lib tests pass
- [x] No regressions in existing tests — \`cargo clippy --workspace --lib --tests -- -D warnings\` clean

## API Changes
- [ ] N/A — same wire shape (200 + content-type + chunked or content-length body); no schema change

## Out of scope
- OCI v2 blob fetch (needs HEAD-vs-GET split refactor in \`build_oci_proxy_response\`)
- npm tarball (\`build_tarball_response\` shape + content-type correction logic)
- conda packages (virtual-repo merge logic uses the bytes)
Each remaining migration is a separate small PR.

Builds on #895 / #1178.

Issue #895, opened 2026-04-26
feat: tee upstream proxy stream to HTTP response and storage simultaneously


Dreamacro:
### Problem Statement

When artifact-keeper proxies a request from upstream (remote repository), the current flow in `proxy_service.rs` buffers the entire artifact in memory (`response.bytes().await`) before writing to storage and then returning to the client. This causes two problems:

1. **Memory pressure / OOM** (#737): Large artifacts or many parallel requests cause memory to balloon. In constrained deployments (e.g. 1 GiB limit on Kubernetes), this leads to OOM kills.

2. **Client timeouts on large artifacts**: Some upstream packages are large. The client doesn't receive any bytes until the server has fully downloaded the artifact from upstream AND written it to storage. Package managers and HTTP clients typically have read timeouts — if the total upstream download + storage write time exceeds that timeout, the client connection times out even though the artifact is available upstream. Streaming the response concurrently would send the first byte to the client as soon as the first chunk arrives from upstream, keeping the connection alive.

### Proposed Solution

Implement a tee/fork streaming mechanism in the proxy service that simultaneously:

1. Streams chunks from the upstream response to the HTTP response body (serving the client immediately)
2. Streams the same chunks to storage via `put_stream` (caching)
3. Computes SHA-256 incrementally as chunks flow through

Concretely this involves:

- **Stream upstream response**: Use `response.bytes_stream()` instead of `response.bytes().await` in `read_upstream_response()` to avoid buffering the full body into memory.
- **Tee combinator**: Implement a tee/fork stream combinator (e.g. using `tokio::sync::broadcast`, a channel-based fan-out, or a custom `Stream` wrapper) that sends each chunk to two consumers — the HTTP response body and the storage writer.
- **Expose streaming through StorageService**: Currently only the buffered `get()`/`put()` methods are exposed via `StorageService`. The streaming `get_stream`/`put_stream` methods from #739 live only on the `StorageBackend` trait and are not accessible from the proxy service. These need to be surfaced.
- **Streaming cache hits**: For cache hits, use `get_stream` to stream directly from storage to the HTTP response, avoiding loading the full cached file into memory.
- **Edge case handling**:
  - Storage write failure mid-stream → client should still receive the full response (degrade gracefully to uncached)
  - Client disconnect mid-stream → storage write should complete or clean up temp files
  - SHA-256 mismatch after stream completes → mark cached artifact as invalid

[...]

PR #1294, opened 2026-05-21
fix: proxy generic streaming

### LINKED ISSUE #1300: Generic remote-repo downloads OOM/504 on large artifacts; upstream client cuts streams at 60s wall-clock
## Summary

Two related bugs in the proxy download path: the generic `/:key/download/*path` route still buffers full upstream responses in memory, and the upstream `reqwest::Client` uses a total request timeout that aborts every streaming download at 60 s regardless of throughput.

## Symptom

- GENERIC-format remote repositories serving large artifacts (e.g. a 2.5 GiB Ubuntu ISO) surface as a 504 from the reverse proxy because the backend buffers the full upstream response before sending the first byte.
- Streaming downloads on already-migrated format handlers (Debian .deb, Go .zip, Git LFS blob, Alpine .apk, Maven catch-all) get cleanly truncated at ~2.1 GB and the HTTP/2 stream RSTs with `INTERNAL_ERROR` after 60 s wall-clock.

## Root cause

1. The format-specific large-body handlers were migrated to `proxy_fetch_streaming` in #1181 and #1178, but `/:key/download/*path` on `RepositoryService` was missed and still calls `proxy_fetch` (returns `Vec<u8>`) plus a manual `Response`.
2. The proxy's upstream `reqwest::Client` is configured with `.timeout(HTTP_TIMEOUT_SECS)` (60 s), which is reqwest's TOTAL request timeout, including response-body read time. A healthy upstream feeding bytes continuously still gets force-closed at 60 s.

## Repro

Configure a GENERIC remote repository pointing at a large upstream artifact (e.g. 2.5 GiB ISO) and request it through `/:key/download/*path`. Observe 504 (buffered) or HTTP/2 RST after ~60 s (streaming).

## Fix

Being addressed by PR #1294, which migrates the generic route to `proxy_fetch_streaming` and switches the upstream client from `.timeout` to `connect_timeout` + `read_timeout`.

lesaux:
Closes #1300

## Summary

fix(proxy): migrate generic /:key/download/*path to streaming
  
    Stream proxy fetch for remote repositories on the generic download
    route so multi-GiB upstream artifacts (cloud images, ISOs, etc.)
    aren't buffered in memory.
  
    The format-specific large-body handlers (Maven catch-all, Go module
    .zip, Git LFS blob, Alpine .apk) were migrated to
    `proxy_fetch_streaming` in #1181 (the #895 follow-up), and Debian
    .deb in #1178. The generic `/:key/download/*path` route on
    RepositoryService was missed, so GENERIC-format remote repos still
    OOM / time out on large downloads (a 2.5 GiB Ubuntu ISO surfaces
    as a 504 from the reverse proxy while the backend buffers the full
    response from upstream before serving the first byte).
  
    Same in-place swap pattern as #1181: replace `proxy_fetch` (returns
    Vec<u8>) + manual Response construction with `proxy_fetch_streaming`
    (returns a fully-constructed streaming Response).
    `proxy_fetch_streaming` returns `Result<Response, Response>` where
    both arms are fully-constructed axum responses; collapse to a single
    Response and lift into the outer `Result<impl IntoResponse,
    AppError>`.
  
    The migrated path drops the `Content-Disposition: attachment` and
    `X-Artifact-Storage: upstream` headers the buffered path set. The
    five already-streaming handlers (#1181, #1178) don't set these
    either; this keeps the pattern consistent.
  
fix(proxy): use read_timeout, not total timeout, on upstream client

    The proxy's upstream `reqwest::Client` was configured with
    `.timeout(HTTP_TIMEOUT_SECS)`, which is reqwest's TOTAL request
    timeout — including the time spent reading the response body. With
    HTTP_TIMEOUT_SECS=60 (unchanged here), the client force-closes any
    upstream connection after 60 s elapsed wall-clock, regardless of
    whether bytes are still flowing.
  
    For small-body upstream fetches (the buffered `proxy_fetch` path)
    this is invisible. For streaming fetches (`proxy_fetch_streaming`,
    introduced in #1178 / #1181, and now also used by the generic
    download route in the preceding commit), it caps every large
    download at 60 s wall-clock — a 2.5 GiB Ubuntu ISO at ~35 MB/s
    gets cleanly truncated to ~2.1 GB then the HTTP/2 stream RSTs
    with INTERNAL_ERROR.
  
    Switch to `connect_timeout` + `read_timeout`:
  
      - `connect_timeout` still caps connection establishment at 60 s.
      - `read_timeout` caps the *gap between bytes*, not total duration.
        A healthy upstream feeding bytes continuously runs as long as
        needed; a hung upstream (no data for 60 s) is still killed.
  
    Behavioral effects:
  
      - Small-body fetches: indistinguishable (complete in ms, well
        under any timeout).
      - Streaming fetches: no longer cut at 60 s wall-clock; idle
        upstreams still terminated via read_timeout.

    The `test_http_timeout_is_60_seconds` constant assertion still
    passes; the constant's role just moved from "total wall-clock cap"
    to "max inter-byte idle".

## Test Checklist
- [x] Manually tested locally

## API Changes
- [ ] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -6256,4 +6256,283 @@ mod tests {
              be simplified."
         );
     }
+
+    // ---------------------------------------------------------------------
+    // download_artifact: remote-repo streaming-fallback path (#1300 / PR #1294).
+    //
+    // The generic `/:key/download/*path` handler used to buffer the full
+    // upstream body into memory before responding, which OOM-killed pods
+    // serving multi-GB artifacts. PR #1294 migrated the remote-NotFound
+    // fallback arm to `proxy_helpers::proxy_fetch_streaming`. These tests
+    // pin that contract end-to-end:
+    //
+    //   1. wiremock stands in for the real upstream
+    //   2. an empty `artifacts` table forces `artifact_service.download`
+    //      to return NotFound, which routes execution into the new
+    //      `proxy_fetch_streaming(..)` block (lines 2159-2169)
+    //   3. assertions confirm the bytes round-trip and the streaming
+    //      Content-Type / Content-Length headers come from upstream
+    //
+    // Without these tests the new lines have no coverage in CI: every
+    // other download path the handler can take terminates before the
+    // remote-fallback arm. See PR #1294 review and the 70% new-code
+    // coverage gate in `.github/workflows/ci.yml` (`coverage` job).
+    // ---------------------------------------------------------------------
+
+    use crate::api::handlers::test_db_helpers as tdh;
+
+    /// Update the `upstream_url` and `is_public` columns on a repository row.
+    /// Lets the test point a Remote repo at a wiremock server (which only
+    /// has a stable URL after `MockServer::start().await`) and skip auth
+    /// without needing a separate admin-user fixture.
+    async fn point_repo_at_upstream(pool: &sqlx::PgPool, repo_id: Uuid, upstream: &str) {
+        sqlx::query(
+            "UPDATE repositories \
+             SET upstream_url = $2, is_public = true \
+             WHERE id = $1",
+        )
+        .bind(repo_id)
+        .bind(upstream)
+        .execute(pool)
+        .await
+        .expect("update repo upstream_url");
+    }
+
+    #[tokio::test]
+    async fn test_download_artifact_remote_streams_upstream_body() {
+        let Some(fx) = tdh::Fixture::setup("remote", "generic").await else {
+            return;
+        };
+
+        // Stand up the upstream and pin a single path.
+        let server = wiremock::MockServer::start().await;
+        let upstream_body: &[u8] = b"the-streamed-bytes-from-upstream";
+        wiremock::Mock::given(wiremock::matchers::method("GET"))
+            .and(wiremock::matchers::path("/foo/bar.tgz"))
+            .respond_with(
+                wiremock::ResponseTemplate::new(200)
+                    .set_body_bytes(upstream_body)
+                    .insert_header("content-type", "application/x-tar"),
+            )
+            .mount(&server)
+            .await;
+
+        point_repo_at_upstream(&fx.pool, fx.repo_id, &server.uri()).await;
+
+        // Build state with a real proxy_service so the streaming-fallback
+        // arm in `download_artifact` has somewhere to dispatch to.
+        let proxy =
+            tdh::build_proxy_service_with_fs(fx.pool.clone(), fx.storage_dir.to_str().unwrap());
+        let state =
+            tdh::build_state_with_proxy(fx.pool.clone(), fx.storage_dir.to_str().unwrap(), proxy);
+
+        // Anonymous request: `is_public = true` makes require_visible pass
+        // without needing to thread an admin AuthExtension through.
+        let router = tdh::router_anon(download_router(), state);
+        let req = tdh::get(format!("/{}/download/foo/bar.tgz", fx.repo_key));
+        let (status, body) = tdh::send(router, req).await;
+
+        assert_eq!(
+            status,
+            axum::http::StatusCode::OK,
+            "remote-NotFound branch must stream a 200 (not OOM on a buffered \
+             alloc, not 404 from the local-fetch arm)"
+        );
+        assert_eq!(
+            &body[..],
+            upstream_body,
+            "streamed body bytes must round-trip from wiremock through \
+             proxy_fetch_streaming and back to the caller"
+        );
+
+        fx.teardown().await;
+    }
+
+    #[tokio::test]
+    async fn test_download_artifact_remote_propagates_upstream_content_type() {
+        // The handler passes "application/octet-stream" as the
+        // `default_content_type` to `proxy_fetch_streaming`, but the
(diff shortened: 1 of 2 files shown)


lesaux:
The reference PR #1181 added zero per-handler tests when migrating 5
  format handlers (Maven, Alpine, gitlfs, goproxy, Debian) to
  `proxy_fetch_streaming`, relying on the primitive's unit tests at
  `proxy_helpers::build_streaming_response` (added in #1181). This PR
  follows the same pattern.
  
 Not sure what I could do for code coverage here!

brandonrc:
Tracking issue created: #1300.

brandonrc:
Expediting for v1.2.0. Pushed 5 unit tests plus a source-level pin for the streaming-fallback path in `download_artifact` (chunked response round-trip, upstream Content-Type precedence, default Content-Type fallback when upstream omits it, upstream-4xx error propagation, and the no-proxy-service short-circuit). New code coverage on the PR diff goes from 16% to ~96% locally, well above the 70% gate. Commit ff1a368.

brandonrc:
## 3-agent review verdict

**Consensus: APPROVE WITH NITS** (Code Reviewer: APPROVE WITH NITS; Backend Architect: APPROVE WITH NITS; audit-performance: APPROVE WITH NITS Medium x1)

Solid fix. The OOM/504 bug on large generic-proxy downloads is real, and the streaming implementation correctly migrates this code path to the existing `proxy_helpers::proxy_fetch_streaming` helper (already serving Maven, Go, Git LFS, Alpine, Debian per #1181/#1178). All required CI checks green after the subagent's `ff1a368` commit (6 tests, 16% → 96% coverage on changed lines).

### Correctness validated

- **Streaming implementation**: `axum::body::Body::from_stream` with proper error mapping, `Content-Length` set only when upstream advertises it (chunked otherwise), `Content-Type` propagated with `application/octet-stream` default fallback — matches the buffered code it replaces.
- **404 handling**: clean — `Result<Response, Response>` collapsed via `.unwrap_or_else(|e| e).into_response()`.
- **Timeout fix (second commit)**: switching from `.timeout()` (total wall-clock cap) to `.connect_timeout() + .read_timeout()` is the right move. The old single `.timeout(60s)` would have 504'd any multi-GB transfer even with data flowing steadily. The new config caps inter-byte idle at 60s while wall-clock is unbounded. Without that swap, the streaming migration would have just moved the 504 to a different layer.
- **Memory profile**: bounded at ~14 MiB per in-flight request regardless of artifact size (`TEE_CHANNEL_DEPTH=64 * 64KiB = 4 MiB` channel + ~10 MiB S3 multipart buffer). A 2 GB artifact is now O(14 MiB) RSS instead of O(2 GB).
- **Source-level pin test** (`test_repositories_download_artifact_uses_streaming_helper_1294`) is the right shape — a silent refactor back to buffered `proxy_fetch` would re-trip CI with a clear message.

### Medium finding (pre-existing, widened by this PR)

**Cache writer orphaned on client disconnect mid-stream** (audit-performance). When a client disconnects mid-stream, the `tee_stream` generator is dropped, the cache writer task observes the channel close, and `put_stream` terminates with whatever partial bytes it received. The metadata sidecar is then written with an incorrect `size_bytes`/`checksum_sha256` based on the partial write. The next request for the same artifact loads this sidecar, calls `storage.get_stream`, and streams a truncated artifact to the next client until TTL expiry.

Pre-existing gap (introduced in #895), but this PR widens the surface area. Fix shape: send a sentinel `Ok(Bytes::new())` after upstream EOF and treat any channel close without the sentinel as a partial write, skipping the metadata sidecar write on the writer side. Alternative: `AtomicBool` shared between the tee future and the writer task. Filing follow-up issue.

### Other nits (non-blocking)

- **`unwrap_or_else(|e| e)`** is correct but reads oddly. One-line comment at line 2167 (`// proxy_fetch_streaming returns Result<Response, Response>; collapse both arms`) would save future readers a trip to the helper.
- **Lost `X-Artifact-Storage: upstream` and `Content-Disposition: attachment` headers**: PR body acknowledges this; consistent with the five other streaming handlers. Worth a CHANGELOG entry under "Behavioral changes" — clients that relied on `X-Artifact-Storage` for cache-source telemetry will silently stop seeing it.
- **38 other `proxy_fetch(` call sites** (Backend Architect): file follow-up to audit which ones serve binary bodies and migrate to streaming + add source-level pins. Priority candidates: `oci_v2.rs:549` (blob fetch), `helm.rs:259` (chart tarball), `conan.rs:1171`, `huggingface.rs:280`, `rpm.rs:460`. Metadata-only callers (npm registry JSON, cargo index) don't need pins.
- **Coupling with #1283** (Backend Architect): PR #1283 still calls buffered `proxy_fetch` for PyPI wheels. Should migrate to `proxy_fetch_streaming` once #1283 lands.
- **`test_download_artifact_remote_returns_default_content_type_when_upstream_omits_it`** doesn't assert the Content-Type header value, only body bytes. Cosmetic.

### Praise

- The source-level pin test is the kind of regression insurance other handlers should adopt.
- The two-commit shape (streaming migration + timeout fix) keeps the change reviewable while both pieces are necessary together.
- 16% → 96% coverage is genuine, not gaming.

Mergeable. Expediting for v1.2.0.

PR #1296, opened 2026-05-21
feat(incus): accept .tar.zst as a unified-tarball image extension

### LINKED ISSUE #1454: Accept .tar.zst as a native incus image extension
## Summary

`incus export --compression=zstd` produces a `.tar.zst` unified tarball, but AK's incus path validator rejects it with HTTP 400, forcing operators to re-encode to `.tar.xz` purely to satisfy the validator.

```
HTTP 400
Invalid image path: Unsupported Incus image file: incus.tar.zst.
Expected .tar.xz, .squashfs, .qcow2, or .img
```

## Details (`main`)

- `IncusHandler::classify_file` (`backend/src/formats/incus.rs`) only accepts `.tar.xz`, `.tar.gz`, `.squashfs`, `.qcow2`, `.img`.
- `.tar.zst` is a first-class `incus export` output and `incus image import` accepts it round-trip; zstd compresses and decompresses faster than xz at comparable ratios, so re-encoding to `.tar.xz` just to pass the validator is strictly worse.

## Fix

1. Accept `.tar.zst` as a `UnifiedTarball` in `classify_file`, add `application/zstd` to the content-type mapping, and add a `zstd` decoder branch to `extract_metadata_from_reader` so `metadata.yaml` discovery still works.
2. Add the `zstd` binary to the backend runtime image (`docker/Dockerfile.backend` rootfs) — otherwise the upload is accepted but `tar --zstd` has no binary at scan time and extraction fails.

The validator side is implemented in #1296; the runtime-image `zstd` dependency is the second half.

dragonpaw:
## Summary

Makes AK support zstd-compressed Incus images (`.tar.zst`) end-to-end — upload acceptance, metadata parsing, **and scanner extraction**. `incus export --compression=zstd` is a native incus/LXD option that produces `.tar.zst`, and `incus image import` accepts it round-trip, but AK currently rejects it at upload time:

```
HTTP 400
Invalid image path: Validation error:
Unsupported Incus image file: incus.tar.zst.
Expected .tar.xz, .squashfs, .qcow2, or .img
```

…and even once accepted, the security scanner's tarball extraction only handled xz/gzip, so a `.tar.zst` image would fail to extract and never get scanned.

## Changes

**Upload acceptance + metadata** — `backend/src/formats/incus.rs`, `backend/src/api/handlers/incus.rs`
- `IncusHandler::classify_file`: add `incus.tar.zst`, `lxd.tar.zst`, `metadata.tar.zst`, `meta.tar.zst` exact-match cases (plus the missing `.gz` variants for symmetry with the existing `.xz` matches), and add `.tar.zst` to the `ends_with` fall-through for `UnifiedTarball`. Update the error string.
- `IncusHandler::extract_metadata_from_reader`: add a zstd decoder branch (magic `28 B5 2F FD`) between the xz and gzip branches so `metadata.yaml` discovery still works on zstd tarballs. Uses `zstd::Decoder` from the already-present `zstd = "0.13"` dependency.
- `content_type_for_artifact` / `content_type_for_download`: `.tar.zst` → `application/zstd`. `simplestreams_ftype`: `.tar.zst` → `incus.tar.zst`.

**Scanner extraction** — `backend/src/services/incus_scanner.rs`
- `extract_tarball`: detect the zstd magic (`28 B5 2F FD`) and extract with `tar --zstd -xf`, alongside the existing xz (`-xJf`) and gzip (`-xzf`) paths. Without this, an accepted `.tar.zst` image fails extraction and the rootfs is never scanned. (The xz/gzip detection and the post-extraction hardening guard are unchanged.)

**Runtime image** — `docker/Dockerfile.backend`
- Add `zstd` to the `rootfs-builder` `dnf install` set: `tar --zstd` shells out to a `zstd` binary the minimal `ubi-micro` rootfs otherwise lacks, so without it the extraction is a no-op in the reference image.

## Tests

- `formats::incus::tests`: `test_parse_unified_tarball_zstd`, `test_parse_unified_tarball_gzip`, `test_parse_metadata_tarball_zstd`, `test_lxd_compat_tarball_zstd`.
- `api::handlers::incus::tests`: `test_content_type_for_artifact_tar_zst`, `test_content_type_for_download_tar_zst`, `test_simplestreams_ftype_tar_zst`.
- Scanner extraction reuses the existing `extract_tarball` magic-detection test pattern (mirrors `test_extract_tarball_detects_xz_magic`); a dedicated zstd-magic extraction test is a reasonable follow-up.

## Context

zstd is a natural output for any pipeline that runs `incus export --compression=zstd`: it compresses and decompresses faster than xz at comparable ratios, and is the format used for multi-GiB OS images published into AK. Re-encoding to `.tar.xz` purely to satisfy AK's validator is strictly worse than supporting the format AK's own dependency tree already bundles.

Rebased onto current `main`. PUT of a `.tar.zst` returns HTTP 201 and the artifact is browsable in the UI; the scanner now extracts the same artifacts that the validator accepts.

Closes #1454

### DIFF
--- a/backend/src/api/handlers/incus.rs
+++ b/backend/src/api/handlers/incus.rs
@@ -298,6 +298,8 @@ pub(crate) fn content_type_for_artifact(artifact_path: &str) -> &'static str {
         "application/x-xz"
     } else if artifact_path.ends_with(".tar.gz") {
         "application/gzip"
+    } else if artifact_path.ends_with(".tar.zst") {
+        "application/zstd"
     } else {
         "application/octet-stream"
     }
@@ -309,6 +311,8 @@ pub(crate) fn content_type_for_download(filename: &str) -> &'static str {
         "application/x-xz"
     } else if filename.ends_with(".tar.gz") {
         "application/gzip"
+    } else if filename.ends_with(".tar.zst") {
+        "application/zstd"
     } else if filename.ends_with(".json") {
         "application/json"
     } else {
@@ -326,6 +330,8 @@ pub(crate) fn simplestreams_ftype(filename: &str) -> &str {
         "incus.tar.xz"
     } else if filename.ends_with(".tar.gz") {
         "incus.tar.gz"
+    } else if filename.ends_with(".tar.zst") {
+        "incus.tar.zst"
     } else {
         filename
     }
@@ -1869,6 +1875,14 @@ mod tests {
         );
     }
 
+    #[test]
+    fn test_content_type_for_artifact_tar_zst() {
+        assert_eq!(
+            content_type_for_artifact("ubuntu/20240215/incus.tar.zst"),
+            "application/zstd"
+        );
+    }
+
     // -----------------------------------------------------------------------
     // content_type_for_download
     // -----------------------------------------------------------------------
@@ -1948,6 +1962,19 @@ mod tests {
     #[test]
     fn test_simplestreams_ftype_tar_gz() {
         assert_eq!(simplestreams_ftype("incus.tar.gz"), "incus.tar.gz");
+    }
+
+    #[test]
+    fn test_simplestreams_ftype_tar_zst() {
+        assert_eq!(simplestreams_ftype("incus.tar.zst"), "incus.tar.zst");
+    }
+
+    #[test]
+    fn test_content_type_for_download_tar_zst() {
+        assert_eq!(
+            content_type_for_download("incus.tar.zst"),
+            "application/zstd"
+        );
     }
 
     #[test]
--- a/backend/src/formats/incus.rs
+++ b/backend/src/formats/incus.rs
@@ -326,20 +326,34 @@ pub struct IncusImageMetadata {
 mod tests {
     use super::*;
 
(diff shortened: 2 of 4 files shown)


brandonrc:

dragonpaw:
Heads-up for when this is picked back up for v1.3.0: full `.tar.zst` support also needs `backend/src/services/incus_scanner.rs` touched — not just the accept/metadata/Dockerfile changes in this PR.

`IncusScanner::extract_tarball` only branches on XZ magic (`-xJf`) with a gzip fallback (`-xzf`); a zstd archive falls into the gzip branch, so `tar -xzf` runs against zstd bytes and extraction fails — the rootfs is never scanned. With this PR alone, AK will *accept and store* a `.tar.zst` incus image but the scanner can't unpack it.

Fix is the same shape as the magic-byte detection this PR adds to `extract_metadata_from_reader`: detect the zstd magic (`28 B5 2F FD`) and extract with `tar --zstd -xf` (the `zstd` binary this PR installs in the runtime image is what makes that work).

Found while validating end-to-end scanning of a multi-GiB `.tar.zst` OS image.

Issue #1160, opened 2026-05-11
feat(grype): proper OCI image scanning via registry mode (follow-up to #966)


brandonrc:
Follow-up to #966.

## Context

#966's minimal fix gates Grype out of OCI-manifest scans (Grype currently runs `dir:<workspace>` and the workspace holds the manifest JSON, not the layers — so Grype reports 0 findings on container images). The minimal fix is correct but leaves Grype useless for image scans; only Trivy (via ImageScanner) is doing the work.

## Goal

Make GrypeScanner actually scan OCI container images, so we have two independent scanners producing comparable findings (defense-in-depth + cross-validation; if Trivy and Grype disagree, that's a signal worth surfacing).

## Approach

Grype supports several input modes:

- `grype dir:<path>` (current) — directory scan, no good for OCI manifests
- `grype docker:<image-ref>` — pulls from a Docker daemon, not applicable in our pod
- `grype registry:<image-ref>` — pulls directly from a remote registry over HTTPS
- `grype docker-archive:<path>` — scans a saved docker tarball (`docker save` output)
- `grype oci-archive:<path>` — scans an OCI bundle

The cleanest path is `registry:<image-ref>` pointing at artifact-keeper's own OCI registry endpoint, mirroring how ImageScanner (Trivy server mode) operates. This requires:

1. In GrypeScanner, detect OCI manifest content from the artifact (same predicate ImageScanner uses).
2. Reconstruct the image ref from artifact path (`v2/<name>/manifests/<ref>` → `<our-host>/<name>:<ref>`).
3. Invoke `grype registry:<our-host>/<name>:<ref>` with appropriate auth.
4. The backend service address: derive from env (`AK_PUBLIC_URL` or similar) or use cluster-internal DNS (`http://artifact-keeper-backend.<ns>.svc.cluster.local:8080/v2/`).
5. If auth is required (private repo), pass via `--registry-auth` / `~/.docker/config.json`.

## Test plan

- E2E: upload `library/nginx:latest` manifest + blobs, trigger Grype scan, assert findings_count > 0 (matches Trivy's ~200).
- E2E: private repo image with token-scoped pull, assert Grype gets the layers via auth, finds CVEs.

## Out of scope

- Switching the existing image-scan path away from Trivy; both should run.
- Grype's SBOM-emission mode (`grype --output cyclonedx-json`) — separate work.

Blocked by #966 (the minimal gate fix). Unblocks proper defense-in-depth scanning for OCI images.

PR #1150, opened 2026-05-11
feat(sbom): scan_packages inventory, full dep tree in SBOM (closes #903)

brandonrc:
## Summary

The SBOM endpoint advertised `content.components` as a Software Bill of Materials but derived it from `scan_findings`, which only carries rows for components with known CVEs. Uploading `express@4.18.2` (30+ direct deps, no active CVEs in Trivy's DB on a given day) produced an empty SBOM. Uploads WITH CVEs produced SBOMs containing only the CVE-bearing subset, not the full dep tree.

This PR rebuilds the SBOM pipeline around a canonical per-scan inventory.

**Key changes:**

- New `scan_packages` table (migration 085) that holds every package the scanner saw, regardless of vulnerability status.
- Both Trivy invocation sites (`trivy_fs_scanner`, `image_scanner`) now pass `--list-all-pkgs`, which adds the `Packages` block to Trivy's JSON.
- `Scanner::scan()` returns `Result<ScanOutput>` (with `findings` AND `packages`) instead of `Result<Vec<RawFinding>>`. All 7 scanner impls migrated.
- `ScanResultService::create_packages` persists with `ON CONFLICT DO NOTHING` on `(scan_result_id, name, COALESCE(version, ''))`.
- `extract_dependencies_for_artifact` reads `scan_packages` first, falls back to `scan_findings` only for legacy artifacts.
- `RawFinding.affected_component` is now the bare package name (no `(target)` suffix); the suffix moves to `RawPackage.source_target`.

**Out of scope** (separate issues): license extraction for non-Trivy scanners, SPDX support, backfill for legacy artifacts.

## Test Checklist
- [x] Unit tests added/updated — 5 new tests in `scanner_service` plus the `incus_scanner` finding-name test updated to assert the new bare-name contract.
- [ ] Integration tests added/updated (if applicable) — follow-up will exercise the express@4.18.2 fixture end-to-end on a live Trivy.
- [ ] E2E tests added/updated (if applicable) — depends on artifact-keeper-test SBOM-correctness epic.
- [x] Manually tested locally — migration applied, sqlx cache regenerated, 9041 lib tests pass.
- [x] No regressions in existing tests — full lib green; one pre-existing test updated to match the new contract.

## API Changes
- [ ] New endpoints have `#[utoipa::path]` annotations — N/A, no new endpoints.
- [ ] Request/response types have `#[derive(ToSchema)]` — N/A, response shape (`DependencyInfo`) unchanged.
[...]

PR #1312, opened 2026-05-21
fix(scanner): image scanner emits bare package name (closes #1311)

### LINKED ISSUE #1311: Grype scanner emits old `name (artifact_type)` format on `affected_component` (broader than the Trivy fix in #903)
## Note on initial filing

This issue was originally filed citing `backend/src/api/services/image_scanner.rs:259` as the bug location. That citation was based on a stale local workspace. On actual `main`, both `image_scanner.rs` and the shared `scanner_service::convert_trivy_findings` helper were already fixed by PR #1188 (merged 2026-05-11). The real remaining offender is the Grype code path.

## Actual bug location

`backend/src/services/grype_scanner.rs:254` (Grype scanner used for OCI registry mode per #1160):

```rust
let affected_component = Some(match &m.artifact.artifact_type {
    Some(t) => format!("{} ({})", m.artifact.name, t),
    None => m.artifact.name.clone(),
});
```

The parenthetical here is `artifact_type` (e.g. `java-archive`, `python`), not `result.target` like the Trivy bug. Same spirit though: cross-source joins on `scan_findings.affected_component` against `scan_packages` cannot reconcile entries from Grype scans because the value carries trailing format-metadata.

## Symptom

Vulnerability scans that go through the Grype path (OCI registry image scans, per #1160) write `scan_findings.affected_component` like `"log4j-core (java-archive)"` instead of `"log4j-core"`. SBOM CycloneDX/SPDX output and the new `scan_packages` join-table cannot dedupe by package across Trivy + Grype results.

## Fix

Strip the parenthetical:

```rust
let affected_component = Some(m.artifact.name.clone());
```

If we want to preserve `artifact_type` somewhere queryable, it should go on a separate field (analogous to how the Trivy fix put `target` on `scan_packages.source_target`). Grype's JSON does not emit a `Packages` block today, so persisting an equivalent `source_target` for Grype requires a separate inventory pass; out of scope for the immediate fix.

## Status

Being addressed by PR #1312.

## Why this blocks PR #1223

PR #1223 adds a CHANGELOG entry claiming `affected_component` "is now the bare package name." True for Trivy paths on `main` today. Becomes true for Grype path once #1312 lands. Both should land together for v1.2.0 so the changelog is accurate when shipped.

## Severity

High. Behavioral contract claimed in CHANGELOG (#1223) does not match Grype-path code; downstream integrations relying on the new format will see inconsistent data depending on which scanner produced the finding.

brandonrc:
## Summary

Closes #1311.

The Grype scanner's image-scanner code path (registry mode, #1160) wrapped `affected_component` as `"<name> (<artifact_type>)"`, e.g. `"log4j-core (java-archive)"`. PR #1150 standardized `scan_findings.affected_component` to the bare package name across the filesystem-scanner path so SBOM CycloneDX/SPDX output and the `scan_packages` join-table reconcile entries by raw name, but the image-scanner path was not updated alongside it. This PR aligns Grype's `convert_findings` with `scanner_service::convert_trivy_findings` so cross-source joins work for OCI image scans.

Note on scope: investigation showed the Trivy image-scanner path already delegates to the shared `convert_trivy_findings` helper, which was fixed in #1150 (commit 8af74fd). The remaining offender is the Grype scanner's own `convert_findings`, which is used for OCI registry-mode scans. Grype's JSON does not emit a package inventory block, so the type metadata is dropped rather than persisted in a `RawPackage.source_target` row; the Trivy filesystem scanner that runs alongside provides the inventory side of the contract.

Added `test_convert_findings_emits_bare_package_name_for_typed_artifact` which pins the bare-name format on a Grype finding that carries a non-empty `artifact_type`, the exact shape that produced the bug. The pre-existing `test_convert_findings_basic` was also tightened from `contains("vulnerable-pkg")` (which would have passed against the buggy `"vulnerable-pkg (python)"` output) to `assert_eq!` against the bare name.

## Test Checklist
- [x] Manually tested locally (`cargo test --workspace --lib grype_scanner` and broader `scanner` suite, 456 passed)

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/services/grype_scanner.rs
+++ b/backend/src/services/grype_scanner.rs
@@ -245,23 +245,31 @@ impl GrypeScanner {
     }
 
     /// Convert Grype matches into `RawFinding` values.
+    ///
+    /// `affected_component` holds the bare package name. Earlier versions of
+    /// this method appended the artifact type in parentheses (e.g. `log4j-core
+    /// (java-archive)`), but #1311 aligned the image-scanner code path with
+    /// the filesystem-scanner contract from #903: cross-source join keys
+    /// (SBOM, CVE-mapping, UI) need the raw name, and any type information
+    /// belongs in a separate column rather than smuggled inside the name
+    /// string. Grype's JSON does not emit a package inventory block so the
+    /// type/target metadata is dropped here rather than persisted alongside
+    /// in a `RawPackage` row; the Trivy filesystem scanner that typically
+    /// runs alongside provides the inventory side of the contract.
     fn convert_findings(report: &GrypeReport) -> Vec<RawFinding> {
         report
             .matches
             .iter()
             .map(|m| {
-                let affected_component = Some(match &m.artifact.artifact_type {
-                    Some(t) => format!("{} ({})", m.artifact.name, t),
-                    None => m.artifact.name.clone(),
-                });
-
                 RawFinding {
                     severity: Severity::from_str_loose(&m.vulnerability.severity)
                         .unwrap_or(Severity::Info),
                     title: format!("{} in {}", m.vulnerability.id, m.artifact.name),
                     description: m.vulnerability.description.clone(),
                     cve_id: Some(m.vulnerability.id.clone()),
-                    affected_component,
+                    // Bare package name; matches scanner_service::convert_trivy_findings
+                    // so SBOM / CVE-mapping consumers can join on the raw name.
+                    affected_component: Some(m.artifact.name.clone()),
                     affected_version: Some(m.artifact.version.clone()),
                     fixed_version: m
                         .vulnerability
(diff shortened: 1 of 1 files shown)


brandonrc:
## 3-agent review verdict

**Consensus: APPROVE WITH NITS** (Security Engineer: APPROVE WITH NITS; Code Reviewer + Backend Architect raised MAJOR concerns based on stale local data, see note below)

### Important verification note

The Code Reviewer and Backend Architect both flagged a MAJOR finding that the bug also exists at `image_scanner.rs:259` and `scanner_service.rs:295`. Independent re-verification against a fresh `--depth 1 --branch main` clone shows that finding is **incorrect** — both reviewers were working from stale local workspace copies.

On actual main today:

- `backend/src/services/image_scanner.rs` does NOT contain `format!("{} ({})", ...)` for `affected_component` anywhere. `pkg_name` appears only in the struct definition (line 47). The `format!` calls in the file (lines 195, 223, 232, 239, 246, etc.) are all for HTTP error messages, not for finding-component naming.
- `backend/src/services/scanner_service.rs:493` already correctly emits `affected_component: Some(vuln.pkg_name.clone())` — the bare-name format. This was fixed by PR #1150 / commit 8af74fd, exactly as the PR body claims.
- `grype_scanner.rs:254` was the only remaining offender, and this PR fixes it.

The PR body's scope is accurate. The fix is complete. Both reviewer mistakes mirror my own earlier verification error when filing #1311 (initial workspace copy was stale).

### Security Engineer's findings (the substantive ones)

Security Engineer correctly verified against main and validated the fix. Real findings to track:

- **Low: information loss is real and undocumented**: Grype's `artifact_type` (`java-archive`, `python`, `apk`, `deb`) is dropped at the row level. Unlike Trivy (where `source_target` moves to `scan_packages.source_target`), Grype has no inventory block, so Grype-scanned-only artifacts permanently lose ecosystem context. Triage queries like "show me all Java vulns" against a Grype-only artifact (image with no Trivy fs run alongside) will not be answerable. Recommend follow-up issue to shim a `RawPackage` from Grype matches with `source_target=artifact_type`, OR extend `RawFinding` with an optional `ecosystem` column.

- **Low: backward compatibility not addressed**: existing `scan_findings` rows written before this fix still contain `"log4j-core (java-archive)"`. Cross-source dedupe and SBOM purl construction will emit garbage purls (`pkg:maven/log4j-core (java-archive)@2.14.1`) for those legacy rows until rescanned. Recommend either a one-shot `UPDATE` migration stripping `" (...)"` from `scan_findings.affected_component WHERE source='grype'`, or document "rescan to reconcile" in CHANGELOG.

- **Info: test coverage gap**: no fixture covers names with embedded `(`, spaces, or non-ASCII (scoped npm `@scope/pkg`, OCI registry refs). The strip is unconditional so behavior is fine, but a single fixture would pin the contract.

### Why this matters

The pre-fix Grype output `"log4j-core (java-archive)"` produced invalid purls (`pkg:maven/log4j-core (java-archive)@2.14.1`) in CycloneDX/SPDX SBOMs and broke cross-source dedupe via `SELECT DISTINCT f.affected_component, f.affected_version, f.source` (scanner_service.rs:2701). The fix is real and overdue.

### Plan

Mergeable as-is. Filing follow-up issues for: (1) Grype ecosystem persistence, (2) one-shot backfill migration for legacy rows, (3) test fixture covering edge-case names. The CHANGELOG entry in PR #1223 becomes fully accurate once this merges.

Expediting for v1.2.0.

brandonrc:
Rebased onto current main (064ebec) to fix the Code Coverage gate.

The branch was forked from main at 38cda69 (older state), so `git diff main...HEAD` was showing 125 lines of unrelated diff from intervening PRs (#1267 pypi virtual union, #1268 migration column rename). Those areas of main don't have lib-level tests, so the per-PR new-code coverage gate measured 41% (31/75 lines hit) even though every line genuinely changed by this PR was covered.

After rebase, the diff against main contains only `backend/src/services/grype_scanner.rs` (+69/-16, matching the GitHub PR diff). Local `cargo llvm-cov --lib` against the new line ranges reports 100% (31/31 instrumented new lines hit) via the existing `test_convert_findings_basic` plus the new `test_convert_findings_emits_bare_package_name_for_typed_artifact` regression test.

No production-code changes pushed, only the rebase. Expediting for v1.2.0.

PR #1318, opened 2026-05-22
fix(lxc): register routes mirroring incus (closes #1272)

### LINKED ISSUE #1272: lxc-format repositories return 404 on all routes — /lxc/* never registered in router
## Description

Creating a repository with `format: lxc` succeeds, but every request under `/lxc/<repo-key>/...` returns 404. The closely-related `incus` format works (`GET /incus/<key>/streams/v1/index.json` returns 200). The backend has full plumbing for `lxc` everywhere *except* the Axum route registration.

## Higher-level question first

Before fixing the routing bug — **should `lxc` even exist as a separate format?**

Looking at the code, `lxc` and `incus` already share *everything* that matters:

- Same handler (`IncusHandler` for both).
- Same scanner (`backend/src/services/incus_scanner.rs`).
- Same on-disk storage key prefix (`incus/<repo-id>/...`).
- Same SimpleStreams protocol wire format.
- Same DB column (`format`), just two enum labels.

Incus is the maintained continuation of LXD, and the LXC/LXD/Incus image tooling has stayed payload-compatible across the rename. A single `format: incus` repo serving the SimpleStreams protocol already satisfies `incus`, `lxd`, and stock `lxc` image clients — so the second format buys users no functionality.

If the answer is "no, `lxc` shouldn't exist as a distinct format", then the right fix is option **(3)** below (collapse into `incus`) and the routing bug evaporates.

If the answer is "yes, they should remain separate", then options **(1)** or **(2)** apply.

## Environment Information

- Deployment type: Kubernetes (GKE) and docker-compose (both reproduce)
- Database: PostgreSQL 15
- Backend version: 1.1.9 (also `dev`)
- Web UI version: 1.1.3 (the floating `dev` tag)

## Expected Behavior

`GET /lxc/<key>/streams/v1/index.json` returns a SimpleStreams index (matching the `/incus/` behavior), **or** `lxc` is removed from the format enum / UI dropdown until handlers exist.

## Observed Behavior

`GET /lxc/<key>/streams/v1/index.json` → 404. No `/lxc/*` routes respond.

## Steps to Reproduce

1. `POST /api/v1/repositories` with `{"key": "test-lxc", "format": "lxc", "repo_type": "local"}` — succeeds (handler accepts `"lxc"` at `backend/src/api/handlers/repositories.rs:649`).
2. `curl http://backend/lxc/test-lxc/streams/v1/index.json` → 404.
3. Same call against an `incus`-format repo at `/incus/<key>/streams/v1/index.json` → 200.

## Additional Context

Backend code already aliases `lxc` to the Incus handler at every other layer:

- Enum variant `RepositoryFormat::Lxc` is defined (`backend/src/models/repository.rs:79`).
- Migration `059_incus_format.sql` adds both `incus` and `lxc` to the Postgres enum.
- `formats::get_core_handler` matches `"incus" | "lxc"` → `IncusHandler` (`backend/src/formats/mod.rs:189`).
- `formats::get_handler_for_format` matches `Incus | Lxc` → `IncusHandler` (`backend/src/formats/mod.rs:247`).
- The Incus handler's repo resolver explicitly accepts both formats (`backend/src/api/handlers/incus.rs:388`).

The structurally minimal change is one line: add `.nest("/lxc", handlers::incus::router())` next to the existing `.nest("/incus", ...)` at `backend/src/api/routes.rs:86`. But a naive route alias has a wrinkle: the Incus handler hardcodes `/incus/...` in returned URLs (`build_download_url` at `backend/src/api/handlers/incus.rs:225-235`, `build_upload_location` at line 277-280). A request to `/lxc/<key>/streams/v1/index.json` would return a SimpleStreams index whose image download paths point at `/incus/<key>/...` — wrong for clients that came in via the `/lxc/` prefix and do prefix-based auth or relative resolution.

### Possible fix shapes

**1. Pure route alias** (one line). Add `.nest("/lxc", handlers::incus::router())`. Cheapest, but SimpleStreams index served via `/lxc/` will reference `/incus/` download URLs. May be acceptable if clients follow absolute URLs blindly.

**2. Route alias + URL prefix awareness.** Add the alias and refactor `build_download_url` / `build_upload_location` to take a prefix (or look up the repo's `format` and pick `/incus` vs `/lxc`). Slightly larger patch; behavior identical to two independent handlers.

**3. Collapse `Lxc` into `Incus`.** Treat `lxc` as a deprecated alias for `incus` at the API boundary, remove from the format enum/UI dropdown, write a migration to relabel any existing `format: 'lxc'` rows. Largest blast radius but matches the reality that the two formats share every implementation detail. (This is the operator-side intuition: one repo serves all versions of incus/lxc clients; two formats is just bookkeeping.)

Happy to send a PR once the preferred direction is confirmed. Personal lean toward **(3)** for the reasons in the "Higher-level question" section above, but absolutely a maintainer call.

## Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate (closest: #203 "Request: Support for Incus/lxc containers" — closed/shipped)
- [x] I am using a supported version (1.1.9)

brandonrc:
## Summary

Closes #1272. Adds `/lxc/<repo-key>/...` routes by mirroring the existing `/incus/<repo-key>/...` registration. The rest of the stack already aliased `lxc` to the Incus handler at every layer (`RepositoryFormat::Lxc` enum variant, `formats::get_handler_for_format` dispatch, `resolve_incus_repo` accepts both `incus` and `lxc`, on-disk `incus/<repo-id>/...` storage key prefix), but `create_router` only mounted `handlers::incus::router()` under `/incus`. Repositories created with `format: lxc` therefore 404'd on every request while identical requests against `incus`-format repos at `/incus/<key>/...` returned 200.

One-line fix: nest the same router under `/lxc` as well. The `IncusHandler` is already format-aware via its repo resolver.

Caveat documented in CHANGELOG and the commit body: the SimpleStreams index served via `/lxc/` still emits absolute `/incus/...` download URLs because `build_download_url` and `build_upload_location` hardcode the prefix. Clients that follow absolute URLs blindly are unaffected; clients doing prefix-based auth or relative resolution against `/lxc/` get redirected to `/incus/`. Making the URL builders prefix-aware (or collapsing the `Lxc` enum variant into `Incus` outright since they share every implementation detail) is a structural decision worth its own discussion and is tracked as a follow-up.

Two source-level meta-tests in `backend/src/api/routes.rs` pin the invariant:

- `lxc_route_is_registered_alongside_incus` asserts both `.nest("/incus", handlers::incus::router())` and `.nest("/lxc", handlers::incus::router())` exist in the router source. Removing the `/lxc` line (reproducing the bug) fails the test loudly. A guard on the `/incus` line catches refactors that would make the `/lxc` assertion vacuously true.
- `lxc_and_incus_share_the_same_handler` pins the count of `handlers::incus::router()` references at >= 2 so a future refactor that introduces a wrapper has to explicitly update this test.

Runtime route-level tests would require full `AppState` + a Postgres fixture (the existing `backend/tests/incus_upload_tests.rs` integration suite is gated on `DATABASE_URL` and `--ignored`) which is heavier than the bug warrants; the source-level meta-tests catch the exact regression class while staying inside `cargo test --workspace --lib`.

## Test Checklist
- [x] Manually tested locally (`cargo test --workspace --lib lxc` -- 2/2 pass; `cargo test --workspace --lib` -- 9655/9655 pass; `cargo fmt --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean)

## API Changes
- [x] N/A - no API changes (the Incus handler has no utoipa annotations today, and this PR only adds a route alias to the same handler; OpenAPI surface is unchanged)


brandonrc:
## 3-agent review verdict

**Consensus: APPROVE** (Code Reviewer: APPROVE WITH NITS; Backend Architect: APPROVE; audit-code-quality: APPROVE WITH NITS)

Clean, surgical 1-line route alias plus 2 source-pin meta-tests. All required CI checks green. Route wiring verified sound across all three reviewers — `Router::nest("/lxc", handlers::incus::router())` picks up every leaf route, `resolve_incus_repo` already accepts both formats, `storage_path_for_key` is prefix-agnostic.

### Convergent finding (Low, file as follow-up)

All three reviewers flagged the same URL-builder cross-prefix issue: `build_download_url` and `build_upload_location` at `incus.rs:225,278` hardcode `/incus/` in returned URLs. A SimpleStreams index requested via `/lxc/<key>/streams/v1/index.json` returns image paths pointing at `/incus/<key>/...`. Most lxc/incus clients follow absolute URLs blindly so this is functional, but strict prefix-relative clients see cross-prefix. The PR body already acknowledges this. Fix is mechanical (thread the mount prefix through the URL builders) but out of scope here.

### Backend Architect's bigger recommendation

The principled long-term cleanup is to **collapse `RepositoryFormat::Lxc` into `Incus`** entirely. Every layer treats them identically: same handler, same scanner, same storage prefix (`incus/<repo-id>/...`), same `artifact_metadata.format = 'incus'`, same SimpleStreams wire protocol. The `Lxc` variant is pure user-facing labeling. Removing it would obviate both the route-alias and URL-builder follow-ups. Worth a separate issue.

### Code-quality nits (Low)

- **Meta-tests are source-text pins**, not behavioral. They `include_str!` `routes.rs` into itself and search for literal strings. A future refactor renaming the call-site string silently breaks the pin without protecting against regression. Module doc should call this out so future readers don't mistake passing tests for behavioral coverage.
- **Docs site guide drift**: `artifact-keeper-site/src/content/docs/docs/guides/incus.mdx` still says "All Incus operations use the `/incus` endpoint" — needs a note that `format: lxc` repos respond on `/lxc/{repo_key}/...`. Site-only change.

### Praise

- Author commentary in the doc comment is unusually honest about test tradeoffs.
- CHANGELOG entry well-scoped and matches sibling entries.
- Storage prefix design (intentional `incus/...` regardless of route prefix) is correctly documented rather than papered over.

Mergeable. Filing 2-3 follow-up issues (URL-builder prefix-awareness, Lxc collapse, docs site update).

PR #1323, opened 2026-05-22
fix(maven): include groupId in virtual local-member match (closes #1287)

### LINKED ISSUE #1287: Maven virtual: resolve_virtual_download matches local member by artifact name ignoring groupId
## Summary

When a **virtual** Maven repository has a **local** member containing an artifact whose `artifactId` (name) collides with an artifact in a **remote** member but under a different `groupId`, the virtual repository returns **404** for the remote artifact instead of resolving it from the remote member.

The root cause appears to be that `resolve_virtual_download` matches local member artifacts by the `name` column (= Maven `artifactId`) without considering the full Maven path (groupId + artifactId + version).

### Setup

1. Create a **local** Maven repository: `maven-local`
2. Create a **remote** Maven repository: `maven-remote` proxying Google Maven (`https://dl.google.com/dl/android/maven2`)
3. Create a **virtual** Maven repository: `maven-virtual` with members: `maven-local` (priority 1), `maven-remote` (priority 2)

### Verify baseline

```bash
# This should return 200 (resolved from remote -> Google Maven)
curl -I https://<ak-host>/maven/maven-virtual/com/android/tools/common/31.4.0/common-31.4.0.pom
# -> HTTP 200
```

### Upload a local artifact with the same artifactId but different groupId

```bash
# Upload an artifact with artifactId="common" but groupId="com.example.mylib"
curl -X PUT \
  "https://<ak-host>/maven/maven-local/com/example/mylib/common/1.0.0/common-1.0.0.pom" \
  -H "Authorization: Bearer <token>" \
  -H "Content-Type: application/xml" \
  --data-binary @common-1.0.0.pom
```

### Try to resolve the remote artifact again

```bash
curl -I https://<ak-host>/maven/maven-virtual/com/android/tools/common/31.4.0/common-31.4.0.pom
# -> HTTP 404 !!!
```

### Delete the local artifact -> remote resolution works again

```bash
curl -X DELETE \
  "https://<ak-host>/api/v1/repositories/maven-local/artifacts/com/example/mylib/common/1.0.0/common-1.0.0.pom" \
  -H "Authorization: Bearer <token>"

curl -I https://<ak-host>/maven/maven-virtual/com/android/tools/common/31.4.0/common-31.4.0.pom
# -> HTTP 200 (works again!)
```

## Expected Behavior

Virtual resolution should match by **full path** (groupId + artifactId + version), not just by `name` (artifactId). The presence of `com/example/mylib/common/1.0.0/` in a local member should NOT affect resolution of `com/android/tools/common/31.4.0/` from a remote member.

## Observed Behavior

- When `maven-local` contains ANY artifact with `name = "common"` (regardless of groupId), virtual resolution returns 404 for `com/android/tools/common/31.4.0/common-31.4.0.pom`
- The virtual repo appears to find a local match by `name` column, then fails to serve the file (because the path does not actually exist in the local member at that path), and does NOT fall through to remote members

## Additional Context

- **AK version**: 1.2.0-rc.1
- **Upload method matters**: Uploading via native Maven PUT (`PUT /maven/{repo}/{path}`) stores `name = "common"` (correct artifactId) and triggers the bug. Uploading via REST API (`POST /api/v1/repositories/{repo}/artifacts/{path}`) stores `name = "com"` (first path segment - incorrect GAV parsing) and does NOT trigger the bug because the name no longer collides.
- The REST API GAV parsing issue is a separate bug but explains why some users may not encounter the virtual resolution issue.

## Workaround

Avoid having artifacts in local members whose `artifactId` matches any `artifactId` in remote member repositories. Alternatively, use the REST API upload endpoint (which mis-parses GAV and accidentally avoids the collision).

brandonrc:
## Summary

Closes #1287.

Maven virtual repos resolving a download were matching local members by artifactId only, ignoring groupId. The download handler called the cross-format shadowing guard `virtual_non_remote_owns_name`, which compares against `artifacts.name`. For Maven that column holds just the artifactId, so two coordinates that share an artifactId across different groupIds (e.g. `com.example.mylib:common:1.0.0` vs. `com.android.tools:common:31.4.0`) collided. Once the local artifact was uploaded, the guard fired for every request that ended in `.../common/...`, causing `proxy_for_virtual = None`, which made `resolve_virtual_download` skip all Remote members. The local path lookup then missed (the requested path lives under a different groupId) and the whole resolution returned 404.

The fix adds a Maven-aware variant `virtual_non_remote_owns_maven_ga` that matches on the full `groupId/artifactId/` path prefix using the existing `(repository_id, path)` btree (`idx_artifacts_repo_path`). Both inputs are LIKE-escaped so a crafted artifactId or groupId cannot widen the match beyond its GA directory. The Maven download handler swaps to this guard. The generic primitive is unchanged because non-Maven formats do not encode a hierarchical GA in the path.

## Test Checklist
- [x] Manually tested locally

Tests:
- 5 pure unit tests on `maven_ga_like_pattern` (groupId/artifactId composition, distinct groupIds produce distinct prefixes, sibling groupIds do not match, `%`/`_`/`\` are LIKE-escaped).
- 3 DB-driven integration tests in `backend/tests/maven_virtual_groupid_shadowing_test.rs` covering: the #1287 reproducer (a local `com.example.mylib:common` artifact must not shadow the remote `com.android.tools:common` GA), the no-non-Remote-members case (guard returns false), and crafted-artifactId LIKE-metachar escaping.

Verified locally:
- `cargo fmt --check` clean.
- `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `cargo test --workspace --lib maven` (180 maven-related tests) passes.
- `cargo test --test maven_virtual_groupid_shadowing_test -- --ignored` against `postgresql://registry:registry@localhost:30432/artifact_registry` passes 3/3.

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/maven.rs
+++ b/backend/src/api/handlers/maven.rs
@@ -1148,24 +1148,30 @@ async fn serve_artifact(
                 let artifact_path = path.to_string();
 
                 // Supply-chain shadowing guard (#1217 follow-up, ak-hv3s).
-                // Maven `artifacts.name` stores `coords.artifact_id` (the
-                // artifactId component of GAV; see the publish-path
-                // around line 1634). To run the guard we have to parse
-                // the GAV out of the request path. The guard fires
-                // whenever a non-Remote member already owns the
-                // artifactId case-insensitively; this is strictly a
-                // safety net rather than an authority check, because
-                // different groupIds may legitimately share an
-                // artifactId. The trade-off is preferable to leaving
-                // the shadowing attack open. If the path fails to parse
-                // as a Maven coordinate (eg. dynamic metadata.xml
-                // requests reach this branch from earlier fall-through),
-                // skip the guard rather than block the request.
+                // Originally this used the generic `name`-only guard
+                // (`virtual_non_remote_owns_name`) keyed off
+                // `coords.artifact_id`. That over-matched across
+                // groupIds: a local `com.example.mylib:common:1.0` shadowed
+                // every remote `com/.../common/...` lookup, returning
+                // 404 instead of falling through to the remote member
+                // (#1287). The Maven-aware variant matches the full
+                // groupId+artifactId path prefix so only true GA
+                // collisions activate the suppression. The guard
+                // remains a safety net rather than an authority check:
+                // different versions under the same GA still
+                // legitimately share a directory, and we accept the
+                // false-positive within a single GA in exchange for
+                // closing the shadowing attack. If the path fails to
+                // parse as a Maven coordinate (eg. dynamic
+                // metadata.xml requests reach this branch from earlier
+                // fall-through), skip the guard rather than block the
+                // request.
                 let local_owns = match MavenHandler::parse_coordinates(path) {
                     Ok(coords) => {
-                        proxy_helpers::virtual_non_remote_owns_name(
+                        proxy_helpers::virtual_non_remote_owns_maven_ga(
                             &state.db,
                             repo.id,
+                            &coords.group_id,
                             &coords.artifact_id,
                         )
                         .await?
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -1305,6 +1305,91 @@ pub async fn virtual_non_remote_owns_name(
     Ok(exists.is_some())
 }
 
+/// Build the SQL `LIKE` pattern that matches every artifact path under
+/// a given Maven `groupId/artifactId/` directory.
+///
+/// Pure helper extracted so the prefix construction has unit-test
+/// coverage without a database. Returns `<group-path>/<artifactId>/%`,
+/// where the dot-to-slash conversion runs before LIKE-escaping so
+/// directory separators in the groupId are preserved, and `%`/`_`/`\`
+/// inside either input become literal characters. Use with
+/// `LIKE ... ESCAPE '\'`.
+pub(crate) fn maven_ga_like_pattern(group_id: &str, artifact_id: &str) -> String {
+    let group_path = group_id.replace('.', "/");
+    let mut prefix = String::with_capacity(group_path.len() + artifact_id.len() + 3);
+    prefix.push_str(&super::escape_like_literal(&group_path));
+    prefix.push('/');
+    prefix.push_str(&super::escape_like_literal(artifact_id));
+    prefix.push('/');
+    prefix.push('%');
+    prefix
+}
+
+/// Maven-aware shadowing guard: returns true if any non-Remote member of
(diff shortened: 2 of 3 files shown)


brandonrc:
## 3-agent review verdict

**Consensus: APPROVE** (Code: APPROVE; DB Optimizer: APPROVE WITH NITS; Backend Architect: APPROVE with architectural follow-up)

Correct, surgical fix for #1287. The LIKE-escaping (dot-to-slash conversion before `escape_like_literal`), `ESCAPE '\'` clause, index choice (`idx_artifacts_repo_path`), and fail-closed semantics all hold up. 5 pure tests + 3 DB integration tests covering reproducer + empty-non-Remote case + metachar escaping. 180 maven tests pass.

### Key validation
- Maven `to_path()` stores `group/path/artifactId/version/filename`; helper's pattern correctly mirrors that exact transform.
- `repository_id = ANY($1)` narrows hard before the LIKE; planner uses prefix scan effectively.
- LIKE escaping handles `%`, `_`, `\` in artifactId/groupId; tests pin this.

### Required architectural follow-up

Backend Architect identified that **Conan and OCI v2 have the same hierarchical-identity issue and currently have NO shadowing guard at all**:
- Conan (`name/version/user/channel`): `recipe-name` collisions across `user/channel` would shadow.
- OCI/Docker v2 (`namespace/image_name`): `library/redis` vs `myorg/redis` collisions.

Filing follow-ups:
1. **Refactor to `ShadowingIdentity` enum** (P2): single entry point with `Name(&str)` and `PathPrefix { prefix }` variants. Collapses the format-specific variants.
2. **Add shadowing guard to Conan** with PathPrefix identity (P1, real attack surface).
3. **Add shadowing guard to OCI v2** with PathPrefix identity (P1).

### DB Optimizer nit

PG only uses btree for prefix LIKE if column collation is `C`/`POSIX` OR index built with `text_pattern_ops`. `artifacts.path` is plain TEXT default collation. On most clusters (`en_US.UTF-8`), planner may skip the index. Recommend verifying with `EXPLAIN ANALYZE` on staging; if it falls back to seq scan, add `idx_artifacts_repo_path_pattern ON artifacts(repository_id, path text_pattern_ops)` in a follow-up migration. The `LIMIT 1` keeps cost bounded but indexing properly avoids cliff.

### Code Reviewer nits

- Big call-site comment in maven.rs is excellent; consider trimming history paragraph post-merge.
- Add a test for "same groupId + same artifactId across two local members" to prove the guard hasn't become groupId-scoped-to-one-member by accident.
- Doc comment "Uses the `(repository_id, path)` btree" slightly overstates given collation caveat above.

Mergeable. Filing 3 follow-ups (ShadowingIdentity refactor, Conan guard, OCI guard).

Issue #976, opened 2026-04-30
why not allowed private/internal ip address for remote repository upstream


wppzxc:
### Problem Statement

backend version: 1.1.6

when i create remote repository and use 10.0.0.0 ip address, get error : ' Upstream URL IP '10.0.0.0' is not allowed (private/internal network)'

### Proposed Solution

will artifactkeeper change the rule and allow private/internal ? i really need this feature because my company network has limit for connect public network

### Component

None

### Pre-submission Checklist

- [x] I have searched existing issues and feature requests to ensure this isn't a duplicate
- [x] I have reviewed the documentation to confirm this feature doesn't already exist

PR #1325, opened 2026-05-22
feat(ssrf): add AK_SSRF_ALLOW_PRIVATE_CIDRS env var (closes #1224)

### LINKED ISSUE #1224: feat(ssrf): add AK_SSRF_ALLOW_PRIVATE_CIDRS env var to unblock in-cluster test fixtures
## Summary

`backend/src/api/validation.rs::validate_outbound_url` rejects every RFC1918 address via Rust's built-in `Ipv4Addr::is_private()`. There is no env-var or config knob to allow specific private CIDRs through. This is correct production behavior, but it blocks an entire class of in-cluster integration tests from running against a real backend.

## Why this matters

artifact-keeper-test ships two security suites that need to point the backend at a controllable mock HTTP upstream:

- `tests/security/test-cache-poisoning.sh` (T2-11)
- `tests/security/test-cache-stampede.sh` (T2-12)

The fixture (`tests/lib/mock-upstream.py`, started by `tests/lib/common.sh::start_mock_upstream`) runs the mock inside the ARC runner pod and exposes it on the pod IP. In our K8s cluster the runner pod CIDR is `10.244.0.0/16`, which is RFC1918, so the backend pod rejects:

```
POST /api/v1/repositories
{ "upstream_url": "http://10.244.0.103:45293" }
-> HTTP 400 "IP '10.244.0.103' is not allowed (private/internal network)"
```

Both tests were stubbed in artifact-keeper-test#126 with a `skip` so release-gate is honest, tracked in artifact-keeper-test#122.

## Existing knobs

The only related env var today is `BLOCK_CGNAT_OUTBOUND`, and it goes the other direction (adds `100.64.0.0/10` to the blocklist). There is no symmetric "remove this CIDR from the blocklist" override.

## Proposed change

Add an opt-in env var `AK_SSRF_ALLOW_PRIVATE_CIDRS` (comma-separated CIDR list) that the validator parses once at startup and consults inside `is_blocked_ipv4` / `is_blocked_ipv6` before returning `true`. Off by default. Production deployments do not set it; test deployments set it to the pod CIDR (`10.244.0.0/16`) so the in-cluster mock-upstream fixture can run.

Acceptance criteria:

1. New env var documented in `.env.example` with a warning that it weakens SSRF protection and should not be set in production.
2. `is_blocked_ipv4` and `is_blocked_ipv6` honor the allowlist for the corresponding family (CIDR parsing supports both `10.0.0.0/8` and `fd00::/8` forms).
3. Unit tests in `validation.rs` covering:
   - empty / unset env var preserves existing block behavior
   - valid CIDR allows IPs inside, still blocks outside
   - malformed CIDR list logs a warning and falls back to "deny all private"
   - cloud-metadata IPs (`169.254.169.254`, `192.0.0.192`, `100.100.100.200`) remain blocked even if the user allowlists their containing CIDR (defense-in-depth)
4. helm chart value plumbed through (`backend.env.AK_SSRF_ALLOW_PRIVATE_CIDRS`) so artifact-keeper-iac's test overlay can set it without backend code changes per test.

## Alternatives considered

- **Service-DNS mock**: deploy the mock as a K8s Service and reference it by hostname. Validator passes (it's not in BLOCKED_HOSTS, not an IP literal). Pro: zero backend changes. Con: requires reworking the test fixture to deploy a Service/Deployment instead of bare-process exec on the runner, and the mock state-dir model (filesystem-shared with the test) breaks because the mock pod no longer shares the runner pod's filesystem. The fixture refactor is significantly more code than the allowlist env var on the backend side.
- **NodePort + node IP**: same DNS problem solved at a different layer; same fixture cost.

The env-var allowlist is the smallest defensible change and unblocks artifact-keeper-test#122 with one helm-values edit on the test side.

## References

- artifact-keeper-test#122 (test-side tracking)
- artifact-keeper-test#126 (PR that stubbed the tests)
- `backend/src/api/validation.rs` (the validator)

brandonrc:
## Summary

Adds the `AK_SSRF_ALLOW_PRIVATE_CIDRS` env var as the canonical name for the comma-separated private-CIDR allowlist that gates `validate_outbound_url`. When set (e.g. `10.244.0.0/16`), upstream URLs that land inside one of the listed CIDRs are permitted; everything else (RFC1918 outside the list, loopback, link-local, cloud metadata IPs) remains blocked.

Motivation: in-cluster integration tests (artifact-keeper-test#122) need to point AK at a mock upstream running on the ARC runner pod IP, which lives in the K8s pod CIDR `10.244.0.0/16` (RFC1918). Without an allowlist, the validator rejects the URL before the request can be issued.

Implementation notes:

- `UPSTREAM_PRIVATE_IP_ALLOWLIST` (issue #976) is kept as a backward-compatible alias. If both are set, `AK_SSRF_ALLOW_PRIVATE_CIDRS` wins so the canonical name is the operator's source of truth.
- Default OFF. The existing security posture is preserved: production deployments do not set the var.
- Cloud metadata IPs (169.254.169.254, 192.0.0.192, 100.100.100.200) and loopback / link-local stay hard-blocked even if an operator allowlists a containing CIDR. The hard-block check fires before the allowlist is consulted.
- Malformed CIDR entries are logged and skipped, never silently widening the allowlist.
- `0.0.0.0/0` and `::/0` are rejected at the helper level so a copy-pasted blanket entry cannot widen the allowlist by accident.
- `.env.example` documents the var with an explicit warning that it weakens SSRF protection and should not be set in production.
- Boot-time SSRF posture log now references both the canonical name and the alias, and includes issue #1224 alongside #976.

Closes #1224.

## Test Checklist
- [x] Manually tested locally

Nine new unit tests in `backend/src/api/validation.rs`:

- `test_ssrf_allow_private_cidrs_default_blocks_private`: production default (var unset) still blocks every RFC1918 address.
- `test_ssrf_allow_private_cidrs_unblocks_listed_pod_cidr`: exact issue-1224 scenario, `AK_SSRF_ALLOW_PRIVATE_CIDRS=10.244.0.0/16` allows `10.244.0.103`, neighbours stay blocked.
- `test_ssrf_allow_private_cidrs_still_blocks_metadata`: hard-block for AWS / Oracle / Alibaba metadata IPs survives even when their containing CIDR is allowlisted.
- `test_ssrf_allow_private_cidrs_still_blocks_loopback`: 127.0.0.1 stays blocked even if 127.0.0.0/8 is in the list.
- `test_ssrf_allow_private_cidrs_malformed_entries_skipped`: typos are logged and skipped; good entries still work.
- `test_ssrf_allow_private_cidrs_empty_falls_through`: blank values fall through to the blanket toggle rather than collapsing to "block all private".
- `test_ssrf_allow_private_cidrs_ipv6_subnet`: IPv6 CIDR (`fd00::/8`) works.
- `test_ssrf_allow_private_cidrs_takes_precedence_over_alias`: new canonical name wins over legacy alias when both are set.
- `test_legacy_alias_still_honored_when_new_var_unset`: legacy `UPSTREAM_PRIVATE_IP_ALLOWLIST` still works on its own.
- `test_private_cidr_allowlist_value_helper`: direct contract test for the helper that main.rs uses to format the boot warning.

Pre-existing baseline tests (`test_rejects_10_network`, `test_rejects_172_16_network`, `test_rejects_192_168_network`, `test_rejects_ipv4_mapped_private_10`) now also take the `AllowlistGuard` so they cannot race parallel allowlist-mutating tests.

Verified locally:
- `cargo fmt --check` clean
- `cargo test --workspace --lib api::validation::` -> 75/75 pass

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -501,19 +501,28 @@ JWT_EXPIRATION_SECS=86400
 # -----------------------------------------------------------------------------
 # By default, upstream URLs (remote repository configs, webhooks, migration
 # sources) are rejected when they target private / internal IPs. Operators
-# on air-gapped or fully-internal networks (issue #976) can opt in to
-# private-IP upstreams via one of these two toggles.
+# on air-gapped or fully-internal networks (issue #976), or in-cluster
+# integration tests pointing at a mock upstream on a pod CIDR
+# (issue #1224), can opt in to private-IP upstreams via one of these
+# two toggles.
+#
+# WARNING: these settings weaken SSRF protection. Do NOT set them in
+# production-facing deployments. They exist for tightly scoped internal
+# mirrors and hermetic test fixtures where the operator owns the network.
 #
 # Cloud metadata IPs (169.254.169.254, 192.0.0.192, 100.100.100.200) and
-# loopback / link-local are NEVER unblocked through these toggles — those
+# loopback / link-local are NEVER unblocked through these toggles, those
 # are SSRF targets, not "internal mirrors".
 #
 # Option 1 (blanket, broadest): allow ALL RFC1918 + IPv6 unique-local.
 # UPSTREAM_ALLOW_PRIVATE_IPS=false
 #
 # Option 2 (narrower, preferred): comma-separated CIDRs. Wins over Option 1
-# if both are set. Bare IPs (no /N) act as /32 or /128.
-# UPSTREAM_PRIVATE_IP_ALLOWLIST=10.50.0.0/16,192.168.7.5
+# if both are set. Bare IPs (no /N) act as /32 or /128. In-cluster test
+# fixtures typically set this to the pod CIDR, e.g. `10.244.0.0/16`.
+# `UPSTREAM_PRIVATE_IP_ALLOWLIST` is accepted as a backward-compatible
+# alias; if both are set, `AK_SSRF_ALLOW_PRIVATE_CIDRS` wins.
+# AK_SSRF_ALLOW_PRIVATE_CIDRS=10.50.0.0/16,192.168.7.5
 #
 # Block the entire RFC 6598 CGNAT range (100.64.0.0/10). Off by default
 # because K8s pod CIDRs in some clusters and CGNAT-served residential ISPs
--- a/backend/src/api/validation.rs
+++ b/backend/src/api/validation.rs
@@ -172,18 +172,22 @@ pub(crate) fn is_blocked_url(url: &reqwest::Url) -> Option<BlockReason> {
 ///   evaluated *first* so `::1` is correctly classified as IPv6 loopback
 ///   rather than IPv4 alias `0.0.0.1`.
 ///
-/// Private-IP allowlist (issue #976): operators on corporate networks
-/// with no public internet may need to point upstreams at internal
-/// mirrors. Two opt-in escape hatches:
+/// Private-IP allowlist (issues #976, #1224): operators on corporate
+/// networks with no public internet, or in-cluster test fixtures that
+/// need to reach a mock upstream on a pod CIDR, may need to point
+/// upstreams at internal mirrors. Two opt-in escape hatches:
 ///
 /// - `UPSTREAM_ALLOW_PRIVATE_IPS=true` — allow all RFC1918 + IPv6
 ///   unique-local. Cloud metadata IPs (169.254.169.254, 192.0.0.192,
 ///   100.100.100.200) and loopback / link-local / unspecified remain
 ///   blocked, since those are SSRF targets, not "internal mirrors".
-/// - `UPSTREAM_PRIVATE_IP_ALLOWLIST=10.0.0.0/8,192.168.7.0/24` — more
+/// - `AK_SSRF_ALLOW_PRIVATE_CIDRS=10.0.0.0/8,192.168.7.0/24` — more
 ///   precise: only the listed CIDRs are exempted. Same metadata /
 ///   loopback hard-blocks apply. Wins over the blanket toggle if both
 ///   are set (allowlist is strictly more restrictive).
+///   `UPSTREAM_PRIVATE_IP_ALLOWLIST` is accepted as a backward-compatible
+///   alias; if both are set, `AK_SSRF_ALLOW_PRIVATE_CIDRS` takes
+///   precedence.
 pub(crate) fn is_blocked_ip(ip: std::net::IpAddr) -> bool {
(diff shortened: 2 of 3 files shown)


brandonrc:
## 3-agent review verdict

**Consensus: APPROVE** (Code: APPROVE; Security: APPROVE; DevOps: APPROVE WITH NITS)

Clean canonicalization. Default-OFF preserved. Canonical name wins over legacy alias. Hard-blocks (metadata + loopback + link-local + IPv4-mapped IPv6) STAY BLOCKED unconditionally even when their containing CIDR is allowlisted — verified in `is_blocked_ipv4`/`is_blocked_ipv6` ordering with tests at validation.rs:1097-1148. Malformed entries warn-and-skip. `0.0.0.0/0` rejected. 10 new unit tests cover pod-CIDR happy path + all defense-in-depth scenarios + precedence + IPv6 + legacy alias.

### Security validation (all 5 checks PASS)

- **Hard-blocks preserved**: AWS/Oracle/Alibaba metadata, loopback, link-local stay blocked even when containing CIDR is allowlisted
- **Backward compat**: `UPSTREAM_PRIVATE_IP_ALLOWLIST` still honored; canonical name wins
- **Default OFF**: both env vars unset → blocking behavior preserved
- **`.env.example` warning**: explicit "Do NOT set in production" block + legitimate-use note
- **Boot-time logging**: `tracing::warn!(target: "security", allowlist=%list, ...)` on startup

### DevOps integration

Helm chart's `backend.env` map ranges through into deployment, so the env var flows automatically — no template change needed. To actually unblock artifact-keeper-test#122, a companion PR in `artifact-keeper-iac` needs to add `backend.env.AK_SSRF_ALLOW_PRIVATE_CIDRS: "10.244.0.0/16"` to the e2e values file.

### Non-blocking nits

- **Pre-existing AllowlistGuard bug** (Security): `prev_ssrf` is restored on Drop but `UPSTREAM_ALLOW_PRIVATE_IPS` clearing is not. Pre-existing, not introduced here. Worth a follow-up.
- **Metrics**: consider Prometheus counter for "allowlisted private IP traversed" — runtime visibility beyond boot logs
- **`pub` vs `pub(crate)`**: `private_cidr_allowlist_value()` is `pub` but only used by main.rs. Could tighten to `pub(crate)`.
- **iac companion PR** needed in artifact-keeper-iac for staging values.

Mergeable. Filing follow-ups for iac plumbing + AllowlistGuard cleanup.

Issue #722, opened 2026-04-13
npm tarballs stored with incorrect Content-Type causing empty SBOM and failed security scans


Firjens:
### Description

All npm tarballs proxied and cached through a remote npm repository are stored with `Content-Type: application/octet-stream` instead of the correct `application/gzip`. This causes downstream failures in both SBOM generation and security scanning, as the backend appears to use the content type to determine how to handle the file for extraction.

### Environment Information

- Deployment type: k8s
- Operating System: Alma
- Architecture: AMD64
- Database: Postgres
- Reverse Proxy: nginx
- Network setup: 

### Backend Version

dev

### Web UI Version

dev

### Package Format

npm

### Expected Behavior

- npm tarballs (.tgz files) should be stored with `Content-Type: application/gzip` or `application/x-tar+gzip`
- SBOM generation should extract the tarball, parse `package.json`, and populate components, dependencies, and licenses
- Security scanning (Trivy, Grype) should extract the tarball and scan its contents for vulnerabilities

### Observed Behavior

- All proxied npm tarballs show `Content-Type: application/octet-stream` in the artifact details UI
- SBOM generation completes with 0 components, 0 dependencies, 0 licenses
- Trivy filesystem scan reports 0 vulnerabilities (`Number of language-specific files: 0`)
- Grype scan fails with exit status: 1 and returns empty findings
- Backend logs showed `Failed to extract archive <name>.tgz: Internal error: Failed to execute extraction command: No such file or directory (os error 2). Scanning raw file instead.` (on earlier builds; current dev build suppresses this)

### Steps to Reproduce

- Create a remote npm repository pointing to `https://registry.npmjs.org`
- Configure a client to use the proxy (`.npmrc`: `registry=https://<host>/npm/<repo-name>/`)
- Run `yarn install` or `npm install` on any project to populate the proxy cache
- In the UI, navigate to the repository → Artifacts → click any `.tg`z artifact
- Observe `Content-Type: application/octet-stream` in the Details tab
- Click SBOM tab → Regenerate → observe 0 components, 0 dependencies, 0 licenses
- Click Security tab → Scan → observe 0 findings despite packages having known CVEs

### Logs

```shell

```

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have redacted any sensitive information (passwords, tokens, private URLs)
- [x] I am using a supported version (not end-of-life)

Firjens:
[...]

PR #1330, opened 2026-05-22
fix(scanner): extract npm tgz via tar crate (closes #1243)

### LINKED ISSUE #1243: NPM packages fail to be extracted during scans due to tar not existing
### Description

Reporting that this is very much still an issue:
https://github.com/artifact-keeper/artifact-keeper/issues/722

Running web 1.1.3 and server 1.1.9

### Environment Information

- Deployment type: 
- Operating System: 
- Architecture: 
- Database: 
- Reverse Proxy: 
- Network setup: 

### Backend Version

1.1.9

### Web UI Version

1.1.3

### Package Format

npm

### Expected Behavior

With scan on proxy turned on, I'd expect the vulnerability scanning to run. No logs indicate automatic scanning is running/exists and manual scans result in same issue as what has been reported in the past.

When using npm as a pull in cache, I'd expect scans to automatically run and block any packages based on my policies.

### Observed Behavior

No automatic scanning is running. Manual scanning completes almost immediately with 0 vulnerabilities report. Exact same issue as reported previously per link above

### Steps to Reproduce

Same as in the original report.

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
Closes #1243. Follow-up to #722.

## Summary

Scanner archive extraction no longer depends on host `tar`/`unzip` binaries. `ScanWorkspace::extract_archive` previously shelled out to `tar xzf` and `unzip` to extract scan inputs. The Alpine variant of the backend container image (`docker/Dockerfile.backend.alpine`) does not install `tar` or `unzip`, so npm `.tgz` and other archive scans silently failed at the `Command::new("tar")` call. The workspace then fell through to scanning the raw archive bytes, and Trivy returned zero findings -- exactly what #1243 and the original #722 report describe.

Extraction now runs in-process via the `tar`, `flate2`, and `zip` crates (already direct deps in `backend/Cargo.toml`). CPU work is offloaded to `tokio::task::spawn_blocking` so large archives can't stall the runtime. Removing the subprocess dependency benefits every deployment variant (slim Debian, Alpine, distroless) without per-image package list maintenance, which is why this is preferred over the alternative fix of `apk add tar unzip` in the Alpine Dockerfile.

Supported extension set is unchanged: `.tar.gz`, `.tgz`, `.crate`, `.gem`, `.zip`, `.whl`, `.jar`, `.war`, `.ear`, `.nupkg`, `.egg`. Path-traversal entries are rejected by `tar::Archive::unpack` (refuses absolute paths and `..` components) and `ZipFile::enclosed_name`.

- [ ] N/A -- this is not a bug fix

Five new tests in `services::scanner_service::tests`:

- `test_extract_archive_npm_tgz` -- builds a real npm-shaped tgz containing `package/package.json` + `package/index.js`, extracts it, asserts contents on disk.
- `test_extract_archive_crate_uses_targz` -- pins that Cargo `.crate` files share the gzipped-tar wire format.
- `test_extract_archive_zip_jar` -- exercises the zip codepath (`META-INF/MANIFEST.MF` + nested class file).
- `test_extract_archive_unknown_extension_is_noop` -- non-archive inputs return `Ok(())` without touching the dest dir.
- `test_extract_archive_corrupt_tgz_returns_error` -- corrupt input surfaces an `AppError::Internal` instead of silently succeeding.

All five would have failed against the buggy `Command::new("tar")` path on a host without `tar` installed (which is exactly what the Alpine image is).

## Test Checklist

- [x] Manually tested locally (`cargo test --workspace --lib`: 9658 passed)

## API Changes

- [x] N/A - no API changes


brandonrc:
## 3-agent review verdict

**Consensus: APPROVE** (Code: APPROVE WITH NITS; Security: APPROVE WITH NITS; Backend Architect: APPROVE with follow-ups)

Solid fix. Eliminates a real silent-failure (Alpine image had no `tar`/`unzip`, npm `.tgz` scans returned zero findings). In-process design works across all image variants. `spawn_blocking` placement correct. No new dependencies (tar/flate2/zip already direct deps). Net security improved: no shell-exec, `set_preserve_permissions(false)`, perms dropped.

### Correctness validated
- npm `package/` prefix handled correctly by `tar::Archive::unpack`
- `.crate` aliasing to TarGz correct (Cargo `.crate` is gzipped tar)
- `.gem` mapped to plain Tar correct
- Path traversal: `tar::Archive::unpack` rejects `..` + absolute; `zip::ZipFile::enclosed_name()` returns None for traversal (correctly skipped)
- Error propagation: `JoinError` → `AppError::Internal`, inner Result flattened with `?`
- Streaming: tar streams via GzDecoder, zip per-entry via `std::io::copy`

### Security hardening follow-ups (file as separate issues)

1. **MEDIUM: Zip-bomb / decompression-bomb cap**. No size ceiling on `std::io::copy` or `unpack`. A 1KB tgz can blow up to many GB. Wrap entry reader in `Read::take(256_000_000)` per entry + cumulative 2GB hard cap. Add tracker.
2. **MEDIUM: Symlink filtering in tar**. Crafted `.gem`/`.tgz` could plant symlinks the scanner read-path follows. Add `archive.set_unpack_xattrs(false)` and filter entry types to `Regular | Directory` only.
3. **LOW-MEDIUM: Entry-count cap**. Pathological 1M-entry archive triggers 1M file syscalls. Add `MAX_ENTRIES = 100_000`.
4. **LOW: Partial-extraction docs**. Mid-archive error leaves partial files. Acceptable for tempdir-rooted scan workspace; document the contract.

### Other follow-ups (Backend Architect)

5. **Scanner binary availability check**: same class as this bug — verify Alpine image ships `trivy`/`grype` binaries since the scanners themselves still subprocess. File tracker.
6. **Format extensibility**: when `.tar.zst`/`.tar.xz` arrives (issue #1296 deferred), refactor to `Decoder` trait returning `Box<dyn Read>`. `zstd` and `xz2` crates already direct deps. Refactor will be mechanical.

### Other nits (Low)

- No regression test for path traversal (the crates handle it but a pinning test defends future refactors)
- `unpack_zip` silently `continue`s on `enclosed_name() == None`; add `tracing::warn!` to surface suspicious archives in logs
- No dedicated `.gem` test (covered by tar path but worth pinning the branch)
- `archive.set_overwrite(true)` is the default; harmless redundancy

Mergeable. Filing 6 follow-ups for security hardening + extensibility.

brandonrc:
Rebased onto current main (`5867929`).

Conflicts resolved:
- `CHANGELOG.md` — keep-both, both #1318 (lxc) and this PR's scanner extraction entry now sit under `### Fixed` for v1.2.0.
- `backend/src/services/scanner_service.rs` — auto-merged cleanly against #1318/#1323/#1324/#1325 (none of those touched `ScanWorkspace::extract_archive`).

Local verification on the rebased commit:
- `cargo fmt --check` clean
- `cargo test --workspace --lib scanner_service` — 359 passed, 0 failed (including the new `extract_archive` unit tests)

Force-pushed with `--force-with-lease`.

Issue #1183, opened 2026-05-12
test(proxy): per-handler behaviour-pin test that streaming migration didn't revert to buffered


brandonrc:
Follow-up to #1181 review. Architect S1 finding.

Each of the 5 migrated handlers (maven catch-all, goproxy .zip, gitlfs blob, alpine .apk, debian pool) now calls `proxy_fetch_streaming`. A silent regression to `proxy_fetch` would compile and pass all existing tests because the streaming primitive itself has its own unit tests in `proxy_service::tests`.

Recommended fix:
- One small test per handler that asserts the response body is a streamed body (e.g. via a recording mock `ProxyService` that tracks which method was called) OR that `Content-Length` matches `size_bytes` from the streaming-path metadata sidecar contract.
- Keeps the migration sticky against accidental rebases / merges that re-introduce buffering.

PR #1334, opened 2026-05-22
fix(proxy): migrate remaining buffered fetches to streaming (closes #1215)

### LINKED ISSUE #1215: resolve_virtual_download and remaining proxy paths still use buffered fetch
### Problem Statement

The tee stream migration (#895) converted Remote proxy fetches to streaming (`proxy_fetch_streaming`) in several format handlers (APT, Maven, Go, Git LFS, Alpine). However, `resolve_virtual_download` — which is the code path used when a Virtual repository resolves an artifact through its Remote members — still buffers the entire upstream response into memory before serving the client.

Additionally, `try_remote_or_virtual_download` (used by rpm, rubygems, puppet, hex, huggingface, cran, ansible) still calls the buffered `proxy_fetch` on its Remote branch rather than `proxy_fetch_streaming`.

This means any artifact fetched through a Virtual repo or through one of the above format handlers' shared fallback path does not benefit from the streaming work already landed on `main`.

### Proposed Solution

1. Migrate the Remote branch in `try_remote_or_virtual_download` to use `proxy_fetch_streaming`.
2. Investigate and migrate `resolve_virtual_download` to stream through Remote members rather than buffering.

### Alternatives Considered

Leave the buffered path for Virtual resolution since it involves iterating multiple members with fallback logic — streaming adds complexity when the first member may 404 and the next member needs to be tried. This is a valid trade-off but limits the usefulness of Virtual repos as proxies for large artifacts.

### Additional Context

Is there a timeline for the next release? The fixes I'm tracking are all submitted — once a release is cut I can validate internally.

brandonrc:
## Summary

The buffered `proxy_fetch` helper still backed two shared download paths after the #1183 / #1294 migrations:

- `try_remote_or_virtual_download` Remote arm, used by rpm, rubygems, puppet, hex, huggingface, cran, and ansible.
- `resolve_virtual_download` Remote-member arm, which fronts every Virtual repository download regardless of format.

Both buffered the full upstream body before responding, re-introducing the OOM-kill class that #895 / #1181 / #1294 closed for direct Remote fetches in the Maven, Go, Git LFS, Alpine, Debian, and generic download handlers.

This PR closes both gaps:

- Adds `proxy_fetch_streaming_with_disposition`, a streaming sibling of `proxy_fetch_streaming` that also forwards the `Content-Disposition: attachment; filename="..."` header so the streaming and buffered code paths emit equivalent outbound headers.
- Adds `resolve_virtual_download_streaming`, a first-match-wins resolver that drives Remote members through the streaming helper and only buffers Local / Staging members. Local members already serve from the local storage backend so buffering them does not regress memory pressure; their bodies are bounded by the artifact already being on disk.
- Routes both arms of `try_remote_or_virtual_download` through the new helpers. Iteration semantics are preserved: each Remote member's streaming fetch can still fail and fall through to the next member; once a member produces an `Ok(response)` the outbound response commits, which matches the buffered path's existing first-success-wins behaviour.

Linked issue (required): closes #1215.

Three source-level pin tests in the #1183 style assert that the migrated arms still call `proxy_fetch_streaming_with_disposition(` / `resolve_virtual_download_streaming(`. A silent revert to `proxy_fetch(` / `resolve_virtual_download(` would fail CI:

- `test_try_remote_or_virtual_download_remote_uses_streaming_helper_1215`
- `test_try_remote_or_virtual_download_virtual_uses_streaming_resolver_1215`
- `test_resolve_virtual_download_streaming_uses_streaming_helper_1215`

## Test Checklist
- [x] Manually tested locally (`cargo test --workspace --lib` clean, 9739 passing)

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -404,12 +404,47 @@ pub async fn proxy_fetch_streaming(
     path: &str,
     default_content_type: &str,
 ) -> Result<Response, Response> {
+    proxy_fetch_streaming_with_disposition(
+        proxy_service,
+        repo_id,
+        repo_key,
+        upstream_url,
+        path,
+        default_content_type,
+        None,
+    )
+    .await
+}
+
+/// Streaming sibling of [`proxy_fetch`] that also forwards a
+/// `Content-Disposition: attachment; filename="…"` header on the
+/// outbound response.
+///
+/// Same body and cache semantics as [`proxy_fetch_streaming`]; only the
+/// outbound response headers differ. Used by [`try_remote_or_virtual_download`]
+/// so format handlers that previously buffered via `proxy_fetch` +
+/// `build_download_response` keep the attachment filename on the
+/// streaming code path (#1215).
+pub async fn proxy_fetch_streaming_with_disposition(
+    proxy_service: &ProxyService,
+    repo_id: Uuid,
+    repo_key: &str,
+    upstream_url: &str,
+    path: &str,
+    default_content_type: &str,
+    content_disposition_filename: Option<&str>,
+) -> Result<Response, Response> {
     let repo = build_remote_repo(repo_id, repo_key, upstream_url);
     let result = proxy_service
         .fetch_artifact_streaming(&repo, path)
         .await
         .map_err(|e| map_proxy_error(repo_key, path, e))?;
-    build_streaming_response(result, default_content_type).map_err(|e| {
+    build_streaming_response_with_disposition(
+        result,
+        default_content_type,
+        content_disposition_filename,
+    )
+    .map_err(|e| {
         map_proxy_error(
             repo_key,
             path,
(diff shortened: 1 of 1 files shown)


brandonrc:
## 3-agent review verdict

**Consensus: APPROVE** (Code: APPROVE with nit; Backend Architect: APPROVE with follow-ups; audit-performance: APPROVE with one LOW finding)

Clean migration closing the last large-body buffering on the shared download surface. Both arms of `try_remote_or_virtual_download` are now streaming. Both buffered `resolve_virtual_download` and new `_streaming` resolver dispatch through `virtual_member_fetch_strategy` — iteration semantics identical, first-match-wins preserved.

### Validated
- **Memory bound**: ~14 MiB per in-flight on S3 path (4 MiB tee + 10 MiB WriteMultipart), 4 MiB on filesystem (TEE_CHANNEL_DEPTH=64).
- **Backpressure**: bounded mpsc → `tx.send().await` blocks when storage falls behind → closes TCP recv window upstream.
- **Cache-write integrity**: tee pattern from #1181 reused without modification. Clean EOF on upstream body end.
- **TTFB improvement**: dropped from `artifact_size / upstream_bandwidth` to effectively `upstream_headers_latency`. For a 500 MiB rpm at 100 MB/s, ~5-second TTFB win.
- **Iteration**: sequential priority-ordered (`ORDER BY vrm.priority`); intentional for first-match semantics.
- **Affected formats** (rpm/rubygems/puppet/hex/huggingface/cran/ansible): no compat concerns; metadata endpoints unaffected.

### Follow-ups (non-blocking)

1. **Backend Architect**: verify `fetch_artifact_streaming` preserves stale-cache fallback parity with `fetch_artifact_with_cache_path` (the buffered cache-or-refetch path with stale fallback at proxy_service.rs:177-191). If missing, rpm/cran lose graceful degradation when upstream is down. File follow-up.
2. **Backend Architect**: 30 buffered `resolve_virtual_download` callers (npm, cargo, conan, maven, helm, etc.) remain. Track them; streaming resolver should eventually absorb them. File issue.
3. **audit-performance LOW**: Local/Staging member arm still buffers (intentional — bounded by local storage size, no OOM concern). A future PR could stream via `StorageService::get_stream`, but it's a scope change.
4. **Code Reviewer nit**: source-pin negative assertion `!window.contains("let (content, content_type) =\n            proxy_fetch(")` is whitespace-fragile. `cargo fmt` change to that line would silently disable the pin. Prefer `!window.contains("= proxy_fetch(")` or similar shape-only check.
5. **Integration test**: add one asserting `Content-Length` absent and `Transfer-Encoding: chunked` present on a large-body fixture.

Mergeable. Significant memory-safety win for rpm/CRAN/huggingface and every Virtual download.

Issue #1179, opened 2026-05-12
Storage GC: multi-arch OCI image index child manifests unprotected


brandonrc:
## Problem

Surfaced during review of #1144. The fix in that PR adds `NOT EXISTS` guards against `oci_tags` and `oci_blobs` to the orphan-collection SQL, which protects directly-tagged manifests and named blobs. It does not protect the per-architecture child manifests inside an OCI image index.

When a multi-arch image is pushed:
- The image index manifest (content-type `application/vnd.oci.image.index.v1+json` or `application/vnd.docker.distribution.manifest.list.v2+json`) gets its own `oci_tags` row pointing at the index digest.
- The child manifests embedded in the index are stored at `oci-manifests/<child-digest>` and have their own `artifacts` rows, but they are not directly tagged.
- If the child `artifacts` rows are soft-deleted (e.g., a tag was overwritten), the new GC guards do not protect them: only the index digest appears in `oci_tags`, not the child digests it references.

A user pulling the multi-arch image after GC would get `MANIFEST_UNKNOWN` for the arch-specific layers.

## Suggested approaches

Two paths:

1. **GC-time JSON inspection.** During the orphan scan, find all live oci_tags rows whose `manifest_content_type` indicates an index, read each manifest file from storage, parse the JSON, and collect `manifests[].digest` values into a protected set. Add a third `NOT EXISTS` clause keyed on that set. Adds I/O per index manifest per GC pass.

2. **Eager child-digest recording.** At push time, when an index manifest is committed, write a row to a new table `oci_manifest_refs (parent_digest, child_digest, repository_id)`. GC adds a fourth `NOT EXISTS` against that table. No I/O at GC time; one extra table.

Option (2) is the cleaner architecture; option (1) is the smaller change.

## Acceptance

- Push a multi-arch image, overwrite the tag to point at a new index, run GC, pull the original index by digest, all arch-specific layers resolve.
- Regression test covering the scenario.

PR #1336, opened 2026-05-22
feat(repositories): server-side Docker tag aggregation (closes #1193)

### LINKED ISSUE #1193: Docker repositories: add server-side tag aggregation (manifest → layer sum)
## Background

Web PR artifact-keeper/artifact-keeper-web#330 added a frontend Docker
tag-grouping view that aggregates artifacts client-side using Docker
registry path conventions
(`<image>/manifests/<tag>` vs `<image>/manifests/sha256:...` vs
`<image>/blobs/sha256:...`).

## The gap

Without backend assistance, the frontend cannot:

1. **Compute true image size (sum of layers).**  The aggregated `size_bytes`
   surfaced in the tag row is the manifest size only — typically a few KB
   while the real image is hundreds of MB.  The UI currently surfaces a
   tooltip caveat:

   > Manifest size only. Total layer size will be aggregated once backend
   > support lands.

2. **Identify which blobs belong to which manifest.**  The Docker manifest
   JSON lists its layer digests, but the frontend only sees flat artifact
   rows and has no way to walk the manifest content without re-fetching.

3. **Paginate efficiently.**  The frontend currently bumps `per_page=500`
   for Docker grouped view so the client aggregation sees the whole repo.
   Repos with thousands of layers silently truncate.  (Tracked as
   pr4-security MINOR-2 in the web PR.)

## Proposed shape

Mirror the Maven `?group_by=maven_component` work from ak#701:

```
GET /api/v1/repositories/{key}/artifacts?group_by=docker_tag
```

Response includes a `docker_tags: DockerTagResponse[]` array alongside
the existing `items`, where each entry aggregates per (image, tag):

```rust
pub struct DockerTagResponse {
    pub image: String,
    pub tag: String,
    pub manifest_digest: String,
    pub manifest_id: Uuid,
    pub total_size_bytes: i64,    // manifest + all referenced layers
    pub layer_count: i32,
    pub last_pushed_at: DateTime<Utc>,
    pub scan_status: Option<String>,
}
```

Behind the scenes the handler walks the manifest JSON (or a cached
manifest_layers join table) to sum layer sizes.

## Frontend impact (when this lands)

- `src/app/(app)/repositories/_lib/docker-grouping.ts` becomes a
  fallback for proxied / cached artifacts that don't have manifest
  metadata; primary aggregation comes from the server.
- The "manifest size only" tooltip can be removed.
- `effectivePageSize = 500` workaround in
  `repo-detail-content.tsx` can be removed.
- Server-side pagination becomes meaningful again.

Refs: artifact-keeper/artifact-keeper-web#330,
artifact-keeper/artifact-keeper#701 (Maven precedent)

brandonrc:
## Summary

Closes #1193.

Adds `?group_by=docker_tag` to the repository artifacts listing for Docker/OCI repositories, mirroring the existing Maven `?group_by=maven_component` pattern (#701). Instead of returning flat artifact rows, the response carries a `docker_tags` array where each entry aggregates per (image, tag) and reports the true on-disk image size (manifest body + every referenced layer blob).

The web Docker tag-grouping view in artifact-keeper-web#330 currently aggregates client-side from flat artifact rows, so it only sees the manifest body size (typically a few KB) for images that are hundreds of MB on disk. That view also has to crank `per_page=500` so the client aggregation sees the whole repo, defeating server-side pagination. Both problems collapse once the backend does the work.

### Response shape

```
GET /api/v1/repositories/{key}/artifacts?group_by=docker_tag
```

```jsonc
{
  "items": [],
  "pagination": { "page": 1, "per_page": 20, "total": 42, "total_pages": 3 },
  "docker_tags": [
    {
      "id": "...",
      "repository_key": "docker-hub",
      "image": "library/postgres",
      "tag": "16-alpine",
      "manifest_digest": "sha256:...",
      "total_size_bytes": 250000000,
      "layer_count": 0,
      "is_index": false,
      "last_pushed_at": "2026-05-20T...",
      "scan_status": "completed"
    }
  ]
}
```

### Implementation notes

- The OCI v2 PUT-manifest handler already computes `config_size + sum(layers.size)` and persists it as `artifacts.size_bytes` on the manifest row (`oci_v2.rs:2140-2160`). The new query is therefore a join, not a re-parse of every manifest body.
- For multi-arch image indexes, the per-arch child manifests recorded in `oci_manifest_refs` (migration 092) are summed in so the surfaced number matches what `docker pull` actually downloads across architectures.
- Digest references (`tag = "sha256:..."`) are filtered out to match the `/v2/<name>/tags/list` spec.
- `layer_count` is in the response struct but always returns 0 for now. Persisting it on the manifest artifact row at push time is a follow-up; computing it here would require re-fetching every manifest body from storage on each list call.
- `scan_status` is joined from the latest `scan_results` row per artifact via `LEFT JOIN LATERAL`.

### Frontend follow-up

Once this ships, `artifact-keeper-web` can:

1. Drop the "Manifest size only" tooltip caveat from the Docker tag-grouping view.
2. Remove the `effectivePageSize = 500` workaround in `repo-detail-content.tsx`.
3. Replace `docker-grouping.ts` client aggregation with the server-side `docker_tags` array (keep it as a fallback for proxied artifacts without manifest metadata).

- [x] N/A, this is not a bug fix

## Test Checklist

- [x] Unit tests added/updated (13 new tests: content-type detection, single-arch vs index sizing, empty-children fallback, scan-status preservation, i64 saturation, query deserialization, response serialization)
- [x] Manually tested locally (full repositories test suite, OCI test suite, OpenAPI validity test pass)

## API Changes

- [x] New endpoints have `#[utoipa::path]` annotations (extends existing `list_artifacts` handler)
- [x] Request/response types have `#[derive(ToSchema)]` (`DockerTagResponse` registered in `RepositoryApiDoc`)
- [ ] Migration is reversible (if applicable) (no schema changes)
- [ ] N/A, no API changes

## Verification

```
cargo fmt --check
cargo test --lib api::handlers::repositories::tests   # 254 passed
cargo test --lib oci                                   # 264 passed
cargo test --lib test_openapi_spec_is_valid            # 1 passed
```

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -1234,9 +1234,16 @@ pub struct ListArtifactsQuery {
     pub per_page: Option<u32>,
     pub q: Option<String>,
     pub path_prefix: Option<String>,
-    /// When set to `maven_component`, Maven/Gradle artifacts are grouped by
-    /// groupId, artifactId, and version.  Individual files (jar, pom, checksums)
-    /// appear in the `artifact_files` array of each component.
+    /// Server-side artifact grouping.
+    ///
+    /// Supported values:
+    /// - `maven_component`: Maven/Gradle artifacts are grouped by
+    ///   groupId, artifactId, and version.  Individual files (jar, pom,
+    ///   checksums) appear in the `artifact_files` array of each component.
+    /// - `docker_tag`: Docker/OCI artifacts are grouped by (image, tag),
+    ///   with `total_size_bytes` summed across the manifest config and
+    ///   referenced layer blobs.  The grouped rows are returned in the
+    ///   `docker_tags` array.
     pub group_by: Option<String>,
 }
 
@@ -1263,6 +1270,52 @@ pub struct ArtifactListResponse {
     /// Maven component grouping.  Only present when `group_by=maven_component`.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub components: Option<Vec<MavenComponentResponse>>,
+    /// Docker tag grouping.  Only present when `group_by=docker_tag`.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub docker_tags: Option<Vec<DockerTagResponse>>,
+}
+
+/// A Docker/OCI tag grouped by (image, tag).
+///
+/// `total_size_bytes` is the server-side aggregation of the manifest body
+/// plus every referenced layer blob.  This is what the UI should display
+/// as the on-disk image size; the previous client-side aggregation that
+/// only summed the manifest body itself reported a few kilobytes for
+/// images that are hundreds of megabytes on disk (artifact-keeper#1193).
+///
+/// For multi-arch image indexes the size is the sum across all per-arch
+/// child manifests recorded in `oci_manifest_refs`, so an `amd64+arm64`
+/// index reports the combined storage cost.
+#[derive(Debug, Serialize, ToSchema, Clone)]
+pub struct DockerTagResponse {
+    /// Representative manifest artifact ID.
+    pub id: Uuid,
+    /// Repository key this tag belongs to.
+    pub repository_key: String,
+    /// Image name (no registry host, no tag).  Maps to the OCI v2 `<name>`
+    /// path segment, which may include slashes (e.g. `library/postgres`).
+    pub image: String,
+    /// Tag string (e.g. `16-alpine`).  Never a `sha256:...` digest;
+    /// digest-only references are filtered out of the grouping.
+    pub tag: String,
+    /// Manifest content digest (e.g. `sha256:abcdef...`).
+    pub manifest_digest: String,
+    /// Total size in bytes of the manifest plus all referenced layer blobs.
+    /// For image indexes, this sums across child manifests.
+    pub total_size_bytes: i64,
+    /// Number of layer blobs referenced by the manifest.  For image indexes
+    /// this is the sum of layer counts across child manifests.  `0` when
+    /// the manifest could not be parsed.
+    pub layer_count: i32,
+    /// Whether this manifest is a multi-arch image index.
+    pub is_index: bool,
+    /// Last push (or update) timestamp from the underlying `oci_tags` row.
+    pub last_pushed_at: chrono::DateTime<chrono::Utc>,
+    /// Latest scan status (`pending`, `running`, `completed`, `failed`) from
+    /// `scan_results`, if the manifest has ever been scanned.  `None` when
+    /// the artifact has never been scanned.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub scan_status: Option<String>,
 }
 
 /// A Maven component grouped by GAV (groupId, artifactId, version).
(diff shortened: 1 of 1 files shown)


brandonrc:
## 3-agent review verdict

**Consensus: APPROVE** (Code: APPROVE WITH NITS; Backend Architect: APPROVE WITH FOLLOW-UPS; DB Optimizer: APPROVE WITH CONCERN — verified spurious)

### DB Optimizer's "migration 092 missing" concern is incorrect
Independent verification: `git ls-tree origin/main backend/migrations/` confirms `092_oci_manifest_refs.sql` exists on main. The table schema (parent_digest, child_digest, repository_id, created_at) matches the PR's query usage exactly. Populated on manifest PUT via `record_oci_manifest_refs()` at `oci_v2.rs:488`. Pre-existing manifests handled by `oci_manifest_refs_backfill.rs`. GC orphan-protection at `storage_gc_service.rs:73`. The reviewer scanned only the PR diff (single file `repositories.rs`) and assumed the referenced table was new — it shipped as part of #1179.

### Validated
- **Reuses precomputed `size_bytes`**: oci_v2 PUT-manifest handler computes `config_size + sum(layers.size)` and persists it. Grouping query is a SQL join, not a re-parse.
- **Multi-arch via `oci_manifest_refs` (092)**: per-arch child manifests summed in. Tag reports full on-disk cost across architectures.
- **OCI spec compliance**: `WHERE POSITION(':' IN tag) = 0` excludes `sha256:...` references, matching `/v2/<name>/tags/list` spec.
- **`scan_status`**: LEFT JOIN LATERAL `ORDER BY created_at DESC LIMIT 1` returns latest scan status correctly.
- **13 new tests** cover content-type detection, single-arch vs index sizing, empty-children fallback, scan-status preservation, i64 saturation.
- **OpenAPI annotation**: registered. Validity test passes.

### Required follow-ups (file as v1.2.1 / v1.3.0)

1. **Typed enum for `group_by`** (Backend Architect): two values now, three is the inflection point. `enum GroupBy { MavenComponent, DockerTag }` before adding Helm chart / NPM scope.
2. **`layer_count: Option<i32>`**: returning 0 is misleading; UI can't distinguish "zero layers" from "not yet computed". Change to Option<i32>; persist at push time as separate scope-bumped issue.
3. **Index follow-up** (DB Optimizer LOW): composite `(artifact_id, created_at DESC)` partial index on `scan_results` to optimize the LATERAL subquery at scale. Currently 10k tags × N scans = 10k tiny sorts. Track for v1.2.1.
4. **MAX_FETCH=10k cap**: silently truncates pagination on huge repos. Push pagination into SQL via `COUNT(*) OVER()` or document the cap.
5. **`size_bytes` correctness on proxy-cached / migration paths**: add a debug-assert / metric counter when `size_bytes = 0` is encountered, so drift surfaces in prod.
6. **Wildcard escaping** on `q` param (Code Reviewer): `LIKE '%' || $2 || '%'` misbehaves if input contains `%`/`_`. UX issue, not security.

### Code-style nit (non-blocking)

- `cargo fmt --check` reports a small formatting nit in the test helper. Author may want to run `cargo fmt` before final merge (or fix in follow-up).

Mergeable. Excellent design pattern (mirroring `maven_component`, reusing precomputed size).

Issue #1340, opened 2026-05-22
Artifact Download Race Condition: Synchronizing File Access via Advisory Locks


axellpadilla:
### Description

To enforce strict resource consistency when an artifact record exists in the database (`artifacts` table) but the physical file is missing or being downloaded by another request. We must prevent race conditions during concurrent reads and ensure data integrity without blocking the entire service or causing refetch + concurrent writes I/O and downloads.

### Environment Information

- Deployment type: 
- Operating System: 
- Architecture: 
- Database: 
- Reverse Proxy: 
- Network setup: 

### Backend Version

latest

### Web UI Version

N/A

### Package Format

None

### Expected Behavior

Use `pg_advisory_xact_lock()` to enforce a **synchronous queue** for the specific artifact ID. This ensures only one process is handling the artifact at a time, eliminating the race condition. Example: https://dbfiddle.uk/KwsOjRRc

#### Mechanism Details:
1.  **Lock Granularity:** The lock will be scoped to the unique `artifact_id` (converted to a numeric key for PostgreSQL).
2.  **Serialization:** Before reading the file after a filesystem miss to keep performance untouch on other cases, the process attempts to acquire the lock within a database transaction.
3.  **Queueing:** If the lock is held by another process, the current process will not fail immediately. It should use `tokio::time::timeout` to bound lock acquisition and wait only for a short, configured duration.
4.  **Timeout/Failure:** This is optional because the pg lock is warranted to sease to exists, if the lock is held and the wait timeout expires, we treat the resource as perpetually unavailable, return a specific `507` status, and notify the client to retry later or fallback to a refetch (patch like PyPi only solution in #1283).
5.  **Lock Safety:** By wrapping the entire operation in a database transaction (`sqlx::Transaction`), the lock is automatically guaranteed to be released (via `commit()` or `rollback()`), preventing dangling locks even if the process crashes.

### Observed Behavior

In the current `download_package` flow, if multiple processes request the same artifact simultaneously, they execute the file retrieval (`storage.get(...)`) independently. This is a **read-time race condition** that can lead to:
- Duplicate refetch/write-back: Multiple concurrent miss handlers can all re-download the same upstream object and write it back concurrently.
- Inconsistent cache state: If one request restores the object while others also race, the proxy cache can become inconsistent and the error condition can persist.
- The current design fails to address this read-time concurrency and relies on the file system's inherent atomicity, which is insufficient for a distributed proxy.

### Steps to Reproduce

[...]

PR #1342, opened 2026-05-22
fix(nuget): Refactor NuGet remote download to use proxy cache helper

### LINKED ISSUE #1343: Replicate PyPi fix for nuget centralizing this on proxy_helpers for 1.2
Exact same problem than PyPi #1283 
Step 1 one of proposal #1340

axellpadilla:
Summary
This PR refactors and hardens remote artifact fetching and caching for NuGet and PyPI:
Closes #1343 

- **Proxy Cache Logic:** Extracts generic proxy cache logic into `proxy_helpers`, making it reusable across package types.
- **NuGet:** Uses the shared `get_cached_or_refetch` helper for remote flatcontainer downloads, ensuring consistent cache behavior and reducing duplication.
- **PyPI:** Updates the PyPI handler to use the new shared helper for remote artifact caching.
- **Tests:** Adds a unit test for NuGet to verify correct fallback to remote cache behavior.

These changes improve maintainability, reduce code duplication, and ensure consistent caching logic for remote package fetches in preparation for next steps to handle race conditions on downloads referenced in #1340 .

- [x] Unit test added for NuGet remote cache fallback behavior

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc:
Three independent reviews completed (code-quality, performance, architecture). Aggregate verdict: **REQUEST_CHANGES** before merge. The centralization itself is the right direction (architectural review approved that part), but two concrete concerns block as-is:

### 1. The 39-failure benchmark in the PR body needs explanation

> Comparison Branch (Patched #1342): NuGet `1136 ms` median, `201 successful / 39 failed` (vs current main `156 ms`, `240/0`)

Reviewer read: this is plausibly a real regression on the Remote-repo path, not benchmark noise. Pre-PR code does a single `storage.get(&artifact.storage_key)` (filesystem/S3 read) and surfaces NotFound as 500 fast. Patched code keeps that read but on NotFound falls through to `proxy_helpers::proxy_fetch` (full upstream HTTP round-trip) plus a separate best-effort `storage.put` write-back. For a remote repo, ANY storage read miss (eviction, lazy-replicated artifact, stat-vs-get races, S3 eventual consistency) now triggers a multi-hundred-ms upstream fetch instead of a 500. The shape — 156ms→1136ms median + new 16% failure rate — is exactly what you'd see if a meaningful fraction of objects were not actually present locally and the old benchmark counted fast 500s as "successful" (and the new path genuinely round-trips to nuget.org which then rate-limits at scale).

Please re-run the benchmark with:
- N=1 concurrency AND N=240 on both branches, side by side
- 2xx-vs-5xx split in the "successful" column (a 500-status fast response is NOT a success)
- Cold-cache vs warm-cache delineation
- Upstream RTT recorded separately

### 2. Test coverage gap on the new branch

The single new unit test in `nuget.rs` (around `:1729+`) seeds storage with bytes and exercises the **cache HIT** path. It does NOT cover the actual new logic: `NotFound → refetch closure → write-back`. That's the branch that contains the 39-failure surface. Please add a second unit test that:
1. Sets `storage.get` to return `NotFound`
2. Confirms the refetch closure is invoked exactly once
3. Confirms `storage.put` is called with the right key
4. Confirms the bytes are returned to the caller unchanged

### 3. Dual write-back warrants a doc comment

`backend/src/api/handlers/proxy_helpers.rs:642-682` does its own `storage.put(storage_key, bytes)` AFTER calling `proxy_fetch` which already writes to its own proxy-cache key. That's two cache writes per miss to different keys. Either:
- Drop the redundant `storage.put` (the proxy layer already cached, downstream readers can pull from there), OR
- Document why both are needed in the helper's doc comment (the artifact storage_key is the canonical "where this lives on disk" per the artifact row, and the proxy cache is a separate moving target — that's the likely justification but it needs to be written down)

### 4. Streaming follow-up (non-blocking)

The architectural review noted the helper buffers responses (`Bytes`) which is consistent with the original `pypi.rs` shape. NuGet packages typically run 50-500 MiB; mass-adopting this helper from `npm`/`deb`/`rpm`/`maven` would step backward from the streaming work in #895 that `alpine.rs` already uses (`proxy_fetch_streaming`). Recommend filing a follow-up issue for a `get_cached_or_refetch_streaming` variant BEFORE adopting this helper for large-body formats. Not a blocker for the current PR scope (NuGet flatcontainer + PyPI), but should be in the follow-ups list.

### Verdict by reviewer

- Code quality: **REQUEST_CHANGES** (test gap, dual-write justification, duplicated else arms)
- Performance: **NEEDS_BENCHMARK** (lean REQUEST_CHANGES — author benchmark suggests a real regression)
- Architecture: **APPROVE** with streaming follow-up

Once the benchmark is re-run cleanly and the NotFound-path test added, this is ready. The structure sets up #1355 (advisory-lock single-flight) very cleanly per the architectural review, so this is unblocking work — just needs the regression question settled first.

axellpadilla:
1. Ok will look into it, but because the root cause of most problems should be the race condition (a harder problem to takle with more changes) I think those are related to the concurrent writes or even upstream rate limits, a problem this pr shouldn't solve. Still, I would like to reproduce consistently and fact base this if needed.
2. Ok
3. Definitely let me look into it, could improve or even statistically hide 1 for the live benchmark.
4. Good to know this, after the race condition is fixed and streaming achieved this will be a strong and pretty efficient proxy.

brandonrc:
Hi @axellpadilla — bumping this. Three review rounds (code, security, architecture, performance) all converged on the same blocker: your benchmark in the PR body shows NuGet `201 successful / 39 failed (1136 ms median)` on this branch vs `240 / 0 (156 ms)` on current main. That's a 7x slowdown + 16% failure rate.

To unblock the merge for v1.2.0, we need:

1. **Fresh benchmark** with methodology documented: N=1 and N=240 concurrency side-by-side, 2xx-vs-5xx split (a 500-status fast response is not a success), cold-vs-warm cache delineation, upstream RTT recorded separately.
2. **Unit test for the NotFound -> refetch -> write-back path** in `proxy_helpers.rs`. The current test only exercises the cache HIT path.
3. **Dual-write justification** in the helper doc-comment, OR drop the redundant `storage.put` (the proxy layer already cached the body, downstream readers can pull from there).

Architecture review confirmed the refactor direction is correct and the helper sets up #1355 nicely once that lands. If you can't re-run the benchmark in the next 48h, ping us and we'll take over from your branch.

brandonrc:
Moving to v1.2.1 milestone.

The PR closes #1343 (a refactor to centralize proxy cache logic) and is the prerequisite for #1355 (race-condition single-flight). Since #1355 was already deferred to v1.2.1, this refactor can land alongside it rather than blocking v1.2.0.

The HIGH perf concern from the three reviewers (7x slowdown + 16% failure rate in your own benchmark) still needs to be resolved — but with v1.2.0 unblocked, there's no time pressure. Please address the benchmark methodology when you can, and we'll land #1342 + #1355 together for v1.2.1.

axellpadilla:
Hi guys, I lost my workspace with my previous benckmark scripts, thats why I took a little longer to try to recover the logs and scripts, preliminary results:
1. The errors where upstream rate limits, I used live real endpoints, so ended up with some 502 errors.
2. I will share the final remade benchmark script and results shortly.
3. Dual-write is intentional. We can’t remove it because of the race condition: readers won’t have anything to fetch until the first refetch completes, and waiting for that would require holding the lock.

axellpadilla:
@brandonrc  had to do deeper research, and found another related race condition related to write, so we should close this PR because it truly worsenes the race downstream and focus on the race condition fix, please check current final proposal on https://github.com/artifact-keeper/artifact-keeper/pull/1355

PR #1355, opened 2026-05-23
fix(fetch) Fix/race condition downloads

### LINKED ISSUE #1340: Artifact Download Race Condition: Synchronizing File Access via Advisory Locks

### LINKED ISSUE #1343: Replicate PyPi fix for nuget centralizing this on proxy_helpers for 1.2

axellpadilla:
## Summary
This PR is the follow-up to the earlier fast fix in #1342 and is intended to close #1343 and close #1340.

It fixes a race condition where an artifact database row exists but the local storage object is temporarily missing during concurrent reads. The previous advisory-lock attempt increased contention; this version switches to coordinated hydration/retry logic to reduce duplicate recovery work and improve reliability per node under concurrent access.

What changed:
- Added coordinated retry/hydration logic for missing artifact file reads.
- Added generic proxy cache recovery helper used by Alpine, NPM, NuGet, and PyPI handlers.
- Kept Maven out of this lock-backed path in this change.
- Fixed filesystem temp-write naming to avoid collisions during concurrent cache writes.
- Refactored affected query paths to include artifact IDs for better traceability in logs/diagnostics.
- Added/updated regression-oriented tests for proxy cache refetch behavior and updated helper call signatures.

### Latest benchmark results (mock upstream, same-artifact herd)
Scenario:
- Upstream mode: mock
- Delay model: 1.500s initial delay + 0.200s per response chunk
- Concurrency: 5

#### Current branch: fix/race-condition-downloads
| Ecosystem | Cache mode | Median (ms) | Success | Failed | Upstream fetches | Cache writes |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| NuGet same-artifact herd | cold | 1755 | 5 | 0 | 2 | 1 |
| NuGet same-artifact herd | stale | 1739 | 5 | 0 | 2 | 1 |
| NuGet same-artifact herd | warm | 33 | 5 | 0 | 0 | 0 |
| PyPI / uv | cold | 6015 | 5 | 0 | 14 | 2 |
| PyPI / uv | stale | 5299 | 5 | 0 | 14 | 2 |
| PyPI / uv | warm | 354 | 5 | 0 | 0 | 0 |

#### main
| Ecosystem | Cache mode | Median (ms) | Success | Failed | Upstream fetches | Cache writes |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| NuGet same-artifact herd | cold | 1760 | 1 | 4 | 10 | 1 |
| NuGet same-artifact herd | stale | 1741 | 3 | 2 | 10 | 3 |
| NuGet same-artifact herd | warm | 39 | 5 | 0 | 0 | 0 |
| PyPI / uv | cold | 5671 | 5 | 0 | 30 | 8 |
| PyPI / uv | stale | 5505 | 5 | 0 | 30 | 6 |
| PyPI / uv | warm | 354 | 5 | 0 | 0 | 0 |

#### fix/nuget-500-on-parallel-get
| Ecosystem | Cache mode | Median (ms) | Success | Failed | Upstream fetches | Cache writes |
| --- | --- | ---: | ---: | ---: | ---: | ---: |
| NuGet same-artifact herd | cold | 1744 | 4 | 1 | 10 | 4 |
| NuGet same-artifact herd | stale | 1742 | 3 | 2 | 10 | 3 |
| NuGet same-artifact herd | warm | 34 | 5 | 0 | 0 | 0 |
| PyPI / uv | cold | 5576 | 5 | 0 | 30 | 9 |
| PyPI / uv | stale | 6023 | 5 | 0 | 30 | 6 |
| PyPI / uv | warm | 347 | 5 | 0 | 0 | 0 |

[benchmark-report.md](https://github.com/user-attachments/files/28360694/benchmark-report.md)
[benchmark-proxy-cache-regression.sh](https://github.com/user-attachments/files/28360475/benchmark-proxy-cache-regression.sh)
[benchmark-proxy-cache-race-mock.sh](https://github.com/user-attachments/files/28360024/benchmark-proxy-cache-race-mock.sh)
[generate-proxy-cache-final-report.sh](https://github.com/user-attachments/files/28360025/generate-proxy-cache-final-report.sh)

Key takeaway:
- NuGet cold/stale failures (502s) observed on main are eliminated in this branch under the same mock-race conditions.
- Upstream fetch amplification is significantly reduced in cold/stale paths (NuGet: 10 -> 2; PyPI: 30 -> 14).

## Regression test (required for fix/* PRs)
- [x] This PR is a fix/* AND adds/updates a test that would have caught the bug

## Test Checklist
- [x] Manually tested locally

## API Changes
- [ ] New endpoints have #[utoipa::path] annotations
- [ ] Request/response types have #[derive(ToSchema)]
- [ ] OpenAPI spec validates: cargo test --lib test_openapi_spec_is_valid
- [x] N/A - no API changes


brandonrc:
Moving to v1.2.1 milestone. The advisory-lock single-flight you add here is the right fix for #1340, but the PR depends on #1342 (NuGet proxy_helpers refactor) which currently has REQUEST_CHANGES from three reviewers — the author's own benchmark showed a 7x slowdown + 16% failure rate on the patched branch (#1342 PR body). Until #1342 is either fixed or replaced, this PR's foundation is unstable. Plan: address #1342, then rebase #1355 directly on main, then merge for v1.2.1.

axellpadilla:
I removed the advisory lock because tthe pool was getting quickly filled, simplified to use inmemory per node serialization, this should help considering the retrial now available. The benchmark script I tried to use to avoid upstream limits is this one, it is still a work in process, and I suggest checking how you can include this on the e2e tests, so we have a roadmap where cross-node race condition is correctly handled or documented. 
[benchmark-proxy-cache-regression.sh](https://github.com/user-attachments/files/28360475/benchmark-proxy-cache-regression.sh)
[benchmark-proxy-cache-race-mock.sh](https://github.com/user-attachments/files/28360024/benchmark-proxy-cache-race-mock.sh)
[generate-proxy-cache-final-report.sh](https://github.com/user-attachments/files/28360025/generate-proxy-cache-final-report.sh)

Current report (purely IA made, the increased time makes no sense could be just stastiscal or machine difference or the mock behavior):
[benchmark-report.md](https://github.com/user-attachments/files/28360694/benchmark-report.md)

axellpadilla:
This is a work in progress with the idea to fix this, follow up:
- Final cleanup if this approach is preferable for a first implementation to avoid complex transaction pool handling or advisory locks right now
- Adding the missing coverage and extra tests like missing file but row exists.
- Improve error semantics and mappings for the new cases.
- Multi-node behavior is either covered with DB advisory locks or explicitly not claimed to keep this smaller for the first phase.

PR #1359, opened 2026-05-23
fix: allow WASM plugin formats in repository creation - resolves #1358

### LINKED ISSUE #1358: Bug: WASM plugin formats cannot be used when creating repositories
### Description

After successfully installing and enabling a WASM plugin, attempting to create a
repository with the plugin's format name via `POST /api/v1/repositories` always
returns a validation error — even though the plugin is loaded and the format is
registered in the `format_handlers` table.

### Root cause

`parse_format()` in `backend/src/api/handlers/repositories.rs` is a static
`match` over the `RepositoryFormat` enum. The enum has no dynamic/plugin variant
and the function has no knowledge of `format_handlers` rows written by the plugin
system at runtime. The two subsystems are completely disconnected:

- **Plugin registration path**: installs WASM, writes `format_handlers` row,
  registers format in the proxy/runtime dispatch layer.
- **Repository creation path**: calls `parse_format()` → static enum match →
  hard-fails with `Invalid format` before any DB lookup occurs.

The `format_handlers` table is only consulted by the proxy/runtime dispatch
layer, never by the repository management API.

### Proposed fix

In `parse_format()`'s call site at `create_repository` (line ~838), fall back to
a `format_handlers` DB query when the static match fails:

- Row found, `is_enabled = true` → accept the format, use
  `RepositoryFormat::Generic` as the DB enum value and carry the plugin's
  `format_key` through to `repositories.format_key` (which the existing plugin
  dispatch path already reads).
- Row found, `is_enabled = false` → return descriptive 400: "Plugin format
  handler '…' is disabled."
- No row → preserve the existing "Invalid format" 400.

This requires no schema changes and no changes to the service layer — only the
`create_repository` handler needs updating.

### Environment Information

- Deployment type: Docker
- Operating System: Linux
- Architecture: linux/arm64 (also reproducible on amd64)
- Database: n/a
- Reverse Proxy: n/a
- Network setup: n/a

### Backend Version

v1.1.9 

### Web UI Version

n/a

### Package Format

None

### Expected Behavior

The repository is created successfully with format `myplugin`, backed by the
installed WASM plugin handler.

### Observed Behavior

```json
HTTP 400 Bad Request
{
  "code": "VALIDATION_ERROR",
  "message": "Invalid format: myplugin"
}
```

### Steps to Reproduce

1. Upload a WASM plugin (e.g. `myplugin-plugin`) via `POST /api/v1/plugins`
2. Enable the plugin via `POST /api/v1/plugins/{id}/enable`
3. Confirm the plugin is active: `GET /api/v1/plugins/{id}` → `status: active`
4. Confirm the format handler exists in the DB:
   `SELECT * FROM format_handlers WHERE format_key = 'myplugin';`
   → returns a row with `is_enabled = true`, `handler_type = 'Wasm'`
5. Attempt to create a proxy repository:
   ```json
   POST /api/v1/repositories
   {
     "key": "my-plugin-proxy",
     "name": "My plugin Proxy",
     "format": "myplugin",
     "repo_type": "Proxy",
     "upstream_url": "https://example.com"
   }
   ```

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

fabiogermann:
# Summary

## What this PR does

Fixes registration of custom formats through a plugin.

## The bug the PR Resolves #1358 

`parse_format()` is a static `match` over the compile-time `RepositoryFormat`
enum. It has zero knowledge of `format_handlers` rows written by the WASM plugin
system at runtime. Result: any format registered by a plugin is always rejected
with `{"code":"VALIDATION_ERROR","message":"Invalid format: myplugin"}`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -7611,20 +7611,6 @@ mod tests {
         )
         .await;
 
-        // Cleanup before asserting so we always delete even on failure.
-        sqlx::query("DELETE FROM repositories WHERE key = $1")
-            .bind(&repo_key)
-            .execute(&pool)
-            .await
-            .ok();
-        cleanup_format_handler(&pool, &format_key).await;
-        sqlx::query("DELETE FROM users WHERE id = $1")
-            .bind(user_id)
-            .execute(&pool)
-            .await
-            .ok();
-        let _ = std::fs::remove_dir_all(&storage_dir);
-
         let Json(resp) = result.expect("create_repository with enabled plugin format must succeed");
         assert_eq!(
             resp.format, "generic",
@@ -7639,6 +7625,21 @@ mod tests {
                 .fetch_optional(&pool)
                 .await
                 .expect("query format_key");
+
+        // Cleanup after reading so we don't delete the row before asserting.
+        sqlx::query("DELETE FROM repositories WHERE key = $1")
+            .bind(&repo_key)
+            .execute(&pool)
+            .await
+            .ok();
+        cleanup_format_handler(&pool, &format_key).await;
+        sqlx::query("DELETE FROM users WHERE id = $1")
+            .bind(user_id)
+            .execute(&pool)
+            .await
+            .ok();
+        let _ = std::fs::remove_dir_all(&storage_dir);
+
         assert_eq!(
             stored.as_deref(),
             Some(format_key.as_str()),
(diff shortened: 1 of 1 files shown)


brandonrc:
Closing in favor of PR #1411 which moves the plugin-format resolution into `RepositoryService::resolve_format` (the right architectural layer). PR #1411 also discovered + fixed a TOCTOU gap in the existing service-layer validation that this PR's wrong-layer approach was masking.

Thanks @fabiogermann for the original analysis — the bug is real and the fix is correct, just lives one layer deeper. PR #1411 attributes the diagnosis to this PR.

PR #1382, opened 2026-05-24
fix(promotion): evaluate quality gate before staging-source check (closes #1376)

### LINKED ISSUE #1376: [bug] Quality-gate promotion: 'must be staging repo' check fires before gate evaluation

### Failing assertions
- `quality-gate-blocks-upload / Strict gate evaluates as VIOLATED (passed=false)`: gate body shows `passed: false` correctly, but the test reads it as `passed=''` — see body for shape.
- `quality-gate-blocks-upload / Promotion of violating artifact is rejected (403/409/422)`: returned HTTP 400 with `"Source repository must be a staging repository"`.
- `quality-gate-blocks-upload / Promotion succeeds after gate loosened`: HTTP 400 with same message.

### Root cause
The promotion handler validates "source must be a staging repository" before evaluating the quality gate. So any non-staging promotion 400s on shape grounds, never reaching the gate. The test sets up a non-staging promotion specifically to assert gate-blocking. Either:
- Test should use a staging-typed source repo so the gate gets a chance to fire, OR
- Promotion should evaluate the gate before the staging-type check (correct behavior — gate violations are higher priority than shape errors).

Recommend handler reorder: gate first, then staging-type validation.

Also there's an `is_enabled` serialization issue in the strict gate response separately — `passed=false` is correct but the field comes back coercing to empty string in jq view; check JSON shape stability.

### Acceptance
- All three assertions pass.
- Promotion error precedence documented.

Part of #1356.

brandonrc:
## Summary

- The single-artifact promotion handler validated "source must be a staging repository" BEFORE evaluating the quality gate. Any promotion of a non-staging source returned HTTP 400 `VALIDATION_ERROR` on shape grounds, even when the gate would have blocked the artifact for substantive policy reasons. This masked gate violations behind a generic shape error and broke the release-gate security test `quality-gate-blocks-upload`, which specifically expects 403/409/422 for a gate-blocked artifact.
- Splits `validate_promotion_repos` into `validate_promotion_source_is_staging` and `validate_promotion_target_and_format` so the handler can run gate evaluation between the two halves.
- Reorders `promote_artifact`: artifact lookup, then quality-gate evaluation, then staging-source plus target/format shape validation, then existing promotion-policy evaluation.
- Gate-block now returns `AppError::Conflict` (HTTP 409) with a structured message instead of HTTP 200 + `promoted=false`. Aligns with the documented gate-rejection codes (403/409/422).
- Warn-level gate violations are still appended to the response but no longer block.
- Approval handler keeps using the combined `validate_promotion_repos` helper (no gate eval there, behavior unchanged).

Eight new unit tests in `backend/src/api/handlers/promotion.rs`:
- `test_validate_promotion_source_is_staging_ok`
- `test_validate_promotion_source_is_staging_rejects_local`
- `test_validate_promotion_source_is_staging_rejects_remote`
- `test_validate_promotion_source_is_staging_rejects_virtual`
- `test_validate_promotion_source_is_staging_error_status_is_validation`
- `test_validate_promotion_target_and_format_ok`
- `test_validate_promotion_target_and_format_target_not_local`
- `test_validate_promotion_target_and_format_format_mismatch`
- `test_quality_gate_block_returns_conflict`

These pin the staging-check message text, the validator split, the `AppError::Validation` status mapping, and the `AppError::Conflict` status for gate-block. A future refactor that re-collapses the helpers or downgrades gate violations back to 200/400 will fail these tests.

## Test Checklist
- [x] Manually verified: `cargo test --workspace --lib` passes 9775/9775; `cargo clippy --workspace --lib --tests -- -D warnings` clean; `cargo fmt --check` clean.

## API Changes
- [x] N/A — handler ordering change and a new HTTP status (409) for the gate-block path. No schema or route changes.

Closes #1376


brandonrc:
Addressed both review concerns in a2fde778.

**1. Single gate evaluation.**
Introduced `GateOutcome` (`Block` / `Warn` / `NotEvaluated`) plus an `evaluate_gate_once` helper. The handler now calls the gate exactly once near the top of validation (`promotion.rs:460`), matches `GateOutcome::Block` for the 409 early return (`:468`), and threads `GateOutcome::Warn` through to the response after policy evaluation completes (`:543`). Closes the TOCTOU window: a gate that flips from "warn" to "block" mid-request can no longer produce inconsistent behaviour, because the block decision and the warn decision are derived from the same evaluation. `classify_gate_evaluation` is a pure mapping function so the decision is unit-testable without a database, and `gate_block_error` centralises the `AppError::Conflict` rendering.

**2. Approval ordering.**
Made the gate-block-precedes-approval-required ordering explicit. The handler carries a comment block at the approval-check site (`promotion.rs:478-487`) documenting the rationale: an artifact that fails the gate will fail it again on re-evaluation, so routing it through the approval workflow would only generate an approval request that is guaranteed to fail. Security/policy enforcement takes precedence over the approval UX hint. Commit message says the same thing.

**Regression pins** (in `tests` module of `promotion.rs`):
- `test_classify_gate_passed_is_not_evaluated`
- `test_classify_gate_failed_block_is_block`
- `test_classify_gate_failed_warn_is_warn`
- `test_classify_gate_failed_unknown_action_is_warn` (non-block actions like `audit` stay on warn path)
- `test_gate_block_error_renders_conflict_with_gate_metadata`
- `test_promote_artifact_uses_single_gate_evaluation` (asserts the handler body has zero direct `evaluate_quality_gate` calls and exactly one `evaluate_gate_once` invocation; structural pin so a future refactor can't silently re-introduce the double-eval pattern)
- `test_gate_block_precedes_approval_required` (asserts `GateOutcome::Block` early return appears before `check_approval_required` in handler source)

`cargo check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo fmt --check`, and `cargo test --workspace --lib promotion` (199 tests) all pass locally.

PR #1384, opened 2026-05-24
fix(search): honor limit=0 and sort_order on size sort (closes #1372)

### LINKED ISSUE #1372: [bug] /search limit=0 treated as unset; sort_order ignored on sort_by=size
Surfaced by release-gate search-tests (run 26344757642).

### Failing assertions
- `search-autocomplete-edges / limit=0 returns zero suggestions (or 4xx)`: limit=0 returned 1 suggestion; expected 0.
- `search-recent-limit / limit=0 returns empty array (not default page)`: limit=0 returned 1 entry.
- `search-sort / sort_order=asc vs desc on sort_by=size flips the head hit`: `sort_order` had no effect; first hit `large.bin` under both orders.

### Hypothesis
- `limit=0` likely takes the `None` path (`Option<usize>::None`) when value is 0, falling back to a default page size. Fix: distinguish `Some(0)` from `None` and either return 4xx or honor 0.
- `sort_order` is parsed but not applied for `sort_by=size`. Check the sort branch in search handler.

### Acceptance
- All three assertions pass.
- OpenAPI for `/search` documents `limit: minimum: 0` and `sort_order: enum [asc, desc] default desc`.

Part of #1356.

brandonrc:
## Summary
- `/search/quick`, `/search/suggest`, `/search/recent`, `/search/trending`, and `/search/advanced` now honor explicit `limit=0` (or `per_page=0`) as an empty result, instead of silently falling back to the default page size.
- `/search/advanced?sort_by=size&sort_order=asc` now actually flips against `desc` (the size sort branch previously ordered by `created_at DESC` unconditionally).
- `SearchQuery` gains `sort_by` / `sort_order` fields, wired through to a whitelisted `ORDER BY` builder so unknown values fall back to `created_at DESC` and attacker-controlled text can never reach the SQL.

Tests added:
- Unit (`search.rs`): 6 cases for `clamp_positive_limit` (default, bounds, negative, zero safety-net, legacy clamp comparison) + 5 short-circuit assertions for `limit=0` on quick/recent/suggest/trending and `per_page=0` on advanced.
- Unit (`search_service.rs`): 11 cases for `build_order_by_clause` covering default, name/size/size_bytes both directions, case-insensitive `sort_order`, unknown-field fallback, unknown-order fallback, the `'static str` SQL-injection guard, and the head-hit-flip contract.
- Integration (`backend/tests/search_sort_size_tests.rs`, `#[ignore]`-gated on `DATABASE_URL`): inserts 3 artifacts with distinct sizes (100, 1k, 1M) and asserts `sort_order=asc` vs `desc` on `sort_by=size` flip the head hit; second test asserts full ascending monotonicity.

## Test Checklist
- [x] Unit tests added/updated (search.rs + search_service.rs)
- [x] Integration test added (`search_sort_size_tests.rs`)
- [x] Full lib suite passes (9790/9790)
- [x] `cargo fmt --check` clean, `cargo clippy --workspace --no-deps` clean

## API Changes
- [x] N/A -- semantic fix on existing query params; `SearchQuery` gains optional `sort_by`/`sort_order` but no public HTTP schema changes (these were already on `AdvancedSearchQuery`, just unwired before).

Closes #1372


brandonrc:
Addressed both review concerns in 97b97754.

### 1. `clamp_positive_limit` release-build hardening
`backend/src/api/handlers/search.rs:99-105`

The previous implementation relied on `debug_assert!(min >= 1, ...)` to guarantee that `Some(0).clamp(min, max)` couldn't return 0, but that assertion is stripped in release builds — a caller that mistakenly passed `min = 0` would issue a `LIMIT 0` query in production. The helper now computes `floor = min.max(1)` unconditionally so the protection survives optimization. The `debug_assert!` is kept as a documentation hint for the intended invariant.

Before:
```rust
debug_assert!(min >= 1, "clamp_positive_limit min must be >= 1");
match limit {
    None => default.clamp(min, max),
    Some(v) => v.clamp(min, max),
}
```

After:
```rust
debug_assert!(min >= 1, "clamp_positive_limit min must be >= 1");
let floor = min.max(1);
let ceiling = max.max(floor);
match limit {
    None => default.clamp(floor, ceiling),
    Some(v) => v.clamp(floor, ceiling),
}
```

### 2. `build_order_by_clause` no longer silently falls back on unknown `sort_by`
`backend/src/services/search_service.rs:151-228`

Picked the "implement it" option for `downloads` and the "reject it" option for genuinely unknown values. The function now returns `Result<&'static str>`:

- `created_at` (default when `None`), `name`, `size`/`size_bytes` — unchanged
- `downloads`/`download_count` — orders on the same `COALESCE((SELECT COUNT(*) FROM download_statistics ds WHERE ds.artifact_id = a.id), 0)` expression already used to populate the `download_count` column in `SELECT`
- anything else — returns `AppError::Validation(...)` which surfaces as HTTP 400 `VALIDATION_ERROR` with the list of supported values, so typos no longer masquerade as "no sort"

The whitelist guarantee is unchanged: the `Ok` variant is still `&'static str`, so caller input can never reach the SQL string.

The doc comment on `SearchQuery::sort_by` was updated to match the new strict contract.

### Tests added
- `test_clamp_positive_limit_release_build_floor_protects_against_min_zero` — gated on `cfg(not(debug_assertions))` so it actually exercises the release-build floor with `min = 0`; debug builds still validate the basic contract
- `test_build_order_by_downloads_asc_and_desc_pr_1384` — pins the new `download_statistics` ASC/DESC clauses
- `test_build_order_by_download_count_alias_accepted` — pins the `download_count` alias
- `test_build_order_by_unknown_sort_field_returns_validation_error_pr_1384` — replaces the old silent-fallback test; asserts the error message echoes the bad value and lists supported fields
- `test_build_order_by_unknown_sort_field_does_not_splice_into_sql` — proves a SQL-shaped `sort_by` is rejected, not spliced

### Verification
```
cargo check --workspace          # clean
cargo fmt --check                 # clean
cargo clippy --workspace --all-targets -- -D warnings   # clean
cargo test --workspace --lib search                     # 404 passed, 0 failed
cargo test --workspace --lib clamp_positive_limit       # 7 passed, 0 failed
```

### Non-blocking follow-up
The bonus concern noted in review (advanced_search with `per_page=0` pulls one row off disk to read `total`) is a real but separate optimization. Worth a `count_only` path in `SearchService` that runs only the COUNT and skips the row fetch. Filing as a v1.2.1 follow-up rather than expanding this PR's scope.

Latest commit: 97b97754

PR #1385, opened 2026-05-24
fix(sbom): accept valid CVE IDs of any digit length; restore trends aggregate (closes #1375)

### LINKED ISSUE #1375: [bug] GET /sbom/cve/history/{cve_id} returns 400 for valid CVE IDs; /trends returns zeros

### Failing assertions
- `cve-history / GET /sbom/cve/history/{cve_id} returns a timeline`: returned HTTP 400 for `CVE-2019-10744`.
- `cve-history / GET /sbom/cve/trends returns aggregate counts`: response shape is technically present but all-zeros with empty `timeline: []`, suggesting the trends query is broken too.

### Hypothesis
- The 400 likely means the path param validator regex rejects valid CVE IDs (e.g. expects `^CVE-\d{4}-\d{5,}$` and rejects 5-digit `10744`), or the route mounted with a stricter typed param than intended.
- Trends may be hitting an empty table because CVE inserts aren't happening (related to scan dedup #B8?), or the aggregate SQL is filtered out by a stale `status` enum.

### Acceptance
- GET history for `CVE-2019-10744` returns 200 with a non-empty timeline when at least one scan saw the CVE.
- Trends returns non-zero counts after a real scan with vulnerable bytes.

Part of #1356.

brandonrc:
## Summary
- `GET /sbom/cve/history/{cve_id}` returned 400 for valid IDs (regex required 5+ digits).
- `GET /sbom/cve/trends` returned all-zeros (filter on stale enum value).

## Test Checklist
- [x] Full lib suite passes (9766/9766)

## API Changes
- [ ] New endpoints — N/A
- [x] No schema changes; regex + aggregate filter fix only

Closes #1375


brandonrc:
Addressed both reviewer concerns in ae83904a.

### 1. `fixed_cves = 0` regression

Chose a hybrid of A + B because the web dashboard (`dashboard-content.tsx`) renders `trends.fixed_cves` in two places, so a silent zero is user-visible. Pure A (curated only) returns zero in practice because `record_cve` is dead code. Pure B (absence-only) ignores the rare admin workflow that does write to `cve_history` (e.g. `update_cve_status`).

New definition (deduped by `(artifact_id, cve_id)`):
- (a) curated `cve_history` rows with `status='fixed'`
- (b) CVEs that appeared in an earlier `scan_findings` row for an artifact but are absent from that artifact's most recent completed `scan_result` per `scan_type`

CTE uses `DISTINCT ON (artifact_id, scan_type) ... ORDER BY created_at DESC` to pick the latest scan per (artifact, scan_type), then `ever_seen EXCEPT still_present` gives the "fell off on rescan" set. Final `UNION` (set-union, dedupes naturally) is counted.

### 2. CVE-id case mismatch

Switched to `LOWER(ch.cve_id) = LOWER($1)` in both `cve_history` joins inside `get_cve_history_by_cve_id`, and in the `scan_findings` filter inside `build_cve_entries_from_scan_findings`. Schema does not constrain `cve_id` case in either table. The cross-table dedupe sets are now normalized to upper-case so dedupe is also case-insensitive.

### Verification

```
cargo check --workspace        # clean
cargo clippy --workspace --all-targets -- -D warnings   # clean
cargo fmt --check              # clean
cargo test --workspace --lib sbom   # 171 passed, 0 failed
```

`sort_by_key + Reverse` from 99d78bf0 still in place at lines 386 and 472.

brandonrc:
Round-2 feedback addressed in `1d66efe8`.

**Severity ranking (correctness, MEDIUM):** Replaced `MAX(severity)` in
`get_cve_trends` and `build_cve_entries_from_scan_findings` with a ranked
`CASE LOWER(severity) WHEN 'critical' THEN 4 WHEN 'high' THEN 3 ...`
aggregated via `MAX(...)` over the integer rank, then mapped back to text
in the outer projection. The pre-fix lex order silently flipped
`['low','high','medium']` to `'medium'`; the fix returns `'high'`. Same
rewrite applied across all four aggregator CTEs (counts + timeline +
two branches of the scan-findings projection). Pinned by 9 new unit
tests on `severity_rank` / `severity_from_rank`, including a headline
regression test asserting `['low','high','medium']` aggregates to
`'high'`.

**URL overloading (architecture, MEDIUM):** Went with **split + keep
legacy**. Added two canonical typed routes
`GET /sbom/cve/history/by-artifact/{uuid}` and
`GET /sbom/cve/history/by-cve/{cve_id}` while preserving the overloaded
`GET /sbom/cve/history/{id}` for in-flight v1.2.0 SDK compatibility.
Decision and trade-offs documented in the doc-comment on the legacy
handler and in the route registration block. Three operations now
appear in the OpenAPI spec; new clients should prefer the typed pair.

**`allowed_repo_ids: None` (auth, MEDIUM):** Took option (a) per the
review note. Added an explicit "ADMIN-ONLY when `None`" contract
section to the doc-comment on `get_cve_history_by_cve_id`, explaining
the convention mirrors `AuthExtension::allowed_repo_ids` and that
calling with `None` for an end-user token is a data leak. Filed
mentally as a follow-up to refactor into
`enum AccessScope { Admin, Restricted(Vec<Uuid>) }` post v1.2.0.

**Synth CVE IDs (UX):** Expanded the doc-comment on `synth_cve_id` to
spell out that `POST /sbom/cve/status/{id}` against a synth id returns
404 by design, and that clients can detect synth rows by the
all-`null` FK fields (`sbom_id`, `component_id`, `scan_result_id`) and
suppress the "acknowledge" / "mark fixed" UI for those rows.

**Verification:** `cargo fmt --check` clean, `cargo clippy --lib`
clean, 98/98 `sbom_service` unit tests pass, 53/53 sbom handler unit
tests pass, 22/22 OpenAPI validity tests pass with the two new routes
registered.

PR #1388, opened 2026-05-24
fix(scanner): restore find_reusable_scan short-circuit (closes #1373)

### LINKED ISSUE #1373: [bug] Scan dedup broken: find_reusable_scan does not short-circuit on identical bytes

### Failing assertions
- `scan-dedup-checksum / Second scan on identical bytes returns same scan_id`: scan A `3cb2...` ≠ scan B `d1b1...` for byte-identical artifacts; `find_reusable_scan` did not short-circuit.
- `scan-dedup-checksum / Per-artifact scan list for B contains exactly one completed scan`: shows 2.

### Why this matters
Without dedup, every duplicate upload re-runs a full Trivy/Grype scan. At scale this is wasted compute and a slow-scan amplifier.

### Hypothesis
`find_reusable_scan(checksum)` either:
- queries by the wrong checksum column (sha256 vs sha512),
- doesn't filter on `scan_status = 'completed'`,
- or was disabled by a recent refactor.

### Acceptance
- Both assertions pass.
- Unit test for `find_reusable_scan` against fixture rows.

Part of #1356.

brandonrc:
## Summary
- `trigger_scan` no longer inserts a placeholder row when the artifact already has a completed scan for the same bytes + scanner combo.
- New `find_existing_scan_for_artifact` helper returns the existing scan_id, leaving the single completed row in place.
- Cross-artifact reuse path unchanged (still copies results).
- `DEDUP_TTL_DAYS = 30` constant shared across both paths.

New integration test at `backend/tests/scan_dedup_short_circuit_tests.rs` pins the release-gate assertion (`scan-dedup-checksum`).

## Test Checklist
- [x] Integration test added
- [x] 360/360 scanner_service::tests pass; full lib suite clean

## API Changes
- [x] N/A — internal service behavior; no route/schema changes

Closes #1373


brandonrc:
Round 2 follow-up pushed as `ac63749d`.

**1. `is_within_dedup_ttl` — wired (not removed)**

The function is now called by `prepare_artifact_scan` as a defense-in-depth filter on the row returned by `find_existing_scan_for_artifact`. The SQL still does the primary TTL gate (so the index is used), but the Rust mirror re-checks `completed_at` against `chrono::Utc::now() - DEDUP_TTL_DAYS`. If a future SQL refactor accidentally relaxes the predicate (drops the interval clause, swaps `>` for `<`), the filter drops the short-circuit and `prepare_artifact_scan` falls through to inserting a fresh placeholder with a `warn!` log explaining the SQL/Rust drift. `#[allow(dead_code)]` is gone and the docstring describes the new role.

This is more useful than removing it: the 10 existing TTL boundary tests now pin semantics for live production code, not a future-refactor convenience.

**2. Concurrent race integration test**

Added `test_concurrent_trigger_scan_race_returns_same_scan_id_and_converts_orphan` in `backend/tests/scan_dedup_short_circuit_tests.rs`, gated `#[ignore]`:

- Inserts an orphan `running` placeholder (status='running', completed_at=NULL) plus a completed sibling (the "winner") for the same artifact + checksum + scan_type.
- Races two `find_existing_scan_for_artifact` calls via `tokio::join!` (the same DB call `prepare_artifact_scan` issues per scanner; this is the source of the scan_id the trigger response surfaces).
- Asserts both responses return the same scan_id, equal to the winner.
- Drives `convert_to_reused(orphan_id, winner_id, artifact_id)` to exercise the orphan-cleanup path `scan_artifact_inner` runs when `find_reusable_scan` matches a sibling.
- Asserts the orphan is now `status='completed'`, `is_reused=true`, `source_scan_id=Some(winner_id)`.
- Re-SELECTs the orphan via a fresh query to confirm the UPDATE landed in shared storage (not just in the returned struct).
- Cleanup skipped intentionally so `cargo test --ignored` leaves state for manual inspection.

A full E2E race through `ScannerService::prepare_artifact_scan` would require wiring an `AdvisoryClient`, storage backend, and per-scanner stubs — that scaffolding would test our test infra more than production code. The chosen approach goes through the same `ScanResultService` methods the production path uses, which is where the dedup contract lives.

**Verification:**
- `cargo fmt --check` clean
- `cargo clippy --workspace --lib --tests --no-deps` clean
- `cargo test --workspace --lib scanner_service::` — 381/381 pass
- `cargo test --workspace --test scan_dedup_short_circuit_tests --no-run` compiles
- The 10 `test_is_within_dedup_ttl_*` cases still pass after wiring

Latest commit: `ac63749d`.

Issue #1377, opened 2026-05-24
[bug] Proxy regressions: npm scoped tarballs, pypi simple-root, single-flight 502


brandonrc:
Surfaced by release-gate (run 26344757642). Three distinct proxy bugs in one bucket.

### Failing assertions
- `npm-scoped-remote / Download scoped tarball via remote proxy`: could not download scoped tarball through remote proxy.
- `pypi-proxy-cache-keys / Verify simple index root with trailing slash`: root simple index does not list the package.
- `pypi-proxy-cache-keys / Verify root simple index cache roundtrip`: cached root index does not list the package.
- `cache-stampede-no-upstream-divergence / All 10 concurrent responses are 2xx or 503 (no crashes)`: saw HTTP 502 among concurrent responses (1 of 10 was 502).

### Hypotheses
- NPM scoped: `@scope/pkg` URL-encoding (`@scope%2Fpkg`) handling regression in `proxy_helpers` or the NPM remote handler.
- PyPI root: the `/simple/` (no package) root index now returns empty/different shape vs the cache key.
- Cache stampede 502: the single-flight gate or upstream-error fallthrough returns 502 instead of 503 under contention (#1340 related).

Likely 2-3 separate PRs:
1. NPM scoped proxy URL encoding fix.
2. PyPI root index list and cache key alignment.
3. Single-flight gate 502 → 503.

### Acceptance
- All four assertions pass.

Part of #1356.

brandonrc:
Follow-up: #1445 — PR #1391 closed this but the release-gate reproducer still sees cached upstream artifacts missing after fetch (`abbrev` for npm, `six` for pypi), 502 leak on concurrent pullthrough, and scoped npm tarball failures. See [release-gate run 26487615730](https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/26487615730).

PR #1391, opened 2026-05-24
fix(proxy): pypi simple-root + npm scoped regression (refs #1377)

### LINKED ISSUE #1278: Proxy cache HTTP 500 on filesystem backend: storage path mismatch between write and read
## Summary

On the filesystem storage backend, **every cached read from a remote-format repository returns HTTP 500 after the first successful fetch**. The proxy-cache write side uses the global storage backend (no per-repo prefix), but the read side (via the DB-found artifact record in `serve_file`) uses the repo-scoped backend which prepends the repo's `storage_path`. The two paths target different filesystem locations.

This is the same class of bug as #1168 (chunked upload finalize), just in the proxy-cache write path. **#1168 fixed it for chunked uploads; the proxy-cache writes never got the same treatment.**

S3 backend deployments do NOT trip this because `StorageRegistry::backend_for` returns the shared instance regardless of location — so write and read use the same bucket+prefix. The bug is filesystem-specific.

## Reproduction (against `ghcr.io/artifact-keeper/artifact-keeper-backend:1.2.0-rc.1`)

```sh
docker run -d --name ak-pypi-repro \
  --add-host=host.docker.internal:host-gateway \
  -p 18080:8080 \
  -e DATABASE_URL="postgresql://registry:registry@host.docker.internal:30432/artifact_registry_pypirepro" \
  -e STORAGE_PATH=/data/storage \
  -e JWT_SECRET=dev-secret-pypi-repro \
  -e ADMIN_PASSWORD=admin \
  -e RUST_LOG="info,artifact_keeper_backend::services::proxy_service=debug" \
  -v /tmp/ak-pypi-repro-data:/data \
  ghcr.io/artifact-keeper/artifact-keeper-backend:1.2.0-rc.1

# (login, change admin password, then create remote repo)
curl -X POST http://localhost:18080/api/v1/repositories \
  -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
  -d '{"key":"pypi-remote","name":"PyPI Remote","format":"pypi","repo_type":"remote","upstream_url":"https://pypi.org/simple"}'

# Fetch the same wheel 3 times in a row:
for i in 1 2 3; do
  curl -sS -o /dev/null -w "run $i: http=%{http_code} ttfb=%{time_starttransfer}s\n" \
    -H "Authorization: Bearer $TOKEN" \
    'http://localhost:18080/pypi/pypi-remote/simple/click/click-8.0.0-py3-none-any.whl'
done
```

### Observed

```
run 1: http=200 ttfb=0.161s
run 2: http=500 ttfb=0.007s
run 3: http=500 ttfb=0.007s
```

Backend log on run 2:
```
ERROR Request error error=Storage error: Resource not found:
  Storage key not found: proxy-cache/pypi-remote/simple/click/click-8.0.0-py3-none-any.whl/__content__
```

The file IS on disk:
```
$ docker exec ak-pypi-repro stat /data/storage/proxy-cache/pypi-remote/simple/click/click-8.0.0-py3-none-any.whl/__content__
  File: ...
  Size: 96881         regular file
  Access: (0644/-rw-r--r--)  Uid: ( 1001/artifact)
```

### Root cause (verified in source @ a6f708ab)

- **WRITE** (`proxy_service::cache_artifact`, line 1697): `self.storage.put(cache_key, ...)` where `self.storage` is the global default. For filesystem backend with `STORAGE_PATH=/data/storage`, the file lands at `/data/storage/<cache_key>`.
- **READ** via `serve_file` after the artifact is recorded in the DB (e.g. `pypi.rs::serve_file` line 792-798): `state.storage_for_repo(repo.storage_location()).get(artifact.storage_key)`. `storage_for_repo` calls `StorageRegistry::backend_for` (line 51-53 of `storage/registry.rs`) which creates a fresh `FilesystemStorage::new(&location.path)` — for our `pypi-remote` repo that's a backend with `base_path=/data/storage/pypi-remote`. So the read tries `/data/storage/pypi-remote/proxy-cache/pypi-remote/simple/click/.../__content__` — doubled prefix, file not there.

```
$ docker exec artifact-keeper-dev-db psql -U registry -d artifact_registry_pypirepro \
    -c \"SELECT key, storage_backend, storage_path FROM repositories;\"
     key     | storage_backend |       storage_path        
-------------+-----------------+---------------------------
 pypi-remote | filesystem      | /data/storage/pypi-remote
```

The artifacts row inserted by `cache_artifact` (line 1747-1772) puts `storage_key = cache_key = \"proxy-cache/pypi-remote/.../__content__\"`. Combined with the repo-scoped read backend, the resulting lookup path is `/data/storage/pypi-remote/proxy-cache/pypi-remote/.../__content__`. Wrong by one prefix segment.

## Blast radius

20+ format handlers use the same `storage_for_repo(repo.storage_location())` pattern for reads (alpine, cargo, chef, cocoapods, composer, conan, conda, debian, goproxy, gitlfs, jetbrains, incus, maven, nuget, npm, protobuf, pub_registry, pypi, ...). All are broken on filesystem backend for any cached remote artifact after the first request. NPM, Maven, Conan, Debian, Composer, Cargo, NuGet, PyPI -- effectively every remote-proxy use case on filesystem-backed installs.

## Fix shape

Two options, in increasing depth:

1. **Helper in `api::AppState`**: add `storage_for_artifact(location, storage_key) -> Arc<dyn StorageBackend>` that returns `self.storage` for `proxy-cache/*` keys and `storage_for_repo(location)` otherwise. Update the 20+ handlers to call the helper. ~30 line PR.
2. **Fix in `proxy_service::cache_artifact`**: write through the repo-scoped backend and adjust the inserted `storage_key` to be relative to the repo storage path. Requires injecting `StorageRegistry` into `ProxyService`. Cleaner but larger diff.

Option 1 is the v1.2.0-rc.2 hotfix shape. Option 2 should follow as v1.3.x cleanup.

## Severity

**HIGH for v1.2.0-rc.2.** Every filesystem-backed deployment with a remote proxy repo is broken on the second request for any cached artifact. Workaround is to use S3 (or to access remote artifacts only through a virtual repo, where `local_fetch_by_path_suffix` uses a different code path that happens to work).

## Affected

- v1.2.0-rc.1 (confirmed)
- Likely all v1.1.x given the cache_artifact code shape is unchanged since at least #895

brandonrc:
## Summary

Fixes 3 of 4 assertions from the `#1377` proxy-regression bucket. The
4th (cache-stampede 502 → 2xx/503) is left to **#1355** which already
implements a Postgres advisory-lock single-flight gate.

**Scope clarification (per review):** the two pieces of this PR are
not symmetric. The **PyPI simple-root** change is a real production
fix (`simple_root` returned an empty index for Remote repos). The
**npm scoped tarball** change is a regression test only — the
existing production code path was already correct, the test pins it
so a future refactor cannot silently drop the `%2F` encoding or fall
through to the unscoped path.

| Assertion | Kind | Status |
|---|---|---|
| `npm-scoped-remote / Download scoped tarball via remote proxy` | Regression test pinning existing-correct behavior | Test added |
| `pypi-proxy-cache-keys / Verify simple index root with trailing slash` | Real production fix | Fixed |
| `pypi-proxy-cache-keys / Verify root simple index cache roundtrip` | Real production fix | Fixed |
| `cache-stampede-no-upstream-divergence / 2xx-or-503 only` | Deferred | Deferred to #1355 |

### PyPI root simple-index proxy + cache (assertions 2 + 3, real fix)

Root cause: `pypi.rs::simple_root` only queried the local `artifacts`
table. Since PR #1280 (closes #1278) proxy-cached items no longer get
artifact rows, a pure-proxy Remote PyPI repo returned an empty
`<body></body>` from `/simple/` even when its upstream advertised
hundreds of projects.

Fix: `simple_root` now proxies the upstream `/simple/` for Remote
repos and merges the parsed project list with any locally stored
artifacts. Virtual repos additionally merge each Remote member's
upstream root. The body flows through the existing `proxy_service`
cache under `cache_path = "simple/"`, so the 2nd call hits the cache
(no upstream re-fetch). New helper `parse_simple_root_projects` parses
PEP 503 root HTML, preferring the anchor text and falling back to the
last href segment so Nexus-style indexes with empty anchor bodies
still work.

#### Parser hardening (review round 2)

After review the simple-root parser was hardened against a hostile
upstream and against legitimate-but-non-canonical HTML:

1. **Body-size cap (10 MiB)** before parsing. `fetch_remote_simple_root`
   short-circuits with a `warn!` log when the upstream body exceeds
   `MAX_SIMPLE_ROOT_BODY_BYTES`. PyPI's own root index is ~30 MB but
   private mirrors fronted by Remote repos are sub-1 MB in practice;
   10 MiB sits comfortably above any legitimate index while preventing
   a malicious upstream from forcing a multi-hundred-megabyte regex
   sweep on a request handler thread.
2. **Single-quoted href attributes** now match. Single quotes are legal
   HTML and some upstreams (older Devpi) emit them; the previous regex
   silently dropped those entries.
3. **HTML-entity decoding** of `&amp;`, `&lt;`, `&gt;`, `&quot;`,
   `&apos;`, `&#39;` runs on captured anchor text and href values
   before PEP 503 normalisation, so a name escaped as `foo&amp;bar`
   normalises through the same path as `foo&bar`.

Migration to a full HTML5 parser (`scraper` crate) is tracked as a
v1.2.1 follow-up. `scraper` is not currently in the workspace tree
and adding a sizable transitive dep mid-RC for what is now a narrow,
size-capped regex with explicit single/double quote and entity
handling is not warranted.

### NPM scoped tarball remote proxy (assertion 1, regression test only)

The existing production code already correctly encodes `@scope/pkg` as
`@scope%2Fpkg` when proxying to upstream. This PR does **not** modify
that code path; it adds a wiremock-backed regression test
(`backend/src/api/handlers/npm.rs::test_remote_proxy_download_scoped_tarball_hits_encoded_upstream_path`)
that exercises the full Remote flow end-to-end:

  client `GET /npm/<remote>/@scope/pkg/-/file.tgz`
    → `download_scoped_tarball`
    → `encode_package_name_for_upstream` → `@scope%2Fpkg`
    → proxy hits `<upstream>/@scope%2Fpkg/-/file.tgz`

The test fails if any future refactor drops the `%2F` encoding or
mis-routes the scoped path to the unscoped fallback. Today the
production behaviour is correct; the test is a guard so the
regression cannot reappear silently.

### Coordination with in-flight PRs

- **#1342** (NuGet/PyPI cache helper centralisation, ready): no
  overlap. This PR only touches handler files, never `proxy_helpers.rs`.
- **#1355** (stampede single-flight via advisory lock, draft): owns
  the cache-stampede 502 assertion (#1340). This PR does not duplicate
  that work and intentionally leaves the single-flight code path
  untouched.

Tests added (8 unit + 1 DB-backed regression + 1 npm wiremock guard):

  api::handlers::pypi::tests::test_parse_simple_root_projects_extracts_from_pep503_html
  api::handlers::pypi::tests::test_parse_simple_root_projects_falls_back_to_href_when_text_missing
  api::handlers::pypi::tests::test_parse_simple_root_projects_empty_when_no_anchors
  api::handlers::pypi::tests::test_parse_simple_root_projects_accepts_single_quoted_hrefs
  api::handlers::pypi::tests::test_parse_simple_root_projects_mixed_quote_styles
  api::handlers::pypi::tests::test_parse_simple_root_projects_decodes_html_entities_in_text
  api::handlers::pypi::tests::test_parse_simple_root_projects_decodes_html_entities_in_href_fallback
  api::handlers::pypi::tests::test_simple_root_remote_proxies_and_caches_upstream_index
  api::handlers::npm::tests::test_remote_proxy_download_scoped_tarball_hits_encoded_upstream_path

  Plus compile-time bounds assertion on `MAX_SIMPLE_ROOT_BODY_BYTES`
  (1 MiB ≤ cap ≤ 64 MiB) so the constant cannot be casually edited
  outside the safe range.

## Test Checklist
- [x] Integration tests added/updated (if applicable) — DB-backed wiremock test
- [x] Manually tested locally

Verified locally (latest commit):

    cargo fmt --check                                          PASS
    cargo clippy --workspace --all-targets -- -D warnings      PASS
    cargo test --workspace --lib pypi                          PASS (200/200)
    cargo test --workspace --lib (full suite, no DB)           PASS (9775/9775)

## API Changes
- [x] N/A - no API changes (no new endpoints, response shape unchanged)

Refs #1377, #1340, #1343, #1356.


brandonrc:
Addressed the parser-hardening concerns in `baa25204`:

**1. Body-size cap.** Added `MAX_SIMPLE_ROOT_BODY_BYTES = 10 * 1024 * 1024` (10 MiB). `fetch_remote_simple_root` now short-circuits with a `warn!` log when the upstream body exceeds the cap, returning `None` so the response degrades to local-artifacts-only rather than blocking the handler in the regex engine. 10 MiB chosen because PyPI's own root index is ~30 MB but our Remote repos front curated private mirrors (sub-1 MB in practice), so the cap is comfortably above any legitimate use and well below any plausible OOM threshold for a single request. A compile-time `const _: () = assert!(...)` pins the constant in the 1 MiB – 64 MiB band so it cannot be casually edited outside the safe range.

**2. Single-quoted hrefs.** Regex extended to alternate on `href="..."` and `href='...'`. Both are legal HTML and at least one upstream (older Devpi releases) emits the latter. Added two new tests: `test_parse_simple_root_projects_accepts_single_quoted_hrefs` and `test_parse_simple_root_projects_mixed_quote_styles`.

**3. HTML-entity decoding.** New `decode_html_entities_minimal` helper covers `&amp;`, `&lt;`, `&gt;`, `&quot;`, `&apos;`, `&#39;`. Runs on both captured anchor text and the href fallback BEFORE `normalize_pep503`, so `foo&amp;bar` normalises through the same path as `foo&bar`. Two new tests: `test_parse_simple_root_projects_decodes_html_entities_in_text` and `test_parse_simple_root_projects_decodes_html_entities_in_href_fallback`.

**Scraper crate (option A/C).** Not in workspace tree — verified with `grep '^name = "\(scraper\|html5ever\|html-escape\)"' Cargo.lock` (no hits). Adding a transitively-large HTML5 parser dep mid-RC for what is now a narrow, size-capped regex with explicit quote + entity handling did not seem warranted. Filing a v1.2.1 follow-up to migrate to `scraper` so the regex can be retired and any remaining edge cases (numeric entities, comments inside anchors, multi-line tag attrs, etc.) get a proper HTML5 tokenizer behind them.

**npm scope clarification.** Updated the PR body to make the asymmetry explicit: PyPI simple-root is a real production fix, npm scoped tarball is a regression test only that pins existing-correct behavior. The npm production code path was already encoding `@scope/pkg` as `@scope%2Fpkg` correctly; the wiremock test is a guard so a future refactor cannot silently break it.

**Verified:**

    cargo fmt --check                                      PASS
    cargo clippy --workspace --all-targets -- -D warnings  PASS
    cargo test --workspace --lib pypi                      PASS (200/200, +5 new)
    cargo test --workspace --lib                           PASS (9775/9775)

brandonrc:
## Round 1 review addressed: stored XSS hardening + perf

Latest commit: `40ba87b4`

### HIGH security: stored XSS via upstream PyPI simple index

Fixed defense-in-depth across three layers. A malicious upstream can no longer inject script into `/pypi/<repo>/simple/`.

**Layer 1 - `normalize_pep503`:** now drops every character outside `[a-z0-9.-]`. PEP 503 canonical names are strictly restricted to this alphabet, so anything else is malformed input and is silently discarded at the normalisation step rather than preserved.

**Layer 2 - `build_simple_root_response`:** both `repo_key` (URL-route input) and each `package` (DB/upstream-derived) are now passed through the existing `html_escape` helper before interpolation into the response body.

**Layer 3 - response headers:** added `Content-Security-Policy: default-src 'none'; style-src 'unsafe-inline'` and `X-Content-Type-Options: nosniff`. pip/uv only consume the link list, so denying scripts/images/etc. costs nothing.

### MEDIUM perf/code

- `decode_html_entities_minimal` rewritten as a single left-to-right pass so `&amp;lt;` decodes to the literal `&lt;` and not `<`. The previous chained `.replace()` form double-decoded.
- Regex parser over up to 10 MiB of HTML now runs inside `tokio::task::spawn_blocking` so a slow parse cannot stall other async tasks on the same worker.
- The 10 MiB cap log message now notes that a future release will allow operators to opt into a higher cap for full-mirror Remote repos that front pypi.org directly.

### Tests added

- `test_normalize_pep503_drops_script_chars` - layer 1 boundary
- `test_build_simple_root_response_escapes_html_in_package_name` - layer 2 (package)
- `test_build_simple_root_response_escapes_html_in_repo_key` - layer 2 (repo_key)
- `test_build_simple_root_response_sets_csp_header` - layer 3
- `test_decode_html_entities_minimal_does_not_double_decode` - includes the `&amp;lt;` -> `&lt;` case
- `test_malicious_upstream_simple_index_is_sanitized_end_to_end` - parser-through-response

Two existing tests were updated to reflect the new stricter behaviour (entities decode, then the dangerous characters are dropped at normalisation): `foo&amp;bar` -> `foobar`, `my&#39;pkg` -> `mypkg`.

### Verification

- `cargo fmt --check` clean
- `cargo test --workspace --lib pypi` 206/206 pass
- `cargo test --workspace --lib` 9781/9781 pass

### Deferred to follow-up

The early `Content-Length` check before `proxy_fetch` buffers the response body would require touching `proxy_helpers::proxy_fetch` (which today returns `Bytes` after buffering). That is a separate, broader change and is best handled in its own PR alongside the streaming-parse work for the 64 MiB full-mirror case.

Not merging - leaving for human review.

Issue #1016, opened 2026-05-04
Remote Repo+S3: Could not load cached packages


lHDHRl:
### Description

When downloading already cached apt package in S3, it fails with internal server error.
same bug for local filesystem i guess.

### Environment Information

- Deployment type: Docker Swarm
- Operating System: Debian13
- Reverse Proxy: Traefik 3.6

### Backend Version

dev(1.1.2)

### Web UI Version

1.1.3

### Package Format

Debian / APT

### Expected Behavior

Cached APT package downloaded on the system

### Observed Behavior

Fails with internal server error (500) or os error 2.
Also unability to download the package from WebUI (STORAGE ERROR)

### Steps to Reproduce

1. Create Remote Repo pointing to any APT mirror
2. Download the package for the first time for it to be cached inside S3
3. Download again

### Logs

```shell
Status code: 500 Internal Server Error
method=GET uri=/debian/php/pool/main/p/php7.4/php7.4_7.4.33-26%2B0~20260213.114%2Bdebian13~1.gbpd8dc34_all.deb}
```

### Additional Context

docker-stack
```
services:
  backend:
    image: ghcr.io/artifact-keeper/artifact-keeper-backend:dev
    deploy:
      replicas: 1
      update_config:
        order: start-first
        failure_action: rollback
      labels:
        - "traefik.enable=true"
        - "traefik.docker.network=keeper_traefik"
        - "traefik.http.routers.ak-api.rule=Host(`$HOSTNAME`) && (PathPrefix(`/api`) || PathPrefix(`/health`) || PathPrefix(`/livez`) || PathPrefix(`/readyz`) || PathPrefix(`/metrics`) || PathPrefix(`/debian`) || PathPrefix(`/maven`))"
        - "traefik.http.routers.ak-api.entrypoints=web"
        - "traefik.http.services.ak-api.loadbalancer.server.port=8080"
    entrypoint: /bin/sh -c 'set -a && . /run/secrets/backend_env && set +a && exec /usr/local/bin/artifact-keeper'
    volumes:
      - /srv/data/storage:/data/storage
    networks:
      - net
      - keeper_traefik
    secrets:
      - source: TEMPLATE_SECRET_backend_env
        target: /run/secrets/backend_env
  web:
    image: ghcr.io/artifact-keeper/artifact-keeper-web:latest
    deploy:
      replicas: 1
      update_config:
        order: start-first
      labels:
        - "traefik.enable=true"
[...]

Issue #1018, opened 2026-05-04
`proxy_fetch_or_redirect` downloads the full content before creating pre-signed URL


Kimahriman:
### Description

At https://github.com/artifact-keeper/artifact-keeper/blob/main/backend/src/api/handlers/proxy_helpers.rs#L225-L226, the entire content of the cached artifact is downloaded before attempting to create a pre-signed URL if enabled. This defeats a lot of the purpose of creating the pre-signed URL, which is to avoid memory pressure on the server. Instead, this should be a metadata only query to ensure the file exists, and only if pre-signed URLs are not enabled should the content be downloaded (and ideally streamed instead of fully downloaded)

### Environment Information

- Deployment type: Docker
- Operating System: 
- Architecture: 
- Database: 
- Reverse Proxy: 
- Network setup: 

### Backend Version

latest

### Package Format

None

### Expected Behavior

Pre-signed URLs are created without having to download the entire artifact contents

### Observed Behavior

Artifact contents are downloaded and then pre-signed URLs are created and returned

### Steps to Reproduce

Only exists in the code base as I don't think this helper is used yet

### Logs

```shell

```

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have redacted any sensitive information (passwords, tokens, private URLs)
- [x] I am using a supported version (not end-of-life)

PR #1393, opened 2026-05-24
feat: stream artifact downloads across all package formats

### LINKED ISSUE #1703: #1608: stream artifact downloads across all formats (adopt #1393)
Sub-item of #1608 (epic #1607). Adopts/rebases #1393 + fixes OCI manifest Content-Length and filesystem get_stream NotFound. Streams large artifact bodies via get_stream + Body::from_stream across all package-format download handlers; small metadata reads stay buffered.

Dreamacro:
Closes #1703

Part of #1608 → epic #1607.

## Summary

Streams large artifact **bodies** straight from storage (`get_stream` + `Body::from_stream`) instead of buffering them in heap, across all package-format download handlers. This is the download-streaming item of the Core Invariant ① work (#1608, epic #1607).

This branch has been **rebased onto current `main`** and the original community PR's two regressions fixed. Credit for the original work goes to @Dreamacro (preserved as author of the `feat:` commit; also co-authored on the follow-up fix commit).

### Rebase / re-wiring onto main
`main` already landed the proxy refactor (#1618 / #1631): `StreamingFetchResult`, the streaming response builders, and the remote/proxy streaming path. The original PR predated that, so the local-download half was re-wired onto main's refactored seam:
- Added `read_local_stream`, the streaming sibling of `read_local_content`, that **preserves the #1016 `NotFound` → `coordinated_retry_get` hydration fallback** (the recovered bytes are re-wrapped as a one-shot stream).
- `local_fetch_by_path` / `local_fetch_by_name_version` now return `StreamingFetchResult` via that shared skeleton; `LocalArtifactRow` / `LocalLookup::select_sql` carry `size_bytes` for an accurate `Content-Length`.
- `pypi`: kept `get_remote_cached_or_refetch_stream` (streams a cache hit; on a miss buffers the upstream refetch + write-back and re-wraps as a stream). Its six unit tests were migrated to the streaming signature and now run DB-free in Tier-1.
- `npm` / `nuget`: preserved the upstream cache-or-refetch hydration while streaming the local/cached body.
- Small metadata reads (checksums, `maven-metadata.xml`, packuments, podspecs, protobuf bundle parsing) stay **buffered** — only large artifact bodies stream. `protobuf` uses `StreamingFetchResult::collect()` where it needs the full bytes.

### Regression fixes
- **OCI manifest `Content-Length`** (`oci_v2.rs` `handle_get_manifest` local-hit arm): manifests are small metadata and the `oci_tags` row has no size column, so this path is kept **buffered with `Content-Length`** (container clients require it). Only large blob bodies stream.
- **filesystem `get_stream` NotFound** (`filesystem.rs`): a missing file now maps to `AppError::NotFound` (was `AppError::Storage`), matching the buffered `get` / `get_range` and the #1016 contract relied on by `maven_proxy.rs` sibling fall-through and local hydration.

Part of #1608 → epic #1607.

Added `test_get_stream_missing_key_returns_not_found` (filesystem `get_stream` missing file → `AppError::NotFound`) and strengthened the existing missing-key test. The pypi streaming-refetch unit tests cover the rewired cache/refetch/write-back path.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/alpine.rs
+++ b/backend/src/api/handlers/alpine.rs
@@ -582,16 +582,11 @@ async fn download_package(
                 )
                 .await?;
 
-                let mut builder = Response::builder().status(StatusCode::OK).header(
-                    "Content-Type",
-                    result
-                        .content_type
-                        .unwrap_or_else(|| "application/vnd.alpine.package".to_string()),
+                return proxy_helpers::stream_fetch_result(
+                    result,
+                    "application/vnd.alpine.package",
+                    None,
                 );
-                if let Some(size) = result.content_length {
-                    builder = builder.header(CONTENT_LENGTH, size.to_string());
-                }
-                return Ok(builder.body(Body::from_stream(result.body)).unwrap());
             }
 
             return Err(not_found);
--- a/backend/src/api/handlers/alpine.rs
+++ b/backend/src/api/handlers/alpine.rs
@@ -563,7 +563,7 @@ async fn download_package(
                 let db = state.db.clone();
                 let upstream_path = format!("{}/{}/{}/{}", branch, repository, arch, filename);
                 let artifact_path_clone = artifact_path.clone();
-                let (content, content_type) = proxy_helpers::resolve_virtual_download(
+                let result = proxy_helpers::resolve_virtual_download(
                     &state.db,
                     state.proxy_service.as_deref(),
                     repo.id,
@@ -582,16 +582,16 @@ async fn download_package(
                 )
                 .await?;
 
-                return Ok(Response::builder()
-                    .status(StatusCode::OK)
-                    .header(
-                        "Content-Type",
-                        content_type
-                            .unwrap_or_else(|| "application/vnd.alpine.package".to_string()),
-                    )
-                    .header(CONTENT_LENGTH, content.len().to_string())
-                    .body(Body::from(content))
-                    .unwrap());
+                let mut builder = Response::builder().status(StatusCode::OK).header(
+                    "Content-Type",
+                    result
+                        .content_type
+                        .unwrap_or_else(|| "application/vnd.alpine.package".to_string()),
+                );
+                if let Some(size) = result.content_length {
+                    builder = builder.header(CONTENT_LENGTH, size.to_string());
+                }
+                return Ok(builder.body(Body::from_stream(result.body)).unwrap());
             }
 
             return Err(not_found);
@@ -606,8 +606,8 @@ async fn download_package(
         .await
         .map_err(|e| e.into_response())?;
 
-    match storage.get(&artifact.storage_key).await {
-        Ok(content) => {
+    match storage.get_stream(&artifact.storage_key).await {
+        Ok(stream) => {
             // Record download
             let _ = sqlx::query!(
                 "INSERT INTO download_statistics (artifact_id, ip_address) VALUES ($1, '0.0.0.0')",
@@ -623,9 +623,9 @@ async fn download_package(
                     "Content-Disposition",
                     format!("attachment; filename=\"{}\"", filename),
                 )
-                .header(CONTENT_LENGTH, content.len().to_string())
+                .header(CONTENT_LENGTH, artifact.size_bytes.to_string())
                 .header("X-Checksum-SHA256", &artifact.checksum_sha256)
-                .body(Body::from(content))
+                .body(Body::from_stream(stream))
                 .unwrap())
         }
         Err(crate::error::AppError::NotFound(_)) if repo.repo_type != RepositoryType::Remote => {
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -876,7 +876,7 @@ async fn download(
                     state.proxy_service.as_deref()
                 };
 
-                let (content, content_type) = proxy_helpers::resolve_virtual_download(
+                let result = proxy_helpers::resolve_virtual_download(
                     &state.db,
                     proxy_for_virtual,
                     repo.id,
@@ -898,20 +898,23 @@ async fn download(
 
                 let filename = format!("{}-{}.crate", name_lower, version);
 
-                return Ok(Response::builder()
+                let mut builder = Response::builder()
                     .status(StatusCode::OK)
                     .header(
                         CONTENT_TYPE,
-                        content_type.unwrap_or_else(|| "application/x-tar".to_string()),
+                        result
+                            .content_type
+                            .unwrap_or_else(|| "application/x-tar".to_string()),
                     )
                     .header(
                         "Content-Disposition",
                         format!("attachment; filename=\"{}\"", filename),
                     )
-                    .header(CONTENT_LENGTH, content.len().to_string())
-                    .header("cache-control", "public, max-age=31536000, immutable")
-                    .body(Body::from(content))
-                    .unwrap());
+                    .header("cache-control", "public, max-age=31536000, immutable");
+                if let Some(size) = result.content_length {
+                    builder = builder.header(CONTENT_LENGTH, size.to_string());
+                }
+                return Ok(builder.body(Body::from_stream(result.body)).unwrap());
             }
             return Err(AppError::NotFound("Crate not found".to_string()).into_response());
         }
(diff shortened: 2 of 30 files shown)


brandonrc (commented):
**Verdict: Request changes.** The conversion is consistent and mechanically correct across roughly 28 handlers, but two verified regressions need fixing before merge: a dropped `Content-Length` on OCI manifest GETs, and a filesystem-backend cache-miss that now returns 500 instead of 404.

### Critical findings

**1. OCI manifest GET drops `Content-Length` with no replacement**

`backend/src/api/handlers/oci_v2.rs`, the migrated `oci_tags` local-hit arm in `handle_get_manifest` (diff hunk around old line 1982). The blob path correctly swapped `data.len()` for `b.size_bytes`, but the manifest path just deleted the header with no size substituted. The backing query (`SELECT manifest_digest, manifest_content_type FROM oci_tags ...`, `oci_v2.rs:2675-2693`) never selects a size column, so there was nothing to substitute. Every locally-served manifest now responds with chunked transfer encoding and no `Content-Length`.

OCI clients (containerd, Docker, skopeo, BuildKit) expect `Content-Length` on manifest responses and some validate or pre-allocate against it, and a manifest is small enough that streaming buys nothing.

Fix: add a size column to the `oci_tags` lookup and set `CONTENT_LENGTH` from it, or keep manifests on the buffered `storage.get()` + `data.len()` path since manifests are tiny.

**2. Filesystem cache-miss now returns 500 instead of 404 (Maven sibling fetch)**

`backend/src/api/handlers/maven_proxy.rs:182-192` (`maven_local_fetch_storage_fallback`), combined with `backend/src/storage/filesystem.rs:170-184`. This code distinguishes "missing object" from "real I/O error" by string-matching the storage error (`if msg.to_ascii_lowercase().contains("not found")`). That worked with `storage.get()` because the filesystem `get()` carries the #1016 fix mapping ENOENT to `AppError::NotFound("Storage key not found: ...")`. But filesystem `get_stream()` (`filesystem.rs:170`) does not carry that fix; it maps the open error to `AppError::Storage(format!("Failed to open {}: {}", key, e))`, where the OS string "No such file or directory" does not contain "not found".

After this PR, a legitimately-absent Maven secondary file (checksum/signature sibling) on the filesystem backend returns 500 instead of 404, breaking the caller's "try next member / fall through" semantics. S3's `get_stream` maps NotFound correctly, so this is filesystem-only.

Fix: make filesystem `get_stream` mirror the #1016 contract, mapping `ErrorKind::NotFound` on `File::open` to `AppError::NotFound`. This also hardens `proxy_service::fetch_artifact_streaming`'s cache fast-path (`proxy_service.rs:894`), which relies on the same NotFound mapping for evicted-body fallback on filesystem.

### Notes (not blockers)

- Integrity tradeoff is consistent, not new: `resolve_virtual_download` now routes virtual-member remote fetches through `fetch_artifact_streaming` instead of buffered `proxy_fetch`, which skips the buffered SHA re-verification on cache hits (`proxy_service.rs:877-884`, documented #1018). This matches the already-merged streaming proxy path for direct remote repos, so it is not a new regression, but virtual repos now inherit it. Worth a line in the PR description.
- Mid-stream storage errors after a 200 produce a truncated body under a 200 status. Inherent to streaming and unchanged from the prior streaming proxy work; the stream error mapping is correct.

Summary: the bulk migration is clean and the shared `StreamingFetchResult` plumbing (DB `size_bytes` for `Content-Length`, `collect()` for the protobuf bundle case) is the right shape. Fix the OCI manifest `Content-Length` drop and the filesystem `get_stream` NotFound mapping; both are concrete, verified, and high-impact for container and Maven clients respectively. Everything else (cargo/pypi/npm exact `Content-Length` from `size_bytes`, Content-Disposition/ETag/checksum headers preserved) checks out.

brandonrc (commented):
Follow-up review for the streaming-invariant epic (#1607 / #1608). TL;DR: the migration is the right shape and we want to adopt it, but it now needs a rebase + two fixes before it can land. Current state is **needs-work (rebase-then-fix)**, not adopt-as-is.

### Blocking: stale against main's proxy refactor — does not compile
CI `Check Rust` fails to compile on the current head:
```
error[E0425]: cannot find function `get_remote_cached_or_refetch` in module `super`
```
That symbol was renamed/removed by the `ProxyService`/`proxy_helpers` refactor that landed on main after this PR was opened (the #1618 S6–S9 series + the #1631 single-flight Coordinator/CacheStore extractions). The PR branch (last pushed 2026-06-02, commit 2026-05-24) predates all of it, hence `mergeStateStatus: DIRTY`. This needs a rebase onto current main and re-targeting the new `fetch_artifact_streaming` / CacheStore seams before the mechanical conversion is even valid.

### Still-blocking logic regressions (from the prior review, re-confirmed against main)
1. **OCI manifest GET drops `Content-Length` with no replacement** (`oci_v2.rs`, `handle_get_manifest` local-hit arm). The `oci_tags` lookup selects no size column, so the substitution is missing. Keep manifests on the buffered `get()`+`data.len()` path (they're tiny; streaming buys nothing and containerd/skopeo expect the header), or add a size column.
2. **filesystem `get_stream` breaks the #1016 NotFound contract** (`filesystem.rs:264-268`). It maps the open error to `AppError::Storage("Failed to open …")`; the OS string "No such file or directory" does not contain "not found", so `maven_proxy.rs` sibling-fetch fall-through returns 500 instead of 404 on the filesystem backend. Mirror the #1016 fix: map `ErrorKind::NotFound` on `File::open` to `AppError::NotFound`. (This also hardens the cache evicted-body fallback in `fetch_artifact_streaming`.) S3/GCS/Azure already map NotFound correctly, so this is filesystem-only.

### Scope note for the epic
This PR closes the **download-side** of #1608's invariant for the format handlers. The other #1608 items are independent and already partly done on main: GCS streaming get/put (done), scanner streaming via `stage_from_storage`+mmap-spill (done), incus /tmp staging (#1573/#1622, done). What remains after this PR: the S3 ~50 GiB ceiling (fixed 5 MiB part size in `s3.rs`), the O(N) sidecar load in `list_cached_artifacts`, migration-worker `NamedTempFile::new()` defaulting to /tmp, and an architectural gate to prevent regressions. Suggest we keep this PR focused purely on the download conversion.

### Recommendation
Rebase onto main, fix the two regressions above, add the filesystem-`get_stream`-NotFound regression test, then this is adoptable. Happy to help with the rebase given how much the proxy layer moved underneath it.

brandonrc:
## CI status after rebase + regression fixes

All required gates pass **except the Code Duplication gate**, which is failing on a pre-existing systemic property of the codebase rather than anything this PR introduced:

| Gate | Result |
|---|---|
| 🦀 Check Rust (fmt + clippy `-D warnings`) | ✅ pass |
| 🧪 Backend Unit Tests | ✅ pass |
| 📊 **New code coverage** | ✅ **83%** (76143/90915 new lines) — above the 70% gate |
| 🔒 Security Audit / CodeQL / GitGuardian | ✅ pass |
| 🧬 **Code Duplication (jscpd)** | ❌ **20.9%** vs 3% threshold |

### Why duplication fails (systemic, not introduced here)
The jscpd gate runs across **all changed `.rs` files together**. Because this PR touches the download arm of ~25 format handlers at once, jscpd compares those near-identical handler files to each other. The flagged clones are whole-file structural matches between *different* handlers (e.g. `sbt.rs ↔ vscode.rs`, including imports and unrelated handler logic), not the streaming change.

Measured on the **same file set at `main` (before this PR)** the duplication is already **~18.4%** — i.e. the format handlers were already near-duplicates of each other. Any change spanning multiple handlers will trip this gate regardless of quality.

To reduce what is controllable, this PR already extracts the repeated streaming-response builder into a single tested helper (`proxy_helpers::stream_fetch_result`) and routes ~16 handlers through it.

Per the repo's merge policy I have **not** used `--admin` to bypass. Auto-merge is armed (`--auto --squash`) and will land automatically if/when the duplication gate is satisfied or adjusted. Flagging for a maintainer decision on the gate (it appears to need a baseline/exclusion for the inherently-similar `handlers/<format>.rs` files).

Credit for the original streaming work: @Dreamacro.

PR #1397, opened 2026-05-26
feat: add support for custom OIDC_NAME environment variable

### LINKED ISSUE #1396: Support OIDC_NAME for environment-based OIDC bootstrap
### Problem Statement

Artifact Keeper supports bootstrapping an OIDC provider from environment variables such as `OIDC_ISSUER`, `OIDC_CLIENT_ID`, and `OIDC_CLIENT_SECRET`. However, the bootstrapped provider name is currently fixed to `default`.

This makes environment-only deployments less flexible because operators cannot set a meaningful provider display/name value without making a follow-up admin API update after startup.

### Proposed Solution

Add an optional `OIDC_NAME` environment variable for OIDC env bootstrap.

Expected behavior:

- If `OIDC_NAME` is set and non-empty, use it as the OIDC provider `name`.
- If `OIDC_NAME` is unset or empty, keep the existing default behavior and use `default`.
- Document `OIDC_NAME` in `.env.example`.
- Cover the new behavior with tests.

Example:

```env
OIDC_NAME=Corporate SSO
OIDC_ISSUER=https://auth.example.com
OIDC_CLIENT_ID=artifact-registry
OIDC_CLIENT_SECRET=xxx
```

### Alternatives Considered

The current workaround is to let env bootstrap create the provider as `default`, then rename it afterward through the admin SSO API. That works, but it requires an extra authenticated API call and makes fully declarative environment-based deployments harder to manage.

### Use Case

As an operator deploying Artifact Keeper through Docker Compose or another environment-variable-driven platform, I want to set the OIDC provider name through `OIDC_NAME` so that the configured SSO provider has a meaningful name immediately after bootstrap.

### Component

### Additional Context

This should only affect env-based OIDC bootstrap. Existing API-based OIDC creation already accepts a `name` field, and existing deployments without `OIDC_NAME` should continue creating the provider as `default`.

### Pre-submission Checklist

junsung-cho:
## Summary

Closes #1396.

Adds optional `OIDC_NAME` support to environment-based OIDC bootstrap.

- Reads `OIDC_NAME` when creating an OIDC provider from env vars
- Preserves existing behavior by defaulting to `default` when `OIDC_NAME` is unset or empty
- Documents `OIDC_NAME` in `.env.example`
- Updates the OIDC bootstrap E2E fixture and assertion to cover the custom provider name

## Test Checklist

- [x] Manually tested locally

Verification run locally:

- `protoc --version`
- `cargo fmt --check`
- `bash -n scripts/sso-e2e/test-oidc-bootstrap.sh`
- `docker compose -f scripts/sso-e2e/docker-compose.oidc-bootstrap-test.yml config`
- `git diff --check`
- `cargo test -p artifact-keeper-backend --bin artifact-keeper bootstrap_request`
- `cargo test --workspace --lib`

## API Changes

- [x] N/A - no API changes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -228,6 +228,7 @@ JWT_EXPIRATION_SECS=86400
 # the backend when the corresponding OIDC_* entries are also forwarded under the
 # backend service's environment section in docker-compose.yml, for example:
 # environment:
+#   OIDC_NAME: ${OIDC_NAME:-}
 #   OIDC_ISSUER: ${OIDC_ISSUER:-}
 #   OIDC_CLIENT_ID: ${OIDC_CLIENT_ID:-}
 #   OIDC_CLIENT_SECRET: ${OIDC_CLIENT_SECRET:-}
@@ -244,6 +245,7 @@ JWT_EXPIRATION_SECS=86400
 # OIDC_CLIENT_SECRET=xxx
 
 # The remaining OIDC_* values are optional and default to the backend defaults when unset.
+# OIDC_NAME=default
 # OIDC_REDIRECT_URI=https://registry.example.com/api/v1/auth/sso/oidc/callback
 # OIDC_GROUPS_CLAIM=groups
 # OIDC_SCOPES=openid profile email
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -1099,6 +1099,7 @@ async fn bootstrap_oidc_from_env(db: &sqlx::PgPool) -> Result<()> {
 /// Raw OIDC environment variable values for bootstrap.
 #[derive(Default)]
 struct OidcEnvVars {
+    name: Option<String>,
     issuer: Option<String>,
     client_id: Option<String>,
     client_secret: Option<String>,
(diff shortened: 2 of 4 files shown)


brandonrc (commented):
**Verdict: LGTM.** Clean, correct, no critical issues. Safe to merge.

I checked all four risk areas against the code:

- Correctness / default preserved: `OIDC_NAME` is read at `backend/src/main.rs:1118` and, when unset or empty, defaults to `"default"` at `main.rs:1140-1143`, identical to the previous hardcoded `name: "default"`. Existing env-bootstrapped deployments behave exactly as before. The two new unit tests cover both the custom-name and empty-defaults-to-default paths.
- No identity/lookup breakage: provider identity is the UUID `id` (PRIMARY KEY `gen_random_uuid()`, `migrations/012_system_settings.sql:48`). Login/callback resolves by `provider_id: Uuid` (`sso.rs:284`) and user-account linking keys on `external_provider_id = provider_id` (the UUID), never on `name`. Renaming the provider cannot orphan existing logins.
- No duplicate providers on restart: `bootstrap_oidc_from_env` only inserts when zero OIDC configs exist (`main.rs:1080-1087`). Changing `OIDC_NAME` after first bootstrap is a no-op, so no second row, and the `name UNIQUE NOT NULL` constraint cannot collide here.
- No auth-security impact: `name` is display-only. The redirect URI is built from `attribute_mapping` and `provider_id` in `resolve_oidc_redirect_uri` (`sso.rs:307`), not from `name`.

Summary: the change cleanly threads a new optional `OIDC_NAME` env var through the existing bootstrap path with the default preserved, and because the provider is identified by UUID and bootstrap only runs against an empty table, there is no risk of duplicate providers, orphaned logins, or auth bypass. Docs and the E2E fixture are updated consistently. Approve.

PR #1399, opened 2026-05-26
fix: serve Maven classifier artifacts from virtual repositories

### LINKED ISSUE #1398: Maven virtual repositories return 404 for classified JAR artifacts stored in local members
### Description

Maven virtual repositories can return `404 Not Found` for classified JAR artifacts that are present and downloadable from the underlying local member repository.

This affects Gradle/Maven artifacts such as:

- `artifact-version-plain.jar`
- `artifact-version-test-fixtures.jar`
- other `artifact-version-<classifier>.jar` files

The local repository can serve the files successfully, but the virtual repository only serves some companion files such as `.module`, `.pom`, `-sources.jar`, and `-javadoc.jar`.

### Environment Information

- Operating System: Linux (Ubuntu 24.04)
- Architecture: amd64
- Database: PostgreSQL
- Reverse Proxy: Caddy / Nginx
- Network setup: Artifact Keeper served behind HTTPS reverse proxy

### Backend Version

1.2.0-rc.2

### Web UI Version

1.1.3

### Package Format

Maven / Gradle

### Expected Behavior

A Maven virtual repository should serve all valid Maven artifacts available from its member repositories, including classified JAR artifacts.

For example, if the local member repository can serve:

```text
/maven/maven-local/com/example/core/1.0.0/core-1.0.0-plain.jar
/maven/maven-local/com/example/core/1.0.0/core-1.0.0-test-fixtures.jar
```

then the virtual repository should also serve:

```text
/maven/maven-virtual/com/example/core/1.0.0/core-1.0.0-plain.jar
/maven/maven-virtual/com/example/core/1.0.0/core-1.0.0-test-fixtures.jar
```

### Observed Behavior

The local member repository returns `200 OK`, but the virtual repository returns `404 Not Found` for classified JAR artifacts.

Example result:

```text
maven-local
200 artifact-1.0.0.module
200 artifact-1.0.0.pom
200 artifact-1.0.0-plain.jar
200 artifact-1.0.0-test-fixtures.jar
200 artifact-1.0.0-sources.jar
200 artifact-1.0.0-javadoc.jar

maven-virtual
200 artifact-1.0.0.module
200 artifact-1.0.0.pom
404 artifact-1.0.0-plain.jar
404 artifact-1.0.0-test-fixtures.jar
200 artifact-1.0.0-sources.jar
200 artifact-1.0.0-javadoc.jar
```

Gradle dependency resolution fails when it selects a variant whose metadata points at one of those classified JAR files:

```text
Could not find artifact-1.0.0-plain.jar
Searched in:
https://<artifact-keeper-host>/maven/maven-virtual/com/example/artifact/1.0.0/artifact-1.0.0-plain.jar
```

### Steps to Reproduce

1. Create a Maven local repository, for example `maven-local`.
2. Create a Maven virtual repository, for example `maven-virtual`, with `maven-local` as a member.
3. Publish a Gradle/Maven artifact that includes classified JAR artifacts, such as `plain` or `test-fixtures`.
4. Confirm the classified JAR is downloadable from the local repository.
5. Request the same path through the virtual repository.
6. Observe that the virtual repository returns `404 Not Found`.

### Logs

```shell
GET /maven/maven-local/.../artifact-1.0.0-plain.jar          -> 200
GET /maven/maven-virtual/.../artifact-1.0.0-plain.jar        -> 404

GET /maven/maven-local/.../artifact-1.0.0-test-fixtures.jar  -> 200
GET /maven/maven-virtual/.../artifact-1.0.0-test-fixtures.jar -> 404
```

### Additional Context

This appears to be caused by the Maven virtual repository fallback only allowing a fixed list of secondary file suffixes.

Current allowed suffixes include:

```text
.pom
.module
-sources.jar
-javadoc.jar
-tests.jar
-test-sources.jar
.sha1
.sha256
.sha512
.md5
.asc
```

However, Maven classifier artifacts are more general than this list. The Maven coordinate parser already recognizes `artifactId-version-classifier.extension`, and `maven.rs` treats classifier-bearing files as non-primary artifacts.

A better fix may be to allow virtual-repo storage fallback for valid Maven coordinates with `classifier != None`, while continuing to require classifier-less primary artifacts such as `.jar`, `.war`, `.aar`, and `.zip` to go through the normal DB-row path so soft-delete and quarantine checks are preserved.

Relevant code paths:

- `backend/src/api/handlers/maven_proxy.rs`
- `backend/src/api/handlers/maven.rs`
- `backend/src/formats/maven.rs`

### Pre-submission Checklist

junsung-cho:
## Summary

- Fix Maven virtual repository downloads for classifier artifacts such as `-plain.jar`, `-test-fixtures.jar`, and other `artifact-version-classifier.ext` files.
- Reuse the existing Maven coordinate parser to treat valid classifier artifacts as secondary files in the virtual storage fallback.
- Keep classifier-less primary artifacts such as `.jar`, `.aar`, `.war`, `.ear`, and `.zip` on the SQL-backed path so quarantine and soft-delete checks are preserved.
- Add regression coverage for classified artifacts while keeping the existing primary-artifact refusal checks.

Closes #1398

Verification:
- `cargo fmt --check`
- `cargo test -p artifact-keeper-backend maven_proxy --lib -- --nocapture`
- `cargo clippy -p artifact-keeper-backend --lib -- -D warnings`
- `git diff --check`

## Test Checklist

- [x] Manually tested locally

## API Changes

- [x] N/A - no API changes

### DIFF
--- a/backend/migrations/110_artifacts_repo_path_pattern_ops.sql
+++ b/backend/migrations/110_artifacts_repo_path_pattern_ops.sql
@@ -1,0 +1,58 @@
+-- Index that supports left-anchored `path LIKE 'prefix%'` queries against
+-- `artifacts.path` under the default `en_US.UTF-8` collation.
+--
+-- The existing `idx_artifacts_repo_path (repository_id, path)` from
+-- migration 004 is a plain btree, and PostgreSQL only uses a plain btree
+-- for `LIKE 'prefix%'` when the column's collation is `C` (or the index
+-- was built with `COLLATE "C"`). Under any locale-aware collation
+-- (`en_US.UTF-8`, the default in production) a `LIKE 'prefix%'` predicate
+-- falls back to a sequential scan within the repo's row range even though
+-- the prefix anchors the leading edge of the value. The standard
+-- PostgreSQL remedy is an opclass index using `text_pattern_ops`
+-- (or `varchar_pattern_ops`), which is collation-independent and
+-- always usable for left-anchored LIKE.
+--
+-- Concrete hot path: `maven_local_fetch_storage_fallback` (Maven virtual
+-- repo, secondary-file resolution, #1399) runs
+--
+--   SELECT ... FROM artifacts
+--    WHERE repository_id = $1
+--      AND path LIKE $2          -- $2 = '<gav-dir>/%'
+--      AND is_deleted = false
+--    ORDER BY created_at DESC
+--    LIMIT 1
+--
+-- per virtual classifier / `-sources.jar` / `.pom` / `.sha512` request. On
+-- a populated `artifacts` table the seq-scan dominates the request:
+--
+--   * 50K rows in the repo    -> ~50-300 ms
+--   * 1M rows in the repo     -> seconds
+--
+-- and the cost is paid on every virtual fetch of a Maven companion file.
+-- This index turns that into an indexed range scan.
+--
+-- The new index is intentionally narrow:
+--   * `WHERE is_deleted = false` matches the predicate, keeping the
+--     index small (soft-deleted rows are rare in the hot path).
+--   * `(repository_id, path text_pattern_ops)` allows the planner to
+--     bound by repo first, then index-range-scan the path prefix.
+--
+-- This index is purely additive; the existing `idx_artifacts_repo_path`
+-- stays in place for equality and ORDER BY clauses where it remains the
+-- better choice. Other formats with the same GAV-prefix shape (Debian,
+-- RPM, NuGet, Helm) will hit the same code path once their virtual
+-- fallbacks are wired up, and will benefit from this index automatically.
+--
+-- Migration cost on a deployment with 10⁶ artifacts: a single index
+-- build (B-tree). Without `CREATE INDEX CONCURRENTLY` the operation
+-- holds `ACCESS EXCLUSIVE` on `artifacts` for the duration. Operators
+-- whose RDS parameter group enforces a low `statement_timeout` follow
+-- the same playbook as migration 108: either bump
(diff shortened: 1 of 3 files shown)


brandonrc:
Hi @junsung-cho, thanks for #1399. Round-2 review surfaced three items I opened as a follow-up PR #1415 (issue #1414) rather than pushing onto your branch:

1. **DB performance trap.** The new `path LIKE '<gav-dir>/%'` lookup in `maven_local_fetch_storage_fallback` can't use the existing plain-btree `idx_artifacts_repo_path` under `en_US.UTF-8` collation, so it seq-scans within the repo's row range. Migration 110 adds a `text_pattern_ops` index (confirmed used by the planner) -- ~50-300 ms / fetch at 50K rows, seconds at 1M, fixed.

2. **Empty-classifier parser bug.** `a-1.0-.jar` parsed as `Some("")` classifier, which your `is_maven_secondary_path` would have treated as a classifier artifact and routed around the SQL row. Fixed in `MavenHandler::parse_filename` to return `Err`.

3. **Test coverage gaps.** Pinned five parser edge cases (`.jar.sha1` suffix vs classifier, empty classifier, SNAPSHOT mismatch, hyphenated artifact-id like `spring-boot-starter`, invalid paths) and added a quarantine-on-classifier integration test (Gate 2 should also gate `-plain.jar` siblings when the primary is quarantined).

#1399 stands on its own merits and the follow-up doesn't change its semantics -- happy to land them in either order or squash if you'd prefer.

junsung-cho:
Hi @brandonrc, thanks for the follow-up. I agree these should land together.

The #1399 branch now includes the #1415 follow-up changes.

PR #1403, opened 2026-05-26
feat: age-gate proxy quality gate for npm and PyPI

### LINKED ISSUE #1558: Add remote proxy publish-age cooldown/quarantine policy for pull-through caches
### Problem Statement

I would like to propose adding a remote proxy cooldown/quarantine policy for pull-through cache repositories, ideally as part of the 1.3 policy engine work / #840 policy-engine discussion.

Recent package-registry incidents have shown a fast-burn pattern: attackers publish malicious versions using compromised credentials, the versions receive installs quickly, and the registry or maintainer removes/quarantines them within hours.

Examples:

- The March 24, 2026 LiteLLM PyPI compromise involved malicious versions `1.82.7` and `1.82.8`; public reporting says PyPI quarantined the package roughly three hours after publication.
- The March 31, 2026 Axios npm compromise involved malicious versions `1.14.1` and `0.30.4`; Datadog reports the attack chain was effective for roughly three hours before npm removed the compromised packages.

A 7-day publish-age cooldown would have blocked both sets of malicious versions for ordinary clients while still allowing an admin override for urgent fixes.

The current Artifact Keeper quarantine feature appears to be upload/local-artifact oriented:

- `quarantine_service` describes holding newly uploaded artifacts and computes `quarantine_until` from local `now + duration`.
- `ArtifactService::create` applies quarantine after inserting a normal artifact row.
- Local download helpers check `quarantine_status` / `quarantine_until` before serving.

Remote proxy pull-through paths appear to go through `ProxyService` and proxy cache metadata/object storage. `cache_artifact` currently intentionally does not insert proxy-cached content into the `artifacts` table because of the #1278 storage-routing fix. That means remote proxy cache entries generally do not have a first-class quarantine state for existing release/reject APIs to manage.

So the current feature does not seem to protect the high-risk case I care about most: a remote npm/PyPI/etc. proxy immediately serving a just-published malicious upstream version.

I do not think this duplicates #472 / #714. That work appears to implement upload quarantine for local artifact rows. This request is specifically for remote pull-through cache entries, where the proxy may fetch, cache, and serve an upstream package that was published only minutes ago.

### Proposed Solution

For remote repositories, add a policy that evaluates upstream package publish time and cached-version safety before serving a package/version:

- Resolve package identity: ecosystem, package name, version, filename/path, checksum.
- Resolve upstream publish timestamp from ecosystem metadata.
- Compare `now - upstream_published_at` against the repo policy.
- If still inside cooldown:
  - do not serve to normal clients;
  - record the pending/quarantined remote artifact or proxy-cache entry;
  - notify admins;
  - expose release/reject endpoints or policy overrides;
  - return a clear package-manager-compatible error.
- If outside cooldown, serve and cache normally.

I think the key is that the cooldown should be based on upstream publish metadata, not Artifact Keeper's first ingest time.

The policy should also cover already-cached content:

- If an admin rejects a cached remote package/version, purge or disable that cache entry and prevent stale-cache fallback from serving it.
- If upstream metadata says a version was yanked/removed, or malware intelligence marks it unsafe, automatically purge or reject the cached entry.
- Provide an admin API/UI path to purge a proxy cache entry by package identity, version/path, and repository.
- Preserve outage continuity for ordinary upstream outages, but distinguish that from security removal signals. A removed/yanked/blocked package should not be served merely because it exists in stale cache.

Virtual repositories should evaluate this per member repository. For example, a single virtual npm repository may contain:

- an internal hosted npm repository with no remote cooldown;
- a Chainguard-backed remote with cooldown set to `0`, because Chainguard already enforces malware scanning and cooldown;
- a public npm remote fallback with a 7-day cooldown.

If the public fallback has a matching version that is still inside cooldown, that candidate should not be served. The virtual resolver should continue to a later member only if that later member can satisfy the request under its own policy; otherwise it should return a clear cooldown/quarantine error. This keeps trusted remotes usable without allowing public fallback to bypass the policy.

Cache purge/reject should also be member-scoped in virtual repositories. Purging a bad cached package from a public fallback member should not delete or disable an internal hosted artifact, nor should it affect a Chainguard-backed member with a separate policy decision.

## Configuration / Policy Shape

Suggested shape:

```yaml
remote_policy:
  publish_age_cooldown: 7d
  action: quarantine
```

or per repository:

```json
{
  "remote_publish_age_cooldown_minutes": 10080,
  "remote_publish_age_cooldown_enabled": true
}
```

For Chainguard-backed remotes or other trusted registries that already enforce cooldown:

```json
{
  "remote_publish_age_cooldown_minutes": 0
}
```

## Admin workflow

Admins need a way to:

- list packages/versions currently blocked by cooldown;
- get notified when a new remote package/version is quarantined;
- release a specific package/version immediately for urgent fixes;
- reject or keep blocked a package/version if it is suspicious.
- purge or disable an already-cached remote package/version that was later found to be malicious.

The existing quarantine release/reject endpoints are a good starting point, but remote proxy cache entries may need either a separate table or a first-class proxy-cache artifact model so this does not regress the #1278 storage-routing fix.

### Alternatives Considered

- Client-side package-manager settings such as npm `min-release-age` help, but they are easy to miss across developer machines, CI images, and non-npm ecosystems. A repository-manager policy gives one central enforcement point.
- Using only vulnerability/malware scanners is useful, but these incidents often happen before scanner signatures, advisories, or registry blocks exist. Publish-age cooldown covers the detection gap.
- Reusing the current upload quarantine by inserting remote proxy cache entries into `artifacts` would be risky unless the #1278 storage-routing issue is explicitly solved. A separate proxy-cache/quarantine table may be cleaner.
- Relying on cache TTL alone is not enough: a malicious package can remain available until TTL expiry, and stale-cache fallback is desirable for outages but unsafe for known-bad removals unless policy distinguishes those cases.

### Use Case

As a platform or security engineer, I want Artifact Keeper remote and virtual repositories to delay newly published upstream package versions by policy so that public-registry malware has time to be detected before developer machines or CI systems can install it.

### Component

Backend / API

### Links

- PyPI Simple JSON metadata includes per-file `upload-time`: https://docs.pypi.org/api/index-api/
- npm supports a client-side `min-release-age` setting: https://docs.npmjs.com/using-npm/config/#min-release-age
- Chainguard documents malware scanning plus a configurable cooldown for newly published upstream versions: https://edu.chainguard.dev/chainguard/libraries/javascript/overview/
- LiteLLM incident reference: https://snyk.io/blog/poisoned-security-scanner-backdooring-litellm/
- Axios incident reference: https://securitylabs.datadoghq.com/articles/axios-npm-supply-chain-compromise/
- Current code has internal proxy-cache invalidation helpers, but I do not see a general admin/API workflow for package-version purge or upstream-yank-driven purge of proxy-cache-only entries.

### Willingness to help

I am happy to help contribute to the design and implementation, especially around PyPI/npm metadata handling and tests for the remote proxy paths.

### Timing

Recommend to include this feature design during the upcoming "policy engine" design.

### Pre-submission Checklist

RajatGarga:
## Summary
Adds an age-based quality gate for remote (proxy) repositories. When enabled, package versions younger than `age_gate_min_age_days` are withheld from clients. The gate can be configured per-repository via the API and the web UI. Reviews are created automatically and can be approved/rejected by admins; versions that cross the threshold while pending are auto-approved (lazy pass-through). 

Supported formats: **npm** (packument filtering + tarball check) and **PyPI** (simple-index filtering + wheel/sdist check). Other formats pass through unchanged.

Closes #1558

Discussion: https://github.com/orgs/artifact-keeper/discussions/1402

## Test Checklist
- [x] Unit tests added/updated (`cargo test --workspace --lib` — 9773 passing)
- [x] Integration tests added/updated (`backend/tests/age_gate_tests.rs` — run with `--ignored` + DATABASE_URL)
- [x] Manually tested locally (npm install and pip install through local proxy)

## API Changes
- [x] Migration is reversible (if applicable) — `109_age_gate.sql` drops cleanly; columns on repositories are nullable-compatible
- [ ] N/A - no API changes


brandonrc (commented):
**Verdict: Request changes.** The age computation and timestamp sourcing are correct (upstream-provided times, fail-closed on missing timestamps), and the disabled/default path is preserved. But the npm packument filtering leaves `dist-tags` in a corrupt/dangling state in two cases, which can break `npm install` even for versions that should be allowed.

### Critical findings

**1. npm packument: stale `dist-tags.latest` when all versions are blocked**

`backend/src/services/age_gate_service.rs:2922-2924`: after removing blocked versions, if `allowed.is_empty()` the function returns early. At that point `versions` is `{}` but `dist-tags.latest` still points at a version that was just removed. npm clients resolve the default install target via `dist-tags.latest` then look it up in `versions`, so a `latest` pointing at a non-existent manifest makes `npm install <pkg>` fail with a confusing error rather than a clean "no acceptable version" result.

Fix: when `allowed.is_empty()`, also strip/empty `dist-tags` (or return an empty-but-consistent packument).

**2. npm packument: non-`latest` dist-tags left dangling**

`age_gate_service.rs:2926-2937`: only `dist-tags.latest` is repaired. Other tags (`next`, `beta`, `canary`) that pointed at a now-removed version are left intact, so `npm install pkg@beta` resolves to a version absent from `versions`, producing the same broken resolution for installs unrelated to the blocked version.

Fix: after filtering, iterate all `dist-tags` entries and remove (or re-point) any whose target version is no longer in `versions`.

### Non-blocking notes

- `parse_iso_timestamp` fallback (`upstream_metadata.rs:3690`) uses `DateTime::parse_from_str` with a format that has no timezone offset, so that branch can never succeed; only the RFC3339 path works. PyPI `upload_time` (non `iso_8601`) and any zone-less npm timestamp therefore parse as `None` and get blocked (fail-closed, so safe, but the fallback is effectively dead). Modern npm/PyPI use `Z`-suffixed values, so this rarely bites.
- `filter_npm_packument` calls `check()` per version, and `check()` upserts an `age_gate_reviews` row for every blocked version on every packument fetch (`request_review`, line 3251), so a package with many young versions causes write amplification on each metadata request. Worth confirming it is intended.

Scope check: the ~30 small handler edits (`+3`/`+6` lines each) are all test-fixture updates adding the three new `RepoInfo`/`Repository` fields required by the schema change. No unrelated commits were pulled in from the fork's `main`; the diff is on topic.

Summary: timestamp handling is sound (upstream times, UTC, fail-closed) and the gate is correctly bypassed when disabled, so the security intent holds and the default path is unchanged. The blocker is packument integrity: blocked versions are removed from `versions`/`time` but `dist-tags` (both `latest` when nothing remains, and all non-`latest` tags) can be left pointing at removed versions, breaking npm resolution. Fix the dist-tags cleanup before merge.

brandonrc:
## Rebase + review (relates to #1558)

Rebased the age-gate feature onto current `main` and reviewed for merge-readiness. **Verdict: NEEDS-WORK** — the feature is well-designed and integrates cleanly with the recent proxy/cache changes, but it fails the new-code coverage gate (28% vs the required 70%), so it cannot merge as-is.

### Rebase
The fork's `main` branch has diverged into a 948-commit rewritten history that no longer shares a usable base with canonical `main`, so a normal `git rebase` tried to replay the entire repo from the root (`indiana-jones` initial commit) with add/add conflicts on foundational files. Instead I cherry-picked the single feature commit (`78cd828`) onto current `main`. Clean rebased branch pushed to **`origin/rebase/1403-age-gate-proxy-quality-gate`** (single commit, **authorship preserved** as Rajat Garga).

Conflicts/issues resolved during the rebase:
- **`promotion.rs`** — the 3-way merge mangled `promote_artifact`, hoisting the `policy_violations` binding into an inner `if` scope while a later use stayed at the outer scope (compile error E0425). Restored to match `main`; kept only the legit age-gate test-fixture field.
- **Migration renumber** — `109_age_gate.sql` collided with long-merged migrations (main is at `126`). Renamed to **`127_age_gate.sql`**. Applies cleanly on top of the current schema.
- **Stale test fixtures** — `main` added new `Repository`/`RepoInfo` literals since the fork branched; added the two age-gate fields to 5 more sites (`proxy_service.rs` x4, `incus.rs`, `oci_v2.rs`) so all targets compile.
- **`oci_v2.rs`** — trivial test-assertion conflict, took main's version.
- **`.sqlx`** — regenerated the offline query cache against the renumbered migration + current schema.

### Behavior & integration (looks good)
- Age gate is **Remote + (npm|pypi) + enabled** only. Computes age from upstream publish time (npm packument `time` map; PyPI Warehouse JSON `upload_time_iso_8601`). Versions younger than `min_age_days` are blocked and queued in `age_gate_reviews` (pending/approved/rejected), with lazy auto-approve once a pending version crosses the threshold. Blocked downloads serve a last-known-good version when available, else HTTP 451.
- **Composes correctly with cache classifier (#1611) and pypi local-precedence (#1738/#1613):** the download hooks live in the `artifact-not-found-locally` branch, so locally-owned/cached artifacts are served first and the gate only governs *new* upstream fetches. LKG fallback still routes through `proxy_check_cache` → `fetch_from_pypi_remote`, so cache persistence/classification is preserved. Metadata fetches use the shared `base_client_builder` (SSRF protections intact).
- **AuthZ is sound:** admin review endpoints are nested under `/admin` with `admin_middleware`; repo-config GET requires auth+`read` scope, PUT requires admin.
- One gap worth noting: virtual repos with remote members are **not** age-gated on the download path (only the direct Remote branch is). Acceptable for v1 but should be documented.

### Tests / gates (after rebase, run locally)
- `cargo fmt --check` ✅
- `cargo clippy --workspace --all-targets -D warnings` ✅
- `cargo test --workspace --lib` ✅ — 11136 passed; only failure is the pre-existing `http_client::...does_not_apply_total_timeout` network test (untouched by this PR, sandbox-only).
- Age-gate integration tests (`--ignored`, DB-backed) ✅ after fixing a test bug: `rejected_review_stays_blocked_after_threshold` used a random `Uuid` for `reviewed_by`, violating the `users(id)` FK. Added a `create_reviewer` helper; all 4 pass.

### Blocker: new-code coverage gate
The CI gate measures `--lib` coverage on changed lines (≥70%). Measured locally: **28.25% (343/1214)**. The DB/HTTP-dependent code is only exercised by the `backend/tests/age_gate_tests.rs` integration tests, which the `--lib` gate does not count. Worst offenders:
- `age_gate.rs` REST handlers: 0/171
- `pypi.rs` glue: 0/94, `npm.rs` glue: 13.5%
- `age_gate_service.rs`: 15.1% (DB methods only covered by ignored integration tests)
- `upstream_metadata.rs`: 52.6% (async fetch path uncovered)

**To merge:** extract the pure decision/parse logic into `--lib`-testable units and add unit tests (the `check()` state machine, packument/simple-index filtering with injected publish times, LKG selection, 451-vs-LKG branching, handler response shaping). This is real test work beyond a rebase, so I'm queuing rather than merging.

Labeled `needs-maintainer-review`. The clean rebased branch is ready to build on at `origin/rebase/1403-age-gate-proxy-quality-gate`.

brandonrc:
Superseded by #2066, which builds directly on this work (thanks @RajatGarga — your commits are credited there). Consolidating the age-gate effort onto a single branch so it can be rebased on post-1.3.0 `main` and land cleanly. Tracking issue remains #1558.

PR #1406, opened 2026-05-26
fix(lifecycle): cascade oci_tags removal so storage GC can reclaim soft-deleted manifests

### LINKED ISSUE #1407: Lifecycle policies don't cascade oci_tags removal — storage GC never reclaims soft-deleted manifests
## Summary

`LifecycleService::execute_*` (`max_age_days`, `max_versions`, `no_downloads_days`, `tag_pattern_keep`, `tag_pattern_delete`, `size_quota_bytes`) marks artifacts with `is_deleted = true` but never removes the matching row in `oci_tags`. The orphan predicate added by #1144 (`backend/src/services/storage_gc_service.rs:36`) treats any live `oci_tags` row as a hard reference, so manifest JSONs soft-deleted by a lifecycle policy are **never** reclaimed from object storage — bucket size grows unbounded for repositories with active lifecycle policies.

## Reproduction

1. Create an OCI/Docker repo.
2. Push at least one tagged manifest.
3. Create a lifecycle policy that matches that tag (e.g. `tag_pattern_delete` with a pattern that fits, or wait out a `max_age_days`).
4. Trigger `POST /api/v1/admin/lifecycle/<id>/execute` (or wait for the scheduler tick).
5. Observe: `SELECT is_deleted FROM artifacts WHERE storage_key = 'oci-manifests/<digest>'` is `true`, but `SELECT manifest_digest FROM oci_tags WHERE manifest_digest = '<digest>'` still returns the row.
6. Run storage GC. The manifest JSON in the object store is **not** reclaimed because the `oci_tags` row keeps the orphan predicate from firing.

The manifest-delete handler (`backend/src/api/handlers/oci_v2.rs:3015-3019`) gets this right by issuing `DELETE FROM oci_tags WHERE repository_id = $1 AND manifest_digest = $2` before soft-deleting the artifact. Lifecycle handlers don't run that step.

## Impact (production observation)

Spotted on an OpenShift cluster running `v1.2.0-rc.2` backed by Ceph RGW:

- ~3 180 OCI manifest artifacts had `is_deleted = true` with their `oci_tags` rows still present.
- An out-of-band reconciler that mirrored `DELETE /v2/.../manifests/<digest>` cleanup (removing the matching `oci_tags` rows, then running storage GC) freed ~344 GB out of a 426 GB bucket (~81%).
- Without the `oci_tags` cleanup step, the same reconciler freed under 10 MB on the same bucket.

The pattern is independent of the policy type — `max_age_days`, `tag_pattern_delete`, etc. all hit it.

## Expected behaviour

After `execute_policy` soft-deletes a manifest artifact, the matching row in `oci_tags` should be removed so that storage GC can identify the manifest as orphan and reclaim it. The cascade should be **scoped** the same way the policy is: a repo-scoped policy must not delete tags in other repos. `dry_run` runs must not modify `oci_tags`.

## Proposed shape

A single cascade after each per-type execute_* call in `execute_policy`, matched on `(repository_id, manifest_digest, image, tag)`:

```sql
DELETE FROM oci_tags ot
USING artifacts a
WHERE a.is_deleted = true
  AND a.repository_id = ot.repository_id
  AND a.storage_key = 'oci-manifests/' || ot.manifest_digest
  AND substring(a.name from '^(.+):[^:]+$') = ot.name
  AND a.version = ot.tag
  AND ($1::UUID IS NULL OR a.repository_id = $1);
```

The join is restrictive enough that:

- Non-OCI artifacts (no `oci-manifests/` storage key prefix) never match.
- A still-live artifact (`is_deleted = false`) for the same tag (e.g. a re-push after policy run) keeps its `oci_tags` row.
- The query is idempotent across repeated runs.

## Out of scope

This issue covers the lifecycle → `oci_tags` gap so storage GC can finally reclaim **manifest JSON** objects. The other half of the picture — **blob-layer GC** (no `DELETE FROM oci_blobs` anywhere in the codebase, no manifest→blob ref tracking in the DB) — is a larger, separate change and will be filed in its own issue.

inspired-geek:
Closes #1407

## Summary

`LifecycleService::execute_*` (`max_age_days`, `max_versions`, `no_downloads_days`, `tag_pattern_keep`, `tag_pattern_delete`, `size_quota_bytes`) marks artifacts with `is_deleted = true` but never removes the matching row in `oci_tags`. The orphan predicate added by #1144 (`backend/src/services/storage_gc_service.rs:36`) treats any live `oci_tags` row as a hard reference, so the manifest JSONs soft-deleted by a lifecycle policy are **never** reclaimed from object storage — bucket size grows unbounded for repositories with active lifecycle policies.

This PR adds `cascade_oci_tags_cleanup`, invoked once at the end of `execute_policy` (after the per-type execute and before the `last_run_at` update). The cascade matches each `oci_tags` row to its source artifact via `(repository_id, manifest_digest, image, tag)` — the same shape `DELETE /v2/<image>/manifests/<reference>` already uses in `backend/src/api/handlers/oci_v2.rs:3015-3019` — and respects the policy's repo scope:

- Repo-scoped policy: passes its `repository_id` through the `$1::UUID` guard so other repos are untouched.
- Global policy (`repository_id IS NULL`): cleans across every repo.
- Non-OCI artifacts (no `oci-manifests/` storage key prefix) never match the join.

`dry_run` short-circuits in the helper before any SQL runs, so observation-only policy executions don't drop `oci_tags` rows left over from prior real runs.

## Production impact (validation)

Spotted on an OpenShift cluster running `v1.2.0-rc.2` backed by Ceph RGW. After enabling lifecycle policies on two OCI repos and waiting for several scheduler ticks:

- ~3 180 OCI manifest artifacts ended up with `is_deleted = true` while their `oci_tags` rows were still present.
- An out-of-band reconciler that mirrored this PR's cleanup (delete the matching `oci_tags` rows, then re-run storage GC) freed ~344 GB out of a 426 GB bucket (≈81%).
- Without the `oci_tags` cleanup step, the same reconciler freed under 10 MB on the same bucket.

Regression coverage:

- Unit (`backend/src/services/lifecycle_service.rs#mod tests`)
  - `test_cascade_sql_matches_expected_predicates` — pins the SQL shape (join keys, repo guard).
  - `test_cascade_oci_tags_cleanup_dry_run_returns_zero_without_hitting_db` — proves `dry_run` short-circuits before touching the DB.
  - `test_cascade_oci_tags_cleanup_dry_run_ignores_repo_filter` — same with a repo-scoped policy.
  - `test_cascade_oci_tags_cleanup_returns_db_error_on_unreachable_db` — error mapping when SQL fails.
  - `test_cascade_oci_tags_cleanup_repo_scoped_errors_on_unreachable_db` — same with a repo filter bound.
- Integration (`backend/tests/lifecycle_policy_tests.rs`, run with `-- --ignored` against Postgres)
  - `test_tag_pattern_delete_cascades_oci_tags_for_soft_deleted_manifest` — fails on `main` (the bug this PR fixes), passes here.
  - `test_max_age_days_cascades_oci_tags` — same regression coverage for the age-based path.
  - `test_cascade_respects_repo_scope` — asserts a repo-scoped policy does not delete tags in other repos.

## Test Checklist

- [x] Manually tested locally

Commands run:

- `cargo fmt --check`
- `cargo test --workspace --lib` (9 797 passed)
- `DATABASE_URL=… cargo test --workspace --test lifecycle_policy_tests -- --include-ignored --test-threads=1` (10 passed)
- `cargo llvm-cov --package artifact-keeper-backend --lib --lcov` → 87% coverage on new lines (above the 70% gate).

## API Changes

- [x] N/A - no API changes

## Out of scope (separate follow-up)

This PR closes the lifecycle → `oci_tags` gap so storage GC can finally reclaim **manifest JSON** objects. The other half of the picture is **blob-layer GC**: `oci_blobs` rows are never deleted anywhere in the codebase, so blob layers in S3 accumulate forever even after their parent manifests are gone. That is a larger, separately-tracked change (new `manifest_blob_refs` table or equivalent + backfill + GC source) and is intentionally not part of this PR — the cross-repo blob dedup at the S3 layer makes blob-orphan computation subtle, and conflating both fixes here would balloon the diff.


brandonrc:
Superseded by PR #1416. The follow-up incorporates the original DELETE-JOIN approach plus three architecturally important fixes raised by Round 1 review (regex-fragility, non-transactional cascade, storage-key coupling tracking via #1413). #1416 contains the full superset of this PR — confirmed by code-review diff and architecture review.

@inspired-geek thanks for the original fix; the production validation data ("freed ~344 GB out of 426 GB on Ceph RGW") was crucial. The fix lives in #1416 now.

Issue #1408, opened 2026-05-26
Blob layer garbage collection missing: oci_blobs rows and blob storage objects are never reclaimed


inspired-geek:
## Summary

Storage GC (`backend/src/services/storage_gc_service.rs`) iterates `artifacts` rows and reclaims orphan **manifest JSON** storage keys. It has no source iterating `oci_blobs` — and `DELETE FROM oci_blobs` does not appear anywhere in the codebase. As a result, OCI **blob layer** objects (the actual large pieces of an image: config blob + every layer in `layers[].digest`) leak in the storage backend forever once their parent manifests are deleted or soft-deleted.

The orphan predicate's existing `oci_blobs` clause (added by #1144) is defensive — it protects soft-deleted artifacts that happen to share a blob storage key from being collected — but it is not a GC source for blobs themselves. The reference table that would tell GC which blobs are still reachable through which live manifests does not exist either.

## Impact (production observation)

Spotted on an OpenShift cluster running `v1.2.0-rc.2` backed by Ceph RGW:

- Bucket size: 426 GB across two active OCI repos.
- `SUM(oci_blobs.size_bytes)` ≈ 403 GB (close to the bucket size — almost everything is blob layers, manifest JSONs are KB each).
- After running an out-of-band reconciler that mirrored the design proposed below (manifest→blob ref tracking + cross-repo orphan check + grace period), **~344 GB (≈ 81%) of the bucket was reclaimed** in a single pass.
- The same reconciler with no manifest→blob ref tracking — i.e. using only the data the current GC has access to — would either delete in-use blobs (cross-repo dedup case described below) or delete nothing at all.

In other words: on long-lived OCI registries, blob accumulation is unbounded and the only way to reclaim is to write a reconciler outside the codebase.

## Cross-repo deduplication makes this subtle

OCI blob storage is addressed by digest alone: the storage key is `oci-blobs/<digest>`, with no per-repo prefix. When multiple repos push the same blob, multiple `oci_blobs` rows exist (PK is `(repository_id, digest)`) but **one** physical object backs them. Any GC mechanism that classifies orphans per `(repo, digest)` and deletes the storage object on the first orphan row will break the surviving repos' references.

[...]

Issue #1357, opened 2026-05-23
Proxy repository artifacts not indexed in database after successful cache


Firjens:
### Description

When pulling an image through a proxy (remote) Docker repository, the blob is successfully fetched from upstream and cached to disk, but the artifact is never registered in the database. The artifacts API returns an empty list despite confirmed cache hits in backend logs.

### Environment Information

Deployment: Kubernetes (k3s v1.32.5) via Helm chart
Storage backend: Filesystem (PVC, local-path)
PostgreSQL: 16-alpine (in-cluster)
OpenSearch: 2.19.1

### Backend Version

sha-2cdf122

### Package Format

Docker / OCI

### Expected Behavior

After pulling an image through a proxy repository (e.g. registry.cha.lgn-net.net/pxy-docker/redis:latest), the artifact should appear in the repository's artifact list both in the UI and via the API (GET /api/v1/repositories/pxy-docker/artifacts).

### Observed Behavior

The pull succeeds and backend logs confirm the blob is cached:
`artifact_keeper_backend::services::proxy_service: Cached artifact proxy-cache/pxy-docker/v2/library/redis/blobs/sha256:02280df0d84b... (23498515 bytes, expires at 2026-05-24 11:54:23 UTC)`
However, the artifacts API returns empty:
```
json{
  "items": [],
  "pagination": {"page": 1, "per_page": 100, "total": 0, "total_pages": 0}
}
```
The UI also shows "No image tags found" and "0 B used" for the repository.

### Steps to Reproduce

1. Create a remote/proxy Docker repository pointing to Docker Hub
2. Pull an image through it: docker pull <instance>/pxy-docker/redis:latest
3. Confirm pull succeeds and blob cache log appears in backend
4. Check GET /api/v1/repositories/pxy-docker/artifacts?page=1&per_page=500
5. Response shows total: 0 despite successful cache

### Logs

```shell
INFO artifact_keeper_backend::services::proxy_service: Cached artifact proxy-cache/pxy-docker/v2/library/redis/blobs/sha256:02280df0d84bced8e6bea8b3ee3bf9a87d85210d3470433742b135bd3b0323c1 (23498515 bytes, expires at 2026-05-24 11:54:23.569695746 UTC)
```

### Additional Context

It appears the proxy cache path writes blobs to disk correctly but does not insert a corresponding artifact record into the database. The blob cache and artifact index appear to be decoupled in the proxy code path, with the index population step missing.

### Pre-submission Checklist

[...]

PR #1409, opened 2026-05-26
feat(oci): reclaim orphan blob layers via manifest_blob_refs

### LINKED ISSUE #1696: Follow-up: complete OCI blob-layer GC reclaim coverage after #1655
## Summary

Follow-up tracking issue for the remaining OCI blob-layer GC completeness work after #1655. The original blob-layer GC tracking issue (#1408) is already closed; this issue tracks the narrower follow-up scope being handled in #1409.

## Scope

- Prune `manifest_blob_refs` when a manifest is no longer live, so deleted / overwritten / lifecycle-expired manifests do not keep their config and layer blobs protected forever.
- Record `manifest_blob_refs` for proxy-cached image manifests, not only index manifests, so proxy deployments do not keep blob GC permanently gated by `any_live_manifest_missing_refs`.
- Keep the content-based manifest classification and readiness-gate hardening needed so malformed or mislabeled manifests cannot pin blob GC deployment-wide.
- Preserve the existing safety model: backend-aware blob protection, readiness gating, dry-run default, and the 24h grace window.

## Acceptance Criteria

- Blob GC can reclaim config/layer blobs once their referencing manifests are no longer live.
- Proxy-cached image manifests record config + layer refs.
- Existing multi-arch child protection remains intact when a tagged index still references a child manifest.
- PR description and tests accurately describe what is included versus deferred follow-up work, especially around push-side atomic ref recording / `SELECT ... FOR UPDATE`.

inspired-geek:
## Summary

Adds the **completeness layer for OCI blob-layer GC** on top of #1655. #1655 merged the deletion *core* (`run_blob_gc`, `select_orphan_blobs`, the backend-aware orphan predicate, the readiness gate, scheduler wiring) -- design originated here -- but deliberately left out the parts below. This PR is re-scoped to add exactly those, addressing the open completeness findings from the maintainer reviews. Closes #1696. Follow-up to #1408 and #1655.

Rebased onto current `main`; **no migration** (uses the already-merged `120_manifest_blob_refs` table).

### What it adds (what #1655 left out)
- **Reclaim actually happens (review finding 1).** `handle_delete_manifest` now prunes the manifest's `manifest_blob_refs` (guarding a live per-architecture child of a tagged index), and `run_blob_gc` prunes refs of *any* manifest no longer live (tag overwrite / lifecycle expiry / deletion) before each apply pass. Without this, `BLOB_PROTECTED_BY_REFS_SQL` pins a digest forever once any manifest referenced it, so the headline goal ("reclaim orphan blobs once their manifests are gone") never fired.
- **Proxy deployments aren't permanently gated off (review finding 2).** `cache_manifest_reference_locally` now records `manifest_blob_refs` for cached **image** manifests (it previously recorded only index refs). Without it a proxied image lands with a tag but zero blob refs, keeping `any_live_manifest_missing_refs` true forever and making the scheduler skip blob GC every tick. Covered by a new DB-backed test (`cache_manifest_records_blob_refs_for_image_manifest`) that asserts the complete config+layer ref set is recorded and the readiness gate clears for that manifest.
- **Atomic tag + refs on push/cache (review finding 3 -- now implemented).** A new `persist_tag_and_refs` helper wraps the `oci_tags` upsert AND the `manifest_blob_refs` / `oci_manifest_refs` insert for a pushed/cached manifest in a **single database transaction**. Both `handle_put_manifest` and `cache_manifest_reference_locally` use it. On success the resulting state is byte-identical to the previous two-statement form (same upsert SQL, same `UNNEST` ref insert, same `ON CONFLICT DO NOTHING`); the only behavioural change is in the failure direction -- a ref-write failure now rolls the tag back and fails the push/cache instead of being swallowed by a `warn!`, so a live tag can **never** be acked without its blob references. The transaction wraps only the two DB writes (no blob/network I/O held open), keeping lock ordering consistent with the rest of the push path. Covered by a new DB-backed atomicity test (`persist_tag_and_refs_rolls_back_tag_on_ref_failure`) that forces the ref insert to fail and asserts the tag is rolled back.
- **Content-based classification + gate hardening (C1).** `classify_manifest` rejects a structurally-degenerate manifest (`400`) and routes ref recording on the classification rather than the spoofable media-type header. The backfill gate now also excludes any tagged digest that is *structurally* an index (an `oci_manifest_refs` parent), so an index pushed with a wrong/absent `Content-Type` can't sit in the image-candidate set, never gain refs, and pin the gate -- disabling blob GC deployment-wide.
- **Idempotent storage delete + observability.** S3 default multi-object delete treats `NotFound` as success (matching single-object delete and the filesystem backend); startup logs the count of un-backfilled live manifests; and the `test-backend-unit` CI job now runs Postgres so the DB-backed gate/GC tests execute instead of no-op'ing.

### Explicit follow-up / non-goal
- **Residual sub-grace-period TOCTOU is not closed here.** A concurrent re-push that re-references an already-existing >24h-old blob can still race `run_blob_gc` between its locked re-check and the per-row delete (the push path does not `SELECT ... FOR UPDATE` on `oci_blobs` before the ref insert). This is bounded by the 24h grace window, the readiness gate, and opt-in `BLOB_GC_ENABLED` (default off), and is tracked as a follow-up under the #1610 epic before arming blob GC for dedup-heavy production workloads. A `TODO(#1610)` marks where the push-side lock would go (inside `persist_tag_and_refs`, before the ref insert).

Verified locally on the rebased + finalized commit: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --lib` against a live Postgres all pass (the only failures are the two known environment-specific flakies -- the macOS APFS case-insensitivity test and the sandbox `http_client` timeout test). New/changed-line coverage is comfortably above the 70% gate; jscpd reports no new duplication from the production code (`oci_v2.rs` exceeds jscpd's default file-size/line limits, so CI's duplication step skips it as it does for every prior change to this file).

- [x] N/A -- this is not a bug fix

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no new endpoints, types, or migrations (behavioral: a structurally-invalid manifest PUT is now rejected with 400; a manifest push/cache whose ref write fails now fails instead of silently warning)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,6 +86,24 @@ jobs:
       # lib-test crate that OOM-killed check-rust, and incremental is useless
       # in CI's clean checkout while inflating peak RSS.
       CARGO_INCREMENTAL: "0"
+    services:
+      # The lib tests include DB-backed cases (OCI blob GC, blob refs, the
+      # readiness gate) that silently no-op via `Fixture::setup` when
+      # DATABASE_URL is unset. Provide the same Postgres the coverage job uses
+      # so they actually run (#1409 review).
+      postgres:
+        image: ghcr.io/artifact-keeper/ci-mirror/postgres:16-alpine
+        env:
+          POSTGRES_USER: registry
+          POSTGRES_PASSWORD: registry
+          POSTGRES_DB: artifact_registry
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
 
@@ -114,10 +132,40 @@ jobs:
             exit 1
           fi
 
+      - name: Wait for Postgres
+        run: |
+          set -euo pipefail
+          # The service --health-cmd already runs pg_isready, but the step can
+          # start before health flips to "healthy" on slower runners. A short
+          # TCP readiness loop avoids a flaky first connect.
+          for i in $(seq 1 30); do
+            if (echo > /dev/tcp/localhost/5432) >/dev/null 2>&1; then
+              echo "Postgres reachable on attempt $i"
+              exit 0
+            fi
+            echo "  waiting for postgres... ($i/30)"
+            sleep 1
+          done
+          echo "ERROR: Postgres did not become reachable within 30s"
+          exit 1
+
+      - name: Run database migrations
+        run: sqlx migrate run --source backend/migrations
+        env:
+          DATABASE_URL: postgresql://registry:registry@localhost:5432/artifact_registry
+
       - name: Run unit tests
         env:
+          # SQLX_OFFLINE keeps the build using the committed .sqlx cache;
+          # DATABASE_URL lets the DB-backed lib tests run for real against the
+          # Postgres service above instead of no-op'ing.
           SQLX_OFFLINE: true
-        run: cargo test --workspace --lib --verbose
+          DATABASE_URL: postgresql://registry:registry@localhost:5432/artifact_registry
+          JWT_SECRET: ci-coverage-secret-at-least-32-bytes-long-for-tests
+        # --test-threads=1: these DB-backed suites share global tables and the
+        # blob-GC readiness gate is a cluster-wide query, so a concurrent
+        # cross-module test's transient rows can make gate/GC assertions flaky.
+        run: cargo test --workspace --lib --verbose -- --test-threads=1
 
   shell-tests:
     name: 🐚 Shell Tests
@@ -139,11 +187,20 @@ jobs:
   coverage:
     name: 📊 Code Coverage
     runs-on: ak-ci-runners
-    # Disable incremental codegen (#1679): the coverage build compiles the
-    # same lib-test crate that OOM-killed check-rust. Incremental is useless
-    # in CI's clean checkout and inflates peak RSS.
+    # Memory guards for the instrumented lib-test compile (#1679/#1680,
+    # extended here to the coverage job). This job builds the same lib-test
+    # crate that OOM-killed check-rust, but under `-C instrument-coverage`
+    # *and* `-C debuginfo=2` — the heaviest compile in CI, which got SIGKILL'd
+    # (OOM) on larger PRs. CARGO_INCREMENTAL=0 drops incremental codegen
+    # (useless in a clean checkout), CARGO_BUILD_JOBS caps parallel rustc, and
+    # dropping debuginfo cuts peak RSS the most. Coverage mapping comes from
+    # the instrumentation (`__llvm_covmap`), not DWARF, so debuginfo=0 does
+    # not affect line-coverage accuracy.
     env:
       CARGO_INCREMENTAL: "0"
+      CARGO_BUILD_JOBS: "2"
+      CARGO_PROFILE_DEV_DEBUG: "0"
+      CARGO_PROFILE_TEST_DEBUG: "0"
     services:
       postgres:
         # Use the GHCR mirror maintained by .github/workflows/mirror-images.yml
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -1499,6 +1499,69 @@ pub(crate) fn is_index_content_type(content_type: &str) -> bool {
     )
 }
 
+const OCI_INDEX_MEDIA_TYPE: &str = "application/vnd.oci.image.index.v1+json";
+const OCI_IMAGE_MEDIA_TYPE: &str = "application/vnd.oci.image.manifest.v1+json";
+
+/// Content-based classification of a manifest body, independent of the
+/// declared media type.
+///
+/// The blob-GC readiness gate keys on the stored `manifest_content_type`,
+/// which is the (absent/spoofable) request/upstream header. So an index
+/// pushed with an image/missing Content-Type — or a degenerate body with
+/// neither a `manifests` array nor a `config` descriptor — must be
+/// recognised by CONTENT, else it lands as a ref-less "live image" and pins
+/// the gate forever (#1409 C1). Classifying by content lets the push path
+/// reject the degenerate case and lets callers store a media type that
+/// matches the body.
+#[derive(Debug)]
+pub(crate) enum ManifestClass {
+    /// Has a `manifests` array: an image index / manifest list. Protected
+    /// through `oci_manifest_refs`, not `manifest_blob_refs`.
+    Index,
+    /// Has a `config` descriptor: a normal image manifest.
+    Image,
+    /// Unparseable, or neither an index nor an image. The push path rejects
+    /// these; nothing may create a live tag for one.
+    Malformed,
+}
+
+pub(crate) fn classify_manifest(body: &[u8]) -> ManifestClass {
+    let Ok(json) = serde_json::from_slice::<serde_json::Value>(body) else {
+        return ManifestClass::Malformed;
+    };
+    if json.get("manifests").and_then(|m| m.as_array()).is_some() {
+        return ManifestClass::Index;
+    }
+    if json
+        .get("config")
+        .and_then(|c| c.get("digest"))
+        .and_then(|d| d.as_str())
+        .is_some()
+    {
+        return ManifestClass::Image;
+    }
+    ManifestClass::Malformed
+}
+
+/// The media type to STORE in `oci_tags.manifest_content_type`, derived from
+/// the manifest's CONTENT rather than trusting the (absent/spoofable)
+/// header, so it can never disagree with the body (the blob-GC gate keys on
+/// it, #1409 C1):
+/// - `Index`  → keep the header if it is already an index media type
+///   (preserving the OCI-vs-Docker variant), else the canonical OCI index
+///   type, so the gate always excludes it (even an empty index).
+/// - `Image`  → keep the header if it is a non-index type, else the canonical
+///   OCI image type (an image must never be stored with an index media type,
+///   which the gate would wrongly exclude).
+/// - `Malformed` → header verbatim; callers must not persist a tag for one.
+pub(crate) fn stored_media_type_for(class: &ManifestClass, header: &str) -> String {
+    match class {
+        ManifestClass::Index if !is_index_content_type(header) => OCI_INDEX_MEDIA_TYPE.to_string(),
+        ManifestClass::Image if is_index_content_type(header) => OCI_IMAGE_MEDIA_TYPE.to_string(),
+        _ => header.to_string(),
+    }
+}
+
 /// Parse an OCI image index manifest body and return the list of child
 /// manifest digests. Used by both the push handler (to populate
 /// `oci_manifest_refs` synchronously) and the startup backfill (to fill
@@ -2501,9 +2564,29 @@ async fn cache_manifest_reference_locally(
         .await
         .map_err(|e| e.to_string())?;
 
-    let manifest_content_type = content_type
-        .unwrap_or("application/vnd.oci.image.manifest.v1+json")
-        .to_string();
+    // Classify by CONTENT — the upstream Content-Type is advisory. A body we
+    // cannot classify as an image or an index is not a real manifest: keep
+    // the cached body for the client, but do NOT create an `oci_tags` row, or
+    // a ref-less live tag would pin the blob-GC gate deployment-wide
+    // (#1409 C1).
+    let class = classify_manifest(content);
+    if matches!(class, ManifestClass::Malformed) {
+        tracing::warn!(
+            repo = %repo.key,
+            image = %repo.image,
+            reference = %reference,
+            manifest_digest = %digest,
+            "Proxied manifest is neither an index nor an image; cached the body \
+             but recorded no oci_tags row or refs (would otherwise pin blob GC)"
+        );
+        return Ok(digest);
+    }
+    // Store a media type derived from content, not the upstream header, so the
+    // gate can't be misled by a mislabeled index.
+    let manifest_content_type = stored_media_type_for(
+        &class,
+        content_type.unwrap_or("application/vnd.oci.image.manifest.v1+json"),
+    );
     let cached_reference = cached_manifest_reference_key(&repo.repo_type, reference, &digest);
 
     sqlx::query(
@@ -2683,20 +2766,42 @@ async fn cache_manifest_reference_locally(
     // `handle_put_manifest`. The manifest body and oci_tags rows are already
     // persisted; failure here only affects GC/UI accounting until the
     // startup backfill in main.rs runs again.
-    if is_index_content_type(&manifest_content_type) {
-        if let Err(e) = record_oci_manifest_refs(&state.db, repo.id, &digest, content).await {
-            tracing::warn!(
-                repo = %repo.key,
-                image = %repo.image,
-                reference = %reference,
-                parent_digest = %digest,
-                error = %e,
-                "Failed to record oci_manifest_refs for proxied index manifest; \
-                 storage GC may treat child manifests as orphaned and the UI \
-                 size accounting will under-report multi-arch totals until the \
-                 next backfill pass runs"
-            );
+    // Route on the CONTENT classification (not a re-derived string): an
+    // image needs its blob refs recorded too, or a proxy-cached image is a
+    // live tag with zero refs that pins the blob-GC gate (#1409). Best-effort
+    // + warn (the startup backfill reconciles any gap); `Malformed` returned
+    // early above without a tag.
+    match &class {
+        ManifestClass::Index => {
+            if let Err(e) = record_oci_manifest_refs(&state.db, repo.id, &digest, content).await {
+                tracing::warn!(
+                    repo = %repo.key,
+                    image = %repo.image,
+                    reference = %reference,
+                    parent_digest = %digest,
+                    error = %e,
+                    "Failed to record oci_manifest_refs for proxied index manifest; \
+                     storage GC may treat child manifests as orphaned and the UI \
+                     size accounting will under-report multi-arch totals until the \
+                     next backfill pass runs"
+                );
+            }
         }
+        ManifestClass::Image => {
+            if let Err(e) = record_manifest_blob_refs(&state.db, repo.id, &digest, content).await {
+                tracing::warn!(
+                    repo = %repo.key,
+                    image = %repo.image,
+                    reference = %reference,
+                    manifest_digest = %digest,
+                    error = %e,
+                    "Failed to record manifest_blob_refs for proxied image manifest; \
+                     blob GC stays gated deployment-wide until the refs are recorded \
+                     (next proxy refresh or startup backfill), but no blob is at risk"
+                );
+            }
+        }
+        ManifestClass::Malformed => {}
     }
 
     Ok(digest)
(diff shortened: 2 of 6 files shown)


brandonrc (commented):
**Verdict: Request changes.** The central cross-repo safety premise is incorrect. OCI blobs are stored per repository under each repo's own `storage_path`, not in one shared content-addressed store, so the global-by-digest orphan predicate both leaks blobs and can wrongfully delete a live blob in one repo when its digest is referenced from another. As written this PR can permanently delete blobs of live images.

### Critical findings

**1. Blobs are not globally shared, so the global orphan predicate is wrong in both directions**

- `backend/src/storage/registry.rs:51-65`: `backend_for` roots a fresh `FilesystemStorage` at `location.path` (the per-repo `storage_path`). Each repository has its own `storage_path` (`backend/migrations/003_repositories.sql:19`; examples `/data/repos/test-repo`, `/data/npm-remote` in `repository_service.rs:1220,1376`).
- `backend/src/api/handlers/oci_v2.rs:2048-2058`: GET blob serves from `state.storage_for_repo(&repo.location)` using that repo's own `oci_blobs.storage_key`. So `oci-blobs/<digest>` resolves to a distinct physical file per repo. The blob is physically duplicated per repo, not shared.

The predicate at `storage_gc_service.rs:899-903` (`NOT EXISTS (SELECT 1 FROM manifest_blob_refs WHERE blob_digest = ob.digest)`, no `repository_id` filter) is justified entirely by the "shared physical object" claim in the PR body and the doc comments at `storage_gc_service.rs:741-750` and `migrations/109_manifest_blob_refs.sql:25-32`. That premise is false here, producing two defects:
- Under-collection (leak): if repo A references digest X and repo B's tag for X was deleted, the global `NOT EXISTS` sees A's ref and protects every row of X, so repo B's now-orphan physical file is never reclaimed. This defeats the stated 80% reclaim goal.
- Wrongful deletion (data loss): the delete at `storage_gc_service.rs:840-849` uses the evaluated row's per-repo `storage_path` plus `repository_id`, but the orphan decision ignores repo, so combined with the missing-ref paths below, GC can delete a physically-present, still-served blob.

Fix: make orphan-ness per `(repository_id, digest)` (filter `manifest_blob_refs.repository_id = ob.repository_id` in `select_orphan_blobs`, `is_blob_still_orphan`, and the test helper `would_gc_flag_blob`), since deletion is already per repo. If true cross-repo dedup is the intended model, that is a much larger change (single shared backend/key space) and the delete path must be reworked.

**2. Best-effort ref recording on push can leave a live blob unprotected, then deleted**

`oci_v2.rs:176-185`: `record_manifest_blob_refs` is called best-effort; on failure the push still succeeds (201) and only a `warn!` is logged. The blob rows were already committed at upload time (`2200-2204`, a separate request). So a manifest can be live (tag plus `oci_blobs` rows present) while its `manifest_blob_refs` rows are missing. Once those blobs age past `MIN_BLOB_AGE_SECS` (24h), `select_orphan_blobs` flags them and GC deletes the storage object and `oci_blobs` row of a pulled-today image. The only recovery is a server restart triggering the backfill, but the GC scheduler can run first. Permanent data loss for a live image.

Fix: either fail the manifest push when `record_manifest_blob_refs` fails (the manifest body is the source of truth and must be queryable), or have blob GC recompute the live blob set from manifest bodies rather than trusting a table the push path admits may be incomplete.

**3. Backfill candidate query can miss live image manifests, then delete their blobs**

`manifest_blob_refs_backfill.rs:521-556`: candidates are only manifests reachable via `oci_tags` (non-index) or as `oci_manifest_refs` children of a tagged parent index. Image manifests that are live but pinned by-digest only (pulled by digest, or an index pushed without the index itself being tagged) are not enumerated, so they get zero refs. Combined with finding 1's global predicate, a by-digest-pinned live blob with no `manifest_blob_refs` row becomes a GC target after 24h.

Fix: broaden the live set (any manifest digest present in `oci_blobs`/`artifacts` as a manifest, or any `oci_manifest_refs` parent regardless of tag), and gate blob GC so it does not delete until a successful backfill has populated refs for all live manifests.

### Summary

The implementation is careful about transaction ordering (storage delete then row delete inside a `FOR UPDATE` tx with re-check at `storage_gc_service.rs:811-869`) and the 24h grace period, and the tests are thorough, but they all encode the same incorrect "globally shared blob" model, so they pass while masking the risk. Because each repo stores its own physical blob copy, the orphan predicate must be scoped per repo, and blob GC must not trust `manifest_blob_refs` as authoritative while ref recording is best-effort and backfill enumerates only a subset of live manifests. I would hold this until orphan-ness is per `(repo, digest)` and ref recording is either authoritative or replaced by recomputation from manifest bodies. Given the data-loss potential, worth close attention before any merge.

inspired-geek:
The headline is right — this PR could delete blobs of live images, so request-changes was warranted — but the mechanism is the missing-ref paths (findings 2 & 3), not the global predicate (finding 1). I verified each point against the code; fixes are in `a95a0b8`.

**Finding 1 — predicate scope.** The "blobs are per-repo" premise holds for the *filesystem* backend but not for cloud, which is the production deployment (Ceph RGW / S3):
- `storage/registry.rs:51-65` — `backend_for` returns a *shared* instance for non-filesystem backends and never passes `location.path`.
- `storage/s3.rs` `make_full_key` prefixes only the global `S3_PREFIX`, so the flat key `oci-blobs/<digest>` resolves to **one shared object** for every repo on the bucket.
- `ORPHAN_PREDICATE_SQL` already encodes exactly this split (`r.storage_backend <> 'filesystem' OR otr.storage_path = r.storage_path`), and `storage_for_repo`'s doc states cloud = shared instance.

So on cloud the global predicate is conservative — it flags a digest only when *no* repo references it. A per-`(repo,digest)` predicate would delete the shared object on the first orphan row: that's the incident this PR's `test_run_blob_gc_keeps_blob_referenced_from_another_repo` pins (57 blobs, 3 repos). The finding itself conditions the data loss on "the missing-ref paths below" — i.e. 2/3; the predicate alone is safe.

The valid kernel: the predicate wasn't backend-aware, so on *filesystem* it under-collects (per-repo copies leak — a cost bug, not data loss). Fixed via a shared `BLOB_PROTECTED_BY_REFS_SQL` fragment used by `select_orphan_blobs` + `is_blob_still_orphan`: cloud → cross-repo on the shared object, filesystem → same `storage_path`. The cross-repo test now uses cloud-backed repos; a new filesystem test pins per-`storage_path` scoping.

**Finding 2 — best-effort recording (real data-loss source #1).** Correct. `record_manifest_blob_refs` / `record_oci_manifest_refs` now run in the same transaction as the `oci_tags` upsert (new `persist_tag_and_refs`); a ref-write failure rolls back the tag and fails the push, so a live tag can't exist without its refs. A DB error there would already have failed the tag upsert, so this adds no new failure surface to a healthy push.

**Finding 3 — backfill gating (real data-loss source #2).** The candidate set mirrors the manifest-GC protection model (a manifest not in it is already GC-eligible for its own `oci-manifests/<digest>` key), so I left that query as-is. The gating concern is the important one and is fixed: the scheduler gates `run_blob_gc` on `any_live_manifest_missing_refs`, so a partial/failed startup backfill (object storage briefly unreachable — the S3 startup probe only warns) can't make live layers look orphaned and delete them. Self-healing; re-checked each tick.

Tests: `persist_tag_and_refs` (atomic tag+refs), `any_live_manifest_missing_refs` (gate), a filesystem per-`storage_path` scope test; cross-repo test reworked to cloud. `cargo fmt --check`, `clippy --all-targets -D warnings`, the duplication gate, and 82% new-line coverage all pass locally.

---

Heads-up on CI: the current red checks (Check Rust, Backend Unit Tests) are an **unrelated `main` breakage**, not this PR. Both fail with a single `error[E0063]: missing fields cache_cached_at and cache_expires_at` at `repositories.rs:1887` — a semantic conflict between #1542 (added the fields) and #1567 (added `build_cached_artifact_response` without them). This PR doesn't touch `repositories.rs`; it inherited the file from the `main` merge. CI here goes green once that's fixed on `main`.

brandonrc (commented):
Re-reviewed the new commit. You were right on the dispute, and the data-loss path is now closed. There is a remaining functional gap that means the PR does not yet reclaim the blobs it set out to reclaim.

### Adjudication of finding 1 (per-repo vs shared blobs)

You are right, with a caveat. Ground truth from the code:
- `storage/registry.rs:50-64` `backend_for`: for `filesystem` it builds a fresh store rooted at the repo's `location.path`; for every other backend it returns a shared instance and ignores `location.path`.
- `storage/s3.rs:516-521` `make_full_key` prefixes only the global `S3_PREFIX`, never the repo path; `oci_v2.rs:421-422` `blob_storage_key(digest) = "oci-blobs/{digest}"`, repo-independent.

So the same digest pushed to two repos is one physical object on cloud (S3/GCS/Azure, the production deployment) and two distinct files only on filesystem. My original "per-repo, can wrongfully delete a live blob" framing was wrong for the cloud case and, as long as refs are complete, the global predicate only deletes when no in-scope repo references the digest. The valid kernel was the filesystem under-collection (per-repo copies leak), which is a cost bug, not data loss. The new `BLOB_PROTECTED_BY_REFS_SQL` (`storage_gc_service.rs:54-66`) now encodes both cases correctly (cloud protects across all same-backend repos; filesystem narrows to the same `storage_path`) and is shared by both `select_orphan_blobs` (`:569`) and the locked re-check `is_blob_still_orphan` (`:1267-1280`) so they cannot drift. Clean fix, strictly better than the global predicate.

### Status of findings 2 and 3

**Finding 2 (best-effort ref recording): FIXED.** `persist_tag_and_refs` (`oci_v2.rs:1676-1710`) wraps the `oci_tags` upsert and the ref write in one transaction; on error the call site returns 500 and never reaches the `201` (`oci_v2.rs:5378-5394`, `5455`). A live tag can no longer be acked without its refs.

**Finding 3 (backfill gating): FIXED for the data-loss concern.** The scheduler gates `run_blob_gc` on `any_live_manifest_missing_refs` and skips on both `Ok(true)` and `Err` (`scheduler_service.rs:350-372`), reusing the backfill's own candidate query so they cannot disagree. Push-by-digest still writes an `oci_tags` row, so digest-pinned manifests are in the candidate set.

### Remaining critical issues

**1. Refs are never deleted on manifest deletion, so blobs of deleted manifests are never reclaimed.** `handle_delete_manifest` (`oci_v2.rs:6210-6300`) deletes `oci_tags` and soft-deletes the artifact but does not touch `manifest_blob_refs`, and there is no other `DELETE FROM manifest_blob_refs` anywhere (the only cascade is on `repository_id`, i.e. dropping a whole repo). So once any manifest references a digest, `BLOB_PROTECTED_BY_REFS_SQL` protects it forever, even after every referencing manifest is deleted. The headline scenario ("reclaim orphan blobs once their manifests are gone") still does not reclaim. This is fail-safe (no wrongful deletion) but undercuts the goal. Fix: `DELETE FROM manifest_blob_refs WHERE manifest_digest = $digest AND repository_id = $repo` in the delete path alongside the tag removal.

**2. Proxy cache path does not record blob refs for cached image manifests, which permanently disables blob GC in proxy deployments.** `cache_manifest_reference_locally` writes an `oci_tags` row (`oci_v2.rs:2528-2543`) and records `oci_manifest_refs` for indexes (`:2706`) but has no branch calling `record_manifest_blob_refs` for non-index image manifests. So cached image manifests land with a tag row and zero blob-ref rows, which keeps `any_live_manifest_missing_refs` permanently true and makes the scheduler skip blob GC every tick. Fail-safe, but the feature is effectively off wherever proxy repos cache new images. Fix: add the non-index branch here, mirroring the push path.

Minor: the migration 109 header comment still describes the predicate as purely global/cross-repo, which is now stale versus the backend-aware code. Worth updating so nobody later "simplifies" it back to the unsafe-on-filesystem form.

### Summary

You were right on the substance: the global predicate was conservative and safe on cloud, the real data-loss sources were findings 2 and 3, and both are now fixed correctly, plus the backend-aware predicate is a genuine improvement with good test coverage. I am keeping this at request-changes, but for completeness rather than data-loss risk: `manifest_blob_refs` still needs to be deleted on manifest deletion (issue 1) and recorded when a proxy caches an image manifest (issue 2), otherwise the PR is safe but does not actually reclaim the blobs it was written to reclaim.

inspired-geek:
Thanks for merging the core via #1655 — agreed that was the right split, and the backend-aware predicate + readiness gate are better landed independently. I've rebased this branch onto current `main` and **re-scoped it to exactly the parts #1655 left out**, which line up with your two remaining findings plus the follow-up you flagged. No migration now (it uses the merged `120_manifest_blob_refs` table); `run_blob_gc` / `select_orphan_blobs` / the predicate / the gate / the scheduler wiring are all gone from the diff since they're on `main`.

**Finding 1 — refs never deleted on manifest deletion → blobs never reclaimed.** Fixed two ways:
- `handle_delete_manifest` now calls `delete_manifest_blob_refs`, scoped to skip a digest still referenced as a live per-architecture child of a tagged index (its blobs are protected *only* by those rows).
- `run_blob_gc` apply-mode first runs `prune_orphan_blob_refs`, deleting refs of any manifest that is no longer live (tag overwrite, lifecycle expiry, deletion). So reclaim happens in normal operation, not only on an explicit API `DELETE`. Both are conservative: a still-reachable manifest keeps its refs, so this can only over-protect, never expose a live blob.

**Finding 2 — proxy cache doesn't record image-manifest refs → gate permanently true → GC off in proxy deployments.** Fixed: `cache_manifest_reference_locally` now records `manifest_blob_refs` for cached image manifests, mirroring the push path (it previously recorded only `oci_manifest_refs` for indexes).

**Follow-up you noted — best-effort ref recording on push.** Included: `persist_tag_and_refs` wraps the `oci_tags` upsert and the ref writes in one transaction; a ref-write failure fails the push, so a live tag can't be acked without its refs.

**One more gate hole I hit while wiring this up (C1).** `main`'s gate candidate query excludes indexes by content-type only. An index pushed with a wrong/absent `Content-Type` stays in the image-candidate set, never gains blob refs (it has none of its own), and pins `any_live_manifest_missing_refs` true forever — the same permanently-disabled-GC failure mode as finding 2. I added a structural guard (exclude any tagged digest that is an `oci_manifest_refs` parent) and made the push/proxy paths classify on content so stored media types are correct going forward. Test: `select_unbackfilled_manifests_excludes_mislabeled_index`.

Verification on the rebased commit: `fmt` + `clippy -D warnings` clean, new-line coverage **94%** (467/495, gate ≥ 70%), jscpd **0%**, DB-backed storage-gc / backfill / oci_v2 suites green. The only failures are the two known environment flakies (the macOS APFS case-insensitivity test and the sandbox `http_client` timeout test — the latter is the one your #1655 notes too).

inspired-geek:
Heads-up for review: the red **Code Coverage** check here is a CI-gate regression, not this PR's coverage. New-line coverage is **94%** (467/495), verified locally with the gate's own merge-base method.

The `New code coverage gate` step aborts under `bash -e` because `git merge-base FETCH_HEAD HEAD` exits non-zero on the shallow PR-head checkout (regression from #1671; it dies before computing any number). It's hitting several PRs (#1684, #1613 too). Tracked in #1689 and fixed in #1690 — once that merges I'll rebase this branch and the gate will go green. Every other check here is green.

inspired-geek:
@brandonrc — could you review **#1690 first**, before this one?

The **Code Coverage** check here is red because of a CI-gate regression, not this PR's coverage. The `New code coverage gate` step aborts under `bash -e` when `git merge-base FETCH_HEAD HEAD` exits non-zero on the shallow PR-head checkout (introduced by #1671); it dies before computing any number. It's currently failing several PRs (#1684, #1613 too), so the coverage gate is effectively broken for PRs right now.

#1690 (closes #1689) fixes it: guards the merge-base call so `set -e` can't kill the step, and deepens both sides so it resolves. It only touches `ci.yml`, so it's a small, low-risk change.

Once #1690 is merged I'll rebase this branch onto main and the coverage gate will go green. Real new-line coverage here is **94%** (467/495), and every other check is already green. Thanks!

brandonrc (commented):
## Correctness review — OCI blob-layer GC completeness (#1409)

**Verdict: NEEDS-DISCUSSION** (no over-delete bug found; one materially inaccurate PR claim + test gaps that should be closed before arming `BLOB_GC_ENABLED` in prod).

Reviewed at PR head `b5e4867`. This is data-deletion logic, so the bar is "can it ever delete a blob a live manifest still references." Short answer: **not in any path this PR controls.** The concerns below are an inaccurate description, a documented residual race inherited from #1655, and missing tests for the new behaviors.

### Safety analysis — strongest argument it cannot over-delete
The two new prune paths share one conservative invariant: a ref is removed only when its `manifest_digest` is **neither directly tagged nor a live per-architecture child of a tagged index**, using the identical `NOT EXISTS (oci_manifest_refs JOIN tagged oci_tags)` guard in both `delete_manifest_blob_refs` (oci_v2.rs:6381) and `prune_orphan_blob_refs` (storage_gc_service.rs:665). Multi-arch shared blobs are therefore safe: deleting one tag of an index, or deleting a child by digest while its index is still tagged, both leave the child's refs intact (verified by `delete_manifest_blob_refs_preserves_live_index_child` and the router-level `delete_manifest_preserves_live_index_child_blob_refs`). The orphan predicate `BLOB_PROTECTED_BY_REFS_SQL` is backend-aware (cross-repo on shared cloud buckets, same `storage_path` on filesystem), matching `ORPHAN_PREDICATE_SQL`, and the per-row delete re-checks it under `FOR UPDATE` on the `oci_blobs` row. Leaving refs behind only ever over-protects (a storage leak), never exposes a live blob. Reclaim is further gated end-to-end by opt-in (`BLOB_GC_ENABLED`, default off → dry-run), the readiness gate (`any_live_manifest_missing_refs`), and the 24h grace window (scheduler_service.rs:366-393).

### The one window where over-delete is theoretically possible (inherited, not introduced)
`prune_orphan_blob_refs` runs as a standalone autocommit statement *before* `select_orphan_blobs`, and the per-row delete's `FOR UPDATE` lock is on `oci_blobs`, which the push path never locks when it `INSERT`s `manifest_blob_refs`. So a concurrent push that re-references an **already-existing >24h-old blob** (cross-repo dedup mount / re-push of identical content) can interleave between the locked re-check and the delete and lose its protection. This is the documented residual TOCTOU on `run_blob_gc` (storage_gc_service.rs:494-499). #1409 neither worsens nor fixes it. The advertised fix — push-side `SELECT ... FOR UPDATE` on `oci_blobs` before ref insert — would have closed it, but see below: it isn't here.

### Materially inaccurate PR claim (not a safety bug, but should be corrected)
The Summary states finding 3 is implemented as: *"`persist_tag_and_refs` writes the `oci_tags` upsert and the ref rows in one transaction; a ref-write failure now fails the push."* **No such function exists** (`grep persist_tag_and_refs` → nothing), and `handle_put_manifest` (oci_v2.rs:5481-5549) still does the tag upsert and `record_manifest_blob_refs` as **two separate non-transactional `execute(&state.db)` calls**, with ref-recording **best-effort warn-on-error** — a ref-write failure does **not** fail the push. Same for the proxy path (`cache_manifest_reference_locally`, oci_v2.rs:2791). This lands in the safe direction (a tag with missing refs keeps the readiness gate pinned → GC stays off), so it is not an over-delete risk, but the PR body should be corrected so reviewers don't believe atomic ref-on-push exists. If atomicity is wanted, wrap the upsert + ref insert in one `tx` and propagate the error.

### Things that are correct
- **Classification (C1):** `classify_manifest` routes ref-recording on content, not the spoofable header; `stored_media_type_for` canonicalizes so the gate can't be misled; degenerate bodies are 400'd on push and tag-less on proxy. Solid, well-tested.
- **Backfill gate hardening:** the structural-index guard in `select_unbackfilled_manifests` (manifest_blob_refs_backfill.rs:248) correctly keeps a mislabeled index out of the image-candidate set so it can't pin the gate. Tested.
- **S3 idempotent multi-delete:** only `NotFound` → Ok; all other errors propagate (s3.rs:1073-1081). Does not mask real failures.
- **No migration:** confirmed — uses existing table 120, no files under `backend/migrations/` in the diff.
- **Closes #1408 — yes:** the prune (`run_blob_gc` apply-mode) is exactly what #1655 left out, so reclaim now genuinely fires end-to-end once `BLOB_GC_ENABLED=true` and the gate clears (`test_run_blob_gc_prunes_orphan_ref_and_reclaims_blob` proves the headline path).

### Test gaps
- **(should close)** No test asserts the *headline* of finding 2: that a proxy-cached **image** manifest records its `manifest_blob_refs`. The only new proxy test covers the malformed case; `cache_manifest_records_refs_for_index_manifest` is the pre-existing #1357 index test. Add a proxy-image-manifest test asserting config+layer refs land.
- **(acceptable follow-up)** No handler-level **tag-overwrite** reclaim test (overwrite `latest` A→B, run GC, assert A's now-orphan blobs reclaimed). The unit-level prune test covers the mechanism, but an end-to-end overwrite test would lock the contract.
- **(acceptable follow-up)** No concurrent-push-vs-GC race test for the documented TOCTOU. Acceptable given the grace window, but worth a tracked issue before relying on it at scale.

### Recommendation
Code is safe to merge from an over-delete standpoint. Before merge: (1) fix the PR description's false `persist_tag_and_refs`/atomicity claim, and (2) add the proxy-image-ref test. Before arming `BLOB_GC_ENABLED` in production: confirm the residual sub-grace-period race is acceptable for dedup-heavy workloads, or land the push-side `FOR UPDATE` follow-up.

PR #1415, opened 2026-05-26
fix(maven): indexable LIKE for virtual classifier fallback + parser edge cases

### LINKED ISSUE #1414: Maven virtual classifier fallback: indexable LIKE + parser edge cases (follow-up to #1399)
Round-2 review findings on top of #1399.

## Background

PR #1399 fixes Maven virtual-repo downloads for classifier artifacts (`-plain.jar`, `-test-fixtures.jar`, …). Reviewing the merged shape surfaced three follow-ups that should land alongside the fix.

### 1. DB performance trap: `path LIKE 'prefix%'` cannot use `idx_artifacts_repo_path` under `en_US.UTF-8`

`maven_local_fetch_storage_fallback` runs:

```sql
SELECT ... FROM artifacts
 WHERE repository_id = $1
   AND path LIKE '<gav-dir>/%'
   AND is_deleted = false
 ORDER BY created_at DESC
 LIMIT 1
```

The existing `idx_artifacts_repo_path` (migration 004) is a plain btree. Under the default `en_US.UTF-8` collation PostgreSQL cannot use it for LIKE-prefix; the query degrades to a seq scan within the repo's row range. Cost on a populated table:

- 50K rows: ~50-300 ms per fetch
- 1M rows: seconds per fetch

Paid on every virtual fetch of a Maven companion (`-sources.jar`, `.pom`, `.sha512`, …). Fix: add a `text_pattern_ops` index (`idx_artifacts_repo_path_pattern`) that is collation-independent and serves left-anchored LIKE.

### 2. Parser edge cases in `MavenHandler::parse_coordinates`

Now that `is_maven_secondary_path` delegates to `parse_coordinates` for classifier detection, three edge cases need pinning:

- `g/a/1.0/a-1.0-.jar` — empty classifier was returned as `Some("")`. Fixed to return `Err` so the helper refuses to route around the SQL row.
- `g/a/1.0/a-1.0-SNAPSHOT.jar` — directory version is `1.0` (no -SNAPSHOT) but filename carries `-SNAPSHOT`. Parser treats SNAPSHOT as a classifier; downstream Gate-2 + Gate-3 keep this safe. Pinned with a test.
- `spring-boot-starter/3.0/...-tests.jar` — hyphenated artifact-id with classifier `tests` parses correctly. Pinned.

### 3. Quarantine on classifier sibling

Gate 2 (live primary lookup) should also gate classifier artifacts, not just the documented companion-suffix list. Added an integration test that quarantines a primary and confirms its `-plain.jar` classifier is also refused.

brandonrc:
## Summary

Round-2 follow-up on top of #1399 (Maven virtual classifier downloads). Three items:

1. **DB performance trap.** `maven_local_fetch_storage_fallback` runs `path LIKE '<gav-dir>/%'` against `artifacts`. The existing `idx_artifacts_repo_path` (migration 004) is a plain btree and PostgreSQL **cannot** use it for left-anchored LIKE under the default `en_US.UTF-8` collation. Result: a seq scan within the repo's row range on every virtual companion fetch (`.pom`, `-sources.jar`, `.sha512`, classifier `.jar`s). Cost scales linearly with repo size: ~50-300 ms at 50K rows, seconds at 1M. Migration 110 adds `idx_artifacts_repo_path_pattern (repository_id, path text_pattern_ops) WHERE is_deleted = false`, which is collation-independent and confirmed used by the planner (rewrites `LIKE 'prefix%'` into byte-wise `~>=~`/`~<~` range conditions).

2. **Parser edge case fix.** `MavenHandler::parse_filename` previously returned `Some("")` for `artifact-version-.ext` (dangling hyphen). With #1399 routing classifier detection through `parse_coordinates`, this would have let empty-classifier paths bypass the SQL row. Now returns `Err`.

3. **Test coverage gaps.** Five new parser-edge tests pinning `is_maven_secondary_path` behavior (`.jar.sha1` suffix vs classifier, empty classifier, SNAPSHOT-mismatch, hyphenated artifact-id with classifier, invalid paths). One new integration test confirming Gate 2 (quarantine check on primary) also gates classifier siblings, not just the documented companion-suffix list. One new `parse_coordinates_rejects_empty_classifier` unit test.

Closes #1414. Sits on top of #1399 (does not modify it).

### Performance validation

EXPLAIN on a populated table (5K rows, single repo) confirms the planner picks the new index:

```
Index Scan using idx_artifacts_repo_path_pattern on artifacts
  Index Cond: ((repository_id = $0)
            AND ((path)::text ~>=~ 'com/example/grp5/art55/v0/'::text)
            AND ((path)::text ~<~  'com/example/grp5/art55/v00'::text))
  Filter: ((path)::text ~~ 'com/example/grp5/art55/v0/%'::text)
Execution Time: 0.043 ms
```

## Test Checklist

- [x] Unit tests added/updated (`test_is_maven_secondary_path_rejects_empty_classifier`, `_checksum_on_primary_classified_via_suffix`, `_snapshot_mismatch_pin`, `_hyphenated_artifact_id_with_tests_classifier`, `_invalid_paths`, `test_parse_coordinates_rejects_empty_classifier`)
- [x] Integration tests added/updated (`test_storage_fallback_honors_quarantine_on_classifier_artifact`)
- [x] Manually tested locally (`cargo test --workspace --lib api::handlers::maven_proxy` → 17/17 pass with `DATABASE_URL` set, including the new quarantine-on-classifier integration test)
- [x] No regressions in existing tests (`cargo fmt --check`, `cargo clippy --workspace --tests -- -D warnings` clean)

## API Changes

- [x] Migration is reversible (if applicable) — `DROP INDEX IF EXISTS idx_artifacts_repo_path_pattern`. The index is additive; the existing `idx_artifacts_repo_path` stays.
- [ ] N/A - no API changes


brandonrc:
## Review round 1 — 4 independent reviewers

### This PR is empty — recommend closing.

All four reviewers independently confirmed: `git diff origin/main...origin/fix/1399-followup` is **0 bytes**. The substantive commit (`8f8eab7b`) is already present on `origin/main` (merged via the #1399 chain that landed in #1455). GitHub shows `changedFiles: 0`, `mergeStateStatus: BEHIND`. There is nothing left for this PR to contribute.

**Action: close this PR.** The Maven indexable-LIKE fix, migration 110, and the empty-classifier parser guard are all live on main.

### Two findings on the now-merged code, for follow-up (NOT blockers, code is already shipped)

1. **LOW (defense-in-depth) — `maven_proxy.rs:154`:** `sibling_prefix = format!("{}/%", gav_dir)` doesn't escape LIKE metacharacters (`_`, `%`) in `gav_dir`, which is request-path-derived. Underscores are legal+common in Maven coordinates (`scala_2.13`), so the Gate-2 sibling existence check can over-match within the repo. Not exploitable for cross-repo disclosure (query is scoped by `repository_id` and bytes are served from an exact `storage_key`, not the matched row), but worth a 3-line fix: escape `\ % _` and append `ESCAPE '\'`. Adding `ESCAPE` also keeps the `text_pattern_ops` index usable.
2. **MEDIUM (perf) — `maven_proxy.rs:161`:** the gate query has `ORDER BY created_at DESC LIMIT 1` but only needs existence of any live sibling. The partial index has no `created_at`, so PG range-scans the prefix then sorts. Drop the `ORDER BY` for a free win on the hot path this migration is optimizing.

If you want these addressed, they're a tiny new PR against main (the LIKE-escape one is arguably worth doing for v1.2.0 since it touches the quarantine contract). This branch itself should just be closed.

brandonrc:
Closing per round-1 review: this branch's diff against main is empty (0 bytes). The Maven indexable-LIKE fix, migration 110, and the empty-classifier parser guard all landed on main via the #1399 chain (#1455). The two follow-up nits noted in the review (LIKE-wildcard escaping at maven_proxy.rs:154, and dropping the ORDER BY in the gate query) are tracked for a small separate PR against main if desired.

PR #1428, opened 2026-05-27
fix(incus): scan incus-export backups; harden extraction against prior-run state and setgid dirs

### LINKED ISSUE #1427: Incus scanner extraction: incus-export backups report zero findings, and a failed scan can wedge the workspace
## Summary

Three related robustness gaps in the incus scanner's workspace extraction (`backend/src/services/incus_scanner.rs`), all hit on real incus artifacts. Grouping them since they're one cohesive change to `prepare_workspace` / `extract_tarball` / cleanup — happy to split if you'd prefer.

### 1. `incus export` container backups scan empty

`prepare_workspace` extracts into `workspace/rootfs/` and hands that to trivy. Correct for `incus image export` archives (`{metadata.yaml, rootfs/…}`), but `incus export <container>` backups put the rootfs at `backup/container/rootfs/…`. For that layout trivy walks a directory with no package DB → **zero OS-package findings**, a silent false-negative.

*Fix:* after extraction, probe candidate roots (`rootfs`, `backup/container/rootfs`, `backup/rootfs`) for a package-DB marker (`var/lib/dpkg/status`, `var/lib/rpm/Packages`, `var/lib/rpm/rpmdb.sqlite`, `etc/apk/installed`, `etc/os-release`) and scan the first match; fall back to `rootfs` with a `warn!`.

### 2. A failed scan wedges the next one

`prepare_workspace` calls `create_dir_all` without wiping first, and the `tar` invocation lacks `--overwrite` / `--no-same-owner`. After a failed scan (OOM, disk-full) leaves a partial tree, the next `tar` chokes (`File exists` / `Cannot mkdir: Permission denied`).

*Fix:* `remove_dir_all(&workspace)` at the top of `prepare_workspace` (best-effort — `NotFound` is fine), and add `--overwrite --no-same-owner` to `tar`.

### 3. setgid bits survive → cleanup silently fails → workspace fills

`tar` passes no `--mode`, so special bits (e.g. `/usr/lib/modules/*` kernel-module dirs at mode `2755`) survive; with `--delay-directory-restore` a directory can end up `d--x--S---`, which the runtime UID can't list or delete. `ScanWorkspace::cleanup` (`backend/src/services/scanner_service.rs:523`) only `warn!`s on `remove_dir_all` failure and doesn't propagate — so every "successful" scan silently leaves its extraction tree on the PVC until it fills (`No space left on device`).

*Fix:* add `--mode=u=rwX,go=rX` to `tar`, and make cleanup resilient (chmod the tree owner-writable before `remove_dir_all`).

## Repro

- **(1)** publish via `incus export <container>` → scan reports 0 OS packages despite a full rootfs.
- **(3)** scan a rootfs containing setgid directories (e.g. `/usr/lib/modules/*`) → the extraction tree remains after the scan; repeated scans eventually hit `No space left on device`.

PR to follow (all three are small, same file).

dragonpaw:
## Summary

Three robustness fixes to the incus scanner's workspace extraction (`backend/src/services/incus_scanner.rs`), all hit on real incus artifacts:

1. **`incus export` container backups now scan their packages.** `prepare_workspace` extracted into `rootfs/` and pointed trivy there — correct for `incus image export` archives, but wrong for `incus export` container backups (rootfs lives at `backup/container/rootfs/…`), where trivy found no package DB and reported **zero findings** (a silent false-negative). A new `find_rootfs` probes the known layouts (`rootfs`, `backup/container/rootfs`, `backup/rootfs`) for a package-DB marker and scans the first match, falling back to `rootfs/` with a warning.
2. **A failed scan no longer wedges the next one.** `prepare_workspace` now wipes the workspace before extracting (best-effort), and `tar` gets `--overwrite --no-same-owner`, so a partial tree from a crashed prior run can't cause `File exists` / `Cannot mkdir: Permission denied`.
3. **setgid dirs no longer break cleanup.** `tar` gets `--mode=u=rwX,go=rX`, so special bits (e.g. a `2755` kernel-module dir) don't survive as `d--x--S---` — which the runtime UID can't list or delete. `--no-same-permissions` alone isn't enough (umask doesn't mask setgid/setuid/sticky). Belt-and-suspenders: `ScanWorkspace::cleanup` now chmods the tree owner-writable before `remove_dir_all`, so a "successful" scan can't silently leave a multi-GiB tree on the PVC until it fills.

Incus extraction stays bespoke (rather than using the shared `ScanWorkspace::prepare`) because it needs squashfs (`unsquashfs`) handling and rootfs-layout probing that the generic archive extraction doesn't do. Consolidating the incus scanner onto the shared workspace abstraction is worth doing but is a larger refactor — tracked separately.

`find_rootfs` unit tests cover the backup-format layout, the image-export layout, and the no-marker fallback. The backup-format test fails on `main` (no layout probe exists there).

## Test Checklist
- [x] Manually tested locally (`cargo test --lib incus_scanner::tests::test_find_rootfs` — 3 passed)

## API Changes
- [x] N/A - no API changes

Closes #1427


brandonrc:

The core fix is sound: `find_rootfs` probing `backup/container/rootfs` makes `incus export` backups produce findings (#1427 issue 1), and the `remove_dir_all` + chmod-before-cleanup correctly addresses the wedged-workspace bug (issues 2/3). But round-1 review found **security gaps in the extraction path** that should block merge.

### 🔴 High (security)
1. **Tar path traversal / symlink escape.** Extraction moved to the workspace root and added `--overwrite` with no traversal guard. A crafted archive can plant a symlink (member A = `rootfs -> /etc`) then write through it (member B = `rootfs/passwd`). `--overwrite` specifically disables GNU tar's in-archive symlink-overwrite protection. The fresh `remove_dir_all` doesn't help — the attacker controls members within one archive. **Drop `--overwrite`** (the clean-dir wipe already makes it redundant) and/or add a post-extraction symlink-target canonicalization check.
2. **No extraction size / file-count cap (decompression bomb).** Untrusted archives extract with no bound on uncompressed size or inode count → fills the PVC (DoS). The cleanup comment itself notes "often multi-GiB." Cap input bytes before write and bound extraction (quota dir, `ulimit -f`, or a streaming tar reader enforcing limits).

### 🟡 Medium
3. **Concurrent scans of the same artifact corrupt each other.** Workspace is `incus-<artifact.id>`, shared across scans. The new top-of-function `remove_dir_all` will delete the tree another in-flight scan is mid-extraction on. Use a per-run unique subdir (append scan UUID) or a per-artifact lock — unless the orchestrator already serializes per-artifact scans (please confirm).

### 🟡 Low
4. Nested-tarball backups (`backup/container/rootfs.tar`) still report zero findings — `find_rootfs` only probes plain-file markers. Not a regression, but #1427 is only half-solved for that variant.
5. `.tar.zst` not handled — zstd magic falls through to gzip and tar fails. (Tracked separately in #1454 / PR #1296.)

### 🔴 Tests
6. No real incus-export fixture test — the literal #1427 regression (extract → `find_rootfs` relocates rootfs) is never exercised end-to-end; the 3 tests only hit `find_rootfs` in isolation. Build a real gzipped tar with `backup/container/rootfs/var/lib/dpkg/status`, run `prepare_workspace`, assert the returned path + marker.
7. Wedged-workspace recovery untested (the stale-wipe branch). Coverage on changed lines likely <70%.

**Verdict: request changes.** Items 1 and 2 (security) block merge; items 6/7 needed for the coverage gate.

Also: #1429 notes incus_scanner is the only filesystem scanner not using the shared extraction helper. Consider folding `find_rootfs` + compression handling into the shared path as part of that refactor.

PR #1431, opened 2026-05-27
fix(storage/gcs): override get_stream + put_stream so multi-GiB artifacts don't buffer in heap

### LINKED ISSUE #1430: GcsBackend buffers whole artifacts in heap (no streaming get/put) — multi-GiB uploads and scans OOM the pod
## Summary

`GcsBackend` overrides neither `put_stream` nor `get_stream`, so it inherits the `StorageBackend` defaults that buffer the entire object in memory. A multi-GiB artifact allocates its full size on the heap on **both** the upload path and the scan-fetch path, OOM-killing the backend on memory-constrained nodes.

## Details (`main`)

- `StorageBackend::put_stream` default (`backend/src/storage/mod.rs:123-150`) collects the stream into a `Vec<u8>` then calls `put()`. `GcsBackend` (`backend/src/storage/gcs.rs`) doesn't override it → a 3.4 GiB upload allocates 3.4 GiB on the heap before the first byte reaches GCS.
- `StorageBackend::get_stream` default (`mod.rs:115-118`) wraps `get()` in a one-item stream; `GcsBackend::get` does `response.bytes().await` (whole body into one `Bytes`). So a "streaming" fetch actually buffers the entire object.
- Naively using `reqwest::Body::wrap_stream` for the upload also fails: GCP's L7 load balancer caps a single request at a few minutes, and a multi-GiB PUT exceeds that (408 / connection reset).

## Proposed fix

- **`get_stream`:** override on `GcsBackend` using `reqwest::Response::bytes_stream()` (pulls ~64 KiB chunks straight off the socket).
- **`put_stream`:** implement the GCS JSON API resumable-upload protocol — initiate a session, PUT in 32 MiB (256 KiB-aligned) chunks with `Content-Range`, hashing incrementally for the checksum. Per-request budget stays small; total time is bounded by GCS throughput, not an HTTP timeout. Bump the `GcsBackend` reqwest client timeout to cover the multi-minute total.

PR to follow (with `wiremock` tests for the initiate → chunk → final → checksum flow, plus a stream-error case).

dragonpaw:
## Summary

`GcsBackend` overrode neither `put_stream` nor `get_stream`, so it inherited the `StorageBackend` defaults that buffer the whole object in memory — a multi-GiB artifact allocated its full size on the heap on **both** the upload and scan-fetch paths, OOM-killing the backend on memory-constrained nodes.

- **`get_stream`:** override using `reqwest::Response::bytes_stream()` — pulls ~64 KiB chunks off the socket, so the in-flight footprint stays tiny regardless of object size. (The migration-mode Artifactory fallback still buffers; wrapped in a one-item stream so the interface stays uniform.)
- **`put_stream`:** implement the GCS JSON API resumable-upload protocol — initiate a session, PUT 32 MiB (256 KiB-aligned) chunks with `Content-Range`, hashing incrementally for the checksum. Heap stays ~32 MiB regardless of artifact size. A single `reqwest::Body::wrap_stream` PUT was rejected up front: GCP's L7 load balancer caps a request at ~3 min, which a multi-GiB upload exceeds (408 / reset); resumable chunks keep each request short.
- Bump the `GcsBackend` reqwest client timeout 30 s → 1800 s to cover the multi-minute total of a large resumable upload (per-chunk PUTs stay well under it).

Adds wiremock tests `test_get_stream_success` / `test_get_stream_not_found` exercising the `GcsBackend::get_stream` override against a stub GCS — they fail on `main`, where the default just wraps `get()`. (`put_stream`'s resumable flow is validated in-cluster; a wiremock test for the initiate → chunk → final flow is a sensible follow-up if you'd like it in this PR.)

## Test Checklist
- [x] Manually tested locally (`cargo test --lib storage::gcs::tests::test_get_stream` — 2 passed; fmt clean)

## API Changes
- [x] N/A - no API changes

Closes #1430


brandonrc:

Thanks for this — the streaming direction is right and the memory bounding (32 MiB chunk + `std::mem::take`) is solid. But round-1 review found **blocking correctness bugs** that mean multi-GiB uploads won't reliably succeed against real GCS yet.

### 🔴 Critical
1. **Chunk boundary not 256 KiB-aligned.** The inner loop stops at `buffer.len() >= CHUNK_SIZE`, but `extend_from_slice` can push past 32 MiB when an incoming `bytes` chunk straddles the boundary. GCS requires every non-final chunk to be an exact multiple of 256 KiB, so an intermediate `Content-Range` of `bytes 0-(33554431+tail)/*` gets rejected with HTTP 400. The doc comment claims alignment but the code doesn't enforce it. Fix: cap the copy at `CHUNK_SIZE` and carry the remainder into the next buffer.
2. **Orphaned resumable session on failure.** Unlike `s3.rs` (which calls `write.abort()` on error ~line 1018), the GCS path never issues `DELETE <session_url>` on a stream-read error or non-2xx chunk. GCS holds the orphaned session + partial data ~7 days. Fire a best-effort `DELETE session_url` on every error branch.
3. **No resume/retry.** The entire point (#1430) is reliability for multi-GiB transfers, but a single transient 503/429 on any chunk aborts the whole upload. GCS resumable is designed to query `bytes */*` and resume from the returned `Range`. Add bounded retry-with-backoff per chunk.

### 🟡 High
4. **Boundary finalization bug:** an upload that ends exactly on a chunk boundary can fall through `buffer.is_empty() { break }` without sending the terminal `PUT Content-Range: bytes */<total>`, leaving the session open and the object never created.
5. **Transient 5xx → opaque 500.** `require_success` maps everything to `AppError::Storage` → HTTP 500. Map upstream 429/503/5xx to `AppError::ServiceUnavailable` (→503) so callers can retry.
6. **1800s timeout on the shared client** now also applies to `get`/`exists`/`delete`/metadata. A hung control-plane call blocks a request thread 30 min. Use a separate long-timeout client only for the streaming PUTs.

### 🟡 Medium
7. No integrity check against GCS `x-goog-hash` on the final response.
8. Small-object 2-RTT regression: the proxy cache routes ALL cached objects through `put_stream`; tiny artifacts now pay initiate+chunk. Short-circuit to single-shot `put()` when the stream ends under one chunk.

### 🔴 Tests
9. **`put_stream` has zero tests** (~150 lines of resumable protocol). `wiremock` is already used elsewhere; add: multi-chunk (assert intermediate `Content-Range: bytes 0-33554431/*` and final `.../<total>`), single-chunk, zero-byte, missing `Location` header, mid-stream `Err`. Extract `fn content_range(start, len, is_final) -> String` as a pure tested helper — that's where the off-by-one risk lives. Coverage gate (>=70% changed lines) will fail without these.

`get_stream` is well-covered (wiremock, verifies chunk reassembly) — nice.

**Verdict: request changes.** Items 1, 2, 3, 9 block merge for v1.2.0.

PR #1419, opened 2026-05-27
feat(oci): virtual repo manifest+blob delegation + digest verify (closes #1348 cluster)

brandonrc:
## Summary

Supersedes #1348 and #1412. Lands the OCI virtual repository resolution feature plus the round-1 review hardening as a single cumulative PR for v1.2.0.

**What this PR contains:**

Commit 1 — `fix(oci): forward-port virtual resolution source fixes`
Original forward-port authored by @danatri in #1348. Adds manifest and blob delegation through virtual OCI repositories that walk their member proxies, including `candidate_upstream_images()` plus three unit tests for it.

Commit 2 — `fix(oci): address PR #1348 round 1 review concerns`
Round-1 review fixes from #1412:
- Digest verification (`sha256:`) on every delegated manifest and blob response so a tampering upstream can never poison the virtual repo.
- Short (5 second) negative cache around resolution misses so a single 404 storm does not get amplified across every member.
- Single canonical Docker Hub candidate (`library/<image>` is no longer emitted twice when the user-supplied path already includes `library/`).
- Four wiremock-backed integration tests (`backend/tests/oci_virtual_resolution_tests.rs`) covering: walk-to-third-member, all-members-tamper, digest-ref-mismatch fallthrough, negative-cache on full miss. These are gated on `DATABASE_URL` per the existing integration test convention.

**Why we're taking over:**

#1348 has been in DRAFT/BEHIND for several weeks and we want the v1.2.0 release to ship with virtual OCI resolution. @danatri's original analysis and forward-port stand intact — commit 1 retains @danatri's authorship metadata (verifiable via `git log --format='%an %ae' 7dc307ec`).

We are **not closing #1348 yet**. @danatri, if you'd prefer to mark your PR ready and land it as-is, we're happy to close this takeover instead. Just leave a comment on #1348.

## Regression test (required for `fix/*` PRs)
- [x] This PR is a `fix/*` AND adds/updates a test that would have caught the bug
- [ ] N/A — this is not a bug fix

The digest-verify and negative-cache regressions are covered by the new wiremock integration tests. The `candidate_upstream_images()` Docker Hub double-emission is covered by a new unit test added in commit 2.

## Test Checklist
- [x] Unit tests added/updated (3 from #1348 + extras from #1412 — 225 oci_v2 unit tests pass)
- [x] Integration tests added/updated (4 new wiremock tests in `backend/tests/oci_virtual_resolution_tests.rs`)
- [ ] E2E tests added/updated (if applicable)
[...]

Issue #1317, opened 2026-05-21
Chunked-upload session does not validate URL repo_key against session.repository_id


brandonrc:
## Summary

Across multiple chunked-upload handlers, the session lookup is keyed on `session_id` only; the URL's `repo_key` is resolved into a repo for the existence/ACL check, then its `id` is discarded. All subsequent side effects use `session.repository_id`. This means an authenticated user with write scope can complete or upload to a session belonging to repo A by calling the endpoint under repo B's URL.

## Affected handlers

- `backend/src/api/handlers/incus.rs:885-894` (`complete_chunked_upload`) — `get_session` keyed on `session_id` alone; `_repo` from URL discarded; `build_storage_key` uses `session.repository_id`. Surface made visually explicit by PR #1297 (the `_repo` underscore prefix).
- `backend/src/api/handlers/incus.rs:848` (`start_session_chunked`) — same pattern.
- `backend/src/api/handlers/oci_v2.rs:1256-1283` (`complete_upload`) — resolves session by id; resolves repo from `image_name` URL; writes blob using `session.repository_id` (line 1332) but uses URL-derived `repo.location` for storage (line 1285).

## Impact

- **Audit log lie**: URL says repo B, write goes to repo A.
- **Bypass of per-repo authorization layers**: any future per-repo policy that the auth layer wraps on top of `incus:write` / `oci:write` scopes is silently sidestepped.
- **Repo isolation contract broken**: principle that URL = resource no longer holds.

Pre-existing since PRs #217 / #242. Not exploited in the wild that we know of. The visibility on incus came from PR #1297's clippy-rename of `_repo`.

## Fix

In `get_session`-style lookups, bind `repo.id` alongside `session_id`:

\`\`\`rust
let session = sqlx::query!(
    "SELECT ... FROM incus_chunked_sessions WHERE id = $1 AND repository_id = $2",
    session_id,
    repo.id,
).fetch_optional(...).await?;
\`\`\`

OR add an explicit guard before any side effect:

\`\`\`rust
if session.repository_id != repo.id {
    return Err(AppError::NotFound("session".to_string()));
}
\`\`\`

(404 is the right response shape; "session does not exist for this repo" is semantically the same as "session does not exist" from the caller's perspective.)

Apply to every chunked-upload handler that takes both a `repo_key` and a `session_id` in the URL.

## Acceptance

- Affected handlers (incus + oci_v2) guard session→repo membership
[...]

PR #1448, opened 2026-05-27
fix(oci): stream blob uploads through storage backend

### LINKED ISSUE #1522: OCI blob upload buffers the whole blob in heap → OOM on multi-GiB layers
### Problem

The OCI blob-upload path holds the entire blob in memory:

- `handle_patch_upload` reads the whole accumulated temp object (`storage.get(&session.storage_temp_key)`), appends the new chunk in heap (`existing.extend_from_slice(&body)`), then writes the whole object back (`storage.put(...)`) — a full read-modify-write **on every PATCH chunk**.
- `handle_complete_upload` buffers the whole blob in heap before the final `storage.put`.
- axum delivers each chunk already fully buffered as `Bytes`.

### Impact

Across *k* chunks this is **O(n²)** memory + IO; a single multi-GiB image layer drives the backend to OOM (or a gateway timeout behind a reverse proxy). This is the same class of issue already fixed for GCS `get`/`put_stream` (#1431), the scanner mmap path (#1453), and migration `put_file` (#1512 / #1422) — but on the **registry push** path, which is still unaddressed.

### Repro

`docker push` an image with a multi-GiB layer to an S3/Azure/GCS-backed repository → memory climbs to (at least) the layer size and the push fails.

### Expected

Blob uploads stream chunks directly to the storage backend (S3 multipart / Azure block blobs / GCS / filesystem temp) with bounded, O(1)-per-chunk memory.

### Affected

- `backend/src/api/handlers/oci_v2.rs` (upload handlers)
- `backend/src/storage/*`

Fixed by #1448.

inspired-geek:
## Summary

Streams OCI blob uploads through the storage backend instead of buffering the whole blob in heap.

Before this PR the registry push path did a full read-modify-write of the entire temp object on **every** PATCH chunk (`get`→`extend`→`put`) and buffered the whole blob again on completion — O(n²) memory/IO that OOMs on multi-GiB layers. Now:

- Each PATCH chunk streams straight to a per-session storage part (S3 multipart / Azure staged blocks / GCS / filesystem temp); no whole-blob buffering, bounded per-chunk memory.
- DB-backed upload sessions + `oci_upload_parts` + a durable cleanup-key journal (`oci_upload_cleanup_keys`), with a storage-GC sweep that reaps abandoned sessions/parts/unreferenced keys and a purge of in-flight temp objects on repository delete.
- Incremental SHA-256 with verification on completion (mismatch keeps the session retryable); the monolithic `?digest=` syntax is validated before streaming the body.
- A storage `copy` primitive and per-backend streaming/multipart paths; an S3 fail-fast guard at the 10,000-part limit; an Azure per-upload block-ID nonce + non-destructive failure cleanup (a failed upload can never clobber a concurrent same-digest writer); a GCS rewrite-loop iteration cap.

Rebased onto current `main`: **supersedes the competing local-temp-file fix #1521** (its code is removed), incorporates the orthogonal OCI fixes #1505 (index proxied manifests) and #1519 (canonical Accept), coexists with virtual-repo delegation (#1419) and GC test-isolation (#1499), and preserves the cross-repo upload-session binding (#1317). Migrations **114–118**.

Two independent, adversarially-verified review rounds were applied; remaining LOW/optional items are tracked in #1533. Larger follow-ups: #1523 (lift the ~50 GiB S3 ceiling), #1527 (reclaim an orphaned final blob on DB failure), #1528 (stream the blob pull path), #1529 (cloud-backend OCI E2E).

Closes #1522.

  - Streaming tests assert chunks land as discrete storage parts; cross-repo session regression; per-backend multipart/block/resumable + abort tests; Azure unique-block-id + non-destructive cleanup + block-id length; S3 part-cap guard; digest-mismatch keeps the session retryable; cancel partial-delete keeps the session non-resumable; repo-delete purges upload temp objects but not committed blobs; GC recoverability of orphaned completion temp objects.

## Test Checklist
- [ ] E2E tests added/updated (if applicable) — tracked in #1529
- [x] Manually tested locally
- [x] No regressions in existing tests (full lib suite + DB-backed suites green; clippy/fmt clean; new-line coverage 80%, duplication 2.61%)

## API Changes
- [ ] New endpoints have `#[utoipa::path]` annotations — N/A: OCI distribution endpoints are spec-defined, not part of the utoipa OpenAPI surface
- [ ] Request/response types have `#[derive(ToSchema)]` — N/A
- [x] Migration is reversible (if applicable) — migrations 114–118 are additive/idempotent (`IF NOT EXISTS` / `DROP NOT NULL`)


brandonrc (commented):
**Verdict: Approve.** The streaming upload path preserves digest integrity, chunk ordering, concurrency safety, and resource cleanup. No critical issues.

I verified each in-scope risk against the code:

- Digest verification is intact on every path. Monolithic POST `?digest=` verifies `provided_digest != computed` and stores under the canonical computed digest, not the raw param (`oci_v2.rs:3483`, `:3501`). Multi-part completion recomputes SHA-256 over the concatenated object via `put_stream` and rejects on mismatch before the blob copy (`:4583`, `:4613`). The single-part fast path only skips re-hashing when `session.computed_digest == requested_digest` AND `parts.len()==1` AND no final body (`:4388`), and that cached digest is the same backend's hash of the same bytes (`:3774`, `:3572`), so the copy cannot smuggle unverified content.
- Chunked offset tracking is enforced. `validate_patch_content_range` requires `start == bytes_received`, rejecting out-of-order/resumed chunks, with overflow-safe span math (`:647`, `:663`, `:683`); negative ranges rejected in the parser (`:641`).
- Concurrent PATCH/complete is safe. PATCH uses a unique constraint on `(upload_session_id, part_index)` plus an optimistic `bytes_received` predicate (`:3819`, `:3837`, `:3855`); the unique violation maps to 409 (`:3831`). Completion atomically claims Open to Committing with a `state_token` lease plus heartbeat and an authoritative `state_token` predicate on the terminal DELETE (`:1004`, `:4762`).
- No truncation or committed partials. Filesystem writes to `.tmp.<uuid>` then fsync plus atomic rename, removing the temp on every error branch (`filesystem.rs:280-337`). S3 `put_stream` aborts the multipart upload on any part/complete error (`s3.rs:1244`, `:1290`) with a 10,000-part guard. Azure stages per-upload-nonce block IDs so a failed writer cannot clobber a concurrent same-key writer (`azure.rs:917`, `:893`). GCS rewrite loop is iteration-capped.
- No leaks. Every temp/part/completion object is registered in the `oci_upload_cleanup_keys` journal before the write and reaped by storage GC; the completion-temp register is correctly ordered before its `put_stream` (`:4445-4452`). Cancel intentionally does not reset a half-deleted session back to Open, leaving it for GC (`:4062`).
- Spec headers preserved. POST returns `Location`, `Docker-Upload-UUID`, `Range` (`:3652-3660`); monolithic returns 201 plus `Location` plus `Docker-Content-Digest` (`:3533`); streaming is genuinely bounded (per-chunk only).

Summary: a large but disciplined change. The digest is verified before any blob is promoted to its canonical key on all three paths (monolithic, single-part fast path, multi-part concat), chunk ordering and concurrency are enforced through DB constraints plus a lease/state-token protocol, and every transient object goes through a durable cleanup journal. I found nothing that would let corrupted content be stored, no truncation or leaked-partial path, and no unbounded buffering reintroduced. Approve.

PR #1453, opened 2026-05-27
fix(scanner): stream scan inputs to an mmap-backed Bytes instead of buffering in heap

### LINKED ISSUE #1452: Scanner buffers the entire artifact in heap (fetch_artifact_content uses get()) — multi-GiB scans OOM the backend
## Summary

`ScannerService::fetch_artifact_content` (`backend/src/services/scanner_service.rs`) calls `storage.get(key)`, which materialises the whole artifact as a `bytes::Bytes` on the heap (anon memory), then holds it alive across every applicable scanner in `scan_artifact_inner`'s loop. For a multi-GiB artifact that's several GiB of **unreclaimable anon-rss**, so the backend OOMs on memory-constrained nodes — a 3.5 GiB incus tarball reliably OOM-kills on a node with ~12 GiB allocatable (`kubectl describe pod` shows `Last State: Terminated, Exit Code: 137`).

## Root cause (`main`)

`fetch_artifact_content` uses buffer-everything `storage.get(key)`. Anon pages can't be evicted by the kernel under cgroup pressure, so the OOM killer is the only outcome. `memmap2` is not currently a dependency.

## Proposed fix

Rewrite `fetch_artifact_content` to:

1. stream `storage.get_stream(key)` into a tempfile under `scan_workspace_path`, then
2. `mmap` the tempfile (`memmap2`) and wrap it in `Bytes::from_owner(mmap)` (bytes 1.6+).

The scanner trait keeps its `&Bytes` signature, but the pages are now **file-rss** (kernel-reclaimable under cgroup pressure) instead of anon-rss. On Linux the unlink-after-mmap pattern auto-cleans the tempfile when the `Bytes` drops.

Depends on `GcsBackend::get_stream` actually streaming rather than buffering (#1431) — otherwise the source `Bytes` is still fully materialised before the mmap stage.

PR to follow.

dragonpaw:
## Summary

`ScannerService::fetch_artifact_content` buffered the whole artifact in heap via `storage.get(...)` and held that allocation alive across every applicable scanner — so a multi-GiB input OOM-killed the backend (anon-rss the kernel can't evict under cgroup pressure).

This streams `storage.get_stream(...)` to a tempfile under `scan_workspace_path`, `mmap`s it (`memmap2`), and hands the scanner trait a `Bytes::from_owner(mmap)` — **file-backed, kernel-reclaimable** pages. Consumers still get the same `&Bytes`; the tempfile is unlinked once mapped, so disk frees when the `Bytes` drops.

The streaming + mmap logic is factored into a `stage_stream_to_mmap` helper so it's unit-testable without a DB or live storage backend. This is most effective alongside a real `GcsBackend::get_stream` (#1431); on `main` it already converts the scan-input allocation from anon to file-backed.

Adds a dependency on `memmap2 = "0.9"`.

`stage_stream_to_mmap` unit tests: a multi-chunk stream is concatenated correctly into the mmap-backed `Bytes`, and an empty stream yields empty `Bytes`. They guard the staging/mmap path against truncation/corruption (no DB needed).

## Test Checklist
- [x] Manually tested locally (`cargo test --lib services::scanner_service::tests::test_stage_stream_to_mmap` — 2 passed; fmt clean)

## API Changes
- [x] N/A - no API changes

Closes #1452


brandonrc:

The mmap-staging approach (stream → tempfile → `Bytes::from_owner(Mmap)`) is the right idea for converting anon-RSS to reclaimable file-RSS, and there's no use-after-unmap. But round-1 review found issues that mean the OOM in #1452 isn't actually fixed for cloud backends, plus safety gaps.

### 🔴 High
1. **Doesn't help GCS/Azure (the regression that matters).** `gcs.rs` and `azure.rs` don't override `get_stream`, so they inherit the default (`storage/mod.rs`) which calls `get()` — full object into anon heap — then this code *additionally* writes a tempfile + mmaps it. For GCS/Azure the peak memory is unchanged AND latency/disk get worse. Only filesystem + S3 truly stream. Either implement chunked `get_stream` for GCS/Azure (see #1431 for GCS) or gate the mmap-staging path to backends that genuinely stream.
2. **Reopen-by-path is racy (TOCTOU).** The code reopens the tempfile by path for writing instead of using the handle. Use `temp.reopen()` (independent fd to the same inode) or `temp.as_file_mut()`.
3. **`unsafe { Mmap::map }` has no `// SAFETY:` comment and relies on "unlink pins the inode" without a platform guard.** SIGBUS if the backing file is truncated by another process (uncatchable, takes down the backend). Document the invariant and require `SCAN_WORKSPACE_PATH` to be a local fs the backend exclusively owns (not NFS/shared emptyDir).

### 🟡 Medium
4. **No size cap before staging** — an attacker artifact forces same-size disk write + same-size VA mapping per scan; concurrent scans multiply it. `Artifact.size_bytes` is available; enforce a ceiling before `fetch_artifact_content`.
5. **Small-file regression** — every scan now does `create_dir_all` + spawn_blocking + tempfile + mmap even for a few-KB package. Add a size threshold: buffer small artifacts, mmap only above ~a few MiB.
6. **Doc comment wrong about unlink timing** — claims the inode is pinned until scanning finishes; it's actually freed at function return. Archive scanners also re-buffer via `prepare_scan_workspace` (`tokio::fs::write` copies the full mmap back to a new file), so the win is partly undone for the exact incus-tarball case in #1452.

### 🔴 Tests
7. `fetch_artifact_content` itself (the bug-fixed fn) is untested — only the private helper is. Add an integration test with a filesystem backend asserting mmap-backed bytes == stored content.
8. Mid-stream `Err` / cleanup path untested. Add `stream::iter([Ok(..), Err(..)])`, assert error propagates and no tempfile leaks.
9. Multi-chunk large-file path (the OOM boundary) untested. Coverage on changed lines likely <70%.

**Verdict: request changes.** Item 1 means the headline bug isn't fixed for GCS/Azure; items 2, 3, 7 are the other blockers.

PR #1494, opened 2026-05-29
fix(incus): return 202 and finalize uploads on a background task, with observable status

### LINKED ISSUE #1471: Incus uploads silently bypass StorageBackend on GCS/S3 (data-loss) — and the obvious fix exposes a multi-GiB gateway-timeout problem
## Part 1: Data-loss bug (current behaviour)

For GCS- or S3-backed deployments, both incus upload paths persist artifact bytes to the upload pod's local `STORAGE_PATH` filesystem instead of the configured `StorageBackend`. After staging the body to a temp file, both handlers call `finalize_temp_file` which performs `tokio::fs::rename(temp_path, final_path)` unconditionally — no `StorageBackend::put` / `put_stream` involvement at all.

`backend/src/api/handlers/incus.rs:1140-1147`:

```rust
async fn finalize_temp_file(temp_path: &Path, final_path: &Path) -> Result<(), Response> {
    ensure_parent_dirs(final_path).await?;
    tokio::fs::rename(temp_path, final_path)
        .await
        .map_err(|e| fs_err("finalize upload", e))?;
    Ok(())
}
```

Called from both `upload_image` (monolithic, line 673) and `complete_chunked_upload` (chunked, line 969).

The matching download handler at line 585-622 is correspondingly compensating: it reads from the same pod-local path rather than calling `storage.get_stream(...)`, because (per the inline comment ref'ing #1441) a bucket GET would 500 since the upload never landed there. So the read/write pair "works" — but only as long as the **same pod** that handled the upload also handles every subsequent download.

### Impact

- **Single-replica, no restart**: works incidentally.
- **Pod restart / reschedule / scale-down**: artifact is on ephemeral storage, **lost** silently. Operator sees the artifact row in the DB; every GET 404s on the new pod.
- **Multi-replica**: GETs that hit a different pod than the PUT did 404 immediately. Intermittent depending on which replica serves the request.

All other format handlers route through the StorageBackend correctly — `oci_v2.rs:1788`, `npm.rs:1144`, `cargo.rs:612`, `conda.rs:3075`, `terraform.rs:543`, `vscode.rs:364`, `cocoapods.rs:356`, `jetbrains.rs:446`, `composer.rs:800`, `promotion.rs:562/742`, `proxy_helpers.rs:1958`. The incus path is the outlier.

GCS already ships a working `impl StorageBackend` with `put_stream` (`backend/src/storage/gcs.rs:1333`, added in #1431 *"override get_stream + put_stream so multi-GiB artifacts don't buffer in heap"*). The trait is in place; the incus handlers just don't call it.

## Part 2: The obvious fix exposes a second problem

The mechanical fix is to wire both upload sites into `StorageBackend::put_stream`. Sketch:

```rust
let storage = state.storage_for_repo(&repo).map_err(/* … */)?;
let file = tokio::fs::File::open(&temp_path).await.map_err(fs_err)?;
let stream = tokio_util::io::ReaderStream::new(file)
    .map_ok(Bytes::from)
    .map_err(/* into storage::Error */);
storage.put_stream(&storage_key, Box::pin(stream), size_bytes).await
    .map_err(/* … */)?;
let _ = tokio::fs::remove_file(&temp_path).await;
```

Matching download handler (line 585-622): switch back to `storage.get_stream(&storage_key)`.

But once that lands, the upload handler's HTTP request lifetime extends from "stage to disk + rename" (sub-second) to "stage to disk + stream-PUT to bucket" (2–5 minutes for a multi-GiB artifact at typical GCS throughput).

Behind an L7 gateway with a fixed request timeout — common defaults are 30–120s, and not always tunable (the GKE internal Gateway class rejects `GCPBackendPolicy.timeoutSec`) — the gateway returns **504 Gateway Timeout** and severs the upstream connection. That abort propagates into the backend's in-flight PUT, so the object never lands even though the client successfully streamed every byte.

Reproduced in dev with a 3.5 GiB image **before** the data-loss fix is in:
- via gateway: client streams 3.5 GiB successfully, gets 504 at ~134s, no artifact row, nothing in bucket.
- direct to in-cluster Service: same PUT, 201, artifact lands.

(Once the data-loss fix is in, the gateway side fails the same way for the same reason — the timing window just becomes the GCS PUT instead of the local rename.)

## Options (maintainer's call)

- **A. Fix data-loss only, document minimum gateway timeout.** Land the `put_stream` rewire, add a chart/README note that operators need a request timeout proportional to `largest_artifact ÷ object_store_throughput`. Cheap. Doesn't help on gateways that don't allow timeout tuning (GKE internal Gateway), and pushes the operational concern entirely onto operators.
- **B. Fix both at once with async completion.** Stage the body to disk on the backend (as today), return `202 Accepted` (or `201` with a finalising status), and complete the object-store PUT on a background task. Expose completion via the existing scan/artifact status. Gives clients a progress signal during the multi-minute upload. Biggest change; most reviewer-correct shape.
- **C. Ship (A) immediately, plan (B) as a follow-up.** Closes the data-loss bug for clusters with tunable gateways without blocking on the bigger design conversation. The async-completion work then becomes a clean follow-on that doesn't have the data-loss bug compounding the urgency.

(B) is the right end-state; (C) is the pragmatic sequencing.

## PR

I have a partial fix for Part 1 in progress on a local branch. Happy to send a PR matching whichever option the maintainer prefers — in particular: same `state.storage_for_repo(...)` lookup as other handlers? Worth a shared `finalize_via_storage_backend` helper? For (B), where should the background task's completion state live (existing artifact status enum, new column, scan_results)?

dragonpaw:
> **Updated 2026-06-03 (reworked + rebased onto `main`):** rebased onto current `main`, which already contains #1524 (download reads via `get_stream` with the same key the upload `put_stream`s, fixing the read/write 404). Dropped the two commits superseded by #1524, so this PR is now only the 202/background-finalize — applied to **both** `upload_image` and `complete_chunked_upload`, with an observable session status (migration 119). Round-trip + failure tests pass against a local Postgres. See the comment below for the point-by-point review response.

---

## Summary

`upload_image` had two bugs, both reported in #1471:

1. **Data loss on cloud backends** — it finalized with `finalize_temp_file` → `tokio::fs::rename`, so on a GCS/S3-backed repo the artifact was renamed onto the pod's local filesystem while reads/scans/downloads go through `StorageBackend` and 404. Route the staged bytes through `StorageBackend::put_stream` instead (the pattern every other format handler uses).
2. **Gateway timeout on multi-GiB uploads** — that persist ran synchronously before the response, so a multi-GiB cloud write outlived an L7 gateway's request timeout (the GKE internal Gateway class doesn't honour `GCPBackendPolicy.timeoutSec`). The gateway returns 504 and severs the upstream connection, aborting the in-flight persist, so nothing lands even though the client streamed every byte (curl reports the full `size_upload`, then HTTP 504).

Fix: once the body is fully staged to local disk, run persist + row upsert on a background task and return `202 Accepted`. Request duration now tracks body-receipt, not object-store latency; completion is observable via the artifact appearing. Verified on a dev cluster: a 3.5 GiB upload through the gateway now returns 202 and the object lands in GCS (previously 504, no object).

Closes #1471

The bug is integration-level (object-store routing + an L7 gateway timeout), so the regression that catches it is an end-to-end test in `artifact-keeper-test`: upload a multi-GiB artifact to a cloud-backed incus repo through the gateway, assert the response is `202` and the artifact is subsequently retrievable (fails on `main`: `tokio::fs::rename` leaves nothing in the bucket, and the sync persist 504s through the gateway). Happy to add that e2e test (or a handler-level test if you'd prefer it in this repo) — flagging here rather than checking the box since I'd like your steer on where you want the coverage. Manually validated on a dev cluster as above.

## Test Checklist

- [ ] E2E tests added/updated (if applicable) — proposed, see above
- [x] Manually tested locally (dev cluster: 3.5 GiB gateway upload → 202 → object in GCS)

## API Changes

- [x] N/A — same route; the success status changes `201 → 202` (both 2xx; format clients treat any 2xx as success)


brandonrc (commented):
**Verdict: Request changes.** The cloud data-loss diagnosis is right and the `put_stream` write change is the correct direction, but the fix is one-sided. The write path now persists through `StorageBackend`, while the read path (`download_image`) was not updated, so uploaded images become unservable on every backend (404). This reframes the data loss rather than fixing it.

### Critical findings

**1. Write/read path mismatch, downloads 404 after upload (both filesystem and cloud)**

- Write (changed): `persist_to_storage` then `storage.put_stream(storage_key, ...)` (`backend/src/api/handlers/incus.rs:1221-1241`). For the filesystem backend this routes through `FilesystemStorage::key_to_path`.
- Read (unchanged): `download_image` still does `storage_path_for_key(&state.config.storage_path, &storage_key)` then `tokio::fs::read(&final_path)` (`incus.rs:597-598`).

The storage key is hierarchical: `incus/{repo_id}/{path}` (`build_storage_key`, `incus.rs:244-246`). The two path functions disagree on multi-component keys:
- `FilesystemStorage::key_to_path` skips the 2-char shard prefix when the key has more than one component (`backend/src/storage/filesystem.rs:64-66`), writing to `<base>/incus/{repo_id}/{path}`.
- `storage_path_for_key` always prepends the shard prefix (`incus.rs:167-168`), reading from `<base>/in/incus/{repo_id}/{path}`.

Result:
- Filesystem backend: write lands at `.../incus/...`, read looks at `.../in/incus/...`, every `GET /images/...` returns 404. A regression on the previously-working local-disk path.
- Cloud backend (S3/GCS): write now correctly goes to the bucket, but `download_image` still `tokio::fs::read`s local disk, which never has the object, so 404. The original #1471 symptom is unchanged.

Fix: update `download_image` (and the SimpleStreams catalog / any other read site) to fetch via `storage.get_stream(&storage_key)` / `storage.get(&storage_key)` on the same `StorageBackend` the upload wrote to, so write and read share one key-resolution scheme. Do not mix `put_stream` writes with `storage_path_for_key` reads.

**2. The added test does not exercise the new code path**

The handler test at `incus.rs:2023-2036` writes and reads via `storage_path_for_key` directly, never calling `put_stream` or `download_image`, so it passes while the real upload-then-download round trip is broken. A regression test must drive the actual upload handler (or at least `persist_to_storage` + `download_image`) end to end.

**3. (Secondary) 202 background-finalize failures are silently lost**

If the spawned task's `put_stream` or `upsert_artifact` fails, it only logs `tracing::error!` and returns, but the client already received `202 {"status":"finalizing"}`. There is no status endpoint for monolithic uploads (`get_upload_progress` at `incus.rs:1054` is keyed by chunked-session UUID, which monolithic uploads do not create), so a failed finalize is unobservable, with no retry across a pod restart (temp file is on emptyDir). Acceptable only if completion is observable; today it is not. Consider persisting a pending/failed marker row so clients can poll. Lower priority than finding 1 because downloads are broken regardless.

### Summary

The `put_stream` write change is right, but only the write half moved onto `StorageBackend` while `download_image` still reads local disk via a different key-to-path scheme, so uploaded images cannot be served on either backend. Migrate the read path to the same backend before this lands, add a true round-trip regression test, and address the 202 observability gap as a follow-up.

dragonpaw:
Reworked and rebased onto current `main`. Thanks for the careful review — addressing each point:

**1. Write/read key mismatch → 404 (critical).** Already resolved on `main` by #1524: `upload_image` writes via `put_stream(storage_key)` and `download_image` now reads via `get_stream(storage_key)` — same key, same backend, byte-aligned (`storage_path_for_key` is now dead-code/operator-only). The review was against this PR's older base. I rebased onto `main` and **dropped the two now-redundant commits** (route-through-StorageBackend / `put_stream`) that #1524 superseded, so this PR is now *only* the 202/background-finalize.

**2. The test didn't exercise the real path.** The round-trip lib tests now drive the **actual handlers** end-to-end via the router: PUT (or POST→PATCH→PUT-complete) → poll `GET /uploads/{id}` until `completed` → GET `download_image` → assert byte-equality. Verified green against a local Postgres (`cargo test --lib -- incus`).

**3. 202 finalize failures were silently lost.** Now observable. Both paths record an `incus_upload_sessions` row in `finalizing`, return 202, and finalize on a background task that flips the row to `completed` (with `artifact_id`) or `failed` (with the error string). `GET /incus/{repo}/uploads/{id}` reports `status`/`finalize_error`/`artifact_id`; the chunked `complete` no longer deletes the session on success (retained for polling, reaped by the existing stale-session sweeper). New test `failed_finalize_marks_session_failed` asserts a failed push surfaces as `status='failed'`.

**Scope note:** I applied this to **both** `upload_image` *and* `complete_chunked_upload` — the chunked `complete` did the same synchronous multi-GiB push, so it had the identical gateway-timeout exposure. Migration 119 adds the three session columns; `upsert_artifact`/`put_temp_file_to_storage` now return `Result<_, String>` so the background task can record the error. This changes the success contract for both paths from `201` to `202` (clients poll for completion).

brandonrc (approved):
Re-reviewed after the rework. All three prior findings are resolved. Approving (PR is BEHIND main; rebase before merge).

Status of prior findings:
- 1. Write/read key mismatch (the blocker): FIXED. `download_image` no longer reads local disk. It reads the artifact's stored `storage_key` from the DB row (`incus.rs:669`) and streams it through the repo's configured backend via `storage.get_stream(&storage_key)` (`incus.rs:694`), the same key the write path persisted from `build_storage_key` (`incus.rs:296`). Because the read uses the literal stored key instead of recomputing a path, the old `key_to_path` vs `storage_path_for_key` shard-prefix divergence is structurally impossible. No `tokio::fs::read` of an artifact object remains. (The core of this landed in #1524, now on main; this PR builds on it.)
- 2. Real round-trip regression test: FIXED. The streaming regression tests and `backend/tests/incus_upload_tests.rs` now PUT an image, poll finalize to `completed`, then GET it back and assert byte equality through the real put_stream/get_stream pipeline (monolithic + chunked + resume + upsert).
- 3. Silent background-finalize failure: FIXED. Migration 119 adds `status`/`finalize_error`/`artifact_id` to `incus_upload_sessions`; `finalize_upload` records `completed`+artifact_id or `failed`+error, surfaced via `GET /incus/{repo}/uploads/{id}`, with a test (`failed_finalize_marks_session_failed`).

No new critical findings. One non-blocking follow-up: a process crash between the 202 and finalize completing leaves the session in `finalizing` with no reaper-driven retry, and the 201 -> 202 status change is a client-visible contract shift (acceptable since format clients treat any 2xx as success). Worth a follow-up issue for restart recovery, not a blocker. Rebase and merge once green.

dragonpaw:
Rebased onto current `main` per your approval note. One substantive resolution beyond a plain replay: #1622 (the `/tmp`→`STORAGE_PATH/.incoming` staging fix) merged after this PR's base and rewrites the same `upload_image` / `complete_chunked_upload` persist block. Integrated the two — kept #1622's `.incoming` staging + `StagedTempFile` guard, and `disarm()` it before spawning the background finalize so the staged file survives for the async push (finalize owns cleanup). Also renumbered the session-columns migration to `124` (it collided with `120_manifest_blob_refs.sql` on current `main`). `cargo check --workspace --all-targets` is green. No behavioral change from the version you approved — same 202 + observable-session contract.

brandonrc:
Rebased onto current `main` and verified for adoption.

**Rebase**: clean — no commits on main since the branch point touch `incus.rs`. The only conflict was a migration-number collision: `124` was taken on main by `124_scan_results_status_not_applicable.sql` (and `125_repo_delete_fk_cascade.sql` also landed), so the finalize-status migration was renumbered `124 -> 126` to keep the sequence monotonic.

**Correctness review**
- 202 response shape: `upload_image` and `complete_chunked_upload` now return `202 Accepted` with `{session_id, status: "finalizing", ...}` plus `Location` and `Upload-UUID` headers, matching the incus/LXD async-operation convention.
- Background finalize: `tokio::spawn(finalize_upload(...))` with fully-owned `FinalizeParams`; `run_finalize` does the StorageBackend `put_stream` + artifact upsert + `scan_on_upload` trigger off the request path, so a multi-GiB backend push no longer trips an L7 gateway 504.
- Observability: `GET /{repo}/uploads/{id}` surfaces `status` / `finalize_error` / `artifact_id`. Session rows are retained on completion (not deleted) so the terminal status is pollable; the stale-session reaper cleans terminal rows after `max_age_hours`.
- Failure path: `finalize_upload` flips the session to `failed` with the error string and removes the temp file on every branch.

**Interaction with the #1573/#1622 staging-under-STORAGE_PATH OOM fix (no regression)**
- The staged temp file is still created via `temp_upload_path(&state.config.storage_path, ...)` (under STORAGE_PATH, not `/tmp`), and the background task reads from that same path. No reintroduction of `/tmp` staging.
- Orphan safety: for the monolithic path the `StagedTempFile` RAII guard stays armed until the session-row INSERT succeeds, so a failed INSERT still removes the temp file; `disarm()` only hands cleanup to `finalize_upload` after the row exists. For chunked, the temp file is always tracked by `storage_temp_path` on the session row. A crash after staging is still caught by the age-based `sweep_orphan_staging_files` sweep (#1573) and the session reaper. No new leak/orphan vector.

**Gates (local)**: `cargo fmt --check` ✅, `cargo clippy --workspace --all-targets -D warnings` ✅, `cargo test --workspace --lib` ✅ (188 incus lib tests pass; the only failure was `http_client::test_large_object_client_builder_does_not_apply_total_timeout`, a loopback-TCP test unrelated to this PR that the sandbox can't run). jscpd: `incus.rs` and test files are gate-exempt; the migration + 1-line `#[derive(Clone)]` on `RepoInfo` add no duplication. Changed-line coverage is exercised by the DB-backed lib tests (monolithic + chunked 202/finalize regression tests, `failed_finalize_marks_session_failed`) which the CI coverage job runs against its Postgres.

Queued for maintainer review (label `needs-maintainer-review`); not merged.

PR #1500, opened 2026-05-29
fix(incus/scanner): accept real OS images while keeping extraction guards

### LINKED ISSUE #1492: Incus extraction guards added in #1428 reject real OS images (caps, perms, absolute symlinks)
### Description

The extraction hardening added during the #1428 round-1 security review (`backend/src/services/incus_scanner.rs`) makes the `incus-image` scanner unable to scan a real multi-GiB OS image. Each guard is individually reasonable against an adversarial mini-tarball, but a legitimate OS rootfs (e.g. an `incus export` container backup) trips three of them in sequence. Filing to @brandonrc directly since these guards are yours (commits `1680e77` / `c765bc1`). The overall scan still reports `completed` because the grype/dependency scanners run, which masks that the format-native rootfs scan was skipped — so this also fails silently.

Three distinct blockers, all in the post-#1428 code:

1. **Compressed-size cap too small** — `MAX_COMPRESSED_INPUT_BYTES = 2 GiB`; a real OS image is several GiB compressed, so it's rejected before extraction even starts.
2. **Extracted tree not owner-traversable** — after `--no-same-owner` extraction the non-root scanner UID can't traverse parts of a real rootfs (and tar's implicitly-created parent dirs miss the archive `--mode`); the post-extraction guard's `read_dir` hits `EACCES`, and trivy would hit the same wall reading the package DBs.
3. **Absolute symlinks rejected as traversal** — `reject_escaping_symlink` resolves an absolute target against the *host* root, so the ubiquitous `/var/run -> /run` (present in essentially every OS rootfs) is flagged as escaping the workspace. No real image can pass this.

### Environment Information

- Deployment type: Kubernetes (GKE, Gateway API / Envoy)
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL 15
- Reverse Proxy: GKE internal Gateway
- Network setup: internal-only gateway, GCS storage backend

### Backend Version

Built from `main` @ `3de5795`.

### Package Format

Other — `incus`

### Expected Behavior

A multi-GiB `incus export` OS image, with the normal absolute symlinks and restrictive permissions any real rootfs contains, extracts and gets scanned by the `incus-image` scanner.

### Observed Behavior

Scanning a 3.5 GiB Ubuntu-based `incus export` image, the `incus-image` scan `failed`:

```
Incus image extraction failed: Incus archive contains symlink
.../backup/container/rootfs/var/run -> /run escaping the workspace; refusing extraction (path traversal)
```

With the symlink guard relaxed, the 2 GiB compressed cap and then the `read_dir` `EACCES` on `backup/container` surface in turn. Throughout, the overall artifact scan still reports `completed` (grype/dependency scanners ran), hiding the skipped rootfs scan.

### Steps to Reproduce

1. Create an `incus` repo with `scan_enabled = true` (and `scan_on_upload = true`).
2. Upload a >= 2 GiB `incus export` OS-image `.tar.zst` (a real container rootfs, with absolute symlinks like `/var/run -> /run`).
3. `GET /api/v1/security/artifacts/<id>/scans` -> overall `completed`, but the `incus` scan_type record is `failed` with the extraction error above.

### Logs

```shell
ERROR scanner_service: Scanner incus-image failed: Incus image extraction failed: Incus archive contains symlink .../backup/container/rootfs/var/run -> /run escaping the workspace; refusing extraction (path traversal)
INFO  grype_scanner: Grype scan complete: 0 vulnerabilities
```

### Additional Context

Follow-up to #1428. Suggested fixes (happy to send a PR for review):

1. Make the caps env-tunable with OS-image-sized defaults (e.g. `MAX_INCUS_SCAN_COMPRESSED_BYTES` / `MAX_INCUS_SCAN_EXTRACTED_BYTES`), keeping the bomb guards in force at a sane ceiling.
2. `chmod -R u+rwX` the extracted tree before the guard walk / trivy scan so a non-root scanner can read a real rootfs.
3. Resolve absolute symlink targets relative to the rootfs root (chroot semantics): `/run` -> `<rootfs>/run` (in-tree, benign), while still rejecting `../`-style escapes that normalise outside the workspace.

All three validated locally against a real 3.5 GiB OS image. Also worth considering: surface a per-scanner failure (rather than overall `completed`) when a format-native scanner errors, so this kind of gap isn't silent.

brandonrc:
## Summary

The #1428 extraction hardening rejected three things every real Incus OS image ships with, so a multi-GiB `incus export` container backup never got past extraction. The overall scan still reported `completed` because grype and dependency scanners ran, hiding the silent skip of the format-native rootfs scan.

Three blockers, all in post-#1428 code, fixed here:

1. **Compressed cap was 2 GiB hardcoded.** A real `incus export` of a desktop OS sits in the 3-6 GiB range compressed and was rejected before extraction even started. The cap, the extracted-tree byte cap, and the extracted-entry cap are now env-tunable (`MAX_INCUS_SCAN_COMPRESSED_BYTES` / `MAX_INCUS_SCAN_EXTRACTED_BYTES` / `MAX_INCUS_SCAN_EXTRACTED_ENTRIES`) with OS-image-sized defaults (8 GiB / 40 GiB / 5M). The bomb posture is preserved: the upper bound is still enforced before any bytes hit disk, and bogus values fall back to the default.
2. **Owner-traversal after extraction.** A real rootfs has restrictive perms on parts of the tree, and tar's implicitly-created parent dirs miss the archive `--mode`, so the post-extract guard's `read_dir` hit `EACCES` and trivy would have hit the same wall on the package DBs. After tar and unsquashfs extraction we now run `chmod -R u+rwX` on the tree. Only owner perms widen; group/other and special bits (setuid/setgid/sticky) are preserved. chmod failures (sockets, fifos, immutable attrs) are logged and ignored rather than failing the scan.
3. **Absolute symlink rejection.** `reject_escaping_symlink` resolved absolute targets against the host root, so the ubiquitous `/var/run -> /run` (in essentially every OS rootfs) was flagged as escaping the workspace. Absolute targets are now resolved with chroot semantics against the extracted root: `/run` becomes `<root>/run`, in-tree and benign. The lexical `starts_with(canonical_root)` check still rejects any `..`-chain target that pops past the root, including absolute ones, so the security invariant holds.

The silent-failure aspect (overall scan reads `completed` because other scanners ran) is real and worth a separate fix; I am tracking that in a follow-up since it touches the orchestrator surface and is independent of the extraction guards themselves.

Closes #1492. Follow-up to #1428.

New tests:

- `test_absolute_symlink_within_root_is_allowed_chroot_semantics`: the canonical `/var/run -> /run` shape passes the guard.
- `test_absolute_symlink_with_dotdot_escape_is_rejected`: an absolute target with a `..` chain still escapes and is still rejected (chroot semantics relax the "no absolute targets" rule, not the lexical reachability check).
- `test_enforce_extraction_limits_accepts_realistic_rootfs_layout`: a realistic rootfs tree with the five absolute symlinks every real Linux distro ships passes the guard end-to-end.
- `test_compressed_cap_is_env_tunable`: the env override is read by the production helper and a value previously above the cap is accepted when the operator raises it.
- `test_compressed_cap_bogus_env_falls_back_to_default`: invalid env values fall back to the default rather than panicking; the cap is still enforced.

Existing test `test_squashfs_extracted_tree_rejects_escaping_symlink` is updated to use a relative `..`-chain target rather than an absolute host path, because under chroot semantics the absolute case is intentionally permitted (it resolves dangling in-tree, which is harmless).

## Test Checklist
- [x] Manually tested locally (`cargo test --lib services::incus_scanner` -> 66 passed; backend boots on a local port with the new env var set)

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/services/incus_scanner.rs
+++ b/backend/src/services/incus_scanner.rs
@@ -23,20 +23,67 @@ use crate::services::scanner_service::{
     cached_trivy_cli_version, fail_scan_path, ScanOutput, ScanWorkspace, Scanner, VersionCache,
 };
 
-/// Maximum compressed input size we will attempt to extract (2 GiB). Untrusted
-/// archives can be decompression bombs; refuse anything larger than this before
-/// writing it to disk. The uncompressed tree is further bounded by
-/// [`MAX_EXTRACTED_BYTES`] / [`MAX_EXTRACTED_ENTRIES`] after extraction.
-const MAX_COMPRESSED_INPUT_BYTES: u64 = 2 * 1024 * 1024 * 1024;
+/// Default compressed-input cap (8 GiB). A real `incus export` container
+/// backup of a desktop-flavoured OS regularly sits in the 3-6 GiB range
+/// compressed, so the pre-#1492 2 GiB cap was tighter than any real image
+/// could pass. The cap is now env-tunable via
+/// [`ENV_MAX_COMPRESSED_INPUT_BYTES`] so operators can raise it for unusual
+/// fleets or lower it on resource-constrained clusters. The bomb posture is
+/// preserved: the upper bound is still enforced before any bytes hit disk.
+const DEFAULT_MAX_COMPRESSED_INPUT_BYTES: u64 = 8 * 1024 * 1024 * 1024;
 
-/// Maximum total uncompressed bytes we tolerate in an extracted rootfs (10 GiB).
-/// Bounds decompression bombs that expand a small archive into a PVC-filling
-/// tree. Checked by walking the extracted tree after `tar` finishes.
-const MAX_EXTRACTED_BYTES: u64 = 10 * 1024 * 1024 * 1024;
+/// Default extracted-tree cap (40 GiB). Real container rootfs trees with
+/// debug packages and language SDKs can exceed the previous 10 GiB cap;
+/// likewise tunable via [`ENV_MAX_EXTRACTED_BYTES`].
+const DEFAULT_MAX_EXTRACTED_BYTES: u64 = 40 * 1024 * 1024 * 1024;
 
 /// Maximum number of filesystem entries we tolerate in an extracted rootfs.
-/// Bounds inode-exhaustion bombs (millions of tiny files).
-const MAX_EXTRACTED_ENTRIES: u64 = 2_000_000;
+/// Bounds inode-exhaustion bombs (millions of tiny files). Tunable via
+/// [`ENV_MAX_EXTRACTED_ENTRIES`]; default raised to 5M so real distro
+/// rootfs trees (Ubuntu desktop with `-dbg` packages can exceed 2M) pass.
+const DEFAULT_MAX_EXTRACTED_ENTRIES: u64 = 5_000_000;
+
+/// Environment variables operators can set to override the bomb caps.
+/// Values are parsed as `u64`; anything unparseable falls back to the
+/// default (and is logged the first time the value is read).
+const ENV_MAX_COMPRESSED_INPUT_BYTES: &str = "MAX_INCUS_SCAN_COMPRESSED_BYTES";
+const ENV_MAX_EXTRACTED_BYTES: &str = "MAX_INCUS_SCAN_EXTRACTED_BYTES";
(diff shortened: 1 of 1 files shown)


brandonrc:
Closing in favor of #1498 (opened first, same #1492 fix). Apologies for the duplicate work — my agent didn't see #1498 was already open before starting.

PR #1534, opened 2026-05-31
fix(oci): stream blob pulls through the storage backend instead of buffering

### LINKED ISSUE #1528: OCI blob pulls buffer the entire layer in heap — stream via get_stream()/Body::from_stream
### Problem

The OCI blob read path serves the whole object from memory: `handle_get_blob` does `storage.get(&storage_key)` (returns the entire object) then `.body(Body::from(data))`. Pulling a multi-GiB layer from object storage loads the full layer into the backend's heap. The virtual-repo local-serve branch and the manifest-blob serve site do the same.

This is the read-side mirror of the upload streaming implemented in #1448 for the write side. `get_stream()` is already on the `StorageBackend` trait and every backend (filesystem/s3/gcs/azure), so the fix is feasible.

### Pre-existing

Present on `main` with byte-identical serving logic; not introduced by #1448 (whose scope is upload/push streaming).

### Proposed fix

In `handle_get_blob` (CAS-hit and virtual-local branches) replace `storage.get()` + `Body::from(data)` with `storage.get_stream()` + `Body::from_stream`, sourcing `Content-Length` from the known `oci_blobs.size_bytes` column rather than `data.len()`. Apply to the other local-serve site too. Leave the upstream-proxy branches (`try_upstream_fetch` / `build_oci_proxy_response`) for a later pass. Mirrors the read-side streaming already done for GCS in #1431.

Sites: `backend/src/api/handlers/oci_v2.rs`.

Surfaced during review of #1448.

inspired-geek:
## Summary

`handle_get_blob` served CAS-hit and virtual-local blobs by reading the whole object into heap (`storage.get(&key)` → `Body::from(data)`), so pulling a multi-GiB layer from object storage loaded the entire layer into the backend's memory.

This streams the blob straight from the backend instead: `storage.get_stream(&key)` → `Body::from_stream(...)`, with `Content-Length` sourced from the authoritative `oci_blobs.size_bytes` column (rather than `data.len()`). Memory is now bounded per chunk. Both the migrated-CAS-hit branch and the virtual-repo local-member branch are converted.

Cloud backends already override `get_stream` for true streaming; the trait default falls back to `get`, so no backend regresses. The upstream-proxy fetch branches are left as-is. The OCI **push** path is handled separately (#1448).

Closes #1528.

  - `get_blob_streams_from_backend_with_size_bytes_content_length`: stores a blob + `oci_blobs` row, GETs it, and asserts 200, a `Content-Length` taken from `size_bytes`, and a body that round-trips through the streaming path.

## Test Checklist
- [x] Manually tested locally
- [x] No regressions in existing tests (full lib suite green; clippy/fmt clean; new-line coverage 87%)

## API Changes
- [ ] New endpoints have `#[utoipa::path]` annotations — N/A: OCI distribution endpoints are spec-defined, not part of the utoipa OpenAPI surface
- [ ] Request/response types have `#[derive(ToSchema)]` — N/A
- [x] N/A - no API changes (response bytes/headers unchanged; only the body is now streamed)


brandonrc (commented):
**Verdict: Approve.** The streaming conversion preserves header, status, and error-mapping parity with the buffered path. No blocking correctness, resource-leak, or integrity regressions. Two non-blocking notes worth a glance.

Notes (verified):

- Content-Length now sourced from `oci_blobs.size_bytes` instead of `data.len()` (`backend/src/api/handlers/oci_v2.rs:2432`, `:2473`). These match for any blob written through `handle_complete_upload`, where `size_bytes` is the verified byte count after the digest check (`2466-2492`). They only diverge if a stored object was truncated/rewritten out of band, a pre-existing data-integrity defect. Acceptable and standard for streaming registries, flagging only because it shifts the source of truth.
- Mid-stream errors surface after the 200 and Content-Length are committed. With `Body::from_stream`, a backend read failure produces a truncated body under an already-sent 200 rather than a clean 5xx. This is inherent to streaming and matches how OCI clients detect truncation (Content-Length shortfall). The pre-fetch failure case is still handled correctly: `get_stream(...)` returning `Err` falls through to the upstream-proxy branch exactly as the old `get()` did (`2439-2441`).

Verified non-issues: HEAD blob is a separate handler using `exists()` and is untouched (`1937-1946`). No Range/206 handling exists in either the old or new GET path, so no partial-content regression. Dropping the response body drops the fs `ReaderStream`/file handle and the S3 stream on client disconnect. Digest-verification parity holds: the buffered local/CAS-hit path never re-verified stored bytes either (verification is only on the upstream-proxy branch via `verify_digest_or_fall_through`), so streaming skips nothing the old path did. `use futures::StreamExt;` is added (`oci_v2.rs:20`), so the `.map(...)` calls compile.

Summary: clean, well-scoped streaming conversion of the two local-serve branches (CAS-hit and virtual-local) in `handle_get_blob`. Headers, status codes, error fall-through, and HEAD behavior all retain parity, and the trait default keeps non-streaming backends working. Approve.

PR #1540, opened 2026-06-01
feat(api): wire HTTP endpoint for ProxyService::invalidate_cache (closes #1539)

### LINKED ISSUE #1539: feat(api): wire HTTP endpoint for ProxyService::invalidate_cache (POST /repositories/:key/cache/invalidate?path=...)
## Problem Statement

`ProxyService::invalidate_cache` exists in the backend (`backend/src/services/proxy_service.rs`, around L1027) and is also covered by unit tests via the sister method `invalidate_cache_by_key` (path-traversal rejection, deletes both content + metadata blobs). However, **no HTTP handler wires it up**. Operators who need to evict a stale cached artifact from a remote (proxy) repository currently have to delete the underlying blob in object storage out-of-band (GCS / S3), which is awkward, easy to get wrong, and impossible to do without storage-level credentials.

## Proposed Solution

Wire up a new endpoint:

```
POST /api/v1/repositories/{key}/cache/invalidate?path=<artifact-path>
```

Behaviour:

- `key` (path) — repository key.
- `path` (query) — the artifact path to evict, identical to the path used by `GET /api/v1/repositories/{key}/artifacts/{path}`.
- Auth: requires authenticated user with `write` scope and `require_repo_access` on the target repo (mirrors `set_cache_ttl`).
- Repo type guard: only valid for `repo_type = Remote`. Other repo types return `400 Validation` (cache invalidation is meaningless on Local / Virtual repos — no proxy cache exists to evict).
- Storage guard: when `state.proxy_service` is `None` (e.g. storage backend not configured) returns `503 ServiceUnavailable` instead of crashing.
- On success, returns `200` with `{"repository_key": "...", "path": "...", "invalidated": true}`. The endpoint is idempotent — a second call for an already-evicted path still returns 200, matching the underlying `ProxyService::invalidate_cache` contract (it ignores delete-of-missing).
- `#[utoipa::path]` annotation so it appears in the generated OpenAPI spec / SDK.

The handler is a thin wrapper:

```rust
let proxy = state.proxy_service.as_ref().ok_or_else(|| {
    AppError::ServiceUnavailable("proxy service not configured".to_string())
})?;
proxy.invalidate_cache(&repo, &q.path).await?;
```

Path validation (rejection of `..` traversal, etc.) is already handled inside `cache_storage_key` and is covered by `test_invalidate_cache_by_key_rejects_invalid_path`. The new handler inherits that behaviour for free.

## Alternatives Considered

- **JSON body instead of query string** (`POST /:key/cache/invalidate` with `{"path": "..."}`). More consistent with `set_cache_ttl`, but the query form is simpler from the CLI / `curl` and is what the operator-facing UX is most likely to look like (paste a path, hit a button). Easy to accept either later if reviewers prefer the body form.
- **`DELETE /:key/cache?path=...`**. Semantically a delete, but the existing route shape under `/:key/cache-ttl` already uses `cache-*` as the noun; nesting `cache/invalidate` keeps cache-related operations under one prefix without colliding with `/:key/artifacts/*path` (which is the actual artifact resource).
- **Bulk invalidate by prefix**. Useful but a separate, larger discussion — out of scope for this issue.

## Use Case

As a repository admin operating a remote (proxy) repo, when upstream serves corrupt or wrong content under a known path (e.g. a tag was force-pushed, an upstream mirror went bad, a malware-scanner just flagged a cached artifact), I want to evict that single cache entry over the API so the next request re-fetches from upstream — without needing GCS / S3 credentials or a backend redeploy.

## Component

Backend / API

## Additional Context

- Existing method to wire: `ProxyService::invalidate_cache` (`backend/src/services/proxy_service.rs` L1027).
- Sibling endpoint to mirror style: `set_cache_ttl` / `get_cache_ttl` at `/{key}/cache-ttl`.
- Companion web-UI work tracked separately in [artifact-keeper-web](https://github.com/artifact-keeper/artifact-keeper-web) — an "Invalidate cache" action on the artifact-details dialog. The web change calls this endpoint.

## Pre-submission Checklist

- [x] Reviewed the docs to confirm there is no existing endpoint.

knowinglyAnonymous:
## Summary

Wires up an HTTP endpoint for the existing `ProxyService::invalidate_cache` method, so operators can evict a single cached artifact entry on a Remote (proxy) repository over the API instead of deleting blobs out-of-band in object storage.

Closes #1539.

```
POST /api/v1/repositories/{key}/cache/invalidate?path=<artifact-path>
```

The underlying `invalidate_cache` (`backend/src/services/proxy_service.rs:1027`) already existed and is covered by the storage-level tests on its sibling `invalidate_cache_by_key` (path-traversal rejection, deletes both content + metadata). Until now nothing called it from an HTTP handler, so the feature was effectively dark.

### Behaviour

- **Auth.** Mirrors `set_cache_ttl`: `require_auth` + `auth.require_scope("write")` + `require_repo_access`.
- **Repo type guard.** Only `Remote` (proxy) repos own a cache. Local / Virtual / Staging return `400 VALIDATION_ERROR` with `"cache invalidation is only supported on remote (proxy) repositories"`. Better than a silent 200 no-op.
- **Storage guard.** When `state.proxy_service` is `None` (storage backend not configured on this deployment), returns `503 SERVICE_UNAVAILABLE` instead of `unwrap`-panicking. Matches the `AppError::ServiceUnavailable` doc-comment guidance ("feature off" vs "server bug").
- **Idempotent.** Invalidating a path that was never cached (or was already evicted) still returns 200, matching the underlying `ProxyService::invalidate_cache` contract.
- **Path traversal.** Already rejected inside `cache_storage_key` (covered by the existing `test_invalidate_cache_by_key_rejects_invalid_path` service test); the handler inherits that for free.

### Why a query parameter and not a JSON body

Single string param, easier from `curl` / shell scripts (which is the operator-facing UX), and what #1539 specifies. Happy to switch to a JSON body in a follow-up if reviewers prefer that for consistency with `set_cache_ttl`.

### Why a separate route under `/:key/cache/...`

`/:key/cache-ttl` is already taken by the existing config endpoint, and `/:key/artifacts/*path` is the actual artifact resource (DELETE there removes the artifact row, not just the proxy cache entry). Nesting a new `cache/invalidate` segment keeps cache-related operations under one prefix without colliding with either.

- [x] N/A — this is `feat/*`, a new endpoint wiring an existing service method.

## Test Checklist

- [ ] Integration tests added/updated (if applicable) — N/A; service-layer behaviour is already covered by the existing `test_invalidate_cache_by_key_*` tests
- [ ] E2E tests added/updated (if applicable) — N/A; companion web change in [artifact-keeper-web#446](https://github.com/artifact-keeper/artifact-keeper-web/issues/446) will add a Playwright spec
- [x] Manually tested locally (`cargo check`, `cargo clippy --lib --tests -- -D warnings`, `cargo fmt --all -- --check`, full `api::handlers::repositories` test suite — 288 passing)

New tests in `api::handlers::repositories::tests`:

- `test_invalidate_cache_query_deserialization` — pins the `path` field name.
- `test_invalidate_cache_query_rejects_missing_path` — required-ness regression.
- `test_invalidate_cache_response_serialization` — pins response shape.
- `invalidate_cache_handler_guards_repo_type_and_proxy_service` — structural test mirroring the `set_cache_ttl_type_check_runs_before_range_check` pattern. Asserts that BOTH the `repo.repo_type != RepositoryType::Remote` guard AND the `AppError::ServiceUnavailable` proxy-service guard run BEFORE `proxy.invalidate_cache(`. Pins the contract this PR adds against future refactors.

## API Changes

- [x] New endpoint has `#[utoipa::path]` annotation
- [x] Request/response types have `#[derive(ToSchema)]` (and `IntoParams` for the query)
- [x] OpenAPI spec validates: `cargo test --lib test_openapi_spec_is_valid` ✓
- [ ] Migration is reversible (if applicable) — N/A, no schema changes
- [ ] N/A - no API changes

## Companion work

- Web UI button on the artifact-details dialog: [artifact-keeper-web#446](https://github.com/artifact-keeper/artifact-keeper-web/issues/446) (uses this endpoint via `apiFetch` until the SDK is regenerated).

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -216,6 +216,11 @@ pub fn router() -> Router<SharedState> {
         )
         // Cache TTL configuration for proxy/remote repositories
         .route("/:key/cache-ttl", put(set_cache_ttl).get(get_cache_ttl))
+        // Cache invalidation for a specific path on a Remote (proxy) repository
+        // (#1539). POST keeps the action explicit; the underlying
+        // `ProxyService::invalidate_cache` is idempotent so a second call for
+        // an already-evicted path still returns 200.
+        .route("/:key/cache/invalidate", post(invalidate_cache))
         // Routing rules for path rewriting on remote repositories
         .route(
             "/:key/routing-rules",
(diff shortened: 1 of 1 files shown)


brandonrc (commented):
**Verdict: Approve.** No critical issues found. Authorization, path handling, correctness, and error handling are all sound and consistent with the existing `set_cache_ttl` endpoint.

What I verified:

- Authorization (`backend/src/api/handlers/repositories.rs:715-719`): the handler runs `require_auth` (rejects anonymous), `auth.require_scope("write")`, and `require_repo_access` (enforces `allowed_repo_ids` scoping). Identical to `set_cache_ttl`. No anonymous or read-only eviction path. Context note (not a blocker, pre-existing across this module): non-API-token JWT sessions pass `require_scope` unconditionally, so any authenticated user with repo access can evict. That matches the established convention for the neighboring repo-mutating endpoints.
- Path traversal / key injection: `proxy.invalidate_cache` (`proxy_service.rs:1027-1036`) derives both keys via `cache_storage_key` then `validate_cache_path` (`1302-1347`), which rejects `..`/`.` segments, NUL, backslash, empty segments, and over-length keys before any delete. The repo key comes from the looked-up `repo.key`, not raw input. Invalid paths return `AppError::Validation` (400), not a panic.
- Correctness / blob removal: `invalidate_cache` deletes both the `__content__` blob and the `__cache_meta__.json` metadata via the same key derivation used on the write path, so it removes the on-disk/S3 cached blob, not just DB state. Idempotent (`let _ = delete(...)`).
- Error handling: non-Remote repos return 400 before touching storage (`728-731`); missing `proxy_service` returns 503 via `ok_or_else` (`738-741`); missing repo returns 404. No panic/500 paths.
- Route conflicts: `/:key/cache/invalidate` (`220`) does not collide with `/:key/cache-ttl`, `/:key/artifacts/*path`, or `/:key/routing-rules`. Pure addition. OpenAPI registration and schema derives present.

Summary: clean wiring of an existing, already-tested service method behind an HTTP endpoint that faithfully copies the auth and repo-type-guard pattern from `set_cache_ttl`. Approve.

PR #1542, opened 2026-06-01
feat(api): expose per-artifact cache metadata on GET /:key/artifacts/:path (closes #1541)

### LINKED ISSUE #1541: feat(api): expose per-artifact cache metadata (cache_expires_at, cache_cached_at) on GET /:key/artifacts/:path
## Problem Statement

`GET /api/v1/repositories/{key}/artifacts/{path}` returns the static artifact metadata (id, path, size, sha-256, content type, download count, created_at, ...) but **does not surface any proxy-cache state**. For a Remote (proxy) repository, the cache layer maintains its own metadata blob (`CacheMetadata { cached_at, expires_at, upstream_etag, storage_etag, content_type, size_bytes }` in `services/proxy_service.rs`) but the field that operators care about most — `expires_at` — is not visible anywhere over the API. Operators cannot tell when a cached entry will be re-validated against upstream without manually computing it from the configured TTL and the artifact's `created_at`, which is wrong any time the cache was refreshed without re-creating the artifact row.

## Proposed Solution

Surface the live cache metadata on the existing per-artifact metadata endpoint, so a single round-trip gives the UI everything it needs to render an artifact details panel:

```
GET /api/v1/repositories/{key}/artifacts/{path}
{
  ... existing ArtifactResponse fields ...
  "cache_expires_at": "2026-06-02T11:55:00Z",   // optional, only for Remote repos with a cache hit
  "cache_cached_at":  "2026-06-01T11:55:00Z"    // optional, ditto
}
```

Implementation outline:

- Promote the existing private `ProxyService::load_cache_metadata` (or add a thin `pub` wrapper, e.g. `pub async fn get_cache_metadata(&self, repo_key, path) -> Result<Option<CacheMetadata>>`) so handlers can read the metadata blob without re-implementing the key derivation.
- Extend `ArtifactResponse` with two `Option<DateTime<Utc>>` fields, `#[serde(skip_serializing_if = "Option::is_none")]`. Local / Virtual / Staging repos always serialise as `None`, so the response stays clean.
- In `get_artifact_metadata` only (not the listing endpoint — see below), when `repo.repo_type == Remote` AND `state.proxy_service.is_some()`, attempt to load the metadata blob and populate the new fields. Failures (blob missing, deserialise error, transient storage error) leave the fields `None` rather than fail the whole request: missing cache metadata is a normal state for an artifact that has never been fetched through the proxy, and we do not want to convert it into a 5xx.

### Cost guardrails

The per-artifact handler today touches the database (repo lookup + artifact lookup) but does **not** touch object storage. This change adds **at most one storage GET**, gated on `repo.repo_type == Remote` AND `state.proxy_service.is_some()`. Local / Virtual / Staging requests stay zero-storage as today.

The artifact **listing** endpoint (`GET /:key/artifacts`) is intentionally NOT modified — fanning out one storage GET per artifact across a paginated listing would multiply the cost on large repos. Listings will continue to omit cache fields, and the UI's listing view only needs the existing data anyway. The new fields land only on the single-artifact endpoint where the marginal cost is one GET per page view.

## Alternatives Considered

- **New dedicated endpoint** (`GET /api/v1/repositories/{key}/cache?path=...`). Cleaner separation, but forces the artifact-details UI into two sequential round-trips per open. Reviewers may prefer this for cache-purity reasons; happy to switch if so.
- **Populate cache fields on the listing endpoint as well.** Rejected for the cost reason above. Could be added later behind an opt-in `?include=cache` query parameter if a real use case shows up.
- **Recompute `expires_at` client-side** from `cache_ttl_seconds` plus `created_at`. Wrong: `created_at` is the artifact-row creation time, not the cache-write time, and the cache may have been refreshed any number of times since.

## Use Case

As a repo admin browsing a remote-proxied artifact, I want to see "expires in 4 hours" (or "expired 12 minutes ago, will re-fetch on next download") directly on the artifact details panel, so I can decide whether to wait for the natural refresh or hit the new "Invalidate cache" action shipped in artifact-keeper-web#446. Without this field, the only signal the UI has is `created_at`, which is not the same thing.

## Component

Backend / API

## Additional Context

- Companion web change tracked separately: artifact-keeper-web issue (will be linked once filed) — surfaces these new fields in the artifact-details dialog as a relative-time row.
- Related shipped work: `POST /:key/cache/invalidate` ([#1539](https://github.com/artifact-keeper/artifact-keeper/issues/1539) / [#1540](https://github.com/artifact-keeper/artifact-keeper/pull/1540)) already gives operators the eviction action. This issue gives them the visibility piece that pairs with it.

## Pre-submission Checklist

- [x] Reviewed the OpenAPI spec to confirm the fields are not already exposed.

knowinglyAnonymous:
## Summary

Surfaces proxy-cache freshness on the per-artifact metadata endpoint so the web client can render "expires in 4 hours" / "expired 12 minutes ago" without a separate round-trip and without operators having to compute it from `created_at` + the configured TTL.

Closes #1541. Companion web work tracked in artifact-keeper-web#449.

## Behavior

```
GET /api/v1/repositories/pypi-remote/artifacts/requests/requests-2.31.0-py3-none-any.whl
{
  ... existing ArtifactResponse fields ...
  "cache_cached_at":  "2026-06-01T11:55:00Z",   // omitted when None
  "cache_expires_at": "2026-06-02T11:55:00Z"    // omitted when None
}
```

The two new fields are `Option<DateTime<Utc>>` with `#[serde(skip_serializing_if = "Option::is_none")]` — for non-Remote repos and Remote repos that have not yet been hit through the proxy, the wire shape is unchanged. The web client doesn't need to handle `null` vs missing.

## Cost guardrails

The handler's existing path touches the database (repo lookup + artifact lookup) but does NOT touch object storage. This change adds **at most one storage GET**, gated on `repo.repo_type == RepositoryType::Remote` AND `state.proxy_service.is_some()`. Local / Virtual / Staging requests stay zero-storage as today.

The artifact LISTING endpoint is intentionally NOT modified — fanning out one storage GET per artifact across a paginated listing would multiply the cost on large repos. The `build_artifact_response` and `build_secondary_artifact_responses` helpers used by listings explicitly default the new fields to `None` so the wire shape stays identical and the cost stays flat.

The metadata read is tolerant of failure: a missing or unreadable metadata blob is a normal state (e.g. a Remote-typed artifact direct-uploaded but never proxy-fetched), so any error or `Ok(None)` collapses to `None` on both fields rather than failing the whole metadata response. The existing 200 / 404 contract for the endpoint is preserved.

## Technical Choices

- **Extended `ArtifactResponse`** rather than adding a new endpoint. Single round-trip from the artifact-details dialog; reviewer concern about cache vs static-metadata coupling is addressed by the `skip_serializing_if = "Option::is_none"` shape (response stays clean for non-cache repos) and by keeping the listing endpoint unchanged.
- **New `pub async fn ProxyService::get_cache_metadata(repo_key, path)`** as a thin wrapper around the existing private `load_cache_metadata`. Encapsulates the metadata-key derivation so handlers don't have to know `proxy-cache/{repo_key}/{trimmed}/__cache_meta__.json`.
- **Type guard before storage guard**: `repo.repo_type == Remote` short-circuits the read on Local / Virtual / Staging without even consulting `state.proxy_service`. A reviewer worried about the cost-control invariant can confirm this with a single `if`-statement scan.

## Test Checklist

- [x] **Unit tests added/updated.** Three relevant cases in `tests` mod:
  - `test_artifact_response_serialization` — pins that absent cache fields serialize cleanly (no `cache_cached_at` / `cache_expires_at` keys in JSON).
  - `test_artifact_response_serialization_with_cache_metadata` — pins the wire shape when populated: ISO-8601 strings keyed `cache_cached_at` / `cache_expires_at`.
  - `get_artifact_metadata_populates_cache_fields_only_for_remote` — structural assertion against the handler source: both gates (`RepositoryType::Remote` and `state.proxy_service.as_ref()`) and the `get_cache_metadata` call are present, and both response fields are populated. The test's pattern matching is deliberately tolerant of `cargo fmt` reformatting (matches on individual identifiers, not multi-line strings) — the same approach we used for the proxy-service guard test in #1540.
- [x] **OpenAPI spec validated** by `utoipa`'s build-time validation: `cargo check --lib` passes; the new fields appear automatically because `ArtifactResponse` already derives `ToSchema`.
- [x] **Manually verified** via `cargo fmt --all -- --check` (clean), `cargo clippy --lib --tests -- -D warnings` (clean), `cargo test --lib test_artifact_response` (3 pass; 2 new + 1 updated existing), `cargo test --lib get_artifact_metadata_populates` (1 pass).
- [x] No regressions in existing tests.

## API Changes

- [x] **`ArtifactResponse`** gains two optional fields: `cache_cached_at` and `cache_expires_at`. Both are `Option<DateTime<Utc>>` with `#[serde(skip_serializing_if = "Option::is_none")]`. Backward compatible — existing callers that didn't request the fields continue to see the same wire shape.
- [x] **`ProxyService::get_cache_metadata(repo_key, path)`** is a new public method. The existing `load_cache_metadata` stays private — `get_cache_metadata` is the supported handler-facing surface.
- [x] No existing fields removed or renamed.

## Companion Work

- artifact-keeper-web#449 — surfaces these fields as relative-time rows on the artifact details dialog. Will follow this PR; it depends on the SDK regenerating with the new optional fields.
- Related shipped work: `POST /:key/cache/invalidate` ([#1539](https://github.com/artifact-keeper/artifact-keeper/issues/1539) / [#1540](https://github.com/artifact-keeper/artifact-keeper/pull/1540)) — operators can now both *see* the cache state and *act* on it.

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -1309,6 +1309,19 @@ pub struct ArtifactResponse {
     pub created_at: chrono::DateTime<chrono::Utc>,
     #[schema(value_type = Option<Object>)]
     pub metadata: Option<serde_json::Value>,
+    /// When the proxy cache entry for this artifact was last written.
+    /// Only populated for Remote (proxy) repositories whose proxy service is
+    /// configured AND that have a cache-metadata blob for this path. None
+    /// for Local / Virtual / Staging repos and for Remote repos whose cache
+    /// hasn't been populated yet (e.g. an artifact that exists as a DB row
+    /// from a direct upload but has never been fetched through the proxy).
+    /// (#1541)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cache_cached_at: Option<chrono::DateTime<chrono::Utc>>,
+    /// When the proxy cache entry for this artifact will expire and be
+    /// re-validated against upstream. Same gating as `cache_cached_at`. (#1541)
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub cache_expires_at: Option<chrono::DateTime<chrono::Utc>>,
 }
 
 #[derive(Debug, Serialize, ToSchema)]
@@ -1693,6 +1706,12 @@ fn build_artifact_response(
         download_count,
         created_at: artifact.created_at,
         metadata: None,
+        // Cache metadata is surfaced only by the per-artifact metadata
+        // endpoint to avoid fanning out a storage GET per artifact in
+        // listings (#1541). Helpers used by listings leave these as None;
+        // get_artifact_metadata populates them after the fact.
+        cache_cached_at: None,
+        cache_expires_at: None,
     }
 }
 
@@ -1737,6 +1756,8 @@ fn expand_maven_secondary_files(
             download_count: 0,
             created_at: artifact.created_at,
             metadata: None,
+            cache_cached_at: None,
+            cache_expires_at: None,
         });
     }
     out
@@ -2423,6 +2444,29 @@ pub async fn get_artifact_metadata(
         let downloads = artifact_service.get_download_stats(artifact.id).await?;
         let metadata = artifact_service.get_metadata(artifact.id).await?;
 
+        // #1541: surface proxy cache freshness on Remote repos so the UI
+        // can render "expires in 4 hours" / "expired" without a separate
+        // round-trip. Single storage GET, gated on `repo.repo_type ==
+        // Remote` AND `state.proxy_service.is_some()`. Tolerant of failure:
+        // a missing or unreadable metadata blob is a normal state for an
+        // artifact that has never been fetched through the proxy (e.g.
+        // direct-uploaded into a Remote repo, edge case but observed),
+        // so any error or `Ok(None)` collapses to None on both fields
+        // rather than failing the whole metadata response.
+        let cache_meta = if repo.repo_type == RepositoryType::Remote {
+            if let Some(proxy) = state.proxy_service.as_ref() {
+                proxy
+                    .get_cache_metadata(&key, &artifact.path)
+                    .await
+                    .ok()
+                    .flatten()
+            } else {
+                None
+            }
+        } else {
+            None
+        };
+
         return Ok(Json(ArtifactResponse {
             id: artifact.id,
             repository_key: key,
@@ -2435,6 +2479,8 @@ pub async fn get_artifact_metadata(
             download_count: downloads,
             created_at: artifact.created_at,
             metadata: metadata.map(|m| m.metadata),
+            cache_cached_at: cache_meta.as_ref().map(|m| m.cached_at),
+            cache_expires_at: cache_meta.as_ref().map(|m| m.expires_at),
         })
         .into_response());
     }
@@ -2649,6 +2695,10 @@ pub async fn upload_artifact(
         download_count: downloads,
         created_at: artifact.created_at,
         metadata: metadata_json,
+        // Just-uploaded artifacts have no proxy cache state yet -- the
+        // cache is populated lazily on the first proxy fetch.
+        cache_cached_at: None,
+        cache_expires_at: None,
     }))
 }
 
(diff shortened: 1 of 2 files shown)


brandonrc (commented):
**Verdict: Approve.** No critical issues. The change is correctly gated, leaks nothing sensitive, adds at most one storage GET on the single-artifact path, and is backward compatible. One non-blocking correctness note below.

I verified each focus area against the code:

- Freshness/expiry math: the backend emits only raw RFC3339 UTC timestamps (`cache_cached_at`, `cache_expires_at`) mapped directly with no arithmetic. The relative-time rendering ("expires in / expired N ago") is client-side (web#449), so there is no signed-duration/overflow/timezone math in this PR to get wrong.
- Information disclosure: `CacheMetadata` (`proxy_service.rs:448-478`) carries `upstream_etag`, `storage_etag`, `checksum_sha256`, etc., but the handler maps only `cached_at` and `expires_at`. No upstream URLs, credentials, or storage keys exposed. Auth reuses the pre-existing `require_visible(&repo, &auth)` gate; the two new timestamps are no more sensitive than the already-returned `created_at`.
- N+1 / unbounded work: `get_cache_metadata` is one `load_cache_metadata` then one `storage.get`. The listing helpers default both fields to `None`, so listings stay zero extra storage calls.
- Null/None for non-proxy repos: double-guarded by `repo.repo_type == Remote` then `state.proxy_service.as_ref()`, with `.ok().flatten()` collapsing both `Err` and `Ok(None)` to `None`. No `unwrap`/`expect`, no panic, no 500 for Local/Virtual/Staging or unconfigured-proxy deployments.
- Breaking change: both fields are `Option<DateTime<Utc>>` with `skip_serializing_if = "Option::is_none"`, and the existing serialization test asserts absence when `None`. No fields removed or renamed.

Non-blocking note: for npm-family repos the handler resolves the DB row via `lookup_path_candidates` (URL shape `<name>/-/<name>-<version>.tgz` to stored shape `<name>/<version>/<name>-<version>.tgz`) and passes the stored `artifact.path` to `get_cache_metadata`, while the proxy writes cache metadata keyed by the fetch path (`cache_metadata_key`). When those differ, the lookup misses and the new cache fields stay `None` even when a valid entry exists. That only suppresses the new UI rows, it does not error, so it is a follow-up rather than a blocker.

Summary: clean, well-scoped addition. The cost guardrail, the None-collapsing failure tolerance, and the backward-compatible wire shape are all implemented as described. Approve.

PR #1549, opened 2026-06-01
fix(maven): skip DB checksum lookup for remote repos and remote virtu…

### LINKED ISSUE #1599: Maven remote/virtual checksum requests do a failing DB lookup before proxying upstream
## Bug

In the Maven `download()` handler, a checksum request (`*.md5` / `*.sha1`) for a **Remote** repo, or for a **remote member of a Virtual** repo, calls `serve_computed_checksum` first. That helper computes the checksum from a row in the `artifacts` table, but proxy-cached artifacts are intentionally stored in the proxy cache and NOT inserted into `artifacts` (#1280). So the DB lookup always misses, wasting a query before the code falls through to proxying the checksum from upstream.

For the virtual case the member loop tried `serve_computed_checksum` for every member (including remote ones) before attempting the upstream proxy, which is both wasteful and, depending on ordering, can mask the correct remote result.

## Expected

For Remote repos and remote virtual members, skip `serve_computed_checksum` and proxy the checksum directly from upstream. Only Local/Staging repos (and local/staging members) should compute checksums from stored artifacts.

## Fix

Gate `serve_computed_checksum` behind `repo_type in (Local, Staging)` on the direct path, and split the virtual member loop so remote members proxy upstream while local/staging members compute locally.

Addressed by PR #1549. Related to the broader 'remote Maven should not do local lookups before proxy-fetch' theme in #1547.

Dreamacro:
Closes #1599

## Summary
Fix Maven checksum lookup (`*.md5` / `*.sha1`) for remote repos and virtual
repos with remote members. Remote repos are pull-through caches — their
artifacts are not stored in the local `artifacts` table, so
`serve_computed_checksum` always fails. The handler now:
- **Direct remote repo**: wraps `serve_computed_checksum` behind a
  `Local | Staging` guard; Remote repos fall through to `proxy_fetch` immediately.
- **Virtual repo members**: branches the member loop — Remote members call
  `proxy_fetch` directly (skipping the DB lookup), Local/Staging members call
  `serve_computed_checksum` as before.
## API Changes
- [ ] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/maven.rs
+++ b/backend/src/api/handlers/maven.rs
@@ -1007,17 +1007,21 @@ async fn download(
             }
         }
 
-        // Otherwise try to compute from a locally-stored artifact
-        if let Ok(response) = serve_computed_checksum(
-            &state,
-            repo.id,
-            &repo.storage_location(),
-            base_path,
-            checksum_type,
-        )
-        .await
-        {
-            return Ok(response);
+        // Compute checksum from locally-stored artifact (Local/Staging only).
+        // Remote repos cache artifacts in the proxy cache, not the `artifacts`
+        // table, so the DB lookup inside serve_computed_checksum always fails.
+        if checksum_compute_eligible(&repo.repo_type) {
+            if let Ok(response) = serve_computed_checksum(
+                &state,
+                repo.id,
+                &repo.storage_location(),
+                base_path,
+                checksum_type,
+            )
+            .await
+            {
+                return Ok(response);
+            }
         }
 
         // Fallback: proxy the checksum file from upstream for remote repos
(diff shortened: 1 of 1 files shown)


brandonrc (commented):
**Verdict: Approve.** Behavior-preserving cleanup, no critical issues found.

I verified the central premise against the code. Proxy-cached artifacts are deliberately excluded from the `artifacts` table (enforced by the `cache_artifact MUST NOT INSERT INTO artifacts` guard at `backend/src/services/proxy_service.rs:5808`, ref #1278), and `serve_computed_checksum` queries only that table (`backend/src/api/handlers/maven.rs:1415-1429`). So for a Remote repo or Remote virtual member, the pre-PR call to `serve_computed_checksum` was guaranteed to return `NotFound` and fall through to `proxy_fetch`. Gating it behind `Local | Staging` (lines 991-1003) is therefore behavior-preserving and simply avoids a wasted DB round trip.

Things I checked and found safe:

- No regression for Local/Staging repos. The direct-repo path runs `serve_computed_checksum` only when `repo_type` is `Local || Staging` (`maven.rs:991`). The only excluded types are `Remote` (correctly proxied) and `Virtual` (handled by the virtual branch at line 1036).
- Stored checksums still served for all types. The stored-file lookup (`maven.rs:982`) and SNAPSHOT resolved-checksum lookup (`992-1004`) run before any repo-type branching, so a remote repo with a locally deployed `.sha1`/`.md5` still serves it. No stale/incorrect-checksum risk.
- Virtual member loop equivalence. Old code called `serve_computed_checksum` for every member then proxied only Remote ones; for Remote members that call always failed (#1278), so dropping it changes nothing. Local/Staging/Virtual members retain the DB-compute path. Nested-virtual members are not recursively resolved in either version, so no new gap.
- No new SSRF or path-traversal surface. The `proxy_fetch` calls and `path` value are identical to the pre-PR fallback; the diff only reorders existing calls and adds a type guard.

Summary: small, correct, reduces an unnecessary DB query on the remote checksum path. Approve.

Issue #113, opened 2026-02-12
Protobuf / Buf Schema Registry format support


brandonrc:
## Summary

Feature request from community discussion: https://github.com/orgs/artifact-keeper/discussions/111

Add support for protobuf modules/schemas, compatible with Buf Schema Registry (BSR).

### Completed
- [x] Database models — uses existing `artifacts` + `artifact_metadata` tables (#114)
- [x] Service layer — business logic in handler (#115)
- [x] Format handler — `backend/src/formats/protobuf.rs` with tests (#116)
- [x] API endpoints — 10 Connect RPC endpoints at `/proto/` (#117)
- [x] E2E tests — `test-protobuf.sh` in native tests (#118)

### Remaining
- [ ] Web UI: Dedicated protobuf module browser (artifact-keeper-web#12)
- [ ] Documentation on artifact-keeper-site (artifact-keeper-site#2)

Bead: `ak-y4f`

PR #1557, opened 2026-06-01
npm: persist and serve dist-tags; derive `latest` by semver not recency

### LINKED ISSUE #1543: npm: only `latest` dist-tag is served (custom dist-tags unsupported), and `latest` is picked by recency not semver
## Summary

The npm packument AK serves only ever contains a single `latest` dist-tag, and `latest` is chosen by **recency (most recently created version)** rather than by semver. Two consequences:

1. **Custom dist-tags are unsupported.** Tags like `next`, `beta`, `canary`, `nightly` cannot be published, listed, set, or installed. `npm install pkg@next` (or `npm pack pkg@next`) fails with `ETARGET` even when the package and its versions exist and are served correctly.
2. **`latest` can point at a prerelease.** If a prerelease version (e.g. `2.0.0-rc.1`) is published or imported *after* a stable release (e.g. `1.1.0`), `latest` resolves to the prerelease. Per npm convention, `latest` should be the highest non-prerelease semver unless explicitly set.

This is especially impactful for registries that imported existing packages from another registry: any package that relied on a non-`latest` dist-tag (common for canary/nightly CI pipelines) breaks for consumers, and `latest` may silently point at a prerelease.

## Where it comes from

`backend/src/api/handlers/npm.rs`, in `build_npm_metadata_response`:

```rust
// versions are iterated in the order returned by the query...
versions.insert(version.clone(), version_obj);
latest_version = Some(version);     // overwritten every iteration
}

let dist_tags = serde_json::json!({
    "latest": latest_version.unwrap_or_default()   // only ever "latest"
});
```

and the source query (`fetch_npm_artifacts`) is `... ORDER BY a.created_at ASC`. So:

- `dist-tags` is hardcoded to a single `latest` key — no other tag can ever appear.
- `latest` ends up being the **last-iterated = most-recently-created** version, not the highest semver.

There is also **no dist-tags endpoint** registered (`GET`/`PUT`/`DELETE /npm/{repo}/-/package/{pkg}/dist-tags`), so `npm dist-tag ls/add/rm` and `npm publish --tag <x>` have nothing to talk to.

## Reproduction

Against any npm repo on AK:

```bash
# publish a stable release, then a prerelease under a custom tag
npm publish            # @acme/widget 1.0.0   -> dist-tag latest
npm publish            # @acme/widget 1.1.0   -> dist-tag latest
npm publish --tag next # @acme/widget 2.0.0-rc.1 -> dist-tag next (intended)

# 1) custom tag is gone
npm view @acme/widget dist-tags
# observed: { latest: '2.0.0-rc.1' }       # no `next`, AND latest is the prerelease
# expected: { latest: '1.1.0', next: '2.0.0-rc.1' }

npm install @acme/widget@next
# npm error code ETARGET
# npm error notarget No matching version found for @acme/widget@next

# 2) dist-tags API is absent
npm dist-tag ls @acme/widget
# 404 from GET /npm/{repo}/-/package/@acme%2fwidget/dist-tags
```

The packument `GET` itself returns `200` with all versions present — only the `dist-tags` object is wrong/incomplete.

## Expected behavior (npm registry spec)

- The packument `dist-tags` object should contain **all** tags the package has, not just `latest` (see the npm registry API: packument `dist-tags`, and the dist-tags endpoints).
- `npm publish --tag <tag>` should record `<tag>`; `npm publish` with no tag should update `latest`.
- `latest` should be the highest **non-prerelease** semver (or the explicitly-set `latest`), never auto-set to a prerelease by recency.
- Implement the dist-tags endpoints so `npm dist-tag ls/add/rm` work:
  - `GET    /npm/{repo}/-/package/{pkg}/dist-tags`
  - `PUT    /npm/{repo}/-/package/{pkg}/dist-tags/{tag}`
  - `DELETE /npm/{repo}/-/package/{pkg}/dist-tags/{tag}`

## Suggested fix

1. Persist dist-tags. The publish `PUT` body already carries a `dist-tags` object (npm sends the tag being published); store it (e.g. a `dist_tags` table or a column keyed by package) instead of discarding it.
2. In `build_npm_metadata_response`, emit the stored dist-tags map rather than a synthesized single `latest`.
3. Derive/repair `latest` as the max non-prerelease semver when no explicit `latest` is recorded, rather than from `created_at` order.
4. Add the three dist-tags routes.

Happy to help with a PR if the approach above sounds right.

ThaSami:
## Summary

Fixes #1543.

npm packuments AK serves only ever contained a single `latest` dist-tag, and `latest` was chosen by **recency** (most-recently-created version) rather than semver. So:

1. **Custom dist-tags were unsupported** — `next`/`beta`/`canary`/`nightly` could not be published, listed, set, or installed (`npm install pkg@next` → `ETARGET`), and there were no dist-tags endpoints for `npm dist-tag ls/add/rm`.
2. **`latest` could point at a prerelease** — if a prerelease was published/imported after a stable release, `latest` resolved to the prerelease.

This is especially painful for registries that imported packages from another registry (any package relying on a non-`latest` tag breaks) and for CI pipelines that pin canary/nightly tags.

## What this PR does

1. **Persist dist-tags** — new migration `114_packages_dist_tags.sql` adds a `dist_tags JSONB NOT NULL DEFAULT '{}'` column to `packages` (already one row per `(repository_id, name)` since #113). `parse_npm_publish_payload` now captures the publish body's `dist-tags`, and `publish_package` merges them into that column (`dist_tags || $patch`, so the published tag wins).
2. **Emit the stored map** — `build_npm_metadata_response` takes the stored dist-tags and emits them verbatim instead of a synthesized single `latest`.
3. **Correct `latest`** — when no explicit `latest` is recorded (or it no longer resolves to a known version), `derive_latest_version` picks the **highest non-prerelease semver** (falls back to highest overall when every version is a prerelease, then to most-recent when nothing parses as semver).
4. **dist-tags endpoints** so `npm dist-tag ls/add/rm` work:
   - `GET    /npm/{repo}/-/package/{pkg}/dist-tags`
   - `PUT    /npm/{repo}/-/package/{pkg}/dist-tags/{tag}`
   - `DELETE /npm/{repo}/-/package/{pkg}/dist-tags/{tag}`

## Design notes

- **No new dependency.** Rather than pull in the `semver` crate, `latest` derivation uses a small, unit-tested comparator over the numeric `major.minor.patch` core (build metadata stripped, a `-prerelease` suffix excludes a version from the stable candidates).
- **Runtime-checked queries** (`sqlx::query`) for the new DB access — consistent with existing code and avoids regenerating the offline `.sqlx` cache.
- **`GET dist-tags` delegates to the packument path**, so it resolves correctly for local, virtual (member walk) and remote (upstream proxy) repos, then returns just the `dist-tags` object.
- **`PUT`/`DELETE`** require the `write` scope and a hosted (local/staging) repo; `PUT` validates the target version exists; `latest` cannot be deleted.
- The packument `GET` already returned all versions — only the `dist-tags` object changes shape, so this is backward-compatible for clients that only read `latest`.

## Tests

- 6 pure-function tests for `derive_latest_version` (stable over prerelease, numeric ordering, all-prerelease fallback, non-semver fallback, build-metadata stripping, empty input).
- 1 DB-backed end-to-end test (`#[tokio::test]` via the `test_db_helpers` fixture): publishes `1.0.0`, `1.1.0`, then `2.0.0-rc.1` with a custom `next` tag, and asserts the packument serves `next: 2.0.0-rc.1` **and** `latest: 1.1.0` (the stable version, not the most-recently-created prerelease).

## Verification (`npm` against a hosted repo)

```
npm publish              # widget 1.0.0          -> latest
npm publish              # widget 1.1.0          -> latest
npm publish --tag next   # widget 2.0.0-rc.1     -> next

npm view widget dist-tags
# { latest: '1.1.0', next: '2.0.0-rc.1' }        # was: { latest: '2.0.0-rc.1' }
npm install widget@next  # resolves 2.0.0-rc.1   # was: ETARGET
npm dist-tag ls widget   # latest: 1.1.0 / next: 2.0.0-rc.1   # was: 404
```


ThaSami:
@brandonrc please review

brandonrc (commented):
**Verdict: Request changes.** The PR is built on a schema assumption that does not hold: `packages` is one row per `(repository_id, name, version)`, not per `(repository_id, name)`. As written this breaks dist-tag persistence for any package with more than one version, and the tests pass only because they seed a single version row.

### Critical findings

**1. Wrong premise: `packages` is keyed by `(repository_id, name, version)`, so dist-tags are stored per version, not per package**

- `backend/migrations/019_builds_packages.sql:40` defines `UNIQUE(repository_id, name, version)`. One row per version.
- `backend/src/services/package_service.rs:47` upserts with `ON CONFLICT (repository_id, name, version)`, confirming a new row per published version.
- The PR (and `migrations/114_packages_dist_tags.sql:9-10`) assert `packages` is one row per `(repository_id, name)` "since migration #113." That migration does not exist (highest pre-PR migration is `111_drop_webhooks_url_repo_unique.sql`), and the constraint says otherwise.

Concrete breakage, since `dist_tags JSONB` now lives on a per-version row but the code treats it as per-package:
- `fetch_npm_dist_tags`: `SELECT dist_tags FROM packages WHERE repository_id = $1 AND name = $2` with `.fetch_optional(...)`. For a package with 2+ versions this matches multiple rows, sqlx returns a multiple-row error, the code swallows it with `.ok().flatten()`, and returns an empty map. Persisted dist-tags silently disappear once a second version exists.
- Publish-time `UPDATE packages SET dist_tags = dist_tags || $1::jsonb WHERE repository_id = $2 AND name = $3` writes the tag onto every version row.
- `dist_tags_put` / `dist_tags_delete` have the same `WHERE repository_id AND name` clause, fanning writes across all version rows.

This is masked by the tests: `seed_artifact` (`test_db_helpers.rs:300`) only inserts into `artifacts`, and the new DB tests manually insert exactly one `packages` row, so `fetch_optional` sees a single row and passes. Real publishes go through `try_create_or_update_from_artifact` and create N rows.

Fix: store dist-tags somewhere genuinely per package, either a dedicated `npm_dist_tags (repository_id, name, tags JSONB)` table with `UNIQUE(repository_id, name)`, or read/write a single canonical row deterministically. The read must not use `fetch_optional` against a multi-row result.

**2. Normal `npm publish` persists `latest` at the just-published version, defeating semver derivation**

`parse_npm_publish_payload` captures the publish body's `dist-tags`, which on a plain `npm publish` is `{"latest": "<this version>"}`. `publish_package` merges it verbatim. In `build_npm_metadata_response`, `derive_latest_version` only runs when the stored `latest` does not resolve, but a stored `latest` from a prerelease publish does resolve, so it is kept. Sequence `publish 1.1.0` then `publish 2.0.0-rc.1` (no `--tag`, which some CI does) persists `latest=2.0.0-rc.1` and the semver fallback never fires. That is the exact bug #1543 targets, reintroduced through the publish path. Consider not persisting `latest` blindly from the publish body, or recomputing `latest` by semver when the published version is a prerelease.

### Summary

The semver comparator in `derive_latest_version` looks correct (stable over prerelease, numeric core ordering, build-metadata stripping, empty/all-prerelease fallbacks) and the endpoint auth/scope checks are sound. But the persistence layer targets a schema shape that does not exist, so stored and custom tags are lost in production despite green tests. Fix the per-package storage model (finding 1) and reconsider persisting `latest` from the publish body (finding 2) before merge.

ThaSami:
Thanks for the careful review — both points were spot on. Pushed a follow-up commit addressing them:

**1. Storage shape — moved off the `packages` column to a dedicated table.**
You're right that `packages` is one row per `(repository_id, name, version)` (migration `019`, `UNIQUE(repository_id, name, version)`). A `dist_tags` column there is per-version: writes fan across every version row, and a per-`(repo, name)` read matches multiple rows, so the `fetch_optional` read was unsound the moment a package had more than one version. Replaced it with a dedicated table:

```sql
CREATE TABLE IF NOT EXISTS npm_dist_tags (
    repository_id UUID        NOT NULL REFERENCES repositories(id) ON DELETE CASCADE,
    name          TEXT        NOT NULL,
    tags          JSONB       NOT NULL DEFAULT '{}'::jsonb,
    created_at    TIMESTAMPTZ NOT NULL DEFAULT NOW(),
    updated_at    TIMESTAMPTZ NOT NULL DEFAULT NOW(),
    PRIMARY KEY (repository_id, name)
);
```

dist-tags are a per-package concept, so `PRIMARY KEY (repository_id, name)` makes the read provably single-row and decouples this from the `packages` row shape entirely. The old `114_packages_dist_tags.sql` migration is removed; the new table is `119_npm_dist_tags.sql`. The three SQL sites (publish persist, `dist_tags_put`, `dist_tags_delete`) and the read all go through the new table; `dist_tags_put` no longer needs the prior `rows_affected() == 0 -> NotFound` guard because the UPSERT creates the row.

**2. `latest` was being clobbered by publish.**
Also correct — every `npm publish` body carries `dist-tags: { latest: <this version> }`, including for prereleases, so persisting the body verbatim defeated the whole semver-derivation point of #1543. The publish path now drops `latest` from the persisted tags and lets the read derive it (highest non-prerelease semver). An explicit `npm dist-tag add pkg@x.y.z latest` still wins, because the read keeps a stored `latest` that resolves to a known version and only derives when there isn't one.

**Tests** (this was the part that masked it before — single-version seeds):
- the dist-tags read test now seeds **three** versions with the prerelease created last, so the multi-version read path is actually exercised;
- the put/get/delete test drops its `packages` pre-seed (the UPSERT creates the row; the target-version check reads `artifacts`);
- new `test_publish_prerelease_does_not_clobber_latest`: publishes `2.0.0-rc.1` with `dist-tags: { latest: 2.0.0-rc.1 }` over an existing `1.1.0`, then asserts the stored tags have **no** `latest` and the served packument's `latest` is still `1.1.0`.

ThaSami:
@brandonrc re-review please

brandonrc (approved):
Re-reviewed after the rework. Both prior blocking findings are now FIXED. Approving, pending CI green (branch is BEHIND main, needs a rebase).

### Status of prior findings

Finding 1 (per-version `packages` table corrupted dist-tags): FIXED. New dedicated table `backend/migrations/119_npm_dist_tags.sql:21-28` with `PRIMARY KEY (repository_id, name)`, genuinely per-package. Migration number 119 does not collide (highest existing was 111). The read path `fetch_npm_dist_tags` now uses `fetch_optional` against a single PK row (correct), and both write paths (`publish_package`, `dist_tags_put`) upsert with `ON CONFLICT (repository_id, name) DO UPDATE`, no fan-out across version rows.

Finding 2 (plain `npm publish` clobbering `latest` with a prerelease): FIXED. `publish_package` removes `latest` from the publish body before persisting, so `latest` is only ever derived by semver (or set explicitly via `dist-tag add`). New test `test_publish_prerelease_does_not_clobber_latest` covers it.

Semver comparator and the concurrent JSONB upsert (`tags || EXCLUDED.tags`, merges keys so concurrent publishers do not lose each other's tags) both check out, and the packument change is additive/back-compatible for clients that read only `latest`.

### One note before merge
The dist-tag persistence and tests are DB-backed. Please confirm the integration tests actually execute in CI rather than skipping when no test database is configured, otherwise the new behavior is effectively untested in the pipeline.

Nice work moving this to a proper table. Approving.

PR #1564, opened 2026-06-02
fix: harden peer replication sync worker

### LINKED ISSUE #1565: Peer replication ignores configured S3 storage backend and reads artifacts from local STORAGE_PATH
## Summary

When Artifact Keeper is configured with `STORAGE_BACKEND=s3`, peer replication still tries to read artifact bytes from the local filesystem using `STORAGE_PATH` instead of using the configured repository/storage backend.

As a result, replication fails for artifacts that are correctly stored in S3, because the sync worker looks for them under a local path such as `/data/storage/...` or `/var/lib/artifact-keeper/artifacts/...`.

Related PR: https://github.com/artifact-keeper/artifact-keeper/pull/1564

### Environment Information

- Artifact Keeper backend: `1.1.9`
- Storage backend: `s3`
- Deployment: two independent Artifact Keeper instances configured as peers
- Replication mode tested: peer repository replication / mirror-style cross-peer sync
- Artifacts are uploaded successfully and stored in S3

### Backend Version

1.1.9

### Web UI Version

1.1.3

### Package Format

None

### Expected Behavior

The sync worker should read artifact bytes through the configured storage backend for the repository/instance.

For S3-backed repositories, it should fetch the artifact from S3 using the same storage abstraction that normal upload/download paths use.

### Observed Behavior

The sync worker attempts to read the artifact directly from local filesystem storage.

The failure looks like this conceptually:

```text
Storage read error: Failed to read '/data/storage/c9/69/c96930630463f0ab9cade553464f91fc18b0137833b608d93e5bd11ad99290ea': No such file or directory
```

The artifact exists in S3 and can be served by the source instance, but replication fails because the worker bypasses the configured S3 backend.

### Steps to Reproduce

1. Run Artifact Keeper with S3 storage enabled, for example:

   ```env
   STORAGE_BACKEND=s3
   S3_BUCKET=...
   S3_ENDPOINT=...
   AWS_ACCESS_KEY_ID=...
   AWS_SECRET_ACCESS_KEY=...
   ```

2. Configure two Artifact Keeper instances as peers.
3. Create a repository on both instances.
4. Enable peer replication for that repository.
5. Upload an artifact to one instance.
6. Wait for the peer sync worker to process the sync task.

### Logs

```shell

```

## Technical details

In `backend/src/services/sync_worker.rs`, the replication worker reads artifact bytes via filesystem access based on `STORAGE_PATH`.

This means peer replication does not honor the configured `StorageRegistry` / repository storage backend. The main application path already has initialized storage backends, but the sync worker is spawned with only the database pool, so it cannot resolve the correct backend for the artifact.

The fix should make the sync worker use the same configured storage abstraction as the rest of the application instead of constructing local filesystem paths itself.

## Why this matters

This breaks peer replication for production deployments where Artifact Keeper uses object storage, which is a common setup for Kubernetes and HA deployments.

In that architecture, local filesystem storage is either empty, ephemeral, or not mounted at all, so replication cannot rely on local paths.

## Proposed fix

Pass the initialized storage registry/backend into the sync worker and read artifact bytes through the artifact repository's configured storage backend.

This is implemented in PR https://github.com/artifact-keeper/artifact-keeper/pull/1564

### Pre-submission Checklist

valfon1393:
## Summary

Fix peer-to-peer replication for repositories backed by non-local storage, especially S3-backed peers, and harden mirror-mode sync behavior.

Previously, peer replication assumed artifacts could be read from the backend local filesystem. That broke S3 peer <-> S3 peer replication because the sync worker did not read artifacts through the repository's configured storage backend. This also exposed a few related mirror-mode problems: heartbeat status was not actively probed, replicated deletes could re-enter the sync pipeline, and failed sync retries were not configurable enough for real peer outages.

Changes included here:

- Make the sync worker read artifacts from the repository's configured storage backend instead of assuming local storage.
- Push replicated artifacts through the path-aware `PUT /artifacts/{path}` endpoint.
- Add a replication marker header so mirrored uploads/deletes are applied on the target peer without enqueueing a new sync task.
- Prevent delete replication loops in mirror mode.
- Cancel superseded pending push tasks when an artifact is deleted.
- Add active peer heartbeat probes so peer online/offline state is refreshed even when no sync task is currently running.
- Make sync retry behavior configurable through environment variables:
  - `SYNC_TASK_MAX_RETRIES`
  - `SYNC_TASK_RETRY_BACKOFF_BASE_SECS`
  - `SYNC_TASK_RETRY_BACKOFF_MAX_SECS`
- Keep the default max retry count conservative at `3`; longer retry windows can be enabled by deployment-specific env vars.

This fixes the main failure mode where S3-backed peers could not replicate artifacts to each other because the backend tried to read files as if they were stored locally. It also makes mirror-mode delete propagation safe by distinguishing replication traffic from user/API traffic and avoiding recursive sync task creation.

## Test Checklist

- [x] Manually tested locally

## API Changes

- [x] N/A - no API changes

## Linked issue

Fixes #1565

### DIFF
--- a/backend/migrations/119_upload_session_artifact_metadata.sql
+++ b/backend/migrations/119_upload_session_artifact_metadata.sql
@@ -1,0 +1,10 @@
+-- Preserve source artifact metadata across resumable upload sessions.
+--
+-- Regular client uploads can omit these nullable columns and retain the
+-- existing fallback behavior. Peer replication sets them from the source
+-- artifact row so chunked replication does not derive name/version from the
+-- path at completion time.
+
+ALTER TABLE upload_sessions
+    ADD COLUMN artifact_name TEXT,
+    ADD COLUMN artifact_version TEXT;
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -1904,6 +1904,8 @@ fn build_cached_artifact_response(
         download_count: 0,
         created_at: entry.cached_at,
         metadata: None,
+        cache_cached_at: Some(entry.cached_at),
+        cache_expires_at: None,
     }
 }
 
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -43,6 +43,14 @@ fn require_auth(auth: Option<AuthExtension>) -> Result<AuthExtension> {
     auth.ok_or_else(|| AppError::Authentication("Authentication required".to_string()))
 }
 
+fn is_replication_request(headers: &HeaderMap) -> bool {
+    headers
+        .get("x-artifact-keeper-replication")
+        .and_then(|value| value.to_str().ok())
+        .map(|value| matches!(value.to_ascii_lowercase().as_str(), "true" | "1" | "yes"))
+        .unwrap_or(false)
+}
+
 /// Coerce the requested `is_public` value against the server-wide guest-access
 /// policy (issue #850).
 ///
@@ -2975,7 +2983,7 @@ pub async fn upload_artifact(
     super::cleanup_soft_deleted_artifact(&state.db, repo.id, &path).await;
 
     let artifact = artifact_service
-        .upload(
+        .upload_with_sync_options(
             repo.id,
             &path,
             &name,
@@ -2983,6 +2991,7 @@ pub async fn upload_artifact(
             &content_type,
             body,
             Some(auth.user_id),
+            !is_replication_request(&headers),
         )
         .await?;
 
@@ -3431,6 +3440,7 @@ pub async fn delete_artifact(
     State(state): State<SharedState>,
     Extension(auth): Extension<Option<AuthExtension>>,
     Path((key, path)): Path<(String, String)>,
+    headers: HeaderMap,
 ) -> Result<()> {
     let auth = require_auth(auth)?;
     auth.require_scope("delete")?;
@@ -3452,7 +3462,9 @@ pub async fn delete_artifact(
     .map_err(|e| AppError::Database(e.to_string()))?
     .ok_or_else(|| AppError::NotFound("Artifact not found".to_string()))?;
 
-    artifact_service.delete(artifact).await?;
+    artifact_service
+        .delete_with_sync_options(artifact, !is_replication_request(&headers))
+        .await?;
 
     Ok(())
 }
--- a/backend/src/api/handlers/upload.rs
+++ b/backend/src/api/handlers/upload.rs
@@ -22,6 +22,7 @@ use uuid::Uuid;
 
 use crate::api::middleware::auth::AuthExtension;
 use crate::api::SharedState;
+use crate::services::package_service::PackageService;
 use crate::services::upload_service::{self, UploadError, UploadService};
 
 // ---------------------------------------------------------------------------
@@ -52,6 +53,16 @@ pub struct CreateSessionRequest {
     pub repository_key: String,
     /// Path within the repository (e.g. "images/vm.ova")
     pub artifact_path: String,
+    /// Artifact name to persist when the upload completes.
+    ///
+    /// Optional for regular client uploads. Peer replication sets this from
+    /// the source artifact row so chunked replication preserves metadata.
+    pub artifact_name: Option<String>,
+    /// Artifact version to persist when the upload completes.
+    ///
+    /// Optional for regular client uploads. Peer replication sets this from
+    /// the source artifact row so chunked replication preserves metadata.
+    pub artifact_version: Option<String>,
     /// Total file size in bytes
     pub total_size: i64,
     /// Expected SHA256 checksum of the complete file
@@ -150,6 +161,8 @@ async fn create_session(
         repo_id: repo.0,
         repo_key: &req.repository_key,
         artifact_path: &req.artifact_path,
+        artifact_name: req.artifact_name.as_deref(),
+        artifact_version: req.artifact_version.as_deref(),
         total_size: req.total_size,
         chunk_size: req.chunk_size,
         checksum_sha256: &req.checksum_sha256,
@@ -387,21 +400,24 @@ async fn complete(
     let _ = tokio::fs::remove_file(&temp_path).await;
 
     // Create artifact record
+    let artifact_name = completed_artifact_name(&session);
+    let artifact_version = completed_artifact_version(&session);
     let artifact_id: Uuid = sqlx::query_scalar(
         r#"
         INSERT INTO artifacts (repository_id, path, name, version, size_bytes,
                                checksum_sha256, content_type, storage_key, uploaded_by)
         VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
         ON CONFLICT (repository_id, path) DO UPDATE SET
-            size_bytes = $5, checksum_sha256 = $6, content_type = $7, storage_key = $8,
-            uploaded_by = $9, updated_at = NOW(), is_deleted = false
+            name = $3, version = $4, size_bytes = $5, checksum_sha256 = $6,
+            content_type = $7, storage_key = $8, uploaded_by = $9,
+            updated_at = NOW(), is_deleted = false
         RETURNING id
         "#,
     )
     .bind(session.repository_id)
     .bind(&session.artifact_path)
-    .bind(artifact_name_from_path(&session.artifact_path))
-    .bind::<Option<String>>(None) // version
+    .bind(artifact_name)
+    .bind(artifact_version)
     .bind(session.total_size)
     .bind(&session.checksum_sha256)
     .bind(&session.content_type)
@@ -410,6 +426,20 @@ async fn complete(
     .fetch_one(&state.db)
     .await
     .map_err(|e| map_err(StatusCode::INTERNAL_SERVER_ERROR, e))?;
+
+    if let Some((package_name, package_version)) = completed_package_catalog_entry(&session) {
+        PackageService::new(state.db.clone())
+            .try_create_or_update_from_artifact(
+                session.repository_id,
+                package_name,
+                package_version,
+                session.total_size,
+                &session.checksum_sha256,
+                None,
+                None,
+            )
+            .await;
+    }
 
     tracing::info!(
         "Finalized chunked upload {} -> artifact {} ({}B, sha256:{})",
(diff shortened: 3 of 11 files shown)


brandonrc (commented):
**Verdict: Request changes.** The core fix (reading through the configured storage backend) is correct and unblocks S3-to-S3 peers, but two issues should be resolved before merge: a severe S3 bandwidth regression on the chunked path, and a silent loss of checksum verification on replicated artifacts.

### Critical findings

**1. Chunked S3 replication re-downloads the entire object once per chunk (O(N^2) egress)**

`backend/src/services/sync_worker.rs` `read_artifact_chunk_from_storage` (diff lines 919-1000) opens a fresh `storage.get_stream(&task.storage_key)` for every chunk and discards all bytes before that chunk's offset. For S3, `get_stream` issues a full-object GET with no range (`backend/src/storage/s3.rs:971-988`). `execute_chunked_transfer` calls this helper once per chunk, so a large S3-backed artifact split into N chunks is downloaded from S3 N times. A 10 GB model at the 50 MB default chunk size (200 chunks) pulls roughly 1 TB from S3.

This is exactly the workload the PR targets (large images, ML models on S3). The previous local-FS path used `File::seek` for true random access, so this is a real egress/time regression for the S3 case the PR is meant to fix, and it compounds under retries.

Fix: add a range-capable read to the `StorageBackend` trait (e.g. `get_range(key, offset, len)` backed by S3 `GetObject` `Range:` / object_store `GetRange`) and use it in the chunk reader; or stream the object once and feed chunks sequentially from a single stream. A default trait impl can fall back to the current behavior for backends without native range support.

**2. Replicated artifacts are stored without checksum verification (header-name mismatch)**

The worker sends the integrity header as `X-Artifact-Checksum-SHA256` (`sync_worker.rs:591`, unchanged). The new push target `upload_artifact` reads the declared checksum from `x-checksum-sha256` (`backend/src/api/handlers/repositories.rs:2462-2467`) before calling `ArtifactService::verify_checksums`. Because the names differ, `declared_sha256` is always `None`, and `verify_checksums` skips the check when absent (`backend/src/services/artifact_service.rs:165`).

The worker already knows the source checksum (`task.checksum_sha256`) but the receiver never compares it against the received bytes, so a truncated or corrupted single-request transfer is accepted silently and later served to clients. Since the PR moves replication onto this endpoint, this is the right moment to close the gap.

Fix: send the checksum under `x-checksum-sha256` (the header the PUT handler reads) so `verify_checksums` runs on replicated uploads. Worth a test that a mismatched body is rejected.

### Non-blocking notes

- Heartbeat probe failures and transfer failures now feed the same `consecutive_failures` counter into `calculate_backoff`, so a peer reachable for `/api/v1/peers` but failing transfers (or vice versa) gets a skewed backoff. Consider tracking probe health separately.
- Please verify end-to-end against an S3 peer that has never seen the artifact: `init_transfer` requires the `artifact_id` to already exist on the target (`transfer_service.rs:86-93`) and `complete_session` only flips session status, so confirm cross-peer chunked push actually materializes an artifact row + storage object on a fresh peer. This is pre-existing behavior the PR does not change, flagging only.
- The default backoff change to a constant 300s base/max is intentional per the description and tests were updated to match. No concern, just noting the behavior change from the old 10s to 300s ramp.

### Summary

Routing replication reads through `StorageRegistry::backend_for` instead of assuming local filesystem is the right fix, and the mirror-loop suppression plus delete-supersedes-push cancellation are sound. Before merge, address the per-chunk full-object S3 re-download and the checksum header-name mismatch. Both are concentrated in `sync_worker.rs` and the storage trait, and neither requires reworking the design.

valfon1393:
@brandonrc Thanks for the detailed review. I updated the PR to address the two blocking findings and re-tested the chunked S3 path end-to-end.

Changes made:

- Added a range-capable `StorageBackend::get_range(key, offset, len)` API with a streaming fallback.
- Implemented native range reads for S3 via `object_store::get_range`, and for filesystem via seek/read.
- Updated chunked peer replication to read each chunk through `get_range` instead of opening a full stream and discarding bytes before the chunk offset.
- Changed the replication checksum header to `x-checksum-sha256`, matching the upload handler, so replicated single-request uploads now go through the existing SHA-256 verification path.
- While validating the fresh-peer chunked case from the non-blocking notes, I also found that the old chunked transfer path could not materialize an artifact on a peer that had never seen it. The updated worker now uses the upload-session API for chunked peer replication, so completion creates the target storage object and artifact row.
- During the same validation I found and fixed two follow-up chunked-session issues: replicated chunked uploads now preserve source `artifact_name` / `artifact_version`, and the completion path now updates `packages` / `package_versions` consistently with normal single-request PUT uploads.

I intentionally did not change the heartbeat/probe backoff coupling in this update, since that was called out as non-blocking and is separate from the S3 replication correctness issues.

Validation performed:

- Local checks:
  - `cargo fmt --check`
  - `cargo test --workspace --lib` (`10692 passed; 0 failed; 4 ignored`)

- End-to-end validation on two real S3-backed peers in separate Kubernetes deployments:
  - started with the test repository clean on both peers: API empty, DB empty, and S3 containing only the expected flag files;
  - checksum-negative test on both peers: `PUT` with an invalid `x-checksum-sha256` returned HTTP 400, and a follow-up `GET` returned 404;
  - small artifact replication worked in both directions, with matching downloaded SHA-256, API metadata, `artifacts`, `packages`, and `package_versions`;
  - 160 MiB chunked artifact replication worked in both directions, selecting the chunked path and completing in 4 chunks;
  - downloaded target artifacts matched source SHA-256 in both directions;
  - API metadata matched across peers for `path`, `name`, `version`, `size_bytes`, `checksum_sha256`, and `content_type`;
  - DB `artifacts`, `packages`, and `package_versions` rows matched across peers;
  - S3 contained the same content-addressed objects on both buckets with matching sizes;
  - sync tasks completed with `retry_count=0`;
  - no transfer failures or permanently failed sync tasks were observed.

This covers both blocking findings and also verifies the fresh-peer chunked replication case: the target peer did not already have the artifact, and the replicated chunked upload created the artifact, package catalog rows, storage object, and downloadable bytes correctly.

brandonrc (commented):
Thanks for the quick turnaround. Both prior blocking findings are resolved for the S3 and filesystem backends. There is one remaining gap on the other object stores.

### Status of prior findings

**Finding 1 (O(N^2) S3 egress in chunk reader): FIXED for S3 and filesystem.**
- `read_artifact_chunk_from_storage` no longer opens a full stream per chunk; it resolves the repo's backend and calls `storage.get_range(&task.storage_key, offset, length)` (`backend/src/services/sync_worker.rs:1324-1377`).
- S3 does a true ranged read: `S3Backend::get_range` calls `self.store.get_range(&path, range)` (object_store GetRange / S3 GetObject Range), not a full get plus slice (`backend/src/storage/s3.rs:1196-1220`). Range bounds are end-exclusive and correct (`byte_range` returns `offset..offset+length`, `s3.rs:822-836`; test at `1695-1707`), overflow rejected.
- Filesystem does a real seek plus bounded read (`filesystem.rs:281-319`), and `compute_chunk_ranges` clamps the last chunk to `total_size - offset` (`sync_worker.rs:1184-1200`), so it never requests past EOF. Short reads are rejected (`1364-1376`). Content-Range uses inclusive end `offset + length - 1` (RFC 9110 correct).

**Finding 2 (checksum header mismatch skipped verification): FIXED.**
- The worker now sends the checksum under `CHECKSUM_SHA256_HEADER = "x-checksum-sha256"` (`sync_worker.rs:508`, used on the single-request PUT at `:980`); the old `X-Artifact-Checksum-SHA256` literal is gone.
- `upload_artifact` reads `x-checksum-sha256` and calls `verify_checksums(&body, declared_sha256, ...)` before storing (`repositories.rs:2462-2467`), so verification now runs and rejects a mismatch (your E2E note confirms HTTP 400). A regression test pins the literal (`sync_worker.rs:1722-1725`). The chunked path verifies the full-file SHA-256 server-side at session complete, so it is covered too.

### New critical finding

**The generic `get_range` default reintroduces the same O(N^2) egress for GCS and Azure.** The default trait impl calls `get_stream` and slices the requested window (`backend/src/storage/mod.rs:148-193`). Only S3 and filesystem override it. `GcsBackend` (`gcs.rs:451`) and `AzureBackend` (`azure.rs:398`) are real, supported backends that do not override `get_range`, and their `get_stream` downloads the whole object over HTTPS (`gcs.rs:1425-1448`, Azure analogous). So a GCS- or Azure-backed repo replicating a large artifact via the chunked path re-downloads the full object per chunk: the same problem finding 1 flagged, on a different backend. Both clients are REST and support a `Range` header, so a native override is feasible.

Fix: add native ranged reads to the GCS and Azure backends, or gate the chunked replication path to backends with native range support. Whether this blocks merge depends on whether GCS/Azure peers are in scope for this release; if they are, it is blocking.

### Summary

The two original issues are genuinely fixed for S3 and filesystem: true ranged reads and a matching checksum header so `verify_checksums` runs and rejects mismatches. The remaining ask is to give GCS and Azure native `get_range` implementations (or restrict chunked replication to range-capable backends) so the egress fix is not silently lost on those object stores.

valfon1393:
@brandonrc Thanks, agreed. I pushed an update for this as well.

The supported object-store backends now have native ranged reads:

- `GcsBackend::get_range` sends an HTTP `Range: bytes=start-end` request and expects `206 Partial Content`.
- `AzureBackend::get_range` sends ranged requests as well, including `Range` / `x-ms-range` for Azure Blob Storage.
- Both implementations reject overflow in range construction and preserve the existing fallback path behavior for migrated storage keys.
- The generic `StorageBackend::get_range` fallback still exists, but S3, filesystem, GCS, and Azure now all override it, so the chunked replication path no longer silently falls back to full-object GETs for the supported storage backends you called out.

I do not have live GCS/Azure environments available for true end-to-end validation, so I did not want to overclaim that. Instead I added mock HTTP regression coverage for both backends:

- GCS tests assert that `get_range` sends the expected `Range: bytes=...` header.
- GCS fallback tests assert the migrated-key fallback also sends a ranged request.
- Azure tests assert that `get_range` sends the expected `Range` and `x-ms-range` headers.
- Azure fallback tests assert the fallback path remains ranged as well.
- Range helper tests cover inclusive end-byte calculation and overflow rejection.

I also reran the local validation suite after the change:

- `cargo fmt --check`
- `cargo test --workspace --lib`

The full lib test run passed (`10734 passed`). I also rebuilt/deployed the patched backend in my S3-backed environment and verified the S3 path still starts and behaves as before. Live GCS/Azure E2E remains untested due to lack of those environments, but the transport-level tests now pin the important regression condition: chunk reads must issue native ranged HTTP requests instead of full-object downloads.

brandonrc:
Superseded by #1665 — a clean reconstruction of this PR's genuine peer-replication work (per-repo `backend_for` storage resolution, `get_range` for all backends, heartbeat probes, retry policy, delete-loop prevention) rebased onto current main. This branch had an evil-merge commit (`e9601c5`) that hid the GCS/Azure `get_range` work in neither parent and tangled in unrelated npm/pypi/migration/CI content, so it couldn't be rebased mechanically. #1665 cherry-picks the two genuine commits + re-applies only the GCS/Azure `get_range` additions, drops the noise, and renumbers the migration (120→123, since 120 is now taken by merged #1641). Original work credited to you (@valfon1393) on the commit + PR. Thanks — this is a solid fix; tracking it forward in #1665.

Issue #461, opened 2026-03-19
Maven Virtual members issues functional and visual


pipelineRat:
### Description

Reference: https://github.com/artifact-keeper/artifact-keeper/issues/450
After updates to issue 450, I pulled the latest changes to main, and the images and ran `docker compose up -d` to make certain the update images were applied to my cluster

Still having both visual and functional issues with maven virtual repo not containing members.

I create the virtual repo via the web interface, and during creation select the repositories to add.
From the web interface, when selecting the virtual repo **virtual-maven-all**, the members tab still shows No member repositories configured.

When I attempt to pull artifacts from that repository, which includes my remote of upstream maven central, I get an error in the maven build:
- either a 401 unauthorized (when set to private)
- or a "could not find artifact in artifact-keeper" (when set to public)

In another browser tab, I attempt to navigate to what would be a full artifact path under the virtual-maven-all endpoint: https://artifact-keeper.internal.com/maven/virtual-maven-all/org/apache/maven/plugins/maven-resources-plugin/3.3.1/maven-resources-plugin-3.3.1.pom

I get `Virtual repository has no members`

In order to populate members I have to either execute a rest call, or add them via the web interface by selecting the repo, then selecting the **Add Member** button.
- still results in web interface reporting No member repositories configured
- however should I attempt to pull artifacts, or navigate to them via a web address, the virtual endpoint now works.

I think what I am seeing here is potentially 3 unique bugs.
I have only tested with maven so far, so I don't know if this applies to all repos of type **virtual**

1. When creating a **virtual** **maven** repo from the user interface, and selecting to add members. The created virtual repo does not contain any members, despite selecting them from the interface during creation.
2. The visual web interface still does not list member repositories, even after they have been added either through a rest call or the web interface via the "+ Add Members" button.
3. virtual maven repos, that are **private** are not respecting authentication, reporting 401 unauthorized. switching to type **public** the virtual repo works as intended with a maven build.
    - assuming user has successfully added members

### Environment Information

- Deployment type: docker compose
- Operating System: Rhel9.7
- Architecture: x86_64
- Database: postgres
- Reverse Proxy: caddy
- Network setup: 

### Backend Version

latest

### Web UI Version

latest

### Package Format

None

### Expected Behavior

1) When creating a virtual maven repo via the web interface, the created repo should be populated by the selected members
2) When members are applied to a virtual maven repo, the visual interface tab **Members**, should accurately reflect the included list of members
[...]

PR #1567, opened 2026-06-02
fix(repositories): surface proxy-cached artifacts in the listing endpoint

### LINKED ISSUE #1548: Proxy-cached artifacts no longer tracked in artifacts table, causing ak_storage_used_bytes to under-report
### Description

Commit `63f1409` (#1280) stopped inserting proxy-cached items into the `artifacts` table to fix a filesystem-backend storage path mismatch (#1278). That fix is correct — the doubled-prefix bug was real — but the acknowledged tradeoff leaves a gap: proxy-cached storage is no longer reflected in any storage metric or listing.

### Expected Behavior

Proxy-cached storage should be accounted for, without re-introducing the doubled-prefix filesystem bug. Options include:
1. A separate Prometheus gauge (e.g. `ak_proxy_cache_used_bytes`) queried from proxy metadata sidecars
2. A scheduler-service query that sums storage from the proxy cache directory via the storage backend
3. A lightweight tracking mechanism (separate table or counter) that avoids the original `storage_for_repo` dispatch issue

### Observed Behavior

After deploying #1280, `ak_storage_used_bytes` stops growing on proxy-cache hits. The underlying fix for #1278 works correctly (no more HTTP 500 on second cached read), but storage accounting is now incomplete.

### Environment Information

- Deployment type: All (Docker, bare metal, etc.)
- Storage backend: All (filesystem, S3, GCS, Azure)
- Backend Version: Any version including #1280

### Steps to Reproduce

1. Deploy with filesystem or S3 backend
2. Create a remote repository (e.g. PyPI proxy to https://pypi.org)
3. Fetch a package through the proxy cache
4. Observe: SELECT COUNT(*) FROM artifacts WHERE is_deleted = false does not increase
5. Observe: ak_storage_used_bytes stays flat despite bytes being written to storage

### Additional Context

Affected surfaces:
- Prometheus gauge ak_storage_used_bytes — computed by SELECT SUM(size_bytes) FROM artifacts WHERE is_deleted = false only covers hosted (directly uploaded) artifacts
- Per-repository storage_used_bytes in GET /api/v1/repositories — same WHERE repository_id = ANY($1) query
- Artifact listing endpoints — proxy-cached items do not appear at all

For deployments that primarily use remote/proxy repositories, this makes storage monitoring effectively blind to the majority of stored bytes.

The commit message for #1280 explicitly calls this out:
> Tradeoff: proxy-cached items no longer surface in the repository artifact-listing endpoint or storage_used_bytes accounting. That UX gap is tracked separately.

This issue tracks that gap. The original #1278 filesystem doubled-prefix bug must not be re-introduced; any fix should correctly route reads through the same storage backend that the write used.

brandonrc:
## Summary

Remote (proxy) repositories stopped recording cached items in the `artifacts` table in #1280 (the fix for the #1278 doubled-prefix storage path bug on filesystem backends). The acknowledged tradeoff was that proxy-cached items no longer surface in the repository artifact-listing endpoint or storage accounting. The user-visible symptom is that packages pulled through a remote repo fill up storage but show 0 artifacts in the UI, so they can't be browsed or scanned (artifact-keeper-web #424).

This reconstructs the listing for `remote` repositories from the storage backend instead of the database:

- `ProxyService::list_cached_artifacts` walks the `proxy-cache/<repo_key>/` prefix, keeps only `__content__` leaf keys, and loads each `__cache_meta__.json` sidecar for size, checksum, content type, and cached-at.
- `list_artifacts` takes an early branch for `RepositoryType::Remote` that maps those entries into `ArtifactResponse`, applying the existing `path_prefix` and `q` filters and pagination in-process.
- Cached entries get a deterministic id (SHA-256 over the cache storage key, first 16 bytes) so the same object reports a stable id across calls without colliding with the random v4 ids hosted artifacts get from the DB.

The #1278 fix is preserved: nothing is re-inserted into `artifacts`, and reads still resolve through the same global storage the proxy wrote to. Local and virtual repository listing paths are untouched.

Unit tests cover the pure helpers: `cached_artifact_paths` (key parsing), `filter_and_paginate_cached` (prefix/query filtering + pagination), and `build_cached_artifact_response` (field mapping + deterministic id). The end-to-end user flow is covered by a new Playwright spec in artifact-keeper-web (#424).

## Test Checklist
- [x] E2E tests added/updated (if applicable) — in artifact-keeper-web
- [x] Manually tested locally — cargo build, cargo clippy, targeted unit tests pass

## API Changes
- [x] N/A - no API changes

The response shape of GET /api/v1/repositories/{key}/artifacts is unchanged; only the data source for remote repos changes.

Closes #1548

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -1479,6 +1479,26 @@ pub async fn list_artifacts(
         .await;
     }
 
+    // Remote (proxy) repositories no longer record cached items in the
+    // `artifacts` table (#1278 / #1280): doing so reintroduced a doubled-prefix
+    // storage path bug on filesystem backends. The cached bodies and metadata
+    // sidecars still live in the storage backend under `proxy-cache/<key>/`, so
+    // the listing is reconstructed from there. Without this, packages pulled
+    // through a remote repo fill up storage but never appear in the UI, so they
+    // can't be browsed or scanned (#1548, web #424).
+    if repo.repo_type == RepositoryType::Remote {
+        return list_remote_cached_artifacts(
+            &state,
+            &repo,
+            &key,
+            query.path_prefix.as_deref(),
+            query.q.as_deref(),
+            page,
+            per_page,
+        )
+        .await;
+    }
+
     let (artifacts, total) = if repo.repo_type == RepositoryType::Virtual {
         // For virtual repositories, aggregate artifacts from all member repos.
         // Members are returned in priority order; local/hosted members are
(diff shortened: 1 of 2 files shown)


brandonrc (commented):
Security/API review (SSRF, path traversal, injection, authz, DoS lens).

Net: safe to include in the re-cut v1.2.0 from a security standpoint. No traversal, injection, or authz defect. One Medium performance/DoS concern worth tracking, not a blocker.

Path traversal / prefix escape — SAFE. `cached_artifact_paths` only consumes keys that `storage.list(Some("proxy-cache/<repo_key>/"))` already returned, then `strip_prefix` + `strip_suffix("/__content__")`. Keys outside the repo's prefix do not match (covered by `test_cached_artifact_paths_skips_other_repos_and_empty`). The recovered logical path is used for display and to rebuild the sidecar key via the same `cache_metadata_key` used on write, never concatenated into a fresh filesystem path here, so a `..` embedded in a stored key cannot escape beyond what the cache writer already placed on disk.

repo_key injection — SAFE. `repo.key` is the validated repository record, not raw request input, and only seeds a prefix passed to a prefix-scoped `storage.list`. No interpolation into SQL or a shell.

Deterministic SHA-256 id — SAFE. 128-bit truncation; birthday collision is negligible across a cache, and the id is not an authorization key (authz is by `repo.key` + path). Spoofing a DB v4 UUID would need a preimage. Fine.

Local/virtual behavior — UNCHANGED. The early return is gated strictly on `RepositoryType::Remote`; the Virtual aggregation and hosted/local DB path are untouched.

Filtering vs pagination order — Medium (DoS / memory), please track before scaling remote caches. `list_cached_artifacts` lists the entire prefix and then calls `load_cache_metadata` for EVERY entry, materializing all sidecars into a `Vec`, and only then does `filter_and_paginate_cached` slice the page. So:
- Every `list_artifacts` request on a remote repo is O(N) storage reads regardless of `page`/`per_page`. On S3 that is one object GET per cached artifact per listing call, trivially amplified by repeated paging.
- Memory and latency grow unbounded with cache size; a large proxied repo can stall the endpoint and inflate object-store cost.
- `path_prefix`/`q` filters run in-process after the full load, so they do not bound the work either.
Suggested follow-up (not required for the re-cut): cap the number of keys/entries scanned (with a "results truncated" signal), or fetch sidecars only for the page slice after sorting by key, or add a guard on cache size. The skip-on-missing-sidecar handling is correct and avoids panics.

Minor: `total_pages` divides by `per_page` without guarding `per_page == 0`; confirm the query layer rejects/clamps `per_page=0` upstream or it yields inf/NaN cast to u32.

brandonrc (commented):
Test/E2E quality review (v1.2.0 batch), backend unit tests for ProxyService::list_cached_artifacts. Verdict: the pure-function tests are genuine and well-targeted; the integration boundary is untested (acceptable, but worth naming).

**What is proven well.**
- `cached_artifact_paths` tests are the strongest: `extracts_content_keys` asserts sidecars (`__cache_meta__.json`) are dropped, prefix+suffix stripped, and sorted. `skips_other_repos_and_empty` is a real negative-case test: it proves a different repo's cache (`other-remote`) does not leak in and that the bare-root `__content__` is skipped. That cross-repo isolation check is exactly the kind of thing that distinguishes a correct fix from a naive `strip_prefix` that would over-match. Good.
- `filter_and_paginate_cached` covers sort, path_prefix filter, case-insensitive `q` (`IS-ODD` matching `is-odd`), and pagination offset (page 2 of size 2 yields just "c"). These would fail against off-by-one or case-sensitive implementations.
- `build_cached_artifact_response` asserts field mapping AND the determinism contract: same repo_key+path => same id, different path => different id. That backs the docstring claim about stable client-side keys.

**Gap (name it, don't block).** Every test here operates on hand-built `CachedArtifactEntry` vecs or raw key strings. Nothing exercises `list_cached_artifacts` itself: the storage `.list()` walk, the `load_cache_metadata` sidecar read, the skip-on-missing-sidecar branch, or the `storage.list` error path returning `Vec::new()`. So the actual reconstruction-from-storage logic (the heart of #1548) has no unit coverage; only its pure helpers do. The web #424 e2e (PR #461) is the only thing covering the real path, and that spec skips when upstream NPM is unavailable. Net result: in a CI run with no upstream, the end-to-end behavior of this fix may never be asserted by any test. Recommend a test with an in-memory/temp storage backend that writes a `__content__` + sidecar pair and asserts `list_cached_artifacts` returns one populated entry, plus the missing-sidecar skip. Not requesting changes since the pure logic is solid, but this gap should be tracked.

PR #1574, opened 2026-06-02
fix(uploads): stage large incus upload bodies under STORAGE_PATH/.incoming, not /tmp

### LINKED ISSUE #1573: Large uploads (incus, OCI) stage to /tmp by default — multi-GiB pushes evict the pod on Kubernetes; add a shared staging dir under STORAGE_PATH + reaper
## Summary

incus (monolithic + chunked) and OCI blob uploads stage the request body to a local temp file whose base directory defaults to `std::env::temp_dir()` (`/tmp`). On Kubernetes `/tmp` is typically a small `emptyDir`, so a multi-GiB upload overruns it and kubelet evicts the pod **mid-receive** — the artifact never lands and the client sees a `503`. The temp file also **leaks** when the receive is killed (OOM / eviction / restart) before the handler's own cleanup runs, and nothing reaps it.

## Detail

- `incus::temp_upload_path` → `$AK_INCUS_UPLOAD_TMP_DIR` else `std::env::temp_dir()`.
- `oci_v2::oci_upload_temp_path` → `$AK_OCI_UPLOAD_TMP_DIR` else `std::env::temp_dir()`.
- Both stream the body to that path (incus `stream_body_to_file` / `append_body_to_file`; OCI `stream_body_to_temp_file`) before handing it to the `StorageBackend` via `put_stream`.
- On a stock K8s deploy `/tmp` is a modest `emptyDir` (256Mi in the reference chart). A multi-GiB PUT overruns it → `Usage of EmptyDir volume "tmp" exceeds the limit` → pod `Evicted` → upload fails (behind an L7 gateway the client sees `503 upstream connect error … connection termination`).
- A receive killed before cleanup leaves an orphan temp file. The per-format helpers only clean on their own success/error paths; the chunked-session reaper only covers DB-tracked chunked sessions, not the monolithic temp file — so orphans accumulate.
- Other formats (helm, npm, pypi, debian, cargo, …) stream straight to the `StorageBackend` or buffer in memory; they don't stage to a local temp file, so they're unaffected.

## Proposed fix — a shared staging facility

- Stage under a reserved dir on the **data volume**, e.g. `<STORAGE_PATH>/.incoming`, instead of `/tmp`. `STORAGE_PATH` is a real local writable mount on every backend (it already holds filesystem-backed artifacts, plugins, backups) and is sized by the deployment's storage provisioning — so staging inherits the same sizing as the artifacts it stages. Keep it an **ephemeral** volume: staging is transient scratch, nothing should persist across restarts.
- One env override (e.g. `AK_UPLOAD_STAGING_DIR`) replacing the per-format `AK_INCUS_UPLOAD_TMP_DIR` / `AK_OCI_UPLOAD_TMP_DIR`.
- Use a dot-prefixed subdir (`.incoming`) so it can't collide with a repository key prefix in the filesystem backend's `<base>/<prefix>/<key>` layout.
- Add an **age-based sweep** of the staging dir to reap crash-orphans, on the same schedule and max-age as the existing chunked-session reaper, so both expire staging on the identical threshold.
- Rewire incus + OCI onto the shared helper, consolidating their near-identical per-format temp-path helpers and body-to-file streamers.

## Impact / workaround

Today operators must enlarge the `/tmp` emptyDir or point the per-format `AK_*_UPLOAD_TMP_DIR` at a larger volume just to upload multi-GiB artifacts. Staging on the data volume removes that, and the reaper closes the orphan leak.

## PR

We have this implemented and will open a PR linked to this issue. It adds a shared `staging` module that **both** the incus and OCI handlers use — a common `<STORAGE_PATH>/.incoming` staging folder (one `AK_UPLOAD_STAGING_DIR` override, replacing the two per-format `/tmp`-defaulting helpers), with the age-based sweep wired into the scheduler on the chunked-session reaper's max-age. incus (monolithic + chunked) and OCI are both rewired onto it, and the two near-identical per-format temp-path helpers and body-to-file streamers are consolidated.

dragonpaw:
> **Updated 2026-06-03 (rebased onto current `main`):** re-scoped to **incus-only**. #1448 (merged 2026-06-02, after this PR was approved) already routes OCI blob uploads through the storage backend, so OCI no longer stages to local `/tmp` and the `oci_v2.rs` changes here were dropped to avoid colliding with that approach. The shared `api::handlers::staging` module stays format-generic (namespaces by `kind`) but incus is its only caller now. `cargo check --tests` green; `main`'s `incus.rs` still stages to `std::env::temp_dir()`, so the incus fix here is still needed.

---

## Summary

incus (monolithic + chunked) and OCI blob uploads staged the request body to a local temp file under `std::env::temp_dir()` (`/tmp`). On Kubernetes `/tmp` is a small `emptyDir`, so a multi-GiB upload overran it and the pod was evicted mid-receive (artifact lost, client `503`); temp files also leaked when a receive was killed (OOM/eviction/restart) before the handler's cleanup ran, with nothing to reap them.

Closes #1573.

Adds a shared `api::handlers::staging` module that **both** the incus and OCI handlers use:

- `staging_temp_path(storage_path, kind, id)` roots staging at `<STORAGE_PATH>/.incoming` (the data volume — a real local writable mount on every backend, sized like the artifacts it stages), not `/tmp`. One env override `AK_UPLOAD_STAGING_DIR` replaces the per-format `AK_INCUS_UPLOAD_TMP_DIR` / `AK_OCI_UPLOAD_TMP_DIR`. `.incoming` is dot-prefixed so it can't collide with a repo key prefix in the filesystem backend's `<base>/<prefix>/<key>` layout.
- `sweep_stale(storage_path, max_age_hours)` reaps crash-orphaned staging, run in the scheduler's hourly upload-cleanup task on `UPLOAD_STAGING_MAX_AGE_HOURS = 24` — the same threshold as the chunked-session reaper (`cleanup_stale_sessions`), which now shares the constant.
- incus (`temp_upload_path`) and OCI (`oci_upload_temp_path`) are rewired onto it, consolidating their near-identical per-format temp-path helpers; OCI's `stream_body_to_temp_file` now ensures the staging dir exists before writing.

Staging stays **ephemeral** (it's transient scratch — nothing persists across restarts); the sweep handles orphans that outlive a crashed receive. Formats that stream straight to the `StorageBackend` or buffer in memory are unaffected.

`staging::tests` assert the default root is `<storage_path>/.incoming` (never `/tmp`), the `AK_UPLOAD_STAGING_DIR` override, per-uuid/kind namespacing, and `sweep_stale` (keeps fresh / reaps aged / tolerates a missing root). The incus and OCI path tests now assert the `.incoming` location. Before this change the staging path was `std::env::temp_dir()`, so these would have failed on the `/tmp` default.

## Test Checklist
- [x] Manually tested locally — on a self-hosted Kubernetes (object-store backend) deploy, a multi-GiB `.tar.zst` PUT now stages under `.incoming`, returns `202`, and finalizes to object storage with no `/tmp` eviction and nothing left in `/tmp`.
- [x] No regressions in existing tests (`cargo check --tests` clean; staging/incus/OCI path tests pass)

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/admin.rs
+++ b/backend/src/api/handlers/admin.rs
@@ -732,9 +732,11 @@ pub async fn run_cleanup(
 
     if request.cleanup_stale_uploads.unwrap_or(false) {
         use crate::api::handlers::incus::cleanup_stale_sessions;
-        result.stale_uploads_deleted = cleanup_stale_sessions(&state.db, 24)
-            .await
-            .map_err(AppError::Internal)?;
+        use crate::api::handlers::staging::UPLOAD_STAGING_MAX_AGE_HOURS;
+        result.stale_uploads_deleted =
+            cleanup_stale_sessions(&state.db, UPLOAD_STAGING_MAX_AGE_HOURS as i64)
+                .await
+                .map_err(AppError::Internal)?;
     }
 
     Ok(Json(result))
--- a/backend/src/api/handlers/incus.rs
+++ b/backend/src/api/handlers/incus.rs
@@ -189,17 +189,16 @@ pub(crate) fn storage_path_for_key(storage_base: &str, key: &str) -> PathBuf {
 /// Uploads stage to this on-disk path so PATCH continuations can append
 /// without re-downloading the in-progress object from the storage backend.
 /// The final, durable copy is written via `StorageBackend::put_stream` at
-/// completion time. The base directory is `$AK_INCUS_UPLOAD_TMP_DIR` if
-/// set, otherwise `std::env::temp_dir()`. This is intentionally NOT the
-/// repository's configured storage path: when the backend is S3/GCS, the
-/// repo's `storage_path` is a bucket-relative key prefix, not a local
-/// filesystem path, so staging there would either fail (the path isn't
-/// writable on disk) or land bytes in a location nothing reads back.
-pub(crate) fn temp_upload_path(session_id: &Uuid) -> PathBuf {
-    let root = std::env::var("AK_INCUS_UPLOAD_TMP_DIR")
-        .map(PathBuf::from)
-        .unwrap_or_else(|_| std::env::temp_dir());
-    root.join(format!("ak-incus-upload-{}", session_id))
+/// completion time.
+///
+/// Staging goes through the shared [`crate::api::handlers::staging`] facility,
+/// which roots under `<STORAGE_PATH>/.incoming` (the data volume) rather than
+/// `/tmp` — a multi-GiB image staged on the small `/tmp` emptyDir evicts the
+/// pod mid-receive. `config.storage_path` here is the local writable mount,
+/// NOT a repo's bucket-relative `storage_path`, so it's always a real on-disk
+/// location regardless of `STORAGE_BACKEND`.
+pub(crate) fn temp_upload_path(storage_path: &str, session_id: &Uuid) -> PathBuf {
+    crate::api::handlers::staging::staging_temp_path(storage_path, "incus", session_id)
 }
 
 /// Open the staged upload temp file as a `BoxStream<Result<Bytes>>` ready
@@ -763,7 +762,7 @@ async fn upload_image(
     // filesystem temp-and-rename), so peak memory per upload is bounded
     // to STREAM_CHUNK_BUDGET regardless of image size.
     let temp_id = Uuid::new_v4();
-    let temp_path = temp_upload_path(&temp_id);
+    let temp_path = temp_upload_path(&state.config.storage_path, &temp_id);
     let (size_bytes, checksum) = stream_body_to_file(body, &temp_path).await?;
 
     // Extract metadata from the file on disk
@@ -963,7 +962,7 @@ async fn start_chunked_upload(
     // The chosen path is persisted to `incus_upload_sessions.storage_temp_path`,
     // so subsequent chunk/complete/cancel calls read it back from the
     // session row and don't need to re-derive it.
-    let temp_path = temp_upload_path(&session_id);
+    let temp_path = temp_upload_path(&state.config.storage_path, &session_id);
 
     // Stream initial body (may be empty) to temp file
     let (initial_bytes, _checksum) = stream_body_to_file(body, &temp_path).await?;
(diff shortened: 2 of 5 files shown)


brandonrc (commented):
**Verdict: Approve.** No critical issues found.

I traced every staging call site in `incus.rs`, `oci_v2.rs`, `staging.rs`, and `scheduler_service.rs`.

What I verified:

- `storage_path` is genuinely local. `config.storage_path` is the global `STORAGE_PATH` (`config.rs:605`), defaulting to `/var/lib/artifact-keeper/artifacts`, not a repo bucket-relative key prefix. So `<STORAGE_PATH>/.incoming` is always a real on-disk path even with `STORAGE_BACKEND=s3`. The core claim holds.
- Staging dir is created on every write path. Incus `stream_body_to_file` calls `ensure_parent_dirs` (`incus.rs:91`); OCI `stream_body_to_temp_file` now `create_dir_all`s the parent before open (`oci_v2.rs:496-503`), covering POST, PATCH appends, and PUT-complete. No "directory missing" gap.
- Cleanup on error paths is intact. Incus removes the temp file on `put_stream` failure and success (`incus.rs:775/778`, `1128/1131`), checksum mismatch (`1104`), and cancel (`1224`). OCI calls `cleanup_upload_temp` on every error branch and on completion. A partial file left by a mid-stream error is the documented orphan case that `sweep_stale` reaps. This matches prior behavior.
- Sweep is safe. `sweep_stale` no-ops on a missing root, is best-effort per entry, and reaps by mtime so in-flight uploads (which bump mtime) are preserved. `checked_sub`/`saturating_mul`/`unwrap_or` guard against time-arithmetic panics.
- The `.incoming` dot-prefix correctly avoids colliding with the filesystem backend's 2-char prefix layout.

Summary: clean, well-scoped fix that consolidates two per-format temp-path helpers into a shared `staging` module and moves staging off the `/tmp` emptyDir onto the data volume. The one residual leak window (process killed mid-receive) is exactly what the new hourly sweep covers. Approve.

dragonpaw:
Rebased onto the latest `main` and **re-scoped to incus-only**.

When you approved this, the OCI half overlapped with #1448 — which merged ~30 min later and routes OCI blob uploads through the storage backend (`upload_part_storage_key`) instead of local `/tmp`. That made the OCI staging changes here both redundant and conflicting, so I dropped the `oci_v2.rs` changes entirely; `oci_v2.rs` is now byte-identical to `main`.

What remains:
- the shared `api::handlers::staging` module (kept format-generic — namespaces by `kind` — for future reuse, but incus is its only caller now);
- incus monolithic + chunked uploads moved off `std::env::temp_dir()` onto `<STORAGE_PATH>/.incoming`;
- the hourly `sweep_stale` reaper for crash-orphaned temp files.

`main`'s `incus.rs` still stages to `/tmp`, so this still fixes a live multi-GiB eviction bug. `cargo check --tests` is green. Flagging the scope change since you'd approved the wider version — happy to instead fold incus staging into the #1448 storage-backend approach if you'd prefer one mechanism across both formats.

brandonrc (approved):
Re-reviewed at head `51416ba` after the rework. Still good, approving (just needs a rebase, it is BEHIND main).

The rework correctly narrowed scope to incus only and dropped the OCI changes (right call, #1448 already routes OCI blob uploads through the storage backend). Re-verified the parts that matter:
- Staging roots at `<config.storage_path>/.incoming` (`staging.rs:48-60`), and `config.storage_path` is the dedicated local filesystem field (`config.rs:97-98`), distinct from the bucket fields, so it is a real on-disk path even with `STORAGE_BACKEND=s3`.
- The staging dir is `create_dir_all`-ed before every write via `ensure_parent_dirs` (`incus.rs:91`), covering monolithic and chunked entrypoints.
- Temp files are removed on success and all error branches (`incus.rs:775,778,1104,1128,1131,1224`), and `sweep_stale` no-ops on a missing dir, reaps by mtime (so in-flight chunked uploads are preserved), and uses saturating/checked time math so it cannot panic (`staging.rs:90-98`).
- Staged filename is a server-generated UUID (`incus.rs:200-201`), so no path traversal or cross-upload collision. Cleanup survived the rebase intact.

No critical issues. Non-blocking: if `STORAGE_BACKEND=s3` and `STORAGE_PATH` is unset, staging falls back to the default `/var/lib/...` path, which on an object-store K8s deploy may be the pod overlay rather than a sized volume; strictly better than the old `/tmp` emptyDir, and `AK_UPLOAD_STAGING_DIR` is the escape hatch, but worth a one-line note for S3 operators. Rebase to clear BEHIND and merge once CI is green.

brandonrc:
Superseded by merged #1622 (#1573), which fixes the same /tmp-staging bug with a stronger implementation (RAII cleanup guard on all paths, prefix-selective reaper, AK_UPLOAD_STAGING_DIR + back-compat env). The one thing #1574 had that #1622 lacks — wiring the orphan sweep into the hourly scheduler — is salvaged as https://github.com/artifact-keeper/artifact-keeper/issues/1654. Thanks for this; the core fix shipped via #1622.

PR #1575, opened 2026-06-02
fix(migration): to_artifact_keeper() returns Nexus types instead of AK enum values

### LINKED ISSUE #1576: bug(migration): to_artifact_keeper() returns Nexus types, breaking all JFrog → AK migration jobs
## Summary

`RepositoryType::to_artifact_keeper()` in `migration_service.rs` returns Nexus terminology (`hosted`, `proxy`, `group`) instead of Artifact Keeper's own `repository_type` enum values (`local`, `remote`, `virtual`). This silently breaks every JFrog → AK migration job.

## DB schema

```sql
-- migration 003_repositories.sql
CREATE TYPE repository_type AS ENUM ('local', 'remote', 'virtual');
```

## The bug

```rust
pub fn to_artifact_keeper(&self) -> &'static str {
    match self {
        Self::Local => "hosted",   // ← Nexus, not AK
        Self::Remote => "proxy",   // ← Nexus, not AK
        Self::Virtual => "group",  // ← Nexus, not AK
    }
}
```

This breaks the migration worker in two places:

### 1. `check_repository_conflict` — always reports TypeMismatch for existing repos

```rust
let target_type = repo_type.to_artifact_keeper(); // "hosted"
if existing_type != target_type {                 // "local" != "hosted" → always true
```

Any existing repo (even one with a perfectly matching type) is treated as a conflict and skipped.

### 2. `create_repository` — INSERT always fails

```rust
let repo_type = config.repo_type.to_artifact_keeper(); // "hosted"
INSERT INTO repositories (..., repo_type, ...) .bind(repo_type)
// → DB error: invalid input value for enum repository_type: "hosted"
```

## Reproduction

1. Configure a JFrog Artifactory source connection
2. Target any local repo that doesn't exist in AK yet
3. Start the migration job → fails in < 1 second, 0 items processed

Backend log:
```
ERROR Failed to create destination repository; skipping
error=Database error: error returned from database: invalid input value for enum repository_type: "hosted"
```

If the repo already exists in AK, you get the conflict path instead:
```
ERROR Destination repository conflict; skipping artifact transfer for this repo
conflict=TypeMismatch Repository 'example-gems-local' exists with type 'local', cannot migrate as 'hosted'
```

## Fix

Return AK's own enum values and update the three test assertions that encode the wrong expected output. See PR #1575.

ysolomon-plat:
Fixes #1576.

## Bug

`RepositoryType::to_artifact_keeper()` returns Nexus terminology (`hosted`, `proxy`, `group`) instead of Artifact Keeper's own `repository_type` enum values (`local`, `remote`, `virtual`).

This breaks **every JFrog → AK migration job** in two places:

### 1. `check_repository_conflict` (line 337)

```rust
let target_type = repo_type.to_artifact_keeper(); // returns "hosted"
if existing_type != target_type {                 // "local" != "hosted" → always true
```

The DB stores `"local"` but the comparison uses `"hosted"`, so any existing repo (even one with a matching type) is flagged as `TypeMismatch` and skipped.

### 2. `create_repository` (line 394)

```rust
let repo_type = config.repo_type.to_artifact_keeper(); // returns "hosted"
INSERT INTO repositories (..., repo_type, ...) VALUES (...)
.bind(repo_type) // "hosted" is not in the repository_type enum → DB error
```

The INSERT always fails with:
```
invalid input value for enum repository_type: "hosted"
```

## Fix

Return AK's own enum values from `to_artifact_keeper()` and update the three test assertions that encoded the wrong expected output.

## Verified against

- DB schema: `CREATE TYPE repository_type AS ENUM ('local', 'remote', 'virtual')` (migration 003)
- AK version: 1.2.0-rc.3 / 1.2.0

### DIFF
--- a/backend/src/services/migration_service.rs
+++ b/backend/src/services/migration_service.rs
@@ -84,9 +84,9 @@ impl RepositoryType {
     /// Convert to Artifact Keeper repository type string
     pub fn to_artifact_keeper(&self) -> &'static str {
         match self {
-            Self::Local => "hosted",
-            Self::Remote => "proxy",
-            Self::Virtual => "group",
+            Self::Local => "local",
+            Self::Remote => "remote",
+            Self::Virtual => "virtual",
         }
     }
 }
@@ -1460,9 +1460,9 @@ mod tests {
 
     #[test]
     fn test_repository_type_to_artifact_keeper() {
-        assert_eq!(RepositoryType::Local.to_artifact_keeper(), "hosted");
-        assert_eq!(RepositoryType::Remote.to_artifact_keeper(), "proxy");
-        assert_eq!(RepositoryType::Virtual.to_artifact_keeper(), "group");
+        assert_eq!(RepositoryType::Local.to_artifact_keeper(), "local");
+        assert_eq!(RepositoryType::Remote.to_artifact_keeper(), "remote");
+        assert_eq!(RepositoryType::Virtual.to_artifact_keeper(), "virtual");
     }
 
     #[test]
@@ -1472,7 +1472,7 @@ mod tests {
             let ak_type = repo_type.to_artifact_keeper();
             // Verify the AK type is valid
             assert!(
-                ["hosted", "proxy", "group"].contains(&ak_type),
+                ["local", "remote", "virtual"].contains(&ak_type),
                 "Unexpected AK type '{}' for '{}'",
                 ak_type,
                 rclass


brandonrc (commented):
**Verdict: Request changes.** The migration mapping fix (commit 1) is correct and complete, but commit 2 bundles unrelated cache-field edits that do not compile.

### Critical findings

**1. Undefined variable `entry` in `build_artifact_response` (compile error)**

`backend/src/api/handlers/repositories.rs:2021`:
```
cache_cached_at: Some(entry.cached_at),
```
This function's parameters are `artifact: &Artifact`, `repo_key`, `download_count`. There is no `entry` in scope, so this is a hard E0425 compile error. It is also semantically wrong: the adjacent comment states listing helpers must leave these `None` and let `get_artifact_metadata` populate them later (#1541). Fix: revert to `cache_cached_at: None`.

**2. Undefined variable `entry` in `upload_artifact` (compile error)**

`backend/src/api/handlers/repositories.rs:3008`: same `Some(entry.cached_at)` where the handler binds `artifact`, not `entry`. Same E0425, and semantically wrong since a just-uploaded local artifact has no proxy cache state. Fix: revert to `cache_cached_at: None`.

### The good part

The actual subject of the PR is correct and complete: `to_artifact_keeper()` (`backend/src/services/migration_service.rs:85-91`) maps `hosted->local`, `proxy->remote`, `group->virtual`, which exactly match the DB enum `repository_type AS ENUM ('local','remote','virtual')` (`migrations/003_repositories.sql:8`). All three variants are covered with no catch-all, both production callers agree with the stored values, and there are no other callers depending on the old Nexus-term return. Data integrity is fine: the prior bad strings could never have been INSERTed (the enum cast would have rejected them), so there are no orphaned rows to clean up.

### Summary

Drop commit 2 (or revert lines 2021 and 3008 to `None`) and ship only the migration fix, which is good to merge on its own. Note this PR collides with #1586/#1579/#1588, which are all editing the same `cache_*` lines in `repositories.rs`; the cache-field work should live in exactly one PR (#1579 is the clean version), not be scattered across migration PRs.

ysolomon-plat:
Thanks for the review — your requested changes have been actioned.

brandonrc (approved):
Re-reviewed after the update. The requested change is addressed: the unrelated `repositories.rs` cache-field edits (the commit-2 `Some(entry.cached_at)` lines that referenced an out-of-scope `entry` and would not compile) are gone. The PR is now scoped to just the migration mapping fix.

Verified the fix is correct and complete:
- `to_artifact_keeper()` now returns `local`/`remote`/`virtual` (migration_service.rs:85-91), which exactly match `CREATE TYPE repository_type AS ENUM ('local', 'remote', 'virtual')` (backend/migrations/003_repositories.sql:8). The old `hosted`/`proxy`/`group` were Nexus terms the `$5::repository_type` cast (migration_service.rs:404) would reject, which is the bug.
- All three variants are covered with no catch-all, so no invalid value can escape.
- Both call sites (`check_repository_conflict` at :337, `create_repository` INSERT at :394/404) consume the corrected value; no other caller depends on the old return.
- The two unit tests are updated to match.

Approving. Only remaining gate is CI going green.

PR #1579, opened 2026-06-03
fix(repositories): restore build — set cache fields in cached-artifact listing (closes #1578)

### LINKED ISSUE #1578: main fails to compile: build_cached_artifact_response missing cache_cached_at / cache_expires_at (E0063)
## Symptom

`main` does not compile. `cargo build`/`cargo test` fail with:

```
error[E0063]: missing fields `cache_cached_at` and `cache_expires_at` in initializer of `repositories::ArtifactResponse`
    --> backend/src/api/handlers/repositories.rs:1887:5
```

The scheduled CI on `main` is red as a result.

## Root cause

Two PRs that were each green on their own branch collided on merge:

- **#1567** added `build_cached_artifact_response()` in `backend/src/api/handlers/repositories.rs`.
- **#1542** added the required fields `cache_cached_at` / `cache_expires_at` to the `ArtifactResponse` struct.

`#1542`'s branch predated `#1567`, so it never saw the new initializer and didn't update it. After both landed, that one initializer is missing the two now-required fields.

## Reproduction

```bash
git checkout main
cd backend && cargo build
# error[E0063]: missing fields cache_cached_at and cache_expires_at
```

## Fix

Populate the two fields in `build_cached_artifact_response`. A cached-listing entry is a live proxy-cache object, so `cache_cached_at` is exactly `entry.cached_at`; the sidecar projection (`CachedArtifactEntry`) carries no expiry, so `cache_expires_at` stays `None` (the per-artifact metadata endpoint from #1541 surfaces expiry). Both fields are `skip_serializing_if = Option::is_none`, so the listing wire shape is unchanged.

PR incoming.

## Additional context

Related: #1542, #1567, #1541.

knowinglyAnonymous:
## Summary

`main` currently fails to compile (`error[E0063]: missing fields cache_cached_at and cache_expires_at`), and the scheduled CI on `main` is red.

Two PRs collided on merge:
- **#1567** added `build_cached_artifact_response()`.
- **#1542** added the now-required `cache_cached_at` / `cache_expires_at` fields to `ArtifactResponse` but, having branched before #1567, never updated that initializer.

This populates the two fields in `build_cached_artifact_response`:
- `cache_cached_at: Some(entry.cached_at)` — a cached-listing entry is a live proxy-cache object, so its freshness timestamp is exactly when it was cached.
- `cache_expires_at: None` — the sidecar projection (`CachedArtifactEntry`) doesn't carry expiry; the per-artifact metadata endpoint (#1541) surfaces it.

Both fields are `skip_serializing_if = "Option::is_none"`, so the listing wire shape is unchanged for existing callers.

## Test plan

- [x] `cargo build` / `cargo test --lib` now compile (were broken on `main`).
- [x] Extended `test_build_cached_artifact_response_maps_fields_and_is_deterministic` to assert `cache_cached_at == entry.cached_at` and `cache_expires_at.is_none()` — this assertion would have caught the collision.
- [x] `cargo fmt --check` clean.

Closes #1578.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -1896,6 +1896,17 @@ fn build_cached_artifact_response(
         download_count: 0,
         created_at: entry.cached_at,
         metadata: None,
+        // #1542 added these required fields to ArtifactResponse after #1567
+        // introduced this initializer, so the two landed together and broke
+        // the build. This entry is, by construction, a live proxy-cache
+        // object, so its cache-freshness timestamp is exactly when it was
+        // cached. The listing's sidecar projection (`CachedArtifactEntry`)
+        // doesn't carry the expiry, so leave `cache_expires_at` to the
+        // per-artifact metadata endpoint (#1541). Both fields are
+        // `#[serde(skip_serializing_if = "Option::is_none")]`, so the listing
+        // wire shape is unchanged for callers that don't read them.
+        cache_cached_at: Some(entry.cached_at),
+        cache_expires_at: None,
     }
 }
 
@@ -4403,6 +4414,12 @@ mod tests {
         assert_eq!(resp.checksum_sha256, "deadbeef");
         assert_eq!(resp.download_count, 0);
         assert!(resp.version.is_none());
+        // A cached-listing entry is a live proxy-cache object, so its
+        // freshness timestamp is exactly when it was cached; the sidecar
+        // projection carries no expiry. (Asserting these guards the
+        // #1542/#1567 field-collision regression that broke the build.)
+        assert_eq!(resp.cache_cached_at, Some(entry.cached_at));
+        assert!(resp.cache_expires_at.is_none());
         // Same repo_key + path always yields the same id.
         let resp2 = build_cached_artifact_response(&entry, "npm-remote");
         assert_eq!(resp.id, resp2.id);


knowinglyAnonymous:
@brandonrc — flagging this one for a quick look since it's blocking everyone, not just this change.

`main` currently fails to compile: #1542 and #1567 each landed green on their own branch but collided on merge, leaving `build_cached_artifact_response` without the `cache_cached_at` / `cache_expires_at` fields that #1542 made required (`error[E0063]`). The scheduled CI on `main` is red as a result, and any new backend branch cut from `main` won't build.

This PR is the minimal fix (populate the two fields in that one initializer + extend its existing unit test). It's a clean, low-risk patch — `cargo build` / `cargo test --lib` / `cargo fmt --check` all pass locally.

The npm follow-up #1581 (the non-blocking note you left on #1542) is stacked on this branch because of the build break; once this merges I'll rebase #1581 onto `main` so it shows only its own commit.

Happy to adjust anything. Thanks!

brandonrc (commented):
**Verdict: Approve.** Correctly fixes the compile break with real cache metadata (not a paper-over), and slightly stronger than the competing #1588 because it adds a regression test.

### Critical findings

None.

### Verification

- Compile break is real. `build_cached_artifact_response` (`backend/src/api/handlers/repositories.rs:1882-1899`) constructs `ArtifactResponse` without `cache_cached_at`/`cache_expires_at`, while the struct (`1449-1475`) declares both as required. That is the reported E0063. This PR adds exactly those two fields.
- Values are correct, not papered over. `CachedArtifactEntry` (`backend/src/services/proxy_service.rs:456-471`) carries a real `cached_at` but no `expires_at`, so `Some(entry.cached_at)` surfaces genuine freshness and `cache_expires_at: None` reflects data truly absent from this projection. The richer per-artifact path (`repositories.rs:2788-2789`) still populates both from the full `CacheMetadata`.
- No wire-shape regression: both fields are `skip_serializing_if = "Option::is_none"`, and the None-omission is already asserted by tests at `repositories.rs:5755-5756`.
- The added test (`cache_cached_at == Some(entry.cached_at)`, `cache_expires_at.is_none()`) is deterministic and would have caught this collision.

### Summary

This and #1588 make the identical functional change and both are correct. I recommend merging this one (it adds the regression test that guards against the collision recurring) and closing #1588 as a duplicate. This is blocking `main` compilation, so it is worth fast-tracking.

brandonrc:
Superseded by #1592, which has merged and restored the build. #1592 applies the same fix to `build_cached_artifact_response()` (`cache_cached_at: Some(entry.cached_at)`, `cache_expires_at: None`) and was verified against the full Tier-1 gate set, including `cargo clippy --workspace --all-targets -- -D warnings` (the `Check Rust` gate that kept this PR red). Thanks for jumping on the broken `main` so quickly. Closing as superseded.

PR #1581, opened 2026-06-03
fix(repositories): look up npm cache metadata under the upstream URL path (closes #1580)

### LINKED ISSUE #1580: npm cache-freshness fields stay null: GET /:key/artifacts/:path keys cache lookup on stored path, not upstream URL path
## Symptom

On npm-family Remote (proxy) repos, `GET /api/v1/repositories/{key}/artifacts/{path}` always returns `cache_cached_at` / `cache_expires_at` as absent, even for a tarball that has been fetched through the proxy and has a valid cache entry. The new artifact-details cache rows (web#449) therefore never render for npm.

This is the non-blocking follow-up flagged by @brandonrc in the review of #1542.

## Root cause

The handler resolves the artifact row via `lookup_path_candidates` and ends up with the **stored** path shape (`<name>/<version>/<name>-<version>.tgz`, written by `npm::store_npm_version`). It passes that stored `artifact.path` to `ProxyService::get_cache_metadata`.

But the proxy writes the cache (and its metadata sidecar) keyed on the **upstream download-URL** path shape (`<name>/-/<name>-<version>.tgz`): `npm::serve_tarball` fetches via `build_tarball_upstream_path` → `ProxyService::fetch_artifact` → `cache_metadata_key(repo_key, upstream_path)`.

Stored shape != fetch/cache shape, so the metadata-blob lookup misses and both fields collapse to `None`. It does not error (the lookup is failure-tolerant), so it only suppresses the UI rows — hence a follow-up, not a blocker.

## Fix

Map the stored npm tarball path back to the URL shape before the cache lookup (the exact inverse the listing endpoint already applies via `formats::npm::tarball_url_path_from_stored`). Non-npm formats and npm non-tarball rows (metadata, raw uploads, already-URL paths) are left unchanged.

PR incoming. Note: it is currently stacked on the `main` build fix in #1578/#1579 because `main` does not compile.

## Additional context

Related: #1542 (introduced the fields), web#449 (the UI rows), #1578 (build regression this stacks on).

knowinglyAnonymous:
## Summary

Follow-up to @brandonrc's non-blocking note on #1542.

For npm-family Remote repos, `GET /:key/artifacts/:path` resolved the row to the **stored** path (`<name>/<version>/<file>.tgz`) and keyed `ProxyService::get_cache_metadata` on it. The proxy, however, caches npm tarballs under the **upstream download-URL** shape (`<name>/-/<file>.tgz`, written by `npm::serve_tarball` → `build_tarball_upstream_path` → `cache_metadata_key`). The shapes differ, so the lookup always missed and `cache_cached_at` / `cache_expires_at` stayed `None` even with a valid cache entry — silently suppressing the artifact-details cache rows (web#449). It never errored, so this was a follow-up rather than a blocker.

This adds `cache_metadata_lookup_path()`, which maps the stored tarball path back to the URL shape for npm-family formats (the exact inverse the listing endpoint already uses via `formats::npm::tarball_url_path_from_stored`), and keys the cache lookup on it. Non-npm formats and npm non-tarball rows (metadata, raw uploads, already-URL paths) fall through to the literal path unchanged.

## ⚠️ Stacked on #1579

`main` currently does not compile (see #1578). This branch is based on the build fix in **#1579**, so its first commit is that fix. **Please merge #1579 first**; I will then rebase this onto `main` so only the npm commit remains.

## Test plan

- [x] `cargo test --lib cache_metadata_lookup_path` — unscoped/scoped npm + yarn/pnpm mapping, plus identity cases (non-npm format, npm metadata path, already-URL path).
- [x] Existing structural test `get_artifact_metadata_populates_cache_fields_only_for_remote` still passes (gates + `.get_cache_metadata(` call unchanged).
- [x] `cargo fmt --check` and `cargo clippy --lib` clean.

Closes #1580.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -1941,6 +1941,29 @@ fn lookup_path_candidates(path: &str, format: &RepositoryFormat) -> Vec<String> 
     out
 }
 
+/// Map a stored artifact path to the path the proxy cache is keyed under,
+/// for the purpose of looking up cache-freshness metadata (#1541 follow-up).
+///
+/// npm-family tarballs are stored under the version-segmented layout
+/// (`<name>/<version>/<file>.tgz`) by `npm::store_npm_version`, but the proxy
+/// cached them under the upstream download-URL shape (`<name>/-/<file>.tgz`)
+/// because `npm::serve_tarball` fetches via `build_tarball_upstream_path`.
+/// Looking up cache metadata by the stored path therefore always missed.
+/// Translating the stored tarball path back to the URL shape makes the cache
+/// key match what the proxy wrote.
+///
+/// Returns the literal `path` unchanged for non-npm formats and for npm rows
+/// that are not stored tarballs (metadata paths, raw uploads, paths already
+/// in URL form) — `tarball_url_path_from_stored` returns `None` for those.
+fn cache_metadata_lookup_path(path: &str, format: &RepositoryFormat) -> String {
+    if is_npm_family_format(format) {
+        if let Some(url_path) = crate::formats::npm::tarball_url_path_from_stored(path) {
+            return url_path;
+        }
+    }
+    path.to_string()
+}
+
 /// npm-family formats share the publish/download path conventions of
 /// the npm registry (`yarn`, `pnpm`, and `bower` all wrap the same
 /// upstream wire format). Keep this in sync with the `format` mapping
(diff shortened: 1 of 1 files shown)


brandonrc (commented):
**Verdict: Approve.** The read key provably matches the write key for npm-family tarballs (scoped and unscoped), non-npm paths are untouched, and all None/error paths degrade to empty rather than panicking.

### Critical findings

None.

### Verification

- Key match (write vs read). Write side: `serve_tarball` (`backend/src/api/handlers/npm.rs:992`) builds `upstream_path = build_tarball_upstream_path(package_name, filename)` = `"{package_name}/-/{filename}"` (`npm.rs:184`), and the proxy keys cache metadata on that via `cache_metadata_key` (`proxy_service.rs:1260`). For scoped packages it is called with `@scope/pkg` un-encoded (`npm.rs:901`), giving `@scope/pkg/-/file.tgz`. Read side: `cache_metadata_lookup_path` (`repositories.rs:1956`) calls `tarball_url_path_from_stored`, producing `{scope}/{name}/-/{filename}` for scoped (`npm.rs:371`) and `{name}/-/{filename}` for unscoped (`npm.rs:382`). Both run the same `cache_metadata_key` formula, so the strings are identical. Scoped handling is correct.
- Non-npm path not broken. The rewrite only happens when `is_npm_family_format` is true (`repositories.rs:1609`); maven/pypi/cargo fall through to the literal path unchanged (confirmed by the maven identity test).
- Other npm shapes safe. Metadata paths, raw uploads, and already-URL-shaped paths return `None` and fall through. The already-URL case is idempotent (parsing yields version segment `-`, which fails the cross-check at `npm.rs:368,379`), verified by the idempotency test (`npm.rs:1270`).
- No traversal/injection. The transform only reorders existing path segments, and `validate_cache_path` still runs on the read side inside `cache_metadata_key`.
- None handling is graceful: `get_cache_metadata` returns `Ok(None)` on validation failure and the handler chains `.ok().flatten()` (`repositories.rs:2811`), no panic/500.

### Summary

Tight, well-scoped follow-up that maps the version-segmented stored npm path back to the upstream `/-/` shape the proxy actually writes under. Both ends derive the same key for scoped and unscoped packages, non-npm formats are provably untouched, and error handling stays graceful. One operational note: this PR is stacked on #1579 and `main` currently does not compile, so #1579 (or #1588) must merge first.

PR #1584, opened 2026-06-03
fix(pypi): version-aware shadowing guard for virtual PyPI downloads

### LINKED ISSUE #1582: bug(pypi): virtual download shadowing guard over-suppresses proxy — locally-published v2 blocks proxy fetch of v1.x
## Summary

The PyPI virtual download handler uses `virtual_non_remote_owns_name` to suppress the remote proxy when any local member owns a package by name. This is too broad: a locally-published `v2.0.0` silently blocks proxy resolution of `v1.x` versions that were never published locally, returning 404 to the client.

## Reproduction

1. Publish `mylib==2.0.0` to a local member of a virtual PyPI repo
2. Leave `mylib==1.x` only available via the remote proxy member (e.g. pypi.org)
3. Request `mylib==1.0.0` through the virtual repo
4. AK returns 404 — the name-only guard fires because `mylib` is locally owned (by v2.0.0), suppressing the remote entirely

## Root cause

In `backend/src/api/handlers/pypi.rs`, the download handler calls:

```rust
let suppress_remote_members = proxy_helpers::virtual_non_remote_owns_name(
    &state.db,
    repo.id,
    &normalized_project,
)
.await?;
```

`virtual_non_remote_owns_name` executes:
```sql
SELECT 1 FROM artifacts
WHERE repository_id = ANY($1)
  AND is_deleted = false
  AND LOWER(name) = LOWER($2)
LIMIT 1
```

This matches on **name alone** — any local version of the package suppresses the proxy for **all** versions.

## Impact

Any organisation that:
- Has an internal package also available on a public registry (forked, vendored, or the same)
- Publishes only some versions locally

…will find that older/newer versions not in their local repo become inaccessible via the virtual, even though the remote proxy member could serve them. This breaks transitive dependency resolution whenever a direct dependency requires a version the local repo doesn't hold.

## Proposed fix

Make the guard version-aware at the download endpoint: only suppress the remote proxy when a local member owns **the specific requested version**, not any version of the name. A request for `mylib==1.0.0` should only be suppressed if `mylib==1.0.0` exists locally.

The simple index listing path (`/simple/{project}/`) can retain the name-only guard since no specific version is being requested there.

Note: A version-aware fix requires correctly extracting the version from the requested filename. PyPI filenames follow PEP 427 (wheels: `{name}-{version}-{python}-{abi}-{platform}.whl`) and PEP 625 (sdists: `{name}-{version}.tar.gz`). The name segment in **sdist** filenames uses hyphens, while the normalized name (PEP 503) also uses hyphens — but wheels use underscores. Any implementation must handle both forms when building the name prefix for parsing.

## See also

PR #1575 for a related migration tool fix.

ysolomon-plat:
Fixes #1582.

## Problem

The PyPI virtual download handler calls `virtual_non_remote_owns_name` to suppress the remote proxy when any local member owns the package name. This is too broad: a locally-published `v2.0.0` caused 404s for `v1.x` that only existed on the upstream registry.

Concrete failure: an organisation publishes `celery-message-consumer==2.0.0` internally. Their services also depend on `celery-message-consumer~=1.1.1` as a transitive dependency. The virtual returns 404 for `1.1.1` because the name-only guard fires on `2.0.0`.

## Fix

- Add `virtual_non_remote_owns_name_version(... version: Option<&str>)` — when `version` is `Some`, only suppresses the proxy if that exact name+version exists locally. All existing non-PyPI callers continue to use the name-only path via the `None` branch (no behaviour change for Maven, npm, Cargo, etc.).
- Add `PypiHandler::version_from_filename(filename, normalized_name)` to extract the version from both wheel filenames (underscores in name, PEP 427) and sdist filenames (hyphens in name, PEP 625). Also strips PEP 658 `.metadata` sidecar extensions before parsing.
- Wire the version-aware guard into the PyPI virtual download path.

## Trade-off

Narrows dependency-confusion protection to exact name+version matches. Operators who want strict name-level isolation should use a dedicated local-only virtual rather than mixing local and proxied packages in the same virtual repo. Documented in a code comment referencing #1582.

### DIFF
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -12,6 +12,7 @@ use crate::api::download_response::try_presigned_redirect;
 use crate::api::handlers::error_helpers::map_storage_err;
 use crate::api::AppState;
 use crate::error::AppError;
+use crate::formats::pypi::PypiHandler;
 use crate::models::repository::{
     ReplicationPriority, Repository, RepositoryFormat, RepositoryType,
 };
@@ -1598,6 +1599,21 @@ pub async fn virtual_non_remote_owns_name(
     virtual_repo_id: Uuid,
     package_name: &str,
 ) -> Result<bool, Response> {
+    virtual_non_remote_owns_name_version(db, virtual_repo_id, package_name, None).await
+}
+
+/// Version-aware variant of [`virtual_non_remote_owns_name`]. When `version`
+/// is `Some`, the guard fires if a local member owns a PEP 440-equal version
+/// (`PypiHandler::canonical_version`), so `1.0`/`1.0.0` still match. The guard
+/// is fail-safe: `version = None`, or a requested version that cannot be
+/// canonicalized, falls back to name-only suppression (any local version of the
+/// name suppresses the proxy) rather than allowing fan-out.
+pub async fn virtual_non_remote_owns_name_version(
+    db: &PgPool,
+    virtual_repo_id: Uuid,
+    package_name: &str,
+    version: Option<&str>,
+) -> Result<bool, Response> {
     let members = fetch_virtual_members(db, virtual_repo_id).await?;
     let non_remote_ids: Vec<Uuid> = members
         .iter()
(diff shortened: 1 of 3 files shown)


ThaSami:
@brandonrc we fixed this on simple for virtual repos
not sure if this is a variation
https://github.com/artifact-keeper/artifact-keeper/pull/1267

ysolomon-plat:
Thanks @ThaSami — #1267 is related but this is a different layer, and the two are complementary.

#1267 fixed `simple_project` (the `simple/<project>/` index): it now unions local + remote member entries so pip *sees* every version. This PR fixes `serve_file` (the download path #1267 didn't touch): the version-aware guard lets pip *download* the version it picked.

Concretely: a local member holds `celery-message-consumer==2.0.1`; a consumer needs `1.1.1` (upstream only). After #1267 the index correctly lists `1.1.1`, but when pip requests the `1.1.1` file, the old name-only guard in `serve_file` still suppresses the remote member for the whole name and the download 404s. This PR narrows that suppression to the exact requested version, so `1.1.1` proxies through while `2.0.1` stays locally owned.

So #1267 = visibility in the index, this = fetchability of the proxied artifact. Without this PR, #1267 surfaces versions that then 404 on download.

ysolomon-plat:
Want to add these have been tested in production currently and are working as expected.

brandonrc (commented):
**Verdict: Request changes.** The guard correctly narrows the intent (suppress the proxy only for the requested version), but the version comparison uses the wrong representation with no PEP 440 normalization, which reopens a dependency-confusion shadowing hole.

### Critical findings

**1. Shadow check compares filename-encoded version against stored metadata version**

- `backend/src/api/handlers/pypi.rs:1014-1020` feeds `PypiHandler::version_from_filename(filename)` into the guard.
- The locally stored `artifacts.version` comes from the multipart `version` form field (twine metadata), not the filename (`pypi.rs:1432-1435`, `1580-1589`).
- The comparison is exact, case-insensitive string match: `LOWER(version) = LOWER($3)` (`backend/src/api/handlers/proxy_helpers.rs:1624-1635`).

The guard's `$3` is parsed from the filename, but the column was populated from the metadata version. These match for simple versions (`1.0.0`) but diverge under PEP 427/503 filename escaping. Example: a local upload of metadata version `1.0+ubuntu.1` stores `version = "1.0+ubuntu.1"`, but the wheel filename is `pkg-1.0+ubuntu_1-...whl` (`+` and non-alphanumeric runs become `_`), so `version_from_filename` returns `1.0+ubuntu_1`. The `=` fails, the guard does not fire, and the request falls through to the Remote member. A version that IS held locally stops being shadowed, which is exactly the dependency-confusion case this code prevents. This is a security regression, not just a 404: an attacker publishing the same version upstream with a filename form that does not string-match the local metadata-derived version gets served instead of the local artifact.

Fix: compare a normalized PEP 440 form on both sides, and derive the requested version from the same source of truth used at upload (the metadata version), not the filename.

**2. No PEP 440 normalization: `1.0` vs `1.0.0` vs `1.0.post0`**

`proxy_helpers.rs:1624-1635` is exact string equality, and no PEP 440 canonicalization exists in `backend/src/formats/pypi.rs` (confirmed by grep). PEP 440 treats `1.0`, `1.0.0`, `1.0.0.0` as equal, and pip resolves requirements to whatever release exists, so a local `1.0` versus an upstream filename carrying `1.0.0` misfires. Both directions are possible: leak (local version not shadowed) and false 404 (a genuinely distinct upstream version withheld).

Fix: canonicalize both operands to PEP 440 normalized form before comparing, and apply normalization at write time so `artifacts.version` is stored normalized.

### Summary

The `$3 IS NULL` fallback correctly preserves name-only behavior for non-PyPI callers, and the narrowing is the right idea. The blocker is the comparison: filename-derived version against metadata-derived stored version, exact string equality, no PEP 440 normalization, so any version whose filename form differs from its stored form silently falls through to the remote and reopens the shadowing hole. Derive both sides from the same source, normalize to canonical PEP 440, and add tests for `1.0` vs `1.0.0`, local-version `+` escaping, and `.post`/epoch forms.

ysolomon-plat:
Added the maintainer-directed fix for the PyPI virtual shadowing guard:

- made the guard version-aware for PyPI virtual downloads
- compare requested/stored versions through PEP 440 canonicalization instead of raw string equality
- fail closed to name-level suppression when the requested filename version cannot be safely canonicalized
- added tests for release padding, local-version separators, post/epoch/pre/dev forms, and the fail-safe local-version case
- fixed the rustfmt issue and squashed the branch back to one commit

ysolomon-plat:
@brandonrc ready to merge

brandonrc (approved):
Dug into this against #1267. Short version: #1584 is complementary, not a duplicate, and my earlier PEP 440 concern is resolved by the current diff. Approving.

### #1584 vs #1267: different layers (confirmed in code)
- #1267 (merged 2026-05-21) rewrote the virtual branch of `simple_project`, the `simple/<project>/` index LISTING, to union all members via `merge_local_into_remote_simple_html` so pip sees every version (local + upstream).
- #1584 changes the per-file DOWNLOAD/resolution path (`serve_file`), adding `virtual_non_remote_owns_name_version` to decide shadowing precedence when a specific filename is fetched.
- No overlap: #1267 never touches `serve_file`/the guard; #1584 never touches `simple_project`/the union. If anything #1267 makes #1584 more necessary: now that the index legitimately advertises both a locally held version and upstream-only versions of the same name, the old name-only guard would 404 the upstream-only file, which is exactly what #1584 fixes. So keep both, and keep the `Fixes #1582` link. ThaSami's question was a good one, but it is not the same fix.

### My prior PEP 440 finding: resolved
The current diff (head `90ec188f`) no longer does the exact-string `LOWER(version) = LOWER($3)` comparison. It canonicalizes via `PypiHandler::canonical_version` (formats/pypi.rs) and compares PEP 440-equal forms in `pypi_version_owned` (proxy_helpers.rs), so `1.0` == `1.0.0`, plus epoch/pre/post/dev spellings and metadata-form local segments all match. Critically it FAILS CLOSED: when a requested filename version is not recognizably PEP 440 (e.g. a PEP 427 escaped local version like `1.0+ubuntu_1`), `pypi_version_owned` returns true and suppresses the proxy rather than letting it fan out to the remote. That is the correct security default and closes the dependency-confusion hole I flagged. Unit tests cover the release-padding, local-segment, post, epoch, and fail-safe-on-unparseable cases.

### Non-blocking follow-ups
1. The fail-closed default means the narrowing does not apply to non-PEP-440 / filename-escaped local versions (they revert to name-only suppression), so the #1582 symptom can still recur narrowly for those packages. Acceptable trade-off, but worth one line in the PR body so reviewers know the limit.
2. Consider a short comment in `serve_file` cross-linking #1267, so future readers see that the index union and the download guard are intentionally separate layers.
3. Pre-existing edge case (not introduced here): if two members hold different content under an identical filename, #1267's index advertises the remote sha256 while #1584 would serve the local bytes. Optional low-priority issue.

Approving. Only CI remains.

PR #1586, opened 2026-06-03
fix(migration): preserve Maven/sbt group prefix in artifact paths

### LINKED ISSUE #1585: bug(migration): Maven/sbt artifact paths lose group prefix, breaking resolution
## Summary

The migration worker's path normalisation reconstructs stored paths as `name/version/filename`, which works for flat formats (PyPI, npm, Helm) but strips the Maven group prefix from Maven/sbt/ivy artifacts. Clients request the full GAV path (e.g. `com/example/artifact/version/file.jar`), so the `/maven/` and `/sbt/` download endpoints return 404 for migrated artifacts.

## Root cause

`migration_worker.rs` line ~1044:

```rust
let path_str = match parsed.version.as_deref() {
    Some(ver) if !ver.is_empty() => {
        format!("{}/{}/{}", parsed.name, ver, filename)
    }
    _ => format!("{}/{}", repo_key, artifact_path),
};
```

`parse_maven` extracts `artifact_id` and `version` from the last 3 path segments, discarding the group prefix. The path reconstruction then produces `example-lib_2.12_1.0/0.0.10/file.jar` instead of `com/example/example-lib_2.12_1.0/0.0.10/file.jar`.

## Reproduction

1. Migrate a Maven/sbt local repo from Artifactory containing artifacts under a group path (e.g. `com/example/...`)
2. Verify artifacts appear in the DB with truncated paths (no group prefix)
3. Attempt to resolve via `/maven/{repo}/com/example/artifact/version/file.jar` → 404
4. Note the API download (`/api/v1/repositories/{repo}/artifacts/...`) works because it matches the stored path directly

## Fix

For Maven-family formats, use the original source path (which already has the correct `group/artifactId/version/filename` layout) instead of reconstructing from `name/version/filename`.

```rust
let path_str = match package_type {
    "maven" | "gradle" | "sbt" | "ivy" => {
        artifact_path.to_string()
    }
    _ => match parsed.version.as_deref() {
        Some(ver) if !ver.is_empty() => {
            format!("{}/{}/{}", parsed.name, ver, filename)
        }
        _ => format!("{}/{}", repo_key, artifact_path),
    },
};
```

## Affected formats

Maven, Gradle, sbt, Ivy — any format where the storage path includes a group/namespace prefix that clients use in resolution URLs.

## Verified against

- AK 1.2.0 with patched migration tool (#1575)
- Source: JFrog Artifactory `my-sbt-releases` (1,252 artifacts under `com/example/`)
- Migration completed successfully (1,252/1,252, 0 failed) but `/sbt/` resolution returned 404

ysolomon-plat:
Fixes #1585.

## Summary

- Maven/sbt/ivy migration paths now use the original source path instead of the reconstructed `name/version/filename` form
- The `name/version/filename` reconstruction strips the group prefix (e.g. `com/example/`), causing `/maven/` and `/sbt/` download endpoints to 404

## Change

One `match` arm in `migration_worker.rs` — for Maven-family formats, keep the original `artifact_path` which already has the correct `group/artifactId/version/filename` layout.

## Test plan

- Migrate a Maven/sbt repo from Artifactory with artifacts under a group path
- Verify `/sbt/{virtual}/com/example/artifact/version/file.jar` returns 200
- Verify PyPI/npm migration paths are unchanged (still use `name/version/filename`)

### DIFF
--- a/backend/src/services/migration_worker.rs
+++ b/backend/src/services/migration_worker.rs
@@ -1041,20 +1041,14 @@ impl MigrationWorker {
                 // Falls back to the legacy `<repo>/<source-path>` shape only
                 // when the format-aware parser couldn't recover a version
                 // (unknown format / unparseable filename).
-                let path_str = match package_type {
-                    // Maven/sbt/ivy paths include a group prefix
-                    // (e.g. com/example/artifact/version/file.jar) that
-                    // clients use verbatim when resolving. Reconstructing
-                    // as name/version/filename strips the group and
-                    // breaks resolution via /maven/ and /sbt/ endpoints.
-                    "maven" | "gradle" | "sbt" | "ivy" => artifact_path.to_string(),
-                    _ => match parsed.version.as_deref() {
-                        Some(ver) if !ver.is_empty() => {
-                            format!("{}/{}/{}", parsed.name, ver, filename)
-                        }
-                        _ => format!("{}/{}", repo_key, artifact_path),
-                    },
-                };
+                let path_str = migration_artifact_path(
+                    package_type,
+                    &parsed.name,
+                    parsed.version.as_deref(),
+                    filename,
+                    repo_key,
+                    artifact_path,
+                );
                 // Path-traversal guard: `path_str` is stored verbatim and
                 // used as the load-bearing lookup path, so reject anything
                 // with a leading `/` or a `..` segment rather than writing it.
@@ -1772,6 +1766,26 @@ pub(crate) fn build_source_path(repo_key: &str, artifact_path: &str) -> String {
     format!("{}/{}", repo_key, artifact_path)
 }
 
+fn migration_artifact_path(
+    package_type: &str,
+    parsed_name: &str,
+    version: Option<&str>,
+    filename: &str,
+    repo_key: &str,
+    artifact_path: &str,
+) -> String {
+    match package_type {
+        // Maven/sbt/ivy paths include a group prefix
+        // (e.g. com/example/artifact/version/file.jar) that clients use
+        // verbatim when resolving.
+        "maven" | "gradle" | "sbt" | "ivy" => artifact_path.to_string(),
+        _ => match version {
+            Some(ver) if !ver.is_empty() => format!("{}/{}/{}", parsed_name, ver, filename),
+            _ => format!("{}/{}", repo_key, artifact_path),
+        },
+    }
+}
+
 /// Reject artifact paths that could escape the repository root once stored
 /// verbatim. The migration now persists the raw external path as the
 /// load-bearing `path`, so a crafted export with a leading `/` or a `..`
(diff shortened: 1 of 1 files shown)


brandonrc (commented):
**Verdict: Approve with one note.** The core path fix is correct and properly scoped to Maven-family formats. One pre-existing risk to confirm, and a scope issue to clean up.

### Critical findings

None blocking.

### Verification

- Round-trip is correct. `download`/`serve_artifact` (`backend/src/api/handlers/maven.rs:679-1082`, `1128-1150`) and `download_by_path` (`sbt.rs:62-90`) resolve via `WHERE path = $2`, where `$2` is the URL segment after the repo key (`group/artifact/version/file.jar`). The PR stores `artifact_path` verbatim (`migration_worker.rs:1047`), which is exactly that value. The previous `name/version/filename` shape dropped the `group/` prefix and could never match. Fix is correct.
- Properly scoped. The new arm is gated to `"maven" | "gradle" | "sbt" | "ivy"` (`migration_worker.rs:1049-1054`); all other formats keep the reconstruction. `package_type` reaching the worker is normalized/lowercased (`migration_service.rs:296-302`), so the match is reliable, and `gradle` is correctly Maven-family.
- Data integrity is safe. Resume/dedup is keyed on `source_path` in `migration_items` via `is_item_already_completed` (`migration_worker.rs:453`), unaffected by this change, so re-runs do not double-write or orphan previously migrated artifacts.

### Notes

1. Unsanitized external path is now load-bearing (worth a small guard). `artifact_path` is built from the external source registry's `artifact.path` + `artifact.name` (`migration_worker.rs:437`) and stored directly as the artifacts `path` for Maven-family formats with no `..`/leading-slash/traversal check. Low risk today because the storage key is content-addressed from the checksum (`storage_key_from_checksum`, `migration_worker.rs:859`), so a crafted path cannot escape the storage root, and the download side trims a leading `/` (`sbt.rs:68`). But since this PR makes the raw source path the authoritative lookup key, a one-line guard rejecting/normalizing entries containing `..` or a leading `/` before INSERT is cheap insurance against a crafted export planting a row that shadows a legitimate `group/...` path.

2. Out-of-scope changes. The diff also touches `repositories.rs:1899-1900,2021,3008` (setting `cache_cached_at`/`cache_expires_at`), unrelated to the Maven/sbt path fix and not mentioned in the body. Note that the same `entry.cached_at` edits at lines 2021 and 3008 are reported as compile errors in PR #1575 (no `entry` in scope in `build_artifact_response`/`upload_artifact`). Please split these cache-field edits out of this PR (or confirm they compile here), since multiple PRs are now colliding on the same lines.

### Summary

The path fix itself is correct, scoped, and safe for data integrity. Please drop or document the unrelated `cache_*` edits in `repositories.rs` (they overlap with #1575/#1579/#1588 and at least one variant does not compile), and consider a traversal guard on the now-authoritative external `artifact_path`.

brandonrc (commented):
Re-reviewed. The fix is correct, the scope is now clean, and you have already addressed both notes from my prior review. The only thing blocking merge is the coverage gate, which is a one-test fix.

### Scope check
The unrelated `repositories.rs` cache-field edits are gone; the PR is a single file now (`backend/src/services/migration_worker.rs`, +51/-5). Good.

### Correctness
Confirmed. The new `match package_type` arm returns `artifact_path` verbatim for `maven | gradle | sbt | ivy` (preserving the `com/depop/...` group prefix) and falls through to the unchanged `name/version/filename` reconstruction for every other format. Round-trip verified: sbt resolves via `SELECT id FROM artifacts WHERE repository_id = $1 AND path = $2` (`sbt.rs:245`) and maven uses the same exact-path and `path LIKE` patterns (`maven.rs:165,224,1733`), all of which expect the full group path. No regression to other formats, and resume/dedup is keyed on `source_path` in `migration_items` (unchanged), so previously migrated items are unaffected.

### Path traversal
Already handled in this PR (thanks). `has_unsafe_path()` rejects a leading `/` and any `..` segment and is called before the `INSERT INTO artifacts`, which is the right guard now that `artifact_path` is the load-bearing lookup key. The segment-based `..` check is correct (`my..pkg` allowed, `a/../b` rejected). No change requested.

### Coverage gate (the one blocker)
The gate failed at 59% (16/27 new lines), threshold is 70% on new lines. The two new unit tests cover only `has_unsafe_path`; the remaining uncovered new lines are the `match package_type { ... }` that builds `path_str` and the guard block, which live inline inside the async DB-insert method and are not reachable from a unit test.

Minimal fix: extract the path-selection `match` into a pure free function, for example:

    fn migration_artifact_path(package_type: &str, parsed_name: &str, version: Option<&str>, filename: &str, repo_key: &str, artifact_path: &str) -> String

call it from the insert site, and unit-test its cases:
- maven/gradle/sbt/ivy return `artifact_path` verbatim (group prefix preserved)
- a non-maven format with a version returns `name/version/filename`
- a non-maven format without a version returns `repo_key/artifact_path`

That moves the branch lines into covered territory and clears 70%. Once coverage is green this is good to merge; no other critical issues.

ysolomon-plat:
@brandonrc ready to merge

ysolomon-plat:
Added the maintainer-directed coverage fix for the Maven-family migration path change:

- extracted the migration artifact path selection into a pure helper
- added tests for Maven/Gradle/sbt/Ivy preserving the source artifact path verbatim
- added tests for non-Maven versioned paths using name/version/filename
- added tests for non-Maven fallback paths without a parsed version
- kept the change scoped to the migration worker path-selection coverage request

PR #1588, opened 2026-06-03
fix(repositories): populate cache fields in build_cached_artifact_response

### LINKED ISSUE #1587: bug(repositories): build_cached_artifact_response omits cache_cached_at/cache_expires_at, breaking lib test build
## Summary

`build_cached_artifact_response()` in `backend/src/api/handlers/repositories.rs` constructs an `ArtifactResponse` without the `cache_cached_at` and `cache_expires_at` fields. The `ArtifactResponse` struct declares those fields, so the initializer fails to compile under the test profile:

```
error[E0063]: missing fields `cache_cached_at` and `cache_expires_at` in initializer of `repositories::ArtifactResponse`
  --> backend/src/api/handlers/repositories.rs:1887:5
```

This breaks `cargo test --lib` (the Backend Unit Tests CI job) on a clean checkout.

## Location

`build_cached_artifact_response()` maps a recovered proxy cache entry to an `ArtifactResponse` but does not set the two cache-timestamp fields that other `ArtifactResponse` initializers in the same file set.

## Fix

Set `cache_cached_at: Some(entry.cached_at)` (or the appropriate cache timestamp) and `cache_expires_at` on the initializer so it matches the struct definition. A minimal fix sets both from the cache entry's timing fields; if the entry lacks an explicit expiry, `cache_expires_at: None` is acceptable.

## Impact

Any branch cut from `main` cannot compile the lib test target until this is set, which blocks unit-test CI for unrelated PRs.

ysolomon-plat:
Fixes #1587.

## Summary

`build_cached_artifact_response()` built an `ArtifactResponse` without the `cache_cached_at`/`cache_expires_at` fields the struct declares, so `cargo test --lib` fails to compile on a clean checkout:

```
error[E0063]: missing fields `cache_cached_at` and `cache_expires_at` in initializer of `repositories::ArtifactResponse`
  --> backend/src/api/handlers/repositories.rs:1887
```

## Change

This function maps a recovered proxy cache entry, so it now surfaces `cache_cached_at: Some(entry.cached_at)`. The cache entry tracks no explicit expiry, so `cache_expires_at: None`.

## Why this matters

Blocks the Backend Unit Tests CI job for any branch cut from main until the fields are set.

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -1896,6 +1896,10 @@ fn build_cached_artifact_response(
         download_count: 0,
         created_at: entry.cached_at,
         metadata: None,
+        // This response represents a recovered proxy cache entry, so surface
+        // when it was cached. The cache entry tracks no explicit expiry.
+        cache_cached_at: Some(entry.cached_at),
+        cache_expires_at: None,
     }
 }
 


ysolomon-plat:
Closing as a duplicate of #1579, which predates this and makes the identical fix (`cache_cached_at: Some(entry.cached_at)`, `cache_expires_at: None`) in the same initializer. Happy to see #1579 land — our other PRs (#1575, #1584, #1586) carry the same one-line fix only to compile against the currently-broken main, and will drop it on rebase once #1579 merges.

brandonrc (commented):
**Verdict: Approve.** This correctly fixes the compile break, and the field values are right, not a paper-over.

### Critical findings

None.

### Verification

- Compile break is real. On `main`, `ArtifactResponse` declares both `cache_cached_at` and `cache_expires_at` (`backend/src/api/handlers/repositories.rs:1470,1474`, added by #1541/#1542), but `build_cached_artifact_response` (`repositories.rs:1882`) constructs the struct without them, so `cargo test --lib` fails with E0063 as described. This PR adds both fields.
- Field values are correct. The input here is `CachedArtifactEntry` (`backend/src/services/proxy_service.rs:456`), which carries only `cached_at` (line 469) and has no `expires_at` at all. Expiry lives on the separate `CacheMetadata` struct, available only on the DB-backed metadata path (`repositories.rs:2788-2789`). So `cache_cached_at: Some(entry.cached_at)` surfaces the real value and `cache_expires_at: None` is the only correct choice on this path, consistent with the field's doc comment and `skip_serializing_if = "Option::is_none"`.
- No response-shape regression. Only the two new optional skip-if-none fields are touched.

### Summary

Correct and minimal. Note there is a competing PR #1579 fixing the same break with the identical functional change; #1579 additionally extends the unit test to assert both field values. Both are correct, but to avoid a second collision I would merge one and close the other. Given #1579 carries regression coverage, my recommendation is to merge #1579 and close this as a duplicate, or cherry-pick #1579's test onto this branch. Either way, whichever lands needs to land fast since `main` is currently red.

brandonrc:
Superseded by #1592, which has merged and restored the build. #1592 applies the same fix to `build_cached_artifact_response()` (`cache_cached_at: Some(entry.cached_at)`, `cache_expires_at: None`) and was verified against the full Tier-1 gate set, including `cargo clippy --workspace --all-targets -- -D warnings` (the `Check Rust` gate that kept this PR red). Thanks for jumping on the broken `main` so quickly. Closing as superseded.

PR #1613, opened 2026-06-04
feat(pypi): PEP 708 dependency-confusion mitigation for virtual repos

### LINKED ISSUE #1600: PyPI virtual unions remote versions into the simple index but shadows the download for locally-owned names (inconsistent + dependency confusion)
## Summary

For a PyPI **virtual** repository with a local member and a remote (pypi.org)
proxy member, when a package **name exists in both** — i.e. an *internal*
package in the local member, and an **unrelated public package that happens to
share the name** on the upstream the remote proxies — the virtual is internally
inconsistent and exposes a dependency-confusion problem:

1. **The simple index UNIONS the remote's versions** for that name (so it lists
   the public distributions), but
2. **the wheel/sdist DOWNLOAD through the virtual is bound to the local member**
   and returns `404` for any version only the remote has (no fall-through to the
   remote for that name).

Net effect: the virtual **advertises files in its `/simple/<name>/` index that
its own download endpoint then 404s**, and an internally-owned package name gets
unrelated public versions merged in. With an unpinned requirement, `pip` selects
the highest version — the public one — and then fails to download it.

## Reproduction

Virtual `pypi-virtual` with members (priority order):
- `pypi-local` (local, priority 1) — contains an internal package `acme-sdk`
  at versions `1.0.0`, `1.1.0a1` … (an in-house package)
- `pypi-remote` (remote, priority 2) — proxies `https://pypi.org/simple`, where
  an **unrelated** public `acme-sdk` exists at `2.3.0`

```
# simple index through the VIRTUAL — UNIONS local 1.x AND public 2.3.0
GET /pypi/pypi-virtual/simple/acme-sdk/
  -> 200, lists acme-sdk-1.1.0a1...  AND  acme-sdk-2.3.0-py3-none-any.whl

# downloading the public 2.3.0 wheel through the VIRTUAL (the URL the index points pip at)
GET /pypi/pypi-virtual/simple/acme-sdk/acme-sdk-2.3.0-py3-none-any.whl
  -> 404            # shadowed: virtual binds the name to pypi-local, which lacks 2.3.0

# the SAME wheel through the remote MEMBER directly — proxies fine
GET /pypi/pypi-remote/simple/acme-sdk/acme-sdk-2.3.0-py3-none-any.whl
  -> 200

# control: a package NOT present in pypi-local downloads fine through the virtual
GET /pypi/pypi-virtual/simple/six/six-1.16.0-py2.py3-none-any.whl
  -> 200            # proxy-download through the virtual works in general
```

End-to-end, `pip install acme-sdk` (unpinned) against the virtual:
- resolves to public `2.3.0` (highest version, listed by the unioned index), then
- `HTTP error 404` downloading `acme-sdk-2.3.0-py3-none-any.whl` from the virtual.

## Two problems

1. **Index/download inconsistency.** `/simple/<name>/` lists a distribution the
   virtual's download path refuses to serve. Whatever the resolution policy is,
   the index and the download must agree.
2. **Dependency confusion for internally-owned names.** A name owned by a local
   member should not silently pull in an unrelated public package's versions. An
   unpinned `pip install <name>` then prefers the public (higher) version — both
   a correctness and a supply-chain concern.

## Expected

**Local-precedence / ownership awareness** for the PyPI virtual: when a local
(hosted) member owns a package name, the virtual should serve **only that
member's distributions** for the name — in both the simple index and the
download — and not merge the remote/public versions for it. (This is the PyPI
analog of the Maven groupId+artifactId ownership guard.) Ideally make it
configurable (e.g. local-takes-precedence, or per-virtual include/exclude
patterns), since the opposite policy — unioning versions across members — is
correct only when both members hold the *same* package (split version ranges),
not for a name collision between an internal package and an unrelated public one.

At minimum: keep the index and the download consistent — don't advertise in
`/simple/<name>/` anything the virtual won't serve.

## Relationship to simple-index union behavior

The simple-index union (listing versions contributed by every member) is correct
for a package that genuinely lives in both members with complementary version
ranges, but harmful for a name collision / internal name. The fix needs to be
ownership-aware rather than a blanket union.

## Workarounds (consumer side)

- Pin the package to the intended internal version, and/or
- Remove the direct public index from the client config and rely solely on the
  virtual (which proxies public packages for names the local member does not own).

brandonrc:
Closes #1600

## Summary

A PyPI **virtual** repository could union an unrelated upstream package of the same name into a locally-owned project, in both the `/simple/` index and the download path, so an unpinned `pip install <name>` could resolve to the higher *public* version. That is a dependency-confusion / supply-chain hole (#1600), and it was on by default. #1584 (version-aware fallthrough) fixed the legitimate split-version case (#1582) but widened this exposure.

This implements PEP 708 (Simple Repository API v1.2) **local-precedence** server-side, which is the correct model: the distinguisher is whether the local project is the *same project* as upstream (an explicit operator `tracks` declaration), not whether a version happens to be local.

## What changed

- **Migration 121** `pypi_project_tracks` — operator-controlled `tracks` declarations, attached to the project-owning local/hosted repo (PEP 708: `tracks` is a property of the project's own repository).
- **`proxy_helpers::pypi_virtual_isolates_name()`** — the single predicate both paths use: a local member owns the PEP 503 normalized name AND no `tracks` declared -> **isolate** (serve only the owning member's distributions; no union, no proxy fallthrough). The simple index and the download make the *same* decision, so they can no longer disagree (closes the index/download inconsistency #1600 also raised).
- **Default flips to isolation** (security). A `tracks` declaration re-enables the cross-member union for a genuinely-shared project, preserving the #1582 case as an explicit opt-in.
- **PEP 708 metadata** — Simple API advertises **v1.2**; `meta.tracks` (JSON / PEP 691) and `<meta name="pypi:tracks">` (HTML / PEP 503) are emitted where a relationship is declared.
- **Management API** (admin-scoped): `GET /:key/pypi-tracks`, `PUT/DELETE /:key/pypi-tracks/:project`.

A web UI to manage `tracks` declarations lands in a companion artifact-keeper-web PR (#1600 is backend + UI per the agreed scope).

## Behavior change (called out for release notes)

Mixed virtuals that relied on the old union for a locally-owned name now isolate that name by default. Operators who genuinely mirror a project re-enable the union with a `tracks` declaration (`PUT /api/v1/repositories/{key}/pypi-tracks/{project}`). Documented in CHANGELOG under Security; this is the security-motivated behavior change to highlight when v1.2.1 is cut.

- [ ] This PR is a `fix/*` AND adds/updates a test
- [x] N/A — `feat/`. Includes a DB test (`pypi_virtual_isolates_locally_owned_name_until_tracks_declared`) asserting isolate-by-default, proxy-when-not-owned, and union-when-tracks-declared.

## Test Checklist
- [x] Manually tested locally

Verified against a real Postgres 16: migration 121 applies; `cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` exit 0; the new isolation test passes against the live DB; `cargo test --lib` 10704 passed. The 2 local failures are macOS-only environment artifacts unrelated to this change and pass on Linux CI: an HTTPS localhost-timeout test, and an OCI test asserting an upper-case storage key is absent (macOS APFS is case-insensitive, so it sees the canonical lower-case blob). `oci_v2.rs` is unchanged in this PR.

## API Changes
- [ ] OpenAPI spec validates locally (count-gated test; new paths/schemas only increase counts)
- [x] Migration is reversible-safe (additive table; guarded)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **PyPI virtual repositories now isolate locally-owned project names by default (PEP 708 dependency-confusion mitigation)** (#1600). Previously, when a local member owned a project name, a virtual repo could union an unrelated upstream package of the same name into its `/simple/` index and serve it on download, so an unpinned `pip install <name>` could resolve to the higher public version (a supply-chain hole). A PyPI virtual now serves only the owning member's distributions for a locally-owned name, in both the simple index and the download, unless an operator declares a PEP 708 `tracks` relationship for that project. The Simple API now advertises v1.2 and emits `meta.tracks` / `pypi:tracks` where declared.
+
+  **Behavior change (action may be required):** if you intentionally relied on a mixed virtual unioning a local project's versions with the same project upstream (e.g. split version ranges of the *same* package), declare it via the new endpoint so the union is restored:
+
+  ```
+  PUT /api/v1/repositories/{local_repo_key}/pypi-tracks/{project}
+  { "tracks_url": "https://pypi.org/simple/{project}/" }
+  ```
+
+  Names a local member does not own are unaffected and continue to proxy normally.
+
 ### Pre-upgrade check
 
 - **Stuck-scan janitor will reap accumulated `running` rows on first tick** (#1015, #1062) -- this release introduces a background janitor that transitions `scan_results` rows wedged in `status='running'` past the configured `STUCK_SCAN_THRESHOLD_SECS` (default 1800s / 30 min) to `status='failed'`. On long-running installs predating this release, previously-stuck rows from crashed scan workers (OOM, pod evicted, deploy mid-scan) will flip to `failed` in batches of up to 1000 rows per janitor tick (the per-tick cap bounds memory and audit-log write volume). With the default `STUCK_SCAN_CHECK_INTERVAL_SECS=600`, a backlog of N stuck rows therefore takes roughly `ceil(N / 1000) * (STUCK_SCAN_CHECK_INTERVAL_SECS / 60)` minutes per replica to fully drain. Multiple janitor-running replicas drain proportionally faster (the cap is per-replica, per-tick). Operators with a large backlog who want to accelerate the drain can lower `STUCK_SCAN_CHECK_INTERVAL_SECS` for one or two ticks after deploy and revert. Alerting on `status='failed'` deltas should expect this drain (one or many ticks depending on backlog size) at upgrade and tune accordingly.
--- a/backend/migrations/127_pypi_project_tracks.sql
+++ b/backend/migrations/127_pypi_project_tracks.sql
@@ -1,0 +1,32 @@
+-- PEP 708 (Simple Repository API v1.2) dependency-confusion mitigation.
+--
+-- A `tracks` declaration is an OPERATOR-controlled statement that a locally
+-- owned project is the same project as an upstream one, so it is safe for a
+-- virtual repository to merge (union) versions across members for that name.
+--
+-- Without such a declaration, a virtual repo must NOT merge an externally owned
+-- project into a name a local member owns (that is the dependency-confusion
+-- hole: an internal `acme-sdk` silently pulling in an unrelated public
+-- `acme-sdk`). See #1600.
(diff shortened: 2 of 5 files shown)


ThaSami:
amazing, waiting for this to get merged and update AK to 1.2.1 which unblocks our NPM consumers without the risk of python pipelines gets compromised.

PR #1651, opened 2026-06-05
fix(replication): read artifacts through configured storage backend (#1565)

### LINKED ISSUE #1565: Peer replication ignores configured S3 storage backend and reads artifacts from local STORAGE_PATH

brandonrc:
Closes #1565

## Summary

Peer replication's sync worker read artifact bytes directly from the local
filesystem (`STORAGE_PATH` + `storage_key`) instead of going through the
configured `StorageBackend`. For deployments using object storage
(`STORAGE_BACKEND=s3`/`gcs`/`azure`) the artifact never exists on the local
filesystem, so replication failed with `No such file or directory` even
though the same object was served correctly by the normal download path
(which uses the storage abstraction, e.g. `storage.get(&artifact.storage_key)`).

Root cause: `backend/src/services/sync_worker.rs` — `read_artifact_from_storage`
and `read_artifact_chunk_from_storage` both built a path from the
`STORAGE_PATH` env var and read via `tokio::fs`. The worker was spawned with
only a `PgPool`, so it had no access to the initialized backend.

Fix:
- Thread the configured `Arc<dyn StorageBackend>` (the same `primary_storage`
  used by the rest of the app) into `spawn_sync_worker` and down through
  `process_pending_tasks` -> `execute_transfer` -> `execute_chunked_transfer`.
- `read_artifact_from_storage` now calls `storage.get(storage_key)`, matching
  the download path and working for filesystem/S3/GCS/Azure.
- The chunked path reads the full object once through the backend and slices
  each chunk in memory via a new pure helper `slice_artifact_chunk`, since
  object stores do not expose the filesystem `seek` semantics the old code
  relied on.

Behavior for filesystem-backed deployments is preserved (the filesystem
backend resolves `storage_key` under its root exactly as before).

A unit test (`test_read_artifact_uses_storage_backend_not_filesystem`) reads an
artifact through an in-memory backend with no corresponding file on disk; the
old filesystem-based implementation would have failed this. Additional tests
cover the chunk-slicing helper, including an end-to-end reassembly invariant
against `compute_chunk_ranges`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -544,6 +544,10 @@ pub async fn run_server(shutdown_token: Option<CancellationToken>) -> Result<()>
 
     // Create application state with WASM plugin support
     let scheduler_storage = primary_storage.clone();
+    // The sync worker reads artifact bytes through the configured storage
+    // backend (filesystem/S3/GCS/Azure) so peer replication honors object
+    // storage instead of assuming a local filesystem path (issue #1565).
+    let sync_worker_storage = primary_storage.clone();
     let mut app_state = api::AppState::with_wasm_plugins(
         config.clone(),
         db_pool.clone(),
@@ -685,7 +689,8 @@ pub async fn run_server(shutdown_token: Option<CancellationToken>) -> Result<()>
     let grpc_db_pool = db_pool.clone();
 
     // Spawn background sync worker for peer replication
-    artifact_keeper_backend::services::sync_worker::spawn_sync_worker(db_pool).await;
+    artifact_keeper_backend::services::sync_worker::spawn_sync_worker(db_pool, sync_worker_storage)
+        .await;
     tracing::info!("Sync worker started");
 
     // Conditionally clone state for the metrics listener before the router takes
--- a/backend/src/services/sync_worker.rs
+++ b/backend/src/services/sync_worker.rs
@@ -694,13 +694,23 @@ async fn execute_chunked_transfer(
         .and_then(|s| Uuid::parse_str(s).ok())
         .ok_or_else(|| "Missing session id in transfer init response".to_string())?;
 
-    // 2. Read the full artifact through the configured storage backend, then
-    //    upload it chunk by chunk. Reading once (rather than per chunk) keeps
-    //    the transfer backend-agnostic: object stores (S3/GCS/Azure) do not
-    //    expose the local-filesystem seek semantics the previous code relied
-    //    on, and re-fetching the object for every chunk would be wasteful.
-    let artifact_bytes = match read_artifact_from_storage(storage, &task.storage_key).await {
-        Ok(bytes) => bytes,
+    // 2. Stream the artifact through the configured storage backend and upload
+    //    it chunk by chunk. We deliberately avoid `storage.get()` here: that
+    //    materializes the entire (potentially multi-GB) object in RAM before
+    //    slicing, which defeats the purpose of the chunked path and would OOM
+    //    the host on large Docker images / ML models. Instead we pull a byte
+    //    stream from the backend (S3/GCS/Azure/filesystem all implement genuine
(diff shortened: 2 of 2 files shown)


brandonrc:
Closing in favor of #1564 (the maintainer's broader replication-hardening PR). Reconciliation found #1651 used `primary_storage`, which is subtly incorrect now that per-repo storage backends are confirmed real (`StorageRegistry::backend_for`, `repositories.storage_backend` since migration 003) — it would read the wrong backend for any repo not on the primary. #1564 resolves storage per-repo (`backend_for`), uses native `get_range` for chunked transfers (no full-object buffering), AND carries real hardening #1651 lacks: heartbeat probes, configurable retry/backoff, and delete-loop prevention. The streaming-chunker work here was solid but the primary_storage approach is the wrong foundation. Adopting #1564 for #1565 instead. Thanks — this surfaced the dup and the per-repo correctness issue.

Issue #1617, opened 2026-06-04
Top Defect Source: Auth/SSO correctness consolidation (sessions, callbacks, audit, scope checks)


brandonrc:
Part of #1615 (Top Defect Sources). Source: `docs/audits/honest-issue-analysis-2026-06-04.md`.

## Why
**Auth / RBAC / SSO is the #2 defect component (45 defects) and the dominant security-severity root cause — 28 of 35 `security` defects.** Recurring theme: SSO callback/login wiring, session/token invalidation, partial-update clobbering, and audit gaps.

## Work
- [ ] **Session/token invalidation on credential change.** Sessions/JWTs are not invalidated after a password change — the old password still works (#505, security). Fold into the unified credential-change session model (#1394). gRPC interceptor missing the same invalidation check (#549/#551).
- [ ] **SSO callback/login wiring.** Keycloak OIDC "Sign in" 404s at `/auth/sso/oidc/{id}/login` (#453); OIDC callback 404s at `/auth/callback` (#530). Add SSO login+callback E2E against a test IdP.
- [ ] **PATCH semantics for provider config.** OIDC `PUT` replaces `attribute_mapping` wholesale, dropping unmentioned keys (#1191); also `PUT /security/policies/{id}` only persists the first changed field.
- [ ] **Audit logging.** `AuditService` exists but auth handlers never log login/logout/token events (#386).
- [ ] **Consolidate the authz check** (deny-by-default, single evaluation point). Land the open hardening issues: extract `require_admin`/`require_self_or_admin` (#1314), pin token-minting endpoint count + reject rules (#1315), release-gate check against legacy `any(|s| s == "admin")` scope checks (#1316), `AccessScope` enum refactor (#1417), self-service namespace consolidation (#1313).

## Done when
SSO login+callback and credential-change-invalidation have E2E coverage; provider-config updates are PATCH-safe; every allow/deny is audit-logged; the legacy scope-check pattern is gone and gated in CI.

brandonrc:
## Closeout assessment (verify-before-act, against merged `main` @ 2a7ac44)

Adjudicated each checklist defect from merged code. Summary:

| Defect | Status | Evidence |
|---|---|---|
[...]

Issue #1615, opened 2026-06-04
Epic: Top Defect Sources — SBOM/scanner integration & Auth/SSO correctness


brandonrc:
## Epic: Top Defect Sources — SBOM/scanner integration & Auth/SSO correctness

**Source:** Honest re-classification of all 641 issues from title+body, then re-judged against the actual fix PRs (`docs/audits/honest-issue-analysis-2026-06-04.md`, dataset in `docs/audits/data/`).

Sibling to the invariants epic #1607. Where #1607 owns the **severe but low-volume** tail (~10% of defects), this epic owns the **two largest defect sources by volume** — which the data shows are *not* architecture/invariant problems but **integration brittleness and auth-logic correctness**:

- **SBOM / security-scanning — 50 defects (largest single component).** Grype / Dependency-Track / OpenSCAP integration brittleness + a data-model rot (`cve_history` read by the findings path but never populated).
- **Auth / RBAC / SSO — 45 defects, dominant security-severity root cause (28 of 35 security defects).** SSO callback 404s, session/token invalidation gaps, PUT-vs-PATCH clobbering, audit-log gaps.

Together ~27% of all defects — higher ROI by volume than any single invariant.

### Sub-issues
- [ ] #1616 — SBOM/scanner integration hardening + retire/repair `cve_history`
- [ ] #1617 — Auth/SSO correctness consolidation (sessions, callbacks, audit, scope checks)

### Relationship
- Risk-tail counterpart: epic #1607 (Core Invariants Hardening).
- Both tracked on the **Hardening Core** project board.

brandonrc:
Closing — both children are done; the checkboxes above were simply never ticked as the work landed.

Verified, both **CLOSED**:
- #1616 — SBOM/scanner integration hardening + retire/repair `cve_history`
- #1617 — Auth/SSO correctness consolidation (sessions, callbacks, audit, scope checks)

The companion epic #1607 (Core Invariants Hardening) is closed on the same basis.

Note that "top defect sources" remains a live theme even though this epic is complete — scanner/verdict correctness and auth/SSO work has continued past it (for example #3243/#3306 on severity gating, #3287 on CVE-DB verdict invalidation, and #2936/#3082 on the guest-access auth challenge). Those are tracked on their own issues rather than under this epic.

Issue #1191, opened 2026-05-12
OIDC PUT replaces attribute_mapping wholesale; needs PATCH/partial semantics or dedicated columns


brandonrc:
## Background

Surfaced while fixing artifact-keeper-web#406 (frontend bug: OIDC edit dialog
wipes backend-managed mapping keys on save).

The OIDC `update` endpoint accepts `attribute_mapping: Record<string, unknown>` and stores it as a JSONB column wholesale. Anything the caller doesn't include — for example mapping entries derived from env vars like `OIDC_REDIRECT_URI` — gets dropped on every PUT.

The web UI has been patched to spread the existing `attribute_mapping` before overlaying the form-rendered claim fields (see [artifact-keeper-web@2029561](https://github.com/artifact-keeper/artifact-keeper-web/commit/2029561)). That fixes the user-visible bug, but the underlying API still has merge-by-omission as its update model. Any other consumer (custom UIs, future CLI/Terraform provider) would hit the same trap.

## Proposed change

Pick one:
1. **PATCH semantics** on `PUT /api/v1/sso/oidc/{id}`: treat any field absent from the body as \"don't change\" rather than \"set to null/empty\". Especially for `attribute_mapping`, a missing key should preserve the existing value.
2. **Dedicated columns** for backend-managed fields (e.g. `redirect_uri`) so they don't live inside the JSONB blob that admins can stomp.

Option 2 is the cleaner long-term shape; option 1 is the smaller change.

## Acceptance criteria

- An update request that omits `attribute_mapping` does not clear it.
- An update request that supplies `attribute_mapping` with only some keys does not delete the unmentioned keys (or, if that's deemed wrong shape, those keys live elsewhere and are unaffected).
- Integration test in `backend/tests/...` covering both cases.

## Related

- artifact-keeper-web#406 — the user-visible report
- artifact-keeper-web@2029561 — frontend mitigation already shipped
- #655 — earlier redirect_uri reverse-proxy fix (same field, different angle)

PR #1661, opened 2026-06-05
fix(auth): reconcile env-managed OIDC/LDAP provider on every boot

### LINKED ISSUE #1656: OIDC/LDAP env vars are silently ignored after first boot — no reconcile, no warning, undocumented
## Problem Statement

`OIDC_*` (and `LDAP_*`) environment variables are only honoured on **first boot**, then silently ignored — with no warning and no documentation of this behaviour.

`backend/src/main.rs` → `bootstrap_oidc_from_env()` builds a provider from env, but only persists it when **no** OIDC config already exists in the DB:

```rust
// Only bootstrap when no OIDC configs exist in the database
let existing = AuthConfigService::list_oidc(db).await?;
if !existing.is_empty() {
    tracing::debug!(
        "OIDC env vars present but {} config(s) already exist in DB, skipping bootstrap",
        existing.len()
    );
    return Ok(());
}
```

On every subsequent boot the env vars are discarded at **`debug`** level. There is no env→DB reconciliation.

**Why this is a real-world trap:** the documentation and Helm chart present `OIDC_ISSUER` / `OIDC_CLIENT_ID` / `OIDC_CLIENT_SECRET` as *the* way to configure OIDC (e.g. the configuration reference shows `OIDC_ISSUER=https://accounts.google.com`). An operator reasonably treats env as the source of truth. We migrated an OIDC issuer + client_id from one IdP tenant to another by changing the Helm env values and redeploying. The deploy was clean, pods restarted with the new values present in the container env — **and SSO kept redirecting to the old IdP**, because the original DB provider (seeded at first boot) was untouched. There was zero error signal; the only hint was a `debug` log line. The fix required manually updating the DB provider via the admin API (`PUT /api/v1/admin/sso/oidc/{id}`).

## Proposed Solution

Make **environment variables definitive when set, with the database as the fallback** for manual configuration:

- **If `OIDC_*` env vars are set** → treat them as the source of truth and **reconcile the DB on every boot**: create the env-managed provider if it doesn't exist (current behaviour), and **update it to match env if it has drifted** (new behaviour). This makes Helm/GitOps deployments declarative — changing an env value and redeploying actually changes SSO, which is the behaviour operators already expect from the docs.
- **If `OIDC_*` env vars are not set** → fall back to the DB config, fully manageable via the admin UI/API (current behaviour for UI-first deployments).

Scope notes to keep this safe:
- The reconcile should target only the **single env-managed provider**, identified stably (e.g. by a well-known name/marker). UI-created *additional* providers must not be deleted or modified.
- An env-managed provider is effectively read-only from the UI (edits would be reverted on next boot) — this is the desired GitOps semantic, but it should be surfaced (e.g. a "managed by environment" flag in the admin API / a banner in the UI) so the revert isn't surprising.
- Apply the same model to `LDAP_*` for consistency.

Even short of full reconcile, the **minimum** acceptable fix is: (a) document clearly that env seeds once and the DB is otherwise authoritative, and (b) raise the skip log to `warn`/`info` **when the env values differ from the stored config**, so drift is visible.

## Alternatives Considered

- **Seed-once + warn-on-drift only** (status quo plus logging): surfaces the problem but still requires a manual DB edit to apply env changes — does not make declarative deploys work.
- **Manage the provider purely via admin API / Terraform and drop the env vars from Helm:** workable, but it forces every operator onto external tooling, contradicts the docs that present env as the primary configuration path, and leaves the silent-no-op footgun in place for anyone who does set the env vars. Not preferred.

## Use Case

As an operator running Artifact Keeper via Helm/GitOps, I want the `OIDC_*` environment values to be authoritative so that rotating an IdP tenant or client ID is a config change + redeploy, not a manual admin-API mutation against production — while still allowing UI-first deployments to manage SSO entirely from the database when no env vars are set.

## Component

## Additional Context

- Confirmed against backend `1.2.0`.
- Relevant code: `backend/src/main.rs` (`bootstrap_oidc_from_env`, `bootstrap_ldap_from_env`); `AuthConfigService::{list_oidc,create_oidc}`.
- Docs presenting env as the configuration path: `getting-started/configuration`, `reference/environment`, `advanced/auth`.

rob-howie-depop:
## Summary

`bootstrap_oidc_from_env` / `bootstrap_ldap_from_env` only persisted the env-derived provider when the table was **empty**. After first boot they logged at `debug` and returned, so changing `OIDC_*` / `LDAP_*` in Helm/env and redeploying was a **silent no-op** — surprising operators because the docs present env vars as the configuration path.

This makes the env-managed provider **definitive when set**: on every boot the provider is matched **by name** and **updated** when it already exists, created when absent. Other (UI-created) providers with a different name are never touched. When the env vars are unset, the DB remains the source of truth (UI-managed deployments are unaffected).

The decision is a small pure function (`plan_provider_reconcile`) in the lib crate so it is unit-testable without a database, mirroring the existing `build_oidc_request_from_values` pattern. The `Create -> Update` mapping is a `From<Create…> for Update…` impl placed with the request types (orphan rule). The skip is now an `info`-level reconcile log rather than a silent `debug`.

For OIDC, `attribute_mapping_replace = true` so a key removed from the env clears in the DB (truly env-definitive). Known limitation: `update_ldap` uses `.or(existing)` for `bind_dn` / `group_base_dn` / `group_filter` / `admin_group_dn`, so clearing those four optional LDAP fields via env is not possible — only changed values apply. Fully clearing them would require an `update_ldap` change, out of scope for this minimal fix.

Closes #1656.

`test_plan_reconcile_update_when_name_matches` is the primary regression test: it asserts that when a provider with the env name already exists the planner returns `Update(id)`. On `main` this symbol does not exist (compile failure) and, behaviourally, the old empty-table-only guard could never produce an update — it could only skip. `test_plan_reconcile_create_ignores_other_named_providers` proves UI-created providers neither block bootstrap nor get targeted.

## Test Checklist
- [x] Manually tested locally

New unit tests (run under `cargo test --workspace --lib`): `test_plan_reconcile_create_when_absent`, `test_plan_reconcile_update_when_name_matches`, `test_plan_reconcile_create_ignores_other_named_providers`, `test_oidc_create_to_update_carries_secret_and_replaces_mapping`, `test_ldap_create_to_update_maps_required_fields`. Full lib suite: 10796 passed; 0 failed (the one skipped test is a pre-existing flaky loopback-timeout unrelated to this change).

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -1085,29 +1085,41 @@ async fn init_peer_identity(db: &sqlx::PgPool, config: &Config) -> Result<uuid::
 /// env vars (OIDC_ISSUER, OIDC_CLIENT_ID, OIDC_CLIENT_SECRET, etc.) without
 /// needing admin API access first.
 async fn bootstrap_oidc_from_env(db: &sqlx::PgPool) -> Result<()> {
-    use artifact_keeper_backend::services::auth_config_service::AuthConfigService;
+    use artifact_keeper_backend::services::auth_config_service::{
+        plan_provider_reconcile, AuthConfigService, ReconcileAction,
+    };
 
     let req = match build_oidc_bootstrap_request() {
         Some(r) => r,
         None => return Ok(()),
     };
 
-    // Only bootstrap when no OIDC configs exist in the database
+    // Reconcile the env-managed provider (matched by name) on every boot so
+    // changing OIDC_* env and redeploying takes effect. Other (UI-created)
+    // providers are left untouched.
     let existing = AuthConfigService::list_oidc(db).await?;
-    if !existing.is_empty() {
-        tracing::debug!(
-            "OIDC env vars present but {} config(s) already exist in DB, skipping bootstrap",
-            existing.len()
-        );
-        return Ok(());
-    }
+    let pairs: Vec<(uuid::Uuid, String)> =
+        existing.iter().map(|c| (c.id, c.name.clone())).collect();
 
-    let config = AuthConfigService::create_oidc(db, req).await?;
-    tracing::info!(
-        "Bootstrapped OIDC provider '{}' (id={}) from environment variables",
-        config.name,
-        config.id
-    );
+    match plan_provider_reconcile(&req.name, &pairs) {
+        ReconcileAction::Create => {
+            let config = AuthConfigService::create_oidc(db, req).await?;
+            tracing::info!(
+                "Bootstrapped OIDC provider '{}' (id={}) from environment variables",
+                config.name,
+                config.id
+            );
+        }
+        ReconcileAction::Update(id) => {
+            let name = req.name.clone();
+            let cfg = AuthConfigService::update_oidc(db, id, req.into()).await?;
+            tracing::info!(
+                "Reconciled env-managed OIDC provider '{}' (id={}) from environment variables",
+                name,
+                cfg.id
+            );
+        }
+    }
 
     Ok(())
 }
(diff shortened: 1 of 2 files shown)


brandonrc (commented):
Closeout review (assessment pass for #1617 / epic #1615 — not an approval).

**Verdict: solid.** This is a clean, minimal fix for #1656 and a good fit for the #1617 SSO-wiring workstream.

Verified against the diff and merged main:
- `plan_provider_reconcile` is a pure, DB-free decision function matched **by name**, with the three meaningful unit tests (create-when-absent, update-on-name-match, create-ignores-other-named). UI-created providers with a different name are correctly left untouched.
- **No secret-clobber risk:** `build_oidc_request_from_values` (main.rs) returns `None` unless `issuer`, `client_id`, AND `client_secret` are all non-empty, so the `Update` path never runs with a blank secret. Good.
- `From<CreateOidcConfigRequest> for UpdateOidcConfigRequest` sets `attribute_mapping_replace = Some(true)`, and `update_oidc` (auth_config_service.rs:578) honors that flag — so env truly becomes definitive for the mapping. Consistent with the #1191 PATCH-vs-replace model.
- CI fully green: coverage, security audit, clippy, unit tests all pass.

Two non-blocking notes:
1. The documented LDAP limitation (`bind_dn`/`group_base_dn`/`group_filter`/`admin_group_dn` can't be cleared via env because `update_ldap` uses `.or(existing)`) is an acceptable scope cut; worth a one-line follow-up issue so it isn't silently lost.
2. Consider an info-level log line in the no-op/unset-env case for operator clarity, but not required.

Recommend arming auto-merge once a backend maintainer signs off.

PR #1640, opened 2026-06-04
fix(promotion): repoint open-CVE gate from never-populated cve_history to scan_findings

brandonrc:
Closes #1620

## Summary
Promotion quality-gates read open CVEs via `SELECT DISTINCT cve_id FROM cve_history WHERE artifact_id = $1 AND status = 'open'` (`promotion_policy_service.rs`). The `cve_history` table has **zero writers**, so `open_cves` was always empty and a configured "block on open CVEs" gate **silently passed** — a security gate that did nothing.

This repoints the read at `scan_findings`, the populated source of truth. An artifact's open CVEs are now the unacknowledged, cve_id-bearing findings from its latest completed scan (the same `scan_findings` shape the SBOM read path already uses). A configured block-on-CVE gate now actually blocks.

It also repoints the remaining dead `cve_history` reads in `sbom_service.rs` flagged by the decision doc (`docs/audits/decision-cve_history-2026-06-04.md`):
- `get_cve_history` / `get_cve_history_by_cve_id` now derive entries directly from `scan_findings`; the dead `cve_history` SELECT blocks are removed while the dedupe + `filter_entries_by_repo` behavior is retained.
- `FIXED_CVES_COUNT_{REPO,ALL}_SQL` drop the always-empty `curated_fixed` CTE and keep the `disappeared`-from-`scan_findings` half.

This also addresses the data source behind #1561 and is part of the #1616 umbrella.

**Explicitly NOT in scope (reversible, code-only):**
- The `cve_history` table is **not** dropped (retained per migration 112; the DROP is a separate v1.3.0 item with a data-preserving migration).
- `record_cve` and the `POST /sbom/cve/status/:id` route are **not** removed.
- No migration, no route changes, no DTO changes.

## Regression test (required for `fix/*` PRs)
- [x] This PR is a `fix/*` AND adds/updates a test that would have caught the bug
- [ ] N/A — this is not a bug fix

The new `backend/tests/promotion_open_cve_gating_tests.rs::test_unacknowledged_scan_finding_blocks_promotion` creates an artifact with an unacknowledged `scan_findings` CVE plus a block-on-CVE policy, asserts `cve_history` is empty (reproducing the bug condition that made the old read pass), then asserts `open_cves` is non-empty and promotion is **BLOCKED** (`PolicyAction::Block`). This fails on `main` and passes here. Companion tests cover acknowledged-finding exclusion, latest-scan-only semantics, and `get_cve_history` returning the scan-derived entry.

## Test Checklist
[...]

Issue #1616, opened 2026-06-04
Top Defect Source: SBOM/scanner integration hardening + retire/repair cve_history


brandonrc:
Part of #1615 (Top Defect Sources). Source: `docs/audits/honest-issue-analysis-2026-06-04.md`.

## Why
**SBOM / security-scanning is the single largest defect component — 50 defects.** It is not an architecture problem; it is integration brittleness plus a data-model rot. The recurring root causes (from the labeled dataset): error-handling, config/integration, data-integrity, and a never-populated table.

## Work
- [ ] **Repoint dead `cve_history` reads → `scan_findings`** (decision: `docs/audits/decision-cve_history-2026-06-04.md`). The table is never populated, so reads return CVE false-positives as NOT_FOUND (#1561) and the promotion gate silently passes (#1620). Repoint the reads — **do NOT drop the table** (kept per `migration 112`; the DROP is a separate v1.3.0 item with a data-preserving migration). Code-only, reversible.
- [ ] **Harden the Grype integration.** Seeded DB ages out after ~5d and auto-update refetch fails in egress-restricted runners (#1251); localhost OCI image resolution failures (#1563, open).
- [ ] **Harden Dependency-Track integration.** Project name set to UUID instead of artifact name so vulns don't map back (#1276); `dtrack-init` can't provision an API key on DT 4.x because keys are masked (#978); make DT opt-in/config explicit (#1432).
- [ ] **Fix scan dedup TTL** reusing completed-but-zero-finding rows from silently-broken scans, masking re-scans for 30 days (#1469).
- [ ] **Scan-on-proxy** doesn't auto-run on pull; only manual works (#1274, open).
- [ ] **Distinct terminal status** for "scanner does not apply to this format" instead of rendering as `failed` (#1470, open).
- [ ] **OpenAPI/handler drift**: `list_findings` advertises a 404 the handler never produces (#914).
- [ ] Submit SBOM to DT even when 0 findings (#965).

## Done when
Scan/SBOM E2E covers Grype + DT + OpenSCAP happy-path and the egress-restricted / zero-finding / wrong-format cases; `cve_history` reads are repointed at `scan_findings` (table retained for v1.3.0 drop); open SBOM defects (#1561, #1620, #1274, #1563, #1470) closed with regression tests.

brandonrc:
## Decision: DROP `cve_history`, repoint reads at `scan_findings`

[...]

Issue #1483, opened 2026-05-29
bug(scanner): Grype/Trivy CLI use ":" separator for digest refs instead of "@", causing "could not parse reference" on all digest-pinned OCI artifacts


arthur-sommer-etc:
### Description

`GrypeScanner::build_registry_image_ref` and `ImageScanner::extract_image_ref` always separate the image name and reference with `:`, producing an invalid OCI reference when the artifact is referenced by digest. The correct separator for digest references is `@`.

**Invalid (current):** `localhost:8080/org/myapp:sha256:8c27668f...`
**Valid (expected):** `localhost:8080/org/myapp@sha256:8c27668f...`

OCI distribution spec and all container tooling (Docker, Grype, Trivy CLI) require the `name@digest` format for digest-pinned references. The `:` form is only valid for tags, which cannot contain colons.

### Environment Information

- Deployment type: docker compose
- Operating System: ubuntu
- Architecture: x86_64
- Database: default
- Reverse Proxy: default
- Network setup: default

### Backend Version

1.2.0-rc.2

### Web UI Version

1.1.3

### Package Format

Docker / OCI

### Expected Behavior

Digest-referenced artifacts should be scanned using the `name@sha256:<digest>` format, e.g.:
```
grype registry:localhost:8080/org/myapp@sha256:8c27668f...
```

### Observed Behavior

Every Grype and Trivy scan of a digest-referenced OCI artifact fails. A single `docker push` with BuildKit creates **3 artifacts** in Artifact Keeper — a tag-based image index and two digest-referenced manifests (platform manifest + attestation manifest). The two digest-referenced artifacts fail for both scanners; the tag-based index succeeds.

| Artifact | Content-Type | Trivy | Grype |
|---|---|---|---|
| `v1.0` (tag / image index) | `vnd.oci.image.index` | ✅ completed | ✅ completed |
| `sha256:cf4501...` (platform manifest) | `vnd.oci.image.manifest` | ❌ failed | ❌ failed |
| `sha256:8c2766...` (attestation manifest) | `vnd.oci.image.manifest` | ❌ failed | ❌ failed |

Trivy CLI mode has the identical bug as Grype, producing the same "could not parse reference" error for digest-pinned manifests (Trivy HTTP API mode is unaffected as it constructs the reference differently).

### Steps to Reproduce

1. Push a Docker/OCI image built with `docker buildx` to an Artifact Keeper registry:
   ```sh
   docker pull alpine:3.20
[...]

PR #1664, opened 2026-06-05
Pass repository to grype scanner for oci artifacts

### LINKED ISSUE #1658: Grype fails with NAME_UNKNOWN repository not found because repository is missing from the image reference
### Description

I've added DOCKER_CONFIG variable to my artifact-keeper backend deployment to make authentication work.
I've triggered a scan for all artifacts in a repository called neondb

Example of a failing scan:
localhost:8080/neondb-operator:6fdd2b47314035c864e4e4ecdecceb43e0b10561
I can reproduce this within the container as well, it seems the repository part is missing

running grype localhost:8080/neondb-operator:6fdd2b47314035c864e4e4ecdecceb43e0b10561 indeed fails with a NAME_UNKNOWN

running grype with localhost:8080/**neondb**/neondb-operator:6fdd2b47314035c864e4e4ecdecceb43e0b10561 succeeds as expected

### Environment Information

- Deployment type:  Kubernetes v1.36.1
- Operating System:  Talos Linux
- Architecture:  x86_x64
- Database:  postgres
- Reverse Proxy:  istio
- Network setup:  LB->istio-ingress->virtualservice->artifact-keeper-backend

### Backend Version

1.2.0

### Web UI Version

1.2.0

### Package Format

Docker / OCI

### Expected Behavior

I expect the repository name to be included in grype scans

### Observed Behavior

repository name is missing from grype scan

### Steps to Reproduce

1. trigger a grype scan for a repository (perhaps with multiple images in it)
2. Observe that the scan fails

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

NickyHeuperman:
## Summary
Fixes #1658
This PR fixes Grype scanning for OCI image artifacts by constructing registry pull targets with the owning repository key included. Grype now receives targets like `registry:<host>/<repo_key>/<image>:<tag>` or `registry:<host>/<repo_key>/<image>@sha256:<digest>` instead of targets that omit the repository route.

## What changed

- Added scanner target context so scan orchestration can pass repository identity alongside the artifact being scanned.
- Updated Grype OCI target construction to qualify image names with the owning repository key before invoking Grype registry mode.
- Preserved existing tag and digest handling so digest references continue to use the `@sha256:<digest>` form.
- Kept non-OCI Grype scans on the existing filesystem scan path.
- Added regression coverage for direct tag references, digest references, explicit registry host overrides, peer endpoint fallback, credential stripping, malformed OCI paths, and repository-key edge cases.

## What this fixes

- Fixes OCI Grype scan failures caused by Artifact Keeper asking Grype to pull `/v2/<image>/...` when the registry route actually requires `/v2/<repo_key>/<image>/...`.
- Prevents direct local or remote OCI repository scans from failing with missing-name or `NAME_UNKNOWN`-style registry errors caused by an incomplete pull reference.
- Avoids generating invalid digest pull references such as `<image>:sha256:<digest>`.

## Why this is needed

OCI artifact paths are stored as repository-internal paths, while external registry pulls must include the repository key in the route. The previous Grype target builder only had the internal artifact path, so it could not reconstruct the external pull identity for direct repositories. Passing repository context into scanner execution restores the missing route information without changing stored artifact paths or the artifact model.

## Tests and validation

- Ran targeted Grype scanner tests covering OCI registry reference construction.
- Ran scanner service tests covering repository context propagation and scanner compatibility.
- Verified digest references use `@` separators and tag references use `:` separators.
- Verified non-OCI Grype scans continue to use filesystem scanning.
- Verified malformed OCI paths are still rejected instead of producing unsafe or invalid Grype targets.


brandonrc:
Triaging this into the **SBOM scanner-integration epic** — #1615 (Top Defect Sources) → #1616. It fixes #1658, which is squarely in the #1 defect area from our issue analysis (SBOM/scanning, the largest defect component), and it's complementary to the other SBOM work landing (cve_history repoint #1640, scan-on-proxy surfacing #1650). An independent review is underway; findings to follow. Thanks for this, @NickyHeuperman — solid fix for a real OCI-route bug.

brandonrc (commented):
## Independent review — ✅ correct & safe to merge (after rebase)

Reviewed against the codebase. Triaged into the SBOM scanner-integration epic (#1615 → #1616).

**Verified correct:**
- **Target route is right.** `build_registry_image_ref_for_repo` produces `registry:<host>/<repo_key>/<image>:<tag>` (and `@sha256:<digest>` for digests), which matches AK's own OCI route — `resolve_repo` splits `image_name` at the first `/` into `(repo_key, image)`, so a Grype pull of `<host>/<repo_key>/<image>` re-derives the original route. Confirmed the stored artifact `path` omits the repo key, so the DB `key` JOIN is genuinely needed. Resolves #1658's `NAME_UNKNOWN`. Tag-vs-digest handling preserved (the #1483 regression test).
- **No credential leak.** Grype auth comes from docker config (`DOCKER_CONFIG`), never embedded in the target; `resolve_registry_host` strips any `user:pass@` prefix so secrets can't land in the logged target or the JSON-report `target` field on error. The cred-strip test confirms it.
- **Non-OCI scans unaffected** — the OCI branch is gated by `is_oci_image_artifact`; everything else delegates to the unchanged legacy path (Trivy etc. untouched).
- **Robust** — malformed OCI paths return `None` → `AppError::Internal`, no panics on data-controlled input.

**Should-fix (non-blocking, here or fast follow-up):**
1. The change swaps a compile-time-checked `query_as!` for a runtime `query` + ~50 lines of manual `try_get` mapping — this loses sqlx compile-time column verification and is a duplication-gate risk. Cleaner: keep `query_as!` for the artifact and add a tiny `query_scalar!` for `key` + `repo_type`, or a small joined struct.
2. `_repository_type` is threaded through but unused — drop it or use it.

**Before merge:** rebase onto `main` (it's behind) and let the full CI gate go green (coverage ≥70% on changed lines, duplication ≤3%, security audit, CodeQL) — no `--admin` bypass.

Thanks @NickyHeuperman — clean fix for a real OCI-route bug. 🙌

brandonrc:
Pushed `7dc3b5e` (preserving original authorship) to clear the Code Coverage gate. Two fixes:

1. **Coverage gate (changed-line):** reverted the runtime `sqlx::query` + manual `try_get` mapping back to compile-time-checked `query_as!(Artifact, …)`, plus a small `query!` for the two routing columns the PR needs. Restores column verification and removes the uncovered manual-mapping lines. Changed-line coverage 38% → **75%** (scanner_service.rs 27% → 85%).

2. **Latent test bug (the real reason the gate never ran):** the Coverage job was dying at `exit 101`, not just failing the threshold. `test_scan_artifact_inner_supplies_repository_context_to_scanners`'s `RecordingContextScanner` used `scan_type() = "recording-context"`, which violates the `scan_results_scan_type_check` constraint when the orchestrator persists a row. This same test failed in the original CI coverage run too (`10905 passed; 1 failed`). Changed it to an allowed value (`"grype"`); the test asserts only on recorded key/type/path so its meaning is unchanged.

Heads-up: this test only runs in CI's coverage job (which sets `DATABASE_URL`), not the unit-tests job — which is why the constraint violation was invisible until coverage ran. `.sqlx` regenerated (one new cache entry). fmt/clippy/`--lib` (10904 passed) all clean.

PR #1670, opened 2026-06-05
fix(ci): resolve clippy -D warnings lints that fail on current stable

### LINKED ISSUE #1669: main fails `clippy -D warnings` on current stable (const_is_empty, collapsible_else_if) — intermittent via sccache
## Problem Statement

`main` does not pass `cargo clippy --workspace --all-targets -- -D warnings` on the **current** stable toolchain. Two lints that are warn-by-default on recent stable fire as errors under `-D warnings`:

- `backend/src/storage/azure.rs:1208` — `clippy::collapsible_else_if` (`else { if .. }` block can be collapsed)
- `backend/src/api/handlers/oci_v2.rs:8712` — `clippy::const_is_empty` (`assert!(!ANONYMOUS_TOKEN.is_empty())` where `ANONYMOUS_TOKEN: &str = "anonymous"` — the expression is always `false`... i.e. the assert is trivially true, but clippy flags the const `is_empty()` call)

These are the only two; a cold `clippy -D warnings` on a clean checkout reports exactly these and nothing else.

**Why this hasn't been caught / why it's intermittent:** the `🦀 Check Rust` CI job uses `dtolnay/rust-toolchain@stable` (unpinned) and `sccache`. When the clippy artifacts for `azure.rs`/`oci_v2.rs` are served from a warm sccache, the lints don't re-evaluate and the job passes; when those two files are compiled cold (e.g. a PR whose cache key misses for them), the now-default lints fire and the job fails. The result is that **the same `main`-based code passes on one PR and fails on another**, with the failure pointing at files the PR never touched. I hit exactly this: two unrelated fix PRs against current `main`, identical trees w.r.t. these files — one went green on `Check Rust`, the other red, purely on cache state. `main`'s own CI is green only because it hits the warm cache.

This erodes trust in the gate (a red `Check Rust` that has nothing to do with the PR's diff) and will recur for every contributor who cold-compiles these files, getting worse as stable advances and promotes more lints.

## Proposed Solution

1. **Fix the two current lints** (this PR): collapse the `else if` in `azure.rs`; change the `oci_v2.rs` test to `assert_eq!(ANONYMOUS_TOKEN, "anonymous")` (asserts the real intent and sidesteps `const_is_empty`). Both are behaviour-preserving; the `azure` `put_stream` tests and the `anonymous_token` test still pass. After this, `clippy -D warnings` is clean cold.

2. **Pin the CI toolchain** (recommended follow-up, maintainer policy call — *not* in this PR): point `dtolnay/rust-toolchain@stable` at a pinned version (e.g. `@1.90` or a `rust-toolchain.toml` at the repo root) so a stable bump that promotes new default lints is an explicit, reviewed change rather than a surprise red on an unrelated PR. Unpinned `@stable` + `-D warnings` means any future lint promotion silently becomes a merge-blocker the moment a cache miss exposes it.

## Alternatives Considered

- **`#[allow(...)]` on the two sites** — hides the lint instead of fixing trivially-correct code; rejected.
- **Drop `-D warnings`** — loses the value of the gate; rejected.
- **Pin only, don't fix** — leaves `main` failing on contributors' current-stable local clippy; the lints should be fixed regardless of whether the toolchain is pinned.

## Component

Docker / Deployment (CI)

## Additional Context

Found while landing two unrelated OIDC fixes (#1656 / #1657). The fix PR for this issue touches only the two flagged lines.

rob-howie-depop:
## Summary

`main` does not pass `cargo clippy --workspace --all-targets -- -D warnings` on the current stable toolchain — two warn-by-default lints fail under `-D warnings`:

- `backend/src/storage/azure.rs:1208` — `clippy::collapsible_else_if`
- `backend/src/api/handlers/oci_v2.rs:8712` — `clippy::const_is_empty` (`assert!(!ANONYMOUS_TOKEN.is_empty())`)

Because the `🦀 Check Rust` job uses unpinned `dtolnay/rust-toolchain@stable` + `sccache`, these only surface on a **cold** compile of those files, so `Check Rust` passes or fails on cache state for otherwise-identical `main`-based code (observed across two unrelated PRs).

This collapses the `else if` in `azure.rs` and changes the `oci_v2.rs` test to `assert_eq!(ANONYMOUS_TOKEN, "anonymous")` — asserting the constant's real value, which is the test's actual intent, and sidestepping `const_is_empty`. Both changes are behaviour-preserving. After this, `clippy -D warnings` is clean on a cold checkout.

See #1669 for the full analysis, including the recommended follow-up to **pin the CI toolchain** (out of scope here — maintainer policy decision).

Closes #1669.

This is a lint/CI fix. The bug is a clippy gate failure, not runtime behaviour, so the "test" is the clippy gate itself going green on a cold compile (verified locally: `cargo clippy --workspace --all-targets -- -D warnings` → 0 warnings on a clean target). The existing `test_anonymous_token_is_non_empty` and the `azure::tests::*` `put_stream` tests guard that the touched code's behaviour is unchanged (all pass).

## Test Checklist
- [x] Manually tested locally

`cargo fmt --all -- --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean (was 2 errors on cold compile before this change); affected lib tests (`anonymous_token`, `azure`) 77 passed / 0 failed.

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -9098,7 +9098,7 @@ mod tests {
 
     #[test]
     fn test_anonymous_token_is_non_empty() {
-        assert!(!ANONYMOUS_TOKEN.is_empty());
+        assert_eq!(ANONYMOUS_TOKEN, "anonymous");
     }
 
     // -----------------------------------------------------------------------


brandonrc:
Rebased onto current `main` and verified locally.

**Note on scope after rebase:** the `azure.rs` `collapsible_else_if` hunk from the original PR was dropped because that exact `} else if let Err(e) = self.commit_stream_blocks(...)` collapse already landed on `main` (via #1665). The remaining change is the `oci_v2.rs` `const_is_empty` fix (`assert_eq!(ANONYMOUS_TOKEN, "anonymous")`), which is the intent-preserving fix for that test.

**#1669 ↔ dependabot connection confirmed:** the failing `🦀 Check Rust` jobs on #1462 (rustls) and #1459 (tracing-appender) fail compiling the `artifact-keeper-backend (lib test)` crate under `clippy-driver ... -D warnings` — the crate where `const_is_empty` (in `test_anonymous_token_is_non_empty`) fires. Landing this should let those PRs go green once they pick up main.

Gates on rebased branch (toolchain 1.94.0 stable, cold compile):
- `cargo fmt --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings` -> exit 0, 0 warnings
- `anonymous_token` (7) + `azure` (74) lib tests pass

Issue #1683, opened 2026-06-05
refactor(oci): track manifest existence in a first-class oci_manifests table instead of deriving it from oci_tags


inspired-geek:
## Summary

The registry has **no first-class record of manifest existence**. A manifest's existence and its `content_type` are both derived from `oci_tags` (`oci_tags.manifest_digest`, `oci_tags.manifest_content_type`); `oci_manifest_refs` and `manifest_blob_refs` are edge tables, not an existence registry; `oci_blobs` tracks blobs, not manifests. Manifests are effectively enumerated as the union of (non-index) `oci_tags` rows + `oci_manifest_refs.child_digest` (see #1642).

Because existence is inferred from tags, a manifest with no surviving tag is invisible to the registry even though its bytes are in storage. This is the **root of the bug class** behind #1681 (digest pull/delete depends on a tag) and #1681's retention sibling (lifecycle can orphan the sole protecting tag).

## Proposal

Introduce a first-class `oci_manifests` table that records each manifest object independently of any tag, e.g.:

- `repository_id`
- `digest` (PK with `repository_id`)
- `content_type` (media type, sniffed at PUT)
- `size_bytes`
- `subject_digest` (for the referrers API / OCI 1.1)
- `created_at`, `last_referenced_at`

`oci_tags` would reference a manifest by `(repository_id, digest)`; `oci_manifest_refs` (index → child) and `manifest_blob_refs` (manifest → blob) become edges off this table. Manifest resolution (GET/HEAD/DELETE by digest), the referrers API, and the storage-GC orphan predicate would key off `oci_manifests` presence/refcount rather than reconstructing existence from tags.

This removes the need for the content-addressable storage fallback that #1681 adds as an interim fix, makes the storage-GC orphan predicate a clean ref-count, and gives a natural home for the manifest `content_type` (currently only on `oci_tags`) and for referrers/`subject`.

## Scope / migration

- New migration adding `oci_manifests` (+ indexes, FKs from `oci_tags`/`oci_manifest_refs`/`manifest_blob_refs`).
- Startup backfill from existing `oci_tags` + `oci_manifest_refs` + storage scan (mirrors the existing `oci_manifest_refs` / `manifest_blob_refs` backfills).
- Migrate `handle_put_manifest` to upsert `oci_manifests`, and GET/HEAD/DELETE + GC + referrers to read from it.
- Pairs with #1413 (replace the string-shaped `storage_key` coupling with real FKs).

## Related

[...]

Issue #1682, opened 2026-06-05
fix(lifecycle): retention can delete the sole oci_tags row protecting a manifest, orphaning a still-referenced digest


inspired-geek:
## Summary

The lifecycle retention cascade in `backend/src/services/lifecycle_service.rs` operates on a **per-tag** (`(repository_id, name, tag)`) model, but a single manifest object (`oci-manifests/<digest>`) can be referenced by **many** tags (and as a child of an index via `oci_manifest_refs`). There is **no guard** preventing retention from deleting the *sole* `oci_tags` row protecting a manifest object, leaving a still-referenced digest untagged and eligible for storage GC.

This is the same untagged-but-still-referenced gap that #1681 makes retrievable by digest, reached here through retention rather than an explicit `DELETE`.

## Mechanism

`execute_*` policies (`execute_max_age`, `execute_max_versions`, `execute_no_downloads`, `execute_tag_pattern_keep`/`_delete`, `execute_size_quota`) soft-delete `artifacts` rows (`UPDATE artifacts SET is_deleted = true …`), matching by `repository_id` + age/grouping/size or a regex on `artifacts.name` (`"{image}:{reference}"`). Then `CASCADE_OCI_TAGS_SQL` removes the matching `oci_tags` rows:

```sql
DELETE FROM oci_tags ot
USING artifacts a
WHERE a.is_deleted = true
  AND a.repository_id = ot.repository_id
  AND a.storage_key = 'oci-manifests/' || ot.manifest_digest
  AND a.path = 'v2/' || ot.name || '/manifests/' || ot.tag
  AND a.version = ot.tag
  AND ($1::UUID IS NULL OR a.repository_id = $1)
```

There is no `NOT EXISTS (another live oci_tags row with the same manifest_digest)` check, and the cascade never reads `content_type` / `is_index_content_type` nor consults `oci_manifest_refs`.

## Why it's a bug

- **Digest still tagged elsewhere:** if digest `D` is tagged `v1` and `latest`, a policy that soft-deletes the last surviving tag-artifact for `D` removes the last `oci_tags` row for `D`. The storage-GC `ORPHAN_PREDICATE_SQL` then sees no tag and treats the object as reclaimable.
- **Index children:** the GC's `oci_manifest_refs → oci_tags` protection for per-arch children holds only while the **parent index is still tagged**. If retention untags a multi-arch index, an entire family of child manifest objects loses protection and becomes GC-eligible, even though the policy never named them.
[...]

Issue #1659, opened 2026-06-05
Blob GC under-reclaims: manifest_blob_refs not removed on untag/manifest-delete (#1655 follow-up)


brandonrc:
**Follow-up from the #1655 blob-GC review.** `manifest_blob_refs` rows are never deleted on untag / manifest-delete (only CASCADE on full repo drop). This is the *safe* direction (monotonic refs = a digest is only GC-orphan if it never had a ref), and it's why #1655's deletion is safe — but it means **blobs from deleted/untagged manifests are never reclaimed** (they keep a protecting ref forever). So GC under-reclaims.

Fix: when a manifest is deleted/untagged (and no other manifest references it), remove its `manifest_blob_refs` rows so its now-unreferenced blobs become GC-eligible after the grace window. Must be careful to only drop refs for a manifest that's genuinely gone, and keep it monotonic-safe (don't drop a ref a live manifest still needs). Document the reclaim boundary for operators meanwhile.

brandonrc:
Verified — this is already handled on both paths:

- **Manifest/tag delete:** `delete_manifest_blob_refs(&mut *tx, repo.id, &digest)` runs inside the delete transaction (`backend/src/api/handlers/oci_v2.rs:6964`), with a `NOT EXISTS` guard that preserves a digest still live as an index child.
- **Tag-overwrite / lifecycle expiry:** `prune_orphan_blob_refs()` sweeps stale `manifest_blob_refs` at the start of `run_blob_gc` (`backend/src/services/storage_gc_service.rs:516`, impl :665).

Covered by `test_delete_manifest_removes_blob_refs` and `delete_manifest_blob_refs_preserves_live_index_child` (`oci_v2.rs:11167`). No under-reclaim remains; closing as already-fixed (follow-up to #1409).

Issue #1610, opened 2026-06-04
Core Invariant ③: blob lifecycle & garbage collection


brandonrc:
Part of #1607 (Core Invariants Hardening). Source: `docs/audits/design-retro-2026-06-04.md`.

## Invariant
Blobs have a real lifecycle: ref-counted **mark-and-sweep GC** with a grace window, single-leader via advisory lock, dry-run + audit log. **Delete from object storage only after the metadata transaction commits.** Bias toward leaking storage over losing data.

## Why
GC is missing today: orphaned blobs, ghost artifacts, and storage that is never reclaimed; S3-backed deletes don't remove objects.

## Work
- [ ] Ref-count (or `blob_ref` join table) + grace-period mark-and-sweep GC.
- [ ] Single-leader GC (advisory lock), dry-run mode, deletion audit log.
- [ ] Reclaim `oci_blobs` rows + objects (#1408); clean orphaned blob on failed OCI upload (#1527).
- [ ] Repository delete on S3 backend actually deletes files (#1551).
- [ ] Admin reindex/recover endpoint for ghost artifacts (#1570).

## Linked open issues
#1408, #1527, #1551, #1570

---
**Swept-in open defect:** #1550 — `DELETE /repositories/{key}` returns 500 (DATABASE_ERROR) on repos with ~13.5k artifacts; the delete/reclaim path needs to scale (batch + the GC machinery above).

brandonrc:
Closing — blob lifecycle & GC is complete on main:
- **GC core:** read-only footprint report (#1621), then ref-counted reclaim on the merged `manifest_blob_refs` table with readiness gate + grace window + single-leader scheduler (#1655).
- **Reclaim actually fires (#1408, ~403 GB leak):** completeness layer — prune `manifest_blob_refs` when manifests die + record refs for proxy-cached image manifests + atomic `persist_tag_and_refs` (#1409).
- **Repo-delete reclaims storage incl. S3 (#1551/#1550):** backend-aware per-repo object purge + FK cascades (#1598).
- **OCI orphan-on-failed-commit (#1527):** final blob key journaled before write so a failed `oci_blobs` commit can't orphan it, with GC predicates guarded against reclaiming live blobs (#1711).

All linked defects (#1408, #1551, #1527) closed with regression tests. Part of epic #1607.

PR #1684, opened 2026-06-05
fix(oci): serve and delete manifests by digest independent of tags

### LINKED ISSUE #1681: fix(oci): manifest pull/delete by digest depends on a surviving tag (OCI Distribution Spec violation)
## Summary

For a **hosted (`local`/`staging`) repository**, `GET`/`HEAD`/`DELETE /v2/<name>/manifests/<digest>` are served **only if a live `oci_tags` row still points at that `manifest_digest`**. The manifest bytes themselves are stored content-addressably at `oci-manifests/<digest>`, but the only index that lets a digest reference resolve is an `oci_tags` row. The moment the protecting tag is moved (overwrite via `PUT`), deleted, or never existed (an index child / an untagged push), the registry returns `404 MANIFEST_UNKNOWN` even though the object is intact in storage.

This violates the OCI Distribution Spec: a manifest that exists in storage MUST be retrievable by its digest until it is explicitly deleted.

**Severity:** High — breaks every digest pin (`@sha256:…`): Tekton bundle resolver (`CouldntGetPipeline`), child manifests of multi-arch indexes, and referrers.

## Reproduction (confirmed against `main`)

1. `PUT /v2/<repo>/manifests/<tag>` with body **A** → digest `sha256:AAAA`, object `oci-manifests/sha256:AAAA` written, `oci_tags` row created. `GET …/manifests/sha256:AAAA` → **200**.
2. Overwrite the same `<tag>` with body **B** (`PUT`, `INSERT … ON CONFLICT (repository_id, name, tag) DO UPDATE SET manifest_digest = EXCLUDED.manifest_digest`) → the tag now points at `sha256:BBBB`.
3. From now on `GET`/`HEAD …/manifests/sha256:AAAA` → **404 MANIFEST_UNKNOWN**, although `oci-manifests/sha256:AAAA` is still on disk. Any consumer pinned to `@sha256:AAAA` breaks (Tekton: `CouldntGetPipeline`).
4. Re-pushing body **A** recreates the `oci_tags` row → `GET` returns **200** again. This proves the lost item is the metadata pointer, not the content.

Not garbage collection: the storage GC orphan scan is driven by soft-deleted `artifacts` rows. On a tag overwrite the artifact row for tag `<tag>` is repointed (via `ON CONFLICT (repository_id, path) DO UPDATE`) to digest **B**, so no `artifacts` row references `oci-manifests/sha256:AAAA` at all — the object is never even a GC candidate, it simply becomes unreachable by digest. The bytes are intact (step 4 proves it).

## Expected behavior (OCI Distribution Spec)

- `GET`/`HEAD` by digest return the manifest whenever the object exists in storage at its digest address, regardless of any tag.
- `404 MANIFEST_UNKNOWN` only when the object genuinely does not exist (explicitly deleted or reclaimed by GC).
- `DELETE` by digest removes an untagged manifest.

## Root cause (`backend/src/api/handlers/oci_v2.rs`)

Manifest *existence* is expressed solely by an `oci_tags` row. There is no first-class manifest table; `oci_blobs` tracks blob objects (`oci-blobs/<digest>`), not manifests, and the manifest `content_type` lives only in `oci_tags.manifest_content_type`. So both *existence* and *content-type* of a manifest are bound to a tag.

- `handle_get_manifest` — the digest lookup is `SELECT manifest_digest, manifest_content_type FROM oci_tags WHERE repository_id = $1 AND manifest_digest = $2 LIMIT 1`. The `storage.get(manifest_storage_key(...))` call is locked **inside** `if let Some((manifest_digest, content_type)) = local_result { … }`. On `None` the `else` branch only logs and falls through to the proxy path; for a hosted repo that path is empty → `MANIFEST_UNKNOWN`.
- `handle_head_manifest` — identical locked structure.
- `handle_delete_manifest` — `DELETE` by digest is also tag-gated: it resolves the digest via `SELECT manifest_digest FROM oci_tags … LIMIT 1`, returns `MANIFEST_UNKNOWN` when absent, and only ever deletes `oci_tags` rows. It never touches the storage object, so an untagged manifest can be neither pulled nor deleted by digest.
- `manifest_storage_key(digest)` = `oci-manifests/<digest>` — the object key for a manifest is its digest, i.e. the request reference for a digest pull already *is* the storage key.

## Proposed fix (minimal, OCI-conformant)

Add a **content-addressable fallback** for digest references on hosted (`local`/`staging`) repositories, taken only when no `oci_tags` row resolves:

1. In `handle_get_manifest` / `handle_head_manifest`: when `is_digest_reference(reference)`, no tag row resolved, and the repo stores its own manifests (not `Remote`/`Virtual`), try `storage.get(manifest_storage_key(reference))`.
2. On success, return `200` (HEAD without body) with `Docker-Content-Digest: <reference>`, `Content-Length`, `Content-Type`.
3. Verify `sha256(body) == reference` before serving (defence against a corrupted object) — reuse the existing `verify_digest_or_fall_through` / `compute_sha256`.
4. Content-type is not stored for an untagged object; derive it from the manifest body's `mediaType` field (`crate::formats::oci::OciHandler::parse_manifest`), defaulting to `media_types::OCI_MANIFEST`.
5. `handle_delete_manifest`: allow deletion by digest with no `oci_tags` row — remove the `oci-manifests/<digest>` object and all `oci_tags` rows pointing at the digest. (This is required for correctness: once `GET` serves untagged objects, `DELETE` must remove the object, or a deleted manifest would resurrect on the next digest pull.)
6. Factor the shared `oci_tags → storage` resolution into one helper used by get/head/delete so behavior cannot drift.

`Remote`/`Virtual` repos keep proxying/caching unchanged (`resolve_virtual_manifest`, `cache_manifest_or_compute_digest`).

## Edge cases

- **Repo-type gating:** fallback only for `local`/`staging`; `Remote`/`Virtual` unchanged.
- **HEAD mirrors GET:** same digest / `Content-Type` / `Content-Length`, no body.
- **Storage GC interaction:** after the fix an untagged manifest is pullable by digest while the object lives; physical reclamation is still the storage GC's job. Note the manifest orphan path has **no grace window** (unlike blob GC's 24h `MIN_BLOB_AGE_SECS`), so an untagged-and-soft-deleted manifest is reclaimed on the next sweep — OCI-correct (untagged is GC-eligible) but worth tracking (see follow-ups). The orphan predicate itself is not changed here.
- **PUT atomicity (related, out of scope):** `handle_put_manifest` does `storage.put` + `INSERT oci_tags` + upsert `artifacts` outside one transaction, so a crash mid-PUT can leave a manifest object with no tag — exactly the untagged-by-digest state this fix makes retrievable. Wrapping PUT in a transaction is a separate hardening item.

## Acceptance criteria

1. `PUT` body **A** under tag `T`; then `PUT` body **B** under `T`; `GET`/`HEAD …/manifests/sha256:<A>` → **200** with correct `Docker-Content-Digest`, `Content-Type`, `Content-Length` (GET returns body **A**), while object **A** lives in storage.
2. `GET …/manifests/sha256:<nonexistent>` → **404 MANIFEST_UNKNOWN**.
3. `Remote`/`Virtual` behavior unchanged (proxy/cache regression tests green).
4. `DELETE` by digest removes the object and all tags pointing at it; a subsequent `GET` by digest → **404**.

## Tests

Unit/integration in `oci_v2`: (a) pull-by-digest with no `oci_tags` row on a hosted repo → 200; (b) HEAD mirrors GET; (c) content-type derived from the body `mediaType` (OCI image manifest, Docker v2, artifact with `artifactType`); (d) hosted repo without the object → 404; (e) `Remote`/`Virtual` do not regress; (f) `DELETE` untagged by digest. The new branching is extracted into pure / `&dyn StorageBackend` helpers so it can be covered without a live database (mirrors `verify_digest_or_fall_through` / `build_oci_proxy_response`).

## Related

- #1413 — extract the shared `oci-manifests/` prefix constant (the `manifest_storage_key` ↔ cascade ↔ GC-predicate coupling this fix also touches).
- #1607 / #1610 — Core Invariants Hardening epic / blob lifecycle & GC; this is a natural sibling.
- #1659 — blob GC under-reclaims `manifest_blob_refs` on untag/manifest-delete (the inverse-direction sibling of this read-availability bug).
- Follow-ups filed alongside this issue: a retention guard so lifecycle cannot orphan the sole tag protecting a manifest, and a first-class `oci_manifests` table to track manifest existence independent of tags (removes the root of this bug class).

inspired-geek:
## Summary

Fixes #1681.

For a hosted (`local`/`staging`) repository, `GET`/`HEAD`/`DELETE /v2/<name>/manifests/<digest>` were served only while a live `oci_tags` row still pointed at that `manifest_digest`. The moment the protecting tag was overwritten (`PUT`), deleted, or never existed (an index child / an untagged push), the registry returned `404 MANIFEST_UNKNOWN` even though the bytes were intact at `oci-manifests/<digest>`. This violates the OCI Distribution Spec (a manifest must be retrievable by digest until explicitly deleted) and breaks every digest pin — Tekton bundle resolver (`CouldntGetPipeline`), multi-arch index children, and referrers.

This PR adds a **content-addressable fallback** for digest references on repos that store their own manifests:

- When no `oci_tags` row resolves, `GET`/`HEAD` of a digest serve the object directly from storage after verifying `sha256(body) == digest`. The media type (absent from `oci_tags` for an untagged object) is sniffed from the manifest body's `mediaType`, defaulting to `application/vnd.oci.image.manifest.v1+json`.
- `DELETE` by digest now removes the `oci-manifests/<digest>` object (in addition to the tag rows). This is required for correctness: once `GET` serves untagged objects, a deleted manifest must not resurrect on the next digest pull.
- The `oci_tags → storage` resolution is factored into shared helpers (`lookup_manifest_tag_row`, `resolve_local_manifest_from_storage`, `build_local_manifest_response`, `remove_local_manifest_object`) so `GET`/`HEAD`/`DELETE` cannot drift; this also de-duplicates the previously-copied lookup between `GET` and `HEAD`.
- `Remote`/`Virtual` repositories keep proxying/federating **unchanged** — the fallback is gated to `Local`/`Staging`, and storage is only resolved when a local serve is actually possible.

Most new branching lives in pure / `&dyn StorageBackend` helpers, unit-tested without a database (mirrors the existing `verify_digest_or_fall_through` pattern). A DB-backed router test drives the full end-to-end scenario through the real OCI router.

Follow-ups filed alongside the root-cause issue (out of scope here): #1682 (retention can orphan the sole tag protecting a manifest) and #1683 (track manifest existence in a first-class `oci_manifests` table). Related: #1413, #1607, #1610, #1659.

`manifest_digest_db_tests::digest_pull_and_delete_survive_tag_overwrite` reproduces the bug end-to-end (push A under a tag, overwrite the tag with B, then pull/HEAD/delete A by digest). It returns `404` on `main` and passes here.

## Test Checklist
- [x] Manually tested locally

`cargo test --workspace --lib` (with Postgres): **10927 passed, 0 failed**. New code coverage on changed lines: **96.2%**. Duplication on the changed file: **0%** (jscpd). `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean.

## API Changes
- [x] N/A - no API changes

Behavior fix to existing OCI Distribution endpoints; no new endpoints, types, or migrations.

### DIFF
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -5342,6 +5342,247 @@ async fn handle_complete_upload(
 // Manifest handlers
 // ---------------------------------------------------------------------------
 
+/// True for hosted repositories that store their own manifests/blobs
+/// (`Local` or `Staging`); false for `Remote` (proxy) and `Virtual`
+/// (federated) repos, which resolve manifests by proxying/federating rather
+/// than from their own content-addressable storage.
+fn stores_own_manifests(repo_type: &str) -> bool {
+    repo_type == RepositoryType::Local || repo_type == RepositoryType::Staging
+}
+
+/// Whether a string is safe to emit verbatim as an HTTP header value. A
+/// manifest `mediaType` sniffed from the (attacker-controlled) body must pass
+/// this before it becomes the `Content-Type` header, or building the response
+/// would panic on an invalid header byte (e.g. an embedded newline).
+fn is_header_safe(value: &str) -> bool {
+    axum::http::HeaderValue::from_str(value).is_ok()
+}
+
+/// Sniff the `mediaType` of a manifest out of its JSON body, if present and
+/// usable as a header value. Returns `None` for non-JSON bodies, a
+/// missing/blank `mediaType`, or a value that is not a valid header.
+fn sniff_manifest_media_type(body: &[u8]) -> Option<String> {
+    crate::formats::oci::OciHandler::parse_manifest(body)
+        .ok()
+        .and_then(|manifest| manifest.media_type)
+        .map(|media_type| media_type.trim().to_string())
+        .filter(|media_type| !media_type.is_empty() && is_header_safe(media_type))
+}
+
+/// Decide the media type to serve for a manifest body. Prefers an explicit
+/// stored value (`oci_tags.manifest_content_type`), falls back to sniffing the
+/// body's `mediaType`, and finally to the OCI image-manifest default. An
+/// untagged manifest served content-addressably has no stored content type, so
+/// the body sniff is what gives it a correct media type.
+fn resolve_manifest_content_type(stored: Option<&str>, body: &[u8]) -> String {
+    if let Some(stored) = stored {
+        let stored = stored.trim();
+        if !stored.is_empty() && is_header_safe(stored) {
+            return stored.to_string();
+        }
+    }
+    sniff_manifest_media_type(body)
+        .unwrap_or_else(|| crate::formats::oci::media_types::OCI_MANIFEST.to_string())
+}
+
+/// Whether THIS repository has committed metadata referencing `digest` beyond a
+/// tag. This gates the content-addressable fallback (#1681).
+///
+/// On shared cloud backends the `oci-manifests/<digest>` object is NOT
+/// namespaced per repository — `StorageRegistry::backend_for` returns a single
+/// shared backend instance for s3/azure/gcs and ignores the per-repo path. So
+/// the mere existence of the object must not authorize a pull: otherwise a
+/// manifest pushed to repo A could be read through repo B on the same backend
+/// by digest. `manifest_blob_refs` (image manifests) and `oci_manifest_refs`
+/// parent rows (image indexes) are written per repository at push time and
+/// persist across tag overwrite/deletion, so they are the durable, repo-scoped
+/// proof that this repo actually holds the manifest. A child edge alone is not
+/// proof of ownership: an index body can reference a digest whose manifest body
+/// was never uploaded to this repo.
+async fn manifest_known_to_repo(
+    state: &SharedState,
+    repository_id: Uuid,
+    digest: &str,
+) -> Result<bool, sqlx::Error> {
+    sqlx::query_scalar::<_, bool>(
+        r#"
+        SELECT (
+          EXISTS (
+            SELECT 1
+            FROM manifest_blob_refs
+            WHERE repository_id = $1
+              AND manifest_digest = $2
+          )
+          OR EXISTS (
+            SELECT 1
+            FROM oci_manifest_refs
+            WHERE repository_id = $1
+              AND parent_digest = $2
+          )
+        )
+        "#,
+    )
+    .bind(repository_id)
+    .bind(digest)
+    .fetch_one(&state.db)
+    .await
+}
+
+/// Look up the `oci_tags` row for a manifest reference (digest or tag name)
+/// within a repository, returning `(manifest_digest, manifest_content_type)`.
+async fn lookup_manifest_tag_row(
+    state: &SharedState,
+    repo: &OciRepoInfo,
+    reference: &str,
+) -> Result<Option<(String, String)>, sqlx::Error> {
+    // Each `sqlx::query!` produces its own anonymous record type, so map to a
+    // tuple inside each branch before merging.
+    if is_digest_reference(reference) {
+        let row = sqlx::query!(
+            "SELECT manifest_digest, manifest_content_type FROM oci_tags WHERE repository_id = $1 AND manifest_digest = $2 LIMIT 1",
+            repo.id, reference
+        )
+        .fetch_optional(&state.db)
+        .await?;
+        Ok(row.map(|t| (t.manifest_digest, t.manifest_content_type)))
+    } else {
+        let row = sqlx::query!(
+            "SELECT manifest_digest, manifest_content_type FROM oci_tags WHERE repository_id = $1 AND name = $2 AND tag = $3",
+            repo.id, repo.image, reference
+        )
+        .fetch_optional(&state.db)
+        .await?;
+        Ok(row.map(|t| (t.manifest_digest, t.manifest_content_type)))
+    }
+}
+
+/// Resolve a manifest from this repository's own storage.
+///
+/// When a tag row resolved, the manifest is served from its digest-addressed
+/// object (any repo type, preserving existing tagged-pull behavior). When no
+/// tag row resolved, a digest reference on a hosted repo is served
+/// content-addressably from `oci-manifests/<digest>` ONLY when
+/// `repo_known_digest` is true — i.e. this repo has committed metadata for the
+/// digest (see [`manifest_known_to_repo`]). Without that gate a shared cloud
+/// backend would let one repo read another's manifest by digest (#1681 review).
+/// The stored bytes are verified to hash to the requested digest before being
+/// served. Returns `(digest, content_type, bytes)`, or `None` to fall through
+/// to the proxy paths.
+async fn resolve_local_manifest_from_storage(
+    storage: &dyn crate::storage::StorageBackend,
+    repo_type: &str,
+    reference: &str,
+    tag_row: Option<(String, String)>,
+    repo_known_digest: bool,
+) -> Option<(String, String, Bytes)> {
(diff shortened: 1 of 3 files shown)


brandonrc:
## Maintainer rebase + review (queued, not merged)

Rebased onto current `main` (was 12 commits behind, including #1711 OCI blob-journaling and #1598/#1724 repo-delete, both touching `oci_v2.rs`). The branch carried two merge commits whose conflict resolutions held the *final* form of the helpers; a naive `git rebase` would have replayed only the original commit and regressed them, so the rebase was done as a 3-way apply of the PR's net tree onto `main`. History is now a single clean commit; original authorship preserved.

Also removed three stale `.sqlx` cache files that matched the superseded first-commit macros (`manifest_known_to_repo`/`clear_repo_manifest_refs` are now runtime `sqlx::query`, not macros, so those caches were dead). The macros that remain (`lookup_manifest_tag_row`'s `oci_tags` lookups) reuse cache already on `main`.

### By-digest fix — correct
- GET/HEAD/DELETE by digest now resolve from a surviving `oci_tags` row **or**, for hosted (`local`/`staging`) repos, content-addressably from `oci-manifests/<digest>` when the repo has committed metadata for it. Tagged behavior is unchanged.
- The content-addressable path verifies the stored bytes hash to the requested digest before serving (corruption → fall through, not a bad 200).
- DELETE is now transactional (`oci_tags` + `oci_manifest_refs` + `manifest_blob_refs` in one tx with rollback); a cleanup failure aborts the delete instead of half-deleting the tag.

### Shared-backend safety (the important part)
On s3/azure/gcs the `oci-manifests/<digest>` object is **not** namespaced per repo. The fix gates the fallback on `manifest_known_to_repo` = `EXISTS manifest_blob_refs(repo,digest) OR EXISTS oci_manifest_refs(repo, parent_digest=digest)`. A bare index **child** edge is deliberately *not* proof of ownership (an index body can name a digest never uploaded here), so one repo cannot read another's manifest by digest. Legit multi-arch children remain pullable because a genuinely-pushed child has its own `manifest_blob_refs` rows. Covered by `untagged_digest_is_repo_scoped_on_shared_backend`.

### Blob-GC / `manifest_blob_refs` interaction — no wrong reclaim
`prune_orphan_blob_refs` deletes a manifest's blob refs once it is neither tagged nor a live child of a tagged index. At that same moment `manifest_known_to_repo` flips to false, so the by-digest fallback stops serving in lockstep with the blobs becoming reclaimable — the two predicates stay coherent (no window where the manifest serves by digest but its blobs were GC'd). Separately, `any_live_manifest_missing_refs` (the destructive-GC readiness gate) was decoupled from the backfill candidate set so live legacy child edges keep destructive GC gated even though they are intentionally not backfillable from shared storage.

### Issue coverage
- **#1681 — closed.** This is exactly the proposed interim content-addressable fallback.
- **#1682 — NOT addressed.** Lifecycle retention (`lifecycle_service.rs` / `CASCADE_OCI_TAGS_SQL`) is untouched; the sole-protecting-tag deletion path still needs a separate guard.
- **#1683 — NOT closed (by design).** This is the interim fix that the first-class `oci_manifests` table would later supersede.

### Gates (local)
- `cargo fmt --check` clean
- `cargo test --workspace --lib` → 11065 passed, 0 failed
- DB-gated tests (`manifest_digest_db_tests`) skip without `DATABASE_URL` locally; they run in CI's coverage job and are forced under coverage. jscpd: `oci_v2.rs` is exempt; the other two changed files add no duplication.

Queued for maintainer review — **not merged**, no `--admin`.

Issue #1646, opened 2026-06-05
CI: coverage gate diffs against main tip (not merge-base) → false failures on stale branches


brandonrc:
**Found while draining a batch of parallel PRs.** The new-code coverage gate produces FALSE failures on stale branches, causing repeated spurious red + forced rebases.

## Problem
The coverage gate (`.github/workflows/ci.yml`, "New code coverage gate") computes changed lines via:
```
git diff -U0 FETCH_HEAD -- '*.rs'
```
where `FETCH_HEAD` is the **tip of `main`** (`git fetch ... origin $base_ref`). When a PR branch is behind `main` (because another PR merged after it was cut), this diff includes the *other* merged PR's changes as a phantom delta — lines that look "added/removed" in this PR but aren't. Those phantom lines are counted as uncovered → the gate fails on code the PR never touched. The only "fix" is to rebase, which is pure churn (and burns a ~13-min CI run each time).

Observed repeatedly while merging PRs #1621/#1639/#1640/#1641 against a fast-moving `main`: each merge re-staled the others and flipped their coverage check red with zero real cause.

## Fix
Diff against the **merge-base**, not the raw base tip:
```
BASE=$(git merge-base FETCH_HEAD HEAD)
git diff -U0 "$BASE" HEAD -- '*.rs'
```
This counts only the lines THIS branch actually changed relative to where it diverged, eliminating phantom deltas from intervening merges. (The lcov cross-reference logic stays the same.)

## Done when
A branch that is behind `main` but whose own changes are fully covered passes the coverage gate without needing a rebase.

PR #1690, opened 2026-06-05
fix(ci): keep coverage gate alive when git merge-base fails on shallow PR head (#1689)

### LINKED ISSUE #1689: CI: coverage 'New code coverage gate' aborts on git merge-base under set -e (shallow PR head) — false failures (regression from #1671)
### Summary

Since #1671 merged (2026-06-05 ~15:58 UTC), the **New code coverage gate** step in the `coverage` job fails on many PRs **before it computes any coverage number**, producing false coverage failures. Real coverage is unaffected.

Currently failing at least: #1409, #1684, #1613 (others pass depending on how much history the shallow checkout happens to contain).

### Root cause

The step runs under GitHub Actions' default shell `bash -e` (errexit). #1671 added:

```sh
MERGE_BASE=$(git merge-base FETCH_HEAD HEAD)
if [ -z "$MERGE_BASE" ]; then
  echo "::warning::Could not compute merge-base; falling back to base tip"
  MERGE_BASE=FETCH_HEAD
fi
```

On a PR the head is checked out **shallow** (fetch-depth: 1). `git merge-base FETCH_HEAD HEAD` then exits **non-zero** because the common ancestor is not present in the shallow history. Under `set -e`, the failing command substitution **aborts the whole step**, so the `if [ -z "$MERGE_BASE" ]` fallback is dead code and never runs.

Evidence (job log of a failing run): the step dies ~4s after the `git fetch` lines, immediately after the deepen-fetch output and *before* any `Diffing new code against merge-base:` / `New code coverage:` output, with `##[error]Process completed with exit code 1`. No coverage percentage is ever printed.

Deepening only the base side (`git fetch --deepen=50 origin`) does not help, because the **PR head's** ancestry is still shallow, so `git merge-base` has nothing to walk on the head side.

### Impact

False "coverage below 70%" / failed-gate signals on PRs whose real new-line coverage is fine, blocking otherwise-green PRs. Not a code-coverage regression — purely the gate's git plumbing.

### Proposed fix

1. Make the merge-base resolution non-fatal so the existing fallback can run: `MERGE_BASE=$(git merge-base FETCH_HEAD HEAD 2>/dev/null || true)`.
2. Deepen **both** the base branch and the PR head enough that merge-base actually resolves in the common case (so we don't silently degrade to the base-tip fallback that #1646 warned about).
3. Resolve the fallback to a concrete SHA (`git rev-parse FETCH_HEAD`).

PR follows.

inspired-geek:
## Summary

Fixes a regression introduced by #1671 (merged ~4h ago): the **New code coverage gate** step now aborts on many PRs *before computing any coverage*, producing false coverage failures. Currently red on at least #1409, #1684, #1613.

**Root cause:** the step runs under GitHub's default `bash -e`. #1671 added `MERGE_BASE=$(git merge-base FETCH_HEAD HEAD)`. On a PR the head is checked out shallow (`fetch-depth: 1`), so `git merge-base` exits non-zero (no common ancestor in the shallow history); under `set -e` the failing command substitution aborts the whole step, making the `if [ -z "$MERGE_BASE" ]` fallback dead code. The job log of a failing run dies ~4s after the fetches, before any `Diffing new code against merge-base:` output.

**Fix:**
1. `MERGE_BASE=$(git merge-base FETCH_HEAD HEAD 2>/dev/null || true)` — a merge-base failure now degrades gracefully to the existing base-tip fallback instead of crashing the step.
2. Deepen **both** the base branch and the PR head (50 → 200) so merge-base actually resolves in the common case (avoiding the base-tip fallback that reintroduces the #1646 over-counting).
3. Resolve the fallback to a concrete SHA (`git rev-parse FETCH_HEAD`).

This is the gate-fix-first step for #1409 (real coverage 94%, blocked only by this false failure); other affected PRs benefit too. Closes #1689.

- [x] N/A — CI workflow change; not covered by the Rust/integration/e2e test tiers. Validation is empirical: the coverage gate now runs to completion on shallow PR checkouts (this PR's own coverage job, and the previously-red #1409/#1684/#1613, go green) instead of aborting at `git merge-base`.

## Test Checklist
- [x] Manually tested locally — reproduced the `git merge-base` non-zero-exit + `set -e` abort; confirmed `2>/dev/null || true` lets the fallback run.

## API Changes
- [x] N/A - no API changes (CI workflow only)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -221,10 +221,12 @@ jobs:
         if: github.event_name == 'pull_request' && steps.detect.outputs.rust_changed == 'true'
         run: |
           # Fetch base branch so we can diff against it (checkout is shallow by default).
-          # Fetch with depth (not --depth=1) and deepen the PR head so that
-          # `git merge-base` can resolve a common ancestor in this shallow checkout.
-          git fetch --no-tags --depth=50 origin ${{ github.base_ref }}
-          git fetch --no-tags --deepen=50 origin || true
+          # Deepen BOTH the base branch AND the PR head: the head is checked out
+          # shallow (fetch-depth: 1), so without head-side history `git merge-base`
+          # has nothing to walk on the head side and finds no common ancestor.
+          # Deepening only the base (the original #1671 behavior) is not enough.
+          git fetch --no-tags --depth=200 origin ${{ github.base_ref }}
+          git fetch --no-tags --deepen=200 origin || true
 
           # Diff against the MERGE-BASE between the PR head and the base branch tip,
           # NOT the raw base tip (FETCH_HEAD). Diffing against the tip counts changes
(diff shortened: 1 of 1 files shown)


brandonrc (commented):
Verified end-to-end — this is a real gate regression, not a coverage problem:
- main `ci.yml:235` has the unguarded `MERGE_BASE=$(git merge-base FETCH_HEAD HEAD)`; GitHub's default `bash -eo pipefail` aborts the step on a non-zero command substitution, so the `if [ -z "$MERGE_BASE" ]` fallback at :236–238 is dead code.
- Confirmed in the failing #1684 coverage run: the step prints the `git fetch --deepen` output then `##[error]Process completed with exit code 1` with **no** `Diffing new code against merge-base:` and **no** coverage % — it dies right at the merge-base line.
- It's currently failing #1691/#1692 (this epic's closeout PRs) too, so the gate is broadly broken for stale/shallow heads.
Fix is correct (guard the substitution so the fallback runs, deepen both sides, `rev-parse FETCH_HEAD`). Arming to land first; will rebase the blocked PRs onto main after.

PR #1747, opened 2026-06-08
fix(proxy): page cached-artifact listing without O(N) sidecar loads + guard per_page=0 (#1571)

### LINKED ISSUE #1571: ProxyService::list_cached_artifacts loads every sidecar before paginating (O(N) per listing)
Follow-up from the security review of #1567 (the #424 fix, shipped in v1.2.0).

`ProxyService::list_cached_artifacts` lists the entire `proxy-cache/<repo_key>/` prefix and loads every sidecar into memory before `filter_and_paginate_cached` slices the requested page. So every remote-repo artifact listing is O(N) storage reads regardless of page size, which does not scale for large proxy caches.

Also noted: a `per_page == 0` path should be guarded against divide-by-zero in the pagination math.

Suggested: scan/stream and stop at the page boundary (or push the prefix+pagination down to the storage backend's list), and add a `per_page` lower-bound guard. Not a correctness or security issue (review confirmed the path is safe); purely scale/perf.

Severity: Medium (perf). Surfaced by the v1.2.0 security review.

brandonrc:
## Summary

The remote-repo **cached-artifact listing** had two defects, both surfaced in #1571:

1. **O(N) sidecar reads per listing.** `ProxyService::list_cached_artifacts` listed the entire `proxy-cache/<repo_key>/` prefix and loaded *every* sidecar into memory before the handler sliced the requested page. A `page=1&per_page=20` request against a 10k-object cache still did ~10k storage reads.
2. **`per_page=0` divide-by-zero.** A `per_page=0` query slipped past the `.min(100)` upper bound and reached `total_pages = ((total as f64) / (per_page as f64)).ceil() as u32`, where `inf as u32` saturates — returning `total_pages = 4294967295` for any non-empty repo (affecting both the DB and cached branches).

**Fix:**
- Two-phase listing: `list_cached_paths` recovers the path set with **zero** sidecar reads; the handler filters + paginates the path strings (both listing filters — `path_prefix` and `q` — are purely path-based); `load_cached_entries` then reads sidecars for **only that page**. Listing is now O(page) reads. `list_cached_artifacts` remains as a thin wrapper for existing callers/tests.
- Clamp `per_page` to `[1, 100]` at the handler and add `cached_total_pages`, a guarded integer ceil-division helper, so neither branch can divide by zero.

Trade-off (documented in code): `total` now counts paths whose sidecar may since have gone missing, instead of paying an O(N) read to pre-exclude them. Such a path is still dropped from the returned page (matching old behavior) — an acceptable approximation in exchange for O(page) reads.

New tests (all pure / no DB):
- `test_cached_total_pages_guards_per_page_zero` — `cached_total_pages(5, 0) == 5` (was `u32::MAX`).
- `test_filter_and_paginate_paths_per_page_zero_does_not_wedge` — `per_page=0` no longer produces a degenerate page.
- `test_list_cached_paths_reads_no_sidecars` — path recovery does 0 sidecar reads (counting storage mock).
- `test_load_cached_entries_reads_only_requested_paths` — a 1-path page reads exactly 1 sidecar out of a 3-object cache (the O(N) regression).
- Migrated the 4 `filter_and_paginate_cached` tests to the path-based `filter_and_paginate_paths`.

## Test Checklist
- [x] Manually tested locally (`cargo test --lib`, `cargo clippy -- -D warnings`, `cargo fmt --check` all pass)

## API Changes
- [x] N/A - no API changes (response shape unchanged; only pagination math + internal read strategy)

Closes #1571

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -1892,7 +1892,12 @@ pub async fn list_artifacts(
     Query(query): Query<ListArtifactsQuery>,
 ) -> Result<Json<ArtifactListResponse>> {
     let page = query.page.unwrap_or(1).max(1);
-    let per_page = query.per_page.unwrap_or(20).min(100);
+    // Clamp per_page to [1, 100]: a `per_page=0` query previously slipped past
+    // the `.min(100)` upper bound and reached the `total_pages` divisions
+    // below as a divide-by-zero, saturating `total_pages` to u32::MAX for any
+    // non-empty repo (#1571). The lower bound makes both the DB and the
+    // proxy-cache listing branches well-defined.
+    let per_page = query.per_page.unwrap_or(20).clamp(1, 100);
     let offset = ((page - 1) * per_page) as i64;
 
     let repo_service = RepositoryService::new(state.db.clone());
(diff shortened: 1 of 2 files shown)


brandonrc:
Note: the ❌ **Code Coverage** check is red due to a pre-existing systemic failure on `main` (instrumented build exits 100, produces no `lcov.info`) — tracked in #1753, not a regression from this PR. All branch-protection-required checks (Backend Unit Tests, Check Rust, CI Complete) are green, and the changed logic ships with passing unit tests. Merging on that basis.

Issue #1607, opened 2026-06-04
Epic: Core Invariants Hardening (streaming, cross-replica coordination, blob lifecycle, cache correctness)


brandonrc:
## Epic: Core Invariants Hardening

**Source:** Design retrospective + honest issue analysis (`docs/audits/design-retro-2026-06-04.md`, `docs/audits/honest-issue-analysis-2026-06-04.md`).

A clean-room architecture exercise reproduced ~85% of Artifact Keeper's design — the architecture is sound and **no rewrite is warranted**. This epic hardens three cross-cutting *disciplines* that were retrofitted rather than designed in, by making them **enforced invariants**.

### Scope correction (read this)
The honest, fix-verified data shows these invariants are **~10% of defect volume (36 of 356)** — NOT the bulk of historical bugs. (The bulk is auth-logic, native-client format-compat, and ordinary error-handling — see sibling epic #1615.) **This epic is justified by RISK, not volume:** the invariants are over-represented in the *severe and still-open* tail — they own the open data-loss leaks, the pod-evicting OOMs, and the cross-replica corruption.

### Sub-issues (ordered by severity of the open bugs they own)
- [ ] **#1618 — Foundational refactor** of `ProxyService`/`proxy_helpers.rs` (from community discussion #1614). The invariants below live in this code; landing them is far cheaper once #1618 establishes clean seams. Do first or alongside ①/②/④.
- [ ] #1608 — **① Streaming invariant** *(P0 — owns the open `outage-oom`: #1573 multi-GiB → pod eviction)*. No artifact-path handler buffers a whole body; convert GCS backend, scanner, `/tmp` staging, OCI ceiling.
- [ ] #1610 — **③ Blob lifecycle & GC** *(P0 — owns the open `data-loss`: #1408 ~403 GB leaked, #1551, #1527)*. Ref-counted mark-and-sweep with grace window + single-leader; S3 delete reclaims.
- [ ] #1609 — **② Cross-replica single-flight** *(P1 — owns open `data-integrity` #1606)*. Postgres advisory-lock single-flight; losers poll + large-artifact fallback.
- [ ] #1611 — **④ Cache correctness** *(P1 — virtual/proxy `wrong-result` family: #1600, #1595, #1562)*. Per-path-pattern immutable-vs-mutable classification + conditional revalidation.

### Non-goals
- No rewrite, no re-architecture. The format-handler / CAS / Postgres-source-of-truth / WASM-plugin design stays.

[...]

PR #1749, opened 2026-06-08
fix(startup): run manifest_blob_refs backfill in the background (#1642)

### LINKED ISSUE #1642: Perf: manifest_blob_refs startup backfill scans every manifest serially before HTTP bind (follow-up to #1635)
**Follow-up to #1635 / PR #1641 (from code review).** Non-blocking, but fix before OCI blob GC (#1408) depends on these edges at scale.

## Problem
The `manifest_blob_refs` startup backfill (`manifest_blob_refs_backfill.rs`) scans **every image manifest** in the registry (union of all non-index `oci_tags` + all `oci_manifest_refs.child_digest`), serially — one storage GET + parse + INSERT per manifest — and it runs **before the HTTP listener binds** (`main.rs`). On a large registry's FIRST post-upgrade boot this can meaningfully delay startup. (The sibling `oci_manifest_refs` backfill only scanned the much smaller index-manifest subset, so it didn't have this cost.)

It is bounded, one-shot (subsequent boots are zero-row no-ops via the `NOT EXISTS` anti-join + `ON CONFLICT DO NOTHING`), and non-fatal — so not a merge-blocker for #1641. But it should be hardened before GC relies on full coverage.

## Options
- Move the first-pass backfill to a **background task** (spawn after HTTP bind) so startup isn't blocked.
- And/or **batch/chunk** the candidate processing and/or parallelize the per-manifest storage reads with a bounded concurrency.
- Optionally a progress/throughput log so operators can see it draining.

## Done when
A large registry's first boot after upgrade is not blocked by the backfill, and edge coverage still converges to complete.

brandonrc:
## Summary

On a large post-upgrade corpus the `manifest_blob_refs` backfill scans **every** image manifest serially (storage GET + JSON parse + DB INSERT per manifest), and it was `.await`ed at `main.rs:499` **before** the HTTP listener binds (`main.rs` ~930). First boot after the migration-120 upgrade could therefore delay readiness by minutes.

This moves the backfill into a `tokio::spawn` so the HTTP listener binds immediately. Deferring it is safe by construction:

- The backfill is **additive-only** (no deletion).
- The blob-GC readiness gate (`any_live_manifest_missing_refs`) keeps blob GC **OFF** until every live manifest has its refs — so GC cannot act on a half-backfilled corpus.
- Failures continue to be logged and escalated per-tick by the scheduler (#1409).

Part of the Core Invariants epic (#1607).

- [x] N/A — this is a **startup-sequencing** change in `main.rs` (binary entrypoint, not unit-testable). The backfill's *correctness* is unchanged and remains covered by the existing `manifest_blob_refs_backfill` tests; the *safety* of deferring is guaranteed by the pre-existing blob-GC readiness gate (unchanged). Verifiable by observing sub-second time-to-first-HTTP-response on a corpus with pending backfill.

## Test Checklist
- [x] No regressions in existing tests (`cargo check`, `cargo clippy -- -D warnings`, `cargo fmt --check` all pass)

## API Changes
- [x] N/A - no API changes

Closes #1642

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -496,32 +496,46 @@ pub async fn run_server(shutdown_token: Option<CancellationToken>) -> Result<()>
     // backends. Failures are logged but do not block startup; on a fresh
     // database or after the first successful run the candidate query
     // returns zero rows and this is a near-instant no-op.
-    let blob_refs_backfill_stats =
-        artifact_keeper_backend::services::manifest_blob_refs_backfill::run_backfill(
-            &db_pool,
-            storage_registry.clone(),
-        )
-        .await;
-    // A failed candidate (body missing from storage, over-cap, DB write
-    // error) leaves that manifest ref-less, which keeps the blob-GC readiness
-    // gate closed — the feature stays OFF. This startup log is the earliest
-    // signal; the scheduler escalates per-tick thereafter (#1409).
-    if blob_refs_backfill_stats.candidates_failed > 0 {
-        tracing::error!(
-            candidates_scanned = blob_refs_backfill_stats.candidates_scanned,
-            edges_inserted = blob_refs_backfill_stats.edges_inserted,
-            candidates_failed = blob_refs_backfill_stats.candidates_failed,
-            "manifest_blob_refs backfill left {} live manifest(s) un-backfilled; \
-             blob GC will stay gated off until they are resolved (re-pushed, or \
-             the offending tag deleted)",
-            blob_refs_backfill_stats.candidates_failed
-        );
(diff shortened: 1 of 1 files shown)


brandonrc:
Note: the ❌ **Code Coverage** check is red due to a pre-existing systemic failure on `main` (instrumented build exits 100, produces no `lcov.info`) — tracked in #1753, not a regression from this PR. All branch-protection-required checks (Backend Unit Tests, Check Rust, CI Complete) are green, and the changed logic ships with passing unit tests. Merging on that basis.

Issue #1631, opened 2026-06-04
Coordination seam: SingleFlight (buffered + streaming fan-out + cross-replica advisory-lock) — foundation for #1608/#1609


brandonrc:
**Foundation for #1608 (streaming) and #1609 (cross-replica single-flight).** Net-new coordination architecture — NOT part of the #1618 de-duplication refactor (decision recorded on #1618).

## Why
The adversarial review of #1618 (`docs/audits/proxy-refactor-plan-1618.md`, Amendment 4) found:
- The **streaming path is uncoordinated today** — `fetch_artifact_streaming` goes straight to `fetch_from_upstream_streaming → tee_upstream_to_cache` with no single-flight; N concurrent cold-cache streamers each open their own upstream connection and race to write the same `__content__`.
- `proxy_hydration.rs` **cannot host stream fan-out**: its contract is "leader produces one owned `T`; followers re-check the cache" — but a streaming follower can't re-check mid-flight (the body isn't in cache until the tee completes). Works for buffered (`T = Bytes`), not streaming.
- There is **zero advisory-lock machinery in the proxy path** (only `main.rs` admin init uses `pg_advisory_xact_lock`), so #1609's cross-replica single-flight has no hook.

So #1611 lands on the #1618 cache seam, but **#1608/#1609 need this seam first.**

## Scope — a `SingleFlight`/`Coordinator` seam with three layers
1. **Buffered re-check impl** — relocate the existing `coordinate_proxy_hydration` behavior behind the trait (leader produces `Bytes`; followers re-check cache). No behavior change.
2. **Streaming broadcast-fan-out impl** — leader streams upstream once; followers subscribe to the leader's chunks (e.g. `tokio::sync::broadcast`) instead of each opening upstream. Requires a new primitive, not an extension of the buffered one.
3. **Cross-replica advisory-lock decorator** — `pg_try_advisory_xact_lock(hash(repo_id‖path))` with the loser-poll + large-artifact passthrough fallback. Full design already written in `docs/audits/design-clean-room-singleflight-cache-2026-06-04.md`.

## Dependencies
- Lands on the seams from **#1618** — its S9 persist / upstream seam must leave a clean injection point (recorded on #1618).
- **Acceptance gate:** makes the multi-replica harness **#1627** go green (currently reproduces #1606 red).

## Done when
[...]

Issue #1609, opened 2026-06-04
Core Invariant ②: cross-replica single-flight for pull-through cache (Postgres advisory lock)


brandonrc:
Part of #1607 (Core Invariants Hardening). Source: `docs/audits/design-retro-2026-06-04.md`.

## Invariant
Pull-through cache fills are coordinated **cluster-wide**, not per-process. Use a PostgreSQL advisory lock keyed on `hash(repo_id‖path)`. Winner streams upstream → tees to client + multipart store, then commits metadata (**the linearization point**). Losers poll for the blob/metadata row with bounded backoff; large artifacts fall back to proxy-without-cache after a timeout. Content-addressing makes a duplicate fetch safe.

## Why
Current single-flight (#1355) is per-process, so across replicas concurrent cold-cache requests race and serve partial / truncated files. Advisory locks auto-release on connection death (crash-safe).

## Work
- [ ] Replace per-process single-flight with `pg_try_advisory_xact_lock`-based coordination.
- [ ] Implement loser poll + backoff and large-artifact proxy-without-cache fallback.
- [ ] Negative-cache upstream 404s with short TTL.
- [ ] Multi-replica concurrency test proving no partial/truncated serves.

## Linked open issues
#1606

brandonrc:
**Prerequisite:** #1631 (coordination seam) — this issue lands on that seam, not directly on #1618. The multi-replica harness #1627 is the acceptance gate.

PR #1752, opened 2026-06-08
docs(plan): #1609 cross-replica single-flight implementation plan

brandonrc:
## Summary

A build-ready implementation plan for **#1609** (Core Invariant ② — cross-replica single-flight), the keystone remaining piece of the #1607 / #1631 seam. Written instead of blind-coding because the feature cannot be behaviourally verified without Postgres + the multi-process concurrency harness.

Grounded in **both** the clean-room design doc (`docs/audits/design-clean-room-singleflight-cache-2026-06-04.md`) and the actual current code. Highlights:

- **Ground truth:** Layers 1 (buffered `Coordinator`) and 2 (streaming fan-out) are already merged and wired; only the Layer-3 advisory-lock decorator is missing (`// #1631 layer 3 seam` comments).
- **Critical correction:** use a `CoordinatorImpl` **enum** — NOT a generic `ProxyService<C>`. Generic would force `SharedState` and ~40 handler signatures generic. The enum has zero blast radius outside `proxy_hydration.rs` / `proxy_service.rs` / `main.rs`.
- **Why naive is wrong:** holding `pg_try_advisory_xact_lock` across the upstream download exhausts the connection pool → the two-transaction + `cache_fill` lease pattern (claim → heartbeat → finalize → reclaim → janitor).
- **AK deltas** vs the generic design (UUID repo ids; existing proxy-cache content-addressing already covers the design's `blob`/multipart layer; negative cache already implemented at `proxy_service.rs:2063`).
- **5-PR breakdown** + a **test matrix** gated on the existing `scripts/concurrency/run-singleflight-race.sh` harness (#1627), which is intentionally red on `main` and flips green when this lands.
- **Open decisions** flagged for maintainer sign-off (lease constants, passthrough budget, `siphasher` dep, failure ceiling).

- [x] N/A — docs-only (planning artifact). No code change.

## Test Checklist
- [x] N/A — documentation only

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/docs/plans/2026-06-08-1609-cross-replica-singleflight-impl-plan.md
+++ b/docs/plans/2026-06-08-1609-cross-replica-singleflight-impl-plan.md
@@ -1,0 +1,159 @@
+# Implementation Plan — #1609 Cross-Replica Single-Flight (Layer 3)
+
+**Issue:** #1609 (Core Invariant ②) · **Epic:** #1607 · **Seam:** #1631 · **Also closes:** #1606
+**Design source:** `docs/audits/design-clean-room-singleflight-cache-2026-06-04.md` (Subsystem 1)
+**Status:** ready to execute in a Postgres-equipped dev loop. Cannot be behaviourally verified without PG + the multi-process harness, so it is intentionally **not** built blind.
+
+---
+
+## 0. Where we are (ground truth, verified against source)
+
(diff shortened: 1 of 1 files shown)


brandonrc:
We are not going to publish plan documents

PR #1787, opened 2026-06-09
fix(npm): resolve npm QA bugs (#1774)

### LINKED ISSUE #1774: QA sweep: npm — 1 confirmed API bug(s) (live-verified on :dev)
Found by the live QA fan-out against the current-main `:dev` container and **verified real on current source** (HEAD 43b0e98f). 1 confirmed bug(s) in the `npm` subsystem.

### 1. [MEDIUM] Virtual repo npm tarballs served with Content-Type: application/octet-stream instead of application/gzip
**Repro/live evidence:** curl -v http://localhost:8080/npm/vf-npm-virtual-02252e31/is-array/-/is-array-1.0.1.tgz returned 'content-type: application/octet-stream'. Same tarball via direct remote repo vf-npm-remote-fc1f4682 returned 'content-type: application/gzip'. Reproduced on two consecutive requests (both warm-cache hits), confirming it is not a first-fetch fluke.
**Source:** `backend/src/api/handlers/npm.rs:1229-1234`
**Fix (trivial):** On line 1232, replace `result.content_type,` with `None,` (or `Some(NPM_TARBALL_CONTENT_TYPE.to_string()),`) so that build_tarball_response_stream always falls back to application/gzip for npm tarballs regardless of what content_type the proxy cache sidecar recorded. Optionally also call correct_cached_tarball_content_type for the virtual path (after resolve_virtual_download) to fix the DB record for future SBOM consumers.

brandonrc:
## Summary

Closes #1774

Fixes the single REAL_ON_CURRENT npm QA bug:

**Virtual repo npm tarballs served with `Content-Type: application/octet-stream` instead of `application/gzip` (medium).**
When a Virtual npm repo satisfies a tarball download from a member repo's proxy cache, the handler passed `result.content_type` (whatever the proxy-cache sidecar recorded from upstream — frequently `application/octet-stream`, sometimes a stale error-page type) straight through to the response. The direct remote-repo and local paths already normalize to `application/gzip`, so the virtual path leaked the wrong type, breaking downstream SBOM/scanner content sniffing (Trivy, Grype).

Fix: introduced `npm_virtual_tarball_content_type()` which always returns `application/gzip` (`NPM_TARBALL_CONTENT_TYPE`), and used it at the virtual-download call site in `download_tarball` (backend/src/api/handlers/npm.rs). Reproduced exactly: `curl -v .../is-array-1.0.1.tgz` previously returned `application/octet-stream` for the virtual repo while the direct remote repo returned `application/gzip`.

Added two unit tests in `api::handlers::npm::tests`:
- `test_npm_virtual_tarball_content_type_always_gzip` — verifies the helper returns gzip for `None`, `application/octet-stream`, `text/html`, and `application/gzip` cached values.
- `test_virtual_tarball_response_overrides_octet_stream` — end-to-end of helper + `build_tarball_response_stream`, asserting an octet-stream cache record is served as `application/gzip`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Verified locally: `cargo fmt --check`, `SQLX_OFFLINE=true cargo clippy --workspace --all-targets -- -D warnings` (lib clean), and `cargo test -p artifact-keeper-backend --lib api::handlers::npm` (88 passed, 0 failed) against a fresh DB. No `sqlx` macros changed, so the offline cache is untouched.

### DIFF
--- a/backend/src/api/handlers/npm.rs
+++ b/backend/src/api/handlers/npm.rs
@@ -987,6 +987,20 @@ fn build_tarball_response_stream(
     builder.body(Body::from_stream(stream)).unwrap()
 }
 
+/// Decide the `Content-Type` for an npm tarball served from a Virtual repo.
+///
+/// Virtual downloads are satisfied by a member repo's proxy-cache, and the
+/// sidecar records whatever content type the upstream sent — for npm that is
+/// frequently `application/octet-stream` (and occasionally a bogus
+/// text/HTML type on error pages that slipped into cache). npm tarballs are
+/// always gzip, and downstream SBOM/scanner tooling (Trivy, Grype) keys off
+/// the content type, so the virtual path must normalize to `application/gzip`
+/// just like the direct remote-repo path does. We therefore ignore the cached
+/// content type entirely and always return [`NPM_TARBALL_CONTENT_TYPE`].
+fn npm_virtual_tarball_content_type(_cached: Option<String>) -> Option<String> {
+    Some(NPM_TARBALL_CONTENT_TYPE.to_string())
+}
+
 /// Build an OK response with a given content type and body.
 fn build_ok_response(content_type: &str, body: impl Into<Body>) -> Response {
     Response::builder()
@@ -1226,10 +1240,13 @@ async fn serve_tarball(
         )
         .await?;
 
+        // Always serve npm virtual-repo tarballs as `application/gzip`,
+        // overriding whatever content type the proxy-cache sidecar recorded
+        // for the cached member artifact (see `npm_virtual_tarball_content_type`).
         return Ok(build_tarball_response_stream(
             result.body,
             filename,
-            result.content_type,
+            npm_virtual_tarball_content_type(result.content_type),
             result.content_length,
         ));
     }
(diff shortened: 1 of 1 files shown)


brandonrc:
Adversarial review — **APPROVE** (with two non-blocking suggestions).

## Correctness ✅
Traced the fix against all three tarball-serving call sites in `serve_tarball`:
- remote path (npm.rs:1171) → passes `None` → falls back to `application/gzip`
- local path (npm.rs:1294) → passes `None` → `application/gzip`
- **virtual path (npm.rs:1232)** → was passing `result.content_type` straight through, leaking the proxy-cache `application/octet-stream`. This was the only leaky site.

`npm_virtual_tarball_content_type()` always returns `NPM_TARBALL_CONTENT_TYPE`, so the virtual path now matches the other two. The bug is genuinely fixed, and the behavior is consistent with how the rest of the handler normalizes npm tarball MIME types.

## Scope ✅
Of the three npm QA entries, only the content-type bug was `REAL_ON_CURRENT`. The dist-tags routes (router lines 59–64) and the prerelease `latest` re-derivation (npm.rs:495–506) are already present on main and were marked `FIXED` in QA. The PR correctly addresses only the live bug. Single file, +61/-1, no unrelated edits.

## CI ✅
Check Rust pass, Backend Unit Tests pass, Code Coverage pending. Required checks green.

## 🟡 Suggestions (non-blocking)
1. **Regression tests guard the helper, not the wiring.** Both new tests exercise `npm_virtual_tarball_content_type` directly. Reverting line 1232 back to `result.content_type` would NOT fail either test, since the helper still exists and works. This file already has a `tdh` DB-fixture harness driving the handler end-to-end for the #1377 scoped path — an analogous end-to-end assertion on the virtual `serve_tarball` branch would be a strictly stronger regression guard.
2. **Cached DB record left stale.** The QA fix-sketch also suggested calling `correct_cached_tarball_content_type` on the virtual path. The PR fixes the served header but leaves the member repo's cached `content_type` as `octet-stream` when a tarball is fetched *only* via the virtual repo (the remote path self-heals it, the virtual path does not). The HTTP response is correct now, so this is an edge case, not a blocker — worth a follow-up if SBOM tooling ever reads the DB record rather than the response header.

Solid, minimal, correct fix. Good helper extraction and clear doc comment explaining the rationale.

Issue #1371, opened 2026-05-24
[security] Deactivated user's API token still accepted after 30s (cache not invalidated)


brandonrc:
**Security regression.** Surfaced by release-gate auth-tests (run 26344757642).

### Failing assertion
`auth-user-deactivation-revokes-tokens / Deactivated user's API token is rejected`: token still accepted (HTTP 200) after 30s; cache should have been flushed on `is_active=false`.

### Why this matters
Setting `is_active = false` on a user is the primary off-boarding lever. If their API tokens keep authenticating for any duration after deactivation, that's a CVE-class hardening miss.

### Hypothesis
The token validation path likely caches the user `is_active` flag (or the token row) in-memory without an invalidation hook on `users.is_active` change. Find the cache (LRU/Moka/etc.) and add an `invalidate_user(user_id)` call from the user-deactivate handler.

### Acceptance
- Token rejected within 5s of deactivation (not 30s).
- Regression test asserts the cache invalidation path.
- Audit log entry for token rejection includes deactivation reason.

Part of #1356.

PR #1788, opened 2026-06-09
fix(cargo): resolve cargo QA bugs (#1777)

### LINKED ISSUE #1777: QA sweep: cargo — 2 confirmed API bug(s) (live-verified on :dev)
Found by the live QA fan-out against the current-main `:dev` container and **verified real on current source** (HEAD 43b0e98f). 2 confirmed bug(s) in the `cargo` subsystem.

### 1. [CRITICAL] Native cargo client authentication fails: server rejects plain Authorization token sent by cargo:token provider
**Repro/live evidence:** 1) curl -s -w '\nHTTP:%{http_code}' http://localhost:8080/cargo/vf-cargo-4015d4a9/config.json -H "Authorization: $TOKEN" => HTTP:401 (raw token, no scheme prefix rejected). 2) curl -s -I http://localhost:8080/cargo/vf-cargo-4015d4a9/config.json returns: www-authenticate: Basic realm="artifact-keeper" and www-authenticate: Bearer realm="artifact-keeper", charset="UTF-8" — no 'Cargo' challenge. Bearer-prefixed token returns HTTP:200.
**Source:** `backend/src/api/middleware/auth.rs:323 (extract_token_from_auth_header only handles Bearer/ApiKey/Basic; returns ExtractedToken::Invalid for plain tokens) and backend/src/api/middleware/auth.rs:1081 (`
**Fix (small):** Two changes needed: (1) In extract_token_from_auth_header (auth.rs:323), add a final else branch that treats the entire auth_header value as a Bearer token when it contains no space (i.e., looks like a raw JWT/token): `else if !auth_header.contains(' ') { ExtractedToken::Bearer(auth_header) }`. (2) In unauthorized_response (auth.rs:1081) and on all cargo routes that return 401, add a third header: `WWW-Authenticate: Cargo`. This signals to cargo 1.67+ to use the Cargo token protocol.

### 2. [MEDIUM] Search endpoint meta.total returns page count, not total match count — breaks cargo search pagination
**Repro/live evidence:** After uploading 3 crates (alpha-crate, beta-crate, gamma-crate) to repo vf-cargo-4015d4a9: GET /cargo/vf-cargo-4015d4a9/api/v1/crates?q=&per_page=1 returns {meta_total: 1, crates_returned: 1} (wrong — should be 3). GET /cargo/vf-cargo-4015d4a9/api/v1/crates?q=&per_page=100 returns {meta_total: 3, crates_returned: 3} (correct only because all crates fit in one page).
**Source:** `backend/src/api/handlers/cargo.rs:446 — `"total": crate_list.len()` uses the LIMIT-truncated result slice length; the SQL query at line 399 applies LIMIT $3 before count, and no separate COUNT(*) quer`
**Fix (small):** Before the paginated query, run a separate COUNT(*) query: `sqlx::query_scalar!("SELECT COUNT(DISTINCT a.name) FROM artifacts a WHERE a.repository_id=$1 AND a.is_deleted=false AND ($2='' OR a.name ILIKE '%'||$2||'%')", repo.id, query).fetch_one(&state.db).await` and use this value as `total` in the JSON response instead of `crate_list.len()`.

brandonrc:
## Summary

Closes #1777

Fixes the two live-verified cargo QA bugs from the QA fan-out sweep.

**1. [CRITICAL] Native cargo client auth rejected (raw token + missing Cargo challenge)**
The native cargo client's `cargo:token` credential provider sends the raw token as the `Authorization` header value with **no** scheme prefix, but `extract_token_from_auth_header` only handled `Bearer`/`ApiKey`/`Basic` and returned `Invalid` for a scheme-less token, causing a 401. Additionally, `unauthorized_response` only emitted `Basic` and `Bearer` `WWW-Authenticate` challenges, never `Cargo`, so cargo 1.67+ never engaged the token protocol.
- `auth.rs`: a non-empty, single-word (no space) Authorization value is now treated as a `Bearer` token.
- `auth.rs`: `unauthorized_response` now also emits `WWW-Authenticate: Cargo`.

**2. [MEDIUM] Search `meta.total` returned page count, not total match count**
`search_crates` set `meta.total` to `crate_list.len()` — the LIMIT-truncated page slice — which breaks cargo's search pagination. It now runs a separate `COUNT(DISTINCT a.name)` query and reports that as `meta.total`. Response construction was extracted into the pure helper `build_search_response` for testability.

New tests:
- `auth::tests::test_extract_plain_token_treated_as_bearer` and `test_extract_plain_token_with_space_is_invalid` — cover the raw-token acceptance.
- `auth::tests::test_unauthorized_response_has_www_authenticate_headers` extended to assert the `Cargo` challenge.
- `cargo::tests::test_search_meta_total_reflects_total_not_page_len` and `test_search_meta_total_zero_when_no_matches` — assert `meta.total` is the total, not the page length.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes (no new endpoints; behavior fixes only)

Verified locally against a fresh DB: `cargo fmt --check`, `SQLX_OFFLINE=true cargo clippy --workspace --all-targets -- -D warnings` (0 warnings), and the cargo + auth module tests (312 passed, 0 failed). `.sqlx` offline cache regenerated for the new COUNT query.

### DIFF
--- a/.sqlx/query-aaedcabcafb76d5c573603c4be2e6e177560d993714f07bc9739754de2eb2182.json
+++ b/.sqlx/query-aaedcabcafb76d5c573603c4be2e6e177560d993714f07bc9739754de2eb2182.json
@@ -1,0 +1,23 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT COUNT(DISTINCT a.name)\n        FROM artifacts a\n        WHERE a.repository_id = $1\n          AND a.is_deleted = false\n          AND ($2 = '' OR a.name ILIKE '%' || $2 || '%')\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "count",
+        "type_info": "Int8"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid",
+        "Text"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "aaedcabcafb76d5c573603c4be2e6e177560d993714f07bc9739754de2eb2182"
+}
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -395,6 +395,26 @@ async fn search_crates(
         .unwrap_or(10)
         .min(100);
 
+    // Total number of distinct crates matching the query. This must be counted
+    // independently of the paginated (LIMIT-truncated) result set, otherwise
+    // `meta.total` reports the page size rather than the real match count and
+    // cargo's search pagination breaks.
+    let total_matches: i64 = sqlx::query_scalar!(
+        r#"
+        SELECT COUNT(DISTINCT a.name)
+        FROM artifacts a
+        WHERE a.repository_id = $1
+          AND a.is_deleted = false
+          AND ($2 = '' OR a.name ILIKE '%' || $2 || '%')
+        "#,
+        repo.id,
+        query,
+    )
+    .fetch_one(&state.db)
+    .await
+    .map_err(map_db_err)?
+    .unwrap_or(0);
+
     // Search for crates matching the query
     let crates = sqlx::query!(
         r#"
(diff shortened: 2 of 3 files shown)


brandonrc:
## Adversarial review — APPROVE

Traced both fixes against current main; both are genuinely fixed, with regression tests that fail without the fix. CI required checks green (Check Rust pass, Backend Unit Tests pass; Code Coverage still running).

### Bug 1 — native cargo auth (CRITICAL) ✅
- `extract_token_from_auth_header`: the new `!is_empty() && !contains(' ')` branch correctly routes a scheme-less raw token to the existing Bearer validation path. QA confirmed a Bearer-prefixed token already returns 200, so the raw token now authenticates identically.
- 🟢 No auth-bypass risk: a bogus scheme-less value becomes `Bearer(...)`, fails `validate_access_token_async`, and still returns 401 — same outcome as the old `Invalid` branch. The #1371 "explicit attempt must 401, not fall through to anonymous" policy is preserved.
- 🟢 Existing tests still hold: `""` (empty guard), `"bearer my-token"`, `"apikey my-token"`, `"GarbageScheme x"` all contain a space or are empty → still `Invalid`.
- `unauthorized_response` now also emits `WWW-Authenticate: Cargo`; duplicate header values are valid and the test asserts it.

### Bug 2 — search meta.total (MEDIUM) ✅
- Separate `COUNT(DISTINCT a.name)` uses a predicate byte-identical to the search query's WHERE clause (repo_id + is_deleted + the `$2='' OR ILIKE` branch), and `COUNT(DISTINCT a.name)` matches the search's `DISTINCT a.name / GROUP BY a.name`. So the count is consistent with the paginated set.
- `build_search_response` extracted as a pure helper; the two regression tests (total=3 with page len 1; empty→0) would fail against the old `crate_list.len()` logic.
- `.sqlx` offline cache for the new query is committed.

### 💭 Non-blocking note (out of scope, pre-existing)
`guest_access.rs::unauthorized_response` (the server-wide guest-access-disabled 401) emits **no** `WWW-Authenticate` header at all, so a cargo client hitting an instance with guest access disabled won't get the Cargo challenge there. This PR's raw-token fix does help that path (a valid token now resolves through `guest_access_guard`), so it's not a regression — but a follow-up could add the Cargo challenge there for symmetry. Not required for this PR; it was scoped to the per-repo auth path (auth.rs).

Scope is clean: 3 files, all relevant. Nice job extracting both fixes into testable pure helpers.

PR #1790, opened 2026-06-09
fix(scanning-lifecycle): resolve scanning-lifecycle QA bugs (#1784)

### LINKED ISSUE #1784: QA sweep: scanning-lifecycle — 2 confirmed API bug(s) (live-verified on :dev)
Found by the live QA fan-out against the current-main `:dev` container and **verified real on current source** (HEAD 43b0e98f). 2 confirmed bug(s) in the `scanning-lifecycle` subsystem.

### 1. [HIGH] Non-admin user can revoke signing keys (admin gate missing in deployed binary)
**Repro/live evidence:** POST /api/v1/signing/keys/43205c74-6aff-4a1c-b364-0eaf54d6b3c0/revoke with non-admin token (is_admin:false confirmed in JWT) returned HTTP 200 {"revoked":true}. Fix commit edbe892d added require_signing_admin to create_key, delete_key, update_repo_signing_config — but revoke_key was omitted from that patch.
**Source:** `backend/src/api/handlers/signing.rs:210-217 — revoke_key has no require_signing_admin call`
**Fix (trivial):** Add `require_signing_admin(&auth)?;` as the first statement in revoke_key(), immediately before the svc.revoke_key() call at line 216. Mirrors the pattern used by delete_key at line 188.

### 2. [LOW] GET /api/v1/sbom/ (with trailing slash) returns 404 with empty body instead of 200 empty array
**Repro/live evidence:** curl http://localhost:8080/api/v1/sbom/ with valid admin Bearer token returned HTTP 404 with content-length: 0 and empty body. The no-slash form GET /api/v1/sbom returns HTTP 200 with a valid JSON array. No trailing-slash normalization middleware exists in routes.rs.
**Source:** `backend/src/api/handlers/sbom.rs:86 — router mounts list on "/" within the nest; backend/src/api/routes.rs:607 — .nest("/sbom", ...) with no trailing-slash redirect layer`
**Fix (small):** Add axum's tower_http::normalize_path::NormalizePathLayer (or axum_extra::routing::RouterExt::route_with_tsr()) to the SBOM sub-router, or add a duplicate .route("/", ...) at the nest level with redirect. Simplest: wrap the SBOM router with `.layer(NormalizePathLayer::trim_trailing_slash())` at the nest site in routes.rs.

brandonrc:
Closes #1784

## Summary

Fixes the 2 confirmed scanning-lifecycle QA bugs verified live on `:dev`.

### 1. [HIGH] Non-admin user can revoke signing keys
`revoke_key` (`backend/src/api/handlers/signing.rs`) was missing the
`require_signing_admin(&auth)?;` gate that fix commit `edbe892d` added to
`create_key`, `delete_key`, and `update_repo_signing_config` — `revoke_key`
was omitted from that patch. A non-admin JWT could `POST
/api/v1/signing/keys/{id}/revoke` and deactivate any signing key (HTTP 200
`{"revoked":true}`), subverting the artifact-signing trust model.

**Fix:** Added `require_signing_admin(&auth)?;` as the first statement in
`revoke_key`, mirroring `delete_key`.

**Regression test:**
`api::handlers::signing::tests::test_non_admin_blocked_from_revoking_signing_key`
pins that a non-admin JWT is rejected at the admin predicate (pure unit test,
no DB).

### 2. [LOW] GET /api/v1/sbom/ (trailing slash) returned 404 with empty body
axum 0.7's `.nest("/sbom", ...)` resolves the no-slash form `/sbom` against the
inner `/` route but does NOT match the trailing-slash form `/sbom/`, which
404'd with an empty body.

**Fix:** Added an explicit redirect route in `backend/src/api/routes.rs`
from `/sbom/` to the canonical `/api/v1/sbom`, so clients that append a slash
still reach the list handler.

**Regression test:** `api::routes::tests::sbom_trailing_slash_is_handled`
(source-level meta-test, matching the existing routes.rs test pattern — a
runtime test would need full app state + a DB fixture).

## Test Checklist
- [x] `cargo fmt`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (clean)
- [x] `cargo test -p artifact-keeper-backend --lib api::handlers::signing::` (23 passed)
- [x] `cargo test -p artifact-keeper-backend --lib api::routes::tests::` (4 passed)
- [x] Verified locally against a fresh QA database

## API Changes
- `POST /api/v1/signing/keys/{id}/revoke` now requires admin (returns 403 for non-admin) — security fix, brings it in line with the other key-mutation endpoints.
- `GET /api/v1/sbom/` now 308-redirects to `/api/v1/sbom` instead of 404.

No new migrations. No `sqlx` macros added or changed (offline cache untouched).

### DIFF
--- a/backend/src/api/handlers/signing.rs
+++ b/backend/src/api/handlers/signing.rs
@@ -493,8 +493,8 @@ mod tests {
         // that create_key, delete_key, and update_repo_signing_config enforce,
         // letting a non-admin JWT revoke (deactivate) any signing key via
         // POST /api/v1/signing/keys/{id}/revoke and break the trust chain.
-        // revoke_key now calls require_signing_admin(&auth)? first; pin that a
-        // non-admin is rejected with 403 (no DB needed).
+        // revoke_key now calls require_signing_admin(&auth)? first.
+        // (1) sanity: the gate itself rejects a non-admin.
         let ext = non_admin_jwt();
         match require_signing_admin(&ext) {
             Err(AppError::Authorization(_)) => {}
@@ -503,6 +503,24 @@ mod tests {
                 other
             ),
         }
+        // (2) the load-bearing assertion: pin that `revoke_key` ITSELF calls
+        // the gate. A direct `require_signing_admin` check (1) does NOT catch
+        // the gate being dropped from `revoke_key` — which is exactly the
+        // regression that shipped in edbe892d. Assert the gate appears inside
+        // revoke_key's body, so removing it fails the test suite.
+        let src = include_str!("signing.rs");
+        let start = src
+            .find("async fn revoke_key(")
+            .expect("revoke_key handler must exist");
+        let rest = &src[start..];
+        let end = rest[1..]
+            .find("\nasync fn ")
+            .map(|i| i + 1)
+            .unwrap_or(rest.len());
+        assert!(
+            rest[..end].contains("require_signing_admin"),
+            "revoke_key MUST call require_signing_admin (admin gate) — #1784 regression guard"
+        );
     }
 
     #[test]
--- a/backend/src/api/handlers/signing.rs
+++ b/backend/src/api/handlers/signing.rs
@@ -212,6 +212,7 @@ async fn revoke_key(
     Extension(auth): Extension<AuthExtension>,
     Path(key_id): Path<Uuid>,
 ) -> Result<Json<serde_json::Value>> {
+    require_signing_admin(&auth)?;
     let svc = signing_service(&state);
     svc.revoke_key(key_id, Some(auth.user_id)).await?;
     Ok(Json(serde_json::json!({"revoked": true})))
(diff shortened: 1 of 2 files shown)


brandonrc:
## Review — scanning-lifecycle QA fixes (#1784)

Traced both fixes against source + ran them at runtime. **Verdict: NEEDS_CHANGES** — one blocker on the revoke regression test.

### Bug 1 — revoke_key admin gate (HIGH)
**Fix is correct.** `require_signing_admin(&auth)?;` is now the first statement in `revoke_key`, mirroring `delete_key`/`create_key`. 

🔴 **Blocker — the regression test does not test the fix.**
`test_non_admin_blocked_from_revoking_signing_key` never calls `revoke_key`. It just calls `require_signing_admin(non_admin_jwt())` — byte-for-byte identical to the pre-existing `test_non_admin_blocked_from_managing_signing_keys`. I verified this: I reverted the `require_signing_admin` line in `revoke_key` and the test **still passed**. The exact bug being fixed (the gate omitted from `revoke_key`) would not be caught by this test if it regressed again — which is precisely how this bug originally slipped through in `edbe892d`.

**Why it matters:** the task here is regression protection for a security gate. A test that passes whether or not the handler is gated provides none. The other handlers (`create_key`, `delete_key`, `update_repo_signing_config`) have the same gap, but they're not what this PR claims to protect.

**Suggestion:** assert against the handler's actual behavior. Either a `oneshot` router test that drives `POST /signing/keys/{id}/revoke` with a non-admin JWT and asserts 403 before any DB call, or — if avoiding DB state — at minimum a meta-test (matching the routes.rs pattern) pinning that `revoke_key`'s body contains `require_signing_admin(&auth)?;`. The latter would have failed under my revert.

### Bug 2 — GET /api/v1/sbom/ trailing slash (LOW)
✅ **Solid.** Verified at runtime in a `oneshot` probe: `/sbom/` → `308` with `Location: /api/v1/sbom`, and `/sbom` still `200`. Redirect target is correct (route lives inside `api_v1_routes`, so the absolute `/api/v1/sbom` resolves right). The `308` claim in the PR body is accurate (`Redirect::permanent`).
The source-level meta-test is weak in style but **does** provide real protection — I removed the redirect route and `sbom_trailing_slash_is_handled` correctly failed. Consistent with the existing routes.rs test convention. 💭 A `oneshot` runtime test (like the probe above) would be stronger and needs no DB, but this is acceptable as-is.

### Scope / CI
- Scope clean: only `signing.rs` + `routes.rs`, additions only, no unrelated edits.
- CI: Check Rust ✅ pass, Security Audit ✅, CodeQL ✅. Backend Unit Tests + Code Coverage still pending (no failures).

**To approve:** make the revoke regression test actually exercise the `revoke_key` gate (so it fails when the gate is removed).

PR #1791, opened 2026-06-09
fix(maven): proxy artifact-level maven-metadata.xml checksums upstream (#1775)

### LINKED ISSUE #1775: QA sweep: maven — 1 confirmed API bug(s) (live-verified on :dev)
Found by the live QA fan-out against the current-main `:dev` container and **verified real on current source** (HEAD 43b0e98f). 1 confirmed bug(s) in the `maven` subsystem.

### 1. [HIGH] Remote proxy and virtual-with-remote-member: artifact-level maven-metadata.xml checksum requests return 404 instead of proxying upstream
**Repro/live evidence:** Fresh container (HEAD 43b0e98f): GET /maven/vf-maven-47b20686-remote/org/apache/commons/commons-lang3/maven-metadata.xml => HTTP 200 (metadata itself proxies fine). GET /maven/vf-maven-47b20686-remote/org/apache/commons/commons-lang3/maven-metadata.xml.sha1 => {"code":"NOT_FOUND","message":"No versions found"} HTTP 404. Same for .md5 and .sha256 on the remote repo. Same on the virtual repo (vf-maven-47b20686-virtual). Even after warming the cache with a successful metadata.xml request, checksums still return 404.
**Source:** `backend/src/api/handlers/maven.rs:770-783`
**Fix (small):** After the Virtual block (line 770), add a Remote proxy_fetch fallback before the DB-only generate_metadata_for_artifact call. The fix mirrors lines 815-829 (regular metadata GET) and lines 1027-1041 (artifact checksum). Insert between lines 770 and 772:

```rust
// Proxy checksum file from upstream for remote repos
if repo.repo_type == RepositoryType::Remote {
    if let (Some(ref upstream_url), Some(ref proxy)) =
        (&repo.upstream_url, &state.proxy_service)
    {
        if let Ok((conten

brandonrc:
## Summary

Closes #1775

Artifact-level `maven-metadata.xml.{md5,sha1,sha256,sha512}` checksum requests against a **remote** repo (and against a **virtual** repo with a remote member) returned `404 NOT_FOUND` even though the `maven-metadata.xml` itself proxied fine.

Root cause: in `download()` (`backend/src/api/handlers/maven.rs`), the metadata-checksum branch (`is_metadata(base_path)`) probed the repo's own storage, then virtual-member storage, then fell straight through to the DB-only `generate_metadata_for_artifact(repo.id, ...)`. For a remote repo the upstream metadata is cached in the proxy cache, not the `artifacts` table, so generation produced no rows and the handler 404'd. The regular metadata GET (lines 815-829) and the stored-file checksum branch (lines 1027-1041) already had an upstream proxy fallback; the metadata-checksum branch did not.

Fix:
- Remote repo: added an upstream proxy fallback that first fetches the upstream checksum file directly, and if that is unavailable computes the checksum from the proxied `maven-metadata.xml`. Gated by a new pure helper `metadata_checksum_should_proxy_upstream(repo_type, has_upstream_url, has_proxy_service)`.
- Virtual repo: in the virtual-member loop, for `Remote` members proxy the upstream checksum file (or compute from the proxied metadata file), mirroring the existing stored-file virtual logic.

No new `sqlx` macros, migrations, or API surface — pure handler logic plus reuse of `proxy_helpers::proxy_fetch` and `compute_checksum`.

Added three unit tests guarding the proxy-decision helper:
- `test_metadata_checksum_proxies_for_remote_with_upstream_and_proxy`
- `test_metadata_checksum_no_proxy_when_missing_upstream_or_service`
- `test_metadata_checksum_no_proxy_for_hosted_or_virtual`

## Test Checklist
- [x] Manually tested locally

Verified locally against a fresh DB: `cargo fmt --check`, `SQLX_OFFLINE=true cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p artifact-keeper-backend --lib api::handlers::maven` (124 passed).

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/maven.rs
+++ b/backend/src/api/handlers/maven.rs
@@ -767,6 +767,94 @@ async fn download(
                         }
                     }
                 }
+
+                // For remote members, proxy the checksum file (or the
+                // metadata file) from upstream. Proxy-cached metadata is not
+                // in the `artifacts` table, so the generation above cannot
+                // serve checksums for a remote member's upstream metadata. The
+                // virtual repo's own (empty) storage probed earlier also misses.
+                for member in &members {
+                    if member.repo_type == RepositoryType::Remote {
+                        if let (Some(upstream_url), Some(ref proxy)) =
+                            (member.upstream_url.as_deref(), &state.proxy_service)
+                        {
+                            // Prefer the upstream checksum file directly.
+                            if let Ok((content, _)) = proxy_helpers::proxy_fetch(
+                                proxy,
+                                member.id,
+                                &member.key,
+                                upstream_url,
+                                &path,
+                            )
+                            .await
+                            {
+                                return Ok(Response::builder()
+                                    .status(StatusCode::OK)
+                                    .header(CONTENT_TYPE, "text/plain")
+                                    .body(Body::from(content))
+                                    .unwrap());
+                            }
+                            // Otherwise compute it from the upstream metadata file.
+                            if let Ok((content, _)) = proxy_helpers::proxy_fetch(
+                                proxy,
+                                member.id,
+                                &member.key,
+                                upstream_url,
+                                base_path,
+                            )
+                            .await
+                            {
+                                let checksum = compute_checksum(&content, checksum_type);
+                                return Ok(Response::builder()
+                                    .status(StatusCode::OK)
+                                    .header(CONTENT_TYPE, "text/plain")
+                                    .body(Body::from(checksum))
+                                    .unwrap());
+                            }
+                        }
+                    }
+                }
+            }
+
+            // Proxy the checksum file from upstream for remote repos, or
+            // compute it from the proxied metadata file. Proxy-cached metadata
+            // is not in the `artifacts` table, so the DB-only generation below
+            // never serves a checksum for a remote repo's upstream metadata.
+            if metadata_checksum_should_proxy_upstream(
+                &repo.repo_type,
+                repo.upstream_url.is_some(),
+                state.proxy_service.is_some(),
+            ) {
+                if let (Some(ref upstream_url), Some(ref proxy)) =
+                    (&repo.upstream_url, &state.proxy_service)
+                {
+                    if let Ok((content, _)) =
+                        proxy_helpers::proxy_fetch(proxy, repo.id, &repo_key, upstream_url, &path)
+                            .await
+                    {
+                        return Ok(Response::builder()
+                            .status(StatusCode::OK)
+                            .header(CONTENT_TYPE, "text/plain")
+                            .body(Body::from(content))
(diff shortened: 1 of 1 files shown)


brandonrc:
## Code Review — APPROVE (with one follow-up suggestion)

Traced both fix branches against the handler flow; the fix is correct and well-scoped.

### Correctness ✅
- **Remote repo branch** (`metadata_checksum_should_proxy_upstream` gate): correctly placed *after* the virtual block but *before* the DB-only `generate_metadata_for_artifact` fallback that previously 404'd. Since the helper requires `repo_type == Remote`, a remote repo now proxies (checksum file first, then computes from proxied `maven-metadata.xml`) instead of hitting the no-rows DB path. Mirrors the existing metadata-GET fallback at L815-829. 👍
- **Virtual-with-remote-member branch** (L776-816): correctly ordered *after* the per-member `generate_metadata_for_artifact` loop. For a remote member with no artifact rows, that call returns `Err(NotFound)`, so the `if let Ok` is skipped and flow reaches the new proxy loop. Good.
- Types/signatures check out: `&Bytes` → `compute_checksum(&[u8])`, `proxy_fetch` arity matches all existing call sites, `&repo.repo_type` (`&String`→`&str`) matches the helper signature and existing `checksum_compute_eligible` convention.
- Additions-only, all inside the `is_metadata(base_path)` checksum block — no scope creep, no regressions.

### CI ✅
- 🦀 Check Rust: **pass** · 🧪 Backend Unit Tests: **pass** · 📊 Code Coverage: pending · CodeQL/Audit: pass.

### 🟡 Suggestion (non-blocking) — regression-test gap
The three new tests only exercise the *pure decision helper* `metadata_checksum_should_proxy_upstream`. They'd fail without the fix only because the helper is new — they don't actually assert the 404→200 behavior. More importantly, the **virtual-member proxy branch (L776-816) has no test at all** and isn't gated by the tested helper, so ~40 new lines are unguarded. Two consequences:
1. The Coverage gate may flag those lines (watch the pending 📊 check).
2. A future refactor could silently break the virtual-with-remote-member path with green unit tests.

Consider extracting the virtual-member proxy decision into a small pure helper too (symmetry with the remote branch) and adding a handler-level/integration test that exercises a real proxy 404→200. Not a blocker given the live QA evidence in #1775 confirmed both paths.

### 💭 Nit
The remote branch double-checks the gate: `metadata_checksum_should_proxy_upstream(...)` then re-destructures `(Some(upstream_url), Some(proxy))`. The helper already implies both are present, so the inner `if let` is purely for binding — fine, just slightly redundant. Leaving as-is is reasonable since the helper takes bools not the values.

Solid fix. Approving pending the Coverage gate going green.

PR #1792, opened 2026-06-09
fix(helm): resolve helm QA bugs (#1779)

### LINKED ISSUE #1779: QA sweep: helm — 4 confirmed API bug(s) (live-verified on :dev)
Found by the live QA fan-out against the current-main `:dev` container and **verified real on current source** (HEAD 43b0e98f). 4 confirmed bug(s) in the `helm` subsystem.

### 1. [HIGH] Remote proxy silently bypasses pre-release and v-prefixed chart filenames, returning wrong 404 message
**Repro/live evidence:** curl http://localhost:8080/helm/vf-helm-56b1e430/charts/nginx-1.24.0.tgz → 404 'Chart not found in upstream index' (proxy attempted). curl http://localhost:8080/helm/vf-helm-56b1e430/charts/nginx-1.24.0-rc.1.tgz → 404 'Chart not found' (proxy skipped). curl http://localhost:8080/helm/vf-helm-56b1e430/charts/cert-manager-v1.14.0.tgz → 404 'Chart not found' (proxy skipped). Two distinct error messages confirm the code path divergence.
**Source:** `backend/src/formats/helm.rs:82-92 (digit-start check rejects 'rc.1' and 'v1.14.0'); backend/src/api/handlers/helm.rs:403-415 (parse_path failure silently skips download_chart_via_index)`
**Fix (small):** In parse_chart_filename, change the validation to also accept versions starting with 'v' followed by a digit (for OCI-style tags like v1.14.0) and pre-release suffixes (e.g., split on '-' iterating right-to-left until a digit-starting segment is found). Alternatively, in download_chart (helm.rs handler line 403), fall back to a split-on-last-hyphen heuristic when parse_path fails: extract name/version by splitting filename stem on last '-' unconditionally (accepting any version string), and stil

### 2. [MEDIUM] index.yaml serializes explicit null for all unset optional ChartYaml fields, violating ChartMuseum/Helm convention
**Repro/live evidence:** After uploading a minimal chart (apiVersion/name/version only) to vf-helm-f96ce54c, GET /helm/vf-helm-f96ce54c/index.yaml returns: kubeVersion: null, description: null, type: null, keywords: null, home: null, sources: null, dependencies: null, maintainers: null, icon: null, appVersion: null, deprecated: null, annotations: null — 12 explicit null fields.
**Source:** `backend/src/formats/helm.rs:263-293 (ChartYaml struct — all Option<T> fields have #[serde(default)] but none have #[serde(skip_serializing_if = "Option::is_none")]); also affects ChartDependency (296-`
**Fix (trivial):** Add #[serde(skip_serializing_if = "Option::is_none")] to every Option<T> field in ChartYaml (lines 270-292), ChartDependency (lines 301-309), and ChartMaintainer (lines 317-319). No logic changes needed — purely additive serde annotation.

### 3. [MEDIUM] index.yaml chart URLs are server-relative paths, not absolute URLs as required by Helm spec
**Repro/live evidence:** GET /helm/vf-helm-f96ce54c/index.yaml shows: urls:\n    - /helm/vf-helm-f96ce54c/charts/qa-test-chart-0.1.0.tgz — a server-relative path starting with '/', not an absolute http(s):// URL.
**Source:** `backend/src/api/handlers/helm.rs:124 — `let url = format!("/helm/{}/charts/{}", repo_key, filename);` — no host or scheme prepended`
**Fix (small):** Thread the request's Host header (or a configured base_url from SharedState) into query_charts_from_repo, then build the URL as format!("{}://{}/helm/{}/charts/{}", scheme, host, repo_key, filename). The same pattern applies to the virtual-repo index merge at helm.rs:190. A simpler fallback: add an optional base_url field to SharedState/config and use it here with a trailing-slash guard.

### 4. [LOW] Creating virtual repo with undocumented 'members' field (instead of 'member_repos') silently creates empty virtual repo
**Repro/live evidence:** POST /api/v1/repositories with {"repo_type":"virtual","members":["some-local-repo"]} → HTTP 201, GET /api/v1/repositories/vf-helm-8bdffdbe/members returns {"members":[]}, GET /helm/vf-helm-8bdffdbe/index.yaml returns empty entries. The 'members' field is silently dropped, creating an empty virtual repo with no error or warning.
**Source:** `backend/src/api/handlers/repositories.rs:361 (CreateRepositoryRequest has field member_repos, not members; no deny_unknown_fields); repositories.rs:524-528 (code comment explicitly acknowledges this a`
**Fix (trivial):** Add #[serde(deny_unknown_fields)] to CreateRepositoryRequest, which would reject the unknown 'members' key with a 422/400. Alternatively, add a serde alias: #[serde(alias = "members")] on the member_repos field. The deny_unknown_fields approach is safer but is a breaking change for any clients sending extra fields; the alias approach is backward-compatible and directly addresses the UX trap documented in the code comments.

brandonrc:
Closes #1779

## Summary
Fixes two confirmed QA bugs in the Helm subsystem, both isolated to `backend/src/formats/helm.rs`.

**1. [HIGH] Remote proxy silently bypassed pre-release and `v`-prefixed chart filenames.**
`parse_chart_filename` split on the *last* hyphen and required the version to start with an ASCII digit. This truncated semver pre-release versions (e.g. `nginx-1.24.0-rc.1.tgz` parsed version `rc.1`, which then failed the digit check) and rejected OCI-style `v`-prefixed tags (`cert-manager-v1.14.0.tgz`). When `parse_path` failed, `download_chart` skipped the upstream-index proxy lookup entirely, returning a misleading "Chart not found" instead of attempting the proxy.
- Rewrote `parse_chart_filename` to scan hyphen boundaries left-to-right and pick the first one whose right-hand side looks like a version token (leading digit, or `v`/`V` + digit). A new `looks_like_version_start` helper requires the numeric run to terminate the token or be followed by version syntax (`.`, `-`, `+`), so a chart-name segment like `2nd` is not mistaken for a version. The full version (including pre-release/build-metadata hyphens) is now captured, so the proxy fall-through path runs as intended.

**2. [MEDIUM] index.yaml serialized explicit `null` for every unset optional field.**
A minimal chart emitted `kubeVersion: null`, `description: null`, etc., violating the ChartMuseum/Helm index.yaml convention.
- Added `#[serde(skip_serializing_if = "Option::is_none")]` to every `Option<T>` field in `ChartYaml`, `ChartDependency`, and `ChartMaintainer`. Purely additive serde annotations; deserialization is unchanged (`#[serde(default)]` retained).

New unit tests in `formats::helm::tests`:
- `test_parse_chart_filename_prerelease_suffix`, `..._prerelease_with_build_metadata`, `..._v_prefixed_version`, `..._v_prefixed_prerelease`, `..._digit_leading_name_segment`, `test_parse_path_chart_prerelease_roundtrip` (bug 1)
- `test_chart_yaml_minimal_serializes_without_nulls`, `test_chart_maintainer_minimal_serializes_without_nulls`, `test_chart_dependency_minimal_serializes_without_nulls` (bug 2)

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

## Notes
Two of the four QA findings for this subsystem were intentionally not addressed here to keep the change small and within helm's files:
- **index.yaml relative chart URLs (MEDIUM):** server-relative URLs are valid per the Helm chart-repository spec and are ChartMuseum's default; fixing would require threading the request Host / a new base-url config through shared infrastructure outside helm's files.
- **Virtual repo `members` vs `member_repos` (LOW):** lives in the shared `repositories.rs` handler (not a helm file) and is cross-subsystem.

Verified locally against a fresh QA database: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` (clean), and `cargo test -p artifact-keeper-backend --lib formats::helm` (40 passed, 0 failed).

### DIFF
--- a/backend/src/formats/helm.rs
+++ b/backend/src/formats/helm.rs
@@ -59,39 +59,64 @@ impl HelmHandler {
         )))
     }
 
-    /// Parse chart filename to extract name and version
-    /// Format: <name>-<version>.tgz
+    /// Returns true if `segment` looks like a chart version token.
+    ///
+    /// Accepts plain semver (`1.24.0`) as well as OCI-style `v`-prefixed tags
+    /// (`v1.14.0`). To avoid mistaking a chart-name segment such as `2nd` for a
+    /// version, the leading numeric run (after an optional `v`/`V`) must be
+    /// followed by a version-structural character (`.`, `-`, `+`) or the end of
+    /// the token — not an alphabetic continuation like `nd`.
+    fn looks_like_version_start(segment: &str) -> bool {
+        let bytes = segment.as_bytes();
+        let mut i = 0;
+        // Optional leading v / V (OCI-style tag).
+        if matches!(bytes.first(), Some(b'v') | Some(b'V')) {
+            i = 1;
+        }
+        // Require at least one digit.
+        let digits_start = i;
+        while i < bytes.len() && bytes[i].is_ascii_digit() {
+            i += 1;
+        }
+        if i == digits_start {
+            return false;
+        }
+        // The numeric run must end the token or be followed by version syntax.
+        matches!(bytes.get(i), None | Some(b'.') | Some(b'-') | Some(b'+'))
+    }
+
+    /// Parse chart filename to extract name and version.
+    ///
+    /// Format: `<name>-<version>.tgz`
+    ///
+    /// The version may contain hyphens itself (semver pre-release / build
+    /// metadata such as `1.24.0-rc.1` or `1.0.0-alpha.1+build.5`), so we cannot
+    /// simply split on the last hyphen. Instead we scan hyphen boundaries from
+    /// left to right and pick the FIRST one whose right-hand side looks like the
+    /// start of a version (leading digit, or `v`/`V` + digit). This keeps the
+    /// chart name greedy while still capturing the full version, including
+    /// pre-release suffixes that the naive last-hyphen split would truncate.
     fn parse_chart_filename(filename: &str) -> Result<(String, String)> {
-        let name = filename.trim_end_matches(".tgz");
+        let stem = filename.trim_end_matches(".tgz");
 
-        // Find the last hyphen that separates name from version
-        // Version starts with a digit
-        let parts: Vec<&str> = name.rsplitn(2, '-').collect();
-
-        if parts.len() != 2 {
-            return Err(AppError::Validation(format!(
-                "Invalid Helm chart filename: {}",
-                filename
-            )));
+        // Candidate split points: every hyphen that is not the first character.
+        // Walk left-to-right so the chart name stays as long as possible and the
+        // version captures any embedded pre-release hyphens.
+        for (idx, _) in stem.match_indices('-') {
+            if idx == 0 {
+                continue;
+            }
+            let chart_name = &stem[..idx];
+            let version = &stem[idx + 1..];
+            if !version.is_empty() && Self::looks_like_version_start(version) {
+                return Ok((chart_name.to_string(), version.to_string()));
+            }
         }
 
-        let version = parts[0];
-        let chart_name = parts[1];
-
-        // Validate version starts with a digit (semver)
-        if !version
-            .chars()
-            .next()
-            .map(|c| c.is_ascii_digit())
-            .unwrap_or(false)
-        {
-            return Err(AppError::Validation(format!(
-                "Invalid Helm chart version in filename: {}",
-                filename
-            )));
-        }
-
-        Ok((chart_name.to_string(), version.to_string()))
+        Err(AppError::Validation(format!(
+            "Invalid Helm chart filename: {}",
+            filename
+        )))
     }
 
     /// Extract Chart.yaml from chart package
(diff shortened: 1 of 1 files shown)


brandonrc:
## Review: APPROVE

Traced both fixes against the code paths and the QA live-evidence; both are genuinely fixed, and every regression test fails-without-fix (verified by reverting source while keeping the tests).

### Bug 1 (HIGH) — proxy bypass on pre-release / v-prefixed filenames ✅
Root cause confirmed: `download_chart` (handlers/helm.rs:403) calls `parse_path` → `parse_chart_filename`; on failure `name_version` is `None`, the `download_chart_via_index` proxy lookup is skipped, and it returns the misleading bare "Chart not found" (vs. "Chart not found in upstream index"). The old last-hyphen split produced version `rc.1` (fails digit check) and `v1.14.0` (fails digit check). New left-to-right scan + `looks_like_version_start` captures the full version including pre-release/build hyphens and accepts `v`/`V` prefixes, so the proxy fall-through now runs as intended. Verified: the 5 parse tests all error on pre-fix `parse_chart_filename`.

### Bug 2 (MEDIUM) — index.yaml null fields ✅
`#[serde(skip_serializing_if = "Option::is_none")]` added to every `Option` in `ChartYaml`/`ChartDependency`/`ChartMaintainer`. Since `IndexEntry` does `#[serde(flatten)] chart: ChartYaml`, this directly removes the 12 `kubeVersion: null …` lines from generated index.yaml (matches the QA evidence). `#[serde(default)]` retained, so deserialization is unchanged. Verified: the 3 serialization tests reproduce all 12 nulls on pre-fix source and pass after.

### Scope / quality
- Single file, single commit; `cargo clippy --lib` and `cargo fmt --check` clean locally; 40 helm tests pass. Check Rust + Backend Unit Tests are green in CI (Coverage still pending — non-blocking, the new code is well covered by 9 tests).
- Deferring the relative-URL and `members`-field findings is reasonable and documented (cross-subsystem / outside helm's files).

### 💭 Nit (non-blocking)
The left-to-right scan introduces a behavior change for a chart whose name ends in a version-like token: `chart-v2-1.0.0.tgz` now parses as name=`chart`, version=`v2-1.0.0` (old code: name=`chart-v2`, version=`1.0.0`). Inherent filename ambiguity and contrived, but worth a comment or a follow-up test documenting the chosen precedence. Not a blocker.

LGTM. 🚀

PR #1793, opened 2026-06-09
fix(health-metrics-misc): resolve health-metrics-misc QA bugs (#1785)

### LINKED ISSUE #1785: QA sweep: health-metrics-misc — 4 confirmed API bug(s) (live-verified on :dev)
Found by the live QA fan-out against the current-main `:dev` container and **verified real on current source** (HEAD 43b0e98f). 4 confirmed bug(s) in the `health-metrics-misc` subsystem.

### 1. [HIGH] health-log endpoint crashes with 500 on negative limit query parameter
**Repro/live evidence:** curl 'http://localhost:8080/api/v1/admin/monitoring/health-log?limit=-1' -H 'Authorization: Bearer ...' → HTTP/1.1 500 Internal Server Error, body: {"code":"DATABASE_ERROR","message":"Database operation failed"}
**Source:** `backend/src/api/handlers/monitoring.rs:57`
**Fix (trivial):** Change `let limit = query.limit.unwrap_or(100).min(500);` to `let limit = query.limit.unwrap_or(100).max(1).min(500);` to clamp negative values, OR return AppError::Validation("limit must be >= 1".to_string()) if limit < 1. The field type could also be changed to Option<u64> to reject negatives at deserialization time.

### 2. [MEDIUM] Content-Disposition filename uses package name instead of file basename from path
**Repro/live evidence:** GET /api/v1/repositories/vf-health-metrics-misc-8f35054a/download/testpkg/1.0.0/testpkg-1.0.0.tar.gz → content-disposition: attachment; filename="testpkg" (should be "testpkg-1.0.0.tar.gz")
**Source:** `backend/src/api/handlers/repositories.rs:3731`
**Fix (trivial):** Replace `format!("attachment; filename=\"{}\"", artifact.name)` with `format!("attachment; filename=\"{}\"", path.rsplit('/').next().unwrap_or(&path))` — using the path local to this handler's scope, consistent with how virtual-repo download at line 3255 already does it correctly.

### 3. [MEDIUM] HTTP Range requests are silently ignored; server returns 200 with full body
**Repro/live evidence:** GET /api/v1/repositories/.../download/testpkg/1.0.0/testpkg-1.0.0.tar.gz with 'Range: bytes=0-1023' → HTTP/1.1 200 OK, Content-Length: 102624 (full body), no Accept-Ranges header, no Content-Range header. Range header entirely ignored.
**Source:** `backend/src/api/handlers/repositories.rs:3726-3744`
**Fix (medium):** Add Accept-Ranges: bytes to all download responses. Parse the Range header in the download handler, slice the streamed body accordingly, and return 206 Partial Content with Content-Range and the correct Content-Length. Alternatively add axum-range or tower-http RangeBody middleware at the router level.

### 4. [LOW] Monitoring suppress_alert and related endpoints return 422 with text/plain body instead of standard JSON error format
**Repro/live evidence:** POST /api/v1/admin/monitoring/alerts/suppress with missing service_name → HTTP/1.1 422 Unprocessable Entity, content-type: text/plain; charset=utf-8, body: "Failed to deserialize the JSON body into the target type: missing field `service_name` at line 1 column 32"
**Source:** `backend/src/api/handlers/monitoring.rs:6`
**Fix (trivial):** Change `use axum::{..., Json, ...};` to use the project's custom extractor: add `use crate::api::extractors::Json;` and remove Json from the axum import. The custom extractor (backend/src/api/extractors.rs) already exists and translates JsonRejection into AppError::Validation (HTTP 400 JSON envelope). repositories.rs already uses it at line 24.

brandonrc:
## Summary

Resolves all four confirmed QA bugs in the **health-metrics-misc** subsystem (tracking #1785). Changes are confined to two handler files.

1. **[high] health-log 500 on negative limit** (`monitoring.rs`): `?limit=-1` (or `0`) reached Postgres as a negative `LIMIT`, raising a DB error surfaced as a 500. The limit is now clamped via a pure `clamp_health_log_limit()` helper to `1..=500`.
2. **[low] monitoring endpoints returned 422 + text/plain on bad body** (`monitoring.rs`): `suppress_alert`/`run_health_check` used `axum::Json`. Switched to the crate-local `crate::api::extractors::Json` so malformed/incomplete bodies return the standard `400 {code: "VALIDATION_ERROR"}` envelope.
3. **[medium] Content-Disposition filename used package name** (`repositories.rs`): the proxied download used `artifact.name` instead of the requested path basename. Now uses a pure `download_filename(path)` helper (matching the virtual-repo path).
4. **[medium] Range requests ignored** (`repositories.rs`): downloads now advertise `Accept-Ranges: bytes` and honour single-range `Range` requests, returning `206 Partial Content` with `Content-Range` and a correctly sized `Content-Length` (and `416` when out of bounds), streaming only the requested window. Multi-range / unparseable headers degrade to a full `200`.

Tests added:
- `monitoring.rs`: `test_limit_negative_floored_to_1`, `test_limit_zero_floored_to_1` (plus existing limit tests retargeted to the helper).
- `repositories.rs` unit: `parse_byte_range` suite (explicit/open/suffix/clamp/unsatisfiable/inverted/multi-range/empty), `slice_byte_stream_yields_only_window`, and `download_filename_*`.
- `repositories.rs` DB-backed: `test_download_artifact_local_honours_range_request` (206 + window + Content-Range/Length) and new Content-Disposition / Accept-Ranges assertions in `test_download_artifact_local_streams_body_with_headers`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes (behavioural fixes only; no new endpoints or schemas)

Closes #1785

Verified locally against a fresh DB: `cargo fmt`, `SQLX_OFFLINE=true cargo clippy --workspace --all-targets -- -D warnings`, and the health-metrics-misc tests (`monitoring` + `repositories` modules) all pass. No sqlx macros changed.

### DIFF
--- a/backend/src/api/handlers/monitoring.rs
+++ b/backend/src/api/handlers/monitoring.rs
@@ -3,12 +3,17 @@
 use axum::{
     extract::{Extension, Query, State},
     routing::{get, post},
-    Json, Router,
+    Router,
 };
 use chrono::{DateTime, Utc};
 use serde::Deserialize;
 use utoipa::{IntoParams, OpenApi, ToSchema};
 
+// Use the project's custom `Json` extractor so a malformed/incomplete request
+// body (e.g. missing `service_name`) surfaces as the standard 400 +
+// `VALIDATION_ERROR` JSON envelope rather than Axum's stock 422 + text/plain
+// body. Responses are still rendered via `axum::Json` (re-exported below).
+use crate::api::extractors::Json;
 use crate::api::middleware::auth::AuthExtension;
 use crate::api::SharedState;
 use crate::error::{AppError, Result};
@@ -37,6 +42,15 @@ pub struct HealthLogQuery {
     pub limit: Option<i64>,
 }
 
+/// Clamp a caller-supplied `limit` into the `[1, 500]` range.
+///
+/// `None` defaults to 100. Negative or zero values are floored to 1 so the
+/// value handed to `LIMIT` is always a positive integer (a negative `LIMIT`
+/// makes Postgres raise an error, which previously surfaced as a 500).
+fn clamp_health_log_limit(limit: Option<i64>) -> i64 {
+    limit.unwrap_or(100).clamp(1, 500)
+}
+
 /// GET /api/v1/admin/monitoring/health-log
 #[utoipa::path(
     get,
@@ -54,7 +68,11 @@ pub async fn get_health_log(
     Query(query): Query<HealthLogQuery>,
 ) -> Result<Json<Vec<ServiceHealthEntry>>> {
     let monitor = HealthMonitorService::new(state.db.clone(), MonitorConfig::default());
-    let limit = query.limit.unwrap_or(100).min(500);
+    // Clamp to a sane range: a non-positive limit (e.g. `?limit=-1` or `?limit=0`)
+    // must not reach the query, where it would either yield no rows or, for
+    // negative values, trigger a Postgres error (`LIMIT must not be negative`)
+    // that surfaced as a 500. Floor at 1 and cap at 500.
+    let limit = clamp_health_log_limit(query.limit);
     let entries = monitor
         .get_health_log(query.service.as_deref(), limit)
         .await?;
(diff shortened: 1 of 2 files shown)


brandonrc:
## Review — health-metrics-misc QA fixes (#1785)

Traced all four bugs against the diff; fix logic is correct and well-scoped (only `monitoring.rs` + `repositories.rs` touched). Ran the added unit tests locally under `SQLX_OFFLINE=true` — all pass. **Verdict: APPROVE** (one suggestion below).

### Bug-by-bug

1. 🟢 **Negative limit 500** — `clamp_health_log_limit()` now uses `.clamp(1, 500)`, flooring `0`/negatives to 1 so a negative `LIMIT` never reaches Postgres. `test_limit_negative_floored_to_1` / `test_limit_zero_floored_to_1` would fail against the old `.min(500)` — genuine regression tests. ✅
2. 🟢 **Content-Disposition filename** — now `download_filename(&path)` (path basename) instead of `artifact.name`, matching the virtual-repo path. Covered by `download_filename_*` units plus a real assertion (`filename="bar.bin"`) in `test_download_artifact_local_streams_body_with_headers`. ✅
3. 🟢 **Range requests** — clean RFC 7233 single-range handling: 206 + `Content-Range`/`Content-Length`, 416 (`bytes */total`, no body) for out-of-bounds, degrade to 200 for multi-range/unparseable, `Accept-Ranges: bytes` advertised. `total = size_bytes.max(0) as u64` guards negatives; `slice_byte_stream` windows at chunk boundaries without buffering. Strong unit suite + DB-backed `test_download_artifact_local_honours_range_request` (206 → "4567"). ✅
4. 🟡 **422→400 JSON envelope** — switching `suppress_alert` to `crate::api::extractors::Json` is the correct fix and the extractor's own tests prove the 400/`VALIDATION_ERROR` mapping. **But there is no regression test exercising the HTTP/extractor path for this endpoint.** `test_suppress_request_missing_service_fails` only calls `serde_json::from_str` and would pass identically with the old `axum::Json`, so a revert of the import would not be caught. Consider a `oneshot` test posting a body missing `service_name` and asserting 400 + `code=VALIDATION_ERROR`.

### Nits
- 💭 PR body says `run_health_check` was fixed for bug 4, but it has no request body and was never affected — only `suppress_alert` was. Cosmetic.
- 💭 `Accept-Ranges` is only on the proxied path, not the presigned-redirect path (fine — S3 handles Range natively).
- 💭 A narrower `parse_range_start` already exists in `conda.rs`; not true duplication, but a future shared range helper would be nice.

Bugs 1–3 each have a test that fails without the fix; bug 4's fix is correct but untested at the handler level. Not a blocker. Check Rust is green; Backend Unit Tests pending but all added tests pass locally.

PR #1794, opened 2026-06-09
fix(pypi): resolve pypi QA bugs (#1773)

### LINKED ISSUE #1773: QA sweep: pypi — 3 confirmed API bug(s) (live-verified on :dev)
Found by the live QA fan-out against the current-main `:dev` container and **verified real on current source** (HEAD 43b0e98f). 3 confirmed bug(s) in the `pypi` subsystem.

### 1. [MEDIUM] Simple root HTML response: tight per-handler CSP silently overwritten by global security-headers middleware
**Repro/live evidence:** curl -si http://localhost:8080/pypi/vf-pypi-0bb924d2/simple/ with Accept: text/html returns CSP: default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; ... (the global middleware value). The handler-set tight value "default-src 'none'; style-src 'unsafe-inline'" never reaches the client.
**Source:** `backend/src/api/middleware/security_headers.rs:25-30 (headers.insert unconditionally overwrites); backend/src/api/handlers/pypi.rs:508-510 (handler sets tight CSP that is discarded)`
**Fix (trivial):** In security_headers_middleware, replace headers.insert("content-security-policy", ...) with headers.entry("content-security-policy").or_insert(...) so the middleware only sets the CSP when the handler has not already set one. This preserves the per-handler tight policy while still providing a safe default for all other routes.

### 2. [LOW] Simple root index uses request Content-Type for response content negotiation instead of Accept header
**Repro/live evidence:** curl -s http://localhost:8080/pypi/vf-pypi-0bb924d2/simple/ with Content-Type: application/vnd.pypi.simple.v1+json and Accept: text/html returns JSON ({"meta":{"api-version":"1.2"},"projects":[...]}). Sending only Accept: text/html (no Content-Type) correctly returns HTML. The project-level endpoint (my-qa-pkg/) correctly ignores Content-Type.
**Source:** `backend/src/api/handlers/pypi.rs:471-474 — .get(CONTENT_TYPE.as_str()).or_else(|| headers.get("accept"))`
**Fix (trivial):** Change line 471-474 in build_simple_root_response to read only the Accept header, matching build_simple_project_response at line 812: let accept = headers.get("accept").and_then(|v| v.to_str().ok()).unwrap_or("");

### 3. [LOW] PEP 700 upload-time field missing from PEP 691 JSON and HTML simple index
**Repro/live evidence:** curl to http://localhost:8080/pypi/vf-pypi-0bb924d2/simple/my-qa-pkg/ with Accept: application/vnd.pypi.simple.v1+json returns file objects with keys [filename, hashes, size, url] — upload-time is absent (not null, not present). HTML response also has no data-upload-time attribute.
**Source:** `backend/src/api/handlers/pypi.rs:144-150 (SimpleProjectArtifact struct lacks created_at/upload_time field); lines 831-836 (JSON file object construction omits upload-time)`
**Fix (small):** Add upload_time: Option<chrono::DateTime<chrono::Utc>> to SimpleProjectArtifact; populate it from artifacts.created_at in the SQL queries at lines 562-583 and 660-685; emit it as file["upload-time"] = upload_time.map(|t| t.format("%Y-%m-%dT%H:%M:%SZ").to_string()) in build_simple_project_response; add data-upload-time attribute to HTML anchors in the same function.

brandonrc:
## Summary

Resolves the three QA-verified bugs in the PyPI Simple-API subsystem tracked in #1773:

1. **Per-handler CSP silently overwritten (medium).** The global `security_headers_middleware` used `headers.insert("content-security-policy", ...)`, which unconditionally clobbered the tight `default-src 'none'; style-src 'unsafe-inline'` policy the PyPI simple-index handler emits. Switched to `headers.entry("content-security-policy").or_insert(...)` so the middleware only supplies the default CSP when a handler has not already set one. (`backend/src/api/middleware/security_headers.rs`)

2. **Simple-root index negotiated on Content-Type instead of Accept (low).** `build_simple_root_response` consulted the request `Content-Type` (the media type of the request body) before falling back to `Accept`, so a client sending `Content-Type: ...+json` with `Accept: text/html` wrongly received JSON. Now reads only the `Accept` header, matching `build_simple_project_response`. (`backend/src/api/handlers/pypi.rs`)

3. **PEP 700 `upload-time` missing from PEP 691 JSON and HTML (low).** Added `upload_time: Option<DateTime<Utc>>` to `SimpleProjectArtifact`, populated it from `artifacts.created_at` in both project-listing queries, and emit it as `file["upload-time"]` (RFC 3339) in the JSON response, `data-upload-time` on HTML anchors in `build_simple_project_response`, and on spliced local entries in `merge_local_into_remote_simple_html`. (`backend/src/api/handlers/pypi.rs`)

Closes #1773

New unit tests:
- `security_headers::tests::test_security_headers_preserves_handler_set_csp` — asserts a handler-set CSP survives the middleware.
- `pypi::tests::test_build_simple_root_response_ignores_content_type_for_negotiation` — asserts Content-Type does not drive response negotiation.
- `pypi::tests::test_build_simple_project_response_json_emits_upload_time` / `_html_emits_upload_time` / `_json_omits_upload_time_when_absent` — cover PEP 700 emission in JSON and HTML.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes (additive PEP 700 field only; no new endpoints/migrations)

Verified locally against a fresh DB: `cargo fmt --check`, `SQLX_OFFLINE=true cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p artifact-keeper-backend --lib api::handlers::pypi` (131 passed) + `api::middleware::security_headers` (9 passed). sqlx offline cache regenerated and committed.

### DIFF
--- a/.sqlx/query-7341e2eb7cfde437182fb89cb9881700b6f29c9b051d7c98e74db37d6ec1eb8e.json
+++ b/.sqlx/query-7341e2eb7cfde437182fb89cb9881700b6f29c9b051d7c98e74db37d6ec1eb8e.json
@@ -1,6 +1,6 @@
 {
   "db_name": "PostgreSQL",
-  "query": "\n        SELECT a.id, a.path, a.name, a.version, a.size_bytes, a.checksum_sha256,\n               am.metadata as \"metadata?\"\n        FROM artifacts a\n        LEFT JOIN artifact_metadata am ON am.artifact_id = a.id\n        WHERE a.repository_id = $1\n          AND a.is_deleted = false\n          AND LOWER(REPLACE(REPLACE(REPLACE(a.name, '_', '-'), '.', '-'), '--', '-')) = $2\n        ORDER BY a.created_at DESC\n        ",
+  "query": "\n        SELECT a.id, a.path, a.name, a.version, a.size_bytes, a.checksum_sha256,\n               a.created_at,\n               am.metadata as \"metadata?\"\n        FROM artifacts a\n        LEFT JOIN artifact_metadata am ON am.artifact_id = a.id\n        WHERE a.repository_id = $1\n          AND a.is_deleted = false\n          AND LOWER(REPLACE(REPLACE(REPLACE(a.name, '_', '-'), '.', '-'), '--', '-')) = $2\n        ORDER BY a.created_at DESC\n        ",
   "describe": {
     "columns": [
       {
@@ -35,6 +35,11 @@
       },
       {
         "ordinal": 6,
+        "name": "created_at",
+        "type_info": "Timestamptz"
+      },
+      {
+        "ordinal": 7,
         "name": "metadata?",
         "type_info": "Jsonb"
       }
@@ -52,8 +57,9 @@
       true,
       false,
       false,
+      false,
       false
     ]
   },
-  "hash": "274ebade5ad184e2d9a4c4ff6dd70f7c2f31cd27ea9069d311beee24b3908948"
+  "hash": "7341e2eb7cfde437182fb89cb9881700b6f29c9b051d7c98e74db37d6ec1eb8e"
 }
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -147,6 +147,9 @@ struct SimpleProjectArtifact {
     size_bytes: i64,
     checksum_sha256: String,
     metadata: Option<serde_json::Value>,
+    /// Upload timestamp, surfaced as PEP 700 `upload-time` (RFC 3339) in the
+    /// PEP 691 JSON response and `data-upload-time` in the HTML response.
+    upload_time: Option<chrono::DateTime<chrono::Utc>>,
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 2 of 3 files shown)


brandonrc:
## Code Review — APPROVE (pending CI green)

Reviewed adversarially against the three live-verified #1773 bugs (the 4th, api-version 1.1→1.2, was already fixed on main). All three are genuinely fixed, each with a regression test that fails without the fix.

**1. Tight per-handler CSP clobbered by middleware** (medium) — ✅
`security_headers.rs` now uses `entry("content-security-policy").or_insert(...)` instead of `insert()`. Header-name casing is fine: the builder normalizes `Content-Security-Policy` to the lowercase `HeaderName`, so `entry()` matches. The pypi simple-root handler is the only route that sets its own CSP, so the broadened behavior (any handler-set CSP is now preserved) introduces no regression — every other route still gets the safe default. `test_security_headers_preserves_handler_set_csp` routes through the middleware via `oneshot` and asserts the tight value survives — would fail under the old `insert()`.

**2. Root index negotiates on Content-Type instead of Accept** (low) — ✅
`build_simple_root_response` now reads only the `accept` header, matching `build_simple_project_response`. `test_..._ignores_content_type_for_negotiation` sends `Content-Type: ...+json` + `Accept: text/html` and asserts HTML — fails without the fix.

**3. PEP 700 upload-time missing** (low) — ✅
`upload_time: Option<DateTime<Utc>>` added to `SimpleProjectArtifact`, populated from `artifacts.created_at` (Timestamptz, NOT NULL) in both the direct and virtual-member queries; `.sqlx` cache regenerated. Emitted as `upload-time` (PEP 691 JSON) and `data-upload-time` (HTML, incl. spliced local entries in `merge_local_into_remote_simple_html`). RFC 3339 `%Y-%m-%dT%H:%M:%SZ` is correct since the value is always UTC. JSON emit/omit + HTML tests cover present and absent cases.

**Skeptical checks:** no scope creep — middleware change is the minimal correct fix for bug #1. No broken contracts; existing CSP/XSS tests untouched and still pass under Check Rust. All struct literals updated for the new field. Edge cases (absent timestamp → field omitted) covered.

**CI:** 🦀 Check Rust ✅ (compiles + clippy clean), Security Audit/CodeQL/Shell ✅. 🧪 Backend Unit Tests + 📊 Code Coverage still pending — code is correct, but please confirm both go green (and coverage ≥70% on changed lines) before merging.

PR #1795, opened 2026-06-09
fix(composer): resolve composer QA bugs (#1781)

### LINKED ISSUE #1781: QA sweep: composer — 4 confirmed API bug(s) (live-verified on :dev)
Found by the live QA fan-out against the current-main `:dev` container and **verified real on current source** (HEAD 43b0e98f). 4 confirmed bug(s) in the `composer` subsystem.

### 1. [HIGH] Virtual repo packages.json returns empty {} instead of aggregating from members
**Repro/live evidence:** HTTP 200 but body is {"metadata-url":"/composer/vf-composer-83f56e63-virt/p2/%package%.json","packages":{}} — empty packages object even though the local member repo has 6 packages (testvendor/mypackage, testvendor/lib1, testvendor/lib2, testvendor/lib3, testvendor/myplugin, testvendor/notypepackage).
**Source:** `backend/src/api/handlers/composer.rs:378`
**Fix (small):** In packages_json, check if repo.repo_type == RepositoryType::Virtual and fan out: call proxy_helpers::fetch_virtual_members, then query artifacts for each member_id (like resolve_virtual_composer_metadata does), and aggregate results. Alternatively add a helper that collects all member repo_ids and runs a single IN ($ids) query.

### 2. [MEDIUM] Search type filter: total count ignores the type parameter
**Repro/live evidence:** type=library query returns results_count=4 but total=6; type=composer-plugin returns results_count=1 but total=6. The total always reflects unfiltered count (all 6 packages in the repo).
**Source:** `backend/src/api/handlers/composer.rs:800-810`
**Fix (small):** The total_count COUNT query at line 800 only filters by name ILIKE but not by type. Either move the type filter into SQL (JOIN artifact_metadata and filter on metadata->'composer'->>'type' = $3), or compute total_count as search_results.len() plus unfetched pages rather than from a separate query. Simplest: count total from search_results after type filtering (requires fetching all matching rows first, or a second SQL query that joins metadata and includes the type predicate).

### 3. [MEDIUM] Search pagination 'next' URL drops per_page and type query parameters
**Repro/live evidence:** GET /search.json?type=library&per_page=1&page=1 returns next="/composer/vf-composer-83f56e63-local/search.json?q=&page=2" — both type=library and per_page=1 are absent from the next URL, breaking paginated type-filtered searches.
**Source:** `backend/src/api/handlers/composer.rs:831-836`
**Fix (trivial):** Build the next URL by appending all non-default active parameters: include per_page when it was provided, include type/package_type when set. Change the format! to: format!("/composer/{}/search.json?q={}&page={}{}{}", repo_key, query_str, page+1, if per_page != 15 { format!("&per_page={}", per_page) } else { String::new() }, params.package_type.as_deref().map(|t| format!("&type={}", t)).unwrap_or_default())

### 4. [LOW] Null metadata fields serialized as JSON null instead of being omitted
**Repro/live evidence:** p2 metadata for a minimal package (only name+version in composer.json) contains 9 null fields: authors, autoload, description, homepage, keywords, license, require, require-dev, type all present as null. This violates Packagist/Composer spec which omits absent optional fields.
**Source:** `backend/src/formats/composer.rs:206-228`
**Fix (trivial):** Add #[serde(skip_serializing_if = "Option::is_none")] to each Option<T> field in ComposerJson (lines 208-227). This prevents None fields from being stored as null in artifact_metadata. Existing stored records will also need a migration or merge_composer_metadata must skip null values (change the condition in composer.rs:128 to skip Value::Null: if let Some(val) = composer.get(*key) { if !val.is_null() { version_entry[*key] = val.clone(); } }).

brandonrc:
## Summary

Closes #1781. Fixes the 4 live-verified composer QA bugs, all confined to the composer subsystem (`backend/src/api/handlers/composer.rs` + `backend/src/formats/composer.rs`).

1. **[HIGH] Virtual repo `packages.json` returned empty `{}`** — the root index handler queried only the virtual repo's own artifacts (it has none). Now, for `RepositoryType::Virtual`, it fans out across local/staging members (`fetch_virtual_members`), aggregates each member's artifacts via the new `fetch_package_index_rows`, and renders them under the virtual repo key (so dist URLs route back through us) via the new pure `build_packages_index` helper.
2. **[MEDIUM] Search `total` ignored the `type` parameter** — the count query filtered only by name ILIKE. Both the page query and the count query now apply the same `type` predicate in SQL (`am.metadata #>> '{composer,type}' = $type`, short-circuited with `$type::text IS NULL`), so `type=library` reports the filtered total and pagination math is correct.
3. **[MEDIUM] Search `next` URL dropped `per_page` and `type`** — extracted a pure `build_search_next_url` helper that appends `per_page` (when explicitly provided) and `type` (when set), so paginated, type-filtered searches keep working past page 1.
4. **[LOW] Null composer metadata fields serialized as JSON `null`** — added `#[serde(skip_serializing_if = "Option::is_none")]` to every optional `ComposerJson` field so absent fields are omitted on upload, and hardened `merge_composer_metadata` to skip `Value::Null` so records stored before this fix also render cleanly per the Packagist/Composer spec.

Tests added (each fails on main, passes here):
- `formats::composer::tests::test_composer_json_omits_none_fields_on_serialize` (bug 4 serialization)
- `composer::tests::test_merge_composer_metadata_skips_json_null` (bug 4 stored-record path)
- `composer::tests::test_build_packages_index_*` (bug 1 aggregation/grouping/dist-url, pure)
- `composer::tests::test_search_next_page_url_preserves_{per_page,type,per_page_and_type}` (bug 3, pure)
- `composer::metadata_db_tests::virtual_packages_json_aggregates_member_packages` (bug 1, DB-backed)
- `composer::metadata_db_tests::search_type_filter_constrains_total_count` (bugs 2 & 3, DB-backed)

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes (response shapes corrected, no new endpoints; no migrations)

Verified locally against a fresh DB: `cargo fmt`, `SQLX_OFFLINE=true cargo clippy --workspace --all-targets -- -D warnings`, and the composer tests (77 handler + 28 formats) all pass. `.sqlx` offline cache regenerated for the two changed search queries.

### DIFF
--- a/.sqlx/query-0b78b1df55a897dc9f8e28ddf35c97ba8981e294b196da98ff346f31ec4cb4b5.json
+++ b/.sqlx/query-0b78b1df55a897dc9f8e28ddf35c97ba8981e294b196da98ff346f31ec4cb4b5.json
@@ -1,0 +1,32 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT DISTINCT a.name,\n               am.metadata as \"metadata?\"\n        FROM artifacts a\n        LEFT JOIN artifact_metadata am ON am.artifact_id = a.id\n        WHERE a.repository_id = $1\n          AND a.is_deleted = false\n          AND a.name ILIKE $2\n          AND ($3::text IS NULL OR am.metadata #>> '{composer,type}' = $3)\n        ORDER BY a.name\n        LIMIT $4 OFFSET $5\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "name",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 1,
+        "name": "metadata?",
+        "type_info": "Jsonb"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid",
+        "Text",
+        "Text",
+        "Int8",
+        "Int8"
+      ]
+    },
+    "nullable": [
+      false,
+      false
+    ]
+  },
+  "hash": "0b78b1df55a897dc9f8e28ddf35c97ba8981e294b196da98ff346f31ec4cb4b5"
+}
--- a/.sqlx/query-c1ac096c425e915cf48ef5d5ac7f2c8238403fe5e4672585cfb6b2a7987f9b4d.json
+++ b/.sqlx/query-c1ac096c425e915cf48ef5d5ac7f2c8238403fe5e4672585cfb6b2a7987f9b4d.json
@@ -1,0 +1,24 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT COUNT(DISTINCT a.name)\n        FROM artifacts a\n        LEFT JOIN artifact_metadata am ON am.artifact_id = a.id\n        WHERE a.repository_id = $1\n          AND a.is_deleted = false\n          AND a.name ILIKE $2\n          AND ($3::text IS NULL OR am.metadata #>> '{composer,type}' = $3)\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "count",
+        "type_info": "Int8"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid",
+        "Text",
+        "Text"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "c1ac096c425e915cf48ef5d5ac7f2c8238403fe5e4672585cfb6b2a7987f9b4d"
+}
--- a/backend/src/api/handlers/composer.rs
+++ b/backend/src/api/handlers/composer.rs
@@ -126,7 +126,14 @@ fn merge_composer_metadata(
 
     for key in COMPOSER_METADATA_KEYS {
         if let Some(val) = composer.get(*key) {
-            version_entry[*key] = val.clone();
+            // Skip JSON null so absent optional fields are omitted from the
+            // version entry rather than serialized as `"field": null`. This
+            // matters for records stored before ComposerJson gained
+            // `skip_serializing_if`, whose metadata still carries null fields
+            // (#1781).
+            if !val.is_null() {
+                version_entry[*key] = val.clone();
+            }
         }
     }
 }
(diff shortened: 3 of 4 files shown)


brandonrc:
## Review: APPROVE (pending CI green)

Traced all four fixes against the source rather than the description. They are correct, minimal, and confined to `composer.rs` + `formats/composer.rs` (+ regenerated `.sqlx` cache). No scope creep, no broken contracts.

### Bug-by-bug verification
- **Bug 1 — virtual `packages.json` empty `{}`** ✅ The `RepositoryType::Virtual` branch now fans out over Local/Staging members via `fetch_virtual_members` and aggregates with `fetch_package_index_rows`, rendering under the *virtual* repo key so dist URLs route back through us. Mirrors the existing `resolve_virtual_composer_metadata` pattern — consistent. Remote members correctly excluded (resolved per-package via metadata-url). `fetch_package_index_rows` reuses the existing cached query (`72ac0d36…`).
- **Bug 2 — search `total` ignored `type`** ✅ Both the page query and the count query now apply the same `am.metadata #>> '{composer,type}' = $3` predicate, short-circuited by `$3::text IS NULL`. JSON path matches how metadata is stored (`{"composer": {"type": …}}`, via `#[serde(rename="type")]`). `$3` is a bound parameter — no injection.
- **Bug 3 — `next` URL dropped `per_page`/`type`** ✅ Extracted pure `build_search_next_url`; appends `per_page` only when explicitly sent and `type` when set. Param ordering (per_page before type) is asserted.
- **Bug 4 — null metadata fields** ✅ Two-pronged and correct: `skip_serializing_if = "Option::is_none"` stops new uploads writing nulls, and the `!val.is_null()` guard in `merge_composer_metadata` keeps pre-fix stored records rendering cleanly. Good that both the write path and the legacy read path are covered.

### Regression tests
Each bug has a test that fails on main: serialization (`len()==2`), null-skip merge, `build_packages_index` (pure), `build_search_next_url` (pure), plus DB-backed `virtual_packages_json_aggregates_member_packages` and `search_type_filter_constrains_total_count` (asserts `total==1`, would be 4/6 on main). Solid.

### Offline sqlx cache
Verified both new `.sqlx` filenames are the exact SHA-256 of the new query strings, and the reused index query is already cached — `Check Rust` (`SQLX_OFFLINE=true`) should compile.

### Minor (non-blocking) notes
- 🟡 **Bug 2's total-count fix is only covered by the DB-backed test**, which no-ops without Postgres — so it's untested in the Tier‑1 unit run. Consider a thin pure helper around the count→pages math (or rely on the integration tier; acceptable).
- 💭 `build_search_next_url` does not URL-encode `q`/`type` — but this matches the pre-existing behavior it replaced, so not a regression.
- 💭 Old inline search-query `.sqlx` entries are now orphaned (harmless).

### CI
`Check Rust`, `Backend Unit Tests`, `Code Coverage` still **pending** (not failed); CodeQL/audit/shell green. Approving on the merits — hold the merge until the required checks go green.

Issue #817, opened 2026-04-21
Enforce permission rules in repo_visibility_middleware


brandonrc:
## Context

Part of the permissions enforcement plan from #794. This is Phase 2.
Depends on #816 (PermissionService).

## Problem

The `repo_visibility_middleware` checks whether a repo is public/private and enforces API token repo-scope restrictions, but it does not consult the permissions table. Users with permission rules granting repository access are not actually granted that access.

## Requirements

After the existing checks in `repo_visibility_middleware` pass:

1. If the user is admin, skip permission checks (backward compatible).
2. If no permission rules exist for this repository (no rows with target_type='repository' AND target_id=repo.id), allow access. This preserves backward compatibility for repos without explicit permission rules.
3. If permission rules exist for this repository, call `PermissionService::check_permission()` with the appropriate action:
   - GET/HEAD: requires "read"
   - PUT/POST: requires "write"
   - DELETE: requires "delete"
4. Return 403 if the permission check fails.

The "no rules means open access" default is critical. Locking down a repo requires explicitly creating at least one permission rule for it.

## Design Reference

See `docs/plans/2026-04-17-permissions-enforcement-design.md`, Phase 2 section.

## Test Plan

- Unit tests for the middleware with mocked PermissionService.
- Integration tests: create a repo, add permission rules, verify access is restricted.
- Backward compatibility test: repos without rules continue to work unchanged.

brandonrc:
Implemented in #826. repo_visibility_middleware now checks permissions when rules exist for a repository, with fail-closed on DB errors, admin action implying all actions, and backward-compatible fallthrough when no rules are defined.

PR #1809, opened 2026-06-09
fix(security): native middleware default-allowed rule-less private repos (#1802)

### LINKED ISSUE #1802: Security: native-protocol middleware default-allows rule-less private repos (cross-team artifact leak)
**Severity:** critical — broken access control (found by internal red-team, reproduced live).

## Summary
`repo_visibility_middleware` treated a **private** repo that has **no fine-grained permission rules** as readable by **any authenticated user**: `should_allow_repo_access(is_public=false, has_auth=true)` returns true, and the `#817` permission block is gated on `has_any_rules_for_target(...)`, so when no rules exist the block is skipped and the request falls through to the format handler. The REST download path (`require_visible` → `user_can_access_repo`) denies the same caller with 404 — so the native protocol paths (`/maven`, `/cargo`, `/helm`, `/pypi`, …) were strictly more permissive than REST for the same private repo.

## Reproduction (live)
A brand-new, zero-grant, non-admin user:
```
# NATIVE maven path -> 200, leaks the private crown-jewel secret
GET /maven/acme-internal-core/com/acme/internal/core-service/3.1.0/application.properties  -> 200 (Pr0d_DB_S3cr3t...)
# REST path, SAME user, SAME artifact -> correctly denied
GET /api/v1/repositories/acme-internal-core/download/.../application.properties            -> 404
```

## Root cause
`backend/src/api/middleware/auth.rs` `repo_visibility_middleware`: the non-admin branch only enforced fine-grained rules when `has_any_rules_for_target` was true; the `has_rules == false` path default-allowed.

## Fix
PR #PLACEHOLDER: for a non-admin on a private repo with no fine-grained rules, require a role assignment (mirror `require_visible`) before falling through; existence-hiding 404 otherwise. Public repos, admins, and rule-governed repos unaffected. DB-backed regression test added.

brandonrc:
## Summary

Fixes a critical broken-access-control hole found by the internal red-team and reproduced live: the native-protocol `repo_visibility_middleware` treated a **private** repository with **no fine-grained permission rules** as readable by **any authenticated user**.

`should_allow_repo_access(is_public=false, has_auth=true)` returns `true`, and the `#817` fine-grained block is gated on `has_any_rules_for_target(...)`. When a private repo has no rules, that block was skipped and the request fell through to the format handler — so `/maven`, `/cargo`, `/helm`, `/pypi`, … were strictly more permissive than the REST download path (`require_visible` → `user_can_access_repo`), which denied the same caller with 404.

A brand-new, zero-grant, non-admin user could read private cross-team artifacts (including a credentials file): native maven path returned `200` + the secret, while REST returned `404` for the same user + artifact.

### Fix
For a non-admin on a **private** repo with **no fine-grained rules**, require a role assignment (mirroring `require_visible`) before falling through; deny with an existence-hiding `404` otherwise. Public repos, admins, and repos governed by fine-grained rules are unaffected. Uses `sqlx::query_scalar` (runtime, no offline-cache entry), same predicate as `RepositoryService::user_can_access_repo`.

Closes #1802.

## Test Checklist
- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings` (changed crate, clean)
- [x] `cargo test --workspace --lib` (auth middleware module: 185 passed)
- [x] New DB-backed regression test: non-admin without a role assignment → `404`; granting the role → `200`. Runs in the CI coverage job (seeds Postgres).

## API Changes
None. Behavior change only: native-protocol reads of a rule-less private repo by a non-admin without a role assignment now return `404` (existence-hiding), matching the REST download path. This is a security tightening, not an API surface change.

### DIFF
--- a/.sqlx/query-0b78b1df55a897dc9f8e28ddf35c97ba8981e294b196da98ff346f31ec4cb4b5.json
+++ b/.sqlx/query-0b78b1df55a897dc9f8e28ddf35c97ba8981e294b196da98ff346f31ec4cb4b5.json
@@ -1,0 +1,32 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT DISTINCT a.name,\n               am.metadata as \"metadata?\"\n        FROM artifacts a\n        LEFT JOIN artifact_metadata am ON am.artifact_id = a.id\n        WHERE a.repository_id = $1\n          AND a.is_deleted = false\n          AND a.name ILIKE $2\n          AND ($3::text IS NULL OR am.metadata #>> '{composer,type}' = $3)\n        ORDER BY a.name\n        LIMIT $4 OFFSET $5\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "name",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 1,
+        "name": "metadata?",
+        "type_info": "Jsonb"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid",
+        "Text",
+        "Text",
+        "Int8",
+        "Int8"
+      ]
+    },
+    "nullable": [
+      false,
+      false
+    ]
+  },
+  "hash": "0b78b1df55a897dc9f8e28ddf35c97ba8981e294b196da98ff346f31ec4cb4b5"
+}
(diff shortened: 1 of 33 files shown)


brandonrc:
Superseded by a clean single-file PR. This branch was accidentally cut from a local main carrying ~16 unmerged QA-fix commits, making it a 4,352-line/33-file diff instead of the intended ~60-line auth.rs change (and inheriting an unrelated #1786 repo_tokens test break). Replaced by #PLACEHOLDER with only the auth.rs fix + regression test.

PR #1816, opened 2026-06-09
fix(security): enforce per-member authz on virtual repo downloads (#1804)

### LINKED ISSUE #1804: Security: public virtual repo serves PRIVATE member-repo artifacts to anonymous/zero-grant users
**Severity:** high — confused-deputy / IDOR via virtual aggregation (verified anonymous + fresh zero-grant non-admin)

## Summary
Public Maven virtual repo leaks private member-repo artifacts to anonymous and zero-privilege users (broken access control / IDOR via virtual aggregation)

## Reproduction
```
Ground truth: maven-virtual is_public=t; its members include the PRIVATE local repo acme-backend-internal (is_public=f). Private artifact path com/acme/platform/acme-core/2.1.0/acme-core-2.1.0.jar exists ONLY in acme-backend-internal.

CONTROL (direct access to private member is correctly denied):
  # anonymous
  $ curl -s -o /dev/null -w "%{http_code}\n" \
     "http://localhost:8080/maven/acme-backend-internal/com/acme/platform/acme-core/2.1.0/acme-core-2.1.0.jar"
  401
  # fresh zero-grant non-admin user vleak1 (verified is_admin=f, 0 permission grants)
  $ curl -s -o /dev/null -w "%{http_code}\n" -H "Authorization: Bearer $UJWT" \
     "http://localhost:8080/maven/acme-backend-internal/.../acme-core-2.1.0.jar"
  403

EXPLOIT (same private bytes leak through the PUBLIC virtual repo):
  # anonymous -- NO credentials at all
  $ curl -s -o /tmp/leak.bin -w "HTTP %{http_code} size=%{size_download}\n" \
     "http://localhost:8080/maven/maven-virtual/com/acme/platform/acme-core/2.1.0/acme-core-2.1.0.jar"
  HTTP 200 size=300
  $ sha256sum /tmp/leak.bin
  762827d604cc593935a72db075cffc8d68366c36bce8c5732dcc63757fdccafe
  # matches DB checksum_sha256 of the PRIVATE artifact exactly:
  $ docker exec ak-rt-db psql -U registry -d artifact_registry -tAc \
     "select checksum_sha256 from artifacts a join repositories r on r.id=a.repository_id \
      where r.key='acme-backend-internal' and a.path='com/acme/platform/acme-core/2.1.0/acme-core-2.1.0.jar';"
  762827d604cc593935a72db075cffc8d68366c36bce8c5732dcc63757fdccafe
  (body is the real JAR: PK zip header, MANIFEST "Created-By: Acme Build System 2.1.0")

  # fresh zero-grant non-admin vleak1 -- same leak
  $ curl -s -o /tmp/u.bin -w "HTTP %{http_code}\n" -H "Authorization: Bearer $UJWT" \
     "http://localhost:8080/maven/maven-virtual/.../acme-core-2.1.0.jar"
  HTTP 200  (sha256 == 762827d6... identical private bytes)

Reproduced on a SECOND private artifact (acme-auth-1.5.2.jar, present only in acme-backend-internal): anon direct=401, anon via maven-virtual=200/312 bytes, leaked sha256 2c3fdc79... == DB sha256. Token-privilege trap avoided: no ext-contractor/admin token used; principals were anonymous and vleak1 (DB-verified is_admin=f with zero rows in permission_grants).
```

## Control
Direct GET /maven/acme-backend-internal/<path> as the SAME principals returns 401 (anonymous) / 403 (vleak1), proving those callers must NOT read the private member. The only difference in the exploit request is the repo_key segment (maven-virtual vs acme-backend-internal); the served bytes are byte-identical (sha256 match) to the private artifact, so aggregation is the sole bypass vector.

## Root cause
Access control is enforced only on the URL-named repo, never on the virtual repo's members.
- backend/src/api/middleware/auth.rs:1195 repo_visibility_middleware resolves repo via extract_repo_key(&path) -> "maven-virtual", then decides everything (visibility at :1356 should_allow_repo_access(is_public=true,..), API-token scope at :1364 ext.can_access_repo(repo.id), and fine-grained rules at :1382 has_any_rules_for_target("repository", repo.id)) using ONLY the virtual repo's id/is_public. Members are never consulted.
- backend/src/api/handlers/maven.rs:683 download() takes no auth principal and performs no permission check. For a Virtual repo it calls proxy_helpers::resolve_virtual_download (maven.rs:1341) over fetch_virtual_members.
- backend/src/api/handlers/proxy_helpers.rs:1310 fetch_virtual_members SELECTs ALL members joined on virtual_repo_members with NO is_public/permission filter, and the download loop serves member bytes (local_fetch_by_path -> storage) with no per-member visibility/permission re-check. Net: a public virtual repo becomes a confused deputy that reads its private members on behalf of anyone allowed to read the virtual.

## Suggested fix
Enforce authorization on every member actually served, not just the URL repo. Concretely: when the resolved repo is Virtual, for each member that produces the response, re-run the same checks the middleware applies to a direct repo — member.is_public OR (authenticated AND can_access_repo(member.id) AND, if has_any_rules_for_target("repository", member.id), holds the read action; admins exempt). Skip/deny members the caller could not read directly (return as if not found so existence isn't leaked). Implement either by passing the AuthExtension into download()/resolve_virtual_download and filtering fetch_virtual_members results through PermissionService before fetching bytes, or by adding a guard in proxy_helpers::resolve_virtual_download that takes the principal and a can_read(member) predicate. Add a regression test: anonymous + zero-grant non-admin GET of a private member's artifact via a public virtual repo must return 404/403, while a member-granted user still succeeds.

brandonrc:
## Summary

A **public** Maven virtual repo leaked **private** member-repo artifacts to anonymous and zero-privilege callers — broken access control / IDOR via virtual aggregation (#1804, HIGH).

The visibility middleware authorizes only the URL-named repo. A public virtual repo therefore became a *confused deputy*: `proxy_helpers::resolve_virtual_download` / `fetch_virtual_members` (and the per-member checksum / stored-file probes) served member content with **no per-member visibility or permission check**, so anyone allowed to read the virtual read its private members' bytes.

Verified leak: anonymous `GET /maven/maven-virtual/.../acme-core-2.1.0.jar` returned `200` with bytes byte-identical (sha256 match) to the private artifact in `acme-backend-internal`, while a direct `GET /maven/acme-backend-internal/...` correctly returned `401`/`403`.

**Fix:** when the resolved repo is `Virtual`, re-run the same read-access model the middleware applies to a direct repo against **every member that would serve the response**:

- `member.is_public` -> readable by anyone (incl. anonymous), else
- caller must be authenticated, and either an **admin**, or pass the API-token repo scope (`can_access_repo`) **and**, if fine-grained rules exist for the member, hold the `read` (or `admin`) action.

Members the caller could not read directly are **dropped**, so a denied member behaves as if it did not contain the artifact (`404`) — existence is not leaked.

New helpers `caller_can_read_member` / `authorize_virtual_members` live in `proxy_helpers`, plus a `resolve_virtual_download_from_members` variant that operates over an already-authorized member list (internal delegation; no caller-signature churn). The maven `download`/`serve_artifact` path threads the `AuthExtension` through and filters members before fetching bytes or checksums.

`test_caller_can_read_member_blocks_private_member_1804` (DB-backed, in `proxy_helpers`) proves:
- anonymous + zero-grant non-admin are **denied** a private member,
- `authorize_virtual_members` keeps **only** public members for an anonymous caller,
- a member-granted user (and admins) still succeed.

Verified it **fails** when the per-member check is bypassed (helper stubbed to always-allow -> `anonymous must NOT read a private member`) and **passes** with the fix.

## Test Checklist
- [x] Manually tested locally (clippy -D warnings clean; targeted test passes)

## API Changes
- [x] N/A - no API changes

Closes #1804

### DIFF
--- a/backend/src/api/handlers/maven.rs
+++ b/backend/src/api/handlers/maven.rs
@@ -682,6 +682,7 @@ fn content_type_for_path(path: &str) -> &'static str {
 
 async fn download(
     State(state): State<SharedState>,
+    Extension(auth): Extension<Option<AuthExtension>>,
     Path((repo_key, path)): Path<(String, String)>,
 ) -> Result<Response, Response> {
     let repo = resolve_maven_repo(&state.db, &repo_key).await?;
@@ -723,7 +724,17 @@ async fn download(
             // generation below only queries `repo.id` (which has no
             // artifact rows for a virtual). #1444.
             if repo.repo_type == RepositoryType::Virtual {
+                // #1804: gate the per-member stored-file probes so a public
+                // virtual repo cannot serve a PRIVATE member's stored checksum
+                // or metadata file to a caller who could not read that member
+                // directly.
                 let members = proxy_helpers::fetch_virtual_members(&state.db, repo.id).await?;
+                let members = proxy_helpers::authorize_virtual_members(
+                    &state.permission_service,
+                    auth.as_ref(),
+                    members,
+                )
+                .await;
                 for member in &members {
                     if let Ok(member_storage) = state.storage_for_repo(&member.storage_location()) {
                         if let Ok(content) = member_storage.get(&checksum_storage_key).await {
@@ -1042,7 +1053,17 @@ async fn download(
 
         // Virtual repo: try each member in priority order
         if repo.repo_type == RepositoryType::Virtual {
+            // #1804: only members the caller could read directly may serve a
+            // checksum. A private member's checksum reveals the existence and
+            // exact content hash of its artifact, so it must be gated the same
+            // way the artifact bytes are.
             let members = proxy_helpers::fetch_virtual_members(&state.db, repo.id).await?;
+            let members = proxy_helpers::authorize_virtual_members(
+                &state.permission_service,
+                auth.as_ref(),
+                members,
+            )
+            .await;
 
             for member in &members {
                 if member.repo_type == RepositoryType::Remote {
@@ -1089,7 +1110,7 @@ async fn download(
     }
 
     // 4. Serve the artifact file
-    serve_artifact(&state, &repo, &repo_key, &path).await
+    serve_artifact(&state, &repo, &repo_key, &path, auth.as_ref()).await
 }
 
 async fn generate_metadata_for_artifact(
@@ -1140,6 +1161,7 @@ async fn serve_artifact(
     repo: &RepoInfo,
     repo_key: &str,
     path: &str,
+    auth: Option<&AuthExtension>,
 ) -> Result<Response, Response> {
     let artifact = sqlx::query!(
         r#"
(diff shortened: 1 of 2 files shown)


brandonrc:
Need more checks on this to make sure we are not breaking other repo types

PR #1848, opened 2026-06-11
fix(incus): honour HTTP Range on image download (resumable large-image downloads)

### LINKED ISSUE #1847: incus image download ignores HTTP Range — no resumable downloads for large images
### Description

The incus image download handler `download_image` (`backend/src/api/handlers/incus.rs:720-802`), serving `GET /incus/<repo>/images/<product>/<version>/<filename>`, ignores the request's `Range` header entirely. It takes no header input, streams the full object via `get_stream`, and unconditionally returns `200 OK` with the full `Content-Length` and no `Accept-Ranges` header. A client whose connection drops partway through a multi-GiB image (e.g. behind a load balancer or gateway with a response/stream timeout) cannot resume — every retry restarts at byte 0.

### Inconsistent within the same backend

The generic artifact handler `download_artifact` (`backend/src/api/handlers/repositories.rs`) **is** range-aware as of #1785: it parses `Range` (`parse_byte_range`, `repositories.rs:3676-3748`), returns `206 Partial Content` with `Content-Range`, advertises `Accept-Ranges: bytes`, and handles `416` (`:3946-3949`, `:3963/3973/3987`, slice at `:3754-3788`). That range support was not carried over to the incus handler. A third path, `serve_local_artifact` → `build_download_response` (`proxy_helpers.rs:2612-2667`), buffers the whole object into memory and is also not range-aware. So the backend has three divergent download paths and no shared range-aware streaming primitive.

### Reproduction

A `Range: bytes=0-1023` GET against a published multi-GiB image in an incus-format repo returns `200 OK` with the full multi-GiB `Content-Length` and no `Content-Range`/`Accept-Ranges`, instead of `206 Partial Content`.

### Proposed fix

Extract a shared range-aware streaming helper (parse `Range` → `206`/`416`/`200`, advertise `Accept-Ranges: bytes`) and route both `download_image` and `download_artifact`'s streaming branch through it. The storage trait already exposes `get_range(key, offset, length)` (`backend/src/storage/mod.rs:190-226`, implemented for GCS and S3), so the helper can read only the requested bytes from object storage. Net effect: incus (and any other format) gains resumable downloads, and the duplicated body-serving logic collapses to one primitive. Clients that ignore `Accept-Ranges` keep getting a full `200`, so it's backward-compatible.

### Regression test

A ranged GET against an incus image returns `206` with the correct `Content-Range` and sliced body, and `Accept-Ranges: bytes` on the `200` path (fails on `main`, passes with the fix).

dragonpaw:
## Summary

The incus image download handler (`download_image`,
`backend/src/api/handlers/incus.rs`) ignored the request's HTTP `Range` header and
always returned a full `200 OK` with no `Accept-Ranges`, so a client whose
connection dropped partway through a multi-GiB image (behind a load balancer /
gateway timeout) could never resume — every retry restarted at byte 0. The
generic artifact download (`download_artifact`) gained full RFC 7233 range support
in #1785, but it was never carried over to the incus handler.

This extracts a shared `ranged_stream_response` helper and routes **both**
`download_image` and `download_artifact`'s streaming branch through it:

- `200 OK` now advertises `Accept-Ranges: bytes`,
- a satisfiable range returns `206 Partial Content` + `Content-Range` + the sliced body,
- an unsatisfiable range returns `416 Range Not Satisfiable` + `Content-Range: bytes */<total>`.

The previously-duplicated 200/206/416 block in `download_artifact` is collapsed
into the helper.

`tests/incus_upload_tests::test_incus_download_honours_range` (DB-backed,
`--ignored`): against a staged incus image, a `Range: bytes=0-1023` GET returns
`206` with `Content-Range: bytes 0-1023/4096` and exactly the first 1024 bytes; a
no-range GET returns `200` with `Accept-Ranges: bytes`; an unsatisfiable range
returns `416` with `Content-Range: bytes */4096`. Verified **failing on `main`**
(the `200` path lacked `Accept-Ranges`) and **passing** with this change.

## Test Checklist
- [x] Manually tested locally (local Postgres; also confirmed pre-fix against a deployed instance: a `Range: bytes=0-1023` GET of a multi-GiB image returned `200` with the full `Content-Length`)
- [x] No regressions in existing tests (range unit tests + existing incus upload tests pass; clippy clean)

## API Changes
- [x] N/A — no API changes (adds `Range` handling + `Accept-Ranges` to an existing endpoint; full `200` responses are unchanged for clients that don't send `Range`)

Closes #1847

### DIFF
--- a/backend/src/api/handlers/incus.rs
+++ b/backend/src/api/handlers/incus.rs
@@ -22,7 +22,7 @@
 
 use axum::body::Body;
 use axum::extract::{DefaultBodyLimit, OriginalUri, Path as AxumPath, State};
-use axum::http::header::{CONTENT_LENGTH, CONTENT_TYPE};
+use axum::http::header::CONTENT_TYPE;
 use axum::http::{HeaderMap, StatusCode, Uri};
 use axum::response::{IntoResponse, Response};
 use axum::routing::{get, patch, post};
@@ -719,6 +719,7 @@ async fn streams_images(
 
 async fn download_image(
     State(state): State<SharedState>,
+    headers: HeaderMap,
     AxumPath((repo_key, product, version, filename)): AxumPath<(String, String, String, String)>,
 ) -> Result<Response, Response> {
     let repo = resolve_incus_repo(&state.db, &repo_key).await?;
@@ -784,21 +785,34 @@ async fn download_image(
         }
     })?;
 
-    let content_type = content_type_for_download(&filename);
-    let body =
-        Body::from_stream(stream.map(|r| r.map_err(|e| std::io::Error::other(e.to_string()))));
-
-    Ok(Response::builder()
-        .status(StatusCode::OK)
-        .header(CONTENT_TYPE, content_type)
-        .header(CONTENT_LENGTH, size_bytes.to_string())
-        .header(
-            "Content-Disposition",
+    // Honour HTTP `Range` so large images are resumable, via the shared
+    // range-aware streaming helper that the generic artifact download uses
+    // (#1847). Previously this handler ignored `Range` and always returned a
+    // full `200`, so a dropped multi-GiB transfer could never resume.
+    let range_header = headers
+        .get(axum::http::header::RANGE)
+        .and_then(|v| v.to_str().ok());
+    let base_headers = vec![
+        (
+            CONTENT_TYPE,
+            content_type_for_download(&filename).to_string(),
+        ),
+        (
+            axum::http::header::CONTENT_DISPOSITION,
             format!("attachment; filename=\"{}\"", filename),
-        )
-        .header("X-Checksum-Sha256", checksum)
-        .body(body)
-        .unwrap())
+        ),
+        (
+            axum::http::header::HeaderName::from_static("x-checksum-sha256"),
+            checksum,
+        ),
+    ];
+    crate::api::handlers::repositories::ranged_stream_response(
+        range_header,
+        size_bytes.max(0) as u64,
+        stream,
+        base_headers,
+    )
+    .map_err(|e| e.into_response())
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


brandonrc (approved):
**Approve** — reviewed the diff and verified the bug reproduces on `main` (7f191b0).

**Bug confirmed:** `download_image` ignores the HTTP `Range` request header. On main its signature has no `HeaderMap`/`Range` param (`incus.rs:720-723`) and it unconditionally returns `200 OK` with full `Content-Length` and no `Accept-Ranges`/`Content-Range` (`incus.rs:787-801`), so a `Range: bytes=0-1023` gets the whole object from byte 0 and a dropped large-image transfer can never resume. The sibling `download_artifact` path has been RFC 7233-aware since #1785 (`repositories.rs:4035-4088`), so Incus is the inconsistent outlier.

**Fix is correct:** adds `headers: HeaderMap` to `download_image`, and factors the shared `ranged_stream_response(range_header, total, body, base_headers)` helper out of `download_artifact`, routing both through it. Reuses the existing `parse_byte_range`/`slice_byte_stream`, returns `206` + `Content-Range: bytes start-end/total` (sliced, streamed body), `416` + `Content-Range: bytes */total` for unsatisfiable, and `200` + `Accept-Ranges: bytes` otherwise. Edge cases handled by the reused parser: open-ended `bytes=N-`, suffix `bytes=-N`, multi-range → degrade to 200, malformed → 200, `total==0` → 200. It streams (`Body::from_stream`), never buffers the image, and the storage stream types line up. A regression test is included (`tests/incus_upload_tests::test_incus_download_honours_range`) and CI is green.

**Non-blocking note:** like the existing `download_artifact`, this slices *after* a full `get_stream` (`slice_byte_stream` consumes and discards), so object storage still fetches the whole image per request — functionally correct and resumable, but no storage-side bandwidth/latency win. The issue suggested wiring `get_range(key, offset, length)` (`storage/mod.rs:190`); that can be a follow-up. (Also: the regression test is `#[ignore]`/DB-backed, so only the parser unit tests run in the fast gate.)

Nice consolidation into a shared helper. 👍

PR #1865, opened 2026-06-11
fix: replicate hosted PyPI packages to peers

### LINKED ISSUE #1864: Hosted PyPI uploads are not queued for peer replication, so HA peers never receive packages
### Description

fixed by #1865 

Hosted PyPI repositories can accept valid wheel/sdist uploads, persist the
artifact, store PyPI metadata, populate the package catalog, and serve the
package through the Simple API on the source instance. However, uploads through
the PyPI-specific hosted endpoint are not queued for peer replication.

In an HA deployment with two Artifact Keeper peers and mirror-style repository
replication enabled for the same hosted PyPI repository, a package uploaded to
the source peer remains only on that peer. The target peer never receives an
artifact row, PyPI metadata row, package/package-version rows, or blob because
the source peer never creates a `sync_tasks` row for the uploaded PyPI artifact.

### Environment Information

- Deployment type: Kubernetes, Argo CD-managed Helm chart.
- Operating System: Linux containers on Kubernetes nodes.
- Architecture: Linux/amd64 backend containers
- Database: PostgreSQL 17.
- Storage backend: S3-compatible object storage for repository blobs.
- Reverse Proxy: Kubernetes Ingress / NGINX.
- Network setup: Two independent Artifact Keeper instances configured as peers,
  with mirror-style peer repository replication enabled for the same hosted PyPI
  repositories on both peers. Private endpoint, bucket, and credential details
  are redacted.

### Backend Version

main branch

### Web UI Version

1.1.3

### Package Format

PyPI

### Expected Behavior

When a hosted PyPI package is uploaded to a repository that has enabled push or
mirror peer replication:

- The source instance should persist the wheel/sdist artifact, PyPI
  `artifact_metadata`, package row, and package-version row.
- The upload should enqueue peer `sync_tasks` for eligible non-local peers, the
  same way generic artifact uploads and fixed Debian/APT uploads do.
- The sync worker should replicate the artifact to the target peer through the
  configured storage abstraction.
- Replication should preserve PyPI-specific metadata needed by the Simple API,
  PEP 658 `.metadata` sidecar endpoint, and package catalog.
- After replication settles, the target peer should expose the same PyPI package
  through:
  - `GET /pypi/{repo}/simple/{project}/`
  - `GET /pypi/{repo}/simple/{project}/{filename}`
  - `GET /api/v1/packages?repository_key={repo}&format=pypi&search={project}`
  - PostgreSQL `artifacts`, `artifact_metadata`, `packages`, and
    `package_versions`
  - native `pip install` / `pip download` flows using the target peer as the
    index URL.

### Observed Behavior

The PyPI upload succeeds on the source peer:

- `POST /pypi/{repo}/` returns HTTP `200` with body `OK`.
- The source peer Simple API returns HTTP `200` for the uploaded project.
- The source peer Simple API page contains the uploaded wheel filename.
- The source peer database has one active artifact row for the uploaded
  `{normalized}/{version}/{filename}` path.
- The source peer database has an enabled mirror subscription for the PyPI repo
  to the target peer.
- The source peer backend logs a successful `PyPI upload: ...` message.

But no replication is attempted:

- The source peer database has `0` `sync_tasks` rows for the uploaded PyPI
  artifact.
- The target peer Simple API continues returning HTTP `404` for the same
  project.
- The target peer database has no artifact row for the uploaded PyPI artifact.
- The target peer database has no `sync_tasks` rows for the uploaded artifact.
- Source backend logs contain the PyPI upload and source Simple API checks, but
  no `Queued sync tasks` or sync-worker transfer attempt for the uploaded PyPI
  package.

Sanitized live repro summary from 2026-06-10:

```text
repository=test-pypi
repository_format=pypi
repository_type=local
source_to_target_subscription=sync_enabled=true, replication_mode=mirror
target_to_source_subscription=sync_enabled=true, replication_mode=mirror

package=ak-pypi-repl-probe-20260610t103514z
version=0.0.20260610103514
filename=ak_pypi_repl_probe_20260610t103514z-0.0.20260610103514-py3-none-any.whl
artifact_path=ak-pypi-repl-probe-20260610t103514z/0.0.20260610103514/ak_pypi_repl_probe_20260610t103514z-0.0.20260610103514-py3-none-any.whl
size=1437
sha256=992f904bbaf5b3bf575ed3e1eada3b58cb568e691b0b9b4591865f539a482954

upload_http=200
upload_body=OK
source_simple_status=200
source_simple_contains_filename=true
source_db_artifacts=1
source_db_sync_tasks=0

target_simple_status=404
target_simple_body={"code":"NOT_FOUND","message":"Package not found"}
target_db_artifacts=0
target_db_sync_tasks=0
wait_window_seconds=180
polls=12
```

Evidence was collected under a local run directory and can be provided if
needed. Important evidence captured during the repro:

```text
source upload response body: OK
source Simple API: project page exists and contains the wheel link
target Simple API: repeated 404 Package not found during the polling window
source DB: artifact row exists, sync_tasks rows for that artifact = 0
target DB: artifact rows for that path = 0
source peer subscription: enabled mirror replication for the PyPI repository
```

### Steps to Reproduce

1. Deploy two Artifact Keeper backend instances from current `main` with
   PostgreSQL and S3-backed repository storage.

2. Configure the two instances as peers.

3. Create the same hosted PyPI repository on both peers, for example:

   ```text
   key: PyPIReplicationSmoke
   format: pypi
   type: local/hosted
   ```

4. Enable peer repository replication for that repository in push or mirror
   mode. Verify the source-side subscription is enabled:

   ```sql
   select r.key,
          r.format::text,
          r.repo_type::text,
          pi.name,
          pi.endpoint_url,
          pi.status::text,
          prs.sync_enabled::text,
          prs.replication_mode::text
   from repositories r
   join peer_repo_subscriptions prs on prs.repository_id = r.id
   join peer_instances pi on pi.id = prs.peer_instance_id
   where r.key = 'PyPIReplicationSmoke';
   ```

   Expected precondition: at least one non-local peer row has
   `sync_enabled=true` and `replication_mode` of `push` or `mirror`.

5. Build a minimal valid wheel:

   ```shell
   python3 -m venv /tmp/ak-pypi-repro-venv
   . /tmp/ak-pypi-repro-venv/bin/activate
   python -m pip install build

   mkdir -p /tmp/ak-pypi-repro/src/ak_pypi_replication_smoke
   cd /tmp/ak-pypi-repro
   cat > pyproject.toml <<'EOF'
   [build-system]
   requires = ["setuptools>=68", "wheel"]
   build-backend = "setuptools.build_meta"

   [project]
   name = "ak-pypi-replication-smoke"
   version = "0.1.0"
   description = "Artifact Keeper PyPI peer replication repro package"
   requires-python = ">=3.8"
   EOF
   printf '__version__ = "0.1.0"\n' > src/ak_pypi_replication_smoke/__init__.py
   python -m build --wheel
   ```

6. Upload the wheel to the source peer through the PyPI hosted endpoint:

   ```shell
   python -m pip install twine
   TWINE_USERNAME='<user>' TWINE_PASSWORD='<token>' \
     twine upload \
       --non-interactive \
       --repository-url 'https://<source-peer>/pypi/PyPIReplicationSmoke/' \
       dist/ak_pypi_replication_smoke-0.1.0-py3-none-any.whl
   ```

   Expected response: successful upload.

7. Verify the source peer serves the project:

   ```shell
   curl -u '<user>:<token>' \
     'https://<source-peer>/pypi/PyPIReplicationSmoke/simple/ak-pypi-replication-smoke/'
   ```

8. Check whether the source peer queued replication for the uploaded artifact:

   ```sql
   with target_artifact as (
       select a.id
       from artifacts a
       join repositories r on r.id = a.repository_id
       where r.key = 'PyPIReplicationSmoke'
         and a.path = 'ak-pypi-replication-smoke/0.1.0/ak_pypi_replication_smoke-0.1.0-py3-none-any.whl'
         and a.is_deleted = false
   )
   select st.status::text,
          st.task_type,
          st.retry_count,
          st.error_message,
          st.created_at,
          st.started_at,
          st.completed_at
   from sync_tasks st
   join target_artifact a on a.id = st.artifact_id;
   ```

   Actual result: zero rows.

9. Wait for the normal sync worker interval, then query the target peer:

   ```shell
   curl -i -u '<user>:<token>' \
     'https://<target-peer>/pypi/PyPIReplicationSmoke/simple/ak-pypi-replication-smoke/'
   ```

   Actual result: HTTP `404`, `{"code":"NOT_FOUND","message":"Package not found"}`.

10. Compare with a generic repository upload on the same peers. Generic upload
    queues peer `sync_tasks` and replication can proceed, which shows the peer
    setup itself is not the missing piece.

### Logs

```shell
Sanitized relevant source-peer backend log:

INFO http_request{method=POST uri=/pypi/<repo>/}: artifact_keeper_backend::api::handlers::pypi: PyPI upload: <project> <version> to repo <repo> (<filename>)
INFO http_request{method=POST uri=/pypi/<repo>/}: Request completed status=200

No source-peer log line was observed for `Queued sync tasks` for the PyPI artifact.

Sanitized target-peer response during the wait window:

HTTP/2 404
{"code":"NOT_FOUND","message":"Package not found"}

No sync-worker transfer, chunked upload, retry, or permanent-failure log was
observed for the PyPI artifact, because no sync task was created.
```

### Additional Context

Code-level notes from current `main` inspection:

- The generic upload path calls `ArtifactService::upload_with_sync_options(...,
  !is_replication_request(&headers))`, so normal client uploads queue
  replication and incoming replication writes do not recurse.
- The hosted Debian/APT path was fixed in #1810 to use the same pattern.
- The PyPI upload handler writes storage/artifact rows directly and therefore
  bypasses the sync-task fan-out path.
- As a result, hosted PyPI uploads are local-only even when the repository has
  push/mirror peer subscriptions.

Potential fix direction:

- Treat normal hosted PyPI uploads like generic uploads: enqueue peer sync tasks
  unless the request is an incoming replication write.
- Preserve PyPI `artifact_metadata` and package catalog metadata on replicated
  writes so the receiving peer can serve Simple API project pages, files, PEP
  658 metadata sidecars, and package catalog/API rows.
- Add focused tests proving:
  - normal PyPI `POST /pypi/{repo}/` queues sync tasks for push/mirror peers,
  - incoming replication-marked PyPI writes do not requeue,
  - replicated PyPI artifacts preserve `artifact_metadata`, `packages`, and
    `package_versions`.

### Pre-submission Checklist

valfon1393:
## Summary

Fixes #1864 

Hosted PyPI uploads were persisted locally but never queued for peer
replication because the PyPI upload handler stored blobs and inserted artifact
rows directly instead of using the shared artifact upload path that fans out
`sync_tasks` for push/mirror peer subscriptions.

This PR makes normal hosted PyPI uploads enqueue peer sync tasks while still
suppressing fan-out for incoming replication writes marked with
`x-artifact-keeper-replication`. It also preserves PyPI-specific metadata on the
artifact and package catalog rows so a fresh peer can serve the replicated
package through the Simple API, PEP 658 `.metadata` sidecars, package APIs, and
native `pip` flows.

While validating the PyPI HA flow, two additional replication consistency issues
showed up and are fixed here as part of making PyPI replication production-safe:

- replication-owned resumable upload sessions are now marked and cleaned up
  after peer replication completes, and stale replication sessions for the same
  artifact path are removed before retrying that path
- `package_versions` now keeps a deterministic representative row for package
  formats such as PyPI that can publish multiple distributions for the same
  package/version, avoiding DB row divergence when peers process wheel/sdist
  artifacts in different orders

Added/updated coverage includes:

- direct PyPI uploads enqueue sync tasks when a repository has a push/mirror
  peer subscription
- replication-marked PyPI uploads persist metadata but do not requeue back to
  peers
- PyPI upload metadata is stored as `artifact_metadata(format = 'pypi')` and in
  package catalog metadata
- retrying peer replication removes stale replication upload sessions/chunks and
  stale temp files for the same repository/path
- ordinary completed resumable upload sessions remain queryable for client
  status, while replication sessions are cleaned up after completion
- `package_versions` representative selection is deterministic for multiple
  artifacts sharing the same package/version

## Test Checklist

- [x] Manually tested locally

Validation performed for this branch:

```text
git diff --check
git diff --cached --check
```

The local Rust toolchain was not available in the preparation shell, so
`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
and `cargo test --workspace --lib` still need to be run in a Rust-enabled local
or CI environment before merge.

Manual/live HA validation was run against two Artifact Keeper peers using hosted
PyPI repositories:

- checksum-negative PyPI uploads were rejected with HTTP 400 in both directions
- single-package PyPI replication succeeded in both directions, including API,
  Simple API, DB rows, downloads, and Docker `pip install`
- large chunked PyPI package replication was interrupted mid-transfer on the
  target peer after 3 of 4 chunks, then recovered after a 20-minute outage
- the interrupted 172,305,868-byte wheel arrived byte-identical on the target
  peer with the expected SHA-256
- post-recovery direct downloads succeeded on both peers: 239/239 each, 0
  failures
- post-recovery Docker `pip` matrix succeeded on both peers: 12/12 each, 0
  failures
- `artifacts`, `artifact_metadata`, `packages`, and `package_versions` were
  equivalent across peers after replication settled
- no stale upload sessions remained on either peer/repository after recovery
- the PyPI HA protocol reported PASS with 0 HA warnings

## API Changes

- [ ] N/A - no API changes

This adds an internal `upload_sessions.is_replication` flag plus an index used
to distinguish replication-owned resumable upload sessions from ordinary client
upload sessions during retry/cleanup. It does not add new endpoints.

### DIFF
--- a/backend/migrations/134_upload_session_replication_flag.sql
+++ b/backend/migrations/134_upload_session_replication_flag.sql
@@ -1,0 +1,23 @@
+-- Mark upload sessions created by peer replication.
+--
+-- The generic resumable upload API is also used by inter-peer replication.
+-- Keeping this flag lets retry/recovery cleanup remove only replication-owned
+-- leftovers without touching ordinary client resumable uploads.
+
+ALTER TABLE upload_sessions
+    ADD COLUMN IF NOT EXISTS is_replication BOOLEAN NOT NULL DEFAULT false;
+
+UPDATE upload_sessions
+SET is_replication = true
+WHERE is_replication = false
+  AND (
+      artifact_metadata_format IS NOT NULL
+      OR artifact_metadata IS NOT NULL
+      OR artifact_metadata_properties IS NOT NULL
+      OR package_description IS NOT NULL
+      OR package_metadata IS NOT NULL
+  );
+
+CREATE INDEX IF NOT EXISTS idx_upload_sessions_replication_path
+    ON upload_sessions (repository_id, artifact_path)
+    WHERE is_replication = true;
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -1569,6 +1569,7 @@ async fn upload(
     State(state): State<SharedState>,
     Extension(auth): Extension<Option<AuthExtension>>,
     Path(repo_key): Path<String>,
+    headers: HeaderMap,
     mut multipart: Multipart,
 ) -> Result<Response, Response> {
     // Authenticate
@@ -1717,21 +1718,12 @@ async fn upload(
         return Err(AppError::Conflict("File already exists".to_string()).into_response());
     }
 
-    // Store the file
-    let storage_key = format!("pypi/{}/{}/{}", normalized, pkg_version, filename);
-    let storage = state
-        .storage_for_repo(&repo.storage_location())
-        .map_err(|e| e.into_response())?;
-    storage
-        .put(&storage_key, content.clone())
-        .await
-        .map_err(map_storage_err)?;
-
     // Build metadata JSON
     let mut pkg_metadata = serde_json::json!({
-        "name": pkg_name,
-        "version": pkg_version,
-        "filename": filename,
+        "name": &pkg_name,
+        "normalized_name": &normalized,
+        "version": &pkg_version,
+        "filename": &filename,
     });
     if let Some(rp) = &requires_python {
         pkg_metadata["pkg_info"] = serde_json::json!({
@@ -1739,10 +1731,15 @@ async fn upload(
         });
     }
     if let Some(s) = &summary {
-        pkg_metadata["pkg_info"]
-            .as_object_mut()
-            .get_or_insert(&mut serde_json::Map::new())
-            .insert("summary".to_string(), serde_json::Value::String(s.clone()));
+        if !pkg_metadata["pkg_info"].is_object() {
+            pkg_metadata["pkg_info"] = serde_json::json!({});
+        }
+        if let Some(pkg_info) = pkg_metadata["pkg_info"].as_object_mut() {
+            pkg_info.insert("summary".to_string(), serde_json::Value::String(s.clone()));
+        }
+    }
+    if !metadata_fields.is_empty() {
+        pkg_metadata["upload_metadata"] = serde_json::Value::Object(metadata_fields);
     }
 
     let content_type = pypi_content_type(&filename);
@@ -1752,42 +1749,28 @@ async fn upload(
 
     super::cleanup_soft_deleted_artifact(&state.db, repo.id, &artifact_path).await;
 
-    // Insert artifact record
-    let artifact_id = sqlx::query_scalar!(
-        r#"
-        INSERT INTO artifacts (
-            repository_id, path, name, version, size_bytes,
-            checksum_sha256, content_type, storage_key, uploaded_by
+    let storage = state
+        .storage_for_repo(&repo.storage_location())
+        .map_err(|e| e.into_response())?;
+    let artifact_service = state.create_artifact_service(storage);
+    let artifact = artifact_service
+        .upload_with_sync_options(
+            repo.id,
+            &artifact_path,
+            &normalized,
+            Some(&pkg_version),
+            content_type,
+            content,
+            Some(user_id),
+            should_enqueue_pypi_sync_tasks(&headers),
         )
-        VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)
-        RETURNING id
-        "#,
-        repo.id,
-        artifact_path,
-        normalized,
-        pkg_version,
-        size_bytes,
-        computed_sha256,
-        content_type,
-        storage_key,
-        user_id,
-    )
-    .fetch_one(&state.db)
-    .await
-    .map_err(map_db_err)?;
+        .await
+        .map_err(|e| e.into_response())?;
 
-    // Store metadata
-    let _ = sqlx::query!(
-        r#"
-        INSERT INTO artifact_metadata (artifact_id, format, metadata)
-        VALUES ($1, 'pypi', $2)
-        ON CONFLICT (artifact_id) DO UPDATE SET metadata = $2
-        "#,
-        artifact_id,
-        pkg_metadata,
-    )
-    .execute(&state.db)
-    .await;
+    artifact_service
+        .set_metadata(artifact.id, "pypi", pkg_metadata, serde_json::json!({}))
+        .await
+        .map_err(|e| e.into_response())?;
 
     // Update repository timestamp
     let _ = sqlx::query!(
@@ -1806,9 +1789,12 @@ async fn upload(
                 &normalized,
                 &pkg_version,
                 size_bytes,
-                &computed_sha256,
+                &artifact.checksum_sha256,
                 summary.as_deref(),
-                Some(serde_json::json!({ "format": "pypi" })),
+                Some(build_pypi_package_catalog_metadata(
+                    &filename,
+                    requires_python.as_deref(),
+                )),
             )
             .await;
     }
@@ -1827,6 +1813,24 @@ async fn upload(
 // ---------------------------------------------------------------------------
 // Helpers
 // ---------------------------------------------------------------------------
+
+fn should_enqueue_pypi_sync_tasks(headers: &HeaderMap) -> bool {
+    !super::is_replication_request(headers)
+}
+
+fn build_pypi_package_catalog_metadata(
+    filename: &str,
+    requires_python: Option<&str>,
+) -> serde_json::Value {
+    let mut metadata = serde_json::json!({
+        "format": "pypi",
+        "filename": filename,
+    });
+    if let Some(rp) = requires_python.filter(|value| !value.trim().is_empty()) {
+        metadata["requires_python"] = serde_json::Value::String(rp.to_string());
+    }
+    metadata
+}
 
 /// Determine the Content-Type for a PyPI filename based on its extension.
 fn pypi_content_type(filename: &str) -> &'static str {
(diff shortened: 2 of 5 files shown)


brandonrc (approved):
**Approve** — reviewed the full diff and verified the bug reproduces on `main`.

**Bug confirmed (#1864):** the hosted PyPI upload handler stored the blob with a raw `storage.put` and inserted the `artifacts` row via raw `INSERT INTO artifacts ... RETURNING id` — it never called `ArtifactService::upload_with_sync_options`, which is the only place that fans out `sync_tasks` for push/mirror peer subscriptions. So a PyPI package uploaded to one peer never produced a `sync_tasks` row and the HA target peer never received it. Debian's `persist_debian_upload` already routes through the shared service; PyPI was the outlier.

**Fix is correct:** the handler now goes through `upload_with_sync_options(... should_enqueue_pypi_sync_tasks(&headers))` + `set_metadata`, i.e. the same shared fan-out path as Debian/generic. Loop suppression is the proven pattern — `should_enqueue_pypi_sync_tasks = !is_replication_request(headers)`, so an incoming replication write (carrying `x-artifact-keeper-replication`) passes `enqueue=false` and isn't re-queued back to the source. Migration 134 is additive (`ADD COLUMN IF NOT EXISTS ... DEFAULT false` + partial index) and its backfill correctly targets only historical replication sessions (on `main` those session-metadata columns are populated only for replication requests). The new session cleanup is properly scoped: `cleanup_stale_replication_upload_sessions` deletes only `is_replication = true` rows, and `cleanup_completed_upload_session` runs only for replication sessions and only on `status = 'completed'` — ordinary client resumable sessions are untouched and stay queryable. Catalog metadata is enriched (filename / requires_python) rather than lost.

**One thing to flag (non-blocking):** the `package_versions` representative-row change in `package_service.rs` lives in the shared cross-format upsert, so it changes "last-writer-wins" → deterministic "smallest-(checksum, size)-wins" for **every** format, not just PyPI. It's the right call for peer convergence and the blast radius is small (only the representative catalog row; `artifacts` and `artifact_metadata` are unaffected), but it's worth being explicit that it touches non-PyPI re-upload semantics — a re-upload of the same version with a higher-sorting checksum will no longer move the representative row.

The two-peer HA validation in the description (interrupted-chunk recovery, byte-identical SHA, equivalent DB rows across peers) is a great touch. 👍

Issue #1608, opened 2026-06-04
Core Invariant ①: enforce streaming — no full-body buffering on artifact paths


brandonrc:
Part of #1607 (Core Invariants Hardening). Source: `docs/audits/design-retro-2026-06-04.md`.

## Invariant
**No artifact-path handler may read a full artifact body into memory** (`.bytes()` / object-store `get()` on artifact content). Uploads stream → incremental SHA-256 → bounded multipart; downloads map HTTP Range → storage Range. Peak memory per transfer ≈ one multipart part.

## Why
16 memory/OOM issues trace to ad-hoc full-buffering. The clean-room design made streaming a structural invariant; we retrofitted it per-call.

## Work
- [ ] Add an architectural test / review rule forbidding full-body reads on artifact paths.
- [ ] Convert `GcsBackend` to streaming get/put.
- [ ] Stream the scanner (`fetch_artifact_content`) instead of buffering in heap.
- [ ] Stop staging large uploads to `/tmp` by default (Kubernetes pod eviction).
- [ ] Lift the OCI ~50 GiB S3 ceiling via adaptive multipart part sizing.
- [ ] Fix O(N) sidecar load in `ProxyService::list_cached_artifacts`.

## Linked open issues
#1523, #1571, #1573 (+ scanner/GCS OOM)

brandonrc:
**Prerequisite:** #1631 (coordination seam) — this issue lands on that seam, not directly on #1618. The multi-replica harness #1627 is the acceptance gate.

Issue #1611, opened 2026-06-04
Core Invariant ④: cache correctness — immutable-vs-mutable path classification


brandonrc:
Part of #1607 (Core Invariants Hardening). Source: `docs/audits/design-retro-2026-06-04.md`.

## Invariant
Cache freshness is a **per-path-pattern property declared by each format handler**: immutable paths (versioned Maven jar, OCI blob-by-digest) cache forever; mutable paths (maven-metadata.xml, PyPI simple index, npm packument, OCI tag→manifest) get short TTL + conditional revalidation (ETag/If-None-Match, Last-Modified). Virtual index merges fan out with per-member timeouts and degrade to partial results.

## Why
Most virtual/proxy correctness bugs are immutable-vs-mutable misclassification (stale or shadowed indexes, needless DB lookups before proxying).

## Work
- [ ] Add an immutability classifier to the `FormatHandler` interface (Maven/PyPI/npm/OCI/Cargo).
- [ ] Conditional revalidation for mutable paths; infinite cache for immutable.
- [ ] Track `etag`, `fetched_at`, `expires_at`, `negative_cached_until` per cached path.
- [ ] Fix the Maven/PyPI virtual family below.
- [ ] Expose cache hit/miss/revalidate/stale metrics per repo.

## Linked open issues
#1547, #1562, #1595, #1599, #1600

---
**Swept-in open defects:** #1554 — Swift virtual read endpoints 404 (queries by virtual id instead of members); #1566 — Tofu/Terraform remote providers fail on init (provider not found in mirror search locations). Both are virtual/remote resolution correctness.

PR #1866, opened 2026-06-11
fix(pypi): stream remote package downloads instead of buffering in memory

### LINKED ISSUE #1863: Proxy stalls on large PyPI wheel responses; nvidia_nccl_cu12-2.30.7 times out with zero bytes until upstream finishes
### Description

Artifact Keeper stalls on large PyPI wheel responses. The `nvidia_nccl_cu12-2.30.7-py3-none-manylinux_2_18_aarch64.whl` request times out from the client side with 0 bytes received, even though the upstream URL starts transferring immediately.

### Environment Information

- Deployment type: Kubernetes
- Operating System: Ubuntu 24.04 nodes
- Architecture: amd64 cluster, fetching an arm64 wheel through PyPI proxy
- Database: PostgreSQL
- Reverse Proxy: none in front of Artifact Keeper
- Network setup: in-cluster service access on the private cluster network

### Backend Version

1.2.0

### Web UI Version

N/A

### Package Format

PyPI

### Expected Behavior

The proxy should begin responding promptly for large artifacts, or stream data to the client as upstream data arrives. A large wheel should not sit at 0 bytes until the upstream fetch finishes.

### Observed Behavior

Direct upstream fetch from `files.pythonhosted.org` starts immediately and transfers data.
Through Artifact Keeper, both `GET` and `HEAD` to the same wheel URL time out with 0 bytes received.
Backend logs show Artifact Keeper fetching:
- `https://pypi.org/simple/nvidia-nccl-cu12/`
- `https://files.pythonhosted.org/packages/.../nvidia_nccl_cu12-2.30.7-py3-none-manylinux_2_18_aarch64.whl`

### Steps to Reproduce

1. Start a pod inside the cluster with network access to Artifact Keeper.
2. Request the proxy URL for the NCCL wheel:
   `GET /pypi/pypi-remote/simple/nvidia-nccl-cu12/nvidia_nccl_cu12-2.30.7-py3-none-manylinux_2_18_aarch64.whl`
3. Compare it to the direct upstream URL on `files.pythonhosted.org`.
4. Observe that upstream begins transferring quickly, while Artifact Keeper returns no bytes before client timeout.

### Logs

```shell
- Proxy logs show the request entering the PyPI proxy path and fetching from upstream.
- No completed response is logged for the NCCL request before the client timeout.
- Example client result through Artifact Keeper:
  `curl: (28) Operation timed out ... with 0 bytes received`
```

### Additional Context

Related / existing work

- #1266, slow read-path lookups in the PyPI proxy helpers
  - adjacent because it affects the same PyPI hot path
  - but it is a lookup-latency problem, not the response-streaming stall here

- #1401, large-transfer stability in migration paths
  - related because it explicitly calls out large-transfer hardening and streaming requirements
  - but it is a migration-path issue, not the proxy response path

- #1393, stream artifact downloads across all package formats
  - this is the closest streaming-related work
  - the PR description says it streams large artifact bodies straight from storage instead of buffering them in heap
- This may be related to the broader streaming effort already underway in #1393 / #1608, but the NCCL repro still shows the proxy path stalling with zero bytes returned to the client.

Current work in our platform repo already reproduced this live in-cluster and confirmed:
- direct upstream response starts immediately
- Artifact Keeper holds the response open until upstream work progresses
- the issue is specific to the proxy response behavior, not DNS or repo policy state

### Pre-submission Checklist

GabrielSosaDevopsFarinter:
Closes #1863
Advances #1608 (Core Invariant ①: no full-body buffering on artifact paths).

## Problem

Remote/proxy PyPI package downloads buffered the entire wheel/sdist body in memory before serving it. On multi-hundred-MiB files this stalled the client with zero bytes until the upstream transfer finished, and OOM-killed memory-constrained pods (#895 built the tee primitive; PyPI never adopted it).

## Change

Route package-file downloads through the streaming proxy path, teeing the upstream body into the cache while serving it to the client — no full-body buffering.

PyPI keys its cache on `simple/{project}/{filename}` while the files themselves live on a different upstream host (e.g. `files.pythonhosted.org`). To make the existing streaming path usable here, this adds cache-path-aware helpers that decouple the fetch URL from the cache key:

- `ProxyService::fetch_artifact_streaming_with_cache_path(repo, fetch_path, cache_path)` — streaming sibling of `fetch_artifact_with_cache_path`; fetches from `fetch_path` but keys cache/TTL/negative-cache/scan decisions on `cache_path`.
- `ProxyService::streaming_cached_artifact_by_path(repo, cache_path)` — streaming cache probe; serves a hit straight from storage, returns `None`/`NotFound` on miss/negative-cache.
- `proxy_helpers::proxy_fetch_streaming_with_cache_key` / `proxy_check_cache_streaming` — handler-layer wrappers with best-effort probe semantics.

Simple-index pages stay buffered by design (small).

## Testing

- `cargo clippy --all-targets` — clean, 0 warnings
- `cargo test --lib pypi` — 234 passed, 0 failed

### DIFF
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -937,6 +937,64 @@ pub async fn proxy_fetch_with_cache_key(
     .await
 }
 
+/// Streaming sibling of [`proxy_fetch_with_cache_key`] (#895 OOM relief for
+/// format handlers whose upstream download URL differs from the canonical
+/// artifact path). Fetches `fetch_path` from the upstream but keys the proxy
+/// cache on `cache_path`, returning the body as a [`StreamingFetchResult`]
+/// that the caller tees to the client without buffering.
+pub async fn proxy_fetch_streaming_with_cache_key(
+    proxy_service: &ProxyService,
+    repo_id: Uuid,
+    repo_key: &str,
+    upstream_url: &str,
+    fetch_path: &str,
+    cache_path: &str,
+) -> Result<crate::services::proxy_service::StreamingFetchResult, Response> {
+    with_proxy_repo(
+        repo_id,
+        repo_key,
+        upstream_url,
+        fetch_path,
+        |repo| async move {
+            proxy_service
+                .fetch_artifact_streaming_with_cache_path(&repo, fetch_path, cache_path)
+                .await
+        },
+    )
+    .await
+}
+
+/// Streaming sibling of [`proxy_check_cache`]: probe the proxy cache for
+/// `cache_path` and stream a hit straight from storage instead of buffering
+/// the cached body in memory. Returns `None` on miss or on any probe error
+/// (including a negative-cache hit) — best-effort semantics matching the
+/// buffered probe, so callers fall through to the full fetch, which
+/// re-applies the negative-cache gate itself.
+pub async fn proxy_check_cache_streaming(
+    proxy_service: &ProxyService,
+    repo_id: Uuid,
+    repo_key: &str,
+    upstream_url: &str,
+    cache_path: &str,
+) -> Option<crate::services::proxy_service::StreamingFetchResult> {
+    let repo = build_remote_repo(repo_id, repo_key, upstream_url);
+    match proxy_service
+        .streaming_cached_artifact_by_path(&repo, cache_path)
+        .await
+    {
+        Ok(result) => result,
+        Err(e) => {
+            tracing::debug!(
+                "Streaming cache probe failed for {}/{}, treating as miss: {}",
+                repo_key,
+                cache_path,
+                e
+            );
+            None
+        }
+    }
+}
+
 /// Fetch from upstream directly, bypassing the proxy cache.
 ///
 /// Use this instead of [`proxy_fetch`] when the caller needs the raw upstream
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -4741,4 +4741,190 @@ mod tests {
 
         do_cleanup().await;
     }
+
+    // -----------------------------------------------------------------------
+    // resolve_pypi_remote_fetch_target / fetch_from_pypi_remote_streaming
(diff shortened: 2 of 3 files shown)


brandonrc (approved):
**Approve** — reviewed the full diff and verified the bug reproduces on `main`.

**Bug confirmed (#1863):** the PyPI remote/proxy package path is the *only* proxy format still buffering the whole body — `proxy_service.rs` does `response.bytes().await` and `pypi.rs` serves it via `Body::from(content)` with a precomputed `Content-Length`, so the client gets zero bytes until the upstream transfer finishes (and RSS spikes by the wheel size → OOM on constrained pods). go / maven / debian / alpine / terraform / gitlfs / generic blobs already go through `proxy_fetch_streaming`; PyPI couldn't, because its fetch URL (`files.pythonhosted.org/...`) differs from its cache key (`simple/{project}/{filename}`) and no streaming + cache-path helper existed.

**Fix is correct:** the PR adds exactly that missing primitive (`fetch_artifact_streaming_with_cache_path` + `proxy_fetch_streaming_with_cache_key` / `proxy_check_cache_streaming`) and routes both the single-repo and virtual-repo arms through the tee path. `fetch_path` (outbound URL) vs `cache_path` (cache identity) is threaded cleanly through the whole streaming machinery — storage key, `cache_ttl_for_path` (#1611), negative-cache, and `warn_if_proxy_scan_unsupported` (#1274) all correctly key off `cache_path`, while only `build_upstream_url` uses `fetch_path`. Single-flight (#1631/#1694) is preserved, the cache key stays the stable `simple/{project}/{filename}`, simple-index pages stay buffered by design, and `Content-Length` is forwarded only when upstream advertises one (else chunked). The buffered `fetch_from_pypi_remote` is correctly retained for the cache-recovery closure that needs owned `Bytes`.

**Non-blocking note:** the streaming cache-hit fast path skips the slow-path SHA re-verification that the buffered variant did — but that's the already-accepted #1018 tradeoff shared by every streaming format, not a regression introduced here.

Clean, well-documented, and tested. 👍

PR #1878, opened 2026-06-12
fix(db): evict pooled Postgres connections that fail a liveness probe

### LINKED ISSUE #1877: Backend sqlx pool does not recover from persistent `Device or resource busy (os error 16)` on Postgres connections — only a pod restart fixes it
# Backend sqlx pool does not recover from persistent `Device or resource busy (os error 16)` on Postgres connections — only a pod restart fixes it

## Summary

On a Kubernetes deployment of v1.1.9 using the official Helm chart with in-cluster Postgres, after ~1.5 hours of healthy operation every database operation started failing with:

```
error communicating with database: Device or resource busy (os error 16)
```

Once this state was entered, **it never recovered**: every scheduler tick, every API call that touched the DB, every sync-worker loop kept emitting the same error for as long as the pod was up (we observed it for ~90 minutes before manually restarting). The only fix was `kubectl delete pod` on the backend.

Postgres itself was completely healthy throughout — it kept doing 5-minute checkpoints, `pg_isready` from inside the Postgres pod returned `accepting connections`, and `psql` could query `pg_stat_activity` and see active connections fine.

I believe what triggered the original socket-level error was an environment-level event (CNI flow reprogramming on a kube-OVN cluster), but the **application-side bug** is that the sqlx pool keeps re-using dead connections rather than evicting them and reconnecting, which turns a transient network glitch into a permanent outage.

## Environment

- Artifact Keeper backend: `v1.1.9` (image `ghcr.io/artifact-keeper/artifact-keeper-backend:1.1.9`)
- Postgres: 16-alpine, in-cluster (deployed by the chart)
- Kubernetes: v1.34.5, single-node, CNI: kube-OVN + kube-proxy
- Storage: NFS-backed PVCs
- Deployment: Helm chart from `artifact-keeper-iac` (`values-production.yaml` + a single-node PoC overlay)

## Symptoms

Backend log excerpt — every scheduler tick fails identically:

```
ERROR services::sync_worker: Sync worker error: Failed to fetch peers: error communicating with database: Device or resource busy (os error 16)
WARN  services::scheduler_service: Webhook retry processing failed: Failed to fetch retry queue: error communicating with database: Device or resource busy (os error 16)
WARN  services::scheduler_service: Curation sync cycle failed: error communicating with database: Device or resource busy (os error 16)
WARN  services::scheduler_service: Backup schedule check failed: Database error: error communicating with database: Device or resource busy (os error 16)
WARN  services::scheduler_service: Lifecycle policy execution failed: Database error: error communicating with database: Device or resource busy (os error 16)
WARN  services::scheduler_service: Health monitoring cycle failed: Database error: error communicating with database: Device or resource busy (os error 16)
ERROR services::sync_worker: Failed to run stale peer detection: Database error: error communicating with database: Device or resource busy (os error 16)
...
```

During the same window:

- `/livez` kept returning 200.
- `/readyz` started timing out (`context deadline exceeded`) — readiness probe correctly flipped the pod to 0/1 Ready.
- `POST /api/v1/auth/login` returned 500 to the UI (the backend tried to hit the DB and saw the same EBUSY).
- `kubectl exec` into the Postgres pod, `psql` against `artifact_registry` worked perfectly. `SELECT count(*) FROM pg_stat_activity` showed live connections.
- No restarts on the Postgres or backend pod. Postgres logs showed only routine `checkpoint complete` lines.

After `kubectl delete pod` on the backend, the new pod connected cleanly and operated normally — but the same pool-broken state recurs eventually under the same conditions.

## What I think is happening

`Device or resource busy` (`EBUSY` / `errno 16`) on a TCP read/write is unusual — normal TCP failure modes are `ECONNRESET`, `EPIPE`, `ETIMEDOUT`. On Linux this code surfaces when the underlying virtual device is busy / momentarily not writable, which is consistent with a CNI/OVS flow reprogramming event on the path. The kernel returns EBUSY → tokio surfaces it as `io::Error` → sqlx wraps it as `sqlx::Error::Io`.

The application bug, regardless of what triggered the first EBUSY, is that sqlx's pool keeps handing back the same connection object on subsequent acquires instead of marking it broken and reconnecting. Every subsequent query gets the cached error.

## Expected behavior

When a pooled Postgres connection encounters a non-recoverable IO error (EBUSY, ECONNRESET, EPIPE, etc.) on read/write, the pool should:

1. Drop that connection (do not return it to the pool on release).
2. Open a new TCP connection on the next acquire.
3. Optionally: trip a circuit breaker / surface the failure on `/readyz` so Kubernetes restarts the pod after N consecutive failures.

A single transient network event on the path should not require manual intervention to recover.

## Repro (best effort)

Hard to deterministically reproduce because the trigger is a low-level network event, but the conditions where we hit it:

- Multi-hour uptime
- Helm chart `production` overlay + single-node Postgres
- kube-OVN CNI with NetworkPolicy enabled (this is the chart's default — the NP enforcement likely contributed to the OVS flow event that broke ESTABLISHED sockets)

A more direct repro would be:

1. Run a synthetic test that injects a transient socket failure on the Postgres connection (e.g. drop traffic via iptables for 30s, or restart the Postgres service mid-query).
2. Observe that the backend never recovers even after the network/Postgres is fully restored.

## Suggested investigation

- Audit how `sqlx::Pool<Postgres>` is configured. In particular `before_acquire` / `test_before_acquire` and `max_lifetime` settings. A reasonable default would be to set `test_before_acquire(true)` and a finite `max_lifetime`/`idle_timeout`.
- Check if there's a code path that holds a `PoolConnection` across an await point and retries on error rather than dropping it.
- Consider adding a periodic pool health-check task that probes a connection and forcibly drains+rebuilds the pool when consecutive failures cross a threshold.

Happy to share more logs or run a more targeted repro if helpful.

kycheng:
## Summary

Adds a `before_acquire` liveness probe to the Postgres pool so a broken pooled connection is evicted before the caller sees an `Error::Io`. Without this, a transient network-path failure (CNI flow reflow, NAT rotation, brief Postgres unavailability) is sticky — every subsequent query keeps hitting the same dead connection and only a pod restart recovers the backend.

The default `test_before_acquire` only sends a protocol PING, which can succeed on a TCP socket the kernel has already marked dead. A real `SELECT 1` surfaces the breakage at acquire time so sqlx drops the connection.

To keep the hot path cheap, the probe only runs on connections that have been idle for `IDLE_LIVENESS_THRESHOLD` (30 s). Acquires of recently-used connections skip the extra round trip.

Closes #1877

## Test plan

- [x] `cargo check --package artifact-keeper-backend --lib` passes
- [x] `cargo fmt --all -- --check` passes
- [ ] CI green
- [ ] Maintainer review on the threshold value (`30s` is a guess — happy to make it configurable via env if you'd prefer)
- [ ] Repro the original failure mode in a follow-up integration test (e.g. kill Postgres mid-query, drop traffic for 30 s, verify pool recovers without restart)

## Notes for review

- I cannot deterministically reproduce the trigger locally (the original repro involved kube-OVN flow reprogramming on a single-node cluster), so this is a targeted defensive change rather than a verified end-to-end fix. If you have an integration-test harness that can fault-inject Postgres traffic, that would be the ideal place to validate.
- Open to making the threshold configurable (e.g. `DATABASE_LIVENESS_PROBE_AFTER_IDLE_SECS`) — left it as a const for the minimal patch.
- Did not touch `test_before_acquire` since it's already `true` by default in sqlx 0.8; the `before_acquire` callback runs after the default test and supersedes it on the idle path.

### DIFF
--- a/backend/src/db.rs
+++ b/backend/src/db.rs
@@ -5,6 +5,11 @@ use crate::error::Result;
 use sqlx::postgres::{PgPool, PgPoolOptions};
 use std::time::Duration;
 
+/// Connections idle longer than this run a full `SELECT 1` probe before being
+/// returned by the pool. Short-idle acquires skip the probe to keep the hot
+/// path free of an extra round trip.
+const IDLE_LIVENESS_THRESHOLD: Duration = Duration::from_secs(30);
+
 /// Create a new database connection pool using the connection pool settings
 /// from [`Config`]. Pool sizing and timeouts are configurable via the
 /// `DATABASE_MAX_CONNECTIONS`, `DATABASE_MIN_CONNECTIONS`,
@@ -18,6 +23,26 @@ pub async fn create_pool(config: &Config) -> Result<PgPool> {
         .acquire_timeout(Duration::from_secs(config.database_acquire_timeout_secs))
         .idle_timeout(Duration::from_secs(config.database_idle_timeout_secs))
         .max_lifetime(Duration::from_secs(config.database_max_lifetime_secs))
+        .before_acquire(|conn, meta| {
+            Box::pin(async move {
+                // sqlx's default `test_before_acquire` only sends a protocol
+                // PING, which can succeed on a TCP socket that has been
+                // silently broken by an upstream event (CNI flow reflow,
+                // NAT rotation, brief Postgres unavailability). When that
+                // happens, the next real query fails with an IO error and
+                // the pool keeps handing back the same dead connection,
+                // turning a transient glitch into a permanent outage that
+                // only a pod restart fixes. Issue #1877.
+                //
+                // For connections that have actually been idle, run a real
+                // query so a stale socket is detected here and the
+                // connection is evicted by sqlx before any caller sees it.
+                if meta.idle_for >= IDLE_LIVENESS_THRESHOLD {
+                    sqlx::query("SELECT 1").execute(&mut *conn).await?;
+                }
+                Ok(true)
+            })
+        })
         .connect(&config.database_url)
         .await?;
 


brandonrc (approved):
**Approve** — correct, idiomatic, low-risk; verified against sqlx 0.8.6 semantics.

The gap is real: `main`'s pool (`backend/src/db.rs:14-25`) relies on sqlx's default `test_before_acquire = true`, but that default test is only a protocol **PING** (`conn.ping()`), which can appear to succeed on a socket the kernel has silently wedged (the kube-OVN `EBUSY`/flow-reprogramming case in #1877) — so the dead connection survives the test and the *next real query* fails until the pod restarts.

The fix adds a `before_acquire` hook running `SELECT 1` only when `meta.idle_for >= 30s` — an additive second-layer probe that forces a real query round-trip (unbuffered read), surfacing a wedged socket the PING misses. Eviction semantics are correct: a failed probe propagates `Err` → sqlx `close_hard()` → fresh reconnect; a live conn returns `Ok(true)`. Latency cost is one trivial round-trip only on conns idle >30s — negligible.

Non-blocking notes (suggest as follow-ups): it's a *targeted* fix (only catches at acquire-time after ≥30s idle; a conn that dies mid-use or is reacquired <30s still relies on the default PING), the `30s` threshold is hardcoded (env-configurable would be nice), and there's no fault-injection integration test proving self-heal. None block this — it's strictly better than `main`. 👍

PR #1883, opened 2026-06-12
test(proxy): add regression test for doubled-prefix storage routing bug (#1881)

### LINKED ISSUE #1881: Proxy-cache read fails with "No such file or directory" even when __content__ exists on disk — all cache hits return 500 after first fetch
## Summary

After a first successful fetch through a proxy (Remote) repository, all subsequent reads of the same proxied artifact return HTTP 500 with `code: \"STORAGE_ERROR\"` and `\"Failed to read proxy-cache/<repo>/<path>/__content__: No such file or directory (os error 2)\"` — even though the `__content__` file demonstrably exists on disk, has the correct content, and is readable by the AK process from the shell. The cache is therefore anti-functional on filesystem-backed deployments: one fetch succeeds, every repeat fetch (mvn rebuild, CI rerun, dependency-resolution re-trigger) fails until manual cache purge.

## Environment

- **AK version**: v1.1.9
- **Storage backend**: filesystem (NFS-mounted volume at `/data/storage`)
- **Repository type**: Maven `remote` (proxy), configured with upstream `https://repo1.maven.org/maven2/`
- **OS**: standard Linux container (Kubernetes pod), uid=1001(artifact), gid=0(root)

## Reproduction Steps

```
# First request — cold cache miss, fetches from upstream, writes cache
GET /maven/team-maven-proxy/org/junit/junit/4.13.2/junit-4.13.2.jar
# → 200 OK, 384581 bytes
# AK log: Fetched 384581 bytes from upstream (content_type: Some("application/java-archive"),
#          etag: Some("\"d98a9a02a99a9acd22d7653cbcc1f31f\""))

# Second request (a few seconds later) — should be a cache hit
GET /maven/team-maven-proxy/org/junit/junit/4.13.2/junit-4.13.2.jar
# → 500 Internal Server Error
# Body: {"code":"STORAGE_ERROR","message":"Storage operation failed"}
# AK log: Storage error: Failed to read
#          proxy-cache/team-maven-proxy/junit/junit/4.13.2/junit-4.13.2.jar/__content__:
#          No such file or directory (os error 2)  code="STORAGE_ERROR"
```

All subsequent requests also return 500. The behavior is 100% reproducible.

## Evidence the File Exists on Disk

Immediately after the second 500, from within the running backend pod:

```
$ ls -la /data/storage/proxy-cache/team-maven-proxy/junit/junit/4.13.2/junit-4.13.2.jar/
drwxr-sr-x 2 artifact root     52 Jun 12 07:21 .
drwxr-sr-x 3 artifact root     30 Jun 12 07:21 ..
-rw-r--r-- 1 artifact root    294 Jun 12 07:21 __cache_meta__.json
-rw-r--r-- 1 artifact root 384581 Jun 12 07:21 __content__

$ id
uid=1001(artifact) gid=0(root) groups=0(root)

$ md5sum /data/storage/proxy-cache/team-maven-proxy/junit/junit/4.13.2/junit-4.13.2.jar/__content__
d98a9a02a99a9acd22d7653cbcc1f31f  /data/storage/proxy-cache/.../__content__

$ cat /data/storage/proxy-cache/team-maven-proxy/junit/junit/4.13.2/junit-4.13.2.jar/__cache_meta__.json
{
  "cached_at":"2026-06-12T07:21:45.697682251Z",
  "upstream_etag":"\"d98a9a02a99a9acd22d7653cbcc1f31f\"",
  "expires_at":"2026-06-13T07:21:45.697682251Z",
  "content_type":"application/java-archive",
  "size_bytes":384581,
  "checksum_sha256":"8e495b634469d64fb8acfa3495a065cbacc8a0fff55ce1e31007be4c16dc57d3"
}
```

The MD5 matches the upstream ETag. The file is correct, world-readable, owned by the process user. The error log path is byte-for-byte identical to the on-disk path. The AK process can read the file from the shell (`cat __content__ | wc -c` → 384581).

## Why This Is in AK Code, Not Infrastructure

- The same container user (`uid=1001`) can read `__content__` via shell — no permissions issue.
- `md5sum __content__` succeeds — no NFS incoherence or corruption.
- The error is immediate and deterministic, not flaky — no race condition or NFS caching delay.
- The 500 occurs on every subsequent request across multiple artifacts (junit, commons-lang3, asm).
- S3-backed deployments are not affected (S3 backend returns the shared instance regardless of `storage_location.path`).

The bug is in the AK Rust code path that handles the cache-hit read for proxy/remote repositories on filesystem backends.

## Root Cause (from code analysis)

In v1.1.9, `ProxyService::cache_artifact` does two things:

1. Writes the cached content to `self.storage` (a `StorageService` backed by `FilesystemBackend` rooted at `STORAGE_PATH`, e.g. `/data/storage`). The `__content__` file lands at: `<STORAGE_PATH>/proxy-cache/<repo_key>/<path>/__content__`.

2. **Also inserts a row into the `artifacts` table** with `storage_key = "proxy-cache/<repo_key>/<path>/__content__"`.

On the second request, the Maven (and every other format) handler queries `artifacts` by path, finds the row, and tries to serve it via:

```rust
let storage = state.storage_for_repo(&repo.storage_location());
storage.get(&artifact.storage_key).await
```

`storage_for_repo` returns a **fresh `FilesystemStorage` rooted at `repo.storage_path`**. In v1.1.9, `repo.storage_path` is set to `<STORAGE_PATH>/<repo_key>` (e.g., `/data/storage/team-maven-proxy`). So the read resolves to:

```
/data/storage/team-maven-proxy/proxy-cache/team-maven-proxy/.../__content__
```

— a **doubled-prefix path** that does not exist. The write path used the global backend at `/data/storage/...`; the read path uses the per-repo backend at `/data/storage/<repo_key>/...`. These are two different directories.

## Severity

- **Affects all proxy/remote repos on filesystem-backed deployments** — Maven, npm, PyPI, Cargo, OCI, NuGet, and all other formats.
- **Every repeat fetch fails** with 500 until the operator manually purges the proxy cache.
- **CI pipelines are broken**: the first `mvn` build primes the cache, every subsequent run fails.
- **Not a flaky / race condition**: it is a deterministic, immediate failure on every cache-hit request.
- **NFS is not the issue**: the same failure reproduces on local-disk filesystem backends.

## Tested Artifacts (all affected)

- `org/junit/junit/4.13.2/junit-4.13.2.jar`
- `commons-lang/commons-lang/2.6/commons-lang-2.6.jar`
- `org/ow2/asm/asm/9.5/asm-9.5.jar`

kycheng:
## Summary

This PR adds a regression test that directly pins the storage-routing invariant broken by the v1.1.9 proxy-cache bug described in #1881.

**What the test does**: Writes a cached artifact via `cache_artifact` (which uses the global `StorageService` / `FilesystemBackend` rooted at `STORAGE_PATH`), then verifies:

1. The content IS readable via `get_cached_artifact_by_path` (the same global backend) — the production cache-hit path must work.
2. The content is NOT readable via a `FilesystemStorage` rooted at a per-repo subdirectory (simulating `repo.storage_path = "<STORAGE_PATH>/<repo_key>"`) — this is the doubled-prefix path that caused the 500s in v1.1.9.

**Background** (from #1881):

In v1.1.9, `cache_artifact` wrote the cached body to `self.storage` (global backend at `STORAGE_PATH`) and also inserted a row into `artifacts` with `storage_key = "proxy-cache/<repo_key>/<path>/__content__"`. On the next request, format handlers found that row and called `state.storage_for_repo(repo.storage_location()).get(&artifact.storage_key)`. On filesystem deployments, `storage_for_repo` returns a fresh `FilesystemStorage` rooted at `repo.storage_path` — typically `<STORAGE_PATH>/<repo_key>`. That resolves the key to a **doubled-prefix path**:

```
<STORAGE_PATH>/<repo_key>/proxy-cache/<repo_key>/.../__content__
```

…which does not exist. The file lives at:

```
<STORAGE_PATH>/proxy-cache/<repo_key>/.../__content__
```

The result was HTTP 500 / `STORAGE_ERROR` on every cache-hit request after the first fetch, affecting all proxy/remote repos on filesystem backends.

**The fix** was already applied in commit 63f1409 (v1.2.0-rc.2): remove the `INSERT INTO artifacts` from `cache_artifact`. The existing `test_cache_artifact_does_not_insert_into_artifacts_table` meta-test guards against the insert being re-introduced. This PR adds a complementary behavioral test that pins the storage-routing invariant directly.

## Test Plan

- [x] `cargo test --lib proxy_service::tests::test_proxy_cache_read_uses_global_storage_not_per_repo_storage` — 1 passed
- [x] `cargo test --lib proxy_service::tests::test_cache_artifact_does_not_insert_into_artifacts_table` — 1 passed
- [x] `cargo test --lib proxy_service::tests::test_cache_artifact_persists_bytes_for_next_request` — 1 passed

Fixes #1881

### DIFF
--- a/backend/src/services/proxy_service.rs
+++ b/backend/src/services/proxy_service.rs
@@ -8663,6 +8663,109 @@ SHA256:
         );
     }
 
+    /// Regression test for the doubled-prefix storage routing bug (v1.1.9).
+    ///
+    /// **Bug**: `cache_artifact` wrote cached content via the global `self.storage`
+    /// backend (a `FilesystemBackend` rooted at `STORAGE_PATH`) and also inserted
+    /// a row into the `artifacts` table with `storage_key =
+    /// "proxy-cache/<repo_key>/<path>/__content__"`.
+    ///
+    /// On the next request, every format handler found that row in the DB and
+    /// tried to serve the file via `state.storage_for_repo(repo.storage_location())`,
+    /// which returns a fresh `FilesystemStorage` rooted at
+    /// `repo.storage_path` (e.g. `/data/storage/team-maven-proxy`).
+    ///
+    /// That storage instance resolved the key to
+    /// `/data/storage/team-maven-proxy/proxy-cache/team-maven-proxy/.../__content__`
+    /// — a **doubled-prefix path** that does not exist on disk, because the file
+    /// lives at `/data/storage/proxy-cache/team-maven-proxy/.../__content__`.
+    ///
+    /// The result was HTTP 500 / `STORAGE_ERROR` on every cached read after the
+    /// first, affecting all 28+ format handlers on filesystem-backed deployments.
+    ///
+    /// **Fix (commit 63f1409, v1.2.0-rc.2)**: remove the DB insert from
+    /// `cache_artifact`. Cached items are served exclusively through
+    /// `proxy_check_cache` → `get_cached_artifact_by_path` → `self.storage.get`,
+    /// which always uses the same global backend that the write used. The
+    /// `storage_for_repo` dispatch path is never taken for proxy-cached content.
+    ///
+    /// This test pins the storage-routing invariant directly: the cache key
+    /// written by `cache_artifact` is readable via the same backend, and NOT
+    /// readable via a `FilesystemStorage` rooted at a per-repo subdirectory
+    /// (simulating the state of the `artifacts` table row's storage resolution).
+    #[tokio::test]
+    async fn test_proxy_cache_read_uses_global_storage_not_per_repo_storage() {
+        use crate::storage::filesystem::FilesystemStorage;
+        use crate::services::storage_service::{FilesystemBackend, StorageService};
+        use crate::storage::StorageBackend as _;
+
+        // Global storage root, mirroring STORAGE_PATH
+        let global_root = std::env::temp_dir().join(format!("ak-1278-global-{}", Uuid::new_v4()));
+        std::fs::create_dir_all(&global_root).expect("create global root");
+
+        // Per-repo storage root, mirroring repo.storage_path = "<STORAGE_PATH>/<repo_key>"
+        let repo_key = "team-maven-proxy";
+        let per_repo_root = global_root.join(repo_key);
+        std::fs::create_dir_all(&per_repo_root).expect("create per-repo root");
+
+        // Build ProxyService with the global backend (what production does)
+        let global_backend = Arc::new(FilesystemBackend::new(global_root.clone()));
+        let storage = Arc::new(StorageService::new(global_backend));
+        let pool = sqlx::PgPool::connect_lazy("postgres://invalid/").unwrap();
+        let proxy = ProxyService::new(pool, storage);
+
+        let artifact_path = "org/junit/junit/4.13.2/junit-4.13.2.jar";
+        let content = Bytes::from_static(b"fake-junit-jar-bytes");
+        let cache_key = ProxyService::cache_storage_key(repo_key, artifact_path).unwrap();
+        let metadata_key = ProxyService::cache_metadata_key(repo_key, artifact_path).unwrap();
+
+        // Write via cache_artifact (uses self.storage = global backend)
+        proxy
+            .cache_artifact(
+                &cache_key,
+                &metadata_key,
+                &content,
+                Some("application/java-archive".to_string()),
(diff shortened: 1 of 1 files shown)


kycheng:
Closing — was opened as a regression test for issue #1881, but the underlying bug is already fixed by [`63f1409`](https://github.com/artifact-keeper/artifact-keeper/commit/63f1409) on main. Not adding noise to your tree; if a regression test for this invariant would be welcome later, happy to re-open.

PR #1884, opened 2026-06-12
fix(maven): forward empty artifact path to upstream for proxy/group repos

### LINKED ISSUE #1880: Maven proxy/group: GET /maven/<repo>/ returns 404 while /maven/<repo>// transparently proxies to upstream root (empty artifact path handled inconsistently)
## Summary

For a Maven `proxy` (or `group` aggregating a proxy) repository, the AK router intercepts requests where the artifact path is **empty** and returns 404, but any other artifact path — including the bizarre value `/` produced by a trailing `//` — is transparently forwarded to the upstream. This makes the single corner case "empty artifact" inconsistent with the rest of the proxy's otherwise-transparent semantics.

Result: `GET /maven/<repo>/` returns 404, but `GET /maven/<repo>//` returns 200 (the upstream's `/maven2/` index page). Same logical endpoint ("the root of the proxied registry"), different responses based on how many trailing slashes the client happens to send.

## Environment

- AK version: **v1.1.9** (image tag \`artifact-keeper-backend:1.1.9\`)
- Test date: 2026-06-12
- Tested both via Ingress and direct pod port-forward — identical behavior, so this is in AK's HTTP routing, not in any intermediate proxy.

## Reproduction

Create a Maven \`remote\` (proxy) repo pointed at Maven Central:

\`\`\`bash
curl -u admin:<pwd> -X POST https://<ak>/api/v1/repositories -d '{
  "key":"team-maven-proxy",
  "format":"maven",
  "repo_type":"remote",
  "upstream_url":"https://repo1.maven.org/maven2/"
}'
\`\`\`

Then probe (all requests authenticated with valid creds):

| Request | AK response | Upstream (\`repo1.maven.org/maven2/...\`) response |
|---|---|---|
| \`GET /maven/team-maven-proxy/\` | **404** | 200 (HTML index of \`/maven2/\`) |
| \`GET /maven/team-maven-proxy//\` | **200** | 200 |
| \`GET /maven/team-maven-proxy///\` | 200 | 200 |
| \`GET /maven/team-maven-proxy/org/\` | 200 | 200 |
| \`GET /maven/team-maven-proxy/org/apache/\` | 200 | 200 |
| \`GET /maven/team-maven-proxy/org/junit/junit/\` | 404 | 404 (no such package) |
| \`GET /maven/team-maven-proxy/nonexistent-pkg-xyz/\` | 404 | 404 |

Every non-empty artifact path matches upstream exactly. Only the **empty artifact path** (single trailing \`/\`) is intercepted by AK and forced to 404, while \`\"/\"\` (trailing \`//\`) or any deeper path goes through transparently.

\`virtual\` (group) repos containing such a proxy behave the same way (proxy member's response surfaces).

\`local\` (hosted) repos return 404 for all of these — that part is fine since no upstream exists.

## Expected behavior

For a proxy repo, the empty artifact path should be treated the same as any other path: forward to upstream, return whatever upstream returns. Upstream Maven Central returns 200 with a directory index at \`/maven2/\`, so AK should return 200 as well. This would also align AK with the de-facto Nexus / Artifactory convention where \`GET /repository/<repo>/\` returns repo root metadata — many Nexus-compatible health-check tools rely on this.

For a local repo, returning 404 on \`GET /maven/<repo>/\` is also acceptable, **but** for consistency it would be better to return a 200 stub (e.g. a minimal directory listing or empty JSON descriptor) so that ecosystem tools that probe repo roots can verify the repo exists and credentials work regardless of repo type.

## Why this matters

1. **Semantic asymmetry inside a single repo**: \`<repo>/\` and \`<repo>//\` represent the same conceptual resource (\"registry root\") yet return different status codes. Clients that happen to render URLs with a single trailing slash get 404; clients that double up get 200.

2. **Nexus / Artifactory compatibility**: The Nexus convention is \`GET /repository/<repo>/\` → 200 (HTML or JSON index). AK markets itself as a drop-in alternative, but tooling built against this convention (health probes, repo discovery, credential verification) breaks on AK only for the empty-artifact case.

3. **Fragility of the current \"200\" path**: Today \`<repo>//\` accidentally returns 200 because the literal artifact \`/\` is forwarded to upstream which happens to serve \`/maven2//\` as its root. If upstream tightens slash handling, or if a downstream LB normalizes consecutive slashes before reaching AK, this 200 disappears with no replacement, and the repo root becomes completely unreachable.

## Suggested fix

In the router for proxy / group repos, treat empty artifact path the same as any other path — forward it to the upstream (concatenated as \`<upstream_url>\` + \`\"\"\` = \`<upstream_url>\` itself). For local repos, optionally return a 200 with a minimal repo-existence response.

The same routing issue likely affects \`npm\` (we observed \`GET /npm/<repo>/\` returning 404 for proxy/group/local across the board, including \`//\` variants; npm proxy/group don't even have the \"trailing-slash escape hatch\" that maven does — possibly a separate but related router gap).

## Impact on us

We hit this trying to integrate AK as a backend for a Nexus-compatible artifact connector. The connector's auth-verification probe hits \`<address>/\`, which for AK returns 404 on hosted repos and works on proxy/group only because of the \`//\` accident described above. Either fixing this in AK or being able to rely on a stable root-metadata endpoint would unblock the integration cleanly.

kycheng:
## Summary

- `GET /maven/<proxy-repo>/` (empty artifact path, single trailing slash) returned 404 while every other path against the same repo proxied transparently to upstream. `GET /maven/<proxy-repo>//` accidentally returned 200 only because the literal `/` path happened to hit the upstream root — making two URLs for the same conceptual resource return different status codes.
- Root cause: in axum 0.7, the `*path` wildcard in `/:repo_key/*path` does **not** match when the path after the repo key is empty. The framework 404-ed before the handler was ever called.
- This fix adds explicit `/:repo_key` and `/:repo_key/` routes backed by a `download_root` handler that, for proxy repos, fetches the upstream root URL and returns it. Local/hosted repos preserve their 404.

## What changed

- **`backend/src/api/handlers/maven.rs`**: Added two new routes (`/:repo_key`, `/:repo_key/`) to the maven router and a `download_root` async handler. The handler calls `ProxyService::fetch_artifact_with_cache_path(fetch_path="", cache_path="_root_")` — `fetch_path=""` resolves to the upstream root via `build_upstream_url(upstream_url, "")`, and the non-empty sentinel `"_root_"` satisfies `validate_cache_path`'s empty-string guard. Virtual (group) repos walk their members and return the root from the first responding Remote member.

## How tested

- `cargo test --lib "api::handlers::maven"` — 118 tests pass (2 new tests pinning route registration and cache-path sentinel).
- `cargo test --lib` — 10,555 unit tests pass, 4 ignored.
- `cargo build` — clean build, no warnings.
- `cargo fmt --check` — formatting unchanged.

Fixes #1880

### DIFF
--- a/backend/src/api/handlers/maven.rs
+++ b/backend/src/api/handlers/maven.rs
@@ -4,6 +4,7 @@
 //! `mvn dependency:resolve`.
 //!
 //! Routes are mounted at `/maven/{repo_key}/...`:
+//!   GET  /maven/{repo_key}      — Repository root probe (proxy/group → upstream root; hosted → 404)
 //!   GET  /maven/{repo_key}/*path — Download artifact, metadata, or checksum
 //!   PUT  /maven/{repo_key}/*path — Upload artifact (mvn deploy)
 
@@ -36,7 +37,16 @@ use crate::models::repository::RepositoryType;
 // ---------------------------------------------------------------------------
 
 pub fn router() -> Router<SharedState> {
-    Router::new().route("/:repo_key/*path", get(download).put(upload))
+    Router::new()
+        // Root probe: `/:repo_key/*path` (axum 0.7 wildcard) does NOT match
+        // when the path segment after the repo key is empty — i.e. a request
+        // for exactly `GET /maven/<repo>/`.  We register the bare key route so
+        // proxy and group repos can forward that root probe to their upstream.
+        // See download_root for details.  The trailing-slash variant is listed
+        // separately because axum treats `/x` and `/x/` as distinct routes.
+        .route("/:repo_key", get(download_root))
+        .route("/:repo_key/", get(download_root))
+        .route("/:repo_key/*path", get(download).put(upload))
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


brandonrc (changes requested):
Nice diagnosis — this is a real, correctly-scoped fix; requesting changes only on test coverage (the Code Coverage gate is red).

**Reproduced on `main`:** the Maven router registers a single `/:repo_key/*path` GET route (`maven.rs:39`) and there's no empty-path guard in `download`. The actual cause is framework-level: **axum 0.7's `*path` wildcard doesn't match an empty segment**, so `GET /maven/<repo>/` 404s in the router before any handler runs (#1880; the `<repo>//` → 200 vs `<repo>/` → 404 asymmetry confirms it). Adding `/:repo_key` + `/:repo_key/` → `download_root` is the right fix, and it's safely scoped: empty path → upstream root via `build_upstream_url(.., "")`, the `_root_` cache sentinel passes `validate_cache_path` (empty would be rejected), Remote/Virtual-only with Local/Staging → 404, and the upstream fetch reuses the existing SSRF-guarded path. No traversal/loop/SSRF concerns.

**What's blocking:** the two added tests are `include_str!` string-pin assertions (route registration + the `_root_` sentinel), so the new `download_root` body — the substantive proxy/virtual logic — is never executed, which fails the >=70% changed-lines coverage gate. Please add a behavioral test using the existing wiremock upstream harness (cf. the SSRF tests around `proxy_service.rs:9697+`): assert `GET /maven/<proxy-repo>/` → 200 from the upstream root and `GET /maven/<local-repo>/` → 404. That exercises the handler and should turn coverage green. No live repro needed — behavior is fully determinable from source.

PR #1885, opened 2026-06-12
feat(npm): route /-/<meta> requests to npm meta namespace handler

### LINKED ISSUE #1882: npm router treats /-/<meta> as package "-" / version "<meta>" — all /-/ping, /-/whoami, /-/search, /-/login endpoints return 404
## Summary

AK's npm router does not recognise the npm registry protocol's `/-/` **meta namespace**. Any request whose path starts with `/<repo>/-/<endpoint>` (where `<endpoint>` is not one of the already-handled audit paths) falls into the `/:package/:version` catch-all with `package="-"` and `version="<endpoint>"`, producing a spurious 404.

## Smoking-gun reproducer

```
$ curl -u admin:<pwd> https://<ak>/npm/team-npm-proxy/-/ping
{"code":"NOT_FOUND","message":"Version 'ping' not found for package '-'"}
```

The response body diagnoses the bug exactly: AK parsed `package="-"` and `version="ping"` instead of recognising `/-/ping` as a registry-level meta request.

**AK version:** v1.1.9 (`artifact-keeper-backend:1.1.9`)  
**Repo:** `team-npm-proxy` — Remote proxy configured against `https://registry.npmjs.org/`, valid Basic auth on every request.

## All affected `/-/` endpoints (same 404 mechanism)

| URL | What npmjs.org returns | What AK returns |
|-----|----------------------|-----------------|
| `/npm/<repo>/-/ping` | 200 `{}` | 404 (parsed as pkg `-` ver `ping`) |
| `/npm/<repo>/-/whoami` | 401 (auth challenge) | 404 (parsed as pkg `-` ver `whoami`) |
| `/npm/<repo>/-/v1/search?text=express` | 200 (search results) | 404 (parsed as pkg `-` ver `v1`) |
| `/npm/<repo>/-/all` | 200 (legacy catalog) | 404 (parsed as pkg `-` ver `all`) |
| `/npm/<repo>/-/user/org.couchdb.user:foo` | 200 (login flow) | 404 |

## Confirmation that the package data plane is fine

The same repo, same auth, same session:

| URL | AK returns |
|-----|-----------|
| `/npm/<repo>/express` | 200 (package metadata, 813 KB) |
| `/npm/<repo>/express/-/express-4.18.2.tgz` | 200 (tarball, 55,804 bytes, SHA-1 matches npm Inc.'s published shasum) |
| `/npm/<repo>/@types/node` | 200 (scoped package metadata, 11 MB) |

This is not a wholesale npm router failure. Only the `/-/<meta>` namespace is missing.

## Root cause

The npm router's catch-all route `/:repo_key/:package/:version` matches `/-/ping` with `package="-"` and `version="ping"` because no route explicitly handles the `/-/` prefix for non-audit requests. The security audit routes (`/-/npm/v1/security/advisories/bulk`, `/-/npm/v1/security/audits/quick`) were already added as explicit literal routes (issue #1400), but the broader `/-/<endpoint>` namespace was never handled.

## Impact

- `npm ping` fails for any AK-backed registry → CI health checks break.
- `npm whoami` fails → users cannot verify credentials are configured correctly.
- `npm search` fails → IDE integrations that show search results break.
- `npm login` / `npm adduser` fail → cannot publish.
- Auth-validation health probes from artifact connectors / proxies break (this is how the bug was discovered — a Nexus-compatible connector that probes `/-/ping` to check registry liveness).
- `npm install` itself still works (uses `/<pkg>` + `/<pkg>/-/<pkg>-<ver>.tgz` paths, both implemented). The bug is *partially* hidden: users doing fresh installs won't notice until they try `npm login`, `npm whoami`, or CI health probes.

## Recommended fix

Implement the npm `/-/<endpoint>` meta namespace by adding a wildcard route `/:repo_key/-/*rest` that fires before the `/:package/:version` catch-all:

- **Proxy / Group (Remote / Virtual) repos** — forward `/-/*` paths transparently to the upstream registry (which already serves them correctly). This is the minimal correct behaviour for proxy repos.
- **Local repos** — at minimum implement `/-/ping` (return `{}` 200) and `/-/whoami` (return the authenticated user). Other endpoints can return 501 initially; full implementation is a follow-up.

The existing audit routes (`/-/npm/v1/security/advisories/bulk`, etc.) already register as explicit literal paths and will continue to shadow the new wildcard — no change needed for those.

I will open a PR with at least the proxy/group forwarding fix and the `/-/ping` + `/-/whoami` stubs for local repos.

kycheng:
## Summary

Fixes #1882.

The npm registry protocol reserves `/-/` as a meta namespace for registry-level operations (`/-/ping`, `/-/whoami`, `/-/v1/search`, `/-/user/…`, etc.). Without an explicit route for this prefix, AK's `/:package/:version` catch-all matched `/-/ping` with `package="-"` and `version="ping"`, returning:

```json
{"code":"NOT_FOUND","message":"Version 'ping' not found for package '-'"}
```

This broke `npm ping`, `npm whoami`, `npm search`, `npm login`, and any health-check tooling that probes `/-/ping`.

## What changed

**`backend/src/api/handlers/npm.rs`** — single-file change:

1. **New route** `GET /:repo_key/-/*rest` registered before all `/:package` catch-alls. More-specific literal `/-/…` routes (audit endpoints, dist-tags) continue to shadow this wildcard unchanged.

2. **`proxy_npm_meta_get`** — helper that forwards a GET request to `<upstream_url>/-/<rest>` and returns the upstream response verbatim. Returns `None` on transport error so the caller can fall through to the local stub.

3. **`npm_meta_get`** handler:
   - **Remote repos** → forward to configured upstream.
   - **Virtual repos** → forward to the first reachable Remote member's upstream.
   - **Local / Staging repos** → built-in stubs:
     - `/-/ping` → `200 {}`
     - `/-/whoami` → `200 {"username":"<name>"}` (authenticated) or `401` (unauthenticated)
     - everything else → `501 Not Implemented` (full local support is a follow-up)

## Intentional scope limit

Full `/-/v1/search`, `/-/user/…`, and other endpoints for **local** repos are not implemented in this PR — they require indexing and auth plumbing that deserve their own issue. Proxy/group repos already get full coverage via transparent upstream forwarding, which is the common case.

## How tested

Four new unit tests in `mod tests`:

| Test | Covers |
|------|--------|
| `test_local_repo_meta_ping_returns_200` | Local repo: `/-/ping` → `200 {}` |
| `test_local_repo_meta_whoami_authenticated_returns_username` | Local repo + auth: `/-/whoami` → `200 {"username":"…"}` |
| `test_local_repo_meta_whoami_unauthenticated_returns_401` | Local repo, no auth: `/-/whoami` → `401` |
| `test_remote_repo_meta_ping_proxied_to_upstream` | Remote repo: `/-/ping` forwarded via wiremock mock upstream, response returned verbatim |
| `test_local_repo_meta_ping_does_not_produce_package_not_found_404` | Regression guard: `/-/ping` must not fall through to the `/:package/:version` catch-all |

Build: `cargo check --package artifact-keeper-backend` passes (0 errors).

### DIFF
--- a/backend/src/api/handlers/npm.rs
+++ b/backend/src/api/handlers/npm.rs
@@ -337,9 +337,10 @@ async fn proxy_npm_audit_post(
 /// fallback.
 async fn proxy_npm_meta_get(upstream_url: &str, meta_path: &str) -> Option<Response> {
     let base = upstream_url.trim_end_matches('/');
-    // meta_path already carries the leading slash from the `*rest` wildcard
-    // (axum includes the slash: `*rest` on `/-/ping` yields `/ping`).
-    let path_with_dash = format!("/-{}", meta_path);
+    // Reconstruct the `/-/<rest>` meta path. axum 0.7 wildcard captures do NOT
+    // include a leading slash (`*rest` on `/-/ping` yields `ping`, not `/ping`),
+    // so prepend `/-/` and strip any stray leading slash to be robust either way.
+    let path_with_dash = format!("/-/{}", meta_path.trim_start_matches('/'));
     let url = format!("{}{}", base, path_with_dash);
     let client = crate::services::http_client::default_client();
 
--- a/backend/src/api/handlers/npm.rs
+++ b/backend/src/api/handlers/npm.rs
@@ -63,6 +63,19 @@ pub fn router() -> Router<SharedState> {
             "/:repo_key/-/package/:package/dist-tags/:tag",
             put(dist_tags_put).delete(dist_tags_delete),
         )
+        // npm /-/ meta namespace: GET /npm/{repo_key}/-/*rest
+        //
+        // The npm registry protocol reserves `/-/` as a meta namespace for
+        // registry-level operations (ping, whoami, search, login, audit, etc.).
+        // Without an explicit catch-all here, requests like `/-/ping` fall into
+        // the `/:package/:version` catch-all with `package="-"` and
+        // `version="ping"`, producing a spurious 404 ("Version 'ping' not found
+        // for package '-'"). This route must appear before every `:package`
+        // catch-all so axum resolves it first. More specific `/-/…` routes
+        // registered above (audit endpoints, dist-tags) continue to shadow this
+        // wildcard because axum prefers literal segments over `*` wildcards.
+        // See the filed issue for the full reproducer and impact analysis.
+        .route("/:repo_key/-/*rest", get(npm_meta_get))
         // Scoped package tarball: GET /npm/{repo_key}/@{scope}/{package}/-/{filename}
         .route(
             "/:repo_key/@:scope/:package/-/:filename",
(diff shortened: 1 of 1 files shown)


brandonrc (changes requested):
Thanks for this — the diagnosis and routing are right, and the local-repo stubs work. But there's a real bug in the **proxy path** that's causing the red CI, so requesting changes.

**Reproduced the bug on `main`:** `GET /npm/<repo>/-/ping` matches the catch-all `/:repo_key/:package/:version` (`npm.rs:84`) as `package="-"`, `version="ping"` → `404 "Version 'ping' not found for package '-'"` (#1882). Your new `/:repo_key/-/*rest` route is correctly placed *before* the catch-alls, and `/-/` can't shadow scoped packages (`@scope/name`), so the routing change is sound.

**The defect:** in `proxy_npm_meta_get` the upstream path is built as
```rust
let path_with_dash = format!("/-{}", meta_path);   // comment assumes meta_path = "/ping"
```
but **axum 0.7.9 wildcard captures have no leading slash** — for `/:repo_key/-/*rest` on `/-/ping`, `rest = "ping"`, not `/ping` (see `rpm.rs:754`, which does `format!("packages/{}", pkg_path)` for exactly this reason). So this produces `/-ping` and the upstream URL becomes `<base>/-ping` instead of `<base>/-/ping`. That's why `test_remote_repo_meta_ping_proxied_to_upstream` fails (wiremock serves `/-/ping`, handler requests `/-ping` → 404 → assert 200 panics → Unit Tests + Coverage exit 101). The local ping/whoami tests pass because there's no slash to mangle.

**One-line fix:**
```rust
let path_with_dash = format!("/-/{}", meta_path.trim_start_matches('/'));
```
After that the proxy/virtual forward hits the right upstream path and the test should pass. Worth re-running `cargo test --lib` with a `DATABASE_URL` set locally, since these tests no-op without a DB.

PR #1888, opened 2026-06-12
fix(auth): cast roles.name to text in apply_role_mapping privilege probe

### LINKED ISSUE #1886: Federated (LDAP/OIDC/SAML) login fails with 500 since #1827: varchar[] vs text[] in apply_role_mapping
## Summary

Every federated login (LDAP, and by code inspection OIDC/SAML too) returns HTTP 500 `Database operation failed` on current main. Regression introduced by e64cc5c6 (#1827, closes #1821).

## Environment

- Backend built from main @ 9a3178c6, PostgreSQL 17 (also reproduced on 16)
- LDAP provider (FreeIPA), login via `POST /api/v1/auth/sso/ldap/:id/login`

## Error

Backend log:

```
Request error error=Database error: error returned from database: operator does not exist: character varying[] = text[] code="DATABASE_ERROR"
```

## Root cause

#1827 added a privilege-change probe to `apply_role_mapping` (`backend/src/services/auth_service.rs`), executed on every federated login:

```sql
WITH prev AS (
    SELECT is_admin AS was_admin,
           ARRAY(
               SELECT r.name FROM user_roles ur
               JOIN roles r ON r.id = ur.role_id
               WHERE ur.user_id = $1
               ORDER BY r.name
           ) AS prev_roles
    FROM users WHERE id = $1
)
SELECT
    COALESCE($2, prev.was_admin) IS DISTINCT FROM prev.was_admin
    OR prev.prev_roles IS DISTINCT FROM $3::text[]
FROM prev
```

`roles.name` is `VARCHAR(255)` (migration `002_roles.sql`), so `prev_roles` is `varchar[]`, while `$3` is bound as `text[]`. PostgreSQL has no `varchar[] = text[]` operator (scalar `varchar = text` resolves via implicit coercion, the array types do not), so `IS DISTINCT FROM` fails at plan time. Standalone repro on any AK database:

```sql
SELECT ARRAY(SELECT name FROM roles ORDER BY name) IS DISTINCT FROM ARRAY['x']::text[];
-- ERROR:  operator does not exist: character varying[] = text[]
```

This breaks on *every* schema produced by the shipped migrations — it is not data- or deployment-specific. CI did not catch it because the query is built with `sqlx::query_scalar` (no compile-time check) and no test exercises `apply_role_mapping` against a real database.

## Suggested fix

Cast the element inside the ARRAY subquery:

```sql
SELECT r.name::text FROM user_roles ur ...
```

plus a DB-gated regression test that calls `apply_role_mapping` (both the no-op and the privilege-change branch). Happy to submit a PR with both.

Refs: #1821, #1827; auth/SSO umbrella #1617.

inspired-geek:
## Summary

Every federated (LDAP/OIDC/SAML) login on current main fails with HTTP 500
"Database operation failed". The privilege-change probe added in #1827
(`apply_role_mapping`) compares the user's previous role list — built as
`ARRAY(SELECT r.name ...)` from `roles.name VARCHAR(255)` — `IS DISTINCT FROM`
a `text[]` bind. PostgreSQL has no `varchar[] = text[]` operator, so the
statement fails at plan time on every schema produced by the shipped
migrations. Cast the array elements to `text` so both sides are `text[]`.

Fixes #1886

`test_apply_role_mapping_probes_varchar_roles_without_db_error` (DB-gated,
same pattern as the surrounding tests): verified to fail on main with
`operator does not exist: character varying[] = text[]` and pass with the fix.
It also pins the watermark semantics: a no-op role sync must not bump
`privileges_changed_at`, an admin grant must.

## Test Checklist

- [x] Manually tested locally

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`,
`cargo test --workspace --lib` with DATABASE_URL: 11507 passed / 0 failed.
Changed-lines coverage 96.3% (52/54; the two uncovered lines are the
DB-unavailable skip arms shared by every DB-gated test in the module).

## API Changes

- [x] N/A - no API changes

### DIFF
--- a/backend/src/services/auth_service.rs
+++ b/backend/src/services/auth_service.rs
@@ -1846,7 +1846,10 @@ impl AuthService {
             WITH prev AS (
                 SELECT is_admin AS was_admin,
                        ARRAY(
-                           SELECT r.name FROM user_roles ur
+                           -- roles.name is VARCHAR(255); without the cast the
+                           -- IS DISTINCT FROM below is varchar[] vs text[],
+                           -- which Postgres has no operator for.
+                           SELECT r.name::text FROM user_roles ur
                            JOIN roles r ON r.id = ur.role_id
                            WHERE ur.user_id = $1
                            ORDER BY r.name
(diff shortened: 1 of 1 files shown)


brandonrc (approved):
**Approve** — verified the bug reproduces on `main` (d0ea5a6).

`roles.name` is `VARCHAR(255)` (`backend/migrations/002_roles.sql:7`, never altered), so the privilege probe's `prev_roles` is typed `character varying[]`, while it's compared `IS DISTINCT FROM $3::text[]` (`$3` binds from `mapping.roles: Vec<String>` → `text[]`). Postgres has no `varchar[] = text[]` operator, so the statement fails at plan time (`operator does not exist: character varying[] = text[]`) — and since it's `query_scalar` (no compile-time type check) with no DB-gated test on `apply_role_mapping`, CI never caught it. Result: **HTTP 500 on every federated (LDAP/OIDC/SAML) login** (#1886), reachable via `sync_federated_user` → `apply_role_mapping`.

The `::text` cast on a `varchar(255)` value is a no-op on content, so distinctness/ordering semantics are unchanged — it just aligns the array element types. The added DB-gated regression test exercises both the no-op and admin-grant paths and pins the `privileges_changed_at` watermark behavior. CI is green. 👍

PR #1900, opened 2026-06-14
fix(maven): sbt short-filename parse and storage fallback security gates

### LINKED ISSUE #1899: fix(maven): sbt cross-versioned plugin filenames fail to parse; storage fallback bypasses quarantine/soft-delete on primaries
## Problem

Two related bug families in Maven handling, both surfaced when using sbt cross-versioned plugins (e.g. `sbt-foo_2.12_1.0`) through virtual repositories.

---

### 1. `parse_filename` fails on sbt cross-versioned plugin short filenames

sbt publishes cross-versioned plugins under a directory named after the full cross-versioned artifact ID (e.g. `com/example/sbt-foo_2.12_1.0/2.0.4/`) but writes the **filename** using only the short base name without the cross-version suffix (e.g. `sbt-foo-2.0.4.jar`).

`parse_filename` builds `expected_prefix = "sbt-foo_2.12_1.0-2.0.4"` and fails to match `sbt-foo-2.0.4.jar`, returning a validation error. This causes `parse_coordinates` to fail for any sbt plugin coordinate parsed via the Maven handler, breaking virtual-repo resolution for sbt plugins.

**Secondary issue:** for exact-SNAPSHOT filenames (`sbt-foo-2.0.4-SNAPSHOT.jar`), the timestamp-snapshot prefix (`sbt-foo-2.0.4`) is tried before the full-version prefix, causing `-SNAPSHOT` to be parsed as a classifier instead of being part of the version.

---

### 2. `maven_local_fetch_storage_fallback` — security gate gaps

The storage fallback that serves companion files (`.pom`, `-sources.jar`, checksums, etc.) from virtual repos has three correctness issues:

**a) SQL LIKE wildcard injection**
The sibling-lookup query uses `path LIKE $2` where `$2` is derived from the artifact's GAV directory. sbt artifact IDs contain `_` (a SQL single-character wildcard), so `sbt-foo_2.12_1.0/2.0.4/` becomes a pattern that can match rows from adjacent coordinates.

**b) Wrong quarantine anchor in Gate 2**
Gate 2 selects a live sibling to anchor the GAV policy check using `ORDER BY created_at DESC`. If the newest sibling is a `-sources.jar` or other companion file rather than the primary `.jar`/`.pom`, quarantine is checked against that row instead of the authoritative primary — allowing a quarantined GAV to be served if a younger, clean companion was uploaded after quarantine was applied.

**c) Soft-delete and quarantine bypass on primaries (Gate 1.5 missing)**
The fallback previously refused all primary extensions (`.jar`, `.aar`, etc.) at Gate 1. To support GAV-grouped publishing (where AK parks the canonical row on the `.pom` and the `.jar` has no own row), Gate 1 now admits bare primary extensions when a live sibling anchors the GAV. However, without an explicit check on the primary's own row:
- A soft-deleted primary (retracted artifact) with a live sibling `.pom` would be served.
- A quarantined primary with a live, younger, CLEAN sibling (e.g. a distribution `.zip` in the same GAV) could be served — Gate 2's `ORDER BY created_at DESC` might select the clean sibling as anchor, and Gate 3 checks the anchor's quarantine, not the primary's.

---

## Expected behaviour

- `parse_coordinates` succeeds for sbt cross-versioned plugin paths regardless of whether the filename uses the full or short artifact ID prefix.
- Exact-SNAPSHOT filenames for sbt plugins are parsed with `classifier = None`, not `classifier = "SNAPSHOT"`.
- The sibling-lookup `LIKE` query escapes `_` and `%` metacharacters.
- Gate 2 prefers primary-extension rows as the quarantine anchor.
- Gate 1.5 checks `is_deleted` and `quarantine_status` on the primary's own row before falling through to Gate 2, so a clean sibling cannot anchor past a retracted or quarantined primary.

## Fix

PR: #1900

ysolomon-plat:
Closes #1899

### `backend/src/formats/maven.rs` — `parse_filename`
- sbt cross-versioned plugins publish under `artifact_2.12_1.0/` but the filename uses the short base name (`artifact-2.0.4.jar`). The parser now strips up to two trailing `_<version-segment>` components, handling single (`name_2.12`) and double (`name_2.12_1.0`) cross-version suffixes, so `parse_coordinates` succeeds for short-form sbt plugin filenames.
- Exact-SNAPSHOT misparse fixed: the full-version short prefix (including `-SNAPSHOT`) is tried before the timestamp-snapshot prefix, so `-SNAPSHOT` is never parsed as a classifier.
- Removed a dead `is_snapshot_timestamp` flag; `strip_snapshot_timestamp` is applied inline.
- Refactored the prefix-priority chain to a labeled block.

### `backend/src/api/handlers/maven.rs` — upload GAV grouping
- Promote an incoming primary artifact to the canonical GAV row whenever the existing anchor is **non-primary** (a POM *or* a classifier artifact), not only when it is a POM. Clients that publish a classifier artifact (e.g. `-tests-sources.jar`) before the main artifact previously left the main artifact rowless and the canonical row anchored on the wrong file.

### `backend/src/api/handlers/maven_proxy.rs` — `maven_local_fetch_storage_fallback`
- **LIKE wildcard injection**: artifact IDs can contain `_` (a SQL single-character wildcard). The sibling-lookup now uses the shared `escape_like_literal` helper with `ESCAPE '\'`.
- **Quarantine anchor**: Gate 2 prefers a **true primary** (primary extension, no classifier) as the anchor. Classifier jars (`-sources.jar`, `-tests.jar`, …) also end in `.jar`; without excluding them, a clean classifier jar with its own row (as produced by row-per-file layouts such as bulk migration imports) could anchor past a quarantined primary and leak the GAV's companion files.
- **Primary serving**: Gate 1 admits bare primary extensions so the fallback can serve a rowless primary when a live sibling anchors the GAV. `MAVEN_PRIMARY_FILE_EXTENSIONS` is aligned with `is_primary_maven_artifact` (adds `.bundle`/`.tar.gz`) so those packaging types are servable rather than 404.
- **Gate 1.5 — retraction/quarantine**: a primary with a soft-deleted or quarantined own row is refused before Gate 2, so a clean sibling cannot anchor past it. Implemented as a single atomic query (one round-trip; no soft-delete race between two reads).

### Tests
- Regression test: a quarantined true primary with a clean classifier-jar sibling in the same GAV must withhold its companions (the clean classifier jar must not anchor the quarantine check).

## Risk / rollback
- No schema changes.
- All changes tighten existing security gates or correct anchor selection; no existing behavior is loosened.
- The LIKE-escape and primary-extension definitions are consolidated to reduce drift between the upload and serving paths.

### DIFF
--- a/backend/src/api/handlers/maven.rs
+++ b/backend/src/api/handlers/maven.rs
@@ -2144,8 +2144,17 @@ async fn upload(
 
     match gav_existing {
         Some((existing_id, existing_path, existing_storage_key, existing_meta)) => {
-            // An artifact record already exists for this GAV.
-            let existing_is_pom = MavenHandler::is_pom(&existing_path);
+            // An artifact record already exists for this GAV. The anchor row is
+            // the first-created file in the directory, which — depending on the
+            // client's upload order — can be a POM or a classifier artifact
+            // (e.g. sbt publishes `-tests-sources.jar` before the main `.jar`).
+            // Determine whether that anchor is itself a primary packaging
+            // artifact so a later-arriving primary can be promoted over it.
+            let existing_coords = MavenHandler::parse_coordinates(&existing_path).ok();
+            let existing_is_primary = existing_coords
+                .as_ref()
+                .map(is_primary_maven_artifact)
+                .unwrap_or(false);
 
             let new_file = make_file_entry(
                 &path,
@@ -2156,15 +2165,18 @@ async fn upload(
                 &checksum_sha256,
             );
 
-            if is_primary && existing_is_pom {
-                // The existing record is a POM-only placeholder. Promote the new
-                // JAR/WAR to primary and demote the POM into the files list.
-                let old_pom_coords = MavenHandler::parse_coordinates(&existing_path).ok();
-                let old_ext = old_pom_coords
+            if is_primary && !existing_is_primary {
+                // The existing record is a non-primary placeholder: either a POM,
+                // or a classifier artifact that happened to be uploaded before the
+                // main artifact (sbt's publish order). Promote the new JAR/WAR to
+                // primary and demote the existing file into the files list, so the
+                // canonical row is always the main artifact regardless of the order
+                // files arrived in.
+                let old_ext = existing_coords
                     .as_ref()
                     .map(|c| c.extension.as_str())
                     .unwrap_or("pom");
-                let old_classifier = old_pom_coords
+                let old_classifier = existing_coords
                     .as_ref()
                     .and_then(|c| c.classifier.as_deref());
 
@@ -2181,7 +2193,7 @@ async fn upload(
                     .unwrap_or("")
                     .to_string();
 
-                let pom_file = make_file_entry(
+                let demoted_file = make_file_entry(
                     &existing_path,
                     old_ext,
                     old_classifier,
@@ -2196,7 +2208,7 @@ async fn upload(
                     .and_then(|f| f.as_array())
                     .cloned()
                     .unwrap_or_default();
-                files.push(pom_file);
+                files.push(demoted_file);
 
                 // Merge POM-parsed fields into the new primary metadata
                 let mut merged = file_metadata.clone();
--- a/backend/src/api/handlers/maven_proxy.rs
+++ b/backend/src/api/handlers/maven_proxy.rs
@@ -174,17 +174,18 @@ pub(crate) async fn maven_local_fetch_storage_fallback(
     // committed between two separate queries can't be observed inconsistently
     // (which would let the quarantine check be skipped).
     if is_primary {
-        let own = sqlx::query_as::<_, (bool, Option<String>, Option<chrono::DateTime<chrono::Utc>>)>(
-            "SELECT is_deleted, quarantine_status, quarantine_until \
+        let own =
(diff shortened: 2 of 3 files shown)


brandonrc:

brandonrc (approved):
Verified against fresh `main`; gave the storage-fallback security gates priority scrutiny. Confirmed at code level:
- sbt short-filename parse: real functional bug — `expected_prefix` never matches sbt cross-versioned coordinates; the `short_base` derivation fixes it and standard Maven still matches `expected_prefix` first (no regression).
- (a) SQL `LIKE` wildcard injection: real — sbt `_` acts as a single-char SQL wildcard, letting a sibling row from an adjacent GAV anchor the policy check; fixed with `escape_like_literal` + `ESCAPE`.
- (b) Wrong quarantine anchor: real bypass — `ORDER BY created_at DESC` anchors on a clean younger companion past a quarantined primary; fixed by preferring a true primary as anchor.
- (c) Primary own-row bypass: not reachable on current `main` (Gate 1 refuses all primaries); this PR opens primary serving and correctly fences it with Gate 1.5, using the same `is_deleted` + `check_download_allowed` gates as the canonical SQL path — parity confirmed.
Regression tests cover each gate; required CI is green. Non-blocking follow-up: generate the Gate 2 suffix `CASE` from `MAVEN_PRIMARY_FILE_EXTENSIONS` rather than inline SQL literals to remove the documented drift risk. Approving.

Issue #1889, opened 2026-06-12
Nexus Migration failed "Unknown repository type: hosted"


christopherfischer:
### Description

ERROR artifact_keeper_backend::services::migration_worker: Failed to prepare repository migration config; skipping job_id=0fc10e6b-15a2-4293-8e23-752fdeab686a repo=my-releases error=Configuration error: Unknown repository type: hosted

Nexus Repository:
```
{
  "name" : "my-releases",
  "format" : "maven2",
  "type" : "hosted",
  "url" : "https://my-nexus/repository/my-releases",
  "attributes" : { }
}
```

### Environment Information

- Deployment type: Docker

### Backend Version

1.2.0

### Web UI Version

1.2.0

### Package Format

None

### Expected Behavior

Migration of Repository

### Observed Behavior

Failed

### Steps to Reproduce

"job_type": "full",
    "config": {
      "include_repos": ["my-releases"],
      "conflict_resolution": "skip",
      "concurrent_transfers": 4,
      "throttle_delay_ms": 100,
      "verify_checksums": true
    }

### Logs

```shell

```

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have redacted any sensitive information (passwords, tokens, private URLs)
- [x] I am using a supported version (not end-of-life)

PR #1902, opened 2026-06-15
fix(migration): unblock Nexus migrations — empty include_repos + repo-type aliases

### LINKED ISSUE #1901: Migration with empty include_repos completes silently with 0/0/0 (Nexus and Artifactory)
### Description

When a migration job is created with an empty `include_repos` list (the default if the caller does not pass one), the worker reports the job as **`completed`** in ~200 ms with `total_items=0`, `completed_items=0`, `failed_items=0`, `skipped_items=0`, and writes **no `migration_items` and no `migration_reports`** rows.

The user-visible effect is the UI showing the job in `Completed` state with `0%` / `0/0` items and **no logs at all** to explain why. The source connection is verified, reachable, and contains repositories — so the silent no-op is surprising.

This was reproduced against a Nexus source on Server `1.2.1-dev (8b922f7)` / Web `1.2.0`, but the bug lives in source-agnostic code so it affects Artifactory sources too.

### Root cause

In `backend/src/services/migration_worker.rs::resolve_repos_for_provisioning`, the function iterates `requested` and looks each key up in `source_repos`. When `requested` is `[]`, the `for` loop is skipped entirely and an empty `ResolveRepoPlan` is returned. Back in `process_job`, `repos_to_process` ends up empty, the per-repo `for` loop is skipped, and the job goes straight to `determine_final_status(0, 0) = Completed`.

The current behavior is even pinned by `test_resolve_repos_empty_request_yields_empty_plan`, so the regression test suite locks the bad semantics in.

By convention for include/exclude filter pairs (apt, yum, Bazel, Helm, …) an **empty include list means "include everything"**, not "include nothing." The current code treats it as the latter.

### Expected behavior

A migration job submitted with `include_repos: []` should enumerate every repository returned by the source's `list_repositories()` and process them (subject to `exclude_repos` and supported-type filtering).

If, after all filtering, no repository remains to be migrated, the worker should at minimum:

- log a `WARN` explaining why (so the operator is not left to guess from the database),
- write a `migration_reports.warnings` entry so the UI can surface it.

### Steps to reproduce

1. Configure a source connection (Nexus or Artifactory) that has at least one repository.
2. `POST /api/v1/migrations` with body:
   ```json
   {
     "source_connection_id": "<uuid>",
     "job_type": "full",
     "config": {
       "dry_run": false,
       "conflict_resolution": "skip",
       "concurrent_transfers": 4,
       "throttle_delay_ms": 100,
       "verify_checksums": true,
       "include_repos": []
     }
   }
   ```
3. `POST /api/v1/migrations/<id>/start`.
4. Observe in `migration_jobs`:
   ```
   status=completed total_items=0 completed_items=0 failed_items=0 skipped_items=0 (~200 ms)
   ```
   and zero rows in `migration_items` and `migration_reports`.

### Backend version

`1.2.1-dev` / commit `8b922f7`

### Pre-submission checklist

- [x] I have searched existing issues to ensure this isn't a duplicate (#1889 is a different bug — Nexus `hosted` type mapping failure when the user *does* pass a specific repo)
- [x] I am using a supported version

kycheng:
Closes #1901 and #1889.

This PR bundles two small, independent fixes that both surface as
"submit a Nexus migration job; nothing happens." They live in the same
module and trip the same end-to-end flow, so they ship together.

## #1901 — empty include_repos treated as "migrate nothing"

`migration_worker::resolve_repos_for_provisioning(&[], ...)` walked
its for-loop over the empty `requested` slice, returned an empty
`ResolveRepoPlan`, and `process_job` walked straight to
`determine_final_status(0, 0) = Completed` — silent success in
~200 ms, no `migration_items`, no `migration_reports`. The UI
submits `include_repos: []` by default.

Fix: treat an empty `include_repos` as "every repository the source
reports" (apt / yum / Bazel / Helm convention). Unmappable source
repos still land in `unsupported` so they surface as errors. Also
emits a WARN when `repos_to_process` is empty after provisioning,
so future regressions of this class are observable.

## #1889 — Nexus repo type vocabulary rejected

`RepositoryType::from_artifactory` only matched the Artifactory
triple `local` / `remote` / `virtual`. Nexus reports
`hosted` / `proxy` / `group`, so every Nexus repo failed
`prepare_repository_migration` with `Unknown repository type: hosted`
(or `proxy` / `group`).

Fix: accept the Nexus aliases in `from_artifactory`. The enum doc
comments above the function had always paired the two vocabularies
("Local repository (hosted)" etc.), so the mapping was known to the
original author and just never wired up.

## Why each bug was sticky

Both bugs were pinned into CI by tests that froze the wrong behavior:

- `test_resolve_repos_empty_request_yields_empty_plan` asserted that
  an empty include list yields an empty plan.
- `test_repository_type_from_artifactory_unknown` asserted
  `from_artifactory("hosted") == None`.
- The three `normalizes_nexus_*` tests built fixtures with
  `repo_type: "local"` (Artifactory naming) rather than the `hosted`
  string a real Nexus client returns, so they never exercised the
  problem.

This PR updates the wrong-shaped tests and adds positive coverage:

- `test_resolve_repos_empty_request_resolves_all_source_repos`
- `test_resolve_repos_empty_request_routes_unsupported_source_repos_to_unsupported_bucket`
- `test_resolve_repos_empty_request_with_empty_source_yields_empty_plan`
- `test_repository_type_from_artifactory_accepts_nexus_aliases`
- `test_prepare_repository_migration_accepts_nexus_proxy`
- `test_prepare_repository_migration_accepts_nexus_group`

All 30 `repository_type | prepare_repository_migration | resolve_repos`
unit tests pass.

## Behavior change & risk

`include_repos: []` now means "migrate all source repos" rather than
"migrate nothing" — this matches what the UI's create-job flow already
sends. No internal caller intentionally relies on the empty-list
no-op.

The Nexus alias change is additive — Artifactory inputs keep working.

## Manual verification on a live cluster

Patched backend image rolled into the affected daily cluster:

- **Before fix**: `migration_jobs` row `completed`, 0/0/0, no logs.
- **After #1901 only**: `migration_jobs` row `failed`, 40+ ERROR
  lines from `prepare_repository_migration` rejecting every Nexus
  repo's `hosted` / `proxy` / `group` type — exactly #1889.
- **After both fixes**: worker enumerates every source repository,
  provisions destination repos, and per-artifact rows land in
  `migration_items` for the first time.

Bundled into a single PR because either fix in isolation still leaves
Nexus migration broken end-to-end, and the second fix exists in the
same module immediately downstream of the first.

### DIFF
--- a/backend/src/services/migration_service.rs
+++ b/backend/src/services/migration_service.rs
@@ -71,12 +71,21 @@ pub enum RepositoryType {
 }
 
 impl RepositoryType {
-    /// Parse from Artifactory repository type string
+    /// Parse a source-side repository type string.
+    ///
+    /// Accepts both the Artifactory vocabulary (`local` / `remote` /
+    /// `virtual`) and the Nexus vocabulary (`hosted` / `proxy` / `group`).
+    /// The two sets denote the same three logical kinds — the enum doc
+    /// comments above have always pinned this mapping. Prior to this fix
+    /// the function only matched the Artifactory triple, so every Nexus
+    /// repository was rejected by `prepare_repository_migration` with
+    /// `Unknown repository type: hosted` and an entire Nexus source was
+    /// effectively un-migratable (issue #1889).
     pub fn from_artifactory(rclass: &str) -> Option<Self> {
         match rclass.to_lowercase().as_str() {
-            "local" => Some(Self::Local),
-            "remote" => Some(Self::Remote),
-            "virtual" => Some(Self::Virtual),
+            "local" | "hosted" => Some(Self::Local),
+            "remote" | "proxy" => Some(Self::Remote),
+            "virtual" | "group" => Some(Self::Virtual),
             _ => None,
         }
     }
@@ -1317,9 +1326,14 @@ mod tests {
         // Regression test for issue #857: when Nexus reports `maven2`, the
         // prepared config should carry the canonical `maven` name so the
         // created repository uses the correct AK format.
+        //
+        // `repo_type` here is the real Nexus vocabulary (`hosted`) — not the
+        // Artifactory `local`. Prior fixtures used `local` and so masked
+        // issue #1889, where `RepositoryType::from_artifactory` rejected
+        // every Nexus repo on a live source.
         let repo = RepositoryListItem {
             key: "releases".to_string(),
-            repo_type: "local".to_string(),
+            repo_type: "hosted".to_string(),
             package_type: "maven2".to_string(),
             url: None,
             description: None,
@@ -1333,7 +1347,7 @@ mod tests {
     fn test_prepare_repository_migration_normalizes_nexus_yum() {
         let repo = RepositoryListItem {
             key: "yum".to_string(),
-            repo_type: "local".to_string(),
+            repo_type: "hosted".to_string(),
             package_type: "yum".to_string(),
             url: None,
             description: None,
@@ -1347,7 +1361,7 @@ mod tests {
     fn test_prepare_repository_migration_normalizes_nexus_raw() {
         let repo = RepositoryListItem {
             key: "resources".to_string(),
-            repo_type: "local".to_string(),
+            repo_type: "hosted".to_string(),
             package_type: "raw".to_string(),
             url: None,
             description: None,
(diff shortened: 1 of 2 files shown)


brandonrc:

brandonrc (approved):
Verified against fresh `main`. Both bugs reproduce at code level: (1) an empty `include_repos` silently yields "Completed 0/0" instead of migrating everything; (2) Nexus repo-type aliases (`hosted`/`proxy`/`group`) are rejected as "Unknown repository type". The fix makes empty mean "all" without breaking an explicit list, adds the aliases additively, and replaces the two tests that had frozen the wrong behavior with positive ones. Required CI is green. Note for the record: empty-means-all is an intentional behavior change (confirmed no automation relied on empty=no-op), and `exclude_repos` remains unimplemented — neither blocks. Approving.

PR #1907, opened 2026-06-16
fix: make Maven hosted repositories production-ready

valfon1393:
Fixes #1906 

## Summary

Makes hosted Maven repositories behave like production package repositories for uploaded Maven assets.

This PR fixes two related indexing gaps:

- hosted Maven uploads now populate the package catalog, so `/api/v1/packages` and the Web UI can show uploaded Maven packages and versions
- each uploaded non-sidecar Maven asset is persisted as its own active `artifacts` row by exact Maven path, instead of hiding secondary files inside one grouped GAV artifact

The Maven component/package view remains grouped by `groupId:artifactId`, while the physical artifact ledger now preserves the real uploaded files. That matters for exact-path repository APIs, scans, retention, download accounting, and future peer replication/sync behavior.

The implementation also keeps legacy `metadata.files` expansion as compatibility for old Maven rows, while avoiding duplicate listing rows when real artifact rows exist.

Reviewer note: this is intentionally broader than Package API visibility only. The first fix made packages visible, then deeper E2E/DB verification found that many valid Maven files were still missing exact `artifacts.path` rows. This PR fixes both issues together so Maven hosted repositories are ready as a Nexus/JFrog-like backend before HA/S3 work continues.

CI unblock note: this PR also includes a patch-only Wasmtime bump from `36.0.10` to `36.0.11` because GitHub CI started failing `cargo audit` after `RUSTSEC-2026-0182` was published. The dependency bump is not part of the Maven behavior change; it keeps this PR mergeable until `main` receives the same security update.

## Regression test (required for `fix/*` PRs)

- [x] This PR is a `fix/*` AND adds/updates a test that would have caught the bug
- [ ] N/A — this is not a bug fix

## Test Checklist

- [x] Unit tests added/updated
- [x] Integration tests added/updated (if applicable)
- [ ] E2E tests added/updated (if applicable)
- [x] Manually tested locally
- [x] No regressions in existing tests

Validation run locally:

```bash
cargo fmt --check
cargo test -p artifact-keeper-backend maven_upload --lib -- --nocapture
cargo test -p artifact-keeper-backend expand_maven_secondary_files --lib -- --nocapture
cargo test -p artifact-keeper-backend package_size_tracks --lib -- --nocapture
cargo clippy -p artifact-keeper-backend --lib -- -D warnings
git diff --check upstream/main...HEAD
```

CI unblock validation:

```bash
cargo update -p wasmtime-wasi --precise 36.0.11
cargo audit
[...]

PR #1909, opened 2026-06-16
feat(maven): populate packages table on artifact upload

### LINKED ISSUE #1910: Maven artifacts missing from Packages tab (packages table not populated)
### Description

After deploying Maven artifacts via `mvn deploy` (Maven 2 repository layout), the **Packages** tab in the Web UI shows "No packages in this repository yet." even though the artifacts are successfully stored and visible in the **Artifacts** tab. The packages API (`GET /api/v1/packages?repository_key=<key>`) returns an empty list.

This affects only the Maven/Gradle format. All other format handlers (PyPI, npm, NuGet, Cargo, Debian, Composer, etc.) correctly populate the `packages` and `package_versions` tables via `PackageService::try_create_or_update_from_artifact`.

### Environment Information

- Deployment type: Kubernetes
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL
- Reverse Proxy: (not configured)
- Network setup: (default)

### Backend Version

latest / main

### Web UI Version

latest / main

### Package Format

Maven / Gradle

### Expected Behavior

When a Maven artifact is uploaded via `PUT /maven/{repo_key}/*path`, a corresponding row should appear in the `packages` table so that:
1. The **Packages** tab in the Web UI shows the uploaded artifact (grouped by artifact name)
2. `GET /api/v1/packages?repository_key=<key>` returns the package in its response
3. Package-level features (version listing, drill-down views, install commands) work for Maven artifacts

### Observed Behavior

The artifacts are stored correctly in the `artifacts` table and the **Artifacts** tab shows them. However, no rows are created in `packages` or `package_versions` tables, so:

- The **Packages** tab shows "No packages in this repository yet."
- `GET /api/v1/packages?repository_key=<key>` returns `{"items":[],"pagination":{"page":1,"per_page":24,"total":0,"total_pages":0}}`
- Package-level UI features (drill-down with versions, install commands, file tree) are not available for Maven artifacts

### Steps to Reproduce

1. Deploy Artifact Keeper (latest/main) on Kubernetes
2. Create a Maven repository (hosted)
3. Upload a Maven artifact using `mvn deploy` or `curl -X PUT` to `/maven/{repo_key}/com/example/lib/1.0.0/lib-1.0.0.jar`
4. Open the Web UI for the repository
5. Click on the **Packages** tab
6. Observe "No packages in this repository yet." message
7. Switch to the **Artifacts** tab — the artifact is visible there
8. Call `GET /api/v1/packages?repository_key={repo_key}` — returns empty list

### Root Cause

The Maven upload handler at `backend/src/api/handlers/maven.rs` handles artifact storage directly by inserting rows into the `artifacts` table, but unlike every other format handler, it never calls `PackageService::try_create_or_update_from_artifact` to populate the `packages` and `package_versions` tables.

The fix is implemented in **PR #1909**.

### Fix (PR #1909)

**`backend/src/api/handlers/maven.rs`** — Added a best-effort call to `PackageService::try_create_or_update_from_artifact` in the `None` branch of the GAV lookup (when a new artifact record is created for a groupId:artifactId:version), after the artifact and its metadata are persisted. The call:

- Uses `coords.artifact_id` as the package name and `coords.version` as the package version
- Extracts the description from POM metadata (if the first uploaded file is a POM — standard Maven deploy behavior orders POM first)
- Passes `groupId` / `artifactId` as package metadata
- Is fire-and-forget (logs a warning on failure, never blocks the upload response), matching the pattern used by all other format handlers
- Is idempotent — `ON CONFLICT (repository_id, name) DO NOTHING` ensures subsequent files for the same GAV (e.g., JAR after POM) do not duplicate the package row

**`scripts/native-tests/test-maven.sh`** — Added a verification step that calls `GET /api/v1/packages?repository_key=test-maven` after a Maven deploy and asserts that the package appears with `total: 1` and the correct `name`.

### Why only the `None` branch?

The Maven handler groups multiple files for the same GAV under a single artifact record. The `None` branch fires exactly once per unique GAV — when the first file (typically the POM) is uploaded. Subsequent files for the same GAV hit the `Some` branch (artifact already exists) and do not need to create the package again, since `PackageService` already handled it via its UPSERT semantics.

### Additional Context

This is a missing feature rather than a regression — the Maven handler was never wired to `PackageService` when the packages abstraction was introduced. All other format handlers were updated at that time, but Maven was overlooked, likely because its path-based layout (`/maven/{repo_key}/groupId/artifactId/version/artifact.ext`) handles grouping differently than the API-based upload formats.

The problem is visible from the frontend Web UI (Packages tab), but the root cause is entirely in the backend — the `packages` and `package_versions` tables are never populated for Maven uploads.

Related PR: #1909

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this is not a duplicate

mymarche:
## Summary

The Maven upload handler was not calling PackageService after storing
artifacts, so the Packages tab always showed 'No packages in this
repository yet.' even when Maven artifacts existed in the repository.

Added a best-effort call to `PackageService::try_create_or_update_from_artifact`
in the None branch (new GAV artifact) to create/update the packages and
package_versions rows, matching the behavior of all other format handlers
(PyPI, npm, NuGet, etc.).

Also added a verification step in the Maven native test script to assert
the package appears in the packages API after upload.

Fixes #1910

### Test Checklist
- [ ] Manually tested locally

### API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/maven.rs
+++ b/backend/src/api/handlers/maven.rs
@@ -28,6 +28,7 @@ use crate::api::SharedState;
 use crate::error::AppError;
 use crate::formats::maven::{generate_metadata_xml, MavenCoordinates, MavenHandler};
 use crate::models::repository::RepositoryType;
+use crate::services::package_service::PackageService;
 
 // TODO: Remaining format handlers (beyond maven, npm, pypi, cargo) still use
 // plain-text error responses and should be migrated to AppError (#553).
@@ -2323,6 +2324,23 @@ async fn upload(
             .bind(&metadata)
             .execute(&state.db)
             .await;
+
+            // Populate packages / package_versions tables (best-effort)
+            let pkg_description = metadata.get("description").and_then(|v| v.as_str());
+            PackageService::new(state.db.clone())
+                .try_create_or_update_from_artifact(
+                    repo.id,
+                    &name,
+                    &coords.version,
+                    size_bytes,
+                    &checksum_sha256,
+                    pkg_description,
+                    Some(serde_json::json!({
+                        "groupId": coords.group_id,
+                        "artifactId": coords.artifact_id,
+                    })),
+                )
+                .await;
         }
     }
 
--- a/scripts/native-tests/test-maven.sh
+++ b/scripts/native-tests/test-maven.sh
@@ -92,6 +92,24 @@ fi
 
 echo "==> Artifact download returned HTTP $HTTP_CODE"
 
+# Verify packages API shows the uploaded artifact
+echo "==> Verifying package appears in packages API..."
+API_BASE="http://localhost:30080/api/v1"
+PACKAGES_JSON=$(curl -s -u "$REGISTRY_USER:$REGISTRY_PASS" \
+  "$API_BASE/packages?repository_key=test-maven")
+
+if ! echo "$PACKAGES_JSON" | grep -q '"total":1'; then
+  echo "FAIL: Expected 1 package in packages API, got: $PACKAGES_JSON"
+  exit 1
+fi
+
+if ! echo "$PACKAGES_JSON" | grep -q '"name":"test-artifact-native"'; then
+  echo "FAIL: Package 'test-artifact-native' not found in packages API"
+  exit 1
+fi
+
+echo "==> Packages API correctly shows the uploaded artifact"
+
 # Verify maven-metadata.xml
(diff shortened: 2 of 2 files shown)


mymarche:
Duplicate #1907

Issue #1365, opened 2026-05-24
Maven remote proxy can serve a zero-byte cached POM after Gradle resolution


junsung-cho:
### Description

A Gradle CI build failed while resolving Maven dependencies through an Artifact Keeper Maven remote repository.

During the incident, the Maven remote endpoint returned `200 OK` for a proxied POM URL but served an empty body (`Content-Length: 0`). Gradle then attempted to parse that response as a POM and failed with:

```text
Content is not allowed in prolog.
```

The upstream Maven Central POM was valid XML and non-empty. The specific artifact observed was `io.sentry:sentry:8.42.0`, but this appears to be a generic Maven remote proxy/cache issue rather than anything specific to Sentry.

The affected cache entry later appeared to self-heal or be evicted/refetched; a later request to the same Artifact Keeper URL returned the correct non-empty body.

### Environment Information

- Deployment type: Docker Compose
- Operating System: Ubuntu 24.04
- Architecture:  x86
- Database: PostgreSQL 16
- Reverse Proxy: Nginx
- Network setup: private

### Backend Version

1.2.0-rc.2

### Web UI Version

1.1.3

### Package Format

Maven / Gradle

### Expected Behavior

Artifact Keeper should not cache or serve an empty body for a non-empty upstream Maven POM.

If a Maven client performs HEAD/GET requests during dependency resolution, a HEAD request should not create a zero-byte proxy-cache content object or metadata sidecar that later GET requests treat as a valid cache hit.

### Observed Behavior

A fresh GitLab CI runner failed during Gradle dependency resolution:

```text
Could not parse POM https://<ak-host>/maven/<remote-repo>/<group>/<artifact>/<version>/<artifact>-<version>.pom
> Content is not allowed in prolog.
```

At the time of debugging, the Artifact Keeper Maven remote endpoint returned:

```text
HTTP/1.1 200 OK
Content-Type: text/xml
Content-Length: 0
```

The upstream Maven Central artifact for the same path returned valid XML with a non-empty body.

Example observed artifact:

```text
io/sentry/sentry/8.42.0/sentry-8.42.0.pom
```

For that example:

```text
Artifact Keeper Maven remote: 200 OK, Content-Length: 0
Maven Central: 200 OK, 1456 bytes, valid XML
```

A later request to the Artifact Keeper endpoint returned the correct non-empty body, so the bad cache entry may have expired, been evicted, or been overwritten by a successful GET.

### Steps to Reproduce

[...]

PR #1913, opened 2026-06-16
fix(proxy): reject truncated streamed cache writes (#1912)

### LINKED ISSUE #1912: Cache corruption ong PyPy, specifically Plotly
### Description

  PyPI wheel cache for plotly 6.8.0 is being poisoned again after cleanup. The cached wheel becomes invalid/truncated even though the upstream wheel is valid.

### Environment Information

  - Deployment type: Kubernetes
  - Operating System: Ubuntu 24.04/26.04 nodes
  - Architecture: amd64
  - Database: PostgreSQL
  - Reverse Proxy: none
  - Network setup: cluster internal

### Backend Version

current dev build (sha-8b922f7@sha256:eab4eb9f11646bace5f240133fee3b84012c4dac3126f4e2def6e87787b03cb9)

### Web UI Version

N/A

### Package Format

PyPI

### Expected Behavior

  The cached wheel should match the upstream plotly-6.8.0-py3-none-any.whl bytes and remain valid.

### Observed Behavior

 The cache entry is rewritten as a smaller invalid blob. The cached object no longer matches the upstream SHA256 and fails zip validation.

### Steps to Reproduce

  1. Request /pypi/pypi-remote/simple/plotly/plotly-6.8.0-py3-none-any.whl
  2. Let the proxy cache the wheel
  3. Re-read the cached object after some time or after a refresh (probably after the default 24hour TTL)
  4. Observe the cached object is invalid/truncated

### Logs

```shell
2026-06-16T14:34:59.422374Z WARN  GET /pypi/pypi-remote/simple/plotly/plotly-6.8.0-py3-none-any.whl
artifact_keeper_backend::services::proxy_service: Unexpected status 404 Not Found checking upstream https://pypi.org/simple/plotly/plotly-6.8.0-py3-none-any.whl, assuming changed

2026-06-16T14:34:59.422395Z INFO  GET /pypi/pypi-remote/simple/plotly/plotly-6.8.0-py3-none-any.whl
artifact_keeper_backend::services::proxy_service: Fetching artifact from upstream: https://pypi.org/simple/plotly/ (accept=None)

2026-06-16T14:34:59.538441Z INFO  GET /pypi/pypi-remote/simple/plotly/plotly-6.8.0-py3-none-any.whl
artifact_keeper_backend::services::proxy_service: Fetched 137815 bytes from upstream (content_type: Some("text/html"), etag: Some("\"YpJzWAsmg1OElHKe939Fbw\""), link: None)

2026-06-16T14:34:59.593542Z INFO  GET /pypi/pypi-remote/simple/plotly/plotly-6.8.0-py3-none-any.whl
artifact_keeper_backend::services::proxy_service: Fetching artifact from upstream (streaming): https://files.pythonhosted.org/packages/f9/14/abe5ce876ab5b66ee3c691bf537fcd43d037aea55d447aacf74630a8f31e/plotly-6.8.0-py3-none-any.whl

2026-06-16T14:34:59.708814Z INFO  GET /pypi/pypi-remote/simple/plotly/plotly-6.8.0-py3-none-any.whl
artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=0c1ae494-9f1a-4671-9446-8254c4e64f84 status=200

Downstream error:
Failed to download `plotly==6.8.0`
Failed to extract archive: plotly-6.8.0-py3-none-any.whl
I/O operation failed during extraction
unexpected BufError
```

### Additional Context

The cached object on disk had sha256 090e4272... and size 9689152, while upstream is 13c5c4a0... and size 9902055.

Around the plotly incident, the logs show the exact sequence:
direct wheel URL got a 404
artifact-keeper fell back to https://pypi.org/simple/plotly/
it fetched the simple page
then it fetched the wheel via streaming from files.pythonhosted.org
the request completed with 200

The important part is that artifact-keeper did not just serve an old cached object there. The backend log shows it fetched the upstream wheel stream again on 2026-06-16T14:34:59Z, which means the bad cache object was likely rewritten or re-poisoned during that refresh path.

### Pre-submission Checklist

GabrielSosaDevopsFarinter:
## Summary
The PyPI (and any) remote-proxy streaming path could poison its own cache with a **truncated** artifact. `CachePersister::tee_stream` wrote whatever bytes `put_stream` received and then committed the metadata sidecar **without comparing the byte count to the upstream `Content-Length`**. When a download was interrupted mid-body (e.g. `plotly-6.8.0-py3-none-any.whl`), the cache was overwritten with a short, SHA-mismatched blob and served to clients, which then failed zip/wheel extraction with `unexpected BufError`.

This adds a pure `classify_stream_write(bytes_written, expected_len)` guard and threads the upstream `Content-Length` into `tee_stream`. On a size mismatch (or a zero-byte body) the writer **skips the metadata sidecar and deletes the partial object**, so the next request refetches upstream and self-heals — exactly like the existing #1365 empty-body guard. Validation only fires when `Content-Length` is known, so chunked / auto-decompressed responses (which strip `Content-Length`) are never falsely rejected.

Fixes the cache-corruption reported in #1912.

Closes #1912

`test_tee_truncated_upstream_is_not_cached` fails on `main` (a truncated body was committed) and passes here (no sidecar, object deleted). Plus `test_tee_complete_upstream_with_matching_length_is_cached` and unit tests for `classify_stream_write`.

## Test Checklist
- [x] Manually tested locally

> Note: this contributor's local Windows/MSYS toolchain cannot build `aws-lc-sys` (BoringSSL asm), so `cargo clippy`/`test` were not run locally — relying on CI (Linux) for the full gate. `cargo fmt --check` passes.

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1329,36 +1329,36 @@ dependencies = [
 
 [[package]]
 name = "cranelift-assembler-x64"
-version = "0.123.10"
+version = "0.123.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "de2be1bdbf929c2a2242cbbe15d6583c56f1cc723c6c8452d0179362de28c9d5"
+checksum = "2ad3e7d446bbdc05f33371b17aae2857e2c5f1dd6750848f555c129c5060725f"
 dependencies = [
  "cranelift-assembler-x64-meta",
 ]
 
 [[package]]
 name = "cranelift-assembler-x64-meta"
-version = "0.123.10"
+version = "0.123.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9a0336914de11298290783a95a9a7154b894da601659eb5f8f8bc62d1bea98f8"
+checksum = "96ee145f9cbf7474350ee48e5c33ec2312177fa32c07e4eb395727250892a201"
 dependencies = [
  "cranelift-srcgen",
 ]
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.123.10"
+version = "0.123.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fb972cba51a52c1b2a329fec993b911e4d1f9cfab3795811a319b6746c28e014"
+checksum = "7025668e34c499fd0960dfaf68bc1689c961002fb67227f3dd262535c4118e52"
 dependencies = [
  "cranelift-entity",
 ]
 
 [[package]]
 name = "cranelift-bitset"
-version = "0.123.10"
+version = "0.123.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "642c920666bfed9aebca39d8c6e7cb76f09314cc7a4074b1db5edcccdde771b9"
+checksum = "4ad0b552e9b906fee4416b68664463e80f9024215da2cebd4065eee99d5d3008"
 dependencies = [
  "serde",
  "serde_derive",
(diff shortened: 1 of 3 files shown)


brandonrc:

brandonrc (approved):
Verified against fresh `main`. Reproduced the bug at code level: `CachePersister::tee_stream` commits a cleanly-terminated-but-short upstream stream as a valid cache entry — there is no `bytes_written` vs `Content-Length` reconciliation, so a truncated 200 response poisons the cache. The fix adds `classify_stream_write`, threads `upstream.content_length`, and rejects + deletes the partial entry (self-healing on next fetch). Added regression test fails on `main` and passes with the fix; required CI is green. No material risk. Approving.

Issue #1911, opened 2026-06-16
OIDC login: /me returns 401 immediately after callback due to token self-invalidation race condition


mymarche:
### Description

After completing an OIDC login flow (Microsoft Entra), the `GET /api/v1/auth/me` endpoint returns 401 Unauthorized when called immediately after the callback redirect. The problem self-resolves after approximately 5 seconds, after which `/me` works normally.

### Environment Information

- Deployment type: Kubernetes
- Operating System: Linux (container)
- Architecture: amd64
- Database: PostgreSQL 15+
- Reverse Proxy: N/A (direct)
- Network setup: Internal cluster

### Backend Version

dev (main branch, post-v1.2.0)

### Web UI Version

N/A

### Package Format

N/A / General

### Expected Behavior

After a successful OIDC login, the freshly-issued JWT access token should be accepted immediately by the auth middleware. `GET /api/v1/auth/me` should return 200 OK with the user profile.

### Observed Behavior

`GET /api/v1/auth/me` returns 401 Unauthorized for ~5 seconds post-login, then starts working. The issue occurs for:

- **New users** (first OIDC login) — always affected
- **Returning users whose role mapping changed** — affected when `privileges_changed_at` is bumped

### Steps to Reproduce

1. Configure an OIDC provider (Microsoft Entra) in Artifact Keeper SSO settings
2. Initiate a login via `GET /api/v1/auth/sso/oidc/{provider}/login`
3. Complete the OIDC authorization flow (redirect to Microsoft Entra, authenticate, redirect back)
4. Immediately call `GET /api/v1/auth/me` with the access token received from the callback
5. Observe 401 Unauthorized
6. Wait ~5 seconds
7. Call `GET /api/v1/auth/me` again with the same token
8. Observe 200 OK with user profile

### Root Cause

The bug is a race condition in the in-memory credential invalidation cache in `backend/src/services/auth_service.rs`.

**The flow:**
1. `sync_federated_user` → `apply_role_mapping` detects a privilege change (always true for new users) → calls `invalidate_user_tokens(user_id)` at line 1944
2. `invalidate_user_tokens` (line 184-186) sets an in-memory watermark to `Utc::now().timestamp().saturating_add(1)` — current second **+ 1**
3. `generate_tokens` (line 1755) creates a JWT with `iat = Utc::now().timestamp()` — current second (truncated)
[...]

PR #1917, opened 2026-06-17
fix(proxy): presign proxy-cache downloads via the no-prefix backend instead of streaming (#1555)

### LINKED ISSUE #1555: feat(proxy): extend S3 presigned-URL redirects to Virtual repository downloads
## Problem Statement

`S3_REDIRECT_DOWNLOADS=true` has no effect when artifacts are fetched through a **Virtual** repository. Virtual repo downloads go through `resolve_virtual_download_streaming` (landed in #1334), which correctly avoids heap buffering but still proxies the full response body through the backend — holding an Akka HTTP dispatcher thread open for the entire duration of the download.

For large artifact downloads (50–500 MB JARs, wheels, tarballs) this means each concurrent download occupies a server thread for 5–30 seconds. Under burst traffic (e.g. a CI fleet doing cold-cache dependency resolution), the dispatcher saturates and all other requests — including lightweight HEAD operations used for dependency resolution — queue behind the long-running streams. The effect is that p99 latency for HEAD requests degrades from ~0.2s to 30–60s, causing cascading build timeouts across the entire consumer fleet.

The presigned URL fast-path introduced in #1055 already solves this for **direct Remote** repo downloads: a metadata-only freshness probe + `StorageBackend::exists` check + presigned 302 redirect means the thread is released in milliseconds. But `resolve_virtual_download_streaming` never calls this path — it always streams the body through the backend regardless of the `S3_REDIRECT_DOWNLOADS` setting.

## Environment

- Storage backend: S3-compatible (confirmed on AWS S3 + MinIO)
- `S3_REDIRECT_DOWNLOADS=true`
- Affected request path: Virtual repository GET (e.g. `/maven/:virtual_repo_key/*path`, `/pypi/:virtual_repo_key/*path`)
- Not affected: direct Remote repository GET (presigned redirect works correctly)

## Observed Behaviour

With `S3_REDIRECT_DOWNLOADS=true` and a virtual repository whose members include S3-backed local and remote repos:

```
GET /maven/my-virtual-repo/com/example/my-lib/1.0/my-lib-1.0.jar
  → resolve_virtual_download_streaming
    → iterates remote members → cache hit on S3-backed local member
    → streams full artifact body through backend (200-300 MB, 10-25s)
    → response header: X-Artifact-Storage: proxy  ← presigned redirect never attempted
```

Expected:

```
GET /maven/my-virtual-repo/com/example/my-lib/1.0/my-lib-1.0.jar
  → resolve_virtual_download_streaming
    → iterates members → cache hit on S3-backed local member
    → is_cache_fresh check (metadata only) + StorageBackend::exists (HEAD)
    → 302 Location: https://bucket.s3.amazonaws.com/...?X-Amz-Signature=...
    → response header: X-Artifact-Storage: redirect-s3  ← thread released immediately
```

## Proposed Solution

Add a presigned-redirect fast-path inside `resolve_virtual_download_streaming` that mirrors the logic from `try_proxy_cache_redirect` (#1055) applied per-member:

```rust
// Inside resolve_virtual_download_streaming, after finding the winning member:
if presigned_enabled {
    let cache_key = build_cache_key(repo_key, path);
    if proxy_service.is_cache_fresh(repo_key, path).await {
        if let Ok(storage) = state.storage_for_repo(winning_member_key) {
            if let Some(redirect) = try_proxy_cache_redirect(
                storage.as_ref(), &cache_key, presigned_enabled, expiry, true
            ).await {
                return Ok(redirect);
            }
        }
    }
}
// Fall through to streaming path if presigned not available or cache stale
```

The streaming fallback (#1334) remains for:
- Filesystem backends (no presigned URL support)
- `S3_REDIRECT_DOWNLOADS=false`
- Cache-miss paths where the artifact must be fetched upstream before redirecting
- Local / Staging members (body already bounded by local storage)

## Relation to Existing Work

| Issue / PR | Scope | Status |
|---|---|---|
| #1018 / #1055 | Presigned redirect for direct Remote downloads | ✅ Fixed in v1.2.0 |
| #1215 / #1334 | Streaming (non-buffered) Virtual downloads | ✅ Fixed in v1.2.0 |
| #1020 | Presigned redirect for PyPI format handler | 🔲 Milestone v1.3.0 |
| **This issue** | Presigned redirect for Virtual repo resolution path | 🔲 Not tracked |

The missing piece is specifically the intersection of Virtual resolution + presigned redirect. #1334 fixed the OOM class; this issue addresses the thread-saturation class that #1334 leaves open for S3 deployments.

## Impact

On S3-backed deployments with `S3_REDIRECT_DOWNLOADS=true` and virtual repositories serving large artifacts at scale, the backend pod count becomes the binding constraint on download concurrency rather than S3 throughput. Increasing pod count is a valid short-term mitigation but the correct fix is to make virtual downloads thread-cheap via presigned redirects, matching the behaviour already available on direct Remote downloads.

ysolomon-plat:
## Problem
Virtual-repo proxy downloads stream the full artifact body through the backend instead of redirecting to object storage. For large artifacts (multi-hundred-MB to multi-GB, e.g. ML wheels), streaming holds a worker for the entire transfer; under concurrent load the dispatcher saturates and downloads start returning 502.

## Issues found and addressed
1. **Streaming large bodies -> 502 under burst** (the original report). Fresh proxy-cache hits should redirect to a presigned object-storage URL instead of streaming.

2. **Presign signed the wrong key when a storage key prefix is configured.** Proxy-cache content is written and read through the proxy's `StorageService`, which `StorageService::from_config` builds with no key prefix, so cached objects live at `proxy-cache/<repo>/<path>/__content__`. The presign path, however, signed via `state.storage_for_repo(...)`, which applies the configured `S3_PREFIX`, producing a signed key of `<prefix>/proxy-cache/...` that does not exist -> the redirect 404s. Streaming masked this because it reads through the no-prefix proxy handle. Deployments without a prefix would not hit it, which is why it was easy to miss.

3. **Capability was tracked by construction path, not the type.** Because the facade `StorageBackend` trait did not expose the presign methods, presign-capability was carried in a side-channel field set only on some construction paths. A handle built another way silently returned "no presign" and fell back to streaming, with no compile error. This is the same failure class as (2).

4. **A naive redirect helper on the PyPI path negative-cached.** Routing PyPI through the generic `proxy_fetch_or_redirect` calls `proxy_fetch` with `upstream/simple/<pkg>/<file>`, but PyPI serves the index at `/simple/<pkg>/` and hosts files on a separate CDN, so that URL 404s and gets negative-cached, wedging the package. The PyPI path must resolve the real file URL via the simple index (as `fetch_from_pypi_remote_streaming` does).

## Fix
- Sign proxy-cache presigns via the proxy's own no-prefix storage handle (the same handle `is_cache_fresh` already probes), not the prefixed per-repo handle. `from_config` builds that handle from the env-derived S3 config (so it carries redirect setting, presign credentials, CloudFront, and path format) with the prefix set to `None`.
- Move `supports_redirect` / `get_presigned_url` onto the facade `StorageBackend` trait, forwarded by the wrapper macro, so capability is type-enforced and no call site can silently fall back to streaming. The side-channel field is removed.
- In mixed handlers, only `proxy-cache/`-prefixed keys switch to the no-prefix handle; hosted artifacts keep the prefixed handle (their presign was already correct).
- The cheap `supports_redirect()` check runs before the freshness probe, avoiding a wasted metadata round-trip when redirect is unavailable.
- The PyPI remote path keeps index-based URL resolution and a streaming fallback; it does not reintroduce the wrong-URL fetch.

## Validation
- Unit and source-guard tests for the request-path wiring and the no-prefix signing.
- A live integration test (ignored, gated on an env var + a bucket) that builds the proxy-cache handle as `from_config` does, puts a `proxy-cache/...` object, presigns via the backend's own signer, and asserts the signed key carries no prefix and that object storage serves it 200.
- Full end-to-end against real object storage and real upstream PyPI: a cold `GET` proxies and caches at `proxy-cache/...` (no prefix); the cache-hit `GET` returns a 302 to a presigned URL with the verbatim no-prefix key; following it returns 200 with the artifact bytes, served by object storage (one redirect), not streamed through the backend.

## History note
The branch contains the original streaming-to-redirect commit, a revert (to restore safe streaming while the presign handling was reworked), and the final no-prefix presign commits. Can squash before merge.

## Related
#1911 / #1915 (negative-cache `now + 1` watermark race) - related but distinct.

Fixes #1555.

### DIFF
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -686,11 +686,19 @@ pub async fn proxy_fetch_or_redirect(
     // cache is already fresh, redirect to the signed URL without ever
     // pulling the cached body into the backend's memory. The freshness
     // probe is metadata-only (HEAD-equivalent on cloud backends).
-    if presigned_enabled && proxy_service.is_cache_fresh(repo_key, path).await {
-        // #1555: proxy-cache content is stored without the global key prefix,
-        // so it must be signed through the proxy's own (no-prefix) backend, not
-        // the prefixed repo handle, or the signed key 404s in the object store.
-        if let Some(storage) = proxy_service.cache_storage_backend() {
+    //
+    // #1555: resolve the no-prefix presign handle FIRST and skip the
+    // freshness probe entirely if we can't redirect (no handle, or the
+    // backend doesn't support redirects). The probe loads the cache-meta
+    // sidecar; on a backend that can't presign it would be a pure wasted
+    // S3 GET, since the slow path below re-reads the same sidecar anyway.
+    if presigned_enabled {
+        let storage = proxy_service.cache_storage_backend();
+        if storage.supports_redirect() && proxy_service.is_cache_fresh(repo_key, path).await {
+            // proxy-cache content is stored without the global key prefix,
+            // so it must be signed through the proxy's own (no-prefix)
+            // backend, not the prefixed repo handle, or the signed key
+            // 404s in the object store.
             if let Some(redirect) = try_proxy_cache_redirect(
                 storage.as_ref(),
                 &cache_key,
@@ -716,13 +724,18 @@ pub async fn proxy_fetch_or_redirect(
     if presigned_enabled {
         // #1555: sign the just-populated cache entry through the proxy's
         // no-prefix backend (same handle that wrote it), not the prefixed
-        // repo handle.
-        if let Some(storage) = proxy_service.cache_storage_backend() {
-            if let Some(redirect) =
-                try_presigned_redirect(storage.as_ref(), &cache_key, true, expiry).await
-            {
-                return Ok(redirect);
-            }
+        // repo handle. The entry was just written, so treat it as fresh.
+        let storage = proxy_service.cache_storage_backend();
+        if let Some(redirect) = try_proxy_cache_redirect(
+            storage.as_ref(),
+            &cache_key,
+            presigned_enabled,
+            expiry,
+            /* cache_is_fresh = */ true,
+        )
+        .await
+        {
+            return Ok(redirect);
         }
     }
 
@@ -749,17 +762,48 @@ pub async fn proxy_fetch_or_redirect(
 ///
 /// Extracted from `proxy_fetch_or_redirect` so the redirect short-circuit can
 /// be exercised in unit tests with recording mock storage backends.
-pub(crate) async fn try_proxy_cache_redirect<S: crate::storage::StorageBackend + ?Sized>(
+///
+/// Generic over the facade `storage_service::StorageBackend` trait (#1555):
+/// proxy-cache presigns flow through the single no-prefix backend handle, which
+/// carries presign capability type-enforced on the facade trait — not a
+/// side-channel field. The redirect is built inline (mirroring
+/// `try_presigned_redirect`) since that helper is bound to the inner storage
+/// trait.
+pub(crate) async fn try_proxy_cache_redirect<
+    S: crate::services::storage_service::StorageBackend + ?Sized,
+>(
     storage: &S,
     cache_key: &str,
     presigned_enabled: bool,
     expiry: Duration,
     cache_is_fresh: bool,
 ) -> Option<Response> {
-    if !presigned_enabled || !cache_is_fresh {
+    if !presigned_enabled || !cache_is_fresh || !storage.supports_redirect() {
         return None;
     }
-    try_presigned_redirect(storage, cache_key, true, expiry).await
+    match storage.get_presigned_url(cache_key, expiry).await {
+        Ok(Some(presigned)) => {
+            tracing::debug!(
+                key = %cache_key,
+                source = ?presigned.source,
+                expiry_secs = expiry.as_secs(),
+                "Serving proxy-cache artifact via presigned redirect"
+            );
+            Some(
+                crate::api::download_response::DownloadResponse::redirect(presigned)
+                    .into_response(),
+            )
+        }
+        Ok(None) => None,
+        Err(e) => {
+            tracing::warn!(
+                key = %cache_key,
+                error = %e,
+                "Failed to generate proxy-cache presigned URL, falling back"
+            );
+            None
+        }
+    }
 }
 
 /// Check whether an artifact is present in the proxy cache under `path`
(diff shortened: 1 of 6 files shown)


ysolomon-plat:
Missed the PyPI-specific download path.

The original fix only covered `resolve_virtual_download_streaming` in `proxy_helpers.rs`, which is used by Maven, npm, and other formats. The PyPI handler (`serve_file` in `pypi.rs`) has its own virtual-member loop that called `local_fetch_by_path_suffix` and streamed directly — bypassing the presigned redirect entirely. Verified against a live deployment: requesting a torch wheel via the virtual PyPI repo still returned `200 + content-length: 899MB` (streaming) after the original fix deployed.

Added `local_fetch_or_redirect_by_suffix` to `proxy_helpers` and swapped the PyPI virtual-member local fetch to use it. Same presigned-redirect logic, same S3 flag gate.

ysolomon-plat:
Squashed and updated. The pypi.rs remote-member proxy cache path is now fixed too — replaced the explicit presign block + proxy_check_cache_streaming pair with a single proxy_fetch_or_redirect call. Eliminates the double is_cache_fresh round-trip on presign-miss and reduces the added code surface.

ysolomon-plat:
### Code Duplication gate: false positive (this PR adds no duplicated code)

The duplication gate fails at **3.50% vs the 3.0% threshold**. I reproduced jscpd locally on the same changed-file set and confirmed **none of the duplicated blocks come from this PR's additions**:

- 0 clones overlap the new code (`pypi_proxy_cache_redirect`, `ProxyService::cache_storage_backend`/`is_proxy_cache_key`, the no-prefix presign call sites).
- The 791 duplicated lines are **pre-existing** intra-file duplication in the large files this PR happens to touch: `proxy_service.rs` ~528, `proxy_helpers.rs` ~233, `s3.rs` ~88 (largely test code). They are in scope only because the gate diffs whole changed files.

This matches the failure mode the gate's own comment documents (it "fails on code it never touched ... identical clones from main"; cf. #1824/#1825/#1827/#1796). Bringing this under 3% would require refactoring pre-existing duplication in files this PR does not functionally change, which adds regression risk for no behavioural benefit.

Requesting a maintainer bypass / admin-merge on the duplication gate for this PR. All other checks (Check Rust, Backend Unit Tests, Shell Tests, Security Audit, linked-issue) pass. Happy to squash the commits before merge.

PR #1932, opened 2026-06-19
fix(npm): serve abbreviated metadata and compress responses (#1931)

### LINKED ISSUE #1931: npm proxy returns full uncompressed packuments (ignores abbreviated/corgi metadata), making installs slow
## Summary

The npm registry proxy returns the **full packument** and ignores npm's abbreviated ("corgi") metadata content negotiation, and it serves package metadata **uncompressed**. As a result, `npm`/`pnpm`/`yarn` download roughly 10-100x more metadata bytes per package from an Artifact Keeper npm proxy than from `registry.npmjs.org`, which makes `install`/resolution slow. The gap is dramatic for clients that are not co-located with the AK instance (e.g. developer laptops or CI in a different region), where transfer time is dominated by bytes-on-the-wire.

`registry.npmjs.org` serves an abbreviated install document (only the fields the installer needs) and gzip/brotli-compressed when the client asks. AK does neither, so the same logical content is far larger on the wire.

## Two concrete behaviours

Against an AK npm proxy/virtual repo that proxies `registry.npmjs.org`:

1. **Abbreviated ("corgi") metadata is not honoured.** Clients send `Accept: application/vnd.npm.install-v1+json` to request the slim install document. AK ignores it and returns the full packument with `content-type: application/json`.
2. **Metadata is served uncompressed.** Even with `Accept-Encoding: gzip, br`, the response has no `Content-Encoding`; the full JSON is sent raw.

## Steps to reproduce

Against any AK instance with an npm remote/virtual repo proxying `registry.npmjs.org` (replace `AK` with your base URL and `REPO` with the repo key):

```sh
AK=https://your-artifact-keeper.example.com
REPO=npm-remote

# 1) AK: request the abbreviated doc, allow compression, dump headers
curl -sS --compressed -H 'Accept: application/vnd.npm.install-v1+json' \
  -o /dev/null -D - -w 'wire_bytes=%{size_download}\n' \
  "$AK/npm/$REPO/vite"
# Observed: content-type: application/json   (NOT application/vnd.npm.install-v1+json)
#           no content-encoding header
#           wire_bytes ~= 38,900,000   (full packument, uncompressed)

# 2) Same request to npmjs for comparison
curl -sS --compressed -H 'Accept: application/vnd.npm.install-v1+json' \
  -o /dev/null -D - -w 'wire_bytes=%{size_download}\n' \
  "https://registry.npmjs.org/vite"
# Observed: content-type: application/vnd.npm.install-v1+json
#           content-encoding: gzip (or br)
#           wire_bytes ~= 400,000    (abbreviated, compressed) -> ~100x smaller for this package
```

End-to-end impact (resolution only, no tarball download):

```sh
# A representative heavy front-end closure (next, react, vite, typescript,
# webpack, eslint, @playwright/test, jest, tailwindcss, @tanstack/react-query, ...)
# in package.json, then with a cold store each time:
pnpm install --lockfile-only --registry "$AK/npm/$REPO/"        # via AK
pnpm install --lockfile-only --registry "https://registry.npmjs.org/"  # direct
```

In my measurements a ~538-package closure pulls ~445 MB of metadata via AK (full, uncompressed) vs roughly an order of magnitude less from npmjs (abbreviated + compressed). From a remote client this made resolution about 7-8x slower; from a client co-located with the AK instance the wall-clock gap shrinks (a fat, low-latency link hides the extra bytes) but the byte volume and per-request parse cost remain.

To isolate which behaviour matters, I built a small localhost harness that re-serves the same packuments under each implementation (full+plain, full+gzip, abbreviated+plain, abbreviated+gzip) and times a cold `pnpm` resolve against each with the network removed:

| Implementation | Resolve time | vs full+plain |
|---|--:|--:|
| full + plain (current AK behaviour) | 4.0s | 1.00x |
| full + gzip | 5.0s | 1.25x (slower) |
| abbreviated + plain | 2.6s | 0.63x |
| abbreviated + gzip | 3.5s | 0.88x |

So **abbreviated metadata is the dominant win** (it cuts both bytes and the JSON the installer must parse, ~37% faster even with the network removed), and compression is an additional large win specifically when the client is far from the server (where wire bytes dominate).

## Hypothesis / likely cause

The npm handler (`backend/src/api/handlers/npm.rs`) appears to return the cached full packument verbatim for the package-document route, without:
- branching on `Accept: application/vnd.npm.install-v1+json` to emit the abbreviated install document (it can be derived from the full packument AK already holds: keep `name`, `dist-tags`, `modified`, and per-version `name`/`version`/`dependencies`/`optionalDependencies`/`peerDependencies`/`peerDependenciesMeta`/`bin`/`directories`/`dist`/`engines`/`deprecated`/`hasInstallScript`/`os`/`cpu`), and
- applying response compression.

TTFB also scales with packument size in my tests, which suggests the full document is read/buffered and served per request; emitting the abbreviated form would reduce that too. (Artifact downloads are already handled well via presigned-redirect; this is specifically the metadata path.)

## Suggested fix

1. Honour `application/vnd.npm.install-v1+json`: when the client requests it, serve the abbreviated install document and set `content-type: application/vnd.npm.install-v1+json`. This is the highest-value change and helps every client.
2. Add response compression for metadata (gzip and/or brotli). `tower-http` (already a dependency) provides a `CompressionLayer`. Caching the compressed bytes avoids per-request compression cost.
3. Optional, smaller: emit `ETag`/`Cache-Control` on metadata responses so clients can do conditional (`304`) revalidation on re-installs.

## Environment

- Artifact Keeper 1.2.x, `STORAGE_BACKEND=s3`, npm remote/virtual repo proxying `registry.npmjs.org`.
- Clients: `pnpm` 11.x and `npm` 11.x (both request abbreviated metadata + advertise gzip/br by default).

Happy to contribute a PR for the abbreviated-metadata path if that would be welcome.

rob-howie-depop:
## Summary
Closes #1931. The npm proxy ignored npm's abbreviated ("corgi") metadata content-negotiation: it returned the full packument regardless of the client's `Accept` header and served metadata uncompressed, so `npm`/`pnpm install` downloaded ~10-100x more metadata bytes per package than from `registry.npmjs.org`.

- Honour `Accept: application/vnd.npm.install-v1+json`: derive and serve the abbreviated ("corgi") install document with the matching `Content-Type`; full-document requests (website, `npm view`) are unchanged. Format reference: https://github.com/npm/registry/blob/main/docs/responses/package-metadata.md
- Compress npm metadata responses (gzip/br) via a `tower-http` `CompressionLayer` scoped to the npm router, excluding already-compressed tarball bodies (`application/gzip`, `application/octet-stream`).

Validated against a live AK 1.2.1 instance: `vite` metadata 38.9 MB → 0.37 MB on the wire; a heavy front-end closure's total metadata 138 MB → 11.6 MB (parity with npmjs's 11.9 MB); full-document requests unchanged; npm publish/install smoke passed.

- [ ] N/A

`test_abbreviated_accept_returns_corgi_document` (handler-level, DB fixture): a request with `Accept: application/vnd.npm.install-v1+json` must return the abbreviated content type, while the default request still returns the full `application/json` packument. It fails on the pre-fix code, where the `Accept` header was ignored and the full packument was always returned.

## Test Checklist
- [x] Unit tests added/updated (abbreviate transform, `Accept` detection, corner cases)
- [x] Integration tests added/updated (handler-level regression test on the DB fixture)
- [x] Manually tested locally (live instance before/after above)

## API Changes
- [x] N/A - no API changes (content negotiation + encoding on the existing package-metadata route; no new endpoints, schema types, or migration)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -71,6 +71,21 @@ dependencies = [
 ]
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "0e76a019e91224d279006ff972f1e984179a6e9feb050adba6ce8274aef23195"
+dependencies = [
+ "alloc-no-stdlib",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -849,6 +864,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "brotli"
+version = "8.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5cc91aac060a7a1e25823bdccbfb6af1875b88f17c6daac97894eed8207166b3"
+dependencies = [
+ "alloc-no-stdlib",
+ "alloc-stdlib",
+ "brotli-decompressor",
+]
+
+[[package]]
+name = "brotli-decompressor"
+version = "5.0.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3a32acac15fe1967bc3986b2a6347dffc965602354ea6f450ad07e8bfd253583"
+dependencies = [
+ "alloc-no-stdlib",
+ "alloc-stdlib",
+]
+
+[[package]]
 name = "bstr"
 version = "1.12.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1214,6 +1250,7 @@ version = "0.4.36"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "00828ba6fd27b45a448e57dbfe84f1029d4c9f26b368157e9a448a5f49a2ec2a"
 dependencies = [
+ "brotli",
  "compression-core",
  "flate2",
  "memchr",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ tokio = { version = "1.50", features = ["full"] }
 # Web framework
 axum = { version = "0.7", features = ["multipart", "tokio"] }
 tower = { version = "0.5", features = ["full"] }
-tower-http = { version = "0.6", features = ["cors", "trace", "compression-gzip"] }
+tower-http = { version = "0.6", features = ["cors", "trace", "compression-gzip", "compression-br"] }
 
 # Database
 sqlx = { version = "0.8", features = ["runtime-tokio", "postgres", "uuid", "chrono", "json", "tls-rustls"] }
--- a/backend/src/api/handlers/npm.rs
+++ b/backend/src/api/handlers/npm.rs
@@ -14,7 +14,7 @@
 
 use axum::body::Body;
 use axum::extract::{Path, State};
-use axum::http::header::{CONTENT_LENGTH, CONTENT_TYPE};
+use axum::http::header::{ACCEPT, CONTENT_LENGTH, CONTENT_TYPE};
 use axum::http::{HeaderMap, StatusCode};
 use axum::response::{IntoResponse, Response};
 use axum::routing::{get, post, put};
@@ -24,6 +24,8 @@ use base64::Engine;
 use bytes::Bytes;
 use sha2::{Digest, Sha256};
 use sqlx::PgPool;
+use tower_http::compression::predicate::{DefaultPredicate, NotForContentType, Predicate};
+use tower_http::compression::CompressionLayer;
 use tracing::{debug, info};
 
 use crate::api::handlers::error_helpers::{map_db_err, map_storage_err};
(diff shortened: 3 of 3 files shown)


rob-howie-depop:
## Install equivalence vs `registry.npmjs.org`

Verified that resolving/installing through this proxy is functionally identical to installing directly from `registry.npmjs.org`. Two checks: a full resolution of the same `package.json`, and a per-package comparison of the abbreviated documents. (Tested against a live AK 1.2.1 instance running this branch.)

### A. Same install (decisive)

`pnpm install --lockfile-only` of the same `package.json` (a ~550-package front-end closure), cold cache, against each registry:

- **555 packages resolved from each, and the `integrity` (content-hash) sets are byte-for-byte IDENTICAL.** Every resolved package is the exact same artifact.
- **The two `pnpm-lock.yaml` files are IDENTICAL** (same versions, same dependency graph, same integrities) after normalising the registry host.

A given install therefore resolves to exactly the same packages and content through the proxy as direct from npmjs.

### B. Abbreviated document, per package (small / large / scoped)

Comparing the `application/vnd.npm.install-v1+json` document for the latest version of `is-number`, `vite`, `eslint`, `webpack`, `@types/node`, `@playwright/test`:

- All install-relevant fields match npmjs (deep, order-insensitive): `dependencies`, `devDependencies`, `optionalDependencies`, `peerDependencies`, `peerDependenciesMeta`, `bin`, `engines`, `funding`, `dist-tags`, and the full version set.
- `dist.integrity` **matches on every package**, so the tarball the client downloads is identical content; only the tarball **URL host** differs (the proxy rewrites it to itself, which is the point of a proxy, and the client verifies the unchanged integrity).

The only residual metadata differences are install-irrelevant: the proxy currently also emits a couple of empty/default-valued fields that npm's corgi prunes (`_hasShrinkwrap: false`, and an empty `peerDependencies: {}` on packages that have none). These are semantically equivalent to absent and do not affect resolution or install; pruning them for byte-exact parity would be a small follow-up.

### Reproduce

```sh
# Resolve the same package.json against the proxy and against npmjs, then diff:
pnpm install --lockfile-only --registry "<ak-instance>/npm/<repo>/"
pnpm install --lockfile-only --registry "https://registry.npmjs.org/"
# the resolved (package@version -> integrity) sets and the lockfiles are identical.

# Per-package abbreviated doc:
curl -sS --compressed -H 'Accept: application/vnd.npm.install-v1+json' "<ak-instance>/npm/<repo>/vite"
curl -sS --compressed -H 'Accept: application/vnd.npm.install-v1+json' "https://registry.npmjs.org/vite"
```

PR #1940, opened 2026-06-21
fix(promotion): enforce promotion_rules on manual single and bulk promote endpoints

brandonrc:
## Summary

Enforce `promotion_rules` on the manual single and bulk promote endpoints.

Previously the rows created via `POST /api/v1/promotion-rules`
(`min_staging_hours`, `require_signature`, `min_health_score`,
`max_cve_severity`, ...) were advisory only: they were consulted by the
`/promotion-rules/{id}/evaluate` dry-run and by a `try_auto_promote` method that
has no callers in the tree. Neither `promote_artifact` nor
`promote_artifacts_bulk` ever evaluated them, so an artifact that fails a
promotion rule was copied straight to the release repository — even without
`skip_policy_check`. The bulk path did no policy evaluation at all per item.

This PR wires the existing rule evaluator into both promote handlers. Before the
storage copy, each handler now evaluates the artifact against every **enabled**
rule for the resolved `(source_repo, target_repo)` pair and, if any rule fails,
returns `200` with `promoted: false` and the violations attached — mirroring the
existing `PromotionPolicyService` block branch. The gate honors the documented
`skip_policy_check` admin override, exactly like the other two promotion gates
(quality gate and `PromotionPolicyService`).

Key design points:
- Enforcement reuses the existing `PromotionRuleService::evaluate_artifact` (the
  same code the dry-run uses) via a new thin wrapper
  `evaluate_for_promotion`, so the live gate and the dry-run can never diverge.
- Rules are matched regardless of `auto_promote` (that flag governs the unused
  background daemon); a governance rule for a pair applies to manual promotion.
- Disabled rules (`is_enabled = false`) and pairs with no rule are unaffected
  (default-allow preserved by the SQL filter returning zero rows).
- Bulk partial success is preserved: a rule-blocked item increments `failed`
  and the batch continues.
- No schema/migration; not retroactive. Adds a runtime `query_as` (not the
  `query!` macro), so no `sqlx prepare` is required.

Added unit tests for the pure `rule_violations_to_policy_violations` mapping
helper (0 / 1 / many failing rules; asserts rule-name + message mapping and the
`high` severity tag). Manually verified end-to-end on an isolated instance: the
strict-rule artifact is now blocked on both single and bulk promote and is not
copied to the release repo, while no-rule pairs, loosened rules, disabled rules,
and `skip_policy_check` break-glass all still promote, and the `/evaluate`
dry-run output is unchanged.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes (response shape unchanged; existing
  `PromotionResponse` / `PolicyViolation` types reused)


brandonrc:
## Blue-team review — NEEDS-WORK (does not fully close the finding)

The fix correctly gates the **single** and **bulk** promote handlers, but review + a live repro found a **third, ungated promote path**:

- `approve_promotion` (`backend/src/api/handlers/approval.rs:474`) copies storage (:563) and inserts the artifact (:576) with **no** `promotion_rules` evaluation — it only consults `PromotionPolicyService`.
- Critically, `promote_artifact` **refuses** to promote from approval-required repos and **forces** the caller into this approval workflow (`promotion.rs:570`). So on exactly the repos most likely to carry governance rules, the new gate is never reached, and an admin can voluntarily route any repo through request→approve to dodge it.
- **Reproduced live:** a `min_staging_hours=720` rule on a fresh artifact was bypassed via request → approve (approval reported `action:"allow"`), and the artifact landed in the release repo.

Also: `max_cve_severity` has a DB default of `'medium'` (`migrations/056_promotion_rules.sql:11`), so a rule created only for `min_staging_hours` **silently also requires a clean scan** and fails closed for unscanned artifacts.

**Rework to merge:** (1) add the rule evaluation to `approve_promotion`; (2) resolve the `max_cve_severity` default-block; (3) add a gate-level regression test (rule-met passes / rule-unmet blocked, incl. the approval path).

PR #1941, opened 2026-06-21
fix(lifecycle): block release-immutability swap via delete + re-upload of different content

brandonrc:
## Summary

Release immutability was enforced in only two places: the live `PUT`-overwrite
check in `artifact_service` (which rejects re-uploading a different version to an
existing, **non-deleted** coordinate) and the `DELETE` gate for
classifier-recognized formats (`#1765`, `#1895`). Neither covers the case where a
released coordinate is first `DELETE`d (soft-deleted to a tombstone) and then a
re-upload of **different** bytes is pushed to the same coordinate.

Two gaps combined to allow this:

- The upload-overwrite check only looks at `is_deleted = false` rows, so after a
  delete it finds nothing and the `ON CONFLICT DO UPDATE` rewrites the coordinate
  with the new bytes.
- Every upload handler first purges the soft-deleted tombstone
  (`cleanup_soft_deleted_artifact`) before its own `INSERT`, destroying the only
  record that the coordinate was previously published — and the per-format upload
  handlers do direct `INSERT`s that never re-check immutability.

This change makes the single pre-insert step every upload shares
immutability-aware. A new `cleanup_soft_deleted_artifact_checked` helper consults
the existing `cache_classifier` (the project's single source of truth for whether a
coordinate is immutable): if the coordinate is structurally immutable and the
incoming bytes differ from the tombstoned bytes, the re-upload is rejected with the
same `409 "Artifact version already exists and is immutable"` the live-overwrite
path returns. Re-uploading the **identical** bytes (idempotent republish / undelete)
and all mutable or unknown-format paths are unchanged — the purge proceeds exactly
as before. A matching backstop is added to `artifact_service::upload_with_sync_options`
for the service-backed upload paths (generic / pypi / debian), which re-queries
including soft-deleted rows before the conflict-update.

The per-format upload handlers (maven, npm, cargo, conan, nuget, composer, conda)
are routed through the checked helper; the pypi service path drops its redundant
pre-cleanup so the service backstop can see the tombstone. Completes the
immutability hardening line begun in `#1765` and `#1895`.

No migration. No change to the classifier or the delete gate.

Added DB-backed tests in `backend/src/api/handlers/mod.rs` covering the helper
directly:
- delete + re-upload of **different** bytes to an immutable coordinate is rejected
  (`409`) and the tombstone is not purged — the bug, now blocked;
- re-upload of **identical** bytes after delete is allowed (idempotent republish);
- a mutable coordinate (`maven-metadata.xml`) re-upload of different bytes proceeds;
- first upload with no tombstone proceeds.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/.sqlx/query-78c759c40a5c20d9f330cc953fcacd8d6e0e8df8d119a0f4c748adfde5f378e7.json
+++ b/.sqlx/query-78c759c40a5c20d9f330cc953fcacd8d6e0e8df8d119a0f4c748adfde5f378e7.json
@@ -1,0 +1,29 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT checksum_sha256, version FROM artifacts WHERE repository_id = $1 AND path = $2 AND is_deleted = true",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "checksum_sha256",
+        "type_info": "Bpchar"
+      },
+      {
+        "ordinal": 1,
+        "name": "version",
+        "type_info": "Varchar"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid",
+        "Text"
+      ]
+    },
+    "nullable": [
+      false,
+      true
+    ]
+  },
+  "hash": "78c759c40a5c20d9f330cc953fcacd8d6e0e8df8d119a0f4c748adfde5f378e7"
+}
--- a/.sqlx/query-ab4717210b7d122dcc27b5b9ee2b81e4e672426ef3c4c463c78991f1f6a65393.json
+++ b/.sqlx/query-ab4717210b7d122dcc27b5b9ee2b81e4e672426ef3c4c463c78991f1f6a65393.json
@@ -1,0 +1,29 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT checksum_sha256, version FROM artifacts WHERE repository_id = $1 AND path = $2",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "checksum_sha256",
+        "type_info": "Bpchar"
+      },
+      {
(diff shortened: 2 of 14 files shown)


brandonrc:
## Blue-team review — NEEDS-WORK (bypass still fully open; that's why re-attack couldn't prove closure)

The exploit is `DELETE` then `PUT /api/v1/repositories/{repo}/artifacts/{path}`, which maps to `upload_artifact`/`delete_artifact` in `backend/src/api/handlers/repositories.rs` — **this PR does not touch that file**; it patches the format handlers + `artifact_service` instead.

- The PR's backstop (`artifact_service.rs:349-376`) re-queries the row incl. soft-deleted tombstones and rejects an immutable-coordinate swap — correct in isolation, but `upload_artifact` first purges the tombstone via the **unchecked** `cleanup_soft_deleted_artifact` (`repositories.rs:3521`), so by the time the backstop's SELECT runs `prior` is `None` and the guard short-circuits. (The author removed this exact pre-purge from `pypi.rs:1853` but missed the identical call on the exploited generic endpoint.) **Live-confirmed the swap still lands** (DELETE 200 → re-upload different bytes 201 → GET returns swapped content).
- Secondary: the immutability oracle uses `cache_classifier::classify` (a proxy-cache-TTL classifier whose default returns **mutable** for Generic/Nuget/Conan/Composer/Go/Rpm/Debian/Helm), not the artifact's real released/lifecycle state.

**Rework to close:** (1) use `cleanup_soft_deleted_artifact_checked` at `repositories.rs:3521` (or drop the pre-purge and let the surviving tombstone trigger the backstop via `ON CONFLICT DO UPDATE`); (2) gate on real released/lifecycle state, not the path classifier; (3) add an e2e `DELETE`→`PUT` integration test on `/repositories/{key}/artifacts/*path`.

PR #1944, opened 2026-06-22
fix(pypi): serve PEP 691 JSON with upload-time for proxied simple indexes

### LINKED ISSUE #1545: fix(pypi): emit upload-time metadata in simple index responses (PEP 691 JSON + PEP 714 HTML)
## Problem Statement

`GET /pypi/{repo_key}/simple/{project}/` does not include `upload-time` metadata in either its HTML or JSON (PEP 691) responses. This breaks client-side **age-gating** features in `uv`, `pip`, and `poetry` that depend on `upload-time` to filter packages uploaded before or after a given date.

### Observed Error

```
ERROR: Index https://<your-artifact-keeper-host>/pypi/<repo>/simple/redis/
       does not provide upload-time metadata
```

Triggered by `uv pip install` / `uv lock` when `--exclude-newer` (or an equivalent `pyproject.toml` constraint) is used.

### Affected Clients

| Client | Option | Requires |
|--------|--------|----------|
| `uv` | `--exclude-newer <date>` | `upload-time` in PEP 691 JSON **or** HTML |
| `pip` | `--uploaded-prior-to` + date-range tooling | `upload-time` |
| `poetry` | `source: url:` + version constraints | `upload-time` |

## Root Cause

The simple-index builder in `backend/src/api/handlers/pypi.rs` constructs per-file records (both HTML `<a>` tags and PEP 691 JSON file objects) without ever including an `upload-time` field:

**JSON branch** (`build_simple_project_response`, ~L519–527):
```rust
let mut file = serde_json::json!({
    "filename": filename,
    "url": format!("/pypi/{}/simple/{}/{}", repo_key, normalized, filename),
    "hashes": { "sha256": &a.checksum_sha256 },
    "size": a.size_bytes,
});
if let Some(rp) = requires_python {
    file["requires-python"] = serde_json::Value::String(rp);
}
// ← no "upload-time" here
```

**HTML branch** (~L560–581):
```rust
html.push_str(&format!(
    "<a href=\"{}\"{}>{}</a><br/>\n",
    url, rp_attr, filename
));
// ← no data-upload-time attr
```

The SQL query that populates `SimpleProjectArtifact` also omits `created_at`:
```sql
SELECT a.id, a.path, a.name, a.version, a.size_bytes, a.checksum_sha256,
       am.metadata as "metadata?"
FROM artifacts a
LEFT JOIN artifact_metadata am ON am.artifact_id = a.id
WHERE ...
```

`artifacts.created_at` already exists in the schema and is a valid proxy for upload-time without any migration.

Additionally, for **Remote (proxy) repositories**, the regex that strips PEP 658 attrs during URL rewriting (`METADATA_ATTR_RE`, ~L1528–1566) does not handle `data-upload-time`, so there is no consistent policy for preserving upstream `upload-time` on proxied HTML responses either.

### 1. Include `created_at` in the simple-index SQL queries

Add `a.created_at` to both SQL queries in `simple_project` (~L283–296 and ~L367–376):

```sql
SELECT a.id, a.path, a.name, a.version, a.size_bytes, a.checksum_sha256,
       a.created_at,
       am.metadata as "metadata?"
FROM artifacts a
LEFT JOIN artifact_metadata am ON am.artifact_id = a.id
WHERE ...
```

### 2. Extend `SimpleProjectArtifact`

```rust
struct SimpleProjectArtifact {
    path: String,
    version: Option<String>,
    size_bytes: i64,
    checksum_sha256: String,
    metadata: Option<serde_json::Value>,
    created_at: DateTime<Utc>,   // ← add
}
```

### 3. Emit `upload-time` in the JSON response (PEP 691)

PEP 691 specifies the field as an RFC 3339 string:

```rust
let mut file = serde_json::json!({
    "filename": filename,
    "url": ...,
    "hashes": { "sha256": &a.checksum_sha256 },
    "size": a.size_bytes,
    "upload-time": a.created_at.to_rfc3339(),  // ← add
});
```

### 4. Emit `data-upload-time` in the HTML response (PEP 714)

PEP 714 defines a `data-upload-time` attribute on `<a>` tags using ISO 8601 / RFC 3339:

```rust
html.push_str(&format!(
    "<a href=\"{}\"{}  data-upload-time=\"{}\">{}</a><br/>\n",
    url, rp_attr,
    a.created_at.to_rfc3339(),
    filename
));
```

### 5. Proxy: preserve `data-upload-time` from upstream

For **Remote** repos the upstream PyPI HTML already contains `data-upload-time` on each `<a>` link. The current rewrite loop (`rewrite_upstream_urls`) does not strip it, so it should pass through correctly once URL rewriting runs — but this should be explicitly verified and covered by a test. No regex change needed for HTML.

For the **JSON proxy path**: upstream PEP 691 responses for remote repos are passed through without transformation today (only HTML is rewritten), so JSON `upload-time` from upstream should already be preserved. This also needs a test to confirm.

### 6. No schema migration required

`artifacts.created_at` is set at insert time by the DB default and is already present. No migration is needed.

## Cost and Compatibility

- **No breaking changes.** `upload-time` is an additive field; clients that do not use it ignore it.
- **Local / Virtual repos**: one extra column in the existing `SELECT`, zero extra storage or network calls.
- **Remote repos**: HTML rewrite already iterates each `<a>` tag; adding attribute preservation is O(1) per tag.
- The **listing endpoint** (`GET /:key/artifacts`) is unaffected.

## Alternatives Considered

- **Store `upload-time` explicitly in `artifact_metadata`** at upload time. Adds redundancy — `artifacts.created_at` is the canonical upload timestamp and requires no migration. Could be revisited if an artifact ever needs a decoupled "published at" vs "stored at" distinction.
- **New dedicated endpoint** (`GET /:key/simple/:project/upload-times`). Forces clients into a second round-trip; standard clients (`pip`, `uv`, `poetry`) expect the field inline per PEP 691 / PEP 714.

## Use Case

As a developer or CI pipeline using `uv pip install --exclude-newer 2025-01-01` against an artifact-keeper PyPI proxy or local repo, I want the simple index to include `upload-time` so that version age-gating works without falling back to PyPI.org or disabling the feature entirely.

## Relevant Specifications

- [PEP 691](https://peps.python.org/pep-0691/) — JSON-based Simple API (`"upload-time"` field)
- [PEP 714](https://peps.python.org/pep-0714/) — `data-upload-time` attribute on HTML simple index links
- [PEP 503](https://peps.python.org/pep-0503/) — Simple Repository API (HTML baseline)

## Component

Backend / API — `backend/src/api/handlers/pypi.rs`

## Pre-submission Checklist

- [x] Reviewed the OpenAPI spec and source to confirm `upload-time` is not currently emitted.
- [x] Verified `artifacts.created_at` is present in the schema — no migration required.

rob-howie-depop:
## Summary

`GET /pypi/{repo}/simple/{project}/` ignores the client's `Accept` header on the **proxy path**. For Remote (and Virtual-with-remote) repositories the handler always fetches the upstream simple index with no `Accept` header, so PyPI returns its **HTML** index, which is relayed verbatim — even when the client asked for `application/vnd.pypi.simple.v1+json`.

PEP 700 `upload-time` exists only in the JSON representation (the HTML index never carries it), so proxied packages never expose `upload-time`. Clients that depend on it break — e.g. `uv --exclude-newer` / `uv lock` silently lose the cutoff:

```
$ echo requests | uv pip compile - \
    --index-url https://<host>/pypi/<proxy-repo>/simple/ \
    --exclude-newer 2024-06-01
warning: requests-2.34.2-... is missing an upload date, but user provided: 2024-06-01T...
# the cutoff is ignored; a version newer than the bound is resolved
```

The hosted/local-package case was fixed in #1794; this extends the same guarantee to the proxy path (the remaining part of #1545).

Closes #1545

## What changed

- `simple_project` now honours `Accept: application/vnd.pypi.simple.v1+json` on the proxy path: it requests the PEP 691 JSON form from upstream and serves JSON, preserving `upload-time` (plus `size`, `hashes`, `requires-python`).
- The JSON form is cached under a **format-qualified key** (`.../simple/<project>/index.v1+json`) so it never collides with the cached HTML index.
- New `rewrite_upstream_simple_json` rewrites `files[].url` to route downloads through the proxy and strips the PEP 658/714 metadata signals the proxy can't serve (`core-metadata`, `data-dist-info-metadata`) — the JSON analogue of `rewrite_upstream_urls`.
- New `merge_local_into_remote_simple_json` unions a virtual repo's local distributions into the proxied JSON (dedup by filename, union `versions`, surface `meta.tracks`).
- Upstreams that ignore the JSON `Accept` and return HTML fall back to the existing HTML rewrite, so their behaviour is unchanged.

No schema or migration changes (`upload-time` is sourced from `artifacts.created_at`, already present).

Unit tests for `rewrite_upstream_simple_json` and `merge_local_into_remote_simple_json` cover URL rewriting, `upload-time` preservation, PEP 658/714 metadata stripping, the local↔remote JSON merge, version union, and `requires-python`/`tracks`.

## Test Checklist

- [x] Integration tests added/updated — `backend/tests/pypi_proxy_upload_time_tests.rs` (wiremock-backed; verified to fail on `main` and pass with the fix)
- [ ] E2E tests added/updated (covered by manual end-to-end verification below)
- [x] Manually verified end-to-end against a live deployment proxying pypi.org (see comment)
- [x] No regressions in existing tests (`cargo test --lib`)

## API Changes

- [x] N/A — additive content negotiation on an existing endpoint; no new routes or types

### DIFF
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -854,6 +854,34 @@ pub async fn proxy_fetch_with_cache_key(
     .await
 }
 
+/// Variant of [`proxy_fetch_with_cache_key`] that also forwards an `Accept`
+/// header to the upstream. The PyPI simple-index proxy uses this to request
+/// the PEP 691 JSON representation while keying the cache on a format-qualified
+/// `cache_path`, so the JSON and HTML forms of the same index never collide in
+/// the proxy cache.
+pub async fn proxy_fetch_with_cache_key_and_accept(
+    proxy_service: &ProxyService,
+    repo_id: Uuid,
+    repo_key: &str,
+    upstream_url: &str,
+    fetch_path: &str,
+    cache_path: &str,
+    accept: Option<&str>,
+) -> Result<(Bytes, Option<String>), Response> {
+    with_proxy_repo(
+        repo_id,
+        repo_key,
+        upstream_url,
+        fetch_path,
+        |repo| async move {
+            proxy_service
+                .fetch_artifact_with_cache_path_and_accept(&repo, fetch_path, cache_path, accept)
+                .await
+        },
+    )
+    .await
+}
+
 /// Streaming sibling of [`proxy_fetch_with_cache_key`] (#895 OOM relief for
 /// format handlers whose upstream download URL differs from the canonical
 /// artifact path). Fetches `fetch_path` from the upstream but keys the proxy
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -554,6 +554,15 @@ async fn simple_project(
     let repo = resolve_pypi_repo(&state.db, &repo_key).await?;
     let normalized = normalize_pep503(&project);
 
+    // PEP 691 content negotiation also governs the proxy path: a JSON client
+    // must get the upstream's JSON representation (which carries PEP 700
+    // `upload-time`), not its HTML index (which never does).
+    let wants_json = headers
+        .get("accept")
+        .and_then(|v| v.to_str().ok())
+        .unwrap_or("")
+        .contains(PEP691_JSON_CONTENT_TYPE);
+
     // Find all artifacts that belong to this package.
     // We normalize the name for matching: replace [_.-]+ with - then lowercase.
     let artifacts = sqlx::query!(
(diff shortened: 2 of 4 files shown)


rob-howie-depop:
Verified end-to-end against a live deployment proxying `pypi.org` through a virtual repo.

**Before** (current behaviour): a JSON request (`Accept: application/vnd.pypi.simple.v1+json`) to a proxied package returns `Content-Type: text/html` with zero `upload-time` fields, and `uv pip compile - --exclude-newer 2024-06-01` warns "missing an upload date" for every file and resolves `requests==2.34.2` (newer than the cutoff — the bound is silently ignored).

**After** (this PR): the same JSON request returns `application/vnd.pypi.simple.v1+json` with `upload-time` on every file and `files[].url` rewritten through the proxy; `uv ... --exclude-newer 2024-06-01` resolves `requests==2.32.3` with no warnings — identical to resolving against `pypi.org` directly.

rob-howie-depop:
@mergifyio rebase

PR #1947, opened 2026-06-22
feat(storage): add tracing spans around storage backend operations

### LINKED ISSUE #1946: Add OpenTelemetry child spans around the S3 storage operations (get/put/head/stream/presign)
## Summary

The storage backend operations are not instrumented with tracing spans. The only request-scoped spans are at the HTTP middleware layer (`backend/src/api/middleware/tracing.rs` and the `TraceLayer` in `backend/src/main.rs`), so an HTTP request span shows the full duration as time spent awaiting I/O with **no child span** to attribute the wait to a specific storage operation. When a request is slow, a trace cannot tell you whether the time went to a storage `GET`, a `HEAD` existence probe, a presign call, or something else.

This is an observability-only gap: adding spans around the storage client methods makes per-operation latency visible in APM with no behavior change.

## Where it is in the code (current `main`)

- The `S3Backend` `StorageBackend` impl (`backend/src/storage/s3.rs:1048`) has no `#[instrument]` on any of its methods; they emit only point-in-time `tracing::debug!` events, which cannot carry a duration or appear as child spans.
- A ripgrep for `instrument|info_span|debug_span|tracing::span` across `backend/src/storage/` returns nothing.
- OTLP export is already wired: `tracing_opentelemetry::layer()` is installed on the subscriber (`backend/src/telemetry.rs:141`), so any span created via the `tracing` crate is exported automatically when `OTEL_EXPORTER_OTLP_ENDPOINT` is set. No new dependencies are required.

## Proposed change

Add `#[tracing::instrument(...)]` to the object-operation methods of the `S3Backend` impl (`put`, `get`, `exists`, `delete`, `copy`, `head_etag`, `get_presigned_url`, `get_stream`, `get_range`, `put_stream`, `health_check`), matching the existing house style (e.g. `services/email_dispatcher.rs` and `services/migration_service.rs`): `skip(self, <byte args>)`, generic semantic-convention fields, low (`debug`) level so a high-frequency storage path does not flood `info` in production while still being exported when enabled.

Fields kept generic and deployment-agnostic, for example:

```
otel.kind = "client", storage.system = "s3", storage.operation = "get", key = %key
```

Deliberately **not** recording bucket name, region, endpoint or credentials (those are deployment-specific). One caveat to document with a short comment: `get_stream` returns a stream, so its span measures `GET` initiation (time-to-first-byte), not the full body transfer, which happens as the caller polls the stream.

This is a small, attribute-only diff: no signature changes, no behavior changes, runtime-gated entirely by the existing `OTEL_*` configuration. I will open a PR.

Note: the same pattern could later be applied to the other storage backends (GCS, Azure, filesystem); this issue/PR scopes the S3 backend first as the most common production target.

rob-howie-depop:
## Summary

The storage backend operations are not instrumented with tracing spans. The only request-scoped spans are at the HTTP middleware layer, so an HTTP request span shows its whole duration as time spent awaiting I/O with no child span to attribute the wait to a specific storage operation. When a request is slow, a trace cannot tell you whether the time went to a storage `GET`, a `HEAD` existence probe, a presign call, or something else.

This adds `#[tracing::instrument]` to the object operations of every `StorageBackend` impl, so each storage call appears as a child span under the request span:

- **S3** (`backend/src/storage/s3.rs`) and **GCS** (`gcs.rs`) and **Azure** (`azure.rs`): `put`, `get`, `exists`, `delete`, `copy`, `head_etag`, `get_presigned_url`, `get_stream`, `get_range`, `put_stream` (and `put_file` where overridden), `health_check`.
- **Filesystem** (`filesystem.rs`): `put`, `get`, `exists`, `delete`, `copy`, `put_file`, `get_stream`, `get_range`, `put_stream`.

Fields are generic OpenTelemetry semantic-convention attributes (`otel.kind`, `storage.system`, `storage.operation`) plus the auto-recorded `key`. `otel.kind` is `"client"` for the object stores and `"internal"` for the local filesystem. Bucket/container name, region, endpoint and credentials are deliberately not recorded.

OTLP export is already wired (`telemetry.rs` installs the `tracing_opentelemetry` layer), so these spans flow to APM automatically when `OTEL_EXPORTER_OTLP_ENDPOINT` is set; with no exporter configured they are near-zero cost. No behaviour change, no new dependencies, no signature changes.

`get_stream`'s span covers `GET` initiation (time-to-first-byte); the body transfers as the caller polls the returned stream, noted in a code comment.

Closes #1946

## Test Checklist
- [x] Manually tested locally

Verified locally: `cargo fmt --check`, `cargo clippy -p artifact-keeper-backend --lib -- -D warnings` (clean), and `cargo test -p artifact-keeper-backend --lib storage::` (369 passed, 0 failed across all four backends). This is an attribute-only, behaviour-neutral change, so no new unit tests are added.

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/storage/azure.rs
+++ b/backend/src/storage/azure.rs
@@ -1118,6 +1118,7 @@ impl AzureBackend {
 
 #[async_trait]
 impl StorageBackend for AzureBackend {
+    #[tracing::instrument(skip(self, content), fields(otel.kind = "client", storage.system = "azure", storage.operation = "put"))]
     async fn put(&self, key: &str, content: Bytes) -> Result<()> {
         let url = self.blob_url(key);
         let response = self.authorized_put(&url, key, &content).await?;
@@ -1134,6 +1135,7 @@ impl StorageBackend for AzureBackend {
         Ok(())
     }
 
+    #[tracing::instrument(skip(self), fields(otel.kind = "client", storage.system = "azure", storage.operation = "get"))]
     async fn get(&self, key: &str) -> Result<Bytes> {
         let url = self.read_url(key, Duration::from_secs(300))?;
         let response = self.authorized_get(&url).await?;
@@ -1187,6 +1189,7 @@ impl StorageBackend for AzureBackend {
         Ok(bytes)
     }
 
+    #[tracing::instrument(skip(self), fields(otel.kind = "client", storage.system = "azure", storage.operation = "get_range"))]
     async fn get_range(&self, key: &str, offset: u64, length: usize) -> Result<Bytes> {
         if length == 0 {
             return Ok(Bytes::new());
@@ -1220,6 +1223,9 @@ impl StorageBackend for AzureBackend {
         )))
     }
 
+    // The span covers GET initiation (time-to-first-byte); the body transfer
+    // happens later as the caller polls the returned stream.
+    #[tracing::instrument(skip(self), fields(otel.kind = "client", storage.system = "azure", storage.operation = "get_stream"))]
     async fn get_stream(&self, key: &str) -> Result<BoxStream<'static, Result<Bytes>>> {
         let url = self.read_url(key, Duration::from_secs(300))?;
         let mut response = self.authorized_get(&url).await?;
@@ -1263,6 +1269,7 @@ impl StorageBackend for AzureBackend {
         Ok(Box::pin(stream))
     }
 
+    #[tracing::instrument(skip(self, stream), fields(otel.kind = "client", storage.system = "azure", storage.operation = "put_stream"))]
     async fn put_stream(
         &self,
         key: &str,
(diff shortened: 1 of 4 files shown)


brandonrc:
This has drifted out of sync with `main` (CONFLICTING) after the v1.3.0 merge window. Could you rebase onto current `main` and resolve conflicts? Once it's green again we'll pull it into the v1.4.0 milestone. Thanks for the contribution — flagging so it doesn't go stale.

rob-howie-depop:
Done — rebased onto current `main` (twice, in fact: once resolving the `put_stream` doc-comment conflict from the MultipartAbortGuard change, and again after today's AccessScope/health-gating commits landed). CI is fully green on the rebased head and the PR shows MERGEABLE/CLEAN. Verified locally with `cargo fmt --check`, `clippy -D warnings`, and the 388 storage:: unit tests. Happy for it to go into v1.4.0 — thanks for flagging.

Issue #1971, opened 2026-06-22
OCI image index/tag scan produces empty package inventory and bogus SBOM components


arthur-sommer-etc:
### Description

When scanning a Docker/OCI image tag that is stored as an OCI image index, Artifact Keeper does not produce an accurate package inventory or SBOM. The image scan completes with zero packages, and SBOM generation falls back to unrelated OpenSCAP/compliance findings, resulting in a CycloneDX SBOM with bogus components instead of the container’s actual packages.

This appears different from the already-fixed digest-reference separator issue (#1483/#1496). That fix handles digest-pinned manifest references; this issue is about scanning a tag/index artifact and resolving it to the correct platform manifest/layers for inventory.

### Environment Information

 - Deployment type: docker-compose
 - Operating System: Linux
 - Architecture: amd64
 - Database: PostgreSQL
 - Reverse Proxy: Caddy / HTTPS
 - Network setup: Docker Compose default network

### Backend Version

1.2.1

### Web UI Version

1.2.0

### Package Format

Docker / OCI

### Expected Behavior

When scanning an OCI image tag/index, Artifact Keeper should resolve the image index to a concrete platform manifest/layers, run Trivy/Grype against that image, persist discovered packages into scan_packages, and generate an SBOM from that real package inventory.

The generated SBOM should contain the container’s OS/application packages, not unrelated scanner/compliance findings.

### Observed Behavior

This also reproduces with a single-platform Docker manifest, not only an OCI index.

Example artifact:

 v2/gitlab-codex-review/manifests/latest

Content type:

 application/vnd.docker.distribution.manifest.v2+json

Observed scan rows:

 image      completed       0 findings, 0 scan_packages
 grype      failed          GrypeLicense bare-string license parse error
 openscap   completed       3 findings
 filesystem not_applicable
 dependency completed       0 findings

Generated SBOM:

 CycloneDX component_count: 3
 dependency_count: 3

Those 3 SBOM components appear to come from OpenSCAP/compliance findings, not from the container package inventory.

This suggests the problem is broader than OCI image index resolution. Even for a concrete Docker manifest, the Trivy image scanner completes with zero package inventory. The likely cause is that the image scanner constructs an unqualified image target such as:

 gitlab-codex-review:latest

instead of a registry-qualified Artifact Keeper reference such as:

 artifact-keeper.etcconnect.com/gitlab-codex-review:latest

[...]

PR #1988, opened 2026-06-22
fix(scanner): qualify Trivy image scan target with owning repository key

### LINKED ISSUE #1964: Trivy image scans omit the owning repository key and resolve against Docker Hub
## Summary

The container-image scanner (`ImageScanner`) inherits the trait-default `scan_target`, so it builds a bare `<name>:<tag>` reference from the stored OCI path (`v2/<image>/manifests/<ref>`), which omits the owning repository key. The Trivy client then resolves that target against **Docker Hub**, not the Artifact Keeper registry.

## Impact

- Internal-only images (both OCI and Docker v2 Schema 2 manifests): the scan **completes with 0 findings and no error** — a silent false-clean.
- Images that share a public name: Trivy scans the **wrong** (public) image.

## Context

Grype's OCI path was made repository-scoped in #1664 and the production dispatch hardened in #1952. The Trivy image path has no equivalent and still uses the bare ref.

## Reproduction

1. Push an internal-only image to a docker repository.
2. Run a Trivy image scan on it.

The scan completes with **0 findings** even though the image carries CVEs (e.g. a Grype scan of the same artifact reports them).

## Proposed fix

Override `ImageScanner::scan_target` to prepend the owning repository key (`<host>/<repo_key>/<name>:<ref>`), mirroring `GrypeScanner`, so the Trivy client pulls the stored artifact from Artifact Keeper's own registry.

brandonrc:
## Summary

The container-image scanner (`ImageScanner`) inherited the trait-default `scan_target`, which delegates to `scan()` and builds a bare `<name>:<tag>` reference from the stored OCI path (`v2/<image>/manifests/<ref>`). That ref omits both the registry host and the owning repository key, so the Trivy CLI resolves it against **Docker Hub** instead of Artifact Keeper's own registry:

- Internal-only images complete with **0 findings and no error** — a silent false-clean.
- Images that share a public name get the **wrong** (public) image scanned.

This mirrors the Grype OCI hardening done in #1664/#1952; the Trivy image path had no equivalent.

### Change

- Override `ImageScanner::scan_target` to build a registry-qualified ref `<host>/<repo_key>/<name>:<ref>` using the owning repository key supplied by scanner orchestration, mirroring `GrypeScanner::scan_target` / `build_registry_image_ref_for_repo`.
- Share Grype's `resolve_registry_host` (now `pub(crate)`) so both scanners resolve the same registry endpoint and cannot diverge.
- Factor the health-check + invoke + report logic into `scan_image_ref`, used by both the legacy `scan()` (bare ref, no repo context) and the new `scan_target` (qualified ref) paths.
- Digest references keep the `@sha256:...` separator via the shared `join_oci_image_ref` (#1483).

Stays narrow per the issue: this only threads the repository key into the Trivy scan ref. OCI image-index → platform-manifest resolution (#1971) is intentionally out of scope.

## Test Checklist
- [x] Manually tested locally

Added `test_extract_image_ref_for_repo_prepends_host_and_repo_key`, `..._digest_uses_at_separator`, and `..._rejects_malformed_path`. The first asserts the qualified ref differs from the Docker Hub-resolving bare ref and contains `<repo_key>/<name>:<tag>` behind a host. `cargo check`, `cargo clippy --lib`, `cargo fmt --check`, and the image_scanner test suite all pass locally.

## API Changes
- [x] N/A - no API changes

Fixes #1964

### DIFF
--- a/backend/src/services/grype_scanner.rs
+++ b/backend/src/services/grype_scanner.rs
@@ -241,7 +241,12 @@ fn classify_grype_spawn_error(err: &std::io::Error) -> AppError {
 ///
 /// The returned value has any scheme (`https://`, `http://`) stripped and
 /// trailing `/` trimmed, because Grype expects `host[:port]`, not a URL.
-fn resolve_registry_host() -> String {
+///
+/// Exposed `pub(crate)` so the Trivy `ImageScanner` builds its registry image
+/// ref against the exact same host as Grype (#1964); a divergent resolver
+/// would let the two scanners target different registries for the same
+/// artifact.
+pub(crate) fn resolve_registry_host() -> String {
     let raw = std::env::var("AK_GRYPE_REGISTRY_HOST")
         .ok()
         .filter(|s| !s.is_empty())
--- a/backend/src/services/image_scanner.rs
+++ b/backend/src/services/image_scanner.rs
@@ -6,7 +6,7 @@ use tracing::{info, warn};
 use crate::error::{AppError, Result};
 use crate::models::artifact::{Artifact, ArtifactMetadata};
 use crate::services::scanner_service::{
-    cached_trivy_cli_version, ScanOutput, Scanner, VersionCache,
+    cached_trivy_cli_version, ScanOutput, ScanTarget, Scanner, VersionCache,
 };
 
 #[cfg(test)]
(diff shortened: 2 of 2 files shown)


brandonrc:
Closing as a duplicate of #1965, which fixes the same issue (#1964) with an equivalent implementation and was opened first by an external contributor. Consolidating on #1965.

PR #2040, opened 2026-06-26
fix(sso): add per-provider use_absolute_acs_url opt-in for stricter SAML IdPs

### LINKED ISSUE #2038: SAML AuthnRequest emits a relative AssertionConsumerServiceURL, rejected by stricter IdPs
### Description

`saml_login` (`backend/src/api/handlers/sso.rs:638`) and `saml_acs` (`sso.rs:694`) embed `/api/v1/auth/sso/saml/<id>/acs` — a path-relative URL — into the SAML `AssertionConsumerServiceURL` attribute of every outbound AuthnRequest, and use the same string when validating the IdP-asserted `Destination` / `Recipient` on the callback. Permissive IdPs resolve the relative path against the SP host they receive the AuthnRequest from and the flow round-trips fine. Stricter SAML 2.0 IdPs — notably Lark AnyCross (Feishu's enterprise SSO) and several common enterprise deployments — reject a relative `AssertionConsumerServiceURL` outright, so for those IdPs the SP is unreachable: the AuthnRequest does not even leave the IdP-side processing pipeline. There is no operator workaround short of editing the source.

### Environment Information

- Deployment type: Docker (docker compose), Kubernetes
- Operating System: Linux
- Architecture: amd64, arm64
- Database: PostgreSQL 16
- Reverse Proxy: Caddy (with `AK_EXTERNAL_URL` configured) — also reproducible under nginx with `X-Forwarded-Proto` / `X-Forwarded-Host`
- Network setup: Public IP behind reverse proxy, IdP reachable over the public internet

### Backend Version

1.2.2 (main also affected)

### Web UI Version

1.2.2 (web UI not involved in this code path)

### Package Format

N/A / General

### Expected Behavior

A configurable absolute ACS URL that the SP emits when the IdP requires it, derived from the same base-URL resolver (`AK_EXTERNAL_URL` env, then `X-Forwarded-Proto` / `Host` headers, then URI authority, then `Host` header) the rest of the SSO stack already uses (`backend/src/api/extractors.rs::RequestBaseUrl`). Per-provider, defaulted to `false` so existing deployments do not change their AuthnRequest wire format on upgrade.

### Observed Behavior

The AuthnRequest carries `AssertionConsumerServiceURL="/api/v1/auth/sso/saml/<id>/acs"` and the strict IdP rejects it before issuing an assertion. End-user sees a generic IdP error page; backend never receives an ACS callback.

### Steps to Reproduce

1. Configure a SAML provider whose IdP enforces strict ACS URL parsing (Lark AnyCross will do, as will a sufficiently locked-down Okta / Azure AD / Keycloak instance — check the IdP's AuthnRequest validator settings).
2. Trigger a login via `/api/v1/auth/sso/saml/<id>/login`.
3. The IdP rejects the AuthnRequest because `AssertionConsumerServiceURL="/api/v1/auth/sso/saml/<id>/acs"` is not a syntactically valid absolute URL.

### Logs

Symptom is on the IdP side. Backend emits a normal redirect for `saml_login` and never sees the ACS callback because the IdP refuses to issue one.

### Additional Context

Proposed fix: add `saml_configs.use_absolute_acs_url BOOLEAN NOT NULL DEFAULT false` (per-provider opt-in) and have the two ACS handlers consult it through a shared pure helper. PR to follow this issue.

### Pre-submission Checklist

### LINKED ISSUE #2096: SAML: validate_response does not check Recipient/Destination/InResponseTo
## Summary
`saml_service.rs::validate_response` performs **no Recipient/Destination (and no InResponseTo) validation** on the incoming SAML Response at the ACS callback. Found during red-team verification of #2040 (pre-existing; not introduced by that PR).

## Impact
The SAML spec expects the SP to verify that the assertion's `Recipient`/`Destination` matches the SP's own ACS URL and that `InResponseTo` matches an outstanding AuthnRequest. Without these, the anti-replay / audience-delivery binding is weaker (an assertion minted for/delivered elsewhere is not rejected on these grounds). Signature + `Audience` checks still apply, so this is defense-in-depth, not an open bypass on its own — but it's the binding that would contain issues like #2040.

## Fix
In `validate_response`, verify `Recipient`/`Destination` == the SP ACS URL (from trusted config, per #2040's fix) and enforce `InResponseTo` against a stored request id / relay-state.

## Related
Surfaced verifying #2040; pairs with that PR's trusted-ACS-URL fix.

StringKe:
## Summary

Hardens the SAML SP (Service Provider) side of the SSO stack. Two related
changes, both scoped to the SAML login/ACS flow:

**1. Per-provider absolute `AssertionConsumerServiceURL` opt-in (#2038)**

`saml_login` / `saml_acs` (`backend/src/api/handlers/sso.rs`) historically
emitted the relative path `/api/v1/auth/sso/saml/<id>/acs` in the outbound
AuthnRequest's `AssertionConsumerServiceURL`, and used the same string when
validating the IdP-asserted delivery target on the callback. Permissive IdPs
resolve the relative path against the SP host and round-trip fine; stricter
SAML 2.0 IdPs (e.g. Lark AnyCross and some enterprise deployments) reject a
relative ACS URL outright.

A per-provider opt-in `saml_configs.use_absolute_acs_url` (**migration 139**,
`DEFAULT false`) governs the wire format: only when an operator enables it does
the AuthnRequest carry `{base}/api/v1/auth/sso/saml/<id>/acs`. Critically, the
absolute base is sourced **only** from the operator-configured `AK_EXTERNAL_URL`
(exposed as `api::extractors::trusted_external_url()`), never from
request-derived headers (`Host`, `X-Forwarded-Host`). Both call sites read the
flag through one pure helper `build_saml_acs_url(use_absolute, trusted_base, id)`
so they cannot drift; trailing slashes and reverse-proxy subpaths normalise
correctly. If the flag is on but `AK_EXTERNAL_URL` is unset, the helper **fails
closed** to the relative form and logs a warning. Existing providers keep their
exact pre-139 wire format because the flag defaults to `false`.

`SamlConfigRow` / `SamlConfigResponse` / `Create*` / `Update*` and the matching
SQL (list / get / get_decrypted / create / update / toggle) carry the new
column together; the `redacted_debug` whitelist gains the non-sensitive bool.

**2. Response-binding + InResponseTo enforcement (#2096)**

The ACS callback previously validated status, issuer, audience, time window and
XML signature, but ignored *where* the IdP said it delivered the response and
*whether it corresponds to a request this SP actually made*:

- The parser (`saml_service.rs`) now extracts the `<Response>` `Destination`
  and the assertion `<SubjectConfirmationData>` `Recipient`.
- `validate_response` rejects a response whose `Destination` or assertion
  `Recipient` does not match this SP's own ACS URL. This is **conditional
  defense-in-depth**: it fires only when a trusted ACS URL is available
  (`AK_EXTERNAL_URL` set) **and** the IdP actually asserted the attribute, so
  permissive IdPs that omit them and deployments without a trusted external URL
  are unaffected. The expected ACS is derived from the same trusted-only helper
  as the emit side (never from request headers). Existing status / issuer /
  audience / time / signature checks are unchanged.
- `InResponseTo` is now mandatory and single-use. `saml_login` persists the
  AuthnRequest id via a new `create_sso_session_with_state` that reuses the
  existing `sso_sessions` table (migration 037, `state` is `UNIQUE`); on the
  callback, `authenticate()` requires the response to carry an `InResponseTo`
  that matches a pending, unexpired, not-yet-consumed session and consumes it
  (`DELETE ... RETURNING`). Missing / unknown / expired / replayed values are
  rejected, and unsolicited (IdP-initiated) assertions — which AK never
  requested — are refused. AK is SP-initiated only.

No new migration is required for #2096 (reuses `sso_sessions`), and no
`sqlx::query!` macros were added (SAML uses `query_as`), so no
`cargo sqlx prepare` is needed.

Closes #2038
Closes #2096

Builds on and updates @StringKe's `fix/saml-absolute-acs-url` branch; the
absolute-ACS half is @StringKe's work, the response-binding half is layered on
top. (Supersedes the earlier draft note referencing "migration 138" and the
`RequestBaseUrl` extractor — the column is migration **139** and the trusted
base comes from `trusted_external_url()` / `AK_EXTERNAL_URL` only.)

New tests:
- Pure: `build_saml_acs_url` (flag off → relative, on+trusted → absolute,
  trailing-slash, subpath, on+no-trusted → relative fallback, and a **red-team
  regression** proving no request header is an input so a spoofed
  `Host`/`X-Forwarded-Host` cannot change the emitted ACS); `acs_urls_match`
  trailing-slash normalisation; `validate_response` Destination/Recipient
  mismatch→reject, match→ok, absent+expected→ok (permissive), no-expected→skip
  (back-compat); parser extracts Destination + Recipient.
- DB-gated (skip without `DATABASE_URL`): `create_sso_session_with_state`
  round-trip + single-use consume; `authenticate` consumes a matching
  `InResponseTo`, rejects replay, rejects unknown, rejects absent;
  `SamlConfigRow` round-trips `use_absolute_acs_url` across every SQL path.

## Test Checklist
- [x] Manually tested locally

Verification:
- `cargo fmt --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `cargo test --workspace --lib` (with a throwaway Postgres): **11817 passed, 0 failed, 5 ignored**.
- Deployed on an isolated worker-pool stack with `AK_EXTERNAL_URL` set and proved:
  (a) flag on → login emits the absolute ACS from `AK_EXTERNAL_URL`;
  (b) spoofed `Host`/`X-Forwarded-Host` → ACS stays the trusted host, never the attacker;
  (c) flag off → relative ACS (unchanged);
  (d) response with mismatched `Destination`/`Recipient` → 401 rejected;
  (e) unknown / replayed `InResponseTo` → 401 rejected;
  (f) a legit full login succeeds (307 with exchange code).

## API Changes
- [ ] N/A - no API changes

Migration 139 is reversible via `ALTER TABLE saml_configs DROP COLUMN use_absolute_acs_url;`
(the value is a per-provider config bool, no data loss). #2096 adds no migration.


brandonrc:
Thanks for this — the `build_saml_acs_url` helper is well done (single shared helper for emit + validate so the two sites can't drift, trailing-slash normalization, and it's correctly default-off).

One security item before we can merge, then this is good to go. When the flag is enabled, the absolute ACS URL is built from `RequestBaseUrl`, whose precedence includes request headers (`X-Forwarded-Proto` / `X-Forwarded-Host` / `Host`). That URL feeds two security-relevant sinks:
- `saml_login` embeds it as `AssertionConsumerServiceURL` in the outbound AuthnRequest — a spoofed `Host` could advertise an attacker ACS, and a permissive IdP would POST the signed assertion there.
- `saml_acs` recomputes the same value from the callback request to validate the asserted `Destination`/`Recipient` — attacker-controlled `Host` on the callback tunes the expected value.

Please source the absolute ACS URL from the trusted `AK_EXTERNAL_URL` config only when the flag is on (fail closed / refuse to enable if it's unset), rather than from request headers. That keeps a security-relevant SAML URL off attacker-influenceable input. With that change this is a clean merge.

(`AK_EXTERNAL_URL` already takes precedence in the extractor, and the `Audience`/`sp_entity_id` binding remains the primary control, so the residual risk is narrow — but for a URL that delivers/validates a signed assertion we want it explicitly config-sourced.)

brandonrc:
Red-team verification on an isolated instance **confirmed this is exploitable** as flagged, with a live PoC:

With `use_absolute_acs_url: true` and `AK_EXTERNAL_URL` unset, `GET /api/v1/auth/sso/saml/<id>/login` reflects request headers straight into the outbound AuthnRequest's `AssertionConsumerServiceURL`:

| headers | emitted ACS URL |
|---|---|
| `X-Forwarded-Host: evil.attacker.com`, `X-Forwarded-Proto: https` | `https://evil.attacker.com/api/v1/auth/sso/saml/<id>/acs` |
| `Host: evil2.attacker.com` | `http://evil2.attacker.com/.../acs` |
| no spoof | real host |
| `AK_EXTERNAL_URL=https://legit.company.com` + spoof | `https://legit.company.com/...` (spoof ignored) |
| flag OFF + spoof | relative path (safe, unchanged) |

So an attacker can craft an AuthnRequest instructing the IdP to POST the SAML assertion to an attacker origin (assertion/identity theft). `sign_requests` doesn't help — the SP signs the attacker-injected host.

**Required change to merge:** when `use_absolute_acs_url=true`, build the ACS URL from a **trusted source only** (`AK_EXTERNAL_URL` / per-provider configured SP base), never from `Host`/`X-Forwarded-Host`. If no trusted base is set, hard-error or fall back to relative. Code: `build_saml_acs_url` + `request_base_url` (proxy_helpers.rs).

`AK_EXTERNAL_URL` mitigates in practice and the flag-off path is safe, so this isn't a live prod critical — but it shouldn't land until the ACS host comes from trusted config.

StringKe:
Thanks for the red-team PoC — that's exactly the failure mode. Fixed in edf8f8b:

- New `pub fn trusted_external_url() -> Option<&'static str>` in `api::extractors` returns `AK_EXTERNAL_URL` if set, otherwise `None`. The existing `RequestBaseUrl` extractor is intact for display-URL callers that legitimately want the request-header fallback.
- `build_saml_acs_url` signature is now `(bool, Option<&str>, Uuid)`. With `use_absolute=true`:
  - `Some(trusted_base)` — rebased URL (unchanged behaviour).
  - `None` — **fail closed to the relative form** and emit `tracing::warn!` naming the misconfiguration. Went with fallback-plus-warn rather than hard error so the SAML flow stays available for the same operators the flag exists to help; the relative form is the pre-138 baseline and cannot steer the IdP anywhere hostile.
- `saml_login` and `saml_acs` no longer take a `RequestBaseUrl` extractor; both call `trusted_external_url()` directly. Same trusted-only value on both sides, so the two sites cannot drift, and neither can be tuned by a spoofed request header.

Same PoC table re-run mentally on this branch:

| headers                                            | new ACS URL                                                                        |
|----------------------------------------------------|------------------------------------------------------------------------------------|
| `X-Forwarded-Host: evil.attacker.com`, `X-Forwarded-Proto: https`, `AK_EXTERNAL_URL` unset | `/api/v1/auth/sso/saml/<id>/acs` (relative fallback, safe) |
| `Host: evil2.attacker.com`, `AK_EXTERNAL_URL` unset | `/api/v1/auth/sso/saml/<id>/acs`                                                   |
| `AK_EXTERNAL_URL=https://legit.company.com` + spoof | `https://legit.company.com/...` (spoof ignored, unchanged from before)             |
| flag OFF + spoof                                   | `/api/v1/auth/sso/saml/<id>/acs` (unchanged)                                       |

Attacker-controllable host never reaches the signed `AssertionConsumerServiceURL`.

Tests: 5 existing `test_build_saml_acs_url_*` cases are adapted to the new `Option<&str>` signature, and one new `test_build_saml_acs_url_absolute_falls_back_to_relative_when_no_trusted_base` pins the fail-closed contract explicitly against the very PoC vector you found (flag ON + no trusted base MUST NOT reach for anything spoofable). `cargo fmt --check` / `cargo clippy --workspace --all-targets -- -D warnings` / `cargo test --workspace --lib -- build_saml_acs_url` (6/6) all clean.

Note re. #2096 (the `Recipient` / `Destination` / `InResponseTo` gap you surfaced during red-team): confirmed pre-existing and dependent on this PR's trusted-URL helper as the expected-value source. I'll open a companion PR against it once this one lands — that PR will `Fixes #2096` and add the missing SP-side validation plus the `RelayState`+`sso_sessions.saml_request_id` wiring the callback needs to compare `InResponseTo` against the AuthnRequest it belongs to.

PR #2041, opened 2026-06-26
feat(sso): per-provider opt-in for OIDC ID tokens signed with sub-2048-bit RSA keys

### LINKED ISSUE #2039: OIDC: opt-in compatibility for IdPs signing ID tokens with sub-2048-bit RSA keys (Lark AnyCross)
### Problem Statement

`backend/src/api/handlers/sso.rs::validate_id_token` verifies OIDC ID tokens through `jsonwebtoken` 10.3 + `aws_lc_rs`, which enforces RFC 7518's effective baseline by rejecting every RSA modulus below 2048 bits. That is correct, and matches NIST SP 800-131A Rev.2 (RSA < 2048 bits disallowed since 2014-01-01) and OWASP ASVS 4.0 V6.2.5 (RSA modulus ≥ 2048 bits for digital signatures) — the default behaviour is the right default.

However, some real-world enterprise IdPs still publish 1024-bit RSA public keys on their JWKS. The motivating case is Lark AnyCross (Feishu's enterprise SSO offering): the OIDC issuer at `https://anycross-sg.larksuite.com/sso/<app-id>` exposes an `alg=RS256` key with a 1024-bit modulus (`kid=03fcdd41-1136-40c9-b6aa-11f3c3a2c9d6` at time of writing); the X.509 certificate's issuer is `CN=anycross.feishu.cn`, valid 2025-12-20 → 2036-05-25 — i.e. this is the IdP vendor's current and intended key, not a misconfiguration on the customer side.

For these IdPs every callback fails with `Authentication("ID token validation failed: InvalidSignature")` with no hint that the rejection is key-strength rather than nonce / audience / clock skew, which makes the failure mode hostile to diagnose. We reached out to Lark to ask them to upgrade their JWKS; in the meantime their other downstream applications (Java Nimbus JOSE, Go go-oidc, several Rust clients) are subject to the same problem, and other Rust-stack OIDC clients (Argo CD, Kargo, Grafana, Vault, Keycloak, etc.) hit the same wall.

### Proposed Solution

Add a per-provider compatibility flag `oidc_configs.allow_legacy_rsa_keys BOOLEAN NOT NULL DEFAULT false` (migration 139). When the flag is enabled on a specific provider, ID tokens signed with an RSA key whose modulus is strictly below 2048 bits are accepted through a restricted manual `rsa` + `sha2` PKCS#1 v1.5 fallback path. The fallback is scope-limited at the application layer:

- Algorithm whitelist is **RS256 / RS384 / RS512 only**. `PS*` (RSA-PSS), `ES*` (ECDSA), `HS*` (HMAC) never engage the legacy path regardless of the flag.
- Engaged only when the JWKS publishes an RSA key whose modulus, after stripping leading-zero padding, is strictly below 2048 bits. A 2048-or-larger RSA key (and every EC / OKP key) keeps the strict `aws_lc_rs` path.
- Audience, issuer, exp and nbf are validated on every legacy-path authentication (60-second leeway matches jsonwebtoken's strict-path default). The nonce check is unchanged.

Default-off keeps every existing deployment on the OWASP-baseline strict path with no behavioural change on upgrade. When the flag is off and a sub-2048-bit RSA key is detected, the resulting `Authentication` error names the flag so the failure is immediately actionable instead of looking like a generic `InvalidSignature`.

### Alternatives Considered

- Adding the field at the SDK / type-layer only and routing through an HTTP proxy or pre-processor: rejected because it pushes the responsibility to the operator and forks the security baseline.
- Replacing `aws_lc_rs` with a different verifier crate that allows smaller moduli: rejected because the global default should keep enforcing the OWASP baseline; the per-provider opt-in scopes the weakening exactly where it is needed.
- Waiting for Lark / Feishu to upgrade their JWKS to ≥ 2048-bit: outside our control; the customer-side outage continues meanwhile.

### Use Case

As a DevOps engineer rolling out Lark AnyCross SSO for an internal team, I want to opt that specific OIDC provider into accepting its 1024-bit RSA-signed ID tokens, so that affected users can log in without lowering the security baseline for every other OIDC provider in the same deployment.

### Component

### Additional Context

PR to follow this issue. Migration 139 adds the column with `DEFAULT false` so existing deployments stay on the strict path on upgrade with zero behavioural change.

### Pre-submission Checklist

StringKe:
## Summary

The OIDC ID-token signature path goes through `jsonwebtoken` 10.3 + `aws_lc_rs`, which enforces RFC 7518's effective baseline by rejecting every RSA modulus below 2048 bits. That keeps Artifact Keeper aligned by default with NIST SP 800-131A Rev.2 (RSA < 2048 disallowed since 2014-01-01) and OWASP ASVS 4.0 V6.2.5 (RSA modulus ≥ 2048 bits for digital signatures), but it also strands operators whose IdP still publishes a 1024-bit RSA key on its JWKS — Lark AnyCross's OIDC offering at the time of writing is one such IdP (`kid=03fcdd41-1136-40c9-b6aa-11f3c3a2c9d6`, 1024-bit RS256). The downstream symptom is `Authentication("ID token validation failed: InvalidSignature")` on every callback, with no actionable hint that the problem is key-strength rather than nonce / audience / clock skew.

This change adds a per-provider compatibility flag: `oidc_configs.allow_legacy_rsa_keys` (migration 144, defaults to `false`). When the flag is enabled on a specific OIDC provider, ID tokens signed with an RSA key whose modulus is strictly below 2048 bits are accepted through a restricted manual `rsa` + `sha2` PKCS#1 v1.5 fallback. The fallback is scope-limited at the application layer:

- Algorithm whitelist is **RS256 / RS384 / RS512 only**. `PS*` (RSA-PSS), `ES*` (ECDSA), and `HS*` (HMAC) tokens never engage the legacy path regardless of the flag — the verifier returns `Authentication("Legacy RSA fallback only accepts RS256 / RS384 / RS512")` instead.
- The legacy path runs only when the JWKS publishes an RSA key whose modulus, after stripping leading-zero padding, is strictly below 2048 bits. A 2048-or-larger RSA key (and every EC / OKP key) keeps the strict `aws_lc_rs` path.
- Audience, issuer, exp, and nbf are validated on every fallback authentication. `jsonwebtoken`'s `Validation::insecure_disable_signature_validation` is deprecated in 10.x; we use the new `jsonwebtoken::dangerous::insecure_decode` to extract claims after the manual RSA verification has succeeded, and run an explicit `validate_legacy_rsa_claims` helper (same 60-second leeway as jsonwebtoken's default validator). Audience accepts both string and string-array forms (RFC 7519 §4.1.3). The nonce check sits above the branching point so the same nonce contract applies to both paths.

Default-off policy:

- Migration 144 column is `DEFAULT false`; every existing OIDC configuration stays on the strict path on upgrade with zero behavioural change. An operator must explicitly opt their provider in.
- When a sub-2048-bit RSA key is detected and the flag is off, the resulting `Authentication` error names the flag so the failure is immediately actionable instead of looking like a generic signature reject.

Implementation surface:

- `OidcConfigRow` / `OidcConfigResponse` / `CreateOidcConfigRequest` / `UpdateOidcConfigRequest` and the matching SQL (list / get / get_decrypted / create / update / toggle) are updated together.
- `From<CreateOidcConfigRequest> for UpdateOidcConfigRequest`, `bootstrap_oidc_from_env`, and the e2e fixture propagate the new field as `None` (= preserve existing).
- `validate_id_token` gains an `allow_legacy_rsa_keys: bool` parameter threaded from the OIDC config row at the callback site.
- The previous `select_jwk_key` (which returned a `DecodingKey` and therefore eagerly rejected unsupported `kty`) is replaced by `select_jwk` (returns the raw JWK so the caller can inspect modulus length before choosing strict vs. legacy path). The OKP-rejection contract moves into the existing `decoding_key_from_jwk` tests; three stale `select_jwk_key` tests are removed in favour of the new `select_jwk` coverage.

This is a `feat/` PR adding new opt-in functionality. 19 new unit tests in `api::handlers::sso::tests::legacy_rsa` cover every behavioural contract introduced here:

- `rsa_modulus_bit_len`: reports ~1024 for 1024-bit, ~2048 for 2048-bit, strips leading-zero padding correctly, returns `None` for non-RSA `kty`, returns `None` when `n` is missing.
- `select_jwk`: returns the JWK matching the requested `kid`, returns an `Authentication` error when `kid` is absent, picks the first RSA / EC key when no `kid` is given, returns an error when no usable key exists.
- `verify_id_token_with_legacy_rsa` (1024-bit RSA fixture, generated once via `OnceLock` so keygen cost amortises across the suite): accepts a valid RS256-signed token, accepts RS384, accepts RS512, rejects ES256 even with the flag on, rejects PS256 even with the flag on, rejects a tampered signature, rejects a wrong audience, rejects a wrong issuer, rejects an expired token, rejects a malformed (non-three-segment) token.

## Test Checklist
- [x] Manually tested locally

Verification:

- `cargo fmt --check` clean
- `cargo clippy --workspace --all-targets -- -D warnings` clean (no deprecated `insecure_disable_signature_validation` warning — we use the new `dangerous::insecure_decode`)
- `cargo test --workspace --lib -- --test-threads=1`: 11806 passed, 5 ignored. The only failure (`http_client::test_large_object_client_builder_does_not_apply_total_timeout`) is a pre-existing TCP timing flake unrelated to this change; passes in isolation.
- `jscpd` on the changed `sso.rs`: 2.23% duplicated lines / 2.50% tokens, under the 3% gate.

## API Changes
- [ ] N/A - no API changes

The new column is reversible via `ALTER TABLE oidc_configs DROP COLUMN allow_legacy_rsa_keys;`. The fallback verification logic is gated by the column, so dropping the column re-pins every provider to the strict path automatically.

Fixes #2039


brandonrc:
This is solid work — I reviewed it carefully and it's default-secure: the migration column defaults `false`, the legacy branch only fires for `<2048`-bit RSA keys on providers that explicitly opt in, and the strict `aws_lc_rs` path is unchanged for everyone else. The legacy path itself looks correct (RS256/384/512 only, PS/ES/HS rejected, signature verified before claims, nonce/aud/iss/exp all checked, good test coverage).

That said, I'm moving this to the **v1.3.0** milestone rather than the v1.2.4 patch, for two reasons:
1. It's a `feat` (new opt-in capability + new column) — semver-wise that belongs in a minor, not a patch.
2. It introduces a hand-rolled RSA signature-verification path into the authentication hot path, deliberately routing around the vetted `aws_lc_rs` implementation. Even when the code reads correctly, bespoke crypto verification in the auth callback is exactly the category we want to land with a dedicated crypto review rather than in a fast patch.

Nothing to fix on your end right now. For 1.3.0 we'll want an independent review of the legacy RSA verification (verification-parity vs the strict validator) and confirmation the `rsa`/`sha2` crates are already workspace deps. Appreciate the thoroughness on the tests and the opt-in gating.

brandonrc:
Red-team verification on an isolated instance: **all four attack classes refuted live** (18 adversarial cases, full auth-code flow against a controlled mock IdP):
- **Isolation:** non-opted 1024-bit provider → rejected; the `<2048` branch checks `allow_legacy_rsa_keys` before anything.
- **alg-confusion:** `alg:none`, HS256-over-RSA-pubkey, ES256-vs-RSA-JWK all rejected; legacy branch only accepts RS256/384/512.
- **Signature bypass:** attacker-key RS256, byte-flipped sig, wrong `aud`, expired, nonce-replay all rejected. The manual PKCS#1 v1.5 verification is sound; aud/iss/exp/nbf/nonce still enforced.
- **Strict path unchanged:** 2048-bit + EC providers still flow through the strict `aws_lc_rs` path.

So the opt-in is truly isolated and the legacy RSA path is cryptographically sound. Keeping this on **v1.3.0** purely for semver (it's a `feat` + new column), not for any security concern — it has red-team sign-off.

PR #2055, opened 2026-06-29
fix(repositories): purge proxy cache on repository delete

### LINKED ISSUE #2047: Recreating a RPM remote with the name of a deleted RPM remote brings back the deleted remote.
### Description

It doesn't seem to delete the files.

### Environment Information

- Deployment type: helm
- Operating System: linux
- Architecture: x64
- Database: postgres
- Reverse Proxy: envoy
- Network setup: cilium

### Backend Version

latest

### Web UI Version

latest

### Package Format

RPM

### Expected Behavior

the repo would come back empty

### Observed Behavior

the repo comes back with the old files

### Steps to Reproduce

create remote repo
download something through it
delete it
recreate it

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

Closes #2047

Deleting a remote (proxy / pull-through) repository did not remove its
proxy-cache objects from storage. Proxy-cached content is keyed by the
repository **key** (`proxy-cache/<repo_key>/<path>/{__content__,__cache_meta__.json}`)
and is intentionally not recorded in the `artifacts` table (#1278). The
repository-delete cleanup only purges objects discoverable through `artifacts`
rows (`purge_repo_artifact_objects`) and OCI upload temp objects, so the entire
`proxy-cache/<repo_key>/` subtree — cached bodies, metadata sidecars, and
negative-cache (404) sidecars — survived the delete.

Because cache keys derive from the repository key, a new repository created
later with the **same key** resolves to the same cache keys and serves the
previous repository's cached upstream content instead of fetching from its own
upstream. When the new repository points at a different upstream this is a
content-integrity / supply-chain hazard: the bytes served no longer correspond
to the configured source.

This change purges the repository's whole proxy-cache subtree on delete:

- Adds `ProxyService::purge_repo_cache(repo_key)`, which lists the
  `proxy-cache/<repo_key>/` prefix on the same global-default storage backend
  the cache writer uses and deletes every returned key. Listing raw keys (rather
  than reconstructing logical paths) covers content bodies, metadata sidecars,
  **and** negative-cache sidecars that have no `__content__` companion, so a
  previously-cached 404 is re-evaluated against the new upstream too.
- `delete_repository` calls it best-effort after the existing artifact/OCI
  purges: a listing or per-object failure is logged but never blocks the delete.
  The prefix is repo-key scoped, so it is a harmless no-op for hosted
  repositories and never touches a sibling repository's cache.

Scope is limited to the delete path (the filed bug). Re-pointing a remote's
upstream via update is a related cache-coherence consideration but is out of
scope here and unchanged by this PR.

Adds unit tests in `proxy_service` covering `purge_repo_cache`:
- purges the entire `proxy-cache/<repo_key>/` subtree (content + metadata +
  negative-cache-only sidecar) while leaving a sibling repo's cache untouched;
- is a clean no-op for a repo with no cache (hosted);
- surfaces a listing failure as `Err` so the caller can log it (best-effort).

## Test Checklist
- [x] Manually tested locally

Manual verification (isolated instance): create a remote `K` -> upstream A,
fetch a path (caches it), delete `K`, recreate `K` -> a different upstream B,
fetch the same path -> serves B's content, not the stale A blob. Storage
inspection confirms the `proxy-cache/<K>/` subtree (including the negative-cache
404 sidecar) is removed on delete; hosted repo create/delete is unaffected; the
existing artifact-object purge on proxy-repo delete still runs.

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -1882,6 +1882,30 @@ pub async fn delete_repository(
     // (#1551). Best-effort: never blocks the delete.
     purge_repo_artifact_objects(&state, repo.id, &repo.storage_location()).await;
 
+    // Purge this repo's proxy-cache subtree from the global default storage
+    // backend. Proxy-cached blobs are keyed by the repo KEY and are not tracked
+    // in `artifacts` (#1278), so the purge above never reaches them; left
+    // behind, a new repository created with the same key would serve the deleted
+    // repo's stale upstream content (#2047). Best-effort: never blocks the
+    // delete. Hosted repos have no proxy cache, so this is a no-op for them.
+    if let Some(proxy) = state.proxy_service.as_ref() {
+        match proxy.purge_repo_cache(&repo.key).await {
+            Ok(purged) if purged > 0 => tracing::info!(
+                repo_id = %repo.id,
+                repo_key = %repo.key,
+                purged,
+                "Purged proxy-cache storage objects for deleted repository"
+            ),
+            Ok(_) => {}
+            Err(e) => tracing::warn!(
+                repo_id = %repo.id,
+                repo_key = %repo.key,
+                error = %e,
+                "Failed to list proxy-cache objects to purge before repository delete"
+            ),
+        }
+    }
+
     service.delete(repo.id).await?;
 
     // Remove the deleted repo from the in-memory cache.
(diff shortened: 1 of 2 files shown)


simonfelding:
Thank you so much @brandonrc!!!

>Re-pointing a remote's
upstream via update is a related cache-coherence consideration but is out of
scope here and unchanged by this PR.

This would be nice too hehe, but it's not as big an issue. Thanks so much for fixing it!

Issue #1770, opened 2026-06-09
Package Age Policy not enforced on remote/proxy repositories; setting not returned on read


younsl:
# Package Age Policy not enforced on remote/proxy repositories; setting not returned on read

## Summary

Package Age Policy (quarantine) does not work on remote/proxy repositories. Two distinct defects:

- A: Packages fetched through a remote/proxy repository are never quarantined, so the cooldown window is never enforced on downloads.
- B: The configured policy is saved but never returned by the repository GET, so the UI toggle resets to OFF on reload.

## Environment

- Backend appVersion 1.2.0 (Helm chart artifact-keeper 1.2.1, image tag latest), self-hosted on Kubernetes.
- Repository: `repo_type: remote`, `format: npm`, upstream `https://registry.npmjs.org`.

## Steps to reproduce

1. Create a remote npm repository (upstream registry.npmjs.org).
2. Enable Age Policy with a 30 day cooldown. Either the UI, or `PATCH /api/v1/repositories/{key}` with `{"quarantine_enabled": true, "quarantine_duration_minutes": 43200}` (returns 200).
3. Reopen the repository settings page.
4. Pull a package version published less than 30 days ago, for example `GET /npm/{key}/axios/-/axios-<recent>.tgz`.
5. Pull it again (now cached).

## Expected

- Step 3: the Age Policy toggle shows enabled with 30 days.
- Step 4 and 5: a package released inside the cooldown window is blocked with 409 Conflict.

## Actual

- Step 3: the toggle shows OFF (Finding B).
- Step 4 and 5: the tarball is served with 200 for any age, including versions published a few days ago (Finding A).

## Root cause

Finding A. The proxy cache insert does not set quarantine columns:

- `backend/src/api/handlers/proxy_helpers.rs`, `INSERT INTO artifacts (repository_id, path, name, version, size_bytes, checksum_sha256, content_type, storage_key, uploaded_by)`. No `quarantine_status` / `quarantine_until`, so they default to NULL.
- Quarantine is applied only on direct upload: `backend/src/services/artifact_service.rs` calls `quarantine_service::should_quarantine` / `set_quarantine` inside `upload_with_sync_options`. The proxy cache path does not go through this.
- `backend/src/services/quarantine_service.rs::check_download_allowed` treats NULL status as downloadable (`_ => Ok(())`), so cached proxy artifacts are never blocked.

Finding B. The setting is persisted but not read back:

[...]

PR #2066, opened 2026-06-29
feat(age-gate): adopt proxy quality gate with coverage

### LINKED ISSUE #1558: Add remote proxy publish-age cooldown/quarantine policy for pull-through caches

cazlo:
## Summary

Continues @RajatGarga's #1403: a publish-age cooldown ("age gate") for **remote (proxy) npm and PyPI repositories**, from #1558. Young upstream versions are withheld from metadata listings and blocked at download until they cross a per-repository age threshold, land in an admin review queue, and auto-approve once aged (5-min idempotent sweep + lazy pass-through). Blocked downloads return HTTP 451 with a structured JSON body.

Not to be confused with `promotion_policy_service::evaluate_age_gates` (staging-promotion min-age, unrelated): this PR's gate is the **remote proxy publish-age gate**, named `age_gate_*` throughout.

Rebased 2026-07-05 onto `main` (8eacc827, 76 commits of drift); see the rebase summary comment for conflict-resolution details and authorship handling.

### What's in the rebase beyond #1403

- **PEP 691 JSON simple-index filtering** — `main` now serves `application/vnd.pypi.simple.v1+json` on the proxy path (#1944) and modern pip prefers it; an HTML-only filter is a bypass. The JSON filter reads PEP 700 `upload-time` from the document itself (usually no extra Warehouse round-trip; the fallback fetch only covers versions without it).
- **Computed-packument cache bypass** — age-gated repos skip the #2166 SWR cache: filtered bytes are never stored in the shared cache, and the SWR fast path can never serve an unfiltered packument around the gate. Ungated repos keep the full #2162 speedup.
- **npm abbreviated (corgi) parity** — verified: the filter runs on the full packument before local abbreviation, so both representations hide the same versions.
- **Bounded reads** — the one uncapped buffered packument read (download-gate path) now uses `proxy_fetch_capped` (#2181), and the in-process publish-time cache is capped at 4096 entries (attacker-influenced key space; #1607 lesson).
- **`min_age_days = 0`** — trusted-remote mode per #1558 (e.g. Chainguard): no age delay, but explicit rejections still block and the review queue stays under admin control (which `age_gate_enabled = false` does not provide). Server validation + DB CHECK are 0..=3650, default 7.
- **CI actually runs the suite** — `age_gate_tests` added to the integration-test enumeration; it was never listed, so none of its cases ran in CI.
- Migration renumbered to `143_age_gate.sql` (upstream took 138).

### Fail-open/fail-closed semantics (deliberate, review welcome)

Downloads are the enforcement point and fail **closed**: a missing or unparseable publish time is treated as younger-than-threshold; future-dated publishes clamp to age 0 (young). Listings are best-effort presentation: if the publish-time fetch fails, the HTML index serves unfiltered (original #1403 semantics) while the JSON index withholds timeless versions. A metadata outage can therefore momentarily reveal young versions in a listing, but can never serve one.

### Review-queue volume note

The listing filter records one review row per young version (debounced, near-write-free re-listing). For a package with a large young-version count (e.g. `@types/node` under a multi-year threshold), that is intentionally thousands of rows; the sweep auto-approves them as they age.

### Non-goals in this PR (tracked for follow-up; part of #1558)

1. Reject ⇒ purge/disable of already-cached entries (touches proxy-cache model + #1278 storage routing).
2. Upstream yank/malware-intel ⇒ auto-purge.
3. Admin API to purge a proxy-cache entry by package identity.
4. Virtual repos: per-member policy evaluation (v1 limitation is documented loudly in the module doc; virtual pass-through unchanged).
5. Member-scoped purge in virtual repos (depends on 1/3).
6. Ecosystems beyond npm/PyPI (other formats pass through).
7. Capping the Warehouse JSON metadata read (passes the #1608 gate — not an artifact path — but is an unbounded buffered read; candidate for the capped-read helpers).

Partially fixes #1558 (the follow-up tranche above stays open on the issue).

## Test Checklist

- [x] `--lib` units for decision/parse/format logic (incl. PEP 691 JSON collect/fill/apply helpers, packument-cache bypass classifier, publish-time cache bound)
- [x] DB-backed integration suite `age_gate_tests` (now actually enumerated in CI): block/queue/lazy-approve/reject/sweep/threshold + HTML and JSON listing-filter regressions (document-times and fallback-times variants)
- [x] Full fmt/clippy/unit/coverage + integration matrix green on private staging CI against this exact tree
- [x] Real `npm` client exercised against a deployed build of this branch: filtered full+corgi packuments, structured 451 on blocked downloads, trusted-remote `min_age_days=0` install, `dist-tags.latest` repointing to newest allowed version, scoped packages, `npm ci` against a lockfile pinning a blocked version fails with `npm error code E451 ... age_gate_blocked`, and the review lifecycle end-to-end (queue → lazy auto-approve → sticky approval → manual approve opens the download)

## API Changes

- `GET/PUT /api/v1/repositories/{key}/age-gate` (admin) — per-repo config.
- `GET /api/v1/admin/age-gate/reviews[?status=...]`, `GET .../reviews/{id}`, `POST .../reviews/{id}/approve`, `POST .../reviews/{id}/reject` (admin).
- Webhook events: `age_gate.queued`, `age_gate.approved`, `age_gate.rejected`.
- Blocked downloads: HTTP 451 `age_gate_blocked` JSON.

Closes #1558


cazlo:
Rebased onto `main` @ 8eacc827 (76 commits of drift). Notes for @RajatGarga and reviewers:

- **Authorship**: all 10 of Rajat's commits keep his authorship; the rebase only re-signs them (committer me). No co-author trailers added.
- **Conflicts** were `npm.rs`, `pypi.rs`, and `.sqlx` churn:
  - `serve_tarball` / `serve_file` moved to streaming helpers upstream (#2192 / #2203 / #895); the download gate + LKG fallback were ported to `proxy_fetch_streaming_with_cache_key` / `proxy_check_cache_streaming` / `fetch_from_pypi_remote_streaming` — same logic, streaming APIs. The LKG path also attempts the #1555 presigned redirect before streaming, matching the main path's ordering.
  - `simple_project` gained the PEP 691 JSON branch upstream (#1944); the HTML filter hook reattached after `rewrite_upstream_urls`, and the JSON branch got its own filter as a new commit — that gap was a real listing bypass for JSON-negotiating clients.
  - `.sqlx` conflicts resolved toward upstream; offline cache verified complete against the rebased tree.
- **New commits on top** (mine): PEP 691 JSON index filter; packument-SWR-cache bypass for gated repos + capped download-gate packument read; `age_gate_tests` CI enumeration (the suite was never listed, so it had never run in CI); `min_age_days = 0` trusted-remote mode; a bound on the in-process publish-time cache; post-rebase test-fixture compile fixes; presign-ordering fix on the LKG path.
- **Migration** renumbered 138 → 143 (upstream took 138 for `scanner_service_account`).
- **PR body updated**: `Part of #1558` with an explicit non-goals list — the purge/yank/virtual-member tranche stays open on the issue (this matches the linked-issue gate precedent from #2202).

Validation: fmt/clippy/unit/coverage plus the full Postgres-backed integration matrix (including the newly-enumerated `age_gate_tests`) green on private staging CI against this exact tree, and a real `npm` client exercised against a deployed build — filtered packuments (full and corgi), structured 451s, `min_age_days=0` pass-through, `dist-tags.latest` repointing, scoped packages, an `npm ci` block with a legible `E451` error, and the queue → lazy-auto-approve → manual-approve lifecycle end-to-end.

Happy to split any of the new commits out if you'd rather keep this strictly-rebase.

cazlo:
To sanity-check this PR's extension points against other package types like nuget, golang, etc., I prototyped the next format tranche on a branch (not proposing it for this PR — it's evidence for the follow-up slicing, and for reviewing whether the extension points here are the right ones):

**Branch:** https://github.com/cazlo/artifact-keeper/tree/age-gate-format-expansion
(4 commits on top of this PR's branch)

**What it adds**

1. **A cross-cutting `first_seen` age-source mode** for formats where upstream  publish metadata is missing or untrustworthy: the cooldown timer starts the   first time *this* server observes a version (per-repo   `age_gate_mode = upstream_publish_time | first_seen`, plus an insert-once   observations table — `ON CONFLICT DO NOTHING`, so concurrent requests and    replicas keep the earliest observation and repeated listings can never move   it). It is an AK-local freshness control, not true publish age, and the   config surface treats it as an explicit mode, never a silent fallback.
2. **Go module proxy gate** — `@v/list` filtering plus   `.info`/`.mod`/`.zip`/`@latest` gates. Publish-time mode reads the `.info`   `Time`; since that is VCS tag time (publisher-suppliable/backdatable), the   Go gate in that mode is documented as advisory — which is exactly the case   `first_seen` exists for, and Go repos accept either mode.
3. **NuGet remote gate via `first_seen`** as the first consumer of the   pattern — flat-container index + inline registration-page filtering and a   download gate on the flat-container route. (NuGet's registration
   `catalogEntry.published` is registry-assigned and is the natural   publish-time source later; kept out of the spike to keep the pattern test   clean.)

Enabling the gate validates the (format, mode) pair server-side and rejects combinations no handler enforces, so a config that would silently never gate is impossible.

**Measured incremental complexity** (on top of this PR)

| Piece | Size |
| --- | --- |
| Cross-cutting core (mode + migration + observation service + Go `.info` fetch/cache + unit tests) — written once | +1,029 / −12 (5 files) |
| Go handler hooks | +213 (goproxy.rs only) |
| NuGet handler hooks | +164 / −9 (nuget.rs only) |
| DB-backed integration tests (5 new scenarios) | +372 |
| **Total** | **+1,778 / −22 across 8 files** |

The important number: once the core exists, a new format costs roughly **~450 lines including tests** (NuGet's increment), and it decomposes into exactly three per-format tasks — identity extraction, filtering *every* listing representation the client can request (the corgi / PEP 691 lesson from this PR), and auditing the handler's serving paths so the download gate runs before local-cache and streaming branches alike. None of this PR's core machinery (review queue, debounced batch upserts, sweep auto-approve, config endpoints, events/metrics) needed changes — the extension points held as-is.

**Validation:** green in private staging CI (workspace clippy/fmt, the full DB-backed age-gate suite — 15 tests incl. the 5 new first_seen/Go/NuGet scenarios). Also exercised end-to-end against a test deployment with the real `go` toolchain: filtered `@v/list` as seen by `go list -m -versions`, structured 451 on `go mod download` of a blocked version, `latest` resolution repointing to the newest allowed version at a mid-range threshold, 0-day trusted-remote pass-through, sticky approvals, and first_seen first-sight blocking with zero upstream metadata fetches. No product bugs surfaced.

**Likely expansion cost for the remaining formats** (first_seen mode), from the three data points so far:

- **Cheap (~half a day each):** Composer p2, RubyGems compact index,   Terraform, Helm index.yaml`, pub.dev — one or two clean JSON/text listing  documents.
- **Medium (~1–2 days):** Maven (path-based identity; `maven-metadata.xml`  filtering must also regenerate its `.sha1`/`.md5` sidecars), Cargo sparse  index (JSON-lines rewrite), OCI tag lists (filterable, but  manifest-by-digest fetches bypass tags — needs a scoping decision).
- **Structurally limited:** Alpine/RPM/Debian signed indexes — listings
  cannot be rewritten without re-signing. For that family I'd propose an  explicit *block-only* sub-mode (downloads gated via a per-format  path→(package, version) extractor, listings untouched) rather than
  pretending listing consistency is achievable there. Block-only is  deliberately not the default anywhere else: resolvers pick versions from  listings, so unfiltered listings + blocked downloads turn "resolve an older  version" into "installation fails".

brandonrc:
Maintainer adoption update (rebased + hardened for merge). Authorship of the existing commits is unchanged; the changes below are one additional commit on top.

**Rebased** onto current `main` (`117bc06`). The rebase was clean; two mechanical post-rebase compile fixes were required where `main` had drifted:
- the LKG branch of `serve_file` now passes `index_path` to `fetch_from_pypi_remote_streaming` (the signature gained that parameter on `main`);
- the `age_gate` test's `AuthExtension` fixture was updated for `main`'s `iat_ms` field and `AccessScope`-typed `allowed_repo_ids`.

**Migration renumbered `143_age_gate.sql` → `146_age_gate.sql`.** `main` now occupies 143/144/145, so two migrations at version 143 would abort `sqlx migrate run` after merge (the differing filenames merge cleanly, which is why the collision is invisible to the mergeability check). Content is unchanged, so the `.sqlx` query cache is unaffected. Verified: `sqlx migrate run` applies 001→146 cleanly.

**Closed a PyPI download-gate asymmetry.** `serve_file` consulted the local `artifacts` table first and only ran the age gate on the cache-*miss* branch, so a young version that had an `artifacts` row (a locally-published, hydrated/replicated, or pre-#1278 cached wheel) streamed ungated — whereas npm's `serve_tarball` gates unconditionally. The gate now runs on both the cache-miss and the `artifacts`-hit branches for remote repos (extracted into `enforce_pypi_download_age_gate`, shared by both), so a withheld version returns 451 (or the last-known-good wheel) even when a row exists. Added a DB-backed `--lib` test covering the `artifacts`-row-present young-version case.

**Defense-in-depth:** explicit `require_admin()` added to `list_reviews` and `get_review` for parity with `approve`/`reject` (they previously relied solely on the `/admin` middleware).

**Verification (manual, since the integration job is push-to-main-only):**
- `cargo build --lib --tests` clean; `cargo clippy --workspace --all-targets -D warnings` clean; `cargo fmt --check` clean.
- Full `--lib` suite: 12469 passed / 0 failed (Postgres-backed).
- Author's `age_gate_tests` integration suite: 10 passed / 0 failed.
- Duplication (jscpd, changed non-exempt files): 0.84%, under the 3% gate.
- Deployed to an isolated instance and smoke-tested both ecosystems: a freshly-published version returns 451 `age_gate_blocked` and an aged version serves 200, for npm and PyPI — including the `artifacts`-hit PyPI path.

brandonrc:
**Maintainer adoption + security verification note**

Adopted and finished this PR against current `main` (red→blue loop). Two commits added on top of the original work (authorship of prior commits preserved):

- `aa5a5bd4` — migration renumber `143→146` (main already occupies 143/144/145; duplicate version would abort `sqlx migrate run`); closed a PyPI download fail-open where `serve_file` gated only the cache-miss branch (an `artifacts`-row hit streamed young bytes ungated — now gated on both branches, matching npm's unconditional gate); added `require_admin` to `list_reviews`/`get_review` for parity with approve/reject.
- `eee4479b` — closed a virtual-repo pass-through: the gate is enforced at the format-handler layer, but virtual repos' per-member fetch loops (pypi download + simple-index, npm download) never applied it, so a young version of a gated member was served through a virtual repo. Threaded the existing per-member helpers into the three virtual loops; extracted `filter_pypi_simple_html_response`; added a `Repository→RepoInfo` adapter. Does not touch the #2166 SWR cache or the shared generic virtual resolver (no maven/hex impact).

**Verification:** full `--lib` suite green; `age_gate_tests` 10/10; targeted security re-test on an isolated build confirmed fail-closed across all surfaces — npm full/corgi/tarball, PyPI PEP-691 + direct wheel (incl. Range + concurrent race), the artifacts-hit path, SWR cache toggle, missing/0-day upload-time, threshold validation, admin authz (403 non-admin / 401 anon), and the virtual paths (index + download, incl. nested/mixed-member/PEP-708/encoding/dist-tag attempts). Aged versions still resolve (no over-block regression).

**Follow-ups (separate issues, not blockers):** move age-gate enforcement into the proxy-service layer (the durable all-format fix — the older #1770 quarantine already lives there and correctly covers virtual); `GET /repositories/<key>/age-gate` config is readable by any authed read-scope user (low-sev posture disclosure).

PR #2078, opened 2026-06-30
fix(proxy): align proxy-cache write/freshness/presign paths when S3_PREFIX is set

### LINKED ISSUE #2076: Proxy-cache presigned URLs 404 when S3_PREFIX is set
## Bug

When `S3_PREFIX` is configured, proxy-cache content is written to one S3 key but presigned download URLs point to a different key, causing every presigned proxy-cache download to return HTTP 404.

## Root Cause

Two S3 backend construction paths apply the prefix inconsistently:

- **Freshness checks + presigning** — go through the primary backend built via `S3Backend::from_env()`, which reads `S3_PREFIX` and prepends it to every key.
- **Cache writes (content + sidecar)** — go through `S3BackendWrapper::from_config`, which hard-codes `prefix = None` and ignores `S3_PREFIX`.

So when `S3_PREFIX = "my-prefix"` and a package at path `simple/foo/foo-1.0.whl` is cached:

| Operation | S3 key |
|-----------|--------|
| Write (content) | `proxy-cache/repo/simple/foo/foo-1.0.whl/__content__` |
| Freshness check | `my-prefix/proxy-cache/repo/simple/foo/foo-1.0.whl/__content__` |
| Presign | `my-prefix/proxy-cache/repo/simple/foo/foo-1.0.whl/__content__` |

The freshness check never finds the content (wrong key), so `is_cache_fresh` always returns `false`. When the entry _is_ considered fresh (e.g. via the sidecar path), the presigned URL resolves to the prefixed key where no object exists → 404.

## Reproduction

1. Deploy with `S3_PREFIX` set to any non-empty value.
2. Enable presigned downloads (`PRESIGNED_DOWNLOADS_ENABLED=true`).
3. Install any package through a PyPI proxy repository — the cached entry will be written at the unprefixed key.
4. A subsequent install attempt that triggers a presigned redirect will 404.

## Expected Behaviour

All proxy-cache operations (write, freshness check, presign) should resolve to the same S3 key.

## Fix

See PR — `S3BackendWrapper::from_config` is changed to read `S3_PREFIX` from the environment, matching what `S3Backend::from_env()` already does. One-line change; existing test suite passes; new regression test added.

## Side Effects

- **`BackupService`** also uses `S3BackendWrapper::from_config`. With this fix, backup/restore operations on prefixed deployments will now correctly target the prefixed key namespace. This is the right behaviour (backups should live alongside the artifacts they back up), but operators should be aware the backup key path changes.
- **Orphaned objects**: existing deployments with `S3_PREFIX` set will have proxy-cache objects at the unprefixed `proxy-cache/…` root that were never actually served (freshness probes never found them). These are safe to prune after upgrading.

ysolomon-plat:
Fixes #2076

## Problem

When `S3_PREFIX` is set, `StorageService::from_config` forced `s3_config.prefix = None` for the proxy-cache storage backend, while the primary S3 backend (used for freshness checks and presigned URLs) applied `S3_PREFIX` via `S3Config::from_env()`. This made proxy-cache writes land at the unprefixed key while presigned URLs referenced the prefixed key — causing every presigned proxy-cache download to 404 when `S3_PREFIX` is non-empty.

## Fix

Remove the `s3_config.prefix = None` override in the `"s3"` arm of `StorageService::from_config`, so the proxy-cache backend uses the same `S3_PREFIX` as the primary backend. Writes, freshness checks, and presigned URLs now all resolve to the same S3 key.

```rust
// before
let mut s3_config = crate::storage::s3::S3Config::from_env()?;
s3_config.prefix = None;   // ← forced key mismatch when S3_PREFIX is set

// after
let s3_config = crate::storage::s3::S3Config::from_env()?;
```

The updated comment also corrects the prior incorrect description of the intended key layout.

## Side effect

`BackupService` uses the same `StorageService` backend. With this fix, backup/restore operations on prefixed deployments will now correctly target the prefixed key namespace rather than the unprefixed root. This is the intended behaviour, but operators upgrading existing prefixed deployments with prior backups should be aware the backup key path changes.

## Regression test

`test_cache_persister_write_buffered_uses_raw_cache_keys` — intercepts at the `StorageBackend` trait boundary via `PutKeyCapture`, confirming `CachePersister::write_buffered` passes the raw `CacheKeys::derive` output to the backend for both the content blob and the sidecar, with no caller-side prefix injected.

## Rollout note

Existing deployments with `S3_PREFIX` set will have orphaned proxy-cache objects at the unprefixed `proxy-cache/…` root (written before this fix, never actually served since freshness probes targeted the prefixed key). These are safe to prune after upgrading.

### DIFF
--- a/backend/src/services/proxy_service.rs
+++ b/backend/src/services/proxy_service.rs
@@ -6877,7 +6877,12 @@ SHA256:
             .expect("write_buffered must succeed");
 
         let written = capture.keys_snapshot();
-        assert_eq!(written.len(), 2, "expected content write + metadata write, got {:?}", written);
+        assert_eq!(
+            written.len(),
+            2,
+            "expected content write + metadata write, got {:?}",
+            written
+        );
         assert_eq!(
             written[0], keys.content,
             "first write must be the raw content key from CacheKeys::derive"
--- a/backend/src/services/proxy_service.rs
+++ b/backend/src/services/proxy_service.rs
@@ -6788,6 +6788,117 @@ SHA256:
     }
 
     // -----------------------------------------------------------------------
+    // CachePersister key-consistency: write path must pass raw CacheKeys keys
+    // to the backend without adding a caller-side prefix. The backend (shared
+    // with the presign path after the S3_PREFIX fix) is the single point that
+    // applies the object-key prefix, so the resulting S3 key is identical in
+    // both directions.
+    // -----------------------------------------------------------------------
+
+    use crate::services::storage_service::{
+        PutStreamResult as ServicePutStreamResult, StorageBackend as ServiceStorageBackend,
+        StorageService as RealStorageService,
+    };
+
+    /// Recording backend that captures every key passed to `put()`.
+    struct PutKeyCapture {
+        keys: std::sync::Mutex<Vec<String>>,
+    }
+
+    impl PutKeyCapture {
+        fn new() -> Arc<Self> {
+            Arc::new(Self {
+                keys: std::sync::Mutex::new(Vec::new()),
+            })
+        }
+
+        fn keys_snapshot(&self) -> Vec<String> {
+            self.keys.lock().unwrap().clone()
+        }
+    }
+
+    #[async_trait::async_trait]
+    impl ServiceStorageBackend for PutKeyCapture {
+        async fn put(&self, key: &str, _content: Bytes) -> Result<()> {
+            self.keys.lock().unwrap().push(key.to_string());
+            Ok(())
+        }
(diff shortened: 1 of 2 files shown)


ysolomon-plat:
Closing — the analysis here was wrong. StorageService::from_config already sets s3_config.prefix = None intentionally to keep proxy-cache content at the bucket root (proxy-cache/...), separate from S3_PREFIX-scoped hosted artifacts. The presign path correctly uses this same no-prefix backend via cache_storage_backend(). No code change needed upstream.

Issue #1618, opened 2026-06-04
Refactor: proper abstractions for ProxyService + proxy_helpers.rs (foundation for #1607)


brandonrc:
## Refactor: extract proper abstractions for `ProxyService` + `proxy_helpers.rs`

**Origin:** Community proposal in discussion #1614 by **@Dreamacro** ("Stop the Copy-Paste Treadmill"). Verified against `main`: `proxy_service.rs` is **6,430 lines**, `proxy_helpers.rs` is **5,094 lines** (11,524 total), and every duplication family below exists as described.

**Why this is foundational, not cosmetic:** this is the highest-defect-density code in the project (proxy-remote + virtual-repo + cache-correctness ≈ 40 defects in the labeled dataset, `docs/audits/honest-issue-analysis-2026-06-04.md`). The current copy-paste surface is *why* fixes are hard to land — @Dreamacro reports that **streaming support couldn't be merged before more buffered `Bytes`-reading logic kept landing in mainline.** Doing the behavioral invariants (#1608 streaming, #1609 single-flight, #1611 cache correctness) on top of this surface will multiply variants and guarantee fork/PR conflicts. **This refactor should land first (or alongside) as their foundation.**

## Duplication to eliminate (evidence)
1. **`fetch_artifact*` wrapper chain** — 5 fns; 4 thin wrappers + `fetch_artifact_streaming` independently duplicating the same `repo_type`/`upstream_url` guards + `cache_storage_key`/`cache_metadata_key` derivation (also repeated verbatim in `fetch_upstream_direct` / `fetch_upstream_direct_with_link`).
2. **`proxy_fetch*` — 8 variants** in `proxy_helpers.rs` with near-identical `build_remote_repo(...)` + `.map_err(map_proxy_error)` bodies.
3. **`get_cached_artifact` vs `get_stale_cached_artifact`** — ~40 lines, differing only by the `expires_at` check. Collapse via an `allow_stale: bool`.
4. **`local_fetch_by_*` family** — `local_fetch_by_path` / `_by_name_version` / `_or_redirect` share the same SQL → quarantine → `storage.get()` skeleton; only the WHERE clause differs.
5. **`cache_storage_key` / `cache_metadata_key`** — same `validate_cache_path` + `check_cache_key_length`, differing only in suffix (`__content__` vs `__cache_meta__.json`); the pair is derived together at 6+ call sites. `invalidate_cache` vs `invalidate_cache_by_key` are the same delete written twice.

[...]

PR #2081, opened 2026-06-30
feat(repositories): allow custom User-Agent on remote repository outbound requests

### LINKED ISSUE #2080: feat(repositories): allow custom User-Agent on remote repository outbound requests
## Summary

Remote repositories currently use the default reqwest User-Agent for all outbound HTTP requests to upstream registries. Some upstreams (corporate proxies, rate-limit middleware) require or benefit from a recognizable User-Agent string.

This issue tracks adding a per-repository `custom_user_agent` field that is sent on every outbound request made by Artifact Keeper when proxying through a remote repository.

## Proposed behaviour

- New optional field `custom_user_agent` on remote repositories, stored in `repository_config` under key `custom_user_agent`.
- Accepted on `POST /api/v1/repositories` (create) and `PATCH /api/v1/repositories/:id` (update).
- Returned in `GET /api/v1/repositories/:id` responses (omitted when unset).
- Passing an empty string on update removes the configured value.
- Only valid for `repo_type: remote`; rejected with a 422 for local/virtual repositories.
- Applied on every outbound `GET` to the upstream: both the buffered and streaming fetch paths, including OCI bearer-token retry requests.
- An in-memory cache (60 s TTL) avoids a DB round-trip on every upstream request.

## Validation

Input is validated against RFC 7230 §3.2.6 `field-value` rules (see [RFC 7230 §3.2.6](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6)):

- Maximum 256 characters (no RFC-imposed limit; pragmatic cap).
- Any byte below SP (0x20) except HTAB (0x09), and DEL (0x7F), is rejected.

We intentionally validate at the `field-value` level rather than enforcing the strict `product *( RWS ( product / comment ) )` grammar from RFC 7231 §5.5.3, since real-world User-Agent strings (e.g. `Mozilla/5.0 (compatible; ...)`) do not always conform to that grammar.

## Example

```http
PATCH /api/v1/repositories/npm-proxy
Content-Type: application/json

{
  "custom_user_agent": "artifact-keeper-proxy/1.0 (org: acme)"
}
```

All subsequent proxy fetches through `npm-proxy` will include:

```http
User-Agent: artifact-keeper-proxy/1.0 (org: acme)
```

v4l3nt1tx:
## Summary

Closes #2080.

Remote repositories can now be configured with a custom `User-Agent` string sent on every outbound HTTP request to the upstream registry. This is useful when the upstream is behind a corporate proxy or rate-limit middleware that routes or throttles traffic based on the User-Agent.

## What changed

**`backend/src/api/handlers/repositories.rs`**

1. New optional field `custom_user_agent: Option<String>` on `CreateRepositoryRequest`, `UpdateRepositoryRequest`, and `RepositoryResponse` (all three already derive `ToSchema`; new field picked up automatically).
2. New helper `with_custom_user_agent(db, repo_id, response)` reads from `repository_config` and populates the response field. Called only in the GET handler to avoid an unnecessary DB round-trip in create/update.
3. New `validate_custom_user_agent(ua: &str)` enforces:
   - Maximum 256 characters (no RFC-imposed limit; pragmatic cap).
   - No control characters per [RFC 7230 §3.2.6](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6) `field-value` rules: any byte below SP (0x20) except HTAB (0x09), and DEL (0x7F), is rejected.
   - Intentionally more permissive than the strict `product *( RWS ( product / comment ) )` grammar from RFC 7231 §5.5.3 to accommodate real-world User-Agent strings.
4. Create handler: validates `repo_type = remote`, validates the UA, upserts to `repository_config`.
5. Update handler: empty string removes the value, non-empty validates and upserts.
6. Get handler: populates `custom_user_agent` via `with_custom_user_agent`.
7. `CUSTOM_USER_AGENT_KEY` constant used in all SQL queries and `upsert_repo_config` calls.

No schema migration needed; `repository_config` (key/value per repo) already exists.

**`backend/src/services/proxy_service.rs`**

1. New field `user_agent_cache: RwLock<HashMap<Uuid, (Option<String>, Instant)>>` on `UpstreamClient` with a 60 s TTL to avoid a DB round-trip on every upstream request (mirrors the existing `token_cache` pattern).
2. New method `get_custom_user_agent(repo_id)` with cache-first lookup and DB fallback, using `CUSTOM_USER_AGENT_KEY`.
3. New `apply_custom_ua(req, ua)` helper replaces four repeated `if let Some(ua) = ... { req.header(...) }` patterns.
4. `fetch_buffered` and `fetch_stream` apply the custom UA on the initial request and inside the `exchange_bearer_then` bearer-token retry closure. The buffered/streaming Accept-header asymmetry from #1618 S8 is preserved.

## Test Checklist

- [x] Manually tested locally

## API Changes

- [x] Request/response types have `#[derive(ToSchema)]` (already present; new field picked up automatically)
- [x] Migration is reversible (if applicable); no migration, uses existing `repository_config` table

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -257,6 +257,8 @@ async fn authorize_virtual_member_mutation(
 }
 
 /// Generic upsert helper for repository_config key-value pairs.
+const CUSTOM_USER_AGENT_KEY: &str = "custom_user_agent";
+
 ///
 /// Inserts a new row or updates an existing one for the given repository and
 /// config key. Used by multiple update paths (index_upstream_url, quarantine
@@ -585,9 +587,10 @@ async fn with_custom_user_agent(
     mut response: RepositoryResponse,
 ) -> RepositoryResponse {
     let result = sqlx::query_scalar::<_, String>(
-        "SELECT value FROM repository_config WHERE repository_id = $1 AND key = 'custom_user_agent'",
+        "SELECT value FROM repository_config WHERE repository_id = $1 AND key = $2",
     )
     .bind(repo_id)
+    .bind(CUSTOM_USER_AGENT_KEY)
     .fetch_optional(db)
     .await;
     if let Ok(Some(ua)) = result {
@@ -1464,7 +1467,7 @@ pub async fn create_repository(
 
     if let Some(ref ua) = payload.custom_user_agent {
         if !ua.is_empty() {
-            upsert_repo_config(&state.db, repo.id, "custom_user_agent", ua).await?;
+            upsert_repo_config(&state.db, repo.id, CUSTOM_USER_AGENT_KEY, ua).await?;
         }
     }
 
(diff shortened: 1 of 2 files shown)


brandonrc:
This has drifted out of sync with `main` (CONFLICTING) after the v1.3.0 merge window. Could you rebase onto current `main` and resolve conflicts? Once it's green again we'll pull it into the v1.4.0 milestone. Thanks for the contribution — flagging so it doesn't go stale.

brandonrc:
Thank you for this contribution! We've adopted the custom User-Agent feature — it shipped via #2452 (merged), which builds on your approach with the remote-only validation, CRLF rejection, 60s TTL cache, and full unit + e2e coverage. Closing this in favor of the merged PR. Much appreciated. 🙏

PR #2119, opened 2026-07-01
perf(auth): throttle last_login_at DB writes to reduce WAL pressure

### LINKED ISSUE #2107: perf(auth): throttle last_login_at DB writes to reduce WAL pressure
## Problem

Every successful authentication (password, LDAP, OIDC, SAML, TOTP) unconditionally fires an `UPDATE users SET last_login_at = NOW()` against the database. Package managers re-authenticate on every request, so resolving a large dependency tree can generate dozens of write operations against the `users` table in a burst, producing unnecessary WAL writes even when the value hasn't meaningfully changed. The field is used for display purposes only; minute-level precision is sufficient.

## Proposed solution

Throttle `last_login_at` writes to at most once every 5 minutes per user, using conditional SQL evaluated server-side (works across all pods with no shared in-memory state):

**Standalone paths** (TOTP, LDAP-verify, federated): add a `WHERE` guard:
```sql
UPDATE users SET last_login_at = NOW()
WHERE id = $1
  AND (last_login_at IS NULL OR last_login_at < NOW() - INTERVAL '5 minutes')
```

**Password login** (also resets lockout counters): combine both conditions so the write is skipped only when nothing needs updating:
```sql
UPDATE users
SET last_login_at = CASE
        WHEN last_login_at IS NULL OR last_login_at < NOW() - INTERVAL '5 minutes'
        THEN NOW() ELSE last_login_at
    END,
    failed_login_attempts = 0, locked_until = NULL, last_failed_login_at = NULL
WHERE id = $1
  AND (
    last_login_at IS NULL OR last_login_at < NOW() - INTERVAL '5 minutes'
    OR failed_login_attempts > 0 OR locked_until IS NOT NULL OR last_failed_login_at IS NOT NULL
  )
```

**LDAP/OIDC/SAML profile sync** (also syncs email/display_name/is_admin): skip when profile is unchanged and timestamp is within the 5-minute window, using `IS DISTINCT FROM` for correct NULL handling:
```sql
WHERE id = $4
  AND (
    email IS DISTINCT FROM $1 OR display_name IS DISTINCT FROM $2 OR is_admin IS DISTINCT FROM $3
    OR last_login_at IS NULL OR last_login_at < NOW() - INTERVAL '5 minutes'
  )
```

This approach requires no in-memory state (multi-pod safe), no schema migration, and no new dependencies.

## Affected files

- `backend/src/api/handlers/totp.rs`
- `backend/src/services/auth_service.rs`
- `backend/src/services/ldap_service.rs`
- `backend/src/services/oidc_service.rs`
- `backend/src/services/saml_service.rs`

v4l3nt1tx:
## Summary

Closes #2107.

Package managers re-authenticate on every request (pip, cargo, npm resolve each dependency with credentials). Previously every successful authentication unconditionally fired an `UPDATE users SET last_login_at = NOW()`, producing unnecessary WAL writes even when the value hadn't meaningfully changed. The field is used for display purposes only; minute-level precision is sufficient.

Writes are now throttled to at most once every 5 minutes per user using conditional SQL evaluated server-side. No in-memory state is needed, so the approach is multi-pod safe and works across all instances without coordination.

**`backend/src/services/auth_service.rs`**

Password login path: the write is skipped entirely when `last_login_at` is fresh and no lockout counters need resetting. Lockout counters are still reset immediately when non-zero.

```sql
WHERE id = $1
  AND (
    last_login_at IS NULL OR last_login_at < NOW() - INTERVAL '5 minutes'
    OR failed_login_attempts > 0 OR locked_until IS NOT NULL OR last_failed_login_at IS NOT NULL
  )
```

LDAP-verify and federated paths: simple WHERE guard on `last_login_at`.

**`backend/src/services/ldap_service.rs`, `oidc_service.rs`, `saml_service.rs`**

Profile sync (email, display_name, is_admin): write is skipped when all synced fields are unchanged and `last_login_at` is fresh. `IS DISTINCT FROM` is used for correct NULL-safe comparison.

```sql
WHERE id = $4
  AND (
    email IS DISTINCT FROM $1 OR display_name IS DISTINCT FROM $2 OR is_admin IS DISTINCT FROM $3
    OR last_login_at IS NULL OR last_login_at < NOW() - INTERVAL '5 minutes'
  )
```

**`backend/src/api/handlers/totp.rs`**

Simple WHERE guard on `last_login_at`.

No schema migration needed. `.sqlx` cache updated for the two new CASE WHEN queries.

## Test Checklist

- [ ] Unit tests added/updated — N/A; the throttle logic is pure SQL evaluated server-side, not unit-testable without a real DB
- [ ] Integration tests added/updated (if applicable) — N/A
- [x] Manually tested locally — `cargo clippy` clean, unit tests pass, `SQLX_OFFLINE=true cargo check --workspace --all-targets` clean

## API Changes

- [x] N/A - no API changes


brandonrc:
Closing as superseded — the `last_login_at` write-throttle this PR implements is already on main via #2133 (`perf(auth): throttle last_login_at DB writes to reduce WAL pressure`, merged 2026-07-01, same `Closes #2107`; #2107 is closed). A rebase confirmed the throttle is functionally identical at all sites (totp, password-login, LDAP, and the federated OIDC/SAML sync paths) — net diff ~zero. The only residual difference is a minor SSO-sync `CASE WHEN fresh THEN NOW() ELSE last_login_at END` refinement, not required for correctness; not landing that on the auth hot path without a specific reason. Thanks @v4l3nt1tx — the fix you targeted did ship. (Assessed during v1.3.0 cut prep.)

Issue #2219, opened 2026-07-04
cluster-safe background work claims for multi-replica deployments


cazlo:
## Summary

Artifact Keeper has several background/asynchronous workers that select pending, due, stale, or retryable work and then perform network, storage, email, deletion, or other non-idempotent side effects before recording completion.

In a deployment with multiple backend replicas sharing one Postgres database, each replica can select the same work in the same window and duplicate the side effect.

This is a cluster-safety/concurrency correctness issue rather than a deterministic single-request bug.

## Proposed invariant

Every durable unit of background work should acquire a durable Postgres claim before side effects, perform work only for rows returned by that claim, and finalize success/failure through a random claim token so stale owners cannot clobber a reclaimed row.

## Affected surfaces

I have a fork-side proof of concept on `cazlo/artifact-keeper@feat/cluster-work-claims` that exercises this invariant across the known surfaces:

- `sync_tasks` execution (`RowClaimedQueue`, P0)
- webhook retry delivery (`RowClaimedQueue`, P0)
- backup schedule occurrences (`DueRun`)
- password-expiry email notifications (`RowClaimedQueue`, pre-send claim)
- generic chunked upload completion and expired-session cleanup (`StateMachineLease` / claimed sweep)
- Incus upload finalization and stale-session cleanup (`StateMachineLease`)
- OCI cleanup-journal destructive sweeps (`RowClaimedQueue`)
- singleton scheduler jobs: health monitor, OpenSearch bootstrap reindex, lifecycle, and curation (`SingletonLease`)

## Reproduction / evidence

This is not a deterministic single-request bug. The failure mode depends on two or more backend replicas selecting the same pending/due/stale work item before either replica records completion.

A representative reproduction is to run concurrent workers against the same Postgres database and force both through the claim window for the same work row. In the fork POC, the relevant coverage is expressed as DB-gated regression tests around exactly-once claims, expired-claim recovery, stale-token fencing, and long-lived renewal.

I did not first produce a complete red-test suite against upstream before applying the abstraction, so the tests should be treated as validation of the proposed invariant and regression coverage for split PRs, not as proof that every listed surface is currently reproducible in isolation.

## Fork-side POC

The POC validates that the abstraction holds across the identified worker shapes. Shared `cluster_work` primitives provide the vocabulary:

- `WorkerIdentity`
- `Claimed<T>`
- `SchedulerLease`
- renewal guards
- row-claimed queues
- state-machine leases
- singleton leases

[...]

PR #2221, opened 2026-07-04
fix(sync): claim sync_tasks before peer side effects to stop multi-replica double execution

### LINKED ISSUE #2275: Sync worker: multi-replica double-processing of sync_tasks (no atomic claim before peer side-effects)
In a multi-replica deployment, the sync worker can process the same `sync_tasks` row on more than one replica because tasks are not atomically claimed before the peer-facing side effect (push/delete) fires. This causes duplicate replication work and redundant peer requests.

**Fix (PR #2221):** atomically claim pending tasks (`FOR UPDATE ... SKIP LOCKED` CTE flipping `pending→in_progress` + stamping a `claim_token`, committed before any side effect), token-fence all finalizers, add a TTL claim-lease (`claim_expires_at`) with a recovery sweep for crash-after-claim, and a per-peer advisory-lock oversubscription cap.

Part of the federation-correctness epic #2219.

cazlo:
## Summary

Multi-replica sync workers could execute the same work twice. The worker SELECTed `pending` tasks and only afterwards marked them `in_progress` with an unconditional UPDATE, so with N backend replicas sharing the database the same task could be pushed to (or deleted from) a peer up to N times per tick. The per-peer concurrency check compounded it by reading the shared `active_transfers` counter before other replicas incremented it, letting replicas collectively oversubscribe a peer.

This is the first slice of a broader "claim work in Postgres before external side effects" effort. It introduces the shared primitive and applies it to the highest-risk P0 path:

- **`cluster_work` primitives** (migration 143): `Claimed<T>` (row payload +   `claim_token` + `claimed_by` + `claim_expires_at`), a per-process   `WorkerIdentity`, and a `SchedulerLease`. Side-effecting workers take   `&Claimed<T>`, so an unclaimed call is untypeable.
- **Atomic `sync_tasks` claim** (migration 144): `claim_pending_sync_tasks`   flips `pending -> in_progress` with a `FOR UPDATE SKIP LOCKED` CTE and a fresh   `claim_token`, so two replicas drain disjoint rows instead of   double-executing. The batch is capped in-statement at the peer limit minus
  live claims, and the claim step runs under a per-peer transaction-scoped   advisory lock so that cap is hard even when two replicas' claim statements   overlap exactly. Finalizers are token-fenced   (`AND claim_token = ... AND status = 'in_progress'`); a stale owner skips peer   bookkeeping instead of double-writing. Chunked transfers extend the claim per   uploaded chunk; expired `in_progress` claims are recovered to `pending` each   tick.

Migration 144 also backfills pre-claim `in_progress` rows with expired claim metadata so old stuck rows do not count as live capacity forever. Runtime recovery treats null claim metadata as recoverable too. Sync policy reconciliation now leaves all `in_progress` tasks untouched; when it resets non-in-progress tasks to `pending`, it clears claim metadata.

Migrations are additive: new table, nullable claim columns, partial indexes, and the legacy `in_progress` backfill. Double execution is prevented by the row lock + claim token; the advisory lock additionally closes the oversubscription window. Different peers use different lock keys and never serialize.

Closes #2275.

Part of #2219.

Written as an explicit red-to-green pass. The inherited claim tests were all sequential, so none opened the multi-replica race. This PR adds DB-gated tests that fail on `main` and pass here:

- `claim_pending_sync_tasks_exactly_once_under_concurrency` - one pending task,   two concurrent claimers -> claimed by exactly one, never both.
- `claim_pending_sync_tasks_caps_peer_under_concurrency` - two concurrent   claimers, each seeing all slots free -> total claims never exceed the peer
  limit.
- `expired_sync_task_claims_recover_and_fence_stale_owner` - expired claims  return to `pending`, and stale owners cannot finalize after recovery.
- `legacy_in_progress_sync_task_without_claim_metadata_recovers` - pre-claim   `in_progress` rows with null claim metadata recover to `pending`.
- `sync_policy_reevaluation_preserves_live_claim_and_finalizes` - policy   re-evaluation preserves a live claim and the owner finalizes without leaking  `active_transfers`.

## Test Checklist
- [x] Manually tested locally
see comment about contrived black box experiment used to valdiate the operational behavior in a real-ish environment

## API Changes
- [x] N/A - no API changes


cazlo:
TDD gives me relatively high confidence this works and is tested, but I am running some targetted https://github.com/artifact-keeper/artifact-keeper-test scripts against a local mesh before I open this for review.

cazlo:
Follow-up on the targeted mesh validation: I was able to reproduce the sync-worker race against the public `ghcr.io/artifact-keeper/artifact-keeper-backend:1.2.5` image, and then confirm this branch holds under the same scenario.

I also ran the standard artifact sync mesh smoke from `artifact-keeper-test` against the fixed image in a temporary source/peer topology. After correcting the test-only peer URL allowance for the in-cluster service address, `test-artifact-sync.sh` passed cleanly (`7 passed, 0 failed, 0 skipped`), and the source/peer DB checks showed the replicated artifacts completed with cleared claim fields. That gives me reasonable regression confidence for the ordinary sync path in addition to the claim-specific testing.

The interesting bit from the race validation is how much amplification it took. With 3 source replicas, peer limit `1`, and large artifacts, I could not reproduce the cap violation. Same with 6 replicas. I only got a clean black-box repro after scaling the source backend to 18 replicas and running a larger batch. At that point the old behavior showed both failure modes this PR is meant to prevent: duplicate concurrent dispatch of the same `sync_tasks` row, and `active_transfers` rising above the peer's configured limit.

Running the same amplified scenario against this PR's image held cleanly: live claims / `active_transfers` stayed within the peer limit, and peer logs did not show duplicate upload sessions for the same artifact.

Given how contrived the repro had to be, I do not think the mesh claim-stress script is worth carrying as long-term release-gate coverage. It is sensitive to wall-clock tick alignment, replica count, cluster capacity, and DB sampling. The DB-gated regression tests added in this PR are a better maintenance tradeoff: they force the race deterministically, and they were written red-to-green against the old claim path.

Issue #2062, opened 2026-06-29
CI: Check Rust clippy OOMs (SIGKILL) compiling lib-test on ak-ci-runners


brandonrc:
The `Check Rust` job's `cargo clippy --workspace --all-targets` deterministically OOM-kills (`signal: 9`) while compiling the `lib-test` crate on the 8Gi `ak-ci-runners` pod. This has forced admin-merges of otherwise-green PRs (#2046, #2054, #2055, #2059) and let a compile break slip into main (#2060). History: #1515/#1594/#1679 already dropped debuginfo, disabled incremental, and lowered CARGO_BUILD_JOBS 4→2.

Root cause: `clippy-driver` uses ~2-3x the memory of `rustc` for the same crate. The `Backend Unit Tests` job compiles the same lib-test crate on the same 8Gi runner at CARGO_BUILD_JOBS=4 (rustc) and passes; Check Rust OOMs at jobs=2 because two parallel clippy-drivers exceed 8Gi. Fix: serialize clippy to CARGO_BUILD_JOBS=1 so only a single clippy-driver is resident at peak (a 16Gi `ak-beefy-runners` move is the alternative if a single driver ever exceeds 8Gi).

brandonrc:
Resolved by raising the ak-ci-runners pod memory limit 8Gi->16Gi (iac arc-ci-runners-values.yaml, #178; deployed live helm rev 28). clippy at the existing CARGO_BUILD_JOBS=2 now fits (two drivers ~12Gi < 16Gi) and passes without OOM — validated on a fresh 16Gi runner (Check Rust 3m39s). No more admin-merges needed for this.

Issue #1753, opened 2026-06-08
CI: Code Coverage job fails on main (exit 100, no lcov.info) — gate unenforceable


brandonrc:
## Summary

The **📊 Code Coverage** CI job is failing on `main` and on every PR that touches Rust, blocking nothing required but producing a permanently-red gate.

## Evidence

- `main` latest completed `ci.yml` runs (e.g. run `27157410246`) show **Code Coverage = failure**.
- The job's only annotations are:
  - `No files were found with the provided path: lcov.info. No artifacts will be uploaded.`
  - `Process completed with exit code 100.`
- So the instrumented run (`cargo llvm-cov nextest --workspace --lib --lcov --output-path lcov.info --test-threads 4`, `ci.yml:311`) is dying **before** producing `lcov.info` — the gate never measures anything.
- The in-file comments (`ci.yml:308-310`) already note this crate "OOM-killed check-rust" on the memory-limited ARC runner; exit 100 + no output is consistent with an OOM/crash of the instrumented build, not a coverage-threshold failure.

## Impact

Coverage is **not** a branch-protection-required check (required: `🧪 Backend Unit Tests`, `✅ CI Complete`, `🦀 Check Rust`), so it does not block merges — but it leaves every Rust PR with a red ❌ and makes the real "new-code coverage" gate unenforceable.

## Suggested investigation

- Confirm OOM vs a DB-dependent test failing under the coverage profile (the coverage job has Postgres + runs migrations; the plain unit-test job does not).
- If OOM: lower `--test-threads` (4 → 2), split the instrumented run, or give the coverage job a larger runner.
- If a test fails only under instrumentation/DB: identify and fix it.

Needs runner log access to diagnose precisely (the failed-step logs were not retrievable via `gh run view --log`).

brandonrc:
Reopening: #1754 fixed the `cleanup_stuck_scans_*` flake but the gate is still red on a sibling case — `storage_gc_service::tests::test_run_gc_removes_abandoned_oci_upload_session_storage` fails under nextest for the same reason (the in-process `STORAGE_GC_TEST_LOCK` doesn't serialize across nextest's per-test processes). Broadening the fix.

PR #2230, opened 2026-07-05
test/ci: fix parallel-run test flakes; pool coverage profraw and raise CI test parallelism

### LINKED ISSUE #2229: Test suite flakes at higher --test-threads
## Summary

Three flake families show up when the lib suite runs at higher parallelism, and each has a small fix. I run this repo's coverage suite in my own CI at `--test-threads 8` to keep runtimes down on my hardware; at that concurrency one of these families fired roughly every other run. Across my last 30+ coverage runs none of the three has recurred, and a full coverage pass takes about 11 minutes (10–13) on my main box. Families 1–2 are test-only; family 3 is a small product fix that helps avoid a test flake also (a virtual repo can currently serve a checksum that disagrees with its own metadata body) — happy to split it into its own PR if you'd rather keep this one purely test-infra.

## Family 1: warm-cache tests race the metadata-sidecar commit

`oci_v2::remote_blob_streaming_fallback_tests::remote_blob_streams_large_layer_and_warms_cache` (and the same-shaped npm/helm/swift warm-cache tests) intermittently fail with wiremock's `expected 1 matching request, got 2`.

`ProxyService::tee_stream` commits a streamed write-back in three ordered steps: the content object reaches full size, the storage ETag is pinned, then the `__cache_meta__.json` sidecar is written — and only the sidecar makes the next lookup a cache HIT (by design per #1618: no sidecar ⇒ miss ⇒ refetch ⇒ self-heal). The tests gate their second request on `wait_for_cached_blob`, which polls file SIZE — a condition already true at step one — so under parallel load the second request slips into the window before the sidecar write, misses the cache, and legitimately refetches. The product behavior is correct; the test waits on the wrong authority.

Fix: a `wait_for_cache_commit` test helper that waits for a full-size content object **whose matching sidecar exists** — the same commit marker the production hit path requires — swapped into all four call sites.

## Family 2: PoolTimedOut from 3s test-pool acquire timeouts

The DB-backed suites build small per-suite pools (`max_connections` 3–5) with `acquire_timeout(3s)`. Under parallel load against one Postgres, a query-time acquire occasionally waits past 3s and an unrelated test dies with `PoolTimedOut` — a different victim each time. Fix: raise the six test-only pool builders to 30s, matching the production default (`database_acquire_timeout_secs: 30`). Production pools are untouched.

## Family 3: virtual maven-metadata checksum races its own `<lastUpdated>`

`api::handlers::maven::tests::test_resolver_virtual_merged_metadata_checksum_matches_body_1922` intermittently fails with `assertion left == right failed: virtual .sha1 must equal sha1 of the merged metadata body`.

For a virtual repo, `maven-metadata.xml` is generated on the fly by merging the members' metadata, and the `.sha1`/`.sha256` are served by a **separate** request that regenerates the same body. The merge stamps `<lastUpdated>` with `Utc::now()`, so when the metadata request and the checksum request land in
different wall-clock seconds, the two regenerated bodies differ and the served checksum no longer matches — a real (if rare) correctness issue for Maven clients, not just a test artifact. #1922 already stabilized the local/hosted path (it derives `<lastUpdated>` from `MAX(updated_at)`), but the virtual
merge path was left on the wall clock.

Fix: derive the merged `<lastUpdated>` from the newest `<lastUpdated>` reported by any member (a small `parse_metadata_last_updated` helper), so the body is byte-reproducible across the metadata and checksum requests; fall back to the wall clock only when no member reports one.

## Possible CI benefit

The coverage job currently runs `cargo llvm-cov nextest ... --test-threads 4` and typically takes ~20 minutes; the same suite at 8 threads finishes in about 11 minutes on my hardware once these races are fixed. I know the 4-thread cap is also about ARC runner memory (#1515), so treat a parallelism bump as an optional follow-up — the fixes stand alone, since all three races are load-dependent and can fire at 4 threads too, just less often.

Repro (probabilistic by nature; this is what surfaced it for me):
`cargo llvm-cov nextest --workspace --lib --test-threads 8`, repeated dozens of times overnight.

I have a branch ready with these fixes as three separate commits and will open a PR shortly.

cazlo:
## Summary

This PR grew out of #2229. It bundles the reliability fixes we validated while soak-testing this repo's coverage job in a CI lab that mirrors it command-for-command (same nextest invocation, same Postgres-backed `--lib` suite, same coverage gates), plus CI changes that remove most of the coverage job's disk I/O and raise test/build parallelism. 15 commits: 12 test-only de-flakes, 2 small product fixes with test evidence, 3 CI commits (the last, the unit-job nextest switch, is independently droppable — see below).

**Why fixes and speedups travel together — measured, not asserted:** on an unmodified `main` tree at this repo's own `--test-threads 4`, with only the profraw pooling applied, 17 of 30 coverage runs failed (13 warm-cache wiremock races across helm/npm/swift/oci, 4 token-watermark 401s in the oci upload-authz fixtures, 2 auth-middleware 401s). With the full stack applied, the same infrastructure ran 60/60 green the same evening at 16 test threads — 4x this repo's CI concurrency. Faster tests tighten these race windows, so de-flaking is a prerequisite for the speedups, and this PR's own CI history demonstrates both.

Fixes #2229.

### Measured on this PR's CI (ak-ci-runners, four consecutive green rounds)

| Job | Before (recent main-era runs) | This PR | Change |
|---|---|---|---|
| 📊 Code Coverage | 18–20 min | **8m54s** | ≈ −55% |
| 🧪 Backend Unit Tests | 15m41s | **10m21s** | ≈ −34% |

Phase level, coverage job: compile 12m24s → ~5m30s (`CARGO_BUILD_JOBS` 2→8), test phase 4m51s → ~2m (4→8 nextest threads), profile merge → **2.2s** with the `%32m` pool. All 12,364 lib tests passed in every round.

### The flake families and fixes

| Family | Root cause | Fire evidence (lab, ~290 runs) | Post-fix |
|---|---|---|---|
| warm-cache wiremock "expected 1 request, got 2" (npm/helm/swift/oci) | test asserts the upstream-request count before the streaming tee's cache write commits; the second request is a legitimate cache miss | 5 early sightings at 8T; **13/30 runs at `--test-threads 4` on stock main** | 0 in ~285 runs |
| `PoolTimedOut` in DB-backed tests | default acquire timeout starves under parallel scheduling bursts | 4 sightings | 0 in ~285 runs |
| straggler failure-path tests | tests that *want* an unreachable DB waited out full 30–60s acquire retries (~510 thread-seconds of sleep per run) | wall-clock, not flake | n/a (runtime win) |
| maven virtual `<lastUpdated>` mismatch | virtual metadata merge stamps `<lastUpdated>` with `Utc::now()` per request, so the body and its `.sha1`, fetched ~1s apart, can disagree | 1s-boundary straddle; intermittent historically | 0 in ~195 runs |
| incus finalize race | `completed` status published before temp-file deletion, so a poller can observe terminal status with the temp file still present | 1 fire + latent ordering | 0 in ~120 runs |
| credential-watermark same-second 401s | fixtures mint JWTs in the same second they insert users, so second-granularity `iat` can equal `privileges_changed_at` under the strict comparison | 2 fires at 8T; 5 fires per 30 runs at 4T on stock main; 4/8 on cross-node runners | 0 in ~100+ runs |
| coverage profraw I/O | one `.profraw` per test process × ~12k processes = tens of GB written per run | runner disks pinned at 100% util during coverage | ≤32 files (~92MB), ~20% peak util, 2.2s merge |

**Scope note on the watermark fixes:** the fixture backdates make the tests model reality (credentials are older than the tokens they gate); they deliberately do NOT change the product's strict app-clock/DB-clock comparison. That product-level question — real deployments with a remote Postgres put `iat` and `privileges_changed_at` in different clock domains — is split out as #2245.

**Scope note on the unit-job commit (`ci: run unit tests with nextest at 8 threads`):** the repo's `.config/nextest.toml` `db-serial` groups (#1753) provide the serialization the old `--test-threads=1` run was buying, and the coverage job already executes the same suite under nextest at 8 threads. If maintainers prefer the conservative single-threaded unit lane, that commit reverts cleanly without affecting anything else in the PR.

**Pooling parity:** the pooled profraw output was validated against an unpooled control run on the same tree — identical LCOV aggregates (266 records, 80.09% lines, identical function-hit sets) and identical new-code-gate output. The `CARGO_BUILD_JOBS` bumps are backed by measured peaks of 7.9–11.0 GiB anonymous RSS for the instrumented compile at 8 jobs with `debuginfo=0`, against the 16Gi runner limit from #2062.

  - maven: the existing virtual-metadata/checksum consistency assertions are the regression test — they fail intermittently on `main` (1s boundary) and pass deterministically with the fix.
  - incus: the finalize test polls within an explicit budget and fails on `main` when cleanup lands after terminal status.

## Test Checklist

- [ ] Integration tests added/updated (if applicable) — N/A, `--lib` scope
- [x] Manually tested locally (mirrored CI lab; ledgers above)
- [x] No regressions in existing tests (four consecutive green CI rounds on this PR; 12,364/12,364 passing each round)

## API Changes

- [x] N/A - no API changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -120,11 +120,13 @@ jobs:
     # Same memory guard as check-rust: drop debuginfo and cap parallel rustc
     # so the lib-test compile does not OOM the runner (#1515). This step
     # previously had no CARGO_BUILD_JOBS cap and could spawn one rustc per
-    # core, which is what tipped it over.
+    # core, which is what tipped it over. Jobs raised 4 -> 8 with the same
+    # measured basis as the coverage job (7.9-11.0 GiB peak RSS at 8 jobs
+    # with debuginfo=0 — and this compile is lighter, no instrumentation).
     env:
       CARGO_PROFILE_DEV_DEBUG: "0"
       CARGO_PROFILE_TEST_DEBUG: "0"
-      CARGO_BUILD_JOBS: "4"
+      CARGO_BUILD_JOBS: "8"
       # Disable incremental codegen (#1679): this job compiles the same
       # lib-test crate that OOM-killed check-rust, and incremental is useless
       # in CI's clean checkout while inflating peak RSS.
@@ -215,10 +217,13 @@ jobs:
           SQLX_OFFLINE: true
           DATABASE_URL: postgresql://registry:registry@localhost:5432/artifact_registry
           JWT_SECRET: ci-coverage-secret-at-least-32-bytes-long-for-tests
-        # --test-threads=1: these DB-backed suites share global tables and the
-        # blob-GC readiness gate is a cluster-wide query, so a concurrent
-        # cross-module test's transient rows can make gate/GC assertions flaky.
-        run: cargo test --workspace --lib --verbose -- --test-threads=1
+        # nextest runs each test in its own process and applies the repo's
+        # .config/nextest.toml db-serial test groups, which serialize the
+        # global-DB-state suites (stuck-scan cleanup, storage GC) that the
+        # previous single-threaded `cargo test -- --test-threads=1` run was
+        # protecting. The coverage job already executes this exact suite
+        # under nextest at 8 threads.
+        run: cargo nextest run --workspace --lib --test-threads 8
 
       # Hit-rate data to evaluate dropping the Swatinem rust-cache step once sccache sustains >80% hits.
       - name: sccache stats
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -341,21 +341,42 @@ jobs:
           # explicitly skips when this is set.
           JWT_SECRET: ci-coverage-secret-at-least-32-bytes-long-for-tests
         # nextest runs each test in its own process for faster scheduling.
-        # Cap test concurrency to 4 so parallel test processes don't OOM the
-        # memory-limited ARC runner (cf. the #1515 compile OOM).
-        run: cargo llvm-cov nextest --workspace --lib --lcov --output-path lcov.info --test-threads 4
+        # Test concurrency was capped at 4 to keep parallel test processes
+        # from OOMing the memory-limited ARC runner (cf. the #1515 compile
+        # OOM); raised to 8 here to measure test-phase scaling now that the
+        # per-process profraw I/O is pooled — revert to 4 if runner memory
+        # pressure appears.
+        #
+        # Profile counters are written through an LLVM %Nm merge pool instead
+        # of cargo-llvm-cov's default one-.profraw-per-test-process: with
+        # ~12k test processes the per-process files cost tens of GB of write
+        # I/O per run, while the pool online-merges the same counters into at
+        # most 32 files (~92MB) with byte-equal coverage aggregates. The
+        # explicit `clean --workspace` keeps the implicit-clean semantics the
+        # single `cargo llvm-cov nextest` invocation had.
+        run: |
+          source <(cargo llvm-cov show-env --sh)
+          export LLVM_PROFILE_FILE="${LLVM_PROFILE_FILE%/*}/%32m.profraw"
+          cargo llvm-cov clean --workspace
+          cargo nextest run --workspace --lib --test-threads 8
+          cargo llvm-cov report --lcov --output-path lcov.info
 
       - name: Coverage summary and gate
         if: steps.detect.outputs.rust_changed == 'true'
         run: |
+          # cargo llvm-cov report resolves the profile directory from the
+          # show-env environment; without it, a fresh step shell looks in
+          # the default location and misses the pooled .profraw files.
+          source <(cargo llvm-cov show-env --sh)
+
           # Total project coverage summary
           echo "### Coverage Summary" >> $GITHUB_STEP_SUMMARY
           echo '```' >> $GITHUB_STEP_SUMMARY
-          cargo llvm-cov --workspace --lib --no-run --summary-only 2>&1 | tee -a $GITHUB_STEP_SUMMARY
+          cargo llvm-cov report --summary-only 2>&1 | tee -a $GITHUB_STEP_SUMMARY
           echo '```' >> $GITHUB_STEP_SUMMARY
 
           # Total coverage floor (never let overall coverage drop below 50%)
-          cargo llvm-cov --workspace --lib --no-run --fail-under-lines 50
+          cargo llvm-cov report --fail-under-lines 50
 
       - name: New code coverage gate
         if: github.event_name == 'pull_request' && steps.detect.outputs.rust_changed == 'true'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,13 +251,18 @@ jobs:
     # crate that OOM-killed check-rust, but under `-C instrument-coverage`
     # *and* `-C debuginfo=2` — the heaviest compile in CI, which got SIGKILL'd
     # (OOM) on larger PRs. CARGO_INCREMENTAL=0 drops incremental codegen
-    # (useless in a clean checkout), CARGO_BUILD_JOBS caps parallel rustc, and
-    # dropping debuginfo cuts peak RSS the most. Coverage mapping comes from
-    # the instrumentation (`__llvm_covmap`), not DWARF, so debuginfo=0 does
-    # not affect line-coverage accuracy.
+    # (useless in a clean checkout) and dropping debuginfo cuts peak RSS the
+    # most. Coverage mapping comes from the instrumentation (`__llvm_covmap`),
+    # not DWARF, so debuginfo=0 does not affect line-coverage accuracy.
+    #
+    # CARGO_BUILD_JOBS was capped at 2 when this job ran with debuginfo and
+    # an 8Gi runner limit; both changed since (#2062 raised limits to 16Gi,
+    # debuginfo=0 above). The same instrumented compile at 8 jobs with
+    # debuginfo=0 measures 7.9-11.0 GiB peak anonymous RSS, comfortably
+    # inside 16Gi, and matches the runner's 8-cpu limit.
     env:
       CARGO_INCREMENTAL: "0"
-      CARGO_BUILD_JOBS: "2"
+      CARGO_BUILD_JOBS: "8"
       CARGO_PROFILE_DEV_DEBUG: "0"
       CARGO_PROFILE_TEST_DEBUG: "0"
     services:
--- a/backend/src/api/handlers/conan.rs
+++ b/backend/src/api/handlers/conan.rs
@@ -3433,10 +3433,16 @@ mod tests {
             let username = format!("conan-test-u-{}", id);
             let password = "conan-test-pw".to_string();
             let hash = bcrypt::hash(&password, 4).expect("bcrypt hash failed");
+            // Watermark columns backdated 60s: conan tests login via the
+            // real endpoint, whose token must not race the DB-clock-stamped
+            // creation watermark across node clock skew.
             sqlx::query(
                 r#"
-                INSERT INTO users (id, username, email, password_hash, auth_provider, is_admin, is_active)
-                VALUES ($1, $2, $3, $4, 'local', false, true)
+                INSERT INTO users (id, username, email, password_hash, auth_provider, is_admin, is_active,
+                                   password_changed_at, privileges_changed_at, created_at, updated_at)
+                VALUES ($1, $2, $3, $4, 'local', false, true,
+                        NOW() - INTERVAL '60 seconds', NOW() - INTERVAL '60 seconds',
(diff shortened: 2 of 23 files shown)


cazlo:
@brandonrc, you've been deep in runner tuning lately, so here's raw data from running this repo's coverage job at `--test-threads 8` on my own runners: **0 OOM across 30+ runs.** Mostly numbers below, happy to dig into any of it.

Job: `cargo llvm-cov nextest --workspace --lib --test-threads 8`, measured on the build sidecar's cgroup (cadvisor).

### Memory per phase (30+ runs, 0 OOMKills / 0 restarts, 24 GiB sidecar limit)

| phase | peak RSS (real) | peak working-set |
|---|---|---|
| compile (rustc + llvm-cov instrumentation) | **~8 GB** (worst ~12) | ~9 GB |
| the 8-thread test run | **~0.7 GB** | ~2 GB |
| `llvm-cov` report generation | ~1 GB | **~22 GB** |

- The 8-thread test phase is the **lightest** part of the job. Peak real memory is the compile, which scales with `cargo -j` (roughly the core count), not `--test-threads`.
- That ~22 GB at report-gen is `llvm-cov` reading `target/` + profdata, which is **reclaimable page cache**; RSS stays ~1 GB. `working_set`/`max_usage` sit pinned at the limit through that phase, so they read as OOM-bound when actual demand (RSS) is ~8 GB. On a 16 GiB pool the coverage job won't OOM, but that phase will thrash page cache against the ceiling.

### Throughput / concurrency (one 31-run batch, coverage step only)

| box | runner pods (capacity 1) | 1 job | 3 concurrent | runs/hr |
|---|---|---|---|---|
| **A** (Zen 5) | 3 | 10.3 min | **11.2 min** median | **14.1** |
| **B** (Zen 3) | 1 | 14.0 min | never ran >1 | 4.1 |

- Box A takes 3x concurrent jobs on one host for only ~9% per-job cost (10.3 to 11.2 min). Run-to-run stdev 0.6 min (CV 5%).
- Box B: 14.0 min flat, stdev ~2 s (CV 0.3%). Very consistent, but it only ever ran one at a time.
- I can't separate IPC from memory-bandwidth here (only A ran concurrent). Clean experiment if it's interesting: force B to 3x concurrent. If it blows past 14 min that's bandwidth-bound (DDR4 vs LPDDR5X), if it holds it's compute-bound. Happy to run it.

### Hardware
- **A**: Ryzen AI MAX+ 395 (Zen 5, 16C/32T), 128 GB LPDDR5X.
- **B**: Ryzen 9 5950X (Zen 3, 16C/32T), 64 GB DDR4.

Both 16C/32T, so the 14 vs 11 min gap isn't core count.

### Setup vs your `ak-ci-runners`, in case it's useful

| | your `ak-ci-runners` | mine |
|---|---|---|
| build env | ARC v2 + privileged dind (`docker:27-dind`) | k8s runner + **rootless buildkitd** sidecar |
| cargo runs | directly on the runner image | inside a buildkit build step |
| rust cache | sccache + cargo-home + npm on hostPath `/cache` | BuildKit `--mount=type=cache`: cargo registry **+ `target/`** |
| pod memory | req 8Gi / limit 16Gi, max 6 runners | buildkitd req 8Gi / limit 24Gi, 3 pods/host |

Two things that might be worth a look:
- I cache the whole **`target/` dir** as a BuildKit cache mount, not just sccache. llvm-cov instrumentation shifts object hashes, so a persisted target dir keeps incremental *coverage* builds warm where sccache tends to miss. That's most of how I hold ~11 min.
- Your 20 to 6 runner cap was sized by limit-sum (20 x 16 GiB). My measured coverage RSS is ~8-12 GB against a 24 GiB limit, so sizing by observed RSS rather than limits leaves more packing headroom than the limit math implies (same overcommit caveat you already ran into).

Sharing as data, not a request on the thread count. Your runner sizing may differ.

cazlo:
Did some more experiments with 16 concurrency instead of 8. teased out a bunch more misc test flakes. testing now a series of patches which in ~19 runs so far has not caused a flake and takes runtime down to ~9m -10m on my hardware without materially changing CPU or Memory requirements. Will push these up once I get some more tuning and data collection in place to confirm the most apparent flakes are resolved.

cazlo:
## Measurement appendix

For anyone who likes the data: I validated this PR with ~315 coverage-lane runs (~3M+ executions of the 12.3k-test `--lib` suite against real Postgres) in a private CI lab that mirrors this repo's coverage job command-for-command — 2 nodes, 4 runner slots, dedicated test Postgres — plus the four CI rounds on this PR itself.

```mermaid
timeline
    title How this PR got measured
    section 8-thread era (June)
        runs 101-166 : CI bring-up, first sightings of warm-cache and PoolTimedOut
        runs 167-258 : 92-run soak era : warm-cache + pool-timeout fixes board : final 30-run soak goes 30/30
    section 16-thread day (Jul 5)
        runs 259-411 : 129 runs in one day (~1.58M test executions) : 8 latent flake classes exposed and fixed : per-run failure rate 30% to 0%
        runs 413-415 : profraw %32m pooling validated : LCOV parity proven vs unpooled control
    section A/B/A night (Jul 5)
        runs 416-445 : full fix stack, 16T, no stagger : 30/30 green in 34 min
        runs 446-475 : unmodified main at 4T : 13/30 green
        runs 476-505 : full fix stack again : 30/30 green
    section this PR (Jul 6)
        CI rounds 1-4 : coverage 18-20m to 8m54s : unit tests 15m41s to 10m21s : 12,364 tests green every round
```

### Test-phase wall clock (lab, 16 threads, medians of green runs)

```mermaid
xychart-beta
    title "Coverage test phase by config (seconds)"
    x-axis ["baseline", "+ LPT scheduling", "+ 1s doomed-pool timeout", "+ PG node-pinned", "+ profraw pooling"]
    y-axis "seconds" 0 --> 350
    bar [315, 251, 182, 137, 27.5]
```

With per-process profraw files the test phase was mostly profile I/O — pooling took it from ~2 minutes to 27.5s, and the report merge to 2.2s (measured on this PR's CI).

### Profile I/O

| | per-process | `%32m` pool |
|---|---|---|
| files/run | ~12,300 | ≤ 32 |
| bytes/run | tens of GB (~300GB disk I/O observed at 16T) | ~92 MB |
| runner disk util during coverage | ~100% sustained, alerts firing | ~20% peak, alerts stopped |
| report merge | minutes | 2.2s |

LCOV aggregates stayed byte-equal against an unpooled control (266 records, 80.09% lines, identical function-hit sets; only diff was a known nondeterministic UUID-ordering branch in a conan test).

### The flake A/B (three 30-run no-stagger batches, same night, same infra)

| Tree | Threads | Green |
|---|---|---|
| full fix stack | 16 | 30/30 |
| **unmodified main** + pooling only | **4** (this repo's CI setting) | **13/30** |
| full fix stack | 16 | 30/30 |

What fired on stock main at 4 threads, per 30 runs: warm-cache wiremock family 13 (helm chart-streaming 7, npm tarball 3, swift archive 2, oci blob 1), credential-watermark 401s 5 (oci upload-authz 3, auth middleware 2). Fail-fast truncation makes these lower bounds. Notable: 8 of the 14 flake classes I catalogued never fired once in 92 runs at 8 threads — they only became visible at 16T or under cross-node clock skew, which is how they got fixed before reaching this repo's CI.

Post-fix recurrences, counted across every run since each fix boarded: warm-cache 0/~285, pool-timeout 0/~285, maven 0/~195, incus 0/~120, watermark 0/~100.

### Memory (why jobs=8 / threads=8 are safe)

Measured via Prometheus over ~130 16-thread runs with the same `debuginfo=0` guards this repo uses:

| Phase | Peak |
|---|---|
| instrumented compile, 8 build jobs | 7.9–11.0 GiB anonymous RSS |
| test phase, 8 threads | ~0.7 GiB |
| test phase, 16 threads | barely above 8T (tests are DB-bound, not memory-bound) |
| test Postgres | 167 MiB, 41 connections, 1.2–1.8 cores |

All-in ~12.5 GiB against the 16Gi runner limit from #2062. The old `CARGO_BUILD_JOBS=2` was calibrated for the pre-#2062 8Gi limit with debuginfo enabled — neither condition exists anymore.

### This PR's CI rounds (bars = coverage job, line = unit tests)

```mermaid
xychart-beta
    title "Job duration by round (minutes)"
    x-axis ["baseline", "R1 pooling@4T", "R2 +8 threads", "R3 +jobs=8", "R4 +unit nextest"]
    y-axis "minutes" 0 --> 22
    bar [19, 18.8, 16.9, 9.3, 8.9]
    line [15.7, 15.7, 14, 14.1, 10.4]
```

Round 1 failed at the summary step (env resolution, fixed in round 2); its generate step completed and is the 4-thread phase data. Phase-level on the coverage job: compile 12m24s → ~5m30s, tests 4m51s → ~2m, merge 2.2s. Coverage held at 82.44% region / 80.38% lines throughout.

Caveats: the stock-main batch measured *pooled* 4T behavior, not the unpooled status quo; lab runners have no CPU limit, so lab compile times don't transfer — which is why rounds 1–4 re-measured everything here.

PR #2232, opened 2026-07-05
fix(apt): simplify the InRelease cache invalidation logic

### LINKED ISSUE #2231: APT proxy: cache delete storm on Release/InRelease change
### Description

The APT proxy cache invalidation logic triggers a large batch of storage deletes every time an upstream `Release`/`InRelease` file changes. On a typical Debian distribution, a single Release change produces 50-200+ storage delete operations in one shot.

#### Why the old code deleted cache entries

APT repository metadata has an **integrity constraint**: the `Release` file's checksum sections (`MD5Sum` / `SHA1` / `SHA256` / `SHA512`) declare the expected hash for every sub-file it references. When upstream publishes a new version, `Release` is overwritten in place and its checksums point to brand-new `Packages.gz`, `Translation-en.xz`, etc. If the proxy still serves stale sub-files, `apt-get update` fails with a checksum mismatch.

To prevent that, the old code eagerly deleted every referenced sibling on each Release change:

```mermaid
flowchart TD
    A[Client requests InRelease] --> B[Force-invalidate InRelease cache]
    B --> C[Fetch InRelease from upstream]
    C --> D{SHA-256 changed?}
    D -- yes --> E[Parse Release checksum sections]
    E --> F[Walk every referenced sub-file path]
    F --> G["Delete storm: N storage deletes<br/>(typically 50-200+ paths)"]
    D -- no --> H[Return cached content]
    G --> H
```

Two problems:

1. **Step B — no TTL, no conditional requests.** Every Release/InRelease request force-invalidated the cache and hit upstream, even when content was unchanged. No `If-None-Match` / ETag was ever sent.

2. **Step G — delete storm.** `invalidate_dist_packages_cache` parsed the Release body, walked every path listed under the checksum sections, and issued one `invalidate_cache_by_key` per path. A typical Debian distribution lists dozens to hundreds of paths (every `component × arch` combination produces `Packages`, `Packages.gz`, `Packages.xz`, plus `Sources`, `Translation-*`, `Contents-*`, `dep11/*`, …), all deleted in one shot.

### Environment Information

- Deployment type: N/A (code-level performance issue)
- Operating System: all
- Architecture: all
- Database: PostgreSQL
- Reverse Proxy: N/A
- Network setup: N/A

### Backend Version

1.2.5

### Web UI Version

N/A

### Package Format

Debian / APT

### Expected Behavior

A Release/InRelease change should not trigger a batch delete of all referenced sibling cache entries. The proxy should use TTL + conditional requests (`If-None-Match` / ETag) to avoid unnecessary upstream fetches, and invalidation of sibling caches should be lazy (per-entry, on next read) rather than eager.

### Observed Behavior

Every Release/InRelease request force-invalidated the cache and hit upstream (no TTL, no conditional requests). When content changed, the old `invalidate_dist_packages_cache` method parsed the Release body and issued one `invalidate_cache_by_key` per referenced path — 50-200+ storage deletes in a single request.

#### Impact

- **Delete storm**: a single upstream Release change triggers N storage deletes (N = number of paths referenced in the Release file, typically 50-200+).
- **No TTL protection**: every request hit upstream even when content was unchanged.
- **No conditional requests**: upstream bandwidth wasted on full body fetches that a 304 would have avoided.
- **Storage I/O spikes**: in high-frequency mirroring scenarios (e.g. CI environments running `apt-get update` constantly), the delete storm saturated storage backend IOPS.

### Steps to Reproduce

1. Configure a remote (proxy) APT repository pointing to an upstream Debian mirror.
2. Trigger `apt-get update` from a client to populate the proxy cache (Release, InRelease, Packages*, etc.).
3. Wait for the upstream mirror to publish a new Release (or simulate by changing the upstream content).
4. Run `apt-get update` again — the proxy detects the Release change and fires N storage deletes for all sibling cache entries.

### Logs

```shell
# No specific log line — the issue is observable via storage backend metrics:
# a burst of delete operations on the proxy-cache bucket/prefix immediately
# following an InRelease/Release fetch that detects a content change.
```

#### Proposed approach: epoch-based lazy invalidation

Introduce a `__release_epoch__.json` sidecar per distribution that acts as an "expiry group" marker:

```
proxy-cache/{repo_key}/dists/{distribution}/__release_epoch__.json
```

Content is a single RFC 3339 timestamp:

```json
{"content_changed_at":"2026-07-02T09:16:18Z"}
```

The new flow:

```mermaid
flowchart TD
    A[Client requests InRelease] --> B[Load cache metadata + epoch]
    B --> C{Epoch expired?<br/>cached_at < content_changed_at}
    C -- yes --> D[Mark as expired]
    C -- no --> E{TTL expired?}
    E -- yes --> D
    E -- no --> F[Cache hit, return content]
    D --> G[Send conditional GET with If-None-Match]
    G --> H{Upstream response?}
    H -- "304 Not Modified" --> I[Update metadata, return cached]
    H -- "200 new body" --> J{SHA-256 changed?}
    J -- yes --> K[Write new cache + update epoch timestamp]
    J -- no --> L[Update metadata only]
    K --> M[Return new content]
    L --> M
    I --> M
```

**Key difference**: when Release content changes, **no sibling caches are deleted eagerly**. Only the epoch timestamp is updated. Each sibling checks the epoch lazily on its next read (`maybe_invalidate_by_epoch`) and deletes only its own stale entry if expired.

Immutable paths (`by-hash/` and `pool/`) skip the epoch check entirely via `cache_classifier::classify(...).is_immutable()` — their content is content-addressed, so a Release change cannot affect them.

The old `invalidate_dist_packages_cache` method and its Release-file parser `parse_release_file_paths` would be removed along with their tests; the eager batch-deletion path no longer exists.

#### Open questions

1. **Epoch file consistency** — `__release_epoch__.json` is written before the Release cache entry (`content_changed_at <= cached_at`, strict `<` comparison prevents self-invalidation). In a crash between writing the epoch and writing the cache body, the epoch may point to a version not yet on disk — siblings would be re-fetched unnecessarily. Acceptable degradation, but worth noting.

2. **Storage backend eventual consistency** — on S3-class backends, the epoch file may be briefly invisible after write (`read_release_epoch` returns `None` → no expiry check). Cold start is also `None`, which is a safe default.

3. **Cross-replica coordination** — the epoch lives in the filesystem alongside cache entries. If future work needs cross-replica epoch coordination (similar to the advisory-lock approach in #2169), it may need to migrate to a database table.

A PR implementing this approach will be opened shortly. If maintainers prefer a different direction, the issue and PR can be closed without concern — this is one possible solution, not a settled design.

#### Related

- #1147 — virtual APT repo support
- #1236 — signed Release cache
- `__cache_meta__.json` sidecar — cache metadata storage

### Pre-submission Checklist

Dreamacro:
## Summary

Replaces the APT proxy cache invalidation logic from eager batch deletion to
epoch-based lazy invalidation.

**The problem**: every time an upstream `Release`/`InRelease` file changed, the
old `invalidate_dist_packages_cache` parsed the Release checksum sections and
called `invalidate_cache_by_key` for every referenced path (each `component ×
arch` pair produces `Packages`, `Packages.gz`, `Packages.xz`, plus `Sources`,
`Translation-*`, `Contents-*`, `dep11/*`, … — typically 50-200+ paths per
distribution). This created a delete storm on every upstream change.

**The fix**: introduce a `__release_epoch__.json` sidecar per distribution
storing a single timestamp. When Release content changes, only the epoch is
updated — no sibling caches are deleted eagerly. Each sibling checks the epoch
lazily on its next read (`maybe_invalidate_by_epoch`) and deletes only its own
stale entry if expired. The Release/InRelease fetch now uses TTL +
conditional requests (`If-None-Match` / ETag) instead of force-invalidating
and always hitting upstream.

```mermaid
flowchart TD
    A[Client requests InRelease] --> B[Load cache metadata + epoch]
    B --> C{Epoch expired?}
    C -- yes --> D[Expired]
    C -- no --> E{TTL expired?}
    E -- yes --> D
    E -- no --> F[Cache hit]
    D --> G[Conditional GET with If-None-Match]
    G --> H{304?}
    H -- yes --> I[Revalidate metadata]
    H -- no --> J{SHA changed?}
    J -- yes --> K[Write cache + update epoch]
    J -- no --> I
```

Immutable paths (`by-hash/` and `pool/`) skip the epoch check entirely via
`cache_classifier::classify(...).is_immutable()` — their content is
content-addressed, so a Release change cannot affect them.

The old `invalidate_dist_packages_cache` method and the
`parse_release_file_paths` parser have been removed along with their tests.

New tests added:
- `test_write_and_read_release_epoch_round_trip` — epoch serialization round-trip
- `test_write_release_epoch_with_traversal_distribution_is_noop` — path-traversal guard
- `test_is_dists_epoch_expired_*` — epoch expiry comparison (expired, fresh, boundary, no epoch)
- `classify_debian` classifier tests (`by-hash` immutable, `pool/` immutable, `dists/` mutable)

Closes #2231.

### DIFF
--- a/backend/src/api/handlers/debian.rs
+++ b/backend/src/api/handlers/debian.rs
@@ -38,10 +38,12 @@ use crate::api::handlers::proxy_helpers::{self, RepoInfo};
 use crate::api::middleware::auth::{require_auth_basic_scope, AuthExtension};
 use crate::api::{SharedState, SIGNED_RELEASE_CACHE_MAX_ENTRIES};
 use crate::formats::debian::{DebControl, DebianHandler};
-use crate::models::repository::RepositoryType;
+use crate::models::repository::{RepositoryFormat, RepositoryType};
 use crate::models::signing_key::SigningKey;
 use crate::services::artifact_service::ArtifactService;
+use crate::services::cache_classifier;
 use crate::services::package_service::PackageService;
+use crate::services::proxy_service::{ProxyService, DEFAULT_DISTS_INDEX_TTL_SECS};
 use crate::services::signing_service::SigningService;
 
 const DEBIAN_BINARY_CONTENT_TYPE: &str = "application/vnd.debian.binary-package";
@@ -538,6 +540,7 @@ impl<'a> DebianProxy<'a> {
                 self.state,
                 repo.id,
                 self.repo_key,
+                self.distribution,
                 &upstream_path,
                 content_type,
             )
@@ -556,6 +559,12 @@ impl<'a> DebianProxy<'a> {
             _ => return Ok(()),
         };
         let upstream_path = format!("dists/{}/{}", self.distribution, suffix);
+
+        // Epoch-based lazy invalidation: if the cached file is older
+        // than the release epoch, invalidate it so the streaming fetch
+        // treats it as a cache miss and re-fetches from upstream.
+        maybe_invalidate_by_epoch(proxy, self.repo_key, self.distribution, &upstream_path).await;
+
         let (content, upstream_ct) = proxy_helpers::proxy_fetch_capped(
             proxy,
             repo.id,
@@ -568,14 +577,13 @@ impl<'a> DebianProxy<'a> {
         Err(build_dists_response(content, upstream_ct, content_type))
     }
 
-    /// Variant of `dists` that also detects whether the upstream content
-    /// changed since the last cached copy. When it has, sibling Packages
-    /// caches under the same distribution are invalidated so subsequent
-    /// `apt-get update` requests refetch them and the Release SHA-256
-    /// list matches what apt sees (#1147).
+    /// Variant of `dists` that uses TTL + conditional-request +
+    /// epoch-based lazy invalidation for Release/InRelease files.
     ///
-    /// Used by the Release / InRelease handlers, where stale Packages
-    /// caches manifest as `Hash Sum mismatch` errors on the client.
+    /// Sibling files compare their own `cached_at` against the release
+    /// epoch timestamp to decide freshness at read time.
+    ///
+    /// Used by the Release / InRelease handlers.
     async fn dists_detecting_change(
         &self,
         suffix: &str,
@@ -584,10 +592,7 @@ impl<'a> DebianProxy<'a> {
     ) -> Result<(), Response> {
         let upstream_path = format!("dists/{}/{}", self.distribution, suffix);
 
-        // Virtual: iterate Remote members. Whichever member serves the
-        // Release also owns the sibling Packages caches we may need to
-        // invalidate, so we run the change-detection probe against that
-        // specific member before returning.
+        // Virtual: iterate Remote members.
         if repo.repo_type == RepositoryType::Virtual {
             if let Some(resp) = try_virtual_dists_detecting_change(
                 self.state,
@@ -614,14 +619,16 @@ impl<'a> DebianProxy<'a> {
 
         let pseudo_repo = proxy_helpers::build_remote_repo(repo.id, self.repo_key, upstream_url);
         let (content, upstream_ct, changed) = proxy
-            .fetch_dists_detecting_change(&pseudo_repo, &upstream_path)
+            .fetch_dists_with_revalidation(
+                &pseudo_repo,
+                &upstream_path,
+                self.distribution,
+                DEFAULT_DISTS_INDEX_TTL_SECS,
+            )
             .await
             .map_err(map_proxy_err)?;
 
-        if let Some(text) = release_invalidation_payload(changed, &content) {
-            proxy
-                .invalidate_dist_packages_cache(self.repo_key, self.distribution, text)
-                .await;
+        if changed {
             // Drop any signed-Release entries for this dist; the next
             // InRelease / Release.gpg fetch will re-sign against the new
             // content (#1236).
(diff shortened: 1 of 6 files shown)


brandonrc:
Reviewed this for v1.4.0 inclusion — it's a genuinely nice win (killing the 50–200-path delete storm per upstream Release change), the single-replica happy path is correct, and the test coverage on the new epoch/lazy/conditional paths is strong (path-traversal on the sidecar is guarded and tested; the 401-Bearer SSRF guard is preserved on the conditional retry; `pool/` and `by-hash/` correctly stay immutable). Two items to resolve before it can land in a multi-replica release:

**🔴 F1 — cross-replica epoch race can reintroduce `Hash Sum mismatch` (#1147).** `write_release_epoch` does a blind `put` of `ReleaseEpoch { content_changed_at: Utc::now() }` with no max-merge, and `read_release_epoch` maps a parse/visibility failure to `None` → `is_dists_epoch_expired` treats `None` as *not expired*. AK runs multiple proxy replicas on shared storage, so: replica A bumps the epoch for a new `InRelease` (T2); the client's next `Packages` request lands on replica B before the sidecar is visible, or B's wall-clock lags → B sees "fresh" → serves the **old `Packages` against the new `InRelease`**. Suggested fix: (a) make the epoch monotonic — read existing, write `max(existing, now)`; (b) fail *safe* — treat a read/parse failure as *expired* rather than *not expired* (or confirm the storage backend is strongly read-after-write consistent for this key). A concurrent/multi-replica test for this path would be worth adding.

**🟠 F2 — this also changes the hosted-repo overwrite-protection oracle, which is outside the stated scope.** The new `Debian` arm in `is_explicitly_mutable_index` (cache_classifier.rs) is also consumed by the soft-delete/overwrite guard at `artifact_service.rs:545` — so hosted Debian `dists/…` index coordinates now skip the tombstone/overwrite protection and can be silently overwritten/resurrected. This may well be intended (dists indexes are legitimately rewritten in place, much like `maven-metadata.xml`), but since the PR is scoped as *proxy cache invalidation* could you confirm the semantic change is deliberate and add a hosted-Debian overwrite/tombstone test asserting the new behavior? `pool/` + `by-hash/` staying protected is correct.

Minor (non-blocking): lazy invalidation leaves never-re-read sibling entries on disk until their own TTL/lifecycle cleanup (self-heals on read, so not a stale-serve bug); the 304 path trusts the upstream ETag with no body re-check (mitigated by apt's GPG verification, and no worse than the prior force-refetch design); and there's a duplicated doc-comment line on `read_upstream_response_capped`.

Milestoning as a **v1.4.0 candidate** — happy to help land F1 (it's a small targeted change). F2 just needs your confirmation on intent + the test.

PR #2256, opened 2026-07-06
feat(npm): cross-replica single-flight for SWR background refreshes

### LINKED ISSUE #2248: npm packument cache: cross-replica single-flight for SWR background refreshes
## Problem Statement

Follow-up to #2162 / #2166. The stale-while-revalidate background refresh deduplicates within a single process (an in-process claim set guarantees at most one in-flight refresh per cache key per pod). With the Redis backend on a multi-replica deployment, a traffic burst on the same stale packument spread across N replicas can trigger up to N concurrent upstream refreshes for the same key. Each is bounded and best-effort, so this is waste rather than incorrectness, but on deployments with many replicas it multiplies upstream load for zero benefit, and the first refresh to complete wins anyway.

## Proposed Solution

Extend the shared-backend path with a cross-replica refresh lease: before spawning a background refresh, attempt `SET refresh-lease:{cache_key} <replica-id> NX EX <ttl>` (ttl a little above the refresh timeout); only the lease holder refreshes, others skip. On completion the holder deletes the lease. Failure modes degrade gracefully: if Redis is unavailable the existing per-process dedup still applies, and an expired lease simply allows the next attempt.

## Alternatives Considered

- Doing nothing: correctness is unaffected; this is purely an upstream-load optimisation, so it may be acceptable for small replica counts.
- Reusing the existing Postgres advisory-lock single-flight (as on the miss path): works, but the refresh path is best-effort and fire-and-forget, so a cheap Redis NX lease fits better than a DB round-trip when the shared backend is already configured.

## Use Case

Multi-replica deployments using the Redis backend under bursty install traffic (CI fan-out hitting the same packuments as their fresh windows lapse).

rob-howie-depop:
## Summary

Closes #2248. Follow-up to #2166.

The stale-while-revalidate background refresh dedups within a single process only (the in-process claim set). With the Redis backend on a multi-replica deployment, a traffic burst on the same stale packument spread across N replicas triggers up to N concurrent upstream refreshes for the same key — bounded and best-effort, so pure waste rather than incorrectness, but it multiplies upstream load for zero benefit.

This gates the background refresh behind a **cross-replica lease** on the shared backend:

- On a stale hit, the spawned refresh task first attempts `SET ak:npm-packument:refresh-lease:{flight_key} {token} NX EX 90` — only the winner refreshes; everyone else skips, and the winner's entry reaches all replicas through the shared cache. (The issue sketched acquiring *before* spawning; acquiring *inside* the spawned task has the same dedup effect with zero added latency on the stale-serving request path.)
- **Release is conditional on a per-acquisition token** (GET-then-DEL; deliberately not Lua `EVAL`, which scripting-restricted Redis deployments refuse — the sub-millisecond non-atomic window's worst case is one duplicate refresh), so a holder that outlived its TTL can never free a successor replica's lease. A release lost to a Redis blip is retried once after the cooldown window, so a transient outage cannot orphan the lease until TTL expiry.
- **Lease TTL (90 s) can never be outlived**: the leased compute is bounded at 80 s, so a slow refresh is abandoned rather than allowed to finish after a successor and overwrite its newer entry. A crashed holder delays the next cross-replica refresh until expiry, never blocks it permanently.
- **A denied replica parks its local claim briefly** before re-arming, so a stale burst on a non-holder costs one shared-backend probe per window, not one spawned task per request.
- **Failure posture matches the rest of the shared backend**: Redis unreachable or in cooldown degrades to a local lease — per-process dedup still applies and refreshes are never lost to a cache outage. The in-process backend (no shared state) is unaffected: the default trait impl grants a local lease unconditionally.

The lease/claim split: the existing `RefreshClaim` dedups a stale burst within one process (no Redis round-trip for followers); the lease dedups across replicas (one `SET NX` per spawned refresh attempt, off the request path).

## Test Checklist
- [x] Unit tests added/updated (lease acquire/skip/degrade/release-on-failure semantics, claim always freed, two-facades-one-shared-store race collapses to exactly one refresh, in-process backend unaffected, denial parks the claim, over-long computes abandoned with the lease released, failed release retried once)
- [x] Integration tests added/updated (env-gated real-Redis test via `NPM_PACKUMENT_CACHE_TEST_REDIS_URL`: NX denial while held, PTTL bound, wrong-token release is a no-op, release/reacquire cycle with unique tokens; DB-backed handler test: a stale hit with the lease held elsewhere serves the cached body and provably issues **no** upstream refresh request)
- [ ] E2E tests added/updated (if applicable) — N/A, no client-visible protocol change
- [x] Manually tested locally (`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --lib` against Postgres 16 + a live Redis 7)
- [x] No regressions in existing tests (the only failures in a full parallel run are pre-existing large-object-streaming/timeout flakes that reproduce identically on `main` and pass in isolation on both)

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/services/npm_packument_cache.rs
+++ b/backend/src/services/npm_packument_cache.rs
@@ -86,6 +86,22 @@ const REDIS_KEY_NAMESPACE: &str = "ak:npm-packument:";
 /// (sibling of the proxy path's `proxy-cache:` / `proxy-stream:` prefixes).
 const FLIGHT_LEASE_NAMESPACE: &str = "npm-packument:";
 
+/// Key prefix (under [`REDIS_KEY_NAMESPACE`]) for cross-replica background
+/// refresh leases (#2248).
+const REFRESH_LEASE_KEY_PREFIX: &str = "refresh-lease:";
+
+/// TTL on the cross-replica refresh lease (#2248). A live holder completes
+/// (and releases) well inside this: a refresh's compute is bounded by the
+/// same limits as the miss path, whose single-flight wait deadline is 65 s.
+/// A holder that dies without releasing only delays the next cross-replica
+/// refresh until expiry — it can never block refreshes permanently.
+const REFRESH_LEASE_TTL: Duration = Duration::from_secs(90);
+
+/// Compare-and-delete for lease release: frees the lease only while `token`
+/// still owns it, so a holder that outlived its TTL cannot drop a successor
+/// replica's lease.
+const REFRESH_LEASE_RELEASE_SCRIPT: &str = r#"if redis.call("GET", KEYS[1]) == ARGV[1] then return redis.call("DEL", KEYS[1]) else return 0 end"#;
+
 /// Bound on each Redis connection attempt, so a request never stalls behind
 /// an unreachable Redis host (the in-process layer answers instead).
 const REDIS_CONNECT_TIMEOUT: Duration = Duration::from_secs(1);
@@ -243,6 +259,20 @@ fn key_invalidation_prefix(key: &str) -> String {
 // Backend traits
 // ---------------------------------------------------------------------------
 
+/// Cross-replica claim on one background refresh (#2248). Complements the
+/// process-local [`RefreshClaim`]: the claim dedups a stale burst within one
+/// process, this lease dedups the same burst across replicas that share a
+/// cache backend.
+#[derive(Debug)]
+pub enum RefreshLease {
+    /// No shared lease is held: either no shared backend is configured, or it
+    /// was unreachable and refresh dedup degraded to per-process only.
+    Local,
+    /// Held in the shared backend; the unique `token` proves ownership, so a
+    /// release can never drop a lease acquired later by another replica.
+    Shared { flight_key: String, token: String },
+}
+
 /// Storage backend for computed packument responses.
 ///
 /// Implementations must degrade gracefully: a backend problem surfaces as a
@@ -258,6 +288,25 @@ pub trait PackumentCacheBackend: Send + Sync {
     async fn set(&self, key: &str, prefix: &str, entry: CachedPackument);
     /// Drop every entry whose key starts with `prefix`.
     async fn invalidate_prefix(&self, prefix: &str);
+    /// Try to acquire the cross-replica background-refresh lease for
+    /// `flight_key` (#2248). `None` means another replica is already
+    /// refreshing this packument and the caller must skip. Backends without
+    /// shared state grant a [`RefreshLease::Local`]: the per-process claim
+    /// set is the only dedup needed there.
+    async fn try_acquire_refresh_lease(
+        &self,
+        flight_key: &str,
+        ttl: Duration,
+    ) -> Option<RefreshLease> {
+        let _ = (flight_key, ttl);
+        Some(RefreshLease::Local)
+    }
+    /// Release a lease returned by [`Self::try_acquire_refresh_lease`]. An
+    /// unreleased lease (holder crash, task cancellation) expires on its own
+    /// after the TTL.
+    async fn release_refresh_lease(&self, lease: RefreshLease) {
+        let _ = lease;
+    }
 }
 
 /// The shared cache is unreachable or misbehaving; the caller should use the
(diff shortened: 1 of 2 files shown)


backend/src/services/npm_packument_cache.rs
```diff
@@ -86,6 +86,37 @@ const REDIS_KEY_NAMESPACE: &str = "ak:npm-packument:";
...
+/// under this TTL by a margin. A holder that dies without releasing (panic,
+/// runtime teardown) only delays the next cross-replica refresh until
+/// expiry — it can never block refreshes permanently.
+const REFRESH_LEASE_TTL: Duration = Duration::from_secs(90);
+
+/// Bound on a leased background refresh. Keeping the compute strictly under
+/// [`REFRESH_LEASE_TTL`] guarantees a lease can never expire mid-refresh, so
+/// a slow holder cannot race a successor and overwrite its newer entry with
+/// an older upstream snapshot.
+const REFRESH_LEASE_COMPUTE_TIMEOUT: Duration = Duration::from_secs(80);
```
rob-howie-depop, line 105: **Review finding (resolved):** the original 90s lease TTL assumed a refresh always completes inside it, but the upstream fetch was only per-read bounded, so a slow-drip download could outlive the lease and let an older compute overwrite a successor's newer entry. Resolved by bounding the leased compute at 80s (`REFRESH_LEASE_COMPUTE_TIMEOUT`), making the invariant true by construction; covered by `refresh_under_lease_abandons_computes_that_outlive_the_lease_bound`.

backend/src/services/npm_packument_cache.rs
```diff
@@ -657,6 +753,57 @@ impl SharedCacheBackend for RedisPackumentCache {
...
+        Ok(acquired.map(|_| token))
+    }
+
+    async fn try_release_refresh_lease(
+        &self,
+        flight_key: &str,
+        token: &str,
+    ) -> Result<(), SharedCacheUnavailable> {
+        let mut conn = self.connection().await?;
+        // GET-then-DEL rather than a Lua compare-and-delete: EVAL is refused
```
rob-howie-depop, line 785: **Review finding (resolved):** the first cut released via Lua `EVAL`, which scripting-restricted Redis deployments (ACLs without `@scripting`) refuse — every refresh cycle would fail its release and arm the cooldown gate. Resolved by switching to GET-then-DEL: the non-atomic window is sub-millisecond and its worst case is one duplicate refresh, strictly better than systemic degradation.

backend/src/services/npm_packument_cache.rs
```diff
@@ -703,6 +850,44 @@ impl PackumentCacheBackend for LayeredPackumentCache {
...
+        if let RefreshLease::Shared { flight_key, token } = lease {
+            if self
+                .shared
+                .try_release_refresh_lease(&flight_key, &token)
+                .await
+                .is_ok()
+            {
+                return;
+            }
+            // One retry after the cooldown window: a release lost to a Redis
```
rob-howie-depop, line 881: **Review finding (resolved):** a release swallowed by a transient Redis blip orphaned the lease, blocking every replica's refresh of that packument until TTL expiry. Resolved with a single bounded retry after the cooldown window (6s > the 5s gate, so the retry actually reaches Redis); covered by `refresh_under_lease_retries_a_failed_release_once`.

backend/src/services/npm_packument_cache.rs
```diff
@@ -891,6 +1084,60 @@ impl NpmPackumentCache {
...
+    /// The refresh is bounded by [`REFRESH_LEASE_COMPUTE_TIMEOUT`], so a
+    /// holder can never outlive its lease and race a successor's newer
+    /// entry. A failed refresh still releases the lease, so any replica
+    /// (including a healthy one, when the holder's upstream path is broken)
+    /// can retry on the next stale hit. Only a crash mid-refresh leaves the
+    /// lease to TTL expiry — bounded, and the stale window keeps serving.
+    ///
+    /// Returns `None` when skipped or timed out, `Some(refresh result)`
+    /// otherwise.
+    pub async fn refresh_under_lease<T, E, F, Fut>(
```
rob-howie-depop, line 1104: **Review findings (resolved):** (1) on lease denial the local claim was dropped immediately, so a stale burst on a non-holder replica spawned one task + one Redis probe per request for the whole duration of the holder's refresh — the claim is now parked for a short hold window (`refresh_under_lease_denial_parks_the_claim_before_rearming`); (2) the API took a redundant `flight_key` parameter that could diverge from the claim's — the lease is now keyed off the claim itself, making mismatch unrepresentable.

rob-howie-depop (commented):
Adversarial self-review record (high-effort, independent multi-agent finder/verifier pass over the diff). Nine findings; each inline comment below documents one and how it was resolved in f51a4d9d. Findings accepted-as-designed: (a) a holder crash leaks the lease until TTL expiry — TTL expiry is the designed recovery per the issue text, and the compute-timeout bound makes panics the only leak path; (b) a broken-upstream replica winning the lease delays repair — mitigated because a failed refresh releases the lease immediately and denied replicas re-probe within the claim-hold window.

Issue #2162, opened 2026-07-02
npm proxy: cache computed packuments with stale-while-revalidate to remove the per-request upstream round-trip


rob-howie-depop:
## Problem Statement

On npm proxy repositories, every packument request (`GET /npm/{repo}/{package}`) re-fetches or revalidates the metadata from the upstream registry and re-runs the full transform pipeline (parse, rewrite `dist.tarball` URLs, abbreviate for `install-v1+json` clients, serialize, compress). Nothing caches the computed result, and responses carry no cache headers. The upstream round-trip is paid on essentially every request.

For install-sized workloads this dominates end-to-end latency. A dependency re-resolve of a typical mid-size app (1,301 resolved packages, 1,101 unique packuments) issues ~1,100 packument requests, and each one blocks on upstream even when the same packument was served seconds earlier.

#1931 fixed the payload size (abbreviated documents + compression), but measurements show bytes were never the latency bottleneck: install speed did not move.

### Evidence

All numbers below are from a self-hosted deployment in AWS us-east-1 proxying registry.npmjs.org, medium stack (1,301 packages), fresh client cache per run, warm server, 10 runs per cell.

**A/B of four server builds, same instance, `npm install` re-resolve (medians):**

| build | npm re-resolve | pnpm re-resolve |
|---|---|---|
| 1.2.1-equivalent (pre-#1931: full docs, no compression) | 100.8 s | 46.0 s |
| 1.2.3 (abbreviated + brotli) | 96.8 s | 29.6 s |
| 1.2.3 + gzip-only control (brotli off) | 103.5 s | 28.3 s |
| **1.2.3 + computed-packument cache (prototype)** | **69.5 s** | **22.4 s** |

Compression choice (none vs brotli vs gzip) does not move install time; only caching the computed response does. That is the empirical proof the cost is the upstream round-trip, not compression compute.

**Where a single request spends its time** (median over 13 representative packuments, fresh connection, WAN client ~98 ms from the server):

| request | server wait | total |
|---|---|---|
| registry.npmjs.org (reference) | 34 ms | 67 ms |
| proxy, current behaviour | 281 ms | 774 ms |
| proxy + cache, miss (blocking refresh) | 542 ms | 941 ms |
| proxy + cache, warm hit | 195 ms | 594 ms |

[...]

Issue #2093, opened 2026-07-01
Private-repo image scanning: minimal scanner service account + short-lived repo-scoped tokens (Harbor-style)


brandonrc:
## Summary

Enable **private-repo container-image scanning** for both scanners (grype + the new Trivy adapter, #2091/#2090) using a **Harbor-style least-privilege model**: a minimal, non-login, read-only **scanner service account** as a DB-resident backing identity, from which the backend mints **short-lived, repo-scoped tokens per scan** and hands them to the scanner over the registry-pull path. This closes the last 1.2.2 parity gap without a long-lived pull-all credential.

## Background
- Scanners pull target images from AK's own registry. grype authenticates via `~/.docker/config.json` (grype_scanner.rs:214); 1.2.2's bundled trivy did the same. The Trivy adapter (#2090) accepts a per-request token via `registry.authorization` but ships with `ImageScanner::with_token_minter` **unwired** → anonymous pulls → private repos fail-closed.
- AK constraint: `validate_access_token_async` re-derives identity from a DB row, so a synthetic token is rejected — a DB-resident backing identity is required.

## Design goals
1. **Minimal scanner service account**: non-login, read-only, pull scope; DB-resident (migration + seed). Never handed to users; audit-logged.
2. **Short-lived, repo-scoped tokens**: mint per scan, scoped to the target repo only, short TTL — the credential that leaves the backend is ephemeral + narrow (Harbor model), NOT the account's standing token.
3. **Wire `ImageScanner::with_token_minter`** so the adapter gets `registry.authorization: Bearer <scoped-jwt>`.
4. **Unify grype + trivy on one identity/token path** — grype should use the same scanner identity (mint an equivalent credential / docker-config) rather than a separate model.
5. First **confirm how grype's credential is actually provisioned today** (backend + iac) — whether a real scanner cred exists or private-repo scanning has been unconfigured.

## Acceptance
- Private-repo image scan → both grype and the Trivy adapter return real findings (authenticated pull succeeds) with a short-lived repo-scoped token.
- No long-lived pull-all credential is introduced; token TTL + scope verified.
- Public/local repos still scan; unreachable/unauthorized still fail-closed (never false-clean).

## Related
- #2091 (adapter, PR #2092), #2088/PR #2090 (adapter client), artifact-keeper-test#237 (gate), iac#185/#186 (deploy).

PR #2258, opened 2026-07-06
feat(npm): upstream feed subscriptions for proactive metadata invalidation

### LINKED ISSUE #2249: proxy repositories: subscribe to upstream event feeds for proactive metadata invalidation (npm replication feed first)
## Problem Statement

Follow-up to #2162 / #2166. The packument cache bounds staleness with a fresh window plus stale-while-revalidate, which is the correct floor for any proxy: upstreams do not tell proxies when packages change. But some upstreams do publish change feeds - npm exposes a public CouchDB-style replication feed (`replicate.npmjs.com/_changes`) streaming package-change events. Nothing consumes it today, so a freshly published version is only picked up when a request triggers revalidation.

## Proposed Solution

A small, generic subscription framework for remote repositories rather than an npm one-off:

- **Feed adapters**: per-ecosystem drivers that turn an upstream feed into a normalised stream of `(package, change)` events. First implementation: npm's `_changes` feed (long-poll/continuous mode, resume from the last sequence number).
- **Pluggable actions** on each event, configured per remote repo. Default action: invalidate the cached computed metadata for that package (for npm, the packument cache entries). An optional eager action could recompute/warm instead of just invalidating.
- **Operational shape**: best-effort by design - reconnect with backoff, tolerate feed re-shapes, and never treat the feed as the correctness mechanism (TTL/SWR remains the floor). On multi-replica deployments a single consumer should hold the subscription (the existing advisory-lock single-flight pattern fits) so N replicas do not open N feed connections.

Any registry ecosystem whose upstream later offers a feed then only needs an adapter, not a new invalidation design.

## Alternatives Considered

- Bespoke npm-only consumer: less code today, but the invalidation plumbing is format-agnostic and other ecosystems have or may grow feeds.
- Polling hot packages' dist-tags: simpler but scales poorly and adds steady upstream load for mostly-unchanged data.
- Doing nothing: acceptable - SWR already bounds staleness to one stale-served request past the fresh window - which is why this is an enhancement, not a fix.

## Use Case

Deployments proxying registry.npmjs.org that want near-real-time visibility of new upstream releases (publish-then-consume CI races, security-patch bumps) while keeping bounded-staleness semantics as the guarantee.

rob-howie-depop:
## Summary

Closes #2249. Follow-up to #2162/#2166.

The packument cache bounds staleness with a fresh window plus stale-while-revalidate — the correct floor when upstreams say nothing about changes. npm's upstream does say: `replicate.npmjs.com/_changes` streams package-change events. This adds a small, generic subscription framework (`services/upstream_feed.rs`) rather than an npm one-off, and consumes nothing by default (opt-in):

- **Feed adapters** — `UpstreamFeedAdapter` turns one upstream feed into a normalised stream of `(package, cursor)` batches. First implementation: `NpmReplicationFeedAdapter`, long-polling `_changes` (`feed=longpoll`, bounded `timeout`/`limit`, resume via `since`). Parsing is tolerant by construction: unknown fields ignored, malformed rows skipped, numeric and string sequence cursors both accepted, response-level `last_seq` with a last-row fallback — a feed re-shape degrades to skipped rows or a missing cursor, never a panic.
- **Pluggable actions** — `FeedAction` per polled batch. Default `PackumentInvalidationAction`: drop the cached computed packuments for the changed package in every remote npm repository whose upstream the feed actually covers (host-scoped to `registry.npmjs.org`/`registry.npmjs.com` — private registries are out of the feed's scope and are never touched) and in every virtual repo containing one. The fan-out is a helper shared with the existing local-write invalidation path (`invalidate_package_and_virtuals`), so "a package changed" means the same thing from both directions.
- **Operational shape** — best-effort by design, exactly as the issue frames it:
  - **Single consumer cluster-wide**: leadership via the existing `ClusterLock` advisory-lock primitive (new lock class `0x2249`); non-leaders re-check on an interval, so consumption fails over when the leader dies. Leadership runs in **bounded terms** (5 min): the leader steps down, releases the lock and re-contends, so a lock silently lost to a dead connection (e.g. an RDS failover killing the detached lock session) converges back to a single consumer within one term instead of consuming in duplicate until restart. A lock-infrastructure error means *no* consumer (never N of them), and the TTL/SWR floor covers the gap.
  - **First enablement starts at the feed head** (`since=now` via the adapter's head-cursor sentinel), never at seq 0 — enabling the flag on an existing deployment must not replay npm's entire multi-million-event change history through the invalidation path.
  - **Resume**: cursor persisted per feed in `upstream_feed_state` (migration 156); a failed save loses only the resume position (idempotent replay), never stops consumption.
  - **Resilience**: capped exponential backoff on poll failures; idle instant-empty responses are paced; a batch that returns events *without* cursor progress (a feed re-shape whose seqs cannot be normalised) escalates into the same backoff so the consumer can never hot-loop replaying one batch; clean shutdown via the server's cancellation token. The adapter uses the shared `http_client::base_client_builder()` (custom-CA bundle + SSRF redirect policy).
  - The feed is never the correctness mechanism: TTL/SWR remains the staleness floor.

Two adjacent hardenings surfaced by review: actions are **batch-shaped** (`FeedAction::apply(&[FeedEvent])`), so the default action resolves the repository set once per batch instead of once per event; and the packument cache's invalidation-generation map now carries a **map-wide epoch** bumped on its cap-clear sweep, so a store guard captured before the sweep can never match the swept map's default generation and re-install data an invalidation dropped — routine once feed-driven invalidations cycle the map at npm churn rates, and covered by a regression test.

Any ecosystem whose upstream later grows a feed needs only an adapter, not a new invalidation design.

### Configuration

| Env var | Default | Meaning |
|---|---|---|
| `NPM_UPSTREAM_FEED_ENABLED` | `false` | Opt-in flag; also requires the packument cache to be enabled |
| `NPM_UPSTREAM_FEED_URL` | `https://replicate.npmjs.com/_changes` | Feed endpoint |

Cross-replica note (documented in the module): with the shared (Redis) packument-cache backend an invalidation propagates to every replica; with the in-process backend it clears the leader's replica and the others fall back to the TTL floor.

## Test Checklist
- [x] Unit tests added/updated (adapter: long-poll param construction, since omitted on first poll, head-cursor bootstrap, batch parsing, malformed-row/feed-reshape tolerance, HTTP/parse failures, invalid URL; consumer: events applied in order + cursor persisted + resume, first-run polls from the head sentinel, recovery after poll errors, stalled-cursor batches back off instead of hot-looping, leader steps down at term end and re-contends, save-failure keeps consuming, non-leader never polls, takeover after leader release, lock-infra error never self-elects, prompt cancellation; cache: cap-clear epoch regression test; pure helpers: backoff, seq normalisation, feed host scoping)
- [x] Integration tests added/updated (DB-gated: `upstream_feed_state` round-trip; end-to-end invalidation scoping — covered remote repo + its virtual evicted, other-upstream repo and unrelated packages survive)
- [ ] E2E tests added/updated (if applicable) — N/A, no client-visible protocol change; the consumer is dark by default
- [x] Manually tested locally (`cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --lib` against Postgres 16 with migration 156 applied)
- [x] No regressions in existing tests (the only failures in a full parallel run are pre-existing large-object-streaming/timeout flakes that reproduce identically on `main` and pass in isolation on both)

## API Changes
- [x] N/A — no API changes
- [x] Migration is reversible (if applicable) — migration 156 is a single additive `CREATE TABLE IF NOT EXISTS upstream_feed_state`; reversal is `DROP TABLE upstream_feed_state` (noted in the migration header). No HTTP API changes.


rob-howie-depop (commented):
Adversarial self-review record (high-effort, independent multi-agent finder/verifier pass over the diff). Eight findings; the four confirmed correctness defects and two hardenings are documented inline below, all resolved in d4e2d4a7 with regression tests. Also resolved from the same review: the adapter now uses the shared `http_client::base_client_builder()` (custom-CA bundle + SSRF redirect policy) instead of a bare `reqwest::Client`, and the two `tests/` fixtures reference `NPM_REPLICATION_FEED_DEFAULT_URL` instead of a string literal. One finding was rejected: removing issue-number anchors from comments — they are this codebase's established convention (`#2162`, `#1608`, `#2093`, ...), so consistency won.

backend/src/services/upstream_feed.rs
```diff
@@ -0,0 +1,1544 @@
...
+
+#[async_trait]
+impl UpstreamFeedAdapter for NpmReplicationFeedAdapter {
+    fn feed_key(&self) -> &str {
+        &self.feed_key
+    }
+
+    /// CouchDB understands `since=now` on long-poll feeds: start at the
+    /// current head instead of replaying the registry's entire history.
+    fn initial_cursor(&self) -> Option<String> {
```
rob-howie-depop, line 267: **Review finding (resolved):** the first cut polled with no `since` on first enablement, which CouchDB treats as seq 0 — replaying npm's entire multi-million-event change history through the invalidation path for days. Resolved with the adapter-owned head sentinel (`initial_cursor()` = `now`); the consumer bootstraps from it whenever no cursor is persisted. Covered by `consumer_first_run_polls_from_the_adapters_head_cursor` and `npm_adapter_bootstraps_at_the_feed_head`.

backend/src/services/upstream_feed.rs
```diff
@@ -0,0 +1,1544 @@
...
+const LEADER_RETRY_INTERVAL: Duration = Duration::from_secs(30);
+
+/// Upper bound on one uninterrupted leadership stint: the leader steps
+/// down, releases the lock and re-contends. A Postgres session lock is
+/// freed server-side when its (detached) connection dies, and the holder
+/// has no way to notice — without a bounded term that silent loss would
+/// leave the old leader consuming alongside its successor until process
+/// restart. The term bounds any such overlap; actions are idempotent, so
+/// the overlap is harmless while it lasts.
+const LEADER_TERM: Duration = Duration::from_secs(300);
```
rob-howie-depop, line 89: **Review finding (resolved):** the leader had no way to notice a silently lost advisory lock (Postgres frees a session lock when its detached connection dies), so a follower's takeover produced permanent dual consumption. Resolved with bounded leadership terms: the leader steps down, releases and re-contends, so any overlap converges within one term. The leader is deliberately sticky (immediate re-contention) for availability. Covered by `consumer_steps_down_at_term_end_and_recontends`.

backend/src/services/upstream_feed.rs
```diff
@@ -0,0 +1,1544 @@
...
+                        backoff = self.backoff_initial;
+                    } else if batch.events.is_empty() {
+                        // Idle long-poll round (a well-behaved feed already
+                        // blocked server-side): gentle fixed pacing so an
+                        // instant-empty feed cannot hot-loop.
+                        if cancelled_within(&self.cancel, self.empty_batch_pacing).await {
+                            return;
+                        }
+                    } else {
+                        // Events but no cursor progress: the next poll would
```
rob-howie-depop, line 620: **Review finding (resolved):** a batch carrying events whose seqs cannot be normalised (feed re-shape) advanced nothing, and the old empty-batch-only pacing let the consumer hot-loop replaying the identical batch at maximum rate. Resolved by escalating events-without-cursor-progress into the poll-failure backoff; idle empty rounds keep gentle fixed pacing. Covered by `consumer_backs_off_when_events_arrive_without_cursor_progress`.

backend/src/services/upstream_feed.rs
```diff
@@ -0,0 +1,1544 @@
...
+
+impl PackumentInvalidationAction {
+    pub fn new(db: PgPool, cache: Arc<NpmPackumentCache>) -> Self {
+        Self { db, cache }
+    }
+}
+
+#[async_trait]
+impl FeedAction for PackumentInvalidationAction {
+    async fn apply(&self, events: &[FeedEvent]) {
```
rob-howie-depop: **Review finding (resolved):** the action ran the full repositories scan once per event (N+1 against the shared Postgres, painful for multi-event batches). Resolved by making `FeedAction` batch-shaped: the repository set (and each covered repo's virtual keys) is resolved once per batch, and the per-event work is pure cache invalidation.

backend/src/services/npm_packument_cache.rs
```diff
@@ -858,18 +876,20 @@ impl NpmPackumentCache {
...
+                && !state.generations.contains_key(&prefix)
             {
-                // Clearing is safe-conservative: any in-flight guard sees a
-                // generation change and skips its store.
-                generations.clear();
+                // Clearing is safe-conservative ONLY together with the epoch
+                // bump: a swept package would otherwise report generation 0
+                // again and in-flight guards captured at 0 would pass.
+                state.generations.clear();
+                state.epoch += 1;
```
rob-howie-depop, line 1083: **Review finding (resolved):** the invalidation-generation map's cap-clear silently defeated `store_guarded` for guards captured at the default generation — a swept package reported generation 0 again, letting a pre-invalidation compute re-install exactly the data an invalidation dropped. Latent before this PR, routine once feed-driven invalidations cycle the map at npm churn rates. Resolved with a map-wide epoch bumped on every sweep, so pre-sweep guards can never match. Covered by `store_guarded_skips_after_generation_map_cap_clear`.

brandonrc:
### Maintainer adoption note

Adopting this contribution from @rob-howie-depop — thanks for the thorough, well-tested design (generic adapter/action framework, single-consumer leadership with bounded terms, head-cursor bootstrap, idempotent replay). All five original commits are preserved with their authorship; this update only rebases onto current `main` and adds the two items below.

**1. Migration renumber 149 → 156.** `149` is taken by `upload_chunks_claimed_at` on `main`, which is also past `147/148` (`scheduler_leases` / `sync_tasks_claim`). Renumbered to `156` — the next free number after `155_projects` — to keep the ascending-merge ordering. In-code references updated; table/schema unchanged (`CREATE TABLE IF NOT EXISTS upstream_feed_state`, reversible via `DROP TABLE`).

**2. Entry-point SSRF guard on the configured feed URL.** The adapter already builds its client with `http_client::base_client_builder()`, so the shared **redirect-hop** SSRF policy (`is_blocked_url` on every hop) and the custom-CA bundle apply — good. But `base_client_builder()` only re-validates *redirects*; it does not validate the *initial* request target. Every other outbound path in the tree (cargo/composer/remote-instances) runs the configured URL through `validate_outbound_url` before issuing, and this one did not — so a `NPM_UPSTREAM_FEED_URL` pointing straight at a private / link-local / cloud-metadata address (e.g. `http://169.254.169.254/…`, an internal service name, RFC1918) would have been fetched directly. `NpmReplicationFeedAdapter::new` now calls `validate_outbound_url(url, "npm upstream feed URL")`; a blocked URL cleanly declines to start the consumer (logged) instead of reaching an internal target, and the TTL/SWR floor still bounds staleness. A `from_url_unchecked` test seam keeps the loopback-mock poll tests working, and `npm_adapter_rejects_ssrf_feed_urls` covers metadata / loopback / IPv4-mapped / internal-hostname inputs.

Note: DNS-rebinding (public at validation, private at fetch) remains the pre-existing, separately-tracked residual gap documented in `validation.rs` for *all* outbound paths — not specific to this PR.

**Validation (isolated pool slot + throwaway PG, `:8080` untouched):**
- `cargo fmt --check`, `cargo clippy --lib --tests -- -D warnings` — clean.
- Unit: 26 `upstream_feed` + 47 `npm_packument_cache` lib tests pass (incl. the new SSRF test and the cap-clear epoch regression test).
- DB-gated integration against the real migrated schema (migration 156 applied by the startup runner): `pg_feed_state_store_round_trips` and `invalidation_action_scopes_to_covered_repos_and_their_virtuals` both pass.
- Live: patched image deployed clean (migration 156 applies in sequence, consumer opt-in-off by default). With the feed enabled + `NPM_UPSTREAM_FEED_URL=http://169.254.169.254/…` the guard blocked it (`outbound URL blocked … reason="ip"`), the consumer did not start, and the backend stayed healthy. With the default `replicate.npmjs.com/_changes` the consumer acquired leadership normally.

Heads-up unrelated to this PR: the `streaming_invariant` source-scan gate currently fails on a clean `main` checkout (`npm.rs` 5-vs-6, `pypi.rs` 1-vs-0 allowlist drift, untouched by this PR) — a pre-existing allowlist reconciliation needed on `main`.

Issue #2247, opened 2026-07-06
Conan remote repository: conan download fails with no conanfile.py although upstream recipe contains it


MarioMiosgaCNH:
### Description

I am using Artifact Keeper as a Conan v2 remote/pull-through repository backed by an authenticated upstream Conan registry.
With Artifact Keeper 1.3.0, package listing and recipe/package metadata resolution appear to work, but conan download still fails when downloading the recipe.
The upstream registry contains export/conanfile.py for the resolved recipe revision, but Conan reports that the recipe downloaded through Artifact Keeper is corrupted because conanfile.py is missing.

### Environment Information

- Deployment type: Docker, deployed via Ansible
- Operating System: Debian 11 host
- Architecture: x86_64
- Database: PostgreSQL
- Reverse Proxy: Traefik
- Network setup: Traefik reverse proxy -> Artifact Keeper Docker containers

### Backend Version

1.3.0

### Web UI Version

1.3.0

### Package Format

Conan (C/C++)

### Expected Behavior

If the upstream recipe revision contains export/conanfile.py, Artifact Keeper should expose it correctly through the Conan v2 remote API.
conan download should successfully download the recipe, including conanfile.py.

### Observed Behavior

`conan list` succeeds and shows the package reference:

```text
<package>
  <package>/<version>@<user>/<channel>
```

`conan download` starts downloading the resolved recipe revision:

```text
Connecting to remote 'artifact-keeper' with user '<user>'
Downloading recipe '<package>/<version>@<user>/<channel>#<rrev>'
```

Then it fails:

```text
ERROR: Error downloading from remote 'artifact-keeper'
ERROR: Corrupted <package>/<version>@<user>/<channel> in 'artifact-keeper' remote: no conanfile.py
```

Backend logs show the request flow reaches the recipe files endpoint successfully:

```text
GET /conan/<repo>/v1/ping
-> 200

GET /conan/<repo>/v2/conans/<name>/<version>/<user>/<channel>/latest
-> 200

GET /conan/<repo>/v2/conans/<name>/<version>/<user>/<channel>/revisions/<rrev>/search
-> 200

GET /conan/<repo>/v2/conans/<name>/<version>/<user>/<channel>/revisions/<rrev>/packages/<package-id>/latest
-> 200

GET /conan/<repo>/v2/conans/<name>/<version>/<user>/<channel>/revisions/<rrev>/files
-> 200
```

[...]

PR #2303, opened 2026-07-08
perf(proxy): cache __cache_meta__.json sidecars in-process for #2301

### LINKED ISSUE #2301: Proxy cache hits on warm cache pay 300ms+ per request due to repeated __cache_meta__.json NAS roundtrips
## Problem

A `Remote` (proxy) repository is supposed to make consecutive Maven builds dramatically faster than the first build by caching upstream responses. In practice, warm-cache requests pay ~300ms each — almost as much as cold requests — so the cold→warm speedup is only ~1.14× instead of the 5-10× users expect.

Measured on a live cluster with the 1.2.3 release image, 50 sequential `maven-metadata.xml` requests against `orig-maven-repo`:

| Run | Cache state | Total | Per-request |
|---|---|---|---|
| 1 | Cold | 17.7s | 354ms |
| 2 | Warm (immediately after) | 15.5s | 309ms |
| 3 | Warm, 4-way parallel | 8.1s | 162ms |

A working cache should make Run 2 ~3-5s total.

## Root cause

Each cache-hit request does 2-3 NAS roundtrips inside `proxy_service::read_cached_with_revalidation_streaming` (`backend/src/services/proxy_service.rs:2438`):

1. Read `__cache_meta__.json` (freshness sidecar)
2. Read `__content__` (cached body bytes)
3. Possibly `fsync` if the metadata TTL was just extended (write-back)

The storage backend in the affected deployment is `STORAGE_BACKEND=filesystem` mounted on Alibaba Cloud NAS (`alibabacloud-cnfs-nas`). Each NAS op is 50-100ms. So 3 × 75ms ≈ 225ms per cache hit, plus framing overhead.

`fetch_artifact_streaming` does cache-miss path more cheaply because it doesn't need to read the sidecar first — it just goes upstream.

For reference, Nexus uses local SSDs / in-memory caches for the same operation → ~1-5ms per cache hit.

## Proposed fix

Add an in-process LRU (moka, already used elsewhere in the codebase) keyed by `(repo_key, cache_path)` that holds the parsed `CacheMetadata` struct. TTL ~60s, size cap ~10k entries (~5MB).

Pseudocode for the hot path:

```rust
// pseudo
let cache_key = (repo.key.clone(), cache_path.to_string());
let metadata = METADATA_SIDECAR_CACHE
    .try_get_with(cache_key, async {
        self.load_cache_metadata(metadata_key).await
    })
    .await?;
let entry = metadata.as_ref().map(|m| m.as_cache_entry(mutability));
// rest of evaluate → Fresh/Stale/...
```

On the write path (`CachePersister::write_metadata`), populate the LRU on success so the next read hits in-memory.

## Affected path

- `backend/src/services/proxy_service.rs::try_streaming_cache_hit` (line 2394)
- `backend/src/services/proxy_service.rs::read_cached_with_revalidation_streaming` (line 2438)
- `backend/src/services/proxy_service.rs::load_cache_metadata` (used by both probe + revalidate)

## Expected impact

- Per cache-hit request: 300ms → 150-200ms (NAS metadata read skipped)
- For a 1000-request Maven build: warm path goes from ~300s to ~150-200s
- Storage backend (NAS) roundtrips per request: 3 → 2

## Out of scope

- Replacing the storage backend (NAS → SSD / S3) — infra change
- In-memory caching of `__content__` bytes — that's much larger; risks RAM pressure. Leave on NAS.
- Buffered `fetch_artifact` path — same fix applies but is a follow-up.

kmtong:
Closes #2301

## Summary

Wraps `ProxyService::load_cache_metadata` in a process-global `moka` LRU keyed by the full storage key. Capacity 10k entries (~5 MB), TTL 30s. Every read goes through the LRU; every write (write_buffered, tee_stream, extend_cache_expiry, write_negative_cache) invalidates the entry.

## Why

Each cache-hit request on a remote (proxy) repo paid ~75 ms per NAS roundtrip just to read the freshness sidecar, and three of those per request summed to ~300 ms — making the warm cache barely faster than cold. Measured on a live 1.2.3 cluster:

| Run | Cache state | 50 sequential `maven-metadata.xml` requests |
|---|---|---|
| 1 | Cold | 17.7 s |
| 2 | Warm (right after) | 15.5 s |

A 1.14× cold→warm speedup. Users expect 5-10×. The 1.14× is the symptom; the root cause is the sidecar roundtrip on every request, fixed here.

Tests added in `backend/src/services/proxy_service.rs`:
- `test_proxy_metadata_lru_caches_positive_entry_across_calls`
- `test_proxy_metadata_lru_caches_definitive_miss` (verifies `Some(None)` semantics for known-absent keys)
- `test_proxy_metadata_lru_invalidate_drops_entry`
- `test_load_cache_metadata_uses_lru_after_first_load` (end-to-end: writes via `cache_artifact`, deletes the on-disk sidecar, asserts the second read still returns the cached metadata from the LRU)

## Test Checklist

- [x] Manually tested locally

`cargo test -p artifact-keeper-backend --lib -- services::proxy_service` → 274 passed, 0 failed (4 new + 270 existing).
`cargo clippy -p artifact-keeper-backend --lib --tests -- -D warnings` → clean.
`cargo fmt --check` → clean.

## API Changes

- [x] N/A - no API changes

## Backward compatibility

- Pure addition: existing call sites of `load_cache_metadata` are unchanged.
- LRU miss always falls through to the existing `cache_store.load_metadata` path; the behaviour on a cold LRU entry is identical to before this PR.
- B6 (storage error → cache miss) preserved: any LRU `try_get_with` error is logged and degraded to `Ok(None)`.

## Out of scope

- HTTP/2 to upstream, gzip/brotli, lower `connect_timeout` — separate follow-up.
- Caching `__content__` bytes in memory — would risk RAM pressure for large JARs.
- Cross-replica cache invalidation — single-process moka is sufficient for this fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -221,6 +221,7 @@ dependencies = [
  "metrics-exporter-prometheus",
  "mimalloc",
  "mime_guess",
+ "moka",
  "native-tls",
  "object_store",
  "once_cell",
@@ -307,6 +308,17 @@ dependencies = [
  "compression-core",
  "pin-project-lite",
  "tokio",
+]
+
+[[package]]
+name = "async-lock"
+version = "3.4.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "290f7f2596bd5b78a9fec8088ccd89180d7f9f55b94b0576823bbbdc72ee8311"
+dependencies = [
+ "event-listener",
+ "event-listener-strategy",
+ "pin-project-lite",
 ]
 
 [[package]]
@@ -2206,6 +2218,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "event-listener-strategy"
+version = "0.5.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8be9f3dfaaffdae2972880079a491a1a8bb7cbed0b8dd7a347f668b4150a3b93"
+dependencies = [
+ "event-listener",
+ "pin-project-lite",
+]
+
+[[package]]
 name = "evmap"
 version = "11.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3826,6 +3848,26 @@ dependencies = [
  "ctutils",
  "hybrid-array",
  "num-traits",
+]
+
+[[package]]
+name = "moka"
+version = "0.12.15"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "957228ad12042ee839f93c8f257b62b4c0ab5eaae1d4fa60de53b27c9d7c5046"
+dependencies = [
+ "async-lock",
+ "crossbeam-channel",
+ "crossbeam-epoch",
+ "crossbeam-utils",
+ "equivalent",
+ "event-listener",
+ "futures-util",
+ "parking_lot",
+ "portable-atomic",
+ "smallvec",
+ "tagptr",
+ "uuid",
 ]
 
 [[package]]
@@ -6269,6 +6311,12 @@ dependencies = [
  "windows-sys 0.52.0",
  "winx",
 ]
+
+[[package]]
+name = "tagptr"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b2093cf4c8eb1e67749a6762251bc9cd836b6fc171623bd0a9d324d37af2417"
 
 [[package]]
 name = "tar"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -134,3 +134,6 @@ bergshamra = "=0.5.1"
 
 # SMTP email delivery
 lettre = { version = "0.11", features = ["tokio1-native-tls", "builder"] }
+
+# In-memory cache with TTL eviction (#2079, #2301).
+moka = { version = "0.12", features = ["future"] }
--- a/backend/Cargo.toml
+++ b/backend/Cargo.toml
@@ -105,6 +105,7 @@ clap.workspace = true
 async-trait = "0.1"
 async-stream = "0.3"
 hex = "0.4"
+moka.workspace = true
 rand = "0.9"
 rand08 = { package = "rand", version = "0.8" }
 pgp = "0.14.2"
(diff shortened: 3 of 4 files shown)


Dreamacro:
Caching `__cache_meta__.json` will cause some problems: for example, APT has the concept of expiration groups, i.e., when an index is expired, the related files should also be expired, so caching `__cache_meta__.json` or `__content__` can't be based on the expiration time of the meta directly. So the cache, whether caching `__cache_meta__.json` or `__content__`, should not be based on the expiration time in the meta.

To realize the correct caching logic is more complicated, currently there is the concept of Mutability, for Mutability::Immutable to do this cache is safer, but the current HEAD has a bug, and can not correctly classify the path, I'm tracking down this problem.

rob-howie-depop:
+1 to the diagnosis — this is the same root-cause shape we hit on the npm side (#2247/#2248): per-request metadata round-trips dominating the warm path. The fix pattern (in-process cache for parsed metadata) is right, but there's prior art in this codebase worth borrowing from, because the naive version of this LRU re-creates two issues the npm packument cache already had to solve (`backend/src/services/npm_packument_cache.rs`):

1. **Invalidation vs. LRU races.** With a 60s sidecar LRU, a purge/eviction (`purge_repo_cache`, single-entry invalidation, repo deletion) leaves the in-memory entry serving "fresh" for up to the TTL unless those paths also invalidate the LRU. And "populate the LRU on `write_metadata` success" races a concurrent invalidation: write completes → invalidation lands → LRU populate re-installs the stale entry. The packument cache handles exactly this with per-key generation counters + a post-store re-check (see `StoreGuard` / `invalidate_package`); the sidecar LRU needs the same guard or purges become unreliable in a way that's very hard to debug from the outside.

2. **Multi-replica drift.** An in-process LRU doesn't observe another replica's writes or purges — bounded by the TTL, but worth stating in the docs/config surface (for the packument cache we created a shared-Redis backend for precisely this). For a freshness sidecar 60s of cross-replica staleness is probably acceptable; for purge semantics it may not be.

brandonrc:
Superseded by #2450 (merged) — rebased onto main (dropped the redundant moka dep re-add, already a workspace dep) + validated. Thanks @kmtong!

Issue #1094, opened 2026-05-08
Map OIDC groups to groups


hboyd2003:
### Problem Statement

Currently, OIDC groups are mapped to Artifact Keeper roles. These mappings are hard coded and lack any fine grain control making permission management difficult.

### Proposed Solution

Simply map groups directly with groups created in Artifact Keeper. This would also remove the need for the "Admin Group" override as you can simply configure the group as admin via the permissions.

For backwards compatibility a admin group can be automatically created that matches the set "Admin Group" override.

### Use Case

As a Sysadmin I want fine grained control of permissions that map well into other existing permission structures/systems.

### Component

Authentication / Authorization

### Pre-submission Checklist

- [x] I have searched existing issues and feature requests to ensure this isn't a duplicate
- [x] I have reviewed the documentation to confirm this feature doesn't already exist

PR #2332, opened 2026-07-09
feat(sso): map SAML assertion groups to local group memberships

### LINKED ISSUE #2333: SAML: map assertion groups to group memberships (parity with OIDC #1094)
## Problem

SAML SSO maps assertion groups only to `is_admin` (via the provider's
`admin_group`). There is no way to grant **non-admin** permissions from SAML
group membership.

OIDC already supports this via `map_groups_to_groups` (#1094): the OIDC `groups`
claim is reflected into Artifact Keeper group memberships, with groups
auto-created and reconciled per login. SAML has no equivalent.

This matters because some IdPs (e.g. Google Workspace) emit group membership
**only over SAML**, not in the OIDC id_token. On those IdPs, SAML users can
authenticate but cannot receive any group-based (non-admin) permissions at all —
an admin must assign every user to groups by hand.

## Proposal

Add a per-provider `map_groups_to_groups` flag to SAML configs, mirroring the
OIDC behavior: when enabled, reflect the SAML assertion group values (via the
`groups` attribute mapping) as Artifact Keeper group memberships, auto-creating
groups tagged `external_source = 'saml'` and reconciling on each login. Reuse
the existing group-sync reconcile, scoped by `(external_source, provider)` so
SAML and OIDC never strip each other's memberships.

AkulovNV:
Closes #2333

## What

Adds a per-provider `map_groups_to_groups` flag to SAML configs (migration 149),
mirroring the OIDC group-to-group mapping from #1094. When enabled, group values
carried in the SAML assertion (`groups` attribute mapping) are reflected as
Artifact Keeper group memberships: groups are auto-created tagged
`external_source = 'saml'` and reconciled on every login. When disabled (default),
only the existing `admin_group -> is_admin` behavior applies.

## Why

SAML currently maps assertion groups only to `is_admin` via `admin_group`; there
is no way to grant non-admin permissions from group membership. OIDC already
supports this (#1094), but IdPs like Google Workspace only emit group membership
over SAML, not in the OIDC id_token — so SAML users could not receive
group-based permissions at all.

## How

- The OIDC group-sync core is generalized into `sync_federated_groups_to_local_groups`
  taking a `source` tag; `sync_oidc_groups_to_local_groups` becomes a thin wrapper,
  so all existing OIDC callers/tests are untouched.
- `saml_acs` invokes it with `source = "saml"` when the provider has
  `map_groups_to_groups` enabled. Sync failure is logged and never blocks login.
- The reconcile DELETE is scoped by `(external_source, external_provider_id)`, so
  SAML and OIDC providers never strip each other's memberships, and operator-managed
  groups (NULL `external_source`) are never touched.

## Tests

Adds `test_sync_saml_tags_source_and_isolates_from_oidc`: verifies SAML groups are
tagged `'saml'` and that an OIDC reconcile with the same provider id does not strip
SAML-sourced memberships. Existing OIDC sync tests unchanged.

## Also fixed (unrelated, CI-blocking)

`clippy` 1.97 (released 2026-07-07) promoted `useless_borrows_in_formatting`,
which now fires on pre-existing code in `backend/src/api/handlers/repositories.rs`
(two `assert!` format args and one `format!()` call). This fails
`cargo clippy --workspace --all-targets -- -D warnings` for **any** PR against the
repo, not just this one. Removed the three redundant `&` references so CI can go
green. This is orthogonal to the SAML change — happy to split it into a standalone
PR if preferred.

### DIFF
--- a/backend/migrations/149_saml_map_groups_to_groups.sql
+++ b/backend/migrations/149_saml_map_groups_to_groups.sql
@@ -1,0 +1,13 @@
+-- SAML group-to-group mapping (parity with OIDC #1094).
+--
+-- Add map_groups_to_groups column to saml_configs. When true, the group
+-- values carried in the SAML assertion (via the `groups` attribute mapping)
+-- are reflected as Artifact Keeper group memberships, with groups auto-created
+-- on first sight and tagged external_source = 'saml'. When false, only the
+-- existing admin_group -> is_admin behavior applies.
+--
+-- Reuses the groups.external_source / external_provider_id columns and the
+-- idx_groups_external index already added in migration 100.
+
+ALTER TABLE saml_configs
+    ADD COLUMN map_groups_to_groups BOOLEAN NOT NULL DEFAULT false;
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -10669,12 +10669,12 @@ mod tests {
             assert!(
                 validate_virtual_repo_member_count("my-repo", &rt, None).is_ok(),
                 "{:?} with no members must be Ok",
-                &rt
+                rt
             );
             assert!(
                 validate_virtual_repo_member_count("my-repo", &rt, Some(&[])).is_ok(),
                 "{:?} with empty members must be Ok",
-                &rt
+                rt
             );
         }
     }
@@ -11597,7 +11597,7 @@ mod tests {
                      VALUES ($1, $2, $3, $4, $5, $6, $7)",
                 )
                 .bind(repo_id)
-                .bind(format!("app/{}", &key))
+                .bind(format!("app/{}", key))
                 .bind("manifest")
                 .bind(16_i64)
                 .bind("f".repeat(64))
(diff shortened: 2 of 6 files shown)


brandonrc:
Superseded by #2448.

Thanks @AkulovNV — this PR seeded the SAML group→group mapping feature, and its
design is the basis for what we're shipping. #2448 is the adopted/hardened
version, tracked against issue #2333 and milestone 1.6.0:

- rebased onto current `main` and composed with the SAML XML Signature Wrapping
  (XSW) hardening (#2449): the assertion `groups` attribute is read only from the
  cryptographically-verified assertion subtree, so an unsigned/out-of-subtree
  `groups` value cannot confer local group membership (a new e2e regression test
  pins this);
- gated behind an opt-in per-provider `map_groups_to_groups` flag (off by
  default — no behavior change for existing SAML providers on upgrade);
- source-scoped membership reconcile (`external_source='saml'` + provider id) so
  SAML, OIDC, and operator-managed memberships never strip one another;
- migration renumbered to 157 (the original number collided with migrations that
  landed on `main`);
- added unit, config round-trip, and live-ACS e2e coverage.

Closing in favor of #2448. Your contribution is credited there. Thank you again!

PR #2377, opened 2026-07-10
ci(release): require a CHANGELOG entry for the release version; add RELEASING.md

brandonrc:
## Summary

Makes "the CHANGELOG must document the release version" an enforced part of the release policy. v1.2.5 and v1.3.0 both shipped with their changes still under `## [Unreleased]` (neither tag has a promoted `## [X.Y.Z]` section); nothing in the release path caught it.

Two changes:

1. **`release.yml` / `verify-images-published`**: new step, before the registry polling, that checks out the tag and asserts `CHANGELOG.md` contains a non-empty `## [X.Y.Z]` section for the tagged version (the `## [X.Y.Z] - DATE` form is tolerated; exact-version match, so `1.2.50` cannot satisfy `1.2.5`). Missing or empty entry fails the job with a clear message ("promote [Unreleased] to a `## [X.Y.Z]` section before tagging"), which keeps the GitHub Release a draft via the existing gate-status logic. Prerelease tags (`-rc.N` etc.) are exempt, and the check only runs on the tag-triggered release path, so normal PRs and non-release runs are unaffected.

2. **`RELEASING.md`** (new): the cut-sequence runbook. Step 3 is the required CHANGELOG promotion, placed right before the tag step, and notes that the gate enforces it.

A companion PR in artifact-keeper-test adds the same assertion to the `version-set-integrity` job of the release gate (`verify-image-set.sh`), covering pre-tag gate dispatches that pass a real release version as `backend_tag`. The `release.yml` check here is needed because the tag-triggered gate call uses `backend_tag: dev`.

Verification: the step body passes shellcheck and was simulated locally against fixtures: version with a changelog section passes; missing section fails with the promote-[Unreleased] message; heading-with-no-content fails; prerelease tag skips; `## [1.2.50]` does not satisfy `1.2.5`; `## [X.Y.Z] - DATE` headings pass. Existing image-verification behavior is untouched (the new step is additive and runs first).

Closes #2376

## Regression test (required for `fix/*` PRs)
- [x] N/A — this is not a bug fix

## Test Checklist
- [ ] Unit tests added/updated
- [ ] Integration tests added/updated (if applicable)
- [ ] E2E tests added/updated (if applicable)
- [x] Manually tested locally
- [x] No regressions in existing tests

## API Changes
- [x] N/A - no API changes

PR #2416, opened 2026-07-11
fix(webhooks): claim retry deliveries atomically before re-POSTing

### LINKED ISSUE #2219: cluster-safe background work claims for multi-replica deployments

cazlo:
Partially fixes #2219 — second slice, following the sync-task claim that merged in #2221.

## Problem

`process_webhook_retries` SELECTs due deliveries, POSTs the webhook, and only then updates the row. With N backend replicas sharing one database, every replica's 30s retry tick selects the same due rows, so a single failed delivery can be re-POSTed up to N times per tick. Exhausting `max_attempts` is worse: the dead-letter path can fire the webhook auto-disable notifier once per replica.

## Fix

Due deliveries are claimed through the `cluster_work` `RowClaimedQueue` pattern that #2221 established for sync tasks:

- **Migration 154** adds `claimed_by` / `claim_token` / `claim_expires_at` to `webhook_deliveries`. The claim is expressed entirely through `claim_expires_at` (in-flight = unexpired claim), so `success` / `attempts` / `next_retry_at` keep their existing meanings, and a crashed owner's delivery simply becomes due again after the TTL with no operator action.
- `claim_due_webhook_deliveries` batches due rows in a single `FOR UPDATE SKIP LOCKED` statement and returns `Claimed<RetryDeliveryRow>` with a random per-claim token.
- The claim is re-extended immediately before each send: a 50-row batch at a 30s per-send timeout can far outlive one TTL, and a lost claim skips the send because the new owner will deliver it.
- Success, retry, and dead-letter updates are all token-guarded and release the claim, so a stale owner cannot clobber a delivery that has been re-claimed. The auto-disable notifier and dead-letter metrics only fire for the claim owner.

Migration numbering: this lands as **154** because open PR #2415 already carries `153_audit_correlation_text.sql`; whichever merges second needs no renumber.

## Tests

Two DB-gated tests cover the invariants directly rather than relying on timing:

- exactly-once claiming — two concurrent claimers over the same due set receive disjoint rows, and every row is claimed exactly once;
- stale-owner fencing — an expired/superseded claim cannot pass the pre-send extension, and its dead-letter finalizer is a no-op (no double auto-disable).

CHANGELOG entry added under `[Unreleased]` per the #2377 release gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Webhook retry deliveries are claimed atomically before the re-POST** (#2219): with multiple backend replicas sharing one database, every replica's retry tick selected the same due deliveries, so one failed delivery could be re-POSTed once per replica per tick — and exhausting `max_attempts` could fire the dead-letter auto-disable notifier once per replica. Due deliveries are now claimed in a single `FOR UPDATE SKIP LOCKED` statement with a per-claim token (the same `cluster_work` pattern as the #2275 sync-task claim), the claim is re-extended immediately before each send, and success/retry/dead-letter updates are token-guarded so a stale owner cannot clobber a re-claimed delivery. A crashed replica's claim simply expires, making the delivery due again with no operator action.
+
 ## [1.5.3] - 2026-07-12
 
 Security hotfix: closes a systemic cross-repository authorization gap where artifact/repository-scoped routes returned (and in some cases mutated) another repository's data for any authenticated non-member. Found by red-team; the HIGH-impact subset (vulnerability/SBOM data) is fixed here, with the remaining medium/low routes tracked for the next feature release (#2443).
--- a/backend/migrations/154_webhook_deliveries_claim.sql
+++ b/backend/migrations/154_webhook_deliveries_claim.sql
@@ -1,0 +1,23 @@
+-- Atomic multi-replica claims for webhook retry deliveries (RowClaimedQueue).
+--
+-- The retry processor used to SELECT due rows (`next_retry_at <= NOW()`),
+-- POST the webhook, and only then update the row, so with N replicas one
+-- failed delivery could be re-POSTed N times per retry tick. Due rows are now
+-- claimed with FOR UPDATE SKIP LOCKED before the send; result updates must
+-- present the claim_token.
+--
+-- The claim is expressed through claim_expires_at (in-flight = unexpired
+-- claim) rather than a new status enum: `success`/`attempts`/`next_retry_at`
+-- keep their existing meanings, and a crashed replica's claim simply expires,
+-- making the delivery due again.
+ALTER TABLE webhook_deliveries
+    ADD COLUMN claimed_by TEXT,
+    ADD COLUMN claim_token UUID,
+    ADD COLUMN claim_expires_at TIMESTAMPTZ;
+
+COMMENT ON COLUMN webhook_deliveries.claimed_by IS
+    'Diagnostic worker identity that claimed the retry; not the correctness guard';
+COMMENT ON COLUMN webhook_deliveries.claim_token IS
+    'Random per-claim token; delivery-result updates must match it';
+COMMENT ON COLUMN webhook_deliveries.claim_expires_at IS
+    'Retry-claim lease deadline; an expired claim makes the delivery due again';
--- a/backend/src/api/handlers/webhooks.rs
+++ b/backend/src/api/handlers/webhooks.rs
@@ -12,6 +12,7 @@ use uuid::Uuid;
 use crate::api::middleware::auth::AuthExtension;
 use crate::api::SharedState;
 use crate::error::{AppError, Result};
+use crate::services::cluster_work::{Claimed, WorkerIdentity};
 use crate::services::webhook_payloads::{self, PayloadTemplate};
 use crate::services::webhook_secret_crypto;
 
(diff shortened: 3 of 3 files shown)


brandonrc:
Adoption review for the 1.6.0 milestone — thanks @cazlo for a careful second slice of #2219, following the #2221 sync-task pattern. Validated as-is; no changes were needed on the branch.

**What was verified**

- **Migration numbering**: `154_webhook_deliveries_claim.sql` is correct — current `main` tops out at `153` (merged via #2415), so 154 is next-free. (Reminder for other 1.6.0 PRs carrying a migration: whichever merges after this one takes the next number.)
- **Correctness**: the claim CTE (`FOR UPDATE SKIP LOCKED` + stamped `claim_token` in one statement) leaves no select-to-update window; success/retry/dead-letter finalizers are all token-guarded and release the claim; the pre-send claim re-extension covers long batches; retry backoff and `success`/`attempts`/`next_retry_at` semantics are unchanged. The dead-letter auto-disable notifier + metric now fire only for the claim owner.
- **Gates** (branch merged with current `main`, i.e. post-#2481/#2479): `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, full lib suite `cargo nextest run --workspace --lib` **12957/12957 passed** against a migrated Postgres (both new DB-gated claim tests ran, not skipped), duplication under threshold. No `sqlx::query!` macros added (runtime `sqlx::query` only), so no offline-cache update is needed.
- **Live validation** on an isolated stack built from this branch: two backend replicas sharing one database; 20 due retry deliveries were re-POSTed **exactly once each** (20 hits, 20 unique delivery IDs at the receiver) and every row finished `success=true`, claim columns released, `next_retry_at` cleared. Two deliberately concurrent claim statements over 10 due rows returned disjoint 5/5 sets with zero overlap on the deployed schema. A normal (non-retry) test delivery still works.

Leaving "Partially fixes #2219" as-is is right — #2219 is the umbrella and several surfaces (backup occurrences, expiry emails, upload leases, OCI cleanup journal, singleton jobs) remain. Ready for maintainer merge.

Issue #2334, opened 2026-07-09
OpenAPI spec fidelity: six endpoints where the published spec disagrees with runtime behavior


brandonrc:
The published OpenAPI spec (exported via `export_openapi_spec` and synced to the api repo) disagrees with actual backend behavior on six endpoints. Strict generated SDKs (Progenitor-based CLI) fail on each mismatch. Verified against a live v1.4.0 instance. Tracker: artifact-keeper/artifact-keeper-cli#98.

1. **GET /api/v1/artifacts/{id} returns a shape the spec does not declare** (HIGH). `handlers/artifacts.rs` defines a local `ArtifactResponse` (raw DB row: `repository_id`, `checksum_md5`, `checksum_sha1`, `updated_at`; missing `download_count`, `analyzable`, `metadata`). The canonical `ArtifactResponse` in `handlers/repositories.rs` wins the shared schema-name slot (utoipa merge is first-wins), so the spec promises the rich shape but the endpoint serves the raw one.

2. **PATCH /api/v1/admin/lifecycle/{id} publishes the wrong request schema** (MED). `services/lifecycle_service.rs::UpdatePolicyRequest` (lifecycle fields) loses the schema-name slot to `handlers/security.rs::UpdatePolicyRequest` (quality-gate fields), so the spec tells clients to send quality-gate fields that the handler silently drops.

3. **POST /api/v1/repositories/{key}/email-subscriptions returns 200, spec declares 201** (LOW).

4. **GET/POST /api/v1/admin/ci-oidc published with a trailing slash** (LOW). Annotations use `path = "/"`; the emitted `/api/v1/admin/ci-oidc/` 404s at runtime (router serves the no-slash path).

5. **GET /api/v1/quality/checks: `artifact_id` optional in spec, required by the handler** (LOW). Handler 400s without it; spec also advertises an ignored `repository_id` param.

6. **PUT /api/v1/repositories/{key}/artifacts/{path} returns 201, spec declares only 200** (LOW). Package-manager clients depend on 201, so the spec must move to 201 (not the handler to 200).

Root cause for 1 and 2 is structural: schema-name collisions across handler modules are silently collapsed first-wins at merge time with no guard, so nothing catches a duplicate name until an SDK consumer breaks. A regression test asserting schema-name uniqueness across module docs would have caught both.

Deferred (tracked, not in scope here): the sso PUT-vs-PATCH verb mismatch — fixing it is a breaking route change with no SDK breakage today.

[...]

PR #2435, opened 2026-07-12
feat(composer): allow version and name override on upload

### LINKED ISSUE #1717: Feature request: allow overriding Composer package version via request parameter on upload
Summary
Please support optional version override for Composer uploads through request parameters or headers.

Motivation
Current docs and some workflows assume publish-time version override is possible. Current implementation extracts version from composer.json in the zip and falls back to dev-main when missing.

Current behavior
In backend/src/api/handlers/composer.rs upload handler:
- version is composer_json.version or dev-main
- request params are not used for version override

Requested behavior
Allow an explicit override on upload, for example:
- query parameter version
- or form/json field for version in a backward-compatible path

Possible rules
- if override is present, validate SemVer or accepted Composer version syntax
- if absent, keep existing behavior from composer.json
- return clear error on conflict between override and immutable policy if policy disallows mismatch

Compatibility considerations
- keep existing raw zip upload behavior intact
- ensure duplicate detection still checks repository, package name, effective version
- update metadata and package index using effective version

Optional enhancement
Accept package name override as well, with strict validation, to support CI release pipelines where archive metadata is fixed upstream.

Thanks for considering this enhancement.

ameshkov:
Closes #1717

## Summary
Add optional `?version=` and `?name=vendor/pkg` query-parameter overrides to the Composer upload endpoint (`PUT/POST /composer/{repo_key}/api/packages`).

When an override is provided it becomes the **canonical** name/version stored in the `artifacts` table and served to clients. The uploaded zip archive is stored **byte-for-byte** (never modified), and the original `composer.json` is preserved verbatim inside the nested `composer` metadata object — matching how Artifactory, Nexus, and Cloudsmith handle overrides (override stored separately, archive untouched).

Both the override values **and** the name/version read from `composer.json` are validated with the same helpers before any storage/DB write:

- `version` — Composer's documented version grammar (semver `v?X.Y[.Z[.W]]` with dot-separated pre-release and build metadata, plus `dev-<branch>` forms); rejects path separators, `..`, null/control chars, and >255 chars.
- `name` — the official Composer name regex (lowercase `vendor/package`); rejects missing slash, uppercase, spaces, special chars, and >512 chars.

Immutability is unchanged: duplicate detection and soft-delete cleanup key off the **effective** (override) name+version, so re-publishing the same name+version still returns `409 Conflict`. No migration is required — `artifacts.name` (`VARCHAR(512)`) and `artifacts.version` (`VARCHAR(255)`) are already sufficient.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/composer.rs
+++ b/backend/src/api/handlers/composer.rs
@@ -12,11 +12,11 @@
 //!   PUT  /composer/{repo_key}/api/packages                            - Upload/register package
 //!   POST /composer/{repo_key}/api/packages                            - Upload/register package
 //!
-//! Uploads accept optional `?version=X` and `?name=vendor/pkg` query overrides
-//! (#1717). When provided, the override values become the canonical name/version
-//! stored in the DB and served to clients; the uploaded archive is stored
-//! byte-for-byte and the original `composer.json` is preserved verbatim in the
-//! nested metadata. Immutability still keys off the effective name+version.
+//! Uploads accept optional `?version=X` and `?name=vendor/pkg` query overrides.
+//! When provided, the override values become the canonical name/version stored
+//! in the DB and served to clients; the uploaded archive is stored byte-for-byte
+//! and the original `composer.json` is preserved verbatim in the nested
+//! metadata. Immutability still keys off the effective name+version.
 
 use axum::body::Body;
 use axum::extract::{Path, Query, State};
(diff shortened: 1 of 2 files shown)


brandonrc:
Thanks @ameshkov — this is cleanly implemented and the validation is solid. But we're going to keep this door closed: #2334/#2412 deliberately made `composer.json` the single source of truth for name/version (documented in the site's composer guide), and we'd rather not reintroduce an override path — even an opt-in one — that reopens the ambiguity that fix removed. If there's a concrete #1717 workflow that composer.json-as-truth genuinely can't serve, let's discuss it on #1717 and find a solution that keeps the contract. Really appreciate the contribution.

PR #2448, opened 2026-07-12
feat(sso): map SAML assertion groups to local group memberships (#2333)

### LINKED ISSUE #2333: SAML: map assertion groups to group memberships (parity with OIDC #1094)

brandonrc:
Closes #2333

## Summary

Adds a per-provider `map_groups_to_groups` flag to SAML configs (migration
**157**), mirroring the OIDC group-to-group mapping from #1094. When enabled,
group values carried in the SAML assertion (via the `groups` attribute mapping)
are reflected as Artifact Keeper group memberships: groups are auto-created
tagged `external_source = 'saml'` and reconciled on every login. When disabled
(the default), only the existing `admin_group` → `is_admin` behavior applies, so
existing providers are unchanged on upgrade.

This matters because some IdPs (e.g. Google Workspace) emit group membership
only over SAML, not in the OIDC id_token — on those IdPs, SAML users could
authenticate but could not receive any group-based (non-admin) permissions.

Adopts and completes the design from @AkulovNV's #2332 (thanks @AkulovNV!):

- migration renumbered to **157** (`154`/`155`/`156` are now taken by
  webhook-deliveries / projects / npm-feed on `main`; the duplicate version
  broke `sqlx migrate`) — verified applying in ascending order on a fresh DB
- added the unit / config round-trip / live-ACS e2e test coverage
- kept the auto-created group description per-source (`OIDC` casing preserved
  byte-identical for the existing OIDC path)
- `redacted_debug!(SamlConfigRow …)` shows the new column
- dropped #2332's bundled repositories.rs clippy hunk (already on `main`)

## Security composition with the SAML XSW fix (#2449)

Rebased onto current `main`, which carries the XML Signature Wrapping (XSW) fix
(#2449). That fix scopes SAML claim extraction to the cryptographically-verified
assertion subtree (`in_assertion` gate, excluding the `<ds:Signature>` subtree,
comment-safe text). **The group-mapping feature reads the assertion `groups`
attribute from that same signed-and-verified path**: `saml_acs` takes
`saml_user.groups`, which is populated by `SamlService::authenticate()` →
`extract_user_info()` from `assertion.attributes` — the map the XSW parser fills
only while `should_consume_claim()` (`in_assertion && in_signature_depth == 0`)
holds. No whole-document or unsigned source is read. An attacker who completes
SSO therefore cannot inject an unsigned/out-of-subtree `groups=[…]` to join
arbitrary local groups. A new e2e regression test
(`test_saml_acs_out_of_subtree_groups_confer_no_membership`) pins this: with the
flag on, a `groups` attribute spliced OUTSIDE the signed assertion confers NO
membership (login still 307s as the signed subject).

## How

- The OIDC group-sync core is generalized into
  `sync_federated_groups_to_local_groups(pool, user_id, provider_id, source, groups)`;
  `sync_oidc_groups_to_local_groups` becomes a thin wrapper, so the OIDC call
  site and all #1094 tests are untouched.
- `saml_acs` invokes it with `source = "saml"` when the provider has
  `map_groups_to_groups` enabled. Sync failure is logged and never blocks login.
- The reconcile DELETE is scoped by `(external_source, external_provider_id)`,
  so SAML and OIDC providers never strip each other's memberships — even when
  they share a provider UUID — and operator-managed groups (NULL
  `external_source`) are never touched.
- Config plumbing mirrors `use_absolute_acs_url` (migration 139) 1:1: row /
  response / create / update structs, all seven SAML SELECT/RETURNING lists,
  INSERT + UPDATE binds, `saml_row_to_response`, and every test fixture.

## Tests

- **Unit (sso.rs `sync_db`)**: SAML groups auto-created tagged `'saml'` +
  membership; selective prune (`[eng,ops]` → `[eng]`); isolation — operator
  (NULL source) and OIDC (same provider UUID) memberships survive a SAML sync;
  symmetric — an OIDC sync does not strip SAML memberships; empty/whitespace
  group names skipped. All existing #1094 OIDC tests unchanged and green.
- **Config round-trip (auth_config_service.rs)**: default false on create;
  create-true round-trips through get / `get_saml_decrypted` (the row
  `saml_acs` reads) / list; update preserves when omitted and flips when set.
- **E2E (sso_saml_acs_e2e_tests.rs)**: signed assertions through the live ACS
  handler — full sync + reconcile + isolation, default-off no-op, and the new
  XSW-composition regression (out-of-subtree `groups` → no membership).

- [x] N/A — this is not a bug fix (new feature; security-regression e2e added
      for the XSW composition guarantee)

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] New endpoints have `#[utoipa::path]` annotations (no new endpoints; new
      field on existing `ToSchema` request/response types)
- [x] Migration is reversible (if applicable) — additive column with default
- [ ] N/A - no API changes


brandonrc:
⛔ HOLDING merge — red-team found a pre-existing CRITICAL XML Signature Wrapping vuln in the base SAML validation (https://github.com/artifact-keeper/artifact-keeper/issues/2449) that this group-sync feature weaponizes into admin escalation. The feature itself validated clean (isolation holds, negative controls 401). This PR will rebase on top of the XSW fix and re-red-team before merging.

Issue #2507, opened 2026-07-14
Separate S3 Buckets for storage and backups


BonjeBongo:
### Problem Statement

Would help to separate artifact storage and backups as we usually have different s3 backup/lifecycle configurations. at the moment they all are dumped into s3.

### Proposed Solution

Add an optional BACKUP_S3_BUCKET variable

### Component

Backend / API

### Pre-submission Checklist

- [x] I have searched existing issues and feature requests to ensure this isn't a duplicate
- [x] I have reviewed the documentation to confirm this feature doesn't already exist

PR #2506, opened 2026-07-14
feat(backups): optional dedicated S3 bucket via BACKUP_S3_BUCKET

BonjeBongo:
## Summary
Closes #2507.

Backups currently always share the primary artifact storage bucket (`S3_BUCKET`), with no way to isolate backup data into a separate bucket. This adds an optional `BACKUP_S3_BUCKET` env var: when set (and `STORAGE_BACKEND=s3`), backups are written to that bucket instead, reusing the same S3 credentials/region/endpoint. Falls back to `S3_BUCKET` when unset, so existing deployments are unaffected.

- Added `Config.backup_s3_bucket` (env `BACKUP_S3_BUCKET`), following the existing `s3_bucket`/`S3_BUCKET` pattern in `backend/src/config.rs`.
- Added `StorageService::from_config_for_backups`, which builds the storage backend the same way as `from_config` but overrides the S3 bucket when `backup_s3_bucket` is set (extracted into a small pure helper, `apply_bucket_override`, for isolated unit testing).
- Switched all backup call sites (admin backup handlers, cleanup-old-backups, and the scheduled-backup scheduler) from `StorageService::from_config` to `StorageService::from_config_for_backups`.
- Documented `BACKUP_S3_BUCKET` in `.env.example` and added a CHANGELOG entry.

## Regression test (required for `fix/*` PRs)
- [ ] This PR is a `fix/*` AND adds/updates a test that would have caught the bug
- [x] N/A — this is not a bug fix

## Test Checklist
- [x] Unit tests added/updated (`Config::from_env` coverage for `BACKUP_S3_BUCKET`, `StorageService::apply_bucket_override`, and a fallback test for `from_config_for_backups` on non-s3 backends)
- [ ] Integration tests added/updated (if applicable)
- [ ] E2E tests added/updated (if applicable)
- [ ] Manually tested locally
- [x] No regressions in existing tests (`cargo test --workspace --lib` passes; 3 pre-existing failures confirmed unrelated and present on `main` as well — timing-sensitive proxy/http-client tests)

## API Changes
- [ ] New endpoints have `#[utoipa::path]` annotations
- [ ] Request/response types have `#[derive(ToSchema)]`
- [ ] OpenAPI spec validates: `cargo test --lib test_openapi_spec_is_valid`
- [ ] Migration is reversible (if applicable)
- [x] N/A - no API changes

PR #2510, opened 2026-07-14
feat(backups): optional dedicated S3 key prefix via BACKUP_S3_PREFIX and S3_PREFIX

### LINKED ISSUE #2508: Allow custom S3 Paths
### Problem Statement

in addition to #2507 it would also be nice to specify the S3 paths for where things are stored.
Artifacts are currently dumped into root which is a problem if you have a backup bucket used by other apps
backups are dumped into `/backups`

### Proposed Solution

add S3_PREFIX (and BACKUP_S3_PREFIX) optional variables to configure s3 paths

### Alternatives Considered

_No response_

### Use Case

_No response_

### Component

Backend / API

### Additional Context

_No response_

### Pre-submission Checklist

BonjeBongo:
Follow-up to #2506 / #2507. Resolves #2508
`S3_PREFIX` already worked for primary artifact storage (`main.rs` builds that backend directly via `S3Backend::from_env()`), but the shared `StorageService` facade used by both the proxy cache and backups forced the S3 key prefix to `None` unconditionally — silently discarding `S3_PREFIX` for backups. This adds `BACKUP_S3_PREFIX`: when set (and `STORAGE_BACKEND=s3`), backups are written under that key prefix instead of the bucket root, falling back to `S3_PREFIX` and then no prefix when unset.

- Added `Config.s3_prefix` (env `S3_PREFIX`) and `Config.backup_s3_prefix` (env `BACKUP_S3_PREFIX`) in `backend/src/config.rs`, following the existing bucket-field pattern.
- Generalized `StorageService::from_config_with_bucket_override` into `from_config_with_overrides`, which now takes both a bucket override and a prefix override. Added `apply_prefix_override`, a small pure helper mirroring `apply_bucket_override`, so the prefix decision is unit-testable in isolation.
- `from_config` (proxy-cache path) still forces the prefix to `None` unconditionally — that invariant is unrelated to this feature and unchanged.
- `from_config_for_backups` now resolves the prefix as `backup_s3_prefix.or(s3_prefix)`, so backups pick up a bucket-root-adjacent path by default and can be overridden independently of primary storage.
- Documented `BACKUP_S3_PREFIX` in `.env.example` and added a CHANGELOG entry.

## Test Checklist
- [x] Unit tests added/updated (`Config::from_env` coverage for `S3_PREFIX`/`BACKUP_S3_PREFIX`, `StorageService::apply_prefix_override`, and the `backup_s3_prefix`/`s3_prefix` fallback-chain resolution)
- [ ] Manually tested locally
- [x] No regressions in existing tests (`cargo test --workspace --lib`: 13,101 passed; the only 3 failures are pre-existing, timing-sensitive proxy/http-client tests confirmed to fail identically on a clean `main` checkout, unrelated to this change)

## API Changes
- [x] N/A - no API changes


brandonrc:
Thanks for this, @BonjeBongo — and apologies it got overtaken by parallel work. The `BACKUP_S3_PREFIX` capability this adds has since landed on main via #2755 (which closed the linked #2508), and the `BACKUP_S3_BUCKET` piece carried in this branch landed via #2903 (closing #2507). The merged approach applies the prefix at backup-archive key construction while keeping the storage facade prefix-less (needed for proxy-cache presigning) and preserves today's key shape when the var is unset, so existing archives stay restorable without migration.

Because the underlying issues are delivered and test-covered on main, I'm closing this as superseded — no reflection on the contribution, which flagged the need first. If you spot a gap between what merged and what you intended here, please reopen or ping and we'll take another look.

PR #2550, opened 2026-07-14
feat(admin-security): enumerate accessible-but-not-downloaded users for restricted-repo CVE blast radius (#2386)

### LINKED ISSUE #2386: CVE blast-radius phase 2: enumerate accessible-but-not-downloaded users for restricted repos
Follow-up to #2364 (CVE blast-radius, phase 1). Phase 1 ships the confirmed-downloaders JOIN + a bounded per-repo `access_scope` classification (public / restricted_acl / restricted_roles). Full enumeration of users who *can access* a vulnerable artifact but haven't downloaded it was deferred here because it's unbounded for public/globally-granted repos.

## Ask (phase 2)
For **restricted** repos only, enumerate accessible-but-not-downloaded users with a `reason: has-access` list:
- ACL'd repos → `permissions` read/admin grants + group expansion (permission_service.rs query_actions UNION over user_group_members).
- Rule-less private repos → users with a scoped OR global (`repository_id IS NULL`) `role_assignments` row (auth.rs:1660-1670), plus admins.
- **Public + globally-granted repos**: do NOT enumerate — keep the phase-1 flagged classification 'exposure = all users' (unbounded).

## Open product decision (needs a call before building)
(a) restricted-only enumeration as above (recommended), or (b) leave 'can access' as classification-only permanently. Public/global exposure stays a flagged 'everyone' either way.

## Acceptance
- Restricted-repo accessible-user enumeration is bounded + matches repo_visibility_middleware's read-access branch (auth.rs:1394-1695).
- Public/global repos flagged 'everyone', never enumerated.
- Admin-only; paginated.

brandonrc:
## Summary

Phase 2 of the CVE blast-radius reporting (#2386). Phase 1 answers "who
**downloaded** a vulnerable artifact"; this adds the latent blast radius: the
principals who **can read** an affected artifact in a **restricted** repository
but have **no download record** yet. This lets an admin responding to a CVE see
the full set of accounts that could have pulled the vulnerable artifact, not
only those already recorded as having done so.

Two new admin-only, paginated, read-only endpoints under the existing
`/admin/security` nest (so `admin_middleware` continues to gate them, with a
`require_admin` defense-in-depth re-check per handler, exactly like phase 1):

- `GET /api/v1/admin/security/cve/{cve_id}/accessible-users?repository_id=&page=&per_page=`
  (`repository_id` required — a CVE spans many repos; validated to actually
  affect the repo).
- `GET /api/v1/admin/security/artifact/{artifact_id}/accessible-users?page=&per_page=`
  (repository implied by the artifact).

Each returns the page of `accessible_not_downloaded` principals with
`reason: "has-access"` and a `via` label (`admin` / `permission` / `role`).

Key design points:

- **Reuses the exact data-plane read predicate, no hand-rolled SQL.** The
  enumeration inverts the shared `permissions_grant_exists` fragment
  (`repository_service.rs`) — promoted to a `pub(crate)` expression-based
  variant `permissions_grant_exists_for` so the same string (project arm + group
  UNION + `actions <> '{}'` fail-closed) drives both the live access check and
  the report. The positional-bind `permissions_grant_exists` wrapper now
  delegates to it, so its generated SQL and existing callers/tests are
  unchanged. Access = `role_assignments` (repo or global NULL) **OR** the shared
  permissions grant **OR** admin, anti-joined against `download_statistics`.
- **Superset (REST) read definition, deliberately.** On an ACL'd repo a user
  holding only a `role_assignment` is allowed by the REST download path but
  denied by the native package-manager middleware branch. For a *security*
  blast-radius report, over-reporting who can pull a vulnerable artifact is the
  safe direction (never hide an exposed principal); the `via` field lets a
  reviewer see which entries are role-only. Called out here per the FixSpec so
  the issue owner can confirm the choice.
- **Never enumerates public / everyone-exposed repos.** A `public` repo returns
  `exposure: "everyone"` with an empty list (no enumeration). When the
  accessible set reaches an `effectively-everyone` threshold (>= 90% of active,
  non-service users — catches a global NULL-scoped role granted to everyone) the
  report collapses to `exposure: "effectively-everyone"` with a count only, no
  page body.
- **On-demand, not materialized.** Two bounded queries per call (a `COUNT` for
  pagination + threshold, one `LIMIT/OFFSET` page, clamped to <=100 via the
  existing `blast_page_bounds`). A materialized table would fan invalidation out
  across the entire authz surface; not worth it for an admin report.

- [x] N/A — this is not a bug fix (feat/)

## Test Checklist
- [x] Unit tests added/updated — `is_effectively_everyone` threshold,
  `classify_exposure` branches, and an enumeration-SQL-shape assertion (role arm
  + shared permissions fragment + download anti-join).
- [x] Integration tests added/updated — `#[sqlx::test]`-style DB test seeding a
  restricted repo with direct-permission / group / role / global-NULL-role /
  admin / downloader / no-access users: asserts the downloader and no-access
  users are excluded, the rest are included with correct `via`, each enumerated
  non-admin user passes `user_can_access_repo` (predicate parity), pagination
  windowing is disjoint, and a public repo returns `everyone`. Plus an
  admin-only router test (non-admin -> 403, missing `repository_id` -> 400,
  unknown artifact -> 404).
- [x] Manually tested locally — deployed on an isolated pool instance:
  restricted repo lists the granted-not-downloaded user (`via: permission`) and
  the globally-granted admin, excludes the downloader; deleting the download row
  makes that user reappear (`via: role`) and re-adding it excludes them again;
  public repo -> `exposure: everyone`; anonymous -> 401; non-admin -> 403;
  `per_page=1` paginates correctly.
- [x] No regressions in existing tests — the shared-fragment shape tests and the
  visibility-clause tests remain green (fragment SQL byte-identical).

## API Changes
- [x] OpenAPI spec validates
- [x] Migration is reversible (if applicable) — `161_role_assignments_repository_id_index.sql`
  adds `idx_role_assignments_repository_id` (`IF NOT EXISTS`, index-only,
  non-blocking; `role_assignments` was the only join table without a
  `repository_id` index, its unique constraint leads on `user_id`).
- [ ] N/A - no API changes

Closes #2386

### DIFF
--- a/backend/migrations/161_role_assignments_repository_id_index.sql
+++ b/backend/migrations/161_role_assignments_repository_id_index.sql
@@ -1,0 +1,12 @@
+-- Index role_assignments(repository_id) for the CVE blast-radius
+-- accessible-users enumeration (#2386).
+--
+-- role_assignments' only index today is its UNIQUE(user_id, role_id,
+-- repository_id) constraint, whose leading column is user_id. The enumeration
+-- filters `repository_id = $1 OR repository_id IS NULL` (a non-leading column),
+-- so that constraint cannot serve it and large deployments seq-scan the table.
+-- This partial-free btree serves both the `= $repo` and `IS NULL` scans.
+--
+-- Non-blocking, idempotent; no data change.
+CREATE INDEX IF NOT EXISTS idx_role_assignments_repository_id
+    ON role_assignments(repository_id);
--- a/backend/src/api/handlers/admin_security.rs
+++ b/backend/src/api/handlers/admin_security.rs
@@ -64,7 +64,8 @@ const MAX_IPS_PER_DOWNLOADER: u32 = 50;
 /// Cap on the `affected_repos` block. `summary.affected_repo_count` remains
 /// the true total even when the list is truncated.
 const MAX_AFFECTED_REPOS: u32 = 200;
-/// Fraction of active, non-service users at/above which a restricted repo's
+/// Fraction of active users (the same population the enumeration draws from —
+/// including admins and service accounts) at/above which a restricted repo's
 /// accessible set is treated as "effectively everyone" — the page body is
 /// suppressed (reporting ~all users is not actionable, and this catches the
 /// global NULL-scoped `role_assignment` granted to every user). The count is
@@ -679,12 +680,17 @@ async fn accessible_users_core(
     count_q = bind_target(count_q, &target);
     let accessible_count: i64 = count_q.fetch_one(db).await.map_err(db_err)?;
 
-    // Active, non-service user population for the effectively-everyone check.
-    let active_user_count: i64 =
-        sqlx::query_scalar("SELECT COUNT(*) FROM users WHERE is_active AND NOT is_service_account")
-            .fetch_one(db)
-            .await
-            .map_err(db_err)?;
+    // Candidate population for the effectively-everyone check: ALL active users.
+    // This must match the population the enumeration draws from (`users u WHERE
+    // u.is_active`) so numerator and denominator are apples-to-apples — the
+    // accessible set includes admins and service accounts (both can pull and
+    // both appear in the list), so both must be counted here too. Excluding
+    // service accounts from only the denominator would trip the 90% collapse
+    // early.
+    let active_user_count: i64 = sqlx::query_scalar("SELECT COUNT(*) FROM users WHERE is_active")
+        .fetch_one(db)
+        .await
+        .map_err(db_err)?;
 
     let exposure = classify_exposure(&access_scope, accessible_count, active_user_count);
     if exposure != "enumerable" {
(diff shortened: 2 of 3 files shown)


brandonrc:
Addressed red-team findings on this branch (commit 9fbd7b7):

**MEDIUM (fixed) — CVE download anti-join is now repo-scoped.** `CVE_AFFECTED_SQL` now intersects the CVE's flagged artifacts with the target repository (`JOIN artifacts a ON a.id = sf.artifact_id WHERE sf.cve_id = $2 AND a.repository_id = $1`). Previously the anti-join was CVE-global, so a user who downloaded *another repo's* (even another tenant's) copy of the same CVE was wrongly hidden despite retaining latent access to this repo's un-downloaded vulnerable artifact — and it disagreed with the `/artifact/{id}/accessible-users` answer for the same artifact. Now the two routes agree. Regression locked by `test_accessible_users_cve_antijoin_is_repo_scoped_db` and a shape assertion.

Live re-verify (isolated slot): repo `RA` and `RB` both flagged by `CVE-2386-XSPAN`; `xspan` has role access on `RA`, downloaded only `RB`'s copy; `ydl` downloaded `RA`'s own copy.
- CVE route `?repository_id=RA` -> lists `[admin, bf-xspan-2386(role)]`, total 2 (**xspan now LISTED**)
- Artifact route for RA's artifact -> lists `[admin, bf-xspan-2386(role)]`, total 2 (**agrees**)
- `ydl` (downloaded the target repo's artifact) correctly excluded in both.

**LOW (fixed) — effectively-everyone population consistency.** The denominator was `WHERE is_active AND NOT is_service_account` while the numerator (and the enumeration) includes admins and service accounts. Changed the denominator to `COUNT(*) WHERE is_active` so it matches the exact population the enumeration draws from (`users u WHERE u.is_active`) — apples-to-apples, no early collapse.

fmt/clippy/dup green; full admin_security lib suite (15 tests) green.

PR #2572, opened 2026-07-15
fix(migration): fetch and register referenced OCI content for Docker imports

### LINKED ISSUE #2457: Nexus→AK migration copies Docker manifests to the store but never registers them in the registry index — migrated images unpullable (MANIFEST_UNKNOWN) despite job completed N/0/0 (server 1.5.4)
## Summary

A Nexus→Artifact Keeper migration of a Docker **hosted** repo reports success (`status: completed`, N migrated, 0 failed, 0 skipped), and the migrated manifests are visible in the UI and servable via the repository download API — **but the Docker Registry V2 API never learns about them**. `docker pull` of a migrated tag fails with `manifest unknown`. Only tags pushed **natively** (`docker push`) are pullable. Server 1.5.4, connection `source_type: "nexus"`.

## Repro (via the documented Nexus migration procedure)

1. Source connection to Nexus 3.x, `source_type: "nexus"` (test → Verified).
2. `POST /api/v1/migrations` with
   ```json
   {"source_connection_id":"…","job_type":"full","config":{"include_repos":["docker-releases"],"conflict_resolution":"overwrite","concurrent_transfers":4,"throttle_delay_ms":100,"verify_checksums":true}}
   ```
   then `POST /api/v1/migrations/{id}/start`.
3. Job finishes: `status: completed`, `completed_items: 305`, `failed_items: 0`, `skipped_items: 0`.

## Observed

The manifest is in the artifact store — visible in the UI and served by the download API:
```
GET /api/v1/repositories/docker-releases/download/docker-releases/v2/<image>/manifests/<version-tag>  -> 200 (file)
```

But the registry index does not have it:
```
GET /v2/docker-releases/<image>/tags/list
  -> {"name":"<image>","tags":["demo","dev","test"]}   # migrated version tag ABSENT (only natively-pushed tags)

GET /v2/docker-releases/<image>/manifests/<version-tag>
  -> 404 {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest not found"}]}

docker pull <host>/docker-releases/<image>:<version-tag>
  -> manifest unknown: manifest not found
```

## Control (registry / auth / blobs / addressing are all fine)

`docker pull <host>/docker-releases/<image>:dev` — a **natively-pushed** tag — succeeds and pulls all layers. So the registry, token auth, blob store and image addressing work; **only migrated tags are affected**.

## Root cause (localized)

AK's Docker registry resolves manifests from its OCI index tables (`oci_tags` / `oci_manifest` / `oci_manifest_refs` — cf. #1682). The migration writes the manifest and blob **files** into the artifact store but does **not** create the corresponding `oci_manifest` / `oci_tags` / `oci_manifest_refs` rows, so `/v2/.../manifests/<ref>` can't resolve tag→digest→manifest for migrated images. Re-running `full` with `conflict_resolution: overwrite` + `verify_checksums: true` re-processes every item (`completed 305/0/0`) and still doesn't register them.

## Expected (per docs)

The [Nexus migration guide](https://artifactkeeper.com/docs/migration/from-nexus/) states "Docker tags get migrated" and tells you to verify by pulling images. Migrated Docker images should be pullable.

## Impact

Migrated Docker repositories are unusable via `docker pull` out of the box: images appear in the UI and via the download API, but the registry returns `MANIFEST_UNKNOWN`. Every migrated tag must be re-pushed natively to become pullable.

## Suggested fix

When migrating Docker/OCI artifacts, register them in the registry index (populate `oci_manifest` / `oci_tags` / `oci_manifest_refs`, linking tag → digest → manifest and referenced blobs), not only copy files into storage.

## Related

#1682 — retention vs `oci_tags` (same OCI index tables, different lifecycle).

brandonrc:
## Summary

Docker/OCI migration imports registered the enumerated **tag manifests**
(`oci_tags`, index→child edges, `manifest_blob_refs`) but never fetched the
content those manifests *reference*. A source registry (Nexus, any conformant
`/v2/` registry) lists only the tag manifests; the image `config` blob, every
`layers[]` blob, and — for a multi-arch image index — the per-arch child
manifests plus their own config/layer blobs are addressed by digest and are
**not** enumerated. The result was a *hollow* image: `oci_blobs` stayed empty,
`GET /v2/<repo>//blobs/<config>` and `.../manifests/<child>` returned 404,
and `docker pull` failed even though the job reported success.

This PR hardens the import to transfer referenced content with full fidelity:

- **Referenced-content walker** (`services/oci_referenced_content.rs`): after a
  Docker/OCI manifest is registered, the walker resolves its referenced content
  by digest — for an image manifest, the `config` blob and every layer; for an
  image index, each child manifest (registered so it resolves by digest, exactly
  as a live `docker push` records a by-digest manifest) and then that child's
  own config/layers. Real bytes land in `oci_blobs` and storage, not just edges.
  A new `SourceRegistry::download_oci_content_stream(repo, image, digest, kind)`
  fetches content by digest (default targets the `/v2/` registry layout Nexus
  serves; Artifactory overrides for its Docker registry API path).

- **Fail-closed**: the walk runs inside the item's transaction. If any
  referenced blob or child manifest cannot be transferred, the whole item rolls
  back (no `oci_tags`, no partial `oci_blobs`) and is marked FAILED — a migrated
  tag is never acked while its content is incomplete.

- **Streaming + bounded**: every blob is spilled through a temp file and
  `put_stream` (O(chunk) memory, matching the existing transfer path). A
  visited-digest set dedups shared layers; recursion depth, per-index fan-out,
  and total blob/manifest counts are capped. Content already present
  (`oci_blobs` rows / stored manifests — e.g. blobs the Artifactory source
  enumerates as their own items) is skipped without a fetch, so the walker is a
  no-op there and cannot regress the Artifactory path.

- **Resurrect soft-deleted rows** (`migration_worker`): re-importing a manifest
  whose `artifacts` row was soft-deleted now upserts `is_deleted = false` (the
  full `UNIQUE(repository_id, path)` arbiter) instead of `DO NOTHING`, matching
  the live push path. The `artifacts` INSERT and the OCI registration for a
  Docker item share ONE transaction, so a tag is never committed without a live
  backing artifact.

- **Startup reconciliation** (`services/oci_migration_reindex.rs`): the existing
  one-shot repair now recurses image indexes (registering present child
  manifests + their blobs), drops orphan `oci_tags` whose backing artifact was
  soft-deleted (narrowly scoped to migrated tags — a natively pushed tag has no
  `artifacts` row and is never touched), and emits a WARN listing repositories
  whose tags remain hollow (content that was never transferred cannot be
  fabricated; those repos need re-migration). `RepairStats` gains
  `children_registered` / `hollow_tags_flagged` / `orphan_tags_reconciled`,
  surfaced in the `main.rs` startup log.

Non-Docker formats and the Artifactory-shaped import keep their existing
behaviour byte-for-byte. The `from_artifactory`/Nexus repository-type mapping is
already merged and is not part of this change.

Closes #2457

New/updated tests:
- Walker unit tests: config + all layers for an image; recurse children then
  their config/layers for an index; shared-layer visited-set dedup; depth /
  fan-out DoS caps reject a pathological index; an unfetchable blob fails the
  walk (fail-closed); a child digest shared with another image is still
  registered under the current image and survives the other tag's deletion.
- Migration integration tests: a source enumerating ONLY the tag manifest leaves
  the image fully pullable (`oci_blobs` populated, child resolves by digest); a
  source that 404s a layer marks the item FAILED with nothing committed; a
  delete-then-re-import resurrects the soft-deleted row with zero orphan tags.
- Reindex tests: image-index repair registers children + their blobs; orphan-tag
  reconciliation drops a tombstoned-backing tag while keeping a healthy one.
- Updated the pinned `extract_blob_refs_empty_for_image_index` with a companion
  asserting the walker classifies the index and recurses its children.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/migration.rs
+++ b/backend/src/api/handlers/migration.rs
@@ -1469,7 +1469,7 @@ async fn stream_migration_progress(
         // Send initial connection event
         yield Ok(Event::default().event("connected").data(format!(r#"{{"job_id":"{}"}}"#, id)));
 
-        let terminal_statuses = ["completed", "failed", "cancelled"];
+        let terminal_statuses = ["completed", "completed_with_errors", "failed", "cancelled"];
 
         loop {
             // Fetch current progress
@@ -1605,7 +1605,10 @@ async fn list_migration_items(
 /// audit report is meaningful. Mirrors the terminal set used by the progress
 /// stream loop.
 fn is_terminal_status(status: &str) -> bool {
-    matches!(status, "completed" | "failed" | "cancelled")
+    matches!(
+        status,
+        "completed" | "completed_with_errors" | "failed" | "cancelled"
+    )
 }
 
 /// Fetch the persisted report row for a job, if one exists.
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -2111,7 +2111,40 @@ pub(crate) async fn persist_tag_and_refs(
     manifest_body: &[u8],
 ) -> Result<(), sqlx::Error> {
     let mut tx = pool.begin().await?;
+    persist_tag_and_refs_in_tx(
+        &mut tx,
+        repo_id,
+        name,
+        tag,
+        manifest_digest,
+        manifest_content_type,
+        class,
+        manifest_body,
+    )
+    .await?;
+    tx.commit().await?;
+    Ok(())
+}
 
+/// Transaction-participating form of [`persist_tag_and_refs`]. Runs the tag
+/// upsert and reference recording against a caller-owned transaction WITHOUT
+/// committing, so the caller can bind the registration to a larger atomic
+/// unit — e.g. the migration worker registers the `artifacts` row, the tag,
+/// and every referenced blob/child manifest in ONE transaction so a migrated
+/// tag is never committed without its backing content (#2457). The pool-taking
+/// [`persist_tag_and_refs`] is a thin wrapper that begins + commits its own
+/// transaction, preserving the live push path's behaviour byte-for-byte.
+#[allow(clippy::too_many_arguments)]
+pub(crate) async fn persist_tag_and_refs_in_tx(
+    tx: &mut sqlx::Transaction<'_, sqlx::Postgres>,
+    repo_id: Uuid,
+    name: &str,
+    tag: &str,
+    manifest_digest: &str,
+    manifest_content_type: &str,
+    class: &ManifestClass,
+    manifest_body: &[u8],
+) -> Result<(), sqlx::Error> {
     // 1. Tag upsert (identical SQL/semantics to the previous standalone form).
     sqlx::query(
         r#"INSERT INTO oci_tags (repository_id, name, tag, manifest_digest, manifest_content_type)
@@ -2126,7 +2159,7 @@ pub(crate) async fn persist_tag_and_refs(
     .bind(tag)
     .bind(manifest_digest)
     .bind(manifest_content_type)
-    .execute(&mut *tx)
+    .execute(&mut **tx)
     .await?;
 
     // 2. Reference recording, in the SAME transaction. A failure here rolls
@@ -2146,7 +2179,7 @@ pub(crate) async fn persist_tag_and_refs(
                 .bind(manifest_digest)
                 .bind(&children)
                 .bind(repo_id)
-                .execute(&mut *tx)
+                .execute(&mut **tx)
                 .await?;
             }
         }
@@ -2178,7 +2211,7 @@ pub(crate) async fn persist_tag_and_refs(
                 )
                 .bind(repo_id)
                 .bind(&blob_digests)
-                .fetch_all(&mut *tx)
+                .fetch_all(&mut **tx)
                 .await?;
 
                 // 2b. (#1660) Resurrect any of these blobs that blob GC has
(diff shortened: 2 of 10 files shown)


brandonrc:
Follow-up round (second commit) — data-fidelity hardening + partial-status:

- **Partial-failure status**: a migration where some items fail while others
  succeed now resolves to `completed_with_errors` instead of a clean
  `completed`, so a hollow/partial import is not surfaced as success. Per-item
  counters, report, and error messages are unchanged. No schema migration
  (`migration_jobs.status` is free-text `VARCHAR(50)`); `is_terminal_status`
  and the progress-stream terminal set recognise the new value.
- **Outbound-URL validation (defense-in-depth)**: before building the by-digest
  OCI fetch URL, the digest is validated against `sha256:<64 lowercase hex>` and
  the image against the OCI repository-name grammar; traversal/injection values
  are rejected. This complements the already digest-verified, discarded response.
- **Mid-stream manifest cap**: the manifest size cap is now enforced while
  streaming (abort before the overflowing chunk is written) so a hostile source
  cannot spill a large body to the staging temp file behind a manifest URL
  before rejection. Blob streaming is unchanged.

New unit tests cover the digest/image validators, the mid-stream abort, and the
partial-status mapping.

brandonrc:
Closing: superseded. Fetching and registering referenced OCI content for Docker imports landed on main as #2582 and shipped in v1.5.8 (this branch is DIRTY against a tree that already contains the fix).

PR #2573, opened 2026-07-15
fix(proxy): allow the configured egress proxy host through the SSRF DNS guard

### LINKED ISSUE #2570: Remote repositories return 502 when an outbound proxy is configured
## Description

Artifact Keeper is deployed on OpenShift using the Helm chart.

External network access is only available through a corporate HTTP/HTTPS proxy. Proxy environment variables are configured in the Artifact Keeper backend container and are visible inside the running pod.

However, Artifact Keeper cannot download artifacts from remote repositories. Requests consistently return:

```text
502 Bad Gateway
```

The primary issue is not `NO_PROXY`. External remote repository traffic should go through the configured proxy, but downloads still fail with HTTP 502.

## Proxy configuration

The following environment variables are configured:

```text
HTTP_PROXY
HTTPS_PROXY
NO_PROXY
http_proxy
https_proxy
no_proxy
```

Example:

```yaml
env:
  - name: HTTP_PROXY
    value: "http://proxy.example.com:8080"

  - name: HTTPS_PROXY
    value: "http://proxy.example.com:8080"

  - name: NO_PROXY
    value: "localhost,127.0.0.1,.svc,.svc.cluster.local"

  - name: http_proxy
    value: "http://proxy.example.com:8080"

  - name: https_proxy
    value: "http://proxy.example.com:8080"

  - name: no_proxy
    value: "localhost,127.0.0.1,.svc,.svc.cluster.local"
```

The variables can be verified inside the backend pod:

```sh
env | grep -i proxy
```

## Expected behavior

The expected request flow is:

```text
Client
  -> Artifact Keeper
  -> configured HTTP/HTTPS proxy
  -> external upstream repository
```

Please verify:

- Whether Artifact Keeper supports `HTTP_PROXY` and `HTTPS_PROXY` for remote repository downloads.
- Whether both uppercase and lowercase proxy environment variables are supported.
- Whether proxy configuration must be applied to another Artifact Keeper component in addition to the backend.
- Whether the HTTP client used by Artifact Keeper automatically reads proxy environment variables.
- Whether proxy configuration must be explicitly defined in the application configuration.
- Whether proxy authentication is supported.
- Whether redirects from upstream repositories continue to use the configured proxy.
- Whether HTTPS connections through an HTTP `CONNECT` proxy are supported.
- Whether a custom corporate CA certificate can be configured for remote repository requests.
- Which component generates the final `502 Bad Gateway` response.

## Requested logging improvements

Please add diagnostic logging that shows:

- Requested upstream hostname.
- Whether direct or proxied routing was selected.
- Selected proxy hostname and port, without credentials.
- Whether the destination matched `NO_PROXY`.
- Proxy connection result.
- HTTP `CONNECT` result, where applicable.
- TLS certificate verification result.
- Upstream HTTP status code.
- Redirect target, where applicable.
- Internal exception that caused the final `502` response.

Passwords, tokens, proxy authorization headers, and repository credentials must not be logged.

## Debugging tools

The current backend image does not contain `curl` or `wget`. This makes it difficult to test connectivity from the same container and network context used by Artifact Keeper.

Please add `curl` or `wget` to the backend image. `curl` is preferred because it provides detailed proxy, HTTP, and TLS diagnostics.

Example diagnostic command:

```sh
curl -v --proxy "$HTTPS_PROXY" https://upstream-repository.example.com/
```

Direct connection comparison:

```sh
curl -v --noproxy '*' https://upstream-repository.example.com/
```

If adding diagnostic *tilities to the production image i* not acceptable, please provide on* of the following supported altern*tives:

- A separate debug variant*of the backend image.
- An optiona* debug container enabled through t*e Helm chart.
- Documentation for *ttaching an ephemeral debug contai*er.
- A built-in remote repository*connectivity test in the UI or API*

## Proxy configuration improveme*ts

In addition to fixing the `502* problem, please make proxy configuration easier in a future Helm chart or application version.

Suggested Helm configuration:

```yaml
proxy:
  enabled: true

  httpProxy: "http://proxy.example.com:8080"
  httpsProxy: "http://proxy.example.com:8080"

  noProxy:
    - "localhost"
    - "127.0.0.1"
    - ".svc"
    - ".svc.cluster.local"

  existingSecret: ""
  automaticallyExcludeInternalServices: true
```

The Helm chart currently generates resource names dynamically, for example:

```yaml
{{ include "artifact-keeper.fullname" . }}
```

If internal Artifact Keeper services must be excluded from the proxy, the chart should automatically add these generated names to `NO_PROXY`. Users should not have to manually duplicate the Helm naming logic.

A UI-based configuration would also be useful for:

- Global HTTP/HTTPS proxy configuration.
- Optional proxy configuration per remote repository.
- Proxy authentication.
- Custom corporate CA certificates.
- `NO_PROXY` entries.
- A **Test connection** action.
- Clear errors for proxy, DNS, TLS, authentication, and upstream failures.

These configuration improvements are secondary. The primary issue is that remote repository downloads currently always return `502 Bad Gateway` even though the outbound proxy is configured.

## Environment

```text
Platform: OpenShift
Deployment method: Helm chart
Component: Artifact Keeper backend
Feature: Remote repositories
Network access: External access only through HTTP/HTTPS proxy
Observed result: 502 Bad Gateway
```

## Acceptance criteria

- Artifact Keeper uses the configured outbound proxy for remote repository downloads.
- Artifacts can be downloaded and cached through the proxy.
- Remote repository requests no longer consistently return `502 Bad Gateway`.
- It is possible to determine which component returned a failed HTTP response.
- Proxy support is documented for all supported remote repository types.
- Logs distinguish proxy, DNS, TLS, authentication, and upstream failures.
- Logs do not expose passwords, tokens, or authorization headers.
- Custom corporate CA certificates can be configured.
- A supported method exists to test connectivity from the backend container context.
- The backend image contains `curl` or `wget`, or an officially supported debug alternative is provided.
- Proxy behavior is covered by an automated or documented integration test.

## Environment

- **Deployment type:** Helm chart on OpenShift
- **Operating System:**  container base OS
- **Database:** PostgreSQL
- **Reverse Proxy:** OpenShift Route
- **Network setup:** External outbound access is restricted. Artifact Keeper must connect to external remote repositories through a corporate HTTP/HTTPS proxy. The proxy variables are configured in the backend container, but remote repository downloads consistently return `502 Bad Gateway`.

### Backend Version

1.5.6

### Web UI Version

1.5.6

### Package Format

_No response_

### Expected Behavior

Artifact Keeper should:

- Use the configured outbound proxy when downloading artifacts from external remote repositories.
- Successfully download and cache artifacts from the upstream repository.
- Respect `NO_PROXY` only for excluded destinations.
- Return a meaningful error when a proxied request fails.
- Identify the component that generated the `502` response.
- Never expose proxy credentials or repository credentials in logs.

## Current behavior

1. Proxy environment variables are configured in the backend container.
2. The variables are visible inside the running pod.
3. An external upstream repository is configured as a remote repository.
4. An artifact that is not available in the local cache is requested.
5. Artifact Keeper returns `502 Bad Gateway`.
6. The artifact is not downloaded or cached.

It is currently unclear whether:

- Artifact Keeper uses the configured proxy for remote repository requests.
- The application HTTP client supports standard proxy environment variables.
- The proxy rejects the upstream request.
- The proxy requires additional authentication.
- The connection fails because of TLS certificate validation.
- The `502` response is generated by Artifact Keeper, an internal reverse proxy, or the outbound proxy.

## Steps to reproduce

1. Deploy Artifact Keeper on OpenShift using the Helm chart.
2. Configure `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` in the backend container.
3. Configure the lowercase equivalents as well.
4. Verify the variables inside the running backend pod:

   ```sh
   env | grep -i proxy
   ```

5. Configure an external remote repository.
6. Request an artifact that is not already present in the local cache.
7. Observe that Artifact Keeper returns:

   ```text
   502 Bad Gateway
   ```

8. Observe that the artifact is not downloaded from the upstream repository.

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

When an outbound egress proxy is configured (`HTTP_PROXY` / `HTTPS_PROXY` /
`ALL_PROXY`), remote-repository fetches failed with `502`/`500` in deployments
where the proxy is reached at a private/internal address (the common corporate
topology: a proxy on an RFC1918 or loopback address). The fail-closed SSRF DNS
guard (`SsrfGuardResolver`) resolves **every** outbound hostname — including the
proxy's own host — and drops private/internal addresses. Because the proxy is
resolved through the same guard, its legitimate private address was dropped and
the connection to the proxy never happened, so no upstream fetch could succeed.

This change adds a narrowly-scoped exemption: the resolver skips the SSRF filter
**only** for the exact configured egress-proxy host token(s), read once at
construction from the proxy environment. Everything else keeps the existing
fail-closed behavior byte-for-byte.

### What changed
- `ssrf_dns.rs`: `SsrfGuardResolver` gains a per-resolver `exempt_proxy_hosts`
  set, populated from a new `configured_proxy_hosts()` env reader that parses the
  host token out of the six proxy vars (upper/lower `HTTP_PROXY`/`HTTPS_PROXY`/
  `ALL_PROXY`), handling scheme, optional `user:pass@` userinfo and `:port`
  (e.g. `http://user:pass@proxy.corp:3128` → `proxy.corp`). In `resolve()`, a host
  that matches an exempt entry by **exact, case-insensitive, full-host equality**
  is resolved and returned unfiltered, *before* the mode-specific filter. A
  `with_mode()` constructor funnels all four trust-class constructors (upstream /
  internal / webhook / SSO), mirroring the existing trust-class pattern.
- Diagnostics: when the guard drops **all** resolved addresses for a host
  (previously an opaque connect failure), it now emits a `warn` on the `security`
  target with the host and mode.
- `proxy_service.rs`: a single shared transport-error classifier, reused across
  the buffered / streaming / token-exchange fetch paths, maps an SSRF-guard
  connect failure to a clear `503` instead of a bare `Storage`→`500`. The
  message is proxy-specific ("egress proxy connection refused by SSRF egress
  policy …") **only when an egress proxy is configured**; a direct / `NO_PROXY`
  upstream that is SSRF-blocked gets a generic message ("connection to upstream
  refused by SSRF egress policy …") with no mention of a proxy, so triage is not
  misled. Either way it is a `503` carrying only the redacted URL (never proxy
  creds/IP). Any other (non-SSRF) transport error keeps the existing mapping.
- No migration; no API/schema change.

### Scope of the exemption (security invariants, all covered by tests)
The exemption is keyed on the exact configured proxy **host token** only. It does
NOT:
- relax the guard for any other private target — a direct/`NO_PROXY` upstream at a
  private IP stays blocked even when a proxy is configured;
- exempt anything when no proxy env is set (exempt-set empty → resolver is
  byte-for-byte the pre-existing fail-closed guard);
- match subdomains/suffixes/prefixes (`proxy.corp.attacker.com`,
  `10.0.0.5.attacker.com` are not exempted);
- widen the loopback / link-local / cloud-metadata hard-blocks for the
  webhook / SSO / internal trust classes.

Added tests:
- `ssrf_dns.rs`: exact / case-insensitive / IP-literal match true; empty exempt-set
  always false; subdomain/suffix/prefix false; `proxy_host_token` parses
  scheme+creds+port+IPv6; `configured_proxy_hosts` reads all six env vars and is
  empty with none set; resolver with `proxy_hosts=['10.0.0.5']` resolves
  `10.0.0.5` in Upstream mode; empty exempt-set still blocks `10.0.0.5`
  (fail-closed regression guard); proxy-host exempt does not unblock a different
  private target.
- `http_client.rs`: a positive integration test (configured proxy **hostname** is
  reached through the guard) plus a **discriminating negative** (an unexempted
  proxy hostname is refused by the resolver before any TCP connect) — together
  these prove the resolver is on the proxy path and the exemption is load-bearing.
- `proxy_service.rs`: classifier units — SSRF-block-with-proxy → proxy-specific
  `503`; no-proxy SSRF-block → generic (non-proxy) `503`; generic upstream error
  → existing `Storage` path.

## Test Checklist
- [x] Manually tested locally

Manual verification used an isolated tinyproxy bound at a private (RFC1918)
address in front of the patched image: an uncached npm/maven/pypi remote fetch
returns `200` (previously `502`), with the proxy access log confirming the fetch
routed through the proxy; a direct `NO_PROXY` private-IP upstream and a private-IP
upstream with no proxy configured both stay blocked.

## API Changes
- [x] N/A - no API changes

---

Bundled into the v1.5.8 hotfix branch alongside #2457; the file sets are disjoint
(this PR touches `http_client.rs` / `ssrf_dns.rs` / `proxy_service.rs`).

Closes #2570

### DIFF
--- a/backend/src/services/http_client.rs
+++ b/backend/src/services/http_client.rs
@@ -652,9 +652,18 @@ mod tests {
             socket.write_all(b"K").await.expect("write second byte");
         });
 
-        let client = large_object_client_builder(true)
-            .build()
-            .expect("build large-object client");
+        // Serialize the client build against `test_configured_proxy_host_is_
+        // exempted_from_guard`, which briefly sets a process-wide `HTTP_PROXY`
+        // while building its own client. Without this, a concurrent build here
+        // could observe that proxy and route this round-trip away from the
+        // test's own listener. The guard is dropped before any `.await`, so it
+        // never crosses an await point.
+        let client = {
+            let _proxy_env_guard = PROXY_ENV_LOCK.lock().unwrap();
+            large_object_client_builder(true)
+                .build()
+                .expect("build large-object client")
+        };
         let request = tokio::spawn(async move {
             client
                 .post(&url)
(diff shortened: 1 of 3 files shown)


brandonrc:
Closing: superseded. The egress-proxy allowance is already on main (via #2581/#2582) and shipped in v1.5.8.

For the record, the red here was NOT the documented DNS or sccache flake: Check Rust and cargo test both pass, and the failure is \`release-branch-gate.yml\` firing correctly against a branch whose content already landed by another route.

Issue #2504, opened 2026-07-14
[SECURITY] Cross-tenant read+write on cloud (S3/GCS/Azure) storage: flat namespace + unguarded Maven/sbt legacy fallback


brandonrc:
**Severity: HIGH (potentially CRITICAL — proven cross-tenant READ and WRITE). Pre-existing; surfaced during #1945/PR #2501 red-team.**

## Summary
On non-filesystem storage backends, repositories share one flat object namespace with no per-repo key prefix, and Maven's legacy hosted fallback fetches by an unguarded flat key. A user scoped to repo A can read AND overwrite artifacts in private repo B (different tenant) on S3/GCS/Azure deployments. Filesystem backends are NOT affected (per-repo directories isolate them).

## Root cause
- `backend/src/storage/registry.rs` `StorageRegistry::backend_for` returns one shared backend instance for cloud backends; object keys are flat `{format}/{path}` with no repo scoping.
- `backend/src/api/handlers/maven.rs` (~1677-1693) legacy hosted fallback: when the repo-scoped artifact-row lookup MISSES, it does `storage.get_stream(format!("maven/{}", path))` with no cross-repo guard, resolving another repo's object in the shared namespace.

## Proof (rig, S3/MinIO backend)
- Repos `mvn-a` (tenant A) and `mvn-b` (tenant B, private). User `alice` scoped to `mvn-a` only.
- **Cross-repo READ:** `GET /maven/mvn-a/com/secret/lib/1.0/lib-1.0.jar` (coordinate exists only in mvn-b) → 200, returns mvn-b's confidential bytes.
- **Cross-repo WRITE/poison:** `PUT /maven/mvn-a/com/secret/lib/1.0/lib-1.0.jar` → 201; mvn-b's jar subsequently serves alice's bytes.
- Served via the legacy streaming fallback (NOT the #2501 redirect path — the redirect path is correctly repo-scoped/confined).

## Scope to confirm (blocking triage)
1. Is this the DEFAULT production behavior for cloud backends, or a rig-config artifact?
2. Which shipped versions (1.5.x?) are affected → **security hotfix candidate** if live.
3. All affected formats (Maven/sbt confirmed; check other flat-key formats).

## Fix direction
- Repo-scoped object keys (include repo id/key in the object prefix) for cloud backends, and/or
- Drop/repo-guard the unguarded legacy hosted fallback for cloud backends.

Origin: PR #2501 adversarial validation.

PR #2578, opened 2026-07-15
fix(maven): ownership-checked flat-key reads on shared cloud namespaces (#2574)

### LINKED ISSUE #2574: 1.5.7 regression on cloud backends: legacy GAV-secondary files without their own catalog row now 404 (read gate from #2504)
## Summary

The v1.5.7 cross-repository isolation fix (#2504 / GHSA-g6ph-ph6w-w6vw) gates the flat-key `maven/{path}` read fallbacks to repo-isolated (filesystem) backends. On cloud backends (S3/GCS/Azure) this **breaks serving of pre-existing artifacts that have no catalog row of their own** — most importantly the secondary files of Maven GAVs uploaded by older releases (`.pom`, `.module`, `-sources.jar`, etc.), which now return **404** after the upgrade. The release notes state "No breaking changes", but any instance with legacy data on a cloud backend loses read access to this class of files.

## Environment

- Backend: S3 storage backend
- Upgrade path: 1.4.1 → 1.5.7 (data on the instance dates back to ~1.2.x)

## Root cause

Older releases' `/maven/{repo}/{path}` PUT created **one catalog row per GAV** — the first file uploaded for a GAV got an `artifacts` row; subsequent sibling files (`.pom`, `.module`, sources/javadoc jars) were stored at their flat `maven/{path}` object keys **without a row of their own**. Reads for those row-less paths were served by the direct-byte flat-key fallback in `serve_artifact`.

v1.5.7 gates that fallback (and the checksum-sidecar / metadata fallbacks) behind `backend_is_repo_isolated(...)`, so on S3 the request falls through to strictly row-scoped resolution — and 404s, because no row exists.

Current releases create a row per file (verified on 1.5.7: uploading `.jar` then `.pom` produces two rows, both serve fine), so **newly published artifacts are unaffected**. Only legacy data breaks.

## Reproduction

1. On a version whose uploader still grouped rows per GAV (observed with 1.2.x-era data), publish a GAV via `PUT /maven/{repo}/com/example/lib/1.0.0/lib-1.0.0.jar` followed by `lib-1.0.0.pom`.
2. Observe the catalog: the artifacts listing shows both files sharing **one** row id (or the `.pom` absent entirely).
3. `GET /maven/{repo}/com/example/lib/1.0.0/lib-1.0.0.pom` → 200 on ≤ 1.5.6.
4. Upgrade to 1.5.7 (S3 backend).
5. Same GET → **404**. The `.jar` (which owns the row) still serves 200.

Impact in practice: Gradle module metadata (`.module`) and POMs of every GAV published during the legacy-uploader era 404 → dependency resolution fails for consumers after the registry upgrade.

## Suggested fix

The write side of #2504 already consults the catalog by `storage_key` (backed by the new `idx_artifacts_storage_key` index from migration 154) to decide ownership. The read fallback could do the same instead of being skipped wholesale on shared namespaces:

- if the flat key's `storage_key` is referenced by a row in a **different** repository (live or soft-deleted) → 404 (the security property of #2504 is preserved);
- if it is referenced only by the **requesting** repository, or by **no repository at all** → serve the bytes as before.

Since the legacy GAV-grouped rows live in the requesting repository (the sibling primary references the same GAV directory), this restores legacy reads without reopening the cross-tenant read hole. An alternative (or complement) would be a startup backfill that materializes per-path rows for row-less GAV-secondary objects, but the ownership-checked fallback is the smaller change and also covers non-Maven flat-key classes (checksum sidecars, `maven-metadata.xml`).

Happy to provide more detail from our instance if useful.

ThaSami:
Fixes #2574.

## Problem

The #2504 security fix gated every unanchored `maven/{path}` storage read to repo-isolated (filesystem) backends. On shared cloud namespaces (S3/GCS/Azure) that broke serving of pre-existing data:

- companion files of GAVs uploaded before Maven uploads indexed every physical asset (`.pom`, `.module`, `-sources.jar`, …) return **404** after upgrading to 1.5.7 — Gradle/Maven dependency resolution fails for consumers;
- stored checksum sidecars stop serving for row-less files;
- deliberately-uploaded `maven-metadata.xml` silently degrades to the synthesized document (a content change, not a 404 — see the second comment on #2574), which can advertise a different SNAPSHOT build than the stored file.

## Approach

Apply the **same catalog ownership rule the #2504 write guard already uses**, on the read side — backed by the same `idx_artifacts_storage_key` index (migration 154, already shipped):

- a flat key referenced by a **different** repository (live *or* soft-deleted — the physical object outlives a soft-delete) stays unreadable → the cross-tenant read hole from GHSA-g6ph-ph6w-w6vw **remains closed**;
- a key that is **unowned** or owned only by the **requesting** repository serves exactly as it did before v1.5.7;
- database errors **fail closed** (skip the fallback / 404);
- repo-isolated (filesystem) backends short-circuit to "readable" with no query — zero behavior or cost change there.

## Changes

- `services/artifact_service.rs`: `flat_key_readable_for_repo()` — read twin of `guard_foreign_storage_key` (same foreign-owner query), plus 4 tests mirroring the existing #2504 guard tests (unowned / own-repo / foreign-live / foreign-soft-deleted).
- `handlers/proxy_helpers.rs`: `flat_key_readable()` — isolation short-circuit + fail-closed wrapper, read twin of `guard_cross_repo_write`.
- `handlers/maven.rs`: the four gated read sites — stored checksum sidecars (plain + snapshot-resolved), virtual-member plugin-prefix and snapshot metadata reads, hosted stored `maven-metadata.xml`, and the legacy hosted direct-byte fallback.
- `handlers/maven_proxy.rs`: Gate 0 of `maven_local_fetch_storage_fallback` now runs the ownership check instead of rejecting cloud backends wholesale; Gate 2's live-sibling anchor in the *requesting* repository still applies on top.
- `CHANGELOG.md`: Unreleased → Fixed entry.

## Cost note

On cloud backends each restored fallback adds one indexed point lookup (`storage_key = $1 AND repository_id <> $2 LIMIT 1`) — and only on paths that already missed the row-scoped lookup (the 404 path today). Hot row-backed reads are untouched.

## Verification

- Ownership-rule unit tests added next to the existing #2504 write-guard tests (same `test_db_helpers` harness, all four polarity cases).
- Verified against a production-scale 1.4.1→1.5.7 dataset exhibiting the regression: row-less GAV companions (`.pom`/`.module`) that 404 on 1.5.7 are readable under this rule, while cross-repo colliding keys remain blocked (the write-guard tests' scenarios, exercised read-side).
- `rustfmt` clean; relying on CI for the full check/clippy/test matrix (a corporate egress proxy here blocks a handful of crates.io downloads, so I could not run the full local build — apologies in advance for any CI nit).


ThaSami:
@brandonrc a fix to unblock this

brandonrc:
Thank you for this @ThaSami — both the precise #2574 diagnosis and a working patch with tests. We ran it through an adversarial security review on a real S3 backend before merging, and unfortunately the catalog-ownership approach can't safely serve the class of files it re-enables, so we can't merge it as-is. Sharing the reasoning since it shapes the fix:

The ownership check (`flat_key_readable_for_repo`) refuses a flat key only when **another repository has a catalog row** for it. But the legacy files this restores — row-less GAV companions (`.pom`, `.module`, `-sources.jar`), verbatim `maven-metadata.xml`, checksum sidecars — have **no row at all**. With no owning row, the "no foreign owner" test passes for *every* repository, so on a shared cloud namespace those objects become readable cross-repository — which reopens the isolation guarantee #2504 closed. The difficulty is structural: a row-less object on a flat namespace carries no inherent repository attribution, so a catalog lookup alone can't tell "my legacy file" from "another tenant's."

The good news is the **call-site wiring here is exactly right** (gating the checksum sidecar, SNAPSHOT resolved checksum, virtual-member fanout, the verbatim metadata read, the `serve_artifact` fallback, and `maven_proxy` Gate 0), and we're building the fix on top of it — replacing the gate's *logic* with a stronger attribution mechanism so a legacy file serves only to its genuine owner, robust against a repo forging attribution for someone else's coordinate. We're targeting this for **v1.5.8** and will credit your report and this PR. I'll update here when the follow-up lands.

brandonrc:
Superseded by #2585, shipped in **v1.5.8** (GHSA-xqg8-382p-vjrp). Thank you again @ThaSami — your #2574 diagnosis and the call-site wiring in this PR directly shaped the shipped fix; the final version swapped the catalog-ownership gate for a pristine-state attribution table (so row-less objects can't be read/claimed cross-tenant) and added the write-side guard. Closing in favor of the released fix.

PR #2602, opened 2026-07-16
feat(rpm): curated snapshot versions with signed immutable @N publication serving (#2358)

### LINKED ISSUE #2358: Phase 3: RPM mirror versions, publications, snapshots
Beads `artifact-keeper-1gx.3`. **Depends on P2.** Epic `artifact-keeper-1gx`.

**Scope:** `repository_versions`(+members), `repo_metadata_files`, `publications`, `RpmPublishService` + `SigningService` (AK-signed), `active_publication_id`, `/@{version}/` snapshot serving.

**Curate track:** regenerate primary+filelists+other from the approved set (reuse hosted generation). Trust-model doc (C2): `repo_gpgcheck` → AK key; package `gpgcheck` → Rocky key.

**Done =** pin `/rpm/epel/@N/`, identical resolvable content later.

Design: §7 P3.

brandonrc:
## Summary

RPM curation Phase-3 (#2358) freezes an approved RPM curation set into a
monotonic, immutable `repository_version`, publishes it as **AK-signed**
repodata, and serves it under `/rpm/{key}/@N/`. A published snapshot is
tamper-evident and immutable: its metadata is a frozen blob (never regenerated
from the live approved set), re-publishing it is refused, and its
`repomd.xml.asc` is a **real detached OpenPGP signature** that `dnf`
(`repo_gpgcheck=1`) / `rpm --import` can verify against the served
`repomd.xml.key`.

Supply-chain hardening is preserved throughout: nothing attacker-influenced from
upstream is ever re-emitted as markup. Each upstream `<package>` is captured at
sync as a **typed, validated struct** (entity expansion off, per-package byte
cap, fail-closed on malformed/oversize/missing-field), and `primary.xml` is
**canonically re-serialized from the struct** at publish — every text node and
attribute escaped, exactly one `<metadata>`/`</metadata>` pair by construction,
and every `<location>` rebuilt purely from validated NEVRA (an absolute URL or
`..` is impossible). Package bytes served under `@N` are fetched from the
curation-config upstream and **sha256-verified against the frozen checksum**,
failing closed (404/502) on any mismatch.

## What changed

1. **Structured capture at sync** (`curation_sync.rs`): each `<package>` parses
   into a typed `RpmPackageMetadata` (NEVRA, summary/description, checksum,
   size, time, full `format` incl. provides/requires/files). Malformed,
   oversize (>256 KiB), or field-missing packages are dropped. The upstream
   `<location href>` is deliberately not captured into the trusted metadata.

2. **Migration** `184_rpm_repository_versions.sql`: `curation_packages` gains
   `primary_metadata JSONB`; new `repository_versions` (+ publication columns)
   and `repository_version_packages` (carrying the **frozen package identity** —
   `frozen_checksum_sha256`, `frozen_upstream_path`, `frozen_filename`, all NOT
   NULL, plus a `(version_id, frozen_filename)` lookup index);
   `repositories.active_publication_id`. Reversible (drops documented in-file).

   **Why the identity is frozen:** `curation_packages` is a *live* table — the
   routine sync upserts the same row (`ON CONFLICT DO UPDATE`). If the `@N` serve
   path resolved a package's checksum/location from that row, a post-publish
   re-sync (or a compromised upstream) could change what an already-published,
   *signed* snapshot serves, handing clients bytes that contradict `@N`'s own
   signed `primary.xml`. `create_version` therefore snapshots the identity from
   the **same structured metadata the signed `primary.xml` is generated from**,
   so the frozen checksum is exactly what `@N` attests; `publish` fails closed
   (409) if a member drifted between snapshot and publish, keeping
   "signed metadata == frozen identity" an enforced invariant.

3. **Signed, immutable publish** (`rpm_publish_service.rs`): `create_version`
   snapshots the approved set under a serializable version allocation;
   `publish` canonically re-serializes `primary.xml`, builds pkgid-consistent
   stub `filelists`/`other`, signs `repomd.xml` with a **detached OpenPGP
   signature**, and stores every blob (signature + public key as-of publish
   time) under an immutable per-version prefix. Re-publishing a published
   version is refused (409); a repo whose active key cannot produce OpenPGP
   (`key_type != gpg`) is refused (fail-closed) rather than emitting an
   unverifiable signature.

4. **`@N` serving** (`rpm.rs`): the `@N` metadata routes serve the frozen
   AK-signed blobs; a `@N/packages/{nevra}.rpm` request resolves the member from
   the **frozen** membership columns (never the live `curation_packages` row),
   fetches from the curation-config upstream using the **frozen** relative path,
   verifies sha256 against the **frozen** checksum, caches into the immutable
   store, and streams — fail-closed on mismatch or an unsafe upstream path. Both
   the fetch and cache-hit paths advertise the frozen `X-Checksum-SHA256`, and
   package requests are routed ahead of the generic stored-blob serve so a cached
   `.rpm` keeps its `application/x-rpm` response contract. This also rejects an
   upstream mirror that silently changes the bytes behind an already-published
   NEVRA. New endpoints: `POST /curation/repos/{key}/versions` and
   `.../versions/{n}/publish` (dual authz: curation capability +
   tenant-ownership).

- [x] Adds tests that would have caught the bug

The publish signing path is anchored by unit/DB tests that exercise the real
signature end-to-end:
- `test_published_at_n_signature_verifies_and_rejects_tamper_db` — the frozen
  `repomd.xml.asc` verifies over the frozen `repomd.xml` under the frozen
  `repomd.xml.key`, and a byte-tampered repomd fails verification.
- `test_publish_rejects_non_openpgp_key_db` — a non-OpenPGP active key is
  refused at publish rather than emitting an unverifiable armor blob.
- `test_publish_stores_signed_repodata_and_sets_active_db` — full publish +
  immutability (re-publish → 409).
- `test_create_version_freezes_package_identity_db` — the snapshot's frozen
  checksum/path/filename survive a simulated re-sync of the live curation row.
- `test_publish_rejects_drift_since_version_created_db` — a member that drifted
  between snapshot and publish fails the publish closed (409).

End-to-end, an `rpm-p3-curation` deployment-test tier drives
create → publish → serve → poisoned-re-sync and asserts `@N` never serves the
mutated bytes (it serves the frozen bytes with the frozen `X-Checksum-SHA256`).

## Test Checklist
- [x] Integration tests added/updated (DB-backed publish/serve)
- [x] Manually tested locally (isolated deploy; create→publish→serve→immutability→gpg-verify)

## API Changes
- [x] OpenAPI spec validates
- [x] Migration is reversible
- [ ] N/A - no API changes

## Related

Closes #2358


brandonrc:
Supply-chain red-team (live deploy of head `00367514`) — **signature verified REAL, but one HIGH holds the merge.**

**Confirmed clean:** the OpenPGP signing fix is genuine (independent `gpg --verify` = Good signature, real v4 packet + CRC24; one-byte tamper → BAD); fail-closed on no-key (400, @N stays 404) and on a non-OpenPGP key (refused at config *and* publish); re-publish → 409; `@N` metadata byte-frozen and key-rotation-safe (serves the frozen public key, still verifies after rotation); dual authz (trigger:sync capability + tenant-ownership) enforced; NEVRA `<location>` canonicalized, absolute-URL/`../` upstream paths refused (no SSRF/traversal).

**Finding (HIGH) — `@N` package payloads are not actually frozen.** `repository_version_packages` (mig 183) stores only `(version_id, curation_package_id)` — no frozen checksum or location. `serve_version_package` (rpm.rs:432-539) resolves `checksum_sha256` + `upstream_path` from the **live** `curation_packages` row, which a routine re-sync (`curation_service.rs` ON CONFLICT DO UPDATE) overwrites in place. Live repro: publish `@1` (signed metadata attests checksum A for pkg); mutate the live cp row to checksum B + evil upstream bytes; GET `@1/packages/<pkg>.rpm` → **200 serving the evil bytes**, and `X-Checksum-SHA256` even advertised the evil hash — bytes that contradict `@1`'s own signed `primary.xml`. A compliant dnf rejects the mismatch (signed metadata is the trust anchor), so client impact is bounded, but the server breaks its stated `@N`-immutable contract, vouches for tampered bytes, and poisons mirroring/proxy tools that trust `@N`.

**Fix:** freeze `checksum_sha256` (+ resolved location/storage_key) into `repository_version_packages` at publish, and have `serve_version_package` verify fetched/cached bytes against the FROZEN checksum (reject on mismatch — this also defends against upstream-mirror poisoning of an already-published NEVRA), advertising the frozen hash. Holding draft until this lands + a re-red-team.

Issue #2580, opened 2026-07-15
RPM repository returns 404


Samekraft:
### Description

RPM repository returns the correct repository, but returns 404 for the package

### Environment Information

- Deployment type: official docker compose
- Operating System: Fedora Server
- Architecture: x86_64
- Database: PostgreSQL
- Reverse Proxy: Caddy
- Network setup: Just port forward

### Backend Version

1.5.7

### Web UI Version

1.5.7

### Package Format

RPM

### Expected Behavior

Installing a package from the RPM repository

### Observed Behavior

Artifact Keeper returned a 404 on .rpm``

### Steps to Reproduce

1. create repo
```
ak repo create test-rpm --pkg-format rpm --repo-type local
Created repository 'test-rpm' (format: rpm, type: local)
```

2.    push rpm
```
ak artifact push test-rpm ./cosign-3.1.1-1.x86_64.rpm 
Uploaded cosign-3.1.1-1.x86_64.rpm [########################################] 54.49 MiB/54.49 MiB (37.03 MiB/s, 0s)                                                       cosign-3.1.1-1.x86_64.rpm (54.5 MB) -> test-rpm:cosign-3.1.1-1.x86_64.rpm
```

3.   setup repo
```
vi /etc/yum.repos.d/test-rpm.repo
[test-rpm]
name=Artifact Keeper - test-rpm
baseurl=http://192.168.50.2:30080/rpm/test-rpm/
enabled=1
gpgcheck=0
```

4.    check repodata
```
dnf info cosign
Updating and loading repositories:
 Artifact Keeper - test-rpm             100% |  57.2 KiB/s |   2.2 KiB |  00m00s
Repositories loaded.
Available packages
Name           : cosign
Epoch          : 0
Version        : 3.1.1
Release        : 1
Architecture   : x86_64
Download size  : 54.5 MiB
Installed size : 0.0   B
Source         : 
Repository     : test-rpm
Summary        : 
URL            : 
License        : 
Description    : 
Vendor         : <NULL>
```

5.    try install 
```
sudo dnf install cosign                 
Updating and loading repositories:
 Artifact Keeper - test-rpm                                                                                                     100% |  77.6 KiB/s |   2.2 KiB |  00m00s
Repositories loaded.
Package                                                  Arch          Version                                                  Repository                          Size
Installing:
 cosign                                                  x86_64        3.1.1-1                                                  test-rpm                         0.0   B

Transaction Summary:
 Installing:         1 package

[...]

PR #2634, opened 2026-07-16
fix(alpine): emit an apk-consumable APKINDEX (C: checksum, tar typeflag, I: size)

### LINKED ISSUE #2633: Alpine APKINDEX is not consumable by a real apk client (C: checksum, tar typeflag, I: size)
### Summary

A real `apk` client cannot install any package from a hosted Alpine repository. The
generated `APKINDEX.tar.gz` is malformed for apk-tools in three independent ways, so
`apk add` (even with `--allow-untrusted`) fails on every hosted alpine repo.

This is the advertised-metadata-vs-real-client class (#2580): existing tests only assert
the index downloads and decompresses; they never run a real `apk add`, so this was missed.

### Root cause (`backend/src/api/handlers/alpine.rs`)

1. **`C:` checksum format.** `generate_apkindex_text()` wrote the per-package checksum as a
   bare hex SHA256 (`C:{checksum_sha256}`). apk-tools requires the apk-native form
   `Q1` + base64(SHA1 of the package's control segment), e.g.
   `C:Q1O3f05G1QIlK5QBrIDIGE0gDWKs4=`. An unparseable `C:` makes apk-tools reject the whole
   index with `BAD signature`.

2. **Index tar entry typeflag.** `create_apkindex_tar_gz()` built the `APKINDEX` tar entry
   with a NUL typeflag (the default for a bare GNU header) instead of a regular-file (`'0'`)
   typeflag. apk-tools rejects the index (`BAD signature`) even once `C:` is correct.

3. **Installed size (`I:`) not populated.** `I:` was hard-coded to `0` because `.PKGINFO`
   was never parsed. apk-tools extracts **zero files** (installs an empty package) when the
   index advertises `I:0`, so even after (1) and (2) the marker file never lands on disk.

### Impact

Hosted Alpine repositories are unusable by a real `apk` client. High for the Alpine format.

### Proof

Rebuilding the same `.apk` with `apk index` (which emits `C:Q1...`, a regular-file typeflag,
and a non-zero `I:`) installs cleanly, while the AK-served index is rejected. Fixing all
three in the backend makes `apk update`/`apk add --allow-untrusted` publish→consume install
the package (verified against apk-tools 2.14.6, arm64).

### Fix direction

In `generate_apkindex_text`/`create_apkindex_tar_gz`/`store_apk`: emit the `C:` value as
`Q1`+base64(SHA1(control segment)); mark index tar entries as regular files; parse `.PKGINFO`
at upload to populate `I:` (installed size) plus description/url/license/depends. Repositories
served unsigned remain consumable with `apk add --allow-untrusted`; a signed, trusted index is
a separate follow-up.

brandonrc:
## Summary

A real `apk` client could not install **any** package from a hosted Alpine repository:
`apk add` failed on every hosted alpine repo with `BAD signature`, even with
`--allow-untrusted`. The generated `APKINDEX.tar.gz` was malformed for apk-tools in three
independent ways, all fixed here in `backend/src/api/handlers/alpine.rs`:

1. **`C:` checksum format.** The index carried a bare hex SHA256 of the `.apk`
   (`C:{checksum_sha256}`). apk-tools expects its native form — `Q1` + base64(SHA1) taken
   over the package's **control segment** — and rejects the entire index when it cannot
   parse the value, which is why one bad field made every package unavailable. The
   checksum is now computed from the uploaded package and recorded at upload time.
2. **Index tar entry typeflag.** The `APKINDEX` entry was built from a bare GNU header,
   which leaves a NUL typeflag. apk-tools only accepts index entries marked as regular
   files (`'0'`), so it rejected the index even once `C:` was correct.
3. **`I:` installed size.** Hard-coded to `0` because `.PKGINFO` was never parsed. `I:` is
   now read from the package, along with description/url/license/depends/provides and the
   origin/maintainer/build-time/commit fields a real index carries.

Two follow-on notes:

- Packages stored before the checksum was recorded are **backfilled** from their stored
  bytes on the next index request, and the result is persisted. Without this, existing
  repositories would stay unusable until every package was re-uploaded. Only the head of
  each package is read, at most once per artifact per parser version, and the number of
  packages read per request is capped — see *Bounding the backfill* below.
- An entry with no usable checksum is now **left out** of the index rather than emitted:
  apk-tools rejects an index it cannot key, so emitting one would break the repository for
  every other package. Uploads are still **not** gated on parsing — an unreadable package
  is stored and logged, it just cannot be indexed.

**Why skip rather than reject the upload.** apk v3 is coming, and its packages are not
readable by this (v2) parser. Rejecting unparseable bytes at upload would hard-fail every
v3 upload the moment the format lands, which is a much worse failure than not indexing
them: the registry would refuse content it is supposed to store. Skipping keeps the upload
working and leaves the artifact recoverable — the failure marker is parser-versioned, so
when a v3-capable parser does land it bumps the version and every package skipped by the
older parser is backfilled and indexed retroactively, with no re-upload and no operator
action. Skip is the reversible choice; reject is not.

Repositories served unsigned remain consumable with `apk add --allow-untrusted`; a signed,
trusted index is a separate follow-up.

### Bounding the backfill

The backfill is driven by an endpoint that is **anonymous on public repositories**, so its
cost per request is part of the contract:

- **A verdict is always recorded.** Previously only the success path persisted, so an
  artifact this parser cannot read stayed checksumless and was re-fetched and re-parsed on
  *every* subsequent index request, forever. One unreadable package — uploaded deliberately,
  or merely corrupt/truncated — made every later `apk update` pay a full read of it. The
  backfill now records a parser-versioned `apk_parse_failed` marker and gates on it, so a
  package's bytes are read **at most once per parser version**.
- **Transient failures are not verdicts.** A storage error says nothing about the package,
  so it is *not* marked: a blip must not permanently de-index a good package. Only a
  deterministic parse failure — bytes successfully read, not a readable package — is
  recorded.
- **Only the head is read.** `storage.get` pulls a whole object into memory and
  `MAX_UPLOAD_SIZE` defaults to 10 GB. An apk v2 package is `[signature] control data`, so
  the index metadata is at the front: the backfill now uses a ranged read for the first
  1 MB, clamped to the object's own size, and never fetches the data segment.
- **Work per request is capped**, so a large repository converges over successive requests
  rather than turning one index GET into an unbounded sequential fetch. Persistence is
  per-artifact, so a cancelled request keeps its progress.

The marker is queryable (`WHERE metadata ? 'apk_parse_failed'`), which also answers the
"uploaded 201 but silently absent from the index" observability question a `warn!` alone
does not.

### `P:`/`V:` index injection

Path parameters are percent-decoded before the handler sees them, so
`PUT .../evil%0AC%3AQ1...-1.0-r0.apk` produced a package **name containing a newline**,
written straight into `P:`. APKINDEX has no escaping, so the name ended its own line and
the remainder was read as further index fields — letting any writer forge entries for
packages they do not own. Reproduced on this branch before the fix; the served index was:

```
C:Q11jnfYL890CWXj/b2WxBdoVucrs4=
P:evil
C:Q1AAAAAAAAAAAAAAAAAAAAAAAAAAA=    <- attacker-chosen
P:openssl                            <- forged entry (duplicate keys: last wins)
V:1.0-r0
```

This is **pre-existing**, but it was unreachable while the index was unparseable — the
`C:` fix in this PR is what makes it live, so it is fixed here rather than deferred.
Name and version are now charset-validated (`[A-Za-z0-9._+-]`) and a violating upload is
rejected with a 400. The same rule filters entries at index generation, covering rows
stored before the gate and the metadata-supplied values, which do not pass through it.
The charset was checked against reality rather than picked: across all ~25,000 packages in
Alpine v3.21 `main` and `community`, **no** name or version uses a character outside it.

### Other review follow-ups

- **`I:` fail-safe.** `installed_size` previously defaulted to `0`, so a package with a
  valid checksum but no recorded size would be advertised as empty and apk would install
  it as an empty package — silent corruption. A missing installed size is now treated like
  a missing checksum: skip the entry.
- **Atomic metadata merge.** The backfill's write was a read-modify-write of a snapshot
  taken *before* the package fetch, so a concurrent upload's fields could be dropped. It
  now merges server-side (`metadata = artifact_metadata.metadata || $2`) and sends only the
  new fields.

### Verifying against apk-tools rather than guessing

The `C:` input is the part worth reviewing carefully, so it was pinned to apk-tools' own
output rather than to a reading of the spec. A marker `.apk` was built with `abuild -F`
inside `alpine:3.21` (apk-tools 2.14.6, aarch64) and indexed with real
`apk index -o APKINDEX.tar.gz`, which emits:

```
C:Q11jnfYL890CWXj/b2WxBdoVucrs4=
I:22
```

Hashing each gzip member of that same `.apk` shows the digest input is unambiguous — it is
the **raw compressed bytes of the control segment** (gzip header and trailer included), not
the inflated tar and not the data segment:

```
segment 0 [0:641]     .SIGN.RSA...     sha1=231fdcab...
segment 1 [641:1069]  .PKGINFO         sha1=d639df60bf3dd025978ff6f65b105da15b9caece  <= base64 => 1jnfYL890CWXj/b2WxBdoVucrs4=
segment 2 [1069:1440] usr/share/...    sha1=60c05a66...
```

The same rule was then cross-checked against real Alpine packages (`busybox`, `musl`,
`zlib`, `ca-certificates`) and reproduces `apk index`'s `C:` for all of them. The `.apk`
fixture and the apk-tools-produced values are committed, so the tests assert against
apk-tools' answer, not ours.

The bug survived because the existing tests only assert that the index downloads and
decompresses — they never assert its *contents* against a real client (the
advertised-metadata-vs-real-client class, #2580). Three pre-existing tests actually
asserted the broken behaviour (`assert!(text.contains("C:abc123"))`) and are updated to the
apk-native contract.

Added (all in the lib crate's `#[cfg(test)]` modules, none `#[ignore]`d):

- `test_apk_checksum_matches_apk_tools_output` — the computed `C:` equals the value real
  `apk index` produced for the committed fixture.
- `test_apk_checksum_has_q1_base64_shape` — matches `Q1[A-Za-z0-9+/]+=*`.
- `test_apk_checksum_is_sha1_of_control_segment_not_sha256_hex` — guards the original bug.
- `test_apk_checksum_covers_control_segment_only` — mutating the data segment must not move
  the checksum.
- `test_apkindex_tar_entry_is_a_regular_file` / `..._signature_tar_entry_...` — assert the
  typeflag byte is `'0'`.
- `test_parse_apk_package_reads_installed_size` — `I:` is the real, non-zero size.
- `.PKGINFO` parsing, gzip member-boundary, and malformed/non-apk input cases.
- `test_publish_apk_indexes_with_apk_tools_checksum` — publish → index round trip over real
  DB + storage.
- `test_index_backfills_apk_checksum_for_legacy_artifact` — a pre-fix artifact is
  backfilled and persisted.

Backfill cost and injection (all fail on the code without the corresponding fix — verified
by reverting each fix in place and re-running):

- `test_unparsable_package_is_read_once_and_not_reread` — the re-read is *observed*, not
  counted: after the first index request the stored bytes are replaced with a package that
  does parse, so a backfill that re-read the artifact would index it. Without the marker
  the swapped bytes are picked up (proving the re-read); with it they are not.
- `test_transient_storage_failure_does_not_permanently_mark_package` — storage is made to
  fail, and the artifact must come back on its own once storage recovers, with no marker
  recorded in between.
- `test_backfill_reads_only_the_head_of_a_package` / `..._is_capped_and_never_exceeds_object`
  — a recording backend asserts the exact ranges read and that the whole-object `get` is
  never called.
- `test_backfill_reports_storage_failure_as_unavailable_not_unreadable` — the
  transient-vs-deterministic split at the unit level.
- `test_needs_backfill_*` — the gate, including that a marker from an older parser version
  is retried and one from a newer version is honoured.
- `test_upload_rejects_apkindex_field_injection_in_filename` — the percent-decoded
  injection is rejected with a 400 and never reaches the index.
- `test_parse_apk_filename_rejects_newline_in_{name,version}`,
  `..._with_multiple_version_boundaries` — including that a bad character cannot be escaped
  by letting the parser try a later version boundary (the name is always a prefix of the
  stem).
- `test_generate_apkindex_text_skips_entry_with_injected_name` /
  `..._with_no_installed_size` — the generation-time guards.
- `test_apk_name_charset_accepts_real_alpine_names_and_versions` — the charset does not
  reject real packages.

### End-to-end proof with a real apk client

Verified on the format-conformance tier that found this, which publishes an `abuild`-built
package and runs a real `apk add` against the hosted repo (`FC_ONLY=apk`, arm64):

| backend image | result |
| --- | --- |
| unpatched (`main` @ e8dd224) | **FAIL** — `BAD signature`; `apk add` cannot resolve the package (3 passed, 5 failed) |
| this branch | **PASS** — 8 passed, 0 failed |

On this branch a real apk client parses the AK-served index and installs the package:

```
fetch http://.../v3.21/main/aarch64/APKINDEX.tar.gz
OK: 64 distinct packages available
(1/1) Installing dtf-marker (1.0-r0)
OK: 251 MiB in 64 packages
```

The upgrade path (`index_regen_on_upgrade`) and `arch_isolation` cases pass too:
`(1/1) Upgrading dtf-marker (1.0-r0 -> 1.1-r0)`. The tier's plugin is currently registered
`FC_ENABLED: 0` (KNOWN-RED) in artifact-keeper-test; flipping it to `1` is a follow-up in
that repo once this lands. Re-verified on this branch's rebuilt image with the plugin
flipped to `FC_ENABLED: 1` locally (reverted afterwards): **8 passed, 0 failed**, real
`apk add` green.

One note for whoever does that flip: the plugin's client container is a bare `alpine` image
that installs `bash` during provisioning, and the harness sometimes `exec`s before that
finishes — `exec: "bash": executable file not found in $PATH`, which fails `fc_publish` in
0s before any backend call. It reproduced intermittently on identical images (2 of 5 runs)
and is independent of this change, but the plugin will need a readiness wait before it can
be flipped green in CI.

## Test Checklist
- [x] E2E tests added/updated (if applicable) — verified on the format-conformance tier; the
      plugin lives in artifact-keeper-test and its `FC_ENABLED` flip is a follow-up there
- [x] Manually tested locally
- [x] No regressions in existing tests — full backend lib suite green (13440 passed)

## API Changes
- [x] N/A - no API changes

No migration: the new fields go into the existing `artifact_metadata.metadata` JSON.

## Gates

`cargo fmt --check` clean · `cargo clippy --workspace --all-targets -D warnings` clean ·
`cargo nextest run --workspace --lib` 13459/13460 pass · offline `SQLX_OFFLINE=true cargo
check` clean (`.sqlx` cache regenerated for the changed backfill query) · new-code coverage
**97%** (891/917 new instrumented lines, gate ≥70%) — the uncovered remainder is defensive
`warn!`/`continue` arms on storage- and DB-error paths.

The one failing test, `repositories::tests::npm_scope_policy_put_get_round_trip_db`, is
**unrelated and pre-existing**: it asserts on the order of rows from an unordered `SELECT`,
so it depends on physical row order. It fails identically on this branch with these changes
stashed, and passes in CI.

Duplication gate: `alpine.rs` is on the `.jscpd.json` ignore list (the format handlers are
known structural near-duplicates), and it is the only `.rs` file changed, so the gate
reports N/A for this PR.

Closes #2633


brandonrc:
Returning to draft while three pre-merge review findings are addressed:

1. **Backfill never records failure** (`alpine.rs:720-735`) — an unparsable `.apk` stays checksumless, so it is re-fetched and re-parsed on every subsequent index GET. Combined with anonymous index reads on public repos, filename-only upload validation, and full-object `storage.get`, one junk package makes every `apk update` pay a repeated full-object read. Needs a parser-versioned negative marker, distinguishing transient storage errors from deterministic parse failures.
2. **Backfill is unbounded on the first index GET** (`alpine.rs:665`) — no LIMIT, N sequential full-object reads inside one request. Bound it per request and/or use `get_range` (the control segment is in the first few KB).
3. **`P:`/`V:` index injection** (`alpine.rs:78`) — no charset validation on the filename; a percent-encoded newline forges index entries. Pre-existing, but unexploitable until now because the index was unparseable — this PR makes it reachable, so it should be closed here or in a linked issue before the index is consumable.

The core fix (checksum/typeflag/`I:`) was independently verified against real apk-tools output and is not in question.

PR #2640, opened 2026-07-17
fix(helm): store and serve .prov provenance instead of discarding it

### LINKED ISSUE #2635: helm: .prov provenance files are accepted with 201 {"saved":true} and silently discarded — helm pull --verify can never succeed
## Summary

`backend/src/api/handlers/helm.rs` declares four routes and the string `prov` appears **nowhere** in the file. The ChartMuseum-compatible multipart upload (`POST /helm/{repo_key}/api/charts`) accepts a `prov` part, returns `201 {"saved":true}` — and **silently discards it**. No route serves `<chart>-<version>.tgz.prov` back.

Consequence: **`helm pull --verify` / `helm verify` can never succeed against a hosted AK helm repo.** A correctly signed chart (`helm package --sign`) round-trips through AK and comes out unverifiable. Chart provenance is unverifiable end-to-end for every hosted helm repo.

## The response is itself a defect

`{"saved":true}` for a part that was dropped is the trap: **a publisher has no way to know provenance was discarded.** They sign the chart, upload it, get a success response, and believe provenance shipped. Silently dropping a security artifact while reporting success is the worst of the available behaviors — it converts a missing feature into a false assurance. Whatever the storage outcome, the response must be honest.

This is also why the gap survived: `tests/formats/test-helm-conformance.sh:362-392` asserts only that the multipart POST returns 2xx — which it does.

## Reproduction (real helm 3.16.4)

Sign a chart, confirm it is good **before** AK sees it:

```
$ helm package ./dtf-provchart --sign --key dtf@example.com --keyring /tmp/dtf-sec.gpg
$ helm verify dtf-provchart-0.1.0.tgz --keyring /tmp/dtf-pub.gpg
Signed by: dtf-helm-prov <dtf@example.com>
Chart Hash Verified: sha256:41231c99c5c741e27c3db77625d061afae86a04da3ffdc15730b7a36576e7a1d
```

Upload chart + prov, then retrieve:

```
POST /helm/<repo>/api/charts (chart=@...tgz, prov=@...tgz.prov) -> 201 {"saved":true}

GET charts/dtf-provchart-0.1.0.tgz      -> HTTP 200
GET charts/dtf-provchart-0.1.0.tgz.prov -> HTTP 404  Chart not found
```

Every plausible provenance path 404s:

```
charts/dtf-provchart-0.1.0.tgz.prov -> 404
dtf-provchart-0.1.0.tgz.prov        -> 404
api/charts/dtf-provchart/0.1.0/prov -> 404
```

The real client:

```
$ helm pull dtfsign/dtf-provchart --version 0.1.0 --verify --keyring /tmp/dtf-pub.gpg
Error: failed to fetch provenance "http://backend:8080/helm/<repo>/charts/dtf-provchart-0.1.0.tgz.prov"
```

## Discrimination — the failure is provenance, not the chart or the repo

The **same** chart, **same** repo, **same** client without `--verify` downloads fine:

```
$ helm pull dtfsign/dtf-provchart --version 0.1.0
-> dtf-provchart-0.1.0.tgz   (200)
```

The only delta is the `.prov`.

## Expected

1. Persist the `prov` multipart part alongside its chart.
2. Serve it at the URL helm actually derives. Verified empirically against helm 3.16.4: helm takes the chart URL from `index.yaml` and **string-appends `.prov`** — so the required path is `/helm/{repo_key}/charts/{chart}-{version}.tgz.prov`. (Note: the helm index schema has **no** provenance field, so there is nothing to "advertise" in `index.yaml`; the requirement is that the advertised chart URL + `.prov` resolves.)
3. Deleting a chart must also remove its provenance.
4. The upload response must not report a save that did not happen.

## Impact

**High** for the Helm format. Chart provenance — the format's entire supply-chain trust mechanism — is unverifiable end-to-end against a hosted AK repo, while the API reports success.

## Test coverage note

`tests/formats/test-helm-provenance-verify.sh` (#72.7) was written for exactly this and has **0 hits in `release-gate.yml`** — it runs nowhere. Coverage that never executes is not coverage. The DTF `signing` tier (artifact-keeper-test#259) LEG D pins this as a KNOWN-RED with a real `helm pull --verify` and a discrimination control; it is designed to fail loudly the moment the bug is fixed so the pin gets removed.

Found by: DTF signing tier LEG D (D2/D3) — `rig/results/format-conformance/signing-helm-finding.md`.

brandonrc:
## Summary

Helm provenance was accepted and silently thrown away. `handlers/helm.rs` declared four routes and the string `prov` appeared **nowhere** in the file: the ChartMuseum multipart upload took a `prov` part, answered `201 {"saved":true}`, and dropped it — and no route served `<chart>-<version>.tgz.prov` back. So `helm pull --verify` / `helm verify` could **never** succeed against a hosted AK helm repo: a correctly signed chart round-tripped through AK and came out unverifiable, while the API reported success.

The misleading `{"saved":true}` is why this was believed to work, and why the corpus missed it — `tests/formats/test-helm-conformance.sh:362-392` asserts only that the POST returns 2xx, which it did.

This PR stores the prov alongside its chart and serves it at the URL the client actually derives.

### The path was verified against the real client, not the spec

Against real helm **3.16.4** (`alpine/helm:3.16`, arm64), `helm pull --verify` takes the chart URL it resolved from `index.yaml` and **string-appends `.prov`**:

```
GET /helm/<repo>/index.yaml                          200
GET /helm/<repo>/charts/<chart>-<ver>.tgz            200
GET /helm/<repo>/charts/<chart>-<ver>.tgz.prov       404   <- the bug
Error: failed to fetch provenance ".../<chart>-<ver>.tgz.prov"
```

Dropping the `.prov` at exactly that path — with `index.yaml` **unchanged** — flips it to `Chart Hash Verified`. Two consequences worth flagging for review:

- The **existing** `charts/{filename}` route is the right place to serve it; no new route is needed once the upload stores the prov under that filename.
- The helm index schema has **no** provenance field, so there is nothing to "advertise" in `index.yaml` (the finding listed this as an optional extra). The real requirement is that *the advertised chart URL + `.prov` resolves*, which is what the tests assert.

## Changes

- `upload_chart` stages the `prov` part and persists it as an artifact at `<chart>-<version>.tgz.prov`. The multipart loop no longer `break`s on `chart`, so part order does not matter — a part that is never read is a part that gets silently discarded.
- **The response no longer lies.** A prov that cannot be stored fails the upload (`400`) instead of vanishing behind a success reply. `prov` is reported only when provenance was actually written; a chart-only upload still returns exactly `{"saved":true}` (ChartMuseum-compatible).
- `query_charts_from_repo` excludes prov rows — the prov shares its chart's `name`/`version`, so without the filter every signed chart would render a **duplicate** `index.yaml` entry.
- **A chart and its provenance commit atomically.** Both objects are written to storage *before* either artifact row is inserted, and both rows are inserted inside **one transaction**. Ordering here is load-bearing — see below.
- `delete_chart` removes the chart together with its provenance. Its previous unordered `LIMIT 1` could just as easily have matched the prov row and left the chart behind once provenance exists, so this was load-bearing rather than cosmetic.

### Why the chart and the prov commit together

Storing the prov after committing the chart's artifact row is not safe, even though the prov is validated first.

The armor validation deliberately runs **before** the chart commits, which catches the overwhelmingly likely failure (a malformed prov) — but it cannot catch a *storage or database* fault on the prov. If one happened, the chart row was already committed: the response was an honest 5xx, but the repository now held a chart with **no provenance**, and the publisher's retry hit `ensure_unique_artifact_path` and got a permanent `409 Chart already exists`. Unretryable — and precisely the "signed chart that cannot be verified" state this issue exists to eliminate, just with an honest error code instead of a lying one.

So:

1. **Both objects go to storage first.** Object storage cannot join a DB transaction, so every fallible storage write now happens while the repository still has no row pointing at the coordinate.
2. **Both rows insert in one transaction.** A fault on the prov row rolls the chart row back with it.

Any fault therefore leaves the coordinate exactly as it was, and the retry succeeds. An object orphaned in storage by an interrupted upload is overwritten by that retry — the storage key is fully determined by chart name and version.

`proxy_helpers::insert_artifact_row` is the body of `insert_artifact` **without** the quarantine hold (which reads the row back through the pool and so must run post-commit). `insert_artifact`'s signature is unchanged and now delegates to it, so the other 44 call sites are untouched — the full `--lib` suite (13451 tests) passes.

13 tests added in the crate's `#[cfg(test)] mod tests` (`--lib`, none `#[ignore]`). The core ones were **confirmed to fail against the unfixed behavior** and pass with the fix:

| test | on unfixed code |
|---|---|
| `test_helm_upload_persists_prov_and_serves_exact_bytes` | **FAIL** |
| `test_helm_index_yaml_advertised_url_resolves_prov` | **FAIL** |
| `test_helm_upload_rejects_invalid_prov_instead_of_dropping_it` | **FAIL** |
| `test_helm_prov_storage_fault_leaves_no_chart_row_blocking_reupload` | **FAIL** (chart row committed; retry → 409) |

- `..._persists_prov_and_serves_exact_bytes` — real multipart upload; the `.prov` comes back byte-for-byte (a single changed byte breaks helm's signature check).
- `..._index_yaml_advertised_url_resolves_prov` — takes the URL **out of** `index.yaml`, appends `.prov` exactly as helm does, fetches it; also asserts the prov does not leak in as a second chart entry.
- `..._rejects_invalid_prov_instead_of_dropping_it` — regression guard: an unstorable prov must not report `saved:true`.
- `test_dropped_prov_can_never_report_saved_prov` — regression guard on the response builder, which takes the *storage outcome* rather than the presence of the part.
- `..._prov_storage_fault_leaves_no_chart_row_blocking_reupload` — **atomicity, storage half.** Injects a real, prov-scoped storage fault (a directory squatting on the prov's destination makes the filesystem backend's final rename fail with `EISDIR`, while the chart's own put succeeds — the residual class the armor pre-validation cannot catch). Asserts the upload fails, **no** artifact row is committed, and the retry after clearing the fault returns `201`, not `409`. Against the pre-atomicity code this test reproduces the defect exactly: `artifact rows committed after prov fault = 1`, and the retry returns **`409`** where `201` is required.
- `..._chart_and_prov_rows_roll_back_together` — **atomicity, DB half.** A `UNIQUE(repository_id, path)` violation on the prov row (the shape a publisher racing between the pre-flight uniqueness check and the insert produces) must take the chart row with it, exercising the same drop-rollback the handler's `?` performs.
- `..._delete_chart_also_removes_prov`, `..._upload_without_prov_is_unchanged_and_prov_404s`, plus pure unit tests for the filename/armor helpers (fixture is a real `helm package --sign` prov).

### Two things reviewers should know

- **`prov: true` means "stored", not "verified".** AK does not validate the signature and should not — verification is helm's job, with the user's keyring. The upload validates only that the part is PGP-armored provenance, so a malformed prov is rejected at the door instead of being stored as garbage that helm will later refuse. Release notes should not imply AK verifies chart signatures.
- **5 of the 13 tests are DB-backed and no-op without `DATABASE_URL`** (`Fixture::setup` returns `None` → early return). That is this crate's existing house idiom rather than something this PR introduces, but it is the same shape as the silently-skipping-coverage problem, so it is worth naming: the results below were produced **with** a live `DATABASE_URL`, and the core tests were independently confirmed to fail on the unfixed code — they are not passing vacuously.

## End-to-end proof: DTF signing tier LEG D

LEG D of the `signing` tier (artifact-keeper-test#259) is the helm provenance leg, pinned as **2 KNOWN-REDs**. Against `ak-backend:fix-helm-prov`, both pins tripped their designed tripwire:

```
FAIL: KNOWN-RED leg now PASSES — the backend bug looks FIXED. Remove the pin ... (D2)
FAIL: KNOWN-RED leg now PASSES — the backend bug looks FIXED. Remove the pin ... (D3)
```

Un-pinning D2/D3 to plain assertions makes the tier **genuinely green** (`TIER signing: PASS (exit 0)`, 0 FAILs), with the assertions un-softened:

```
D1: POST /api/charts (chart + prov) -> 201  {"prov":true,"saved":true}
D2: GET charts/dtf-provchart-0.1.0.tgz.prov -> HTTP 200  -----BEGIN PGP SIGNED MESSAGE-----
D3: REAL `helm pull --verify` ->  Signed by: dtf-helm-prov <dtf@example.com>
                                  Chart Hash Verified: sha256:42b37ef1c253e5f9...
D4: discrimination — same chart also pulls without --verify
```

Re-run against the atomicity change (same un-pin, `artifact-keeper-backend:fix-helm-prov`) — the real client flow is unaffected by the reordering + transaction:

```
D1: POST /api/charts (chart + prov) -> 201  {"prov":true,"saved":true}
D2: GET charts/dtf-provchart-0.1.0.tgz.prov -> HTTP 200  -----BEGIN PGP SIGNED MESSAGE-----
D3: REAL `helm pull --verify` ->  Signed by: dtf-helm-prov <dtf@example.com>
                                  Using Key With Fingerprint: 7EF60D643F29A39CABE268A212BA41048F10D764
                                  Chart Hash Verified: sha256:f33ae766cd7b2e6eb933b5e76c348f0a7f24edaa9ca1ea8a6835cd06546e78ef
D4: discrimination — same chart also pulls without --verify

Results: 24 passed, 0 failed, 0 skipped (24 total, 41s)
=== TIER signing: PASS (exit 0) ===
```

The matching tier un-pin needs to land alongside this PR, otherwise the tripwire fires on merge (which is the design working as intended).

## Test Checklist
- [x] E2E tests added/updated (if applicable) — DTF `signing` tier LEG D un-pinned
- [x] Manually tested locally
- [x] No regressions in existing tests — `cargo nextest run --workspace --lib`: **13451 passed, 0 failed** (the `insert_artifact` delegation touches 44 other call sites; the full suite is the blast-radius gate)

## API Changes
- [x] N/A - no API changes

No new routes, no schema change, no migration: provenance is served by the existing `charts/{filename}` route. The upload response gains an additive `prov` field on chart+prov uploads only; the chart-only reply is unchanged.

Closes #2635

### DIFF
--- a/backend/src/api/handlers/helm.rs
+++ b/backend/src/api/handlers/helm.rs
@@ -38,6 +38,7 @@ use crate::api::SharedState;
 use crate::formats::helm::{generate_index_yaml, ChartYaml, HelmHandler, HelmIndex};
 use crate::models::repository::{RepositoryFormat, RepositoryType};
 use crate::services::proxy_service::ProxyService;
+use crate::services::quarantine_service;
 
 // ---------------------------------------------------------------------------
 // Router
@@ -639,6 +640,23 @@ async fn upload_chart(
         .await?;
     }
 
+    // A chart and its provenance are one publish, so they must land as one
+    // unit (#2635). Ordering below is load-bearing:
+    //
+    //   1. BOTH objects go to storage first. Object storage cannot join a DB
+    //      transaction, so every fallible storage write happens while the
+    //      repository still has no row pointing at this coordinate.
+    //   2. BOTH rows are inserted inside ONE transaction. A fault on the prov
+    //      row rolls the chart row back with it.
+    //
+    // The property that matters is that a failure is *retryable*: nothing
+    // commits a chart row until the prov is safely stored, so the publisher's
+    // retry cannot collide with a half-finished predecessor in
+    // `ensure_unique_artifact_path` and get a 409 it can never clear. An
+    // orphaned object left in storage by an interrupted upload is overwritten
+    // by that retry -- the storage key is fully determined by chart name and
+    // version.
+
     // Stream the staged archive into the repo's StorageBackend via `put_stream`,
     // which computes the SHA-256 incrementally as it copies (no re-hash pass).
     let storage_key = format!("helm/{}/{}/{}", chart_name, chart_version, filename);
@@ -647,9 +665,28 @@ async fn upload_chart(
 
     let size_bytes = put.bytes_written as i64;
 
-    // Insert artifact record
-    let artifact_id = proxy_helpers::insert_artifact(
-        &state.db,
+    // #2635: persist the provenance BEFORE the chart row is committed. Any
+    // failure here is propagated with `?` -- the handler must never fall
+    // through to a `{"saved":true}` reply while the prov is on the floor, and
+    // must never leave behind a chart that advertises provenance it cannot
+    // serve.
+    let prov_storage_key = format!("helm/{}/{}/{}", chart_name, chart_version, prov_filename);
+    let prov_put = match staged_prov {
+        Some(prov) => {
+            Some(proxy_helpers::put_artifact_stream(&state, &repo, &prov_storage_key, prov).await?)
+        }
+        None => None,
+    };
+
+    // Both objects are in storage. Commit both rows together or neither.
+    let mut tx = state
+        .db
+        .begin()
+        .await
+        .map_err(|e| proxy_helpers::internal_error("Database", e))?;
+
+    let artifact_id = proxy_helpers::insert_artifact_row(
+        &mut tx,
         proxy_helpers::NewArtifact {
             repository_id: repo.id,
             path: &artifact_path,
@@ -663,6 +700,37 @@ async fn upload_chart(
         },
     )
     .await?;
+
+    let prov_artifact_id = match prov_put.as_ref() {
+        Some(prov_put) => Some(
+            proxy_helpers::insert_artifact_row(
+                &mut tx,
+                proxy_helpers::NewArtifact {
+                    repository_id: repo.id,
+                    path: &prov_artifact_path,
+                    name: chart_name,
+                    version: chart_version,
+                    size_bytes: prov_put.bytes_written as i64,
+                    checksum_sha256: &prov_put.checksum_sha256,
+                    content_type: PROV_CONTENT_TYPE,
+                    storage_key: &prov_storage_key,
+                    uploaded_by: user_id,
+                },
+            )
+            .await?,
+        ),
+        None => None,
+    };
+
+    // Until this returns, a fault has left the repository exactly as it was.
+    tx.commit()
+        .await
+        .map_err(|e| proxy_helpers::internal_error("Database", e))?;
+
+    // Post-commit follow-ups. The quarantine hold reads the artifact row back
+    // through the pool, so it can only run once the rows are visible; metadata
+    // recording is best-effort by contract.
+    quarantine_service::apply_upload_hold_hosted(&state.db, repo.id, artifact_id).await;
 
     // Build metadata JSON including the full Chart.yaml data
     let helm_metadata = serde_json::json!({
@@ -680,30 +748,8 @@ async fn upload_chart(
     )
     .await;
 
-    // #2635: persist the provenance alongside its chart. Any failure here is
-    // propagated with `?` -- the handler must never fall through to a
-    // `{"saved":true}` reply while the prov is on the floor.
-    let mut prov_stored = false;
-    if let Some(prov) = staged_prov {
-        let prov_storage_key = format!("helm/{}/{}/{}", chart_name, chart_version, prov_filename);
-        let prov_put =
-            proxy_helpers::put_artifact_stream(&state, &repo, &prov_storage_key, prov).await?;
-
-        let prov_artifact_id = proxy_helpers::insert_artifact(
-            &state.db,
-            proxy_helpers::NewArtifact {
-                repository_id: repo.id,
-                path: &prov_artifact_path,
-                name: chart_name,
-                version: chart_version,
-                size_bytes: prov_put.bytes_written as i64,
-                checksum_sha256: &prov_put.checksum_sha256,
-                content_type: PROV_CONTENT_TYPE,
-                storage_key: &prov_storage_key,
-                uploaded_by: user_id,
-            },
-        )
-        .await?;
+    if let Some(prov_artifact_id) = prov_artifact_id {
+        quarantine_service::apply_upload_hold_hosted(&state.db, repo.id, prov_artifact_id).await;
 
         proxy_helpers::record_artifact_metadata(
             &state.db,
@@ -718,9 +764,9 @@ async fn upload_chart(
             }),
         )
         .await;
-
-        prov_stored = true;
     }
+
+    let prov_stored = prov_artifact_id.is_some();
 
     info!(
         "Helm upload: {} {} to repo {} (provenance: {})",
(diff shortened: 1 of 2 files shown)


brandonrc:
Review follow-up — the atomicity item is fixed in c050ae1, and the two out-of-scope findings are filed.

## chart+prov commit is now atomic

Adopted **both** halves of the suggestion rather than picking one, because they close different faults:

1. **Both objects go to storage before either row is inserted.** Object storage cannot join a DB transaction, so every fallible storage write now happens while the repository still has no row pointing at the coordinate. This closes the storage-fault half.
2. **Both rows insert in one transaction** (`proxy_helpers::insert_artifact_row` on a `db.begin()` tx). This closes the DB-fault half — including the `UNIQUE(repository_id, path)` race between the pre-flight uniqueness check and the insert, which reordering alone would not have covered.

Any fault now leaves the coordinate exactly as it was, so the publisher's retry succeeds instead of hitting an unclearable 409. An object orphaned in storage by an interrupted upload is overwritten by that retry — the storage key is fully determined by chart name and version.

`insert_artifact_row` is the body of `insert_artifact` minus the quarantine hold, which reads the row back through the pool and so must run post-commit. `insert_artifact`'s signature is unchanged and delegates to it, leaving the other 44 call sites untouched — full `--lib` suite is 13451 passed / 0 failed.

## The residual was real — reproduced before fixing it

`test_helm_prov_storage_fault_leaves_no_chart_row_blocking_reupload` injects a genuine, prov-scoped storage fault (a directory squatting on the prov's destination → the filesystem backend's final rename fails `EISDIR`, while the chart's own put succeeds). Against the **pre-atomicity** code it reproduces the described wedge exactly:

```
PREFIX-EVIDENCE: artifact rows committed after prov fault = 1
assertion `left == right` failed: the retry after a transient storage fault must succeed, not 409
  left: 409
 right: 201
```

So both claimed consequences are confirmed: the chart row committed on its own, and the publisher was then permanently stuck on `409 Chart already exists`. With the fix, 0 rows are committed and the retry returns `201`.

A second test (`..._chart_and_prov_rows_roll_back_together`) covers the DB half via a unique-violation on the prov row, exercising the same drop-rollback the handler's `?` performs.

## Verification

- **DTF signing tier LEG D** re-run against the patched image with the local D2/D3 un-pin: `Results: 24 passed, 0 failed` / `TIER signing: PASS (exit 0)`. Real `helm pull --verify` still reports `Signed by: dtf-helm-prov` + `Chart Hash Verified` — the reordering and transaction do not disturb the client flow.
- fmt / clippy `-D warnings` / `nextest --workspace --lib` (13451 passed) all green. No `.sqlx` regen: the added queries are runtime-bound, not `query!` macros (`SQLX_OFFLINE=true cargo check --lib --tests` is clean).

## Filed, not fixed here

- **#2653** — remote/proxy helm repos still cannot serve provenance (`download_chart` early-returns 404 for `is_prov` before `download_chart_via_index`). Not a regression; this PR closes **hosted** only. Milestone 1.6.0.
- **#2654** — `build_download_response` interpolates the filename into `Content-Disposition` unsanitized and `.unwrap()`s the builder. Pre-existing and equally reachable via `.tgz`. Filed with the reachability caveat stated explicitly — and noting that on a modern `http` crate the realistic outcome is a **panic** (the builder rejects invalid header values rather than emitting them), which is a smaller but still real problem. Milestone 1.7.0, re-triage to 1.6.0 if a repro confirms CRLF can reach the header.

Left as a draft.

PR #2610, opened 2026-07-16
fix: complete authorization hardening

valfon1393:
## Summary

Closes #2603 

This follow-up completes the authorization hardening that remained after the
earlier work was split into upstream changes. It keeps public repository
visibility read-only, centralizes action-aware repository authorization, and
applies the same decision consistently to generic, resumable, native, and OCI
mutation paths.

It also requires repository administration for delegated repository tokens and
repository configuration subresources, and applies early global-admin guards to
system-wide peer, signing-key, security-configuration, and migration routes.
Protected routes now stop unauthorized callers before request validation or
resource lookup.

The detailed reproduction matrix is intentionally not included in this public
PR. A sanitized endpoint/status report is available privately to maintainers.

### Authorization model

- Public visibility permits reads; it does not imply `write` or `delete`.
- Repository actions are evaluated as `read`, `write`, `delete`, or `admin`.
- Applicable direct, group, service-account, and inherited project rules are
  authoritative; repository `admin` implies the narrower actions.
- Repositories without applicable fine-grained rules retain legacy role
  compatibility for read/write, but `developer` no longer implies delete.
- Upload sessions recheck repository write access on chunk, status, and complete
  operations; their owner can still cancel staged data after access is revoked.
- Global administrators retain the existing bypass.

### Explicit non-goals

- Remote instances remain authenticated user-owned self-service records; their
  service scopes operations by the caller's user id.
- SBOM generation retains the current upstream artifact-visibility policy,
  including generation for artifacts visible in public repositories.

### CI notes and existing test updates

Several existing positive-path tests used a legacy developer fixture while
expecting a destructive operation to reach an immutability, promotion, OCI
cleanup, or npm-path assertion. Those fixtures now explicitly grant `delete`
(and `write` where a later republish is part of the case). Their original
business assertions are unchanged; an ungranted developer remains denied.

The in-process full suite also exposed seven proxy quarantine tests that shared
one process-wide metadata-LRU key across different mock storage instances. Their
test repository keys are now unique. Production cache behavior and the expected
hold/no-hold results are unchanged.

No schema migration or new endpoint is introduced.

## Regression test (required for `fix/*` PRs)

- [x] This PR is a `fix/*` AND adds/updates a test that would have caught the bug
- [ ] N/A - this is not a bug fix

## Test Checklist

- [x] Unit tests added/updated
- [x] Integration tests added/updated (if applicable)
- [ ] E2E tests added/updated (if applicable)
- [x] Manually tested locally
[...]

PR #2675, opened 2026-07-17
fix: add durable repository owner capability

### LINKED ISSUE #2666: Design: repo owners have no durable owner capability (roles.permissions is never populated)
Surfaced while reviewing #2610 (external contribution, authz hardening). Not a bug in that PR — a pre-existing structural gap in the ownership model that **blocks the good half of #2610 and will block any future action-aware authz work**.

## The gap

1. A repo creator receives a `developer` grant scoped to the repo — `backend/src/services/repository_service.rs:756-772`, commented "Owner auto-grant... so the creator retains access under per-repo authorization". **`developer` is the strongest thing an owner ever gets.**
2. `backend/migrations/002_roles.sql:44-47` seeds `admin` / `developer` / `reader` **without a `permissions` value**, so `roles.permissions = {}`.
3. **No migration anywhere populates it.** (Grepped all of `backend/migrations/`.)
4. Therefore any authz model that reads `roles.permissions` sees an empty set for every legacy role, and can only fall back to matching on the role *name*.
5. Remediation requires a fine-grained `permissions` row, and `create_permission` is `require_admin()`-gated (`backend/src/api/handlers/permissions.rs:265`) — **global admins only**. A repo owner cannot grant themselves anything on their own repo.

## Why it bites

Any change that makes authorization action-aware (rather than role-name-aware) collapses `delete` and repo-`admin` to global-admins-only on a default install that never adopted fine-grained rules. In #2610 that concretely means a repo owner can no longer delete artifacts, can no longer `handle_delete_manifest`, and **can no longer mint repo-scoped tokens** — repo-scoped token self-service silently becomes a global-admin feature.

There is also a second-order trap: if legacy-role fallback is gated on "this repo/project has no fine-grained rules", then adding one rule to fix an owner drops **every other legacy principal on that repo** to no-access. Repo-wide or project-wide, all-or-nothing. That is our familiar rules-empty-falls-open pattern inverted into first-rule-falls-closed.

## Options to weigh (not yet decided)

- **A real `owner` role / capability** on the repo, granted at create, that does not depend on rules a global admin must write.
- **A backfill migration** populating `roles.permissions` for the seeded roles, giving the action-aware model something true to read.
- **Let repo owners write fine-grained rules scoped to their own repo** (relax `create_permission`s global-admin gate to repo-admin-on-that-repo).
- Any combination — plus a documented, staged upgrade path for installs that never adopted fine-grained rules.

## Constraint for whoever takes this

Whatever we pick needs a **backfill + staged migration path**, not a flag day. And it needs a test that actually runs: `backend/tests/*` integration targets are not on `ci.yml`s allowlist, coverage runs `--lib` only, and `detect-changes` is `if: github.event_name == push` so integration + E2E skip on every PR. A test for this placed in the wrong target would be fiction-green.

Refs: #2610, #2603.

valfon1393:
## Summary

This PR proposes the staged ownership foundation requested in #2666. If
accepted and merged, it resolves #2666 and provides the prerequisite needed to
revisit the authorization work originally proposed in #2610.

The change makes repository ownership durable, gives the seeded roles explicit
capabilities, preserves legacy access during upgrade, and adds one
principal-scoped repository action decision that later authorization hardening
can adopt consistently.

### Chosen design

This PR combines the first two options from #2666:

1. Add a real repository-scoped `repository-owner` role with
   `read`, `write`, `delete`, and `admin` capabilities.
2. Backfill `roles.permissions` for the seeded roles:
   - `admin`: `read`, `write`, `delete`, `admin`
   - `developer`: `read`, `write`
   - `reader`: `read`

It then adds the staged compatibility needed to make that combination safe:

- New repository creators receive both `repository-owner` and the existing
  `developer` assignment in the same transaction as repository creation.
- Repositories with an authoritative `created_by` value backfill that creator
  as owner.
- Older repositories without `created_by` promote their repository-scoped
  developers only when neither the repository nor its project has a
  fine-grained rule. This preserves their previous effective control without
  guessing through an ACL that is already authoritative.
- Legacy repositories governed by repository or project rules are deliberately
  left unchanged.

This PR does not relax the global-admin gate on permission-rule management.
That policy can be considered separately after endpoints have a durable
repository-admin capability to check; it is not required to establish owner
identity or complete this migration.

### Staged upgrade path

The rollout is additive rather than a flag day:

1. Migration `172_repository_owner_capability.sql` establishes truthful role
   capabilities and backfills ownership without removing existing assignments.
2. Repository creation dual-writes owner and developer grants, so old and new
   authorization code can coexist during the transition.
3. Existing handlers keep their current behavior in this PR. The new
   `check_repository_action` decision is available for later endpoint-by-endpoint
   adoption.
4. An application rollback remains compatible with the migrated database:
   existing columns and developer assignments are retained, and older code can
   ignore the additional owner role.

The migration uses conflict-safe inserts, and its regression test executes the
exact migration twice to verify replay safety.

### Repository action semantics

`PermissionService::check_repository_action` defines the transition target for
future handler hardening:

- Global administrators retain the existing bypass.
- A role carrying `admin` is a durable owner capability and cannot be
  accidentally narrowed by an ordinary fine-grained rule.
- An applicable direct, service-account, or group rule on the repository or
  inherited project is authoritative for that principal only.
- A rule for another principal does not disable an unrelated user's role
  fallback. This avoids the first-rule-falls-closed failure described in #2666.
- Without an applicable rule, the decision reads the assigned role's explicit
  capabilities rather than inferring actions from the role name.
- `admin` implies the narrower repository actions.

### Explicit non-goals and follow-up

- This prerequisite does not rewire repository mutation handlers or change
  their externally visible authorization behavior.
- It does not include the endpoint-level hardening from #2610.
- It does not change the permission-management API or add HTTP endpoints.

After this prerequisite merges, the endpoint hardening originally proposed in
#2610 can be adapted to the durable owner/action model in a fresh, focused
follow-up from the then-current `upstream/main`. Keeping those handler changes
separate lets this PR settle the ownership and migration model first.

### CI placement and regression coverage

The regression tests live in the library target rather than `backend/tests/*`:

- `test_repository_action_uses_owner_roles_and_principal_scoped_rules` covers
  owner durability, developer/reader capabilities, group and project
  inheritance, principal-scoped overrides, the global-admin bypass, and the
  first-rule-falls-closed regression.
- `test_repository_owner_migration_backfills_without_flag_day` executes the
  exact migration against isolated temporary tables, checks creator and legacy
  backfill behavior, verifies ruled repositories are not promoted, and reruns
  the migration to check idempotency.
- The existing repository-create regression now asserts that a creator receives
  exactly `developer` plus `repository-owner` during the staged rollout.
- Shared handler fixtures continue to grant ordinary `developer` membership;
  they no longer describe every test principal as an owner.

Both the Backend Unit Tests and Code Coverage jobs provision PostgreSQL, apply
migrations, set `DATABASE_URL`, and run `cargo nextest run --workspace --lib`.
These tests therefore execute in PR CI instead of landing in an integration or
E2E target that the workflow does not run.

### CI notes and test changes

No existing test was disabled, weakened, or rewritten to make this change pass.
The only changes to existing test code add the repository-create dual-write
assertion and keep shared handler fixtures as ordinary `developer` members.
The repository-action and migration/backfill checks are new DB-backed library
regressions.

Local Docker verification passed formatting, full workspace Clippy, a fresh
database migration through the complete branch migration set rebased onto
current `upstream/main`, and the three focused DB-backed owner/action,
migration/backfill, and repository-create regressions. The new DB-backed
regressions execute in the library target, so the same tests are included in PR
CI.

### Manual E2E verification

Before the latest upstream rebase, I deployed an earlier revision of this
feature patch as a `linux/amd64` image to a non-production Kubernetes
environment backed by an existing PostgreSQL database at migration `162`.

The feature migration was numbered `167` in that image and was later
renumbered to `172` to avoid conflicts with current `main`. Its SQL is
unchanged, and the current migration file is exercised directly by the
DB-backed library regression. The observations below retain the number used by
the deployed image.

Before rollout, I inserted controlled upgrade fixtures for:

- a repository with an authoritative creator;
- a rule-less legacy repository with a scoped developer;
- legacy developers governed by repository and project rules;
- a rule-less legacy repository with only a reader assignment.

Observed after rollout:

- migrations `163` through `167` completed successfully;
- all `41/41` repositories with `created_by` received a matching owner grant;
- the rule-less legacy developer was promoted;
- repository-rule, project-rule, and reader-only fixtures were not promoted;
- duplicate role-assignment groups remained at zero;
- a repository created through the authenticated REST endpoint received exactly
  one `developer` and one `repository-owner` grant for its creator;
- after replacing the backend pod, migration `167` remained recorded once,
  role and assignment counts were unchanged, and readiness returned HTTP 200;
- primary and replica agreed on the migrated state.

All disposable repositories, permission rows, project, and user records were
removed after verification. The environment retained only the intended role
capabilities and owner grants for its existing repositories.

## Test Checklist

- [ ] Integration tests added/updated
- [ ] E2E tests added/updated
- [x] Manually tested locally

## API Changes

- [ ] Migration is reversible (migration 172 is additive and forward-only;
      application rollback compatibility is retained through dual-write)
- [x] N/A - no HTTP API changes

### DIFF
--- a/backend/migrations/172_repository_owner_capability.sql
+++ b/backend/migrations/172_repository_owner_capability.sql
@@ -1,0 +1,68 @@
+-- Establish durable, action-aware repository ownership without switching
+-- existing installations to the new authorization model in one step.
+--
+-- The built-in roles were originally seeded with empty `permissions` arrays,
+-- so action-aware checks had to infer capabilities from role names. Populate
+-- the intended capabilities first; endpoint enforcement can then migrate to
+-- these values independently.
+UPDATE roles
+SET permissions = CASE name
+        WHEN 'admin' THEN ARRAY['read', 'write', 'delete', 'admin']::TEXT[]
+        WHEN 'developer' THEN ARRAY['read', 'write']::TEXT[]
+        WHEN 'reader' THEN ARRAY['read']::TEXT[]
+    END,
+    updated_at = NOW()
+WHERE name IN ('admin', 'developer', 'reader');
+
+-- Keep repository ownership distinct from the legacy developer role. The
+-- `admin` capability is durable: a fine-grained rule may narrow an ordinary
+-- role, but it must not silently strip the repository owner of control.
+INSERT INTO roles (name, description, permissions, is_system)
+VALUES (
+    'repository-owner',
+    'Full control of assigned repositories',
+    ARRAY['read', 'write', 'delete', 'admin']::TEXT[],
+    true
+)
+ON CONFLICT (name) DO UPDATE
+SET description = EXCLUDED.description,
+    permissions = EXCLUDED.permissions,
+    is_system = true,
+    updated_at = NOW();
+
+-- Repositories created after migration 128 have an authoritative creator.
+-- Add the owner role alongside the existing developer assignment so upgrades
+-- are additive and remain compatible with code that still reads legacy roles.
+INSERT INTO role_assignments (user_id, role_id, repository_id)
+SELECT repo.created_by, owner_role.id, repo.id
+FROM repositories repo
+CROSS JOIN roles owner_role
+WHERE owner_role.name = 'repository-owner'
+  AND repo.created_by IS NOT NULL
+ON CONFLICT (user_id, role_id, repository_id) DO NOTHING;
+
+-- Older repositories have no `created_by` value. On a rule-less repository,
+-- every repository-scoped developer currently receives the legacy mutation
+-- capability and may therefore be the historical owner. Preserve that
+-- effective access during the staged migration by promoting those principals
+-- to explicit owners. Repositories already governed by repository/project
+-- rules are deliberately left alone because their ACL is authoritative.
+INSERT INTO role_assignments (user_id, role_id, repository_id)
+SELECT legacy.user_id, owner_role.id, legacy.repository_id
+FROM role_assignments legacy
+JOIN roles legacy_role ON legacy_role.id = legacy.role_id
+JOIN repositories repo ON repo.id = legacy.repository_id
+CROSS JOIN roles owner_role
+WHERE legacy_role.name = 'developer'
+  AND owner_role.name = 'repository-owner'
+  AND repo.created_by IS NULL
+  AND NOT EXISTS (
+      SELECT 1
+      FROM permissions permission
+      WHERE (permission.target_type = 'repository' AND permission.target_id = repo.id)
+         OR (
+             permission.target_type = 'project'
+             AND permission.target_id = repo.project_id
+         )
+  )
+ON CONFLICT (user_id, role_id, repository_id) DO NOTHING;
--- a/backend/src/api/handlers/test_db_helpers.rs
+++ b/backend/src/api/handlers/test_db_helpers.rs
@@ -546,11 +546,9 @@ pub async fn send(app: Router, req: Request<Body>) -> (StatusCode, Bytes) {
     (status, body)
 }
 
-/// Grant `user_id` the `developer` role scoped to `repo_id`, mirroring the
-/// owner auto-grant that `RepositoryService::create` performs for real
-/// callers. Handler smoke tests authenticate as the fixture user, so without
-/// this grant the per-repo authorization check in `require_visible` /
-/// `require_repo_write_access` would reject them on private repositories.
+/// Grant `user_id` the `developer` role scoped to `repo_id`. Handler smoke
+/// tests use this for an ordinary read/write repository member; owner-specific
+/// tests should grant the `repository-owner` role explicitly.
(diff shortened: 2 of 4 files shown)


valfon1393:
Status note: this implementation remains ready for design review as the proposed prerequisite for #2666.

Resolving this foundation is important because the remaining endpoint hardening cannot be completed safely without a durable repository-owner capability and a staged migration path. Until that chain is completed, affected deployments may retain serious authorization gaps.

GitHub currently reports the branch as mergeable. The existing aggregate CI failure came from jobs being cancelled or timing out before test assertions ran, rather than from a failing regression test. I will refresh CI and address review feedback as needed.

Issue #2671, opened 2026-07-17
Mixed-backend deployments: maven_flat_object_owner.storage_key is not qualified by backend, so both tenants 404 their own objects


brandonrc:
In a deployment with two cloud backends registered (e.g. a repo on `s3` and another on `gcs`), two **physically distinct objects in two different clouds** collapse into one row in the Maven attribution namespace. Both tenants then lose access to their own artifacts.

## Root cause

`backend/migrations/163_maven_flat_object_attribution.sql:21-26`:

```sql
CREATE TABLE IF NOT EXISTS maven_flat_object_owner (
    storage_key   TEXT PRIMARY KEY,
    repository_id UUID NOT NULL REFERENCES repositories(id) ON DELETE CASCADE,
    ...
);
```

**`storage_key` is the primary key and is not qualified by storage backend.** Every backfill branch filters only `r.storage_backend <> filesystem` (163:48, 163:68, 163:104, 163:126, 163:153) and groups with `GROUP BY storage_key HAVING COUNT(DISTINCT repository_id) = 1`. All three runtime lookups key on `storage_key` alone (`maven_flat_attribution.rs:47`, `:54`, `:66`).

## Impact — fail-closed, not a leak

Repo A on `s3` and repo B on `gcs` publish the same Maven GAV:

- **Backfill:** `COUNT(DISTINCT repository_id) = 2`, so the key stays unattributed. `attributed_owner` returns `None`, `flat_key_readable` returns false. **Both tenants get 404 on their own objects.**
- **Live rows:** `OWNER_BY_ARTIFACT_ROW_SQL` returns 2 rows → `LayerResult::Ambiguous` → resolves to `None`. Same 404.
- **Write:** `guard_flat_key_writable` sees a foreign owner and returns `AppError::Authorization` (403), refusing a write that is physically harmless (different cloud, different bucket).

This is an availability/correctness bug, **not** a cross-tenant leak — the failure is closed, not open. Lower severity than #2504, but real.

## Reachability

Mixed-backend is a supported, reachable topology:
- `main.rs:445-476` auto-registers **additional** backends whose `from_env()` succeeds — an operator with both AWS and GCP credentials in the environment silently gets a two-cloud registry.
- `repositories.rs:2256-2276` lets admins create a repository on a non-default backend, validated against `storage_registry.is_available`.

## This is live on current HEAD

[...]

PR #2707, opened 2026-07-19
fix(maven): files[] attribution matches camelCase storageKey the upload handler writes (#2706)

### LINKED ISSUE #2706: Maven flat-key files[] attribution never matches real data: reads/backfill query snake_case storage_key, the upload handler writes camelCase storageKey
### Summary

The Maven flat-key attribution added by #2585 (for #2574/#2584) resolves row-less legacy companions via the parent artifact's metadata `files[]` array — but it matches `f->>'storage_key'` (snake_case), while the GAV-grouped upload handler has **always written those entries with camelCase `"storageKey"`**. The spelling never matches real data, so:

1. **Read-time layer 2** (`OWNER_BY_METADATA_FILES_SQL` in `backend/src/services/maven_flat_attribution.rs`) never resolves an owner from `files[]`.
2. **Migration 163 category (2)** (`metadata_files` backfill) inserts **zero rows**: `elem->>'storage_key'` is NULL for every element, so nothing passes the `LIKE 'maven/%'` filter.

Net effect on cloud backends: every legacy row-less GAV-grouped companion (`.pom`, `.module`, `-sources.jar`, `-javadoc.jar`) that #2585 was specifically designed to keep serving still fails closed with 404 after upgrading to ≥ 1.5.7. For a deployment with a history of native Maven/Gradle publishes this is a large set — the primary `.jar`/`.aar` resolves (live row) while its `.pom`/`.module` 404s, which breaks dependency resolution for those GAVs.

### Evidence (all in-tree)

- The writer: #418 (`a9c845f6`, "group Maven deploy artifacts by GAV coordinates") builds each `files[]` entry as `serde_json::json!({... "storageKey": storage_key ...})` — camelCase from day one.
- Other readers of the same data already use camelCase:
  - `backend/src/api/handlers/repositories.rs:3169` — `f.get("sizeBytes")`
  - `backend/src/api/handlers/sbom.rs:1617` — `f.get("storageKey")`
  - test fixtures at `backend/src/api/handlers/repositories.rs:8525` — `"storageKey": "maven/com/example/demo/1.0.0/demo-1.0.0.pom"`
- Only the two #2585 sites use snake_case: `maven_flat_attribution.rs` (`OWNER_BY_METADATA_FILES_SQL`) and `migrations/163_maven_flat_object_attribution.sql` category (2).

Quick check against any pre-1.5.7 dataset:

```sql
SELECT count(*) FILTER (WHERE elem ? 'storageKey')  AS camel,
       count(*) FILTER (WHERE elem ? 'storage_key') AS snake
FROM artifact_metadata am
CROSS JOIN LATERAL jsonb_array_elements(
  CASE WHEN jsonb_typeof(am.metadata->'files') = 'array'
       THEN am.metadata->'files' ELSE '[]'::jsonb END) AS elem;
```

returns `camel = <all>`, `snake = 0`.

### Repro (unit-level)

Seed an artifact + `artifact_metadata` row shaped like the upload handler writes it (`{"files": [{"storageKey": "maven/g/a/1.0/a-1.0.pom", ...}]}`), then call `attributed_owner(pool, "s3", "maven/g/a/1.0/a-1.0.pom")` → returns `None` (expected: the parent's repository). Consequently `flat_key_readable` is false for the genuine owner and the companion 404s.

### Suggested fix

- Read-time: match both spellings — `WHERE COALESCE(f->>'storageKey', f->>'storage_key') = $1`.
- A follow-up migration that re-runs 163's category (2) with the corrected spelling (plus the derived-checksum expansion for the newly attributed keys), `ON CONFLICT DO NOTHING`, against the post-#2671 `(storage_backend, storage_key)` schema.

Notably the `metadata_files` layer has no test today (the casing mismatch is invisible to the suite) — the fix should add one with a camelCase fixture matching the writer.

I have a PR ready and will link it here.

ThaSami:
Closes #2706.

## What

The `files[]` attribution layer from #2585 never matches real data: it queries `f->>'storage_key'` (snake_case), but the GAV-grouped upload handler has always written those entries with camelCase `"storageKey"` (#418 — and the other in-tree readers, `repositories.rs:3169` / `sbom.rs:1617`, already read camelCase). So read-time layer 2 resolves nothing and migration 163's `metadata_files` backfill inserts zero rows — every legacy row-less GAV-grouped companion (`.pom`, `.module`, `-sources.jar`) still fails closed on cloud backends, despite #2574's fix being designed to keep exactly those served. The primary `.jar`/`.aar` resolves (live row) while its `.pom`/`.module` 404s, breaking dependency resolution for those GAVs.

## How

1. **`maven_flat_attribution.rs`** — `OWNER_BY_METADATA_FILES_SQL` matches both spellings: `COALESCE(f->>'storageKey', f->>'storage_key') = $1`. camelCase is what the writer produces; snake_case is kept so any hand-repaired rows (inserted following the migration's spelling) don't regress.
2. **Migration 169** — re-runs 163's category (2) with the corrected spelling, plus the derived-checksum expansion (3) scoped to the newly attributed `metadata_files` keys. Written against the current post-#2671 schema (`storage_backend`-qualified composite PK, backend taken from the parent's repository). Idempotent and replay-safe: pure `INSERT ... ON CONFLICT DO NOTHING`, same single-owner-per-backend / ambiguous-stays-unattributed semantics as 163/168.
3. **Tests** — the metadata-files layer had no test (which is how the mismatch shipped invisibly). Added:
   - `test_attributed_owner_from_metadata_files_camelcase` — fixture shaped exactly like the writer's output; **fails on pre-fix SQL**, passes with the fix; also asserts owner readability (`flat_key_readable`), `.sha1` sidecar inheritance, and that a foreign repository still cannot read the key (isolation unchanged).
   - `test_attributed_owner_from_metadata_files_snake_case` — the migration-spelling fallback keeps working.

## Security posture

No change to the isolation model: attribution still requires a single provable owner per backend; ambiguous and orphan keys stay unattributed (fail closed). This only makes the already-intended layer actually fire on real data.

## Operator note

Deployments that upgraded through 1.5.7/1.5.8 with cloud storage get the missing `metadata_files` attributions on next start via migration 169. On large catalogs the backfill is a few `GROUP BY` passes over `artifact_metadata`/`artifacts` (same cost profile as 163).


ThaSami:
@brandonrc can you please review this, its blocking our upgrade

brandonrc:
> @brandonrc can you please review this, its blocking our upgrade 

On it

brandonrc:
Goal is to have this in the 1.6.1 release which should happen tonight/Texas  @ThaSami

PR #2804, opened 2026-07-21
feat(setup): configurable first-run password retrieval hint

### LINKED ISSUE #2802: First-run setup instructions assume docker compose; make the retrieval hint deployment-aware or configurable
The first-time setup screen instructs:

    docker exec artifact-keeper-backend cat /data/storage/admin.password

On Kubernetes (Helm chart deployments) this is wrong; the equivalent is kubectl exec into the backend pod. On managed or packaged deployments the operator may want to direct users elsewhere entirely (documentation link, support contact).

Proposal: keep the current text as the default for compose deployments, add an env var (for example SETUP_PASSWORD_HINT) that replaces the retrieval instruction when set, and have the Helm chart set a kubectl-appropriate default. Falls back to current behavior when unset.

brandonrc:
## Summary

The first-time-setup login screen instructs operators to retrieve the generated admin password with:

```
docker exec artifact-keeper-backend cat /data/storage/admin.password
```

That instruction assumes a Docker Compose deployment. On Kubernetes the equivalent is a `kubectl exec` into the backend pod, and on packaged or managed installs the operator may want to point users at internal documentation or a support contact entirely.

This adds an optional `SETUP_PASSWORD_HINT` env var (parsed in `config.rs` with the existing `env::var(...).ok()` pattern, trimmed, blank treated as unset). When set, the public unauthenticated `GET /api/v1/setup/status` endpoint the login screen already consults now returns the operator-supplied string in a new optional `setup_password_hint` field. When unset the field is omitted from the response, so the web UI keeps its current built-in Docker Compose instruction and existing deployments are unaffected.

The web frontend change that renders this field lives in a separate PR against `artifact-keeper-web` (filed as a dependent issue there); this backend PR is the dependency and is safe to merge on its own.

`.env.example` documents the new variable and the CHANGELOG has an entry under the 1.6.0 Added section.

## Test Checklist
- [x] Manually tested locally

Added `config::tests::test_config_setup_password_hint` (unset/blank/real-value parsing) and `auth::tests::test_setup_status_response_serialize_with_hint` (plus updated the existing serialize tests to assert the field is omitted when absent). `cargo fmt --check` and `cargo clippy --workspace` are clean.

## API Changes
- [ ] N/A - no API changes

Additive only: `setup_password_hint: Option<String>` was added to the existing `SetupStatusResponse` schema (already annotated). No new endpoints, no schema removals, no migration.

Closes #2802


brandonrc:
Frontend consumer of the new `setup_password_hint` field: artifact-keeper-web#620 (tracked by artifact-keeper-web#619). That PR falls back to the current default when this field is absent, so the two can merge in either order.

Issue #2748, opened 2026-07-20
uv cannot install packages through a virtual PyPI repository


dshvubsk:
### Description

When using Artifact Keeper as a Python package registry, uv cannot install packages through a virtual PyPI repository.

The same package can be installed successfully when uv is configured to use either:

the local repository directly;
the remote repository directly.

However, when the virtual repository that combines the local and remote repositories is used, uv cannot find or download the package.

The virtual repository works differently with pip, so the problem appears to be specific to the interaction between Artifact Keeper virtual PyPI repositories and uv.

Repository structure

The following repositories are configured:

Local repository: bgd-pip-local
Remote repository: bgd-pip-remote
Virtual repository: bgd-pip

The virtual repository contains both bgd-pip-local and bgd-pip-remote.

uv configuration using the virtual repository
[tool.uv]
allow-insecure-host = ["artifact-keeper.example.com"]

[[tool.uv.index]]
name = "artifact-keeper"
url = "https://artifact-keeper.example.com/pypi/bgd-pip/simple/"
default = true

Installation command:

uv add <package-name>

or:

uv pip install <package-name>

Working configurations
Local repository
[[tool.uv.index]]
name = "artifact-keeper-local"
url = "https://artifact-keeper.example.com/pypi/bgd-pip-local/simple/"
default = true

This configuration works for packages stored in the local repository.

Remote repository
[[tool.uv.index]]
name = "artifact-keeper-remote"
url = "https://artifact-keeper.example.com/pypi/bgd-pip-remote/simple/"
default = true

This configuration works for packages fetched through the remote repository.

Additional observations
Direct access to the local repository works with uv.
Direct access to the remote repository works with uv.
Access through the virtual repository does not work with uv.
Installation through pip works, when the same virtual repository is used.
This may be related to the HTML response, redirects, package links, content type, or PEP 503/PEP 691 compatibility of the virtual repository endpoint.

### Environment Information

- Deployment type:  Helm Chart Openshift
- Operating System: Linux
- Database: PostgreSQL
- Reverse Proxy: Openshift Route
- Network setup: Corporate network with outbound HTTP/HTTPS proxy, internal DNS, custom CA certificates, and NO_PROXY configured for internal domains and Kubernetes service networks

### Backend Version

1.5.6

### Web UI Version

1.5.6

### Package Format

None

### Expected Behavior

The virtual repository should expose packages from both the local and remote repositories and should be compatible with standard Python package clients, including uv.

[...]

PR #2810, opened 2026-07-22
fix(pypi): sniff simple-index body instead of trusting upstream Content-Type; never serve un-rewritten upstream URLs

### LINKED ISSUE #2801: PyPI simple-index proxy serves raw un-rewritten upstream bytes when upstream Content-Type isn't JSON/HTML (offsite URL leak + uv rejects)
### Summary

The PyPI simple-index proxy render trusts the upstream `Content-Type` verbatim. When a member/remote upstream returns a simple-index body whose `Content-Type` is **neither** `application/vnd.pypi.simple.v1+json`/`json` **nor** `text/html` — e.g. a corporate proxy or CDN that mangles it to `application/octet-stream` — the fallthrough at `backend/src/api/handlers/pypi.rs:985-993` (virtual) and `:705-720` (direct-remote) serves the **raw upstream bytes with the upstream Content-Type, un-rewritten**. Two consequences, both reproduced (real uv 0.11.30):

1. **Offsite download URLs leak** — the served `files[].url` still point at `files.pythonhosted.org` (or whatever the upstream returns) instead of being rewritten through `/pypi/<repo>/...`. SSRF-adjacent, and fatal in air-gapped / NO_PROXY installs where the client can only reach Artifact Keeper.
2. **uv rejects the response** — uv (PEP 691/503-strict) errors `Unsupported Content-Type "application/octet-stream" ... Expected JSON or HTML`, so installs fail. (pip is lenient and may still work — matching the pip-works/uv-fails asymmetry in #2748.)

### Fix

Sniff the body (JSON vs HTML) instead of trusting the upstream Content-Type, via one shared helper applied symmetrically to the direct-remote and virtual render sites; rewrite download URLs through the repo and set the correct simple-index Content-Type; on unrecognizable/binary bodies return 502 rather than serving raw offsite bytes. Must not change the conformant happy path (json/html Content-Types keep their existing branches) or the PEP 708 isolate / union merge.

### Relation to #2748

Likely the real root cause of #2748 (uv-through-virtual fails) for reporters behind a Content-Type-rewriting corporate proxy. #2748 itself could not be reproduced with a conformant upstream; this hardening should be shipped and the #2748 reporter asked to re-test + provide `uv --version` + full `uv -v` error if it persists.

brandonrc:
## Summary

The PyPI simple-index proxy renders an upstream response by **trusting the
upstream `Content-Type`**. When a member/remote upstream returns a simple-index
body whose `Content-Type` is neither JSON (`application/vnd.pypi.simple.v1+json`)
nor `text/html` — e.g. a corporate outbound proxy that mangles it to
`application/octet-stream` — both render sites fall through and serve the **raw
upstream bytes with the raw upstream `Content-Type`, un-rewritten**:

- the direct-remote render site, and
- the virtual render site.

That fallthrough has two consequences:

1. **Offsite download-URL leak.** The raw body still carries absolute
   `files.pythonhosted.*` URLs that were never rewritten to route through the
   repo. This is unreachable (and a data-exfil / SSRF-adjacent surface) in
   air-gapped / `NO_PROXY` installs, which rely on every download URL pointing
   back at Artifact Keeper.
2. **PEP 691 clients break.** `uv` rejects the response with
   `Unsupported Content-Type "application/octet-stream" ... Expected JSON or HTML`.

### Fix

Replace the content-type-*trusting* fallthrough with content-*sniffing*, via one
shared classifier (`sniff_simple_index_body`) used by **both** render sites so
they cannot diverge:

- Leading `{`/`[` (trimmed) → treat as PEP 691 JSON; run it through the **same**
  JSON rewrite/merge path the conformant branch already uses (rewrite
  `files[].url` through `/pypi/<repo>/…`, splice local members for a virtual),
  and serve `application/vnd.pypi.simple.v1+json`.
- `<a `/`<!DOCTYPE`/`<html` → treat as PEP 503 HTML; run it through the **same**
  HTML rewrite/merge path and serve `text/html; charset=utf-8`.
- Anything else (binary / non-simple-index) → **`502 Bad Gateway`**. The proxy
  never passes raw upstream bytes / offsite URLs to the client.

The conformant happy path is unchanged: bodies whose `Content-Type` already
contains `json` or `text/html` keep their existing branches (which run first).
The PEP 708 dependency-confusion isolate and the local+remote union merge are
untouched. No migration.

Likely related to #2748 (a customer report of `uv` failing through a virtual
PyPI repo). If that customer's corporate proxy rewrites the upstream
simple-index `Content-Type`, this is the code path that would bite them — but
#2748's exact repro is still unconfirmed pending the reporter's `uv -v` error, so
this PR does **not** claim to close it. It closes the reproduced offsite-URL leak
+ non-conformant-upstream robustness gap tracked as #2801.

Closes #2801.

New unit tests in `backend/src/api/handlers/pypi.rs`:
- `test_sniff_classifies_pep691_json_body`, `_pep503_html_body`,
  `_binary_garbage_as_unknown` — the shared classifier.
- `test_sniff_then_json_rewrite_strips_offsite_urls` — a PEP 691 body labelled
  `application/octet-stream` is classified JSON and the rewrite path reroutes
  `files[].url` through the repo; asserts **no** `files.pythonhosted` URL
  survives.
- `test_sniff_then_html_rewrite_strips_offsite_urls` — same for a PEP 503 HTML
  body labelled `application/octet-stream`.
- `test_sniff_binary_body_maps_to_bad_gateway_never_serves_offsite` — binary
  garbage classifies as Unknown (→ 502) and yields no servable rewrite, so an
  embedded offsite marker can never reach the client.

## Test Checklist
- [x] Manually tested locally

Manual verification on an isolated instance: a mock upstream returning a JSON
(and an HTML) simple-index body with `Content-Type: application/octet-stream`,
requested through both a direct-remote and a virtual repo with `uv`'s `Accept`
header → `200` with the correct simple-index `Content-Type` and `files[].url`
rewritten under `/pypi/<repo>/…`; a binary body → `502`; no `files.pythonhosted`
URL in any response body.

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -719,8 +719,8 @@ async fn simple_project(
                     }
                 }
 
-                // Rewrite absolute download URLs to route through our proxy
-                let body_bytes: Vec<u8> = if ct.contains("text/html") {
+                // Rewrite absolute download URLs to route through our proxy.
+                if ct.contains("text/html") {
                     let html = String::from_utf8_lossy(&content);
                     let rewritten = rewrite_upstream_urls(&html, &repo_key, &project);
                     let rewritten = filter_pypi_simple_html_response(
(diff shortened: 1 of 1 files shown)


brandonrc:
Closing as a duplicate of #2807 (same #2801 content-sniffing fix). #2807 is already integrated on main + backported to release/1.6.x and validated by the DTF pypi-contenttype tier (17/17 green). Thanks — both implementations were equivalent.

Issue #2759, opened 2026-07-20
Federated group reconciler over-grants: attaches SSO/LDAP users to operator-managed (external_source=NULL) groups by name alone


brandonrc:
Found by red-team verify of #2756 (AD group sync). `sync_federated_groups_to_local_groups` (sso.rs ~1216-1314) upserts `groups(name) ON CONFLICT(name) DO UPDATE RETURNING id` and then UNCONDITIONALLY inserts `user_group_members` — so a federated (OIDC/SAML/LDAP) group name that collides with a pre-existing OPERATOR-MANAGED privileged local group (external_source IS NULL) silently adds the federated user as a MEMBER of that privileged group. Grants group memberships only (not is_admin), but the group's permissions are inherited. PRE-EXISTING for OIDC/SAML; #2756's cn-only AD mapping amplifies collision probability (CN=developers,OU=Contractors → local 'developers'). Fix: key the group lookup/upsert by (name, external_source, external_provider_id) — OR refuse to add federated members to external_source IS NULL groups — so federated cns can never attach users to operator-managed groups. This will be fixed as part of hardening #2756 (shared code). Milestone 1.7.0. SECURITY (MEDIUM).

PR #2828, opened 2026-07-22
fix(sso): per-provider trust_group_names opt-in for OIDC operator-group attachment

### LINKED ISSUE #2823: OIDC (GitLab) group sync no longer attaches operator-created same-named groups (1.6.1 #2759 guard)
**1.6.1 regression (behavioral).** Our 1.6.1 backport (#2781/eaa42a1) bundled the #2759 ownership guard into federated group sync: it refuses to attach an OIDC user to a same-named group NOT tagged as owned by that OIDC provider (i.e. operator-created groups). Pre-1.6.1 it attached by name. Reported by Jon Craig (GitLab OIDC) on 1.6.1.

**Proposed fix (pending customer confirmation that his AK groups are operator-created):** a per-provider `trust_group_names` opt-in, default OFF (keeps #2759 secure); ON restores name-matching for that trusted provider only. **Auth fix → red-team verify before merge.** If Jon's groups are provider-owned, this is a config/expectation issue, not code.

brandonrc:
## Summary

Adds a per-provider, default-false `trust_group_names` opt-in to OIDC providers
(`oidc_configs`) that restores pre-1.6.1 name-matching against **operator-managed**
groups for a single IdP the operator explicitly trusts.

Background: #2759 (1.6.1) added an ownership guard to the group-sync upsert
(`ON CONFLICT (name) DO UPDATE ... WHERE groups.external_source = EXCLUDED.external_source
AND groups.external_provider_id = EXCLUDED.external_provider_id`). For an
operator-created group `external_source IS NULL`, so `NULL = 'oidc'` is never true, the
`DO UPDATE` is skipped, no `RETURNING id` comes back, and the federated user is never
attached. That is the intended hardening for the general case, but it regressed
deployments that pre-create operator groups and expect the OIDC `groups` claim to
attach users to them by name (GitLab-OIDC, reported in #2823).

This change keeps that guard on and default-on for every provider. When an admin sets
`trust_group_names = true` for one provider, the sync — before the guarded upsert —
resolves an existing group whose name matches and whose `external_source IS NULL`
(operator-managed only) and inserts a `user_group_members` row for it, leaving the
`groups` row untouched (`external_source` / `external_provider_id` stay NULL, so the
group is not adopted into the OIDC namespace). Groups owned by a different provider
(`external_source = 'oidc'`, non-NULL) are never matched by this path and still fall
through to the guarded upsert, which requires a matching `external_provider_id`, so
cross-provider attachment remains impossible even with trust on.

Implementation mirrors the existing `allow_legacy_rsa_keys` per-provider flag end to
end (Row / Response / Create / Update / all SELECT/INSERT/UPDATE/RETURNING lists /
`oidc_row_to_response`). All OIDC CRUD uses runtime `sqlx::query`/`query_as`, so no
`.sqlx` regeneration is needed. Admin mutation continues to ride the existing
`require_admin` + `admin_middleware` gating (no handler changes). Migration `176`
adds the column `NOT NULL DEFAULT false`.

Closes #2823

Four DB-backed unit tests were added to `sso.rs::tests::sync_db`, calling the core
`sync_federated_groups_to_local_groups` directly:
- `test_trust_off_operator_group_not_attached` — trust OFF: operator group not attached, source stays NULL (the #2759 guard).
- `test_trust_on_operator_group_attached_owner_unchanged` — trust ON: user attached, group's `external_source`/`external_provider_id` stay NULL.
- `test_trust_on_cross_provider_oidc_group_not_attached` — trust ON but another provider's oidc-owned group: NOT attached, owner unchanged.
- `test_trust_on_new_name_auto_creates_oidc_group` — trust ON, brand-new name: auto-created oidc-owned + attached (create path intact).

## Test Checklist
- [x] Manually tested locally

Local verification (DTF tier `oidc-group-sync`, image `artifact-keeper-backend:fix-2823`):

```
--- CASE 1 happy path: claim [gitlab-devs] auto-creates group (external_source=oidc) and attaches user ---
  PASS
--- CASE 2 (#2759 guard): OIDC claim matching an operator NULL-source group does NOT attach the user (trust OFF) ---
  PASS
--- CASE 3 (pending #2823): trust_group_names=true attaches the user to the operator group ---
  PASS
--- CASE 4 cross-provider: a claim matching another OIDC provider's group does NOT attach the user ---
  PASS

  Results: 6 passed, 0 failed, 0 skipped (6 total)
=== TIER oidc-group-sync: PASS (exit 0) ===
```

CASE 3 flips from SKIP to PASS (POST accepts + persists `trust_group_names`, GET
round-trips `== true`, OIDC login attaches the user to the pre-created operator group
and `groups.external_source` for that group stays NULL). CASE 1 / CASE 2 (#2759 guard)
/ CASE 4 (cross-provider isolation) stay green. All 20 `sync_db` unit tests pass
against a throwaway Postgres; `cargo fmt --check`, `cargo clippy --lib --tests -D
warnings`, and jscpd (<3%) are green.

## API Changes
- [ ] N/A - no API changes

`CreateOidcConfigRequest` / `UpdateOidcConfigRequest` / `OidcConfigResponse` gain an
optional `trust_group_names` field (existing `ToSchema` derives auto-update the spec;
no new endpoints). Migration `176_oidc_trust_group_names.sql` is a single additive
`ADD COLUMN IF NOT EXISTS ... NOT NULL DEFAULT false` and is reversible via `DROP COLUMN`.

### DIFF
--- a/backend/migrations/176_oidc_trust_group_names.sql
+++ b/backend/migrations/176_oidc_trust_group_names.sql
@@ -1,0 +1,17 @@
+-- Fix C (#2823): per-provider opt-in restoring pre-1.6.1 name-matching for an
+-- operator-TRUSTED OIDC IdP only. #2759 (1.6.1) added an ownership guard to the
+-- group-sync upsert so an IdP-supplied group name can no longer attach a
+-- federated user to a same-named operator-managed group (external_source IS
+-- NULL); that regressed GitLab-OIDC deployments that pre-create operator groups
+-- and expect the OIDC `groups` claim to attach users by name (Jon Craig, #2823).
+--
+-- This column re-enables name-matching against operator-managed groups for a
+-- SINGLE provider the operator explicitly trusts. DEFAULT false: every existing
+-- and every other provider stays fully guarded exactly as 1.6.1/1.6.2 -- no
+-- grandfathering, no global relaxation. When true, the sync attaches the user to
+-- the existing operator group by resolving its id; the group's external_source /
+-- external_provider_id stay NULL (the group is NOT adopted into the OIDC
+-- namespace), and another provider's oidc-owned groups are never attached
+-- regardless of this flag.
+ALTER TABLE oidc_configs
+    ADD COLUMN IF NOT EXISTS trust_group_names BOOLEAN NOT NULL DEFAULT false;
--- a/backend/src/api/handlers/sso.rs
+++ b/backend/src/api/handlers/sso.rs
@@ -551,8 +551,15 @@ async fn oidc_callback_inner(
     //     Auto-create groups (tagged with external_source = 'oidc') on first
     //     sight and reconcile membership so removed groups drop their members.
     if row.map_groups_to_groups {
-        if let Err(e) =
-            sync_oidc_groups_to_local_groups(&state.db, user.id, provider_id, &groups).await
+        if let Err(e) = sync_federated_groups_to_local_groups(
+            &state.db,
+            user.id,
+            provider_id,
+            "oidc",
+            &groups,
+            row.trust_group_names,
+        )
+        .await
         {
             tracing::warn!(
                 error = %e,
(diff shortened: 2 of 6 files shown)


brandonrc:
Closing without merge. Re-diagnosis with the customer confirmed their OIDC groups do not exist at all: the actual root cause is a claim-name mismatch (GitLab publishes group memberships under the groups_direct claim; AK reads the groups claim by default and never calls userinfo), which this change does not address. The operator-created-group case this PR targets is a real but separate behavior, and enabling it re-opens the #2759 ownership hardening behind an opt-in with no confirmed demand. Deferring as a possible future feature rather than shipping it in the 1.6.2 patch. Branch fix/oidc-trust-group-names retained; DTF oidc-group-sync CASE 3 stays a capability-gated skip.

Issue #2754, opened 2026-07-20
cran/rubygems: bare-path generic uploads yield client-reconstructed download paths that 404 (audit #2589 follow-up)


brandonrc:
From the #2589 audit (PR #2753 fixed helm): cran and rubygems index formats carry only name/version and the CLIENT reconstructs the download filename. A bare-path generic upload whose derived version is the `sha256-<prefix>` fallback produces a client-side path that 404s. Unlike helm, there is no server-advertised URL to correct at serve time — the proper fix is accurate coordinate (name/version/filename) extraction at generic-upload time for these formats (or a suffix-resolve fallback that matches the stored bare path). Milestone 1.7.0.

Issue #2589, opened 2026-07-16
Audit on-demand index/metadata generation for other suffix-resolved formats (npm/helm/pypi/...) on bare-path uploads


brandonrc:
Follow-up to #2580 / #2587.

#2587 added an exact-path fallback to the shared suffix resolver so bare-path (generically-uploaded) artifacts are downloadable, and fixed RPM `primary.xml` `<location>`. Other formats that resolve by filename suffix (npm, helm, pypi, rubygems, etc.) now DOWNLOAD correctly via the same fallback, but each format's own on-demand index/metadata generation (analogous to rpm `primary.xml`) should be audited to ensure the advertised URLs/paths match the served route for bare-path uploads.

PR #2864, opened 2026-07-23
fix(formats): advertise stored basename in ansible/puppet/rubygems on-demand metadata download URLs

brandonrc:
## Summary

Follow-up to #2587 (rpm `primary.xml`) and #2754 (cran `PACKAGES` / rubygems
`specs.4.8.gz`), closing the remaining half of the #2589 audit.

Those PRs fixed the **repo-wide** on-demand indices: an artifact pushed through
the generic chunked-upload flow is stored at its **bare filename** with the
whole basename as `name` and a `sha256-<prefix>` fallback `version`
(`upload.rs::completed_format_artifact_version`), so any download URL/coordinate
reconstructed from `{name}-{version}` points at a path the suffix-resolving
download route cannot serve.

This PR audits the **name-addressed metadata endpoints** of the other
suffix-resolved formats and fixes the same class where the on-demand response
advertises a **reconstructed** download URL instead of the stored object's real
filename:

| Format | Endpoint | Field | Before | After |
|--------|----------|-------|--------|-------|
| Ansible Galaxy | `collection_info`, `version_info` | `download_url`, `artifact.filename` | `{namespace}-{name}-{version}.tar.gz` | stored basename |
| Puppet Forge | `module_info`, `release_list`, `release_info` | `file_uri` | `{owner}-{name}-{version}.tar.gz` | stored basename |
| RubyGems | `gem_info`, `gem_versions` | `gem_uri` | `{name}-{version}.gem` | stored basename |

### Fix (serve-time, no migration — retroactive for all historical bare-path artifacts)

- New shared helper `proxy_helpers::served_download_filename(path, fallback)`:
  returns the basename of the stored object `path` when present, else the
  format's reconstructed `{name}-{version}.<ext>`. For a natively-published
  artifact stored at `{name}/{version}/{file}` the basename already equals the
  reconstruction, so the advertised URL is **byte-identical** for native
  uploads; it only diverges for bare-path uploads, where it now matches the
  served route. This is the same fix already applied to helm `index.yaml`
  (`chart_download_url`), cran (`source_index_coordinates`) and rubygems
  (`spec_index_coordinates`).
- `ArtifactWithMetadata` (shared by the ansible/puppet/rubygems name-lookup
  helpers) now carries the stored `path`; the three `sqlx::query` lookups select
  `a.path`. All runtime `sqlx::query` (no `sqlx::query!` macros), so the offline
  `Check Rust` gate is unaffected.

No schema change: entirely serve-time, so it also repairs previously-uploaded
bare-path artifacts.

### Audited and already correct (no change)
- **rpm** `primary.xml` (#2587), **cran** `PACKAGES` + **rubygems**
  `specs.4.8.gz` (#2754), **helm** `index.yaml` — repo-wide indices already
  derive the served location/coordinates from the stored filename.
- **npm** packument and **pypi** simple index already advertise the stored
  basename (`a.path.rsplit('/')`).

### Deliberate follow-ups (out of scope here)
- **hex**: the JSON `/api/packages/{name}` release list (and its virtual-repo
  variant `build_package_info_json`) reconstructs `/tarballs/{name}-{version}.tar`.
  The virtual variant is fed `(version, checksum)` tuples with no stored `path`,
  and hex's primary install path is the **signed registry** payload (which is
  name/version-reconstructed by protocol), so a coherent fix needs a wider query
  change — tracked separately rather than bundled here.
- Formats whose download route does **not** use the shared suffix resolvers
  (debian `pool/`, conda, cocoapods, nuget, …) cannot serve bare-path uploads at
  all today; surfacing them is a separate, larger concern.

Coverage that fails on `main` and passes here:
- **Unit (`proxy_helpers.rs`)** — `served_download_filename`: native
  `{name}/{version}/{file}` layout → basename is byte-identical to the
  reconstruction; bare path → prefers the stored basename over the
  `{name}-{version}` (sha256-fallback) reconstruction; empty/trailing-slash path
  → falls back.
- **DTF tier (`artifact-keeper-test` `ondemand-index-formats`)** — a real
  generic (bare-path) chunked upload of an Ansible collection and a Puppet
  module; reads the served metadata JSON, extracts the advertised
  `download_url` / `file_uri`, and GETs it. Proven **RED → GREEN**:
  - RED (base image): both advertise `{name}-{version}-sha256-….tar.gz` → `404`.
  - GREEN (this image): both advertise the stored basename → `200`.

## Test Checklist
- [x] E2E tests added/updated (if applicable) — DTF tier in artifact-keeper-test
- [x] Manually tested locally

Gates: `cargo fmt --check` pass · `cargo clippy --lib --tests` clean ·
`cargo test --lib` (changed logic) pass · `jscpd --min-lines 10 --threshold 3`
on changed `.rs` under threshold.

## API Changes
- [x] N/A - no API changes (JSON field values now resolve; shapes unchanged)

### DIFF
--- a/backend/src/api/handlers/ansible.rs
+++ b/backend/src/api/handlers/ansible.rs
@@ -232,8 +232,12 @@ async fn collection_info(
             repo_key, namespace, name
         ),
         "download_url": format!(
-            "/ansible/{}/download/{}-{}-{}.tar.gz",
-            repo_key, namespace, name, latest_version
+            "/ansible/{}/download/{}",
+            repo_key,
+            proxy_helpers::served_download_filename(
+                &artifact.path,
+                &format!("{}-{}-{}.tar.gz", namespace, name, latest_version)
+            )
         ),
     });
 
@@ -322,16 +326,20 @@ async fn version_info(
     .unwrap_or(Some(0))
     .unwrap_or(0);
 
+    // Advertise the stored basename so the suffix-matching download route
+    // resolves the object for bare-path (generic) uploads as well as native
+    // ones (#2589); byte-identical to the reconstruction for native uploads.
+    let served_filename = proxy_helpers::served_download_filename(
+        &artifact.path,
+        &format!("{}-{}-{}.tar.gz", namespace, name, version),
+    );
     let json = serde_json::json!({
         "namespace": namespace,
         "name": name,
         "version": version,
-        "download_url": format!(
-            "/ansible/{}/download/{}-{}-{}.tar.gz",
-            repo_key, namespace, name, version
-        ),
+        "download_url": format!("/ansible/{}/download/{}", repo_key, served_filename),
         "artifact": {
-            "filename": format!("{}-{}-{}.tar.gz", namespace, name, version),
+            "filename": served_filename,
             "size": artifact.size_bytes,
             "sha256": artifact.checksum_sha256,
         },
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -3792,16 +3792,47 @@ pub async fn try_remote_or_virtual_download(
 }
 
 /// Artifact row exposing the columns most metadata endpoints need:
-/// id, version, size, checksum, and the raw `artifact_metadata.metadata`
-/// JSON. Returned by [`find_artifact_by_name_lowercase`] and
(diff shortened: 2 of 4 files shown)


brandonrc:
Superseded — advertising the stored basename for suffix-resolved formats (incl. ansible/puppet/rubygems) landed in #2891 (closing #2589). Closing as duplicate.

PR #2883, opened 2026-07-24
chore: versions

rockdrilla:
## Summary

Update many container images to more recent versions.

Part 1 (trivial):
- fix(ci): update aquasecurity/trivy to 0.72.0
- fix(ci): update anchore/grype to 0.116.0
- fix(ci): update opensearch to 2.19.6
- fix(ci): update dependencytrack to 4.14.3
- feat(ci): fresh image sync

Part 2 (non-trivial) - expect breakage here and there:
- feat(ci): update postgresql to 17
- feat(ci): update alpine to 3.24
- feat(ci): update debian to 13
- feat(ci): update keycloak to 26.7
- feat(ci): update python to 3.13
- feat(ci): update helm to 3.21
- feat(ci): update terraform to 1.15.8
- feat(ci): update golang to 1.26
- feat(ci): update nodejs to 24

Part 3 (almost trivial) - may break as with part 2:
- fix(ci): decouple versions
- feat(ci): update rust to 1.97
- feat(ci): update rust to 1.97.1

Part 4 - likely a follow-up to part 2 but after part 3:
- feat(ci): update protocolbuffers to 35.1

## Test Checklist
- [ ] Manually tested locally

## API Changes
- [ ] N/A - no API changes


rockdrilla:
@brandonrc I know it's heavy MR but I'm sure you are able to pick these commits, for example one by one or so.

brandonrc:
Thanks for this, and sorry it's sat for five weeks — that delay is on us, and it's the direct cause of most of what's below.

I ran two review passes over it (breakage-risk and supply-chain/pinning). First, credit where it's due, because both passes independently landed on the same thing: **the mechanical work is careful.** All 19 bumped tags actually resolve upstream (verified by manifest HEAD, including the non-obvious `rust:1.97-alpine3.24`), there are no stale references left behind in the files you touched, and the `mirror-images.yml` matrix is an exact superset of every `ci-mirror/*` reference on your head — I checked that in both directions. That's more care than a bulk-version PR usually gets.

The blockers aren't tag correctness. They're that `main` has moved a long way underneath the branch, and two of your four "parts" no longer mean what the description says.

You offered to let us pick commits, so the useful output here is a split plan rather than a verdict. It's at the bottom.

---

## The one thing I'd ask for first

**Rebase onto `d8c497e8`.** In five weeks main has: rewritten both Dockerfiles' grype stages to build from source, moved trivy to 0.69.3 behind a new retry guard, added `rust-toolchain.toml` pinning 1.98.0, added a `dtrack-init` service to local-dev, and moved every `ci.yml` postgres block by ~300 lines. Several items below only surface as conflicts after a rebase — better you resolve them with context than us guess in review.

## Blockers

**B1 — `820efd52` (grype 0.116.0) would revert supply-chain hardening. Please just drop it.**
Main no longer copies grype from the published image at all. `Dockerfile.backend.alpine:86` and `Dockerfile.backend:239` now set `ARG GRYPE_VERSION=0.117.0` and **build from source** at a pinned commit, with a `GO_MIN_PATCH=1.26.6` floor and an `x/mod` override (#3352, #3465) — the comment at `Dockerfile.backend.alpine:57-80` explains a Go stdlib CVE in Anchore's published binary that "cannot be bumped away." Taking this commit is a downgrade 0.117.0 → 0.116.0 that re-introduces those CVEs and desyncs the two Dockerfiles, tripping a guard in `config.rs`. Already done, better, on main.

**B2 — Keycloak 26.7 is a tag-only bump and cannot boot or authenticate.**
`0329fd40` leaves the 24.x-era config intact:
- `scripts/sso-e2e/docker-compose.yml:36-37` and `docker-compose.oidc-bootstrap-test.yml:10-11` still set `KEYCLOAK_ADMIN`/`KEYCLOAK_ADMIN_PASSWORD`. Keycloak 26.0 renamed these to `KC_BOOTSTRAP_ADMIN_USERNAME`/`KC_BOOTSTRAP_ADMIN_PASSWORD`. With the old names no bootstrap admin is created, so `setup-keycloak.sh:18-27` gets a null token and exits 1 — every OIDC/SAML e2e dies at setup before any of our code runs.
- `oidc-bootstrap-test.yml:14` sets `KC_HOSTNAME_STRICT_HTTPS`, removed in the 26 hostname-v2 rework; an unrecognised option aborts `start-dev`.
- `docker-compose.yml:43` mounts `keycloak_data:/opt/keycloak/data`, persisting the 24.0 H2 dev DB. 24.0 → 26.7 across two majors on the same H2 file isn't a supported path, so this fails locally even after the env fix.

The renames and the volume reset need to be in the same commit as the tag.

**B3 — `415f28be` "decouple versions" isn't a refactor; it moves two versions.**
It introduces `ENV RUST_VERSION=1.93.1` and `ENV PROTOC_VERSION=29.6`, where main has Rust **1.93.0** (`Dockerfile.backend:18`) and protoc **29.3** (`:15`). So cherry-picking "part 3, almost trivial" silently ships a Rust patch bump and a protoc minor bump into the image built on every PR — under a commit message saying the opposite. This is exactly the mixing that makes "just take the safe commits" fail. Re-cut it keeping 1.93.0 and 29.3 and it becomes the most valuable commit in the PR (see the plan).

**B4 — the mirror must land *and be dispatched* before anything consumes it.**
`85ce59ca` correctly adds all five new tags to `mirror-images.yml:23-42`, but that workflow only fires on `schedule: '0 4 * * 1'` and `workflow_dispatch`. Merge any part-2 commit first and `ci.yml:156,290,693` plus the smoke stack fail `manifest unknown` on every PR and push until the following Monday 04:00 UTC.

**B5 — Rust 1.97.x now contradicts `rust-toolchain.toml`.**
Main added `rust-toolchain.toml:30` `channel = "1.98.0"` (#3551) after you cut the branch. The PR sets 1.97.x at `Dockerfile.backend:4`, `.alpine:4`, `.dev:3`, and `docker-compose.test.yml:193`. `check-toolchain-pin.sh` won't catch it — it only greps `.github/workflows` for `dtolnay/rust-toolchain@` refs, never Dockerfiles. At build time rustup reads `rust-toolchain.toml` and it outranks the installed default, so the images download 1.98.0 anyway and the pinned `RUST_VERSION` becomes dead weight that merely *looks* authoritative. Purely a rebase artifact — 1.97 was current on 2026-07-24.
(`docker-compose.test.yml:193` is a native cargo *client* test, unrelated to MSRV — 1.75 → 1.97 there is fine.)

## Worth knowing

**PostgreSQL 17 splits into two very different risks.** Safe: `docker-compose.test.yml:33` is `tmpfs`, and the concurrency/mesh/migration-e2e stacks plus all three `ci.yml` service blocks use anonymous mounts — fresh `initdb` every run, no upgrade path needed. Unsafe: `docker-compose.yml:78` (+`postgres_data`, the *published* self-host compose), `docker-compose.local-dev.yml:29`, and `docker-compose.s3-maven-test.yml:50` carry named volumes. An existing user running `docker compose pull && up` gets `FATAL: database files are incompatible with server` and a stack that never starts, because the backend's `depends_on: service_healthy` never satisfies. No data loss, but down with no in-band explanation. Those two groups should be separate PRs, and the second needs a 1.9.0 upgrade note.

*On sqlx:* no PG17 interaction risk — `Cargo.toml:26` is sqlx 0.8, which speaks the stable v3 wire protocol and doesn't gate on server major; `.sqlx/` offline entries key on query text plus built-in type OIDs, stable across majors, so no cache regeneration. The real interaction is with **#2857 (sqlx 0.8.6 → 0.9.0)**: same blast radius (`test-backend-integration`). Land them in separate PRs and separate CI runs, PG17 first since it's revertible in one line.

**A green check on this PR proves much less than it looks.** On `pull_request` only `ci.yml` and `codeql.yml` run; `e2e.yml`, `mesh-e2e.yml`, `concurrency-e2e.yml` and `cache-correctness-e2e.yml` are all dispatch/schedule-only. Inside `ci.yml` the only compose invoked is the `smoke` profile, waiting on four containers. So CI covers postgres 17, alpine 3.24, python 3.13, node 24 and rust 1.97-slim — and covers **nothing** for Go 1.26, Debian 13, Helm 3.21, Terraform 1.15.8, protobuf, or any of `scripts/sso-e2e/`, `scripts/migration-e2e/`, the mesh/concurrency stacks, or `proof/*.yml`. Worth stating in the description so nobody mistakes green for evidence.

**`Dockerfile.backend.alpine:7` is the largest change and has zero PR coverage.** It moves the builder to `rust:${RUST_VERSION}-alpine${ALPINE_VERSION}`, and that file isn't in `ci.yml`'s `detect-changes` filters (`:953-962`) — it's built only by `build-backend-alpine` on push to main, whose publish is currently suspended. So a musl/openssl-static break surfaces post-merge on main. **That said, this change fixes real drift**: main builds the static musl binary on alpine 3.21 and runs it on an alpine 3.24 runtime. Aligning them is right; it just needs a local `docker build -f docker/Dockerfile.backend.alpine .` pasted into the PR as evidence.

One caution on the coupling, though: the Alpine suffix on the official `rust` image tracks *which Alpine that Rust release was built against*, not which Alpine you run on. Main's deliberate `alpine3.21` builder next to an `alpine:3.24` runtime is the evidence they were intentionally decoupled. Safer: keep `ALPINE_VERSION` for the seed and runtime stages, give the builder its own `ARG RUST_IMAGE_TAG`.

**The ARG decouple is currently inert.** `Dockerfile.backend.alpine:3-4` and `.dev:3` use `ARG` (overridable); `Dockerfile.backend:4-5` uses `ENV` — a `--build-arg` can't override an `ENV`, and it leaks into the final image env. And nothing supplies them: the only `build-args` blocks in the repo (`docker-publish.yml:384-386`, `:629-631`) pass `GIT_SHA` and `APP_VERSION` only. So post-PR each version still lives in one place per Dockerfile, just spelled differently. Good readability step; doesn't yet deliver "bump in one place." Wiring `ALPINE_VERSION`/`RUST_VERSION` through those two `build-args` blocks would.

**Two undeclared changes.** `291ae0b6` (debian 13) also adds `apt-get upgrade -y` and swaps the apt cleanup at `Dockerfile.backend.dev:8,16`. Neither is a version bump, and `apt-get upgrade -y` makes the dev image non-reproducible across rebuilds. Please split those out.

**protobuf 29.3 → 35.1** is a six-major jump on one of four build paths — `Dockerfile.backend:5` fetches 35.1 while `Dockerfile.backend.dev:14`, `release.yml:448`, `publish-binaries.yml:62` and `sync-openapi-spec.yml:36` all use the distro compiler. proto3 descriptors are stable and prost/tonic should be fine, but it's the widest jump with the least coverage, so it earns its own commit. Also worth confirming 35.1 publishes assets under the `protoc-35.1-linux-x86_64.zip` name.

**Leftovers after the sweep** (all post-rebase artifacts, not errors on your side):

| `path:line` | Leftover |
|---|---|
| `docker-publish.yml:771,774` | `trivy:0.69.3` in main's new pre-pull retry guard — after rebase it warms 0.69.3 while `:780` creates 0.72.0 |
| `docker-compose.local-dev.yml:136` | `dtrack-init` still on `alpine:3.23` — service added to main after you branched |
| `scripts/concurrency/README.md:25` | documents `python:3.12-slim`; the Dockerfile is now 3.13 |
| `scripts/migration-e2e/docker-compose.yml:19` | `artifactory-db` on PG17 backing `jfrog/artifactory-oss:latest` — worth verifying against Artifactory's supported-DB matrix; a failure there reads like a migration bug |
| `mirror-images.yml:21,25,27,31,35,39` | six now-unconsumed old tags, cleanup only |

(`scripts/native-tests/test-docker*.sh` alpine pins are deliberate test fixtures — correctly left alone.)

## Two notes that aren't about this PR

**Pinning.** Across the 24 files here there are **zero** `@sha256:` pins on either side; repo-wide there's exactly one (`Dockerfile.scanner-adapter:123`). So this PR is reproducibility-neutral for ~95% of its surface — floating tag in, floating tag out. Not something to fix here, but worth naming: we ship `check-floating-tag-promotion.sh`, `check-published-image-signature.sh` and cosign-signed checksums for our *own* outputs while pinning none of our *inputs*.

**`mirror-images.yml` is a trust boundary.** Least privilege is already right (`permissions: contents: read, packages: write`, both actions SHA-pinned, `GITHUB_TOKEN` not a PAT, no `pull_request` trigger) — credit for that. The gap is what it copies: `imagetools create --tag <target> <image>` resolves e.g. `postgres:17-alpine` from Docker Hub **by floating tag at run time, no digest, no signature check**, and republishes it under our namespace. Downstream then trusts an `artifact-keeper`-namespaced image whose contents were whatever Docker Hub served that Monday, with nothing recording which digest. Adding a `digest:` column to the matrix, or just echoing the resolved digest into the job summary, would make it auditable. Separate issue.

**Also no checksum verification on downloaded tools:** `Dockerfile.backend:17-20` fetches protoc over HTTPS and `unzip`s it with no `sha256sum`, and rustup is `curl | sh`. Since this PR already introduces `PROTOC_VERSION`, adding `PROTOC_SHA256_AMD64`/`_ARM64` beside it plus a `sha256sum -c` is ~4 lines and turns the largest unverified fetch in the build into a pinned one. Optional, but it's right there.

## Dependabot ordering

No conflict — the PR modifies zero `uses:` lines, and the action-SHA differences are pure staleness that a normal merge resolves toward main. Merge #3579–#3585 whenever. The one thing that *would* revert them is honouring "pick the commits you want" by cherry-picking whole files or `git checkout refs/pr/2883 -- <file>` on the four workflow files. We won't do that.

## Split plan

Each step independently mergeable and revertible:

1. **Mirror sync alone, first.** `85ce59ca`. Merge, then manually dispatch `mirror-images.yml` and confirm all five tags resolve in GHCR. Nothing else is safe until this has actually run.
2. **Trivial bumps.** `07f49c7b` (opensearch), `7b2fd1ba` (dtrack). Add `404c917a` (trivy) only after also fixing `docker-publish.yml:771,774`. **Drop `820efd52`.**
3. **Decouple, as a true no-op — and land it before every bump.** Re-cut `415f28be` keeping 1.93.0/29.3, change `ENV` → `ARG` in `Dockerfile.backend`, fold in the `ARG ALPINE_VERSION` currently riding in `37f7279a`, and wire both through the `build-args` at `docker-publish.yml:384`/`:629`. **This is the highest-leverage change in the PR** — after it, every remaining bump is a one-line default change we can review and revert independently.
4. **Rust → 1.98.0** (reworking `a8e0d11c`+`f5f09c71`) to match `rust-toolchain.toml`. Mirror `rust:1.98-slim` in step 1 instead of `1.97-slim`.
5. **Low-coupling client images**, batched: python 3.13, node 24, helm 3.21, terraform 1.15.8, golang 1.26. Fix the concurrency README here.
6. **Alpine 3.24** (`37f7279a` minus the ARG hunk). Add `docker-compose.local-dev.yml:136`. Paste a local alpine build into the PR — CI won't do it.
7. **protobuf 35.1** alone, so a `build.rs` failure is unambiguous.
8. **Debian 13**, with `apt-get upgrade -y` split out or dropped.
9. **PG17, CI + ephemeral stacks only.** All tmpfs/anonymous, no upgrade path needed. Verify Artifactory OSS accepts PG17 in migration-e2e.
10. **PG17 for persistent stacks**, separately, with a 1.9.0 upgrade note. Consider holding `docker-compose.yml` on 16 for one release and shipping the note first.
11. **Keycloak 26.7** plus the env renames, the `KC_HOSTNAME_STRICT_HTTPS` removal, and the volume reset. Verify with `scripts/sso-e2e/run.sh` locally — no workflow runs it on a PR.

Steps 1–4 are genuinely low-risk and could land this week, and they unblock the rest. If you only do one thing: re-cut `415f28be` as a true no-op and let it land first.

Happy to take these as separate PRs from you, or to split them out ourselves if you'd rather hand it off — your call, and either is fine. Thanks again for the thorough sweep.

<sub>Reviewed with two automated passes (breakage-risk + supply-chain/pinning); every version claim re-verified against the PR head and current main.</sub>

brandonrc:
@rockdrilla — I took you up on "pick these commits" and rebased the branch onto current `main`, reworked as 5 commits. Your authorship is on all of them via `Co-Authored-By`. **Your original 18 commits are recoverable at `3f764e89`** if you want anything back.

The branch was 5 weeks and 286 commits behind `main`, which caused most of what needed changing. It's now `MERGEABLE` (was `CONFLICTING`).

### Everything went to current latest — several of the targets had gone stale

| | branch had | now |
|---|---|---|
| PostgreSQL | 17 | **18** |
| Python | 3.13 | **3.14** |
| Node | 24 | **26** |
| Go | 1.26 | **1.27** |
| Rust | 1.97.1 | **1.98.0** |
| protoc | 35.1 | **36.1** |
| Trivy | 0.72.0 | **0.74.0** |
| Keycloak | 26.7 | **26.7.3** |
| Terraform | 1.15.8 | **1.16.0** |

Alpine 3.24, Debian trixie, OpenSearch 2.19.6 and Dependency-Track 4.14.3 are as you had them.

### What changed beyond the version strings

**Keycloak needed config, not just a tag.** The bump alone couldn't boot or authenticate: `KEYCLOAK_ADMIN`/`_PASSWORD` were renamed to `KC_BOOTSTRAP_ADMIN_USERNAME`/`_PASSWORD` in 26.0, so under the old names no bootstrap admin is created and `setup-keycloak.sh` fails at "Failed to get admin token" before any of our code runs. `KC_HOSTNAME_STRICT_HTTPS` was removed in the 26.x hostname-v2 rework and an unrecognised option aborts `start-dev`. And `keycloak_data` persisted a 24.0 dev-mode H2 store across two majors, so it's rolled to `keycloak_data_v26`.

**Rust went to 1.98.0, not 1.97.x.** `main` gained `rust-toolchain.toml` with `channel = "1.98.0"` (#3551) after you branched. That file wins: rustup reads it at build time and it outranks the toolchain the Dockerfile installs — so a 1.97 pin would have silently downloaded a second toolchain and built with 1.98.0 anyway, while looking authoritative. `check-toolchain-pin.sh` doesn't cover Dockerfiles, so nothing would have caught it.

**Your `decouple versions` commit wasn't a no-op** — it moved Rust 1.93.0→1.93.1 and protoc 29.3→29.6 inside a refactor. Easy to miss; I kept the ARG idea (it's the best structural change in the branch) and made the version moves explicit. Also switched `Dockerfile.backend` from `ENV` to `ARG`, since `--build-arg` can't override an `ENV` and it leaks into the final image.

**Dropped the grype commit.** `main` now builds grype **from source** at 0.117.0 with a commit pin and an `x/mod` override (#3352/#3465), so `ghcr.io/anchore/grype:v0.116.0` would have been a downgrade that re-introduced the CVEs that work cleared. Nothing lost — `main` is ahead of where the commit was going.

**Added:** protoc downloads are now checksum-verified per architecture (it was the largest unverified fetch in the build); the mirror matrix was deduped (3.19 and 3.23 both collapsed to 3.24) and is now an exact two-way match with what the tree consumes; and `docker-publish.yml`'s trivy retry guard picked up the bump, so it no longer warms 0.69.3 while creating 0.74.0.

### Three deliberate non-latest calls, each commented in-tree

- `migration-e2e`'s `artifactory-db` stays on **PostgreSQL 16** — it backs `jfrog/artifactory-oss`, and the supported version is JFrog's matrix to set, not ours. A newer major there fails as an Artifactory start-up error that reads like a migration bug.
- **Helm, OpenSearch, Dependency-Track** take latest *within their current major*. 4.x, 3.x and 5.x respectively are migrations with their own config and API changes, not version bumps — each deserves its own PR.

Also left your `scripts/native-tests/test-docker*.sh` alpine pins alone; you were right that those are registry test payloads, not toolchain.

### Verification

All 22 tags confirmed to resolve upstream. Every changed YAML parses. `docker compose config` passes on the four main stacks. And I built the risky download steps against `ubi9:9.8`: protoc checksum `OK`, `libprotoc 36.1`, `rustc 1.98.0`, `cargo 1.98.0`. (That build ran on aarch64, so the x86_64 protoc checksum is computed-and-correct but not build-exercised.)

### Before merge

**`mirror-images.yml` must be dispatched on `main` first** — it only fires on a weekly cron or manual dispatch, and until it runs, the five new `ci-mirror/*` tags don't exist and every job referencing them fails `manifest unknown`. That's why it's commit 1. I'll run it.

One thing CI can't tell us either way: `e2e.yml`, `mesh-e2e.yml` and `concurrency-e2e.yml` are dispatch/schedule-only, so a green check here doesn't cover Go, Debian, Helm, Terraform, or any of `scripts/sso-e2e/`. I'll dispatch those before merging.

Thanks for this — the sweep itself was thorough and the mirror matrix was already an exact match for what the tree consumed in both directions, which is more care than these usually get.

brandonrc:
Merged as `85d5e7e8` — thank you for this, @rockdrilla. Final shape relative to your original branch, for the record: Rust pin aligned to `rust-toolchain.toml` (1.98.0) with per-arch protoc checksums, the coordinated Alpine 3.24 / Debian trixie / Python 3.14 / Node 26 / Keycloak 26 set, the PostgreSQL 18 mount move to `/var/lib/postgresql` that the Smoke E2E caught, and the `ci-mirror` image set split out and merged first (#3639) so CI could pull them. The grype build stage was deliberately left as-is; Go 1.27 for it landed separately via #3579.

Issue #2524, opened 2026-07-14
PF-008 [P2]: Maintenance and large schema upgrades are not uniformly online and bounded


brandonrc:
# PF-008: Maintenance and large schema upgrades are not uniformly online and bounded

Priority: P2  
Affected area: artifact/OCI garbage collection, large index migrations

## Finding

Garbage collection has careful safety checks, but several candidate queries fetch every matching orphan into memory. The general artifact GC then processes keys serially and performs object-storage deletion while a database transaction and row locks remain open. Separately, million-row index migrations are created non-concurrently and explicitly block artifact writes.

These operations may be acceptable at small scale or during a maintenance window, but they do not satisfy a continuously available one-million-artifact operating model.

## Evidence

- Blob mark and sweep candidate queries have no `LIMIT`/cursor and call `fetch_all`: [storage_gc_service.rs](../backend/src/services/storage_gc_service.rs#L933), [storage_gc_service.rs](../backend/src/services/storage_gc_service.rs#L966).
- General orphan selection aggregates the entire orphan set into a vector: [storage_gc_service.rs](../backend/src/services/storage_gc_service.rs#L993).
- General GC iterates serially, opens a per-key transaction, rechecks under locks, then awaits `storage.delete` before committing: [storage_gc_service.rs](../backend/src/services/storage_gc_service.rs#L290).
- The default DB pool has 50 connections and a five-second acquire timeout, making long maintenance transactions consequential to foreground traffic: [config.rs](../backend/src/config.rs#L869), [db.rs](../backend/src/db.rs#L19).
- The lower-name migration states that its non-concurrent build takes `ACCESS EXCLUSIVE` and blocks uploads for millions of rows: [106_artifacts_lower_name_index.sql](../backend/migrations/106_artifacts_lower_name_index.sql#L27).
- Prefix/suffix index migrations give the same warning and require an out-of-band concurrent-index playbook for online installations: [108_artifacts_filename_index.sql](../backend/migrations/108_artifacts_filename_index.sql#L44), [110_artifacts_repo_path_pattern_ops.sql](../backend/migrations/110_artifacts_repo_path_pattern_ops.sql#L46).

## Impact at scale

- A deletion wave can create a large candidate vector and a very long maintenance run.
- Awaiting cloud storage while holding row locks and a pooled connection extends lock/transaction lifetime by network latency and retries.
- Serial object deletion underutilizes safe provider batch APIs yet can still monopolize one pool connection for a long time.
[...]

Issue #2516, opened 2026-07-14
Perf Epic: Million-artifact readiness — critical requirements, release gates & test plan


brandonrc:
# Critical performance requirements and audit matrix

## Workload model

“One million artifacts” is not enough by itself to define readiness. The acceptance workload should include:

- 1,000,000 live catalog rows, including at least one 500,000-row repository and a virtual repository spanning several large members.
- 10–100 million retained download events, because an artifact registry normally has far more downloads than artifacts.
- A remote repository with 1,000,000 cached artifacts. The current cache layout commonly has a content object and metadata sidecar, so this means roughly two million object keys.
- 100,000 distinct package names and high-version-count packages to exercise PyPI roots, Composer roots, Helm indexes, RPM/Conda metadata, Maven components, and OCI tags.
- A size distribution from 1 KiB through 1 MiB and 100 MiB to 5–10 GiB, with concurrent large transfers.
- Read-heavy steady state, burst publishing, proxy cold-cache fan-in, scan-on-upload, replication, lifecycle, and GC running at the same time.

Test both filesystem and one production object-store backend. Filesystem directory walking and cloud object-list latency have very different failure modes.

## Release gates

Targets below are proposed critical requirements, not measurements of the current code. Final values can be adjusted to the production topology, but the boundedness properties should not be relaxed.

| Requirement | Proposed acceptance gate |
|---|---|
| Exact artifact resolution | p95 ≤ 100 ms and p99 ≤ 250 ms at 1M rows; query plan is an index lookup and does not vary materially with repository size. |
| Catalog browsing | First and deep cursor pages p95 ≤ 250 ms for 100 items; work and rows examined are proportional to page size, not page number or repository size. |
| Search | p95 ≤ 500 ms; substring/full-text search uses a matching index or OpenSearch; no million-row scan for a 100-item page. |
| Remote cache browsing | No bucket-wide/prefix-wide list in an interactive request; p95 and memory are independent of total cache keys. |
| Whole-repository protocol metadata | Requests serve a prebuilt immutable revision or a cached object; foreground work is O(response bytes), not O(catalog rows + regeneration). Concurrent misses coalesce. |
| Upload memory | RSS increase ≤ 16 MiB per large upload, excluding explicitly bounded shared buffers; no route materializes a full blob. |
| Download/proxy memory | RSS increase remains bounded per transfer; slow clients and slow storage apply backpressure rather than queueing unbounded chunks. |
[...]

PR #2887, opened 2026-07-24
perf(search): back artifact full-text search with a GIN-indexed search_vector (#2871)

### LINKED ISSUE #2871: PF: artifact full-text search recomputes to_tsvector per row (no GIN-indexed tsvector)
Part of the million-artifact perf epic #2516; online-upgrade path relates to #2524 (PF-008).

**Measured (PTF `search-at-scale` profile):** artifact full-text search p95 **272ms → 1.09s → 3.63s** at 10k/100k/500k artifacts, pegging Postgres ~11–17 cores. Backend itself stays light — this is a pure query-plan problem.

**Root cause (EXPLAIN ANALYZE @ 517k rows):** `SearchService::search` (both /search/quick and /search/advanced) runs an **inline functional predicate** `to_tsvector('english', name || path || version) @@ to_tsquery(...)` with **no stored/indexed tsvector column**. Every search is a **Parallel Seq Scan recomputing to_tsvector per row**, plus a **second full scan for the pagination COUNT**.

**Fix:** add a **trigger-maintained** `search_vector tsvector` column on `artifacts` (nullable → instant catalog change, no table rewrite) + a **partial GIN index** (`WHERE is_deleted=false`) + rewrite the two search queries to match the stored column (semantics/order preserved; count becomes an index-backed Bitmap Index Scan). Migration adds column + trigger + in-migration backfill; large-table CONCURRENTLY/batched online-upgrade path documented (per migration 173 precedent).

**Validation:** DTF parity tier (identical results/ordering/exact total, prefix match, metachar→empty-not-500, fresh upload immediately searchable) + PTF `search-at-scale` re-run vs the committed baseline — **target ~3.6s → sub-100ms @500k**, cores collapse, EXPLAIN shows Bitmap Index Scan.

brandonrc:
Closes #2871

## Summary

Artifact full-text search (`/search/quick` and `/search/advanced`, both served
by `SearchService::search`) filtered on an **inline functional predicate**:

```sql
to_tsvector('english', a.name || ' ' || a.path || ' ' || COALESCE(a.version, ''))
  @@ to_tsquery('english', $1)
```

Nothing on `artifacts` matched that expression, so the planner had no choice but
a Parallel Seq Scan that recomputes `to_tsvector(...)` for every live row on
every request -- **twice**, because the exact-count query for pagination runs
the same predicate a second time. EXPLAIN at ~517k rows showed this pegging
Postgres at 11-17 cores, with p95 climbing 272ms -> 1.09s -> 3.63s at
10k / 100k / 500k artifacts (perf epic #2516).

This PR materializes the exact same tsvector into a stored, GIN-indexed column:

- **Migration `176_artifacts_search_vector.sql`**
  - `ALTER TABLE artifacts ADD COLUMN IF NOT EXISTS search_vector tsvector`
    (nullable, no default -> catalog-only change, no table rewrite).
  - `ak_artifacts_search_vector_update()` trigger function + a
    `BEFORE INSERT OR UPDATE OF name, path, version` trigger that maintains the
    column from the identical expression the inline predicate used.
  - In-migration backfill of existing rows.
  - Partial GIN index `idx_artifacts_search_vector_gin ... WHERE is_deleted = false`
    matching every caller's filter.
  - A documented `CREATE INDEX CONCURRENTLY` / batched-backfill online-upgrade
    path in a comment for very large tables (follows the migration 173
    precedent; relates to #2524 PF-008).
- **`backend/src/services/search_service.rs`** -- both the item query
  (`execute_search`) and the pagination `count_results` now filter on
  `a.search_vector @@ to_tsquery('english', $1)` via one shared
  `SEARCH_VECTOR_MATCH` constant, so the two can never diverge. The planner now
  satisfies the predicate with a Bitmap Index Scan and the COUNT collapses to
  the same index-backed bitmap instead of a second full scan.

Result semantics and ordering are unchanged: the stored vector is byte-for-byte
the vector the inline predicate used to compute per row.

- [x] N/A -- this is a `perf/*` change (behavior preserved). Unit tests pin the
      rewritten predicate; a deployment-test tier (search-gin-2871, tracked
      separately for the artifact-keeper-test repo) validates identical
      results/ordering, exact count, prefix match, metachar-safety, and
      fresh-upload searchability against a live stack.

## Test Checklist
- [x] Unit tests added/updated (`SEARCH_VECTOR_MATCH` contract: targets the
      stored column, short-circuits on NULL `$1`, uses only `$1`, balanced +
      `'english'` config)
- [x] E2E tests added/updated (deployment-test tier search-gin-2871, batched
      into a later artifact-keeper-test MR)
- [x] Manually tested locally (isolated stack: migration applies; column,
      trigger, and partial GIN index present; quick/advanced return identical
      ordered results; exact total count; prefix match; metacharacter query ->
      HTTP 200 empty, not 500; unrelated artifacts excluded; fresh upload
      immediately searchable with its search_vector populated by the trigger;
      `EXPLAIN` confirms the GIN index is used for the predicate)
- [x] No regressions in existing tests (`cargo fmt --check`, `cargo clippy
      --lib -D warnings`, lib unit tests, `jscpd` duplication all clean)

## API Changes
- [ ] OpenAPI spec validates
- [x] Migration is reversible (if applicable) -- forward-only catalog change
      (column + trigger + partial index); all statements are idempotent
      (`IF NOT EXISTS` / `CREATE OR REPLACE`). No request/response contract
      change: search inputs and outputs are byte-for-byte identical.
- [ ] N/A - no API changes

### DIFF
--- a/backend/migrations/176_artifacts_search_vector.sql
+++ b/backend/migrations/176_artifacts_search_vector.sql
@@ -1,0 +1,107 @@
+-- Stored, trigger-maintained full-text search vector for artifacts
+-- (PF-009 / #2871, part of the million-artifact perf epic #2516).
+--
+-- SearchService::search (both /search/quick and /search/advanced) filtered on
+-- an INLINE functional predicate:
+--
+--   to_tsvector('english', name || ' ' || path || ' ' || COALESCE(version, ''))
+--     @@ to_tsquery('english', $1)
+--
+-- Nothing on `artifacts` matched that expression, so the planner had no choice
+-- but a Parallel Seq Scan that recomputes to_tsvector('english', ...) for every
+-- live row on every request -- twice, because the exact-count query for
+-- pagination runs the same predicate a second time. EXPLAIN at ~517k rows
+-- showed this pegging Postgres at 11-17 cores with p95 climbing 272ms -> 1.09s
+-- -> 3.63s at 10k / 100k / 500k artifacts.
+--
+-- Fix: materialize the exact same tsvector into a stored `search_vector` column
+-- maintained by a BEFORE INSERT/UPDATE trigger, and back it with a partial GIN
+-- index (WHERE is_deleted = false, matching every caller of this path). The two
+-- search queries then filter on `a.search_vector @@ to_tsquery('english', $1)`,
+-- which the planner satisfies with a Bitmap Index Scan; the COUNT collapses to
+-- the same index-backed bitmap instead of a second full scan. Result semantics
+-- and ordering are unchanged: the stored vector is byte-for-byte the vector the
+-- inline predicate used to compute per row.
+--
+-- The column is nullable and carries no default, so `ADD COLUMN` is a
+-- catalog-only change (no table rewrite, no long ACCESS EXCLUSIVE hold). The
+-- backfill and the (non-concurrent) index build below are the only heavy steps.
+
+-- 1. Catalog-only column add (instant; no rewrite because it is nullable with
+--    no default).
+ALTER TABLE artifacts ADD COLUMN IF NOT EXISTS search_vector tsvector;
+
+-- 2. Trigger function: recompute the stored vector from exactly the same
+--    expression the old inline predicate used. Kept in one place so the stored
+--    column can never drift from the query semantics.
+CREATE OR REPLACE FUNCTION ak_artifacts_search_vector_update() RETURNS trigger AS $$
+BEGIN
+    NEW.search_vector := to_tsvector(
+        'english',
+        NEW.name || ' ' || NEW.path || ' ' || COALESCE(NEW.version, '')
+    );
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+-- Fire only when a component of the vector actually changes (plus every
+-- INSERT). Soft-delete (is_deleted) and download/updated_at churn do not touch
+-- name/path/version, so they never pay the recompute -- and a soft-deleted row
+-- drops out of the partial index regardless.
+DROP TRIGGER IF EXISTS ak_artifacts_search_vector_trg ON artifacts;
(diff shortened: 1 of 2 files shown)


brandonrc:
## Retest results — PTF `search-at-scale`, fix branch vs `1.6.2-rc` baseline

The PR body has the *before* curve; here's the measured *after* — this is the PTF delta that validates the fix (first PTF-driven perf fix).

**App-level, @500k artifacts (median-of-≥3, warm):**

| metric @500k | 1.6.2-rc (before) | #2871 GIN (after) | change |
|---|---|---|---|
| app p95 | 2831 ms | **790 ms** | 3.6× faster |
| app p99 | 3772 ms | 911 ms | 4.1× |
| tail p99 (under concurrency) | 4220 ms | 980 ms | 4.3× |
| p99/p50 (latency cliff) | 4.44× | **1.53×** | cliff flattened |

Everything at 500k is now **under 1 s**, and the latency curve is **flat across 10k / 100k / 500k** instead of the old 272 ms → 1.09 s → 3.63 s climb.

**DB-level EXPLAIN @~500k rows (post-fix):**
- Selective term → **0.089 ms** via Bitmap Index Scan on `idx_artifacts_search_vector_gin` (was a Parallel Seq Scan recomputing `to_tsvector` per row, ×2 for the COUNT).
- Broad-prefix term matching ~250k rows → COUNT collapses to **75 ms**.

**Parity:** 0 result/ordering drift across 12 query shapes (identical results, exact counts, prefix match, metachar-safety, fresh-upload searchability).

**What's left:** the residual ~790 ms app p95 is the broad-term stress mix sorting/serializing large match sets (`COUNT` + `ORDER BY`), **not** tsvector recompute — that cost is eliminated. The old bottleneck was Postgres query CPU (~11–17 cores), not the pool or backend.

Measured with the `search-at-scale` profile in `artifact-keeper-test` (`deploy-test/perf/`).

PR #2897, opened 2026-07-24
ci: move portable control, smoke, and AMD64 image jobs to hosted runners

### LINKED ISSUE #2895: CI: portable control and smoke jobs plus AMD64 image builds depend on self-hosted runners, so a runner outage deadlocks CI
During the 2026-07-24 ARC outage, release-branch CI run 30055109653 waited about eleven hours for ak-e2e-runners, and after cancellation its `if: always()` CI Complete job queued on ak-ci-runners for another 4h43m, keeping the cancelled run non-terminal and holding the branch concurrency group. Replacement run 30089358441 sat pending with zero jobs the whole time. Docker Publish AMD64 builds (backend, OpenSCAP, scanner adapter) also queued behind the same pool while hosted ARM64 builds completed.

Portable jobs must not depend on ARC:

- Move `detect-changes` and `ci-complete` to ubuntu-latest
- Move `smoke-e2e` to ubuntu-24.04 (scheduled-tests already proves hosted compose smoke works)
- Move the AMD64 matrices in docker-publish.yml from ak-ci-runners to ubuntu-24.04
- Preserve all existing required job display names (branch protection)
- Ensure ci-complete reaches a terminal result after upstream failure, cancellation, and legitimate skips
- Make smoke wait for its required prerequisites so a known-red run does not consume E2E capacity
- Add .github/actionlint.yaml declaring the custom runner labels

Rollback note: if a Docker build exceeds hosted disk or memory limits, move only that build to a dedicated image-builder scale set, not back to the shared Rust control path.

Corresponds to PR AK-CI-1 in docs/plans/2026-07-24-ci-pipeline-reliability-implementation-plan.md.

brandonrc:
## Summary

Moves the portable, non-compute CI control jobs and the AMD64 container image
builds off the self-hosted ARC runners and onto GitHub-hosted runners, so a
self-hosted runner outage can no longer stall change detection, smoke tests,
image publishing, or the required CI Complete gate. Implements PR AK-CI-1 from
the CI pipeline reliability plan.

Changes in `.github/workflows/`:

- `ci.yml`
  - `detect-changes` moved to `ubuntu-latest` (portable path-filter job).
  - `ci-complete` moved to `ubuntu-latest`. The display name "CI Complete"
    is unchanged (branch protection). Kept `if: always()`. Rewrote the status
    step so it always reaches a terminal result: required jobs (check-rust,
    test-backend-unit, shell-tests, security-audit) must be success or the gate
    fails, and a non-run such as a skip or cancellation on a required job is an
    explicit failure rather than a silent pass; conditionally-required jobs
    (smoke, integration, image builds) pass on a legitimate skip but fail on a
    failure or cancellation so an aborted upstream job is never masked. Job
    results are now passed via env vars instead of inlined into the shell body.
  - `smoke-e2e` moved to `ubuntu-24.04`. This job runs `docker compose
    -f docker-compose.test.yml --profile smoke`, which `scheduled-tests.yml`
    already runs on hosted runners. Added `check-rust` to `needs` and tightened
    the `if` so smoke only launches when Check Rust, Backend Unit Tests, and the
    backend image build all succeeded, so a known-red Rust or unit-test run no
    longer consumes E2E capacity.

- `docker-publish.yml`
  - AMD64 matrix runner changed from `ak-ci-runners` to `ubuntu-24.04` for the
    backend, OpenSCAP, and scanner-adapter build jobs. ARM64
    (`ubuntu-24.04-arm`) is unchanged, and the disabled Alpine job (`if: false`)
    is left untouched.
  - Added a "Free disk space" step before the backend AMD64 build only (gated
    on `matrix.platform == 'linux/amd64'`) because the backend image compiles
    Rust inside Docker and can strain a hosted runner's ~14GB disk. The build
    already uses registry buildcache (`cache-from`/`cache-to type=registry`),
    which keeps compile time reasonable on hosted runners. If hosted disk or
    memory limits are still exceeded, the fallback per the plan is a dedicated
    image-builder scale set, not a return to the shared Rust runner path.

- Added `.github/actionlint.yaml` declaring the self-hosted runner labels
  (`ak-ci-runners`, `ak-e2e-runners`, `ak-beefy-runners`, `ak-docker-runners`)
  so actionlint accepts them.

The compute-heavy Rust jobs (Check Rust, Backend Unit Tests, Code Coverage,
Backend Integration Tests, the two image build jobs, Windows cross-build) remain
on `ak-ci-runners` and are out of scope for this PR. All required check display
names are preserved: "Check Rust", "Backend Unit Tests", "CI Complete".

Validation run locally:

- `git diff --check`: clean.
- YAML parse (`python3 -c "import yaml; yaml.safe_load(...)"`) on both workflows
  and the actionlint config: all parse.
- `actionlint` (v1.7.12): no unknown-runner or label errors; the self-hosted
  labels are accepted via the new config. Total findings dropped from 105 to 85
  because the ci-complete rewrite (env vars, quoting, `set -euo pipefail`)
  cleared 20 pre-existing shellcheck warnings and introduced none. The remaining
  findings are all pre-existing shellcheck style/info notes and the disabled
  Alpine job's `if: false`, none in the lines changed here.

Closes #2895

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc:
Updated the branch with main (was 79 commits behind) and resolved the `ci.yml` conflict: kept this PR's env-hardened `require_success`/`allow_skip` rewrite of `ci-complete` and folded in main's docs-only gate semantics (`docs_only` detection via a new `CODE_CHANGED` env var, docs-only skips allowed for Tier 1 jobs and security-audit).

The previous 📊 Code Coverage failure was a stale-merge-base artifact, not a real gap: this PR touches no `.rs` files, but the gate diffed against a 79-commit-old merge base and attributed main's code ("12% (3/24 new lines)") to this PR. The refresh should clear it.

PR #2900, opened 2026-07-24
fix(debian): pass strict mode at the two test call sites missed by the relpath signature change

### LINKED ISSUE #2898: main does not compile: debian.rs test call sites missed the normalized_debian_relpath signature change
Main is red since the #2770 merge (d095d0ce). normalized_debian_relpath gained a third parameter (allow_encoded_separators: bool, backend/src/api/handlers/debian.rs:392) but two test call sites still pass 2 arguments:

- backend/src/api/handlers/debian.rs:3614 (encoded-separator rejection loop)
- backend/src/api/handlers/debian.rs:3620 (legitimate-path assertion)

cargo clippy --workspace --all-targets fails with E0061 x2 and Check Rust is red on every main push since run 30111206190 (and on every PR based on current main, which currently blocks all open PRs).

Cause: #2838 (f1c0ef45) added these tests against the 2-argument signature while #2770 changed the signature; both merged without a combined build. Fix: pass false (strict mode) as the third argument at both call sites. The tests assert rejection of encoded separators and acceptance of a clean path, which is exactly strict-mode behavior.

brandonrc:
## Summary

Fixes the `main` compile break in the backend. Two `#[test]` call sites in `backend/src/api/handlers/debian.rs` still called `normalized_debian_relpath` with two arguments after its signature grew a third.

This is a two-PR semantic collision:

- #2770 changed `normalized_debian_relpath` to take a third parameter, `allow_encoded_separators: bool` (base_url, relative, allow_encoded_separators), so callers can opt into or out of accepting encoded `%2f` / `%5c` separators.
- #2838 landed tests that were written against the old two-argument signature.

Both merged cleanly on their own but together left two test call sites (in `test_normalized_relpath_rejects_encoded_separators`) calling the function with the old arity, which fails to compile. Because the break is in test code under `--all-targets`, it blocks Check Rust for every other open PR.

The fix passes `false` (strict mode) at both call sites, which is the correct value: both assertions verify that encoded separators are rejected, so the test must exercise the strict path.

The updated tests (`test_normalized_relpath_rejects_encoded_separators`) now compile and assert strict-mode rejection of encoded `%2f`/`%5c` separators, and pass on this branch.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Closes #2898

### DIFF
--- a/backend/src/api/handlers/debian.rs
+++ b/backend/src/api/handlers/debian.rs
@@ -3611,14 +3611,18 @@ mod tests {
             "pool/main/m/mypkg%5C..%5Cevil_1_amd64.deb",
         ] {
             assert!(
-                normalized_debian_relpath(TEST_BASE, bad).is_err(),
+                normalized_debian_relpath(TEST_BASE, bad, false).is_err(),
                 "should reject encoded separator: {bad:?}"
             );
         }
         // Legitimate request with real separators still resolves.
         assert_eq!(
-            normalized_debian_relpath(TEST_BASE, "dists/bookworm/main/binary-amd64/Packages.gz")
-                .unwrap(),
+            normalized_debian_relpath(
+                TEST_BASE,
+                "dists/bookworm/main/binary-amd64/Packages.gz",
+                false
+            )
+            .unwrap(),
             "dists/bookworm/main/binary-amd64/Packages.gz"
         );
     }


brandonrc:
Closing as duplicate of #2893, which already fixes the same lib-test compile break (#2898). #2893 was opened first. No action needed here.

Issue #2084, opened 2026-06-30
v1.2.3: backend and dtrack-init fail to start — runtime image dropped /bin/sh but their compose entrypoints require it


nicexe2e4:
## Summary

On a clean `v1.2.3`, `docker compose up` cannot start the `backend` (and `dtrack-init`) container:

```
OCI runtime create failed: runc create failed: unable to start container process:
error during container init: exec: "/bin/sh": stat /bin/sh: no such file or directory
```

The hardened runtime image from #2059 no longer contains `/bin/sh` (or `/bin/bash`), but the shipped `docker-compose.yml` starts these services through a `/bin/sh` entrypoint.

## Reproduction (published release image, no local build)

```console
$ docker run --rm --entrypoint /bin/sh ghcr.io/artifact-keeper/artifact-keeper-backend:1.2.3 -c 'true'
... exec: "/bin/sh": stat /bin/sh: no such file or directory
```

The image's own `CMD` is fine — `--entrypoint /usr/local/bin/artifact-keeper` boots and reaches normal config validation (`Error: Config("JWT_SECRET not set")`). It's only the shell entrypoints that break.

## Root cause

#2059 ("chore(docker): harden backend runtime image") removed the shells (`bash`, plus the `/bin/sh` symlink). But two services run this image **through a shell**:

- `backend` → `entrypoint: ["/bin/sh", "-c", <wait-for-DT-key script>]`
- `dtrack-init` → `entrypoint: ["/bin/sh", "/init-dtrack.sh"]` (and `docker/init-dtrack.sh` is `#!/bin/sh`)

## Impact

- Default `docker compose up` does not start out of the box.
- The `--profile dtrack` path is also blocked: `dtrack-init` can't run `init-dtrack.sh` to provision the API key, so even overriding just the backend entrypoint doesn't fully recover Dependency-Track.

## Question

@brandonrc — was dropping `/bin/sh` in #2059 intentional (the compose entrypoints to be reworked to not need a shell), or an oversight?

## Happy to help

If unintended, I'd gladly open a PR that **keeps** the no-shell hardening rather than restoring a shell — e.g. drop the `/bin/sh -c` wrappers and rely on the image `CMD` (`["artifact-keeper"]`), move the DT-API-key wait into the binary via a `DEPENDENCY_TRACK_API_KEY_FILE` env (mirroring existing `*_FILE` patterns), and run `dtrack-init` from a small shell-bearing image. Just say the preferred direction.

PR #2909, opened 2026-07-25
fix(docker): get docker-compose.local-dev.yml working again

### LINKED ISSUE #2911: docker-compose.local-dev.yml is broken end to end: dtrack-init reuses the shell-less hardened image, dev protoc lacks well-known types
Reported and fixed by @Mo0rBy in PR #2909; issue filed for tracking per repo convention.

The local-dev compose stack cannot start:

- dtrack-init reuses the hardened production backend image, which has no /bin/sh (#2059), but its init script needs sh, curl, and jq. backend and web transitively depend on dtrack-init completing, so the whole stack blocks. Same bug #2126 already fixed in docker-compose.yml; docker-compose.local-dev.yml carried an independent, never-updated copy of the pattern.
- docker/Dockerfile.backend.dev installs protoc via Debian protobuf-compiler, which lacks the google/protobuf well-known types the build imports; production vendors the official protoc release which bundles them.
- The default JWT_SECRET was 31 characters, one below the backend's hard 32-character minimum, so startup validation always failed.
- The dev compile gets OOM-killed under memory-constrained Podman VMs without scoped build tuning.

Fix in #2909 also generalizes the compose regression test to scan every top-level docker-compose*.yml so this drift class cannot silently recur.

Mo0rBy:
## Summary
- `docker-compose.local-dev.yml` was broken end to end for `backend`/`web` (confirmed via `podman compose -f docker-compose.local-dev.yml up`). Root causes:
  - `dtrack-init` reused the hardened production backend image, which has no `/bin/sh` (#2059), but its init script needs sh+curl+jq. Since `backend`/`web` transitively depend on `dtrack-init` completing, this blocked the whole stack. This is the exact bug #2126 fixed in `docker-compose.yml` — `docker-compose.local-dev.yml` carried an independent, never-updated copy of the same broken pattern. Fixed by switching `dtrack-init` to `alpine:3.23`, mirroring `docker-compose.yml`.
  - Simplified the `backend` entrypoint to use the `DEPENDENCY_TRACK_API_KEY_FILE` convention (added in #2126) instead of a shell wrapper reading the key file.
  - `docker/Dockerfile.backend.dev` installed `protoc` via Debian's `protobuf-compiler`, which doesn't ship the `google/protobuf/*.proto` well-known types (unlike production's Dockerfile, which vendors the official protoc release). Added `libprotobuf-dev`, which supplies them at protoc's default include path.
  - Backend's dev compile was getting OOM-killed under memory-constrained Podman VMs (this is a large workspace: wasmtime, git2, sqlx, tonic all in one crate, built with full debug info). Added `CARGO_PROFILE_DEV_DEBUG=line-tables-only` and `CARGO_BUILD_JOBS=3` to the backend service, scoped to this compose file only.
  - The default `JWT_SECRET` in the compose file was 31 characters, one short of the backend's 32-character minimum, so it always failed config validation. Padded it.
- The existing #2084 regression test (`shipped_compose_does_not_run_hardened_image_through_a_shell`) only parsed `docker-compose.yml`, which is exactly why it didn't catch `docker-compose.local-dev.yml` drifting back into the same bug. Generalized it to discover every `docker-compose*.yml` at the repo root and check each one's `dtrack-init` service, so a future compose file can't reintroduce this unnoticed.
- Audited every other Dockerfile and compose file in the repo for both bug patterns (hardened image run through a shell; protoc missing well-known types from a split-off package). Nothing else was affected — details in the third commit's message.
- Added a `CLAUDE.md` rule: changing one Dockerfile/compose file's assumptions about the runtime image must prompt checking every sibling file for the same assumption, since that's exactly how this regressed.

Not part of this diff, but required for other local-dev users: a real `AK_WEBHOOK_SECRET_KEY` must be set via a gitignored `.env` at the repo root (compose only auto-loads `.env`, not `.env.local-dev`, and the default is intentionally invalid). Podman machine users on memory-constrained hosts may also need `podman machine set --memory 11264` or similar.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Closes #2911

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,6 +119,36 @@ gh workflow run e2e.yml -f profile=all -f include_stress=true
 
 Rust 1.75+: Follow standard conventions
 
+## Docker & Compose Files
+
+This repo ships multiple Dockerfiles (`docker/Dockerfile.backend`,
+`docker/Dockerfile.backend.dev`, `docker/Dockerfile.backend.alpine`, plus
+scanner/mock images) and multiple compose files
+(`docker-compose.yml`, `docker-compose.local-dev.yml`,
+`docker-compose.*-e2e.yml`, `docker-compose.test.yml`, etc.) that each
+independently encode assumptions about what the backend image provides (e.g.
+whether it has a shell, which packages install which files). These have
+drifted out of sync before: #2126 fixed production's `docker-compose.yml`
+after the hardened runtime image (#2059) dropped `/bin/sh`, but
+`docker-compose.local-dev.yml` carried an independent copy of the same
+`dtrack-init` pattern that nobody updated, so it silently reintroduced the
+exact bug #2126 had already fixed.
+
+**When changing any Dockerfile or compose file, check every other one for the
+same assumption.** Concretely:
+- If a Dockerfile's runtime image gains or loses a shell, package manager, or
+  any binary (e.g. `/bin/sh`, `curl`, `jq`, `protoc`), grep all other
+  Dockerfiles and compose files for the same pattern before assuming only the
+  one you're editing is affected.
+- If a compose file's `dtrack-init` (or any init container) is fixed to stop
+  reusing the hardened backend image through a shell, check every other
+  compose file for a service doing the same thing.
+- Prefer a regression test that scans *all* matching files (e.g. every
+  `docker-compose*.yml` at the repo root) over one hardcoded to the single
+  file you just fixed — see
+  `shipped_compose_does_not_run_hardened_image_through_a_shell` in
+  `backend/src/config.rs` for the pattern.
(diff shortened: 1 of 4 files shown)


brandonrc:
@Mo0rBy  awesome stuff! Thanks for this PR. There are alot of docker files all over the place and it is hard to update them all. I might be removing the e2e docker files and keep the new Test Framework system I am implementing in artifact-keeper-test repo. For now this is great and appreciate the time you put in:)

Mo0rBy:
All good @brandonrc , happy to help!

I found this project and I like the look of it. Very disappointed that there are no other open source solutions for package registries, at least none that cover a large range of package types.

PR #2915, opened 2026-07-26
Fix conda-forge and Hugging Face pull-through proxies; preserve Content-Length on compressed upstreams

### LINKED ISSUE #2914: Proxy pull-through broken for conda-forge and Hugging Face; compressed upstreams lose Content-Length
## Summary

Several pull-through proxy flows do not work against real upstreams. Found while validating a developer onboarding path (pip, conda/pixi, Hugging Face, NuGet all pointed at Artifact Keeper). pip worked; conda and Hugging Face did not, and both trace to backend defects rather than client configuration.

### 1. conda-forge discovery through a remote conda repo returns an empty index

`serve_repodata` proxies a Remote repo's `repodata.json` with an 8 MiB cap (`DEFAULT_METADATA_MAX_BYTES`). conda-forge's real `repodata.json` is ~24-30 MB, so the capped fetch fails and the code silently falls through to `build_repodata` (the DB-only index, empty for a fresh remote repo), returning a valid 200 with zero packages. A solver then finds no candidates. Separately, the CEP-16 sharded-repodata endpoint serves a valid-but-empty 200 for Remote repos (AK never implemented upstream shard fetch), which pixi treats as authoritative and never falls back from.

### 2. Hugging Face model downloads fail for any namespaced model

Three layers:
- The router used single-segment `:model_id` placeholders, so any `org/name` model 404s.
- `model_info` was DB-only and never proxied the upstream HF API, so an uncached model had no file list.
- The `/api/models/{id}/revision/{rev}` route the client needs did not exist.

### 3. Compressed upstream responses lose Content-Length (affects every format)

The `opensearch` dependency enables `reqwest`'s `gzip` feature. Cargo unifies features across the binary, so every `reqwest::Client` silently switches into content-negotiation: it sends `Accept-Encoding: gzip`, transparently decodes compressed responses, and strips `Content-Encoding` and `Content-Length` from the response headers before the proxy captures them. Any upstream or CDN that compresses above a size threshold then reaches the client with no Content-Length. This broke `huggingface_hub`'s HEAD metadata check (`FileMetadataError: Distant resource does not have a Content-Length`) and is a latent correctness issue for every proxied format.

## Fix

1. Serve conda Remote repodata with the shared `LARGE_METADATA_MAX_BYTES` tier (128 MiB, already used by debian/npm/maven/pypi/composer) and propagate real upstream fetch failures instead of masking them as an empty index. 404 the sharded endpoint for Remote repos so clients fall back to `repodata.json`.
2. Accept namespaced (`org/name`) HF model ids on all routes, add the `/api/models/{id}/revision/{rev}` routes, and make `model_info` a real upstream pull-through for Remote repos.
3. Forward the upstream ETag and inject `X-Repo-Commit` on HF resolve responses, and disable reqwest auto-decompression (`no_gzip`/`no_brotli`/`no_deflate`/`no_zstd`) on the base client so upstream serves identity and Content-Length survives for all formats.

## Verification

All four developer flows (pip, conda/pixi via a proxied conda-forge, `hf download` of a namespaced model, `dotnet restore`) verified end to end through a locally built image. `hf download` pulls the full model (30 files, ~1 GB) through the proxy.

## Scope

Backend only, targeting `main`. Independent of the security-gate enforcement work in #2912/#2913.

brandonrc:
## Summary

Makes the conda-forge and Hugging Face pull-through proxies work against real upstreams, and fixes a binary-wide reqwest bug that stripped Content-Length from compressed upstream responses. Found while validating a developer onboarding path (pip, conda/pixi, Hugging Face, NuGet all pointed at Artifact Keeper). Closes #2914.

1. HF namespaced (org/name) model routing.
2. conda Remote repodata served in full (128 MiB tier, budget-reserved) with real upstream failures propagated instead of masked as an empty index; sharded-repodata endpoint 404s for non-hosted repos so clients fall back to repodata.json.
3. HF model-info and tree listings are now real upstream pull-throughs for Remote repos.
4. Upstream `ETag`, `Content-Encoding` and `X-Repo-Commit` are forwarded on resolve responses (huggingface_hub's HEAD metadata check requires an ETag, a size, and a commit).
5. Disable reqwest auto-decompression (`no_gzip`/`no_brotli`/`no_deflate`/`no_zstd`) on the base HTTP client. The `opensearch` crate enables reqwest's `gzip` feature, which Cargo unifies across the binary, silently switching every client into content-negotiation that decodes compressed responses and drops Content-Encoding and Content-Length before the proxy captures them.
6. conda Remote **package** downloads stream instead of buffering at the 8 MiB metadata tier.

## Content-Encoding is now carried end to end

Turning off reqwest's decompression removed the only component in the process that handled `Content-Encoding`, and nothing took over: a coded body was served and cached labelled as identity. Before, such a response reached the client with correct bytes and no Content-Length; after commit 5 alone it would have reached the client with *wrong bytes* and a correct Content-Length. Object stores return a stored `Content-Encoding` regardless of the request's `Accept-Encoding`, so this needed no hostile upstream — an S3-hosted channel mirror or `nginx gzip_static always` was enough, and conda would have re-gzipped on top of it.

The coding is now captured on both the buffered and streaming paths, persisted in the cache sidecar (`CacheMetadata::content_encoding`, `#[serde(default)]` for wire-compat — absent means identity, which is correct for entries written while reqwest was still decoding), and replayed on cache hits and stale-if-error serves. The base client also advertises `Accept-Encoding: identity` explicitly, because sending *no* `Accept-Encoding` means "any content coding is acceptable" per RFC 9110 §12.5.3 — the opposite of what a client that cannot decode should say.

`fetch_artifact_capped` keeps its existing `(Bytes, Option<String>)` shape so the ~8 callers that parse bodies themselves are untouched; a `_with_encoding` sibling serves the handlers that forward bytes onward.

## Behavior changes worth a reviewer's attention

- **conda `repodata.json` / `channeldata.json` now fail closed.** A real upstream failure surfaces (503 on a 5xx) instead of a DB-only empty index with 200. A single failing channel can now fail a solve that previously "succeeded" against a silently empty index. The proxy's own RFC 5861 stale-if-error window still covers transient upstream errors.
- **Sharded-repodata and JLAP 404 for all non-hosted repo types**, not just Remote. Virtual was the real fail-open (a virtual repo owns no artifacts, so it served an authoritative empty CEP-16 index that a client will not fall back from). Extending it to Staging is conservative rather than a bug fix — Staging is hosted, so its shard index would have been correct; the guard is written to fail closed on an unreadable `repo_type` and the rationale is in its doc comment so a reviewer can dissent knowingly.
- **`GET /api/models/{id}` `"sha"` changed for hosted repos** from the artifact's `checksum_sha256` to a stable per-`(model_id, revision)` commit id. `snapshot_download` takes the snapshot directory from model-info's `sha` and files each blob under the commit its own resolve reported, so the two must agree or the download produces an empty directory. The old value was per-file and 64 chars, and was never a commit identifier. No existing test asserted it.
- Upstream `ETag` is never synthesized for Remote artifacts — inventing one would let clients cache proxied bytes under an identity we made up. Hosted/Virtual local serves use the stored `checksum_sha256`.

## Verification

All four developer flows verified end to end through a locally built image: pip download, `pixi install` resolving conda-forge + PyPI, `hf download` of a namespaced model (pulls the full model, 30 files, ~1 GB, through the proxy), and `dotnet restore`.

Targets `main`. **Expect merge conflicts with #2913 (merged after this branch was cut)**: `proxy_helpers.rs`'s `RepoInfo` gained `curation_enabled`/`curation_default_action`, and the conda/HF test fixtures gained fields. #2913's curation enforcement is PyPI-only, so no conda/HF path needs a curation gate.

Each fix has a test that fails on the pre-fix code. Two were verified by reverting the fix and observing the failure:

- conda package download above the 8 MiB tier — fails with `502 … Upstream metadata response exceeded the 8388608-byte limit`.
- Scan-policy-style wiring aside, the gzip test now asserts the coding is forwarded **and** that the body inflates to the original bytes, rather than only asserting Content-Length.

Also added: conda `channeldata.json` fail-closed and large-body serve; sharded-shard and JLAP guards for Remote, Local and Virtual; a **HEAD** request carrying the full `etag`/`content-length`/`x-repo-commit` trio (no test issued a HEAD before, though the bug being fixed is huggingface_hub's HEAD probe); a second cache-hit request replaying the same ETag and commit with the mock `.expect(1)` proving it came from cache; forwarded-header preference with the model-info mock `.expect(0)`; model-info failure not failing the download; tree pull-through and its upstream failure surfacing; hosted ETag + commit with `model_info.sha == resolve x-repo-commit`; slash-bearing revision re-encoding; and `validate_cache_path` unit tests for the literal and encoded dot-segment sets.

Three pre-existing namespaced-404 tests passed unchanged at the merge base (their URIs matched no route, so axum's fallback already produced the 404). They now assert the handler's own error body and `.expect(1)` on the upstream mock, so they fail at base.

## Test Checklist
- [x] Manually tested locally
- [x] No regressions in existing tests — full `cargo test --lib`: **14,122 passed, 2 failed**, both pre-existing and environment-specific (`test_hex_same_version_case_variants_advertise_one_downloadable_release_db` fails identically at the merge base and passes in CI; `test_large_object_client_builder_does_not_apply_total_timeout` fails with a loopback TCP connect timeout, verified identical with this PR's `http_client.rs` change reverted)

## API Changes
- [x] N/A — no new OpenAPI-documented endpoints. `huggingface.rs` and `conda.rs` carry no `#[utoipa::path]` annotations at all and are not registered in `ApiDoc`, consistent with every other format handler. (The previous revision of this description checked the utoipa/ToSchema boxes and named "HF revision routes"; that was wrong — those annotations do not exist.)
- [x] `cargo test --lib test_openapi_spec_is_valid` passes (nothing was added to the spec)
- [x] N/A - no migration
- [x] `cargo sqlx prepare --check --workspace` clean — no new checked macro queries

## Known follow-ups (not blocking)

- **LFS commit header.** The Hub serves LFS objects as a 302 to its CDN with `X-Repo-Commit` on the *redirect*, and reqwest follows redirects internally, exposing only the final hop's headers. So for model weights the forwarded header is unavailable and a bounded 2s model-info fallback is used instead — strictly weaker, since model-info is a second observation of a mutable ref. Capturing intermediate-hop headers needs `Policy::none()` plus a manual redirect loop that re-runs the SSRF check per hop; once that lands the fallback can be deleted outright.
- **The buffered-download audit found two more instances of the conda bug class outside this PR**, filed separately: `cargo.rs`'s Remote crate download buffers at 8 MiB (so any crate over that 502s — crates.io's default publish ceiling is 10 MiB), and nuget's flatcontainer cache-repair path both buffers at 8 MiB and builds its upstream URL without `PackageBaseAddress` discovery. Maven, goproxy and the other ~25 formats were checked and stream correctly.
- **Budget-reservation gaps** beyond conda: `debian.rs` has two unbudgeted 128 MiB buffered dists paths (missed when its siblings were budgeted), `protobuf.rs` buffers 128 MiB unbudgeted and its Virtual arm collects with no cap at all, and `pub_registry.rs` is the last caller of the uncapped `proxy_fetch`.
- conda's Remote `repodata.json` (identity, not `.zst`) can exceed the 128 MiB tier for a large channel; modern clients request `.zst` first and are unaffected.
- Only the first page of `/api/models/{id}/tree/{rev}` is proxied (`Link: rel="next"` is not followed). `hf download` does not use this endpoint.
- `GET /pypi//simple/` (empty repo-key segment) returns 500 rather than 404 — matchit matches the empty param and the middleware's early return skips inserting the auth extension. Not exploitable; no repository can hold the empty key.
- No end-to-end router test for a Virtual HF repo's local-member resolve headers; the new member-priority checksum SQL is covered directly instead.

### DIFF
--- a/backend/src/api/handlers/conda.rs
+++ b/backend/src/api/handlers/conda.rs
@@ -1567,23 +1567,32 @@ async fn serve_repodata(
         return Ok(cacheable_response(body, ct, headers));
     }
 
-    // For remote repos, proxy repodata from upstream
+    // For remote repos, proxy repodata from upstream. Real conda-forge
+    // repodata.json.zst commonly runs 20-30 MiB (e.g. noarch/osx-arm64), well
+    // past the 8 MiB DEFAULT_METADATA_MAX_BYTES ceiling used by most other
+    // formats' metadata proxying, so this uses the LARGE tier (128 MiB) the
+    // same way debian/npm/maven/pypi do for their oversized index documents.
+    //
+    // A fetch failure here must propagate to the client instead of silently
+    // falling through to `build_repodata` below: that fallback only reflects
+    // artifacts already cached in our own DB, so masking a real upstream
+    // failure (cap exceeded, upstream down, etc.) behind it would serve a
+    // valid-looking but empty/incomplete index and make the client believe
+    // there are no packages, rather than surfacing the actual problem.
     if repo.repo_type == RepositoryType::Remote {
         if let Some(ref upstream_url) = repo.upstream_url {
             if let Some(ref proxy) = state.proxy_service {
                 let upstream_path = format!("{}/{}", subdir, encoding.upstream_filename());
-                if let Ok((content, _ct)) = proxy_helpers::proxy_fetch_capped(
+                let (content, _ct) = proxy_helpers::proxy_fetch_capped(
                     proxy,
                     repo.id,
                     repo_key,
                     upstream_url,
                     &upstream_path,
-                    proxy_helpers::DEFAULT_METADATA_MAX_BYTES,
+                    proxy_helpers::LARGE_METADATA_MAX_BYTES,
                 )
-                .await
-                {
-                    return Ok(cacheable_response(content.to_vec(), ct, headers));
-                }
+                .await?;
+                return Ok(cacheable_response(content.to_vec(), ct, headers));
             }
         }
     }
@@ -1894,11 +1903,36 @@ fn group_artifacts_by_name<'a>(
     by_name
 }
 
+/// CEP-16 sharded repodata is only built from artifacts already present in
+/// our own DB (see [`list_conda_artifacts`]) - it never proxies an upstream's
+/// real shard index. For a Local/hosted repo that's exactly right: our DB is
+/// the source of truth. For a Remote repo it's not: an upstream like
+/// conda-forge that has not yet been mirrored into our DB would otherwise get
+/// a syntactically valid but semantically empty (or incomplete) shard index
+/// served with 200, which a CEP-16-aware client treats as authoritative and
+/// will not fall back from - silently hiding every package that hasn't
+/// already been cached. Until sharded repodata is actually proxied from the
+/// upstream, Remote repos must 404 here so clients (pixi, conda) take the
+/// documented fallback to `repodata.json`, which IS proxied.
+#[allow(clippy::result_large_err)]
+fn reject_unsupported_remote_sharding(repo: &RepoInfo) -> Result<(), Response> {
+    if repo.repo_type == RepositoryType::Remote {
+        return Err((
+            StatusCode::NOT_FOUND,
+            "Sharded repodata (CEP-16) is not available for remote/proxy conda repositories; \
+             use repodata.json",
+        )
+            .into_response());
+    }
+    Ok(())
+}
+
 async fn sharded_repodata_index(
     State(state): State<SharedState>,
     Path((repo_key, subdir)): Path<(String, String)>,
 ) -> Result<Response, Response> {
     let repo = resolve_conda_repo(&state.db, &repo_key).await?;
+    reject_unsupported_remote_sharding(&repo)?;
     let all_artifacts = list_conda_artifacts(&state.db, repo.id).await?;
     let subdir_artifacts = artifacts_for_subdir(&all_artifacts, &subdir);
     let by_name = group_artifacts_by_name(&subdir_artifacts);
@@ -1950,6 +1984,7 @@ async fn sharded_repodata_shard(
     }
 
     let repo = resolve_conda_repo(&state.db, &repo_key).await?;
+    reject_unsupported_remote_sharding(&repo)?;
     let all_artifacts = list_conda_artifacts(&state.db, repo.id).await?;
     let subdir_artifacts = artifacts_for_subdir(&all_artifacts, &subdir);
     let by_name = group_artifacts_by_name(&subdir_artifacts);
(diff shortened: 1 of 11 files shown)


brandonrc:
Internal review notes (non-blocking follow-ups, filing as fast-follow rather than blocking this PR since hf download is verified live end to end):

Important (fast-follow):
1. `fetch_upstream_commit_sha` adds a model-info cache lookup on every cache-miss resolve of a Remote-repo file, keyed independently from the file's cache entry. Confirm the model-info cache classification/TTL matches the file's, so a divergence can't trigger stale-revalidation upstream round-trips on every resolve.
2. Confirm `fetch_upstream_commit_sha`'s HTTP call has a bounded timeout independent of the file fetch, so a slow model-info endpoint cannot stall file downloads. Likely already bounded by the shared client config, but worth verifying.

Minor:
- The route-list doc comment at the top of huggingface.rs was not updated with the new `/revision/{revision}` variants (cosmetic).
- `fetch_upstream_commit_sha` and `proxy_model_info` both build `api/models/{id}/revision/{rev}` independently; could share a helper.

Review confirmed: the reqwest gzip feature is pulled in transitively by opensearch (verified via cargo tree); the no_gzip change applies only to the untrusted-upstream base client (internal service client untouched); SSRF redirect policy and CA cert application are unaffected; no code double-handles Content-Encoding; the metadata cap still bounds buffered reads regardless of compression.

Issue #2192, opened 2026-07-04
Phase 4c (#1608): stream the buffered blob FALLBACK paths instead of capping them


brandonrc:
Follow-up to #2181. Retiring the unbounded buffered read gave every buffered caller a per-caller byte cap (8/16 MiB). Most are metadata (correct). But several are package **blobs** on buffered recovery/fallback paths whose PRIMARY download streams (Phase 4) and is unaffected — the buffered fallback now caps: fetch_from_pypi_remote (wheel recovery, 16 MiB), npm.rs:1388 (tarball, 16 MiB), swift.rs:516 (.zip, 8 MiB), helm.rs:276 (chart, 8 MiB), oci_v2.rs:2607/2720/3077 (blob/manifest, 8 MiB). A genuinely large artifact (e.g. a >16 MiB npm/pypi package) that hits one of these buffered fallbacks would now **502** instead of OOM — better than a crash, but ideally these fallbacks should **stream** (via fetch_artifact_streaming) so large blobs succeed. Also: the secondary pypi simple-index path (pypi.rs:1530) inherited the 8 MiB default vs 16 MiB on the primary — align to 16 MiB. Low priority (fallback paths are rarely hit; primary streams); tracks the honest caveat from #2181.

PR #2919, opened 2026-07-26
Repair a missing cached .nupkg via the discovered PackageBaseAddress, streaming

### LINKED ISSUE #2917: nuget: flatcontainer cache-repair path requests the wrong upstream URL and caps .nupkg at 8 MiB
`flatcontainer_download` has a cache-**repair** path that fires when a Remote repository holds an `artifacts` row for a package but the blob is missing from storage (a pre-#1278 `proxy-cache/…` storage key read through the per-repo backend, a hydrated/replicated copy, or a package pushed into the remote). `FilesystemStorage` maps the missing key to `NotFound` (#1016), which invokes `get_cached_or_refetch`'s refetch closure.

Two defects at that one site:

**1. Wrong upstream URL — the more fundamental one.** The closure builds `v3/flatcontainer/{id}/{version}/{file}` and appends it to `upstream_url`, bypassing `discover_upstream_resources` / the service index's `PackageBaseAddress`. `ProxyService::build_upstream_url` is a bare concatenation, and a NuGet remote's `upstream_url` is normally the service-index *document* — so against nuget.org the repair requests `https://api.nuget.org/v3/index.json/v3/flatcontainer/{id}/{version}/{file}`, which 404s. This path is therefore broken regardless of package size.

**2. 8 MiB cap on a `.nupkg` binary.** The refetch uses `proxy_fetch_capped` at `DEFAULT_METADATA_MAX_BYTES`, which 502s rather than truncating. Realistic exceedances include the `Microsoft.CodeAnalysis.*` and `Microsoft.ML.*` packages and most native-runtime packages.

The URL bug masks the cap bug (404 before any body is read), so the two need isolating to demonstrate independently.

**Impact.** A first `dotnet restore` against a clean Remote repository works — the primary arm discovers and streams correctly. The failure shows up on a *later* restore after a storage wipe, GC, or migration, when the row survives but the blob does not.

The primary Remote arm and the V2 arm both stream correctly and are unaffected.

brandonrc:
## Summary

Closes #2917.

`flatcontainer_download`'s cache-**repair** path had two defects. It now routes through `proxy_v3_flatcontainer(streaming = true)` — the same call the primary Remote arm makes — which fixes both in one go, because that helper performs the service-index discovery *and* streams.

1. **Wrong upstream URL.** The old refetch closure appended `v3/flatcontainer/{id}/{version}/{file}` to `upstream_url`, bypassing `discover_upstream_resources`. Since `build_upstream_url` is a bare concatenation and a NuGet remote's `upstream_url` is normally the service-index *document*, the repair requested `…/v3/index.json/v3/flatcontainer/…` and 404'd against nuget.org — broken regardless of size.
2. **8 MiB cap on a `.nupkg`.** `proxy_fetch_capped` 502s rather than truncating.

A warm Remote hit also no longer buffers the whole `.nupkg` through `storage.get`; the whole branch is now a single `match storage.get_stream(...)`.

## How the repair path is reached

`flatcontainer_download` looks the package up by `(repository_id, LOWER(name), version)`. When a row **is** found, the primary Remote cache-miss arm is never entered. A Remote repository can hold such a row with the object gone: a pre-#1278 `proxy-cache/…` storage key read through the per-repo backend, a hydrated/replicated/imported copy, or a package pushed into the remote. The new test proves reachability empirically — with a row inserted and no blob written, the wiremock upstream receives the request.

So a first `dotnet restore` against a clean remote always worked; this fails on a *later* restore after a storage wipe, GC or migration.

## Reviewer notes

- **The write-back target changed, and it is worth a decision.** The old code re-`put` the bytes at `artifact.storage_key`; routing through the proxy helper warms the *proxy* cache instead, so the row's blob is not re-materialized — a later request re-enters the repair branch and is served warm from the proxy cache with no upstream traffic. For the dominant real case (a pre-#1278 `proxy-cache/…` key) the old write-back could never heal anything anyway, so this is a net improvement. If you want the row's blob genuinely re-materialized, the PyPI shape (`get_remote_cached_or_refetch_stream` + `tee_refetch_to_storage`) is the pattern.
- **`get_cached_or_refetch` now has no production caller** (it was nuget-only), so it is documented as superseded and marked `#[allow(dead_code)]` with its tests retained. Deleting it instead is a legitimate alternative — flagging rather than deciding.
- Response headers on the repair path are the proxy's streaming shape, identical to the primary Remote arm. `record_download` is preserved, recorded after the upstream body opens so a failed repair is not counted.
- Single-flight moves from the buffered hydration lease to the streaming coordinator (#1631 layer 2), the same trade PyPI accepted in #2192.
- Quarantine/Package-Age gating is equal or stricter: `check_artifact_download` still runs before the branch, and `open_streaming_leader` refuses to open a cold streaming fetch when the age policy is enabled.
- Cache key is preserved character-for-character, so existing entries stay hits.

Both defects are demonstrated **independently**, since the URL bug masks the cap bug:

```
# pre-fix — URL shape
repair must fetch the discovered PackageBaseAddress path
(/flat/newtonsoft.json/13.0.1/newtonsoft.json.13.0.1.nupkg); upstream saw
["/v3/index.json/v3/flatcontainer/newtonsoft.json/13.0.1/newtonsoft.json.13.0.1.nupkg"]

# discovery fixed, still buffered — isolates the cap
repair of a 9437184 byte .nupkg must not be capped, got 502 Bad Gateway
```

New: `test_flatcontainer_download_repairs_missing_blob_via_discovered_package_base` and `test_flatcontainer_download_repair_streams_body_above_metadata_cap` (9 MiB non-uniform body, byte equality). Renamed `…_routes_through_cached_or_refetch_helper` → `test_flatcontainer_download_remote_warm_cache_hit_served_from_storage`: it seeds the blob, so the refetch closure was never invoked and the old name was inaccurate.

## Test Checklist
- [x] Manually tested locally
- [x] No regressions in existing tests — `nuget` 139 passed, `proxy_helpers` 227 passed, clippy clean workspace-wide

## API Changes
- [x] N/A — no new endpoints, no schema change, no migration
- [x] `cargo sqlx prepare --check --workspace` clean — the repair path uses no SQL

## Note
Branched before #2915 merged, so it may need a trivial rebase onto `main`.


brandonrc:
Pipeline complete: reproduced #2917 on the base, verified the fix, reviewed, rebased and pushed `50d16bd`, all 16 checks green.

Stage 1 reproduced **both** reported defects independently against the base — the concatenated path never reaching the discovered `PackageBaseAddress`, and the 8 MiB cap turning a large `.nupkg` repair into a 502. Stage 2 confirmed both closed, with a negative control: a package genuinely absent upstream still fails cleanly with a 404 in ~19ms, no hang and no 500.

**The important finding was not in the diff — it was in the branch's age.** A reviewer found a working SSRF-plus-credential-exfiltration PoC against this branch: `guard_upstream_base` checked only the URL scheme, so a hostile upstream service index could point `PackageBaseAddress` at any host, the connect-time DNS guard didn't fire on an IP literal, and the repository's configured upstream credentials were forwarded to it verbatim.

That PoC is real, but the conclusion needed correcting, and it changes the remedy: **main is not vulnerable.** Main already carries a three-argument origin-pinned `guard_upstream_base` that rejects off-origin with *"refusing to send upstream credentials off-host"*, plus its own regression tests. The branch was ~230 PRs stale and predated that hardening — so merging it as written would have **reverted** main's origin pin while looking like a fix, which is worse than introducing a new hole because it reads as forward progress.

So the fix was a rebase, not a hand-written guard. Merging main in restored the pinned three-arg version, and I confirmed the PR's repair tests and main's off-origin rejection tests now coexist: 154 nuget tests pass. I'd rather inherit hardening that already exists than re-derive it.

Also dropped `.pr2919.diff` — a 20 KB working file committed at the repo root — and moved the CHANGELOG entry from `[1.6.0]` to `[Unreleased]`, since the branch was cut before 1.7.0 and 1.7.1 shipped.

**Two follow-ups I did not fold in.** The repaired body is cached with `expected_checksum: None` even though the artifact row carries a checksum and a `_verified` sibling helper exists one argument away — worth closing given the row is exactly what's being healed. And the streamed repair is disk-unbounded: the synthesized repo carries `quota_bytes: None`, which makes `exceeds_single_object_quota` return false unconditionally, so the per-object quota guard never fires on this path. Memory is properly bounded (the 4 MiB tee ceiling is good design); disk isn't. Both pre-existing and shared with other streaming paths.

I can't approve this one: `gh` runs as `brandonrc` and GitHub rejects self-approval. Validated, reviewed and labeled `ready-to-merge`; the approval is yours.

PR #2922, opened 2026-07-26
Forward upstream Content-Encoding on cargo download responses

### LINKED ISSUE #2920: Cargo download arms drop upstream Content-Encoding
## Problem

The cargo `.crate` download handler builds its response by hand on both the Remote arm and the Virtual arm, and neither forwards `StreamingFetchResult.content_encoding`. The field is read one line after `content_length` is taken from the same struct and then discarded.

The struct's own doc states the requirement:

> Upstream `Content-Encoding` when the cached/streamed bytes are content coded. Must be forwarded with the body on every serve (fresh, cache hit, stale-if-error) or the client cannot decode what it receives.

The shared helper `build_streaming_response_with_disposition` forwards it, but these two arms do not go through that helper (deliberately, so the content type stays pinned to `application/x-tar` rather than taking upstream's `application/gzip`).

## Impact

The proxy no longer decodes upstream bodies, so an undeclared content coding reaches cargo as compressed bytes it will not inflate. The client writes them to disk and fails the index `cksum` check. This needs an intermediary that applies a transfer coding: a corporate proxy, an nginx with gzip on, or a mirror. crates.io itself does not trigger it.

## History

This is the same class fixed elsewhere in #2915, whose commit message notes the gzip test now asserts the coding is forwarded and the body inflates to the original bytes. #2918 converted the cargo Remote arm to streaming and preserved the existing behaviour here, so the gap predates it on both arms and was not introduced by it.

## Fix

Forward `result.content_encoding` in both hand-rolled builders. Two lines plus the `CONTENT_ENCODING` import.

## Notes

Not a regression: the pre-#2918 buffered arm dropped the header too. Filing so the fix is tracked rather than folded silently into unrelated work.

brandonrc:
## Summary

Closes #2920.

The Remote and Virtual cargo download arms build their responses by hand rather than routing through `build_streaming_response_with_disposition`. That is deliberate: the shared helper prefers upstream's content type, and crates.io serves `application/gzip`, so going through it would un-pin `application/x-tar` and break parity with the local-hit arm. But both hand-rolled builders read `content_length` off `StreamingFetchResult` and then discarded `content_encoding`, whose own doc comment states it "must be forwarded with the body on every serve (fresh, cache hit, stale-if-error) or the client cannot decode what it receives."

Since #2915 the proxy no longer decodes upstream bodies, so an undeclared coding reaches cargo as compressed bytes it will not inflate. The client writes them to disk and fails the index `cksum`. Triggering it requires an intermediary that applies a transfer coding: a corporate proxy, an nginx with gzip on, or a mirror. crates.io itself does not.

**Not a regression from #2918.** The pre-existing buffered arm dropped the header too, and the Virtual arm has always dropped it. #2918 preserved existing behaviour here; this closes the gap on both arms.

`test_remote_crate_download_forwards_upstream_content_encoding` serves a gzip-coded upstream body and asserts the coding is forwarded and the bytes arrive still compressed and byte-identical, so a regression that re-enables transparent decoding fails here too.

Verified failing with the fix reverted, on the header assertion specifically:

```
upstream Content-Encoding must be forwarded, or cargo writes undecodable
bytes to disk and fails the checksum
  left: None
 right: Some("gzip")
```

It also pins `Content-Type: application/x-tar`, so a future move to the shared helper cannot silently start forwarding upstream's `application/gzip`.

## Test Checklist
- [x] Manually tested locally
- [x] No regressions in existing tests: `cargo test --lib handlers::cargo::` 132 passed, clippy clean, rustfmt clean

## API Changes
- [x] N/A - no API changes

## Note

The CHANGELOG entry amends the merged #2918 entry in place rather than adding a new bullet, since it describes the same release-visible behaviour. It also documents two operator-visible consequences of that merged change which were not written down at the time: `Content-Length` is omitted when upstream does not advertise one, and a Remote cargo repository with the Package Age Policy enabled now returns 409 for any crate not already cached (streaming has no buffered `Last-Modified` to compute the hold from, and because nothing is cached the hold does not elapse on its own). `QUARANTINE_ENABLED` acts as a global default, so that can affect every Remote cargo repository at once.


brandonrc:
Pipeline complete: reproduced #2920 on the base — on **both** arms — verified the fix, reviewed, pushed `6cf4c65`, all 16 checks green.

Stage 1 reproduced the Remote arm with the PR's own test against reverted production code (`left: None, right: Some("gzip")`), and the Virtual arm with a throwaway probe. Stage 2 confirmed both arms forward, plus the cache-hit leg replays the coding from the sidecar with `upstream_hits` unchanged, plus a negative control where no upstream coding produces no manufactured header and a byte-identical body.

The fix itself is minimal and correct: `Content-Length` and `Content-Encoding` come from the same source on every serve path, and there's no double-encoding risk — the only `CompressionLayer` is npm's, and the outbound client pins `Accept-Encoding: identity` and disables transparent decoding, so the bytes genuinely arrive coded.

**What I pushed:** the Virtual-arm regression test the PR was missing. The fix touched both arms but shipped a test for one, so the Virtual arm could regress silently — exactly the gap that let this bug exist on two arms in the first place. It discriminates cleanly: reverting only the Virtual arm fails only the Virtual test, reverting only the Remote arm fails only the Remote one. Also merged the branch up to main (90 commits behind; resolved `test_db_helpers` by keeping the branch's `send_with_headers`, which main lacks, plus main's newer `grant_repo_access` doc) and moved the CHANGELOG entry from `[1.6.0]` to `[Unreleased]`, since this ships in 1.7.2+.

**One thing I did not touch and would flag.** The CHANGELOG diff also back-fills a paragraph documenting **#2918's** behavior — including the operator-facing claim that a Remote cargo repo with Package Age Policy enabled *"now returns 409 for any crate not already cached"* and that `QUARANTINE_ENABLED` can trigger it globally. I left it in the `1.6.0` section, which is the correct home for #2918's text. But **I have not verified that claim** — it describes a different PR's behavior riding on a two-line fix, and it deserves independent confirmation before it ships as release copy.

Filed **#3149** for the same defect class in npm, OCI, generic repos and the cargo sparse index, rather than widening this PR. The sparse-index arm is a different mechanism — `proxy_fetch_capped` returns a 2-tuple that structurally discards the encoding.

I can't approve this one: `gh` runs as `brandonrc` and GitHub rejects self-approval. Validated, reviewed and labeled `ready-to-merge`; the approval is yours.

PR #2943, opened 2026-07-28
perf: index the metadata-files attribution lookup and consult the claims ledger first (#2942)

### LINKED ISSUE #2942: Maven flat-key attribution: metadata-files layer full-scans artifact_metadata on every row-less read, and runs before the indexed owner-table lookup
## Summary

On cloud storage backends, `attributed_owner` (`backend/src/services/maven_flat_attribution.rs`) resolves ownership through three layers in this order:

1. `OWNER_BY_ARTIFACT_ROW_SQL` — live `artifacts` row (indexed, fast)
2. `OWNER_BY_METADATA_FILES_SQL` — scan of `artifact_metadata.metadata->'files'` (**unindexable full scan**)
3. `OWNER_BY_ATTRIBUTION_TABLE_SQL` — `maven_flat_object_owner` PK lookup (indexed, fast)

Layer 2 cannot use any index in its current form, so **every flat-key read that is not satisfied by a live artifact row pays a sequential scan of the entire `artifact_metadata` table** — before the cheap, indexed attribution-table lookup in layer 3 is ever tried.

On a deployment with ~2M `artifact_metadata` rows, each layer-2 execution takes 3.7–6.6 s (sqlx slow-statement log):

```
WARN sqlx::query: slow statement: execution time exceeded alert threshold
  summary="SELECT DISTINCT a.repository_id FROM …"
  db.statement="SELECT DISTINCT a.repository_id FROM artifact_metadata am
    JOIN artifacts a ON a.id = am.artifact_id
    JOIN repositories r ON r.id = a.repository_id
    WHERE a.is_deleted = false AND r.storage_backend = $2
      AND jsonb_typeof(am.metadata->'files') = 'array'
      AND EXISTS ( SELECT 1 FROM jsonb_array_elements(am.metadata->'files') f
                   WHERE COALESCE(f->>'storageKey', f->>'storage_key') = $1) LIMIT 2"
  rows_returned=0 elapsed=6.210230502s
```

The `EXISTS (SELECT 1 FROM jsonb_array_elements(...) WHERE COALESCE(...) = $1)` predicate is evaluated per row; no GIN/expression index can serve it, so the planner has no choice but a seq scan with per-row jsonb detoasting.

## Impact

- Every read of a row-less legacy object (GAV-grouped companions, checksum sidecars, verbatim `maven-metadata.xml`) stalls for seconds — **even when the key is already attributed in `maven_flat_object_owner`** (migration 163/170 backfill or a write-time claim), because the indexed table is consulted *after* the scan.
- Worse, the scan also runs for keys that don't exist at all: a virtual-repository walk probes each local member before falling through to remote members, so **every cache-miss request for a public coordinate through a virtual pays the full scan on each cloud-backed local member** — with `rows_returned=0`, always.
- Under concurrent CI load these multi-second queries queue up and saturate the connection pool; unrelated endpoints (e.g. repository listing) degrade from <1 s to tens of seconds, and reads start timing out.

Verified unchanged on `main` as of this report.

## Suggested fixes

Either of these resolves the practical problem; both together are ideal:

1. **Reorder the layers: attribution table before the metadata scan.** `maven_flat_object_owner` is populated by migrations 163/168/170 and maintained by write-time claims, so on a healthy deployment layer 3 already answers nearly every key that layer 2 could — at PK-lookup cost. A conclusive owner (or a conclusive ambiguity) from the table would short-circuit the scan; the metadata scan remains as the last-resort fallback for keys written after backfill by a version that predates write-time claiming.

2. **Make the metadata query indexable.** Replace the `EXISTS`/`COALESCE` predicate with jsonb containment that a GIN index can serve:

   ```sql
   AND (am.metadata->'files' @> jsonb_build_array(jsonb_build_object('storageKey', $1::text))
     OR am.metadata->'files' @> jsonb_build_array(jsonb_build_object('storage_key', $1::text)))
   ```

   plus a migration:

   ```sql
   CREATE INDEX IF NOT EXISTS idx_artifact_metadata_files_gin
     ON artifact_metadata USING gin ((metadata->'files') jsonb_path_ops);
   ```

   The two containment branches preserve the dual-spelling behavior from #2706 / migration 170 (`storageKey` written by the upload handler, `storage_key` accepted as fallback), and `jsonb_path_ops` serves both via a BitmapOr. Semantics match the current predicate for both spellings; rows where an element carries *both* spellings with different values would differ from `COALESCE` (which only checks `storageKey`), but no writer has ever produced such rows.

3. (Optional, alongside 1) **Negative-result short-circuit for virtual walks**: when a local member has neither a live row nor an attribution-table row, skip the metadata scan during virtual member iteration and treat the member as a miss — the scan's only purpose there is to rescue pre-backfill legacy keys, which the table already covers after 163/170.

## Environment

- artifact-keeper 1.6.2 (also inspected `v1.6.3` and `main` — layer order and query unchanged)
- PostgreSQL 14, S3 storage backend
- ~2M rows in `artifact_metadata`, ~4M in `artifacts`, ~27M in `maven_flat_object_owner`

## Related

- #2504 (flat-key fail-closed attribution), #2574/#2584 (migration 163), #2671 (backend-qualified PK, migration 168), #2706 (`storageKey` casing, migration 170)

ThaSami:
Closes #2942.

## Problem

`resolve_direct` runs the attribution layers as live-row → metadata-files → owner-table. The metadata-files predicate — `EXISTS (SELECT 1 FROM jsonb_array_elements(metadata->'files') f WHERE COALESCE(f->>'storageKey', f->>'storage_key') = $1)` — is evaluated per row and cannot be served by any index, so every row-less flat-key read not satisfied by a live artifact row sequentially scans the entire `artifact_metadata` table. On a deployment with ~2M metadata rows that is 3.7–6.6 s per request; the scans queue up under concurrent load and saturate the connection pool. The scan also fires for keys that don't exist at all — every cloud-backed local member probed during a virtual-repository walk pays it on each cache miss, with `rows_returned=0` (details and measurements in #2942).

## Changes

1. **`OWNER_BY_METADATA_FILES_SQL` rewritten to jsonb containment**, one branch per `files[]` key spelling (camelCase `storageKey` as written by the upload handler since #418, snake_case `storage_key` as the #2706 fallback), served by a new GIN `jsonb_path_ops` expression index on `(metadata->'files')` — **migration 179**. The planner turns the two branches into a BitmapOr of two index probes. Containment against an array operand is false for non-array `files` values, so the previous `jsonb_typeof` guard is implicit.

2. **`resolve_direct` consults the `maven_flat_object_owner` claims ledger (a primary-key probe) before the metadata derivation.** Beyond cost, this closes a read/write disagreement: `guard_flat_key_writable` enforces against the *recorded* claim, but on read the derivation layer ran first, so
   - a claim recorded by `claim_flat_key_on_write` could be overridden by an attribution derived from another repository's metadata, and
   - a key that is metadata-ambiguous resolves `None` for the write guard (so a first writer may legitimately claim it), yet the old read order still hit the ambiguous derivation first and failed closed — leaving a legitimately claimed key unreadable by its claimant.

   A recorded claim now outranks derived attribution on read, matching the write guard. On deployments with a populated ledger (163/168/170 backfills + write-time claims) the derivation only runs for keys the ledger has never seen.

## Semantics notes

- The containment rewrite matches either spelling; `COALESCE` checked `storageKey` first and ignored `storage_key` when both were present with different values. No writer has ever produced an entry with both spellings, so behavior is identical on real data.
- Layer-2 ambiguity handling is unchanged (`LIMIT 2` → two distinct owners → fail closed) — it just runs after the ledger now.
- Migration 179 is `IF NOT EXISTS`-idempotent. The in-transaction build blocks writes to `artifact_metadata` for the build duration; the migration header documents the out-of-band `CREATE INDEX CONCURRENTLY` alternative for operators who can't take the write pause (same pattern as migration 166).

## Tests

- New `test_attribution_table_claim_outranks_metadata_derivation` — fails under the old layer order, passes with the ledger consulted first, and asserts the claimant/deriver read polarity via `flat_key_readable`.
- Existing `test_attributed_owner_from_metadata_files_camelcase` / `_snake_case` (from #2707) exercise the containment rewrite for both spellings.


ThaSami:
@brandonrc if possible can we have it as a patch in 1.6.x please

brandonrc:
> @brandonrc if possible can we have it as a patch in 1.6.x please 

Should not be a problem. Let me take a look at this.

brandonrc:
Looks solid, going to use my red-team rig on it. If all passes it will go into main. Then I will look at backporting it to 1.6.x. I will keep you posted on that.

brandonrc:
Merged — thank you, @ThaSami, this is a genuinely valuable fix. 🙏

We ran it through an isolated red-team verification of the layer reorder (both orderings executed against the real migrated schema on a throwaway DB, plus a live build with migration 179 applied). It confirmed the reorder is not just a perf win but a **tenant-isolation hardening**, and it actually surfaced that the *pre-fix* order was worse than a read inconsistency: because `flat_key_readable` and `guard_flat_key_writable` share `attributed_owner`, the metadata-first order let a non-owning repo that merely referenced a claimed key in its own `files[]` **overwrite/poison** that key (the write guard saw it as "own key"), while locking the true claimant out of writing its own object. Consulting the ledger first closes the read-leak, the write-poison, and the owner-lockout together — exactly as your `test_attribution_table_claim_outranks_metadata_derivation` asserts (verified RED on the old order / GREEN on yours).

**Release split:** this lands in **1.7.0** as-is. We're also backporting the **layer reorder only** to **1.6.x** as a security patch, but leaving migration 179 (the `jsonb_path_ops` GIN index) and the `@>` containment rewrite on 1.7.0 — a write-blocking in-transaction index build is heavier than we want on a patch release, and the reorder is the security-relevant part (it's independent of the SQL form, which the verification confirmed resolves identically with or without the index). 1.6.x users still get the isolation fix; the indexed fast-path ships with 1.7.0. The backport PR will credit you and reference this one.

Issue #2947, opened 2026-07-28
Epic: curation policy rule types — publisher-trust + popularity/typo-squat


brandonrc:
Roadmap from a customer CISO security evaluation (2026-07-28): extend the curation engine (`CurationRule`, currently pattern/version/arch allow-block) with two new **rule types** to reduce manual approval overhead:
1. **publisher-trust** (#PUB) — allow/flag/block by trusted publisher (e.g. Microsoft, NumFOCUS), keyed on provenance/attestation where available (NOT spoofable self-asserted author fields).
2. **popularity / typo-squat** (#POP) — download-count threshold (pluggable source: pypistats / npm downloads) + lexical-distance typo-squat detection vs a popular-package list.

Shared foundation (this epic): a `rule_type` discriminator + `config` JSONB on `curation_rules`, an evaluation-dispatch seam, and typed-rule CRUD API. MVP targets PyPI + npm; deeper sigstore/attestation provenance is a fast-follow. Target: **1.7.0**.
## Definition of done / validation (required before these ship)
Beyond unit tests, each rule type must be validated **end-to-end**:
1. **DTF E2E tiers** in artifact-keeper-test (`deploy-test/harness/tiers/`), discriminating (RED pre-fix / GREEN post-fix):
   - `curation-global-publisher-trust` — stand up the stack, configure a **global** `publisher_trust` rule, drive a real pull of a package from a trusted vs untrusted publisher, assert the allow/flag/block decision is enforced at the pull gate; assert a non-applicable format (raw/generic) is unaffected (`NotApplicable`).
   - `curation-global-popularity` — configure a **global** `popularity` rule, drive a pull of a below-threshold and a typo-squat-named package, assert flag/block; assert `Unknown`-source fails open (flag, not block) and non-applicable formats pass through.
   - **Determinism:** the external data sources (pypistats / npm downloads / registry metadata) must be exercised via a controlled fixture or seeded state so the tiers are not flaky on external-API availability (mock/stub upstream, not live internet in CI).
2. **Red-team pass** after integration: adversarially attempt to (a) forge a trusted publisher / bypass the provenance check, (b) evade the popularity/typo-squat gate, (c) escape the global scope. Findings block ship until closed.

PR #2951, opened 2026-07-28
feat(curation): popularity + typo-squat rule evaluator and download-count sources

### LINKED ISSUE #2949: Curation rule: popularity threshold + typo-squat detection (download counts)
Part of #2947. New `popularity` curation rule type: (a) block/flag packages below a download-count threshold, and (b) typo-squat detection — flag a package whose name is within a small lexical distance (Damerau-Levenshtein) of a popular package but is itself unpopular. Pluggable download-count source (pypistats.org API / PyPI BigQuery public dataset; npm downloads API), with caching + graceful degradation on source outage. Config shape: `{min_downloads:N, window:"month", typosquat_check:true, action:"flag|block"}`. Best framed as flag-for-review by default (legit new packages have low counts). From a customer security evaluation.

brandonrc:
## Summary

Adds the self-contained evaluator + data-source layer for the `popularity` curation rule type (download-count threshold + typo-squat detection), from the curation rule-types epic. Motivated by a customer security evaluation: reduce manual approval overhead by letting curation policy catch unpopular and lexically-suspicious packages automatically.

Closes #2949
Depends on #2947 (curation foundation: `rule_type` discriminator + dispatch seam). This PR deliberately does **not** touch the `CurationRule` model, migrations, or `curation_service` dispatch — it provides the evaluator and data source the foundation wires in, so the two PRs compose with only a `mod` merge.

New module `backend/src/services/curation/`:

- **`popularity_source.rs`** — `#[async_trait] PopularitySource { async fn downloads(&self, format, name) -> PopularityResult }` with `PopularityResult::{Known(u64), Unknown}`. Real impl `HttpPopularitySource`: PyPI via `GET https://pypistats.org/api/packages/{name}/recent` (last-month total, PEP 503 name normalization) and npm via `GET https://api.npmjs.org/downloads/point/last-month/{name}` (scoped names path-encoded). Uses the repo-standard SSRF-guarded `base_client_builder()` with a 5s total / 3s connect timeout. Every failure (timeout, non-2xx incl. 429, malformed body) degrades to `Unknown` — a source never fabricates a count. `CachedPopularitySource<S>` TTL decorator (15 min positive / 60 s negative cache so an outage is not amplified but recovery is quick). `FakePopularitySource` test double with call counting.
- **`typosquat.rs`** — `damerau_levenshtein` (optimal-string-alignment: transposition = 1 edit, so `reqeusts`→`requests` is distance 1), `nearest_popular`, `is_typosquat(name, popular, max_distance) -> Option<target>` (self-exclusion: a name that is itself on the popular list is never flagged — `request` vs `requests`), plus built-in top-package seed lists for PyPI and npm (overridable per rule via config).
- **`popularity.rs`** — `pub async fn evaluate(config, format, name, version, source) -> Decision` with `Decision::{Allow, Flag(String), Block(String), NotApplicable}` (variants mirror the foundation's `CurationDecision`; the integration layer maps 1:1).

### Decision semantics (doc-commented + tested)

- **`NotApplicable`** for any format without a public download-count source (currently applicable: `pypi`/`poetry` and `npm`/`yarn`/`pnpm`) — popularity rules run as instance-wide policy, so a global rule silently passes generic/docker/maven/etc. through instead of flagging everything.
- Known count below `min_downloads` → configured `action` (`flag` default, `block` opt-in; flag-first because legit new packages are legitimately low-count), reason includes the count and threshold.
- Typo-squat match (within `max_distance`, clamped 1–2, of a popular name while not itself popular) → **Flag** naming the likely target; advisory-only, never a block on its own.
- `Unknown` downloads → `Flag("popularity unknown …")`, never Block, regardless of `action`: fail-open on the data source, fail-safe on the decision, so a stats-API outage cannot hard-block installs.
- Block > Flag when signals combine; reasons are joined.

Config shape: `{"min_downloads": N, "window": "month", "typosquat_check": true, "max_distance": 2, "action": "flag"|"block", "popular_packages": [...]}` — all keys optional.

No new crate dependencies (reuses `reqwest`, `async-trait`, `serde_json`). No migration.

### Reviewer note: external data sources

pypistats.org and api.npmjs.org are public, unauthenticated APIs with informal rate limits. The TTL cache plus short-TTL negative caching bounds request volume, and every degradation path is `Unknown` → flag-for-review, but operators with heavy ingest may want a curated `popular_packages` list and a higher cache TTL; a BigQuery-backed source can be added behind the same trait later.

## Test Checklist
- [x] Unit tests added/updated (33 new tests: threshold flag/block/allow, typo-squat incl. transposition + self-exclusion, Unknown→Flag-not-Block, NotApplicable formats with zero source calls, cache hit/TTL-expiry/negative-cache, distance properties, API-body parsers, PEP 503 normalization, path-segment encoding)
- [x] Manually tested locally (`cargo nextest run --lib`: 14188 passed; fmt + clippy `-D warnings` clean; docker backend image builds)

## API Changes
- [x] N/A - no API changes (rule CRUD API lands with the foundation PR)

### DIFF
--- a/backend/src/services/curation/mod.rs
+++ b/backend/src/services/curation/mod.rs
@@ -1,0 +1,17 @@
+//! Curation rule-type evaluators.
+//!
+//! Each rule type lives in its own submodule and exposes a self-contained
+//! `evaluate` entry point plus whatever data sources it needs. The curation
+//! foundation (`rule_type` discriminator + `config` JSONB dispatch on
+//! `curation_rules`, epic #2947) calls into these evaluators; the evaluators
+//! themselves stay free of database and model dependencies so they can be
+//! unit-tested in isolation.
+//!
+//! Currently implemented:
+//! - [`popularity`] — download-count threshold + typo-squat detection
+//!   (issue #2949), backed by [`popularity_source`] (pluggable download-count
+//!   providers) and [`typosquat`] (lexical-distance matching).
+
+pub mod popularity;
+pub mod popularity_source;
+pub mod typosquat;
--- a/backend/src/services/curation/popularity.rs
+++ b/backend/src/services/curation/popularity.rs
@@ -1,0 +1,344 @@
+//! `popularity` curation rule evaluator (#2949): download-count threshold +
+//! typo-squat detection.
+//!
+//! # Decision semantics
+//!
+//! - **Not applicable** — popularity rules run as instance-wide policy, but
+//!   only some ecosystems have a public download-count source (currently
+//!   PyPI via pypistats.org and npm via the npm downloads API, plus their
+//!   registry-compatible aliases). For any other format `evaluate` returns
+//!   [`Decision::NotApplicable`] so a global rule silently passes raw/
+//!   generic/private-only formats through instead of flagging everything.
+//! - **Below threshold** — a known download count under `min_downloads`
+//!   applies the configured `action` (`"flag"` default, `"block"` opt-in).
+//!   Flag-first is deliberate: legitimately new packages have low counts.
+//! - **Typo-squat** — a name within `max_distance` (1–2) edits of a popular
+//!   package, while not itself popular, is *flagged* for review. The signal
+//!   is advisory (never a block from this check alone): lexical proximity has
+//!   false positives by construction.
+//! - **Unknown popularity** — a source outage/rate limit/unlisted package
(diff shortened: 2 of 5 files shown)


brandonrc:
Folded into #2952 (the integrated typed-curation feature branch), which wires this module into the dispatch. Closing as superseded; the work + attribution live on in #2952 (Closes #2947/#2948/#2949).

Issue #1480, opened 2026-05-29
Dependency-track integration shown as disabled when the healthcheck is displayed as passing


flopma:
### Description

Dependency-track health check is listed as passing (green) in the monitoring section; but in the Security Dashboard, it says that Dependency-track is unavailable.

### Environment Information

- Deployment type: containers, following the docket compose file - but with podman
- Operating System: RHEL9
- Architecture: x86-64
- Database: Postgresql
- Reverse Proxy: Caddy
- Network setup: all containers running on the same server
- Dependency-track version: 4.14.2

### Backend Version

1.2.0-rc.2

### Web UI Version

1.1.3

### Package Format

None

### Expected Behavior

Dependency-track should be available for Artifact-Keeper

### Observed Behavior

Dependency-track is up and running, it is accessible through its web interface

Artifact-Keeper monitoring , regular health checks, displays Dependency-track as being reachable

Artifact-Keeper can't make use of Dependency-track

### Steps to Reproduce

1. Deploy dependency-track with its frontend as a standalone product, as described in https://docs.dependencytrack.org/getting-started/deploy-docker/
2. Login into dependency-track and change it's admin password
3. Generate a API key for the Automation Team
4. Deploy Artifact-keeper components through containers, as described in the docker compose file (opensearch, trivy, openscap, backend, web, caddy) - make sure to configure the environment variables concerning dependency-track needed by the artifact-keeper backend (URL, API KEY, ENABLE flag)
5. In Artifact keeper web interface, check in the monitoring section if all is green
6. In Artifact keeper web interface, go into the security dashboard and see that dependency-track is described as unavailable

### Logs

```shell
2026-05-29T15:24:51.914330Z  WARN artifact_keeper_backend::services::dependency_track_service: Dependency-Track base_url is not HTTPS and not a private network address. Set ALLOW_HTTP_INTEGRATIONS=1 to allow plain HTTP connections. url=http://cefpost2029-rh9.redacted.domain:8291
2026-05-29T15:24:51.924948Z  INFO artifact_keeper_backend::services::dependency_track_service: Dependency-Track integration initialized url=http://cefpost2029-rh9.redacted.domain:8291
2026-05-29T15:24:51.924972Z  INFO artifact_keeper: Dependency-Track integration enabled
2026-05-29T15:25:36.733186Z DEBUG http_request{method=GET uri=/api/v1/dependency-track/status}: tower_http::trace::on_request: started processing request
[...]

Issue #1395, opened 2026-05-25
Dependency track not fully disabled


bsosnowski:
### Description

There are mixed results with Dependency Track integration when `DEPENDENCY_TRACK_ENABLED=false` is set on the backend.
When we go to:
1. Registry -> Artifact > Choose Artifact > Security tab
The following information appears: 'Dependency-Track: Disabled' in red at the top and 'Dependency Track is unavailable' in yellow at the bottom. 
2. Security -> DT Projects
Information stating 'Dependency-Track is unavailable' appears in yellow.

Shouldn't there only be information that this feature is disabled? Or maybe these 'tabs' should be greyed/disabled? Maybe this belong in 'web'?

### Environment Information

- Deployment type: Docker
- Operating System: ArchLinux
- Architecture: x64
- Database: Postgres in Docker
- Reverse Proxy: Traefik
- Network setup: Traefik+Docker networking

### Backend Version

dev

### Web UI Version

dev

### Package Format

Docker / OCI

### Expected Behavior

The application should clearly indicate whether DT integrations are disabled. 

### Observed Behavior

At the moment, we can only guess whether it is disabled or not working. In some places it shows 'disablesd', in others 'The Dependency-Track integration is currently disconnected. Projects cannot be displayed until the service recovers.'

### Steps to Reproduce

1. Registry -> Artifact > Choose Artifact > Security tab
The following information appears: 'Dependency-Track: Disabled' in red at the top and 'Dependency Track is unavailable' in yellow at the bottom. 
2. Security -> DT Projects
Information stating 'Dependency-Track is unavailable' appears in yellow.

### Logs

```shell

```

### Additional Context

DT Projects:
<img width="832" height="296" alt="Image" src="https://github.com/user-attachments/assets/bac6b7e0-1ba5-482a-8762-cd00e223eb9f" />

Artifact:

<img width="814" height="986" alt="Image" src="https://github.com/user-attachments/assets/3a8ce7c9-1676-4ad6-8ac0-b96d50df405e" />

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have redacted any sensitive information (passwords, tokens, private URLs)
- [x] I am using a supported version (not end-of-life)

PR #2978, opened 2026-07-29
Resolve the Dependency-Track API key lazily so a late bootstrap is picked up

### LINKED ISSUE #2975: Dependency-Track API key is read once at startup, so a successful bootstrap still needs a backend restart
Third of the three bugs described in artifact-keeper-iac#202. The other two are chart-side and are fixed in artifact-keeper-iac#265; this one is ours.

## Problem

`DependencyTrackConfig::from_env()` in `backend/src/services/dependency_track_service.rs` resolves `DEPENDENCY_TRACK_API_KEY_FILE` exactly once, at startup, via `resolve_api_key()`. If the file is not readable at that moment the function returns `None`, `main.rs` wires no Dependency-Track service into application state, and the integration reports disabled for the life of the process.

In a Kubernetes deployment the key file is written by a bootstrap Job that races the backend, and in practice the Job finishes after the backend has already started. So a correct, fully working deployment still ends up with Dependency-Track disabled until someone restarts the backend by hand. The user-visible symptom is "DT is unavailable in the UI even though everything deployed fine."

## Requirements for a fix

- When Dependency-Track is enabled and a key file path is configured but the file is not yet readable, the backend recovers on its own once the file appears. No restart.
- Must not block backend startup indefinitely. A registry that will not serve because an optional SBOM integration is slow is a worse bug than this one. Any startup wait must be short and bounded.
- No behavior change when `DEPENDENCY_TRACK_ENABLED` is false or unset. That is the documented kill-switch (see #1395, #1480) and the `from_env()` doc comment describes the contract precisely.
- `DEPENDENCY_TRACK_API_KEY` keeps taking precedence over the file, and the existing edge cases in `resolve_api_key()`'s doc comment still hold: blank variable, empty path, missing file, and whitespace-only file all yield no key.

Two designs are plausible: a bounded retry at startup, or constructing the service and resolving the key lazily or in the background so it is picked up when it lands. Lazy resolution is more robust to a slow bootstrap; bounded retry is simpler. If lazy, a key that appears later must actually be observed by in-flight requests.

## Note

artifact-keeper-iac#265 closes iac#202. This issue exists so the backend half is not lost when that happens.

brandonrc:
## Summary

`DependencyTrackConfig::from_env()` read `DEPENDENCY_TRACK_API_KEY_FILE` exactly once, at startup. If the file was not readable at that instant it returned `None`, `main.rs` wired no service into application state, and Dependency-Track reported disabled for the life of the process.

In a Kubernetes deployment that file is written by a bootstrap Job that races the backend, and in practice the Job finishes *after* the backend has started. So a correct, fully working deployment still ended up with DT disabled until someone restarted the backend by hand. The user-visible symptom was "DT is unavailable in the UI even though everything deployed fine."

The config now carries the key file path and resolves it **on demand** rather than at construction. Once the file yields a key the result is cached; while it does not, re-reads are throttled so a permanently missing file cannot become a per-request stat storm. Startup is never blocked: the service is constructed immediately and reports unhealthy until the key lands.

### Why lazy rather than a startup retry

A bounded startup retry was the simpler option and was rejected. Blocking backend startup on an optional SBOM integration trades a cosmetic bug for an availability one: a registry that will not serve because Dependency-Track is slow is worse than the problem being fixed. Lazy resolution also handles the unbounded case, where the bootstrap Job is delayed far beyond any retry window anyone would pick.

### Preserved by design

- `DEPENDENCY_TRACK_API_KEY` still takes precedence over the file.
- `DEPENDENCY_TRACK_ENABLED` remains the kill-switch and still short-circuits before any other variable is read (#1395, #1480).
- Blank variable, empty path, missing file and whitespace-only file all still yield no key.

### Related

This is the backend third of artifact-keeper-iac#202. The two chart-side bugs (a bootstrap that installed packages at runtime as non-root and so failed on every cluster, and an empty admin password rejected with 406) are fixed in artifact-keeper-iac#265, now merged.

`from_env_defers_when_key_file_is_not_written_yet` and `api_key_is_picked_up_after_the_file_appears` are the two that fail on `main`: today `from_env()` returns `None` when the file is absent, so there is no service to later pick anything up.

## Test Checklist
- [x] Manually tested locally

Eight new tests, all green alongside the 107 existing tests in the module (115 passed, 0 failed):

| Test | Covers |
|---|---|
| `from_env_defers_when_key_file_is_not_written_yet` | the actual bug |
| `api_key_is_picked_up_after_the_file_appears` | recovery without a restart |
| `requests_recover_once_the_key_file_appears` | in-flight requests observe a late key |
| `health_status_is_unhealthy_until_the_key_lands` | honest health reporting while waiting |
| `key_file_reads_are_throttled_while_the_file_is_missing` | no stat storm |
| `from_env_direct_key_wins_over_key_file` | precedence preserved |
| `from_env_none_when_disabled_even_with_a_readable_key_file` | kill-switch preserved |
| `from_env_none_when_no_key_source_is_configured` | no-source case preserved |

`cargo fmt --check` clean. Built and tested with `SQLX_OFFLINE=true` per the repo convention.

## API Changes

N/A. No endpoints, schemas, or migrations. `DependencyTrackConfig` gains one field, which is internal; the one other construction site is updated in this PR.

Closes #2975

### DIFF
--- a/backend/src/services/dependency_track_service.rs
+++ b/backend/src/services/dependency_track_service.rs
@@ -18,6 +18,15 @@
 //! `/bin/sh` entrypoint wrapper (issues #2059/#2084). The direct
 //! `DEPENDENCY_TRACK_API_KEY` value takes precedence when both are set.
 //!
+//! The bootstrap step that writes the key file runs alongside the backend and
+//! can finish after it, so the file is allowed to appear late: when a key file
+//! is configured but not yet readable, the integration is still wired up and
+//! the key is resolved on first use, retried at most once every
+//! [`DT_KEY_FILE_RETRY_INTERVAL`] until the file lands. No restart is needed.
+//! Requests made before the key is available fail with `ServiceUnavailable`
+//! and the status endpoint reports the integration as unhealthy with the
+//! reason.
+//!
 //! ## Required Dependency-Track team permissions (#1472)
 //!
 //! The API key passed via `DEPENDENCY_TRACK_API_KEY` must belong to a team
@@ -60,7 +69,8 @@ use reqwest::{Client, StatusCode};
 use serde::de::DeserializeOwned;
 use serde::{Deserialize, Serialize};
 use std::error::Error as _;
-use std::time::Duration;
+use std::sync::RwLock;
+use std::time::{Duration, Instant};
 use tracing::{debug, error, info, warn};
 use url::Url;
 use utoipa::ToSchema;
@@ -179,8 +189,14 @@ fn dt_upstream_parse_err(operation: &str, err: impl std::fmt::Display) -> AppErr
 pub struct DependencyTrackConfig {
     /// Base URL of the Dependency-Track API server
     pub base_url: String,
-    /// API key for authentication (X-Api-Key header)
+    /// API key for authentication (X-Api-Key header). Empty when the key is
+    /// supplied through [`Self::api_key_file`] and that file was not readable
+    /// at startup; see [`DependencyTrackService::api_key`].
     pub api_key: String,
+    /// Path the API key is read from when `api_key` is empty. Re-read until it
+    /// yields a key, so a bootstrap step that writes the file after the backend
+    /// starts does not leave the integration dead for the life of the process.
+    pub api_key_file: Option<String>,
     /// Whether integration is enabled
     pub enabled: bool,
 }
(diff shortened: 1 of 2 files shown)


brandonrc:
Two known limitations and one adjacent finding, recorded here rather than left implicit.

**Key rotation is not handled.** The cache is never invalidated after a successful read, so a key rotated on disk is not picked up until the process restarts. Invalidating on a 401/403 would cover it, but that is a different bug from the startup race and the error constructor is a free function rather than a method, so it would touch all eleven call sites again. Worth its own issue if rotation becomes a real workflow.

**The health endpoint now gates on the key.** `/api/version` is unauthenticated, so before this change the status endpoint could report Dependency-Track healthy for an instance we had no usable key for. It now reports unhealthy until the key lands, then flips on its own. That is a behavior change, and it is the honest direction, but it means a deployment mid-bootstrap will briefly show DT unhealthy where it previously showed healthy-but-broken.

**Unrelated, found while verifying:** `cargo test --workspace --lib`, the command documented in CLAUDE.md as requiring no database, does not complete without `DATABASE_URL`. Three modules hang indefinitely at 0% CPU rather than skipping: `storage_gc_service`, `scan_result_service::tests::db`, and `scanner_service::tests::e2e_scan_sbom_matrix`. The db helpers are documented to no-op without `DATABASE_URL`, so something there waits instead of skipping. Also `http_client::test_large_object_client_builder_does_not_apply_total_timeout` fails deterministically in isolation on an unmodified file: it mixes a real loopback TCP connect with `tokio::time::advance`, and the connect times out when the clock jumps. Neither is touched by this PR.

brandonrc:
Closing as superseded: #2975 was completed through a separate change that landed after this PR was opened. Keeping the issue trail clean as part of the release-train triage (#3089).

Issue #2523, opened 2026-07-14
PF-007 [P1]: Quota-enabled uploads scan all live artifact sizes twice (non-atomic admission)


brandonrc:
# PF-007: Quota-enabled uploads scan all live artifact sizes twice

Priority: P1  
Affected area: upload admission and finalization for repositories with quotas

## Finding

Repository storage usage is computed as `SUM(size_bytes)` across every live artifact. A quota-enabled upload invokes it during preflight, then invokes it again after insert for warning telemetry. At one million rows, write admission becomes O(repository artifacts), and concurrent admissions are not atomic with the subsequent artifact insert.

## Evidence

- `get_storage_usage` aggregates all live artifact sizes for a repository: [repository_service.rs](../backend/src/services/repository_service.rs#L1414).
- `check_quota` calls that aggregate whenever a positive quota is configured: [repository_service.rs](../backend/src/services/repository_service.rs#L1446).
- Every shared-service upload calls quota checking in preflight: [artifact_service.rs](../backend/src/services/artifact_service.rs#L568).
- Finalization fetches repository data and repeats the full usage aggregate for the warning threshold: [artifact_service.rs](../backend/src/services/artifact_service.rs#L813).

## Impact at scale

- Every quota-enabled upload reads/aggregates the repository range twice, increasing latency and shared-buffer/CPU pressure as the catalog grows.
- Upload bursts multiply those scans and occupy the small shared DB pool.
- Two concurrent uploads can both observe the old total, both pass admission, and jointly exceed quota. This is a correctness symptom of the same non-atomic accounting design.
- Logical usage and physical CAS-deduplicated usage are different; the current sum enforces logical usage. That semantic should be made explicit before optimizing.

## Recommendation

Maintain an O(1) repository usage ledger updated in the same transaction as artifact insert/resurrection/delete/size replacement. Lock the repository or ledger row for atomic quota admission and delta application. Correctly handle idempotent re-upload and overwrite deltas rather than always adding incoming size.

Keep a cursor-batched reconciliation job that compares the ledger to authoritative artifact rows, repairs drift, and emits a metric. Use the new post-update value for warning telemetry instead of a second aggregate.

If physical storage quota is desired, maintain a separate CAS reference/physical-byte ledger; do not silently change logical quota semantics.

## Acceptance criteria

- Quota admission touches one repository/ledger row and remains constant-time at 1M artifacts.
- Concurrent near-limit uploads cannot over-admit.
[...]

PR #2991, opened 2026-07-29
perf(quota): O(1) admission via the maintained usage-ledger counters

brandonrc:
## Summary

Part of #2516 (Million-artifact readiness), slice S2: make quota admission O(1).

`RepositoryService::check_quota_locked` already serializes same-repo uploads on the
`repository_usage_ledger` row (`SELECT ... FOR UPDATE`, migration 171 / #2523) — but
then ignored the locked row's counters and re-ran a live 3-way `SUM(size_bytes)` over
`artifacts` + `proxy_cache_artifacts` + `oci_blobs` on **every quota-enabled upload**.
In a 500K-row repository that is an O(rows) aggregate executed *inside* the critical
section, so every same-repo upload serializes behind a ~53 ms full scan (measured
below). The unlocked preflight `check_quota` ran the same aggregate again per upload.

Since migration 182 (#2995), row-level triggers on the three source tables keep the
ledger counters exact inside every write's own transaction — for all write paths,
including the format handlers, proxy-cache fills, OCI blob pushes, and deletes. This
PR routes admission through those authoritative counters:

- **`check_quota_locked` is O(1):** it locks the ledger row and decides off
  `hosted_bytes + proxy_bytes + oci_bytes` (PK lookup) plus the existing per-path
  netting (unique-index lookup, so overwrites are still charged only their net delta).
  Admission itself writes **no** counter: the caller's artifact INSERT — performed in
  the same transaction while the row lock is still held — fires the migration-182
  trigger, so the charge and the artifact row commit (or roll back) together, and the
  next admission serialized on the row lock always observes them. There is no
  double-counting: the bytes are charged exactly once, by the trigger. The
  `quota <= 0` unlimited fast path stays lock-free and never touches the ledger.
- **Lazy-create seeds from the live sums, not zeros:** a repository whose ledger row
  is missing (pre-171 repo never backfilled, or manual surgery) pays a one-time
  reconcile on its first quota-checked upload (row created + locked in the same
  transaction). Trusting a default-zero row would have admitted everything until the
  first reconciler pass.
- **`reconcile_usage_ledger` locks before it computes:** the previous
  compute-then-upsert (unlocked, on the pool) could overwrite a concurrent admission's
  just-committed charge with stale sums. It now takes the same ledger-row lock
  admission holds, *then* reads the source tables, in one transaction (shared helper
  `reconcile_usage_ledger_in_tx`). Post-182 the reconciler is a drift safety net only.
- **Configuring a finite quota trues the ledger up synchronously** (`update()`):
  defensive against residual pre-182 drift, so enforcement never starts from stale
  counters. Best-effort; the background reconciler stays the safety net.
- **The chunked-upload `complete` handler runs the locked admission** in the same
  transaction as its artifact INSERT. Previously this path's only quota gate was the
  unlocked session-create preflight, so (pre-existing) two concurrent near-limit
  sessions could both complete past the quota — #2523's race, closed for the
  direct-write path in `finalize_upload` but never here. Over-quota completes now
  return 507 `QUOTA_EXCEEDED`. Peer replication keeps its documented exemption, same
  as the session-create gate.
- **Preflight `check_quota`** (its only callers are the upload preflights) reads the
  same counters (O(1) point read), falling back to the live sum only for repositories
  with no ledger row yet. Display paths (`get_storage_usage`,
  `get_display_storage_usage`, stats) keep the live figures unchanged.
- **`ArtifactService::delete_with_sync_options` keeps its simple soft-delete:** the
  migration-182 trigger decrements `hosted_bytes` in the soft-delete's own
  transaction, so freed space is admissible by the very next upload with no manual
  ledger write in the service.

All three usage components (hosted / proxy-cache / OCI blobs) remain part of the
enforced total, exactly as before — and, unlike pre-182, uploads that land through
format handlers which bypass this admission path are charged immediately by the
triggers, so admission always sees them. No migration needed here: 171 carries the
columns, 182 carries the triggers and the true-up backfill.

Plan-shape witness on a seeded 500K-row repository (ANALYZEd, filesystem profile):

- before — admission read under the held lock:
  `Finalize Aggregate -> Parallel Seq Scan on artifacts` (rows=500000,
  shared hit=15745 read=335, **53.1 ms** per upload, scales with repo size)
- after — admission read under the held lock:
  `LockRows -> Index Scan using repository_usage_ledger_pkey` (shared hit=4,
  **0.08 ms**, invariant to repo size); the overwrite net-delta probe stays an
  `Index Scan using idx_artifacts_repo_path_pattern` (1 row, 0.1 ms)

New DB-backed tests (skip cleanly without `DATABASE_URL`, run for real in CI). They
assert **end-state** ledger values (what the triggers must produce), not the charging
mechanism:

- `test_quota_admission_o1_under_then_over` — under-quota admitted, over-quota
  rejected, and the ledger ends exact after each admission (charged once by the
  insert trigger, never double-counted by admission).
- `test_quota_admission_concurrent_joint_excess_admits_exactly_one` — two concurrent
  uploads that would jointly exceed the quota: the second blocks on the ledger-row
  lock and is rejected once the first commits; exactly one artifact lands.
- `test_quota_admission_frees_space_after_delete` — a delete outside the admission
  path returns the ledger to the exact live value in the delete's own transaction;
  the freed space is admissible by the very next upload, with no phantom credit.
- `test_quota_admission_counts_proxy_and_oci_components` — proxy-cache + OCI-blob
  bytes still count against the quota.
- `test_quota_admission_lazy_seed_reads_live_sums_not_zero` — a data-holding
  repository without a ledger row is not over-admitted through a fresh zero row.
- `test_quota_admission_overwrite_charges_net_delta` — overwrites end at the net
  value; no double-counting.
- `test_quota_admission_unlimited_skips_ledger` — `NULL`/`0`/negative quotas stay
  lock-free and never create the ledger row.
- `test_update_with_finite_quota_reconciles_ledger` — setting a finite quota trues
  the ledger up from injected stale counters before enforcement begins.
- `test_delete_releases_ledger_bytes_immediately` (artifact_service) — the service
  delete path releases the bytes in its own transaction; a double-delete maps to
  NotFound and never releases the same bytes twice.
- `test_check_quota_locked_has_no_live_union_aggregate` — source-contract pin: the
  admission critical section must never reintroduce the 3-way `UNION ALL` aggregate.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

(Behavior note: chunked-upload completes that exceed the quota now return 507
`QUOTA_EXCEEDED` at `PUT /uploads/{id}/complete` in the race window the unlocked 413
session preflight cannot see — previously they were silently admitted past the quota.)


brandonrc:
Security review (red-team, live deploy of head `77757d7`) — **FINDINGS, holding for a follow-up before merge.**

**Held / could not be broken (good):** admission race (10 concurrent uploads into quota=1000 → exactly 5 admit / 5 reject; 2-way joint-excess → exactly one wins), chunked-upload `complete` lock (held in the same tx as its INSERT), lazy-seed race, cross-tenant isolation (every ledger read/charge keyed `WHERE repository_id = $1`). The DTF tier is genuinely green.

**Finding 1 (HIGH) — freshness-window over-admission.** This PR moves admission from a live 3-way SUM under the lock to an O(1) read of the ledger counters. But many write paths (maven/npm/cargo/goproxy/oci, promotion, proxy fills) do not charge the ledger in-tx, so the counters can lag real usage. Repro: repo quota=1000 → enforced PUT 200B (ledger=200) → uncharged maven PUT 5000B (real=5200, ledger still 200) → enforced PUT 400B **admitted** off the stale 200 → final real=5600 vs quota=1000 (**5.6×**). On the prior code the under-lock re-SUM (5200) returned 507. Self-heals on the 30-min reconciler, so bounded per window but re-openable.

**Finding 2 (MEDIUM) — decremented delete of an un-charged artifact drives the ledger below reality** → phantom free quota (GREATEST(charged−uncharged_size,0)=0). The added tests reason only about the conservative/idempotent direction and miss this.

Both findings share one root cause: the ledger is not authoritative because not all write paths charge it. Correct fix = make every artifact-write path charge the ledger in the same tx as its INSERT; then admission is O(1) **and** exact, and Finding 2 dissolves (charged == decremented). Routing to a follow-up spec; this stays draft until it lands.

Issue #2954, opened 2026-07-28
Inline scan-and-block on download: enforce scan results + policy at serve time (proxy + hosted)


brandonrc:
## Problem (confirmed live on v1.6.3)
Artifact Keeper has **no inline, download-time security enforcement**. Scanning + policy only block at the **promotion gate** (staging→release); nothing gates the actual pull:
- **Proxy:** `scan_on_proxy` is a surfaced-but-unimplemented flag — it logs the #1653/#1274 "cached UNSCANNED" warning and serves the package. Proxy-cached content isn't recorded in `artifacts` (#1280), so the scanner (which operates on `artifacts` rows) never runs. Verified: `Pillow-9.0.0` (known criticals) pulled through a remote PyPI returned HTTP 200, 0 scan_results.
- **Hosted:** scan-on-upload genuinely runs (Grype found 2 crit/15 high on a test wheel), but the **download path does not consult `scan_results`** — the vulnerable hosted wheel still downloads HTTP 200. `block_on_policy_violation`/`block_unscanned`/`block_on_fail` are not enforced on any download path.
- **Curation typed rules (#2947):** same shape — publisher-trust/popularity decisions enforce at the staging-catalog/sync layer, not inline on the pull.

The requested behavior (a  through the registry is scanned against the CVE DB and blocked if the resolved version is vulnerable) does not exist.

## Goal
A **serve-time policy enforcement gate** on the download path that, per the repo's scan config + security policy, blocks delivery of an artifact that violates policy — for **proxy** and **hosted** — and, as a natural extension, lets the #2947 typed curation rules (publisher-trust/popularity) enforce inline too.

## Design questions / phasing (design spike to resolve)
1. **Proxy scanning seam:** scan-on-fetch (scan upstream bytes before first serve) vs. a scannable proxy-cache record + async scan + gate. Handle the #1280 constraint (proxy content not in `artifacts`).
2. **Verdict caching keyed by content digest** so repeat pulls are fast (no re-scan); TTL/invalidation.
3. **First-pull-of-unknown behavior:** fail-closed (block until scanned) vs fail-open (serve then scan) — driven by `block_unscanned`/`block_on_fail`.
4. **Hosted download gate:** the smaller, immediately-valuable piece — download path consults existing `scan_results` + policy and blocks. (Also fixes the `block_on_policy_violation` false-affordance.)
5. **Latency budget** for inline scanning on the hot serve path; bounded wait + async completion.
[...]

Issue #2555, opened 2026-07-14
Scan concurrency: add per-user/tenant fairness (global FIFO is monopolizable → degraded scan freshness)


brandonrc:
Follow-up from #2540/#2554 red-team (LOW, non-blocking). The #2540 global scan-extraction semaphore is one global FIFO with no per-user fairness, so a single authenticated user can saturate all N permits and delay other users'/admin's scan COMPLETION (freshness), though not availability (triggers return 200 <30ms, all scans complete FIFO, permits release correctly, /health unaffected). Add a per-user/per-tenant sub-quota or a small guaranteed permit reserve so one flooder can't monopolize scan capacity. Origin: PR #2554.

PR #2994, opened 2026-07-29
fix(scan): thread live CVE-scanner version into proxy verdict freshness check

### LINKED ISSUE #2976: Proxy scan verdict cache ignores live scanner-DB version → clean verdict reused up to TTL after a CVE-DB bump
## Summary
Follow-up from the #2954 inline scan-and-block red-team review (LOW; pre-existing, not introduced by #2954).

At serve time, `lookup_verdict` passes `current_version=None` (`backend/src/api/handlers/pypi.rs` ~L2806), so `verdict_is_fresh` never compares the verdict's scanner version against the **live** Grype DB version. A cached `clean` verdict for a given content digest is therefore reused for the full `DEDUP_TTL_DAYS` (30d) window **even after the vulnerability DB has been updated** to newly know about a CVE in that content.

## Impact
LOW. A package that was genuinely clean at scan time but is later revealed vulnerable by a CVE-DB update keeps serving `200 clean` from cache until the 30-day TTL expires (or the digest changes). Bounded by TTL; does not defeat fail-closed on a fresh/unknown digest.

## Fix direction
Thread the live scanner version into the serve-time `lookup_verdict` call so `verdict_is_fresh` invalidates a cached verdict when the scanner DB version has advanced past the one that produced it (re-scan on DB bump). Pairs naturally with the #2954 proxy_scan_results freshness logic.

## Notes
Found during authorized red-team of PR #2970 (commit d232dbf); the primary CRITICAL (fail-closed serve-of-vulnerable on scanner error) is fixed and re-red-teamed CLEAN — this is the residual LOW recommended as a separate follow-up.

brandonrc:
## Summary

Closes #2976.

At serve time the PyPI proxy verdict fast path called `verdict_is_fresh` with
`current_version=None`, so the scanner-version comparison the function was
built for never ran: a cached `clean` verdict for a content digest was reused
on TTL alone (`DEDUP_TTL_DAYS` = 30d), even after the CVE-DB advanced to newly
know a CVE in those bytes. The proxy kept serving `200 X-AK-Scan: clean` from
the stale cache until TTL expiry or a digest change.

This PR threads the live CVE-engine version into the freshness check:

- `ScannerService::cve_scanner_version()` (new): returns the version string of
  the CVE-authoritative scanner (Grype), i.e. the SAME `grype-X.Y.Z`
  provenance string `run_inline_proxy_scanners` persists on
  `proxy_scan_results.scanner_version`. It reads `Scanner::version()`, which
  is `VersionCache`-backed (1h hit TTL), so the hot path never shells out per
  download.
- The serve-time call site in `serve_scanned_pypi_file` now passes that value
  as `current_version` instead of `None`. The version comparison itself is
  unchanged and still lives in `verdict_is_fresh`: a stored/live mismatch
  (both known) marks the verdict stale.
- A stale verdict falls through to the existing re-scan branches, preserving
  fail-closed alignment: an invalidating re-scan that finds CVEs is a 403, an
  inconclusive re-scan is a 423 — never a 200 of stale-clean bytes.
- `verdict_is_reusable` (new, pure): the policy-aware gate the serve path now
  asks. `verdict_is_fresh` deliberately fails OPEN on an unknown version
  (probe failed / legacy row) — it cannot prove a mismatch, so it relies on
  the TTL. That default is right for fail-open but is a hole under
  `fail_closed`, because the freshness fast path runs BEFORE the inline scan:
  an unprovable `clean` verdict short-circuited the entire fail-closed gate
  (including the #2954 "the CVE engine actually ran" condition) and served
  cached bytes nothing on the node could vouch for — while a fresh pull on the
  same node returned 423. That window is common exactly when it matters: a
  Grype upgrade (the CVE-DB advance this issue is about) transiently fails
  `grype --version` for at least the 60s miss-TTL, a missing binary makes it
  permanent, and a loaded host can push the probe past its 5s timeout. So
  under `fail_closed` a CLEAN verdict is reusable only when its provenance is
  PROVEN current (both versions known and equal); otherwise the pull
  re-scans, and the inconclusive outcome correctly locks. It self-heals: the
  re-scan records the live version, so the next pull hits the cache.

`fail_open` is deliberately untouched (its re-scan branch serves-with-pending
anyway, so availability is unaffected), and non-`clean` verdicts keep their
existing handling — a cached `vulnerable` verdict still blocks.

No schema change, no new sqlx macros.

`test_serve_file_proxy_scan_rescans_when_scanner_version_advances` drives
`serve_file` end-to-end with a seeded `clean` verdict at stored version V and
a mock CVE-authoritative scanner reporting V+1: on `main` the pull serves 200
stale-clean from cache; with this PR the verdict is invalidated, the inline
re-scan runs, and the pull is blocked 403.

## Test Checklist
  - `test_cve_authoritative_scanner_version_picks_cve_engine_only` — the
    accessor returns the CVE engine's version (skipping non-authoritative
    scanners regardless of registration order) and `None` when no engine is
    registered / the probe failed.
  - `unknown_version_is_not_reusable_for_clean_under_fail_closed` — the full
    matrix: unknown on either side (or both) is still "fresh" per
    `verdict_is_fresh` but NOT reusable for a `clean` verdict under
    `fail_closed`; fail-open and cached-`vulnerable` handling are unchanged.
  - `reusable_matches_fresh_when_both_versions_known` — a proven match is
    reusable under both actions, a proven mismatch and an expired TTL are not.
  - `test_serve_file_proxy_scan_rescans_when_scanner_version_advances` —
    stored V, live V+1, re-scan vulnerable → 403 (was 200 stale-clean).
  - `test_serve_file_proxy_scan_unknown_live_version_rescans_under_fail_closed`
    — stored V, live probe returns None, `fail_closed`: must not serve
    cached-clean; re-scan → 403.
  - `test_serve_file_proxy_scan_unknown_live_version_serves_under_fail_open` —
    the availability control for the above: `fail_open` still serves 200 from
    the cache on an unknown probe.
  - `test_serve_file_proxy_scan_reuses_cached_verdict_same_version` — stored
    version equals live → served from cache (the mock would 403 on a re-scan,
    so the 200 proves no needless re-scan / no perf regression).
  - `test_serve_file_proxy_scan_version_bump_inconclusive_locks` — stored V,
    live V+1, inconclusive re-scan → 423 under fail-closed (never
    stale-clean).
  - `test_serve_file_proxy_scan_serves_cached_clean_digest` (existing, #2954)
    now exercises the cached-clean fast path on a `fail_open` repo. It
    previously asserted the same 200 under `fail_closed`, which passed only
    because a state with no scanner service reports an unknown version — i.e.
    it encoded the fail-open hole described above.
- [x] Manually tested locally
  - Isolated instance, fail-closed remote PyPI repo over pypi.org: seeded a
    `clean` verdict for the Jinja2-2.10 wheel digest at an old version while
    the bundled engine is `grype-0.116.0` → pull returned `403 scan_blocked`
    and the row was upserted to `vulnerable | grype-0.116.0` (re-scan
    occurred). A second CVE wheel seeded `clean` AT `grype-0.116.0` served
    `200 X-AK-Scan: clean` with the row untouched (matching provenance is
    honoured).
  - Unknown-probe repro on the same instance: grype binary stubbed and the
    backend restarted (dropping the in-process version cache), so the probe
    returns None. A FRESH digest returns 423 (the node is provably
    fail-closed) and the cached-clean digest now also returns 423 with a
    `live CVE-scanner version unknown; not reusing the cached verdict` warn —
    it returned `200 X-AK-Scan: clean` before this change. The same digest on
    a `fail_open` repo on that same node still serves 200.
- [x] No regressions in existing tests (`cargo nextest run --workspace --lib`
  with a migrated Postgres: 14341 passed / 0 failed; fmt + clippy
  `-D warnings` clean; jscpd 0.98%, no clones in the changed files)

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -2918,20 +2918,41 @@ async fn serve_scanned_pypi_file(
         // below (fail-closed: inline re-scan → 403/423; fail-open: serve
         // pending + async re-scan) instead of serving stale-clean for the full
         // TTL window. `Scanner::version()` is VersionCache-backed, so this is
-        // a memory read on the hot path, never a per-download subprocess. When
-        // either side is unknown, `verdict_is_fresh` falls back to the TTL.
+        // a memory read on the hot path, never a per-download subprocess.
+        //
+        // The repo's action is part of the decision, not just the versions:
+        // this fast path runs BEFORE the inline scan, so on a fail-closed repo
+        // an UNKNOWN live version (probe failed mid-upgrade, binary absent,
+        // probe past its timeout) must NOT let a cached `clean` verdict
+        // short-circuit the gate — see `verdict_is_reusable`.
+        //
+        // No per-digest singleflight here: N concurrent pulls of the same
+        // newly-stale digest each re-scan. They queue behind the #2555
+        // fair-share extraction semaphore so this cannot starve upload scans,
+        // and the window closes as soon as the first re-scan upserts the row.
+        // Coalescing is a possible follow-up, not a correctness requirement.
         let current_version = match state.scanner_service.as_deref() {
             Some(scanner) => scanner.cve_scanner_version().await,
             None => None,
         };
-        let fresh = crate::services::proxy_scan_service::verdict_is_fresh(
+        let reusable = crate::services::proxy_scan_service::verdict_is_reusable(
+            &row.verdict,
             row.scanned_at,
             row.scanner_version.as_deref(),
             current_version.as_deref(),
+            action,
             crate::services::scanner_service::DEDUP_TTL_DAYS as i64,
             Utc::now(),
         );
-        if fresh {
+        if !reusable && current_version.is_none() && action.is_fail_closed() {
+            warn!(
+                repo_id = %repo_id, file = %filename, digest = %digest,
+                stored_version = ?row.scanner_version,
+                "live CVE-scanner version unknown; not reusing the cached verdict \
+                 on a fail-closed repo (re-scanning)"
+            );
+        }
+        if reusable {
             if crate::services::proxy_scan_service::verdict_blocks(&row.verdict) {
                 warn!(repo_id = %repo_id, file = %filename, digest = %digest, "blocking proxy pull: cached vulnerable verdict");
                 return Err(scan_blocked_response(filename));
@@ -9800,6 +9821,19 @@ mod tests {
         }
     }
 
+    /// The cached-`clean` fast path serves 200 `X-AK-Scan: clean` (#2954) —
+    /// exercised here on a FAIL-OPEN repo with no scanner service, where the
+    /// header is the discriminator: without the cached verdict this pull would
+    /// still be a 200, but marked `pending`.
+    ///
+    /// This test used to assert the same 200 under FAIL-CLOSED. That passed
+    /// only because a state with no scanner service reports an unknown live
+    /// scanner version, which the freshness check treated as "not provably
+    /// stale" and served — i.e. it encoded the #2976 fail-open hole: a node
+    /// with no working CVE engine served cached-clean bytes through a
+    /// fail-closed policy that 423s a fresh pull. The fail-closed side of that
+    /// pair now lives in
+    /// `test_serve_file_proxy_scan_unknown_live_version_rescans_under_fail_closed`.
     #[tokio::test]
     async fn test_serve_file_proxy_scan_serves_cached_clean_digest() {
         use crate::api::handlers::test_db_helpers as tdh;
@@ -9815,10 +9849,7 @@ mod tests {
 
         let upstream = wiremock::MockServer::start().await;
         mount_scan_upstream(&upstream, project, filename, wheel).await;
-        // Fail-closed: without the fresh clean verdict this pull would 423
-        // (no scanner service on the state), so a 200 here proves the cached
-        // verdict fast path served it.
-        enable_proxy_scan(&fx.pool, fx.repo_id, "fail_closed").await;
+        enable_proxy_scan(&fx.pool, fx.repo_id, "fail_open").await;
 
         ProxyScanService::new(fx.pool.clone())
             .record_verdict(
@@ -9865,7 +9896,7 @@ mod tests {
             Err(r) => {
                 let status = r.status();
                 fx.teardown().await;
-                panic!("fresh clean verdict must serve under fail-closed, got {status}");
+                panic!("fresh clean verdict must serve from the cache, got {status}");
             }
         };
         let status = resp.status();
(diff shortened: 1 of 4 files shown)


brandonrc:
Security review (red-team, live deploy of head `69650ab`) — **one HIGH, holding for a fix before merge.**

**Confirmed on happy paths (good):** same-version cache reuse → 200; probe=Some + DB-version advance → re-scan → 403 vulnerable / 423 inconclusive (both live-verified). Version-string spoofing HELD (stored `scanner_version` comes from the scanner's own `version()`, not attacker input).

**Finding (HIGH) — probe-failure fallback re-opens #2976.** When `cve_scanner_version()` returns `None`, `verdict_is_fresh`'s `_ => true` arm treats the verdict as fresh (fail-OPEN) and serves a stale **clean** verdict (200) for genuinely-vulnerable content — on a `fail_closed` repo. Live repro: seed a stale `clean|grype-OLD` verdict for a vulnerable wheel; with grype working → 403 (re-scan blocks); with the version probe forced to `None` (grype absent/restarting) → **200 X-AK-Scan: clean, served the vulnerable bytes, no re-scan**, while a fresh pull on the same node returns 423 (proving fail-closed posture). `None` is reachable and common precisely when #2976 matters: a grype **upgrade** (the CVE-DB advance) transiently fails `grype --version` → ≥60s (`VERSION_CACHE_MISS_TTL`) window of stale-clean; grype absent = permanent no-op; `--version` past the 5s timeout under load. The freshness fast-path also short-circuits before `proxy_scan_action` is consulted and before the #2970 `cve_scanner_applicable && !cve_scanner_ran` gate.

**Fix direction:** thread `proxy_scan_action` into the freshness decision — on `fail_closed`, an UNKNOWN live version must NOT be treated as `version_ok=true` for a CLEAN verdict; force the re-scan branch (whose inconclusive→423 fail-closes) instead of serving cached-clean. Do not let `current_version=None` short-circuit the fail-closed gate.

**Non-blocking (MEDIUM, note for follow-up):** the forced re-scan has no per-digest coalescing → N concurrent pulls of the same stale digest each re-scan (bounded by the #2555 extraction semaphore, so no worker-starvation, but wasteful); and a routine DB bump on a fail_closed repo can cause a mass-423/latency bump as many distinct stale digests re-queue. Consider singleflight per digest. Holding draft until the HIGH lands.

brandonrc:
Re-red-team (live deploy of head `c72fb5a` on an isolated slot) — **CLEAN, HIGH closed.** The exact original exploit now fail-closes: stale `clean|grype-OLD` verdict for a genuinely-vulnerable wheel + version-probe forced to `None` on a `fail_closed` repo → **HTTP 423**, vulnerable bytes NOT served (log: 'live CVE-scanner version unknown; not reusing the cached verdict on a fail-closed repo'). Working-probe legs unregressed (version-advance → 403 + row self-heals; same-version → cached 200, no re-scan); fail_open unchanged (200, availability preserved); no leftover `verdict_is_fresh` serve-path caller; `proxy_scan_action` is read from scan_configs by repo_id and is not request-spoofable; the legacy-NULL re-scan is bounded by the #2555 fair-share semaphore (10 concurrent → all 403, ~2.9s, no 5xx) and self-heals; and the re-scan branch is backstopped by the #2970 `cve_scanner_applicable && !cve_scanner_ran` gate so grype-absent → 423, never spurious-clean. No new bypass or availability-to-serve regression. Clear to merge.

PR #2999, opened 2026-07-29
fix(lifecycle): retain Docker tags per image

### LINKED ISSUE #2998: Docker max_versions groups each tag separately and never prunes an image
## Description

`max_versions` retention groups artifacts by `artifacts.name`.

For Docker/OCI manifest artifacts, `name` includes the tag:
`namespace/image:v1`, `namespace/image:v2`, `namespace/image:v3`.

Therefore every tag becomes a separate one-item partition. A repository-scoped
`max_versions` policy can be created and executed successfully, but it finds
nothing to remove regardless of the configured `keep` value.

The existing OCI cascade regression test masks this: it rewrites the manifest
artifact names to the same synthetic value before executing `max_versions`.

## Environment Information

- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL
- Reverse Proxy: N/A
- Network setup: private deployment

## Backend Version

1.5.1 (the same behaviour is present on current `main`)

## Web UI Version

N/A

## Package Format

Docker / OCI

## Expected Behavior

For a Docker repository, `max_versions: {"keep": 1}` should retain the newest
tag of each image independently.

Three tags of one image — `namespace/image:v1`, `namespace/image:v2`, and
`namespace/image:v3` — should be treated as three versions of
`namespace/image`, leaving only `v3`.

Different images must remain independent partitions. A registry port in an
image reference must be preserved, for example
`registry.example:5000/namespace/image:v3`.

## Observed Behavior

The policy completes without an error but removes zero artifacts. Each full
`image:tag` value is considered a distinct artifact name, so each group has
only one item and is already within the retention limit.

## Steps to Reproduce

1. Create a local Docker repository.
2. Push three manifest tags for the same image:
   `namespace/image:v1`, `namespace/image:v2`, and `namespace/image:v3`.
3. Create a repository-scoped lifecycle policy:

   ```json
   {
     "name": "Keep latest Docker tag",
     "repository_id": "<docker-repository-id>",
     "policy_type": "max_versions",
     "config": { "keep": 1 }
   }
   ```

4. Run its preview or execute endpoint.
5. Observe `artifacts_matched: 0` and `artifacts_removed: 0`.

## Logs

No backend error is emitted; execution reports success with zero matched
artifacts.

## Additional Context

The relevant query partitions by `a2.name = a.name`. For Docker repositories,
the partition key should instead be the image reference without its final tag.

A candidate fix is to use:

```sql
regexp_replace(name, ':[^:]+$', '')
```

only when `repositories.format = 'docker'`; this removes the final tag while
preserving a registry port.

A regression test should use actual distinct Docker tag names rather than
rewriting them to one synthetic name. The existing OCI cascade test is related,
but its name rewrite currently hides this bug.

## Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate.
- [x] I have redacted any sensitive information (passwords, tokens, private URLs).
- [x] I am using a supported version (not end-of-life).

ivolnistov:
## Summary

Closes #2998.

Docker manifest artifacts encode `image:tag` in `artifacts.name`, so the original
`max_versions` query treated every tag as a separate one-item partition and never
pruned old tags. For Docker repositories this PR partitions valid tag-shaped
references by image name while preserving registry ports.

The partition key deliberately keeps `image:sha256:<encoded>` digest references
as distinct full-name partitions. This avoids grouping multi-architecture child
manifests, which can be stored by digest without an `oci_tags` row. It also avoids
collapsing tagless synthetic names containing a registry port.

## Test Checklist

- [x] Manually tested locally

Validation performed:

- `cargo fmt --check`
- `cargo test --lib services::lifecycle_service::tests::test_max_versions_groups_docker_tags_by_image --no-run`
- Fresh PostgreSQL schema (all 168 migrations), then
  `cargo test --lib services::lifecycle_service::tests::test_max_versions_groups_docker_tags_by_image -- --exact`
- Built the production backend Dockerfile on Linux and ran the resulting image
  with an isolated PostgreSQL container. Through the admin API, `max_versions:
  {"keep": 1}` soft-deleted exactly the two old tags while retaining the newest
  tag, another image, two digest references, and two tagless names with a
  registry port.

## API Changes

- [x] N/A - no API changes


brandonrc (approved):
Approving. Full pipeline: reproduced #2998 on the base (Docker repos matched 0 while the Maven control correctly matched 2), verified the fix with negative controls, reviewed, then pushed remediation as `1d24853` and re-verified.

The grouping diagnosis is right and the root cause — `oci_v2` writing `artifact_name = format!("{}:{}", image, reference)` so every tag is its own singleton partition — is exactly it.

Four things changed after review, each verified against a live database:

* **Shape.** The correlated `NOT IN` wraps the indexed `name` in `regexp_replace` on both sides, so the predicate is non-sargable and the subplan degrades to a seq scan per outer row: 4 ms → 212 s at 2,000 artifacts, timing out at 10,000 — inside the transaction `execute_policy` deliberately keeps short so `artifacts` row locks release promptly, and paid twice per run. Now a `row_number()` window: one pass, one regex per row.
* **Cosign.** `image:sha256-<digest>.sig`/`.att` match the tag regex and are written *after* the manifest they sign, so they're newer and win the slots. On a `keep = 2` fixture the pre-fix query deleted **all four real image tags and kept both signatures**. They're now singleton partitions, like digest references already were.
* **Formats.** All six formats routing to the OCI handler share this naming — a `podman` repo matched zero. Gating on `docker` alone left #2998 open for podman, buildx, oras, wasm_oci and helm_oci.
* **Recency.** Ordering stayed on `created_at`, which #3027 establishes is wrong for a mutable tag: a manifest PUT upserts the row without touching `created_at`, so a rolling tag is always the *oldest* row of its image and is evicted first however recently it was pushed. Now `COALESCE(oci_tags.updated_at, created_at)` with `a.id` as a deterministic tie-break.

Also fixed `test_max_versions_cascades_oci_tags`, which **fails on this branch as it stands** — its two "surviving sibling" artifacts were both named `other:*`, so they shared a partition and `keep = 1` pruned one, making `artifacts_removed` 3 rather than 2. Renamed them to genuinely distinct images so it still guards the #1682 cascade. Worth knowing: that suite only runs on main and release branches, so this would have merged green and broken main on arrival — a gating gap worth addressing separately.

The new lib test covers all three behavioural changes and fails if any one is reverted; the existing docker-only test cannot see them. All 10 checks green, clippy clean, 14,396 lib tests / 0 failures.

Not merging — labeled `ready-to-merge`.

PR #2993, opened 2026-07-29
fix(auth): accept colon-form write:artifacts scope on artifact upload paths

brandonrc:
## Summary

Repo-scoped tokens minted via `POST /api/v1/repositories/{key}/tokens` carry the colon-form scope vocabulary (`write:artifacts`, `read:artifacts`, ...), but the artifact upload/mutation handlers enforced the bare `write` scope via `require_scope("write")`. The two vocabularies never matched, so a legitimate least-privilege `write:artifacts` repo-scoped token received 403 on every upload path. The mismatch failed closed (no privilege escalation), but it broke the repo-scoped-token flow entirely.

This PR fixes the vocabulary mismatch with one centralized rule plus a handler-side scope rename:

1. **Central matcher** (`token_service::scopes_grant_access`, the single scope-decision choke-point used by `AuthExtension::has_scope`, `require_scope_response`, and the gRPC interceptor): a held scope now satisfies a required scope iff
   - it equals the required scope exactly, or
   - it is `*` / `admin` (unchanged wildcard short-circuit), or
   - the required scope is colon-form (`action:resource`) and the held scope is its **bare action parent** (held `write` satisfies required `write:artifacts`).

   The direction is deliberately **broad-covers-specific only**. A held colon-form scope satisfies neither a bare requirement (`write:artifacts` does NOT satisfy `write`, which still gates settings-class handlers) nor a colon-form requirement for a different resource (`write:artifacts` does NOT satisfy `write:repositories`, and vice versa). Read scopes are untouched beyond the same symmetric parent rule.

2. **Artifact-write handlers** now require the resource-specific `write:artifacts` instead of bare `write` (13 call sites): the three chunked-upload verbs (`create_session` / `upload_chunk` / `complete`), the three generic upload handlers (`upload_artifact`, both multipart variants), artifact label mutation, and the format publish endpoints that used `require_scope_response(_, "write")` (npm publish + dist-tags put/delete, cargo publish, nuget push, goproxy PUT). Non-artifact `require_scope("write")` sites (repository settings, projects, groups, permissions, promotion release-target) are intentionally unchanged, so bare `write` remains their only accepted vocabulary and a `write:artifacts` token still cannot reach them.

Satisfaction table implemented:

| held | required `write:artifacts` | required bare `write` | required `write:repositories` |
|---|---|---|---|
| `write` (bare, global) | allow | allow | allow |
| `write:artifacts` | allow | deny | deny |
| `write:repositories` | deny | deny | allow |
[...]

Issue #2989, opened 2026-07-29
Repo-scoped tokens carry colon-form scopes (write:artifacts) but upload handlers check the bare scope (write) → token 403s on upload


brandonrc:
## Summary
Pre-existing scope-vocabulary mismatch found during the #2603 mutation-authz red-team (not introduced by that work — present ~17× on the base branch).

A repository-scoped token minted via `POST /repositories/{key}/tokens` carries **colon-form** scope strings such as `write:artifacts`. But the artifact upload handlers enforce the **bare** scope via `require_scope("write")` (the bare-`write` check appears ~17× across handlers). The two vocabularies don't match, so a repo-scoped token that legitimately holds `write:artifacts` is rejected **403** when used to upload — the repo-scoped-token feature is effectively broken for its primary use case (CI pushing to a single repo with a narrowly-scoped token).

## Impact
Functional (not a security hole — it fails *closed*): users who follow the documented repo-scoped-token flow cannot upload with that token and must fall back to a broader personal/global token, defeating the least-privilege intent.

## Fix direction
Normalize the scope vocabulary end to end: either (a) mint repo-scoped tokens with the bare scope the handlers check, or (b) teach `require_scope`/the scope-matching to treat `write:artifacts` as satisfying a `write` requirement (colon-form ⊇ bare). Pick one canonical form and make mint-time and check-time agree; add a test that a repo-scoped `write:artifacts` token can actually upload to its repo (and still can't to another).

## Notes
Surfaced during authorized red-team of PR #2987 (#2603 Part 1); that PR's G3 delegation-ceiling is correct and unaffected — this is the orthogonal mint-vs-check vocabulary bug.

Issue #2603, opened 2026-07-16
Security hardening follow-up: complete repository RBAC and admin-surface authorization


valfon1393:
### Description

Proposed fix #2610 

This is a follow-up to #2321 after the related hardening work was split into
upstream changes and released. I retested the released `v1.5.7` backend with the
same authorization matrix.

The release contains meaningful improvements, and the matrix confirms that
several previously reported areas are now protected. However, some
authorization gaps remain across repository mutations, repository
administration, and selected global administrative surfaces.

The remaining hardening areas are:

1. Public repository visibility should grant read visibility only. It should
   not become implicit write or delete permission when a repository has no
   explicit permission rules.
2. Generic artifact APIs, resumable uploads, and native package publish
   handlers should enforce the same action-specific repository authorization
   model.
3. Repository administration subresources and repository-scoped token
   delegation should require repository-admin or global-admin authorization.
4. Selected replication, signing, security-configuration, and migration
   surfaces still need consistent global-admin gates.
5. Protected handlers should reject unauthorized callers before request-body
   validation, resource lookup, or mutation.

I am intentionally keeping this public issue high-level. I am not publishing
endpoint-by-endpoint reproduction commands, request bodies, deployment
identifiers, account details, repository names, group names, or tokens. I have
prepared a sanitized private report with the exact affected request classes and
can share it directly with a maintainer through the existing private security
thread or by email.

### Environment Information

- Deployment type: containerized backend deployment
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL
- Reverse Proxy: present; implementation details redacted
- Network setup: private deployment; topology and hostnames redacted
- Auth modes tested: local authenticated non-admin user and SSO authenticated
  non-admin user

### Backend Version

1.5.7

### Web UI Version

1.5.7

### Package Format

N/A / General

### Expected Behavior

- Public visibility permits read/download according to repository policy, but
  does not imply mutation permission.
- Artifact create or overwrite requires explicit repository `write` permission
  or global admin.
- Artifact deletion requires explicit repository `delete` permission or global
  admin.
- Repository configuration and repository-scoped token delegation require
  repository `admin` permission or global admin.
- Global administrative and security surfaces require global admin unless a
  narrower privilege model is explicitly designed and tested.
- Authorization runs before validation, lookup, or state changes on protected
  endpoints.

### Observed Behavior

The fresh authorization matrix identified inconsistent authorization gates in
[...]

PR #3001, opened 2026-07-29
fix(auth): validate and bound token scopes at the mint choke-point

### LINKED ISSUE #2996: Token mint path does not validate requested scopes (no ALLOWED_SCOPES check, no scope ceiling, admin-only-mint boundary porous via parent rule)
## Summary
Pre-existing, surfaced during the #2989 (PR #2993) authz red-team. `auth_service::generate_api_token` never validates the requested scope set, which creates three related gaps. **Impact is contained today** — the independent repo-permission middleware (`auth.rs::check_repository_action`) is the true enforcing control and held on every cross-tenant attempt in testing — so this is defense-in-depth hardening, not an active privilege escalation.

## Gaps
1. **No ALLOWED_SCOPES validation at mint.** `generate_api_token` accepts arbitrary scope strings (no `validate_scopes_pure`). Unknown/malformed scopes are persisted.
2. **admin-only-mint boundary is porous via the parent rule.** `delete:artifacts` is in `ADMIN_ONLY_SCOPES`, but a non-admin can mint the bare parent `delete` (not admin-only), which satisfies `delete:artifacts` via the broad-covers-specific matcher. So the "destructive scopes are admin-only to mint" property does not actually hold. NOTE: the fix is subtle — simply making every bare parent admin-only would break the common bare-`write` CI token (it parents non-privileged `write:artifacts`). The correct model is likely: the admin-only-mint check should trigger when a requested scope grants access to an admin-only scope AND is not itself a normal delegable capability — needs a deliberate design pass.
3. **No scope ceiling.** A presenting token can mint a token with scopes it does not itself hold (e.g. a `read:artifacts` token mints `write:artifacts`). Consider enforcing minted ⊆ presenter for non-admin mint flows (verify this doesn't break legitimate admin/UI mint paths first).

## Why not folded into PR #2993
#2993's core scope-vocabulary fix is red-team-verified clean (no over-granting) and actually tightens delete authz. These mint-path gaps are orthogonal, pre-existing, contained by the repo-permission layer, and (esp. #2) require a design decision that shouldn't be rushed. Tracking here for a focused fix + red-team.

Related: #2989 / PR #2993.

brandonrc:
## Summary

Hardens the token mint path behind the repo-permission enforcing layer (#2996), building on the #2993 scope model:

1. **Vocabulary backstop at the mint choke-point** — `AuthService::generate_api_token` now rejects any requested scope outside the canonical `ALLOWED_SCOPES` list, so no endpoint can persist arbitrary or unknown scope strings even if it forgets its per-endpoint validation. Bare action parents (`read`/`write`/`delete`) are not vocabulary and become un-mintable — relevant because `scopes_grant_access` treats a held bare parent as covering every colon-form child of that action (#2989/#2993).
2. **Delegation ceiling on personal-token minting** — new `AuthExtension::enforce_mint_ceiling`, called from `POST /auth/tokens`, `POST /profile/access-tokens`, and `POST /users/{id}/tokens`: a non-admin presenting a scoped credential cannot mint a token carrying a scope the presenting credential does not hold. Interactive/console logins (`scopes: None`) are action-unrestricted and completely unaffected; the repo-token endpoint already enforces its own stricter repo-action ceiling and is unchanged.
3. **Mintability-classification invariants** — pure unit tests pin that (a) no non-admin-mintable scope grants any `ADMIN_ONLY_SCOPES` entry via `scopes_grant_access` (evaluated with the #2993 broad-covers-specific parent rule active), and (b) the bare parent of every admin-only colon-form scope is un-mintable by a non-admin. A future vocabulary or matcher change that would reopen the gap fails these tests and forces a reviewed decision.

**Regression fix included:** `POST /profile/access-tokens` defaulted omitted `scopes` to bare `read`, which is outside the vocabulary; the default is now the canonical `read:artifacts` (doc examples and tests updated). A service-account test fixture minting bare `read` was updated to `read:artifacts` for the same reason.

**Behavior change (intended hardening):** the primitive-level vocabulary check applies to all callers, including admins — a non-`ALLOWED_SCOPES` string (e.g. `hack:system`) now returns 400 on every mint endpoint. Newly minted tokens must use canonical colon-form scopes; existing persisted tokens are unaffected.

Closes #2996

Unit: `validate_scopes_pure` bare-parent rejection; the two mintability invariants in `token_service`; `enforce_mint_ceiling` matrix (interactive `None` unaffected, scoped token bound, wildcard/admin covered, effective-admin bypass, GHSA-vvc3 narrow-admin-token still bound).
Integration (DB): on `/auth/tokens` — read-scoped credential minting `write:artifacts` → 403 while the same user interactive → 200; bare/unknown scopes → 400 for non-admin and admin; admin-only scopes still 403; `write:artifacts` still 200. On `/profile/access-tokens` — omitted scopes → 200 persisting `read:artifacts`; ceiling enforced.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes (error responses on existing endpoints only)

### DIFF
--- a/backend/src/api/handlers/auth.rs
+++ b/backend/src/api/handlers/auth.rs
@@ -556,6 +556,11 @@ pub async fn create_api_token(
     crate::services::token_service::enforce_admin_only_scopes(&payload.scopes, auth.is_admin)
         .map_err(AppError::Authorization)?;
 
+    // Delegation ceiling (#2996): a scoped credential may not mint a token
+    // that exceeds its own scopes. Interactive sessions (`scopes: None`)
+    // are unaffected.
+    auth.enforce_mint_ceiling(&payload.scopes)?;
+
     let auth_service = AuthService::new(state.db.clone(), Arc::new(state.config.clone()));
 
     let (token, id) = auth_service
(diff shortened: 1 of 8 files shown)


brandonrc:
Authz red-team (live deploy of head `aa57200`) — **core CLEAN, one LOW uniformity gap to close before merge.**

**Closed (verified live):** `validate_scopes_pure` sits inside `generate_api_token` and all five mint handlers route through it — bare `delete`/`write`/`read` → 400; `*`/`admin`/`write:users`/`promote:artifacts`/`delete:artifacts` → 403; mixed-case/whitespace/empty/custom → 400. The ceiling holds on /auth, /profile, /users; the parent-rule×ceiling coupling is safe (admin-only children caught by `enforce_admin_only_scopes` first; bare parents un-mintable). No admin lockout; the scan/registry-offline JWT issuers are unaffected. `escalation_closed=true`.

**Finding (LOW) — ceiling not uniform.** `enforce_mint_ceiling` was wired into auth.rs/profile.rs/users.rs but NOT `repo_tokens.rs::create_repo_token`, which keeps the older #2603-G3 `require_delegatable_scopes` check (guards only write:artifacts/write:repositories via repo-permission, ignores read-family). Live: a non-admin mints a `write:repositories`-only token, then via `POST /repositories/{key}/tokens` mints `read:users`/`read:artifacts`/`read:repositories` tokens it never held (200, DB-verified) — while the same request is 403 on /auth and /profile. Bounded: write:artifacts via repo path stays 403, all ADMIN_ONLY stay 403, and the laundered `read:users` token still hits 'Admin access required' on the admin-gated user endpoints — so no enumeration/admin reach. It's a least-privilege/confinement escape on one handler, not a cross-principal escalation.

**Fix:** add `auth.enforce_mint_ceiling(&payload.scopes)` to `create_repo_token` (before `generate_api_token`), OR extend `require_delegatable_scopes` to also require the presenting credential holds each requested read scope — so all four mint handlers enforce one identical ceiling. Holding for the fix + a quick re-verify.

Issue #3003, opened 2026-07-29
Phase 2: extend inline scan-and-block on download to npm and Docker/OCI (parity with proxy-PyPI)


brandonrc:
## Summary
#2954 / #2970 shipped inline scan-and-block on download for **proxy + hosted PyPI** (Phase 1 MVP): a download of a package with an enforced CVE-authoritative scan verdict is blocked (403) or held (423) fail-closed, with a digest-keyed verdict cache and policy-driven fail-open/closed. #2954's own scope notes the next phase explicitly: *"PyPI first, then npm + Docker/OCI (same Trivy/Grype support)."* This issue tracks that Phase 2.

A customer evaluation raised active blocking for **Node.js (npm)** and **Docker** as a primary requirement — today those formats proxy/serve without the inline scan-gate that PyPI now has, so a vulnerable npm/OCI artifact can be pulled through without being blocked on the serve path.

## Scope
Replicate the PyPI serve-path scan-gate for:
1. **npm** — the proxy/registry serve handler (tarball fetch) must run the same `run_inline_proxy_scanners` gate + verdict cache + fail-open/closed policy as `serve_scanned_pypi_file`, and 403/423 on an enforced verdict.
2. **Docker/OCI** — the blob/manifest serve path must gate on the scan verdict for the referenced image content (Trivy/Grype already support OCI), blocking a vulnerable pull fail-closed.

Reuse the Phase-1 machinery: `Scanner::is_cve_authoritative` gate (#2970), `proxy_scan_results` verdict cache keyed by content digest, `ProxyScanAction` fail-open/closed, and the live-scanner-version freshness check (#2976) so a CVE-DB bump re-scans. Do NOT reintroduce the Phase-1 defects that were caught and fixed (non-authoritative Ok masking an authoritative error; probe-None serving stale-clean).

## Acceptance
- A known-vulnerable npm package and a known-vulnerable OCI image, pulled through a proxy repo with scan enabled + fail_closed, are BLOCKED (403) / held (423), not served 200.
- Clean artifacts serve normally; verdicts are cached per digest; a scanner-DB advance invalidates a stale clean verdict.
- Parity behavior + tests mirroring the PyPI DTF tiers.

Related: #2954, #2970, #2913, #2976, #2947.

PR #3004, opened 2026-07-29
feat(npm): inline scan-and-block on proxy tarball pulls + shared scan-gate core

brandonrc:
## Summary

Part of #3003. Extends the inline scan-and-block gate — proven on the proxy-PyPI
serve path in #2954/#2970/#2976 — to **npm proxy tarball pulls**, and factors the
shared machinery into two reusable seams so npm (and, in the follow-up PR-2, OCI)
are thin call-sites instead of re-implementations.

Before this change, `npm.rs serve_tarball` never consulted a proxy scan verdict:
a known-vulnerable npm tarball pulled through a scan-enabled, fail-closed remote
repo streamed straight through with a `200` on the first `npm install`. The gate
existed only for PyPI.

### Two shared seams (single-sourced defenses)

1. **Scan seam** — `run_inline_proxy_scanners_target(...)` is now THE single
   implementation of the inline-proxy scan contract, reached by file artifacts via
   `ScannerService::scan_content{,_expecting}` (context-free `ScanTarget`) and by
   repository-context callers via `scan_content_target`. The
   `cve_scanner_applicable && !cve_scanner_ran` fail-closed gate therefore exists in
   exactly ONE place: a non-authoritative scanner's `Ok(default)` cannot mask an
   authoritative Grype error → inconclusive → fail-closed, for every format.

2. **Serve-decision seam** — `proxy_scan_service::decide_serve(row, current_version,
   action, ttl_days, now) -> ServeDecision` lifts the freshness + fail-open/closed
   state machine out of `serve_scanned_pypi_file`. `serve_scanned_pypi_file` is
   **rewritten to call it** (proving it is behavior-preserving for PyPI), and npm
   calls it too — so npm inherits the #2976 probe-None-fail-closed fix and the
   #2970 authoritative gate without re-implementing them. The handler-side glue
   (digesting, block/lock responses, scan-and-record) moved into `proxy_helpers`.

### npm wiring

`serve_tarball`'s remote branch, when `scan_on_proxy` is enabled, routes to
`serve_scanned_npm_tarball`: buffered capped fetch via
`fetch_artifact_with_cache_path_capped` (instead of streaming, so the bytes are
available to scan), verdict keyed on `sha256(tarball bytes)` (never the packument
or anything the upstream index controls), the shared `decide_serve` gate, then
`403` (vulnerable) / `423` (inconclusive fail-closed) / `200 + X-AK-Scan` (clean or
fail-open pending). Scoped packages and dist-tags need no special-casing — the
concrete `@scope/pkg/-/file.tgz` GET is the single seam and the content digest keys
the verdict. Repos without `scan_on_proxy` keep today's untouched streaming path.

### "The engine ran" is not "the artifact was assessed"

A review of the first cut found the gate could still pass vulnerable artifacts,
because `is_vulnerable()` is `findings_count > 0`: an engine that RUNS but has
nothing to grade reports zero findings, which is indistinguishable from a
genuinely clean package. That is not hypothetical — syft/grype do not catalog a
bare npm `package/package.json`, nor an sdist's root `PKG-INFO`. Two live
consequences, both serving `200` under a fail-closed policy:

* **npm, crafted.** Real vulnerable `lodash-4.17.11` bytes served at the lodash
  coordinate graded "clean" whenever its identity could not be pinned: identity
  rewritten to something benign, an empty root `package-lock.json` packed in as a
  decoy, `package/package.json` removed, or `version` shipped as a JSON number.
* **PyPI, uncrafted.** An ordinary vulnerable sdist (`PyYAML-5.3.1.tar.gz`,
  CVE-2020-14343) graded clean while the same release's wheel was correctly
  blocked — a pre-existing gap in the merged PyPI path, which this PR's shared
  core now owns.

Closed with one principle in the shared core: **under fail-closed, a verdict is
only `clean` when the CVE-authoritative engine actually assessed the identity
being served.** Three parts:

1. **Catalog capture.** Grype's `json` report names only CVE-MATCHED artifacts,
   so the full catalog comes from its `cyclonedx-json` presenter, requested in
   the SAME invocation (no second scan). It lands on `ScanOutput::cataloged`.
   `None` means "this scanner reports no catalog" (every other scanner, and the
   OCI paths) and changes nothing; `Some(vec![])` means the engine cataloged
   NOTHING, which is now inconclusive rather than clean.
2. **Identity pin + cross-check.** `ExpectedComponent` is derived from the
   REQUEST coordinate (route package name + the registry's invariant
   `{basename}-{version}.tgz` / the PyPI filename version), never from bytes the
   upstream controls. It is materialized into the scan workspace as a
   `package-lock.json` / `<name>-<version>.dist-info/METADATA` so the engine has
   something to grade — written AUTHORITATIVELY, so a decoy lockfile inside the
   archive cannot suppress grading. The engine's catalog must then actually
   contain that identity.
3. **npm content agreement.** The tarball's own `package/package.json` must
   state the identity it is being served as. Absent, unparseable, non-string
   version, or disagreeing => `ProxyScanIdentity::Unestablished` => inconclusive.

### Archive namespace isolation

The archive used to unpack into the very directory the scan writes its control
files to, so a crafted tarball could ship an entry AT a control path — a
regular file named `.ak-scan-shrinkwrap/package`, or a directory where the
staged file goes — and make the control write fail. Staging was best-effort, so
the failure silently reduced the scan to "identity pin only"; because the
assessment gate only requires the PIN to be cataloged, a tarball shipping a
vulnerable `npm-shrinkwrap.json` came back a confident, cached `200`-clean.

Fixed structurally rather than by name-matching: when a scan writes control
files, the archive is unpacked into a dedicated `<workspace>/pkg/` and every
control file goes to the workspace ROOT, a sibling. Extraction already rejects
`..`/absolute entries, so the archive can only write beneath `pkg/` — the two
namespaces are disjoint and no entry can collide with any control path, present
or future. This protects the npm pin, the staged shrinkwraps, and the Python
dist-info pin alike.

Staging is now fail-closed too: any failure to stage a shipped shrinkwrap is a
hard error (→ inconclusive → 423), not warn-and-continue. The old reasoning
that "a failed copy leaves less signal, never wrong signal" was false — the
dropped signal was exactly the vulnerable shrinkwrap.

Unpinned scans (every hosted upload) write no control files, have no collision
surface, and keep the original flat layout unchanged.

### Pinning without touching what the archive shipped

The pin has to land at a lockfile path so the CVE engine catalogs it. Earlier
revisions merged it into a shipped `package-lock.json` and normalized that
file's `lockfileVersion`; both were the wrong shape. Every rewrite is a chance
to parse npm's lockfile formats slightly differently from the scanner, and each
mismatch is a silent mask — two bypasses came out of it, each serving a
vulnerable transitive `200`-clean under fail-closed:

* Bumping a v1 lockfile to v3 made the engine read only `packages` and ignore
  the `dependencies` tree where the vulnerable transitives actually were. A
  dummy `packages` map was enough to trigger the bump, flipping identical
  content from blocked to clean.
* Only `package-lock.json` was considered. The engine does not catalog
  `npm-shrinkwrap.json` at all, yet npm HONORS a shrinkwrap over a lockfile, so
  a vulnerable shrinkwrap was completely invisible.

So nothing shipped is read, parsed, merged, re-versioned or overwritten. Each
signal is presented as its own file, at a path nothing else claims:

* **Shipped `package-lock.json`** (root, `package/`, any depth) — untouched.
  The engine already grades v1 `dependencies` AND v2/v3 `packages` natively; it
  only failed to when we rewrote them.
* **Shipped `npm-shrinkwrap.json`** — copied VERBATIM (no parsing, so no format
  for us to get wrong) to a `package-lock.json` under `.ak-scan-shrinkwrap/`,
  one subdirectory per source location since a root and a `package/` shrinkwrap
  can legitimately differ. The original is left in place.
* **Top-level identity pin** — always `.ak-scan-pin/package-lock.json`, never a
  path an archive can occupy, so a decoy can neither shadow nor suppress it.

The engine globs `package-lock.json` at any depth and grades every one it finds
(verified: shipped lock + shrinkwrap copy + pin all cataloged in one scan), so
the union is what gets assessed — nothing normalized, nothing masked. The
Python path has no lockfile-merge concept and is unchanged.

Ordering is deliberate: the cached-verdict fast path still runs FIRST, so a
known-vulnerable digest is blocked from cache (403) without re-fetching, and the
identity work only happens when a scan is actually required.

Scope is deliberately narrow. The assessment gate fires only when a serve path
supplied a coordinate AND the engine reported a catalog; hosted upload scans,
legacy callers, and scanners with no catalog signal behave exactly as before, and
`prepare` (unpinned) fabricates nothing. Under `fail_open` an unassessable
artifact still serves, loudly marked `X-AK-Scan: pending` — the tightening is
fail-closed only. A clean package whose identity WAS cataloged stays `200 clean`;
holding a genuinely-empty artifact under fail-closed is the accepted trade.

**OCI (Docker/OCI manifest scan-block) is intentionally NOT in this PR** — it is a
separate unit-of-scan (whole image via `oci-dir:`, cold-proxy blob-availability
timing, recursion avoidance) tracked as PR-2, so this stays reviewable. Because
OCI remains, this PR is `Part of #3003`, not `Closes`.

PyPI behavior is proven preserved through the `decide_serve` refactor: the full
existing `serve_scanned_pypi_file` suite (fail-closed 423, fail-open pending,
cached-vulnerable 403, cached-clean 200, the four #2976 version-freshness cases)
stays green. New npm handler tests mirror them 1:1, plus one per crafted shape.

Unit coverage for the assessment gate: zero-cataloged => inconclusive; cataloged
identity mismatch (name AND version) => inconclusive; **cataloged match with no
findings stays CLEAN** (the control that keeps the gate from withholding
everything); no-catalog-signal and no-expected-coordinate keep prior behavior; a
scanner error still outranks the catalog check. Plus the cyclonedx catalog
parser, the ecosystem-aware identity comparison (PEP 503 vs npm), the pin bodies,
that an archive shipping entries at control paths (file-at-staging-dir,
directory-at-staged-file, decoy-at-pin-path) cannot suppress grading, that a
staging failure is a hard error, that a shipped lockfile is preserved
byte-for-byte at both the root and
`package/` (including the v1-`dependencies`-plus-dummy-`packages` bypass shape),
that shrinkwraps are staged verbatim from both locations with the originals
intact, that nothing is staged when no shrinkwrap ships, that a decoy lockfile
cannot suppress the pin, and the "unpinned scans fabricate nothing"
blast-radius control.

A local deployment-test tier (`npm-scan-block`) reproduces everything end-to-end
against live upstreams and real Grype. Against the pre-fix build all five
blindness cases serve `200`; against this build all are withheld, while both
clean controls pass on BOTH builds — so the cases discriminate the fix, not the
environment.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Reuses the existing `proxy_scan_results` store (digest-keyed, format-agnostic) —
no migration, no new columns, no new `sqlx::query!` macros. `X-AK-Scan` and `X-NPM-Tarball-SHA256` are response
headers on the npm tarball serve path, not new endpoints.

Relates to a customer evaluation requiring active download-block parity across
proxy formats.

### DIFF
--- a/backend/src/api/handlers/npm.rs
+++ b/backend/src/api/handlers/npm.rs
@@ -2762,6 +2762,7 @@ async fn serve_tarball(
                     repo.id,
                     repo_key,
                     upstream_url,
+                    package_name,
                     &fetch_path,
                     &response_filename,
                     ctx,
@@ -3044,6 +3045,54 @@ async fn correct_cached_tarball_content_type(db: &PgPool, repository_id: uuid::U
 // synthetic-artifact shape, and response builder.
 // ---------------------------------------------------------------------------
 
+/// The npm version a tarball filename encodes for `package_name`, per the
+/// registry's invariant filename shape `{basename}-{version}.tgz` (#3003).
+///
+/// `basename` is the unscoped half of the name (`@acme/widget` → `widget`), so
+/// this resolves scoped and unscoped packages identically. Returns `None` when
+/// the filename does not have that shape for this package — which the serve
+/// path treats as inconclusive rather than guessing, because the version is
+/// half of the identity the CVE engine must grade.
+fn npm_version_from_tarball_filename(package_name: &str, filename: &str) -> Option<String> {
+    let basename = package_name.rsplit('/').next().unwrap_or(package_name);
+    let stem = filename.strip_suffix(".tgz")?;
+    let version = stem.strip_prefix(&format!("{basename}-"))?;
+    (!version.is_empty()).then(|| version.to_string())
+}
+
+/// The `name`/`version` an npm tarball claims for ITSELF, read from the
+/// `package/package.json` that `npm pack` always writes (#3003).
+///
+/// Returns `None` when the entry is absent, unparseable, or carries a
+/// non-string / empty name or version — including the JSON-number version
+/// shape. Every one of those is "this tarball does not state a usable
+/// identity", which the caller must treat as inconclusive, never clean.
+fn npm_claimed_identity(tarball: &Bytes) -> Option<(String, String)> {
+    let body = crate::util::bounded_archive::read_metadata_from_tar_gz(&tarball[..], |path| {
+        path == std::path::Path::new("package/package.json")
+    })
+    .ok()??;
+    let v: serde_json::Value = serde_json::from_slice(&body).ok()?;
+    let name = v.get("name")?.as_str()?.trim().to_string();
+    let version = v.get("version")?.as_str()?.trim().to_string();
+    (!name.is_empty() && !version.is_empty()).then_some((name, version))
+}
+
+/// Whether the identity a tarball claims for itself agrees with the coordinate
+/// it is being served at (#3003).
+///
+/// npm names are lowercase by construction, but compare case-insensitively so a
+/// legacy mixed-case publication is not spuriously withheld. Version equality
+/// is exact: a tarball claiming a different version than the URL pins is not
+/// the artifact the consumer asked for.
+fn npm_identity_agrees(
+    requested_name: &str,
+    requested_version: &str,
+    claimed: &(String, String),
+) -> bool {
+    claimed.0.eq_ignore_ascii_case(requested_name) && claimed.1 == requested_version
+}
+
 /// Build the synthetic in-memory [`Artifact`](crate::models::artifact::Artifact)
 /// the leaf scanners run over for a proxied npm tarball. There is NO
 /// `artifacts` row (proxy-cached bytes are deliberately not persisted as
@@ -3120,6 +3169,7 @@ async fn serve_scanned_npm_tarball(
     repo_id: uuid::Uuid,
     repo_key: &str,
     upstream_url: &str,
+    package_name: &str,
     fetch_path: &str,
     filename: &str,
     ctx: &crate::api::middleware::download_telemetry::DownloadContext,
@@ -3199,9 +3249,53 @@ async fn serve_scanned_npm_tarball(
     correct_cached_tarball_content_type(&state.db, repo_id, fetch_path).await;
 
     let digest = proxy_helpers::sha256_hex(&bytes);
+
+    // #3003: establish WHAT these bytes are being served as.
+    //
+    // The coordinate comes from the REQUEST (route package name + the
+    // registry's invariant `{basename}-{version}.tgz` filename), never from
+    // bytes the upstream controls, and the tarball's own `package/package.json`
+    // must agree with it. A tarball that states a different identity, states
+    // none, or states an unusable one (missing/empty/non-string version) is
+    // unassessable: any scan of it would grade something other than the package
+    // the consumer is about to install under this coordinate.
+    //
+    // This is only consulted when the shared gate actually needs to SCAN — a
+    // cached vulnerable verdict for this digest still blocks first, from cache.
+    let identity = match npm_version_from_tarball_filename(package_name, filename) {
+        Some(version) => match npm_claimed_identity(&bytes) {
+            Some(claimed) if npm_identity_agrees(package_name, &version, &claimed) => {
+                proxy_helpers::ProxyScanIdentity::Established(
+                    crate::services::scanner_service::ExpectedComponent::new(
+                        crate::services::scanner_service::ComponentEcosystem::Npm,
+                        package_name,
+                        &version,
+                    ),
+                )
+            }
+            other => {
+                tracing::warn!(
+                    repo_id = %repo_id, file = %filename, digest = %digest,
+                    requested = %format!("{package_name}@{version}"),
+                    claimed = ?other,
+                    "npm proxy tarball does not state the identity it is served as"
+                );
+                proxy_helpers::ProxyScanIdentity::Unestablished
+            }
+        },
+        None => {
+            tracing::warn!(
+                repo_id = %repo_id, file = %filename, digest = %digest,
+                package = %package_name,
+                "npm proxy tarball filename does not encode a version for this package"
+            );
+            proxy_helpers::ProxyScanIdentity::Unestablished
+        }
+    };
+
     let synthetic = npm_synthetic_artifact(repo_id, filename, &digest, bytes.len() as i64);
     match proxy_helpers::gate_proxy_scan_serve(
-        state, repo_id, filename, &digest, synthetic, &bytes, action,
+        state, repo_id, filename, &digest, synthetic, &bytes, action, identity,
     )
     .await
     {
@@ -9175,7 +9269,7 @@ mod proxy_scan_block_tests {
         upstream: &wiremock::MockServer,
         package: &str,
         filename: &str,
-        tarball: &'static [u8],
+        tarball: &[u8],
         expected_fetches: Option<u64>,
     ) {
         use wiremock::matchers::{method, path};
@@ -9184,7 +9278,7 @@ mod proxy_scan_block_tests {
             .and(path(format!("/{package}/-/{filename}")))
             .respond_with(
                 ResponseTemplate::new(200)
-                    .set_body_bytes(tarball)
+                    .set_body_bytes(tarball.to_vec())
                     .insert_header("Content-Type", "application/octet-stream"),
             );
         if let Some(n) = expected_fetches {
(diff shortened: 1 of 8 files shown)


brandonrc:
Red-team (live deploy of head `8b1286b` on an isolated slot) — **core fail-closes correctly on honest packages + scanner errors, but two HIGH grype-blindness gaps hold the merge.**

**Held clean (good):** npm blocks honest vulnerable packages (lodash 4.17.11 → 403, minimist 1.2.5 → 403, scoped @babel/traverse 7.23.0 → 403 — scoped is not a bypass); fail-closed inheritance intact (#2954 broken-grype → 423, #2976 probe-None+cached-clean → 423 not stale-200, cached-vulnerable → 403); PyPI unregressed by the shared-core refactor; no digest TOCTOU; 220MB body → 423 at the 200MiB cap, no OOM.

**Finding 1 (HIGH) — grype-blindness served CLEAN under fail_closed (npm).** A tarball Grype cannot catalog is graded 0-findings = clean. Live: real vulnerable lodash under 4 shapes that defeat `derive_npm_package_lock` — (a) package.json name/version rewritten benign, (b) an empty root package-lock.json added (derive early-returns if a lock exists), (c) `package/package.json` removed, (d) version as a JSON number — each served 200 `X-AK-Scan: clean`, verdict findings=0, grype RAN (0 gradeable components, so the #2954 scanner-error gate doesn't fire). Root cause: `is_vulnerable()` = `findings_count > 0`, so '0 components' == CLEAN.

**Finding 2 (HIGH) — pypi sdist grype-blindness (pre-existing, in merged #2970).** A normal unmodified vulnerable sdist serves 200-clean: `GET .../PyYAML-5.3.1.tar.gz` (CVE-2020-14343) → 200 clean; `grype dir:` on the extracted tree finds nothing (no .dist-info/METADATA). The wheel → 403. No crafting needed.

**Fix (one principle, both formats — the shared core now serves PyPI too):** under fail_closed, CLEAN may be returned only when the CVE-authoritative scanner actually **assessed the served artifact's real identity+contents**. So:
1. Track the scanner's **cataloged component count**. If the authoritative scanner ran but cataloged **0 components** for content that should be gradeable (an npm tarball, a pypi sdist/wheel) → **INCONCLUSIVE → 423** under fail_closed (not clean). Genuinely-empty artifacts being conservatively held under fail_closed is acceptable.
2. **Identity cross-check** (closes shape a): the graded component identity must match the requested package name@version; if Grype graded a *different* identity than what's being served → inconclusive.
3. npm: if `derive_npm_package_lock` couldn't pin name@version → inconclusive. pypi: add a PKG-INFO/name@version sdist derive analogous to the npm one (so sdists get graded), backstopped by the 0-component gate.

fail_open behavior unchanged (serves + pending). Holding for the fix + a re-red-team of all four npm shapes and the pypi sdist.

brandonrc:
Re-red-team (partial, code-confirmed) — the 2 grype-blindness HIGHs are closed and the fix is a big net win, but there's a **bounded MEDIUM to fix before merge: the pin clobbers a shipped lockfile's resolved transitives.**

**Confirmed by source at 857bfbd:** `ScanWorkspace::prepare_pinned` extracts the tarball, then `write_component_pin` UNCONDITIONALLY writes `package-lock.json` = `npm_package_lock_pin_json(name, version)`, a single-entry lock naming only the pinned top-level (`node_modules/{name}: {version}`), overwriting any `package-lock.json` the archive shipped (the doc comment says so). So a tarball that ships its OWN `package-lock.json` listing a vulnerable RESOLVED transitive → that lockfile is overwritten → Grype only sees the clean top-level → the transitive is masked. Pre-#3004, Grype would have resolved the shipped lockfile and caught it. The 'deps are semver ranges, don't fabricate findings' rationale is correct for `package.json` but does not apply to a shipped lockfile (resolved versions, real signal).

**Severity bounded:** published npm registry tarballs strip `package-lock.json`, and a consumer installing a dep doesn't use that dep's lockfile — so the common case is unaffected and #3004 remains a large improvement over the two HIGHs it closes. But it's a real scan-fidelity regression a security-focused evaluator could probe.

**Fix:** don't clobber — MERGE. After extraction, if a `package-lock.json`/`npm-shrinkwrap.json` with resolved `packages` exists, keep its resolved entries (so transitives are graded) and inject/override ONLY the top-level `node_modules/{name}` = the pinned (request-coordinate) version — this both preserves real transitives AND still defeats an empty/decoy/benign-top-level lockfile. Fall back to the single-entry pin lock if the shipped lock is absent/malformed. Then re-verify: a tarball shipping a vulnerable-transitive lockfile → 403; the original 4 shapes stay 423/403; clean controls stay 200. Holding for this + a re-red-team.

brandonrc:
Final re-red-team (live 48a7e9c, fail_closed) — **two HIGH bypasses in the merge logic; do not merge.** The natural nested-lockfile case, unmergeable-preserve path, and no-over-restriction all HOLD, but a vulnerable npm transitive (lodash 4.17.11, GHSA-jf85 Critical) still serves 200 `X-AK-Scan: clean` via two paths:

**HIGH 1 — merge `lockfileVersion 1→3` normalization launders a `dependencies` vuln.** A lock shipping a dummy `packages` map (to take the mergeable branch) + a `dependencies` tree with vulnerable lodash: `merge_npm_lock_pin` (scanner_service.rs:1233-1239) bumps lv 1→3 while RETAINING `dependencies`. grype at lv≥2 parses only `packages` → the `dependencies` vuln is invisible → 200. Adding the dummy `packages` map flips the identical content 403→200 (live).

**HIGH 2 — `npm-shrinkwrap.json` transitives never graded.** grype does not catalog npm-shrinkwrap.json at all, and npm HONORS it over package-lock.json. `write_npm_lock_pin` (scanner_service.rs:983) only inspects package-lock.json → a tarball shipping a vulnerable shrinkwrap + no lock → single-entry pin only → 200. (live)

Both persist a shared digest-keyed `clean` row → cross-tenant cache poison.

**Fix — STOP normalizing; preserve + pin separately (avoids all format-parse pitfalls):**
1. NEVER rewrite/re-version a shipped lockfile. Leave any shipped `package-lock.json` byte-for-byte — grype grades it NATIVELY (v1 `dependencies` AND v2/v3 `packages` are both read correctly at their own version; the bug was ONLY introduced by bumping the version).
2. `npm-shrinkwrap.json`: grype won't read it, so COPY it verbatim to a `package-lock.json` grype will catalog, at a distinct path (e.g. `.ak-scan-shrinkwrap/package-lock.json`) — same format, now graded. Grade BOTH if a tarball ships lock + shrinkwrap.
3. ALWAYS write the top-level identity pin to a distinct nested path (`.ak-scan-pin/package-lock.json`, as you already do for the unmergeable case) so the served identity is graded authoritatively regardless.
Net: grype grades shipped-lock transitives + shrinkwrap transitives + top-level pin; nothing normalized, nothing masked, decoy top-level still overridden by the separate authoritative pin.
4. Live re-verify the PyPI path too (it was refactored to the shared gate but not re-tested): sdist→403, vuln wheel→403, clean wheel→200.

Holding for this + another re-red-team.

brandonrc:
Final re-red-team (live 4a3256b, fail_closed) — prior HIGH-1/HIGH-2 direct repros stay closed, over-restriction clean, PyPI ok, BUT a new **HIGH archive-controlled path collision** reopens the shrinkwrap-blindness class. Do not merge.

**HIGH — prevent-the-copy via a colliding archive entry.** A tarball ships the vulnerable `package/npm-shrinkwrap.json` PLUS an extra entry named `.ak-scan-shrinkwrap/package` (a regular file; or ships `.ak-scan-shrinkwrap/package/package-lock.json` as a directory). Extraction writes it first; `stage_shrinkwraps_for_grading` (scanner_service.rs ~L1015-1053) then `create_dir_all`/`tokio::fs::copy` FAILS (`File exists`/`Is a directory`) and warn+continues → the vulnerable shrinkwrap is never copied to a gradeable path → grype (which doesn't catalog npm-shrinkwrap.json) sees only the benign pin → findings=0 → **200 X-AK-Scan: clean**, cached digest-keyed. Live-confirmed both variants; a consumer installing the package gets the vulnerable minimist while AK served clean under fail_closed.

**Root cause (asymmetry, proven live):** the identity-PIN write IS a hard error (shipping `.ak-scan-pin` as a file → scan errors → inconclusive → 423, correct), but the shrinkwrap staging is best-effort warn+continue, and the assessment gate only requires the PIN to be cataloged, never that a shipped shrinkwrap was graded — so 'grype grades the union' silently degrades to pin-only. The comment 'a failed copy leaves strictly less signal, never wrong signal' is the flawed premise: for a vulnerable shrinkwrap, less signal = wrong (clean) verdict.

**Fix — structural namespace isolation (+ fail-closed):**
1. **Isolate extraction from the control namespace.** Extract the archive into a DEDICATED subdir (e.g. `<workspace>/pkg/`) and write ALL `.ak-scan-*` control files (identity pin, shrinkwrap-copies) at the workspace ROOT (a SIBLING of the extraction dir). Then NO archive entry can pre-create or collide with a control path. grype scans the whole workspace → catalogs `pkg/` (extracted shipped locks, at native version) + the root control files. This future-proofs the pin against the same trick too.
2. **Fail-closed on any control-write failure.** Make a shrinkwrap staging failure a HARD error exactly like the pin — a shipped `npm-shrinkwrap.json` (or `package-lock.json`) that cannot be staged/graded → inconclusive → 423, NEVER clean. Never warn+continue past a control-write failure.
3. Optionally have the assessment gate assert that every shipped lockfile/shrinkwrap it detected was actually represented in the catalog (defense in depth), but (1)+(2) close it.

Re-verify: both collision variants (`.ak-scan-shrinkwrap/package` as file, and target as dir) → 403/423; plus a colliding `.ak-scan-pin` decoy → 423; all prior cases (HIGH-1/2, nested lock, decoy, controls, PyPI) unchanged. Add the collision shapes to the DTF tier. Holding for this + a re-red-team.

Issue #2930, opened 2026-07-26
curation_enabled is settable per repository but silently inert on 26 of 27 proxy formats


brandonrc:
## Problem

`curation_enabled` is settable per repository, on any repository of any format, and is silently inert on 26 of the 27 proxy-capable formats. Only PyPI acts on it. Nothing in the API, the response payload, or the UI tells an operator which of those two situations they are in.

The framing matters, so to be precise about what this issue is and is not: it is **not** "26 formats were forgotten". #2912 scoped enforcement to PyPI deliberately as a first increment, and #2264 tracks the cross-format consolidation. The finding is that a security control is accepted, stored, and echoed back as repository state while enforcing on one format, with no signal to the operator. An operator who enables curation on a NuGet or Cargo repository and adds a typosquat block rule has a false sense of enforcement, and the system agrees with them at every point they could check.

That gap exists regardless of whether the consolidation is already planned, which is why this is filed separately rather than folded into #2264 alone.

Pre-existing. Not introduced by #2918, #2919, or #2922.

## Scope of the gap

Curation is designed to cover remote and virtual repositories only. PyPI's own gate returns early on hosted repos:

```
if !repo.curation_enabled || !matches!(repo.repo_type.as_str(), "remote" | "virtual")
```

`backend/src/api/handlers/pypi.rs:259`

So "hosted enforced, proxy not" is not the shape of this. Proxy serve is the only surface curation was ever meant to cover, and on the other 26 formats it covers none of it. For those formats the gap is total: no enforcement on any arm.

## Evidence

The fields reach every format. `resolve_repo_by_key` selects both columns and populates them on every `RepoInfo`:

- `backend/src/api/handlers/proxy_helpers.rs:161-162` (struct fields)
- `backend/src/api/handlers/proxy_helpers.rs:204` (SELECT)
- `backend/src/api/handlers/proxy_helpers.rs:240-244` (populated from the row)
- `backend/src/api/handlers/proxy_helpers.rs:2528-2529` (carried across the virtual-member conversion)

Only PyPI reads them:

- `backend/src/api/handlers/pypi.rs:253-283` (`enforce_pypi_curation`)
- Call sites: `pypi.rs:701` (simple index), `:1006` (virtual member index), `:1455` (download), `:1896` (virtual member download)

Both shared primitives have exactly one caller each:

- `curation_blocked_response` (`proxy_helpers.rs:4779`) from `pypi.rs:278`
[...]

Issue #2912, opened 2026-07-25
Security gates do not enforce: scan policies, admin quarantine transition, curation on proxies


brandonrc:
## Summary

Three security features expose full API and schema surface but do not actually enforce at runtime. Each is a half-wired feature: the configuration is accepted and stored, but nothing acts on it, so an operator who sets a policy reasonably believes they are protected when they are not.

### 1. Scan policies are never enforced

`scan_policies` (name, max_severity, block_unscanned, block_on_fail) are full CRUD via `POST /api/v1/security/policies`, and `PolicyService::evaluate_artifact` implements the evaluation completely, but nothing ever calls it. A completed scan that finds a critical CVE sets `quarantine_status = 'flagged'` (non-blocking); downloads are not blocked. The policy is decorative.

### 2. No flagged-to-quarantined transition for admins

The quarantine state machine has `clean`, `flagged`, and `quarantined` states with admin release/reject endpoints, but no path to move a flagged artifact to quarantined. `POST /api/v1/quarantine/{id}/reject` returns 409 unless the artifact is already quarantined, so an admin who spots a bad flagged artifact has no lever short of deleting it.

### 3. Curation block rules do not gate proxy downloads

Repositories carry `curation_enabled` / `curation_default_action` columns and `POST /api/v1/curation/rules` accepts block rules, but only the RPM/Debian curation-sync paths consult them. `CurationService` has zero references in `pypi.rs`, so a block rule has no effect on PyPI proxy traffic. A typosquat block rule silently does nothing.

## Proposed fix

Complete all three so the existing schema and API tell the truth:

1. Call `PolicyService::evaluate_artifact` at scan completion; a violation quarantines the artifact (download-blocked) with the policy reason. Fail open on evaluation error so a broken policy engine never takes scanning down.
2. Add `POST /api/v1/quarantine/{artifact_id}` (admin-only) to quarantine a flagged/clean artifact manually.
3. Enforce curation block rules on the PyPI proxy index and download paths (403 with the rule reason, before upstream lookup), gated on `curation_enabled`.

Plus a `quarantine_reason` column surfaced in blocked-download errors, and guards so a later scan (clean result, or a scanner failure) cannot silently downgrade an admin- or policy-set quarantine.

## Related

- Relates to #2264 (unify proxy-layer enforcement across formats): this is a concrete increment toward a single enforcement point.
- Follows the #1770 age-gate-on-proxy precedent, which the curation hook mirrors.
[...]

PR #3022, opened 2026-07-30
fix(oci): inline scan-and-block on Docker/OCI proxy manifest pulls

### LINKED ISSUE #3003: Phase 2: extend inline scan-and-block on download to npm and Docker/OCI (parity with proxy-PyPI)

brandonrc:
## Summary

Completes #3003 (Phase 2) by extending inline scan-and-block on download to
**Docker/OCI proxy pulls**, the last format in the parity set. PR-1 (#3004)
landed the shared scan-gate core and the npm half; this PR wires the OCI proxy
serve path onto that same core. Proxy-PyPI (#2954/#2970/#2976), proxy-npm
(#3004), and now proxy-OCI all reach the identical digest-keyed verdict state
machine, so the fail-closed contract lives in one place and cannot drift per
format.

Requested during a customer evaluation: a scan-enabled, `fail_closed` Docker
mirror repository must refuse to serve a vulnerable image, not just record a
finding after the fact.

### The gap

`oci_v2.rs handle_get_manifest` enforced curation (#2930) and quarantine
(#2912) but never consulted a proxy scan verdict, so a known-vulnerable image
pulled through a scan-enabled, `fail_closed` remote repo resolved and pulled
normally on the very first `docker pull`.

### Block point

The **manifest GET** — the single seam every `docker pull` resolves first
(blobs are content-addressed by digest and carry no image identity), and where
curation and quarantine already sit. The unit of scan is the **whole image,
once**, keyed on the manifest's content digest in `proxy_scan_results`; never
per-blob (layers are not independently gradable, and per-blob would be N×
expensive). Blob GETs are untouched.

Both manifest serve branches are gated — the cold upstream fetch and the warm
local-cache serve — so a repo whose first pull happened under `fail_open`
still blocks on every later pull once the verdict exists.

Gated only when the served body is a **runnable single-image manifest**:

* a multi-arch **index** passes through unscanned (it carries no config or
  layers); the client's follow-up platform-child manifest GET re-enters this
  same gate and is scanned at its own digest — which is the precise image it
  would actually run;
* non-image OCI payloads (Helm charts, cosign signatures, SBOM/attestation,
  WASM) pass — they are not container images;
* **hosted** repos are not gated here at all. Hosted push→pull is already
  blocked by scan-on-upload → quarantine, and double-gating it would change
  hosted serve behavior.

### Cold-proxy blob availability, without registry recursion

`docker pull` fetches the manifest **before** the layer blobs, so at gate time
the image's blobs are not yet local and Grype's local `oci-dir:` reassembly
has nothing to read. The pre-existing fallback for that case is `registry:`
mode pointed at our own registry endpoint — which would re-enter
`handle_get_manifest` and recurse.

Instead the gate **eager-fetches** the config + layer blobs from upstream into
local storage before scanning, then scans the local `oci-dir:` layout. Every
staged blob is content-address-verified against the digest the manifest
declares before it is written, the total is bounded by the same
`PROXY_SCAN_MAX_BYTES` cap the other formats use (with a cheap declared-size
pre-check before any fetch), and staging is idempotent. The eager fetch warms
the cache, so the client's following blob GETs are served locally — the extra
latency is front-loaded onto the manifest GET, which is the correct
fail-closed posture. A scanner-scoped pull token (`scan_pull_repo`, #2093) is
exempt from the gate so the residual `registry:` fallback can never deadlock
against it.

### Carried-over fail-closed contract

* **#2954** — a non-CVE-authoritative scanner's `Ok` cannot mask an
  authoritative failure. Inherited unchanged by calling through the single
  `run_inline_proxy_scanners_target` loop.
* **#2976** — a cached-clean verdict plus an unknown live scanner version
  under `fail_closed` re-scans rather than serving stale-clean. Inherited via
  `decide_serve`.
* **#3003 assessment gate, OCI form** — an image has no `name@version`
  coordinate to pin, so instead the OCI caller sets `require_nonempty_catalog`
  on the `ScanTarget`: the CVE engine must have actually **cataloged** at
  least one gradeable component before a *zero-finding* result is trusted as
  clean. "Ran but graded nothing" and "reported no catalog at all" are both
  INCONCLUSIVE → 423 under `fail_closed`. A verdict *with* findings is
  vulnerable regardless of the catalog signal, so blocking is never weakened
  by it. To make that signal exist for images, the `oci-dir:` scan now
  captures Grype's CycloneDX catalog from the same single subprocess the dir
  mode already did — no second scan, no extra extraction.
* Any staging/control-plane failure (upstream fetch, digest mismatch,
  unverifiable digest algorithm, storage or DB write) is a hard error → 423,
  never warn-and-serve-clean.
* `fail_open` is unchanged in posture: serve immediately with a loud
  `X-AK-Scan: pending` header and scan asynchronously, so the next pull of
  that digest blocks.

### Round 2: three bypasses found by review, and what changed

A red-team pass against the first cut found three ways a vulnerable image
still reached a client. All three are closed here; each has a unit test, a
handler test, and a deployment-test discriminator that fails on the previous
head and passes now.

**The shared root cause of the first two: the gate predicate described the
SHAPE of the document, and the shape is attacker-controlled.** The classifier
disagreed with what `docker pull` actually runs, so dressing a real vulnerable
image up in the right decoration skipped the gate entirely and the image was
served `200` unscanned (falling through to the pre-#1274 cached-unscanned
path).

* **HIGH-1 — a cosmetic `manifests` array.** `classify_manifest` tested
  `manifests` *first*, so a runnable image manifest carrying `"manifests": []`
  classified as an Index and was waved through. (The same ordering was also a
  latent GC bug: that body recorded `oci_manifest_refs` for its zero children
  and no `manifest_blob_refs`, i.e. exactly the ref-less live tag that pins the
  blob-GC readiness gate, #1409 C1.)
* **HIGH-2 — a decoy non-image layer.** `oci_manifest_is_scannable_image`
  returned false if *any* layer carried a non-image marker, so bolting one
  dummy `vnd.dev.cosign…` / `…+wasm` / `spdx` descriptor next to the real
  `tar+gzip` layer made a genuine image look like a signature.

The fix replaces the shape test with a **consumer-faithful** one,
`oci_manifest_is_runnable_image`: an image config **and** at least one real
rootfs layer (a `tar*` layer that is not a known non-image payload). Everything
else in the body is ignored on purpose. A co-present `manifests` array does not
make it an index; a decoy descriptor is not an exemption, it is simply not a
catalog source. `classify_manifest` now lets "is a runnable image" win for the
same reason, which also closes the latent GC hole. A **pure** index (a
`manifests` array with no image config of its own) still passes ungated so the
platform child re-enters the gate at its own digest, and genuine cosign
signatures, SBOMs, Helm charts and WASM modules still pass — they have no
rootfs layer.

* **HIGH-3 — the blob path was ungated.** Blocking the manifest was not
  enough: the client took the 403 and then fetched every config and layer blob
  `200` by digest, reconstructing the image. `handle_get_blob` now refuses a
  blob that belongs to an image this repository has already scanned and found
  vulnerable.

  The blocklist is **derived, not stored** — an `EXISTS` join over the
  `manifest_blob_refs` edges the gate already records and the digest-keyed
  `proxy_scan_results` verdict. No new table, no migration, and no drift: a
  re-scan that flips a verdict back to clean un-blocks the blobs immediately.
  Both existing indexes (`idx_manifest_blob_refs_blob`,
  `idx_proxy_scan_checksum`) already cover the lookup. A control-plane error on
  that lookup fails closed (423).

  Two scoping choices, both deliberate and both tested: the reference set is
  per repository, so a layer another repo shares via a *different, clean* image
  keeps serving there; and within a repository it is ANY-not-ALL, so a layer
  shared between a vulnerable and a clean image is blocked. The second
  over-blocks the clean sibling, which is the fail-closed direction — the
  alternative would let a crafted clean sibling manifest unlock a
  known-vulnerable layer. Note the verdict store itself is content-addressed
  and therefore global by design (#2954): the *same* manifest digest proxied
  into two repositories is the same image and is correctly blocked in both.

**Also fixed while verifying the above (found live, not reported):** the
`oci-dir:` layout descriptor took its `mediaType` from the **upstream
Content-Type header**. Grype refuses to parse a layout whose descriptor is not
a manifest type, so an upstream answering a manifest GET with
`application/octet-stream` made every image scan inconclusive — every pull
423 under `fail_closed`, and, worse, a permanent "serve unscanned with
`X-AK-Scan: pending`" under `fail_open`. The mediaType is now derived from the
manifest body (which the digest covers), falling back to the header only when
it already looks like a manifest type, then to the canonical OCI type.

### Scope of the manifest gate (honest limits)

Same threat model as the curation (#2930) and quarantine (#2912) gates that
already live at this seam, stated explicitly so review can assess it:

* `GET` manifest and `GET` blob are both gated. `HEAD` manifest is not — it
  returns no image bytes, only existence plus the digest, and a client cannot
  obtain the layers without the manifest body or the blobs. Gating `HEAD`
  would also charge a full image scan for a cheap existence probe.
* A blob can only be known-vulnerable once the image that names it has been
  scanned. A blob for an image nobody has pulled through the gate yet is
  served — the same seam limit curation and quarantine have.
* **Virtual repositories are out of scope in this PR.** The gate applies to
  `Remote` (proxy) repositories; a `Virtual` repo resolving a manifest from a
  remote member returns before the gate, exactly as npm's virtual-member path
  was deferred in PR-1 (#3004). This is a known gap, not an oversight, and is
  the obvious next slice.
* Verdicts are keyed strictly on the content digest we compute, never on
  anything the upstream advertises, so a lying upstream cannot bind a clean
  verdict to different bytes.

### Behavior worth calling out for review

Under `fail_closed`, an image the CVE engine can catalog **nothing** for — a
`scratch`/`hello-world`-style image with no package database — is withheld
with **423**, not served as clean. That is deliberate and is the same posture
PR-1 took for an npm tarball whose identity could not be established: a
zero-finding scan of something the engine could not grade is not a clean
verdict. Operators who need such images through a `fail_closed` proxy should
use `fail_open` for that repository. Called out explicitly because it is a
real behavior change for minimal images, and it is pinned by a test in both
directions.

### No migration

Reuses `proxy_scan_results` (digest-keyed, already format-agnostic) plus the
existing `oci_blobs` / `manifest_blob_refs` tables.

Unit + handler tests (`oci_v2::proxy_scan_block_tests`, 22 tests) drive the
real router end-to-end over a DB + wiremock upstream: cached-vulnerable digest
→ 403 on both the cold and warm serve branches with no upstream re-fetch;
inline scan of a vulnerable image → 403 with blobs staged and the verdict
persisted; cached-clean → 200 with `Docker-Content-Digest`; index → passes
ungated; `fail_closed` + inconclusive → 423; `fail_closed` + cached-clean +
unknown live scanner version → 423 (#2976); `fail_open` → 200 +
`X-AK-Scan: pending`; scan-scoped pull token exempt while the same pull
without the scope is withheld; staging hard-failures (over-cap, digest
mismatch) error rather than degrade; staging is idempotent on re-scan.

`scanner_service::tests` adds the OCI assessment-gate truth table (empty
catalog → inconclusive, absent catalog → inconclusive, cataloged clean → clean,
findings block even without a catalog, flag off → prior behavior unchanged).

Round 2 adds: the gate truth table for both bypass shapes (image +
`manifests` array, image + each decoy layer type, and both at once) plus the
non-runnable controls (pure index, real cosign, real Helm, SBOM, image with no
layers); an end-to-end handler test driving five crafted shapes through the
real router and asserting none is served; a real-signature control; the blob
gate (vulnerable image's config and layer blobs refused, clean image's served);
the blocklist scoping test; and the layout-mediaType derivation tests.

A deployment-test tier (`oci-scan-block`) exercises the same flow against a
live stack proxying Docker Hub, including both crafted bypass shapes served
from a local origin over a real vulnerable payload. 9 of its assertions fail
on the pre-PR `main` and 4 more fail on the round-1 head; all 28 pass on this
branch.

## Test Checklist
- [x] Manually tested locally

Manual verification on an isolated instance proxying Docker Hub with
`scan_on_proxy` + `fail_closed`:

| probe | result |
| --- | --- |
| `docker pull <ak>/library/alpine:3.9` | fails: `denied: … blocked by scan policy: vulnerabilities found` |
| index GET, then the amd64 child GET | 200, then **403** (132 findings, 7 critical / 69 high, from a local `oci-dir:` reassembly of 2 eagerly-staged blobs) |
| repeat pull (local-cache serve branch) | **403** from the cached verdict |
| **HIGH-1** real vulnerable image + `"manifests": []` | **403** — scanned and graded vulnerable (was **200 unscanned**) |
| **HIGH-2** real vulnerable image + decoy cosign layer | **423** — withheld; Grype cannot parse a layout carrying a foreign layer type, so fail-closed refuses rather than serving (was **200 unscanned**) |
| **HIGH-3** config blob of the blocked image | **403** (was 200) |
| **HIGH-3** layer blob of the blocked image | **403** (was 200) |
| clean distroless image, and its config blob | **200** / **200** — no over-block |
| blob in a repo with no scan history | **200** — no over-block |
| `docker pull` of the clean image, end to end | succeeds |
| pure index | 200, child gated at its own digest |
| real cosign signature artifact (no rootfs layer) | 200, ungated |
| zero-catalog image (`hello-world`) | 423 |
| `fail_open` first pull | 200 + `X-AK-Scan: pending`, then 403 after the async scan |
| scanner-scoped pull token | 200 on its own repo (exempt), 403 on another (scope gate holds) |
| hosted push→pull round trip | unaffected, not double-gated |
| `registry:`-mode scans in the logs | zero — the recursion path is never taken |

The two crafted shapes are served from a local origin over a **real** vulnerable
image payload (sourced through a separate un-gated proxy repo), so a 200 there
would be a genuine compromise rather than a synthetic shape.

## API Changes
- [x] N/A - no API changes

The manifest GET can now additionally return `403 DENIED` (vulnerable) or
`423 DENIED` (inconclusive under fail-closed) as OCI error bodies, and a
`fail_open` serve carries `X-AK-Scan: pending` — only for repositories that
have opted into `scan_on_proxy`. No route, schema, or migration changes.

Closes #3003


brandonrc:
Red-team (live d18a916, fail_closed) — **3 HIGH bypasses; do not merge.** The core manifest scan+block works (plain runnable image → 423/403; index children gated; no scan-vs-serve TOCTOU; scan-scoped token not forgeable; byte-cap/#2976 correct), but a vulnerable image reaches a client THREE ways:

**HIGH-1 — manifest-type confusion.** `classify_manifest` (oci_v2.rs:1829) checks `.get("manifests").as_array()` FIRST, so a runnable image manifest carrying a cosmetic `"manifests":[]` classifies as Index → `oci_manifest_requires_proxy_scan`=false → gate SKIPPED → falls through to the #1274 'cached UNSCANNED' path → **200, image manifest served verbatim, no grype run**. Identical content without the field → 423. (live: manifest `frank`)

**HIGH-2 — layer-marker confusion.** `oci_manifest_is_scannable_image` (scanner_service.rs:259) returns false if ANY layer mediaType contains a non-image marker (wasm/cosign/in-toto/spdx/cyclonedx). Add ONE such dummy layer alongside the real tar+gzip vulnerable layer → gate skipped → **200 'cached UNSCANNED'**, vulnerable layer still pullable. (live: manifest `cosign`)

**HIGH-3 — blob-GET bypass.** `handle_get_blob` (oci_v2.rs:4419) has no scan gate. A manifest blocked 403, yet `GET .../blobs/<config>` and `.../blobs/<vulnerable-layer>` both return **200** with digest-verified bytes → the image is reconstructable from public descriptors. (live confirmed)

Root cause of 1/2 = the same npm-lesson: the 'is this a gated runnable image?' predicate is UNDER-INCLUSIVE vs what docker/containerd runs (image-manifest mediaType + image-config descriptor + real layer = runnable, regardless of a cosmetic index-array or a decoy non-image layer).

**Fix:**
1. **Consumer-faithful gating (1/2):** gate ANY manifest docker would run as an image — i.e. mediaType is an image manifest AND it has an image-**config** descriptor AND ≥1 real image layer — EVEN IF it also carries a `manifests` array or an extra non-image-marker layer. A manifest that is simultaneously image-and-index → GATE (or reject). Only a PURE index (config absent, only `manifests` entries) passes to child re-gating. When scanning, scan the real image layers; a decoy non-image layer must not exempt the image.
2. **Blob-path gate (3):** the manifest gate must imply a blob gate. When a manifest scan yields VULNERABLE, record its referenced config+layer digests as blocked-for-this-repo (you already eager-stage them, so you have the digests); `handle_get_blob` returns 403 for a digest on that repo's vulnerable-blocklist. (Document the residual: a blob for an image whose manifest was NEVER scanned can't be known-vulnerable — same limit as curation/quarantine; but a scanned-vulnerable image's blobs must not serve.)

Re-verify all 3 repros → 403/423, the plain/clean/index/child/fail-open/#2976 cases unchanged, no over-restriction. Holding for the fix + re-red-team.

PR #3027, opened 2026-07-30
fix(lifecycle): use OCI tag last push for max age

### LINKED ISSUE #3026: OCI max_age_days expires mutable tags based on their first push
## Description

A mutable Docker/OCI tag such as `latest`, `stable`, or `nightly` is stored in a single `artifacts` row keyed by `(repository_id, path)`, where the path is `v2/<image>/manifests/<tag>`.

Re-pushing that tag executes an upsert that updates the digest, storage key, uploader, `is_deleted`, and `updated_at`, but intentionally preserves the row's original `created_at`. The corresponding `oci_tags.updated_at` is also refreshed on every manifest PUT.

`max_age_days`, however, selects and soft-deletes artifacts exclusively by `artifacts.created_at`. As a result, a manifest pushed today through a tag first created months ago is treated as months old. The current image behind the moving tag can therefore be removed immediately by the next lifecycle run. Re-pushing the tag resurrects the row but preserves the same old `created_at`, so the next run removes it again.

This affects every mutable OCI tag, not only `latest`.

## Environment Information

- Deployment type: Source-level and PostgreSQL integration-test reproduction
- Operating System: N/A; the behavior follows from the persisted timestamps and SQL predicates
- Architecture: N/A
- Database: PostgreSQL
- Reverse Proxy: N/A
- Network setup: N/A

## Backend Version

Current `main` at `855a8aee1589fdcd1606c1c26ad3d0792567c753`.

## Web UI Version

Current `artifact-keeper-web` `main` at `26d27c71e8e2caf315eb1166e2f17b3ff10f6a0a` for the stale "Last pushed" display symptom. The lifecycle bug itself is backend-only.

## Package Format

Docker / OCI

## Expected Behavior

For an OCI manifest referenced by a human-readable tag, `max_age_days` should measure age from the most recent successful manifest PUT for that tag, using `oci_tags.updated_at` or a dedicated `last_pushed_at` value.

Re-pushing `image:latest` today should reset its retention age. A `max_age_days = 90` policy must not match that tag until 90 days have elapsed without another successful push, regardless of when the tag path was first created.

The UI's "Last pushed" value should use the same last-push timestamp so display and retention semantics agree.

## Observed Behavior

- `oci_tags.updated_at` and `artifacts.updated_at` advance on each manifest PUT.
- `artifacts.created_at` remains the timestamp of the first push to that tag path.
- The current web grouped view renders `group.manifest.created_at`, so "Last pushed" remains stale.
- `max_age_days` compares the unchanged `artifacts.created_at` against the cutoff and soft-deletes the row even when its current digest was pushed recently.
- If a fresh version tag points to the same digest, the lifecycle cascade may remove the stale moving-tag alias because the version tag remains as a surviving digest reference. `docker pull image:latest` then fails while the version tag still works.
- If the moving tag is the sole digest reference, the current last-protecting-tag guard can retain `oci_tags`, but the soft-deleted artifact disappears from repository listings, leaving inconsistent UI and registry state.
- A later re-push sets `is_deleted = false` but still preserves the old `created_at`, so the behavior repeats on the next lifecycle sweep.

## Steps to Reproduce

1. Create a local Docker/OCI repository.
2. Push `image:latest` with digest `D0`.
3. Simulate passage of time by setting the matching artifact row's `created_at` and `updated_at` to more than 90 days ago.
4. Push a fresh manifest `D1` to `image:latest`. Optionally also push a fresh immutable version tag such as `image:1.2.3` pointing to `D1`.
5. Verify the timestamp split:
   - `artifacts.created_at` for `v2/image/manifests/latest` is still older than 90 days.
   - `artifacts.updated_at` and `oci_tags.updated_at` are current.
   - the artifact row points to `D1` and is live.
6. Create a `max_age_days` lifecycle policy with `days = 90` for the repository.
7. Run a dry-run or execute the policy.
8. Observe that `image:latest` is matched and its artifact row is soft-deleted even though `D1` was pushed moments earlier. With a surviving version-tag sibling, the `latest` tag row can also be pruned.
9. Re-push `image:latest`, run the policy again, and observe that it is immediately matched again because `created_at` never moved.

## Logs

No runtime error is emitted. The behavior is driven by the timestamp and lifecycle SQL predicates.

Relevant write path:

```sql
ON CONFLICT (repository_id, path) DO UPDATE SET
  checksum_sha256 = EXCLUDED.checksum_sha256,
  storage_key = EXCLUDED.storage_key,
  is_deleted = false,
  updated_at = NOW()
-- created_at is preserved
```

Relevant lifecycle predicate:

```sql
AND created_at < NOW() - make_interval(days => ...)
```

### Current source evidence

- OCI tag upsert refreshes `oci_tags.updated_at`: https://github.com/artifact-keeper/artifact-keeper/blob/855a8aee1589fdcd1606c1c26ad3d0792567c753/backend/src/api/handlers/oci_v2.rs#L2111-L2116
- Manifest artifact upsert refreshes `updated_at` but not `created_at`: https://github.com/artifact-keeper/artifact-keeper/blob/855a8aee1589fdcd1606c1c26ad3d0792567c753/backend/src/api/handlers/oci_v2.rs#L7804-L7815
- `max_age_days` matches and deletes by `created_at`: https://github.com/artifact-keeper/artifact-keeper/blob/855a8aee1589fdcd1606c1c26ad3d0792567c753/backend/src/services/lifecycle_service.rs#L900-L952
- The backend already exposes the correct `last_pushed_at` from `oci_tags.updated_at`: https://github.com/artifact-keeper/artifact-keeper/blob/855a8aee1589fdcd1606c1c26ad3d0792567c753/backend/src/api/handlers/repositories.rs#L6256-L6264
- The current web view still renders `artifact.created_at`: https://github.com/artifact-keeper/artifact-keeper-web/blob/26d27c71e8e2caf315eb1166e2f17b3ff10f6a0a/src/app/%28app%29/repositories/_components/docker-tag-list.tsx#L215-L217

### Related work

- The same moving-tag reproduction was documented in #1682: https://github.com/artifact-keeper/artifact-keeper/issues/1682#issuecomment-4688394291
- #1987 closed #1682 by adding a last-protecting-tag guard, but it did not change the `max_age_days` timestamp predicate: https://github.com/artifact-keeper/artifact-keeper/pull/1987
- The web-side switch to the backend's correct `last_pushed_at` is already proposed in artifact-keeper-web#561: https://github.com/artifact-keeper/artifact-keeper-web/pull/561

## Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate.
- [x] I have redacted any sensitive information (passwords, tokens, private URLs).
- [x] I am using a supported version.

junsung-cho:
## Summary

Fixes #3026.

- Age live OCI tag-backed manifest artifacts by `oci_tags.updated_at`, which is refreshed on every successful manifest PUT, instead of the path-keyed artifact row's original `created_at`.
- Match the exact live repository, digest, image path, and tag mapping; retain `artifacts.created_at` as the fallback for non-OCI artifacts and unmatched legacy rows.
- Use the same effective timestamp expression for repository-scoped/global preview and deletion queries.
- Update existing max-age OCI fixtures so both the artifact and authoritative tag timestamp are genuinely old.

`test_max_age_days_uses_last_push_for_moving_oci_tag` reproduces a 91-day-old `latest` artifact row being moved to a fresh digest. It first proves that the pre-fix `created_at` predicate incorrectly matches exactly that row, then verifies that both dry-run and execution match/remove zero artifacts and preserve `latest` plus its immutable sibling.

## Test Checklist

- [ ] E2E tests added/updated (not applicable; this lifecycle SQL path is covered by PostgreSQL integration tests)
- [x] Manually tested locally

Validation performed:

- `cargo fmt --check`
- `cargo test --workspace --lib` — 14,550 passed, 6 ignored
- `cargo test --test lifecycle_policy_tests -- --ignored --nocapture` — 23 passed
- Focused moving-tag regression against PostgreSQL 17.10 — passed
- Expanded `jscpd` measurement for the changed service file — 2.2% duplication

## API Changes

- [x] N/A - no API changes

### DIFF
--- a/backend/src/services/lifecycle_service.rs
+++ b/backend/src/services/lifecycle_service.rs
@@ -30,12 +30,20 @@ use crate::storage::keys::prefix_matches;
 
 /// Build a max-age query with a shared effective timestamp expression.
 ///
-/// Mutable OCI tags reuse the same `artifacts` row for every manifest PUT.
-/// Their `created_at` therefore describes when the tag path was first seen,
-/// not when its current digest was pushed. `oci_tags.updated_at` is refreshed
-/// by every successful manifest PUT, so it is the authoritative age for an
-/// exact live `(repository, digest, image, tag)` mapping. All other artifacts
-/// retain the historical `created_at` behavior.
+/// An OCI manifest artifact ages from the matching logical reference row in
+/// `oci_tags`, keyed by `(repository, image, reference)`, rather than from the
+/// path-keyed artifact row's first-seen `created_at`. This applies to both
+/// human-readable tags and digest-shaped references. When no matching tag row
+/// exists, `COALESCE` retains the historical `artifacts.created_at` fallback
+/// used by non-OCI and legacy/content-addressed rows without tag metadata.
+///
+/// `oci_tags.updated_at` is deliberately a last-publish/cache-fill clock. Every
+/// successful hosted manifest PUT refreshes it, including a re-push of an
+/// unchanged digest because the tag upsert has no `IS DISTINCT FROM` guard. A
+/// valid Remote manifest cache fill (or refill after a local miss) that records
+/// a tag row also refreshes it; a normal warm Remote pull returns from local
+/// storage before the cache-upsert path and therefore does not extend
+/// retention.
 ///
 /// The tag row is reached with a `LEFT JOIN`, not a correlated subquery. The
 /// subquery form is quadratic here: `oci_tags_repository_id_name_tag_key` is
@@ -59,14 +67,15 @@ use crate::storage::keys::prefix_matches;
 ///
 /// The join deliberately does NOT also require
 /// `a.storage_key = 'oci-manifests/' || ot.manifest_digest`. That conjunct adds
-/// no identification -- path and version already pin the row -- and it converts
-/// a partial write into data loss: the `oci_tags` upsert is transactional while
-/// `upsert_manifest_artifact` is best-effort and only logs on failure, so the
-/// two tables can disagree on digest while the push still returns 201. With the
-/// conjunct, that disagreement makes the join miss, `COALESCE` falls back to the
-/// stale `created_at`, and a tag pushed seconds ago is soft-deleted -- exactly
-/// the bug this code exists to fix, silently resurrected on any partial write.
-/// Verified both ways against a live database.
+/// no identification for the logical reference -- path and version already pin
+/// the row -- and makes a partial write fail toward deletion: the `oci_tags`
+/// upsert is transactional while `upsert_manifest_artifact` is best-effort and
+/// only logs on failure, so the two tables can disagree on digest while the push
+/// still returns 201. With the conjunct, that disagreement makes the join miss,
+/// `COALESCE` falls back to the stale `created_at`, and the live reference's
+/// artifact row is soft-deleted despite its fresh tag mapping. The `oci_tags`
+/// row still protects the current manifest bytes, but the reference disappears
+/// from artifact-backed listings. Verified both ways against a live database.
 macro_rules! max_age_from_where {
     ($repository_predicate:literal, $days_parameter:literal) => {
         concat!(
@@ -1426,7 +1435,7 @@ mod tests {
     // -----------------------------------------------------------------------
 
     #[test]
-    fn test_max_age_timestamp_uses_exact_oci_tag_update_with_created_fallback() {
+    fn test_max_age_timestamp_uses_oci_reference_update_with_created_fallback() {
         for sql in [
             MAX_AGE_SCOPED_SELECT_SQL,
             MAX_AGE_GLOBAL_SELECT_SQL,
@@ -1463,7 +1472,7 @@ mod tests {
             // created_at -- the very bug this predicate exists to fix.
             assert!(
                 !sql.contains("ot.manifest_digest"),
-                "joining on manifest_digest turns a partial write into data loss: {sql}"
+                "joining on manifest_digest makes a partial write fail toward deletion: {sql}"
             );
         }
     }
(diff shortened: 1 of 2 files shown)


brandonrc:
Ran this through the review pipeline: reproduced #3026 on the base, verified the fix, then three independent review passes (SQL, Rust correctness, test quality) against `80cfa131` on real Postgres with all 170 migrations.

**The diagnosis and the fix are right.** I confirmed the bug on base — a manifest pushed *today* soft-deleted by a 90-day policy — and confirmed the fix spares it while a genuinely old OCI tag still expires. The negative control is real, so this isn't "stop expiring OCI tags."

Two things I want to credit specifically, because they're the parts reviews usually find missing:

* `test_max_age_days_uses_last_push_for_moving_oci_tag` asserts `legacy_match_count == 1` against the *pre-fix* predicate before running the policy. That's a genuine precondition oracle — it proves the fixture actually reproduces #3026 rather than passing vacuously. Reverting the fix makes it fail with the exact bug signature. This is the right pattern.
* `simulate_oci_manifest_push` is faithful to the real write path — same conflict targets, same `DO UPDATE` sets, same write order, same `path`/`version`/`storage_key` shapes. It's materially better than the older helper and I'd support migrating the rest of the file onto it.

**Also worth stating plainly, because a reviewer might suggest otherwise:** the correlated subquery is the *correct* design here. The obvious simplification — read `artifacts.updated_at` and skip the join — does not work, because `artifacts.updated_at` is written by roughly thirty call sites across the codebase (quarantine, scans, metadata, per-format handlers). It means "last touched by anything," not "last pushed." Keying on `oci_tags.updated_at` is what makes the signal mean what the fix needs it to mean.

---

**One blocker, and it's performance.** Two reviewers measured it independently on different fixtures and got the same shape:

| fixture | pre-fix | this PR | remedy |
|---|---|---|---|
| 40k artifacts / 20k tags | 14 ms | **13,770 ms** | 84.8 ms (expression index) |
| 80k artifacts / 40k tags | 21.5 ms | **55,930 ms** | 54 ms (LEFT JOIN) |

Three compounding causes visible in the plan:

1. **Parallelism is lost.** The base plans a `Parallel Seq Scan` with workers; this plans a serial `Seq Scan`.
2. **The subplan runs once per artifact row** (`loops=80000`), including every non-OCI row that can never match. There's no cheap pre-filter such as `a.storage_key LIKE 'oci-manifests/%'` to short-circuit them.
3. **The index probe is unbounded in the middle column.** `a.path = 'v2/' || ot.name || '/manifests/' || ot.tag` is non-sargable on the `ot` side, so `Index Cond` constrains `repository_id` and `tag` but not `name`, and each probe walks the whole repository range — 31.9M buffer hits.

Why this is a blocker rather than a nit: it runs twice per execution (count, then UPDATE), inside the transaction whose own comment says it commits immediately "so the row locks on `artifacts` release before any further pool work, unblocking concurrent uploads/scans", on a 6-hour cron — and the global variant has no repository scope at all, so it's deployment-wide. Concurrent `docker push` upserts the same rows.

The `LEFT JOIN` rewrite is the one I'd take, since at-most-one match is already guaranteed by `UNIQUE(repository_id, name, tag)` — which is what the `MAX()` comment concedes — and it needs no migration. Verified result-identical on the same data.

**One high-severity correctness issue.** When `artifacts` and `oci_tags` disagree on digest, the subquery returns NULL and `COALESCE` falls back to `a.created_at` — the exact stale timestamp this PR exists to stop trusting. It fails *toward deletion*. Demonstrated: a manifest pushed seconds ago marked for deletion. This is reachable, not hypothetical — the `oci_tags` write is transactional while the `artifacts` upsert is best-effort and its failure is only logged, so one failed upsert leaves the tag fresh and the artifact row stale permanently.

**On the two new `--lib` tests.** They're text pins, and it's worth knowing exactly what they do and don't buy. A reviewer mutated the production SQL two ways and ran them verbatim:

* swap the `COALESCE` arguments — since `created_at` is `NOT NULL` it always wins and #3026 is fully restored → both tests **pass**
* flip `<` to `>` — the policy then soft-deletes everything *newer* than N days → both tests **pass**

They survive because `contains(...)` is true regardless of *arrangement*, and nothing asserts the comparison operator. They do have real value against "someone edits this string back", which is a legitimate guard for a `sqlx::query` constant that gets no compile-time checking — so keep them, just don't read them as semantic coverage. Asserting the whole predicate as one exact substring (or snapshotting the constant) catches both mutations for the same cost.

This matters more than usual because the coverage gate runs `cargo nextest --lib`, so `backend/tests/` scores zero toward it. These two tests are the fix's entire `--lib` footprint. The good news the gate can't see: the integration test *does* run in CI as its own `--ignored` job, so the genuine guard is enforced.

Also: `assert!(sql.contains("a.created_at"), "non-OCI artifacts must retain created_at age semantics")` is individually vacuous — it passes in the mutation where `a.created_at` is the *only* predicate, i.e. it asserts the opposite of what its message claims.

---

**To land this:** the query-shape fix for the regression, and a `COALESCE` fallback that doesn't fail toward deletion on digest drift. Both are contained, and the macro's SELECT/UPDATE-share-one-definition structure — which is a real improvement over the two hand-maintained copies — makes the first one a single edit.

Two adjacent gaps I'm filing separately rather than loading onto you: `no_downloads_days` has the identical unfixed `created_at` assumption (and is pinned by a currently-green test that backdates only `artifacts.created_at`), and the global unscoped max-age constants are executed by no test anywhere.

Heads up on a collision: PR #2999 rewrites `execute_max_versions` in this same file. Git merges both cleanly, but they'd disagree about which Docker tag is "newest" — #2999 leaves that ordering on `created_at`, the signal this PR establishes is wrong for mutable tags. Whichever lands second should carry this expression across, ideally as a shared helper rather than two definitions in one file.

brandonrc:
Follow-up: a third review pass (SQL/schema) finished after my comment above. It adds material, and it **refutes** one concern, so recording both.

**Refuted — the proxy concern is unfounded.** An earlier pass flagged (as plausible, unverified) that on Remote repos every cache-fill bumps `oci_tags.updated_at`, turning `max_age_days` into "max *idle* days" so a regularly-pulled proxy tag would never age out. That does not hold. A warm pull resolves locally first — `lookup_manifest_tag_row` → `resolve_local_manifest_from_storage` → `build_local_manifest_response` — and returns *above* `cache_manifest_or_compute_digest`, which is the only caller that upserts `updated_at`. There's no TTL revalidation on that path, and the `#3003 PR-2` comment says so directly: "A Remote repo caches the manifest on the first pull, so every warm pull lands here." So `updated_at` records last cache-*fill*, not last pull, and this is not a storage-reclaim behavior change. Good — that was the one open question on the design.

**The blocker's root cause, precisely.** `oci_tags_repository_id_name_tag_key` is `btree (repository_id, name, tag)`. The subquery constrains `repository_id` and `tag`, but `name` — the *middle* column — appears only inside the concatenation, so it's unconstrained and Postgres scans the entire `repository_id` range on every outer row: 851 buffers per row, 85M total. Third independent measurement, third fixture: 36 ms → 194,800 ms at 200k artifacts / 100k tags. `LEFT JOIN` rewrite measured at **81 ms** with an identical result set (`count=142491` both ways); an expression index on `(repository_id, (('v2/'||name||'/manifests/'||tag)))` gets 413 ms without touching the query text.

**Two things confirmed good, worth knowing they were checked:**

* SELECT/UPDATE parity is **byte-identical** — the four generated constants were compiled standalone and md5'd from `WHERE` onward; scoped and global each hash-match their UPDATE counterpart. The macro genuinely eliminates the drift risk the four hand-maintained copies carried.
* `a.version = ot.tag` is **load-bearing**, not redundant. Without it a slash-bearing image name produces two matches — `name='a/manifests/b', tag='c'` and `name='a', tag='b/manifests/c'` both reconstruct `v2/a/manifests/b/manifests/c`. With `tag` pinned, `name` is fully determined and `UNIQUE` guarantees ≤1. Also on `MAX()`: with a duplicate present it resolves to the newest, i.e. retains rather than deletes — the conservative direction for a retention policy. Keep it.

**One more semantics item (medium).** By-digest manifests don't reliably keep `created_at`. `v2/app/manifests/sha256:EEE` ages by last push when an `oci_tags` row exists with `tag = <digest>`, and by `created_at` when it doesn't — and that row shape isn't exotic: `cached_manifest_reference_key` returns the digest as the tag for Remote repos, so every proxy-cached manifest gets one. Two structurally identical rows end up on different clocks. Defensible for content-addressed objects, but the doc comment says "all other artifacts retain the historical `created_at` behavior", which no longer describes it.

**And a semantic worth naming in the comment:** the tag upsert is `ON CONFLICT ... DO UPDATE SET updated_at = NOW()` with no `IS DISTINCT FROM` guard, so re-pushing an *unchanged* digest still bumps the clock. A CI job that re-pushes an identical image on a schedule keeps that tag alive indefinitely under `max_age_days`. Arguably correct — it is still actively published — but it should be a stated choice rather than a side effect.

brandonrc:
Pushed `85aafe0` addressing the blocker and the digest-drift issue. Both changes are measured and mutation-tested rather than argued — details so you can check my work.

**The join rewrite.** Reproduced the regression independently on a third fixture (50k artifacts / 25k tags): **58,946 ms** as a correlated subquery, **28 ms** as a `LEFT JOIN`, both returning exactly 25,000 rows. The `UPDATE` can't carry a `LEFT JOIN`, so the shared `FROM`/`WHERE` is reused verbatim inside an `IN` subselect — both statements still derive from one definition, so the dry-run count and the real run can't drift, which is what your macro was for in the first place. Verified: `UPDATE 25000`, with all 25,001 freshly-pushed OCI manifests still live.

**The digest conjunct is gone, and this is the part I'd most want your eyes on.** `a.storage_key = 'oci-manifests/' || ot.manifest_digest` adds no identification — `a.path` and `a.version` already pin `(repository_id, name, tag)` uniquely, which is exactly why `a.version = ot.tag` is load-bearing and stays. What the conjunct does add is a failure mode: the `oci_tags` upsert is transactional while `upsert_manifest_artifact` is best-effort and only logs on failure, so the two tables can disagree on digest while the push still returns 201. With the conjunct the join then misses, `COALESCE` falls back to the stale `created_at`, and a tag pushed seconds ago gets soft-deleted — this bug, silently resurrected on any partial write. Verified both ways against a live database:

```
              variant               | would_delete
------------------------------------+--------------
 digest conjunct KEPT               | t     <- manifest pushed today, deleted
 digest conjunct DROPPED            | f
```

**On the two unit tests** — I tightened rather than removed them, because the "don't edit this string" guard is genuinely useful for a `sqlx::query` constant that gets no compile-time checking. The problem was the *form*: independent `contains(...)` fragments pass on any arrangement of the same vocabulary. Three mutations that fully restore the bug all passed the fragment version and all fail the exact-substring version:

| mutation | effect | fragments | exact |
|---|---|---|---|
| swap `COALESCE` args | `created_at` wins unconditionally (it's `NOT NULL`) — original bug restored | pass | **fail** |
| flip `<` to `>` | soft-deletes everything *newer* than the window | pass | **fail** |
| re-add digest conjunct | destructive fallback returns | — | **fail** |

I also dropped `assert!(sql.contains("a.created_at"), "non-OCI artifacts must retain created_at age semantics")` — it passes in the mutation where `created_at` is the *only* predicate, so it asserted the opposite of its message.

**What I did not change**, deliberately: the by-digest manifest inconsistency (rows age by last push when an `oci_tags` row exists with `tag = <digest>`, which the proxy path routinely writes, and by `created_at` otherwise), and the missing positive control — every new assertion in the integration test is `== 0`, so a predicate matching *nothing* would satisfy them all. The `legacy_match_count` pre-assert compensates substantially, which is why I left it rather than reworking your test. Both are worth a follow-up; neither blocks.

Gates on the pushed commit: `cargo fmt --check` clean, `cargo clippy --lib --all-features -- -D warnings` clean, `cargo test --lib` 14,550 passed, and the `lifecycle_policy_tests` integration suite 23/23 including your `test_max_age_days_uses_last_push_for_moving_oci_tag`.

Thanks for a well-diagnosed fix — the `legacy_match_count` precondition oracle and the faithful `simulate_oci_manifest_push` fixture are both better than what the file had, and I'd like to see the rest of the suite migrated onto that helper.

brandonrc (approved):
Approving. Full pipeline: reproduced #3026 on the base, verified the fix with a real negative control, three independent review passes, then two remediations pushed as `85aafe0` and re-verified.

The diagnosis was right and the integration test's `legacy_match_count` precondition oracle is the pattern I'd like to see more of — it proves the fixture reproduces the bug against the old predicate before asserting the new behavior, so it can't pass vacuously.

What changed after review: the correlated subquery became a `LEFT JOIN` (58,946 ms → 28 ms on 50k artifacts / 25k tags, identical result sets), the digest conjunct came out because it converted a best-effort partial write into silent data loss, and the unit assertions are pinned as exact substrings — three bug-restoring mutations that passed the original `contains` fragments now all fail.

All 10 checks green. Coverage failed once on an `aws-lc-sys` C build error at 2m22s rather than on the threshold; the rerun passed.

Two follow-ups deliberately left off this PR: the by-digest manifest ages inconsistently depending on whether an `oci_tags` row exists with `tag = <digest>` (the proxy path routinely writes one), and every new assertion in the integration test is `== 0`, so a positive control would strengthen it. Neither blocks.

Not merging — labeled `ready-to-merge` for the maintainer.

junsung-cho:
Thanks for the thorough review, and for pushing `85aafe0`. I fetched and reviewed that commit before adding `34b02e43` on top.

`85aafe0` addresses the blocking SQL issue with the set-based `LEFT JOIN`, removes the digest conjunct, and strengthens the exact SQL-string mutation guards.

My follow-up commit covers the remaining confirmed items:

- documents that exact digest-shaped references use `oci_tags.updated_at`, while rows without matching tag metadata fall back to `artifacts.created_at`;
- documents that re-pushing an unchanged digest refreshes the publish clock;
- clarifies that a valid Remote cache fill refreshes the clock, while a warm Remote pull returns locally before the cache-upsert path;
- adds PostgreSQL-backed scoped/global tests covering human-readable tags, digest references, missing-tag fallback, digest drift, slash-bearing path ambiguity, dry-run/UPDATE parity, and positive deletion controls.

I kept `a.version = ot.tag` and added a DB regression for the ambiguous `/manifests/` path case. I did not reintroduce `MAX(ot.updated_at)`: once `tag` is pinned, the path determines `name`, and `UNIQUE(repository_id, name, tag)` guarantees at most one matching row. Re-aggregation would therefore add work without preserving a reachable schema state.

Issue #2783, opened 2026-07-21
Nexus migration: Nexus 'group' repos do not correlate to AK Virtual repos


brandonrc:
Reported by customer Jon Craig (on AK 1.6.0) — 2026-07-21 feedback. CRITICAL. When migrating from Nexus, Nexus 'group' repositories don't map correctly to AK 'Virtual' repositories — imported virtual repos are broken (frontend + backend errors), forcing a full clean redeploy. NOTE: the DELETE-failure half is fixed separately (#2776/#2778, 1.6.1); THIS issue is the migration MAPPING/correlation defect (member correlation wrong or incomplete). Reproduce a Nexus group import and fix the virtual-repo member correlation. 1.6.x backport candidate.

PR #3038, opened 2026-07-30
feat(migration): rename repos on import via repo_mappings

### LINKED ISSUE #3035: Migration: allow renaming repos on import (source → target mapping)
### Problem Statement

Migration always keeps the source repo's name — `prepare_repository_migration` hardcodes `target_key = repo.key`. 

So there's no way to import a repo under a different name (e.g. `network-team-python` → `network-pypi`), which you'd want when tidying up names during a Nexus/Artifactory move

### Proposed Solution

Optional `repo_mappings` (source key → target key) on the job config. 

When set, override the mapped repo's target before provisioning; the existing conflict-reuse logic transfers into the renamed repo. Unmapped repos unchanged, empty map == today. 

Config is a JSON blob, so no DB migration.

### Alternatives Considered

_No response_

### Use Case

_No response_

### Component

None

### Additional Context

_No response_

### Pre-submission Checklist

nicola-preda:
## Summary
Adds optional `repo_mappings` (source repo key → target repo key) so a repo can
migrate under a new name. Before, the destination always reused the source's key
(`prepare_repository_migration` set `target_key = source_key`), so there was no
way to rename during a migration.

- `MigrationConfig` gains `repo_mappings` (`#[serde(default)]`) — `models/migration.rs`.
- `apply_repo_mappings` retargets each resolved repo's `target_key` after resolve,
  before provisioning — `migration_worker.rs`.
- The transfer path now separates source vs destination via a `RepoKeys{source,target}`
  threaded through `process_repository_artifacts`/`process_single_artifact`/`transfer_artifact`:
  it **reads the source by source key** (listing, download, properties, OCI
  referenced-content walk) and **writes AK by target key** (storage, dup-check,
  repo-id + artifacts insert, stored path). Without this split a renamed repo would
  query the source for the new name and migrate empty.
- Unmapped repos / empty map == current behavior.

Config is a JSON blob (`MigrationJob.config`), so this is additive with **no DB
migration**; `MigrationConfig` already derives `ToSchema`, so OpenAPI picks it up.

Closes #3035

- [x] N/A — not a bug fix

## Test Checklist
- [x] Unit tests added/updated (`apply_repo_mappings` override + no-op; migration_worker suite green)
- [x] Manually tested locally

## API Changes
- [x] Request/response types have `#[derive(ToSchema)]` (`MigrationConfig`)
- [x] N/A - additive config field, no new endpoints or DB changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Migration jobs can rename repositories on import** (#3035): `MigrationConfig.repo_mappings` maps a source repository key to the key it lands under (e.g. `network-team-python` -> `network-pypi`); unmapped repositories keep their source name, and a renamed member of a migrated virtual repository keeps its membership. Targets are validated like a hand-created repository key and two sources may not share one destination; either fails the job rather than migrating under the original name. A target that already exists is reused and the source's artifacts are merged into it, logged as a warning naming both repositories.
+
 ### Changed
+
+- **`MigrationConfig.dry_run` no longer writes** (#3035): it gated only the artifact upload, so a preview created every destination repository and wrote `artifacts` plus OCI index rows pointing at bytes it never uploaded. A dry run now logs the destination plan (create or reuse, under which key), skips provisioning and virtual member correlation, and enumerates the source without downloading or writing, so it still reports the artifact count and byte total it would transfer.
 
 - **BREAKING API SEMANTICS: `PUT /api/v1/repositories/{key}/members` (virtual repository members) is a full-set replace** (#2795, ratified in #2899): as of #2795 the request body is the *complete* desired member list, not a partial update. Any existing member whose key is absent from the body is **removed**; members present in the body are inserted if missing, or have their priority updated if already present; an empty `members` array removes every member. Previously the endpoint only reordered members that already existed and returned 404 for a member that was not already present, which made it impossible to add or remove a member when editing a virtual repository after creation. **API clients that send a partial member list in order to reorder priorities will now delete the members they omit** and must be updated to always send the complete member list. The web UI (`virtual-members-panel.tsx` move-up/move-down) and the CLI (`members reorder`, documented as a bulk full-set operation) already send the full list and are unaffected. The behavior is now stated in the OpenAPI operation description and pinned by a regression test.
 
--- a/backend/src/models/migration.rs
+++ b/backend/src/models/migration.rs
@@ -121,6 +121,8 @@ pub struct MigrationConfig {
     pub exclude_paths: Vec<String>,
     /// Rename repos as they migrate: source key -> target key
     /// (e.g. network-team-python -> network-pypi). Empty keeps source names.
+    /// Targets are validated like a hand-created repository key, and two
+    /// sources may not map onto one target.
     #[serde(default)]
     pub repo_mappings: std::collections::HashMap<String, String>,
     #[serde(default = "default_true")]
--- a/backend/src/models/migration.rs
+++ b/backend/src/models/migration.rs
@@ -119,6 +119,10 @@ pub struct MigrationConfig {
     pub exclude_repos: Vec<String>,
     #[serde(default)]
     pub exclude_paths: Vec<String>,
+    /// Rename repos as they migrate: source key -> target key
+    /// (e.g. network-team-python -> network-pypi). Empty keeps source names.
+    #[serde(default)]
+    pub repo_mappings: std::collections::HashMap<String, String>,
     #[serde(default = "default_true")]
     pub include_users: bool,
     #[serde(default = "default_true")]
--- a/backend/src/services/migration_worker.rs
+++ b/backend/src/services/migration_worker.rs
@@ -407,7 +407,10 @@ impl MigrationWorker {
         let mut plan = resolve_repos_for_provisioning(&repos, &source_repos);
         // Honor any source -> target repo renames before provisioning destinations,
         // so the conflict pre-pass below reuses the renamed repo instead of the source.
-        apply_repo_mappings(&mut plan.resolved, &config.repo_mappings);
+        if let Err(why) = apply_repo_mappings(&mut plan.resolved, &config.repo_mappings) {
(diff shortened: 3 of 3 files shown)


brandonrc:
Ran this through the pipeline. **The part you worked hardest on is right:** the `RepoKeys{source,target}` split through `process_repository_artifacts` → `process_single_artifact` → `transfer_artifact` is correct and revert-proved — source reads use `keys.source`, destination writes use `keys.target`, and I checked every `repo_key` use in that range for mis-threading. It also meets every acceptance criterion in #3035 literally.

I've pushed two fixes as `a5fab81` rather than sending you round again, because both are small and one is a security boundary.

**`repo_mappings` values reached repository identity with no validation.** This is the first place an operator supplies a repository key to the import path directly, and that path never calls `validate_repository_key` — its only two call sites are repository create and repository rename. So a mapped target reached `create_repository` verbatim, landed in `repositories.key` (VARCHAR(255), no CHECK) and, for filesystem backends, was concatenated into `storage_path` as `{storage_base}/{target_key}`. Verified end-to-end:

```
ACCEPTED key="../../../../tmp/ak-escape"  storage_path="/var/lib/ak/storage/../../../../tmp/ak-escape"
bytes landed at: /tmp/ak-escape/pkg/1.0/file.txt   inside configured root? false
```

`FilesystemStorage` sanitizes the object key but never the base path, so nothing downstream catches it. All six probe values (`..`, `/`, leading `-`, spaces, a Cyrillic homoglyph, >128 chars) are rejected by `validate_repository_key` and were accepted here. It's admin-gated, so this is admin → filesystem-write-outside-storage-root rather than anything unauthenticated — but AK admin isn't meant to be equivalent to shell on the server, and the homoglyph case is separately nasty since `аdmin-python` renders identically to `admin-python` in the UI.

Targets now clear the same bar a hand-created key does, and a bad mapping **fails the job** rather than skipping the repository — a migration that silently lands under the original name is worse than one that refuses to start. Two sources mapped onto one target are rejected too; both would have transferred into the same destination and collided by artifact path.

**Renaming a repo silently dropped it from migrated virtual/group repos.** `RepositoryListItem.members` are source-side names, and `correlate_virtual_repo_members` resolves each with `SELECT id FROM repositories WHERE key = $1` — which only holds while `target_key == source_key`. The deleted comment at the old `resolve` site said exactly this: *"We don't currently rename repos; documenting the contract so a future change that breaks it gets caught here."* Verified:

```
lookup by SOURCE name (what correlate uses) -> 0
lookup by TARGET name (where the repo is)   -> 1
virtual_members_written                     -> 0
```

The migrated group ends up empty — the failure #2783 exists to fix. Worse, if an unrelated repo already occupies the source key the lookup resolves to *it*, and the group aggregates a repository that was never part of it. Members are now translated through the mapping before correlation. Also added a warning for mapping keys that matched no repository, which were previously ignored in silence.

---

**Two things I deliberately left for you, because they're semantics rather than defects.**

A mapping onto an **existing** repository currently logs "reusing" and transfers into it. `check_repository_conflict` compares only `repo_type` and `format` — never `created_by`, owner, or any ACL — so the answer to "is the target's ownership checked at map time or at create time" is neither, at any time. That's benign today because the nest is admin-gated and a global admin bypasses `check_repository_action` anyway, but it's a silent data merge into a production repo with no operator confirmation. I'd argue it should require an explicit opt-in flag; that's your call, not mine to make in someone else's PR.

Related: `dry_run` is checked in exactly one place, inside `transfer_artifact`. The provisioning loop calls `create_repository` unconditionally, so a dry run with a typo'd mapping still permanently creates the renamed repository. Pre-existing, but this feature is the first thing that makes it consequential — there's now no safe way to preview a rename plan.

Also worth knowing, not a PR defect: `migrate_permissions` / `migrate_groups` / `migrate_users` have no callers today, but `process_permission_target` resolves source repo keys the same way `correlate_virtual_repo_members` did. It'll inherit this bug whenever it's wired back into `process_job`, and `include_permissions` still defaults to `true`, which makes the dead code look live to a reader.

Gates: `cargo fmt --check` clean, `clippy --lib --all-features -D warnings` clean, `migration_worker` 182 passed / 0 failed. Both existing guards still revert-prove; the two new tests cover invalid targets and duplicate targets, which nothing covered before.

There's no documentation for `repo_mappings` anywhere outside the field's doc comment — worth a line before the UI follow-up lands.

nicola-preda:
@brandonrc Thanks for pushing `a5fab81`, kept both fixes. Two nits in it: a doubled `#[test]` left `test_apply_repo_mappings_overrides_only_mapped_target_key` unattributed so it stopped running (`clippy --all-targets` flags it, `--lib` doesn't), and the duplicate-target error string has a run of spaces in the middle. Both fixed.

On your two:

**Existing target: no flag.** Migrating into a repo that already exists is what we want, so an opt-in would just mean always setting it. It logs at `warn` now naming source and destination, instead of the same `info` line a plain re-run gets. If destination ownership should matter, that belongs in `check_repository_conflict` for all callers, not in a mapping-only flag.

**`dry_run`: went further than repo creation.** With an existing target the artifacts INSERT and the OCI registrations ran too, only the put was gated, so a preview could leave rows pointing at bytes it never uploaded. A dry run now logs the destination plan, skips provisioning and member correlation, and enumerates the source without downloading or writing, so you still get the artifact count and byte total it would transfer.

One wart: the transfer loop resolves storage for a destination that doesn't exist during a dry run, so it falls back to a placeholder handle from the registry. Nothing reads or writes through it, the dry-run branch is the first statement in the artifact loop. Happy to make the parameter `Option` instead if you'd rather.

**Docs:** CHANGELOG under Unreleased. Nothing else in the repo documents migration config.

Also: your duplicate-target check only sees the mappings, so `a -> b` while `b` migrates too slipped through and merged them. Rejected now, checked against the resolved plan after mapping, with the same-repo-requested-twice case excluded.

`process_job` had no test driving it, so the dry-run behavior is pinned by a new DB-gated one (`test_process_job_dry_run_reports_counts_and_writes_nothing`): renames a repo, asserts no destination repository, no `migration_items`, and reports the count and bytes it would move, with a mock that panics on download. Revert-proved against both new guards.

Still open: `process_permission_target` resolving source keys the same way, agreed it breaks when it's wired back into `process_job`. Want an issue for it?

Rebased on current main. `fmt`, `clippy --workspace --all-targets -D warnings`, and the migration tests green locally; the 5 unrelated failures I see (hex, pypi remote-metadata, usage-ledger) reproduce identically on `a5fab81`.

Issue #2120, opened 2026-07-01
Pull-through PyPI cache thrashes into permanent slow-path under multi-replica load — production evidence for #1609


ysolomon-plat:
## Description

Since our PyPI pull-through repo (backed by a horizontally-scaled backend, ~5-10 replicas behind a shared S3 object store) started taking sustained multi-replica concurrent traffic, the fast-path ETag revalidation added in #1051 has been almost continuously falling back to the slow path (full upstream re-fetch + re-cache) for a very large fraction of hosted PyPI packages, not just occasionally-changed ones.

We believe this is a **production-scale manifestation of #1609** (cross-replica single-flight is still per-process per #1355/#1631): concurrent cold-cache requests for the same artifact land on different backend replicas, each replica independently streams the artifact from upstream and completes its own multipart upload to the shared S3 backend, then pins whatever ETag `head_etag` returns immediately after its own write. Because **S3 multipart ETags are not content-addressable** (an opaque per-upload value, not a hash of the bytes — unlike single-part-PUT ETags, which equal the object's MD5), two replicas independently re-uploading byte-identical content can legitimately produce two different ETags. Whichever replica's sidecar gets read next by `is_fresh` almost always disagrees with whatever the S3 object currently holds, so the cache never "settles" — it re-fetches from upstream and re-uploads on nearly every request for the same artifact.

## Environment Information

- Deployment type: Kubernetes, S3-compatible object storage backend, ~5-10 backend replicas (HPA-scaled)
- Operating System: Linux (container)
- Architecture: amd64
- Database: PostgreSQL (managed)
- Reverse Proxy: internal ALB/ingress in front of the backend Service
- Network setup: private VPC, backend not directly internet-exposed

## Backend Version

Not directly observable from our vantage point (log-based, external to the deployment). The `pinned_etag` / `current_etag` fields in the warn log below indicate a version at or after #1051.

## Package Format

PyPI

## Expected Behavior

For a pull-through PyPI repo under concurrent multi-replica load, previously-cached artifacts that have not actually changed upstream should be served from the fast path (cache hit, no re-fetch) on repeat requests, even when multiple replicas raced to fill the same cache key initially.

## Observed Behavior

The backend logs a continuous stream of:

```
[...]

Issue #2075, opened 2026-06-30
Presigned-redirect fast path bypasses Package-Age-Policy quarantine (#1770) on redirect-capable backends


inspired-geek:
## Description

The proxy-cache **presigned-redirect fast path** gates only on cache freshness
(`is_cache_fresh`), not on Package-Age-Policy quarantine (#1558 / #1770). When
presigned downloads are enabled on a redirect-capable storage backend (S3 / GCS
/ Azure), a *fresh-but-held* cache entry is served as a presigned redirect to
the cached object — bypassing the 409/403 that the streaming and buffered fetch
paths enforce via `check_quarantine_until`. So a quarantined artifact can be
downloaded during its cooldown window on those deployments.

This is **pre-existing** behavior (not introduced by #2069 / PR #2074); it was
surfaced during the independent review of PR #2074, where the quarantine-aware
fast path was relocated into the `try_member_cache_redirect` helper but the gap
itself was left unchanged (out of scope for #2069). The Package-Age-Policy
quarantine is a supply-chain hardening feature (delaying freshly-published, and
possibly compromised, versions), so this bypass is hardening/security-relevant —
maintainers may want to handle it accordingly.

## Environment Information

- Deployment type: any with a **redirect-capable** object store (S3 / GCS / Azure)
- Storage: presigned downloads enabled (`presigned_downloads_enabled = true`)
- Policy: a repository with Package-Age-Policy quarantine configured (#1558)
- Network setup: pull-through cache / virtual repo serving proxied artifacts

## Backend Version

dev (main) — pre-existing; affects the `proxy_helpers` redirect fast path.

## Web UI Version

N/A (backend-only)

## Package Format

Any proxied format (Maven / npm / PyPI / OCI / ...).

## Expected Behavior

A quarantined (held) artifact returns 409/403 (or is skipped) on **every** serving
path, including the presigned-redirect fast path — the Package-Age-Policy block
must not be bypassable by the storage backend / download mode in use.

## Observed Behavior

On a redirect-capable backend with presigned downloads enabled, the fast path
(`try_member_cache_redirect` → `try_proxy_cache_redirect`, and likewise
`proxy_fetch_or_redirect` / `local_fetch_or_redirect`) issues a presigned
redirect for a fresh cache entry **without** consulting `quarantine_until`, so a
held artifact is served instead of returning 409/403.

## Steps to Reproduce

1. Configure Package-Age-Policy quarantine on a remote/virtual repository (#1558).
[...]

PR #3068, opened 2026-08-01
fix(proxy): probe object existence before presigned proxy-cache redirect in local_fetch_or_redirect

### LINKED ISSUE #3067: Presigned proxy-cache redirects are served for objects deleted out-of-band (e.g. S3 lifecycle expiry), with no upstream re-fetch
## Summary

When a proxy-cache object is deleted from the object store out-of-band — most commonly by an S3 lifecycle rule expiring cache entries by age — the corresponding `artifacts` row still exists, and `local_fetch_or_redirect` keeps issuing presigned 302 redirects to the deleted key. Every download of that package then 404s at the object store, permanently, until the object is manually restored or the row invalidated. The server never notices and never re-fetches from upstream.

This makes it unsafe to run any retention policy on the proxy-cache prefix, even though pull-through cache entries are exactly the kind of data a retention policy should be able to expire (they are re-fetchable from upstream by design).

## Reproduction

1. Remote (or virtual-with-remote-member) PyPI repo on an S3 backend with `presigned_downloads_enabled = true`.
2. Download any package through the proxy so it is cached (creates the `proxy-cache/...` object and an `artifacts` row).
3. Delete the cached `__content__` object directly in S3 (simulating a lifecycle expiration) — leave the DB untouched.
4. Download the same package again.

**Expected:** treated as a cache miss — re-fetch from upstream, repopulate cache, serve the file.
**Actual:** 302 redirect to a presigned URL for the deleted key → the object store returns 404 to the client. Repeats on every request.

## Root cause

`try_proxy_cache_redirect`'s doc contract says the caller must have "already done a metadata-only freshness check" before passing `cache_is_fresh = true`. Two of its three production callers honor that:

- `proxy_fetch_or_redirect` gates on `proxy_service.is_cache_fresh(...)` (metadata + existence/ETag probe, #1051/#1555) — a deleted object degrades into the slow path and an upstream re-fetch.
- `try_member_cache_redirect` does the same.

But `local_fetch_or_redirect` (`backend/src/api/handlers/proxy_helpers.rs`, the proxy-cache branch used by e.g. PyPI's `serve_file` for cached remote artifacts resolved via DB rows) passes `/* cache_is_fresh = */ true` **hardcoded**, with no probe of any kind. A DB row alone is sufficient to sign and serve the redirect.

Catching the failure after the redirect isn't possible: once the 302 is issued the client talks directly to the object store, so the server is out of the loop.

## Proposed fix

Gate the redirect in `local_fetch_or_redirect` on a metadata-only existence probe (`exists()`, HEAD-equivalent) through the same no-prefix cache handle it already signs with — capability-check first so non-redirecting backends never pay the probe, mirroring the #1555 ordering in `proxy_fetch_or_redirect`:

```rust
let object_present = b.supports_redirect()
    && matches!(b.exists(&artifact.storage_key).await, Ok(true));
try_proxy_cache_redirect(
    b.as_ref(),
    &artifact.storage_key,
    /* presigned_enabled = */ true,
    expiry,
    /* cache_is_fresh = */ object_present,
)
.await
```

On a missing object this falls through to the buffered read, whose `NotFound` lets virtual-member resolution continue to the remote member's upstream fetch — the cache repopulates and the DB row heals, with no manual intervention.

Cost: one HEAD request per redirect on this path — the same price the two sibling paths already pay for their freshness probe.

(A full `is_cache_fresh` probe would also check sidecar expiry/ETag, but it needs `(repo_key, path)` to derive `CacheKeys`, and this call site only has the row's `storage_key`; the existence probe covers the out-of-band-deletion failure mode precisely without re-deriving keys.)

PR incoming.

ysolomon-plat:
## Why

Fixes #3067. `local_fetch_or_redirect` calls `try_proxy_cache_redirect(..., cache_is_fresh = true, ...)` with a hardcoded `true`, so a DB row alone is enough to sign and serve a redirect — the object store is never probed. The sibling paths (`proxy_fetch_or_redirect`, `try_member_cache_redirect`) already gate on `is_cache_fresh`; this was the only call site that skipped the check, so an object deleted out-of-band (e.g. by a lifecycle rule) produced a dead 302 with no server-side fallback.

## What

Gate the redirect on an `exists()` probe before signing it, mirroring the sibling paths. A miss falls through to the buffered read, whose `NotFound` lets virtual-member resolution continue to an upstream fetch — cache and DB row both self-heal, no manual intervention. Cost: one extra HEAD per redirect on this path, same as the sibling paths already pay.

`cargo check`/`cargo fmt` pass.

## Follow-up: three unconditional cache-key deletes

Found while validating the fix: three separate spots call `storage.delete(cache_key)` unconditionally on a reject/miss path, without checking whether the key currently holds good content:

- `write_negative_cache` — every negative-cache renewal, even on a path that never had real content
- `tee_stream`'s empty/truncated-body reject arm
- `tee_stream`'s digest-mismatch reject arm

On a versioned backend, an unconditional delete doesn't clean up — it stacks a delete marker over the whole version history, which can silently hide real, previously-good content underneath (same failure mode as the bug above, different trigger).

`write_negative_cache` gets an `exists()`-before-`delete()` guard, since that path never writes content itself first — `exists()` genuinely reflects whether a prior positive cache entry is there to invalidate. Now covered by a test.

The two `tee_stream` reject arms are different: by the time either runs, the write they're rejecting has *already been committed* as the key's current version, so an `exists()` check there is always true and can't distinguish "only this bad write is present" from "a prior good version sits one behind it." Gating the delete doesn't fix anything on that path — it still fires. The actual fix is to not delete at all: the missing metadata sidecar (written separately, and skipped on every reject path) is what forces the self-heal refetch, so nothing depends on the object being gone too. Verified via the existing reject-path tests, updated to assert zero deletes instead of one.

Audited every other storage-delete call site in the backend (all format handlers, GC/backup/sync/upstream-feed jobs); everything else is already safely guarded or scoped to keys that structurally can't hold prior content.

### DIFF
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -3145,12 +3145,30 @@ pub async fn local_fetch_or_redirect(
         };
         let redirect = match &proxy_cache_backend {
             Some(b) => {
+                // The artifacts row alone does not prove the cached object
+                // still exists: retention tooling (e.g. an object-store
+                // lifecycle policy expiring cache entries by age) deletes
+                // proxy-cache objects without updating the database. A
+                // redirect signed for a missing key 404s at the object store,
+                // where the client is beyond any fallback we control — and the
+                // row keeps serving dead redirects until manual repair. So
+                // honor `try_proxy_cache_redirect`'s contract (caller performs
+                // a metadata-only freshness check first, as
+                // `proxy_fetch_or_redirect` does via `is_cache_fresh`): probe
+                // existence through the same no-prefix handle before signing.
+                // Capability check first, mirroring #1555 — never pay the
+                // probe on a backend that cannot redirect anyway. On a miss
+                // fall through to the buffered read below, whose NotFound lets
+                // virtual-member resolution continue to the remote member's
+                // upstream re-fetch, repopulating the cache.
+                let object_present = b.supports_redirect()
+                    && matches!(b.exists(&artifact.storage_key).await, Ok(true));
                 try_proxy_cache_redirect(
                     b.as_ref(),
                     &artifact.storage_key,
                     /* presigned_enabled = */ true,
                     expiry,
-                    /* cache_is_fresh = */ true,
+                    /* cache_is_fresh = */ object_present,
                 )
                 .await
             }
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -7135,14 +7135,12 @@ mod tests {
             }
         }
 
-        /// Redirect-capable but reports the key as absent — for the #2274/#1691
-        /// style existence-gate tests, where `supports_redirect() == true` must
-        /// not be enough on its own to trigger a presign.
-        fn new_with_exists(supports: bool, exists_result: bool) -> Self {
-            Self {
-                exists_result,
-                ..Self::new(supports)
-            }
+        /// Override the `exists()` result — for the #3067/#3068 existence-gate
+        /// tests, where `supports_redirect() == true` must not be enough on
+        /// its own to trigger a presign.
+        fn with_exists(mut self, exists_result: bool) -> Self {
+            self.exists_result = exists_result;
+            self
         }
     }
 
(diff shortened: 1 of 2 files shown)


ysolomon-plat:
Note on the two red checks — both appear unrelated to this change (one .rs file, no dependency changes):

- **Security Audit** fails on RUSTSEC-2026-0222 (severity 3.8, low), a new advisory published after `main`'s last CI run — the same audit is green on `main`. Any PR run today would hit it until the dep is bumped/allowlisted on `main`.
- **CI Complete** is currently failing on `main` as well, and this fork PR additionally has the secret-gated jobs (image build / integration / smoke) skipped.

The checks that exercise the change all pass: Check Rust, Backend Unit Tests (incl. the #1555 source-scan tests), Coverage. Happy to rebase once the advisory is handled on `main`.

ysolomon-plat:
Validated this fix end-to-end using `docker-compose.s3-maven-test.yml` (MinIO + Postgres, S3 storage). Setup notes for anyone reproducing:

- Requires **both** `PRESIGNED_DOWNLOADS_ENABLED=true` and `S3_REDIRECT_DOWNLOADS=true` — without the latter, `supports_redirect()` is false and the buffered-read fallback masks the bug (requests self-heal via upstream re-fetch).
- Requires an `artifacts`-table row whose `storage_key` points at the `proxy-cache/` object (e.g. from a registry migration/import). A fresh proxy fetch only writes `proxy_cache_artifacts` and resolves through `pypi_proxy_cache_redirect`, which probes `is_cache_fresh` correctly — so pure pull-through traffic never hits this path.

Repro: remote pypi repo (upstream pypi.org) + virtual wrapper; download a wheel through the virtual (caches it); insert the `artifacts` row; delete the `__content__` object in MinIO directly (simulating lifecycle expiry); re-request.

**Unpatched (v1.6.0):**
```
AK: 302 redirect
presigned target (valid signature): 404   <- dead redirect, repeats forever
```

**With this patch (cherry-picked onto v1.6.0):**
```
request 1: 200 direct (7,526 bytes)       <- probe fails, falls through, re-fetched upstream
cache object present ~3s later: True       <- tee repopulated the cache
request 3: 302 -> presigned target: 200    <- fast path healthy again
```

The existence probe converts a permanently-dead redirect into one slow request followed by a fully healed cache entry.

brandonrc:
Reviewed this across five independent passes — security, reliability, blast radius, Rust correctness, and test quality — against a clean Postgres with all 178 migrations. I've pushed a follow-up commit (2a75efe) rather than sending you round again. The short version: **the existence probe is good and I kept it; the reject-arm delete removal I've reverted**, for a reason I don't think was visible from where you were standing.

**On the deletes.** Your diagnosis of the intermediate `exists()` guard is exactly right — `put_stream` has already committed by the time the reject arm runs, so `exists()` is always true and the guard never fires. Where I ended up differing is the conclusion: that makes the *guard* redundant, not the *delete*.

The removal rests on "the missing sidecar alone forces the refetch." That's true for `CacheStore::get`, which does gate on the sidecar. But three readers reach the content key without consulting it:

- `handle_streaming_leader_upstream_error` streams `get_stream(cache_key)` as an RFC 5861 stale-if-error response and loads the sidecar *afterwards*, best-effort, only for headers. So a truncated body left at the key gets served on the next transient upstream failure — and that's a correlated sequence, not a coincidence: the same flaky upstream causes both the truncation and the 502 that triggers stale-if-error.
- The PyPI Remote arm streams `artifacts.storage_key` directly.
- `is_fresh` degrades to a bare `exists()` when either ETag is multipart-shaped (#2120) or a legacy sidecar has no pinned ETag — after which the presigned paths sign a 302 straight at the object with no checksum check at all. Multipart means large objects, which are exactly the ones that truncate.

Pre-fix, the delete made all three unreachable: `get_stream` returned `NotFound` and the code fell through to the error. It was load-bearing rather than belt-and-braces, which I don't think reads from the diff.

The versioned-bucket rationale also doesn't hold up, and this is the part I'd most want you to check my work on: `grep -rn "version_id\|versionId\|version-id" backend/src/storage/` returns nothing. The storage layer has no version concept and never reads a non-current version, so a delete marker can't mask content — that content was already unreachable. The marker simply reads as `NotFound`, which *is* the self-heal. Marker accumulation is a real operational cost, but it's a lifecycle-policy problem, and strictly smaller than serving corrupt bytes.

Worth noting the module's own contract still said so: `StreamWriteOutcome`'s doc reads "A rejected object is deleted (and its metadata sidecar skipped) so the next request refetches upstream and self-heals." And `write_negative_cache`'s 404 branch deletes the body for precisely this reason, which left the transient-error branch one line below as the odd one out.

**What I kept, because both are right:**

- The existence probe. It removes the dead-redirect case, and your two-direction test with the `presigned_calls == 1` positive control is genuinely well built — I revert-proved it and it fails correctly.
- The `write_negative_cache` guard. That path never writes content first, so `exists()` really does mean "a prior positive body is there to invalidate." The #1611 marker stacking is real and this is the right place to fix it.

**Also in the commit:**

- The `exists()` `Err` arm now logs instead of folding into "absent". `Err` means "unknown", not "gone", and an object-store brownout would otherwise silently convert every redirect on this path into a fully-buffered read through the backend with no signal that it happened. It still falls through on `Err` — never sign a redirect we couldn't verify.
- Corrected two comments. The fallthrough doesn't return `NotFound`: it routes through `coordinated_retry_get`, which takes a cluster-wide advisory lease (#1609) and returns 507 — the self-heal actually comes from the caller discarding that error. And an existence probe isn't the metadata-only freshness check the siblings apply; they also check TTL, revalidate the ETag, and apply `cache_quarantine_gate` for the #2075 hold. That gap is pre-existing, but the comment claimed parity, so I've stated it plainly instead. Filed separately.
- A positive control for the negative-cache guard — with a prior body present the delete must still fire. Without it, replacing the guard with `if false` passed the entire suite. That's the same critique you correctly applied to your own `local_fetch_or_redirect` test.
- The existence test now uses the real content-key shape (`proxy-cache/<repo>/<path>/__content__`) rather than one that only satisfies the prefix check.

**One correction to the PR body:** the audit claim ("audited every other storage-delete call site... everything else is already safely guarded") doesn't hold — `pypi.rs`'s `tee_refetch_to_storage` deletes a proxy-cache key unconditionally after its own `put_stream` commits, same shape as the arms this PR changed, and its test explicitly asserts "a truncated write-back must be deleted, not served warm". It's benign there (that path only runs after `get_stream` already returned `NotFound`, so there's nothing to mask), but the two policies would have contradicted each other.

Full `cargo test --lib` is green (14555 passed) and `cargo fmt --check` is clean. Each of the three guards is revert-proved: disable it and the corresponding test fails.

PR #3077, opened 2026-08-02
fix(pypi): fall back when advertised metadata is unavailable

### LINKED ISSUE #3076: PyPI proxy emits missing or incorrect PEP 658 metadata hashes
## Summary

The PyPI proxy does not preserve or correctly generate PEP 658 metadata hashes in the Simple API.

## Reproduction

1. Proxy a PyPI remote repository backed by https://pypi.org.
2. Request `/simple/dbt-sqlserver/`.
3. Compare the proxy response with PyPI for `dbt_sqlserver-1.12.0rc2-py3-none-any.whl`.
4. Resolve with uv 0.12.1 using `--index-strategy first-index --prerelease allow`.

Observed proxy behavior:

- The wheel URL keeps the artifact `#sha256` hash.
- The anchor omits `data-dist-info-metadata` / `data-core-metadata`.
- The corresponding `.whl.metadata` endpoint returns 404.
- uv reports that `dbt-sqlserver==1.12.0rc2` has no compatible version, although the wheel is listed and directly downloadable.

## Expected behavior

- Preserve upstream PEP 658 metadata hashes and metadata endpoints, or generate the metadata hash from the actual `.metadata` response.
- Never use the artifact/wheel SHA256 as `data-dist-info-metadata`.
- Add a regression test that validates the advertised metadata hash against the served `.metadata` bytes and verifies uv first-index resolution.

The current `generate_simple_package_index` implementation appears to map a generic artifact hash directly to `data-dist-info-metadata`, and its test only checks that the attribute exists, not that the value is the metadata-file hash.

axellpadilla:
## Summary

Fix remote PyPI PEP 658 metadata requests when an upstream advertises `.whl.metadata` but returns 404 for that resource.

Remote metadata requests now:

- Fetch upstream `.whl.metadata` when available.
- Fall back to extracting `METADATA` from the upstream wheel on 404.
- Continue propagating non-404 upstream failures.
- Preserve the existing local repository metadata extraction path.

This prevents pip from receiving a hard metadata 404 while retaining the PEP 658 metadata benefits.

Closes #3076

The regression test covers an upstream-only package where PEP 658 metadata is advertised, `.whl.metadata` returns 404, and the wheel remains available for fallback extraction.

## Test Checklist

- [x] Manually tested locally
- [x] No regressions in existing focused tests

Focused result:

```text
2 passed; 0 failed
```

Commands used:

```text
cargo fmt --check
git diff --check
cargo test -p artifact-keeper-backend --test pypi_proxy_upload_time_tests -- --ignored
```

## API Changes

- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -233,8 +233,20 @@ pub(crate) fn version_from_pypi_filename(filename: &str) -> Option<String> {
     None
 }
 
+/// The distribution a request ultimately refers to, with any PEP 658
+/// `.metadata` suffix removed.
+///
+/// ALL trailing `.metadata` suffixes are stripped, not just one. The serve path
+/// resolves a metadata request back to a distribution the same way, and the
+/// curation gate must evaluate the distribution that will actually be served:
+/// when the two disagree, a request naming the same wheel through a different
+/// suffix shape (`…-1.0-py3-none-any.whl.metadata.metadata`) is version-parsed
+/// as `None` — which skips version-constrained rules — while still resolving to
+/// the blocked wheel, serving a blocked version's metadata. Both callers in
+/// `download_or_metadata` share one call to this function so they cannot drift
+/// apart again.
 fn pypi_distribution_filename(filename: &str) -> &str {
-    filename.strip_suffix(".metadata").unwrap_or(filename)
+    filename.trim_end_matches(".metadata")
 }
 
 /// Curation gate for PyPI proxy requests (#2912). Returns
@@ -1538,8 +1550,14 @@ async fn download_or_metadata(
     // either branch below. The version is parsed from the distribution filename so
     // exact- and range-constrained rules apply on this path; an unrecognized
     // filename shape yields `None`, which matches on name only.
+    //
+    // `distribution` is resolved ONCE and reused by the metadata branch below:
+    // the gate must evaluate exactly the distribution the serve path resolves,
+    // or a suffix shape that parses to a different (or no) version slips past a
+    // version-constrained rule and is then served anyway.
     let normalized = normalize_pep503(&project);
-    let requested_version = version_from_pypi_filename(pypi_distribution_filename(&filename));
+    let distribution = pypi_distribution_filename(&filename);
+    let requested_version = version_from_pypi_filename(distribution);
     enforce_pypi_curation(&state, &repo, &normalized, requested_version.as_deref()).await?;
 
     // PEP 658: serve metadata from the remote upstream when possible. If the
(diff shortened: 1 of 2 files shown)


axellpadilla:
## What the final fix does

PEP 658/714 metadata signals are advertisements in the Simple API response. They are not the embedded wheel `dist-info/METADATA` file. The proxy now preserves those signals and makes the advertised `.metadata` request usable:

```mermaid
sequenceDiagram
    participant Client as pip or uv
    participant AK as Artifact Keeper
    participant Index as Upstream Simple API
    participant CDN as Upstream file host

    Client->>AK: Request Simple index
    AK->>Index: Fetch Simple index
    Index-->>AK: Wheel link + PEP 658/714 metadata signal
    AK-->>Client: Rewritten wheel link, metadata signal preserved

    Client->>AK: GET package.whl.metadata
    AK->>Index: Resolve wheel URL from Simple index
    Index-->>AK: Actual wheel URL
    AK->>CDN: GET package.whl.metadata

    alt Upstream metadata exists
        CDN-->>AK: 200 METADATA
        AK-->>Client: 200 METADATA
    else Upstream metadata returns 404
        CDN-->>AK: 404
        AK->>CDN: GET package.whl
        CDN-->>AK: 200 wheel
        AK->>AK: Extract embedded METADATA
        AK-->>Client: 200 extracted METADATA
    end
```

The resulting invariant is: **if the rewritten index advertises PEP 658/714 metadata, the proxy either returns the upstream `.metadata` resource or extracts equivalent `METADATA` from the upstream wheel.** Only an upstream 404 triggers the wheel fallback; authentication, network, and server errors are propagated instead of hidden.

axellpadilla:
@brandonrc  This fixes the issue but I'm not sure is the correct way and edges, for example, I believe there was a streaming pull already and this would use buffered to fix the inconsistence (and I believe prepare the cache?), left editable.

brandonrc:
Thanks for this — the PEP 658 fallback approach is the right call. I pushed two commits (`maintainerCanModify`) to green the coverage gate and to close two issues found while hardening the metadata path.

**1. Coverage gate (the failing check).** The new-code gate ran at 35% because the coverage job runs `cargo nextest --workspace --lib`, and `--lib` never compiles `backend/tests/*.rs` — so the `#[ignore]`d integration test couldn't move coverage regardless of the `--ignored` flag. Moved the coverage-critical tests into in-crate `#[cfg(test)]` tests in `pypi.rs` (using the existing `test_db_helpers` harness + wiremock, guarded by a runtime DB-skip so they run in CI and skip gracefully without a DB): a 404→wheel-extract test and a **new 200-direct** test that covers the `Ok` arm the original test never hit. Removed the now-dead ignored copy. New-code coverage is now **95.5%**.

**2. Curation bypass via stacked `.metadata` suffix (the one real bug).** With a version-constrained block rule, `pkg-1.0-...whl.metadata` was correctly blocked, but `pkg-1.0-...whl.metadata.metadata` returned **200** and served the blocked version's METADATA. Root cause was two different strippers: the curation gate used `strip_suffix(".metadata")` (one) so `version_from_pypi_filename` returned `None` and the version rule was skipped, while the serve branch used `trim_end_matches(".metadata")` (all) and resolved back to the real wheel. Fixed by resolving the distribution **once** and handing that same value to both the curation gate and the serve branch (they can no longer drift), plus a guard that a stacked-suffix request never resolves to a wheel. Blocked versions now 403 on every `.metadata` shape; name-only rules and legit non-blocked versions are unaffected.

**3. Content-type on the direct arm (hardening).** The 200-direct arm forwarded the upstream `Content-Type` verbatim; a malicious upstream serving `text/html` would render under this origin (blunted by the global CSP + nosniff today, but fragile). PEP 658 metadata is always the METADATA text file, so the response is now pinned to `text/plain; charset=utf-8` on both arms.

All three have red→green tests (reverting the prod edits reproduces the exact prior symptom). `fmt`/`clippy -D warnings`/full `--lib` suite green locally.

Separately — out of scope here, worth its own issue: `CurationService::version_matches` strips only a single leading `=`, so a PEP 440 `==1.0` rule leaves target `=1.0` and matches nothing (fails open). A `=1.0` rule works; `==1.0` silently never fires.

PR #3088, opened 2026-08-03
feat(stats): report proxy-cache storage and artifact count in system stats

### LINKED ISSUE #3087: System stats exclude proxy-cached storage and artifact counts
## Problem Statement
`GET /api/v1/admin/stats` derives `total_artifacts` and `total_storage_bytes` from the `artifacts` table only. Proxy-cached objects have no `artifacts` row, so they never appear in the stats even though they consume storage.

## Proposed Solution
Add `proxy_artifact_count` and `proxy_storage_bytes` to `SystemStats`, aggregated from `proxy_cache_artifacts`, kept separate from the existing hosted totals so the numbers stay additive.

## Component
Admin / Statistics

## Pre-submission Checklist
- [x] I have searched existing issues to make sure this is not a duplicate.
- [x] I have reviewed the documentation to confirm this feature does not already exist.

nicola-preda:
## Summary
Adds `proxy_artifact_count` and `proxy_storage_bytes` to `SystemStats`, summed from `proxy_cache_artifacts`. The existing totals stay hosted-only, so the two fields are purely additive. Lets the dashboard show a local + remote breakdown.

Fixes #3087

## Test Checklist
- [x] Manually tested locally

## API Changes
- [ ] N/A - no API changes

### DIFF
--- a/backend/.sqlx/query-2b47a8130e7d52549c75c6692fb03ef094dd204f44c19be4c86ad3d8ead44d59.json
+++ b/backend/.sqlx/query-2b47a8130e7d52549c75c6692fb03ef094dd204f44c19be4c86ad3d8ead44d59.json
@@ -1,0 +1,26 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT\n            COUNT(*) as \"count!\",\n            COALESCE(SUM(size_bytes), 0)::BIGINT as \"size!\"\n        FROM proxy_cache_artifacts\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "count!",
+        "type_info": "Int8"
+      },
+      {
+        "ordinal": 1,
+        "name": "size!",
+        "type_info": "Int8"
+      }
+    ],
+    "parameters": {
+      "Left": []
+    },
+    "nullable": [
+      null,
+      null
+    ]
+  },
+  "hash": "2b47a8130e7d52549c75c6692fb03ef094dd204f44c19be4c86ad3d8ead44d59"
+}
--- a/backend/src/api/handlers/admin.rs
+++ b/backend/src/api/handlers/admin.rs
@@ -789,8 +789,8 @@ pub struct SystemStats {
     pub total_users: i64,
     pub active_peers: i64,
     pub pending_sync_tasks: i64,
-    /// Proxy-cached (pull-through) objects. They have no `artifacts` row, so
-    /// they're excluded from the totals above and reported separately here.
+    /// Proxy-cached (pull-through) objects, tracked in `proxy_cache_artifacts`
+    /// rather than `artifacts`, so they are reported separately here.
     pub proxy_artifact_count: i64,
     pub proxy_storage_bytes: i64,
 }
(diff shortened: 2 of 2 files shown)


brandonrc:
Nice PR — thank you! 🙌 Clean, genuinely additive change: the existing totals stay hosted-only, the new fields carry `ToSchema`, and you committed the `.sqlx` offline metadata for the new query. That last one is the single most common thing that trips contributors up here (CI builds with `SQLX_OFFLINE=true`, so a `query!` macro without its metadata fails to compile in a way that's invisible locally) — nice catch.

Both red checks turned out to be unrelated to your logic, so I pushed one commit (`3c48c301`) to sort them out. Hope you don't mind the maintainer edit.

**Clippy — not your fault, an infra flake.** It wasn't a lint:

```
error: failed to select a version for the requirement `regex = "^1.11"` (locked to 1.12.4)
candidate versions found which didn't match: 1.8.4, 1.8.3, 1.8.2, ...
```

That's a stale crates.io index on the runner. Your PR touches neither `Cargo.toml` nor `Cargo.lock`, `main` is green with the same `regex = "1.11"` / lock `1.12.4`, and Backend Unit Tests passed on the very same commit. It cleared on a re-run.

**Coverage — the real one.** The gate flagged 23% (4/17 new lines): the serialization test covers the struct, but nothing exercised the handler or the new aggregate. I added two in-crate tests, now at **94% (82/87)**:

- one that seeds `proxy_cache_artifacts` rows and calls `get_system_stats`, asserting the **delta** (+2 count, +10,000 bytes) rather than absolute totals — the endpoint aggregates the whole table and `nextest` shares one database across processes, so absolute equality would flake while a genuine wiring bug still fails deterministically;
- one for the empty-table branch, proving `COALESCE(SUM(size_bytes), 0)` can't hand a NULL to the non-nullable `"size!"` decode, run inside a rolled-back transaction so it can't disturb concurrent tests.

Worth knowing for next time: the coverage job runs `cargo nextest run --workspace --lib`, and `--lib` never compiles `backend/tests/*.rs` — so a test placed in an integration target contributes exactly zero coverage. New-code coverage has to come from in-crate `#[cfg(test)]` tests.

I also softened one line of your doc comment. It said proxy-cached objects "have no `artifacts` row, so they're excluded from the totals above" — that's true for anything written since #1280, but legacy `storage_key LIKE 'proxy-cache/%'` rows do still exist in `artifacts`, and the existing `total_storage_bytes` query (`WHERE is_deleted = false`) doesn't filter them out. It now just says they're tracked in `proxy_cache_artifacts` rather than `artifacts`. No behaviour change.

One thing I deliberately did **not** change: your `COUNT(*)`/`SUM` over `proxy_cache_artifacts` is a full-table aggregate, but so are the six aggregates already in this handler, so you're being consistent with the surrounding code. Separately, the trigger-maintained `repository_usage_ledger` (migration 182) already carries per-repository `hosted_bytes`/`proxy_bytes`/`oci_bytes`, so this whole endpoint could become O(repositories) instead of O(rows) — that's the same win #3079 just landed for the repository listing. I'll file that as its own issue rather than push it into your PR.

All checks are green now. Thanks again — this is a useful addition, and the local/remote split on the dashboard will be genuinely nice to have. 🚀

PR #3113, opened 2026-08-04
fix(logging): silence k8s probe request logs, drop CA-load line to debug

### LINKED ISSUE #3112: Probe-endpoint request logs bury real request logs; custom-CA load logged per client build
Kubernetes and Prometheus poll the probe endpoints (`/health`, `/healthz`, `/ready`, `/readyz`, `/livez`, `/metrics`) on a tight loop forever. Each hit emits a per-request `Request completed` info line from the correlation-id middleware, so real request logs get buried under probe noise.

Separately, `apply_custom_ca_cert` logs `Loaded custom CA certificate(s)` at info level. That runs on every HTTP client build (many per request, e.g. `/health` 's Trivy sub-check), not once at startup, so it floods the logs too.

### Expected
- Probe-endpoint completion logs suppressed by default, with an opt-back-in (`LOG_PROBE_REQUESTS=true`) for operators who want them.
- Custom-CA load line demoted to debug so it does not fire at info on every client build.

nicola-preda:
Fixes #3112

## Summary
Kubernetes and Prometheus poll the probe endpoints (`/health`, `/healthz`, `/ready`, `/readyz`, `/livez`, `/metrics`) continuously, and each hit emitted a per-request `Request completed` info line that buried real request logs. The correlation-id middleware now skips the completion log for probe paths; set `LOG_PROBE_REQUESTS=true` to restore it. Same probe set the setup/guest-access middlewares already treat as unauthenticated.

Also demotes the `Loaded custom CA certificate(s)` line in `apply_custom_ca_cert` from info to debug: it fires on every HTTP client build (many per request), not once at startup, so at info it floods the logs.

`probe_paths_match_routes_and_nothing_else` asserts every probe path is suppressed and that real paths (`/api/v1/packages`, `/health/dashboard`, `/`, `/readyz/x`) are not.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -531,6 +531,14 @@ JWT_EXPIRATION_SECS=86400
 # LOG_FORMAT=pretty
 
 # -----------------------------------------------------------------------------
+# Log the per-request "Request completed" line for the health/readiness/liveness
+# probe endpoints (/health, /healthz, /ready, /readyz, /livez). Off by default
+# so Kubernetes probe polling does not bury real request logs; a non-2xx probe
+# response is logged regardless. Set to 1/true/yes/on to restore probe request
+# logging (e.g. mid-incident). Read once at first request.
+# LOG_PROBE_REQUESTS=false
+
+# -----------------------------------------------------------------------------
 # Structured audit-event stream (backend, #2413)
 # -----------------------------------------------------------------------------
 # Emit every recorded audit action as a self-contained NDJSON record so it can
--- a/backend/src/api/middleware/tracing.rs
+++ b/backend/src/api/middleware/tracing.rs
@@ -155,6 +155,11 @@ pub async fn correlation_id_middleware(mut request: Request, next: Next) -> Resp
 
     tracing::Span::current().record("correlation_id", tracing::field::display(&correlation_id));
 
+    // K8s/Prometheus poll the probe endpoints on a tight loop forever; their
+    // per-request "Request completed" lines bury real request logs. Skip them
+    // unless the operator opts back in via LOG_PROBE_REQUESTS (see #544).
+    let log_completed = !is_probe_path(request.uri().path()) || log_probe_requests();
+
     // Scope the task-local around the whole downstream future so audit
     // emitters anywhere under this request observe the same correlation ID
     // the span carries and the response header echoes (#2414).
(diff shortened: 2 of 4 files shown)


brandonrc (commented):
Thanks for this — the probe-log noise is a real problem and the core change is right. I reviewed against `97ecbbaa` and verified the surrounding behavior rather than just the diff. No security concerns: I checked whether the suppression could be evaded or abused, and it's closed in both directions — trailing-slash, double-slash and percent-encoded variants (`/health/`, `//health`, `/%68ealth`) all fail `is_probe_path` *and* fail routing, and `/health?x=1` is correctly suppressed since the query string isn't part of `path()`. Correlation-ID propagation and audit logging are untouched.

One substantive item, then some smaller ones.

**1. `/metrics` never matches, so the Prometheus half of this doesn't take effect.**

The reachable metrics endpoint on the main router is `/api/v1/admin/metrics` (`routes.rs:724`, via `nest("/api/v1")` → `nest("/admin")`), and the middleware sits outside every `nest()` so it sees the full unstripped path. The other `/metrics` — the `METRICS_PORT` listener at `main.rs:1161` — is a bare `Router::new()` with no correlation middleware, so it never emitted `Request completed` to begin with. Net effect: no request that reaches this middleware can ever match the `"/metrics"` arm, and a Prometheus scrape of `/api/v1/admin/metrics` still logs on every poll.

Worth fixing the two doc claims alongside it: `/metrics` isn't a route in `api::routes`, and the "same set the setup/guest-access middlewares treat as unauthenticated" line only holds for `setup.rs:41` — `guest_access.rs:77-88` doesn't include `/metrics`.

Either drop the entry, or match `/api/v1/admin/metrics` if silencing scrape logs is actually wanted.

**2. 5xx probe responses lose their only INFO record.**

When `/readyz` starts 503-ing because Postgres is gone, the line that carried `status=503` is now suppressed. The signal does survive, but only via `TraceLayer`'s default `ServerErrorsAsFailures` emitting at ERROR — an implicit dependency that a future custom `on_failure` or an `EnvFilter` tweak would quietly break. Suggest making it explicit:

```rust
if log_completed || !response.status().is_success() {
```

**3. For the CA line, consider `Once` at info rather than demoting to debug.**

`log_proxy_env()` right above it already solves exactly this ("log once at startup, not per build") with `static LOG_ONCE: Once`, and it's called from the same builder entry points. Demoting loses the positive confirmation of `path` + `count` at the level operators actually run — and the silent case is a valid-but-empty bundle, where `count = 0` produces no warn and now no info either. Both failure arms correctly stay at `warn!`, so a missing or corrupt file is still caught; it's the "loaded nothing, said nothing" case that goes dark.

Two notes on the rationale: the demotion is a no-op under the crate default filter (`telemetry.rs` sets `artifact_keeper_backend=debug`) and only bites in the shipped images that set `RUST_LOG=info`. And whatever wrapper you use, applying it to the whole function would also stop the `warn!` arms from repeating per client build, which is the louder flood on a misconfigured path.

**4. `LOG_PROBE_REQUESTS` should go in `.env.example`.**

It's the escape hatch someone reaches for mid-incident, so discoverability matters. `.env.example` already documents `CUSTOM_CA_CERT_PATH` (line 502) and `LOG_LEVEL`/`LOG_FORMAT`, so it slots in next to those. Minor related nit: only `1|true|yes` are accepted, so `LOG_PROBE_REQUESTS=on` silently means off, and the `OnceLock` initializes on first request rather than at startup so a typo is never surfaced. `sync_worker.rs` has an `env_bool` helper if you'd rather reuse it.

**5. Two test suggestions.**

`log_probe_requests()` isn't executed by any test — the `||` short-circuit means it's only reached on a probe path, and the added test covers `is_probe_path` alone. That's a possible brush with the 70% new-code coverage gate (though it may fall under the gate's <10-instrumented-line skip). Extracting the parse into a pure helper makes it directly testable without env or `OnceLock` state.

Also worth adding `assert!(!is_probe_path("/api/v1/admin/metrics"))` — that's the assertion that fails loudly if anyone ever moves this layer inside a `nest()`, where the prefix would be stripped to `/metrics` and the admin metrics endpoint would start being silenced.

---

Unrelated to this PR, but it surfaced while tracing the probe paths, so I filed it separately as #3127 rather than loading it onto you here: `/health` is unauthenticated, sits outside `/api/v1` so none of the `rate_limit_*` layers apply, and per request runs an uncached Trivy probe, an uncached OpenSearch `cluster_health()`, and an uncached `LdapService::new(...).check_health()` — a fresh directory bind on every anonymous request. Only the storage probe is TTL-cached. Nothing for you to do on this PR.

nicola-preda:
@brandonrc thanks for the thorough review, all addressed in the follow-up commit:

1. Dropped /metrics entirely (confirmed it never floods these logs; Prometheus scrapes the separate METRICS_PORT listener). Doc comment corrected; added the !is_probe_path("/api/v1/admin/metrics") guard-assertion.
2. Non-success probe responses now always log via should_log_completed.
3. Reverted the debug demotion; the CA line logs once at info via Once (warns included), and count==0 stays visible.
4. LOG_PROBE_REQUESTS documented in .env.example, now accepts 1/true/yes/on, and an unrecognized value warns at startup (init_probe_request_logging) instead of silently meaning off.
5. Extracted pure parse_log_probe_requests/resolve_log_probe_requests with unit tests (also clears the new-code coverage gate).

PR #3138, opened 2026-08-06
fix(security): evaluate scan gates across all scanners, not the newest row

### LINKED ISSUE #3140: Scan gates evaluate a single scan_results row and fail open with multiple scanners enabled
## Summary

Three security gates derive an artifact's verdict from a **single** `scan_results` row. Once more than one scanner is enabled, each one fails **open** rather than closed.

## 1. Download gate: a newer non-completed row switches severity blocking off

`policy_service::evaluate_artifact` selects the newest `scan_results` row of **any** status, then runs the `max_severity` check only when that row's status is `completed`.

Any scanner that records a `pending` / `running` / `not_applicable` row *after* a completed one therefore disables the severity gate for that artifact entirely, and the artifact is served **200 with critical findings on record**.

`block_unscanned` does not compensate: `classify_scan_state` is any-completed-wins, so an earlier completed scan satisfies it.

## 2 & 3. Promotion gates: the last engine to finish supplies the whole CVE verdict

`promotion_policy_service::LATEST_SCAN_COUNTS_SQL`, `OPEN_CVES_SQL` and `promotion_rule_service::get_latest_scan` each take the latest completed row **regardless of `scan_type`**. With several engines enabled, whichever finished last supplies the entire verdict for `max_cve_severity` and the promotion CVE summary; every other engine's findings are invisible to the gate.

## Reproduction

Against a clean database with all migrations applied, seed one artifact with:

- an **older** `grype` scan, status `completed`, 1 critical / 2 high, carrying `CVE-2026-0001`
- a **newer** `dependency` scan, status `completed`, clean
- a **newest** `image` scan, status `pending`

Observed:

| Gate input | Result |
|---|---|
| Promotion severity counts | `0` critical — the clean newer row wins |
| Promotion open CVEs | 0 rows — the critical CVE is invisible |
| Download gate | newest row is `pending` ⇒ severity check skipped ⇒ served 200 |

All three should reflect the critical finding that is on record for the artifact.

## Impact

Any deployment with two or more scan engines enabled. An artifact with critical findings can be downloaded and promoted as though it were clean. The gates fail open, so there is no error or log to indicate the check was skipped.

## Expected

- The severity gate keys on whether a completed scan **exists** for the artifact, not on whether the newest row happens to be completed.
- The promotion queries window **per `scan_type`** so every engine's latest completed scan contributes to the verdict.
- An artifact that has never been scanned must remain distinguishable from one that was scanned and found clean.

YoussefEG-Eurostar:
## Summary

Three security gates decide an artifact's verdict from a **single** `scan_results` row. That is wrong as soon as more than one scanner is enabled, and each case fails **open** rather than closed.

I found these while designing an additional (asynchronous, external) scanner backend against this codebase — an async engine makes all three fire routinely, but none of them require one to reproduce: any deployment with two engines enabled is already affected.

## 1. Download gate: a newer non-completed row switches severity blocking off

`policy_service::evaluate_artifact` selected the newest row of **any** status:

```sql
SELECT status, ... FROM scan_results WHERE artifact_id = $1 ORDER BY created_at DESC LIMIT 1
```

and then ran the `max_severity` check only `if scan.status == "completed"`.

So any scanner that records a `pending` / `running` / `not_applicable` row *after* a completed one disables the severity gate for that artifact entirely — `enforce_download_gate` returns allowed and the artifact is served **200 with critical findings on record**. `block_unscanned` does not compensate: `classify_scan_state` is any-completed-wins, so an earlier completed scan satisfies it.

Fix: read the two gate inputs separately, since they are answering different questions.

* `block_on_fail` — "did the most recent attempt fail?" → still keys on the newest row's status.
* `max_severity` — "what is on record for this artifact?" → keys on whether a completed scan **exists**.

Both come from one round trip, and the decision is a pure, unit-tested function (`post_scan_gates`).

**Intended behaviour change:** a `failed` newest row with `block_on_fail` *off* now still evaluates severity against the older completed scan's findings, where previously the check was skipped. A crashed rescan must not clear an artifact's finding history.

## 2 & 3. Promotion gates: the last engine to finish supplies the whole CVE verdict

`promotion_policy_service::LATEST_SCAN_COUNTS_SQL`, `OPEN_CVES_SQL` and `promotion_rule_service::get_latest_scan` all took the latest completed row **regardless of `scan_type`**. With several engines enabled, whichever finished last supplies the entire verdict for `max_cve_severity` / the promotion CVE summary, and every other engine's findings are invisible to the gate.

Fix: window per `scan_type` with `DISTINCT ON`, matching what `recalculate_score` already does (`scan_result_service.rs`).

Two deliberate details:

* **`MAX`, not `SUM`.** The per-scanner count columns describe *overlapping* finding sets, so summing double-counts a CVE that two engines both report. `MAX` cannot weaken the gate — `evaluate_cve_thresholds` only tests "more than allowed", and a severity is non-zero here whenever any engine reported it.
* **`HAVING COUNT(*) > 0`.** An aggregate with no `GROUP BY` always returns one all-zeros row, which would turn the caller's `None` ("never scanned") into "scanned and clean". This keeps zero rows for an unscanned artifact.

## Verification

Applied all 170 migrations to a clean Postgres 16 and seeded the minimal reproducing fixture: one artifact, an **older** `grype` scan with 1 critical / 2 high carrying `CVE-2026-0001`, a **newer, clean** scan from a second engine, then a newer `pending` row.

| Query | Before | After |
|---|---|---|
| Promotion severity counts | `0` critical (the clean newer row won) | **1 critical, 2 high** |
| Promotion open CVEs | 0 rows | **CVE-2026-0001** |
| Download-gate input | newest row `pending` ⇒ severity check skipped | `has_completed_scan = true` ⇒ gate armed |
| Artifact with no scans | — | 0 rows ⇒ `None` preserved |

New pure unit tests cover the gate-input decision (including each of `pending` / `running` / `not_applicable` as the newest row) and pin the SQL shape so a future `LIMIT 1` regression fails the build. `cargo fmt --check` clean; `cargo test --lib` passes (409 tests in the touched modules, 0 failures).

## Notes for reviewers

Happy to split this into three PRs if you'd prefer them separately — they're grouped because they're one class of bug with one root cause.

Three adjacent things I did **not** change, but which look like the same family, if you want issues opened for them:

* `artifact_service::prepare_download` (the generic REST download path) enforces quarantine but never calls `enforce_download_gate`, so scan policy is not applied there — unlike the ~25 per-format handlers.
* `rollup_scan_status` does not handle `not_applicable`, so it falls through to `has_unknown` and reports `partial` for any artifact with a not-applicable scanner row.
* `scan_configs.block_on_policy_violation` and `scan_configs.severity_threshold` have CRUD but no enforcement reader; `Severity::meets_threshold` / `ScanConfig::threshold()` are referenced only from tests.

Closes #3140

### DIFF
--- a/backend/src/services/policy_service.rs
+++ b/backend/src/services/policy_service.rs
@@ -26,18 +26,20 @@ fn block_unscanned_violated(block_unscanned: bool, scan_state: ScanState) -> boo
 struct PostScanGates {
     /// The newest scan row (any status) is `failed`, so `block_on_fail` may fire.
     block_on_fail_applies: bool,
-    /// A completed scan exists, so the `max_severity` threshold check must run.
+    /// Findings are on record for this artifact, so the `max_severity`
+    /// threshold check must run.
     severity_applies: bool,
 }
+
 /// Decide which post-scan gates apply from the artifact's scan rows.
 ///
 /// The two gates deliberately read DIFFERENT inputs:
 ///
 /// * `block_on_fail` is about the most recent attempt, so it keys on the newest
 ///   row's status.
-/// * `max_severity` is about the findings on record, so it keys on "a completed
-///   scan EXISTS for this artifact" — not on "the newest row happens to be
-///   completed".
+/// * `max_severity` is about the findings on record, so it keys on whether
+///   there is anything on record to grade — not on "the newest row happens to
+///   be completed".
 ///
 /// Both used to read the single newest row of any status, which made the
 /// severity gate silently inert whenever the newest row was not `completed`.
@@ -45,17 +47,32 @@ struct PostScanGates {
 /// `not_applicable` row AFTER a completed one — an asynchronous or external
 /// scanner does this routinely — therefore disabled severity blocking for that
 /// artifact entirely, and a download with critical findings on record was
-/// served with a 200. Keying on existence closes that fail-open.
+/// served with a 200. Keying on what is on record closes that fail-open.
 ///
 /// Note the consequence for a `failed` newest row with `block_on_fail` off:
 /// severity now still evaluates against the findings of the older completed
 /// scan, where it previously skipped the check. That is the intended, strictly
 /// safer direction — a crashed rescan must not clear an artifact's history.
+///
+/// `has_unacknowledged_findings` is the second half of that same argument.
+/// Findings are persisted BEFORE the scan row is flipped to `completed`
+/// (`scanner_service` calls `create_findings` and only then `complete_scan`),
+/// so a scanner that dies in between — or is later reaped from `running` to
+/// `failed` by the stuck-scan janitor — leaves unacknowledged findings on
+/// record with NO completed row anywhere. Keying the gate on `completed`
+/// alone would leave exactly the reported fail-open one layer down: an
+/// unacknowledged critical on record, served 200. Grading whenever there is
+/// something to grade cannot produce a false block, because the threshold
+/// query returns zero when no finding meets the policy's severity.
 /// Pure / unit-testable.
-fn post_scan_gates(latest_status: Option<&str>, has_completed_scan: bool) -> PostScanGates {
+fn post_scan_gates(
+    latest_status: Option<&str>,
+    has_completed_scan: bool,
+    has_unacknowledged_findings: bool,
+) -> PostScanGates {
     PostScanGates {
         block_on_fail_applies: latest_status == Some("failed"),
-        severity_applies: has_completed_scan,
+        severity_applies: has_completed_scan || has_unacknowledged_findings,
     }
 }
 
@@ -152,6 +169,7 @@ impl PolicyService {
         struct ScanGateRow {
             latest_status: Option<String>,
             has_completed_scan: bool,
+            has_unacknowledged_findings: bool,
         }
 
         let gate_row: ScanGateRow = sqlx::query_as(
@@ -167,7 +185,13 @@ impl PolicyService {
                       FROM scan_results
                      WHERE artifact_id = $1
                        AND status = 'completed'
-                ) AS has_completed_scan
+                ) AS has_completed_scan,
+                EXISTS (
+                    SELECT 1
+                      FROM scan_findings
+                     WHERE artifact_id = $1
+                       AND NOT is_acknowledged
+                ) AS has_unacknowledged_findings
             "#,
         )
         .bind(artifact_id)
@@ -178,6 +202,7 @@ impl PolicyService {
         let gates = post_scan_gates(
             gate_row.latest_status.as_deref(),
             gate_row.has_completed_scan,
+            gate_row.has_unacknowledged_findings,
         );
 
         // #1649: classify the artifact's overall scan state from ALL its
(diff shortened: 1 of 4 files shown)


brandonrc:
Reviewed this across five independent passes — security, SQL/schema, Rust correctness, blast radius, and test quality — and validated it against a clean Postgres 16 with all 170 migrations applied. Nice find, and the writeup made it fast to verify. I've pushed a follow-up commit rather than sending you round again; details below so nothing is a surprise.

**All three fail-opens reproduce and all three are closed.** Independently confirmed with your fixture shape (older `grype` completed carrying a critical, newer clean scan from a second engine, newest row `pending`):

| Gate input | Before | After |
|---|---|---|
| Download gate | newest `pending` ⇒ severity check skipped | `has_completed_scan` ⇒ gate armed |
| Promotion counts | 0 critical | 1 critical / 2 high |
| Open CVEs | 0 rows | the seeded CVE |

The details worth calling out as correct, because each is a trap: `HAVING COUNT(*) > 0` is genuinely load-bearing (removing it returns one all-zeros row, turning "never scanned" into "scanned and clean" — verified both ways); `MAX` really cannot weaken the gate, since every consumer thresholds at `> 0` or `> i32::MAX` and `findings_count` is display-only; and `ORDER BY scan_type, completed_at DESC NULLS LAST, created_at DESC` matches `idx_scan_results_latest_per_artifact_type` from migration 101 column-for-column, so the new form is index-only and measured at the same buffer count as the old `LIMIT 1`. `fetch_optional` → `fetch_one` is right, and these are runtime `query_as` calls so no `.sqlx` regeneration is needed.

**What I pushed (38201d5):**

1. **The same fail-open one layer down.** `severity_applies` keyed on "a completed scan exists", but `scanner_service` persists findings *before* flipping the row to `completed`. A scanner that dies in between — or gets reaped `running` → `failed` by the stuck-scan janitor — leaves unacknowledged findings on record with no completed row anywhere, and the gate skipped them. Confirmed live: `latest_status=failed, has_completed_scan=false`, and an unacknowledged critical sitting there. It now grades whenever there's something on record, which can't produce a false block because the threshold query already returns zero when nothing meets the policy severity.

2. **`ELSE` on the `max_severity` CASE.** No `ELSE` meant an unknown value yielded `unnest(NULL)` → `IN (<empty>)` → false, i.e. the gate passes an artifact it was asked to block. Unreachable today thanks to `scan_policies_max_severity_check`, so purely defence in depth — but it's a security gate and it's one line.

3. **Tests that fail when the fix is reverted.** This is the substantive one. The three pure `post_scan_gates` tests pin the helper, but the bug never lived in the helper — it lived in the wiring, so **all three pass with the fail-open fully restored** (I ran it). Your SQL-shape test does fail on revert, and it earns its place, but it can't catch a query that's syntactically valid and semantically wrong. I added DB-backed tests driving the real `enforce_download_gate` and `get_cve_summary`, each verified to fail on revert of the corresponding change — the CVE one catches the wrong *value* (`left: 0, right: 1`), not just the wrong string. Included a zero-findings negative control so it can't pass by blocking unconditionally. Also widened the shape pin to every count column; it pinned only `critical_count`, so leaving one un-aggregated would have silently narrowed the verdict back to one scanner.

**One thing for the release note, not a code change.** The PR frames this as "wrong as soon as more than one scanner is enabled," which undersells it. The shipped `docker-compose.yml` already enables four scanners, and OpenSCAP registers *last* and writes a terminal `not_applicable` row for anything that isn't a container/rpm/deb. So on a default deployment the newest row is permanently non-`completed` for every npm/pypi/maven/cargo/generic artifact, and the `max_severity` download gate has been inert for essentially all of them. This turns it back on for all of them at once — correct, but a step-change in enforcement rather than an edge case. It's bounded by the fact that no default `scan_policies` row is seeded, so anyone who never created a policy sees no change; the exposure is operators who opted in and have been silently unprotected.

**On your offer to split:** no need, the grouping is right — one root cause, one review.

**Adjacent items** — I verified all three of yours are real and am filing them separately so they don't ride along here. Adding two more found during review: `block_on_fail` still keys on a single global newest row, so a crashed scanner masked by a later clean one goes unnoticed (same class, but pre-existing — the old code behaved identically, so not a regression); and the download gate's severity count is unscoped to any scan while the promotion gate now windows per scanner, which will read as "download 403s but the dashboard looks clean".

Full `cargo test --lib` is green (14554 passed) and `cargo fmt --check` is clean.

PR #3151, opened 2026-08-07
fix(helm): reject OCI pushes to classic repositories

### LINKED ISSUE #3150: OCI pushes into classic Helm repositories generate an invalid index.yaml
## Description

A repository configured with `format=helm` accepts uploads through the OCI Distribution API under `/v2/...`, including Helm chart pushes made with `helm push oci://...`.

The resulting OCI manifest rows are then enumerated by the classic Helm endpoint at `/helm/{repo}/index.yaml`. This produces invalid chart names and versions, and a valid tagged chart cannot be resolved with classic `helm pull`.

The backend already distinguishes `helm` from `helm_oci`, so the classic Helm repository should not accept OCI writes.

## Environment Information

- Deployment type: Kubernetes
- Client operating system: macOS
- Client architecture: arm64
- Database: PostgreSQL
- Reverse proxy: HTTPS ingress (hostname redacted)
- Network setup: TLS endpoint

## Backend Version

`1.7.0`

## Web UI Version

N/A

## Package Format

Helm / OCI

## Expected Behavior

- OCI Distribution endpoints should only accept repositories backed by OCI-capable formats, including `helm_oci`.
- An OCI request targeting a classic `format=helm` repository should fail with a clear 4xx response before any blob, manifest, or artifact data is written.
- A classic Helm `index.yaml` should contain only valid chart-package entries and must not expose OCI manifest rows.
- Helm charts stored as OCI artifacts should be accessed with an `oci://` reference.

## Observed Behavior

An OCI Helm push to a classic `format=helm` repository succeeds. The generated classic index then contains entries shaped like:

```yaml
entries:
  keycloak:26.7.0:
    - name: keycloak:26.7.0
      version: 26.7.0
      urls:
        - /helm/helm-local/charts/26.7.0
  keycloak:sha256:<digest>:
    - name: keycloak:sha256:<digest>
      version: sha256:<digest>
```

Digest references are rejected by Helm as invalid chart versions. Tagged entries also use `keycloak:26.7.0` as the chart name instead of `keycloak`, so this fails even immediately after a successful repository update:

```console
$ helm repo update local
...Successfully got an update from the "local" chart repository

$ helm pull local/keycloak --version 26.7.0
Error: chart "keycloak" matching 26.7.0 not found in local index.
(try 'helm repo update'): no chart name found
```

The same chart remains accessible through OCI:

```console
$ helm show chart oci://registry.example.com/helm-local/keycloak --version 26.7.0
apiVersion: v2
name: keycloak
version: 26.7.0
```

## Steps to Reproduce

1. Create a local repository with `format=helm`.
2. Authenticate to the Artifact Keeper registry.
3. Push a packaged chart through the OCI endpoint:

   ```shell
   helm push keycloak-26.7.0.tgz oci://registry.example.com/helm-local
   ```

4. Add the same repository as a classic HTTP Helm repository:

   ```shell
   helm repo add local https://registry.example.com/helm/helm-local
   helm repo update local
   ```

5. Attempt to pull the tagged chart:

   ```shell
   helm pull local/keycloak --version 26.7.0
   ```

6. Inspect `GET /helm/helm-local/index.yaml` and observe that OCI tag/digest references were emitted as chart names and versions.

## Logs

```text
level=WARN msg="skipping loading invalid entry for chart \"keycloak:sha256:<digest>\" \"sha256:<digest>\" ... validation: chart.metadata.version \"sha256:<digest>\" is invalid"

Error: chart "keycloak" matching 26.7.0 not found in local index.
(try 'helm repo update'): no chart name found
```

Helm client: `v4.2.3`

## Current Source Analysis

Verified against current `main` at commit [676f12ca69e985e46d2725ee758ee94d1275dc03](https://github.com/artifact-keeper/artifact-keeper/commit/676f12ca69e985e46d2725ee758ee94d1275dc03):

1. OCI repository resolution selects a repository by key but does not select or validate its format:
   [`backend/src/api/handlers/oci_v2.rs#L2400-L2407`](https://github.com/artifact-keeper/artifact-keeper/blob/676f12ca69e985e46d2725ee758ee94d1275dc03/backend/src/api/handlers/oci_v2.rs#L2400-L2407)

2. OCI manifest artifact rows are written with `name = "{image}:{reference}"` and `version = reference`:
   [`backend/src/api/handlers/oci_v2.rs#L7921-L7940`](https://github.com/artifact-keeper/artifact-keeper/blob/676f12ca69e985e46d2725ee758ee94d1275dc03/backend/src/api/handlers/oci_v2.rs#L7921-L7940)

3. Classic Helm index generation selects every non-deleted, non-`.prov` artifact in the repository. If chart metadata is absent, it synthesizes `ChartYaml` from the artifact row's name and version:
   [`backend/src/api/handlers/helm.rs#L175-L214`](https://github.com/artifact-keeper/artifact-keeper/blob/676f12ca69e985e46d2725ee758ee94d1275dc03/backend/src/api/handlers/helm.rs#L175-L214)

4. The repository model already has distinct `Helm` and `HelmOci` variants:
   [`backend/src/models/repository.rs#L19-L33`](https://github.com/artifact-keeper/artifact-keeper/blob/676f12ca69e985e46d2725ee758ee94d1275dc03/backend/src/models/repository.rs#L19-L33)

## Suggested Fix

- Include the repository format in OCI repository resolution and reject repositories whose format does not map to the OCI handler.
- In particular, reject `format=helm` while accepting `format=helm_oci`.
- Defensively exclude OCI manifest rows (for example, OCI manifest paths/media types) from classic Helm index generation.
- Add regression tests covering wrong-format OCI pushes and index generation after an attempted OCI push.
- Consider a repair or migration path for repositories already containing OCI manifest rows under `format=helm`.

## Impact

- Every `helm repo add` or `helm repo update` emits repeated warnings.
- Charts cannot be resolved through the classic repository name/version interface.
- A successful OCI push leaves the repository in a protocol-inconsistent state.
- OCI implementation details and digest references leak into the classic Helm index.

## Additional Context

Existing issue #2941 concerns `cm-push` URL construction and appears unrelated to this format-validation/index-integrity problem.

## Pre-submission Checklist

- [x] I have redacted sensitive information (passwords, tokens, private URLs)

junsung-cho:
## Summary

- Reject OCI Distribution requests when the target repository format is not backed by the OCI handler, returning `400 UNSUPPORTED` before writes.
- Continue accepting `helm_oci` and the other OCI-backed repository formats.
- Exclude legacy OCI manifest rows from classic Helm `index.yaml` generation so already affected repositories produce a valid index without destructive data migration.
- Fixes #3150.

## Test Checklist

- [ ] Manually tested locally

Validated locally:

- `cargo fmt --check`
- `cargo test --workspace --lib` (14,722 passed; 0 failed; 6 ignored)

## API Changes

The existing OCI endpoints now reject classic `helm` repositories. No endpoints, request/response schemas, or migrations were added.

- [ ] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/helm.rs
+++ b/backend/src/api/handlers/helm.rs
@@ -94,6 +94,30 @@ fn is_prov_filename(filename: &str) -> bool {
     filename.ends_with(PROV_SUFFIX)
 }
 
+/// OCI manifest rows are not classic Helm chart packages and must never be
+/// rendered into `index.yaml`.
+///
+/// The path check repairs the read side for rows already written through the
+/// OCI Distribution API before repository-format validation was added (#3150).
+/// The media-type check is a second guard in case an imported manifest uses a
+/// non-standard path.
+fn is_oci_manifest_artifact(path: &str, content_type: &str) -> bool {
+    let media_type = content_type
+        .split(';')
+        .next()
+        .unwrap_or(content_type)
+        .trim();
+
+    (path.starts_with("v2/") && path.contains("/manifests/"))
+        || matches!(
+            media_type,
+            "application/vnd.oci.image.manifest.v1+json"
+                | "application/vnd.oci.image.index.v1+json"
+                | "application/vnd.docker.distribution.manifest.v2+json"
+                | "application/vnd.docker.distribution.manifest.list.v2+json"
+        )
+}
+
 /// Validate that an uploaded `prov` part really is a clearsigned provenance
 /// document before it is stored.
 ///
@@ -174,7 +198,7 @@ async fn query_charts_from_repo(
 ) -> Result<(), Response> {
     let rows = sqlx::query(
         r#"
-        SELECT a.id, a.name, a.version, a.path, a.size_bytes, a.checksum_sha256,
+        SELECT a.id, a.name, a.version, a.path, a.content_type, a.checksum_sha256,
                a.created_at,
                am.metadata
         FROM artifacts a
@@ -194,9 +218,14 @@ async fn query_charts_from_repo(
         let name: String = row.get("name");
         let version: Option<String> = row.get("version");
         let path: String = row.get("path");
+        let content_type: String = row.get("content_type");
         let checksum_sha256: String = row.get("checksum_sha256");
         let created_at: chrono::DateTime<chrono::Utc> = row.get("created_at");
         let metadata: Option<serde_json::Value> = row.get("metadata");
+
+        if is_oci_manifest_artifact(&path, &content_type) {
+            continue;
+        }
 
         let version = match version {
             Some(v) => v,
@@ -1312,6 +1341,31 @@ wsDcBAEBCgAQBQJqWW7VCRA8wAoTVPCkgwAAVAoMACmQbvnhlkWncOkVJXfissGD\n\
     }
 
     #[test]
+    fn test_oci_manifests_are_not_classic_helm_chart_artifacts() {
+        for content_type in [
+            "application/vnd.oci.image.manifest.v1+json",
+            "application/vnd.oci.image.index.v1+json; charset=utf-8",
+            "application/vnd.docker.distribution.manifest.v2+json",
+            "application/vnd.docker.distribution.manifest.list.v2+json",
+        ] {
+            assert!(is_oci_manifest_artifact("imported/manifest", content_type));
+        }
+
+        assert!(is_oci_manifest_artifact(
+            "v2/keycloak/manifests/26.7.0",
+            "application/json"
+        ));
+        assert!(!is_oci_manifest_artifact(
+            "keycloak/26.7.0/keycloak-26.7.0.tgz",
+            "application/gzip"
+        ));
+        assert!(!is_oci_manifest_artifact(
+            "keycloak-26.7.0.tgz",
+            "application/octet-stream"
+        ));
+    }
+
+    #[test]
     fn test_sha256_computation() {
         let mut hasher = Sha256::new();
         hasher.update(b"chart content");
@@ -1452,6 +1506,58 @@ wsDcBAEBCgAQBQJqWW7VCRA8wAoTVPCkgwAAVAoMACmQbvnhlkWncOkVJXfissGD\n\
             "the URL helm derives from index.yaml must serve the chart"
         );
         assert_eq!(&got[..], b"helm-chart-bytes");
+
+        f.teardown().await;
+    }
+
+    /// Regression guard for #3150: OCI manifest rows written into a classic
+    /// Helm repository by older versions must not leak into `index.yaml`.
+    #[tokio::test]
+    async fn test_helm_index_excludes_existing_oci_manifest_rows() {
+        let Some(f) = tdh::Fixture::setup("local", "helm").await else {
+            return;
+        };
+        let repo = f.repo_info("local", None);
+
+        tdh::seed_artifact(
+            &f.state,
+            &f.pool,
+            &repo,
+            "helm/keycloak/26.7.0/keycloak-26.7.0.tgz",
+            "keycloak/26.7.0/keycloak-26.7.0.tgz",
+            "keycloak",
+            "26.7.0",
+            "application/gzip",
+            bytes::Bytes::from_static(b"classic-chart"),
+            f.user_id,
+        )
+        .await;
+        tdh::seed_artifact(
+            &f.state,
+            &f.pool,
+            &repo,
+            "oci/manifests/abcdef",
+            "v2/keycloak/manifests/26.7.0",
+            "keycloak:26.7.0",
+            "26.7.0",
+            "application/vnd.oci.image.manifest.v1+json",
+            bytes::Bytes::from_static(br#"{"schemaVersion":2}"#),
+            f.user_id,
+        )
+        .await;
+
+        let app = f.router_anon(super::router());
+        let (status, body) = tdh::send(app, tdh::get(format!("/{}/index.yaml", f.repo_key))).await;
+        assert_eq!(
+            status,
+            StatusCode::OK,
+            "index generation failed: {}",
+            String::from_utf8_lossy(&body)
+        );
+        let index: HelmIndex = serde_yaml::from_slice(&body).expect("valid Helm index");
+        assert_eq!(index.entries.len(), 1);
+        assert!(index.entries.contains_key("keycloak"));
+        assert!(!index.entries.contains_key("keycloak:26.7.0"));
 
         f.teardown().await;
     }
--- a/backend/src/api/handlers/helm.rs
+++ b/backend/src/api/handlers/helm.rs
@@ -101,6 +101,21 @@ fn is_prov_filename(filename: &str) -> bool {
 /// OCI Distribution API before repository-format validation was added (#3150).
 /// The media-type check is a second guard in case an imported manifest uses a
 /// non-standard path.
+///
+/// This filter is LOAD-BEARING, not redundant with the `/v2` write gate. That
+/// gate only covers the Distribution API; `services/oci_referenced_content.rs`
+/// (migration import) and `services/peer_instance_service.rs` (federation
+/// replication) also write OCI manifest rows and apply NO repository-format
+/// check. Removing this filter would let those two paths put manifests back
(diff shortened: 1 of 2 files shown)


brandonrc:
Ran this through the pipeline against `4a02c296`. **Stage 1 reproduced, stage 2 passed, no blocker.** The fix is correct and the format gate is exactly right.

Reproduction on the base — one legitimate classic chart, then `PUT /v2/<helm-repo>/keycloak/manifests/26.7.0`, then `GET /helm/<repo>/index.yaml`:

```
OCI PUT status = 201 Created
index.yaml:
  keycloak:        { name: keycloak, version: 26.7.0 }
  keycloak:26.7.0: { name: "keycloak:26.7.0", urls: [/helm/<repo>/charts/26.7.0] }
```

That's #3150 verbatim. At head, with a fixture per format:

```
docker 201 · podman 201 · buildx 201 · oras 201 · wasm_oci 201 · helm_oci 201
helm 400 UNSUPPORTED (oci_tags=0)   generic 400 (oci_tags=0)
```

All six OCI-backed formats still push, helm is blocked with no rows and no storage object written. Reusing `format_handler_key` instead of hand-rolling a list is the right call — I diffed it against `get_handler_for_format`'s OCI arm and the sets are identical, not one format wide or narrow. Also confirmed `"oci"` is only a handler-table key, never a `repository_format` label, and that all 52 enum values round-trip through `parse_format_str`. Ordering is genuinely pre-write: the gate is in `resolve_repo`, before `StorageLocation` is built, and all twelve `/v2` handlers route through it.

Three findings, none blocking.

**The gate covers reads and deletes too, which is broader than the title suggests.** `resolve_repo` is on every path, so for the exact operators this PR is for — those who already pushed OCI into a classic helm repo — content that was pullable yesterday is now unreachable *and* undeletable through the OCI API:

```
GET    .../manifests/26.7.0 -> 400 UNSUPPORTED
DELETE .../manifests/26.7.0 -> 400 UNSUPPORTED
```

The `oci_tags` rows survive, so the repo still appears in `/v2/_catalog` while every pull against it 400s. There's a partial escape via the path-based `delete_artifact` REST route, but it doesn't cascade to `oci_tags`/`oci_manifest_refs`/`oci_blobs`. Your call whether to gate writes only and let reads drain, or keep it and add the repair path the issue itself contemplates — but it's worth stating in the PR description either way, since this is a behavior break landing in a patch line.

**The index exclusion predicate can drop a legitimate chart.** `is_oci_manifest_artifact` matches on path shape *or* content-type, and classic upload builds `artifact_path` straight from `Chart.yaml` with no SemVer validation. A chart named `v2` at version `manifests` yields `v2/manifests/v2-manifests.tgz`, which trips the path arm. Verified — seeded through the exact path and content-type the upload handler writes:

```
adversarial entries = ["ok"]     # the v2 chart is gone
```

Upload returns 201, the file downloads fine, but `helm repo update` never sees it. Contrived inputs, which is why this isn't a blocker — but an authoritative discriminator already exists and is nearly free: OCI manifest rows always carry `storage_key = "oci-manifests/" + digest`, classic charts carry `helm/<name>/<version>/<file>`. Anchoring on the storage-key prefix can't false-positive on a chart, and the constant is already pinned by `prefix_matches` guards. Keep the path/media-type checks as a secondary arm for imported rows if you want, but let the storage key carry the load.

**Minor:** the rejection message includes the repository format, and on read paths the gate runs before the visibility check — so `Authorization: Bearer anonymous` against a **private** repo now returns its `repository_format` where it previously returned a 401 challenge. Existence was already distinguishable, so the delta is just the format string. Dropping `format` from the message costs nothing.

**On the tests:** three of the six are real guards — `test_helm_index_excludes_existing_oci_manifest_rows`, `test_resolve_repo_rejects_classic_helm_repository`, and `manifest_put_rejects_classic_helm_repository_before_writes` all fail on a production-only revert, and together they cover both halves of the fix plus the pre-write property. That's the important part and it's satisfied. The other three are pure-predicate tests that pass with the call sites deleted — worth keeping (`..._accepts_oci_handler_formats` exercises production `format_handler_key`, so it would catch a gate that got too *narrow*), just not guards of the wiring. The gap: `helm_oci` is the only DB-backed positive control. I verified all six formats by hand above, but that verification doesn't live in the repo — turning that test into a loop over the six format strings is a few lines and locks the set in.

Gates: `cargo fmt --check` clean, `clippy --lib --all-features -D warnings` clean, `oci_v2` 520 passed, `helm` 54 passed, `formats::helm` 47 passed.

brandonrc:
Ran this through the review pipeline at `4a02c29`. **Bug verified at merge-base**, and correctly diagnosed — not a misdiagnosed cause.

I confirmed by reading the merge-base rather than trusting the issue: `resolve_repo`'s SELECT has no `format` column at all, and there is no format gate anywhere in the OCI data plane, so a `format=helm` repo is fully writable over `/v2`. `upsert_manifest_artifact` writes `name = {image}:{ref}`, and `query_charts_from_repo` synthesizes a `ChartYaml` from the row's name/version, which `generate_index_yaml` keys by `chart.name`. That composes to exactly the reported `entries: {keycloak:26.7.0: …, keycloak:sha256:<digest>: …}`.

Several things here are notably well done for a first pass: the gate sits at the single choke point (`resolve_repo`) rather than being sprinkled across 12 handlers; the accept set is *derived* from the existing `format_handler_key` instead of a fresh hardcoded list that would drift; both the write side and the already-corrupted read side are addressed; the no-write assertion queries `artifacts` and `oci_tags` directly instead of trusting the status code; and the SQL avoids the macros that would have forced a `cargo sqlx prepare` regeneration.

**One question is a product decision, and it's the maintainer's to make — I'm not going to decide it in review.**

The gate is in `resolve_repo`, which is also called by `handle_get_manifest`, `handle_get_blob`, `handle_head_manifest`, `handle_head_blob`, and `handle_tags_list`. So it blocks **reads**, not just pushes. A customer who already hit this issue — the issue explicitly notes "the same chart remains accessible through OCI" — upgrades and finds `helm pull oci://…` now returns `400 UNSUPPORTED`, while the same charts were simultaneously dropped from `index.yaml`. The bytes still exist and still count against quota, reachable by neither protocol, with no repair or export path. The PR title says "reject OCI *pushes*", and the issue asked to "consider a repair or migration path for repositories already containing OCI manifest rows."

Two defensible answers: scope the gate to the mutating handlers so pre-existing charts stay pullable, or keep the full block and ship a repair path plus an "am I affected" release note. **@brandonrc — your call.**

Three test items, all small:

1. **The `&&` in `is_oci_manifest_artifact` is untested.** The DB fixture seeds one row satisfying *both* the path arm and the media-type arm, so neither is independently load-bearing — deleting `path.starts_with("v2/") &&`, or flipping `&&` to `||`, leaves the whole new suite green. A legitimately uploaded chart stored at any path containing `/manifests/` would then silently vanish from `index.yaml` with nothing failing. Pin each arm with its own unit case. Also, `"v2/"` is a second hand-written copy of the literal `upsert_manifest_artifact` builds — the module already imports `OCI_MANIFEST_STORAGE_PREFIX`, so use that and the two can't drift.
2. **`manifest_put_rejects_classic_helm_repository_before_writes` has no positive control in its own fixture.** Mutating the validator to reject *everything* satisfies all four of its assertions (400, `UNSUPPORTED`, zero artifacts, zero tags). Only out-of-file docker tests kill that. Add a `helm_oci` PUT asserting 201 + one artifact, so blanket-reject dies inside the test that claims the behavior.
3. **`test_resolve_repo_rejects_classic_helm_repository` asserts only the status code** — any unrelated 400 satisfies it. One line: assert `errors[0].code == "UNSUPPORTED"`, as its pure-unit sibling already does.

Worth one comment line in this PR, because it will otherwise get deleted later as redundant: enforcement is `/v2`-only, but `oci_referenced_content.rs` (migration import) and `peer_instance_service.rs` (federation replication) also write OCI manifest rows with no format gate. The read-side filter is what actually keeps `index.yaml` clean for those two paths — it is not belt-and-braces.

Follow-ups, not blocking: the error message discloses the repo's format before the authz check on private repos, and `/v2/_catalog` still advertises a classic helm repo whose every request now 400s.

Happy to push items 1–3 myself if you'd rather — say the word. Item 1 is the one I'd genuinely insist on.

Issue #3147, opened 2026-08-07
Proxy-cache reads can bypass the metadata sidecar: stale-if-error and PyPI Remote serve the content key unverified


brandonrc:
Surfaced while reviewing #3068. The proxy cache's safety model assumes readers gate on the `__cache_meta__.json` sidecar, but three paths reach the `__content__` key without consulting it. Nothing here is a regression — filing so the invariant is explicit and testable.

**1. `handle_streaming_leader_upstream_error` (stale-if-error)**

On a transient upstream failure it does:

```rust
if let Ok(body) = self.storage.get_stream(cache_key).await {
    let metadata = self.load_cache_metadata(metadata_key).await.unwrap_or(None);
    ...
    return Ok(StreamHandle { body, headers });
}
```

The body is read **first**; the sidecar is loaded afterwards, best-effort, and `unwrap_or(None)` swallows both "absent" and "unreadable". No expiry gate, no `checksum_sha256` verification, no `quarantine_until` check, no ETag revalidation. Any object sitting at that key is served — including one whose sidecar was deliberately withheld.

**2. PyPI Remote arm** streams `artifacts.storage_key` via a raw `get_stream`, with no sidecar, digest, or freshness check.

**3. `is_fresh` multipart fallback.** When either ETag is multipart-shaped, an ETag mismatch is treated as inconclusive and degrades to `matches!(exists(cache_key), Ok(true))`. Multipart means large objects. Combined with a surviving sidecar (immutable entries are `Fresh` regardless of `expires_at`), a presigned 302 can point at an object that was never verified.

**Why it matters**

The reject arms in `CachePersister` rely on skipping the sidecar to force a refetch. That works only for `CacheStore::get`, which re-verifies SHA-256. For the paths above, the object being *absent* is what protects them — which is why the reject arms also delete. #3068 originally proposed removing those deletes on the grounds that the sidecar skip was sufficient; it isn't, and the deletes were retained for this reason.

**Suggested direction**

- Make the sidecar gate explicit for stale-if-error: require a loadable sidecar before serving a stale body, and treat `None` as "nothing to serve" rather than "serve with unknown headers".
- Give the PyPI Remote arm the same freshness gate the other read paths use.
- Add a test that asserts a rejected object is never served: today `TeeRecordingBackend::get()` returns `NotFound` unconditionally, so the tee test module is structurally incapable of covering it.

**Adjacent, lower priority**

[...]

PR #3153, opened 2026-08-07
fix(proxy-cache): publish streamed writes via a staging key, never live

### LINKED ISSUE #3152: Streaming proxy-cache writer can leave a permanently corrupt object as the live cache entry
## Summary

`CachePersister::tee_stream` writes a streamed upstream body directly to the live `cache_key` via `put_stream`, and only *afterward* classifies whether the write was good (`classify_stream_write`: empty / truncated / digest mismatch => reject, else commit). On reject, it calls `storage.delete(&cache_key)` as best-effort cleanup — no version pinning, failure swallowed at `debug!`, no retry.

Under concurrent cold-cache requests for the same hot artifact — a thundering herd, since the streaming path's leader-election (`proxy_hydration.rs`) is per-process/per-pod, not cluster-wide — a rejected writer's bad bytes land directly on the live key. If the cleanup delete itself then fails (which it did, silently, in a real occurrence of this), a corrupted (0-byte or truncated) object becomes the permanent "current" version at that key, with no delete marker and no trace anything went wrong.

Once that happens, the entry does not self-heal: the streaming cache-hit path (`open_cached_stream`) does zero integrity verification against the metadata sidecar before streaming bytes to the client — it trusts whatever object is currently at the key. The buffered path (`CacheStore::get`) does recompute the checksum on every hit, but large binaries go through the streaming path, which doesn't.

## Observed impact

Confirmed in production: two independent artifacts (a Maven jar and a `.pom`-adjacent binary) ended up with a 0-byte or truncated `__content__` object as the current version, with the real content sitting underneath as an older, non-current version. Both were manually repaired by identifying and restoring the correct version. No delete marker was present in either case — the self-heal delete had failed silently and left no trace.

## Root cause

`backend/src/services/proxy_service.rs`, `CachePersister::tee_stream`:
1. `put_stream` writes the streamed body directly to `cache_key`, before any validation.
2. `classify_stream_write` determines the outcome after the write has already landed.
3. On reject, `storage.delete(&cache_key)` is attempted, but it's unconditional, unpinned to a specific version, and its failure is only logged at `debug!` with no retry — effectively invisible to operators.

## Suggested fix

Write the streamed body to a private, per-write staging key instead of `cache_key` directly. Only `copy()` the staging object onto `cache_key` once the write is classified as good (and, for the digest-gated commit case, the checksum matches). A rejected write, or a failed publish, then just discards the staging object — the live key is never touched by an unvalidated write, so there's nothing to race or leave corrupted regardless of whether cleanup succeeds.

## Related

#3147 documents a complementary read-side gap: several readers (stale-if-error, PyPI Remote, `is_fresh`'s multipart-ETag fallback) serve the `__content__` key directly without consulting the sidecar. That issue is about defense-in-depth on the read side, assuming a bad object *could* be present. This issue is about the write-side mechanism that actually put one there in the first place — fixing the write side (as proposed above) removes the specific scenario #3147 worries about for content written via `tee_stream`, though the broader read-side hardening in #3147 remains independently valuable.

ysolomon-plat:
## Why

Fixes #3152.

`CachePersister::tee_stream` wrote a streamed upstream body directly to the live `cache_key` via `put_stream`, then classified the result afterward. On reject, it called `storage.delete(&cache_key)` as best-effort cleanup with no version pinning and failures swallowed at `debug!`.

Under concurrent cold-cache requests for the same hot artifact, a rejected writer's bad bytes land directly on the live key, and if the cleanup delete itself fails silently, a corrupted (0-byte/truncated) object becomes the permanent "current" version — confirmed in production on two separate occurrences. The streaming cache-hit path (`open_cached_stream`) does zero integrity verification against the sidecar, so an entry corrupted this way never self-heals.

## What

`put_stream` now targets a private, per-write staging key under its own dedicated prefix (`proxy-cache-staging/<uuid>`, unrelated to any content key). The live `cache_key` is only ever touched via `StorageBackend::copy`, once the write is classified `Commit` (and, for the digest-gated commit case, the checksum validates). A rejected write, or a failed publish, discards the staging object and never touches the live key at all.

The staging key started out suffix-based (`{cache_key}.tee-staging.<uuid>`) and was reworked to the dedicated-prefix form after review — the suffix shared a prefix with real cache content, which meant: no S3 lifecycle rule could target orphaned staging objects without also expiring real cache entries; the PyPI project listing needed a workaround to keep staging objects out of it (briefly regressed nested distributions, fixed along the way); and the content-key length budget had to grow to cover the staging suffix, pushing a band of long paths from "caches fine" to a hard 400. The dedicated prefix removes all three: staging objects never appear in any `proxy-cache/`-scoped listing, are independently lifecycle-sweepable, and their length no longer depends on any content path.

Also:
- `FilesystemBackend::copy` was a bare `fs::copy` straight onto the destination — non-atomic, reopening the same race on the filesystem backend now that `copy()` sits on the hot path. Rewritten to copy-to-temp + fsync + rename.

Also extends the existing multi-replica concurrency harness (`scripts/concurrency/`, `docker-compose.concurrency-e2e.yml`) with an opt-in fault-injection knob (`FAIL_EVERY_NTH`/`FAIL_MODE`) so it can reproduce a genuinely bad upstream response racing a concurrent good one against real S3-compatible storage.

## Known carve-outs (not fixed here)

- **>5 GiB S3 objects**: `S3Backend::copy` falls back to `get_stream` + `put_stream` streamed directly onto the live key for objects above the single-copy threshold, checking the digest only after that write completes. On a digest mismatch, the corrupt bytes are already live and the sidecar-bypassing read paths (stale-if-error, PyPI Remote, `is_fresh`'s multipart-ETag fallback → presigned 302) can serve them — narrowly reopening the bug class this PR fixes, for a rare case (S3-object corruption during an internal restream on an object over 5 GiB). Not adding a reactive delete-on-mismatch here: that delete could clobber a concurrent writer's just-published good object, recreating the exact pattern this PR removes. The real fix is server-side `UploadPartCopy` (never restreams through the app), which the `object_store` crate doesn't currently expose — worth its own follow-up issue.
- **Orphaned staging objects**: nothing in-process reaps a staging object if the writer task is killed before cleanup (a pod eviction mid-write, a crash). The dedicated prefix makes this cheaply solvable at the infra layer — an S3/GCS/Azure lifecycle rule expiring `proxy-cache-staging/` after a day — rather than needing in-app sweep code (`StorageBackend::list` doesn't return object age, so an in-app sweep would need a HEAD per key or a trait extension). Filesystem-backed deployments have no equivalent lifecycle mechanism and would need a cron/tmpwatch-style sweep instead if that's a supported production configuration.

## Test plan

- [x] New regression test asserting two concurrent writers (one reject, one commit) for the same key never `put_stream` or `delete` the literal live key.
- [x] New tests for the previously-untested `copy()`-failure path and digest-mismatch commit arm.
- [x] `copies` recorder on the test double asserts the publish step itself — previously entirely unverified.
- [x] `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --lib` (14,720 passed) all clean.
- [x] Every `storage.list()` call site in the codebase checked directly against the new `proxy-cache-staging/` prefix to confirm staging objects are unreachable from any of them.

## Related

#3147 documents readers that bypass the sidecar gate on the content key. This PR removes the specific scenario that issue worries about for content written via `tee_stream` (the live key can only ever hold validated content now, modulo the >5 GiB carve-out above), though the broader read-side hardening in #3147 remains independently valuable.

### DIFF
--- a/backend/src/services/proxy_service.rs
+++ b/backend/src/services/proxy_service.rs
@@ -9178,6 +9178,87 @@ mod tests {
         assert_eq!(copies[0].1, "cache-key");
     }
 
+    /// The `Err` arm: an upstream error mid-stream must leave the live key
+    /// untouched and discard whatever was staged.
+    ///
+    /// NOT a revert-proof guard, and labelled so deliberately: it passes on
+    /// the pre-fix shape too. At merge-base this arm had no delete at all (the
+    /// writer's only two deletes were the digest-mismatch and reject arms),
+    /// which looks like a third corruption path -- a failed `put_stream`
+    /// leaving partial bytes at the live key forever. It is not one, because
+    /// every `put_stream` implementation is atomic at the destination:
+    /// `FilesystemBackend` writes a temp file and only renames on success
+    /// (removing the temp on any error), `S3Backend` runs multipart behind an
+    /// abort guard, `GcsBackend` aborts the resumable session, and the default
+    /// trait impl buffers before calling `put`. So there was never anything at
+    /// `cache_key` for the missing delete to remove. Verified by running this
+    /// test against the pre-fix shape rather than by reading the arms.
+    ///
+    /// It earns its place as an invariant pin, not as a gate: it is the
+    /// assertion that fails if a future backend gains a `put_stream` that
+    /// streams straight to the destination key, at which point the arm really
+    /// would need its own cleanup. It also covers what the pre-existing
+    /// `test_tee_upstream_error_mid_stream_aborts_cache` leaves out -- that
+    /// one asserts only `metadata_writes.is_empty()`, and `#3068` established
+    /// that three readers reach `__content__` without consulting the sidecar,
+    /// so "no sidecar" is not sufficient protection on its own.
+    ///
+    /// Uses the real filesystem backend so the assertion is about bytes on
+    /// disk rather than about which calls a mock recorded.
+    #[tokio::test]
+    async fn test_upstream_error_mid_stream_leaves_live_key_untouched() {
+        use crate::services::storage_service::FilesystemBackend;
+
+        let temp_dir = tempfile::TempDir::new().unwrap();
+        let root = temp_dir.path().to_path_buf();
+        let storage = Arc::new(RealStorageService::new(Arc::new(FilesystemBackend::new(
+            root.clone(),
+        ))));
+
+        // One good chunk, then the upstream dies.
+        let upstream: BoxStream<'static, Result<Bytes>> = Box::pin(futures::stream::iter(vec![
+            Ok(Bytes::from_static(b"first-chunk")),
+            Err(AppError::Storage("upstream exploded".to_string())),
+        ]));
+
+        let mut client = CachePersister::new(test_catalog_pool(), Arc::clone(&storage)).tee_stream(
+            upstream,
+            "proxy-cache/repo/errored".to_string(),
+            "proxy-cache/repo/errored.meta".to_string(),
+            template(),
+            None,
+        );
+        let mut saw_error = false;
+        while let Some(chunk) = client.next().await {
+            if chunk.is_err() {
+                saw_error = true;
+            }
+        }
+        assert!(saw_error, "the upstream error must surface to the client");
+
+        tokio::time::sleep(Duration::from_millis(300)).await;
+
+        assert!(
+            !storage.exists("proxy-cache/repo/errored").await.unwrap(),
+            "an aborted upstream must not leave a partial object at the live key"
+        );
+        assert!(
+            !storage
+                .exists("proxy-cache/repo/errored.meta")
+                .await
+                .unwrap(),
+            "an aborted upstream must not leave a metadata sidecar"
+        );
+        let staging_root = root.join(TEE_STAGING_KEY_PREFIX.trim_end_matches('/'));
+        let leaked: Vec<_> = std::fs::read_dir(&staging_root)
+            .map(|d| d.flatten().map(|e| e.path()).collect())
+            .unwrap_or_default();
+        assert!(
+            leaked.is_empty(),
+            "the staged partial must be discarded: leaked {leaked:?}"
+        );
+    }
+
     /// The test that actually distinguishes this fix from `main`.
     ///
     /// Everything else in this module -- including the real-backend round trip
--- a/backend/src/services/proxy_service.rs
+++ b/backend/src/services/proxy_service.rs
@@ -311,6 +311,15 @@ const TEE_CHANNEL_DEPTH: usize = 64;
 /// docstring above. Smaller upstream chunks pass through unchanged.
 const TEE_MAX_CHUNK_BYTES: usize = 64 * 1024;
 
+/// Suffix `CachePersister::tee_stream` appends to a content key to derive
+/// its private staging key. Shared with `ProxyService::check_cache_key_length`
+/// so the two can't drift apart.
+const TEE_STAGING_KEY_SUFFIX: &str = ".tee-staging.";
+
+/// Byte length of a hyphenated UUID v4 string (`Uuid::new_v4().to_string()`),
+/// e.g. "550e8400-e29b-41d4-a716-446655440000". Always exactly 36.
+const UUID_STRING_LEN: usize = 36;
+
 /// Default TTL for APT InRelease/Release index files (5 minutes).
 /// Controls how often the proxy re-checks upstream for changes.
 pub const DEFAULT_DISTS_INDEX_TTL_SECS: i64 = 300;
@@ -948,6 +957,20 @@ async fn pin_storage_etag(storage: &StorageService, cache_key: &str) -> Option<S
     })
 }
 
+/// Best-effort cleanup of a `CachePersister::tee_stream` staging object. A
+/// failure here only leaks a staging object — the live cache key is never
+/// touched by this call. `warn`, not `debug`: nothing else sweeps leaked
+/// staging keys, so this is the only signal one is accumulating.
+async fn discard_staging(storage: &StorageService, staging_key: &str) {
+    if let Err(e) = storage.delete(staging_key).await {
+        tracing::warn!(
+            staging_key = %staging_key,
+            error = %e,
+            "best-effort delete of proxy-cache staging object failed"
+        );
+    }
+}
+
 /// Storage keys for a single proxy-cache entry: the artifact body and its
 /// `__cache_meta__.json` sidecar.
 ///
@@ -1502,6 +1525,16 @@ impl CachePersister {
     ///   metric buckets honest: a flaky mirror does not inflate the
     ///   `STORAGE_ERROR` rate, and a genuine cache backend failure does
     ///   not get hidden as `BAD_GATEWAY`.
+    ///
+    /// Publishes via a staging key: the body is written to a private
+    /// per-write key, never to `cache_key` directly, and only
+    /// [`StorageBackend::copy`]'d onto `cache_key` once classified
+    /// [`StreamWriteOutcome::Commit`]. A rejected write never touches the
+    /// live key at all, so it can't leave a corrupt object at that key
+    /// regardless of whether the subsequent cleanup succeeds — closing the
+    /// scenario the sidecar-bypassing read paths described on the reject
+    /// arm below actually depend on (there is nothing bad ever sitting at
+    /// `cache_key` for them to serve).
     fn tee_stream(
         &self,
         upstream: BoxStream<'static, Result<Bytes>>,
@@ -1523,10 +1556,16 @@ impl CachePersister {
         let (tx, rx) = tokio::sync::mpsc::channel::<Result<Bytes>>(TEE_CHANNEL_DEPTH);
 
         // Spawn the storage writer. It consumes the channel as a stream
-        // and calls put_stream. On completion, writes the metadata sidecar
-        // with the observed SHA-256 + byte count.
+        // and calls put_stream against a private staging key, never
+        // `cache_key` directly (see the `copy` call below). On completion,
+        // writes the metadata sidecar with the observed SHA-256 + byte count.
         let storage_clone = storage.clone();
         let cache_key_for_writer = cache_key.clone();
+        // Unique per write so concurrent writers for the same hot key never share one.
+        let staging_key = format!(
+            "{cache_key_for_writer}{TEE_STAGING_KEY_SUFFIX}{}",
+            Uuid::new_v4()
+        );
         tokio::spawn(async move {
             // Adapter: receiver -> futures::Stream<Result<Bytes>>.
             let rx_stream = futures::stream::unfold(rx, |mut rx| async move {
@@ -1534,7 +1573,7 @@ impl CachePersister {
             });
 
             let put_result = storage_clone
-                .put_stream(&cache_key_for_writer, Box::pin(rx_stream))
+                .put_stream(&staging_key, Box::pin(rx_stream))
                 .await;
 
             match put_result {
@@ -1549,26 +1588,9 @@ impl CachePersister {
                         // the digest — but the proxy cache must never be poisoned
                         // with mismatched content addressed by that digest. Treat
                         // a mismatch exactly like the truncated/empty reject path:
-                        // delete the object and skip the metadata sidecar so the
-                        // next request refetches upstream (self-heal).
-                        //
-                        // The delete is load-bearing, NOT belt-and-braces: the
-                        // missing sidecar is not sufficient on its own, because
-                        // several read paths reach the content key without
-                        // consulting it. `handle_streaming_leader_upstream_error`
-                        // serves `get_stream(cache_key)` as an RFC 5861
-                        // stale-if-error response and only loads the sidecar
-                        // afterwards, best-effort, for headers; `is_fresh`
-                        // degrades to a bare `exists()` when either ETag is
-                        // multipart-shaped (#2120) or when a legacy sidecar has
-                        // no pinned ETag; and the presigned-redirect paths sign a
-                        // 302 straight to the object with no checksum check at
-                        // all. Leaving rejected bytes in place therefore makes
-                        // them servable. Deleting is also what self-heals on a
-                        // versioned bucket: nothing in this codebase ever reads a
-                        // non-current version (there is no version_id support in
-                        // the storage layer), so a delete marker simply reads as
-                        // NotFound and forces the refetch.
+                        // discard the staging object and skip the sidecar so
+                        // the next request refetches (self-heal); `cache_key`
+                        // was never touched.
                         if let Some(expected) = template.expected_checksum.as_deref() {
                             if result.checksum_sha256 != expected {
                                 tracing::warn!(
@@ -1577,18 +1599,29 @@ impl CachePersister {
                                      not caching (no metadata sidecar) so the cache entry cannot \
                                      be poisoned"
                                 );
-                                if let Err(e) = storage_clone.delete(&cache_key_for_writer).await {
-                                    tracing::debug!(
-                                        cache_key = %cache_key_for_writer,
-                                        error = %e,
-                                        "best-effort delete of digest-mismatched proxy-cache \
-                                         object failed; the missing metadata sidecar still forces \
-                                         a refetch"
-                                    );
-                                }
+                                discard_staging(&storage_clone, &staging_key).await;
                                 return;
                             }
                         }
+                        // Publish: copy the validated staging object onto the
+                        // live key via the backend's native copy (atomic —
+                        // readers see the old or new version, never a partial
+                        // write). A copy failure degrades like a reject.
+                        if let Err(e) = storage_clone
+                            .copy(&staging_key, &cache_key_for_writer)
+                            .await
+                        {
+                            tracing::warn!(
+                                cache_key = %cache_key_for_writer,
+                                error = %e,
+                                "failed to publish staged proxy-cache object to the live key; \
+                                 not caching (no metadata sidecar) so the next request refetches \
+                                 upstream"
+                            );
+                            discard_staging(&storage_clone, &staging_key).await;
+                            return;
+                        }
+                        discard_staging(&storage_clone, &staging_key).await;
                         let now = Utc::now();
                         // #1618 S9 / #1051: pin the storage backend's ETag at
                         // write time so the fast path can re-HEAD on each hit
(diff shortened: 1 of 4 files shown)


brandonrc:
Ran this through the pipeline: reproduced #3152 on the merge-base, verified the fix with a negative control, reviewed, and pushed one fix as `013fcc6`.

**The core change is right, and the tests are real guards.** Reverting the staging write (put back to the live key, drop the `copy`, restore the live-key deletes) fails six tests, including `test_tee_concurrent_reject_and_commit_never_touch_live_key_directly`. Reproduction on the base: upstream advertises 100 bytes, delivers 11, and those 11 corrupt bytes are the live key's current version with no sidecar. At head, the live key is untouched **even when the cleanup delete fails** — which is the whole point, and is what the production report described.

Worth stating explicitly for anyone reading the history: **this supersedes the reject-arm deletes that #3068 restored, and retiring them is not just tidy — keeping them would now be actively wrong.** Under a thundering herd a rejecting writer's `delete(cache_key)` would destroy a *concurrent good writer's* object. Your concurrency test pins exactly that. I'd also note the staging approach is the better fix than the deletes were: it removes the producer rather than racing to clean up after it.

On **#3147** — this closes the `tee_stream` producer by construction, so the three sidecar-bypass readers have nothing bad to serve *from this source*. It does not close the issue: `write_negative_cache` still deletes the live key directly, and those readers still serve `cache_key` with zero integrity verification, so any other corruption source is still served blind. I'll keep #3147 open and note this narrowed it.

---

**One fix pushed — a listing regression that ships green.**

`pypi_package_names_from_cache_keys` was tightened from `tail.is_empty()` to an exact match on the whole tail. But PyPI caches distribution files one level deeper than the index: `build_pypi_proxy_cache_path` yields `simple/<name>/<filename>`, so a cached wheel keys as

```
proxy-cache/<repo>/simple/requests/requests-2.31.0.tar.gz/__content__
```

and its tail is `requests-2.31.0.tar.gz/__content__`, matching neither literal. Every project represented only by cached distributions silently vanished from `GET /pypi/<repo>/simple/` — and since proxy-cached items have no `artifacts` rows (#1278), that listing is their only local record. Both existing tests use flat keys, which is why CI was green.

Now matches the last segment, which still excludes staging (final segment `__content__.tee-staging.<uuid>`). Added tests for both directions.

---

**Two things I did not change, one of which I'd weigh before merging.**

**Orphaned staging objects are unbounded and unswept.** I confirmed there's no reaper that can reach them: the only `storage.list()` deleter is `purge_repo_cache`, which runs on repository deletion, and the hourly `sweep_orphan_staging_files` is scoped to `<storage_path>/.incoming`, a different keyspace entirely. They're also invisible to accounting — `proxy_catalog::upsert` keys on the live key inside the Commit arm, so orphans cost real bytes and appear in nothing. The `warn!` isn't the backstop the comment suggests: it only fires when `delete()` returns `Err`, while the dominant producer is the detached `tokio::spawn` writer with no cancellation token, so every rolling deploy and pod eviction orphans every in-flight staging object silently.

The shape of the fix matters here. Because `.tee-staging.` is a *suffix*, the surrounding prefix is byte-identical to the live key, so no S3 lifecycle rule can express it without also expiring the real cache. Putting staging under its own prefix (`proxy-cache-staging/<uuid>`) instead would make it prefix-sweepable, remove it from the `proxy-cache/` listing space entirely — which also retires the regression I just fixed rather than patching around it — and decouple it from the key-length budget. That last point is the second item: the worst-case suffix grew from 19 to 60 bytes, so a band of long paths flips from "caches fine" to a hard 400, making those artifacts unfetchable through the proxy rather than merely uncached.

I'd take the dedicated-prefix variant if you're willing, since it collapses three findings into one change. If not, an hourly sweep over `proxy-cache/` for staging keys older than N hours plus threading the shutdown token into the writer would cover the leak.

Also worth a carve-out in the PR description rather than a code change: on S3 above 5 GiB, `copy` falls back to `get_stream` + `put_stream` **directly to the live key** and only then verifies the digest, so the "never touches the live key" invariant has an exception at the top end.

Gates on the pushed commit: `cargo fmt --check` clean, the `tee_stream` and PyPI listing suites pass.

ysolomon-plat:
Pulled in 013fcc6 and took the dedicated-prefix suggestion. Changes since your review:

**Staging key moved to its own prefix.** `{cache_key}.tee-staging.<uuid>` → `proxy-cache-staging/<uuid>` (`TEE_STAGING_KEY_PREFIX`), fully decoupled from any content key. This closes all three things you flagged in one move:
- No longer shares a prefix with real cache content, so it's now cleanly targetable by an S3/GCS/Azure lifecycle rule without touching real entries.
- Never matches any `proxy-cache/`-scoped listing at all (checked every `storage.list()` call site: `list_cached_pypi_packages`, `list_cached_paths`, `purge_repo_cache` — none of them can see it).
- Length no longer depends on the content path, so `check_cache_key_length` and its two boundary tests are reverted to their exact original (pre-#3153) form. The 19→60-byte budget growth is gone entirely rather than just patched.

**Kept 013fcc6 as defense in depth.** The last-segment matching in `pypi_package_names_from_cache_keys` is still independently correct for real inputs even though staging can no longer reach that code path at all now — updated the comments to say so rather than leaving them implying it's still a live concern.

**Doc fix.** The staging-key constant's comment said "lifecycle rule after a few hours" — S3 expiration granularity is days, not hours. Fixed, and added a note that filesystem-backed deployments have no lifecycle mechanism and would need a cron-style sweep instead.

**Carve-outs written up in the PR description rather than code:**
- The >5GiB `copy()` exception (streams onto the live key before verifying digest) — agreed a reactive delete-on-mismatch would risk clobbering a concurrent writer's good object, so this stays a known gap. Filed #3164 for the real fix (`UploadPartCopy`, once/if `object_store` exposes it).
- Orphaned staging objects — going with the lifecycle-rule approach you suggested rather than in-app sweep code, since `StorageBackend::list` has no object-age info to sweep on anyway. Documented as an operator requirement.

Also fixed a stray-whitespace typo in the existing `skips_staging_objects` test data while I was in there.

All green: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo test --workspace --lib` (14,720 passed).

brandonrc:
Re-reviewed at `9a585a3` after the staging-prefix change. **SHIP WITH FIXES** — the core fix is real and I re-verified it end-to-end against a real `FilesystemBackend` rather than the PR's mock.

**First, a correction I owe you.** I earlier called the PyPI listing change a merge blocker and described it as shipping green, which implied `main` was affected. **It wasn't.** At merge-base `676f12ca` the predicate is:

```rust
if name.is_empty() || tail.is_empty() {
```

which *admits* nested distributions. The whole-tail exact match was introduced by `deda24da` — this PR's own first commit — so `013fcc6` fixed a self-inflicted regression inside the PR, not a defect against main. Net effect versus main is still a small genuine improvement, because the exact-leaf match also excludes in-flight `.tmp` files that `tail.is_empty()` let through. But the severity I assigned was wrong, and the comment I left at `proxy_service.rs:4336-4349` now narrates history that never happened — worth rewriting to state the invariant instead, and renaming `..._skips_staging_objects`, whose input shape is unreachable now that staging lives outside `proxy-cache/`.

**The bug is worse than first recorded, in your favour.** A rejected refetch of an already-cached path didn't merely leave corrupt bytes — it *destroyed* a previously-good entry (overwrite, then best-effort delete). Reverting `tee_stream` to merge-base behaviour against a real filesystem backend:

```
#3152: the previously-cached good object was DESTROYED by a rejected write:
NotFound("Storage key not found: proxy-cache/pypi-remote/simple/requests/.../__content__")
```

Control passes on both sides, so no false-positive blocking.

**The staging-prefix change is a clear improvement** — it retires the key-length finding entirely (`check_cache_key_length` is now byte-identical to merge-base, so the band of long paths that flipped to a hard 400 is gone). One regression came with it, though: the old suffix key lived under `proxy-cache/<repo>/`, so `purge_repo_cache` *did* sweep staging on repository delete. The relocation traded partial coverage for none.

---

**One pre-merge ask, and it's the important one.** The mock accepts a mutant that destroys the feature. Moving `discard_staging` to *before* the `copy` that reads it — which on any real backend means nothing is ever cached again — leaves **all 18 tee tests passing**:

```
test result: ok. 18 passed; 0 failed
```

`TeeRecordingBackend::copy` records `(src, dst)` and returns `Ok(())` without checking `src` exists, and `delete` only appends to a `Vec`. The suite asserts *which* keys were touched, never that the sequence is executable. A real-`FilesystemBackend` control catches it instantly. That's ~30 lines and closes the whole blind spot.

**Two cheap corrections while you're there:** the log at `:1626` and the docstring at `:1541-1549` say `cache_key` was never touched, which is false above 5 GiB on S3 — `copy` degrades to `get_stream` + `put_stream` **directly to the live key** and only then verifies the digest, returning `Err` without deleting dest. Still strictly better than main (which wrote direct-to-live at all sizes), so it's a follow-up — but the log will misdirect incident response.

**Follow-ups rather than this diff:**

The lifecycle mitigation named at `:323-331` is the **wrong rule**. S3 `put_stream` starts a multipart upload unconditionally on the first non-empty chunk, and `MultipartAbortGuard` only fires on drop — a SIGKILL gives no drop. An `Expiration` rule does not reap abandoned MPUs; that needs `AbortIncompleteMultipartUpload`. Abandoned parts are billed and invisible to `ListObjects`. And `grep -rn "proxy-cache-staging"` over the whole repo returns **exactly one hit** — the const itself. No lifecycle config, no chart, no docs, no reaper; filesystem has no mechanism at all.

On filesystem, publish is `copy` → `sync_all` → `rename`. The atomicity fix is right and valuable, but staging and live share one `base_path`, so a plain **`rename` would have been atomic, O(1), and free** — and would retire the discard step and the filesystem leak together. As written it's 2× writes, 3× total I/O, and publishing an object of size S now needs **2S** free; below that floor the copy fails, staging is discarded, space recovers, and the next request repeats forever.

**Merge order with #3172, and a trap.** They're complementary — without #3153, #3172's mid-stream abort would leave a partial over-ceiling object at the *live* key. **Merge #3153 first.** But a trial merge produces exactly one conflict (a docstring collision), and once resolved the tree still **does not compile**: #3172 adds a 6th `tee_stream` parameter and #3153's four new test call sites pass five arguments.

```
error[E0061]: this method takes 6 arguments but 5 arguments were supplied
  --> proxy_service.rs:9459 / 9492 / 9501 / 9578
```

Whoever merges second adds `None` at those four sites; git will not warn them.

**Refuted, for the record:** no re-opening of #2504 — `from_config` builds S3 with `prefix = None`, per-repo routing is a different handle, and the staging key is a v4 UUID never derived from input. `"proxy-cache-staging/"` also does not match `is_proxy_cache_key`'s `starts_with("proxy-cache/")`, so there's no accidental prefix collision.

Gates clean at head: `cargo fmt --check` and `clippy --workspace --all-targets -- -D warnings` both exit 0.

brandonrc:
Re-reviewed at `9a585a3`. The dedicated `TEE_STAGING_KEY_PREFIX` addresses everything I raised last round — the staging objects are out of `proxy-cache/`, `discard_staging` documents that nothing in-process reaps them, and the log at the publish failure and the `tee_stream` docstring now both describe what the code actually does. The PyPI listing comment is correct too.

One gap left, which I've pushed a test for (`7e0cb04`).

Every tee test asserts on the *calls* the persister makes — `put_stream` went to a staging key, `copy` went to the live key — against `TeeRecordingBackend`. None of them proves those calls move any bytes. A mock whose `copy()` is a no-op that appends `(src, dest)` to a `Vec` satisfies all eleven of them while storing nothing, so the suite would stay green if publishing were broken end to end. That is worth closing on a PR whose entire subject is *where* bytes land.

The new test runs the same path through a real `FilesystemBackend` and checks the state a later cache hit actually depends on: the bytes readable at `cache_key`, and the absence of a leaked staging object. Both arms, since the reject arm is where the bug lived — a rejected write must leave the live key **absent**, and before this PR the truncated body went straight to `cache_key`, so a subsequent hit served 5 bytes as though they were the whole artifact.

Revert-proved rather than eyeballed. Setting `staging_key = cache_key` (the pre-fix shape) fails it:

```
assertion `left == right` failed: a committed write must be readable at the live key
with the exact upstream bytes -- this is what the next cache hit serves
  left: None
 right: Some([102, 105, 114, 115, 116, 45, 99, 104, 117, 110, 107])
```

`None` rather than a corrupt object, because with the two keys collapsed the best-effort staging discard deletes what it just published — which is a nice illustration of why the separate prefix matters.

I also checked the atomicity claim per backend rather than taking the comment's word for it, since the reject arm leans on it: `FilesystemBackend::copy` writes a temp file, `sync_all`s it, `rename`s onto the destination and fsyncs the parent directory; S3/GCS/Azure `copy` is a server-side atomic operation. A reader sees the old object or the new one, never a partial write. The comment is accurate.

Gates on this head: `cargo fmt --check` clean, `clippy --workspace --all-targets -D warnings` exit 0, 436 `proxy_service` + `rpm` tests pass.

For the record on my own earlier report: the two rpm failures I mentioned were me pointing this workspace at another PR's test database, not anything in your change. And the PyPI nested-distribution "blocker" I filed earlier was wrong, as I corrected above — main was never broken.

From my side this is done pending the review round I'm running on `9a585a3`.

brandonrc:
Correcting my own last comment. **The round-trip test I pushed does not distinguish this fix from `main`**, and a review round caught it.

The pre-fix reject arm already called `delete(&cache_key)` (merge-base `proxy_service.rs:1580` and `:1730`). Streaming straight to the live key and then deleting it reaches the same final state as never writing there — so every assertion in that test holds under a faithful revert. Its staging-leak check is worse than neutral: pre-fix the staging directory never exists, `read_dir` errors, and the `let Ok(..) else { continue }` swallows it into an empty vec. It passes vacuously.

The mutation I proved it against — `staging_key := cache_key` while *keeping* the copy and discard — is a shape the pre-fix code never had. It failed only because the discard then deleted what the copy had just published. That's an artifact of self-deletion, not detection of the bug. I checked the merge-base source rather than taking the report's word for it, and it's right.

`3093f2d` adds the test that does gate it.

The pre/post difference is only observable when the **cleanup delete fails** — which is the incident this PR cites. Pre-fix, a failed delete leaves the truncated body at the live cache key for the sidecar-bypassing readers to serve. Post-fix the live key was never written, so a failed delete leaks a staging object and nothing more. So: a real `FilesystemBackend` with exactly one fault injected, every `delete` errors. Under a faithful revert it fails with the assertion that matters:

```
a rejected write must leave the live cache key absent even when the cleanup delete
fails ... the next request served 5 bytes as though they were the whole artifact.
```

The writer-finished barrier counts delete **attempts** rather than polling for a staging object. A staging-shaped signal would make the synchronization assume the fix, so the revert would fail on the barrier instead of on the claim — which is exactly how my first version went wrong. Both shapes attempt one delete on the reject arm, so the counter is fair to either. It also replaces the sleep-then-assert-absence pattern, which couldn't tell "the writer correctly avoided the live key" from "the writer never ran."

Also pinned `is_proxy_cache_key(staging_key) == false`. That exclusion keeps in-flight staging objects out of the presigned-redirect path, and it holds *only* because of the trailing slash in `"proxy-cache/"` — renaming the constant to `proxy-cache/staging/` would silently invert it.

Gates: `fmt --check` clean, `clippy --workspace --all-targets -D warnings` exit 0, 334 `proxy_service` tests pass.

**Two things I'd like your call on before I sign this off**, both from the same round and both outside what a test can settle:

1. **Staging orphans have no reaper.** `storage_gc_service` contains no `.list(` at all — every delete it makes is driven by a DB row, so an object with no row is unreachable. Nothing in lifecycle, the prefix sweeper, or `purge_repo_cache` reaches `proxy-cache-staging/`, and no lifecycle rule ships in the repo. Meanwhile the writer's post-client tail grew from one sidecar `put` to `copy` + `delete` + sidecar, so the window in which a SIGTERM leaks an in-flight staging object is now seconds-to-minutes on large objects instead of milliseconds — every rolling deploy leaks some. The repo already solved this exact pattern for `oci-uploads/<uuid>` with the `oci_upload_cleanup_keys` journal (migration 116) plus the claim-based reaper in `storage_gc_service`. `template.repository_id` is already in scope at the staging-key mint. Worth folding in, or a follow-up issue?

2. **A pre-existing HIGH the new docstring now denies.** `classify_stream_write` only enforces truncation when `expected_len.is_some()`. On a chunked upstream `Content-Length` is stripped, so a client disconnect mid-fetch yields any non-zero byte count → `Commit` → the partial body is published *and* gets a blessing sidecar. Not caused by this PR, but the docstring at `:1536-1546` now asserts "there is nothing bad ever sitting at `cache_key` for them to serve," which is stronger than what the code guarantees. I'd narrow it to "no *classified-reject* write reaches `cache_key`" and file the disconnect case separately — the enumerated sidecar-bypassing-reader notes that the old comments carried were the valuable part.

I can do both; say the word and I'll either fold them in or file them.

brandonrc:
Follow-up on one more review finding, which I'm **refuting** — with the caveat that its underlying observation about the code shape was correct.

The claim: the writer's `Err` arm is a third corruption path this PR closes. At merge-base it had no delete at all (the only two deletes were the digest-mismatch and reject arms), so a `put_stream` that failed mid-write would leave partial bytes at the live `cache_key` permanently — reachable by the three sidecar-bypassing readers from `#3068`.

The code shape is exactly as described. It still isn't a corruption path, because **every `put_stream` implementation is atomic at the destination**:

- `FilesystemBackend` writes to a temp file and only `rename`s on success, removing the temp on any error
- `S3Backend` runs multipart behind an abort guard
- `GcsBackend` aborts the resumable session
- the default trait impl buffers the whole body before calling `put`

So nothing ever reached `cache_key` for the missing delete to remove. The merge-base arm was harmless, not lucky.

I settled this by running a new test against the pre-fix shape rather than by reading the arms — it passes there too. `f47f5c2` keeps the test but its docstring says plainly that it is **not** a revert-proof guard, so nobody re-derives this or mistakes it for a gate.

It earns its place as an invariant pin: it's the assertion that fails if a backend ever gains a `put_stream` that streams straight to the destination key, which is when that arm really would need its own cleanup. It also covers what `test_tee_upstream_error_mid_stream_aborts_cache` leaves out — that one asserts only `metadata_writes.is_empty()`, and `#3068` established that "no sidecar" is not sufficient protection on its own.

The same review noted several other tee tests pass under a faithful revert (`test_tee_forwards_all_chunks_to_client_and_storage`, `test_tee_single_chunk_round_trip`, and the two I already corrected). That's accurate but not, I think, a merge gate: those predate this PR and pin the tee's *forwarding* contract, which this change doesn't touch. The gate is `test_rejected_write_leaves_live_key_absent_even_when_cleanup_fails`, which does fail on a faithful revert.

Gates on this head: `fmt --check` clean, `clippy --workspace --all-targets -D warnings` exit 0, 335 `proxy_service` tests pass.

That closes everything from my side except the two questions in my previous comment — the missing staging reaper, and narrowing the docstring that currently over-claims against the chunked-upstream disconnect case. Both are yours to call.

PR #3155, opened 2026-08-07
fix(pub): accept non-string dependency forms in pubspec.yaml

### LINKED ISSUE #3058: bug(pub): Flutter packages with SDK dependencies fail to publish with 400 Bad Request
### Description

Publishing a valid Flutter plugin to a hosted Pub repository fails during the multipart upload step with `400 Bad Request`.

The upload initialization request succeeds, but the actual package upload fails when the `pubspec.yaml` contains standard Flutter SDK dependencies such as:

```yaml
dependencies:
  flutter:
    sdk: flutter

dev_dependencies:
  flutter_test:
    sdk: flutter
```

This appears to happen because Artifact Keeper deserializes `dependencies` and `dev_dependencies` as `HashMap<String, String>`, while valid Pub dependency declarations may also be YAML mappings for SDK, Git, path, and hosted dependencies.

### Environment Information

* Deployment type: Docker Compose
* Dart Pub version: 3.9.2
* Flutter version: 3.35.7

### Backend Version

1.6.4

### Web UI Version

1.5.8

### Package Format

Pub (Dart/Flutter)

### Expected Behavior

Artifact Keeper should accept valid Flutter SDK dependency mappings in `pubspec.yaml`.

Publishing should complete using the standard Pub upload flow:

1. `GET /api/packages/versions/new` returns `200 OK`.
2. `POST /api/packages/versions/newUpload` accepts the multipart package archive and returns `204 No Content` with a `Location` header.
3. `GET /api/packages/versions/newUploadFinish` returns `200 OK`.
4. The published package version can be queried and downloaded from the hosted repository.

### Observed Behavior

Client-side package validation completes successfully.

The upload initialization request succeeds:

```text
GET /pub/pub-hosted/api/packages/versions/new
200 OK
```

However, the multipart archive upload fails:

```text
POST /pub/pub-hosted/api/packages/versions/newUpload
400 Bad Request
content-type: text/plain; charset=utf-8
content-length: 129
```

The Dart client then reports:

```text
Failed to upload the package.
```

The package archive is approximately 152 KB, so this does not appear to be an upload-size limit.

The package is not registered after the failed upload.

### Steps to Reproduce

1. Create a Flutter plugin with a standard Flutter SDK dependency:

   ```yaml
   name: test_flutter_plugin
   version: 1.0.0

   environment:
     sdk: ">=3.0.0 <4.0.0"
     flutter: ">=3.3.0"

   dependencies:
     flutter:
       sdk: flutter
     plugin_platform_interface: ^2.0.2

   dev_dependencies:
     flutter_test:
       sdk: flutter
     flutter_lints: ^5.0.0
   ```

2. Configure the hosted Artifact Keeper Pub repository:

   ```yaml
   publish_to: https://example.com/pub/pub-hosted/
   ```

3. Add a valid Artifact Keeper token:

   ```bash
   dart pub token add https://example.com/pub/pub-hosted/
   ```

4. Run:

   ```bash
   dart pub publish
   ```

5. Confirm publishing when prompted.

6. Observe that the upload initialization request returns `200 OK`, but the multipart `newUpload` request returns `400 Bad Request`.

### Logs

```shell
Dart Pub: 3.9.2
Flutter: 3.35.7

Package validation completed successfully.

HTTP GET https://example.com/pub/pub-hosted/api/packages/versions/new
HTTP response 200 OK

HTTP POST https://example.com/pub/pub-hosted/api/packages/versions/newUpload
HTTP response 400 Bad Request
content-type: text/plain; charset=utf-8
content-length: 129

Failed to upload the package.
Exception type: ApplicationException
package:pub/src/command/lish.dart
```

### Additional Context

The current `PubSpec` model in:

```text
backend/src/formats/pub.rs
```

defines Pub dependency values as strings:

```rust
pub dependencies: Option<HashMap<String, String>>,

#[serde(skip_serializing_if = "Option::is_none")]
pub dev_dependencies: Option<HashMap<String, String>>,
```

Valid Pub dependency declarations are not always strings. Flutter packages normally use YAML mappings:

```yaml
dependencies:
  flutter:
    sdk: flutter
```

Other valid mapping forms include:

```yaml
dependencies:
  local_package:
    path: ../local_package

  git_package:
    git:
      url: https://example.com/project.git
      ref: main
```

The suspected deserialization error is:

```text
invalid type: map, expected a string
```

A possible fix would be:

```diff
- pub dependencies: Option<HashMap<String, String>>,
+ pub dependencies: Option<HashMap<String, serde_yaml::Value>>,

  #[serde(skip_serializing_if = "Option::is_none")]
- pub dev_dependencies: Option<HashMap<String, String>>,
+ pub dev_dependencies: Option<HashMap<String, serde_yaml::Value>>,
```

It may also be appropriate to support:

```rust
pub dependency_overrides: Option<HashMap<String, serde_yaml::Value>>,
```

This is related to, but distinct from, #1997.

Issue #1997 addressed Pub upload protocol compliance, including the upload initialization method and upload response status. In this reproduction, upload initialization already succeeds and the failure occurs while processing the uploaded Flutter package.

### Pre-submission Checklist

andrlange:
## Summary

Fixes #3058.

Publishing any Flutter package to a hosted Pub repository failed with `400 Bad Request` during the multipart upload step. The reporter's hypothesis was correct, and the failure is reproducible at the exact line they suspected.

`PubSpec` modelled dependency values as strings:

```rust
pub dependencies: Option<HashMap<String, String>>,
pub dev_dependencies: Option<HashMap<String, String>>,
```

Pub allows a dependency value to be either a version string (`http: ^0.13.0`) **or** a mapping — for SDK, git, path and hosted dependencies. Every Flutter package declares `flutter: {sdk: flutter}`, so `serde_yaml` rejected the pubspec outright:

```
dependencies.flutter: invalid type: map, expected a string at line 10 column 5
```

`extract_pubspec_from_staged` turns that error into the `400` the reporter saw, which is why step 1 of the upload flow (`GET /api/packages/versions/new`) succeeded while step 2 failed.

This types both fields as `serde_json::Value` so every dependency form round-trips. That mirrors how `CargoToml::dev_dependencies` already solves the identical problem with `toml::Value` (`backend/src/formats/cargo.rs:406`), so it introduces no new pattern.

**Scope of impact:** `PubSpec` is only deserialized in the publish path — `extract_pubspec_from_staged` / `extract_pubspec_from_archive` are its sole call sites. Proxied metadata from an upstream registry is passed through as raw JSON and only has `archive_url` rewritten, so remote and virtual Pub repositories are unaffected by this change.

Two tests, at different levels:

1. `formats::pub::tests::test_deserialize_pubspec_with_flutter_sdk_dependencies_3058` — deserializes a Flutter pubspec directly.
2. `api::handlers::pub_registry::tests::test_extract_pubspec_from_flutter_plugin_archive_3058` — runs a real gzipped tar through `extract_pubspec_from_archive`, the function whose error becomes the 400. It covers all four non-string dependency forms: sdk, git, path and hosted.

Both were confirmed to fail on `main` before the fix. The archive-level test fails with the production error string:

```
Failed to parse pubspec.yaml: dependencies.flutter: invalid type: map, expected a string at line 10 column 5
```

## Test Checklist
- [ ] Manually tested locally

Local gate results:

```
cargo clippy --workspace --all-targets -- -D warnings    no warnings
cargo test --workspace --lib                            14721 passed; 0 failed; 6 ignored
```

One box deliberately left unchecked, so a reviewer does not have to guess: **Manually tested locally** — I did not run a real `dart pub publish` against a running instance. The tests cover the parse stage that produced the 400, but not the remainder of the upload protocol (`newUpload` → `newUploadFinish` → query → download) that the reporter enumerated under Expected Behavior. If you would like that verified end to end before merging, an E2E case in artifact-keeper-test exercising a Flutter plugin would be the right place, and I am happy to add one.

## API Changes
- [x] N/A - no API changes

`PubSpec` is an internal parsing type; it is not part of the OpenAPI surface. The wire format of the Pub metadata this affects is unchanged for every pubspec that parsed before — a string dependency still serializes as a string.

## Follow-up (not in this PR)

While tracing this I noticed a separate defect worth its own issue: `PubSpec` has no catch-all field, and the metadata stored at publish time is the re-serialized struct. Fields Pub permits but the struct does not declare — `flutter:`, `topics:`, `executables:`, `publish_to:` — are silently dropped from the pubspec served by `/api/packages/<name>`. It does not affect version solving (`dependencies` and `environment` survive), which is why it is not bundled here. Happy to file it separately if that is useful.

### DIFF
--- a/backend/src/api/handlers/pub_registry.rs
+++ b/backend/src/api/handlers/pub_registry.rs
@@ -1167,6 +1167,68 @@ mod tests {
         encoder.finish().expect("gzip finish")
     }
 
+    /// Negative controls for the #3058 widening.
+    ///
+    /// `dependencies` moved from `HashMap<String, String>` to an untyped
+    /// value, which is the right fix -- but it also means a fully
+    /// attacker-controlled pubspec now drives an untyped deserializer at
+    /// publish time. These pin that malformed input still degrades to a clean
+    /// `Err` (a 400) rather than a panic or a stack overflow, which is the
+    /// property the widening could plausibly have broken.
+    ///
+    /// Deliberately assert `is_err()` and not a specific message: the point is
+    /// that the request fails gracefully, not how the error is worded.
+    #[test]
+    fn test_malformed_pubspec_dependencies_fail_cleanly_3058() {
+        // Deeply nested map. serde_yaml caps nesting at 128 and returns
+        // RecursionLimitExceeded; without a limit this is a stack overflow,
+        // which aborts the process rather than the request.
+        let deep = format!(
+            "name: x\nversion: 1.0.0\ndependencies:\n  bomb:{}",
+            // Indentation must INCREASE per level -- a constant indent is 200
+            // siblings, not 200 levels deep, and parses fine.
+            (0..200)
+                .map(|i| format!("\n{}k:", " ".repeat(4 + i * 2)))
+                .collect::<String>()
+        );
+        assert!(
+            extract_pubspec_from_archive(&archive_with_pubspec(&deep)).is_err(),
+            "a 200-deep dependency map must be rejected, not overflow the stack"
+        );
+
+        // Non-string mapping keys, at both levels. Measured rather than
+        // assumed: serde coerces a scalar key to a string in BOTH the outer
+        // `HashMap<String, _>` and the untyped value, so these parse rather
+        // than erroring. That is benign -- what matters is that neither
+        // panics on the conversion, which is the failure mode a widened
+        // untyped deserializer could plausibly introduce.
+        for coerced in [
+            "name: x\nversion: 1.0.0\ndependencies:\n  1: foo\n",
+            "name: x\nversion: 1.0.0\ndependencies:\n  foo:\n    1: bar\n",
+        ] {
+            assert!(
+                extract_pubspec_from_archive(&archive_with_pubspec(coerced)).is_ok(),
+                "scalar mapping keys are coerced to strings, not rejected: {coerced:?}"
+            );
+        }
+
+        // Wrong container shape where a mapping is required.
+        let sequence = "name: x\nversion: 1.0.0\ndependencies:\n  - a\n  - b\n";
+        assert!(
+            extract_pubspec_from_archive(&archive_with_pubspec(sequence)).is_err(),
+            "a sequence where a mapping is required must be rejected cleanly"
+        );
+
+        // Positive control. Without this the three assertions above are
+        // satisfied by a parser that rejects everything -- including every
+        // real Flutter pubspec, which is the bug #3058 fixed.
+        let good = "name: x\nversion: 1.0.0\ndependencies:\n  flutter:\n    sdk: flutter\n";
+        assert!(
+            extract_pubspec_from_archive(&archive_with_pubspec(good)).is_ok(),
+            "the legitimate map form must still parse"
+        );
+    }
+
     /// Regression test for #3058: a real Flutter plugin archive must survive
     /// pubspec extraction. Every non-string dependency form Pub permits is
     /// present here; each one previously aborted the upload with 400.
--- a/backend/src/api/handlers/pub_registry.rs
+++ b/backend/src/api/handlers/pub_registry.rs
@@ -1145,6 +1145,90 @@ mod tests {
         assert!(result.is_err());
     }
 
+    /// Build a gzipped tar containing exactly one `pubspec.yaml` with `body`.
+    fn archive_with_pubspec(body: &str) -> Vec<u8> {
+        use flate2::write::GzEncoder;
+        use flate2::Compression;
+        use std::io::Write;
+
+        let mut tar_builder = tar::Builder::new(Vec::new());
+        let mut header = tar::Header::new_gnu();
+        header.set_path("pubspec.yaml").unwrap();
+        header.set_size(body.len() as u64);
+        header.set_mode(0o644);
+        header.set_cksum();
+        tar_builder
+            .append(&header, body.as_bytes())
+            .expect("append pubspec");
+        let tar_bytes = tar_builder.into_inner().expect("finish tar");
+
+        let mut encoder = GzEncoder::new(Vec::new(), Compression::default());
+        encoder.write_all(&tar_bytes).expect("gzip write");
+        encoder.finish().expect("gzip finish")
+    }
+
+    /// Regression test for #3058: a real Flutter plugin archive must survive
+    /// pubspec extraction. Every non-string dependency form Pub permits is
+    /// present here; each one previously aborted the upload with 400.
+    #[test]
+    fn test_extract_pubspec_from_flutter_plugin_archive_3058() {
+        let pubspec = r#"
+name: my_flutter_plugin
+version: 2.1.0
+description: A Flutter plugin with the full range of dependency forms.
+environment:
+  sdk: '>=3.0.0 <4.0.0'
+  flutter: '>=3.10.0'
+dependencies:
+  flutter:
+    sdk: flutter
+  http: ^0.13.0
+  some_git_pkg:
+    git:
+      url: https://example.com/pkg.git
+      ref: main
+  some_path_pkg:
+    path: ../local_pkg
+  some_hosted_pkg:
+    hosted:
+      name: some_hosted_pkg
+      url: https://pub.example.com
+    version: ^1.0.0
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  test: ^1.16.0
+"#;
+
+        let spec = extract_pubspec_from_archive(&archive_with_pubspec(pubspec))
+            .expect("Flutter plugin archive must yield a pubspec");
+
+        assert_eq!(spec.name, "my_flutter_plugin");
+        assert_eq!(spec.version, "2.1.0");
+
+        let deps = spec.dependencies.expect("dependencies present");
+        assert_eq!(deps["flutter"], serde_json::json!({ "sdk": "flutter" }));
+        assert_eq!(deps["http"], serde_json::json!("^0.13.0"));
+        assert_eq!(
+            deps["some_path_pkg"],
+            serde_json::json!({ "path": "../local_pkg" })
+        );
+        assert_eq!(
+            deps["some_git_pkg"]["git"]["url"],
+            serde_json::json!("https://example.com/pkg.git")
+        );
+        assert_eq!(
+            deps["some_hosted_pkg"]["hosted"]["url"],
+            serde_json::json!("https://pub.example.com")
+        );
+
+        let dev_deps = spec.dev_dependencies.expect("dev_dependencies present");
+        assert_eq!(
+            dev_deps["flutter_test"],
+            serde_json::json!({ "sdk": "flutter" })
+        );
+    }
+
     #[test]
     fn test_pub_error_response_format() {
         let resp = pub_error_response(StatusCode::NOT_FOUND, "NOT_FOUND", "Package not found");
--- a/backend/src/formats/pub.rs
+++ b/backend/src/formats/pub.rs
@@ -243,9 +243,18 @@ mod tests {
                     .collect(),
             ),
             dependencies: Some(
-                vec![("http".to_string(), serde_json::json!("^0.13.0"))]
-                    .into_iter()
-                    .collect(),
+                vec![
+                    ("http".to_string(), serde_json::json!("^0.13.0")),
+                    // The map form #3058 added support for. It has to survive
+                    // SERIALIZATION too, not just parsing: `newUpload`
+                    // persists `serde_json::to_value(&pubspec)` and the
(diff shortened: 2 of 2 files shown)


brandonrc:
Ran this through the review pipeline. **The production fix is correct and I wouldn't change a line of it** — I've pushed test coverage only (`76e05ef`).

**Bug verified at merge-base.** `dependencies: Option<HashMap<String, String>>` sends a mapping value into serde's `String` visitor → `invalid type: map, expected a string` → the 400 at `pub_registry.rs:656`. The 200-then-400 split in the report is explained too: `GET /api/packages/versions/new` never parses the archive. Every Flutter package was unpublishable. Correctly diagnosed.

**Your type choice is better than the one the issue proposed**, and it's worth saying why. `newUpload` does `serde_json::to_value(&pubspec).unwrap_or_default()`. With `serde_yaml::Value` as the issue suggested, a pubspec containing a non-string mapping key would make that conversion fail — and `unwrap_or_default()` would silently persist `"pubspec": null`. Silent metadata corruption on publish. `serde_json::Value` makes it infallible.

What was missing was negative coverage. A fully attacker-controlled document now drives an untyped deserializer at publish time, and nothing pinned that malformed input degrades to a clean 400 rather than a panic or a stack overflow. Added:

- a 200-deep nested dependency map → must be `Err`. serde_yaml caps nesting at 128 and returns `RecursionLimitExceeded`; without a limit that's a stack overflow, which takes the *process* down rather than the request.
- a sequence where a mapping is required → must be `Err`.
- **a positive control in the same test**, so those rejections can't be satisfied by a parser that rejects everything — including every real Flutter pubspec, which is the bug you're fixing.

**Two behaviours I pinned as measured rather than assumed.** I expected `dependencies: {1: foo}` and a non-string key inside a dependency value to error, and wrote the assertions that way. They failed — serde *coerces* scalar mapping keys to strings at both levels, so both parse fine. The test now records what the parser actually does. Benign, and worth pinning so a future change that turns coercion into a panic gets caught.

**Also asserted the serialize side**, which turned out to matter more than the parse side. `test_pubspec_serialization` built dependencies and then only checked `name`/`version`/`description` — so adding `#[serde(skip_serializing)]` to `dependencies` left the entire suite green while every published package served an empty dependency set and broke client version solving. It now carries the map form through `serde_json::to_value`, the exact call `newUpload` persists, and fails under that mutation.

Both your tests are genuine red→green, and both are inline `mod tests`, so they count toward the `--lib` coverage gate — nothing landed in `backend/tests/`, which is the trap a recent PR here hit.

Two things for you, neither blocking:

1. **`dependency_overrides`** — the issue asked for it and it isn't in the diff. No functional break (there's no `deny_unknown_fields`, so it's ignored rather than 400ing), but it joins `flutter:`, `topics:`, `executables:`, `platforms:` in being dropped from the pubspec served at `/api/packages/<name>`, since the persisted metadata is the re-serialized struct. You flagged this as a follow-up yourself and I think you're right that pub's solver only needs `dependencies` + `environment` — it just deserves an explicit answer in the thread rather than staying an open ask.
2. **YAML alias expansion** is bounded by a *jump* budget, not a node budget, so one anchored subtree referenced many times can still materialize far more than the 8 MiB input suggests. I'm explicitly **not** blocking on this: it's not novel — `artifact_metadata.rs:151` and `helm.rs:174` already do untyped `serde_yaml` on attacker-supplied `Chart.yaml` in production today — and the work runs on `spawn_blocking` behind an 8-permit semaphore. I'll file it as one hardening issue covering every YAML ingestion site, plus the fact that serde_yaml is `+deprecated` and unmaintained.

From my side this is ready once CI is green on the new head. Thanks — clean, well-scoped contribution with an honest PR body.

brandonrc (approved):
Approving. Verification round confirms every claim in the remediation against source, including the two I flagged as measured rather than assumed.

The deep-nesting fixture genuinely nests — indentation increases 2 per level to depth 202, and serde_yaml's 128 cap is enforced on the `deserialize_any` path that `serde_json::Value` actually uses, so it errors for the right reason (and at 41 KB it's nowhere near the 8 MiB entry cap, so it isn't a size-limit false positive).

The scalar-key coercion I recorded is confirmed at both levels: `deserialize_str` hands raw scalar bytes to `visit_str` without consulting `scalar.style` or `scalar.tag`, so YAML type resolution never runs — for the outer `HashMap<String, _>` and for `serde_json`'s `KeyClassifier` alike. That's genuinely non-obvious and would have been easy to get backwards, which is why it's pinned as measured.

And the serialize assertion bites: with `skip_serializing`, `json["dependencies"]` is absent, `Index for str` yields `Null`, and the assertion fails. The justification holds too — `newUpload` persists `serde_json::to_value(&pubspec)` and the registry serves it straight back, so the silent-empty-dependency-set failure mode was real.

Your production diff is unchanged and correct. Thanks for a clean, well-scoped contribution with an honest PR body — and for picking `serde_json::Value` over the `serde_yaml::Value` the issue proposed, which would have made `to_value` fallible and silently persisted `"pubspec": null` on a non-string key.

One-line follow-up, not a hold: `dev_dependencies` was widened identically and is serialized through the same call, but nothing asserts it — `skip_serializing` on that field alone still leaves the suite green. Same class of hole, much lower consequence.

andrlange:
Thanks for the depth here — particularly for verifying at the merge-base rather than taking the diagnosis on trust, and for the negative-coverage commit. That gap was real and I should have seen it: widening a parser for fully attacker-controlled input and then only proving that *valid* documents now pass is half a test suite. Noted for next time.

Also useful: the note that both tests being inline `mod tests` is what makes them count toward the `--lib` coverage gate. I did that by following the surrounding file rather than by knowing the rule, so thanks for making it explicit.

**On `dependency_overrides`** — you're right that it was in the issue and that it deserves an answer rather than staying an open ask. I missed it: I read the reproduction and expected-behavior sections and acted on those without reading the suggested-fix section to the end. That's on me.

Having looked at it properly, I'd argue against adding the field on its own, for two reasons.

First, it would not change anything for consumers. `dependency_overrides` applies only to the root package being developed — pub ignores it when the package is consumed as a dependency. So even a faithfully served `dependency_overrides` would not affect any client's version solving, which is the same reasoning we both applied to `flutter:` and `topics:`.

Second, and this is the stronger reason: adding it field-by-field treats one symptom of a general defect. `PubSpec` has no catch-all and the persisted metadata is `serde_json::to_value(&pubspec)` (`pub_registry.rs:706`), so `dependency_overrides` is dropped for exactly the same reason `flutter:`, `topics:`, `executables:` and `platforms:` are. Adding one named field would leave the other four dropped while making the struct look complete.

What closes the class is a `#[serde(flatten)]` catch-all so unknown keys round-trip. That changes the shape of served metadata for every existing package, so it doesn't belong in an already-approved `fix/` PR. If you agree, I'll open it as its own issue + PR against the lossy-re-serialization defect, naming `dependency_overrides` explicitly so this ask is tracked rather than dropped.

Confirming the non-functional half of your note: there's no `deny_unknown_fields` on `PubSpec`, so a pubspec carrying `dependency_overrides` publishes fine today — it's silently omitted from the served metadata, not rejected.

**On the YAML alias-expansion point** — agreed it's out of scope here, and thanks for framing it as a pre-existing property of every YAML ingestion site rather than something this PR introduces. Happy to help on that issue once you've filed it, especially the `serde_yaml` deprecation half, which is the part that won't get smaller by waiting.

andrlange:
**CI note on the new head — this looks like a runner timeout, not a test failure.** Posting the evidence so you don't have to dig, and specifically so the new negative tests aren't left under suspicion.

`🧪 Backend Unit Tests` ran for **25m16s** against `timeout-minutes: 25` (`.github/workflows/ci.yml:337`). The log matches a hard timeout rather than a failing assertion:

- no assertion failure anywhere in the log
- no `test result:` summary — the output stops mid-run and jumps straight to container teardown
- the postgres service logged `checkpoint complete: wrote 1242 buffers (7.6%); write=123.877 s, total=123.880 s`

A checkpoint taking two minutes for 1242 buffers points at a badly starved runner rather than anything in the diff. (The `role "root" does not exist` lines are just the healthcheck probing, and `Security Scan` failed on `main` in the same window.)

I checked out `76e05ef` and ran it locally to be sure:

```
cargo test --lib pub_registry     35 passed; 0 failed
cargo test --workspace --lib      14722 passed; 0 failed; 6 ignored — finished in 153.40s
```

Worth stating explicitly, since it was my first hypothesis and it would have looked identical in the log: I suspected the 200-deep nesting fixture might overflow the stack and abort the test process, which would also produce an abrupt end with no summary. It doesn't — `test_malformed_pubspec_dependencies_fail_cleanly_3058` passes cleanly, so `RecursionLimitExceeded` is doing its job on that path. Your negative controls are fine.

So: a rerun should be all it needs; the suite takes ~2.5 minutes when the runner is healthy, against a 25-minute budget. I can't trigger one from a fork, so it's yours whenever convenient. Nothing on my side needs changing, and `✅ CI Complete` is just mirroring the same job.

PR #3158, opened 2026-08-07
fix(oci-gc): re-check blob liveness immediately before the cleanup storage delete

### LINKED ISSUE #3085: OCI cleanup sweep: blob liveness is not re-checked immediately before the storage delete
The cleanup-journal sweep re-checks blob liveness when it **claims** a key and again at the guarded row `DELETE`, but not in the window immediately before the destructive storage delete itself. A digest re-pushed during that window can therefore lose its bytes: the sweep deletes the object for a digest that has just become live again.

Pre-existing, but #3072 rewrote exactly this path and added a per-key renew hook right before the delete — which is the natural place to re-assert liveness.

Identified during the #3072 multi-replica red-team.

brandonrc:
## Summary

Closes #3085 as titled -- the missing liveness re-check in the cleanup-key
sweep. It does NOT close the underlying race; see "The window" below and the
follow-up in #3187, which tracks the actual closure.

The OCI cleanup-journal sweeps re-checked blob liveness when they **claimed** a key and again at the guarded row `DELETE` — but the row `DELETE` runs *after* the storage object has already been deleted. Nothing re-consulted liveness in the window immediately before the destructive delete, so a digest re-pushed during that window lost its bytes.

**The window.** A push commits an `oci_blobs` row for the very same `storage_key` (the push path re-uses the existing journal row via `ON CONFLICT (storage_key) DO NOTHING`, so a stale claimed row survives a re-push). The sweep's per-key pre-delete hook — `renew_cleanup_key_claim`, added by #3072 — extended the claim lease without consulting liveness, returned `true`, and `storage.delete()` destroyed a live blob. The guarded row `DELETE` that follows then correctly refused to remove the journal row because `oci_blobs` now referenced the key: it saved the row, not the data. The result is a live `oci_blobs` row pointing at bytes that no longer exist.

A claimed batch is up to `OCI_UPLOAD_CLEANUP_KEY_SCAN_LIMIT` (1000) rows walked sequentially against an object store, so the window is the whole batch walk, not an instant.

Both sweeps were affected: `cleanup_unreferenced_oci_upload_keys` and `reap_pending_oci_upload_cleanup_keys` share the hook.

**The fix.** The pre-delete hook (renamed `hold_cleanup_key_claim_for_delete`) now re-asserts the sweep's full liveness predicate in the *same atomic* `UPDATE` that extends the claim lease. A key that has become live cannot have its lease renewed. `CleanupSweepKind` selects the predicate matching the calling sweep's own row `DELETE`, so the pre-delete check and the post-delete row removal agree on what "still reclaimable" means.

## What this does and does not guarantee

**This narrows the window. It does not close it.** Read that literally — the code's own doc comment on `hold_cleanup_key_claim_for_delete` says the same thing, and this section exists so the PR does not claim more than the code delivers.

The `UPDATE` runs on the pool as a single autocommit statement. Its row lock is on `oci_upload_cleanup_keys`, never on `oci_blobs`, and it is released the instant the statement commits. The caller then performs the object-store `DELETE` holding **nothing**. A push that commits its `oci_blobs` row inside that gap still has its bytes destroyed. Nothing in the push path serializes against the sweep.

What changes is the size of the exposure, not its existence:

| | before | after |
|---|---|---|
| liveness evaluated | once per batch, in the candidate `SELECT` | once per key, immediately before its delete |
| window spans | the entire batch walk — up to 1000 sequential object-store deletes, i.e. minutes | one storage round trip, order tens of milliseconds |

That is a large reduction and it is worth shipping on its own. It is not a proof.

**Closing it properly** requires the push and the sweep to serialize on the journal row. Two options, both shapes this codebase already runs elsewhere:

1. **`FOR UPDATE` serialization** — an explicit transaction holding `SELECT ... FOR UPDATE` on the journal row across the storage delete, with the push taking that same row lock before committing `oci_blobs`.
2. **A `pending_delete_at`-style two-phase tombstone** — mark, then delete storage, then reap, with a push clearing the marker to resurrect. This is exactly what migration 141 gave `oci_blobs` for blob GC (`run_blob_gc_mark` / `run_blob_gc_sweep`), and it holds no lock across the storage call.

Tracked in **#3187**, which costs both.

**This is not the same guarantee as `is_blob_still_orphan`.** An earlier draft of this PR described the new hook as "the same shape as" the per-row re-check in the two-phase blob GC. That comparison is wrong and has been corrected in the doc comment. `is_blob_still_orphan` runs inside an open transaction that took `SELECT ... FOR UPDATE` on the `oci_blobs` row and **holds that lock across the storage delete** (`storage_gc_service.rs`, `run_blob_gc_sweep`), so a racing pusher blocks. This new hook holds nothing. Same predicate, different guarantee.

## Ordering

Unchanged: storage delete first, then the guarded row `DELETE`. That order fails safe.

- A crash between the two leaves the journal row behind; the next sweep re-issues the delete (idempotent — `NotFound` is treated as success) and removes the row. An orphaned storage object is recoverable.
- Deleting the DB row first would leave an untracked storage object that no sweep can ever find again (the journal *is* the sweep's key list) — a permanent leak — and would not close the same window, since a push can still land between the row `DELETE` commit and the storage call.

**Correction: a key going live does not keep the journal row visible.** An earlier draft argued the current order preserves evidence, because the row `DELETE`'s `oci_blobs` guard keeps the journal row when a key goes live. It does not, and this should not be counted as a benefit of the ordering:

- The push's success path calls `clear_oci_upload_cleanup_key_best_effort` — `backend/src/api/handlers/oci_v2.rs:978`, a bare `DELETE FROM oci_upload_cleanup_keys WHERE storage_key = $1` (lines 981-983) with no `claim_token` guard and no claim check. The push removes the row itself, out from under an in-flight sweep.
- The sweep then never notices. Its guarded row `DELETE` does not inspect `rows_affected`; it increments `cleanup_rows_removed` and `result.storage_keys_deleted` unconditionally — `backend/src/services/storage_gc_service.rs:1726-1727`. A zero-row match is counted as a **clean success**.

So in the exact interleaving where a push wins the race, there is no surviving journal row and no error. #3187 covers making that observable.

## Schema migration

**This PR adds one:** `backend/migrations/193_oci_blobs_storage_key_index.sql`.

```sql
CREATE INDEX IF NOT EXISTS idx_oci_blobs_storage_key ON oci_blobs (storage_key);
```

**Why.** The liveness predicate contains `NOT EXISTS (SELECT 1 FROM oci_blobs b WHERE b.storage_key = ...)`. That subquery used to run once per batch inside the candidate `SELECT`, so a sequential scan was amortised across the whole batch; the per-key re-check runs it once per candidate. `oci_blobs` had no index on `storage_key` — migration 118 indexed `oci_upload_sessions.storage_temp_key` and `oci_upload_parts.storage_key`, but nothing needed this one per-row until now.

Measured on 300k blobs: **87 ms per hold without the index vs 0.19 ms with it**; a full 1000-row batch went from 30.7 s to 0.19 s (~163x). The cost scales linearly with registry size.

**Lock behaviour.** `CREATE INDEX CONCURRENTLY` is not usable here (sqlx runs each migration inside a transaction; `CONCURRENTLY` is rejected in a transaction block), matching migration 166.

A plain `CREATE INDEX` takes a **`SHARE` lock** on `oci_blobs` for the duration of the build — **not `ACCESS EXCLUSIVE`**, which is what an earlier revision of the migration comment claimed. The difference matters for anyone planning a maintenance window:

- **Writes block.** `SHARE` conflicts with `ROW EXCLUSIVE`, so `INSERT`/`UPDATE`/`DELETE` on `oci_blobs` wait — blob commits on the push path, and blob GC's row deletes.
- **Reads do not block.** Plain `SELECT` takes `ACCESS SHARE`, compatible with `SHARE`. Pulls and manifest resolution are unaffected.

`oci_blobs` is far smaller than `artifacts` (one row per stored blob, deduplicated by `(repository_id, digest)`), so the build is short — but the write stall is real and should be scheduled on a large registry.

**Recommended, not optional.** Correctness does not depend on the index; forward progress does. Without it the re-check still returns the right answer via a sequential scan, but per the measurement above a batch on a large registry approaches the 15 minute claim TTL, and a sweep whose leases lapse mid-batch stops reclaiming storage. Skipping it trades a bounded write stall for an unbounded GC stall — the migration comment no longer offers that as an equivalent choice.

Operators who cannot accept the write stall should build the index out of band **before** deploying, at which point `IF NOT EXISTS` makes the migration a no-op:

```sql
CREATE INDEX CONCURRENTLY idx_oci_blobs_storage_key ON oci_blobs (storage_key);
```

That is the only supported way to avoid it — the migration cannot simply be skipped, since sqlx tracks applied migrations by checksum.

No API change. Patch-shaped for 1.7.2.

Tests are in-crate (`backend/src/services/storage_gc_service.rs` `mod tests`) so they count toward the `nextest --lib` coverage gate.

`MidBatchPushStorage` reproduces the production shape deterministically: it records every storage delete and, when the sweep deletes the **head** of a claimed batch, performs the concurrent push (commits an `oci_blobs` row) for a **later** key in the same batch. No timing or real concurrency needed.

- `unreferenced_sweep_keeps_bytes_of_blob_that_went_live_after_claim`
- `pending_reaper_keeps_bytes_of_blob_that_went_live_after_claim`

Each seeds three keys — `trip` (drives the push), `revive` (`oci-blobs/<digest>`, goes live mid-batch), and `control` (stays genuinely unreferenced). **Positive control:** the assertion that `control` IS still swept is in the same test, so the guard cannot pass by refusing to delete anything.

Two no-DB tests (`cleanup_sweep_liveness_predicates_*`) pin the predicate shape on every CI run — that the hook actually names `oci_blobs` / `oci_upload_parts` / `oci_upload_sessions`, and that each sweep keeps its own `storage_write_completed_at` marker so the kinds cannot be swapped.

### Test isolation and ordering (fixed in this revision)

Both regression tests passed alone and under `cargo nextest`, but **failed** in the single-process `cargo test --workspace --lib` run that CI also executes. Two independent causes, both in the test code — no production behaviour was involved:

1. **Wrong serialization primitive.** They took only `blob_gc_serial_lock` (a Postgres advisory lock), while the sibling `test_run_gc_*` tests — which run the *same* unscoped, cluster-wide cleanup-key sweeps — take `storage_gc_test_guard` (an in-process mutex). The two do not exclude each other, so a sibling's sweep reaped these tests' fixtures first and the recorded delete list came back empty. Under nextest the `db-serial` group already serialized the module, which is exactly why that run stayed green and hid it. Fixed by taking both guards and scoping each sweep to the fixture's own repository, so foreign rows cannot join the batch at all.

2. **Dependence on batch walk order.** `MidBatchPushStorage` fired the simulated push only when it saw one specific "head" key. The claim query orders its candidate CTE by `created_at` but returns rows from `UPDATE ... RETURNING`, which carries **no row-order guarantee** — the `ORDER BY` does not survive it, and the observed order flips with the plan once the table holds other tests' rows. When `revive` was walked before the head it was deleted before the push made it live, and the test reported #3085 data loss against a guard that was working correctly. Fixed by firing the push on the first delete of any key that is not `revive`, seeding `revive` newest so an order-preserving plan puts it last, and asserting the push actually fired — so a degenerate order fails as a clear precondition rather than a false #3085 red.

**Scope of the tests.** They cover the post-claim gap — a push landing after the batch was claimed. They do **not** cover the residual post-hold, pre-storage-delete gap described above, because no lock is held there and the test harness cannot deterministically interleave it. That coverage belongs with the fix in #3187.

**Revert-proof (re-verified after the test changes above).** With only the production guard removed (the `{liveness}` clause dropped from the hook, tests untouched), both regression tests fail on exactly the assertion above:

```
running 2 tests
test ...::pending_reaper_keeps_bytes_of_blob_that_went_live_after_claim ... FAILED
test ...::unreferenced_sweep_keeps_bytes_of_blob_that_went_live_after_claim ... FAILED

#3085: blob liveness must be re-checked immediately before the storage delete. An
oci_blobs row committed after the claim makes the blob live, so its bytes must
survive the sweep: ["oci-uploads/gc3085u-trip/2cd734fc...", "oci-blobs/sha256:2cd734fc...", "oci-uploads/gc3085u-control/2cd734fc..."]

test result: FAILED. 0 passed; 2 failed
```

The `oci-blobs/...` key appearing in the recorded delete list is the data loss. With the guard in place it is absent while `trip` and `control` are still deleted.

## Test Checklist
- [x] Manually tested locally

Local gates against a dedicated Postgres 16 (`sqlx migrate run` applied cleanly through 193):

- `cargo fmt --check` — clean
- `cargo test --lib -- storage_gc_service` — `110 passed; 0 failed`, stable across repeated runs

The `storage_gc_service::tests` module is already pinned to the `db-serial` nextest group.

The documentation corrections in this revision touch only doc comments and the migration's SQL comment block — no `sqlx::query!` macro changed, so `.sqlx` is untouched and remains valid for the offline build.

## API Changes
- [x] Migration is reversible (if applicable) — `193_oci_blobs_storage_key_index.sql` is a pure additive index; reverse with `DROP INDEX idx_oci_blobs_storage_key;`
- [ ] N/A - no API changes (schema migration added; see **Schema migration** above)

### DIFF
--- a/backend/migrations/193_oci_blobs_storage_key_index.sql
+++ b/backend/migrations/193_oci_blobs_storage_key_index.sql
@@ -1,0 +1,38 @@
+-- Migration: index oci_blobs.storage_key for the OCI cleanup sweep's
+-- pre-delete liveness re-check (#3085).
+--
+-- The sweep re-asserts blob liveness immediately before each destructive
+-- storage delete. That predicate contains
+-- `NOT EXISTS (SELECT 1 FROM oci_blobs b WHERE b.storage_key = k.storage_key)`.
+--
+-- Before #3085 that subquery ran once per batch, inside the candidate SELECT,
+-- so a sequential scan was amortised across the whole batch. The re-check runs
+-- it once per candidate, and oci_blobs has no index on storage_key -- migration
+-- 118 indexed oci_upload_sessions.storage_temp_key and oci_upload_parts
+-- .storage_key, but not this one, because nothing needed it per-row until now.
+--
+-- Measured on 300k blobs: 87 ms per hold without the index versus 0.19 ms with
+-- it; a full 1000-row batch (OCI_UPLOAD_CLEANUP_KEY_SCAN_LIMIT) went from
+-- 30.7 s to 0.19 s, ~163x. The cost scales linearly with registry size, so on a
+-- large registry a batch would begin to approach the 15 minute claim TTL and
+-- leases would lapse mid-sweep.
+--
+-- CREATE INDEX CONCURRENTLY is intentionally not used here, matching migration
+-- 166: sqlx::migrate runs each migration file inside a transaction, and
+-- CONCURRENTLY is rejected inside a transaction block. The non-concurrent build
+-- takes ACCESS EXCLUSIVE on `oci_blobs` for the duration of the build, so blob
+-- uploads block until it finishes. This is a single-column btree on a table that
+-- is far smaller than `artifacts` (one row per stored blob, deduplicated by
+-- (repository_id, digest)), so the window is short -- but on a large registry,
+-- plan the upgrade accordingly.
+--
+-- Operators who cannot accept the lock window can skip this migration and build
+-- the index out of band against a quiescent table:
+--
+--   CREATE INDEX CONCURRENTLY idx_oci_blobs_storage_key ON oci_blobs (storage_key);
+--
+-- Correctness does not depend on it: without the index the liveness re-check
+-- still returns the right answer, just via a sequential scan. The index is
+-- purely a query-plan accelerator.
+CREATE INDEX IF NOT EXISTS idx_oci_blobs_storage_key
+    ON oci_blobs (storage_key);
--- a/backend/migrations/193_oci_blobs_storage_key_index.sql
+++ b/backend/migrations/193_oci_blobs_storage_key_index.sql
@@ -17,22 +17,44 @@
 -- large registry a batch would begin to approach the 15 minute claim TTL and
 -- leases would lapse mid-sweep.
 --
--- CREATE INDEX CONCURRENTLY is intentionally not used here, matching migration
--- 166: sqlx::migrate runs each migration file inside a transaction, and
--- CONCURRENTLY is rejected inside a transaction block. The non-concurrent build
--- takes ACCESS EXCLUSIVE on `oci_blobs` for the duration of the build, so blob
--- uploads block until it finishes. This is a single-column btree on a table that
--- is far smaller than `artifacts` (one row per stored blob, deduplicated by
--- (repository_id, digest)), so the window is short -- but on a large registry,
--- plan the upgrade accordingly.
+-- LOCK BEHAVIOUR. CREATE INDEX CONCURRENTLY is intentionally not used here,
+-- matching migration 166: sqlx::migrate runs each migration file inside a
+-- transaction, and CONCURRENTLY is rejected inside a transaction block.
 --
--- Operators who cannot accept the lock window can skip this migration and build
--- the index out of band against a quiescent table:
+-- A plain CREATE INDEX takes a SHARE lock on `oci_blobs` for the duration of
+-- the build. SHARE conflicts with ROW EXCLUSIVE, so for as long as the build
+-- runs:
+--
+--   * writes to oci_blobs BLOCK -- INSERT/UPDATE/DELETE, which means blob
+--     commits on the push path wait, as do blob GC's row deletes;
+--   * reads do NOT block -- plain SELECT takes ACCESS SHARE, which is
+--     compatible with SHARE, so pulls and manifest resolution are unaffected.
+--
+-- (It is not ACCESS EXCLUSIVE. Nothing here rewrites the table or takes the
+-- table fully offline; an earlier revision of this comment said otherwise and
+-- was wrong. Plan the window against blocked writers, not blocked readers.)
+--
+-- This is a single-column btree on a table far smaller than `artifacts` (one
+-- row per stored blob, deduplicated by (repository_id, digest)), so the build
+-- is short -- but the write stall is real, and on a large registry the upgrade
+-- should be scheduled accordingly.
+--
+-- Operators who cannot accept even that write stall should build the index out
+-- of band, BEFORE deploying this migration, against a live table:
 --
 --   CREATE INDEX CONCURRENTLY idx_oci_blobs_storage_key ON oci_blobs (storage_key);
 --
--- Correctness does not depend on it: without the index the liveness re-check
--- still returns the right answer, just via a sequential scan. The index is
--- purely a query-plan accelerator.
+-- The IF NOT EXISTS below then makes this migration a no-op. Note that this is
+-- the only supported way to avoid the stall: the migration itself cannot simply
+-- be skipped, since sqlx tracks applied migrations by checksum and a missing
+-- entry blocks the ledger.
+--
+-- DO build it, one way or the other. Correctness does not depend on the index
+-- -- without it the liveness re-check still returns the right answer, just via
+-- a sequential scan -- but forward progress does. Per the measurement above, on
+-- a large registry the un-indexed re-check pushes a batch toward the 15 minute
+-- claim TTL, and a sweep whose leases lapse mid-batch stops reclaiming storage.
+-- Skipping the index outright is not an equivalent option; it trades a bounded
+-- write stall for an unbounded GC stall.
 CREATE INDEX IF NOT EXISTS idx_oci_blobs_storage_key
     ON oci_blobs (storage_key);
--- a/backend/src/services/storage_gc_service.rs
+++ b/backend/src/services/storage_gc_service.rs
@@ -7447,16 +7447,24 @@ mod tests {
     /// the same batch live before the sweep reaches it.
     struct MidBatchPushStorage {
         db: PgPool,
-        trip_key: String,
         revive_key: String,
         revive_repo: Uuid,
         revive_digest: String,
         deleted: std::sync::Mutex<Vec<String>>,
+        pushed: std::sync::atomic::AtomicBool,
     }
 
     impl MidBatchPushStorage {
         fn deleted_keys(&self) -> Vec<String> {
             self.deleted.lock().expect("deleted lock").clone()
+        }
+
+        /// Did the simulated concurrent push actually commit its `oci_blobs`
+        /// row? If it never fired, the sweep reached the revive key first and
+        /// the test never entered the window it exists to cover — a false red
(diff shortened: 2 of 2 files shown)


brandonrc:
Addressed both review blockers in `185a397`.

**The index.** The re-check moved `NOT EXISTS oci_blobs` from once per batch to once per candidate, and `oci_blobs` has no index on `storage_key` — migration 118 indexed `oci_upload_sessions.storage_temp_key` and `oci_upload_parts.storage_key`, but not this one, because nothing needed it per-row until now. I verified the gap independently against a fully-migrated database:

```
Indexes:
    "oci_blobs_pkey" PRIMARY KEY, btree (id)
    "idx_oci_blobs_digest" btree (digest)
    "idx_oci_blobs_repo_digest" btree (repository_id, digest)
    "oci_blobs_pending_delete_idx" btree (pending_delete_at) WHERE pending_delete_at IS NOT NULL
```

Migration 193 adds it, and the PR body's "no schema migration" claim is now wrong and should be dropped.

Worth recording a mistake I made here: I first wrote `CREATE INDEX CONCURRENTLY`, which is what you'd want for a hot table. Migration 166 documents why that cannot work — `sqlx::migrate` runs each file inside a transaction and `CONCURRENTLY` is rejected there. My psql test passed only because psql runs in autocommit, so it never exercised the real runner. Corrected to a plain index, verified applying **inside** a transaction, with the lock window and the out-of-band `CONCURRENTLY` escape hatch documented the same way 166 does it.

**The atomicity claim.** The review is right and the probe settles it — the claim was false. The `UPDATE` runs on the pool as a single autocommit statement, so its row lock (on `oci_upload_cleanup_keys`, never on `oci_blobs`) releases at commit, and the storage `DELETE` then runs with no lock held. `register_oci_upload_cleanup_key` is `ON CONFLICT DO NOTHING`, so nothing in the push path serializes against it.

I've rewritten the doc comment to say what is and is not guaranteed, because the difference decides whether anyone finishes the job. What the re-check buys is real and worth keeping: exposure went from the entire batch walk — up to 1000 sequential object-store deletes, minutes — to a single storage round trip. It also names the two ways to close it properly: row-lock serialization across the delete, or the two-phase tombstone this codebase already implements for blob GC via `oci_blobs.pending_delete_at` (migration 141).

**Not changed, and worth a follow-up rather than this PR:** the two no-DB `cleanup_sweep_liveness_predicates_*` tests assert only on the pure function's return value and passed under every sabotage, including the predicate fully unwired. They pin the string, not the guard — keep them, but they are not protection. Same for the transient-error path parking a claim for the full 15 minutes without calling `release_cleanup_key_claim`, unlike the storage-resolve failure path.

Gates on the pushed commit: `cargo fmt --check` clean, `storage_gc_service` 110 passed, all 193 migrations apply clean from scratch.

PR #3159, opened 2026-08-07
fix(maven): accept both files[] key spellings in flat-object delete guards

### LINKED ISSUE #3156: Maven flat-key delete guards match only snake_case storage_key, so live GAV companions can be purged
Found while confirming #2938 is already fixed. This is a separate, live defect — a delete guard that fails **open**.

## The divergence

The **read** path deliberately accepts both spellings (`services/maven_flat_attribution.rs`, `OWNER_BY_METADATA_FILES_SQL`):

```sql
AND (am.metadata->'files' @> jsonb_build_array(jsonb_build_object('storageKey',  $1::text))
  OR am.metadata->'files' @> jsonb_build_array(jsonb_build_object('storage_key', $1::text)))
```

and its own doc comment states which one production actually writes:

> The GAV-grouped upload handler **has always** serialized `files[]` entries with camelCase keys (`"storageKey"`, since #418); snake_case `storage_key` is accepted as a fallback spelling for hand-repaired rows (#2706).

The **delete guards** match only snake_case. Three sites:

1. `api/handlers/repositories.rs` — `collect_repo_maven_flat_keys`, run before a repository delete:
   ```sql
   AND NOT EXISTS ( ... EXISTS (
       SELECT 1 FROM jsonb_array_elements(am.metadata->'files') f
       WHERE f->>'storage_key' = o.storage_key ) )
   ```
2. `services/storage_gc_service.rs` — `ORPHAN_MAVEN_FLAT_PREDICATE_SQL`, the parent-artifact guard, same shape.
3. `services/storage_gc_service.rs` — the checksum-sidecar guard in the same statement, `f2->>'storage_key' = regexp_replace(o.storage_key, ...)`.

## Why it matters

The guards are `NOT EXISTS` — they answer *"is anything still referencing this key?"* and a **miss means delete**. Since they miss the spelling the upload handler actually writes, a GAV-grouped companion file (`.pom`, `.module`, `-sources.jar`) recorded as `storageKey` is invisible to them. The key reads as unreferenced and is purged, while a parent artifact — potentially in a *different* repository — still serves it. That is the exact row-less-companion case #2574/#2584 exist to keep served.

Polarity is the problem: the read path fails safe (an unmatched spelling means "not attributed", so it 404s), while the delete guards fail **destructive**. And it is the primary spelling that is missed, not an edge case.

## Scope

Not a request-path performance concern — these run per repository-delete and per GC sweep, which is why #2942's index work correctly left them alone. This is correctness, and it is reachable on any deployment that has ever used a GAV-grouped Maven upload on a cloud backend.

## Suggested fix

Accept both spellings in all three guards, matching the read path. The cheapest correct form mirrors what the read path already does — jsonb containment against both keys — which also makes the guards index-eligible via `idx_artifact_metadata_files_gin` from migration 179 rather than the current per-row `jsonb_array_elements` scan.

Worth extracting the predicate into one shared constant so the read path and the delete guards cannot drift again. The drift here is exactly the shape of #418 → #2706: two places encoding the same fact, one of them updated.

A regression test should assert that a companion recorded as camelCase `storageKey` is **not** collected for purge by either guard.

brandonrc:
## Summary

Fixes #3156.

Three `NOT EXISTS` delete guards for row-less Maven flat objects matched only the
snake_case `files[]` key spelling (`f->>'storage_key'`), while the affected rows record
the companion under camelCase `storageKey` and the read path deliberately accepts both
(#2706). A `NOT EXISTS` guard answers *"is anything still referencing this key?"* and
**a miss means DELETE**, so a live companion file — and its checksum sidecars — read as
unreferenced and were purged while a live parent artifact still referenced them.

**Which rows these are.** No handler in the current tree writes `files[].storageKey` —
the #418-era GAV-grouped upload handler that did has since been removed, and `files[]`
is now read-only (`sbom.rs`, `expand_maven_secondary_files`, and the attribution SQL all
read it, nothing writes it). So this fix protects **legacy rows on deployments upgraded
through that version**, plus the hand-repaired snake_case rows of #2706. It is not
guarding against writes a current handler is still making. That also bounds the blast
radius: affected deployments are those with pre-#418-removal Maven history on a shared
cloud namespace (S3/GCS/Azure); filesystem backends isolate each repository's key space
and are unaffected.

Sites fixed:

1. `backend/src/api/handlers/repositories.rs` — `collect_repo_maven_flat_keys`, the
   purge list built before a repository delete (cross-repository reference test).
2. `backend/src/services/storage_gc_service.rs` — `ORPHAN_MAVEN_FLAT_PREDICATE_SQL`
   guard 2, the parent-artifact test.
3. Same statement, guard 5 — the checksum-sidecar base-key test.

### Reproduced first (against the base commit)

Bare `postgres:16-alpine` with all 178 migrations applied, seeded with the `files[]`
entry shape the repo's own fixture uses
(`repositories.rs` `test_expand_maven_secondary_files_emits_each_file`), and the
`maven_flat_object_owner` rows migration 170 backfills. Running each production
predicate verbatim:

```
READ PATH  OWNER_BY_METADATA_FILES_SQL for maven/com/example/demo/1.0.0/demo-1.0.0.pom
  gav-a, gav-b            <- attributed under camelCase

SITE 1  collect_repo_maven_flat_keys(gav-a)  -> purged on repo delete
  maven/com/example/demo/1.0.0/demo-1.0.0-sources.jar
  maven/com/example/demo/1.0.0/demo-1.0.0.pom          <- gav-b's LIVE parent lists it
  maven/com/example/demo/1.0.0/demo-1.0.0.pom.sha1
  maven/com/example/gone/9.9.9/gone-9.9.9.pom          <- genuine orphan (control)

SITES 2+3  select_orphan_maven_flat_objects()  -> collected by the GC sweep
  maven/com/example/demo/1.0.0/demo-1.0.0-sources.jar   metadata_files
  maven/com/example/demo/1.0.0/demo-1.0.0.pom           metadata_files
  maven/com/example/demo/1.0.0/demo-1.0.0.pom.sha1      derived_checksum
  maven/com/example/gone/9.9.9/gone-9.9.9.pom           metadata_files
```

Every live companion is collected for purge alongside the genuine orphan. For site 1
the loss is not hypothetical: once the deleted repository's attribution rows CASCADE
away, the *other* repository becomes the key's sole read-path owner —

```
DELETE FROM repositories WHERE id = <gav-a>;
  attribution_rows_left = 0
  OWNER_BY_METADATA_FILES_SQL -> gav-b   (sole owner, now serving bytes just purged)
```

After the fix the same predicates collect only the genuine orphan (sites 2+3), and
site 1 spares the cross-repository-referenced key.

### Shared predicate

The read path and the guards encoded the same fact in two places, which is how they
drifted (#418 -> #2706). They now share one definition,
`maven_flat_attribution::metadata_files_name_key_sql(files_expr, key_expr)`, which
renders the containment test against both spellings. It is parameterised by the SQL
expressions for the array and the key so all four callers can use it: a bind parameter
(`$1::text`) on the read path, an outer column (`o.storage_key`) in guards 1–2, and a
derived expression (`regexp_replace(...)`) in guard 3. Both fragments are composed by
this crate — no caller input reaches the SQL.

### Query plans

Containment is index-eligible via `idx_artifact_metadata_files_gin` (migration 179),
so the fix carries a genuine plan improvement as well. Measured on 20 002
`artifact_metadata` rows. The per-key form is
`is_maven_flat_object_still_orphan` — the locked re-check that runs once per GC
candidate:

| | before (`jsonb_array_elements` EXISTS) | after (containment) |
|---|---|---|
| plan | `Seq Scan on artifact_metadata` | `BitmapOr` of two `idx_artifact_metadata_files_gin` probes |
| buffers | 1000 | 12 |
| execution | 11.053 ms | 0.114 ms |

The whole-table sweep predicate stays a hash anti-join (no single key to probe with),
but its buffer count still drops from 61 009 to 1 891 because the containment test no
longer expands every `files[]` array per candidate row.

### Deliberately out of scope

`collect_repo_maven_flat_keys` has no sidecar-base guard at all — in either spelling —
so `<companion>.sha1` is still collected when the companion itself is spared. That is a
distinct pre-existing gap (the GC predicate has such a guard; the repository-delete
path does not), unchanged by this PR.

**Filed as #3188** rather than widened into this fix. Note this PR makes that gap
reachable in a new way: previously the camelCase-referenced companion was itself purged,
so losing its sidecar was the lesser half of a larger loss; now the companion is
correctly spared and the sidecar alone is destroyed. Deleting repo A leaves repo B
serving `demo-1.0.0.pom` with its `.sha1` gone — a warning under Maven's default
`checksumPolicy=warn`, a hard failure under `checksumPolicy=fail` and under Gradle
dependency verification. #3188 carries the reproduction and the suggested fix.

Four tests, all in-crate (`src/... mod tests`) so they count toward the
`nextest --lib` coverage gate:

- `repositories.rs` `collect_repo_maven_flat_keys_spares_camelcase_referenced_companion_3156`
- `storage_gc_service.rs` `test_orphan_maven_flat_scan_spares_camelcase_referenced_companion_3156`
  (covers guards 2 and 5)
- `maven_flat_attribution.rs` `metadata_files_name_key_sql_matches_both_spellings_by_containment`
- `maven_flat_attribution.rs` `owner_by_metadata_files_sql_uses_the_shared_fragment`

Each DB test asserts a camelCase-recorded companion is **not** collected, that the
snake_case spelling keeps working, and carries a **positive control** — a genuinely
unreferenced key that must still be collected, so the fix cannot pass by never purging
anything.

**The connective is pinned.** `metadata_files_name_key_sql_matches_both_spellings_by_containment`
originally asserted only that both spellings appear, that both use containment, and that
`jsonb_array_elements` is gone — every one of which still holds if the two-branch
predicate is "simplified" from `OR` to `AND`. A `files[]` entry carries **one** spelling,
so an `AND` matches nothing on real data: it would silently restore the #3156 data loss
across all three guards *and* break the read path, while looking like a tidy-up. The test
now asserts the connective directly and pins the whole rendered fragment.

Mutation-proved: flipping `OR` to `AND` in `metadata_files_name_key_sql` fails the test
with

```
the two spellings must be OR-ed, never AND-ed: a files[] entry carries one spelling,
so an AND matches nothing on real data and re-opens the #3156 purge of live companions
```

This matters because it is the only guard-correctness check that runs **without** a
database: both DB-backed tests above skip silently when `DATABASE_URL` is unset.

**Revert-proved.** With the three production guards reverted to their snake_case-only
form (tests untouched), both DB tests fail:

```
collect_repo_maven_flat_keys_spares_camelcase_referenced_companion_3156 ... FAILED
  a companion another repository's live parent lists as camelCase `storageKey` must
  NOT be purged by this repository's delete (#3156); collected:
  ["maven/gav3156/.../demo-1.0.0.pom", "maven/gav3156/.../gone-9.9.9.pom"]

test_orphan_maven_flat_scan_spares_camelcase_referenced_companion_3156 ... FAILED
  guard 2: a companion its live parent lists as camelCase `storageKey` must NOT be
  collected for purge (#3156); collected:
  ["maven/gav3156/.../demo-1.0.0.pom", "maven/gav3156/.../demo-1.0.0.pom.sha1",
   "maven/gav3156/.../gone-9.9.9.pom"]

test result: FAILED. 0 passed; 2 failed
```

Note in both failure sets that the snake_case key is absent (it was already spared) and
the control key is present — the assertions distinguish the two spellings rather than
just counting.

## Test Checklist
- [x] Integration tests added/updated (if applicable) — the two DB-backed `--lib` tests
- [x] Manually tested locally

Local gates, isolated Postgres 16 (not a shared instance):

```
cargo test --workspace --lib                            ok. 14727 passed; 0 failed; 6 ignored
```

Duplication measured with the gate's own settings (`jscpd --min-lines 10 --format rust`)
over `backend/src/services` and `backend/src/api/handlers`, before and after: unchanged
at 0.51 % and 1.8 % respectively (identical duplicated-line counts), both well under
3 %. The shared fixture the two new DB tests need was factored into
`test_db_helpers::attach_maven_files_metadata` rather than copied.

## API Changes
- [x] N/A - no API changes

No schema migration: `idx_artifact_metadata_files_gin` already exists (migration 179).
Patch-shaped, targeting 1.7.2.

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -3955,8 +3955,19 @@ async fn purge_storage_object_keys(
 /// artifact's metadata `files[]` reference): deleting this repository must
 /// never destroy an object another tenant still serves. Best-effort: a DB
 /// error logs and returns the empty set.
+///
+/// The `files[]` guard is a `NOT EXISTS`, so a *missed* reference means
+/// DELETE. It therefore has to match every spelling the read path accepts, and
+/// gets that from the shared [`metadata_files_name_key_sql`] fragment rather
+/// than an inline copy: the inline copy matched snake_case `storage_key` only,
+/// so the camelCase `storageKey` entries the GAV-grouped upload handler
+/// actually writes were invisible to it and this repository's delete purged
+/// companions another repository's live parent artifact still referenced
+/// (#3156). Once this repository's attribution rows CASCADE away with it, that
+/// other repository becomes the key's sole read-path owner -- of bytes the
+/// delete has already destroyed.
 async fn collect_repo_maven_flat_keys(state: &SharedState, repo_id: Uuid) -> Vec<String> {
-    sqlx::query_scalar(
+    let sql = format!(
         "SELECT o.storage_key \
          FROM maven_flat_object_owner o \
          WHERE o.repository_id = $1 \
@@ -3973,24 +3984,25 @@ async fn collect_repo_maven_flat_keys(state: &SharedState, repo_id: Uuid) -> Vec
                JOIN repositories pr ON pr.id = pa.repository_id \
                WHERE pa.repository_id <> $1 \
                  AND pr.storage_backend = o.storage_backend \
-                 AND jsonb_typeof(am.metadata->'files') = 'array' \
-                 AND EXISTS ( \
-                     SELECT 1 FROM jsonb_array_elements(am.metadata->'files') f \
-                     WHERE f->>'storage_key' = o.storage_key \
-                 ) \
+                 AND {files_match} \
            )",
-    )
-    .bind(repo_id)
-    .fetch_all(&state.db)
-    .await
-    .unwrap_or_else(|e| {
-        tracing::warn!(
-            repo_id = %repo_id,
-            error = %e,
-            "Failed to list Maven flat-object keys to purge before repository delete"
-        );
-        Vec::new()
-    })
+        files_match = crate::services::maven_flat_attribution::metadata_files_name_key_sql(
+            "am.metadata->'files'",
+            "o.storage_key",
+        ),
+    );
+    sqlx::query_scalar(&sql)
+        .bind(repo_id)
+        .fetch_all(&state.db)
+        .await
+        .unwrap_or_else(|e| {
+            tracing::warn!(
+                repo_id = %repo_id,
+                error = %e,
+                "Failed to list Maven flat-object keys to purge before repository delete"
+            );
+            Vec::new()
+        })
 }
 
 /// Derived row-less Maven keys for a FILESYSTEM repository being deleted:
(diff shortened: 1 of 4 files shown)


brandonrc:
Review follow-ups — two applied, one **not** applied with evidence.

**1. Assertion that fails if `OR` becomes `AND`** — done (a5a137d). The test asserted
both spellings, containment, and no `jsonb_array_elements`; all three survive an
`OR`→`AND` flip. It now asserts the connective directly and pins the full rendered
fragment. Mutation-proved: flipping the production fragment fails the test; restored.
Worth noting this is the only guard-correctness check that runs without a database —
both DB-backed guard tests skip silently when `DATABASE_URL` is unset.

**3. Sidecar-base gap** — filed as #3188 with a reproduction, and linked from the PR body
in place of the "I will file it separately" promise.

**2. Scoping the three guards to Maven — I did not apply this, because it reintroduces
#3156 along a new axis.** The direction of harm inverts in a `NOT EXISTS` guard:

- A *spurious* match spares a key ⇒ storage leak. Benign.
- A *missed* match purges a key ⇒ data loss. This is #3156.

`AND am.format = 'maven'` narrows the anti-join, so it can only ever cause **more**
purging. It is the data-loss direction, and the guards should over-anchor, not
under-anchor.

Concretely, it would desynchronize the guards from the two things that define what
anchors a flat key — neither of which is format-scoped:

- **Migration 170** (and 163) backfills `maven_flat_object_owner` from *any*
  `artifact_metadata` row whose `files[]` names a `maven/%` key. It filters on the key
  prefix, not `am.format`.
- **The read path**, `OWNER_BY_METADATA_FILES_SQL`, has no format filter either.

So a key attributed by migration 170 through a non-`maven`-format parent row still
resolves to an owner on the read path, but a format-scoped guard loses the anchor and
purges it. Verified against a migrated schema — parent row with `format='sbt'` whose
`files[]` names `maven/com/example/demo/1.0.0/demo-1.0.0.pom`:

```
(1) migration 170 attributed the key despite format=sbt:  metadata_files
(2) READ PATH resolves an owner:                          <repo B>
(3) guard as merged (no format filter):  anchor_found = t   -> spared
(4) guard + AND am.format = 'maven':     anchor_found = f   -> PURGED
```

That is the same read-path/guard disagreement this PR exists to close, just along the
`format` axis instead of the spelling axis.

On the premise: I checked, and no other format writes a `files` array into
`artifact_metadata` at all. The `files` arrays in `pypi.rs` and `conan.rs` are HTTP
response bodies, not stored metadata. In the current tree `artifact_metadata.metadata->'files'`
has **no writer** — it is read-only (`sbom.rs`, `expand_maven_secondary_files`, the
attribution SQL). And even a hypothetical foreign-format row would need a member object
containing exactly `{"storageKey": "maven/..."}` matching a live attribution row, whose
only consequence would be sparing a genuine orphan.

If we do want format scoping, it needs to be applied **symmetrically** — read path,
guards, and a migration to clean up rows 163/170 mis-attributed — as its own change with
its own analysis, not a one-line addition to three delete guards. Happy to file that if
you want it; I didn't want to land the asymmetric half.

Gates (isolated Postgres 16): `cargo fmt --check` clean · `cargo clippy --workspace
--all-targets -- -D warnings` clean · `cargo test --lib -- maven_flat_attribution
storage_gc_service` 123 passed / 0 failed. No `sqlx::query!` macros touched, so no `.sqlx`
regeneration.

Issue #3078, opened 2026-08-03
Repository listing recomputes storage_used_bytes live (~10s on a 2M-artifact catalog) instead of reading the trigger-maintained usage ledger it ships with


ThaSami:
## Summary

`GET /api/v1/repositories` computes `storage_used_bytes` for every repository live, on every call:

1. a grouped `UNION ALL` SUM over `artifacts` + `proxy_cache_artifacts` + `oci_blobs` for the page's repositories (`backend/src/api/handlers/repositories.rs`, kept in lockstep with `RepositoryService::get_storage_usage`, #2218), and
2. for each **virtual** repository, `get_virtual_storage_usage` — a recursive CTE over the membership graph followed by the same three-component SUM over the union of every reachable leaf member's rows (#2785 / #2795).

On a deployment with ~2M live `artifacts` rows and 38 repositories (8 of them virtuals whose members include multi-million-row locals), the listing takes **~10 s per call** (9.86–10.27 s across repeated samples at `per_page=50`; ~6.4 s at the default page size). On the 1.4.x line — before either aggregation existed — the same endpoint on the same dataset returned in ~1 s.

The listing is the landing call for the web UI and a common first call for API clients, so this is paid constantly, and the cost scales with total artifact count rather than repository count.

## The same release already ships the O(1) answer

1.7.0's own migrations build a trigger-maintained per-repository usage ledger:

* migration 171 creates `repository_usage_ledger (repository_id PK, hosted_bytes, proxy_bytes, oci_bytes)` and seeds it with exactly the same three component sums;
* migration 182 installs row-level triggers on `artifacts`, `proxy_cache_artifacts`, and `oci_blobs` that keep it current, precisely so quota admission stops rescanning live artifact sizes ("O(1) quota admission on an authoritative ledger", per the 1.7.0 release notes).

The ledger's three components are by construction the same three branches the listing SUMs — including the `proxy-cache/%` legacy-key exclusion — so reading it is not an approximation:

* **non-virtual repository**: `hosted_bytes + proxy_bytes + oci_bytes` from one ledger row (primary-key probe);
* **virtual repository**: the existing recursive membership CTE, but joined against ledger rows of the DISTINCT leaves instead of aggregating their `artifacts`/`proxy_cache`/`oci_blobs` rows — the aggregation input shrinks from `O(total artifacts under members)` to `O(member repositories)`.

[...]

Issue #2785, opened 2026-07-21
Virtual repo: reported total does not match sum of child repos; cannot edit a virtual repo after creation


brandonrc:
Reported by customer Jon Craig (on AK 1.6.0) — 2026-07-21 feedback. Two virtual-repo defects: (1) a virtual repo shows it contains data but the total does NOT match the combined total of its child repos (aggregation/count bug); (2) a virtual repo cannot be EDITED after creation (members/config not updatable). Fix both. 1.6.x backport candidate.

PR #3162, opened 2026-08-07
fix(repositories): scope virtual repo storage total to members the caller can see

### LINKED ISSUE #3081: Virtual repository storage total discloses byte sizes of member repos the caller cannot read
A user granted access to a **virtual** repository but not its members sees the virtual's aggregate `storage_used_bytes` — which sums those members — while `GET` on each member repository correctly 404s. Neither the pre-existing live-SUM path nor the ledger-backed path introduced in #3079 ACL-filters virtual members before aggregating.

Live-repro'd during the #3078 red-team with a low-privilege user: the member repos 404, but the virtual reports their combined total.

**Pre-existing, not introduced by #3079** — the new code is byte-identical to the old oracle here. Filing so the disclosure is tracked on its own rather than riding along with a performance change.

Scope question for whoever picks it up: whether the virtual total should be ACL-filtered per-caller (correct but makes the figure caller-dependent and uncacheable) or whether virtual membership should be treated as implying read access to members' aggregate size.

brandonrc:
## Summary

Fixes #3081.

A virtual repository's `storage_used_bytes` was aggregated over **every** member reachable through the membership graph, with no per-caller authorization filter. A user granted the virtual parent but not its members therefore learned those members' exact byte sizes, even though `GET /api/v1/repositories/{member_key}` correctly 404s for them via `require_visible`. Both aggregation paths were affected:

* `RepositoryService::get_virtual_storage_usage` — the recursive walk behind `GET /api/v1/repositories/{key}` (and `PATCH`);
* `RepositoryService::get_virtual_storage_usage_batch` — the ledger-backed batch behind `GET /api/v1/repositories` (added in #3079, byte-identical to the older oracle here, so this is pre-existing and not a #3079 regression).

**Reproduced first**, against a bare migrated Postgres with a private virtual `v3081` (members `m3081a` = 500,000 B, `m3081b` = 250,000 B) and a user holding a `role_assignments` grant on the virtual only:

```
--- user_can_access_repo predicate for the user, per repo ---
  key   | is_public | can_access
--------+-----------+------------
 m3081a | f         | f
 m3081b | f         | f
 v3081  | f         | t

--- get_virtual_storage_usage (live SUM)          -> 750000
--- get_virtual_storage_usage_batch (ledger path) -> 750000
```

Both members are invisible to the caller; both aggregates hand back their combined size.

### The fix

Both functions now take the caller's `RepoVisibility` and filter the `leaves` CTE through `build_visibility_clause_for` — the *same* predicate the repository listing already uses to decide which repositories a caller may see (public ∪ `role_assignments` ∪ fine-grained `permissions`, direct/group/project), rather than a new hand-rolled check. A member contributes bytes only when the caller could have read it directly.

* Only the **leaves** are filtered. The membership walk is unchanged, so an invisible intermediate virtual still contributes the leaves the caller *can* see.
* `RepoVisibility::All` renders the literal `true` clause, so admin callers — and the internal live-SUM oracles used by the #3078 equivalence tests — get the byte-identical pre-fix aggregate.
* A repo-scoped API token gets `RepoVisibility::Ids`, matching what the listing already does for that principal.
* The derivation of the caller's visibility is extracted out of `list_repositories` into `visibility_for_auth` and shared by the detail, update, and listing handlers, so the three cannot drift.

Patch-shaped, per the 1.7.2 target: **no schema migration, no API break, no new endpoint or field.** The only behaviour change is that the value of an existing field is now correct for non-admin callers. Cost: the batch query gains one `EXISTS` per resolved leaf, i.e. O(member repositories), not O(artifacts) — it does not undo the #3078/#3079 N+1 work.

### Adjacent leaks found while checking siblings — NOT fixed here

Verified against source, reported so they get their own issues:

1. **`GET /api/v1/repositories/{key}/members`** (`backend/src/api/handlers/repositories.rs:8297`) — the parent gate is `require_repo_access` (token scope) and the per-member filter is `auth.can_access_repo`, also token scope. The handler's own comment at :8342 says so: *"Tokens with allowed_repo_ids = None (admins, JWT sessions, unrestricted API tokens) see everything by virtue of `can_access_repo` returning true."* For an ordinary JWT session the filter is a no-op, so any authenticated user can enumerate the key/name/type of every member of any virtual repo. Same class as this issue, wider blast radius (member identity, not just size).
2. **`GET /api/v1/repositories/{key}/artifacts`** on a virtual (`:4589`) and its Maven-grouped variant (`:5525`) — `proxy_helpers::fetch_virtual_members` (`proxy_helpers.rs:2623`) resolves members with no ACL predicate and feeds `list_for_repos_page` / `count_for_repos`; only the virtual parent is gated by `require_visible`. `proxy_helpers::authorize_virtual_members` exists for exactly this and is called only from `maven.rs` and `pypi.rs`.

Checked and **not** vulnerable: `GET /{key}/storage` and `/{key}/storage/tree` (both bound to `repo.id` against `repository_storage_stats` / `repository_path_storage_stats`, no member walk), and quota (`check_quota` → `get_ledger_usage(repo_id)`, single-repo ledger read).

Three in-crate tests (`src/api/handlers/repositories.rs`, so they count toward the `nextest --lib` coverage gate):

* `test_virtual_storage_total_excludes_invisible_members_3081` — HTTP-level, exercising **both** the detail walk and the listing batch and asserting they agree. Private virtual with a 1,000 B visible member and a 500,000 B hidden member:
  * caller with the virtual only → `0`
  * caller with the virtual + one member → `1,000` (**control against an always-zero "fix"**: the visible member's bytes are still counted)
  * admin → `501,000` (**positive control**: a caller who can see everything still gets the full, correct total)
  * plus the premise: `GET` on each invisible member returns 404 for that caller.
* `test_public_virtual_total_excludes_private_members_3081` — anonymous caller on a *public* virtual with one public and one private member → `3,000`, not `903,000`. Also exercises the `PublicOnly` bind shape of both queries.
* `test_visibility_for_auth_maps_each_principal_3081` — DB-free unit test pinning the anonymous / admin / user / repo-scoped-token / admin-scoped-token arms of `visibility_for_auth`.

**Revert-proved.** With the production guard disabled (`AND (true)` in place of the visibility clause) and the tests untouched:

```
test test_visibility_for_auth_maps_each_principal_3081 ... ok
test test_public_virtual_total_excludes_private_members_3081 ... FAILED
test test_virtual_storage_total_excludes_invisible_members_3081 ... FAILED

assertion failed: virtual total must not disclose bytes of members the caller cannot read
  left: 501000   right: 0
assertion failed: anonymous total must count the public member only, not the private one
  left: Some(903000)   right: Some(3000)
```

## Test Checklist
- [x] Manually tested locally

Gates run locally against a dedicated Postgres with all 178 migrations applied:

| Gate | Result |
|---|---|
| `cargo fmt --check` | clean |
| `cargo clippy --workspace --all-targets -- -D warnings` | clean |
| `cargo test --workspace --lib` | **14726 passed; 0 failed; 6 ignored** |
| `cargo test --lib services::repository_service::` (DB required) | 125 passed; 0 failed |
| `cargo test --lib api::handlers::repositories::` (DB required) | 469 passed; 0 failed |

The existing `#3078` equivalence matrix still passes: its oracles now pass `RepoVisibility::All`, which generates the identical unfiltered SQL. Three repeated seeding blocks in the virtual-storage tests (`virtual_repo_members` insert, `oci_blobs` seed, repo cleanup) were extracted into shared helpers in the same commit so the new tests do not add near-duplicates for the jscpd gate.

## API Changes
- [x] N/A - no API changes

No schema change, no route change, no response-shape change. `RepositoryService::get_display_storage_usage` / `get_virtual_storage_usage` / `get_virtual_storage_usage_batch` gain a `&RepoVisibility` parameter (internal Rust API only).

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -218,6 +218,33 @@ pub(crate) async fn require_repo_admin(
     }
 }
 
+/// Resolve the caller's repository visibility (issue #3081).
+///
+/// This is the single derivation of [`RepoVisibility`] from an authenticated
+/// (or anonymous) principal, shared by the repository listing and by every
+/// path that aggregates over OTHER repositories on the caller's behalf — today
+/// the virtual-repository storage total, whose members are separate
+/// repositories with their own ACLs.
+///
+/// * anonymous              -> public repositories only
+/// * global admin           -> everything
+/// * repo-scoped API token  -> exactly the token's allowed set (checked before
+///   the general user arm; admin tokens are handled above and bypass scope)
+/// * any other principal    -> public repositories plus their own grants
+pub(crate) fn visibility_for_auth(auth: Option<&AuthExtension>) -> RepoVisibility {
+    match auth {
+        None => RepoVisibility::PublicOnly,
+        Some(a) if a.is_admin => RepoVisibility::All,
+        Some(a) if matches!(a.allowed_repo_ids, AccessScope::Restricted(_)) => RepoVisibility::Ids(
+            a.allowed_repo_ids
+                .as_allowed_repo_ids()
+                .unwrap_or_default()
+                .to_vec(),
+        ),
+        Some(a) => RepoVisibility::User(a.user_id),
+    }
+}
+
 /// Ensure a repository is visible to the current user.
 ///
 /// Public repos are visible to everyone. Private repos require authentication
@@ -2184,21 +2211,10 @@ pub async fn list_repositories(
         .map(|t| parse_repo_type(t))
         .transpose()?;
 
-    let visibility = match &auth {
-        None => RepoVisibility::PublicOnly,
-        Some(a) if a.is_admin => RepoVisibility::All,
-        // Repo-scoped token: the listing must reflect ONLY the token's
-        // allowed repositories, not every repo the owning user can reach.
-        // Checked before the general `User` arm. Admin tokens are handled
-        // above and bypass scope restrictions.
-        Some(a) if matches!(a.allowed_repo_ids, AccessScope::Restricted(_)) => RepoVisibility::Ids(
-            a.allowed_repo_ids
-                .as_allowed_repo_ids()
-                .unwrap_or_default()
-                .to_vec(),
-        ),
-        Some(a) => RepoVisibility::User(a.user_id),
-    };
+    // The listing must reflect ONLY what this caller may see — including, for
+    // a repo-scoped token, only the token's allowed repositories rather than
+    // every repo the owning user can reach.
+    let visibility = visibility_for_auth(auth.as_ref());
     let service = RepositoryService::new(state.db.clone());
     let (repos, total) = service
         .list(
@@ -2206,7 +2222,7 @@ pub async fn list_repositories(
             per_page as i64,
             format_filter,
             type_filter,
-            visibility,
+            visibility.clone(),
             query.q.as_deref(),
             query.project,
         )
@@ -2272,9 +2288,12 @@ pub async fn list_repositories(
         for id in &virtual_ids {
             storage_map.insert(*id, 0);
         }
+        // #3081: aggregate only over the members THIS caller may see — a
+        // member the caller cannot read must not disclose its byte size
+        // through the virtual's total.
         storage_map.extend(
             service
-                .get_virtual_storage_usage_batch(&virtual_ids)
+                .get_virtual_storage_usage_batch(&virtual_ids, &visibility)
                 .await?,
         );
     }
@@ -2793,7 +2812,11 @@ pub async fn get_repository(
     require_visible(&repo, &auth, &service).await?;
     // #2785: virtual repos report the union of their members' contents, not
     // their own (empty) rows, so the displayed total matches the child repos.
-    let storage_used = service.get_display_storage_usage(&repo).await?;
+    // #3081: that union is scoped to the members THIS caller may see, so the
+    // aggregate cannot disclose the size of a member whose own GET 404s.
+    let storage_used = service
+        .get_display_storage_usage(&repo, &visibility_for_auth(auth.as_ref()))
+        .await?;
     let auth_type =
         crate::services::upstream_auth::get_upstream_auth_type(&state.db, repo.id).await?;
 
@@ -3621,7 +3644,10 @@ pub async fn update_repository(
     }
 
     // #2785: virtual repos report the union of their members' contents.
-    let storage_used = service.get_display_storage_usage(&repo).await?;
+    // #3081: scoped to the members this caller may see.
+    let storage_used = service
+        .get_display_storage_usage(&repo, &visibility_for_auth(Some(&auth)))
+        .await?;
 
     state.event_bus.emit_repository_event(
         "repository.updated",
@@ -21110,7 +21136,7 @@ mod apt_validation_tests {
         for (id, tag, expected) in matrix {
             let oracle = if tag == "virt" || tag == "hollow" {
                 service
-                    .get_virtual_storage_usage(id)
+                    .get_virtual_storage_usage(id, &RepoVisibility::All)
                     .await
                     .expect("virtual oracle")
             } else {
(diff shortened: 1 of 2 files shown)


brandonrc:
Review round on `16dab50`. Bug verified at merge-base — `get_virtual_storage_usage` took no caller parameter at all, so an **unauthenticated** caller could `GET` a public virtual and read the exact combined byte size of its private members. The fix uses `build_visibility_clause_for` / `visibility_for_auth`, not `can_access_repo`, and the `User` arm is byte-identical to `user_can_access_repo` — so for browser sessions and unrestricted tokens the leaf filter exactly mirrors `require_visible`. That part is right, and the regression test asserts **exact** sums (0 / 1,000 / 501,000) with genuine controls in both directions rather than "non-zero".

Holding on one thing, which is the same root cause as #3173's blocker: **the `Ids` arm is token scope, not visibility**, and here it produces a new wrong number in both directions.

**Under-report — a regression this PR introduces.** Virtual `maven-all` = public `maven-central` (8 GB) + private `internal-libs` (2 GB, user granted). CI holds a token scoped to `[maven-all]` — the natural way to scope a token for a virtual. Visibility is `Ids([maven-all])`, and `maven-all` is itself virtual so it's excluded by `leaf.repo_type <> 'virtual'` before the clause even runs. **Reported total: 0 bytes**, where it was correctly 10 GB. The sharpest form: because the `Ids` arm drops the `is_public` arm, an authenticated scoped token now reports a *smaller* total than an anonymous caller would for the same public virtual.

**Leak — the class survives for that principal.** Token scoped to `[internal-libs]`, minted while the user had a grant; the grant is later revoked. `GET /internal-libs` correctly 404s. `GET /maven-all` still reports its 2 GB, tracking uploads in real time. Scope is a mint-time snapshot; the aggregate re-derives entitlement from the snapshot rather than from current grants.

The `Ids` arm was imported from the listing, whose documented contract is "reflect only the token's scope" — a genuinely different question from "which members may this caller read." Correct leaf predicate: `leaf.is_public OR (leaf.id = ANY($scope) AND (<role_assignments> OR <permissions>))`. Since #3173 needs the same thing, it's worth adding one `RepoVisibility::ScopedUser { user_id, scope }` variant and using it in both.

**The byte channel isn't actually closed, and I'd rather say so than let #3081 close on it.** Run this PR's own fixture: the `outsider` who now correctly gets `storage_used_bytes: 0` can call `GET /repositories/{virt}/artifacts` and receive every artifact of the hidden member individually — `size_bytes`, `name`, `path`, `version` — plus an exact row count via `?count=exact`. Summing that reconstructs precisely the 500,000 the fix removed, and polling it reveals upload activity better than the total ever did. `require_visible` gates only the parent. The body discloses this accurately and doesn't overclaim — my concern is only that shipping it as "fixes #3081" will read as the channel being closed.

Test gaps, each with the mutation that survives:
- Change the `Ids` arm to return `"true"` → green. No test uses a restricted token against the query. Highest-value addition.
- Delete `OR {perms}` from the `User` arm → green. Fixtures grant via `role_assignments` only, so the fine-grained/group/project branch is never exercised.
- **Remove the `virtual_repo_members` join entirely** — make `leaves` "every repo the caller can see" — and the `outsider` (0) and `member` (1,000) assertions both still pass. Only the admin 501,000 assertion catches it, and only *incidentally*, via unrelated repos left in the shared DB. That's the missing negative test: add a visible, byte-holding repo that is **not** a member and re-assert 1,000, so membership is pinned by construction rather than by shared-DB accident.
- Reverting `update_repository` to `&RepoVisibility::All` → green; the PATCH response path is untested.

Also worth doing here since the PR's stated rationale is anti-drift: `repositories.rs:16009/16020/16031` each re-implement the visibility match inline and assert against their own **stale three-arm** copy with no `Ids` arm. The PR extracted `visibility_for_auth` "so the three cannot drift" and left a fourth copy that already has. `test_visibility_for_auth_maps_each_principal_3081` supersedes them — delete them. And `packages.rs:28-41` / `:45-50` are byte-identical clones of the two extracted helpers, which is relevant to the jscpd gate.

One operational note: the detail path still computes a live 3-way `UNION ALL` sum while the batch path reads the O(members) ledger. There's no index whose key list contains `size_bytes`, no plain `(repository_id, is_deleted)` btree, and the negated `storage_key NOT LIKE 'proxy-cache/%'` can't use `idx_artifacts_storage_key` — so a virtual fronting a large proxy makes a plain repository-detail `GET` scan millions of rows. Pre-existing (#2785), but this is the natural place to collapse the asymmetry.

**Merge-order: #3173 defines `visibility_for_auth` identically in this same file. Whichever lands second won't compile until rebased.**

Happy to push the `ScopedUser` variant plus the three tests across both PRs — it's one change made twice otherwise. Say the word and I'll take it.

brandonrc:
Pushed the fix for the shared `Ids`-arm defect. Both PRs are done from my side.

**The root cause, stated once.** `require_visible` is `is_public OR (in_scope AND (is_admin OR grants))`. The `Ids` arm is `in_scope` **alone** — it drops the public arm *and* the grant conjunct. That single gap produced opposite-direction bugs in the two PRs: #3173 discarded an admin's deliberately-scoped token (because `is_admin` was matched ahead of `Restricted`, and `can_access_repo` keys on the *scope*, never on the flag), and #3162 reported **0 bytes** where the correct answer was 10 GB.

**I did not change the shared `Ids` arm.** The repository listing's documented contract genuinely *is* "reflect only the token's scope" — a different question from "which members may this caller read." Instead both PRs now compose the two halves: `member_grant_visibility` answers the grant half in SQL, and in #3173 `member_passes_token_scope` answers the scope half per row. Together they are `require_visible` exactly.

**The one product call, made deliberately:** a token scoped to a *virtual* reads through to that virtual's members. The by-path download through the same virtual already serves member bytes to such a token — so without read-through the listing returns `200 OK` with zero items while the download of those same items succeeds. Two answers to one question. The grant half still applies on top, so this widens which members a token may **enumerate**, never which a user may **read**. It's documented on both helpers rather than left implicit.

**Both fixes are revert-proved, not argued.** #3173's new `AccessScope::Restricted` test fails under the revert with `got []` while all four pre-existing tests stay green — that arm had no coverage at all, which is why both defects shipped. #3162 gained the missing negative control (a visible, byte-holding repo that is *not* a member); removing the `virtual_repo_members` join now fails with `left: 8000, right: 1000`, where before only the admin assertion caught it, incidentally, via unrelated rows in the shared database.

Gates on both: `fmt --check` clean, `clippy --workspace --all-targets -D warnings` exit 0, 469/471 `repositories::` tests pass.

⚠️ **Merge order still applies and is now load-bearing:** both PRs define `member_grant_visibility` in `repositories.rs`. **Whichever lands second must drop its copy or it will not compile.** Noted in both commit messages.

Still open and yours: #3173's `caller_can_read_member` follow-up (rules-empty falls open, so member *content* is still reachable by path even though enumeration is now closed), and #3162's sibling artifact-listing channel — I'd keep #3081 open rather than close it on the total alone.

Issue #3154, opened 2026-08-07
Code Coverage job intermittently fails building aws-lc-sys (cc-rs), masquerading as a coverage-gate failure


brandonrc:
The `📊 Code Coverage` job intermittently fails while compiling the vendored C in `aws-lc-sys 0.39.0`, before it ever measures coverage. Observed twice today on different PRs:

| PR | run | job duration | failing step |
|---|---|---|---|
| #3027 | 31148018643 | 2m22s | Generate coverage report |
| #2922 | 31175059925 | 1m27s | Generate coverage report |

Both die the same way:

```
error occurred in cc-rs: command did not execute successfully (status code exit status: 1):
  LC_ALL="C" "cc" -O0 ... -c .../aws-lc/crypto/asn1/a_bitstr.c
error: command `cargo test --no-run --message-format json-render-diagnostics --workspace --lib` exited with code 101
```

(the exact translation unit differs between runs — `a_bitstr.c` in one, `bignum_montsqr_p384_alt.S` in the other, which is what suggests an environment/parallelism issue rather than a genuine compile error.)

**Why this is worth fixing rather than tolerating.** On the same commit, `🦀 Check Rust` and `🧪 Backend Unit Tests` both pass — they compile the same crate successfully. Only the coverage job fails, and it fails *fast*: ~1-2 minutes versus the ~14 minutes a real coverage run takes. A rerun clears it every time.

The failure surfaces as `Code Coverage: fail`, which reads as "your new code is under the 70% threshold". The natural reaction is to go add tests that are not needed. The only reliable way to tell the difference is to open the log and check the job duration, which is not discoverable.

**Suggested fixes**, roughly in order of cheapness:

- Make the job distinguish the two cases — if the `cargo test --no-run` build step fails, fail with a message that says "coverage build failed" rather than letting it land under the coverage-gate check name.
- The coverage job runs under `cargo llvm-cov`, which sets different `RUSTFLAGS`/`CFLAGS` than the other jobs and therefore does not share their build cache — so it rebuilds `aws-lc-sys` from scratch every run while the others hit cache. Giving the coverage job its own warm cache key would both fix the flake surface and cut several minutes off the job.
- If it is a parallelism/OOM issue in the C build specifically, capping `CARGO_BUILD_JOBS` or `make` parallelism for that job would confirm it.

[...]

Issue #3129, opened 2026-08-05
Flaky: storage-GC footprint test asserts a cluster-wide oci_blobs delta, races concurrent row deletes


brandonrc:
`services::storage_gc_service::tests::test_oci_blob_footprint_report_executes_dedups_and_ages`
is flaky under CI parallelism. It failed on the `main` run at 2026-08-04T20:39Z:

```
panicked at backend/src/services/storage_gc_service.rs:6908:9:
assertion failed: r1.total_blob_rows - r0.total_blob_rows >= 5
```

and **passed on a re-run of the identical commit**, so it is a race, not a deterministic break.

**Cause.** The test asserts on a *cluster-wide delta*: it reads `oci_blob_footprint_report()`
into `r0`, seeds 5 `oci_blobs` rows across four repos, reads `r1`, then asserts
`r1.total_blob_rows - r0.total_blob_rows >= 5`. `total_blob_rows` counts every row in the table,
not just the seeded ones. Any test that DELETES `oci_blobs` rows between the two reads drives the
delta below 5 and fails the assertion, even though the seeding worked perfectly.

`.config/nextest.toml` places `test(storage_gc_service::tests)` in the `db-serial` group, but
that only serializes those tests **against each other** — not against the many `oci_v2` handler
tests that insert `oci_blobs` rows and then delete them in `cleanup_all`. Those run concurrently
at the CI thread count.

**Why now.** The fragility is pre-existing (the monolithic-upload tests have always churned
`oci_blobs`), but two tests added in #3125 for cross-repo blob mount create and delete rows in
the same table, increasing the collision surface enough to make it manifest. Worth noting the
same class of latent global-state coupling likely affects the other delta assertions in this
test (`distinct_digests`, `logical_bytes`, `physical_bytes`).

**Suggested fix.** Make the assertions independent of global table state rather than widening the
serial group (which would slow CI). The seeded digests are locally generated UUIDs, so the test
can assert on exactly those — e.g. count rows `WHERE digest IN (shared_fs, only_a, shared_cloud)`
— instead of differencing a cluster-wide total.

brandonrc:
Another occurrence, on a PR that cannot possibly have caused it — useful as a clean data point.

**PR #3165**, run `31188086025`, job `92897677627`:

```
Summary [ 440.051s] 13169/14723 tests run: 13168 passed (3 slow), 1 failed, 6 skipped
[...]

PR #3165, opened 2026-08-07
ci: run backend integration tests on backend-touching PRs

### LINKED ISSUE #3124: Integration tests never run on PRs, so a PR can break main and show green — this is how #3118 shipped
## Summary

`🔗 Backend Integration Tests` cannot gate a pull request. It only runs on pushes to `main` / `release/*`:

```yaml
# .github/workflows/ci.yml:910
if: (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/'))
    && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
```

So a PR can break every integration test and still show all-green. The breakage only appears *after* merge, on the `main` push — at which point `main` is red and the author has moved on.

## This is not hypothetical — it is exactly what happened

#3118 traced 11 days of red `main` to #2905. Checking that PR's own CI:

```
PASS      🧪 Backend Unit Tests
PASS      🦀 Check Rust
SKIPPING  🔗 Backend Integration Tests     <-- the job it was about to break
```

#2905 changed `audit_fire_and_forget` to spawn its INSERT rather than await it. That is a change to **audit-write timing**, and the tests that assert audit-write behaviour are integration tests. They were skipped on the PR, ran for the first time post-merge, and `main` went red on 2026-07-24 and stayed red through 2026-08-04.

The author did notice the async-timing hazard — #2905 added an `audit_count_eventually` polling helper for it. But because the integration suite never ran on the PR, there was no signal that one call site had been missed.

## Why it stayed broken for 11 days

Once `main` is red, it stops being a signal. Over the last 100 CI runs on `main`: **1 success, 51 failures**, the rest cancelled. Every one of those 51 failed on the same three tests. A permanently-red `main` is indistinguishable from a newly-red `main`, so a second, unrelated blocker (#3121, which stopped all image publishing) accumulated underneath it unnoticed.

## Compounding factor

`ci-complete` treats `integration-tests` as `allow_skip` — a skip **passes**:

```bash
allow_skip "integration-tests" "$RESULT_INTEGRATION"
```

That is correct for PRs given the `if:` above, but it means the only enforcement point is the post-merge `main` run, whose failure blocks nothing and notifies nobody.

## Options

1. **Run integration tests on PRs that touch backend code.** Most direct. Cost: a ~2 min Postgres-service job (it runs in ~115s today) on backend PRs. Given it just cost 11 days of red `main`, this looks like a good trade.
2. **Run on merge queue / `merge_group`** if enabling one, so it gates the merge without paying on every push.
3. **Alert on red `main`.** Whatever the gating choice, a `main` CI failure should notify someone — Slack, an auto-filed issue, or a check on the next PR. 51 consecutive failures should not be able to pass unremarked.
4. **Label-gated opt-in** (`run-integration`) for risky PRs. Cheapest, but relies on humans predicting which PRs are risky — #2905 would likely not have been labelled.

Recommend 1 + 3.

## Note for #3120

The fix for #3118 is #3120, and by this same gap **its own PR CI cannot exercise the tests it repairs** — they will first run post-merge. That PR was therefore verified locally instead: reverting only the change reproduces the exact CI panic on a clean database, and with it applied the suite passes 13/13 on two consecutive runs leaving zero stranded rows.

brandonrc:
## Summary

`🔗 Backend Integration Tests` could not gate a pull request. It was scheduled only on pushes to `main` / `release/*`:

```yaml
# .github/workflows/ci.yml:910 (before)
if: (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
```

and `ci-complete` treated the resulting skip as a pass (`allow_skip "integration-tests"`). So a PR could break every integration test and still show all-green, with the breakage appearing only after merge on a `main` push whose failure blocks nothing and notifies nobody.

This PR:

1. **`changes`** (the existing docs-only gate) now also classifies the PR as backend-touching, from the same `pulls/files` listing it already fetches. `backend` is a strict subset of `code` — the docs arm matches first, so `backend/README.md` is docs, not backend. Backend paths are `backend/**`, `Cargo.toml`, `Cargo.lock`, `.sqlx/**`, and `.github/workflows/ci.yml` (so a change to this gate exercises the gate).
2. **`test-backend-integration`** keeps its main/release push arm byte-for-byte unchanged and gains a `pull_request` arm gated on `needs.changes.outputs.backend`. The suite, the service container, and every step are identical on both arms. It **fails open to running**: the PR arm tests `!= 'false'`, so if the `changes` job dies outright the suite runs anyway.
3. **`ci-complete`** **fails closed**: on a backend-touching PR a *skipped* integration job is now a hard failure rather than a pass. Everywhere else — docs-only PRs, version-bump-only release PRs, PRs touching no backend path, and pushes to branches other than `main`/`release/*` — a skip still passes.

`smoke-e2e` / `build-backend-image` are **also** effectively push-only (`detect-changes` is `if: github.event_name == 'push'`, despite its section header reading "CONTAINER BUILDS (PR only, ...)"). That is a separate hole with a much larger cost profile (a full image build per PR) and is deliberately **not** changed here; it wants its own issue.

## Cost analysis

Measured from the last 30 `push`-to-`main` CI runs via the Actions API — not estimated:

| Job | n | Mean duration |
|---|---|---|
| 🦀 Check Rust | 22 | 5.2 min |
| 🧪 Backend Unit Tests | 20 | 14.5 min |
| 📊 Code Coverage | 20 | 15.2 min |
| **🔗 Backend Integration Tests** | **21** | **~10 min** (10.3 success / 9.5 failure) |

Note this **corrects the estimate in #3124**, which put the job at "~115s". That is the test *execution* time; the job also compiles the integration test binaries from a clean checkout, and the real wall clock on `ak-ci-runners` is ~10 minutes.

**Wall-clock cost: ~0 added.** The job runs in parallel with the existing heavy jobs, and the PR critical path is already `max(5.2, 14.5, 15.2) ≈ 15.2 min` — longer than the ~10 min this job takes. A backend PR should finish in the same wall clock it does today, provided a runner slot is free.

**Runner cost: +~10 runner-minutes per backend PR run**, i.e. one more concurrent `ak-ci-runners` slot alongside the three heavy jobs already there. Peak concurrency observed across the last 40 CI runs is **11 simultaneous `ak-ci-runners` jobs**, so the fleet has headroom — but it is not free: per-job queue waits of up to 14.4 min (Check Rust) and 11.4 min (Coverage) already occur in that sample, so the fleet does saturate at times and this adds to the queue under load.

**The path filter is what keeps this affordable.** Of the last 54 merged PRs, **37 (69%)** touch `backend/**` / `Cargo.toml` / `Cargo.lock` / `.sqlx/**`; the other 31% (web, docs, scripts, IaC) skip the job entirely and pay nothing. Running the job unconditionally on every PR would be ~45% more runner-minutes than this for zero added signal.

Rejected alternatives: `merge_group` (no merge queue is enabled on this repo — a larger change); label opt-in `run-integration` (cheapest, but it relies on humans predicting which PRs are risky, and #2905 would not have been labelled).

The bug is in CI configuration, so the "test" is the gate itself. It is revert-proved below, and the `ci-complete` decision logic is exercised as a truth table.

### Evidence 1 — the gate fires on this very PR

This PR touches `.github/workflows/ci.yml`, which is in the backend path set, so `🔗 Backend Integration Tests` must run on this pull request. GitHub evaluates `pull_request` workflows from the head, so the new `if:` is the one in effect. See this PR's own checks — that is the live proof, and it is the check to look at before approving.

### Evidence 2 — it would have caught PR #2999

#2999 rewrote `execute_max_versions` in `lifecycle_service.rs`. Checks recorded at its original head **6ff1b253**:

```
skipped   🔗 Backend Integration Tests     <-- the job it was about to break
success   🧪 Backend Unit Tests
success   🦀 Check Rust
success   ✅ CI Complete
```

Running that head's own integration suite against a migrated Postgres — `cargo test --workspace --test lifecycle_policy_tests -- --ignored --test-threads=1`, exactly what the job runs:

```
running 22 tests
...
test test_max_versions_cascades_oci_tags ... FAILED

---- test_max_versions_cascades_oci_tags stdout ----
thread 'test_max_versions_cascades_oci_tags' (891843) panicked at backend/tests/lifecycle_policy_tests.rs:1331:5:
  left: 3
 right: 2

test result: FAILED. 21 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.65s
```

The PR's own base — `a791799` (#2993), its merge base with `main` — is green on the same suite, so the red was introduced by the PR and not inherited:

```
running 22 tests
test test_max_versions_cascades_oci_tags ... ok
...
test result: ok. 22 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.17s
```

Base green, head red. With this change that PR shows a red required check instead of a green one.

### Evidence 3 — the `ci-complete` gate, revert-proved

The `ci-complete` script was extracted from the workflow and executed under a matrix of job results. With this change:

```
backend PR, integration SKIPPED  (the #3124 hole)  backend=true  integ=skipped  -> FAIL (want FAIL)
backend PR, integration FAILED   (the #2999 case)  backend=true  integ=failure  -> FAIL (want FAIL)
backend PR, integration success                    backend=true  integ=success  -> PASS (want PASS)
non-backend PR, integration skipped                backend=false integ=skipped  -> PASS (want PASS)
docs-only PR, integration skipped                  backend=false integ=skipped  -> PASS (want PASS)
bump-only PR, integration skipped                  backend=true  integ=skipped  -> PASS (want PASS)
changes job died (backend empty), integ skipped    backend=      integ=skipped  -> FAIL (want FAIL)
push to main, integration success                  backend=true  integ=success  -> PASS (want PASS)
push to develop, integration skipped               backend=true  integ=skipped  -> PASS (want PASS)
```

All 9 match. Reverted to the script on `main`, the first case passes — the hole:

```
BEFORE fix: backend PR + integration skipped -> exit=0 (0 = green, i.e. the hole)
⏭️ integration-tests (skipped)
✅ **CI Passed**
```

### Evidence 4 — the suite is green on `main` today, so this does not immediately red-wall every PR

Turning on a gate that fails on arrival would be worse than leaving it off. The full job command list was run against `main` at `452b001` on a freshly migrated Postgres:

```
lifecycle_policy_tests               12 passed  0 failed
api_token_cache_flush_tests           8 passed  0 failed
grpc_sbom_tests                      13 passed  0 failed
search_reindex_tests                  7 passed  0 failed
scan_convert_to_reused_tests         23 passed  0 failed
stuck_scan_janitor_tests              1 passed  0 failed
virtual_members_atomicity_test        3 passed  0 failed
virtual_members_duplicate_test        5 passed  0 failed
oci_chunked_upload_cross_repo_tests   6 passed  0 failed
sso_oidc_login_callback_e2e_tests    13 passed  0 failed
sso_saml_acs_e2e_tests               16 passed  0 failed
cache_invalidation_fanout_tests       7 passed  0 failed
age_gate_tests                        5 passed  0 failed
scan_policy_validation_tests          4 passed  0 failed
                                    ---------
                                    123 passed  0 failed   (exit 0)
```

plus the job's second step, `incus_upload_tests::test_chunked_upload_cross_repo_session_rejected`: `1 passed; 0 failed`.

**Nothing is currently red.** Note this is a local run on a 20-core host, not `ak-ci-runners`; it does not rule out infrastructure flake on the fleet, only currently-failing tests.

## Test Checklist
- [x] Manually tested locally — `actionlint` reports 27 findings before and after this change, byte-identical (all pre-existing shellcheck warnings in the untouched `coverage` job); YAML parses; `ci-complete` truth table above
- [x] No regressions in existing tests — full integration suite green on `main` (Evidence 4)

## API Changes
- [x] N/A - no API changes

## Follow-ups deliberately not in this PR

- `smoke-e2e` / `build-backend-image` are push-only for the same structural reason (see Summary). Separate issue, separate cost argument.
- Option 3 in #3124 — alerting on a red `main` — is not addressed here. Gating PRs stops `main` going red on arrival; it does nothing about a `main` that is already red. Both are wanted.

Fixes #3124

## No branch-protection change needed

`main`'s required contexts are already:

```json
["🧪 Backend Unit Tests", "✅ CI Complete", "🦀 Check Rust"]
```

`✅ CI Complete` is required and gates on `test-backend-integration`, so the new enforcement flows through an existing required check. `🔗 Backend Integration Tests` does **not** need to be added to branch protection — and should not be, since it is legitimately skipped on non-backend PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -966,6 +966,13 @@ jobs:
     # peak RSS.
     env:
       CARGO_INCREMENTAL: "0"
+      # This job carries no rust-cache (see the sccache note below), so it
+      # resolves and downloads the whole dependency graph from crates.io on
+      # every run — more registry exposure than any other Rust job here. Now
+      # that a failure blocks a PR rather than just reddening a `main` push,
+      # a transient registry hiccup costs a developer a rerun, so let cargo
+      # retry the fetch itself before the step fails.
+      CARGO_NET_RETRY: "10"
     services:
       postgres:
         # Use the GHCR mirror to avoid Docker Hub anonymous-pull rate limits.
@@ -984,6 +991,38 @@ jobs:
 
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      # Resolve and download the dependency graph in its own step, ahead of the
+      # test steps, with a retry.
+      #
+      # Why: on 2026-08-07 this job failed on `main` (run 31176621357) with
+      #   error: failed to select a version for the requirement `axum = "^0.7"`
+      #          (locked to 0.7.9)
+      #   candidate versions found which didn't match: 0.6.6, 0.6.5, ...
+      # i.e. cargo saw a stale/truncated crates.io index. No test ever ran.
+      # That was the only non-test failure of this job on `main` in the last
+      # 13 runs, and under this PR the same hiccup would red a developer's PR
+      # instead of a post-merge push.
+      #
+      # This retries the *resolution and download* only. It deliberately does
+      # NOT wrap `cargo test`: retrying the test steps would mask exactly the
+      # real regressions this gate exists to catch. A genuine dependency
+      # problem still fails after the last attempt.
+      #
+      # No `--locked`: the test steps below do not use it either, so adding it
+      # here would newly fail PRs with a legitimately-updated Cargo.lock.
+      - name: Fetch dependencies
+        run: |
+          set -euo pipefail
+          for attempt in 1 2 3; do
+            if cargo fetch; then
+              exit 0
+            fi
+            echo "cargo fetch failed (attempt ${attempt}/3) — retrying" >&2
+            sleep $(( attempt * 10 ))
+          done
+          echo "ERROR: cargo fetch failed after 3 attempts" >&2
+          exit 1
 
       # The integration suites under `backend/tests/` insert into real tables
       # (users, api_tokens, repositories, ...) via the SQLx pool. The service
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,7 +40,7 @@ env:
 
 jobs:
   # ════════════════════════════════════════════════════════════════════════════
-  # DOCS-ONLY GATE
+  # DOCS-ONLY / BACKEND-CHANGE GATE
   # ════════════════════════════════════════════════════════════════════════════
 
   # Decides whether a PR touches anything beyond the docs paths that used to
@@ -57,6 +57,11 @@ jobs:
   # `code != 'false'` so an empty output also runs full CI. The file list
   # comes from the pulls/files API (server-side, same source GitHub's own
   # path filtering uses) — not from attacker-influenced checkout content.
+  #
+  # The same pass also classifies the PR as backend-touching or not (#3124).
+  # `backend` drives whether the integration suite runs on the PR; see
+  # `test-backend-integration` below. It is a strict subset of `code`: the
+  # docs arm is matched first, so `backend/README.md` is docs, not backend.
   changes:
     name: 🔍 Docs-Only Gate
     runs-on: ubuntu-latest
@@ -66,6 +71,7 @@ jobs:
       pull-requests: read
     outputs:
       code: ${{ steps.filter.outputs.code }}
+      backend: ${{ steps.filter.outputs.backend }}
     steps:
       - name: Determine whether non-docs files changed
         id: filter
@@ -74,25 +80,45 @@ jobs:
         run: |
           if [ "${{ github.event_name }}" != "pull_request" ]; then
             echo "code=true" >> "$GITHUB_OUTPUT"
+            echo "backend=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
           if ! files=$(gh api "repos/${GITHUB_REPOSITORY}/pulls/${{ github.event.pull_request.number }}/files" --paginate --jq '.[].filename'); then
             echo "PR file listing failed — failing open to full CI" >&2
             echo "code=true" >> "$GITHUB_OUTPUT"
+            echo "backend=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
           code=false
+          backend=false
           while IFS= read -r f; do
             [ -z "$f" ] && continue
-            # Keep in sync with the `push` trigger's paths-ignore list above
-            # (CLAUDE.md is already covered by *.md).
+            # First arm: keep in sync with the `push` trigger's paths-ignore
+            # list above (CLAUDE.md is already covered by *.md).
+            #
+            # Second arm: anything that can change what the integration
+            # suites compile to, what schema they run against, or how the
+            # job itself is defined. Keep in sync with `detect-changes`'
+            # `backend` filter, minus docker/Dockerfile.backend (the
+            # integration job compiles with cargo, it does not build an
+            # image) and plus .github/workflows/ci.yml (so a change to this
+            # gate exercises the gate).
             case "$f" in
-              site/*|*.md|.beads/*|LICENSE) ;;
-              *) code=true ;;
+              site/*|*.md|.beads/*|LICENSE)
+                ;;
+              backend/*|Cargo.toml|Cargo.lock|.sqlx/*|.github/workflows/ci.yml)
+                code=true
+                backend=true
+                ;;
+              *)
+                code=true
+                ;;
             esac
           done <<< "$files"
           echo "code=$code" >> "$GITHUB_OUTPUT"
+          echo "backend=$backend" >> "$GITHUB_OUTPUT"
           echo "non-docs changes: $code"
+          echo "backend changes: $backend"
 
   # ════════════════════════════════════════════════════════════════════════════
   # RELEASE-BUMP GATE
@@ -900,14 +926,41 @@ jobs:
           path: lcov.info
 
   # ════════════════════════════════════════════════════════════════════════════
-  # TIER 2: INTEGRATION (Main Branch Push Only)
+  # TIER 2: INTEGRATION (main/release push + every backend-touching PR)
   # ════════════════════════════════════════════════════════════════════════════
 
+  # Until #3124 this job was gated to `main` / `release/*` pushes only, so it
+  # could not gate a pull request: a PR could break every integration test and
+  # still show all-green, with the breakage surfacing only after merge on a
+  # `main` push that blocks nothing and notifies nobody. That is exactly how
+  # #2905 shipped — its own CI shows `🔗 Backend Integration Tests` SKIPPING
+  # next to a green `✅ CI Complete` — and `main` stayed red for 11 days
+  # (#3118).
+  #
+  # The push arm is unchanged. The new arm runs the same suite, unmodified, on
+  # pull requests that touch backend code (`needs.changes.outputs.backend`).
+  # Non-backend PRs — web, docs, scripts, IaC — still skip it and pay nothing.
+  #
+  # Fails open to RUNNING: the PR arm tests `!= 'false'`, so if the `changes`
+  # job dies outright and the output is empty the suite runs anyway. The
+  # matching enforcement is in `ci-complete`, which fails closed: on a
+  # backend-touching PR a skipped integration job is a hard failure, not a
+  # pass.
   test-backend-integration:
     name: 🔗 Backend Integration Tests
     runs-on: ak-ci-runners
     timeout-minutes: 40
-    if: (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/')) && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+    needs: [changes, release-bump]
+    if: >-
+      ${{ !cancelled()
+          && needs.release-bump.outputs.bump_only != 'true'
+          && (
+               (
+                 (github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/heads/release/'))
+                 && (github.event_name == 'push' || github.event_name == 'workflow_dispatch')
+               )
+               || (github.event_name == 'pull_request' && needs.changes.outputs.backend != 'false')
+             ) }}
     # Disable incremental codegen (#1679): consistent with check-rust and the
     # other heavy Rust compile jobs. Useless in CI's clean checkout, inflates
     # peak RSS.
@@ -1337,6 +1390,7 @@ jobs:
           RESULT_BUILD_OPENSCAP: ${{ needs.build-openscap-image.result }}
           RESULT_SECURITY: ${{ needs.security-audit.result }}
           CODE_CHANGED: ${{ needs.changes.outputs.code }}
+          BACKEND_CHANGED: ${{ needs.changes.outputs.backend }}
           BUMP_ONLY: ${{ needs.release-bump.outputs.bump_only }}
         run: |
           set -euo pipefail
@@ -1404,7 +1458,25 @@ jobs:
           require_success "security-audit" "$RESULT_SECURITY"
 
           allow_skip "smoke-e2e" "$RESULT_SMOKE"
-          allow_skip "integration-tests" "$RESULT_INTEGRATION"
+
+          # Integration tests are REQUIRED on a pull request that touches
+          # backend code (#3124). A skip there is precisely the hole that let
+          # #2905 merge green and leave `main` red for 11 days (#3118), so it
+          # must fail rather than pass. Fails closed: an empty BACKEND_CHANGED
+          # (the `changes` job itself died) counts as backend-touching, which
+          # matches the job's own fail-open-to-running condition.
+          #
+          # A skip still passes everywhere else: docs-only and version-bump-only
+          # PRs, PRs that touch no backend path, and pushes to branches other
+          # than main / release/*, none of which schedule the job.
+          if [[ "$GITHUB_EVENT_NAME" == "pull_request" \
+                && "$docs_only" != "true" && "$bump_only" != "true" \
+                && "$BACKEND_CHANGED" != "false" ]]; then
+            require_success "integration-tests" "$RESULT_INTEGRATION"
+          else
+            allow_skip "integration-tests" "$RESULT_INTEGRATION"
+          fi
+
           allow_skip "build-backend-image" "$RESULT_BUILD_BACKEND"
           allow_skip "build-openscap-image" "$RESULT_BUILD_OPENSCAP"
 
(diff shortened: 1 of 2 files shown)


brandonrc:
### CI note: the red `🧪 Backend Unit Tests` on this PR is #3129, not this change

This PR touches only `.github/workflows/ci.yml` and `CLAUDE.md` — no Rust. The unit-test failure is the already-filed flake [#3129](https://github.com/artifact-keeper/artifact-keeper/issues/3129):

```
FAIL services::storage_gc_service::tests::test_oci_blob_footprint_report_executes_dedups_and_ages
panicked at backend/src/services/storage_gc_service.rs:69
assertion failed: r1.total_blob_rows - r0.total_blob_rows >= 5
```

Same test, same assertion, same cluster-wide-delta race described in #3129, which records it passing on a re-run of an identical commit. 13168/13169 passed. Re-run triggered.

**The job this PR is actually about passed.** `🔗 Backend Integration Tests` ran on a pull request for the first time and succeeded:

| | |
|---|---|
| Result | success |
| Duration | 14:35:00 → 14:45:42 = **10.7 min** |

Two things that confirms empirically rather than by argument:

1. **The ~10 min cost estimate in the description is right** (10.7 min measured), and #3124's "~115s" figure is low by ~5x.
2. **The new job is not on the critical path.** It finished at 14:45:42 while `🧪 Backend Unit Tests` was *still running*. Wall-clock cost to a backend PR is ~0, as claimed.

On runner isolation, since a reviewer may reasonably ask whether adding a fourth Postgres-service job worsens the #3129 race: it does not, and the pattern is not new. `test-backend-unit` and `test-backend-integration` each get their own service container on their own ephemeral ARC pod, and they already overlap in time on every `main` push today — e.g. run 31179808751:

```
🔗 Backend Integration Tests  12:49:19 -> 13:00:32  runner=ak-ci-runners-9526x-runner-lg6bm
🧪 Backend Unit Tests         12:56:43 -> 13:09:09  runner=ak-ci-runners-9526x-runner-9rsv5
```

Distinct pods, overlapping windows. This change extends an existing concurrency pattern to PRs; it does not create one.

brandonrc:
## Review — gate mechanics verified, one hardening commit pushed

Reviewed at `980b0b4`, focusing on the two ways a change like this goes wrong: a gate that does not gate, or one that bricks the queue. **Neither is present.** I re-derived the claims independently rather than taking the PR body's word for them.

### 1. Path filter — correct

Extracted the `case` arms verbatim and ran them against real paths. `case` globs match `/`, so `backend/*` covers the whole subtree:

| path | `code` | `backend` |
|---|---|---|
| `backend/src/api/handlers.rs` | true | **true** |
| `backend/tests/lifecycle_policy_tests.rs` | true | **true** |
| `backend/migrations/2026_x.sql` | true | **true** |
| `backend/Cargo.toml` | true | **true** |
| `Cargo.toml` / `Cargo.lock` | true | **true** |
| `.sqlx/query-abc.json` | true | **true** |
| `backend/.sqlx/query-abc.json` | true | **true** |
| `.github/workflows/ci.yml` | true | **true** |
| `web/src/App.tsx`, `scripts/foo.sh`, `docker/Dockerfile.backend` | true | false |
| `docs/guide.md`, `README.md`, `backend/README.md`, `site/*`, `LICENSE`, `.beads/*` | false | false |

`.sqlx/**` and `Cargo.lock` are both covered, so a dependency bump cannot slip through ungated. No docs-only path sets `backend`. Parity with `detect-changes`' `backend` filter is exactly as documented (minus `docker/Dockerfile.backend`, plus `ci.yml`).

Checked for gaps: `workspace.members = ["backend"]` only, and there is no `rust-toolchain.toml` and no `.cargo/config.toml` — so there is no other path that changes what this suite compiles to. Nothing missing.

### 2. Required, not advisory — confirmed

Branch protection on `main` is `["🧪 Backend Unit Tests", "✅ CI Complete", "🦀 Check Rust"]`. `ci-complete` already has `test-backend-integration` in `needs`, so enforcement flows through an existing required context. No `continue-on-error` anywhere on the path. The PR is right that `🔗 Backend Integration Tests` should *not* be added to branch protection directly.

### 3. Skipped-vs-success — traced, and this is the part that most often bricks a queue

`ci-complete` is `if: always()`, so a skipped `test-backend-integration` cannot skip it and cannot leave a required check pending forever. I extracted the real `ci-complete` script and ran it under a matrix — including three cases beyond the PR's own table:

```
backend PR, integration SKIPPED  (the #3124 hole)   -> FAIL (want FAIL) ok
backend PR, integration FAILED   (the #2999 case)   -> FAIL (want FAIL) ok
backend PR, integration success                     -> PASS (want PASS) ok
backend PR, integration CANCELLED                   -> FAIL (want FAIL) ok   <- added
non-backend PR (web), integ skipped                 -> PASS (want PASS) ok
docs-only PR, integ skipped                         -> PASS (want PASS) ok
bump-only PR, integ skipped                         -> PASS (want PASS) ok
changes job DIED (backend empty), integ skipped     -> FAIL (want FAIL) ok
push to main, integ success                         -> PASS (want PASS) ok
push to main, integ FAILED                          -> FAIL (want FAIL) ok   <- added
push to develop, integ skipped                      -> PASS (want PASS) ok
workflow_dispatch feature branch, integ skipped     -> PASS (want PASS) ok   <- added
```

12/12. And revert-proved — swapping the new block back for `main`'s unconditional `allow_skip` reproduces the hole:

```
REVERTED (main's logic): backend PR + integration skipped -> exit 0 => PASS  <-- #3124 reproduced
```

The `!cancelled()` / `needs` interaction is also right: a docs-only PR skips the job *and* passes `ci-complete`, while a dead `changes` job runs the suite (job `if` fails open to running) *and* requires it (`ci-complete` fails closed). Those two halves agree, which is the thing that usually gets mismatched.

### 4. Runtime and cost

Concurrency is already handled — `group: ci-${{ github.event_name }}-${{ github.ref }}` with `cancel-in-progress` true for `pull_request`, so superseded PR runs are cancelled and a busy day does not stack runners. The wall-clock argument holds: this job (~10 min) runs in parallel and sits under the existing `max(check-rust, unit, coverage) ≈ 15.2 min` critical path.

### 5. Flakes — I checked this specifically, and the picture is better than it first looks

Pulled the last 60 `push`-to-`main` runs and read the logs. Raw totals look alarming — 34 failure / 11 success / 13 cancelled — but they are not what they seem, and **ordering is decisive**:

- The long red streak (2026-07-29 → 08-04) is the real `main` breakage, mostly `test_oidc_callback_map_groups_sync_and_prune`, `..._map_groups_disabled_creates_no_groups`, `..._emits_audit_records`. That is the #3118 outage this PR exists to prevent, not flake.
- The two SAML failures (08-05, 08-06 `test_saml_acs_emits_audit_records`) both landed **before** `e3f8e44` "poll for detached SAML audit writes" (#3133, merged 2026-08-06T02:05Z). They are the pre-fix state, not a live flake.

Post-`e3f8e44`, the suite is clean: **12 success / 1 failure across 13 runs**, and the single failure is not a test at all — run `31176621357` died in dependency resolution:

```
error: failed to select a version for the requirement `axum = "^0.7"` (locked to 0.7.9)
candidate versions found which didn't match: 0.6.6, 0.6.5, ...
```

So: **the two flakes named as live in this repo do not touch this gate.** `test_oci_blob_footprint_report_executes_dedups_and_ages` (#3129) is a `#[cfg(test)]` unit test in `backend/src/services/storage_gc_service.rs` — it runs in 🧪 Backend Unit Tests (already required) and Code Coverage, not in this job's `--test` list. The `aws-lc-sys` coverage failure (#3154) is in Code Coverage, which `ci-complete` treats as non-blocking.

**I do not think this needs to land advisory first.** The suite is green, the one failure class is infrastructure, and the alternative is another 11-day red `main`.

### Pushed: `dcb4a4a`

One hardening commit for that single failure class. This job carries no rust-cache, so it hits crates.io on every run — more registry exposure than any other Rust job here — and once it gates PRs a registry hiccup costs a developer a rerun instead of reddening a post-merge push.

- `CARGO_NET_RETRY: "10"`
- a separate `Fetch dependencies` step running `cargo fetch` with 3 attempts and backoff

Deliberately scoped: the retry covers **resolution and download only, never `cargo test`** — retrying the test steps would mask exactly the regressions this gate exists to catch. No `--locked`, since the test steps do not use it and adding it would newly fail PRs with a legitimately-updated `Cargo.lock`.

`python3 -c "import yaml; yaml.safe_load(...)"` parses; `shellcheck -s bash` clean on the new script; the two findings on the pre-existing `changes` / `ci-complete` scripts are artifacts of stubbing out `${{ }}` and `env:` vars, not real.

### Not blocking, worth a follow-up issue

The stale-index failure mode itself is unexplained — a truncated sparse index suggests either a corrupt cached index on a persistent `ak-ci-runners` home or a transient upstream serving issue. I did not add a cache purge, because if runner homes are shared across the ~11 concurrent jobs seen here, purging would race other jobs. Worth pinning down separately; the retry above is mitigation, not a root-cause fix.

I have not merged and have not applied `ready-to-merge`.

brandonrc:
### Live gate confirmation + one operational note (not a defect in this PR)

**The gate fires on this PR.** `🔍 Docs-Only Gate` on `dcb4a4a` logged:

```
non-docs changes: true
backend changes: true
```

and `🔗 Backend Integration Tests` is **queued**, not `skipping`. That is Evidence 1 from the description, confirmed live on the current head.

**I could not get it to green, because the `ak-ci-runners` fleet is currently stalled.** On this run every `ubuntu-latest` job completed while all four self-hosted jobs sit queued:

```
completed  success  🔍 Docs-Only Gate        (ubuntu-latest)
completed  success  🔍 Release-Bump Gate     (ubuntu-latest)
completed  success  🔒 Security Audit        (ubuntu-latest)
completed  success  🐚 Shell Tests           (ubuntu-latest)
queued     -        🦀 Check Rust            (ak-ci-runners)
queued     -        🧪 Backend Unit Tests    (ak-ci-runners)
queued     -        📊 Code Coverage         (ak-ci-runners)
queued     -        🔗 Backend Integration Tests (ak-ci-runners)
```

Repo-wide there are **zero runs in progress and 9 queued**, the oldest from `2026-08-08T02:29Z`. So this is a stalled fleet, not saturation, and it is unrelated to this change.

To be clear about the blast radius, since it would be easy to over-read: this does **not** make the fleet outage worse. `🦀 Check Rust` and `🧪 Backend Unit Tests` are already required and already run on `ak-ci-runners`, so a stalled fleet already blocks every backend PR with or without this PR. The only practical consequence is merge timing — this PR's own required checks cannot go green until the fleet recovers.

Two asks before merge:

1. Recover the fleet, then confirm `🔗 Backend Integration Tests` goes green on this PR — that is the one piece of live validation still outstanding, including for the `Fetch dependencies` step I added in `dcb4a4a`, which has not yet executed anywhere.
2. Independently, the deferred-scope claim in the description checks out: `detect-changes` is `if: github.event_name == 'push'`, `build-backend-image` needs its `backend` output, and `smoke-e2e` needs `build-backend-image` to have succeeded — so both are genuinely push-only on PRs and correctly left for a separate issue.

Issue #2270, opened 2026-07-07
Proxy-cached artifacts have no DB representation -- storage accounting, download tracking, and cache visibility are broken


mymarche:
### Description

Since PR #1280 (fix for #1278), proxy-cached artifacts have no database representation. Three problems result:

1. **storage_used_bytes = 0 for remote repositories** — the per-repo storage accounting query (`SUM(artifacts.size_bytes)`) misses all proxy-cached content. Remote repos always report 0 even with gigabytes cached. (#2218)

2. **Download tracking impossible for proxy-served artifacts** — `DownloadTracker::record_download()` requires an `artifact_id`, but proxy-cached items have no row in the `artifacts` table. (#2260)

3. **No cache visibility** — operators cannot list cached entries, check their age or size, or apply lifecycle policies. The only way to inspect the proxy cache is to walk the S3/filesystem storage directly.

PR #2051 is a partial workaround for Maven — it indexes cached artifacts into the `packages`/`package_versions` catalog so they appear in the Maven component listing. But it does not fix storage accounting or download tracking.

The PR description for #1280 called out: *"If a future feature wants proxy-cache listing/UX, it should be built via a separate `proxy_cache_artifacts` view or table."*

### Environment Information

- Deployment type: any (all deployments with remote/proxy repositories)
- Operating System: N/A
- Architecture: N/A
- Database: all PostgreSQL versions
- Reverse Proxy: N/A

### Backend Version

1.2.1+ (regression introduced in #1280)

### Steps to Reproduce

1. Create a remote (proxy) repository
2. Download an artifact through it (so it gets cached)
3. Call `GET /api/v1/repositories/{key}` — observe `storage_used_bytes: 0` even though files exist in the proxy cache
4. Check `download_statistics` — no rows for the proxy-served download
5. No API endpoint exists to list what was cached

### Expected Behavior

- `storage_used_bytes` should include proxy-cached content size per repository
- Proxy-served downloads should appear in `download_statistics`
- Operators should be able to query proxy cache contents via API

### Actual Behavior

- `storage_used_bytes` is always 0 for remote repos (#2218)
- Proxy downloads are invisible in statistics (#2260)
- No cache query capability exists

### Additional Context

Related:
- **#2218** — Per-repo storage_used_bytes excludes proxy-cached storage
- **#2260** — Download statistics silently miss many artifact downloads
- **PR #2051** — Maven-only workaround via package catalog indexing
[...]

PR #3172, opened 2026-08-07
fix(proxy-cache): enforce known digests and bound streamed cache writes

### LINKED ISSUE #2928: No byte bound on streamed proxy-cache writes, and the one quota guard is unreachable (quota_bytes hardcoded None on every proxy path)
## Problem

Two separate gaps that compound:

1. The streaming proxy-cache write path has no per-object byte ceiling. The buffered path has one (`cache_classifier::exceeds_single_object_quota`); the streaming leader has no equivalent.
2. That buffered guard reads `repo.quota_bytes`, and every handler-driven proxy fetch synthesizes its `Repository` with `quota_bytes: None` hardcoded. So the one guard that exists is structurally unreachable on the paths that actually serve proxy traffic.

Net effect: an upstream response of any size is written to the proxy cache, on both paths, regardless of the repository's configured quota.

This is pre-existing. It was not introduced by #2918, #2919, or #2922.

> **Line references corrected.** Every `file:line` in the original filing was taken from a stale working tree (branch `feat/repo-export-import`), not from `main`. All references below are re-verified against `origin/main` at `ca87b456`. The analysis and conclusions are unchanged; only the numbers moved.

### The guard has exactly one call site, on the buffered path

`cache_classifier::exceeds_single_object_quota` is defined at `backend/src/services/cache_classifier.rs:478`. A repo-wide grep finds exactly one non-test call site: `backend/src/services/proxy_service.rs:2893`, inside `fetch_artifact_with_cache_path_and_accept_capped` (declared at `backend/src/services/proxy_service.rs:2745`).

```rust
let quota_ok = !cache_classifier::exceeds_single_object_quota(
    repo.quota_bytes,
    resp.content.len() as i64,
);
```

The function's own doc comment (`cache_classifier.rs:463-477`) is accurate about its scope: it is a Phase-1 interim single-object guard, not usage accounting, with full enforcement deferred to P4 because proxy-cached content is deliberately absent from the `artifacts` table (#1278, see also #2270).

### `quota_bytes` is hardcoded `None` on every handler proxy path

`build_remote_repo_with_format` at `backend/src/api/handlers/proxy_helpers.rs:4815` sets `quota_bytes: None` at `:4832`. `build_remote_repo` at `:4804` delegates to it.

Every buffered `proxy_fetch*` wrapper routes through `with_proxy_repo` at `backend/src/api/handlers/proxy_helpers.rs:528`, which builds the repo at `:540`:

```rust
let repo = build_remote_repo(repo_id, repo_key, upstream_url);
```

That covers `proxy_fetch` (`:558`), `proxy_fetch_with_accept` (`:577`), `proxy_fetch_capped` (`:602`), `proxy_fetch_with_cache_key` (`:1299`), `proxy_fetch_with_cache_key_and_accept` (`:1326`), `proxy_fetch_capped_with_cache_key` (`:1351`), and `proxy_fetch_capped_with_cache_key_and_accept` (`:1377`). The streaming entry points build theirs directly: `proxy_fetch_streaming_with_cache_key_verified` at `:1468` and `proxy_check_cache_streaming` at `:1529`, both via `build_remote_repo_with_format`. The one non-`proxy_helpers` caller, `maven.rs:841`, also uses `build_remote_repo`.

`Repository` rows loaded from the database do carry the real column (`fetch_virtual_members` at `backend/src/api/handlers/proxy_helpers.rs:2534` selects `r.quota_bytes` at `:2546`), so the value is available; it is simply never threaded into the synthesized proxy repo.

### The streaming leader has no equivalent guard at all

`open_streaming_leader` at `backend/src/services/proxy_service.rs:3393-3489` opens upstream, classifies the TTL, and hands the body to `CachePersister::tee_stream` at `:3467`. There is no quota check, no size check, and `fetch_from_upstream_streaming` is called at `:3417` with no byte cap, unlike the buffered `fetch_from_upstream_with_accept`, which takes a `max` argument.

## What bounds actually exist

Stated precisely, because the memory bound is real and should not be confused with a byte bound:

- **Memory, per stream: bounded.** `tee_stream` (`backend/src/services/proxy_service.rs:1465`) splits each upstream chunk at `TEE_MAX_CHUNK_BYTES` = 64 KiB (`:312`, applied at `:1690`) into an mpsc channel of depth `TEE_CHANNEL_DEPTH` = 64 (`:298`, created at `:1483`). That caps in-flight bytes per stream at roughly 4 MiB. Unit tests at `:10315-10334` pin both constants specifically because they are the per-request OOM budget.
- **Total bytes transferred and written to disk: not bounded.** The chunk splitting preserves the total payload; it only changes the piece size. Nothing counts cumulative bytes or aborts the write.
- **Buffered path with an explicit cap: bounded.** The `*_capped` variants take a `max: usize` and pass it into `fetch_from_upstream_with_accept` (called at `proxy_service.rs:2823`, defined at `:5035`). This is a per-call-site cap chosen by the handler, not a repository policy, and it has no streaming counterpart.
- **Proxy-cache size cap: none.** No `max_cache_size` / cache-size-limit configuration exists in the backend.
- **Proxy-cache eviction: none.** `lifecycle_service::execute_size_quota` (`backend/src/services/lifecycle_service.rs:1189-1250`) does implement LRU size-quota eviction, but it queries `FROM artifacts WHERE repository_id = $1 AND is_deleted = false` (`:1207-1209`, `:1228-1235`). Proxy-cached content is deliberately not in the `artifacts` table (#1278), so proxy-cache entries are invisible to it. There is no eviction service that covers the proxy cache.

## Impact

A single upstream response of arbitrary size is written to proxy-cache storage. Repeated across paths, this fills the cache volume with no ceiling and no reclaim path, since nothing evicts proxy-cache entries.

The secondary effect is that the repository `quota_bytes` setting silently does nothing for remote repositories. An operator who configures a quota on a remote repo gets no enforcement and no warning.

## Fix

Three pieces, roughly in order of cost:

1. Thread a real `quota_bytes` into `build_remote_repo_with_format` (`backend/src/api/handlers/proxy_helpers.rs:4815`) from the source repository row, so the guard at `proxy_service.rs:2893` stops being inert. This is the smallest change and makes the existing buffered guard actually do what its doc comment says.
2. Add a per-object ceiling to the streaming write path. `tee_stream` already sees every chunk; accumulating a byte count and abandoning the cache write (while still serving the body to the client, matching the buffered path's behavior at `proxy_service.rs:2897-2912`) past a threshold is a contained change. `upstream.content_length` is already passed into `tee_stream` at `proxy_service.rs:3485` and can short-circuit when the upstream advertises a length.
3. Longer term, a proxy-cache size cap with eviction. This depends on the proxy-cache accounting work tracked by #1278 / #2270, since there is no per-repo proxy-cache total to check against today. Related: #2474 (project-scoped quotas and storage accounting).

## Severity note

Low, stated honestly. Exploiting this requires a hostile or compromised upstream, and configuring a Remote repository against an upstream is an explicit operator trust decision. It is not reachable by an unauthenticated client against a well-behaved upstream: a client cannot choose the response size, only which path is fetched.

This is not a regression. The 8 MiB ceilings recently removed from artifact paths were metadata caps misapplied to binaries; removing them was the bug fix, not the introduction of a gap. The accurate framing is that the streaming path never had the guard the buffered path nominally had, and the buffered guard was inert on every handler-driven fetch anyway.

The reason to file it rather than leave it implicit is that the `quota_bytes` field is user-configurable and reads as an enforced setting when it currently enforces nothing on remote repositories.

### LINKED ISSUE #2929: Streaming proxy cache commits without verifying a known digest, and commits truncated bodies when upstream omits Content-Length
## Problem

Two related gaps in the streaming proxy cache commit path. Both are pre-existing on `main` and neither was introduced by #2918, #2919, or #2922. All line numbers below are against `origin/main` at `ca87b456`.

1. The digest-gated commit machinery exists and is wired up on only two of the streaming proxy paths, including on paths where a SHA-256 for the content is already known locally.
2. When an upstream omits `Content-Length`, a body that is truncated mid-transfer commits to the cache as authoritative and is served as a warm hit from then on.

Finding 1 is the cheapest available mitigation for finding 2 on any path that can supply a digest, which is why they are filed together.

## Finding 1: digest gate is available but unused where a digest is known

`proxy_fetch_streaming_with_cache_key_verified` (`backend/src/api/handlers/proxy_helpers.rs:1468`) takes `expected_checksum: Option<String>` and threads it to `fetch_artifact_streaming_with_cache_path_gated`. The gate itself is at `backend/src/services/proxy_service.rs:1514`: on a `StreamWriteOutcome::Commit`, if an expected checksum was pinned and the streamed SHA-256 does not match, the object is deleted and the metadata sidecar is skipped, so the entry reads as a miss and the next request refetches.

Only two call sites pass a real value:

* `backend/src/api/handlers/oci_v2.rs:2879`
* `backend/src/api/handlers/debian.rs:2787`

Everything else reaches the same function through the wrapper at `backend/src/api/handlers/proxy_helpers.rs:1439`, which hardcodes `None`.

Two paths where a digest is already in hand and is not used:

**NuGet flatcontainer repair.** The SELECT at `backend/src/api/handlers/nuget.rs:899` fetches `checksum_sha256` (and `content_type`), and neither column is read anywhere in the function. The repair branch at `nuget.rs:1026` calls `get_cached_or_refetch`, whose refetch closure pulls from upstream via `proxy_fetch_capped` (`nuget.rs:1041`). That is the buffered helper, which has no digest gate at all, not even the optional one. `get_cached_or_refetch` then writes the bytes back to the row's own `storage_key` (`proxy_helpers.rs:1215`) without ever comparing them to `artifact.checksum_sha256`.

There is an authorization consequence worth stating precisely. `check_artifact_download(artifact.id)` runs at `nuget.rs:1008` and authorizes based on a row whose `checksum_sha256` and `size_bytes` describe a blob that is no longer present in storage. The bytes actually served, and written back under that row's storage key, come from a fresh upstream pull (or from a warm proxy-cache object keyed under `v3/flatcontainer/...`) that is never compared against the row. An admin who reviews an artifact and releases it from quarantine is approving a hash that is never enforced against what clients subsequently receive. Scan-on-proxy does not close this either: it is a documented no-op for proxy-cached content, since #1278 keeps that content out of `artifacts` (see `should_warn_proxy_scan_skipped` at `proxy_service.rs:616` and the message builder `proxy_scan_skipped_warning` at `:627` onward, plus the rationale on the former).

**Cargo remote crate download.** `backend/src/api/handlers/cargo.rs:905` calls the unverified wrapper, so it resolves to the verified function with `expected_checksum: None`. The sparse index entry for `{name}/{version}` served by this same handler already carries `cksum` (`build_index_entry`, `cargo.rs:1091`), which is what cargo itself verifies the downloaded `.crate` against.

To be accurate about what this does and does not buy: it cannot catch a lying upstream on a first fetch in the general case, because the digest and the bytes can come from the same compromised source. What it does is convert an already-recorded digest from decorative into enforced. For NuGet the digest is locally recorded and independent of the upstream that serves the repair. For cargo the index and the download can be different hosts (crates.io splits `index.crates.io` from the download host), so gating catches a compromised or misbehaving download host while the index is intact, and it catches ordinary corruption on either.

## Finding 2: a truncated body commits when upstream omits Content-Length

`classify_stream_write` (`backend/src/services/proxy_service.rs:580`) rejects a write whose byte count disagrees with the upstream `Content-Length`, and the reject arm deletes the object and skips the sidecar so the entry self-heals. But the length comparison only runs when `expected_len` is `Some`, and `expected_len` comes solely from the `Content-Length` header parsed in `extract_streaming_headers` (`proxy_service.rs:539`, the `content_length` field).

An upstream that frames by connection close, or that uses chunked transfer encoding and drops mid-body, produces `bytes_written > 0` with `expected_len == None`. `classify_stream_write` falls through to `StreamWriteOutcome::Commit` (`proxy_service.rs:591`), the commit arm at `proxy_service.rs:1502` writes the metadata sidecar, and the truncated artifact is served as a warm cache hit on every subsequent request. Only the zero-byte case is caught.

The existing doc comment on `classify_stream_write` is honest that a stripped `Content-Length` cannot be validated and must not be falsely rejected, which is correct as far as it goes. The gap is that there is no second signal, so nothing else catches the truncation.

This is systemic across every streaming proxy path, not specific to one format.

## Impact

* A truncated archive, a corrupted proxy-cache object, or bytes from a download host that disagrees with a locally recorded digest can be committed and then served warm indefinitely, until the TTL expires or the entry is manually evicted.
* For NuGet specifically, quarantine review and release operate on a hash that is never enforced against served bytes.
* Clients that verify independently (cargo checks `cksum`, OCI clients check the blob digest, `dotnet restore` will surface a corrupt `.nupkg`) will fail rather than silently accept bad content, so the practical outcome is usually a hard-to-diagnose repeated failure rather than a silent compromise. Clients that do not verify get the bad bytes.

## Fix

1. Pass the known checksum into the verified variant on the cargo path (`cargo.rs:905`), sourcing it from the sparse index entry for `{name}/{version}`.
2. On the NuGet repair path, compare the refetched bytes against `artifact.checksum_sha256` before the write-back at `proxy_helpers.rs:1215`, and fail the repair rather than persisting a mismatch. If that path is converted to streaming, route it through the verified variant instead.
3. Gate on the value matching `^[0-9a-f]{64}$` before using it. A row carrying a `sha256:`-prefixed or otherwise non-bare value must be treated as "no digest available" and fall through to the current behaviour. Otherwise it would never match, the entry would never cache, and every request would re-pull upstream.
4. For finding 2, consider a fallback signal when `Content-Length` is absent: for example, treat a stream that terminates with a transport error rather than a clean EOF as a reject, so a mid-body connection drop does not reach the commit arm. Finding 1 closes the truncation case only for paths that supply a digest, and there is no cheap general fix, so that narrower fix plus an explicit fallback is probably the realistic scope.

## Related, not part of this issue

There is currently no byte bound at all on streamed proxy-cache writes. `exceeds_single_object_quota` has a single production call site (`proxy_service.rs:2893`) on the buffered path only, and every handler-driven proxy fetch builds its `Repository` through `build_remote_repo_with_format` (`proxy_helpers.rs:4832`), which hardcodes `quota_bytes: None`, so even that guard is inert for these paths. Noted here for cross-reference; it is a separate concern and may be filed separately.

#2917 and #2921 cover the same NuGet repair site for different defects (wrong upstream URL, the 8 MiB cap, and the dangling `artifacts` row). The integrity gap described here is a third, independent defect at that site. If the in-flight repair work relocates or restructures that code, the digest check should land with it rather than against the current line numbers.

## Severity

Defense in depth, not a critical remote exploit. Both findings require either a compromised or misbehaving upstream, or write access to the storage backend, and the most likely real-world trigger is ordinary network truncation rather than an attacker. The NuGet quarantine interaction is the part most worth fixing on its own merits, because it makes an approval control weaker than it appears to be. Please do not treat this as a P0.

brandonrc:
## Summary

Fixes #2929 and Fixes #2928 — one family: a digest that is already recorded locally was never enforced against cached bytes, and the streamed cache write had no byte bound while the one guard that existed was unreachable.

### #2929 — enforce digests that are already in hand

**cargo.** The Remote `.crate` download called `proxy_fetch_streaming_with_cache_key`, the wrapper that hardcodes `expected_checksum: None`. This same handler serves the sparse-index entry carrying `cksum` — exactly what cargo verifies the download against — so the digest was decorative. It now resolves the index `cksum` for `{name}/{version}` and routes the download through the digest-gated `_verified` sibling. On a split-host registry (crates.io serves `index.crates.io` separately from the download host) this catches a misbehaving or compromised download host while the index is intact, and it catches ordinary corruption or a clean-EOF truncation on either. A body that fails the gate is still streamed to the client, which verifies independently; it is only kept out of the cache.

**nuget.** The missing-file repair refetched from upstream and wrote the bytes back under the artifact row's own storage key without ever comparing them to that row's `checksum_sha256` — a column the SELECT already fetched and nothing read. `check_artifact_download` authorises on that row, so an admin releasing an artifact from quarantine was approving a hash never enforced against what clients then received. `get_cached_or_refetch` now verifies before the write-back and fails the repair on a mismatch rather than persisting and serving it.

A digest is enforced only when it is a bare lowercase 64-hex SHA-256 (`normalize_expected_sha256`). A `sha256:`-prefixed, uppercase, or otherwise non-canonical value reads as "no digest available" and falls through to the previous behaviour. That fallback is load-bearing: a value the comparison can never match would reject the entry on every fetch, nothing would cache, and every request would re-pull upstream.

### #2928 — bound the streamed write, and make the quota reachable

`exceeds_single_object_quota` read `repo.quota_bytes` off the `Repository` that every handler-driven proxy fetch synthesizes via `build_remote_repo_with_format`, which hardcodes `quota_bytes: None`. `exceeds_single_object_quota(None, _)` is `false` unconditionally, so the guard was **structurally unreachable** on every path that serves proxy traffic — an operator who configured a quota on a Remote repo got no enforcement and no warning. The value is now resolved from the repository row by `repo.id`, in the same shape as the existing `get_cache_ttl_override`, rather than read off the synthesized field — so a future synthesizing helper cannot reintroduce the hole.

The streaming leader had no size check at all. `TEE_CHANNEL_DEPTH * TEE_MAX_CHUNK_BYTES` bounds *memory* per stream (~4 MiB in flight); the chunk splitting preserves the whole payload, so nothing counted cumulative bytes. The resolved ceiling now applies on both sides:

* **pre-flight**, when the upstream advertises a `Content-Length` over the ceiling — the tee is skipped entirely and not one byte is written;
* **mid-stream** otherwise (the chunked / connection-close case, where no length is advertised) — `tee_stream` counts what it hands the writer and abandons the cache write the moment the count would pass the ceiling.

Enforcing mid-stream rather than after `put_stream` returns is the point: a post-hoc check only fires once the unbounded object has already landed on disk. The client is served in full either way, matching the buffered path's over-quota behaviour.

### Scope notes

* **No schema migration, no API break.** Patch-shaped.
* **#2929 Fix item 4** (a fallback reject signal when the stream ends with a transport error rather than a clean EOF) is **already implemented on main** and already tested — `test_tee_upstream_error_mid_stream_aborts_cache`. `put_stream` returns `Err` on a stream error, so the writer skips the sidecar. The irreducible residual is a *clean* EOF short of the true length with no `Content-Length`, which is indistinguishable from a complete response at the protocol level; only a digest catches it, which is what the #2929 half of this PR does for the paths that have one. Nothing was changed there.
* **#2928 Fix item 3** (a global proxy-cache size cap with eviction) is deliberately **not** in scope — it depends on the proxy-cache accounting work in #1278 / #2270, as the issue itself states. Repositories with no configured quota keep today's unbounded behaviour; a positive control test pins that.
* **Composition with #3153.** #3153 rewrites `tee_stream` to stage-then-publish. This PR builds on top of it and does not fight it: every edit here is outside every hunk #3153 touches. The signature gains one parameter, and the enforcement lives in the client-facing `async_stream::try_stream!` forward loop, which #3153 leaves untouched; #3153's changes are confined to the spawned writer task. The mechanism is also complementary — this abandons the write by making `put_stream` fail, which under #3153 fails against the staging key and never reaches the live key at all.

Both issues were reproduced against this PR's base (`452b001`) before any code was written. Real output from that run:

```
REPRO-2928-A synthesized repo.quota_bytes = None
REPRO-2928-A exceeds_single_object_quota(repo.quota_bytes, 1 TiB) = false
REPRO-2928-A exceeds_single_object_quota(Some(1 MiB), 1 TiB) = true   <- guard is correct; the value it is fed is not
REPRO-2928-B client bytes = 8388608 ; bytes written to cache = 8388608 ; sidecar writes = 1
REPRO-2929-2 classify_stream_write(11, None) = Commit
REPRO-2929-2 sidecar writes = 1 ; deletes = []
```

Every new guard was revert-proved — the guard reverted, the test re-run, and confirmed to fail:

| Guard reverted | Test that fails |
|---|---|
| `tee_stream` ceiling check removed | `test_tee_abandons_cache_write_past_the_object_ceiling` — *"an over-ceiling object MUST NOT write a metadata sidecar"* |
| `resolve_quota_bytes` → `repo.quota_bytes` | `test_resolve_quota_bytes_reads_the_row_for_a_synthesized_proxy_repo` — `left: None, right: Some(1048576)` |
| refetch digest check removed | `test_get_cached_or_refetch_rejects_a_refetch_that_fails_the_recorded_digest` |
| cargo back on the unverified helper | `test_cargo_remote_download_uses_the_verified_streaming_helper_2929` |

Positive controls are present so no guard can be satisfied by rejecting everything: an object **under** the ceiling still caches; an object **exactly at** the ceiling is accepted (inclusive, matching `exceeds_single_object_quota`); with **no** ceiling configured an arbitrarily large object still caches; a digest-**matching** refetch is still written back; a non-canonical digest still repairs.

## Test Checklist
- [x] Unit tests added/updated — 20 new in-crate tests across `proxy_service.rs`, `proxy_helpers.rs`, `cargo.rs`
- [x] Manually tested locally

Gates run locally against an isolated Postgres:

```
cargo nextest run --workspace --lib                     14739 passed, 0 failed, 6 skipped
```

`nextest` rather than `cargo test --workspace`: with `AK_TESTS_REQUIRE_DB=1` the whole-suite runner produces bogus "Connection refused" failures because `config.rs` tests set `DATABASE_URL` process-wide.

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -146,6 +146,77 @@ async fn resolve_upstream_dl_url(
     Some(dl_url)
 }
 
+/// Extract the `cksum` recorded for `version` from a sparse-index document
+/// (#2929).
+///
+/// The sparse index is NDJSON: one JSON object per line, each carrying `vers`
+/// and `cksum`. `cksum` is the SHA-256 of the `.crate` file — it is exactly
+/// what cargo itself verifies the download against, and on a split-host
+/// registry (crates.io serves the index from `index.crates.io` and the
+/// `.crate` from a separate download host) it is sourced independently of the
+/// host that serves the bytes.
+///
+/// Returns `None` when the document is unparseable, the version is absent, or
+/// the recorded value is not a bare lowercase SHA-256 — all of which mean "no
+/// enforceable digest", so the caller falls back to an unverified fetch rather
+/// than failing the download.
+fn cksum_for_version(index_ndjson: &[u8], version: &str) -> Option<String> {
+    let text = std::str::from_utf8(index_ndjson).ok()?;
+    for line in text.lines() {
+        let line = line.trim();
+        if line.is_empty() {
+            continue;
+        }
+        let Ok(entry) = serde_json::from_str::<serde_json::Value>(line) else {
+            // A single malformed line must not hide a later good one.
+            continue;
+        };
+        if entry.get("vers").and_then(|v| v.as_str()) != Some(version) {
+            continue;
+        }
+        return entry
+            .get("cksum")
+            .and_then(|v| v.as_str())
+            .and_then(proxy_helpers::normalize_expected_sha256);
+    }
+    None
+}
+
+/// Resolve the sparse-index `cksum` for `{name}/{version}` from the upstream
+/// registry so the streamed `.crate` download can be digest-gated (#2929).
+///
+/// Best-effort by construction: any failure (no proxy, index fetch error,
+/// missing version, non-canonical digest) returns `None` and the download
+/// proceeds unverified, exactly as it did before. The index fetch uses the
+/// same proxy-cached, capped metadata helper the sparse-index handler uses, so
+/// a cargo client — which always reads the index immediately before
+/// downloading — finds it warm.
+async fn resolve_index_cksum(
+    state: &SharedState,
+    repo: &RepoInfo,
+    repo_key: &str,
+    name_lower: &str,
+    version: &str,
+) -> Option<String> {
+    let proxy = state.proxy_service.as_ref()?;
+    let base_url = repo
+        .index_upstream_url
+        .as_deref()
+        .or(repo.upstream_url.as_deref())?;
+    let index_path = cargo_sparse_index_path_upstream(name_lower);
+    let (content, _content_type) = proxy_helpers::proxy_fetch_capped(
+        proxy,
+        repo.id,
+        repo_key,
+        base_url,
+        &index_path,
+        proxy_helpers::DEFAULT_METADATA_MAX_BYTES,
+    )
+    .await
+    .ok()?;
+    cksum_for_version(&content, version)
+}
+
 /// Build the full download URL for a crate, using the upstream `dl` template
 /// when available. Falls back to `{upstream_url}/api/v1/crates/{name}/{version}/download`.
 ///
@@ -916,13 +987,34 @@ async fn download(
                     // The *metadata* fetches on this handler (config.json, sparse
                     // index) stay capped and buffered — an 8 MiB ceiling is
                     // correct for those.
-                    let result = proxy_helpers::proxy_fetch_streaming_with_cache_key(
+                    // #2929: gate the proxy-cache commit on the sparse index's
+                    // recorded `cksum` for this exact `{name}/{version}`. The
+                    // index entry this same handler serves already carries the
+                    // digest cargo verifies the download against, but the
+                    // download arm reached the streaming helper that hardcodes
+                    // `expected_checksum: None`, so the digest was decorative:
+                    // a `.crate` whose bytes disagreed with it was committed to
+                    // the cache and served warm from then on.
+                    //
+                    // This cannot catch a wholly compromised registry — index
+                    // and bytes could both be forged — but on a split-host
+                    // registry (crates.io: `index.crates.io` vs the download
+                    // host) it catches a misbehaving or compromised download
+                    // host while the index is intact, and it catches ordinary
+                    // corruption or a clean-EOF truncation on either. A body
+                    // that fails the gate is still streamed to the client
+                    // (which verifies it independently and errors); it is only
+                    // kept out of the cache.
+                    let expected_cksum =
+                        resolve_index_cksum(&state, &repo, &repo_key, &name_lower, &version).await;
+                    let result = proxy_helpers::proxy_fetch_streaming_with_cache_key_verified(
                         proxy,
                         repo.id,
                         &repo_key,
                         &dl_base,
                         &dl_path,
                         &cache_path,
+                        expected_cksum,
                         RepositoryFormat::Cargo,
                     )
                     .await?;
@@ -1701,6 +1793,95 @@ mod tests {
     // -----------------------------------------------------------------------
     // Test helpers
     // -----------------------------------------------------------------------
+
+    // ── #2929: the sparse-index cksum must gate the .crate cache commit ──
+
+    const CKSUM_A: &str = "abc1230000000000000000000000000000000000000000000000000000000def";
+
+    #[test]
+    fn test_cksum_for_version_picks_the_matching_version() {
+        let ndjson = format!(
+            "{{\"name\":\"serde\",\"vers\":\"1.0.0\",\"cksum\":\"{}\"}}\n\
+             {{\"name\":\"serde\",\"vers\":\"1.0.1\",\"cksum\":\"{}\"}}",
+            CKSUM_A,
+            "2".repeat(64)
+        );
+        assert_eq!(
+            super::cksum_for_version(ndjson.as_bytes(), "1.0.0").as_deref(),
+            Some(CKSUM_A),
+            "the digest must be the one recorded for the requested version"
+        );
+        assert_eq!(
+            super::cksum_for_version(ndjson.as_bytes(), "1.0.1").as_deref(),
+            Some("2".repeat(64).as_str())
+        );
+        assert_ne!(
+            CKSUM_A,
+            CKSUM_A.to_uppercase(),
+            "the fixture must contain hex letters or the case test below is vacuous"
+        );
+        assert_eq!(
+            super::cksum_for_version(ndjson.as_bytes(), "9.9.9"),
+            None,
+            "an absent version means no enforceable digest, not a wrong one"
+        );
+    }
+
+    #[test]
+    fn test_cksum_for_version_degrades_on_unusable_input() {
+        // A non-canonical cksum must read as "no digest available" rather than
+        // being enforced: it could never equal the streamed lowercase-hex
+        // SHA-256, so enforcing it would reject the crate on every fetch and
+        // re-pull upstream forever.
+        let prefixed = format!("{{\"vers\":\"1.0.0\",\"cksum\":\"sha256:{CKSUM_A}\"}}");
+        assert_eq!(super::cksum_for_version(prefixed.as_bytes(), "1.0.0"), None);
+
+        let upper = format!(
+            "{{\"vers\":\"1.0.0\",\"cksum\":\"{}\"}}",
+            CKSUM_A.to_uppercase()
+        );
+        assert_eq!(super::cksum_for_version(upper.as_bytes(), "1.0.0"), None);
+
+        assert_eq!(super::cksum_for_version(b"not json at all", "1.0.0"), None);
+        assert_eq!(super::cksum_for_version(b"", "1.0.0"), None);
+        assert_eq!(
+            super::cksum_for_version(b"{\"vers\":\"1.0.0\"}", "1.0.0"),
+            None,
+            "an entry with no cksum is not a digest"
+        );
+    }
+
+    /// A malformed line must not hide a good entry further down: the sparse
+    /// index is NDJSON produced by a third party, and a single bad line
+    /// silently disabling the digest gate for the whole crate would make the
+    /// guard trivially bypassable by a hostile index.
+    #[test]
+    fn test_cksum_for_version_skips_a_malformed_line() {
+        let ndjson = format!("this-is-not-json\n{{\"vers\":\"1.0.0\",\"cksum\":\"{CKSUM_A}\"}}");
+        assert_eq!(
+            super::cksum_for_version(ndjson.as_bytes(), "1.0.0").as_deref(),
+            Some(CKSUM_A)
+        );
+    }
+
+    /// Source pin (#2929): the Remote `.crate` download must go through the
+    /// DIGEST-GATED streaming helper AND actually hand it a resolved digest.
+    /// The unverified sibling hardcodes `expected_checksum: None`, so a revert
+    /// to it — or keeping the verified helper but passing `None` — would
+    /// compile, pass the rest of the suite, and silently un-enforce the index
+    /// cksum again.
+    #[test]
+    fn test_cargo_remote_download_uses_the_verified_streaming_helper_2929() {
+        let src = include_str!("cargo.rs");
+        assert!(
+            src.contains("proxy_helpers::proxy_fetch_streaming_with_cache_key_verified("),
+            "the cargo remote download MUST call the digest-gated streaming helper (#2929)"
+        );
+        assert!(
+            src.contains("let expected_cksum =\n                        resolve_index_cksum("),
+            "the digest passed to the gate must come from the sparse index, not be None (#2929)"
+        );
+    }
 
     /// Regression test for the cargo instance of the buffered-download class
     /// (#895 / #2192).
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -1864,22 +1864,176 @@ mod tests {
         );
     }
 
-    /// Source pin (#2929): the Remote `.crate` download must go through the
-    /// DIGEST-GATED streaming helper AND actually hand it a resolved digest.
-    /// The unverified sibling hardcodes `expected_checksum: None`, so a revert
-    /// to it — or keeping the verified helper but passing `None` — would
-    /// compile, pass the rest of the suite, and silently un-enforce the index
-    /// cksum again.
-    #[test]
-    fn test_cargo_remote_download_uses_the_verified_streaming_helper_2929() {
-        let src = include_str!("cargo.rs");
-        assert!(
-            src.contains("proxy_helpers::proxy_fetch_streaming_with_cache_key_verified("),
-            "the cargo remote download MUST call the digest-gated streaming helper (#2929)"
+    /// Fixture shared by the two #2929 cache-gate tests below.
+    ///
+    /// Serves a sparse-index document recording `index_cksum` for
+    /// `{name}/{version}` and a download host serving `body`, then issues two
+    /// cold requests through the real router. The upstream download mock is
+    /// pinned to `expected_download_hits`, which is what actually distinguishes
+    /// the two cases: a committed cache entry makes the second request a hit
+    /// (1 upstream fetch), a refused commit makes it another miss (2).
+    ///
+    /// Returns the served bodies plus the proxy cache dir so the caller can
+    /// assert on what was persisted.
+    async fn run_cargo_download_with_index_cksum(
+        name: &str,
+        version: &str,
+        body: Vec<u8>,
+        index_cksum: &str,
+        expected_download_hits: u64,
+    ) -> Option<(Vec<u8>, Vec<u8>, tempfile::TempDir)> {
+        use crate::api::handlers::test_db_helpers as tdh;
+        use wiremock::matchers::{method as wm_method, path as wm_path};
+        use wiremock::{Mock, MockServer, ResponseTemplate};
+
+        let fx = tdh::Fixture::setup("remote", "cargo").await?;
+
+        let server = MockServer::start().await;
+
+        // The sparse index this same handler proxies. `cksum` is the digest the
+        // download is gated on.
+        let index_doc = format!(
+            "{{\"name\":\"{name}\",\"vers\":\"{version}\",\"cksum\":\"{index_cksum}\",\"deps\":[],\"features\":{{}},\"yanked\":false}}\n"
         );
+        Mock::given(wm_method("GET"))
+            .and(wm_path(format!(
+                "/{}",
+                cargo_sparse_index_path_upstream(name)
+            )))
+            .respond_with(ResponseTemplate::new(200).set_body_string(index_doc))
+            .mount(&server)
+            .await;
+
+        Mock::given(wm_method("GET"))
+            .and(wm_path(format!("/api/v1/crates/{name}/{version}/download")))
+            .respond_with(ResponseTemplate::new(200).set_body_bytes(body.clone()))
+            .expect(expected_download_hits)
+            .mount(&server)
+            .await;
+
+        let (state, dir) = tdh::rewire_remote_proxy(&fx, &server.uri()).await;
+        let request_path = format!(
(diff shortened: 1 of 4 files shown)


brandonrc:
Reviewed adversarially at `d2c92b9` on an isolated clone and database, with 811 blast-radius tests and all four guards independently revert-proved. **The `#2928` diagnosis is right and the shape of the fix is right** — `build_remote_repo_with_format` hardcodes `quota_bytes: None` across ~15 production call sites, so `exceeds_single_object_quota` was inert on every proxy path, and resolving by `repo.id` fixes the class rather than one call site. Enforcing mid-stream rather than post-`put_stream` is also the correct call.

Three things block merging.

**1. The ceiling inverts a documented `quota_bytes` sentinel, and re-opens #1970.**

`0` and negative are a shipped, tested, documented **"unlimited"** sentinel:

```rust
match quota_bytes {
    Some(quota) if quota > 0 => current_usage + additional_bytes <= quota,
    // NULL or a non-positive sentinel (0 / negative) => unlimited.
    _ => true,
}
```

with `("zero_quota_one_byte", Some(0), 0, 1, true)` carrying the comment *"0 sentinel must NOT behave as a zero-byte cap (bug #1)"*, a CHANGELOG entry at 1.2.2 (*"`quota_bytes` 0/null treated as unlimited"*, #1970), and an update validator that explicitly admits `0`.

`quota_to_cache_ceiling` maps those same values to a **zero-byte ceiling**:

```
quota_bytes=Some(0)   upload admission says unlimited=true   proxy-cache ceiling=Some(0)
quota_bytes=Some(-1)  upload admission says unlimited=true   proxy-cache ceiling=Some(0)
with quota_bytes=0 the FIRST byte trips the tee ceiling: true
```

An operator who set `0` meaning *unlimited* gets their proxy cache **completely disabled** on both paths — every object abandoned at byte 1, every request re-pulling upstream forever. This was harmless while the guard was unreachable; this PR is what makes it live. And the new test enshrines it as intended behaviour.

Worth naming the trap, because it is subtle and general: parity was validated against `exceeds_single_object_quota`, whose 0/negative branch is **untested and was never reachable**. Parity was measured against dead code rather than against the product's contract. `quota_to_cache_ceiling` should return `None` for `q <= 0`, and the missing 0/negative cases belong in `test_exceeds_single_object_quota_table`.

**2. Revert-proof row 4 is false — the cargo test is not a guard.**

`test_cargo_remote_download_uses_the_verified_streaming_helper_2929` is an `include_str!` source-text assertion. The digest gate was disabled while keeping both asserted strings verbatim — resolve the cksum, then discard it and pass `None` — and the test **passed**. Assertion 1 is also the exact self-match shape the PR claims to have eliminated: revert the call site and the needle survives only inside the test's own assertion literal. Assertion 2 escapes that only by accident (its `\n` is a source escape) and is brittle against any rustfmt reflow.

**3. It does not merge.** `mergeable=CONFLICTING`, `state=DIRTY` — conflicts in `nuget.rs` and `proxy_helpers.rs` against `main`. Needs a rebase before anything else.

---

**On #2928's actual reach, stated plainly:** `quota_bytes` has no DEFAULT, `create_repository` passes it through unvalidated, and nothing in the Helm chart, docker-compose, `init-db.sql`, or any fixture sets it. The realistic production state is `NULL`, so both guards no-op. The PR converts a structurally-unreachable guard into a never-configured one. If #2928 is meant to bound streamed writes, it needs a config-driven cap independent of `quota_bytes`.

**Claims I checked that held up:** the client is *not* broken when the ceiling trips (`served == TOTAL`, verified), and no partial object is orphaned — filesystem removes its temp file, S3 aborts multipart, GCS aborts the resumable session. Digest normalization's fail-open is real but no current caller trips it: OCI strips the `sha256:` prefix, nuget stores bare lowercase hex.

**Composition with #3153 is better than claimed** — #3153 adds `discard_staging` to the very `Err` arm this PR drives, so a partial lands on a staging key and is deleted, never touching the live key. Note whoever merges second adds `None,` to four `tee_stream` call sites.

**Smaller items worth folding in:** log on the `resolve_quota_bytes` error arm (`.ok().flatten().flatten()` swallows a DB error into "no ceiling" — unbounded exactly when the DB is stressed); emit `record_proxy_cache_quota_exceeded` on the mid-stream trip, which currently produces no metric and a misleading `"put_stream failed"` log; and `resolve_index_cksum` runs on every remote `.crate` download bounded only by the global timeout.

One for a follow-up rather than this PR: the #2929 cargo integrity claim holds only when `index_upstream_url` is separately configured. Otherwise index and bytes share an origin, making the gate a corruption detector rather than an integrity control — the PR body oversells it.

brandonrc:
Pushed `a472694`, which fixes a blocker I found while reviewing this: **`quota_to_cache_ceiling` inverted the "unlimited" sentinel.**

It mapped a `quota_bytes` of `0` or negative to `Some(0)` — a zero-byte per-object ceiling that rejects every object. Platform-wide those values mean *unlimited*, not "allow nothing": `RepositoryService::quota_allows` reads `Some(quota) if quota > 0` and treats everything else as unbounded, with the comment `// NULL or a non-positive sentinel (0 / negative) => unlimited.` (CHANGELOG 1.2.2, #1970).

What makes this more than a rounding argument is that **this PR is what makes the column reachable in the first place.** `build_remote_repo_with_format` hardcodes `quota_bytes: None`, so before `resolve_quota_bytes` the single-object guard was structurally inert on every path that serves proxy traffic — which is exactly the finding this PR set out to fix. So activating it without honouring the sentinel wouldn't have preserved existing behaviour; it would have silently inverted it. Every repository explicitly configured as unlimited would have stopped caching proxied objects entirely, with no error and no log line to explain why.

Normalized in `resolve_quota_bytes` rather than at each consumer, because **both** consumers read it: the streaming ceiling here, and `cache_classifier::exceeds_single_object_quota` at the buffered write site, which this PR also newly feeds real values. That one is `object_len > quota`, so `Some(0)` would have rejected everything there too — the same inversion in a second place. `quota_to_cache_ceiling` repeats the filter rather than trusting its caller.

The test table now pins `None` for both sentinels and cross-checks that the two helpers agree on them, so they can't drift back apart silently.

Two things still outstanding on this PR, and the second one I'd like your call on:

1. The `include_str!`-based cargo test still needs replacing — noted from the earlier round, not addressed here.
2. **The branch is `CONFLICTING` and the rebase is not mechanical.** `main` has since moved the NuGet cached-artifact path to a streaming re-fetch (`storage.get_stream` with a `NotFound` re-fetch arm), while this branch replaces the same block with a buffered `proxy_helpers::get_cached_or_refetch` that enforces the row's recorded digest against the repaired bytes (#2929). Both are wanted; combining them means enforcing the digest over a stream rather than a buffer, which is a design decision about where the hash gets computed and what happens when it fails mid-body. I didn't want to make that call inside a conflict resolution and quietly change the digest-enforcement semantics you wrote. Happy to take a swing at it if you tell me which way you want it — verify-then-serve (buffer, bounded) or serve-and-abort-on-mismatch (stream, client sees a truncated body).

Issue #2439, opened 2026-07-12
HIGH cross-repo authz batch: ungated /security scan+finding reads + /sbom generate/cve writes (+ weak-gate upgrade)


brandonrc:
**Severity: HIGH. Live in shipped <=1.5.2.** The expedited (v1.5.3) subset of the systemic cross-repo authorization audit — routes that query artifact/repo-scoped data with NO per-repo authorization, exposing CVE/supply-chain data and allowing cross-repo writes to any authenticated non-member.

Canonical gates to apply: artifact-level `check_artifact_visibility` (artifacts.rs:25-77) / repo-level `require_visible` (repositories.rs:214-243) — token-scope + admin-bypass + **membership (`user_can_access_repo`)** + existence-hiding 404.

**HIGH routes (this hotfix):**
1. `sbom.rs:349` `generate_sbom` — `POST /api/v1/sbom` **(WRITE)** — generate/regenerate SBOM attestation on ANY repos artifact.
2. `security.rs:792` `list_findings` — `GET /security/scans/:id/findings` — CVE findings for any scan.
3. `security.rs:761` `get_scan` — `GET /security/scans/:id`.
4. `security.rs:723` `list_scans` — `GET /security/scans?repository_id=&artifact_id=`.
5. `security.rs:1138` `list_artifact_scans` — `GET /security/artifacts/:artifact_id/scans`.
6. `sbom.rs:1006` `update_cve_status` — `POST /sbom/cve/status/:id` **(WRITE)** — mirror the gating of the sibling finding-acknowledge/status handlers (member-visibility vs require_admin — match siblings).

**Weak-gate upgrade (closes the "also weak" sbom routes at once):** `sbom.rs:1600` `require_repo_access` currently checks only `can_access_repo` (token scope) — upgrade it to also run `user_can_access_repo` for private repos (i.e. `check_artifact_visibility` semantics). This closes get_sbom/get_sbom_by_artifact/delete_sbom/get_sbom_components/list_sboms/get_cve_history_by_artifact/update_cve_status_by_artifact_cve without patching each call site.

**Keep working:** public repos stay public (gate returns Ok for is_public); admins + members still 200; do not double-gate already-correct repo-nested twins (security::list_repo_scans etc.). Existence-hiding 404, never 403. The MEDIUM/LOW rest of the class (promotion/approval/curation/signing/quality-remainder rows 8,12) → 1.6.0 (#2440).

Issue #913, opened 2026-04-27
Backend: virtual member endpoints check parent access only, not member access


brandonrc:
## Bug

The three virtual-membership handlers in `backend/src/api/handlers/repositories.rs`:

- `add_virtual_member` at `:2051`
- `remove_virtual_member` at `:2136`
- `update_virtual_members` at `:2185`

all call `require_repo_access` on the virtual parent repo V only. None of them check the caller's access to the member repo M being attached, removed, or reordered.

## Impact

A caller with write scope on virtual repo V can add, remove, or reorder a member repo M to which they have no access rights. They cannot read M's contents directly, but they can change which repos a virtual repo aggregates, indirectly affecting downstream resolution.

This is either:

1. Intentional (the membership graph is considered V-owned metadata, and M's ACL only gates direct reads). If so, document this in the API spec and pin a test that asserts the behavior.
2. A bug. If so, add `require_repo_access(&auth, member_repo.id, RequiredScope::Read).await?;` alongside the existing parent check in all three handlers.

Decide which, and pin a test either way.

## Provenance

Surfaced by multi-agent review of artifact-keeper-test#80.

Tracking: ak-7093.

brandonrc:
Re-milestoned from v1.1.9 to v1.2.0. The fix tightens AuthZ in a way that flips some 200s to 403s for callers with parent-only access — that's a behavior change unsuitable for a maintenance release. Will land on main only and ship in v1.2.0 with a release-notes callout.

PR #3173, opened 2026-08-07
fix(repositories): scope virtual repo member and artifact listings to caller-visible members

### LINKED ISSUE #3163: Virtual repo member list and artifact listing enumerate members the caller cannot see (JWT sessions bypass the filter entirely)
Found while fixing #3081 (PR #3162). Same class, wider blast radius: #3081 disclosed member *byte sizes*, these disclose member *identity and contents*.

## 1. `GET /api/v1/repositories/{key}/members`

The handler resolves every member of the virtual repo, then filters with:

```rust
.filter(|row| auth.can_access_repo(row.member_repo_id))
```

`can_access_repo` is **token scope**, not repository visibility:

```rust
pub fn can_access_repo(&self, repo_id: Uuid) -> bool {
    self.access_scope().grants(repo_id)
}
```

The handler's own comment states the consequence:

> Tokens with `allowed_repo_ids = None` (admins, JWT sessions, unrestricted API tokens) see everything by virtue of `can_access_repo` returning true.

A browser session is exactly that shape. So **any authenticated user** who can reach a virtual repository can enumerate the `key`, `name` and `repo_type` of every member — including private repositories they hold no grant for and cannot read. The parent gate is `require_repo_access`, which is also token scope, so it does not compensate.

This is not the same control as the one used everywhere else. Repository visibility is decided by `build_visibility_clause_for` / `user_can_access_repo` (public ∪ `role_assignments` ∪ fine-grained `permissions`, including the group and project arms). The members endpoint never consults it.

## 2. `GET /api/v1/repositories/{key}/artifacts` on a virtual repo

Same root cause, one layer down. `proxy_helpers::fetch_virtual_members` resolves members with **no ACL predicate at all**; only the parent repository is `require_visible`d. The Maven-grouped variant has the same shape.

Notably `proxy_helpers::authorize_virtual_members` exists for precisely this purpose and is called only from `maven.rs` and `pypi.rs` — the generic listing paths skip it.

## Impact

Read-only disclosure, but of exactly the thing multi-tenant isolation is supposed to protect: which private repositories exist, what they are called, what format they are, and — via the artifact listing — what is in them. An operator who partitions tenants by repository and grants each tenant a virtual view has no reason to expect the member list to leak sideways.

## Verified as NOT affected

`/{key}/storage` and `/{key}/storage/tree` bind to `repo.id` with no member walk; quota goes through `check_quota` → single-repo ledger read. Search was reported clean by one reviewer but **I have not verified it myself** — treat as unconfirmed.

## Suggested fix

Use the same predicate as everything else. #3162 threads `RepoVisibility` into the virtual storage aggregates and splices `build_visibility_clause_for(visibility, "leaf", N)` into the CTE; the members and artifact-listing paths want the same treatment. For the artifact listing specifically, routing through the existing `authorize_virtual_members` rather than `fetch_virtual_members` may be sufficient — worth checking whether that helper is complete before extending it.

The general lesson worth encoding: `can_access_repo` answers *"is this repo within my token's scope?"*, which is **not** *"am I allowed to see this repo?"*. Anywhere the second question is meant, the visibility clause is the control. A grep for `can_access_repo` used as a visibility filter would be worth doing beyond these two sites.

brandonrc:
## Summary

Fixes #3163.

A virtual repository disclosed its members — and those members' contents — to callers who are not allowed to see the member repositories.

The root cause is one confusion, in three places: `AuthExtension::can_access_repo` resolves to `access_scope().grants(repo_id)`. That answers **"is this repository within my TOKEN's scope?"**, which is not **"am I allowed to see this repository?"**. For any *unscoped* principal — a browser JWT session, an unrestricted API token, an admin — `AccessScope::Admin` means "no repo restriction", so `can_access_repo` returns `true` for **every repository in the instance**. A browser session is exactly that shape.

**Site 1 — `GET /api/v1/repositories/{key}/members`**

- The member set was filtered with `.filter(|row| auth.can_access_repo(row.member_repo_id))`. The handler's own comment admitted the consequence ("Tokens with `allowed_repo_ids = None` … see everything"). Any authenticated user who could reach a virtual repo could enumerate the `key`, `name` and `repo_type` of every member, including private repositories they hold no grant on.
- The parent was gated with `require_repo_access` (token scope) and **never** with `require_visible`. So a private virtual repository's membership was readable by any authenticated principal — while `GET /{key}` on that same repository correctly 404s.

**Site 2 — `GET /api/v1/repositories/{key}/artifacts` on a virtual repo, flat and `group_by=maven_component`**

Both resolve members via `proxy_helpers::fetch_virtual_members`, which applies **no ACL predicate at all**; only the parent is `require_visible`d. The private members' artifacts (and, in the grouped view, their Maven GAV coordinates) were aggregated straight into the response.

### The fix

All three sites now narrow the member set through the **same predicate the repository listing uses** — `build_visibility_clause_for` (public ∪ `role_assignments` ∪ fine-grained `permissions`, including the group and project arms) — rather than inventing a second rule:

- `RepositoryService::filter_visible_repo_ids(candidate_ids, visibility)` (new) evaluates that clause over a candidate id set. It **preserves input order**, which is load-bearing: `list_for_repos_page` documents that "the priority is determined by the position in the `repo_ids` slice".
- `visibility_for_auth(Option<&AuthExtension>)` is the single derivation of `RepoVisibility` from a principal, extracted from the body of `list_repositories` (which now calls it).
- `RepoVisibility::All` short-circuits before the query, so the **admin path is byte-identical to before** and costs no extra round trip.
- On site 1 the parent now runs **both** gates. `require_repo_access` is deliberately kept *ahead* of `require_visible`, because `require_visible` short-circuits `Ok` on a public repository — dropping it would have let a repo-scoped token reach a public virtual outside its allowed set. Neither gate subsumes the other.

No schema migration, no API break, no response-shape change. Which members a non-admin *sees* changes — that is the point of the issue.

### Note for the reviewer: overlap with #3162

`visibility_for_auth` is also introduced by the still-open #3162 (for the #3081 storage-total leak), with the same body in the same file. That is intentional — #3163 explicitly asks these paths to reuse #3162's predicate rather than grow a second one. **Whichever of the two merges first, the other should drop its copy on rebase.** I did not stack this branch on #3162 because stacked PRs on this repo do not get Rust CI, and a security fix should not merge on an unrun gate.

### `can_access_repo` audit (requested in the issue)

I audited every production call site. 12 total; 10 are correct (token scope is genuinely what is meant, and a separate visibility check covers the path). Two are the **same bug class as this issue and are NOT fixed here** — they are unrelated endpoints and deserve their own issues rather than riding along:

| file:line | function | what an unscoped non-admin can read |
|---|---|---|
| `backend/src/api/handlers/quarantine.rs:106` and `:122` | `get_quarantine_status` | `!repo.is_public && !auth_ext.can_access_repo(...)` is the **only** gate (routes run under plain `auth_middleware`, not `repo_visibility_middleware`). Any logged-in user who knows an artifact UUID in another tenant's private repo learns it exists, plus `quarantine_status`/`quarantine_until`/`is_blocked`, and at `:122` the free-text `quarantine_reason` (the malware/CVE finding). The comment at `:119-121` states the intended semantics precisely — and `can_access_repo` does not implement them. |
| `backend/src/api/handlers/sbom.rs:441` | `list_sboms` | On the `repository_id`-only path (no `artifact_id`), `if !auth.can_access_repo(repo_id)` is the sole check before up to 100 `sbom_documents` rows are returned: SBOM ids, artifact ids, formats, component/dependency/license counts and license lists for any private repository. The sibling helper at `sbom.rs:1666-1670` documents this exact class as fixed for the `ensure_*` wrappers under #2439; `list_sboms` was never converted. |

Four in-crate (`src/…`, so they count toward the `nextest --lib` coverage gate) DB-backed tests in `mod virtual_member_visibility_tests`, each driving the **real `router()`** over `axum` `oneshot` as a non-admin JWT-session principal (`tdh::make_auth` — `is_admin: false`, `allowed_repo_ids: AccessScope::Admin`, i.e. exactly a browser session).

Fixture: a private virtual parent, a **private** member (`viewer` has no grant) and a **public** member, plus `viewer` / `insider` / `outsider` / admin principals. Every leak assertion ships with positive controls, so a fix that simply hid everything fails: the public member must still be listed, `insider` (who holds a grant on the private member) must still see it, and an admin must still see the whole set.

Revert-proof — each guard was reverted **individually** and fails exactly one test, and only that one:

| reverted guard | failing test |
|---|---|
| member filter → back to `can_access_repo` | `members_endpoint_hides_members_the_caller_cannot_see_3163` |
| parent `require_visible` removed | `members_endpoint_hides_the_virtual_itself_from_outsiders_3163` |
| flat-listing member filter removed | `virtual_artifact_listing_hides_invisible_member_artifacts_3163` |
| Maven-grouped member filter removed | `virtual_maven_grouped_listing_hides_invisible_member_components_3163` |

All four fail on the base commit and pass on this branch.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Response shapes and status codes are unchanged. The behavioural change is which members appear in them, plus `GET /{key}/members` now returning the existence-hiding `404` (instead of `200`) to a caller with no grant on a private virtual repository, matching `GET /{key}`.

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -337,28 +337,32 @@ pub(crate) fn member_grant_visibility(auth: Option<&AuthExtension>) -> RepoVisib
 
 /// The SCOPE half, applied per member row.
 ///
-/// A public member bypasses scope entirely, matching `require_visible`'s
-/// early return. Otherwise the member passes when it is itself in scope, OR
-/// when the PARENT virtual is — read-through.
+/// A public member bypasses scope entirely, matching `require_visible`'s early
+/// return. Otherwise the member must itself be in the token's scope.
 ///
-/// Read-through is a deliberate choice, not an oversight. Scoping a token to a
-/// virtual repository is the canonical way to point CI at one, and the
-/// by-path download through that virtual already serves member bytes for such
-/// a token. Without read-through the listing returns `200 OK` with zero items
-/// while the download of those same items succeeds — two answers to one
-/// question. The grant half still applies on top, so this widens which members
-/// a token may ENUMERATE, never which a user may READ.
+/// An earlier revision also passed when the PARENT virtual was in scope,
+/// arguing that the by-path download already read through. That premise was
+/// wrong: `proxy_helpers::caller_can_read_member` gates on
+/// `can_access_repo(member.id)` and has no parent term anywhere, so a token
+/// scoped to a virtual but not to a member is DENIED that member's bytes.
+/// Read-through would have created the inconsistency it claimed to remove,
+/// reversed — and widened enumeration past `require_visible`, letting a token
+/// list artifact names, versions and sizes for a member it is scoped away
+/// from. Strict scoping is what agrees with both.
+///
+/// Consequence, accepted deliberately: a token scoped only to a virtual sees
+/// an empty listing. That is consistent with the download, which also refuses.
+/// Scope tokens to the members, or to both.
 pub(crate) fn member_passes_token_scope(
     auth: Option<&AuthExtension>,
     parent_repo_id: Uuid,
     member_id: Uuid,
     member_is_public: bool,
 ) -> bool {
+    let _ = parent_repo_id;
     match auth {
         None => member_is_public,
-        Some(a) => {
-            member_is_public || a.can_access_repo(parent_repo_id) || a.can_access_repo(member_id)
-        }
+        Some(a) => member_is_public || a.can_access_repo(member_id),
     }
 }
 
@@ -8701,7 +8705,8 @@ pub async fn add_virtual_member(
             vrm.created_at,
             r.key as member_key,
             r.name as member_name,
-            r.repo_type
+            r.repo_type,
+            r.is_public
         FROM virtual_repo_members vrm
         INNER JOIN repositories r ON r.id = vrm.member_repo_id
         WHERE vrm.virtual_repo_id = $1 AND vrm.member_repo_id = $2
@@ -22919,6 +22924,10 @@ mod virtual_member_visibility_tests {
         let at = fx.scoped_admin_token(vec![fx.virt_id]);
         let (adm_status, adm_keys) = fx.members_seen(at).await;
 
+        // Positive control: same caller, member now IN scope.
+        let it2 = fx.scoped_token(&fx.insider, vec![fx.virt_id, fx.private_id]);
+        let (ins2_status, ins2_keys) = fx.members_seen(it2).await;
+
         let (private_key, public_key) = (fx.private_key.clone(), fx.public_key.clone());
         fx.cleanup().await;
 
@@ -22926,11 +22935,13 @@ mod virtual_member_visibility_tests {
         assert_eq!(ins_status, StatusCode::OK);
         assert_eq!(adm_status, StatusCode::OK);
 
-        // Read-through + the public arm the `Ids` arm had dropped.
+        // The public arm the `Ids` arm had dropped: a public member is
+        // readable regardless of scope, matching `require_visible`'s early
+        // return on `is_public`.
         assert!(
             virt_keys.contains(&public_key),
-            "a token scoped to the virtual must reach its public member, or \
-             the listing contradicts the download path; got {virt_keys:?}"
+            "a public member is readable regardless of token scope; \
+             got {virt_keys:?}"
         );
         // The grant half still applies.
         assert!(
@@ -22938,14 +22949,32 @@ mod virtual_member_visibility_tests {
             "token scope must not substitute for a grant: {private_key} has no \
              grant for this caller; got {virt_keys:?}"
         );
+        // STRICT scoping: the private member is NOT in this token's scope, so
+        // even a caller who holds the grant must not enumerate it. This is what
+        // agrees with the download path -- `caller_can_read_member` gates on
+        // `can_access_repo(member.id)` with no parent term, so read-through
+        // here would let a token list what it cannot fetch.
         assert!(
-            ins_keys.contains(&private_key),
-            "a scoped token whose owner HOLDS the grant must still see \
-             {private_key}; got {ins_keys:?}"
+            !ins_keys.contains(&private_key),
+            "a token scoped only to the virtual must not enumerate a private \
+             member outside its scope, because the download refuses it too; \
+             got {ins_keys:?}"
         );
+        // Scope the token to the MEMBER and the same caller does see it --
+        // the positive control, without which the assertion above is
+        // satisfied by a filter that hides everything.
+        assert_eq!(ins2_status, StatusCode::OK);
         assert!(
-            adm_keys.contains(&public_key) && adm_keys.contains(&private_key),
-            "an admin token scoped to the virtual reaches its members; \
+            ins2_keys.contains(&private_key),
+            "with {private_key} in scope AND a grant, it must be listed; \
+             got {ins2_keys:?}"
+        );
+        // An admin token scoped to the virtual only: the scope still binds.
+        // Before the fix `is_admin` was matched ahead of `Restricted`, so the
+        // scope was discarded entirely and every member was listed.
+        assert!(
+            !adm_keys.contains(&private_key),
+            "an admin's deliberately-scoped token must stay scoped; \
              got {adm_keys:?}"
         );
     }
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -4656,6 +4656,7 @@ pub async fn list_artifacts(
             count_exact,
             page,
             per_page,
+            &visibility_for_auth(auth.as_ref()),
         )
         .await;
     }
@@ -4729,7 +4730,16 @@ pub async fn list_artifacts(
                 AppError::Internal("Failed to resolve virtual repository members".to_string())
             })?;
 
+        // #3163: `fetch_virtual_members` applies NO access predicate — only
+        // the virtual PARENT was `require_visible`d above. Aggregating over
+        // the raw member set therefore served the contents of private member
+        // repositories to anyone who could see the parent. Narrow to the
+        // members this caller may see, using the same predicate as the
+        // repository listing and `GET /{key}/members`.
         let member_ids: Vec<uuid::Uuid> = members.iter().map(|m| m.id).collect();
+        let member_ids = repo_service
+            .filter_visible_repo_ids(&member_ids, &visibility_for_auth(auth.as_ref()))
+            .await?;
 
         let page_rows = artifact_service
             .list_for_repos_page(
@@ -5579,6 +5589,10 @@ async fn list_artifacts_grouped_by_maven_component(
     count_exact: bool,
     page: u32,
     per_page: u32,
+    // The CALLER's repository visibility (#3163). Only consulted for a
+    // Virtual repo, whose members are separate repositories with their own
+    // ACLs; a non-virtual repo is already gated by `require_visible`.
+    visibility: &RepoVisibility,
 ) -> Result<Json<ArtifactListResponse>> {
     // Remote (proxy) repositories do NOT record cached items in the `artifacts`
     // table (#1278 / #1280), so `artifact_service.list` returns nothing for them
@@ -5660,7 +5674,14 @@ async fn list_artifacts_grouped_by_maven_component(
             .map_err(|_| {
                 AppError::Internal("Failed to resolve virtual repository members".to_string())
             })?;
-        members.iter().map(|m| m.id).collect()
+        // #3163: same unfiltered member walk as the flat listing — narrow to
+        // the members this caller may see before the catalog is queried, so a
+        // private member's GAV coordinates are not disclosed through the
+        // grouped view.
+        let member_ids: Vec<Uuid> = members.iter().map(|m| m.id).collect();
+        RepositoryService::new(state.db.clone())
+            .filter_visible_repo_ids(&member_ids, visibility)
+            .await?
     } else {
         vec![repo.id]
     };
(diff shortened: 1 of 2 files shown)


brandonrc:
Verified both extra sites the audit turned up — they are real, and I've filed them as **#3174** (p1, security, 1.7.2) rather than widening this PR.

Two things from this work I want recorded where they will be found later.

**`authorize_virtual_members` was the wrong tool, and the reason generalises.** Its `caller_can_read_member` has `Ok(false) => true` — rules-empty falls **open**, so a private member with no fine-grained rules reads as visible to any authenticated caller. Using it here would have left the leak half-closed while looking fixed. That fall-open is still live on the maven and pypi download paths and is worth its own issue.

**Keeping `require_repo_access` alongside `require_visible` is load-bearing.** Replacing one with the other is the obvious simplification and it is wrong: `require_visible` short-circuits `Ok` on public repos, so a repo-scoped token would reach a public virtual repo outside its allowed set. Neither gate subsumes the other. Good catch before committing.

**On the overlap with #3162:** both add `visibility_for_auth` with the same body in the same file. Not stacking this branch was the right call — stacked PRs here get no Rust CI, and a security fix should not merge on an unrun gate. Whichever lands first, the other drops its copy on rebase. I'd suggest merging #3162 first since it is already green, then rebasing this one.

Also worth flagging for the maintainer: my original issue named the parent gate but did not score it. `list_virtual_members` never called `require_visible` **at all**, so an outsider with no relationship to a private virtual repo got a 200 listing its membership while `GET /{key}` on the same repo correctly 404s. That is an existence oracle, and it is fixed here.

brandonrc:
Reviewed adversarially against `b863d0a` on an isolated clone and database. **Everything the PR claims about its own diff is true** — I independently revert-proved all four guards (each fails exactly one test, no cross-talk), and confirmed both directions of the gate-subsumption argument live:

* drop `require_repo_access` → out-of-scope token on a **public** virtual goes `403 → 200`
* drop `require_visible` → unscoped outsider on a **private** virtual goes `404 → 200`, and it lists a member

Neither is dead code; keeping both was correct.

**But the fix hides the index while the bytes stay open**, and that changes what this PR can claim.

`GET /{key}/artifacts/*path` and `/{key}/download/*path` resolve through `proxy_helpers::resolve_virtual_download`, which takes **no `auth` parameter at all**. Against this PR's head:

```
listing  200  {"items":[],"pagination":{...,"total":0}}   <- fix works
download 200  TOP-SECRET-PAYLOAD                          <- bytes served
```

and with no credentials at all, against a public virtual with a private member:

```
anon listing  200  (private member correctly absent)
anon download 200  ANON-VISIBLE-SECRET
```

51 of 62 `fetch_virtual_members` call sites serve caller-visible data with no per-member ACL; 11 serve raw bytes. Filed as **#3178**.

**A second finding: the `can_access_repo` audit stopped one call level short of a privilege escalation.** It enumerated 12 *direct* call sites but never looked at `require_repo_access`, which is a two-line wrapper over the identical predicate — and it is the only member check in `authorize_virtual_member_mutation`. A non-admin with repo-admin on their own virtual can attach any private repository and read it:

```
direct GET of victim  -> 404      add_member -> 200      download -> 200 VICTIM-CROWN-JEWELS
```

Filed as **#3177** (p1). Worth renaming the wrapper to say what it actually checks, so the next audit is not misled the same way.

---

**Two things to fix on this PR before merge:**

**1. A pre-existing regression guard is now satisfied by a comment.** `test_list_virtual_members_requires_auth_and_filters` (the #913 guard) asserts the handler body contains `auth.can_access_repo(row.member_repo_id)`. After this change that string survives at exactly one place — inside a comment explaining that it *used* to be there. The test would pass with the entire new filter deleted. Retarget it at `filter_visible_repo_ids`. A security PR should not leave a security guard passing on a comment.

**2. Repo-scoped tokens now see empty listings.** `visibility_for_auth` maps a `Restricted` token to `RepoVisibility::Ids`, whose clause is `{alias}.id = ANY($N)` with **no `is_public` arm**. A token scoped to the virtual repo itself — exactly what `POST /{key}/tokens` mints — now gets `{"items":[],"total":0}` and `{"members":[]}`, where pre-fix it got everything. Meanwhile npm/OCI/conan still *serve* those artifacts to the same token, so the API says empty while the package manager pulls fine. That arm has zero test coverage here.

**On merge order with #3162 — not symmetric.** Merging #3162 into this branch produces exactly one conflict hunk (the duplicate `visibility_for_auth`), which fails loudly and is fine. But `repository_service.rs` **auto-merges clean**, silently leaving #3162's `split_visibility_bind` helper beside this PR's byte-identical inline `match` — duplicated logic the jscpd gate may flag. **Merge #3162 first**, then rebase this to drop the duplicate helper and call `split_visibility_bind`. The other order leaves #3162 owing two cleanups instead of one.

**Cost is fine, measured not estimated:** at 100k repositories / 100k role_assignments / 50k permissions, `filter_visible_repo_ids` runs 0.53 ms for 20 members and 11.4 ms for 500. `All` short-circuits with no query. The empty-set path fails **closed** — both `list_for_repos_page` and `maven_component_keys_from_catalog` early-return on `is_empty()`.

**Test gaps worth knowing:** `filter_visible_repo_ids` has 65 new lines and zero direct tests despite a documented order-preservation contract; the `Ids` and `PublicOnly` arms are untested (the latter works only because Postgres tolerates an unused declared bind — by luck, not by test); and the fixture drives only the `role_assignments` arm, never the `permissions`, group or project arms the doc comment advertises.

Net: this should land — it strictly reduces disclosure and breaks no existing test. But **#3163 must stay open**, or the changelog will claim member *contents* are protected while an anonymous request disproves it.

brandonrc:
Review round on `ad282f9`. Bug verified at merge-base on both sites, the predicate choice is right, and the tests are the good shape — real positive controls in every fixture, and they genuinely run in CI. But there are **two arm-level defects, one of which is a regression this PR introduces**, so I'm holding.

**HIGH-1 — an admin's deliberately-scoped token now sees everything. Verified against source, not inferred.**

`visibility_for_auth` matches `is_admin` *before* `Restricted`:

```rust
Some(a) if a.is_admin => RepoVisibility::All,
Some(a) if matches!(a.allowed_repo_ids, AccessScope::Restricted(_)) => RepoVisibility::Ids(...),
```

`can_access_repo` — what merge-base used — is `access_scope().grants()`, and `AccessScope::grants` is `Admin => true, Restricted(ids) => ids.contains(&repo_id)`. It keys on the **scope**, never on the `is_admin` flag. So an admin who mints a CI token with `repository_ids = [V]` specifically to bound blast radius had members outside `[V]` filtered at merge-base, and now gets all of them. The handler enforces token scope on the parent and ignores it on the members.

**HIGH-2 — a repo-scoped non-admin token gets a silently empty virtual.** `visibility_for_auth` → `Ids([V])` → `filter_visible_repo_ids([M1, M2], Ids([V]))` → `[]`. A service account scoped to the virtual repo — the canonical "point CI at the virtual" pattern — gets `200 OK` with zero items where it previously got the full aggregate. It reads as "the repo went empty", not as a 403, and the by-path download through the same virtual still serves bytes, so the two disagree.

Both trace to the same root cause: **the `Ids` arm is token scope, not visibility.** `require_visible` is `is_public OR (in_scope AND user_grant)`; the `Ids` arm drops both the public arm and the grant conjunct. That is the `can_access_repo`-vs-visibility confusion this PR set out to fix, surviving in the arm the fix routes through.

The shape that actually matches `require_visible`:

| principal | correct visibility |
|---|---|
| anonymous | `PublicOnly` |
| admin + **restricted** token | `Ids(scope)` — all repos in scope, no grant check |
| non-admin + restricted token | `scope ∩ (public ∪ grants)` — needs a new arm |
| admin, unrestricted | `All` |
| everyone else | `User(id)` |

Reordering alone fixes HIGH-1; the non-admin row needs a `ScopedUser { user_id, scope }` variant. HIGH-2 is then a product decision rather than a bug: does a token scoped to a virtual implicitly reach its members? I'd say yes (otherwise scoping to a virtual is useless), but that's your call and it should be stated in the release notes either way, not left silent.

**Two more, both verified:**

- **`visibility_for_auth` has no direct test.** The three that look like they cover it (`repositories.rs:16066/16077/16088`) each re-implement the match *inline* against a stale three-arm version with no `Restricted` arm. Deleting the `Restricted` arm from production leaves all three green **and** all four new DB tests green. That's the same "test asserts a copy, not production" species the second commit was written to fix one level up.
- **`members_endpoint_hides_the_virtual_itself_from_outsiders_3163` doesn't guard what the PR body says it does.** Deleting `require_repo_access` at `:8385` leaves it green — `outsider` is unscoped, so `require_visible` alone produces the 404. The body's justification for keeping both gates has no test behind it.

**Merge-order hazard, please sequence explicitly:** #3162 defines `visibility_for_auth` identically in this same file. Whichever lands second **will not compile** until rebased.

**And one the linked issue specifically asked about that the audit table missed:** `proxy_helpers::caller_can_read_member` has `Ok(false) => true` — rules-empty falls open — on top of another `can_access_repo`. It's used by four `maven.rs` sites and `pypi.rs`; the other ~40 `fetch_virtual_members` call sites apply no predicate at all. So after this PR `viewer` can no longer *enumerate* member `M` through `V`, but can still *retrieve* `M`'s bytes through `GET /V/artifacts/<path>` or a native client. Enumeration closed, retrieval open. Out of scope to fix here — but it belongs in the audit table, and #3163 shouldn't close as "the class is handled."

Unblock list: (1) reorder the arms, (2) decide HIGH-2 and state it, (3) add tests for a `Restricted` principal, an anonymous caller, and a member granted only via a fine-grained `permissions` row, (4) retarget `:16066-16088` at the real function, (5) file the `caller_can_read_member` issue.

I can push 1, 3 and 4 — say the word. 2 is yours.

brandonrc:
Pushed the fix for the shared `Ids`-arm defect. Both PRs are done from my side.

**The root cause, stated once.** `require_visible` is `is_public OR (in_scope AND (is_admin OR grants))`. The `Ids` arm is `in_scope` **alone** — it drops the public arm *and* the grant conjunct. That single gap produced opposite-direction bugs in the two PRs: #3173 discarded an admin's deliberately-scoped token (because `is_admin` was matched ahead of `Restricted`, and `can_access_repo` keys on the *scope*, never on the flag), and #3162 reported **0 bytes** where the correct answer was 10 GB.

**I did not change the shared `Ids` arm.** The repository listing's documented contract genuinely *is* "reflect only the token's scope" — a different question from "which members may this caller read." Instead both PRs now compose the two halves: `member_grant_visibility` answers the grant half in SQL, and in #3173 `member_passes_token_scope` answers the scope half per row. Together they are `require_visible` exactly.

**The one product call, made deliberately:** a token scoped to a *virtual* reads through to that virtual's members. The by-path download through the same virtual already serves member bytes to such a token — so without read-through the listing returns `200 OK` with zero items while the download of those same items succeeds. Two answers to one question. The grant half still applies on top, so this widens which members a token may **enumerate**, never which a user may **read**. It's documented on both helpers rather than left implicit.

**Both fixes are revert-proved, not argued.** #3173's new `AccessScope::Restricted` test fails under the revert with `got []` while all four pre-existing tests stay green — that arm had no coverage at all, which is why both defects shipped. #3162 gained the missing negative control (a visible, byte-holding repo that is *not* a member); removing the `virtual_repo_members` join now fails with `left: 8000, right: 1000`, where before only the admin assertion caught it, incidentally, via unrelated rows in the shared database.

Gates on both: `fmt --check` clean, `clippy --workspace --all-targets -D warnings` exit 0, 469/471 `repositories::` tests pass.

⚠️ **Merge order still applies and is now load-bearing:** both PRs define `member_grant_visibility` in `repositories.rs`. **Whichever lands second must drop its copy or it will not compile.** Noted in both commit messages.

Still open and yours: #3173's `caller_can_read_member` follow-up (rules-empty falls open, so member *content* is still reachable by path even though enumeration is now closed), and #3162's sibling artifact-listing channel — I'd keep #3081 open rather than close it on the total alone.

PR #3176, opened 2026-08-07
Forward upstream Content-Encoding on npm, PyPI, OCI, generic and cargo-index download responses

### LINKED ISSUE #3149: Streaming download handlers drop upstream Content-Encoding for npm, OCI, generic repos (and the cargo sparse index)
Same defect class as #2920 (cargo crate downloads), found while validating its fix in #2922. Filing separately rather than widening that PR.

A `StreamingFetchResult` carries `content_encoding` alongside `content_length`, but several response builders read the length and discard the encoding. The proxy's HTTP client pins `Accept-Encoding: identity` and disables transparent decoding, so upstream bytes arrive still compressed — serving them without the header means the client writes undecodable bytes and fails its checksum, exactly as reported in #2920.

**Confirmed by signature (the builder structurally cannot forward it):**

* **npm** — `build_tarball_response_stream` takes `content_type` and `content_length` parameters and no `content_encoding` at all. Affects the Remote and Virtual tarball arms.
* **OCI/Docker** — `oci_v2.rs` contains **zero** references to `content_encoding`. Affects blob GET, virtual blob GET, and the virtual-blob HEAD builder (which should keep header parity with GET).
* **generic repositories** — `repositories.rs` likewise has **zero** references. Affects virtual artifact download and HEAD.

**Also in scope, different mechanism:**

* **cargo sparse index** — `try_remote_index` fetches via `proxy_fetch_capped`, whose return type is a 2-tuple `Result<(Bytes, Option<String>), Response>` that structurally discards the encoding. The bytes are then re-served verbatim by `index_response`, which sets only `Content-Type` and `cache-control`, and are pinned into `index_cache`. A 3-tuple variant that preserves the coding already exists in `proxy_helpers.rs`. `config.json` uses the same 2-tuple path but is parsed server-side, so it degrades to the fallback URL rather than corrupting a client — lower severity.

**Reported as affected but verified only at call-site level, not confirmed by me:** pypi's `build_streaming_file_response` (6 call sites covering LKG, remote and virtual sdist/wheel serves).

**Not affected** — these route through `build_streaming_response_with_disposition`, which does forward: maven, goproxy, nuget, huggingface. conda has its own explicit handling.

## Suggested fix

Thread `content_encoding` through each builder the way #2922 does for cargo, or route these handlers through the shared `build_streaming_response_with_disposition` helper that already gets it right. Keep `Content-Length` and `Content-Encoding` sourced from the same place — the upstream `Content-Length` is the *coded* length, so forwarding one without the other is a protocol defect.

Each arm needs its own regression test. #2922 fixed two arms but shipped a test for only one, which is the gap to avoid repeating here.

## Related open question (worth answering once, applies to all formats)

None of these handlers consult the **client's** `Accept-Encoding` — no forwarding path references it on the response side. So the server would emit `Content-Encoding: gzip` regardless of whether the client accepted a coding. Declaring the coding is strictly better than dropping it, but whether it fully resolves the user-visible symptom depends on the client auto-inflating. Worth confirming against a real cargo/npm client before closing this out.

brandonrc:
## Summary

Fixes #3149. Same defect class as #2920, in the response builders that #2922 did not reach.

Since #2915 the proxy pins `Accept-Encoding: identity` and disables transparent decoding (`http_client.rs:93-115`), so a body coded by an intermediary (a corporate proxy, nginx with `gzip on`, a CDN) arrives still coded. `StreamingFetchResult::content_encoding` carries that coding and its own doc requires it be forwarded on every serve — but several builders read `content_length` and dropped the coding one line later. Since the upstream `Content-Length` is the **coded** length, forwarding one without the other is a protocol defect.

I verified each site named in the issue against the source rather than taking the report as given. Every one reproduced; **pypi, which the issue flagged as unverified, is genuinely affected too**.

| Site | Builder | Verdict |
|---|---|---|
| npm | `build_tarball_response_stream` | no `content_encoding` parameter at all — 4 proxied arms affected (remote, virtual, virtual-LKG, scan-pending); the hosted arm is not |
| pypi | `build_streaming_file_response` | **affected** — takes the whole `StreamingFetchResult`, read only `content_length`. Single builder behind all 6 call sites |
| OCI | `build_oci_streaming_proxy_response` + virtual-blob HEAD | zero references to `content_encoding` in `oci_v2.rs` |
| generic | `download_artifact`, `get_artifact_metadata` | zero references in `repositories.rs`; two independent hand-rolled builders |
| cargo sparse index | `try_remote_index` | 2-tuple `proxy_fetch_capped` structurally discards the coding |

### Notes on two non-mechanical changes

**OCI virtual-blob HEAD** now builds the exact GET response and drops the body, instead of re-deriving headers by hand. The hand-rolled copy had already drifted from GET (it forwarded type and length but not the coding); sharing the builder makes that class of drift unrepresentable. Dropping the body transfers nothing — the stream's upstream read is lazy and is never polled.

**cargo sparse index** had a second cause beyond the discarded coding. The in-process `index_cache` stores bare `Bytes` with no room for a coding, and its fast path serves them as `application/json`, so a coded body pinned there would hand *every later hit* an undeclared coded index. Coded responses now skip that memo; the ProxyService's own cache (whose sidecar does carry the coding) still serves them warm through the same arm, which re-declares it. Uncoded indexes are still memoized — there is a test asserting upstream is hit exactly once across two requests, so this cannot silently become "never cache".

### Deliberately out of scope

`config.json` (cargo) uses the same 2-tuple path, but the bytes are parsed server-side with `serde_json::from_slice(...).ok()?` and a coded body degrades to the fallback download URL. It never reaches a client, so there is no protocol defect to fix — forwarding a header would change nothing, because nothing there emits one. The real fix is server-side *decoding*, which is a different change with its own decompression-bomb budget. Worth a follow-up issue, not this patch.

### Does this fully resolve the user-visible symptom?

**Partly — it makes the response well-formed and self-describing, which is a necessary precondition, but this PR does not prove end-to-end client recovery.**

None of these handlers consult the **client's** `Accept-Encoding` on the response side, so the server will declare `Content-Encoding: gzip` even to a client that asked for `identity`. That is a deviation from RFC 9110 §12.5.3, and it is still strictly better than the status quo: before, coded bytes arrived undeclared and the client wrote garbage and failed its checksum with no indication why; now the response is honest, so a client that handles the coding recovers and one that does not gets a diagnosable protocol error instead of a corrupt artifact.

What this PR does **not** establish is that real cargo/npm/pip clients auto-inflate here — the issue explicitly flags that as needing confirmation against a live client, and I did not do that. Full correctness for a client that cannot decode would require content negotiation (decode server-side when the client did not accept the coding), which is a larger change than a 1.7.2 patch.

Note the npm router already applies a `CompressionLayer`; it excludes `application/gzip`/`application/octet-stream` by predicate and tower-http additionally skips any response that already carries a `Content-Encoding`, so the forwarded header cannot cause double-encoding.

### Adjacent, UNVERIFIED

`pypi.rs:2572` (`get_remote_cached_or_refetch_stream` → `tee_refetch_to_storage`) is the only remaining place that reads `content_length` off a fetch result without the coding. It is a cache-recovery path (cached entry missing on disk) that returns a bare stream, so a coding would be lost. I did not chase whether the artifact row's `size_bytes` describes coded or decoded bytes, so I am **not** claiming it is a bug — flagging it rather than guessing.

Every fixed arm has its own guard, and **each was revert-proved individually**: the arm was reverted in isolation and the suite re-run, confirming that arm's test fails and that no arm rides on another's coverage. This is the #2922 gap the issue asks not to repeat (it fixed two arms and shipped a test for one).

| Reverted arm | Tests that fail |
|---|---|
| npm builder | 3 (builder + remote + virtual) |
| npm remote call site | 2 |
| npm virtual-LKG call site | 1 (call-site guard) |
| npm virtual call site | 2 |
| npm scan-pending call site | 1 (call-site guard) |
| pypi builder | 1 |
| OCI GET builder | 2 |
| OCI virtual-blob HEAD | 1 |
| generic `get_artifact_metadata` | 1 |
| generic `download_artifact` | 2 |
| cargo sparse index | 1 |

The virtual-LKG and scan-pending npm arms are impractical to drive end to end, so they are covered by a source-level call-site guard that pins how every `build_tarball_response_stream` call site is shaped — reverting either one fails it.

One test in this PR was **caught being hollow by that process**: the first OCI GET/HEAD parity test rendered the resolution itself instead of calling the handlers, and passed with the HEAD arm reverted. It now drives `handle_get_blob`/`handle_head_blob` directly, and issues HEAD *before* GET so a warmed cache cannot route it onto a different branch.

Every arm also has a negative control: no upstream coding produces no manufactured header and a byte-identical body.

## Test Checklist
- [x] Manually tested locally

In-crate (`src/... mod tests`) so they count toward the `nextest --lib` coverage gate. Gates run locally against a dedicated Postgres: `cargo fmt --check` pass, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo nextest run --workspace --lib` **14743 passed / 0 failed**.

## API Changes
- [x] N/A - no API changes

Patch-shaped for 1.7.2: no schema migration, no API break. The only behavior change is that responses which were already serving coded bytes now say so.


brandonrc:
Review round on `19df120`. The production diff is directionally right and I found no path where it introduces corruption. But two of this PR's central claims are weaker than stated, and I've pushed a fix for the more serious one (`28d1586`).

**Every one of the fourteen tests survived replacing the forwarded value with a hardcoded `"gzip"` literal.** Every fixture was gzip and every assertion was `== Some("gzip")`, so the builders were pinned to *"some coding is declared"*, not *"the upstream's coding is declared"*. The negative controls survive too, because the `is_some()` guard stays intact. "Each arm was revert-proved individually" is true and insufficient — reverting isn't the only mutation, and it isn't the one that matters here.

Not academic: `http_client.rs`'s own comment records that object stores, S3 especially, return a stored `Content-Encoding` regardless of what the request advertised — and S3 commonly stores `br`. A brotli-stored wheel served labelled `gzip` makes urllib3 select the gzip decoder and raise `DecodeError` mid-stream, so pip hard-fails on a package that previously at least wrote raw bytes to disk. Strictly worse than the bug being fixed.

Added `coded_fixture(encoding, seed)` and switched one test per builder to `deflate` (flate2 is already a dep; there's no brotli crate in the tree). Verified by mutation: hardcoding `"gzip"` now fails npm, cargo, oci_v2 and repositories — one per builder, where before it failed nothing. Also made one test actually **decode** its body, since nothing proved the client receives bytes it can decode with the coding it was handed.

**The completeness claim is wrong, and I've filed it as #3184 rather than patching it here.** The table says *"pypi | `build_streaming_file_response` | Single builder behind all 6 call sites"*. There's a seventh. `build_scanned_file_response` (pypi) and `build_scanned_tarball_response` (npm) are buffered proxied-serve builders that set `Content-Length` from the coded length and never set `Content-Encoding` — the npm one doesn't even take the parameter. They structurally can't forward it, because `fetch_artifact_with_cache_path_capped` discards it in its signature (`let (content, content_type, _encoding) = ...`). So on any repo with `scan_on_proxy` enabled, `pip install` / `npm install` still gets the exact #3149 symptom. Widening that signature is a bigger change than this PR should carry. Worth correcting the body so it doesn't read as closed — and note this PR *did* fix npm's streaming pending arm inside the scanned flow, which is what makes the buffered arm easy to miss.

Three things I checked that came back **clean**, since they're the obvious worries: the cache stores **coded** bytes with `checksum_sha256` and `size_bytes` both over the coded bytes and agreeing with each other, so no silent corruption inside the cache; double-encoding is genuinely refuted — tower-http 0.6.11 hard-skips on `!res.headers().contains_key(CONTENT_ENCODING)` before the predicate runs; and `Vary` buys nothing here because the proxy pins identity upstream, so the response is invariant w.r.t. the client's `Accept-Encoding`.

Remaining items I did **not** push, in rough priority:

1. **`get_artifact_metadata` has no negative control** (`repositories.rs:6726`). Replace it with an unconditional `.header(CONTENT_ENCODING, "gzip")` and the whole PR stays green while every *uncoded* virtual serve through that handler is corrupted. Its sibling `download_artifact` has one; these two near-verbatim copies deserve to be one function anyway — neither forwards `result.etag` either, which is the same silently-dropped-field class #3149 is about.
2. **The cargo memo control test doesn't test the memo.** `test_uncoded_sparse_index_is_still_memoized...` claims "without this, 'never cache' would satisfy the test above" — but delete `index_cache_set` entirely and it still passes: the sparse index classifies Mutable with a 300s TTL, so request 2 lands on a **Fresh disk-cache** hit and never reaches upstream. The `.expect(1)` is measuring the disk cache. Assert `index_cache_get(...).is_some()`, or set `cache_ttl_secs = 0`.
3. **`try_virtual_index` (`cargo.rs:1414`) still calls the 2-tuple `proxy_fetch_capped`** — the exact call this PR's own comment condemns. `merge_index_lines` starts with `from_utf8(content)` and bails on error, so a gzip body makes the member's crate versions **silently vanish** from the aggregated virtual index, and the truncated result is then memoized. `cargo add serde` reports the version doesn't exist. Pre-existing, but the "deliberately out of scope" section names only `config.json`.
4. **Coded OCI blobs are now pullable but permanently uncacheable** — `expected_checksum` is the OCI digest, compared against a checksum computed over the *coded* stream, so it always mismatches → object deleted, sidecar skipped, warn logged, every layer pull re-hits upstream forever. Fails safe, but it was unreachable before this PR. The RPM analogue additionally fires `record_proxy_cache_checksum_mismatch`, a *security* metric, on a benign `gzip on`.
5. `oci_v2.rs:3553` says `Docker-Content-Digest` names the digest of the **decoded** layer; `proxy_service.rs:1572` treats it as covering the coded bytes. Both can't hold, and the comment is what the next reader will trust.
6. `Content-Encoding: identity` is now forwarded verbatim — not a valid response coding, and it newly disables npm's `CompressionLayer` for those responses since tower-http keys on presence, not value.
7. Please label this `type:bug`. Branch is `fix/*`, body says "Fixes #3149", and the issue is `type:bug` — it just isn't labelled.

Happy to take 1 and 2 now; 3–6 I'd file as follow-ups.

PR #3179, opened 2026-08-07
fix(pypi): PEP 658 virtual repo handling

### LINKED ISSUE #3175: PEP 658 metadata routing for virtual repos
### Description

PR #3077 fixed metadata routing for remote repositories, but fails on virtual repos referencing those remote repo artifacts.

```rust
if repo.repo_type == RepositoryType::Remote {
    return serve_remote_metadata(...);
}

return serve_metadata(... repo.id ...);
```

For a virtual repository, repo.id is the virtual repository ID, which owns no artifacts, so the fallback necessarily returns 404.

### Environment Information

- Deployment type: 
- Operating System: 
- Architecture: 
- Database: 
- Reverse Proxy: 
- Network setup: 

### Backend Version

1.7.1

### Web UI Version

1.7.0

### Package Format

PyPI

### Expected Behavior

For virtual repositories, it must resolve the owning member and:
- Call serve_remote_metadata for remote members.
- Call serve_metadata for local members.

### Observed Behavior

404 for metadata

### Steps to Reproduce

1. Create remote repo for PyPI
2. Create virtual repo and include above as member
3. uv pip install --index-url virtual-repo setuptools==80.10.2 wheel==0.47.0

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

jfgordon2:
## Summary
Closes #3175 - extends #3077 to virtual repos.

For full-disclosure, I'm no rust programmer, so this was AI assisted with additional review by codex.

#3077 fixed PEP 658 metadata routing for remote repositories. Virtual repositories still queried metadata against the virtual repository itself, which owns no artifacts, resulting in a 404.

New logic:
- Resolves metadata through authorized virtual members in priority order.
- Preserves remote `.metadata` fetching and wheel-extraction fallback.
- Applies the same priority-aware dependency-confusion isolation used by simple indexes and artifact downloads.
- Allows remote platform-specific wheels for locally owned versions when they support a platform not available locally, while blocking remote-only versions and competing builds.
- Stabilizes the large-object timeout test by separating virtual timeout advancement from wall-clock socket readiness.

Regression coverage verifies:
- Metadata resolves through a virtual repository’s remote member.
- A remote metadata 404 falls back to extracting metadata from the wheel.
- Remote-only metadata remains hidden when a higher-priority local member owns the project.

## Test Checklist
- [ ] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -1705,6 +1705,26 @@ async fn serve_virtual_metadata(
     let members =
         proxy_helpers::authorize_virtual_members(&state.permission_service, auth, members).await;
 
+    // Keep PEP 658 metadata resolution symmetric with the simple-index and
+    // distribution-download paths. A higher-priority local owner suppresses a
+    // lower-priority Remote member unless the requested distribution is an
+    // admitted Case-A platform wheel. Without this gate, a direct or stale
+    // `.metadata` URL could expose a remote-only Case-B version that neither
+    // the index advertises nor the download route serves.
+    let owning_local_min_priority =
+        proxy_helpers::pypi_virtual_isolates_name(&state.db, virtual_repo.id, normalized_project)
+            .await?;
+    let member_priorities = if owning_local_min_priority.is_some() {
+        proxy_helpers::fetch_virtual_member_priorities(&state.db, virtual_repo.id).await?
+    } else {
+        Default::default()
+    };
+    let owned_profile = if owning_local_min_priority.is_some() {
+        pypi_owned_wheel_profile(&state.db, &members, normalized_project).await
+    } else {
+        OwnedWheelProfile::default()
+    };
+
     for member in members {
         let member_info = proxy_helpers::repo_info_from_member(&member);
         if enforce_pypi_curation(state, &member_info, normalized_project, requested_version)
@@ -1715,6 +1735,20 @@ async fn serve_virtual_metadata(
         }
 
         let result = if member.repo_type == RepositoryType::Remote {
+            let remote_suppressed = match owning_local_min_priority {
+                Some(local_min) => {
+                    local_min
+                        < member_priorities
+                            .get(&member.id)
+                            .copied()
+                            .unwrap_or(i32::MAX)
+                        && !owned_profile.admits(filename)
+                }
+                None => false,
+            };
+            if remote_suppressed {
+                continue;
+            }
             match (&member.upstream_url, state.proxy_service.as_ref()) {
                 (Some(upstream_url), Some(proxy)) => {
                     serve_remote_metadata(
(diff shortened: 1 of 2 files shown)


brandonrc:
Ran this through the review pipeline at `45d9e09`. Stage 1 and 2 pass — the bug is real and your fix closes it — and I want to lead with that, because the parts of `serve_file` you ported were ported carefully.

**Issue verified.** On `main` the dispatch is `if repo.repo_type == Remote { serve_remote_metadata } ... serve_metadata(repo.id)`, so a virtual repo — which owns no artifacts — necessarily 404s. Exactly as #3175 describes.

**Fix verified, by running your own test against `main`'s dispatch.** I deleted the `Virtual` arm and left everything else, and `test_virtual_pypi_metadata_resolves_remote_member` fails with `left: 404, right: 200`. That's a genuine red→green regression test and it satisfies the `fix/*` gate on its own.

**One finding from my side: your second test doesn't gate anything against `main`.** `test_virtual_pypi_metadata_suppresses_remote_only_version` asserts a 404 — and with the `Virtual` arm removed, *everything* 404s, so the bug itself satisfies it. It passed on the reverted shape.

To be fair to it, I then checked the mutation that actually matters — setting `remote_suppressed = false` to disable the isolation outright — and it **does** catch that (`left: 200, right: 404`). So it is a real guard on the new suppression logic, just not a regression test. Worth moving out of the "Regression coverage" list in the body so nobody later reads it as red→green protection.

Independent review also raised two things I'd want addressed before merge:

**1. Remote members skip the local-artifact lookup, so metadata and the wheel can come from different bytes.** Your loop dispatches on `member.repo_type` alone. `serve_file` doesn't — at `pypi.rs:2378` it calls `local_fetch_or_redirect_by_suffix` for *every* member before the Remote branch, with the in-code note that this "works for hosted repos and cached remote artifacts." Remote-typed repos do legitimately carry `artifacts` rows (direct upload, replication, promotion). So for a virtual whose highest-priority member is a Remote repo holding a replicated wheel: the download serves the stored bytes, while `.metadata` skips the row and fetches the *upstream's* metadata. pip gets one package's dependency set and installs another's, with nothing to detect it — PEP 658 metadata isn't hash-tied to the distribution. The fix is to mirror `serve_file`: try the local lookup for every member first, fall through to `serve_remote_metadata` only on 404 *and* Remote.

**2. One bad member aborts metadata for the whole virtual repo.** `Err(response) => return Err(response)` propagates any non-404. That's the safe instinct and I'd rather have it than a swallow — but `serve_file`'s loop logs and continues, so the two paths now disagree, and metadata is the strict one. Upstream 401/403/5xx map to `BadGateway`, so a Remote member with expired credentials at priority 0 makes a purely-local package's metadata 502 even though its download succeeds fine. Record the first non-404 and return it only if no member serves.

Also worth a comment correction rather than code: the docstring says this uses "the same authorization and curation boundaries as the regular download path." Authorization and curation are both there — but the age gate, quarantine check, and inline scan-and-block gate are not, so a quarantined wheel returns 403 on download and 200 with its full `METADATA` here. The age and scan gaps are parity with the existing direct-Remote metadata route, but quarantine is newly reachable. I'd file those as follow-ups and just narrow the sentence.

Two things I checked that came back **clean**, since they're the usual suspects: the wheel-extraction fallback is properly bounded (8 MiB download cap, `bounded_archive` zip-inflation limit, extraction permit — no zip-bomb exposure, malformed wheels degrade to a clean 404), and the `http_client.rs` change is entirely inside `#[cfg(test)]` — a `tokio::time` sequencing fix with zero production blast radius. The PR body's "separating virtual timeout advancement from wall-clock socket readiness" reads like a shared-client change; it isn't, and saying so plainly would save the next reviewer the trip.

For a first Rust contribution this is solid work — no new unwraps on reachable paths, no panics on malformed input, no locks held across `.await`, and wiring `authorize_virtual_members` in closes a confused-deputy hole on a route that had never been exposed. The gap is structural, not a Rust problem: dispatching on repo type where the reference implementation dispatches on storage reality first.

Happy to push both fixes myself if you'd rather — say the word.

brandonrc:
Follow-up on the security pass. It found a real bypass in the new function, I reproduced it, and **I've pushed the fix to your branch as `2b5ba69`** — hope that's welcome; revert it if you'd rather take it differently.

**The dependency-confusion gate could be turned off with a single space in the URL.**

Every gate in `serve_virtual_metadata` keys on `normalized_project` (`normalize_pep503`), but the upstream fetch was handed the **raw** path segment, and `serve_remote_metadata` re-normalizes with `PypiHandler::normalize_name` — a different function:

```
input        gate (normalize_pep503)   fetch (normalize_name)
"acme-sdk"   acme-sdk                  acme-sdk      same
"acme sdk"   acmesdk                   acme-sdk      DIVERGE
```

`normalize_pep503` silently *drops* characters outside `[A-Za-z0-9._-]` (a deliberate XSS defense); `normalize_name` turns every non-alphanumeric into `-`. So for a private `acme-sdk` owned by a higher-priority local member, `GET /simple/acme%20sdk/acme_sdk-9.9.9-py3-none-any.whl.metadata` evaluates the gate as `acmesdk` — owned by nobody, nothing suppressed — then fetches `acme-sdk` and serves a public shadow package's metadata under the private name. The 404 fallback also pulls the attacker's wheel into the member's proxy cache.

This is not something you invented — it's the same class as the `#3077` stripper divergence, "the string used for the policy check differs from the string used to fetch." It just landed in a new place.

I reproduced it before fixing rather than arguing from the code: the new test fails on `45d9e09` with `left: 200, right: 404`.

**One fixture detail worth knowing, because it cost me a false pass first:** the private package name must contain a separator. With `demo`, the two normalizers can't diverge in the direction that misses the gate *and* hits the private package, so a test built on that name passes and tells you nothing. My first version did exactly that.

The fix passes `normalized_project` — which the gates already used — and `normalize_name` is idempotent on PEP 503 output, so upstream URLs and cache keys are unchanged for well-formed requests. I also dropped the raw `project` parameter from the function entirely, since not having the raw name in scope is what stops this drifting apart again.

`clippy -D warnings` exit 0, 369 pypi tests pass.

**Separately: the same divergence is live on the download path in 1.7.1**, where it serves package *bytes* rather than metadata. That's pre-existing, not yours — I confirmed it with the same test shape (`left: 200`) and filed it as #3183 with the reproduction.

Still open from the earlier comment, and unchanged by this push — the local-lookup-for-every-member mismatch, and the strict error propagation (a second reviewer independently reached the same conclusion on both, and added that the age gate has no counterpart here, so a withheld version's METADATA is served while its wheel 451s). Happy to push those too if you want them handled the same way.

brandonrc:
Last round of the test audit came back, and it ran an actual mutation matrix against your branch rather than reading: **8 of 13 production mutations left the suite green.** I've pushed `d273bba` closing the two that block.

It independently reproduced what I reported earlier — `test_virtual_pypi_metadata_suppresses_remote_only_version` passes with the fix reverted — and identified exactly why. Pre-fix, a virtual `.metadata` request falls through to `serve_metadata(repo.id)` against a repo that owns no artifacts, so *everything* 404s. The test asserts 404. Its asserted outcome is byte-identical to the bug, so it can't fail on `main`, and it can't distinguish "the guard suppressed the remote" from "nothing resolved at all."

**The whole fix was a positive control in the same fixture.** Seed the local owner's wheel bytes, and assert its own `.metadata` resolves 200 with the *local* member's body. That one addition kills two mutations:

- reverting the virtual dispatch → `left: 404, right: 200`
- passing `virtual_repo.id` instead of `member.id` into `serve_metadata` → `left: 404, right: 200`

The second is worth dwelling on: that is *precisely the bug class this PR fixes*, on the local branch, and nothing could previously detect it. Both existing tests only ever see the local member as a miss, so the entire local resolution path was unexecuted.

**Second test: the member authorization filter.** `serve_virtual_metadata` is a new way to read member content through a virtual repo, so `authorize_virtual_members` is load-bearing here rather than inherited — but both fixtures set every member `is_public = true`, so no denial was ever exercised. With the filter removed the new test fails and prints the private member's metadata straight into the assertion message.

A process note that cost me a cycle and may save you one: my first attempt at that mutation patched the wrong occurrence. `grep authorize_virtual_members` returns the *download* path first, and mutating that one proves nothing about this function. Same trap as the fixture-name issue I hit earlier — a mutation that fails for the wrong reason is as misleading as one that passes.

Gates: `clippy --workspace --all-targets -D warnings` exit 0, 370 pypi tests pass.

Also confirmed clean, since they're the usual suspects: both new tests are inline `mod tests`, so they're `--lib` and **do** run in the Code Coverage gate, and CI sets `AK_TESTS_REQUIRE_DB=1` so the DB fixtures can't silently skip. No tests were added under `backend/tests/`, which is the trap #3077 hit. And the `http_client.rs` de-flake was measured against injected timeouts at 5s/15s/25s/30s/300s — detection power is identical before and after, so it's a genuine flake fix, not a weakened assertion.

Left for you, unchanged: the local-lookup-for-every-member mismatch, the strict error propagation, and the missing age gate. Three reviewers converged on the first two independently. Two smaller test gaps I didn't add — nothing pins the Case-A "remote platform-specific wheel for a locally-owned version" allowance your summary describes (deleting that conjunct entirely stays green), and nothing pins priority *ordering*. Say the word and I'll take any of them.

PR #3189, opened 2026-08-08
fix(npm): reject non-ASCII hosted package names

### LINKED ISSUE #3007: npm hosted publish accepts non-ASCII package names (asymmetry with PyPI + the npm proxy path)
## Summary
Surfaced during the #2956 typo-squat red-team. Hosted npm `publish`/`publish_scoped` (`backend/src/api/handlers/npm.rs` ~3015-3040) validate the package name only via `validate_package_name` (empty/length/null/traversal/slash/leading-dot) — there is **no ASCII check**. The strict validator `is_valid_npm_name` exists but is applied only on the proxy-ownership path (~npm.rs:2803), not at publish.

By contrast: PyPI hosted upload enforces ASCII (`PypiHandler::normalize_name`, `backend/src/formats/pypi.rs:184`, folds non-ASCII to separators), and the npm proxy path already uses the strict validator. So a caller can publish an npm package with a Unicode homoglyph name (`nｕmpy`) to a hosted repo — enabling the impersonation the #2956 typo-squat detection then has to catch downstream.

## Impact
Low/defense-in-depth. The #2956 homoglyph + proximity-gated mixed-script curation detection now flags these, so it's not an open hole — but rejecting non-ASCII names at the hosted npm publish boundary (matching PyPI and the npm proxy path) is the cleaner primary defense and removes the asymmetry. Registry npm itself restricts names to a URL-safe/ASCII set.

## Fix direction
Apply `is_valid_npm_name` (or the registry's documented name grammar) at the hosted npm `publish`/`publish_scoped` entry, consistent with the proxy path and with PyPI's normalize. Confirm no legitimate existing hosted package relies on a non-ASCII name first. Keep it a validation-at-boundary change; the curation detection stays as the downstream backstop.

Related: #2956 (typo-squat detection), the shared curation typosquat work.

cazlo:
## Summary

- reject non-ASCII package names at the decoded hosted npm publish boundary
- cover both unscoped and scoped publish routes, including percent-encoded Unicode input
- preserve the handler's existing ASCII and uppercase compatibility

## Why

The hosted publish validator checked length and path shape but did not require ASCII, so Unicode homoglyph names such as `nｕmpy` could reach publication. The proxy path already uses a stricter npm-name validator, leaving an avoidable asymmetry at the primary write boundary.

This uses an explicit ASCII check instead of replacing the hosted validator with `is_valid_npm_name`: that helper also rejects uppercase names, while existing hosted behavior and tests accept `MyPackage` and `@Scope/Package`. A lowercase/grammar migration would be a separate compatibility decision.

Validation happens after URL decoding and before `publish_package`, so rejected names cannot reach repository lookup, payload parsing, metadata persistence, or object storage.

## Validation

Local:

- `cargo nextest run --workspace --lib -E 'test(/validate_npm_package_name|publish_.*non_ascii|publish_ascii_names_accepted/)'` — 11 passed
- `cargo fmt --all --check`
- `cargo check --workspace`

- 14,750/14,750 backend tests passed; 8 skipped
- 86.94% overall line coverage
- 96% changed-line coverage (1,878/1,941)
- strict lint and all nine integration shards passed
- image build, scan, sign, and verification passed
- PyPI/npm/Cargo native-client smoke passed
- Playwright web E2E passed 11/11

Closes #3007

### DIFF
--- a/backend/src/api/handlers/npm.rs
+++ b/backend/src/api/handlers/npm.rs
@@ -761,10 +761,10 @@ fn normalize_package_name(raw: &str) -> String {
 
 /// Validate a decoded npm package name.
 ///
-/// Rejects names with path traversal sequences, null bytes, and names that
-/// violate the npm naming rules (empty, too long, leading dot/underscore,
-/// non-lowercase for unscoped packages). Called after URL decoding to catch
-/// percent-encoded attacks like `%2e%2e%2f`.
+/// Rejects non-ASCII names, path traversal sequences, null bytes, and names
+/// that violate the handler's existing shape rules (empty, too long, leading
+/// dot/underscore). Called after URL decoding to catch percent-encoded attacks
+/// like `%2e%2e%2f` and percent-encoded Unicode homoglyphs.
 #[allow(clippy::result_large_err)]
 fn validate_package_name(name: &str) -> Result<(), Response> {
     if name.is_empty() {
@@ -775,6 +775,12 @@ fn validate_package_name(name: &str) -> Result<(), Response> {
     }
     if name.len() > 214 {
         return Err(map_status(StatusCode::BAD_REQUEST, "Package name too long"));
+    }
+    if !name.is_ascii() {
+        return Err(map_status(
+            StatusCode::BAD_REQUEST,
+            "Package name must contain only ASCII characters",
+        ));
     }
     if name.contains('\0') {
         return Err(map_status(
@@ -813,6 +819,25 @@ fn validate_package_name(name: &str) -> Result<(), Response> {
         ));
     }
     Ok(())
+}
+
+#[allow(clippy::result_large_err)]
+fn validate_publish_package_name(raw: &str) -> Result<String, Response> {
+    let package = normalize_package_name(raw);
+    validate_package_name(&package)?;
+    Ok(package)
+}
+
+#[allow(clippy::result_large_err)]
+fn validate_scoped_publish_package_name(
+    raw_scope: &str,
+    raw_package: &str,
+) -> Result<String, Response> {
+    let scope = normalize_package_name(raw_scope);
+    let package = normalize_package_name(raw_package);
+    let full_name = build_scoped_package_name(&scope, &package);
+    validate_package_name(&full_name)?;
+    Ok(full_name)
 }
 
 fn map_status(status: StatusCode, msg: &str) -> Response {
@@ -3624,8 +3649,7 @@ async fn publish(
     headers: HeaderMap,
     body: Bytes,
 ) -> Result<Response, Response> {
-    let package = normalize_package_name(&package);
-    validate_package_name(&package)?;
+    let package = validate_publish_package_name(&package)?;
     publish_package(&state, auth, &repo_key, &package, &headers, body).await
 }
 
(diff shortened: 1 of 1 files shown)


brandonrc:
Reviewed and pushed one remediation commit (d18859d). The fix is right; the placement was broader than the issue.

## What I changed

`validate_package_name` is not the publish validator — it is the **shared** validator behind 11 call sites, and only 2 of them are publish:

| route | fn | line |
|---|---|---|
| `GET` packument | `get_metadata` / `get_scoped_metadata` | 1691, 1704 |
| `GET` version metadata | `get_version_metadata` | 1714, 1726 |
| `GET` tarball | `download_tarball` / `download_scoped_tarball` | 2847, 2859 |
| `GET`/`PUT`/`DELETE` dist-tags | `dist_tags_get` / `dist_tags_put` / `dist_tags_delete` | 4003, 4041, 4116 |

Putting `!name.is_ascii()` inside it therefore applies the rule to **reads**, for local, virtual *and* remote repositories. Two consequences beyond what #3007 asks for:

1. Any non-ASCII package already published becomes permanently unreadable — packument and tarball both `400`. The remediation for an existing homoglyph package should be deleting it, not making it a silent `400` that also hides it from the curation tooling.
2. `download_tarball` serves **remote/proxy** repos. npm grandfathers names that predate its URL-safe rule (`validate-npm-package-name` only rejects non-URL-safe characters for *new* packages), so this would block proxied pulls of legacy upstream names on name shape alone — a proxy behaviour change nobody asked for.

The PR already had the right structure for this — `validate_publish_package_name` / `validate_scoped_publish_package_name` — but those wrappers did nothing the previous inline code did not. I moved the ASCII rule into `assert_publishable_package_name`, called only from those two wrappers, so the wrappers now carry the rule that justifies them.

## Tests

Kept yours; added two.

- `test_non_ascii_names_remain_readable_on_the_shared_validator` — the guard against re-introducing this. Asserts `validate_package_name` still accepts `nｕmpy`, `@ｓcope/package`, `@scope/pａckage`, `日本語`.
- `test_same_names_are_still_blocked_at_publish` — its positive control. Without it, deleting the publish-side check would leave the read guard passing on its own and prove nothing.

## Revert-proof

Both directions, faithfully reverted against the merge base.

Revert the publish check (`assert_publishable_package_name` → `Ok(())`):

```
FAIL test_same_names_are_still_blocked_at_publish
FAIL test_publish_unscoped_non_ascii_names_rejected
FAIL test_publish_scoped_non_ascii_names_rejected
PASS test_non_ascii_names_remain_readable_on_the_shared_validator
4 tests run: 1 passed, 3 failed
```

Restore this PR’s original placement (ASCII check back inside `validate_package_name`):

```
FAIL test_non_ascii_names_remain_readable_on_the_shared_validator
     panicked: read path must still resolve nｕmpy; the ASCII rule is publish-only
PASS test_same_names_are_still_blocked_at_publish
PASS test_publish_unscoped_non_ascii_names_rejected
PASS test_publish_scoped_non_ascii_names_rejected
4 tests run: 3 passed, 1 failed
```

So each guard fails for its own reason and neither is hollow.

Full run green: 13/13 on the name-validation selector, `cargo fmt --all`, `SQLX_OFFLINE=true cargo build --lib --tests` clean.

One thing I left alone deliberately: renaming `test_validate_npm_package_name_uppercase_rejected` → `_accepted` was correct. The old name contradicted its own body, which asserted `is_ok()` — a mislabelled test, not a behaviour change.

Issue #3186, opened 2026-08-08
PyPI: project names are coerced, never validated (PEP 508) — the root of the normalizer-divergence class


brandonrc:
Follow-up to #3183 and the #3179 metadata fix. Both of those make the *policy* string and the *fetch* string agree. This issue is about why they could disagree at all.

## The spec position

**PEP 503** defines normalization as exactly:

```python
re.sub(r"[-_.]+", "-", name).lower()
```

That collapses runs of `-`, `_`, `.` and lowercases. It says nothing about any other character — because it assumes the input is already a **valid** project name.

**PEP 508** defines what valid means:

```
^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$   (case-insensitive)
```

So `acme sdk`, `acme!sdk`, `acmeésdk` are **not project names**. There is no correct normalization for them, because they are not in the input domain.

## What we do instead

Neither of our two normalizers rejects. Both coerce, and they coerce differently:

| input | `normalize_pep503` (gate) | `PypiHandler::normalize_name` (fetch) |
|---|---|---|
| `acme-sdk` | `acme-sdk` | `acme-sdk` |
| `acme sdk` | `acmesdk` | `acme-sdk` |

`normalize_pep503` drops characters outside `[A-Za-z0-9._-]` without marking a separator — a deliberate XSS boundary, and correct as a *sanitizer*. `normalize_name` maps every non-alphanumeric to `-`. Neither matches PEP 503, because PEP 503 has no opinion here.

Grepping for `is_valid_project_name` / any PEP 508 pattern returns nothing. We have two coercions and zero validation.

## Why that's the actual root cause

Every fix so far has been "make both sides use the same coercion." That works, but it's fragile by construction: any third code path that needs a project name has to *know* which of the two helpers is the blessed one, and picking wrong is a silent security bug rather than a compile error. That has now happened at least three times (#3077 suffix stripping, #3179 metadata, #3183 download).

Validating at the edge eliminates the class:

1. Reject a project segment that fails the PEP 508 pattern with **404** (matching pypi.org, which has no route for an invalid name) before any gate or fetch runs.
2. Then a single `NormalizedProjectName` newtype, constructible only from a validated name, so a raw `&str` cannot reach a fetch helper at all.

After that there is no divergence to keep in sync, because there is only one legal input shape.

## Compatibility note

[...]

PR #3199, opened 2026-08-08
feat(curation): verify publisher attestations

### LINKED ISSUE #2955: publisher_trust: cryptographically verify attestations before granting trust (presence != trust)
Follow-up from the #2952 red-team (ship-blocker, deferred half). `publisher_trust match:attestation` currently grants trust (`verified=true`) on the **structural presence** of a provenance field (`provenance.attestation_bundles[].publisher.repository` for PyPI, `dist.attestations.provenance` for npm) — it never verifies the attestation envelope. A forged/planted provenance blob is therefore trusted ("attestation theater").

**Required:** in attestation mode, cryptographically verify the bundle — verify the sigstore/DSSE envelope and bind the certificate identity (Fulcio SAN / OIDC issuer) to the claimed repository owner — before setting verified=true (PyPI PEP 740 attestations; npm sigstore provenance). Presence of the field must never equal trust.

Interim (shipping in #2952): attestation mode is made **fail-safe** — unverified structural presence is treated as untrusted (blocks/flags like self-asserted metadata), so nothing false ships. This issue tracks the real verification, which pairs naturally with wiring pypi/npm catalog ingestion (currently rpm/debian only). Until this lands, do NOT advertise 'verified provenance'.

cazlo:
## Summary

I added on-demand PyPI and npm curation ingestion so publisher trust rules can evaluate packages seen by the proxy.

PyPI PEP 740 attestations are now verified in-process with sigstore-rs. A package is marked verified only after the DSSE signature, Fulcio chain, SCT, Rekor entry and inclusion proof, integrated time, OIDC issuer, certificate identity, subject digest, and publisher owner all match. Results are persisted on the curation package.

sigstore-rs 0.14 does not verify the Rekor inclusion proof in `verify_digest`, so the verifier also calls the crate's `InclusionProof::verify` primitives. The offline red-team fixtures cover forged proofs, signatures, identities, replayed subjects, time bounds, malformed input, and size limits.

npm ingestion is included, but npm attestations remain unverified and fail safe because their subjects use sha512. The upstream limitation is tracked in [sigstore-rs#596](https://github.com/sigstore/sigstore-rs/issues/596), with a fix proposed in [sigstore-rs#615](https://github.com/sigstore/sigstore-rs/pull/615).

The migration-repair test now disables migrator locking only for its isolated scratch database's intentionally failing checksum run. sqlx 0.8.6 returns on `VersionMismatch` without releasing the session advisory lock, which could block the later green migrator run.

## Validation

- `cargo +1.95.0 fmt --all -- --check`
- `cargo check --workspace`
- PostgreSQL 16 full library suite: 14,791 passed, 6 skipped
- CI coverage suite: 14,789 passed, 8 skipped; 88% new-code coverage
- Backend integration shards, smoke tests, iand web E2E passed

Fixes #2955


brandonrc:
## Review + remediation

I took this one over, reviewed it adversarially, and pushed fixes to the branch. Summary first: **the verification core is sound and does what the PR says it does** — I confirmed the sigstore-rs claim against the crate source *and* by execution, and every red-team fixture rejects for the right reason. The problems I found are all in the plumbing around it, plus one migration collision that would have broken startup.

Below: what I verified, the revert-proof output, what I changed, and what I deliberately left as follow-ups.

---

### 1. The sigstore-rs 0.14 claim — CONFIRMED, twice

The PR states `verify_digest` does not verify the Rekor inclusion proof. It is exactly right, and it is still true on `main` upstream:

```rust
// sigstore-0.14.0/src/bundle/verify/verifier.rs:196-202
// 5) Verify the inclusion proof supplied by Rekor for this artifact,
//    if we're doing online verification.
// TODO(tnytown): Merkle inclusion; sigstore-rs#285

// 6) Verify the Signed Entry Timestamp (SET) supplied by Rekor for this
//    artifact.
// TODO(tnytown) SET verification; sigstore-rs#285
```

What the crate *does* do at step 4 is `CheckedBundle::tlog_entry`, which only checks the tlog body is *consistent* with the bundle (`envelopeHash`, `payloadHash`, signature, verifier cert) — the CVE-2022-36056 defence. A forger controls every one of those inputs end to end, so consistency is not inclusion.

The hand-rolled call is correct and reached. `rekor_glue::verify_inclusion` selects the log key by the entry's own `logId.keyId` out of the pinned trusted root, then `InclusionProof::verify` does all three things that matter (`sigstore-0.14.0/src/rekor/models/inclusion_proof.rs`): requires a checkpoint, verifies the checkpoint signature against that key, calls `is_valid_for_proof` to bind the proof's root+size to the *signed* root, and only then runs the RFC 6962 recomputation. That is assembly of the crate's own vetted primitives, not new crypto — which is what the scope guardrail on #2955 asked for.

I also confirmed the reachability the hard way, below.

### 2. Every mutation fixture — per-fixture result

Offline, against the pinned vendored root. `mask` is the coverage bitmask, LSB = `BundleWellFormed`.

| Fixture | verdict | mask | failing check / reason |
|---|---|---|---|
| *(control: good bundle)* | **verified** | `1111111` | — |
| `flipped-signature-byte` | failed | `0000011` | signature/certificate chain: signature verification failed |
| `flipped-payload-byte` | failed | `0000001` | subject binding: subject sha256 `fa45b207…` != artifact `f045b207…` |
| `forged-inclusion-proof` | failed | `0000111` | transparency (Rekor inclusion): checkpoint envelope decode |
| `forged-merkle-path` | failed | `0000111` | transparency: `Inclusion Proof error: MismatchedRoot` |
| `roothash-vs-checkpoint-mismatch` | failed | `0000111` | transparency: `Consistency proof error: MismatchedRoot` |
| `forged-set-and-proof` | failed | `0000111` | transparency (Rekor inclusion): checkpoint envelope decode |
| `no-inclusion-proof` | failed | `0000011` | bundle: must contain inclusion proof (crate rejects first) |
| `tampered-tlog-body` | failed | `0000011` | signature: transparency materials are inconsistent |
| `integrated-time-out-of-window` | failed | `0000011` | certificate: expired before time of signing |

No fixture passes verification. One nuance worth recording, because the test names overstate it slightly: **4c and 4f also mangle the checkpoint envelope**, so the glue rejects them at checkpoint *decode*, not at the Merkle arithmetic. Only **4d** (hashes rewritten, checkpoint intact) exercises RFC 6962 recomputation and only **4e** (rootHash rewritten, checkpoint intact) exercises the checkpoint↔proof root binding. I've pinned the specific error string in both so they can't be satisfied by an incidental parse failure, and documented which row proves which mechanism.

### 3. Revert-proof

**Mutation A — delete the crate's `verify_digest` call** (keep the mask bit set):

```
FAIL row2a_flipped_signature_fails_crypto
FAIL row3_self_signed_non_fulcio_chain_rejected
FAIL row4a_inclusion_proof_stripped_rejected
FAIL row4b_tampered_tlog_body_rejected
FAIL row9_cert_outside_validity_at_integrated_time
Summary: 24 tests run: 19 passed, 5 failed

FIXTURE flipped-signature-byte          state=verified  mask=1111111 err=-
FIXTURE integrated-time-out-of-window   state=verified  mask=1111111 err=-
```

**Mutation B — delete `rekor_glue::verify_inclusion`**, i.e. rely on sigstore-rs 0.14 alone as the crate intends:

```
FAIL row4c_forged_inclusion_proof_caught_by_our_glue_not_the_crate
FAIL row4d_forged_merkle_path_rejected_by_glue
FAIL row4e_roothash_vs_checkpoint_mismatch_rejected_by_glue
FAIL row4f_forged_set_and_proof_rejected_by_glue
Summary: 24 tests run: 20 passed, 4 failed

FIXTURE forged-inclusion-proof            state=verified  mask=1111111 err=-
FIXTURE forged-merkle-path                state=verified  mask=1111111 err=-
FIXTURE roothash-vs-checkpoint-mismatch   state=verified  mask=1111111 err=-
FIXTURE forged-set-and-proof              state=verified  mask=1111111 err=-
```

Four wholly forged inclusion proofs reach `verified` without the glue. Both guards are load-bearing, and the PR's headline claim is demonstrated rather than asserted. On the unmodified tree all 26 tests pass.

**Mutation C — remove the marker sanitization I added** (see below): `test_upsert_package_strips_self_asserted_verification_marker_db` FAILS with the hostile blob persisted verbatim, `{"_ak_attestation_verification":{"owner":"Microsoft","state":"verified"},...}`.

### 4. npm — confirmed FAIL SAFE, and now tested

npm attestations fail safe; they are not silently verified. Two independent walls, and either one moving would matter, so both are now asserted:

1. **Ours** — npm binds its subject with `sha512` only and names it by purl (`pkg:npm/sigstore@5.0.0`). `subject_of` requires `subject[0].digest.sha256`, so it dies at `Check::SubjectDigestBound` before any crypto runs.
2. **The crate's** — `CheckedBundle::try_from` calls `InTotoStatementV1::subject_sha256_digest()`, which errors without a `sha256` entry, so the bundle does not even construct.

The existing `npm_never_overclaims` only proved a hardcoded stub returns `Failed`, which says nothing about real material — and the three captured npm fixtures in `testdata/` were unused. New `npm_real_bundle_cannot_reach_a_verified_state` drives the real `sigstore@5.0.0` bundle through **both** doors into the verifier (bundle path against three different artifact byte strings; provenance-document path), asserts the fixture genuinely has the sha512/no-sha256 shape so it can't pass by being malformed, asserts `verified_marker()` is `None` in every case, and carries a **positive control** on the PyPI bundle in the same fixture so the negative assertions can't be satisfied by a broken harness. Plus `npm_provenance_ingestion_never_yields_a_verified_publisher`, end-to-end from the real packument through `build_npm_curation_entry` to `extract_publisher`.

---

## Blockers found and fixed

**1. Migration 193 collides with `main` — would abort backend startup.** `main` has taken both 193 (`193_oci_blobs_storage_key_index.sql`) and 194 (`194_oci_cleanup_keys_pending_delete.sql`) since this branch was cut. Different filenames means no textual conflict, GitHub reports `mergeable: clean`, and nothing catches it at merge time: sqlx `resolve` sorts by version without deduping, so both entries take the `apply` arm on a fresh database (duplicate key on `_sqlx_migrations.version`) or trip `VersionMismatch` on an existing one. Either way the backend aborts before any newer migration runs — the same duplicate-slot class `migration_repair.rs` exists to clean up after. **Renumbered to 195**, with the history recorded in the file header. Verified applying cleanly on `postgres:16-alpine`.

No `CREATE INDEX CONCURRENTLY` (the string appears only in an explanatory comment), no other statement that cannot run in a transaction, `IF NOT EXISTS` throughout, and `DEFAULT 'unverified'` with a `CHECK` on the three-value domain is the fail-safe direction. No new `query!`/`query_as!` macros anywhere in the diff, so no `cargo sqlx prepare` needed — `record_attestation` deliberately uses the runtime API and says so.

**2. The PyPI on-demand enqueue ran before authorization and before any upstream contact.** In `download_or_metadata`, the enqueue sat immediately after `enforce_pypi_curation`. That helper is a content gate, not an access gate, and `resolve_pypi_repo` is a bare key lookup — the repository access check lives inside `serve_file`, ~50 lines later. And nothing had contacted the upstream yet: the name and version come straight out of the request path via `normalize_pep503` / `version_from_pypi_filename`. So `GET /pypi/{remote}/simple/anything/anything-1.0-py3-none-any.whl` wrote a `curation_packages` row for a package that does not exist, into a staging repo belonging to whoever pointed `curation_source_repo_id` at that proxy, with no cap and no rate limit, from an anonymous caller on a public repo. Unbounded catalog inflation plus a poisoned review queue.

**Moved after `serve_file` returns a served response.** The npm seam was already correctly shaped this way (it enqueues only after a version document resolves out of a real packument) — this brings PyPI in line. Note the gate is `is_success() || is_redirection()`, not `is_success()` alone: #1555 answers a fresh proxy-cache hit on a remote member with a 307 to a presigned URL, so gating on 2xx would have silently switched ingestion off for every operator running object-storage downloads.

## Also fixed

**3. Attacker-controlled path components interpolated unencoded into credentialed upstream URLs.** `verify_pypi_curation_row` built `{base}/pypi/{name}/{version}/json` and `{base}/integrity/{name}/{version}/{filename}/provenance` and fetched both with `apply_upstream_auth`. `version_from_pypi_filename` returns whatever sits between the dashes with no charset check, so `foo-1.0?x-py3-none-any.whl` injects a query string and `foo-..-py3-none-any.whl` traverses a path segment (`Url::parse` normalises `..` away) — a credentialed request-forgery primitive against arbitrary paths on an internal mirror. Added `is_safe_upstream_path_segment`, rejecting rather than encoding: every value that can legitimately appear is already inside PEP 503 normalized names / PEP 440 versions / PEP 427-625 filenames, so a value outside them is not a package we can verify anyway. Unit-tested in both directions, including that real-world identifiers like `2!1.0.dev4` and `ruamel.yaml.clib-0.2.8-cp312-cp312-manylinux_2_17_x86_64.whl` still pass — a false negative here fails quietly, which is worse than failing loudly.

**4. Upstream-chosen `dist_url` fetched with the operator's upstream credentials attached.** `_ak_dist.url` is copied verbatim out of the upstream's own JSON (`urls[].url`), so a hostile or compromised mirror picks it. Connect-time SSRF is already refused by the shared DNS guard, but nothing stopped the configured Basic/token header being handed to whatever public host the mirror named. Credentials are now gated on `same_upstream_origin`, same rule as the NuGet remote-search fix (#3130/#2925), fail-closed on anything unparseable. Unit-tested. (Public PyPI serves distributions from `files.pythonhosted.org`, a different origin from `pypi.org`, so that download simply goes unauthenticated — which is correct.)

**5. The verification marker was never stripped, and the comment saying it was is now true.** `evaluate_ondemand_curation` inserted `_ak_attestation_verification` on success but never removed a pre-existing one, so on the failure and npm paths anything already in the stored blob survived into evaluation — and `extract_publisher` short-circuits on that marker and returns `verified = true` for whatever owner it names, landing directly on the `action: allow` arm.

I want to be precise: **this is not currently exploitable.** Both ingestion builders construct metadata from a fixed key allowlist rather than passing upstream JSON through, so a hostile registry has no channel to plant the key today. But the security property rested entirely on that distant invariant while `publisher_source.rs` asserted the opposite ("a package's own metadata can never set it"), and any future ingestion path that stores richer metadata converts silently into an attestation-forgery bypass. Added `strip_verification_marker` and called it at both chokepoints — the evaluation context, and `CurationService::upsert_package`, which is the single production write path for `curation_packages.metadata`. Now true by construction. Revert-proved above; the unit test carries its own positive control (it asserts the planted blob *is* trusted before stripping, so it cannot pass in a world where nothing is trusted).

**6. `signal_label` told operators a verified attestation was unverified.** Reachable via the watch-mode arm. Now keyed on `publisher.verified` too.

**7. Doc accuracy.** The migration header claimed the persisted record is what lets `extract_publisher` emit `verified=true`; it isn't — nothing reads those columns back (`CurationPackage` doesn't carry the fields), the decision is made from the in-memory verdict in the same tick. Corrected, with a pointer to the follow-up. Also added a "known residual" section to the `attestation_verify` module doc about `integratedTime` (see #3231).

## Not changed, filed instead

- **#3230** — the persisted attestation record is write-only, so re-evaluating a row loses `verified=true` and an `approved` package silently returns to `review`. Fixing it properly means making the columns authoritative and deciding a re-verification interval; too much to bolt onto this PR.
- **#3231** — Rekor SET is not verified, so `integratedTime` is unauthenticated and the "cert was valid when it signed" check is corroborating rather than independent. Low severity (satisfying it still needs a Fulcio cert naming the claimed publisher), and per the #2955 guardrail the right answer is upstreaming, not hand-rolling.
- **#3232** — unblocking npm when sigstore-rs#615 lands, with notes on what the npm path needs beyond the version bump (purl subject-name binding, an npm reader in `bundle_convert`).
- **#3233** — a handler-level regression test for the enqueue gate in fix 2, plus a per-repo row cap. I could not write the test in this PR: `download_or_metadata` is not exercised directly by the existing tests and the connect-time SSRF guard refuses a `wiremock` loopback upstream, so the usual remote-proxy harness does not reach the path. Flagging that gap explicitly rather than leaving it implied.

## Minor, left alone

- `MAX_PENDING_PER_TICK = 500` with serial per-row network I/O inside a sequential repo sweep — a backlog delays other staging repos. Bounded in practice because a row leaves `pending` after one evaluation. Covered by #3233.
- `ON CONFLICT DO UPDATE SET metadata = EXCLUDED.metadata` lets a later proxy request clobber an already-approved row's richer metadata with the thin `{name, version, _ak_dist}` blob. Status is preserved, so this costs audit context, not a decision.
- A package that failed verification because the upstream was briefly down is parked in `review` with no retry.
- I reordered the enrichment fetches so provenance is fetched before the JSON metadata — provenance is absent for the overwhelming majority of distributions, and the early return on that miss now saves the JSON round-trip entirely.
- The `.bind(x.to_string_lossy().as_ref())` → `.bind(&*…)` lint churn spans 17 files and dominates the file count here, burying the ~120 lines that matter. All of it is semantically identical and ~95% in test code — I checked every hunk, including the four in `security_regression_tests.rs`, and no assertion was weakened. Worth its own commit next time.

## Verified, no finding

- **`migration_repair.rs` `set_locking(false)` is test-only and NOT production-reachable.** All seven added lines are inside `#[cfg(test)] mod tests`, in a helper operating on a per-test `CREATE DATABASE` scratch database; `set_locking` appears exactly once in the whole repo. Production keeps the default at `main.rs`. The justification also checks out: in sqlx 0.8.6 `run_direct` takes the lock and then has three early returns that bypass the unlock — `Dirty`, `validate_applied_migrations`, and `VersionMismatch` — and the Postgres lock is *session*-scoped (`SELECT pg_advisory_lock($1)`), so the connection goes back to the pool still holding it. Real hazard, correctly scoped.
- The ~12 touched format handlers share no signature change at all — it is entirely the lint churn above. No threaded-but-unused parameter.
- The 11 positional args from `enqueue_curation_on_demand` line up correctly with `upsert_package`.
- Enqueue failures are logged and swallowed, and the task is detached, so ingestion cannot break a download. Correct.
- The scheduler job inherits the existing cluster lease on `run_curation_sync_cycle`; no double-run.
- Not dead code and not behind a flag that defaults off.

---

## Validation

`cargo fmt --all` · `SQLX_OFFLINE=true cargo build --lib --tests` · `cargo clippy --workspace --all-targets -- -D warnings` clean · full `cargo nextest run --workspace --lib` against a freshly-migrated PostgreSQL 16.

---

## Credit where it is due

This is careful work and it is worth saying which parts specifically. Going to the crate source, finding the `TODO(tnytown)`, and then *not* concluding "sigstore-rs verifies it for me" is the exact failure this issue was filed to prevent, and it would have been very easy to miss. Assembling the crate's own `InclusionProof::verify` / `rekor_keys` / `CosignVerificationKey` primitives instead of writing Merkle code is the right call under the scope guardrail. The coverage bitmask with `from_mask` as the single place that can mint `Verified` is a genuinely good structural idea — it survived both of my mutations by failing closed rather than silently, and it made the revert-proof legible. Ordering the subject-digest compare *before* the crate call so a replay reports as a replay rather than as the crate's misleading `Transparency` error is a real usability win for whoever reads these reasons at 3am. And reporting the npm limitation honestly, with the upstream issue and PR linked, rather than shipping a path that half-works, is exactly the judgement call the guardrail comment asked for.

The blockers I found are all around the edges — a migration slot that moved under the branch, and an ingestion seam placed a few lines too early. The security core held up to everything I threw at it.

brandonrc:
## Rebased onto current main

The branch had gone dirty. Rebased and force-pushed (`58eb46b` → `f08881c`); the author's three commits are unchanged, and my review commit carries one deliberate conflict resolution.

### The conflict: `serve_file` in `pypi.rs`

#3227 / #3186 landed between the branchpoint and now. They replace the raw `&project` path segment with a validated `NormalizedProjectName` newtype (`parse_project_segment`), and change `serve_file` to take `&NormalizedProjectName` instead of `&str`. My review fix had rewritten the same call to bind its result, so the on-demand curation enqueue could gate on a served status.

Resolved by keeping **both** sides: main's validated-name argument and its explanatory comment verbatim, plus the response binding. Nothing from #3186 was dropped — `normalize_pep503` is untouched at `pypi.rs:139` (its character dropping is the deliberate stored-XSS boundary and stays exactly as it was), `parse_project_segment` remains the single validator, and I did not add a second one.

One consequence is worth calling out, because git would not have caught it and the compiler did. The enqueue block built its catalog row from `normalized`, which is now a newtype with no `Deref`, no `AsRef<str>` and no `Serialize` — deliberately, per that module's doc. It now reads `normalized.as_str().to_string()` explicitly. **That is the right value independently of the rebase**: the curation catalog is a gate input, so recording the raw path segment while every gate above keys on the validated canonical name would have reintroduced precisely the gate-on-one-string / fetch-another divergence #3186 exists to make unrepresentable. This is the fourth code path that module was written to catch, and it caught it at compile time rather than in a review.

### Verifying the auto-merges

Auto-merge results here are not trustworthy by default, so I checked rather than assumed:

- **`scheduler_service.rs` auto-merged with +460 / −0** against current main. Purely additive, so nothing of #3228's format-handler registry work was clobbered.
- Across the whole branch, the **only** lines removed relative to main are: the PR's own `.bind(x.to_string_lossy().as_ref())` → `.bind(&*…)` lint churn, the three `pypi.rs` lines rewritten deliberately above, `publisher_trust.rs`'s `signal_label` match, and one line in `migration_repair.rs`. Every other file is additive-only.
- **`pypi.rs` has exactly one `enqueue_curation_on_demand` call site**, at the post-serve position. The rebase did not leave a duplicate from the earlier commit that introduced it at the pre-auth position — which is the failure mode that would have quietly reinstated the bug I had just removed.
- For every file main did **not** touch (the whole `attestation_verify` module, `curation_service.rs`, `curation_sync.rs`, `publisher_source.rs`, `migration_repair.rs`, the migration), the branch's delta is byte-identical before and after the rebase.

### The migration renumber now has a positive control

`main` carries 193 (`oci_blobs_storage_key_index`) and 194 (`oci_cleanup_keys_pending_delete`); this branch carries 195, and `sqlx migrate run` applies all three in order on a fresh `postgres:16-alpine`. Had the file still been at 193, this rebase would have produced two migrations at the same version with no textual conflict and no CI signal — which is exactly why it was worth renumbering before the rebase rather than discovering it after.

### Re-verification on the rebased tree

- `cargo fmt --all` — no changes
- `SQLX_OFFLINE=true cargo build --lib --tests` — clean
- fresh `postgres:16-alpine`, all 195 migrations applied in order
- attestation / curation / publisher / scheduler subset — **267 passed**
- full `cargo nextest run --workspace --lib` with `DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` set, so the DB-backed tests hard-fail rather than skipping — **15,028 passed, 0 failed, 7 skipped**

The `AK_TESTS_REQUIRE_DB=1` part matters for reading that number: without it the DB tests return PASS in ~0.04s without executing, and the marker-sanitization guard would have looked green while proving nothing. In this run `test_upsert_package_strips_self_asserted_verification_marker_db` took 1.6s, i.e. it really ran.

Main has since moved on again to `e82e3d2` (#3200), but GitHub reports the PR `mergeable: true` against it, so no further rebase is needed right now.

wolfv:
Would love to hear if `sigstore-rust` works for your use case: https://github.com/sigstore/sigstore-rust

cazlo:
@wolfv thank you for the idea. I think this would close the npm sha-512 gap we have no on first glance. Also signstore-rust using FIPS 140-3 is pretty great. It's definitely worth a look. I'll try and check this out this weekend.

Issue #2503, opened 2026-07-14
Harden SA-grant path: validate principal_type/principal_id correspondence + surface direct service_account grants on target_type=group


brandonrc:
Follow-ups from the #2433 red-team validation (PR #2499, both fail-closed, non-blocking):

1. **Defense-in-depth (LOW):** `create_permission` does not validate that `principal_id` corresponds to a principal of the declared `principal_type`. Since #2433 widened grant matching to include `service_account`, a mistyped grant (e.g. `principal_type='service_account'` naming a real *user* id) becomes effective. Not attacker-reachable (ids are globally-unique UUIDs; grant creation is admin-only), but a type/id-correspondence check in `create_permission` would prevent admin data-entry mistakes.
2. **Functional gap:** a *direct* `service_account`-typed grant with `target_type='group'` does not surface the group in `list_groups`/`get_group` for that SA (under-grants, fails closed). The `groups.rs:53` widening in #2433 covers group-mediated repo access but not this direct-SA-grant-to-group listing path.

Origin: #2433 / PR #2499 adversarial isolation validation (verdict CLEAN; these are hardening/UX, not isolation holes).

PR #3200, opened 2026-08-08
fix(permissions): populate service-account principal names

### LINKED ISSUE #2826: Adding permissions for service accounts through the api does not populate the principal_name field in the permissions response
### Description

Related to https://github.com/artifact-keeper/artifact-keeper-web/issues/621#issue-4951479148 with evidence shown there

### Environment Information

AK backend 1.6.1
AK Web 1.5.8

### Backend Version

1.6.1

### Web UI Version

1.5.8

### Package Format

None

### Expected Behavior

Principal name should be populated in the permissions response, this might be an issue on save already where the db record is not populated

### Observed Behavior

principal_name is null for newly added service_account permissions

### Steps to Reproduce

See description

### Logs

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

cazlo:
## Summary

- resolve service-account usernames in permission list and get responses with a `users` join keyed by principal id, matched to how permission enforcement itself resolves the row
- return the same hydrated permission response immediately after create and update, each collapsed into a single autocommit CTE statement instead of a four-round-trip transaction
- add a DB-backed response matrix for human users, service accounts, groups, legacy `principal_type = 'user'` rows naming a service account, repository targets, and global targets
- dedupe the SELECT/CASE/JOIN hydration SQL into one shared fragment reused by list, get, create, and update, dropping the file's jscpd duplication from 3.98% to within the 3% gate

## Why

Service accounts share the `users` table with human users, but permission response queries only joined `users` for `principal_type = 'user'`. As a result, service-account permissions returned `principal_name: null`; create and update also hard-coded both display-name fields to null.

I initially guarded the join with `users.is_service_account = TRUE/FALSE` per principal type, on the theory that a mismatched type/ID pair (e.g. `principal_type = 'service_account'` naming a human user's id) shouldn't acquire a display name. Review caught that this guard actively hides real grants: migration 057 (Feb 2026) introduced `is_service_account`, but `principal_type = 'service_account'` wasn't honored by any resolver until #2499 (Jul 2026). In that window the only way to grant a service account fine-grained access was `principal_type = 'user'` + the SA's `users.id`. No migration re-types those rows, `validate_principal` (#2503) only guards new writes, and `query_actions` (`permission_service.rs`) still matches such rows by bare `principal_id` — they are live, enforced grants today. The type-guarded join rendered exactly those rows `principal_name: null`, hiding from admins who a still-effective grant empowers. I removed the guard: the join is keyed by the `users` primary key, so a guardless join can never attach the wrong name, only a missing one. Display now matches enforcement. A regression test pins the legacy direction (`principal_type = 'user'` naming a service account); the existing mismatched-type test flips from asserting `null` to asserting the real username, with a comment explaining why.

Create and update reuse the same shared hydration SQL as list and get, each collapsed into one `WITH ... AS (INSERT|UPDATE ... RETURNING *) SELECT ...` statement on the pool instead of BEGIN / INSERT-or-UPDATE / re-SELECT / COMMIT, so they no longer hold a connection and row lock across four round trips. On create, a zero-row result from the CTE is an invariant failure (mapped to a 500), not a legitimate 404 — the INSERT always returns exactly one row on success. On update, a zero-row result still means the id doesn't exist and stays a 404.

No migration, frontend change, or response-field schema change is required.

## Validation

- `cargo fmt --check`
- `cargo nextest run` filtered to the permissions handler tests, DB-backed cases confirmed to run (not skipped) against a local postgres
- jscpd on `backend/src/api/handlers/permissions.rs` at CI's `--min-lines 10 --threshold 3` settings, measured locally with `--max-lines` raised past the file's line count (jscpd's default 1000-line cap otherwise skips it): 1.68% (23/1367 lines, 2 clones), down from 3.98% pre-fix

fixes #2826

### DIFF
--- a/backend/src/api/handlers/permissions.rs
+++ b/backend/src/api/handlers/permissions.rs
@@ -88,6 +88,53 @@ impl From<PermissionRow> for PermissionResponse {
     }
 }
 
+/// Shared SELECT/JOIN fragment that hydrates a `permissions` row's
+/// `principal_name` / `target_name` display columns.
+///
+/// `from_clause` supplies both the row source and its alias, always `p`:
+/// either the base table (`"permissions p"`, used by the plain read paths
+/// [`fetch_permission_row`] and [`list_permissions`]) or a data-modifying
+/// CTE named `p` (used by [`create_permission`] / [`update_permission`] to
+/// collapse INSERT/UPDATE + hydration into one round trip). Every caller
+/// appends its own `WHERE`/`ORDER BY`/pagination clause (or, for the CTE
+/// callers, prefixes a `WITH p AS (...)`) via `format!`. Centralizing this
+/// keeps the principal/target CASE+JOIN logic in exactly one place.
+fn hydrate_select(from_clause: &str) -> String {
+    format!(
+        r#"
+        SELECT p.id, p.principal_type, p.principal_id, p.target_type, p.target_id,
+               p.actions, p.created_at, p.updated_at,
+               CASE
+                   WHEN p.principal_type IN ('user', 'service_account') THEN u.username
+                   WHEN p.principal_type = 'group' THEN g.name
+               END as principal_name,
+               CASE
+                   WHEN p.target_type = 'repository' THEN r.name
+               END as target_name
+        FROM {from_clause}
+        LEFT JOIN users u ON p.principal_id = u.id
+                         AND ((p.principal_type = 'user' AND u.is_service_account = FALSE)
+                           OR (p.principal_type = 'service_account' AND u.is_service_account = TRUE))
+        LEFT JOIN groups g ON p.principal_type = 'group' AND p.principal_id = g.id
+        LEFT JOIN repositories r ON p.target_type = 'repository' AND p.target_id = r.id
+        "#
+    )
+}
+
+async fn fetch_permission_row<'e, E>(executor: E, id: Uuid) -> Result<Option<PermissionRow>>
+where
+    E: sqlx::Executor<'e, Database = sqlx::Postgres>,
+{
+    sqlx::query_as(&format!(
+        "{} WHERE p.id = $1",
+        hydrate_select("permissions p")
+    ))
+    .bind(id)
+    .fetch_optional(executor)
+    .await
+    .map_err(|e| AppError::Database(e.to_string()))
+}
+
 #[derive(Debug, Serialize, ToSchema)]
 pub struct PermissionListResponse {
     pub items: Vec<PermissionResponse>,
@@ -135,30 +182,16 @@ pub async fn list_permissions(
         }));
     }
 
-    let permissions: Vec<PermissionRow> = sqlx::query_as(
-        r#"
-        SELECT p.id, p.principal_type, p.principal_id, p.target_type, p.target_id,
-               p.actions, p.created_at, p.updated_at,
-               CASE
-                   WHEN p.principal_type = 'user' THEN u.username
-                   WHEN p.principal_type = 'group' THEN g.name
-               END as principal_name,
-               CASE
-                   WHEN p.target_type = 'repository' THEN r.name
-               END as target_name
-        FROM permissions p
-        LEFT JOIN users u ON p.principal_type = 'user' AND p.principal_id = u.id
-        LEFT JOIN groups g ON p.principal_type = 'group' AND p.principal_id = g.id
-        LEFT JOIN repositories r ON p.target_type = 'repository' AND p.target_id = r.id
-        WHERE ($1::text IS NULL OR p.principal_type = $1)
+    let permissions: Vec<PermissionRow> = sqlx::query_as(&format!(
+        "{} WHERE ($1::text IS NULL OR p.principal_type = $1)
           AND ($2::uuid IS NULL OR p.principal_id = $2)
           AND ($3::text IS NULL OR p.target_type = $3)
           AND ($4::uuid IS NULL OR p.target_id = $4)
         ORDER BY p.created_at DESC
         OFFSET $5
-        LIMIT $6
-        "#,
-    )
+        LIMIT $6",
+        hydrate_select("permissions p")
+    ))
     .bind(&query.principal_type)
     .bind(query.principal_id)
     .bind(&query.target_type)
@@ -210,18 +243,6 @@ pub struct CreatePermissionRequest {
     pub target_type: String,
     pub target_id: Uuid,
     pub actions: Vec<String>,
-}
-
-#[derive(Debug, FromRow, ToSchema)]
-pub struct CreatedPermissionRow {
-    pub id: Uuid,
-    pub principal_type: String,
-    pub principal_id: Uuid,
-    pub target_type: String,
-    pub target_id: Uuid,
-    pub actions: Vec<String>,
-    pub created_at: chrono::DateTime<chrono::Utc>,
-    pub updated_at: chrono::DateTime<chrono::Utc>,
 }
 
 /// Create a permission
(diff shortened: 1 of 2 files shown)


brandonrc:
## Review + remediation — pushed `50f7328`

Reviewed adversarially against merge base `0a0d80a` with the head pinned at `25b01d2`, four independent review passes (security / SQL+schema / correctness / test analysis), and everything below executed against a dedicated Postgres 16 with all 194 migrations applied. **The fix is correct and the diagnosis is right.** I pushed one commit for three real gaps plus some comment accuracy.

### Stage 1 — the bug reproduces, and the fix closes it

Merge-base read SQL vs head read SQL against the same three seeded rows:

```
===== MERGE-BASE (pre-fix) =====
 id     | principal_type  | principal_name | target_name
 ...001 | user            | rev-human      | rev-repo
 ...002 | service_account |                | rev-repo     <-- #2826
 ...003 | user            | rev-sa         | rev-repo

===== PR HEAD (post-fix) =====
 ...002 | service_account | rev-sa         | rev-repo
```

Create/update were separately returning hard-coded `null` for both display fields; the CTE fixes that. `principal_name` is produced nowhere else in the backend, so the fix is complete for the field.

### What the review confirmed as correct

Recording these because they were the plausible failure modes and they all check out:

- **The `users` join cannot attach a wrong name.** All three joins key on a primary key (`users.id`, `groups.id`, `repositories.id`), so at most one match each — no cross-attachment and no row fan-out, which also means the separate `COUNT(*)` stays consistent with `items`.
- **The rationale for dropping the type guard holds against source.** `query_actions` (`permission_service.rs`) matches `principal_type IN ('user','service_account') AND principal_id = $1`, i.e. by bare `principal_id`. Both a legacy `'user'` row naming a service account and a `'service_account'` row naming a human really are live, enforced grants, so displaying the name matches enforcement. `validate_principal` gates only new writes, so neither shape is reachable via the API going forward. Display-only change, no authz effect. This was the right call and the doc comment earns its length.
- **The untyped `query_as` is safe.** All four generated statements `PREPARE` with byte-identical result types; the ten column names match `PermissionRow` exactly (`FromRow` matches by name), and both `CASE`s resolve to `character varying`, not `unknown`. No runtime-500 mismatch.
- **The CTE is genuinely atomic.** Demonstrated by forcing the outer half to fail: `WITH p AS (INSERT ... RETURNING *) SELECT 1/0 FROM p` leaves the table at 0 rows. No error path can leave a half-written grant.
- **Zero-row update still 404s** (empty CTE → empty outer SELECT), and **`duplicate key` survives the CTE wrapper** so create's 409 mapping still fires.
- **Plans are unchanged.** `EXPLAIN (ANALYZE, COSTS OFF)` on head vs base is identical apart from `Join Filter: (principal_type = 'user')` → `= ANY ('{user,service_account}')`. Still three PK index scans.
- `CreatedPermissionRow` was a dangling registered schema — no operation ever referenced it, zero remaining refs repo-wide. Removal is clean. One downstream note: the generated `@artifact-keeper/sdk` will drop that exported model, so it's worth a line in release notes even though nothing used it.

### What I changed, and why

**1. Update-path `target_name` hydration was completely unguarded.** Every update assertion in the PR ran against a `target_type = "global"` target, whose `target_name` is null under the fix **and** under the pre-#2826 code that hard-coded the field to `None`. `assert!(...is_null())` there is satisfied by the bug as well as by the fix. Proof — keep the whole fix, un-hydrate only `target_name` on the update path:

```
Mutation: `permission.target_name = None;` in update_permission only

  PASS  create_and_update_return_hydrated_permission_names        <-- as written, blind
  PASS  list_and_get_hydrate_principal_names_by_id_..._mismatch
  PASS  legacy_user_typed_grant_naming_a_service_account_...
```

I re-pointed the same row at the repository as a positive control. Same mutation, after:

```
  FAIL  create_and_update_return_hydrated_permission_names
        panicked at permissions.rs:943
          left: Null
         right: "ph-test-npm-12622992-c98c-4332-b3eb-834abef78563"
```

The other four tests still pass under that mutation, so the new assertion is the one doing the work.

**2. An update that collides on the unique key returned 500, not 409.** `permissions` is `UNIQUE(principal_type, principal_id, target_type, target_id)`. Create maps that violation to `Conflict`; update fell through to `AppError::Database` and answered `500 DATABASE_ERROR` with the raw Postgres error text. It's a client-caused conflict on both paths, so I extracted `map_permission_write_error` and share it. This is pre-existing rather than something the PR introduced — flagging it explicitly since it's the one behaviour change beyond #2826, and happy to split it out if you'd rather keep this PR to the one issue.

**3. Two behaviours the CTE rewrite is responsible for had no coverage at all.** The 409 mapping is a substring match on the driver's error text, and nothing anywhere exercised it — a change in how a unique violation surfaces from inside a CTE would have silently downgraded 409 → 500. Same for the 404 path. Added `duplicate_grants_are_conflicts_on_both_create_and_update` (which also asserts the rejected write leaves no row, pinning the atomicity claim) and `update_of_a_missing_permission_is_not_found_not_a_server_error`.

**4. Comment accuracy.** The create/update comments say the previous code held "a dedicated transaction/connection across four round trips (BEGIN, INSERT ... RETURNING id, re-SELECT via `fetch_permission_row`, COMMIT)" and constructed an "unreachable NotFound". None of that is true of `0a0d80a`: base create was a single `sqlx::query_as(INSERT ... RETURNING ...).fetch_one(&state.db)` — no transaction, one round trip, `fetch_permission_row` is introduced *by this PR*, and create constructed no `NotFound`. (I assume this described an intermediate revision of your own branch.) The round-trip count is 1 → 1; what the CTE actually buys is hydration inside the write statement, which is a good reason on its own. Rewritten to say that, plus the atomicity property and a note that the two name fields are snapshot-best-effort — the outer SELECT reads `users`/`repositories` at the statement snapshot, so a principal or repo committed concurrently mid-statement can echo back `null` and hydrate on the next GET. Worth writing down so nobody later adds a test asserting they're non-null under concurrency.

**5. Relabelled `legacy_user_typed_grant_naming_a_service_account_hydrates_its_username`.** It's green on the merge base. A service account *is* a `users` row, so the old `p.principal_type = 'user' AND p.principal_id = u.id` join already hydrated that shape — the row `...003` in the stage-1 table above shows it hydrating pre-fix. It carries no coverage of #2826 itself. I kept it, because it *is* a real forward guard, and proved that:

```
Mutation: AND u.is_service_account = (p.principal_type = 'service_account')  (the rejected type-guard design)

  FAIL  legacy_user_typed_grant_naming_a_service_account_hydrates_its_username
        left: Null   right: "ph-test-sa-df179ffc-582c-461c-8290-e344234fd32d"
  FAIL  list_and_get_hydrate_principal_names_by_id_regardless_of_type_mismatch
  PASS  create_and_update_..., duplicate_grants_..., update_of_a_missing_...
```

Its docstring now says plainly that it's green pre-fix and what it does guard, so it isn't mistaken for #2826 coverage later.

### Revert-proof — faithful full revert

Merge-base production code spliced under the branch's test module (`diff` against `0a0d80a` confirms the production section is byte-identical to the merge base):

```
Summary [0.401s] 34 tests run: 31 passed, 3 failed

FAIL  list_and_get_hydrate_principal_names_by_id_regardless_of_type_mismatch
      permissions.rs:670  left: Null  right: "ph-test-sa-cd33122c-..."
FAIL  create_and_update_return_hydrated_permission_names
      permissions.rs:808  left: Null  right: "ph-test-sa-8042c0d5-..."
FAIL  duplicate_grants_are_conflicts_on_both_create_and_update
      permissions.rs:978  "an update colliding with an existing grant must map to 409, not 500:
                           {"code":"DATABASE_ERROR","message":"Database operation failed"}"
                          left: 500  right: 409
PASS  legacy_user_typed_grant_naming_a_service_account_hydrates_its_username   (green pre-fix by design, see above)
PASS  update_of_a_missing_permission_is_not_found_not_a_server_error           (guards the rewrite, not the fix — 404 pre-existed)
```

With the fix in place: `34 tests run: 34 passed`, DB-backed cases confirmed executing (~4s each, not skipped) under `AK_TESTS_REQUIRE_DB=1`.

### Gates

```
cargo fmt --all                                        clean, no further changes
SQLX_OFFLINE=true cargo build --lib --tests            Finished (also clean at 25b01d2 as authored)
cargo clippy --workspace --all-targets -- -D warnings  exit 0
cargo nextest run --workspace --lib --test-threads 8   14846 run: 14844 passed, 2 failed
```

The two failures are `proxy_service::test_tee_publish_round_trip_on_a_real_filesystem_backend` and `rpm_publish_service::test_create_version_concurrent_racers_both_succeed_db` — neither module is touched here, and both pass on rerun in isolation (`2 tests run: 2 passed`). Parallel-run flakes on a contended shared database, not this change. No `cargo sqlx prepare` needed: every query in the diff is a runtime `sqlx::query_as(&String)`, not a `query!`/`query_as!` macro. The corollary is worth stating — because they're runtime queries there is no compile-time check that `hydrate_select`'s column list matches `PermissionRow`, so these DB-backed tests are the only guard.

### Filed separately

#3229 — `list_permissions` and `get_permission` take no `AuthExtension` at all and have no admin gate, so any authenticated principal (including a read-scoped service-account token) can page the entire ACL table; and `delete_permission` checks `require_scope("delete")` without `require_admin()`, which a JWT session satisfies unconditionally since `scopes: None` is action-unrestricted. Both predate this PR and are untouched by it — this PR does marginally widen the first by adding service-account usernames to what that already-too-broad read returns, but human usernames, group names, repository names and full action arrays were already flowing to the same callers.

One judgement call worth your eyes: the `mismatched_permission` case pins that a row declared `principal_type: "service_account"` renders a **human** user's username. That is the audit-visible truth (`query_actions` really does enforce it for that human) and I agree with the call, but it is a product decision about what the settings UI shows, not just a test detail.

Nice catch on the legacy `principal_type = 'user'` window between migration 057 and #2499 — that's the kind of thing a narrower fix would have papered over, and the reasoning for keeping the join guardless is exactly right.

brandonrc (approved):
Approving: stages 1-5 of the review pipeline passed. Bug reproduced against the merge base and confirmed closed at head; four review passes reconciled; the three findings I pushed as `50f7328` are each revert-proved (details in the review comment above); full CI green including Code Coverage/jscpd. The pre-existing authorization gaps on this router are tracked separately in #3229 and are out of scope here.

Issue #3184, opened 2026-08-08
Content-Encoding still dropped on the scan-enabled proxy serve paths (npm, PyPI)


brandonrc:
Follow-up to #3149 / PR #3176. That PR fixed the streaming builders, but two **buffered** proxied-serve builders remain, so the original symptom is still live on any repository with `scan_on_proxy` enabled.

## The gap

- `build_scanned_file_response` (`backend/src/api/handlers/pypi.rs`, called from the scanned-serve path)
- `build_scanned_tarball_response` (`backend/src/api/handlers/npm.rs`, same shape)

Both set `Content-Length` from `bytes.len()` — the **coded** length — and neither sets `Content-Encoding`. `build_scanned_tarball_response` does not even take the parameter.

They can't forward it as written, because the coding is discarded upstream of them in the signature:

```rust
// proxy_service.rs
let (content, content_type, _encoding) = fetch_artifact_with_cache_path_capped(...)
```

So the fix is to widen `fetch_artifact_with_cache_path_capped` to return the coding and thread it through both builders — a bigger change than #3176 should carry, which is why it's split out.

## Impact

On a repo with `scan_on_proxy` enabled, behind any gzipping upstream or intermediary:

- `pip install` gets a body labelled with no coding but carrying gzip bytes, and a `Content-Length` that describes the coded length
- `npm install` gets the same on the tarball path

That is exactly the #3149 symptom. #3176's own body claims "single builder behind all 6 call sites" for PyPI; there is a seventh. Worth correcting the record there so this doesn't read as closed.

Note #3176 *did* fix `npm.rs`'s streaming pending arm inside the scanned flow, which is what makes the remaining buffered arm easy to miss.

## Adjacent, surfaced while tracing this — probably worth its own issue

A coded-on-arrival npm tarball fails `npm_claimed_identity` (a single `GzDecoder`, so gzip-over-tgz never parses) → `ProxyScanIdentity::Unestablished` → `proxy_helpers.rs` returns `Serve { pending: true }` **before** the `tokio::spawn`. No scan runs, no verdict row is ever written, permanently. Bounded to scan-enabled repos on the default `fail_open` action, but it is the same coded-bytes-meet-identity-assumption class.

## Suggested test

[...]

PR #3210, opened 2026-08-09
fix(pypi): forward upstream Content-Encoding on PEP 658 remote metadata, and decode coded wheels before extraction

### LINKED ISSUE #3193: PyPI PEP 658 serve_remote_metadata drops upstream Content-Encoding, and its wheel fallback cannot parse a coded wheel
Split out of #3184, where it was flagged while tracing the buffered proxied-serve builders. It is the same "coded bytes meet an assumption of plaintext" class as #3149 / #3184, but it does **not** fit in that fix, for two independent reasons.

## The gap

`serve_remote_metadata` (`backend/src/api/handlers/pypi.rs`) serves PEP 658 `.metadata` bytes for a Remote PyPI repo. Both of its arms are affected.

### Arm 1 — direct `.metadata` fetch: coding is dropped

```rust
match proxy.fetch_upstream_direct_with_link(&remote_repo, &target.fetch_path).await {
    Ok((content, _content_type, _)) => {
        Ok(Response::builder()
            .status(StatusCode::OK)
            .header(CONTENT_TYPE, "text/plain; charset=utf-8")
            .body(Body::from(content))
            .unwrap())
    }
```

The bytes are forwarded verbatim, but no `Content-Encoding` is set. `fetch_upstream_direct_with_link` returns `(Bytes, Option<String>, Option<String>)` where the third element is the **`Link`** header, not the coding — the coding is not in the return type at all, so as with #3184 the handler cannot forward what it never receives. Since the shared HTTP client no longer lets reqwest decode upstream bodies, a coded `.metadata` resource reaches the client as compressed bytes labelled `text/plain; charset=utf-8`.

### Arm 2 — wheel-extraction fallback: needs decoding, not forwarding

On upstream 404 the handler falls back to fetching the wheel and extracting `METADATA` from it:

```rust
let (wheel, _) = proxy.fetch_artifact_with_cache_path(&wheel_repo, ...).await?;
let metadata = crate::util::bounded_archive::with_ingest_extraction(|| extract_metadata_from_wheel(&wheel))?;
```

A content-coded wheel is not a parseable zip, so `extract_metadata_from_wheel` fails and the request 404s ("Metadata not available") even though the wheel is perfectly fine. This arm needs the body **decoded before parsing** — forwarding a header does nothing for it. That is a different fix from #3184's, which is the main reason this is not a one-line addition to that PR.

## Why it wasn't folded into #3184

1. It requires widening a **different** shared primitive. `fetch_upstream_direct_with_link` is also used by the OCI tag-pagination path (`proxy_helpers.rs`), so widening its 3-tuple is its own blast radius, separate from the `fetch_artifact_with_cache_path_capped` widening #3184 already carries.
2. Arm 2 is a decode-before-parse bug, not a forward-the-header bug. Bundling them would put two different remediations behind one "forward Content-Encoding" title.

## Impact

Remote PyPI repositories, behind a gzipping intermediary or an object-store upstream that returns a stored `Content-Encoding` regardless of the request's `Accept-Encoding`:

- arm 1: `pip` receives coded bytes it will not decode, and PEP 658 metadata parsing fails
- arm 2: metadata resolution 404s for a wheel that is present and valid

Not gated on `scan_on_proxy` — unlike #3184, this path is on the ordinary remote-PyPI metadata route.

## Suggested test

Same discipline as #3184: use a **non-gzip** coding (`tdh::coded_fixture("deflate", ..)`, `tdh::inflate_deflate(..)`). A gzip fixture cannot distinguish "forwards the upstream coding" from "hardcodes gzip", which is how the gap in #3176 survived fourteen tests. Cover, for each arm:

- deflate upstream ⇒ served coding matches upstream (arm 1)
- uncoded upstream ⇒ no `Content-Encoding` header (negative control)
- a deflate-coded wheel ⇒ metadata extraction still succeeds (arm 2)

## Also still open, same class

The issue body of #3184 notes a related one that is likewise unfixed: a coded-on-arrival npm tarball fails `npm_claimed_identity` (a single `GzDecoder`, so a coding layered over the `.tgz` never parses) → `ProxyScanIdentity::Unestablished` → `proxy_helpers.rs` returns `Serve { pending: true }` **before** the `tokio::spawn`, so no scan ever runs and no verdict row is written. Bounded to scan-enabled repos on the default `fail_open` action. Worth its own issue if it is not already tracked.

brandonrc:
Fixes #3193

`serve_remote_metadata` (`backend/src/api/handlers/pypi.rs`) carried two independent content-coding bugs, one per arm, needing opposite remediations. Unlike #3184 this route is **not** gated on `scan_on_proxy` — it is the ordinary remote-PyPI metadata path, so it is the wider exposure.

## Arm 1 — forward the upstream coding

The `.metadata` arm forwards the upstream body **verbatim** while pinning `text/plain; charset=utf-8` and declaring no coding. The coding was not in `fetch_upstream_direct_with_link`'s return type at all — the third slot of its triple was the `Link` header — so the handler could not forward what it never received. Since the shared HTTP client no longer lets reqwest decode upstream bodies, a coded `.metadata` from a gzipping CDN or an object-store upstream reached pip as compressed bytes labelled as plain text.

The handler now re-declares `upstream.content_encoding`; `None` upstream means the header stays **absent** rather than manufactured, which is the common case.

## Arm 2 — decode before parse

On upstream 404 the handler extracts `METADATA` from the wheel. `extract_metadata_from_wheel` opens the bytes as a zip, and a coded wheel is not a parseable zip, so a present and perfectly valid wheel answered 404 "Metadata not available". Forwarding a header does nothing here — the bytes have to be decoded before the parser sees them.

The wheel fetch moves to `fetch_artifact_with_cache_path_capped` at `DEFAULT_METADATA_MAX_BYTES` — the ceiling the uncapped wrapper already delegates with, so the byte budget is unchanged — because that is the variant #3184 widened to report the coding. Decode and parse run under **one** ingest-extraction permit, so both halves of the CPU work on upstream-controlled bytes stay admission-controlled together.

## The widening, and the caller audit

`fetch_upstream_direct_with_link` is widened **in place**, following #3194's reasoning: widening breaks every caller at compile time and forces each to declare its intent, whereas a `_with_encoding` sibling lets the next caller silently reintroduce the bug — which is how #3149 outlived #3176.

It is widened to a named `DirectUpstreamBody { content, content_type, content_encoding, link }` rather than a fourth positional `Option<String>`. That is a deliberate deviation in *form*, not reasoning: the mechanism that produced this bug was positional confusion — the old triple's third slot was `Link` while the sibling `CachedBody` triple's third slot is the coding, so `Ok((content, _ct, _))` read as "coding deliberately dropped" when it was "coding never available". A 4-tuple of three consecutive `Option<String>`s makes that failure mode more likely, not less.

Every caller, and what it declared:

| Caller | Bytes | Treatment |
|---|---|---|
| `pypi.rs::serve_remote_metadata` arm 1 | **verbatim** | forwards `content_encoding` on the response (the #3193 fix) |
| `proxy_helpers::proxy_fetch_uncached_with_link` | pass-through wrapper | returns the whole `DirectUpstreamBody`; documents that callers must forward *or* strip |
| `oci_v2.rs::fetch_upstream_tags_page` (via that wrapper) | **parses** (`serde_json::from_slice`, rebuilds the page) | **strips the coding before parsing** — see below |

The OCI tag-pagination caller had the same decode-before-parse gap: a coded `tags/list` document reached `serde_json::from_slice` still compressed and the whole tag listing 502'd. It is fixed here, with its own test, because the honest outcome of auditing it was "transforms, and cannot currently handle coded input" — leaving a known break in a caller this PR touches would be worse than the small scope increase. Happy to split it out if a reviewer prefers.

Two related primitives are documented rather than changed, matching how #3194 documented `proxy_fetch_capped_with_cache_key`:

- `ProxyService::fetch_artifact_with_cache_path` now states that it DROPS the coding and is otherwise identical to the capped form at `DEFAULT_METADATA_MAX_BYTES`, so a verbatim-forwarding or parsing caller must use the capped one.

## New: `util::content_coding`

Coding removal was open-coded nowhere and needed in two places, so it is centralised. It handles `gzip`/`x-gzip`, `deflate` (both the RFC's zlib wrapper and the raw form real servers emit — chosen by sniffing the zlib header, then falling back, with the *first* attempt's error reported so a budget breach is not masked), `zstd`, `identity`, and layered codings stripped right-to-left per RFC 9110 §8.4.

Two properties worth review attention:

- **Bounded.** The input is upstream-controlled, so every decode runs through the shared ingest decompressed-byte budget (`bounded_archive`, 128 MiB default, env-tunable). An unbounded decode here would be a remote memory-exhaustion primitive on the serve path. A `strip_content_coding_to` variant takes an explicit budget so the bomb test drives a tiny budget against a tiny fixture instead of mutating a process-global env var and racing the rest of the binary.
- **`Unsupported` is not an `Err`.** "This build has no decoder for `br`" is a different decision per caller than "the stream is corrupt or is a bomb". The PyPI wheel arm degrades an unsupported coding to its **pre-existing** "Metadata not available" answer rather than a new hard error: pip recovers from that by downloading the wheel itself, which succeeds because the download path forwards the coding for the client to strip (#3176). A 502 there would turn a recoverable case into a failed install.

## Tests

Every fixture is **`deflate`, never gzip** — a gzip fixture cannot distinguish "forwards the upstream's coding" from "hardcodes gzip", which is how the gap in #3176 survived fourteen tests. `tdh::gzip_fixture` is for GET/HEAD parity arms and is deliberately untouched. `tdh::coded_fixture`'s encoder block is factored into `tdh::code_bytes` so the wheel test can code a *specific* artifact (a real zip) rather than a generated payload.

The PyPI tests run the **real router end to end** against a mock upstream, so they fail against the handler rather than against a response-builder helper.

- `test_remote_pypi_metadata_forwards_upstream_deflate_coding` — deflate upstream ⇒ served `Content-Encoding: deflate`; body is the coded bytes verbatim; **body decodes** under the declared coding.
- `test_remote_pypi_metadata_omits_coding_when_upstream_uncoded` — negative control: uncoded upstream ⇒ **no** `Content-Encoding`.
- `test_remote_pypi_metadata_extracts_from_deflate_coded_wheel` — arm 2: a deflate-coded wheel's METADATA is extracted (200, correct bytes), and the wheel's coding is not echoed onto the freshly-built plaintext.
- `test_fetch_upstream_direct_with_link_reports_upstream_content_encoding` — the widened primitive reports the upstream coding and does **not** decode; the pre-existing `Link` test gains an uncoded negative control.
- `test_upstream_tags_page_decodes_coded_body_before_parsing` — OCI: a deflate-coded `tags/list` parses instead of 502ing.
- Eight `util::content_coding` unit tests: identity/absent pass-through, all supported tokens against *different* encoders' output, deflate in both zlib and raw forms, case/whitespace tolerance, layered right-to-left stripping, `br` reported as `Unsupported` (not silently passed through as "decoded"), corrupt stream as an error, and bomb rejection with a same-fixture control proving the rejection is the budget and not a broken fixture.

### Mutation results

Each mutation was applied to the fixed tree and the suite re-run; all five were caught.

**A — hardcode `"gzip"` in arm 1:**
```
test test_remote_pypi_metadata_forwards_upstream_deflate_coding ... FAILED
assertion `left == right` failed: the served `.metadata` must declare the UPSTREAM's coding; `gzip` means the value is hardcoded, `None` means #3193 is live
  left: Some("gzip")
 right: Some("deflate")
```

**B — emit the header unconditionally in arm 1:**
```
test test_remote_pypi_metadata_omits_coding_when_upstream_uncoded ... FAILED
assertion `left == right` failed: an uncoded upstream must be served with NO Content-Encoding; manufacturing one makes pip try to inflate plain text
  left: Some("identity")
 right: None
```

**C — revert arm 2's decode (parse the coded wheel directly):** reproduces the reported symptom exactly.
```
test test_remote_pypi_metadata_extracts_from_deflate_coded_wheel ... FAILED
assertion `left == right` failed: a deflate-coded wheel is still a valid wheel; its METADATA must be extracted rather than 404'd as unavailable; body: {"code":"NOT_FOUND","message":"Metadata not available"}
  left: 404
 right: 200
```

**D — drop the coding in the widened primitive (the pre-fix state):** fails at both layers.
```
test test_remote_pypi_metadata_forwards_upstream_deflate_coding ... FAILED
  left: None
 right: Some("deflate")

test test_fetch_upstream_direct_with_link_reports_upstream_content_encoding ... FAILED
assertion `left == right` failed: the UPSTREAM's coding must reach the caller; `gzip` here would mean it is hardcoded, `None` means #3193 is live
  left: None
 right: Some("deflate")
```

**E — revert the OCI decode:**
```
test test_upstream_tags_page_decodes_coded_body_before_parsing ... FAILED
a deflate-coded tags/list document must be decoded before parsing, not 502'd: 502
```

## Gates

- `cargo fmt --check` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean (exit 0)
- `AK_TESTS_REQUIRE_DB=1 cargo test --lib -- pypi` — 400 passed, 0 failed
- `AK_TESTS_REQUIRE_DB=1 cargo test --lib -- oci_v2` — 525 passed, 0 failed
- `AK_TESTS_REQUIRE_DB=1 cargo test --lib -- proxy_service` — 344 passed, 0 failed
- `cargo test --lib -- content_coding` — 8 passed, 0 failed
- No `sqlx::query!` macro text changed, so `.sqlx` is untouched; `SQLX_OFFLINE=true cargo build --lib --tests` with `DATABASE_URL` unset succeeds.

## Adjacent, deliberately not fixed here

`maven.rs::download_root` (both the Remote and Virtual-member arms) forwards the upstream body **verbatim** — it copies the upstream `Content-Type` and sets `Content-Length` from `content.len()` — through `fetch_artifact_with_cache_path`, which drops the coding. That is the same #3149 class on a different handler and outside this issue's scope; worth its own issue. It is why `fetch_artifact_with_cache_path` gained the explicit "DROPS the coding" doc rather than being widened.

### DIFF
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -8666,7 +8666,7 @@ async fn fetch_upstream_tags_page(
     })?;
 
     let upstream_path = format!("v2/{}/{}", ctx.image, build_remote_tags_list_path(n, last));
-    let (content, _ct, link) = proxy_helpers::proxy_fetch_uncached_with_link(
+    let upstream = proxy_helpers::proxy_fetch_uncached_with_link(
         proxy,
         ctx.repo_id,
         ctx.repo_key,
@@ -8687,6 +8687,34 @@ async fn fetch_upstream_tags_page(
         ),
     })?;
 
+    // This arm PARSES the upstream body (it rebuilds the tag page rather than
+    // forwarding bytes), so the upstream coding must come OFF here rather than
+    // be forwarded: `serde_json::from_slice` over a gzipped tags document fails
+    // and the page 502s. Reachable because the shared HTTP client no longer
+    // decodes upstream bodies and object stores return a stored
+    // `Content-Encoding` regardless of `Accept-Encoding` (#3193; same
+    // decode-before-parse class as the PyPI wheel fallback, surfaced by the
+    // `fetch_upstream_direct_with_link` widening that made the coding visible
+    // to callers at all).
+    let content = match crate::util::content_coding::strip_content_coding(
+        &upstream.content,
+        upstream.content_encoding.as_deref(),
+    ) {
+        Ok(crate::util::content_coding::Decoded::Bytes(bytes)) => bytes,
+        Ok(crate::util::content_coding::Decoded::Unsupported) | Err(_) => {
+            warn!(
+                "Upstream tags/list response for {} used a content coding this build \
+                 cannot decode ({:?})",
+                ctx.image, upstream.content_encoding
+            );
+            return Err(oci_error(
+                StatusCode::BAD_GATEWAY,
+                "UNKNOWN",
+                "invalid upstream tags response",
+            ));
+        }
+    };
+    let link = upstream.link;
     let parsed = serde_json::from_slice::<serde_json::Value>(&content).map_err(|e| {
         warn!(
             "Invalid upstream tags/list response for {}: {}",
@@ -26799,6 +26827,59 @@ mod content_encoding_forwarding_tests {
         );
         let (_s, body, _h) = tdh::collect_response(resp).await;
         assert_eq!(&body[..], &coded[..], "layer bytes must pass through");
+    }
+
+    /// #3193, the OTHER caller of the widened
+    /// `fetch_upstream_direct_with_link`. Tag pagination PARSES the upstream
+    /// body (it rebuilds the page rather than forwarding bytes), so it is the
+    /// mirror image of the blob arm above: the coding must come OFF, not be
+    /// forwarded.
+    ///
+    /// Before the widening the coding was not in the primitive's return type at
+    /// all, so a `deflate`-coded tags document reached `serde_json::from_slice`
+    /// still compressed and the whole tag listing 502'd. Fixture is `deflate`,
+    /// never gzip, for the same reason as everywhere else in this class.
+    #[tokio::test]
+    async fn test_upstream_tags_page_decodes_coded_body_before_parsing() {
+        let Some(fx) = tdh::Fixture::setup("remote", "docker").await else {
+            return;
+        };
+        let server = MockServer::start().await;
+        let page = br#"{"name":"myimage","tags":["v1","v2"]}"#;
+        let coded = tdh::code_bytes("deflate", page);
+        assert_ne!(
+            &coded[..],
+            &page[..],
+            "fixture must actually be coded or the test proves nothing"
+        );
+
+        Mock::given(method("GET"))
+            .and(wm_path("/v2/myimage/tags/list"))
+            .respond_with(
+                ResponseTemplate::new(200)
+                    .insert_header("content-encoding", "deflate")
+                    .set_body_bytes(coded),
+            )
+            .mount(&server)
+            .await;
+
+        let upstream_url = server.uri();
+        let (state, _cache) = tdh::rewire_remote_proxy(&fx, &upstream_url).await;
+        let ctx = super::TagsFetchCtx {
+            state: &state,
+            repo_id: fx.repo_id,
+            repo_key: &fx.repo_key,
+            upstream_url: &upstream_url,
+            image: "myimage",
+        };
+        let result = super::fetch_upstream_tags_page(&ctx, 10, None).await;
+
+        fx.teardown().await;
+
+        let page = result
+            .map_err(|r| r.status())
+            .expect("a deflate-coded tags/list document must be decoded before parsing, not 502'd");
+        assert_eq!(page.tags, vec!["v1".to_string(), "v2".to_string()]);
     }
 
     /// Negative control: no upstream coding, no manufactured header, and a
--- a/backend/src/api/handlers/proxy_helpers.rs
+++ b/backend/src/api/handlers/proxy_helpers.rs
@@ -17,8 +17,8 @@ use crate::models::repository::{
     ReplicationPriority, Repository, RepositoryFormat, RepositoryType,
 };
 use crate::services::proxy_hydration::{Coordinator, HydrationCoordinator};
-use crate::services::proxy_service::ProxyService;
 pub use crate::services::proxy_service::StreamingFetchResult;
+use crate::services::proxy_service::{DirectUpstreamBody, ProxyService};
 // Re-export the per-format buffered-metadata byte ceilings (#1608 Phase 4b /
 // #2181) so format handlers select a cap via `proxy_helpers::<CONST>`.
 pub use crate::services::proxy_service::{DEFAULT_METADATA_MAX_BYTES, LARGE_METADATA_MAX_BYTES};
@@ -1719,13 +1719,18 @@ pub async fn proxy_fetch_uncached(
 }
 
 /// Fetch from upstream directly, preserving the upstream `Link` header.
+///
+/// Returns the whole [`DirectUpstreamBody`], including the upstream
+/// `Content-Encoding` (#3193). Callers must handle the coding explicitly:
+/// forward it if they pass `content` through verbatim, or strip it with
+/// [`crate::util::content_coding::strip_content_coding`] before parsing.
 pub async fn proxy_fetch_uncached_with_link(
     proxy_service: &ProxyService,
     repo_id: Uuid,
     repo_key: &str,
     upstream_url: &str,
     path: &str,
-) -> Result<(Bytes, Option<String>, Option<String>), Response> {
+) -> Result<DirectUpstreamBody, Response> {
     with_proxy_repo(repo_id, repo_key, upstream_url, path, |repo| async move {
         proxy_service
             .fetch_upstream_direct_with_link(&repo, path)
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -2023,17 +2023,30 @@ async fn serve_remote_metadata(
         .fetch_upstream_direct_with_link(&remote_repo, &target.fetch_path)
         .await
     {
-        Ok((content, _content_type, _)) => {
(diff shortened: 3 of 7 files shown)


brandonrc:
On your open question — **keep the OCI tags fix in this PR.**

The reasoning is the same one that justified widening in place: the signature change breaks every caller at compile time precisely so each declares whether it forwards bytes verbatim or transforms them. `fetch_upstream_tags_page` transforms — it hands the bytes to `serde_json::from_slice`. Splitting would land the widened signature alongside a caller that provably cannot handle coded input, which is a coded `tags/list` 502ing the whole listing. That is a worse intermediate state than a slightly broader PR, and 'the next caller silently reintroduces it' is exactly how #3149 outlived #3176.

The `DirectUpstreamBody` deviation from a 4-tuple is right too. The bug's mechanism *was* positional confusion — the old triple's third slot was `Link` while the sibling `CachedBody` triple's third slot is the coding, so `Ok((content, _ct, _))` read as a deliberate drop. Three consecutive `Option<String>` would make that likelier, not less. A named struct removes the failure mode rather than relocating it.

Two other calls I'd endorse explicitly, since both are the kind of thing a reviewer might 'simplify' later:

`Unsupported` deliberately not being an `Err` is correct. Degrading an unstrippable `br` to the pre-existing "Metadata not available" lets pip recover by fetching the wheel itself; turning it into a hard 502 would be a new failure mode introduced by a fix. Worth keeping the comment that says so.

Bounding the decode against the shared ingest budget is not optional. The input is upstream-controlled, so an unbounded decode here is a remote memory-exhaustion primitive — a decompression bomb served by a compromised or hostile mirror. Sharing one permit across decode and zip-walk is the right shape.

I've filed the `maven.rs::download_root` finding as its own issue rather than widening this further.

PR #3223, opened 2026-08-09
fix(nuget): repair the flatcontainer_download test call site broken on main

### LINKED ISSUE #3221: main test build is broken: nuget flatcontainer_download call site missed by #3205
`cargo build --lib --tests` fails on a pristine `origin/main` (`1743098`), so the whole test suite is uncompilable and `cargo test` / nextest cannot run at all.

```
error[E0277]: the trait bound `Path<(String, String, String, String)>: Default` is not satisfied
    --> backend/src/api/handlers/nuget.rs:4659:17
4659 |                 Default::default(),
     |                 ^^^^^^^^^^^^^^^^^^

error[E0061]: this function takes 4 arguments but 3 arguments were supplied
    --> backend/src/api/handlers/nuget.rs:4651:24
4651 |               let resp = super::flatcontainer_download(
     | argument #4 of type `download_telemetry::DownloadContext` is missing

error: could not compile `artifact-keeper-backend` (lib test) due to 2 previous errors
```

## Cause: parallel-merge signature drift

- #3205 (`5760936`) added `Extension(auth): Extension<Option<AuthExtension>>` to `flatcontainer_download` and updated four of its five test call sites (`nuget.rs:3305`, `:4498`, `:4837`, `:4982`).
- #2919 (`a6acc9f`) concurrently **added** a fifth call site at `nuget.rs:4651` using the old 3-argument signature.

Both PRs were green independently; the conflict is semantic, not textual, so git merged them cleanly and main broke.

## Fix

One line — add the missing argument at `nuget.rs:4652`, matching the other four call sites:

```rust
axum::Extension(tdh::admin_auth_ext()),
```

This is already included in PR #3217 and PR #3218 (both needed it to run their own tests), so merging either one repairs main. Filing separately because main being red is worth its own record, and because it recurs: this is the same class as the previously-seen handler-vs-test signature drift from parallel green PRs.

## Suggested follow-up

CI appears not to be gating on a test-target **compile** for merges to main, or these would have been caught. Worth confirming that the required checks include something that compiles `--all-targets`, and considering a post-merge tripwire build after a burst of parallel merges touching shared handler signatures.

brandonrc:
fix(nuget): repair the flatcontainer_download test call site broken on main

main does not compile. `cargo build --lib --tests` fails at
backend/src/api/handlers/nuget.rs:4651 with E0061/E0277 — the call passes
State, Path, Default and omits the Extension, while the function takes
State, Extension, Path, ctx.

Classic parallel-merge signature drift: #3205 added the Extension parameter and
updated the call sites it could see, while a concurrently-merged PR added
another one. Both were individually green; the combination is not. The lib
still builds, so only `--all-targets` consumers fail — which is Check Rust,
Backend Unit Tests and Code Coverage, i.e. every required check on every open
PR.

Fixes #3221.

### DIFF
--- a/backend/src/api/handlers/nuget.rs
+++ b/backend/src/api/handlers/nuget.rs
@@ -4650,6 +4650,7 @@ mod read_db_tests {
         for _ in 0..calls {
             let resp = super::flatcontainer_download(
                 axum::extract::State(state.clone()),
+                axum::extract::Extension(None),
                 axum::extract::Path((
                     fx.repo_key.clone(),
                     package_id.to_string(),


brandonrc:
Closing as a duplicate of #3215, which is identical (+1/-0, same call site) and was already CLEAN. Two workers found main red independently and both filed a fix — merging #3215.

Leaving the note here for the record since it is the same root cause either way: #3205 added the `Extension` parameter to `flatcontainer_download` and updated the call sites visible on its base, while a concurrently-merged PR added another. Both were individually green; the combination was not. The lib still built, so only `--all-targets` consumers broke — Check Rust, Backend Unit Tests and Code Coverage, i.e. every required check on every open PR, failing for a reason unrelated to those PRs.

Issue #3180, opened 2026-08-07
OCI upload to S3 hangs at 100% and then fails


skroll:
### Description

While trying to push to an OCI/Docker repository, which is backed by S3, a layer that is 2.33GB reaches 100%, but then hangs, after a minute or so, an error is returned from the docker client:

```
Failed to copy 'oci-uploads/ed953571-5ea3-4f7f-ac63-e6f9ffe303c5.complete.8d83ca92-2d3d-4f3e-8f93-20969069f704' to 'oci-blobs/sha256:1b31e0aa4b34a899fd9a210679c80ca2f96891e19495cb39cb900734751c317a':
Generic S3 error: Error performing PUT https://s3.us-east-2.amazonaws.com/REDACTED/oci-blobs/sha256%3A1b31e0aa4b34a899fd9a210679c80ca2f96891e19495cb39cb900734751c317a in 30.00131275s - HTTP error: request or response body error
```

In the ak backend logs, I see the following (linebreaks added for clarity):
```
http_request{method=PUT uri=/v2/docker-sandbox/mingc/android-build-box/blobs/uploads/3b9054cb-848a-4b35-b2ea-60b00af4f17a?digest=sha256%3A1b31e0aa4b34a899fd9a210679c80ca2f96891e19495cb39cb900734751c317a correlation_id=33c64341-a810-4172-bd4a-78ca92830545
}:
put_stream{
  key="oci-uploads/3b9054cb-848a-4b35-b2ea-60b00af4f17a.complete.b899758f-1aac-40b0-8d78-f68d0caacf13"
  otel.kind="client"
  storage.system="s3"
  storage.operation="put_stream"
}:
object_store::client::get: Encountered error while reading response body: HTTP error: request or response body error. Retrying in 0.1s
```

Every other layer pushes just fine.

### Environment Information

- Deployment type: Rootless Podman in AWS
- Operating System: Amazon Linux 2023
- Architecture: aarch64, c6gn.large AWS instance, but also noticed same issue on amd64 t3.medium
- Database: Postgres 16
- Reverse Proxy: Caddy
- Network setup: frontend and backend ak containers exposed port, caddy running natively

### Backend Version

1.7.1

### Web UI Version

1.8.0

### Package Format

Docker / OCI

### Expected Behavior

Expected the upload to complete.

### Observed Behavior

Error is returned after hanging.

### Steps to Reproduce

`docker pull mingc/android-build-box:1.24.0`
[...]

PR #3237, opened 2026-08-10
fix(s3): copy objects >5GiB server-side via UploadPartCopy instead of restreaming through the app

### LINKED ISSUE #3164: S3 copy() for objects >5GiB restreams through the app before verifying digest
## Summary

`S3Backend::copy` (backend/src/storage/s3.rs) uses native `CopyObject` for objects up to the single-copy threshold (5 GiB), which is atomic and never touches application memory. Above that threshold, it falls back to `get_stream(source)` + `put_stream(dest)` — reading the whole object back through the app and re-uploading it to `dest`, checking the digest only after that write completes (`ensure_copy_digest_matches`).

## Why it matters

Surfaced during review of #3153 (`CachePersister::tee_stream`'s publish-via-staging-key fix). That fix relies on `copy()` being safe to call against a live, already-populated key: publish a validated staging object onto `cache_key` via `copy()`, and if anything goes wrong, the live key should never end up holding bad bytes.

For objects ≤5 GiB this holds: native `CopyObject` is atomic, so a failure never touches `dest`. For objects >5 GiB, the fallback restreams onto `dest` directly and only checks whether the result matches *afterward*. If a mismatch occurs (a corruption during the restream itself, not the original write — the source was already validated before `copy()` was called), the bad bytes are already live at `dest` by the time the check runs, and #3153 deliberately does not add a reactive delete-on-mismatch (that delete could clobber a concurrent writer's just-published good object, recreating the exact live-key race #3153 exists to remove).

## Suggested fix

Use S3's server-side `UploadPartCopy` for the >5GiB case instead of restreaming through the app — parts are copied entirely server-side and the object never passes through application memory, closing this gap the same way native `CopyObject` already does for smaller objects. The `object_store` crate this backend is built on doesn't currently expose `UploadPartCopy`; this may need either a crate feature request/PR upstream, or a hand-rolled multipart-copy implementation directly against the S3 API for this one path.

## Scope

Narrow in practice: requires both an object over 5 GiB going through the proxy-cache streaming path AND a corruption during `copy()`'s own internal restream (not the original upstream write, which is independently validated before `copy()` is ever called). Not urgent, but worth tracking rather than leaving implicit.

brandonrc:
Fixes #3164

## What was actually wrong (verified against source before fixing)

Both halves of the issue are real, confirmed on current `main` (`929c0bb`, on top of #3222's client split):

1. **Restream**: `S3Backend::copy` for objects over `S3_MAX_SINGLE_COPY_SIZE` (5 GiB) fell back to `get_stream(source)` + `put_stream(dest)` (`backend/src/storage/s3.rs:1753-1792` at merge-base) — every byte of a >5 GiB object came back through the application and was re-uploaded.
2. **Integrity ordering (the part that matters)**: `ensure_copy_digest_matches` ran only **after** `put_stream` returned — i.e. after `CompleteMultipartUpload` had already published `dest`. On a mismatch the corrupt bytes were already live, and #3153 deliberately rules out a reactive delete (it could clobber a concurrent writer's just-published good object). Worse, the check compared the tap hasher against `put_stream`'s hasher — **two SHA-256s of the same in-process byte stream** — so it was structurally unable to detect a transfer that diverged from the source at all. The check was theater; this PR fixes the ordering problem by removing the window entirely rather than reordering the theater.

## Fix

Copy `>5 GiB` objects **entirely server-side** via ranged `UploadPartCopy`, per the issue's suggested direction. AWS caps single copies at 5 GiB:

> "You create a copy of your object up to 5 GB in size in a single atomic action using this API. However, to copy an object greater than 5 GB, you must use the multipart upload Upload Part - Copy (UploadPartCopy) API." — *Amazon S3 API Reference, CopyObject*

- One `UploadPartCopy` per 5 GiB slice (`x-amz-copy-source` + `x-amz-copy-source-range`, "the form bytes=first-last, where the first and last are the zero-based byte offsets to copy" — *S3 API Reference, UploadPartCopy*). Maximal slices keep the count minimal: S3's 5 TiB object ceiling needs 1,024 parts, far under the 10,000-part cap.
- **No payload byte passes through the application**, and `dest` only becomes visible at `CompleteMultipartUpload`, after every part copied successfully (S3 validates the supplied part ETags at completion) — the same publish-is-the-last-step atomicity native `CopyObject` gives the ≤5 GiB path, which is exactly the property #3153's staging-key publish relies on.
- Failures abort the multipart upload (inline on error paths, `MultipartAbortGuard` on future drop), leaving an existing `dest` untouched.

`object_store` 0.13 implements `UploadPartCopy` internally only as the un-ranged whole-object variant behind `copy_if_not_exists`, so the ranged call is hand-rolled against the S3 REST API — but signed with the crate's **own public SigV4 signer** (`AwsAuthorizer`) and the store's own credential chain (`AmazonS3::credentials()`), so static keys, IRSA, ECS/EKS container credentials and session tokens all keep working ("All headers with the `x-amz-` prefix, including `x-amz-copy-source`, must be signed" — *S3 API Reference, UploadPartCopy*; a presigned URL cannot satisfy that, header signing does). The request URL is derived from the configured store via `Signer::signed_url` rather than rebuilt, so path-style/virtual-hosted/custom-endpoint shapes carry over. The raw client honors `S3_CA_CERT_PATH` / `S3_INSECURE_TLS`. `partNumber` is 1-based per the API reference; a `200 OK` body is only accepted with a parseable `<CopyPartResult><ETag>` (S3 can embed an `<Error>` document in a 200 for copy operations).

`ensure_copy_digest_matches` and its tests are removed with the restream they guarded (see the tautology note above — there is nothing app-side left to hash, by design).

## Tests (in-file `#[cfg(test)]`, count toward the `--lib` coverage gate)

All wiremock storage doubles; large sizes are `Content-Length` headers only — **no multi-GB buffers are materialized**.

- `test_copy_over_5gib_is_server_side_upload_part_copy` — 10 GiB + 1 B copy: asserts **zero GET requests** (no restream), **exactly three bodyless ranged part copies** whose `(partNumber, x-amz-copy-source, x-amz-copy-source-range, body length)` tuples exactly partition `[0, size-1]`, and exactly one `CompleteMultipartUpload`. No GET mock is mounted, so the old shape fails on the first source read.
- `test_copy_over_5gib_failed_part_copy_never_publishes_dest` — the integrity claim. Second part copy 500s; asserts the copy errors, `CompleteMultipartUpload` count == **0** (dest never published), abort count == 1. The fixture deliberately keeps the *old* shape fully green (source GET + un-ranged part PUTs + complete all mocked OK), so pre-fix code returns `Ok` and publishes — the test cannot pass unless the publish is actually gated on the part copies.
- `test_copy_at_or_under_threshold_stays_single_native_copy_object` — positive control at exactly 5 GiB: exactly one native `CopyObject` PUT (`x-amz-copy-source`, no range header, no `uploadId`), zero GET/POST — a fix that degraded or multipart-ified the common path fails these counts.

The prior `test_copy_streams_large_source_instead_of_single_copy_object` asserted the bug as a requirement ("large copy should read the source as a stream", "must not use \[x-amz-copy-source\]") and is replaced by the above.

## Revert-proof (faithful, insertions-only)

Production `s3.rs` restored byte-identical to `main`, keeping only the new tests:

```
$ git diff main --shortstat
 1 file changed, 294 insertions(+)
$ git diff main -U0 | grep ^@@
@@ -4168,0 +4169,294 @@ mod tests {
```

(single hunk, entirely inside `mod tests`). Result on merge-base production code:

```
PASS test_copy_at_or_under_threshold_stays_single_native_copy_object   (control)
FAIL test_copy_over_5gib_is_server_side_upload_part_copy
     -> NotFound("Storage key not found: source-object")  [old code restreams the source]
FAIL test_copy_over_5gib_failed_part_copy_never_publishes_dest
     -> "#3164: a failed server-side part copy must fail the whole copy
         (the old restream shape returned Ok here)"       [old code published dest]
```

With the fix restored: all 424 `storage::` `--lib` tests pass; `cargo fmt` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean. No `sqlx::query!` text touched.

## Live validation against real MinIO (SigV4 proof)

The wiremock suite asserts request *shape* but cannot validate signatures, so the multipart-copy path is also proven against a live server. To make that testable without moving 5 GiB, the single-copy ceiling is now configurable — `S3Config::max_single_copy_bytes` + `with_max_single_copy_bytes()` + `S3_MAX_SINGLE_COPY_BYTES` env (the `bulk_timeout_secs` shape from #3222), defaulting to AWS's 5 GiB cap and clamped into [5 MiB, 5 GiB] (a range copy requires a source over 5 MB, and no single copy may exceed 5 GiB). The multipart part slice follows the same knob, since each `UploadPartCopy` is itself a single copy operation under the same ceiling. Defensible on its own: it is the operator lever for S3-compatible stores with lower copy limits.

`test_large_copy_upload_part_copy_live_3164` (`AK_S3_E2E`-gated + `#[ignore]`, same pattern as #3180's roundtrip test) sets a 6 MiB ceiling via the env lever, copies a 16 MiB object of position-dependent bytes against live MinIO, and asserts: (1) the copy is **accepted** — real SigV4 validation of the hand-rolled requests; (2) destination bytes equal source bytes — a range mix-up would change the content; (3) the dest ETag carries the `-3` multipart suffix, so the `UploadPartCopy` path demonstrably ran rather than a native-copy fallback.

**The live run immediately caught a real bug the doubles could not:** the part-copy requests were signed correctly and accepted, but `CompleteMultipartUpload` failed with `InvalidPart` — MinIO (Go `encoding/xml`) escapes the ETag quotes as decimal entities (`&#34;`) where AWS uses `&quot;`, and the parser passed the entity through literally. `parse_copy_part_etag` now decodes both; the two wiremock fixtures use one encoding each. Fixed output:

```
#3164 live multipart copy OK: len=16777216 dest_etag="c59bad0d7b9c262935d141050bf39bf9-3"
test result: ok. 1 passed; 0 failed
```

MinIO's server-side API trace (`mc admin trace --call s3`) for the same run shows the exact call sequence, confirming the hand-rolled requests were accepted and executed server-side — each `CopyObjectPart` (MinIO's name for `UploadPartCopy`) uploads ~131 B of headers and no payload:

```
[200 OK] s3.NewMultipartUpload   .../large-copy-3164/dst-...?uploads=
[200 OK] s3.CopyObjectPart       .../large-copy-3164/dst-...?partNumber=1&uploadId=...  ↑ 131 B ↓ 228 B
[200 OK] s3.CopyObjectPart       .../large-copy-3164/dst-...?partNumber=2&uploadId=...  ↑ 131 B ↓ 228 B
[200 OK] s3.CopyObjectPart       .../large-copy-3164/dst-...?partNumber=3&uploadId=...  ↑ 131 B ↓ 228 B
[200 OK] s3.CompleteMultipartUpload .../large-copy-3164/dst-...?uploadId=...
```

**Note for reviewers:** this live test is `AK_S3_E2E`-gated and `#[ignore]`d, so it does **not** run in CI — it is a deliberate-invocation check, the same pattern as `test_oci_blob_roundtrip_live_3180`. A green CI run does not constitute signature validation; the validation is the recorded run above (executed 2026-08-09/10 against MinIO at `http://127.0.0.1:39164`, path-style, static `S3_ACCESS_KEY_ID`/`S3_SECRET_ACCESS_KEY` credentials).

Revert-proof was re-run for the full test set (production restored byte-identical to main, now 387 insertions / 0 deletions across the two test modules — `@@ mod tests` and `@@ mod integration_tests`): both wiremock #3164 tests fail as before, and the live test fails on assertion (3) with `dest ETag "\"bccf03b2...\"" does not carry the 3-part multipart suffix` — on merge-base the env lever is ignored, the 16 MiB copy silently rides native `CopyObject`, and copies (1) and (2) alone cannot distinguish the shapes. The control still passes.

## Behavior notes

- ≤5 GiB copies are byte-for-byte unchanged (native `CopyObject` on the bulk client, #3222 timeouts intact).
- Part copies run sequentially; each rides the bulk timeout (`S3_BULK_TIMEOUT_SECS`) since a 5 GiB server-side copy is bulk work. Parallelizing part copies is possible later but irrelevant to correctness.
- S3-compatible stores that lack `UploadPartCopy` would now fail the >5 GiB copy loudly instead of silently restreaming; every mainstream target (AWS, MinIO, Ceph RGW, R2, OBS) implements it.

### DIFF
--- a/backend/src/storage/s3.rs
+++ b/backend/src/storage/s3.rs
@@ -27,6 +27,10 @@
 //!   CopyObject (default: 1800). Control-plane calls (head/list/delete/exists)
 //!   keep a short [`S3_CONTROL_TIMEOUT`] so a wedged endpoint still fails fast.
 //!   Set to 0 to disable the bulk timeout entirely.
+//! - S3_MAX_SINGLE_COPY_BYTES: Largest object copied with one server-side
+//!   CopyObject (default: 5 GiB, AWS's cap). Larger sources are copied via
+//!   multipart UploadPartCopy in slices of this size. Clamped into
+//!   [5 MiB, 5 GiB]. Lever for S3-compatible stores with lower copy limits.
 //!
 //! For redirect downloads (302 to presigned URLs):
 //! - S3_REDIRECT_DOWNLOADS: Enable 302 redirects (default: false)
@@ -67,6 +71,13 @@ const S3_MULTIPART_CHUNK_SIZE: usize = 5 * 1024 * 1024;
 /// this regardless of how large the object is.
 const S3_MULTIPART_MAX_PART_SIZE: u64 = 5 * 1024 * 1024 * 1024;
 const S3_MULTIPART_MAX_IN_FLIGHT_PARTS: usize = 4;
+/// AWS's ceiling for a single copy operation (5 GiB): "You create a copy of
+/// your object up to 5 GB in size in a single atomic action using this API.
+/// However, to copy an object greater than 5 GB, you must use the multipart
+/// upload Upload Part - Copy (UploadPartCopy) API" (Amazon S3 API Reference,
+/// CopyObject). Default for the configurable
+/// [`S3Config::max_single_copy_bytes`]; also the hard upper clamp, since no
+/// single copy — `CopyObject` or one `UploadPartCopy` part — may exceed it.
 const S3_MAX_SINGLE_COPY_SIZE: u64 = 5 * 1024 * 1024 * 1024;
 /// Overall HTTP request timeout for control-plane S3 calls: head, exists, list,
 /// delete, multipart create/abort. These are all small, fixed-cost round trips,
@@ -367,17 +378,19 @@ impl Drop for MultipartAbortGuard {
 /// Each range becomes one `UploadPartCopy`'s `x-amz-copy-source-range` header,
 /// which the S3 API Reference (UploadPartCopy, `x-amz-copy-source-range`)
 /// defines as "the form bytes=first-last, where the first and last are the
-/// zero-based byte offsets to copy". Every part is
-/// [`S3_MULTIPART_MAX_PART_SIZE`] (5 GiB, S3's per-part ceiling) except a
-/// shorter final part; maximal parts keep the count minimal — S3's 5 TiB
-/// object ceiling needs 1,024 parts, far under [`S3_MULTIPART_MAX_PARTS`].
-/// Every non-final part is 5 GiB, so the 5 MiB minimum-part-size rule is
-/// trivially satisfied for the `> 5 GiB` objects this path handles.
-fn copy_part_ranges(size: u64) -> Vec<(u64, u64)> {
+/// zero-based byte offsets to copy". Every part is `part_size` except a
+/// shorter final part. `part_size` is the (clamped) single-copy ceiling — each
+/// `UploadPartCopy` is itself a single copy operation bound by the same limit
+/// — so maximal parts keep the count minimal: at the default 5 GiB ceiling,
+/// S3's 5 TiB object ceiling needs 1,024 parts, far under
+/// [`S3_MULTIPART_MAX_PARTS`]. Callers guarantee `part_size >= 5 MiB`
+/// (see [`S3Backend::effective_single_copy_ceiling`]), satisfying the
+/// minimum-part-size rule for every non-final part.
+fn copy_part_ranges(size: u64, part_size: u64) -> Vec<(u64, u64)> {
     let mut ranges = Vec::new();
     let mut start = 0u64;
     while start < size {
-        let end = (start + S3_MULTIPART_MAX_PART_SIZE).min(size) - 1;
+        let end = (start + part_size).min(size) - 1;
         ranges.push((start, end));
         start = end + 1;
     }
@@ -401,8 +414,14 @@ fn s3_copy_source_value(bucket: &str, full_key: &str) -> String {
 /// A successful response is `<CopyPartResult><ETag>...</ETag></CopyPartResult>`
 /// (S3 API Reference, UploadPartCopy response syntax). S3 copy operations can
 /// return `200 OK` with an embedded `<Error>` document instead, so a 200
-/// status alone is not success — only a parseable ETag is. The ETag's XML
-/// entity escapes are undone (S3 wraps ETags in `&quot;`).
+/// status alone is not success — only a parseable ETag is.
+///
+/// The ETag's XML entity escapes are undone: implementations differ on how
+/// they escape the quotes around the ETag — AWS emits the named entity
+/// (`&quot;`), MinIO (Go's `encoding/xml`) the decimal reference (`&#34;`) —
+/// and a passed-through entity poisons `CompleteMultipartUpload` with
+/// `InvalidPart` ("the specified entity tag may not match"), caught live
+/// against MinIO by `test_large_copy_upload_part_copy_live_3164`.
 fn parse_copy_part_etag(body: &str) -> Option<String> {
     if body.contains("<Error") {
         return None;
@@ -412,6 +431,10 @@ fn parse_copy_part_etag(body: &str) -> Option<String> {
     let etag = body[start..end]
         .trim()
         .replace("&quot;", "\"")
+        .replace("&#34;", "\"")
+        .replace("&#x22;", "\"")
+        .replace("&apos;", "'")
+        .replace("&#39;", "'")
         .replace("&lt;", "<")
         .replace("&gt;", ">")
         .replace("&amp;", "&");
@@ -468,6 +491,15 @@ pub struct S3Config {
     /// so this stays short to fail fast on a wedged endpoint. `0` disables it.
     /// Default: [`S3_CONTROL_TIMEOUT`] (30s).
     pub control_timeout_secs: u64,
+    /// Largest object copied with a single server-side `CopyObject`; larger
+    /// sources use the multipart `UploadPartCopy` path, whose per-part slice
+    /// size is this same value (each part copy is a single copy operation
+    /// under the same ceiling). Operator lever for S3-compatible stores with
+    /// a lower single-copy limit than AWS's. Clamped at use into
+    /// [5 MiB, 5 GiB]: a range copy requires a source over 5 MB, and no
+    /// single copy may exceed AWS's 5 GiB cap. Default:
+    /// [`S3_MAX_SINGLE_COPY_SIZE`] (5 GiB).
+    pub max_single_copy_bytes: u64,
 }
 
 /// CloudFront CDN configuration for signed URLs
@@ -532,6 +564,10 @@ impl S3Config {
             .ok()
             .and_then(|v| v.parse().ok())
             .unwrap_or(S3_CONTROL_TIMEOUT.as_secs());
+        let max_single_copy_bytes: u64 = std::env::var("S3_MAX_SINGLE_COPY_BYTES")
+            .ok()
+            .and_then(|v| v.parse().ok())
+            .unwrap_or(S3_MAX_SINGLE_COPY_SIZE);
 
         Ok(Self {
             bucket,
@@ -551,6 +587,7 @@ impl S3Config {
             pool_idle_timeout_secs,
             bulk_timeout_secs,
             control_timeout_secs,
+            max_single_copy_bytes,
         })
     }
 
@@ -616,6 +653,7 @@ impl S3Config {
             pool_idle_timeout_secs: 90,
             bulk_timeout_secs: S3_DEFAULT_BULK_TIMEOUT_SECS,
             control_timeout_secs: S3_CONTROL_TIMEOUT.as_secs(),
+            max_single_copy_bytes: S3_MAX_SINGLE_COPY_SIZE,
         }
     }
 
@@ -677,6 +715,13 @@ impl S3Config {
     /// Override the control-plane request timeout (seconds). `0` disables it.
     pub fn with_control_timeout_secs(mut self, timeout_secs: u64) -> Self {
         self.control_timeout_secs = timeout_secs;
+        self
+    }
+
+    /// Override the single-copy ceiling in bytes (see
+    /// [`Self::max_single_copy_bytes`]). Clamped at use into [5 MiB, 5 GiB].
+    pub fn with_max_single_copy_bytes(mut self, bytes: u64) -> Self {
+        self.max_single_copy_bytes = bytes;
         self
     }
 }
@@ -834,6 +879,11 @@ pub struct S3Backend {
     /// does not expose (ranged `UploadPartCopy`, #3164). Built with the same
     /// custom-CA / insecure-TLS options as the object_store clients.
     raw_http: reqwest::Client,
+    /// Single-copy ceiling (already clamped into [5 MiB, 5 GiB], see
+    /// [`Self::effective_single_copy_ceiling`]): sources at or under it use
+    /// native `CopyObject`; larger ones use multipart `UploadPartCopy` with
+    /// parts of this size.
+    max_single_copy_bytes: u64,
     prefix: Option<String>,
     redirect_downloads: bool,
     cloudfront: Option<CloudFrontConfig>,
@@ -873,6 +923,19 @@ impl S3Backend {
         let timeout =
             (config.bulk_timeout_secs > 0).then(|| Duration::from_secs(config.bulk_timeout_secs));
         Self::build_store_with_timeout(config, access_key, secret_key, timeout)
+    }
+
+    /// Clamp the configured single-copy ceiling into what the S3 API can
+    /// honor: at least 5 MiB ("You can copy a range only if the source object
+    /// is greater than 5 MB" — S3 API Reference, UploadPartCopy — and 5 MiB
+    /// is the minimum non-final part size) and at most
+    /// [`S3_MAX_SINGLE_COPY_SIZE`] (no single copy may exceed AWS's 5 GiB
+    /// cap). Used for both the CopyObject-vs-multipart threshold and the
+    /// multipart part slice size.
+    fn effective_single_copy_ceiling(config: &S3Config) -> u64 {
+        config
+            .max_single_copy_bytes
+            .clamp(S3_MULTIPART_CHUNK_SIZE as u64, S3_MAX_SINGLE_COPY_SIZE)
     }
 
     /// HTTP client for hand-rolled S3 API calls that `object_store` does not
@@ -1147,6 +1210,7 @@ impl S3Backend {
             bulk_timeout: (config.bulk_timeout_secs > 0)
                 .then(|| Duration::from_secs(config.bulk_timeout_secs)),
             raw_http,
+            max_single_copy_bytes: Self::effective_single_copy_ceiling(&config),
             prefix: config.prefix,
             redirect_downloads: config.redirect_downloads,
             cloudfront: config.cloudfront,
@@ -1829,7 +1893,7 @@ impl S3Backend {
     /// Copy content from one key to another
     pub async fn copy(&self, source: &str, dest: &str) -> Result<()> {
         let size = self.size(source).await?;
-        if size > S3_MAX_SINGLE_COPY_SIZE {
+        if size > self.max_single_copy_bytes {
             tracing::debug!(
                 source = %source,
                 dest = %dest,
@@ -1853,8 +1917,10 @@ impl S3Backend {
         Ok(())
     }
 
-    /// Copy an object larger than [`S3_MAX_SINGLE_COPY_SIZE`] entirely
-    /// server-side via multipart `UploadPartCopy` (#3164).
+    /// Copy an object larger than the single-copy ceiling
+    /// ([`S3Config::max_single_copy_bytes`], default
+    /// [`S3_MAX_SINGLE_COPY_SIZE`]) entirely server-side via multipart
+    /// `UploadPartCopy` (#3164).
     ///
     /// AWS caps single copies at 5 GiB: "You create a copy of your object up
     /// to 5 GB in size in a single atomic action using this API. However, to
@@ -1876,7 +1942,7 @@ impl S3Backend {
     /// earlier failure aborts the upload and leaves an existing `dest`
     /// object untouched.
     async fn multipart_server_side_copy(&self, source: &str, dest: &str, size: u64) -> Result<()> {
-        let ranges = copy_part_ranges(size);
+        let ranges = copy_part_ranges(size, self.max_single_copy_bytes);
         if ranges.len() > S3_MULTIPART_MAX_PARTS {
             // Unreachable for real S3 (5 TiB object ceiling / 5 GiB parts =
             // 1,024), but fail fast rather than opaquely at part 10,001 if an
@@ -3703,6 +3769,7 @@ mod tests {
             region: config.region.clone(),
             bulk_timeout: Some(Duration::from_secs(S3_DEFAULT_BULK_TIMEOUT_SECS)),
             raw_http: reqwest::Client::new(),
+            max_single_copy_bytes: S3_MAX_SINGLE_COPY_SIZE,
             prefix: None,
             redirect_downloads: false,
             cloudfront: None,
@@ -3980,6 +4047,7 @@ mod tests {
                 bulk_timeout: (config.bulk_timeout_secs > 0)
                     .then(|| Duration::from_secs(config.bulk_timeout_secs)),
                 raw_http: reqwest::Client::new(),
+                max_single_copy_bytes: S3_MAX_SINGLE_COPY_SIZE,
                 prefix: None,
                 redirect_downloads: false,
                 cloudfront: None,
@@ -4058,6 +4126,7 @@ mod tests {
             bulk_timeout: (config.bulk_timeout_secs > 0)
                 .then(|| Duration::from_secs(config.bulk_timeout_secs)),
             raw_http: reqwest::Client::new(),
+            max_single_copy_bytes: S3_MAX_SINGLE_COPY_SIZE,
             prefix: None,
             redirect_downloads: false,
             cloudfront: None,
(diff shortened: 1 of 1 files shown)


brandonrc:
## Independent validation review — head `c121c57`

Reviewed from a fresh clone (`git fetch origin pull/3237/head`), diffed against merge-base `929c0bb` (one file, `backend/src/storage/s3.rs`, +732/-103). I set out to refute the PR, and mostly failed to.

**Verdict: no blockers. The copy path is correct — I could not break the range arithmetic, the SigV4 signing, the ETag parsing, the abort handling, or the publish gating, and I tried each against a live server rather than against the mocks.** Two defects and two robustness gaps are described below; none of them can corrupt data or publish a bad object. I'd like **D1** fixed in this PR (~10 lines) and **D2/D3** either folded in or tracked as follow-ups.

Credit where it's due: **the `&#34;` vs `&quot;` ETag-escaping find is the real thing, and it is exactly the class of bug that only a live run can produce.** I reproduced it independently (see below) — MinIO emits the decimal character reference, and feeding it through unmodified gets `CompleteMultipartUpload` rejected with `InvalidPart`. A reviewer working from the wiremock suite alone would never have seen it. The decision to hand-roll the request rather than fake it, and then go validate the signature against a real server, is the right call for this change.

### What I did to test it

Because signing bugs pass mocks, I built a standalone binary that replicates `upload_part_copy` / `multipart_server_side_copy` call-for-call against the same `object_store` 0.13.2 (`signed_url` → `set_query(None)` → `query_pairs_mut` → `credentials().get_credential()` → `AwsAuthorizer::new(..).authorize(&mut req, None)` → `reqwest.put().headers(..)`) and ran it against a live MinIO. Results:

- **Signature accepted.** All three ranged part copies returned `200 OK`, `CompleteMultipartUpload` succeeded, destination bytes were byte-identical to a 16 MiB position-dependent source, and the destination ETag carried the `-3` multipart suffix. Repeated with a key containing a space, `+`, `&`, `=`, and parentheses (`x-amz-copy-source: <bucket>/val3237/sp%20ace%2Bplus/src%20%281%29%26x%3Dy.bin`) — also `200 OK` and byte-equal, so the `s3_copy_source_value` percent-encoding survives SigV4 header signing.
- **Claim 5 revert-proved live.** I re-ran the same copy while skipping only the entity unescaping, passing MinIO's raw `&#34;55b8dc36...&#34;` straight into `PartId`. `CompleteMultipartUpload` returned `400 InvalidPart`: *"the specified entity tag may not match the part's entity tag"*. The `replace` chain in `parse_copy_part_etag` is load-bearing, not defensive padding.
- **Range arithmetic proved exhaustively.** I extracted `copy_part_ranges` verbatim and checked the partition invariant (starts at 0, ends at `size-1`, no gap, no overlap, lengths sum to `size`, every non-final part exactly `part_size`) for all `size ∈ [1,400) × part_size ∈ [1,40)` plus the named boundaries: threshold, threshold+1, threshold+2, exact multiples, 5 MiB floor, and exactly 10,000/10,001 parts. **Zero failures.** There is no off-by-one.
- **Abort semantics** verified against MinIO: after `abort_multipart`, the upload id is no longer completable and no destination object exists.
- CI is green on `c121c57` (Check Rust, Backend Unit Tests, Code Coverage, Integration, Smoke E2E), so the crate compiles and the three new wiremock tests pass. My probe independently compiles the novel `object_store` API usage.

---

## Claims

**1. Both halves of #3164 were real — CONFIRMED.**
Merge-base `copy()` did `get_stream` → `put_stream` (removed at diff lines 290–310). And `put_stream` calls `complete_multipart` at `s3.rs:1841-1851` and only *then* returns `PutStreamResult` (`s3.rs:1858`), so the merge-base `ensure_copy_digest_matches(...)` ran strictly after the destination had been published. The integrity check was a post-mortem, not a gate.

**2. Signed with the crate's own `AwsAuthorizer` + `AmazonS3::credentials()` — CONFIRMED, with one exception the claim doesn't mention.**
`s3.rs:2057-2067` and `s3.rs:2085`. `AmazonS3::credentials()` returns the store's own `AwsCredentialProvider` (object_store `aws/mod.rs:96`), which is whatever the chain resolved at build time — static keys, web identity/IRSA, `AWS_CONTAINER_CREDENTIALS_*`, EKS Pod Identity, IMDS. Session tokens work: `authorize` inserts `x-amz-security-token` when `credential.token` is `Some` (object_store `aws/credential.rs:176-180`) and `canonicalize_headers` signs every header present. Static keys are live-verified above. **The exception is anonymous mode — see D1.**

**3. URL derived from the configured store via `Signer::signed_url` — CONFIRMED.**
`s3.rs:2042-2051`. `signed_url` → `path_url` → `bucket_endpoint`, which is exactly where object_store resolves path-style vs virtual-hosted vs custom endpoint (`aws/builder.rs:1211-1218`). Nothing is rebuilt. `set_query(None)` then discards the presign params, which is correct — presigned URLs cannot cover `x-amz-*` headers, as the doc comment says.

**4. `200 OK` only accepted with a parseable `<CopyPartResult><ETag>` — CONFIRMED in code, UNTESTED.**
`s3.rs:2115-2123` — a 200 with no parseable ETag is an error, not a success. `parse_copy_part_etag` bails on `<Error` (`s3.rs:426-428`) and on empty/whitespace ETags (`s3.rs:441`), which I verified directly. But **no test in the PR feeds a 200-with-`<Error>` body**, so the specific hazard the claim is about has zero regression coverage. See D4.

**5. Handles both `&#34;` and `&quot;` — CONFIRMED and independently reproduced.**
`s3.rs:433-434`. The unescape ordering is also right: `&amp;` is replaced last (`s3.rs:441`), so `&amp;quot;` correctly decodes to the literal `&quot;` rather than being double-unescaped into a quote.

**6. New `S3Config` field + `S3_MAX_SINGLE_COPY_BYTES` + hard 5 GiB clamp — CONFIRMED.**
Field `s3.rs:502`, env parse `s3.rs:567-570`, builder `s3.rs:723-726`, clamp `s3.rs:935-940` (`clamp(5 MiB, 5 GiB)`), applied at construction `s3.rs:1213`. Note the clamp lives on the backend, not the config — `S3Config::max_single_copy_bytes` itself holds the raw value; only `S3Backend::new` clamps it. That's fine today (one consumer) and matches the doc comment, but the clamp has no test at all (D4).

**7. ≤5 GiB copies byte-for-byte unchanged — CONFIRMED.**
The only change to the small-copy path is the threshold source (`s3.rs:1896`, `S3_MAX_SINGLE_COPY_SIZE` → `self.max_single_copy_bytes`, same default value). `s3.rs:1906-1917` is byte-identical to merge-base, still `self.bulk_store.copy(...)`, so #3222's bulk timeout is intact. `test_copy_at_or_under_threshold_stays_single_native_copy_object` (`s3.rs:4648`) pins it at exactly 5 GiB.

**8. Live test is `AK_S3_E2E`-gated and does not run in CI — CONFIRMED.**
`s3.rs:5163-5168`: `#[ignore]` plus an `AK_S3_E2E == "1"` early return. It lives in `mod integration_tests` (`s3.rs:4941`), which is a `--lib` module. CI's `--lib` runs are `cargo nextest run --workspace --lib` (`ci.yml:479,644`) with no `--ignored`; the only `-- --ignored` invocations (`ci.yml:1114-1143`) are pinned to an explicit list of `--test <binary>` integration targets, none of which is the lib. It cannot fire in CI.

---

## Defects

### D1 — MEDIUM: `S3_ALLOW_ANONYMOUS=true` deployments lose the >ceiling copy path (functional regression)

When no credentials are configured, `build_store` sets `with_skip_signature(true)` (`s3.rs:1080`). But object_store still populates `config.credentials` from the chain, and with no keys/profile that falls through to the **IMDS `InstanceCredentialProvider`** (`aws/builder.rs:1163-1178`). The new path bypasses `skip_signature` entirely: both `signed_url` (`s3.rs:2042`) and `credentials().get_credential()` (`s3.rs:2057`) resolve credentials unconditionally.

Live-confirmed against an anonymous store on a non-EC2 host:

```
signed_url on anonymous store ERRORED: Generic S3 error: Error performing PUT
http://169.254.169.254/latest/api/token in 13.760208029s, after 10 retries,
max_retries: 10, retry_timeout: 180s - HTTP error: error sending request
```

**Failure scenario:** operator runs AK against an unauthenticated S3-compatible store with `S3_ALLOW_ANONYMOUS=true` and promotes an artifact larger than the ceiling. Before this PR the restream path worked (unsigned GET/PUT). After it, `copy()` stalls on the IMDS probe (13.7 s here, up to the 180 s `retry_timeout` depending on how the link-local address behaves on the host) and then fails with `Failed to build UploadPartCopy URL for '<key>': ...169.254.169.254...`, which reads like a network fault rather than a config one. Below-ceiling copies keep working, so it presents as "only big artifacts fail to promote."

**Fix:** carry the unsigned decision onto the backend and honour it. `build_store` already computes it, so thread a `sign_requests: bool` field alongside `bucket`/`region` (`s3.rs:875-886`, set at `s3.rs:1213`). In `upload_part_copy`, when `!sign_requests`, skip both the `signed_url` call and the `AwsAuthorizer::authorize` line and build the URL directly from a `bucket_endpoint: String` captured at construction (`endpoint.trim_end_matches('/') + "/" + bucket`, or `https://s3.{region}.amazonaws.com/{bucket}` when `config.endpoint` is `None` — AK never sets `virtual_hosted_style_request`, so those are the only two shapes), appending `object_store`-style percent-encoded path segments. If that's more than you want here, the acceptable minimum is an explicit guard at the top of `multipart_server_side_copy` returning "server-side copy of objects over the single-copy ceiling requires S3 credentials; unset S3_ALLOW_ANONYMOUS or lower S3_MAX_SINGLE_COPY_BYTES" — still a regression, but a legible one instead of a 3-minute stall.

### D2 — LOW: the 10,000-part guard runs *after* the range vector is materialised

`s3.rs:1945-1956` builds the whole `Vec<(u64,u64)>` and only then checks `ranges.len() > S3_MULTIPART_MAX_PARTS`. The guard is correct (`> 10_000` permits part numbers 1..=10,000, which matches the API), but it cannot fire before the allocation it is meant to protect against.

**Failure scenario:** `size` comes from a `HEAD` `Content-Length` (`s3.rs:1895`). An S3-compatible endpoint that reports a bogus or enormous length makes `copy_part_ranges` loop `size / part_size` times pushing 16-byte tuples before anything rejects it. Measured: `copy_part_ranges(u64::MAX/2, 5 GiB)` produces 1,717,986,919 ranges (~27 GiB resident) and completes rather than erroring. At the 5 MiB floor a reported 1 PiB gives ~2.1e8 ranges (~3.4 GB). There is also a latent `start + part_size` overflow, though the allocation gets you first. Not attacker-reachable through the API — the S3 endpoint is operator-configured — but it's a three-line fix:

```rust
let part_count = size.div_ceil(self.max_single_copy_bytes); // ceiling is clamped >= 5 MiB, never 0
if part_count > S3_MULTIPART_MAX_PARTS as u64 {
    return Err(AppError::Storage(format!(
        "Multipart copy of '{}' would need {} parts, exceeding S3's {}-part limit",
        source, part_count, S3_MULTIPART_MAX_PARTS,
    )));
}
let ranges = copy_part_ranges(size, self.max_single_copy_bytes);
```

### D3 — MEDIUM-LOW: the hand-rolled part copy is the only S3 call in the file with no retry

Every other request here goes through object_store, which marks its own `put_part` `idempotent(true)` and retries per `retry_config` (default 10 retries / 180 s). `upload_part_copy` (`s3.rs:2087-2101`) sends once; any transient `500`/`503 SlowDown`/connection reset becomes a hard error, and `multipart_server_side_copy` (`s3.rs:1979-1982`) then aborts the entire upload with no resume.

**Failure scenario:** a 3 TiB object at the default ceiling is 600 sequential part copies. One `503 SlowDown` on part 400 — routine on a hot prefix — discards ~40 minutes of completed server-side work and surfaces as a failed promotion that succeeds on manual retry. `UploadPartCopy` is idempotent by part number, so a bounded retry (say 3 attempts with backoff on connection errors and 429/5xx) is safe and would match the reliability of every neighbouring call.

Related, and worth a sentence in the doc comment rather than a code change: parts are copied strictly sequentially (`s3.rs:1973`) with a per-part timeout but **no overall deadline**, so a very large copy can run for hours inside one `copy()` call. The old restream path had the same unbounded property, so this is not a regression — but the multipart machinery right above it uses `S3_MULTIPART_MAX_IN_FLIGHT_PARTS = 4`, and a reader will wonder why this one doesn't.

### D4 — LOW: new pure helpers and the new operator knob have no direct tests

I hunted for hollow tests and **found none** — see the test assessment below. The gap is the opposite problem: the PR deletes two pure-function tests (`test_ensure_copy_digest_matches_*`) and adds **zero** pure-function tests for three new pure helpers. Uncovered by anything:

- `copy_part_ranges` boundary behaviour — no direct test (only the two 5 GiB-ceiling wiremock cases).
- `parse_copy_part_etag` — the `<Error`-in-a-200 early return (**the entirety of claim 4**), the empty/whitespace rejection, and the `&#x22;`/`&apos;`/`&#39;`/`&lt;`/`&gt;`/`&amp;` branches. Only `&#34;` and `&quot;` are hit, incidentally, via the two wiremock fixtures.
- `s3_copy_source_value` — no test; percent-encoding of an exotic key is asserted nowhere in CI (I had to verify it out-of-band).
- `effective_single_copy_ceiling` — **the 5 GiB clamp and the 5 MiB floor are asserted by nothing.** This is the one I'd most want covered: the clamp is what stops `S3_MAX_SINGLE_COPY_BYTES=10GiB` from emitting parts AWS rejects, and `=1` from emitting sub-minimum parts. Deleting `.clamp(..)` keeps the whole suite green.
- The `> S3_MULTIPART_MAX_PARTS` error branch (`s3.rs:1946`).

These are cheap table-driven `#[test]`s in the same module and would have caught D2 as well.

Minor: `test_large_copy_upload_part_copy_live_3164` calls `std::env::set_var` (`s3.rs:5175`), which is process-global. It's safe as written (behind `#[ignore]` + `AK_S3_E2E`, and nextest is process-per-test), but this repo has been bitten before by a test mutating process env under `cargo test`, so it's worth a comment noting the single-test invocation is the only supported one.

---

## Test assessment (hollow-test hunt)

I checked all three new wiremock tests against the six species. **No hollow tests.**

- `test_copy_over_5gib_is_server_side_upload_part_copy` (`s3.rs:4419`) — mounts **no GET mock**, so on the merge-base tree `get_stream` hits wiremock's 404 fallback and the `.expect()` panics. Fails on revert. The range assertion is a full `assert_eq!` on the observed vector (`bytes=0-5368709119`, `bytes=5368709120-10737418239`, `bytes=10737418240-10737418240`, all with zero-length bodies), not a shape check — an off-by-one anywhere changes it.
- `test_copy_over_5gib_failed_part_copy_never_publishes_dest` (`s3.rs:4553`) — the important one, and it is built correctly. The fixture **deliberately keeps the pre-fix path viable**: the source GET is mocked green, and `put_stream` creates a multipart upload on the first non-empty chunk regardless of size (`s3.rs:1723-1731`), so on merge-base every request the old shape makes is answered and `copy()` returns `Ok`. All three assertions therefore fail on revert independently — `is_err()` (false), `complete_guard == 0` (would be 1), `abort_guard == 1` (would be 0). That's the "make sure the old code can succeed, then assert it doesn't" discipline, and it's the difference between a guard and a decoration. I also confirmed the mock-precedence assumption it relies on: wiremock sorts by priority with a **stable** sort and breaks on first match (`mock_set.rs:63-71`), and all mocks default to equal priority, so the range-matched 500 registered first does win over the generic 200 PUT.
- `test_copy_at_or_under_threshold_stays_single_native_copy_object` (`s3.rs:4648`) — passes on merge-base too, but it is correctly labelled a positive control, and it is the assertion that stops a future "simplification" from routing the common path into multipart. Exactly the negative-assertion-needs-a-positive-control pattern, and it pins the boundary at *exactly* 5 GiB.

---

## Things I checked that turned out to be non-issues

Recording these so the next reviewer doesn't re-spend the time:

- **`x-amz-copy-source` format.** No leading slash, `bucket/percent-encoded-key`. Byte-identical in construction to object_store's own `CopyObject`/`put_part` (`aws/client.rs:701-703`, `format!("{}/{}", bucket, encode_path(path))`), and `urlencoding::encode` preserves exactly the RFC 3986 unreserved set that `STRICT_PATH_ENCODE_SET` does (`util.rs:303-308`). Live-verified with an exotic key.
- **Canonical-query vs wire-query encoding.** `canonicalize_query` decodes and re-encodes with `STRICT_ENCODE_SET` (`aws/credential.rs:389-415`) while the wire query is form-urlencoded, so they diverge on space/`~`/`*`. Not reachable: `partNumber` is digits and S3 upload ids are base64url-ish. object_store's own `put_part` builds its query the same way (`aws/client.rs:696`), so any divergence would already have broken every multipart upload.
- **Payload hash.** `HttpRequestBody::empty()` → `authorize` takes the `body().is_empty()` branch → `EMPTY_SHA256_HASH` (`aws/credential.rs:195-196`), which matches the bodyless PUT reqwest actually sends. Correct, and live-confirmed.
- **`x-amz-request-payer` / SSE-C copy-source headers / S3 Express session token.** The hand-rolled request omits all three, and object_store's `CopyObject` sets the SSE-C ones (`aws/client.rs:605-633`). All latent, none live: AK's builder never sets `request_payer`, any `encryption_*` key, `s3_express`, `checksum`, `unsigned_payload`, or `virtual_hosted_style_request` (`s3.rs:1016-1054`). Worth a code comment so a future SSE-C or S3 Express change doesn't silently skip this path.
- **`unwrap`/`expect`/panic reachable from a malformed upstream response.** None. `response.text().await.unwrap_or_default()` (`s3.rs:2104`) fails closed — an unreadable body yields `""`, which `parse_copy_part_etag` rejects, which becomes an error. `parse_copy_part_etag` is all `?`/`Option` with no slicing panic: both `find` results are used as forward offsets from a matched ASCII delimiter, so the `body[start..end]` slice is always on a char boundary and `end >= start`. I fuzzed the reversed-tag case (`</ETag>junk<ETag>x`) → `None`, no panic.
- **Abort coverage.** `abort_now()` on part failure (`s3.rs:1980`) and on completion failure (`s3.rs:1995`), `disarm()` only after success (`s3.rs:2001`), and `MultipartAbortGuard`'s `Drop` (`s3.rs:350-374`) spawns a best-effort abort for cancellation. So no path leaks a pending upload: no abort, no accruing storage charges.
- **"Does anything still verify the digest?"** Yes, and it's stronger than what was removed. Each `UploadPartCopy` ETag is captured from the server (`s3.rs:2124`) and handed back to `CompleteMultipartUpload` (`s3.rs:1990-1993`), which the server validates — my `InvalidPart` experiment above *is* the proof that this validation is real and not advisory. The bytes never enter the process, so there is nothing app-side left to hash; the author's "by design" is sound. The check also moved from after-publish to before-publish, which was the actual point of the issue.

Happy to re-check whatever you change for D1.

PR #3240, opened 2026-08-10
fix(scanner): hand the scanner-adapter a registry URL reachable from its own container

### LINKED ISSUE #3169: Scanner-adapter fails to find images
### Description

After updating to 1.7.1 and latest scanner-adapter image, scanner adapter is failing with multiple errors, first is a permissions error on `/home/scanner/.cache/trivy/fanal/fanal.db` I exec'd into container and gave it 666 permissions which resolved that, but now it fails to find docker images with: 

```
Bad gateway: Scanner adapter returned 500 Internal Server Error fetching report: scan failed: trivy scan failed for localhost:8080/docker-remote/library/alpine@sha256:266f29255458134745f2bf588cb23ed1ed1768b96ff2580a05d70a8aba59e145: exit status 1: 2026-08-07T14:19:31Z FATAL Fatal error run error: image scan error: scan error: unable to initialize a scan service: unable to initialize artifact: unable to initialize container image: unable to find the specified image "localhost:8080/docker-remote/library/alpine@sha256:266f29255458134745f2bf588cb23ed1ed1768b96ff2580a05d70a8aba59e145" in ["docker" "containerd" "podman" "remote"]: 4 errors occurred: * docker error: unable to inspect the image (localhost:8080/docker-remote/library/alpine@sha256:266f29255458134745f2bf588cb23ed1ed1768b96ff2580a05d70a8aba59e145): failed to connect to the docker API at unix:///var/run/docker.sock; check if the path is correct and if the daemon is running: dial unix /var/run/docker.sock: connect: no such file or directory * containerd error: containerd socket not found: /run/containerd/containerd.sock * podman error: unable to initialize Podman client: no podman socket found: stat podman/podman.sock: no such file or directory * remote error: Get "https://localhost:8080/v2/": http: server gave HTTP response to HTTPS client; GET http://localhost:8080/v2/: unexpected status code 404 Not Found: 404 page not found
```

### Environment Information

- Deployment type: Docker
- Operating System: Tried on both Mac and Fedora
- Architecture: arm64/amd64
- Database: default
- Reverse Proxy: Caddy 
- Network setup: Docker default

### Backend Version

latest

### Web UI Version

latest

### Package Format

Docker / OCI

### Expected Behavior

Image scans should complete.

### Observed Behavior

Docker image scans fail with errors

### Steps to Reproduce

1. enable scanner-adapter in docker-compose
2. bring up stack
3. scan already existing docker image
4. review scan history tab

### Logs

```shell
2026-08-07T14:19:30.870959Z  INFO artifact_keeper_backend::services::image_scanner: Starting adapter image scan for docker-remote/library/alpine@sha256:266f29255458134745f2bf588cb23ed1ed1768b96ff2580a05d70a8aba59e145
2026-08-07T14:19:35.877752Z ERROR artifact_keeper_backend::services::scanner_service: Scanner container-image failed for artifact 4fa0f60c-87c7-4ee3-9cf0-993cad024aef: Bad gateway: Scanner adapter returned 500 Internal Server Error fetching report: scan failed: trivy scan failed for localhost:8080/docker-remote/library/alpine@sha256:266f29255458134745f2bf588cb23ed1ed1768b96ff2580a05d70a8aba59e145: exit status 1: 2026-08-07T14:19:31Z	FATALFatal error	run error: image scan error: scan error: unable to initialize a scan service: unable to initialize artifact: unable to initialize container image: unable to find the specified image "localhost:8080/docker-remote/library/alpine@sha256:266f29255458134745f2bf588cb23ed1ed1768b96ff2580a05d70a8aba59e145" in ["docker" "containerd" "podman" "remote"]: 4 errors occurred:
	* docker error: unable to inspect the image (localhost:8080/docker-remote/library/alpine@sha256:266f29255458134745f2bf588cb23ed1ed1768b96ff2580a05d70a8aba59e145): failed to connect to the docker API at unix:///var/run/docker.sock; check if the path is correct and if the daemon is running: dial unix /var/run/docker.sock: connect: no such file or directory
	* containerd error: containerd socket not found: /run/containerd/containerd.sock
	* podman error: unable to initialize Podman client: no podman socket found: stat podman/podman.sock: no such file or directory
	* remote error: Get "https://localhost:8080/v2/": http: server gave HTTP response to HTTPS client; GET http://localhost:8080/v2/: unexpected status code 404 Not Found: 404 page not found
```

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
Fixes #3169

## Root cause (mechanism, not symptom)

"Scanner-adapter fails to find images" is two deployment-visible defects, matching the two errors in the report in order:

**1. The backend hands the adapter a registry URL that resolves to the adapter itself.**

`ImageScanner::registry_url()` reused grype's host fallback (`backend/src/services/grype_scanner.rs` → `resolve_registry_host()`), which defaults to `http://localhost:8080` when neither `AK_GRYPE_REGISTRY_HOST` nor `PEER_PUBLIC_ENDPOINT` is set. That default is correct **only for grype**, which execs inside the backend container. The scanner-adapter is a separate container that itself listens on `:8080` (`SCANNER_ADAPTER_ADDR=:8080`, `docker/Dockerfile.scanner-adapter`), so when trivy inside the adapter dials `localhost:8080` it reaches **the adapter's own Go server**, whose default mux answers `404 page not found` on `GET /v2/` — the exact string in the issue log (the axum backend never emits that body; it does serve `/v2` at root, `backend/src/api/routes.rs:146`). Hence trivy's `unable to find the specified image "localhost:8080/docker-remote/library/alpine@sha256:…"`.

No shipped config sets a registry host: the base `docker-compose.yml` enable path (uncomment `TRIVY_ADAPTER_URL` + the `scanner-adapter` service, per README) could never produce a working image scan. The internal test deployments only work because they set `AK_GRYPE_REGISTRY_HOST: backend:8080` explicitly.

**2. The compose adapter block reuses the root-owned trivy cache volume.**

The commented `scanner-adapter` block mounted `trivy_cache:/home/scanner/.cache/trivy` — the same volume the trivy *server* service (running as root) owns at `/root/.cache/trivy`. The adapter runs as UID 1001, producing the reported `permission denied` on `fanal/fanal.db`.

## Fix

- `ImageScanner::registry_url()` now resolves, in order:
  1. **`TRIVY_ADAPTER_REGISTRY_URL`** — new dedicated override (full URL, scheme preserved; bare `host[:port]` gets `http://`), so operators can point the adapter path somewhere without re-pointing grype.
  2. The **explicitly configured** grype chain (`AK_GRYPE_REGISTRY_HOST` / `PEER_PUBLIC_ENDPOINT`) — byte-identical behavior to before.
  3. Nothing configured: when the adapter host is **non-loopback** (i.e. it cannot share this network namespace), the backend advertises its own route-facing address toward the adapter (connected-UDP route lookup, no packet sent) with its `BIND_ADDRESS` port, and logs the substitution. A **loopback** adapter (local dev, `cargo run`) keeps the historical `http://localhost:8080` fallback unchanged. Derivation failure (unresolvable host / no route) also degrades to the historical fallback — never a new failure mode.
- `grype_scanner`: factored `configured_registry_host()` out of `resolve_registry_host()` so callers can distinguish explicit configuration from the dev fallback. Grype behavior unchanged (73/73 module tests green).
- `docker-compose.yml`: dedicated `scanner_adapter_cache` volume for the adapter (fixes the `fanal.db` permission error) + `TRIVY_ADAPTER_REGISTRY_URL` documented next to `TRIVY_ADAPTER_URL`.
- `README.md`: documents the reachability requirement and the cache-volume footgun.

With this, the issue's repro (uncomment two compose blocks, scan an existing docker image) works with zero extra configuration, and a backend-image-only upgrade fixes existing broken deployments.

## Tests

In-file `#[cfg(test)]` (`--lib`, counted by the coverage gate):

- `test_scan_request_registry_url_honors_adapter_registry_override` — **wire-level revert-proof anchor**: captures the actual `POST /api/v1/scan` body via wiremock and pins `registry.url`; the same fixture completes the scan with findings (positive control — a legitimately addressed scan still succeeds, so the fix cannot pass by making everything resolve to a dummy).
- `test_scan_request_registry_url_defaults_loopback_for_loopback_adapter` — same-netns control: a loopback adapter still receives `http://localhost:8080` (passes on both trees by design).
- `test_registry_url_override_shapes`, `test_registry_url_grype_chain_when_configured` — exact-string precedence pins.
- `test_registry_url_remote_adapter_substitutes_own_address` — remote adapter (TEST-NET-1 host) gets this host's route-facing IP + `BIND_ADDRESS` port, exact string; degrades to the documented fallback assertion when the environment has no route.
- `test_registry_url_unresolvable_adapter_keeps_fallback`, `test_host_is_loopback`, `test_url_host_port`.

## Revert-proof (faithful, DB-backed run)

Production code restored byte-for-byte to merge-base (`757eb9f`), keeping only the test subset expressible against the old API. Faithfulness: `git diff origin/main --shortstat` on the **reverted tree** = `1 file changed, 115 insertions(+)` — zero deletions, all inside the test module.

Reverted tree (with `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1` exported):

```
FAIL services::image_scanner::tests::test_scan_request_registry_url_honors_adapter_registry_override
assertion `left == right` failed: the adapter must be told to pull from the configured registry URL, not the loopback dev fallback (#3169)
  left: String("http://localhost:8080")
 right: "http://ak-backend:8080"
Summary: 2 tests run: 1 passed, 1 failed
```

Fixed tree: 8/8 new tests pass; full `image_scanner` (35/35), `grype_scanner` (73/73) modules green; `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean.

## Not covered here (follow-ups worth their own issue if wanted)

- The Helm chart lives outside this repo; chart deployments already set `AK_GRYPE_REGISTRY_HOST` explicitly.
- The adapter image itself is unchanged; the `fanal.db` fix is compose-level (fresh volume). Existing deployments that already share `trivy_cache` should switch to the dedicated volume rather than chmodding inside the container.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,7 +222,15 @@ flowchart LR
 > network namespace (#3169). When nothing is configured the backend now
 > advertises its own container address automatically; set
 > `TRIVY_ADAPTER_REGISTRY_URL` (or the shared `AK_GRYPE_REGISTRY_HOST`) to
-> override, e.g. `http://backend:8080`. Give the adapter its **own** trivy
+> override, e.g. `http://backend:8080`. The automatic derivation is skipped
+> when the adapter's own URL is a loopback address, because that normally means
+> the adapter shares this network namespace (`cargo run` next to a local
+> adapter, or `network_mode: host`). It is **not** skipped correctly for an
+> adapter container published on loopback (`docker run -p 127.0.0.1:8081:8080
+> …scanner-adapter`): there the backend reaches the adapter over loopback but
+> the adapter cannot reach the backend that way, so set
+> `TRIVY_ADAPTER_REGISTRY_URL` explicitly. The backend logs which fallback it
+> took on every scan. Give the adapter its **own** trivy
 > cache volume — reusing the trivy server's root-owned `trivy_cache` fails
 > with `permission denied` on `fanal/fanal.db` because the adapter runs
 > unprivileged (UID 1001).
--- a/README.md
+++ b/README.md
@@ -216,6 +216,17 @@ flowchart LR
 > Trivy image report is produced. Uncomment the `TRIVY_ADAPTER_URL` line and the
 > `scanner-adapter` service in `docker-compose.yml` to enable it.
 >
+> The adapter pulls images back from the registry, so it must be told an
+> address of the backend that is reachable *from the adapter container* —
+> `localhost` is not, since the adapter itself listens on `:8080` in its own
+> network namespace (#3169). When nothing is configured the backend now
+> advertises its own container address automatically; set
+> `TRIVY_ADAPTER_REGISTRY_URL` (or the shared `AK_GRYPE_REGISTRY_HOST`) to
+> override, e.g. `http://backend:8080`. Give the adapter its **own** trivy
+> cache volume — reusing the trivy server's root-owned `trivy_cache` fails
+> with `permission denied` on `fanal/fanal.db` because the adapter runs
+> unprivileged (UID 1001).
+>
 > **Not-applicable scanners.** A scanner that does not apply to an artifact's
 > format (e.g. the filesystem/incus/openscap scanners on a Docker image) records
 > a `not_applicable` result — a benign terminal state, distinct from `failed`.
--- a/backend/src/services/grype_scanner.rs
+++ b/backend/src/services/grype_scanner.rs
@@ -361,6 +361,17 @@ fn parse_grype_output(
 /// `pub(crate)` so `ImageScanner::registry_url` reuses the same host-resolution
 /// logic when telling the Harbor scanner-adapter which registry to pull from.
 pub(crate) fn resolve_registry_host() -> String {
+    configured_registry_host().unwrap_or_else(|| normalize_registry_host("http://localhost:8080"))
+}
+
+/// The *explicitly configured* registry host, if any: `AK_GRYPE_REGISTRY_HOST`
+/// then `PEER_PUBLIC_ENDPOINT`, normalized to `host[:port]`. Returns `None`
+/// when neither is set, i.e. the caller would be on the `localhost:8080` dev
+/// fallback. Split out of [`resolve_registry_host`] so `ImageScanner` can tell
+/// "operator chose this host" apart from "dev fallback" — the fallback is only
+/// correct for pullers that share the backend's network namespace (grype execs
+/// in-process; the scanner-adapter does not — #3169).
+pub(crate) fn configured_registry_host() -> Option<String> {
     let raw = std::env::var("AK_GRYPE_REGISTRY_HOST")
         .ok()
         .filter(|s| !s.is_empty())
@@ -368,9 +379,13 @@ pub(crate) fn resolve_registry_host() -> String {
             std::env::var("PEER_PUBLIC_ENDPOINT")
                 .ok()
                 .filter(|s| !s.is_empty())
-        })
-        .unwrap_or_else(|| "http://localhost:8080".to_string());
+        })?;
+    Some(normalize_registry_host(&raw))
+}
 
+/// Strip scheme, trailing `/`, and any embedded `user[:pass]@` credentials
+/// from a registry endpoint so it can be used as a bare `host[:port]`.
+fn normalize_registry_host(raw: &str) -> String {
     let no_scheme = raw
         .trim_end_matches('/')
         .trim_start_matches("https://")
--- a/backend/src/services/image_scanner.rs
+++ b/backend/src/services/image_scanner.rs
@@ -260,6 +260,89 @@ pub(crate) fn build_adapter_scan_artifact(
     })
 }
 
+/// `host[:port]` of a URL-ish string (scheme optional). IPv6 hosts keep their
+/// brackets. Pure fn.
+fn url_host_port(url: &str) -> Option<(String, u16)> {
+    let candidate = if url.contains("://") {
+        url.to_string()
+    } else {
+        format!("http://{}", url)
+    };
+    let parsed = url::Url::parse(&candidate).ok()?;
+    let host = parsed.host_str()?.to_string();
+    let port = parsed.port_or_known_default().unwrap_or(80);
+    Some((host, port))
+}
+
+/// True when `host` (with or without IPv6 brackets) is a loopback name or
+/// address. Loopback is per-network-namespace: it never names this backend
+/// from inside a different container. Pure fn.
+fn host_is_loopback(host: &str) -> bool {
+    let bare = host.trim_start_matches('[').trim_end_matches(']');
+    bare.eq_ignore_ascii_case("localhost")
+        || bare
+            .parse::<std::net::IpAddr>()
+            .map(|ip| ip.is_loopback())
+            .unwrap_or(false)
+}
+
+/// The local IP this host would use to reach `host:port`, via a connected UDP
+/// socket (route lookup only — no packet is sent). `None` when the host does
+/// not resolve or no route exists.
+fn local_ip_toward(host: &str, port: u16) -> Option<std::net::IpAddr> {
+    use std::net::{SocketAddr, ToSocketAddrs, UdpSocket};
+    let bare = host.trim_start_matches('[').trim_end_matches(']');
+    let addrs: Vec<SocketAddr> = (bare, port).to_socket_addrs().ok()?.collect();
+    for addr in addrs {
+        let bind = if addr.is_ipv4() {
+            "0.0.0.0:0"
+        } else {
+            "[::]:0"
+        };
+        if let Ok(sock) = UdpSocket::bind(bind) {
+            if sock.connect(addr).is_ok() {
+                if let Ok(local) = sock.local_addr() {
+                    return Some(local.ip());
+                }
+            }
+        }
+    }
+    None
+}
+
+/// This backend's own listening port, from `BIND_ADDRESS` (`host:port`,
+/// default `0.0.0.0:8080` — mirrors `Config::from_env`).
+fn own_listen_port() -> u16 {
+    std::env::var("BIND_ADDRESS")
+        .ok()
+        .and_then(|addr| addr.rsplit_once(':').and_then(|(_, p)| p.parse().ok()))
+        .unwrap_or(8080)
+}
+
+/// Registry URL for a *remote* scanner adapter when no registry endpoint is
+/// configured: this backend's own address as routed toward the adapter, with
+/// the backend's listening port. Returns `None` (caller keeps the loopback
+/// dev fallback) when the adapter itself is loopback — i.e. it shares this
+/// network namespace, where `localhost` is correct — or when the adapter host
+/// cannot be resolved/routed.
+fn adapter_visible_registry_url(adapter_url: &str) -> Option<String> {
+    let (host, port) = url_host_port(adapter_url)?;
+    if host_is_loopback(&host) {
+        return None;
+    }
+    let ip = local_ip_toward(&host, port)?;
+    if ip.is_loopback() {
+        // The adapter name resolved to a loopback alias after all; a loopback
+        // registry URL is then no worse than the substitution.
+        return None;
+    }
+    let own_port = own_listen_port();
+    Some(match ip {
+        std::net::IpAddr::V4(v4) => format!("http://{}:{}", v4, own_port),
+        std::net::IpAddr::V6(v6) => format!("http://[{}]:{}", v6, own_port),
+    })
+}
+
 /// Container image scanner that delegates to a Harbor Pluggable Scanner API v1
 /// adapter (e.g. `harbor-scanner-trivy`) over HTTP.
 ///
@@ -340,14 +423,57 @@ impl ImageScanner {
         crate::services::scanner_service::is_oci_image_artifact(artifact)
     }
 
-    /// Registry base URL the adapter should pull from. Host comes from the
-    /// shared `grype_scanner::resolve_registry_host` (AK_GRYPE_REGISTRY_HOST /
-    /// PEER_PUBLIC_ENDPOINT / `localhost:8080`), which strips any scheme and
-    /// embedded credentials; we re-add an `http://` scheme for the adapter's
-    /// registry client.
-    fn registry_url() -> String {
-        let host = crate::services::grype_scanner::resolve_registry_host();
-        format!("http://{}", host)
+    /// Registry base URL the adapter should pull from, in precedence order:
+    ///
+    /// 1. `TRIVY_ADAPTER_REGISTRY_URL` — dedicated override for the adapter
+    ///    path (full URL, scheme preserved; a bare `host[:port]` gets
+    ///    `http://`). Lets operators point the adapter at a different endpoint
+    ///    (e.g. a TLS edge) without re-pointing grype.
+    /// 2. The shared grype chain (`AK_GRYPE_REGISTRY_HOST` /
+    ///    `PEER_PUBLIC_ENDPOINT`) when *explicitly configured* — scheme and
+    ///    credentials stripped, `http://` re-added, exactly as before.
+    /// 3. Nothing configured: the historical `http://localhost:8080` dev
+    ///    fallback is only correct when the adapter shares this process's
+    ///    network namespace (`cargo run` + local adapter). In the documented
+    ///    compose topology the adapter is a SEPARATE container that itself
+    ///    listens on `:8080` (`SCANNER_ADAPTER_ADDR`), so handing it a
+    ///    loopback registry URL makes trivy dial the adapter *itself* and
+    ///    every image scan fails with "unable to find the specified image"
+    ///    (#3169). When the adapter host is non-loopback, advertise this
+    ///    backend's own address on the interface that routes toward the
+    ///    adapter instead.
+    fn registry_url(&self) -> String {
+        if let Some(explicit) = std::env::var("TRIVY_ADAPTER_REGISTRY_URL")
+            .ok()
+            .map(|s| s.trim().trim_end_matches('/').to_string())
+            .filter(|s| !s.is_empty())
+        {
+            return if explicit.starts_with("http://") || explicit.starts_with("https://") {
+                explicit
+            } else {
+                format!("http://{}", explicit)
+            };
+        }
+
+        if let Some(host) = crate::services::grype_scanner::configured_registry_host() {
+            return format!("http://{}", host);
+        }
+
+        match adapter_visible_registry_url(&self.adapter_url) {
+            Some(url) => {
+                info!(
+                    "No registry endpoint configured (TRIVY_ADAPTER_REGISTRY_URL / \
+                     AK_GRYPE_REGISTRY_HOST / PEER_PUBLIC_ENDPOINT); scanner adapter at {} \
+                     is remote, so advertising this backend as {} for image pulls",
+                    self.adapter_url, url
+                );
+                url
+            }
+            None => format!(
+                "http://{}",
+                crate::services::grype_scanner::resolve_registry_host()
+            ),
+        }
     }
 
     /// Mint the `registry.authorization` value for a scan request, or `None`
@@ -611,7 +737,7 @@ impl ImageScanner {
         // BadGateway rather than mid-stream.
         self.check_adapter_health().await?;
 
-        let registry_url = Self::registry_url();
+        let registry_url = self.registry_url();
         // Mint a per-repository scoped pull token only when we know which repo
         // is being scanned (the repository-aware `scan_target` path). The
         // legacy keyless `scan` path pulls anonymously as before.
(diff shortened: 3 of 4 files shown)


brandonrc:
## Independent validation review — head `5e1536de6b202209e7a6c9311e97fd9a99dece31`

**Verdict: diagnosis and code hold up. All six author claims CONFIRMED against source; no regression found in the grype path, and the two `BIND_ADDRESS`/TLS attack lines I went in expecting to break this are refuted by the code.** Three findings, none of them a blocker: one residual topology the fix does not cover, one blocking-DNS-on-the-runtime nit, and one test that cannot fail on the thing it claims to gate. Reviewed from a fresh clone; nothing built or pushed.

Credit where due: the `404 page not found` fingerprint is a genuinely sharp diagnosis. The adapter builds its router with `http.NewServeMux()` (`docker/scanner-adapter/server.go:50`), so that string is stdlib `http.NotFound`, and the same log line carries `http: server gave HTTP response to HTTPS client` — a plain-HTTP Go server on `:8080`, which is the adapter and cannot be the axum backend. That single string identifies the whole bug.

### Claim-by-claim

**1. `registry_url()` reused grype's `localhost:8080` fallback, and the adapter is what listens there — CONFIRMED.**
Base `backend/src/services/image_scanner.rs:349` called `resolve_registry_host()`; base `backend/src/services/grype_scanner.rs:383` ends `.unwrap_or_else(|| "http://localhost:8080".to_string())`. Adapter: `docker/Dockerfile.scanner-adapter:97` `ENV SCANNER_ADAPTER_ADDR=:8080`, `:101` `USER 1001`, `:102` `EXPOSE 8080`; the commented compose block sets `SCANNER_ADAPTER_ADDR: ":8080"` explicitly.

**2. Go-mux fingerprint — CONFIRMED**, with one citation correction: the `GET /v2/` version check is `backend/src/api/routes.rs:145` (`.route("/v2/", handlers::oci_v2::version_check_handler())`); `:146` is the `.nest("/v2", …)` on the next line. Same statement, but worth pinning exactly since the whole argument rests on "the backend does answer `/v2/`".

**3. No shipped config sets a registry host — CONFIRMED.** `grep -rn 'AK_GRYPE_REGISTRY_HOST\|PEER_PUBLIC_ENDPOINT'` across every `*.yml/*.yaml/*.env/*.sh/Dockerfile*` in the repo returns zero hits. The documented compose enable path could not have produced a working image scan.

**4. Root-owned `trivy_cache` reused by the UID-1001 adapter — CONFIRMED**, and I checked the way this fix could have been cosmetic: a *dedicated* empty named volume only solves the permission error if the image's directory is already owned by 1001, because Docker copies ownership from the image when it initialises an empty volume. `docker/Dockerfile.scanner-adapter:74-77` does `adduser -D -u 1001 scanner && mkdir -p /home/scanner/.cache/trivy && chown -R scanner:scanner /home/scanner`, so it is. The base mount collided with `docker-compose.yml:351` (`trivy_cache:/root/.cache/trivy`, trivy server as root). No other shipped manifest mounts `trivy_cache` into an adapter.

**5. New precedence — CONFIRMED** as written (`image_scanner.rs:445-486`): dedicated override → explicit grype chain → route-derived own address for a non-loopback adapter → historical loopback fallback on loopback-adapter or derivation failure.

**6. Grype behaviour unchanged — CONFIRMED, byte-for-byte equivalent.** New `resolve_registry_host()` is `configured_registry_host().unwrap_or_else(|| normalize_registry_host("http://localhost:8080"))`; `normalize_registry_host` (`grype_scanner.rs:387-405`) is the old body verbatim, and the old code applied that same normalisation to the same literal. Both trees return `localhost:8080`. No precedence step reordered or dropped, `AK_GRYPE_REGISTRY_HOST` still beats `PEER_PUBLIC_ENDPOINT`, empty-string filtering intact, and the three other grype call sites (`grype_scanner.rs:664,693,716`) still go through `resolve_registry_host()`. `ImageScanner::registry_url` becoming `&self` is safe: the only call site in the tree is `image_scanner.rs:740`.

### Attack lines that did NOT hold up (refuted)

- **`BIND_ADDRESS` unset / unix socket / wrong port.** `backend/src/config.rs:1052` defaults to `0.0.0.0:8080`, and `backend/src/main.rs:1076` parses it as a `SocketAddr` — so it is *always* `host:port` and a unix socket would fail the process at boot, never reach `own_listen_port()`. `rsplit_once(':')` also handles `[::]:8080` correctly. The published-vs-bound port worry does not apply either: the derived address is the backend's *container* address on the compose network, which pairs with the *container* port, not the published one. (Unrelated pre-existing oddity: `docker-compose.yml:298-299` sets `HOST`/`PORT`, which the backend never reads. Not this PR's problem; a follow-up issue if anyone cares.)
- **Hardcoded `http://` in the substitution.** The backend has no TLS listener — `main.rs` installs a rustls provider for *outbound* only, and Caddy terminates. `http://` is correct.
- **Loopback detection misfiring on `127.0.0.1` vs `localhost` vs `::1`.** `host_is_loopback` (`image_scanner.rs:280-290`) strips IPv6 brackets, case-folds `localhost`, and uses `IpAddr::is_loopback` for the whole `127.0.0.0/8` and `::1`. There is also a second guard on the *derived* IP at `:334-337`. That part is right — the problem is semantic, see F1.
- **Token leaking to the wrong host after the substitution.** The adapter keys its docker `config.json` on `trimRegistryHost(req.Registry.URL)` (`docker/scanner-adapter/scan.go:113-119,137-146`), which yields the same `IP:port` string as the image ref host, so the repo-scoped bearer still applies to exactly one host and still does not leak into trivy's vuln-DB pull.

### Findings

**F1 — MEDIUM: the loopback carve-out conflates "reached over loopback" with "shares my network namespace", leaving a plausible topology still broken and silent.**
`adapter_visible_registry_url` returns `None` whenever the adapter host is loopback (`image_scanner.rs:330-331`), on the stated rationale that loopback implies the same netns. That equivalence does not hold for a container published on loopback. Concrete scenario: backend on the host (`cargo run`, systemd, or `network_mode: host`) bound `0.0.0.0:8080`, adapter run as `docker run -p 127.0.0.1:8081:8080 …scanner-adapter`, `TRIVY_ADAPTER_URL=http://127.0.0.1:8081`. Loopback → `None` → `http://localhost:8080` → inside the adapter's own netns that is the adapter's Go server on `:8080` → the exact #3169 failure, unfixed. The operator also gets no hint, because the `info!` at `:464` only fires on the substitution branch. Cheap remedy: log on the loopback branch too — one `debug!`/`warn!` saying "adapter at {} is loopback, advertising the historical `http://localhost:8080`; if the adapter is a separate container reached through a published port rather than sharing this network namespace, set `TRIVY_ADAPTER_REGISTRY_URL`". Not a regression (that topology was broken before), but it is inside the class the PR claims to close, so it should be named in the README rather than left implied.

**F2 — LOW/MED: blocking `getaddrinfo` on a tokio worker, once per image scan, uncached.**
`local_ip_toward` resolves via `(bare, port).to_socket_addrs()` (`image_scanner.rs:295`) — a synchronous libc `getaddrinfo`. It is reached from `registry_url()`, which is called at `image_scanner.rs:740` inside `async fn run_image_scan`, with no `spawn_blocking` and no memoisation. Healthy compose DNS makes this sub-millisecond, but a slow or unreachable resolver parks a runtime worker for the full resolver budget (`resolv.conf` default 5s × 2 attempts) on *every* scan, and this is now the default path for exactly the compose users the PR targets. This repo has a documented worker-starvation failure class of precisely this shape. Three cheap fixes, any of which is fine: `tokio::net::lookup_host` (the caller is already async), `spawn_blocking`, or memoise the derived URL per `adapter_url` in a `OnceCell`/`RwLock` — the value cannot change without a restart in practice.

**F3 — MEDIUM (test quality): the only test of the new derivation computes its expected value from the production helper it is testing, and the "revert-proof anchor" gates the new env var rather than the reported bug.**
`test_registry_url_remote_adapter_substitutes_own_address` does `let Some(ip) = local_ip_toward("192.0.2.1", 8080) else { … }` and then builds `expected` from that same `ip`. It genuinely pins the string format, the `BIND_ADDRESS` port, and non-loopbackness — but it cannot detect a wrong-interface derivation, which is the single highest-risk property of this change. Its no-route branch also asserts the fallback and returns, so in an environment without a default route the test passes vacuously. Separately, the test the PR body names as the revert-proof anchor (`…honors_adapter_registry_override`) proves that a *newly added* env var is read — true by construction of any new feature — not that #3169 is fixed. The actual reported configuration (nothing set, remote adapter) has no wire-level test at all. Suggestion: bind the wiremock server to a route-facing interface address instead of loopback, run the same `run_scan_and_capture_request` fixture with no env set, and assert the on-the-wire `registry.url` is non-loopback and carries the backend's own port. That gates the real branch end-to-end and would fail on the base tree.

**F4 — process note, not a code defect: the evidence here is unit-level only.**
No live `docker compose` reproduction is recorded on either tree, and the change is entirely about deployment topology. Worth one real run before merge, because of one interaction I could not settle by reading: the OCI 401 challenge realm comes from `RequestBaseUrl` (`backend/src/api/extractors.rs:203-226`), which prefers `AK_EXTERNAL_URL` over the request `Host`. A deployment that sets `AK_EXTERNAL_URL=https://…` (the reporter runs a Caddy edge) will hand trivy a realm at the external hostname even though it dialled the derived container IP; if the adapter container cannot reach that hostname, an anonymous pull fails one step later than before. The shipped compose does not set `AK_EXTERNAL_URL`, and repo-scoped scans carry a bearer so they never take the challenge path, so the default topology should be fine — but "should be" is what a compose run would replace with "is". Uncommenting the two blocks and scanning an existing `docker-remote` image is a ten-minute check that also exercises the fresh `scanner_adapter_cache` volume.

Minor, no action needed: the compose volume fix only reaches deployments that uncomment the block fresh; existing installs keep the root-owned `trivy_cache` until they `docker volume rm` and recreate. The PR body already says so; one sentence in the README aimed at existing users would close it.

CI is fully green on this SHA (Backend Unit Tests, Integration, Coverage, Check Rust), so compilation and the new tests are settled — I did not rebuild locally.

brandonrc:
## Review remediation — pushed `1f964d4`

All three findings addressed. Merge base `757eb9f`; nothing merged, nothing rebased.

---

### F3 — the derivation test was self-referential, and the named anchor gated the wrong thing

**Confirmed as reported.** `test_registry_url_remote_adapter_substitutes_own_address` built `expected` by calling `local_ip_toward` — the production route helper under test — so a wrong-interface derivation produced the same string on both sides of the `assert_eq!`. Its no-route branch also asserted the fallback and `return`ed, so it could pass vacuously. And the anchor the PR body named (`…honors_adapter_registry_override`) only proves a newly added env var is read.

**What changed.**

1. **The decision is now injectable (review option 1).** The pure core came out of `adapter_visible_registry_url`:

   ```rust
   fn build_adapter_registry_url(
       adapter_host: &str,
       local_ip: Option<std::net::IpAddr>,
       own_port: u16,
   ) -> Option<String>
   ```

   Production passes the real resolver and `own_listen_port()`; the tests pass values they choose. The expectations are now **string literals**:

   ```rust
   assert_eq!(
       build_adapter_registry_url("scanner-adapter", Some("198.51.100.7".parse()?), 18080),
       Some("http://198.51.100.7:18080".to_string())
   );
   assert_eq!(
       build_adapter_registry_url("scanner-adapter", Some("2001:db8::5".parse()?), 9443),
       Some("http://[2001:db8::5]:9443".to_string())
   );
   ```

   `198.51.100.7` is TEST-NET-2 (RFC 5737 §3) and `2001:db8::5` is the documentation prefix (RFC 3849 §4). Neither is ever assigned to the host running the test, so **no route derivation can produce these expectations**, correct or wrong — which is exactly the property the old test lacked. The second case uses a third distinct port so `own_port` cannot be ignored or hardcoded either.

2. **The substitution is proved to actually substitute.** `test_build_adapter_registry_url_replaces_the_adapter_host` runs stubbed adapter hosts (`scanner-adapter`, `trivy-adapter.svc`, `203.0.113.9`) against a stubbed routed address and asserts the advertised host is not loopback, is **not the adapter's own host**, and carries this backend's port.

3. **Destination-sensitivity is tested directly.** The highest-risk property — "the interface facing the adapter", not "some interface" — is that the derivation *depends on its destination*. `test_local_ip_toward_is_destination_sensitive` asserts the source address toward loopback is loopback, the source address toward TEST-NET-1 is not, and the two differ. It then binds the derived address: the kernel returns `EADDRNOTAVAIL` for an address not assigned to a local interface, so this is an OS-verified check that is independent of how the address was derived.

4. **The reported configuration now has a wire-level test.** `test_scan_request_registry_url_substitutes_for_remote_adapter_on_the_wire` sets **nothing**, binds the mock adapter to a non-loopback local address (`MockServer::builder().listener(…)`, same pattern as the PyPI remote tests), runs a full scan and asserts the submitted `POST /api/v1/scan` body. The expected URL is anchored to *the address the mock adapter is actually listening on* — a fact the scan succeeding through that listener proves — not to anything the derivation computes. The same fixture is the positive control (one finding returned).

5. **No vacuous pass left.** Both route-dependent tests now `eprintln!` an explicit SKIP with a reason and return, instead of asserting the fallback. Neither skipped in this environment (verified with `--nocapture`: no `SKIP` line printed, 38/38 ran).

6. The env-var test's doc comment no longer calls itself the `#3169` revert-proof anchor and points at the wire test instead.

#### Revert-proof (faithful, against the merge base)

Production restored to `757eb9f` **byte for byte**; only the new wire-level test and its fixtures added. (The wire test deliberately uses no helper introduced by this PR, so it compiles against the pre-fix tree — the pure-core tests cannot be revert-proved this way because the function they test does not exist before the fix; they are mutation-proved below instead.)

Faithfulness, run on the **reverted** tree:

```
$ git diff 757eb9f --shortstat
 1 file changed, 154 insertions(+)

$ git diff 757eb9f -U0 backend/src/services/image_scanner.rs | head -5
@@ -1424,0 +1425,154 @@ mod tests {
```

Insertions only, zero deletions, and the single hunk opens inside `mod tests {` (which begins at line 771) — production is identical to the merge base, so the failure below can only come from the claim under test.

```
$ DATABASE_URL=postgresql://…@127.0.0.1:30441/artifact_keeper AK_TESTS_REQUIRE_DB=1 \
  cargo test --lib services::image_scanner::tests::test_scan_request_registry_url -- --nocapture

thread '…::test_scan_request_registry_url_substitutes_for_remote_adapter_on_the_wire'
panicked at backend/src/services/image_scanner.rs:1570:9:
assertion `left == right` failed: with nothing configured, a remote adapter must be
told to pull from the address it can actually reach this backend on — not the loopback
fallback, which inside the adapter's own network namespace is the adapter (#3169)
  left: String("http://localhost:8080")
 right: "http://192.168.42.14:18080"

test result: FAILED. 0 passed; 1 failed
```

It fails on the derivation claim itself — the loopback fallback on the wire versus the routable address — not on a missing env var. (`DATABASE_URL` + `AK_TESTS_REQUIRE_DB` were exported per house rules; nothing on this path is DB-backed, the scan runs against wiremock.)

#### Mutation proofs for the pure-core tests

Three wrong derivations, each applied to the fixed tree on its own:

| Mutation | Caught by |
|---|---|
| A. advertise the **adapter's** host instead of the routed local address | `…uses_supplied_address_verbatim`, `…replaces_the_adapter_host`, `…substitutes_own_address` — 3 FAILED |
| B. route toward a **fixed destination** (`127.0.0.1`) instead of the adapter | `…substitutes_own_address`, `…on_the_wire` — 2 FAILED |
| C. destination-**insensitive** route lookup (`connect` to a hardcoded address) | `…is_destination_sensitive` — 1 FAILED |

Worth recording: mutation A does **not** fail the wire test, because in that fixture the adapter's host *is* the backend's address, so advertising the adapter coincidentally gives the right answer. That is precisely why the literal-valued pure-core assertions had to exist alongside the end-to-end one.

---

### F1 — the loopback carve-out failed silently

**Confirmed, and behaviour deliberately unchanged.** A scanner-adapter container published on loopback (`docker run -p 127.0.0.1:8081:8080 …`) still takes the fallback and still hits #3169 — nothing observable at that point distinguishes it from a genuine same-namespace adapter — but it is no longer silent. Every `None` branch now logs at `info` (the level the substitute branch already used, visible at the default filter) and names the override:

* loopback adapter host,
* adapter host that *resolves* to loopback,
* no derivable route,
* and an adapter URL with no parseable host (`warn`).

Each message names `TRIVY_ADAPTER_REGISTRY_URL` and states the failure mode. A code comment on the loopback branch records why loopback is read as same-network-namespace (`cargo run` next to a local adapter, `network_mode: host`) and the published-port topology where that assumption breaks; the README paragraph added by this PR now says the same for operators.

---

### F2 — blocking `getaddrinfo` on a tokio worker

**Fixed — it was a clean change.** `local_ip_toward` split in two:

* name resolution now uses `tokio::net::lookup_host`, so a slow or unreachable resolver can no longer park a worker for the resolver budget on every scan (this is now the default path);
* the route lookup stays synchronous in `local_ip_for`, because a UDP `connect` to an *already resolved* `SocketAddr` is a kernel route lookup — no name resolution, no packet, no blocking.

`registry_url` became `async` (single call site, `run_image_scan:740`, already async). Not memoised: the env vars and the routing table are read per scan, which keeps the existing tests honest and costs one non-blocking syscall.

---

### Gates

```
cargo fmt --all -- --check      clean
cargo clippy --workspace --all-targets -- -D warnings    clean
cargo test --lib scanner        748 passed; 0 failed
cargo test --lib services::image_scanner    38 passed; 0 failed (no SKIPs taken)
```

No new `sqlx::query!` macros, so no `cargo sqlx prepare` needed.

PR #3241, opened 2026-08-10
fix(maven): declare upstream Content-Encoding on download_root verbatim forwards

### LINKED ISSUE #3211: Maven download_root forwards bytes verbatim but drops upstream Content-Encoding
Found while fixing #3193 (PR #3210), deliberately left out of scope there.

## The gap

`backend/src/api/handlers/maven.rs::download_root` — both the Remote and Virtual-member arms — forwards upstream bytes **verbatim**: it copies the upstream `Content-Type` and sets `Content-Length` from `content.len()`. But it fetches through `fetch_artifact_with_cache_path`, which **drops** the upstream `Content-Encoding`.

So a coded artifact reaches a Maven client as undeclared compressed bytes with a `Content-Length` describing the *coded* length. Same class as #3149, a different handler.

## Why it wasn't fixed in #3210

#3210 widened `fetch_upstream_direct_with_link` (the PEP 658 / OCI-tags primitive). `fetch_artifact_with_cache_path` is a different primitive with its own caller set, and widening it is a separate blast radius that deserves its own audit rather than riding along in a PyPI fix.

#3210 documented that primitive as "DROPS the coding" precisely so this is visible at the call site instead of being rediscovered later.

## Suggested approach

Follow the pattern already established by #3194 and #3210: widen `fetch_artifact_with_cache_path` **in place** rather than adding a `_with_encoding` sibling. In-place widening breaks every caller at compile time and forces each to declare whether it forwards bytes verbatim or transforms them. A sibling lets the next caller silently reintroduce the bug — which is how #3149 outlived #3176.

Audit every caller and classify it:
- **forwards verbatim** → must re-declare the coding
- **parses/rewrites** → must decode before parsing, and must not re-declare

## Test requirement

Use a **non-gzip** coding. An earlier PR shipped fourteen tests that all passed with the forwarded value hardcoded to `"gzip"`, because every fixture was gzip — they pinned "a coding is declared", not "the UPSTREAM's coding is declared". Helpers exist on main: `tdh::coded_fixture("deflate", seed)` and `tdh::inflate_deflate(bytes)`. Include a negative control (uncoded upstream ⇒ no header) and at least one test that actually decodes the served body.

brandonrc:
Fixes #3211

## The bug

`maven.rs::download_root` — both the Remote arm and the Virtual-member arm — forwards the upstream root body **verbatim**: it copies the upstream `Content-Type`, recomputes `Content-Length` from `content.len()`, and streams the bytes untouched. But it fetched through `ProxyService::fetch_artifact_with_cache_path`, which dropped the upstream `Content-Encoding` (the primitive's own doc said so, left as a visible seam by #3210).

Per RFC 9110 **§8.4**, `Content-Encoding` is a representation header describing the coding of the bytes *as transferred*; per **§8.6**, `Content-Length` must describe the octets actually sent. The handler sent the coded octets with a matching (coded) `Content-Length` but no coding declaration — so a client receives compressed bytes it has no reason to decode and stores a compressed document under an identity implying it is plain. Same class as #3149, in a different handler.

Verified against source before fixing: `Content-Type` survived the forward, `Content-Length` was recomputed from the coded bytes (and therefore *matched* what was sent — §8.6 was satisfied; the only violation was the missing §8.4 declaration). `Content-Encoding` was the one header dropped.

## The fix

Followed the in-place-widening pattern established by #3194 / #3210 / #3184: **no `_with_encoding` sibling**, so every caller breaks at compile time and must classify itself.

* `ProxyService::fetch_artifact_with_cache_path` and `fetch_artifact_with_cache_path_and_accept` now return `CachedBody` (`body, content_type, content_encoding`), delegating to the already-widened capped primitive. The cache sidecar already persists the coding, so cache hits re-emit it identically to cold fetches.
* `maven.rs::download_root` (both arms) — **verbatim forwarders, the bug** — now re-declare the upstream coding via a shared `forward_root_verbatim` builder: `Content-Encoding` set iff the upstream (or cache sidecar) reported one; `Content-Length` remains the coded length, which is correct because the coded bytes are what is sent (§8.6).

### Caller audit (the full blast radius — nothing silently widened)

Every caller of the two widened methods, classified per the issue:

| Caller | Classification | Action |
|---|---|---|
| `maven.rs::download_root` Remote arm | forwards verbatim | re-declares the coding (**the fix**) |
| `maven.rs::download_root` Virtual arm | forwards verbatim | re-declares the coding (**the fix**) |
| `ProxyService::fetch_artifact` | convenience wrapper for parse/rewrite callers (virtual-metadata merge closures, pub registry JSON rewriter) | keeps its `(Bytes, Option<String>)` shape, now explicitly destructures-and-drops with a doc stating so (mirrors `fetch_artifact_capped`) |
| `ProxyService::fetch_artifact_with_accept` | same stance | same |
| `proxy_helpers::proxy_fetch_with_cache_key` | **no callers in-tree** | widened to the triple + doc, so a future adopter sees the coding instead of silently reintroducing #3149 |
| `proxy_helpers::proxy_fetch_with_cache_key_and_accept` | **no callers in-tree** | same |
| in-file `proxy_service.rs` tests (8 destructures) | test call sites | third element added |

So the *behavioral* change is Maven-only (`download_root`); the *seam* change is shared, and no other format's serving behavior changes in this PR.

### Known residuals (documented, deliberately out of scope)

* `proxy_helpers::proxy_fetch_or_redirect` (no production caller, per its own comment) forwards verbatim through `fetch_artifact` and would drop a coding if adopted; its buffered fallback arm carries the same latent class.
* `pub_registry.rs::proxy_pub_meta_get` has a verbatim pass-through fallback for non-JSON upstream bodies.

Both go through `fetch_artifact`, which is documented as dropping; fixing them means the decode-or-declare treatment per caller and belongs to a follow-up, not riding along here.

## Tests

DB-backed `--lib` test in `maven.rs` (counts toward the coverage gate), `test_download_root_forwards_upstream_content_encoding_verbatim`, per the issue's test requirement:

* **Non-gzip coding** (`tdh::coded_fixture("deflate", …)`) — a handler hardcoding `"gzip"` cannot pass.
* **Uncoded upstream control in the same fixture** — no spurious `Content-Encoding`, body byte-identical — so hardcoding `"deflate"` cannot pass either.
* Asserts the actual bytes received (verbatim coded bytes, `Content-Length` == coded length) and **decodes the served body** with `tdh::inflate_deflate`, proving the declared coding matches the bytes.
* Covers **both** arms that had the bug (Remote, and Virtual via a member), plus the control.

### Revert-proof (faithful, DB-backed)

Production files restored byte-identical to merge-base `757eb9f`, tests kept. On the reverted tree:

```
$ git diff 757eb9f --shortstat
 1 file changed, 154 insertions(+)      # single hunk, @@ mod tests — insertions only, all in the test module
```

```
FAIL api::handlers::maven::tests::test_download_root_forwards_upstream_content_encoding_verbatim   [0.31s]
  assertion `left == right` failed: remote root must re-declare the upstream Content-Encoding (#3211)
    left: None
    right: Some("deflate")
PASS api::handlers::maven::tests::test_download_root_forwards_remote_and_virtual_but_404s_local    [0.39s]
```

Run with `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1` (timings show real DB execution, not the try_pool skip). With the fix applied the test passes; full `cargo nextest run --workspace --lib`, `cargo fmt --all --check`, and `cargo clippy --workspace --all-targets -- -D warnings` are green locally.

### DIFF
--- a/backend/src/api/handlers/maven.rs
+++ b/backend/src/api/handlers/maven.rs
@@ -13,7 +13,7 @@ use std::time::Duration;
 
 use axum::body::Body;
 use axum::extract::{Path, State};
-use axum::http::header::{CONTENT_LENGTH, CONTENT_TYPE};
+use axum::http::header::{CONTENT_ENCODING, CONTENT_LENGTH, CONTENT_TYPE};
 use axum::http::{HeaderMap, StatusCode};
 use axum::response::{IntoResponse, Response};
 use axum::routing::get;
@@ -915,17 +915,15 @@ async fn download_root(
             // under the non-empty sentinel key "_root_" to satisfy the cache-
             // path validation that rejects empty strings.
             let remote = proxy_helpers::build_remote_repo(repo.id, &repo_key, upstream_url);
-            let (content, content_type) = proxy
+            let (content, content_type, content_encoding) = proxy
                 .fetch_artifact_with_cache_path(&remote, "", "_root_")
                 .await
                 .map_err(|e| e.into_response())?;
-            let ct = content_type.unwrap_or_else(|| "text/html".to_string());
-            return Ok(Response::builder()
-                .status(StatusCode::OK)
-                .header(CONTENT_TYPE, ct)
-                .header(CONTENT_LENGTH, content.len().to_string())
-                .body(Body::from(content))
-                .unwrap());
+            return Ok(forward_root_verbatim(
+                content,
+                content_type,
+                content_encoding,
+            ));
         }
     }
 
@@ -938,17 +936,15 @@ async fn download_root(
                 {
                     let remote =
                         proxy_helpers::build_remote_repo(member.id, &member.key, upstream_url);
-                    if let Ok((content, content_type)) = proxy
+                    if let Ok((content, content_type, content_encoding)) = proxy
                         .fetch_artifact_with_cache_path(&remote, "", "_root_")
                         .await
                     {
-                        let ct = content_type.unwrap_or_else(|| "text/html".to_string());
-                        return Ok(Response::builder()
-                            .status(StatusCode::OK)
-                            .header(CONTENT_TYPE, ct)
-                            .header(CONTENT_LENGTH, content.len().to_string())
-                            .body(Body::from(content))
-                            .unwrap());
+                        return Ok(forward_root_verbatim(
+                            content,
+                            content_type,
+                            content_encoding,
+                        ));
                     }
                 }
             }
@@ -956,6 +952,31 @@ async fn download_root(
     }
 
     Err(AppError::NotFound("Repository root not available".to_string()).into_response())
+}
+
+/// Build the response for an upstream root body that is forwarded VERBATIM.
+///
+/// The bytes are sent exactly as the upstream (or the proxy cache) produced
+/// them, so the upstream `Content-Encoding` must be re-declared when present
+/// (RFC 9110 §8.4 — the representation header describes the coding applied to
+/// the bytes as transferred), and `Content-Length` is the length of those
+/// coded bytes (RFC 9110 §8.6). Dropping the coding while keeping the coded
+/// bytes was #3211: clients stored a compressed root document as if it were
+/// plain.
+fn forward_root_verbatim(
+    content: Bytes,
+    content_type: Option<String>,
+    content_encoding: Option<String>,
+) -> Response {
+    let ct = content_type.unwrap_or_else(|| "text/html".to_string());
+    let mut builder = Response::builder()
+        .status(StatusCode::OK)
+        .header(CONTENT_TYPE, ct)
+        .header(CONTENT_LENGTH, content.len().to_string());
+    if let Some(enc) = content_encoding {
+        builder = builder.header(CONTENT_ENCODING, enc);
+    }
+    builder.body(Body::from(content)).unwrap()
 }
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


brandonrc:
## Independent validation — verdict: fix is correct; one substantive finding (documentation/scope, not a regression)

Reviewed from a **fresh clone at the PR head SHA `426588e8627684aa91ab0ec65e3e8fbab08a85a6`**, against its own Postgres. I tried to break each claim rather than confirm it. The production change is minimal, correct, and I could not find a way to make it serve a wrong body or a wrong length. The one real finding is that the **"Known residuals" list is not exhaustive and the classification that justifies leaving `fetch_artifact` dropping is inverted for one of its callers** — five live verbatim forwards in three other formats have exactly this defect and are not listed or tracked anywhere.

---

### Claim 1 — `Content-Length` recomputed, only `Content-Encoding` dropped — **CONFIRMED**

At merge-base `757eb9f`, `backend/src/api/handlers/maven.rs` `download_root` both arms build the response with `.header(CONTENT_LENGTH, content.len().to_string())` over the buffered `content`. The upstream's own `Content-Length` is never copied onto the response — it is parsed into `UpstreamHeaders.content_length` (`backend/src/services/proxy_service.rs:601`) and used only for cache accounting. So RFC 9110 §8.6 was satisfied before the fix and still is; there is no response-smuggling-adjacent length/body divergence here, on either arm.

This holds on the **cache-hit** path too, not just cache-miss: a hit returns the stored `Bytes` (`backend/src/services/proxy_service.rs:2990`) and the same `content.len()` is what gets declared.

I also checked the two ways a length could go wrong and neither applies:

* **Truncation.** The bounded buffered read *rejects* an over-cap body with `BadGateway` rather than truncating (`backend/src/services/proxy_service.rs:2143-2157`), so a partial body is never paired with a full length.
* **Re-encoding downstream.** The only `CompressionLayer` in the tree is npm-scoped (`backend/src/api/handlers/npm.rs:121`); maven routes are not behind it, so nothing re-codes the bytes after this builder.

### The inverse failure (declaring a coding on already-decoded bytes) — **ruled out**

This was the most dangerous possibility and it does not occur. `ProxyService` builds its client from `http_client::base_client_builder()` (`backend/src/services/proxy_service.rs:2648`), which sets `.no_gzip().no_brotli().no_deflate().no_zstd()` (`backend/src/services/http_client.rs:93-96`) and pins an explicit `Accept-Encoding: identity`. Nothing between the socket and `forward_root_verbatim` touches the body. Re-declaring the upstream coding on those bytes is therefore correct, not corrupting.

### Cache-hit path re-declares the coding — **CONFIRMED**

Worth stating explicitly because a fix that only covered cold fetches would be half a fix:

* the sidecar persists it — `CacheMetadata { content_encoding, .. }` in `write_buffered` (`backend/src/services/proxy_service.rs:1478`), wired positionally from `upstream_encoding` through `cache_artifact` (`:5583`) with the parameter order matching;
* every read path returns it — fresh hit (`:1198-1202`, `:5125`), 304 revalidation (`:5226`), stale-if-error (`:5239` and `:3228`).

Incidentally the test's Virtual probe *is* a cache-hit probe: the Remote probe already populated `_root_` under the same member key, so the second assertion exercises the sidecar replay. That is extra coverage, not a gap — but it means the "cold virtual fetch" arm is covered only transitively.

### Claim 3 — every caller of the two widened methods audited — **CONFIRMED**

Enumerated independently rather than trusting the table. `fetch_artifact_with_cache_path` / `..._and_accept` callers in the whole tree: `maven.rs:919`, `maven.rs:942`, `proxy_helpers.rs:1444`, `proxy_helpers.rs:1472`, plus the in-file test destructures. That is the complete set; the table is right. `cargo check --workspace --all-targets` (SQLX_OFFLINE) on the head SHA: **0 errors, 0 warnings**.

### Claim 5 — `proxy_fetch_with_cache_key{,_and_accept}` have zero in-tree callers — **CONFIRMED**

Only the definitions and doc references exist. Widening them defensively costs nothing.

### Claim 7 — the test — **CONFIRMED, and it is a good one**

Non-gzip coding so a hardcoded `"gzip"` cannot pass; an uncoded control **in the same fixture** so a hardcoded `"deflate"` cannot pass either; asserts the verbatim coded bytes and the coded `Content-Length`; and actually inflates the served body, so it pins "the declared coding matches the bytes" rather than "a header is present". Both arms that had the bug are covered. Using the fully-qualified `axum::http::header::CONTENT_ENCODING` so the test compiles unchanged against pre-fix production code is the right call for a faithful revert-proof.

**Independently executed and independently revert-proved**, against my own Postgres (`DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1`, so these are real runs, not the `try_pool` early return — the timings confirm it):

* head SHA, as shipped — `2 tests run: 2 passed` (`0.673s` / `0.603s`, not the ~0.04s of a skipped DB test);
* **faithful revert** — `proxy_service.rs` and `proxy_helpers.rs` restored from `757eb9f`, `download_root` restored to its merge-base shape, test module kept. `git diff 757eb9f --shortstat` on the reverted tree: `1 file changed, 154 insertions(+)` — insertions only, single hunk at `@@ -4783,0 +4784,154 @@ mod tests`, so production is byte-identical to the merge base and the red can only come from the claim. Result: `FAIL … assertion left == right failed: remote root must re-declare the upstream Content-Encoding (#3211) / left: None / right: Some("deflate")`;
* **mutation, the one the issue explicitly warns about** — hardcoding the forwarded value to `"gzip"` instead of echoing the upstream's. Result: `FAIL … left: Some("gzip") / right: Some("deflate")`. The test is not the "a coding is declared" shape that let fourteen earlier tests pass.

I also looked for the failure mode where a test *asserts the bug*: there is none. Grepping `goproxy.rs`, `rubygems.rs`, `hex.rs` and `pub_registry.rs` returns **zero** `Content-Encoding` assertions of any kind — the class is simply untested in those handlers, which is the subject of the finding below.

### Ranges / partial responses — not applicable

`download_root` takes no `HeaderMap`, never inspects `Range` and never emits `Content-Range` or `206`, so the range-vs-coding interaction cannot arise. A `HEAD` (routed by `get()`) now returns the declared coding alongside the coded length, which is the consistent answer and matches the stance taken in #3226/#3209.

---

## Finding — the residuals list is not exhaustive, and the `fetch_artifact` classification is inverted for one caller

**Severity: medium. Not a regression from this PR — nothing here changes behaviour — but the PR states a scope that is wider than what it verified, and the un-listed cases are live.**

The PR's new doc on `fetch_artifact` says it is *"kept for the callers that parse or rewrite the body themselves (virtual metadata merge closures, the pub registry JSON rewriter)"*, and the residuals section lists only `proxy_fetch_or_redirect` (no production caller) and `pub_registry`'s non-JSON fallback. That reads as exhaustive. It is not.

`proxy_fetch` — the only handler-facing wrapper over `fetch_artifact` — has four production call sites. Two are in `proxy_helpers.rs`: `collect_virtual_metadata` (`:2790`), which genuinely merges, and `resolve_virtual_metadata` (`:2711`), which does **not** transform anything itself. Its `transform` is caller-supplied, and **all four in-tree callers pass an identity pass-through** that forwards the buffered upstream bytes with a hardcoded `Content-Type` and no coding:

| Call site | Route | Shape |
|---|---|---|
| `backend/src/api/handlers/goproxy.rs:406-415` | `try_proxy_go_metadata`, virtual arm (`@v/list`, `.info`, `.mod`) | `Body::from(bytes)` + hardcoded CT |
| `backend/src/api/handlers/rubygems.rs:133-139` | `gem_info` — `GET /:repo_key/api/v1/gems/:name` | same |
| `backend/src/api/handlers/rubygems.rs:806-812` | `quick_spec` — `GET /:repo_key/quick/Marshal.4.8/:spec_file` | same; the code comment above it reads *"Otherwise proxy the upstream quick spec verbatim"* |
| `backend/src/api/handlers/hex.rs:412-420` | `package_info` | same |

Note also that `resolve_virtual_metadata`'s Pass-1 cached probe already receives the coding and discards it — `Ok(Some((bytes, _ct, _enc)))` at `backend/src/api/handlers/proxy_helpers.rs:2670` — so on that path the value is in hand and thrown away, not merely unavailable.

And a fifth of the same class one layer over: `goproxy.rs:376-393` (`try_proxy_go_metadata`, **Remote** arm) forwards verbatim through `proxy_fetch_capped` → `fetch_artifact_capped`, the known dropper — even though a `_with_encoding` sibling already exists at `proxy_service.rs:2716`.

Each of these is #3211's exact shape (verbatim bytes, recomputed length, no coding declared) in three formats other than Maven, reachable on exactly the same premise this PR argues for Maven: an object store or CDN returns a stored `Content-Encoding` regardless of the `identity` the proxy advertises.

The narrower sentence in the PR body — *"no other format's serving behavior changes in this PR"* — is accurate and I confirmed it. The problem is the sentence next to it, and the residuals section that implies the rest is already understood.

**Asks (neither blocks merge):**

1. Amend the residuals section, and correct the `fetch_artifact` doc — "virtual metadata merge closures" is true of `collect_virtual_metadata` and false of `resolve_virtual_metadata`, which is precisely the caller a future reader would trust that sentence about.
2. File a follow-up issue covering the five call sites above. There is no open issue for them today: #3149, #3184, #3193, #3210 and #2920 are all closed and none touches goproxy, rubygems or hex. Without one, the seam is documented as clean in the one place someone would look.

---

## Nits (non-blocking, no change requested)

* `forward_root_verbatim` `.unwrap()`s the builder and now feeds it **two** upstream-derived header values. It is safe: both are captured through `HeaderValue::to_str().ok()` (`proxy_service.rs:604`, `:2115`), which admits only visible ASCII, so `header()` cannot fail. A one-line comment saying so would save the next reader the trip.
* `Content-Encoding` is captured with `headers.get(..)`, so an upstream sending two separate `Content-Encoding` field lines yields only the first, and the re-declared value would understate the coding stack. Pre-existing at the primitive, not introduced here.
* An upstream `Content-Encoding: identity` would now be echoed; RFC 9110 §16.6 registers `identity` as reserved for use in `Accept-Encoding`. Cosmetic only.

## Credit

The parts that are easy to get wrong were got right: widening **in place** rather than adding a `_with_encoding` sibling (so the blast radius is a compile error, not a silent regression), verifying that `Content-Length` was already correct before claiming §8.6 was satisfied, a non-gzip fixture with a same-fixture uncoded control, decoding the served body instead of only asserting the header, and a revert-proof run with a real database rather than the silent `try_pool` skip.

Issue #3244, opened 2026-08-10
Add an admin endpoint to inspect/invalidate a proxy_scan_results verdict by digest (split from #3105 Bug 3)


brandonrc:
Split out from #3105 (Bug 3).

**Claim (confirmed at source):** there is no API route to list, get, or delete a `proxy_scan_results` row by `checksum_sha256`. Verified across backend/src/api/handlers/security.rs and admin.rs — the only related endpoint, `DELETE .../proxy-cache`, clears `proxy_cache_artifacts` (admin.rs), an unrelated table. So an operator with a stale/incorrect verdict has no non-DB recourse.

**Note:** #3105's Bug 2 fix (honor `scan_on_proxy` in the blob re-block, PR to follow) substantially relieves the acute pain here: once merged, setting `scan_on_proxy: false` unblocks a previously-flagged image at BOTH the manifest and blob levels, removing the 'permanently blocked short of a DB delete' situation. This issue remains for the general operator-tooling gap (viewing/clearing a specific verdict), which is a new admin endpoint = feature-shaped; scope it accordingly.

brandonrc:
# Decision brief: admin inspect/invalidate endpoint for `proxy_scan_results` (#3244)

Analysis only — no code changes. All `file:line` citations are against `main` @ `e91e970`.

## Premise check: confirmed

There is no route anywhere that reads or deletes a `proxy_scan_results` row by digest. The only readers/writers are the serve-path internals (`ProxyScanService::lookup_verdict` / `record_verdict`, `backend/src/services/proxy_scan_service.rs:259`, `:286`); every `DELETE FROM proxy_scan_results` in the tree is test fixture cleanup (e.g. `backend/src/api/handlers/oci_v2.rs:25534`, `pypi.rs:13400`). One phrasing correction: the cache-clear surface in the current tree is `POST /api/v1/repositories/{key}/cache/invalidate` (`backend/src/api/handlers/repositories.rs:647`, handler `:2016`), not a `DELETE .../proxy-cache`; either way it clears cached blobs + the `proxy_cache_artifacts` catalog row only (`backend/src/services/proxy_service.rs:4039-4055`) — and the verdict *deliberately* survives cache eviction (`proxy_scan_service.rs:9-12`).

## 1. What actually unblocks a blocked image today

| Path | Clears the block? | Latency / cost |
|---|---|---|
[...]

Issue #3243, opened 2026-08-10
Proxy/OCI inline scan gate ignores scan_configs.severity_threshold (split from #3105 Bug 1)


brandonrc:
Split out from #3105 (Bug 1) so #3105's live-reproduced blob re-block bug (Bug 2) can land on its own.

**Claim (confirmed at source):** the proxy/OCI inline scan-and-block gate blocks on ANY finding (`ProxyScanVerdict::is_vulnerable()` = `findings_count > 0` → `verdict = 'vulnerable'` → `verdict_blocks()`), and never consults `scan_configs.severity_threshold`. Verified: `decide_serve` / `gate_proxy_scan_serve` (backend/src/api/handlers/proxy_helpers.rs) and `verdict_blocks` (backend/src/services/proxy_scan_service.rs) take no severity input.

**Important nuance for whoever picks this up — this is a behavior change, not a pure bug fix, and it moves the gate toward fail-OPEN:**
- `severity_threshold` currently has NO production consumer *anywhere* — `ScanConfig::threshold()` (backend/src/models/security.rs) is referenced only by tests; hosted downloads use the separate `PolicyService`/`max_severity` path (#2912). Making the proxy gate the first consumer is new behavior.
- The column default is `'high'` (migrations/022_security_scanning.sql). Honoring it unconditionally would silently change the *default* posture: a repo on defaults that today blocks an image with only a low/medium finding would start SERVING it. That is a weakening of a security gate and is not appropriate for a silent patch.

**Requirements when implemented (must fail CLOSED):** a `vulnerable` verdict must still block whenever `max_severity` is unknown/NULL (the stored `proxy_scan_results.max_severity` can be absent on legacy rows) — serve only when `max_severity` is KNOWN and strictly below the configured threshold. Recommend shipping in a minor with an explicit release-note callout, not a patch.

Data is already present: `ProxyScanVerdict.max_severity` is computed and persisted to `proxy_scan_results.max_severity`.

brandonrc:
## Decision brief: disposition of `scan_configs.severity_threshold`

Investigated against `e91e9703a273dd23c2bc93eaad3a5abdc939bc24` (main). Analysis only — no code changes, no PR. Line references are to that SHA.

---

### 0. Premise check — mostly confirmed, with three corrections that matter

**Confirmed:**

* The column default *is* `'high'`, with a CHECK over `critical|high|medium|low|info` — `backend/migrations/022_security_scanning.sql:12-13`.
[...]

Issue #3023, opened 2026-07-30
Virtual repos bypass the inline scan-and-block gate (systemic, all formats): a vulnerable artifact blocked via a Remote repo serves 200-unscanned via a Virtual repo aggregating it


brandonrc:
## Summary
The inline scan-and-block-on-download gate (#2954/#2970 PyPI, #3004 npm, #3022 Docker/OCI) is wired **only into the Remote serve path**. A **Virtual** repo that aggregates a Remote member resolves and serves the member's bytes **before** the gate — so a known-vulnerable artifact that is correctly blocked (403) when pulled through the Remote repo is served **200 UNSCANNED** when pulled through a Virtual repo that has that Remote as a member.

**This is the standard recommended topology** — clients point at a single virtual repo that aggregates local + remote sources — so in typical use the scan-block is bypassed. Surfaced during the #3022 OCI re-red-team.

## Evidence
- **OCI — CONFIRMED live** (on the #3022 head): `GET /v2/dockremote/<img>/manifests/latest` → **403** 'blocked by scan policy'; the IDENTICAL image via `GET /v2/dockvirt/<img>/manifests/latest` (virtual aggregating dockremote) → **200 unscanned**; config+layer blobs via the virtual likewise 200.
- **Root cause:** `oci_v2.rs handle_get_manifest` — the `if repo.repo_type == Virtual { resolve_virtual_manifest → build_oci_proxy_response }` branch returns before `maybe_gate_remote_manifest_scan`, which itself early-returns `Ok(false)` for any `repo_type != Remote`. The blob-blocklist join keys on the virtual repo id (empty `manifest_blob_refs`) so virtual blobs also serve.
- **npm + PyPI — CONFIRMED by code (same architecture):** the gate is `if repo.repo_type == RepositoryType::Remote` (`npm.rs:2723`, `pypi.rs:1812/1838`); the Virtual serve branches return without the gate.

## Impact
HIGH. The flagship active-blocking control is bypassable across all three formats via the recommended virtual-repo topology — a false sense of protection for exactly the deployment shape evaluators use.

## Fix direction
[...]

PR #3245, opened 2026-08-10
fix(oci): honor scan_on_proxy in the blob-level scan re-block (#3105 Bug 2)

### LINKED ISSUE #3105: Inline scan-and-block on Docker/OCI proxy pulls issues
### Description

**Bug 1 — `severity_threshold` is never consulted by the proxy/OCI inline scan-and-block gate.** Any scan finding of any severity — including `low` or `info` — blocks a Docker/OCI proxy pull with a 403, with no way to configure a severity floor for this specific enforcement path, even though `PUT /api/v1/repositories/{key}/security` accepts and validates a `severity_threshold` field that reasonably implies it should apply here.

**Bug 2 — the blob-level re-block check is unconditional and ignores `scan_on_proxy` entirely**, unlike the manifest-level gate. Disabling scanning on a repository (`scan_on_proxy: false`) stops new manifest pulls from being gated, but a blob GET for an image that was **ever previously** scanned and flagged vulnerable — even once, even before scanning was disabled — is blocked forever regardless of current config, because `blob_belongs_to_vulnerable_image` checks only for a pre-existing `verdict = 'vulnerable'` row, with no reference to whether scanning is currently enabled.

**Bug 3 — there is no way, via the API, to inspect or invalidate a stored `proxy_scan_results` verdict.** Combined with Bug 2, this means an operator who disables scanning to unblock a pull has no recourse for an already-flagged image short of direct database access — there's no admin endpoint to list, view, or delete a verdict by digest, and the one existing cache-clear endpoint (`DELETE .../proxy-cache`) clears a different table (`proxy_cache_artifacts`) that has no effect on this block.

### Environment Information

- Deployment type: docker-compose.yml

### Backend Version

1.7.0

### Web UI Version

1.7.0

### Package Format

Docker / OCI
*(the same underlying gate is shared with npm and PyPI proxy pulls per `gate_proxy_scan_serve` / `proxy_scan_and_record` in `api/handlers/proxy_helpers.rs`, so this likely reproduces on those formats too)*

### Expected Behavior

- **Bug 1:** Setting `severity_threshold` to, e.g., `critical` should mean only `critical`-severity findings cause a Docker/OCI proxy pull to be blocked.
- **Bug 2:** Setting `scan_on_proxy: false` on a repository (and, for a virtual repo, on the virtual and all its Remote members — the current OR-of-both-sides semantics from #3025 are fine and not being disputed here) should mean NO pull through that repository is blocked by the scan gate anymore, including images that were previously flagged.
- **Bug 3:** An operator should have some way — even a manual admin action — to view or clear an incorrect/stale scan verdict for a specific digest without direct database access, the same way `DELETE .../proxy-cache` exists for the proxy content cache.

### Observed Behavior

**Bug 1:** A docker pull through a Remote/proxy (or Virtual repo resolving to a Remote member) Docker repository is blocked with:
```
pull of <image>:<tag> blocked by scan policy: vulnerabilities found
```
as soon as the scanner reports **any** finding, regardless of the repo's configured `severity_threshold`.

**Bug 2 (reproduced live):** After pulling an image once through a Virtual repo (containing both a Local and a Remote member) while scanning was enabled and the image was flagged vulnerable, we then set `scan_on_proxy: false` on every relevant repo (the virtual and the Remote member). The manifest pull is now correctly unblocked, but the blob GET for that same image still returns:
```
blob belongs to an image blocked by scan policy: vulnerabilities found
```
i.e. disabling scanning entirely does not undo a block recorded before the config change.

**Bug 3:** There is no endpoint to confirm this diagnosis directly (list/view `proxy_scan_results` by digest) or to clear it (the only related endpoint, proxy-cache clear, deletes `proxy_cache_artifacts`, an unrelated table).

Root cause (traced in the 1.7.0 source):

Bug 1 — `ProxyScanVerdict::is_vulnerable()` in `backend/src/services/scanner_service.rs`:
```rust
pub fn is_vulnerable(&self) -> bool {
    self.findings_count > 0
}
```
This is persisted as the digest-keyed verdict (`proxy_scan_results.verdict`) via `verdict_token()`, and `verdict_blocks()` in `services/proxy_scan_service.rs` blocks on any `"vulnerable"` verdict. Nowhere in `gate_oci_proxy_manifest_scan` → `maybe_gate_remote_manifest_scan` / `effective_virtual_scan_policy` (`api/handlers/oci_v2.rs`, `api/handlers/proxy_helpers.rs`) is `scan_configs.severity_threshold` read. The only fields consulted are `scan_on_proxy` (via `is_proxy_scan_enabled`) and `proxy_scan_action` (fail_open/fail_closed for pending scans).

Bug 2 — the blob GET handler in `api/handlers/oci_v2.rs` (the `#3003 round 2 (HIGH-3)` block, guarding against reconstructing a blocked image one blob at a time) calls `blob_belongs_to_vulnerable_image` / `blob_belongs_to_vulnerable_image_any` unconditionally:
```rust
if !oci_pull_is_scan_scoped(&claims) {
    let blocklist = /* ... */ blob_belongs_to_vulnerable_image(state, repo.id, &lookup_digest).await;
    match blocklist {
        Ok(true) => { return oci_error(StatusCode::FORBIDDEN, "DENIED",
            "blob belongs to an image blocked by scan policy: vulnerabilities found"); }
        ...
```
This is a pure `EXISTS` query against `proxy_scan_results` joined through `manifest_blob_refs`, with no `is_proxy_scan_enabled` check and no freshness/TTL check (unlike the manifest-level `verdict_is_reusable` path, which at least considers TTL and scanner-DB-version freshness for reuse). The manifest gate and the blob gate are inconsistent with each other: one respects `scan_on_proxy`, the other does not.

Bug 3 — confirmed by reading `api/handlers/security.rs` and `api/handlers/admin.rs` in full: no route exists to list, get, or delete a `proxy_scan_results` row by `checksum_sha256`. The table (`migrations/181_proxy_scan_results.sql`) also has no TTL/expiry column beyond what `verdict_is_fresh` computes at *reuse* time in the manifest path — the blob path doesn't even consult that.

### Steps to Reproduce

**Bug 1:**
1. Create a Remote (proxy) Docker/OCI repository pointed at a public upstream.
2. `PUT /api/v1/repositories/{key}/security` with
   `{"scan_on_proxy": true, "severity_threshold": "critical"}`.
3. `docker pull` an image through the repo with a known `low`/`medium` finding
   and no `critical` finding.
4. Pull is blocked despite `severity_threshold: "critical"`.

**Bug 2 + 3:**
1. Same setup as above, but as a Virtual repo containing a Local repo and a Remote repo as members.
2. `docker pull` the image once through the virtual repo — it gets scanned, flagged vulnerable, and blocked (manifest-level).
3. Disable scanning everywhere it could apply:
   `PUT /api/v1/repositories/{virtual-key}/security {"scan_on_proxy": false}`
   `PUT /api/v1/repositories/{remote-member-key}/security {"scan_on_proxy": false}`
4. `docker pull` the same image/tag again.
5. The manifest now resolves (Bug 1's gate is skipped, as expected), but the blob GET for the same image still returns 403 `"blob belongs to an image blocked by scan policy: vulnerabilities found"`.
6. There is no API call available to confirm or clear the stale verdict causing this (Bug 3) — the only recourse is a direct `DELETE FROM proxy_scan_results WHERE checksum_sha256 = '<manifest digest>';` against the database.

### Additional Context

- Relevant source (main branch, 1.7.0 tag):
  - `backend/src/services/scanner_service.rs` — `ProxyScanVerdict::is_vulnerable()`
  - `backend/src/services/proxy_scan_service.rs` — `verdict_blocks()`, `verdict_is_fresh()`, `verdict_is_reusable()`
  - `backend/src/api/handlers/oci_v2.rs` — `maybe_gate_remote_manifest_scan`, `gate_oci_proxy_manifest_scan`, `oci_scan_deny_response`, `blob_belongs_to_vulnerable_image`, `blob_belongs_to_vulnerable_image_any`
  - `backend/src/api/handlers/proxy_helpers.rs` — `gate_proxy_scan_serve`, `proxy_scan_and_record`, effective_virtual_scan_policy`, `stricter_scan_policy`
  - `backend/src/services/scan_config_service.rs` — confirms `severity_threshold` is stored/validated but never read by `is_proxy_scan_enabled` / `proxy_scan_action`
  - `backend/migrations/181_proxy_scan_results.sql` — no delete/admin path, global content-addressed table shared across all repos/tenants
- Related issues: #2954 (original design goal — "per the repo's scan config + security policy" — contradicted by Bug 1 for the proxy path specifically), #3003 (Phase 2, npm/Docker parity; no severity-threshold behavior specified or implemented), #3023/#3025 (virtual-repo gating — working as designed, not disputing this), #2976 (a related but different staleness direction:
  stale-*clean* verdicts after a CVE-DB bump, not stale-*vulnerable* verdicts with no invalidation path at all), #2912 (the older, separate `PolicyService`/`max_severity` mechanism used by hosted downloads, which does not share this gap).
- Workaround in use: disabling `scan_on_proxy` on both the virtual repo and every Remote member stops *new* blocks, but does not clear an already-recorded vulnerable verdict (Bug 2) — that image tag remains permanently blocked short of a direct DB delete.

### Pre-submission Checklist

brandonrc:
## Summary

`#3105` reports three distinct problems in inline scan-and-block on Docker/OCI proxy pulls. This PR resolves the one that is a live-reproduced, fail-closed-safe patch bug (**Bug 2**), and splits the other two — which are behavior changes / new surface, not patch bug fixes — into dedicated tracking issues so they get the review they warrant instead of riding along.

## Enumeration of the three reported bugs

| Bug | Verdict | Disposition |
|---|---|---|
| **Bug 2** — blob-level re-block ignores `scan_on_proxy` | **Confirmed, fixed here** | This PR |
| **Bug 1** — proxy gate ignores `severity_threshold` | **Confirmed at source, deferred** | → #3243 |
| **Bug 3** — no admin endpoint to inspect/clear a verdict | **Confirmed at source, deferred** | → #3244 |

### Bug 2 (fixed) — the blob re-block must honor `scan_on_proxy`

The #3003-round-2 blob re-block (a config/layer blob belonging to an image with a stored `vulnerable` proxy-scan verdict must not serve, so a blocked image can't be reconstructed one blob at a time) ran **unconditionally** in `handle_get_blob` — it only checked `oci_pull_is_scan_scoped`, never `is_proxy_scan_enabled`.

The **manifest** gate (`maybe_gate_remote_manifest_scan`) already short-circuits when `is_proxy_scan_enabled` is false. So the two halves of the same gate disagreed: after an operator set `scan_on_proxy: false` to unblock a previously-flagged image, the manifest served again but every blob `GET` stayed `403 DENIED "blob belongs to an image blocked by scan policy"` **forever** — the image was un-pullable but not "blocked", with no recourse short of `DELETE FROM proxy_scan_results` against the database. Reproduced live in the issue.

**Fix:** gate the blob re-block on the current `scan_on_proxy` of the repository that **owns the refs**, mirroring the manifest gate. For a Virtual repo the #3023/#3025 OR-of-both-sides is preserved: `scan_on_proxy` on the virtual itself keeps the gate on across the whole member set, otherwise each member's refs are gated by that member's own policy. The member set is resolved once and reused for both the enabled-check and the blocklist span.

See **Scope of the gate** below for why "owns the refs" and not "this request" — that distinction is what keeps hosted repositories on their pre-#3105 behaviour.

**Fails closed, by construction:**
- When scanning is **on** (the security-relevant state), the blocklist runs exactly as before — no weakening.
- A config-read fault is treated as "assume enabled" (`unwrap_or(true)`), so a DB blip keeps the block. `is_proxy_scan_enabled` returns `Ok(false)` only for an absent row, `Err` only on a real DB error.
- Virtual member-enumeration failure returns `423` (unchanged posture).

### Scope of the gate — `scan_on_proxy` is applied to proxy-served refs only

`scan_on_proxy` is a **proxy** knob. Every other consumer reads it on a proxy serve path, and the manifest half of this very gate (`maybe_gate_remote_manifest_scan`) is itself `Remote`-only. Critically, `ScanConfigService::is_proxy_scan_enabled` returns `Ok(false)` for an **absent** `scan_configs` row, and hosted (`Local`/`Staging`) repositories normally have no such row and no workflow that creates one.

So an earlier revision of this PR, which applied the flag as one whole-request boolean, did not "unblock a proxy pull" on a hosted repository — it switched the #3003 blob re-block **off by default for every hosted OCI repository on the instance**. That is the only blob-level scan enforcement those repositories have: `maybe_gate_remote_manifest_scan` bails on non-`Remote`, `handle_get_blob` has no quarantine seam, and #3235's `enforce_scan_policy_gate` hangs off `artifacts` rows while OCI blobs resolve from `oci_blobs`. It was proved live on one tree and one database, with only the production hunk differing: merge-base **403**, that revision **200**. Reachable via the reporter's own topology, because the verdict join is deliberately global by content digest (no `psr.repository_id` predicate) — an image graded vulnerable through a proxy also blocks the copy that lives in a hosted repo.

The predicate is therefore evaluated per **repository that owns the refs** (`blob_reblock_applies`):

| Repo type | Blob re-block | Rationale |
|---|---|---|
| `Local` / `Staging` | always applies | `scan_on_proxy` says nothing about hosted content; an absent row must not read as "disabled". Identical to pre-#3105 behaviour, and costs **no** extra query on the hosted hot path. |
| `Remote` | honors `scan_on_proxy` | This is the path #3105 Bug 2 is about. |
| `Virtual` | virtual's own flag ORs the whole member set; otherwise per-member by the rows above | #3025 stricter-of-two, and it stops a virtual with proxy scanning off from becoming a bypass route around a hosted member's enforcement. |

**#3105's reported topology still resolves.** A virtual containing a Local *and* a Remote member: the image was proxied, so its `manifest_blob_refs` are recorded under the **Remote** member. Setting `scan_on_proxy: false` on the virtual and that member drops those refs out of the blocklist, and the blob serves — which is exactly what the issue asked for. What does *not* happen any more is the same setting silently disabling enforcement on hosted content the operator never mentioned.

Two adjacent gaps found during review were filed rather than folded in: #3258 (`handle_head_blob` has no re-block at all — pre-existing #3003 scope) and #3259 (the blob blocklist has no freshness/TTL check, unlike the manifest path's `verdict_is_reusable`).

### Bug 1 (deferred → #3243)

Confirmed: the proxy gate blocks on any finding and never reads `severity_threshold`. But `severity_threshold` has **no production consumer anywhere** today (`ScanConfig::threshold()` is test-only; hosted downloads use the separate `PolicyService`/`max_severity` path, #2912), and its column default is `'high'`. Honoring it unconditionally would silently **weaken** the gate and change the default posture (repos on defaults would stop blocking low/medium findings). That is a fail-open-direction behavior change, not a patch bug fix, and must itself fail closed on unknown `max_severity`. Tracked in #3243.

### Bug 3 (deferred → #3244)

Confirmed: no route lists/gets/deletes a `proxy_scan_results` row by digest; the existing `DELETE .../proxy-cache` clears the unrelated `proxy_cache_artifacts` table. This is a new admin endpoint (feature-shaped). Its acute motivation is largely removed by the Bug 2 fix — disabling `scan_on_proxy` now actually unblocks a flagged image at both levels. Tracked in #3244.

## Tests

All DB-backed `--lib` tests driving the real `handle_get_blob` through the router, run with `DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` exported (0.3–0.6 s each, so they genuinely executed rather than early-returning).

**The fix (#3105 Bug 2):**
- `test_blob_reblock_honors_scan_on_proxy` — Remote repo. `scan_on_proxy` **enabled** → vulnerable image's blob `403` (control) **and** a clean image's blob `200` (control); `scan_on_proxy` **disabled** → the previously-flagged blob serves `200`.
- `blob_via_virtual_all_scanning_off_unblocks` — the same claim through a virtual, with the block observed **first** while the member has scanning on, so it cannot be satisfied by a fixture that never blocked.

**The scope (no hosted regression):**
- `test_hosted_blob_reblock_ignores_absent_scan_config` — hosted repo, **no `scan_configs` row at all** (asserted as a fixture precondition). Vulnerable image's config and layer blobs `403`; a clean image's blob in the same fixture `200`.
- `blob_via_virtual_still_blocks_hosted_member_refs_with_scanning_off` — a hosted member's refs stay blocked through a virtual that has proxy scanning off, with the clean sibling as the positive control.

**The virtual OR, previously untested in both directions:**
- `blob_via_virtual_enabled_blocks_when_member_scanning_off` — the virtual's flag holds the block on across the member set (stricter-of-two).
- `blob_via_virtual_all_scanning_off_unblocks` — the unblock direction, above.

The four virtual blob-seam cases share a `VirtualBlobFixture` so the ~90 lines of staging exist once rather than four times.

**Test-isolation fix.** `proxy_scan_results` is keyed on the content digest alone and `cleanup_proxy_scan_row` deletes by that key globally, so three tests built from identical `image_manifest(b"bad-cfg", b"bad-layer")` bytes shared one verdict row and raced under nextest — measured at 3 failures in 9 runs, always presenting as "the gate did not block", the worst possible shape for a flake on a security gate. All fixture bytes in both modules are now UUID-derived (`unique_fixture_bytes`), which also de-collides the pre-existing pair. **12/12 green** on the same selector after the change; full `oci_v2` suite 530/530 green on three consecutive runs.

**Revert-proof 1 — the fix (faithful merge-base revert).** Production code reverted to the exact merge-base form, all tests kept:

```
test_blob_reblock_honors_scan_on_proxy ... FAILED
  #3105 Bug 2: with scan_on_proxy disabled, the previously-flagged blob must serve again
    left: 403   right: 200

blob_via_virtual_all_scanning_off_unblocks ... FAILED
  #3105 Bug 2 through a virtual: scanning off on the virtual AND its proxy member must unblock the blob
    left: 403   right: 200
```

Both fail on their **target** assertion; every positive control still passes, and the two hosted guards pass (they pin behaviour the merge base already had, which is the point of them). Faithfulness on the reverted tree: `git diff <merge-base> --shortstat` → `1 file changed, 617 insertions(+), 93 deletions(-)`. Every hunk is inside `mod proxy_scan_block_tests` / `mod virtual_scan_gate_tests`; the deletions are the test-only refactor (shared `VirtualBlobFixture`, UUID-derived bytes). Verified directly rather than inferred: stripping every `#[cfg(test)] mod … { }` block from both files leaves **63,459 bytes each, byte-identical** — production code on the reverted tree is exactly the merge base, so the failures can only come from the claim under test.

**Revert-proof 2 — the scope.** Production restored to this PR's earlier whole-request-boolean form, tests kept:

```
test_hosted_blob_reblock_ignores_absent_scan_config ... FAILED
  hosted: the CONFIG blob of a vulnerable image must stay blocked with no scan_configs row
    left: 200   right: 403

blob_via_virtual_still_blocks_hosted_member_refs_with_scanning_off ... FAILED
  a hosted member's vulnerable blob must stay blocked through the virtual
    left: 200   right: 403
```

That reproduces the reviewer's live 403 → 200 result as a test, and adds the virtual-over-hosted bypass. Both #3105 guards still pass on that tree, so the two experiments are independent.

`cargo fmt --all --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `SQLX_OFFLINE=true cargo check --workspace --all-targets` clean. No `sqlx::query!` text changed, so no `cargo sqlx prepare`; no migration.

Fixes #3105

### DIFF
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -4616,51 +4616,97 @@ async fn handle_get_blob(
     // fetch. Scanner-scoped pulls are exempt for the same reason the manifest
     // gate exempts them.
     if !oci_pull_is_scan_scoped(&claims) {
+        let scan_cfg =
+            crate::services::scan_config_service::ScanConfigService::new(state.db.clone());
         // #3023: for a Virtual repo the `manifest_blob_refs` are recorded under
-        // the resolving MEMBER (the virtual owns no refs), so the blocklist must
-        // span the virtual's member set. A direct Remote/Local repo keeps the
-        // single-id check verbatim. If the members cannot be enumerated we cannot
-        // prove the blob is safe -> fail closed (same posture as an unreadable
-        // verdict store below).
-        let blocklist = if repo.repo_type == RepositoryType::Virtual {
+        // the resolving MEMBER (the virtual owns no refs), so BOTH the
+        // scan-enabled check and the blocklist span the virtual's member set. A
+        // direct Remote/Local repo keeps the single-id check. If the members
+        // cannot be enumerated we cannot prove the blob is safe -> fail closed
+        // (same posture as an unreadable verdict store below).
+        let member_ids: Option<Vec<Uuid>> = if repo.repo_type == RepositoryType::Virtual {
             match proxy_helpers::fetch_virtual_members(&state.db, repo.id).await {
-                Ok(members) => {
-                    let ids: Vec<Uuid> = members.iter().map(|m| m.id).collect();
-                    blob_belongs_to_vulnerable_image_any(state, &ids, &lookup_digest).await
+                Ok(members) => Some(members.iter().map(|m| m.id).collect()),
+                Err(_) => {
+                    tracing::warn!(
+                        repo = %repo.key, digest = %lookup_digest,
+                        "virtual member enumeration failed during blob scan-verdict check; withholding"
+                    );
+                    return oci_error(
+                        StatusCode::LOCKED,
+                        "DENIED",
+                        "blob scan status is unavailable; retry shortly",
+                    );
                 }
-                Err(_) => Err(sqlx::Error::Protocol(
-                    "virtual member enumeration failed during blob scan-verdict check".to_string(),
-                )),
             }
         } else {
-            blob_belongs_to_vulnerable_image(state, repo.id, &lookup_digest).await
+            None
         };
-        match blocklist {
-            Ok(true) => {
-                tracing::warn!(
-                    repo = %repo.key, digest = %lookup_digest,
-                    "blocking blob pull: blob belongs to an image with a vulnerable scan verdict"
-                );
-                return oci_error(
-                    StatusCode::FORBIDDEN,
-                    "DENIED",
-                    "blob belongs to an image blocked by scan policy: vulnerabilities found",
-                );
+
+        // #3105 (Bug 2): the blob re-block must honor `scan_on_proxy`, exactly as
+        // the manifest gate does (`maybe_gate_remote_manifest_scan` short-circuits
+        // when `is_proxy_scan_enabled` is false). Before this, the EXISTS check
+        // ran unconditionally, so a blob belonging to an image flagged vulnerable
+        // ONCE stayed 403 forever even after an operator set `scan_on_proxy: false`
+        // -- while the manifest for the same image served again, leaving the image
+        // un-pullable but not "blocked" with no non-DB recourse. For a Virtual
+        // repo the effective policy is enabled iff the virtual OR ANY member
+        // enables it (#3023/#3025 OR-of-both-sides), matching
+        // `effective_virtual_scan_policy`. Fails CLOSED on a config-read fault:
+        // `is_proxy_scan_enabled` returns Ok(false) only for an absent row and Err
+        // for a real DB error, so `unwrap_or(true)` keeps the block on a fault.
+        let scan_active = match &member_ids {
+            Some(ids) => {
+                let mut active = scan_cfg
+                    .is_proxy_scan_enabled(repo.id)
+                    .await
+                    .unwrap_or(true);
+                for id in ids {
+                    if active {
+                        break;
+                    }
+                    active = scan_cfg.is_proxy_scan_enabled(*id).await.unwrap_or(true);
+                }
+                active
             }
-            Ok(false) => {}
-            // Fail CLOSED on a control-plane error: we cannot show this blob is
-            // safe to serve, and this is the same posture the manifest gate
-            // takes for an unreadable verdict store.
-            Err(e) => {
-                tracing::warn!(
-                    repo = %repo.key, digest = %lookup_digest, error = %e,
-                    "blob scan-verdict lookup failed; withholding rather than serving"
-                );
-                return oci_error(
-                    StatusCode::LOCKED,
-                    "DENIED",
-                    "blob scan status is unavailable; retry shortly",
-                );
+            None => scan_cfg
+                .is_proxy_scan_enabled(repo.id)
+                .await
+                .unwrap_or(true),
+        };
+
+        if scan_active {
+            let blocklist = match &member_ids {
+                Some(ids) => blob_belongs_to_vulnerable_image_any(state, ids, &lookup_digest).await,
+                None => blob_belongs_to_vulnerable_image(state, repo.id, &lookup_digest).await,
+            };
+            match blocklist {
+                Ok(true) => {
+                    tracing::warn!(
+                        repo = %repo.key, digest = %lookup_digest,
+                        "blocking blob pull: blob belongs to an image with a vulnerable scan verdict"
+                    );
+                    return oci_error(
+                        StatusCode::FORBIDDEN,
+                        "DENIED",
+                        "blob belongs to an image blocked by scan policy: vulnerabilities found",
+                    );
+                }
+                Ok(false) => {}
+                // Fail CLOSED on a control-plane error: we cannot show this blob
+                // is safe to serve, and this is the same posture the manifest gate
+                // takes for an unreadable verdict store.
+                Err(e) => {
+                    tracing::warn!(
+                        repo = %repo.key, digest = %lookup_digest, error = %e,
+                        "blob scan-verdict lookup failed; withholding rather than serving"
+                    );
+                    return oci_error(
+                        StatusCode::LOCKED,
+                        "DENIED",
+                        "blob scan status is unavailable; retry shortly",
+                    );
+                }
             }
         }
     }
@@ -26037,6 +26083,10 @@ mod proxy_scan_block_tests {
             .execute(&fx.pool)
             .await
             .expect("make repo public");
+        // #3105 (Bug 2): the blob re-block now honors `scan_on_proxy`, so the
+        // block only applies while scanning is enabled. Enable it (as production
+        // always is when a `vulnerable` verdict could have been recorded).
+        enable_proxy_scan(&fx.pool, fx.repo_id, "fail_open").await;
 
         let (vuln_manifest, vuln_config, vuln_layer) = image_manifest(b"bad-cfg", b"bad-layer");
         let (clean_manifest, clean_config, clean_layer) =
@@ -26164,6 +26214,138 @@ mod proxy_scan_block_tests {
             clean_layer_status,
             StatusCode::OK,
             "a clean image's layer blob must still serve (no over-block)"
+        );
+    }
+
+    /// #3105 (Bug 2): the blob-level re-block MUST honor `scan_on_proxy`, the
+    /// same way the manifest gate does. A blob belonging to an image with a
+    /// stored `vulnerable` verdict is re-blocked ONLY while scanning is enabled;
+    /// after an operator sets `scan_on_proxy: false` the block is lifted, so
+    /// disabling scanning is not left as a half-measure that unblocks the
+    /// manifest but strands the image's blobs at 403 forever.
+    ///
+    /// The same fixture carries the two positive controls the negative assertion
+    /// needs (per the pipeline's "a negative assertion needs a positive control"
+    /// rule): with scanning ENABLED the vulnerable blob is 403 AND a clean blob
+    /// is 200, so a fix that blocked everything (or nothing) cannot pass.
+    #[tokio::test]
+    async fn test_blob_reblock_honors_scan_on_proxy() {
+        let Some(fx) = tdh::Fixture::setup("remote", "docker").await else {
+            return;
+        };
+        sqlx::query("UPDATE repositories SET is_public = true WHERE id = $1")
+            .bind(fx.repo_id)
+            .execute(&fx.pool)
+            .await
+            .expect("make repo public");
+
+        let (vuln_manifest, vuln_config, vuln_layer) = image_manifest(b"bad-cfg", b"bad-layer");
+        let (clean_manifest, clean_config, clean_layer) = image_manifest(b"ok-cfg", b"ok-layer");
+        let vuln_digest = compute_sha256(&vuln_manifest);
+        let clean_digest = compute_sha256(&clean_manifest);
+
+        for (mdigest, body) in [
+            (&vuln_digest, &vuln_manifest),
+            (&clean_digest, &clean_manifest),
+        ] {
+            record_manifest_blob_refs(&fx.pool, fx.repo_id, mdigest, body)
+                .await
+                .expect("record blob refs");
+        }
+        let storage = fx
+            .state
+            .storage_for_repo(&crate::storage::StorageLocation {
+                backend: "filesystem".to_string(),
+                path: fx.storage_dir.to_string_lossy().into_owned(),
+            })
+            .expect("storage");
+        for (d, bytes) in [
+            (&vuln_config, b"bad-cfg".to_vec()),
+            (&vuln_layer, b"bad-layer".to_vec()),
+            (&clean_config, b"ok-cfg".to_vec()),
+            (&clean_layer, b"ok-layer".to_vec()),
+        ] {
+            let key = blob_storage_key(d);
+            storage
+                .put(&key, Bytes::from(bytes.clone()))
+                .await
+                .expect("put blob");
+            sqlx::query(
+                "INSERT INTO oci_blobs (repository_id, digest, size_bytes, storage_key) \
+                 VALUES ($1, $2, $3, $4)",
+            )
+            .bind(fx.repo_id)
+            .bind(d)
+            .bind(bytes.len() as i64)
+            .bind(&key)
+            .execute(&fx.pool)
+            .await
+            .expect("insert oci_blobs");
+        }
+
+        // Only the first image is vulnerable.
+        ProxyScanService::new(fx.pool.clone())
+            .record_verdict(
+                vuln_digest.strip_prefix("sha256:").unwrap(),
+                "grype",
+                "vulnerable",
+                2,
+                1,
+                1,
+                0,
+                0,
+                Some("critical"),
+                Some("grype-1.0.0-test"),
+                Some(fx.repo_id),
+            )
+            .await
+            .expect("seed vulnerable verdict");
+
+        let get_blob = |d: String| {
+            let state = fx.state.clone();
+            let key = fx.repo_key.clone();
+            async move {
+                let app = tdh::router_anon(router(), state);
+                let req = Request::builder()
+                    .method("GET")
+                    .uri(format!("/{key}/app/blobs/{d}"))
+                    .header(AUTHORIZATION, format!("Bearer {ANONYMOUS_TOKEN}"))
+                    .body(Body::empty())
+                    .unwrap();
+                app.oneshot(req).await.expect("oneshot").status()
+            }
+        };
+
+        // ── scan_on_proxy ENABLED: the re-block applies (positive controls). ──
+        enable_proxy_scan(&fx.pool, fx.repo_id, "fail_open").await;
+        let vuln_when_enabled = get_blob(vuln_layer.clone()).await;
+        let clean_when_enabled = get_blob(clean_layer.clone()).await;
+
+        // ── scan_on_proxy DISABLED: the re-block must NOT apply anymore. ──
+        sqlx::query("UPDATE scan_configs SET scan_on_proxy = false WHERE repository_id = $1")
+            .bind(fx.repo_id)
+            .execute(&fx.pool)
+            .await
+            .expect("disable scan_on_proxy");
+        let vuln_when_disabled = get_blob(vuln_layer.clone()).await;
+
+        cleanup_proxy_scan_row(&fx.pool, vuln_digest.strip_prefix("sha256:").unwrap()).await;
+        fx.teardown().await;
+
+        assert_eq!(
+            vuln_when_enabled,
+            StatusCode::FORBIDDEN,
+            "positive control: with scan_on_proxy enabled, a vulnerable image's blob stays blocked"
+        );
+        assert_eq!(
+            clean_when_enabled,
+            StatusCode::OK,
+            "positive control: a clean image's blob serves even while scanning is enabled"
+        );
+        assert_eq!(
+            vuln_when_disabled,
+            StatusCode::OK,
+            "#3105 Bug 2: with scan_on_proxy disabled, the previously-flagged blob must serve again"
         );
     }
 
@@ -26769,6 +26951,10 @@ mod virtual_scan_gate_tests {
             .expect("member storage");
         let (virt_id, virt_key) = insert_virtual(&pool).await;
         link_member(&pool, virt_id, member_id).await;
+        // #3105 (Bug 2): the blob re-block now honors `scan_on_proxy`; for a
+        // virtual it applies iff the virtual OR any member enables scanning.
+        // Enable it on the member (its refs carry the vulnerable image).
+        enable_proxy_scan(&pool, member_id, "fail_open").await;
 
         let (vuln_manifest, vuln_cfg, _vuln_layer) = image_manifest(b"bad-cfg", b"bad-layer");
         let (clean_manifest, clean_cfg, _clean_layer) = image_manifest(b"ok-cfg", b"ok-layer");
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -4618,15 +4618,28 @@ async fn handle_get_blob(
     if !oci_pull_is_scan_scoped(&claims) {
         let scan_cfg =
             crate::services::scan_config_service::ScanConfigService::new(state.db.clone());
+        // #3105 (Bug 2): the blob re-block must honor `scan_on_proxy` for
+        // PROXY-served content. Before this, the EXISTS check ran
+        // unconditionally, so a blob belonging to an image flagged vulnerable
+        // ONCE stayed 403 forever even after an operator set
+        // `scan_on_proxy: false` -- while the manifest for the same image served
+        // again (`maybe_gate_remote_manifest_scan` short-circuits on the same
+        // flag), leaving the image un-pullable but not "blocked", with no
+        // non-DB recourse.
+        //
+        // The knob is applied per REPOSITORY THAT OWNS THE REFS, not as one
+        // whole-request boolean -- see `blob_reblock_applies` for why hosted
+        // refs must NOT be gated on it.
+        //
         // #3023: for a Virtual repo the `manifest_blob_refs` are recorded under
-        // the resolving MEMBER (the virtual owns no refs), so BOTH the
-        // scan-enabled check and the blocklist span the virtual's member set. A
-        // direct Remote/Local repo keeps the single-id check. If the members
-        // cannot be enumerated we cannot prove the blob is safe -> fail closed
-        // (same posture as an unreadable verdict store below).
-        let member_ids: Option<Vec<Uuid>> = if repo.repo_type == RepositoryType::Virtual {
-            match proxy_helpers::fetch_virtual_members(&state.db, repo.id).await {
-                Ok(members) => Some(members.iter().map(|m| m.id).collect()),
+        // the resolving MEMBER (the virtual owns no refs), so the blocklist
+        // spans the virtual's member set. A direct repo keeps the single-id
+        // check. If the members cannot be enumerated we cannot prove the blob
+        // is safe -> fail closed (same posture as an unreadable verdict store
+        // below).
+        let blocklist = if repo.repo_type == RepositoryType::Virtual {
+            let members = match proxy_helpers::fetch_virtual_members(&state.db, repo.id).await {
+                Ok(members) => members,
                 Err(_) => {
                     tracing::warn!(
                         repo = %repo.key, digest = %lookup_digest,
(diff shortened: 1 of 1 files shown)


brandonrc:
## Independent validation review — PR #3245 @ `308adb8`

Reviewed adversarially (assume-wrong-until-proven) from a fresh clone at `308adb8b2a6073f794d90a78a1725a0086b791c9`, diffed against merge-base `6c8a774`, with a local Postgres 16 and `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1` exported (DB tests took 0.3–0.8 s each, so they genuinely executed rather than early-returning).

**Verdict: not ready to merge as-is. Two blockers, both fixable in this PR. The core fix is correct and the new test is a genuine guard — I revert-proved it.**

| # | Blocker | Severity |
|---|---|---|
| B1 | The re-block also goes inert on **hosted (Local) OCI repos**, which `scan_on_proxy` does not describe. Live-proved: base 403 → head 200. | Medium — undeclared scope change on a security gate |
| B2 | The new test shares a content-addressed verdict digest with two existing tests → **reproducible CI flake** (3 failures in 9 local runs; base was 0/5). | Medium — introduces a red that teaches people to ignore red |

---

## B1 (read this first) — the unblock is wider than the PR claims: hosted repos lose the blob re-block entirely

`is_proxy_scan_enabled` returns `Ok(false)` for an **absent** `scan_configs` row (`backend/src/services/scan_config_service.rs:231-241`). Hosted (Local) repositories normally have no such row, and `scan_on_proxy` is a proxy-only knob — before this PR it had **no** consumer on any hosted path (`is_proxy_scan_enabled` / `effective_virtual_scan_policy` call sites are all npm/pypi/oci *proxy* serve paths). This PR is the first place a Local repo's behaviour depends on it, and the value is false by default with no workflow that would set it. Net effect: on hosted repos the `#3003` blob re-block is now unconditionally off.

The manifest half does not compensate: `maybe_gate_remote_manifest_scan` bails on `repo.repo_type != RepositoryType::Remote` (`backend/src/api/handlers/oci_v2.rs:7863`) — so hosted repos have no proxy-verdict block at either seam now. And `handle_get_blob` has no quarantine check either (`oci_manifest_quarantine_block` is manifest-only), so there is now **zero** blob-level scan/quarantine enforcement on hosted OCI. #3235's `enforce_scan_policy_gate` does not cover this route — it hangs off `local_lookup_artifact` on `artifacts` rows (`proxy_helpers.rs:3192-3213`), while OCI blobs resolve from `oci_blobs` in `handle_get_blob`. (Answering the "do #3235 and #3245 double-block or mutually bypass?" question: **neither** — they are disjoint mechanisms on disjoint code paths.)

**Live proof, not reasoning.** I added a throwaway probe (hosted `local`/`docker` fixture, public, one image with `manifest_blob_refs` + `oci_blobs` + bytes on disk, one `vulnerable` grype verdict on its manifest digest, no `scan_configs` row at all) and did an anonymous `GET /<repo>/app/blobs/<layer>`:

```
merge-base 6c8a774 :  LOCAL-REPO PROBE RESULT: status=403 Forbidden
PR head    308adb8 :  LOCAL-REPO PROBE RESULT: status=200 OK
```

Same tree, same database, only the production hunk differing.

**Concrete scenario** (it is the reporter's own topology in #3105 — a virtual with a Local *and* a Remote member): image `X` is pulled through the Remote proxy with `scan_on_proxy: true`, is graded `vulnerable`, and its digest-keyed verdict lands in `proxy_scan_results`. The same image also lives in a hosted repo (pushed, promoted, or mirrored) so `manifest_blob_refs` rows exist under the Local repo id. The verdict join is deliberately global by content digest — no `psr.repository_id` predicate (`oci_v2.rs:7475-7498`), and `test_blob_block_scoping_shared_digest_vs_shared_layer` pins that as intended. Before this PR, `GET /<local-repo>/v2/.../blobs/<layer>` → 403; after, → 200, and every config and layer of the vulnerable image is retrievable by digest. That is the reconstruct-the-image-one-blob-at-a-time hole #3003 round 2 was written to close, reopened on the hosted side by a knob the operator never touched.

I do not think this is *necessarily* wrong — hosted content is meant to be governed by scan-on-upload + quarantine, and #3105 is entirely about proxy pulls. But it is a decision, it is not stated anywhere in the PR, no claim in the description covers it, and no test pins it. Options:

1. Scope the new gate the way the manifest gate is scoped — apply the `scan_on_proxy` short-circuit only for `Remote`/`Virtual`, leaving `Local` on the pre-#3245 behaviour. Smallest diff, keeps the security posture identical for hosted, but leaves hosted repos with the same "flagged once, blocked forever, no API recourse" complaint that #3244 tracks.
2. Keep the current behaviour deliberately: say so in the PR body and in the code comment (which today says "exactly as the manifest gate does" — the manifest gate is Remote-only, so that sentence is not accurate as written), add a hosted-repo test pinning `200`, and file a follow-up for blob-level enforcement on hosted OCI, since `handle_get_blob` has no quarantine seam at all.

My recommendation is (2) plus the follow-up, because (1) recreates the no-recourse problem on the hosted side. Either way it needs to be an explicit, tested decision rather than a side effect.

---

## B2 — the new test is flaky: it collides with two existing tests on a shared content-addressed digest

`test_blob_reblock_honors_scan_on_proxy` builds its fixture with `image_manifest(b"bad-cfg", b"bad-layer")` (`oci_v2.rs:26242`). Two other tests use the *same* bytes through byte-identical `image_manifest` helpers:

* `proxy_scan_block_tests::test_get_blob_blocked_for_vulnerable_image_but_not_clean_one` (`oci_v2.rs:26091`)
* `virtual_scan_gate_tests::blob_via_virtual_blocks_vulnerable_but_not_clean` (`oci_v2.rs:26959`)

`proxy_scan_results` is keyed on `checksum_sha256` alone and `cleanup_proxy_scan_row` deletes by that key globally (`oci_v2.rs:25017-25023`). nextest runs these concurrently against one database, so one test's teardown deletes the verdict row another is mid-assertion on. The new test widens the window considerably — it holds the row across three HTTP round-trips plus a `scan_configs` UPDATE — and adds a second deleter.

Reproduced, not theorised:

```
head 308adb8, -E 'test(proxy_scan_block_tests) or test(virtual_scan_gate_tests)'
  9 runs -> 3 failures, always the same one:
  test_blob_reblock_honors_scan_on_proxy
  assertion `left == right` failed: positive control: with scan_on_proxy enabled,
                                    a vulnerable image's blob stays blocked
    left: 200
   right: 403

merge-base 6c8a774, same selector: 5 runs -> 0 failures
```

Note the failure is on the **positive control**, i.e. it presents as "the fix does not block", which is the worst possible shape for a flake on a security gate to take — someone will read that red as a real fail-open.

Fix confirmed locally: give the test its own fixture bytes.

```rust
let (vuln_manifest, vuln_config, vuln_layer) =
    image_manifest(b"reblock-bad-cfg", b"reblock-bad-layer");
let (clean_manifest, clean_config, clean_layer) =
    image_manifest(b"reblock-ok-cfg", b"reblock-ok-layer");
```

With that change: **8/8 green** on the same selector, and the test still revert-proves (below). Worth disambiguating the pre-existing `26091` ⟷ `26959` pair in the same pass — it is the same latent race, just narrower.

CI is currently green on this head. That is luck, not evidence: at the ~1-in-3 rate I measured, this will start failing unrelated PRs.

---

## The claims, checked one at a time

### Claim 1 — the re-block ran unconditionally, gated only on `oci_pull_is_scan_scoped` — **CONFIRMED**

Merge-base `backend/src/api/handlers/oci_v2.rs:4618-4666`: inside `if !oci_pull_is_scan_scoped(&claims)` the code goes straight to `blob_belongs_to_vulnerable_image[_any]`, with no config read anywhere in the block. Confirmed behaviourally twice: on the reverted tree a blob whose repo has `scan_on_proxy = false` still returns 403, and the hosted-repo probe above returns 403 with no `scan_configs` row at all.

### Claim 2 — the manifest gate already short-circuits, so the two halves disagreed — **CONFIRMED, with two caveats on the wording**

`maybe_gate_remote_manifest_scan` (`oci_v2.rs:7855-7873`) returns `Ok(false)` when `is_proxy_scan_enabled` is false. So yes: manifests served while blob GETs stayed 403, and #3244 confirms there is no API route that reads or deletes a `proxy_scan_results` row — I grepped every handler; the table is touched only by `ProxyScanService::lookup_verdict` / `record_verdict` and by test cleanup. DB access really was the only recourse.

Caveats on "exactly as the manifest gate does":
* the manifest gate is **Remote-only** (`:7863`); the new blob gate is not — that is B1;
* the manifest gate uses `unwrap_or(false)` (`:7869`), which fails **open** on a config-read fault. The blob gate uses `unwrap_or(true)`. The PR is stricter, which is the right call, but it is not "exactly" the same posture and the comment should not claim it is.

### Claim 3 — virtual = `virtual OR any member`, member set resolved once — **CONFIRMED**

`oci_v2.rs:4658-4676` is a correct short-circuiting OR, and `member_ids` is computed once at `:4627` and reused for both the policy decision and the blocklist — no second `fetch_virtual_members`. Semantics match `stricter_scan_policy`'s `enabled = virtual_enabled || member_enabled` (`proxy_helpers.rs:2870-2886`) and match what #3105 explicitly asked for ("the virtual and all its Remote members — the current OR-of-both-sides semantics from #3025 are fine").

On the "could one permissive member unblock what a stricter member holds?" question: **no**. OR means a single scanning-enabled member keeps the block on for the whole virtual, and the blocklist then spans *all* members' refs. The residual is over-blocking (a member with scanning off has its refs blocked because a sibling has it on), which is the fail-closed direction and is unchanged from #3023.

Also note this is *broader* than `effective_virtual_scan_policy`, which ORs the virtual with the single **resolving** member; this ORs across every member. Broader = stricter here, so fine — but the code comment says it "matches" that function, which slightly undersells it.

**Untested, though.** The only virtual coverage added is one `enable_proxy_scan` call on the member so the existing test keeps passing. Nothing exercises virtual-enabled/members-disabled, nor all-disabled-through-a-virtual-unblocks. Two cheap additions to `virtual_scan_gate_tests` would pin the OR in both directions.

### Claim 4 — fail-closed preserved at every error path — **CONFIRMED for error paths**

I traced every `?`, `unwrap_or`, `unwrap_or_default`, `if let Ok(..)` and swallowed `Err` in the changed block. There are no `?` and no `unwrap_or_default` in it. The three fallible operations:

| Operation | Failure | Result |
|---|---|---|
| `fetch_virtual_members` (`:4627-4642`) | `Err` | early `return` 423 — **withheld** |
| `is_proxy_scan_enabled` (`:4661,4668,4673`) | `Err` | `unwrap_or(true)` → `scan_active = true` → blocklist runs — **withheld/blocked** |
| `blob_belongs_to_vulnerable_image[_any]` (`:4679-4682`) | `Err` | 423 at `:4699-4709` — **withheld** |

The `unwrap_or(true)` reasoning holds, and I checked the premise rather than taking it: `is_proxy_scan_enabled` maps every sqlx error to `Err` and only `fetch_optional`'s `None` (absent row) becomes `Ok(false)` (`scan_config_service.rs:231-241`). `scan_on_proxy` is NOT NULL — if it were nullable the scalar would be `Option<Option<bool>>` and `result.unwrap_or(false)` would not compile against `Result<bool>`. So a NULL cannot masquerade as "disabled".

**I found no path where a failure results in the blob being served.** The only way to reach a serve is `Ok(false)` from a successful config read or `Ok(false)` from a successful blocklist read. That is the claim, and it survives.

The claim is true as stated about *errors*. What it does not cover is the non-error case where `scan_on_proxy` is legitimately false on a repo type where it was never a meaningful setting — B1.

### Claim 5 — deferring `severity_threshold` to #3243 — **CONFIRMED, and it is the right call**

Verified independently rather than trusting the author:

* `scan_configs.severity_threshold` has **no production consumer**. Grepping the whole backend, it appears only in the API response DTO (`api/handlers/security.rs:420,587`) and in `ScanConfig::threshold_severity()` (`models/security.rs:164-167`) — and `threshold_severity` has **zero** callers anywhere in `backend/src`. It is a stored, validated, round-tripped, entirely inert column.
* Wiring it in would weaken the gate. The persisted verdict is `findings_count > 0` (`ProxyScanVerdict::is_vulnerable` → `verdict_token()`, persisted at `proxy_helpers.rs:5925-5940`); it carries no notion of a floor. Introducing one at the column's default of `'high'` would stop blocking every image whose worst finding is `medium`/`low`/`info`, **for every repository that never set the field** — a silent, instance-wide relaxation shipped as a bug fix.

So: correct to split, and correct to split it *away from* a fix that is otherwise a pure unblock-what-the-operator-disabled change. #3243 and #3244 are both open and both accurately scoped.

---

## Timing / consistency (asked, checked)

`scan_on_proxy` is read at request time on both seams, and the two seams are independent requests. Flipping it mid-pull is benign in both directions:

* **off → on** between manifest and blob: manifest already served, blobs now blocked. Fail-closed; the client gets a 403 partway through a pull, which is correct-but-noisy.
* **on → off**: manifest was gated, blobs now serve. Availability, not exposure — and it is precisely the recovery path #3105 asked for.

No state is cached across the two, and nothing is memoised inside the request, so there is no window where a stale "enabled" decision is paired with a fresh verdict read or vice versa.

---

## Revert-proof (done properly)

I reverse-applied **only the production hunk** onto the PR tree, then checked the reverted tree against the merge base:

```
git diff 6c8a774 --shortstat   ->   1 file changed, 140 insertions(+)
```

Zero deletions, all 140 insertions inside test modules — production code byte-identical to the merge base, so a failure can only come from the claim under test. On that tree:

```
test_blob_reblock_honors_scan_on_proxy ... FAILED
assertion `left == right` failed:
  #3105 Bug 2: with scan_on_proxy disabled, the previously-flagged blob must serve again
    left: 403
   right: 200
```

It fails on the **target** assertion, and the two positive controls (403 vulnerable-with-scanning-on, 200 clean-with-scanning-on) still pass on the reverted tree — so the test cannot be satisfied by a change that blocks everything or nothing. That is a real guard, and the "negative assertion needs a positive control in the same fixture" rule is properly honoured here. Re-verified after the digest de-collision fix: still fails on the same assertion.

No hollow-test species found in the new test: it is not asserting on shape, it executes the changed path through the real router, the fixture carries the precondition (`enable_proxy_scan` before the positive control), it does not pass when reverted, and it is not asserting the bug as a requirement.

One pre-existing observation, not a regression: `test_blob_block_scoping_shared_digest_vs_shared_layer` calls `blob_belongs_to_vulnerable_image` directly rather than through the router, so it is blind to the handler-level gate and stays green whatever this PR does to it. Fine as SQL coverage; just not evidence about the handler.

---

## Non-blocking notes

1. **Extra query per blob GET.** Non-virtual: +1 round trip on every blob GET. Virtual: +1..N+1. Docker pulls issue one of these per layer. Checking the blocklist *first* and only reading the config when it returns `true` (the rare case) is semantically identical on every path except one — with the current order, a blocklist DB fault on a scanning-disabled repo is skipped, whereas reversed it would 423. Your call; the current order is the more available one, the reversed one is cheaper on the hot path.
2. **`HEAD /v2/<name>/blobs/<digest>` has no re-block at all.** `blob_belongs_to_vulnerable_image[_any]` has exactly one production call site — `handle_get_blob`. `handle_head_blob` (`oci_v2.rs:4385`) returns 200 with `Content-Length` for a blob whose GET is 403. Pre-existing (#3003 scope), not caused here, no bytes leak — but it confirms existence and size of blocked content and is worth its own issue.
3. **No freshness check on the blob side.** #3105 notes this in passing: the manifest path considers TTL / scanner-DB version via `verdict_is_reusable`, the blob blocklist is a bare `EXISTS`. So a verdict the manifest gate would treat as stale and re-scan still blocks blobs indefinitely. Out of scope here, but it is the same "no recourse" family as #3244 and might belong on that issue.

---

## What the author got right

The diagnosis is exact and matches the source: the two seams genuinely disagreed, and the reporter genuinely had no non-DB recourse. The fail-closed posture was thought through rather than assumed — choosing `unwrap_or(true)` over the manifest gate's `unwrap_or(false)`, and reasoning explicitly from "`is_proxy_scan_enabled` only returns `Ok(false)` for an absent row", is the correct and non-obvious call, and I confirmed the premise holds at the type level. Restructuring `member_ids` so the member set is resolved once and reused by both the policy check and the blocklist is cleaner than the shape it replaced. The virtual OR matches both `stricter_scan_policy` and what #3105 asked for. The new test was built with its positive controls in the same fixture and it genuinely fails when the fix is reverted. And splitting Bugs 1 and 3 out — rather than "fixing" a `severity_threshold` that would have quietly relaxed the gate for every repository on the instance — is the right instinct and the right reason.

The two blockers are both narrow: pin the test's fixture bytes, and make the hosted-repo scope an explicit, tested decision.

brandonrc:
## Both blockers addressed — pushed as `3f3b710`

Worked from a fresh shallow clone at the reviewed head `308adb8`, own Postgres 16, `DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` exported throughout (DB tests land at 0.3–0.6 s each, so nothing early-returned). The PR body has been updated with the reasoning; this comment is the evidence.

---

## B1 — decision: **scope the gate, do not keep the current behaviour**

I took option (1) from the review, not the recommended (2), because the argument for (2) does not survive checking what `scan_on_proxy` actually is on a hosted repository.

The reviewer's own premise is the deciding fact: `is_proxy_scan_enabled` returns `Ok(false)` for an **absent** row, hosted repos have no row, and no workflow creates one. So on the hosted side the flag is not a setting an operator turned off — it is a setting that has never existed and cannot be turned on through any UI or API path an operator would reach. "Keep the behaviour and document it" therefore documents a gate that is **off by default for every hosted OCI repository on the instance**, with no operator action anywhere in the causal chain. That is not a scope decision, it is a silent removal, and #3105 asked for neither: Bug 2 is entirely about proxy pulls, and the reporter's expected behaviour is scoped to "no pull **through that repository**".

The counter-argument for (2) was that (1) recreates the "flagged once, blocked forever, no API recourse" problem on the hosted side. It does — but that problem is *already* there today and is *already* tracked (#3244 for the manual escape hatch; I have filed #3259 for the automatic one). Shipping a default-open gate as the workaround for a missing admin endpoint trades a known availability annoyance for an unknown security posture, and it does it on the exact route (`handle_get_blob`) that is the **only** blob-level enforcement hosted OCI has — `maybe_gate_remote_manifest_scan` bails on non-`Remote`, there is no blob quarantine seam, and (agreeing with the review) #3235's `enforce_scan_policy_gate` is disjoint: it hangs off `artifacts` rows via `local_lookup_artifact`, while OCI blobs resolve from `oci_blobs`.

### The predicate

Not "is this a Remote request?" but **"which repository owns the refs this blocklist would match?"** — `blob_reblock_applies(scan_cfg, repo_type, repo_id)`:

| Repo type | Re-block | Why |
|---|---|---|
| `Local` / `Staging` | always applies | `scan_on_proxy` says nothing about hosted content, and an absent row must not read as "disabled". Identical to pre-#3105 behaviour. |
| `Remote` | honors `scan_on_proxy` | The path #3105 Bug 2 is about. |
| `Virtual` | the virtual's own flag ORs the whole member set; otherwise per-member by the rows above | #3025 stricter-of-two, preserved. |

Two things fall out of doing it per-owner rather than as one whole-request boolean:

1. **It closes a bypass the request-level version opened.** A `Virtual` with proxy scanning off (or with no config row) spans its members' refs — including a `Local` member's. Under the whole-request boolean, `GET /<virtual>/…/blobs/<layer>` served a hosted member's blocked blob 200 while `GET /<local-member>/…/blobs/<layer>` was 403. Now pinned by `blob_via_virtual_still_blocks_hosted_member_refs_with_scanning_off`.
2. **It removes the extra query from the hosted hot path.** The `Local`/`Staging` arm short-circuits without touching `scan_configs`, so the non-blocking-note #1 about "+1 round trip on every blob GET" now applies only to proxy repos. Docker pulls against hosted repos pay nothing.

### #3105's reported topology still resolves

The reporter's virtual has a `Local` **and** a `Remote` member. The image was *proxied*, so its `manifest_blob_refs` are recorded under the **Remote** member. Setting `scan_on_proxy: false` on the virtual and that member drops those refs out of the blocklist and the blob serves — the requested outcome, unchanged. What no longer happens is the same setting quietly disabling enforcement on hosted content the operator never mentioned. (If the same image had *also* been pushed to the Local member, it stays blocked there — correctly, and reachable via the same DB delete as before, pending #3244.)

### Revert-proof for the scope

Production restored to this PR's earlier whole-request-boolean shape, tests kept:

```
test_hosted_blob_reblock_ignores_absent_scan_config ... FAILED
  hosted: the CONFIG blob of a vulnerable image must stay blocked with no scan_configs row (#3245 B1)
    left: 200   right: 403

blob_via_virtual_still_blocks_hosted_member_refs_with_scanning_off ... FAILED
  a hosted member's vulnerable blob must stay blocked through the virtual (#3245 B1)
    left: 200   right: 403
```

That is your live 403 → 200 result, now a test. Both #3105 guards still pass on that tree, so the two experiments are independent rather than one mutation knocking everything over.

Also fixed the inaccuracy you flagged: the code comment no longer claims the blob gate behaves "exactly as the manifest gate does". It now states the two differences explicitly — the manifest gate is `Remote`-only and uses `unwrap_or(false)` (fails open on a config-read fault); this one is scoped per ref-owner and uses `unwrap_or(true)` (fails closed).

---

## B2 — flake fixed, 12/12 green

Root cause is exactly as diagnosed. Rather than hand-picking literals I made the fixture bytes **UUID-derived**, so the class cannot come back when someone adds a fourth test:

```rust
fn unique_fixture_bytes(label: &str) -> Vec<u8> {
    format!("{label}-{}", Uuid::new_v4()).into_bytes()
}
```

Applied to all three colliding tests — the new one plus the pre-existing `test_get_blob_blocked_for_vulnerable_image_but_not_clean_one` ⟷ `blob_via_virtual_blocks_vulnerable_but_not_clean` pair you suggested disambiguating in the same pass. Both modules now have the helper with a comment explaining *why* (digest-keyed table + global cleanup + concurrent nextest), so the next author sees the trap.

```
selector: -E 'test(proxy_scan_block_tests) or test(virtual_scan_gate_tests)'
  12 runs -> 12 passed, 0 failed   (55 tests per run)

full oci_v2 suite: 530/530 green on 3 consecutive runs
```

(One unrelated flake surfaced once during the sweep — `remote_blob_streaming_fallback_tests::remote_blob_streams_large_layer_and_warms_cache` — then passed in isolation and in all three repeat runs. Not this change; the code path it exercises is untouched by the diff.)

### Still revert-proves after the de-collision

Faithful merge-base revert of the production code, all tests kept:

```
test_blob_reblock_honors_scan_on_proxy ... FAILED
  #3105 Bug 2: with scan_on_proxy disabled, the previously-flagged blob must serve again
    left: 403   right: 200

blob_via_virtual_all_scanning_off_unblocks ... FAILED
  #3105 Bug 2 through a virtual: scanning off on the virtual AND its proxy member must unblock the blob
    left: 403   right: 200
```

Both fail on the **target** assertion; every positive control still passes.

**Faithfulness, verified not asserted.** On the reverted tree:

```
git diff 6c8a774 --shortstat
  1 file changed, 617 insertions(+), 93 deletions(-)
```

Every hunk sits inside `mod proxy_scan_block_tests` / `mod virtual_scan_gate_tests`. The 93 deletions are test-only — the shared `VirtualBlobFixture` refactor and the byte-literal swaps — not the "insertions only" shape the rule expects, so I checked the underlying property directly instead of relying on the count: stripping every `#[cfg(test)] mod … { }` block from the merge-base file and from the reverted file leaves **63,459 bytes each, byte-identical**. Production code on the reverted tree *is* the merge base, so the two failures can only come from the claim under test.

---

## Claim 3 — the virtual OR, now tested in both directions

You noted it was untested either way. Added:

- `blob_via_virtual_enabled_blocks_when_member_scanning_off` — virtual enabled, member has no config row → blocked; clean sibling 200 as the positive control.
- `blob_via_virtual_all_scanning_off_unblocks` — off on both → serves; the block is observed **first** with the member enabled, in the same fixture, so it cannot be satisfied by a fixture that never blocked.
- `blob_via_virtual_still_blocks_hosted_member_refs_with_scanning_off` — the hosted-member case above.

The four virtual blob-seam cases share a `VirtualBlobFixture` so the ~90 lines of staging exist once instead of four times (jscpd).

---

## Non-blocking notes — dispositions

1. **Extra query per blob GET** — now zero for `Local`/`Staging` (short-circuits before touching `scan_configs`), unchanged for proxy repos. Left the check order as-is: your read is right that checking the blocklist first is cheaper but turns a blocklist DB fault on a scanning-disabled repo into a 423, and availability wins on a path a scanning-disabled repo has opted out of.
2. **`HEAD /v2/<name>/blobs/<digest>` has no re-block** — filed as **#3258**. Confirmed the single production call site. Not folded in: pre-existing #3003 scope, and it deserves its own revert-proof.
3. **No freshness/TTL on the blob side** — filed as **#3259**, cross-referenced to #3244 as the same "no recourse" family (manual vs automatic escape hatch).

## Gates

`cargo fmt --all --check` clean · `cargo clippy --workspace --all-targets -- -D warnings` clean · `SQLX_OFFLINE=true cargo check --workspace --all-targets` clean (no `sqlx::query!` text changed, so no `cargo sqlx prepare`) · no migration.

Thanks for the live probe on B1 — the 403 → 200 pair on one tree and one database is what made the scope question decidable instead of arguable, and it is now a regression test.

PR #3248, opened 2026-08-10
ci: stop shared-cargo-registry mutation flake and unmask coverage build failures

### LINKED ISSUE #3154: Code Coverage job intermittently fails building aws-lc-sys (cc-rs), masquerading as a coverage-gate failure

brandonrc:
Fixes #3154

## What the flake actually is

The Code Coverage job — and, less often, every other Rust job on `ak-ci-runners` — intermittently dies **1–2.5 minutes into its build** with one of three signatures, on commits that pass identical builds in sibling jobs:

| Signature | Example | Root diagnostic in the log |
|---|---|---|
| `error occurred in cc-rs` building vendored `aws-lc-sys` C | job 92771558547 (run 31148018643, 2026-08-07 04:40), job 92855151814 (run 31175059925, 2026-08-07 11:42) | `aws-lc/include/openssl/bn.h:11: error: unterminated #ifndef` — the header under `/cache/cargo-home/registry/src/...` was **truncated mid-read** |
| `could not execute process 'cargo-llvm-cov ... rustc ...' — No such file or directory` | job 93166712233 (2026-08-09 00:09:20) | spawn failed because the child's **cwd** — the crate's registry src dir (`cranelift-codegen-meta`) — had been **deleted** |
| `could not parse/generate dep info at .../deps/<crate>.d — No such file or directory` | jobs 93166712233 + 93167009499 (two different PRs' coverage jobs, two different pods, failing within 13 s of each other), job 93341706470 (Backend Unit Tests, PR #3237, 2026-08-10 03:19:43) | sources listed in the `.d` (all **registry** crates: hyper, regex-automata, rustls, wasmtime, curve25519-dalek…) had vanished |

**Frequency:** in the last ~7 days of `ci.yml` failure history I can attribute at least 6 failed jobs to this class (2× aws-lc-sys on 2026-08-07, 2× dep-info on 2026-08-09, 1× dep-info unit-tests on 2026-08-10, 1× a 1m35s Check Rust death in the same 2026-08-10 window), out of ~16 Code-Coverage-job failures overall — the rest were genuine (real coverage-gate misses at ~15 min, real compile errors on dependabot bumps, a real test failure). Every instance cleared on rerun.

## Root cause (shared by both variants)

The ARC runner pods set `CARGO_HOME=/cache/cargo-home` on a **hostPath volume shared by up to 6 concurrent runner pods** (`artifact-keeper-arc-runners/argocd/arc-ci-runners-values.yaml`). Two things treat that registry as private, mutable state:

1. **`Swatinem/rust-cache`** — written for GitHub-hosted runners with a private `~/.cargo`:
   - its **restore** tar-extracts registry index caches, `registry/cache/*.crate`, and `-sys` crate sources (unlink + rewrite) **over the live shared copy** at every job start;
   - its **post-save clean** (`cleanRegistry` in the pinned v2.9.1 source) **deletes `registry/src`** for all non-`-sys` crates and prunes `registry/cache` and the index `.cache` down to *the saving job's own package set* — which, on a registry shared with other jobs and other repos, deletes files others still need.
2. **cargo itself** — whenever those files go missing it re-extracts them (`rm_rf` + rewrite) under the package-cache lock. The lock protects extraction against other *extractions*, but sibling pods' **compile phases read registry sources without holding any lock**, so they see files mid-delete or mid-rewrite.

Every observed symptom is a direct read of that mutation: cc reading a half-written vendored header (`unterminated #ifndef`), cargo spawning rustc with a deleted registry dir as cwd (`could not execute process`, ENOENT), and dep-info translation statting deleted registry sources. It also explains why the failure window is always the first ~1–2.5 min of a build (maximum overlap with sibling restores/extraction storms), why reruns always clear it, and why two pods failed within 13 s of each other on 2026-08-09. The integration job's existing "stale/truncated crates.io index" fetch-retry comment (added for run 31176621357) is the same corruption class caught at the index instead of the sources.

**Do the aws-lc-sys and dep-info variants share a root cause? Yes.** Same shared-registry mutation, different read path: `cc` reads the vendored C/headers directly (truncation ⇒ cc-rs error), rustc/cargo read `.rs` sources and spawn cwds (deletion ⇒ dep-info / could-not-execute). The failing invocation is the same `cargo test --no-run --workspace --lib` in both.

**What it is NOT (all log-verified):** not sccache — every failing job ran with `RUSTC_WRAPPER: ""` (the existing bypass), so the "sccache dep-info flake" diagnosis is wrong for these instances; not OOM — no `signal: 9` anywhere, and the failures are ENOENT/truncation, not kills; not disk pressure — ENOSPC never appears; not the monthly registry-cleanup CronJob — it runs 05:00 on the 1st, matching none of the failure times.

## The fix (narrowest that addresses the cause; CI-config-only)

1. **Private per-job `CARGO_HOME`** — a one-line step in each of the 5 Rust jobs (`check-rust`, `test-backend-unit`, `coverage`, `test-backend-integration`, `build-windows-dev`) sets `CARGO_HOME=$RUNNER_TEMP/cargo-home` before rust-cache runs. `RUNNER_TEMP` is on the pod-local emptyDir work volume and ARC pods are ephemeral (one job each), so each job's cargo home is private and self-consistent; nothing can mutate it mid-build. This removes both the *writers* (this repo's rust-cache restores/saves now target private state) and the *exposure* (builds no longer read shared state), so it also shields against mutations from other repos' jobs on the same runners.
   - Warm path unchanged for the three rust-cache jobs: the tarball already contains registry index+cache and now restores into the private path (restore was already how every job start worked).
   - `prefix-key` bumped `v0-rust` → `v1-rust`: existing tarballs archived the old shared `/cache/cargo-home` with absolute paths; restored under an old key they would write into the shared volume again and leave the private home cold. First run per key is a cold build (~9 min for clippy per the existing comment), then warm.
   - Tool binaries are unaffected: the runner image deliberately bakes nextest/llvm-cov/sqlx-cli into `/usr/local/cargo/bin`, which stays on PATH regardless of `CARGO_HOME` (its Dockerfile comment says exactly this).
   - The two uncached jobs (integration, windows) trade the shared warm registry for a per-run `cargo fetch` from crates.io (~1–2 min); their existing `CARGO_NET_RETRY`/fetch-retry absorb network hiccups.

2. **Unmask the failure mode** (the issue's "masquerading" half): the coverage job now builds the instrumented test binaries in their own step (`cargo nextest run --no-run`). If the build fails, a `::error::` annotation titled **"Coverage BUILD failed — not a coverage-gate failure"** appears on the PR checks page, stating that no coverage was measured and the 70% gate / 50% floor / jscpd duplication gate never ran. A genuine compile error still fails the job (and Check Rust alongside it); nothing is retried or swallowed.

## Alternatives evaluated (per the issue's checklist)

- **sccache knobs / per-job sccache dir / cache-size tuning:** moot — sccache is bypassed (`RUSTC_WRAPPER: ""`) in every failing job, and `cc` invocations never went through sccache. The failure lives in the cargo registry, not the compile cache.
- **`CARGO_INCREMENTAL`:** already `0` in the affected jobs; incremental state is in the (pod-local) target dir and is not implicated.
- **cc-rs parallelism / `CARGO_BUILD_JOBS` caps:** would only shrink the race window; the diagnostic is truncation/ENOENT, not resource exhaustion, and single-threaded builds read the same shared files.
- **Not caching `*-sys` crates:** backwards — rust-cache already *keeps* `-sys` sources on clean; it's the restore *rewriting* them and the deletion of everything else that breaks builds. Excluding more crates increases re-extraction churn on shared state.
- **Warm cache key for coverage (issue suggestion 2):** already exists (`shared-key: coverage`, full-match hits observed in the failing jobs' logs) — it did not prevent the flake because the mutation is cross-job, not cold-cache.
- **Narrow signature-matched retry:** rejected while a config fix exists that removes the cause; retry stays the workaround only for unfixed instances (`gh run rerun --failed`).
- **Fixing it in the runner infra instead (per-pod hostPath subdirs):** equivalent effect, but requires an `artifact-keeper-arc-runners` change + redeploy and would still leave workflow-level rust-cache assumptions in play; the workflow-side change is self-contained and reviewable here. A follow-on infra cleanup (retiring the now-mostly-idle shared `cargo-home`) can ride the next runner-image update.

## Proved vs. reasoned

**Proved from logs/config:** the three signatures and their locations under `/cache/cargo-home/registry/src`; the truncated-header diagnostic; the cross-pod simultaneity (13 s apart on 2026-08-09); `RUSTC_WRAPPER=""` in all failing jobs (sccache exonerated); the shared hostPath `CARGO_HOME` (iac values); rust-cache v2.9.1's `cleanRegistry` deleting non-`-sys` `registry/src` and pruning `registry/cache`/index to the saver's package set (read at the pinned commit); tools living in `/usr/local/cargo/bin` (runner Dockerfile); nextest supporting `--no-run`.

**Reasoned but not directly observed:** which specific mutation event (a restore, a save-clean, or a cargo re-extraction) fired at each individual failure instant — the runner host is not reachable from here and GitHub's API does not expose sibling repos'/pods' filesystem activity, so I could not catch a deleter in the act. The mechanism is the only one consistent with all observed evidence, and the fix removes the entire class rather than one deleter. I could not pre-validate on the PR's own CI either: this PR necessarily changes the workflow it needs to exercise, and the flake is a low-probability race — a green run here demonstrates the changed config builds and gates normally (including the new cold-cache path), not the absence of the race. Validation plan: after merge, watch for any recurrence of the three signatures; each pre-merge instance was ≤3 days apart, so two quiet weeks is strong evidence.

**Validated locally:** workflow YAML parses (PyYAML); `actionlint` is not available on this host — flagging that honestly. No Rust code touched, so no build/fmt was needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -396,7 +396,7 @@ jobs:
   test-backend-unit:
     name: 🧪 Backend Unit Tests
     runs-on: ak-ci-runners
-    timeout-minutes: 25
+    timeout-minutes: 40
     needs: [changes, release-bump]
     # Skip on docs-only PRs and version-bump-only release PRs — see check-rust.
     if: ${{ !cancelled() && needs.changes.outputs.code != 'false' && needs.release-bump.outputs.bump_only != 'true' }}
@@ -560,7 +560,7 @@ jobs:
   coverage:
     name: 📊 Code Coverage
     runs-on: ak-ci-runners
-    timeout-minutes: 35
+    timeout-minutes: 45
     needs: [changes, release-bump]
     # Skip on docs-only PRs and version-bump-only release PRs — see check-rust.
     if: ${{ !cancelled() && needs.changes.outputs.code != 'false' && needs.release-bump.outputs.bump_only != 'true' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -293,10 +293,46 @@ jobs:
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
+      # PRIVATE PER-JOB CARGO_HOME (#3154). The ARC runner pods set
+      # CARGO_HOME=/cache/cargo-home, a hostPath volume SHARED by up to six
+      # concurrent runner pods on the node — but two things treat that
+      # registry as private, mutable state:
+      #   1. Swatinem/rust-cache: its restore tar-extracts registry files
+      #      (unlink+rewrite) over the live shared copy, and its post-save
+      #      clean DELETES registry/src for non `-sys` crates and prunes
+      #      registry/cache and index caches down to the saving job's own
+      #      package set (see cleanRegistry in the pinned rust-cache src).
+      #   2. cargo itself: whenever those files go missing it re-extracts
+      #      them (rm_rf + rewrite) under a package-cache lock that sibling
+      #      pods' COMPILE phases do not hold.
+      # Any job compiling while another pod restores, saves, or re-extracts
+      # reads half-written or deleted sources. Observed as: aws-lc-sys cc
+      # "unterminated #ifndef" on a truncated vendored header, "could not
+      # execute process ... (os error 2)" (spawn cwd = deleted registry src
+      # dir), and "could not parse/generate dep info ... No such file or
+      # directory" cascades — the #3154 flake family. All failing jobs had
+      # RUSTC_WRAPPER="" (sccache was NOT involved).
+      #
+      # RUNNER_TEMP is on the pod-local emptyDir work volume and the pods
+      # are ephemeral (one job each), so this makes every job's cargo home
+      # private and self-consistent. Warm-path cost is unchanged for jobs
+      # with rust-cache (the tarball already contains registry index+cache
+      # and restores into the private path). Tool binaries are unaffected:
+      # the runner image bakes them into /usr/local/cargo/bin, which stays
+      # on PATH regardless of CARGO_HOME (ak-runner-rust Dockerfile).
+      - name: Isolate CARGO_HOME per job (#3154)
+        run: echo "CARGO_HOME=$RUNNER_TEMP/cargo-home" >> "$GITHUB_ENV"
+
       # Cache the dependency build so only the workspace crate recompiles.
       # ~635 deps are byte-identical run-to-run; cold compile was ~9m.
+      #
+      # prefix-key bumped v0->v1 (#3154): pre-existing tarballs archived the
+      # old shared /cache/cargo-home with absolute paths, so restoring one
+      # under an old key would write into the shared volume again and leave
+      # the private CARGO_HOME cold. A fresh namespace repopulates cleanly.
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
+          prefix-key: v1-rust
           shared-key: clippy
 
       - name: Check formatting
@@ -403,10 +439,19 @@ jobs:
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
+      # Shield this job from concurrent mutation of the shared registry —
+      # see the #3154 comment on the same step in check-rust. A dep-info
+      # ENOENT instance of that flake hit THIS job on 2026-08-10
+      # (job 93341706470, PR #3237).
+      - name: Isolate CARGO_HOME per job (#3154)
+        run: echo "CARGO_HOME=$RUNNER_TEMP/cargo-home" >> "$GITHUB_ENV"
+
       # Separate cache key from check-rust: same target dir but a different
       # build profile (test vs clippy), so they must not share a cache.
+      # prefix-key: see the #3154 note in check-rust.
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
+          prefix-key: v1-rust
           shared-key: unit
 
       # Catch migration-slot collisions before they reach the e2e/nightly job.
@@ -585,13 +630,25 @@ jobs:
             echo "### Coverage: skipped (no Rust source changes in this PR)" >> "$GITHUB_STEP_SUMMARY"
           fi
 
+      # Shield this job from concurrent mutation of the shared registry —
+      # see the #3154 comment on the same step in check-rust. This job is
+      # the most frequent victim: its instrumented build is the longest in
+      # CI, so its compile phase overlaps other jobs' cache restores/saves
+      # most often (observed 2026-08-07 twice as aws-lc-sys cc failures,
+      # 2026-08-09 twice as dep-info ENOENT, both pods within seconds).
+      - name: Isolate CARGO_HOME per job (#3154)
+        if: steps.detect.outputs.rust_changed == 'true'
+        run: echo "CARGO_HOME=$RUNNER_TEMP/cargo-home" >> "$GITHUB_ENV"
+
       # Distinct cache key: the instrumented coverage build is incompatible
       # with the clippy/unit profiles, and cargo-llvm-cov builds into
       # target/llvm-cov-target, so point rust-cache there or it caches an
       # empty target/ and silently no-ops.
+      # prefix-key: see the #3154 note in check-rust.
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         if: steps.detect.outputs.rust_changed == 'true'
         with:
+          prefix-key: v1-rust
           shared-key: coverage
           workspaces: ". -> target/llvm-cov-target"
 
@@ -604,6 +661,30 @@ jobs:
         run: sqlx migrate run --source backend/migrations
         env:
           DATABASE_URL: postgresql://registry:registry@localhost:5432/artifact_registry
+
+      # Build the instrumented test binaries as a step of their own, so a
+      # BUILD failure is visibly distinct from a coverage-gate failure
+      # (#3154's "masquerading" half). Before this split, an aws-lc-sys /
+      # dep-info build flake surfaced as "Code Coverage: fail", which reads
+      # as "your new code is under the 70% threshold" and sends people off
+      # to write tests they do not need. The only tell was the job duration
+      # (~1-2 min vs ~14 min for a real gate run), which is not
+      # discoverable. The ::error:: annotation below shows up directly on
+      # the PR checks page.
+      - name: Build instrumented test binaries
+        if: steps.detect.outputs.rust_changed == 'true'
+        env:
+          SQLX_OFFLINE: true
+          RUSTC_WRAPPER: ""
+        # The explicit `clean --workspace` keeps the implicit-clean
+        # semantics the single `cargo llvm-cov nextest` invocation had.
+        run: |
+          source <(cargo llvm-cov show-env --sh)
+          cargo llvm-cov clean --workspace
+          if ! cargo nextest run --workspace --lib --no-run; then
+            echo "::error title=Coverage BUILD failed — not a coverage-gate failure::cargo could not build the instrumented test binaries, so no coverage was measured; the 70% new-code gate, the 50% floor, and the jscpd duplication gate never ran. A genuine compile error will also fail Check Rust / Backend Unit Tests on this commit; if those are green and this log shows ENOENT / 'dep info' / cc errors in vendored sources, this was the #3154 flake class — rerun the job."
+            exit 1
+          fi
 
       - name: Generate coverage report
         if: steps.detect.outputs.rust_changed == 'true'
@@ -634,13 +715,13 @@ jobs:
         # of cargo-llvm-cov's default one-.profraw-per-test-process: with
         # ~12k test processes the per-process files cost tens of GB of write
         # I/O per run, while the pool online-merges the same counters into at
-        # most 32 files (~92MB) with byte-equal coverage aggregates. The
-        # explicit `clean --workspace` keeps the implicit-clean semantics the
-        # single `cargo llvm-cov nextest` invocation had.
+        # most 32 files (~92MB) with byte-equal coverage aggregates.
+        #
+        # NO `cargo llvm-cov clean` here: it ran in the build step above and
+        # must not run again between build and test execution.
         run: |
           source <(cargo llvm-cov show-env --sh)
           export LLVM_PROFILE_FILE="${LLVM_PROFILE_FILE%/*}/%32m.profraw"
-          cargo llvm-cov clean --workspace
           cargo nextest run --workspace --lib --test-threads 8
           cargo llvm-cov report --lcov --output-path lcov.info
 
@@ -1000,6 +1081,19 @@ jobs:
 
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      # Shield this job from concurrent mutation of the shared registry —
+      # see the #3154 comment on the same step in check-rust. The
+      # "stale/truncated crates.io index" failure documented on the fetch
+      # step below is the same corruption class (a sibling pod rewriting
+      # the shared registry index mid-read). This job carries no
+      # rust-cache, so a private CARGO_HOME means it now downloads the
+      # dependency graph from crates.io each run instead of reading the
+      # shared warm registry; CARGO_NET_RETRY and the fetch retry below
+      # already absorb network hiccups, and correctness beats the ~1-2
+      # minutes of download.
+      - name: Isolate CARGO_HOME per job (#3154)
+        run: echo "CARGO_HOME=$RUNNER_TEMP/cargo-home" >> "$GITHUB_ENV"
 
       # Resolve and download the dependency graph in its own step, ahead of the
       # test steps, with a retry.
@@ -1654,6 +1748,13 @@ jobs:
     if: github.event_name == 'push' && github.ref == 'refs/heads/main'
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+
+      # Shield this job from concurrent mutation of the shared registry —
+      # see the #3154 comment on the same step in check-rust. No rust-cache
+      # here either (cf. the integration job note): the graph is downloaded
+      # per run, which this main-push-only job can afford.
+      - name: Isolate CARGO_HOME per job (#3154)
+        run: echo "CARGO_HOME=$RUNNER_TEMP/cargo-home" >> "$GITHUB_ENV"
 
       - name: Build Windows binary (cross-compile)
         env:


brandonrc:
## Independent validation review

I reviewed this as an adversarial reviewer — fresh shallow clone at `/tmp/ak-val-3248`, `git fetch origin pull/3248/head`, head SHA `8937cf452d31a05d2422ea52fb999a1447f87789`, merge-base `01ebcb23b783b3191cb6a0d61d4e2d7a91b6f622`. I did not push anything and did not trigger CI. Every claim below was checked against a primary source (job logs via the Actions API, the `Swatinem/rust-cache` tree at the pinned commit, and the ARC scale-set values / runner Dockerfile in `artifact-keeper-arc-runners`), not against the PR description.

### Verdict

**The diagnosis holds up — better than the PR body claims — and the change is mechanically correct, complete, and YAML-valid. There is one operational blocker to resolve before merge, and it is not about correctness: the `prefix-key` v0→v1 bump makes every Rust job cold, and `test-backend-unit`'s `timeout-minutes: 25` is very likely too small for a cold run.** With 9 runs queued, a wave of jobs that burn a runner slot for 25 minutes and then time out would make the queue materially worse, not better.

Recommended before merge (both one-line changes in this PR):

1. Raise `test-backend-unit` `timeout-minutes: 25` → `40`, and `coverage` `35` → `45`, for the transition. Numbers below.
2. Right after merge, delete the dead `v0-rust-*` cache entries so they do not LRU-evict the fresh `v1-rust` ones — the repo cache budget is already effectively full (~10.6 GiB across 14 Rust entries).

Everything else I checked is sound.

---

### Claim-by-claim

**1. One bug, not two: concurrent mutation of a shared cargo registry — CONFIRMED.**

The infrastructure premise is exactly right. `argocd/arc-ci-runners-values.yaml`: `CARGO_HOME: /cache/cargo-home` (l.79-80), mounted from `volumes.runner-cache` → `hostPath: /home/runner-cache` (l.255-258), `maxRunners: 6` (l.56). So up to six concurrent pods on one node share one mutable cargo registry.

I also checked the *other* scale sets, which the PR does not mention: `arc-beefy-runners` and `arc-e2e-runners` mount **no** `/cache` volume and set no `CARGO_HOME`. So `ak-ci-runners` is the only pool with the shared registry, which bounds the blast radius of the bug and makes the completeness argument tractable.

I went further than the PR body and reconstructed the concurrency for the 2026-08-09 incident. At 00:09:20 the only `ak-ci-runners` jobs alive were:

| job | pod | window |
|---|---|---|
| 93167009499 Code Coverage | `...-cmzgc` | 00:07:42 → 00:10:15 (FAIL) |
| 93166712233 Code Coverage | `...-7z6kz` | 00:08:21 → 00:10:01 (FAIL) |
| 93171033796 Check Rust | `...-x4hs9` | 00:05:33 → 00:10:27 (pass) |

Both coverage pods restored **the same cache key** (`v0-rust-coverage-Linux-x64-d2790a82-0cd4c402`, `full match: true`) — pod A's restore completed 00:08:23, pod B's at **00:08:53**. Pod A began compiling at 00:08:29 and died at 00:09:20. So pod B tar-extracted the registry over `/cache/cargo-home` **while pod A was mid-compile out of it**. The Check Rust pod got a full match and therefore never ran the post-save clean, so in *this* incident the mutator was a concurrent **restore** (and/or cargo's own re-extraction), not the clean. The clean path is real (see claim 3) but was not the actor here. That does not weaken the fix — both mutators are removed by the same change — but the PR comment should not be read as having pinned which one fired.

**2. Not sccache — CONFIRMED, for all four instances I pulled.**

`RUSTC_WRAPPER:` (empty) in the env block of every failing job log: 93167009499 l.534, 93166712233 l.532, 93341706470 l.132/237/333/571, 92771558547 l.534, 92855151814 l.532. Job 93341706470 additionally prints `sccache --show-stats` with `Cache read errors 0 / Cache write errors 0 / Cache errors 0`. Inside the coverage job the wrapper is `cargo-llvm-cov`, not sccache (`RUSTC_WRAPPER = Some(/usr/local/cargo/bin/cargo-llvm-cov)` in the aws-lc build-script env dump). The standing "sccache dep-info flake" diagnosis is wrong for these instances.

**3. rust-cache mutates the live shared copy — CONFIRMED from source at the pinned commit.**

Read `Swatinem/rust-cache` at `c19371144df3bb44fab255c43d04cbc2ab54d1c4` (tag `v2.9.1`):

- `src/config.ts:15` — `CARGO_HOME = process.env.CARGO_HOME || ~/.cargo`, so it targets whatever `CARGO_HOME` says.
- `src/config.ts:272-278` — `cachePaths = [CARGO_HOME/registry, CARGO_HOME/git]`, **plus `CARGO_HOME/bin`, `.crates.toml`, `.crates2.json`** because `cache-bin` defaults to true. Restore is a tar extract straight over those live paths.
- `src/cleanup.ts:150-160` (`cleanRegistry`) — `registry/src` entries are kept only for packages whose name `endsWith("-sys")`; everything else is `rmRF`'d. That is precisely the shape of the `bergshamra-pkcs12-0.5.1/src/lib.rs` ENOENT.
- `src/cleanup.ts:162-180` — `registry/cache` pruned to the saving job's exact `{name}-{version}.crate` set; `cleanRegistryIndexCache` prunes the index `.cache` to the saving job's package set.
- `src/cleanup.ts:106-111` (`cleanBin`) — deletes binaries in `CARGO_HOME/bin` not in the job's own set. On the shared volume that has been deleting into a directory the cleanup CronJob explicitly protects as "baked/installed tool binaries". Another good reason to stop pointing this at a shared path.

`src/save.ts` wraps every clean in `try/catch` and `saveCache` in an outer `reportError`, so a private `CARGO_HOME` with no `bin/` directory yet will not fail the post step. Checked specifically, because that was my first guess at a way this change breaks jobs.

**4. Log evidence — CONFIRMED, with one correction to the timestamps.**

`bn.h:11: error: unterminated #ifndef` is real. Job **92771558547** (run 31148018643 **attempt 1**, 2026-08-07 04:40:20), truncated header at `/cache/cargo-home/registry/src/index.crates.io-.../aws-lc-sys-0.39.0/aws-lc/include/openssl/bn.h`. The second aws-lc instance, job **92855151814** (run 31175059925 attempt 1, 11:42:27), is the same class on a different file: `generated-include/openssl/boringssl_prefix_symbols_asm.h:7: unterminated #ifndef` **and** `:4: unterminated #else` — a header truncated mid-file, which is what a concurrent rewrite looks like and is not something a deterministic compile produces. Note both are on **attempt 1**; the current run/job listings show attempt 2, so `gh run view` on the run id will not show them.

Correction: the two 2026-08-09 coverage jobs did **not** fail 13 seconds apart. Their *first* errors are 00:09:20.074 (93167009499) and 00:09:20.655 (93166712233) — **0.6 s apart**, on different pods. The 00:09:33 in the PR body is 93167009499's *third* error, not the second job's. The corrected numbers make the point more strongly than the ones in the comment; worth fixing the comment text so the next reader can reproduce it.

**5. Not OOM, not disk, not the monthly CronJob — CONFIRMED.**

Zero matches for `ENOSPC|No space left|signal: 9|Killed|out of memory` across all five failing job logs. `e2e/runner-cache-cleanup-cronjob.yaml` is `schedule: "0 5 1 * *"` (05:00 on the 1st); the failures are 08-07 04:40, 08-07 11:42, 08-09 00:09 (×2), 08-10 03:19. None on the 1st.

**6. `$RUNNER_TEMP` per-job `CARGO_HOME` removes the class — CONFIRMED, including the `$RUNNER_TEMP` question.**

This was the thing most likely to make the PR silently a no-op, so I checked it empirically rather than by reasoning. Job 93171033796's log contains `/home/runner/_work/_temp/git-credentials-....config`, so `RUNNER_TEMP=/home/runner/_work/_temp` on these pods. `/home/runner/_work` is the `work` volume, `emptyDir: {}` (values l.126-127, l.249-250) — pod-local, and the pods are ephemeral (one job each). `RUNNER_TEMP` is set for every step and the runner only wipes `_temp` at *job* start, so it persists across steps; the `Isolate` step runs immediately after `checkout`, before any cargo step, in all five jobs (verified by dumping the parsed step order).

It is also stable *across* pods (same repo ⇒ same `GITHUB_WORKSPACE` depth), which matters for claim 7.

**7. `prefix-key` v0→v1 — CONFIRMED, with a mechanism correction.**

The reason is right, the description is slightly off. `@actions/cache` does not store literally absolute paths; `resolvePaths` stores each path *relative to `GITHUB_WORKSPACE`*, so `/cache/cargo-home/registry` is archived as `../../../../cache/cargo-home/registry` and extracted with `-C $GITHUB_WORKSPACE`. Net effect is identical to what the comment says: restoring a v0 tarball under a private `CARGO_HOME` would write into `/cache/cargo-home` and leave the private home cold. So the bump is **necessary**, not cosmetic. And because `GITHUB_WORKSPACE` is the same on every pod, v1 tarballs saved under `../../_temp/cargo-home/...` restore correctly on any other pod. Good.

The cost of the bump is that it also throws away the **target dirs**, which are the expensive part — see the risk section.

**8. Tools resolve from `/usr/local/cargo/bin` — CONFIRMED, from the image definition *and* from a live log.**

`runner-images/rust/Dockerfile` sets `ENV CARGO_HOME=/usr/local/cargo`, `PATH=/usr/local/cargo/bin:$PATH`, `RUSTUP_HOME=/usr/local/rustup`, and installs `sccache`, `cargo-nextest` 0.9.137, `cargo-llvm-cov`, `cargo-audit`, and `sqlx-cli` into `/usr/local/cargo/bin` — with a comment saying this is deliberate precisely because `CARGO_HOME` is overridden at runtime. Crucially this is not just theory: the aws-lc log shows `RUSTC_WRAPPER = Some(/usr/local/cargo/bin/cargo-llvm-cov)` resolving *while* `CARGO_HOME` was `/cache/cargo-home`. `rustup`/`rustc` come from `RUSTUP_HOME`, independent of `CARGO_HOME`. No step in `ci.yml` does `cargo install` on a self-hosted runner (the one `taiki-e/install-action` is in `security-audit`, on `ubuntu-latest`).

I also confirmed `cargo nextest run --no-run` is a valid flag on the pinned 0.9.137 (`--no-run  Compile, but don't run tests`).

**9. No retry added — CONFIRMED.** The diff adds none. The only retry in the file is the pre-existing 3-attempt `cargo fetch` loop in the integration job, untouched. Genuine compile errors still fail, and the new step re-raises with `exit 1`.

---

### Job coverage: **COMPLETE**

This is the thing that had to be right, so I enumerated rather than trusted. Parsing the head `ci.yml`, there are 14 jobs. Seven run on `ak-ci-runners` (the only pool with the shared registry):

| job | host cargo? | isolated |
|---|---|---|
| `check-rust` | yes | ✅ |
| `test-backend-unit` | yes | ✅ |
| `coverage` | yes | ✅ |
| `test-backend-integration` | yes | ✅ |
| `build-windows-dev` | yes | ✅ |
| `build-backend-image` | **no** | n/a |
| `build-openscap-image` | **no** | n/a |

The two image jobs are `docker/build-push-action` only; `docker/Dockerfile.backend` runs `cargo chef` / `cargo build` **inside** the build container with the image's own `CARGO_HOME`, with no bind or cache mount to the host. They cannot touch `/cache/cargo-home`. The remaining seven jobs are `ubuntu-latest` / `ubuntu-24.04` (GitHub-hosted), including `security-audit`'s `cargo audit`.

Widening beyond this file: `docker-publish.yml` has one `ak-ci-runners` matrix entry but zero host `cargo` invocations; `mesh-e2e.yml` uses `ak-e2e-runners` (no shared registry). Across the whole org, `gh search code --owner artifact-keeper "ak-ci-runners"` returns only this repo, `-iac` (Helm/YAML), `-site` (Next.js), and `-web` (Node) — no other repo compiles Rust on this pool. **So after this PR there is no remaining writer to the shared registry, and the fix is complete.**

One fragility worth noting, not a blocker: in `coverage` the `Isolate` step is gated on `steps.detect.outputs.rust_changed == 'true'`. That is correct today because every cargo step in that job carries the same guard — but it means a future unguarded cargo step would silently fall back to the shared home. An unconditional `Isolate` step would be strictly safer and costs nothing (it is one `echo`).

---

### Operational risk: cold cache. This is the real cost, and it is front-loaded onto a queue that is already the bottleneck

Measured warm baselines, last six successful `main` runs: Check Rust 4–5 min, Backend Unit 13–16 min, Code Coverage 14–16 min, Integration 9–11 min, Windows 12–13 min.

Current cache inventory (`GET /actions/caches`) — this is the useful number:

| key | size | warm | partial-restore (`full match: false`) |
|---|---|---|---|
| `v0-rust-unit-*` | **1.12–1.33 GiB** | 13–14 min | **18–19 min** (7 samples, 2026-08-10) |
| `v0-rust-clippy-*` | 0.63–0.70 GiB | 4–5 min | — |
| `v0-rust-coverage-*` | **0.16–0.18 GiB** | 14–17 min | 14–17 min |

Three consequences:

- **`coverage` barely loses anything.** At 0.18 GiB its entry is essentially just the pruned registry — its `target/llvm-cov-target` is not being cached usefully today, which independently corroborates #3154's "it rebuilds `aws-lc-sys` from scratch every run". Expected cold delta: **+1–3 min** (registry download/extract). Comfortably inside `timeout-minutes: 35`.
- **`check-rust` is fine.** Warm 4–5 min plus the workflow's own measured ~9 min cold dep compile ⇒ ~13–14 min against a 20-min timeout. ~6 min margin.
- **`test-backend-unit` is the problem.** A *partial* restore already costs 18–19 min. A `prefix-key` bump is not a partial restore — the restore-keys fallback cannot match either, so it is a true cold: ~635 deps compiled from scratch with `RUSTC_WRAPPER: ""` (no sccache). Estimate **22–27 min against `timeout-minutes: 25`.** I could not find a genuinely cold historical run to pin this exactly, so treat it as an estimate — but it straddles the timeout, and the failure mode is the expensive one: the job holds a runner slot for the full 25 minutes and *then* fails.

Queue impact. Actions caches are ref-scoped: a PR can read `refs/heads/main`'s caches but not another PR's. So until this merges **and** a `main` CI run completes and saves under `v1-rust`, **every** queued PR job is cold. With 9 runs queued against `maxRunners: 6`, the added CPU-time is roughly 9 × (check-rust +9 min, unit +9 min, coverage +2 min) ≈ **3 runner-hours ≈ +30–45 min of extra queue drain**, plus whatever is lost to `test-backend-unit` timeouts. That is worth paying once for a flake that has cost ≥6 job failures in 7 days — but it should be a decision, not a surprise.

Mitigations, cheapest first:

1. **Raise the two timeouts in this PR** (`test-backend-unit` 25→40, `coverage` 35→45). One line each, revert later. This is the only item I would call a blocker.
2. **Merge, then let `main`'s CI finish before rerunning PRs** so PR runs restore from `main`'s `v1-rust` entries instead of each paying full cold.
3. **Purge the `v0-rust-*` entries immediately after merge.** The 14 Rust entries currently total ~10.6 GiB against GitHub's 10 GB per-repo budget, so v1 entries will start evicting things by LRU while v0 corpses are still occupying the budget. `gh api -X DELETE "repos/artifact-keeper/artifact-keeper/actions/caches?key=<v0-key>"`.
4. **Optional, and the cheapest transition of all:** the `RUSTC_WRAPPER: ""` overrides in `check-rust` / `test-backend-unit` are labelled "Revert once the runner is restored", and sccache now looks healthy (job 93341706470, 2026-08-10: `Cache location Local disk: "/cache/sccache"`, 0 read/write/total errors). The shared 50 GiB sccache still holds objects for all these deps, so re-enabling it would make the cold compile nearly free. That is a separate change and should not ride along in this PR — but it is the natural follow-up, and it would also make the `prefix-key` cost a non-event.

### Smaller risks

- **Permanent, not transitional:** `test-backend-integration` and `build-windows-dev` have no rust-cache and previously read a warm shared registry. They will now download and extract the full graph on **every** run, forever. The PR estimates 1–2 min; 1–3 min is realistic. Correct trade, and the PR is honest about it — just note it does not go away after the transition.
- **Disk — UNVERIFIED RISK.** `RUNNER_TEMP` is on the `work` `emptyDir`, which is backed by node ephemeral storage. The `dind` block in the same values file records that the root disk is ~70 GB and that filling it previously caused kubelet disk-pressure eviction of *other* pods — which is why the docker roots were deliberately moved to `/home`. Each pod now keeps its own extracted registry there (order 1.5–3 GB), so up to ~9–18 GB extra at 6 concurrent pods. `target/` already dominates that volume so this is incremental, and no failure I read showed ENOSPC. But I have no cluster access to check headroom or whether `/var/lib/kubelet` is on the root disk — flagging it as unverified rather than guessing. There is no `ephemeral-storage` limit on the pod spec to catch it.
- **Build/test split, low risk.** `Build instrumented test binaries` omits `DATABASE_URL` / `AK_TESTS_REQUIRE_DB` / `JWT_SECRET`, which `Generate coverage report` sets. Cargo fingerprints do not include arbitrary env vars, and the only `build.rs` in the workspace (`backend/build.rs`) declares `rerun-if-env-changed` for `GIT_SHA` and `SOURCE_DATE_EPOCH` only, so this should not trigger a rebuild in the second step. If CI shows the second step recompiling, the split is doubling work and the env blocks should be made identical.
- **`::error::` annotation semantics — checked, correct.** Default shell is `bash -e`; `if ! cmd` does not trip `errexit`, and the explicit `exit 1` fails the step. The original exit code (101) is replaced by 1, which is immaterial since the log carries the diagnosis. The annotation title contains no comma and no `::`, so GitHub will not mis-parse the property list. Moving `cargo llvm-cov clean --workspace` into the build step preserves the previous ordering exactly (`show-env` → `clean` → build → run), and the comment correctly warns against a second `clean` between build and run. `Upload coverage artifact` runs with `if: always()` and no `if-no-files-found`, so a missing `lcov.info` warns rather than adding a second confusing failure.
- **Nothing else in the repo assumes `CARGO_HOME=/cache/cargo-home`** — no `.cargo/config.toml`, no registry source replacement anywhere in the workflows or in the ARC repo, so the private home does not lose a mirror or auth config.

### Validation performed

YAML validated with Python `yaml.safe_load` (parses; 14 jobs; step order and `if:` guards dumped and inspected per job). `actionlint` is **not** installed on this host, so the workflow was not linted for expression/context errors — the repo does ship `.github/actionlint.yaml` declaring the `ak-*` self-hosted labels, and the diff introduces no new labels, expressions, or action refs, so the residual risk there is low but it is not zero and I am not claiming otherwise.

brandonrc:
Applied the review's blocker fix in `9e3dc37`, and closed one of the two unverified risks with cluster access the reviewer did not have.

## Blocker fixed: cold-cache timeouts

`test-backend-unit` 25 → 40, `coverage` 35 → 45. The reasoning in the review holds: a *partial* restore already costs 18–19 min across 7 samples, so a true cold build straddles the 25-minute limit — and the failure mode burns a full runner slot before failing. Both bumps are transitional headroom; once a `main` run saves under the `v1` prefix, PR jobs warm again.

Left `check-rust` at 20 (est. 13–14 min cold) and the others unchanged.

## Unverified risk #1 — ephemeral storage: RESOLVED, not a concern

The review flagged that `RUNNER_TEMP` lives on the `work` emptyDir → node ephemeral storage, against a values file recording a ~70 GB root disk and a prior disk-pressure eviction. Checked directly on the node:

```
alloc-ephemeral: 1535140420829   (~1.53 TB)
/dev/mapper/rl00-root   70G   57G   13G  82%  /
/dev/mapper/rl00-home  1.8T  437G  1.4T  25%  /var/lib/kubelet
DiskPressure: False
```

Kubelet — and therefore pod ephemeral storage — lives on `/var/lib/kubelet`, which is the **1.8 TB `rl00-home` volume with 1.4 TB free**, not the 70 GB root. The estimated 9–18 GB of per-pod extracted registries at 6 pods is negligible against that. The 82%-full root is real but unrelated: no pod storage is on it.

## Root cause confirmed at the infrastructure layer

The diagnosis is not just log-inferred — it is visible in the live pod spec:

```
env:    CARGO_HOME  = /cache/cargo-home
mount:  runner-cache -> /cache
volume: hostPath /home/runner-cache      <- shared by ALL runner pods
```

`/home/runner-cache/cargo-home` is a real 1.3 GB directory containing `registry/{cache,index,src}`, shared by every concurrent runner in the `ak-ci-runners` scale set (live cap: 6). That is exactly the shared-writer hazard this PR removes.

It also independently confirms claim 2: the scale set sets `RUSTC_WRAPPER: sccache` at the pod level, and `ci.yml` overrides it to `""` in four places — so sccache is genuinely not in the path for the failing jobs.

## Remaining unverified risk

`actionlint` is unavailable here as well; the workflow was validated with `yaml.safe_load` only. The diff adds no new labels, expressions, or action refs, so residual risk is low — but stating it rather than implying it was checked.

## Two follow-ups worth doing separately

1. **Correct the timestamp in the code comment.** The review found the two coverage pods' first errors are **0.6 s** apart, not 13 s (the 00:09:33 figure is pod A's *third* error), and that the mutator in that instance was a concurrent **restore**, not the post-save clean. Worth fixing in the comment so the incident stays reproducible for the next person.
2. **`coverage`'s Isolate step is gated on `rust_changed == 'true'`.** Correct today, but an unconditional `echo` would be strictly safer.

Neither blocks merge.

Issue #794, opened 2026-04-19
Permissions table CRUD is wired up in UI/API but never enforced — silent security gap


nikitatsym:
## Summary

The `permissions` table (migration `018_groups_permissions.sql`) has a fully working CRUD story — admins can create/read/update/delete permissions via `/api/v1/permissions` and the Admin → Permissions page in the web UI. However, the backend **never checks these records** when handling requests to repositories or artifacts. Any authenticated user can read/write any repository they aren't otherwise blocked from, regardless of what the `permissions` table says.

This is a silent security gap: the UI looks like it's protecting resources, but it isn't.

## Reproducer

```
1. Create group "readers", add user U
2. POST /api/v1/permissions
   { "principal_type":"group", "principal_id":"<readers>",
     "target_type":"repository", "target_id":"<repo>",
     "actions":["read"] }
3. Authenticate as U, PUT /maven/<repo>/g/a/1.0/a-1.0.jar
   -> 200 OK  (expected 403)
```

## Root cause

- `repo_visibility_middleware` in `api/middleware/auth.rs` only checks `is_public`, write-requires-auth, and API-token `allowed_repo_ids`. It never queries the `permissions` table.
- Format handlers rely on the middleware; `/api/v1/repositories/{key}/...` handlers rely on per-handler `require_auth` / `require_scope` / `require_repo_access` helpers. Neither path consults `permissions`.
- Authorization today is spread across middleware, ad-hoc helpers in handlers, and implicit `is_public` checks. There's no single place that answers "can this principal do X on this resource?", which makes it easy for new code paths (gRPC, background jobs, new format handlers) to bypass whatever rules do exist.

## Why this matters

This is worse than not having the feature at all. Admins see a working UI, configure group permissions, and reasonably conclude that the repository is protected. The UX actively communicates a security guarantee that the backend does not honour.

## Design questions (before implementation)

1. **Enforcement point — service layer?** Moving the permission check into the service layer (`RepositoryService`, `ArtifactService`, etc.) would give a single choke point that covers format handlers, the API, gRPC, and any future entry point. The existing middleware / handler helpers could be kept for coarse auth (is-authenticated, token scope) but the "can this principal do X on this repo" question would be answered in one place. Is that the direction you'd like?
2. **Single or dual permission model?** Both `permissions` and `role_assignments` exist; #754 used `role_assignments` for the repo-list filter. Which is the canonical system going forward, or are they expected to coexist?
[...]

PR #3250, opened 2026-08-10
fix(auth): remove stale 'enforcement is NOT active' startup warning (#3185)

### LINKED ISSUE #3185: Startup warning says permission enforcement is off, but #794 says it was fixed
### Description

I just upgraded to 1.7.1 (from 1.6.0) and noticed the startup log says permission enforcement is off, but the issue it links to was closed months ago as fixed. Filing this so someone with more context can tell which one's actually true.

### Environment Information

- Deployment type: Kubernetes (GKE), via Helm chart.
- Operating System: Linux (container).
- Architecture: amd64.
- Database: PostgreSQL 18 (CloudSQL).
- Reverse Proxy: nginx-ingress.
- Network setup: public ingress behind a GCP LB with IP allowing list at the ingress.

It's a fresh install.

### Backend Version

1.7.1

### Web UI Version

1.8.0

### Package Format

None

### Expected Behavior

Either the log shouldn't say enforcement is off (if #794 actually got fixed), or if there's a real remaining gap, it should be clear what part of enforcement it covers.

### Observed Behavior

Every boot, with at least one row in permissions, shows:

> Found N permission rule(s) in the database, but enforcement is NOT active. Permission rules created via /api/v1/permissions are stored but not consulted during request authorization. This will be addressed in a future release. See https://github.com/artifact-keeper/artifact-keeper/issues/794

But #794 is closed, with a comment saying it was fixed across #824/#826/#827/#819 and enforcement is active.

I tried to just test it myself instead of guessing, so I gave one service account a read grant on a repo via /api/v1/permissions, made a second account with zero grants, and the second one still gets denied pulling from that repo while the first one works fine. So enforcement looks like it's actually doing something here, at least for this case. It makes me think the warning is just stale text that got left behind, but I can't be 100% sure I'm not hitting some other code path that happens to gate it correctly for unrelated reasons.

### Steps to Reproduce

1. Create service account A, grant it read on repo R via POST /api/v1/permissions.
2. Create service account B, no grants.
3. Try pulling a package from R with each account's token.
4. Account A works, B gets denied.
5. Restart the backend, check logs.

### Logs

```shell
Found 1 permission rule(s) in the database, but enforcement is NOT active.
Permission rules created via /api/v1/permissions are stored but not
consulted during request authorization. This will be addressed in a future
release. See https://github.com/artifact-keeper/artifact-keeper/issues/794
permission_rules=1
```

### Additional Context

Related to #794 (closed), #816, #817, #818, #819 (the PRs/issues the closing comment credits for the fix). If the warning is just stale text left in main.rs after those landed, probably worth deleting it. It's actively confusing to read it in prod logs. If there's a real gap, would help to know exactly what it does and doesn't cover, since the test didn't turn up one.

### Pre-submission Checklist

brandonrc:
## What this fixes

Every boot with at least one row in `permissions`, the backend logged:

> Found N permission rule(s) in the database, but enforcement is NOT active. Permission rules created via /api/v1/permissions are stored but not consulted during request authorization. This will be addressed in a future release. See #794

That warning is **false**. It was inert log text left in `main.rs` after the enforcement work landed, and it misleads operators about their security posture — the concern the reporter raised.

## Which world we are in: the warning is STALE; enforcement IS active

I established this from the enforcement code, not the warning text or #794's closing comment:

- **Native-protocol / format-handler path** — `repo_visibility_middleware` (`backend/src/api/middleware/auth.rs`) gates writes through `PermissionService::check_repository_action` (deny-by-default when a rule exists) and reads through `has_any_rules_for_target` + `check_permission`.
- **REST path** — `require_repo_action` (`backend/src/api/handlers/repositories.rs`) calls the same `check_repository_action`.
- **`check_repository_action`** (`backend/src/services/permission_service.rs`) runs a real SQL decision over the `permissions` table (direct + group + inherited-project grants), with a `CASE` that is authoritative-and-deny-by-default once any applicable rule exists.
- The runtime posture endpoint `/api/v1/system/config` already reports `permissions.enforcement_enabled: true` (`backend/src/api/handlers/system_config.rs`), directly contradicting the startup warning.

So a permissions rule **is** consulted during request authorization. The reporter's own black-box test (grantee allowed, ungranted account denied) matches this; the warning is the thing that is wrong.

I did not take the mere presence of a gate-shaped call as proof — `check_repository_action` runs an actual grant-resolving query and the middleware routes writes through it deny-by-default.

## What I changed

- Removed the stale warning block in `backend/src/main.rs`, replacing it with a short comment pointing at the enforcement paths and the regression test (so nobody re-adds it).
- Added a DB-backed regression test `permission_service::tests::permission_rule_is_consulted_during_authorization` that pins the enforcement the warning denied, so the message and behaviour cannot drift apart again. If enforcement is ever silently regressed (rules stored but not honoured — the warning becoming true again), this test goes red instead of a log line quietly becoming accurate.

## Scope note

This is a log-message correction backed by an enforcement regression guard. I did **not** find a live authorization gap on the surfaces the warning covers — enforcement is genuinely active on both the native and REST paths. No enforcement code was changed.

Fixes #3185

### DIFF
--- a/backend/src/main.rs
+++ b/backend/src/main.rs
@@ -256,26 +256,19 @@ pub async fn run_server(shutdown_token: Option<CancellationToken>) -> Result<()>
     let peer_id = init_peer_identity(&db_pool, &config).await?;
     tracing::info!("Peer identity: {} ({})", config.peer_instance_name, peer_id);
 
-    // Warn when permission rules exist but enforcement is not yet active (#794).
-    // This makes the gap visible in server logs so administrators do not
-    // assume their rules are protecting anything.
-    match sqlx::query_scalar::<_, i64>("SELECT COUNT(*) FROM permissions")
-        .fetch_one(&db_pool)
-        .await
-    {
-        Ok(count) if count > 0 => {
-            tracing::warn!(
-                permission_rules = count,
-                "Found {} permission rule(s) in the database, but enforcement is NOT active. \
-                 Permission rules created via /api/v1/permissions are stored but not consulted \
-                 during request authorization. This will be addressed in a future release. \
-                 See https://github.com/artifact-keeper/artifact-keeper/issues/794",
-                count,
-            );
-        }
-        Ok(_) => {}  // no rules, nothing to warn about
-        Err(_) => {} // table may not exist on old schema, ignore
-    }
+    // Note: a startup warning claiming "permission rules are stored but not
+    // consulted during request authorization (#794)" used to live here. That
+    // gap was closed — fine-grained `permissions` rules are now enforced on
+    // both the native-protocol path (`repo_visibility_middleware` in
+    // `api/middleware/auth.rs`, via `PermissionService::check_repository_action`
+    // for writes and `has_any_rules_for_target`/`check_permission` for reads)
+    // and the REST path (`require_repo_action` in `api/handlers/repositories.rs`).
+    // `/api/v1/system/config` reports `enforcement_enabled: true` accordingly.
+    // The warning was left behind after #817/#824/#826/#827/#819/#2603 landed and
+    // was misleading operators (#3185), so it has been removed. The regression
+    // test `permission_service::tests::permission_rule_is_consulted_during_authorization`
+    // pins the enforcement so this message and the behaviour cannot drift apart
+    // again.
 
     // Initialize WASM plugin system (T068)
     let plugins_dir =
--- a/backend/src/services/permission_service.rs
+++ b/backend/src/services/permission_service.rs
@@ -2054,42 +2054,30 @@ mod tests {
             "the just-inserted permissions rule must be observable"
         );
 
-        // Negative: the principal the rule does NOT name is denied. This is the
-        // enforcement the warning claimed did not exist. A rule-less private
-        // repo would deny `other_user` too, so the negative alone is not proof —
-        // hence the positive control below in the SAME fixture.
-        assert!(
-            !service
-                .check_repository_action(other_user, repo_id, "write", false)
-                .await
-                .expect("write check must reach the DB"),
-            "a principal without a matching rule must be denied write (rule IS consulted)"
-        );
-        assert!(
-            !service
-                .check_permission(other_user, "repository", repo_id, "read", false)
-                .await
-                .expect("read check must reach the DB"),
-            "a principal without a matching rule must be denied read once rules exist"
-        );
-
-        // Positive control (mandatory): the granted principal succeeds in the
-        // same fixture, so the negatives above cannot be satisfied by a gate
-        // that simply denies everyone.
-        assert!(
-            service
-                .check_repository_action(granted_user, repo_id, "write", false)
-                .await
-                .expect("write check must reach the DB"),
-            "the granted principal must be allowed write (positive control)"
-        );
-        assert!(
-            service
-                .check_permission(granted_user, "repository", repo_id, "read", false)
-                .await
-                .expect("read check must reach the DB"),
-            "the granted principal must be allowed read (positive control)"
-        );
+        // Collect every verdict BEFORE tearing down, then assert after. A
+        // failing assert must not skip cleanup: this suite runs
+        // process-per-test against a SHARED database, so leaked rows surface
+        // as unrelated failures elsewhere (the #3129 class).
+        let other_write = service
+            .check_repository_action(other_user, repo_id, "write", false)
+            .await
+            .expect("write check must reach the DB");
+        let other_read = service
+            .check_permission(other_user, "repository", repo_id, "read", false)
+            .await
+            .expect("read check must reach the DB");
+        let granted_write = service
+            .check_repository_action(granted_user, repo_id, "write", false)
+            .await
+            .expect("write check must reach the DB");
+        let granted_read = service
+            .check_permission(granted_user, "repository", repo_id, "read", false)
+            .await
+            .expect("read check must reach the DB");
+        let granted_delete = service
+            .check_repository_action(granted_user, repo_id, "delete", false)
+            .await
+            .expect("delete check must reach the DB");
 
         // Scope teardown to this fixture's own rows only (#3129: no
         // cluster-wide deletes on a shared database).
@@ -2106,5 +2094,42 @@ mod tests {
             .bind(repo_id)
             .execute(&pool)
             .await;
+
+        // These two are CONTROLS, not the discriminating assertions — the
+        // comments here previously had it the other way round. `other_user`
+        // has no applicable rule, so it never takes the rule arm: it falls
+        // through to the role-assignment `ELSE` (see the `CASE` around
+        // permission_service.rs:216-224) and is denied for want of a role.
+        // Mutating the rule arm to `WHEN false` leaves BOTH of these green.
+        assert!(
+            !other_write,
+            "a principal with neither a rule nor a role must be denied write"
+        );
+        assert!(
+            !other_read,
+            "a principal with neither a rule nor a role must be denied read"
+        );
+
+        // THESE gate the claim. `granted_user` holds no role assignment, so
+        // the only way it can be allowed is if the `permissions` rule was
+        // actually consulted. Mutating the rule arm to `WHEN false` fails
+        // exactly here.
+        assert!(
+            granted_write,
+            "the granted principal must be allowed write — the rule IS consulted"
+        );
+        assert!(
+            granted_read,
+            "the granted principal must be allowed read — the rule IS consulted"
+        );
+
+        // The rule is AUTHORITATIVE for the principal it names, not merely
+        // additive: an action it does not carry is denied. Without this, a
+        // resolver returning "any applicable rule ⇒ allow" would satisfy
+        // everything above.
+        assert!(
+            !granted_delete,
+            "the granted principal holds read+write only; delete must be denied"
+        );
     }
 }
--- a/backend/src/services/permission_service.rs
+++ b/backend/src/services/permission_service.rs
@@ -2012,4 +2012,99 @@ mod tests {
             tx.rollback().await.expect("rollback migration fixture");
         }
     }
+
+    // -----------------------------------------------------------------------
+    // #3185: enforcement-vs-message drift guard.
+    //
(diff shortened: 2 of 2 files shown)


brandonrc:
## Independent validation — verdict: **the deletion is correct, the justification is over-broad**

Reviewed at head `f5c4022` in a fresh clone with its own Postgres (migrations 001–181). I tried to refute the premise rather than confirm it.

**The warning is genuinely stale and removing it is right.** Independent corroboration the PR does not cite: `CHANGELOG.md:1006` ("Permission gap warnings" #794/#820) and `CHANGELOG.md:1025` ("Permission enforcement gap closed" #794/#816/#817/#818/#819/#820/#824/#826/#827/#828) are entries in the **same release** — the warning was stale on arrival, not stale by later drift. And `check_repository_action` demonstrably consults the `permissions` table (mutation-proved below).

**But "enforcement IS active" is not a whole-system property here, and the PR states it as one.** The single most-used read path in the product — `docker pull` — consults neither `permissions` nor role assignments. Details in Finding 1. That does not resurrect the warning (a blanket "rules are not consulted" is still false and still misleading), but the replacement comment and PR description should describe coverage rather than assert a global fact, which is also literally what the reporter asked for as the second branch of their expected behaviour.

---

## Claims, verified independently from the enforcement code

**1. `repo_visibility_middleware` routes writes deny-by-default through `check_repository_action`, reads through `has_any_rules_for_target` + `check_permission` — CONFIRMED, but its coverage is narrower than the comment implies.**
Writes: `backend/src/api/middleware/auth.rs:1942-1953`. Reads: `auth.rs:1963-2014`. Fail-closed on DB error (503) on both arms.
Coverage caveat: the middleware has **exactly one** attach site, `backend/src/api/routes.rs:109-112`, covering the `format_routes` router built at `routes.rs:64-108`. **OCI/Docker `/v2` is registered at `routes.rs:145-146`, before `.merge(format_routes)` at `routes.rs:148`, so it is not under this layer at all.** `oci_v2::router()` (`backend/src/api/handlers/oci_v2.rs:9714-9730`) attaches only `DefaultBodyLimit`. There is a *third* seam the PR comment does not mention: OCI writes self-gate via `require_oci_repo_write_access` → `check_repository_action` (`oci_v2.rs:550-577`).
Two by-design carve-outs inside the middleware that also mean "rule not consulted": public-repo reads skip the ACL entirely (`public_read_satisfies_acl`, `auth.rs:1688-1690`, applied at `auth.rs:1990`), and admins skip the whole block (`auth.rs:1915-1916`).

**2. REST `require_repo_action` calls the same `check_repository_action` — CONFIRMED, but it is not "the REST path".**
`backend/src/api/handlers/repositories.rs:154-177`. Call sites exist in only three files: `repositories.rs:7251` (`authorize_generic_upload`), `repositories.rs:8464` (`delete_artifact`), `upload.rs:239/362/532`, `gitlfs.rs:314`.
The dominant REST repo gate is `require_repo_write_access` (`repositories.rs:110-128`, 29 call sites) which **does not consult `permissions` at all** — it is token scope + `user_can_access_repo` (role assignment) — and returns `Ok` early for public repos. Some handlers correctly stack a fine-grained check on top (`set_cache_ttl` inline `check_permission("admin")` at `repositories.rs:1537-1545`; `require_repo_admin` at `repositories.rs:2190`, `2254`, `9108`, `9310`, `9363`, `security.rs:1284`). Others do not: `repository_labels.rs:185/234/274` and `email_subscriptions.rs:218` are role-assignment-only.

**3. `check_repository_action` runs a real grant-resolving SQL `CASE` over direct + group + inherited-project grants — CONFIRMED. "Deny-by-default once an applicable rule exists" needs one correction.**
`backend/src/services/permission_service.rs:162-237`. The `CASE` is authoritative once a rule applies **to that principal**, not once a rule exists on the target (`permission_service.rs:216-224`). With no applicable rule the `ELSE` arm falls back to role assignments — deliberate and documented at `permission_service.rs:153-161` ("principal-scoped transition"), so the first rule on a repo does not drop every legacy principal. Consequence worth stating plainly: it is **additive per principal**, so a caller holding a global role assignment carrying the action is unaffected by anyone else's restricting rules.
On the *no rule at all* case you asked about: it does not fall open. Writes still require a role carrying the action (`permission_service.rs:222-224`; `auth.rs:1942-1953`). Rules-less **private** reads on the native path require a role assignment via a bare `EXISTS role_assignments` query (`auth.rs:2032-2042`) — note that query does not check the role actually carries `read`, but it is not open.
Project inheritance is consistent across all three resolvers (`permission_service.rs:285-291`, `316-321`, `489-494`), so a project-only grant does not create a has-rules/no-actions false deny.

**4. `/api/v1/system/config` reports `enforcement_enabled: true` — CONFIRMED.** `backend/src/api/handlers/system_config.rs:267`, hardcoded (not derived), alongside a real `rules_exist` probe at `system_config.rs:259-263`. It does contradict the warning.

**5. The warning fired on `COUNT(permissions) > 0` regardless of actual enforcement — CONFIRMED.** The deleted block, `backend/src/main.rs:259-278`.

---

## The new test — not hollow, but it does not pin what the PR says it pins

Run with `DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` against a dedicated migrated Postgres.

- Baseline: **PASS in 0.345s** — a real DB run (the early-return skip is ~0.04s).
- Fixture preconditions check out: `create_repo` relies on `is_public NOT NULL DEFAULT false` (`backend/migrations/003_repositories.sql:21`), so the repo really is private; `create_user` (`test_db_helpers.rs:636`) creates no role assignment, so the positive control cannot be satisfied by a role.
- **Faithful mutation of the regression the warning describes** — rules stored but not consulted: in `check_repository_action`, `WHEN EXISTS (SELECT 1 FROM applicable_rules)` → `WHEN false`, so the decision falls back to role assignments only. Result: **FAIL** at `permission_service.rs:2079`, *"the granted principal must be allowed write (positive control)"*. The test genuinely gates the claim at the service layer.

Three things to fix, none of which change the verdict:

**(a) The test's own comments have the discriminating assertion backwards.** Under that mutation **both negative assertions passed**. `other_user` holds no applicable rule, so it takes the `ELSE` role-fallback arm — those asserts are green even with the `applicable_rules` CTE deleted outright. The comment calls the negative *"the enforcement the warning claimed did not exist"* and demotes the positives to *"positive control"*; empirically it is exactly the reverse. Worth correcting so the next person does not delete the positives as redundant.

**(b) The test never touches either enforcement seam.** It calls `PermissionService` directly. Delete the `check_repository_action` call from `repo_visibility_middleware` (`auth.rs:1944`) or from `require_repo_action` (`repositories.rs:154-177`) — the precise regression that would make the warning true again at the HTTP layer — and this test stays green. The PR says it "pins the enforcement so this message and the behaviour cannot drift apart again"; it pins the service function. Please either add a seam-level assertion or scope the claim.

**(c) The "authoritative" property in claim 3 is untested, and is free in this fixture.** `granted_user` holds `{read, write}`, so:
```rust
assert!(!service.check_repository_action(granted_user, repo_id, "delete", false).await.unwrap(),
        "a rule granting read+write must not confer delete");
```
I added exactly this and ran it: **passes on head**. It is the only assertion in the test that would exercise the deny-by-default arm rather than the role-fallback arm.

**(d) Teardown hygiene.** Cleanup is post-assert best-effort deletes (`permission_service.rs:2094-2107`). Any assert failure skips it and leaks a repository, two users and a `permissions` row into a shared CI database. The sibling DB test in this same module uses `tx.rollback()` (`permission_service.rs:2011`); please match it.

Revert-faithfulness note: the production change is inert log text, so a revert of it fails nothing — the mutation approach was the right call, and the 95 added lines in `permission_service.rs` are entirely inside `#[cfg(test)] mod tests`, so production bytes are unchanged.

---

## Finding 1 (needs its own issue) — `permissions` rules are not consulted on OCI/Docker `/v2` reads

Verified from source, not inferred:
- `/v2` is not under `repo_visibility_middleware` (`routes.rs:145-146` vs `routes.rs:148`).
- `handle_get_blob` gates on `if is_anon && !repo.is_public` and nothing else (`oci_v2.rs:4596-4598`), then only token-scope checks (`enforce_scan_pull_scope` 4604, `enforce_token_repo_scope` 4610).
- `handle_get_manifest` is identical (`oci_v2.rs:7889-7891`, 7897, 7903).
- `authenticate_oci` (`oci_v2.rs:316-324` → `332-370`) authenticates only; it performs no repository authorization.
- Writes/deletes **are** gated (`require_oci_repo_write_access`, `oci_v2.rs:550-577`).

**Failure scenario:** private OCI repository `team-a/app`. Account B has zero `permissions` rules and zero role assignments on it. `docker login` with B's credentials, `docker pull <host>/team-a/app:tag` → served. A repo-scoped API token is still blocked by `enforce_token_repo_scope`, but an ordinary JWT session carries `allowed_repo_ids = AccessScope::Admin`, which satisfies that check unconditionally — so this is not mitigated for interactive/CI logins.

This is the reporter's own experiment run against Docker instead of pip, and it returns the opposite answer. I could not find an existing issue for it. It is pre-existing and not caused by this PR, but it is the reason the unqualified wording should not ship.

Lesser surfaces where a stored rule is likewise not consulted: `repository_labels.rs:185/234/274` and `email_subscriptions.rs:218` (role-assignment only, and public repos short-circuit in `require_repo_write_access`), and three repository-config GETs that take no auth extractor at all — `get_cache_ttl` (`repositories.rs:1600-1603`), `list_pypi_tracks` (`repositories.rs:2127`), `get_routing_rules` (`repositories.rs:9234-9237`).

---

## Requested changes

1. Rewrite the `main.rs` replacement comment and the PR description to state **coverage**, not a global fact — and name all three seams, including OCI `/v2` self-gating at `oci_v2.rs:550` (it is not under the middleware, and the current comment reads as if all native protocols are).
2. File the OCI `/v2` read gap as its own issue and reference it from the linked issue, so the closing answer to "what it does and doesn't cover" is on the record instead of implied.
3. Test: add the `delete` deny-by-default assertion; fix the comments that invert which assertion discriminates; switch teardown to `tx.rollback()`; and either add a seam-level assertion or drop the "pins the enforcement" claim to "pins the service-layer decision".

Nothing here argues for keeping the warning. It was a blanket false statement and it should go.

Issue #3249, opened 2026-08-10
Analytics daily storage snapshot excludes OCI blob and proxy-cache bytes (same gap as #3134)


brandonrc:
## Description

`AnalyticsService::capture_daily_snapshot` (`backend/src/services/analytics_service.rs:112`) writes the daily `storage_metrics` row from a raw `artifacts` aggregate:

```sql
(SELECT COALESCE(SUM(size_bytes), 0) FROM artifacts WHERE is_deleted = false)
```

This is the same accounting gap as #3134 (dashboard `total_storage_bytes`), on the analytics surface: OCI layer/config blobs live in `oci_blobs`, not `artifacts` (only manifests land there), and proxy-cached bytes live in `proxy_cache_artifacts`. A docker-heavy instance therefore records a daily storage snapshot that misses the overwhelming majority of its stored bytes, and every growth figure derived from `storage_metrics` (`GrowthSummary`: `storage_bytes_start/end`, `storage_growth_bytes`, `storage_growth_percent`) under-reports accordingly. Legacy backfilled `proxy-cache/%` rows in `artifacts` are also still counted here (the per-repo predicate from #2218 excludes them).

`total_artifacts` in the same snapshot has the parallel manifest-only blind spot.

## Suggested fix

Use the same source #3134 moved the dashboard to, so there is exactly one accounting implementation:

```sql
(SELECT COALESCE(SUM(hosted_bytes + proxy_bytes + oci_bytes), 0)::BIGINT
   FROM repository_usage_ledger)
```

The ledger components are maintained by the migration-182 triggers and trued up by `reconcile_all_usage_ledgers` from the authoritative 3-way source sums, so the snapshot becomes consistent with the dashboard total and the per-repository storage figures by construction.

Note when fixing: historical `storage_metrics` rows keep the old (under-reported) basis, so growth series will show a one-day step when the corrected accounting lands. Worth a doc comment on `GrowthSummary` rather than any attempt to rewrite history.

Found while fixing #3134; scoped out of that PR because the snapshot upsert is keyed on `CURRENT_DATE` cluster-wide, which needs its own test strategy (a per-date global row cannot be fixture-scoped the way the stats handler deltas were).

Issue #3094, opened 2026-08-03
GET /admin/stats does 7 full-table aggregates; byte totals could read repository_usage_ledger


brandonrc:
`get_system_stats` (`backend/src/api/handlers/admin.rs`) issues seven independent full-table aggregates on every call — `COUNT(*)` over `repositories`, `users`, `download_statistics`, `peer_instances`, `sync_tasks`, a `COUNT(*) + SUM(size_bytes)` over `artifacts`, and (as of #3088) the same over `proxy_cache_artifacts`. Cost scales with total row count, so it degrades on exactly the large catalogs where an admin dashboard is most useful.

The byte totals in particular no longer need to be computed live. `repository_usage_ledger` (migration 171, made authoritative by row-level triggers in #2992 / migration 182) already carries per-repository `hosted_bytes`, `proxy_bytes` and `oci_bytes`, so:

- `total_storage_bytes` → `SELECT COALESCE(SUM(hosted_bytes), 0) FROM repository_usage_ledger`
- `proxy_storage_bytes` → the same over `proxy_bytes`

turning both into O(repositories) index scans instead of O(rows) aggregates. This is the same change #3079 just made for the repository listing (measured there: 56 ms → 0.9 ms on a 220k-row seed), and migration 182's contribution rules are byte-for-byte equivalent to the live sums, so it is not an approximation.

The `COUNT(*)` figures have no ledger equivalent and would need separate treatment (approximate counts from `pg_class.reltuples`, or a counters table) — worth deciding whether exact counts are actually required on a dashboard.

Not urgent, and explicitly **not** a criticism of #3088, which follows the existing pattern in the handler. Filing so the endpoint gets the same treatment the listing did.

Context: #3078 / #3079, #2992 (migration 182).

PR #3251, opened 2026-08-10
fix(admin): include OCI blob bytes in dashboard Storage Used total (#3134)

### LINKED ISSUE #3134: Dashboard total Storage Used excludes OCI blob bytes (docker repos under-reported)
## Description

The dashboard's **Storage Used** tile can under-report total storage by orders of magnitude, because `GET /api/v1/admin/stats` derives `total_storage_bytes` from the `artifacts` table only. OCI layer/config blobs live in `oci_blobs`, not `artifacts` — only manifests land there — so a Docker/OCI repository contributes a few KiB of manifests to the total while holding GiBs of layers.

This is the same class of gap as #3087 (proxy-cached bytes), for the third component that issue did not cover.

`get_system_stats` (`backend/src/api/handlers/admin.rs:816`, assigned at `:869`):

```sql
SELECT
    COUNT(*) as "count!",
    COALESCE(SUM(size_bytes), 0)::BIGINT as "size!"
FROM artifacts
WHERE is_deleted = false
```

Meanwhile the per-repository `storage_used_bytes` returned by `GET /api/v1/repositories` reads all three usage components (`backend/src/api/handlers/repositories.rs:2240`):

```sql
SELECT repository_id, (hosted_bytes + proxy_bytes + oci_bytes)::BIGINT
  FROM repository_usage_ledger
 WHERE repository_id = ANY($1)
```

which is the ledger-backed equivalent of the authoritative 3-way UNION in `RepositoryService::get_storage_usage` (`backend/src/services/repository_service.rs:1612`) — `artifacts` + `proxy_cache_artifacts` + `oci_blobs`. The comment at `repository_service.rs:1603` already spells out why the third branch exists:

> OCI layer/config blobs live in `oci_blobs`, not `artifacts` (only manifests land there), so without the third branch a docker repo reports a few KiB of manifests while holding GiBs of layers.

So the two figures on the same dashboard page are computed from different sets of tables, and the aggregate is the one missing components.

## Environment Information

- Operating System: macOS (host), Linux containers
- Architecture: arm64
- Database: PostgreSQL 16
- Reverse Proxy: none
- Network setup: local

## Backend Version

1.7.0 (`main`)

## Web UI Version

1.7.0

## Package Format

Docker / OCI

## Expected Behavior

The **Storage Used** total on the dashboard reflects all stored bytes, and is consistent with the per-repository **Storage** figures shown in the *Recent Repositories* table on the same page.

## Observed Behavior

On an instance with 4 repositories, the tile reads **3.55 GB** while the repository table on the same screen lists:

| Repository | Format | Type | Storage |
| --- | --- | --- | --- |
| xxx-docker-registry | docker | local | 100.17 GB |
| npm-xxx-registry | npm | local | 6.91 MB |
| npm-registry.npmjs.org | npm | remote | 0 B |
| raw-xxx | generic | local | 3.54 GB |

3.54 GB + 6.91 MB + a few MB of Docker manifests ≈ the 3.55 GB shown, i.e. the docker repository's **100.17 GB of layers is entirely absent from the total**. The tile reports ~3 % of actual usage.

`total_artifacts` (361 here) has the same blind spot: it counts manifests but no OCI blobs.

## Steps to Reproduce

1. Create a local Docker repository.
2. `docker push` an image of non-trivial size (a few hundred MB is enough to make it obvious).
3. Open **Repositories** — the repository's *Storage* column shows the full size including layers.
4. Open the **Dashboard** — the *Storage Used* tile counts only the manifest bytes for that repository.

## Additional Context

**Suggested fix.** Sum the ledger instead of the raw `artifacts` aggregate:

```sql
SELECT COALESCE(SUM(hosted_bytes + proxy_bytes + oci_bytes), 0)::BIGINT
  FROM repository_usage_ledger
```

This makes the total consistent with the per-repo column by construction, and is the O(repositories) shape #3094 asks for, so the two could be fixed together. Note that #3094's suggested mapping (`total_storage_bytes` → `SUM(hosted_bytes)`, `proxy_storage_bytes` → `SUM(proxy_bytes)`) would preserve this bug — `oci_bytes` needs to land somewhere in the reported totals.

If the ledger sum is adopted, `proxy_storage_bytes` becomes a *breakdown* of `total_storage_bytes` rather than a disjoint bucket, so the "reported separately here" doc comment on `SystemStats` (`admin.rs:792`) and anything that adds the two figures need updating. Alternatively keep proxy out of the total and use `SUM(hosted_bytes + oci_bytes)` — either is fine as long as OCI bytes are counted and the semantics are documented.

**Do not** compute the total by summing the repository listing's displayed column: that endpoint deliberately overwrites a virtual repository's figure with the union of its members (`repositories.rs:2256`, #2785), so summing it would double-count every virtual repo. Summing the ledger avoids this, since virtual repositories own no rows in the three source tables.

**Two smaller issues in the same query, for whoever picks this up:**

1. It does not exclude `storage_key LIKE 'proxy-cache/%'` rows, which the per-repo predicate does exclude (#2218). Legacy backfilled proxy objects are therefore counted once in `total_storage_bytes` *and* once in `proxy_storage_bytes`.
2. The web UI drops `proxy_storage_bytes` entirely in its stats adapter (`src/lib/api/admin.ts:22` in artifact-keeper-web), so even the proxy figure that #3087 added is never displayed. Happy to file that separately in the web repo if it's worth tracking on its own.

## Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate — closest are #3087 (proxy bytes, closed, different component) and #3094 (performance of the same handler, open); neither covers the OCI gap.

brandonrc:
Fixes #3134

## What was wrong

`GET /api/v1/admin/stats` derived `total_storage_bytes` from a raw aggregate over `artifacts` only (`backend/src/api/handlers/admin.rs`, `get_system_stats`):

```sql
SELECT COUNT(*), COALESCE(SUM(size_bytes), 0)::BIGINT
FROM artifacts WHERE is_deleted = false
```

OCI layer/config blobs live in `oci_blobs` — only manifests land in `artifacts` — so a docker repository contributed a few KiB of manifests to the dashboard tile while holding GiBs of layers (the reporter's instance showed 3.55 GB against ~104 GB actual). The same query also counted legacy backfilled `proxy-cache/%` rows, double-counting those bytes against `proxy_storage_bytes` (#2218's per-repo predicate excludes them; this aggregate did not).

## The fix

Sum `repository_usage_ledger` instead — the issue's suggested shape:

```sql
SELECT
    COALESCE(SUM(hosted_bytes + proxy_bytes + oci_bytes), 0)::BIGINT as "total!",
    COALESCE(SUM(proxy_bytes), 0)::BIGINT as "proxy!"
FROM repository_usage_ledger
```

The ledger is the accounting implementation the per-repository `storage_used_bytes` column (`repositories.rs`) and quota admission (`check_quota`/`check_quota_locked`) already read; its components are maintained inline by the migration-182 triggers and trued up every 30 min by `reconcile_all_usage_ledgers` from the authoritative 3-way source sums (`artifacts` excluding `proxy-cache/%` keys + `proxy_cache_artifacts` + `oci_blobs`). Reusing it — rather than writing a second global aggregate — makes the dashboard tile equal the sum of the per-repo column by construction and keeps exactly one accounting implementation (the divergent-duplicate-accounting failure mode of #3077/#3179). Virtual repositories own no rows in the source tables, so nothing is double-counted (#2785). It is also the O(repositories) handler shape #3094 asks for; `proxy_storage_bytes` now reads `SUM(proxy_bytes)` from the same scan, and the two remaining raw-table queries are count-only.

### Logical, not physical bytes — and why

The tile now shows **logical** bytes: each repository's blobs counted once per repository (`oci_blobs` is `UNIQUE(repository_id, digest)`, so N tags/manifest references to one blob in a repo count it once; a blob cross-repo-mounted into N repos counts once per repo). Chosen because the issue's stated expectation is consistency with the per-repository Storage column on the same page, and that column — like quota enforcement — is logical. **Physical** bytes (digest deduped per shared backend / per filesystem `storage_path`) answer a capacity-planning question and remain the storage-GC footprint report's concern (`oci_blob_footprint_report`, which reports both figures explicitly). Showing physical here would make the tile disagree with the per-repo column it sits next to, and with what quota charges. The `SystemStats` doc comments now state this, and that `proxy_storage_bytes` is a breakdown of the total rather than a disjoint bucket (the "reported separately" comment was updated per the issue).

## Other surfaces checked for the same omission

- **Quota enforcement: NOT affected.** `check_quota` and `check_quota_locked` (`repository_service.rs`) both read `hosted_bytes + proxy_bytes + oci_bytes` from the ledger, so docker repos cannot exceed quota through the blob gap.
- **Per-repository storage: NOT affected.** Both the listing column (ledger) and `get_storage_usage` (3-way UNION) include `oci_blobs`.
- **Analytics daily snapshot: AFFECTED.** `AnalyticsService::capture_daily_snapshot` sums `artifacts` only, so `storage_metrics` / growth trends have the identical gap. Filed as #3249 (scoped out: the snapshot is keyed cluster-wide on `CURRENT_DATE`, which needs its own test strategy).
- Prometheus metrics (per-upload histograms only) and project quotas (stored, not enforced) have no storage-total accounting to be wrong.
- `total_artifacts` intentionally still counts `artifacts` rows only (for OCI that is manifests); blobs are content-addressed objects, not artifacts. Documented on the field.

## Tests

Both in-file `#[cfg(test)] mod tests` (the `--lib` coverage gate), run with `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1`:

- `test_system_stats_total_storage_includes_oci_blob_bytes` — handler-level before/after delta over a fixture of 4 repos: a docker repo with a manifest + 2 blobs, a second docker repo cross-mounting one of those digests (logical counting: counted once per repo), 2 `manifest_blob_refs` rows for the shared digest (per-reference counting must NOT multiply the figure), a generic repo with a plain hosted artifact (**positive control**: non-OCI storage unchanged), and a remote repo with both a legacy `proxy-cache/%` `artifacts` row (must be excluded) and a `proxy_cache_artifacts` catalog row (must be counted once). Fixture sizes are terabyte-scale integers with a ±10 GB tolerance window and retry, so the assertion is decided by this fixture's own rows: every wrong composition lands ≥ 23 TB outside the window while concurrent suites move KBs–GBs (the #3129 cluster-wide-delta flake class — an exact-equality version of this test demonstrably flaked under a 199-test parallel run; this version passed it).
- `test_ledger_totals_query_empty_table_coalesces_to_zero_db` — the ledger aggregate COALESCEs to 0 over an empty `repository_usage_ledger` (fresh install), since the handler decodes both columns as non-nullable BIGINT; same rolled-back-transaction shape-forcing pattern as the existing proxy-cache test.

### Revert-proof

Production code reverted faithfully to merge-base `01ebcb2` (tests kept). Faithfulness: `git diff 01ebcb2 --shortstat` on the reverted tree = `1 file changed, 251 insertions(+)` — zero deletions, single hunk inside `mod tests`. Result (against a live DB, 0.73 s — not a skip):

```
FAIL api::handlers::admin::tests::test_system_stats_total_storage_includes_oci_blob_bytes
attempt saw total delta 105000000000000 (expected 196000000000000) / proxy delta 61000000000000 (expected 61000000000000)
```

The reverted delta is exactly the artifacts-only sum (manifest 11 TB + hosted 41 TB + legacy 53 TB = 105 TB): all 83 TB of OCI blob bytes and 61 TB of proxy bytes absent from the total, legacy row double-counted. With the fix restored the test passes (196 TB delta), including under the full parallel suite.

`cargo sqlx prepare -- --all-targets` regenerated; `SQLX_OFFLINE=true cargo build --lib --tests`, `cargo fmt --all`, and `cargo clippy --workspace --all-targets -- -D warnings` all clean. No migration.

## Behavior change note

`total_storage_bytes` now includes proxy-cached bytes (it previously excluded them while `proxy_storage_bytes` reported them separately). Consumers that added the two fields would now double-count — the shipped web UI does not (it drops `proxy_storage_bytes` entirely, per the issue's point 2, tracked web-side). On instances upgraded mid-flight the figure is as fresh as the ledger, which the 182 triggers keep exact and the 30-minute reconciler self-heals.


brandonrc (commented):
Reviewed the diff against the tree — the approach is sound and every load-bearing claim checks out. Posting findings below; nothing blocking from my side.

**Verified:**

- The ledger is maintained as described: migration 171 (backfill from the same 3-way source sums, `proxy-cache/%` excluded), migration 182 (row-level AFTER triggers on all three source tables charging the ledger inline, `GREATEST(..., 0)` flooring, self-seeding upsert on growth), plus the 30-minute `reconcile_all_usage_ledgers` safety net in `scheduler_service.rs`.
- "Dashboard total equals the sum of the per-repo column by construction" is literally true — `repositories.rs` reads the same `hosted_bytes + proxy_bytes + oci_bytes` expression from the same table.
- The pre-existing `test_get_system_stats_reports_proxy_cache_totals_db` still holds under the new code path: the 182 triggers charge the ledger inline on INSERT, so no reconcile is needed for its delta assertion to pass.
- `.sqlx` offline cache is consistent: the two deleted query hashes are exactly the two removed queries; the three new ones match the new SQL verbatim.
- The semantic overlap change (`proxy_storage_bytes` is now a breakdown of the total, not disjoint) is documented on the struct fields feeding OpenAPI via `ToSchema`.
- The revert-proof in the PR body is the right shape, and the empty-ledger COALESCE test covers the fresh-install NULL case.

Reusing the single accounting implementation instead of writing a second global aggregate is the right call, and the comment block on the new query documents the semantics (logical per-repo counting, virtual repos, staleness bounds) thoroughly.

**Minor, non-blocking:**

1. Test cleanup on failure: in `test_system_stats_total_storage_includes_oci_blob_bytes`, the `assert!(matched, ...)` fires before the `DELETE FROM repositories` cleanup, so a failing run leaks 4 repos into the shared test DB. Same shape as the pre-existing proxy test, so consistent with the file — but cleanup-before-assert or a guard would be tidier.
2. The 5-attempt retry with ±10 GB slack is well-justified by the TB-scale fixture (every wrong composition lands ≥23 TB out of window), though it does mean a hypothetically intermittent wrong implementation landing in-window would pass silently. Theoretical at these magnitudes — fine as-is.
3. `total_storage_bytes` now includes proxy bytes — a behavior change for external API consumers. The body notes the shipped web UI doesn't add the two fields, good. Consider a `CHANGELOG.md` entry since the repo maintains one.

Note: the main CI checks (Rust, unit, integration, coverage) were still pending at review time — worth letting those go green before merge.

brandonrc (commented):
Reviewed against the tree — the approach is sound and all the infrastructure claims check out (ledger schema in migration 171, migration-182 triggers covering `artifacts`/`proxy_cache_artifacts`/`oci_blobs` with the `proxy-cache/%` exclusion, 30-min `reconcile_all_usage_ledgers` that inserts missing rows for all repos, per-repo `storage_used_bytes` and `check_quota_locked` reading the same ledger, `manifest_blob_refs` not involved in accounting so no per-reference multiplication). The test's expected-delta arithmetic and flake defenses look right. Approving directionally; a few notes, none blocking the core fix:

1. **Semantic change worth calling out in release notes** (non-blocking): `total_storage_bytes` now includes OCI blob bytes (the point of the fix) but also *drops* legacy `proxy-cache/%` artifact rows, and `proxy_storage_bytes` becomes a strict subset of the total. Any external consumer that adds the two fields will now double-count. None exists in this repo, but the API semantics did change.

2. **`analytics_service.rs` `capture_daily_snapshot` divergence widens** (~lines 113-146): it still computes its `total_storage_bytes` as `SUM(size_bytes)` over `artifacts` only. After this PR, `/admin/stats` and the daily `storage_metrics` snapshots will disagree by proxy + OCI bytes. Pre-existing gap, but this PR makes it user-visible. I'd suggest fixing it here — same ledger sum keeps one accounting implementation, per the PR's own rationale. I'm happy to push that fix to this branch.

3. **Minor doc nit**: the ledger table itself is migration 171 (182 adds the triggers + true-up). The code comment in `admin.rs` says "migration-182 triggers", which is accurate — only the PR body's phrasing could be read otherwise. No change needed.

4. **Count/byte scope mismatch** (pre-existing, slightly more visible now): `total_artifacts` counts *all* non-deleted `artifacts` rows including legacy `proxy-cache/%` ones, while the byte total excludes them. The new doc on `total_artifacts` covers the OCI side; consider one clause noting the legacy-row inclusion for completeness.

brandonrc:
Pushed 98be8f41 addressing review note 2: `capture_daily_snapshot` now sums `repository_usage_ledger` (`hosted_bytes + proxy_bytes + oci_bytes`) instead of a raw `SUM(artifacts.size_bytes)`, so the daily `storage_metrics` trend stays consistent with the dashboard total. Note 4's doc clause (legacy `proxy-cache/%` rows still included in `total_artifacts`) and the CHANGELOG consumer note are in d30fd33f. Verified with `SQLX_OFFLINE=true cargo check --lib` and `cargo fmt --check`.

PR #3253, opened 2026-08-10
Feat/vscode openvsx gateway

### LINKED ISSUE #3252: feat(vscode): make Artifact Keeper a governed VS Code extension gateway
### Problem Statement

Artifact Keeper has VS Code-format exact-version pull-through code, but it is not a complete proxy for a real extension gallery and cannot yet enforce organization extension policy or the repository publish-age gate.

The useful outcome is an administrator-controlled extension gateway, not only a cooldown hook. Administrators should be able to apply one baseline across the organization, further restrict individual repositories, review or delay new versions, and have supported clients obtain extension metadata and VSIX assets through Artifact Keeper rather than reaching public gallery/CDN hosts directly.

The current handler:
- serves `GET /api/extensionquery` from local artifact rows instead of the client's POST gallery query;
- serves `latest` from local rows only;
- synthesizes an upstream download path real galleries do not use;
- does not represent target-platform-specific versions;
- does not enforce curation or age-gate policy; and
- does not provide a gallery manifest or documented client setup.

As a result, an operator cannot point a supported VS Code-family client at an AK Remote/Virtual repository, search/install/update through it, and know that allowlists, emergency denies, cooldown, reviews, cache hits, and rewritten asset downloads all use the same server-side decision.

#### Proposed Policy & Gateway Behavior

- Open VSX gallery proxy foundation with gallery manifest, POST extension query, latest/version metadata, and target-platform resolution.
- Rewrite all client-visible metadata and VSIX URLs back through AK.
- Enforce repository age gate (`enforce_repository_age_gate`) during resolution and exact VSIX delivery.
- Instance-wide baseline for VS Code repositories and optional repository restrictions.
- Canonicalize `publisher.extension` case-insensitively.
- Fail closed when an active policy is malformed or cannot be loaded.

#### Why this is distinct from built-in VS Code enterprise controls

Current VS Code provides valuable client controls (`extensions.allowed`, `extensions.autoUpdateDelay`, and `ExtensionGalleryServiceUrl` policy). Those should be documented as defense in depth. 

An AK gateway adds a uniform distribution boundary across explicitly supported clients — including open-source distributions and IDE forks (VSCodium, code-server, Antigravity IDE, Cursor, etc.) that are excluded from proprietary Private Marketplaces. AK applies decisions to fresh/manual installs as well as updates, centralizes cache/review/audit and an emergency kill switch, and reduces developer-machine egress to the AK host for extension discovery and installation.

This scope does not restrict network calls made by an installed extension at runtime.

#### Prior Art

| Project | License | Role & Reuse Category | Notes / Analysis |
|---|---|---|---|
| `microsoft/vscode` (gallery client) | MIT | Authoritative Client Contract | Flags bitmask, FilterType, AssetType/PropertyType strings, and gallery-manifest resource IDs transcribed per official client spec. |
| `dinvlad/vscode-cooldown` | MIT | Logic & Fixture Reference | Small zero-dependency Go proxy (v0.1.5) filtering gallery queries by version age (`lastUpdated`) on Microsoft and Open VSX upstreams. |
| `eclipse-openvsx/openvsx` | EPL-2.0 | Protocol Reference & Interop Peer | Primary upstream protocol reference. Endpoint shapes and OpenAPI 3.1 specs (`/vscode/...`) serve as wiremock fixture standards. Official Docker image used as interop test peer. |
| `gitpod/openvsx-proxy` | AGPL-3.0 | Architectural Reference | Legacy proxy prior art. Highlights POST body hashing for query cache keys and two-tier regular/backup TTL. |
| `coder/code-marketplace` | AGPL-3.0 | Conformance Target | Independent gallery implementation used to validate client-side query behavior. |
| Microsoft Private Marketplace | Proprietary | Feature Prior Art | Establishes policy modes (`None`, `Search-only`, `SearchAndAssets`). Entitlement-gated SKUs create the market gap AK fills for open IDEs. |
| JFrog AI Editor Extension Repos | Proprietary | Feature Prior Art | Validates the "proxy-the-Open-VSX-gallery" architecture (`https://open-vsx.org/vscode/gallery`) and confirms the anonymous read-only client constraint in VS Code. |
| Sonatype Nexus | Proprietary | Negative Prior Art | Declined VS Code extension repository support because POST-driven dynamic galleries do not fit static storage models—justifying AK's dynamic gateway. |

### Use Case

_No response_

### Component

None

### Additional Context

A complete implementation of the Open VSX gateway and age-gate proxy has been completed, real-client dogfooded (VSCodium 1.126 & code-server 4.131), and validated **all green** on integration CI (14,874 DB-backed tests passed).

### Pre-submission Checklist

cazlo:
Fixes #3252

### Summary

This PR turns Artifact Keeper into an organization-controlled gateway for VS Code extensions by adding a protocol-faithful Open VSX gallery proxy, stable asset URL rewriting, and age-gate cooldown enforcement.

- **Open VSX Gallery Proxy:** Implements POST extension query, latest version resolution, gallery manifest, and platform-qualified asset delivery.
- **Asset Egress Control:** Rewrites client-visible metadata and VSIX delivery URLs back through AK so clients do not contact public CDNs directly.
- **Age-Gate Integration:** Calls `enforce_repository_age_gate` during query/latest resolution and exact VSIX delivery for Remote/Virtual repositories.
- **Real-Client Dogfooding:** Tested and validated against contained official VSCodium 1.126 and code-server 4.131 CLI install and update flows.

### Testing & Validation

- **Focused Suite:** 55 DB-backed unit and integration tests passing (`vscode.rs`).
- **Real-Client Egress Test:** Verified VSCodium and code-server CLI update/install flows through AK with external egress blocked.
- **CI Signal:** Full local check/clippy and private CI runs passed cleanly (14,874 DB tests passed).


scripts/dogfood-vscode-openvsx.sh
cazlo, line 1: this should get moved to the release gate tests

docs/vscode-openvsx-gateway.md
cazlo, line 1: this should go away

cazlo:
need to do human qa pass still to see if this misbehaving in real vscode.
also adversarial review not done yet.

since vscode apparently needs this to be public, virtual should just not be supported for this IMO, otherwise that would open all kinds of risk for unintentionally exposing private extensions.

brandonrc:
## Review + remediation pushed

I took this through the full review pipeline: fresh shallow clone at the PR head (`04234d2`), private read-only extracts of both the head and merge-base trees for four independent review passes (security, backend architecture, correctness, test quality), and an isolated Postgres 16 for DB-backed verification. Contradictions between the passes were reconciled against source rather than averaged — one pass reported the `auth.rs` change as correctly narrow; three did not, and reading the merge-base settled it.

**Migration ledger: CLEAN.** This PR adds zero migrations, and `main` ends at `195_curation_attestation_verification.sql`. No duplicate number, so no `_sqlx_migrations` duplicate-key abort at startup. Verified with `git diff <merge-base> --name-only -- backend/migrations/` (empty).

### What is genuinely solid here — this is careful work

Calling this out first because it is the reason the remaining findings are as narrow as they are:

- **The SSRF answer is structural, not a check.** `rewrite_gallery_asset_urls` overwrites `assetUri` / `fallbackAssetUri` / `files[].source` and never retains an upstream absolute URL; delivery reconstructs the target from repository configuration via `openvsx_asset_url`. There is no upstream-supplied URL whose origin needs validating. That is a stronger answer than the `same_upstream_origin` gating in #3130, and it is the right one.
- **`validate_relative_post_endpoint` is exactly right, for exactly the right stated reason** ("this primitive always attaches repository credentials"). It rejects absolute, protocol-relative, rooted, backslash, dot-segment *and* percent-encoded dot-segment targets, with a 14-case table test. I could not get anything past it.
- **The common gallery gate cannot drift.** All six routes funnel through `gallery_upstream`, which does a live `SELECT is_public` rather than trusting the (TTL-stale) repo cache, and rejects private repos even for an authenticated member — which sidesteps both of this codebase's standing traps (`can_access_repo` is token scope, not visibility; `scopes: None` satisfies `require_scope` unconditionally) by never reaching them.
- **Ordering is right in the places it matters.** The age-gate filter runs before URL rewriting, so a blocked version never gets an AK asset URL minted. The delivery gate runs before cache construction, and the `.expect(1)` on the vspackage mock in `gallery_delivery_age_gate_blocks_before_cache_then_honors_manual_review` is genuine proof the post-rejection request re-read neither the cache nor the upstream. That test — full 451 → approve → 200 → reject → 451 with a stable `review_id` — is the best test in the PR.
- **`test_openvsx_extensionquery_posts_and_rewrites_every_asset_url` does a whole-body byte scan**, not a spot check, and verifies what was *sent* upstream (method, path, `Accept`, exact body) as mock matchers. It also re-dispatches the *composed* `assetUri + "/README.md"` through the real router, which is what catches the trailing-slash `//README.md` class.
- **Both age-gate seams fail closed**, and `gallery_last_updated` deliberately returns `None` for malformed timestamps rather than defaulting to "old enough".

### What I changed and why

**1. `auth.rs` — narrowed the anonymous-write exemption back to the gallery query alone (HIGH).**

The diff changed `let is_write = is_write_method(request.method())` to `... && !non_mutating_post`. But `is_non_mutating_format_post` matches **three** path families, not one — it also matches `/lfs/<key>/objects/batch` and `/conan/<key>/v2/users/authenticate`. So the #508 anonymous-write 401 was removed from git-LFS and Conan on every public repository, and the base doc comment that guaranteed this could not happen ("this exemption never loosens the anonymous contract") was deleted in the same hunk. No test at any layer pinned the LFS or Conan behaviour, so the change was silent.

Neither is an upload hole — `gitlfs.rs` self-gates `operation: "upload"` with `require_auth_basic` + `require_repo_action(..., "write", ...)`, and `conan.rs` calls `require_auth_basic` unconditionally. But anonymous callers now reach the *body* of the LFS batch handler, which buffers up to its router-local `DefaultBodyLimit::max(2 GB)` into `Bytes` before the handler's first line, then does one `sqlx` round-trip per object with no cap on `objects.len()`, on a router that is not behind `rate_limit_middleware`. That is an anonymous memory-and-DB amplifier that the merge-base 401 prevented.

I added `is_anonymous_readable_format_post` as a strict subset (gallery query only) and left `is_non_mutating_format_post` in its original permission-check-only role. Anonymous public gallery search still works — that positive control is the author's own test and it still passes.

If the LFS/Conan loosening was intended, it deserves its own PR with its own tests; I did not want to decide that inside a VS Code feature.

**2. `vscode.rs` — the gallery download path skipped the curation gate (HIGH).**

Every other curated proxy download path calls it from the handler: `npm.rs:3044`, `pypi.rs:335`, `cargo.rs:906`, `nuget.rs:1274`/`1993`, `oci_v2.rs`. The shared streaming helper the gallery uses (`proxy_fetch_streaming_response_with_cache_key`) takes no `db` and therefore structurally *cannot* apply it — same shape as #3235. `RepoInfo` already carries `curation_enabled` and `curation_default_action`, so the data was in hand and unused.

The consequence: an operator who writes an emergency-deny rule for a publisher on a `vscode` Remote gets no enforcement at all, silently. Issue #3252 explicitly promises that "allowlists, emergency denies, cooldown, reviews, cache hits, and rewritten asset downloads all use the same server-side decision" — this was the missing half. Added `enforce_curation` to `proxy_gallery_asset`, before the age gate so a denied publisher costs no upstream round-trip, keyed on the same casefolded `publisher.extension` identity the age gate uses so one operator rule governs both.

**3. `vscode.rs` — the legacy download route became an age-gate bypass (MEDIUM, newly created by this PR).**

`download_vsix`'s Remote pull-through branch is unchanged by this PR and has no gate. That was harmless before, because `vscode` was not in the age-gate matrix at all — `supports_format_mode` returned false and the admin API rejected enabling it. This PR adds `VSCODE_AGE_GATE_SPEC` and relaxes the validation message, so an admin can now turn the gate on for a `vscode` repo, and `/vscode/{repo}/extensions/{pub}/{name}/{ver}/download` serves right past it on the same repository.

The legacy route synthesizes its own upstream path, has no gallery document to resolve publish time from, and has no platform in its coordinate — so it genuinely cannot evaluate the gate. Rather than fake an evaluation, it now refuses with the shared `age_gate_unavailable` fail-closed response while the gate is enabled, pointing at `/gallery/...`. Zero regression risk: no existing deployment can have the gate enabled on a `vscode` repo. Locally-hosted artifacts are untouched — the gate governs proxy pull-through only.

**4. `vscode.rs` — `/vscode/{repo}/item` was a header-controlled open redirect (MEDIUM).**

`gallery_item` built its `Location` from `RequestBaseUrl`, which falls back to `X-Forwarded-Host` / `Host` when `AK_EXTERNAL_URL` is unset. This is the first format-handler route to put that value in a `Location` a browser follows — everywhere else it only lands in JSON bodies, where being self-referential is harmless. Live-confirmed below: `X-Forwarded-Host: evil.example` produced `Location: https://evil.example/repositories/<key>` on an unauthenticated request. Changed to a relative target, which reaches the same page and cannot leave the origin.

**5. `vscode.rs` — `gallery_vspackage` put two different upstream URLs behind one permanently-immutable cache key (MEDIUM).**

The upstream query string was derived from the raw `Option` (omitted when absent) while the cache path always used the normalized value. So `.../vspackage` and `.../vspackage?targetPlatform=universal` were two distinct upstream requests sharing `gallery/P/N/V/universal/vspackage` — and `classify_vscode_gallery_asset` marks that shape `Immutable`, i.e. retained without revalidation. Whichever landed first defined the bytes served for the other, forever. Now sends the same normalized platform the cache key is built from, matching what `openvsx_asset_url` already does on the sibling asset endpoint.

**6. Four test strengthenings.**

- `gallery_age_gate_without_service_fails_closed_for_metadata_and_delivery` asserted only `503` on the delivery half — but the fixture state has no `ProxyService` either, and "Proxy service is unavailable" is *also* a 503. Deleting the entire delivery age-gate seam left it green. It now asserts the structured `error: "age_gate_unavailable"` body, which only the fail-closed seam produces. (Revert-proof below.)
- `gallery_routes_reject_non_remote_repositories` asserted only `501` — but `unsupported_gallery_repo` emits 501 for both the non-Remote and the non-public check, and `create_repo` leaves `is_public` at its `false` default, so either guard satisfied it. Now pins the discriminating clause.
- `gallery_latest_returns_one_extension_not_query_envelope` checked `assetUri` but never `fallbackAssetUri`, which the mock also carries — a regression dropping that field from the rewrite loop would have leaked a CDN URL to the client on the most-used route while staying green. Added the same whole-body host scan the extensionquery test uses.
- `gallery_extensionquery_rejects_malformed_client_json_without_proxying` had no `expect(0)` mock, so the `without_proxying` half of its name was a claim, not an assertion. Added one.

**7. `scripts/dogfood-vscode-openvsx.sh` — inverted assertion.** `.assetUri?` on an object *lacking* the key yields `null` (the `?` suppresses type errors, not missing keys), and `null | startswith(_)` is a jq runtime error → non-zero exit → a "gallery response retains a non-AK asset URL" failure that is the opposite of the truth. Only masked today because `flags: 511` includes `IncludeAssetUri`. Now drops nulls first.

**8. `cargo clippy -- -D warnings` was failing, and Check Rust is red on this PR for exactly that reason.**

```
error: the `Err`-variant returned from this closure is very large
   --> backend/src/api/handlers/vscode.rs:599:22   (filter_gallery_response_age_gate)
    = note: `-D clippy::result-large-err` implied by `-D warnings`
```

Newer clippy applies `result_large_err` to the inner `.map(|version| ...)` closure that collects into `Result<Vec<_>, Response>`. I reproduced it locally on stable 1.97.0 and confirmed against the Check Rust job log that CI hit the same lint on the same toolchain. Added `#[allow(clippy::result_large_err)]` on the function, consistent with the six other helpers in this file that already carry it. Worth flagging that the PR description says local clippy passed cleanly — likely an older stable locally.

### Revert-proof

Six guards, each reverted **faithfully** — the reverted tree differs from the PR head by insertions only (my new tests), with production code byte-identical to `04234d2`:

```
$ git diff pr-3253 --shortstat        # on the reverted auth.rs tree
 1 file changed, 96 insertions(+)
$ git diff pr-3253 -- backend/src/api/middleware/auth.rs | grep '^-' | grep -v '^---' | wc -l
0
```

All runs with `DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` exported against a dedicated Postgres 16 — the DB-backed tests run in 0.23–0.47s, not the ~0.01s a skipped `try_pool()` early-return would show.

**Guard 1 — `auth.rs` anonymous-write narrowing.** Reverted to the PR-head `&& !non_mutating_post`:

```
PASS [0.014s] test_repo_visibility_public_vscode_gallery_query_no_auth_passes      <- positive control holds
PASS [0.020s] test_repo_visibility_public_vscode_publish_no_auth_returns_401
PASS [0.025s] test_repo_visibility_private_vscode_gallery_query_no_auth_returns_401
FAIL [0.025s] test_repo_visibility_public_lfs_batch_no_auth_still_returns_401
FAIL [0.025s] test_repo_visibility_public_conan_authenticate_no_auth_still_returns_401
Summary: 5 tests run: 3 passed, 2 failed
```

The three vscode controls staying green is the point — the fix did not simply re-close everything.

**Guard 2 — curation gate removed from `proxy_gallery_asset`:**

```
FAIL gallery_delivery_honors_curation_block_before_upstream_fetch
  assertion `left == right` failed: a curation block rule must stop the gallery download
    left: 404      <- reached upstream, which has no mock for the blocked coordinate
   right: 403
```

**Guard 3 — `reject_ungateable_legacy_pull_through` removed:**

```
FAIL legacy_remote_download_refuses_while_age_gate_is_enabled
  assertion `left == right` failed: the ungateable legacy route must fail closed while the gate is on
    left: 200      <- the VSIX is served, gate enabled. The bypass, demonstrated live.
   right: 503
```

**Guard 4 — `/item` redirect restored to the base-URL-derived absolute form:**

```
FAIL gallery_item_redirect_ignores_a_forged_forwarded_host
  Location must not echo a client-supplied host:
  https://evil.example/repositories/ph-test-vscode-d5254012-...
```

**Guard 5 — the author's delivery age-gate seam removed, to prove the strengthened assertion now catches it:**

```
FAIL gallery_age_gate_without_service_fails_closed_for_metadata_and_delivery
  panicked at vscode.rs:2010   <- the new body assertion
```

Note where it failed: line 2010 is the added `error == "age_gate_unavailable"` check. The pre-existing `assert_eq!(delivery_status, SERVICE_UNAVAILABLE)` a few lines above **passed** on the same run, which is precisely why the original assertion was hollow.

**Guard 6 — `gallery_vspackage` upstream query restored to the raw-`Option` shape:**

```
FAIL gallery_delivery_honors_curation_block_before_upstream_fetch
  left: 404 / right: 200
```

The mock requires `query_param("targetPlatform", "universal")` for a request that sent no `targetPlatform`, so the cache-key/upstream-URL disagreement is now pinned.

Each guard was restored to the fixed state and re-verified before moving to the next; the whole set green together:

```
Summary [1.494s] 91 tests run: 91 passed, 15008 skipped
```

Gates: `cargo fmt --all` clean, `SQLX_OFFLINE=true cargo build --lib --tests` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean. No new `sqlx::query!` macro text, so no `cargo sqlx prepare` needed.

### Deliberately left as follow-ups

Filed rather than folded in, because each is a design decision or a tuning question rather than a defect with an obvious right answer:

- **#3254** — scan-on-proxy is inert for `vscode` repos (the toggle is accepted and does nothing); and the legacy download route has no boundary charset gate, relying on `validate_cache_path` downstream.
- **#3255** — every gated asset serve costs a fresh uncached upstream POST, including on warm cache hits (~5 per extension page); and gallery queries reserve 32x their wire cap from the shared 1 GiB metadata budget with an unbounded wait, which can park other formats' metadata fetches. The 32x multiplier is assumed rather than measured — worth setting from the ~400 KiB live response you observed.
- **#3256** — `target_platform` is not canonicalized (case/whitespace), so one artifact can occupy unbounded permanently-immutable cache entries, and the policy coordinate is casefolded while the cache key is not; plus a missing `versions` array 502s a legitimate flag combination, `find_gallery_exact_version` aborts on one malformed sibling, and `TotalCount` is reconciled against a cross-page denominator.
- **#3257** — the private-repository gallery story: `501` vs `403`, and whether to support it at all. The current test wording ("must remain unsupported") turns a scoping decision into a requirement.

I also corrected the `rewrite_gallery_asset_urls` doc comment, which claimed "we do not retain an upstream absolute URL in the response". The install/update path is fully rewritten, but `versions[].properties[]` links (`...Services.Links.Source`, `...Code.SponsorLink`) and `publisher.domain` pass through unchanged — which is a reasonable scope, just not what the sentence said.

### Status

Ready to merge once CI is green, which it now can be — Check Rust was red before this push and the cause is fixed. The gateway design is sound; the gaps were the shared-gate omission that this codebase keeps re-finding (#3235, #3068), one behaviour change that reached well outside the feature, and four assertions that could not fail.

cazlo:
yo thank you for your help with review and fix @brandonrc . just a heads up I'm probably not gonna have time to do the testing I was talking about until next weekend :(

brandonrc:
## Follow-up cluster #3254–#3257: analysis posted, deliberately no parallel PR

I was assigned to ship a consolidated fix for the four follow-ups filed off this review (#3254, #3255, #3256, #3257). First step was to check where the code lives, and the answer changes the plan:

**`origin/main` (`b0643226`) does not contain the gallery gateway.** `backend/src/api/handlers/vscode.rs` there is 873 lines with none of the surface these issues describe — no `proxy_gallery_asset`, `gallery_upstream`, `normal_target_platform`, `find_gallery_exact_version`, or `GALLERY_METADATA_BUDGET_RESERVATION_BYTES` — and `docs/vscode-openvsx-gateway.md` does not exist. All of it is only here, in this still-draft PR.

So I have **not opened a PR.** Fixing these against `main` would mean either forking your 2,474-line `vscode.rs` diff or duplicating helpers into a file this PR rewrites — a guaranteed conflict and a duplicate of your work either way. Instead I have posted a full analysis on each of the four issues so the fixes are ready to apply here (or as a fast follow-up once this merges). Summary of what is in them:

**#3254 — scan-on-proxy inert; legacy route boundary gate.** Confirmed both. The scan gate is not a one-line insert: `proxy_gallery_asset` ends in a *streaming* serve and `gate_proxy_scan_serve` needs buffered bytes + digest, so it needs npm's branch-instead-of-stream shape (`npm.rs:3305-3325`, gate call at `npm.rs:3961`). Two scoping notes: gate the **VSIX asset type only** (that path also serves icons/README/changelog and buffering those is pure cost), and `ProxyScanIdentity` has to be `NotApplicable` for now because `ComponentEcosystem` only has `Npm`/`Python` — worth landing anyway, with "pin the embedded `extension/package.json`" as a separate follow-up. Please reuse `gate_proxy_scan_serve` / `direct_scan_policy` / `ProxySeverityGate` rather than a parallel gate — #3375 and #3376 just consolidated that logic into one seam. On the legacy route: its `format!` interpolation is on `main` today, but the fix reuses `validate_gallery_request_segment` which only exists here — and this PR already edits exactly the right spot, since `reject_ungateable_legacy_pull_through(&repo)?` goes at the top of the same `Err(not_found)` arm.

**#3255 — re-query amplification and the 32× reservation.** Confirmed. Unconditional cheap win: `gallery_upstream`'s `SELECT is_public` runs twice per request (once from the handler, once inside `fetch_gallery_query`). For the memo, the framing I would suggest is **memoize the evidence, not the verdict** — cache `find_gallery_exact_version`'s `lastUpdated` for an immutable coordinate, keep `download_basis`/`enforce_age_gate` per request. That collapses the five-POSTs-per-page fan-out while preserving exactly the property `gallery_delivery_age_gate_blocks_before_cache_then_honors_manual_review` defends. On the budget: `try_reserve` already exists on `main` (`proxy_helpers.rs:118`) and `MetadataWorkingSetLimits` is new in this PR, so bounding the wait is nearly free and worth doing before the multiplier is retuned.

**#3256 — targetPlatform and strict parsing.** Confirmed all three. One disagreement worth flagging before anyone writes it: **do not 400 on unknown `targetPlatform` values.** That set is not closed (`alpine-arm64`, `web` were both later additions) and a strict allowlist makes AK the thing that breaks next time one is added — while an unknown platform already 404s from upstream, which is the right answer for free. Canonicalize (trim + lowercase, `Cow<'_, str>`) and constrain the *character class* instead; that bounds the immutable-cache-key space, which is the actual concern. Also, on `reconcile_gallery_result_count` I would delete it rather than fix it — page-local minus cross-page has no correct value, and `saturating_sub` turns a filtered page into `TotalCount: 0`, which a client reads as "does not exist". The issue's own "leave `TotalCount` alone" option is the right one. The `versions`-retain fix has a concrete keep-mask shape in the issue comment.

**#3257 — private-repository galleries.** I made the call rather than listing options: **do not implement private galleries** (the client cannot present a credential, and the reverse-proxy shape is both unverifiable and already achievable by fronting a public repo), and **split the status**: keep 501 for the non-Remote repo-type arm, return **403** for the Remote-but-private arm. The deciding argument is not aesthetics — `repo_visibility_middleware` 401s the anonymous case first (your own `test_repo_visibility_private_vscode_gallery_query_no_auth_returns_401` pins it), so this branch is only ever reached by an authenticated, authorized member, which is the definition of 403; and **RFC 9110 makes 501 heuristically cacheable by default while 403 is not**, so a cached 501 survives an operator flipping the repository to public. The split also retires the body-substring workaround your remediation commit had to add ("501 is emitted by both the repo-type and the public-only guard"). Full reasoning and the concrete edits are on #3257.

Happy to send any of these as PRs into `feat/vscode-openvsx-gateway`, or to pick them up as a follow-up once this lands — whichever is less disruptive to you. Just say which.

brandonrc:
@cazlo — status check from the 1.8.0 milestone sweep: this branch is currently **green on all 14 checks** (unit, integration, coverage, security audit, smoke E2E) and reports `MERGEABLE` / `CLEAN` against today's `main`, which has moved a lot in the last day.

Is it ready to come out of draft, or are you still working on it?

Asking because there are now **four follow-up issues queued directly behind this PR** — #3254 (scan-on-download inert for vscode repos), #3255 (per-asset upstream re-query amplification + the 32x metadata budget reservation), #3256 (targetPlatform canonicalization + strict upstream parsing), and #3257 (the private-repository gallery story). All four describe code that exists only on this branch, so none of them can be fixed against `main` until this lands. Rather than fork your `vscode.rs` work, we've posted the analysis for each on the issues themselves so it's ready to pick up the moment this merges.

No pressure on the timing — just want to avoid two people editing the same 2,400-line diff. If you'd rather land this as-is and take the four follow-ups separately, say the word and we'll pick them up after the merge.

cazlo:
hey @brandonrc I didn't want to sign off on this until I get some manual tests done to make sure this is working well for the use cases I was building out. from initial automated tests it looks like it should. I'll be getting these sorted out shortly, like probably before end of day.

brandonrc:
Thanks @cazlo , let me know once you get the testing done. I am happy to put this back in the 1.8.0 milestone if there is time to get it in there. Moving it to 1.9.0 for now just to be safe. 

Keep me posted!

cazlo:
Update from real-client QA: I've been running this branch with VSCodium 1.126 pointed at AK and a 7-day age gate on an Open VSX remote.

Real use found a real bug: searches touching version-heavy extensions 502'd, because VSCodium's IncludeVersions follow-up makes Open VSX return complete version histories (28 MiB / 12,592 versions for rust-analyzer alone, 55 MiB for one measured @popular page) and no bounded metadata cap survives that. The latest commits replace buffered passthrough with bounded per-extension composition — a typed version skeleton plus a latest-per-platform detail query, newest 30 per (platform, channel), asset URLs synthesized from the gateway's own templates. The failing searches now return 200 in ~10 s at well under 1 MiB, and /latest on a gate-withheld extension resolves the newest allowed version instead of 404ing.

While folding in the QA fixes I also picked up the bounded parts of the follow-up cluster: curation enforcement plus boundary validation on the legacy download route (#3254), targetPlatform canonicalized to a single coordinate for cache, upstream, and policy (#3256 item 1), and the public-only contract made explicit — private Remote gallery routes 403, non-Remote 501 (#3257). The rest of #3254/#3256 and all of #3255 stay as follow-ups.

Also fixed from live testing: gallery assetUri was built from X-Forwarded-Host (reproduced through my real edge), and explicit "assetUri": null from Open VSX — emitted whenever a client omits IncludeAssetUri — was rejected as malformed instead of passed through.

brandonrc:
Merged — thanks @cazlo, and thanks especially for doing real-client QA rather than stopping at green CI.

The 502 you found is exactly the kind of thing automated tests were never going to catch: Open VSX returning complete version histories (28 MiB / 12,592 versions for rust-analyzer alone, 55 MiB on a measured `@popular` page) means no fixed metadata cap survives contact with a version-heavy extension. Replacing buffered passthrough with bounded per-extension composition is the right shape, and the `assetUri`-from-`X-Forwarded-Host` fix is a genuine header-injection bug that only shows up behind a real edge.

This is now in the **1.8.0** milestone (#3252 closed with it).

Also worth noting: your 501-vs-403 split matches, independently, the conclusion an internal analysis had reached on #3257 — including the deciding argument that **501 is heuristically cacheable by default while 403 is not**, so a cached 501 would survive an operator flipping the repository public. #3257 is closed as decided-and-implemented.

Remaining follow-ups stay open with the split recorded on each:
- **#3254** — curation enforcement and boundary validation landed; the scan-on-download half remains (the obstacle is that the asset route streams while the scan gate needs buffered bytes plus a digest — npm's branch-instead-of-stream pattern is the model).
- **#3256** — `targetPlatform` canonicalization landed; relaxed parsing of upstream gallery documents remains.
- **#3255** — per-asset upstream re-query amplification and the 32x metadata-budget reservation, untouched.

Issue #3258, opened 2026-08-10
OCI: HEAD /v2/<name>/blobs/<digest> has no scan re-block (pre-existing #3003 gap)


brandonrc:
Found while reviewing #3245.

`blob_belongs_to_vulnerable_image` / `blob_belongs_to_vulnerable_image_any` have exactly one production call site: `handle_get_blob` (`backend/src/api/handlers/oci_v2.rs`). `handle_head_blob` (`oci_v2.rs`, the `HEAD /v2/<name>/blobs/<digest>` route) performs no equivalent check.

## Impact

For a blob belonging to an image with a stored `vulnerable` verdict:

- `GET  /v2/<name>/blobs/<digest>` -> `403 DENIED` ("blob belongs to an image blocked by scan policy")
- `HEAD /v2/<name>/blobs/<digest>` -> `200 OK` with a `Content-Length` and `Docker-Content-Digest`

No blob bytes leak, so this is not the image-reconstruction hole #3003 round 2 closed. What it does leak is **existence and exact size** of blocked content, and it makes the two verbs disagree on the same resource. The distribution-spec (`opencontainers/distribution-spec`, "Pulling blobs") describes `HEAD` as the existence check for exactly the resource `GET` serves, so a client that HEADs first sees the blob as available and then takes a 403 — the same "un-pullable but not "blocked"" confusion #3105 reported at the manifest/blob seam, one layer down.

## Expected

`handle_head_blob` should apply the same re-block as `handle_get_blob` (including the `scan_on_proxy` scoping added in #3245: proxy-served refs honor the flag, hosted refs do not), and answer `403` where `GET` would.

## Scope

Pre-existing since #3003; not introduced by #3245. Filed separately rather than widening that PR.

Related: #3003, #3105, #3245.

Issue #2329, opened 2026-07-09
Public Repository Read Access Inconsistency: Authenticated User Denied by ACL While Anonymous Read Is Allowed


danatri:
## Summary

When a repository is marked `public` but has any fine-grained `permissions` rules configured, an authenticated non-admin user without matching rule actions is denied (`403`), while anonymous access may still succeed for read paths.

This creates an inconsistent access model where authenticated users can be blocked even though the repository is public.

---

## Why This Looks Like a Bug

For public repositories, operators generally expect read access to be at least as permissive for authenticated users as for anonymous users.

Current behavior appears to prioritize fine-grained ACL checks for authenticated users whenever any repo rules exist, which can produce:

- `anonymous read` -> allowed
- `authenticated read without matching ACL` -> denied

That is counterintuitive and can break npm clients that always send auth headers/tokens.

### Environment Information

Deployment type: docker compose
Operating System: Ubuntu 24.04
Architecture: x86
Database: postgres 16
Reverse Proxy: nginx
Network setup: container/host

### Backend Version

1.2.1

### Web UI Version

1.2.0

### Package Format

npm

### Expected Behavior

One of the following should be true and explicitly documented/configurable:

1. Public means public for read, regardless of whether caller is anonymous or authenticated.
2. If ACL mode is enabled for a repository, both anonymous and authenticated access should be evaluated consistently against ACL policy (no auth-state-dependent inversion).

At minimum, authenticated users should not have less read access than anonymous users for a `public` repository unless this is an explicit, documented policy mode.

### Observed Behavior

- Repo has permission rules (`permissions` rows) for `target_type='repository'`.
- Authenticated non-admin user without matching actions resolves to `actions=[]`.
- Request denied in permission middleware branch.

Backend logs observed:

- `permission denied: rules exist but no actions granted`
- `permission cache hit ... actions=[]`

Manual grant of `read` to user/group immediately resolves issue.

### Steps to Reproduce

1. Create/set a repository as public (for example `<repo-key>`).
2. Add at least one entry in `permissions` for this repository (group/user ACL).
3. Use user A without matching ACL action.
4. Call with bearer token:

```bash
curl -L -H "Authorization: Bearer $TOKEN" \
   "https://<ak-host>/npm/<repo-key>/<package-path>"
```

5. Optionally compare with anonymous call (if endpoint/repo allows):

```bash
curl -L "https://<ak-host>/npm/<repo-key>/<package-path>"
```

[...]

PR #3268, opened 2026-08-10
fix(oci): authorize /v2 reads instead of gating on anonymity alone

### LINKED ISSUE #3261: OCI /v2 reads authorize on anonymity alone: any authenticated account can pull any private image
Found during independent review of #3250. **Pre-existing — not caused by that PR.** No existing issue covers it.

## What

`/v2` is registered at `backend/src/api/routes.rs:145-146`, **before** `.merge(format_routes)` at `routes.rs:148`. `repo_visibility_middleware` is attached at `routes.rs:109-112` and covers `format_routes` only, so the entire OCI/Docker read surface sits **outside** it.

Inside the handlers, the only authorization on reads is an anonymity check:

- `handle_get_blob` — `backend/src/api/handlers/oci_v2.rs:4596-4598`
- `handle_get_manifest` — `backend/src/api/handlers/oci_v2.rs:7889-7891`

Both gate on `is_anon && !repo.is_public` and nothing else. `authenticate_oci` (`oci_v2.rs:316-370`) establishes *who* the caller is but never asks whether they may read this repository.

The only other check on the path is `enforce_token_repo_scope`, which is **not** an authorization gate for session callers: a JWT session carries `AccessScope::Admin`, which `grants()` satisfies unconditionally. (Same trap as #3236, where `require_scope("delete")` looked like a gate and was not.)

**Net effect: any authenticated account — including a read-scoped service account with no grant on the repository — can pull any private image by name and digest.**

Writes are *not* affected: `oci_v2.rs:550-577` gates them properly.

## Why it was not noticed

The `permissions` table genuinely is consulted for native format routes (pip, npm, maven…) through `repo_visibility_middleware`, and for the REST surface through `require_repo_action`. So "permission enforcement is active" is true for those paths — which is why the reporter of #3185 saw correct behaviour when testing with pip. Running the same experiment against Docker returns the opposite answer.

Also worth noting for whoever picks this up: the dominant REST gate `require_repo_write_access` (29 call sites) does not consult `permissions` at all — `require_repo_action` exists in only 3 files. Enforcement coverage is per-surface, not global.

## Reproduction

1. Create a **private** OCI repository, push an image.
2. Authenticate as a second account with **no** grant and **no** role assignment on that repository.
3. `GET /v2/&lt;repo&gt;/manifests/&lt;tag>` and `GET /v2/&lt;repo&gt;/blobs/&lt;digest>`.

Expected: 401/403. Actual: 200 with the manifest and layer bytes.

## Suggested direction

Move `/v2` inside the middleware, or add an explicit read authorization call in `handle_get_blob` / `handle_get_manifest` mirroring what `repo_visibility_middleware` does for reads (`has_any_rules_for_target` + `check_permission`, falling back to role assignments). Do **not** use `can_access_repo` or a bare `require_scope` — neither is a visibility check.

Because the fix changes who can read existing private repositories, it needs a test asserting both directions in one fixture: a granted principal still pulls 200, an ungranted authenticated principal gets 403, and anonymous access to a public repo is unchanged.

## Related

- #3185 / #3250 — the stale startup warning whose review surfaced this
- #3236 — `scopes: None` satisfies `require_scope` unconditionally
- #3235 — download gate skipped on virtual-repository member serves

brandonrc:
Fixes #3261

## The bug

`/v2` is nested in `routes.rs` **before** `.merge(format_routes)`, and
`repo_visibility_middleware` is layered on `format_routes` only — so the entire
OCI/Docker read surface sits outside the middleware that authorizes every other
format's reads. Inside the handlers the only check on the pull path was

```rust
if is_anon && !repo.is_public { /* 401 challenge */ }
```

i.e. *anonymity*, not authorization. `authenticate_oci` establishes **who** the
caller is and never asks whether they may read this repository. The neighbouring
gates do not close it either: `enforce_token_repo_scope` is `None`-unrestricted
for a JWT session, `require_scope` is satisfied unconditionally by
`scopes: None` (#3236), and `can_access_repo` is a token-scope ceiling
(`AccessScope::Admin => true`), never a visibility check.

Net effect: **any authenticated account could pull any private image** by name
and digest. Writes were never affected (`require_oci_repo_write_access`).

The bug was reproduced as a unit test against the real router before fixing —
see the revert-proof below, where the pre-fix code answers `200` to an ungranted
authenticated principal.

## Design decision: explicit read authorization in the handlers, not the middleware

The issue offered two directions. I took the handler gate, deliberately:

* **Moving `/v2` inside `repo_visibility_middleware` does not work.** The
  middleware reads the *first path segment* as a repository key, which would
  swallow `/v2/token` and `/v2/_catalog`; it resolves credentials through
  `try_resolve_auth_outcome`, which does not understand the OCI bearer minted by
  `/v2/token` or the `anonymous` sentinel token; and it answers with
  `WWW-Authenticate: Basic`, which would break the Docker/OCI challenge
  handshake. It would have been a rewrite of the OCI auth path disguised as a
  routing change — on a p1 security fix.
* The write path already established the pattern on this surface
  (`require_oci_repo_write_access`, #2603 G1), so the read gate is its
  read-side twin and sits in the same place in every handler.

`require_oci_repo_read_access` reproduces the decision
`repo_visibility_middleware` makes for a read, through the **same predicates**
so the two surfaces cannot answer differently:

| Case | Decision | Native-path counterpart |
|---|---|---|
| global admin | allow | the middleware's `!ext.is_admin` guard |
| scanner-scoped pull token | allow (see below) | n/a — OCI-only |
| public repository | allow | `public_read_satisfies_acl` (#2329) — reused, not reimplemented |
| private + fine-grained rules exist | `check_permission(read)` \|\| `check_permission(admin)` | the same two calls (#817/#827) |
| private + no rules | role assignment scoped to the repo, or global | the same `role_assignments` EXISTS predicate |
| lookup error | 503, fails closed | same |

No new predicate was invented: `has_any_rules_for_target` + `check_permission`
for the rules branch, and the identical `role_assignments` EXISTS query for the
rules-less branch.

## Protocol correctness (the 401 challenge is preserved)

The anonymous arm is **untouched**: an anonymous caller against a private repo
still gets `401` +
`WWW-Authenticate: Bearer realm="<base>/v2/token",service=...,scope="repository:<name>:pull"`,
so the challenge → token-exchange → retry handshake works exactly as before.

The new gate never re-issues that challenge, on purpose. The caller has already
authenticated successfully, so a 401 would send docker/podman back to
`/v2/token` for a credential it already holds, and loop. It returns terminal
answers instead, matching the distribution-spec error codes
(`opencontainers/distribution-spec`, Error Codes):

* `403 DENIED` — "requested access to the resource is denied" — when a
  fine-grained rule exists and denies the read. The same code the OCI write gate
  already returns.
* `404 NAME_UNKNOWN` — for a rules-less private repo the caller holds no
  assignment on. The body is **byte-identical** to the one `resolve_repo_inner`
  emits for a key that does not exist (`repository not found: <key>`), so
  403-vs-404 cannot be used to probe which private repositories exist. This
  mirrors `not_found_response` on the native path and REST `require_visible`.

## Coverage: all five OCI read verbs

`GET`/`HEAD` blob, `GET`/`HEAD` manifest, and `GET tags/list`. The HEAD verbs
had the *identical* gap — they are separate handlers carrying the same
`is_anon && !repo.is_public` line — and `tags/list` leaked a private
repository's tag names to any authenticated caller for the same reason (it
authenticated and then discarded the claims; they are now bound and gated).
That is the complete read surface of `catch_all`.

This is **unrelated to #3258**, which is about the scan re-block not being
applied on HEAD — a different gate in the same handlers. #3245's
`blob_reblock_applies` is untouched; the new check sits above it.

## The regression this fix had to avoid

`_ak_scanner` (migration 138) is a **non-admin service account seeded with no
role assignments, no permission rules and no API token by design** — its only
credential is the ephemeral, single-repository token
`AuthService::generate_scan_token` mints per scan. Subjecting it to the grant
check would have broken vulnerability scanning of *every private image*. The
`scan_pull_repo` claim is therefore accepted as the authorization it is: it was
minted by this backend for exactly one repository key, and
`enforce_scan_pull_scope` — which runs immediately before — has already
rejected the token if that key is not this repository. Both directions are
tested.

## Behaviour change (operators should read this)

This changes **who can read existing private repositories**. An authenticated
principal with no role assignment and no permission rule on a private
repository could pull from it and now cannot. That is the vulnerability, but
deployments that relied on it (e.g. a CI service account holding only an API
token and no grant) must be given a grant. Public repositories, anonymous
pulls, admin pulls and every granted principal are unaffected.

## Tests

Four DB-backed `--lib` tests driving the **real `/v2` router** (the same
dispatch a `docker pull` takes), in `oci_read_authz_tests`:

1. `private_repo_reads_denied_for_ungranted_authenticated_principal` — the bug.
   All five read verbs refused for a second account with no grant, and the
   refused bodies asserted to carry **no manifest/layer bytes**. In the *same
   fixture*: the granted member still pulls 200 with the real manifest bytes and
   the layer verbatim, and a global admin still pulls 200 — so a gate that
   denied everyone could not satisfy this test.
2. `public_repo_reads_unchanged_for_anonymous_and_ungranted` — the anonymous
   pull token a logged-out client presents still gets 200 with content, and an
   authenticated ungranted caller is never left *below* that anonymous baseline
   (#2329).
3. `fine_grained_rules_decide_private_reads` — a `read` rule pulls 200 with
   content; a `write`-only rule gets 403 carrying the OCI `DENIED` code.
4. `scanner_scoped_pull_token_still_reads_private_repo` — a scan token for this
   repo pulls 200; the same identity's token pinned to a *different* repo is
   still rejected 403, so the test cannot pass by neutering the scan-token path.

### Revert-proof

The production change was reverted to the merge base (`1115203`) with the test
module kept, which is provably faithful:

```
$ git diff 1115203 --shortstat -- backend/src/api/handlers/oci_v2.rs
 1 file changed, 417 insertions(+)
$ git diff 1115203 -- backend/src/api/handlers/oci_v2.rs | grep -c '^-[^-]'
0
```

Zero deletions, insertions only, all inside the appended
`#[cfg(test)] mod oci_read_authz_tests` — production code byte-identical to the
merge base, so the failure below can only come from the claim under test.

Both runs with `DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` exported (a
DB-less run would silently early-return PASS; the ~0.4 s timings confirm the
tests really ran):

**Reverted (production code = merge base):**

```
FAIL [0.423s] fine_grained_rules_decide_private_reads
  assertion `left == right` failed: a write-only rule must not confer read on a private repo
    left: 200
   right: 403
FAIL [0.425s] private_repo_reads_denied_for_ungranted_authenticated_principal
  assertion `left == right` failed: an ungranted authenticated principal must not pull a private manifest
    left: 200
   right: 404
Summary [0.455s] 4 tests run: 2 passed, 2 failed
```

`left: 200` **is** the vulnerability: the private manifest served to an account
with no grant. The two controls (public/anonymous unchanged, scanner token)
pass on both trees, as controls should.

**With the fix:**

```
PASS [0.355s] scanner_scoped_pull_token_still_reads_private_repo
PASS [0.358s] fine_grained_rules_decide_private_reads
PASS [0.362s] public_repo_reads_unchanged_for_anonymous_and_ungranted
PASS [0.408s] private_repo_reads_denied_for_ungranted_authenticated_principal
Summary [0.430s] 4 tests run: 4 passed
```

No regressions: `534 tests run: 534 passed` for `api::handlers::oci_v2`, and
`15087 tests run: 15087 passed` for the full `cargo nextest run --workspace
--lib` suite against a freshly migrated database. `cargo fmt --all`, `cargo clippy --workspace
--all-targets -- -D warnings` and `SQLX_OFFLINE=true cargo build --lib --tests`
are clean; no migration, and no `sqlx::query!` macro text was added
(`query_scalar` only, so the offline cache is unchanged).

## Deliberately left as follow-ups

* `GET /v2/_catalog` still lists **every** repository key to any authenticated
  caller — repository-name enumeration, not content disclosure, and it needs a
  per-row visibility filter on `catalog_local_entries` rather than a gate.
* `handle_tags_list` still does not apply the #2290/#504 token repo-scope
  ceiling (`enforce_token_repo_scope`) that the manifest/blob handlers do.
* On a Virtual repository the read decision is made against the virtual's own
  id; a private *member*'s own visibility is not consulted (pre-existing, and
  the same shape as the native path).

All three are filed as #3269.

### DIFF
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -576,6 +576,148 @@ async fn require_oci_repo_write_access(
     }
 }
 
+/// The `permissions` action an OCI pull requires. Matches what
+/// `middleware::auth::action_for_method` maps `GET`/`HEAD` to, so the OCI read
+/// gate asks the permission service the same question the native-protocol
+/// middleware asks.
+const OCI_READ_ACTION: &str = "read";
+
+/// 503 for an authorization lookup that could not be completed. Reads fail
+/// CLOSED: an unreachable permission store must not serve a private image.
+fn oci_authorization_unavailable() -> Response {
+    oci_error(
+        StatusCode::SERVICE_UNAVAILABLE,
+        "DENIED",
+        "repository authorization temporarily unavailable",
+    )
+}
+
+/// OCI v2 **read** authorization — parity with the native-protocol read gate
+/// (#3261).
+///
+/// `/v2` is mounted OUTSIDE `repo_visibility_middleware`: `routes.rs` nests it
+/// before `.merge(format_routes)`, and the middleware is layered on
+/// `format_routes` only. So no middleware ever authorizes an OCI pull. Before
+/// this gate the only check on the read path was `is_anon && !repo.is_public`
+/// — *anonymity*, not authorization — and every authenticated principal could
+/// pull every private image by name and digest. The neighbouring scope gates
+/// do not close it: `enforce_token_repo_scope` is `None`-unrestricted for a
+/// JWT session, and `require_scope` is satisfied unconditionally by
+/// `scopes: None` (#3236). `AuthExtension::can_access_repo` is a token-scope
+/// ceiling (`AccessScope::Admin => true`), never a visibility check.
+///
+/// This reproduces the decision `repo_visibility_middleware` makes for a read,
+/// through the same predicates, so the two surfaces cannot answer differently:
+///
+/// * a global admin bypasses (as the middleware's `!ext.is_admin` guard does);
+/// * a scanner-scoped pull token bypasses — see below;
+/// * a **public** repository confers a read baseline that fine-grained rules
+///   must not take away from an authenticated caller (#2329,
+///   [`public_read_satisfies_acl`]), otherwise an authenticated user would end
+///   up with *less* access than an anonymous one on the same repository;
+/// * a **private** repository *with* fine-grained rules requires the `read`
+///   action (or `admin`, which implies it, #827) → otherwise 403 `DENIED`;
+/// * a **private** repository with *no* rules requires a role assignment
+///   scoped to it (or a global one) → otherwise the existence-hiding 404
+///   `NAME_UNKNOWN`, byte-identical to the response a nonexistent repository
+///   key produces in `resolve_repo_inner`. This mirrors the middleware's
+///   rules-less branch (`not_found_response`) and REST `require_visible`.
+///
+/// **Protocol**: the 401 + `WWW-Authenticate: Bearer` challenge is deliberately
+/// NOT reused here. The caller has already authenticated, so re-challenging
+/// would send Docker/Podman back to `/v2/token` for a credential it already
+/// holds and loop. The OCI distribution spec defines `DENIED` as "requested
+/// access to the resource is denied" (`spec.md`, Error Codes), which is the
+/// correct terminal answer for an authenticated-but-unauthorized pull; the
+/// anonymous arm above still returns the 401 challenge unchanged, so the
+/// challenge → token-exchange → retry handshake is untouched.
+///
+/// **Scanner exemption**: `_ak_scanner` (migration 138) is a NON-admin service
+/// account seeded with no role assignments, no permission rules and no API
+/// token *by design* — its only credential is the ephemeral, single-repository
+/// token minted per scan (`AuthService::generate_scan_token`). Running the
+/// grant check against it would break scanning of every private repository.
+/// The `scan_pull_repo` claim is therefore accepted as the authorization it
+/// is: it was minted by this backend for exactly one repository key, and
+/// [`enforce_scan_pull_scope`] — which MUST run before this helper — has
+/// already rejected the token if that key is not this repository.
+async fn require_oci_repo_read_access(
+    state: &SharedState,
+    claims: &crate::services::auth_service::Claims,
+    repo: &OciRepoInfo,
+) -> Result<(), Response> {
+    if claims.is_admin || claims.scan_pull_repo.is_some() {
+        return Ok(());
+    }
+
+    // #2329: never leave an authenticated caller below the anonymous read
+    // baseline a public repository already grants.
+    if crate::api::middleware::auth::public_read_satisfies_acl(repo.is_public, OCI_READ_ACTION) {
+        return Ok(());
+    }
+
+    let has_rules = match state
+        .permission_service
+        .has_any_rules_for_target("repository", repo.id)
+        .await
+    {
+        Ok(v) => v,
+        Err(e) => {
+            tracing::error!("OCI repository read authorization lookup failed: {}", e);
+            return Err(oci_authorization_unavailable());
+        }
+    };
+
+    if has_rules {
+        let allowed = state
+            .permission_service
+            .check_permission(claims.sub, "repository", repo.id, OCI_READ_ACTION, false)
+            .await
+            .unwrap_or(false)
+            || state
+                .permission_service
+                .check_permission(claims.sub, "repository", repo.id, "admin", false)
+                .await
+                .unwrap_or(false);
+        return if allowed {
+            Ok(())
+        } else {
+            Err(oci_denied_repo_access())
+        };
+    }
+
+    // Rules-less private repository: a role assignment is required. Same
+    // predicate as the middleware's rules-less branch and
+    // `RepositoryService::user_can_access_repo`. `query_scalar` (not the
+    // macro) so no new sqlx offline-query cache entry is needed.
+    match sqlx::query_scalar::<_, bool>(
+        "SELECT EXISTS ( \
+             SELECT 1 FROM role_assignments ra \
+             WHERE ra.user_id = $1 \
+               AND (ra.repository_id = $2 OR ra.repository_id IS NULL) \
+         )",
+    )
+    .bind(claims.sub)
+    .bind(repo.id)
+    .fetch_one(&state.db)
+    .await
+    {
+        Ok(true) => Ok(()),
+        // Existence-hiding: identical to the `resolve_repo_inner` 404 for an
+        // unknown key, so 403-vs-404 cannot be used to probe which private
+        // repositories exist.
+        Ok(false) => Err(oci_error(
+            StatusCode::NOT_FOUND,
+            "NAME_UNKNOWN",
+            &format!("repository not found: {}", repo.key),
+        )),
+        Err(e) => {
+            tracing::error!("OCI repository read access check failed: {}", e);
+            Err(oci_authorization_unavailable())
+        }
+    }
+}
+
 /// Build a Docker/OCI scope string for a repository resource.
 fn pull_scope(image_name: &str) -> String {
     format!("repository:{}:pull", image_name)
@@ -4432,6 +4574,13 @@ async fn handle_head_blob(
         if let Err(resp) = enforce_token_repo_scope(claims, repo.id) {
             return resp;
         }
+        // #3261: authentication is not authorization. `/v2` is mounted outside
+        // `repo_visibility_middleware`, so this is the ONLY read gate on the
+        // OCI surface; without it any authenticated principal could pull any
+        // private image.
+        if let Err(resp) = require_oci_repo_read_access(state, claims, &repo).await {
+            return resp;
+        }
     }
 
     // Check oci_blobs table. Look up by the canonical digest so an upper-case
@@ -4608,6 +4757,13 @@ async fn handle_get_blob(
         // read-side parity of the REST #504 scope gate. No-op for unscoped
         // tokens (allowed_repo_ids == None).
         if let Err(resp) = enforce_token_repo_scope(claims, repo.id) {
+            return resp;
+        }
+        // #3261: authentication is not authorization. `/v2` is mounted outside
+        // `repo_visibility_middleware`, so this is the ONLY read gate on the
+        // OCI surface; without it any authenticated principal could pull any
+        // private image.
+        if let Err(resp) = require_oci_repo_read_access(state, claims, &repo).await {
             return resp;
         }
     }
@@ -7137,6 +7293,13 @@ async fn handle_head_manifest(
         if let Err(resp) = enforce_token_repo_scope(claims, repo.id) {
             return resp;
         }
+        // #3261: authentication is not authorization. `/v2` is mounted outside
+        // `repo_visibility_middleware`, so this is the ONLY read gate on the
+        // OCI surface; without it any authenticated principal could pull any
+        // private image.
+        if let Err(resp) = require_oci_repo_read_access(state, claims, &repo).await {
+            return resp;
+        }
     }
 
     // Curation enforcement (#2930): mirror the GET-manifest gate so a HEAD probe
@@ -7982,6 +8145,13 @@ async fn handle_get_manifest(
         if let Err(resp) = enforce_token_repo_scope(claims, repo.id) {
             return resp;
         }
+        // #3261: authentication is not authorization. `/v2` is mounted outside
+        // `repo_visibility_middleware`, so this is the ONLY read gate on the
+        // OCI surface; without it any authenticated principal could pull any
+        // private image.
+        if let Err(resp) = require_oci_repo_read_access(state, claims, &repo).await {
+            return resp;
+        }
     }
 
     // Curation enforcement (#2930): a `block` rule on this remote/virtual repo
@@ -8607,14 +8777,18 @@ async fn handle_tags_list(
     let scope = pull_scope(image_name);
     // #1776: mirror handle_head_manifest — anonymous tokens are allowed past the
     // auth gate so a public repository's tags can be listed without credentials.
+    // #3261: bind the claims instead of discarding them — the tag list of a
+    // private repository is readable content and needs the same read gate the
+    // manifest/blob handlers apply.
     let is_anon = is_anonymous_token(headers);
-    if !is_anon
-        && authenticate_oci(&state.db, &state.config, headers)
-            .await
-            .is_err()
-    {
-        return unauthorized_challenge_with_scope(base_url, Some(&scope));
-    }
+    let claims = if is_anon {
+        None
+    } else {
+        match authenticate_oci(&state.db, &state.config, headers).await {
+            Ok(c) => Some(c),
+            Err(()) => return unauthorized_challenge_with_scope(base_url, Some(&scope)),
+        }
+    };
 
     let repo = match resolve_repo(&state.db, image_name).await {
         Ok(r) => r,
@@ -8624,6 +8798,15 @@ async fn handle_tags_list(
     // Anonymous tokens may only list tags on public repositories.
     if is_anon && !repo.is_public {
         return unauthorized_challenge_with_scope(base_url, Some(&scope));
+    }
+
+    // #3261: authentication is not authorization — an authenticated principal
+    // with no grant on a private repository must not be able to enumerate its
+    // tags either.
+    if let Some(claims) = &claims {
+        if let Err(resp) = require_oci_repo_read_access(state, claims, &repo).await {
+            return resp;
+        }
     }
 
     let (n, last) = match parse_pagination_params(query) {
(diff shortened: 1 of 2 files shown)


brandonrc:
## Independent validation — REQUEST CHANGES

Reviewed at head `7b8bba3` from a fresh clone with its own Postgres. Every result below is
from a DB-backed `--lib` run against the real `/v2` router (`DATABASE_URL` **and**
`AK_TESTS_REQUIRE_DB=1` exported; runtimes 0.3–1.3 s, so nothing silently skipped).

**The fix is real and the guard is not hollow.** Mutating `require_oci_repo_read_access` to
`if true { return Ok(()); }` fails exactly the two tests the description claims, with the same
signal:

```
private_repo_reads_denied_for_ungranted_authenticated_principal  left: 200  right: 404
fine_grained_rules_decide_private_reads                          left: 200  right: 403
Summary: 4 tests run: 2 passed, 2 failed
```

Unmutated, all four pass in 1.07 s. The positive controls are load-bearing.

But three things came back red. One is a defect this PR introduces, one leaves the #3261 class
**live**, and one will break legitimate pulls on upgrade for principals the description does not
mention — including a repository's own owner.

---

## Claim-by-claim

| # | Claim | Verdict |
|---|---|---|
| 1 | `/v2` could not simply move under `repo_visibility_middleware` | **CONFIRMED** (mechanism restated below) |
| 2 | The read gate reuses the **same predicates** as the write gate | **REFUTED** — see F1 |
| 3 | Anonymous arm untouched; 401 + `Bearer` challenge preserved | **CONFIRMED** |
| 4 | Never re-challenges; terminal 403/404; 404 body non-enumerable | **CONFIRMED in substance**, wording imprecise |
| 5 | Covers all five read verbs | **CONFIRMED for `catch_all`** — but see F2 |
| 6 | Scanner tokens exempt, and a token pinned elsewhere still rejected | **REFUTED on `tags/list`** — see F3 |
| 7 | #3245 `blob_reblock_applies` untouched, new check above it | **CONFIRMED** |

**Claim 1 — confirmed, with a correction worth recording.** `/v2` is nested at
`backend/src/api/routes.rs:145-146`, the middleware is layered on `format_routes` only at
`routes.rs:108-112`. The repo-key problem is *worse* than described, not the same:
`extract_repo_key` (`backend/src/api/middleware/auth.rs:1398-1415`) skips one prefix segment and
returns the **second**, so `/v2/<repo>/<image>/manifests/<ref>` would resolve the repo key as
`<image>`, not `<repo>`. The anonymous-sentinel half is exactly right —
`ANONYMOUS_TOKEN = "anonymous"` (`oci_v2.rs:161`) is not a JWT, not an API token, and not
base64 credentials, so `try_resolve_auth_outcome` (`auth.rs:938-965`) returns
`InvalidCredential` → 401. And `unauthorized_response` (`auth.rs:1589-1601`) emits
`Basic realm="artifact-keeper"` with no `service=`/`scope=`, which no OCI client can act on.
The decision to gate in the handlers is right.

**Claim 4 — the important half holds, the wording does not.** An expired or malformed bearer
still hits `authenticate_oci` → `Err(())` → `unauthorized_challenge_with_scope`, i.e. a **401
challenge**, not a terminal 403 (`oci_v2.rs:4549`, `4734`, `7268`, `8118`, `8785`). That is the
correct behaviour and is preserved.

The 404 bodies are *not* byte-identical — they differ by the echoed key (measured
`content-length: 122` vs `127`). But they are not an oracle either, because both echo the
segment the **client supplied**:

```
private, no grant : {"errors":[{"code":"NAME_UNKNOWN","message":"repository not found: <k>"}]}
nonexistent key   : {"errors":[{"code":"NAME_UNKNOWN","message":"repository not found: <k>"}]}
```

Same code, same status, same headers modulo the attacker's own input. No enumeration. One narrow
divergence to be aware of: `resolve_repo_inner` echoes the **requested** first segment
(`oci_v2.rs:2712`) while the gate echoes the **resolved** `repo.key` (`oci_v2.rs:709-713`). With
`AK_DEFAULT_DOCKER_MIRROR_REPO` set, a missed literal key falls back to the mirror repo, so an
ungranted caller probing a nonexistent key gets `repository not found: <mirror-repo-key>` and
learns the configured mirror's key. Cosmetic, but the description's "byte-identical" should say
"echoes the same client-supplied key" instead.

---

## F1 — HIGH, breaks legitimate pulls on upgrade: an unrelated ACL row revokes read for every role-assignment principal, including the repository owner

The description says the read gate is the "read-side twin" of `require_oci_repo_write_access`
using "the **same predicates**". It is not. It is a copy of the *middleware's read branch*, and
that branch uses a strictly more brittle predicate than the write gate on the same surface:

| | new read gate (`oci_v2.rs:664-716`) | `check_repository_action` (write gate, `oci_v2.rs:561-567`) |
|---|---|---|
| "do rules exist?" | `has_any_rules_for_target` — **principal-blind** (`permission_service.rs:285-296`) | `applicable_rules` bound to the caller + their groups (`permission_service.rs:175-199`) |
| rules branch | `check_permission` → `query_actions`, **`permissions` table only** (`permission_service.rs:478-502`) | rules authoritative **for that principal only** |
| role fallback | none once any rule exists | `assigned_roles` join, and `'admin' = ANY(role.permissions)` **always wins** |

`permission_service.rs:157-161` states the design intent explicitly: *"This principal-scoped
transition prevents the first rule on a target from dropping every unrelated legacy principal to
no access."* The read gate reintroduces exactly that.

This is not theoretical, because `role_assignments` is written by **one** production code path:
repository creation (`repository_service.rs:938-948`, inserting `repository-owner` +
`developer` for `created_by`). Every other `INSERT INTO role_assignments` in `backend/src` is
inside a `#[cfg(test)]` module. And `repository-owner` carries `{read,write,delete,admin}`.

So the owner of a private Docker repo, the moment they grant one teammate `read` through the
fine-grained `permissions` API, loses the ability to pull their own images — while still being
able to push. Measured:

```
VAL1  member GET manifest = 403  (write gate allows push = true)
      OWNER  GET manifest = 403  (write gate allows push = true)
```

Fixture: private docker repo; `member` holds the `developer` `role_assignments` row; `owner`
holds the `repository-owner` + `developer` rows the creation path writes; a single `permissions`
row granting `read` to **a third, unrelated user**. Neither the member nor the owner is named by
that rule.

Yes, this matches `repo_visibility_middleware` — but that is an argument for fixing the
middleware, not for propagating the predicate onto a surface that did not have it. Today these
principals pull fine over `/v2`; after this merges they get 403.

**Fix:** make the private-repo branch call the choke-point the write gate already uses —
`state.permission_service.check_repository_action(claims.sub, repo.id, OCI_READ_ACTION, claims.is_admin)`
— keeping the `public_read_satisfies_acl` short-circuit above it. To preserve the existence-hiding
split, keep `has_any_rules_for_target` for the **status code only** (403 `DENIED` when a rule
exists, 404 `NAME_UNKNOWN` when none does), not for the decision. That makes the claim in the
description true and removes the pull-denied/push-allowed asymmetry.

---

## F2 — HIGH, the #3261 class is still live: a public Virtual repo serves a private member's bytes to an **anonymous** caller

`resolve_virtual_manifest` (`oci_v2.rs:3181`) and `resolve_virtual_blob` (`oci_v2.rs:3077`) take
no `claims` argument and call `proxy_helpers::fetch_virtual_members` (`oci_v2.rs:3092`, `3204`)
with no authorization filter. The new gate authorizes the **virtual parent** only. Call sites:
`oci_v2.rs:4646`, `4923`, `7401`, `8292`.

```
VAL4  via PUBLIC virtual:  anonymous = 200 (private manifest bytes served = true)
                     ungranted authed = 200 (private manifest bytes served = true)
      DIRECT pull of the same private member = 404
```

The direct route is correctly closed by this PR; the aggregated route is not. Because the virtual
is public, `is_anon && !repo.is_public` is false and `public_read_satisfies_acl` allows, so the
new gate is skipped entirely and the private member's manifest is returned to a caller with no
credential at all.

`#3269` item 3 says this is "the same shape the native-protocol path has". That is not accurate.
`proxy_helpers::authorize_virtual_members` exists precisely for this and is called by
`maven.rs:1188,1351,1870`, `npm.rs:3201`, `pypi.rs:1889,2643`, `debian.rs:2821`,
`goproxy.rs:1051`. `oci_v2.rs` is the **only** format handler that never calls it. Its doc at
`proxy_helpers.rs:2995-3007` even names this exact scenario as the thing it fixed: *"through a
public Virtual parent, an ANONYMOUS caller — able to read a private member's bytes."*

**Fix:** thread `Option<&AuthExtension>` (or the `Claims`) into the two virtual resolvers and
filter members through `authorize_virtual_members`, treating a denied member as not containing the
artifact — the same shape the five other handlers use. This should not ride along as a follow-up
note on a p1 read-authorization fix; it is the same vulnerability on a second route.

---

## F3 — MEDIUM, introduced by this PR: `handle_tags_list` grants on `scan_pull_repo` without ever enforcing the pin

`require_oci_repo_read_access` returns `Ok(())` for any token carrying `scan_pull_repo`
(`oci_v2.rs:650-652`), and its own doc says `enforce_scan_pull_scope` *"MUST run before this
helper"* (`oci_v2.rs:641-643`). That precondition holds at four of the five call sites
(`4568`, `4753`, `7287`, `8139`) and **not** at `handle_tags_list` (`oci_v2.rs:8807`), which
calls neither `enforce_scan_pull_scope` nor `enforce_token_repo_scope`.

```
VAL3  scan token pinned to "totally-other-repo":
        GET /v2/<private-repo>/app/tags/list  ->  200 {"name":"app","tags":["v1"]}
        GET /v2/<private-repo>/app/manifests/v1 -> 403   (correct)
```

This refutes claim 6 for the tag-listing verb. Before this PR the omission was inert — `tags/list`
discarded its claims and had no gate at all. This PR makes it load-bearing by turning
`scan_pull_repo` into an authorization **grant**, so a leaked or compromised scan token (it is
handed to an external scanner container, `image_scanner.rs:604-607`, `grype_scanner.rs:769`) now
enumerates any private repository's tags. `#3269` item 2 records the missing
`enforce_token_repo_scope` but not this one, which is the security-relevant half.

**Fix (three lines):** add `enforce_scan_pull_scope(claims, &repo.key)?` — and, per #3269,
`enforce_token_repo_scope(claims, repo.id)?` — to `handle_tags_list`, or move both inside
`require_oci_repo_read_access` so the precondition cannot be missed at a call site again.

---

## Confirmed non-issues (checked, cleared)

* **`_ak_scanner` is genuinely safe.** It is the only `INSERT INTO users` in all 181 migrations
  (`138_scanner_service_account.sql:18-40`), non-admin, `password_hash NULL`, and no migration
  ever grants it a `role_assignments` or `permissions` row. Both scanner backends mint its
  credential *only* via `generate_scan_token` (`image_scanner.rs:566-586`,
  `grype_scanner.rs:661-681`); minting failure degrades to an anonymous pull, never to an
  admin/API-token fallback. The exemption and its justification are correct.
* **Replication / federation is unaffected.** `sync_worker` is push-only over
  `/api/v1/repositories/.../artifacts/...` (`sync_worker.rs:1724`, `1131`, `1204`); its only GET
  is the `GET /api/v1/peers` heartbeat (`sync_worker.rs:353`). The inbound peer surface
  (`routes.rs:697-703`, `handlers/peer.rs`) has no `/v2` route. Nothing in replication reads
  `/v2` in either direction.
* **Migration subsystem, promotion/approval/lifecycle, webhooks:** none of them make a
  self-referential `/v2` request. Every `v2/<image>/manifests/...` string in
  `lifecycle_service.rs` / `storage_gc_service.rs` / `migration_worker.rs` is an `artifacts.path`
  SQL literal, and the migration clients read external Nexus/Artifactory layouts only.
* **Cross-repo blob mount** already authorizes the source independently
  (`oci_v2.rs:5036-5049`), so it is not a way around the new gate.
* **API-token principals resolve to the owning user.** `authenticate_oci` rebuilds claims from
  `generate_tokens(&validation.user)` (`oci_v2.rs:357-368`, `385-396`), so `claims.sub` is the
  user id and the `role_assignments` lookup is against the right principal.

## Operator notes worth adding to the description

* The stated behaviour change ("a CI service account holding only an API token and no grant must
  be given a grant") understates how to *give* that grant. `POST /api/v1/users/:id/roles`
  (`users.rs:771`) writes `user_roles`, which no authorization path reads; SSO/LDAP/OIDC role
  mapping writes the same table (`auth_service.rs:2748-2769`). The only grant the gate honours
  for a non-creator is a fine-grained `permissions` row (`POST /api/v1/permissions`). Worth
  saying explicitly, or upgraders will "assign the role" and still get 404.
* `svc-*` service accounts (`service_account_service.rs:52-89`) and auto-provisioned `ci-*` OIDC
  identities (`ci_auth.rs:134-143`) are created with no grant of any kind. For CI OIDC,
  `mapping.allowed_repo_ids` is a token **ceiling** consumed by `enforce_token_repo_scope`, not a
  grant, so an admin's explicit "this mapping may use repo X" does not satisfy the gate.
* The read gate bypasses on raw `claims.is_admin`, while the middleware uses the scope-gated
  `AuthExtension::is_admin` (`auth.rs:232-235`, GHSA-vvc3). Consistent with the existing OCI write
  gate, so not a regression — noting it so the "same decision as the middleware" framing is not
  taken as exact.

---

### Summary

F3 is small and squarely in scope. F1 changes the shape of the gate and, on the evidence above,
is the difference between a fix that ships cleanly and one that locks repository owners out of
their own private images. F2 means #3261 is not closed — the same bytes are still reachable
anonymously through a Virtual repo — and should not be left as a follow-up bullet on the PR that
claims to close it.

Method: fresh shallow clone at `7b8bba3`, dedicated Postgres, all 181 migrations applied, four
independent tests driving the real `/v2` router. Nothing was pushed to this branch.

brandonrc:
## Remediation pushed — `1755d01`

All three findings fixed on `fix/3261-oci-read-authz`. Fresh shallow clone at
`/tmp/ak-fix-3268`, its own Postgres with all 181 migrations, every run below with
**both** `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1` exported (runtimes 0.3–1.0 s per
test, so nothing silently skipped). One new DB-backed test per defect, each driving
the real `/v2` router, each revert-proved against the **PR-head** shape.

| Test | Defect | Reverted (PR head) | Fixed |
|---|---|---|---|
| `role_assignment_principals_keep_pull_when_an_unrelated_rule_exists` | F1 | FAIL `left: 403 / right: 200` | PASS |
| `public_virtual_does_not_launder_a_private_members_bytes` | F2 | FAIL `200 OK, private bytes served = true` | PASS |
| `scan_token_pinned_elsewhere_cannot_list_private_tags` | F3 | FAIL `left: 200 / right: 403` | PASS |

---

### F1 — read gate now asks what the write gate asks

The rules-exist / `check_permission` / `EXISTS role_assignments` ladder is gone. The
private-repo branch calls the choke-point the write gate already uses:

```rust
state.permission_service
     .check_repository_action(claims.sub, repo.id, OCI_READ_ACTION, claims.is_admin)
```

so a role carrying `admin` (the `repository-owner` row repository creation seeds)
always wins, an applicable rule stays authoritative **for the principals it names**,
and every other principal keeps its role-based capabilities.
`has_any_rules_for_target` survives only to pick the *status code* on a denial —
403 `DENIED` when rules govern the repository, the existence-hiding 404
`NAME_UNKNOWN` when none do — so the split you confirmed under claim 4 is unchanged.
A failed rules lookup there degrades to the 404 rather than a retryable 503: the
denial is already decided, and a flaky query must not turn a firm refusal into
"try again". Claim 2 is now true as written.

Fixture is exactly yours: private docker repo, an owner holding the
`repository-owner` + `developer` rows creation writes, a `developer` member, and a
single `permissions` rule naming a **third, unrelated** user.

```
--- REVERTED (PR-head decision body restored) ---
FAIL role_assignment_principals_keep_pull_when_an_unrelated_rule_exists
assertion `left == right` failed: a repository OWNER must not lose docker pull on
their own private repo because an unrelated read rule was added (#3268 review F1)
  left: 403
 right: 200
```

Positive controls in the same fixture: the rule's own grantee reads (rules path still
works), the `developer` member reads, the owner's blob pull is asserted too — and the
principal with **neither** a rule nor a role assignment is asserted `403`, so a gate
that simply fell open cannot satisfy this test either.

### F2 — `authorize_virtual_members` on the OCI read path

`resolve_virtual_manifest` and `resolve_virtual_blob` now take the caller and filter
members through `proxy_helpers::authorize_virtual_members`, the same helper maven,
npm, pypi, debian and goproxy call. A denied member is dropped from the walk, so it
reads as "this member does not have the artifact" and member existence is not leaked
either. The caller is derived with the standard `From<Claims> for AuthExtension`, so
the `allowed_repo_ids` ceiling and the GHSA-vvc3 scope-gated admin demotion come
along; anonymous (`None`) maps to `RepoVisibility::PublicOnly`.

One thing worth flagging, because it would have been a self-inflicted wound:
`VirtualResolveKey` carries **no principal term**. Filtering members without touching
the negative cache would have made a denied caller's miss get replayed to a granted
one as a false "not found" — i.e. turned the authorization filter into a
cache-poisoning primitive. The cache is therefore consulted *and* populated only when
the caller sees every member, where the outcome is principal-independent; a narrowed
walk records nothing. That moves the member fetch ahead of the cache probe (one
indexed query, against the N-member upstream fan-out the cache exists to bound).

```
--- REVERTED (members unfiltered, PR-head shape) ---
FAIL public_virtual_does_not_launder_a_private_members_bytes
an ANONYMOUS caller must not receive a private member's manifest through a public
Virtual parent (#3261 / #3268 review F2); got status 200 OK with private manifest
bytes served = true
```

Positive controls in the same fixture: the **public** member is still served
anonymously through the same virtual (manifest and layer bytes, compared verbatim),
so the walk is filtered rather than broken; and the private member's own granted
principal still reads it through the virtual, so the filter is not merely denying
everyone. The private member's blob is asserted refused as well as its manifest.

The `blob_reblock_applies` blocklist at `handle_get_blob` deliberately keeps its
**unfiltered** `fetch_virtual_members` — that path only ever *adds* blobs to the deny
set, so narrowing it could only weaken it.

### F3 — the ordering requirement is removed, not documented

Both ceilings now run as the first act of `require_oci_repo_read_access`, ahead of
the admin and scanner bypasses (matching the write gate, where
`enforce_token_repo_scope` applies even to admins):

```rust
enforce_scan_pull_scope(claims, &repo.key)?;
enforce_token_repo_scope(claims, repo.id)?;
```

Both are pure functions of the claims, so the identical calls the four
manifest/blob call sites still make before their own bookkeeping stay correct and
idempotent. No call site can omit them again. This also lands #3269 item 2.

```
--- REVERTED (ceilings back at the 4 call sites only) ---
FAIL scan_token_pinned_elsewhere_cannot_list_private_tags
assertion `left == right` failed: a scan token pinned to another repository must not
enumerate this repository's private tags (#3268 review F3)
  left: 200
 right: 403
```

Positive control in the same fixture: the scanner still lists the tags of the
repository it **is** pinned to (`200`, carrying the real tag list), and the manifest
verb is asserted to stay `403` for the foreign token.

---

### Revert faithfulness

The three fixes were also reverted **together**, producing a tree whose production
code is byte-identical to `7b8bba3` with only the new tests added:

```
$ git diff --no-index --shortstat <7b8bba3 oci_v2.rs> <reverted oci_v2.rs>
 1 file changed, 390 insertions(+), 39 deletions(-)

$ ... | grep '^@@'
@@ -28630,0 +28631,72 @@ mod oci_read_authz_tests {
@@ -28657,39 +28729,2 @@ mod oci_read_authz_tests {
@@ -28999,0 +29035,316 @@ mod oci_read_authz_tests {
```

Every hunk is inside `mod oci_read_authz_tests`. The 39 deletions are test-only: the
manifest+blob seeding in `setup()` was extracted into a shared `seed_image` helper the
new virtual-member fixture reuses. Because the shortstat heuristic stops applying once
a revert also refactors tests, the real property was verified directly — strip every
`#[cfg(test)] mod … { }` block and compare the remainders:

```
PR-head 7b8bba3 production remainder : 403328 bytes
REVERTED tree   production remainder : 403328 bytes   identical=True
FIXED tree      production remainder : 409689 bytes   identical=False
```

So on the reverted tree the failures can only come from the claims under test. That
run:

```
Summary [1.049s] 7 tests run: 4 passed, 3 failed, 15091 skipped
  FAIL  scan_token_pinned_elsewhere_cannot_list_private_tags            (F3)
  FAIL  role_assignment_principals_keep_pull_when_an_unrelated_rule_exists (F1)
  FAIL  public_virtual_does_not_launder_a_private_members_bytes         (F2)
  PASS  private_repo_reads_denied_for_ungranted_authenticated_principal
  PASS  fine_grained_rules_decide_private_reads
  PASS  public_repo_reads_unchanged_for_anonymous_and_ungranted
  PASS  scanner_scoped_pull_token_still_reads_private_repo
```

The four original tests still pass on the reverted tree — each new test fails for its
own defect, not for a shared reason. Each fix was additionally reverted **individually**
and produced the same single failure in isolation.

### Regression

```
cargo nextest run --lib                     15090 tests run: 15090 passed, 0 failed
cargo nextest run --lib -E 'test(/oci_v2/)'   540 tests run:   540 passed
cargo nextest run --test oci_virtual_resolution_tests --run-ignored all
                                                4 tests run:     4 passed
cargo fmt --all                              clean
```

Behaviour you confirmed and that is preserved: anonymous → private still gets the
401 + `Bearer` challenge (the anonymous arm is untouched); an authenticated-but-
unauthorized caller still gets a terminal 403/404 and is never re-challenged; the
#3245 blob re-block path is unchanged. No new `sqlx::query!` macros, so no
`cargo sqlx prepare` churn.

### #3269

Edited. Items 2 and 3 are folded into this PR and are recorded there as shipped
rather than residual — item 3 in particular was the reported vulnerability still open
on a second route, not a follow-up. `GET /v2/_catalog` remains the one genuine
residual, and #3269 is now scoped to it.

PR #3272, opened 2026-08-10
fix(proxy): declare upstream Content-Encoding on verbatim forwards in goproxy, rubygems and hex

### LINKED ISSUE #3260: Content-Encoding dropped on verbatim upstream forwards in goproxy, rubygems and hex
Found during independent review of #3241 (which fixes the Maven instance of this). Not a regression from that PR — these are pre-existing and live.

## The shape

#3211 was: a handler forwards upstream bytes **verbatim** but drops the upstream `Content-Encoding`, so the client stores coded bytes under a name implying they are not coded (RFC 9110 §8.4). #3241 fixes `maven::download_root`. The same shape exists in three other formats.

`resolve_virtual_metadata` (`backend/src/api/handlers/proxy_helpers.rs:2711`) takes a caller-supplied `transform`. Its documentation classifies it as being for callers that "parse or rewrite the body themselves" — but **all four in-tree callers pass an identity pass-through** that forwards upstream bytes unchanged:

| caller | note |
|---|---|
| `goproxy.rs:406` | identity pass-through |
| `rubygems.rs:133` (`gem_info`) | identity pass-through |
| `rubygems.rs:806` (`quick_spec`) | comment literally says "verbatim" |
| `hex.rs:412` | identity pass-through |

Its Pass-1 cached probe even receives the coding and discards it (`proxy_helpers.rs:2670`).

A fifth of the same class, via a different primitive: `goproxy.rs:376` (the Remote arm) calls `proxy_fetch_capped`, despite a `_with_encoding` sibling already existing at `proxy_service.rs:2716`.

## Why it matters

An upstream that responds `Content-Encoding: gzip` (or `deflate`, `br`, `zstd`) to these routes has its coded bytes forwarded with the header stripped. Go module clients, `gem` and `mix` then persist or parse bytes that are not in the form the response claims. For `goproxy` in particular the payloads are `.info`/`.mod`/`.zip` module data, so a silently-compressed artifact on disk is a correctness problem, not just a header nit.

Nothing in the codebase decodes on this path — `http_client::base_client_builder()` sets `.no_gzip().no_brotli().no_deflate().no_zstd()` and an explicit `Accept-Encoding: identity` (`http_client.rs:93-96`) — so the bytes really are forwarded as received.

## Suggested fix

Same treatment as #3241: widen the `resolve_virtual_metadata` path to carry `CachedBody` (the #3194/#3210/#3184 pattern) and have the identity-pass-through callers re-declare the coding, and point `goproxy.rs:376` at the existing `_with_encoding` sibling. Callers that genuinely parse or rewrite the body should keep dropping it, and should say so explicitly rather than by omission.

Also worth correcting in the same change: `resolve_virtual_metadata`'s doc comment, which describes a classification its actual callers do not match.

## Related

- #3211 / #3241 — the Maven instance and its fix
- #3149, #3184, #3193, #3210, #2920 — all closed, none covers goproxy/rubygems/hex

Cites RFC 9110 §8.4 (Content-Encoding describes the bytes as transferred) and §8.6 (Content-Length describes the encoded bytes actually sent).

brandonrc:
Fixes #3260

Same class as the Maven instance (#3211, fixed by #3241): handlers that forward upstream bytes **verbatim** dropped the upstream `Content-Encoding`. Nothing on this path decodes — `http_client::base_client_builder()` sets `.no_gzip().no_brotli().no_deflate().no_zstd()` and advertises `Accept-Encoding: identity` — so the coded bytes really are what the client receives, labelled as if they were plain. Per RFC 9110 §8.4 `Content-Encoding` describes the coding applied to the representation as transferred, and per §8.6 `Content-Length` describes the coded bytes actually sent. For goproxy the payloads are `.info`/`.mod` module data, so a silently-compressed document on the `go` client's disk is a correctness problem, not a header nit.

## Callers changed (now re-declare the upstream coding)

All the issue's sites, plus four more instances of the identical shape found in the same three formats while verifying the table against main:

| site | arm | primitive |
|---|---|---|
| `goproxy::try_proxy_go_metadata` | Remote (`.info`, `@latest` fallback) | `proxy_fetch_capped` → **`proxy_fetch_capped_encoded`** (issue's "fifth instance") |
| `goproxy::try_proxy_go_metadata` | Virtual | `resolve_virtual_metadata` transform |
| `goproxy::get_mod_file` | Remote (`.mod`) | `proxy_fetch_capped` → `proxy_fetch_capped_encoded` *(additional, same shape)* |
| `rubygems::gem_info` | Virtual | `resolve_virtual_metadata` transform |
| `rubygems::quick_spec` | Virtual (comment says "verbatim") | `resolve_virtual_metadata` transform |
| `hex::package_info` | Remote *(additional)* + Virtual | `proxy_fetch_capped_encoded` / transform |
| `hex::list_names` | Remote *(additional)* | `proxy_fetch_capped_encoded` |
| `hex::list_versions` | Remote *(additional)* | `proxy_fetch_capped_encoded` |

The hex Remote arms are the #2658 "signed registry bytes pass through as-is" contract — the same verbatim property that makes the coding load-bearing.

## Callers that keep dropping the coding — now explicitly

- `goproxy::fetch_go_metadata_with_source` — every caller **parses** the body (`@v/list` line filtering, `@latest` JSON) and serves a rebuilt document; doc comment now says the drop is deliberate.
- `goproxy::go_version_exists_upstream` — existence probe; the body is discarded entirely.
- `proxy_helpers::collect_virtual_metadata` — every `extract` closure parses and the caller serves a merged document it built itself; doc comment now says so.

## Seam widening (the #3241 / #3194 / #3210 / #3184 pattern)

- `resolve_virtual_metadata`'s `transform` now receives `(body, content_encoding, member_key)` — widened **in place** so every caller breaks at compile time and must state whether it forwards verbatim (declare the coding) or parses (drop it). Its Pass-1 cached probe no longer discards the coding, and its Pass-2 upstream fetch goes through `proxy_fetch_with_cache_key` (already `CachedBody`-shaped since #3241) instead of the coding-dropping `proxy_fetch`. The doc comment no longer claims a parse-or-rewrite caller population its actual (all-verbatim) callers never matched.
- New `proxy_fetch_capped_encoded`: the plain-path sibling of `proxy_fetch_capped_with_cache_key_encoded`, for Remote arms that forward the capped buffered body verbatim.
- New `forward_verbatim_metadata`: shared verbatim-200 builder (the general form of maven's `forward_root_verbatim`), setting `Content-Type`, `Content-Length` of the coded bytes (§8.6), and the coding when present (§8.4). Used by all nine sites, so the response shape cannot drift per format.

## Tests (DB-backed, `--lib`, one per format)

Each uses a **deflate** (non-gzip) coded upstream **plus an uncoded control in the same fixture** (`tdh::coded_and_plain_upstreams`), asserts the exact bytes the client receives, and **decodes the coded body with the declared coding** (`tdh::inflate_deflate`) — a hardcoded `"gzip"`/`"deflate"` literal or an `if false` guard cannot pass:

- `goproxy: test_go_metadata_forwards_upstream_content_encoding_verbatim_db` — Remote `.info`, Remote `.mod`, Virtual `.info`, plus uncoded controls through all three arms.
- `rubygems: test_rubygems_virtual_forwards_upstream_content_encoding_verbatim_db` — Virtual `gem_info` + `quick_spec`, plus controls.
- `hex: test_hex_registry_forwards_upstream_content_encoding_verbatim_db` — Remote `packages`/`names`/`versions`, Virtual `packages`, plus controls.

Shared fixture/assertion helpers live in `test_db_helpers` (`coded_and_plain_upstreams` -> `CodedUpstreams`, its `assert_coded_forward` / `assert_plain_forward` / `coded_hits` methods, `decode_coded`, `create_remote_and_virtual`) so the three suites do not each carry a copy of the mock/assert blocks. The fixture carries the coding it mounted and every assertion reads it from there, so no assertion can be satisfied by a handler that emits a *different* coding; goproxy and rubygems mount `deflate`, hex mounts `gzip`.

## Revert-proof (faithful, DB-backed)

Reverted tree = merge-base production code + the new tests only, proven by stripping every `#[cfg(test)] mod { }` block from merge-base and reverted files and comparing remainders byte-identical: `goproxy` 55,460 B, `rubygems` 34,200 B, `hex` 70,407 B, `proxy_helpers` 377,068 B each side (`git diff origin/main --shortstat` on the reverted tree: `4 files changed, 357 insertions(+)`, all in test modules / `#[cfg(test)]` `test_db_helpers`).

With `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1` (runtimes 0.28–0.30 s — the tests ran, not skipped):

Re-run after the review fixes below (dropping the coding at the sink — the pre-#3260 shape of all nine builders). Maven's pre-existing `#3211` test now fails alongside them because its duplicate helper was folded into the shared one:

```
FAIL api::handlers::goproxy::tests::test_go_metadata_forwards_upstream_content_encoding_verbatim_db
  assertion `left == right` failed: remote .info: the UPSTREAM's Content-Encoding must be re-declared, not some other coding (#3260)
    left: None    right: Some("deflate")
FAIL api::handlers::hex::tests::test_hex_registry_forwards_upstream_content_encoding_verbatim_db
  assertion `left == right` failed: remote packages/pkg3260: the UPSTREAM's Content-Encoding must be re-declared, not some other coding (#3260)
    left: None    right: Some("gzip")
FAIL api::handlers::rubygems::db_cov_tests::test_rubygems_virtual_forwards_upstream_content_encoding_verbatim_db
  assertion `left == right` failed: virtual gem_info (cold, Pass 2): the UPSTREAM's Content-Encoding must be re-declared, not some other coding (#3260)
    left: None    right: Some("deflate")
FAIL api::handlers::maven::tests::test_download_root_forwards_upstream_content_encoding_verbatim
  assertion `left == right` failed: remote root must re-declare the upstream Content-Encoding (#3211)
    left: None    right: Some("deflate")
```

On the fix branch all three pass, along with the 918 tests of the touched modules (`goproxy`/`rubygems`/`hex`/`maven`/`proxy_helpers`/`proxy_service`) and the full `--workspace --lib` suite.

## No-decode property verified (the inverse failure would be worse)

If any layer decoded, re-declaring the coding would corrupt every client. Verified it does not: (1) `http_client.rs` disables all four reqwest codecs and sends `Accept-Encoding: identity`; (2) `proxy_service.rs` documents "the proxy no longer decompresses upstream bodies" and threads `content_encoding` through `UpstreamResponse` and the cache sidecar untouched; (3) empirically, the tests assert the client-received body is byte-identical to the deflate-coded mock body through the full upstream → proxy → cache → handler stack, and the controls prove an uncoded body gains no header.

## Out of scope (same class, other formats — follow-up)

`pub_registry::proxy_pub_meta_get`'s non-JSON "pass through verbatim" fallback arm and `proxy_helpers::proxy_fetch_or_redirect`'s buffered slow path also drop the coding on bytes they may forward; they are outside this issue's goproxy/rubygems/hex scope and are filed as #3273.

## Review response — the goproxy `@latest` / `@v/list` decision (#3280)

An earlier revision of this PR carried a doc comment on `fetch_go_metadata_with_source` claiming every caller rebuilds its document. That is **false for `latest_version`**, which parses the JSON only to run the age gate and then forwards the upstream bytes verbatim (`goproxy.rs:1219-1223`). Measured on this branch with wiremock upstreams (temporary probe, since removed):

```
upstream @latest plain=50 coded=48 | @v/list plain=21 coded=18

CODED @latest  -> status=404  content-encoding=None  len=27  verbatim=false
CODED @v/list  -> status=200  content-encoding=None  len=30  verbatim=false   <- 18 coded bytes served as 30 bytes of U+FFFD
PLAIN @latest  -> status=200  content-encoding=None  len=50  verbatim=true    <- proves @latest IS a verbatim forwarder
PLAIN @v/list  -> status=200  content-encoding=None  len=21  verbatim=true
```

**Decision: filed as #3280, not folded in.** Both are pre-existing (this PR changed neither path), and neither can be fixed the re-declare-only way this PR is built on: `@latest` 404s because `serde_json::from_slice` rejects the coded bytes before the forward is reached, and `@v/list` corrupts because `String::from_utf8_lossy` mangles them before `filter_go_version_list` passes them through unchanged. Fixing either means **decoding an upstream-controlled body** on a metadata path, which needs its own size bounds and unknown-coding handling — a different change from one that provably alters no bytes. The doc comment now states plainly what the code does and points at #3280.

Also recorded, not fixed: `resolve_virtual_metadata` still drops the member's `Content-Type` (both passes bind it `_ct`), so the four Virtual verbatim forwards hardcode a literal and disagree with the Remote arm of the same endpoint — hex serves a signed protobuf as `application/json`. That is a client-visible behaviour change with its own tests, filed as #3281 so the seam is widened once more rather than twice. The RFC 9110 §12.5.3 residual (AK does not honour the client's `Accept-Encoding`) belongs to #3273.

### DIFF
--- a/backend/src/api/handlers/goproxy.rs
+++ b/backend/src/api/handlers/goproxy.rs
@@ -453,11 +453,27 @@ fn filter_version_list(body: &str, blocked: &std::collections::HashSet<String>) 
 /// virtual repository each member's own gate configuration governs its
 /// contribution (#2264).
 ///
-/// Deliberately DROPS the upstream `Content-Encoding` (#3260): every caller
-/// PARSES the returned body (`@v/list` line filtering, `@latest` JSON) and
-/// serves a document it rebuilt itself, so the upstream coding never
-/// describes the bytes that leave the handler. Verbatim forwards go through
-/// [`try_proxy_go_metadata`] / `get_mod_file`, which re-declare the coding.
+/// DROPS the upstream `Content-Encoding`, and unlike the verbatim forwards
+/// #3260 fixed ([`try_proxy_go_metadata`] / `get_mod_file`, which re-declare
+/// it) that is NOT sound for both callers today:
+///
+/// * `@v/list` ([`list_versions`]) does rebuild its document — but through
+///   `String::from_utf8_lossy`, so a coded body is not rejected, it is
+///   replaced character by character with U+FFFD and served as a 200. The
+///   coding is correctly dropped; the body is corrupt.
+/// * `@latest` ([`latest_version`]) does NOT rebuild anything. It parses the
+///   JSON only to run the age gate and then forwards the upstream bytes
+///   verbatim, so it is a verbatim forwarder that drops the coding. It does
+///   not currently mislabel a coded body only because `serde_json::from_slice`
+///   rejects one first and the request 404s — i.e. the endpoint is broken for
+///   exactly the coded-upstream population #3260 is about.
+///
+/// Both are pre-existing (they predate #3260, which changed neither) and both
+/// are tracked in #3280. Fixing them needs this helper to report the coding
+/// AND a decoder on the parse path, which is a larger change than #3260's
+/// re-declare-only fix: it introduces decompression of an upstream-controlled
+/// body, so it needs its own bounds. Until then this comment describes what
+/// the code does rather than what it ought to do.
 async fn fetch_go_metadata_with_source(
     state: &SharedState,
     repo: &RepoInfo,
@@ -2608,42 +2624,71 @@ mod tests {
     /// as if it were plain.
     ///
     /// Deflate (non-gzip) coded upstream plus an uncoded control in the SAME
-    /// fixture — see `tdh::coded_fixture` for why gzip would prove less.
+    /// fixture — see `tdh::coded_fixture` for why gzip would prove less. The
+    /// hex arm mounts `gzip` for the same reason in reverse: with all three
+    /// suites on one coding, pinning the production line to that literal
+    /// keeps every assertion green.
+    ///
+    /// The Virtual `.info` URI is probed TWICE: cold (Pass 2, the upstream
+    /// fan-out) and warm (Pass 1, `cached_metadata_if_servable`). Pass 1 is a
+    /// line this fix changed — it used to discard the member's coding — and
+    /// only the second probe reaches it.
     #[tokio::test]
     async fn test_go_metadata_forwards_upstream_content_encoding_verbatim_db() {
         let Some(fx) = tdh::Fixture::setup("remote", "go").await else {
             return;
         };
-        let (plain, coded, coded_mock, plain_mock) =
+        let up =
             tdh::coded_and_plain_upstreams("deflate", "application/json", b"go-meta-3260 ").await;
 
         // fx repo = the coded Remote (Remote-arm probes); a second coded
         // Remote wrapped by a Virtual (Virtual-arm probe, cold cache so the
         // upstream pass of `resolve_virtual_metadata` runs); a plain Remote +
         // Virtual pair as the uncoded control.
-        let (state, _cache) = tdh::rewire_remote_proxy(&fx, &coded_mock.uri()).await;
+        let (state, _cache) = tdh::rewire_remote_proxy(&fx, &up.coded_mock.uri()).await;
         let (coded_member_id, _cm_key, virt_coded_id, virt_coded_key) =
-            tdh::create_remote_and_virtual(&fx.pool, "go", &coded_mock.uri()).await;
+            tdh::create_remote_and_virtual(&fx.pool, "go", &up.coded_mock.uri()).await;
         let (plain_id, plain_key, virt_plain_id, virt_plain_key) =
-            tdh::create_remote_and_virtual(&fx.pool, "go", &plain_mock.uri()).await;
+            tdh::create_remote_and_virtual(&fx.pool, "go", &up.plain_mock.uri()).await;
 
         // Remote arm, `.info` (`try_proxy_go_metadata`).
         let uri = format!("/{}/example.com/coded/@v/v1.0.0.info", fx.repo_key);
         let (body, headers) =
             tdh::probe_ok(tdh::router_anon(super::router(), state.clone()), uri).await;
-        tdh::assert_coded_forward(&headers, &body, &coded, &plain, "remote .info");
+        up.assert_coded_forward(&headers, &body, "remote .info");
 
         // Remote arm, `.mod` (`get_mod_file`).
         let uri = format!("/{}/example.com/coded/@v/v1.0.0.mod", fx.repo_key);
         let (body, headers) =
             tdh::probe_ok(tdh::router_anon(super::router(), state.clone()), uri).await;
-        tdh::assert_coded_forward(&headers, &body, &coded, &plain, "remote .mod");
+        up.assert_coded_forward(&headers, &body, "remote .mod");
 
-        // Virtual arm, `.info` (`resolve_virtual_metadata` transform).
+        // Virtual arm, `.info`, COLD: `resolve_virtual_metadata` Pass 2 (the
+        // upstream fan-out) transforms the fetched bytes.
         let uri = format!("/{}/example.com/coded/@v/v1.0.0.info", virt_coded_key);
         let (body, headers) =
             tdh::probe_ok(tdh::router_anon(super::router(), state.clone()), uri).await;
-        tdh::assert_coded_forward(&headers, &body, &coded, &plain, "virtual .info");
+        up.assert_coded_forward(&headers, &body, "virtual .info (cold, Pass 2)");
+
+        // Virtual arm, `.info`, WARM: the same URI again now resolves through
+        // `resolve_virtual_metadata` Pass 1 (`cached_metadata_if_servable`),
+        // which reads the coding off the cache sidecar. Cold and warm are
+        // different lines of production code and only this probe covers the
+        // warm one. The unchanged upstream hit count is the barrier proving
+        // the response really came from the cache rather than a second
+        // fan-out — a barrier both the fixed and the coding-dropping shape of
+        // Pass 1 reach.
+        let upstream_path = "/example.com/coded/@v/v1.0.0.info";
+        let hits_before = up.coded_hits(upstream_path).await;
+        let uri = format!("/{}/example.com/coded/@v/v1.0.0.info", virt_coded_key);
+        let (body, headers) =
+            tdh::probe_ok(tdh::router_anon(super::router(), state.clone()), uri).await;
+        assert_eq!(
+            up.coded_hits(upstream_path).await,
+            hits_before,
+            "second probe must be served from the proxy cache (Pass 1), not refetched"
+        );
+        up.assert_coded_forward(&headers, &body, "virtual .info (warm, Pass 1)");
 
         // Controls: uncoded upstream through the same three arms.
         for (key, what) in [
@@ -2653,12 +2698,18 @@ mod tests {
             let uri = format!("/{}/example.com/coded/@v/v1.0.0.info", key);
             let (body, headers) =
                 tdh::probe_ok(tdh::router_anon(super::router(), state.clone()), uri).await;
-            tdh::assert_plain_forward(&headers, &body, &plain, what);
+            up.assert_plain_forward(&headers, &body, what);
         }
+        // ... including the warm virtual path: a cache hit on an uncoded
+        // member must not invent a coding either.
+        let uri = format!("/{}/example.com/coded/@v/v1.0.0.info", virt_plain_key);
+        let (body, headers) =
+            tdh::probe_ok(tdh::router_anon(super::router(), state.clone()), uri).await;
+        up.assert_plain_forward(&headers, &body, "control virtual .info (warm, Pass 1)");
         let uri = format!("/{}/example.com/coded/@v/v1.0.0.mod", plain_key);
         let (body, headers) =
             tdh::probe_ok(tdh::router_anon(super::router(), state.clone()), uri).await;
-        tdh::assert_plain_forward(&headers, &body, &plain, "control remote .mod");
+        up.assert_plain_forward(&headers, &body, "control remote .mod");
 
         for id in [virt_coded_id, coded_member_id, virt_plain_id, plain_id] {
             let _ = sqlx::query("DELETE FROM repositories WHERE id = $1")
--- a/backend/src/api/handlers/goproxy.rs
+++ b/backend/src/api/handlers/goproxy.rs
@@ -368,34 +368,37 @@ async fn try_proxy_go_metadata(
     upstream_path: &str,
     default_content_type: &str,
 ) -> Result<Response, ()> {
-    // Remote repo: proxy to upstream
+    // Remote repo: proxy to upstream. The upstream body is forwarded
+    // VERBATIM (`.info` JSON / `.mod` bytes as the upstream served them), so
+    // the upstream `Content-Encoding` must be re-declared when present
+    // (RFC 9110 §8.4, #3260) — nothing on this path decodes.
     if repo.repo_type == RepositoryType::Remote {
         if let (Some(ref upstream_url), Some(ref proxy)) =
             (&repo.upstream_url, &state.proxy_service)
         {
-            if let Ok((content, content_type)) = proxy_helpers::proxy_fetch_capped(
-                proxy,
-                repo.id,
-                &repo.key,
-                upstream_url,
-                upstream_path,
-                proxy_helpers::DEFAULT_METADATA_MAX_BYTES,
-            )
-            .await
+            if let Ok((content, content_type, content_encoding)) =
+                proxy_helpers::proxy_fetch_capped_encoded(
+                    proxy,
+                    repo.id,
+                    &repo.key,
+                    upstream_url,
+                    upstream_path,
+                    proxy_helpers::DEFAULT_METADATA_MAX_BYTES,
+                )
+                .await
             {
-                return Ok(Response::builder()
-                    .status(StatusCode::OK)
-                    .header(
-                        "Content-Type",
-                        content_type.unwrap_or_else(|| default_content_type.to_string()),
-                    )
-                    .body(Body::from(content))
-                    .unwrap());
+                return Ok(proxy_helpers::forward_verbatim_metadata(
+                    content,
+                    content_type,
+                    default_content_type,
+                    content_encoding,
+                ));
             }
         }
     }
 
-    // Virtual repo: try each member in priority order
+    // Virtual repo: try each member in priority order. Same verbatim-forward
+    // contract as the Remote arm above: the member's coding is re-declared.
     if repo.repo_type == RepositoryType::Virtual {
         let ct = default_content_type.to_string();
         if let Ok(resp) = proxy_helpers::resolve_virtual_metadata(
@@ -403,14 +406,15 @@ async fn try_proxy_go_metadata(
             state.proxy_service.as_deref(),
             repo.id,
             upstream_path,
-            |bytes, _key| {
+            |bytes, content_encoding, _key| {
                 let ct = ct.clone();
                 async move {
-                    Ok(Response::builder()
-                        .status(StatusCode::OK)
-                        .header(CONTENT_TYPE, ct)
-                        .body(Body::from(bytes))
-                        .unwrap())
+                    Ok(proxy_helpers::forward_verbatim_metadata(
+                        bytes,
+                        None,
+                        &ct,
+                        content_encoding,
+                    ))
                 }
             },
         )
@@ -448,6 +452,12 @@ fn filter_version_list(body: &str, blocked: &std::collections::HashSet<String>) 
 /// source id is what the age-gate listing filter resolves policy from: for a
 /// virtual repository each member's own gate configuration governs its
 /// contribution (#2264).
+///
+/// Deliberately DROPS the upstream `Content-Encoding` (#3260): every caller
+/// PARSES the returned body (`@v/list` line filtering, `@latest` JSON) and
+/// serves a document it rebuilt itself, so the upstream coding never
+/// describes the bytes that leave the handler. Verbatim forwards go through
+/// [`try_proxy_go_metadata`] / `get_mod_file`, which re-declare the coding.
 async fn fetch_go_metadata_with_source(
     state: &SharedState,
     repo: &RepoInfo,
@@ -623,7 +633,8 @@ async fn enforce_go_zip_age_gate(
 
 /// Positive existence evidence for `module@version`: the upstream serves its
 /// `.info` document. Failures are treated as "no evidence" — the gate then
-/// blocks without starting a clock, never the reverse.
+/// blocks without starting a clock, never the reverse. The body (and hence
+/// its `Content-Encoding`, #3260) is discarded: only reachability matters.
 async fn go_version_exists_upstream(
     state: &SharedState,
     params: &crate::services::age_gate_service::AgeGateRepoParams,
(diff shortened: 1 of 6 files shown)


brandonrc:
## Independent validation — PR #3272 @ `a857d7b`

**Verdict: the fix is correct and the inverse failure does not exist — but do not merge as-is.** Three defects, all in the test/documentation layer, one of which asserts something about the production code that is measurably false.

Method: fresh shallow clone at `a857d7b`, own Postgres, `cargo nextest run --lib` with **both** `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1` (runtimes 0.44–0.63 s — the tests ran). All mutations below were compiled and run; the tree was restored between each.

---

### Claims, verified independently

| # | Claim | Verdict |
|---|---|---|
| 1 | 9 sites, not the issue's 5 | **CONFIRMED** |
| 2 | Callers that keep dropping the coding, now documented | **REFUTED** (see D3) |
| 3 | `resolve_virtual_metadata` widened in place | **CONFIRMED** |
| 4 | Nothing decodes | **CONFIRMED**, on every path incl. cache-hit and stale-if-error |
| 5 | Tests assert byte-identity and decode | **CONFIRMED but weak** (see D1, D2) |

**Claim 1 — CONFIRMED.** `grep -c 'proxy_helpers::forward_verbatim_metadata('` returns exactly 9: `goproxy.rs:390`, `goproxy.rs:412`, `goproxy.rs:906`, `rubygems.rs:136`, `rubygems.rs:812`, `hex.rs:354`, `hex.rs:423`, `hex.rs:1123`, `hex.rs:1279`. Finding the four beyond the issue's table is a good call.

**Claim 3 — CONFIRMED.** I enumerated the callers myself rather than trusting the table: `resolve_virtual_metadata` has exactly four call sites — `rubygems.rs:130`, `rubygems.rs:806`, `goproxy.rs:404`, `hex.rs:417` — and all four now declare the coding; none of the four parses, so none should be dropping it. The Pass-2 swap is behaviour-preserving: `ProxyService::fetch_artifact(repo, path)` is *literally* `fetch_artifact_with_cache_path(repo, path, path)` with the coding thrown away (`proxy_service.rs:2684-2693`), so `proxy_fetch` → `proxy_fetch_with_cache_key(.., path, path)` changes nothing but the third tuple element. Widening in place rather than adding a `_encoded` sibling is the right call.

**Claim 4 — CONFIRMED, and it holds beyond the two greps in the PR body.** This is the one that would have been catastrophic to get wrong, so I checked each layer rather than the client builder alone:

- `http_client.rs:93-121` — all four codecs disabled plus explicit `Accept-Encoding: identity`. As claimed.
- **Router layers.** The only `CompressionLayer` in the tree is npm's (`npm.rs:114`, applied *inside* `npm::router()`, scoped to `/npm`). There is no `DecompressionLayer` anywhere. `routes.rs:73,76,89` mount `/gems`, `/go`, `/hex` with no compression layer, and the global layers (`routes.rs:109-264`) are body-limit / guest-access / setup / correlation-id / load-shed only. None of the 9 routes can be re-encoded or decoded by middleware.
- **Cache-hit path** — `proxy_service.rs:3015` returns `CacheReadOutcome::Hit(content, ct, enc)` from `read_cached_with_revalidation:5150` → `get_cached:1198-1201`, which reads `metadata.content_encoding` off the sidecar.
- **304 revalidation** — `:5251`. **stale-if-error** — `:5264` and `:3253`. Both carry the coding.
- A sidecar cannot declare a coding for bytes it does not describe: `get_cached` re-verifies `checksum_sha256` over the body (`:1167`) and `cache_artifact` writes the coding and the content in one call (`:5583-5615`), so the two cannot drift apart.
- **Empirically**: I patched the goproxy test to probe the same virtual `.info` URI *twice* (cold → Pass-2, warm → Pass-1) and the cache-hit response re-declares `deflate` and returns the coded bytes byte-identically. Then I mutated Pass-1 to `transform(bytes, None, ..)` and that second probe went red — so the warm path genuinely runs and is genuinely correct.

**Content-Length — CONFIRMED on all 9.** `forward_verbatim_metadata` (`proxy_helpers.rs:697-706`) computes `content.len()` from the same `Bytes` value it moves into the body, so a CL/body mismatch is structurally impossible. The goproxy arms this PR does *not* touch (Virtual `.mod`, `.zip`) go through `build_streaming_response_with_disposition` (`proxy_helpers.rs:1015-1028`), which already sets the coded length and the coding — no gap there.

**hex `list_names` / `list_versions` and the #2658 signed-registry contract — risk REFUTED.** The bytes are untouched: the tests assert the client-received body equals the mock body through the full upstream → proxy → cache → handler stack, and `hex.rs:327-335` is unchanged in keeping Remote/Virtual a pass-through, so the client's pinned *upstream* signing key stays the one that validates. Declaring the coding is what makes AK's response identical to what `mix` would receive talking to the upstream directly — which is the definition of a transparent proxy, and the only defensible behaviour. It cannot be worse than upstream: if a client decoded at the HTTP layer and then hit `:zlib.gunzip/1`, it would already be broken against hex.pm itself. And because AK sends `Accept-Encoding: identity` upstream, the header can only appear at all when the origin object store has one *stored*, which is precisely the case that must be forwarded.

---

## Defects

### D1 — MEDIUM (hollow test): all three new tests pass with the coding pinned to a literal

`tdh::assert_coded_forward` hardcodes the expected value (`test_db_helpers.rs:1761-1766`, `Some("deflate")`) and `tdh::inflate_deflate` is deflate-only, so nothing in the fixture ties the *served* coding to the *upstream's* coding.

Proven by mutation. In `forward_verbatim_metadata` (`proxy_helpers.rs:706-708`):

```rust
-    if let Some(enc) = content_encoding {
-        builder = builder.header(axum::http::header::CONTENT_ENCODING, enc);
+    if content_encoding.is_some() {
+        builder = builder.header(axum::http::header::CONTENT_ENCODING, "deflate");
     }
```

Result: **574 / 574 pass** across the whole `goproxy | hex | rubygems | proxy_helpers | maven` suites, including all three new tests and all six uncoded controls. The controls do not help — the mutation stays inside the `is_some()` arm, so an uncoded upstream still grows no header.

This is exactly the regression `coded_fixture`'s own doc comment exists to prevent (`test_db_helpers.rs:1649-1658`: *"every #3149 fixture was gzip and every assertion was `== Some("gzip")`, so replacing the forwarded value with a hardcoded `"gzip"` literal left all fourteen tests green"*), and the consequence it names applies verbatim here: S3 commonly stores `br`, and a brotli body labelled `deflate` makes the client throw mid-decode — strictly worse than the bug being fixed.

**Fix:** give `assert_coded_forward` an `encoding: &str` parameter instead of the literal, and make at least one of the three suites use a second coding. `code_bytes` already supports `"gzip"`; add a `gunzip` next to `inflate_deflate` and select the decoder from the same parameter. Re-run the mutation above to confirm it goes red.

### D2 — MEDIUM (coverage gap on a line this PR changed): Pass-1 is untested

The PR changes the cache-hit arm at `proxy_helpers.rs:2753` (`Ok(Some((bytes, _ct, enc))) => transform(bytes, enc, ..)`) — a real behavioural change, since that arm previously discarded the coding. No shipped test reaches it: every probe in all three tests is a cold-cache first request.

Proven by mutation — `transform(bytes, None, member.key.clone())` in Pass-1 only:

```
PASS  rubygems::db_cov_tests::test_rubygems_virtual_forwards_upstream_content_encoding_verbatim_db
PASS  hex::tests::test_hex_registry_forwards_upstream_content_encoding_verbatim_db
```

Both green with the arm neutered. (Only my extra second-request probe caught it.) The arm is **correct** — I verified it end to end — so this is a test gap, not a bug, but it is a changed line with no guard.

**Fix:** two lines. Probe one virtual URI a second time in one of the three tests and assert the same coded forward, labelled `"virtual … cache hit"`. Cold → Pass-2, warm → Pass-1, both covered by one fixture.

### D3 — MEDIUM (claim 2 is false, and it blesses a live in-scope gap)

`goproxy.rs:456-460` now states that `fetch_go_metadata_with_source`'s callers *"PARSE the returned body … and serve a document it rebuilt itself"*. `latest_version` (`goproxy.rs:1155`) does not. It parses the JSON **only to run the age gate** and then serves the original upstream bytes verbatim:

```rust
// goproxy.rs:1219-1223
return Ok(Response::builder()
    .status(StatusCode::OK)
    .header(CONTENT_TYPE, "application/json")
    .body(Body::from(content))      // <- upstream bytes, untouched
    .unwrap());
```

Measured against the PR head (temporary probe, wiremock upstreams, since removed):

```
CODED @latest  -> status=404  content-encoding=None  len=27
PLAIN @latest  -> status=200  len=50  verbatim=true      <- proves it IS a verbatim forwarder
CODED @v/list  -> status=200  content-encoding=None  len=80 (upstream coded body was 48 bytes)  verbatim=false
```

Two consequences:

1. **`@latest` returns 404 for any coded upstream.** The mislabel is not reachable (`serde_json::from_slice` on coded bytes fails first, `goproxy.rs:1196-1205`), so the *symptom* differs from #3260 — but the endpoint is broken for exactly the upstream population #3260 is about, and the fix is the same one-line change to `proxy_fetch_capped_encoded`.
2. **`@v/list` returns 200 with a corrupted body.** `goproxy.rs:732` does `String::from_utf8_lossy(&content)`, and `filter_go_version_list` returns `body.to_string()` unchanged when the age gate is off (`goproxy.rs:519` / `:525`) — the default. So the "rebuild" is an identity pass-through through a lossy decoder: 48 coded bytes became 80 bytes of U+FFFD. The drop is technically defensible (the bytes emitted are not the bytes received) but the endpoint is unusable either way.

Neither is a regression from this PR. What *is* new is that the PR reviewed both, wrote a doc comment asserting a property the code does not have, and marked them deliberate — which is the comment that stops the next reader from looking.

**Fix (minimum):** correct the comment — say plainly that `latest_version` forwards the upstream bytes verbatim and only survives today because the JSON gate rejects coded bodies first. **Fix (proper, and in scope by this PR's own standard of "same shape, same format"):** route `fetch_go_metadata_with_source` through `proxy_fetch_capped_encoded` and either decode a coded body before parsing, or fail closed with 502 on an unrecognised coding.

---

## Nits

- **N1.** `forward_verbatim_metadata` is introduced as *"the general form of maven's `forward_root_verbatim`"*, but `maven::forward_root_verbatim` (`maven.rs:966-980`) is left in place as a near-identical duplicate. Fold it into the shared helper — that is the whole point of the shared helper, and it is a jscpd surface.
- **N2.** The seam was widened for the coding but still discards the member's `Content-Type` (`proxy_helpers.rs:2753`, `:2808` — both `_ct`). So the four Virtual verbatim forwards re-declare §8.4 while overriding §8.3 with a hardcoded literal: `hex.rs:426` serves a gzipped signed protobuf as `application/json`. Pre-existing, but if you are widening the seam so callers must state their intent, widen it once to `(body, content_type, content_encoding, member_key)` rather than twice.
- **N3.** RFC 9110 §12.5.3 residual: AK does not honour the *client's* `Accept-Encoding`, so a client that sent `identity` can now receive a declared coding. Strictly better than the pre-fix silent mislabel, and identical to what that client would get from the upstream directly — but it is the same class as #3273 and belongs in that issue.

---

## What was verified green

- Revert-proof reproduced independently. Dropping the coding at the sink (the pre-#3260 shape of all nine builders) fails all three: `left: None, right: Some("deflate")` for `remote .info`, `remote packages/pkg3260`, `virtual gem_info`. Maven's untouched `test_download_root_forwards_upstream_content_encoding_verbatim` stays green in the same run, so the failure is attributable to this change and not to the harness.
- All 9 sites do carry a coded assertion, plus 6 uncoded controls — the "3 tests for 9 sites" ratio is not the problem here; D1 and D2 are.
- No regressions: **1065 / 1065** pass on the pristine PR tree across `goproxy | hex | rubygems | proxy_helpers | maven | npm | pypi`.

Good calls worth keeping: finding four sites past the issue's table, and widening `resolve_virtual_metadata` in place so every caller has to state verbatim-vs-parse at compile time rather than by omission.

brandonrc:
## Review response — all three defects fixed, pushed as `4131186`

Fresh shallow clone at `a857d7b`, own Postgres, `cargo nextest run --lib` with **both** `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1`. Runtimes 0.27–0.58 s throughout, so the DB-backed tests ran rather than early-returning. The tree was restored between every mutation and is clean against the pushed commit. No production behaviour changed — the diff is tests, doc comments, and one duplicate helper folded away.

Taking the confirmed findings as read: 9 call sites, in-place seam widening, nothing decodes, `Content-Length` structurally correct, hex #2658 risk refuted.

---

### D1 — hollow test: FIXED, and killed in both directions

The fixture now carries the coding it mounted. `coded_and_plain_upstreams` returns a `CodedUpstreams` whose `encoding` field is the single source every assertion reads, so there is no literal left to pin. `assert_coded_forward` / `assert_plain_forward` became methods on it, and the body is decoded with a new `decode_coded(encoding, bytes)` — using **the coding the response declared**, not the one the fixture mounted, so a mislabel that somehow slipped past the header assertion still fails at the decode.

Per the `coded_fixture` doc's own prescription, the suites no longer agree on one coding: **goproxy and rubygems mount `deflate`, hex mounts `gzip`.** That is what makes the class of bug unsatisfiable rather than just this instance of it — no single hardcoded literal keeps the tree green.

Mutation, exactly as prescribed, in `forward_verbatim_metadata` (`proxy_helpers.rs:706-708`):

```rust
-    if let Some(enc) = content_encoding {
-        builder = builder.header(axum::http::header::CONTENT_ENCODING, enc);
+    if content_encoding.is_some() {
+        builder = builder.header(axum::http::header::CONTENT_ENCODING, "deflate");
     }
```

```
FAIL  hex::tests::test_hex_registry_forwards_upstream_content_encoding_verbatim_db
  assertion `left == right` failed: remote packages/pkg3260: the UPSTREAM's Content-Encoding
  must be re-declared, not some other coding (#3260)
    left: Some("deflate")   right: Some("gzip")
PASS  goproxy … / rubygems … / maven …
Summary: 4 tests run: 3 passed, 1 failed
```

And the same mutation with `"gzip"` instead — the direction a deflate-only fixture set would miss:

```
FAIL  goproxy …    left: Some("gzip")  right: Some("deflate")   (remote .info)
FAIL  rubygems …   left: Some("gzip")  right: Some("deflate")   (virtual gem_info (cold, Pass 2))
FAIL  maven …      left: Some("gzip")  right: Some("deflate")   (#3211 root, via the folded helper)
PASS  hex …
Summary: 4 tests run: 1 passed, 3 failed
```

Every hardcode is now caught by at least one suite, and no hardcode is caught by all of them — which is the property that was missing.

### D2 — untested changed line: FIXED, mutation isolates Pass 1

`goproxy` and `rubygems` now probe one Virtual URI **twice**: cold (Pass 2, the upstream fan-out) and warm (Pass 1, `cached_metadata_if_servable`). The barrier is `up.coded_hits(path)` — the coded upstream's request count for that exact path must be **unchanged** across the second probe, which proves the response came from the proxy cache rather than a second cold fan-out. Both the fixed and the coding-dropping shape of Pass 1 reach that barrier, so it cannot mask the regression it is there to expose.

Mutation — `transform(bytes, None, member.key.clone())` in Pass 1 **only**:

```
FAIL  rubygems …  assertion failed: virtual gem_info (warm, Pass 1): the UPSTREAM's
                  Content-Encoding must be re-declared …   left: None  right: Some("deflate")
FAIL  goproxy  …  assertion failed: virtual .info (warm, Pass 1): …
                  left: None  right: Some("deflate")
PASS  hex …       (no warm probe — correctly unaffected)
PASS  maven …
Summary: 4 tests run: 2 passed, 2 failed
```

The failing assertion is the one labelled `(warm, Pass 1)` in both cases, so the red is attributable to the cache-hit arm specifically and not to Pass 2. Uncoded warm controls were added in the same fixtures: a cache hit on an uncoded member must not invent a coding either.

### D3 — false doc claim: CORRECTED, and the gap it blessed is filed as #3280

Reproduced independently before touching anything — temporary probe on this branch, wiremock upstreams, deflate-coded vs byte-identical uncoded bodies, since removed:

```
upstream @latest plain=50 coded=48 | @v/list plain=21 coded=18

CODED @latest  -> status=404  content-encoding=None  len=27  verbatim=false
CODED @v/list  -> status=200  content-encoding=None  len=30  verbatim=false   <- 18 coded bytes served as 30 bytes of U+FFFD
PLAIN @latest  -> status=200  content-encoding=None  len=50  verbatim=true    <- proves @latest IS a verbatim forwarder
PLAIN @v/list  -> status=200  content-encoding=None  len=21  verbatim=true
```

Same shapes as reported (my body sizes differ only because the probe fixture differs). `PLAIN @latest` returning bytes identical to the upstream is the load-bearing line: `latest_version` is a verbatim forwarder, so the claim that every caller "serves a document it rebuilt itself" was false as written.

**Decision: filed as #3280, not folded in.** The comment now says what the code does:

> * `@v/list` does rebuild its document — but through `String::from_utf8_lossy`, so a coded body is not rejected, it is replaced character by character with U+FFFD and served as a 200. The coding is correctly dropped; the body is corrupt.
> * `@latest` does NOT rebuild anything. […] It does not currently mislabel a coded body only because `serde_json::from_slice` rejects one first and the request 404s — i.e. the endpoint is broken for exactly the coded-upstream population #3260 is about.

The reasoning for filing rather than folding: this PR's entire argument is that it changes no bytes, only the headers describing them. Neither goproxy path can be fixed that way — `@latest` 404s before the forward is reached and `@v/list` corrupts before the filter runs, so both need the body **decoded** before it is parsed. That introduces decompression of an upstream-controlled body on a metadata path, which needs a decompressed-size bound and an unknown-coding policy of its own. Bolting that onto a re-declare-only fix would make the "nothing decodes" property this PR rests on untrue of the very PR that asserts it. #3280 carries the measurements above and a suggested fix.

---

### Nits

- **N1 — folded in.** `maven::forward_root_verbatim` is now a one-line delegate to `proxy_helpers::forward_verbatim_metadata` with maven's `text/html` default; the duplicate body is gone. Useful side effect visible above: maven's pre-existing `#3211` test now guards the shared helper too, and it goes red under the D1 mutations.
- **N2 — recorded, filed as #3281, not folded in.** Right that the seam should be widened once rather than twice, but this one is a **client-visible behaviour change** (hex's Virtual arm would stop serving `application/json` and start serving the upstream's type, correctly disagreeing with what clients see today), so it needs a test per arm and a fallback story for upstreams that declare nothing. #3281 has the four affected call sites and the Remote/Virtual inconsistency. `resolve_virtual_metadata`'s doc now names the drop and points there, so it is not left blessed by omission — which was D3's actual lesson.
- **N3 — agreed, belongs to #3273.** Not touched here.

### Regression status

- `goproxy | hex | rubygems | proxy_helpers | maven | npm | pypi`: **1776 / 1776 pass**.
- Revert-proof re-run after these changes (coding dropped at the sink, the pre-#3260 shape of all nine builders): **all four go red**, including maven's via the folded helper. Output is in the PR body.
- `cargo fmt --all` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean.

Issue #3108, opened 2026-08-04
OCI: Referrers API (/v2/<name>/referrers/<digest>) not implemented


brandonrc:
`parse_oci_path` has no `referrers` branch, so the OCI 1.1 Referrers discovery API 404s. This fails 2 distribution-spec conformance content-discovery tests (live run: 59 pass / 6 fail). Feature gap.

---
Surfaced by the conformance corpus (artifact-keeper-test#333). 🤖 filed via Claude Code.

PR #3274, opened 2026-08-10
fix(oci): emit the spec error envelope on scope-denial and curation-block /v2 paths

### LINKED ISSUE #3110: OCI: some 400 error responses are not OCI-conforming JSON
The distribution-spec conformance test 'Pull / Error codes / 400 response body should contain OCI-conforming JSON message' fails: some 400 paths do not emit the spec's {"errors":[{"code","message","detail"}]} body shape. Also noted: chunked-range errors return 416 where the spec pins SIZE_INVALID/400.

---
Surfaced by the conformance corpus (artifact-keeper-test#333). 🤖 filed via Claude Code.

brandonrc:
Fixes #3110

Companion (**merge this second** — see [Merge order](#merge-order)): artifact-keeper/artifact-keeper-test#337.

> **Rebased onto `8b9e0ff`.** The original merge-base was `1115203`; #3268, #3272, #3277 and #3278 landed underneath. Conflict resolution and the post-rebase re-audit are described in [Rebase](#rebase) at the end.

## What the issue claimed, and what verification showed

#3110 makes two claims. Both were verified against a live build of current main (`1115203`) by running the pinned upstream conformance suite (`opencontainers/distribution-spec` v1.1.0, the same suite and pin the corpus oracle `deploy-test/harness/tiers/oci-conformance/oracle.sh` adopts) against a freshly migrated stack:

1. **"The conformance test 'Pull / Error codes / 400 response body should contain OCI-conforming JSON message' fails."** — **Not reproducible at current main.** The suite run finished 59 passed / 6 failed, and this test **passes**: `GET /v2/<name>/manifests/sha256:totallywrong` returns `404 {"errors":[{"code":"MANIFEST_UNKNOWN","message":"manifest not found"}]}`, which the test accepts (its envelope assertions sit inside an `if StatusCode == 400`, and a 404 satisfies the `SatisfyAny(400, 404)` check unconditionally — `conformance/01_pull_test.go:250-268`). The 6 real failures are unrelated functionality gaps, not body-shape bugs: 5 are the Referrers API (already tracked as #3108), 1 is the upload-status GET (#3275).

2. **"Chunked-range errors return 416 where the spec pins SIZE_INVALID/400."** — **Refuted by the spec text.** distribution-spec v1.1.0 `spec.md` (Pushing a blob in chunks): *"If a chunk is uploaded out of order, the registry MUST respond with a `416 Requested Range Not Satisfiable` code."* The endpoints table pins end-5 (`PATCH .../blobs/uploads/<reference>`) failure codes as `404`/`416`, and the conformance suite itself asserts 416 ("Out-of-order blob upload should return 416", "Retry previous blob chunk should return 416" — both **pass** against main, and both of AK's 416 arms already carry the `BLOB_UPLOAD_INVALID` envelope). `SIZE_INVALID` (code-10) belongs to length-mismatch cases such as end-6, not to out-of-order chunks. **No status codes are changed by this PR.**

## What was actually wrong (full enumeration of /v2 error paths)

Auditing every error return on the `/v2` surface at current main (`oci_error` call sites, raw `Response::builder` sites, shared-gate returns, extractors, middleware):

**Conforming already** (no change): all `catch_all` dispatch errors, `resolve_repo*` (NAME_UNKNOWN/INTERNAL_ERROR incl. 503 shed), digest validation (DIGEST_INVALID), manifest classification (MANIFEST_INVALID), pagination (400 `PAGINATION_NUMBER_INVALID` — see the note below), all five 416 range arms (BLOB_UPLOAD_INVALID), 401 challenge (UNAUTHORIZED + WWW-Authenticate), quarantine block (DENIED), scan gates (`oci_scan_deny_response`: DENIED 403/423), blob scan re-block (DENIED), token endpoint errors (DENIED), 405 fallback (UNSUPPORTED). There are zero `.into_response()` calls in the file's production code, and every `if let Err(resp) = <gate>` helper returns an envelope.

**Non-conforming — fixed here:**

- `oci_forbidden_scope` (GHSA-vvc3 API-token scope gate; 6 call sites: monolithic blob POST, upload PATCH, upload PUT, blob DELETE via upload-session, manifest PUT, manifest DELETE) returned a **bare-string 403** `Token does not have required scope: ...`. Now `{"errors":[{"code":"DENIED","message":"Token does not have required scope: ..."}]}` — status unchanged, legacy text preserved verbatim inside the envelope (`test-ghsa-vvc3-scope-checks.sh:317` greps the raw body and still matches).
- The shared curation gate (called from manifest GET and HEAD) leaked the REST-shaped **`{"error":"curation_blocked","package":…,"reason":…}`** 403 body onto the OCI wire. Per the spec's Error Codes section, *"If the response body is in JSON format, it MUST have the following format"* (`{"errors":[...]}`) — a differently-shaped JSON object is a MUST violation. The OCI seam now renders the verdict itself via `oci_curation_deny_response` (sibling of `oci_scan_deny_response`): code `DENIED`, the `curation_blocked` marker kept in the message, **and the rule's `reason` carried through** (see below).
- `/v2/token`'s two **fallible** extractors. `Query<TokenQuery>` + `Bytes` produce axum's own `text/plain` rejections: `?service=a&service=b` yields `400 Failed to deserialize query string: duplicate field \`service\`` (the serde_urlencoded mechanism the module comment at `oci_v2.rs:3741-3753` already documents for `scope`), and a body over `TOKEN_REQUEST_BODY_LIMIT_BYTES` yields a plaintext 413. Both are now re-shaped into the envelope with the rejection's own status and text preserved.

## Changes since the independent review ([comment](https://github.com/artifact-keeper/artifact-keeper/pull/3274#issuecomment-5245426938))

The review confirmed claims 1, 2, 3 and 5, reproduced the revert-proof independently, and confirmed the handler-surface enumeration is complete. It found one blocking defect, one observability regression, and two inaccurate statements in this description. All four are addressed:

### 1. The DTF `curation-multiformat-enforce` oracle does **not** still match — fixed in the companion PR

This description previously claimed the `curation_blocked` marker in the message kept that oracle matching. **That was wrong.** The oracle is a JSON field lookup, not a grep (`oracle.sh:142-143`): `jq -r '.error // empty'` yields `''` against the new body, so the docker arm of a **blocking** gate would have gone red on the first release carrying this PR.

Fixed in **artifact-keeper/artifact-keeper-test#337**, validated red→green locally before opening against bodies captured from this backend (not hand-written). The oracle now selects the expected shape from the deployed backend version — envelope **required** at ≥ 1.7.2, either shape accepted below it, since that oracle runs at `@main` against 1.7.x cuts that predate this fix — and asserts it with `jq` field lookups (`errors[0].code`, `errors[0].message`, `errors[0].detail.reason`), never a substring match on the raw body.

The other consumer, `test-ghsa-vvc3-scope-checks.sh`, greps the raw body and survives; independently re-verified. `curation_blocked` has no other consumer in either repo.

### 2. The curation rule `reason` was being discarded — fixed here

At `301f3d4` the OCI body dropped `reason` and hardcoded `detail: None`, so an operator debugging a blocked `docker pull` got "blocked by this repository's curation policy" while every other proxy format still reported which rule fired. The reason was unreachable only because the shared seam handed callers an already-rendered `Response`.

Split at the seam: `proxy_helpers::evaluate_curation{,_lookup}` now return a structured `CurationBlock { package, reason }`, and `enforce_curation{,_lookup}` are thin wrappers that render it through `CurationBlock::into_rest_response`. Every non-OCI format's wire body is byte-identical (pinned by a new test). The OCI seam builds its own envelope from the structured verdict:

```json
{"errors":[{"code":"DENIED",
            "message":"curation_blocked: pull of library/hello-world is blocked by this repository's curation policy: #2930 tier block",
            "detail":{"package":"library/hello-world","reason":"#2930 tier block"}}]}
```

`spec.md` "Error Codes": *"The `detail` field is OPTIONAL and MAY contain arbitrary JSON data providing information the client can use to resolve the issue."* The reason is in **both** `message` (what `docker pull` prints) and `detail` (what a machine reads). `oci_error` gained an `oci_error_detail` sibling and is otherwise byte-identical — `detail` is `skip_serializing_if = "Option::is_none"`, pinned by `test_oci_error_omits_detail_when_absent`.

### 3. "Extractors in the `/v2` router are `Infallible`" — that claim was **false**, and the underlying gap is now fixed

Confirmed against the running router, not just by reading: `GET /v2/token?service=a&service=b` returned `400 text/plain` *"Failed to deserialize query string: duplicate field `service`"*, while `?service=artifact-keeper` returned `200` with a token. `Query<TokenQuery>` and `Bytes` are both fallible; only `catch_all` / `handle_catalog`'s `Query<HashMap<String,String>>` is effectively infallible. The sentence has been removed from the enumeration above.

Neither rejection was a spec *violation* — `spec.md`: a 4XX *"MAY return a body in any format"* — but they were the last bare-string 4XX bodies a docker client could meet on this router, which is the same ergonomics argument the rest of this PR makes, so they are re-shaped rather than merely documented. Status and axum's explanatory text are preserved; `UNSUPPORTED` (code-14, *"The operation is unsupported."*) is the closest registered code, as the closed table has no malformed-request entry.

### 4. `PAGINATION_NUMBER_INVALID` — "cosmetic" was the wrong justification

`spec.md` "Error Codes" reads *"The `code` field MUST be one of the following:"* over a closed 14-row table, and `PAGINATION_NUMBER_INVALID` is not in it, so calling the mismatch cosmetic was not defensible. The real reason to leave it is the reference implementation, which is what clients are written against: `distribution/distribution` registers exactly this code, string and status —

```go
// registry/api/errcode/register.go:219-226 (main); same in registry/api/v2/errors.go:137-144 on release/2.8
ErrorCodePaginationNumberInvalid = register(errGroup, ErrorDescriptor{
    Value:          "PAGINATION_NUMBER_INVALID",
    Message:        "invalid number of results requested",
    ...
    HTTPStatusCode: http.StatusBadRequest,
})
```

AK's body is already the conforming envelope; only the code is an upstream-compatible extension. Changing it would move away from the behaviour Docker Hub-oriented clients see. Left as-is, on that ground.

## Known residuals (deliberately not in this diff)

- **`setup_guard` / `demo_guard` emit REST-shaped JSON 4XX on `/v2`.** `routes.rs:146` nests `/v2` inside the global layer stack, so a request can be refused before reaching any handler. `setup_guard` (`middleware/setup.rs:32-79`) has no `/v2` entry in its allowlist and 403s the `docker login` probe with `{"error":"SETUP_REQUIRED",…}` while setup is pending; `demo_guard` (`middleware/demo.rs:71-77`) does the same on write verbs in demo mode. Both are narrow states and both are middleware, not handler surface — filed as **#3284** rather than grown into this PR. (`guest_access_guard` is already correctly allowlisted for `/v2`.)
- Referrers API 404s (#3108); upload-status GET 405 (#3275).
- The shared REST bodies of `curation_blocked_response` / `scan_blocked_response` for non-OCI formats — their consumers expect that shape, and this PR now pins it with a test.

## Spec grounding

OCI Distribution Specification (opencontainers/distribution-spec, `spec.md` § "Error Codes"): *"A `4XX` response code from the registry MAY return a body in any format. If the response body is in JSON format, it MUST have the following format: `{"errors":[{"code":..., "message":..., "detail":...}]}`. The `code` field MUST be a unique identifier, containing only uppercase alphabetic characters and underscores"* — with `DENIED` registered as code-12 ("requested access to the resource is denied") and `UNSUPPORTED` as code-14.

## Tests (7 new/rewritten, all `--lib` so the coverage gate sees them)

- `tests::test_oci_forbidden_scope_status_and_body` — rewritten: previously asserted the bare-string body *as the contract*; now asserts the **parsed** envelope (`errors[0].code == "DENIED"`, content-type, legacy message preserved).
- `tests::test_oci_curation_deny_response_is_oci_error_envelope` — envelope + 403 + `curation_blocked` marker + `detail.reason` / `detail.package`, all parsed.
- `tests::test_oci_error_omits_detail_when_absent` — positive control on the `detail` plumbing: `oci_error`'s output is byte-for-byte what it was.
- `proxy_helpers::tests::curation_block_rest_body_is_unchanged` — the shared REST body every other format serves is byte-identical after the seam split. This is what stops the `evaluate`/`enforce` refactor from silently moving npm's wire shape (which the DTF tier's npm arm reads).
- `oci_error_envelope_db_tests::scope_denied_manifest_put_returns_oci_error_envelope` (DB, real router): read-scoped API token PUT manifest → 403 envelope DENIED; **positive control in the same fixture**: wildcard-scoped token, byte-identical request → 201 CREATED.
- `oci_error_envelope_db_tests::curation_blocked_manifest_get_returns_oci_error_envelope` (DB, real router): curation-blocked image GET → 403 envelope DENIED with marker **and `detail.reason == "blocked by test rule"`** from the actual rule row; **positive control**: non-matching image on the same repo passes the gate and its miss is the spec-shaped `404 MANIFEST_UNKNOWN` envelope.
- `oci_error_envelope_db_tests::token_query_rejection_returns_oci_error_envelope` (DB, real router): `?service=a&service=b` → 400 envelope `UNSUPPORTED` carrying axum's own text; **positive control**: `?service=artifact-keeper` still mints a token, so the assertion cannot be satisfied by a route that rejects everything.

Every assertion goes through `serde_json::from_slice` and indexes `errors[0]…`; none is a substring match on the raw body. That matters concretely: the REST body this PR removes contains the literal `curation_blocked` **and** the reason at the top level, so a whole-body `contains` check would pass on the shape being removed.

## Revert-proof (run with `DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1`)

Three independent reverts to the pre-fix shape in one tree — the curation body back to `301f3d4` (reason discarded, no `detail`), `/v2/token` back to `Query(query): Query<TokenQuery>, body: Bytes`, and the shared REST body's `"error"` key moved. Exactly the four guards that cover them fail, with real DB timings; the three that cover the untouched scope gate keep passing:

```
FAIL [ 0.013s] proxy_helpers::tests::curation_block_rest_body_is_unchanged
  left:  "{\"curation_error\":\"curation_blocked\",\"package\":\"left-pad\",…}"
  right: "{\"error\":\"curation_blocked\",\"package\":\"left-pad\",…}"

FAIL [ 0.018s] oci_v2::tests::test_oci_curation_deny_response_is_oci_error_envelope
  message must name the rule that fired (docker prints message):
  {"errors":[{"code":"DENIED","message":"curation_blocked: pull of someimage is blocked by this repository's curation policy"}]}

FAIL [ 0.280s] oci_error_envelope_db_tests::token_query_rejection_returns_oci_error_envelope
  body must be the OCI error envelope with code UNSUPPORTED;
  got: Failed to deserialize query string: duplicate field `service`

FAIL [ 3.883s] oci_error_envelope_db_tests::curation_blocked_manifest_get_returns_oci_error_envelope
  detail.reason must name the curation rule that fired:
  {"errors":[{"code":"DENIED","message":"curation_blocked: pull of blockedimg is blocked by this repository's curation policy"}]}
  left: None   right: Some("blocked by test rule")

PASS [ 4.842s] oci_error_envelope_db_tests::scope_denied_manifest_put_returns_oci_error_envelope
PASS [ 0.030s] oci_v2::tests::test_oci_forbidden_scope_status_and_body
PASS [ 0.030s] oci_v2::tests::test_oci_error_omits_detail_when_absent
```

**3.8s and 4.9s, not 0.04s** — the DB guards executed against a real database in both directions. Fixed tree: 7/7 pass. Both directions re-run **after** the rebase, on the rebased tree.

Full `oci_v2` + `proxy_helpers` battery on the rebased fixed tree: **795/795** (up from 788 — #3268 added a module). Whole backend `--lib` suite on the rebased tree: **15104/15104 passed, 0 failed** — run because the `evaluate`/`enforce` seam split touches a helper with 9 callers across the format handlers, so the blast radius is wider than the two files in the diff. `cargo fmt --all --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean. No `sqlx::query!` macro text touched (runtime queries only), no migrations.

## Merge order

The DTF release gate is invoked **only** on a `v*` tag push (`release.yml` → `artifact-keeper-test/.github/workflows/release-gate.yml@main`) and by `release-gate-rehearsal.yml`. There is no PR-level DTF gate, so no merge into `main` turns it red by itself — but the order still matters for the next tag:

1. **artifact-keeper/artifact-keeper-test#337 first.** It is green against every backend that exists today (1.7.1 and older take its below-floor branch and pass on the REST body), so it is a no-op until a backend carrying this PR is deployed.
2. **Then this PR.**
3. Both before the next `v*` tag. The reverse order leaves a window in which a release cut runs the old `.error` lookup against the new envelope and fails a blocking gate.

If this PR slips out of milestone 1.7.2, `CURATION_OCI_ENVELOPE_MIN_VERSION` in #337 must move with it.

## Rebase

Rebased from merge-base `1115203` onto `8b9e0ff`. What landed underneath and what it cost:

- **#3268** (`f75f010`) — `require_oci_repo_read_access` on GET/HEAD blob, GET/HEAD manifest and `GET tags/list`; `authorize_virtual_members` on the OCI read path; `enforce_scan_pull_scope` / `enforce_token_repo_scope` moved to run first.
- **#3272** (`118a8b0`) — `resolve_virtual_metadata`'s transform signature is now `(body, content_encoding, member_key)`, plus `forward_verbatim_metadata`.

**Conflicts: two, both in `oci_v2.rs`, both "we each appended a new test module at EOF"** — #3268's `oci_read_authz_tests` against this PR's `oci_error_envelope_db_tests`. Resolved by keeping both, each with its own header comment and attributes. `proxy_helpers.rs` merged without conflict (#3272's change is ~2,800 lines away from the curation seam).

**Verified rather than assumed**, because a clean auto-merge is not evidence:

- Every deleted line in `git diff 8b9e0ff` across both files was audited one by one. **All 21 are this PR's own** (the bare-string `oci_forbidden_scope` body, the two `return resp;` arms, the old `Query`/`Bytes` extractor params, the old `detail: None`, the two `enforce_curation_lookup` call heads, the pre-rewrite test asserts, and the two seam-internal lines). Nothing from #3268 or #3272 was dropped or rewritten.
- Gate ORDERING preserved: `enforce_scan_pull_scope` → `enforce_token_repo_scope` → `require_oci_repo_read_access` → the curation seam, in both `handle_get_manifest` and `handle_head_manifest`. The curation gate stays last, where it was.
- The negative-cache constraint in `authorized_virtual_members` is byte-identical: `VirtualResolveKey` carries no principal term, so the cache is still consulted and populated only when `members.len() == total`. This PR never enters that region — confirmed by the deletion audit above, not by inspection alone.
- **The `/v2` enumeration was re-run against the new code.** #3268's error returns are `oci_denied_repo_access()` (403 DENIED), `oci_error(404, "NAME_UNKNOWN", …)` and `oci_authorization_unavailable()` (503 DENIED) — all already the envelope, so nothing new to fix and the "conforming already" list above holds. A denied virtual member is dropped from the walk, so it surfaces as the existing `MANIFEST_UNKNOWN` 404 envelope.
- `cargo fmt --all --check`, `clippy --workspace --all-targets -D warnings`, the 795-test battery and the full revert-proof were all re-run **after** the rebase, on the rebased tree, with `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1`. The numbers quoted above are the post-rebase ones.

The DTF companion (#337) is unaffected: it changes only `artifact-keeper-test`, and the wire body it asserts is unchanged by the rebase.

### DIFF
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -455,14 +455,20 @@ fn oci_scopes_grant(scopes: &Option<Vec<String>>, required: &str) -> bool {
 
 /// Build a 403 Forbidden response when an API-token scope check fails on
 /// an OCI write/delete path. See GHSA-vvc3-h39c-mrq5.
+///
+/// Uses the OCI error envelope (#3110): the distribution-spec "Error Codes"
+/// section requires any JSON body on a 4XX to be
+/// `{"errors":[{"code","message","detail"}]}`, and a bare-string body is
+/// opaque to conforming clients (docker/podman surface `errors[0].message`).
+/// `DENIED` is the registered code (code-12, "requested access to the
+/// resource is denied"). The legacy message text is preserved verbatim
+/// inside the envelope so operators/scripts that match on it keep working.
 fn oci_forbidden_scope(required: &str) -> Response {
-    Response::builder()
-        .status(StatusCode::FORBIDDEN)
-        .body(Body::from(format!(
-            "Token does not have required scope: {}",
-            required
-        )))
-        .unwrap()
+    oci_error(
+        StatusCode::FORBIDDEN,
+        "DENIED",
+        &format!("Token does not have required scope: {}", required),
+    )
 }
 
 /// Build a 403 Forbidden response when a caller is authenticated but lacks
@@ -7404,7 +7410,9 @@ async fn handle_head_manifest(
     )
     .await
     {
-        return resp;
+        // #3110: re-shape the shared gate's REST-style body into the OCI
+        // error envelope this surface is contractually required to emit.
+        return oci_curation_deny_response(resp.status(), &repo.image);
     }
 
     // Reference can be a tag or a digest. Resolve locally first: a surviving
@@ -7880,6 +7888,25 @@ fn oci_manifest_synthetic_artifact(
     }
 }
 
+/// Map the shared curation gate's deny response onto the OCI error envelope
+/// (#3110). `proxy_helpers::enforce_curation_lookup` builds a REST-shaped
+/// `{"error":"curation_blocked",...}` body shared by every proxy format; on
+/// the `/v2` surface that shape violates the distribution-spec "Error Codes"
+/// MUST (a 4XX JSON body must be `{"errors":[...]}`) and docker clients
+/// cannot render it. Status is preserved; the `curation_blocked` marker is
+/// kept inside the message so existing oracles/scripts that grep for it
+/// still match. Sibling of [`oci_scan_deny_response`].
+fn oci_curation_deny_response(status: StatusCode, image: &str) -> Response {
+    oci_error(
+        status,
+        "DENIED",
+        &format!(
+            "curation_blocked: pull of {} is blocked by this repository's curation policy",
+            image
+        ),
+    )
+}
+
 /// Map the shared gate's deny response onto the OCI error body shape docker
 /// clients surface. Status is preserved: 403 = cached/inline vulnerable
 /// verdict, 423 = inconclusive under fail-closed (scan pending/failed —
@@ -8269,7 +8296,9 @@ async fn handle_get_manifest(
     )
     .await
     {
-        return resp;
+        // #3110: re-shape the shared gate's REST-style body into the OCI
+        // error envelope this surface is contractually required to emit.
+        return oci_curation_deny_response(resp.status(), &repo.image);
     }
 
     // Resolve locally first: a surviving tag row, or — for a digest this hosted
@@ -10792,16 +10821,54 @@ mod tests {
 
     #[tokio::test]
     async fn test_oci_forbidden_scope_status_and_body() {
-        // The body string is part of the contract; clients (and clients of
-        // clients) parse it to know whether they hit an auth wall.
+        // #3110: on the /v2 surface the error body is contractually the OCI
+        // error envelope (distribution-spec "Error Codes": a 4XX JSON body
+        // MUST be `{"errors":[{"code","message","detail"}]}` with a
+        // registered code). Assert the PARSED structure, not a substring —
+        // a substring match would pass on the pre-#3110 bare-string body.
         let resp = oci_forbidden_scope("write");
         assert_eq!(resp.status(), StatusCode::FORBIDDEN);
+        assert_eq!(
+            resp.headers()
+                .get(CONTENT_TYPE)
+                .and_then(|v| v.to_str().ok()),
+            Some("application/json"),
+            "OCI error responses must be application/json"
+        );
         let body = axum::body::to_bytes(resp.into_body(), 4096).await.unwrap();
-        let s = String::from_utf8_lossy(&body);
+        let parsed: serde_json::Value =
+            serde_json::from_slice(&body).expect("OCI error body must be valid JSON");
+        assert_eq!(
+            parsed["errors"][0]["code"].as_str(),
+            Some("DENIED"),
+            "scope denial must use the registered DENIED code (code-12): {parsed}"
+        );
+        // The legacy message text is preserved verbatim inside the envelope
+        // (test-ghsa-vvc3-scope-checks.sh greps the raw body for it).
         assert!(
-            s.contains("Token does not have required scope: write"),
-            "unexpected body: {}",
-            s
+            parsed["errors"][0]["message"]
+                .as_str()
+                .unwrap_or_default()
+                .contains("Token does not have required scope: write"),
+            "unexpected message: {parsed}"
+        );
+    }
+
+    #[tokio::test]
+    async fn test_oci_curation_deny_response_is_oci_error_envelope() {
+        // #3110: the shared curation gate's `{"error":"curation_blocked"}`
+        // REST shape must never reach the /v2 wire; the OCI mapping keeps the
+        // status and re-shapes the body into the spec envelope.
+        let resp = oci_curation_deny_response(StatusCode::FORBIDDEN, "someimage");
+        assert_eq!(resp.status(), StatusCode::FORBIDDEN);
+        let body = axum::body::to_bytes(resp.into_body(), 4096).await.unwrap();
+        let parsed: serde_json::Value =
+            serde_json::from_slice(&body).expect("OCI error body must be valid JSON");
+        assert_eq!(parsed["errors"][0]["code"].as_str(), Some("DENIED"));
+        let message = parsed["errors"][0]["message"].as_str().unwrap_or_default();
+        assert!(
+            message.contains("curation_blocked") && message.contains("someimage"),
+            "message must keep the curation_blocked marker and name the image: {parsed}"
         );
     }
 
@@ -29454,5 +29521,237 @@ mod oci_read_authz_tests {
             String::from_utf8_lossy(&granted_private_body).contains(MANIFEST_MARKER),
             "the granted principal's 200 must carry the private manifest bytes"
         );
+    }
+}
+
+// ---------------------------------------------------------------------------
+// OCI error-envelope conformance through the router (#3110).
+//
+// The distribution-spec "Error Codes" section pins the error body contract:
+// a 4XX response body that is JSON MUST be
+// `{"errors":[{"code","message","detail"}]}` with `code` drawn from the
+// registered set. Two /v2 deny paths violated it: the GHSA-vvc3 API-token
+// scope gate returned a bare-string 403, and the shared curation gate
+// returned the REST-shaped `{"error":"curation_blocked",...}` body. These
+// tests drive both through the real router and assert the PARSED envelope
+// (never a substring, which the malformed shapes would also satisfy), each
+// with a positive control in the same fixture so an over-blocking gate
+// cannot satisfy the deny assertion by failing everything.
+// ---------------------------------------------------------------------------
+
+#[allow(clippy::disallowed_methods)]
+// streaming-invariant: test module exempt — buffering response bodies in test assertions is not an artifact path (#1608)
+#[cfg(test)]
+mod oci_error_envelope_db_tests {
+    use super::*;
+    use crate::api::handlers::test_db_helpers as tdh;
+    use crate::services::auth_service::AuthService;
+    use axum::http::Request;
+    use std::sync::Arc;
+    use tower::ServiceExt;
+
+    const IMAGE_MANIFEST: &[u8] = br#"{"schemaVersion":2,"mediaType":"application/vnd.oci.image.manifest.v1+json","config":{"mediaType":"application/vnd.oci.image.config.v1+json","digest":"sha256:3110311031103110311031103110311031103110311031103110311031103110","size":4},"layers":[]}"#;
+
+    async fn send(app: Router, request: Request<Body>) -> (StatusCode, serde_json::Value, Bytes) {
+        let resp = app.oneshot(request).await.expect("oneshot");
+        let status = resp.status();
+        let body = to_bytes(resp.into_body(), 1024 * 1024)
+            .await
+            .expect("response body");
+        let parsed = serde_json::from_slice(&body).unwrap_or(serde_json::Value::Null);
+        (status, parsed, body)
+    }
+
+    fn assert_oci_envelope(parsed: &serde_json::Value, raw: &Bytes, expected_code: &str) {
+        let code = parsed["errors"][0]["code"].as_str();
+        assert_eq!(
+            code,
+            Some(expected_code),
+            "body must be the OCI error envelope with code {expected_code}; got: {}",
+            String::from_utf8_lossy(raw)
+        );
+    }
+
+    /// GHSA-vvc3 scope gate: a read-scoped API token on `docker push`'s
+    /// manifest PUT must be refused with the OCI envelope (403 DENIED), and
+    /// a wildcard-scoped token on the byte-identical request must succeed.
+    #[tokio::test]
+    async fn scope_denied_manifest_put_returns_oci_error_envelope() {
+        let Some(fx) = tdh::Fixture::setup("local", "docker").await else {
+            return;
+        };
+        tdh::grant_repo_actions(&fx.pool, fx.repo_id, fx.user_id, &["read", "write"]).await;
+        let auth_service = AuthService::new(fx.state.db.clone(), Arc::new(fx.state.config.clone()));
+        let (read_token, _) = auth_service
+            .generate_api_token(
+                fx.user_id,
+                "envelope-read",
+                vec!["read:artifacts".to_string()],
+                None,
+            )
+            .await
+            .expect("generate read-scoped API token");
+        let (full_token, _) = auth_service
+            .generate_api_token(fx.user_id, "envelope-full", vec!["*".to_string()], None)
+            .await
+            .expect("generate wildcard API token");
+
+        let uri = format!("/{}/image/manifests/v3110", fx.repo_key);
+        let put = |token: String, uri: String| {
+            Request::builder()
+                .method(Method::PUT)
+                .uri(uri)
+                .header(
+                    AUTHORIZATION,
+                    format!(
+                        "Basic {}",
+                        base64::engine::general_purpose::STANDARD
+                            .encode(format!("{}:{}", fx.username, token))
+                    ),
+                )
+                .header(CONTENT_TYPE, "application/vnd.oci.image.manifest.v1+json")
+                .body(Body::from(IMAGE_MANIFEST))
+                .expect("build request")
+        };
+
+        let (denied_status, denied_json, denied_raw) = send(
+            router().with_state(fx.state.clone()),
+            put(read_token, uri.clone()),
+        )
+        .await;
+
+        let (allowed_status, _allowed_json, allowed_raw) = send(
+            router().with_state(fx.state.clone()),
+            put(full_token, uri.clone()),
+        )
+        .await;
+
+        let _ = sqlx::query("DELETE FROM api_tokens WHERE user_id = $1")
+            .bind(fx.user_id)
+            .execute(&fx.pool)
+            .await;
+        for table in ["oci_manifest_refs", "manifest_blob_refs", "oci_tags"] {
+            let _ = sqlx::query(&format!("DELETE FROM {} WHERE repository_id = $1", table))
+                .bind(fx.repo_id)
+                .execute(&fx.pool)
+                .await;
+        }
+        fx.teardown().await;
+
+        assert_eq!(
+            denied_status,
+            StatusCode::FORBIDDEN,
+            "read-scoped token must be refused on manifest PUT: {}",
+            String::from_utf8_lossy(&denied_raw)
+        );
+        assert_oci_envelope(&denied_json, &denied_raw, "DENIED");
+        assert!(
+            denied_json["errors"][0]["message"]
+                .as_str()
+                .unwrap_or_default()
+                .contains("Token does not have required scope"),
+            "envelope message must carry the scope-denial text: {denied_json}"
+        );
+        // Positive control, same fixture, byte-identical request: the gate
+        // must open for a token that HAS the scope — otherwise the deny
+        // assertion above would also be satisfied by a gate that rejects
+        // everything.
+        assert_eq!(
+            allowed_status,
+            StatusCode::CREATED,
+            "wildcard-scoped token must complete the same PUT: {}",
+            String::from_utf8_lossy(&allowed_raw)
+        );
+    }
+
+    /// Curation block on a proxy pull: the manifest GET for a blocked image
+    /// must return the OCI envelope (403 DENIED, message carrying the
+    /// `curation_blocked` marker), while a non-blocked image on the same
+    /// repository falls through the gate (and its miss is itself the
+    /// spec-shaped 404 MANIFEST_UNKNOWN envelope).
+    #[tokio::test]
+    async fn curation_blocked_manifest_get_returns_oci_error_envelope() {
+        let Some(fx) = tdh::Fixture::setup("remote", "docker").await else {
+            return;
+        };
+        sqlx::query(
+            "UPDATE repositories SET curation_enabled = true, curation_default_action = 'allow' \
+             WHERE id = $1",
+        )
+        .bind(fx.repo_id)
+        .execute(&fx.pool)
+        .await
+        .expect("enable curation on fixture repo");
+        sqlx::query(
+            "INSERT INTO curation_rules \
+                 (staging_repo_id, package_pattern, version_constraint, architecture, \
+                  action, priority, reason, enabled, created_by) \
+             VALUES ($1, 'blockedimg*', '*', '*', 'block', 10, 'blocked by test rule', true, $2)",
+        )
+        .bind(fx.repo_id)
+        .bind(fx.user_id)
+        .execute(&fx.pool)
+        .await
+        .expect("insert curation block rule");
+
+        let auth_service = AuthService::new(fx.state.db.clone(), Arc::new(fx.state.config.clone()));
+        let (api_token, _) = auth_service
+            .generate_api_token(fx.user_id, "envelope-curation", vec!["*".to_string()], None)
+            .await
+            .expect("generate API token");
+        let basic = format!(
+            "Basic {}",
+            base64::engine::general_purpose::STANDARD
+                .encode(format!("{}:{}", fx.username, api_token))
+        );
+        let get = |image: &str| {
+            Request::builder()
+                .method(Method::GET)
+                .uri(format!("/{}/{}/manifests/latest", fx.repo_key, image))
+                .header(AUTHORIZATION, basic.clone())
+                .body(Body::empty())
+                .expect("build request")
+        };
+
+        let (blocked_status, blocked_json, blocked_raw) =
+            send(router().with_state(fx.state.clone()), get("blockedimg")).await;
+        let (allowed_status, allowed_json, allowed_raw) =
+            send(router().with_state(fx.state.clone()), get("goodimg")).await;
+
+        let _ = sqlx::query("DELETE FROM curation_rules WHERE staging_repo_id = $1")
+            .bind(fx.repo_id)
+            .execute(&fx.pool)
+            .await;
+        let _ = sqlx::query("DELETE FROM api_tokens WHERE user_id = $1")
+            .bind(fx.user_id)
+            .execute(&fx.pool)
+            .await;
+        fx.teardown().await;
+
+        assert_eq!(
+            blocked_status,
+            StatusCode::FORBIDDEN,
+            "curation-blocked image pull must be 403: {}",
+            String::from_utf8_lossy(&blocked_raw)
+        );
+        assert_oci_envelope(&blocked_json, &blocked_raw, "DENIED");
+        assert!(
+            blocked_json["errors"][0]["message"]
+                .as_str()
+                .unwrap_or_default()
+                .contains("curation_blocked"),
+            "envelope message must keep the curation_blocked marker: {blocked_json}"
+        );
+        // Positive control, same fixture: the non-matching image must NOT be
+        // caught by the gate. Its upstream miss (no reachable upstream in
+        // this fixture) is the spec-shaped MANIFEST_UNKNOWN 404 — which also
+        // pins the 404 arm's envelope.
+        assert_eq!(
+            allowed_status,
+            StatusCode::NOT_FOUND,
+            "non-blocked image must pass the curation gate: {}",
+            String::from_utf8_lossy(&allowed_raw)
+        );
+        assert_oci_envelope(&allowed_json, &allowed_raw, "MANIFEST_UNKNOWN");
     }
 }
--- a/backend/src/api/handlers/oci_v2.rs
+++ b/backend/src/api/handlers/oci_v2.rs
@@ -14,6 +14,7 @@ use std::sync::{
 use std::time::Duration;
 
 use axum::body::{to_bytes, Body, HttpBody};
+use axum::extract::rejection::{BytesRejection, QueryRejection};
 use axum::extract::{DefaultBodyLimit, Query, State};
 use axum::http::header::{AUTHORIZATION, CONTENT_ENCODING, CONTENT_LENGTH, CONTENT_TYPE, LOCATION};
 use axum::http::{HeaderMap, Method, StatusCode};
@@ -58,11 +59,27 @@ struct OciErrorEntry {
 }
 
 fn oci_error(status: StatusCode, code: &str, message: &str) -> Response {
+    oci_error_detail(status, code, message, None)
+}
+
+/// [`oci_error`] with the spec's OPTIONAL `detail` member populated.
+///
+/// distribution-spec `spec.md` "Error Codes": *"The `detail` field is
+/// OPTIONAL and MAY contain arbitrary JSON data providing information the
+/// client can use to resolve the issue."* It is serialized only when
+/// present (`skip_serializing_if`), so a `None` body is byte-identical to
+/// what [`oci_error`] emitted before this helper existed.
+fn oci_error_detail(
+    status: StatusCode,
+    code: &str,
+    message: &str,
+    detail: Option<serde_json::Value>,
+) -> Response {
     let body = OciErrorResponse {
         errors: vec![OciErrorEntry {
             code: code.to_string(),
             message: message.to_string(),
-            detail: None,
+            detail,
         }],
     };
     let json = serde_json::to_string(&body).unwrap_or_default();
@@ -4134,9 +4151,35 @@ async fn handle_refresh_grant(state: &SharedState, form: &TokenForm) -> Response
 async fn token(
     State(state): State<SharedState>,
     headers: HeaderMap,
-    Query(query): Query<TokenQuery>,
-    body: Bytes,
+    query: Result<Query<TokenQuery>, QueryRejection>,
+    body: Result<Bytes, BytesRejection>,
 ) -> Response {
+    // #3110: these two extractors are the only FALLIBLE ones on the /v2
+    // surface, and axum renders its own rejections as `text/plain` — a
+    // repeated `?service=` yields "400 Failed to deserialize query string:
+    // duplicate field `service`" (the same serde_urlencoded mechanism
+    // documented on `TokenQuery::scope` above), and a request body over
+    // TOKEN_REQUEST_BODY_LIMIT_BYTES yields a plaintext 413. Neither is a
+    // spec VIOLATION — distribution-spec "Error Codes" says a 4XX "MAY
+    // return a body in any format" — but they are the last bare-string 4XX
+    // bodies a docker client can meet on this router, so they are re-shaped
+    // into the envelope here. The rejection's own status and text are
+    // preserved; `UNSUPPORTED` (code-14, "The operation is unsupported.") is
+    // the closest registered code, as the spec's closed table has no
+    // malformed-request entry.
+    let Query(query) = match query {
+        Ok(q) => q,
+        Err(rejection) => {
+            return oci_error(rejection.status(), "UNSUPPORTED", &rejection.body_text());
+        }
+    };
+    let body = match body {
+        Ok(b) => b,
+        Err(rejection) => {
+            return oci_error(rejection.status(), "UNSUPPORTED", &rejection.body_text());
+        }
+    };
+
     // Per the OCI Distribution token spec, clients must request a token
     // scoped to the same `service` the server advertises. Reject mismatches
     // with 400 DENIED. Allow a missing `service` so curl-style and older
@@ -7400,7 +7443,7 @@ async fn handle_head_manifest(
     // Curation enforcement (#2930): mirror the GET-manifest gate so a HEAD probe
     // (which many clients issue before the GET) is blocked identically. No-op
     // for hosted repos / curation off.
-    if let Err(resp) = proxy_helpers::enforce_curation_lookup(
+    if let Err(block) = proxy_helpers::evaluate_curation_lookup(
         &state.db,
         repo.id,
         &repo.key,
@@ -7410,9 +7453,10 @@ async fn handle_head_manifest(
     )
     .await
     {
-        // #3110: re-shape the shared gate's REST-style body into the OCI
-        // error envelope this surface is contractually required to emit.
-        return oci_curation_deny_response(resp.status(), &repo.image);
+        // #3110: this surface is contractually required to emit the OCI error
+        // envelope, so it renders the verdict itself rather than forwarding the
+        // shared gate's REST-shaped body — carrying the rule reason through.
+        return oci_curation_deny_response(&block);
     }
 
     // Reference can be a tag or a digest. Resolve locally first: a surviving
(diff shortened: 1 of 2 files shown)


brandonrc:
## Independent validation review — head `301f3d4`

**Verdict: the code change is correct and well-evidenced, but it ships a broken release gate.** One of the two backwards-compatibility claims is false: the DTF `curation-multiformat-enforce` oracle does **not** still match the new body, and it will go red on the docker arm. That needs a companion change in `artifact-keeper-test` before this merges. Everything else I could attack held up.

Method: fresh shallow clone at `/tmp/ak-val-3274`, `pull/3274/head` (`301f3d4`), own Postgres on `:30474` migrated from `backend/migrations/*.sql`, `--lib` runs with **both** `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1`. No commits pushed, no CI triggered.

---

### Claim-by-claim

| # | Claim | Verdict |
|---|---|---|
| 1 | #3110's "400 body not conforming" headline is not reproducible | **CONFIRMED** (structurally) |
| 2 | The 416 side-note is refuted by the spec; no status codes changed | **CONFIRMED** |
| 3 | Two real defects fixed at their construction seams | **CONFIRMED** |
| 4 | Both legacy consumers still match | **REFUTED (one of two)** |
| 5 | The pre-existing bare-string test was rewritten | **CONFIRMED** |
| 6 | Deliberate exclusions | **CONFIRMED, with one mischaracterised** |

---

### 1 — CONFIRMED. Refuting the issue's own premise was the right call.

I did not re-run the conformance suite, but I verified the load-bearing part from the pinned source. `opencontainers/distribution-spec@v1.1.0`, `conformance/01_pull_test.go:250-268`:

```go
Expect(resp.StatusCode()).To(SatisfyAny(
    Equal(http.StatusBadRequest),
    Equal(http.StatusNotFound)))
if resp.StatusCode() == http.StatusBadRequest {
    ... envelope assertions ...
}
```

The envelope assertion is **inside** an `if StatusCode == 400`. A 404 satisfies the test unconditionally. And `handle_get_manifest`'s terminal arm is `oci_error(NOT_FOUND, "MANIFEST_UNKNOWN", …)` (`backend/src/api/handlers/oci_v2.rs:8247`) — an unresolvable reference never reaches a 400. So the named test cannot fail for the reason #3110 asserts. Good catch; a lot of fixes get written for this kind of unverified issue text.

I also corroborated both follow-ups from source rather than taking the run report on trust: `catch_all`'s dispatch (`oci_v2.rs:9723-9800`) has arms for `("POST"|"PATCH"|"PUT"|"DELETE", "uploads")` but **no** `("GET", "uploads")` and no `referrers` arm at all — consistent with the 405 in #3275 and the 5 referrers failures in #3108.

### 2 — CONFIRMED. Declining the status change was also right.

`spec.md` v1.1.0 line 372: *"If a chunk is uploaded out of order, the registry MUST respond with a `416 Requested Range Not Satisfiable` code."* Line 750 pins end-5's failure codes as `404`/`416`. `SIZE_INVALID` is code-10, *"provided length did not match content length"* (line 797) — a different condition. Changing 416→400 would have broken conformance, not fixed it.

Independently verified across both fixed seams that no status moved: `oci_forbidden_scope` still builds `StatusCode::FORBIDDEN`, and both curation seams pass `resp.status()` straight through. `curation_blocked_response` (`proxy_helpers.rs:5399`) only ever emits 403, so `oci_curation_deny_response`'s status parameter has exactly one reachable value.

### 3 — CONFIRMED. Six call sites, both seams, sibling consistent.

`oci_forbidden_scope` has 6 call sites (`oci_v2.rs:4964, 5386, 5635, 5855, 8404, 9465`), all passing static literals — no injection surface. `enforce_curation_lookup` has exactly 2 callers in this file (`7151` HEAD, `8021` GET), both converted.

I diffed `oci_curation_deny_response` (`oci_v2.rs:7638`) against its sibling `oci_scan_deny_response` (`7657` on head): same `oci_error` constructor, same `DENIED` code, same status-preserving shape, same `application/json` from `oci_error`. **No drift.**

### 4 — REFUTED for the DTF oracle. This is the blocking finding.

**`test-ghsa-vvc3-scope-checks.sh` — holds.** `assert_403_denied` matches on the raw body with `grep -Eq` (upstream `tests/security/test-ghsa-vvc3-scope-checks.sh:317`), and `Token does not have required scope` survives verbatim inside the JSON message. Verified by running that exact `grep -E` against the real body: **pass**.

**`curation-multiformat-enforce` — BREAKS.** The oracle does not grep. It extracts a JSON field (upstream `deploy-test/harness/tiers/curation-multiformat-enforce/oracle.sh:142-143`):

```sh
err="$(echo "$body" | jq -r '.error // empty' 2>/dev/null || true)"
if [ "$st" = "403" ] && [ "$err" = "curation_blocked" ]; then
```

I captured the actual wire body by instrumenting the PR's own DB test and printing `blocked_raw`:

```
{"errors":[{"code":"DENIED","message":"curation_blocked: pull of blockedimg is blocked by this repository's curation policy"}]}
```

Piped through the oracle's own expression:

```
ORACLE: FAIL  (err='')
```

There is no top-level `.error` key in the new body, so `jq -r '.error // empty'` yields the empty string and the test *"docker: BLOCKED image GET manifest -> 403 curation_blocked (#2930)"* fails. Keeping the `curation_blocked` marker in the message only helps a substring consumer; this consumer is a field lookup. The tier's `manifest` documents the expected shape as `403 {"error":"curation_blocked"}` too, so both files need updating.

The HEAD arm (`oracle.sh:150`) and the non-blocked control (`:162`) only check status and are unaffected — it is exactly one testcase, but the tier asserts zero failures and the DTF gate is blocking.

Scope of the breakage — I checked every other consumer I could find and this is the only one:
- All other DTF oracles that touch `/v2` were scanned for `jq -r '.error'`; `curation-multiformat-enforce` is the sole hit.
- `token-colon-scope/oracle.sh:459` and `5f2q-upload-scope` match on `grep -q` / status only, and neither exercises `/v2`.
- No consumer of `curation_blocked` exists inside `artifact-keeper` itself.

**Fix:** companion MR to `artifact-keeper-test` teaching the docker arm the OCI shape while leaving the npm arm (unchanged REST shape) alone:

```sh
code="$(echo "$body" | jq -r '.errors[0].code // empty' 2>/dev/null || true)"
msg="$(echo "$body"  | jq -r '.errors[0].message // empty' 2>/dev/null || true)"
if [ "$st" = "403" ] && [ "$code" = "DENIED" ] && \
   printf '%s' "$msg" | grep -q 'curation_blocked'; then
```

plus the corresponding line in that tier's `manifest`. That MR must land with (or before) this PR.

### 5 — CONFIRMED, and the tests are real.

All four tests pass on head, and I reproduced the revert-proof independently rather than trusting the reported one. Faithful production revert only (bare-string `oci_forbidden_scope` restored to the merge-base shape; both seams back to `return resp;`; the now-unused helper kept so the pure test compiles). Stripping every `#[cfg(test)]` block from the merge-base and reverted files leaves only doc comments and that one unused helper differing — no executable production divergence.

Reverted tree:

```
FAIL [ 0.014s] tests::test_oci_forbidden_scope_status_and_body
PASS [ 0.018s] tests::test_oci_curation_deny_response_is_oci_error_envelope
FAIL [ 3.868s] oci_error_envelope_db_tests::curation_blocked_manifest_get_returns_oci_error_envelope
FAIL [ 4.843s] oci_error_envelope_db_tests::scope_denied_manifest_put_returns_oci_error_envelope
  got: {"error":"curation_blocked","package":"blockedimg","reason":"blocked by test rule"}
  got: Token does not have required scope: write:artifacts
```

Head: 4/4 pass, `3.972s` / `4.989s` on the DB pair. **3.9s and 4.8s, not 0.04s** — these executed against a real database in both directions. Assertions go through `serde_json::from_slice` and index `errors[0].code`, never a substring; both positive controls (201 CREATED on the wildcard token, 404 `MANIFEST_UNKNOWN` on the non-ruled image) are in the same fixture and would catch a gate that denies everything. The one pass on the reverted tree is the pure helper test, which the PR body already flags as non-behavioural — accurate, not oversold.

### 6 — CONFIRMED, but one exclusion is described wrongly.

`PAGINATION_NUMBER_INVALID` (`oci_v2.rs:2318`, a **400** on `GET /v2/<name>/tags/list?n=…`) is called *"conforming shape, unregistered code — cosmetic."* `spec.md:784` reads *"The `code` field MUST be one of the following:"* over a closed 14-row table (`code-1`…`code-14`), and this code is not in it — so by the same MUST this PR cites twice, it is a violation, not cosmetic. The defensible ground is the reference implementation (distribution defines `ErrorCodePaginationNumberInvalid` with this exact string), which is a fine reason to leave it — it just needs to be the reason given. Worth a one-line correction to the PR body rather than a code change.

---

### Other findings

**M1 — the fix silently discards the curation rule `reason` (observability regression).**
The reverted-tree output above shows what the wire used to carry: `{"error":"curation_blocked","package":"blockedimg","reason":"blocked by test rule"}`. The new message keeps the marker and the image but drops `reason` entirely, and `oci_error` hardcodes `detail: None` (`oci_v2.rs:65`). An operator debugging a blocked `docker pull` now gets "blocked by this repository's curation policy" where every other proxy format still reports which rule fired. The spec provides the field for exactly this (`spec.md:782`: *"The `detail` field is OPTIONAL and MAY contain arbitrary JSON data providing information the client can use to resolve the issue"*).

The reason is unreachable at the OCI seam only because `enforce_curation_lookup` returns an opaque `Response`. Suggested fix: have `enforce_curation` surface the structured `(package, reason)` alongside the response — or add an `enforce_curation_lookup_detailed` — and populate `detail: {"package": …, "reason": …}` on the OCI side. That restores parity, is spec-legal, and gives the DTF oracle a stable structured field to assert on.

**L1 — "Extractors in the `/v2` router are `Infallible`" is false.**
`/v2/token` takes `Query<TokenQuery>` and `Bytes` (`oci_v2.rs:3918-3919`), both fallible. `GET /v2/token?service=a&service=b` produces an axum `QueryRejection`: a **400 `text/plain`** *"Failed to deserialize query string: duplicate field `service`"*. The module's own comment at `oci_v2.rs:3725-3733` documents this exact mechanism ("serde_urlencoded errors on the repeated known field with '400 duplicate field `scope`'") — it is why `scope` was removed from the struct, but `service` still has it. A `Bytes` rejection past `TOKEN_REQUEST_BODY_LIMIT_BYTES` is likewise a plaintext 413. Neither is a spec violation (`spec.md:764`: a 4XX *"MAY return a body in any format"*), so no code change is required — but it is a bare-string 4XX on `/v2`, which is the ergonomics argument this PR makes, and the enumeration claim should be softened. `catch_all` and `handle_catalog` use `Query<HashMap<String,String>>`, which is effectively infallible — that part of the claim holds.

**L2 — middleware was not fully enumerated; two layers still emit REST-shaped JSON on `/v2`.**
`routes.rs:146` nests `/v2` *inside* the global layer stack, and `setup_guard` (`backend/src/api/middleware/setup.rs:62-79`) has no `/v2` entry in its allowlist. While setup is pending, every `/v2` request — including the `docker login` challenge probe — gets `403 {"error":"SETUP_REQUIRED","message":…,"instructions":[…]}`, a JSON 4XX body that violates the same MUST. `demo_guard` (`middleware/demo.rs:71-77`) does the same with `{"error":"Write operations are disabled…"}` on every `/v2` write verb in demo mode. Both are narrow states, so a follow-up issue rather than scope creep here.

Credit where due: `guest_access_guard` **is** correctly allowlisted for `/v2` (`middleware/guest_access.rs`, `allowlist_oci_v2_challenge`), so the third global layer is clean — the enumeration was not wrong everywhere.

**L3 — flake note.** `curation_blocked_manifest_get_returns_oci_error_envelope`'s positive control pulls `goodimg` from the fixture's `https://upstream.example.test`, i.e. a live outbound attempt whose failure produces the asserted 404. It resolved fast here (3.9s total) but it is a real network attempt on a fixture that is otherwise offline-safe; if `.test` resolution ever hangs on a runner this test pays the proxy timeout.

---

### What I could not break

- No status code changed anywhere in the diff — I read all 377 lines and re-derived it from the reverted-tree comparison.
- `oci_curation_deny_response` does not drift from `oci_scan_deny_response`.
- The `/v2` handler surface really is otherwise complete: stripping the test modules leaves 9,976 production lines, and every non-2xx `StatusCode::` in them resolves to `oci_error`/`oci_internal_error` except the 401 challenge (which carries `WWW-Authenticate` correctly) and the quarantine tuple at `oci_v2.rs:7356-7362`, which feeds `oci_error`. There are zero `.into_response()` calls in the file's production code. Every `if let Err(resp) = <gate>` helper (`require_oci_repo_write_access`, `enforce_token_repo_scope`, `enforce_scan_pull_scope`, the three upload-cleanup-key helpers, `reset_oci_upload_session_state`) returns an envelope. The `UNKNOWN` code appears 6 times but only on 502s, outside the 4XX MUST. `enforce_curation_lookup` was genuinely the only shared gate leaking a foreign shape.

---

### To clear this

1. **Blocking:** companion MR to `artifact-keeper-test` for `curation-multiformat-enforce` (`oracle.sh:142-143` + `manifest`), landed with or before this PR.
2. **Should fix:** carry the curation `reason` into `detail` (M1) — it also gives that oracle something structured to key on.
3. **Body corrections:** drop the `Infallible` extractor claim (L1); restate the `PAGINATION_NUMBER_INVALID` exclusion as a reference-implementation extension rather than cosmetic.
4. **Follow-up issue:** `setup_guard` / `demo_guard` REST bodies on `/v2` (L2).

Cleaned up: container `ak-val3274-pg` removed, clone and `target/` deleted.

brandonrc:
## Remediation for the independent validation review

Thanks — the blocking finding was real and the two body corrections were fair. Everything is addressed; evidence per finding below. New head: `552edef` (rebased onto `8b9e0ff`). The production fix you confirmed (claims 1, 2, 3, 5) is untouched; nothing was re-litigated.

Method: fresh shallow clone at `/tmp/ak-fix-3274`, own Postgres `ak-fix3274b-pg` on `:30674` migrated from `backend/migrations/*.sql`, every `--lib` run with **both** `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1`. Separate clone for `artifact-keeper-test`.

The branch also had to be **rebased** off `1115203` onto `8b9e0ff` mid-way (#3268 and #3272 landed in both of these files); see [the last section](#rebase-onto-8b9e0ff) — everything below was re-run after the rebase, on the rebased tree.

---

### B1 (blocking) — the DTF oracle. Fixed in artifact-keeper/artifact-keeper-test#337, validated locally before opening.

Your reading is exactly right and my claim was wrong: `oracle.sh:142-143` is `jq -r '.error // empty'`, a field lookup, and the marker in the message does nothing for it.

**The oracle now selects the expected shape from the deployed backend version** rather than picking one:

```sh
CURATION_OCI_ENVELOPE_MIN_VERSION="1.7.2"      # the milestone #3110 ships in

curation_block_oci_ok() {                       # structural, not a grep on the body
  code="$(  printf '%s' "$body" | jq -r '.errors[0].code // empty')"
  msg="$(   printf '%s' "$body" | jq -r '.errors[0].message // empty')"
  reason="$(printf '%s' "$body" | jq -r '.errors[0].detail.reason // empty')"
  [ "$code" = "DENIED" ] || return 1
  printf '%s' "$msg" | grep -q 'curation_blocked' || return 1
  [ -n "$reason" ] || return 1
}
```

At or above the floor the envelope is **required** — a regression back to the REST body fails the tier. Below it either shape passes, each still checked structurally. That second branch is not laxness, it is forced by how the gate is wired: `release.yml` calls `artifact-keeper-test/.github/workflows/release-gate.yml@main` on any `v*` tag, so this same oracle runs against 1.7.x patch cuts that do not contain this PR, and against `dev` images built from main before the version bump (main's workspace version is still `1.7.1`). "Envelope only" false-reds every one of those; "accept either, always" lets the regression ride through on 1.8.0+.

**Red → green against captured real bodies.** The harness `eval`s the helper bodies straight out of the shipped `oracle.sh` and the version helpers straight out of `tests/lib/common.sh`, so what runs is the bytes that land. None of the bodies is hand-written:

| body | captured from |
|---|---|
| REST | `proxy_helpers::CurationBlock::into_rest_response` — the renderer every non-OCI format serves |
| OCI (DB) | printed out of this PR's own DB test — what the **real router** returned |
| OCI (tier) | `oci_curation_deny_response` with the tier's identity (`library/hello-world`, `#2930 tier block`) |
| OCI (long) | same, with a realistic long operator reason — 594 bytes |

```
floor: CURATION_OCI_ENVELOPE_MIN_VERSION=1.7.2

== RED: the OLD oracle expression against the REAL new wire body ==
  OK   old jq '.error' vs new OCI envelope (DB-test capture)            -> fail
  OK   old jq '.error' vs new OCI envelope (tier identity)              -> fail
  OK   old jq '.error' vs legacy REST body                              -> pass

== GREEN: the NEW oracle, backend 1.7.2 (at the floor: envelope REQUIRED) ==
  OK   new oracle vs OCI envelope (DB-test capture)                     -> pass (shape=oci-envelope)
  OK   new oracle vs OCI envelope (tier identity)                       -> pass (shape=oci-envelope)
  OK   new oracle vs legacy REST body (REGRESSION)                      -> fail

== GREEN: the NEW oracle, backend 1.8.0 (above the floor) ==
  OK   new oracle vs OCI envelope (tier identity)                       -> pass (shape=oci-envelope)
  OK   new oracle vs legacy REST body (REGRESSION)                      -> fail

== GREEN: the NEW oracle, backend 1.7.1 (below the floor: either shape) ==
  OK   new oracle vs legacy REST body (shipped 1.7.1)                   -> pass (shape=rest)
  OK   new oracle vs OCI envelope (dev image off main)                  -> pass (shape=oci-envelope)

== Negative controls (must FAIL at every version) ==       [1.7.1 / 1.7.2 / 1.8.0]
  OK   403 with no curation verdict at all                              -> fail
  OK   OCI envelope, wrong code (UNAUTHORIZED)                          -> fail
  OK   OCI envelope, no curation_blocked marker                         -> fail
  OK   OCI envelope, reason dropped from detail                         -> fail
  OK   not JSON at all                                                  -> fail

== Truncation control: why head -c 300 had to grow ==
  OK   head -c 300 of a long-reason envelope (would be a FALSE red)     -> fail
  OK   head -c 2000 of the same body                                    -> pass

ALL EXPECTATIONS MET
```

Two things worth calling out:

- **The "reason dropped from detail" negative control is the body this PR emitted at `301f3d4`.** Fixing B2 gave the oracle something structural to key on, so it now rejects the pre-review shape too.
- **`head -c 300` had to grow to 2000.** The envelope repeats the image and the operator-authored `reason` in both `message` and `detail`; the tier's own values land at 225 bytes, 75 from truncation. A truncated body makes every `jq` lookup return empty — a false red that reads exactly like a regression. That one nearly shipped as a latent flake.

`test-ghsa-vvc3-scope-checks.sh` untouched, as you said — re-verified that `:317` is `grep -Eq` on the raw body and the legacy text survives verbatim. `curation_blocked` has no other consumer in either repo (re-checked both).

Manifest updated too, since it documented the old shape.

### B2 — the discarded `reason`. Fixed, with a parsed assertion.

You identified the cause precisely: the reason was unreachable because the seam returned an opaque `Response`. Split it there rather than re-parsing:

- `evaluate_curation{,_lookup}` → `Result<(), CurationBlock { package, reason }>`
- `enforce_curation{,_lookup}` → thin wrappers rendering via `CurationBlock::into_rest_response`

so every non-OCI format is byte-identical, and the OCI seam builds its own envelope from the structured verdict. Real wire body now:

```json
{"errors":[{"code":"DENIED",
            "message":"curation_blocked: pull of blockedimg is blocked by this repository's curation policy: blocked by test rule",
            "detail":{"package":"blockedimg","reason":"blocked by test rule"}}]}
```

The reason is in **both** places on purpose: `message` is what `docker pull` prints to a human, `detail` is what a machine reads — and it is what `spec.md` provides the member for, as you quoted.

Guarding it needed two things beyond the obvious assertion:

- `test_oci_error_omits_detail_when_absent` — positive control that `oci_error`'s output is byte-for-byte unchanged for every *other* error on the surface (`detail` is `skip_serializing_if`). Without it, "added a field to the error type" is an unbounded change.
- `proxy_helpers::tests::curation_block_rest_body_is_unchanged` — pins the shared REST body the seam split could silently have moved. That is the body the DTF tier's **npm** arm reads, so moving it would have broken the other half of the same tier while I was fixing the docker half.

### B3 — the `Infallible` claim. Confirmed false against the running router; fixed rather than caveated.

Not taken on reading — driven through `router()`:

```
GET /v2/token?service=a&service=b   -> 400  Failed to deserialize query string: duplicate field `service`   (text/plain)
GET /v2/token?service=artifact-keeper -> 200 {"token":"anonymous",...}
```

Your pointer to the module's own comment on `TokenQuery::scope` was the giveaway — same mechanism, and `service` still has the field.

The sentence is gone from the description. Since it is a 5-line change with no effect on the success path, I re-shaped the rejections instead of only documenting them: `Result<Query<TokenQuery>, QueryRejection>` / `Result<Bytes, BytesRejection>`, mapped through `oci_error` preserving `rejection.status()` and `rejection.body_text()`. That covers the 413 you noted as well. Agreed it was never a spec violation (4XX MAY be any format) — the argument for fixing is the ergonomics one this PR already makes, so the enumeration is now true rather than merely honest.

`UNSUPPORTED` (code-14) is the closest registered code; the closed table has no malformed-request entry, and that is stated in the code comment rather than glossed.

New guard `token_query_rejection_returns_oci_error_envelope` runs through the real router with the well-formed request as a positive control in the same fixture, so it cannot be satisfied by a route that rejects everything.

### B4 — middleware residuals. Recorded as #3284.

Confirmed both, and confirmed the credit you gave `guest_access_guard` (`is_allowlisted` covers `/v2`, `/v2/`, and `/v2/<name>/manifests/<ref>`, pinned by `allowlist_oci_v2_challenge`). `setup_guard`'s allowlist is `/api/v1/auth*` plus the password route and a small `matches!` set — no `/v2`. Filed **#3284** with both, the `routes.rs:146` nesting that makes them reachable, and `guest_access_guard` named as the pattern to copy. Not pulled into this diff: they are middleware, in narrow states, and the diff is already 588 lines.

### B5 — `PAGINATION_NUMBER_INVALID`. "Cosmetic" retracted; reference implementation cited instead.

You are right that `spec.md`'s "The `code` field MUST be one of the following:" over a closed table makes it a violation by the same MUST I cite twice. Rather than take the reference-implementation claim on trust I fetched it:

```go
// distribution/distribution, registry/api/errcode/register.go:219-226 (main)
// identical in registry/api/v2/errors.go:137-144 on release/2.8
ErrorCodePaginationNumberInvalid = register(errGroup, ErrorDescriptor{
    Value:          "PAGINATION_NUMBER_INVALID",
    Message:        "invalid number of results requested",
    HTTPStatusCode: http.StatusBadRequest,
})
```

Same string, same status. AK's body is already the conforming envelope; only the code is an upstream-compatible extension, and clients are written against that implementation. Left as-is, on that ground, stated that way in the description.

### L3 — the flake note

Kept as-is. The `goodimg` control's upstream attempt is load-bearing: it is what makes the 404 arm's envelope assertion meaningful, and replacing the outbound attempt with a mock would weaken the positive control into something the fixed code alone produces. Noted for whoever sees it time out on a runner.

---

### Revert-proof of the new guards

Three independent reverts in one tree — curation body back to `301f3d4` (reason discarded, `detail: None`), `/v2/token` back to `Query(query): Query<TokenQuery>, body: Bytes`, and the shared REST body's `"error"` key moved. Exactly the four guards covering them fail; the three covering the untouched scope gate keep passing:

```
FAIL [ 0.013s] proxy_helpers::tests::curation_block_rest_body_is_unchanged
  left:  "{\"curation_error\":\"curation_blocked\",\"package\":\"left-pad\",…}"
  right: "{\"error\":\"curation_blocked\",\"package\":\"left-pad\",…}"

FAIL [ 0.018s] oci_v2::tests::test_oci_curation_deny_response_is_oci_error_envelope
  message must name the rule that fired (docker prints message):
  {"errors":[{"code":"DENIED","message":"curation_blocked: pull of someimage is blocked by this repository's curation policy"}]}

FAIL [ 0.280s] oci_error_envelope_db_tests::token_query_rejection_returns_oci_error_envelope
  body must be the OCI error envelope with code UNSUPPORTED;
  got: Failed to deserialize query string: duplicate field `service`

FAIL [ 3.883s] oci_error_envelope_db_tests::curation_blocked_manifest_get_returns_oci_error_envelope
  detail.reason must name the curation rule that fired:
  {"errors":[{"code":"DENIED","message":"curation_blocked: pull of blockedimg is blocked by this repository's curation policy"}]}
  left: None   right: Some("blocked by test rule")

PASS [ 4.842s] scope_denied_manifest_put_returns_oci_error_envelope
PASS [ 0.030s] test_oci_forbidden_scope_status_and_body
PASS [ 0.030s] test_oci_error_omits_detail_when_absent
```

3.9s / 4.8s / 0.28s, not 0.04s — the DB guards ran against a real database in both directions. Fixed tree 7/7 pass. Full `oci_v2` + `proxy_helpers` battery **795/795** (788 before the rebase; #3268 added a module), and the whole backend `--lib` suite **15104/15104 passed, 0 failed** on the rebased tree — worth the 3.5 minutes because the seam split touches a helper with 9 callers across the format handlers; `cargo fmt --all --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean; `bash -n` and `shellcheck -S warning` clean on the oracle.

### Merge order (also stated in the description)

There is no PR-level DTF gate — the gate runs only on a `v*` tag push and by rehearsal dispatch, always at `artifact-keeper-test@main`. So:

1. **artifact-keeper/artifact-keeper-test#337 first** — it is green against every backend that exists today, so it is a no-op until a backend carrying this PR is deployed.
2. **Then this PR.**
3. Both before the next `v*` tag. The reverse order leaves a window where a release cut runs the old `.error` lookup against the new envelope and fails a blocking gate.

If this PR slips out of milestone 1.7.2, `CURATION_OCI_ENVELOPE_MIN_VERSION` in #337 moves with it — flagged in both descriptions.

---

### Rebase onto `8b9e0ff`

The PR went dirty while this was in flight: #3268 (`f75f010`, the `/v2` read-authz gate) and #3272 (`118a8b0`, `resolve_virtual_metadata`'s transform signature) landed in exactly these two files. Rebased off `1115203` onto `8b9e0ff`; head is now `552edef` and the PR is MERGEABLE again.

**Two conflicts, both in `oci_v2.rs`, both the same shape:** #3268 and this PR each appended a new test module at EOF, so git could not order `oci_read_authz_tests` against `oci_error_envelope_db_tests`. Resolved by keeping both, each with its own header comment and attributes. `proxy_helpers.rs` merged without conflict — #3272's change is ~2,800 lines from the curation seam.

I did not take the clean auto-merge on trust:

- **Every deleted line** in `git diff 8b9e0ff` across both files was audited individually. All 21 are this PR's own — the bare-string `oci_forbidden_scope` body, the two `return resp;` arms, the old `Query`/`Bytes` params, the old `detail: None`, the two `enforce_curation_lookup` call heads, the pre-rewrite test asserts, and two seam-internal lines. Nothing from #3268 or #3272 was dropped or rewritten. That check is what would have caught a wrongly-merged hunk; reading the result for plausibility would not.
- **Gate ordering preserved**: `enforce_scan_pull_scope` → `enforce_token_repo_scope` → `require_oci_repo_read_access` → curation seam, in both `handle_get_manifest` and `handle_head_manifest`. Curation stays last, where it was.
- **The negative-cache constraint is byte-identical.** `authorized_virtual_members` still fetches members before probing, and still consults/populates only when `members.len() == total` — `VirtualResolveKey` has no principal term, so a narrowed walk touching it would turn the member filter into a cache-poisoning primitive. This PR never enters that region, which the deletion audit proves rather than asserts.
- **The `/v2` enumeration was re-run against the new code**, since the whole premise of this PR is that the enumeration is complete. #3268's three error returns are `oci_denied_repo_access()` (403 DENIED), `oci_error(404, "NAME_UNKNOWN", …)` and `oci_authorization_unavailable()` (503 DENIED) — all already the envelope, so nothing new to fix and the "conforming already" list still holds. A denied virtual member is dropped from the walk and surfaces as the existing `MANIFEST_UNKNOWN` 404 envelope, not as a new shape.

fmt, clippy, the 795-test battery and the full three-way revert-proof were all re-run after the rebase. #337 is unaffected — it touches only `artifact-keeper-test`, and the wire body it asserts did not change.

Issue #1808, opened 2026-06-09
Security (low): native download paths leak private-repo existence via 401-vs-404 oracle


brandonrc:
**Severity:** low — information disclosure (anonymous). Partially related to the rule-less-private-repo fix; filed separately for the anonymous existence oracle.

## Summary
Anonymous private-repository name enumeration oracle (401 vs 404) on native download paths

## Reproduction
```
Server http://localhost:8080. No credentials used (anonymous). Status code alone is the oracle; timing not needed.

# Existing PRIVATE repo (is_public=f) -> 401 with WWW-Authenticate challenges
$ curl -s -o /dev/null -w "%{http_code}\n" http://localhost:8080/maven/acme-internal-core/a/b/1.0/b-1.0.jar
401
$ curl -s -i http://localhost:8080/maven/acme-internal-core/x/y/1.0/y-1.0.jar | head -4
HTTP/1.1 401 Unauthorized
www-authenticate: Basic realm="artifact-keeper"
www-authenticate: Bearer realm="artifact-keeper", charset="UTF-8"
www-authenticate: Cargo

# Guaranteed-NONEXISTENT repo -> 404 "Repository not found"
$ curl -s -o /dev/null -w "%{http_code}\n" http://localhost:8080/maven/zzz-nonexistent-repo-9/a/b/1.0/b-1.0.jar
404

Enumeration sweep (anonymous), status perfectly tracks DB existence:
  acme-internal-core    anon=401  db_exists=1
  acme-web-private      anon=401  db_exists=1
  globex-secret-sauce   anon=401  db_exists=1
  acme-does-not-exist   anon=404  db_exists=no
  payments-prod-private anon=404  db_exists=no
  random-xyz-123        anon=404  db_exists=no

Reproduced on all four named formats (existing-private / nonexistent):
  MAVEN /maven/acme-internal-core/...           -> 401 / 404
  PYPI  /pypi/acme-ml-models/simple/p/          -> 401 / 404
  NPM   /npm/acme-web-private/p                  -> 401 / 404
  CARGO /cargo/acme-crates-local/config.json     -> 401 / 404
```

## Control
Control 1 (must differ, and does): a nonexistent repo key returns 404 while an existing private repo returns 401 to the SAME anonymous caller — the differing status is the leak.
[...]

PR #3278, opened 2026-08-10
fix(ansible): serve Galaxy discovery at /api without trailing slash and accept the Token auth scheme

### LINKED ISSUE #3137: Ansible galaxy discovery
### Description

Hi, thanks for your work on 
https://github.com/artifact-keeper/artifact-keeper/issues/1451
https://github.com/artifact-keeper/artifact-keeper/pull/1506

it seems the galaxy client adds /api to its discovery of the configured url, however the artifactkeeper only handles /api/ (trailing slash).

### Environment Information

- Deployment type: docker-compose
- Operating System: arch linux
- Architecture: amd64
- Database: docker-compose default
- Reverse Proxy: docker-compose default
- Network setup: docker-compose default

### Backend Version

latest

### Web UI Version

latest

### Package Format

Other / WASM Plugin

### Expected Behavior

status 200 for every query

### Observed Behavior

```
❯ ANSIBLE_DEBUG=1 ansible-galaxy collection publish \
  ~/Downloads/community-hashi_vault-7.1.0.tar.gz \
  --server artifact_keeper -vvvv
...
Calling Galaxy at http://localhost:30080/ansible/galaxy/
Calling Galaxy at http://localhost:30080/ansible/galaxy/api
Initial connection to galaxy_server: http://localhost:30080/ansible/galaxy/
Calling Galaxy at http://localhost:30080/ansible/galaxy/
Calling Galaxy at http://localhost:30080/ansible/galaxy/api
[ERROR]: Error when finding available api versions from artifact_keeper (http://localhost:30080/ansible/galaxy/) (HTTP Code: 401, Message: Unauthorized)
```

```
❯ curl -s -o /dev/null -w "%{http_code}\n" \
  -H "Authorization: Bearer $BEARER_TOKEN" \
  http://localhost:30080/ansible/galaxy/api
404
❯ curl -s -o /dev/null -w "%{http_code}\n" \
  -H "Authorization: Bearer $BEARER_TOKEN" \
  http://localhost:30080/ansible/galaxy/api/
200
```

Additionally, note how there is 401, 404 and 200 in the picture. Galaxy sends the header in the format of "Token <token>" while AK expects "Bearer <token> (took me a while to figure out, but could confirm this with mitmproxy).

### Steps to Reproduce

same as referenced issues

### Logs

```shell

```

### Additional Context

Image

### Pre-submission Checklist

brandonrc:
Fixes #3137

## Scope — read this first

This PR fixes the two defects #3137 reports and diagnoses: the discovery route only
answered `/api/` (the client requests `/api`), and `Authorization: Token <t>` was
rejected as a malformed header. Both are closed, and `ansible-galaxy collection
install` now works end to end.

**`ansible-galaxy collection publish` still fails**, at the step immediately after the
two fixed here. The upload POST succeeds and the collection is stored, but AK's `202`
omits the `task` field that `GalaxyAPI.publish_collection` unconditionally dereferences,
so the client dies with `Unexpected Exception, this is probably a bug: 'task'`.
`--no-wait` does not help — the dereference happens inside `api.publish_collection`,
before the caller ever consults `wait` — and adding a `task` key without also serving
`v3/imports/collections/{id}/` converts the crash into an indefinite hang, because
`wait_import_task` retries a 404 forever and `--import-timeout` defaults to `0`.

Fixing that needs a new route plus an import-status contract, which warrants its own
review, so it is deliberately **not** in this diff. It is filed as **#3282** with the
full mechanism and a proposed shape. #3283 tracks the missing Ansible integration/DTF
coverage that let all three breaks reach a user report.

`Fixes #3137` is kept because the issue — by title, by body, and by the failure in its
own transcript (`Error when finding available api versions ... HTTP Code: 401`) — is
about discovery and the auth scheme, and both are resolved. If you would rather hold
#3137 open until `publish` works end to end, say so and I will retarget it to #3282.

### 1. Discovery route only registered with a trailing slash

The CLI discovers the API root at `<server>/api` **without** a trailing slash:
ansible-core's `g_connect` builds the URL as `_urljoin(n_url, '/api/')`, and `_urljoin`
strips `/` from every component (`lib/ansible/galaxy/api.py`, stable-2.18):

```python
def _urljoin(*args):
    return '/'.join(to_native(a, errors='surrogate_or_strict').strip('/') for a in args + ('',) if a)
```

(the appended `''` is falsy and filtered by `if a`, so no trailing slash survives). The
router registered only `/:repo_key/api/`, and axum matches `/api` and `/api/` as
distinct paths, so version negotiation 404'd — exactly the reporter's `curl` matrix
(`/api` → 404, `/api/` → 200).

**Fix:** register both spellings on `api_root`.

Two upstream details, both raised in review, that make the one-line alias sufficient
and that the earlier draft of this body got half-right:

- `g_connect`'s `except` block opens with
  `if n_url.endswith('/api') or n_url.endswith('/api/'): raise`. A user who configures
  the server URL **as** `http://host/ansible/<repo>/api` therefore got no fallback probe
  at all — the first request 404'd and the CLI aborted. The alias fixes that
  configuration too, not just the bare-server-URL one.
- `g_connect` ends with `self.api_server = n_url`. **That reassignment**, not the
  `+ '/'` in individual call sites, is why only the discovery route needs the alias:
  every later v3 URL is built from whichever discovery URL succeeded.
  (`publish_collection` does append `+ '/'`, but `get_collection_versions` instead
  passes `'/?%s=%d'` as its final `_urljoin` component — the original claim was true of
  one call site, not of all of them. Net effect is the same.)

Registering both spellings matches existing precedent in this repo —
`backend/src/api/handlers/pypi.rs:55-59` registers `/simple` and `/simple/` the same way.

### 2. `Authorization: Token <api_key>` rejected as malformed

The CLI authenticates Galaxy API calls with the `Token` scheme — ansible-core
`lib/ansible/galaxy/token.py`:

```python
class GalaxyToken(object):
    token_type = 'Token'
    ...
    headers['Authorization'] = '%s %s' % (self.token_type, self.get())
```

(only the Keycloak/Automation-Hub variant uses `Bearer`). `extract_token_from_auth_header`
recognized no `Token` scheme; the value contains a space so it fell through to
`ExtractedToken::Invalid`, and `repo_visibility_middleware` → `try_resolve_auth_outcome`
answered 401 on every authenticated call — the reporter's `HTTP Code: 401` during
discovery.

That is also why the reporter's transcript shows 401 *before* 404:
`_add_auth_token(headers, url, required=auth_required)` attaches the header whenever a
token is configured, **including on the `auth_required=False` discovery probe**, so a
configured token 401'd discovery itself rather than only the later authenticated calls.

**Fix:** treat `Token <t>` as Bearer-equivalent so the credential flows through the same
JWT → API-token validation chain.

**No weakening, by construction.** The new branch returns the *same*
`ExtractedToken::Bearer` variant, not a parallel one, so no downstream consumer can
diverge — `auth_middleware`, `try_resolve_auth_outcome`, `optional_auth_middleware`,
`admin_middleware` and `extract_visibility_token` all reach identical validation.
Expiry, revocation, deactivation, the `prefix_secret` lookup, scope-gated admin
demotion (GHSA-vvc3), `allowed_repo_ids` and the permission-service checks apply
unchanged. Every credential now accepted under `Token` was already accepted under
`Bearer`: the change adds a *presentation*, not a capability. `Token <bad>` → 401,
`Token ` (empty) → 401. Nothing relied on `Token <x>` being rejected — the one test
enumerating unrecognised schemes uses `"Bogus scheme-with args"` and `"!!! bad !!!"`,
neither of which the new branch matches.

Scheme matching stays case-sensitive, consistent with the pre-existing `Bearer` arm.
RFC 9110 §11.1 makes auth schemes case-insensitive, so `token <x>` is strictly
non-conformant to reject — but that inconsistency predates this PR (`Basic`/`basic` are
both accepted, `Bearer` is not), ansible-core emits exactly `Token`, and being stricter
fails closed. Normalizing the whole extractor belongs in its own change.

## Tests

- `test_3137_api_discovery_without_trailing_slash` (DB-backed router test):
  `GET /{repo}/api` serves the discovery document and advertises `v3`; positive control
  in the same fixture (`/api/` still 200); unknown repo still 404s at the handler.
- `test_3137_galaxy_api_through_visibility_middleware` (DB-backed) — **both fixes
  together, through the middleware that produced the reported 401**:
  `GET /ansible/{repo}/api` (no trailing slash) with
  `Authorization: Token <real_api_token>` against a **private** repository → 200 with
  `v3` advertised. Fail-closed controls in the same fixture so the 200 cannot be
  explained by a permissive middleware: anonymous → 401, bogus `Token` credential →
  401, and unknown repo key → **401, not 404** (the #1808 existence-oracle closure; the
  handler-level 404 asserted in the router test above is not the deployed status, and
  that comment is corrected).
- `test_3137_galaxy_token_scheme_authenticates_api_token` (DB-backed): drives the exact
  header bytes the CLI sends through `extract_token` → `try_resolve_auth_outcome`;
  resolves to the token owner. Both fail-closed assertions now run **before** the
  positive `panic!`, so a revert still evaluates them, and the discriminating one is an
  **empty** credential (`Token `), which must be `InvalidCredential` — not
  `NoCredential`, which `repo_visibility_middleware` would let through as an anonymous
  read on a public repository.
- `test_3137_extract_galaxy_token_scheme_as_bearer` (pure): `"Token abc"` →
  `Bearer("abc")`.

## Revert-proof (`DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1`, private Postgres)

**Faithfulness.** With both production hunks reverted to merge-base shape (`1115203`)
and the tests kept, stripping every `#[cfg(test)] mod tests` block leaves the two files
differing from the merge base **only in `//` / `//!` comment lines** — zero executable
production statements differ. (Stronger than the shortstat heuristic, which does not
apply here because this round rewrites test code.)

**Reverted run — 4/4 FAIL:**

| test | failure | time |
|---|---|---|
| `test_3137_extract_galaxy_token_scheme_as_bearer` | `matches!(result, ExtractedToken::Bearer(..))` | 0.012s |
| `test_3137_api_discovery_without_trailing_slash` | left `404`, right `200` | 0.216s |
| `test_3137_galaxy_api_through_visibility_middleware` | left `401`, right `200` | 1.090s |
| `test_3137_galaxy_token_scheme_authenticates_api_token` | `expected Resolved for `Token <valid_api_key>`, got InvalidCredential` | 1.141s |

**Per-fix reverts**, to prove the new middleware test gates *each* fix and not just
their union:

- route alias removed only → middleware test fails `left: 404, right: 200` (2.906s)
- `Token` branch removed only → middleware test fails `left: 401, right: 200` (1.094s),
  and the auth test fails on its *positive* assertion, which confirms both fail-closed
  assertions ran first and passed.

**Mutation proof for the empty-credential control** (a revert cannot exercise it — the
pre-fix tree rejects everything). Mutating the production `Token` branch to fail open on
an empty credential (`if token.is_empty() { ExtractedToken::None }`):

```
FAIL [3.741s] test_3137_galaxy_token_scheme_authenticates_api_token
  `Token ` with an empty credential must fail closed as InvalidCredential
  (not anonymous, not resolved), got NoCredential
```

The old `"Token not-a-valid-credential"` assertion passes under that mutation. All
timings are 0.2–4.8s, not the ~0.04s silent-skip path.

**Green run:** `api::handlers::ansible::` + `api::middleware::auth::` → **277/277
passed**. `cargo fmt --all --check` and
`cargo clippy --workspace --all-targets -- -D warnings` clean. No migration; no
`sqlx::query!` text touched.

No collision with #3272 (goproxy/hex/rubygems/proxy_helpers).

## Follow-ups filed

- **#3282** — `collection publish` fails with `KeyError: 'task'`; needs the `task` field
  plus a `v3/imports/collections/{id}/` route (milestoned 1.7.2: it is a bug on the
  reported command path and needs no migration, though the new route is arguably
  feature-shaped for a patch line — move it to 1.8.0 if you read it that way).
- **#3283** — no integration/DTF/conformance coverage and no docs page for Ansible,
  which is why three mechanical breaks reached a user report behind green CI.


brandonrc:
## Independent validation review — head `6427dd6`

**Verdict: the two production changes are correct, and I found no way for the `Token`
scheme to reach weaker validation than `Bearer`.** Both fixes are grounded in
ansible-core's own source and I re-derived every citation from upstream rather than
from the PR body. All five claims **CONFIRMED**.

One substantive problem, and it is a scope problem rather than a code problem:
**`ansible-galaxy collection publish` — the exact command in #3137's repro — still
fails after this PR**, at the step immediately after the two that were fixed. Details
in Defect 1. Everything else is test-strength and documentation polish.

Method: fresh shallow clone, `pull/3278/head`, private Postgres, revert-proof run with
`DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1`. Upstream quotes are from
`ansible/ansible@stable-2.18` (and `devel` where noted), not from memory.

---

### Claim 1 — discovery is requested at `<server>/api`, no trailing slash — **CONFIRMED**

`lib/ansible/galaxy/api.py` is exactly as quoted:

```python
def _urljoin(*args):
    return '/'.join(to_native(a, errors='surrogate_or_strict').strip('/') for a in args + ('',) if a)
```

`if a` filters the appended `''` before the `.strip('/')` runs, so no trailing slash
survives, and `g_connect`'s fallback `n_url = _urljoin(n_url, '/api/')` yields
`<server>/api`. Route registered at `backend/src/api/handlers/ansible.rs:58`.

Two upstream details the PR body does not mention, both of which *strengthen* it:

1. The same `except` block opens with
   `if n_url.endswith('/api') or n_url.endswith('/api/'): raise`. So a user who
   configures the server URL **as** `http://host/ansible/<repo>/api` got no fallback
   at all — the first probe 404'd and the CLI aborted immediately. The alias fixes
   that configuration too, not just the bare-server-URL one.
2. `g_connect` ends with `self.api_server = n_url`. **This is the actual reason Fix 1
   is sufficient on its own**, and it is stronger than the body's argument. The body
   says "later v3 URLs get an explicit `+ '/'` client-side", which is true of
   `publish_collection` (`_urljoin(...) + '/'`) but not literally true of
   `get_collection_versions`, which instead passes `'/?%s=%d'` as the final `_urljoin`
   component. The reassignment is what guarantees every subsequent v3 URL is built
   from `<server>/api`, whatever the trailing-slash mechanism. Net effect identical;
   worth correcting in the body since the stated reason is only half the reason.

Registering both spellings also matches existing repo precedent —
`backend/src/api/handlers/pypi.rs:55-59` registers `/simple` and `/simple/` the same
way. No route-table conflict: `/:repo_key/api` cannot shadow `/:repo_key/api/v3/...`
or `/:repo_key/download/*file_path`, no catch-all exists in this router, and the
Caddy front end proxies `/ansible/*` wholesale (`docker/Caddyfile:41`), so nothing
upstream of axum needs a matching change.

### Claim 2 — only discovery needs the alias — **CONFIRMED** (see the correction above)

### Claim 3 — `Authorization: Token <api_key>` was rejected as malformed — **CONFIRMED**

`lib/ansible/galaxy/token.py`: `GalaxyToken.token_type = 'Token'`;
`KeycloakToken.token_type = 'Bearer'`; `BasicAuthToken.token_type = 'Basic'`. Header
built as `'%s %s' % (self.token_type, self.get())`.

The chain checks out end to end: `auth.rs:466-493` — the value contains a space, so
the scheme-less cargo fallback (`!auth_header.contains(' ')`, `auth.rs:490`) does not
apply and it falls to `ExtractedToken::Invalid`; `auth.rs:1038` maps that to
`InvalidCredential`; `repo_visibility_middleware` returns 401 at `auth.rs:1883`.

One more upstream detail that explains the reporter's confusing 401-**then**-404
transcript, and which is worth recording: `_add_auth_token(headers, url, required=auth_required)`
attaches the header whenever `self.token` is set, **even for the `auth_required=False`
discovery probe**. That is why a configured token 401'd discovery itself rather than
only the later authenticated calls.

### Claim 4 — `Token` is Bearer-equivalent, no weakening — **CONFIRMED**

This is the claim I attacked hardest, and it holds by construction: the branch returns
the *same* `ExtractedToken::Bearer` variant (`auth.rs:478`), not a parallel one. There
is no second validator that could diverge. I traced every consumer of `ExtractedToken`:

| Consumer | `Bearer` arm |
|---|---|
| `auth_middleware` (`auth.rs:640`) | `validate_access_token_async` → `validate_api_token_with_scopes` |
| `try_resolve_auth_outcome` (`auth.rs:949`) | same, plus the base64 `user:pass` fallback |
| `optional_auth_middleware`, `admin_middleware` | both delegate to `try_resolve_auth_outcome` |
| `extract_visibility_token` (`auth.rs:1514`) | header credential wins; unchanged |

Nothing downstream sees the scheme. Expiry, revocation, deactivation and the
`prefix_secret` lookup all live in `AuthService::validate_api_token`
(`backend/src/services/auth_service.rs:2148+`); scope-gated admin demotion (GHSA-vvc3)
is in `validate_api_token_with_scopes` (`auth.rs:841-866`); `allowed_repo_ids`
(`can_access_repo`), `has_scope`, and the permission-service checks all run against the
resolved `AuthExtension` further down `repo_visibility_middleware`. A `Token`
presentation reaches all of them identically.

- **No privilege gain.** Every credential now accepted under `Token` was already
  accepted under `Bearer`. The change adds a *presentation*, not a capability.
- **No anonymous elevation.** An unresolvable `Token <x>` still yields
  `InvalidCredential` → 401 — the same status the pre-fix `Invalid` produced. There is
  no optional-auth path where the pre-fix `Invalid` was treated as anonymous and the
  post-fix `Bearer(bad)` would authenticate; both `auth.rs:1312` and `auth.rs:1883`
  short-circuit on `credential_invalid`.
- **`Token ` with an empty credential** → `Bearer("")` → `""` has no `_`, so no
  `prefix_secret` lookup matches, and `decode_basic_credentials("")` returns `None` →
  `InvalidCredential`. Same as the pre-existing `"Bearer "` case already pinned by
  `test_extract_bearer_empty_token` at `auth.rs:2128-2130`.
- **Whitespace.** `"Token  abc"` yields `Bearer(" abc")`, which fails validation
  (fail-closed). ansible-core never emits it.

**Cross-format regression — verified independently, agree with the author.** Nothing
relied on `Token <x>` being rejected. The one test that enumerates unrecognised schemes
(`auth.rs:2255-2281`) uses `"Bogus scheme-with args"` and `"!!! bad !!!"`, neither of
which the new branch matches, and the `X-NuGet-ApiKey` no-fallback invariant it guards
is untouched. `grpc/auth_interceptor.rs:114`, `oci_v2.rs:173`,
`services/image_scanner.rs:1035` and `services/proxy_service.rs:2286,2495` each parse
`Bearer` themselves and are unaffected — a REST/gRPC parity gap, but a strictly
*stricter* one, so not a security issue.

### Claim 5 — no collision with #3272 — **CONFIRMED**

`#3272` touches `goproxy.rs`, `hex.rs`, `proxy_helpers.rs`, `rubygems.rs`,
`test_db_helpers.rs`. Disjoint from this PR's two files.

### Revert-proof — independently reproduced

I reverted the production hunks to merge-base shape (`1115203`) and kept the tests.
`git diff 1115203 --shortstat` on the **reverted** tree: `2 files changed, 106
insertions(+)` — insertions only, every one inside a `mod tests` block, so production
code is byte-identical to the merge base. Matches the PR body exactly.

Reverted run (`DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1`): **3/3 FAIL**, with the DB test
at **1.176s** and the router test at 0.261s — genuinely executed, not the ~0.04s
silent-skip path.

```
FAIL [0.030s] test_3137_extract_galaxy_token_scheme_as_bearer
FAIL [0.261s] test_3137_api_discovery_without_trailing_slash
             left: 404, right: 200
FAIL [1.176s] test_3137_galaxy_token_scheme_authenticates_api_token
             expected Resolved for `Token <valid_api_key>`, got InvalidCredential
```

---

## Defects

### 1. HIGH (scope/claim) — `ansible-galaxy collection publish` still fails; the PR body says otherwise

The body opens: *"`ansible-galaxy` could not talk to an Artifact Keeper Ansible
repository at all — neither `install` nor `publish` got past API discovery."* That
reads as "both work now". `install` does. `publish` does not, and it is the command in
#3137's repro.

`GalaxyAPI.publish_collection` ends with a hard index (stable-2.18):

```python
resp = self._call_galaxy(n_url, args=b_form_data, headers=headers, method='POST',
                         auth_required=True, error_context_msg=...)
return resp['task']
```

(`devel` is the same, wrapped: `return urljoin(self.api_server, resp['task'])`.)

AK's `upload_collection` returns `202` with
`{namespace, name, version, href, download_url}` — **no `task` key**
(`backend/src/api/handlers/ansible.rs:666-679`). So the POST succeeds, the artifact is
stored, and the CLI then dies on `KeyError: 'task'` → *"Unexpected Exception, this is
probably a bug: 'task'"*.

`--no-wait` does **not** help: `resp['task']` is dereferenced inside
`api.publish_collection`, before `publish_collection(collection_path, api, wait, timeout)`
in `lib/ansible/galaxy/collection/__init__.py` ever consults `wait`.

And even once a `task` field is added, the default path polls
`_urljoin(self.api_server, 'v3/', 'imports/collections', task_id, '/')`. There is **no
`imports/collections` route** in `ansible.rs`, and `wait_import_task` treats 404 as
"import hasn't started yet" and retries:

```python
except GalaxyError as e:
    if e.http_code != 404:
        raise
    display.vvv('Galaxy import process has not started, wait %s seconds before trying again')
    time.sleep(wait); continue
```

with `while timeout == 0 or ...` and
`publish_parser.add_argument('--import-timeout', dest='import_timeout', type=int, default=0)`
plus `'--no-wait', dest='wait', action='store_false', default=True`. So the naive fix
(add `task`, no route) converts a crash into an **indefinite hang**.

**Suggested resolution — no need to grow this PR.** Keep the diff as is, but (a) soften
the body to say discovery and authentication are fixed and `install` now works
end-to-end, while `publish` remains blocked at the import-task contract; and (b) open a
follow-up issue for: return `{"task": "/ansible/{repo}/api/v3/imports/collections/{id}/"}`
from `upload_collection`, and register a `GET /:repo_key/api/v3/imports/collections/:id/`
returning at minimum `{"state": "completed", "finished_at": <ts>, "messages": []}` —
`wait_import_task` breaks on `finished_at` and raises on `state == 'failed'`. Since
AK's upload is synchronous, a synthetic already-finished task keyed off the artifact id
is sufficient. If the reporter is told #3137 is fixed without this, their original
command still fails.

### 2. MEDIUM (test) — the negative control cannot fail on the unfixed tree, and is never reached on a revert

`auth.rs:4962-4988`. `"Token not-a-valid-credential"` contains a space, so on the merge
base it takes `ExtractedToken::Invalid` → `AuthOutcome::InvalidCredential`
(`auth.rs:1038`); with the fix it takes `Bearer("not-a-valid-credential")` → also
`InvalidCredential`. **Identical outcome on both trees**, so it distinguishes nothing.

It is also positioned *after* the `panic!` in the positive `match` (`auth.rs:4983`), so
on the reverted tree it is never evaluated at all — my reverted run confirms the test
aborts on the positive assertion.

To be fair: as a forward-looking guard against a future "`Token` ⇒ auto-resolve"
mistake it has real value, and I would keep it. It just should not be described as a
"must not fail open" control that the revert-proof exercised — the revert-proof never
reached it. A control that *can* discriminate would be an empty credential
(`"Token "` → must be `InvalidCredential`, not `Resolved`), asserted before the
positive `match` so a revert still evaluates it.

### 3. LOW (test) — neither fix is tested through the middleware that actually returned 401/404

`test_3137_galaxy_token_scheme_authenticates_api_token` calls `try_resolve_auth_outcome`
directly; `test_3137_api_discovery_without_trailing_slash` uses `router_anon`. Neither
exercises `repo_visibility_middleware`, which is where the reporter's 401 came from,
and **no test covers the two fixes together** — a `GET /ansible/{repo}/api` carrying
`Authorization: Token <valid>` returning 200 is the actual customer contract and is
currently unpinned.

Relatedly, the third assertion in the router test —
`GET /no-such-repo/api` must 404, commented *"the alias must not bypass repo
resolution"* — is true of the bare handler router but **not of the deployed stack**:
`repo_visibility_middleware` returns **401** for an unknown repo key with no credential
(`auth.rs:1834-1836`, the #1808 repo-existence-oracle closure), before the handler ever
runs. Harmless as written, but the comment overstates what it proves.

There is also no integration or DTF coverage for ansible anywhere (`backend/tests/` has
no ansible file; `docs/` has no ansible page) — which is exactly why a bug this
mechanical survived to a user report. A conformance entry driving the real
`ansible-galaxy` URL shapes would be the durable guard, and would have caught Defect 1
as well.

### 4. LOW (docs) — the extractor's own doc comments now understate what it accepts

`auth.rs:465` still reads *"supports Bearer, ApiKey, and Basic schemes"*, `auth.rs:498`
*"Checks: Authorization (Bearer/ApiKey), X-API-Key"*, and the `ExtractedToken::Bearer`
variant doc (`auth.rs:456`) *"JWT or API token from Bearer scheme"*. Given how much of
this file's correctness is carried by its comments, worth a one-line update each.

### 5. LOW (process) — no CHANGELOG entry

`CHANGELOG.md` `[Unreleased] / ### Fixed` carries per-issue entries for comparable
client-compatibility fixes (#3251 added one in the same series). Not CI-gated — the
release-bump gate at `.github/workflows/ci.yml:198` only lists `CHANGELOG.md` as
*allowed*, never required — but a user-visible "your Galaxy client now works" fix is
squarely the kind of thing that section exists for.

### Note (not a defect) — scheme case sensitivity

`strip_prefix("Token ")` is case-sensitive. RFC 9110 §11.1 makes the auth scheme
case-insensitive, so a strict reading wants `token <x>` accepted too. **This is not
introduced here**: `Bearer` has always been matched case-sensitively at `auth.rs:467`,
while `Basic`/`basic` (`auth.rs:483-484`) and `extract_bearer_credentials`
(`auth.rs:391`) are not — the inconsistency predates this PR. ansible-core emits
exactly `Token`, so there is no client impact, and being stricter fails closed. Worth
a separate normalization issue for the whole extractor rather than a change here.

---

## Credit where it is due

The diagnosis method was the right one and is the main reason this review was short:
the author read ansible-core's `_urljoin`, `g_connect` and `GalaxyToken` and quoted the
normative lines instead of guessing at client behaviour, which is what turned a vague
"404/401" report into two precise one-line changes. The revert-proof was genuine —
faithful production revert, DB actually connected, timings that prove it. And the
security instinct was correct: reusing the existing `ExtractedToken::Bearer` variant
rather than adding a parallel one is precisely what makes "no weakening" provable by
construction instead of by argument.

Nothing here blocks the code. Recommend: correct the `publish` claim in the body, file
the import-task follow-up, and optionally tighten the negative control.

brandonrc:
## Review remediation — pushed `324a6e2`

Thanks — the `publish` finding is the one that mattered, and you are right that shipping
this while telling the reporter their command works would have been worse than shipping
nothing. Production code is unchanged in this round; everything below is scope
correction, test strength and documentation.

Fresh shallow clone, `pull/3278/head`, own Postgres (`:30678`), every run with
`DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1`.

### Defect 1 (HIGH, scope) — corrected, and split out

**`Fixes #3137` is kept**, deliberately, and I want the reasoning on the record because
it is a judgement call rather than an obvious one.

#3137 is titled "Ansible galaxy discovery". Its body names exactly two defects — "the
galaxy client adds /api to its discovery of the configured url, however the
artifactkeeper only handles /api/", and "Galaxy sends the header in the format of
`Token <token>` while AK expects `Bearer <token>`" — and the failure in its own
transcript is `Error when finding available api versions ... (HTTP Code: 401)`. Both
defects are closed by this diff. `collection publish` is the *command* used to surface
them, but the report is not about the publish contract, which the client never reached.

So: issue closed as titled, residual stated explicitly, follow-up filed — which is also
what keeps `require-linked-issue` satisfied. **If you would rather hold #3137 open until
`publish` works end to end, that is a defensible read of "the reporter's command still
fails" and I will retarget the link to #3282 on request.**

The body no longer claims `publish` works. It now opens with a scope section stating
that `install` works end to end and `publish` does not, with the mechanism, why
`--no-wait` does not help, and why a naive `task` key makes it worse. I also folded in
your two upstream corrections: the `n_url.endswith('/api')` early-`raise` (the alias
fixes that configuration too), and `self.api_server = n_url` as the actual reason one
alias suffices — the old "later v3 URLs get an explicit `+ '/'`" line was true of
`publish_collection` but not of `get_collection_versions`. Same for the
`_add_auth_token(..., required=auth_required)` detail that explains the 401-then-404
ordering in the reporter's transcript.

**#3282** — `publish` fails with `KeyError: 'task'`. Full mechanism, the `--no-wait`
frame-ordering, the 404-retry hang (`wait_import_task` + `--import-timeout` default `0`),
and your proposed shape: return
`{"task": "/ansible/{repo}/api/v3/imports/collections/{id}/"}` and register a
repo-scoped `GET .../imports/collections/:id/` returning
`{"state": "completed", "finished_at": ..., "messages": []}`, synthetic and
already-finished since AK's upload is synchronous. Acceptance criteria included.

Milestoned **1.7.2**, with the reasoning stated in the issue: it is a bug on the exact
command the reporter ran, reachable in 1.7.1, and needs no schema migration — AK already
advertises `v3` in its discovery document, so serving a non-conformant publish response
is a conformance bug rather than a missing feature. The counter-argument is real and I
noted it in the issue: it does add a route, which the patch-line rule treats as feature
work. Re-milestone to 1.8.0 if you read it that way.

**#3283** — the coverage gap, filed separately since it outlives the publish fix:
no integration/DTF/conformance coverage and no docs page for Ansible anywhere. It
tabulates all three breaks (two fixed here, one open) against what the unit tests
structurally could not see, and proposes an Ansible conformance entry driving the real
client URL shapes. Milestoned 1.8.0.

### Defect 2 (MEDIUM, test) — negative control now discriminates, and runs first

Confirmed on both counts: `"Token not-a-valid-credential"` produced `InvalidCredential`
on the fixed *and* unfixed trees, and it sat after the positive `panic!` so a revert
never reached it.

Both fail-closed assertions now run **before** the positive `match`. The discriminating
input is an empty credential, `"Token "`, and the assertion is specifically
`InvalidCredential` — **not** `NoCredential`. That distinction is the operationally
meaningful one: `repo_visibility_middleware` 401s on `InvalidCredential` but lets
`NoCredential` through as an anonymous read on a public repository, so an implementation
that mapped an empty `Token` credential to "no credential presented" would silently
downgrade a broken client to anonymous instead of challenging it. That is the fail-open
shape worth pinning, and it is stated in the test comment.

A revert cannot exercise this (the pre-fix tree rejects everything), so here is the
mutation proof you asked for. Production branch mutated to accept an empty credential:

```rust
} else if let Some(token) = auth_header.strip_prefix("Token ") {
    if token.is_empty() { ExtractedToken::None } else { ExtractedToken::Bearer(token) }
```

```
FAIL [3.741s] api::middleware::auth::tests::test_3137_galaxy_token_scheme_authenticates_api_token
  panicked at auth.rs:5004:22:
  `Token ` with an empty credential must fail closed as InvalidCredential
  (not anonymous, not resolved), got NoCredential
```

3.741s — genuinely DB-executed. The old assertion passes under that same mutation, which
is the point. I kept the unknown-credential assertion, relabelled in-code as a
forward-looking guard against a future "the `Token` scheme is trusted" change rather
than as a control the revert-proof exercised.

### Defect 3 (LOW, test) — the actual user path is now pinned

New `test_3137_galaxy_api_through_visibility_middleware` in `ansible.rs`. It mounts the
**real** handler table under the **real** `repo_visibility_middleware` at the real
`/ansible` prefix (so `extract_repo_key` sees the production path shape) and issues the
request the CLI actually makes: `GET /ansible/{repo}/api`, no trailing slash,
`Authorization: Token <api_key>`, against a **private** repository → 200 with `v3`
advertised. First test covering both fixes together, and the first covering either one
through the middleware that produced the reported 401.

Fail-closed controls in the same fixture, so the 200 cannot be explained by a permissive
middleware: anonymous → 401; bogus `Token` credential → 401; **unknown repo key → 401,
not 404**. That last one is your #1808 point, pinned as production truth. I left the
bare-router 404 assertion in `test_3137_api_discovery_without_trailing_slash` — it
correctly proves the alias is repo-scoped like every other route in that table — but
rewrote its comment to say so and to name the middleware test as the place the deployed
status lives.

Per-fix reverts, to prove the new test gates each fix rather than only their union:

| revert | middleware test | time |
|---|---|---|
| route alias removed only | `left: 404, right: 200` | 2.906s |
| `Token` branch removed only | `left: 401, right: 200` | 1.094s |
| both (faithful merge-base shape) | `left: 401, right: 200` | 1.090s |

Under the `Token`-only revert the auth test fails on its **positive** assertion — which
is the direct evidence that both fail-closed assertions ran first and passed.

Full faithful revert: **4/4 FAIL** (0.012s / 0.216s / 1.090s / 1.141s). Faithfulness
checked the strong way rather than by shortstat, which does not apply this round because
the tests were rewritten: stripping every `#[cfg(test)] mod tests` block from the
reverted files and diffing against `1115203` leaves **only `//` and `//!` comment
lines** — zero executable production statements differ.

### Defect 4 (LOW, docs) — done

`ExtractedToken::Bearer`'s variant doc, the `extract_token_from_auth_header` doc and the
`extract_token` header list all name `Token` now (and the scheme-less cargo credential,
which was also unmentioned).

### Defect 5 (LOW, process) — done

`[Unreleased] / ### Fixed` entry, which states plainly that `collection publish` is not
fixed by this change and points at #3282.

### Note (scheme case sensitivity)

Agreed, and deliberately left alone: `Bearer` has always been matched case-sensitively
while `Basic`/`basic` are both accepted, so normalizing here would fix half of an
inconsistency inside an unrelated change. ansible-core emits exactly `Token`, and being
stricter fails closed. The PR body now says this explicitly so the omission reads as a
decision rather than an oversight.

### Verification

- `api::handlers::ansible::` + `api::middleware::auth::` → **277/277 passed**
  (276 before, +1 new test), timings 0.2–4.8s.
- `cargo fmt --all --check` clean; `cargo clippy --workspace --all-targets -- -D warnings`
  clean.
- No migration, no `sqlx::query!` text touched, no production behaviour change in this
  round.

Issue #3283, opened 2026-08-10
Ansible Galaxy format has no integration/DTF/conformance coverage and no docs page


brandonrc:
## Summary

The Ansible Galaxy format has **no integration, DTF, or conformance coverage at all**,
and no documentation page. `backend/tests/` contains no ansible file; `docs/` has no
ansible page. The only coverage is `#[cfg(test)]` unit tests inside
`backend/src/api/handlers/ansible.rs`, which exercise handlers against the bare router
— never the real client request shapes, and never the middleware stack in front of
them.

That gap is the direct cause of three separate user-visible breaks, all of them
mechanical, all of them behind green CI:

| Break | Issue | What the unit tests could not see |
|---|---|---|
| Discovery registered only at `/api/`, client requests `/api` | #3137 | the route table was tested with the spelling the code registered |
| `Authorization: Token <t>` rejected as malformed | #3137 | no test ever presented a real client's header bytes |
| `publish` response omits the `task` import-task field | #3282 | no test ever ran the client past the upload POST |

Each was a one- or two-line defect that any run of the real `ansible-galaxy` binary
would have caught immediately.

## Proposed work

An Ansible entry in the conformance corpus (the program that landed pypi/oci in
artifact-keeper-test #333), driving the request shapes ansible-core actually emits
rather than the ones the handler happens to register:

- `GET <server>/api` **and** `GET <server>/api/` — both spellings, since ansible-core's
  `_urljoin` strips trailing slashes and the CLI reassigns `api_server` from the
  discovery URL it succeeded on.
- `Authorization: Token <api_key>` on every authenticated call (`GalaxyToken.token_type
  = 'Token'`), not `Bearer` — including on the `auth_required=False` discovery probe,
  which `_add_auth_token` attaches the header to whenever a token is configured.
- The full `install` path: discovery → `v3/collections/{ns}/{name}/versions/` →
  version detail → artifact download.
- The full `publish` path: upload POST → `task` field → `v3/imports/collections/{id}/`
  poll → completion (blocked on #3282).
- A private-repository variant through `repo_visibility_middleware`, which is where
  the reported `401` came from, and where the format handlers' own tests do not reach.

Plus a `docs/` page covering `ansible.cfg` / `--server` configuration against an AK
repository and how to mint the API key the `Token` scheme carries.

## Notes

- The conformance-corpus approach is preferred over hand-written integration tests:
  driving the published client contract is what makes these breaks impossible to
[...]

PR #3288, opened 2026-08-11
fix(ansible): return the Galaxy v3 import task and serve the import-status route

### LINKED ISSUE #3282: Ansible: `ansible-galaxy collection publish` fails with KeyError 'task' — upload response omits the Galaxy v3 import-task contract
## Summary

`ansible-galaxy collection publish` still fails against an Artifact Keeper Ansible
repository, **after** #3278 fixed the two discovery/authentication breaks reported in
#3137. The upload itself succeeds — the artifact is stored — but the CLI then crashes
on the response body, because AK's `202` does not carry the `task` field the Galaxy v3
publish contract requires.

This is the step immediately after the two that #3278 fixed, on the exact command in
#3137's repro. Discovery and `install` work; `publish` does not.

## Mechanism

`GalaxyAPI.publish_collection` (ansible-core `lib/ansible/galaxy/api.py`, `stable-2.18`)
ends with a hard index into the response:

```python
resp = self._call_galaxy(n_url, args=b_form_data, headers=headers, method='POST',
                         auth_required=True, error_context_msg=...)
return resp['task']
```

(`devel` is the same, wrapped: `return urljoin(self.api_server, resp['task'])`.)

AK's `upload_collection` returns `202` with
`{namespace, name, version, href, download_url}` — no `task` key
(`backend/src/api/handlers/ansible.rs`). The POST succeeds, the collection is stored,
and the client dies with:

```
ERROR! Unexpected Exception, this is probably a bug: 'task'
```

### `--no-wait` does not work around it

`resp['task']` is dereferenced **inside** `api.publish_collection`, before
`publish_collection(collection_path, api, wait, timeout)` in
`lib/ansible/galaxy/collection/__init__.py` ever consults `wait`. The flag is applied
one frame too late to help.

### A naive fix turns the crash into an indefinite hang

Adding a `task` key on its own is worse than the crash. After `publish_collection`
returns, the default path polls the import task at
`_urljoin(self.api_server, 'v3/', 'imports/collections', task_id, '/')`. There is **no
`imports/collections` route** in `ansible.rs`, and `wait_import_task` treats a 404 as
"the import has not started yet" and retries forever:

```python
except GalaxyError as e:
    if e.http_code != 404:
        raise
    display.vvv('Galaxy import process has not started, wait %s seconds before trying again')
    time.sleep(wait); continue
```

with `while timeout == 0 or ...`, and the CLI defaults are
`--import-timeout` → `default=0` (never time out) and `--no-wait` →
`dest='wait', action='store_false', default=True` (wait by default). So the loop never
exits. **The route and the response field have to land together.**

## Proposed fix

AK's upload is synchronous — by the time the `202` is written the artifact is already
stored — so a synthetic, already-finished import task keyed off the stored artifact is
sufficient. No queue, no worker, no schema migration.

1. `upload_collection` returns
   `{"task": "/ansible/{repo_key}/api/v3/imports/collections/{id}/"}` alongside the
   existing fields.
2. Register `GET /:repo_key/api/v3/imports/collections/:id/` returning at minimum:
   ```json
   {"state": "completed", "finished_at": "<ts>", "messages": []}
   ```
   `wait_import_task` breaks out of its loop on `finished_at` and raises on
   `state == "failed"`, so those three fields are the whole client-visible contract.
3. Keep the route repo-scoped and behind the same visibility middleware as every other
   `/ansible/{repo}/...` route, so the task id is not a cross-repo read oracle.

The `task` value must be resolvable by the client: `devel` joins it against
`api_server` while `stable-2.18` passes it to `_urljoin(self.api_server, 'v3/',
'imports/collections', task_id, '/')`, so verify the emitted shape against **both**
before landing.

## Acceptance criteria

- `ansible-galaxy collection publish <tarball> --server <ak-repo>` exits 0 against a
  local AK Ansible repository, with no `--no-wait`, and the collection is then
  installable from the same repository.
- `--no-wait` also exits 0.
- A publish that fails validation surfaces as a client-visible error, not a hang.
- The import-status route is repo-scoped and does not leak task existence across repos.
- A regression test drives the real client URL shapes end to end (see the coverage gap
  below), not just the handler in isolation.

## Why this was not caught

There is no integration or DTF coverage for the Ansible format anywhere:
`backend/tests/` has no ansible file and `docs/` has no ansible page. The unit tests
cover handlers in isolation, which is exactly why two mechanical URL/header breaks
survived to a user report — and why this third one is still open. Tracked separately in #3283; a conformance entry driving the real `ansible-galaxy`
request shapes is the durable guard for all three.

## References

- Found by the independent review of #3278 (which fixed #3137's discovery and
  `Token`-scheme breaks). #3278 deliberately does **not** attempt this: it needs a new
  route and an import-status contract, which warrants its own review.
- ansible-core `stable-2.18`: `lib/ansible/galaxy/api.py`
  (`publish_collection`, `wait_import_task`), `lib/ansible/galaxy/collection/__init__.py`
  (`publish_collection`), `lib/ansible/cli/galaxy.py` (`--import-timeout`, `--no-wait`).

brandonrc:
## What broke

`ansible-galaxy collection publish` crashed with

```
ERROR! Unexpected Exception, this is probably a bug: 'task'
```

immediately after a **successful** upload — the collection was stored, and the client
died reading the response. #3278 fixed discovery and the `Token` auth scheme, which
unblocked `install`; this is the next step in the same command.

The Galaxy v3 publish contract is two calls, not one. `GalaxyAPI.publish_collection`
(`lib/ansible/galaxy/api.py`) ends with a hard index into the upload response, and
`upload_collection` returned `202` with no `task` key:

| branch | last line of `publish_collection` |
|---|---|
| `stable-2.17`, `stable-2.18`, `stable-2.19`, `stable-2.20` | `return resp['task']` |
| `devel` | `return urljoin(self.api_server, resp['task'])` |

`--no-wait` cannot work around it. `wait` is consulted one frame upstream, in
`lib/ansible/galaxy/collection/__init__.py::publish_collection`, after the dereference
has already raised.

## Why both halves had to land together

Adding a `task` key on its own is *worse* than the crash. After `publish_collection`
returns, the client polls the import task, and `wait_import_task` treats a `404` as
"the import has not started yet":

```python
except GalaxyError as e:
    if e.http_code != 404:
        raise
    display.vvv('Galaxy import process has not started, wait %s seconds before trying again' % wait)
    time.sleep(wait)
    continue
```

with `while timeout == 0 or ...`, and the CLI defaults are `--import-timeout`
`type=int, default=0` (never time out) and `--no-wait`
`dest='wait', action='store_false', default=True` (wait by default,
`lib/ansible/cli/galaxy.py`). So a missing route is an indefinite hang, not an error.
The field and the route ship in the same commit.

## The verified schema

Read off `wait_import_task`, which is identical on `stable-2.17` through `devel`.
These are the only fields the client touches:

| expression | how it is used |
|---|---|
| `data.get('state', 'waiting')` | anything but `'waiting'` or `'failed'` is success; `'waiting'` at loop exit raises `Timeout while waiting for the Galaxy import process to finish` |
| `data.get('finished_at', None)` | **truthy breaks the poll loop** — this is the only exit |
| `data.get('messages', [])` | iterated unconditionally; each entry is hard-indexed for `['level']` and `['message']` |
| `data['error']` | read **only** when `state == 'failed'`, then `.get('code')` / `.get('description')` |

The served document is:

```json
{
  "id": "<uuid>", "state": "completed",
  "created_at": "...", "started_at": "...", "finished_at": "...", "updated_at": "...",
  "namespace": "community", "name": "hashi_vault", "version": "7.1.0",
  "href": "/ansible/{repo}/api/v3/imports/collections/{id}/",
  "messages": []
}
```

`state`/`finished_at`/`messages` are the client-visible contract; the rest is identity
for humans and UIs, named after the fields Galaxy NG serves.

## Why `task` is a root-absolute path

One spelling has to satisfy both live client generations, and they consume the value
differently:

* **`stable-2.17`..`stable-2.20`** — the caller in `collection/__init__.py` reduces the
  value to its **last non-empty path segment**
  (`for path_segment in reversed(import_uri.split('/'))`), and `wait_import_task(task_id)`
  rebuilds `_urljoin(api_server, 'v3/', 'imports/collections', task_id, '/')`. `g_connect`
  has already rewritten `self.api_server` to the URL that served `available_versions`
  (`self.api_server = n_url`), so that resolves to
  `/ansible/{repo}/api/v3/imports/collections/{task_id}/`.
* **`devel`** — `urljoin(self.api_server, resp['task'])` is passed to `wait_import_task`
  **verbatim**; there is no segment extraction on this branch. A root-absolute reference
  resolves against the origin, giving the same URL.

So `/ansible/{repo_key}/api/v3/imports/collections/{id}/` is correct for both, and it is
the shape ansible-core itself documents as the v3 reference response:

```python
# v3: {"task": "/api/automation-hub/v3/imports/collections/838d1308-a8f4-402c-95cb-7823f3806cd8/"}
```

A **bare id** would break `devel` (relative resolution drops the last segment of
`api_server`); a **fully-qualified URL** would make the stable line's `_urljoin` emit a
doubled path. Both are covered by mutations below.

## Synthetic vs real: it is synthetic, and the code says so

Artifact Keeper has **no asynchronous import pipeline**. `upload_collection` validates,
stores and inserts the artifact row inline, so by the time the `202` is written the
"import" is already finished. The route therefore reports a **completed import derived
from the stored artifact**, not progress from a job queue. It reports no intermediate
states because there are none, and the single timestamp is the artifact's real
`created_at` rather than three invented ones. This is stated plainly in the module docs,
the handler comment and the CHANGELOG — nothing here implies async processing exists.

A publish that fails validation never reaches this route: it is returned as a `4xx` from
the upload itself, which `_call_galaxy` raises on, so a rejected publish is a
client-visible error rather than a task the client would poll forever.

The artifact id doubles as the task id — the import *is* the upload — so no new table,
no queue, and **no migration**.

## Authorization

The route is registered in `handlers::ansible::router()`, which `routes.rs` nests under
`/ansible` inside `format_routes` — the router that carries `repo_visibility_middleware`
(`backend/src/api/routes.rs`). That middleware is the credential authority for every
format route, so a private repository's import tasks are unreachable without a
credential that can read the repository, exactly like the collection endpoints beside
it. This is **pinned by a test through the real middleware**, not assumed. No
`can_access_repo` (token scope, not visibility) and no bare `require_scope`
(`scopes: None` satisfies it unconditionally) is involved.

Repo scoping: the lookup is keyed on `(id, repository_id)`, so a task id minted in one
repository is a plain `404` under another and the route is not a cross-repo existence
oracle.

A malformed id answers `404` rather than `400`, so the route does not confirm which ids
are even well-formed.

## Tests

Four DB-backed `--lib` tests (so they count for the coverage gate). The two helpers
`stable_line_poll_path` / `devel_poll_path` **re-implement ansible-core's own URL
derivation**, so the tests drive the client's real request shapes rather than echoing
back whatever the handler formats — no assertion compares the response to the same
constant the handler emits.

| test | claim |
|---|---|
| `test_3282_publish_task_resolves_for_both_client_generations` | publish → `task` → both client derivations produce the **same** URL → it resolves → `state != waiting/failed`, `finished_at` truthy, `messages` an array. Positive control in the same fixture: the #3278 `install` path (version metadata + download of the published bytes) still works. |
| `test_3282_import_status_is_repo_scoped` | repo A's task `404`s under repo B, with a caller holding the **widest** access (`AccessScope::Admin`, `scopes: None`) so the `404` can only come from the SQL predicate. Positive controls in the same fixture: A's task resolves under A, and B's own publish round-trips. |
| `test_3282_import_status_through_visibility_middleware` | the real client URL through the real `repo_visibility_middleware` against a **private** repo, publishing with `Authorization: Token <api_key>`. Fail-closed controls in the same fixture: anonymous → `401`, bogus credential → `401`. |
| `test_3282_failed_publish_returns_an_error_not_a_task` | a rejected publish (sha256 mismatch) is a `400` carrying **no** `task`. Positive control in the same fixture: an accepted publish **does** carry one — so the negative assertion can actually fail. |

### Revert-proof

Faithful revert = merge-base production code + the new tests only. Production code on
the reverted tree is **byte-identical** to the merge base:

```
$ git diff e91e970 --shortstat          # REVERTED tree
 1 file changed, 439 insertions(+)      # insertions only, all inside #[cfg(test)] mod tests

$ # strip the #[cfg(test)] module from merge-base and from the reverted file:
mergebase production bytes = 26110   sha256 79d10eba...3eaff4ca
reverted  production bytes = 26110   sha256 79d10eba...3eaff4ca   -> IDENTICAL
```

Before (reverted, `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1` both exported):

```
FAIL [0.271s] test_3282_failed_publish_returns_an_error_not_a_task
FAIL [0.278s] test_3282_publish_task_resolves_for_both_client_generations
FAIL [0.385s] test_3282_import_status_is_repo_scoped
FAIL [2.962s] test_3282_import_status_through_visibility_middleware
Summary [2.981s] 4 tests run: 0 passed, 4 failed
  panicked at ansible.rs:1458: task          <- the missing key, i.e. the claim
```

After (fix restored):

```
PASS [0.401s] test_3282_publish_task_resolves_for_both_client_generations
PASS [0.403s] test_3282_failed_publish_returns_an_error_not_a_task
PASS [0.589s] test_3282_import_status_is_repo_scoped
PASS [3.888s] test_3282_import_status_through_visibility_middleware
Summary [3.906s] 4 tests run: 4 passed
```

Timings are 0.4–3.9s, not the ~0.04s signature of a DB test that skipped.

Two further mutations, each isolating one half of the fix:

* **The trap — emit `task`, unregister the two routes:** 3/4 fail with
  `import-status poll of /ansible/{repo}/api/v3/imports/collections/{id}/ must resolve;
  a 404 makes wait_import_task retry forever`. So the route half is load-bearing, and
  the "field alone" shape is caught.
* **Emit a bare task id instead of a root-absolute path:** 3/4 fail with
  ``` `task` must be root-absolute so devel's urljoin(api_server, task) resolves against
  the origin instead of relative to the api root; got "394ec164-..." ```. So the shape
  assertion is load-bearing, not decorative.

Full ansible + galaxy suite on this branch: **37 tests run, 37 passed** — including
every #3137/#3278 test, so `install` is not regressed.

`cargo fmt --all`, `cargo clippy --workspace --all-targets -- -D warnings` clean.
One `sqlx::query!` was added, so `cargo sqlx prepare -- --all-targets` was run and
`SQLX_OFFLINE=true cargo build --lib --tests` verified.

## Follow-up: #3283's premise needs a correction

#3283 records that ansible has "no integration or DTF coverage anywhere". That is true of
`backend/tests/`, but **not** of the DTF repo: `tests/formats/test-ansible.sh` and
`tests/formats/test-ansible-conformance.sh` both exist and both already `POST` to
`/api/v3/artifacts/collections/`.

They did not catch this because they drive the endpoint with `curl` and assert only on
the **HTTP status** — they never read the upload response body, and never poll an import
task. A `202` with a missing `task` key is invisible to them, which is exactly the shape
of this bug.

The cheap, durable guard is a few lines added to `test-ansible-conformance.sh` after its
existing step 1: assert `.task` is present and root-absolute in `upload-resp.json`,
derive the poll URL the way ansible-core does (last non-empty path segment appended to
`${ANSIBLE_BASE}/v3/imports/collections/`), `GET` it, and assert `200` with
`.state != "waiting"` and a non-empty `.finished_at`. No new client dependency — it needs
no `ansible-core` install. I have not opened that MR here because a DTF change must be
validated against a deployed stack and this branch has not been through one; it is
written up on #3283 so it is not silently left uncovered.

Fixes #3282


brandonrc:
## Independent validation — #3288 @ `979e898`

**Verdict: the fix is correct and I could not break it.** I set out to find a cross-repository
or cross-format read on the new `v3/imports/collections/{id}/` route and failed to: the
`(id, repository_id)` predicate is real, and I proved it load-bearing by mutation. The
revert-proof reproduces independently. **What I did find is a factual error in the branch
attribution that is repeated in the shipped CHANGELOG, a self-contradicting claim inside the
test module, and one small scope widening.** None of it blocks; all of it is a text/one-line fix.

Method: fresh shallow clone at `/tmp/ak-val-3288`, `pull/3288/head` pinned at `979e898`,
own Postgres on `:30488` migrated from `backend/migrations/*.sql`, all runs with
`DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` exported (timings 0.27–4.8s, not the ~0.04s
skip signature). ansible-core sources pulled fresh from `raw.githubusercontent.com` for
`stable-2.17`, `stable-2.18`, `stable-2.19`, `stable-2.20`, `devel`. Nothing pushed, no CI triggered.

---

### Claim-by-claim

**1. `wait_import_task` reads exactly `state` / `finished_at` / `messages` / `error` — CONFIRMED.**
Read on all five branches. The bodies are semantically identical (`stable-2.20`/`devel` only
swap the literal `wait = 2` / `* 1.5` for `C.GALAXY_COLLECTION_IMPORT_POLL_INTERVAL` /
`C.GALAXY_COLLECTION_IMPORT_POLL_FACTOR`). `messages` is consumed by a plain
`for message in data.get('messages', []):` — no branch indexes `messages[0]` and no branch
requires a non-empty list, so **`"messages": []` is safe on all five**; the hard
`message['level']` / `message['message']` indexes are inside the loop body and simply never run.
`data['error']` is guarded by `if state == 'failed':`.

**2. The stable/devel split — REFUTED as stated.** `stable-2.20` is on the *devel* side, not the
stable side:

| branch | `publish_collection` returns | caller in `collection/__init__.py` |
|---|---|---|
| `stable-2.17` `api.py:699`, `stable-2.18` `:698`, `stable-2.19` `:692` | `return resp['task']` | `:631` last-non-empty-segment → `wait_import_task(task_id)` |
| **`stable-2.20` `api.py:668`** | **`return urljoin(self.api_server, resp['task'])`** | **`:630` `wait_import_task(import_uri, timeout)` — verbatim, no extraction** |
| `devel` `api.py:678` | `return urljoin(self.api_server, resp['task'])` | `:633` verbatim |

`stable-2.20`'s signature is `def wait_import_task(self, task_url, timeout=0)` and the
`# v2: … / # v3: …` extraction comment is gone from that branch entirely.

**This does not change the fix** — a root-absolute path is still the one spelling both
*derivations* accept, and both derivations are covered. But the attribution is wrong in four
places, one of which ships to users:

- `CHANGELOG.md` (Unreleased → Fixed): "`return resp['task']` on `stable-2.17` through
  `stable-2.20`, `return urljoin(self.api_server, resp['task'])` on `devel`"
- the PR body's table, same wording
- `backend/src/api/handlers/ansible.rs:50` — "`stable-2.17`..`stable-2.20`: the caller …
  reduces the value to its last non-empty path segment"
- `backend/src/api/handlers/ansible.rs:1447` (`stable_line_poll_path` doc) — "The path
  `ansible-core` polls on `stable-2.17`..`stable-2.20`"

Why it matters beyond tidiness: this doc block is the *reasoning* the next maintainer will use.
As written it says "only `devel` needs a root-absolute value" — which invites exactly the
simplification that breaks the newest **released** line. Suggested edit: `stable-2.17`–`stable-2.19`
on one side, `stable-2.20` **and** `devel` on the other, and rename the helper pair to
`segment_extraction_poll_path` / `verbatim_poll_path` so the labels can't rot again.

**2b. "A fully-qualified URL would make the stable line's `_urljoin` emit a doubled path" — REFUTED.**
This appears in the PR body and, more importantly, in the source at
`backend/src/api/handlers/ansible.rs:1547-1549`. It is false, and it contradicts the module doc
20 lines earlier at `ansible.rs:66-68`, which gets it right ("a full path would break the stable
line only if the stable line used it verbatim, which it does not"). The stable line reduces the
value to its last non-empty segment *before* any `_urljoin`, so an absolute URL derives the
identical URL everywhere. Running ansible-core's own `_urljoin` / `urljoin` against
`api_server = https://ak.example/ansible/myrepo/api`:

| emitted `task` | 2.17–2.19 derives | 2.20 / devel derives |
|---|---|---|
| `/ansible/myrepo/api/v3/imports/collections/ID/` (this PR) | `…/api/v3/imports/collections/ID/` | `…/api/v3/imports/collections/ID/` |
| `ID` (bare) | `…/api/v3/imports/collections/ID/` | `https://ak.example/ansible/myrepo/ID` ✗ |
| `https://ak.example/ansible/myrepo/api/v3/imports/collections/ID/` | `…/api/v3/imports/collections/ID/` | `…/api/v3/imports/collections/ID/` |

So the real constraint is "root-absolute **or** absolute; never bare", not "root-absolute only".
The PR's choice is still right and is what ansible-core documents for v3 — the *justification*
just needs correcting. (Worth noting the trade-off the corrected text should record: an absolute
URL would additionally survive a sub-path deployment, where the root-absolute form drops the
prefix under `urljoin`. That is moot here because AK has no external-base-URL config and
`download_url` is already root-absolute and resolved by ansible-core the same way —
`api.py:790-797` — so `install` shares the constraint. Consistent, not a regression.)

**3. All three `--server` spellings converge — CONFIRMED.** `g_connect` rewrites
`self.api_server = n_url`, and `_urljoin(server, '/api/')` yields `…/myrepo/api` (no trailing
slash — the #3278 alias is what makes this land). Derivation run for
`…/myrepo`, `…/myrepo/api`, `…/myrepo/api/`: all three give
`https://ak.example/ansible/myrepo/api/v3/imports/collections/ID/` under **both** derivations.

**4. Synthetic document, no migration — CONFIRMED, and `finished_at` can never be falsy.**
`artifacts.created_at` is `timestamp with time zone NOT NULL DEFAULT now()`; the generated
`backend/.sqlx/query-6a0ab36e….json` records `"nullable": [false, true, false]`, which is why
`row.created_at.to_rfc3339()` compiles without an `unwrap`. The only nullable column read is
`version`, which no client branch touches. **The infinite-poll scenario is not reachable through
a null timestamp.**

**5. Authorization — CONFIRMED, and I added a control the PR does not have.**
- `backend/src/api/routes.rs:94` nests `/ansible` inside `format_routes`; the
  `repo_visibility_middleware` layer is applied to that whole router at `routes.rs:110`. There is
  no second mount point (grepped).
- `extract_repo_key` (`backend/src/api/middleware/auth.rs:1413-1430`) takes path segment 2
  generically, so the new URL resolves its repo key with no per-route allowlist involved.
- The fixture repo really is private: `create_repo` (`test_db_helpers.rs:705`) does not set
  `is_public`, and the column default is `false`. So the anon-401 in
  `test_3282_import_status_through_visibility_middleware` is genuine and self-proving — on a
  public repo that assertion would fail.
- **Extra control I ran (not in the PR):** a *stranger* — a second authenticated user with a valid
  API token, no `role_assignments` row and no `permissions` rule on the private repo — polling the
  real client URL through the real middleware. Result: **`404 Repository not found`**, the
  existence-hiding branch at `auth.rs:2031-2068`. So the PR body's "unreachable without a
  credential that can read the repository" is accurate, not just "any valid credential".
  I'd suggest folding that third control into the test — it is ~15 lines, and that branch's own
  comment records it as closing a red-team cross-tenant private-artifact leak, so it is the one
  worth pinning on a brand-new route.
- Malformed id → `404` (uuid parse before the query), so the route does not confirm well-formedness.

**Cross-repo — REFUTED as a defect, and the guard is load-bearing.** The predicate is real SQL,
not prose (`ansible.rs:823-833`, and the committed `.sqlx` cache shows both `Uuid` parameters).
I neutralised it — `AND (repository_id = $2 OR true)` — and rebuilt against a live database:

```
FAIL [0.390s] test_3282_import_status_is_repo_scoped
  ansible.rs:1682 assertion `left == right` failed:
  an import task minted in another repository must not be an existence oracle under this one
    left: 200   right: 404
```

**6. Revert-proof — CONFIRMED independently.** Merge-base (`e91e970`) production section, cut at
`#[cfg(test)]\nmod tests`, is **26,110 bytes / sha256 `79d10ebadc479eee…`** — exactly the author's
figure. I rebuilt the reverted tree as *merge-base production + the PR's test module verbatim*
(the strict property, not the shortstat heuristic) and ran it:

```
Summary [3.021s] 4 tests run: 0 passed, 4 failed
  FAIL [0.290s] test_3282_publish_task_resolves_for_both_client_generations
  FAIL [0.303s] test_3282_failed_publish_returns_an_error_not_a_task
  FAIL [0.392s] test_3282_import_status_is_repo_scoped
  FAIL [3.003s] test_3282_import_status_through_visibility_middleware
  panicked at ansible.rs:1464: task
```

My reconstruction shows `450 insertions(+), 0 deletions(-)` where the body says 439 — an
immaterial boundary difference; the byte-identity of the production section is the property that
matters and it holds exactly.

I also reproduced the **route-unregistered trap** rather than taking it on trust, because the
plain revert fails on the missing `task` key first and so does *not* independently exercise the
route half:

```
Summary [2.947s] 4 tests run: 1 passed, 3 failed
  import-status poll of /ansible/{repo}/api/v3/imports/collections/{id}/ must resolve;
  a 404 makes wait_import_task retry forever (--import-timeout defaults to 0)
```

Both halves are independently load-bearing, as claimed.

Baseline: `cargo nextest run --lib ansible` on the PR head = **35 run, 35 passed**, timings up to 3.9s.

---

### Defects

**D1 — MEDIUM (ships in the CHANGELOG): `stable-2.20` is attributed to the wrong client generation.**
Four locations, listed under claim 2. Failure scenario: not a runtime failure today; it is a
documentation trap. The next person to touch this reads "only `devel` needs the root-absolute
form", trims `task` to a bare id for the four branches the comment calls "stable", and silently
reintroduces the indefinite publish hang on 2.20 — the exact bug this PR exists to prevent, on
the newest released line. Fix: correct the split to 2.17–2.19 vs 2.20+devel in `CHANGELOG.md`,
the PR body, `ansible.rs:50` and `ansible.rs:1447`.

**D2 — LOW: a false claim inside the test module.** `ansible.rs:1547-1549` says a fully-qualified
URL "makes the stable line's `_urljoin` produce a doubled path". It does not (table above), and
`ansible.rs:66-68` says the opposite in the same file. Fix: delete the doubled-path clause and
state the real invariant — *the value must not be bare; root-absolute is chosen because it is the
shape ansible-core documents for v3.*

**D3 — LOW: the route resolves any artifact row in the repository, not just collections.**
`ansible.rs:824-830` constrains `id`, `repository_id`, `is_deleted` — nothing ties the row to an
Ansible collection. Demonstrated: I inserted a non-collection row (`name = internalonlyname`,
`version = 9.9.9-secret`, no hyphen) into an Ansible repository and asked the import-status route
for its id:

```
list_collections    -> 200 {"data":[], "meta":{"count":0}}          <- deliberately filtered out
import_status       -> 200 {"state":"completed","namespace":"","name":"internalonlyname",
                            "version":"9.9.9-secret","created_at":"2026-08-11T15:41:52…"}
```

`list_collections` drops rows whose `name` has no hyphen (`ansible.rs:276`); the new route does
not, so generic-upload / migration-imported rows in an Ansible repository become readable by id
with their coordinates and creation time, and are served as a Galaxy import document with an
empty `namespace`. **Not a boundary crossing** — the caller already holds read on that repository,
and v4 ids are not guessable — so this is a scope-widening nit, not a vulnerability. Fix, if you
want it tight: require the row to carry `artifact_metadata` of format `ansible` (or at minimum
require `name` to split into `namespace-name`, matching `list_collections`), and 404 otherwise.
That also keeps the served document from ever claiming a nameless namespace.

**D4 — LOW: the second registered spelling is untested.** `ansible.rs:147` registers
`…/collections/:task_id` without the trailing slash and the comment justifies it, but no test
requests it. I verified it works (`200 OK`). One `assert_eq!` on
`task.trim_end_matches('/')` inside the existing publish test covers it; otherwise a future
router edit can drop the alias with the suite still green.

**D5 — LOW: the middleware test's fail-closed set is missing the stranger case.** anon-401 and
bogus-401 are there; the authenticated-non-member case is not. I ran it and it passes (404) — see
claim 5. Worth adding for the reason given there.

---

### Credit where it is due

Two things here are well above the bar. Verifying the URL derivation against ansible-core's own
`_urljoin` / `urljoin` rather than eyeballing it is what makes the root-absolute choice defensible
at all — and the `stable_line_poll_path` / `devel_poll_path` helpers reimplementing the client's
derivation, instead of comparing the response to the constant the handler emitted, is the
difference between a test and an echo. Second, spotting that the naive fix (add the field, skip
the route) converts a crash into an **indefinite** hang — because `wait_import_task` treats 404 as
"not started yet", `--import-timeout` defaults to `0` and `--no-wait` is read one frame too late —
and then *mutation-testing that specific trap* rather than asserting it in prose, is the reasoning
that makes this PR safe to land. Both replicated exactly as described. The `--no-wait` analysis
holds too: on the 2.17–2.19 line the task-id extraction is inside the `if wait:` block, so the
flag genuinely cannot reach the dereference.

The follow-up note on #3283 is also correct: `tests/formats/test-ansible.sh` and
`test-ansible-conformance.sh` do exist in the DTF repo, so #3283's "no coverage anywhere" premise
does need the correction the body gives it.

**Recommendation:** land after the D1/D2 text corrections (CHANGELOG + the two comment blocks).
D3–D5 are fine as a follow-up issue.

Issue #2218, opened 2026-07-04
Per-repo storage_used_bytes excludes proxy-cached storage


Dreamacro:
### Description

#1548 was closed by #1567, but #1567 only fixed the **artifact listing** for remote repos (reconstructed from the storage backend). The per-repository `storage_used_bytes` field called out in #1548 is still broken: it's computed exclusively from the `artifacts` table, which has not included proxy-cached items since #1280 stopped inserting them.

### Expected Behavior

Per-repo `storage_used_bytes` in `GET /api/v1/repositories` should reflect bytes cached under `proxy-cache/<repo_key>/` for remote/proxy repositories, without re-introducing the #1278 doubled-prefix filesystem bug.

### Observed Behavior

For a remote/proxy repo, `artifacts` has zero rows, so `storage_used_bytes` reads `0` regardless of how many bytes are actually cached on disk. The proxy cache sidecar (`__cache_meta__.json`) already records `size_bytes` per cached entry (`CacheMetadata.size_bytes`, `backend/src/services/proxy_service.rs:3284`), so the data exists — it's just not wired into the accounting query.

The accounting query at `backend/src/api/handlers/repositories.rs:1238`:

```sql
SELECT repository_id, COALESCE(SUM(size_bytes), 0)::BIGINT
FROM artifacts
WHERE repository_id = ANY($1) AND is_deleted = false
GROUP BY repository_id
```

### Environment Information

- Deployment type: All (Docker, bare metal, etc.)
- Storage backend: All (filesystem, S3, GCS, Azure)
- Backend Version: Any version including #1280

### Steps to Reproduce

1. Create a remote repo (e.g. PyPI proxy to https://pypi.org)
2. Pull a package through the proxy cache (bytes land under `proxy-cache/<repo_key>/...`)
3. `GET /api/v1/repositories` → that repo's `storage_used_bytes` is `0`
4. `SELECT SUM(size_bytes) FROM artifacts WHERE repository_id = <that repo> AND is_deleted = false` returns `0` (or NULL)

PR #3289, opened 2026-08-11
fix(proxy): bound the lazy proxy-cache catalog backfill spawned per cache hit

### LINKED ISSUE #3298: Proxy cache-hit catalog backfill spawns unbounded background DB writers, contributing to "Server overloaded" 503s under load
### Description

Every streaming proxy cache **HIT** (npm, PyPI, Maven, OCI, or any other
proxy-capable format) unconditionally spawns a detached `tokio::spawn` task
(`ProxyService::spawn_lazy_catalog_backfill`) to lazily self-heal the
`proxy_cache_artifacts` catalog table for objects that were cached before
that table existed (#2218/#2270 back-compat).

This spawn has no concurrency bound at all, and each spawned task:

1. Re-reads the sidecar metadata object from storage a second time — the
   caller already loaded the identical `CacheMetadata` moments earlier to
   evaluate cache freshness, so this is a redundant read on every single hit.
2. Acquires a connection from the **same** `PgPool` that serves normal
   request-handling queries (`DATABASE_MAX_CONNECTIONS`, default 50) to run
   an upsert.

Under a burst of concurrent cache hits — e.g. a popular package suddenly
getting hit hard through a proxy repo — this can spawn a correspondingly
large number of background tasks that compete with real request traffic for
the same fixed DB connection budget. That contention appears able to starve
the pool enough that in-flight requests are held longer, which in turn pushes
concurrent-request count up against the app's global concurrency backstop
(`tower::load_shed` + `GlobalConcurrencyLimitLayer`, `GLOBAL_MAX_CONCURRENCY`,
default 512), tripping it and surfacing to clients as `SERVICE_UNAVAILABLE`.

### Environment Information

- Deployment type: Kubernetes / Helm
- Operating System: Linux
- Architecture: AMD
- Database: Postgresql
- Reverse Proxy: AWS ALB
- Network setup: AWS ALB -> Target Group Binding -> Artifact Keeper pods

### Backend Version

1.7.1

### Web UI Version

N/A

### Package Format

N/A / General

### Expected Behavior

A cache-hit's best-effort catalog backfill should never be able to starve the
shared DB connection pool. Since the backfill is explicitly best-effort and
safe to skip (it's `ON CONFLICT DO NOTHING` on body fields — a missed
backfill just gets retried on the next hit), it should degrade gracefully
under load rather than piling up unbounded background work.

### Observed Behavior

Under sustained high proxy cache-hit throughput, the backend intermittently
returns:

"Service unavailable: Server overloaded, please retry: service overloaded"
"SERVICE_UNAVAILABLE"

from `backend/src/error.rs:303`, on the `tokio-rt-worker` thread pool, with
no corresponding user-facing error on the request that actually trips it —
consistent with background tasks (not the foreground request itself)
consuming shared capacity.

### Steps to Reproduce

1. Configure a proxy/remote repository (e.g. npm) against a real upstream.
2. Warm the cache for one artifact (one request so it's cached).
3. Send a large burst of concurrent GET requests for that same cached
   artifact (all of which will be streaming cache **hits**), e.g.
   `./scripts/stress/run-concurrent-uploads.sh`-style concurrency, adjusted
   to hit the same cached proxy path repeatedly.
4. Observe DB pool saturation (`POOL_EXHAUSTED`) and/or the global backstop's
   `SERVICE_UNAVAILABLE` ("Server overloaded, please retry") under
   sufficient concurrency, correlated with cache-hit volume rather than
   genuine request-handling load.

### Logs

```shell
Service unavailable: Server overloaded, please retry: service overloaded
SERVICE_UNAVAILABLE
backend/src/error.rs:303
thread.name: tokio-rt-worker
```

### Additional Context

- Relevant code: `backend/src/services/proxy_service.rs`,
  `ProxyService::try_streaming_cache_hit` (calls
  `spawn_lazy_catalog_backfill` unconditionally on every hit) and
  `spawn_lazy_catalog_backfill` itself (the unbounded `tokio::spawn`).
- The catalog write itself is `proxy_catalog::backfill_from_sidecar` in
  `backend/src/services/proxy_catalog.rs`.
- The DB pool is shared: `ProxyService`'s `db: PgPool` is a clone of the same
  pool used by `AppState.db` for all normal request handlers
  (`backend/src/main.rs`).
- There are two distinct 503 sources in this codebase worth not conflating
  when investigating further: the global router-wide concurrency/load-shed
  backstop (`backend/src/api/routes.rs`, `apply_global_backstop`) vs. a
  `sqlx` pool `.acquire()` timeout (`AppError::is_pool_timeout`,
  `backend/src/error.rs`, `POOL_EXHAUSTED`). Both are plausibly downstream of
  the same root cause described here.
- A fix bounding the spawn with a semaphore and removing the redundant
  storage read is proposed in a companion PR.

### Pre-submission Checklist

AppliNH:
fixes #3298

## Summary

Every streaming proxy cache **HIT** unconditionally spawns a detached
`tokio::spawn` task (`spawn_lazy_catalog_backfill`) to self-heal the
`proxy_cache_artifacts` catalog for objects that were cached before that
table existed (#2218/#2270 back-compat). Under load this pattern has two
problems:

1. **Unbounded spawn.** There was no cap on how many of these tasks could be
   in flight at once. A burst of concurrent cache hits — e.g. a popular
   package suddenly getting hammered through a proxy repo — spawns one task
   per hit, and each task acquires a connection from the **same** `PgPool`
   that serves normal request-handling queries. That pool defaults to 50
   connections; this background bookkeeping was competing directly with real
   traffic for that same fixed budget.
2. **Redundant storage read.** Each spawned task re-read the sidecar metadata
   object from storage a second time, even though the caller
   (`read_cached_with_revalidation_streaming`) had already loaded the exact
   same `CacheMetadata` moments earlier to evaluate freshness. Every cache hit
   was paying for two storage reads where one sufficed.

Under sustained high cache-hit throughput this background load can starve the
shared DB pool, contributing to requests piling up against the app's global
concurrency backstop (`tower::load_shed` + `GlobalConcurrencyLimitLayer`,
`GLOBAL_MAX_CONCURRENCY`, default 512) and surfacing to clients as
`"Server overloaded, please retry"` (`SERVICE_UNAVAILABLE`).

### What changed

- Added a bounded `tokio::sync::Semaphore` (`backfill_limiter`, capacity
  `MAX_CONCURRENT_CATALOG_BACKFILLS = 8`) on `ProxyService`.
  `spawn_lazy_catalog_backfill` now does a synchronous `try_acquire_owned`
  before spawning. If the limiter is saturated, the backfill for that hit is
  skipped (logged at `debug`) instead of piling up another task.
  - This is safe to skip: the backfill is strictly best-effort self-healing
    (`ON CONFLICT DO NOTHING` on the body fields — see
    `proxy_catalog::backfill_from_sidecar`'s doc comment). Skipping it once
    just means the next hit on the same object retries.
- Threaded the already-loaded `CacheMetadata` through
  `StreamingCacheReadOutcome::Hit` (boxed, per `clippy::large_enum_variant`)
  instead of having the spawned task re-read the sidecar from storage,
  removing one redundant storage round trip per cache hit.

### Why this bound and not a queue

The existing design already treats this as strictly best-effort
(fire-and-forget, errors swallowed at `debug`). A bounded semaphore that
*drops* work under saturation matches that contract — queuing instead would
just delay the same pool pressure rather than relieving it, and would add
complexity for a self-healing path that is explicitly allowed to be lossy.

Added two unit tests in `backend/src/services/proxy_service.rs`:
- `test_lazy_catalog_backfill_permit_is_acquired_then_released` — asserts a
  permit is claimed synchronously when spawning a backfill and released once
  the task completes.
- `test_lazy_catalog_backfill_is_skipped_when_limiter_saturated` — saturates
  the limiter first, then asserts a further call does not spawn (permit count
  is unaffected, no panic).

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/services/proxy_service.rs
+++ b/backend/src/services/proxy_service.rs
@@ -20,7 +20,7 @@ use reqwest::header::{
 use reqwest::{Client, StatusCode};
 use serde::{Deserialize, Serialize};
 use sqlx::PgPool;
-use tokio::sync::RwLock;
+use tokio::sync::{RwLock, Semaphore};
 use uuid::Uuid;
 
 use crate::error::{AppError, Result};
@@ -869,8 +869,11 @@ enum CacheReadOutcome {
 /// three actions are otherwise identical to the buffered path: serve a body,
 /// return a cached 404, or fetch upstream.
 enum StreamingCacheReadOutcome {
-    /// Serve this cached (or freshly revalidated) streamed body.
-    Hit(StreamingFetchResult),
+    /// Serve this cached (or freshly revalidated) streamed body. Carries the
+    /// sidecar [`CacheMetadata`] that was already loaded to reach this hit, so
+    /// the caller's lazy catalog backfill (#2218/#2270) can reuse it instead
+    /// of re-reading the same sidecar object from storage a second time.
+    Hit(StreamingFetchResult, Box<CacheMetadata>),
     /// A negative-cached upstream 404 is still within its TTL: respond 404.
     NegativeHit,
     /// No usable entry; the caller must elect a streaming leader / fetch
@@ -2629,6 +2632,14 @@ pub struct ProxyService {
     /// //               replaces this field with a wrapping `Coordinator` impl,
     /// //               selected by config via the [`HydrationCoordinator`] enum.
     coordinator: HydrationCoordinator,
+    /// Bounds how many [`Self::spawn_lazy_catalog_backfill`] tasks may be
+    /// in flight at once. Those tasks are opportunistic (#2218/#2270
+    /// self-healing) but draw a connection from the same shared `db` pool
+    /// as real request handlers; left unbounded, a burst of concurrent
+    /// cache hits could spawn enough of them to starve that pool under
+    /// load. Saturating this limiter simply skips the backfill for that
+    /// hit rather than queuing — never blocks the serve.
+    backfill_limiter: Arc<Semaphore>,
 }
 
 impl ProxyService {
@@ -2642,6 +2653,12 @@ impl ProxyService {
     /// denominator so a switch from filesystem to S3 cannot turn an
     /// existing repo into a broken one (#1044).
     const MAX_STORAGE_KEY_BYTES: usize = 1024;
+
+    /// Max concurrent lazy catalog-backfill tasks (see `backfill_limiter`).
+    /// Small relative to `DATABASE_MAX_CONNECTIONS` (default 50) so this
+    /// best-effort background work can never claim more than a slice of the
+    /// pool away from real request handlers, even at full saturation.
+    const MAX_CONCURRENT_CATALOG_BACKFILLS: usize = 8;
 
     /// Create a new proxy service
     pub fn new(db: PgPool, storage: Arc<StorageService>) -> Self {
@@ -2661,6 +2678,7 @@ impl ProxyService {
         // with `PROXY_SINGLEFLIGHT_ADVISORY_LOCKS_ENABLED=true`. Call sites are
         // unchanged — both variants implement the same `Coordinator` seam.
         let coordinator = HydrationCoordinator::from_env(db.clone());
+        let backfill_limiter = Arc::new(Semaphore::new(Self::MAX_CONCURRENT_CATALOG_BACKFILLS));
 
         Self {
             db,
@@ -2669,6 +2687,7 @@ impl ProxyService {
             cache_persister,
             upstream_client,
             coordinator,
+            backfill_limiter,
         }
     }
 
@@ -3470,15 +3489,24 @@ impl ProxyService {
             )
             .await?
         {
-            StreamingCacheReadOutcome::Hit(result) => {
+            StreamingCacheReadOutcome::Hit(result, metadata) => {
                 // #2218/#2270 back-compat: a cache hit on an object cached
                 // BEFORE this catalog existed has no `proxy_cache_artifacts`
                 // row. Fire-and-forget a best-effort backfill from the sidecar
                 // so pre-upgrade objects self-heal into the catalog on first
                 // access (also bumps `last_accessed_at`). Never blocks the
                 // serve; `ON CONFLICT DO NOTHING` leaves authoritative rows
-                // untouched.
-                self.spawn_lazy_catalog_backfill(repo.id, cache_path, cache_key, metadata_key);
+                // untouched. Bounded by `backfill_limiter` (#3xxx overload
+                // fix): under a cache-hit burst this is skipped rather than
+                // spawning unbounded tasks that starve the shared DB pool —
+                // it is strictly best-effort self-healing, safe to drop.
+                self.spawn_lazy_catalog_backfill(
+                    repo.id,
+                    cache_path,
+                    cache_key,
+                    metadata_key,
+                    *metadata,
+                );
                 Ok(Some(result))
             }
             StreamingCacheReadOutcome::NegativeHit => Err(AppError::NotFound(format!(
@@ -3491,31 +3519,44 @@ impl ProxyService {
 
     /// Fire-and-forget lazy backfill of the persisted proxy-cache catalog from a
     /// cache-hit's sidecar (#2218/#2270 back-compat). Spawned so it never blocks
-    /// the serve; reads the metadata sidecar for the true size/checksum and
-    /// upserts `ON CONFLICT DO NOTHING` (existing authoritative rows only get
-    /// their `last_accessed_at` bumped). Any failure is logged at `debug` and
-    /// swallowed — a warm-hit backfill is strictly best-effort.
+    /// the serve; upserts `ON CONFLICT DO NOTHING` (existing authoritative rows
+    /// only get their `last_accessed_at` bumped). Any failure is logged at
+    /// `debug` and swallowed — a warm-hit backfill is strictly best-effort.
+    ///
+    /// Takes the [`CacheMetadata`] the caller already loaded to determine the
+    /// hit, instead of re-reading the same sidecar object from storage a
+    /// second time — every cache hit was paying for two storage reads before
+    /// this (one to evaluate freshness, one here).
+    ///
+    /// Bounded by `backfill_limiter`: a burst of concurrent cache hits used to
+    /// spawn an unbounded number of these tasks, each contending with normal
+    /// request handlers for a connection out of the *same* shared DB pool —
+    /// under sustained load this starved the pool for real traffic and fed
+    /// into the global concurrency backstop tripping "Server overloaded". This
+    /// is opportunistic self-healing, not correctness-critical, so it is
+    /// simply skipped (not queued) once the limiter is saturated; the next
+    /// cache hit on the same object will retry.
     fn spawn_lazy_catalog_backfill(
         &self,
         repository_id: Uuid,
         path: &str,
         cache_key: &str,
         metadata_key: &str,
+        metadata: CacheMetadata,
     ) {
+        let Ok(permit) = Arc::clone(&self.backfill_limiter).try_acquire_owned() else {
+            tracing::debug!(
+                cache_key = %cache_key,
+                "skipping proxy-cache catalog backfill: limiter saturated under load"
+            );
+            return;
+        };
         let db = self.db.clone();
-        let storage = Arc::clone(&self.storage);
         let path = path.to_string();
         let cache_key = cache_key.to_string();
         let metadata_key = metadata_key.to_string();
         tokio::spawn(async move {
-            let data = match storage.get(&metadata_key).await {
-                Ok(data) => data,
-                Err(_) => return, // sidecar gone -> nothing to backfill from
-            };
-            let metadata: CacheMetadata = match serde_json::from_slice(&data) {
-                Ok(m) => m,
-                Err(_) => return,
-            };
+            let _permit = permit; // held for the task's lifetime, released on drop
             if let Err(e) = proxy_catalog::backfill_from_sidecar(
                 &db,
                 repository_id,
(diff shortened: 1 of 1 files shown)


brandonrc:
## Review — verdict: sound fix, two small changes requested before merge

Thanks for this, @AppliNH — and thanks especially for filing #3298 with a real reproduction shape, the two distinct 503 sources called out separately, and the code pointers. That issue did most of the work of reviewing this PR, which is not the norm for an external report.

I reviewed this against head `752c5c0` in a fresh clone (no CI triggered), read the base for comparison, and ran your two new tests locally.

**The premise checks out.** On `main`, `backend/src/services/proxy_service.rs:3481` calls `spawn_lazy_catalog_backfill` unconditionally inside the `StreamingCacheReadOutcome::Hit` arm, and there is no semaphore, dedup set, or once-flag anywhere in the path. Because the task is detached, in-flight backfills genuinely can exceed the request-concurrency limit — the request returns as soon as it is spawned, so the next request's task stacks on top of the previous one's still queued on `db.acquire()`. That is the part that makes it unbounded rather than merely "one per hit".

**Things you got right that are easy to get wrong:**

- **The limiter is process-wide.** `ProxyService` is constructed exactly once and shared behind an `Arc` (`backend/src/main.rs:743`), so `backfill_limiter` is a single semaphore for the whole process. Had the service been built per-request, an `Arc<Semaphore>` field would have looked identical in the diff and bounded nothing. I checked all non-test `ProxyService::new` call sites — only that one.
- **It skips, it never blocks.** `try_acquire_owned` (`proxy_service.rs:3547`) is non-blocking and runs on the request task, so a saturated limiter costs a cache hit one atomic op and nothing else. Latency and body on the hit path are unchanged.
- **No permit leak.** The permit is moved into the spawned future and dropped when it completes; there is no early-return path between the acquire and the `tokio::spawn`, and if the runtime drops the task the future (and permit) go with it.
- **The metadata threading is behaviour-preserving.** I specifically checked the `Stale` → `ServeRevalidated` arm, where the sidecar is rewritten between the load and the backfill: `extend_cache_expiry` (`proxy_service.rs:5390-5397`) clones the metadata and mutates only `expires_at`, and `backfill_from_sidecar` reads only `size_bytes` / `checksum_sha256` / `content_type`. So the in-memory copy you thread through is equivalent to the re-read you removed. Nothing on the MISS path changed, and `Hit` has only the two construction sites you updated.
- **`Box`ing the variant** for `clippy::large_enum_variant`, and writing `fixes #3298` as plain text — the linked-issue workflow strips inline code spans, so a backticked keyword silently fails that gate. Both details people miss routinely.

---

### 1. Requested change: the two new tests do not gate the bound (proven)

This is the one thing I'd like fixed before merge. I ran both tests on your branch — green, 0.09s / 0.11s. Then I mutated only the guard so the limiter becomes completely inert (acquired when available, but the backfill spawns unconditionally either way):

```rust
// mutation: the bound no longer bounds anything
let permit = Arc::clone(&self.backfill_limiter).try_acquire_owned().ok();
```

Both tests still pass:

```
PASS [0.055s] test_lazy_catalog_backfill_permit_is_acquired_then_released
PASS [0.081s] test_lazy_catalog_backfill_is_skipped_when_limiter_saturated
Summary [0.100s] 2 tests run: 2 passed
```

The reason is in the saturated test's own assertion. It asserts `available_permits() == 0` after the call — but the buggy implementation never touches the semaphore in that state either, so `0` is exactly what the bug produces. The assertion is satisfied by the thing it is meant to catch. (`test_..._acquired_then_released` has the same issue from the other side: acquiring a permit and spawning anyway still moves the count down and back up.) Neither test can observe whether a task was actually spawned, so neither can distinguish "skipped" from "spawned regardless".

**Concrete fix — I built and validated this, it's about six lines.** Have `spawn_lazy_catalog_backfill` return whether it spawned:

```rust
    ) -> bool {
        let Ok(permit) = Arc::clone(&self.backfill_limiter).try_acquire_owned() else {
            tracing::debug!(/* … unchanged … */);
            return false;
        };
        // … unchanged …
        tokio::spawn(async move { /* … unchanged … */ });
        true
    }
```

Call site becomes `let _ = self.spawn_lazy_catalog_backfill(…);` (or feed it a metric — a counter of skipped backfills would be genuinely useful for anyone diagnosing this in production). Then assert on the return value in both tests:

```rust
// unsaturated
assert!(service.spawn_lazy_catalog_backfill(…), "an unsaturated limiter must actually spawn the backfill");
// saturated
assert!(!service.spawn_lazy_catalog_backfill(…), "a saturated limiter must SKIP the spawn, not spawn anyway");
```

Keep your existing permit-count assertions as-is; they're still worth having. With that change I get, on your code:

```
Summary [0.108s] 2 tests run: 2 passed
```

and with the same inert-bound mutation applied:

```
thread '…test_lazy_catalog_backfill_is_skipped_when_limiter_saturated' panicked at
backend/src/services/proxy_service.rs:8124:9:
a saturated limiter must SKIP the spawn, not spawn anyway
Summary [0.078s] 2 tests run: 1 passed, 1 failed
```

That is the property the PR is claiming, and it now fails when the claim is false.

### 2. Requested change: `#3xxx` placeholder

`proxy_service.rs:3499` — the inline comment reads `Bounded by backfill_limiter (#3xxx overload fix)`. Looks like a placeholder that outlived the issue being filed; should be `#3298`.

---

### 3. Worth knowing: "the next hit retries" is weaker than the PR body claims

Not a blocker, and it is **pre-existing** rather than introduced here, but it undercuts the stated safety rationale so it's worth being accurate about.

`backfill_from_sidecar`'s conflict arm (`backend/src/services/proxy_catalog.rs:112-113`) is:

```sql
ON CONFLICT (repository_id, path) DO UPDATE SET
    last_accessed_at = now()
```

It only repairs a row that does not yet exist. Meanwhile, on the very same request, `proxy_helpers::record_proxy_download` is awaited inline immediately after the streaming serve (`backend/src/api/handlers/proxy_helpers.rs:4467`), and it inserts a **placeholder** catalog row with `size_bytes = 0` and `checksum NULL` (`proxy_catalog.rs:360-366`) for the same `(repository_id, path)`. The comment there says the tee later refines that placeholder in place — but the tee only runs on a cache MISS, and this whole path is a HIT.

So once the placeholder lands, `backfill_from_sidecar` can never fill in the real size/checksum; every later hit just bumps `last_accessed_at`. The practical effect is a legacy pre-catalog object stuck at `size_bytes = 0` in `sum_by_repo` accounting and in the remote-repo browse listing (`repositories.rs:5217-5223` documents exactly that rendering).

Two consequences for this PR:

- **Your change makes this better, not worse, in the common case.** Removing the storage read means the backfill task now races straight to the DB instead of doing an object-store round trip first, so it wins the race against `record_proxy_download` far more often than it did on `main`. On `main` the backfill was arguably losing that race almost every time, which means the #2218 self-healing has been largely inert in practice.
- **In the saturated case, the skip is deterministic** — the placeholder wins for certain, and that object's real size never lands. That's a narrow, transient-burst-only window, and I don't think it should block this PR, but I'd soften "skipping it once just means the next hit on the same object retries" in the body and the doc comment to something like "the next hit retries, unless a placeholder row has since been created by the download recorder — see #NNNN".

The real fix is in the conflict arm, and belongs in its own PR:

```sql
ON CONFLICT (repository_id, path) DO UPDATE SET
    last_accessed_at = now(),
    size_bytes       = CASE WHEN proxy_cache_artifacts.size_bytes = 0
                            THEN EXCLUDED.size_bytes ELSE proxy_cache_artifacts.size_bytes END,
    checksum_sha256  = COALESCE(proxy_cache_artifacts.checksum_sha256, EXCLUDED.checksum_sha256)
```

Happy to file that as a separate issue if you'd rather not grow this PR — it's a clean, self-contained follow-up and it's a nicer first fix than it looks.

### 4. Minor: the bound is a constant, the pool is configurable

`MAX_CONCURRENT_CATALOG_BACKFILLS = 8` (`proxy_service.rs:2661`) and the doc comment reasons about it as "small relative to `DATABASE_MAX_CONNECTIONS` (default 50)". That default is set in `backend/src/config.rs:1216` and is very much tunable — an operator running a small pod with `DATABASE_MAX_CONNECTIONS=5` (exactly the deployment most vulnerable to what #3298 describes) gets a background bound larger than their entire pool. Deriving it (`max(1, pool/8)`) or at least noting the assumption in the comment would close that. Fine as a follow-up.

### 5. Minor: the removed sidecar read was also an incidental existence check

Dropping the in-task `storage.get(metadata_key)` also drops a late "is this entry still there?" check. `invalidate_cache` deletes the catalog row (`proxy_service.rs:4089`) *before* evicting the blob, so an in-flight backfill that read its metadata pre-purge can re-insert an orphan row for an object that no longer exists — inflating storage accounting until the next reconcile. This race exists on `main` too; you've widened the window somewhat, and the cheap mitigation is on the other side (move the `delete_by_key` to after `cache_store.invalidate`). Worth an issue, not worth holding this PR.

### 6. Scope note on #3298 itself

I'd expect this to reduce the pressure but not necessarily to close the reported symptom on its own. `record_proxy_download` runs one unconditional `INSERT` into `proxy_download_statistics` per proxy serve against the same pool, in the foreground, on every hit and miss. Under the burst you describe that is a larger and more constant source of pool contention than the backfill spawns. Suggest keeping #3298 open (or opening a sibling) after this merges, rather than auto-closing and assuming it's done — otherwise the next person to see the 503s will start from scratch.

---

### Recommendation

Approve and merge once (1) and (2) are addressed. (1) is the only one I'd actually hold for: the fix itself is correct, but as it stands the repo would carry a bound that nothing prevents a future refactor from quietly making inert, which is the specific failure mode the regression-test requirement exists to prevent. (3)–(6) are follow-ups, and (3) is a pre-existing bug you happened to surface rather than anything you introduced.

Nice piece of work — the diagnosis, the choice of drop-over-queue for a best-effort path, and the incidental removal of a per-hit storage read are all the right calls.

Issue #3277, opened 2026-08-10
Flaky: proxy-cache stats test asserts a cluster-wide empty table, races concurrent inserts


brandonrc:
Same class as #3129 (fixed by #3242), in a different file. Currently failing unrelated PRs — it took down `Backend Unit Tests` on #3274, which does not touch this code.

## The race

`api::handlers::admin::tests::test_proxy_cache_stats_query_empty_table_coalesces_to_zero_db` (`backend/src/api/handlers/admin.rs:1881`, added by #3088):

```rust
let mut tx = pool.begin().await.expect("begin transaction");
sqlx::query("DELETE FROM proxy_cache_artifacts").execute(&mut *tx).await…;
let (count, size): (i64, i64) = sqlx::query_as(
    "SELECT COUNT(*), COALESCE(SUM(size_bytes), 0)::BIGINT FROM proxy_cache_artifacts",
).fetch_one(&mut *tx).await…;
tx.rollback().await…;
assert_eq!(count, 0);
```

The transaction wrapper looks like isolation but is not. Postgres defaults to READ COMMITTED, so each statement sees rows committed by *other* transactions since the statement began. The suite runs process-per-test against a SHARED database, so:

1. this test `DELETE`s every row visible to it;
2. a concurrent test inserts a `proxy_cache_artifacts` row and commits;
3. the `SELECT` sees it → `count > 0` → `assertion left == right failed` at `admin.rs:1901`.

The rollback correctly avoids destroying other tests' data, but it does nothing for the read.

## Observed

```
test api::handlers::admin::tests::test_proxy_cache_stats_query_empty_table_coalesces_to_zero_db ... FAILED
thread ... panicked at backend/src/api/handlers/admin.rs:1901:9:
assertion `left == right` failed
```

Full 7-minute run, single test failed — not the build-stage flake fixed by #3248.

## What the test is actually for

It pins that `COALESCE(SUM(size_bytes), 0)` maps `SUM`'s `NULL` to `0` rather than erroring or returning `NULL` on an empty set. That property is worth keeping — it just does not need a globally empty table to demonstrate.

## Suggested fix

Do not assert a cluster-wide count. Options in preference order:

1. **Scope the aggregate** to rows this test controls — filter on a UUID-derived key prefix or `repository_id`, seed nothing, and assert the `COALESCE` behaviour over that empty *subset*. Same property, no global assumption. This is what #3242 did for the storage-GC footprint test.
[...]

PR #3293, opened 2026-08-12
fix(scanner): classify Negligible and UNKNOWN scanner severities deliberately (#3243 stage 1)

### LINKED ISSUE #3294: OpenSCAP downgrades a critical finding to info, and Negligible reaches the floor by parse miss (#3243 stage 1a: mapping consolidation)
> **Scope narrowed 2026-08-13.** This issue was originally written as the whole
> of #3243 stage 1 — consolidate the three adapter fallbacks **and** move
> unrecognised tokens off the floor. The second half turned out to be a
> high-volume, gate-flipping behaviour change: `PolicyService::evaluate_artifact`
> is the hosted-artifact **download** gate (via
> `quarantine_service::enforce_download_gate`), `info` violates no policy at any
> threshold while `medium` violates at `medium` and `low`, and Artifact Keeper's
> own OpenSCAP wrapper defaults a rule's severity to the literal `unknown` — so
> the remap can begin returning `403` on downloads that succeed today, and can
> drop a repository from grade A to F on the next scan. That is not a patch, so
> stage 1 has been split:
>
> * **stage 1a — this issue**, milestone 1.7.2: the OpenSCAP `critical` arm,
>   `Negligible` as a recognised token, and the classifier consolidation. No
>   other token's classification changes on any adapter.
> * **stage 1b — #3306**, milestone 1.8.0: unrecognised / `UNKNOWN` tokens move
>   off the floor, with the operator upgrade note.
>
> #3243 remains open for stage 2 (option F) and stage 3 (option B1).

Split out of #3243. This is the half that is safe to ship in a patch because it
is either a near-zero-volume fail-open or behaviour-preserving.

## Defect 1 — OpenSCAP downgrades `critical` to `info` (the behaviour change)

`backend/src/services/openscap_scanner.rs::convert_findings` classified
severities with its own private match:

```rust
let severity = match f.severity.to_lowercase().as_str() {
    "high" => Severity::High,
    "medium" | "moderate" => Severity::Medium,
    "low" => Severity::Low,
    _ => Severity::Info,
};
```

There is **no `critical` arm**, so content reporting a `critical` rule failure
had its *most* severe result filed as its *least* severe one. `Info` is inert
everywhere it matters: `penalty_weight()` is `0`, it appears in none of
`policy_service`'s severity arrays at any threshold, and no gate's projection
selects `info_count`. So a `critical` OpenSCAP finding contributed to nothing —
no policy violation, no promotion block, no score penalty.

Volume is near zero, which is what makes this patch-safe: XCCDF 1.2 (NISTIR
7275r4 §6.4.4.2) defines the Rule `severity` attribute as
`unknown | info | low | medium | high`, so `critical` is not in the spec's
vocabulary and only non-standard content or a non-default wrapper produces it.
Near-zero volume, unambiguous fail-open when it fires.

## Defect 2 — `Negligible` reaches the floor by parse miss (behaviour-preserving)

`Severity::from_str_loose` recognises only
`critical|high|medium|moderate|low|info|informational|none`, so Grype's and
Harbor's `Negligible` — the Debian and Ubuntu security trackers' explicit
"triaged, not worth fixing" grade — landed in `Info` by falling off the end of a
match. The lowest bucket is the *right* destination (Harbor documents
`negligible` as never blocking a pull), but nothing in the code recorded that it
was a decision rather than a miss, so it was indistinguishable from an ungraded
finding.

Making it a **recognised** token changes no observable behaviour today — it
lands in `Info` either way. It matters because it is what keeps `Negligible` at
the floor when #3306 moves the unrecognised bucket off it.

## Defect 3 — three private fallbacks (the reason 1 and 2 drifted)

Each adapter carried its own catch-all, which is how they diverged:

* `backend/src/services/grype_scanner.rs` — `.unwrap_or(Severity::Info)`
* `backend/src/services/scanner_service.rs` (`convert_trivy_findings`) — `.unwrap_or(Severity::Info)`
* `backend/src/services/openscap_scanner.rs` — `_ => Severity::Info`, with the missing `critical` arm above

## Scope

1. Add a `critical` arm to the OpenSCAP path so a `critical` result classifies
   as `Severity::Critical`.
2. Make `negligible` a **recognised** token that maps to the lowest bucket
   deliberately.
3. Collapse the three per-adapter fallbacks onto one shared classifier, and give
   the "what does an unrecognised token mean" decision a single named home so
   #3306 is a one-symbol edit.

## Out of scope

**Moving where an unrecognised token lands.** That is #3306. `UNKNOWN`,
XCCDF's `unknown`, an empty severity and any garbage token must keep
classifying at `Info` here, and the tests must pin that so #3306 has to change
it on purpose.

## Acceptance

* One classifier used by all three adapters.
* Tests drive the **literal tokens** Grype, Trivy and OpenSCAP emit through each
  adapter's real parse path — a test that only exercises `from_str_loose` misses
  the three adapter fallbacks, which is where the bug lives. All `--lib`; the
  coverage gate is `cargo nextest run --workspace --lib`.
* **Token-by-token behaviour neutrality except `critical`.** Enumerate every
  token each adapter handled before and after and show that only OpenSCAP
  `critical` moves. The evidence is that a faithful revert of all three adapters
  to their merge-base expressions reds exactly one test — the `critical` one —
  and every other new test stays green, because every other new test also holds
  against the pre-fix code.
* Revert-proved against the merge base with `DATABASE_URL` **and**
  `AK_TESTS_REQUIRE_DB=1` exported.
* A `CHANGELOG.md` entry sized to what actually ships — no grade-band upgrade
  note, because no grade moves.

brandonrc:
**Closes #3294** (stage 1a) — **Refs #3243**, which stays open for stages 2 and 3.
The deferred half is **#3306**, milestone 1.8.0.

> **This PR was reduced.** It previously carried the whole of #3243 stage 1,
> including moving unrecognised / `UNKNOWN` scanner tokens from `Info` to
> `Medium`. That half is a high-volume, gate-flipping behaviour change and has
> been split out to #3306 for 1.8.0. What remains is the low-volume,
> unambiguous half, and it is behaviour-neutral for every token except OpenSCAP
> `critical` — proved below by a revert rather than asserted.

---

### 1. OpenSCAP had no `critical` arm — the one behaviour change

`openscap_scanner::convert_findings` classified severities with its own private
match:

```rust
let severity = match f.severity.to_lowercase().as_str() {
    "high" => Severity::High,
    "medium" | "moderate" => Severity::Medium,
    "low" => Severity::Low,
    _ => Severity::Info,
};
```

No `critical` arm, so content reporting a `critical` rule failure had its **most**
severe result filed as its **least** severe one. `Info` is inert everywhere it
matters — `penalty_weight()` is `0`, it appears in none of `policy_service`'s
severity arrays at any threshold, and no gate's projection selects `info_count` —
so a `critical` OpenSCAP finding contributed to nothing at all: no policy
violation, no promotion block, no score penalty.

Volume is what makes this patch-safe. XCCDF 1.2 (NISTIR 7275r4 §6.4.4.2) defines
the Rule `severity` attribute as `unknown | info | low | medium | high`, so
`critical` is **not in the spec's vocabulary** and reaches this adapter only from
non-standard content or a non-default wrapper. Near-zero volume, unambiguous
fail-open when it does fire, fail-closed correction.

### 2. `Negligible` becomes a recognised token — behaviour-preserving

Grype's `Negligible` (the Debian and Ubuntu security trackers' explicit
"triaged, not worth fixing" grade) landed at `Info` by falling off the end of
`from_str_loose`. The lowest bucket is the right destination — Harbor documents
`negligible` as never blocking a pull — but nothing recorded that it was a
decision rather than a parse miss, so it was indistinguishable from an ungraded
finding.

**Correction to an earlier revision of this description:** this arm does *not*
govern the Harbor path. `image_scanner::normalize_harbor_severity` rewrites
Harbor's `Negligible` to the string `"Low"` **before** the token reaches the
classifier (`image_scanner.rs:170`, pinned by the pre-existing test at `:1190`),
so a Harbor `Negligible` classifies as `Severity::Low` and always has. That
mapping is deliberately unchanged here — `Low` violates a `max_severity = 'low'`
policy where `Info` violates nothing, so reconciling the two would have relaxed
a gate inside a patch — which leaves two `Negligible` mappings in the tree,
Grype's at `Info` and Harbor's at `Low`. The comment this PR adds at
`image_scanner.rs:160-168` records that; the CHANGELOG bullet now says it too.

It now arrives there deliberately. **No observable behaviour changes**: it maps
to `Info` before and after. It matters because it is what keeps `Negligible` at
the floor when #3306 moves the unrecognised bucket off it. Stated plainly: no
revert can distinguish this arm today, and the test that covers it says so in
its own doc comment rather than pretending otherwise.

### 3. One classifier for the three scanner adapters

`Severity::from_scanner_token` replaces the three per-adapter catch-alls that
let 1 and 2 drift apart in the first place, and
`Severity::UNRECOGNIZED_SCANNER_SEVERITY` gives the "what does an unrecognised
scanner token mean" decision a named home — currently `Severity::Info`, i.e.
exactly where each adapter's own fallback put it.

**Scoped honestly: this is not every unrecognised-severity fallback in the
crate.** There is a fourth, at a *different* bucket, which this PR deliberately
does not touch: `scanner_service::scan_dependencies` — the OSV/GitHub advisory
ingestion path — is
`Severity::from_str_loose(&advisory_match.severity).unwrap_or(Severity::Medium)`.
Advisory feeds are a different producer with a different vocabulary, so folding
them in would be a behaviour change for that path rather than a consolidation.
The consequence worth recording is for the follow-up, not for this PR: **#3306
is a one-symbol edit for the adapters, but repointing that symbol does not by
itself make the crate uniform** — whoever does it has to decide explicitly
whether the advisory fallback converges or stays independent. The doc comment on
the constant says so, and #3306 has been corrected to say so too.

`from_str_loose` is left strict and is now documented as the parser for
*operator-configured* thresholds (`policy_service` uses it with
`unwrap_or(Critical)`, `ScanConfig::threshold()` with `unwrap_or(High)`) — those
call sites are deliberately untouched, because "what an unparseable *configured*
value means" is a different question from "what an unparseable *scanner* token
means".

---

## Token-by-token proof that only `critical` moves

This is the whole basis for calling the PR patch-safe, so it is enumerated, not
sampled. `from_scanner_token` lowercases and **does not trim** — deliberately,
matching `from_str_loose` and therefore every adapter's prior behaviour, so a
whitespace-padded token stays unrecognised exactly as before.

### Grype (`grype_scanner::convert_findings`)

Before: `Severity::from_str_loose(&m.vulnerability.severity).unwrap_or(Severity::Info)`
After: `Severity::from_scanner_token(&m.vulnerability.severity)`

| token | before | after | moved? |
|---|---|---|---|
| `Critical` | Critical | Critical | no |
| `High` | High | High | no |
| `Medium` / `moderate` | Medium | Medium | no |
| `Low` | Low | Low | no |
| `info` / `informational` / `none` | Info | Info | no |
| `Negligible` | Info (parse miss) | Info (recognised arm) | no |
| `Unknown` | Info (parse miss) | Info (fallback) | no |
| `""`, `Severe`, `P1`, ` high `, anything else | Info (parse miss) | Info (fallback) | no |

Every token: unchanged. The two functions are extensionally equal.

### Trivy (`scanner_service::convert_trivy_findings`)

Before: `Severity::from_str_loose(&vuln.severity).unwrap_or(Severity::Info)`
After: `Severity::from_scanner_token(&vuln.severity)`

Same expression as Grype, so the same table applies verbatim — including
`UNKNOWN`, `Negligible` and the empty string. Every token: unchanged. This is
the shared converter for native Trivy, `trivy_fs_scanner`, `incus_scanner` and
Harbor reports, so all four producers are covered by that one equality.

### OpenSCAP (`openscap_scanner::convert_findings`)

Before: the private match above. After: `Severity::from_scanner_token(&f.severity)`.

| token | before | after | moved? |
|---|---|---|---|
| **`critical`** | **Info** | **Critical** | **YES — the one change** |
| `high` | High | High | no |
| `medium` / `moderate` | Medium | Medium | no |
| `low` | Low | Low | no |
| `info` / `informational` / `none` | Info (catch-all) | Info (recognised) | no |
| `unknown` (XCCDF's default) | Info (catch-all) | Info (fallback) | no |
| `negligible` | Info (catch-all) | Info (recognised arm) | no |
| `""` and anything else | Info (catch-all) | Info (fallback) | no |

The old match's image was exactly `{High, Medium, Low, Info}` and the new
classifier agrees on every one of those preimages; the only preimage that
changes image is `critical`, which the old match had no arm for.

### Unchanged files

`image_scanner.rs` carries a comment-only change (the `normalize_harbor_severity`
divergence note); `incus_scanner.rs` and `docker/scanner-adapter/harbor.go` are
untouched — both previously carried edits that only made sense with the deferred
half.

---

## What the `critical` arm changes for an operator

An earlier revision of this description said no gate outcome changes. That was
wrong, and it is corrected here and in the CHANGELOG. For every token except
OpenSCAP `critical` the claim holds by construction — nothing moves, so nothing
downstream can. For `critical` itself the direction is fail-**closed** on every
consumer, but the magnitude is not marginal and it ships with an upgrade note
rather than silently.

**The score moves, and the score is gated.** `recalculate_score`
(`scan_result_service.rs:1812-1849`) selects `DISTINCT ON (sr.artifact_id,
sr.scan_type)`, so OpenSCAP competes *alongside* Trivy and Grype rather than
replacing them, and every OpenSCAP finding is counted. It runs automatically
after every scan (`scanner_service.rs`, "Recalculate repository security
score"). `compute_security_score` charges **25** per critical against
`Grade::from_score`'s `90+ A / 75-89 B / 50-74 C / 25-49 D / else F`:

| OpenSCAP `critical` findings | score | grade |
|---|---|---|
| 0 | 100 | **A** |
| 1 | 75 | **B** |
| 2 | 50 | **C** |
| 3 | 25 | **D** |
| 4 | 0 | **F** |

Note this is a *different and steeper* scale than the one in the upgrade note
that was removed from this branch: that note's `4 → B, 9 → C, 17 → D, 26 → F`
was the deferred half's 3-point `Medium` arithmetic. Per finding, what actually
ships moves grades roughly 8x faster. The offsetting factor is population, not
size: XCCDF 1.2 does not define `critical`, so only custom or vendor-extended
SCAP content emits it, where the deferred half fires on the XCCDF *default*.

**And `critical` is a member of every policy severity array**, so unlike the
deferred half — which only bites at `max_severity` medium or low — a single
OpenSCAP `critical` finding violates a scan policy at **any** threshold, where
as `info` it violated none.

### Every severity consumer, enumerated

1. **Proxy/OCI inline download gate — neutral for any severity change at all.**
   `ProxyScanVerdict::is_vulnerable()` is `self.findings_count > 0` →
   `verdict_token()` → `proxy_scan_service::verdict_blocks`. Neither
   `decide_serve` nor `gate_proxy_scan_serve` takes a severity argument;
   `proxy_scan_results.max_severity` is written and read back into
   `ProxyScanRow` but no decision consults it. Re-bucketing cannot change a
   count. Pinned by `test_proxy_verdict_blocking_is_independent_of_severity`.
2. **Hosted-artifact download gate** — `quarantine_service::enforce_download_gate`
   → `enforce_scan_policy_gate` → `PolicyService::evaluate_artifact`, rendered
   as `403`. Reach: `check_artifact_download` (`quarantine_service.rs:624`) has
   30 call sites across 25 handler files, `artifact_service::prepare_download`
   (`:1455`) calls it directly, and `proxy_helpers.rs:3314` / `:5372` and
   `maven_proxy.rs:341` call the policy half. `info` is in none of its severity
   arrays; `critical` is in all of them.
3. **`promotion_rule_service::check_cve_severity`** — reads the `scan_results`
   count columns; `info_count` is never consulted, `critical_count` is.
4. **`promotion_policy_service::evaluate_cve_thresholds`** — reached from
   `api/handlers/promotion.rs` and `api/handlers/approval.rs`;
   `LATEST_SCAN_COUNTS_SQL` does not select `info_count`, and `max_critical` is
   `0` at every threshold.
5. **`promotion_rule_service::check_min_health_score`** (`:447-449`) — a
   promotion rule's `min_health_score` against
   `artifact_health_scores.health_score`, whose `security` component is fed from
   `repo_security_scores.score` (`quality_check_service.rs:392-402`) — the value
   `recalculate_score` writes.
6. **`quality_check_service` gate evaluation** (`:635-648`) — `min_health_score`
   **and** `min_security_score`, the latter reading the security score directly
   with no weighting in between.

Consumers 5 and 6 were missing from the previous revision of this list, which is
what made the score look display-only. The whole value of enumerating is that it
is complete, so: consumers 2, 3, 4, 5 and 6 all tighten for an OpenSCAP
`critical` finding, and 1 is neutral. Nothing anywhere becomes more permissive.

---

## Revert-proof

Merge base `36f7d12`. Own throwaway Postgres (`ak-split3293-pg`, port 30693),
freshly migrated from `backend/migrations/*.sql`, with `DATABASE_URL` **and**
`AK_TESTS_REQUIRE_DB=1` exported for every run so no DB-backed `--lib` test
could early-return a false PASS (DB tests came back in 0.27–0.48s, not 0.01s).

**Baseline (fixed tree)** — `cargo nextest run --workspace --lib --no-fail-fast`:

```
Summary [212.471s] 15126 tests run: 15126 passed (3 slow), 8 skipped
```

**R1 — faithful revert of all three adapters to their merge-base expressions**,
same command, same database:

```
Summary [246.286s] 15126 tests run: 15125 passed (2 slow), 1 failed, 8 skipped

FAIL services::openscap_scanner::tests::test_convert_findings_openscap_critical_is_not_downgraded
```

**Exactly one test fails, and it is the `critical` one.** That single run proves
both halves of the claim at once:

* the `critical` arm is genuinely gated — restore the old code and the test goes
  red;
* **and everything else really is behaviour-neutral** — every other test added by
  this PR (the Grype table, the Trivy table, the OpenSCAP graded/`unknown`
  tables, the classifier tests, the proxy invariant) passes against the
  *pre-fix* production code, which is what "no other token moves" means
  operationally. 15126 − 1 = 15125 passed, so there is no collateral failure
  anywhere in the suite either.

**Faithfulness of the revert**, checked on the **reverted** tree rather than on
the PR diff. Every `#[cfg(test)] mod … { }` block stripped from the merge-base
file and from the reverted file, remainders compared byte-for-byte:

```
IDENTICAL    72035 ->   72035 bytes  backend/src/services/grype_scanner.rs
IDENTICAL   260155 ->  260155 bytes  backend/src/services/scanner_service.rs
IDENTICAL    12854 ->   12854 bytes  backend/src/services/openscap_scanner.rs
IDENTICAL    40423 ->   40423 bytes  backend/src/services/incus_scanner.rs
IDENTICAL     4830 ->    4830 bytes  docker/scanner-adapter/harbor.go
DIFFERS      40474 ->   41145 bytes  backend/src/services/image_scanner.rs   [non-comment production: IDENTICAL]
DIFFERS      10826 ->   15160 bytes  backend/src/models/security.rs
```

Both differences are accounted for:

* **`image_scanner.rs`** — the entire production delta is comment text. Filtering
  the stripped production source to non-comment lines yields byte-identical
  remainders, so no executable byte changed.
* **`models/security.rs`** — R1 reverts the three *adapters*; the classifier is
  retained so the file still compiles for the classifier-level tests. Safe
  because the addition is **purely additive** (diffing the test-stripped
  production source against the merge base: **0** removed or changed merge-base
  lines, 72 added) and **unreachable** — grepping the reverted tree for
  `from_scanner_token` outside `models/security.rs` returns exactly one hit, a
  doc comment in `image_scanner.rs`, and **no call sites**. An uncalled function
  cannot make an adapter test fail.

**Why there is no M1 for the `negligible` arm.** With
`UNRECOGNIZED_SCANNER_SEVERITY == Severity::Info`, deleting
`"negligible" => Severity::Info` leaves every observable result unchanged — the
arm is behaviour-preserving by construction, which is precisely why it is
allowed in a patch. It becomes load-bearing, and independently revert-provable,
in #3306. The test says this in its own doc comment rather than implying a guard
it does not have.

---

## Neutrality pins for the deferred half

Four tests pin the *unmoved* behaviour so #3306 cannot happen by accident inside
a patch release, and each names #3306 in a comment so it is deleted on purpose:

* `models::security::test_unrecognized_tokens_still_classify_at_the_floor`
* `grype_scanner::test_convert_findings_grype_unrecognized_severity_is_unchanged_pending_3306`
* `scanner_service::test_convert_trivy_findings_unrecognized_severity_is_unchanged_pending_3306`
* `openscap_scanner::test_convert_findings_xccdf_unknown_is_unchanged_pending_3306`

Each carries a positive control in the same fixture (a graded token that must
still classify correctly), so none of them can be satisfied by a converter that
stopped grading anything.

---

## CHANGELOG

Rebased onto `5405584` (#3312, "add missing 1.7.2 operator upgrade notes"), and
this change now carries an operator note of its own.

The branch previously carried an upgrade note titled *"repositories with
OpenSCAP findings may drop a security-grade band"*, whose arithmetic
(4 findings → B, 9 → C, 17 → D, 26 → F) is the **deferred** half's 3-point
`Medium` scale and would have been wrong here. It was removed on the basis that
no grade moves — which was itself wrong, as the table above shows. Both halves
are now fixed:

* the deferred material (grade bands, four gates, `min_health_score`) is carried
  into #3306 in full, so the 1.8.0 work does not re-derive it;
* what ships carries a **new note with the correct 25-point arithmetic**, placed
  as the fourth item of #3312's collected *"smaller operational changes worth
  two minutes before you deploy"* list — retitled seven → eight — directly after
  the `scan_policies` sizing item it shares a table with. It states the
  mechanism, the `1 → B … 4 → F` bands, that `critical` violates a policy at any
  threshold, the gates it reaches, an **"Am I affected?"** paragraph scoping it
  to OpenSCAP deployments whose content declares a non-standard
  `severity="critical"`, the fact that nothing is backfilled, and a sizing query
  an operator can run before upgrading.

The `### Fixed` bullet is also corrected: it previously said `Negligible` "lands
in the same bucket it always did (`Info`)", which is false for Harbor-sourced
reports, where `normalize_harbor_severity` has always mapped it to `Low` ahead
of the classifier. No regression — but wrong in the direction that matters,
since `Low` violates a `max_severity='low'` policy, and it contradicted a
comment this same PR adds.

*(The reduction also dropped a duplicated `#3282` CHANGELOG bullet the branch
had picked up in a rebase.)*

---

## Checks

* `cargo fmt --all` clean.
* `cargo clippy --workspace --all-targets -- -D warnings` — exit 0.
* `SQLX_OFFLINE=true cargo build --lib --tests` clean.
* No `sqlx::query!` macro text touched, so `cargo sqlx prepare` was not required.
* Milestone 1.7.2.


brandonrc:
## Independent validation — VERDICT: fix is correct, direction argument holds; three things to fix before merge, none of them a security regression

I set out to refute this. I could not find any input, on any path, that becomes
**more permissive**. Every claim in the description reproduced against source or
against a live run, and the R1 / M1 / anti-vacuity experiments reproduce exactly
as reported. What I did find is that the blast radius is **larger and more
concentrated than the description says**, that one severity-reading gate is
missing from the proof table, and that a strictly worse adjacent fail-open sits
one line away from the code this PR touches.

Method: fresh shallow clone, `pull/3293/head` at `a2283fda`, own Postgres
(`ak-val3293-pg`, port 30493) migrated from scratch, every run with
`DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` exported (DB-backed tests came
back in 0.38–0.42s, not 0.02s, so they really ran). No pushes, no CI triggered.

---

## The monotonicity proof, done independently

The direction claim reduces to two questions, and both are decidable.

**(a) Does any input's classification move *down* in severity?** No — provably.

* Grype / Trivy (`grype_scanner.rs:1251`, `scanner_service.rs:1963`): pre-fix was
  `from_str_loose(s).unwrap_or(Info)`. If `from_str_loose(s) = Some(x)` the new
  path returns the same `x` (`from_scanner_token` lowercases and trims, then
  delegates to a parser that lowercases again — trimming cannot un-parse a
  string that already parsed). If it was `None`, the old value was `Info`, the
  floor, so the new value can only be equal or more severe.
* OpenSCAP (`openscap_scanner.rs:249`): the old private match produced exactly
  `{High, Medium, Low, Info}`; the new classifier agrees on `high`, `medium`,
  `moderate`, `low`, and everything else was already at the floor.
* `negligible` is the one token deliberately pinned at the floor in both shapes.

So the changed-input set is exactly: **unrecognised tokens Info → Medium**, plus
**OpenSCAP `critical` Info → Critical**. Nothing else moved. Claim 1 CONFIRMED.

**(b) Is any consumer non-monotone in severity?** No. `info` is not a member of
any severity array in `policy_service`, and there is no `info_count` in any
`scan_results`-reading gate — so an `Info` finding contributes to *nothing*, and
every consumer is `count > limit` / `level < threshold` / additive penalty.
Confirmed by running the actual `policy_service` CASE expression against Postgres:

```
 finding_severity | violates at max_severity critical/high/medium/low
------------------+---------------------------------------------------
 info             | 0 / 0 / 0 / 0
 medium           | 0 / 0 / 1 / 1
```

Since classification is monotone non-decreasing and every consumer is monotone,
"nothing relaxes" follows structurally rather than by enumeration. That is a
stronger statement than the description makes, and it survives the finding below.

---

## Claim-by-claim

| # | Claim | Verdict | Evidence |
|---|---|---|---|
| 1 | one classifier, all three adapters; `negligible` recognised; rest → `Medium` | **CONFIRMED** | `models/security.rs:114,138`; call sites `grype_scanner.rs:1251`, `scanner_service.rs:1963`, `openscap_scanner.rs:249`; no `unwrap_or(Severity::Info)` remains anywhere |
| 2 | `from_str_loose` stays strict, now the operator-config parser only | **CONFIRMED** | `policy_service.rs:260` `unwrap_or(Critical)`, `models/security.rs:236-238` `threshold()` `unwrap_or(High)`, `scan_config_service.rs:70` (400-on-invalid) — all untouched |
| 3 | `Medium` over a new variant; four CHECKs, a sixth count column, and the `Ord` total order | **CONFIRMED** | `022:13`, `022:46`, `022:84`, `053:52` all verified verbatim; `Severity` derives `Ord` with `Critical = 0 … Info = 4` (`models/security.rs:41-53`) and `meets_threshold` is `self <= threshold` (`:160-162`) |
| 4 | the #3243 gate is unchanged | **CONFIRMED** | `is_vulnerable()` = `findings_count > 0` (`scanner_service.rs:2503`); `proxy_scan_results.max_severity` is written (`proxy_helpers.rs:6106`) and read into `ProxyScanRow` but no decision consults it — I re-grepped every use |
| 5 | three severity-reading paths, all fail-closed | **CONFIRMED in direction, INCOMPLETE in enumeration** | see F1 |
| 6 | Harbor `Negligible` → `Low` deliberately not reconciled | **CONFIRMED** | `image_scanner.rs:159-176`; `Low` violates a `max_severity='low'` policy and `Info` violates nothing, so reconciling really would have relaxed a gate |
| 7 | OpenSCAP had no `critical` arm; XCCDF's default severity is `unknown` | **CONFIRMED** — and see F2, because the second half is much bigger than the description makes it sound | |

### Full suite, reproduced independently

`cargo nextest run --workspace --lib --no-fail-fast` on a freshly migrated
database: **`Summary [220.555s] 15122 tests run: 15122 passed (3 slow), 8
skipped`** — the same 15122/8 the description reports, no collateral failures,
and the `storage_gc_service` flake it mentions did not recur on a clean database.

### Revert-proofs, reproduced independently

* **R1** — reverted the three adapter expressions to the merge-base shapes.
  Faithfulness checked on the *reverted* tree, not on the PR diff: production
  code (test modules stripped, comments stripped) is **byte-identical to
  `b9d4289` for all five service files**; only `models/security.rs` differs, by
  the purely-additive classifier, and I confirmed it has **no call sites** in the
  reverted tree. Result: **exactly the 8 failures listed**, no collateral.
* **M1** — removed only `"negligible" => Severity::Info`: **exactly the 4
  failures listed**, and every `UNKNOWN` test stayed green. The two halves really
  are independently gated.
* **Anti-vacuity** — repointed `UNRECOGNIZED_SCANNER_SEVERITY` to `Info` on the
  otherwise-fixed tree. `test_proxy_verdict_blocking_is_independent_of_severity`
  fails on `fixture must actually re-bucket the findings`, i.e. the anti-vacuity
  assertion fires first, as designed. Not hollow.

---

## F1 — a **fourth** severity-reading gate is missing from the proof table (medium)

The description says "**Two** do", then enumerates `PolicyService`,
`promotion_rule_service::check_cve_severity` and the health score. There is a
fourth, and it is a separately-configured gate on a separate surface:

**`promotion_policy_service::evaluate_cve_thresholds`**
(`backend/src/services/promotion_policy_service.rs:153-193`), fed by
`build_cve_summary` (`:140-148`) over `LATEST_SCAN_COUNTS_SQL` (`:79-94`), which
selects `critical_count, high_count, medium_count, low_count, findings_count` —
**`info_count` is not in the projection**. Reached from
`api/handlers/promotion.rs:664` **and** `api/handlers/approval.rs:442`.

It is driven by `scan_policies.max_severity` (not `promotion_rules.max_cve_severity`),
and at `medium` or `low` it sets `max_medium = 0` (`:159-163`), so:

> an ungraded finding, before: contributes to no count column, no violation at
> any threshold. After: `medium_count += 1` → a new `cve-severity-threshold`
> violation on **both** the promotion gate and the approval gate, for every
> repository with an enabled `scan_policies` row at `max_severity` medium or low.

Direction is fail-closed, so the PR's conclusion survives intact. But the whole
value of the proof section is that it is exhaustive, and it is not. Fix: add this
row to the table with its before/after.

(Two smaller precision notes on the same section: the `critical*25 + high*10 +
medium*3 + low*1` formula is `compute_security_score` in
`scan_result_service.rs:143-156`, which computes the **repo security score /
grade**; it reaches `min_health_score` only indirectly, as the `security` component
of `compute_weighted_health_score` in `quality_check_service.rs:388-425`. And the
`Info`-excluded severity arrays are at `policy_service.rs:266-282`.)

---

## F2 — the OpenSCAP half is the common case, not a long tail, and nothing tells an operator (medium)

The description frames XCCDF `unknown` as "the single most common token in real
SCAP content". It is stronger than that in **this** product, because AK's own
wrapper manufactures it:

* `scripts/openscap-wrapper.py:211` — `severity = rule.get("severity", "unknown")`
* `scripts/openscap-wrapper.py:231` — `severity = rule_result.get("severity") or meta.get("severity", "unknown")`

So every rule with no `severity` attribute is emitted to the backend as the
literal string `"unknown"`, and `openscap_scanner.rs:240` reports every rule whose
result is `fail | error | unknown`. On the next scan, **all of them move Info →
Medium at once**.

Priced through the code:

* `compute_security_score` (`scan_result_service.rs:143-156`) charges 3 per
  medium and 0 per info; `Grade::from_score` (`models/security.rs:186-193`) is
  `90+ = A, 75-89 = B, 50-74 = C, 25-49 = D, else F`.
* **26 previously-`Info` OpenSCAP findings take a repository from score 100 /
  grade A to score 22 / grade F.** 9 findings is already a C.
* That score is the `security` component of `artifact_health_scores`, so any
  `min_health_score` promotion rule tightens with it — plus the four gates above.

Scoping, stated honestly because it materially limits the impact: OpenSCAP is
opt-in (`scanner_service.rs:4148` registers it only when a sidecar URL is
configured), and existing `scan_findings` rows keep their stored `'info'` — the
flip happens on the next scan, with no backfill. Both facts belong in the note
rather than being left for an operator to discover.

**There is no CHANGELOG entry.** The four fix commits immediately beneath this
one — `b9d4289` (#3288), `8b9e0ff` (#3278), `a44bb33` (#3251), `f5abd2f` (#3151) —
all update `CHANGELOG.md` under `## [Unreleased] / ### Fixed`. This PR touches six
files and none of them is `CHANGELOG.md`. Given the milestone is **1.7.2, a
patch**, and given four gates tighten, that entry is the only place an operator
would ever see this. Please add one, and say in it: the `UNKNOWN` → `Medium`
re-bucketing, the OpenSCAP unrated-rule case by name, that scores/grades will move
on the next scan, and that existing findings are not rewritten.

Related scope note for the maintainer, not a blocker: #3294 lists "**Out of
scope:** anything that changes what the product blocks", and four gates do change.
The PR is completely honest about this in its proof section, and the direction is
the one #3218 / #3138 established as arguable — but the acceptance criterion as
written is not met, and #3245 / #3247 both declined a behaviour change in a patch.
Whether 1.7.2 is the right vehicle is your call, not the pipeline's.

---

## F3 — the adjacent fail-open this PR's premise walks past (medium, follow-up issue)

On **three of the four** Trivy producers, `UNKNOWN` never reaches AK at all,
because the CLI invocation filters it out before the report is written:

* `backend/src/services/trivy_fs_scanner.rs:104-105` — `--severity CRITICAL,HIGH,MEDIUM,LOW`
* `backend/src/services/incus_scanner.rs:174-176` — same list
* `docker/scanner-adapter/scan.go:316` — `const fsSeverity = "CRITICAL,HIGH,MEDIUM,LOW"`, whose own comment says it "mirrors the backend's legacy `trivy filesystem` CLI invocation … so routing through the adapter (#2363) does not change which findings are reported"

Only the **image** path admits it: `docker/scanner-adapter/config.go:77` defaults
`SCANNER_TRIVY_SEVERITY` to `"UNKNOWN,LOW,MEDIUM,HIGH,CRITICAL"`, and
`harbor.go:92-107` maps `UNKNOWN → "Unknown"`, which
`image_scanner::normalize_harbor_severity` passes through to the new classifier.
So the fix is real and reachable — via Harbor/image reports — but on the
hosted-artifact and incus filesystem paths an ungraded Trivy CVE is **not
mis-bucketed, it is absent**. It never becomes a `RawFinding`, so it is invisible
to `scan_findings`, to every count column, and to `findings_count > 0` — i.e. to
the proxy gate too. **That is strictly worse than filing it at `Info`**, which is
the defect this PR exists to fix, and it is untouched here.

Consequence for the tests as written:
`test_convert_trivy_findings_does_not_bucket_unknown_as_lowest` drives a token
that three of that converter's four producers cannot emit. The test is correct and
worth keeping — it does gate the Harbor path — but it reads as though the Trivy
`UNKNOWN` fail-open is now closed generally, and it is not. Suggest a
follow-up issue to add `UNKNOWN` to those three `--severity` lists, and a
sentence in the doc comment on `convert_trivy_findings` recording which producers
can actually reach the `UNKNOWN` arm today.

---

## F4 — stale cross-component comment, one line, fixable in this PR (low)

`docker/scanner-adapter/harbor.go:105`:

```go
// Empty or unrecognized -> Unknown (backend maps Unknown -> Info).
```

That parenthetical is now false, and it is precisely the comment a future reader
of the adapter will trust when deciding what the adapter may safely emit. Update
it to say the backend classifies `Unknown` at
`Severity::UNRECOGNIZED_SCANNER_SEVERITY`.

---

## F5 — the invariance test couples itself to the constant it is meant to outlive (low)

`scanner_service.rs:10678`:

```rust
assert_eq!(after.max_severity, Some(Severity::Medium));
```

The classifier tests are deliberately written against the *ordering* so they
survive repointing `UNRECOGNIZED_SCANNER_SEVERITY`
(`models/security.rs:483-497`), and the standing-invariant test should hold to the
same standard: today, moving the constant to `High` reds the proxy-invariance test
for a reason that has nothing to do with the invariant it guards. Prefer
`assert!(after.max_severity < Some(Severity::Info))`, or compare against
`Some(Severity::UNRECOGNIZED_SCANNER_SEVERITY)`. (The `assert_eq!(before.max_severity,
Some(Severity::Info))` half is fine — `Info` is a fixed endpoint of the order.)

---

## Credit where it is due

Two calls here are better than the average PR that crosses this pipeline.

**Pricing the new-variant option against the `Ord` total order.** The migration
inventory would have been enough to justify `Medium` on cost, and most authors
would have stopped there. The stronger argument is the one that was actually
made: `Severity` derives `Ord` and `meets_threshold` is `self <= threshold`, so a
sixth variant must be assigned an honest rank in a **total** order, and "ungraded"
has none — it is an absence of a severity, not a position between two of them.
That is a real modelling objection, not a cost objection, and it is the reason a
distinct `Unknown` variant would have been wrong here even if the schema were
free. Keeping `UNRECOGNIZED_SCANNER_SEVERITY` as the single symbol to repoint is
the right hedge.

**Refusing to reconcile `normalize_harbor_severity`.** Passing Harbor's
`Negligible` through to the shared classifier is the tidier-looking change, and
it would have been wrong: `Low` violates a `max_severity = 'low'` policy and
`Info` violates nothing, so tidiness would have silently relaxed a gate inside a
PR whose entire premise is that it relaxes nothing. Leaving the divergence in
place, with a comment recording *why* it is deliberate and where it should be
settled, is the correct trade — and noticing it at all required actually tracing
the Harbor path rather than trusting that "one classifier" meant one mapping.

brandonrc:
## Remediation pushed — F1, F2, F3, F4, F5

Thanks for the validation pass; the enumeration gap and the missing operator note
were both real. Everything below is pushed to `fix/3243-severity-mapping`
(commit 7950987). No production logic changed — the classifier, the three adapters and
the `normalize_harbor_severity` divergence are byte-for-byte what you validated.

Method: fresh shallow clone at `pull/3293/head`, own Postgres
(`ak-fix3293b-pg`, port 30593) migrated from scratch, every test run with
`DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` exported.

---

### F2 (highest priority) — the operator-visible impact is now documented

`CHANGELOG.md` had no entry for this PR at all. It now has two, under
`[Unreleased]`:

1. A `### Fixed` bullet for the classification change itself, in the shape of the
   surrounding entries.
2. An **`### Upgrade note — repositories with OpenSCAP findings may drop a
   security-grade band`** section, placed and structured like the 1.7.1
   `### Upgrade note — proxy downloads may start being blocked` that preceded it,
   including its **"Am I affected?"** paragraph.

The note states the mechanism in the terms an operator can act on: unrecognised
tokens previously landed at `Info`, which has `penalty_weight() == 0` and is a
member of no policy severity array, so an unrated finding contributed to
**nothing**; they now land at `Medium`. It says that this is fail-closed and
deliberate, that no gate anywhere becomes more permissive, and that the inline
proxy/OCI gate — `findings_count > 0` — is untouched.

It also says explicitly, because your point stands that this is the common case
and not a tail: AK's own wrapper manufactures the token
(`scripts/openscap-wrapper.py:211` and `:231` both default to the literal
`"unknown"`), so on a typical SCAP profile the *unrated* rules are the norm.

**I re-derived the grade figure rather than quoting yours, and I get the same
number: 26.** Working, from source rather than from the PR body —
`compute_security_score` (`scan_result_service.rs:143-156`) is
`100 - (critical*25 + high*10 + medium*3 + low*1)` clamped to `[0,100]`, with
weights read off `Severity::penalty_weight` (`models/security.rs:58-66`), banded
by `Grade::from_score` (`:186-193`):

| previously-`Info` findings | penalty | score | grade |
|---|---|---|---|
| 0 | 0 | 100 | **A** |
| 4 | 12 | 88 | **B** |
| 9 | 27 | 73 | **C** |
| 17 | 51 | 49 | **D** |
| 26 | 78 | 22 | **F** |

Each row is the *first* count that reaches that band, so 26 is exactly the
A-to-F threshold on an otherwise-clean repository. Your 9-is-already-a-C
checkpoint reproduces too.

On the `min_health_score` half, one addition worth having in the note: the path
is `promotion_rule_service.rs:447-449`, which reads
`artifact_health_scores.health_score` — and since the license component is never
populated today (`quality_check_service.rs` hardcodes `license_score = None`),
an artifact with no quality or metadata check results has a health score **equal**
to its repository's security score, because `compute_weighted_health_score`
averages only the components that have data. So in that (common) configuration
the gate tracks the security score 1:1 rather than at weight 40/100. The note
also points at quality gates' `min_security_score`, which reads the security
score with no weighting in between at all.

Scoping is stated in the note as well, since it materially limits the blast
radius: OpenSCAP is opt-in (`OPENSCAP_URL`), acknowledged findings are excluded
from the score, and stored `scan_findings` rows keep their recorded `'info'` —
no backfill, no migration, so a repository moves only when its artifacts are
next scanned.

### F1 — the proof table is now exhaustive

Added **§4 `promotion_policy_service::evaluate_cve_thresholds`** with the same
before/after treatment as the other three, and renumbered the health score to §5.
Verified against source rather than taken from the report:

* `LATEST_SCAN_COUNTS_SQL` (`:79-94`) projects
  `critical_count, high_count, medium_count, low_count, findings_count` —
  no `info_count`, so an `Info` finding was invisible here exactly as it was in §3.
* `evaluate_cve_thresholds` (`:153-193`): `"medium" | "low" => (0, 0, 0)`, so at
  either threshold one re-classified finding is a new `cve-severity-threshold`
  violation.
* Reached from `promotion.rs:664` **and** `approval.rs:442` — two gates, not one.
* Driven by `scan_policies.max_severity`, a different surface from §3's
  `promotion_rules.max_cve_severity`. Confirmed.

One scoping detail I added while verifying: with no `scan_policies` row,
`evaluate_cves_against_policy` (`:368-380`) falls through to a default policy that
blocks only on `critical_count > 0`, so it is unaffected — worth stating so the
row is not read as "every deployment".

The body's "**Two do**" is corrected to "**Four do**", and the section now says
in as many words that the enumeration is meant to be exhaustive rather than
illustrative, and that §4 was added after review. Your two precision notes are
folded in as well: the formula is `compute_security_score` in
`scan_result_service.rs`, producing the **repository** score/grade, reaching
`min_health_score` only via `compute_weighted_health_score`.

### F3 — filed as #3296, not fixed here

**#3296 — "Trivy `--severity` allowlist drops UNKNOWN findings entirely on
filesystem and incus scans"**, milestone **1.8.0**.

I confirmed all three call sites independently (`trivy_fs_scanner.rs:104-105`,
`incus_scanner.rs:174-176`, `scan.go:316` — plus `config.go:77`, which is why the
image path is the only one that admits `UNKNOWN`). The issue records why this is
strictly worse than mis-bucketing rather than merely adjacent to it: a
mis-bucketed finding is still a `scan_findings` row and still increments
`findings_count`, which is what the severity-blind proxy gate reads, so even at
the floor it was doing work; a filtered finding produces no `RawFinding` at all
and is invisible to every gate including that one.

It carries the testing requirement you asked for, stated as the blocking
condition: **the guard must assert the ungraded finding is _present_**, not that
a scan succeeded — a scan over ungraded content completes normally today with the
finding silently dropped, so every "scan ran / returned a report /
`findings_count > 0`" assertion is satisfied by the bug and cannot fail on the
unfixed tree. It also asks for a graded finding in the same fixture as a positive
control.

1.8.0 rather than the patch line because the fix changes *which findings are
reported*, which is not patch-line work on the #3245 / #3247 precedent, and the
exposure is a missing signal on an opt-in path rather than an active bypass.

### F4 — stale cross-component comment

`docker/scanner-adapter/harbor.go`, `mapSeverity`'s `default` arm. The
parenthetical "(backend maps Unknown -> Info)" is false as of this PR, and it is
exactly the comment a future reader of the adapter would trust when deciding what
the adapter may safely emit. It now records that the backend routes `Unknown`
through `Severity::from_scanner_token` and classifies it at
`Severity::UNRECOGNIZED_SCANNER_SEVERITY`, and that emitting `Unknown` is
therefore not a way to make a finding harmless.

### F5 — the invariance test no longer couples to the constant

```rust
-        assert_eq!(after.max_severity, Some(Severity::Medium));
+        assert!(
+            after.max_severity < Some(Severity::Info),
+            "fixture must re-bucket strictly above the floor, whatever bucket \
+             UNRECOGNIZED_SCANNER_SEVERITY currently names"
+        );
```

Same standard `test_unrecognized_scanner_severity_is_not_the_lowest_bucket`
already holds itself to. The `before` half still pins `Info` literally, per your
note that `Info` is a fixed endpoint of the order.

**Mutation-proved, both directions**, because a weakened assertion is the obvious
failure mode here:

| mutation | old assertion | new assertion |
|---|---|---|
| `UNRECOGNIZED_SCANNER_SEVERITY = Severity::High` | **FAIL** — `left: Some(High)` / `right: Some(Medium)` at `:10686` | **PASS** |
| `UNRECOGNIZED_SCANNER_SEVERITY = Severity::Info` | **FAIL** — anti-vacuity `assert_ne!` at `:10673` | **FAIL** — anti-vacuity `assert_ne!` at `:10673` |

The first column is the defect you reported: repointing the constant to `High`
reds a test whose subject is "the proxy gate ignores severity". The second row is
the control that the new assertion did not simply become vacuous — with the
constant back at the floor, the fixture stops re-bucketing and the test still
fails, on the anti-vacuity assertion, exactly as it did before.

---

### Verification

* `cargo fmt --all -- --check` — exit 0.
* `SQLX_OFFLINE=true cargo build --lib --tests` — clean, no warnings.
* `cargo clippy --workspace --all-targets -- -D warnings` — **exit 0**.
* `cargo nextest run --lib` over `test(/scanner/) or test(/severity/) or test(/security/) or test(/promotion/) or test(/health_score/)` — **1243 passed**, DB-backed tests returning in 0.3-1.1s rather than 0.03s, so they really ran.
  * One run showed `api::handlers::admin_security::tests::test_accessible_users_enumeration_db` red. It passes in isolation and passed on an immediate re-run of the identical filter (1243/1243), and this push contains **zero** production changes, so it is cross-test interference on a shared database — the same species as #3277 / #3279 — not a regression from this branch.
* `docker/scanner-adapter` — Go is not installed in this environment, so `gofmt`/`go test` were not run. The change is a comment-only edit inside an existing `default:` arm, tab-indented to match its neighbours; the diff contains no code lines.
* No `sqlx::query!` macro text touched, so `cargo sqlx prepare` was not required.

Nothing in `models/security.rs`, `grype_scanner.rs`, `openscap_scanner.rs`,
`incus_scanner.rs` or `image_scanner.rs` changed in this push, and the only
change in `scanner_service.rs` is inside `#[cfg(test)]`, so R1 / R2 / M1 stand as
validated.

brandonrc:
## This PR has been split — the `UNKNOWN` → `Medium` half moved to 1.8.0 (#3306)

Per the maintainer decision, #3293 now carries only the low-volume, unambiguous
half of #3243 stage 1. The description has been rewritten to match the diff.
Force-pushed to `fix/3243-severity-mapping`; still milestone 1.7.2.

### What stayed

1. **The OpenSCAP `critical` arm.** The adapter's private match was
   `high | medium | moderate | low` with `_ => Severity::Info` and no `critical`
   case, so a `critical` rule failure was filed as the *least* severe finding on
   record — and `Info` contributes to nothing: `penalty_weight() == 0`, no
   membership in any `policy_service` severity array at any threshold, no
   `info_count` in any gate's projection. XCCDF 1.2 (NISTIR 7275r4 §6.4.4.2)
   defines the Rule `severity` attribute as `unknown | info | low | medium |
   high`, so `critical` is outside the spec's vocabulary and only non-standard
   content reaches that arm. Near-zero volume, unambiguous fail-open when it
   fires. **This is the only behaviour change in the PR.**
2. **`Negligible` recognised.** Same bucket before and after (`Info`) — it just
   gets there by an explicit arm instead of by falling off the end of a match.
   No observable behaviour change; it is what keeps `Negligible` at the floor
   when #3306 moves the unrecognised bucket.
3. **The consolidation** onto `Severity::from_scanner_token`, with
   `UNRECOGNIZED_SCANNER_SEVERITY` pinned at `Severity::Info` — exactly where
   each adapter's own fallback put it — so #3306 is a one-symbol edit.

### What moved to #3306 (1.8.0)

Unrecognised / `UNKNOWN` tokens classifying above the floor. I re-derived the
blast radius from source rather than taking the previous description's word for
it, and it is worse than that description said in one specific way: **the
`PolicyService` path is a download gate, not just a reporting surface.**

`quarantine_service::enforce_download_gate` → `enforce_scan_policy_gate` →
`PolicyService::evaluate_artifact`, rendered as
`403 "Artifact download is blocked by the repository's scan policy"`. Reach:
`quarantine_service::check_artifact_download` (`:624`) delegates to it and has
**30 call sites across 25 handler files** — the per-format download handlers;
`artifact_service::prepare_download` (`:1455`) calls it directly for the generic
buffered and streaming download paths; and `proxy_helpers.rs:3314` / `:5372` and
`maven_proxy.rs:341` call the policy half. Its severity `CASE` run directly
against Postgres:

```
 finding_severity | policy_max_severity | violates_policy
------------------+---------------------+-----------------
 info             | critical            | f
 info             | high                | f
 info             | medium              | f
 info             | low                 | f
 medium           | critical            | f
 medium           | high                | f
 medium           | medium              | t
 medium           | low                 | t
```

So an artifact whose only findings are ungraded is **served today** and would
be **`403`'d after the remap**, on the next scan, in any repository with an
enabled `scan_policies` row at `max_severity` `medium` or `low`. Add the three
gates the previous description already listed — `check_cve_severity`,
`evaluate_cve_thresholds` (promotion *and* approval), and the security score,
where 4 previously-`Info` findings take a clean repository to grade B and 26 to
F — and the volume driver that Artifact Keeper's own OpenSCAP wrapper defaults a
rule's severity to the literal `unknown` (`scripts/openscap-wrapper.py:211,231`),
and it is not a patch. The direction is fail-closed, which is the defensible
direction, but the size is the problem, not the sign.

Worth stating explicitly because it was the obvious counter-argument: the
**inline proxy/OCI download gate is not affected either way**.
`ProxyScanVerdict::is_vulnerable()` is `findings_count > 0` and consults no
severity, so an ungraded finding already blocks there today and re-bucketing
cannot change a count. That is why the remap looked cheaper than it is — the
gate #3243 is named after is the one gate that does not move.
`test_proxy_verdict_blocking_is_independent_of_severity` lands here as a
standing invariant.

### Proof that only `critical` moves

The description carries the full token-by-token table for all three adapters.
The compact version: the Grype and Trivy call sites are extensionally identical
before and after (`from_str_loose(..).unwrap_or(Info)` vs `from_scanner_token`,
which agrees on every input including `UNKNOWN`, `Negligible` and `""`), and on
OpenSCAP the old match's image was exactly `{High, Medium, Low, Info}` with the
new classifier agreeing on every one of those preimages — `critical` is the only
preimage whose image changes.

The experiment that establishes it: baseline on the fixed tree is
`15126 tests run: 15126 passed, 8 skipped`; then a faithful revert of all three
adapters to their merge-base expressions, same full `--workspace --lib
--no-fail-fast` command against the same freshly-migrated Postgres, with
`DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` exported:

```
Summary [246.286s] 15126 tests run: 15125 passed (2 slow), 1 failed, 8 skipped

FAIL services::openscap_scanner::tests::test_convert_findings_openscap_critical_is_not_downgraded
```

One failure, and it is the `critical` one. That single run does double duty: it
gates the fix, and it proves the neutrality claim, because every *other* test
this PR adds passes against the pre-fix production code.

Revert faithfulness was checked on the reverted tree with test modules stripped:
`grype_scanner.rs`, `scanner_service.rs`, `openscap_scanner.rs`,
`incus_scanner.rs` and `harbor.go` are byte-identical to the merge base;
`image_scanner.rs` differs only in comment text (non-comment remainders
identical); `models/security.rs` differs only by the retained classifier, which
is purely additive (0 merge-base production lines removed or changed) and has no
call sites in the reverted tree.

I have **not** claimed a revert-proof for the `negligible` arm. With
`UNRECOGNIZED_SCANNER_SEVERITY == Severity::Info`, deleting that arm changes
nothing observable — it is behaviour-preserving by construction, which is
exactly why it is allowed in a patch. It becomes load-bearing and independently
provable in #3306, and the test's own doc comment says so.

### CHANGELOG

The branch carried an upgrade note titled *"repositories with OpenSCAP findings
may drop a security-grade band"* whose arithmetic (4 findings → B, 9 → C,
17 → D, 26 → F) describes the **deferred** half. No grade moves in this PR, so
shipping it would have been false. It is not lost: the grade-band table, the
four gates with their before/after, the `min_health_score` 1:1 relationship and
the scoping caveats are all carried into #3306 in full, so the 1.8.0 work
assembles the note from there rather than re-deriving it. What ships instead is
one `### Fixed` bullet sized to the actual change — net CHANGELOG diff versus
the merge base is **one added line**. (The reduction also removed a duplicated
`#3282` bullet the branch had picked up in a rebase.)

### Issue scope

#3294's original scope required unrecognised tokens to map above the floor, so
it no longer matched what ships. Rather than leave a PR closing an issue it only
half-addresses, #3294 has been **narrowed to stage 1a** — retitled, body
rewritten, an explicit note recording what was carved out and where it went,
milestone unchanged at 1.7.2 — and #3306 filed for stage 1b at 1.8.0. #3243 is
updated with the revised 1a / 1b / 2 / 3 staging table.

### Checks

`cargo fmt --all` clean; `cargo clippy --workspace --all-targets -- -D warnings`
exit 0; `SQLX_OFFLINE=true cargo build --lib --tests` clean; no `sqlx::query!`
macro text touched.

brandonrc:
## Adversarial review — pinned to `bfb2efd`

Brief was to refute. Every line cited below was read at `bfb2efd` (or `36f7d12` where stated), not from a working tree.

**Headline:** the code is right and the neutrality claim survives — I re-derived it rather than accepting it. What does not survive is the *release-notice* half. The PR asserts "no grade moves in this PR," and that is false: the one behaviour change it ships moves a repository's security grade by 25 points per finding, which is a **steeper** per-finding move than the deferred half whose grade-band upgrade note was removed on the grounds that it doesn't apply here.

---

### 1. CONFIRMED — "no grade moves in this PR" is false, and the shipped entry carries no upgrade note

PR body, CHANGELOG section: *"All of that describes the deferred half and does not ship here — no grade moves in this PR."* #3294's acceptance criteria repeat it: *"no grade-band upgrade note, because no grade moves."* Traced end to end at `bfb2efd`:

- `backend/src/services/openscap_scanner.rs:258` — `Severity::from_scanner_token(&f.severity)`; `critical` now yields `Severity::Critical`.
- `backend/src/services/scan_result_service.rs:1812-1849` — `recalculate_score`'s `latest_scans` CTE is `DISTINCT ON (sr.artifact_id, sr.scan_type)`, so **`openscap` is one of the competing scan types**, and the projection counts `scan_findings` with `FILTER (WHERE severity = 'critical' AND NOT is_acknowledged)`.
- `backend/src/services/scan_result_service.rs:143-156` — `compute_security_score`: 25 points per critical.
- `backend/src/services/scanner_service.rs:5049` — `.recalculate_score(artifact.repository_id)` runs after every scan, so this lands on the next scan with no operator action.

Arithmetic: **1** OpenSCAP `critical` → score 75 → grade B. **2** → 50 → C. **3** → 25 → D. **4** → 0 → F. The removed note's arithmetic (4→B, 9→C, 17→D, 26→F) is the 3-point `Medium` scale of the deferred half. The half that actually ships moves grades ~8x faster per finding; it was dropped because it was believed to move none.

The consumer enumeration in the PR body is also incomplete. It lists the score at item 5 as a scoring concern, but the score is a **gate input**:

- `backend/src/services/quality_check_service.rs:392-402` reads `repo_security_scores.score` as the `security` component of the composite, which is upserted to `artifact_health_scores` at `quality_check_service.rs:455-485`.
- `backend/src/services/promotion_rule_service.rs:447-449` — `check_min_health_score(min_score, score)` against `artifact_health_scores.health_score` (read at `:658-661`). That is a **live promotion gate** that appears nowhere in the enumeration.
- `backend/src/services/quality_check_service.rs:635-648` — `min_health_score` and `min_security_score` quality-gate conditions off the same row.

And the download path the PR does name is confirmed live: `policy_service.rs:266-290` puts `'critical'` in the severity array at **every** threshold while `info` appears in none, reached from `quarantine_service.rs:664 enforce_scan_policy_gate` via `enforce_download_gate`. So for a repository whose SCAP content emits `critical`, this patch can turn a succeeding download into a 403 and a passing promotion into a block.

None of that is wrong to fix — the direction is fail-closed and the fix is correct. The defect is that it ships with **no operator upgrade note**, in a `## [Unreleased]` section that already carries three `### Upgrade note —` headings for exactly this class of change, while a sibling PR is concurrently landing more of them. The volume argument (XCCDF 1.2 §6.4.4.2 has no `critical`) is sound for *standard* content, but `OPENSCAP_URL` (`backend/src/config.rs:1085`) points at an arbitrary sidecar, so "near-zero volume" is an assumption about deployments, not a property of the code — which is precisely when a one-paragraph note is cheap and its absence is not.

**Ask:** add an `### Upgrade note` covering (a) a repository whose OpenSCAP content emits `critical` may drop up to four grade bands on the next scan, (b) those findings now violate a `max_severity` policy at any threshold and can 403 a download, (c) they can fail a `min_health_score` / `min_security_score` gate. Correct the "no grade moves" line in the PR body and in #3294's acceptance criteria.

---

### 2. CONFIRMED — the CHANGELOG's `Negligible` claim is false for the Harbor path

Shipped bullet: *"`Negligible` … lands in the same bucket it always did (`Info`), so nothing about it changes behaviour."*

At `bfb2efd`, `backend/src/services/image_scanner.rs:170` — `normalize_harbor_severity` maps `"negligible" => "Low"` **before** the token ever reaches `convert_trivy_findings`. Pinned by the existing test at `image_scanner.rs:1167,1190`: a Harbor `"Negligible"` asserts `Severity::Low`.

So a Harbor-proxied `Negligible` is `Low`, before and after — not `Info`. No behaviour regression, but the operator-facing sentence is wrong in the direction that matters: a reader concludes `Negligible` violates nothing, when at `Low` it violates a `max_severity = 'low'` policy. The comment this PR *adds* at `image_scanner.rs:160-168` states the divergence correctly; the CHANGELOG contradicts it. Scope the sentence to the Grype/native-Trivy path.

---

### 3. CONFIRMED — there is a fourth scanner-token fallback, it lands at `Medium`, and it is unmentioned

PR body: *"`Severity::UNRECOGNIZED_SCANNER_SEVERITY` gives the 'what does an unrecognised token mean' decision a single named home … exactly where each adapter's own fallback put it."* #3294 Defect 3 lists three.

`backend/src/services/scanner_service.rs:3988` at `bfb2efd`:

```rust
let severity =
    Severity::from_str_loose(&advisory_match.severity).unwrap_or(Severity::Medium);
```

This is the OSV / GitHub-advisory dependency-scan path (`query_osv` / `query_github`, same fn). Its input is a feed-supplied severity token — a *scanner* token by the PR's own distinction, not an operator-configured threshold — and its private catch-all is `Medium`, not `Info`. It is the exact species of per-adapter fallback #3294 says it is eliminating, it is untouched, and it disagrees with `UNRECOGNIZED_SCANNER_SEVERITY` today.

No regression here. Two consequences worth fixing before merge:

- The "single named home" / "exactly where each adapter's own fallback put it" claims are not true of the tree. There are four fallbacks and two homes.
- #3306's stated premise — *"a one-symbol edit"* — is wrong. Repointing the constant to `Medium` leaves this call site as an independent, uncoordinated copy of the same decision, and #3243 stage 2/3 will still find the tree divergent. Either fold it in or state explicitly in #3306 that it is out of scope and why (my read: it is a *different* question — an advisory with no CVSS is not the same as a scanner that failed to grade — but that reasoning needs to be written down, not left as an unremarked inconsistency).

---

### 4. CONFIRMED (low) — the positive-control claim does not hold for one of the four neutrality pins

PR body: *"Each carries a positive control in the same fixture (a graded token that must still classify correctly), so none of them can be satisfied by a converter that stopped grading anything."*

`backend/src/models/security.rs:562-585`, `test_unrecognized_tokens_still_classify_at_the_floor`: the fixture is eight unrecognised tokens plus `assert_eq!(Severity::UNRECOGNIZED_SCANNER_SEVERITY, Severity::Info)`. That second assertion is a restatement of the constant, not a control. `from_scanner_token` rewritten as `|_| Severity::Info` passes this test in full.

The sibling `test_from_scanner_token_preserves_every_recognized_token` (`security.rs:503-525`) does catch that, so there is no real hole — but the methodology statement is wrong for one of the four, and this repo's failure taxonomy is exactly "a test that cannot fail in the way its doc comment claims." Either move a graded token into the fixture (one line) or drop the "each" from the body.

The other three pins were checked and **do** carry real in-fixture controls that can fail: `grype_scanner.rs:3313-3320` (`"Critical"` → `Critical`), `scanner_service.rs:11814` (`findings[3]` → `High`), `openscap_scanner.rs:11xx` / `test_convert_findings_xccdf_unknown_is_unchanged_pending_3306` (`findings[2]` → `High`).

---

### 5. Nit — double `to_lowercase()` allocation on the highest-traffic converter

`security.rs:145-150` lowercases, then hands the lowercased `other` to `from_str_loose`, which lowercases again (`security.rs:77`). Two `String` allocations per finding on `convert_trivy_findings`, the shared converter for four producers. Behaviourally inert (`to_lowercase` is idempotent), so this is genuinely a nit, but a `matches!`-first or a single binding would remove it.

---

## What I checked and found clean — the negative space

**The neutrality claim itself: holds, and I re-derived it rather than accepting the table.** At `bfb2efd`, `from_scanner_token` (`security.rs:145-150`) is `match token.to_lowercase() { "negligible" => Info, other => from_str_loose(other).unwrap_or(UNRECOGNIZED_SCANNER_SEVERITY) }`, and `from_str_loose` (`security.rs:76-85`, byte-identical to `36f7d12:69-78`) returns `None` for `negligible` — so the new arm is a no-op while the constant is `Info`. The prior Grype/Trivy expression was `from_str_loose(s).unwrap_or(Severity::Info)`. `to_lowercase` is idempotent, so the double-lowercase changes nothing. The two functions are equal on **every** input: `""`, `"   "`, `" high "`, `"UNKNOWN"`, `"Negligible"`, `"moderate"`, `"informational"`, `"none"`, `"very-high"`, `"sev:9"`. For OpenSCAP, the merge-base match's image is exactly `{High, Medium, Low, Info}` and the new classifier agrees on every one of those preimages; `critical` is the sole preimage whose image changes.

**The dropped `.trim()`: the author's reasoning is right, and I checked the thing that would have made it wrong.** Nothing upstream trims. `scripts/openscap-wrapper.py:231` is `severity = rule_result.get("severity") or meta.get("severity", "unknown")` — a raw XML attribute, no `.strip()`; the Grype and Trivy tokens are serde-deserialized JSON strings with no normalisation between deserialization and the converter. So no adapter previously trimmed anywhere else in the call chain, and adding `.trim()` would have moved `" high "` from `Info` to `High` — a second, unintended behaviour change. Correct call.

**Per-adapter merge-base fallback (hazard 2): `Info` in all three touched adapters.** Grype `36f7d12` and Trivy `36f7d12` both `unwrap_or(Severity::Info)`; OpenSCAP `_ => Severity::Info`. Verified. (The fourth, untouched one is finding 3.)

**Proxy/OCI inline gate is genuinely severity-independent.** `scanner_service.rs:2507-2509` — `is_vulnerable()` is `self.findings_count > 0`; `verdict_token()` at `:2512-2518` branches on it alone. `proxy_scan_service.rs` carries `max_severity` only as a stored/returned column (`:34`, `:269`, `:296-328`) and no decision reads it. `test_proxy_verdict_blocking_is_independent_of_severity` (`scanner_service.rs:10660-10710`) is well built: it asserts the fixture *actually* re-bucketed (`assert_ne!` on `max_severity`) before asserting invariance, has a blocking positive control and an empty-population negative control at the same call. That one can fail and would catch a real regression.

**`info_count` really is unprojected by the CVE-count gates** (`promotion_rule_service::check_cve_severity`, `promotion_policy_service` `LATEST_SCAN_COUNTS_SQL`). Claim holds.

**Reverted files are genuinely at merge-base.** `incus_scanner.rs` and `docker/scanner-adapter/harbor.go` are absent from the PR's file list against `36f7d12`, which is the strongest possible form of that claim.

**`image_scanner.rs` +17/-7 is comment-only.** Every added and removed line in that hunk is inside a doc comment or a `//` line; the sole executable line in range, `"negligible" => "Low".to_string()`, is context, not a change.

**Coverage mechanics (hazard 5): clean.** All new tests live in `#[cfg(test)] mod tests` inside `backend/src/models/security.rs`, `backend/src/services/{grype,openscap}_scanner.rs` and `backend/src/services/scanner_service.rs`. Nothing under `backend/tests/`. `cargo nextest run --workspace --lib` counts them.

**Revert-proof scope, stated precisely (hazard 6).** It proves the `critical` arm is gated and that nothing else in 15,126 tests observes a change. It cannot demonstrate Grype/Trivy neutrality — as the brief notes, if the consolidation is neutral no test *can* fail — so neutrality there is asserted by construction. I closed that by reading both functions rather than trusting the run; see above. The `AK_TESTS_REQUIRE_DB=1` + 0.27-0.48s timing discipline is the right guard and is the one that matters most on this repo.

**Two process notes, not defects:**
- **CI is not green.** At review time `🧪 Backend Unit Tests`, `📊 Code Coverage`, `🔗 Backend Integration Tests` and `🐳 Build Backend Image` are all `pending`. Nothing should be recorded as "passed full green CI" until they land.
- **#3312 collision: textual risk low, content risk real.** #3312 touches `CHANGELOG.md` only, at hunks anchored near old lines 7 / 15 / 31 (the `### Upgrade note` block); this PR inserts at old line 72 under `### Fixed`. Non-overlapping, so git should auto-merge either order. But #3312 is *"add missing 1.7.2 operator upgrade notes"* and adds three such headings — the note finding 1 asks for belongs in exactly that set. Whichever lands second should add it rather than assume the other did.

**#3294 vs the PR (context item):** the narrowed issue and the PR agree with each other cleanly — scope, out-of-scope, and acceptance all match what shipped. They agree on one wrong premise, though: #3294's final acceptance bullet, *"no grade-band upgrade note, because no grade moves,"* is the same claim finding 1 refutes. Both need the same correction.

brandonrc:
## Review findings addressed — rebased onto #3312, upgrade note restored with the correct arithmetic

All five items confirmed against source at `bfb2efd` before acting on them. Three
of the five were factual errors in my own description or CHANGELOG, and the first
one mattered: **I asserted no grade moves, and grades do move.**

Rebased onto `5405584` (#3312).

---

### 1. The `critical` arm moves grades — note restored, with the right numbers

Confirmed. `recalculate_score`'s `latest_scans` CTE is
`DISTINCT ON (sr.artifact_id, sr.scan_type)` (`scan_result_service.rs:1812-1849`),
so `openscap` is a competing scan type rather than one that replaces or is
replaced by Trivy/Grype — every OpenSCAP finding is counted. It runs
automatically after every scan (`scanner_service.rs`, "Recalculate repository
security score"), writes `repo_security_scores`, and `compute_security_score`
charges `Severity::Critical.penalty_weight()` = **25**. Against
`Grade::from_score`:

| OpenSCAP `critical` findings | score | grade |
|---|---|---|
| 1 | 75 | **B** |
| 2 | 50 | **C** |
| 3 | 25 | **D** |
| 4 | 0 | **F** |

The removed note's `4 → B, 9 → C, 17 → D, 26 → F` was the deferred half's
3-point `Medium` scale, so per finding this ships roughly **8x steeper**. The
offset is population, not magnitude — XCCDF 1.2 does not define `critical`, so
only custom or vendor-extended content emits it, where the deferred half fires on
the XCCDF *default*.

One thing that makes it sharper than the deferred half and was not in either
version of the note: **`critical` is a member of every `policy_service` severity
array**, so a single such finding violates a scan policy at *any* `max_severity`,
including the strictest, where as `info` it violated none.

The note is placed as the fourth item of #3312's collected *"smaller operational
changes worth two minutes before you deploy"* list (retitled seven → eight),
directly after the `scan_policies` sizing item it shares a table with, matching
that list's voice. It carries the mechanism, the band table, the any-threshold
policy point, the gates reached, an **"Am I affected?"** paragraph scoping it to
OpenSCAP deployments whose content declares a non-standard `severity="critical"`,
the fact that nothing is backfilled so it lands one repository at a time, and a
sizing query (`scan_type = 'openscap'` verified against `OpenScapScanner::scan_type`).

I agree with the framing: this is a correct fail-closed fix — a `critical`
compliance failure filed at the floor is a gate silently not firing — and the
error was shipping it silently, not shipping it.

### 2. Consumer list corrected — it was incomplete, which is what made it look safe

Confirmed; both omissions are live gates, and the chain behind them is
`recalculate_score` → `repo_security_scores.score` →
`quality_check_service.rs:392-402` (the `security` component) →
`compute_weighted_health_score` → `artifact_health_scores.health_score`:

* **`promotion_rule_service::check_min_health_score`** (`:447-449`) — a promotion
  rule's `min_health_score`.
* **`quality_check_service`** (`:635-648`) — `min_health_score` **and**
  `min_security_score`, the latter reading the security score directly with no
  weighting in between.

The list in the description is now six entries: the proxy/OCI inline gate
(neutral by construction), the download gate, both promotion CVE gates, and these
two. Five tighten for a `critical` finding, one is neutral, none relaxes. #3306
and the #3243 staging comment carry the same correction.

### 3. CHANGELOG `Negligible` claim — false for Harbor, fixed

Confirmed. `image_scanner::normalize_harbor_severity` rewrites `Negligible` to
`"Low"` before the token reaches the classifier (`image_scanner.rs:170`), pinned
by the pre-existing test at `:1190`. My bullet said it "lands in the same bucket
it always did (`Info`)", which is true for Grype and native Trivy and false for
Harbor — and wrong in the direction that matters, since `Low` violates a
`max_severity='low'` policy. It also contradicted the comment this PR itself adds
at `:160-168`. Both the CHANGELOG bullet and the description now state the split
explicitly and say the two `Negligible` mappings are left to #3243 stage 2.

### 4. There is a fourth fallback, at `Medium` — #3306's premise corrected

Confirmed: `scanner_service::scan_dependencies` (OSV/GitHub advisory ingestion)
is `from_str_loose(&advisory_match.severity).unwrap_or(Severity::Medium)`.

This does not change anything in this PR — the advisory path is a different
producer and folding it in would be a behaviour change rather than a
consolidation — but it falsifies "three fallbacks / a single named home", and
more importantly it falsifies **#3306's one-symbol-edit premise**, which is
load-bearing for how cheap the 1.8.0 work looks. Corrected in three places: the
section heading and text here, a paragraph on the constant's doc comment
(`UNRECOGNIZED_SCANNER_SEVERITY` now says it governs the adapters only and names
the advisory fallback), and a call-out block in #3306 requiring whoever repoints
the constant to decide explicitly whether the advisory path converges. Note the
direction this also implies: the advisory path is cited in #3306 as evidence that
`Medium` is the convention, so 1.8.0 would be the adapters *converging onto*
existing behaviour rather than cleaning up an inconsistency.

### 5. Neutrality pin had no working positive control

Confirmed, and it is the sharper version of the finding than "low": the test
asserted only that a set of tokens map to `Info` and then restated the constant,
so `from_scanner_token` rewritten as `|_| Severity::Info` satisfied it in full.
It now carries `critical/high/medium/low` positive controls in the same test, and
the restated-constant assertion is gone as redundant.

Proved rather than asserted — **M2**, `from_scanner_token` replaced by
`|_token| Severity::Info` on the otherwise-fixed tree:

```
FAIL [0.012s] models::security::tests::test_unrecognized_tokens_still_classify_at_the_floor
Summary [0.030s] 1 test run: 0 passed, 1 failed, 15135 skipped
```

---

## Revert-proof re-run on the rebased tree

Own Postgres (`ak-split3293-pg`, port 30693) re-migrated from scratch,
`DATABASE_URL` **and** `AK_TESTS_REQUIRE_DB=1` exported for every run.

```
Baseline (fixed tree):  Summary [215.444s] 15128 tests run: 15128 passed (2 slow), 8 skipped
R1 (three adapters reverted):
                        Summary [230.921s] 15128 tests run: 15127 passed (2 slow), 1 failed, 8 skipped
                        FAIL services::openscap_scanner::tests::test_convert_findings_openscap_critical_is_not_downgraded
```

Same result as before the rebase: **exactly one failure, and it is the `critical`
one.** That single run still does double duty — it gates the fix, and every other
test this PR adds passes against the pre-fix production code, which is what "no
other token moves" means operationally.

Faithfulness on the **reverted** tree (test modules stripped, compared to
`5405584`):

```
IDENTICAL    72035 ->   72035 bytes  backend/src/services/grype_scanner.rs
IDENTICAL   260155 ->  260155 bytes  backend/src/services/scanner_service.rs
IDENTICAL    12854 ->   12854 bytes  backend/src/services/openscap_scanner.rs
IDENTICAL    40423 ->   40423 bytes  backend/src/services/incus_scanner.rs
IDENTICAL     4830 ->    4830 bytes  docker/scanner-adapter/harbor.go
DIFFERS      40474 ->   41145 bytes  backend/src/services/image_scanner.rs   [non-comment production: IDENTICAL]
DIFFERS      10826 ->   16026 bytes  backend/src/models/security.rs
```

`cargo fmt --all` clean; `cargo clippy --workspace --all-targets -- -D warnings`
exit 0; `SQLX_OFFLINE=true cargo build --lib --tests` clean.

`image_scanner.rs` differs only in comment text — 7 merge-base lines changed, all
of them doc/inline comments, with the non-comment remainder byte-identical.
`models/security.rs` differs only by the retained classifier: **0 merge-base
production lines removed or changed, 85 added**, and its only mention outside
that file on the reverted tree is a doc comment in `image_scanner.rs:163` — no
call sites, so an uncalled function cannot make an adapter test fail.

Issue #3310, opened 2026-08-13
Two e2e suites still soft-fail after their blocking issues closed: cache-correctness (#1611) and concurrency (#1609)


brandonrc:
Two workflows soft-fail themselves with an in-comment promise to un-soften once a specific fix lands. Both fixes landed. Neither `continue-on-error` was removed, so both suites have been reporting green-regardless ever since.

## The two sites

**`.github/workflows/cache-correctness-e2e.yml:42`** — `continue-on-error: true` on the `🧊 Cache-correctness + virtual-resolution` job. The comment above it (lines 38-41) reads:

> Do NOT fail the workflow run if the suite is red on `main`: a red result is the EXPECTED outcome until #1611 lands (it proves the bugs reproduce). The per-step logs carry the PASS/FAIL detail. Once #1611 lands, remove this `continue-on-error` so regressions hard-fail.

#1611 closed 2026-06-06.

**`.github/workflows/concurrency-e2e.yml:68`** — `continue-on-error: true` on the single-flight race job. The comment (lines 65-67) reads:

> On `main` this job is EXPECTED to fail (it reproduces #1606). Do not let a red result mark the whole workflow failed in a way that pages people; the purpose is observability, not gating. Once #1609 lands, remove this.

#1609 closed 2026-07-03; #1606 closed 2026-07-03.

## Why this matters beyond the two files

Both of these were deliberate, well-reasoned, correctly-documented decisions at the time they were made. The failure is that a soft-fail justified by a *temporary* condition has no mechanism attached to it that notices when the condition ends. The comment is the only record, and comments do not expire.

That is the same shape as the `resilience-tests` problem being fixed in artifact-keeper-test: a suite listed as if it gates something, which cannot actually report failure. The difference is only that these two are off the release path, so the blast radius is narrower — these are observability suites, not release gates.

## Requested fix

Do **not** simply delete the two lines. Establish first whether the suites actually pass now:

1. Pull the recent run history for both workflows and read the per-step PASS/FAIL detail the comments refer to. Sample enough runs to distinguish "passes" from "passes sometimes".
2. If a suite is green and stable, remove its `continue-on-error` so regressions hard-fail — which is what the comment always intended.
[...]

Issue #3309, opened 2026-08-13
release-preflight check 1's superset rule makes .trivyignore entries un-removable: 9 dead suppressions are stuck on main


brandonrc:
## Summary

`scripts/ci/release-preflight.sh` check 1 requires main's `.trivyignore` suppression set to be a **superset** of every active `release/*` branch's, and hard-fails otherwise. That is correct for its purpose — it stops a hotfix suppression being lost on the way to main, which is the #3039 gap. But it has a side effect nobody chose: **a suppression can never be removed from main while any release branch still lists it.** The set only ever grows.

That is the wrong direction for a security gate. A `.trivyignore` entry is a documented exception to a security control; letting them accumulate irreversibly means the file drifts from "the residual set" toward "everything we have ever had to work around", and the gate quietly weakens as it goes.

## How it bit us

Found while fixing #3307. Two groups of entries on main are now provably dead:

**1. `GHSA-hrxh-6v49-42gf`** — suppressed because grype v0.116.0 shipped grpc-go v1.80.0. The #3307 bump to grype v0.117.0 ships grpc-go **v1.82.1**, the fixed version. A scan of `ghcr.io/anchore/grype:v0.117.0` does not report it at any severity.

**2. The entire trivy block** — `CVE-2026-48978`, `CVE-2026-50151`, `CVE-2026-50162`, `GHSA-vh4v-2xq2-g5cg`, `CVE-2026-56852`, `CVE-2026-46600`, `CVE-2026-49834`, `CVE-2026-49835`. All written against `ghcr.io/aquasecurity/trivy:0.72.0`, which the scanner-adapter stopped shipping at #3123. It now ships `ghcr.io/artifact-keeper/trivy`. Scanning the exact pinned digest with no ignorefile and all severities: **not one of them appears in the image.** They are not suppressed findings, they are absent findings.

Nine dead tokens. Deleting them is the obviously right change, and it fails the pre-tag check, because `release/1.5.x` and `release/1.6.x` still carry them:

```
[FAIL] main is MISSING suppressions that release/1.6.x has:
         - CVE-2026-46600
         - CVE-2026-56852
         - GHSA-hrxh-6v49-42gf
         ...
```

So #3307 marks them `RETIRED` in comments and leaves the tokens in place. That works, but it means the file's real content now lives in its comments — a reader has to parse prose to know which of the 19 tokens are live (10) and which are inert (9). That is not a durable state.

## Why the rule over-reaches

[...]

Issue #3039, opened 2026-07-30
security(images): forward-port trivy-CLI CVE suppressions to main + bump 1.7.0-rc.2


brandonrc:
The first 1.7.0 release candidate (v1.7.0-rc.1) stalled at docker-publish's **Security Scan** gate: the scanner-adapter image vendors the trivy 0.72.0 CLI, which carries `CVE-2026-46600` (golang.org/x/net) and `CVE-2026-56852` (golang.org/x/text). These were suppressed on `release/1.6.x` for v1.6.4 (#2997) with a documented vendored-binary-scope justification, but were **never forward-ported to `main`**.

Because Security Scan is a hard gate that the multi-arch manifest jobs depend on, a failure there skips the manifest → `release.yml`'s `resolve-candidate-digest` has no manifest-list digest → the entire release gate chain (E2E, DTF, verify-images) cannot start.

Fix: forward-port the two `.trivyignore` entries (verbatim, incl. #2997's justification comments) to `main`, and bump the version set `1.7.0-rc.1` → `1.7.0-rc.2` to re-cut the RC off fixed main.

The suppressions are an already-reviewed accepted exception (vendored-binary scope; no fixed trivy tool release exists); this is a faithful forward-port, not a new exception.

PR #3315, opened 2026-08-13
fix(ci): clear go-git CVE-2026-71556 from the bundled scanner CLIs, and make a measured-red Docker Publish blocking

### LINKED ISSUE #3307: Docker Publish on main fails Security Scan: go-git CVE-2026-71556 (HIGH) in bundled grype and trivy CLIs -- blocks the v1.7.2 cut
## Summary

`Docker Publish` on `main` is red, and it blocks the v1.7.2 cut. Both `Trivy gate - Backend` and `Trivy gate - Scanner Adapter` fail on the **same** advisory: **CVE-2026-71556** (HIGH, CVSS 7.1) in `github.com/go-git/go-git/v5` v5.19.1, fixed in **5.19.2**.

`Security Scan` hard-gates the manifest jobs, so the failure cascades. A tag cut now stalls exactly the way `v1.7.0-rc.1` did in #3039.

## Evidence

Run [31657226105](https://github.com/artifact-keeper/artifact-keeper/actions/runs/31657226105), head `7255a23c` (current main), job `94316458479`:

| Step | Result |
|---|---|
| Trivy gate - Backend | **failure** |
| Trivy gate - OpenSCAP | success |
| Trivy gate - Scanner Adapter | **failure** |

Backend image, `usr/local/bin/grype` (gobinary):

```
Total: 1 (HIGH: 1, CRITICAL: 0)
github.com/go-git/go-git/v5  CVE-2026-71556  HIGH  fixed  v5.19.1  ->  5.19.2
```

Scanner-adapter image, `usr/local/bin/trivy` (gobinary): byte-for-byte the same finding.

Everything else in both images is clean. The backend's RHEL 9.8 layer reports 0; the scanner-adapter's alpine 3.24.1 layer and its own `scanner-adapter` binary both report 0.

Skipped as a consequence: `Backend Multi-Arch Manifest`, `Backend Alpine Multi-Arch Manifest`, `Scanner Adapter Multi-Arch Manifest`, `OpenSCAP Multi-Arch Manifest`, `Verify Published Tags`, `Backend Alpine`.

Two things in that log look like findings and are not, recording them so the next person does not chase them: the `digest-mismatch: error` line is an echoed **input parameter** of `actions/download-artifact`, and the block of CVE identifiers printed mid-log is the contents of `.trivyignore` being echoed by `trivy-action`, not results.

## The advisories

| CVE | Severity | CVSS | What |
|---|---|---|---|
| [CVE-2026-71556](https://github.com/advisories/GHSA-hc8v-wwc9-vgxm) | HIGH | 7.1 | Worktree operations (checkout, status, add) resolve symlinks without confining resolution to the worktree boundary, so a crafted repository can make go-git read or write outside the intended working directory. |
| [CVE-2026-71557](https://github.com/advisories/GHSA-qgq7-7hm3-q39j) | MEDIUM | 6.3 | Reference names are not sanitised before being joined into on-disk paths under the reference-storage directory, so a crafted ref name writes outside it. |

Both fixed in go-git 5.19.2. Only the HIGH gates; the MEDIUM clears with it.

## Classification

**A genuine new CVE with a real upstream fix.** Not a trivy-DB flake, not an expired suppression, and explicitly **not** a `.trivyignore` candidate — a fixed version exists for both bundled tools, so the file's own rule ("only vendored-scanner-binary CVEs that have no fixed tool release are listed") excludes it.

Worth recording the shape, because it will recur: **nothing regressed.** Neither image changed. `7255a23c` is a repositories fix that touches no Dockerfile. The advisory was published against a dependency the images already shipped, days after they last scanned clean. The scanner-adapter's pinned trivy image was published gate-clean on 2026-08-07 and started failing on 2026-08-13 without a byte changing. A green scan is a statement about a moment, not a property of an image.

## Remediation

Real fixes on both sides, no suppression.

**Backend** (`docker/Dockerfile.backend`, `docker/Dockerfile.backend.alpine`) — bump `ghcr.io/anchore/grype` v0.116.0 to **v0.117.0** (released 2026-08-10), whose `go.mod` pins go-git v5.19.2. Verified with the gate's exact configuration:

```
$ trivy image --severity CRITICAL,HIGH --ignore-unfixed --exit-code 1 \
      --ignorefile .trivyignore ghcr.io/anchore/grype:v0.117.0
grype (gobinary)   0
exit status 0
```

v0.117.0 also moves grpc-go 1.80.0 to 1.82.1, which retires the `GHSA-hrxh-6v49-42gf` suppression.

**Scanner adapter** (`docker/Dockerfile.scanner-adapter`) — no upstream trivy release carries the fix: v0.73.0 is still the latest and pins go-git v5.19.1. This is exactly the case `artifact-keeper/trivy` was built for, and all three of its override-policy conditions hold (the CVE gates our image, upstream `main` has already made the bump, no upstream release carries it). Add the go-git override there, publish `v0.73.0-r2`, and re-pin the digest here.

Tracked in artifact-keeper/trivy#3.

## Also in scope

`.trivyignore` has drifted from what the images actually ship. Its header still describes the scanner-adapter as carrying `ghcr.io/aquasecurity/trivy:0.72.0`, but since #3123 it carries our own v0.73.0 build. Scanning the pinned digest shows every trivy-attributed entry in the file — the oras-go group, x/text, x/net, the sigstore pair — is absent from the image. Those suppressions are dead and should be removed with this bump, per the file's own "re-evaluate this whole file on every scanner bump".

### LINKED ISSUE #3308: release-preflight.sh reports READY while warning that a cut will stall: a measured-red Docker Publish must be blocking, not advisory
## Summary

`scripts/ci/release-preflight.sh` measures the exact condition it was written to catch, and then throws the answer away. On current `main` it prints both of these, in this order:

```
[warn] main's last Docker Publish did NOT cleanly publish the manifest
       -- a tag cut will likely stall the same way. Investigate before tagging.

READY: no blocking problems. Safe to cut the tag.
```

Those two statements contradict each other and the warn is the truthful half. `Docker Publish` run 31657226105 on `main` failed `Security Scan` and skipped every manifest job (#3307). A cut would stall.

## The line

`scripts/ci/release-preflight.sh:156`:

```bash
warn "main's last Docker Publish did NOT cleanly publish the manifest -- a tag cut will likely stall the same way. Investigate before tagging."
```

`warn()` (line 59) prints and returns. `bad()` (line 58) prints **and** increments `problems`. Only `problems` reaches the verdict block at lines 163-168, so the exit status is 0 and the script says READY.

## Why this is the wrong call

**1. This is the motivating incident, verbatim.** The script's own header (lines 7-14) says it exists because of #3039, where `v1.7.0-rc.1` stalled ~90 minutes on `Security Scan` and *"the signal was already visible (`Docker Publish [push/main]` had been red for 5 commits); nothing checked it before tagging. This gate is that check."* It now performs the check, gets the red answer, and exits 0. The 90 minutes are still on the table.

**2. The hard check cannot catch this class.** Check 1 (trivyignore drift) is a *proxy* for "will Security Scan pass" and only detects one cause: a suppression on `release/*` not forward-ported to `main`. Check 3 is the *direct measurement*. Right now the proxy is hard and the measurement is soft, which is backwards — and this incident proves the gap is reachable. #3307 is a newly-published advisory affecting both branches equally, so there is no drift and check 1 passes clean. Every hard check passes while the release is definitively blocked.

**3. The comment calling it "advisory" is load-bearing in the wrong direction.** Line 26 labels check 3 "best-effort/advisory" because `gh` may be missing or unauthenticated. That justifies softness for *not knowing*. It does not justify softness for *knowing the answer is red*. The two are currently the same code path.

**4. It is the pattern the release-freeze policy exists to remove.** The freeze revoked "gate failure is cosmetic, promote anyway" after it produced the v1.5.8 integrity failure. A preflight that prints a red measurement and then blesses the cut is that same reasoning wearing a different hat — worse, because it hands over an explicit "Safe to cut the tag."

## Proposed fix

Split check 3's three outcomes, which are currently collapsed into one:

| Outcome | Meaning | Today | Should be |
|---|---|---|---|
| `PREFLIGHT_SKIP_DOCKER_HEALTH=1`, or `gh` absent | deliberately not measured | note, exit 0 | unchanged |
| `gh` present but the query failed (line 146) | could not measure | warn, exit 0 | INFRA, exit 2 -- matches the documented contract for "tooling/network failure; retryable, NOT a readiness verdict" |
| Measured, and Security Scan or the manifest is not `success` (line 156) | **measured red** | warn, exit 0 | `bad`, exit 1 |

Only the third row changes behaviour for a healthy `main`. Concretely, line 156 becomes `bad` and the exit-code contract at lines 31-36 gains a sentence saying a measured-red Docker Publish is blocking.

Keep an escape hatch, but make it explicit and separate from "the probe did not run" — e.g. `PREFLIGHT_ALLOW_RED_PUBLISH=1` — so overriding it is a recorded decision by a human rather than the default. `PREFLIGHT_SKIP_DOCKER_HEALTH=1` should not be that hatch: "do not look" and "I looked, it is red, proceed anyway" deserve different names in the transcript.

## Acceptance

- With `main`'s latest Docker Publish red, the script exits 1 and prints NOT READY.
- With it green, the script exits 0 and prints READY (no change).
- With `gh` unavailable or the query failing, the script exits 2 with an INFRA verdict, not a readiness one.
- A test covers the measured-red path, so a future refactor cannot quietly turn it back into a warn.

Found while diagnosing #3307.

brandonrc:
Closes #3307
Closes #3308

## The blocker

`Docker Publish` on main (run [31657226105](https://github.com/artifact-keeper/artifact-keeper/actions/runs/31657226105), head `7255a23c`) fails `Security Scan`. Both `Trivy gate - Backend` and `Trivy gate - Scanner Adapter` fail on the **same** finding:

```
github.com/go-git/go-git/v5  CVE-2026-71556  HIGH  fixed  v5.19.1  ->  5.19.2
```

Backend carries it in `usr/local/bin/grype`; scanner-adapter in `usr/local/bin/trivy`. Security Scan hard-gates every multi-arch manifest, so a v1.7.2 cut stalls the way v1.7.0-rc.1 did (#3039).

**Nothing regressed.** Neither image changed — `7255a23c` touches no Dockerfile. [GHSA-hc8v-wwc9-vgxm](https://github.com/advisories/GHSA-hc8v-wwc9-vgxm) was published against a dependency both images already shipped. A clean scan is a statement about a moment, not a property of an image.

## Version landscape

| Tool | First release carrying go-git 5.19.2 | Evidence |
|---|---|---|
| grype | **v0.117.0** (2026-08-10) | `go.mod` at v0.117.0 pins v5.19.2. v0.116.1 and v0.116.0 are both still on **v5.19.1**. |
| trivy | **none exists** | v0.73.0 (2026-08-03, latest) pins v5.19.1. `main` pins v5.19.2, unreleased. |

**Dependabot #3031 is not this fix.** It bumps grype v0.116.0 → v0.116.1, which is still on the vulnerable go-git. Measured, not inferred — see the table below. It should be closed or rebased onto v0.117.0.

## What changed

**Backend** (`Dockerfile.backend`, `Dockerfile.backend.alpine`) — grype v0.116.0 → **v0.117.0**. The comment blocks claiming "v0.116.0, the latest upstream release" and "no fixed grype release yet" were stale and are rewritten; the alpine file's comment had also drifted to `v0.115.0` while its pin said v0.116.0.

**Scanner adapter** (`Dockerfile.scanner-adapter`) — re-pinned to `artifact-keeper/trivy` **0.73.0-r2** (`sha256:d6c56ef9…`), which adds a go-git v5.19.2 override on top of the existing oras-go one. Built and published via artifact-keeper/trivy#3.

Since no fixed upstream trivy exists, `.trivyignore`'s own rule ("only vendored-scanner-binary CVEs that have **no** fixed *tool release* are listed") would technically have permitted a suppression here. It is still the wrong answer, and the Dockerfile now says why: a fixed *dependency* exists, we can build against it, and an override that `assert-buildinfo.sh` proves is linked into the shipped binary is a stronger claim than a comment asserting the code path is unreachable. **Suppress when there is no fix; override when there is one and only the packaging is missing.**

## Verification — measured, not inferred

Every row is the gate's exact configuration: `--severity CRITICAL,HIGH --ignore-unfixed --exit-code 1 --ignorefile .trivyignore`.

| Target | Gate exit | Detail |
|---|---|---|
| `grype:v0.116.0` (control) | **1** | reproduces the CI failure exactly |
| `grype:v0.116.1` (#3031's target) | **1** | **still v5.19.1** — this is why #3031 is not the fix |
| `grype:v0.117.0` | **0** | clean |
| `artifact-keeper/trivy@sha256:202c…` (old pin) | **1** | the finding was its *only* gating result |
| `artifact-keeper/trivy@sha256:d6c5…` (new pin) | **0** | redhat 0, `usr/local/bin/trivy` 0 |
| **scanner-adapter built from this branch** | **0** | alpine 0, `scanner-adapter` 0, `trivy` 0 |

Beyond the version string, the shipped binary's own build info:

```
dep  github.com/go-git/go-git/v5  v5.19.2  h1:wkfn7vOlUBu8ivAWKBWisTiwJK4jYHzTF8Ndv1LyGqY=
```

### No new red traded in

The only HIGH+fixed findings remaining in grype v0.117.0 are `CVE-2026-27145`, `CVE-2026-39822`, `CVE-2026-42504` (stdlib, still go1.26.3) and `CVE-2026-34040` (docker, still v28.5.2) — all already suppressed, none new. The residual set is otherwise unfixed or below HIGH.

### Output shape is unchanged

The backend parses grype `-o json` into `GrypeReport`. Same lodash 4.17.4 fixture through v0.116.0 and v0.117.0: **10 matches each**, identical `vulnerability.{id,severity,fix}`, `artifact.{name,version,type,purl}`, and `relatedVulnerabilities` still **top-level on the match** — the exact trap #1375 fixed.

### The adapter still works

Built from this branch, it starts, probes `trivy version 0.73.0`, downloads its DB, serves `/api/v1/metadata`, and completes real scans over HTTP:

- scanning `grype:v0.116.0` through it → **14 findings, including CVE-2026-71556 (High) and CVE-2026-71557 (Medium)**
- scanning `grype:v0.117.0` through it → **11 findings, zero go-git**

The first is the positive control. Without it, "0 go-git findings" is equally consistent with a broken scanner returning nothing; with it, the clean result is real.

## .trivyignore

Re-measured and re-annotated per the file's own "re-evaluate this whole file on every scanner bump".

Nine tokens are now **provably dead**: `GHSA-hrxh-6v49-42gf` (grype v0.117.0 ships grpc-go v1.82.1, the fixed version) and the entire trivy block, which was written against `ghcr.io/aquasecurity/trivy:0.72.0` — an image we stopped shipping at #3123. Scanning the pinned digest with no ignorefile and all severities finds **none** of them. They are absent findings, not suppressed ones.

They are marked **RETIRED** rather than deleted. `release-preflight.sh` check 1 hard-fails when main is not a superset of every `release/*` branch, and 1.5.x/1.6.x still list them, so deleting them here breaks the pre-tag check. The suppression token set is therefore **byte-identical** to before (19 tokens) and check 1 still passes. Actually removing them is #3309.

## Also: the preflight soft-fail (#3308)

`release-preflight.sh` measured the red Docker Publish and then printed:

```
[warn] main's last Docker Publish did NOT cleanly publish the manifest
       -- a tag cut will likely stall the same way.
READY: no blocking problems. Safe to cut the tag.
```

Line 156 called `warn` (which does not increment `problems`) where it needed `bad`. It now blocks. Three outcomes are kept deliberately distinct, because *"I did not look"*, *"I could not look"* and *"I looked and it is red"* are different claims:

| Outcome | Verdict |
|---|---|
| not measured (skipped / `gh` absent) | note, not blocking |
| could not measure (query failed) | **INFRA, exit 2** — not a readiness verdict |
| measured red | **exit 1, NOT READY** |

`PREFLIGHT_ALLOW_RED_PUBLISH=1` is a separately-named human override, deliberately not the same switch as `PREFLIGHT_SKIP_DOCKER_HEALTH`.

Worth noting *why* the hard checks did not catch this: check 1 is a **proxy** for "will Security Scan pass" and detects exactly one cause — a suppression stranded on `release/*`. Check 3 is the **direct measurement**. Having the proxy hard and the measurement soft was backwards, and this incident is the proof: #3307 affects main and the release branches equally, so there was no drift, and check 1 passed clean while the release was definitively blocked.

Live run of the fixed script against the real red main:

```
1) .trivyignore forward-port drift ...
[ok]   trivyignore drift: main is a superset of every release/*
2) version-set consistency ...
[ok]   version set is consistent at 1.7.1
3) latest main Docker Publish health
  run 31657226105: Security Scan=failure, Backend Multi-Arch Manifest=skipped
[FAIL] main's last Docker Publish did NOT cleanly publish the manifest -- a tag cut WILL stall on the same gate.

NOT READY: 1 blocking problem(s). Fix on main before tagging.
```

### The test is revert-proved

`scripts/ci/test-release-preflight.sh` covers all five cases against a stubbed `gh`, wired into `release-preflight.yml` ahead of the real check. Restoring the single `bad` to `warn` — the exact pre-fix shape — reproduces the original contradictory output and turns **two** cases red, while the **green-path case still passes**. That asymmetry is the point: it shows the test isolates the measured-red property rather than merely detecting that the script runs.

## Follow-ups filed

- #3309 — check 1's superset rule makes `.trivyignore` entries un-removable; the nine dead tokens above are stuck on main until it lands.
- artifact-keeper/trivy — `verify-image.sh` still asserts only a single hardcoded `OVERRIDE_MODULE`. Not a coverage hole (`assert-buildinfo.sh` iterates `overrides.yaml` generically and did verify go-git), but worth generalising before a third override.

Adjacent, for cross-reference: #3310 on `continue-on-error` outliving its justification is the same species of soft-fail as #3308.

## Notes

No Rust or Cargo changes — `cargo build` / `cargo fmt` are not applicable to this diff.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -557,6 +557,20 @@ jobs:
       - name: Check Trivy gate severity filter
         run: ./scripts/ci/check-trivy-gate-severity.sh
 
+      # Same species as the check above, one layer out (#3308). That one asks
+      # "does the publish gate apply the filter it declares?"; this one asks
+      # "does the PRE-TAG gate block on what it just measured?" It did not:
+      # release-preflight.sh printed "a tag cut will likely stall" and
+      # "READY: ... Safe to cut the tag." in the same run, because the
+      # measured-red branch called `warn` (non-incrementing) instead of `bad`.
+      #
+      # It needs a test rather than a live run because the interesting case is
+      # invisible whenever main is green: with a green Docker Publish the buggy
+      # and fixed scripts are byte-identical in behaviour. Stubs `gh`, so no
+      # network and no dependence on main's current health. ~1s.
+      - name: Check release-preflight blocks on a measured red
+        run: bash scripts/ci/test-release-preflight.sh
+
   coverage:
     name: 📊 Code Coverage
     runs-on: ak-ci-runners
--- a/.github/workflows/release-preflight.yml
+++ b/.github/workflows/release-preflight.yml
@@ -35,6 +35,14 @@ jobs:
           ref: main
           fetch-depth: 1
 
+      # Runs first, and offline. If check 3 has regressed into a soft warn
+      # again (#3308), this fails here rather than silently blessing a cut
+      # below. The interesting case -- a MEASURED-RED Docker Publish -- is
+      # invisible whenever main happens to be green, so it has to be tested
+      # against stubs rather than waited for.
+      - name: Self-test the preflight gate
+        run: bash scripts/ci/test-release-preflight.sh
+
       - name: Run release preflight
         env:
           GH_TOKEN: ${{ github.token }}
--- a/.trivyignore
+++ b/.trivyignore
@@ -4,9 +4,10 @@
 #
 # * Every entry below is a Go-dependency CVE compiled INTO one of the vendored
 #   scanner CLIs that AK images ship and invoke:
-#     - /usr/local/bin/grype  (ghcr.io/anchore/grype:v0.116.0, backend images)
-#     - /usr/local/bin/trivy  (ghcr.io/aquasecurity/trivy:0.72.0,
-#       scanner-adapter image only — the backend images do NOT bundle trivy)
+#     - /usr/local/bin/grype  (ghcr.io/anchore/grype:v0.117.0, backend images)
+#     - /usr/local/bin/trivy  (ghcr.io/artifact-keeper/trivy, pinned by digest
+#       in docker/Dockerfile.scanner-adapter — scanner-adapter image only; the
+#       backend images do NOT bundle trivy)
 #   Both are short-lived CLI subprocesses the services shell out to. They are
 #   NOT part of the network-reachable service surface (axum HTTP :8080 / gRPC
 #   :9090 / adapter HTTP :8080). The vulnerable code paths (docker client
@@ -23,12 +24,31 @@
 #   every build), or in the UBI/Alpine base OS packages. Only
 #   vendored-scanner-binary CVEs that have no fixed *tool release* are listed.
 #
-# * Both tools are pinned to the latest upstream release (grype v0.116.0,
-#   trivy 0.72.0, bumped for #2391). For the CVEs below the dependency fix
-#   exists upstream but no tagged tool release has been rebuilt against it
-#   yet (or, where noted, no upstream fix exists at all). They drop off
-#   automatically when the tools ship a release rebuilt against the fixed
-#   dependency — re-evaluate this whole file on every scanner bump.
+# * grype is pinned to the latest upstream release (v0.117.0, bumped for #3307).
+#   trivy is no longer taken from Aqua's release stream at all: since #3123 the
+#   scanner-adapter ships ghcr.io/artifact-keeper/trivy, our own build of the
+#   pinned upstream tag with declared dependency overrides. That means a trivy
+#   Go-dependency CVE with a fixed dependency version is now FIXABLE by us and
+#   must NOT be added here — file it as an override in that repo's
+#   overrides.yaml instead. This file is for findings with no fix available in
+#   the tool we can actually build.
+#
+# * For the CVEs below the dependency fix exists upstream but no tagged grype
+#   release has been rebuilt against it yet (or, where noted, no upstream fix
+#   exists at all). They drop off when grype ships a release rebuilt against
+#   the fixed dependency — re-evaluate this whole file on every scanner bump.
+#
+# * ⚠ ENTRIES HERE CANNOT CURRENTLY BE DELETED, even once they are provably
+#   dead. scripts/ci/release-preflight.sh check 1 is a HARD gate that requires
+#   main's suppression set to be a SUPERSET of every active release/* branch's.
+#   release/1.5.x and release/1.6.x still carry the older set, so removing a
+#   retired token from main makes the pre-tag check fail with "main is MISSING
+#   suppressions". The rule was written to stop a hotfix suppression being lost
+#   on the way to main (#3039), and for that it is right — but it also ratchets
+#   the set monotonically upward, which is the wrong direction for a security
+#   gate. Entries that measurement shows are no longer present in any shipped
+#   image are therefore marked RETIRED below and kept in place rather than
+#   removed. Do not read a RETIRED entry as an active exception. See #3309.
 #
 # * The CI publish GATE scans with `ignore-unfixed: true` and
 #   `severity: CRITICAL,HIGH`, so the genuinely no-fix-available and
@@ -47,14 +67,19 @@
 #   there without gating.
 #
 # Grype tracker: https://github.com/anchore/grype/releases
-# Trivy tracker: ghcr.io/aquasecurity/trivy tags (GitHub repo has no releases)
+# Trivy tracker: https://github.com/artifact-keeper/trivy — overrides.yaml is
+#   the live record of what we diverge from upstream on, and its
+#   upstream-watch workflow tracks new Aqua releases.
 
 # ---------------------------------------------------------------------------
-# grype v0.116.0 binary — Go stdlib @ go1.26.3
-# Fixed in Go 1.26.4/1.26.5 (1.25.11/1.25.12). grype v0.116.0 (latest) is
-# still built with go1.26.3. Drops off when grype ships a release built on
-# Go >= 1.26.5. (CVE-2026-39822 / CVE-2026-42505 also cover the trivy 0.72.0
-# binary, which is built with go1.26.4.)
+# grype v0.117.0 binary — Go stdlib @ go1.26.3
+# Fixed in Go 1.26.4/1.26.5 (1.25.11/1.25.12). grype v0.117.0 (latest) is
+# still built with go1.26.3 — the v0.116.0 -> v0.117.0 bump for #3307 did not
+# move the toolchain. Drops off when grype ships a release built on Go >=
+# 1.26.5. Re-measured against ghcr.io/anchore/grype:v0.117.0 on 2026-08-13:
+# all five still present, CVE-2026-27145 / -39822 / -42504 at HIGH+fixed (i.e.
+# these three are the ones actually holding the gate open), -42505 / -42507 at
+# MEDIUM.
 CVE-2026-27145
 CVE-2026-39822
 CVE-2026-42504
@@ -62,10 +87,12 @@ CVE-2026-42505
 CVE-2026-42507
 
 # ---------------------------------------------------------------------------
-# grype v0.116.0 binary — github.com/docker/docker @ v28.5.2+incompatible
+# grype v0.117.0 binary — github.com/docker/docker @ v28.5.2+incompatible
 # CVE-2026-41567 / -41568 / -42306: NO upstream-fixed docker release exists yet
 # (FixedVersion=none). CVE-2026-33997 / -34040 are fixed in docker 29.3.1 but
-# grype v0.116.0 (latest) has not rebuilt against it. Code path: grype's
+# grype v0.117.0 (latest) has not rebuilt against it — still v28.5.2. Of these,
+# only CVE-2026-34040 is HIGH+fixed and therefore actually gating; the rest are
+# unfixed (dropped by --ignore-unfixed) or MEDIUM. Code path: grype's
 # Docker engine client for `docker:` image sources; the backend invokes grype
 # against on-disk artifacts/SBOMs, not the Docker daemon.
 CVE-2026-33997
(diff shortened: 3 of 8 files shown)


brandonrc:
Independent verification of both pins, at the **binary** level rather than from `go.mod` — pulled each image by the digest/tag this PR lands on, extracted the scanner binary, and read the linked module version out of the Go build info embedded in it:

| Binary | Source | Linked go-git |
|---|---|---|
| `grype` | `ghcr.io/anchore/grype:v0.117.0` (new pin) | `go-git/go-git/v5@v5.19.2` ✅ |
| `grype` | `ghcr.io/anchore/grype:v0.116.0` (old pin) | `go-git/go-git/v5@v5.19.1` |
| `trivy` | `ghcr.io/artifact-keeper/trivy@sha256:d6c56ef9…` (new pin) | `go-git/go-git/v5@v5.19.2` ✅ |

The old-pin row is the control. Without it, a green result only shows the new images are clean; with it, the bump is demonstrably both necessary and sufficient rather than coincidental.

This matters most for the scanner-adapter half. There, the fix is a rebuild with a dependency override rather than a version bump, so `go.mod` saying `v5.19.2` and the *shipped* binary linking `v5.19.2` are genuinely separate claims — an override can be declared and then not take effect. The second is the one Trivy actually scans, and it holds. That is consistent with what `assert-buildinfo.sh` asserts in the build pipeline; this is the same property confirmed from outside it, against the published artifact.

Also confirming the version landscape independently, since it drives the two different remedies: grype v0.117.0 (2026-08-10) is the first upstream release carrying the fix — v0.116.1 and v0.116.0 are both still on v5.19.1, so dependabot #3031 is genuinely not this fix. For trivy, upstream v0.73.0 (2026-08-03, latest) still pins v5.19.1, so no fixed release exists to bump to and the override is the only available remedy, not a shortcut around an easier one.

brandonrc:
## End-to-end verification: full Docker Publish, green

Triggered `Docker Publish` on this branch via `workflow_dispatch` — the same workflow, the same runners, the same gate that is red on main. [Run 31662197998](https://github.com/artifact-keeper/artifact-keeper/actions/runs/31662197998): **success**.

The two previously-failing gate steps, in `Security Scan` job `94330960540`:

| Step | main (run 31657226105) | this branch |
|---|---|---|
| Trivy gate - Backend | **failure** | **success** |
| Trivy gate - OpenSCAP | success | success |
| Trivy gate - Scanner Adapter | **failure** | **success** |

And the whole cascade that was skipped as collateral:

| Job | main | this branch |
|---|---|---|
| Security Scan | **failure** | **success** |
| Backend Multi-Arch Manifest | skipped | **success** |
| OpenSCAP Multi-Arch Manifest | skipped | **success** |
| Scanner Adapter Multi-Arch Manifest | skipped | **success** |
| Publish Summary | skipped | **success** |

`Backend Alpine` and its manifest remain skipped — unrelated, the alpine build is suspended in the workflow (`if: false`). `Verify Published Tags` is skipped because this is a candidate build, not a version tag.

This is the claim that actually matters, and it is now measured rather than argued from component scans: a v1.7.2 cut off this tree does not stall at Security Scan.

### Interpretation note

Run 31662197998 builds all three images from source on this branch and gates the *resulting* images, so it also settles the composition question the per-component scans left open — that the bumped grype and re-pinned trivy do not interact badly with the UBI/alpine base layers, and that no new CRITICAL/HIGH appears once assembled.

Issue #2290, opened 2026-07-07
OCI /v2/token exchange does not enforce API-token scope/allowed_repo_ids (backstopped by downstream per-repo checks)


brandonrc:
Follow-up from #2276 review (non-blocking, pre-existing). The `GET /v2/token` handler mints a full-identity JWT from the authenticated principal's privileges and does NOT narrow the token to the API token's declared scope / `allowed_repo_ids` (there's a standing TODO in oci_v2.rs ~3827-3838). It is currently backstopped by per-repository authorization at every resource handler (`require_oci_repo_write_access` on push/upload, visibility + `enforce_scan_pull_scope` on pull), so it is not exploitable for access the principal doesn't already have — but a restricted API token can obtain a broader-than-declared OCI bearer. Harden the exchange to intersect the minted token's effective scope with the API token's `allowed_repo_ids` (ties into the AccessScope work from epic #1617). Target 1.5.

brandonrc:
✅ **Validated / reproduced** on v1.4.0 (main `7a1eb22`).
Repro: API token restricted to repo A → `GET /v2/token?scope=repository:<repoB>:pull,push` → **200**, mints a full-identity JWT with no `allowed_repo_ids`/`scopes` claim (broader than the source token). Backstop holds: downstream per-repo push/pull gates still 403.
Root cause: `handlers/oci_v2.rs:3834` — explicit TODO; only warns on restricted scopes, then `generate_tokens(&user)` mints unrestricted. `validate_api_token` already returns scopes+allowed_repo_ids.
Fix size: **SMALL** — use `generate_tokens_with_repo_scope(&user, validation.allowed_repo_ids.as_allowed_repo_ids())` (`auth_service.rs:1130`).
Severity **LOW** (defense-in-depth; no privilege gain today) — issue targets 1.5 hardening; consider deferring past 1.4.1. No open PR. (automated validation)

Issue #3269, opened 2026-08-10
OCI /v2 read-authz residual: GET /v2/_catalog enumerates every repository key to any authenticated caller


brandonrc:
Residual from #3261 (PR #3268), which closed the OCI read-authorization hole on
`GET`/`HEAD` blob, `GET`/`HEAD` manifest and `GET tags/list`.

### 1. `GET /v2/_catalog` enumerates every repository to any authenticated caller

`handle_catalog` (`backend/src/api/handlers/oci_v2.rs`) authenticates and then
returns `catalog_local_entries(...)` unfiltered, so any account — including one
with no grant anywhere — receives the full list of repository keys, private ones
included. Unlike the manifest/blob handlers this cannot be fixed with a gate: the
decision is per row, so it needs a visibility filter pushed into
`catalog_local_entries`' query (public OR the caller holds a rule/role assignment,
admin sees all), and the `n`/`last` pagination has to stay correct once rows are
filtered out.

Impact is name enumeration, not content disclosure — after #3268 the caller still
cannot pull anything it lists.

---

### Folded into #3268 — no longer residuals

Two items originally filed here were resolved in #3268 itself, after review found
that neither was safe to defer.

**`handle_tags_list` does not apply the token repo-scope ceiling** — shipped in
#3268. It was worse than "the ceiling is missing": #3268 turned `scan_pull_repo`
into an authorization *grant*, and `handle_tags_list` was the one call site that
reached that grant without running `enforce_scan_pull_scope` first, so a scan
token pinned to repository A listed repository B's private tags with `200` while
the manifest verb correctly answered `403`. Both ceilings (the scan pin and the
#2290 `allowed_repo_ids` allow-list) are now enforced inside
`require_oci_repo_read_access`, so no call site can omit them.

**Virtual repositories authorize on the virtual, not the resolving member** —
shipped in #3268. The claim above that this is "the same shape the
native-protocol path has" was wrong: `proxy_helpers::authorize_virtual_members`
exists for exactly this and is called by `maven.rs`, `npm.rs`, `pypi.rs`,
`debian.rs` and `goproxy.rs`; `oci_v2.rs` was the only format handler that never
called it. Through a **public** Virtual parent the read gate is satisfied by
`public_read_satisfies_acl` and never runs, so an **anonymous** caller received a
private member's manifest and layer bytes with `200` while the direct pull of
that same private repository correctly `404`'d. That is the #3261 vulnerability
[...]

PR #3317, opened 2026-08-13
fix(oci): confine directly-presented scoped API tokens to their repositories

### LINKED ISSUE #3316: OCI /v2 direct-credential path discards API-token repository scope (allowed_repo_ids): scoped token reads/writes any repo the owner can reach
## Summary

A repository-scoped API token's `allowed_repo_ids` allow-list is **silently discarded** when the token is presented **directly** to the `/v2/` OCI endpoints (Docker `Bearer` slot and Basic-auth password slot). The repository-scope ceiling (`enforce_token_repo_scope`, the gate #2290/#3261 rely on) becomes a no-op on exactly the credential CI uses most, so a token scoped to repository A can **read and write any other repository the owning identity is otherwise authorized for**.

The action-scope ceiling (read-only vs write) is *not* affected — it travels through a separate channel and still bites. Only the **repository** ceiling is lost.

Found during a customer evaluation / hardening pass.

## Root cause

`authenticate_oci_with_scopes` (`backend/src/api/handlers/oci_v2.rs`) has two branches that accept a raw API token directly:

- Bearer slot (`docker login --password-stdin` with an API token), ~line 374–385
- Basic password slot (service accounts / CI / curl), ~line 402–413

Both re-mint OCI claims with:

```rust
let claims = auth_service
    .generate_tokens(&validation.user)   // <-- allowed_repo_ids: None, scopes: None
    ...
return Ok((claims, Some(scopes)));
```

`generate_tokens` → `generate_tokens_with_family_and_scope(user, .., None, None)`, so the minted `Claims.allowed_repo_ids` is `None`. The read gate (`require_oci_repo_read_access` → `enforce_token_repo_scope`) and the write gate (`require_oci_repo_write_access` → `enforce_token_repo_scope`) both treat `None` as **unrestricted**. `validation.allowed_repo_ids` (the real ceiling, an `AccessScope::Restricted(..)` allow-list) is thrown away.

Contrast the intended `/v2/token` exchange path (same file, ~line 4396–4407), which correctly threads it:

```rust
let allowed_repo_ids: Option<Vec<Uuid>> = validation.allowed_repo_ids.clone().into();
let token_scopes = Some(validation.scopes.clone());
let tokens = auth_service.generate_tokens_with_scope(&user, token_scopes, allowed_repo_ids)?;
```

`#2290` and `#2430` fixed the **exchange** path; `#3261` added the read gate. None of them covers the **direct-credential** path, which never goes through `/v2/token`.

## Pre-existing vs. new

**Pre-existing — NOT a 1.7.2 regression.** Verified live against the `1.7.1` release image: the same repository-scoped token writes and reads a non-scoped repository through the direct-credential path. Both `authenticate_oci_with_scopes` (mints via `generate_tokens`) and the write gate (`require_oci_repo_write_access` → `enforce_token_repo_scope`) existed at `v1.7.1`. What 1.7.2 (#3261) added is the read-side gate; the ceiling is structurally present there but reads empty claims on the direct path, so the 1.7.2 changelog statement that "the token repo-scope ceiling runs as the first act of the read gate ahead of the admin bypass" is **optimistic for direct credentials** — it holds only for tokens laundered through `/v2/token`.

## Live reproduction (pinned to main HEAD `7255a23c075f594929bbece6a825ebef6345e0b4`, built from source)

Setup: user `ci.scoped` (non-admin) granted read+write on private docker repos `scope-repo-a` and `scope-repo-b`; token minted via `POST /api/v1/repositories/scope-repo-a/tokens` (scoped to A only); both repos seeded with an image.

Negative control (token used on its own repo A — works):
```
GET /v2/scope-repo-a/manifests/v1   Authorization: Bearer <A-scoped-token>   -> 200
GET /v2/scope-repo-a/tags/list      Authorization: Bearer <A-scoped-token>   -> 200 {"name":"scope-repo-a","tags":["v1"]}
```

Direct-credential path, A-scoped token against **B** (should be 403; all return 200):
```
# Bearer slot
GET  /v2/scope-repo-b/tags/list       Bearer <tok> -> 200 {"name":"scope-repo-b","tags":["v1"]}
GET  /v2/scope-repo-b/manifests/v1    Bearer <tok> -> 200 (manifest)
HEAD /v2/scope-repo-b/manifests/v1    Bearer <tok> -> 200
GET  /v2/scope-repo-b/blobs/<digest>  Bearer <tok> -> 200 (config blob)
HEAD /v2/scope-repo-b/blobs/<digest>  Bearer <tok> -> 200
# Basic password slot (curl -u ci.scoped:<tok>) — identical results
GET  /v2/scope-repo-b/manifests/v1    -u ci.scoped:<tok>  -> 200
# username is ignored (token identifies the user):
GET  /v2/scope-repo-b/manifests/v1    -u nobody:<tok>     -> 200
```

Write path, A-scoped token against **B** (should be 403; all succeed):
```
POST /v2/scope-repo-b/blobs/uploads/         Bearer <tok> -> 202 (Location: .../uploads/<uuid>)
POST /v2/scope-repo-b/blobs/uploads/         -u ci.scoped:<tok> -> 202
PUT  /v2/scope-repo-b/manifests/pwned-by-a   Bearer <tok> -> 201 Created
GET  /v2/scope-repo-b/tags/list  (admin)     -> {"name":"scope-repo-b","tags":["pwned-by-scoped-a","v1"]}
```

Positive control — the intended `/v2/token` exchange DOES enforce the ceiling:
```
GET /v2/token?service=artifact-keeper&scope=repository:scope-repo-b:pull  -u ci.scoped:<A-scoped-token>
  -> minted JWT: allowed_repo_ids=["<A-id>"], scopes=["read:artifacts","write:artifacts"]
GET /v2/scope-repo-b/manifests/v1  Authorization: Bearer <minted-jwt>  -> 403 {"errors":[{"code":"DENIED","message":"You do not have access to this repository"}]}
GET /v2/scope-repo-a/manifests/v1  Authorization: Bearer <minted-jwt>  -> 200
```

Action-scope differential (only the repo ceiling is lost, not the action ceiling): a **read-only** A-scoped token presented directly:
```
GET  /v2/scope-repo-a/tags/list      -> 200   (own repo, read)
POST /v2/scope-repo-a/blobs/uploads/ -> 403 {"code":"DENIED","message":"Token does not have required scope: write:artifacts"}   (action-scope enforced)
GET  /v2/scope-repo-b/tags/list      -> 200   (repo-scope discarded — the bug)
```

## Impact / severity: HIGH

A repository-scoped token — the standard containment primitive for CI service accounts and mirrors — provides **no repository containment** when used the normal way (`docker login`, `-u user:token`, raw Bearer). It can read and write every repository the owning identity holds any grant on. For a broadly-granted CI account this is effectively repo-wide read + write from a credential the operator believes is pinned to one repo. Not anonymous→admin, so HIGH rather than CRITICAL; but it defeats a control the product advertises and that #2290/#3261 were built to provide.

## Proposed fix

Thread the validated ceiling into the minted claims on both direct-credential branches of `authenticate_oci_with_scopes`, mirroring the `/v2/token` path:

```rust
let allowed_repo_ids: Option<Vec<Uuid>> = validation.allowed_repo_ids.clone().into();
let claims = auth_service
    .generate_tokens_with_repo_scope(&validation.user, allowed_repo_ids)
    .and_then(|tokens| auth_service.validate_access_token(&tokens.access_token))?;
```

Notes:
- On the OCI surface the **action** scope is already preserved via the returned `Option<Vec<String>>` tuple element (`oci_scopes_grant`), so `Claims.scopes: None` does not currently void the action ceiling *here*. Using `generate_tokens_with_scope(&user, Some(validation.scopes.clone()), allowed_repo_ids)` instead is preferable anyway for defense-in-depth and parity with the exchange path, so the minted `Claims` is self-consistent if any future code path reads `claims.scopes` directly.
- Add a regression test that a raw API token scoped to repo A is denied (403 DENIED) on both the read (`GET manifest/tags/blob`) and write (`POST blob upload` / `PUT manifest`) handlers of repo B, presented in both the Bearer and Basic slots — matching the existing `enforce_token_repo_scope` unit tests but exercised end-to-end through `authenticate_oci_with_scopes`.

brandonrc:
## Summary

A repository-scoped API token's `allowed_repo_ids` allow-list was silently discarded on the OCI `/v2/` **direct-credential** path, so `enforce_token_repo_scope` — the ceiling the read gate (#3261) and write gate (#2290) rely on as their first act — was a no-op there. A token scoped to repository A could read and write any other repository the owning identity could reach, when the token was presented the normal way: `docker login`, `-u user:token`, or a raw `Authorization: Bearer`.

Plain text: Fixes #3316.

Root cause is isolated to `authenticate_oci_with_scopes` in `backend/src/api/handlers/oci_v2.rs`. Its two raw-API-token branches (Bearer slot and Basic password slot) re-minted claims through `generate_tokens(&user)` — i.e. `generate_tokens_with_family_and_scope(user, .., None, None)` — so the minted `Claims.allowed_repo_ids` was `None`, which `enforce_token_repo_scope` reads as unrestricted. `validation.allowed_repo_ids` (the real `AccessScope` ceiling) was thrown away. The intended `/v2/token` exchange path already threads it via `generate_tokens_with_scope`; the direct path never goes through `/v2/token`.

The action-scope ceiling was **not** lost — it travels through the second tuple element (`oci_scopes_grant`), not `claims.scopes`, so the direct path still enforced read-only-vs-write. Only the repository ceiling was lost. This is confirmed differentially below: a read-only token still cannot write to its own repository.

### The change

Each of the three direct-credential arms now threads **only the repository ceiling** onto the minted claims, via `generate_tokens_with_repo_scope` (which sets `Claims.scopes = None`), mirroring what the exchange path threads for the repository dimension:

```rust
let allowed_repo_ids: Option<Vec<Uuid>> = validation.allowed_repo_ids.clone().into();
generate_tokens_with_repo_scope(&validation.user, allowed_repo_ids)
```

`AccessScope::Admin` (an unscoped token) maps to `None` = unrestricted, preserving prior behaviour for full-scope tokens; `AccessScope::Restricted(vec![])` maps to `Some(vec![])` = deny-all, so a token whose allow-list resolves to nothing denies everything rather than falling open. That inversion is covered by an explicit test.

**Deliberately NOT threading `Claims.scopes`.** An earlier revision passed `Some(scopes)` for parity with the exchange path. That is a fail-closed *policy change* riding on this fix, not part of it, and it must be separated. `virtual_caller_auth` (`oci_v2.rs`) converts these claims through `From<Claims> for AuthExtension` (`auth.rs`), which ends in `.with_scope_gated_admin()` → `is_admin = is_admin && has_scope("admin")`. Setting `Claims.scopes = Some(["read:artifacts","write:artifacts"])` therefore **demotes a global admin** whose token is action-scoped but carries no repository restriction: on a virtual repository the member set drops from `All` to `User(id)`, every member is filtered, and the request returns `404` where merge-base returned `200`. That end state may well be desirable (it matches `/v2/token`), but it hits tokens *outside* this issue's blast radius (no repo scope), surfaces as a 404 rather than the 403 this fix promises, and would contradict — three times over (PR, issue, CHANGELOG) — the claim that this change cannot affect unscoped tokens. It is filed as a standalone decision for 1.8.0 (issue #3322) carrying the demotion analysis, rather than shipped unremarked here.

A sibling in the same function is fixed too: the JWT-as-Docker-password fallback (`validate_access_token_async` arm) re-minted via `generate_tokens`, dropping the presented JWT's repository ceiling. A JWT that was itself exchanged from a scoped token could therefore be laundered through the password slot to reach outside its scope — a bypass of the raw-token fix. The Bearer-JWT arm already returns the original claims and keeps the ceiling; this arm now threads the repository ceiling explicitly (again `Claims.scopes` left `None`). An interactive/keyless JWT carries `None` and stays unrestricted, so the CI/CD keyless push flow is unchanged. The presented JWT's `scan_pull_repo` pin is not carried onto the re-mint (as with the pre-existing `generate_tokens` here); it is inert — scan tokens reach OCI via the Bearer slot, not as a Docker password, and a dropped pin fails closed — and is documented in a comment.

I checked the other `generate_tokens(&user)` call sites, and specifically went looking for a fourth discarded-ceiling site: the Conan and OCI `/v2/token` exchanges thread both ceilings; the legacy refresh path threads `token_data.claims.allowed_repo_ids`; the auth middleware never re-mints. The remaining production callers in this file are the username/password login arm (a full login — unrestricted is correct) and unit-test fixtures; `totp.rs` and `guest_access.rs` are login flows with no token ceiling to preserve. No other discarded-ceiling site exists.

### Upgrade note

A CHANGELOG entry is added under `[Unreleased]` → Security with an "am I affected" framing. This is a **fail-closed behaviour change**: a CI pipeline whose scoped token has been reaching repositories outside its declared scope will start receiving a denial on those repositories after upgrading. The note states the denial *shape* explicitly: a direct (Local/Remote) repository outside scope answers `403 DENIED`; a **Virtual** repository answers `404`, because a token scoped to the virtual clears the parent gate and then has each member re-checked against its scope and dropped (`member_passes_token_scope`), so the parent resolves as if empty. Both are refusals; only the status code differs.

Five DB-backed `#[cfg(test)]` tests in `backend/src/api/handlers/oci_v2.rs` (module `direct_credential_repo_scope_regression_3316`) drive `authenticate_oci_with_scopes` end to end and assert the minted claims carry the ceiling and that `enforce_token_repo_scope` (the first act of both the read and write gates) denies the out-of-scope repository:

- `bearer_slot_confines_scoped_token_to_its_repos` — Bearer slot; A allowed, B denied; also asserts `Claims.scopes == None` (pins the no-demotion decision above).
- `basic_password_slot_confines_scoped_token_to_its_repos` — Basic password slot; username ignored; A allowed, B denied.
- `deny_all_repo_selector_denies_every_repo` — `Restricted(vec![])` → `Some(vec![])`, denies every repo (guards against the fail-open inversion).
- `jwt_as_password_preserves_repo_ceiling` — the sibling; an exchanged JWT presented as the password stays confined.
- `unscoped_token_stays_unrestricted` — an unscoped (`AccessScope::Admin`) token maps to `Claims.allowed_repo_ids == None` and passes on every repository, through the real mint path. This pins the non-regression side: if someone later "hardens" the `Admin → Option` conversion to `Some(vec![])` (plausible after reading the deny-by-default `Default` on `AccessScope`), every unscoped token would be locked out of every repository and none of the restricted-token tests would catch it. This test correctly passes on both merge-base and head — it guards a future inversion, not the current bug.

These are DB-backed and were run under `AK_TESTS_REQUIRE_DB=1` (0.3–5s each, not the ~0.04s skip). Reverting the production change against the merge-base and re-running makes the four bug-pinning tests fail (`left: None`), including the deny-all case; `unscoped_token_stays_unrestricted` correctly still passes (its guard is for a future change). The reverted tree differs from the merge-base only by the added test module (313 insertions, insertions only — the production code is byte-identical to merge-base).

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc:
## Adversarial review — pinned to `d0f94dc2158a3e6edc041fb5dc993a706da1fbf9`

Every file quoted below was fetched from `raw.githubusercontent.com` at that SHA and byte-compared against the checkout, so none of this is working-tree drift.

The core of the fix is correct. The `AccessScope` conversion is not inverted, the sweep of `generate_tokens` really is complete, every read and write verb is covered, and the four tests are real and will actually execute in CI. Details of what I checked and cleared are at the bottom — I'd rather that be read than skimmed.

But **the `Some(scopes.clone())` half of the change is not redundant on this surface**, and the PR body, the issue and the CHANGELOG all state that it is. That claim is load-bearing for the "this only fails closed for out-of-scope repositories" framing of the upgrade note, and it is wrong.

---

### F1 — `Some(scopes.clone())` demotes global admins on Virtual OCI reads. Confirmed. Medium.

The PR body says:

> The action-scope ceiling was **not** affected — it travels through the second tuple element (`oci_scopes_grant`), not `claims.scopes`, so `Claims.scopes: None` was harmless.

`claims.scopes` is read on this path. `oci_v2.rs:3160-3166`:

```rust
fn virtual_caller_auth(
    claims: Option<&crate::services::auth_service::Claims>,
) -> Option<crate::api::middleware::auth::AuthExtension> {
    claims
        .cloned()
        .map(crate::api::middleware::auth::AuthExtension::from)
}
```

called at `oci_v2.rs:4816` (`handle_head_blob`), `5137` (`handle_get_blob`), `7621` (`handle_head_manifest`) and `8730` (`handle_get_manifest`) — all four on the `repo.repo_type == RepositoryType::Virtual` branch, all four fed the claims these two changed arms now mint.

`From<Claims> for AuthExtension` (`backend/src/api/middleware/auth.rs:269-293`) ends with:

```rust
        scopes: claims.scopes,
        allowed_repo_ids: AccessScope::from(claims.allowed_repo_ids),
        iat_ms,
    }
    .with_scope_gated_admin()
```

and `with_scope_gated_admin` (`auth.rs:232-235`) is:

```rust
self.is_admin = self.is_admin && self.has_scope("admin");
```

`has_scope` (`auth.rs:111-120`) returns `true` unconditionally for `None`, and otherwise delegates to `scopes_grant_access`, which short-circuits only on `*` or `admin` (both in `ALLOWED_SCOPES`, `token_service.rs:91-104`, but neither is a default).

So the before/after on the two changed arms:

| | `claims.scopes` | `AuthExtension.is_admin` for a global admin | `member_grant_visibility` |
|---|---|---|---|
| merge-base `5405584` | `None` | `true` | `RepoVisibility::All` |
| `d0f94dc2` | `Some(["read:artifacts","write:artifacts"])` | **`false`** | `RepoVisibility::User(id)` |

`member_grant_visibility` is `repositories.rs:330-336`. From there `authorize_virtual_members` (`proxy_helpers.rs:3119-3153`) narrows the member set through `filter_visible_repo_ids`, and a dropped member is treated as not holding the artifact.

**Failure scenario.** Admin `ops` holds an API token with scopes `["read:artifacts","write:artifacts"]` and **no repository restriction at all** (`AccessScope::Admin`). `docker-virtual` is a Virtual repo whose members were created by other users, so `ops` holds no `role_assignments` row and no `permissions` row on them.

```
curl -u ops:$TOKEN https://ak/v2/docker-virtual/manifests/v1
```

At `5405584`: `200`, manifest served (admin ⇒ `RepoVisibility::All`).
At `d0f94dc2`: every member is filtered out of the walk, `resolve_virtual_manifest` returns `None`, and the caller gets `404 MANIFEST_UNKNOWN`. Same for blob `GET`/`HEAD`.

Three things make this worth blocking on rather than waving through:

1. It hits tokens with **no repository scope**, i.e. entirely outside the blast radius the upgrade note describes ("a repository-scoped token…"). An operator reading the note has no reason to think an unscoped admin token is affected.
2. The failure is `404 MANIFEST_UNKNOWN`, not the `403 DENIED` the note promises. That reads as "the image isn't there", which is the worst possible diagnostic for a post-upgrade CI break.
3. The demotion is not what this PR set out to do. It rides in on a line the author describes as parity/defence-in-depth and the issue explicitly assessed as inert.

The direction is fail-closed and it matches what `/v2/token` already does, so I am **not** arguing the end state is wrong — the standard `docker` CLI does the `/v2/token` handshake and already gets this demotion, so only direct-credential clients (`curl -u`, raw `Authorization: Bearer`, mirrors) see a change. But shipping it as an unremarked side effect of a patch, on top of a claim in three places that it cannot happen, is the problem. Either drop `Some(scopes.clone())` from the two raw-token arms (the tuple element already carries the action ceiling — the PR is right that it is redundant *for `oci_scopes_grant`*, just not for `virtual_caller_auth`), or keep it and say so in the upgrade note.

### F2 — The upgrade note's symptom is wrong for Virtual repos. Confirmed. Low/Medium.

Independently of F1, the repository ceiling itself does not produce `403 DENIED` in the Virtual case. `member_passes_token_scope` (`repositories.rs:356-367`):

```rust
Some(a) => member_is_public || a.can_access_repo(member_id),
```

with the deliberate note at `repositories.rs:353-355` — *"a token scoped only to a virtual sees an empty listing"*.

So a token scoped to virtual `V` clears `enforce_token_repo_scope(claims, V.id)` at the gate (`oci_v2.rs:746`), then loses every private member in the walk, and the pull answers `404 MANIFEST_UNKNOWN`. A token scoped to member `A` instead fails at the parent with `403`. Either way a Virtual pull that worked before now needs the token scoped to **both** parent and members.

Again: parity with `/v2/token`, so not new behaviour for that path — but new for the direct path this PR is about, and the note's "will start receiving `403 DENIED` on those repositories" does not describe it. Worth one sentence: *scope the token to the virtual repository **and** its members; a member outside the scope answers `404`, not `403`.*

### F3 — Pre-existing, adjacent, and I think worth its own issue: `tags_list_virtual` has no member authorization at all. Confirmed. Medium/High.

Not introduced here, but it directly contradicts what #3261 and this PR's CHANGELOG entry claim about the Virtual surface, so it should not stay unfiled.

`tags_list_virtual` (`oci_v2.rs:9783-9829`) takes no `auth` parameter. It calls `proxy_helpers::fetch_virtual_members` and merges tags from **every** member — no `authorize_virtual_members`, no `member_passes_token_scope`, nothing. It is reached unconditionally via `resolve_tags_page` → `resolve_virtual_tags_page` (`oci_v2.rs:9295-9354`).

And `handle_tags_list` skips its gate entirely for anonymous callers (`oci_v2.rs:9253-9257`):

```rust
    if let Some(claims) = &claims {
        if let Err(resp) = require_oci_repo_read_access(state, claims, &repo).await {
            return resp;
        }
    }
```

with `claims == None` whenever `is_anonymous_token(headers)`, and the earlier anonymous guard only rejecting when `!repo.is_public`.

**Scenario:** `docker-virtual` is public and has a private member `docker-internal`. `curl https://ak/v2/docker-virtual/tags/list` **with no credential** returns the merged tag list, including every tag of `docker-internal`.

This is the same shape as the private-member laundering #3261 closed for `resolve_virtual_manifest` and `resolve_virtual_blob` — the third virtual walker was missed. Content bytes are still protected (those two do filter); this is tag-name and version disclosure, which for private release repos is not nothing. The fix is the same one-liner shape the other two got: thread `virtual_caller_auth(claims.as_ref())` into `resolve_tags_page` and filter through `authorize_virtual_members`.

### F4 — No test pins `AccessScope::Admin → None` through the production path. Confirmed. Low (test gap).

All four new tests use a *restricted* token. `access_scope.rs:130-186` covers the conversion in isolation, but nothing asserts that `authenticate_oci_with_scopes` yields `allowed_repo_ids: None` for an **unscoped** token.

The entire risk of this fix is the direction of that conversion. If a later edit "hardened" `AccessScope::Admin` to `Some(vec![])` — exactly the kind of change someone makes after reading the deny-by-default `Default` impl at `access_scope.rs:71-80` — every unscoped API token would be locked out of every repository on `/v2/`, and **all four new tests would still pass**. That is a total-outage regression with no guard.

One more test closes it: mint a token, pin no repos, assert `claims.allowed_repo_ids == None` and `enforce_token_repo_scope(&claims, Uuid::new_v4()).is_ok()`.

### F5 — The JWT-as-password arm still drops `scan_pull_repo`. Confirmed. Informational.

`oci_v2.rs:465-495` now threads `scopes` and `allowed_repo_ids`, but `generate_tokens_with_scope` hard-codes `scan_pull_repo: None` (`auth_service.rs:1319`). The `/v2/token` exchange treats this as important enough to branch on (`oci_v2.rs:4322-4340`, re-issuing via `generate_scan_token` precisely so the pin is not re-widened); this arm does not.

Currently inert: the scanner presents `Bearer <scan_token>` (`image_scanner.rs:570-577`), which lands in the Bearer-JWT arm that returns the original claims untouched. And if a scan token ever *did* arrive in the password slot, dropping the pin makes it fail **closed** (`_ak_scanner` holds no grants, so `check_repository_action` denies) rather than open. So: no action required, but the PR's own thesis for this arm is "it re-mints, so thread the ceiling explicitly" — and it threaded two of the three ceilings. Worth a comment noting the third is deliberately left, or a `scan_pull_repo` passthrough for symmetry.

---

## Negative space — what I checked and found clean

**The conversion, which is the whole fix.** `From<AccessScope> for Option<Vec<Uuid>>` (`access_scope.rs:111-118`) is `Admin => None`, `Restricted(ids) => Some(ids)`. `Restricted(vec![])` maps to `Some(vec![])`; there is no collapse to `None`. `enforce_token_repo_scope` (`oci_v2.rs:580-587`) is `Some(ids) if !ids.contains(&repo_id) => Err`, and `![].contains(x)` is `true`, so deny-all genuinely denies every repository including one that does not exist. No inversion. The runtime path and the test agree.

**The producer side.** `validate_api_token` (`auth_service.rs:2276-2308`) yields exactly three states and the round-trip is faithful: no join-table rows or an empty selector → `None` → `Admin`; a non-empty selector resolving to zero repos → `Some(vec![])` → `Restricted(vec![])`; an explicit list → `Some(list)`. So `Admin` on the wire really does mean "no restriction was ever declared", not "restriction lost".

**The `generate_tokens` sweep — independently verified, and I confirm it.** Every remaining `.generate_tokens(&user)` in the workspace: production callers are `oci_v2.rs:450` (username/password login arm — a full login, unrestricted is right), `totp.rs:512` (TOTP login completion), `auth_service.rs:1216` and `:2562` (both inside `authenticate`). Everything else is `#[cfg(test)]` — `guest_access.rs:577`, `auth.rs:5940`/`6127`, `oci_v2.rs:14565`/`16285`/`16942`/`17236`/`24886`, `test_db_helpers.rs:929`, and the `auth_service.rs` test module. No further site re-mints from a presented credential.

I specifically went looking for other exchange endpoints, since that is where the sibling was found. The Conan exchange (`conan.rs:367-372`) already threads both ceilings off the presented `AuthExtension`. The legacy refresh path (`auth_service.rs:1806-1810`) threads `claims.scopes` and `claims.allowed_repo_ids`. `authenticate_federated_with_scope` (`auth_service.rs:2602`) uses `generate_tokens_with_repo_scope`. The middleware path (`auth.rs:856-881`) never re-mints — it copies `validation.allowed_repo_ids` straight onto the `AuthExtension`. No fourth discarded-ceiling site exists.

**Route coverage — all eleven verbs.** The ceiling rides on the claims, so the `authenticate_oci` wrapper discarding the tuple is harmless. Reads: `handle_head_blob` (gate at `4748`), `handle_get_blob` (`4939`), `handle_head_manifest` (`7517`), `handle_get_manifest` (`8587`), `handle_tags_list` (`9254`). Writes/delete: `handle_start_upload` (`5342`), `handle_patch_upload` (`5776`), `handle_cancel_upload` (`6021`), `handle_complete_upload` (`6262`), `handle_put_manifest` (`9000`), `handle_delete_manifest` (`10078`). Both gates run `enforce_token_repo_scope` as their first act (`oci_v2.rs:624` and `:746`).

I also checked the cross-repo blob mount, which is the natural way to sidestep a per-repo ceiling: `try_mount_blob` runs `enforce_token_repo_scope(claims, source.id)` on the **source** repo at `oci_v2.rs:5255`, so a token scoped to A cannot mount B's blobs into A. Covered. `GET /v2/_catalog` (`oci_v2.rs:9883-9895`) authenticates but runs no gate — that is the known #3269 residual, unchanged here.

**The ceiling still caps admins.** `require_oci_repo_read_access` runs `enforce_scan_pull_scope` and `enforce_token_repo_scope` at `oci_v2.rs:745-746`, *before* the `if claims.is_admin || claims.scan_pull_repo.is_some()` bypass at `:748`. The write gate does the same at `:624`. This survived.

**The scanner is untouched.** `_ak_scanner` presents `Bearer <scan_token>` (`image_scanner.rs:574`), which hits the first Bearer arm — `validate_access_token_async` succeeds and the original claims are returned verbatim, with `scan_pull_repo` intact and `allowed_repo_ids: None` (`generate_scan_token`, `auth_service.rs:1371-1398`). Neither changed branch is on its path. Private-image scanning is unaffected.

**Unscoped tokens still work.** `AccessScope::Admin → None`, and `enforce_token_repo_scope` is a no-op on `None`. The only behavioural delta for an unscoped token is F1.

**`/v2/token` is untouched.** The diff contains three production hunks, all inside `authenticate_oci_with_scopes`; the exchange path at `oci_v2.rs:4408-4460` is byte-identical to merge-base.

**The tests are real and will run.** They are in `backend/src/api/handlers/oci_v2.rs:30819+`, so they count for `cargo nextest run --workspace --lib`. CI sets both `DATABASE_URL` and `AK_TESTS_REQUIRE_DB: "1"` on **Backend Unit Tests** (`ci.yml:505-524`) and on **Code Coverage** (`ci.yml:697-725`), so `try_pool` → `require_db_url` → `enforce_db_available` (`testing.rs:59-80`) panics rather than silently skipping. No fiction-green.

I checked each of the four named failure classes against the fixtures:
- *Precondition missing.* `create_repo` returns `(Uuid, String, PathBuf)` (`test_db_helpers.rs:695-720`), so `repo_a` really is the repository UUID being asserted. `generate_api_token` returns `(String, Uuid)` (`auth_service.rs:2329-2335`) and `["read:artifacts","write:artifacts"]` passes `validate_scopes_pure`. The join-table insert is the same store `validate_api_token` reads when `repo_selector` is null.
- *Silently-dropped fixture.* The deny-all test's `{"match_repos": [...]}` is a real field (`repo_selector_service.rs:34`). Had the name been wrong, `serde_json::from_value(...).unwrap_or_default()` would have produced an empty selector → `is_empty` → `None` → unrestricted, and the test would have **failed loudly** on `Some(vec![])`, not passed vacuously. The fixture is self-checking.
- *Tautology.* Every expected value is a test-local literal (`Some(vec![repo_a])`, `Some(vec![])`, `StatusCode::FORBIDDEN`, `scoped_scopes()`), never read back from the code under test.
- *Assertion after an early return.* None; each test's `expect(...)` on `authenticate_oci_with_scopes` fails the test if the path did not run.

The token cache is a per-`AuthService` field (`auth_service.rs:980`, `:992`), not a process global, so the test-local `AuthService` cannot pre-poison the one `authenticate_oci_with_scopes` constructs internally — the DB is re-read. The revert-proof's `left: None` for all four is consistent with `generate_tokens` → `generate_tokens_with_family_and_scope(user, .., None, None)` (`auth_service.rs:1232-1234`), so the revert was faithful and they fail for the right reason.

**Not yet independently confirmed:** Check Rust, Backend Unit Tests, Code Coverage and Backend Integration Tests were all still `pending` at review time, so I have not seen the new module compile or the four tests go green in CI. The module uses `use super::*` alongside explicit `use base64::Engine;`, `use std::sync::Arc;` and `use crate::services::auth_service::AuthService;`, all of which are already glob-visible from the file header (`oci_v2.rs:10,24,40`) — explicit-over-glob shadowing, which is legal, but worth watching the Check Rust result rather than assuming.

PR #3321, opened 2026-08-13
fix(oci): filter virtual tags/list members through the caller authz gate

### LINKED ISSUE #3320: Anonymous tags/list on a public Virtual repo leaks private members' tag names (#3261 class, third walker)
### Summary

An **unauthenticated** caller can enumerate the tags of **private** OCI/Docker repositories by listing tags on a **public Virtual** repository that has them as members. This is the tag-name analogue of the manifest/blob laundering that #3261 (#3268 review, F2) closed — the third virtual walker, `tags_list_virtual`, was missed.

This leaks repository/tag **names**, not artifact bytes, so it is lower severity than the manifest/blob laundering. But it is a genuine private-metadata disclosure to unauthenticated callers, and the 1.7.2 `[Unreleased]` CHANGELOG currently tells operators that virtual laundering was closed for OCI reads. Either the code matches that claim or the claim gets qualified.

### Mechanism

Two independent gaps combine.

1. **`handle_tags_list` skips the read gate for anonymous callers.** In `backend/src/api/handlers/oci_v2.rs`, the anonymous check is only `if is_anon && !repo.is_public { return challenge }` — an anonymous caller on a **public** repo passes. The #3261 read gate immediately below is wrapped in `if let Some(claims) = &claims`, and for an anonymous caller `claims` is `None`, so `require_oci_repo_read_access` never runs.

2. **`tags_list_virtual` takes no auth and filters no members.** Its signature was `(state, repo, n_limit, last)` — no claims, no `AuthExtension`. It called `proxy_helpers::fetch_virtual_members(&state.db, repo.id)` **unfiltered** and merged every member's tags into the response.

So `GET /v2/<public-virtual>/<image>/tags/list` with no credential returns tags drawn from private members.

This is exactly the class `proxy_helpers::authorize_virtual_members` documents as the thing it fixed. `resolve_virtual_manifest` and `resolve_virtual_blob` were fixed to narrow the member set through that helper; the tags walker was not.

### Live reproduction

A public Virtual `virt-docker` with two members: a private local repo `team-private-docker` (tag `secret-rel-2026`) and a public local repo `pub-docker` (tag `public-1`). Built from `main` and deployed on an isolated instance.

Unauthenticated flow, exactly as a logged-out OCI client (follow the `/v2/` challenge to the token endpoint with no credential, use the anonymous token it hands out):

```
$ curl .../v2/virt-docker/app/tags/list -H "Authorization: Bearer anonymous"
{"name":"app","tags":["public-1","secret-rel-2026"]}
HTTP 200
```

The private member's tag `secret-rel-2026` is disclosed. A direct anonymous tags/list of the private member is correctly refused:

```
$ curl .../v2/team-private-docker/app/tags/list -H "Authorization: Bearer anonymous"
HTTP 401
```

### Relationship to #3261 / #3268

#3261 authorized the OCI `/v2` read surface; the #3268 review found the same laundering class through a public Virtual parent and filtered `resolve_virtual_manifest` / `resolve_virtual_blob` through `authorize_virtual_members`. `tags_list_virtual` is the third walker of the same shape and was missed. The fix follows the two existing correct call sites verbatim: thread the caller through and narrow members via `authorize_virtual_members` rather than inventing a fourth predicate.

### Sibling walkers checked

Within `oci_v2.rs` the only remaining `fetch_virtual_members` call sites are:

- `authorized_virtual_members` (already narrows through `authorize_virtual_members`; used by the manifest/blob walkers).
- The blob scan-verdict fan-out in `handle_get_blob`, which enumerates members to compute the vulnerable-verdict **blocklist** scope. That one must stay unfiltered by caller: narrowing it would let a caller who cannot see a member escape that member's block gate. It is a security-enforcement fan-out, not a content-serving read, and is correct as-is.

So `tags_list_virtual` was the single missed content-serving walker.

### Scope

- Affects: `GET /v2/<public-virtual>/<image>/tags/list` for any anonymous or under-privileged caller against a public Virtual with private members. Names only; not artifact bytes.
- Not affected: writes; direct private-repo tags/list (already gated by #3261); the manifest/blob walkers (already filtered).

brandonrc:
## Summary

An **unauthenticated** caller could enumerate the tags of **private** OCI/Docker repositories by listing tags on a **public Virtual** repository that has them as members. This is the tag-name analogue of the manifest/blob laundering #3261 (#3268 review, F2) closed — the third virtual walker was missed.

Two gaps combined:

1. `handle_tags_list`'s anonymous check is only `if is_anon && !repo.is_public { return challenge }`, and the #3261 read gate below it is wrapped in `if let Some(claims) = &claims`. An anonymous caller on a *public* parent passes the first check and has `claims == None`, so `require_oci_repo_read_access` never ran.
2. `tags_list_virtual` took no caller and called `proxy_helpers::fetch_virtual_members` **unfiltered**, merging every member's tags — including private members' — into the response.

So `GET /v2/<public-virtual>/<image>/tags/list` with no credential returned tags drawn from private members.

The fix follows the two call sites that already do this correctly — `resolve_virtual_manifest` and `resolve_virtual_blob`. `tags_list_virtual` now threads the caller (`virtual_caller_auth(claims)`, mapping anonymous → `PublicOnly`) and narrows the member set through the same shared visibility filter Maven, npm, PyPI, Debian and goproxy use. A denied member contributes nothing, so neither tag names nor member existence leak. An anonymous caller on a public virtual still sees the tags of its *public* members — legitimate public pulls are unchanged. This is a metadata (names) disclosure, not a bytes leak, so it is lower severity than #3261's manifest/blob class, but it contradicted the CHANGELOG claim that OCI virtual laundering was closed.

**What the filter actually grants (correcting an earlier overstatement).** The shared `authorize_virtual_members` predicate admits a member when it is public **or** the caller holds *any* grant on it — a `role_assignments` row, or any fine-grained `permissions` rule, **regardless of the actions that rule carries** (`permissions_grant_exists_for` gates on `p.actions <> '{}'` with no action term). That is **not** the same as "listable exactly when a direct tags/list would succeed": the direct OCI read gate uses `check_repository_action(…, "read", …)`, where an applicable fine-grained rule is authoritative for the `read` action. So a principal holding only a **write-scoped** rule on a private member is refused a *direct* tags/list of that member but **does** see its tags through the virtual. This predicate gap is deliberately reused, not introduced here — it is shared with the manifest/blob walkers and with repository listing — and narrowing it belongs in that shared code, tracked separately; this PR does not touch it. The PR intentionally uses the shared predicate rather than inventing a fourth one.

**DB-failure posture (regression this PR avoids).** `authorize_virtual_members` returns a `Vec`, flattening a failed `filter_visible_repo_ids` query into the empty set. That is safe for the *walking* callers (manifest/blob resolution), where a missing member reads as "artifact not found here". But this walker **aggregates** the filtered set into the response, and a Virtual repo owns no `oci_tags` rows of its own, so an empty first page routes to `404 NAME_UNKNOWN` — meaning a transient DB blip would tell clients their image no longer exists (before this PR the only DB call on that path, `fetch_virtual_members`, surfaced a retryable **503**). This PR therefore adds and uses `try_authorize_virtual_members`, the fallible sibling: same filter, but a failed visibility query is surfaced as `Err` (the `map_db_err` 5xx shape, matching `fetch_virtual_members`) rather than an empty result. Both forms fail closed; only the definitiveness of the answer differs.

Tag pages are computed per request and never enter the shared negative resolve cache, so the principal-independence concern that governs `authorized_virtual_members` (`VirtualResolveKey` carries no principal term) does not apply here.

**Sibling walkers checked.** Within `oci_v2.rs` the remaining `fetch_virtual_members` call sites are `authorized_virtual_members` (already filters; used by the manifest/blob walkers) and the blob scan-verdict fan-out in `handle_get_blob`. The latter must stay unfiltered by caller — it computes the vulnerable-verdict **blocklist** scope (`gated_ids` → `blob_vulnerable_verdict_status_any`, whose only consumer is the `403 DENIED` deny path; serving happens later through the filtered `resolve_virtual_blob`), and narrowing it by caller would let a caller who cannot see a member escape that member's block gate. It is a security-enforcement deny-set, not a content-serving read. `tags_list_virtual` was the single missed content-serving walker.

**Out of scope, noted (pre-existing, low severity).** The deny-set being unfiltered makes the manifest/blob `403`-vs-`404` distinction an oracle: an anonymous caller can learn that *some* invisible member of a public virtual holds a given blob digest. The fix is to gate the denial *shape*, not to filter the deny-set — that would reintroduce the scan-verdict bypass above — so it is left for a separate change.

Closes #3320. Fixes #3320

Two tests, both in `#[cfg(test)]` inside `src/` so `cargo nextest run --workspace --lib` scores them:

- `oci_read_authz_tests::public_virtual_does_not_leak_a_private_members_tags`: a public Virtual over a private member (tag `v1`) and a public member (tag `zpublic`). Asserts the anonymous caller's list omits `v1` and still contains `zpublic` (public pulls not broken), an ungranted authenticated caller likewise sees only the public tag, and the private member's granted principal sees **both** tags (the filter narrows to the caller, it does not deny everyone). Reverting the tags filter makes it fail with `got tags ["v1", "zpublic"]` (~0.4s against a real DB).
- `proxy_helpers::tests::try_authorize_virtual_members_surfaces_query_failure_as_err`: with the pool severed, the fallible form returns `Err` carrying a 5xx while the flattening form still returns the empty set — pinning the 503-not-404 posture above. Reverting the helper's error arm to `Ok(empty)` makes it fail.

Revert-proof was re-run after this round: each production hunk reverted against merge-base fails the matching test for the right reason; the additions are insertions inside test modules plus the new fallible helper.

## Test Checklist
- [x] Manually tested locally

## Coordination with #3317

#3317 (scoped-token repository ceiling) is still open at the time of writing and threads `allowed_repo_ids` into OCI claims, which flows into `virtual_caller_auth` → `member_passes_token_scope` — the token-scope arm of this same filter. Neither PR tests the combination. I will rebase on `main` once #3317 merges; if adding a combined test (a direct scoped token whose ceiling excludes a member the visibility grant would otherwise admit) is cheap on the rebased tree I will add it here, otherwise I will flag it for a separate follow-up.

## API Changes
- [x] N/A - no API changes


brandonrc:
## Adversarial review — pinned to `7b890428`

Read-only review of the head commit (every line below verified against `7b890428`, not a working tree). The core fix is right: `virtual_caller_auth(None)` → `member_grant_visibility(None) = RepoVisibility::PublicOnly` (`repositories.rs:330-336`) conjoined with `member_passes_token_scope(None, …) = member_is_public` (`repositories.rs:356-367`), so an anonymous caller reaches public members only. That closes the reported leak. Five findings below, `confirmed` (read off the source at this SHA) kept separate from `plausible`.

---

### F1 — MEDIUM (confirmed, pre-existing, but this PR re-asserts the invariant it breaks)

**The member filter and the direct read gate are not the same predicate, and the filter is the looser one.**

- `proxy_helpers::authorize_virtual_members` (`proxy_helpers.rs:3119-3154`) resolves visibility through `build_grant_predicate` (`repository_service.rs:403-420`), which is *existence of any grant*: a `role_assignments` row, **or** a `permissions` row with `p.actions <> '{}'` (`repository_service.rs:372-391`). No action term.
- A direct `GET /v2/<member>/tags/list` goes through `require_oci_repo_read_access` → `PermissionService::check_repository_action(sub, repo, "read", …)` (`permission_service.rs:162-237`), where `WHEN EXISTS (SELECT 1 FROM applicable_rules) THEN …` makes an applicable fine-grained rule **authoritative** — the role fallback is skipped entirely.

Concrete failure: private OCI member `M` inside public virtual `V`. A deploy account holds one fine-grained rule on `M` with `actions = {'write'}`.

| request | predicate | result |
|---|---|---|
| `GET /v2/M/app/tags/list` | `check_repository_action(read)` → rule applicable, `'read' ∉ {'write'}` | denied |
| `GET /v2/V/app/tags/list` | `permissions_grant_exists` → `actions <> '{}'` → true | **`M`'s private tags returned** |

So the PR body's and CHANGELOG's claim — *"a member's tags are listable through the virtual exactly when a direct tags/list of that member would succeed"* — is false for any principal governed by a fine-grained rule that withholds `read`. This is not introduced here: `resolve_virtual_manifest` / `resolve_virtual_blob` already use the same helper, so for that principal the virtual also serves `M`'s **manifest and layer bytes**, which is materially worse than tag names. Worth either narrowing the CHANGELOG claim or filing the predicate-divergence separately; shipping the sentence as written tells operators a boundary holds that does not.

### F2 — LOW/MEDIUM (confirmed, **newly introduced by this PR**)

**A transient DB error in the new authorization query renders as `404 NAME_UNKNOWN` on a healthy repository.**

`authorize_virtual_members` returns `Vec`, not `Result`; a `filter_visible_repo_ids` failure logs a warn and returns the empty set (`proxy_helpers.rs:3137-3146`). At `oci_v2.rs:9770-9771` that empty set flows into an empty merged page, and `handle_tags_list` then hits:

```rust
// oci_v2.rs:9246-9259
if page.is_empty() && !has_more && last.is_none()
    && repo.repo_type != RepositoryType::Remote
    && !local_tags_exist(&state.db, repo.id, &repo.image).await.unwrap_or(false)
{ return oci_error(StatusCode::NOT_FOUND, "NAME_UNKNOWN", …) }
```

A Virtual repo owns no `oci_tags` rows, so `local_tags_exist` is always false for it — the handler answers `404 repository name not known to registry` for a virtual that exists and has tags. Before this PR `tags_list_virtual` had no swallowing call: `fetch_virtual_members` routes through `map_db_err`, so pool saturation surfaced as the retryable `503` that helper exists to produce (`proxy_helpers.rs:3053-3055`). This PR adds a second DB call on the same path whose failure mode is a semantically wrong, non-retryable `404`. A client or CI job reads `NAME_UNKNOWN` as "the image is gone", not "retry".

Cheapest fix that keeps fail-closed: at this call site, compare against the pre-filter member count and distinguish "authorized to nothing" from "could not evaluate" — return the `503` shape for the latter, or suppress the `NAME_UNKNOWN` synthesis for a Virtual parent that has ≥1 member and answer an empty `200`.

### F3 — LOW (confirmed, pre-existing) — direct answer on the scan-verdict fan-out

**Leaving `handle_get_blob`'s fan-out unfiltered is correct. It is also an oracle.**

Traced it: `oci_v2.rs:4941-4980` enumerates all members into `gated_ids`/`ref_owners` and feeds `blob_vulnerable_verdict_status_any` (`oci_v2.rs:7880-7912`), whose only consumers are the `403 DENIED` at `4989-5040`. Nothing in it selects *what gets served* — serving happens later at `5092-5098` through `resolve_virtual_blob`, which **is** filtered by `virtual_caller_auth`. So the fan-out is purely a deny-set. Narrowing it by caller visibility would let a caller who cannot see member `B` escape `B`'s block gate, and the author's reasoning holds. **Do not filter it.**

But the two outcomes are distinguishable, and the blocklist spans members the caller cannot see:

- anonymous `GET /v2/<public-virtual>/<image>/blobs/sha256:D`, where invisible private member `P` holds a `manifest_blob_refs` row for `D` under a manifest with a `vulnerable` verdict → `Blocking` → `403 "blob belongs to an image blocked by scan policy: vulnerabilities found"` (`5019-5023`);
- same request where `D` is simply absent → filtered `resolve_virtual_blob` finds nothing → `404 BLOB_UNKNOWN`.

`403` vs `404` therefore confirms "an invisible member of this virtual holds blob `D`". Digests are content-addressable and often publicly known, so this is a probeable existence oracle over private members' layer inventory. Low severity — no names, no bytes, and it needs a vulnerable verdict on a referencing manifest — but it is exactly the inference channel the question anticipated, and it is not closed by "don't filter".

Direction that preserves both properties: keep the blocklist computed over **all** members (unchanged), and gate only the *shape* of the denial on whether the caller could have resolved the blob at all — if no authorized member holds `D`, answer the `BLOB_UNKNOWN` such a caller would otherwise have received. A caller who can see member `A` still gets `403` when invisible member `B` supplied the verdict, so the escape-the-gate property is intact.

Related, out of scope: `handle_head_blob` (`oci_v2.rs:4712-4770`) has no scan-verdict gate at all — `HEAD` on a blocked blob returns `200` with `Content-Length`. No oracle there, but the block is `GET`-only.

### F4 — LOW (confirmed, pre-existing) — the invariant fails in the permissive direction too

`require_oci_repo_read_access` runs the token-scope ceiling **first**, ahead of the public-repository shortcut (`oci_v2.rs:702-715`): a repo-scoped token is denied even a *public* repository outside its allow-list. `member_passes_token_scope` short-circuits the opposite way — `member_is_public || a.can_access_repo(member_id)` (`repositories.rs:363-366`).

So a token scoped to `[V]` is refused `GET /v2/P/app/tags/list` for public member `P` (403 `DENIED`), yet lists `P`'s tags through `V`. Public content only, so the impact is containment rather than confidentiality — but it is a second counterexample to "listable exactly when a direct tags/list would succeed."

### F5 — informational: the "single missed walker" claim is scoped narrower than it reads

Within `oci_v2.rs` the audit is right — `fetch_virtual_members` appears at `3157` (filtered), `4942` (enforcement, F3) and `9770` (fixed here), and all five OCI read verbs (`HEAD`/`GET` blobs, `HEAD`/`GET` manifests, `GET` tags — the complete set in `catch_all`, `oci_v2.rs:10305-10381`) now thread `virtual_caller_auth`.

Repo-wide it does not hold. Content-serving virtual walkers still calling `fetch_virtual_members` with no caller, verified by reading each: `helm.rs:323` (`index.yaml` merge) and `helm.rs:530` (chart download member walk), `rubygems.rs:595` (`specs_index`, `query_local_member_specs` over every member), `nuget.rs:108` (`effective_local_repo_ids` — federates local listing/search across all members), plus the shared `proxy_helpers::resolve_virtual_metadata` (`2722`) and `collect_virtual_metadata` (`2869`), neither of which takes an `auth` parameter. Those handlers sit behind `repo_visibility_middleware`, which authorizes the URL repo only — the same confused-deputy shape `authorize_virtual_members` documents. Not a blocker for this PR, but the sentence "`tags_list_virtual` was the single missed content-serving walker" should say "within `oci_v2.rs`".

(The `virtual_non_remote_owns_*` / `pypi_virtual_isolates_name` helpers at `proxy_helpers.rs:3951/4077/4181/4316` are correctly classified the other way — dependency-confusion enforcement, blocklist-shaped, must stay unfiltered.)

### F6 — plausible, low

`authorize_virtual_members` has no scanner-token exemption, while `require_oci_repo_read_access` does (`oci_v2.rs:707`). A `scan_pull_repo` token pinned to a Virtual repo would therefore see only its public members. I could not find a production path that scans through a virtual parent (scan tokens are minted for the repo owning the artifact, and artifacts live in members), so I am filing this as plausible rather than confirmed.

---

### The test

`oci_read_authz_tests::public_virtual_does_not_leak_a_private_members_tags` (`oci_v2.rs:30506`) is a real test, not a tautology. Checked each of the recurring failure classes:

- **Precondition present.** `ReadFixture::setup()` (`29859`) creates the fixture repo with `is_public` at its column default (false) and the new test never publishes it — only `public_id` and `virt_id` are flipped (`UPDATE … WHERE id = ANY($1)`). The private member really is private.
- **Not asserting the bug.** Without the production hunk, `tags_list_virtual` merges both members unfiltered and the first assertion fails on `["v1","zpublic"]`. The failure mode matches what the PR body reports.
- **No early-return hole.** `tag_names` on a non-200 body yields an empty vec, which would satisfy the first `!contains(TAG)` assertion — but `assert_eq!(anon_status, StatusCode::OK)` immediately after closes it, and `anon_tags.contains(PUBLIC_TAG)` forces a real populated page.
- **Positive control is real.** `tdh::grant_repo_access` writes a `role_assignments` row for role `developer` (`test_db_helpers.rs:825-836`), which `build_grant_predicate`'s first `EXISTS` arm honours — so "granted principal sees both tags" genuinely exercises the allow path rather than a filter that denies everyone.
- Assertions run after teardown, so a failure does not leak fixture rows. Good.

Gaps worth closing while the fixture is in hand:

1. **No pagination coverage** — `tags_list_virtual` takes `n_limit`/`last` and the test never passes either. I convinced myself by construction that filtering cannot corrupt the cursor (each authorized member is queried with `member_limit = n+1` and the merged cursor; `apply_cursor_pagination` at `oci_v2.rs:2527-2543` sets `has_more = end < tags.len()`, and `has_more` can only be false when every authorized member returned fewer than `n+1` remaining, i.e. the page really is the tail) — but that is my reasoning, not a test. One `?n=1` call asserting the `Link` header and the second page would pin it.
2. **No Remote member** — only local members are exercised, so the `fetch_tags_from_remote_member` arm of the filtered walk is uncovered.
3. **No direct-vs-virtual contrast** — the issue's whole framing is "the direct read is refused, the virtual is not". Asserting the anonymous direct `tags/list` of the private member 401s in the same fixture would make the invariant explicit (and would have surfaced F1/F4 as soon as anyone tried to generalise it).
4. The `let Some(mut f) = ReadFixture::setup().await else { return; }` shape means the test **passes silently with no database**. Any revert-proof run on this test needs `AK_TESTS_REQUIRE_DB=1` (or an asserted `DATABASE_URL`) to mean anything — a sub-100ms "pass" is the tell that it never ran.

### Other checks that came back clean

- Negative cache: the tags path never reads or writes `virtual_negative_cache` (it is confined to `authorized_virtual_members`, `oci_v2.rs:3151-3167`), so the principal-independence constraint that governs the manifest/blob walkers genuinely does not apply. The PR's reasoning here is correct.
- Admin: `member_grant_visibility(Some(a)) if a.is_admin => RepoVisibility::All` renders a literal `true`, so an admin's tag list is byte-identical to before. `AuthExtension::from(Claims)` (`auth.rs:269-293`) carries `is_admin`, `scopes` and `allowed_repo_ids` and applies `with_scope_gated_admin()`, so the GHSA-vvc3 demotion is not lost.
- The `if let Some(claims)` shape at `9212-9216` is shared by all five read verbs (`4707`, `4898`, `7476`, `8546`, `9213`). It is benign on every one of them: for an anonymous caller the repo has already been proven public by `is_anon && !repo.is_public`, and `require_oci_repo_read_access` returns `Ok` for a public repo via `public_read_satisfies_acl` (`oci_v2.rs:713`, `auth.rs:1703-1705`) with no scope ceilings to apply. The defect was never the `if let` — it was the unfiltered walk behind it, and all three walkers now filter. Nothing further to fix in that shape.
- The `404 NAME_UNKNOWN` that a fully-filtered-out member set now produces is not an oracle: a virtual whose members are all invisible and a virtual with no such image answer identically.
- Anonymous callers can no longer trigger upstream fan-out to private Remote members via `fetch_tags_from_remote_member` — an unadvertised side benefit.

### Coordination with #3317

#3317 (`d0f94dc2`) touches `authenticate_oci_with_scopes` (~`373-480`) and appends a test module after `30778`; #3321's hunks are at `9225-9771` and inside `oci_read_authz_tests` at `30489`. No textual overlap — expect a clean merge apart from both inserting a bullet into the same `### Security` block in `CHANGELOG.md`, which is a trivial resolve.

The semantic interaction is worth a note: #3317 makes a directly-presented scoped API token carry its `allowed_repo_ids` into the OCI claims, and those flow straight into `virtual_caller_auth` → `member_passes_token_scope`. That is the desirable direction, but neither PR's tests cover the combination, and F4's public-member scope asymmetry gets a wider blast radius once #3317 lands.

### CI

At review time the Rust jobs (`Check Rust`, `Backend Unit Tests`, `Backend Integration Tests`, `Code Coverage`) were still `pending` on `7b890428`; only the CodeQL/shell/gate jobs had reported. Nothing here is blocked on that, but the PR is not green yet.

---

**Verdict on the fan-out question:** the reasoning is sound and the code should stay unfiltered — it is a deny-set with no serving role, and filtering it would be a privilege escalation. The residual is F3's `403`-vs-`404` oracle, which is a separate defect with a separate fix, not an argument for filtering.

brandonrc:
Heads-up on a merge interaction with #3326 (issue #3325), which changes the same function. Both are green independently; the hazard is in the reconciliation, and it is not just textual.

**What each PR does to `authorize_virtual_members`:**

- This PR splits it: the body moves into a new `try_authorize_virtual_members`, and `authorize_virtual_members` becomes a thin `.unwrap_or_default()` wrapper.
- #3326 appends a second gate to the body — a `check_repository_action(.., "read", ..)` pass over the members that clear the existing grant filter — because the grant filter admits any grantee, so a `{write}`-only grant on a private member currently yields that member's manifest and blob bytes through a virtual.

**The trap.** Whoever rebases second will get a conflict in that function. If #3326's action gate is resolved into the **wrapper** rather than into `try_authorize_virtual_members`, the code compiles, both PRs' test suites still pass, and `tags_list_virtual` — the aggregating caller this PR is adding, which calls the fallible form — silently skips the action gate. That reintroduces the #3325 leak on exactly the path this PR is hardening, with no failing test to catch it, because #3326's regressions exercise `caller_can_read_member` / `authorize_virtual_members` and would still be green.

**Resolution rule:** the action gate belongs at the end of `try_authorize_virtual_members`, so both forms inherit it. The wrapper stays a pure `.unwrap_or_default()`.

Worth noting the two changes are complementary rather than overlapping in effect. Measured live on a stack built from `main`, a **public** virtual with a **private** member:

```
no-grant authenticated user, VIRTUAL tags/list      200  <- this PR
no-grant authenticated user, VIRTUAL manifest/blob  404  (already filtered)
{write}-only user,           VIRTUAL manifest/blob  200  <- #3326
{write}-only user,           DIRECT  tags/list      403
```

So this PR closes the unfiltered aggregation path and #3326 closes the action-blindness in the filter itself; neither subsumes the other. After both, a member is readable through a virtual exactly when a direct read of it would succeed, on every verb.

No objection to either merge order — flagging only so the conflict is resolved deliberately rather than by taking one side of the hunk.

PR #3326, opened 2026-08-13
fix(authz): require the read action for virtual-repo member resolution

### LINKED ISSUE #3325: Virtual-repo member reads admit any grant, not the read action: a write-only grantee gets a private member's bytes
## Summary

Virtual-repository member resolution admits a member on the **existence of any grant**, while a direct read of that same member requires the **`read` action**. A principal whose only entitlement on a private member is `{write}` is refused a direct read and is then handed that member's manifest and blob bytes through a virtual repository that contains it.

The member filter is doing the job of a *tenant* gate but is being used as the *action* decision. This repository already recognises that distinction and has a standard remedy for it — see "Why this is the right seam" below.

Live-confirmed on a stack built from `main` at 8796568. Both halves ship in 1.7.2: the member-visibility predicate (#3081, #3163, #3178) and the OCI virtual read walkers (#3261).

## Reproduction

Private `docker` repo `M` holding `payments-api:secret-tag`; **public** virtual `V` with member `M`; user `U` holding exactly one fine-grained rule on `M`: `actions = {write}`.

```
[U] DIRECT  GET /v2/M/payments-api/tags/list                403 DENIED
[U] VIRTUAL GET /v2/V/payments-api/manifests/secret-tag     200  <full manifest>
[U] VIRTUAL GET /v2/V/payments-api/blobs/sha256:c5b1d6…     200  <config bytes>
```

Controls, same run — the gap is specific, not a blanket authorization failure:

```
[read-granted user] VIRTUAL manifest / blob                 200 / 200   (legitimate, must not regress)
[no-grant user]     VIRTUAL manifest / blob                 404 / 404   (filter works — it is only action-blind)
```

So the filter is not absent. It resolves the right principal against the right member and correctly refuses a stranger. It just never asks *which* action the grant carries.

## Root cause

`proxy_helpers::authorize_virtual_members` (`backend/src/api/handlers/proxy_helpers.rs:3119`) composes its entitlement half from `RepositoryService::filter_visible_repo_ids` → `build_visibility_clause_for` → `build_grant_predicate` (`backend/src/services/repository_service.rs:403`). Neither arm of that predicate carries an action term:

* the `permissions` arm (`permissions_grant_exists_for`, `repository_service.rs:372`) tests `AND p.actions <> '{}'` — any non-empty action array qualifies, including `{write}`, `{delete}`, or an unrecognised string, since `permissions.actions` is `TEXT[]` and not the `permission_type` enum;
* the `role_assignments` arm does not join `roles` at all, so an assignment to a role carrying no `read` also qualifies.

The direct read gates ask a different and stricter question — `PermissionService::check_repository_action(user, repo, "read", is_admin)` (`permission_service.rs:162`), reached from `oci_v2::require_oci_repo_read_access` (`oci_v2.rs:725`) and from `repo_visibility_middleware` (`auth.rs:1959`) for the native formats.

`build_grant_predicate` ⊋ `check_repository_action(read)` strictly: every input the action gate admits also satisfies the grant predicate, so this is purely a widening, and closing it can only ever remove access that the direct path already refuses.

### Blast radius

`authorize_virtual_members` is the single choke point for virtual byte resolution across **every** format, so the gap is not OCI-specific — it is reachable through 11 direct call sites and ~30 more transitively:

| Surface | Call site |
|---|---|
| OCI manifest + blob (GET/HEAD ×2 each) | `oci_v2.rs:3161` via `resolve_virtual_blob` :3182, `resolve_virtual_manifest` :3283 |
| Maven download / metadata merge / serve | `maven.rs:1185`, `:1348`, `:1867` |
| npm tarball | `npm.rs:3229` |
| PyPI simple index + file download | `pypi.rs:1889`, `:2643` |
| Debian pool download | `debian.rs:2821` |
| Go module zip | `goproxy.rs:1080` |
| Generic virtual download (+ streaming) | `proxy_helpers.rs:2353`, `:2568` — reached from ~28 further format handlers (cargo, conan, nuget, rpm, rubygems, terraform, alpine, hex, composer, …) and from REST `GET /repositories/{key}/artifacts/*path` |

## Caller inventory of the shared predicate

The predicate is deliberately shared, and the sharing is correct — this is why the fix is **not** a narrowing of the fragment. Every production consumer, classified:

### Enumeration — "may you see that this repository exists". Any grant is the intended contract; **unchanged by this fix**.

| Call site | Surface |
|---|---|
| `repository_service.rs:1173` (`list`) → `repositories.rs:2427` | `GET /api/v1/repositories` |
| `packages.rs:180`, `:181` | `GET /api/v1/packages` (page + count) |
| `repositories.rs:8794` (`list_virtual_members`) | `GET /api/v1/repositories/{key}/members` |
| `admin_security.rs:524`, `:717` | `GET /api/v1/admin/security/{cve,artifact}/…/accessible-users` — an admin blast-radius report that must mirror the read predicate exactly |

### Content resolution — "may you receive these bytes / this metadata". Needs the action.

| Call site | Surface | Status |
|---|---|---|
| `proxy_helpers.rs:3133` (`authorize_virtual_members`) | all virtual byte resolution, every format | **fixed here** |
| `repositories.rs:4933`, `:5887` | `GET /repositories/{key}/artifacts` on a virtual (incl. `group_by=maven_component`) | residual, see below |
| `packages.rs:282`, `:377` | `GET /api/v1/packages/{id}`, `/versions` — returns `metadata` JSON and `checksum_sha256` | residual |
| `artifacts.rs:62` (`check_artifact_visibility`) | `GET /api/v1/artifacts/{id}` and the label / security / SBOM / quality-gate siblings | residual |
| `repositories.rs:400` (`require_visible`) | the general REST read gate, 33 call sites | residual |
| `tree.rs:91` | `GET /api/v1/tree/content` — serves artifact bytes | residual |
| `sbom.rs:1695`, `quarantine.rs:156` | SBOM component lists; quarantine reason | residual |

### Other

| Call site | Surface |
|---|---|
| `repository_service.rs:1846`, `:1937` (`build_member_visibility_clause`) | virtual `storage_used_bytes` aggregate — a size oracle, summed inside the database, so it cannot be filtered row-wise |
| `repositories.rs:120` (`require_repo_write_access`) | tenant gate for ~16 REST write/delete handlers — **already** paired with an action gate at the two handlers that need one (`repositories.rs:7314`, `:8527`) |
| `repo_tokens.rs:237`, `webhooks.rs:111`, `curation.rs:757`/`:889`, `promotion.rs:495` | token-mint / ownership / tenant gates |

## Why this is the right seam

Three options were considered.

**Rejected — add an action term to the shared fragment.** It would silently narrow repository *listing*, which shares the fragment and for which "any grant means you may see this repository exists" is the intended contract; that is a fail-closed policy change shipping unremarked in a patch. Worse, it would not even be correct: a bare `'read' = ANY(p.actions)` cannot express `check_repository_action`'s actual semantics — *an applicable rule is authoritative for the principals it names, every other principal keeps its role-based capabilities, and a role carrying `admin` always wins* — and it would leave the action-blind `role_assignments` arm untouched. The result would be a fourth predicate that drifts from the three that already exist. The fragment's own comments warn against exactly this.

**Rejected — narrow only the claim.** The gap hands out private bytes on a path whose direct equivalent refuses them, it is reachable from every package format, and it is closable without touching shared SQL. Documenting it is not proportionate.

**Chosen — layer the canonical action choke-point on top of the tenant gate, at the content-resolution call site.** This is not a new pattern; it is *this repository's established idiom for this exact situation*, introduced by #2603 G1 and currently applied at two REST handlers whose comments describe the identical problem verbatim:

> `repositories.rs:8527` — *"Action gate (#2603 G1): the tenant gate above admits any grantee (incl. a write-only or read-only one), collapsing write/delete. Route the `delete` decision through the canonical `check_repository_action` choke-point (deny-by-default)…"*

The virtual read path is the same shape — `require_repo_write_access` and `authorize_virtual_members` both use the action-blind grant predicate as a tenant gate — and it is the one content path that never received the second gate. Applying it here means member reads and direct reads are answered by *the same function*, so they cannot drift; the shared fragment is untouched, so listing behaviour is byte-identical; and the subtle role-fallback semantics come along for free instead of being re-implemented.

Public members short-circuit ahead of the action check, preserving the anonymous read baseline (#2329) exactly as `public_read_satisfies_acl` does in the direct `/v2` gate. Global admins short-circuit too. Failure is closed.

## Fail-closed impact

A principal whose only entitlement on a **private** virtual member is a grant without `read`/`admin` — a `{write}`-only or `{delete}`-only fine-grained rule, or a role assignment to a role carrying no `read` — loses read of that member through a virtual. Those principals are **already refused the same bytes on a direct read**, so this removes an inconsistency rather than a capability. Publish-only CI identities are the population to check; grant them `read` on the member if they are also expected to pull through a virtual.

Unaffected: read-granted users, admins, public members, anonymous callers, token scoping, and every listing/enumeration surface.

## Known residuals (not fixed here)

Filed separately rather than folded in, because they change enumeration semantics or span many surfaces and do not belong in a patch:

1. **`user_can_access_repo` is action-blind on REST content paths.** `GET /api/v1/artifacts/{id}`, `GET /api/v1/tree/content` and `require_visible`'s 33 call sites gate bytes on "any grant". After this fix the virtual path is *stricter* than the REST direct read of the same member — the CHANGELOG's "readable through a virtual exactly when its own `GET` would succeed" is inexact in both directions and is being corrected.
2. **Virtual `storage_used_bytes`** still aggregates over any-grant members (`build_member_visibility_clause`), disclosing private members' exact byte totals.
3. **`GET /v2/{virtual}/tags/list` applies no member filter at all** (`oci_v2.rs:9748`, `tags_list_virtual` fans out over the raw `fetch_virtual_members` set) — any authenticated caller enumerates a private member's tags. Confirmed live; already being fixed in #3321, referenced here only so the boundary between the two is on the record.
4. `resolve_virtual_metadata` (`proxy_helpers.rs:2722`) takes the unfiltered member set and relies on callers to pre-filter; `hex.rs:417` and `goproxy.rs:404` reach it with no visible pre-filter.

## Verification

* Live red→green on an isolated stack built from source: write-only principal loses virtual manifest/blob (200 → 404) while the read-granted principal keeps them (200 → 200), the write-only principal keeps its direct push to the member, and the no-grant principal stays denied.
* Two DB-backed regression tests in `#[cfg(test)]` inside `src/`, covering both authz stores and pairing every deny assertion with its allow control.
* Revert-proof against merge-base with `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1`: production hunks reverted, tests fail on the write-only assertion; the revert diff is 223 insertions / 0 deletions, entirely inside `mod tests`.

brandonrc:
Fixes #3325

Virtual-repository member resolution admits a member on the **existence of any grant**; a direct read of that same member requires the **`read` action**. A principal holding exactly one rule on a private member — `actions = {write}` — is refused a direct read and then handed that member's manifest and blob bytes through a virtual repository containing it.

## The two predicates

`proxy_helpers::authorize_virtual_members` builds its entitlement half from `build_grant_predicate`, and neither arm carries an action term:

* the `permissions` arm (`permissions_grant_exists_for`) tests `AND p.actions <> '{}'` — any non-empty array qualifies, including `{write}`, `{delete}` or an unrecognised string, since `permissions.actions` is `TEXT[]` rather than the `permission_type` enum;
* the `role_assignments` arm never joins `roles`, so an assignment to a role carrying no `read` also qualifies.

The direct read gates ask `PermissionService::check_repository_action(user, repo, "read", is_admin)` — `oci_v2::require_oci_repo_read_access` for `/v2`, `repo_visibility_middleware` for the native formats. `build_grant_predicate` is a strict superset of that, so this is purely a widening and closing it can only remove access the direct path already refuses.

## Why the shared fragment is not touched

`build_grant_predicate` is shared with repository **listing**, and that sharing is deliberate and correct. For a listing, "any grant means you may see this repository exists" is the intended contract — someone with write-only access should still see the repository in their list. Narrowing the fragment would have changed enumeration behaviour, fail-closed and unremarked, in a patch.

It would also not have been correct. A bare `'read' = ANY(p.actions)` cannot express what the choke-point actually decides — *an applicable rule is authoritative for the principals it names, every other principal keeps its role-based capabilities, and a role carrying `admin` always wins* — and it would leave the action-blind `role_assignments` arm untouched. The result would be a fourth predicate drifting from the three that exist, which is the outcome the fragment's own comments warn about.

## The seam

This is not a new pattern. It is the repository's established idiom for exactly this situation, introduced by #2603 G1 and currently applied at two REST handlers whose comments describe the identical problem:

> `repositories.rs:8527` — *"Action gate (#2603 G1): the tenant gate above admits any grantee (incl. a write-only or read-only one), collapsing write/delete. Route the `delete` decision through the canonical `check_repository_action` choke-point (deny-by-default)…"*

`require_repo_write_access` and `authorize_virtual_members` have the same shape — both use the action-blind grant predicate as a **tenant** gate. The virtual read path is the one content path that never received the second gate. It gets it here: the same function, with the same arguments, that the direct read gates call. Member reads and direct reads are now answered by one function and cannot drift.

Public members short-circuit ahead of the action check, preserving the anonymous read baseline (#2329) exactly as `public_read_satisfies_acl` does in the direct `/v2` gate. Global admins short-circuit. A failed per-member lookup drops that member rather than widening the set.

Because the helper is the single choke point for virtual byte resolution, one change covers OCI manifest/blob (GET+HEAD), Maven, npm, PyPI, Debian, Go, and the ~28 formats routing through `resolve_virtual_download`.

## Placement: the gate is in the shared body, and that is asserted mechanically

Rebased onto `main` after #3321, which split the helper: `authorize_virtual_members` is now a thin `.unwrap_or_default()` wrapper over `try_authorize_virtual_members`, and `tags_list_virtual` calls the **fallible form directly**.

The gate is therefore in `try_authorize_virtual_members`. Putting it in the wrapper instead would compile, keep every test written against the walking (manifest/blob) callers green, and silently leave tags/list ungated — a defect that survives compilation, review and CI.

That is not left to inspection. Two tests pin the placement, and both were verified against a deliberately misplaced variant with the gate moved into the wrapper:

```
gate in the WRAPPER (deliberately wrong)  -> cargo check: clean, compiles fine
  FAIL  proxy_helpers::…::try_authorize_virtual_members_applies_the_read_action_gate
  FAIL  oci_v2::oci_read_authz_tests::write_only_grant_does_not_enumerate_a_private_members_tags
        panicked: … so this failing means the read-action gate is in the
        authorize_virtual_members WRAPPER rather than in the shared body;
        got ["v1", "zpublic"]
  PASS  proxy_helpers::…::test_virtual_member_read_requires_read_action
  PASS  proxy_helpers::…::test_virtual_member_read_requires_read_action_via_role_assignment

gate in the shared BODY (as shipped)      -> all four PASS
```

The two behaviour-only tests passing against the misplaced variant is the point: the pre-existing suite could not have caught this, which is why the two placement guards exist.

## Live red → green

Isolated stack built from source, private member `M` holding `payments-api:secret-tag`, **public** virtual `V` containing `M`. Identical script both sides.

| Principal / request | before | after |
|---|---|---|
| `{write}`-only — DIRECT `/v2/M/…/tags/list` | 403 | 403 |
| **`{write}`-only — VIRTUAL manifest** | **200 + manifest** | **404** |
| **`{write}`-only — VIRTUAL blob** | **200 + bytes** | **404** |
| `{read}` — VIRTUAL tags / manifest / blob | 200 / 200 / 200 | 200 / 200 / 200 |
| `{read}` — DIRECT tags/list | 200 | 200 |
| no grant — VIRTUAL manifest / blob | 404 / 404 | 404 / 404 |
| `{write}`-only — direct blob + manifest PUSH to `M` | 201 / 201 | 201 / 201 |

The write-only principal keeps everything it legitimately had and loses only the read it was already refused directly; the read-granted principal is untouched.

## Tests

Four DB-backed regressions in `#[cfg(test)]` inside `src/` (`backend/tests/` scores zero on the `--lib` coverage gate), each pairing every deny assertion with its allow control so a fix that simply denied more would fail rather than pass:

* `test_virtual_member_read_requires_read_action` — `{write}` denied; `{read}`, `{write,read}`, `{admin}`, and a public member with a write-only rule all allowed; the set form agrees with the single form; the global-admin bypass is preserved.
* `test_virtual_member_read_requires_read_action_via_role_assignment` — a custom role carrying `write` but not `read` is denied; the stock `developer` role still works.
* `try_authorize_virtual_members_applies_the_read_action_gate` — placement guard at the unit boundary: the fallible form must apply the gate.
* `oci_v2::oci_read_authz_tests::write_only_grant_does_not_enumerate_a_private_members_tags` — placement guard end-to-end through the real `tags/list` route, in #3321's own fixture: a `{write}`-only principal must not enumerate the private member's tag, while the `{read}` principal still sees it and both still see the public member's tag.

**Revert-proof**, with `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1` both exported:

```
production hunks reverted to merge-base, tests kept:
  git diff --numstat  ->  223  0  backend/src/api/handlers/proxy_helpers.rs
  single hunk @@ -13095,0 +13096,223 @@ mod tests   (module opens at 6287, runs to EOF)

FAIL test_virtual_member_read_requires_read_action                    [0.231s]
  panicked: a {write}-only grant must NOT confer read of a private member through a virtual
FAIL test_virtual_member_read_requires_read_action_via_role_assignment [0.240s]
  panicked: a role assignment whose role carries no `read` must NOT confer read …
```

223 insertions / 0 deletions, the single hunk entirely inside `mod tests` — the revert touched no test code. Timings of 0.23s (not ~0.04s) confirm the tests actually reached the database.

With the fix restored: `252/252` in `proxy_helpers::tests` and `183/183` across `repository_service::` + `permission_service::`. `cargo fmt --all` and `cargo clippy --workspace --all-targets -D warnings` clean. No migration, no `sqlx::query!` macros added.

## Operator impact — fail-closed

A principal whose only entitlement on a **private** virtual member is a grant without `read`/`admin` loses read of that member through a virtual. Publish-only CI identities are the population to check. They are already refused the same bytes on a direct read, so this removes an inconsistency rather than a capability; grant them `read` on the member if they are also expected to pull through the virtual. CHANGELOG entry carries the "am I affected" framing.

## Coordination

Rebased onto `main` at 0ef9789 (post-#3321, post-#3317).

* **#3321 (merged).** It split the helper; this PR's gate lands in the body it created, so #3321's new `tags_list_virtual` path inherits the action gate. The placement guards above exist specifically because of this interaction.
* **#3328 (open).** It dropped its duplicate helper in favour of the merged one and now calls `try_authorize_virtual_members` from `helm.rs`, `nuget.rs` and `swift.rs`. There is exactly one definition and no ambiguity about which is authoritative. Because the gate is in that shared body, those three new byte-download call sites inherit the read-action check for free — worth knowing when reviewing #3328, since its scope is now slightly wider than its description. Its `test_db_helpers.rs` changes are pure additions (`link_virtual_member`, `cleanup_member_repo`) and do not touch the helpers these tests use; this PR does not modify that file.

The CHANGELOG entry states plainly that this **supersedes** the #3163 claim that "a member is readable through a virtual exactly when its own `GET` would succeed": that was true of the REST `GET /api/v1/repositories/{key}`, which is also action-blind, but not of the `/v2` read gate, which is not.

## Residuals, deliberately not in this PR

Recorded in #3325 rather than folded in, because they change enumeration semantics or span many surfaces:

1. **`user_can_access_repo` is action-blind on the REST content paths — filed as #3331 (1.8.0), and it is the larger half of this defect class.** `GET /api/v1/artifacts/{id}`, `GET /api/v1/tree/content`, `GET /api/v1/sbom`, `GET /api/v1/quarantine/{id}` and `require_visible`'s 33 call sites all gate bytes and private contents on "any grant". After this PR the virtual path is deliberately *stricter* than the REST direct read of the same member: the virtual was aligned with the stricter of the two existing gates rather than the looser one. Closing #3331 is a per-call-site classification across ~35 surfaces and a real capability removal, which is why it is not in a patch.
2. Virtual `storage_used_bytes` still aggregates over any-grant members — it sums inside the database, so it cannot be filtered row-wise the way this path can.
3. `GET /v2/{virtual}/tags/list` applies no member filter at all (`tags_list_virtual` fans out over the raw member set). Confirmed live in both runs above — it is 200 for a no-grant caller before *and* after this change, because it never reaches `authorize_virtual_members`. That is #3321's, left untouched on purpose.
4. `resolve_virtual_metadata` takes the unfiltered member set and relies on callers to pre-filter; `hex.rs:417` and `goproxy.rs:404` reach it with no visible pre-filter.


brandonrc:
Heads-up on a merge interaction with #3326 (issue #3325), which changes the same function. Both are green independently; the hazard is in the reconciliation, and it is not just textual.

**What each PR does to `authorize_virtual_members`:**

- This PR splits it: the body moves into a new `try_authorize_virtual_members`, and `authorize_virtual_members` becomes a thin `.unwrap_or_default()` wrapper.
- #3326 appends a second gate to the body — a `check_repository_action(.., "read", ..)` pass over the members that clear the existing grant filter — because the grant filter admits any grantee, so a `{write}`-only grant on a private member currently yields that member's manifest and blob bytes through a virtual.

**The trap.** Whoever rebases second will get a conflict in that function. If #3326's action gate is resolved into the **wrapper** rather than into `try_authorize_virtual_members`, the code compiles, both PRs' test suites still pass, and `tags_list_virtual` — the aggregating caller this PR is adding, which calls the fallible form — silently skips the action gate. That reintroduces the #3325 leak on exactly the path this PR is hardening, with no failing test to catch it, because #3326's regressions exercise `caller_can_read_member` / `authorize_virtual_members` and would still be green.

**Resolution rule:** the action gate belongs at the end of `try_authorize_virtual_members`, so both forms inherit it. The wrapper stays a pure `.unwrap_or_default()`.

Worth noting the two changes are complementary rather than overlapping in effect. Measured live on a stack built from `main`, a **public** virtual with a **private** member:

```
no-grant authenticated user, VIRTUAL tags/list      200  <- this PR
no-grant authenticated user, VIRTUAL manifest/blob  404  (already filtered)
{write}-only user,           VIRTUAL manifest/blob  200  <- #3326
{write}-only user,           DIRECT  tags/list      403
```

So this PR closes the unfiltered aggregation path and #3326 closes the action-blindness in the filter itself; neither subsumes the other. After both, a member is readable through a virtual exactly when a direct read of it would succeed, on every verb.

No objection to either merge order — flagging only so the conflict is resolved deliberately rather than by taking one side of the hunk.

Issue #3323, opened 2026-08-13
Virtual-repo member authorization is missing on ~40 read paths across 20 formats (3 are anonymous BYTE leaks)


brandonrc:
> **TOP LINE — three of these are ANONYMOUS BYTE leaks, not metadata leaks.** `GET /helm/{virtual}/charts/{f}`, `GET /nuget/{virtual}/v2/package/{id}/{ver}`, and `GET /swift/{virtual}/{scope}/{name}/{ver}/Package.swift` will each stream a **private** member repository's stored artifact bytes to an **unauthenticated** caller through a public virtual parent. Same shape as #3320, one step worse. Milestone/priority for those three needs a decision before this issue is scheduled.

## Summary

#3261 filtered the OCI virtual **download** walkers (`resolve_virtual_manifest`, `resolve_virtual_blob`) through `proxy_helpers::authorize_virtual_members`. #3320/#3321 found the third OCI walker (`tags_list_virtual`) took no `auth` at all.

That was not "one missed walker". It is a **class**. `proxy_helpers::fetch_virtual_members` returns every member of a virtual repository with no access control; `authorize_virtual_members` is the filter. The filter is applied on **13** call sites and **absent on ~40**, spanning 20 formats. The two shared metadata primitives — `resolve_virtual_metadata` and `collect_virtual_metadata` — take no `auth` parameter at all, so every caller of either inherits the gap.

Note: the helpers live in `backend/src/api/handlers/proxy_helpers.rs`, not `backend/src/services/`.

### Why this is reachable anonymously

`repo_visibility_middleware` (`backend/src/api/middleware/auth.rs:1725`) is layered over every format router (`backend/src/api/routes.rs:109-112`) and authorizes **only the URL repo** — the virtual parent. `should_allow_repo_access(is_public, has_auth) = is_public || has_auth` (auth.rs:1538), so an anonymous `GET` against a **public** virtual passes the gate. Everything the virtual then aggregates is a separate repository with its own ACL.

Nothing prevents the required configuration. `add_virtual_member` (`repositories.rs:8838`) requires the *actor* to see the member and hold `repository:admin` on the virtual; it imposes **no** constraint that the member be at least as visible as the parent. A public virtual over private members is a normal, supported setup — it is the setup #3261 and #3320 were about. No format restriction on virtual repos exists either.

### Classification rule used

* **Content-serving** — decides what bytes or metadata the caller receives. **Must** filter by caller visibility.
[...]

PR #3328, opened 2026-08-13
fix(virtual): authorize members on Helm, NuGet V2 and Swift byte downloads

### LINKED ISSUE #3324: Anonymous byte leak through public virtual repos in helm, nuget V2 and swift (split from #3323)
Split out of #3323 so the three **anonymous byte leaks** can ship in 1.7.2 without waiting on the ~40-site systemic remediation, which is not patch-release work.

## The three routes

Each streams a **private member's stored artifact bytes to an unauthenticated caller** through a public virtual parent:

| Route | Bytes served | Mechanism |
|---|---|---|
| `GET /helm/{virtual}/charts/{filename}` | chart `.tgz` / `.prov` | `helm.rs:530` `download_chart_via_index` → `local_fetch_by_path_suffix(member.id)`; handler `download_chart` (~L599) has **no auth extractor** |
| `GET /nuget/{virtual}/v2/package/{id}/{ver}` | `.nupkg` | `nuget.rs:108` `effective_local_repo_ids` → `v2_download` (~L2019) → `storage.get`; `v2_odata` (~L1813) has **no auth extractor** |
| `GET /swift/{virtual}/{scope}/{name}/{ver}/Package.swift` | manifest source from the member's zip | `swift.rs:696` `resolve_manifest_owner` → `fetch_manifest` (~L758) |

Preconditions confirmed at source, not assumed:
- `should_allow_repo_access = is_public \|\| has_auth` lets an anonymous caller past the middleware, which then authorizes only the **URL** repository — the virtual parent — never its members.
- `add_virtual_member` imposes no constraint that a member be at least as visible as its parent, so public-virtual-over-private-member is a supported configuration. It is precisely the configuration #3261 and #3320 addressed for OCI.

Two of the three are especially worth noting because a correct sibling exists **in the same file**, which is the strongest available evidence that these are oversights rather than design:
- NuGet **V3** `flatcontainer_download` **is** authorized; the legacy **V2** / Chocolatey path is the bypass.
- Swift's own `download_archive` **is** authorized; `version_path_handler` even **binds** `Extension(auth)` and then never threads it through.

## Why these three and not the other ~37

#3323 catalogues roughly 40 unfiltered call sites across 20 formats. Nearly all leak *metadata* — package names, versions, index documents. That is a real disclosure and it is tracked there, but it is a different severity from an unauthenticated caller downloading the private artifact itself, and remediating 40 sites across 20 formats carries regression risk that does not belong in a patch.

These three are bounded, share one shape, and each has a correct sibling to copy.

## Requirements

- Thread the caller through and filter members with the existing `authorize_virtual_members` helper. Do **not** invent a fourth predicate — match `resolve_virtual_manifest` / `resolve_virtual_blob` / `tags_list_virtual`, and for NuGet and Swift copy the authorized sibling already in the file.
- **Anonymous access to *public* members must keep working.** These are public virtual repositories; breaking anonymous pulls of legitimately public content would be a worse regression than the leak. Controls in both directions are mandatory.
- Fail **closed** on an authorization lookup error — an empty member set must not be reported as "not found". See the regression discussed on #3321, where swallowing a DB failure into an empty list turned a retryable `503` into a `404 NAME_UNKNOWN`.
- Live red→green per route, plus regression tests in `#[cfg(test)]` inside `src/` (the coverage gate is `--lib`; `backend/tests/` scores zero).

## Note for the release notes

Until this lands, 1.7.2 must not claim that virtual-repository laundering is closed generally. It is closed for **OCI**. That distinction belongs in the changelog either way, because #3323 will still be open.

brandonrc:
## Summary

Closes three anonymous artifact-byte leaks through public Virtual repositories in the Helm, NuGet V2 and Swift download paths.

The format-route middleware admits an unauthenticated caller whenever the URL repository is public, then authorizes only that repository — never the members a Virtual repository resolves through. Three download handlers walked the member set unfiltered from there, so a **public** Virtual parent streamed a **private** member's stored artifact to an anonymous caller:

- `GET /helm/{repo}/charts/{filename}` → `download_chart_via_index` served a hosted member's chart `.tgz`/`.prov` (handler `download_chart` bound no auth extractor at all).
- `GET /nuget/{repo}/v2/package/{id}/{version}` → `v2_download` resolved the `.nupkg` against every non-remote member via `effective_local_repo_ids` (`v2_odata` bound no auth extractor). The V3 sibling `flatcontainer_download` was already authorized.
- `GET /swift/{repo}/{scope}/{name}/{version}/Package.swift` → `resolve_manifest_owner` served a member's manifest source. Swift's own `download_archive` was already authorized, and `version_path_handler` already bound `Extension(auth)` but never threaded it through.

Each now threads the caller through and narrows the member walk with the existing `proxy_helpers::authorize_virtual_members` predicate — the same filter the OCI virtual walkers (`resolve_virtual_manifest`/`resolve_virtual_blob`/`tags_list_virtual`, #3261/#3320) and `resolve_virtual_download` use. No new predicate is introduced; this branch is rebased on the merged #3321 and uses its `try_authorize_virtual_members` directly (an identical interim copy carried by this branch was dropped on rebase). The fallible form is used so a failed visibility query surfaces as a retryable server error rather than being flattened into an empty set that these handlers would report as a definitive "not found". A member contributes bytes only when it is public or the caller holds a grant on it and it passes the token-scope ceiling; a denied member reads as an ordinary miss, so neither the bytes nor the member's existence leak. Anonymous pulls of **public** members through the same virtual repositories are unchanged, as are authorized callers.

**What this does NOT close: the metadata paths that lead to these downloads still enumerate private members.** Helm's merged `index.yaml` (`index_yaml`), the NuGet V2 OData feed (`load_hosted_v2_entries`, plus the V3 registration/search/version listings) and Swift's release listings (`list_releases` / version enumeration) all still aggregate private members' package names and versions to any caller who can reach the public virtual parent. A reviewer should not read this PR as closing the format: it closes the **byte** paths only. The metadata-level enumeration is a distinct, lower-severity gap tracked in #3323 (the systemic ~40-site catalogue), and the changelog entry states the same scope limit: virtual laundering is closed for OCI plus these three byte paths, not generally.

One pre-existing test is edited: `helm::tests::test_virtual_member_chart_download_applies_scan_policy_3220` probed the virtual chart download **anonymously** against a member repository that the test fixture creates private-by-default — i.e. it depended on exactly the unauthorized walk this PR removes. Its subject is the scan-policy gate, not the anonymous/private direction, so the fixture now publishes the member; the anonymous-vs-private direction is pinned by the new `virtual_member_authz_tests` in the same file. No other existing test needed changes (full `--lib` suite green).

Fixes #3324

Three DB-backed `#[cfg(test)]` regression tests live inside `src/` (the `--lib` coverage gate; `backend/tests/` scores zero):

- `helm::virtual_member_authz_tests::virtual_chart_download_does_not_leak_a_private_members_bytes_to_anon`
- `nuget::virtual_member_authz_tests::v2_package_download_does_not_leak_a_private_members_nupkg_to_anon`
- `swift::virtual_member_authz_tests::virtual_manifest_does_not_leak_a_private_members_source_to_anon`

Each asserts all three directions: an anonymous caller is denied a private member's bytes, a public member's bytes are still served anonymously (byte-for-byte), and the private member's granted principal still resolves it. Revert-proof, re-run after the rebase onto the merged #3317/#3321: reverting only the production hunks against current `main` (test modules kept) turns all three red — the anonymous request receives 200 with the private bytes — which also demonstrates that neither merged OCI change covers these three routes. The revert diff is insertions-only inside test modules and test scaffolding. Run with `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc:
Small heads-up now that this PR consumes the merged `try_authorize_virtual_members` rather than its own copy: #3326 (issue #3325) adds a read-action gate **inside that shared body**, so the three call sites this PR adds — `helm.rs`, `nuget.rs`, `swift.rs` — will inherit it automatically once both land.

Concretely, after both merge those byte downloads deny a principal whose only entitlement on a private member is a grant without `read` (a `{write}`-only fine-grained rule, or a role assignment to a role carrying no `read`), in addition to the anonymous/no-grant cases this PR is closing. That is intended and matches what a direct read of the member already does — flagging it only because it makes this PR's effective scope slightly wider than its description, which is worth a line in the CHANGELOG entry here if you want the behaviour attributed accurately.

No conflict between the two: #3326 touches `proxy_helpers.rs` and `oci_v2.rs` only, and does not modify `test_db_helpers.rs`, so your `link_virtual_member` / `cleanup_member_repo` additions are unaffected. Merge order does not matter.

Issue #3170, opened 2026-08-07
Backup silently discards unreadable artifact bytes and still reports completed


brandonrc:
Split out of #2863. That issue's root cause (backup resolves bytes from the deployment-wide storage root instead of the per-repo one) needs an archive-format change and is correctly milestoned 1.8.0. **This is the part that is patch-shaped, and it is arguably the more dangerous half.**

`backup_service.rs:605-608`:

```rust
for key in storage_keys {
    if let Ok(content) = self.storage.get(&key).await {
        artifact_data.push((key, content.to_vec()));
    }
}
```

A failed read is discarded. Not counted, not logged, not surfaced in the manifest, not reflected in the job status. Observed end to end on a filesystem deployment:

```
backup completed: artifact_count=0
restore: artifacts_restored=0 errors=[]
```

The backup says **completed**. The restore says **zero errors**. The artifacts are gone. An operator has no signal at any point that the archive is empty of content.

## Why this is worth fixing on its own

A backup that fails loudly is an incident. A backup that reports success while containing nothing is a **false sense of safety** — it is discovered only during a restore, which is exactly the moment when there is no recourse. The underlying path bug (#2863) means every filesystem deployment is in this state today and cannot tell.

This fix does not repair the mismatch. It converts an undetectable failure into a detectable one, which is independently valuable and does not touch the archive format.

## Suggested change

* Count read failures alongside successes while collecting.
* Record both in `BackupManifest` — e.g. `artifact_count` plus `artifacts_unreadable` — so an existing archive can be audited without re-running.
* Log each miss at `warn` with the key and the underlying error, rather than discarding the `Err`.
* Fail the job, or mark it `completed_with_errors`, when any blob could not be read. A backup missing content should not present as clean. If a partial backup must remain possible, it should be opt-in rather than the silent default.
* Apply the same treatment to restore, which currently reports `errors=[]` while restoring nothing.

## Test

Seed an artifact row whose bytes are not readable at the resolved root, run a backup, and assert the job does **not** report clean success and the manifest records the miss. That test fails today and will keep passing after #2863's real fix lands, since it asserts on honesty rather than on the path.

## Related

* #2863 — the root-resolution bug, 1.8.0, needs the archive-format change
* #3171 — the same miss class on S3 deployments with `S3_PREFIX` set
* OCI blobs live in `oci_blobs` rather than `artifacts`, so `artifact_storage_keys` never enumerates them; they are missing from every archive for a third, independent reason

PR #3330, opened 2026-08-13
fix(backup): report a failed backup run on the resource instead of a 500

### LINKED ISSUE #2863: Backups skip repo-scoped filesystem artifact bytes (storage_key scoping mismatch)
Found while implementing #2772 (backup repo exclusion), pre-existing and unrelated to that fix.

For repo-scoped filesystem storage, format handlers (e.g. Maven) write objects under a **repo-scoped** key like `<repo_key>/maven/...`, but `artifacts.storage_key` holds the **unscoped** path. The backup path (`get_artifact_storage_keys` → object fetch in backup_service) reads `artifacts.storage_key` and therefore looks for the bytes at the unscoped location, so repo-scoped filesystem artifact **bytes are not captured in the backup** (the DB rows are, the blobs aren't).

Impact: a restore from such a backup would be missing artifact content for repo-scoped filesystem repositories. Discovered because a DTF tier had to realign `storage_key` in-oracle to make exclusion observable.

Fix direction: reconcile the backup's object-key resolution with the actual repo-scoped write path (or normalize storage_key at write time). Needs a repro tier proving backup→restore round-trips the bytes for a repo-scoped filesystem Maven repo.

### LINKED ISSUE #3327: Backup execute returns 500 on any instance holding artifacts: a recorded run failure is reported as a transport error and its message discarded
Found by the v1.7.2 Release Gate rehearsal (`artifact-keeper-test` run 31666770835, `admin-tests` → `tests/admin/test-backup-lifecycle.sh`):

```
--- POST /admin/backups/{id}/execute returns 202 ---
  FAIL: expected 202/200 from execute, got 500
```

`admin-tests` passed in the v1.7.0 and v1.7.1 release runs and the test script has not changed since 2026-05-16, so this is new in this release.

## What actually happens

Two things compose.

**1. On a filesystem deployment the backup cannot read *any* artifact's bytes — #2863, pre-existing.** Format handlers write objects under the repository's own root, so the object lives at `<repo storage_path>/<storage_key>`, while `BackupService::do_backup` reads `artifacts.storage_key` through the deployment-wide handle and looks for it at `<deployment root>/<storage_key>`. Every read misses. Confirmed on a stack built from current `main`:

```
/data/storage/test-stats-.../a1/36/a136b2c2a535f8fa103a4c272f9a7d809f5af8346e2cafe51ce4f8009e323d8e   <- where the bytes are
storage_key = a1/36/a136b2c2a535f8fa103a4c272f9a7d809f5af8346e2cafe51ce4f8009e323d8e                    <- where the backup looks
```

**2. #3170 (this release) turned that miss from silent into fatal, and the failure surfaces as a transport error.** The read failure is now counted, logged, recorded in the manifest, and fails the job unless `allow_partial_artifacts` is set — all intended. But `execute_backup` calls `BackupService::execute`, whose `Err` propagates through `Result<Json<_>>`, so the caller gets:

```
HTTP 500
{"code":"INTERNAL_ERROR","message":"Internal server error"}
```

The message #3170 composes for exactly this moment — naming every object it could not read — is written to `backups.error_message` and to the server log, and then discarded at the HTTP boundary. Reproduced against `main`, server log:

```
ERROR backup_service: backup is missing artifact content: 3 artifact object(s) could not be read:
a1/36/a136b2c2… (Resource not found: Storage key not found: a1/36/a136b2c2…); … .
The archive does NOT contain these bytes. Set `allow_partial_artifacts` in the backup metadata
to accept a partial archive deliberately.
```

## Scope

`POST /api/v1/admin/backups/{id}/execute` returns 500 on **any** instance holding at least one artifact. An empty instance still returns 200 — verified — so this is the opposite of "nothing to back up looks like a failure": nothing to back up is the only case that succeeds.

## The defect being fixed here

A run that reaches a terminal FAILED state has already recorded its own outcome on the backup row (`status`, `error_message`). That is a fact about the backup, not about the request, which was understood and carried out. Reporting it as `500 INTERNAL_ERROR` both breaks the endpoint's own documented `BackupResponse` contract and throws away the causal message, leaving an operator with a bare 500 and no way to learn why.

`execute_backup` will answer with the backup resource: HTTP 200, `status: "failed"`, `error_message` carrying the message that names the unreadable objects. A successful run answers `status: "completed"`. Nothing becomes silent — the failure is more visible than it is today, because it now reaches the caller instead of only the log. `BackupService::execute` keeps its `Err`-on-failed-run shape for the scheduler, whose `backup_schedule_runs` bookkeeping depends on it.

## What this does not fix

#2863. Backups on filesystem deployments still capture no artifact bytes; they now say so, loudly and specifically, on the resource the operator is looking at. The archive-format and restore-side work that makes the bytes actually round-trip stays where it is, in 1.8.0.

brandonrc:
Fixes #3327

Release-blocking: `admin-tests` → `tests/admin/test-backup-lifecycle.sh` fails in the v1.7.2 Release Gate rehearsal (`artifact-keeper-test` run 31666770835) with `expected 202/200 from execute, got 500`. It passed in the v1.7.0 and v1.7.1 release runs and the test script has not changed since 2026-05-16.

## Root cause

Reproduced on a stack built from `main`, not inferred from the diff.

`POST /api/v1/admin/backups/{id}/execute` returns 500 on **any** instance holding at least one artifact. An empty instance returns 200, so this is the opposite of "nothing to back up looks like a failure" — nothing to back up is the only case that still succeeds.

Two things compose:

1. **#2863 (open, 1.8.0, pre-existing):** format handlers write objects under the repository's own root, so the bytes live at `<repo storage_path>/<storage_key>`, while the backup reads `artifacts.storage_key` through the deployment-wide handle at `<deployment root>/<storage_key>`. On a filesystem deployment *every* artifact read misses. Observed directly:

   ```
   /data/storage/test-stats-…/a1/36/a136b2c2…   <- where the bytes are
   storage_key = a1/36/a136b2c2…                <- where the backup looks
   ```

2. **#3170 (this release):** the miss is no longer discarded — it is logged, counted into the manifest's `artifacts_unreadable`, and fails the job unless `allow_partial_artifacts` is set. All intended. But the trigger handler called `BackupService::execute`, whose `Err` propagates through `Result<Json<_>>`, so the caller got:

   ```
   HTTP 500
   {"code":"INTERNAL_ERROR","message":"Internal server error"}
   ```

   The message #3170 composes for exactly this moment — naming every object it could not read — reached `backups.error_message` and the server log, and was then discarded at the HTTP boundary.

## The change

A run that reaches a terminal state has already recorded its own outcome on the backup row. That is a fact about the backup, not about the request, which was understood and carried out. So the endpoint answers with the `BackupResponse` it already documents:

* `200` + `status: "failed"` + `error_message` naming the cause, for a failed run.
* `200` + `status: "completed"` for a clean one. (Previously the success path answered `status: "in_progress"` — `do_backup` loads the row before the status is written. Re-read on both arms, so a client told to trust `status` gets a status that is true.)
* `409` / `404` / `500` unchanged: another backup in progress, no such backup, and "the run could not be started or its outcome could not be recorded".

Nothing became quieter. The failure now reaches the caller instead of living only in the log.

`BackupService::execute` keeps its `Err`-on-failed-run shape — the scheduler's `backup_schedule_runs` bookkeeping keys off it — and is now a thin wrapper over the new `execute_run`, which returns `Ok(BackupRun { backup, failure })`. The existing #3170 tests assert `execute()` still errs and still pass unchanged.

**This does not fix #2863.** Backups on filesystem deployments still capture no artifact bytes; they now say so, specifically, on the resource the operator is looking at. The archive-format and restore-side work stays in 1.8.0.

## Live proof

Same stack, same database and storage volume, only the image swapped. 214 artifacts present, seeded by replaying the e2e suites that run before `admin-tests` in the gate.

Before (`main`):

```
--- POST /admin/backups/{id}/execute returns 202 ---
  FAIL: expected 202/200 from execute, got 500
Results: 3 passed, 1 failed (4 total)
```

After (this branch):

```
Results: 4 passed, 0 failed, 0 skipped (4 total)          test-backup-lifecycle.sh
Results: 6 passed, 0 failed, 0 skipped (6 total)          test-backup-lifecycle-depth.sh
```

Response body on the failing-run path:

```
HTTP 200
status:        failed
error_message: backup is missing artifact content: 214 artifact object(s) could not be read:
               a1/36/a136b2c2… (Resource not found: Storage key not found: …); …
```

and on a clean run (backup scoped to an empty repository):

```
{"status":"completed","artifact_count":0,"error_message":null}
```

## Regression test

`api::handlers::admin::tests::execute_reports_a_failed_run_as_the_backup_resource_not_a_500` — in `#[cfg(test)]` inside `src/`, so `cargo nextest run --workspace --lib` scores it. It drives the real router, asserts the status is not 5xx **and** that the body carries `status: "failed"` plus the message naming the unreadable key; asserting only the status code would pass for a handler that reported success.

Revert-proof, with `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1` exported:

```
production code reverted to merge-base, test kept:
  FAIL [0.251s] … panicked: a recorded run outcome must not surface as a transport error;
                 got 500 Internal Server Error body={"code":"INTERNAL_ERROR",…}

with the fix:
  PASS [0.362s]
```

The revert was faithful — `backup_service.rs` byte-identical to merge-base, and `admin.rs` differing by a single 121-line insertion at line 2583, inside the `#[cfg(test)] mod tests` that begins at line 1481:

```
 backend/src/api/handlers/admin.rs | 121 +++++++++++++++++++++++++
 1 file changed, 121 insertions(+)
@@ -2582,4 +2582,125 @@ mod tests {
```

Full backup suite: `120 tests run: 120 passed`. `cargo fmt --all` and `clippy --workspace --all-targets -D warnings` clean. No new `sqlx::query!` macros, so no `.sqlx` regeneration.


brandonrc:
Added the upgrade note to the CHANGELOG (`### Upgrade note — backups on filesystem storage start reporting ``failed``, and your existing archives never held artifact bytes`), matching the depth of the existing `[Unreleased]` notes: mechanism, "Am I affected?", how to audit an archive you already have, why object storage is not an in-place workaround, and exactly what `allow_partial_artifacts` does and does not buy.

The read-side fixability assessment for #2863 is recorded on that issue: https://github.com/artifact-keeper/artifact-keeper/issues/2863#issuecomment-5276480465 — short version, the read change is an afternoon but cannot ship alone, because the archive's flat key namespace silently collides once reads succeed, the restore side writes to the deployment root and would report a clean run over misplaced bytes, and `artifacts_unreadable` would drop to 0 so the archive would stop declaring itself incomplete. Recommend it stays whole in 1.8.0.

PR #3333, opened 2026-08-13
fix(pypi): serve PEP 658 .whl.metadata through the proxy cache

### LINKED ISSUE #3300: PyPI PEP 658 .whl.metadata requests bypass cache and refetch upstream
## Description

Artifact Keeper 1.7.1 does not reuse cached PyPI PEP 658 `.whl.metadata` responses. Repeated requests for the same metadata URL trigger the complete upstream resolution path again, significantly slowing uncached `uv lock` operations.

## Environment Information

- Deployment type: Kubernetes
- Backend version: 1.7.1
- Package format: PyPI
- Database: PostgreSQL 16
- Client: uv 0.12.x
- Backend replicas: 1
- Storage: filesystem on persistent volume

## Expected Behavior

After the first successful request, subsequent requests for the same `.whl.metadata` URL should be served from Artifact Keeper cache without refetching the upstream simple index or the external metadata file.

## Observed Behavior

The same URL was requested twice:

```text
/pypi/pypi-remote/simple/requests/requests-2.34.2-py3-none-any.whl.metadata
```

Measured response times:

```text
Request 1: 331 ms
Request 2: 115 ms
```

Both requests produced the following upstream sequence in the backend logs:

```text
Fetching upstream: https://pypi.org/simple/requests/
Fetched 82826 bytes

Fetching upstream: https://files.pythonhosted.org/.../requests-2.34.2-py3-none-any.whl.metadata
Fetched 4806 bytes
```

The response bodies were identical, but the second request was not served as a cache hit.

## Steps to Reproduce

1. Configure an Artifact Keeper PyPI remote repository.
2. Run Artifact Keeper 1.7.1.
3. Enable debug logging:

```text
RUST_LOG=info,artifact_keeper::services::proxy_service=debug,artifact_keeper::api::handlers::pypi=debug
```

4. Request the same PEP 658 metadata URL twice:

```text
/simple/requests/requests-2.34.2-py3-none-any.whl.metadata
```

5. Compare the backend logs.
6. Observe that both requests refetch the upstream simple index and `.whl.metadata` file.

## Impact

Our CI regenerates multiple `uv.lock` files concurrently using `uv lock --no-cache`. Each resolver requests many PEP 658 metadata files. With four concurrent runners, the backend receives repeated requests for the same metadata URLs, causing hundreds of redundant upstream requests, increased backend CPU usage, and significantly slower dependency resolution.

## Related Issue

- #3076: incorrect or missing PEP 658 metadata hashes.

## Requested Fix

Please verify that PEP 658 `.whl.metadata` requests use the normal proxy cache path and cache key. Add a regression test that requests the same metadata URL twice, verifies only the first request reaches upstream, verifies the second request is a cache hit, and confirms the returned bytes and headers remain identical.

## Checklist

- [x] Searched existing issues; no duplicate found for this cache behavior.
- [x] Redacted internal URLs and credentials.
- [x] Reproduced on supported backend version 1.7.1.
- [x] Included logs, timing evidence, and reproduction steps.

axellpadilla:
## Summary

Closes #3300.

`serve_remote_metadata` fetched the PEP 658 sidecar with `fetch_upstream_direct_with_link`, which bypasses the proxy cache in **both** directions: the bytes were never stored, so no request could ever hit. Every `.whl.metadata` request re-ran the entire upstream resolution — the simple index (read through `proxy_fetch_uncached` to rediscover the distribution's download URL) plus the sidecar itself — two upstream round-trips for a resource that cannot change. An uncached `uv lock` fan-out requests one sidecar per candidate distribution, so a CI job regenerating several lock files multiplied that across every resolver run, which is what the issue reports.

This is the path #3077 added, so it affects **both direct Remote repositories and Virtual members** — virtual metadata resolution routes through the same function.

Two changes:

- **Probe the proxy cache before resolving the fetch target.** The ordering is load-bearing: resolving reads the simple index uncached, so a probe behind it would still pay an upstream round-trip on every hit. Keyed on `simple/{project}/{dist}.metadata` — what `resolve_pypi_remote_fetch_target` already computes, and distinct from the distribution's own key. A miss then fetches through `fetch_artifact_with_cache_path_capped`, which stores the body together with its upstream `Content-Encoding`, so a later hit re-emits the coding #3193 established rather than serving coded bytes bare.
- **Classify the sidecar as immutable.** `simple/<pkg>/<dist>.whl.metadata` matched no PyPI distribution extension and took the mutable default — a five-minute TTL and a revalidation against upstream after it, even once stored. A PEP 658 sidecar holds the `METADATA` of an immutable distribution, so the suffix is now stripped ahead of the extension test, exactly as the Maven checksum/signature sidecars are.

The three arms that produce a sidecar (cache hit, upstream fetch, wheel extraction) now share one response builder, so a cached serve is byte- and header-identical to the upstream serve that filled it.

Deliberately out of scope: the upstream-`404`-to-wheel-extraction fallback is unchanged, so an upstream that advertises metadata it does not serve still re-extracts from its (cached) wheel per request. Caching a synthesized body under a key upstream 404s for would mask an upstream that later starts serving it.

`test_remote_pypi_metadata_is_served_from_cache_on_repeat` asserts the upstream request **count**, which is the real discriminator — the sidecar's bytes are correct either way, so only the count separates a cache hit from a silent refetch. On `main` it fails with `left: 4, right: 2`; on this branch it is 2. It lives in-crate beside the other `serve_remote_metadata` tests so the coverage `--lib` run exercises the async body, following the precedent set in #3077.

## Test Checklist

- [x] Manually tested locally

Local run of the PR CI surface, against a real Postgres:

```text
cargo fmt --all -- --check                                   pass
scripts/ci/check-no-legacy-admin-scope.sh                    pass
scripts/ci/check-token-mint-surface.sh                       pass
scripts/ci/check-auth-event-audit.sh                         pass
scripts/ci/check-migration-ledger.sh                         pass
cargo clippy --workspace --all-targets -- -D warnings        pass
cargo nextest run --workspace --lib --test-threads 8         15144 passed, 0 failed
integration suites (14 named, --ignored --test-threads=1)    123 passed, 0 failed
incus cross-repo session regression                          pass
shell-tests (dtrack-init, trivy-gate, release-preflight)     pass
cargo audit                                                  exit 0 (4 allowed warnings)
new-code coverage (CI's diff-vs-merge-base calc)             96% (91/94), gate 70%
total coverage floor                                         87.89%, gate 50%
jscpd duplication on changed measurable files                0.0%, gate 3%
```

Not run locally: CodeQL, the image builds, and smoke-e2e (needs the built image and full stack). Rust-only change; no Dockerfile, compose, or migration edits.

## API Changes

- [x] N/A - no API changes


axellpadilla:
@brandonrc could you include a version of this one on next release, it was a heavy impact on performance, and rollbacks are only possible with a database drop or backup (separate issue, it checks for database migration version and doesn't start, safely doable for a cache but want to avoid it)

brandonrc:
I will get this one out asap! Thanks @axellpadilla

brandonrc:
Going to do a quick review, if I do not find any blockers I will merge this into main tonight. Then hopefully have a 1.7.5 release soon. Catching some security issues thanks to kimi k3 and some redteaming.

brandonrc:
Reviewed against `4cc0d41`. This looks good to merge.

The two ways a change like this fails silently both check out:

- **Probe key equals write key.** The probe builds `simple/{project.as_str()}/{dist}.metadata` (`pypi.rs:2241`) while the write re-normalizes through `PypiHandler::normalize_name` (`pypi.rs:3388`). Those two normalizers genuinely disagree on non-canonical input (`normalize_pep503("acme sdk")` is `acmesdk`, `normalize_name` is `acme-sdk`), so this is only safe because `project` is a `NormalizedProjectName` and `normalize_name` is the identity on canonical input, which `pypi_name.rs` pins with a test. Worth knowing that the newtype is what keeps the cache hitting: relaxing that parameter back to `&str` would turn the probe into a permanent miss with every test still green.
- **The 404 fallback survives the fetch function swap.** `fetch_artifact_with_cache_path_capped` still returns `AppError::NotFound` on an upstream 404, including the negative-cache short circuit (`proxy_service.rs:3019`), so the wheel extraction arm is still reachable. It also has leader/follower single flight, so a cold fan-out does not stampede.

Also confirmed: curation runs at `pypi.rs:1689`, ahead of `serve_remote_metadata` at 1727, so the early cache return does not let a cached sidecar outlive a block rule added later. The virtual arm reaches the same function with `member.id` / `member.key` (`pypi.rs:1982`), so the claim about virtual members holds and probe and write agree per member.

Two review notes, neither blocking, both filed as #3356 rather than asked for here:

1. The regression test does not discriminate the classifier half of this PR. `cached_metadata_if_servable` serves only on `Freshness::Fresh`, and `evaluate` returns `Fresh` for a mutable entry while `now < expires_at` with a 300s default. The warm request follows the cold one within milliseconds, so a sidecar classified mutable is still fresh and the probe hits either way. Revert `cache_classifier.rs` alone and the test still passes. The predicate has table coverage, so this is an end to end gap rather than untested code, and covering it properly needs clock injection this path does not have.
2. The `.metadata` strip in `is_pypi_package_file` also applies to Conda, which shares `classify_pypi` (`cache_classifier.rs:182`). Poetry is a correct beneficiary, Conda is not. Exposure is close to nil since no conda resource has that shape and storing one needs an upstream 200, so gating it on format is probably more churn than it is worth.

#3356 also tracks the pre-existing uncached simple index read on the cold path, which this PR correctly does not claim to address.

PR #3346, opened 2026-08-14
feat(security): proxy-cached artifact scan visibility and SBOM support

### LINKED ISSUE #3344: Stale user-facing message: says proxy-cached artifacts cannot be security scanned, but proxy scanning shipped in #2954
## Summary

`ARTIFACT_NOT_ANALYZABLE_MSG` (`backend/src/api/handlers/sbom.rs`) tells callers:

> Artifact not found or not eligible for analysis: SBOM generation and security scanning are available only for artifacts hosted in this registry, not proxy-cached remote artifacts.

The SBOM half is still correct. **The security-scanning half has been false since #2954.**

## Why it is wrong

`backend/src/services/proxy_scan_service.rs` exists precisely to scan proxy-cached bytes, and says so in its own header:

> Proxy-cached bytes are deliberately NOT written to `artifacts` (#1278/#1280), so the `artifact_id`-keyed `scan_results` pipeline cannot hold a verdict for a proxied object. This service persists a content-addressed (`checksum_sha256`) verdict in `proxy_scan_results`, independent of `artifacts`, so that a repeat pull of a known-vulnerable digest is blocked WITHOUT re-fetching upstream or re-scanning, and a verdict is shared across repos/tenants pulling identical bytes.

The serve-time gate is `proxy_helpers::gate_proxy_scan_serve`, fed by `scanner_service::run_inline_proxy_scanners`, with per-repository enablement (`is_proxy_scan_enabled`) and a configurable fail direction (`ProxyScanAction::FailOpen` / `FailClosed`).

## How it drifted

Ordinary sequencing, worth recording so the fix targets the cause:

- The message was added by **#2227**, closed **2026-07-07**, when the statement was true.
- Proxy scanning shipped in **#2954**, closed **2026-07-29** — three weeks later — and deliberately built a *parallel*, digest-keyed pipeline rather than extending the `artifacts`-keyed one. Because it touched neither `sbom.rs` nor the `artifacts` table, nothing pointed at this string.

## Why it matters beyond wording

This message is the answer a user gets when they ask "why can't I scan this?" It currently tells them proxy-cached artifacts **cannot** be scanned, when the product's answer is that they are scanned by a different subsystem and can be blocked on download. A user acting on it would reasonably conclude they must mirror everything into hosted repositories to get coverage they already have — which is a significant and unnecessary architectural change to make on the strength of a stale sentence.

## Requested

1. Correct the security-scanning clause. Keep the SBOM clause, which is accurate, and say what *does* apply to proxied objects — digest-keyed verdicts in `proxy_scan_results`, enforced at serve time when proxy scanning is enabled for the repository.
2. Point the reader somewhere useful. The message's value is that it explains the expected case rather than saying "not found"; that intent is right and worth preserving.
3. Check `signing.rs`, which contains the same "proxy-cached remote artifacts" phrasing, for the same drift.
4. Worth considering: the two pipelines are keyed differently by design (`artifacts.id` vs `checksum_sha256`), so any user-facing text that says "scanning" without qualifying which one will drift again. Naming the subsystem in the message would make the next divergence self-evident.

### LINKED ISSUE #3394: Proxy-cached artifact security visibility and SBOM support
Tracking issue for proxy-cached artifact security visibility, shipped on `feat/proxy-scan-visibility`.

## Background

Proxy-cached artifacts are scanned at download time and blocked on policy violation (#2954, #3003), but none of it was visible. Worse, the API told users scanning was unavailable for them, and the UI rendered a green all-clear for artifacts the gate had just returned 403 for.

Proxy content deliberately has no `artifacts` row (#1278/#1280), so every artifact-keyed read path — CVE history, SBOM, security score — is structurally empty for it. That invariant is the root of everything below.

## Shipped

- [x] Correct the stale API messaging; scope claims to the formats that actually have gate wiring (PyPI, npm, Docker/OCI)
- [x] Fix inflated `findings_count` — a wheel is cataloged twice (own METADATA + synthetic pin), so counts were doubled
- [x] `GET /repositories/{key}/security/proxy-scans` — verdict summary and cached-path listing, path-keyed with enforcement context
- [x] `GET /repositories/{key}/security/proxy-sbom` — CycloneDX/SPDX generated on demand from the recorded inventory
- [x] Persist the package inventory the inline scan already computes (`proxy_scan_packages`, migration 197)
- [x] Report grype's full component catalog rather than only CVE-matched packages
- [x] Exclude cached upstream index responses from artifact listings and scan counts

## Design notes

Proxy SBOMs are **not** stored in `sbom_documents`. That table's `artifact_id` is `NOT NULL` with an FK to `artifacts`; making it nullable would touch every SBOM query in the codebase and put a cross-tenant `ON DELETE CASCADE` in the path. The document is regenerated per request from the stored inventory instead. No existing SBOM handler was modified.

Lookups are **path-keyed, never digest-keyed**. Verdicts and inventories are global by content digest, so accepting a digest parameter would turn these endpoints into a cross-tenant lookup oracle.

## Known gaps

Tracked separately — see linked issues.

Design docs: `docs/plans/2026-08-13-proxy-scan-visibility-design.md`, `docs/plans/2026-08-14-proxy-sbom-design.md`

brandonrc:
## Summary

Makes proxy-cached artifact security visible, and gives proxy-cached artifacts an SBOM.

Proxy-cached artifacts are scanned at download time and blocked on policy violation, but none of it was visible — and the API actively told users scanning was unavailable for them. Proxy content deliberately has no `artifacts` row (#1278/#1280), so every artifact-keyed read path is structurally empty for it.

Fixes #3344
Fixes #3394

### API

- `GET /repositories/{key}/security/proxy-scans` — verdict summary and cached-path listing, with `scan_on_proxy` / `proxy_scan_action` enforcement context so `vulnerable` is not ambiguous between a repository that blocks and one that serves anyway
- `GET /repositories/{key}/security/proxy-sbom` — CycloneDX or SPDX, generated on demand

Both are **path-keyed, never digest-keyed**: verdicts and inventories are global by content digest, so a digest parameter would be a cross-tenant lookup oracle. Both authenticate unconditionally, including on public repositories, where `require_visible` would otherwise return early.

### Scanner

- Report grype's full component catalog rather than only CVE-matched packages. Grype builds the catalog on every scan; requesting `-o json` alone discarded it, so clean artifacts produced no inventory and vulnerable ones produced an SBOM listing only their vulnerable components
- Persist that inventory (`proxy_scan_packages`, migration 197), digest-keyed so identical content cached in many repositories stores one copy
- Fix inflated `findings_count`: a wheel is cataloged twice — its own METADATA plus the synthetic pin — so counts were doubled
- Exclude cached upstream index responses from artifact listings and scan counts. They were being counted as `not_scanned` with reason `unknown`, a permanent false positive no action could clear

### Design

Proxy SBOMs are **not** stored in `sbom_documents`. `artifact_id` there is `NOT NULL` with an FK to `artifacts`; making it nullable would touch every SBOM query in the codebase and put a cross-tenant `ON DELETE CASCADE` in the path. Documents are regenerated per request from the stored inventory. **No existing SBOM handler or query was modified.**

Docs: `docs/plans/2026-08-13-proxy-scan-visibility-design.md`, `docs/plans/2026-08-14-proxy-sbom-design.md`

## Test Checklist

- [x] `cargo fmt --check`
- [x] `cargo clippy --workspace --all-targets -- -D warnings`
- [x] `cargo test --workspace --lib` — 15201 passing
- [x] Verified end to end on a live stack: blocking, verdict listing, SBOM generation, index exclusion

## API Changes

Two new authenticated, repository-scoped GET endpoints. One new table and index (migration 197). No existing endpoint changed shape.

Deferred and tracked separately: per-CVE detail for proxy content (#3395), on-demand rescan (#3396), blast-radius proxy coverage (#3397).

### DIFF
--- a/.sqlx/query-0ad0c56e6850c91ee5dab6835aeb1f3b84d22fbfdfdef2b691e813630edc1999.json
+++ b/.sqlx/query-0ad0c56e6850c91ee5dab6835aeb1f3b84d22fbfdfdef2b691e813630edc1999.json
@@ -1,0 +1,14 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "DELETE FROM users WHERE auth_provider = 'saml' AND username LIKE $1",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": [
+        "Text"
+      ]
+    },
+    "nullable": []
+  },
+  "hash": "0ad0c56e6850c91ee5dab6835aeb1f3b84d22fbfdfdef2b691e813630edc1999"
+}
--- a/.sqlx/query-33cf2fa69cd12631e8891288d7de68f9ff9f9a95f957c0b6f98b4257b493267f.json
+++ b/.sqlx/query-33cf2fa69cd12631e8891288d7de68f9ff9f9a95f957c0b6f98b4257b493267f.json
@@ -1,0 +1,26 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT\n            COALESCE(SUM(hosted_bytes + proxy_bytes + oci_bytes), 0)::BIGINT as \"total!\",\n            COALESCE(SUM(proxy_bytes), 0)::BIGINT as \"proxy!\"\n        FROM repository_usage_ledger\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "total!",
+        "type_info": "Int8"
+      },
+      {
+        "ordinal": 1,
+        "name": "proxy!",
+        "type_info": "Int8"
+      }
+    ],
+    "parameters": {
+      "Left": []
+    },
+    "nullable": [
+      null,
+      null
+    ]
+  },
+  "hash": "33cf2fa69cd12631e8891288d7de68f9ff9f9a95f957c0b6f98b4257b493267f"
+}
--- a/.sqlx/query-3b8135df633be47e2a2094e401cfc2dd41c62649519ee43ca02782c6afb2da19.json
+++ b/.sqlx/query-3b8135df633be47e2a2094e401cfc2dd41c62649519ee43ca02782c6afb2da19.json
@@ -1,0 +1,12 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "DELETE FROM users WHERE email = '' AND auth_provider = 'oidc' AND username LIKE 'stranded\\_%'",
+  "describe": {
+    "columns": [],
+    "parameters": {
+      "Left": []
+    },
+    "nullable": []
+  },
+  "hash": "3b8135df633be47e2a2094e401cfc2dd41c62649519ee43ca02782c6afb2da19"
+}
--- a/.sqlx/query-45fb62f316727f921d1f61acd90eb3aea0524dc0b1319f3cef2904a2a49453ad.json
+++ b/.sqlx/query-45fb62f316727f921d1f61acd90eb3aea0524dc0b1319f3cef2904a2a49453ad.json
@@ -1,0 +1,22 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "SELECT COUNT(*) AS \"count!\" FROM proxy_cache_artifacts\n           WHERE repository_id = $1 AND checksum_sha256 IS NOT NULL",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "count!",
+        "type_info": "Int8"
+      }
+    ],
+    "parameters": {
+      "Left": [
+        "Uuid"
+      ]
+    },
+    "nullable": [
+      null
+    ]
+  },
+  "hash": "45fb62f316727f921d1f61acd90eb3aea0524dc0b1319f3cef2904a2a49453ad"
+}
--- a/.sqlx/query-6a0ab36eb33a16e2e2004aaf55f65d0f48163433baa4763a06173231749728ba.json
+++ b/.sqlx/query-6a0ab36eb33a16e2e2004aaf55f65d0f48163433baa4763a06173231749728ba.json
@@ -1,0 +1,35 @@
+{
+  "db_name": "PostgreSQL",
+  "query": "\n        SELECT name, version, created_at\n        FROM artifacts\n        WHERE id = $1\n          AND repository_id = $2\n          AND is_deleted = false\n        ",
+  "describe": {
+    "columns": [
+      {
+        "ordinal": 0,
+        "name": "name",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 1,
+        "name": "version",
+        "type_info": "Varchar"
+      },
+      {
+        "ordinal": 2,
+        "name": "created_at",
+        "type_info": "Timestamptz"
+      }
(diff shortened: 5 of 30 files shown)


brandonrc:
Superseded by #3412 (`fix/proxy-cache-scan-visibility`), which **carries every commit from this branch forward by merge** — nothing here was reimplemented, and the history and authorship are preserved.

Two things had to change, and neither could be done on this branch without also renumbering it:

1. **Migration slots 196 and 197 are taken.** `main` has since landed `196_oci_manifest_subjects.sql` and `197_oidc_env_seeded.sql`, so this branch's `196_proxy_cache_repo_checksum_index.sql` and `197_proxy_scan_packages.sql` collide. That is what the `duplicate migration version slot(s)` error in the failing Backend Unit Tests job was reporting. Both are consolidated into the single slot **202** on the new branch, together with the `proxy_scan_findings` table.

2. **`proxy_scans_path_lookup_is_repo_scoped_and_floors_scanned_at` was failing for a real reason, not flaking.** It compared a `chrono::Utc::now()` carrying nanoseconds against its own round trip through a `timestamptz` column, which stores microseconds:

   ```
   left:  Some(2026-08-14T14:30:31.670554Z)
   right: Some(2026-08-14T14:30:31.670554965Z)
   ```

   It fails whenever the clock hands back a non-zero nanosecond remainder — near-always on CI, rarely on a developer's machine, which is exactly the asymmetry that makes a wrong test look flaky. The timestamp is now truncated at the source. The Code Coverage job was red only because the test run it wraps exited non-zero, and CI Complete only because of those two.

The new PR also closes the three gaps this one explicitly deferred:

- **#3395** — per-CVE detail, persisted to a digest-keyed `proxy_scan_findings` table and served on a single-path read, in the shape this PR's description proposed.
- **#3396** — on-demand rescan of already-cached bytes, recording through the same `proxy_scan_and_record` write path the inline gate uses.
- **#3397** — proxy exposure in the blast-radius report, which #3395 unblocked.

Plus the follow-through on #3344 point 2: the capability messages now name the endpoint that answers the question rather than only asserting one exists.

**This PR should be closed in favour of #3412.** Please don't merge both — the migration slots would collide again.

brandonrc:
Closing as **superseded by #3412**, which is now merged.

To be clear about what happened to the work here: #3412 did **not** reimplement it. It merged this branch forward, so all 24 commits and their authorship are preserved in the merged history, then extended the result to cover #3395, #3396 and #3397 as well.

Both CI failures on this branch were real, not flakes, and are worth recording:

1. **Migration slot collision.** This branch claimed 196/197, which `main` has since taken (`196_oci_manifest_subjects`, `197_oidc_env_seeded`) — that is the `duplicate migration version slot(s)` error in the Backend Unit Tests log. The migrations were consolidated into slot 202 in #3412.
2. **`proxy_scans_path_lookup_is_repo_scoped_and_floors_scanned_at`** compared a nanosecond `Utc::now()` against its own microsecond `timestamptz` round trip. That fails nearly always on CI and only rarely locally, which is exactly why it read as flaky. Truncated at the source in #3412.

Code Coverage and CI Complete were red only downstream of the second failure.

Issues closed by #3412: #3344, #3394, #3395, #3396, #3397.

PR #3357, opened 2026-08-14
fix(maven): treat non-unique SNAPSHOT artifacts as mutable in the cache classifier

### LINKED ISSUE #3295: Non-unique Maven SNAPSHOT redeploy is rejected as immutable
Found while investigating #3182. Conclusion is from reading the code path, not a live reproduction — **worth confirming against a running instance before acting on it**, but the path looks unambiguous.

## The path

`classify_maven` (`backend/src/services/cache_classifier.rs:291-308`) for a **non-unique** snapshot artifact — e.g. `com/example/app/1.0-SNAPSHOT/app-1.0-SNAPSHOT.jar`, the shape produced when `uniqueVersion=false`:

1. leaf is not `maven-metadata.xml*` → falls through;
2. `lower.contains("-snapshot/") && !has_artifact_extension(leaf)` → **does not fire**, because `.jar` *is* an artifact extension;
3. `has_artifact_extension(leaf)` → returns `Mutability::Immutable`.

The comment above that branch says concrete **timestamped** SNAPSHOT artifacts are immutable, which is correct — `app-1.0-20240101.120000-3.jar` never gets rewritten, a new filename is minted per deploy. The rule does not distinguish the non-unique spelling, where the *same* filename is overwritten on every deploy by design.

That classification then feeds the release-immutability backstop (`backend/src/services/artifact_service.rs:681-730`):

```rust
let is_released = prior.version.is_some()
    || cache_classifier::classify(&repo.format, path).is_immutable();
if is_released && !prior.checksum_sha256.eq_ignore_ascii_case(checksum_sha256) {
    return Err(AppError::Conflict("Artifact version already exists and is immutable"));
}
```

Both disjuncts are satisfied — `version` is `Some("1.0-SNAPSHOT")` *and* the classifier says immutable — so a second `mvn deploy` of the same snapshot with changed bytes is refused.

## Impact

Redeploying a snapshot is the defining behaviour of a snapshot. Affected users see `409 Conflict` with a message stating the version is immutable, which reads like intended policy rather than a defect.

`uniqueVersion=false` was deprecated in Maven 3, so the default `mvn deploy` path (timestamped filenames) is unaffected — each deploy writes a new coordinate. But the non-unique spelling is still honoured by Maven, is common in older internal builds, and `RepositoryFormat::Gradle` and `RepositoryFormat::Sbt` share `classify_maven` (`cache_classifier.rs:175`), so any of their publishing configurations that reuse a fixed snapshot filename hit the same wall.

Identical-bytes republish is unaffected — the guard only fires when the checksum differs — so a no-op redeploy succeeds and the failure only appears once the artifact actually changes, which is the confusing case.

## Suggested direction

Make the SNAPSHOT test independent of the extension check. A path under a `-SNAPSHOT/` directory is a snapshot coordinate whether or not the leaf carries a timestamp; only the *timestamped* form should be immutable:

- `.../1.0-SNAPSHOT/app-1.0-20240101.120000-3.jar` → immutable (unique per deploy)
- `.../1.0-SNAPSHOT/app-1.0-SNAPSHOT.jar` → mutable (overwritten by design)

Distinguishing them means detecting the `yyyyMMdd.HHmmss-N` timestamp in the leaf rather than relying on "has an extension".

## Verification note

A test for this must assert the **redeploy with different bytes**, not merely that the classifier returns a value — the current behaviour is a `Conflict` at upload time, several layers below the classifier. It also needs a positive control in the same fixture: a timestamped snapshot artifact must stay immutable, and a genuine release redeploy must still be refused. Asserting only "snapshot redeploy succeeds" would pass if the backstop were removed entirely.

## Related

- #3182 — where this surfaced; the reporter's actual problem was a separate storage-key collision fixed in 1.7.0
- The backstop itself is correct and valuable for releases; this is narrowly about which Maven paths it treats as released

brandonrc:
## Summary

Fixes #3295.

Redeploying a non-unique Maven SNAPSHOT artifact (e.g. `app-1.0-SNAPSHOT.jar`) over a previously deleted version was rejected with `409 "Artifact version already exists and is immutable"`. Per Maven SNAPSHOT semantics, a non-unique SNAPSHOT keeps the literal `-SNAPSHOT` token in its filename and is redeployed in place — it is a mutable coordinate, not a released one.

Root cause: `classify_maven` in `backend/src/services/cache_classifier.rs` classified **any** file with a Maven artifact extension as `Immutable`; its SNAPSHOT-directory mutable branch only covered non-artifact files (metadata/listings). A non-unique SNAPSHOT jar/pom therefore fell through to `Immutable`, and the shared release-immutability guard (`cleanup_soft_deleted_artifact_checked`) then 409'd a redeploy over a soft-deleted tombstone.

Fix (single condition in the shared classifier): a file inside a `-SNAPSHOT/` version directory whose filename still carries the `-snapshot` token is classified mutable. Unique (timestamped) SNAPSHOT artifacts (`app-1.0-20240101.120000-3.jar` — the token is replaced by the timestamp) and release coordinates are unchanged and stay `Immutable`, so release immutability is preserved. Because every consumer (upload guard, service-backed mirror, proxy cache, overwrite-permission check) reads through `classify()` / `is_explicitly_mutable_index()`, the correction flows through at one point.

Deliberate, benign consequence: the proxy cache now revalidates remote non-unique SNAPSHOT artifacts on TTL instead of caching them forever, which matches Maven's volatile-SNAPSHOT semantics and removes a latent stale-SNAPSHOT cache issue.

No migration; no sqlx prepared-query changes (the changed function is pure; new DB-backed tests use runtime `sqlx::query`).

Red on `main`, green here:
- `services::cache_classifier::tests::classify_matches_table` — new rows: non-unique SNAPSHOT `.jar`/`.pom`/`-sources.jar`/`.jar.sha1` under a `-SNAPSHOT/` directory classify mutable; existing timestamped-SNAPSHOT and release rows kept immutable (regression guards).
- `services::cache_classifier::tests::explicitly_mutable_index_only_for_real_index_files` — new assertion: `is_explicitly_mutable_index(Maven, ".../1.0-SNAPSHOT/app-1.0-SNAPSHOT.jar") == true`.
- `api::handlers::tests::checked_cleanup_allows_nonunique_snapshot_redeploy_different_bytes` (DB-backed) — tombstone + different incoming bytes at a non-unique SNAPSHOT path now purges and proceeds; was `Err(Conflict)` on `main`. The existing `checked_cleanup_blocks_immutable_swap_different_bytes` (release path) stays green — release redeploys with different bytes still 409.

## Test Checklist
- [x] Manually tested locally

Full `cargo nextest run --workspace --lib` with a real database (`AK_TESTS_REQUIRE_DB=1`): 15172 passed. Manually verified on an isolated deployment: SNAPSHOT deploy → version delete → redeploy with new bytes now succeeds and serves the new bytes (was 409); the same flow on a release coordinate still returns 409 immutable.

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/mod.rs
+++ b/backend/src/api/handlers/mod.rs
@@ -729,6 +729,49 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn checked_cleanup_allows_nonunique_snapshot_redeploy_different_bytes() {
+        let Some(pool) = crate::api::handlers::test_db_helpers::try_pool().await else {
+            return;
+        };
+        let repo = make_repo(&pool, "maven").await;
+        // Non-unique SNAPSHOT artifact: redeployed in place per Maven semantics.
+        let path = "com/x/app/1.0.0-SNAPSHOT/app-1.0.0-SNAPSHOT.jar";
+        insert_tombstone(
+            &pool,
+            repo,
+            path,
+            "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
+        )
+        .await;
+
+        // Redeploy with DIFFERENT bytes -> allowed (SNAPSHOTs are mutable).
+        let res = cleanup_soft_deleted_artifact_checked(
+            &pool,
+            &RepositoryFormat::Maven,
+            repo,
+            path,
+            "bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb",
+        )
+        .await;
+        assert!(
+            res.is_ok(),
+            "non-unique SNAPSHOT redeploy over a tombstone must be allowed, got {res:?}",
+        );
+        let remaining: i64 =
+            sqlx::query_scalar("SELECT COUNT(*) FROM artifacts WHERE repository_id = $1")
+                .bind(repo)
+                .fetch_one(&pool)
+                .await
+                .unwrap();
+        assert_eq!(
+            remaining, 0,
+            "SNAPSHOT tombstone purged so redeploy can land"
+        );
+
+        cleanup_repo(&pool, repo).await;
+    }
+
+    #[tokio::test]
     async fn checked_cleanup_no_tombstone_proceeds() {
         let Some(pool) = crate::api::handlers::test_db_helpers::try_pool().await else {
             return;
--- a/backend/src/services/cache_classifier.rs
+++ b/backend/src/services/cache_classifier.rs
@@ -294,10 +294,16 @@ fn classify_maven(lower: &str) -> Mutability {
     if leaf.starts_with("maven-metadata.xml") {
         return Mutability::mutable_default();
     }
-    // A SNAPSHOT path that is NOT a concrete timestamped artifact is mutable
-    // (the directory listing / metadata is republished). Concrete timestamped
-    // SNAPSHOT artifacts (e.g. `app-1.0-20240101.120000-3.jar`) are immutable.
-    if lower.contains("-snapshot/") && !has_artifact_extension(leaf) {
+    // Files inside a `-SNAPSHOT/` version directory. A NON-UNIQUE SNAPSHOT
+    // artifact keeps the literal `-SNAPSHOT` token in its filename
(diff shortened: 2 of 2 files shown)


cstamas:
I find this overkill: since Maven 3.0 the "non unique" snapshots are _not supported_. Do you plan to support Maven 2? Otherwise, I'd advise to not do this. (or, if you have some broken client that does this? Maven 3.0 -- from 2010 -- does not does this for sure).

PR #3367, opened 2026-08-14
fix(pypi): map a missing metadata source object to 404 instead of 500 on the PEP 658 route

### LINKED ISSUE #3366: PyPI: PEP 658 metadata via a virtual repo returns permanent 500 instead of falling back to the remote member's upstream (regression in 1.7.4)
## Summary

On 1.7.4, a PEP 658 `.whl.metadata` request through a **virtual** PyPI repository returns a permanent `500 {"code":"STORAGE_ERROR"}` when the remote member has an `artifacts` table row for the wheel but the storage read for that row fails. The same request served directly against the remote member returns 200 (metadata fetched from upstream), so the upstream fallback works — the virtual resolution path just never reaches it.

This is a regression relative to 1.7.0, where the `.metadata` route on a virtual repo simply 404'd (virtual repos own no artifacts) and pip fell back to downloading the wheel. It was introduced by the PEP 658 virtual-repo handling rework (#3179 and follow-ups).

## Mechanism (v1.7.4)

Three things compound in `backend/src/api/handlers/pypi.rs`:

1. `serve_virtual_metadata` resolves members **local-first, including Remote members**: it calls `serve_metadata(member.id, ...)` before the member's `serve_remote_metadata` branch. (The local-first ordering itself is well-motivated — the comment explains the bytes-mismatch it prevents — this issue is about its failure handling.)

2. `serve_metadata` maps **every** storage error, including `AppError::NotFound`, to a 500 via `map_storage_err`:

   ```rust
   let content = storage
       .get(&artifact.storage_key)
       .await
       .map_err(map_storage_err)?;   // NotFound -> AppError::Storage(...) -> 500
   ```

   A missing/unreadable object behind an artifacts row is indistinguishable from "this member does not have it", but it surfaces as `Storage error: Resource not found: Storage key not found: proxy-cache/<repo>/simple/<pkg>/<wheel>/__content__` with HTTP 500.

3. Back in `serve_virtual_metadata`'s member loop, a non-404 from the local-first check records `first_member_error` and `continue`s **to the next member** — skipping the *same* member's `serve_remote_metadata` branch. So the one code path that could serve the metadata (fetching it from the member's upstream) is never tried, and the request fails identically on every retry. No self-healing, unlike the download path (#3147, which re-fetches upstream when the cached object is missing).

## Why this isn't just "swallow the error"

The non-404 propagation in the member loop is a **deliberate security decision** from #3179, and this issue is not asking to revert it. The in-code comment is explicit:

> A non-404 here is a storage or DB fault, or a 503 from the decode-permit fast-fail — not "this member does not have it". Swallowing it would fall through to a lower-priority REMOTE member and serve the upstream's metadata for a distribution this member holds locally, which is exactly the bytes-mismatch the local-first ordering exists to prevent, reappearing under load.

That rationale is sound for *transient* faults: if a member really holds the wheel locally and storage hiccups, falling through to a public upstream's metadata would let an installer resolve against different bytes than the wheel it then downloads.

But a storage `NotFound` on `artifact.storage_key` is the one error class where that threat model does not apply: **there are no local bytes to mismatch against**. The row promised bytes that storage cannot produce — durably, not transiently — so "this member does not have it" is precisely the right reading. The regression is that `NotFound` gets lumped in with transient faults by `map_storage_err`, which appears to be a long-standing blanket mapping (its only documented nuance is ENAMETOOLONG → 400) that predates the virtual-metadata rework and was never revisited when #3175/#3179 pointed `serve_metadata` at remote members carrying `artifacts` rows. The #3175 changelog entry likewise assumes row ⇒ bytes ("virtual repositories still queried metadata against the virtual itself, which owns no artifacts, so every request 404'd") and never considers the row-without-object state.

Net effect: one artifacts row pointing at an unreadable object turns every `pip install` that requests that wheel's PEP 658 metadata through the virtual repo into a hard 500 — permanently, until the row is manually deleted. pip treats the metadata 500 as fatal, so installs fail even though the wheel itself downloads fine.

## Reproduction

1. Create a remote PyPI repository `pypi-remote` proxying `https://pypi.org`, and a virtual repository `pypi-virtual` with `pypi-remote` as its only member.
2. Arrange for `pypi-remote` to have an `artifacts` row for a wheel whose `storage_key` object is not readable (e.g. delete the object from the backing store after it is indexed, or point the row at a nonexistent key).
3. `GET /pypi/pypi-virtual/simple/<project>/<wheel>.whl.metadata` → `500 {"code":"STORAGE_ERROR","message":"Storage operation failed"}` on every attempt.
4. `GET /pypi/pypi-remote/simple/<project>/<wheel>.whl.metadata` (member directly) → `200`, metadata served from upstream.
5. `GET /pypi/pypi-virtual/simple/<project>/<wheel>` (the wheel itself) → succeeds (download path self-heals per #3147).

Expected: step 3 should serve the metadata from the member's upstream (matching step 4), or at worst return 404 so installers fall back to downloading the wheel — never a permanent 500.

## Suggested fix

Carve `NotFound` out of the propagated error class; keep everything else propagating exactly as #3179 intended:

- In `serve_metadata`, map a storage `NotFound` on `artifact.storage_key` to 404 rather than routing it through the blanket 500 mapping. The virtual loop's existing `Err(404) => {}` arm then falls through naturally, and the #3179 protection against transient storage/DB faults is preserved untouched. This also mirrors #3147's stance on the download path: a missing cached object is treated as absent, not as a server fault.
- Optionally, harden the loop as well: when the local-first check fails with a non-404 on a **Remote** member, that member's own `serve_remote_metadata` branch could still be attempted before recording the error — falling back to *the same member's* upstream is not the cross-member bytes-mismatch the comment guards against. This is defence-in-depth for whatever error class comes next; the `NotFound` carve-out alone fixes the reported regression.

## Additional observation

In the deployment where we hit this, the storage read failed with `NotFound` even though the object **does** exist at `artifact.storage_key` in the backing object store (the server's own presigned redirects for the same key serve it fine). That suggests a possible second bug — the handle obtained via `storage_for_repo(&repo.storage_location())` resolving somewhere other than where the proxy-cache store wrote the object. We have not fully isolated that part, so this issue focuses on the error-handling regression, which turns any such read failure into a permanent, retry-proof 500. Happy to file the storage-resolution part separately if useful.

Observed on: 1.7.4 (S3 storage backend). Not reproducible on 1.7.0 (route returned 404 for virtual repos).

ysolomon-plat:
Fixes #3366.

## What

`serve_metadata` routed every storage error through the blanket `map_storage_err` 500 mapping, `NotFound` included. Since #3175/#3179 point it at remote members carrying `artifacts` rows, a row whose `storage_key` object no longer exists (deleted cache entry, lost object, stale backfill) turned the virtual `.whl.metadata` route into a **permanent 500**: the member loop deliberately propagates non-404s (the #3179 transient-fault guard), so the same member's `serve_remote_metadata` fallback was never tried, and no retry could heal it. `uv` treats a 500 metadata response as fatal, so installs of the affected distribution fail outright.

This PR carves `NotFound` out of the propagated class and maps it to 404. The virtual loop's existing `Err(404) => {}` arm then falls through to the member's upstream fetch. Every other error keeps propagating exactly as #3179 intended — a storage `NotFound` on `artifact.storage_key` is the one class where the cross-member bytes-mismatch threat model cannot apply, because there are no local bytes to mismatch against. Same stance #3147 takes on the download path for a missing cached object.

## Test

The regression test seeds an orphaned `artifacts` row on a remote member (object never written) and asserts the virtual metadata route serves the upstream's PEP 658 metadata. Mutation-checked: with the fix hunk reverted and the test kept, it fails with exactly the production symptom (`500 {"code":"STORAGE_ERROR"}`); with the fix it passes, and all pre-existing `test_virtual_pypi_metadata_*` tests stay green (run serially — the suite has an unrelated parallelism flake around the shared SSRF allowlist).

### DIFF
--- a/backend/src/api/handlers/pypi.rs
+++ b/backend/src/api/handlers/pypi.rs
@@ -3831,7 +3831,24 @@ async fn serve_metadata(
     let content = storage
         .get(&artifact.storage_key)
         .await
-        .map_err(map_storage_err)?;
+        .map_err(|e| match e {
+            // A NotFound here means the artifacts row promised bytes that
+            // storage cannot produce — durably, not transiently. For the
+            // caller that is "this repository does not have it", the same
+            // stance #3147 takes on the download path for a missing cached
+            // object. Mapping it through the blanket storage-error 500 made
+            // the virtual member loop's non-404 propagation (a deliberate
+            // #3179 dependency-confusion guard against TRANSIENT faults)
+            // pin every retry to a permanent 500 and suppress the same
+            // member's own upstream fallback (#3366). There are no local
+            // bytes for a fallback to mismatch against, so 404 is the one
+            // honest answer.
+            AppError::NotFound(_) => {
+                AppError::NotFound(format!("Metadata source not available: {}", filename))
+                    .into_response()
+            }
+            other => map_storage_err(other),
+        })?;
 
     // #2561: permit-scoped decode on the serve path, fast-fail 503 on
     // saturation. Only taken for the branches that actually decode an archive.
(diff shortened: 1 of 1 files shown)


ysolomon-plat:
Local test results backing the PR, for reviewer convenience.

**Environment:** `postgres:16-alpine` (same image CI uses), all 181 migrations applied, `SQLX_OFFLINE=true cargo test --lib` against `DATABASE_URL` so the DB-backed fixtures run for real rather than no-op'ing.

**Mutation check** — with the fix hunk reverted and the regression test kept, the test fails with exactly the reported production symptom:

```
assertion `left == right` failed: an orphaned artifacts row must fall through to the member's
upstream metadata, not pin the route to a 500: {"code":"STORAGE_ERROR","message":"Storage operation failed"}
  left: 500
 right: 200
```

**With the fix** — the full virtual-metadata set is green:

```
cargo test --lib test_virtual_pypi_metadata -- --test-threads=1
test result: ok. 6 passed; 0 failed  (1.74s)
```

One pre-existing observation, unrelated to this change: `test_virtual_pypi_metadata_resolves_remote_member` intermittently fails with a 502 when the suite runs with default parallelism (it passes reliably in isolation and serially). It looks like contention on shared process state around the SSRF-allowlist/wiremock setup between concurrent DB-backed tests — mentioning it so a flaky neighbor in CI isn't attributed to this diff. Happy to file it separately if useful.

ysolomon-plat:
Scope note for reviewers: while validating this we root-caused *why* the storage read misses on our deployment — a key-layout mismatch (`S3_PREFIX`-applying handle reading a root-layout proxy-cache key), now filed as #3368 with full details. This PR is intentionally confined to the error mapping: it is correct for every miss cause (deleted object, lost object, or the #3368 layout mismatch, where the read is structurally guaranteed to miss and upstream fallback is the only possible serving path). The mis-routed read itself is left to #3368 rather than expanded into this diff.

Issue #1437, opened 2026-05-27
[bug,hardening] Backend returns 5xx (not rate-limit) under modest concurrent load above ~10 simultaneous requests


brandonrc:
## Summary

Backend returns 5xx errors under modest concurrent load instead of either succeeding or returning 503 / rate-limit responses. The threshold is around 10 simultaneous requests.

## Reproduction

Test cluster: backend `dev` (commit `sha-54f3848`), release-gate run [26487615730](https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/26487615730).

Stress test output:

| Concurrency | Successful | 5xx | Timeouts | Max response |
|---|---|---|---|---|
| 20 | 8 | 12 | 0 | 1351 ms |
| 40 | 8 | 32 | 0 | 1417 ms |

Suite: `stress-tests`, [job 77999424637](https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/26487615730/job/77999424637).

The fact that successful response count stays flat at 8 across both concurrency levels suggests the backend has a hard internal limit (semaphore, connection pool, or executor worker count) that returns 5xx for requests over the limit instead of queueing.

## Expected

Under concurrent load above capacity, the backend should either:
1. Queue requests up to a bound and serve them in order
2. Return HTTP 503 with `Retry-After` if overloaded
3. Apply rate limiting and return 429

5xx for healthy in-bound requests is a stability bug.

## Severity

High. Real customer traffic regularly bursts past 10 concurrent requests for any active package format. Conan stress in the same gate run also shows ~50-60% failure on concurrent ops (separate issue, but likely same root cause).

## Affected modules

- Connection pool config (sqlx, http server)
- Tower/Axum request budget middleware if any
- Any per-handler semaphores

PR #3382, opened 2026-08-14
feat(maven): Support prefix files

### LINKED ISSUE #3383: (Maven) Add support for RRF/prefix files
### Problem Statement

Maven 3.9+ supports "remote repository filtering" (RRF), and one filter relies on `.meta/prefixes.txt` file, the list of prefixes advertised by repositories, see here https://maven.apache.org/resolver/remote-repository-filtering.html#the-prefixes-filter

A demo project https://github.com/cstamas/rrf-demo

Content should come as:
* local/staging - generates content for their stored Gs (altered into paths)
* proxy - proxies content from upstream
* virtual - merges the union of the member prefixes

### Proposed Solution

Backend should support this path to provide proper content for Maven. Maven 3.9 needs user config to enable RRF, but in 3.10+ and onwards, the default is flipped and Maven will by default look for prefixes file, unless disabled by user.

### Alternatives Considered

Simply return 404. DO NOT return spurious files, as it may break Maven clients, see https://maven.apache.org/resolver/remote-repository-filtering.html#broken-maven-repository-managers

In linked case, a commercial MRM "leaks" prefixes file via virtual repositories (so exposes prefixes file as of virtual one, while content is from one -- random -- member), effectively wrecking Maven, so users of it must disable prefix download.

### Use Case

Performance increase for Maven, by avoiding "known to result in 404" HTTP roundtrips.

### Component

Backend / API

### Additional Context

_No response_

### Pre-submission Checklist

cstamas:
## Summary

Maven format: Added `.meta/prefixes.txt` support for RRF/prefixes, see https://maven.apache.org/resolver/remote-repository-filtering.html#the-prefixes-filter

Closes #3383

Behaviour:
* local/staging - generates content for their stored Gs
* proxy - proxies content from upstream
* virtual - merges the union of the member prefixes

## Test Checklist
- [ ] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc:
Thanks for this — RRF prefix support is a genuinely valuable addition, and the shape of the change is right. I ran three review passes (design/integration, security, correctness/tests). Let me lead with what's solidly done, because it's most of the PR:

- **The interception point is well chosen and carries almost no regression risk.** `is_prefixes_file` is exact-match after trimming leading slashes (`formats/maven.rs:218-220`), matching a 2-segment path, while `parse_coordinates` requires ≥4 segments — so no legal Maven coordinate can ever be shadowed, and `PUT /maven/{repo}/.meta/prefixes.txt` is already rejected `400` by the upload handler. The `-1` in `+401/-1` is a **comment** (`// 2.` → `// 2a.`), verified by diffing the blobs directly. Nothing was removed or reordered. Exactly five paths change behavior: the file and its four checksum siblings, all of which previously 404'd.
- **Authorization is correct**, and I checked this specifically since the file is a complete groupId inventory. No new route is registered; the branches live inside `download`, behind `repo_visibility_middleware` on the whole `format_routes` nest. The virtual arm additionally re-authorizes members via the #1804/#3325 helper with the identical call the neighbouring metadata path makes. No leak.
- **Curation is correctly unaffected** — `maven_proxy_package_name(".meta/prefixes.txt")` returns `None` via the same `parse_coordinates` failure, so `enforce_curation` is skipped exactly as it already is for metadata.
- `cargo check --all-targets` on the head is clean, 0 warnings. `staging` correctly lands in the hosted arm. Output is sorted and deduped, so the body and its ETag are deterministic. No `unwrap()`, panicking index, blocking I/O, or missing `.await` anywhere in the new async code.

The findings are semantic, and they cluster on one theme: **a prefixes file is an exhaustive allowlist, and this code emits one in several cases where it doesn't actually know the full set.**

---

## 1. 🔴 The magic header is `1.0`; Resolver requires `2.0` and discards anything else

`formats/maven.rs:229` emits `## repository-prefixes/1.0`.

I verified this against upstream rather than inferring it. `PrefixesSource.Parser` accepts exactly three first lines:

```java
PREFIX_MAGIC        = "## repository-prefixes/2.0"
PREFIX_LEGACY_MAGIC = "# Prefix file generated by Sonatype Nexus"
PREFIX_UNSUPPORTED  = "@ unsupported"
```

and anything else returns `invalid(origin, path, "No expected magic in file")`, halting immediately without reading entries. The page your own doc comment links (`formats/maven.rs:223`) says the same: *"Valid prefix files start with following 'magic' on their very first line: `## repository-prefixes/2.0`. If the first line in file is not this string, the prefix file is discarded."*

**Consequence:** every file this endpoint serves is discarded. The filter falls to `NO_INPUT_PREFIXES`, and since `CONFIG_PROP_NO_INPUT_OUTCOME` defaults to `true`, Resolver abstains and allows everything. The feature is a silent no-op, with a client-side `WARN` as the only signal — which is also why it wouldn't have shown up in manual testing.

It compounds in the virtual arm: `parse_prefixes_lines` (`maven.rs:1691-1697`) strips the upstream's real `2.0` header and `generate_prefixes_txt` re-emits `1.0`, so **a proxy of Central whose prefixes worked standalone stops working once merged through a group.**

The literal changes in six places: `formats/maven.rs:229`, plus assertions at `formats/maven.rs:851`, `:860` and `maven.rs:4902`, `:4988`, `:5069` — those assertions are precisely why CI is green.

**Sequencing note:** because the header is invalid, findings 2–4 are currently latent. Fixing this line alone converts all three from inert to breaking builds in production, so they should land together.

## 2. 🔴 The virtual union publishes a partial allowlist rather than declining to publish

`maven.rs:1655-1670` — both member arms end in `.unwrap_or_default()`, so a member that *fails* contributes `[]`, indistinguishable from one that genuinely contains nothing. The result still returns an authoritative `200`. `fetch_remote_member_metadata` has already swallowed its own error via `.ok()?` (`:1291`) by that point, and it additionally requires valid UTF-8 (`:1292`), so a non-UTF-8 body degrades the same way.

This matters more here than in the `maven-metadata.xml` merge it's modeled on. There, a short `<versions>` list means a version is missing from a *listing* but a client that pins it still resolves fine. Here the semantics invert: an omitted prefix means the request is **never sent**.

**Concrete failure:** virtual `maven-all` = [local `releases`, remote `central`, remote `google` (`dl.google.com/dl/android/maven2/`, which publishes no prefixes file)]. The `google` fetch 404s → contributes `[]` → `maven-all/.meta/prefixes.txt` returns `200` with `/com/android` absent. Clients stop requesting `com.android.tools.build:gradle` from `maven-all` and the build fails with "Could not resolve dependencies" — while a direct GET of that jar still succeeds, because the artifact walk has no such filter. The symptom points away from the cause. A 30-second timeout to Central during one request has the same effect, durably, via Resolver's client-side cache.

Resolver 2.0.11+ softens this (`CONFIG_PROP_VERIFY_DENIED` defaults `true`, so a proven-wrong prefix tree is dropped for the session), but that doesn't exist in 1.9.x clients and defeats the performance purpose anyway. It's a safety net, not a reason to ship the partial union.

**Suggested fix:** the union is only publishable when *every* member's set is known. If any member's prefixes are unavailable for any reason, return `404` — which is what the sibling metadata arm already does when it can't assemble a complete answer (`maven.rs:1551`), and which RRF reads as "don't filter, ask the repo." The format also has a first-class escape hatch for exactly this: `@ unsupported` as the first line makes the parser return `invalid(...)` and Resolver abstains for that repository. Either is better than a knowingly-incomplete file. Note this needs a prefixes-specific fetch that keeps the upstream status, so a genuine 404 (member publishes nothing → contributes nothing, fine) is distinguishable from a 502/timeout (unknown → bail).

## 3. 🔴 Aggregating caller uses `authorize_virtual_members`; the contract says `try_authorize_virtual_members`

`maven.rs:1650-1651`. The helper's own doc comment is explicit:

> *"Flattening Err into the empty set is correct for the WALKING callers ... Callers that AGGREGATE the filtered set into a response (OCI tags/list, #3320) must use `try_authorize_virtual_members` instead: for them an empty set is indistinguishable from 'nothing exists'..."*

This is an aggregating caller — same bug class as #3320.

**Failure:** `filter_visible_repo_ids` hits a saturated pool during a CI burst → `Err` → flattened to `vec![]` → `200` with a header-only body. A retryable `503` becomes a definitive "this group contains nothing," which clients then cache. One DB blip becomes a fleet-wide resolution outage that outlives the blip.

## 4. 🔴 An unknown or unreadable `repo_type` falls through to the hosted generator and answers "empty"

`maven.rs:1630-1686` is `if Remote {…} / if Virtual {…} /` unconditional Local-Staging fallthrough. `RepoInfo.repo_type` is a raw `String`, and `RepositoryType::from_db_str` exists precisely for this — documented at `models/repository.rs:270-275`: *"Returns `None` for anything that is not a known repository type so a caller reading a raw `repo_type` string can fail closed ... `resolve_repo_by_key` yields an empty string when the column read fails, and a variant added later is unknown to every existing match."*

**Failure:** the column read fails and yields `""`; the request lands in the hosted arm; `collect_local_group_prefixes` finds no rows for what may be a remote repo; the client gets an authoritative empty allowlist for a repository that proxies all of Central. Same outcome the day a `federated` type is added.

**Fix:** `match RepositoryType::from_db_str(&repo.repo_type)` with an explicit `None => 404`.

## 5. 🟠 Generated lines aren't validated against the parser's limits — one bad groupId invalidates the whole file

`maven.rs:1722` does `format!("/{}", g.replace('.', "/"))` on the raw `metadata->>'groupId'`, which originates in a client-supplied POM, and `generate_prefixes_txt` sorts, dedupes and emits. Neither applies any of the parser's constraints (verified upstream):

| Parser rule | What breaks it |
|---|---|
| `MAX_ENTRIES = 100_000` → "Contains too many entries" | one line per distinct groupId, no depth collapsing — Central's own file uses depth 2–4 for this reason |
| `MAX_LINE_LENGTH = 250` → "Contains too long line" | `artifacts.path` is `VARCHAR(2048)` |
| ASCII-only → "Contains non-ASCII characters" | nothing rejects a non-ASCII groupId at deploy |
| no `:` `<` `>` `\` `//` → "Contains forbidden characters" | `com..example` → `/com//example`; an empty groupId (passes the `IS NOT NULL` filter) → `/` |

Every one is **whole-file** rejection, not per-line. So a single malformed artifact anywhere in a 100k-artifact repo silently disables filtering for that repo — and via finding 2, for every virtual repo including it.

Relatedly, the path is percent-decoded by axum's `Path` extractor, so a newline is reachable: `PUT /maven/{repo}/com%0A%2F/lib/1.0/lib-1.0.jar` yields groupId `com\n/`, injecting a bare `/` line that matches every groupId. Requires write permission (deny-by-default), and because the merge is additive an attacker can only *widen* the filter, never remove a legitimate prefix — so the dangerous direction isn't available. Low severity on its own, but it's the same missing validation.

Worth noting the asymmetry: the remote-member branch *does* sanitize (`parse_prefixes_lines` drops lines not starting with `/`); the local branch doesn't.

**Fix:** apply the length/charset/forbidden-character checks next to the existing `sort`/`dedup` in `generate_prefixes_txt` so one place owns "this output is parseable", and when the entry count would exceed 100k either collapse to a shallower depth (truncate each path to N segments before dedupe — what Central does, and cheap given you already sort) or emit `@ unsupported`.

## 6. 🟠 Uncached full-repo scan per request, fanned out across virtual members

`maven.rs:1701-1724` runs `SELECT DISTINCT am.metadata->>'groupId' … WHERE a.repository_id = $1` with no LIMIT and no cache. No index serves it: `artifact_metadata` has no `repository_id` column, and `idx_artifact_metadata_maven_gav` (migration 140) is an equality index on the *values*. The plan is an index scan on `idx_artifacts_repo_path`, then a per-row `artifact_metadata` probe, jsonb detoast and parse, then a hash aggregate.

The rationale at `:1620-1623` ("fetched rarely, once per repo by a client/group") holds for a developer with a warm `~/.m2`, but Resolver caches the file in the *local repository* — so it's one fetch per ephemeral CI agent. The contrast inside this very file is the strongest argument: `maven-metadata.xml` generation is cached (`MAVEN_METADATA_CACHE`, 60s) **and** the virtual merge is separately cached (`VIRTUAL_MAVEN_METADATA_CACHE`, 60s). Prefixes is the only Maven fan-out with neither, while costing orders of magnitude more per call.

Amplifiers: five URLs per repo each re-run it (base + four checksums); the ETag doesn't help, since `cacheable_response` computes it *from the already-generated body*, so even a 304 pays full cost; and a virtual repo fans out to every member in batches of 16. Format routes also carry no rate limit — `rate_limit_middleware` is applied only inside `api_v1_routes`.

**Failure:** a 1M-artifact repo with 200 concurrent CI jobs → 200 × (1M-row join + aggregate) to produce ~400 lines. There's also a sharper variant: operators commonly set a server-side `statement_timeout` (our own code comments cite 10s on RDS as typical, and `create_pool` sets none), in which case the query is *killed* and finding 8 turns every prefixes GET on a large repo into a 500 carrying raw Postgres text.

**Two fixes, ideally both.** (a) Source from the `packages` catalog instead: it already holds one row per `(repository_id, "groupId:artifactId")` for maven/gradle (`UNIQUE(repository_id, name)`, migration 113; Maven names normalized by 177), served directly by `idx_packages_repository_id`. `maven_component_keys_sql` (`repositories.rs:6347`) with `MAVEN_CATALOG_NAME_SHAPE_SQL` is almost exactly this query, already written and index-served — roughly a 50× row reduction and no jsonb work. (b) Add the Moka cache keyed on `repo_id` with the same 60s TTL, and register invalidation in `invalidate_maven_metadata_cache` — that one hook already covers both write paths (deploy at `maven.rs:2811`, delete via `clear_stored_maven_metadata`). That also fixes finding 9 for free.

## 7. 🟡 The branch is behind main and won't compile against it

Three call sites moved:
- `maven.rs:1634-1641` calls `proxy_fetch_capped_budgeted` with 6 args; main's takes a 7th, `format: RepositoryFormat` (#3459), driving `cache_classifier` TTL selection. Pass `RepositoryFormat::Maven`.
- `maven.rs:1649-1651` uses the `fetch_virtual_members` + `authorize_virtual_members` pair; main's `authorized_virtual_members` composes them, and its doc block says that's the form every content-serving path should use. (Ties into finding 3 — `try_authorize_virtual_members` is the aggregating variant.)
- Once the cache from finding 6 lands, the virtual arm should also consult `virtual_aggregate_cacheable` (#3323), which stops one caller's authorization-dependent merge being served to the next. A full groupId inventory makes that matter more here than for a single GA's version list.

## 8. 🟡 Raw sqlx error text returned to the client

`maven.rs:1677-1685` builds `(INTERNAL_SERVER_ERROR, format!("Database error: {}", e))`, bypassing `AppError`. `map_db_err` is already imported at `maven.rs:31`. `AppError::Database` deliberately redacts to `"Database operation failed"` (`error.rs:265`) with a regression test asserting raw SQL never reaches the client (`error.rs:338`), and `fetch_virtual_members` carries the comment *"to avoid leaking raw DB error text (#1437)"*. It also returns 500 on pool timeout instead of the 503/`POOL_EXHAUSTED` mapping, so this path pages during ordinary capacity shed.

In fairness `generate_metadata_for_artifact` has the same wart, so this is "don't add a second instance." Having `collect_local_group_prefixes` return `sqlx::Error` rather than `String` lets both call sites use `map_db_err`.

## 9. 🟡 Body and `.sha1` are generated independently and can disagree

`maven.rs:1024-1031` handles the sidecar by calling `fetch_maven_prefixes_bytes` a second time and hashing the result. With no cache between the two requests, a deploy landing in between yields a checksum that doesn't match the body the client just fetched. This is the invariant #1922 established for `maven-metadata.xml`, which gets it free from the 60s cache. Fixing finding 6 fixes this.

Also: Resolver never requests this sidecar (`PrefixesRemoteRepositoryFilterSource` fetches only `PREFIX_FILE_TYPE`), so the branch serves generic clients only while doubling finding 6's cost. Dropping it is also defensible.

## 10. 💭 Smaller items

- **`parse_prefixes_lines` turns "I can't answer" into "I have nothing."** It keeps only lines starting with `/`, so an upstream publishing `@ unsupported` — the standard way a repo says it can't produce prefixes — parses to zero entries and contributes nothing while the merge claims completeness. Same class as finding 2; fix together.
- **Empty local repo returns a valid empty allowlist** (`200`, header-only, pinned by `test_empty_repo_prefixes_txt_is_header_only`). Spec-wise arguably true; operationally a trap — a new repo hands out an empty allowlist, CI caches it, the first `mvn deploy` lands `com.acme:lib:1.0`, and the next build still filters `/com/acme` out while a direct GET works. Worth an explicit decision (I'd suggest 404 when the generated set is empty, consistent with finding 2), with the test asserting the *decided* behavior and the reasoning attached.
- **`Cache-Control: public, max-age=60`** on a private repo's full groupId inventory. Pre-existing and shared with the metadata path, not a regression — but main just landed #3406 (`negotiated_cache_control`, citing RFC 9111 §3.5 on shared caches storing credentialed responses), and an inventory is a higher-value target for that hazard than a version list. One-line adoption.
- **Upstream buffer cap:** `fetch_remote_member_metadata` allows 128 MiB (`LARGE_METADATA_MAX_BYTES`) while holding a budget permit that drops when the function returns — the surviving `String` and the `Vec<String>` from `parse_prefixes_lines` are unaccounted. A 128 MiB body of `/a\n` lines is ~44M `String`s, well over 1 GB resident, ×16 concurrent members. `DEFAULT_METADATA_MAX_BYTES` (8 MiB) is ample for this file; a line-count cap would also do.
- **`.sha1` interception shadows a stored sidecar.** `PUT /maven/{repo}/.meta/prefixes.txt.sha1` is accepted today (the upload checksum branch stores any `*.sha1` with no coordinate parsing). Before this PR a GET returned those bytes; now it returns the generated digest. Narrow, and the generated-wins direction is the safe one, but it means a repo can hold an orphaned sidecar that can never be read back. Worth rejecting the write or documenting the precedence.

## Tests

Six added; two are solid, three are pinned to the wrong constant, one is self-confirming.

`test_is_prefixes_file` is real and useful — mutating to `ends_with` fails it. `test_local_repo_serves_prefixes_txt` is the strongest handler test: it genuinely pins the groupId→path conversion and cross-member ordering, and deleting `.replace('.', "/")` fails it.

`test_prefixes_txt_sha1_matches_body_digest` is weak — it computes `expected` from the body the same handler just produced, both sides running `fetch_maven_prefixes_bytes` → `compute_checksum`, so it can only detect the two arms disagreeing. Green under: header `1.0`/`9.9`/none; deleting `.replace('.', "/")`; dropping the `/` prefix; making `generate_prefixes_txt` return `String::new()`.

`test_virtual_repo_merges_prefixes_from_members` uses two *local* members, so it exercises `collect_local_group_prefixes` twice and never touches the code the virtual arm exists for. Green under: deleting the entire `Remote` branch at `:1656-1660`; making `parse_prefixes_lines` return `vec![]` (that function has **zero** coverage); replacing `.chunks(MAX_VIRTUAL_FANOUT)`. Disjoint member groupIds also mean cross-member dedupe is untested.

All four handler tests early-return when `try_pool()` is `None`, so a green check only means something in a job with Postgres.

One correction to the PR checklist, offered helpfully: **the proxy and virtual arms *can* have unit coverage in this repo.** There's an established wiremock pattern in this exact file on main — `test_resolver_remote_metadata_checksum_matches_body_1922` (`maven.rs:5749`) for a remote repo against a mock upstream, and `test_virtual_metadata_merges_local_and_remote_versions_2069` (`maven.rs:6041`) for a virtual merging a local member with a wiremock-proxied remote, no real egress. Plus `SsrfTestAllowlistGuard` in `test_db_helpers.rs` for the non-loopback mock. Near drop-in templates.

**Minimum missing before merge:**
1. `test_prefixes_txt_header_is_repository_prefixes_2_0` — assert line 1 exactly. This is the test whose absence let finding 1 through.
2. `test_parse_prefixes_lines_drops_header_and_keeps_paths` — pure unit test; given a realistic `2.0` body with comments and blanks, assert exactly `["/com/foo", "/org/bar"]`. Currently zero coverage.
3. `test_virtual_prefixes_merges_local_and_remote_member` — copy `..._2069`; wiremock serves a valid `2.0` body; assert the merge contains both, deduped, sorted, under a `2.0` header.
4. `test_virtual_prefixes_when_a_member_has_no_prefixes_file` — wiremock 404s; assert whatever finding 2 resolves to. **If only one lands, make it this one** — it pins the semantic decision the whole feature rests on.
5. `test_remote_prefixes_proxies_upstream_verbatim` — assert bytes are byte-identical and the header isn't rewritten; upstream 404 → 404.
6. `test_prefixes_unknown_repo_type_is_404` — finding 4.

---

**Path forward:** findings 1–4 are really one coherent fix — emit `2.0`, and make every arm that cannot enumerate a complete set return `404` (or `@ unsupported`) instead of a partial or empty `200`, matching what the metadata arm already does. Finding 5 is validation next to the existing sort/dedup; finding 6 is reuse of the cache already at the top of the file. With tests 1, 3 and 4 this is a solid feature.

And thank you for citing the spec inline in the doc comment — that's what made finding 1 findable at all.

<sub>Reviewed with three automated passes (design/integration, security, correctness/tests). Every Resolver claim above was re-verified against `apache/maven-resolver` upstream and the Apache RRF documentation directly, not taken from the agents.</sub>

cstamas:
Thanks for review! All "red" findings 1-4 are right (my focus, did not verify rest, let's go baby steps), my bad:
https://github.com/apache/maven-resolver/blob/master/maven-resolver-impl/src/main/java/org/eclipse/aether/internal/impl/filter/prefixes/PrefixesSource.java#L83

Will update

cstamas:
@brandonrc pls review (branch rebased but kept commits; commit message have explanations)

cstamas:
Just an explanation regarding (IMHO expected; validate) outcomes:
* hosted repo - always has "definitive" answer, where empty hosted repo simply emits "header only" 200 (don't ask me, I have nothing)
* proxy repo - has "definitive" answer only if upstream delivers it, otherwise 404
* virtual (aggregation) - answer is definitive only, if all member answers are definitive (even one 404 should result in 404 from virtual)

Please review and if needed (for any reason), overrule/change this.

brandonrc:
## Round 2 — close, one blocker

Thanks for the rebase and the three fix commits. Every round-1 finding is resolved and all six requested tests are present, with each one shown to catch its mutation (header rewrite, dropped dedup, partial union on a member 404). The `2.0` header, the 404-on-incomplete rule, `@ unsupported`, validation beside sort/dedup and cache reuse all check out at the cited lines. "A complete-but-empty prefix set is 200" is right per Resolver: a header-only file is valid and denies everything, which is truthful for an empty repository, and the only cost is client-side metadata staleness, the same as for any prefix change. The auth gate matches the metadata route's, a private Virtual cannot leak a private member's path set, and the Remote arm goes through the normal proxy path so only a true upstream 404 reaches the "member has no file" arm. 500/500 DB-backed tests green, `check` and `clippy -D warnings` clean. (The "Nope, this is Maven spec" commit is your own wording fix predating round 1 — noted, no disagreement.)

**Blocker — promoted artifacts are denied by their own repository's prefixes file.** `collect_local_group_prefixes` (`maven.rs` ~2013-2035) derives prefixes from the `packages` catalog, but promotion (`promotion.rs` ~799 and ~1079, `promote_artifact`/`_bulk`) inserts into `artifacts` only, never `packages`. A Staging→Release promotion therefore lands an artifact whose groupId is absent from the Release repo's `prefixes.txt`; Resolver loads the file as valid, `PrefixTree.acceptedPath` returns false, and the just-promoted artifact is unresolvable through that repository while a direct GET works. Narrowing direction, no time bound. This is not a regression from round 1's `packages` suggestion (the original join had the same gap). Two ways out, either is fine: derive from `artifacts.path` (`SELECT DISTINCT regexp_replace(path, '/[^/]+/[^/]+/[^/]+$', '') FROM artifacts WHERE repository_id = $1`), or make promotion write the catalog row. A test that promotes and then asserts the prefix is present would pin it.

**Should-fix (both reviewers)** — a **nested Virtual member yields a partial-union 200.** `fetch_maven_prefixes_bytes_uncached` (~1954-1961) sends every non-Remote member to `collect_local_group_prefixes(member.id)`; a Virtual member owns no catalog rows, so it counts as confirmed-empty and the group publishes an authoritative allowlist missing everything behind it — the spurious-file case #3383 forbids. Nesting is supported (`MAX_VIRTUAL_DEPTH = 32`). Bail with 404 or recurse on `RepositoryType::Virtual`.

**Should-fix** — a **member body without the magic is accepted as empty, and `@ unsupported` is only checked on line 1.** `fetch_remote_member_prefixes` (~1774) and `parse_prefixes_lines` (~1786-1792) will take a soft-404 HTML body as "confirmed empty" and publish a 200 union. Upstream rejects the whole file without the `## repository-prefixes/2.0` magic (`PrefixesSource.java:102-105`) and on `@ unsupported` anywhere (`:107-112`). Same class as round-1 finding 2, reached via a body instead of a status.

**Should-fix** — **transient DB errors are cached for 60 s.** `PrefixesError::Db` is stored in `MAVEN_PREFIXES_CACHE` (~1897-1903), so one pool timeout pins a 503 for that repository's prefixes for a minute; the sibling `MAVEN_METADATA_CACHE.try_get_with` never caches errors.

**Housekeeping**
- The branch now conflicts with `main`: #3711 and this PR both insert functions directly after `fetch_maven_metadata_bytes` in `maven.rs`. Purely positional (keep both blocks), but CI cannot run until it is rebased.
- Please add a CHANGELOG entry under `## [Unreleased]` → `### Added` (new subsection at the top of that section), first reference `#3383`, in the house style: a bold one-sentence lead, then what the file is, that Hosted/Remote/Virtual each serve it (Remote proxied verbatim, Virtual as the union over members and 404 when any member's set is unknown), and that Resolver 1.9+ / Maven 3.9+ consult it to skip repositories that cannot serve an artifact.

Nits, no action needed for merge: `.meta/prefixes.txt` lands in the mutable 5-minute cache class only via the classifier's fallback, and cross-member dedup is exercised at the parser level but not at the handler level.

<sub>Reviewed with two automated passes (Resolver spec/design; tests/correctness/security), each verified against `apache/maven-resolver` upstream directly.</sub>

brandonrc:
I've pushed the round-2 work onto your branch rather than asking you for another cycle — 1.9.0 is at the cut and this deserves to be in it. Your six commits are untouched; mine sit on top as a merge with `main` plus one fix commit, both co-authored to you.

**The blocker (promoted artifacts absent from their own prefixes file).** Fixed by deriving prefixes from `artifacts.path` rather than the `packages` catalog, in preference to making promotion write the catalog row. Three reasons: it is one query instead of a write path; it repairs repositories that were promoted into *before* this ships, which the promotion-side fix cannot; and it honours `is_deleted`, so the file stops advertising groupIds whose artifacts were deleted — nothing deletes the `packages` row today, so the catalog was over-advertising in both directions. The `≥4` path-segment guard is load-bearing: without it a GA-level `maven-metadata.xml` would collapse `/com/acme` to `/com` and hand Resolver a prefix far wider than the repository serves.

**The other three.** A member body must now carry one of upstream's accepted first lines and is rejected on `@ unsupported` anywhere, so a soft-404 HTML body can no longer be published as a confirmed-empty set. A nested Virtual member bails the union with 404 rather than counting as empty — 404 reads as "don't filter" to Resolver, which is the safe direction. And `CachedPrefixes::Db` is gone entirely, so caching a transient DB error is unrepresentable rather than merely unreached.

**Verified:** 782 tests pass on the merged tree, including all six you added; reverting only the SQL hunk fails exactly the new promotion test and nothing else; clippy and fmt clean.

Left alone deliberately: the leading-dot stripping upstream does in `PrefixesSource.java` (cosmetic and widening-only), the classifier arm for `.meta/prefixes.txt`, and handler-level cross-member dedup.

Thank you for this one. The spec citations in your doc comments are what made the first review findable at all, and the design — one helper, three surfaces, 404 when any member's set is unknown — is yours and survived intact.

cstamas:
Thanks!

Issue #3313, opened 2026-08-13
CI: fail on duplicate sqlx migration version numbers (invisible to git, mergeable CLEAN, aborts the backend at startup)


brandonrc:
## Summary

Two migration PRs that pick the same version number produce **no textual merge conflict**, both report `mergeable: CLEAN`, both pass CI, and the backend then **aborts at startup** after the second one merges. There is no automated check for this anywhere in the tree. The only mitigation today is a comment inside one migration file.

## The near-miss on this release

`backend/migrations/195_curation_attestation_verification.sql` carries a header recording that it was renumbered **193 → 195 during review**, because main took 193 and 194 after the branch was cut. The header goes on to spell out the mechanism: sqlx sorts by version without deduplicating, so duplicate versions are not a merge conflict and the backend aborts at startup.

That is the correct diagnosis, written by someone who had just been bitten. But the mitigation it produced is a comment in the file that collided — visible only to someone already editing that file, which is exactly the person who no longer needs the warning.

## Why review does not catch it

Every signal a reviewer normally trusts reads clean:

* `git` reports no conflict — the two files have different names, so there is nothing to conflict on.
* The GitHub UI reports the branch mergeable.
* CI passes on both PRs independently, because each is internally consistent.
* The failure appears only after the *second* merge, in a *third* context (backend startup), with a message that names sqlx rather than either PR.

The blast radius is a hard startup abort, so it fails safe — but it fails safe in production or in the release gate, long after the two changes were each individually approved.

## Requested check

A CI job that fails when `backend/migrations/` contains two files sharing a version prefix. The whole check is roughly:

```sh
dupes=$(ls backend/migrations/ | sed -n 's/^\([0-9]\+\)_.*/\1/p' | sort | uniq -d)
[ -z "$dupes" ] || { echo "::error::duplicate migration version(s): $dupes"; exit 1; }
```

Two design points worth getting right:

**It must run against the merge result, not the PR branch.** A duplicate is by definition invisible on either branch alone — each branch has exactly one file at that number. GitHub's default `pull_request` checkout is already the merge commit, which gives this for free, but it is the property the check depends on, so it deserves an explicit comment in the job rather than being left implicit.

**Consider also asserting contiguity from the previous release tag**, or at minimum warning on a gap. A gap usually means a migration was renumbered on one side and the ledger is drifting — the same underlying condition, caught one step earlier.

[...]

PR #3385, opened 2026-08-14
fix(gc): singleton-lease the scheduled storage GC tick, index maven_flat_object_owner.storage_key

### LINKED ISSUE #3384: Scheduled storage GC has no cross-replica coordination — Maven flat-object orphan scan runs duplicated every hour, stuck 20+ min
## What happened

Observed in production via `pg_stat_activity`: the Maven flat-object orphan-scan query shows up multiple times simultaneously, all with the identical `xact_start`, each running longer than 20 minutes.

```sql
SELECT pid, xact_start, now() - xact_start AS running_for, query
FROM pg_stat_activity
WHERE query ILIKE '%maven_flat_object_owner%' AND state <> 'idle';
```

returns one row per replica, all sharing the same `xact_start` — i.e. every replica fired the same scan at the exact same instant.

## Root cause

`backend/src/services/scheduler_service.rs`, the "Storage garbage collection" block (cron `GC_SCHEDULE`, default hourly `"0 0 * * * *"`), has **no cross-replica coordination at all**. Every replica independently computes the same `next` cron occurrence and fires `service.run_gc(false)` at the same wall-clock instant. Every other periodic job in the same file (`lifecycle_policy_execution`, `health_monitor`, curation sync) acquires a `SchedulerLease` singleton lease (`cluster_work.rs`, added in #3072) before running — this block never got that treatment.

`run_gc()` → `cleanup_orphan_maven_flat_objects()` → `select_orphan_maven_flat_objects()` scans `maven_flat_object_owner` for objects older than 1h with no live artifact/metadata/checksum reference, via several correlated `NOT EXISTS` subqueries per candidate row (one with jsonb containment, one with `regexp_replace`), `LIMIT 1000`.

This is compounded by an indexing gap: `maven_flat_object_owner`'s only index is on `repository_id`; its PK is `(storage_backend, storage_key)` (migration 168), but the query filters `storage_key LIKE 'maven/%'` without constraining `storage_backend`, so the PK's leading column can't be used and the scan falls back to a sequential scan + sort of the whole table before the `LIMIT`. Combined with the per-row `NOT EXISTS` cost, a single pass runs 20+ minutes on our dataset. With every replica firing simultaneously each hour and no lock, that becomes N concurrent 20+ minute copies every hour (N = replica count).

Confirmed via `git log v1.7.2..main -- backend/src/services/scheduler_service.rs backend/src/services/storage_gc_service.rs` (empty) that neither the coordination gap nor the missing index has been touched since v1.7.2.

## Impact

Recurring, roughly-hourly Postgres CPU/lock pressure proportional to replica count, on a query whose per-row cost has grown across releases as the orphan predicate gained more `NOT EXISTS` clauses (#3159, #3192, #3197, #3212, #3225) without a corresponding look at concurrency or indexing.

## Suggested fix

- Wrap only the scheduler's own `run_gc()` call in a `SchedulerLease` (e.g. job name `storage_gc`), following the exact pattern already used for `lifecycle_policy_execution`. Scope it to the scheduled tick only, not the on-demand admin/per-repo GC endpoints, so an explicit human-triggered GC never no-ops while a scheduled tick holds the lease.
- Add `CREATE INDEX IF NOT EXISTS idx_maven_flat_object_owner_storage_key ON maven_flat_object_owner (storage_key);` so the candidate scan can use an index scan with early `LIMIT` termination instead of a full sequential scan + sort every tick.

Happy to follow up with a PR for both changes.

sergeishestakov-gh:
## Summary

Fixes #3384.

Every replica's scheduler computes the same cron occurrence independently, so on a multi-replica deployment every replica fired the hourly storage-GC pass at the same instant and ran the same Maven flat-object orphan scan concurrently — visible in `pg_stat_activity` as several identical copies of that query sharing one `xact_start` and running for 20+ minutes.

- **Singleton-lease the scheduled tick.** `run_gc()` is now wrapped in the cluster-wide `SchedulerLease` already used for `lifecycle_policy_execution`, under the job name `storage_gc`. Exactly one replica runs it per tick; the rest return immediately; the lease is renewed for as long as the pass runs and explicitly released at the end rather than left to lapse.
- **The on-demand endpoints are deliberately NOT gated.** `POST /api/v1/admin/storage-gc` and the per-repository panel call the GC service directly and must keep working while a scheduled tick holds the lease. There is a regression assertion for this (below).
- **Trigram index for the rollup guard.** The guard that spares a `maven-metadata.xml` while any live artifact remains under its directory builds its `LIKE` pattern per candidate row, and Postgres only rewrites a `LIKE` into an indexable range scan when the pattern is a plan-time constant — so the guard was a hash anti-join against the entire `artifacts` table, re-evaluated once per candidate. A partial `gin_trgm_ops` index on `artifacts.storage_key` (`WHERE is_deleted = false`, matching the guard's own filter) turns it into a bitmap index scan with the correlated pattern as an `Index Cond`. Measured on a 400,000-row `artifacts` table at the attribution-table size the issue reports (136 rows):

  ```
  without idx_artifacts_storage_key_trgm : 54,744 / 54,905 ms
    Hash Right Anti Join -> Seq Scan on artifacts a2 (400,000 rows)
    Rows Removed by Join Filter: 54,400,000   (= 400,000 x 136)
  with    idx_artifacts_storage_key_trgm :     151 /     167 ms   (~345x)
    Nested Loop Anti Join -> Bitmap Index Scan on idx_artifacts_storage_key_trgm
    Index Cond: storage_key ~~ like_escape(substr(...) || '%', '')
  ```

  The un-indexed plan is quadratic in the candidate count, so the gap widens with table size: at 20,000 attribution rows over the same `artifacts` table the indexed scan still ran in 2.5 s while the un-indexed one had not finished after 52 minutes. The `Index Cond` line also shows the `ESCAPE ''` clause from #3492/#3493 still plans as an indexable `~~` — the two changes compose.
- The block's fixed 120-second startup sleep now uses the same jittered startup delay as every other periodic block in the scheduler, and the tick body moved into `StorageGcService::run_scheduled_tick` where it is directly testable.

### What the lease does NOT cover — #3384 is only PARTIALLY closed

The lease covers `run_gc` and nothing else in the tick. **`run_blob_gc_mark`, `run_blob_gc_sweep` and `stats_service.recompute_all()` still run on EVERY replica in the same tick.** Live-measured with two replicas: the leased `maven_flat_object_owner` scan drops to 1/minute while the un-leased `oci_blobs` scans stay at 4–6/minute. Blob GC is also destructive when `BLOB_GC_ENABLED=true`.

There is a second-order effect worth stating plainly: because a losing replica now returns immediately instead of after a ~2-minute GC, **the un-leased half of the tick now fires on every replica at the same instant with none of the accidental stagger the slow path used to provide.** Extending coordination over the rest of the tick is tracked separately in **#3503**; it is not done here because blob GC has its own readiness gate and dry-run default whose interaction with a skipped tick needs its own analysis. Both the call site and `run_scheduled_tick`'s doc comment carry this scope note so it is not mistaken for a full fix.

### Operator note — the migration briefly blocks writes to `artifacts`

`sqlx::migrate` runs each migration inside a transaction, so `CREATE INDEX CONCURRENTLY` cannot be used. The plain build takes a **`SHARE`** lock on `artifacts` for its duration (verified in `pg_locks` during a build: `ShareLock` on the table, `AccessExclusiveLock` only on the new index relation). Reads are unaffected — `SELECT` takes `ACCESS SHARE`, which is compatible — but **uploads across every format block until it completes, and this runs at each replica's startup migration, so the stall lands in the middle of a rolling deploy.**

Measured on this branch:

| | rows | avg key | build | index size |
|---|---|---|---|---|
| realistic Maven corpus | 400,000 | 51 B | **1.87 s** | 28 MB |
| fixed 20k, typical keys | 20,000 | 76 B | 127 ms | 2.0 MB |
| fixed 20k, max-length high-entropy keys | 20,000 | 2,046 B | **3,632 ms** | **60 MB** |

Build time is roughly linear in row count, so ~10M artifacts ≈ 45 s and ~50M ≈ 4 min. Note the third row: **build time and index size are user-influenced**, because `artifacts.storage_key` is `varchar(2048)` — 29× on both axes between typical and max-length high-entropy keys at identical row counts. `artifacts` already carries `idx_artifacts_lower_path_trgm`, so this is the second trigram index on that table whose size a tenant can influence.

Sites that cannot accept the stall should build the index out of band before upgrading; the migration is `IF NOT EXISTS`, so it then becomes a no-op:

```sql
CREATE INDEX CONCURRENTLY idx_artifacts_storage_key_trgm
  ON artifacts USING gin (storage_key gin_trgm_ops)
  WHERE is_deleted = false;
```

### The second migration was dropped after measurement

An index on `maven_flat_object_owner.storage_key` was originally proposed alongside this change, on the theory that it would let the candidate scan terminate early into its `ORDER BY … LIMIT 1000`. **Measured at 400,000 artifacts / 20,000 attribution rows, the planner does not choose it:**

```
with    idx_maven_flat_object_owner_storage_key : 2550.7 / 2527.9 / 2487.9 ms
without idx_maven_flat_object_owner_storage_key : 2452.8 / 2484.0 / 2467.6 ms
plan, both ways: Seq Scan on maven_flat_object_owner o -> top-N heapsort
```

The index does not appear anywhere in the `EXPLAIN (ANALYZE, BUFFERS)` output. The five-way anti-join estimates `rows=1`, so an ordered index scan is never believed to terminate early. It would have cost a second write-blocking build and ongoing write amplification for no measurable gain, so it is not in this PR. The remaining migration was renumbered 207 → **206** (main's highest is 205; no other open PR adds a migration).

### Migration comment corrections

The trigram migration's comment carried three factual errors, all corrected here:

1. It claimed the build takes **ACCESS EXCLUSIVE** on `artifacts`. It takes **SHARE** — measured above, and the dropped migration in the same PR stated this correctly, so the two contradicted each other.
2. It claimed the index also serves guard 3's leading arm (`{base} LIKE '%/maven-metadata.xml'`). It cannot: that arm's operand is `regexp_replace(o.storage_key, …)`, a correlated column of `maven_flat_object_owner`. In both plans it appears only as a `Join Filter` / `One-Time Filter`.
3. Its safety argument cited the bare *"can only ever over-PROTECT"* invariant — which was **false** for a literal backslash until #3492/#3493 landed. The comment now states that guard 3's soundness rests on the `ESCAPE ''` clause `metadata_rollup_dir_anchor_sql` emits, and that dropping that clause reintroduces a silent data-loss bug which this index makes reachable far faster, while dropping the index only makes the sweep slow.

### Interaction with #3431

This PR is the reachability multiplier for the sweep behind the #3431 data-loss incident, so both of that fix's guards were re-verified **through the new leased entry point**, not only through `run_gc`:

- `MAVEN_FLAT_GC_ENABLED` unset ⇒ a tick that *wins* the lease still deletes nothing and only reports (`maven_flat_objects_gated`).
- Opted in, the `metadata_rollup` row is reclaimed while an `operator_repair` row is preserved — the source allowlist still holds.

Neither gate is touched by this change; both are now pinned by a test.

`run_scheduled_tick_respects_the_singleton_lease` now runs against the **real** `SCHEDULED_GC_JOB_NAME` rather than a random job name, so the argument the scheduler actually passes is under test, and it additionally asserts that the on-demand admin endpoint performs a real sweep (not an empty no-op result) while another replica holds the lease. `run_scheduled_tick_honours_the_flat_gc_gate_and_source_allowlist` is new and covers the #3431 interaction above. `scheduled_gc_lease_job_name_is_the_documented_wire_name` pins the operator-visible lease name.

All three are `--lib` tests, i.e. in the set CI actually runs, and none is hollow — four independent production-code reverts each fail the matching assertion:

| production code reverted | result |
|---|---|
| lease acquire/skip removed | `FAIL [0.241s]` … `tick must skip while another replica holds the lease` |
| `lease.release()` removed | `FAIL [0.272s]` … `a completed tick must release the lease, not hold it forever` |
| `MAVEN_FLAT_GC_ENABLED` gate removed | `FAIL [0.333s]` … `left: [false, true]  right: [true, true]` (plus both pre-existing #3431 tests) |
| `system_written_source_sql` allowlist neutered | `FAIL [0.320s]` … `left: [false, false]  right: [false, true]` |
| on-demand endpoint lease-gated | `FAIL [0.228s]` … `must actually run the flat-object sweep …, not return an empty no-op result` |

Restored, all pass in 0.226–0.351 s — well clear of the ~0.04 s "test never ran" tell, with `DATABASE_URL` set and `AK_TESTS_REQUIRE_DB=1`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Migration 206 is additive and idempotent (`CREATE INDEX IF NOT EXISTS`); reversing it is a single `DROP INDEX` and functionality is unaffected without it — it is purely a query-plan accelerator.


sergeishestakov-gh:
## Correction on the "3 duplicate copies" evidence

A closer look at `pg_stat_activity` for the reported instance (single replica) shows the "3 copies" of this query are actually one leader connection plus 2 Postgres **parallel workers** for the same query (`client_addr` is NULL for the workers, all three share an identical `duration` to the microsecond — the standard signature of a parallel leader+worker group, not 3 independent client connections). Details in artifact-keeper/artifact-keeper#3384.

This doesn't change the recommendation in this PR: the missing index on `maven_flat_object_owner.storage_key` is what makes the planner estimate a sequential scan as expensive enough to parallelize in the first place; adding the index removes the incentive to pick that (multi-hour) parallel seq-scan plan.

It does mean the `SchedulerLease` half of this PR is a correctness fix for genuine multi-replica deployments (every replica's scheduler loop really does compute the same cron occurrence independently, with no coordination — that part of the code is unchanged and still worth fixing), but it is not what caused the specific 3-process symptom observed on the single-replica instance that prompted this PR. Keeping both changes since both address real, independent gaps — just correcting the causal claim for the record.

brandonrc:
Heads-up on a merge-order collision that landed while this PR was open, plus one CI note.

**Migration number 196 is now taken on `main`.** `196_oci_manifest_subjects.sql` merged with #3376 (OCI Referrers API), so this branch's `196_maven_flat_object_owner_storage_key_index.sql` now duplicates that version number. This is the failure mode that is invisible in git — the two files have different names, so there is no textual conflict and the merge looks CLEAN — but sqlx keys on the numeric version, so the duplicate aborts the backend at startup on `_sqlx_migrations_pkey`.

Please renumber this migration when you next rebase. `198` is the safe choice right now: 197 is claimed by open PR #3389, and #3390 claims none.

Note that `main` also gained an automated gate for exactly this (#3313, merged today) — once you rebase, CI will check duplicate migration versions numerically, so it will catch a renumber that misses.

Separately, the current CI red here is the **Code Coverage** gate, which is independent of the above. The repo requires >=70% coverage on added/changed lines; extracting the new logic into a pure helper is usually the quickest way to make the changed lines coverable.

Thanks for the contribution — the singleton-lease on the scheduled GC tick is a good catch.

sergeishestakov-gh:
Thanks for the heads-up, both addressed:

- Renumbered 196 → 206 (rebased onto current main, confirmed 197-205 are all taken and no other open PR claims a migration number right now, so 206 is clear).
- Extracted the lease-acquire/run_gc/release sequence out of the scheduler's `tokio::spawn` loop into `StorageGcService::run_scheduled_tick(job_name)`, parameterized on the lease job name so a test can exercise it under a throwaway name instead of the real `storage_gc` job. Added a test that holds the lease as a second replica would, asserts the tick skips, releases it, then asserts the next tick wins the lease, runs, and releases it. The scheduler loop is now just `service.run_scheduled_tick("storage_gc").await;`.

Pushed (rebased + force-pushed the branch, since the migration renumber needed the rebase anyway).

sergeishestakov-gh:
@brandonrc now all checks successfully finished
Thank you in advance

sergeishestakov-gh:
Rebased onto current `main` (now on top of `88153dc8` / #3431 / #3451). This branch previously sat 8 commits behind, predating that fix — flagging since #3431's changelog entry explicitly names this PR's indexing as what "activates" the data-loss bug it fixes (a scan that used to time out before it could delete anything now completes). With this rebase, `run_gc()`'s Maven flat-object sweep already has the system-source allowlist and the `MAVEN_FLAT_GC_ENABLED` opt-in gate underneath it, so the lease + indexing changes here can't reintroduce that regression once merged.

`cargo check` clean post-rebase, no migration numbering conflicts (207 still free, verified against `main` and all other open PRs touching `backend/migrations/`).

brandonrc:
Surfaced while correcting the v1.7.8 Maven-GC advisory, and it argues for triaging
this PR on its own merits rather than as part of the data-loss fix.

**The cost this addresses is real, and it has shipped nowhere.** Migration 168
moved the `maven_flat_object_owner` primary key from `(storage_key)` to
`(storage_backend, storage_key)`. The orphan sweep's candidate scan ends with
`ORDER BY o.storage_key LIMIT 1000`, and a composite PK leading with
`storage_backend` cannot serve that ordering. So on a large attribution table the
scheduled sweep does a full predicate pass plus a sort — hourly, on **every**
replica that runs the tick, with no singleton lease (`scheduler_service.rs` calls
`run_gc(false)` directly).

`backend/migrations/206_maven_flat_object_owner_storage_key_index.sql` from this
PR is what fixes that. It exists on no branch — not `main`, not `release/1.7.x`.

**Separately: it never masked the data-loss bug, and the v1.7.8 notes no longer
say it did.** The 1.7.7 prep's advisory claimed "the 1.7.3 guard rebuild plus the
#3385 indexing made this scan *complete* on large attribution tables where it
previously timed out," and used that to tell operators on 1.7.0–1.7.2 the bug was
largely masked for them. That was wrong three ways:

- the 1.7.3 guard rebuild (#3156/#3197) only ever made the predicate *more*
  protective and *more* expensive — a widening cannot convert a timing-out scan
  into a completing one;
- there is no runtime `statement_timeout` on the GC path at all (the only
  occurrences are session-local in the migration runner), so there was nothing to
  time out against;
- and the indexing it credited has never existed on any branch.

v1.7.8 now states 1.7.0 through 1.8.0 are uniformly affected. Recording it here
because this PR was cited as load-bearing for a claim it could not support, and
anyone reading that history should not conclude the PR is unnecessary — the
opposite. It addresses a genuine recurring cost; it simply is not the reason
anyone lost data.

Worth deciding on its own terms: the sort happens every hour on every replica
today.

Issue #3377, opened 2026-08-14
Flaky lib tests: http_client sets HTTP_PROXY process-wide, racing test_read_json_capped_* under parallel test runs


brandonrc:
## Summary

`cargo test --workspace --lib` intermittently fails 4 tests — `services::http_client::tests::test_read_json_capped_*` — when run with default parallelism, while all of them pass in isolation (`5/5 ok`).

## Root cause

A test in `backend/src/services/http_client.rs` (around line 896) sets `HTTP_PROXY` via `std::env::set_var`, which is process-wide. Other tests in the same binary that construct HTTP clients concurrently observe the leaked proxy env and fail. Classic cross-test env mutation race.

## Evidence

Observed on `fix/oci-v2-spec-compliance-cluster` (rebased on main @ 3871639) during the pre-push check for #3376: 15192 passed, 4 failed, all four `test_read_json_capped_*`; immediately green when re-run in isolation. The failing tests do not touch the diff of that branch.

## Suggested fix

Serialize the env-mutating tests (e.g. `serial_test` crate or a shared `Mutex`), or refactor the client to take proxy settings via injected config in tests instead of reading process env, and scope/restore the var with a guard.

brandonrc:
Additional evidence from the 1.8.0 fix waves: this suite does not only fail intermittently, it **hangs indefinitely**.

Two independent observations on separate branches today:

- A `cargo test --workspace --lib` run wedged inside `services::http_client::tests` and sat at **0% CPU for 7 hours** before being killed manually. Because the rig serializes Rust builds behind a 2-token semaphore, that one wedged binary held a build slot for its entire lifetime and throttled every other concurrent build.
- A second worker's first full run wedged on the same suite for 32 minutes before it was killed, and completed normally once the suite was filtered out.

So the impact is worse than a flaky red: on any machine that runs the suite with default parallelism, a hung test binary can occupy a CI runner or a local build slot until something reaps it, and "still running" is indistinguishable from "hung".

Suggested addition to the fix: alongside serializing the env-mutating tests (or removing the process-wide `set_var`), give the local-server tests an explicit **timeout** so a wedge fails fast and loudly instead of hanging. Anything that blocks on a socket that never gets connected — a server task that failed to bind while `HTTP_PROXY` was leaked into the client — should surface as a failed assertion, not an indefinite block.

Issue #3331, opened 2026-08-13
REST content paths gate artifact bytes on any grant, not the read action (user_can_access_repo is action-blind)


brandonrc:
## Summary

`RepositoryService::user_can_access_repo` answers **"does this principal hold any grant on this repository"**, not **"may this principal read it"**. It is the gate in front of REST paths that return artifact **bytes** and private repository **contents**, so a principal whose only entitlement is `{write}` or `{delete}` — or a role assignment to a role carrying no `read` — can read those bytes.

This is the same defect class as #3325, which was closed for the virtual-repository byte-resolution path. It is filed separately because it is **larger than the one that was fixed** and cannot ship in a patch: closing it changes behaviour on ~35 REST surfaces including enumeration ones.

It is not a footnote to a closed issue. #3325 closed one path; this is the rest of the class, and it is the bigger half.

## The predicate

`backend/src/services/repository_service.rs:1008`:

```
SELECT EXISTS (
    SELECT 1 FROM role_assignments ra
    WHERE ra.user_id = $1 AND (ra.repository_id = $2 OR ra.repository_id IS NULL)
) OR <permissions_grant_exists("$2", 1)>
```

Neither arm carries an action term:

* `permissions_grant_exists_for` (`repository_service.rs:372`) tests `AND p.actions <> '{}'` — any non-empty array qualifies. `permissions.actions` is `TEXT[]`, not the `permission_type` enum, so an unrecognised string qualifies too.
* the `role_assignments` arm never joins `roles`, so the role's own `permissions` array is never consulted.

The canonical action decision is `PermissionService::check_repository_action(user, repo, "read", is_admin)` (`permission_service.rs:162`), which the OCI `/v2` read gate (`oci_v2.rs:725`) and the native-format `repo_visibility_middleware` (`auth.rs:1959`) both call. `user_can_access_repo` is a strict superset of it.

## Affected surfaces

**Content paths — these return bytes or private contents on any grant:**

| Path | Gate |
|---|---|
| `GET /api/v1/artifacts/{id}` and its label / security / SBOM / quality-gate siblings | `artifacts.rs:62` `check_artifact_visibility` |
| `GET /api/v1/tree/content` — serves artifact bytes | `tree.rs:91` |
| `GET /api/v1/repositories/{key}` and 33 further call sites of `require_visible` | `repositories.rs:400` |
[...]

PR #3399, opened 2026-08-15
fix(security): authorize virtual-repo members on the read paths across formats

### LINKED ISSUE #3323: Virtual-repo member authorization is missing on ~40 read paths across 20 formats (3 are anonymous BYTE leaks)

brandonrc:
## Summary

Closes the virtual-repository member-authorization gap on the READ paths, across every format.

The format-route visibility middleware (`repo_visibility_middleware`) authorizes only the repository named in the URL. For a Virtual repository that is the **parent** — while everything the virtual aggregates is a separate repository with its own ACL. `should_allow_repo_access(is_public, has_auth)` admits an anonymous caller on a public repo, so a public virtual over private members resolved each member's content with no further check.

The predicate was not missing — `proxy_helpers::authorize_virtual_members` already composes exactly what a direct read of that member would apply (tenant gate + token scope + the `read` action gate from #3326). It was applied on a minority of call sites, and the two shared metadata primitives took **no caller parameter at all**, so all of their callers inherited an unfiltered walk without ever making a decision.

### What changed

**One shared helper.** `proxy_helpers::authorized_virtual_members(db, auth, virtual_repo_id)` is `fetch_virtual_members` composed with `try_authorize_virtual_members` (fallible form, so a failed visibility query is a retryable error rather than a false "not found"). Every content-serving virtual read path now calls it:

PyPI simple index (root + per-project) · npm packument and version metadata · cargo sparse index · Helm `index.yaml` · NuGet V2/V3 search, registration and flat-container version indexes · Conda `repodata`/`channeldata` · CRAN `PACKAGES` · Hex names/versions/package · RubyGems `specs.4.8.gz` family and quick gemspecs · Conan search/revisions/package listings (8 sites) · Composer `packages.json` and p1/p2 metadata · Pub package and version info · Swift release listing and metadata · Go `@v/list` and `@v/{v}.info` · Debian `Release`/`InRelease`/`Release.gpg`/`Packages` · Alpine `APKINDEX` · Maven root listing · Hugging Face `ETag`/`X-Repo-Commit` resolution.

**The two primitives now require a caller.** `resolve_virtual_metadata` and `collect_virtual_metadata` take `auth` as a required parameter, so each of their ~15 callers had to make the decision explicitly (a compile error, not a review question) and cannot silently regress.

**Caller-independent caches in front of now caller-dependent documents.** Narrowing the aggregation alone would have moved the leak, not closed it: three caches are keyed by repository/document and not by caller, so the first authorized request would store a private member's contribution and every later anonymous request would be served it. The npm computed-packument cache, the cargo in-process sparse-index cache and the Maven virtual `maven-metadata.xml` merge cache are now used only when the aggregate cannot vary by caller. That condition is exact and is `virtual_aggregate_is_cacheable`: every member public — `try_authorize_virtual_members` admits a public member for every caller unconditionally, so an all-public member set authorizes identically for anonymous and authenticated callers. All-public virtuals keep their cache; a virtual with a private member computes per request.

**Enforcement walks are deliberately left unfiltered.** The OCI scan-verdict deny-set, the PEP 708 isolation guard, the `virtual_non_remote_owns_*` shadowing guards, priority maps, cache invalidation and cycle detection all need the full member set: narrowing a deny-set or a shadowing decision by caller visibility would let a caller who cannot see a member escape that member's gate, or turn a shadowing decision into a dependency-confusion vector. The OCI blob scan-verdict fan-out carries an explicit `UNFILTERED-ENFORCEMENT` marker saying so.

### Not included

The RPM `repodata` walk (`repomd.xml`, `primary`, `filelists`, `other`, `updateinfo`). Its document is rendered once and cached per **virtual repo id**, and `repomd.xml.asc` is a detached signature over the rendered bytes — a per-caller document breaks both the shared cache and the reproducible-signature invariant (#2636). Making it caller-dependent requires re-keying that cache by the authorized member-id set, which is a design decision rather than a mechanical filter. It carries an explicit `UNFILTERED-DEFERRED` marker in the source and stays tracked on #3323.

The three anonymous BYTE leaks called out at the top of #3323 (Helm chart download, NuGet V2 `.nupkg`, Swift `Package.swift`) were already fixed on `main` by #3328; this PR is the metadata and remote-member remainder plus the shared-helper consolidation. The read-action half raised in #3331 landed in #3326 and is inherited here by construction, since every path now goes through the same gate.

The class-level test is `every_raw_virtual_member_walk_is_authorized_or_explicitly_exempt` (`proxy_helpers.rs`): it reads every format handler at compile time and fails on any `fetch_virtual_members` call site that neither pairs with an authorization call nor carries an `UNFILTERED-ENFORCEMENT` / `UNFILTERED-DEFERRED` marker. On `main` it reports ~25 unguarded sites. A test pinned to the handful of sites one PR touches would go green while the next format re-opened the hole, so the gate scans all of them — the pattern CLAUDE.md asks for.

Behavioural tests that fail on `main`:
- `pypi::virtual_index_member_authz_tests::simple_index_hides_a_private_members_projects_from_an_anonymous_caller` — anonymous root index and per-project index over a public virtual with one private and one public local member; the private member's project name and filename must be absent, the public member's present (positive control), and an admin must still resolve the private one.
- `npm::virtual_packument_member_authz_tests::packument_hides_a_private_members_versions_from_an_anonymous_caller` — the merged packument must not carry the private member's version for an anonymous caller, must still carry the public member's, and must carry both for an admin.
- `proxy_helpers::virtual_read_authz_tests::authorized_virtual_members_drops_a_private_member_for_an_anonymous_caller` — the shared predicate itself, plus `virtual_has_private_member` / `virtual_aggregate_cacheable` in both directions.
- `the_shared_metadata_primitives_require_a_caller` — pins the required `auth` parameter on both primitives.
- `packument_cache_eligible` and `virtual_aggregate_is_cacheable` unit tests — the cache-shareability decision, DB-free.

## Test Checklist
- [x] Manually tested locally

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings` and `cargo test --workspace --lib -- --skip services::http_client` (#3377) are all green locally: 15346 passed, 0 failed. No database was available locally, so the DB-backed tests no-op here and run for real in the coverage job (`AK_TESTS_REQUIRE_DB=1`).

## API Changes
- [x] N/A - no API changes

No route, request or response shape changes: the handlers gain an `Extension<Option<AuthExtension>>` extractor that the middleware already inserts on every format route. `backend/tests/swift_virtual_lookup_tests.rs` now seeds its members public, since its fixtures probe with no caller and the subject there is the #1554 fan-out.

Fixes #3323


brandonrc:
### Follow-up: DB-backed fixtures updated (`cf2cdb4`)

The first push was red. Root cause, verified against a real Postgres 16 rather than reasoned about: **31 pre-existing DB-backed fixtures across 12 formats linked a member created by `test_db_helpers::create_repo` — which leaves `is_public` at its `false` default, so the member is PRIVATE — and then probed with no caller at all.** They passed before this PR only because the member walk was unauthorized. Under the fix they correctly resolve nothing.

I explicitly tested for the over-narrowing hypothesis first, and it does not hold. On the same fixture, with the member left private:

| member | caller | result |
|---|---|---|
| private | anonymous | denied ← the #3323 fix |
| **public** | anonymous | **resolves** ← public short-circuit intact |
| **private** | **granted user** | **resolves** ← full read predicate intact |

So legitimate access is unaffected in both directions; only the unauthorized walk stopped working.

No assertion in any of those tests was weakened or removed. Each fixture now supplies the caller entitlement it always implicitly assumed:

- fixtures that probe **with** a caller (composer, conan) `grant_repo_access` that user on the member — this keeps the member private and exercises the *full* predicate rather than the public short-circuit, which is strictly stronger than publishing;
- fixtures that probe **anonymously** publish the member through a new `test_db_helpers::publish_repo`, the same remedy #3328 applied to the helm fixture for exactly this reason.

Two shared fixture helpers absorbed six failures on their own (`create_remote_and_virtual`, the Debian virtual builder). The npm packument-cache fixtures needed the member published for a second, less obvious reason: a virtual with a private member is no longer packument-cache eligible, so those tests would otherwise have warmed and asserted nothing.

`composer::virtual_p2_resolves_from_local_member` additionally gained the negative direction — the same request with no caller must not resolve the private member — so the boundary is pinned in the file where the regression surfaced.

**Verification** (Postgres 16, all 197 migrations, `AK_TESTS_REQUIRE_DB=1`): `15365 passed, 1 failed`. The single failure is `services::oci_migration_reindex::tests::test_run_repair_drops_orphan_tag_keeps_healthy`, which is **pre-existing and unrelated** — I reproduced it on unmodified `main` against the same database, and it passes in isolation on both. It is a DB-state-dependent flake in that module; CI's `cargo nextest` runs each test in its own process against a fresh database. `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` are clean.

Issue #3352, opened 2026-08-14
Build grype from source with a patched Go toolchain (as #3123 does for trivy)


brandonrc:
The bundled `grype` CLI in the backend images is the last vendored scanner binary whose CVEs we cannot fix ourselves.

## Problem

`ghcr.io/anchore/grype:v0.117.0` (latest upstream, published 2026-08-10) is built with Go 1.26.3. Six Go-stdlib CVEs are present, four of them HIGH+fixed. Because Anchore has not rebuilt against a patched toolchain, our only lever is `.trivyignore`, so each new Go stdlib advisory becomes a release blocker that we resolve by writing a suppression rather than by shipping a fix.

This has now happened repeatedly:

- #3307 — bumped v0.116.0 -> v0.117.0, which did not move the toolchain
- 1.7.4 publish — CVE-2026-39821 (x/net/idna, Punycode label processing) landed in the Trivy DB and failed `Security Scan` on an otherwise-clean build

The backend and openscap images scan clean at 0 findings apart from this binary.

## Why suppression is weak here

The current runtime scope argument is real but narrow: the images set `GRYPE_DB_AUTO_UPDATE=false` and `GRYPE_CHECK_FOR_APP_UPDATE=false` against a build-time pre-seeded DB (#1001), and grype is invoked against on-disk artifacts and SBOMs, so it makes no outbound HTTP at runtime. That means the affected `net/http` path is not driven — but the vulnerable code is still linked into a binary we ship, and the argument has to be re-litigated on every advisory.

## Proposal

Do for grype what #3123 did for trivy: build it ourselves from the pinned upstream tag with a current Go toolchain, publish to `ghcr.io/artifact-keeper/grype`, and consume that by digest. A stdlib CVE then becomes a rebuild instead of a suppression.

Scope:

- [ ] Build repo/workflow for grype mirroring `artifact-keeper/trivy` (pinned tag + toolchain, `overrides.yaml` equivalent for Go-dependency bumps)
- [ ] Repoint `docker/Dockerfile.backend` and `docker/Dockerfile.backend.alpine` at the digest-pinned image
- [ ] Upstream-watch workflow for new Anchore releases
- [ ] Delete the grype blocks from `.trivyignore` once the rebuilt image scans clean

## Acceptance

`.trivyignore` contains no grype entries, and a new Go stdlib advisory is cleared by rebuilding rather than by editing a suppression file.

brandonrc:
[...]

Issue #3339, opened 2026-08-13
Release preflight does not catch a version-pinned image tag colliding with published content (killed the v1.7.2 tag)


brandonrc:
## What happened

`v1.7.2` was tagged on `a0dfc7b3`. Its Docker Publish failed:

```
ghcr.io/artifact-keeper/artifact-keeper-scanner-adapter:1.2.2 already exists at
676f12ca69e985e46d2725ee758ee94d1275dc03, but scanner-adapter sources changed.
Bump docker/scanner-adapter/VERSION; exact version tags are never republished.
```

The go-git CVE fix changed `docker/Dockerfile.scanner-adapter` — it re-pinned the bundled trivy base image from `sha256:202c9b4c…` to `sha256:d6c56ef9…` (the in-house `0.73.0-r2` build carrying the go-git `v5.19.1` → `v5.19.2` override) — but left `docker/scanner-adapter/VERSION` at `1.2.2`.

Exact version tags are never republished, so the publish failed. That skipped every remaining multi-arch manifest, `resolve-candidate-digest` had nothing to pin, and the whole release chain stopped. The tag was immutable, so it could not be amended and had to be deleted by hand.

## Why nothing caught it

`Docker Publish` run `31675941091` on `main`, for the **same commit** `a0dfc7b3`, passed `Scanner Adapter Multi-Arch Manifest`.

The publish job derives `stable_requested` from the ref:

```bash
if [[ "$GH_REF" == refs/tags/v* && "$GH_REF" != *-* ]]; then
  echo "stable_requested=true" >> "$GITHUB_OUTPUT"
```

Only a clean `refs/tags/v*` sets it. On `main` the job publishes dev/sha tags and returns before the collision check ever runs. A `-rc.N` tag contains a hyphen, so it takes the same early return — an RC would **not** have caught this either.

The defect was reachable by exactly one kind of run: the real tag. That is the most expensive place to discover it and the only place it cannot be amended.

## Fix

1. Bump `docker/scanner-adapter/VERSION` to `1.2.3`. The content genuinely differs — the published `1.2.2` ships the older trivy build, so republishing it would be wrong on the merits, not merely blocked by policy.
2. Add a check to `scripts/ci/release-preflight.sh` that fails **before tagging** when a component's pinned version tag already exists in the registry with different content, following the file's existing `ok`/`warn`/`bad`/`note` vocabulary and its `INFRA` exit 2 for "could not measure".
[...]

Issue #3338, opened 2026-08-13
release-preflight check 3 selects the latest Docker Publish run, not the run for the commit being tagged (blocking false red on the 1.7.2 cut)


brandonrc:
## Summary

`release-preflight.sh` check 3 asks *"is the latest main Docker Publish healthy?"* when the question that matters before a tag is *"did Docker Publish succeed for the commit I am about to tag?"* Those differ, and the difference produced a blocking false red during the v1.7.2 cut.

## What happened

Preflight run `31694643135` (2026-08-13 11:14:57Z) reported:

```
run 30830506078: Security Scan=failure, Backend Multi-Arch Manifest=skipped
[FAIL] main's last Docker Publish did NOT cleanly publish the manifest -- a tag cut WILL stall on the same gate.
NOT READY: 1 blocking problem(s).
```

Run `30830506078` is from **2026-08-03**, head `b5c30748` — ten days stale. The actual latest Docker Publish on `main` at that moment was `31675941091` (head `a0dfc7b3`, 2026-08-13T06:59Z, **success**, all manifests published).

Ruled out, by checking rather than assuming:
- **Not a duplicate workflow name.** Only one workflow is named `Docker Publish` (id `224314536`, `.github/workflows/docker-publish.yml`), and both runs belong to it.
- **Not a deterministic query bug.** Re-running the same `gh run list --workflow "Docker Publish" --branch main --limit 1` a few minutes later returned the correct `31675941091`, matching the REST endpoint.

So it was transient API inconsistency in the "give me the newest run" query. Which means it can recur, silently, at the worst possible moment.

## Why "latest" is the wrong question anyway

Even with a perfectly consistent API, "latest run on main" is a proxy. A tag is cut on a **specific commit**, and what determines whether the release chain stalls is whether *that commit's* images published. Those diverge whenever main has moved since the last publish, when a publish is still in flight, or when an unrelated later run failed.

The precise query already exists and needs no new machinery:

```sh
head_sha="$(gh api "repos/$REPO/commits/main" --jq .sha)"
run_id="$(gh api "repos/$REPO/actions/workflows/docker-publish.yml/runs?head_sha=$head_sha" \
            --jq '.workflow_runs[0].id')"
```

Verified against the v1.7.2 candidate: `head_sha=a0dfc7b3` resolves to exactly one run, `31675941091`, `completed/success`.

## Requested

1. Select the Docker Publish run by **`head_sha` of the ref being tagged**, not by recency.
[...]

PR #3421, opened 2026-08-16
fix(ci): select the Docker Publish run by head_sha on release/1.7.x

### LINKED ISSUE #3422: release/1.7.x: port the per-commit Docker Publish preflight check (#3338) so v1.7.5 can be tagged
## Problem

`v1.7.5` is prepared on `release/1.7.x` (`chore(release): prepare 1.7.5`, #3379) but has never been tagged, because **Pre-tag readiness** fails:

```
[FAIL] main's last Docker Publish did NOT cleanly publish the manifest
       -- a tag cut WILL stall on the same gate.
```

This is #3338 on the maintenance branch: check 3 selects the Docker Publish run **by recency on `main`** rather than the run for the commit being tagged. #3364 fixed that on `main`; `release/1.7.x` still has the old copy.

For this cut the two questions give opposite answers:
- the 1.7.5 tip's own Docker Publish **succeeded** (2026-08-14 19:31)
- `main` is red for an unrelated reason — bundled grype carries 6 HIGH Go stdlib CVEs from its build toolchain (#3352)

Judging a 1.7.x tag on main's publish health measures the wrong branch entirely.

## Why this is urgent

**Three security fixes are fixed-but-unreleased, all waiting on this tag:**

- GHSA-8jm4-4x6c-6787 — client-controlled `X-Forwarded-For` prefix bypasses IP rate limiting
- GHSA-95fx-g94v-8jqg — backup restore inserts rows into tables outside the export allowlist
- GHSA-fv45-mwhh-q23r — percent-encoded repository key bypasses private-repository visibility

They are on `release/1.7.x` and on `main`, and in no tag. Every 1.7.4 deployment is still exposed.

## Complication worth recording

The natural fix — porting `scripts/ci/release-preflight.sh` and its self-test from `main` — cannot satisfy the **"Verify commits trace back to main"** gate, which matches by `git patch-id` against main's history. Only a true cherry-pick passes, and the containing commit (`3a61fa3`) also carries a `.trivyignore` rewrite retiring 8 suppression tokens, deletion of the committed scanner-adapter binary, a VERSION bump, and e2e waiver changes — none of which belong on a patch release.

So this is a genuine gap in the branch policy rather than a mistake by either gate: a maintenance-branch-only port of a *subset* of a main commit is unrepresentable. The `release-process: approved` escape hatch exists for this case.

## Acceptance

- `release/1.7.x` evaluates check 3 against the tagged commit's own publish run
- `scripts/ci/test-release-preflight.sh` passes on the branch (verified locally: all cases pass, including the "no run for this commit" and "still in flight" states the old copy could not distinguish)
- `v1.7.5` can be tagged and the three advisories above can be published with a real patched version

PR: #3421

brandonrc:
## Summary

Ports the release-preflight fixes from #3364 (plus #3339/#3340 refinements) onto `release/1.7.x`, so the pre-tag gate asks the right question when cutting a 1.7.x release.

**Why this is blocking right now:** `v1.7.5` is prepared on this branch (`chore(release): prepare 1.7.5`, #3379) but has never been tagged, and "Pre-tag readiness" fails with:

```
[FAIL] main's last Docker Publish did NOT cleanly publish the manifest
       -- a tag cut WILL stall on the same gate.
```

That is the #3338 defect: check 3 selected the Docker Publish run **by recency on `main`** rather than the run for the commit being tagged. For this cut the two diverge completely — the 1.7.5 tip's own Docker Publish **succeeded** (2026-08-14 19:31), while `main` is separately red for an unrelated reason (bundled grype carries 6 HIGH Go stdlib CVEs from its build toolchain; tracked as #3352). Judging a 1.7.x tag on main's publish health is measuring the wrong branch entirely.

With this change check 3 resolves `head_sha` and selects that commit's run, reporting *no run for this commit* and *run still in flight* as distinct blocking states.

**Release impact:** three security fixes are currently fixed-but-unreleased, all waiting on this tag — GHSA-8jm4-4x6c-6787 (X-Forwarded-For rate-limit bypass), GHSA-95fx-g94v-8jqg (backup restore table allowlist), and GHSA-fv45-mwhh-q23r (percent-encoded repository-key visibility bypass).

Scope is CI tooling only: two scripts, no product code, no migration.

`scripts/ci/test-release-preflight.sh` comes with the fix and covers both directions, including the two states this branch's copy could not distinguish (no run for the commit vs. run in flight). Executed on this branch: **all release-preflight cases passed**, including checks 1, 3 and 4.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [ ] OpenAPI spec validates
- [x] N/A - no API changes

Closes #3422

### DIFF
--- a/scripts/ci/release-preflight.sh
+++ b/scripts/ci/release-preflight.sh
@@ -14,20 +14,41 @@
 # nothing checked it before tagging. This gate is that check.
 #
 # What it verifies:
-#   1. TRIVYIGNORE DRIFT (hard): main's CVE/GHSA suppression set must be a
-#      SUPERSET of every active `release/*` branch's. A CVE suppressed on a
-#      hotfix branch but not forward-ported to main means main's images fail
-#      Security Scan -> no release can be cut from main. This is the exact
-#      #3039 gap. Enumeration mirrors check-migration-ledger.sh.
+#   1. TRIVYIGNORE DRIFT (hard): every suppression on every active `release/*`
+#      branch must be ACCOUNTED FOR on main -- either as a live suppression
+#      token, or as an explicit `# RETIRED: <token>` tombstone line. A CVE
+#      suppressed on a hotfix branch and neither carried nor tombstoned on
+#      main means main's images fail Security Scan -> no release can be cut
+#      from main. This is the exact #3039 gap. Enumeration mirrors
+#      check-migration-ledger.sh.
+#
+#      The tombstone form exists because a plain superset rule makes the set
+#      monotonic (#3309): a suppression provably dead for main's images could
+#      never be DELETED while any release branch still listed it, so the file
+#      drifted from "the residual set" toward "everything ever worked
+#      around". A `# RETIRED: <token>` line is the machine-read record that
+#      main removed the token DELIBERATELY (the finding is absent from main's
+#      images), as opposed to never having received it (the #3039 forward-port
+#      miss, which still hard-fails). Retiring is a reviewable diff, and a
+#      token listed BOTH live and retired is a contradiction and fails.
 #   2. VERSION-SET CONSISTENCY (hard): the workspace version in Cargo.toml,
 #      the OpenAPI info version in backend/src/api/openapi.rs, and the
 #      artifact-keeper-backend entry in Cargo.lock must all agree. A partial
 #      bump ships a stale version string (see RELEASING.md step 2).
-#   3. MAIN DOCKER-PUBLISH HEALTH (hard, when it can be measured): whether the
-#      most recent `Docker Publish` run on main published its multi-arch
-#      manifest (i.e. Security Scan passed). A red/skipped manifest here does
-#      not merely "predict" the same stall on the tag -- it is the same job,
-#      on the same images, with the same gate. If it is red, the cut stalls.
+#   3. DOCKER-PUBLISH HEALTH FOR THIS COMMIT (hard, when it can be measured):
+#      whether the `Docker Publish` run for the EXACT commit being tagged
+#      (HEAD) published its multi-arch manifest (i.e. Security Scan passed).
+#      A red/skipped manifest here does not merely "predict" the same stall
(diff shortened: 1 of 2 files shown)


brandonrc:
Applying the **`release-process: approved`** escape hatch, with the reasoning recorded here so the bypass is reviewable rather than silent.

**What is being bypassed:** the "Verify commits trace back to main" gate, which matches by `git patch-id` against main's history and therefore only accepts a true cherry-pick.

**Why a cherry-pick is not the right answer here.** The fix this PR needs is already on `main`, but it lives inside `3a61fa3`, which also carries:

- a `.trivyignore` rewrite retiring 8 suppression tokens,
- deletion of the committed `docker/scanner-adapter/scanner-adapter` binary,
- a scanner-adapter `VERSION` bump,
- e2e soft-fail waiver changes.

Cherry-picking it wholesale would drag all of that onto a patch release. The `.trivyignore` retirements are the sharpest objection: they were measured against **main's** images, and applying them to 1.7.x could un-suppress findings on images that were never re-measured — changing what the 1.7.x scan gate reports as a side effect of a CI-tooling port.

So the gate is not wrong and this PR is not wrong; the policy simply cannot express "port a *subset* of a main commit to a maintenance branch". That gap is written up in #3422.

**Why the risk here is low:** the diff is two CI scripts, no product code, no migration, no workflow behaviour change beyond the check itself. `scripts/ci/test-release-preflight.sh` ships with the fix and was executed on this branch before pushing — **all release-preflight cases passed**, including the "no run for this commit" and "still in flight" states the branch's old copy could not distinguish.

**What it unblocks:** three fixed-but-unreleased security issues, all waiting on the `v1.7.5` tag — GHSA-8jm4-4x6c-6787, GHSA-95fx-g94v-8jqg, and GHSA-fv45-mwhh-q23r. Every 1.7.4 deployment is exposed to all three today, and the advisories cannot honestly name a patched version until this tag exists.

Issue #3287, opened 2026-08-11
Proxy scan verdicts never invalidate on CVE-DB updates: scanner_version records the CLI version only


brandonrc:
Surfaced while preparing the decision brief for #3244. Not caused by any recent PR — this is how the freshness check has always behaved.

## The claim vs the code

`verdict_is_fresh` (`backend/src/services/proxy_scan_service.rs:55-57`) documents its own reuse rule as:

> a CVE-DB bump changes the version string

and migration 181 describes the stored token as "grype-0.8x + db build". The freshness comparison is therefore intended to invalidate a cached verdict whenever the vulnerability database moves.

**It does not.** The token that is stored and compared is the scanner **CLI** version only — `grype --version` yields `grype-0.83.0` (`backend/src/services/grype_scanner.rs:1517-1523`, consumed at `backend/src/services/scanner_service.rs:3197-3220`). The DB build is not part of it.

## Consequence

Grype ships CVE-database updates daily; the CLI version changes rarely. So:

- A verdict recorded today is reused unchanged until the **grype binary itself** is upgraded, or until `DEDUP_TTL_DAYS = 30` expires (`proxy_helpers.rs:6190-6195`, `scanner_service.rs:52`).
- A **withdrawn or corrected advisory** — the common false-positive case — does not unblock the image. It stays blocked for up to 30 days even on the manifest path, which is the path that *does* have freshness checking.
- Symmetrically, a **newly published advisory** does not re-block an image already graded clean within the window.

This undermines the intent of #2976: the freshness signal exists precisely so that a moving CVE database re-decides a cached verdict, and it is inert with respect to the database.

## Suggested direction

Include the vulnerability-database build identifier in the stored token, so it changes when the DB does. Grype exposes DB metadata (`grype db status` / the DB checkpoint, which carries a build timestamp and schema version) — capture that alongside the CLI version and store the pair, e.g. `grype-0.83.0+db-2026-08-10`.

Then `verdict_is_fresh` behaves as its own doc comment already claims, and the practical stale window drops from ~30 days to ~1.

Worth noting this also reduces the operational need discussed in #3244: a large share of "this image is wrongly blocked" cases resolve themselves within a day once the DB build participates in freshness, rather than requiring manual intervention.

## Verification note

[...]

Issue #1001, opened 2026-05-03
Pre-seed Grype vulnerability DB in backend Dockerfile (v1.1.10)


brandonrc:
## Problem

The Grype vulnerability scanner subprocess (\`grype\`) fails on first invocation in network-restricted ARC runner pods because the vulnerability database is not pre-populated. \`grype\` tries to fetch the DB from \`grype.anchore.io\` at scan time, which fails on egress-restricted pods.

This causes \`test-scan-completes.sh\` in the release-gate security suite to fail deterministically:
- \`Internal error: Grype scan failed (exit exit status: 1):\`
- 5 passed, 3 failed within the test
- Surfaces because #996 removed silent-success — the previous symptom was \`scan_type=image\` rows with \`findings_count=0\` and no error

This was confirmed in the v1.1.9-rc.1 release-gate runs (artifact-keeper-test runs 25265469440 and 25265748133), both deterministically failing on the same Grype path.

## Why not blocking v1.1.9

- Quality gate is LAST-scanner-wins (\`policy_service.rs:67-79\` reads \`LIMIT 1 ORDER BY created_at DESC\`). Trivy succeeds and satisfies \`block_unscanned\`.
- Coverage gap is low-to-medium (Trivy covers most of the same CVEs).
- 6-agent triage earlier in the v1.1.9 work converged on \"NOT a v1.1.9 regression\" — longstanding gap unmasked by #996's correctly-working silent-success removal.

## Proposed fix (~5 LoC)

In \`backend/Dockerfile\` (or wherever \`grype\` is installed):

\`\`\`dockerfile
RUN curl -sSfL https://raw.githubusercontent.com/anchore/grype/main/install.sh | sh -s -- -b /usr/local/bin v0.110.0 && \\
    grype db update
\`\`\`

(or whatever the current install pattern is; the key is \`grype db update\` at build time so the DB ships in the image)

## Bonus fix

The error message in \`grype_scanner.rs\` produces \`exit exit status:\` because it uses \`{}\` formatter on \`ExitStatus\` whose Display already includes \`exit status: N\`. Cosmetic but distracting:

\`\`\`rust
// Before:
format!(\"Grype scan failed (exit {}): {}\", output.status, stderr)
// After:
format!(\"Grype scan failed ({}): {}\", output.status, stderr)
\`\`\`

## Related

- artifact-keeper#886 (v1.1.9 release coordination)
- artifact-keeper-test#... (companion test-fix issue)

PR #3424, opened 2026-08-16
fix(security): build grype from source with a patched Go toolchain (1.7.x)

### LINKED ISSUE #3352: Build grype from source with a patched Go toolchain (as #3123 does for trivy)

### LINKED ISSUE #3422: release/1.7.x: port the per-commit Docker Publish preflight check (#3338) so v1.7.5 can be tagged

brandonrc:
## Summary

Cherry-picks `b050758` (#3423, `Fixes #3352`) from `main` onto `release/1.7.x` so `v1.7.5` can be cut with clean images.

**Why this blocks the release.** `Security Scan` in `docker-publish.yml` fails on this branch (run 31919400928 on `a5aaf18`) with 6 HIGH Go stdlib CVEs against `usr/local/bin/grype` — CVE-2026-33818, -56853, -56858, -56859, -56860, -56862. The binary was copied from `ghcr.io/anchore/grype:v0.117.0`, which is compiled with go1.26.3. These are **compiled-with** CVEs, so bumping the pinned grype tag cannot fix them; v0.117.0 is already current.

Note the publish on `7aab4b7` **succeeded** on 2026-08-14. Nothing about the image changed — the vulnerability database caught up. "It passed last week" was never evidence.

Both backend Dockerfiles now compile upstream grype v0.117.0 from source on a Go toolchain floored at 1.26.6, with `GO_MIN_PATCH` asserted at build time so a lagging mirror fails loudly rather than silently shipping vulnerable stdlib.

**Measured on `main` with the gate's own trivy invocation, no ignorefile:** findings against the grype binary drop from **12 → 1** at HIGH+fixed and **21 → 8** across all severities, with zero added. All thirteen removed are stdlib. `grype --version` still prints exactly `grype 0.117.0` and `Supported DB Schema` is still 6, so `GrypeScanner` parsing, the `scanner_version` composition from #3287, and the air-gapped DB pre-seed (#1001) are all unaffected.

## Deliberately narrowed for a patch release

The upstream commit also carried changes that belong only on `main`; those hunks were resolved to this branch's versions rather than imported:

- `backend/src/config.rs` — the rewritten Dockerfile guard test depends on scaffolding introduced by #3413/#3415, neither of which is on this branch.
- `.github/workflows/scheduled-container-scan.yml` — scan-report prose from #3413.
- `.trivyignore` — this branch keeps its own suppression set. Retiring tokens here is unnecessary: `main` carries `# RETIRED:` tombstones for them, which is exactly what preflight check 1's superset rule accepts (#3309).
- `CHANGELOG.md` — the entry is filed under `## [1.7.5]`, where it actually ships, not under `## [Unreleased]`.

Net effect on this branch is the three Dockerfiles, the publish workflow, and one changelog entry.

The guard test that pins this behaviour lives on `main` (rewritten in #3423 to fail if any Dockerfile pulls a prebuilt grype binary). It is deliberately not backported, per the narrowing above; the release-preflight and publish `Security Scan` are what enforce it on this branch.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Closes #3422


brandonrc:
Applying **`release-process: approved`**, with the reasoning recorded rather than left implicit.

**What failed:** "Verify commits trace back to main" — it indexes `git patch-id` over main's last 1500 commits and requires an exact match. This commit carries the correct `(cherry picked from commit b0507587...)` trailer, but the gate matches on the *diff*, not the trailer, and this backport was **deliberately narrowed**, so the patch-ids differ by construction.

**Why an unmodified cherry-pick is not available here.** Taking `b050758` whole would import two things that cannot work on this branch:

1. **`backend/src/config.rs`** — #3423 rewrote the Dockerfile guard test, which depends on `discover_dockerfiles` and the surrounding scan-all scaffolding introduced by #3413/#3415. Neither is on `release/1.7.x`, so the test would not compile.
2. **`CHANGELOG.md`** — the upstream hunk targets main's `## [Unreleased]` section. Applying it here either creates an `[Unreleased]` heading on a release branch or files the entry outside `## [1.7.5]`, where it actually ships. The entry has been placed under `[1.7.5]` instead, alongside the three advisories it unblocks.

Also excluded, and worth stating so the omissions are auditable: `.trivyignore` (this branch keeps its own suppression set — main carries `# RETIRED:` tombstones for the retired tokens, which is precisely what preflight check 1's superset rule accepts, #3309) and `scheduled-container-scan.yml` (scan-report prose from #3413).

**Net change on this branch:** three Dockerfiles, the publish workflow, one changelog entry. The security-relevant content — compiling grype from source on a go1.26.6-floored toolchain — is carried over intact and unmodified.

This is the same structural gap #3422 documents: the policy cannot express "port a subset of a main commit to a maintenance branch", and narrowing is the correct engineering choice for a patch release rather than a shortcut around it.

**What it unblocks:** `v1.7.5`, carrying GHSA-8jm4-4x6c-6787, GHSA-95fx-g94v-8jqg and GHSA-fv45-mwhh-q23r — all fixed, none released, every 1.7.4 deployment currently exposed.

PR #3430, opened 2026-08-17
chore(release): prepare 1.7.6 (v1.7.5 is a dead tag)

### LINKED ISSUE #3429: v1.7.5 tag blocked: backported comment-only edit tripped the scanner-adapter version-collision gate
## What happened

`v1.7.5` was tagged at `2a069e9`. Both tag workflows failed and **no release object was created**:

- **Docker Publish** → `Scanner Adapter Multi-Arch Manifest`:
  ```
  ghcr.io/artifact-keeper/artifact-keeper-scanner-adapter:1.2.3 already exists at
  04b6cba994ecc9f150775a2ddce8ee53db7bebff, but scanner-adapter sources changed.
  Bump docker/scanner-adapter/VERSION; exact version tags are never republished.
  ```
- **Release** → `Resolve candidate image digest`: refused to continue while docker-publish was red. Correct behaviour — it should not resolve a digest from a failed publish.

## Cause

The #3423 backport (#3424) carried a **comment-only** edit to `docker/Dockerfile.scanner-adapter`. The gate from #3340 treats any change under the adapter's source set as a source change, which is the right rule: it exists to stop an exact version tag being republished with different content.

Nothing was functionally wrong with the images. The branch publish on the same commit had already passed `Security Scan` — the grype source build (#3352) works. This is purely release mechanics.

## Fix

#3426 reverts the comment on `release/1.7.x`. The comment documents a Go patch floor that matters on `main`, where the adapter sits beside the grype source build; it has no effect on this branch and is not part of the security fix. Reverting beats bumping `VERSION`, which would republish an adapter image for a documentation change and consume a version number on a maintenance branch.

## Lesson worth keeping

**When backporting to a release branch, restrict the cherry-pick to files that are functionally required.** A comment carried along for tidiness is not free: it can invalidate a published, immutable version tag. The narrowing in #3424 excluded four files deliberately and still kept this one without weighing it against the publish gate.

Suggested follow-up (separate from this fix): have the backport checklist in `RELEASING.md` call out the adapter/component source sets explicitly, so "does this touch a versioned component's source set?" is asked before the tag rather than discovered by a failed publish.

## Consequence for the tag

`v1.7.5` points at a commit whose publish fails, so the tag must be moved to the fixed tip once #3426 lands. It produced no release object and no assets, so nothing can have consumed it. If moving it proves problematic, the fallback is to cut `v1.7.6` and leave `v1.7.5` as a dead tag, documented as such.

Blocks: the release of GHSA-8jm4-4x6c-6787, GHSA-95fx-g94v-8jqg and GHSA-fv45-mwhh-q23r, all fixed and unreleased.

brandonrc:
## Summary

Bumps `release/1.7.x` to **1.7.6** so the security content intended for 1.7.5 can actually ship.

**Why the version number changed.** `v1.7.5` was tagged at `2a069e9`, its Docker Publish failed on the scanner-adapter version-collision gate, and the Release workflow correctly refused to resolve a digest from a failed publish — so **no images and no release object were ever produced**. Full detail in #3429.

The tag cannot be moved or reused:

```
remote: - Cannot delete this tag
! [remote rejected] v1.7.5 (push declined due to repository rule violations)
```

A repository ruleset forbids deleting or force-moving tags — which is the correct policy, and the reason the version is burned rather than recycled. Reusing a tag that consumers may already have fetched is exactly what that rule exists to prevent, even though in this case nothing consumed it.

**Nothing about the content changes.** 1.7.6 carries precisely what 1.7.5 was meant to:

- GHSA-8jm4-4x6c-6787 — client-controlled `X-Forwarded-For` prefix bypassing IP rate limiting
- GHSA-95fx-g94v-8jqg — backup restore inserting outside `ALLOWED_EXPORT_TABLES`
- GHSA-fv45-mwhh-q23r — percent-encoded repository key bypassing private-repository visibility
- the grype source build (#3352) that unblocked image publication on this branch

The dead tag is called out in **both** the CHANGELOG entry and `.github/release-notes/1.7.6.md`, so an operator who finds `v1.7.5` in the tag list learns immediately not to use it.

Changes: version in `Cargo.toml` / `Cargo.lock` / `openapi.rs`, the CHANGELOG heading plus a dead-tag note, and the release-notes file renamed.

## Test Checklist
- [x] Manually tested locally

Version-only change; the adapter source set is deliberately untouched, so the collision gate that killed 1.7.5 cannot fire again here.

## API Changes
- [x] N/A - no API changes

Closes #3429

### DIFF
--- a/.github/release-notes/1.7.6.md
+++ b/.github/release-notes/1.7.6.md
@@ -1,8 +1,18 @@
-# Artifact Keeper 1.7.5
+# Artifact Keeper 1.7.6
 
 A security release: **three advisories**, plus fixes to PyPI PEP 658 metadata handling. No schema migrations.
 
 Most deployments can take this without changing anything. Two situations need a look first — both are below.
+
+## Why this is 1.7.6 and not 1.7.5
+
+`v1.7.5` was tagged and is a **dead tag**: do not use it. Its Docker Publish
+failed because a comment-only edit backported alongside the grype fix tripped
+the scanner-adapter version-collision gate, so no images and no release object
+were ever produced for it. Repository rulesets forbid deleting or moving a tag,
+so the version was burned rather than reused. Full detail in #3429.
+
+The content intended for 1.7.5 ships here, unchanged.
 
 ## Advisories
 
@@ -56,7 +66,7 @@ So this narrows which tables an archive can reach; it does **not** make a hostil
 
 See the [CHANGELOG](CHANGELOG.md) for the mechanism behind each entry, the reasoning, and the exact scope of every fix.
 
-**Full Changelog**: https://github.com/artifact-keeper/artifact-keeper/compare/v1.7.4...v1.7.5
+**Full Changelog**: https://github.com/artifact-keeper/artifact-keeper/compare/v1.7.4...v1.7.6
 
 ---
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [1.7.5] - 2026-08-14
+## [1.7.6] - 2026-08-17
+
+> **`v1.7.5` is a dead tag — do not use it.** It was tagged, its Docker Publish
+> failed on a scanner-adapter version-collision gate, and no images or release
+> object were ever produced. Repository rulesets forbid deleting or moving a
+> tag, so the version number was burned rather than reused. Everything intended
+> for 1.7.5 ships in 1.7.6 unchanged. See #3429.
 
 ### Security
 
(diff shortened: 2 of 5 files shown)


brandonrc:
Applying **`release-process: approved`**.

"Verify commits trace back to main" cannot pass for a release-preparation commit by construction: bumping `release/1.7.x` to 1.7.6 (`Cargo.toml`, `Cargo.lock`, `openapi.rs`, the CHANGELOG heading, the release-notes filename) describes a version that exists only on this maintenance branch. There is no corresponding commit on `main` to trace back to, and there should not be — `main` is on 1.8.0-track.

The prior release-prep commit on this branch (#3379, "prepare 1.7.5") is in the same position for the same reason.

This is the fourth use of the escape hatch in this release sequence, all for the one structural gap tracked in #3422: the policy has no way to express a maintenance-branch-only change. Worth fixing rather than continuing to wave through — the obvious candidate is exempting `chore(release):` prep commits and narrowed backports carrying a `(cherry picked from …)` trailer, so the gate keeps its teeth for genuine drift.

Content is a version bump plus documentation of why 1.7.5 is dead (#3429). The adapter source set is deliberately untouched, so the collision gate that killed 1.7.5 cannot fire on this tag.

PR #3435, opened 2026-08-17
feat(docker): make backend and openscap images arbitrary-UID compatible

### LINKED ISSUE #3434: make backend and openscap images arbitrary-UID compatible for OpenShift restricted-v2
Part of the OpenShift compatibility effort (collaboration with Red Hat).

OpenShift's default `restricted-v2` SCC ignores the image's `USER` and runs the container as a random high UID that is a member of GID 0. The backend and openscap images already run as numeric non-root UID 1001 with GID 0 as primary group, but their runtime-writable directories are owned `1001:0` at the default mode 0755, which denies group write. An arbitrary UID in GID 0 therefore gets EACCES on `/data`, the Grype/Trivy caches, and the scan workspace.

## Scope
- `docker/Dockerfile.backend`: make writable dirs and the seeded Grype DB group-writable (`g=rwX`).
- `docker/Dockerfile.openscap`: same for `/home/oscap` and `/scan-workspace`.
- Add a regression guard that scans every `docker/Dockerfile*` and asserts the arbitrary-UID contract (numeric non-root USER, GID 0, group-writable dirs, UBI base) for cluster-deployable images.

## Out of scope (later phases)
- UBI conversion of the Alpine backend and scanner-adapter images (Phase 2).
- Container certification / Operator packaging.

brandonrc:
## Summary

OpenShift's default `restricted-v2` SCC ignores the image's `USER` and runs the container as a random high UID that belongs to GID 0. The backend and openscap images already run as numeric non-root UID 1001 with GID 0 as their primary group, but their runtime-writable directories were owned `1001:0` at the default mode 0755, which denies group write. A random UID in GID 0 therefore hit EACCES on `/data`, the Grype/Trivy caches, and the scan workspace.

For the backend this was a **boot** failure, not a degraded-runtime one: `PLUGINS_DIR` defaults to `./plugins` relative to `WORKDIR /app`, and `ensure_plugins_dir` `create_dir_all`s it at startup, so `/app` at 0755 meant CrashLoopBackOff before one request was served. `/app` is the load-bearing entry in the chmod list, not an incidental one.

- `Dockerfile.backend`: `chmod -R g=rwX` on the writable dirs, then `find -type d -exec chmod g+s` so a PVC stays group-owned by GID 0 after OpenShift reallocates the namespace a different UID. `HOME`, `XDG_CACHE_HOME` and `TRIVY_CACHE_DIR` are set explicitly (an arbitrary UID has no passwd entry, so `HOME` otherwise arrives as `/`). The seeded Grype DB is copied `--chmod=0550`, read-only: the same file disables auto-update and says the DB is refreshed by rebuilding the image.
- `Dockerfile.openscap`: the same `g=rwX` + directories-only `g+s` on `/home/oscap` and `/scan-workspace`, `HOME`/`XDG_CACHE_HOME`, and the setuid/setgid strip `Dockerfile.backend` already had (10 binaries).
- `grype_scanner.rs`: inject `GRYPE_DB_CACHE_DIR` onto the grype subprocess, joining the three DB-related variables already pinned there for exactly this reason. It was image-`ENV`-only, so a deployment whose container `env:` replaces rather than appends sent grype to `$HOME/.cache/grype` and failed the scan outright.
- `config.rs`: `openshift_runtime_images_are_arbitrary_uid_compatible` classifies every `docker/Dockerfile*` as `OPENSHIFT_RUNTIME` (asserted) or `EXCLUDED` with a reason, and fails on a new one left unclassified. For the asserted images it compares the *set* of directories created against the set made group-writable and setgid, reads the base image off the runtime stage's `FROM` rather than matching any stage, cross-references the passwd entry's UID against the `USER` directive, and asserts `/usr/local/bin`, `/`, `/etc`, `/usr`, `/bin` and `/licenses` are **not** reachable by the chmod.

## Scope: two of the stack's three runtime images

`Dockerfile.backend.alpine` and `Dockerfile.scanner-adapter` are built on Alpine with `adduser -u 1001`, which gives primary **GID 1001, not 0**, and their trees are chowned `1001:1001`. They are **not** arbitrary-UID compatible, and that is a functional gap rather than a packaging preference.

The scanner-adapter matters most: it is a published, deployed cluster workload and is how the backend performs every container-image scan (`TRIVY_ADAPTER_URL`). On OpenShift under `restricted-v2` after this PR, the backend and openscap pods start — which they could not before — while the scanner-adapter pod gets EACCES writing trivy's DB to `SCANNER_TRIVY_CACHE_DIR` and fails its readiness probe.

**So the stack is not yet OpenShift-deployable end to end.** Converting those two images (UBI base with a GID-0 user, or at minimum GID 0 plus group-writable dirs) is Phase 2 on the same issue, and the `EXCLUDED` reasons in `config.rs` now say so in those terms.

Part of the OpenShift compatibility effort (collaboration with Red Hat).

Closes #3434

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

<sub>Verified locally: `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo nextest run --workspace --lib` 16211 passed / 0 failed. Both images built and run as `--user 1000730000:0`; evidence in the audit-response comment.</sub>


brandonrc:
## Audit round 1 — arbitrary-UID compatibility (2 reviewers: container/OpenShift + security)

Both passes went beyond inspection — the container reviewer built and ran replicas of these images. Summary of the empirical work, since it's what the verdict rests on:

- **Permission replica built** (stage-for-stage: same `mkdir -p` set → `chown -R 1001:0` → `chmod -R g=rwX` → `COPY --from` → the grype-DB `COPY --chown=1001:0 --chmod=0770`). `/data`, `/data/storage`, `/data/backups`, `/scan-workspace`, `/app`, `/shared`, `/home/artifact`, `/home/artifact/.cache/grype` all land **0775 uid1001:gid0**, files 0664, DB files 0770. Modes survive `COPY --from` intact — the load-bearing assumption of the design.
- **Run as `--user 1000670000:0`** (the shape restricted-v2 produces): every path the app writes is `WRITABLE`; `/` correctly `DENIED`; `HOME=[/]`, `USER=[]`.
- **Real UBI chain built** (`ubi9/ubi:9.8` → `dnf --installroot` → `ubi-micro:9.8`) to check `/tmp`, which neither Dockerfile chmods: **1777** in the rootfs, the base, and the final image. The three code paths writing there are safe.
- **`g=rwX` semantics verified** on coreutils 9.4: dirs 0755→0775, files 0644→0664, exec bit never added to a plain file, **nothing world-writable**, nothing setuid/setgid introduced.
- **No `VOLUME`** in either Dockerfile, so no anonymous-volume masking. `USER` stays numeric non-root in both.

**So: for the two images it targets, the fix is real and it works.** The findings are that it's one Helm `env:` block from failing, the guard can't detect its own regression, and the stack still isn't OpenShift-deployable.

---

### 1. 🔴 `GRYPE_DB_CACHE_DIR` is the one grype env var not re-injected at the subprocess level — and it's the one this PR depends on

`grype_scanner.rs:1356-1358` pins three env vars onto the grype child (`GRYPE_DB_AUTO_UPDATE`, `GRYPE_DB_VALIDATE_AGE`, `GRYPE_CHECK_FOR_APP_UPDATE`). The doc comment right above at `:1444-1449` says why, in as many words: *"The Dockerfile also sets these vars, but injecting them here means the scanner keeps working under deployment configs that wipe inherited env (Helm charts, k8s `env:` blocks that replace rather than append)."* There's even a source-scanning test pinning all three at `:4296-4313`.

`GRYPE_DB_CACHE_DIR` gets neither. It exists **only** as image ENV at `Dockerfile.backend:353`.

**Failure:** an OpenShift Deployment with a `containers[].env:` block — which every real deployment has, since `DATABASE_URL` and `JWT_SECRET` are both mandatory (`config.rs:1113`, `1129`). If that block replaces rather than appends, the var is gone. Grype falls back to `$XDG_CACHE_HOME/grype/db` → `$HOME/.cache/grype/db`; `XDG_CACHE_HOME` is set nowhere and `HOME=/` for the arbitrary UID (verified: `mkdir /.cache` → Permission denied as UID 1000670000). Grype gets EACCES, never reads the seeded DB at `/home/artifact/.cache/grype` that this PR just spent a `--chmod=0770` making writable, and per the CAUTION comment at `grype_scanner.rs:1334-1340` **exits 1 with empty stdout — the whole scan fails, findings included.**

We already learned this lesson for three sibling variables and wrote it down. The fourth didn't get carried across. One line in `run_grype_with_catalog`, plus extending the `:4296` test.

### 2. 🔴 The anti-drift test cannot detect the regression it exists to prevent

`config.rs:4869` is `assert!(content.contains("g=rwX"))` — a whole-file substring search asserting the *string* appears, not that any particular directory is chmod'ed.

Proven, not asserted. Mutation applied to `Dockerfile.backend` at PR head: collapse the chmod list from the five `/mnt/rootfs/...` paths down to `/mnt/rootfs/shared` alone, **and** add a new writable dir to `mkdir -p` with no chmod. All four assertions still pass:

```
content.contains(':1001:0:')                        -> True
content.contains('g=rwX')                           -> True
content.contains('registry.access.redhat.com/ubi9') -> True
final USER -> 1001 | numeric & non-zero             -> True
```

**Failure:** someone refactors that RUN and drops `/mnt/rootfs/app`. CI green, image ships, and on OpenShift the backend fails **during startup** — `PLUGINS_DIR` defaults to `./plugins` (`main.rs:284`), relative to `WORKDIR /app`, and `ensure_plugins_dir` `create_dir_all`s it at boot (`main.rs:1385` → `wasm_plugin_service.rs:1975`). `/app` at 0755 → `os error 13` → CrashLoopBackOff before one request is served. Drop `/mnt/rootfs/data` instead and you get the same error later, on the first upload staging into `<STORAGE_PATH>/.incoming` (`incus.rs:337-344`).

Note this also corrects the PR description: `/app` isn't an incidental entry in that list, it's **load-bearing for boot**. Pre-PR it was `1001:0` mode 0755, so an arbitrary UID could not create `/app/plugins` — the backend failed at startup. This chmod is what fixes that.

The other two content assertions are equally weak:
- `content.contains(":1001:0:")` (`:4862`) matches anywhere including comments, and never cross-checks the passwd UID against the `USER` directive — `USER 1002` plus a `1001:0:` passwd line passes.
- `content.contains("registry.access.redhat.com/ubi9")` (`:4876`) matches any stage. The file already has `golang:1.26-alpine` and `alpine:3.23` stages, so switching the **runtime** stage (`:300`) to Alpine while leaving the UBI builders passes cleanly.

What it does genuinely catch: `USER 1001` → `USER appuser` (the `is_ascii_digit` check at `:4857`) and an unclassified new Dockerfile (`:4830-4843`). Roughly a third of the stated contract.

**Fix:** parse the runtime stage only (last `FROM`→EOF), extract the path set from `mkdir -p` and from `chmod -R g=rwX`, assert subset. ~20 lines against helpers already in the module. Also worth asserting `/usr/local/bin`, `/etc` and `/licenses` are *not* in the chmod set, so the test bounds the blast radius rather than merely mandating the mechanism.

### 3. 🔴 `Dockerfile.scanner-adapter` is a deployed cluster workload, excluded as if it weren't

`config.rs:4815-4818` excludes it as *"non-UBI Alpine variant; Phase 2 UBI + GID 0 conversion pending."* Factually true, but it frames a functional blocker as a packaging preference.

`Dockerfile.scanner-adapter:104-135`: `FROM alpine:3.24`, `adduser -D -u 1001 scanner` — primary **GID 1001, not 0** — `chown -R scanner:scanner /home/scanner`, `ENV SCANNER_TRIVY_CACHE_DIR=/home/scanner/.cache/trivy` at 0755, `USER 1001`. It's published as a first-class image (`docker-publish.yml:439-442`) and is how the backend performs every container-image scan (`TRIVY_ADAPTER_URL`, `config.rs:1146`).

**Failure:** deploy the stack on OpenShift under restricted-v2. The backend pod now starts — this PR's win. The scanner-adapter pod runs as UID 1000670000/GID 0, trivy writes its DB to `/home/scanner/.cache/trivy` owned `1001:1001` mode 0755 → EACCES → every image scan errors and the pod restarts on its readiness probe. Same for `Dockerfile.backend.alpine:143-148` (`chown -R 1001:1001`) for anyone selecting that variant.

The PR title is honest about scope; the release note should be too. Two of the stack's three runtime images are arbitrary-UID compatible, so the stack is not yet OpenShift-deployable. (Related, from the #2883 review: `Dockerfile.backend.alpine:104` still does `chown -R 1001:1001` on main, on this PR, and on #2883 — worth deciding explicitly whether the alpine half is in scope here or a stated Phase 2.)

### 4. 🟠 No setgid on the writable directories — a UID change makes the volume unwritable

`Dockerfile.backend:119-121` and `Dockerfile.openscap:49` use `g=rwX`, which yields 0775/0664 with **no `s` bit** (confirmed: `775`, not `2775`). Runtime-created files take the process umask (0022 from the runtime; the STIG `UMASK 077` written to `/etc/login.defs` at `:142-144` is inert for a non-login container process) and land 0644.

**Failure:** a PVC bound to `/data` is populated by UID 1000670000. The namespace is later recreated, or the app is promoted dev→prod across an RWX volume, and OpenShift allocates UID 1000850000. Every artifact under `/data/storage` is `1000670000:0` mode 0644 — readable, **not writable**. Blob overwrite, storage GC and `.incoming` cleanup all fail with `os error 13`.

Red Hat's own container guidelines specify `chmod -R g=u` **plus** `g+s` for exactly this. One character: `g=rwXs`.

### 5. 🟠 `HOME` is never set; trivy is spawned with no env; the trivy cache dir this PR chmods is unreachable by design

No `ENV HOME` in either Dockerfile (the only `HOME` reference is the build-stage `$HOME/.cargo/env` at `:19`). Verified at runtime: `HOME=[/]`, `USER=[]`, `$HOME/.cache` not creatable.

`Command::new("trivy")` at `trivy_fs_scanner.rs:125` and `incus_scanner.rs:185`, plus the version probe at `scanner_service.rs:3499`, carry **no `.env(...)` at all**, and `TRIVY_CACHE_DIR` appears nowhere in the repo. Trivy defaults to `$HOME/.cache/trivy` → `/.cache/trivy` → EACCES.

Meanwhile `Dockerfile.backend:114` creates `/home/artifact/.cache/trivy` and this PR makes it group-writable. Nothing can reach it: no env names it, and `HOME` never resolves there. **The PR made a directory writable for a consumer that cannot find it.**

Latent today — `:174-179` deliberately omits the trivy binary and the spawn fails cleanly via `classify_trivy_spawn_error`. It goes live the moment anyone re-bundles trivy or builds a derived image, and the pre-created cache dir reads as an invitation to do exactly that. `ENV HOME=/home/artifact XDG_CACHE_HOME=/home/artifact/.cache` plus `TRIVY_CACHE_DIR` costs nothing and is standard OpenShift practice.

(Checked rather than assumed: **zero** non-test uses of `HOME`, `dirs::home_dir()` or `std::env::temp_dir()` in `backend/src`.)

### 6. 🟡 `COPY --chmod=0770` is correct only because the destination happens to pre-exist

`Dockerfile.backend:318`. Both cases tested:
- **Destination pre-created by the rootfs COPY (the real case):** `--chmod` applies to copied *contents only*; the directory keeps the 0775 it got from `chmod -R g=rwX`. Works.
- **Destination absent:** BuildKit auto-creates intermediates at **0750** with `--chown` applied and **ignores `--chmod` for them**. Verified: `750 1001:0 .../cache/grype`, `770 .../grype/5`.

**Failure:** someone drops `/mnt/rootfs/home/artifact/.cache/grype` from the `mkdir -p` at `:113`, or reorders the grype COPY above the rootfs COPY at `:303`. The parent silently becomes 0750 group-read-only, grype's DB refresh gets EACCES, and finding 2's test stays green.

Two sub-points: the mode also makes data files executable (`rwxrwx---` on `vulnerability.db`, `metadata.json`) — `0660` is correct given the dirs are already 0775, and the current mode is the kind of thing a Red Hat container-certification `file_permissions` rule picks at. Second, the security pass notes the stated justification is **contradicted by the same file**: `:316-317` says the chmod is needed "so an arbitrary OpenShift UID can refresh the DB at runtime", but `:354-355` sets `GRYPE_DB_AUTO_UPDATE=false`/`GRYPE_DB_VALIDATE_AGE=false` and `:339-348` says outright the DB is refreshed by rebuilding the image. Re-asserted per invocation at `grype_scanner.rs:1356`. The DB is never written at runtime — `0550` would do, and the comment should be corrected either way.

### 7. 🟡 Layer ordering is correct — but verify it stays that way

Walked both stages. In `rootfs-builder`: `dnf install` (`:85`) → mkdir/chown/chmod (`:105-121`) → STIG hardening (`:127-171`). Every STIG step touches only `/etc`, `/var`, `/usr` — nothing lands in the chmod'd trees afterward at default perms. In `runtime`: rootfs COPY (`:303`) first, then `/usr/local/bin` (`:307`), `/licenses` (`:313`), grype DB with its own `--chmod` (`:318`), app binary (`:321`). No COPY lands in a writable directory without its own mode, and the rootfs COPY precedes the grype DB so it can't overwrite it. Right today; finding 6 is how fragile that is.

**`COPY --chmod` support: confirmed on every build path.** `setup-buildx-action` + `build-push-action@v7` at `docker-publish.yml:210/257`, `332/357`, `414/439`, `499/516`; also `ci.yml:1455` and `e2e.yml:122`. Compose builds are BuildKit-only under Compose v2. The only bare `docker build` in the repo is `scripts/native-tests/test-docker.sh:106`, which builds a generated `$WORK_DIR`, not this file. No `DOCKER_BUILDKIT=0` anywhere. Nothing breaks.

Minor: `# syntax=docker/dockerfile:1` is on **line 2** of `Dockerfile.backend`, behind a banner comment on line 1, so the parser directive is inert and degrades to an ordinary comment. Pre-existing and harmless in practice (BuildKit's built-in frontend has supported `COPY --chmod` for years), but it means the frontend this PR's `--chmod` relies on is whatever the builder ships, not the pinned `dockerfile:1`. Moving it to line 1 makes the pin apply.

### 8. 💭 Answers to the rest, and things worth knowing

**Test is wired into CI — yes, no #3620-style gap.** `ci.yml:521` runs `cargo nextest run --workspace --lib` in `test-backend-unit`, gated on `changes.outputs.code != 'false'`; a Dockerfile-only PR falls into the `*)` arm of the path filter at `:107-116` and sets `code=true`, so the guard runs even on a `docker/`-only change. `config.rs` is reachable via `lib.rs:11`, and the coverage job at `:746` runs the same command.

**Test can't silently pass on a missing directory.** `config.rs:4802` resolves `CARGO_MANIFEST_DIR/..`; `discover_dockerfiles` (`:4489-4491`) uses `read_dir(...).expect(...)`, so a missing `docker/` panics rather than yielding an empty vec. Correct design. Side effect: the lib suite is un-runnable from a `cargo package` tarball — pre-existing, shared with three other guards using the same helper.

**EXCLUDED list is complete and arithmetically closed.** Six `docker/Dockerfile*` files; 2 asserted + 4 excluded = 6, each with a stated reason, and `:4836-4843` fails on a seventh. The `.dev` and `.redteam` reasons are honest (`Dockerfile.redteam:49` is genuinely `USER root`). Scope note: `discover_dockerfiles` walks only `docker/`, so `.assets/docker/Dockerfile` and the two `scripts/*/mock-upstream` Dockerfiles escape entirely — none are cluster workloads, so that's a scope statement, not a gap.

**Non-OpenShift path: no meaningful regression.** Nothing world-writable; a plain-Docker `USER 1001` already has primary GID 0 (`:105`, `artifact:x:1001:0:`), so this widens access to nobody who didn't have it. CIS Docker 4.x and Trivy misconfig don't flag group-writable-by-root-group. Note `docker-compose.yml:155-175` already ships a root `scan-workspace-init` sidecar chowning to `1001:0` + `chmod 0775`; now partly redundant for fresh volumes, still needed for the Podman/SELinux case its comment describes.

**Nothing in CI would have flagged any of this.** No hadolint, dockle, checkov or conftest anywhere in `.github/`, `scripts/`, or `Makefile` (zero hits). `scheduled-container-scan.yml` and the `docker-publish.yml` gate both run trivy with **no `scanners: misconfig`** — default is vuln + secret only, so group-writable directories are never reported. So the PR broke no expectation; but nothing bounds these permissions either, which is finding 2's real stakes.

**Untrusted-content handling around `/scan-workspace`: clean, checked.** Symlink/hardlink/special-file entries are rejected in the tar path (`scanner_service.rs:1732`) and skipped in the zip path (`:1855`); `..`/absolute entries rejected via `enclosed_name` (`:1700`, `:1812`), with regression tests at `:13662`/`:13737`. Workspaces are namespaced per artifact UUID (`:925-931`) and `cleanup`/`cleanup_path` (`:1229`, `:1237`) run after each scan. Nothing reads config from `/home/oscap`. Also worth knowing: `/usr/local/bin` is deliberately in neither the chown nor the chmod list, which matters — `python3 /usr/local/bin/openscap-wrapper.py` puts that dir at `sys.path[0]`, so had it been group-writable any GID-0 process could shadow `json`/`re`/`subprocess` and own the scanner. It isn't.

**Latent hazard for later:** `g=rwX` over a tree containing a setuid binary would make it group-writable — trivial root escalation. Zero impact today (both trees are empty at chmod time), and `Dockerfile.backend:167` strips setuid/setgid afterward as a backstop. **`Dockerfile.openscap` has no such strip step**, so only emptiness protects it. Worth adding `find -xdev -perm /6000 -exec chmod a-s {} +` there for symmetry.

**Two of the five chmod'd directories are dead** (pre-existing, out of scope, but relevant while you're in here): `Dockerfile.backend:110` creates `/data/plugins` but nothing points `PLUGINS_DIR` at it, so uploaded WASM plugins land on the ephemeral layer at `/app/plugins` and are lost on restart rather than persisting to the `/data` PVC. And `BACKUP_PATH=/data/backups` (`:352`) is read nowhere in `backend/src` — backups go through the storage backend under key `backups/YYYY/MM/DD/<name>` (`backup_service.rs:629-631`). Separately, `artifactory_import.rs:164-196` `.keep()`s its `tempfile::tempdir()` and never removes it, so every Artifactory ZIP import leaks an extracted tree into `/tmp` — on Kubernetes that's an ephemeral-storage eviction, not a permissions error.

---

**Bottom line:** `chmod -R g=rwX` is the right mechanism — it's Red Hat's documented pattern, and there's no better in-image alternative (`fsGroup` is a deployment-side control an image can't assume; setgid dirs solve GID inheritance for new files, not base-directory write permission). The path list is right except that it's worth dropping nothing and *adding* the `s` bit.

To actually call this "runs under restricted-v2": findings 1, 2, 3. Then 4 and 5 as cheap durability. Worth considering alongside: `readOnlyRootFilesystem: true` with `emptyDir` for `/scan-workspace` and `/tmp`, plus `fsGroup: 0` and `runAsNonRoot: true` in the chart.

<sub>Round 1 of an automated multi-reviewer audit (container/OpenShift + security passes). The container pass built and ran permission replicas, the real UBI base chain, grype under a broken `HOME`, and a mutation test against the guard; artifacts kept locally.</sub>

brandonrc:
## Audit round 1 — dispositions, with build/run evidence

Branch merged up from `main` (it was 3 weeks behind; the merge was clean, including today's `GRPC_TO v1.83.1 -> v1.83.2` bump in `Dockerfile.backend{,.alpine}` from #3792, which is preserved untouched). `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo nextest run --workspace --lib` **16211 passed / 0 failed**.

Everything below was run against **images built from this branch and from `main`**, not reasoned about. One disclosure on method: the runtime stage `COPY --from=builder`s the Rust binary, so to keep the loop short I built both images with `--build-context builder=<dir>` supplying a small static probe in place of `artifact-keeper`. Every other stage is the real one — `rootfs-builder` (the entire permission story), the UBI micro base, the grype source build, the DB seed, and the real runtime stage. The probe's job is only to `stat` and write; the binary's *content* is irrelevant to file modes, and the grype runs below use the real grype binary out of the real image.

| # | | Disposition |
|---|---|---|
| 1 | `GRYPE_DB_CACHE_DIR` not re-injected | **Fixed** — and it was worse than described |
| 2 | Guard can't detect its own regression | **Fixed** — your mutation now fails it |
| 3 | scanner-adapter excluded as if it weren't a workload | **Framing fixed**, conversion stays Phase 2 |
| 4 | No setgid | **Fixed, but not with `g=rwXs`** — see below |
| 5 | `HOME` never set / trivy cache unreachable | **Fixed** |
| 6 | `COPY --chmod=0770` correct only by accident | **Fixed** — 0550, not 0660 |
| 7 | Layer ordering; `# syntax` on line 2 | Verified; directive moved to line 1 |
| 8 | Answers + latent hazards | openscap setuid strip added; the rest triaged below |

---

### 1 🔴 `GRYPE_DB_CACHE_DIR` — fixed, and the fallback you'd expect doesn't work

Fixed in `run_grype_with_catalog`. It can't be pinned to a literal the way the other three are — outside the image (`cargo run`, CI, a dev laptop) grype's own `$HOME/.cache/grype` default is the right answer and forcing the image path would break it — so it goes through `resolve_grype_db_cache_dir`: an explicit non-empty value passes through, otherwise the seeded directory is named if it exists on disk, otherwise nothing is set. Pure function, six unit tests, plus a guard reading the path back off `Dockerfile.backend` so the constant can't drift from the image.

The empirical part is the bit worth reading. Four runs, real grype binary, real seeded DB, `--user 1000730000:0`:

```
$ docker run --rm --user 1000730000:0 [-e GRYPE_DB_CACHE_DIR=] \
    --entrypoint /usr/local/bin/grype ak3435-backend:<tag> dir:/licenses -o json

[main | image ENV intact]         exit=1
    ERROR failed to load vulnerability db: failed to access database file:
          stat /home/artifact/.cache/grype/6/vulnerability.db: permission denied
[main | GRYPE_DB_CACHE_DIR wiped] exit=1
    ERROR failed to load vulnerability db: database does not exist
[fix  | image ENV intact]         exit=0
[fix  | GRYPE_DB_CACHE_DIR wiped] exit=1
    ERROR failed to load vulnerability db: database does not exist
```

Two things fall out of that table.

First, on `main` grype fails as an arbitrary UID **even with the variable intact** — the seeded DB was not merely un-writable, it was un-*readable*. So the scan-fails-with-empty-stdout outcome you described for the wiped-env case was already happening on every OpenShift scan, unconditionally. This PR's chmod is what turns row 3 green.

Second, and this is why the Rust-side injection is load-bearing rather than belt-and-braces: **row 4 is still red on the fixed image even though `XDG_CACHE_HOME` is now set.** Finding 5's `ENV XDG_CACHE_HOME=/home/artifact/.cache` does not rescue finding 1, because grype's XDG fallback is `$XDG_CACHE_HOME/grype/db` while the seeded DB lives at `$GRYPE_DB_CACHE_DIR/6/`. Different path. Naming the directory on the child is the only thing that fixes row 4, which is exactly what the code now does.

### 2 🔴 Guard rewritten — your mutation fails it

Rewritten as you specified, plus the two weaker assertions. It now locates the provisioning `RUN` by its `chown -R 1001:0` signature, joins the backslash continuations, extracts the path sets from `mkdir -p` / `chown -R 1001:0` / `chmod -R g=rwX` / the setgid `find`, and asserts subset relations with `-R` prefix semantics. The base check reads the **last** `FROM`, so the golang/alpine builder stages no longer satisfy it. The passwd check is `format!(":x:{user}:0:")` against the parsed `USER`, so `USER 1002` with a `1001:0:` line fails.

Your mutation, applied verbatim to the real `Dockerfile.backend` at this branch's head:

```
--- old guard's four assertions against the mutated file ---
  content.contains(':1001:0:')                        -> True
  content.contains('g=rwX')                           -> True
  content.contains('registry.access.redhat.com/ubi9') -> True
  final USER -> 1001 | numeric & non-zero             -> True
--- new guard against the same mutated file ---
  FAIL config::tests::openshift_runtime_images_are_arbitrary_uid_compatible
  Dockerfile.backend creates /app but no `chmod -R g=rwX` reaches it
  (group-writable roots: ["/shared"]). Under restricted-v2 it is mode 0755 and
  the arbitrary UID gets EACCES on it. Either add it to the chmod list or, if
  it must not be writable, add it to NEVER_GROUP_WRITABLE here
```

Blast radius is bounded as you suggested: `/usr/local/bin` is asserted to stay **out** of the group-writable set (with your `sys.path[0]` reasoning written into the failure message), and `/`, `/etc`, `/usr`, `/usr/bin`, `/bin`, `/licenses` are asserted unreachable by the chmod. A second test, `arbitrary_uid_guard_helpers_catch_the_mutations_the_substring_version_missed`, pins all three mutations as fixtures so the guard's own regression is caught in-suite rather than by the next auditor.

I took your correction about `/app` into the PR body: it is now stated as the boot-critical entry, with the `ensure_plugins_dir` path, rather than one item in a list.

### 3 🔴 Framing fixed; the conversion stays Phase 2

You're right that the reason was true and misleading in the same breath. Both reasons are rewritten to state what is functionally wrong — GID 1001 rather than 0, trees chowned `1001:1001` — and the scanner-adapter's entry now says it is a published cluster workload reached via `TRIVY_ADAPTER_URL`, that trivy gets EACCES on `SCANNER_TRIVY_CACHE_DIR` under restricted-v2, and, explicitly: *"with this image excluded the STACK is not yet OpenShift-deployable end to end, only the backend and openscap pods are."* The PR body carries the same statement under its own heading, and so does the CHANGELOG bullet.

I did **not** convert the two Alpine images here. Not because it's hard — `-G root` plus `chown -R 1001:0` and the same chmod would do it without a UBI rebase — but because the scanner-adapter ships a binary from a separate source repo under a pinned digest, and changing its runtime identity belongs in a PR that can build and gate that image on its own rather than riding along in one whose title says "backend and openscap". The honest statement of the gap is in three places now; the fix is Phase 2 on #3434.

### 4 🟠 setgid — accepted, but `g=rwXs` is the wrong character

Substance accepted, literal fix rejected, and I'd push back on this one. `chmod -R g=rwXs` sets the setgid bit on **plain files** as well. Measured on the real base:

```
$ docker run --rm registry.access.redhat.com/ubi9/ubi:9.8 bash -c '... chmod -R g=rwXs /t/dir ...'
--- before                     --- after g=rwXs
755 /t/dir                     2775 /t/dir
755 /t/dir/exec                2775 /t/dir/exec      <-- setgid executable
644 /t/dir/file                2664 /t/dir/file      <-- setgid data file
644 /t/dir/sub/f2              2664 /t/dir/sub/f2
```

A setgid executable owned by GID 0 is a group-privilege escalation primitive, and setgid data files are a `file_permissions` certification finding — the same class of hazard your own finding 8 flags. In `Dockerfile.backend` the later `find -perm /6000 -type f -exec chmod a-s` would have scrubbed the files again, which means the setgid bits would have been invisible in the shipped image and load-bearing on step ordering; in `Dockerfile.openscap` there was no such step at all and they'd have shipped.

So it's Red Hat's actual two-step form, directories only:

```dockerfile
find /mnt/rootfs/data /mnt/rootfs/scan-workspace \
     /mnt/rootfs/home/artifact /mnt/rootfs/app \
     /mnt/rootfs/shared -type d -exec chmod g+s {} +
```

The guard asserts the `-type d`, so `g=rwXs` cannot come back in a later refactor.

**On what setgid actually buys, measured** — because it is less than the finding claims. Named volume on `/scan-workspace`, populated by UID 1000730000, then reopened as 1000850000 (both GID 0), which is your namespace-recreated scenario:

```
===== main image =====
-- UID 1000730000 populates the volume --
  dir /scan-workspace mode=0o755 owner=1001:0
  PermissionError: [Errno 13] '/scan-workspace/artifact.blob'      <-- can't even populate it
===== fix image =====
-- UID 1000730000 populates the volume --
  dir /scan-workspace mode=0o2775 owner=1001:0
  created /scan-workspace/artifact.blob mode=0o644 owner=1000730000:0
-- namespace recreated: OpenShift allocates UID 1000850000 --
  existing /scan-workspace/artifact.blob mode=0o644 owner=1000730000:0
  IN-PLACE APPEND:     DENIED (Permission denied)
  REPLACE-VIA-RENAME:  ok
  NEW SUBDIR:          ok mode=0o2755 owner=1000850000:0
```

setgid fixes *group inheritance*, not the file mode: new files still land 0644 under the runtime's 0022 umask, so the second UID cannot write an existing file in place. It can replace it and it can create directories, which inherit GID 0 and the bit. That is enough for this application: `storage/filesystem.rs` writes every blob to a temp path and `fs::rename`s it into place (`:203`/`:214`, `:405`/`:439`) precisely because the old in-place `File::create` was non-atomic, and GC/`.incoming` cleanup are unlinks, which need directory write. So the three failures the finding predicts don't occur on the real write path — but there is no in-image fix for the umask half, and a chart wanting belt-and-braces should set `fsGroup: 0`.

### 5 🟠 `HOME` / `XDG_CACHE_HOME` / `TRIVY_CACHE_DIR` — fixed

`ENV HOME=/home/artifact XDG_CACHE_HOME=/home/artifact/.cache TRIVY_CACHE_DIR=/home/artifact/.cache/trivy` in the backend, `HOME=/home/oscap XDG_CACHE_HOME=/home/oscap/.cache` in openscap. Before/after under `--user 1000730000:0`:

```
main:  env HOME = [/]                  fix:  env HOME            = [/home/artifact]
       env XDG_CACHE_HOME = []               env XDG_CACHE_HOME  = [/home/artifact/.cache]
       env TRIVY_CACHE_DIR = []              env TRIVY_CACHE_DIR = [/home/artifact/.cache/trivy]
```

This closes the "made a directory writable for a consumer that cannot find it" gap: `TRIVY_CACHE_DIR` now names the directory the rootfs-builder creates, so a derived image that re-bundles trivy lands in the group-writable cache instead of `/.cache/trivy`. As noted under finding 1, it does **not** fix grype — that needed the subprocess injection.

### 6 🟡 grype DB mode — `0550`, and the comment corrected

The security pass is right that the justification contradicted the same file, so the mode changed rather than the comment. It is `--chmod=0550`, read-only.

Not `0660`, though — `--chmod` applies one mode to every copied entry including directories, and the DB lives under a schema subdirectory. Any mode without group-execute makes `6/` untraversable and the DB unreadable; `0550` is the least-privilege mode that still traverses. Shipped result, read by real grype as UID 1000730000 (this is the `exit=0` row from finding 1):

```
2775 artifact:root /home/artifact/.cache/grype
 550 artifact:root /home/artifact/.cache/grype/6
 550 artifact:root /home/artifact/.cache/grype/6/vulnerability.db
 550 artifact:root /home/artifact/.cache/grype/6/import.json
...  "path": "/home/artifact/.cache/grype/6/vulnerability.db", "valid": true
```

The comment now says the DB is never written at runtime and cites the two ENV knobs that make that true, and the fragility you identified — that this only works because the destination pre-exists from the rootfs COPY — is written down as a "do not drop this from the mkdir, do not reorder this COPY" note at the line itself. The guard's mkdir-vs-chmod set comparison catches the first half of that automatically.

### 7 🟡 Ordering verified; `# syntax` moved to line 1

Re-walked after the merge: still `dnf` → mkdir/chown/chmod/setgid → STIG (which touches only `/etc`, `/var`, `/usr`) in the builder, and rootfs COPY → `/usr/local/bin` → `/licenses` → grype DB → binary in the runtime. The `-type f` on the setuid strip is what keeps it from undoing the new directory setgid bits, and that's now stated in the comment above it.

`# syntax=docker/dockerfile:1` moved from line 2 to line 1, so the pin applies instead of degrading to a comment.

### 8 💭 Latent hazards

**openscap setuid strip added** (your symmetry suggestion). It was the one worth doing: openscap ships shadow-utils where the backend doesn't, so the `g=rwX` hazard there was protected only by the trees happening to be empty at chmod time. Measured:

```
-- MAIN openscap setuid/setgid files --      -- FIX openscap setuid/setgid files --
/usr/libexec/utempter/utempter               (none)
/usr/bin/write  /usr/bin/umount  /usr/bin/su
/usr/bin/chage  /usr/bin/mount   /usr/bin/newgrp
/usr/bin/gpasswd /usr/sbin/unix_chkpwd
/usr/sbin/pam_timestamp_check
```

Ten setuid/setgid binaries out of a scanner image that runs one Python process as a non-login user, none of which it can use.

**Not fixed, deliberately.** The dead `/data/plugins` directory and the unread `BACKUP_PATH`, and the `artifactory_import.rs` `.keep()`ed tempdir leak, are all pre-existing and none are permission bugs — the tempdir leak in particular is an ephemeral-storage eviction with its own failure mode and deserves its own PR and CHANGELOG line rather than being buried in a Dockerfile permissions change. Same for `readOnlyRootFilesystem` / `fsGroup: 0` / `runAsNonRoot: true` in the chart: deployment-side controls, correctly out of an image PR. All worth issues; say the word and I'll file them.

---

### Full runtime evidence, both images, before and after

```
$ docker run --rm --user 1000730000:0 ak3435-backend:{main,fix}

                                    main                  fix
uid=1000730000 gid=0 groups=[0]     is_root=no            is_root=no
/data                               DENIED  0755 1001:0   WRITABLE 2775 1001:0
/data/storage                       DENIED  0755 1001:0   WRITABLE 2775 1001:0
/data/backups                       DENIED  0755 1001:0   WRITABLE 2775 1001:0
/data/plugins                       DENIED  0755 1001:0   WRITABLE 2775 1001:0
/scan-workspace                     DENIED  0755 1001:0   WRITABLE 2775 1001:0
/shared                             DENIED  0755 1001:0   WRITABLE 2775 1001:0
/app                                DENIED  0755 1001:0   WRITABLE 2775 1001:0
/home/artifact                      DENIED  0755 1001:0   WRITABLE 2775 1001:0
/home/artifact/.cache/grype         DENIED  0755 1001:0   WRITABLE 2775 1001:0
/home/artifact/.cache/trivy         DENIED  0755 1001:0   WRITABLE 2775 1001:0
/tmp                                WRITABLE 1777 0:0     WRITABLE 1777 0:0
/                                   DENIED  0755 0:0      DENIED   0755 0:0     <-- correct
mkdir /app/plugins                  EACCES (CrashLoop)    ok  2755 1000730000:0
mkdir /data/storage/.incoming       EACCES                ok  2755 1000730000:0
mkdir /scan-workspace/<uuid>        EACCES                ok  2755 1000730000:0

$ docker run --rm --user 1000730000:0 --entrypoint /usr/bin/python3 ak3435-openscap:{main,fix}

/home/oscap                         DENIED  0755 1001:0   WRITABLE 2775 1001:0
/scan-workspace                     DENIED  0755 1001:0   WRITABLE 2775 1001:0
/usr/local/bin                      DENIED  0755 0:0      DENIED   0755 0:0     <-- correct
/                                   DENIED  0755 0:0      DENIED   0755 0:0     <-- correct
oscap --version                     (n/a)                 OpenSCAP 1.3.14 as uid 1000730000
```

`USER` stays numeric non-root in both, `getuid()` is the arbitrary UID with `groups=[0]` and nothing else, and neither image falls back to root anywhere.

### CHANGELOG

One bullet added under the current `## [Unreleased]` → `### Changed` (verified with the `awk '/^## \[/{h=$0} /^- \*\*/{print h}'` check — the branch predates the 1.9.0 cut on 2026-09-08, so this was the #3797 trap). It carries the scope limitation from finding 3 in the same terms as the PR body.

PR #3439, opened 2026-08-18
fix(auth): reconcile OIDC_ADMIN_GROUP from environment

### LINKED ISSUE #3420: OIDC_ADMIN_GROUP is documented but ignored by environment-based OIDC bootstrap
### Description

`OIDC_ADMIN_GROUP` is documented as an environment variable for OIDC
configuration, but the environment-based OIDC bootstrap/reconciliation path
does not consume it.

In a Kubernetes deployment, the backend container has
`OIDC_ADMIN_GROUP=ArtifactKeeperAdmins`, while the env-managed OIDC provider
retains a different admin-group value in its persisted configuration.

Other OIDC environment settings such as scopes, groups claim, redirect URI, and
group-reflection settings are reconciled into the provider, but
`OIDC_ADMIN_GROUP` is not.

This makes the documented environment variable ineffective for GitOps /
environment-managed OIDC providers.

### Environment Information

- Deployment type: Kubernetes / Helm / Argo CD
- Operating System: Linux container
- Architecture: amd64
- Database: PostgreSQL bundled with the Artifact Keeper Helm deployment
- Reverse Proxy: Traefik
- Network setup: AWS EKS; Artifact Keeper and Keycloak communicate over HTTPS; OIDC provider is Keycloak

### Backend Version

1.7.1

### Web UI Version

N/A

### Package Format

N/A / General

### Expected Behavior

When an environment-managed OIDC provider is configured with `OIDC_ADMIN_GROUP=ArtifactKeeperAdmins`, the environment bootstrap/reconciliation process should apply `ArtifactKeeperAdmins` as the provider's admin group.

If `OIDC_ADMIN_GROUP` is changed and the backend is restarted or redeployed, the persisted env-managed OIDC provider should be reconciled accordingly, consistent with the other supported `OIDC_*` settings.

### Observed Behavior

The backend container receives `OIDC_ADMIN_GROUP=ArtifactKeeperAdmins` correctly, but the persisted OIDC provider does not use that value.

For example, after deployment the provider configuration still showed `Admin group: artifact-keeper-admins`, while the backend container environment contained `OIDC_ADMIN_GROUP=ArtifactKeeperAdmins`.

Other environment-managed OIDC properties are reconciled successfully, but the admin group remains unchanged.

No startup error or warning indicates that `OIDC_ADMIN_GROUP` is ignored.

### Steps to Reproduce

1. Configure an environment-managed OIDC provider with valid OIDC settings, including `OIDC_NAME`, `OIDC_ISSUER`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET`, `OIDC_SCOPES`, `OIDC_GROUPS_CLAIM`, `OIDC_ADMIN_GROUP`, and `OIDC_MAP_GROUPS_TO_GROUPS`.

2. Set `OIDC_ADMIN_GROUP=ArtifactKeeperAdmins`.

3. Start or restart the Artifact Keeper backend.

4. Verify that the backend container receives `OIDC_ADMIN_GROUP=ArtifactKeeperAdmins`.

5. Open the env-managed OIDC provider configuration through the Artifact Keeper administration UI or API.

6. Observe that the provider's admin-group value has not been reconciled from `OIDC_ADMIN_GROUP` and instead retains its previous or default value.

7. Change `OIDC_ADMIN_GROUP` to another value and restart the backend.

8. Observe that the persisted admin-group value still does not follow the environment variable, while other supported `OIDC_*` settings continue to be reconciled normally.

### Logs

```shell

```

### Additional Context

The environment configuration documentation includes `OIDC_ADMIN_GROUP`, but
the environment-bootstrap implementation does not currently read that variable
when constructing/reconciling the OIDC provider.

A possible fix would be to include `OIDC_ADMIN_GROUP` in the environment values
read by the OIDC bootstrap path and apply it to the same persisted provider
field used by the admin UI/API.

The important behavior is that an env-managed provider should reconcile the
admin group in the same way it already reconciles settings such as scopes,
groups claim, redirect URI, username/email claims, and
`OIDC_MAP_GROUPS_TO_GROUPS`.

The fix should remain backward-compatible when `OIDC_ADMIN_GROUP` is unset.

### Pre-submission Checklist

vkhilin-top:
## Summary

The env-based OIDC bootstrap read every documented `OIDC_*` setting except `OIDC_ADMIN_GROUP`, so the env-managed provider's persisted `attribute_mapping.admin_group` — the key the OIDC callback reads for admin elevation — never reflected the variable. Fixes #3420.

`admin_group` is now an optional field on `CreateOidcConfigRequest`/`UpdateOidcConfigRequest`, folded into `attribute_mapping.admin_group`. The bootstrap reads `OIDC_ADMIN_GROUP` and applies it on every boot. A missing JSON field still deserializes to `None`, and mapping semantics (including replace-based deletion) are unchanged, so existing API callers are unaffected.

### The revocation semantic is deliberate: `OIDC_ADMIN_GROUP` is env-definitive

An earlier revision of this branch carried the persisted admin group over when the variable was unset. That was wrong, and it is worth stating plainly rather than leaving it in the history: removing `OIDC_ADMIN_GROUP` and redeploying is the natural way an operator *revokes* group-based admin, and a carry-over would have made the last known group elevate its members forever, with no log line. It also contradicted the contract three lines away in the same `From` impl — *"Env is definitive: replace the whole mapping so a key removed from the environment is cleared from the stored config"* — and made `.env.example`'s "the remaining `OIDC_*` values … default to the backend defaults when unset" false for exactly one variable.

So `admin_group` behaves like every other key the bootstrap writes: **unset ⇒ cleared**. This is the fail-closed direction and it restores, rather than bends, the documented contract. Measured on this branch:

| boot of the env-managed provider with the variable **removed** | persisted mapping | log |
|---|---|---|
| `main` (before this PR) | `{}` | none |
| earlier revision of this branch | `{"admin_group": "ArtifactKeeperAdmins"}` — still elevating | none |
| this branch | `{}` | **WARN**, naming the group being cleared |

Because a change to who is admin should never be silent in *either* direction, the reconcile now logs the transition: INFO when the variable sets or changes the group, WARN when the variable is gone and a persisted group is being cleared. `.env.example` documents the rule and the escape hatch — an operator who wants the admin group owned by the admin API rather than the environment gives that provider a name other than `OIDC_NAME`, which keeps the env reconcile away from it entirely (verified live: an API-created provider's `admin_group` survives reboots untouched).

### Correction to this PR's original description

The original summary said the bug meant updating the variable "silently changed nothing about who is admin". That is not accurate and is worth correcting so reviewers judge the right change. `sso.rs:509-513` already had `.or_else(|| std::env::var("OIDC_ADMIN_GROUP").ok())` before this PR, and because the env reconcile replaced the attribute mapping wholesale on every boot, an env-managed provider's mapping was always empty — so the fallback engaged and `OIDC_ADMIN_GROUP` **was** effective at login on `main`.

What was actually broken, and what this PR fixes:

- The **persisted configuration did not reflect the variable.** `GET /api/v1/admin/sso/oidc` reported `"attribute_mapping": {}` no matter what `OIDC_ADMIN_GROUP` was set to, so the effective admin group was invisible to the admin API and the UI. It now reads back correctly.
- Correct behaviour depended entirely on a **login-time fallback that only engages when the mapping carries no value at all** — one admin-API write of any mapping key that included `admin_group`, or any future change to the mapping's lifecycle, and the variable would silently stop mattering. The value is now stored where it is read.
- The admin group was only settable by hand-writing the untyped `attribute_mapping` blob; it now has a typed request field.

The CHANGELOG entry has been rewritten accordingly.

### Scope: this may not reach the reporter

The `Skip` path is untouched. If a provider already exists under a name that is not `OIDC_NAME`, the bootstrap skips it entirely and logs a WARN — the mapping is never reconciled and `OIDC_ADMIN_GROUP` genuinely does nothing. Reproduced on this branch: a provider named `Corporate SSO` with `OIDC_NAME=default` still reads `{"admin_group": "OldGroup"}` after a reboot with `OIDC_ADMIN_GROUP=NewGroup`, and the log carries `env bootstrap skipped to avoid creating a duplicate`. That is precisely #3420's reported symptom, so **before closing the issue, confirm the reporter's provider name matches `OIDC_NAME`** (the report says there was no startup warning, and on that path there would have been one).

### Reported separately, not fixed here

The env reconcile clobbers admin-API-set `groups_claim` / `username_claim` / `email_claim` on every boot of an env-managed provider (pre-existing — `attribute_mapping_replace` ships in v1.7.1). It matters because the documented remedy for a claim-name mismatch seen in a customer evaluation is setting `groups_claim=groups_direct`, which therefore does not survive a restart. That is a separate structural fix (an env reconcile that builds the full desired mapping, or deep-merges with explicit deletion) and is filed on its own.

**These tests now actually run in CI.** `backend/Cargo.toml` declares `[[bin]] name = "artifact-keeper"`, and CI's only Rust test invocations are `cargo nextest run --workspace --lib` and an explicit `--test` allowlist. `--lib` does not build bin targets, so every test in `src/main.rs` — including this PR's — was unenforced. The OIDC env-bootstrap builder and the reconcile planner therefore move into `services::oidc_env_bootstrap`; `main.rs` keeps only the `std::env` reads. 36 tests move from a target CI never runs into one it does:

```
$ cargo nextest list --workspace --lib -E 'test(admin_group)'   # before: 2 (both unrelated DB tests)
$ cargo nextest list --workspace --lib -E 'test(oidc_env_bootstrap)' | wc -l
36
$ cargo nextest list --workspace --lib -E 'test(insecure_default_admin)'
(empty — a still-bin-resident test, showing the gap is real and general)
```

Each production hunk was reverted independently with the tests left in place; every one produces a failure:

| reverted hunk | failing test |
|---|---|
| `OIDC_ADMIN_GROUP` plumb-through in the builder | `test_bootstrap_request_admin_group`, `test_bootstrap_request_all_optional_fields` |
| the empty-value filter | `test_bootstrap_request_empty_admin_group_is_none` |
| carry-over restored in the reconcile planner | `test_plan_admin_group_env_unset_clears_persisted` |
| `create_oidc` mapping insert | `test_create_oidc_persists_admin_group_in_mapping` |
| `update_oidc` mapping insert | `test_update_oidc_admin_group_wins_over_replaced_mapping` |
| carry-over restored inside `update_oidc` | `test_update_oidc_env_reconcile_shape_clears_admin_group` |

Two tests are **forward guards, not evidence**, and say so in their doc comments: `test_bootstrap_request_no_admin_group` (asserts an absent variable contributes neither a request field nor a smuggled mapping key) and `test_oidc_create_to_update_absent_admin_group_stays_none`. They hold on both sides of the change and exist to keep the clear-on-unset path from being reopened.

## Test Checklist

- [x] Manually tested locally

Unit: the builder's `admin_group` cases plus six `plan_admin_group_reconcile` cases covering set / change / no-op / clear / non-string mapping value. DB-backed: create persists `admin_group` into the mapping; the field wins over a wholesale mapping replace; and the env-reconcile shape with the variable unset clears it, asserted on a re-read of the row and not just the response DTO. Full suite green: `cargo nextest run --workspace --lib` with a real `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1` — 15758/15758 passed.

Manual: deployed on an isolated instance and walked create ⇒ change ⇒ revoke, plus the `Skip` path and the API-owned-provider path, with the results in the table above.

**Test-hygiene fix included.** `make_create_req` now appends a `Uuid::new_v4()` to the provider name. Its doc comment already promised uniqueness, but the caller-supplied suffix was fixed, and `oidc_configs.name` is `UNIQUE` — so a test that failed before reaching `cleanup_oidc` left a row that made *every subsequent run* fail on a unique violation instead of on the original assertion. That cost a reviewer of this PR a false red.

## API Changes

- [ ] N/A - no API changes

No new endpoints and no DB migration — one optional field on two existing `ToSchema` request types.


brandonrc:
Two notes for whoever picks this up.

**Relation to #3494.** The test-target move in this PR is a local instance of that issue, not a fix for it: 36 of the 276 test functions #3494 counts as never executing move out of the bin target and into `--lib` here. The general fix (making CI build the bin target and the unlisted integration targets) still belongs there.

**Filed separately:** #3507 — the env reconcile discards admin-API-set `groups_claim`/`username_claim`/`email_claim` on every boot. Pre-existing, and deliberately not touched here so this PR stays scoped to the `admin_group` direction question. It matters because the supported remedy for a claim-name mismatch is an admin-API write that does not survive a restart.

Issue #2260, opened 2026-07-06
Download statistics silently miss many artifact downloads — counting is unreliable


mymarche:
# Bug Report: Download statistics silently miss many artifact downloads — counting is unreliable

## Description

Artifact download statistics (`download_statistics` table) are incomplete. A significant fraction of downloads are never recorded, making download counts unreliable for billing, analytics, and audit.

**Three distinct bugs cause data loss:**

**Bug 1. Presigned redirect path writes to non-existent table** — [`repositories.rs` (line 4314)](https://github.com/artifact-keeper/artifact-keeper/blob/main/backend/src/api/handlers/repositories.rs#L4314-L4323) does `INSERT INTO download_events` but the table `download_events` does not exist. The error is discarded (`let _ = sqlx::query(...)...execute().await;`). **All S3/CloudFront presigned downloads are silently lost.**

**Bug 2. Virtual and remote (proxy) repository downloads are never recorded** — [`resolve_virtual_download()`](https://github.com/artifact-keeper/artifact-keeper/blob/main/backend/src/api/handlers/proxy_helpers.rs#L1715-L1800) and [`proxy_fetch_streaming()`](https://github.com/artifact-keeper/artifact-keeper/blob/main/backend/src/api/handlers/proxy_helpers.rs#L531-L560) in `proxy_helpers.rs` do not write to `download_statistics` at all. Any artifact served through a virtual or remote repository is invisible in analytics.

[...]

PR #3447, opened 2026-08-18
fix(proxy): record proxied downloads and guard the whole class

### LINKED ISSUE #3446: backend: proxy/remote downloads are never recorded in ~19 formats, so Downloads always reads 0
## Summary

The Downloads count stays at 0 for proxy/remote repositories in most formats. There are **two independent halves**, one already fixed and one not.

**Half 1 — display (FIXED, unreleased).** `build_catalog_artifact_response` hardcoded `download_count: 0` for every proxy-cached row. Fixed in #3388 via a batched `download_counts_by_paths` lookup per page. **That fix is on `main` and ships in 1.8.0 — it is in no released version**, so on 1.7.x the column reads 0 even where downloads *were* recorded.

**Half 2 — recording (NOT fixed).** Most remote-capable formats never call `record_proxy_download` at all, so there is nothing for the display to show even after upgrading.

## Recording coverage today

Proxied downloads are recorded only via:
- `proxy_helpers::try_remote_or_virtual_download` (the shared path), used by **ansible, conda, cran, rpm, rubygems**
- explicit calls in **maven** (added by #3388), **npm**, **pypi**

Remote-capable formats that call **neither**:

`alpine`, `cargo`, `chef`, `cocoapods`, `composer`, `conan`, `debian`, `gitlfs`, `goproxy`, `helm`, `incus`, `jetbrains`, `nuget`, `oci_v2`, `pub_registry`, `sbt`, `swift`, `terraform`, `vscode`

That includes the highest-traffic proxy types — Docker/OCI, cargo, NuGet, Debian, Helm, Go.

## Verified by reading the code, not inferred

**`cargo.rs`** — the Remote arm returns the streamed upstream body directly:
```rust
return Ok(builder.body(Body::from_stream(result.body)).unwrap());
```
and `record_download` sits ~15 lines later on the hosted path, so a proxied crate download never reaches it. This is the same shape #3388 found and fixed in Maven, where "the Maven remote branch in `serve_artifact` was the only proxying format that never called `record_proxy_download`" — that conclusion was correct about Maven but the class is much wider.

**`oci_v2.rs`** — `record_oci_manifest_download` resolves the artifact by
```sql
SELECT id FROM artifacts WHERE repository_id = $1 AND storage_key = $2 AND is_deleted = false
```
and does nothing on `Ok(None) => {}`. Proxy-cached objects **have no `artifacts` row** — stated in `build_catalog_artifact_response`'s own comment — so every proxied manifest pull silently records nothing.

## Why this hid

`record_download` (hosted, keyed on `artifacts.id`, writes `download_statistics`) and `record_proxy_download` (proxy, keyed on repo+path, writes `proxy_download_statistics`) are different functions writing different tables. A format that calls only the first *looks* instrumented on inspection, and fails silently at runtime because the id lookup returns `None` for cached content. Nothing errors; the counter simply never increments.

## Suggested fix

1. Route proxied serves through `try_remote_or_virtual_download` where the format's shape allows, rather than adding N more explicit call sites.
2. Where a format must serve upstream bytes directly, call `record_proxy_download` on that arm.
3. **Add a class-level guard**, in the shape of the ones added this cycle (`every_user_can_access_repo_call_names_its_action`, `every_raw_virtual_member_walk_is_authorized_or_explicitly_exempt`): scan every format handler and fail if a remote/proxy serve path records nothing, unless it carries a written exemption marker. A per-format fix without this will drift again — the Maven fix in #3388 did not generalise, and 19 formats stayed broken behind it.

## Note for operators

Even after 1.8.0, only the formats listed as instrumented will report proxy downloads. Repositories in the other formats will continue to show 0 until half 2 is fixed, which is a reporting gap rather than a serving problem — artifacts are served correctly either way.

brandonrc:
## Summary

Fixes #3446 — the backend half. Proxied downloads were never recorded in most formats, so the Downloads column read 0 forever.

**Why it hid.** Two functions count a download and they are not interchangeable. `record_download` is keyed on an `artifacts.id` and writes `download_statistics`; `record_proxy_download` is keyed on `(repository, path)` and writes `proxy_download_statistics`. Proxy-cached content is deliberately not registered in `artifacts` (#1278), so a format calling only the first *looks* instrumented on inspection and records nothing at runtime — the id lookup returns `None`, nothing errors, and the counter simply never moves. That is exactly how ~19 formats stayed broken behind #3388, which fixed Maven and concluded Maven was the only one.

**Recording, for the six formats people actually proxy:**

- **cargo** — the Remote arm returned the upstream stream directly, so it never reached the `record_download` fifteen lines below on the hosted path. That call could not have helped anyway: a proxy-cached crate has no `artifacts` row.
- **oci_v2** — `record_oci_manifest_download` resolves a manifest through `artifacts.storage_key` and does nothing on `Ok(None)`, so **every** proxied `docker pull` counted zero on the highest-traffic proxy format we have. A new `record_oci_manifest_pull` picks the recorder matching where the bytes live, on both the warm (local-storage) and cold (upstream pass-through) manifest arms.
- **nuget** — the V3 flat-container `.nupkg` arm and the legacy V2 / Chocolatey seam. Version lists stay uncounted as metadata, and the #2929 repair path passes no context so it is not double-counted alongside its hosted `record_download`.
- **debian**, **helm**, **goproxy** — their pool / chart / module-zip arms.

Each records against the proxy-cache path the fetch commits under, so the count lands on the catalog row the artifact listing renders (#3388's display half), and *after* the fetch resolves so a 404 or 502 is not counted. The unit of a Docker pull is unchanged — once at the manifest, never per blob (#2260) — and a HEAD counts nothing anywhere.

**The class guard is the main deliverable.** `every_proxy_serve_path_records_a_download_or_is_explicitly_exempt` scans every format handler for the primitives that put upstream bytes in front of a client and fails unless each site either records (directly, via the recording `try_remote_or_virtual_download`, or one call hop away — pypi records in the *caller* of its streaming helper, npm in the *callee*, and both are correct, so both pass unmarked) or carries a written `UNRECORDED-PROXY-SERVE:` marker saying why. Not counting is sometimes right — a HEAD serves no body; an OCI blob must not count because a pull is counted at the manifest — so the exemption is a greppable token that keeps the judgement in review rather than in an allowlist that drifts.

The guard immediately found a serve path the per-format audit in #3446 had missed: **RPM's catch-all `/rpm/{repo}/*path` route records nothing**, even though RPM's main `download_package` path routes through `try_remote_or_virtual_download` and does. #3446 lists RPM as covered; it is covered on one of its two serve paths.

**Remaining surface is now visible and bounded.** The 14 formats not fixed in this pass — alpine, chef, cocoapods, composer, conan, conda, gitlfs, jetbrains, pub_registry, rpm's catch-all, sbt, swift, terraform, vscode — each carry a marker naming #3446, and `the_deferred_format_count_only_shrinks` pins that count at 14 so a new format cannot silently join them. Those repositories keep reporting 0: a reporting gap, not a serving defect.

No migration — `proxy_download_statistics` and `proxy_cache_artifacts` already exist. No new `sqlx::query!` macros, so no `cargo sqlx prepare` needed.

## Live validation

Validated on a real running backend, not just unit tests. Two isolated stacks, identical exercise script: create a remote/proxy repo per format, pull a real artifact from a real upstream twice, then read `GET /api/v1/repositories/{key}/artifacts` and check `download_count`.

| format | upstream | bytes served | before (`main`, ghcr `:dev`) | after (this branch) |
|---|---|---|---|---|
| goproxy | proxy.golang.org | 17866 | **0** | **2** |
| nuget | api.nuget.org | 2441966 | **0** | **2** |
| cargo | crates.io | 46741 | **0** | **2** |
| helm | charts.jetstack.io | 80746 | **0** | **2** |
| debian | deb.debian.org | 53080 | **0** | **2** |
| oci_v2 | registry-1.docker.io | 8077 | **0** | **2** |

Every request returned HTTP 200 with the correct body on *both* builds — the artifacts were always served correctly; only the counter was missing.

Additional live checks on the patched build:

- **A real `docker pull`** of an image the local daemon had never seen produced exactly one count, on the digest the client actually fetched (`manifests/sha256:8fc3…` = 1), with both layer blobs at 0 and the tag it only HEAD-probed at 0. One pull, one count, no per-blob inflation.
- **A real `docker pull` of a digest the daemon already had** issued only a HEAD and correctly counted nothing.
- **3× HEAD** on a manifest left the count unchanged at 2; the next GET took it to 3.
- **helm's `index.yaml`** stayed at 0 while the chart counted — metadata is not a download.

`test_proxied_crate_download_is_counted_3446` and `test_proxied_go_module_zip_is_counted_but_metadata_is_not_3446` are DB-backed, drive the real handler through a wiremock upstream, and assert through `download_counts_by_paths` — the same lookup the artifact listing uses — so they pin the number an operator sees rather than just the fact that a row exists. Both cover a cold miss and a warm cache hit, which take different paths through the handler. **Negative control performed:** with the two `record_proxy_download` calls removed, both tests fail with `proxied crate download 1 must be counted` / `proxied module zip download 1 must be counted`; restored, both pass.

The goproxy test also asserts the *negative* half — a `.mod` metadata fetch must leave the counter alone — so a future "record everything the proxy serves" change cannot pass.

## Test Checklist
- [x] Manually tested locally

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace --lib` all green against a real PostgreSQL with `AK_TESTS_REQUIRE_DB=1` (15688 passed, 0 failed).

## API Changes
- [x] N/A - no API changes

Response shapes are unchanged; `download_count` was already in the proxy listing payload (#3388) and simply always read 0.

## Not in this PR

- The other 14 formats (marked, tracked by the pinned count).
- The 1.7.x backport — handled separately after this lands.
- The UI half, artifact-keeper/artifact-keeper-web#808: distinguishing *measured zero* from *not tracked*. That distinction now matters more, not less — after this ships a bare `0` means "nobody used this proxy" for six formats and "we do not count this format yet" for fourteen, and rendering them identically could get a busy repository deleted.


brandonrc:
### Note on the failing Security Audit check

The `🔒 Security Audit` failure (and the `✅ CI Complete` rollup that depends on it) is **not from this PR**.

`cargo audit` flags **[RUSTSEC-2026-0258](https://rustsec.org/advisories/RUSTSEC-2026-0258)** — "h2 unbounded empty DATA frames" — in the transitive dependency `h2 0.4.13`. The advisory was **published 2026-08-17** and is patched in `h2 >= 0.4.16`.

Evidence that it is pre-existing and fleet-wide:

- This branch changes **no dependency files** — `git diff --name-only main..HEAD` matches neither `Cargo.toml` nor `Cargo.lock`.
- Every open PR whose audit ran after the advisory landed fails identically (#3445, #3444, #3440, #3439). #3435 passes only because its audit ran before publication.

Every other gate is green on this branch: Check Rust, Backend Unit Tests, Backend Integration Tests, Code Coverage, Build Backend Image, Smoke E2E, Shell Tests, CodeQL, GitGuardian.

Fixing it belongs in its own `chore(deps)` PR bumping `h2` to `>= 0.4.16`, not folded into a behavioural change — and per `CLAUDE.md` it should not be `--admin` bypassed either. Flagging rather than acting on it.

Issue #2391, opened 2026-07-10
Remediate 188 Trivy code-scanning CVEs (base image + bundled scanner tools + Go deps)


brandonrc:
Burn down the open code-scanning backlog. **All 188 alerts are Trivy** (dependency/base-image CVEs); **CodeQL is clean (0)**. All have a fix available. They cluster into a handful of high-leverage bumps:

## Buckets (by scanned binary)
- `usr/local/bin/trivy` — **70** (bundled Trivy binary's Go deps)
- `artifact-keeper-backend` — **44** (embedded Go tooling: cosign/sigstore-go + Go stdlib)
- `usr/local/bin/grype` — **39** (bundled Grype binary's Go deps)
- `usr/local/bin/scanner-adapter` — **35** (in-repo Go adapter deps)

## Top vulnerable packages (the actual levers)
- **Go `stdlib`** — 62 alerts across v1.23.12 / v1.25.7 / v1.26.2 (Fixed: 1.25.12, 1.26.5, 1.27.0-rc.2) → rebuild/re-pin Go binaries with patched Go toolchain
- **Alpine `libssl3`/`libcrypto3`** (3.5.6-r0) — 30 → refresh Alpine base (apk upgrade)
- **`golang.org/x/crypto`** (v0.46.0/v0.50.0) — 26 → go get -u
- **Alpine `curl`/`libcurl`** (8.17.0-r1) — 14 → apk upgrade
- **`golang.org/x/net`** (v0.48.0/v0.53.0) — 14 → go get -u
- **`go-git/v5`** (v5.16.5/v5.18.0) — 14 → go get -u
- **`oras-go/v2`** (v2.6.0) — 8 → go get -u
- sigstore-go, plus residual

Severity: 2 critical, 73 high, 80 medium, 31 low, 2 note.

## Plan
1. Refresh Alpine base image(s) so apk provides patched openssl3 + curl.
2. Bump the pinned **trivy / grype / cosign** versions (built with patched Go) in the Dockerfile(s) that install them.
3. Bump the scanner-adapter **Go toolchain + go.mod** deps (x/crypto, x/net, go-git, oras-go) and any embedded Go tooling in the backend image.
4. Rebuild affected images, re-scan with Trivy, confirm the code-scanning count drops; triage/accept any residual with no upstream fix.

May land as 1-3 PRs (base-image, bundled-tools, go-deps). Closes this issue when the backlog is cleared/triaged.

PR #3468, opened 2026-08-20
fix(security): override golang.org/x/mod to v0.40.0 in the grype build (#3465) (#3469)

### LINKED ISSUE #3465: v1.7.7 tag burned: CVE-2026-56864/-56865 (golang.org/x/mod) in the bundled grype binary fails the Docker Publish Security Scan
`v1.7.7` is a dead tag. Its `Security Scan` failed and no images or release object were produced; repository rulesets forbid deleting or moving a tag, so the version number is burned. This is the **third** burned tag in this series (`v1.7.5` on the scanner-adapter version-collision gate, #3429; `v1.7.7` here).

```
Trivy gate - Backend           Total: 2 (HIGH: 2)
Trivy gate - Scanner Adapter   Total: 2 (HIGH: 2)

usr/local/bin/grype (gobinary)
  golang.org/x/mod  CVE-2026-56864  HIGH  fixed  v0.38.0 -> 0.40.0
  golang.org/x/mod  CVE-2026-56865  HIGH  fixed  v0.38.0 -> 0.40.0
```

### Why the usual two levers do not work

`docker/Dockerfile.backend` stage 5 (`grype-bin`) builds grype from source at `GRYPE_VERSION=0.117.0` on a `GO_MIN_PATCH=1.26.6` toolchain floor (#3352). Neither of the two levers that stage already has clears this:

* **It is not a Go stdlib CVE**, so raising `GO_MIN_PATCH` does nothing. `golang.org/x/mod` is a module in grype's own dependency graph, pinned at `v0.38.0` in grype's `go.mod`.
* **There is no release to bump to.** `v0.117.0` (2026-08-10) is the latest anchore/grype release — `v0.118.0` does not exist. And anchore/grype `main` (`ffbca56`, 2026-08-14) *also* pins `golang.org/x/mod v0.38.0`; there is no merged or open PR bumping it. Unlike the sibling `artifact-keeper/trivy` case, where upstream `v0.74.0` carried `x/mod v0.40.0` and a plain version bump sufficed, upstream has not made this bump anywhere.

That leaves a **dependency override in our build** — `go get golang.org/x/mod@v0.40.0` before `go build` — which diverges from upstream and therefore needs the removal condition and build-time guards the `artifact-keeper/trivy` override policy requires.

### Reachability (context, not a reason to suppress)

The vulnerable code is **not linked into the grype binary**. Both advisories are scoped to specific import paths — `golang.org/x/mod/sumdb` (`Client.Lookup`, GO-2026-6180) and `golang.org/x/mod/sumdb/tlog` (`tileHashReader.ReadHashes`, GO-2026-6179) — and neither is in grype's build graph:

```
$ go mod why golang.org/x/mod/sumdb
(main module does not need package golang.org/x/mod/sumdb)

$ go list -deps ./cmd/grype | grep ^golang.org/x/mod
golang.org/x/mod/internal/lazyregexp
golang.org/x/mod/semver
golang.org/x/mod/module
golang.org/x/mod/modfile
```

`x/mod` enters through `cmd/grype/cli/commands/internal/jsonschema` -> `x/tools/go/packages` -> `x/tools/internal/gocommand` -> `x/mod/semver`. Trivy flags Go binaries at **module** granularity from the embedded build info, so it reports the module version without regard to which packages were linked; `govulncheck`, which works at symbol granularity, would report this binary as not affected.

So the practical urgency was low. It is still being **fixed at source rather than suppressed**, per the standing decision: `.trivyignore` is for findings with no fix available in a tool we can build, and this one has a fix.

### Scope

Backend image only. The scanner-adapter's copy of the same finding is cleared separately by repointing `docker/Dockerfile.scanner-adapter` at a new `ghcr.io/artifact-keeper/trivy` digest built from upstream `v0.74.0`, which carries `x/mod v0.40.0` with no override needed.

brandonrc:
## Summary

**Cherry-pick of 904cfab5 from `main` (#3469) onto `release/1.7.x`.** Fixes #3465.

This PR was originally authored directly against the maintenance branch and correctly failed `Verify commits trace back to main`. It has been rewritten as a `git cherry-pick -x` of the merged main commit, so it now carries a generated `(cherry picked from commit 904cfab5ead1823219bf9eac0d5da7ee01cae17d)` trailer and traces back through the gate's **narrowed-backport path (D)** — no `release-process: approved` label, and none has been applied.

Verified locally against the gate as merged in #3467:

```
$ MAIN_REF=origin/main scripts/ci/check-release-branch-commits.sh <base> <head>
  ✓ c394c45 narrowed backport of 904cfab on main (patch differs; hunks were resolved away):
    fix(security): override golang.org/x/mod to v0.40.0 in the grype build (#3465) (#3469)
::notice::All 1 PR commit(s) trace back to origin/main. Gate passed.
```

**Only the CHANGELOG hunk was resolved.** `docker/Dockerfile.backend` and `docker/Dockerfile.backend.alpine` applied without conflict and are **byte-identical to `origin/main`'s post-merge state** (verified with `cmp` against `git show origin/main:<path>`). The CHANGELOG conflicted because this branch's `## [Unreleased]` already carries a `### Fixed` entry from #3467 and because the entry's framing named `main`'s blocked `v1.8.1`; the resolution keeps #3467's `### Fixed` intact, adds `### Security` beneath it, and narrows the blocking sentence to this line's own history (`v1.7.7` burned, second after `v1.7.5`). Nothing was renamed — the dead-`v1.7.7` section rename belongs to the `chore(release): prepare 1.7.8` prep, not here.

---

### What the change is

`docker/Dockerfile.backend` stage 5 (`grype-bin`) builds grype v0.117.0 from source (#3352). grype's `go.mod` pins `golang.org/x/mod v0.38.0`, and two HIGH advisories fixed in v0.40.0 are reported against the compiled binary:

```
usr/local/bin/grype (gobinary)   Total: 2 (HIGH: 2, CRITICAL: 0)
  golang.org/x/mod  CVE-2026-56864  HIGH  fixed  v0.38.0 -> 0.40.0
  golang.org/x/mod  CVE-2026-56865  HIGH  fixed  v0.38.0 -> 0.40.0
```

**This is an override, not a bump, and that took verifying.** The stage already has two levers and neither reaches this:

* Not a Go **stdlib** CVE, so raising `GO_MIN_PATCH` does nothing — `x/mod` is a module in grype's dependency graph.
* No release to bump to. `v0.117.0` (2026-08-10) is the latest anchore/grype release, `v0.118.0` does not exist, **and `main` (`ffbca56`, 2026-08-14) still pins `x/mod v0.38.0`** with no open or merged PR moving it. Upstream has not made this bump anywhere. Contrast the scanner-adapter's copy of the identical finding, which is cleared by a plain version bump because upstream Trivy `v0.74.0` carries `x/mod v0.40.0` itself.

So both backend Dockerfiles apply `go get golang.org/x/mod@v0.40.0` after the commit-pin verification (left intact — it is a supply-chain statement) and before `go build`, modelled on the `artifact-keeper/trivy` override policy. Note that policy asks upstream to have made the same bump first; here they have not, on `main` or in any PR, and the comment block records that this goes one step further than those rules allow.

`go get` rather than `go mod edit -require`, for two reasons: it resolves through MVS and writes checksum-database-verified hashes into `go.sum` in the same step (a `-require` edit writes no `go.sum` and leaves `go.mod` inconsistent — `x/mod v0.40.0` requires `x/tools v0.49.0`, which `-require` would not select), and unlike a `replace` directive it records the **real** upgraded version in the binary's Go build info, which is exactly what a scan and any published SBOM read. `go mod tidy` afterwards is a verified no-op on `go.mod` and only costs a full test-dependency download, so it is not run.

Three build-time guards, so this cannot rot:

| Guard | Behaviour |
|---|---|
| Pinned source no longer requires `v0.38.0` (upstream shipped the fix) | **fails the build** — the block must be deleted. The tag bump that makes the override redundant is what stops the build until someone removes it. |
| `go get` moved a module outside the declared set (`x/mod`, `x/tools`, `x/crypto`, `x/net`, `x/text` — the deterministic MVS consequence) | **fails the build** rather than shipping an undeclared divergence |
| The compiled binary's Go build info does not record `v0.40.0` | **fails the build**, with a dependency-record plausibility check so it cannot report "verified" against build info it failed to parse |

### Reachability — context for how urgent this actually was

**The vulnerable code was never linked into the binary.** Both advisories are scoped to specific import paths — `golang.org/x/mod/sumdb` (`Client.Lookup`, GO-2026-6180) and `golang.org/x/mod/sumdb/tlog` (`tileHashReader.ReadHashes`, GO-2026-6179), the GOSUMDB transparency-log client:

```
$ go mod why golang.org/x/mod/sumdb
(main module does not need package golang.org/x/mod/sumdb)

$ go list -deps ./cmd/grype | grep ^golang.org/x/mod
golang.org/x/mod/internal/lazyregexp
golang.org/x/mod/semver
golang.org/x/mod/module
golang.org/x/mod/modfile
```

`x/mod` arrives only via `cmd/grype/cli/commands/internal/jsonschema` -> `x/tools/go/packages` -> `x/tools/internal/gocommand` -> `x/mod/semver`. Trivy flags Go binaries at **module** granularity from the embedded build info and cannot see which packages were linked; `govulncheck`, which works at symbol granularity, would call this binary unaffected.

It is still fixed at source rather than suppressed. `.trivyignore` is for findings with no fix available in a tool we can build; this one has a fix. Reachability changes how alarming the finding is, not whether a fixable HIGH gets fixed.

### Scope

Backend images only. `docker/Dockerfile.scanner-adapter` and `docker/scanner-adapter/VERSION` are deliberately untouched — the adapter's copy of the same CVE pair is being cleared separately by repointing at a Trivy image built from upstream `v0.74.0`. `docker/Dockerfile.backend.alpine` carries the identical stage per its own byte-for-byte requirement and the drift rule in `CLAUDE.md`; the two stages share a build-cache entry, which is the proof.

One inaccuracy noticed but **not** fixed here, to keep this diff to one thing: `Dockerfile.backend.alpine`'s comment claims "the guard in `backend/src/config.rs` fails the build if they disagree". No such guard exists on this branch — nothing under `backend/` references either Dockerfile. Worth a follow-up; the drift rule is currently enforced by convention only.

The regression test is in the Dockerfile itself and runs on every build, which is the only place that can catch this class: a Rust unit test cannot observe what a Go linker put in a binary. The build-info assertion fails the build if the override is not present in the compiled artifact, and the `from`-version guard fails it if the override has gone stale — the two ways this fix could silently stop working. Both were exercised: the build fails as intended when either is violated, and passes on this branch.

## Test Checklist
- [x] Manually tested locally

**Gate reproduction (linux/arm64, full backend image, CI's `no-cache-filters: rootfs-builder,grype-bin,grype-db-seed,runtime`, trivy `0.69.3`, `--severity CRITICAL,HIGH --ignore-unfixed --ignorefile .trivyignore --exit-code 1`):**

| | `usr/local/bin/grype` | UBI 9.8 base | gate exit |
|---|---|---|---|
| before | **2 (HIGH: 2)** | 0 | **1** |
| after | **0** | 0 | **0** |

Reproduced identically on a `linux/amd64` cross-compiled `grype-bin`, so this is not arch-specific.

**No collateral.** With all severities and no ignorefile, the only remaining findings in the image are the pre-existing, already-documented `github.com/docker/docker v28.5.2` pair (`CVE-2026-34040` HIGH, `CVE-2026-33997` MEDIUM) that `.trivyignore` covers and for which no resolvable fixed version exists. Zero Go **stdlib** findings — the `GO_MIN_PATCH=1.26.6` floor holds and the binary reports `GoVersion: go1.26.6`. Bumping the five `golang.org/x` modules introduced nothing new.

One thing worth flagging for whoever cuts the tag: an earlier local build reported **4 HIGH** in the UBI base (`curl-minimal`/`libcurl`, `CVE-2026-8286` and `CVE-2026-9547`, fixed in `7.76.1-40.el9_8.5`) purely because the `rootfs-builder` layer was a local cache hit. A fresh `rootfs-builder` installs `curl-minimal-7.76.1-40.el9_8.5` and the findings vanish. CI is already correct here — `no-cache-filters` includes `rootfs-builder` — but it is a live demonstration of #2391's failure mode, and a reason not to trust a locally-cached scan when clearing a release gate.

**grype still works** (verified against the rebuilt arm64 binary):
- `grype --version` prints exactly `grype 0.117.0` — asserted as a string equality, since `GrypeScanner` parses it into `scanner_version`.
- `grype version` reports `GitCommit: b5fa92bb…`, `GitDescription: v0.117.0`, `Supported DB Schema: 6` (so the DB seeded in stage 5b still loads), `GoVersion: go1.26.6`.
- `grype db update` succeeds and a real scan works on both a `docker-archive:` target (correctly reports the `docker/docker` advisories in a Go binary) and a `dir:` target (correctly reports apk packages). Stage 5b's `grype db update` also ran fresh in the no-cache build, so the pre-seed path is exercised end to end.

**Pre-push checks:** `cargo fmt --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, `cargo test --workspace --lib` **15202 passed / 0 failed / 8 ignored** (no `--skip services::http_client` needed — #3377 did not reproduce). **DB-backed tests were not run**: no `AK_TESTS_REQUIRE_DB=1` and no Postgres, so anything requiring a database silently no-opped. This PR changes no Rust, so that gap is not load-bearing here, but it is stated rather than implied.

## API Changes
- [x] N/A - no API changes

---

The release-branch gate is expected to **pass** now (#3467 landed the repaired gate; this commit takes its path D). If it does not, that is a gate defect worth investigating rather than labelling around — please do not apply `release-process: approved`. This PR does not touch `.github/workflows/docker-publish.yml`, `.github/workflows/release-preflight.yml`, `scripts/ci/check-release-branch-commits.sh`, `scripts/ci/release-preflight.sh` or this branch's `ci.yml`, all of which have another PR in flight against them.

The `CHANGELOG.md` entry sits under the existing `## [Unreleased]` heading and is deliberately not pre-numbered; the `chore(release): prepare 1.7.8` commit should retitle that heading and handle the dead-`v1.7.7` section rename.


brandonrc:
Converting to draft — **this PR is out of order and the gate failure is correct.**

`Verify commits trace back to main` fails because these are new commits authored directly on `release/1.7.x`, which is none of the three shapes the repaired gate (#3467) accepts: clean cherry-picks by patch-id, `chore(release):` preps, or narrowed backports carrying a `(cherry picked from commit <sha>)` trailer. Reaching for the bypass label on the first PR after that gate was repaired is exactly the habit the repair exists to end, so we are not doing that.

`main` needs the identical fix anyway — its Docker Publish is failing on the same two CVEs ([run 32419016885](https://github.com/artifact-keeper/artifact-keeper/actions/runs/32419016885)) and has published no images since 2026-08-19, so `v1.8.1` is blocked too. Main-first is both the compliant path and the one that fixes both release lines.

The change now lives in #3469 against `main`. Once that merges it will be cherry-picked here with `git cherry-pick -x`, so the trailer is generated rather than hand-written and this lands through the gate's clean-cherry-pick path with no label. The `docker/Dockerfile.backend` and `docker/Dockerfile.backend.alpine` contents are byte-identical between the two branches, and the grype binary built from each is byte-identical too (`sha256:cadb1a16…`), so nothing is lost in the move.

No review effort needed here in the meantime — please review #3469 instead.

Issue #3452, opened 2026-08-19
pypi/npm/maven native content routes fail to resolve packages, with inconsistent error messages per format


Chris-sentbe:
### Description

Native package-format content routes fail to resolve packages/artifacts for
at least three formats — pypi, npm, maven — each with a different error
signature:

- npm & maven: requesting actual package/artifact content from a `virtual`
  repo returns `{"code":"NOT_FOUND","message":"Virtual repository has no
  members"}`, even though the repo is `virtual` with members confirmed
  configured via the admin API (at least for one of the two).

- pypi: every request to `/pypi/*` returns a generic, unstructured
  `Repository not found` (text/plain, 20 bytes) — regardless of whether the
  target repo is `virtual` or a standalone `remote` repo, and regardless of
  URL shape (with/without the `/simple/` segment). No configuration
  variation tried produced a successful (200) response, which suggests a
  broader failure than the npm/maven "virtual repo members" case (see
  comparison table below).

Filing together since both symptoms sit in the same area (native
format-mount content resolution) and may share a root cause, even though
the exact failure point differs by format.

### Environment Information

- Deployment type: Docker Compose
- Operating System: Amazon Linux 2023
- Architecture: x86-64
- Database: AWS RDS for PostgreSQL 18
- Reverse Proxy: embedded Caddy (bundled with the app)
- Network path: client → VPN → ALB → Artifact Keeper

### Backend Version

1.8.0

### Web UI Version

1.7.0

### Package Format

None

### Expected Behavior

Native content routes should resolve packages/artifacts through a virtual
repo's configured members (local and/or remote):

  GET /pypi/{repo}/simple/{package}/                              → 200
  GET /npm/{repo}/{package}                                       → 200
  GET /maven/{repo}/{group-path}/{artifact}/{version}/{file}.pom  → 200

If a repo genuinely has no members or the requester lacks access, the three
formats should fail consistently and with a clear message — not diverge
between a structured JSON message, an empty body, and a generic "not
found" that's indistinguishable from "this repo key doesn't exist."

### Observed Behavior

| Format | Repo type tested            | Deep content request       | Result |
|--------|------------------------------|------------------------------|--------|
| maven  | virtual (members present*)   | GET .../org/.../x.pom       | 404 `{"code":"NOT_FOUND","message":"Virtual repository has no members"}` |
[...]

PR #3483, opened 2026-08-21
feat(cargo): age-gate enforcement for Cargo sparse registries via pubtime

### LINKED ISSUE #3480: Cargo: add remote publish-age gating using sparse-index pubtime
### Problem Statement

Artifact Keeper's remote publish-age gate currently supports npm, PyPI, and VS Code extensions from upstream publish timestamps and Go from first-seen observations. Cargo is explicitly unsupported, so an administrator cannot apply the same cooldown to a Cargo Remote repository even though Cargo's registry index now provides the timestamp needed to make the decision.

This leaves a meaningful gap for fast-burn Rust supply-chain attacks. On 2026-08-20, compromised releases of `arrayref` and `append-only-vec` introduced the malicious build dependency `proc-macro1 1.0.107`; its `build.rs` downloaded and executed a payload during compilation. Public crates.io-index history shows that the affected coordinates were corrected or removed quickly, but not before they were available to ordinary clients:

| Coordinate | Registry `pubtime` | Correction/removal | Observed window |
| --- | --- | --- | --- |
| `proc-macro1 1.0.107` | 07:11:15Z | deleted 08:03:09Z | 51m 54s |
| `arrayref 0.3.10` | 07:15:00Z | deleted 08:41:40Z | 1h 26m 40s |
| `append-only-vec 0.1.9` | 07:37:49Z | dependency corrected 09:22:36Z; version deleted 09:25:24Z | 1h 44m 47s to correction |

A minimum age longer than those windows would have withheld these versions while the public response caught up. This is not a claim about how many systems were affected; package download totals do not establish compromise count.

This is a narrow format follow-up to #1558, #2264, and merged PR #3075. It does not require another policy subsystem. Cargo's sparse-index schema defines an optional registry-assigned `pubtime` field, stabilized for Cargo clients in Rust 1.94, in the same newline-delimited metadata Artifact Keeper already proxies and parses for checksums.

### Proposed Solution

Add Cargo to the existing age-gate capability registry in `upstream_publish_time` mode and enforce the policy at both Cargo resolution surfaces:

1. **Sparse-index metadata**
   - Parse each entry's `vers` and optional `pubtime`.
   - Omit young versions from Remote index responses so stable Cargo resolves the newest eligible version naturally.
   - For Virtual repositories, evaluate every Remote member using that member's own policy before merging/deduplicating its index lines. Local and Staging members remain unaffected.

2. **Exact `.crate` downloads**
   - Enforce the gate before serving from the proxy cache or fetching upstream.
   - Reuse the exact sparse-index lookup already used to obtain `cksum`, also returning `pubtime` for the requested version.
   - This second check is required because an existing `Cargo.lock` or direct download URL can request an exact version without resolving a fresh index.

The implementation should reuse the existing review queue, manual approval and rejection behavior, structured HTTP 451 response, metrics, and audit events. Missing or malformed `pubtime` should fail closed while the gate is enabled and enter the normal review flow rather than silently exposing an unevaluable version.

One deliberate divergence from the npm/PyPI download seam: a blocked Cargo download should always return the structured 451 and never substitute the last-known-good artifact's bytes, because cargo verifies the downloaded `.crate` against the sparse-index `cksum` for the exact requested version — substituted bytes would surface as a checksum mismatch that looks like corruption rather than policy. The filtered index already provides the resolver-level equivalent: stable Cargo naturally resolves the newest eligible version.

Ungated Cargo repositories should retain their current byte-preserving fast path. Gated metadata responses may need to decode upstream content coding before filtering and return coherent uncoded headers. Raw upstream metadata can remain in the proxy cache, but policy-filtered bodies must not be shared through a cache key that omits repository policy, review state, or time.

I recommend limiting the first implementation to `upstream_publish_time`. Artifact Keeper's existing `first_seen` mode can be considered separately after its coordinate-identity semantics are agreed; it is not needed to address this incident class.

### Alternatives Considered

- **Cargo's native minimum publish-age setting:** RFC 3923 is implemented behind `-Zmin-publish-age`, but remains nightly-only today. It is also a per-client control, whereas Artifact Keeper can enforce one policy for developer and CI clients centrally. Cargo's documented lockfile exception is another reason the proxy download check remains useful.
- **Client configuration only:** this is difficult to keep consistent across developer machines, CI images, and older/stable toolchains, and direct downloads can bypass resolver-only policy.
- **Malware/advisory scanning only:** valuable defense in depth, but inherently reactive during the interval before detection, removal, or advisory issuance.
- **Use a crates.io versions side API:** unnecessary and registry-specific now that the standardized index document carries `pubtime` directly.

### Use Case

As a rust developer, I want to protect myself from fast-burn supply chain attacks on the cargo ecosystem, without having to rely on Cargo's own nightly "min-publish-age" feature.

As a platform or security administrator, I want a Cargo Remote or Virtual repository to delay newly published crate versions for a configurable period so that developers and CI cannot compile a fast-burn malicious release before the registry and security ecosystem have time to respond. I also want an explicit review override for urgent legitimate releases.

### Component

Backend / API

### Additional Context

Four design points where maintainer guidance would be useful before code:

1. Is a publish-time-only Cargo first slice the preferred scope?
2. Should an unparseable sparse-index line be omitted/fail closed when the gate is enabled, rather than preserved as today's ungated aggregation does?
3. For Virtual repositories, npm's existing download gate treats a blocked higher-priority Remote member as authoritative (451, no fallthrough to a lower-priority member with the same coordinate). I propose Cargo match that precedent for consistency — any objection?
4. The registry-index spec defines `pubtime` strictly (no fractional seconds, `Z` only), but Cargo itself parses it leniently as RFC 3339. Should the gate parse spec-strict (blocking-and-queueing spec-violating timestamps from third-party registries) or match Cargo's leniency?

I am willing to implement the backend and Cargo-client regression coverage once the scope and these semantics are confirmed.

Protocol and incident references:

- Cargo registry-index schema (`pubtime`): https://doc.rust-lang.org/cargo/reference/registry-index.html
- Cargo RFC 3923 tracking issue: https://github.com/rust-lang/cargo/issues/17009
- Cargo nightly `min-publish-age` documentation: https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#min-publish-age
- Rust 1.94 release notes: https://github.com/rust-lang/rust/releases/tag/1.94.0
- Incident analysis: https://www.aikido.dev/blog/two-popular-rust-crates-arrayref-and-append-only-vec-compromised-in-supply-chain-attack
- The Hacker News report: https://thehackernews.com/2026/08/rust-supply-chain-attack-puts-build.html
- `proc-macro1 1.0.107` removal in crates.io-index: https://github.com/rust-lang/crates.io-index/commit/c34a466f5f18685a1ac4cf07b77730f13c8e3c1f
- `arrayref 0.3.10` removal in crates.io-index: https://github.com/rust-lang/crates.io-index/commit/ce7217451cc9b077c5e48e274b756488bcfe12e9
- `append-only-vec 0.1.9` publication, correction, and removal: https://github.com/rust-lang/crates.io-index/commit/156bffa3dbbc3748e84822b9adf7f440b4d013d9 https://github.com/rust-lang/crates.io-index/commit/728d24764edb4ecfb224df1bac7eadc813f31640 https://github.com/rust-lang/crates.io-index/commit/189d50adbfcae8f855f07afba44930eb97a262e0

### Pre-submission Checklist

cazlo:
fixes #3480

This implements the Cargo sparse-registry age gate proposed in #3480, extending the existing per-Remote-repository gate to Cargo using the registry-index `pubtime` field (stabilized in Rust 1.94). The four design questions on the issue are still open, so I'm opening this as a draft wired to conservative defaults; each question maps to a single, isolated change site if the answers land differently (details below).

## What it does

Registers Cargo in the age-gate capability matrix for `upstream_publish_time` only. `first_seen` stays rejected: crates.io permits administrative deletion of a published (name, version) and the registry-index spec doesn't guarantee a deleted coordinate can't be republished with different content, so trusting elapsed time since this server's first observation would be unsound. The config endpoint's supported-format diagnostic is updated accordingly.

Filters Remote sparse-index responses through the repository's gate. Entries whose `pubtime` is younger than the configured minimum age — or missing — are dropped from the served NDJSON, so stable cargo naturally resolves the newest eligible version. Allowed lines are preserved byte-for-byte and in order. Malformed lines (invalid UTF-8, unparseable JSON, missing `vers`) fail closed: an enabled gate never passes a line whose version/time it couldn't evaluate, and one bad line can't hide the rest of the document from evaluation. Ungated repositories keep today's verbatim fast path untouched.

Enforces the gate on direct `.crate` downloads before any proxy-cache read, artifacts-row serve, or upstream stream — index filtering alone is bypassable by a `Cargo.lock`-pinned exact version or a hand-written URL, and a `.crate` already warm in the proxy cache must not outlive the gate. The gated exact-index lookup deliberately inverts `resolve_index_cksum`'s documented fail-open contract: under an enabled gate, an index that can't be fetched, decoded, or parsed is a 503, never an unverified serve (an upstream 404 stays a definitive 404). It also switches to the encoded fetch variant plus the shared bounded decoder, because the coding-blind 2-tuple fetch would let a stored-coded upstream body "parse" as an absent version. `br` and unknown codings fail closed with a 502, same as the Go metadata path — no brotli decoder is added.

Blocked downloads always return 451 and never substitute the last-known-good artifact. The shared seam offers an LKG that npm/PyPI use, but cargo verifies the downloaded `.crate` against the sparse-index `cksum` for the exact requested version, so LKG bytes would present as corruption or a MITM rather than policy. Index filtering already provides the resolver-level analog.

Virtual repositories filter each authorized Remote member's index contribution with that member's own policy before merge/dedup (Local and Staging members are unaffected), and pre-evaluate gated Remote members before download resolution. A blocked higher-priority member is authoritative — the 451 propagates rather than falling through to a lower-priority member, matching npm's virtual tarball loop. A version genuinely absent from a gated member's index drops that member from the candidate list (the index is authoritative existence evidence), and the local-ownership shadowing guard is unchanged.

Cache correctness: the raw ProxyService body cache is retained — raw upstream metadata is policy-neutral, so a gated repo costs a re-filter, not a re-fetch. The 5-minute in-process index cache (keyed `repo_key:crate`, no policy/review/clock component) is bypassed for reads and writes on gated Remote repos and on Virtuals with any gated member, so approvals and config changes are visible on the next request server-side. Clients honoring `max-age=300` may still see the prior view briefly; that's inherent to the response caching, not the gate.

Blocks and approvals emit the same review rows, metrics, and audit events as the existing npm/PyPI/Go/VS Code paths, with the label sourced from the capability registry.

## Open questions from #3480 and where each lands in the code

`pubtime` parse strictness: I implemented spec-strict (`yyyy-mm-ddThh:mm:ssZ` exactly; anything else is missing evidence → block and queue). Cargo itself deserializes via jiff and accepts broader RFC 3339, so a conforming third-party registry could emit timestamps cargo accepts but this rejects. The choice is isolated in `parse_cargo_pubtime` with tests covering the deliberately-rejected forms; flipping to cargo-lenient is a one-function change.

Virtual priority: blocked-is-authoritative produces a deliberate asymmetry when a gated and an ungated member both publish the same coordinate — the merged index lists it (the ungated member's line survives) while the download 451s. That's the conservative reading and matches npm; it's documented at `eligible_virtual_download_members`, which is the single place a different answer would change.

Malformed-line handling and publish-time-only scope are implemented as proposed on the issue.

## Tests

53 new tests across the three commits: pure coverage of the strict parser (boundaries, leap days, rejected lenient forms), the NDJSON collector/filter (ordering, blank lines, invalid UTF-8 isolation, yanked entries, duplicate-`vers` alignment between the index filter and the download gate), and the decoder wrapper; plus DB-backed tests through the real router covering young-omitted + 451 direct block (with an assertion that a last-known-good candidate genuinely resolves, so never-LKG can't pass vacuously), old allowed, manual approval visible without waiting out the index-cache TTL while the raw proxy cache is provably retained, missing `pubtime` blocked, gate service unavailable → 503, br-coded gated index → 502, disabled gate byte-for-byte passthrough, a warm pre-gate index body unable to leak after enabling, a warm artifacts row of a young version still blocked (locking the gate-before-artifacts-lookup ordering), mixed gated/ungated/Local virtual members, blocked-member no-fallthrough with the lower-priority member's download mock pinned to zero hits, and private-member metadata kept out of the shared index cache.

## Noticed in passing (pre-existing, unchanged here)

The ungated Virtual member index fetch uses the coding-blind 2-tuple fetch, so a stored-coded member body silently contributes zero index entries today. I kept ungated behavior byte-identical in this PR; happy to fix that separately.

No new dependencies, no migration, no frontend change.

### DIFF
--- a/backend/src/api/handlers/age_gate.rs
+++ b/backend/src/api/handlers/age_gate.rs
@@ -444,7 +444,8 @@ pub async fn update_repo_age_gate(
         if !AgeGateService::supports_format_mode(&format, mode) {
             return Err(AppError::Validation(format!(
                 "The age gate cannot be enforced for format {:?} in '{}' mode; \
-                 supported today: npm, pypi, and vscode (both modes), go (first_seen)",
+                 supported today: npm, pypi, and vscode (both modes), go (first_seen), \
+                 cargo (upstream_publish_time)",
                 repo.format,
                 mode.as_str()
             )));
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -251,7 +251,7 @@ fn split_url(url: &str) -> Option<(String, String)> {
 }
 
 // ---------------------------------------------------------------------------
-// Age-gate sparse-index filtering (#3480, Phase 1 — pure, no IO)
+// Age-gate sparse-index filtering (#3480)
 // ---------------------------------------------------------------------------
 //
 // These helpers extract `(version, publish_time)` pairs from a sparse-index
@@ -259,12 +259,9 @@ fn split_url(url: &str) -> Option<(String, String)> {
 // (`crate::services::age_gate_service`, `pub(crate)`), and rebuild a
 // filtered document once the caller knows which versions are blocked — the
 // same split as the Go `@v/list` filter (`parse_version_list` /
-// `filter_version_list` in `goproxy.rs`). Not yet wired into `serve_index`,
-// `try_remote_index`, or `try_virtual_index`: that DB-backed enforcement,
-// the content-coding-aware `_encoded` fetch variant, and the `resolve_index_
-// cksum` fail-open-to-fail-closed inversion are Phase 2. `#[allow(dead_code)]`
-// mirrors `package_name_from_crate_filename` in `formats::cargo` — a pure
-// primitive kept ready for a call site that lands in a later phase.
+// `filter_version_list` in `goproxy.rs`). They stay pure and IO-free; the
+// handler wiring that fetches, decodes, and enforces lives under
+// "Age-gate enforcement" below.
 
 /// Minimal fields extracted from one Cargo sparse-index NDJSON line for
 /// age-gate evaluation: the version being described and its optional
@@ -276,7 +273,6 @@ fn split_url(url: &str) -> Option<(String, String)> {
 /// full `IndexEntry` requires) still parses far enough to be age-gate
 /// evaluated.
 #[derive(serde::Deserialize)]
-#[allow(dead_code)]
 struct CargoAgeGateLineFields {
     vers: String,
     #[serde(default)]
@@ -303,7 +299,6 @@ struct CargoAgeGateLineFields {
 /// Isolated as a single function — with its own tests below, including the
 /// forms it deliberately rejects — so switching to Cargo's own leniency
 /// later, if that is the maintainers' answer, is a one-function change.
-#[allow(dead_code)] // not yet wired into a handler; see the module comment above.
 fn parse_cargo_pubtime(raw: &str) -> Option<chrono::DateTime<chrono::Utc>> {
     use chrono::{NaiveDate, NaiveTime, TimeZone, Utc};
 
@@ -363,7 +358,6 @@ fn parse_cargo_pubtime(raw: &str) -> Option<chrono::DateTime<chrono::Utc>> {
 /// `evaluate_versions_batch`, [`filter_cargo_index_lines`] independently
 /// omits any line this function could not parse too — an enabled gate must
 /// never pass a line whose version/time it never evaluated.
-#[allow(dead_code)] // not yet wired into a handler; see the module comment above.
 fn collect_cargo_index_versions(
     content: &[u8],
 ) -> Vec<(String, Option<chrono::DateTime<chrono::Utc>>)> {
@@ -399,6 +393,11 @@ fn collect_cargo_index_versions(
 /// removed, preserving each surviving line byte-for-byte (no
 /// re-serialization) and in original order.
 ///
+/// Phase 2 note: [`gated_remote_index_response`] calls this on EVERY gated
+/// response, including one with an empty `blocked` set, so the malformed-line
+/// contract below applies uniformly rather than only when something is
+/// withheld.
+///
 /// This is the enabled-gate contract: it is deliberately more aggressive
 /// than [`collect_cargo_index_versions`] on malformed input. That function
 /// only SKIPS a line it cannot parse, so one bad release does not hide the
@@ -408,12 +407,10 @@ fn collect_cargo_index_versions(
 /// NDJSON, per the plan). Blank lines are dropped, matching
 /// [`collect_cargo_index_versions`] and the Go `@v/list` filter's shape.
 ///
-/// The caller (Phase 2's handler-level wiring — not yet added) is expected
-/// to invoke this only when the age gate is enabled and enforceable; an
-/// ungated repository keeps today's verbatim proxy fast path untouched.
-/// `content` must already be decoded bytes for the same reason as
-/// [`collect_cargo_index_versions`].
-#[allow(dead_code)] // not yet wired into a handler; see the module comment above.
+/// Only [`gated_remote_index_response`] calls this, and only once the gate is
+/// known enabled and enforceable; an ungated repository keeps today's verbatim
+/// proxy fast path untouched. `content` must already be decoded bytes for the
+/// same reason as [`collect_cargo_index_versions`].
 fn filter_cargo_index_lines(content: &[u8], blocked: &HashSet<String>) -> Bytes {
     let mut out: Vec<u8> = Vec::with_capacity(content.len());
     let mut wrote_any = false;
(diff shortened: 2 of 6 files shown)


brandonrc:
Thanks for this — it is substantially right, so I have taken it the last mile rather than send it back: rebased onto current `main` (zero conflicts, your three commits are intact and unmodified) and pushed one fix commit on top, `98d127c5`, so it can merge into 1.10.0 with the rest of the age-gate work. Your authorship stays on your commits and the release notes credit you.

What the fix commit changes, and why:

1. **`age_gate_format_from_str` had no `cargo` arm** — a latent fail-open. Your enforcement seam resolves policy through `resolve_repo_params`, which reads the typed `repositories.format` column, so nothing misbehaves today; but any caller that builds `AgeGateRepoParams` from a `RepoInfo` *string* (`proxy_helpers::age_gate_params`, as npm/pypi do) would have resolved a Cargo repo to `Generic` — outside the matrix, never gated. Added the arm and a test that walks the capability registry and asserts every entry maps back from its own spec label, so a future entry cannot be added without a match arm.
2. **The Virtual download arm reopened the #3452 existence oracle.** The inlined member walk emitted a bespoke `(404, "Artifact not found in any member repository")` text/plain tuple where `resolve_virtual_download` emits `no_accessible_members_response()` (JSON envelope), so an anonymous caller could distinguish "empty virtual" from "members you may not see" by body/content type. Routed through the shared helper (as `npm.rs`/`pypi.rs` do) with a test comparing the private-member and zero-member responses — it fails on the original code with exactly that text/plain-vs-JSON difference.
3. CHANGELOG bullet under `### Added`, crediting you and this PR.

Kept your design answers to #3480's open questions as they are (spec-strict `pubtime`, malformed lines dropped from gated responses, publish-time-only scope with `first_seen` refused for Cargo, blocked-is-authoritative for Virtual) — each is isolated to one site if the maintainers decide differently later. The one pre-existing defect you called out (the ungated Virtual member index fetch going through the coding-blind `proxy_fetch_capped`) is filed separately.

Verified: `cargo clippy --lib --bins --tests -- -D warnings` clean; `test(cargo) or test(age_gate)` 380/380 including all 18 `*_3480` DB-backed cases; the broader npm/PyPI age-gate sweep 468/468.

brandonrc:
Rebased onto main after #3932 merged. The only conflict was in backend/src/api/handlers/cargo.rs: both sides append a new test module at the end of the file (#3932's catalog_registration_tests, this PR's age_gate_tests). Resolved keep-both, no logic changed. CHANGELOG rebuilt on top of main's copy. New head c5656f03.

Issue #3459, opened 2026-08-19
Maven `.sha1` files are cached for only 5 minutes, so every checksum request goes back to Maven Central


vldqq:
### Description

Checksum files for released Maven artifacts are cached with a 300-second expiry. Every `.sha1` request after that window makes the server contact Maven Central again.

Maven asks for a checksum after every artifact it downloads, so this puts an upstream round-trip on almost every dependency in a build.

### Environment Information

- Deployment type: docker
- Operating System: linux
- Architecture: amd64
- Database: postgres 16.13
- Reverse Proxy: Caddy

### Backend Version

1.8.0

### Web UI Version

1.8.0

### Package Format

Maven / Gradle

### Expected Behavior

A `.sha1` for a released artifact should be cached with the same indefinite lifetime as the artifact it describes, and served from cache without contacting upstream.

### Observed Behavior

The `.pom` and `.jar` files themselves are cached indefinitely and never re-fetched. Only their checksums expire.

Measured over one 5-minute build: **1000 upstream requests, every one of them a `.sha1`, and zero for `.pom` or `.jar`.** Every one came back `304 Not Modified` - the checksum had not changed, because it cannot.

A released Maven coordinate is immutable, so the checksum of that file is immutable too. Caching the hash of a file for less time than the file itself has no benefit: the extra request can never return anything different.

### Steps to Reproduce

Virtual Maven repo `libs-release` → Remote member `maven-remote` → `https://repo.maven.apache.org/maven2`, S3 storage

### Logs

```shell
21:35:58.454  GET /maven/libs-release/avalon-framework/avalon-framework/4.1.3/avalon-framework-4.1.3.pom.sha1
              Cached artifact .../avalon-framework-4.1.3.pom.sha1/__content__
              (167 bytes, expires at 2026-08-19 21:40:58 UTC)

21:35:58.699  GET /maven/libs-release/javax/servlet/servlet-api/2.3/servlet-api-2.3.pom.sha1
              Cached artifact .../servlet-api-2.3.pom.sha1/__content__
              (150 bytes, expires at 2026-08-19 21:40:58 UTC)

21:35:59.069  GET /maven/libs-release/commons-chain/commons-chain/1.1/commons-chain-1.1.pom.sha1
              Cached artifact .../commons-chain-1.1.pom.sha1/__content__
              (44 bytes, expires at 2026-08-19 21:40:59 UTC)

[...]

Issue #3335, opened 2026-08-13
proxy cache: a just-cached entry is unreadable until after the response completes, so a closely-following request refetches upstream


brandonrc:
## Summary

A proxy-cache entry is not readable until after the client response that
created it has already completed. A request that arrives inside that window is
classified as a cache miss and goes back to the upstream, even though the
object was just fetched and cached successfully.

Not a regression — this is long-standing behaviour of the streaming proxy path,
and it is not a v1.7.2 blocker. Filing it because it was surfaced while
diagnosing an intermittent release-gate failure, and because the window has two
real consequences.

## Mechanism

`CachePersister::tee_stream` (`backend/src/services/proxy_service.rs`) forks the
upstream body to the client and to a `tokio::spawn`ed writer. The client
response is complete as soon as the last byte is teed. The writer then still has
to, in order:

1. finish `put_stream` into a private staging key,
2. `copy` the staged object onto the live `__content__` key,
3. `discard_staging` (a delete),
4. `pin_storage_etag` (a HEAD),
5. `put` the `__cache_meta__.json` sidecar,
6. `invalidate_proxy_metadata_lru`.

Freshness is decided from that sidecar:
`read_cached_with_revalidation_streaming` loads it, and
`cache_classifier::evaluate` returns `Freshness::Miss` when it is absent. So for
the whole span of steps 1-6 the just-cached object reads as "not cached".

The metadata LRU compounds it slightly: `load_cache_metadata` uses
`try_get_with` over a `Cache<String, Option<CacheMetadata>>`, so the `None`
observed by the request that opened the fetch is itself memoised until step 6
invalidates it.

## Consequences

1. **Freshness contract.** For a path classified `Mutable` (which includes every
   path in a `Generic` repo, via the conservative `_ => mutable_default()` arm),
   the entry carries a 300 s TTL. Inside the publish window that TTL is
   effectively zero: a re-request refetches upstream and serves whatever the
   upstream now returns. If the upstream content changed in between, the new
   bytes are served and cached — which is exactly what
   `tests/security/test-cache-poisoning.sh` reports as cache poisoning.

2. **Upstream amplification.** A burst that arrives just behind a completing
   fetch is not coalesced by the single-flight coordinator (the leader has
   already finished) and each request opens its own upstream fetch.

## Reproduction

Against `ghcr.io/artifact-keeper/artifact-keeper-backend@sha256:816aed4f...`,
[...]

PR #3488, opened 2026-08-21
fix(proxy): publish the tee digest at Content-Length, not only at EOF

### LINKED ISSUE #3487: Proxy-cache commit gate refuses every gated object when the consumer stops reading at Content-Length
### Description

The non-SHA-256 proxy-cache commit gate added for GHSA-qxv7-p3mq-88fv in 1.7.4 refuses to cache every gated object, not just mismatched ones. On our instance that means no proxied Maven release `.pom` or `.jar` has been cached since we upgraded. The bytes are served correctly, so nothing is broken from a client's point of view; the cache has simply stopped caching.

The tee-computed digest that the gate compares against is only finalized in the block after the upstream loop, which runs when the consumer polls the stream once more and receives `None`. A consumer holding a fixed-length body need not do that: once it has `Content-Length` bytes it can drop the stream, and then that block never runs. The writer reads an empty digest slot, and because a missing digest deliberately fails the gate closed, the object is discarded.

`.sha1` sidecars are the only SHA-1 expectation in normal use, so Maven is where it shows. The SHA-512 path used by npm `dist.integrity` looks equally exposed. SHA-256 expectations are unaffected because they read the observed digest from the storage layer rather than the tee.

Two things make it easy to miss. The gate emits a `warn` with no metric, which the 1.7.4 notes call out as intentional. And checksum sidecars are excluded from gating, so they keep caching normally: every gated fetch caches its sidecar and discards the artifact, so `proxy_artifact_count` keeps climbing and the instance looks healthy.

### Environment Information

- Deployment type: kubernetes (Helm)
- Operating System: linux
- Architecture: amd64
- Database: Aurora PostgreSQL 16.13
- Reverse Proxy: AWS ALB
- Storage backend: S3

### Backend Version

1.7.6

### Web UI Version

1.8.0

### Package Format

Maven / Gradle

### Expected Behavior

A release `.pom` or `.jar` whose bytes match its published `.sha1` is committed to the proxy cache and served warm afterwards.

### Observed Behavior

It is never committed. Reproduced on fresh, never-cached coordinates across four groups, poms and jars alike: `org.apache.commons:commons-lang3:3.12.0`, `com.fasterxml.jackson.core:jackson-core:2.15.2`, `org.slf4j:slf4j-api:2.0.9`, `at.yawk.lz4:lz4-java:1.11.0`.

The digests agree in every case. For `lz4-java-1.11.0.pom`:

```
sha1sum of the served body      d66ad1b6800d46b96eff3723c9fad202592155c5
published .sha1 sidecar         d66ad1b6800d46b96eff3723c9fad202592155c5
upstream X-Checksum-Sha1        d66ad1b6800d46b96eff3723c9fad202592155c5
```

Bare canonical lowercase hex, 40 characters, so `parse_maven_sha1_sidecar` accepts it.

After the fetch, `GET /api/v1/repositories/{key}/artifacts` lists `lz4-java-1.11.0.pom.sha1` but not `lz4-java-1.11.0.pom`. Repeat requests never cache it.

### Steps to Reproduce

1. Remote Maven repository proxying an upstream that publishes `.sha1` sidecars.
2. `GET /maven/{repo}/org/apache/commons/commons-lang3/3.12.0/commons-lang3-3.12.0.pom` for a coordinate not already cached.
3. The body returns 200 and is correct.
4. `GET /api/v1/repositories/{repo}/artifacts` lists the `.sha1` but not the `.pom`.

It also reproduces as a unit test against `CachePersister::tee_stream`, with no HTTP or storage backend involved. Read exactly `Content-Length` bytes from the returned stream and drop it without the final poll, with a matching `CacheCommitDigest::Sha1Hex` in the template. No metadata sidecar is written and nothing is published to the live key. Polling the same stream to `None` commits normally, which is why the existing gate tests pass.

### Logs

```
WARN artifact_keeper_backend::services::proxy_service: streamed body checksum
does not match the requested digest; not caching (no metadata sidecar) so the
cache entry cannot be poisoned
cache_key=proxy-cache/{repo}/at/yawk/lz4/lz4-java/1.11.0/lz4-java-1.11.0.pom/__content__
```

One per gated fetch, and no other cache-write warnings. The `#2928` object ceiling is never hit, there is no `Content-Encoding` on either hop, and cache invalidation is not involved: fresh coordinates fail identically.

### Notes

The existing coverage explains how it shipped. The mismatch posture is tested for both SHA-1 and SHA-512, but the matching posture only for SHA-512, and every gate test drives the stream with `while let Some(chunk) = client.next().await`, which always reaches the end-of-stream publish. Nothing exercised a consumer that stops at `Content-Length`.

A patch that publishes the digest as soon as the advertised length has been hashed fixes it, with the full lib suite green. I will open a PR against this issue.

travisluecke:
Fixes #3487.

The GHSA-qxv7-p3mq-88fv commit gate compares a tee-computed digest against the expected one for SHA-1 and SHA-512 expectations. That digest was only finalized after the upstream loop, in the block that runs when the consumer polls the stream once more and gets `None`.

A consumer holding a fixed-length body need not poll again. Once it has `Content-Length` bytes it can drop the stream, and then the end-of-stream block never runs. The writer reads an empty slot, and since a missing digest fails the gate closed, the object is refused. Not intermittently: every object on a gated format. In practice that is every proxied Maven release `.pom` and `.jar`, because `.sha1` sidecars are the only SHA-1 expectation in normal use. The npm `dist.integrity` SHA-512 path has the same exposure. SHA-256 is unaffected, since it takes the observed digest from the storage layer.

## The fix

Publish the digest as soon as the advertised length has been hashed:

```rust
hashed_bytes += slice.len() as u64;
if let Some(total) = expected_len {
    if hashed_bytes >= total {
        if let Some(hasher) = tee_hasher.take() {
            if let Ok(mut slot) = tee_digest_slot.lock() {
                *slot = Some(hasher.finalize_hex());
            }
        }
    }
}
```

A consumer can only stop early when it knows the length, so this covers exactly the case that breaks and leaves a chunked response on the existing EOF publish. `take()` keeps the finalize once-only, and the EOF block already skips a `None` hasher. Publication also moves earlier relative to `tx.send`, so the ordering the original comment relies on, digest visible before `put_stream` can return, is preserved rather than weakened.

## Tests

`test_tee_sha1_gate_commits_when_consumer_stops_at_content_length` reads exactly `Content-Length` bytes and drops the stream without the final poll. It fails on `main` with `left: 0, right: 1` and passes with the fix.

Four more fill the gap that let this through. The mismatch posture was covered for both SHA-1 and SHA-512, but the matching posture only for SHA-512, and Maven is the only format that exercises SHA-1:

- `test_tee_sha1_gate_match_commits`, the SHA-1 counterpart of the existing SHA-512 match test
- `test_tee_sha1_gate_match_commits_multi_chunk`, since a real body arrives as many chunks
- `test_tee_sha1_gate_match_commits_across_chunk_split`, covering a body over `TEE_MAX_CHUNK_BYTES`, the shape a `.jar` takes
- the early-drop test above

Full `--lib` suite: 15197 passed, 1 failed, 8 ignored. The single failure is `api::handlers::health::tests::test_probe_storage_health_filesystem_unwritable`, which fails the same way on an unmodified v1.7.6 checkout in the same container because the test needs a path the process cannot write to and the build runs as root. Unrelated to this change.

## Notes for reviewers

Worth deciding whether the missing-digest case should stay silent. It currently logs the same "checksum does not match" warning whether the digests genuinely differ or the digest was never published, which is what made this hard to spot. A distinct message for an absent digest, or the `ak_proxy_cache_checksum_mismatch_total` counter the 1.7.4 notes mention this path skips, would have surfaced it immediately. Happy to add either if you want it in the same PR.

### DIFF
--- a/backend/src/services/proxy_service.rs
+++ b/backend/src/services/proxy_service.rs
@@ -2081,6 +2081,9 @@ impl CachePersister {
             // point of the guard is to bound bytes-to-disk: a post-hoc check
             // only fires once the unbounded object has already been written.
             let mut cached_bytes: u64 = 0;
+            // Bytes fed to the tee hasher. Tracked separately from
+            // `cached_bytes`, which only advances when a ceiling is configured.
+            let mut hashed_bytes: u64 = 0;
             let mut cache_abandoned = false;
             while let Some(chunk_result) = upstream.next().await {
                 match chunk_result {
@@ -2123,6 +2126,29 @@ impl CachePersister {
                             if !cache_abandoned {
                                 if let Some(hasher) = tee_hasher.as_mut() {
                                     hasher.update(&slice);
+                                }
+                                hashed_bytes += slice.len() as u64;
+                                // Publish the tee digest as soon as the
+                                // advertised length has been hashed. A consumer
+                                // that stops polling once it holds
+                                // `Content-Length` bytes -- which a server
+                                // serving a fixed-length body may legitimately
+                                // do -- drops this stream without a final poll,
+                                // so the end-of-stream block below never runs.
+                                // The writer then compares against an empty
+                                // slot and, since a missing digest fails the
+                                // gate closed, refuses to cache EVERY object.
+                                // Only a known length lets a consumer stop
+                                // early, so this covers exactly that case; the
+                                // chunked path still publishes at EOF below.
+                                if let Some(total) = expected_len {
+                                    if hashed_bytes >= total {
+                                        if let Some(hasher) = tee_hasher.take() {
+                                            if let Ok(mut slot) = tee_digest_slot.lock() {
+                                                *slot = Some(hasher.finalize_hex());
+                                            }
+                                        }
+                                    }
                                 }
                                 let _ = tx.send(Ok(slice.clone())).await;
                             }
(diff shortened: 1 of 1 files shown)


brandonrc:
Independently reproduced live, from a different starting point — this was hit while tracing an unrelated Maven checksum-caching issue (#3459), not while looking for it. Recording the evidence here because it corroborates the diagnosis from the outside.

Deployed `dev` image, a Remote Maven repo to Maven Central and a Remote npm repo, reading the cache volume directly:

```
GET /maven/mvn2/.../slf4j-api-1.7.36.jar   -> 200, sha1 of served bytes ==
                                              upstream .sha1 (6c62681a...)
WARN streamed body checksum does not match the requested digest; not caching
     cache_key=proxy-cache/mvn2/.../slf4j-api-1.7.36.jar/__content__
GET /npm/npm-remote/is-odd/-/is-odd-3.0.1.tgz -> 200, same WARN, no cache entry
```

Two things worth adding to the record:

**The served bytes were correct.** The SHA-1 of what the client received matched the upstream `.sha1` exactly. So the gate is not catching a corrupt body — it is refusing a byte-perfect one because the slot it compares against is empty. That matches the "fails closed on a missing digest" reading in the PR description, observed rather than inferred.

**The blast radius is every request, not every object.** The only things that ended up in the cache were the `.sha1` sidecars and the npm packument (no gate). Every release `.jar`/`.pom` and every npm tarball through a direct Remote repo is re-fetched from upstream on *every single request* — the cache never populates, so there is no steady state where it starts working.

That last part is the reason I'd argue this outranks #3459 on impact. #3459 is a report about upstream amplification (a checksum re-fetched every 5 minutes, ~1000 redundant round-trips per build); this is the same failure mode without the 5-minute floor, on the artifacts themselves rather than their sidecars. A proxy repo that never caches also silently removes the availability property the proxy exists for — if upstream is down or rate-limits, there is nothing to serve.

On the open question in the PR notes about whether the missing-digest case should stay silent: it should not. The warning currently reads `checksum does not match`, which sent me looking for a corruption bug for a while before I realised the slot was simply empty. Those are different operational conditions — one means "upstream or transport is lying to you", the other means "the gate is broken" — and they should not share a message.

No comment on the fix itself; I haven't reviewed the diff.

brandonrc:
## Audit round 1 — tee digest published at Content-Length (2 reviewers: correctness / security+integration)

Both passes read the head, merged `origin/main` (64 commits) into a throwaway branch — **clean, zero conflicts; `cargo check --lib --tests` exit 0**; the three main commits touching `proxy_service.rs` since the merge base (#3609, #3599, #3513) are cache-key/catalog changes outside `tee_stream` — and ran the tee tests under the lock: **10–11 pass**. Both independently reverted the 26 production lines with the tests kept: exactly one test fails, `test_tee_sha1_gate_commits_when_consumer_stops_at_content_length`, with `left: 0, right: 1` — precisely the claim in the PR. The diagnosis is correct and the fix is minimal.

**Security: could not break it.** The `>=` trigger can hash a prefix if upstream over-delivers, but `classify_stream_write` (`:818-826` / `:836-847`) rejects on `bytes_written != expected` in *both* directions, so an over- or under-length body never reaches the gate; hashing is paired with `tx.send` in the same block, so the published digest is always the hash of exactly the bytes stored. `take()` guarantees a single publication. `Content-Length` is read verbatim from the upstream header; absent/non-numeric → `None` (branch disabled, correct); `0` publishes then hits `RejectEmpty`/`RejectTruncated`. Truncated bodies, chunked/no-length, HEAD, 304, the #2928 ceiling and client-visible bytes are all unchanged. `streaming_invariant.rs` stays intact (no new full-body reads).

### Should-fix (both reviewers)

1. **#3335's `tail` flag has the identical EOF-only defect and is left in the block this PR just proved never runs.** `publish_entry.tail.store(true, …)` at `proxy_service.rs:2183-2185` / `:2210-2212` sits in the post-loop block; `await_tee_publish` (`:184-195`) returns `false` unless it's set. Before this PR that was harmless (the write always ended in `discard_staging`); this PR is what makes the publish tail real, so it also makes #3335's mitigation dead exactly where it now matters — a concurrent cache read during the publish tail classifies the entry as a miss and re-fetches upstream. One line moved next to the new digest publish fixes it; just mind that `tail` must not be set while the body is still streaming (`:95-101`).

2. **Publish on `== total` rather than `>=`.** After `tee_hasher.take()` (`:2146`) further bytes are forwarded but not hashed; today that can't commit because of `classify_stream_write`, but that's a two-function invariant with nothing local pinning it, and #3605 is open on the same path. An overshoot can never commit anyway, so `==` costs nothing and makes the "no digest over a prefix" argument local. (Or clear the slot on any later byte.)

### Nits
- `:1844-1851` and `:1909-1911` still say the tee digest is published "at upstream EOF" — load-bearing prose for a security gate (GHSA-qxv7-p3mq-88fv).
- The gate's `warn` at `:1928-1934` still can't distinguish "digests differ" from "no digest was published" — the ambiguity that hid #3487 for two releases. Your offer of a distinct message or the `ak_proxy_cache_checksum_mismatch_total` counter is worth taking up.
- Three of the four new tests pass with the fix reverted — they're coverage backfill, which is fine, just not regression tests.

### Adjacent, not this PR
The single-flight leader has the same latent shape: `leader_handle` sends `StreamItem::Done` only after its own loop ends (`proxy_hydration.rs:596-598`), so a consumer stopping at `Content-Length` leaves followers on `RecvError::Closed` → `BadGateway` inside an already-200 body. Untouched here, but 1 + this argue for a `Drop`-based publish covering digest, tail flag and `Done` at once. Also first-hand corroboration of #3607: `streaming_invariant.rs` fails on main today on a pre-existing `npm.rs allowlist=7 actual=6` drift.

### Bookkeeping
No `## [Unreleased]` bullet for #3487 — needed before the cut. I'll push it once the `tail` move is in. Milestoning 1.9.0.

brandonrc:
I owe you an apology and an explanation.

While this PR was open and awaiting the `tail` move I asked for, I ran a fix wave over the remaining 1.9.0 issues without checking which of them already had a contributor PR in flight. #3487 was one of them, and I merged my own fix for it (#3738, commit `136bf68e`) this evening. That should not have happened with your PR sitting here in good shape, and the process gap was mine, not yours.

For the record, what landed is your diagnosis and very nearly your patch: publish the tee digest when the forwarded byte count reaches `Content-Length`, before the last chunk is yielded, so a consumer that stops at the advertised length no longer leaves the gate reading an empty slot. It also carries the `tail` move I asked you for here, and it uses `==` on the byte count exactly as you argued. The three points you raised that I had marked "adjacent" are now filed: the single-flight leader's `StreamItem::Done` shape and the `Drop`-based publish idea are still unowned, and the gate's `warn` ambiguity is unowned as well — I would rather you took either of those than have you spend another cycle here.

Closing this as superseded, with the credit going to you: you found a bug that had silently refused every gated Maven and npm object since 1.7.4, and you were right about the mechanism before anyone else looked. If you want the follow-ups, say so and I will assign them.

travisluecke:
No apology needed, honestly — the bug's fixed and gated objects are caching again, which was the whole point. Shipping the fix that was ready rather than waiting on my PR was the right call; I'd much rather users get unblocked today. And thanks for the writeup — that's more than generous.

Glad the `==` on the byte count and the tail move held up in the version that landed.

On the follow-ups: I'll take the gate's `warn` ambiguity if you'll assign it. That's the thing that cost me the most time tracking this down — an absent digest logging the same "checksum does not match" line as a genuine mismatch. I'd already sketched the two options in the PR (a distinct message for the missing-digest case, or wiring the `ak_proxy_cache_checksum_mismatch_total` counter the 1.7.4 notes say this path skips), so it's a natural one for me to carry. Happy to leave the `StreamItem::Done` shape and the Drop-based publish for others — though if they're still unowned down the line, ping me.

Thanks again, this was a good one to work.

Issue #2665, opened 2026-07-17
Buffered proxy metadata is unbounded across concurrency: ~512x the per-request cap (rpm/debian/npm/composer/maven/pypi)


brandonrc:
## Summary

Every buffered metadata proxy path holds the whole document in memory per in-flight request, with no per-path concurrency or memory bound. Worst-case resident memory is **object size x global concurrency limit**, on an anonymous, unrate-limited path.

Surfaced while reviewing #2623 (RPM 8 MiB -> 128 MiB tier switch). **The tier switch is correct and should not be reverted** — it fixes #2622, and you cannot serve 30 MB of metadata while holding 8 MiB. This issue is the pre-existing, format-wide gap it makes more visible.

## The math

- **Read is well-behaved.** `proxy_service.rs:1783-1797` accumulates frame-by-frame into a `BytesMut` and aborts as soon as `buf.len() + chunk.len() > max`. No eager allocation. Memory = `min(actual body, cap)`.
- **Nothing gates this path.** The existing semaphores (auth/bcrypt `api/mod.rs:159`, scanner extraction, admin rescan, remote_instances) do not cover it. `bounded_archive.rs` provides byte budgets for decompression, not permits. The only bound is the router-wide `GlobalConcurrencyLimitLayer` + load-shed backstop (`routes.rs:194-210`), default **512** (`config.rs:883`).
- **No rate limiting.** The limiters are scoped inside `api_v1_routes`; `/rpm/*` (`routes.rs:71`) only gets `repo_visibility_middleware`.
- **Anonymously reachable** on public repos (`routes.rs:57-58`).
- **Single-flight does not help.** `proxy_service.rs:2589` returns `CacheReadOutcome::Hit` **before** `coordinator.coordinate` at `:2603`. Miss-path fetches coalesce; **cache hits do not.** N concurrent GETs for an already-cached document each buffer it independently.

| | per-request cap | x512 concurrency |
|---|---|---|
| RPM pre-#2623 | 8 MiB | 4 GiB |
| RPM post-#2623 | 128 MiB | 64 GiB |
| Realistic (~30 MB OL8 `filelists.xml.gz`, cached) | 30 MB | **~15 GB** |

The realistic case needs no hostile upstream: once a large real-world repodata file is cached, ~512 concurrent GETs transiently buffer ~15 GB and OOM the pod.

## Affected (all share the constant and the unbounded buffered path)

[...]

PR #3515, opened 2026-08-23
fix(helm): serve the upstream index from a Remote Helm repository

### LINKED ISSUE #3448: Remote Helm repository serves an empty index.yaml (entries: {}), so helm pull cannot resolve any chart through a proxy
## Summary

A **Remote** (proxy) Helm repository serves an `index.yaml` containing `entries: {}`. Since `helm` resolves chart names and versions entirely from that index, `helm pull` cannot find any chart through a proxied repository — `helm repo add` and `helm repo update` both succeed, which makes the failure look like a client or chart-name problem rather than a server one.

## How it was found

While validating proxy download recording (#3446 / #3447) on a live instance. The intent was to drive a real `helm pull` through a Remote repo pointed at `charts.jetstack.io`; `helm repo add` and `helm repo update` succeeded, then no chart could be resolved. Fetching the chart tarball by direct path **worked** (HTTP 200, 80746 bytes), so the artifact-serving path is fine — only the index is empty.

That distinction is the useful part: **serving works, discovery does not.** Anything that resolves a chart by name through the index fails, while anything addressing the tarball path directly succeeds.

## Impact

`helm pull`, `helm install` and `helm upgrade` against a Remote Helm repository are effectively unusable, because all three resolve through the index. Direct-path automation is unaffected, which is likely why this has survived — a scripted deployment pinning a full chart URL would never notice.

## Expected

A Remote Helm repository's `index.yaml` should reflect what the upstream index advertises (proxied, subject to the usual caching and any configured filters), so `helm` can resolve charts by name and version the way it does against the upstream directly.

## Notes for whoever picks this up

- Compare against the PyPI and npm proxy paths, which do serve upstream-derived metadata for Remote repos — the Helm index generator appears to build only from locally-catalogued charts, which for a proxy is empty until something has been cached.
- Worth checking whether a **Virtual** repo containing a Remote Helm member has the same gap.
- If the index is intended to be built from cached content only, then it is not a bug so much as a design that makes proxy Helm repositories non-functional for their normal use — in which case the fix is to proxy or merge the upstream index, and the current behaviour should at minimum be documented.

brandonrc:
## Summary

A Remote (proxy) Helm repository answered `index.yaml` with `entries: {}`. `helm` resolves a chart name and version entirely from that index before it requests any tarball, so `helm repo add` and `helm repo update` both succeeded and `helm pull` / `install` / `upgrade` could then resolve nothing — the failure looked like a bad chart name rather than a server problem. Fetching the same tarball by its full path worked, so serving was fine and only discovery was broken (#3448).

`index_yaml`'s fall-through Remote branch called `query_charts_from_repo`, a purely local artifacts query, and never contacted upstream. The Virtual branch already fetched and merged its Remote members' upstream indexes, which is why a Virtual-with-Remote-member worked and only a repository pointed straight at an upstream did not — confirmed live, both topologies, below.

## Fix

The Remote branch merges the upstream index over whatever the repository holds locally, rewriting every URL to this repository's own `charts/` route so the client comes back through the proxy for the tarball.

**Why merge rather than proxy verbatim, and who wins.** `download_chart` resolves a local artifact row *first* and only falls through to the upstream index otherwise, so the index should advertise exactly the union it can serve. Locally-held entries win their `(name, version)`: that entry carries the artifact's REAL stored filename, which is what the download route resolves, whereas an upstream entry advertises the `{name}-{version}.tgz` convention and could name a path the download route cannot resolve (the same distinction `chart_download_url` already draws). A different version of the same chart is a different entry and is always merged.

**What happens when upstream is unreachable — argued, not silently picked.** Three options were on the table: empty index, hard 502, or stale-if-error.

- *Empty index* is rejected outright. `entries: {}` is indistinguishable from "this upstream publishes no charts", so `helm repo update` reports success and the operator is told nothing. That is the defect being fixed; reproducing it in the failure path would be the same bug wearing a different hat.
- *With charts held locally*, they are served and a warning is logged. An upstream outage must not stop a chart the repository already has from resolving; that is the point of a cache. This is the handler-level half of stale-if-error.
- *With nothing held*, the upstream's own error is surfaced (502/503 as `map_proxy_error` classifies it). There is no honest index to serve, and a loud failure is strictly better than a success that resolves nothing.

The proxy cache already applies RFC 5861 `stale-if-error` to the `index.yaml` object itself, so a warm proxy keeps serving the last good upstream index before the handler-level fallback is reached at all — the fallback only covers a cold or long-dead proxy.

**Byte ceiling.** The upstream index fetch is factored out (`fetch_upstream_index`) and shared with the chart-download path, which resolves real chart URLs from the same document. It moves from `DEFAULT_METADATA_MAX_BYTES` (8 MiB) to `LARGE_METADATA_MAX_BYTES`, and now reserves against the shared buffered-metadata budget (#2665/#2684), which the previous unbudgeted read did not. A Helm index is the whole-repository listing — the direct analogue of the PyPI simple-index and npm packument documents already on the larger ceiling — and `charts.bitnami.com/bitnami/index.yaml` is **27,282,002 bytes** today (`curl -sI`), so the 8 MiB default would have 502'd exactly the repositories this fix exists for, on both the index and the download path. What that reservation does and does not bound is measured below rather than asserted.

The Virtual branch's inline merge loop is replaced by a call to the same helper, so there is one merge path. Its output is unchanged for the existing single-remote-member case (the vector is empty at that point); with two members advertising the same chart+version it now emits one entry instead of two, which is what `helm` resolves to either way.

## Reproduction and live verification

Deployed on an isolated pool slot (`:8102`), never `:8080`. Baseline is `ghcr.io/…/artifact-keeper-backend:dev`; patched is a Docker build of this branch. Remote repo → `https://charts.jetstack.io`.

**Baseline (unpatched), real `helm` client:**

```
$ curl -s .../helm/helm-remote/index.yaml
apiVersion: v1
generated: 2026-08-23T01:54:59Z
entries: {}

$ helm repo add akbase http://127.0.0.1:8103/helm/helm-remote -u admin -p ***
"akbase" has been added to your repositories
$ helm repo update akbase
Update Complete. ⎈Happy Helming!⎈
$ helm pull akbase/cert-manager --version v1.16.2
Error: chart "cert-manager" matching v1.16.2 not found in akbase index. (try 'helm repo update'): no chart name found

# but the tarball by direct path is fine — serving works, discovery does not
$ curl -o c.tgz .../helm/helm-remote/charts/cert-manager-v1.16.2.tgz
status=200 size=136961
```

The reporter's open question, answered on the same instance: a **Virtual** repo with that Remote as a member returned the full upstream index (11,877 lines) on the SAME baseline build. Only the direct Remote was broken.

**Patched:**

```
$ curl -s .../helm/helm-remote/index.yaml   # 12 charts, 486 versions
sample url: ['/helm/helm-remote/charts/cert-manager-v1.21.1.tgz']

$ helm repo add akproxy http://127.0.0.1:8102/helm/helm-remote -u admin -p ***
$ helm repo update akproxy
...Successfully got an update from the "akproxy" chart repository
$ helm search repo akproxy/cert-manager --versions
akproxy/cert-manager   v1.21.1   v1.21.1   A Helm chart for cert-manager
akproxy/cert-manager   v1.21.0   v1.21.0   A Helm chart for cert-manager
...
$ helm pull akproxy/cert-manager --version v1.16.2
$ ls -l
-rw-r--r-- 136961 cert-manager-v1.16.2.tgz
```

**Live regression checks on the patched build:**

| check | result |
|---|---|
| Virtual repo with the Remote as member | 200, 12 charts / 486 versions — unchanged |
| Hosted (local) repo: `cm-push` upload then index | `201 {"saved":true}`, chart advertised — unchanged |
| Remote whose upstream host does not resolve, nothing held | **502** `Failed to fetch from upstream …` — not a silent empty 200 |
| Remote whose `upstream_url` path is wrong (upstream 404), nothing held | **404** `Artifact not found upstream` |

## Adversarial review follow-up (rev 2)

Every point below was raised on `f7cc2c2c` and is addressed here. Where a finding's *diagnosis* did not survive measurement, that is stated rather than papered over.

### G-A (blocker) — the budget permit outlived by the largest allocations. Fixed, and the claim corrected to what is measurable.

The finding is structurally correct: `fetch_upstream_index` bound `_budget_permit` locally, so the parsed tree, the cloned chart vector and the ~27 MB rendered document all lived outside the reservation.

**Fixed.** `fetch_upstream_index` now *returns* the permit. `index_yaml` holds it across the merge and hands it to `build_index_response`, which rides it on the response body so it is released only after the rendered document reaches the response writer. The chart-download path drops it explicitly once it has copied out the one URL it needs. The body-riding helper is exactly the RPM precedent the review cited, so rather than re-derive it, `rpm.rs::metadata_body_stream` is **extracted into `proxy_helpers::budgeted_body`** and both callers share it — rpm's existing `buffered_metadata_response_holds_budget_until_body_dropped` test is unchanged and still passes, which is the proof the extraction is behaviour-preserving.

**But the measurement does not support the attribution, and I am not going to ship a claim that says it does.** I rebuilt the counterfactual (permit released just before the body is built, everything else identical) and measured both against a 27 MB upstream index, restarting the backend before each run so the baseline is cold rather than settled:

| concurrent index GETs | this PR (permit rides the body) | counterfactual (released early) |
|---|---|---|
| 16 | baseline 0.186 → peak 2.122 GiB (**+1.94**) | 0.179 → 2.303 GiB (**+2.12**) |
| 64 | 0.186 → 2.529 GiB (**+2.34**) | 0.187 → 2.631 GiB (**+2.45**) |
| 256 | 0.187 → 2.668 GiB (**+2.48**) | 0.186 → 2.652 GiB (**+2.47**) |
| control: 450 KB index, 64 | 2.711 → 2.719 GiB (**+0.008**) | — |

Two things follow, and they are not the same thing:

1. **Peak stops scaling with connection count.** A sixteenfold increase in connections (16 → 256) costs 28% more memory. That is the reservation's admission control working — `reserve()` awaits, so only eight buffered index requests are ever in flight process-wide. The control run reproduces the review's observation that growth tracks index size, not connection count.
2. **The permit's lifetime is not the dominant term at these concurrencies.** Both variants plateau at the same level, so the memory the review measured is accounted for by the eight *admitted* requests, not by the unreserved tail. The real gap is that the reservation is sized to the 128 MiB **ceiling** rather than the body, so a 27 MB index whose working set (buffer + parsed tree + cloned vector + rendered document) exceeds 128 MiB is under-charged. Eight of those peak above the 1 GiB budget.

So the sentence the review objected to is gone. The CHANGELOG now says the budget caps **concurrency, not bytes**, quotes the numbers above, and names the reason. Closing the byte gap would mean the whole-request working-set reservation `MetadataWorkingSetLimits` provides — a change to the shared budget's contract that would also take reservation share from RPM/npm/PyPI/debian/composer/conda — so it is deliberately not made in this PR. I kept the lifetime fix regardless: it costs nothing, it is the correct shape, and it removes the window even though the window was not what the numbers were showing.

New test `test_index_response_holds_the_budget_until_the_body_is_dropped_3448` asserts the reservation survives the merge and the render and is released only when the response is dropped. Revert-proof — release the permit before building the body:
```
FAIL test_index_response_holds_the_budget_until_the_body_is_dropped_3448
  budget must stay debited while the rendered response body is alive
  left: 4096   right: 3096
```

### G-B — the chart-download path's new budget participation, disclosed.

Correct, and it was under-described: `fetch_chart_via_index` went from unbudgeted-8 MiB to budgeted-128 MiB, so its `index.yaml` lookup now waits on a reservation shared with RPM repodata, npm packument, PyPI simple, debian, composer and conda — eight slots at 1 GiB / 128 MiB. `reserve()` awaits, so this is head-of-line blocking under concurrent metadata load, not failure, and it is the hotter of the two paths. Now stated explicitly in both the PR text above and the CHANGELOG as a deliberate trade of tail latency for participation in a bound the path was previously outside.

### G-D — factual error, corrected; test widened.

Confirmed on the rig, and there are three statuses rather than two:

```
upstream .../does-not-exist-xyz   -> 404  "Artifact not found upstream"
upstream host does not resolve    -> 502  "Failed to fetch from upstream …"
upstream 5xx                      -> 503  (map_proxy_error)
```

My original "nothing held ⇒ 502" was true only for the DNS case I happened to test. The live table above and the CHANGELOG now name all three, and say why the difference is worth keeping: it lets an operator tell a wrong URL from an unreachable host from an upstream that is merely down. `test_remote_index_yaml_surfaces_an_upstream_failure_when_nothing_is_held_3448` is now table-driven over `(503 → 503, 404 → 404)`, asserting both `!= 200` (the policy) and the exact status (the classification), so the most common misconfiguration shape is covered.

### G-C — fixed rather than noted.

Agreed the amplification is mine: the same strict parser only cost one chart download before this PR, and now costs an entire repository's discovery. `urls`, `created`, `digest` and `generated` take `#[serde(default)]`. `name` and `version` stay required — an entry without them cannot be advertised at all. Defaulting invents nothing: a missing `urls` yields an empty list, which the download path already answers 404 for, and the proxied index rebuilds URLs from name/version anyway.

Revert-proof (defaults removed, test kept):
```
FAIL test_upstream_index_entry_missing_optional_fields_still_parses
  Error("entries.k[0]: missing field `urls`", line: 4, column: 7)
```
plus `test_upstream_index_entry_without_a_name_is_still_rejected` as the negative control.

### G-E — fixed, but not by encoding.

I tried percent-encoding first and my own test caught it as a regression: `urlencoding::encode` is `quote(safe="")`, so `1.0.0-rc.1+build.5` became `1.0.0-rc.1%2Bbuild.5` and every legitimate semver-build-metadata chart URL moved. An encoder weak enough to leave `+` alone would also have to leave `%` alone, which defeats the point since the router percent-decodes before matching.

Instead, `merge_upstream_index_entries` **drops** upstream entries whose `name`/`version` cannot form a single path segment (`/`, `\`, `?`, `#`, `%`, control chars, empty, `.`/`..`), with a warning. That is strictly stronger than encoding: rather than making a hostile name inert, it restores the invariant outright — everything the proxied index advertises resolves back to the entry it came from. It costs nothing real, because an entry that cannot be addressed could not have been downloaded through this repository anyway. `chart_download_url` is left byte-identical, so no hosted URL moves.

Revert-proof (filter removed, test kept):
```
FAIL test_merge_drops_upstream_entries_that_cannot_be_addressed
  the unaddressable entry must be omitted, got
  [(… name: "../../../../etc/passwd" …, "/helm/helm-remote/charts/../../../../etc/passwd-1.0.0.tgz", …)]
```
with `test_ordinary_chart_names_and_versions_stay_advertisable` pinning `+build.5` and friends as unaffected.

### G-F / G-G / G-H — acknowledged, disclosed, not changed.

- **Dedup key vs download resolution.** The merge dedups on `(chart.name, chart.version)` while `download_chart` resolves by filename suffix. Different key spaces, agreed. Not reachable on a Remote, where writes are 405'd, so no local row can be created with a filename that disagrees with its name/version. Left alone deliberately: aligning them would mean re-deriving the filename for the dedup key, which reintroduces the very `{name}-{version}.tgz` assumption `chart_download_url` exists to avoid.
- **How local rows reach a Remote repository at all.** Worth saying plainly, since it bounds how often the stale-if-error branch matters: direct upload to a Remote is rejected, so `test_remote_index_yaml_still_serves_locally_held_charts_when_upstream_is_down_3448` covers rows that arrive by promotion, peer replication, migration import, or from a repository that was converted. Real, but not the common case; the common case is the upstream branch.
- **`generate_index_yaml` regroups by `chart.name`, discarding the upstream map key.** Confirmed pre-existing and unchanged by this PR — the Virtual branch has always fed it upstream-derived entries the same way. An upstream can file a chart under key `innocent` with inner `name: popular-chart` and it is re-advertised under the inner name. Not fixed here because it is a property of the shared generator and would change the Virtual path's output too; it belongs in its own issue rather than smuggled into this one.

### Also in rev 2

Rebased onto `main` (`20820369`) after #3516 landed; the CHANGELOG conflict is resolved with both bullets present, and the Helm bullet verified at line 41 — under `## [Unreleased]` (line 8), above `## [1.8.1]` (line 114).

One correction to the review's "verified in your favour" list, in the interest of not banking credit I did not earn: the note that my new tests use `router_anon(super::router(), …)` **without** `repo_visibility_middleware` is right, and they still do. They are index-shape tests, not authorization tests; the authorization guarantee rests on the reviewer's live check and on the existing `virtual_member_authz_tests`, not on anything this PR adds.

Five new `--lib` tests (CI's runner; DB-backed ones run for real under `AK_TESTS_REQUIRE_DB=1`, 0.28–0.29s each, so they are not the ~0.04s silent-skip shape):

- `test_merge_upstream_index_entries_rewrites_urls_to_this_repo` (pure)
- `test_merge_upstream_index_entries_keeps_the_locally_held_entry` (pure) — the locally-held chart deliberately has a filename that is *not* the `{name}-{version}.tgz` convention, so an overwrite would be visible
- `test_remote_index_yaml_serves_upstream_entries_3448` — wiremock upstream, driven through the real router
- `test_remote_index_yaml_surfaces_an_upstream_failure_when_nothing_is_held_3448`
- `test_remote_index_yaml_still_serves_locally_held_charts_when_upstream_is_down_3448`

Revert-proof — production hunk (the Remote branch in `index_yaml`) deleted, tests kept:

```
FAIL test_remote_index_yaml_serves_upstream_entries_3448
  a Remote repository's index must carry the upstream's charts; an empty `entries`
  is what makes `helm pull` unable to resolve anything (#3448)
FAIL test_remote_index_yaml_surfaces_an_upstream_failure_when_nothing_is_held_3448
  with an unreachable upstream and nothing held locally there is no honest index to
  serve … Got 200 OK
PASS test_remote_index_yaml_still_serves_locally_held_charts_when_upstream_is_down_3448
```

The third passing under the revert is expected and stated deliberately: it is the no-regression control for the pre-existing local-only behaviour, not a revert-proof. The two that flip are.

## Test Checklist
- [x] Manually tested locally

- `cargo nextest run --workspace --lib --test-threads 8`: **15814 passed, 0 failed, 8 skipped**
- `cargo fmt --check`, `cargo clippy --lib --tests -- -D warnings`: clean
- `jscpd --min-lines 10 --threshold 3` on the changed file: no clones reported

## API Changes
- [x] N/A - no API changes

Closes #3448


brandonrc:
Pushed rev 2 (`514796da`), rebased onto `main` after #3516 landed. Thanks — G-A was a real defect and the extraction it prompted is a genuine improvement. Two things in the review did not survive measurement, and I have written both up rather than quietly shipping around them.

**G-A is fixed, but the numbers do not support the attribution.** The permit now rides the response body (`proxy_helpers::budgeted_body`, extracted from `rpm.rs` so the two callers share one implementation — rpm's existing budget test is unchanged and still passes). I then built the counterfactual and measured both, cold-restarting the backend so the baseline is not a settled one:

| concurrent index GETs (27 MB index) | this PR | counterfactual (permit released early) |
|---|---|---|
| 16 | +1.94 GiB | +2.12 GiB |
| 64 | +2.34 GiB | +2.45 GiB |
| 256 | +2.48 GiB | +2.47 GiB |
| 450 KB control, 64 | +0.008 GiB | — |

Both plateau at the same level. 16 → 256 connections is a sixteenfold increase for 28% more memory, so the reservation *is* bounding things — but by **concurrency**, not by bytes, and the memory you measured is accounted for by the eight *admitted* requests rather than by the unreserved tail. The actual gap is that the reservation is sized to the 128 MiB ceiling, not the body, so a 27 MB index whose working set exceeds 128 MiB is under-charged; eight of those peak above the 1 GiB budget. Closing that needs `MetadataWorkingSetLimits`-style whole-request reservation, which changes the shared budget's contract and would take share from RPM/npm/PyPI/debian/composer/conda — deliberately not done here.

So the offending sentence is gone. The CHANGELOG now states the concurrency bound, quotes the table, and names the limitation. I kept the lifetime fix anyway: it costs nothing and it is the correct shape.

**G-E: I tried encoding first and my own test caught it as a regression.** `urlencoding::encode` is `quote(safe="")`, so `1.0.0-rc.1+build.5` became `…%2Bbuild.5` — every legitimate semver-build-metadata URL moved. Any encoder weak enough to spare `+` also has to spare `%`, which defeats the purpose since the router decodes before matching. Rev 2 **drops** unaddressable upstream entries instead. That is stronger than encoding: instead of making a hostile name inert it restores the invariant outright — everything the proxied index advertises resolves back to the entry it came from — and `chart_download_url` stays byte-identical, so no hosted URL moves.

**G-B** disclosed explicitly (chart-download path moved from unbudgeted-8 MiB to budgeted-128 MiB; `reserve()` awaits, so head-of-line blocking on the hotter path, traded for a bound it was previously outside). **G-D** corrected — there are three statuses, not two: upstream 404 → 404, connect/DNS failure → 502, upstream 5xx → 503; the test is now table-driven over 503 and 404 and asserts both `!= 200` and the exact status. **G-C** fixed rather than noted, since the amplification is mine — the strict parser cost one chart download before this PR and an entire repository's discovery after it. **G-F/G-G/G-H** acknowledged in the PR body with the reasoning for leaving each alone; G-H in particular is a property of the shared generator that also affects the Virtual path, so it belongs in its own issue rather than being smuggled into this one.

Four new revert-proofs, each with the production hunk removed in isolation and the test kept: budget-lifetime, serde defaults, unaddressable-entry filter, and the pre-existing empty-index one. Full suite `15821 passed, 0 failed`; `helm pull` of the 136,961-byte chart re-verified on the rev-2 build.

One correction to the "verified in your favour" list: you are right that my tests use `router_anon` without `repo_visibility_middleware`, and they still do. They are index-shape tests. The authorization guarantee rests on your live check and the existing `virtual_member_authz_tests`, not on anything this PR adds — I did not want that credited to this diff.

Issue #2824, opened 2026-07-22
docker-publish cosign step fails → 1.6.1 image unsigned; root-cause before 1.6.2 tag


brandonrc:
The 1.6.1 backend image published but the cosign keyless signing step failed; the image is UNSIGNED and can't be re-signed (the 'refuse to overwrite a published version' guard blocks a job re-run against the immutable tag). Root-cause the cosign step (id-token perms / keyless OIDC / Fulcio-Rekor / registry-auth ordering) on main + cherry-pick to release/1.6.x so the 1.6.2 tag build signs cleanly.

brandonrc:
Investigation: the docker-publish manifest job **does** grant `id-token: write` (docker-publish.yml:694, same job as the cosign step at :805) and `packages: write` for the signature push, so the keyless cosign OIDC→Fulcio→Rekor path is correctly wired. The 1.6.1 sign failure was therefore most likely a **transient sigstore outage**, not a config defect (its original step log was overwritten when I re-ran the failed jobs). 

Disposition: **downgraded from release-blocker to watch-item.** No workflow change needed pre-1.6.2. Expectation: the 1.6.2 `v*` build signs cleanly; if the cosign step fails again we'll capture the fresh error (do NOT re-run the whole job — the 'refuse to overwrite' guard blocks it against an immutable tag; instead diagnose from the first-run log). 1.6.1's image stays unsigned (it's a short-lived patch superseded by 1.6.2); re-signing it would need an out-of-band `cosign sign` with a signing identity.

PR #3518, opened 2026-08-23
fix(ci): make provenance attestation fail the publish job instead of vanishing

### LINKED ISSUE #3496: docker-publish: 6 step-level continue-on-error on provenance attestation and VEX means an image publishes green even when its attestation failed
## Summary

`.github/workflows/docker-publish.yml` carries **six STEP-level `continue-on-error: true`** on supply-chain steps. Step-level (unlike job-level) leaves `needs.<job>.result` green, so if provenance attestation or VEX attachment fails, **the image publishes anyway, the job is green, and nothing reports it.**

## The six steps

| line | job | step |
|---|---|---|
| 1288 | `merge-backend` | Generate artifact attestation (`actions/attest-build-provenance@…v4`, `push-to-registry: true`) |
| 1301 | `merge-backend` | Attach VEX attestations (`docker scout attestation add`) |
| 1510 | `merge-openscap` | Generate artifact attestation |
| 1874 | `merge-scanner-adapter` | Generate artifact attestation |
| 1985 | `merge-backend-alpine` | Generate artifact attestation |
| 2002 | `merge-backend-alpine` | Attach VEX attestations |

That is provenance attestation for **all four published images** (backend, OpenSCAP, scanner-adapter, backend-alpine) plus both VEX attachment steps.

## Correction to an earlier reading of this file

The adjacent `Sign image with cosign (keyless)` steps (lines 1277, 1499, 1863, 1974) are **not** `continue-on-error` — they are bare `run:` steps and a cosign failure does fail the job. That was fixed in #2824 after the 1.6.1 image shipped unsigned. Only attestation and VEX are still soft.

So the accurate claim is narrower than "signing has never been proven to work", but it is still bad: **build provenance has never been proven to be attached to any published image.** An `attest-build-provenance` failure (OIDC token unavailable, registry push rejected, action outage, rate limit) is indistinguishable from success in the run summary.

## Why step-level is the wrong tool here

This is the same shape as the `resilience-tests` problem: a STEP-level `continue-on-error` keeps the job's own result green, so downstream `needs:` and branch protection cannot see it. Job-level `continue-on-error` at least surfaces as a non-success job result. Neither is right for a control that is supposed to be a supply-chain guarantee.

There is also no post-publish verification that would catch it: nothing in the workflow runs `gh attestation verify` / `cosign verify-attestation` against the published digest, so a silently-missing provenance predicate survives indefinitely.

## Suggested direction

1. Decide per step whether it is a **control** or **best-effort telemetry**, and write the answer into a comment.
   - Provenance attestation is a control. Remove `continue-on-error` and let it fail the publish, or gate it behind an explicit, named, documented condition.
   - VEX attachment via `docker scout` depends on an external service and may legitimately be best-effort — but then it should emit a `::warning::` and a step-summary line so a failure is visible, rather than vanishing.
2. Add a verification step to the tag path: after the merge jobs, verify the published digest actually carries a provenance attestation (and its cosign signature), and hard-fail if not. The `verify-published` job at the bottom of the file is the natural home.
3. Because the control has been soft since it was introduced, treat existing published images as unverified until checked — worth a one-off audit of the current `:latest` and the most recent release tags.

## Related

- #2824 — cosign step failed and 1.6.1 shipped unsigned; that one was caught only because someone looked.
- #3494, #3495, #3437 — same family: controls that cannot fail.

Found while verifying the 1.8.2 milestone; line numbers re-verified against `main` @ `a9f3876` on 2026-08-22.

brandonrc:
## Summary

Closes #3496.

`docker-publish.yml` carried **six STEP-level `continue-on-error: true`** on supply-chain steps: `Generate artifact attestation` on all four published images (backend, OpenSCAP, scanner-adapter, backend-alpine) and `Attach VEX attestations` on the two that carry VEX. Step-level is the worst available shape for a control — unlike job-level it leaves the **job** result `success`, so `needs.<job>.result`, required checks and every downstream gate see green, nothing lands in the run summary, and no annotation is raised. An OIDC-token outage, a registry rejection, a rate limit or an `actions/attest-build-provenance` outage was indistinguishable from success. **Build provenance had therefore never been proven to attach to any published image.**

Confirming the issue's own correction: the four `Sign image with cosign (keyless)` steps are bare `run:` blocks with no `continue-on-error` and do hard-fail the job (#2824). They are untouched here.

### The decision, and why

**Provenance attestation → hard-fail.** The usual argument against is "an attestation hiccup should not block a release". That argument does not apply at this position in the job: the attestation step runs *after* `Apply ghcr tags` and `Copy to Docker Hub`, so the image is already on both registries by the time it runs. Hard-failing does not prevent a publish — it converts a silent supply-chain gap into a visible red job. The recovery is `Re-run failed jobs`, which is cheap here because every step in these jobs is idempotent by construction: `imagetools create` re-points the *same* digest, `cosign sign` and `attest` are safe to repeat. That idempotence is what makes hard-fail affordable, and it is now written down next to the step.

**VEX attachment → best-effort, but loud.** `docker scout attestation add` depends on an external SaaS and on Docker Hub credentials; an outage there is not a reason to redden a release. What was wrong was not the softness but that the softness was invisible. The decision now lives in the script rather than in a YAML key: failures are collected per file and reported with a `::warning::` annotation and a `$GITHUB_STEP_SUMMARY` section naming each VEX document that did not attach, and the step still exits 0. `docker-publish.yml` now carries **no `continue-on-error` at all**.

### Making the fix falsifiable

`scripts/ci/check-supply-chain-soft-fail.sh` (new, wired into `shell-tests`) pins two invariants and a non-vacuity clause:

1. no attestation (`actions/attest-*`) or cosign signing/attesting step anywhere in `.github/workflows/` may set `continue-on-error` to anything but a literal `false` — an expression that *could* evaluate true counts as soft;
2. `docker-publish.yml` may carry no `continue-on-error` on any step **or** job;
3. if no attestation step and no cosign step exists at all, the check **fails** rather than passing vacuously — that being the exact defect this PR is about.

The job-level and step-level `continue-on-error` elsewhere in the repo (the two documented e2e waivers, the nightly issue-filing step) are deliberately out of scope and unaffected.

**The gate fails on the pre-fix workflow.** Run against `main`'s `docker-publish.yml`:

```
$ git show origin/main:.github/workflows/docker-publish.yml > /tmp/mainwf/docker-publish.yml
$ ./scripts/ci/check-supply-chain-soft-fail.sh /tmp/mainwf
ERROR: supply-chain step(s) can fail without failing the build:

  - docker-publish.yml :: merge-backend :: Generate artifact attestation
  - docker-publish.yml :: merge-backend :: Attach VEX attestations
  - docker-publish.yml :: merge-openscap :: Generate artifact attestation
  - docker-publish.yml :: merge-scanner-adapter :: Generate artifact attestation
  - docker-publish.yml :: merge-backend-alpine :: Generate artifact attestation
  - docker-publish.yml :: merge-backend-alpine :: Attach VEX attestations
exit 1
```

All six reported steps, by name. On this branch: `OK: 9 supply-chain step(s) hard-fail their job; no soft-failed steps in docker-publish.yml.`

**The gate's own self-test proves it can fail** (`scripts/ci/test-check-supply-chain-soft-fail.sh`, 10 cases, all from the failing side): soft attestation, soft VEX, an unrelated soft step in the publish workflow, an expression-valued `continue-on-error`, a soft cosign step copied into another workflow, job-level `continue-on-error`, a workflow set with no supply-chain step (vacuity), and a missing directory — plus two accept cases (`continue-on-error: false`, and the corrected shape).

**The self-test was revert-proofed** by breaking the guard three ways and confirming the self-test goes red each time:

| deliberate breakage | self-test result |
|---|---|
| `is_soft()` always returns `False` | 6 of 10 cases FAIL |
| non-vacuity clause removed | 1 case FAILS (`no supply-chain step is rejected`) |
| soft test narrowed to a literal `== "true"` | 1 case FAILS (`expression-valued continue-on-error`) |

restored → `passed: 10   failed: 0`.

**The rewritten VEX step was executed locally** with a stub `docker` on the PATH, driving the exact script extracted from the merged YAML:

- docker succeeds → exit 0, no annotation, empty step summary;
- docker fails → exit 0, `::warning title=VEX attestation not attached::… failed for .vex/a.vex.json .vex/b.vex.json …`, and a `### VEX attestations NOT attached` summary section listing both files;
- no `.vex/*.json` present → exit 0, no output.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

## Deliberately not in this PR

Two follow-ups the issue raises are left out on purpose, and tracked separately in **#3519**:

- **Post-publish verification.** Nothing verifies that a published digest actually carries the provenance predicate — `verify-published` probes tag existence only. Adding an unvalidated `gh attestation verify` hard gate to the tag path cannot be demonstrated to work before a real cut, and shipping a control I cannot show failing would repeat the mistake this PR fixes.
- **Audit of already-published digests.** Because the control has been soft since it was introduced, existing images should be treated as unverified until checked.


brandonrc:
Merge note: #3523 adds a step to the same `shell-tests` list, immediately after `Check the docs/CI drift gate catches what it must`. Whichever lands second will need a trivial additive rebase there — no semantic overlap.

brandonrc:
### Live CI proof that the new gate blocks a regression

Two throwaway commits on this branch (`d4e1…` re-adding a soft-fail, then its revert) exist to demonstrate this on the runner rather than only locally. The diff against `main` is unchanged by them.

Re-adding a single `continue-on-error: true` to `merge-openscap`'s `Generate artifact attestation` — one of the six steps this PR removes — turned `🐚 Shell Tests` red on [run 32653826374](https://github.com/artifact-keeper/artifact-keeper/actions/runs/32653826374/job/97229415318):

```
12 Check supply-chain steps cannot fail silently -> failure   (exit 1)
13 Check the supply-chain soft-fail gate catches what it must -> skipped
```

with the guard naming the step:

```
docker-publish.yml :: merge-openscap :: Generate artifact attestation
    is a supply-chain control — attestation step (`uses: actions/attest-build-provenance`) — and sets
    `continue-on-error`. A STEP-level continue-on-error leaves the JOB result `success`, ...
```

Reverted; `Shell Tests` is green again and the guard reports `OK: 9 supply-chain step(s) hard-fail their job`.

**What is *not* demonstrated live, stated plainly:** that an actual `attest-build-provenance` failure now reddens `merge-backend`. That is GitHub Actions step semantics (a failing step with no `continue-on-error` fails its job) and cannot be exercised without a real tag publish. The falsifiable artifact this PR ships is the static guard above, plus the fact that the same guard reports all six pre-fix steps when pointed at `main`'s copy of the file.

Issue #3122, opened 2026-08-04
CVE gate fails on ANY severity despite declaring CRITICAL,HIGH — trivy-action unsets TRIVY_SEVERITY for SARIF


brandonrc:
## Summary

The container CVE gate in `docker-publish.yml` does **not** do what it says. It declares `severity: 'CRITICAL,HIGH'` and its step summary reports *"Scope: CRITICAL and HIGH severity, fixed CVEs only"* — but it actually fails on a fixable CVE of **any** severity, including LOW.

This is not theoretical: #3121 is a **LOW** CVE (`security-severity: 3.1`, SARIF level `note`) that has blocked all image publishing from `main` for days.

## Root cause

`aquasecurity/trivy-action@ed142fd` (v0.36.0), `entrypoint.sh:74-82`:

```sh
# Handle SARIF
if [ "${TRIVY_FORMAT:-}" = "sarif" ]; then
  if [ "${INPUT_LIMIT_SEVERITIES_FOR_SARIF:-false,,}" != "true" ]; then
    echo "Building SARIF report with all severities"
    unset TRIVY_SEVERITY          # <-- severity input silently discarded
  else
    echo "Building SARIF report"
  fi
fi
```

All four scan steps use `format: sarif` and do not set `limit-severities-for-sarif`, so `TRIVY_SEVERITY` is unset on every one of them. `exit-code: '1'` then fires on any fixable finding at any severity. The failing job log contains the giveaway line verbatim: `Building SARIF report with all severities`.

## Evidence this has already cost the team

`.trivyignore` (currently ~line 110) documents a previous encounter with this exact behaviour, misdiagnosed:

> *Note on severity, learned the hard way on the v1.6.4 cut: this CVE exports to SARIF as UNKNOWN/note, so it looks non-gating in the code-scanning UI. It DOES gate. Trivy filters on the vendor severity, not the NVD one … under `severity: CRITICAL,HIGH` this entry still failed the scan.*

The conclusion drawn there — that Trivy filters on vendor severity — is wrong. The severity filter was never applied at all. Entries have since been accumulating in `.trivyignore` to work around a gate nobody realised was misconfigured, which means low-severity findings are being papered over with reachability arguments rather than either fixed or consciously accepted.

## Why it matters

1. **Wrong blast radius.** Any fixable LOW/MEDIUM CVE in any of the three images halts all publication, with no signal that the gate is stricter than documented.
[...]

Issue #2609, opened 2026-07-16
Trivy container scans report but do not enforce; make the stated policy real


brandonrc:
Split out of PR #2604 review. The summary text was corrected there; the scan behaviour deliberately was not touched.

## What happens

All Trivy scans in `.github/workflows/docker-publish.yml` run with `exit-code: 0`, so findings never fail the job. The step summary claimed:

> **Policy**: Fail on CRITICAL or HIGH severity CVEs

which the workflow does not do. PR #2604 corrected the summary to say reporting-only, because a false claim in a security summary is the same class of problem that PR exists to fix. It did not change enforcement — that is a real policy decision with a real blast radius and deserves its own change.

Also worth noting: `merge-backend` does not `need` `scan-containers`, so images publish regardless of scan outcome.

## Direction

Decide and then implement one of:

1. Enforce: set `exit-code: 1` for CRITICAL/HIGH with `ignore-unfixed: true`, keep `.trivyignore` as the documented exception path, and make the merge jobs depend on the scan.
2. Keep reporting-only and rely on code scanning + VEX, with the summary continuing to say so plainly.

Option 1 needs a burndown of what currently fires, or the first release after it lands will be blocked by a backlog.

## Acceptance

- [ ] Decision recorded
- [ ] Workflow behaviour and summary text agree
- [ ] If enforcing: publication depends on the scan result

PR #3520, opened 2026-08-23
fix(ci): teach the #2609 scan guard the two-pass Trivy shape and actually run it

### LINKED ISSUE #3437: workflow_scan_gate_tests fails on main: the #2609 Trivy guard cannot express docker-publish's two-pass scan
## Summary

`cargo test --test workflow_scan_gate_tests` **fails on pristine `main`**:

```
trivy_scan_steps_enforce_via_exit_code
  scan-containers / Trivy SARIF - Backend: Trivy must run with exit-code '1' ... (#2609)
  left: Some("0")  right: Some("1")
```

Verified by stashing all local changes and re-running against unmodified `main`, so this is not branch-local.

## Cause

`docker-publish.yml` deliberately runs Trivy in **two passes** per image (documented in the workflow around lines 645-663):

1. a **SARIF visibility** pass at `exit-code: '0'`, which uploads findings without blocking, and
2. an **enforcement** pass at `exit-code: '1'`, which is what actually gates the publish.

The #2609 guard predates that split and assumes **one Trivy step per image**, so it sees the visibility pass and reports the gate as disabled. The workflow is correct; the guard can no longer express what it does.

## Why it stayed invisible

It is an **integration** test, so Tier 1 (the per-PR suite, `cargo test --workspace --lib`) never runs it. PRs stay green while `cargo test --workspace` — the Tier 2 gate for `main` and release branches — is red. A guard that is broken where nobody looks is worse than no guard: #2609 exists to stop image publishing from silently becoming non-blocking, and right now it would not catch that regression.

This is the same species as #3422 and #3433: a gate that quietly stopped matching reality and produced no signal where anyone was watching.

## Decision needed before fixing

Two defensible shapes, and the choice is a judgement call rather than mechanical:

1. **Scan only enforcement-pass steps** — filter to steps whose `exit-code` is `'1'` and assert each image has one. Simple, but a workflow that accidentally dropped its enforcement pass entirely would pass vacuously.
2. **Assert the pair** — for each image, require exactly one visibility pass and one enforcement pass. Stronger and matches the workflow's actual contract, but couples the guard to the two-pass design.

Option 2 looks right, with the non-vacuity concern in option 1 being the deciding argument. Whoever takes it should also confirm no other guard in that file has drifted the same way.

## Also worth considering

Tier 1 not running integration tests is the structural reason this hid. Worth asking whether the workflow-scanning guards specifically — which are pure file parsing, need no database, and run in milliseconds — belong in the per-PR suite rather than behind the Tier 2 boundary.

Found while hardening the release pipeline (#3436).

brandonrc:
## Summary

Closes #3437.

Two independent defects in `backend/tests/workflow_scan_gate_tests.rs`, and the #2609 guard was worthless for both reasons at once.

### 1. The test was wrong

It asserted that **every** `aquasecurity/trivy-action` step in `docker-publish.yml` carries `exit-code: '1'`. #3122 deliberately split the scan into two passes per image, documented in the workflow at lines 645-663:

- a **visibility** pass (`format: sarif`, `exit-code: '0'`, no `severity:`) that feeds the Security tab at all severities and never blocks — it declares no severity on purpose, because `trivy-action` runs `unset TRIVY_SEVERITY` for SARIF and writing one there only documents a filter that does not exist;
- an **enforcement** pass (`format: table`, `severity: CRITICAL,HIGH`, `exit-code: '1'`) which is the control the `merge-*` jobs depend on.

Three of the guard's four assertions only ever made sense on the enforcement pass, so from #3122 onward it failed on the visibility pass — reporting a working gate as disabled.

The guard now models the **pair**, which is the stronger of the two shapes the issue proposed. Filtering to enforcement-pass steps and checking only those would accept a workflow that had silently lost its enforcement pass altogether, which is exactly the regression #2609 exists to catch. Concretely, per scanned image (grouped by `image-ref`): exactly one visibility pass and exactly one enforcement pass; the enforcement pass must have `exit-code: '1'`, `severity: CRITICAL,HIGH`, `ignore-unfixed: true`, `trivyignores: .trivyignore` **and a non-SARIF format** (new — it is what keeps the severity filter alive, per #3122); the visibility pass must be SARIF, must not declare a `severity:`, and must agree with the gate on `ignore-unfixed` and `.trivyignore`.

`scripts/ci/check-trivy-gate-severity.sh` is deliberately left alone. It covers the same policy from the other side — every *gating* Trivy step in *every* workflow applies the filter it declares — with no Rust compile. The two now fail from either direction.

I checked the other guards in the file for the same drift, as the issue asked: `merge_jobs_depend_on_container_scan` and `scan_job_covers_all_actively_published_builds` are still accurate against the current workflow (both pass unmodified, and both are shown failing under mutation below).

### 2. It never ran — and the issue's diagnosis of *why it stayed invisible* is not quite right

The issue says PRs stay green "while `cargo test --workspace` — the Tier 2 gate for `main` and release branches — is red". **No job runs `cargo test --workspace`.** The only two `cargo test --workspace` invocations in `ci.yml` (lines 1270 and 1297) both carry an explicit `--test` allowlist that does not name this file, and the unit/coverage jobs are `cargo nextest run --workspace --lib`. So there was no red signal anywhere — not on PRs, not on `main`. The guard against image publishing silently becoming non-blocking was itself silently non-blocking.

**Adding it to the Tier 2 allowlist would have re-created the defect in a new place.** That job invokes `-- --ignored`, and every one of the 15 suites in its allowlist is 100% `#[ignore]`d by construction (verified file by file). These three tests are not. Measured on this branch:

```
$ cargo test --workspace --test workflow_scan_gate_tests -- --ignored --test-threads=1
running 0 tests
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out
$ echo $?
0
```

A green, permanently vacuous gate — the exact defect being fixed. So it runs in **Tier 1** instead, as its own step in `test-backend-unit`, with `--no-tests=fail` so a rename or deletion of the target is red rather than a silent zero-test pass. Tier 1 is also right on the merits, and is the direction the issue's own "Also worth considering" section points: pure YAML parsing, no database, no network, ~25 ms, and it gates every PR rather than only backend-touching ones. `CLAUDE.md` records the extra target and why it is not in the Tier 2 list.

**On `main`, the test fails** (this is the reported bug, reproduced on a pristine checkout):

```
test trivy_scan_steps_enforce_via_exit_code ... FAILED
  scan-containers / Trivy SARIF - Backend: Trivy must run with exit-code '1' ... (#2609)
  left: Some("0")  right: Some("1")
test result: FAILED. 2 passed; 1 failed
```

**On this branch, all five pass** in 25 ms: `every_scanned_image_is_covered_by_both_trivy_passes`, `trivy_enforcement_pass_gates_on_the_stated_policy`, `trivy_visibility_pass_never_blocks_and_never_claims_a_filter`, `merge_jobs_depend_on_container_scan`, `scan_job_covers_all_actively_published_builds`.

**The rewritten guard is falsifiable.** Ten mutations applied to `docker-publish.yml`, each reverted after the run — every one is caught, and the failing assertion is the right one:

| # | mutation to `docker-publish.yml` | caught by |
|---|---|---|
| M1 | backend **gate** pass downgraded to `exit-code: '0'` (the #2609 regression) | `every_scanned_image_is_covered_by_both_trivy_passes` — "expected exactly ONE enforcement Trivy pass … Found 0" |
| M2 | backend **gate** pass deleted entirely | same — the case a filter-based guard would have passed vacuously |
| M3 | openscap gate switched to `format: sarif` (the #3122 shape) | `trivy_enforcement_pass_gates_on_the_stated_policy` |
| M4 | scanner-adapter gate `severity` narrowed to `CRITICAL` | `trivy_enforcement_pass_gates_on_the_stated_policy` |
| M5 | SARIF pass gains a `severity:` it cannot honour | `trivy_visibility_pass_never_blocks_and_never_claims_a_filter` |
| M6 | backend gate drops `trivyignores` | `trivy_enforcement_pass_gates_on_the_stated_policy` |
| M7 | backend gate drops `ignore-unfixed` | `trivy_enforcement_pass_gates_on_the_stated_policy` |
| M8 | `merge-backend` stops `needs: scan-containers` | `merge_jobs_depend_on_container_scan` |
| M9 | scan step count changed (extra/removed pass) | pairing assertions |
| M10 | `aquasecurity/trivy-action@` renamed away (vacuity) | all three Trivy tests fail — "found 0 distinct image-refs" — rather than passing on an empty set |

Restored → `5 tests run: 5 passed, 0 skipped`.

`cargo fmt --check` clean; `cargo clippy --lib --tests -- -D warnings` clean.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc:
Live CI confirmation that the guard now executes (it previously ran in no job at all). From `🧪 Backend Unit Tests` on [run 32651403003](https://github.com/artifact-keeper/artifact-keeper/actions/runs/32651403003), step `Run workflow contract tests`:

```
Run cargo nextest run --workspace --test workflow_scan_gate_tests --no-tests=fail
 Nextest run ID f1d42d63-... with nextest profile: default
    Starting 5 tests across 1 binary
        PASS [0.024s] (1/5) workflow_scan_gate_tests merge_jobs_depend_on_container_scan
        PASS [0.025s] (2/5) workflow_scan_gate_tests trivy_enforcement_pass_gates_on_the_stated_policy
        PASS [0.025s] (3/5) workflow_scan_gate_tests trivy_visibility_pass_never_blocks_and_never_claims_a_filter
        PASS [0.025s] (4/5) workflow_scan_gate_tests every_scanned_image_is_covered_by_both_trivy_passes
        PASS [0.031s] (5/5) workflow_scan_gate_tests scan_job_covers_all_actively_published_builds
     Summary [0.032s] 5 tests run: 5 passed, 0 skipped
```

32 ms, and the runner's nextest accepts `--no-tests=fail`. Whole run green.

brandonrc:
### Live CI proof that the guard now blocks the #2609 regression

Two throwaway commits on this branch (a temporary downgrade, then its revert) exist to demonstrate this on the runner rather than only locally. The diff against `main` is unchanged by them.

Downgrading **one** step — `scan-containers / Trivy gate - Backend` from `exit-code: '1'` to `'0'`, which is exactly the "image publishing silently becomes non-blocking" regression #2609 exists to catch — turned `🧪 Backend Unit Tests` red on [job 97229730900](https://github.com/artifact-keeper/artifact-keeper/actions/runs/32653951640/job/97229730900):

```
10 Run workflow contract tests -> failure     (Process completed with exit code 100)
```

with

```
FAIL every_scanned_image_is_covered_by_both_trivy_passes
  ${{ steps.refs.outputs.backend }}: expected exactly ONE enforcement Trivy pass
  (non-zero exit-code) - that is the step the merge jobs' `needs:` edge reacts to
  (#2609). Found 0: ["scan-containers / Trivy SARIF - Backend (exit-code 0, format sarif)",
                     "scan-containers / Trivy gate - Backend (exit-code 0, format table)"]
FAIL trivy_visibility_pass_never_blocks_and_never_claims_a_filter
```

On `main` today that same downgrade is invisible: no job runs this file.

Reverted; `5 tests run: 5 passed`.

Merge note: #3518 and #3523 each add steps to `shell-tests` in `ci.yml`; this PR adds a step to `test-backend-unit`. No overlap, but whichever of the three lands last may need a trivial additive rebase in that file.

Issue #1704, opened 2026-06-06
CI: jscpd duplication gate fails on any multi-handler change (format handlers are inherently near-duplicates)


brandonrc:
## Problem
The Code Duplication gate (`jscpd --min-lines 10 --threshold 3`) in `ci.yml` runs across **all changed `.rs` files together**. The per-format handlers under `backend/src/api/handlers/<format>.rs` are structurally near-identical to one another (same import block, same virtual-resolve / download / publish skeleton). jscpd therefore reports large **cross-file** clones between different handlers whenever a change touches several of them.

Measured on `main` (no PR), the same handler file set already sits at **~18%** duplication. So any PR that legitimately edits the download/upload path across many formats (e.g. #1703 / #1608 streaming) will exceed the 3% threshold regardless of code quality.

Observed on #1393 (streaming downloads): **20.9%**, with the dominant clones being whole-file matches like `sbt.rs <-> vscode.rs` (including imports), not the actual change.

## Options
- Have jscpd compare each changed file against the **base** version of itself (true *new* duplication) rather than changed-files-against-each-other.
- Or exclude/relax the threshold for the inherently-similar `backend/src/api/handlers/*.rs` family.
- Or raise `--min-lines` so import blocks and tiny skeletons don't count.

## Why this matters
The current behavior blocks otherwise-green PRs (fmt/clippy/tests/coverage all pass) and pressures contributors toward `--admin` bypasses, which the repo policy forbids.

Surfaced while adopting #1393 (sub-item #1703 of epic #1607).

Issue #1619, opened 2026-06-04
CI: duplication gate only measures changed lines — add periodic full-repo jscpd scan


brandonrc:
## Make the duplication gate catch pre-existing / whole-file duplication

**Origin:** Surfaced by the proxy-refactor analysis (#1618, discussion #1614). `proxy_service.rs` (6,430 lines) and `proxy_helpers.rs` (5,094 lines) accumulated large-scale copy-paste (5 duplication families) **despite** the jscpd CI gate.

## Why the gate missed it
Per `CLAUDE.md`, the duplication gate measures **only changed lines on the PR diff** (≤3% on changed files). A pre-existing duplicated surface therefore passes indefinitely — every PR only nudges a few lines, so the giant duplicated baseline is never measured.

## Proposal
- Add a **periodic full-repo jscpd scan** (scheduled workflow, not the per-PR gate) that reports total duplication % and the top duplicated blocks, so structural debt is visible and trackable over time.
- Optionally: a non-blocking "duplication budget" trend check that fails only when whole-repo duplication regresses beyond a threshold.
- Keep the existing per-PR changed-lines gate as-is (it prevents *new* localized duplication); this is a complementary whole-repo view.

## Done when
A scheduled job publishes repo-wide duplication metrics, and the proxy hotspots (#1618) are tracked against that baseline.

PR #3523, opened 2026-08-23
fix(ci): stop the duplication gate silently skipping 45% of backend/src

### LINKED ISSUE #3495: jscpd duplication gate silently skips the three largest source files (maxLines/maxSize defaults), including two the config explicitly refuses to exempt
## Summary

The "Code duplication gate" (`.github/workflows/ci.yml:948`) invokes jscpd without overriding its file-size limits, so jscpd silently skips the repo's largest source files. Two of them are files `.jscpd.json` explicitly says must **not** be exempt.

This is about the gate being unfalsifiable, not about a duplication problem — see "Measured result" below.

## Detail

`ci.yml:1012-1014`:

```
jscpd --min-lines 10 --threshold 3 --reporters json \
  --format rust --output /tmp/jscpd-report \
  $(echo "$CHANGED_RS" | tr '\n' ' ')
```

jscpd v4 defaults to `maxLines: 1000` and `maxSize: "100kb"`. A file exceeding either is skipped without an error and without appearing in the report — it simply is not analysed, and the gate reports whatever the remaining files measured.

Files on `main` @ `a9f3876` that exceed both limits:

| file | lines | bytes |
|---|---|---|
| `backend/src/api/handlers/oci_v2.rs` | 33,983 | 1,345,702 |
| `backend/src/api/handlers/repositories.rs` | 25,497 | 1,049,829 |
| `backend/src/services/artifact_service.rs` | 4,722 | 188,134 |

`oci_v2.rs` is *additionally* on `.jscpd.json`'s `ignore` list — that part is deliberate and documented (format handlers are inherently structural near-duplicates, #1704). The size/line skip is not deliberate, and it reaches two files the config goes out of its way to keep in scope. `.jscpd.json`'s own comment block says:

> Do NOT add non-format handlers (admin, auth, users, repositories, ...) or anything under `services/`, `storage/`, `middleware/` to this ignore list -- duplication there is a real smell and must keep failing the gate.

`repositories.rs` and `artifact_service.rs` are exactly those, and exactly what the gate is not measuring. Any PR whose Rust changes land only in one of those files gets a duplication verdict computed from an empty or near-empty file set.

## Measured result once the limits are lifted

Re-running the same gate with `--max-lines`/`--max-size` raised so the files are actually analysed: **2.57% duplication, 59 clones** — under the 3% threshold. So there is no hidden duplication emergency; the defect is that for the repo's three biggest files the gate has never been able to fail, and nobody would find out until it was needed.

Of the 59 clones, exactly one touched recently-changed code, and it was a pre-existing block moved verbatim — i.e. the raised-limit run is also usable in practice, not a wall of noise.

## Suggested fix

1. Pass explicit `--max-lines` and `--max-size` (or set `maxLines`/`maxSize` in `.jscpd.json`) high enough to cover the whole tree, and note in the config comment that the defaults silently skip rather than warn.
2. Optionally emit a warning when a changed file was skipped for any reason, so a future size regression is visible instead of silent.
3. Consider whether `oci_v2.rs`'s explicit ignore entry should stay now that its skip is no longer accidental — that part is a separate judgement call and does not block this fix.

## Related

- #1619 — the gate only measures changed lines; add a periodic full-repo scan. Complementary: that issue is about *scope*, this one is about the changed-file scope not being honoured either.
- #1704 — why format handlers are on the ignore list.
- #3494, #3437 — same family: controls that cannot fail.

Found while verifying the 1.8.2 milestone; re-verified against `main` @ `a9f3876` on 2026-08-22.

brandonrc:
## Summary

Closes #3495.

jscpd defaults to `maxLines: 1000` and `maxSize: "100kb"` and skips a file over **either** limit silently — no error, no warning, no entry in the report — after which the gate reports whatever the remaining files measured. `ci.yml`'s "Code duplication gate" never overrode those limits.

### What I found vs what the issue claimed

The issue named **three** files, on the basis that they exceed *both* limits. jscpd applies the two checks independently (`@jscpd/finder`: `bytes(stats.size) > maxSize` and `lines < minLines || lines > maxLines`), so the real set is every file over *either*:

| | count |
|---|---|
| `.rs` files under `backend/src` | 336 |
| skipped by `maxLines: 1000` | 152 |
| skipped by `maxSize: 100kb` | 66 |
| **skipped by either** | **152 (45%)** |
| of those, **not** on `.jscpd.json`'s ignore list | **125** |

That includes `services/scanner_service.rs`, `services/proxy_service.rs`, `services/storage_gc_service.rs`, `services/auth_service.rs`, `api/middleware/auth.rs`, `services/repository_service.rs`, `storage/s3.rs` — exactly the set the config's own comment says must keep failing the gate.

The issue's framing that this is about falsifiability rather than a duplication problem is right, and I verified it directly. With `repositories.rs` and `services/artifact_service.rs` as the entire changed-file list, `main`'s gate:

```
### Code Duplication: 0% (no duplicates found)      # exit 0, zero files analysed
```

and with a deliberate 79-line clone injected into `artifact_service.rs`, `main`'s gate gives the identical output and exit 0.

### The part the issue's fix would have missed

**Raising the limits alone makes the gate unusable.** `backend/src` is 474k lines, of which **265k (56%) are inline `#[cfg(test)]` modules** — `repositories.rs` is 26,328 lines with 22,164 of them test code. Measured with the limits raised and nothing else changed:

- the non-ignored tree scores **3.38%**, over the 3% threshold;
- **23 of the last 44** code-touching commits on `main` would have failed the gate, several at 4–10%;
- 58 of the 61 clones in the issue's own sample lie entirely inside `#[cfg(test)]` modules; only 2 are production-to-production.

The issue's "2.57%, under threshold" holds for those three files in isolation, but not for real changed-file sets. Shipping a limits-only change would have red-flagged the majority of PRs on test-module boilerplate — the #1704 pressure the config warns about.

So the measurement changed too: **inline test modules are now excluded, explicitly.** That is applying existing policy, not weakening the gate — `.jscpd.json` already ignores `**/*_test.rs` and `**/tests/**`, and the CI step already drops `_test`/`tests/` from the changed-file list. Test code is out of scope **by policy**; inline test modules were out of scope only **by accident**, because the files holding them were too big for jscpd to open. `scripts/ci/jscpd-prepare-sources.py` copies each changed source with its top-level `#[cfg(test)]` items removed and keeps a line map so failures still report real line numbers. **Only column-0 attributes count as boundaries** — an indented `#[cfg(test)]` is inside another item and treating it as top-level would delete production code up to the enclosing item's closing brace; 22 files under `backend/src` have one, and the self-test pins that case.

With that in place: tree-wide **2.27%**, and **2 of those 44 commits** exceed the threshold — both on genuine production-code clones (`storage_gc_service.rs`, `auth_config_service.rs`), which is the gate doing its job.

### Two silent passes are now failures

- A changed file that jscpd did **not** analyse — for any reason, not only size — fails the gate, naming the file and the reason, instead of being averaged out of the result.
- A run that produces **no report at all** fails, instead of printing `0% (no duplicates found)`. That output is precisely what a whole-file-set skip looked like.
- A changed-file set where nothing reaches jscpd's 10-line floor is still N/A and passes: that is the one legitimate omission, and jscpd genuinely emits no report for it.

`maxLines`/`maxSize` live in `.jscpd.json` next to the ignore list, and the CI step reads them from there and passes them explicitly, the same belt-and-braces the `ignore` list already gets ("deterministic regardless of jscpd version").

`oci_v2.rs` stays on the ignore list. That exemption is deliberate and documented (#1704) and is unaffected either way, since the CI step filters ignore-listed handlers out before jscpd runs.

Every run below drives the **actual `run:` block** extracted from `ci.yml` (via `yaml.safe_load`), with only the `npm install` and `git diff` preamble replaced by a fixed `CHANGED_RS`, against a real jscpd 4.3.0.

**1. The gate can now fail — inject, watch, revert.** A 79-line clone of `upload_with_sync_options` added to `artifact_service.rs`:

```
Duplication: 3.7% (99/2653 lines, 3 clones, 1 files analysed)
::error::Code duplication is 3.7%, exceeding 3% threshold
  backend/src/services/artifact_service.rs:545-622 <-> backend/src/services/artifact_service.rs:468-545
  ...
EXIT=1
```

(line numbers translated back through the stripped test module to the real file — the injection was at line 466). Reverted:

```
Duplication: 0.8% (22/2574 lines, 1 files analysed)
EXIT=0
```

**2. The same clone under `main`'s gate:** `### Code Duplication: 0% (no duplicates found)`, `EXIT=0`.

**3. The silent-skip detector fires.** With `maxLines` lowered to 2000 in `.jscpd.json` so a changed file exceeds it:

```
::error::The duplication gate did not analyse every changed file, so its verdict is not a measurement of this change (#3495).
  backend/src/services/artifact_service.rs: 2576 lines exceeds maxLines 2000
Raise maxLines/maxSize in .jscpd.json, or exempt the file there deliberately.
EXIT=1
```

Restored → `EXIT=0`.

**4. Legitimate omissions still pass.** A single 4-line changed file: `### Code Duplication: N/A (no changed file has 10+ production lines)`, `EXIT=0`.

**5. The source preparation has its own self-test** (`scripts/ci/test-jscpd-prepare-sources.sh`, 13 cases, wired into `shell-tests`) covering the trailing test module, no test module, `#[cfg(all(test, ...))]`, a mid-file module with production code after it, several modules in one file, `#[cfg(test)] use ...;`, line-map translation, usage error — and the dangerous case, an **indented** `#[cfg(test)]`, which must remove nothing.

**The self-test was revert-proofed** by breaking the stripper three ways:

| deliberate breakage | self-test result |
|---|---|
| allow indented `#[cfg(test)]` as a boundary (the over-removal regression) | 2 of 13 FAIL |
| strip nothing (gate reverts to measuring test code) | 6 of 13 FAIL |
| stop recording the line map | 3 of 13 FAIL |

restored → `passed: 13   failed: 0`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

## Known consequence, stated plainly

Files the gate now measures may hold pre-existing clones that the PR touching them did not introduce, because its scope is whole changed **files**, not changed lines (#1619). On the last 44 code-touching commits that is 2 — both real production duplication worth a look. The failure output now says so explicitly and points at #1619.


brandonrc:
### Live CI proof that the gate can fail

Two throwaway commits on this branch (`097b7bc5` then its revert `fbe98316`) exist to demonstrate this on the real runner rather than only locally. They are in the branch history deliberately; the diff against `main` touches no Rust source.

**Why it was needed:** this PR changes no `.rs` file, so its own `Code duplication gate` step short-circuits at "No measurable .rs source files changed" and never exercises the new code. Without an injected change the gate's own CI would be as unfalsifiable as the defect being fixed.

`097b7bc5` added a deliberate 79-line clone of `upload_with_sync_options` to `backend/src/services/artifact_service.rs` — one of the files `main`'s gate cannot open at all. [Coverage job 97226045082](https://github.com/artifact-keeper/artifact-keeper/actions/runs/32652440392/job/97226045082):

```
13 Code duplication gate -> failure
::error::Code duplication is 3.7%, exceeding 3% threshold
```

(`New code coverage gate` failed alongside it, as expected: 79 uncovered new lines. Different gate, unrelated.)

On `main`'s version of the step the identical commit produces `### Code Duplication: 0% (no duplicates found)` and exit 0, because the file is over `maxLines` and is skipped without a word.

`fbe98316` reverts it; the branch is back to workflow, config, scripts and CHANGELOG only.

PR #3527, opened 2026-08-23
feat(oci): configure virtual negative cache bounds

### LINKED ISSUE #1424: [enhancement] Make OCI virtual negative cache TTL + max entries configurable
Follow-up to PR #1419.

Constants `VIRTUAL_NEGATIVE_CACHE_TTL_MS = 5_000` and `VIRTUAL_NEGATIVE_CACHE_MAX_ENTRIES = 4096` are hardcoded. Move into `Config` for operator tunability without recompile.

Surfaced by Round 3 architecture review.

cazlo:
## Summary

- add `OCI_VIRTUAL_NEGATIVE_CACHE_TTL_MS` and
  `OCI_VIRTUAL_NEGATIVE_CACHE_MAX_ENTRIES` backend configuration
- preserve the existing 5000 ms and 4096-entry defaults
- thread the configured policy through OCI virtual blob and manifest negative
  cache lookups/inserts without changing cache keys, storage, or authorization
- document and test zero-value behavior: zero TTL disables hits and zero
  maximum disables new inserts

## Testing

- `cargo fmt --all --check`
- workspace/all-targets Clippy with warnings denied
- backend coverage suite: 15,845 passed, 10 skipped; changed-source coverage
  93%
- backend integration suites, including virtual-member concurrency coverage
- signed backend image native-client and proxy/virtual runtime smoke

Closes #1424


brandonrc:
## Audit round 1 — configurable OCI virtual negative cache (2 reviewers: correctness / security+API+integration)

Both passes read head `6612cb21`, merged `origin/main` (45 commits) into a throwaway branch — **zero conflicts, `cargo check --lib --tests` exit 0** — and ran the negative-cache tests under the lock (31 and 10 pass respectively, including all new ones). It does exactly what #1424 asks: both constants move to `Config` as `OCI_VIRTUAL_NEGATIVE_CACHE_TTL_MS` / `_MAX_ENTRIES`, defaults byte-identical (5000 / 4096), both live call sites threaded (no third exists), all nine `Config` fixtures updated, the deleted tripwire test replaced in `config.rs`. Env-only is the right surface — the cache is one process-global map. `.env.example` documentation is above the repo's own bar. No new authorization weakness: the caller-sees-every-member invariant is unchanged and sound against poisoning. Two things to add before merge, both found independently by both reviewers.

### Should-fix

1. **Clamp both knobs.** `config.rs:1435-1442` takes them straight from `env_parse` with no ceiling, while the repo's own convention is the opposite — `blob_gc_sweep_grace_secs` is `.min(7 days)` at `config.rs:1284-1285` with the comment *"so a fat-fingered enormous value can't silently disable the sweep forever"*, and three more `.min()`ed knobs sit next to it. Without a clamp:
   - a large TTL holds a stale 404 on a freshly published tag — `resolve_virtual_blob`/`resolve_virtual_manifest` return the negative hit *before* the local `oci_blobs`/`oci_tags` lookups (`oci_v2.rs:3395`/`:3516`) and `virtual_negative_cache_clear()` has no production caller, so a push to a Local member 404s through the Virtual for the whole TTL. The comment at `oci_v2.rs:3184-3188` justifying the design ("expires quickly so a real upstream publishing event is not blocked for long") is now only true at the default.
   - the cache **freezes** at the cap: `negative_cache_evict_and_has_room` (`oci_v2.rs:3271-3282`) only evicts past-TTL entries, so at 4096 entries × 1 h TTL nothing is ever evicted and every further insert is refused; conversely a huge `MAX_ENTRIES` removes the memory bound the constant existed for, on a path an **unauthenticated** caller reaches (anonymous → `PublicOnly`; on an all-public virtual `cacheable = true`) with attacker-chosen `image`/`reference` strings in the key.
   Suggested ceilings: TTL ≤ 60 s or so, entries ≤ 65 536 — your call, but something.

2. **Pin the wiring.** All the new tests call the pure helpers or `Config` directly; reverting `oci_v2.rs:3366`/`:3468`/`:3615` to literal `5_000`/`4096` passes the entire added suite. `backend/tests/oci_virtual_resolution_tests.rs:241-294` already has the harness (wiremock `.expect(1)`, `test_config` with `..Default::default()`); a sibling with `ttl_ms: 0` and `.expect(2)` closes it in ~30 lines.

### Worth knowing (pre-existing, not this PR's to fix)
The key carries no principal term; the safety argument holds only at a fixed member set. Anonymous caches a miss over 3-of-3 public members → a private 4th member holding that tag is added → an admin sees 4-of-4, hits the anonymous-written entry, and gets a 404 without walking the new member. Fails closed, leaks nothing — but the window was hard-capped at 5 s and is now whatever the operator types, which is the other reason for the clamp.

### Bookkeeping
No CHANGELOG entry for #1424/#3527 — needed before the cut, under `### Added` (there isn't one yet in `[Unreleased]`). I'll push the `docs(changelog)` commit once the clamp lands. Milestoning 1.9.0.

brandonrc:
Checking in on this one. The audit for the configurable OCI virtual negative cache has been sitting since 2 September with no reply, and 1.9.0 shipped yesterday — I would like to get this into 1.10.0 rather than let it age further.

cazlo:
Hi @brandonrc sorry for delayed response. I'm not gonna have bandwidth to work on this until like week after next, so please feel free to take over this PR, I'm not planning to push here in meantime...

brandonrc:
Thanks @cazlo — taking this over as you offered, and your commit `6612cb21` is untouched at the base of the branch. Everything below is one commit on top of it (`38d47ca8`), plus a `main` merge. Release notes will credit you.

`origin/main` merged in first (the branch was from 23 August, 130 commits behind): **zero conflicts**, `cargo check -p artifact-keeper-backend --lib --tests` exit 0.

## Finding 1 — clamp both knobs

Done, in `config.rs`, the same way `blob_gc_sweep_grace_secs` is `.min()`ed beside it.

The entry ceiling is **65536** (16x the default), which is the memory bound on a key that carries caller-chosen `image` / `reference` strings on a path an anonymous puller reaches.

The TTL ceiling is **45000 ms**, and rather than pick a round number I tied it to `cache_classifier::NEGATIVE_CACHE_TTL_SECS` — the proxy layer's own negative-cache window. The argument: `write_negative_cache` is gated on `AppError::NotFound`, i.e. the proxy records a negative only for a *definitive* upstream 404. A virtual miss is the weaker statement "no member resolved this key", which a throttled (429) or broken (5xx) member produces too, because `proxy_fetch_streaming_with_cache_key_verified` folds every failure into `Err(_) => continue`. A weaker negative must not outlive the status-gated one beneath it. That is also the answer to whether a configurable bound reopens the #3756 / #3760 narrowing: it does not, because 45 s is the most the virtual layer can now hold, and the proxy layer would hold a real 404 for exactly that long anyway. See below for what I deliberately did *not* fix there.

Both clamps also close the second half of your finding — the freeze at the cap. `negative_cache_evict_and_has_room` only evicts past-TTL entries, so "at cap and nothing expiring" was unbounded before and is now bounded by 45 s.

Config-validation behaviour, all pinned by tests:

| value | result |
|---|---|
| unset | 5000 / 4096 — byte-identical to the constants you replaced |
| `0` | preserved; disables hits / inserts respectively (your tests) |
| `-1`, `invalid` | falls back to the default (your test) |
| `3600000` / `100000000` | clamped to 45000 / 65536 |

Two `const _: () = assert!(...)` guards keep each default strictly inside its ceiling, so the clamp can never silently move today's behaviour.

## Finding 2 — pin the wiring

You were right that restoring the `5_000` / `4096` literals at `oci_v2.rs:3546`/`:3649`/`:3846` passed the entire added suite. Three integration tests in `backend/tests/oci_virtual_resolution_tests.rs` now close it, over a Virtual whose members are one 404ing Remote and one empty Local:

- `resolve_virtual_blob_honors_configured_zero_ttl` — pins the hit site
- `resolve_virtual_blob_honors_configured_zero_max_entries` — pins the blob insert site
- `resolve_virtual_manifest_honors_configured_zero_max_entries` — pins the manifest insert site

One note on the shape you suggested: `ttl_ms: 0` with `.expect(2)` does **not** work. The proxy layer negative-caches the member 404 for 45 s, so the second walk is upstream-silent whatever the virtual bounds say, and the mock sees one call either way. They instead publish to the Local member between two probes and assert the second probe serves it — which is directly the harm your first finding describes ("a push to a Local member 404s through the Virtual for the whole TTL"), and is immune to the layer below.

## Verification

```
cargo nextest run -p artifact-keeper-backend --lib --test oci_virtual_resolution_tests \
  --run-ignored all -E 'test(oci_virtual_negative_cache) + test(virtual_negative_cache) +
  test(negative_cache_) + test(resolve_virtual)' --no-tests=fail
  -> Summary [0.410s] 41 tests run: 41 passed, 16240 skipped

cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings   -> clean
cargo fmt -p artifact-keeper-backend --check                           -> clean
```

Fail-without-fix, on the same tree with both production hunks reverted (clamps dropped, literals restored) and the tests left in place:

```
Summary [0.248s] 9 tests run: 5 passed, 4 failed, 16272 skipped
  FAIL config::tests::test_oci_virtual_negative_cache_enormous_values_are_clamped
  FAIL resolve_virtual_manifest_honors_configured_zero_max_entries
  FAIL resolve_virtual_blob_honors_configured_zero_ttl
  FAIL resolve_virtual_blob_honors_configured_zero_max_entries
```

Exactly the four new behavioural tests, and nothing else.

Postgres 16 on a throwaway container with all 216 migrations applied; the integration tests are `#[ignore]`d and were run with `--run-ignored all`.

## Bookkeeping

CHANGELOG bullet added under `[Unreleased]` / `### Added`, verified with `awk '/^## \[/{h=$0} /^- \*\*/{print h}' CHANGELOG.md` — so you do not need a separate `docs(changelog)` commit. `.env.example` documents both ceilings. Milestone is already 1.10.0.

## Left alone, deliberately

The negative cache does not distinguish *why* a member failed: a 429 or 5xx from every member is recorded as "not found" exactly as a 404 is. That conflation predates this PR and is now bounded at 45 s by the clamp, which is why I did not chase it here. Fixing it properly means threading the upstream status out of `try_upstream_fetch_with_accept` (five call sites, currently `.ok()`-discarding) so the manifest resolver can see it too — a separate change with its own review. Same for the principal-free key you flagged under "worth knowing"; the clamp is what caps that window now.

I will merge this once CI is green — nothing further needed from you. Thanks again for the contribution.

Issue #3529, opened 2026-08-23
Flaky unit test: completion_with_unknown_length_empty_final_put_skips_zero_byte_part intermittently fails with 'blob storage was being reclaimed concurrently'


brandonrc:
## Summary

`api::handlers::oci_v2::oci_blob_upload_streaming_tests::completion_with_unknown_length_empty_final_put_skips_zero_byte_part` intermittently fails in the Backend Unit Tests job. It failed once and passed on a re-run of the same commit with no code change.

## Observed

Run [32658416189](https://github.com/artifact-keeper/artifact-keeper/actions/runs/32658416189), job 97240751268, on a PR whose diff touches only `backend/tests/workflow_scan_gate_tests.rs`, `.github/workflows/ci.yml`, `CLAUDE.md` and `CHANGELOG.md` — nothing that can reach the OCI blob upload path:

```
FAIL [ 0.756s] ( 2856/15842) artifact-keeper-backend
  api::handlers::oci_v2::oci_blob_upload_streaming_tests::completion_with_unknown_length_empty_final_put_skips_zero_byte_part

thread '...' panicked at backend/src/api/handlers/oci_v2.rs:22228:9:
assertion `left == right` failed: empty unknown-length final PUT should complete:
  b"{\"errors\":[{\"code\":\"BLOB_UPLOAD_INVALID\",\"message\":\"blob storage was being reclaimed concurrently; retry the upload\"}]}"

Summary [109.194s] 2863/15842 tests run: 2862 passed, 1 failed, 8 skipped
```

Re-running the same job on the same commit: **pass** (job 97246792793, 19m26s). `main` was green at the commit the branch was based on.

## Why this is worth fixing rather than re-running

The failure message is a real concurrency guard firing — the test hits the blob-reclamation path rather than the completion path it is asserting. That points at test isolation (shared storage state or a GC/reclaim interaction between concurrently-running tests) rather than randomness.

The practical cost is misattribution: the failure surfaces on whatever PR happens to be in CI when it fires, and that PR's author starts by suspecting their own diff. This session it fired on a docs-and-workflow change. #3377 was the same shape — a process-wide `HTTP_PROXY` leak between tests made an intact SSRF control look broken, and the reported symptom pointed at the wrong subsystem.

## Suggested investigation

- Identify what sets `blob storage was being reclaimed concurrently` on the completion path and under what state it can be true for a test that just created its own upload session.
[...]

PR #3530, opened 2026-08-23
fix(upload): deduplicate chunked completion writes

### LINKED ISSUE #3517: Chunked upload completion rewrites blobs the server already has: the dedup exists() guard present on both other write paths is missing
## Summary

The chunked-upload completion path derives the content-addressed storage key and then writes unconditionally, skipping the `storage.exists()` short-circuit that both other write paths perform. Identical content uploaded through a chunked session is rewritten to the backend every time.

`backend/src/api/handlers/upload.rs:555-557` derives the CAS key, and `:585` calls `storage.put_file(&storage_key, &temp_path)` with no preceding existence check. The comment at `:553-554` states the key shape matches the artifact service's "so future dedup … see the upload like any other write" — the key shape does match; the guard does not exist.

Compare the two paths that do have it:

- buffered: `backend/src/services/artifact_service.rs:502-508`
- streaming: `backend/src/services/artifact_service.rs:570-583`

Both call `storage.exists(&storage_key)` and skip the `put` when the blob is already present.

## Impact

Storage-backend write amplification on every chunked re-upload of unchanged content. Chunked sessions are the path large artifacts take, so this is the case where a redundant write costs the most. On object stores it is a billable PUT and a needless overwrite of an object that is by definition byte-identical (the key is the sha256). No correctness problem — the bytes are the same — which is likely why it has gone unnoticed.

## Not the same as the format-flat key gap

Several format paths (Go, npm, maven, rpm, proxy cache) use non-content-addressed keys and cannot dedup at all by construction. That is a separate design question. This issue is narrower: the chunked path *does* use the CAS key and *does* have the machinery available, it simply omits the check.

## Suggested fix

Add the same `exists()` short-circuit before `put_file` at `upload.rs:585`, matching `artifact_service.rs:502-508`. Worth a test that asserts the second completion of an identical payload performs no backend write, since the observable behaviour is otherwise identical and a regression would be invisible.

## Related

Found while scoping #3427 (bulk/air-gap ingest), where re-import traffic is the central concern — a client-side skip-existing diff would reduce the transfers, and this closes the server-side half for the chunked path.

cazlo:
## Summary

- check the repository-scoped content-addressed key before completing a
  chunked upload and skip `put_file` when the object is already present
- verify filesystem hits by SHA-256 so truncated or equal-length corrupt
  legacy objects are repaired instead of reused
- propagate Azure, S3, and GCS existence-probe failures, including migration
  fallbacks, rather than treating operational errors as cache misses
- cover write-count deduplication, corrupt filesystem repair, and retryable
  existence/write failures

## Why

The buffered and streaming upload paths both call `storage.exists()` before
writing a content-addressed blob, but chunked completion wrote it
unconditionally. That makes re-uploading unchanged large artifacts perform a
redundant backend write even though their SHA-256 storage key is already
present.

A blind existence check is not sufficient for the filesystem backend: its
existing `put_file` implementation copies directly to the final key, so an
interrupted legacy write can leave truncated or equal-length corrupt bytes
that the old unconditional completion path would overwrite. Filesystem hits
are therefore streamed through SHA-256 before reuse; mismatches take the
existing write path and repair the object. Cloud hits remain metadata-only so
deduplicating a large object does not require downloading it.

Existence checks also return `false` only for definite not-found responses.
Authorization, throttling, transport, and service failures remain storage
errors, including on migration fallback probes, so completion retains its
staged payload and can be retried instead of treating an unverified object as
a miss.

## Test Checklist

- [x] Integration tests added/updated (DB- and filesystem-backed handler coverage)
- [ ] E2E tests added/updated (not applicable)
- [x] Manually tested locally

## Validation

- `cargo nextest run --workspace --lib` (15,850 passed; 10 skipped)
- backend integration suites (nine shards; 64 passed)
- backend Compose smoke and web E2E
- line coverage: 87.72% overall; 94% on changed lines

## API Changes

- [x] N/A - no API changes

Closes #3517


brandonrc:
## Audit round 1 — chunked-completion dedup (3 reviewers: correctness / security+integration / architecture+tests)

All three passes read head `59be6412`, merged `origin/main` (45 commits) into a throwaway branch — **zero conflicts; `cargo check --lib --tests` exit 0**; no main commit touches the four files — and ran the ten new tests against a live Postgres with `AK_TESTS_REQUIRE_DB=1`: **10/10, three consecutive runs, no flakiness**. It closes #3517 exactly as specified, including the write-counting regression test the issue asked for.

**What checks out.** The guard keys off the *final, server-verified* digest — `complete_session` re-hashes the assembled temp file and refuses a mismatch (`upload_service.rs:692-716`) before the storage key is derived (`upload.rs:555-557`) and before the guard (`:594`) — so there is no client-supplied-digest hazard: a client cannot name a digest it lacks the preimage of. Every side effect survives the skip: `size_bytes`/`checksum_sha256`/`storage_key` still written (`upload.rs:731-746`), quota admission still runs, the #3134 ledger is trigger-maintained off the `artifacts` INSERT so accounting is unchanged, the staged temp file is removed unconditionally at `:647`. Concurrent completions are safe on content-addressed keys with atomic cloud writes; no lock or conditional put is needed. `exists()` is one HEAD per completion on every backend and depends on no LIST, so #3593/#3594 are not on this path. #3529 (the flaky OCI completion test) is in `oci_v2.rs`, untouched. And a point worth crediting: **the Azure `exists()` hardening is load-bearing, not incidental** — on main Azure returned `Ok(false)` for *any* non-2xx HEAD, so a naive guard on top of that would have skipped the write on a throttled HEAD and left an artifact row pointing at bytes never stored. You found that before adding the guard.

### Should-fix (all three reviewers, independently)

1. **Fix `FilesystemStorage::put_file` instead of working around it.** `filesystem.rs:314-323` is a bare `fs::copy` onto the final CAS path — the *only* write in that backend that isn't stage-and-rename (`put` `:178-220`, `copy` `:265-310`, `put_stream` `:392-449` all are; `temp_path_for_dest` is right there). The PR's comments call the hazard "the **old** filesystem `put_file`" (`upload.rs:584`) and "**legacy** in-place copies" (`:1008`), but it is live: two concurrent completions both entering `fs::copy` with `O_TRUNC` is still open, and the re-hash at `upload.rs:601-617` becomes permanent O(n) overhead on the default backend rather than an upgrade-window repair. The handler-level re-hash also only protects the chunked path — after an interrupted copy leaves a truncated CAS object, a **direct** upload of the same content still hits `artifact_service.rs:505` → `exists()` → skip, and the truncated blob is served. `put_file` has one production caller: stage+rename like `copy()`, or delete the override entirely (the trait default at `storage/mod.rs:215-228` already routes through the atomic `put_stream`). That removes the flag, the branch, the helper and two of the five handler tests.

2. **The filesystem dedup *skip* is never asserted.** The write-counting test (`upload.rs:3757`) points the repo at a fake backend named `chunked-complete-recording`, so `verify_existing_filesystem_object` (`:571`) is false and the verify-then-skip branch never runs; the two filesystem tests only cover mismatch → rewrite. A mutation making `storage_object_matches_checksum` always return `Ok(false)` — "filesystem never dedups", i.e. exactly #3517 — passes all ten tests, on the backend most deployments run.

3. **Azure `exists()` now fails closed for *every* caller, and that has a cost the PR doesn't state.** The change at `azure.rs:1496-1504` is right for the dedup guard (see above), but `exists()` is a shared trait method: an ordinary non-chunked `PUT` during an Azure 503/429 window now fails at `artifact_service.rs:505`/`:574` where it previously fell through to a redundant-but-successful write, and `authorized_head` (`azure.rs:883-905`) has no retry beneath it. Two acceptable resolutions: a bounded retry in `authorized_head`, or have the two direct paths treat an `exists()` *error* as "write it" (idempotent on a CAS key). Either way, update the trait doc (`storage/mod.rs:149-150` still says only "Check if key exists") and give it a sentence in the CHANGELOG bullet — it is a storage-contract change riding on an upload fix.

4. *(narrow)* In Artifactory-migration path mode on cloud, `artifactory_fallback_path` (`s3.rs:918-928`) maps `ab/cd/<sha>` → `ab/<sha>`, so an `exists()` hit can be on the *fallback* key and the skip leaves the canonical key permanently unwritten — reads work only while migration mode is on. Pre-existing on the two direct paths, newly introduced for the chunked one. One-line stopgap: force `write_object = true` at `upload.rs:601` when the backend has a fallback.

### Nits
- No test covers the concurrent-completion race (harmless today, but it's the scenario the issue title describes).
- Third independent copy of the dedup decision alongside `artifact_service.rs:504-510` and `:571-577`; the trait doc wasn't updated.

### Bookkeeping
No `## [Unreleased]` bullet for #3517/#3530 — needed before the cut, with the Azure note. I'll push it once 1–3 are in. Milestoning 1.9.0 (it closes a milestone bug).

brandonrc:
Checking in on this one. The audit for the chunked-completion dedup has been sitting since 2 September with no reply, and 1.9.0 shipped yesterday — I would like to get this into 1.10.0 rather than let it age further.

cazlo:
Hi @brandonrc sorry for delayed response. I'm not gonna have bandwidth to work on this until like week after next, so please feel free to take over this PR, I'm not planning to push here in meantime...

brandonrc:
## Taken over per @cazlo's go-ahead above — audit round 1 worked through

Thanks @cazlo. The dedup itself, the write-counting regression test, and the Azure `exists()` hardening are yours and are unchanged in substance — commit `59be6412` stays as the first commit on this branch and the release notes will credit you. Everything below is added on top as `08c604c7`, plus a merge of `origin/main` (`71f9ec1e`).

**Merge**: `origin/main` had moved 130 commits since 23 August (storage-GC lease work, the plugin-service removal, the 1.9.0 cut). **Zero conflicts** — no main commit touches `upload.rs`, `azure.rs`, `gcs.rs`, `s3.rs`, `filesystem.rs` or `storage/mod.rs`.

### Finding 1 — fix `put_file` instead of working around it ✅ fixed

`FilesystemStorage::put_file` now copies into a per-writer temp sibling, fsyncs, and renames into place, exactly like `put`, `copy` and `put_stream`. The `fs::copy` straight onto the final CAS path is gone, so the `O_TRUNC` window — truncated object under a content-addressed key after an interrupted copy, torn reads between concurrent writers of the same key — is closed at the storage layer where it belongs.

With the write fixed, the workaround comes out: `verify_existing_filesystem_object`, the verify-then-skip branch, `storage_object_matches_checksum`, the `sha2` import, and the two legacy-repair tests. As the audit noted, that re-hash only ever protected the chunked path — a *direct* upload of the same content hit `exists()` and skipped the write regardless — so it was a partial mitigation at permanent O(n) cost on the default backend.

**One thing that is now genuinely gone**: objects already corrupted on disk by the old `put_file` are no longer repaired by a chunked re-upload the way an unconditional write used to repair them. They were never repaired on the direct paths, so this is consistency rather than a new hole, but if you want existing installs scrubbed it wants its own issue — say the word and I will open one.

### Finding 2 — the filesystem dedup *skip* is never asserted ✅ fixed

`complete_does_not_rewrite_an_existing_filesystem_object` completes the same payload twice against the **real** `FilesystemStorage` and asserts the CAS object's inode is unchanged. Stage-and-rename is what makes this observable: a rewrite replaces the directory entry, so "filesystem never dedups" can no longer pass the suite.

### Finding 3 — `exists()` failing closed for every caller ✅ fixed (option 2)

Took the second of the two resolutions you offered. The two direct paths route their probe through `dedup_probe`, which treats a failed probe as "write it" and logs a warning: the key is the content's SHA-256, so the write is idempotent, an Azure 503/429 window costs one redundant-but-successful write exactly as it did before, and a backend that is genuinely down still surfaces its error from the write itself. Chunked completion keeps the fail-closed behaviour — it has a staged temp file and a retryable lease, which the direct paths do not.

The trait doc at `storage/mod.rs` now states the contract (`Ok(false)` means definite not-found; operational failures are errors; dedup-hint callers should fall back to writing). CHANGELOG carries it as a storage-contract change, as asked.

I did not add the bounded retry in `authorized_head` — that is a broader change to the Azure client with its own backoff/jitter questions, and it is not needed once callers handle a failed probe correctly.

### Finding 4 — migration-mode fallback leaves the canonical key unwritten ✅ fixed for the chunked path

Added `StorageBackend::exists_may_match_fallback_key()` (default `false`; S3/GCS/Azure return `path_format.has_fallback()`), and completion always writes when it is true. A one-line guard needed *something* to ask, and a trait method beat threading `StoragePathFormat` through the handler.

**Left alone on the two direct paths**, as your finding scoped it: pre-existing there, a behavioural change in migration mode that the issue is not about, and it deserves its own regression test and CHANGELOG line. Happy to open that issue.

### Nits

- **Concurrent-completion race** — covered, but at the storage layer rather than the handler: `test_put_file_replaces_the_destination_atomically` pins the actual hazard (a reader holding the old object must still see all of it). A handler-level two-completions race test would be nondeterministic, so I did not add one.
- **Third copy of the dedup decision** — not fixed. Unifying the three call sites is a refactor, not this PR; the two direct ones do now share `dedup_probe`.
- **Trait doc** — updated (finding 3).
- Adjacent observation, not changed: the `unreachable!()` after `require_success` in `gcs.rs::exists` is provably unreachable but is a panic in a request path. Flagging rather than touching it.

### Bookkeeping

CHANGELOG bullet added under `## [Unreleased]` / `### Fixed`, leading `#3517` then `#3530`, with the `exists()` storage-contract paragraph and the "no dedup under `STORAGE_PATH_FORMAT=migration`" note.

```
$ awk '/^## \[/{h=$0} /^### /{s=$0} /^- \*\*Completing a chunked/{print h" / "s}' CHANGELOG.md
## [Unreleased] / ### Fixed
```

### Verification

Postgres 16, `AK_TESTS_REQUIRE_DB=1`.

```
$ cargo fmt --all -- --check                                              # exit 0
$ cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings    # exit 0
$ cargo nextest run -p artifact-keeper-backend --lib -E '<the 12 tests>'
     Summary [1.675s] 12 tests run: 12 passed, 16267 skipped
$ cargo nextest run -p artifact-keeper-backend --lib \
    -E 'test(storage::) + test(upload::) + test(artifact_service::)'
     Summary [34.187s] 648 tests run: 648 passed, 15631 skipped
```

**Each behavioural fix fails without its production hunk** (hunk reverted, test re-run, hunk restored):

| Reverted hunk | Test | Failure |
|---|---|---|
| `put_file` back to in-place `fs::copy` | `test_put_file_replaces_the_destination_atomically` | `a reader holding the old object must still see all of it: put_file truncated the destination in place` |
| completion writes unconditionally | `complete_does_not_rewrite_an_existing_filesystem_object` | `an existing filesystem CAS object must be reused, not rewritten` |
| " | `complete_skips_put_file_for_existing_content_addressed_object` | `existing content-addressed object must not be written again` |
| drop the `exists_may_match_fallback_key` clause | `complete_writes_when_an_exists_hit_may_be_a_migration_fallback` | `an exists hit that may be a migration fallback must still write the canonical key` |
| `dedup_probe` back to `exists().await?` | `test_3517_dedup_probe_failure_still_writes_and_completes_the_upload` | `a failed dedup probe must not fail a content-addressed upload: Storage("backend throttled the existence probe for '1e/68/…'")` |

Note that reverting `put_file` alone makes `complete_does_not_rewrite_an_existing_filesystem_object` pass vacuously — `fs::copy` preserves the inode — so the reverts were run in three separate batches rather than one.

Not run here: the Compose smoke, the web E2E, and the nine integration shards. Those are CI's.

Over to you for the merge — I am not merging my own takeover.

PR #3534, opened 2026-08-24
fix(helm): uploaded charts missing from the Packages page

### LINKED ISSUE #3531: Uploaded Helm charts missing from the Packages page
Uploaded Helm charts appear on the Artifacts page but are absent from the Packages page. `upload_chart` inserts the `artifacts` rows but never writes to the `packages`/`package_versions` catalog, so the catalog (and Packages page) never sees them.

Expected: an uploaded chart is indexed into the package catalog like the other format handlers do, so it shows up on the Packages page.

xinnj:
## Summary

Uploaded Helm charts appeared on the Artifacts page but were absent from the Packages page: `upload_chart` wrote the `artifacts` rows but never indexed them into the `packages`/`package_versions` catalog. This populates the catalog after commit, matching the other format handlers.

Closes #3531

## Test Checklist

- [ ] Manually tested locally

## API Changes

- [x] N/A - no API changes


brandonrc:
## Audit round 1 — Helm charts in the Packages catalog (2 reviewers: correctness / integration+security)

Both passes read head `c21d9cba`, merged `origin/main` (63 commits, including the 1.8.2 helm changes #3515/#3444) into a throwaway branch — **zero conflicts, `cargo check --lib --tests` exit 0** — and ran the helm suite against a live Postgres: **60/60 pass**. The fix is the right one-call change in the right place: it uses `PackageService::try_create_or_update_from_artifact` like npm/NuGet/Composer/Incus, takes name/version from `Chart.yaml` (`helm.rs:807-808`) not the filename, reaches both upload routes (native `POST /helm/{repo}/api/charts` and the `/api/helm` cm-push alias — there is no raw-PUT route), is idempotent (re-push is a 409 before the write; the write is an upsert), and cannot fail a publish. **The test genuinely fails on main:** with the PR's 16 lines deleted, `panicked at helm.rs:2159: uploaded chart must be indexed into the package catalog: RowNotFound`. Visibility is inherited correctly — all three catalog read endpoints apply `build_visibility_clause_for` on both page and count queries, and `format` comes from `r.format`, not the metadata the PR passes.

### Should-fix (one line)
**The `Chart.yaml` description is discarded.** `helm.rs:954` passes `None`; `ChartYaml.description` is parsed (`formats/helm.rs:350`), `chart_yaml` is borrowed 15 lines earlier for `artifact_metadata`, and `list_packages` returns `p.description`. Helm is the only handler in the set hard-coding `None`. Fix: `chart_yaml.description.as_deref()`.

### Not blockers, but you should know
- **No backfill exists.** `POST /api/v1/admin/reindex` is the OpenSearch reindex and never touches `packages`; no migration inserts catalog rows. Charts pushed before this fix stay off the Packages page until re-pushed (safe: upsert). The CHANGELOG bullet must say so.
- **`helm delete` leaves a permanent ghost row** — probed empirically: after a 200 delete, `packages=1 package_versions=1 live_artifacts=0`. Nothing in the backend ever deletes catalog rows (`DELETE FROM packages` has zero runtime hits), so npm/nuget/conan/debian share this; this PR just makes it reachable for Helm. Follow-up issue, not this PR.
- The catalog write runs unconditionally after `apply_upload_hold_hosted` (`helm.rs:925` → `943-957`), so a quarantine-held chart is advertised before it's servable — the same shape #3599 hoisted behind the serve gate for OCI. Also follow-up material.
- rpm/cargo/alpine/conda/terraform insert their own `artifacts` rows with zero `try_create_or_update_from_artifact` references — the #3531 defect likely exists there too. Separate issue.

### Nits
Test asserts `packages` but not `package_versions` (nuget/composer/conan/debian tests do); nothing covers a signed upload's `.tgz` + `.tgz.prov` sharing `(name, version)` yielding exactly one row.

### Plan
This is a merge once the description line is in. Since it's a one-liner and the branch is nine days old, I'll push that change plus the `docs(changelog)` bullet (with the no-backfill upgrade note) to your branch today and merge on green — shout if you'd rather do it yourself. Milestoning 1.9.0.

Issue #3500, opened 2026-08-22
Two read paths build LIKE patterns without ESCAPE (tree listing, RubyGems quick gemspec), so wildcards in a path or a stored gem name return the wrong rows


brandonrc:
## Summary

Two read paths build a SQL `LIKE` pattern by concatenating an unescaped value with `'%'`, with no `ESCAPE` clause. This is the same shape as #3492 (Maven rollup delete guards), but on read paths, so the consequence is **wrong results / wrong artifact served**, not deletion.

A full audit of derived-pattern `LIKE` sites during the 1.8.2 campaign found exactly these two beyond the ones #3492 covers. Constant patterns (`LIKE 'maven/%'`, `LIKE 'oci-blobs/%'`, `LIKE '%/maven-metadata.xml'`) are unaffected — escaping applies to the pattern, not the operand.

## Site 1 — repository tree listing

`backend/src/api/handlers/tree.rs:229`, in the `GET /tree` handler:

```sql
WHERE a.repository_id = $1
  AND a.is_deleted = false
  AND ($2 = '' OR a.path LIKE $2 || '%')
```

`$2` is the request's `path` query parameter with a trailing `/` appended (`tree.rs:232-237`). It goes into the pattern verbatim.

- `%` or `_` in the requested path act as wildcards, so the listing can include entries from sibling folders — the folder view stops corresponding to the folder.
- A backslash is Postgres's default `LIKE` escape character, so a path containing one changes the meaning of the following character and the folder's own contents can fail to match.

Authorization is unaffected: `authorize_tree_read` runs first (`tree.rs:208`) and the query is `repository_id = $1`-scoped, so this cannot cross a repository boundary. It is a correctness bug within one already-authorized repository.

## Site 2 — RubyGems quick gemspec lookup

`backend/src/api/handlers/rubygems.rs:710`, in `find_local_quick_spec`, serving `GET /gems/{repo_key}/quick/Marshal.4.8/{full_name}.gemspec.rz`:

```sql
WHERE a.repository_id = $1
  AND a.is_deleted = false
  AND $2 LIKE a.name || '-%'
```

Here the **pattern is derived from stored data** (`a.name`, the artifact's own name) — precisely the #3492 shape. The comment above it states the intent: *"Narrow to artifacts whose name is a prefix of the requested full name so hyphenated gem names disambiguate correctly against the version suffix."*

[...]

Issue #3494, opened 2026-08-22
CI runs no bin-target tests and only 16 of 61 integration targets: 276 test functions never execute, so a green CI does not prove a new test ran


brandonrc:
## Summary

CI's Rust test surface is narrower than it looks. Two whole classes of test **never execute in any CI job**, and a third can pass without running. Together they mean **a green CI does not imply that a newly added test ran at all** — a test can be written, reviewed, merged, and quietly never execute again.

Measured on `main` @ `a9f3876`: **276 test functions are never executed by CI** (55 bin-target + 221 integration).

## CI's only Rust test invocations

| where | command |
|---|---|
| `.github/workflows/ci.yml:518` (Backend Unit Tests) | `cargo nextest run --workspace --lib --test-threads 8` |
| `.github/workflows/ci.yml:773` (Code Coverage) | `cargo nextest run --workspace --lib --test-threads 8` |
| `.github/workflows/ci.yml:1256` (integration) | `cargo test --workspace --verbose --test <15 explicit targets> -- --ignored --test-threads=1` |
| `.github/workflows/ci.yml:1283` (integration) | `cargo test --workspace --verbose --test incus_upload_tests <one named case>` |

`.github/workflows/sync-openapi-spec.yml:45` runs one named `--lib` test. That is the complete list.

## Class 1 — bin-target tests never run (55 functions)

`backend/Cargo.toml:7-9` declares a second compilation target:

```toml
[[bin]]
name = "artifact-keeper"
path = "src/main.rs"
```

`backend/src/main.rs` contains **55** `#[test]` / `#[tokio::test]` functions. `--lib` selects only the library target, and the integration job is `--test`-only, so nothing in CI ever runs them. Only `clippy --all-targets` *compiles* them.

Proof from the campaign, on a local checkout:

```
cargo nextest list --workspace --lib -E 'test(admin_group)'   # lists only the auth_config_service tests
cargo test --bin artifact-keeper -- admin_group               # runs 5 more, from main.rs
```

This is not hypothetical: a 1.8.2 PR shipped its two revert-proof-positive tests for its core behaviour into `main.rs`, i.e. into a target CI never executes.

## Class 2 — integration tests outside the 15-name allowlist never run (221 functions across 45 files)

`backend/tests/` holds **61** integration targets. The allowlist at `ci.yml:1256`/`:1283` names **16**. The other **45 never run anywhere**:

```
[...]

Issue #2520, opened 2026-07-14
PF-003 [P1]: Grouped Maven and Docker listings are incomplete or fully materialized


brandonrc:
# PF-003: Grouped Maven and Docker listings are incomplete or fully materialized

Priority: P1  
Affected area: Maven/Gradle component view, Docker/OCI tag view

## Finding

Grouped views paginate after loading and grouping source rows in memory. Hosted and virtual Maven views cap the source set at 10,000 artifact files, then report totals derived only from that subset. Docker tag grouping applies the same 10,000-row cap. Remote Maven removes the cap but fetches the entire package catalog and only then paginates.

This is both a performance and correctness defect: at million-artifact scale, users can receive plausible-looking but incomplete results and totals.

## Evidence

- Maven declares `MAX_FETCH = 10_000`, fetches only that many artifacts, groups them in Rust, then paginates the groups: [repositories.rs](../backend/src/api/handlers/repositories.rs#L3467).
- The response's `total_components` is simply the length of that in-memory grouped subset: [repositories.rs](../backend/src/api/handlers/repositories.rs#L3538).
- Remote Maven selects every matching package row without `LIMIT/OFFSET`, allocates a result vector, and sends it to the same in-memory paginator: [repositories.rs](../backend/src/api/handlers/repositories.rs#L3578).
- Docker declares `MAX_FETCH = 10_000`, expands multi-arch sizes, sorts all fetched tags, and then paginates: [repositories.rs](../backend/src/api/handlers/repositories.rs#L3746).

## Impact at scale

- A Maven repository with more than 10,000 artifact files can hide components entirely; source files per component make the threshold arrive much earlier than 10,000 components.
- Docker repositories with more than 10,000 tags return incomplete totals/pages.
- Remote Maven memory and latency grow linearly without a ceiling.
- Filtering after truncation can omit valid matches depending on ordering.

## Recommendation

- Define canonical relational keys for the grouped entity and aggregate/page in SQL. Maven can use the package catalog or indexed GAV metadata; Docker can page `(image, tag)` from `oci_tags` and calculate size/status for only the page.
- Use a separate `COUNT(*)` only if an exact group total is contractually required; otherwise use cursor + `has_more`. If exact totals are needed frequently, maintain group counters.
- Apply search before pagination using a matching index. Avoid `ILIKE '%q%'` on an unbounded package set.
[...]

PR #3535, opened 2026-08-24
fix: virtual repositories don't aggregate member content in listing views

### LINKED ISSUE #3532: Virtual repositories don't aggregate member content in listing views
A virtual repository owns no rows of its own — pushes to it route to a hosted member. Two listing views filter by the virtual repo's own key/id instead of expanding to its members, so they come back empty:

1. **Packages page** — the `repository_key` filter matches `packages` on the virtual repo's key, which owns no catalog rows.
2. **Docker grouped view** (`group_by=docker_tag`) — filters `oci_tags` on the virtual repo's id, which owns no rows.

Expected: both views expand the virtual repo to its members (priority-ordered, de-duplicated) so member content appears — matching the flat artifacts view, which already aggregates.

xinnj:
## Summary

A virtual repository owns no rows of its own — pushes route to a hosted member — but two listing views filtered by the virtual repo's own key/id instead of expanding to members, so they came back empty while the flat artifacts view listed the content:

- **Packages page** filtered `packages` by the virtual repo's key (which owns no catalog rows).
- **Docker grouped view** (`group_by=docker_tag`) filtered `oci_tags` by the virtual repo's id (which owns no rows).

Both now expand the virtual repo to its priority-ordered members and de-duplicate by member priority, matching the flat view.

Closes #3532

## Test Checklist

- [x] Manually tested locally

## API Changes

- [x] N/A - no API changes


brandonrc:
## Audit round 1 — virtual repositories in listing views (3 reviewers: correctness / security+integration / architecture+perf)

All three read head `f6150232` and reproduced their findings as SQL against a seeded Postgres (and, for the security pass, through the real handler on a merged tree). The aggregation *architecture* is right — one indexed query per view with `repository_id = ANY($members)` and `array_position()` for priority, members resolved through the existing `fetch_virtual_members`, page and `total` from the same `FROM`, nothing merged in Rust — and **the docker listing's authorization holds**: members are narrowed by `filter_visible_repo_ids` + per-member `member_passes_token_scope` *before* the query (`repositories.rs:6861-6891`), and a live probe (anonymous caller, public Virtual, one public + one private Local member) returned only the public member's tag, with `?count=exact` not counting the private one. Both views #3532 names do start aggregating. But the PR needs a round of work before it can land; the findings below were reached independently by two or three reviewers each.

### Blocker — does not build against `main`
Four merge conflicts, all `sqlx::AssertSqlSafe` sites from main's sqlx 0.9 bump (#2857), plus two errors the conflicts don't surface: main's #3555 regression test `test_docker_tag_search_treats_wildcards_literally_3500` calls `fetch_docker_tag_rows`/`count_docker_tag_rows` with a `Uuid`, which this PR retypes to `&[Uuid]` (`E0308` ×2). With all six fixed the merged tree checks clean. **Rebase hazard:** #3555 added `ESCAPE '\'` LIKE-escaping into exactly the two functions this PR rewrites; reapplying the PR's inline LIKE fragment silently reverts it.

### Should-fix

1. **De-duplicating `oci_tags` *before* the `artifacts` join erases tags.** `DISTINCT ON (t.name, t.tag)` (`repositories.rs:7041-7057`) runs over `oci_tags` alone; `a.is_deleted = false` is applied only by the outer join. A priority-1 member whose tag row has no live artifact row shadows a live priority-2 copy and is then eliminated — the tag disappears entirely. That state is documented production behaviour: retention soft-deletes `artifacts` without touching `oci_tags` (`lifecycle_service.rs:144-151`), `ArtifactService::delete` never removes the tag row (`artifact_service.rs:1985`), and the proxy manifest cache writes the two tables as independent best-effort statements (`oci_v2.rs:3719-3850`). Reproduced three ways: PR shape **0 rows**, join-then-dedupe **1 row**; through the real handler on the merged tree, `0 tags, total=0`. The flat listing filters `is_deleted` *inside* its subquery — this view's doc comment claiming parity with it is wrong.

2. **The Packages half never de-duplicates**, despite the issue's "priority-ordered, de-duplicated" expectation and the PR body's claim. `packages.rs:308-323` / `:341-349` have no `DISTINCT ON` and no `array_position`; the `ORDER BY vrm.priority` in `resolve_package_filter` (`:138-141`) feeds `= ANY(…)`, a set test, so it's dead. Reproduced: `express 1.0` in two members → two rows, both stamped with the virtual's key, `total: 2` — while a pull through the virtual resolves only the priority-1 copy. The Maven grouped catalog one file away (`repositories.rs:6285`, `:6316-6341`) does this correctly over the same table.

3. **The keyset early-exit #2520 built is destroyed for every docker repo, virtual or not.** `array_position($1::uuid[], t.repository_id)` as a sort key is non-indexable, so the non-virtual branch (`repositories.rs:6886`) takes the same path. Measured on a 50k-tag repo: main `Index Only Scan` **0.12–0.24 ms** per page; PR, *same single repo*, `Sort` over ~20k rows **7.6–22 ms**; virtual over 10 members, parallel seq scan + external merge to disk, **61 ms** (114 ms with `count=exact`). One `if` fixes it — gate the aggregating SQL on `repo_type == Virtual`, as the flat listing already does (`:5039` vs `:5063`).

4. **New anonymous existence oracle on private virtual repos.** `list_packages` applies no `require_visible` to the repo named by `?repository_key=`; `resolve_package_filter` (`packages.rs:85`) now expands it to members regardless of whether the *parent* is visible, so a private virtual with a public member returns content where a nonexistent key returns empty — distinguishable. Same class on the net-new `GET /api/v1/packages/{id}?repository_key=` (`package_in_virtual_repo`, `packages.rs:158`), which answers membership with no ACL on the virtual. This is the #1808 / GHSA-fv45-mwhh-q23r shape that #3525 has a byte-identical-response test for; apply the same guard.

5. **`fetch_index_child_sizes` double-counts multi-arch sizes** — widened to `ANY($1)` while keeping `GROUP BY parent_digest`, so an index mirrored in K members reports K× its size (measured 1000 vs 500; 6000 vs 3000).

6. **The new integration test never runs.** It is `#[ignore]`d but absent from CI's `--test` allow-list (`ci.yml:1251-1272`), so both regression tests are dead as merged (#3494 is the milestone issue about exactly this set).

### Not claimed by the issue, for the record
`GET /:key/storage/tree` and `GET /:key/versions/*path` remain empty for virtuals.

### Bookkeeping
No `## [Unreleased]` bullet for #3532/#3535 — needed before the cut. Given the rebase hazard in the blocker, I'd suggest rebasing first, then addressing 1–4 (2 and 3 are each a few lines; the Maven grouped view is the template). Happy to push the `docs(changelog)` commit when the code is settled. Milestoning 1.9.0.

brandonrc:
Same check-in as on #3536, with one extra note specific to this PR.

The audit's first item was that the branch no longer builds against main: four `sqlx::AssertSqlSafe` conflicts from main's sqlx 0.9 bump (#2857), plus two errors the conflicts do not surface. Main has moved a good deal further since then — 1.9.0 shipped yesterday — so a rebase is needed before any of the other findings can even be evaluated.

If you would rather not do that rebase, say so and I will take it over: rebase onto main, apply the fixes from the audit, keep your commits and authorship, and note in the PR exactly what I changed. The tag-erasing `DISTINCT ON` finding is the one I would want to talk through first either way, since it is a behaviour question rather than a mechanical fix.

brandonrc:
@xinnj — heads up: I pushed to your branch (`maintainer_can_modify` was on) rather than opening a competing PR. It was conflicting against `main` and did not build there, and it needed the CHANGELOG entry plus one substantive fix before it could go into 1.10.0. Force-pushed after a rebase, so please `git fetch && git reset --hard xinnj/fix/virtual-repo-aggregate-member-listings` before you touch it again. Sorry for the rewrite.

## 1. Rebase onto `main`

The branch was based on `a9f38766` (pre-1.9.0). Two conflicts, both from `main` bumping sqlx to 0.9, which requires dynamically-built SQL to go through `sqlx::AssertSqlSafe`:

- `packages.rs` — the page and count query construction. Resolved as **main's wrapper + your member-id bind**: `sqlx::query_as::<_, PackageRow>(sqlx::AssertSqlSafe(&*page_sql)).bind(filter.repo_ids.clone())`, same for the count arm.
- `repositories.rs` — the same collision in `fetch_docker_tag_rows` and `count_docker_tag_rows`. Resolved the same way: `sqlx::query(sqlx::AssertSqlSafe(&*sql)).bind(repository_ids)`.

`main`'s #3500 work (the `docker_tag_search_sql` LIKE-escaping fragment) auto-merged cleanly and is kept. One non-conflicting rebase adaptation was needed: `test_docker_tag_search_treats_wildcards_literally_3500` arrived on `main` after you branched and calls both helpers with a single `Uuid`, so it now passes `&[repo_id]`. Nothing of `main`'s was reverted.

## 2. CHANGELOG

The branch had no entry. Added one bullet under `## [Unreleased]` → `### Fixed`, leading with `(#3532)`. Worth noting for next time: because the branch predates the 1.9.0 cut, an entry written on it would have anchored to a heading the cut renamed and landed *inside the released 1.9.0 section* with no merge conflict — so it is worth re-checking placement after any rebase (`awk '/^## \[/{h=$0} /^- \*\*/{print h}' CHANGELOG.md | sort -u`).

## 3. The substantive fix: the tag dedup could erase a tag entirely

This is the one behavioural change I made. `DOCKER_TAG_ROWS_FROM_SQL` de-duplicates `oci_tags` **before** the `artifacts` join that proves the manifest is live, and that join is an INNER join:

```sql
FROM (
    SELECT DISTINCT ON (t.name, t.tag) ...
    FROM oci_tags t
    WHERE t.repository_id = ANY($1)
      AND POSITION(':' IN t.tag) = 0
    ORDER BY t.name, t.tag, array_position($1::uuid[], t.repository_id)
) t
JOIN artifacts a
  ON a.repository_id = t.repository_id
 AND a.path = 'v2/' || t.name || '/manifests/' || t.tag
 AND a.is_deleted = false
```

So the election runs before liveness is known. **Rows dropped:** any `(name, tag)` present in two or more members where the *highest-priority* member's `oci_tags` row has no matching live `artifacts` row at `v2/<name>/manifests/<tag>`. That member wins the `DISTINCT ON`, the join then discards it, and the lower-priority member's live copy was already eliminated — so the tag disappears from the page *and* from `?count=exact`, even though the flat listing and `GET /v2/<name>/manifests/<tag>` both still resolve it. It is reachable rather than theoretical: for a proxy-cached manifest the tag row and the artifacts row are written by two separate best-effort statements (#1357), and the surrounding comment says as much.

Reproduced on a real Postgres with two members for `app:latest`, the priority-1 member's manifest artifact soft-deleted: **0 rows**, while `list_for_repos_page`'s dedup over the same data returns the priority-2 member's live copy.

The flat listing avoids this by testing `a.is_deleted = false` **inside** its `DISTINCT ON (a.path)` subquery, so a dead higher-priority row can never shadow a live lower-priority one — which is exactly the contract your doc comment cites. The fix matches it, minimally: an `EXISTS` liveness predicate inside the de-duplicating subquery. The outer join, the aliases, and every caller are untouched, and for a non-virtual repository the id set is a singleton so nothing changes there.

Regression test: `test_docker_tag_dedup_ignores_member_without_live_manifest` (in `docker_tag_search_escape_tests`, next to the #3500 test, so it runs in the normal DB-backed lib suite rather than behind `--ignored`). Stashing the `EXISTS` hunk makes it fail with `left: []` / `right: ["latest"]`.

## 4. Verification

Against Postgres 16 with all 216 migrations applied:

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(/^api::handlers::repositories::/) or test(/^api::handlers::packages::/) or test(/^api::handlers::oci_v2::/)'` — **1214 passed**.
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(docker_tag) or test(virtual) or test(package)'` — **855 passed**, including the #3163 member-visibility tests and the `virtual_storage_usage` dedup/diamond/cycle set.
- `cargo nextest run --test virtual_docker_grouped_listing_test --run-ignored all` — 3 passed. I added `virtual_docker_grouped_view_unions_disjoint_members`: two members with one shared and two distinct images, asserting the union in `(image, tag)` order, the shared image collapsing onto priority 1, and `?count=exact` agreeing.

**Member expansion is not recursive**, which you may want to note in the issue. `proxy_helpers::fetch_virtual_members` is a single-level `INNER JOIN virtual_repo_members`, so a virtual repository nested inside another contributes nothing — probed end to end through the handler: outer virtual → inner virtual → local member with one tag lists **0 rows**. I deliberately left that alone: the flat listing (`list_artifacts`) and the Maven component grouping call the same helper, so the grouped view now matches them exactly, which is what #3532 asks for. Making expansion recursive is a platform-wide change across every listing path and belongs in its own issue — nesting is permitted and cycle-checked at membership-set time, so it is a real configuration.

Nothing else was touched. The packages-side changes (`resolve_package_filter`, `package_in_virtual_repo`, the `?repository_key=` detail context) are yours as written, only rebased.

brandonrc:
Filed the nested-expansion limitation as #3840 so it is not lost — it is out of scope here and this PR does not make it worse, it just brings the grouped view into line with the flat and Maven paths that already behaved this way. Nothing needed from you on it.

CI is running on `15f9486`; I will merge once it is green.

brandonrc:
CI caught one thing I missed: `backend/tests/virtual_docker_grouped_listing_test.rs` is new in this branch and no workflow named it, so the `every_integration_target_is_wired_into_ci_or_exempt` gate (#3494) failed the unit-test job — correctly, since the suite would otherwise never execute anywhere. Pushed `b5e325ba` adding it to the Tier 2 `-- --ignored` step next to the other Postgres-only virtual-repository suites. Verified the gate passes locally (7/7 in `ci_test_surface`).

Issue #2596, opened 2026-07-16
Docker re-migration of a hollow repo is a no-op: duplicate-skip bypasses the referenced-content walker, blobs never fetched


brandonrc:
## Summary

PR #2582 (shipped in v1.5.8) fixed the hollow-image import bug from #2457 by adding a referenced-content walker that fetches config/layer blobs after a Docker/OCI manifest is registered. The startup reindex correctly flags repos whose tags remain hollow and states that those repos need re-migration.

Re-migration does not repair them under the default settings. The duplicate check in the migration worker decides skip/transfer purely from the manifest artifact row's checksum. A hollow repo has all its manifest rows present with matching checksums, so every manifest is skipped and `process_single_artifact` never runs. The referenced-content walker only runs inside that path, so no blobs are ever fetched. The job then reports success.

## Evidence (production logs, Nexus Docker hosted source, server pre-1.5.8)

Startup reindex on a hollow repo:

```
INFO  oci_migration_reindex: OCI migration reindex: registering migrated Docker/OCI manifests candidate_count=899
WARN  oci_migration_reindex: referenced blob has no artifacts row; layer will 404 until re-migrated
      repository_id=ff0d4cf5-... manifest_path=docker-repo/v2/... blob_digest=sha256:...
```

9,498 missing-blob warnings across 526 distinct manifests in one repository.

Re-run of the migration against the same still-online Nexus source:

```
INFO  migration_worker: Starting migration job processing job_id=d7a8d397-ea15-426a-80d0-3217decf7e51
INFO  migration_worker: Destination repository already exists with matching type+format; reusing repo=docker-repo
DEBUG migration_worker: Skipping duplicate artifact (already exists with matching checksum) ... (x888)
INFO  migration_worker: Migration job completed job_id=d7a8d397-... completed=11 failed=0 skipped=888
```

Zero `oci_referenced_content` activity in the run. The repo remains hollow and `docker pull` still 404s on every layer, but the job reads as a clean success.

## Root cause

`backend/src/services/migration_worker.rs`:

- `check_artifact_duplicate` (around line 908) matches only the `artifacts` row for the manifest path and compares sha256/sha1. It has no knowledge of whether the manifest's referenced OCI content (`oci_blobs`, child manifests) exists.
- `decide_duplicate_match` with the default `ConflictResolution::Skip` returns true on checksum match, and the caller `continue`s (around line 583) before `process_single_artifact`, so the #2582 walker never executes for that item.

[...]

PR #3536, opened 2026-08-24
fix(migration): migrated Docker/OCI artifacts render with wrong path, name, and type

### LINKED ISSUE #3533: Migrated Docker/OCI artifacts render with wrong path, name, and content type
Docker/OCI artifacts imported from Nexus/Artifactory are stored with the legacy repo-prefixed path, the filename (tag) as the name, a NULL version, and a hardcoded `application/octet-stream` content type. On the artifacts page this renders as a redundant repo name in the path/download URL, a tag-only name, and the wrong content type.

Expected: the artifacts-row identity is derived from the OCI role (manifests use the live push path `v2/<image>/manifests/<ref>` and an `<image>:<ref>` name with the sniffed media type; blobs don't produce an artifacts row), matching the live push path.

xinnj:
## Summary

Docker/OCI artifacts imported from Nexus/Artifactory were stored with the legacy repo-prefixed path, the filename (tag) as the name, a NULL version, and a hardcoded `application/octet-stream` content type — so they rendered on the artifacts page with a redundant repo name in the path/download URL, a tag-only name, and the wrong content type. This derives the artifacts-row identity from the OCI role, matching the live push path.

Closes #3533

## Test Checklist

- [x] Manually tested locally

## API Changes

- [x] N/A - no API changes

### DIFF
--- a/backend/src/services/migration_worker.rs
+++ b/backend/src/services/migration_worker.rs
@@ -1419,8 +1419,47 @@ impl MigrationWorker {
         conflict_resolution: ConflictResolution,
         package_type: &str,
     ) -> Result<bool, MigrationError> {
+        // Docker/OCI blobs no longer produce an `artifacts` row (they live
+        // only in `oci_blobs`, matching the live push path), so their
+        // duplicate check is against `oci_blobs` keyed by digest rather than
+        // the `artifacts` checksum the manifest path uses. Blobs are
+        // untouched by the #2596 completeness gate — a matching digest still
+        // skips under `Skip`, and `Overwrite`/`Rename` still re-process.
+        if is_oci_package_type(package_type) {
+            if let OciRole::Blob { digest } = classify_oci_source_artifact(artifact_path) {
+                let existing: Option<String> = sqlx::query_scalar(
+                    "SELECT b.digest FROM oci_blobs b \
+                     JOIN repositories r ON r.id = b.repository_id \
+                     WHERE r.key = $1 AND b.digest = $2",
+                )
+                .bind(repo_key)
+                .bind(&digest)
+                .fetch_optional(&self.db)
+                .await?;
+                return Ok(existing.is_some()
+                    && decide_duplicate_match(
+                        expected,
+                        digest.strip_prefix("sha256:").unwrap_or(&digest),
+                        None,
+                        conflict_resolution,
+                    ));
+            }
+        }
+
         // Match artifacts in the same repository by repository-relative path.
-        // Keep a fallback for legacy rows where path was saved as repo-prefixed.
+        // Keep a fallback for legacy rows where path was saved as repo-prefixed,
+        // plus the canonical `v2/<image>/manifests/<reference>` shape Docker/OCI
+        // manifests are now stored under (the live push path's layout).
+        let oci_canonical_path = if is_oci_package_type(package_type) {
+            match classify_oci_source_artifact(artifact_path) {
+                OciRole::Manifest { image, reference } => {
+                    Some(format!("v2/{}/manifests/{}", image, reference))
+                }
+                _ => None,
+            }
+        } else {
+            None
+        };
         let existing: Option<(String, Option<String>)> = sqlx::query_as(
             r#"
             SELECT a.checksum_sha256, a.checksum_sha1
@@ -1428,14 +1467,19 @@ impl MigrationWorker {
             JOIN repositories r ON r.id = a.repository_id
             WHERE r.key = $1
               AND a.is_deleted = false
-              AND (a.path = $2 OR a.path = $3)
-            ORDER BY CASE WHEN a.path = $2 THEN 0 ELSE 1 END
+              AND (a.path = $2 OR a.path = $3 OR ($4::text IS NOT NULL AND a.path = $4))
+            ORDER BY CASE
+                WHEN a.path = $2 THEN 0
+                WHEN a.path = $4 THEN 1
+                ELSE 2
+            END
             LIMIT 1
             "#,
         )
         .bind(repo_key)
         .bind(artifact_path)
         .bind(legacy_source_path)
+        .bind(oci_canonical_path.as_deref())
         .fetch_optional(&self.db)
         .await?;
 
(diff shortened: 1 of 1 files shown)


brandonrc:
## Audit round 1 — migrated Docker/OCI artifact rows (3 reviewers: correctness / security+integration / architecture)

All three passes read head `3656c2d6`, merged `origin/main` (39 commits) into a throwaway branch — **zero conflicts, `cargo check --lib --tests` exit 0**, `migration_worker.rs` untouched on main since the merge base — and ran the migration-worker suite against a live Postgres: **205/205 pass**, including the eight new pure unit tests and every DB-backed Docker import test. The DB-backed Artifactory-layout test is a genuine regression test (its new assertions fail on main at `assert_eq!(artifact_count.0, 3)`).

**What checks out.** `migration_artifact_identity` produces `artifacts` rows byte-identical to the live `PUT /v2/<name>/manifests/<ref>` path on every load-bearing column — `path`, `name`, `version`, `content_type`, `size_bytes`, `storage_key`, `checksum_sha256` — and `oci_tags` / `manifest_blob_refs` / `oci_manifest_refs` / `packages` were already identical. Security is sound: digests are computed from downloaded bytes and path-named digests must match; the composed path hits `has_unsafe_path` before the INSERT; blob keys are fixed-length digest-derived so there is no 1024-byte key pressure. Two fixes the PR doesn't claim but does deliver, both verified in SQL: the WebUI docker-tags panel (`a.path = 'v2/' || t.name || '/manifests/' || t.tag`) was **empty for every migrated Docker repo** and now populates; and a child manifest enumerated as its own item no longer produces two same-checksum rows that `index_child_artifact_size_sum` counted twice. jscpd reproduced CI's way: 0.62% vs 0.66% at the merge base — but the gate copies only changed files, so it cannot see the duplication below.

### Should-fix

1. **Reuse `oci_v2::upsert_manifest_artifact` instead of the hand-rolled INSERT at `migration_worker.rs:1984-2013`.** That function (`oci_v2.rs:9384`) already composes the identical path/name/version/checksum and upserts on `(repository_id, path)`; its live-push caller says "*Shared with the migration import path … so a pushed and a migrated manifest are recorded identically*" (`oci_v2.rs:9587-9590`) — now false. And the two writers collide for real: the sibling migration module `services/oci_referenced_content.rs:262-274` already calls the shared upsert for child manifests inside the same item transaction, and Artifactory's `<image>/sha256__<hex>/manifest.json` classifies to `reference: "sha256:<hex>"`, so both write byte-identically the same path. They differ on `checksum_sha1`, `uploaded_by` and whether `name` updates on conflict, so the surviving row depends on item ordering. The reuse is cheap: generic over `PgExecutor`, every argument in scope at `:1951`; the only loss (`sha1`) is behaviourally free because `decide_duplicate_match` treats NULL sha1 as a duplicate (`:3521`).

2. **Drop or narrow `migration_content_type` (`:3136`).** It retypes *every* non-OCI format — scope beyond #3533 — and it is a fourth Maven MIME table that contradicts the publish path: a migrated `.tar.gz` becomes `application/zip` where `mvn deploy` stores `application/octet-stream`; `.asc` → `application/pgp-signature` vs `text/plain`; `.aar` → java-archive vs octet-stream (`maven.rs:863-873`; sbt diverges further, `sbt.rs:275-280`). The npm/helm arms match the live handlers and are fine. Narrow fix: make `maven::content_type_for_path` `pub(crate)` and call it, or keep the OCI arm only. Related: the new `content_type = EXCLUDED.content_type` in the DO UPDATE lets a re-migration downgrade a natively-published `.whl` from `application/zip` to `application/octet-stream`.

3. **Only `docker` and `oci` are reached.** `is_oci_package_type` (`:3276`) hardcodes two names; the repo's canonical predicate is `format_handler_key(&format) == "oci"` (`oci_v2.rs:2787-2790` → `models/repository.rs:227-243`), which covers `podman | buildx | oras | wasm_oci | helm_oci`. Those imports still show the reported symptom.

4. **Nothing repairs already-migrated repositories, and a re-run is a silent no-op.** Every deployment hit by #3533 already has the bad rows. Replaying the PR's `check_artifact_duplicate` SELECT against the exact legacy row main writes: it matches via the `$3` arm (source key == target key, the common case), the checksum matches, the #2596 completeness gate passes → item **skipped**, canonical row never written. Under `overwrite` the corrected row lands *beside* the legacy one; nothing deletes the old manifest or per-blob rows, and via the `ak_usage_ledger_artifacts` trigger those stale blob rows keep billing to `hosted_bytes` while `oci_blobs` bills the same bytes to `oci_bytes`. #2457 — the same class of defect in this worker — shipped its write-path fix together with its repair (`services/oci_migration_reindex.rs`); this ships only the write half. Either add the repair, or the CHANGELOG must carry a forward-only upgrade note (drafted below either way).

5. **An over-long source `mediaType` now aborts the whole item transaction.** `artifacts.content_type` is `VARCHAR(255)` but the value the PR newly binds there comes from `sniff_manifest_media_type`, which only checks `HeaderValue::from_str` — no length bound anywhere in the chain. A 300-char `mediaType` fails with a raw Postgres `value too long` and rolls back the tag, its `oci_blobs` rows and `manifest_blob_refs`. On main the artifacts row got the literal `application/octet-stream` and the sniffed value went only to `oci_tags.manifest_content_type` (`TEXT`). Truncate or fall back.

### Not this PR, but adjacent
- **#3654 shares the exact root cause**: `migration_artifact_path`'s version-less fallback `format!("{}/{}", repo_key, artifact_path)` (`migration_worker.rs:3057`). This PR routes Docker/OCI manifests around that line and leaves it intact for `generic`/raw — correct scoping, but worth saying on #3654.
- Nits: blobs no longer pass `has_unsafe_path` (not exploitable — digests validated); `parse_name_and_version` computed twice; `checksum_sha1` can go stale when a later `docker push` upserts a migrated row.

### Bookkeeping
No `## [Unreleased]` bullet for #3533/#3536 — the release preflight blocks a cut without one. I'll push the `docs(changelog)` commit (with the no-repair upgrade note) once 1–3 and 5 are in. Milestoning 1.9.0.

brandonrc:
Checking in on this one. The audit for the migrated Docker/OCI artifact rows has been sitting since 2 September with no reply, and 1.9.0 shipped yesterday — I would like to get this into 1.10.0 rather than let it age further.

brandonrc (dismissed):
## Review — changes requested

Thanks for this, and for picking the right bar: the claim to test isn't "is the row correct in the abstract" but "is a migrated manifest indistinguishable from a pushed one." That claim **holds, field for field**, and I tried hard to break it and couldn't. The problems are all around it — the fix is inert on the repositories that actually have the bug, and it introduces a uniqueness collision that didn't exist before.

### What I verified and could not break

Comparing the `OciRole::Manifest` arm against `upsert_manifest_artifact` (`backend/src/services/oci_v2.rs:9384-9426`):

| field | live push | migration | |
|---|---|---|---|
| `path` | `v2/{image}/manifests/{reference}` `:9398` | same, `migration_worker.rs:3105` | ✅ no repo prefix on **either** side |
| `name` | `{image}:{reference}` `:9399` | same, `:3106` | ✅ |
| `version` | `Some(reference)` `:9417` | same, `:3107` | ✅ |
| `checksum_sha256` | digest minus `sha256:` `:9400` | `sha256_hex` `:1719` | ✅ |
| `storage_key` | `manifest_key` | `manifest_storage_key` `:1867` | ✅ |
| `size_bytes` | `manifest_total_size(&body)` `:9584` | same fn via `:3185-3188` | ✅ |
| `content_type` | `stored_media_type_for(…)` | same, header `None` `:1934-1944` | ✅ |

The size case is the one I most expected to be wrong, and it's right: an index has no config or layers so `manifest_total_size` is ~0 — but the live push *also* stores that ~0 on the artifacts row and reserves `index_child_artifact_size_sum` for the catalog row (`:9637`). You mirror both halves. Media types are the correct spec strings, Docker schema2 and OCI image manifest are **not** conflated, and a Docker manifest *list* keeps its own type rather than being rewritten to `…index.v1+json`, because `is_index_content_type` (`oci_v2.rs:2103-2115`) accepts the Docker string. Blobs correctly produce no artifacts row, and the blob duplicate check being re-keyed onto `oci_blobs` by digest (`:1428-1446`) is the right consequence of that.

I also revert-proved your test rather than trusting green: with the production change reverted to the base shape and the test kept, it **fails** at `migration_worker.rs:8251` with `left: 3, right: 1` — the three legacy rows the fix collapses to one. That's a real guard.

---

### F1 (High) — the fix never repairs an already-migrated repository, and nothing says so

This is an import-time transform only. Nothing rewrites existing rows, and `oci_migration_reindex.rs:189-210` reconciles *tags*, never paths. The part that matters: **a re-run doesn't help either.**

Under the default `ConflictResolution::Skip`, `check_artifact_duplicate` still finds the legacy row through the `legacy_source_path` fallback — `$3` bound at `migration_worker.rs:1481`, matched at `:1469` — the checksum matches, the #2596 completeness gate passes for a healthy repo, and the item is **skipped**. The canonical row is never written. You added `$4` for the new canonical shape (`:1453-1461`) but left the legacy shape as a positive duplicate match. `$3` is `<source_key>/<path>` while the pre-fix write used `<target_key>/<path>`, so this bites whenever source and target repository keys match — the normal case.

Under `Overwrite`/`Rename` the canonical row does get inserted, but the legacy row is never removed, so the UI shows a correct row *and* a stale broken one for the same manifest, and storage/quota/search double-count. Previously-migrated blob rows survive forever too — meaning the spurious flat-listing entries your own doc comment at `:3059-3066` says must not exist stay put.

There's no backfill migration in the diff, nothing in the PR body or #3533 acknowledges existing rows, and I couldn't find a filed backfill issue. Pick one: add the legacy shape to the reprocess decision (treat a legacy-shaped path as *not* a duplicate, and soft-delete it when the canonical row lands), ship a backfill, or at minimum state the limitation in the PR body and file the issue. I'd rather it not merge silently implying migrated repos are fixed when they aren't.

### F2 (High) — two source manifests for one `(image, reference)` now collide silently

`classify_oci_source_artifact` maps **both** `<image>/<ref>/manifest.json` and `<image>/<ref>/list.manifest.json` to the same `OciRole::Manifest { image, reference }` (`:3356-3367`). Pre-fix those produced two distinct rows; post-fix both derive `path = v2/<image>/manifests/<ref>` (`:3105`).

`artifacts` carries exactly one unique key — `UNIQUE(repository_id, path)`, `backend/migrations/004_artifacts.sql:21`, never dropped or replaced across all 191 migrations, and there's no unique key on `(repository_id, name, version)`, `checksum_sha256` or `storage_key`. The INSERT is `ON CONFLICT (repository_id, path) DO UPDATE` (`:1988-2018`), so the later arrival **silently overwrites** the earlier one's checksum, size, storage_key and content_type. `persist_tag_and_refs_in_tx` (`:2089`) then re-points `oci_tags` to whichever manifest was imported last — and import order comes from the source listing, so a multi-arch tag can end up resolving to a single-arch manifest non-deterministically.

The same shape applies to a Nexus export enumerating both `v2/<image>/manifests/<ref>` and `v2/-/<image>/manifests/<ref>`, since the `-` catch-all is stripped at `:3378-3380`.

Please detect it rather than merge it: a second *distinct* `checksum_sha256` arriving for an existing canonical path should fail or warn the item. Silent last-write-wins on a manifest identity is the kind of thing that shows up months later as "the tag points at the wrong arch."

For the record, the `packages` projection is **not** affected — its key is `UNIQUE (repository_id, name)` and `migration_catalog_entry` (`:3240-3250`) still uses `name = image` / `version = reference`, identical to the live push. I also checked #3816, which reworks a catalog projection concurrently: it touches none of these files, so there's no conflict to coordinate.

### F3 (Medium) — reimplements `upsert_manifest_artifact` instead of calling it, and already diverges

`oci_v2.rs:9586-9589` says the row is *"Shared with the migration import path via `upsert_manifest_artifact` so a pushed and a migrated manifest are recorded identically."* After this PR that comment is false — the migration hand-writes its own INSERT at `:1988-2018`. They already differ: `checksum_sha1` written vs not, `name` updated on conflict vs not, `uploaded_by` NULL vs `Some(claims.sub)`, and no `apply_upload_hold_hosted` (`grep -n quarantine migration_worker.rs` is empty, so a migrated manifest bypasses the quarantine hold a pushed one gets).

`upsert_manifest_artifact` is `pub(crate)` and generic over `E: PgExecutor<'e>`, so it takes `&mut *tx`. Call it. The parity this PR is built on should be structural, not maintained by copy-paste — otherwise the next change to the push path silently re-opens #3533.

### F4 (Medium) — the non-OCI content-type table is out of scope and has a wrong mapping the test locks in

#3533 is Docker/OCI, but this also rewrites stored `content_type` for maven/gradle/sbt/ivy/npm/yarn/pnpm/bower/helm (`:3143-3167`). `.tar.gz` under the maven arm maps to `application/zip` (`:3155`) — a gzipped tar is not a zip, and the npm/helm arm maps the same extension to `application/gzip` (`:3164`), so the function contradicts itself. `test_migration_content_type_maven_extensions` then enshrines it. Either drop the non-OCI table from this PR or fix `.tar.gz` → `application/gzip` and the assertion with it.

### Lower

- **F6** — coverage is one role, one layout. Not covered end to end: an OCI image manifest, a multi-arch index (`list.manifest.json` → index media type + `index_child_artifact_size_sum`), a digest-referenced child manifest, the Nexus `v2/...` layout. Those exist only as pure-fn tests that never reach the INSERT. Given F2 lives exactly in the index case, an index test would be worth adding here.
- **F7** — no CHANGELOG entry. `grep -n 3533 CHANGELOG.md` is empty; add an issue-number-first bullet under `## [Unreleased]`. Our release cut reconciles that section, so without it the fix won't appear in the notes.
- **F8** — `parse_name_and_version` now runs twice per non-OCI artifact (`:3122` and `:2168`). Cost is negligible; the hazard is that identity and catalog entry are derived from two separate calls. Threading the `ParsedArtifact` out removes it.
- **F9** — `oci_migration_reindex.rs:281-284` and `:327-331` still describe stripping a `<repo_key>/` prefix "the migration writes." It doesn't, after this PR. The logic survives (the `LIKE '%/manifests/%'` match still fires and `strip_repo_prefix` becomes a no-op), so this is drift, not breakage — but it's drift in the one module whose job is cleaning up after the code you changed.
- Edge worth knowing: an Artifactory `sha256__<hex>` reference folder whose hex isn't exactly 64 lowercase hex fails `normalize_digest_segment` (`:3289-3293`) and falls through to the literal string (`:3360`), which is a valid Docker tag — so it now becomes the `path`, the `name`, and a catalog version instead of being rejected. Pre-existing, but newly load-bearing on identity.

---

**What I ran:** 8/8 new pure-fn tests; the Artifactory-layout DB regression test against a real Postgres with `AK_TESTS_REQUIRE_DB=1` set deliberately so a skip would fail rather than go fiction-green (it passed, 1.7s); all 191 migrations applied clean; the revert-proof above. Worth flagging that `let Some(pool) = tdh::try_pool().await else { return; }` at `:8165` means that guard silently passes with no `DATABASE_URL` — pre-existing pattern, but it's only a real guard where the harness runs.

**What I reasoned from source rather than executed:** the parity table, F1's skip path (traced through `check_artifact_duplicate`/`decide_duplicate_match`; I didn't seed a legacy row and re-run a job), and F2's collision (derived from the classifier plus the single unique key; I didn't build a source export containing both filenames under one tag folder — how often a real Artifactory export emits both is unverified, though the collision is unguarded either way).

F1 and F2 are what I need resolved before merge. F3 I'd like done now while the context is fresh, since it's the thing that keeps #3533 from coming back. F4 just needs a decision either way.

brandonrc:
Taking this over as described on 9 September (option 2), since there has been no reply to the 2 September audit or the 12 September review and it is the last open item in the 1.10.0 milestone.

What I am going to do, on top of the two existing commits (authorship untouched):

- Rebase onto current `main` (161 commits behind; the rebase was clean).
- **F3** — replace the hand-rolled `artifacts` INSERT with the shared `upsert_manifest_artifact`, so a pushed and a migrated manifest are written by the same code path and the parity comment in the push path is true again.
- **F1** — make a re-run repair a repository that was migrated before this fix: a legacy-shaped row (`<repo_key>/<image>/<ref>/manifest.json` and friends) is no longer treated as a duplicate of the canonical `v2/<image>/manifests/<ref>` row, and it is removed when the canonical row lands.
- **F2** — detect, rather than silently overwrite, a second distinct manifest digest arriving for the same `(image, reference)` path within one job.
- **F4** — drop the non-OCI content-type table; the Maven/npm/Helm publish paths already own their MIME mapping and the `.tar.gz → application/zip` entry contradicts them.
- Cover the `format_handler_key(...) == "oci"` family (podman / buildx / oras / wasm_oci / helm_oci), not just the two literal names.
- Bound the sniffed `mediaType` before it reaches `artifacts.content_type` (`VARCHAR(255)`) so a long value cannot abort the item transaction.
- Add an index-layout DB test alongside the Artifactory one, refresh the `oci_migration_reindex` comments the review flagged as drifted, and add the CHANGELOG bullet.

I will note here exactly what changed once it is pushed. Push back at any point if any of this reads as the wrong call.

brandonrc:
Pushed the takeover on top of your two commits (authorship intact), rebased onto current `main`. What changed, mapped to the review:

- **F3** — the `OciRole::Manifest` arm now calls `oci_v2::upsert_manifest_artifact` (the same function the live manifest-PUT path and `oci_referenced_content` use) instead of the hand-rolled INSERT, so the "recorded identically" comment in the push path is true again and a child manifest imported both as its own item and by the walker lands in one row. The `NotOci` arm keeps `main`'s INSERT byte-for-byte (sha1 recorded, `content_type` not refreshed on conflict).
- **F1** — `check_artifact_duplicate` no longer treats a legacy `<repo_key>/<source path>` row as a duplicate of a Docker/OCI manifest: a manifest is a duplicate only under its canonical `v2/<image>/manifests/<reference>` path. When the canonical row lands, `retire_legacy_migration_rows` soft-deletes the pre-fix manifest row and (in the blob arm) the pre-fix per-blob rows in the same transaction — soft-delete is `artifact_service::delete`'s own mechanism, so the usage ledger and GC follow. Re-running a migration is therefore the repair; there is no separate backfill.
- **F2** — `reject_conflicting_manifest_identity`: when a *different* digest arrives at an already-written canonical path, the worker looks for a completed sibling item of the same job whose source path classifies to the same `(image, reference)`; if one exists the item fails naming both source paths and both digests. Without such a sibling the existing row predates the job (tag moved upstream, or a native push) and the overwrite is the re-migration the operator asked for. `transfer_artifact` takes an `Option<Uuid>` item id to scope this.
- **F4** — `migration_content_type` is gone; non-OCI rows keep `application/octet-stream` as before this PR.
- **Scope** — `is_oci_package_type` resolves through `RepositoryFormat::handler_key() == "oci"`, so `podman`/`buildx`/`oras`/`wasm_oci`/`helm_oci` take the OCI path.
- **Audit 5** — `bounded_manifest_media_type` falls back to the canonical media type for the manifest's class when the sniffed `mediaType` exceeds `VARCHAR(255)`.
- **F6** — `test_docker_import_multiarch_index_nexus_layout` now asserts the `artifacts` rows (index under its tag, child under its digest, exactly two rows); new DB tests `test_docker_reimport_repairs_pre_3533_legacy_rows` (F1 end to end, including the duplicate check before and after) and `test_docker_import_rejects_second_manifest_for_same_reference` (F2, with a seeded job and both Artifactory manifest files).
- **F7/F8/F9** — CHANGELOG bullet under `## [Unreleased]` → `### Fixed` (credits you); `(name, version)` is parsed once and threaded into both the row and the catalog entry; the two `oci_migration_reindex` comments now describe both path shapes.

Verified locally: `cargo clippy --lib --bins --tests -- -D warnings` clean; the migration-worker, `oci_migration_reindex` and `oci_referenced_content` suites against a fresh Postgres 16 with `AK_TESTS_REQUIRE_DB=1` (so a missing DB fails rather than skips).

Issue #3464, opened 2026-08-20
release/1.7.x has diverged from main's release engineering: pushes publish nothing, no backport can pass the gate, and the branch cannot be preflighted


brandonrc:
GitHub executes a workflow file **at the ref that triggered it**, not at `main`. Every release-engineering fix of the last few weeks landed on `main`; 1.7.x releases are cut from `release/1.7.x`. The two have silently diverged, and the divergence is invisible from `main` — nothing on `main` is wrong, and nothing on `main` reports that the maintenance line is broken.

Verified on `release/1.7.x` today:

### 1. Commits to `release/1.7.x` publish no images at all

```yaml
# release/1.7.x  .github/workflows/docker-publish.yml
on:
  push:
    branches: [main, master, 'release/1.1.x']
```

`release/1.1.x` no longer exists. `main` has `release/**` (the #3422 fix), which is inert for this branch because the branch runs its own copy. So a commit landing on `release/1.7.x` triggers Docker Publish **never**.

That makes release-preflight check 3 correct every time it reports "no Docker Publish run exists for HEAD" (#3338) — and it means every 1.7.x cut has been preceded by an undocumented manual `workflow_dispatch` to create the images the tag needs to resolve a digest from.

`main` also derives the `X.Y-dev` floating alias from the branch name in the same change; on this branch it is still hardcoded to `1.1-dev`/`release/1.1.x`, so even with the glob fixed, `release/1.7.x` would publish a series-dev tag for the wrong series (i.e. none).

### 2. No compliant backport can pass the release-branch gate

`release/1.7.x` carries the gate's logic **inlined** in `release-branch-gate.yml`, in its pre-#3422 form: a commit passes only if it is on `main` or matches a `main` commit by patch-id.

Two shapes that a maintenance branch produces *by construction* can do neither:

- a **release prep** (`chore(release): prepare 1.7.7`) prepares a version that exists only on this branch — there is nothing on `main` to trace to, and there should not be;
- a **narrowed backport** — hunks that could not apply were resolved away (a guard test depending on scaffolding this branch does not have, a CHANGELOG hunk targeting `[Unreleased]` where the branch needs `[1.7.6]`), so patch-id cannot match.

[...]

PR #3542, opened 2026-08-24
ci(release): require a green Release Preflight for the commit being tagged

### LINKED ISSUE #3538: release.yml does not require a Release Preflight: the guard that would have saved v1.7.5 and v1.7.7 is optional
`RELEASING.md` step 1 has said "run the release preflight and get a `READY` before you tag" since #3043. Nothing enforces it. `release.yml` has no dependency of any kind on `release-preflight.yml`, and a guard's absence is invisible — only its failure is visible. Two of the most expensive incidents in the 1.7.x line are that shape, together ~46 hours of delay and two permanently burned version numbers.

### v1.7.5 — the check existed and was never run

`docker/scanner-adapter/VERSION` was not bumped when `Dockerfile.scanner-adapter` changed (a comment-only edit, #3429). Preflight **check 4** (#3339) exists precisely to catch that, and it was merged on 2026-08-13. The Actions API shows the gap:

```
$ gh api ".../actions/workflows/release-preflight.yml/runs?per_page=30" \
    --jq '.workflow_runs[] | "\(.created_at) \(.event) \(.status)/\(.conclusion)"'
2026-08-17T13:51:16Z schedule         completed/failure
2026-08-14T14:19:26Z schedule         completed/failure
```

Zero runs between 2026-08-14T14:19Z and 2026-08-17T13:51Z. The tag's Docker Publish hit the never-republish guard, the chain stopped, and because the tag is immutable the number was abandoned. Cost: 38h41m.

### v1.7.7 — the check ran, was right, and was ignored

- `2026-08-20T13:57Z`, scheduled run **32377238021**: `[FAIL] Docker Publish for HEAD did NOT cleanly publish the manifest — a tag cut WILL stall on the same gate.`
- `2026-08-20T21:22Z`, dispatches **32419067572** (main) and **32419076623** (release/1.7.x): both `[FAIL] … verdict does not exist yet`.
- `2026-08-20T21:40:17Z`: **`v1.7.7` was tagged**, 18 minutes later.
- 19 seconds after the tag push, the in-flight docker-publish concluded `failure`.

The pipeline called it correctly, hours early. Nothing required it to be listened to.

### The compounding factor: the verdict was about a different tree

`release-preflight.yml` hardcoded `ref: main` in its checkout (#3464), so dispatching it from `release/1.7.x` audited **main's** tree. v1.7.5, v1.7.6 and v1.7.7 were each cut from `release/1.7.x` on a `READY` transcript that was a statement about another branch. That was fixed on `release/1.7.x` in v1.7.8 — and **never forward-ported to `main`**, which still has `ref: main` today. Any evidence-based gate on `main` inherits that hole unless it is closed first.

## What is missing

Nothing anywhere asserts *"a preflight ran against this commit."* Specifically:

- `release.yml`'s job graph is `release-preflight → resolve-candidate-digest → release-gate → build-binaries → verify-images-published → release`. `release-preflight` there is the *identity* preflight (moved-tag, existing-release); it has nothing to do with the Release Preflight workflow.
- The three integrity gates all run *after* `docker-publish` — i.e. after the point where an unbumped `VERSION` or a red Security Scan has already stalled the chain and spent the tag.

## Proposal

1. Forward-port the ref-awareness fix (`ref: ${{ inputs.ref || github.ref }}` + `PREFLIGHT_REF`) from `release/1.7.x` to `main`, so a preflight's verdict is about the tree it names. Partially closes #3464 for `main`.
2. Have a green Release Preflight run leave machine-readable evidence naming the sha it actually checked out (`git rev-parse HEAD` computed inside the run, not the API's `head_sha`, which is the sha of the ref the run was *dispatched* on).
3. Add a job at the head of `release.yml`, ahead of `resolve-candidate-digest`, that refuses to proceed without that evidence for `github.sha`. **`in_progress` must not pass** — that is the v1.7.7 shape exactly.
4. Preserve the preflight's `exit 1` (NOT READY) vs `exit 2` (INFRA) distinction: an indeterminate verdict must not read as a pass.
5. Provide a *documented, loud* break-glass with a required reason, per the freeze policy's "skip schedule, never integrity". A silent bypass is not acceptable; neither is no bypass at all.
6. Self-test every leg — including each leg that must block — wired into `shell-tests`, following `test-check-changelog-unreleased.sh`. A gate nobody has watched fail is the defect this issue is about.

## Scope note

`main` only. Whether `release/1.7.x` needs the same is a separate decision (see #3464): the branch's own `release.yml` runs for tags cut from it, and three of the last six releases were cut from that branch.

brandonrc:
## Summary

`RELEASING.md` step 1 has said *"run the release preflight and get a `READY` before you tag"* since #3043. **Nothing enforced it.** `release.yml` had no dependency of any kind on `release-preflight.yml`, and a guard's absence is invisible — only its failure is visible. The two most expensive incidents in the 1.7.x line are both that shape, together ~46 hours of delay and two permanently burned version numbers:

- **v1.7.5** — `docker/scanner-adapter/VERSION` was not bumped when `Dockerfile.scanner-adapter` changed. Preflight **check 4** (#3339) exists precisely to catch that, was merged 2026-08-13, and was simply **never run**: the Actions API shows zero Release Preflight runs between `2026-08-14T14:19Z` and `2026-08-17T13:51Z`.
- **v1.7.7** — the preflight **ran and was right**. Run `32377238021` (2026-08-20T13:57:42Z) printed `[FAIL] Docker Publish for HEAD did NOT cleanly publish the manifest -- a tag cut WILL stall on the same gate.`; two dispatches at 21:22Z reported the follow-up publish's `verdict does not exist yet`. `v1.7.7` was tagged at **21:40:17Z**, 18 minutes later, and that docker-publish concluded `failure` 19 seconds after the tag push.

Closes #3538. Partially addresses #3464 (the `ref: main` half, for `main`).

## What this does

A `preflight-evidence` job at the head of `release.yml`, ahead of `resolve-candidate-digest`, and named again in the `release` job's condition. **The rule is one sentence:** every Release Preflight run for the commit being tagged must have **completed**, and the newest completed one must have concluded `success` **and** left a `release-preflight-<sha>` artifact naming that commit.

Each clause is load-bearing, and each has a self-test that goes red if it is reverted:

| clause | why | reverting it breaks |
|---|---|---|
| every run must have **completed** | an in-flight run is a verdict that *does not exist yet* — the v1.7.7 shape exactly | 2 cases |
| the **newest completed** one decides | checks 1/3/4 measure state outside the tree, so a commit can legitimately go red after being green; honouring a superseded green re-imports "the red is cosmetic, promote anyway" at the level of run selection | 1 case |
| it must carry an **artifact naming this commit** | the API's `head_sha` is the sha of the ref a run was *dispatched* on, not the sha it checked out | 2 cases |
| an unreadable answer is **INFRA**, not a pass | preserves `release-preflight.sh`'s exit 1 / exit 2 distinction end to end | 1 case |
| an override needs a **reason** | a silent bypass is the thing being removed | 3 cases |
| ordering comes from `created_at`, not the API's response order | a stale-ordered listing must not flip the verdict | 2 cases |

## Design decision: require the evidence, don't re-run the preflight inline

The two options are running the preflight inline in `release.yml`, or querying the Actions API for evidence that it already ran green on `github.sha`. **Inline is simpler and cannot be stale, and it is still the wrong choice**, for a reason that only shows up when you ask what each option would have done to v1.7.5:

> An inline preflight runs **after the tag exists**. The tag is immutable, so the version number is already spent. Inline would have moved v1.7.5's red *later*, not earlier — same burned number, plus the cost of discovering it in the most expensive place.

The whole value of the preflight is that it is cheap and **before** the tag. Requiring its evidence is what actually puts it back in front of the cut, because the only way to hold a green preflight for commit `X` is to have run it on `X` — which is necessarily pre-tag. The gate does not merely detect the omission; it makes the omission impossible to proceed through.

Secondary reasons, honestly weighed:

- **Cost.** Two API calls (~2s) versus ~40s of duplicated work on every release. Small, but it is paid every cut, forever.
- **Drift.** One definition of "releasable" instead of two copies that diverge. This repo has already been bitten by exactly that: the ref-awareness fix below lived on `release/1.7.x` for a week and never reached `main`.
- **Against inline**, the strongest point is staleness — and the gate closes it: it pins on `github.sha`, so a green run for an older commit is not accepted, and "green then main moved" is a *block*, not a pass.
- The hazards you'd expect the API route to fumble — *ran on a different ref*, *ran but still `in_progress`*, *ran green then main moved* — are all handled explicitly, and each has a test that fails if the handling is removed.

**`in_progress` does not pass.** It blocks, naming the run it is waiting on and linking it. That leg is the single most important one here, because it is precisely what happened at 21:22Z on 2026-08-20.

## Break glass: where the line sits

The freeze policy calls for a documented hotfix path — *"skip schedule, never integrity."* This gate is on the **schedule** side of that line: it predicts whether the chain will stall; it does not certify the bytes. The three gates that *do* certify bytes — `resolve-candidate-digest`, `release-gate`, `verify-images-published` — are untouched and stay required, and the override is scoped to this one job.

The override is a trailer in the **annotated tag's own message**:

```
git tag -a v1.8.2 -m "Release 1.8.2

Preflight-Override: cut off-branch at 635496d0, which has no branch to
dispatch the workflow on; preflight run locally against that exact tree,
transcript on #3538"
```

Why there and nowhere else:

- `release.yml` has **no `workflow_dispatch`** — it was deliberately removed so the version released cannot drift from the ref tested. There is no input to hang a bypass on, and re-adding one to carry a bypass would be a strange thing to reintroduce.
- A repository variable or secret is **mutable and not scoped to a release**: set once, it silently disarms the gate for every future tag, and the next operator cannot see it in anything they read.
- The tag object is **immutable** (ruleset `19144026` restricts updates and deletions on `refs/tags/v*`), is created by the same command that starts the release, and is scoped to exactly one version. An override cannot be added after the fact, cannot leak to the next release, and is readable forever with `git show v1.8.2`.

A reason is **required** — at least 20 characters. A trailer with an empty or token reason is **refused, not honoured**; a lazy bypass must not be easier than the fix. Wrapped reasons are folded rather than truncated (`git tag -a -m` hard-wraps what a human types, and half a sentence in the audit trail is not a stated reason). Every honoured override prints the verdict it overrode, raises a `::warning`, and is written to the run summary. An override **cannot** convert an `INFRA` result into a pass: there is no measurement to accept the risk of.

## Prerequisite fixed here too: the preflight audited the wrong tree on `main`

`release-preflight.yml` hardcoded `ref: main` in its checkout, so dispatching it from `release/1.7.x` audited **main's** tree — v1.7.5, v1.7.6 and v1.7.7 each got a `READY` verdict that was a statement about a different branch (#3464). That was corrected on `release/1.7.x` for v1.7.8 and **never forward-ported**; `main` still had `ref: main` before this PR. Verified:

```
$ git show origin/main:.github/workflows/release-preflight.yml | grep -A4 Checkout
      - name: Checkout main
        uses: actions/checkout@3d3c42e5... # v7.0.1
        with:
          ref: main
```

The workflow now audits the ref it was dispatched on (with an optional `ref` input), and the verdict line names the ref and sha it evaluated. The `release/1.7.x` change is ported verbatim, including its four self-test cases, so the branches converge rather than diverge further.

## Falsifiability

A gate that has only ever been watched to pass is the defect this audit found repeatedly, so every leg was exercised for real.

### Against the live Actions API, on production data

| leg | fixture | result |
|---|---|---|
| no preflight run | `main@1475ee3c` | `BLOCKED: No Release Preflight has ever run against this commit.` exit 1 |
| preflight **failed** | `88153dc8` (run `32377238021`) | `BLOCKED: The newest Release Preflight for this commit concluded 'failure'.` exit 1 |
| **superseded green** | `88153dc8` has a green run on 08-19 *and* a red run on 08-20 | picks the red, exit 1 |
| green but **no evidence** | `3364edd7` (run `32489518200`, green, predates this PR) | `BLOCKED: ... is green but carries no evidence that it audited 3364edd7...` exit 1 |

### On a real runner, running `release.yml`'s job body verbatim

A throwaway `push`-triggered workflow ran the exact `preflight-evidence` job body (sparse checkout, `permissions: actions: read`, same env) — run `32683011388`, all three jobs `success`:

```
success  Require a green Release Preflight for this commit (override fixture)
success  Require a green Release Preflight for this commit (no-run fixture)
success  Require a green Release Preflight for this commit (green fixture)
```

- **green** — a real harness run produced a real `release-preflight-<sha>` artifact; the gate resolved it: `READY: Green Release Preflight run 32682792581 audited 0141cbe8...`
- **in_progress** — queried while a run was genuinely mid-flight: `BLOCKED: A Release Preflight for this commit is still 'in_progress' -- its verdict does not exist yet.` / `waiting on run 32682797643`
- **failed** — a red harness run: exit 1
- **no run** — inverted assertion (the job asserts the gate exits **nonzero**), so a fail-open regression fails the job
- **override** — read out of a **real annotated tag object** over the API, full reason folded and echoed into the `::warning`

The throwaway workflow, its three fixture branches and the proof tag have all been deleted; nothing from the harness is in this PR.

That harness earned its keep immediately: the first green run produced **no artifact at all**, because `actions/upload-artifact` defaults `include-hidden-files: false` and the evidence path was `.preflight-evidence/`. The gate refused the commit — correct outcome, wrong reason. Fixed in `76200585`; a comment records why the path is not a dot-directory.

### Revert-proof

Each production branch reverted in turn; the self-test goes red every time.

```
REVERT: in-flight preflight treated as absent            -> 2 case(s) failed
REVERT: green run needs no evidence artifact             -> 2 case(s) failed
REVERT: any green counts, not the newest completed       -> 1 case(s) failed
REVERT: override needs no reason                         -> 3 case(s) failed
REVERT: unreadable artifact listing is not INFRA         -> 1 case(s) failed
REVERT: override can launder an INFRA result into a pass -> 1 case(s) failed
REVERT: runs ordered by API response, not created_at     -> 2 case(s) failed
```

One of those was found *by* this exercise: the INFRA guard was initially dead code (every INFRA path `exit 2`-ed before the override was ever parsed), so the test passed with the guard deleted. `INFRA` is now a verdict that flows through the same reporting path, which makes the guard live, testable and visible in the run summary.

## What this does NOT do

- It does not make the preflight's own checks stronger — only required.
- It cannot make the tag-creation moment safe on its own. It fires after the tag is pushed, like every other gate in `release.yml`; its effect is on operator behaviour (you cannot hold green evidence for a commit without having preflighted it) and on refusing to build/publish over a known-red verdict.
- It says nothing about `release/1.7.x`. **That branch runs its own copy of `release.yml`, so this change does not protect it** — and three of the last six releases were cut from it. Porting is a separate decision that belongs with #3464 (which also covers the branch's inert `docker-publish` triggers and its pre-#3422 release-branch gate); this PR is `main` only, as scoped.
- The first cut after this merges needs one fresh Release Preflight run: green runs from before this PR carry no evidence artifact and are correctly refused (demonstrated above against `3364edd7`).

- [x] N/A — this is not a bug fix (`ci/*`), though it does ship a self-test per leg

## Test Checklist
- [x] Unit tests added/updated — `scripts/ci/test-assert-preflight-evidence.sh`, 24 cases, wired into `shell-tests`; plus the 4 ported ref-awareness cases in `test-release-preflight.sh`
- [x] Manually tested locally — and against the live Actions API and a real runner, see Falsifiability
- [x] No regressions in existing tests — `shellcheck -x` clean on all four scripts; no Rust touched

## API Changes
- [x] N/A - no API changes


brandonrc:
### Rebased onto post-#3539 `main`; integration collision resolved

#3539 and this PR both touch `scripts/ci/release-preflight.sh` and `.github/workflows/release-preflight.yml`. Rebased, and the resulting 4 self-test failures are fixed in `d9c9b2bf`.

**The checkout block.** The two changes were on independent axes and both are needed, so the merged block takes both: `ref: ${{ inputs.ref || github.ref }}` (*which* commit — this PR) with `fetch-depth: 0` (*how much history* — #3539's check 5 needs `git describe` and `git log <tag>..HEAD`). Verified that neither side's tests assumed the other's value: both suites build synthetic repos and drive the script through `PREFLIGHT_*` env, and neither reads the workflow file. Check 4 is also fine at depth 0 — its commit fetch is already guarded by `git cat-file -e`, so it simply stops fetching.

**Why the 4 failures, and what I did.** All were `expected exit 0|1, got 2`: the ref-awareness fixtures have no `CHANGELOG.md`, so check 5 returned INFRA and masked the verdict each case asserts. Two fixes, for two different problems the one symptom was hiding:

1. **Scope the cases to the check they test** (`PREFLIGHT_SKIP_CHANGELOG_RECON=1`). This is the convention every other helper in the file already follows — check 5's own cases neutralise checks 3 and 4 symmetrically. The alternative (give the fixture a `CHANGELOG.md`, a reachable stable tag and a `gh` stub answering check 5's GraphQL PR→issue query) would couple a check-1 assertion to check 5's fixture requirements — which is precisely the coupling that made these two changes collide. A case should be a statement about one check. A comment now records that, so the next person does not "fix" it by bolting on a CHANGELOG fixture.

2. **Fix the precedence defect the collision exposed.** Every INFRA site exits immediately, discarding whatever `problems` the earlier checks had counted. So a run where check 1 found real forward-port drift **and** a later check could not be measured reported **exit 2 — "retryable"** — when the true answer was already "you are not ready, and retrying will not help". That is not hypothetical on a real repo: check 3's and check 4's INFRA conditions are transient network states.

   `infra_exit` now reports the strongest claim the run has actually earned: **exit 1 if a blocking problem has been counted, exit 2 otherwise**, with the transcript saying which checks did not run so an incomplete audit is not mistaken for a clean sweep. Checks only ever *add* problems, so a verdict reached this way cannot be withdrawn by a check that did not run.

   **This is a behaviour change to #3539's exit-code contract and is called out as such** in `RELEASING.md`, the script header and the CHANGELOG. It does not weaken INFRA in the direction anyone cares about: an indeterminate result still never reads as a pass, nothing that used to block stops blocking, and the only conversions are exit-2 → exit-1. It is pinned from **both** sides — forcing `infra_exit` to always exit 2 fails the 2 new cases; forcing it to always exit 1 fails **5 pre-existing INFRA cases, including #3539's own two**. Their tests are the guardrail against over-promotion, which is the right shape.

**Also confirmed, as asked.** `gh` absent degrades to exit 2, never 0, on both paths — evidence gate: `INFRA: gh is not on PATH -- cannot look for preflight evidence`, exit 2; preflight check 5: `INFRA: gh is not on PATH; the issue<->PR map cannot be built`, exit 2 (both exercised with `gh` genuinely off `PATH`). The `preflight-evidence` job deliberately keeps `sparse-checkout` at depth 1 and must **not** be "fixed" to `fetch-depth: 0` by analogy: it reads the Actions API, and its only git call (deriving the repo slug from `origin`) is bypassed because `GATE_REPO` is passed explicitly. A comment now says so.

All 9 shell self-tests and all 5 standalone gates pass locally; `shellcheck -x` clean.

---

### CI status: one unrelated red

`✅ CI Complete` is red **solely** because of `🦀 Check Rust`. Its rollup env shows every other result green:

```
RESULT_CHECK_RUST: failure
RESULT_UNIT: success        RESULT_SHELL: success
RESULT_COVERAGE: success    RESULT_INTEGRATION: success
RESULT_BUILD_BACKEND: success   RESULT_SECURITY: success
```

That failure is **#3551**, a pre-existing repo-wide breakage: Rust 1.98's clippy `result_large_err` fires **555 errors** across existing handlers. Same error, same count, on `main` itself at `000484b9` — a commit that predates this branch's base — and PRs #3544 and #3536 are red identically. **This PR changes zero Rust files** (`git diff --name-only origin/main...HEAD | grep -E '\.rs$|Cargo\.(toml|lock)$'` → empty), so nothing here can influence it, and fixing 555 clippy errors inside a CI-gating PR would be exactly the kind of unrelated change that ought not to ride along.

This PR cannot go green until #3551 lands on `main`. Not marking ready; no merge.

Issue #3546, opened 2026-08-24
Silent SSO auto-login (OIDC prompt=none check-sso) with fail-open probe and operator kill switch


brandonrc:
## Problem

When an OIDC provider (e.g. Keycloak) is configured, a user who already holds a live IdP session still has to click **Sign in with Keycloak** every time they open the web UI. Platforms that embed Artifact Keeper next to other SSO-protected apps expect to land already signed in.

A naive fix — auto-redirecting to the sole provider on load — is wrong for a registry: anonymous browsing is a first-class state (public repos, pypi/npm proxy indexes), and bouncing every anonymous visitor to a visible Keycloak login page uninvited breaks that.

## Proposal (check-sso, the Keycloak JS-adapter semantic)

Backend support for an invisible probe the web UI runs once per browser session:

- `GET /api/v1/auth/sso/oidc/{id}/login` accepts an optional `prompt=none` query parameter, forwarded to the IdP authorize URL. Only `none` is accepted; the explicit sign-in flow (no `prompt`) is unchanged byte-for-byte.
- The OIDC callbacks map the OIDC Core 3.1.2.6 interaction-required error family (`login_required`, `interaction_required`, `consent_required`, `account_selection_required`) — the IdP's "no live session" answer to a `prompt=none` probe — to a clean 307 to `/callback?silent_denied=1` with **no error payload**, burning the single-use SSO session. The web UI treats it as "remain anonymous".
- A live IdP session completes the normal code flow invisibly and the user is signed in with no click.

## Guardrails (the anticipated objections)

- **Anonymous stays anonymous**: no visible IdP page is ever shown uninvited; a denied probe produces no error UI.
- **IdP down ⇒ fail open**: the frontend runs the probe in a hidden, timeout-guarded iframe, never a top-level redirect, so an unreachable IdP cannot strand a visitor on a browser error page (frontend: artifact-keeper-web).
- **Kill switch**: `OIDC_SILENT_SSO=false` disables the feature fleet-wide; advertised to the frontend as `auth.silent_sso_enabled` in `GET /api/v1/system/config`. Default on.

Frontend counterpart tracked in artifact-keeper-web.

Issue #2686, opened 2026-07-17
In-place upgrade v1.5.7 → 1.6.0 fails to boot: _sqlx_migrations VersionMismatch(154) from release/1.5.x ledger divergence


brandonrc:
## Summary
An in-place upgrade of a **v1.5.7** deployment to a **1.6.0** (main) build **fails to start**:

```
Error: Migration(VersionMismatch(154))
```

Found by the Deployment Test Framework `upgrade` tier (boots the v1.5.7 image, seeds data, swaps to the 1.6.0 image on the same volume). CI never exercises prior-release→current in-place upgrade, so this passes every green gate.

## Root cause — divergent migration ledgers
The #2504 storage-key-index migration was renumbered on the hotfix branch, so the same logical migration sits at **different version numbers** on the two lines:

| version | release/1.5.x (applied on customer DBs) | main (1.6.0) |
|---|---|---|
| 154 | `artifacts_storage_key_index` (#2504) | `webhook_deliveries_claim` |
| 155 | — | `projects` |
| 156 | — | `upstream_feed_state` |
| 157 | — | `artifacts_storage_key_index` (#2504) |

A v1.5.7 DB has version 154 applied with the *storage_key_index* checksum; main's 154 file is *webhook_deliveries_claim* → sqlx checksum mismatch → `VersionMismatch(154)` → backend refuses to boot.

`release/1.5.x:154_artifacts_storage_key_index.sql` is **byte-identical** to `main:157_artifacts_storage_key_index.sql` (same checksum).

## Fix
Add `repair_release_1_5_x_divergence()` to `backend/src/migration_repair.rs`, wired into the pre-migrate repair sequence in `main.rs` (alongside the existing `repair_release_1_1_9_divergence`, which solved the identical class for the release/1.1.x 73–75 divergence):

1. Skip on fresh install (no `_sqlx_migrations`) or when version 154 is already `webhook_deliveries_claim` (already a 1.6.0 DB).
2. Detect the v1.5.7 signature: version 154 applied with the storage_key_index checksum and main's 155/156/157 not yet applied.
3. `UPDATE _sqlx_migrations SET version = 157 WHERE version = 154` (checksum already matches main-157).
4. The migrator then applies 154/155/156 fresh and skips 157 → boots.

Idempotent, transaction-guarded, no-op on fresh/1.6.0 DBs.

## Verification
[...]

Issue #2430, opened 2026-07-12
Token exchange launders API-token action scope → read-only token gains write/delete (CRITICAL)


brandonrc:
**Severity: CRITICAL. Live in shipped v1.5.1 and earlier.** Confirmed by red-team on the rig.

A scope-restricted API token can be exchanged for a JWT session that drops the action-scope ceiling, so a **read-only** token gains full **write/delete** across every format and every repo the principal can read.

**Root cause:** `Claims` (auth_service.rs) has no action-`scopes` field; `From<Claims>` hardcodes `scopes: None` (middleware/auth.rs:186); `has_scope()` short-circuits `if !self.is_api_token { return true }` (auth.rs:97). Every JWT minted from an API-token credential therefore loses the read/write/delete scope and can never re-enforce it.

**Vectors (confirmed):**
1. `POST /conan/{repo}/v2/users/authenticate` (conan.rs:362 `generate_tokens`) — launders **both** action scope and repo restriction → a read-only, repo-scoped token becomes a fully unrestricted `Bearer` JWT.
2. `GET /v2/token` (docker login, oci_v2.rs) — action-scope laundering (read→write); #2290 already threads `allowed_repo_ids` so repo restriction holds, but the minted JWT has no `scopes` claim, and `authenticate_oci_with_scopes` returns `(claims, None)` when re-presented on `docker push`.

**Audited clear:** npm/nuget/pypi/gitlfs/gem/etc. do not mint JWTs (they consume AuthExtension directly + already scope-check). Interactive password/TOTP login and federated CI-OIDC are full-privilege by design and unaffected.

**Fix:** add `Claims.scopes: Option<Vec<String>>` (None = interactive/full, Some = ceiling copied from the presenting token); make `has_scope`/`From<Claims>` honor it; thread scopes (+ allowed_repo_ids) through every exchange mint (conan, /v2/token, OCI re-presentation) and refresh. No DB migration. Pairs with the GHSA-vvc3 residual handler hardening (#2417). Ships in **v1.5.2**.

Issue #1819, opened 2026-06-09
Security: auth(2FA): TOTP-verify session skips refresh-jti persistence — replay-detection disabled for 2FA users


brandonrc:
**Severity:** high — found by internal red-team (round 3, source-verified).

## Summary
Refresh-token rotation / replay-detection (RFC 6819) is silently disabled for the TOTP-verify auth path — stolen refresh tokens of 2FA users are infinitely replayable and the token family is never revoked

## Reproduction / evidence
Principals (DB-verified): rsec1 is_admin=f totp_enabled=t; rsec2 is_admin=f totp_enabled=f. Both reset to password "Verify!2026xZ" (password_changed_at backdated 2 days to avoid the credential-change invalidation gate). rsec1 TOTP secret recovered from DB (GHGRCNJFEKZDFCFLUFL2DU5XU7CULNMH, SHA1/6/30).

TOTP user rsec1 (the vulnerable path):
1) POST /api/v1/auth/login {rsec1} -> 200, totp_required:true, totp_token.
2) POST /api/v1/auth/totp/verify {totp_token, code} -> 200; refresh_token RT0 carries jti + family_id; access token claims show username=rsec1, is_admin=false, token_type=access (a full non-admin session, NOT the limited pending token).
3) DB check: select count(*) from refresh_token_jti where jti='<RT0.jti>'  => 0 (NO ROW — never persisted).
4) POST /api/v1/auth/refresh {RT0} -> 200 (rotate#1 -> RT1); POST /api/v1/auth/refresh {RT1} -> 200 (rotate#2 -> RT2).
5) REPLAY the already-rotated original RT0: -> 200, 200, 200 (and a follow-up loop returned 200 x7). The original jti still has 0 rows in refresh_token_jti, so it is never marked consumed; the family is never revoked and the latest sibling RT2 still refreshes (200).

CONTROL — non-TOTP user rsec2 (same non-admin privilege, password path):
- RT0.jti row IS persisted (row present). rotate#1 -> 200, rotate#2 -> 200. REPLAY rotated RT0 -> 401 ("Refresh token replay detected"); the family is killed so even the latest CRT2 -> 401. Correct RFC 6819 behavior.

The sole difference between vulnerable and safe is whether the session was minted via /auth/totp/verify vs password /auth/login. Result: every 2FA-enabled account gets multi-use, replayable refresh tokens with no breach detection.

## Control
[...]

PR #3549, opened 2026-08-24
feat(auth): implement OIDC Device Authorization Grant (RFC 8628)

### LINKED ISSUE #3461: OIDC Device Authorization API (RFC 8628) for Artifact Keeper
## Feature Request: OIDC Device Authorization API for Artifact Keeper

Artifact Keeper should expose a standards-compliant OAuth 2.0 Device Authorization Grant API, using OpenID Connect to authenticate the end user and then minting an Artifact Keeper-native access token.

The implementation should follow RFC 8628, RFC 9700, and OpenID Connect Core.

The API must be compatible with standard OAuth clients, including oauth2c.

### Suggested Endpoints

```
POST /api/v1/auth/oidc/device/code
POST /api/v1/auth/oidc/device/token
```

### Device Authorization Response

The device authorization response should contain the standard RFC 8628 fields:

```json
{
  "device_code": "...",
  "user_code": "ABCD-EFGH",
  "verification_uri": "https://ak.example.com/device",
  "verification_uri_complete": "https://ak.example.com/device?user_code=ABCD-EFGH",
  "expires_in": 600,
  "interval": 5
}
```

### Token Endpoint

The token endpoint must accept the standard request:

```
POST /api/v1/auth/oidc/device/token
Content-Type: application/x-www-form-urlencoded
grant_type=urn:ietf:params:oauth:grant-type:device_code
device_code=<device-code>
client_id=<client-id>
```

It must return standard device-flow errors, including:
- `authorization_pending`
- `slow_down`
- `expired_token`
- `access_denied`

### Example Usage with oauth2c

```bash
oauth2c https://ak.example.com \
  --client-id artifact-keeper-cli \
  --grant-type urn:ietf:params:oauth:grant-type:device_code \
  --device-authorization-endpoint \
    https://ak.example.com/api/v1/auth/oidc/device/code \
  --token-endpoint \
    https://ak.example.com/api/v1/auth/oidc/device/token \
  --scopes openid,read:artifacts
```

The exact command should be validated against the supported oauth2c version. The API must treat the CLI as a public client and must not require a client secret embedded in the CLI.

## Security Requirements

- Accept only configured OIDC providers
- Do not accept arbitrary issuer URLs, usernames, email addresses, sub claims, or user-supplied OIDC claims as proof of identity
- Validate issuer, signature, JWKS key rotation, audience, authorized party where applicable, expiration, and provider/client binding
- Bind device transactions to the provider, client, requested scopes, and transaction state
- Generate high-entropy, short-lived, single-use device and user codes
- Rate-limit device-code creation, user-code attempts, and token polling
- Prevent user-code enumeration and brute-force attacks
- Require explicit user approval and display requested scopes
- Enforce least-privilege, allow-listed scopes server-side
- Do not grant wildcard or administrative scopes by default
- Issue short-lived AK access tokens and enforce a maximum refresh-token lifetime
- Do not allow "never expires" tokens through this flow
- Rotate refresh tokens and detect refresh-token reuse
- Support revocation and account/provider disablement
- Do not log or expose device codes, tokens, or OIDC assertions unnecessarily
- Use TLS and publish OAuth/OIDC metadata where possible
- Record audit events for device-code creation, approval, token issuance, refresh, revocation, expiration, and failed attempts

### Token Minting

The API must mint an AK-native token after successful OIDC validation. An upstream OIDC access token must not be accepted directly as an Artifact Keeper API token.

## Acceptance Criteria

- oauth2c can complete the device flow using the standard grant type and endpoint parameters
- A configured OIDC user receives a valid AK access token after approving the request
- Unknown providers, invalid issuers, invalid signatures, wrong audiences, expired assertions, reused device codes, and unauthorized scopes are rejected
- Polling abuse returns slow_down and is rate-limited according to RFC 8628
- "Never" expiration is unavailable for tokens issued through this flow
- Existing browser OIDC, CI OIDC JWT exchange, API-token, and password-login flows remain unaffected

## Integration Testing Coverage

Integration tests must cover:
- Success path: complete device flow
- Pending authorization
- Slow-down handling and rate limiting
- User denial
- Token expiration
- Replay attacks
- Scope escalation
- Provider mismatch
- Refresh token rotation
- Token revocation
- Audit events
- oauth2c interoperability

vivekdsr:
## Summary

Implements OIDC Device Authorization Grant (RFC 8628) as requested in #3461.

This adds a new device flow for CLI tools, IoT devices, and CI systems that cannot open a browser — allowing them to authenticate against Artifact Keeper's OIDC provider without direct user-agent interaction.

### New files

- **`backend/migrations/209_oidc_device_sessions.sql`** — DB table `oidc_device_sessions` with:
  - `device_code` (64-char hex), `user_code` (`XXXX-XXXX` confusion-free alphabet)
  - `status`: `pending` → `approved` → `consumed` (or `denied`/`expired`)
  - Poll rate-limiting columns (`poll_count`, `last_polled_at`, `slow_down_until`)
  - Expiry at 10 minutes, interval 5 seconds
  - FK to `oidc_configs` and `users`

- **`backend/src/services/oidc_device_service.rs`** — `OidcDeviceService`:
  - `create_session` — creates device + user codes, persists to DB
  - `poll_token` — RFC 8628 §3.4 compliant polling with `authorization_pending` / `slow_down` / `expired_token` / `access_denied`
  - `approve_session` — authenticated endpoint for users to approve a pending device code
  - `deny_session` — authenticated endpoint to deny
  - `cleanup_expired` — housekeeping

- **`backend/src/api/handlers/oidc_device.rs`** — 4 HTTP handlers:
  - `POST /api/v1/auth/oidc/device/code` — RFC 8628 §3.1 device authorization request (per-IP rate-limited: 10/min)
  - `POST /api/v1/auth/oidc/device/token` — token polling with RFC 8628 error format (not AK standard envelope)
  - `POST /api/v1/auth/oidc/device/approve` — authenticated approval (user submits `user_code`)
  - `GET /device` — browser placeholder page for the verification URI

### Modified files

- **`backend/src/services/mod.rs`** — `pub mod oidc_device_service`
- **`backend/src/api/handlers/mod.rs`** — `pub mod oidc_device`
- **`backend/src/api/routes.rs`** — nested public router + authenticated approve route + /device page route

## RFC 8628 Compliance

- §3.1: device authorization request → `device_code`, `user_code`, `verification_uri`, `verification_uri_complete`, `expires_in`, `interval`
- §3.4: token polling → proper `authorization_pending`, `slow_down`, `expired_token`, `access_denied` error responses with HTTP 400 (not standard AK error envelope per RFC requirement)
- §3.5: rate limiting via `slow_down` with 5-second interval increase
- Confusion-free user code alphabet (no `I`, `O`, `0`, `1`)

## Security

- Per-IP rate limit: 10 device code creations per minute (in-memory, via `OnceLock<Mutex<HashMap<IpAddr, ...>>>`)
- Scope allow-list: `openid`, `profile`, `email`, `read:artifacts`, `write:artifacts`
- Device codes expire after 600 seconds
- Audit event fired on successful token issuance
- Approval requires authenticated session — unauthenticated users cannot approve device codes

## Verification

`cargo check` passes with **0 errors, 0 warnings**.

Closes #3461

## Design note: why `GET /device` is served by the backend

The verification page intentionally lives in the backend (the repo's only human-facing
page — existing backend HTML is protocol-mandated, e.g. the PEP 503 simple index):

1. **Self-contained PR**: the web frontend ships from its own repo on its own cadence;
   a cross-repo version-coupled change would create a window where a new backend +
   old web = broken `verification_uri`.
2. **Headless deployments**: `web.enabled=false` (edge) must still be able to complete
   the device flow — `verification_uri` always resolves.

The page is deliberately isolated in two handlers (`device_page`, `device_page_script`)
with zero coupling to the flow logic, and drives only public/authenticated JSON
endpoints (`/auth/me`, `/device/approve`) — so a first-class page in
`artifact-keeper-web` can replace it later with the backend page kept as the headless
fallback. Happy to file that follow-up on the web repo if maintainers prefer that split.


vivekdsr:
CI status note:

- ✅ All jobs green except **🦀 Check Rust**: it fails with 555 pre-existing `clippy::result_large_err` errors (Rust 1.98 toolchain drift on the CI runners) — **zero** in this PR's files, and `main` is red with the identical failure. Tracked in #3551.
- 📊 Code Coverage gate now passes after adding DB-backed tests for the device-flow service and handlers (commit 2370d6b5).

vivekdsr:
@brandonrc Could you please help in reviewing this PR if this is a good solution ?

brandonrc:
⚠️ **Migration number collision with another open PR — this will not show up as a conflict.**

Two open PRs currently claim migration **209**:

- **#3578** — `backend/migrations/209_storage_metrics_proxy_columns.sql` (plus `210_storage_metrics_proxy_backfill.sql`)
- **#3549** — `backend/migrations/209_oidc_device_sessions.sql`

`main` currently tops out at `208_backups_payload_checksum.sql`.

Because the filenames differ, git sees no conflict. Both PRs will show `mergeable: CLEAN`, both will pass CI independently, and both will merge without complaint. The failure appears only **after the second one lands**: sqlx records one version `209` in `_sqlx_migrations`, encounters a second file claiming the same version, and the backend **aborts at startup**. On `main`.

Whoever merges second needs to renumber before merging. Given #3578 also uses 210, the clean split is:

- #3578 keeps 209 + 210, or moves to 210 + 211
- #3549 moves to whichever remains free

Note **#3599** already claims **211**, so please avoid it — checked and confirmed against the current ledger.

Worth doing the check by hand rather than trusting CI here: `gh pr view <n> --json files -q '.files[].path' | grep migrations` across the open PRs is the whole check, and it is the only thing that catches this before it reaches `main`.

brandonrc:
Update: the 1.8.2 blockers have now taken their numbers, so the free range has moved.

Currently claimed against `main` (which tops at `208_backups_payload_checksum.sql`):

- **209** — #3604 `209_scan_results_pin_identity.sql` (1.8.2 blocker, merging first)
- **211** — #3599 `211_packages_version_oci_tag_length.sql` (1.8.2 blocker)
- **209** — #3578 `209_storage_metrics_proxy_columns.sql` + **210** `210_storage_metrics_proxy_backfill.sql`
- **209** — #3549 `209_oidc_device_sessions.sql`

So three PRs hold 209. The 1.8.2 pair merges first, which means **#3578 and #3549 both need to renumber before they land**.

Free after 1.8.2: **210**, and **212** onward.

Suggested split — #3578 takes 210 + 212 (it needs two), #3549 takes 213. A gap at 212 or elsewhere is harmless; the ledger already has several.

Restating why this is worth a manual check: the filenames differ, so git reports no conflict, both PRs show `mergeable: CLEAN`, and CI passes on each independently. The failure only appears after the second one merges, when sqlx finds two files claiming one version and the backend aborts at startup — on `main`.

vivekdsr:
Done — renumbered to **213** as suggested (commit 03436bdd, after merging latest `main`).

- `209_oidc_device_sessions.sql` → `213_oidc_device_sessions.sql` (rename only, content unchanged; no code references the number)
- Leaves **210 + 212** free for #3578's pair per the proposed split
- Verified on a fresh database: the ledger now applies `208 → 209 (scan_results_pin_identity) → 211 → 213` cleanly, `scripts/ci/check-migration-ledger.sh` passes, and the device-flow tests are green.

Thanks for catching the silent-collision window — agreed that `mergeable: CLEAN` + independent CI would have masked this until it hit `main`.

brandonrc:
## Audit round 1 — OIDC Device Authorization Grant (3 reviewers: security / correctness+protocol / integration+architecture)

Thank you for taking this on — #3461 is a real need and the shape is right: AK acting as the RFC 8628 *authorization server* and minting AK-native tokens (the architecture the issue asked for, not a relay to the IdP's device endpoint), reusing `RequestBaseUrl` and `AuthService` as the mint choke point, a correct migration slot (213 — 210/212 stay free for #3578), the token endpoint's wire shape largely per RFC 6749 §5.2, and the ten inline tests run in CI's unit job with a real database (all three reviewers ran them: 10/10). The merge with `origin/main` is textually clean; the *only* compile break is `use rand::RngCore;` at `oidc_device_service.rs:4` → `use rand::Rng;` after main's rand 0.10 bump (#2852).

But this is a token-minting surface reachable unauthenticated, and the security envelope #3461 spends most of its body specifying is not there yet. **Verdict: do not merge in this state.** Every blocker below was demonstrated against a live database with a throwaway probe test, by two reviewers independently, not inferred.

### Blockers

1. **An approved `device_code` is never consumed.** `poll_token` (`oidc_device_service.rs:145-165`) returns `Approved` on every call while `status='approved'`; there is no `consumed` transition anywhere in the diff or in migration 213 (`grep -c consumed` = 0), although the PR body describes `pending → approved → consumed`. Measured: polling one approved code three times returned three distinct token pairs, status still `"approved"`. At the 5 s interval that is ~120 independently-refreshable token pairs per approval. The repo's own pattern is `validate_sso_session` (`auth_config_service.rs:1775`) — atomic `DELETE … RETURNING`.

2. **The minted token ignores the scopes it just validated and carries no ceiling.** `create_device_code` validates `read:artifacts`/`write:artifacts` and `create_session` persists them; issuance then calls `generate_tokens(&user)` (`oidc_device.rs:512`), which is `…with_family_and_scope(user, …, None, None)` — `None` being the *unrestricted* sentinel. Measured: a session requesting `["openid","read:artifacts"]` minted a JWT with **no `scopes`, no `allowed_repo_ids`, `is_admin: true`**. The repo already has `generate_tokens_with_scope`, used by `conan.rs:366-371` and the OCI `/v2/token` exchange, whose doc comment (`auth_service.rs:1315-1322`) states exactly the invariant this breaks (#2430).

3. **Approval is unbound to credential, provider, or scope.** `approve_session_handler` (`oidc_device.rs:555-570`) passes `auth.user_id` straight through — it never reads `auth.is_api_token` / `auth.scopes` / `auth.is_service_account` — and `approve_session` matches on `user_code` alone: no `provider_id` predicate, `oidc_sub`/`email`/`name` hardcoded to `""` ("oidc_sub not available in this path"). Since `/approve` sits under `auth_middleware`, which accepts API tokens, **a `read:artifacts`-only API token can approve its own device session and collect a full unscoped session JWT plus refresh family** — the escalation `scripts/ci/check-token-mint-surface.sh` exists to prevent, and it does not cover this endpoint because it keys on `generate_api_token`. A session bound to a Keycloak provider is approvable by a local-password user.

4. **The refresh token is invisible to replay detection and revocation.** `:512`/`:543` return `tokens.refresh_token` without `persist_refresh_jti_from_pair`; every replay of a leaked device-flow refresh token hits the `auth_service.rs:1773-1788` "row missing → rotate normally" branch and mints a fresh pair, and `logout` has no family row to revoke. This is a verbatim re-occurrence of #1819 — `totp.rs:604-612` carries the fix with a comment naming the consequence.

5. **No throttle on `user_code` guessing, and the audit trail is empty.** Measured: 500 wrong `user_code` guesses → 500 plain `NotFound`, **0 audit rows**, then the correct code approved a different client's session and bound it to the guesser. The per-IP device-code limiter keys on the raw socket peer, so behind a reverse proxy it becomes a fleet-wide 10/min DoS. Of the seven auth events the repo's audit gate expects for a surface like this, one is emitted.

### Should-fix
- The approval page shows no `client_id`, no scopes, no expiry, and has no Deny — `deny_session` has no handler or route, so the user-denial path the issue requires is unreachable, and `access_denied` can never be returned.
- The `sso.rs` `?return_to=` change composes with 1.8.2's `prompt=none` silent SSO (#3546) into a one-click chain onto a pre-filled consent page; it has zero tests, including for `sanitize_return_to`.
- `/device` is missing from the guest-access allowlist, so the verification URI 401s on `AK_GUEST_ACCESS_ENABLED=false` — defeating the headless rationale.
- No per-provider enable flag, no `system/config` advertisement, no OpenAPI/utoipa annotations (so no SDK, and no linked web PR for the verification page), `cleanup_expired` has no scheduler caller (unbounded table), the device-authorization endpoint returns AK's error envelope instead of the RFC 6749 §5.2 shape the token endpoint already gets right.

### Suggested order
Rebase (one-line rand fix) → consume on poll (atomic `UPDATE … WHERE status='approved' RETURNING`) → mint via `generate_tokens_with_scope` and persist the refresh JTI → bind approval (reject API-token / service-account principals, check `provider_id`, record identity) → throttle + audit → Deny route + page context. The CHANGELOG `### Added` bullet should describe the *fixed* feature; I'll draft it once the blockers are in. Milestoning 1.9.0 so it's tracked, but it lands only when the above is closed.

brandonrc:
Heads-up on migration numbering: main now ends `209, 211, 212, 214` after today's merges (#3786 took 212, #3788 took 214). `213_oidc_device_sessions.sql` in this PR is still free and needs no change — flagging it only because #3578 does have to renumber, and I would rather you hear it from me than from a red `check-duplicate-migrations` on the merge commit.

brandonrc (changes requested):
## Security review — changes requested, do not merge yet

Thank you for this — RFC 8628 is a genuinely useful gap to close for CI and headless clients, the page is careful about CSP in a way most contributions aren't, and the `return_to` sanitizer is well built. I reviewed it as an authentication-bypass hunt rather than a code review, because it mints tokens.

**There is no authentication bypass** — every step requires a valid credential. What I did find is a **privilege escalation** that a scoped API token can walk end to end without a browser, an approved `device_code` that can be redeemed an unlimited number of times, and a one-click phishing path assembled from this PR's own parts. Details below, ranked by exploitability.

First, two corrections to my own starting assumptions, so nobody carries them forward:

- **There are tests** — 9 of them, in two commits (`oidc_device_service.rs:312-534`, `oidc_device.rs:574-1049`). The PR *body's* "Verification: `cargo check` passes" is stale and undersells the branch. Please update it.
- **Migration 213 is not a collision.** 213 is a genuine gap on `main` (210–212, 214–216 exist), free on every `release/*` branch, unclaimed by other open PRs, and the full set applies cleanly against a real Postgres 16 — I ran it. It should still be **217** per `CLAUDE.md:269`: filling a historical gap is how two in-flight PRs end up on the same slot, which is the ledger divergence #2686/#2689 exist to prevent. Also note the branch is stale — it has 213 but is missing 210, 212, 214, 215, 216 from current `main`, so the `cargo check` you ran was against an older base.

---

## 1. Blocker — the issued token ignores both the requested scopes and the approver's own privilege ceiling

`oidc_device.rs:508` calls `auth_service.generate_tokens(&user)`, which is `generate_tokens_with_family_and_scope(user, Uuid::new_v4(), None, None)` (`auth_service.rs:1301`) — **no repository allow-list, no action-scope ceiling.** The comment on its sibling says exactly why that matters (`auth_service.rs:1315-1322`):

> the minted JWT inherits both the token's repository allow-list AND its action-scope allowlist, so a read-only token can never be laundered into a write/delete-capable JWT.

Every other mint-from-a-credential path in the tree honours that — `conan.rs:367`, `oci_v2.rs:432`, `:475`, `:535`, `:4649`, `:4745`, `:4797`. The device flow is the only one that doesn't.

Three things fall out:

- **The scope allow-list is decorative.** `ALLOWED_SCOPES` (`oidc_device.rs:59-65`) is validated on input (`:333-341`), persisted to `oidc_device_sessions.scopes`, and **never read again**. A device asking for `scope=read:artifacts` gets a token that can write, delete and promote. Pass the requested scopes into `generate_tokens_with_scope` as the ceiling.
- **It's a second, divergent copy of the scope vocabulary.** The canonical list is `token_service.rs:91-103` with `validate_scopes_pure` at `:108`. This invents a five-entry list mixing OIDC scopes (`openid`, `profile`, `email`, which mean nothing to AK's authorizer) with two AK scopes, omitting the other nine. It will drift.
- **Admin is inherited silently.** Claims are built from the `users` row, so an admin approver yields an admin token, with no consent step naming what is being delegated.

## 2. Blocker — `/approve` accepts any authenticated principal and discards its constraints

`approve_session_handler` (`oidc_device.rs:546-564`) extracts `AuthExtension` and reads **exactly one field**, `auth.user_id`. `AuthExtension` (`middleware/auth.rs:68-92`) also carries `is_api_token`, `is_service_account`, `scopes` and `allowed_repo_ids` — all discarded.

Combined with finding 1, that is the escalation:

> A caller holding a `read:artifacts`-only API token (or a repo-scoped token, or a service account) can call `POST /device/code` (unauthenticated) → `POST /device/approve` with `Authorization: Bearer <read-only token>` → `POST /device/token`, and receive an **action-unrestricted, repo-unrestricted JWT plus a multi-day refresh token** for that token's owner. `violates_csrf_contract` (`auth.rs:641`, tested at `:4446`) deliberately exempts header-authenticated clients from `X-Requested-With`, so nothing stands in the way.

This is precisely the laundering #2430 closed on the Conan and OCI token endpoints, reopened on a new one. No victim, no social engineering, no race — which is why I rate it the most exploitable finding here.

At minimum: refuse approval from a non-interactive credential (`is_api_token` / `is_service_account`), and propagate `auth.scopes` and `auth.allowed_repo_ids` into the mint as a ceiling.

**Related:** the OIDC in "OIDC Device Grant" never actually happens — `oidc_sub`, `oidc_email` and `oidc_name` are passed as `String::new()`. The flow resolves a `provider_id`, checks `is_enabled`, and then authenticates the approver by whatever AK already accepts, including local password sessions and API tokens. The `oidc_configs` FK is decoration. Either drive a real OIDC authorization at approval time, or rename this to a plain AK device grant and drop the provider coupling.

## 3. Blocker — an approved `device_code` is redeemable without limit; there is no `consumed` state at all

`poll_token` (`oidc_device_service.rs:107-165`) reads the session, updates poll bookkeeping, and returns `Approved { user_id }`. **Nothing ever transitions the row out of `approved`.** The migration has no `consumed` status and no `consumed_at`; `grep -rn consumed` over the three new files returns nothing.

**Executed:** five polls produced five token pairs, with `status` still `approved`. The only throttle is `interval_secs = 5`, so one approval yields ~120 token pairs inside the 600-second window, each carrying a refresh token good for `JWT_REFRESH_TOKEN_EXPIRY_DAYS`.

The PR body states the opposite in its own words — "`status`: `pending` → `approved` → `consumed`" and "poll rate-limiting columns (`poll_count`, `last_polled_at`, `slow_down_until`)". Neither `consumed` nor `slow_down_until` exists anywhere in the diff. A reviewer reading the body ticks this box and moves on, which is the part that worries me most: **the single most important property of the flow is documented as implemented and is not.**

RFC 8628 §3.5 is admittedly silent on single-use, but RFC 6749 §5.2 defines `invalid_grant` for a grant that is "invalid, expired, revoked", every reference implementation treats the device code as one-shot, and *this codebase's own* web refresh tokens are single-use rotating (`auth_service.rs:1351-1356`). A device grant replayable 120 times while the web session it clones can't be replayed once is an internal inconsistency, not just a spec debate. In practice a `device_code` lives in a CLI's memory, its `--verbose` output, a CI log, or shell history — under one-shot semantics a leaked code the device already redeemed is dead; here it's a ten-minute key to unlimited fresh sessions, plus refresh tokens that outlive the window by days.

Make redemption one conditional statement:

```sql
UPDATE oidc_device_sessions
   SET status = 'consumed', consumed_at = now(), updated_at = now()
 WHERE device_code = $1 AND client_id = $2
   AND status = 'approved' AND expires_at > now()
RETURNING approved_user_id
```

`fetch_optional` then decides; a second caller gets zero rows and `invalid_grant`. Add `consumed` to the migration with a `CHECK`, and a test that polls twice and asserts the second fails.

(I probed the concurrency question the other way too — my two-connection double-redeem probe found only one winner, but that's incidental serialisation by the `last_polled_at` gate, not a guard. There's nothing to race because there's nothing to consume.)

## 4. High — one-click device-code phishing, from this PR's own parts

CSRF on `/approve` is **correctly handled** — it's inside the `/api/v1` nest that layers `csrf_middleware` (`routes.rs:1087`), a cookie-authenticated POST without `X-Requested-With` gets 403, and the page's `fetch` sends the header (`oidc_device.rs:273`). Credit where it's due.

But the undisclosed `sso.rs` change hands an attacker a same-origin, first-party URL:

```
https://ak.example.com/api/v1/auth/sso/oidc/<id>/login?return_to=%2Fdevice%3Fuser_code%3DABCD-EFGH
```

The victim clicks a link on our own domain, transits their IdP silently if they have a live session (#3546 shipped `prompt=none`), returns to `/device` already authenticated by cookies set on the redirect, sees the attacker's code **already in the box** (`oidc_device.rs:222`), and needs one click. The attacker's device then polls and gets a token as the victim, with the victim's full privileges per finding 1.

The page does carry the RFC 8628 §5.4 warning — "Never enter a code sent to you by someone else" (`:185`) — but the user didn't enter it; the page filled it in. The warning describes the attack the flow then performs on their behalf.

Fix: don't accept `user_code` from the query string on a page whose only button approves it (or require re-typing/confirming it character by character), refuse `return_to` values pointing at `/device`, and add an explicit confirmation naming the `client_id` and the scopes being granted.

## 5. High — `sso.rs` is a 96-line change to an existing auth file that the PR body never mentions

What changed: `OidcLoginQuery` gains `return_to` (`sso.rs:136`); new `pub(crate) sanitize_return_to` (`:145-156`); `oidc_login`'s return type changes `Redirect` → `Response` and sets a `Set-Cookie: ak_sso_return_to` (`:307-320`); `return_to_from_cookie` (`:159-171`) reads it in both callbacks; and `oidc_callback_inner` gains a 7th argument and **replaces `build_frontend_callback_url(&exchange_code)` as the redirect target** (`:819-826`).

**The sanitizer itself holds** — I tried `//evil`, `/\evil`, `/\/evil`, `\r\n` header splitting, `/%2f%2fevil.com`, `/ /evil`, and the 512-byte cap, and couldn't construct an open redirect. It should still ship with the tests its `pub(crate)` visibility implies; right now it has **zero**.

Undisclosed behaviour changes to existing SSO:

- When `return_to` is present the one-time exchange code is still minted and then **discarded**, so the `/exchange` POST that "re-sets cookies (idempotent)" (`:833-838`) no longer runs on that path. That should be a decision, not a side effect.
- The cookie is `HttpOnly; SameSite=Lax; Path=/; Max-Age=600` and **never `Secure`**, even though `oidc_callback_inner` already threads `client_is_https` for exactly this. The doc comment says it's "scoped to the SSO callback paths"; the code sets `Path=/`. Fix both.
- It is the enabling half of finding 4.

A generic "land anywhere same-origin after SSO" primitive added to the login endpoint, in a PR about a device grant, with no tests and no mention in the body, is exactly the shape an unrelated regression hides in. It belongs in the body at minimum; ideally it's a separate PR with its own tests.

## 6. Medium — the bespoke rate limiter should be deleted in favour of the existing one

`DEVICE_CODE_RATE_LIMITER: OnceLock<Mutex<HashMap<IpAddr, …>>>` (`oidc_device.rs:37-53`):

- **Unbounded growth** — nothing prunes; an IPv6 /64 gives 2⁶⁴ keys. Compare the comment at `routes.rs:503-526`, which is literally *"prevent unbounded HashMap growth from unique client IPs over time."*
- **Proxy attribution is inverted from what I expected** — it uses `ConnectInfo(addr).ip()`, the real TCP peer, so it is *not* spoofable via `X-Forwarded-For`. But behind the reverse proxy every AK deployment runs, every request carries the proxy's IP, so it's **10 device-code requests per minute for the entire internet** — a trivially triggered self-DoS on the feature. `RateLimitState` carries `trusted_proxies` CIDRs and a hardened XFF walk (`rate_limit.rs:606-800`, written for GHSA-8jm4-4x6c-6787) to get this right.
- No `RATE_LIMIT_ENABLED` kill switch (`routes.rs:444-447`), no exemption list, and it rejects with `AppError::Validation` → **400** in AK's envelope rather than 429 in the RFC 6749 §5.2 shape.

Registering a `RateLimiter` + `RateLimitState` the way the other six are (`routes.rs:440-530`) fixes growth, attribution, the kill switch and cleanup in one move. For the record: it does **not** hold the `Mutex` across an `await` — that's a synchronous fn and the guard drops first.

**Separately, RFC 8628 §5.1 rate-limiting on the `user_code` is absent.** The user code is 24⁸ ≈ 2³⁶·⁷ (the `% 24` modulo bias costs 0.01 bits — cosmetic, not a finding; the RNG is a proper CSPRNG and the 256-bit device code is fine). With only the generic 100/min `/api/v1` limiter, an attacker gets ~1000 guesses per IP per 600s code lifetime → P ≈ 2⁻²⁶·⁹ per IP per code, ~32× worse than the spec's stated target, degrading linearly with attacker IPs and with the number of live codes — a hit on *any* live code wins, since the attacker doesn't care whose device it is. Add a per-session failed-attempt counter that kills the session after ~5 wrong submissions.

## 7. Medium — `cleanup_expired` is wired to nothing

`grep -rn cleanup_expired backend/src` finds the rate-limiter calls at `routes.rs:513-523` and `token_service.rs:500`, but **no call to `OidcDeviceService::cleanup_expired`**, and no reference to `oidc_device_sessions` outside the two new files. The table is written by an **unauthenticated** endpoint and rows are never deleted, so it grows forever. One line inside the existing 60-second loop fixes it.

## 8. The `/device` page — no injection, and the architecture argument is sound

I traced every sink and found nothing. The HTML is a const raw-string literal with **no server-side interpolation at all**, so there's no reflected-XSS sink; the `user_code` is read client-side from `URLSearchParams` and passed through `normalize()` (`:209-212`), which upper-cases and strips everything outside `[A-Z]` before assigning to `input.value` as a property. All dynamic text goes through `textContent`, and the one `insertAdjacentHTML` inserts a static literal. CSP is inherited from `security_headers_middleware` and is correct — and commit `c8a27764` moving the script out of line into `/device/app.js` specifically to satisfy `script-src 'self'`, with a test, is good work.

On the architecture: your argument is **correct and better made than most**. Headless `web.enabled=false` deployments must still resolve `verification_uri`, the cross-repo version window is real, and the PEP 503 precedent holds. I'd accept the page in the backend. What I won't accept is the query-string pre-fill, which is an independent decision from where the page lives.

One line worth adding: assert `loginUrl.startsWith('/')` before `window.location.assign` (`:235`, `:263`).

## 9. Migration details

- `approved_user_id → users(id) ON DELETE SET NULL` is wrong for this row shape — deleting an approved user leaves `status='approved'` with NULL, which `poll_token` turns into `AppError::Internal("approved session missing user_id")` (`:158-160`) → a **500** with `error: "server_error"`. Use CASCADE, or map NULL to `access_denied`.
- The extra `CREATE INDEX ON oidc_device_sessions (user_code)` is **redundant** — the UNIQUE constraint already builds one. The poll path filters on `(device_code, client_id)`, covered by the unique index on `device_code`.
- Indexes are unnamed and without `IF NOT EXISTS`, and there's no `CREATE TABLE IF NOT EXISTS`; every other recent migration uses `CREATE INDEX IF NOT EXISTS idx_<table>_<cols>`. As written this migration is **not re-runnable**, which matters to the `SKIP_MIGRATIONS=true` path `main.rs:207-233` supports.
- No `CHECK` on `status` — land one with `consumed` from finding 3.

## 10. Lower

- **Fast pollers are permanently locked out.** `poll_token` updates `last_polled_at = now()` on the `slow_down` branch too (`:132-140`), so a client polling faster than `interval_secs` resets the clock every attempt and can never escape `slow_down`, even after approval. Only advance it when the gate passes.
- **No `Cache-Control: no-store` on the token response**, which RFC 6749 §5.1 makes a MUST. No other AK endpoint does it either, so it's a house gap — but a new endpoint claiming RFC conformance should set it.
- **`/device/code` errors aren't OAuth-shaped** — `{code, message}` rather than `{error, error_description}`, while `/device/token` correctly uses `OAuthErrorResponse`. The tooling you name at `:82-84` can't parse the former.
- **`deny_session` is unreachable dead code** — the body describes it as an endpoint, but `grep -n deny backend/src/api/routes.rs` finds no route and the page has no Deny button.
- **Unauthenticated provider enumeration** — omitting `provider_id` with several providers enabled returns every enabled provider's name and UUID to an unauthenticated caller.

## 11. Bookkeeping

- **CHANGELOG missing.** Add an `### Added` entry under `## [Unreleased]` citing **#3461** first, house style `- **<Title>** (#<issue>).`
- **OpenAPI not registered.** Nothing in `openapi.rs` for the four new endpoints, and `handler_sources` (`:631-700`) gains no `/api/v1/auth/oidc/device/` prefix. This will **not** fail CI — the coverage test walks spec→handlers, not handlers→spec — so it has to be caught here.
- A regression test isn't required for `feat/*` per `CONTRIBUTING.md:113-131`, so your 9 tests are above the floor. That doesn't cover finding 3's missing single-use test, which is a test for a security property the body claims to have.

---

**What I ran:** the full migration set against a real Postgres 16 (clean); the repeated-poll probe that produced five token pairs from one approval; a two-connection concurrent-poll probe; migration-number collision checks against `main` and every `release/*`; the CSP and sink trace; and `sanitize_return_to` probing. **What I reasoned from source without driving over HTTP:** the escalation chain in findings 1 and 2 — the code path is unambiguous (`generate_tokens` → `None, None`, and the handler reading only `user_id`), but I did not mint a scoped API token and walk it end to end. If you'd like, I'll do that before re-review so we're both certain.

Findings 1, 2 and 3 are what block merge. 4 and 5 I'd like resolved in the same pass, since they're coupled. The rest can follow.

I'd also ask you to split the `sso.rs` `return_to` work into its own PR. It's a reasonable primitive, it needs its own tests, and reviewing it inside a device-grant PR is how it got this far without anyone looking at it.

This is a lot of findings, but the shape of the feature is right and most of these are mechanical. The two that need design thought are the scope ceiling and the approval binding.

brandonrc:
## Where this lands, and what we'll accept

Thanks again for this — the device flow is a feature we want, the CSP and headless-hosting reasoning in the branch is better than most first submissions, and the defects are fixable rather than fundamental. I've taken it through a wider review (security, product, operations, standards) and here is the decision, so the scope is not a moving target.

**It will not ship in 1.10.0.** I'm moving it off that milestone. The fixes below are a design-level pass, not review nits, and 1.10.0 is closing.

**The shape we'll merge:**

1. **A plain AK device authorization grant, not "OIDC".** Today `oidc_sub`/`oidc_email`/`oidc_name` are `String::new()` and the approver authenticates by whatever AK already accepts — the provider FK is decoration. Naming it OIDC creates a client expectation nothing can meet: AK publishes no `/.well-known/openid-configuration`, and a client configured against it as an OIDC device endpoint sends `scope=openid`, gets no `id_token`, and either fails or — worse — accepts a token whose subject is "whoever clicked approve". Rename to `POST /api/v1/auth/device/{code,token}`, drop the `oidc_configs` coupling. A real IdP-backed approval leg can be added later on the same endpoint.
2. **Scope ceiling on the mint.** `generate_tokens(&user)` → `generate_tokens_with_scope`, capped at min(approver's scopes, requested scopes), with repo allow-list propagated — the same thing every other credential-exchange path in the tree does (`conan.rs:367`, `oci_v2.rs:432` …). And `/approve` must refuse non-interactive principals (`is_api_token`, `is_service_account`). As it stands a read-only API token can mint itself an unrestricted JWT plus a multi-day refresh token in three unauthenticated-then-header-authenticated calls; that's the laundering #2430 closed elsewhere.
3. **Single-use redemption.** A `consumed` state with a `CHECK`, and redemption as one conditional `UPDATE … WHERE status='approved' … RETURNING`. Measured on the branch: five polls of one approved code returned five token pairs. The PR description says `pending → approved → consumed` — that needs to be true.
4. **No query-string pre-fill of `user_code`, and no `verification_uri_complete`, for now.** The pre-fill plus the `return_to` change together make a one-click phish from a first-party URL. A confirmation screen doesn't fix it while `client_id` is unregistered (it would render attacker-chosen text in trusted chrome). `verification_uri_complete` is OPTIONAL in RFC 8628 §3.3.1 and every mainstream client falls back cleanly; we'll bring it back together with a client registry and a real consent screen.
5. **The `sso.rs` `return_to` change comes out into its own PR**, with tests for the cases that were probed by hand (`//evil`, `/\evil`, CRLF, `%2f%2f`, the 512-byte cap, `Secure` on HTTPS). It's a useful primitive; it just can't ride along undocumented.
6. **Use the existing `RateLimitState`** instead of the bespoke `OnceLock<Mutex<HashMap>>`. Behind the reverse proxy every deployment runs, the current one throttles the entire internet to 10 device-code requests/minute, answers 400 in AK's envelope rather than 429 in the OAuth shape, and has no kill switch. Plus: wire `cleanup_expired` into the existing 60-second sweep, `ON DELETE CASCADE` for `approved_user_id`, migration renumbered to the next free slot, CHANGELOG and OpenAPI entries, and `{error, error_description}` on `/device/code` to match `/device/token`.

**What I'd ask of you:** reply within two weeks (by 2026-09-30) if you want to carry this. If I don't hear back, I'll take it over on this branch — your commits and authorship stay — and land it in 1.11.0 with the changes above. Either way the work you did here is the base.

vivekdsr:
Hello @brandonrc , i tried to fix the comments but i will not be able to contribute to this PR further, could you please take over so that its available in the e next  releases

brandonrc:
> Hello @brandonrc , i tried to fix the comments but i will not be able to contribute to this PR further, could you please take over so that its available in the e next  releases 

Sorry soending alot of my time improving and optimizing the workflows (pipelines)

brandonrc:
Thanks for the work on this, @vivekdsr, and for saying so clearly. We'll take it from here.

The plan is the shape from the 2026-09-16 comment: a plain AK device authorization grant (`/api/v1/auth/device/{code,token}`) with no OIDC coupling, and the security fixes listed there. Those are a scope ceiling on the minted token, interactive approvers only, a single-use device code, and the remaining review items. The work goes on your branch as new commits on top of yours, so your authorship stays. It gets a security review before merge, and your name goes in the changelog entry. I'll post here what changed once it's pushed.

Issue #3559, opened 2026-08-25
Add a gate asserting published image tags resolve to cosign-signed bytes


brandonrc:
## Summary

Every published backend release tag on ghcr.io resolves to an image index that carries **no cosign signature**, despite `docker-publish.yml` running `cosign sign --yes` as a hard-failing step. The signatures are being produced — they are just attached to a different digest than the one the release tag ends up pointing at.

## Evidence

Paged the full tag list for `ghcr.io/artifact-keeper/artifact-keeper-backend` (3,507 tags). There are **472** `sha256-<digest>.sig` tags, so signing runs and succeeds. None of them match the digest any release tag resolves to:

| tag | index signed | per-arch signed |
|---|---|---|
| 1.6.4 | no | 0/2 |
| 1.7.0 | no | 0/2 |
| 1.7.1 | no | 0/2 |
| 1.7.3 | no | 0/2 |
| 1.7.4 | no | 0/2 |
| 1.7.6 | no | 0/2 |
| 1.7.8 | no | 0/2 |
| 1.8.0 | no | 0/2 |
| 1.8.1 | no | 0/2 |
| `latest` | no | 0/2 |

Checked both the tag-based scheme (`sha256-<digest>.sig`) and the OCI referrers API (`/v2/.../referrers/<digest>` returns `MANIFEST_UNKNOWN`). Neither carries a signature for the published digests.

The signed digests are themselves valid 4-child indexes (2 platform manifests + 2 attestation manifests), i.e. the same shape as a published index — so cosign is signing a real multi-arch index, just not the one that ships.

## What is NOT broken

Attestations are fine, and this is worth stating because the two are easy to conflate. The published `1.8.1` index carries per-architecture in-toto attestations that survived the `imagetools` merge:

```
application/vnd.in-toto+json  2481269  in-toto.io/predicate-type: https://spdx.dev/Document
application/vnd.in-toto+json   108311  in-toto.io/predicate-type: https://slsa.dev/provenance/v1
```

So `sbom: true` and `provenance: mode=max` are working end to end. Only the cosign signature is missing from what ships.

## Why this matters more than an ordinary bug

[...]

PR #3563, opened 2026-08-25
ci: sign and verify the Docker Hub mirrors, not just ghcr (#3562)

### LINKED ISSUE #3562: Docker Hub mirrors are published unsigned: cosign sign targets only the ghcr ref
## Summary

`docker-publish.yml` mirrors every published image to Docker Hub, but `cosign sign` targets **only the ghcr.io reference**. The Docker Hub mirrors therefore carry no signature at all: `docker.io/artifactkeeper/backend`, `docker.io/artifactkeeper/openscap` and `docker.io/artifactkeeper/scanner-adapter` are unsigned for every tag they have ever published.

Same invariant as #3559, different distribution channel: **the bytes a user pulls by tag should be the bytes that were signed, regardless of which registry they pull from.**

## Evidence (read-only, 2026-08-25)

The bytes are identical across the two registries — only the signature is missing on one side:

| tag | ghcr.io digest | docker.io digest | ghcr signed | Hub signed |
|---|---|---|---|---|
| `backend:1.8.1` | `sha256:44e433bc…` | `sha256:44e433bc…` | yes | **no** |
| `backend:1.8.0` | `sha256:178fb460…` | `sha256:178fb460…` | yes | **no** |
| `backend:1.7.8` | `sha256:2ed8a904…` | `sha256:2ed8a904…` | yes | **no** |
| `backend:latest` | `sha256:44e433bc…` | `sha256:44e433bc…` | yes | **no** |

Run against the gate added by #3560, unmodified — it already resolves Docker Hub refs correctly:

```
$ scripts/ci/check-published-image-signature.sh docker.io/artifactkeeper/backend:1.8.1
  docker.io/artifactkeeper/backend:1.8.1
    tag resolves to sha256:44e433bc856b5e8dceb63b803e5ff1d96b0584fbfe3ea676d5d52592555ca333
error: Published image is not verifiably signed: ... cosign found no acceptable signature for it.
BLOCKED: 1 published tag(s) do not resolve to signed bytes
EXIT=1

$ scripts/ci/check-published-image-signature.sh ghcr.io/artifact-keeper/artifact-keeper-backend:1.8.1
    signed: 1 cosign signature(s) over sha256:44e433bc...
EXIT=0
```

A cosign signature is scoped to the repository it was pushed to. The ghcr signature covers `sha256:44e433bc…` in `ghcr.io/artifact-keeper/artifact-keeper-backend`; it says nothing about the identical digest in `docker.io/artifactkeeper/backend`, because a verifier looking at Docker Hub never sees it. Anyone pulling from Docker Hub has no way to verify what they got.

## Why the existing controls do not catch it

* `cosign sign` is hard-failing (#2824, #3496) — but it is only ever handed the ghcr ref, so it succeeds while doing nothing for Hub.
* `check-supply-chain-soft-fail.sh` asserts that signing steps cannot fail silently. It does not assert *what* they sign.
* The #3559 gate re-resolves published tags, but is wired with the ghcr tag set only. A gate that checks one of two registries leaves exactly this hole.

## Registry behaviour worth recording

Measured directly, because the two registries differ and the difference drives what signing writes:

* **ghcr.io does NOT implement the OCI referrers API.** `GET /v2/<name>/referrers/<digest>` answers `404 MANIFEST_UNKNOWN` for every digest, signed or not. cosign v3 falls back to the `sha256-<digest>` tag scheme (no `.sig` suffix).
* **Docker Hub DOES implement it.** `GET /v2/<name>/referrers/<digest>` answers `200` with a real `application/vnd.oci.image.index.v1+json`. It returns an empty `manifests` array for our (unsigned) images and a populated one for repositories that have referrers — e.g. `bitnami/nginx:latest` returns one `application/vnd.cncf.notary.signature`.
* On Docker Hub the same referrer index is **also** addressable as the `sha256-<digest>` tag: for `bitnami/nginx:latest` the referrers response and `GET manifests/sha256-<digest>` return byte-identical content. So signing on Hub should be expected to make a `sha256-<digest>` tag visible in the **public** repository's tag list.

That last point is a visible change to what users see listed on Docker Hub and is a maintainer decision, not an implementation detail.

## Scope

`merge-backend`, `merge-openscap` and `merge-scanner-adapter` all mirror to Docker Hub and all sign ghcr only. `merge-backend-alpine` has the same shape but is `if: false` (Alpine builds suspended), so it publishes nothing today. `apply-floating-tags` only re-points tags at an already-published digest, so it needs no signature of its own — but it does mean a Hub signature must be over the *digest*, so that floating tags inherit it.

## Acceptance criteria

1. Docker Hub refs are signed with cosign keyless, by digest, the same way ghcr refs are.
2. The #3559 published-tag gate covers the Docker Hub tag set as well as the ghcr one, proven red against a genuinely unsigned published Hub image.
3. Something prevents silent regression at the workflow level: a job that mirrors to Docker Hub and does not sign/verify it must fail CI.
4. Decide and document whether already-published Hub images are retro-signed.
5. The Docker Hub tag-surface change is stated explicitly so a maintainer can accept or reject it.

brandonrc:
Closes #3562.

**Stacked on #3560** — base branch is `ci/3559-verify-published-image-signature`, not `main`. Review that one first; the diff here is only this commit.

## TL;DR

`cosign sign` was only ever handed the ghcr reference, so **every Docker Hub tag this project has published is unsigned**. The bytes are right — `backend:1.8.1` resolves to the same `sha256:44e433bc…` on both registries — but a cosign signature lives in the repository it was pushed to, so the ghcr signature is invisible to anyone verifying from Docker Hub. This signs the `docker.io` digest too, extends the #3559 gate to cover the Hub tag set, and adds a static gate so the absent half cannot quietly come back.

Two things need a maintainer decision rather than a review: **the Docker Hub tag surface changes** (§"Surface change") and **already-published Hub images are not retro-signed** (§"Retro-signing").

## Why stack on #3560 rather than rebase later

Every line this change touches in `docker-publish.yml` is a line #3560 adds or rewrites — the `Sign image with cosign (keyless)` steps and the `Verify the published tags resolve to signed bytes` steps in all three merge jobs. Independent branches would conflict in three places on a 2,600-line workflow, and the more useful half of the fix (the Hub refs in the verify list) has nothing to attach to on `main` at all. #3560 also already does the hard part: its `check-published-image-signature.sh` resolves `docker.io` refs correctly, `registry-1.docker.io` endpoint and token dance included. That claim was verified live rather than taken on trust — see the first block under "Proof". No part of that script is duplicated or re-implemented here.

## The gap, measured (read-only, 2026-08-25)

| tag | ghcr digest | docker.io digest | ghcr | docker.io |
|---|---|---|---|---|
| `backend:1.8.1` | `sha256:44e433bc…` | `sha256:44e433bc…` | signed | **unsigned** |
| `backend:latest` | `sha256:44e433bc…` | `sha256:44e433bc…` | signed | **unsigned** |
| `openscap:1.8.1` | `sha256:faf3e4b2…` | `sha256:faf3e4b2…` | signed | **unsigned** |
| `scanner-adapter:latest` | `sha256:8b9d82d0…` | `sha256:8b9d82d0…` | signed | **unsigned** |

Identical digests on both sides, for all three images, including the scanner-adapter whose Hub copy re-applies index annotations. That matters twice over: it is why one digest is the right subject on both registries, and it is why the missing signature is the *whole* of the defect.

### Why nothing caught it

Nothing was broken. A whole distribution channel was simply absent from the controls, and absence is invisible to a runtime check:

* `cosign sign` is hard-failing (#2824, #3496) and succeeded — it did exactly what it was asked.
* `check-supply-chain-soft-fail.sh` asserts that a signing step cannot fail silently. Not *what* it signs.
* #3559's gate is a real assertion, but it can only assert about the refs it is handed, and it was wired with the ghcr tag set. A gate over one of two registries, staying green over the other for the project's entire history.

## What changed

* **`docker-publish.yml`** — in `merge-backend`, `merge-openscap` and `merge-scanner-adapter`: `cosign sign` the `docker.io` **digest** alongside the ghcr one, and pass the Docker Hub tag set to the verification gate alongside the ghcr one. No `continue-on-error` anywhere; `check-supply-chain-soft-fail.sh` passes (9 control steps).
* **`scripts/ci/check-registry-publish-parity.sh`** (new) — reads the workflow and fails if any publish job mirrors to a registry it does not sign *and* verify. Exit 1 = drift, exit 2 = INFRA (file missing/unparsable). Non-vacuity floor: scope is defined by the presence of a cosign step, so fewer than 3 in-scope jobs is itself a failure — otherwise the gate would be satisfiable by deleting a signing step.
* **`scripts/ci/test-check-registry-publish-parity.sh`** (new) — 13 offline cases, `expect "<name>" <exit> "<expected>"`, workflow fixtures. ~1s.
* **`scripts/ci/check-published-image-signature.sh`** — one behaviour change, described under "Credential scoping" below.
* **`scripts/ci/test-check-published-image-signature.sh`** — 19 → 25 cases (Docker Hub refs; both halves of the credential scoping).
* **`ci.yml`** — the parity gate and its self-test run in 🐚 Shell Tests.

### Signing by digest is load-bearing

`apply-floating-tags` re-points `:latest` / `:X.Y` on both registries at a digest the merge jobs already published, uploading nothing — so a signature over the digest keeps covering the floating tags and that job stays build-free. It is also self-asserting: `imagetools create` copies a manifest list without rewriting it, so if a Hub copy ever produced different content, `cosign sign` fails immediately on a manifest that is not there instead of attaching a signature to the wrong bytes.

### Credential scoping (a real hole this change would otherwise have opened)

The verification gate takes an optional registry credential for resolving tags on a private registry. The token endpoint that credential is presented to is named by a `WWW-Authenticate` header **the registry sends**. ghcr's realm is on `ghcr.io`; Docker Hub's is `auth.docker.io` — a different origin. Dropping a `docker.io` ref into an invocation carrying `${{ github.actor }}:${{ secrets.GITHUB_TOKEN }}` would have handed a GITHUB_TOKEN to Docker's auth server. The credential is now pinned to the single host it was issued for (`PUBLISHED_SIG_BASIC_AUTH_HOST`, default `ghcr.io`); public mirrors resolve anonymously, so nothing is lost. Both directions are tested: case 22 asserts the credential *is* still offered on its own host (so a private ghcr package does not turn into an INFRA failure every run), case 23 asserts it is *not* sent to Docker Hub. Removing the scope turns case 23 red.

## Proof it is not theater

**1. #3560's script, unmodified, against the two live published images:**

```
$ scripts/ci/check-published-image-signature.sh docker.io/artifactkeeper/backend:1.8.1
  docker.io/artifactkeeper/backend:1.8.1
    tag resolves to sha256:44e433bc856b5e8dceb63b803e5ff1d96b0584fbfe3ea676d5d52592555ca333
error: Published image is not verifiably signed: ... cosign found no acceptable signature for it.
BLOCKED: 1 published tag(s) do not resolve to signed bytes
EXIT=1

$ scripts/ci/check-published-image-signature.sh ghcr.io/artifact-keeper/artifact-keeper-backend:1.8.1
    signed: 1 cosign signature(s) over sha256:44e433bc...
EXIT=0
```

**2. The new `merge-backend` verify step, replayed verbatim against the live registries** — same ref construction (`jq -r '.[]' /tmp/ghcr-tags.json /tmp/hub-tags.json`), cosign v3.1.3:

```
  ghcr.io/artifact-keeper/artifact-keeper-backend:1.8.1
    tag resolves to sha256:44e433bc...
    signed: 1 cosign signature(s) over sha256:44e433bc...
  ghcr.io/artifact-keeper/artifact-keeper-backend:latest
    signed: 1 cosign signature(s) over sha256:44e433bc...
  docker.io/artifactkeeper/backend:1.8.1
    tag resolves to sha256:44e433bc...
error: Published image is not verifiably signed: ...
  docker.io/artifactkeeper/backend:latest
    tag resolves to sha256:44e433bc...
error: Published image is not verifiably signed: ...

BLOCKED: 2 published tag(s) do not resolve to signed bytes:
  - docker.io/artifactkeeper/backend:1.8.1
  - docker.io/artifactkeeper/backend:latest
EXIT=1
```

Same run, ghcr green and Hub red, on identical digests. `openscap:1.8.1` and `scanner-adapter:latest` behave the same. `docker.io/artifactkeeper/backend:1.8.1` is a real, published, genuinely unsigned image and the gate goes red on it.

**3. The static gate against `docker-publish.yml` as it stood before this commit** — it names the defect per job:

```
$ git stash && ./scripts/ci/check-registry-publish-parity.sh
  - docker-publish.yml :: merge-backend
    mirrors an image to Docker Hub but its `cosign sign` step(s) reference only ghcr. ...
  - docker-publish.yml :: merge-backend
    verifies published tags on ghcr only. ...
  (same for merge-openscap and merge-scanner-adapter)
EXIT=1

$ git stash pop && ./scripts/ci/check-registry-publish-parity.sh
OK: 3 publish job(s) in docker-publish.yml sign and verify every registry they publish to.
EXIT=0
```

That configuration is case 2 of the self-test, and it is also case 11: the self-test runs the gate against the repository's *real* workflow, so the logic stays anchored to the thing it is about rather than to fixtures.

**Revert-proof.** Removing the `docker.io` line from a sign step, removing the Hub tag set from a verify step, removing a verify step entirely, dropping below 3 in-scope jobs, or removing the credential host-scope each turns a specific named case red. 25/25 and 13/13 pass; `shellcheck -S style` clean; both workflows parse and no `needs:` dangles.

## Surface change — maintainer's call

**Signing writes a signature manifest into the *public* Docker Hub repositories, and on Docker Hub it is very likely to appear as a `sha256-<digest>` tag in the public tag listing.** Measured rather than assumed, and the two registries genuinely differ:

* **ghcr.io does NOT implement the OCI referrers API.** `GET /v2/<name>/referrers/<digest>` → `404 MANIFEST_UNKNOWN` for every digest, signed or not. cosign v3 falls back to the `sha256-<digest>` tag (no `.sig` suffix) — confirmed on our own signed 1.8.1, where that tag returns an index containing `application/vnd.dev.sigstore.bundle.v0.3+json`.
* **Docker Hub DOES implement it.** `GET /v2/artifactkeeper/backend/referrers/sha256:44e433bc…` → `200`, `Content-Type: application/vnd.oci.image.index.v1+json`, with an empty `manifests` array (correct: nothing is signed). It is a real implementation, not a stub — `bitnami/nginx:latest` returns one `application/vnd.cncf.notary.signature`.
* **On Docker Hub the referrer index is also addressable as the `sha256-<digest>` tag.** For `bitnami/nginx:latest`, the referrers response and `GET manifests/sha256-<digest>` return byte-identical content, and that repository's tag listing is full of `sha256-…` tags.

So the honest statement is: whichever path cosign takes on Hub, the referrer becomes readable both ways, and on this evidence you should **expect a `sha256-<digest>` tag per signed digest to show up in the public tag list** on Docker Hub — one per image per release, plus one per floating-tag digest. It is the signature, and it is what makes `cosign verify docker.io/artifactkeeper/backend:<version>` work. If that tag noise on a public repository is not acceptable, this is the thing to say no to; the gate and the parity check are independent of which storage path is used.

## Retro-signing — recommendation: **no**

Consistent with the ghcr position on `1.6.1` in #3560, and for the same reason. `cosign sign docker.io/artifactkeeper/backend@sha256:44e433bc…` would work without a rebuild, but the signature would be produced **today**, by a workflow run that did not build those bytes, asserting a relationship between an identity, a commit and an artifact that never existed. That trades an accurate "Docker Hub tags before this change are unsigned" for a signature whose provenance is fiction — the exact failure mode this whole area exists to prevent. Applying it to a whole registry's history rather than one release makes it worse, not better: it would manufacture a clean-looking signed history for tags nobody verified at the time.

The recommendation is therefore: **tags published from the first release carrying this change onward are signed on both registries; earlier Docker Hub tags stay unsigned and are documented as such.** A consumer who needs a verifiable image today can pull the identical bytes from ghcr.io, where they are signed — and "identical" is not a hope, it is the digest table above.

Retro-signing would need a registry **write**. This work was read-only throughout; nothing was pushed, deleted or re-tagged. If it is wanted anyway, it should be a deliberate, logged, separately-approved action, not a side effect of this PR.

## Notes

* `merge-backend-alpine` has the same defect and is **not** fixed here: it is `if: false` (Alpine builds suspended), so it publishes nothing. The parity gate skips literal `if: false` jobs and says so in its output; re-enabling that job changes the `if:` and brings it into scope on the same PR, so the exemption expires by itself. Case 7 of the self-test pins both halves.
* `apply-floating-tags` is deliberately out of scope: it writes no new bytes, only re-points tags at an already-signed digest, and it already asserts digest identity itself. Case 8 pins that it is not dragged in.
* Draft: the live behaviour of the new signing step is only observable on a real `main`/tag push run of `docker-publish.yml`.
* No overlap with #3561 — `release.yml` and the binary-signing scripts are untouched.


brandonrc:
CI note: because this PR targets `ci/3559-verify-published-image-signature` rather than `main`, the full workflow set does not trigger — only the linked-issue gate and GitGuardian ran (both green). The gates and self-tests were run locally against this exact tree:

```
./scripts/ci/check-supply-chain-soft-fail.sh          OK: 9 supply-chain step(s) hard-fail their job
./scripts/ci/check-registry-publish-parity.sh         OK: 3 publish job(s) sign and verify every registry
bash scripts/ci/test-check-supply-chain-soft-fail.sh  passed: 10   failed: 0
bash scripts/ci/test-check-published-image-signature.sh    25/25
bash scripts/ci/test-check-registry-publish-parity.sh      13/13
shellcheck -S style (4 scripts)                       clean
workflow YAML parse + needs: reachability             clean
```

Shell Tests will run for real once #3560 merges and this rebases onto `main`.

Issue #3565, opened 2026-08-25
A scan that exceeds its wall-clock budget leaks its workspace directory permanently


cazlo:
A scan cancelled by its wall-clock budget never removes its workspace directory, and nothing else ever will. On an instance where scans time out regularly, this accumulates without bound on the scan-workspace volume.

`ScanWorkspace` is a unit struct (`pub(crate) struct ScanWorkspace;` — `scanner_service.rs:934`), so it holds no state and has no `Drop`. Cleanup is an explicit async call, and across four scanner implementations it sits *after* the scan completes at five sites:

- `grype_scanner.rs:1907` — `ScanWorkspace::cleanup(&self.scan_workspace, None, artifact).await`, after the grype run and findings conversion
- `trivy_fs_scanner.rs:325`, `openscap_scanner.rs:349`, and `incus_scanner.rs:336` and `:811` — same shape (two Incus sites)

The inline proxy scan path wraps that whole future in `tokio::time::timeout` (`proxy_helpers.rs`, `proxy_scan_and_record`). When the budget expires, the future is dropped before control ever reaches the cleanup line, so the directory survives. There is no `Drop` guard, and no filesystem reaper anywhere in the codebase — the "janitor" service reaps stale `scans` rows in Postgres, not files on disk.

Each timed-out attempt leaks a *fresh* directory rather than reusing one: `proxy_rescan_synthetic_artifact` mints a new `Uuid::new_v4()` per rescan, so the workspace path is unique per attempt and repeated attempts never collide or overwrite.

What is left behind is the staged scan input — the artifact bytes written under their real filename, the CycloneDX BOM sidecar, and for archive formats the entire extracted tree, which can be substantially larger than the artifact itself.

This matters operationally because the scan workspace is a dedicated fixed-size volume (`scanWorkspace` in the Helm chart), not an `emptyDir` that a pod restart would clear. It fills, and the instances most likely to time out — CPU-starved, IO-starved, or memory-constrained — are exactly the ones that will time out repeatedly.

**Reproduction:** run a scan under a wall-clock budget shorter than the scan takes (throttling block-device reads to 50 MB/s puts a trivial input past 30s on its own), then inspect the workspace base. The directory for the cancelled attempt is present, and a second attempt adds another rather than reusing it.

[...]

PR #3566, opened 2026-08-25
fix(scanner): distinguish inconclusive proxy rescans and give them an operator budget

### LINKED ISSUE #3455: Proxy rescan returns 503 inconclusive with grype: scanner invoked on extensionless __content__ blob
## Summary

`POST /api/v1/repositories/{key}/security/proxy-scans/rescan` returns a 503 inconclusive result for proxy-cached wheel artifacts on a 1.8.0 instance whose filesystem scanning is backed by grype. The scan never actually inspects the package contents, because the code path hands the scanner an extensionless file it cannot identify as an archive.

## Environment

- Backend 1.8.0, with a grype-based filesystem scanner. The runtime image does not bundle a Trivy CLI binary; container/OCI image scanning goes through a separate scanner-adapter service, so grype is what filesystem-type artifact scans actually use here.
- A demo deployment with a PyPI proxy repository.

## Steps to reproduce

1. Configure a PyPI proxy (remote) repository.
2. Pull a wheel through the proxy so it gets cached, for example `urllib3==1.24.1` as a wheel.
3. Call `POST /api/v1/repositories/{key}/security/proxy-scans/rescan` with `{"path": "<the cached path>"}`.
4. After about 30 seconds (the inline scan budget), the request returns:

   `503 {"code":"SERVICE_UNAVAILABLE","message":"Rescan was inconclusive: no scanner is configured, or the scan failed or exceeded its budget. The stored verdict is unchanged."}`

## Root cause

The proxy cache stores each cached artifact as a directory rather than a single file: `<cache path>/{__cache_meta__.json,__content__}`, where `__content__` holds the actual bytes with no file extension.

I reproduced this directly by running grype against the stored content inside the running container:

- Running grype against the wrapping directory path produces a `directory` source scan (0 packages found), matching the API's roughly 30 second timeout.
- Running grype against `__content__` directly (the actual bytes) is correctly identified as a `file` source, but still returns 0 matches after about 30 seconds, even though the urllib3 1.24.1 wheel used in the test has known published CVEs.

The cause is that `__content__` carries no file extension, so grype's archive cataloger does not recognize it as a zip-based Python wheel and never unpacks its dist-info/METADATA to identify the package. It is treated as an opaque unknown file, so the scan completes with no findings and no error, and the whole call still spends the full budget doing so.

This is isolated to the proxy-cache code path. As a control, I uploaded the same wheel to a hosted (non-proxy) repository and ran the normal artifact scan endpoint. That completed correctly through grype, with status completed and accurate findings. The hosted-artifact scan path evidently materializes the artifact with its original filename and extension before invoking the scanner; the proxy-cache rescan path does not.

## Suggested direction

Before invoking the scanner against proxy-cached content, materialize a temporary hardlink or copy of `__content__` that carries the artifact's original filename and extension (the filename is already known and recorded, since it is part of the cache path), or pass an explicit format or type hint to the scanner invocation so archive detection does not depend on inferring the type from the file name.

## Notes

- I did not test whether a Trivy-CLI-backed filesystem scanner has the same problem; this instance had no Trivy CLI available to compare against. The root cause, an extensionless input file defeating archive-type detection, is a property of what gets handed to the scanner, so it plausibly affects any scanner invoked the same way, not only grype.
- `scan_on_proxy` (the inline gate applied at download time) likely goes through the same code path and is probably affected the same way, though I only directly tested the explicit rescan endpoint.

cazlo:
fixes #3455

## Summary

Brandon's correction on #3455 is right: the synthetic artifact already preserves the real filename, so the extensionless cache-key theory is not the cause.

This changes `proxy_scan_and_record` from an undifferentiated `Option` to `Result<ProxyScanVerdict, ProxyScanInconclusive>`. The rescan endpoint returns the standard `{code, message}` 503 body with one of `NO_SCANNER_CONFIGURED`, `RESCAN_SCAN_FAILED`, or `RESCAN_BUDGET_EXCEEDED`; the OpenAPI response points at `ErrorResponse`. `RESCAN_THROTTLED` remains the precedent for a throttled HTTP status with `Retry-After`, not for the response-body schema.

Rescans now have their own 120-second budget instead of the inline download gate's 30-second latency budget. The repository cooldown is claimed only after the cached object is read and prepared, so an unreadable object still returns the canonical 404 without consuming a slot. `NoScanner` releases its just-claimed slot; `ScanFailed` and `BudgetExceeded` remain conservatively throttled. `Retry-After` is emitted only when cooldown time remains and rounds fractional seconds up so clients do not retry early.

The observed fixed scanner startup cost can exceed 30 seconds under CPU constraint. Separate block-read throttling on local NVMe also exceeded 30 seconds: 40.8 seconds at 50 MB/s and 101 seconds at 20 MB/s, with full CPU and memory. That was a local-NVMe throttle experiment, not a test of network-attached storage; network-attached storage behavior remains untested. The 120-second budget clears the slowest completed local-throttle run by about 19 seconds. A 1 MB/s run was still incomplete after 129 seconds and was stopped, so it is not a completed worst-case measurement.

Every timeout-reachable scanner, extraction, workspace-permission `chmod`, and shared CLI version-probe child now uses `kill_on_drop(true)`, so cancellation owns the process rather than leaving it running after the future is dropped.

The retained red regression proves the previous handler conflated the inconclusive causes. New focused coverage pins the typed 503 shapes, conditional upward-rounded `Retry-After`, missing-storage 404 followed by an immediate unthrottled retry, the rescan-specific budget, and cancellation ownership for cleanup and shared version-probe commands.

## Test Checklist

- [ ] Integration tests added/updated (not applicable; the endpoint regressions are DB-backed lib tests)
- [ ] Manually tested locally

Validation completed all 18 jobs, including six focused assertions under `--no-tests=fail`: 15,903/15,903 tests passed, 10 were skipped, line coverage was 87.89%, and changed-line coverage was 93% (51,037/54,306). The inline gate remains tested at its actual layers: fail-closed decision coverage observes 423, while fail-open returns pending before its background scan observes a timeout; this does not claim every inconclusive arm runs end-to-end through `gate_proxy_scan_serve`.

## API Changes

- [ ] Migration is reversible (not applicable; no migration)
- [ ] N/A - no API changes

## Not in this PR

- Inline download-gate budgets and fail-open/fail-closed posture.
- An asynchronous rescan API.
- Scanner warm-up or database pre-paging.
- Cancellation-safe workspace-directory cleanup is tracked separately in #3565.


brandonrc:
Reviewed at `503804fa52292dfc0a672ab046fd32f5761779e5` on a live deployment. The direction here is right and most of the PR holds up under attack — the `Retry-After` rounding is correct in every branch, slot accounting has no leak on `?` and no double-release, the error bodies leak nothing, and the new tests genuinely execute in CI against a real Postgres. Details of what I could not break are at the end.

Two findings are blocking, both reproduced live, and both are introduced by this PR rather than pre-existing.

## F1 — the cooldown claim moved behind the expensive work, so every *rejected* rescan now does a 200 MiB read + SHA-256

`security.rs:2199` (`storage.get`) → `:2211` (`sha256_hex`) → `:2225` (`claim_rescan_slot`).

The bullet "cooldown is claimed only after the cached object is read and prepared" moves the endpoint's **only throttle** behind its most expensive operation. Same binary, same rig, 16 concurrent POSTs against a 200 MiB cached object, only the claim ordering differing:

| | order | result |
|---|---|---|
| base `a46a252c` | size-check → **claim** → read → hash | 15 × `429` in 0.047–0.118 s; 1 × `503` at 10.3 s; peak RSS **534 MiB** |
| this PR | size-check → read → hash → **claim** | **16 × `429`, each 12.6–16.1 s**; peak RSS **3334 MiB** (idle baseline 137 MiB) |

Every rejected request buffers `PROXY_SCAN_MAX_BYTES` (200 MiB, `scanner_service.rs:93`) into heap and hashes it before being told no. On S3/GCS/Azure that is also a full object GET per rejected request.

The reach is wider than the PR assumes. It describes the endpoint as authenticated and write-gated, but `require_repo_write_access` (`repositories.rs:110-117`) returns `Ok` immediately when `repo.is_public`:

```rust
require_repo_access(auth, repo.id)?;
if repo.is_public || auth.is_admin {
    return Ok(());
}
```

That helper is documented as deliberately tenant-gate-only and action-blind, with the action half layered separately at the upload/delete handlers — so unless the rescan handler layers it too, any authenticated user reaches this on a public repo. Confirmed live: a non-admin with no grant at all drove the full 200 MiB read on a public remote repo (9.5 s); the same user on a private repo correctly gets 403. With `GLOBAL_MAX_CONCURRENCY` at 512 that is a large amount of concurrent buffer, and `RATE_LIMIT_API_PER_MIN` defaults to 10000, so the limiter is not a control here. (Worth noting separately: the comment at `routes.rs:1071` says "100 req/min" and is stale.)

The suggested fix keeps the stated goal: claim first, then call the `release_rescan_slot` this PR already adds on the 404 / read-failure path. That still avoids consuming a slot for an unreadable object, without doing the work first.

One thing to change regardless of the fix chosen: `security.rs:3067-3070` pins the current ordering as a permanent test, reasoning that "a pure claim/release test cannot prove the handler keeps the storage read ahead of the claim." That asserts the amplification as correct behaviour and would block the fix above.

## F2 — the 120 s budget equals the 120 s global request timeout, so `RESCAN_BUDGET_EXCEEDED` is unreachable and #3455's own scenario still returns the old generic 503

- `PROXY_RESCAN_BUDGET` = 120 s (`scanner_service.rs:125`)
- `global_request_timeout_secs` default = 120 s (`config.rs:1338`, and the `1075` default), applied by `conditional_request_timeout` (`routes.rs:317`) to every path where `is_byte_transfer_path` is false. The rescan path is false — `routes.rs:267-300` exempts only the `artifacts` / `download/` tails.

The outer clock starts at request arrival. The budget clock starts only after auth, two DB lookups, the full object read and the SHA-256, and the verdict/SBOM/CVE writes (`proxy_helpers.rs:6917-6968`) also sit outside the budget. The outer timeout therefore always wins. With `GLOBAL_REQUEST_TIMEOUT_SECS=5` to make it observable:

```
503 {"code":"SERVICE_UNAVAILABLE","message":"Server overloaded, please retry: request timed out"}
```

which is exactly the undifferentiated 503 this PR exists to replace. So a rescan in the 110–120 s band — precisely the CPU-starved case the PR's own measurements target — is killed by the backstop, records no verdict, and leaves the slot claimed, since nothing releases on cancellation. The budget test calls the handler function directly rather than through the router, which is why it cannot see this.

Suggested fix: derive `PROXY_RESCAN_BUDGET` from `config.global_request_timeout_secs` with headroom rather than hardcoding it, or exempt the route, or make the rescan async.

## Non-blocking

**F3** — `scanner_service.rs:1342` uses `spawn_blocking(move || extract_archive_blocking(...))`. Dropping a `spawn_blocking` handle cannot cancel it, so on budget expiry extraction keeps running and keeps writing to the workspace while its `ScanExtractionPermit` has already been released, which breaks the N × per-archive bound documented at `:1455-1464`. `kill_on_drop` covers subprocesses only. Widening 30 → 120 s widens that window, so the CHANGELOG line about every timeout-reachable extraction using `kill_on_drop(true)` overclaims.

**F4** — `scan_content_expecting` (`:4820`) takes a global extraction permit inside the caller's timeout, and the global cap defaults to 4 (`:1464`). Four concurrent rescans now hold all four for up to 120 s, after which fail-closed inline proxy downloads exhaust their own 30 s budget and return 423. At ≤30 s this was marginal; at 120 s it is deterministic. Cooldown being exactly equal to the budget (`security.rs:1904`) also lets a timed-out rescan immediately re-claim.

**F5** — a scanner that errors in milliseconds now costs a 120 s repository lockout, up from 30. Measured: `RESCAN_SCAN_FAILED` returned in 11.7 ms with `Retry-After: 120`. That is the most common operator-facing failure — an absent or misconfigured scanner, which is #3455's reporter — so it makes the debug loop four times worse.

**F6** — the endpoint now emits two error envelopes: the new 503 uses `{"code", "message"}` and the new tests assert `error` is absent (`:2905`, `:2952`), while the 429 from the same handler still emits `{"error": "RESCAN_THROTTLED", …}` (`:2033`) and its OpenAPI entry declares no body (`:2147`).

**F7** — three of the six `kill_on_drop` hunks have no revert-proof: `trivy_fs_scanner.rs:131`, `incus_scanner.rs:157`, `incus_scanner.rs:201`.

**F8** — `inline_gate_budget_exceeded_arm_fail_open_serves_pending_without_waiting` (`proxy_helpers.rs:7513`) passes in both branches. Its own comment concedes `ServePendingScanAsync` never awaits the scan, so `Serve { pending: true }` holds for a hanging, failing or succeeding scanner alike.

## What I could not break

`Retry-After` is correct in every branch — `retry_after_delay_seconds` (`:2017`) maps 0 → 1 and 7.2 → 8, saturates on `Duration::new(u64::MAX, 1)`, and never emits 0 or negative; live runs gave 111, 91 and 120 (rounded up from 119.99…), with no header on the arms that release. Slot accounting has no `?` between claim and match, `release_rescan_slot` is reachable only from `NoScanner`, and `HashMap::remove` is idempotent, so there is no leak and no double-release apart from the F2 cancellation case. Error bodies name no scanner internals, paths or config; repo-scoped `find_cached_entry` plus the canonical 404 closes cross-tenant probing, and a private repo 403s a non-member. And the new tests genuinely run: both `--lib` jobs set `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1` after `sqlx migrate run` (`ci.yml:543-565`, `949-982`), none of the new tests is a bin or non-allowlisted integration target, and all 18 new/touched tests pass against a real Postgres — the 12 `rescan_*` taking 32.60 s, which proves the 32 s `MockCveRescan::Hang` actually executes.

Happy to take F1 and F2 on if that is easier than reworking them here — this is a 1.8.2 blocker so we would rather move it along than sit on it.

brandonrc:
Thanks for this — the direction and most of the change hold up well, so we picked it up to land the two blocking review findings (plus a couple of the non-blocking ones) since it is the last 1.8.2 blocker. Your commits are preserved; I rebased onto current `main` (it carries migrations 209 and 211 now) and added one commit on top. Summary of what changed and why:

**F1 — claim moved back ahead of the read.** The cooldown claim is the only bound on how many `PROXY_SCAN_MAX_BYTES` (200 MiB) buffers are in flight, and it had moved behind the `storage.get` + `sha256_hex`, so every *rejected* rescan did a full read + hash just to answer 429. Reproduced on a deployed build: 16 concurrent rescans of one 200 MiB cached object went from **714 MiB peak RSS / 15×429 in 3–17 ms** to **3429 MiB / 16×429 each 0.57–1.0 s** with the claim after the read. The claim now runs before the read again; the unreadable-object path calls the `release_rescan_slot` this PR added, so a missing object still returns the canonical 404 without consuming a slot — the goal your comment described, without exporting the read outside the throttle. (Note the reach: `require_repo_write_access` returns `Ok` on a public repo, so any authenticated user drove this on a public remote.)

**F2 — budget derived from the outer timeout.** `PROXY_RESCAN_BUDGET` (120 s) equalled the default `global_request_timeout_secs` (120 s), which *is* applied to this route and whose clock starts at request arrival — so the outer backstop always won and `RESCAN_BUDGET_EXCEEDED` was unreachable. The budget is now `proxy_rescan_budget(global_request_timeout_secs)`: a quarter-window of headroom reserved for the pre/post-scan work, capped at a 120 s ceiling (90 s at the default), with a compile-time and unit-tested relation. Confirmed reachable live through the full router+timeout stack.

**F5** — a scan that fails in milliseconds no longer costs a 120 s lockout; a failed scan's cooldown is capped at 30 s (the pre-existing window). `NoScanner` still releases outright; `BudgetExceeded` still keeps the full window.

**F6** — the 429 now uses the same `{code, message}` `ErrorResponse` envelope (`RESCAN_THROTTLED`) as the 503 arms, and its OpenAPI entry documents the body.

**Tests.** The ordering-pinning test is rewritten to assert the intended invariant (unreadable object → 404 and no slot left claimed). New coverage: a rejected rescan asserts **zero** storage reads via an instrumented backend (measured, not source-order); `RESCAN_BUDGET_EXCEEDED` is exercised **through the router** so the timeout layer is in play; and the budget/timeout relation, the failure-cooldown cap, and the missing-object release are pinned. Full `--lib` suite green (`AK_TESTS_REQUIRE_DB=1`), clippy `-D warnings` clean, duplication at baseline.

F3/F4/F7/F8 from the review are left as noted (F3/F4 are pre-existing extraction/permit-bound concerns beyond this PR's scope; happy to file follow-ups).

Issue #3498, opened 2026-08-22
Migration worker's cancel branch is unreachable: spawn_migration_worker's CancellationToken is never retained, so real cancels log "Migration paused by user"


brandonrc:
## Summary

The migration worker has a cancel branch that can never execute, and as a result every real user cancel is logged as `"Migration paused by user"`. Cosmetic — the job row is already `cancelled` before the worker gets there — but it makes cancel/pause indistinguishable in logs, which is exactly what you want to distinguish when diagnosing a stopped migration.

## Mechanism

`backend/src/services/migration_worker.rs:721-731`:

```rust
Ok(RepoProcessOutcome::Interrupted) => {
    if self.cancel_token.is_cancelled() {
        tracing::info!(job_id = %job_id, "Migration cancelled by user");
        self.migration_service
            .update_job_status(job_id, MigrationJobStatus::Cancelled)
            .await?;
    } else {
        tracing::info!(job_id = %job_id, "Migration paused by user");
    }
    return Ok(());
}
```

`self.cancel_token` comes from `spawn_migration_worker` (`backend/src/api/handlers/migration.rs:550`):

```rust
let cancel_token = CancellationToken::new();
```

It is constructed fresh, moved into `MigrationWorker::new`, and **nothing retains a handle to it**. There is no registry of running workers, and no production code path calls `.cancel()` on it — the only `.cancel()` in the file is in the test fixture (`migration_worker.rs:5378`, `MockInterrupt::CancelToken`).

The real cancel path is a pure database write: `cancel_migration` (`backend/src/api/handlers/migration.rs:1408-1440`) sets `status = 'cancelled'`, and the worker observes it through `is_paused(job_id)` (`migration_worker.rs:836`), which is what actually produces `RepoProcessOutcome::Interrupted`. Since the token is never cancelled, that path always takes the `else` arm.

The same dead condition appears at `migration_worker.rs:666-672`, the between-repositories check, with the same consequence.

## Impact

- A cancelled job logs `"Migration paused by user"`. Operator-visible and misleading.
- The `update_job_status(Cancelled)` inside the dead branch never runs. Harmless today only because `cancel_migration` already wrote `cancelled` synchronously, but it means the worker contributes nothing to the terminal state and a future refactor that relied on it would be silently wrong.
- The branch is untestable through any production path, so the compiler and the test suite both give it a pass. #3380's new test reaches it only by injecting a token cancel that production cannot produce.

## Related wording bug in the same function

[...]

Issue #2926, opened 2026-07-26
redact_url_for_diagnostics keeps userinfo, leaking upstream_url credentials to clients in 502 bodies


brandonrc:
## Problem

`redact_url_for_diagnostics` (`backend/src/services/proxy_service.rs:377`) strips only the query string and the fragment:

```rust
fn redact_url_for_diagnostics(url: &str) -> String {
    if let Ok(mut parsed) = reqwest::Url::parse(url) {
        parsed.set_query(None);
        parsed.set_fragment(None);
        return parsed.to_string();
    }
    ...
}
```

`Url::to_string()` preserves the userinfo component, so `https://svc:hunter2@nuget.internal.corp/v3/index.json` comes back with `svc:hunter2@` intact.

That redacted string is woven into error messages by `validate_upstream_status` (`backend/src/services/proxy_service.rs:471`), which formats it into `AppError::NotFound`, `AppError::ServiceUnavailable`, and `AppError::BadGateway` for non-2xx upstream responses. The catch-all arm of `map_proxy_error` (`backend/src/api/handlers/proxy_helpers.rs:505`) then returns the error's `Display` output to the client verbatim:

```rust
(
    StatusCode::BAD_GATEWAY,
    format!("Failed to fetch from upstream: {}", e),
)
```

## Impact

Minor, and conditional on operator configuration. An operator who embeds credentials directly in a repository's `upstream_url` has those credentials echoed in a 502 body to any client authorized to download from that repository. The internal upstream hostname is disclosed on the same path regardless of whether credentials are present.

This requires the upstream to return a non-2xx status, which is not attacker-controlled in the general case, though an attacker who can reach the repo can retry until the upstream is transiently unhealthy.

## Failure scenario

1. Operator sets a remote repository's `upstream_url` to `https://svc:hunter2@nuget.internal.corp/v3/index.json`.
2. The upstream returns 403 (expired credential, IP restriction, rate limit, anything non-2xx and non-5xx).
3. `validate_upstream_status` builds `BadGateway("Upstream returned error status 403 Forbidden: https://svc:hunter2@nuget.internal.corp/v3/index.json")`.
4. `map_proxy_error` returns that string as the 502 body to the downloading client.

## Fix

Strip userinfo in `redact_url_for_diagnostics` alongside query and fragment. The codebase already uses this pattern in `backend/src/services/upstream_feed.rs:384`:

```rust
let mut display = parsed.clone();
let _ = display.set_username("");
[...]

PR #3570, opened 2026-08-26
fix(migration): harden Nexus transfers against flaky sources

### LINKED ISSUE #3569: Nexus source: large transfers time out and transient failures are never retried
Migrating from a Nexus source has two failure modes that the Artifactory source does not have. Both live in `backend/src/services/nexus_client.rs`.

### 1. The request timeout is really a body-read deadline

`NexusClient::new` builds the client with `.timeout(Duration::from_secs(config.timeout_secs))`. In reqwest that bound covers reading the response body, not just establishing the request, so any artifact whose download takes longer than `timeout_secs` (default 30) is killed mid-body and surfaces as `error decoding response body`.

The practical effect is that the largest artifact a migration can move depends on the throughput between the two instances rather than on any configured limit, so the same job succeeds or fails run to run. The error message points at deserialization, which sends you looking in the wrong place.

### 2. Transient upstream failures are fatal

Nothing in the Nexus client retries. A source restart, a 502 from a proxy in front of Nexus, or a 429 under load permanently fails whatever item was in flight. `ArtifactoryClient` already has `RetryConfig` for exactly this, so the two source types behave differently for the same interruption.

### Suggested fix

Bound the connect and per-read phases instead of the whole request, so a slow but healthy download survives while a dead connection still fails fast. Then give `NexusClient` the same `RetryConfig` the Artifactory client uses and retry 5xx, 429 (honouring `Retry-After`) and connect/read timeouts with exponential backoff.

Retrying inside `send_authenticated` is enough to cover `get`, `download_artifact` and `download_artifact_stream`, since all three go through it. A body that dies mid-stream should not be re-issued, because the caller is already consuming chunks; `read_timeout` bounds that case and item-level retry belongs to the caller.

nicola-preda:
## Summary

Fixes #3569

Two fixes to `NexusClient`, both scoped to `backend/src/services/nexus_client.rs`.

**Timeouts are now per-phase.** `.timeout()` covered reading the body, so any download slower than `timeout_secs` died as `error decoding response body` and the effective size ceiling depended on throughput. Replaced with `connect_timeout` plus `read_timeout`, so a slow download survives while a stalled one still fails. `timeout_secs` keeps its name and default and now means "how long a read may stall with no bytes arriving"; the new `connect_timeout_secs` defaults to 10.

**Transient failures now retry.** `NexusClient` picks up the same `RetryConfig` the Artifactory client already uses, and `send_authenticated` retries 5xx, 429 (honouring `Retry-After`) and connect/read timeouts with exponential backoff. `get`, `download_artifact` and `download_artifact_stream` all route through it, so one change covers every caller. A 404 stays a definitive answer and is not retried. A body that dies mid-stream is deliberately not re-issued, since the caller is already consuming chunks.

Both new config fields land behind `Default`, and the only external construction site (`api/handlers/migration.rs`) already spreads `..Default::default()`, so nothing else changes.

Three new tests in `nexus_client::tests`, all against a `wiremock` server so they need no database:
- `test_download_artifact_retries_transient_5xx` - 503 twice then 200, which is what a source restart looks like mid-migration. Fails on `main`, where the first 503 is returned to the caller.
- `test_download_artifact_gives_up_after_max_retries` - the 503 still surfaces once the budget is spent, so retry cannot mask a genuinely broken source.
- `test_download_artifact_does_not_retry_404` - asserts exactly one upstream request, pinning the "definitive answer" boundary.

`test_nexus_config_default` also now asserts `max_retries > 0`, so the retry path cannot be silently defaulted off.

## Test Checklist
- [ ] Manually tested locally

`cargo test --lib nexus_client` is 31 passed / 0 failed, and `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` are clean.

## API Changes
- [x] N/A - no API changes


brandonrc:
## Audit round 1 — Nexus transfer hardening (2 reviewers: correctness / security+integration)

Both passes read head against `origin/main`, merged main into a throwaway branch (**clean, zero conflicts; `cargo check --lib --tests` exit 0**; `nexus_client.rs`/`artifactory_client.rs` untouched on main since the merge base), and ran the client tests under the compile lock (31/31 pass). The fix is the right shape and closes both halves of #3569 exactly as the issue prescribed; the retry is genuinely idempotent (it returns before any body byte is read, so no double write and no immutability conflict is reachable, and the worker's write is an upsert anyway). Credentials are **not** re-sent across redirects — verified in vendored reqwest 0.12.28, `redirect.rs:239-252` strips `AUTHORIZATION` on any host/port change. Three small changes turn this into a merge.

### Should-fix (both reviewers found the first two independently)

1. **`Retry-After` is trusted unclamped, and the sleep is not cancellable.** `nexus_client.rs:200-209` uses `retry_after.map(|s| s * 1000)` directly; `retry.max_delay_ms` (30 s default) bounds only the fallback backoff (`:244-247`), never the header value. A rate-limiting proxy answering `429 Retry-After: 3600` parks the worker for an hour, three times over with default `max_retries: 3`; `Retry-After: 18446744073709551615` overflows the multiply (panic in debug, wrap in release). The operator cannot escape it: pause/cancel is checked only *between* artifacts (`migration_worker.rs:990-993`) and the sleep at `:242` isn't `select!`-ed against the token. One line: `s.saturating_mul(1000).min(retry.max_delay_ms)`. (Same unclamped multiply exists on main at `artifactory_client.rs:395` — happy to take that as a follow-up rather than widen this PR.)

2. **Dropping `.timeout()` removes the only bound on total request duration** — correct for the streaming path the production migration uses, but `get()` (`:262`, `response.json`) and `download_artifact` (`:521`, `response.bytes()`) still buffer unbounded, and now nothing bounds their total time. `read_timeout` resets on every successful read, so a source delivering ≥1 byte per 30 s holds the worker and a growing allocation forever. Either keep a generous total `.timeout()` on those two call sites or cap the bodies; the streaming path is fine as-is.

3. **The two new WARN lines log the source URL verbatim** (`nexus_client.rs:226`, `:235`, `url = %url`). The URL is `source_connections.url`, and `validate_outbound_url` (`validation.rs:343-360`) neither rejects nor strips userinfo, so `https://svc:hunter2@nexus.corp/` is accepted and the first 503 writes the password to the log. Main's `nexus_client.rs` logs no URLs, so this is a regression. The repo already has `redact_url_for_diagnostics` (#2926) for this.

### Nits
- Doc comment at `:177-179` says "every request goes through here", but `ping()` (`:280-289`) builds its own request and bypasses `send_authenticated` — the connection test still fails on the first blip.
- No jitter; `throttle_delay_ms` isn't applied between retries.
- Tests: a mutation run (retry loop short-circuited to main's behaviour) shows only `test_download_artifact_retries_transient_5xx` catches the bug; the give-up and no-retry-404 tests pass either way (they pin boundaries, which is fine — just know that's what they do). Nothing covers the timeout half or the 429 path.

### Not this PR's problem
- #3654 (Nexus raw migration doubles the repo name into the path) is in `migration_artifact_path` at `migration_worker.rs:2977-2994`, not in the client; this PR is right not to touch it.
- The `CancellationToken` arm being dead in production (`api/handlers/migration.rs:550`/`:572`) is #3498, already on the milestone.

### Bookkeeping
No `## [Unreleased]` bullet for #3569 — needed before the cut. I can push a `docs(changelog)` commit once the three fixes are in.

brandonrc:
Checking in on this one. The audit for the Nexus transfer hardening has been sitting since 2 September with no reply, and 1.9.0 shipped yesterday — I would like to get this into 1.10.0 rather than let it age further.

brandonrc:
Taking this over as option 2, per my note last week — @nicola-preda, thank you, the design here is yours and it is unchanged. The connect/read timeout split, the retry classification, the "only the header phase is retried" boundary and the three original tests are your commit (`fbf22031`), which stays first in the history and in the release notes. I pushed two commits on top rather than rebasing or squashing, so your authorship is intact and you can still push back on any of this.

## What I changed, per audit item

**1. `Retry-After` clamp + cancellable backoff** (`nexus_client.rs`). `retry_after.map(|s| s * 1000)` is now `s.saturating_mul(1000).min(retry.max_delay_ms)` — the same 30 s ceiling the fallback backoff already obeyed, and no overflow on an absurd header. The backoff sleep is now a `tokio::select!` against the migration's `CancellationToken` (the shape `cache_invalidation.rs:312` and `upstream_feed.rs:241` already use); on cancel it stops retrying and lets the last response surface, rather than sleeping out the full wait. The token is threaded from the job that owns the client: `create_source_client` takes an `Option<&CancellationToken>`, `spawn_migration_worker` now receives the token instead of minting its own, and the three callers with no job behind them (test-connection, list-repositories, assessment) pass `None`. That wiring is inert until #3498 makes the token live in production, but the knob is real rather than dead.

**2. Total-duration bound restored on the buffered paths only.** `send_authenticated` takes a `total_timeout: Option<Duration>`; `get()` and `download_artifact()` pass `buffered_timeout_secs` (new, default 300 s), `download_artifact_stream()` passes `None` and stays on `connect_timeout` + `read_timeout`. So a source dribbling one byte per read can no longer hold a worker and a growing allocation forever, and a large streamed artifact is still allowed to take as long as it takes. `test_streaming_download_is_not_bounded_by_the_total_timeout` pins that the streaming path did *not* pick the ceiling up.

**3. Source URL no longer logged verbatim.** Both WARN lines (and the new cancel line) log `redact_url_for_diagnostics(&url)`, computed once outside the loop. The helper is promoted from `fn` to `pub(crate) fn` in `proxy_service.rs`; nothing else about it changed.

**4. Nits.** Corrected the `send_authenticated` doc comment instead of routing `ping()` through it — a connection test should report what the source is doing right now, not spend the retry budget before answering; the comment now says so explicitly. **Left out:** jitter and `throttle_delay_ms` between retries. Jitter is not a two-line change here — the only jitter in the tree is `apply_jitter`/`deterministic_jitter_seed`, private to `webhooks.rs`, and lifting them into a shared helper is a wider diff than this PR should carry. Applying `throttle_delay_ms` on top of the backoff would push the wait past the `max_delay_ms` clamp item 1 just added. Both are happy follow-ups. The unclamped multiply at `artifactory_client.rs:395` is untouched, as you offered.

**5. CHANGELOG.** One `### Fixed` bullet under `## [Unreleased]`, first ref `#3569`, covering both halves of the issue plus the four changes above.

## Verification

All under the compile lock, `SQLX_OFFLINE=true`:

```
cargo fmt --check                                                    clean
cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings  clean
cargo check  -p artifact-keeper-backend --lib --tests                 exit 0
cargo nextest run -p artifact-keeper-backend --lib -E 'test(nexus)'
    Summary [2.236s] 52 tests run: 52 passed, 15949 skipped
scripts/ci/check-changelog-unreleased.sh                              exit 0
```

New tests, each verified to fail with only its own fix reverted:

| reverted | fails |
|---|---|
| clamp → `s * 1000` | `test_retry_after_is_clamped_to_max_delay`, `test_retry_after_does_not_overflow` |
| `select!` → plain `sleep` | `test_retry_backoff_is_abandoned_when_migration_is_cancelled` |
| drop the buffered `.timeout()` | `test_buffered_download_is_bounded_by_a_total_timeout` |
| `url = %url` | `test_retry_warnings_redact_source_credentials` |

The redaction test captures WARN output through a scoped `tracing_subscriber` writer and asserts the password never reaches it.

**Your mutation, re-run** (retry loop short-circuited to a single send, everything else intact) — was caught by `test_download_artifact_retries_transient_5xx` alone; now caught by four:

```
FAIL test_download_artifact_retries_transient_5xx
FAIL test_retry_after_is_clamped_to_max_delay
FAIL test_retry_warnings_redact_source_credentials
FAIL test_retry_after_does_not_overflow
Summary 37 tests run: 33 passed, 4 failed
```

`test_download_artifact_gives_up_after_max_retries` and `test_download_artifact_does_not_retry_404` still pass under it, as your audit said — they pin boundaries, which is the right job for them.

## One thing to know before merging

The branch is still based on `d8c497e8`, i.e. before 1.9.0 was cut, so `CHANGELOG.md` now conflicts with `main` — `main` opened a `### Added` in `[Unreleased]` at the same anchor. `git merge-tree` says that is the *only* conflict; all three source files auto-merge. Resolution is to keep both blocks, `### Added` from `main` first and this `### Fixed` after it. I left it rather than pushing a merge or a rebase into @nicola-preda's branch.

I will merge this for 1.10.0 once CI is green.

PR #3572, opened 2026-08-26
fix(proxy): serve stale when upstream refuses revalidation

### LINKED ISSUE #3571: Proxy revalidation treats a throttled or refused upstream as "content changed"

```rust
status => {
    Ok(true)
}
```

### Suggested fix

nicola-preda:
## Summary

Fixes #3571

`check_etag_changed`'s catch-all arm answered "changed" for 403, 429 and 5xx on the revalidation HEAD. Those are the upstream declining to answer, not an answer, so the result was a discarded-but-valid cache entry plus a refill attempt against an upstream that had just refused a far cheaper request. On a throttling upstream that makes every stale-entry read compound the throttling.

New predicate `revalidation_status_carries_no_content_information` classifies those three, and the probe returns an error for them instead of `Ok(true)`. That routes them into the stale-if-error path `revalidate_verdict` already implements, so the cached copy is served for up to `STALE_IF_ERROR_GRACE_SECS` (1 hour) past expiry and falls through to a refill after that. Staleness stays bounded by the existing window; this PR adds no new knob and no new unbounded-stale path.

Deliberately excluded:
- **401** drives the OCI bearer-token re-fetch and must keep meaning "retry with a token".
- **404 / 410** are real statements about the resource and keep their current handling.

Scope worth noting for review: `check_etag_changed` has exactly one production caller, `revalidate_verdict`. The other caller, `check_upstream`, is `pub` but is only reached from tests. So the behaviour change is confined to one path.

The one judgement call here is **403**. Treating it as "no information" means that if a repository's upstream credentials are revoked, reads are served from cache for up to the 1 hour grace window before the failure surfaces, rather than failing immediately. That seemed clearly better than the current behaviour, which evicts the cache entry and then fails the download anyway, but it is a policy choice and I am happy to narrow the predicate to 429 and 5xx only if you would rather 403 stayed fatal.

`test_revalidation_status_carries_no_content_information` pins both directions of the classifier: 403/429/500/502/503/504 are refusals, and 401/404/410/200/304 must keep their existing handling. The second half is the load-bearing one, since the risk in this change is over-reaching and swallowing a status that legitimately means "re-fetch".

## Test Checklist
- [ ] Manually tested locally

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/services/proxy_service.rs
+++ b/backend/src/services/proxy_service.rs
@@ -747,6 +747,23 @@ fn validate_upstream_status(status: StatusCode, url: &str) -> Result<()> {
     Ok(())
 }
 
+/// Whether a revalidation response status says anything about whether the
+/// cached content changed.
+///
+/// A throttle (429), a refusal (403) or an upstream fault (5xx) is the upstream
+/// declining to answer the question, not answering "yes". Treating them as
+/// "changed" discards a valid cached entry and spends a full refill on an
+/// upstream that is already refusing us, which is strictly worse than serving
+/// the copy we hold: the refill cannot succeed where the probe was refused.
+///
+/// 401 is deliberately excluded, the OCI bearer-token exchange depends on it
+/// meaning "re-fetch with a token".
+fn revalidation_status_carries_no_content_information(status: StatusCode) -> bool {
+    status == StatusCode::FORBIDDEN
+        || status == StatusCode::TOO_MANY_REQUESTS
+        || status.is_server_error()
+}
+
 /// Whether a single path segment is a dot segment, in the same sense the WHATWG
 /// URL parser uses when it normalizes a path.
 ///
@@ -3096,6 +3113,18 @@ impl UpstreamClient {
                     url
                 );
                 Ok(true)
+            }
+            status if revalidation_status_carries_no_content_information(status) => {
+                tracing::warn!(
+                    "Upstream returned {} on the ETag check for {}; serving the cached copy instead of refilling",
+                    status,
+                    redact_url_for_diagnostics(url)
+                );
+                Err(AppError::ServiceUnavailable(format!(
+                    "Upstream returned {} on revalidation: {}",
+                    status,
+                    redact_url_for_diagnostics(url)
+                )))
             }
             status => {
                 tracing::warn!(
@@ -12425,6 +12454,38 @@ mod tests {
                 );
             }
             other => panic!("403 must map to redacted AppError::BadGateway; got {other:?}"),
+        }
+    }
+
+    #[test]
+    fn test_revalidation_status_carries_no_content_information() {
+        for status in [
+            StatusCode::FORBIDDEN,
+            StatusCode::TOO_MANY_REQUESTS,
+            StatusCode::INTERNAL_SERVER_ERROR,
+            StatusCode::BAD_GATEWAY,
+            StatusCode::SERVICE_UNAVAILABLE,
+            StatusCode::GATEWAY_TIMEOUT,
+        ] {
+            assert!(
+                revalidation_status_carries_no_content_information(status),
+                "{status} is a refusal to answer, not a 'yes, it changed'"
+            );
+        }
+
+        // 401 drives the OCI bearer-token re-fetch, and 404/410 are real
+        // statements about the resource. None of them may take the stale path.
+        for status in [
+            StatusCode::UNAUTHORIZED,
+            StatusCode::NOT_FOUND,
+            StatusCode::GONE,
+            StatusCode::OK,
+            StatusCode::NOT_MODIFIED,
+        ] {
+            assert!(
+                !revalidation_status_carries_no_content_information(status),
+                "{status} must keep its existing handling"
+            );
         }
     }
 


nicola-preda:
Coverage job went red on `policy_service::tests::test_ungraded_finding_blocked_by_high_policy_3306`, which this PR doesn't go anywhere near (the diff is one predicate plus a match arm in `proxy_service.rs`).

Looks like a flake rather than anything here: it passes locally for me, the Backend Unit Tests job passed the same test on this same commit, and the coverage job on #3570 passed it off the same base. Note the failure was in the "Generate coverage report" step, so the actual coverage gates were skipped, not breached.

Could you kick off a re-run when you get a chance? I don't have the rights to do it from my side.

brandonrc:
## Audit round 1 — stale-serve on refused revalidation (2 reviewers: correctness / security+integration)

Both passes read the PR head `da77d1e8` against `origin/main`, merged main into a throwaway branch (**clean, zero conflicts; `cargo check --lib --tests` exit 0** — none of the 18 commits touches `proxy_service.rs`), and ran the revalidation tests against a live Postgres. The mechanism is right and security-neutral-to-positive; two things to change and two framings to correct before this lands.

**What checks out.** The new arm sits below `NOT_MODIFIED`/`OK`/`UNAUTHORIZED` and above the catch-all (`proxy_service.rs:3117`), feeds `revalidate_verdict`'s existing `stale-if-error` branch (`:6140-6156`), and deliberately does not re-stamp `expires_at`, so the grace window stays anchored on the original expiry. 404/410 still fall through to refill + negative cache; 401 is untouched for the OCI bearer re-fetch. No scan-gate, quarantine, curation or visibility check is bypassed — the `ServeStaleIfError` body is quarantine-gated identically to a fresh hit (`:5973-5975`, `:4284-4288`), and it moves the serve onto a *gated* path from the ungated refill fallback. There is exactly one upstream conditional-revalidation probe in the backend, so no other format carries a copy of this bug.

### Should-fix

1. **Drop 403 from the set (you already offered — please take it).** 429 and 5xx describe the upstream's *state* and are method-agnostic; 403 is also returned per-request — a method-filtering WAF/CDN rule, a signed URL or object-store policy that grants GET but not HEAD. On such an upstream a `Mutable` npm packument (5-min TTL) that main refreshes within 5 minutes is now frozen for 65: every read repeats the same 403 probe and serves the same bytes, and `expires_at` is never advanced. The revoked-credentials trade you describe in the PR body is a cost main does not avoid either — main's refill path *already* serves stale on 403 via its own `stale-if-error` arm (`origin/main` `proxy_service.rs:3860-3874`), with no grace bound at all. So 403 buys nothing there and costs freshness here.

2. **The added test cannot catch the bug — proved, not argued.** With the new match arm deleted (main's exact behaviour), `test_revalidation_status_carries_no_content_information` still passes; it pins the predicate table, not the wiring. A behaviour test is cheap on the existing `revalidate_verdict` + wiremock harness (`test_revalidate_verdict_replays_accept_of_cached_variant_3290`, `:18430`): a HEAD returning 429 against a sidecar expired 600 s ago must yield `ServeStaleIfError`; expired 7200 s ago must yield `Refill`; 401/404/410 must still `Refill`. The reviewer wrote those three, all pass on your head, and the first one fails with the arm removed.

### Framing corrections (for the CHANGELOG, not the code)

- **Nothing is evicted.** `write_negative_cache` fires only for `NotFound` (`:3883-3888`); a 403/429/5xx on the refill GET maps to `BadGateway`/`ServiceUnavailable` and lands on the stale-serve arm. Clients were never failing on main. The real bug is **upstream request amplification** — one HEAD plus one full-body GET per stale read against a throttling upstream — and this PR halves it (every read in the grace hour still issues the HEAD, since `expires_at` is correctly not re-stamped; a refusal backoff would need a *new* sidecar field).
- **Docker Hub, which the issue leads with, is not reached.** `check_etag_changed` does no OCI bearer exchange (`exchange_bearer_then` is called only from the fetch paths, `:2454`/`:2674`); the probe gets **401** and exits at the pre-existing 401 arm before the new predicate is consulted. The 429 arrives on the refill, which is unchanged. Fixed for Maven Central / npm / PyPI / Basic-auth'd upstreams; OCI is byte-for-byte as before.

### Nits
- New arm returns `ServiceUnavailable` for 403/429 where `validate_upstream_status` maps non-404 4xx to `BadGateway` (`:727-748`) — harmless today, inconsistent if `check_upstream` ever gains a production caller.
- `pub fn check_upstream` changes contract (403/429/5xx now `Err`); no production callers, but worth a doc line.

### Bookkeeping
No `## [Unreleased]` bullet for #3571 — the release preflight blocks a cut without one. Happy to push a `docs(changelog)` commit to the branch once the 403 change lands, or paste it in if you'd rather own it.

brandonrc:
Heads-up: #3661 (gilesw) landed a second fix for #3571 this afternoon with the wiremock behaviour tests the review above asked for, and a CHANGELOG bullet — but it also includes 403 (and 408). I've asked there to drop both for the reasons in finding 1 above. Two fixes for one bug won't both merge; whichever of the two carries the narrowed predicate first is the one I'll take, with the bullet crediting both. Your call whether to rebase this onto that harness or let #3661 carry it.

brandonrc:
Apology and a correction here too.

While this PR was open on #3571 I ran a fix wave over the remaining 1.9.0 issues without checking for contributor PRs already in flight, and merged my own fix (#3747, commit `429192ba`). That was my process failure.

And the review you got on this PR was right in a way my merge was not: I merged a predicate covering 429, 403, 408 and 5xx, when the finding here argued 403 and 408 do not belong — they are per-request answers, so an upstream that grants GET but refuses HEAD leaves a mutable entry frozen for the grace window instead of refreshing, and the refill path already serves stale on a failed GET anyway. A PR narrowing the merged set to 429 and 5xx is open now, crediting this review and #3661.

Closing as superseded. The diagnosis in this PR — that the probe was mapping "the upstream declined to answer" onto "the content changed", and spending a full GET on an upstream that had just refused a cheap HEAD — is what got fixed, and it came from you.

brandonrc:
Follow-up as promised: #3756 is open and narrows the merged set to 429 and 5xx only. 403 and 408 go back to the refill path, with a test that pins the case the broad predicate broke — probe 403, refill 200, new bytes served and the sidecar validator updated — and one that keeps the degradation, probe 403 with refill 403 serving stale. The reasoning is credited to this review by number in the commit and the PR body.

PR #3574, opened 2026-08-26
fix(rpm): walk virtual members on the upstream catch-all

### LINKED ISSUE #3573: Virtual RPM repository over Remote members serves an empty index, dnf installs nothing
A Virtual RPM repository over Remote members cannot serve any real yum layout. `dnf` reports the repository as empty and installs nothing, with no error to explain it.

### Why

RPM has three download routes. `/packages/…` and `/repodata/…` are handled specifically, and everything else falls through to the `upstream_proxy` catch-all. That catch-all serves only Remote repositories:

```rust
// A normal (non-`@N`) request only serves from Remote repos.
if repo.repo_type != RepositoryType::Remote {
    return Err((StatusCode::NOT_FOUND, "Not found").into_response());
}
```

Real yum layouts put packages under a prefix, not under `/packages/`:

- `9-stream/BaseOS/x86_64/os/Packages/…`
- `elrepo/el9/x86_64/…`
- `rpm/<arch>/…`

All of those reach the catch-all, so on a Virtual repo they 404 even when a member holds the artifact.

### Why it's silent rather than a visible 404

The 404 on the package is the second-order problem. Before fetching packages, the client fetches `/repodata/`, which the Virtual repo *does* answer, generating the document from the `artifacts` table. That table never holds proxy-cached content (#1278), so a Virtual repo over Remote members generates a valid, empty repodata document. `dnf` is told the repository legitimately contains nothing, exits 0, and installs nothing.

This is the same class of failure as #1447, which was fixed for Remote repositories. The Virtual case on this route was missed.

### Note

`download_package` already walks Virtual members on the `/packages/` route, and the debian handler already handles Virtual on its own catch-all paths. RPM's `upstream_proxy` is the outlier, so the fix is to make it match the sibling that already works rather than to introduce a new pattern.

nicola-preda:
## Summary

Fixes #3573

`upstream_proxy` (the RPM catch-all for any path that is not `/packages/` or `/repodata/`) served Remote repositories only, so a Virtual repo 404'd every real yum layout: `9-stream/BaseOS/x86_64/os/…`, `elrepo/el9/x86_64/…`, `rpm/<arch>/…`. Combined with generated repodata rendering from the `artifacts` table, which never holds proxy-cached content (#1278), the client is handed a valid empty index and installs nothing without erroring.

The fix walks Virtual members on this route, using the same `try_remote_or_virtual_download` call `download_package` already makes on `/packages/`, with identical options (`VirtualLookup::PathSuffix`, `suppress_upstream_proxy: false`). `PathSuffix` is the documented convention for filename-keyed handlers including rpm, and it only affects Local members, since Remote members resolve by proxy-fetching `upstream_path`.

No new pattern is introduced. `download_package` in this same file already does this, and the debian handler already handles Virtual on its catch-all paths; RPM's catch-all was the outlier.

Placement: the Virtual branch goes after the existing `@N` pinned-member handling so that path is untouched, and before the `repo_type != Remote` rejection it replaces for the Virtual case. Local and Remote behaviour on this route is unchanged.

`test_rpm_virtual_upstream_path_resolves_from_later_member` builds a Virtual repo over two Local members and stores the artifact in the **second** one, at a nested path (`el/9/x86_64/deep-2.0-1.x86_64.rpm`) that reaches the catch-all rather than `/packages/`. Seeding the second member is deliberate: a fix that only consulted the first member would still pass a single-member test.

Verified in both directions against a live Postgres:
- on this branch, passes
- with the new branch condition neutralized, fails with `404 != 200` at the documented assertion

## Test Checklist
- [x] Manually tested locally

`cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` are clean. The test is DB-backed, so it was run with `AK_TESTS_REQUIRE_DB=1` to ensure it could not pass by skipping.

## API Changes
- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/rpm.rs
+++ b/backend/src/api/handlers/rpm.rs
@@ -1288,7 +1288,9 @@ async fn repodata_proxy(
 
 async fn upstream_proxy(
     State(state): State<SharedState>,
+    Extension(auth): Extension<Option<AuthExtension>>,
     Path((repo_key, upstream_path)): Path<(String, String)>,
+    ctx: crate::api::middleware::download_telemetry::DownloadContext,
 ) -> Result<Response, Response> {
     let repo = resolve_rpm_repo(&state.db, &repo_key).await?;
 
@@ -1331,12 +1333,45 @@ async fn upstream_proxy(
         return Err((StatusCode::NOT_FOUND, "Not found").into_response());
     }
 
-    // A normal (non-`@N`) request only serves from Remote repos.
+    let filename = upstream_path.rsplit('/').next().unwrap_or(&upstream_path);
+
+    // Virtual repos walk their members here, same as `download_package` does for
+    // `/packages/`. Without this a group over Remote members 404s every real yum
+    // layout (`9-stream/BaseOS/…`, `elrepo/el9/…`, `rpm/<arch>/…`), because those
+    // paths reach this catch-all rather than the `/packages/` or `/repodata/`
+    // routes above. Worse than the 404: the generated `/repodata/` document a
+    // client fetches first renders from the `artifacts` table, which never holds
+    // proxy-cached content (#1278), so dnf is told the repository is empty and
+    // silently does nothing — the #1447 failure, fixed for Remote and missed here.
+    //
+    // `PathSuffix` matches this file's convention (the Remote arm below and
+    // `download_package` both key on the filename); it only affects Local members,
+    // since Remote members are resolved by proxy-fetching `upstream_path`.
+    if repo.repo_type == RepositoryType::Virtual {
+        return match proxy_helpers::try_remote_or_virtual_download(
+            &state,
+            auth.as_ref(),
+            &repo,
+            &ctx,
+            proxy_helpers::DownloadResponseOpts {
+                upstream_path: &upstream_path,
+                virtual_lookup: proxy_helpers::VirtualLookup::PathSuffix(filename),
+                default_content_type: "application/octet-stream",
+                content_disposition_filename: None,
+                suppress_upstream_proxy: false,
+            },
+        )
+        .await?
+        {
+            Some(resp) => Ok(resp),
+            None => Err((StatusCode::NOT_FOUND, "Not found").into_response()),
+        };
+    }
+
+    // A normal (non-`@N`) request otherwise only serves from Remote repos.
     if repo.repo_type != RepositoryType::Remote {
         return Err((StatusCode::NOT_FOUND, "Not found").into_response());
     }
-
-    let filename = upstream_path.rsplit('/').next().unwrap_or(&upstream_path);
 
     // Cache hit by filename: serve the local copy.
     if let Some(hit) =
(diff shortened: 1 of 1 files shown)


brandonrc:
## Audit round 1 — RPM virtual catch-all (2 reviewers: correctness / security+integration)

Both passes read head `bfdc093c`, merged `origin/main` into a throwaway branch (**clean, zero conflicts, `cargo check --lib --tests` exit 0**; no main commit touches `rpm.rs`), and ran the RPM tests against a live Postgres (8/8 pass). The code change is correct and reuses the right shared helper — the design questions all resolved in your favour. The substantive issue is scope, not code.

**What checks out.** `try_remote_or_virtual_download` is the canonical resolver; priority order, entitlement narrowing (members are filtered *before* the walk, so there is no 403 to fall through), and 404/5xx-continue vs quarantine-terminate semantics all come from it. `PathSuffix` over `ExactPath` is correct because Local rpm members store at `packages/<filename>`. **Security verified live, not just read:** a temporary probe test — anonymous caller, public Virtual with a *private* Local member holding the artifact — returned 404 with no member bytes (`authorize_virtual_members` at `proxy_helpers.rs:2729`, the #3525 unified 404). Path handling: Remote fetches go through `validate_cache_path` (`proxy_service.rs:5783`, rejects `..`, NUL, backslash and percent-encoded dot segments); Local members only see the trailing filename via an escaped LIKE. The added test is well built (artifact seeded in the *second* member) and the reviewer confirmed it fails without the fix (`404 != 200` at `rpm.rs:3833`).

### Should-fix

1. **`Fixes #3573` over-claims — please make it `Part of #3573`.** The issue's headline is "*serves an empty index, dnf installs nothing*" and says explicitly the package 404 is the second-order problem. This PR fixes the second-order problem. The top-level `/repodata/*` routes (`rpm.rs:48-61`) bypass the catch-all entirely; for a Virtual they generate from `collect_repodata_artifacts` (`rpm.rs:702-736`), which reads `artifacts` rows — and proxy-cached bytes are never written there (#1278; `rpm.rs:248-254` already documents this as an open follow-up). So `dnf` against a Virtual-over-Remote still sees 0 packages and never issues the package request this branch serves. It narrows further: AK's own generated `primary.xml` always advertises `packages/<file>` (`rpm.rs:1815-1825`, pinned at `:2619`/`:2636`), so the catch-all is only reached for externally constructed mirror-layout URLs. That is a real fix — just not the one the issue title names. The comment the PR *adds* at `rpm.rs:1342-1345` describes the empty-repodata defect directly above a branch that doesn't address it; reword or move it.

2. **The one test uses two Local members.** `test_rpm_virtual_upstream_path_resolves_from_later_member` (`rpm.rs:3783-3838`) only exercises the `PathSuffix` → `local_fetch_by_path_suffix` branch; the Remote-member branch — the issue's topology — is untested, and the default fixture wires no `proxy_service`. The file already has the wiremock machinery (`rpm.rs:3172-3210`). A negative authz test (private member, anonymous caller → 404) would match what every prior virtual-member fix pinned.

3. **The #3446 recording gate for `rpm.rs` is now vacuous.** Its predicate (`proxy_helpers.rs:16657`) is function-scoped: `upstream_proxy`'s body now contains `try_remote_or_virtual_download(` (a listed recorder), so the Remote arm's still-unrecorded serve at `rpm.rs:1432` passes on the recorder rather than on its `UNRECORDED-PROXY-SERVE` marker — which now reads stale and invites deletion. Worth a one-line note or a narrower predicate.

### Nits
- "identical options" isn't accurate: the new branch passes `application/octet-stream` / `content_disposition_filename: None`, where `download_package`'s Virtual arm (`rpm.rs:1463-1466`) and this handler's own Remote arm (`:1432-1442`) use `application/x-rpm` / `Some(filename)`.
- The `None => Err(404)` arm at `rpm.rs:1367` is unreachable (the Virtual arm always returns `Ok(Some(_))`, `proxy_helpers.rs:5143`) and would emit `text/plain` rather than the unified JSON envelope.

### Bookkeeping
No `## [Unreleased]` bullet for #3573/#3574 — needed before the cut. I can push a `docs(changelog)` commit once the scope wording is settled.

brandonrc:
Apology first: while this PR was open awaiting the round-1 revisions, I ran a fix wave over the remaining 1.9.0 issues and did not check which of them already had a contributor PR in flight. #3573 was one, and I merged my own fix for it (#3745, commit `524f4fd9`) this evening. That was my process failure, not a judgement on your work.

What landed is the shape your PR established: the RPM catch-all resolves a Virtual through `proxy_helpers::try_remote_or_virtual_download` with the same `DownloadResponseOpts` `download_package` uses, first member wins, caller-authorized walk. The audit added two documentation points on top — the helper records a hosted-member serve but not a Remote-member serve through a Virtual (#1278), and a prefixed `repodata/repomd.xml` is first-member-wins rather than merged, so a Virtual over two Remotes exposes only the higher-priority member's index to `dnf`.

Closing as superseded. Your diagnosis of the three RPM download routes and the Remote-only gate was exactly right and is what made the fix a five-line change.

PR #3576, opened 2026-08-26
feat(peers): expose replication_mode on the assigned-repos list

### LINKED ISSUE #3575: Peer subscriptions list returns only repository IDs, so replication_mode is unavailable
`GET /api/v1/peers/{id}/repositories` returns only repository IDs:

```json
["3f2a...", "9c81..."]
```

The sibling endpoint for a single subscription, `GET /api/v1/peers/{id}/repositories/{repo_id}`, already returns the full `SubscriptionResponse` with `replication_mode`, `sync_enabled`, `replication_schedule`, `replication_filter` and `last_replicated_at`.

So the list view of a collection carries strictly less information than the item view of the same collection, and the field that matters most for display, `replication_mode`, is only reachable one repository at a time.

### Impact

A replication dashboard listing a peer's subscriptions cannot tell which direction any of them replicates. The options are to issue one extra request per repository just to learn the mode, or to display a placeholder. Ours defaults the display to `pull`, which is wrong for every `push` and `mirror` subscription and silently so.

### Suggested fix

Have the list endpoint return `Vec<SubscriptionResponse>`, the same type the item endpoint already returns. No new schema type is needed, and the list becomes a list of the thing the item endpoint serves.

This does change the response shape of an existing endpoint, so it is a breaking change for any client reading the bare-UUID array. Given that `SubscriptionResponse` already exists and is already served by the neighbouring route, the alternative of a parallel `?detail=true` parameter or a second endpoint seemed worse than making the two routes agree, but that is a maintainer call.

nicola-preda:
## Summary

Fixes #3575

`GET /api/v1/peers/{id}/repositories` returned `Vec<Uuid>`, while the neighbouring `GET /api/v1/peers/{id}/repositories/{repo_id}` already returns a full `SubscriptionResponse`. The list view of a collection therefore carried less information than the item view, and `replication_mode` was only reachable one repository at a time. A dashboard listing a peer's subscriptions cannot show replication direction without N extra requests.

This returns `Vec<SubscriptionResponse>` from the list endpoint, backed by a new `PeerInstanceService::list_subscriptions`. No new schema type is introduced; it reuses the one the item endpoint already serves, so the OpenAPI surface gains no new component.

`list_subscriptions` uses a runtime `query_as` rather than the `query!` macro, so it needs no offline SQLx cache entry (same approach as `principal_must_change_password` in `api/middleware/auth.rs`). Ordered by `created_at` for a stable list.

**This is a breaking response-shape change** for any client reading the bare-UUID array, so it wants a coordinated SDK regen and frontend update. If you would rather not break it, I am happy to rework this as a `?detail=true` parameter or a separate route instead. Making the two neighbouring routes agree seemed cleaner than adding a third shape, but it is your call and I would rather ask than assume.

The second commit is a small cleanup rather than part of the fix: mapping the row into `SubscriptionResponse` was written out field-by-field in both handlers, so this replaces both with one `From` impl. It removes 22 duplicated lines, keeps the changed file under the duplication gate, and leaves `get_subscription` a one-liner.

## Test Checklist
- [x] Manually tested locally

`list_subscriptions_returns_rows_with_replication_mode` covers the new service method against a live database. Worth noting explicitly: the `test_pool()` helper in that module returns `None` and skips when it cannot connect, so a green run does not by itself prove the test executed. I verified it does by making the query return no rows and confirming the test then fails, rather than trusting the pass.

Also run: the 55 `peers` handler tests, and `cargo test --lib test_openapi_spec_is_valid` since this changes a documented response body. `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` are clean.

## API Changes
- [ ] N/A - no API changes

The `#[utoipa::path]` `responses(...)` entry for the list endpoint is updated from `body = Vec<Uuid>` to `body = Vec<SubscriptionResponse>`.


nicola-preda:
Pushed a fix for the coverage gate: new-code coverage was 68%, the `From` impl's field assignments weren't exercised by anything. Added two pure unit tests for the mapping (no DB), which took it to 91% (110/120). Both coverage gates pass now.

The duplication gate is still red at 4.3% and I don't think I should be the one to fix it. All five clones it reports pre-date this PR:

```
peer_instance_service.rs:269-285 <-> 240-256
peer_instance_service.rs:298-314 <-> 240-256
peer_instance_service.rs:334-349 <-> 241-256
peer_instance_service.rs:796-808 <-> 601-613
peers.rs:404-421 <-> 358-375
```

Four of them are the `query_as!(PeerInstance, "SELECT <24 columns> FROM peer_instances WHERE ...")` blocks that differ only in their WHERE clause, and the fifth is the `rows_affected() == 0 -> NotFound` pattern. The gate prints its own caveat about this (it measures whole changed files, #1619), and my `From` impl actually removed 22 duplicated lines from `peers.rs`; it still reads 4.3% because of what was already there.

I'd rather not deduplicate those `query_as!` blocks inside this PR: doing it properly means either parameterizing a macro string literal or dropping to runtime `query_as` and losing the compile-time SQL checking, which is a real trade-off that deserves its own review and doesn't belong bundled with an API change.

Happy to do it as a separate PR if you want it, or to leave it. Let me know how you'd prefer to handle the gate here.

brandonrc:
## Audit round 1 — `replication_mode` on the assigned-repos list (2 reviewers: correctness+API / security+integration)

Both passes read the head, merged `origin/main` into a throwaway branch (**clean, zero conflicts; `cargo check --lib --tests` exit 0**; nothing on main has touched peers/mesh since the merge base), and ran the peer tests against a live Postgres (14/14, and the DB-backed one genuinely executed and confirmed `Some("push")`, so the runtime `query_as` decodes every column). The field itself is right: read from the authoritative `peer_repo_subscriptions` row via the same projection the item endpoint uses, serialised in the same lowercase casing as the PG enum and every other reader, `SubscriptionResponse` untouched and already registered in `components(schemas(...))`. It closes #3575. Two pieces of collateral ride along undeclared, both one-line fixes.

### Should-fix (both reviewers found these independently)

1. **`scripts/mesh-e2e/test-sync-policy.sh` is broken by the shape change and not updated.** At `:128-141` it does `jq '[.[] | select(. == $repo_id)] | length'` on what it comments as "a flat array of repository UUIDs". With objects that comparison is always false → `SUB_COUNT=0` → `fail "peer-b subscription not found"` on a correctly working system. It is the only in-tree consumer of the endpoint. It didn't gate this PR because `mesh-e2e.yml` is dispatch-only. Fix: `select(.repository_id == $repo_id)`.

2. **The `sync_enabled = true` filter is silently dropped.** The replaced `get_assigned_repositories` (`peer_instance_service.rs:614-616`) filtered on it; the new `list_subscriptions` (`:635-648`) does not, so the endpoint widens from "repos this peer syncs" to "all subscription rows" — `{"sync_enabled": false}` is a supported assignment (`peers.rs:626`), and every other reader of that table keeps the filter (`:819`, `:849`, `peer_service.rs:334`, `maven.rs:2398`). Defensible either way, but it's a second breaking change mentioned nowhere. Restore it, or document the widening in the PR body and the OpenAPI description.

3. `get_assigned_repositories` is now dead code with divergent semantics (`pub`, so no warning). Remove it or route the new list through it. Consider `query_as!` — the byte-identical sibling `get_subscription` (`:517-532`) gets compile-time column checking and this one doesn't.

### The red `📊 Code Coverage` job is not this PR's fault
Both coverage gates pass; the failing step is the **jscpd duplication gate** (4.3% vs 3% threshold), and both reviewers reproduced it with CI's flags: PR head **4.3%**, `origin/main` on the same two files **4.4%**, the identical five pre-existing clones (the 20-column `peer_instances` projection ×4, the `rows_affected()==0 → NotFound` tail, the `PeerInstanceResponse` literal). Your `From`-impl refactor actually improves the ratio. A rerun won't change it; it needs a de-dup of those clones, a knowing merge, or an exemption. I'll take the knowing-merge route once the two should-fixes land, since the gate is measuring main's debt, not yours.

### Nits
- New DB test spliced under the doc comment belonging to `assign_repository_with_missing_repo_returns_not_found`.
- Tests assert only `push`; nothing covers the handler's response shape or the row-set change.
- `ORDER BY created_at` has no tiebreaker.

### Bookkeeping
No `## [Unreleased]` bullet — needed before the cut, and given the response-shape change it belongs under `### Changed` with a breaking-change note. Not a slice of #1619 (that's the jscpd-scan CI issue, cited only to explain the red gate). I'll milestone this 1.9.0 and push the `docs(changelog)` commit once the two fixes are in.

brandonrc:
On the duplication gate: leave it. `📊 Code Coverage` is not a required check, and jscpd measures whole changed *files*, not changed lines — so touching one line inside `peer_instance_service.rs` re-reports every pre-existing `query_as!` clone in it. #3711 hit exactly this and merged over it. Your change removing 22 duplicated lines while the percentage goes up is the gate working as designed and reading backwards.

Please don't spend a PR on the dedupe either: both ways out cost something real (a macro string literal loses compile-time SQL checking, which is most of why `query_as!` is there), and #1619 already tracks the gate measuring changed files rather than changed lines. If you want a follow-up worth having, that issue is the one.

brandonrc:
## Audit round 1 addressed — pushed on top of your branch (`220a7b42` → `5f6b3a3b`)

@nicola-preda, thanks for this — the design call was the right one and the implementation was sound; the list and item views agreeing is exactly what I wanted, and your `From` impl is a genuine improvement to `peers.rs`. You haven't been able to get back to the audit, and I want this in 1.10.0, so I've finished the three should-fix items myself rather than let it sit. Your three commits are untouched and the work stays yours; mine sit on top.

### 1. `scripts/mesh-e2e/test-sync-policy.sh` — updated to the new shape

`select(. == $repo_id)` → `select(.repository_id == $repo_id)`, plus the two comments that described the response as a flat UUID array.

**I could not run the mesh harness.** It targets `backend-peer-a`/`backend-peer-b` in the ArgoCD-managed `ak-mesh-*` cluster, and `mesh-e2e.yml` is dispatch-only against `ak-e2e-runners`; there is no local equivalent. What I did execute is the jq itself, against a synthetic response in the new shape:

```
OLD jq vs NEW payload  : 0     <- the bug: fail "peer-b subscription not found" on a healthy mesh
NEW jq vs NEW payload  : 1
NEW jq vs no-match     : 0
NEW jq vs empty []     : 0     <- the curl-failure fallback still degrades correctly
```

`sh -n` is clean and shellcheck reports nothing new (the SC2015/SC2034 notes it prints are all pre-existing lines).

### 2. `sync_enabled = true` filter — restored, pinned by a DB test

Restored on `list_subscriptions`, so the endpoint keeps meaning "repos this peer syncs" rather than widening to all subscription rows. New DB-backed test `list_subscriptions_omits_sync_disabled_rows` assigns one repo with `sync_enabled = true` and one with `false` to the same peer and asserts only the enabled one is listed.

Fails without the fix — I reverted just the predicate and re-ran:

```
FAILED services::peer_instance_service::tests::list_subscriptions_omits_sync_disabled_rows
panicked at peer_instance_service.rs:1309:
  sync_enabled = false subscription must not be listed, got [581a3ac8-…, bcd30a58-…]
```

Sync policies insert with `sync_enabled = true` (`sync_policy_service.rs:646,769,1022`), so the restored filter doesn't change what the mesh script above sees.

### 3. `get_assigned_repositories` — removed

Removed rather than routed through. Routing would mean the new list mapping whole rows down to ids and then widening them again, which is the larger diff for no gain, and the function had no callers left. Nothing else in the tree referenced it; `cargo check --lib --tests` confirms.

I left the runtime `query_as` as you wrote it. `query_as!` would give compile-time column checking, but this repo carries **two** committed offline caches (`.sqlx/` and `backend/.sqlx/`, 464 and 462 entries), and `cargo sqlx prepare` risks churning both for one query. Your precedent (`principal_must_change_password`) is real and there is no `sqlx prepare --check` gate, so this is fine as is.

### Nits
- `ORDER BY created_at, id` — deterministic tiebreaker.
- The new DB test no longer sits under the doc comment belonging to `assign_repository_with_missing_repo_returns_not_found`; both now have their own.
- OpenAPI response description says "Sync-enabled subscriptions…" so the row set is documented, not just the shape.

### Bookkeeping
Merged `origin/main` in (1.9.0 cut on the 8th, so your branch predated it) and added the `## [Unreleased]` → `### Changed` bullet, first reference **#3575**, with an upgrade note for the breaking response shape. Kept in Keep-a-Changelog order ahead of the existing `### Security` section.

### Verified
`cargo fmt --all --check` clean · `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` clean · `cargo check -p artifact-keeper-backend --lib --tests` exit 0 · `cargo nextest run --lib -E 'test(peer) or test(subscription) or test(replication) or test(openapi)'` against a live Postgres with migrations through 214: **393 passed, 0 failed**, including `test_openapi_spec_is_valid` and both DB-backed `list_subscriptions_*` tests (they genuinely executed — 0.13s/0.19s against the pool, not the ~0.02s of the skip path).

The red `📊 Code Coverage` job stays red and that is fine: it is not a required check and, as on #3711, jscpd is re-reporting `peer_instance_service.rs`'s pre-existing `query_as!` clones because the file was touched. Nothing for you to do there.

I'll merge once CI is green. Thanks again for the PR and for the careful write-up on the coverage gate.

Issue #3537, opened 2026-08-24
Release bookkeeping is unverified: 4 of the last 8 releases shipped with wrong or missing notes/CHANGELOG state


brandonrc:
Four of the last eight releases shipped with wrong release bookkeeping. Every
one was found by a human, days later; none was found by the pipeline.

| Release | What went wrong | Found |
|---|---|---|
| **v1.7.1** | No `.github/release-notes/1.7.1.md` existed on the tag. `release.yml`'s "Resolve release notes" step silently fell back to `generate_release_notes: true`. Auto-notes do not promote the CHANGELOG, so 8 `[Unreleased]` entries that had already shipped stayed under `[Unreleased]` — including an upgrade note telling 1.7.1 operators to act "before upgrading to 1.7.2" when they had already been exposed for a week. | 6 days later (#3318) |
| **post-v1.7.3** | `bb11616e` merged ~5h *after* the tag and filed its bullet under `## [1.7.3]`. The published notes claim a fix that release does not contain. | later |
| **v1.7.5 prep** | `7aab4b79` renamed `## [Unreleased]` → `## [1.7.5]` without opening a fresh `[Unreleased]`. Four later PRs anchored on the old heading and merged conflict-free into the released section: 29–35 bullets of 1.8.0 work filed under `[1.7.5]`. | #3433 |
| **v1.8.1** | The two scanner-CVE fixes that unblocked the publish were filed under `[Unreleased]` rather than `[1.8.1]`. | later |

**Why nothing caught it.** The only automated check is `release.yml`'s
"Verify CHANGELOG.md documents this release", which asserts exactly two
things: a `## [X.Y.Z]` heading exists, and it has at least one non-empty
line under it. `artifact-keeper-test`'s `version-set-integrity` runs a
byte-identical awk block and falls back to *main's* CHANGELOG when the
tag's copy cannot be read. Neither can detect an entry filed in the wrong
section, a missing entry, or a missing notes file. #3433 added
`check-changelog-unreleased.sh`, which closes the v1.7.5 shape at PR time
only — and that script does not exist on `release/1.7.x`, the branch three
of the last six releases were cut from.

### Proposed

1. **Hard-fail a stable `vX.Y.Z` tag with no curated notes file.** Keep the
   `generate_release_notes` fallback for `-rc`/prerelease tags only. This
   alone removes the v1.7.1 mechanism.
2. **Port `check-changelog-unreleased.sh` (and its self-test) to
[...]

PR #3578, opened 2026-08-26
fix(analytics): count proxy-cached objects in storage stats

### LINKED ISSUE #3577: Analytics storage breakdown and trend exclude proxy-cached objects, so remotes read as empty
Proxy-cached objects carry no `artifacts` row (#1280). Both analytics reads aggregate `artifacts`, so on an instance with pull-through remotes they exclude every proxy-cached object:

- `GET /analytics/storage/breakdown` reports a remote repository's `artifact_count` and `storage_bytes` as 0 no matter how much it has cached. A remote serving terabytes reads as empty, and because the breakdown is ordered by `storage_bytes`, it also sorts last.
- `GET /analytics/storage/trend` reads `storage_metrics`, whose daily snapshot counts only hosted objects, so the storage trend under-reports both object count and bytes, and `total_downloads` misses every proxy serve.

The inconsistency is visible inside the product: `/admin/stats` already reports the proxy half separately (#3088), so the dashboard and the analytics page disagree about the same storage.

### Suggested fix

Keep the existing `artifact_count` / `storage_bytes` / `total_*` fields hosted-only so their meaning does not change for existing consumers, and add the proxy half beside them as sibling fields. That mirrors the split `/admin/stats` already exposes.

`storage_metrics` needs three columns for the snapshot case, since the figure has to be captured at snapshot time rather than derived later. Historic rows can be partly reconstructed: `proxy_cache_artifacts.cached_at` and `proxy_download_statistics.downloaded_at` carry their own timestamps, so the cumulative figure as of each snapshot date is recoverable for anything still on record. That is best-effort by construction, since an object since evicted or a serve since pruned is gone, so reconstructed rows are lower bounds rather than what a live capture would have written.

nicola-preda:
## Summary

Fixes #3577

Both analytics reads aggregated only `artifacts` rows, and proxy-cached objects have no `artifacts` row (#1280), so a remote repository serving terabytes reported 0 artifacts and 0 bytes. `/admin/stats` already reports the proxy half separately (#3088), so the two views disagreed about the same storage.

`GET /analytics/storage/breakdown` now carries `proxy_artifact_count` and `proxy_storage_bytes` beside the existing `proxy_download_count`, and `storage_metrics` records the same three halves per snapshot (migration 210) for `GET /analytics/storage/trend`.

**Additive, not a behaviour change for existing consumers.** The hosted `artifact_count` / `storage_bytes` / `total_*` fields keep their exact current meaning and the proxy half rides alongside as new fields, mirroring the split `/admin/stats` already exposes. Both structs default the new fields to 0 when deserializing an older payload.

The breakdown is now ordered by hosted + proxy bytes, so a remote repository no longer sorts last regardless of how much it holds. That is the one ordering change.

Migration 212 reconstructs the proxy halves of historic snapshots from `proxy_cache_artifacts.cached_at` and `proxy_download_statistics.downloaded_at`, rather than leaving every pre-existing row reporting zero remote usage. It is best-effort by construction: an object since evicted, or a serve dropped by download retention, is no longer on record, so a reconstructed row is a lower bound rather than what a live capture would have written. It only touches rows still entirely at 0, so it cannot overwrite a figure captured live, and re-running is safe. Cost is one aggregating pass per proxy table plus a small scan per snapshot row; on an install with long retention and a busy proxy, treat it as seconds-to-minutes.

Both migrations are additive with no data loss, so they are reversible by dropping the three columns.

`test_analytics_reads_surface_proxy_download_count_db` covers the breakdown, and `test_capture_daily_snapshot_includes_oci_blob_and_proxy_cache_bytes` covers the snapshot. Both are DB-backed and assert the proxy halves are populated where the old code reported 0.

One limitation worth flagging rather than leaving for you to find: `test_storage_metrics_proxy_backfill_fills_historic_rows_db` verifies migration 212's logic by re-executing a **copy** of its statement scoped to a single snapshot date, because the migration itself has already run by the time tests execute. So it proves the reconstruction is correct, but it would not catch the migration file and the test copy drifting apart. I did not find a better option that stays a unit test; happy to take a suggestion.

## Test Checklist
- [x] Manually tested locally

20 `analytics_service` tests pass against a live Postgres with both migrations applied. `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` are clean. All new queries are runtime `sqlx::query`, not the macro, so no `.sqlx` cache entry is needed.

## API Changes
- [ ] N/A - no API changes

No new endpoints; two existing response structs gain fields.


brandonrc:
⚠️ **Migration number collision with another open PR — this will not show up as a conflict.**

Two open PRs currently claim migration **209**:

- **#3578** — `backend/migrations/209_storage_metrics_proxy_columns.sql` (plus `210_storage_metrics_proxy_backfill.sql`)
- **#3549** — `backend/migrations/209_oidc_device_sessions.sql`

`main` currently tops out at `208_backups_payload_checksum.sql`.

Because the filenames differ, git sees no conflict. Both PRs will show `mergeable: CLEAN`, both will pass CI independently, and both will merge without complaint. The failure appears only **after the second one lands**: sqlx records one version `209` in `_sqlx_migrations`, encounters a second file claiming the same version, and the backend **aborts at startup**. On `main`.

Whoever merges second needs to renumber before merging. Given #3578 also uses 210, the clean split is:

- #3578 keeps 209 + 210, or moves to 210 + 211
- #3549 moves to whichever remains free

Note **#3599** already claims **211**, so please avoid it — checked and confirmed against the current ledger.

Worth doing the check by hand rather than trusting CI here: `gh pr view <n> --json files -q '.files[].path' | grep migrations` across the open PRs is the whole check, and it is the only thing that catches this before it reaches `main`.

brandonrc:
Update: the 1.8.2 blockers have now taken their numbers, so the free range has moved.

Currently claimed against `main` (which tops at `208_backups_payload_checksum.sql`):

- **209** — #3604 `209_scan_results_pin_identity.sql` (1.8.2 blocker, merging first)
- **211** — #3599 `211_packages_version_oci_tag_length.sql` (1.8.2 blocker)
- **209** — #3578 `209_storage_metrics_proxy_columns.sql` + **210** `210_storage_metrics_proxy_backfill.sql`
- **209** — #3549 `209_oidc_device_sessions.sql`

So three PRs hold 209. The 1.8.2 pair merges first, which means **#3578 and #3549 both need to renumber before they land**.

Free after 1.8.2: **210**, and **212** onward.

Suggested split — #3578 takes 210 + 212 (it needs two), #3549 takes 213. A gap at 212 or elsewhere is harmless; the ledger already has several.

Restating why this is worth a manual check: the filenames differ, so git reports no conflict, both PRs show `mergeable: CLEAN`, and CI passes on each independently. The failure only appears after the second one merges, when sqlx finds two files claiming one version and the backend aborts at startup — on `main`.

brandonrc:
## Audit round 1 — proxy-cached bytes in storage stats (2 reviewers: correctness / security+integration+perf)

Both passes read head `e29120f2`, merged `origin/main` into a throwaway branch (**clean, zero conflicts; `cargo check --lib --tests` exit 0** — every added statement is a static literal, so the sqlx 0.9 `SqlSafeStr` guard is satisfied and no `.sqlx` entry is missing), applied all four migrations in order on a fresh PG16, and ran the analytics tests against it (20/20, both new DB-backed tests genuinely executing). The `/analytics/storage/breakdown` half is correct and fixes the reported symptom. The `storage_metrics` half ships a duplicate figure under a wrong doc contract, and two of the new queries have measured performance cliffs. Nothing here is architectural.

### Blocker

**`StorageSnapshot.proxy_storage_bytes` is documented as a disjoint half of `total_storage_bytes`, but it is a subset already inside it.** `capture_daily_snapshot` sums `repository_usage_ledger (hosted_bytes + proxy_bytes + oci_bytes)` (`analytics_service.rs:173-174`, from #3134/#3249), and migration 182 defines `proxy_bytes` as exactly `SUM(proxy_cache_artifacts.size_bytes)`. Reproduced on a seeded DB by both reviewers: one 1 MB cached object → `total_storage_bytes = 1000000` **and** `proxy_storage_bytes = 1000000`. The new doc at `:43-47` ("the `total_*` figures stay hosted-only") contradicts the unchanged doc at `:30-39` ("**includes** … proxy-cached bytes") and `/admin/stats`, whose identically named field is documented at `admin.rs:1048-1050` as "a **breakdown** of `total_storage_bytes` (**do not add the two**)"; main's own CHANGELOG (#3249) says "clients that added the two fields must stop doing so". The repo's pre-existing test (`analytics_service.rs:1201`, `EXPECTED_TOTAL_DELTA` at `:1224`) already asserts `CACHED` is a term in the total — and still passes. Fix is the doc comment and the CHANGELOG text, not the query: the field is a breakdown, same as `/admin/stats`. (On `/storage/breakdown` the same field name *is* disjoint from `storage_bytes`, so one doc comment can't cover both structs.)

### Should-fix

1. **The breakdown double-counts legacy `proxy-cache/%` `artifacts` rows, including in the new sort key.** The hosted join at `:332` has no `storage_key NOT LIKE 'proxy-cache/%'` predicate, which migration 182 (`:104`, `:111`) and the struct's own `total_artifacts` doc (`:24-28`) both require. On an instance upgraded from pre-#1280: `artifact_count 1 / storage_bytes 1000000 / proxy_artifact_count 1 / proxy_storage_bytes 1000000`, sort key 2 MB, while the ledger says `hosted 0, proxy 1000000`.

2. **Migration 212's `serve_daily` CTE is referenced once, so Postgres inlines it and re-aggregates the whole `proxy_download_statistics` table once per snapshot row.** Measured on PG16, 100k serves × 365 snapshots: **4,711 ms** with `Seq Scan … loops=365`; with `MATERIALIZED` on both CTEs: **68 ms**. It scales as snapshots × serves, there is no retention job on that table, and it runs at boot inside one `UPDATE`.

3. **The new `LEFT JOIN LATERAL` sits after the `artifacts` fan-out** (`:331-340`), so the per-repo aggregate is evaluated once per (repo, artifact) pair and only survives because PG14+ inserts a Memoize node. Measured with 200k artifacts: as written **117 ms**; `enable_memoize=off` **113,775 ms**; LATERAL moved above the artifacts join **71 ms with `loops=20`** and no Memoize dependence.

4. **Migration 212 reconstructs from `cached_at`, which is "last written", not "first cached"** — `proxy_catalog.rs:70` resets it to `now()` in the upsert's conflict arm, so every TTL revalidation and mutable-index refresh moves it. Reproduced: reconstruction for a 2024-06-01 snapshot = 1000000 before a refresh, **0** after. Still a usable lower bound, but the migration header and CHANGELOG attribute the shortfall only to eviction; say what it actually is, and give operators the one-line note that this full-table aggregate runs at startup.

5. **The CHANGELOG bullet is filed inside `## [1.8.2]`** (`CHANGELOG.md:176`, the section released 2026-09-01), not `## [Unreleased]`, and carries no `#NNNN`. That is exactly the "bullet under an already-released heading" shape the preflight gates after #3537 — it will report #3578 as undocumented and block the cut. Move it up and cite #3577.

### Verified fine
Scoped `proxy-cache/<scope>/…` keys (#3609) are counted (both aggregates key on `repository_id`, not the prefix); eviction is symmetric (`delete_by_key` + the 182 delete trigger); migration 210 is a gap-fill below the already-shipped 211 and sqlx applies it out of order without error on a simulated v1.8.2 database; no object-store enumeration is introduced (pure SQL — #3593's S3 LIST fragility isn't in this path); both endpoints are admin-gated.

### Bookkeeping
Once the doc/CHANGELOG text is corrected and the two `MATERIALIZED` / join-order changes are in, this is a merge. I'll milestone it 1.9.0 now.

brandonrc:
Migration renumbering looks right: 210 for the columns, 212 for the backfill, 213 left clear for #3549, and applying the full ledger against a fresh database is the check that matters. Thanks for doing that rather than assuming. Nothing blocking from my side; it stays on 1.10.0 with #3577.

brandonrc:
Correction to my note above, and it is my fault: the numbering advice I gave you on 8 September is now stale.

`212` was free when I wrote it. It no longer is — `212_jupyter_format.sql` merged to main today (#3786), and `214_artifact_metadata_pypi_classifiers_gin.sql` merged with it (#3788). Main's migration directory now ends `209, 211, 212, 214`, with `210` and `213` free.

For this PR that means:

- `210_storage_metrics_proxy_columns.sql` — still fine, `210` is genuinely free.
- `212_storage_metrics_proxy_backfill.sql` — **needs renumbering**. `scripts/ci/check-duplicate-migrations.sh` runs on the merge commit, so it goes red the moment this lands on top of main.

`215` is the safe target: `213` is claimed by #3549 (OIDC device grant), which is also in flight. Nothing else in the file needs to change — the ledger keys on the filename, not the number's neighbours.

Happy to push the rename myself if you would rather not do another round trip; say the word and I will do exactly that one file rename and nothing else. Sorry for sending you down the wrong number.

brandonrc:
## Taking this over — audit round 1 addressed, authorship intact

@nicola-preda hasn't been able to pick this back up since the audit, and it is wanted for 1.10.0, so I've finished it rather than let it sit. **The work is theirs**: `e29120f2` is untouched at the base of the branch and stays the substance of this PR — the breakdown half was already correct, both DB-backed tests were real, and the PR description flagged the test/migration drift risk before anyone else had to find it. My three commits sit on top; nothing was rewritten or squashed.

New head: **`b5858749`** (`e29120f2` → `709003dd` renumber → `174c8f61` review fixes → `b5858749` merge of `main`).

### A. Migration renumbering (`709003dd`)

`212_storage_metrics_proxy_backfill.sql` → **`215_storage_metrics_proxy_backfill.sql`**, since `212_jupyter_format.sql` (#3786) and `214_artifact_metadata_pypi_classifiers_gin.sql` (#3788) took 212/214 on main. `213` left alone for #3549. `210_storage_metrics_proxy_columns.sql` keeps its number. The old number is updated everywhere it is referenced: migration 210's header, the DB-backed test that re-executes the statement, and the CHANGELOG bullet.

```
$ MIGRATIONS_DIR=<union of origin/main and this branch> scripts/ci/check-duplicate-migrations.sh
::warning::gap(s) in migration version sequence (... 213): usually a renumber on one side of a merge...
migration versions: 197 unique, no duplicates
exit=0
```

Same result run on the real merge result after `b5858749`: `209, 210, 211, 212, 214, 215`, exit 0, `213` the only in-range gap.

### Blocker — `StorageSnapshot.proxy_storage_bytes` is a breakdown, not a disjoint half

Confirmed exactly as reported, and the query is not what's wrong. `capture_daily_snapshot` sums `repository_usage_ledger (hosted_bytes + proxy_bytes + oci_bytes)` and migration 182 defines `proxy_bytes` as `SUM(proxy_cache_artifacts.size_bytes)`, so the cached bytes were already inside `total_storage_bytes`:

```
     key     | hosted_bytes | proxy_bytes
-------------+--------------+-------------
 demo-hosted |      6000002 |           0
 demo-remote |            0 |     6000001
```

What changed is the contract text — per field, because the three halves do not behave alike:

* `proxy_storage_bytes` — **a breakdown of `total_storage_bytes`, do not add the two**, same wording as `admin.rs`'s identically named field (#3134/#3249).
* `proxy_artifact_count` / `proxy_download_count` — genuinely disjoint, because `total_artifacts` and `total_downloads` read only `artifacts` / `download_statistics`.
* Noted explicitly that on `RepositoryStorageBreakdown` the same field name **is** disjoint from that row's `storage_bytes`, since one comment can't cover both structs.

Updated in the struct doc, migration 210's header, the three `COMMENT ON COLUMN` texts (so `\d+ storage_metrics` agrees), and the CHANGELOG. To keep it from drifting back into prose-only, `test_capture_daily_snapshot_includes_oci_blob_and_proxy_cache_bytes` now also asserts the seeded `CACHED` bytes show up in **both** the `total_storage_bytes` delta and the `proxy_storage_bytes` delta. Honest note: that assertion pins the corrected contract, it does not fail on `e29120f2` — the code there was already right, only the doc was wrong.

### 1. Breakdown double-counted legacy `proxy-cache/%` `artifacts` rows

Added the `AND a.storage_key NOT LIKE 'proxy-cache/%'` guard to the hosted join — the same predicate `repository_usage_ledger.hosted_bytes` has used since migration 182 — which fixes the count, the bytes, the sort key and `last_upload_at` in one place.

One cached object, carried both ways (catalog row + legacy `artifacts` row) next to a hosted repo holding genuinely more:

```
--- BEFORE ---
 rank |     key     | artifact_count | storage_bytes | proxy_artifact_count | proxy_storage_bytes | sort_key
    1 | demo-remote |              1 |       6000001 |                    1 |             6000001 | 12000002
    2 | demo-hosted |              1 |       6000002 |                    0 |                   0 |  6000002

--- AFTER ---
    1 | demo-hosted |              1 |       6000002 |                    0 |                   0 |  6000002
    2 | demo-remote |              0 |             0 |                    1 |             6000001 |  6000001
```

New DB test **`test_storage_breakdown_excludes_legacy_proxy_cache_artifact_rows_db`**, which asserts both the figures and the ordering. With the guard removed it fails:

```
assertion `left == right` failed: the remote repository holds ONE object: hosted count/bytes
must be 0 and the legacy proxy-cache/% artifacts row must not be added to the catalog's
  left: (1, 6000001, 1, 6000001)
 right: (0, 0, 1, 6000001)
```

### 2. Backfill CTE inlined and re-aggregated per snapshot row

Both CTEs are now `AS MATERIALIZED`, and the test's copy of the statement was changed with it so the two stay byte-equivalent. `EXPLAIN (ANALYZE, BUFFERS)` on PG16, 100k serves × 365 snapshots (the whole `UPDATE`, in a rolled-back transaction):

```
BEFORE
 Update on storage_metrics sm  (actual time=4592.915..4592.917 rows=0 loops=1)
   Buffers: shared hit=308405
   ->  Seq Scan on storage_metrics sm  (actual time=167.139..4578.574 rows=365 loops=1)
         ->  HashAggregate  (actual time=12.044..12.055 rows=183 loops=365)
               ->  Seq Scan on proxy_download_statistics  (actual time=0.002..8.623 rows=50141 loops=365)
 Execution Time: 4612.263 ms

AFTER
 Update on storage_metrics sm  (actual time=43.220..43.221 rows=0 loops=1)
   Buffers: shared hit=5342
   CTE serve_daily
     ->  HashAggregate  (actual time=16.413..16.433 rows=365 loops=1)
           ->  Seq Scan on proxy_download_statistics  (actual time=0.002..8.757 rows=100000 loops=1)
 Execution Time: 43.340 ms
```

**4612 ms → 43 ms**, `loops=365` → `loops=1`, 308k → 5.3k shared buffers.

### 3. `LEFT JOIN LATERAL` below the `artifacts` fan-out

Moved above the `artifacts` join. 200k artifacts across 20 repositories:

```
BEFORE                          131.528 ms   ->  Memoize (actual time=0.000..0.000 rows=1 loops=200000)
BEFORE, enable_memoize=off     4258.507 ms   ->  Aggregate (actual time=0.019..0.019 rows=1 loops=200000)
AFTER                            76.967 ms   ->  Aggregate (actual time=0.022..0.022 rows=1 loops=20)   [no Memoize node]
AFTER, enable_memoize=off        77.317 ms   ->  identical plan
```

So: **131 ms → 77 ms** on the good path, and the 32× cliff behind the Memoize node is gone rather than merely hidden — the plan no longer contains a Memoize node at all.

### 4. `cached_at` is "last written", not "first cached"

Reproduced against `proxy_catalog::upsert`'s conflict arm — reconstruction for a 2025-07-01 snapshot, before and after a single revalidation:

```
 reconstructed_bytes | reconstructed_objects
             6000001 |                     1     <- before
                   0 |                     0     <- after UPDATE ... SET cached_at = now()
```

The query keeps its shape (there is no first-cached timestamp to reconstruct from), so what changed is the claim. Migration 215's header now states the shortfall has **two** independent causes — eviction/retention pruning, *and* `cached_at` moving forward on every TTL revalidation or mutable-index refresh, which walks an object out of earlier snapshots it was in fact occupying — and says plainly that both only ever under-report. The operator note is there too: this is a full-table aggregate over both proxy tables, running at startup inside the migration transaction, budget seconds to minutes on a busy long-retention install. The CHANGELOG carries the same as an upgrade note, and the `COMMENT ON COLUMN` texts now say reconstructed rows are lower bounds.

### 5. CHANGELOG bullet

Moved out of the released `## [1.8.2]` into a fresh `### Fixed` under `## [Unreleased]`, citing **#3577** first. `scripts/ci/check-changelog-unreleased.sh` passes.

`b5858749` also merges current `main`, because the 1.9.0 cut turned this into a conflict: main promoted `## [Unreleased]` to `## [1.9.0]` while this branch was moving its bullet into `## [Unreleased]`. Resolved by keeping both, with the bullet in a `### Fixed` between main's `### Added` and `### Security` (Keep-a-Changelog order). Nothing else conflicted — `analytics_service.rs` auto-merged. The PR reports `MERGEABLE` again.

### Verified

Everything below on the merged tree at `b5858749`, PG16 in Docker, migrations applied in order from empty (`209, 210, 211, 212, 214, 215`):

```
cargo fmt --check                                                        clean
cargo check -p artifact-keeper-backend --lib --tests                     ok
cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings     clean
cargo nextest run -p artifact-keeper-backend --lib \
  -E 'test(analytics) or test(openapi)'                     75 passed, 0 failed
```

All four DB-backed tests genuinely execute (`AK_TESTS_REQUIRE_DB=1`), including `test_openapi_spec_is_valid`. Counterfactual run for the new test is quoted under item 1.

Not run: the full workspace suite and CI's shell-test job beyond the two migration/CHANGELOG gates I invoked directly.

Not merging — over to you.

PR #3598, opened 2026-08-31
fix(ansible): serve upstream collection versions from a Remote repository

### LINKED ISSUE #3365: Ansible remote repository don't show versions collection upstream
### Description

I create an Ansible remote repository.

When I try to install collections from this new repository it doesn't get the collection from upstream repository.

### Environment Information

- Deployment type: docker compose
- Operating System: Debian 13
- Architecture: amd64
- Database: Postgresql 18 on vm
- Reverse Proxy: Caddy
- Network setup: Behind corporate proxy

### Backend Version

1.7.4

### Web UI Version

1.7.0

### Package Format

Other / WASM Plugin

### Expected Behavior

```
ansible-galaxy collection install -r requirements.yml --force -v
Using /etc/ansible/ansible.cfg as config file
Starting galaxy collection install process
Process install dependency map
Starting collection install process
Downloading https://galaxy.ansible.com/api/v3/plugin/ansible/content/published/collections/artifacts/ansible-posix-1.5.1.tar.gz to /home/user/.ansible/tmp/ansible-local-321159fqa7avs/tmp6jh4cxad/ansible-posix-1.5.1-4patjk1e
Installing 'ansible.posix:1.5.1' to '/home/user/.ansible/collections/ansible_collections/ansible/posix'
ansible.posix:1.5.1 was installed successfully
```

### Observed Behavior

```
ansible-galaxy collection install -r requirements.yml --force
Starting galaxy collection install process
Process install dependency map
ERROR! Failed to resolve the requested dependencies map. Could not satisfy the following requirements:
* ansible.posix:1.5.1 (direct request)
Hint: Pre-releases hosted on Galaxy or Automation Hub are not installed by default unless a specific version is requested. To enable pre-releases globally, use --pre.
```

### Steps to Reproduce

1. create file requirements.yml :
```
collections:
  - name: ansible.posix
    version: "1.5.1"
    source: https://repos.ak.org/ansible/ansible_galaxy_v3
```
2. create ansible remote repository
```
ak repo show ansible_galaxy_v3
Key:          ansible_galaxy_v3
Name:         ansible_galaxy_v3
Format:       ansible
Type:         remote
Upstream URL: https://galaxy.ansible.com/
Public:       yes
Description:  
Storage Used: 0 B
Quota:        unlimited
Created:      2026-08-14 06:07:28 UTC
Updated:      2026-08-14 06:07:28 UTC
```
3. try install collection from requirement file :
ansible-galaxy collection install -r requirements.yml --force -v

### Logs

```shell
artifact-keeper-backend  | 2026-08-14T06:44:44.976573Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3}: tower_http::trace::on_request: started processing request
artifact-keeper-backend  | 2026-08-14T06:44:44.978645Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3 correlation_id=78d25dfa-39b1-4682-b118-9ca00ffeb5c3}: sqlx::query: summary="SELECT id, format::text as …" db.statement="\n\nSELECT id, format::text as format, repo_type::text as repo_type, upstream_url, storage_backend, storage_path, is_public, (SELECT value FROM repository_config WHERE repository_id = repositories.id AND key = 'index_upstream_url') AS index_upstream_url FROM repositories WHERE key = $1\n" rows_affected=1 rows_returned=1 elapsed=1.402648ms elapsed_secs=0.001402648
artifact-keeper-backend  | 2026-08-14T06:44:44.978717Z  INFO http_request{method=GET uri=/ansible/ansible_galaxy_v3 correlation_id=78d25dfa-39b1-4682-b118-9ca00ffeb5c3}: artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=78d25dfa-39b1-4682-b118-9ca00ffeb5c3 status=404
artifact-keeper-backend  | 2026-08-14T06:44:44.978759Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3 correlation_id=78d25dfa-39b1-4682-b118-9ca00ffeb5c3}: tower_http::trace::on_response: finished processing request latency=2 ms status=404
artifact-keeper-backend  | 2026-08-14T06:44:45.048069Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3/api}: tower_http::trace::on_request: started processing request
artifact-keeper-backend  | 2026-08-14T06:44:45.049636Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3/api correlation_id=c0ec6acd-8727-42f5-8acc-46b6e67f59ac}: sqlx::query: summary="SELECT id, key, storage_backend, …" db.statement="\n\nSELECT id, key, storage_backend, storage_path, format::text as format, repo_type::text as repo_type, upstream_url, promotion_only, age_gate_enabled, age_gate_min_age_days, age_gate_mode, curation_enabled, curation_default_action FROM repositories WHERE key = $1\n" rows_affected=1 rows_returned=1 elapsed=977.76µs elapsed_secs=0.00097776
artifact-keeper-backend  | 2026-08-14T06:44:45.049746Z  INFO http_request{method=GET uri=/ansible/ansible_galaxy_v3/api correlation_id=c0ec6acd-8727-42f5-8acc-46b6e67f59ac}: artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=c0ec6acd-8727-42f5-8acc-46b6e67f59ac status=200
artifact-keeper-backend  | 2026-08-14T06:44:45.049811Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3/api correlation_id=c0ec6acd-8727-42f5-8acc-46b6e67f59ac}: tower_http::trace::on_response: finished processing request latency=1 ms status=200
artifact-keeper-backend  | 2026-08-14T06:44:45.133938Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3/api/v3/collections/ansible/posix/versions/?limit=100}: tower_http::trace::on_request: started processing request
artifact-keeper-backend  | 2026-08-14T06:44:45.135147Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3/api/v3/collections/ansible/posix/versions/?limit=100 correlation_id=60d35c17-4bca-4ceb-ba8c-2ca30c38d65f}: sqlx::query: summary="SELECT id, key, storage_backend, …" db.statement="\n\nSELECT id, key, storage_backend, storage_path, format::text as format, repo_type::text as repo_type, upstream_url, promotion_only, age_gate_enabled, age_gate_min_age_days, age_gate_mode, curation_enabled, curation_default_action FROM repositories WHERE key = $1\n" rows_affected=1 rows_returned=1 elapsed=675.62µs elapsed_secs=0.00067562
artifact-keeper-backend  | 2026-08-14T06:44:45.137594Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3/api/v3/collections/ansible/posix/versions/?limit=100 correlation_id=60d35c17-4bca-4ceb-ba8c-2ca30c38d65f}: sqlx::query: summary="SELECT a.id, a.name, a.version, …" db.statement="\n\nSELECT a.id, a.name, a.version, a.path, a.size_bytes, a.checksum_sha256, am.metadata FROM artifacts a LEFT JOIN artifact_metadata am ON am.artifact_id = a.id WHERE a.repository_id = $1 AND a.is_deleted = false AND LOWER(a.name) = LOWER($2) ORDER BY a.created_at DESC\n" rows_affected=0 rows_returned=0 elapsed=2.126766ms elapsed_secs=0.002126766
artifact-keeper-backend  | 2026-08-14T06:44:45.137676Z  INFO http_request{method=GET uri=/ansible/ansible_galaxy_v3/api/v3/collections/ansible/posix/versions/?limit=100 correlation_id=60d35c17-4bca-4ceb-ba8c-2ca30c38d65f}: artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=60d35c17-4bca-4ceb-ba8c-2ca30c38d65f status=200
artifact-keeper-backend  | 2026-08-14T06:44:45.137735Z DEBUG http_request{method=GET uri=/ansible/ansible_galaxy_v3/api/v3/collections/ansible/posix/versions/?limit=100 correlation_id=60d35c17-4bca-4ceb-ba8c-2ca30c38d65f}: tower_http::trace::on_response: finished processing request latency=3 ms status=200
artifact-keeper-backend  | 2026-08-14T06:44:46.789035Z DEBUG sqlx::query: summary="WITH recovered AS ( …" db.statement="\n\n\n        WITH recovered AS (\n            UPDATE sync_tasks\n            SET status = 'pending',\n                started_at = NULL,\n                claimed_by = NULL,\n                claim_token = NULL,\n                claim_expires_at = NULL\n            WHERE status = 'in_progress'\n              AND (\n                  claim_token IS NULL\n                  OR claimed_by IS NULL\n                  OR claim_expires_at IS NULL\n                  OR claim_expires_at <= NOW()\n              )\n            RETURNING peer_instance_id\n        ),\n        counts AS (\n            SELECT peer_instance_id, COUNT(*)::INT AS n\n            FROM recovered\n            GROUP BY peer_instance_id\n        )\n        UPDATE peer_instances p\n        SET active_transfers = GREATEST(p.active_transfers - counts.n, 0),\n            updated_at = NOW()\n        FROM counts\n        WHERE p.id = counts.peer_instance_id\n        RETURNING counts.n\n        \n" rows_affected=0 rows_returned=0 elapsed=1.806721ms elapsed_secs=0.001806721
artifact-keeper-backend  | 2026-08-14T06:44:46.790342Z DEBUG sqlx::query: summary="SELECT id, name, endpoint_url, …" db.statement="\n\n\n        SELECT\n            id, name, endpoint_url, api_key,\n            sync_window_start, sync_window_end, sync_window_timezone,\n            concurrent_transfers_limit, active_transfers\n        FROM peer_instances\n        WHERE is_local = false\n          AND status IN ('online', 'syncing')\n          AND (backoff_until IS NULL OR backoff_until <= NOW())\n        \n" rows_affected=0 rows_returned=0 elapsed=1.009488ms elapsed_secs=0.001009488
```

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
## Summary

`ansible-galaxy collection install` against a Remote (proxy) Ansible repository failed with `Could not satisfy the following requirements: * ansible.posix:1.5.1 (direct request)` (#3365). The failure names the *requirement*, so it reads as a bad version pin rather than a server that never asked its upstream.

Reproduced live on an isolated pool slot, Remote repo → `https://galaxy.ansible.com/`, real `ansible-galaxy` (core 2.21.3):

```
$ curl '.../ansible/ansible-remote/api/v3/collections/ansible/posix/versions/?limit=100'
{"data":[],"links":{...},"meta":{"count":0}}          # HTTP 200

$ ansible-galaxy collection install -r requirements.yml --force
ERROR! Failed to resolve the requested dependencies map. Could not satisfy the
following requirements: * ansible.posix:1.5.1 (direct request)
```

byte-identical to the reporter's output. Meanwhile the tarball by full path already worked — and the bytes were correct:

```
$ curl -o c.tgz .../ansible/ansible-remote/download/ansible-posix-1.5.1.tar.gz
status=200 size=175162
$ sha256sum c.tgz
112653d7e4e462827f1521130849c650e9dbc403f8b2e9dd5040f2ddabff33ed   # == upstream's declared sha256
```

So serving was fine and only **discovery** was broken — the Ansible analogue of the Helm index defect fixed in #3448/#3515, whose structure and conventions this follows.

## The prior diagnosis was RIGHT, but incomplete

"`version_list` queries local artifacts only and has no Remote branch" is correct and confirmed. It is not sufficient: fixing only that still leaves the install broken, in two ways I found by driving the real client rather than the endpoint.

**1. `versions/{version}/` 404s.** The client hits it immediately after resolving a pinned version, to read the download URL, the digest it verifies the tarball against, and the dependency map. Confirmed in the request log on the unpatched build.

**2. `collections/{ns}/{name}/` 404s, and that silently empties the version list.** When the client has a response cache configured — which it does whenever the server is declared under `[galaxy_server.*]` in `ansible.cfg`, the ordinary enterprise setup and the shape the report describes — `get_collection_versions` reads that route *first* for `updated_at` as a cache key, and treats a 404 from it as "no collection found" by `return []`. I only caught this because I re-tested with a configured server after the ad-hoc `source:` case passed:

```
# patched build, version list already correct — ad-hoc `source:`   → installs
# patched build, same requirements via [galaxy_server.akproxy]     → ERROR! Could not satisfy…
```

A fix without this branch works on a cold client and fails on the reporter's actual configuration.

## Fix

The Remote branch on all three discovery routes, merging what the upstream publishes over whatever the repository holds locally, advertising everything under this repository's own routes.

**Locally-held wins.** The local row is what `download_collection` resolves before it ever consults upstream and carries the artifact's real stored filename; re-advertising the same version from upstream would offer two candidates for one resolvable artifact.

**The version list is walked to completion.** Galaxy clamps `limit` to 100 server-side — verified live, `limit=1000` on a 239-version collection returns 100 items plus a `links.next` — and a pin on an older version lives on a later page. The walk re-requests the same path with a growing `offset` rather than following `links.next`, because that link is spelled relative to the upstream **origin**, not to the configured `upstream_url`: following it means issuing a request outside the configured prefix on the say-so of the upstream's own response body, which is the shape of an SSRF rather than a paging strategy. A private Automation Hub makes the same point without malice — it serves Galaxy under `/api/galaxy/`, so a root-relative `next` resolved against the origin drops the prefix and 404s. Ceiling of 20 pages (2000 versions); truncation is logged, never silent.

**Nothing is forwarded verbatim.** The upstream's `download_url` is an absolute upstream URL; serving it would send the client straight past the proxy — no cache, no download accounting, and none of this repository's authorization on the bytes. Every advertised URL is rebuilt against this repository's own routes. Upstream `name`/`version` values that cannot form a single path segment are dropped with a warning, on #3448's reasoning: encoding was rejected there because an encoder covering `%` also rewrites `+`, which is legal semver build metadata, so the index would have moved every legitimate build-metadata URL.

### Upstream-unreachable behaviour — argued, not silently picked

Consistent with #3515.

* **Empty list rejected outright.** `{"data": [], "meta": {"count": 0}}` at HTTP 200 is indistinguishable from "this collection has no versions", and it is what makes the client blame the requirement. Reproducing it in the failure path would be the same bug wearing a different hat.
* **With versions held locally, they are served** and a warning is logged. An upstream outage must not stop a collection the repository already has from resolving; that is the point of a cache.
* **With nothing held, the upstream's own error is surfaced** — 503 for an upstream 5xx, 404 for a definitive 404, 502 for a connect failure, as `map_proxy_error` classifies it — so an operator can tell an outage from a wrong upstream URL. `version_info` and `collection_info` return the locally-held answer first, so their unreachable path only ever has nothing to degrade to.

The proxy cache already applies RFC 5861 `stale-if-error` to these documents, so a warm proxy keeps serving the last good response before the handler-level fallback is reached at all.

One case deliberately still answers 200 with an empty list: a collection the upstream genuinely does not publish. `galaxy.ansible.com` itself answers `{"meta":{"count":0},"data":[]}` at HTTP 200 for `nosuchns/nosuchcoll`, so passing that through is faithful reporting, not silence — the distinction from the bug is that the server now actually asked.

### One thing that had to change beyond the Remote branch

`GalaxyAPI.get_collection_version_metadata` indexes five fields unconditionally:

```python
CollectionVersionMetadata(data['namespace']['name'], data['collection']['name'], data['version'],
                          download_url, data['artifact']['sha256'],
                          data['metadata']['dependencies'], data['href'], signatures)
```

`namespace` was emitted as a bare string and `href` was absent entirely, so the client raised `TypeError: string indices must be integers` and aborted with "Unexpected Exception, this is probably a bug". This was **not** Remote-specific — installing from a **hosted** Artifact Keeper Ansible repository failed the same way, which I confirmed by publishing a collection and installing it back on the unpatched build. The Remote branch would have been useless without correcting it, so one builder now emits that document for both branches and the contract no longer depends on which copy answered.

**Deliberately not fixed here:** the native upload never reads `MANIFEST.json` out of the collection tarball, so a hosted collection's real dependency map is unknown and is emitted as `{}`. That is a defect in the **upload** path with its own root cause and its own reproduction, and is filed separately rather than smuggled in. Listing a Remote repository's whole collection catalogue also stays local-only: that is a ~30,000-entry upstream enumeration, not a lookup, and no install path reads it.

## Live verification (patched build, isolated pool slot)

```
$ ansible-galaxy collection install -r requirements.yml --force
Downloading http://127.0.0.1:8106/ansible/ansible-remote/download/ansible-posix-1.5.1.tar.gz
Installing 'ansible.posix:1.5.1' to '.../ansible_collections/ansible/posix'
ansible.posix:1.5.1 was installed successfully
```

— matching the issue's "Expected Behavior" block, and note the download goes **through** the proxy.

| check | result |
|---|---|
| Remote version list, `ansible.posix` | 200, **79** versions, `1.5.1` present, hrefs under `/ansible/ansible-remote/…` |
| Pagination, `community.general` (239 upstream) | **239** returned; page-1 `13.3.0` and page-3 `1.3.0` both present |
| Install via ad-hoc `source:` | installed successfully |
| Install via configured `[galaxy_server.*]` (client cache on) | installed successfully |
| **Hosted** repo: publish then install | `202` then installed successfully — was `TypeError` before |
| Remote whose upstream host does not resolve, nothing held | **502**, `Failed to fetch from upstream …` — not a silent empty 200 |
| Same, with a version held locally | **200** serving the held version, with `upstream collection version list unavailable; serving locally-held versions only` logged |
| **Private** Remote repo, anonymous | **401** on both `versions/` and `versions/{v}/` |
| Same, authorized | 200, 79 versions |

## Authorization / visibility

This change adds upstream content to listings, so the question is whether the additions are gated. They are, and by the pre-existing gate rather than anything added here: these routes are nested under `format_routes`, so `repo_visibility_middleware` runs **before** the handler and is unchanged. The private-repo rows above are the live proof — anonymous 401, authorized 200, same upstream behind both. There is no cross-member aggregation on these routes (each answers for exactly one repository), so there is no per-item visibility check to miss.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Closes #3365

---

## Revert-proof, per production hunk

Each hunk was reverted **individually** with the tests kept, rebuilt, and run against a migrated Postgres with `AK_TESTS_REQUIRE_DB=1`. The DB-backed tests run at 0.29–0.41s, not the ~0.04s shape of a test that silently skipped for want of a `DATABASE_URL`.

**Hunk A — the Remote branch in `version_list`**
```
FAIL test_remote_version_list_serves_upstream_versions_3365
  a Remote repository must advertise the upstream's versions; an empty `data` is
  what makes `ansible-galaxy` report 'Could not satisfy the following requirements' (#3365)
FAIL test_remote_version_list_surfaces_an_upstream_failure_when_nothing_is_held_3365
  assertion `left != right` failed: upstream 503 with nothing held locally must not
  answer 200: an empty `data` reads as 'this collection has no versions' and the
  install fails blaming the requirement (#3365)
    left: 200   right: 200
PASS test_remote_version_list_still_serves_held_versions_when_upstream_is_down_3365
```
The third passing under the revert is expected and deliberate: it pins the pre-existing local-only behaviour, so it is the no-regression control, not a revert-proof.

**Hunk B — the Remote branch in `version_info`**
```
FAIL test_remote_version_info_serves_upstream_metadata_3365
  assertion `left == right` failed: a Remote repository must answer for a version it
  does not hold; a 404 here fails the install one step after the version list
  resolved it (#3365); body: Collection version not found
    left: 404   right: 200
```

**Hunk C — the addressability filter in `merge_upstream_versions`**
```
FAIL test_merge_drops_upstream_versions_that_cannot_be_addressed_3365
  the unaddressable version must be omitted, got [Object {
    "href": String("/ansible/gx/api/v3/collections/testns/testcoll/versions/../../../../etc/passwd/"),
    "version": String("../../../../etc/passwd")}]
PASS test_ordinary_collection_versions_stay_advertisable_3365
Summary: 2 tests run: 1 passed, 1 failed
```
The negative control passes under the revert *by design* — it asserts that real versions (`1.5.1`, `2.0.0-rc.1`, `1.0.0+build.5`, `0.0.1-alpha.1+sha.abcdef`, `10.20.30`) are untouched, using strings deliberately disjoint from the rejection fixtures so it cannot collapse onto its own boundary.

**Hunk D — the Remote branch in `collection_info`**
```
FAIL test_remote_collection_info_serves_upstream_metadata_3365
  assertion `left == right` failed: a Remote repository must answer the collection
  detail route; a 404 makes a cache-enabled client discard the whole version list
  (#3365); body: Collection not found
    left: 404   right: 200
```

**Hunk E — the client-contract document shape (`namespace` object, `href`)**
```
FAIL test_version_info_document_satisfies_the_client_contract_3365
  assertion `left == right` failed: `namespace` must be an OBJECT carrying `name`;
  a bare string makes the client raise TypeError before the install starts
    left: Null   right: "testns"
```

**Hunk F — rewriting `download_url` to this repository's route**
```
FAIL test_rewrite_upstream_version_info_points_download_at_this_repo_3365
  assertion `left == right` failed: the advertised download_url must be this
  repository's own route
    left: String("https://upstream.invalid/api/v3/plugin/ansible/content/published/collections/artifacts/testns-testcoll-1.5.1.tar.gz")
    right: "/ansible/gx/download/testns-testcoll-1.5.1.tar.gz"
```

**Hunk G — the multi-page upstream walk** (reverted to stopping after page one)
```
FAIL test_remote_version_list_walks_every_upstream_page_3365
  assertion `left == right` failed: both pages must be collected, not just the first
    left: 100   right: 101
PASS test_remote_version_list_serves_upstream_versions_3365
Summary: 2 tests run: 1 passed, 1 failed
```
The single-page test passing under a single-page revert is the control: it proves the pagination test is measuring pagination and not just re-testing the basic path.

---

## Notes for the reviewer

* **CHANGELOG conflict is expected.** This PR and its sibling both insert a bullet at the top of `## [Unreleased]` → `### Fixed`, so whichever merges second needs a trivial rebase keeping both bullets.
* **`🔒 Security Audit` fails on the base commit too.** It is red on `1aa0aa3` (the `main` this branched from), so it is a pre-existing advisory in the lockfile rather than anything this PR introduces.
* **Duplication gate:** the changed file is on the `.jscpd.json` `ignore` list (the deliberate per-format-handler exemption), so the changed-file duplication gate does not measure it. Stating that rather than claiming a gate run that does not apply.
* **No `sqlx::query!` macros were added**, so `cargo sqlx prepare` is not required; the new queries use runtime `sqlx::query_as`/`query_scalar` or existing helpers.
* **Full `--lib` suite:** `cargo nextest run --workspace --lib` against a migrated Postgres with `AK_TESTS_REQUIRE_DB=1` — **15906 tests run: 15905 passed, 1 failed, 8 skipped**. The one failure is `services::age_gate_service::tests::db_sweep_requires_current_identity`, unrelated to this change; it **passes on its own** against the same database (`1 test run: 1 passed`), so it is cross-test interference from sharing one database between two concurrent suite runs, not a regression. CI's own `🧪 Backend Unit Tests` job — which runs the identical command with `AK_TESTS_REQUIRE_DB=1` — is **green** on this branch.
* **CI dead zones were checked:** every new test lives in a `#[cfg(test)] mod` inside the changed `.rs` file, so `cargo nextest run --workspace --lib` runs it — not a bin target, not an integration target outside the `--test` allowlist. The DB-backed ones were run with `AK_TESTS_REQUIRE_DB=1` against a migrated Postgres and take 0.3–0.4s, well clear of the ~0.04s signature of a test that silently skipped.


brandonrc:
Reviewed at `7fe420fac1f0a91c5db81e00ef0663eb7eb30d29` against a live backend built from that SHA, a mock upstream, and the real `ansible-galaxy` 2.21.3 client.

**Every claim the PR makes, I reproduced and it held** — authorization, traversal rejection, the unreachable-upstream arms, pagination against the real `galaxy.ansible.com`, cache scoping, and all seven revert-proofs. The full list is at the end, so none of it needs re-litigating. The defects are in what the new Remote branch *doesn't* validate.

Two of them block merge, and together they are an install-arbitrary-code path.

## HIGH-1 — the advertised `download_url` is not bound to the coordinate it names

`ansible.rs:699-704`. The comment immediately above states the invariant:

> *"every `download_url` emitted here has to resolve back through this repository's own route to the version it names."*

The code does not enforce it:

```rust
let filename = upstream
    .pointer("/artifact/filename")
    .and_then(|v| v.as_str())
    .filter(|f| galaxy_coordinate_is_addressable(f))   // only "is this one path segment?"
    .unwrap_or(&canonical)
    .to_string();
```

The upstream's `artifact.filename` is trusted verbatim and never required to correspond to the `{namespace}/{name}/{version}` the client just asked for. `test_rewrite_upstream_version_info_rejects_an_unaddressable_filename_3365` covers traversal shapes only, never correspondence.

Live, with a mock upstream answering the `victimns/victimcoll/versions/1.0.0/` request with `artifact.filename = "testns-testcoll-1.0.0.tar.gz"`:

```
AK doc for victimns.victimcoll:1.0.0 →
  "download_url": "/ansible/ans-e2e4/download/testns-testcoll-1.0.0.tar.gz"

$ ansible-galaxy collection install victimns.victimcoll:1.0.0
Installing 'victimns.victimcoll:1.0.0' to '.../ansible_collections/victimns/victimcoll'
victimns.victimcoll:1.0.0 was installed successfully

$ cat .../victimns/victimcoll/MANIFEST.json
testns.testcoll 1.0.0
```

Different code installed under the requested name, reported as success. And `download_collection` resolves that filename **locally first** via `find_local_by_filename_suffix`, so on a Remote repo holding artifacts an upstream-supplied filename can steer the fetch onto a locally stored artifact of a different collection in the same repository.

This is new surface: before this PR `version_info` 404'd on a Remote miss, so no such URL was ever emitted.

Fix is one line — require the emitted filename to equal `collection_filename(ns, name, version)`, which is already computed just above as `canonical`.

## HIGH-2 — a missing upstream `sha256` silently disables the client's only integrity check

`ansible.rs:710-713`:

```rust
let sha256 = upstream
    .pointer("/artifact/sha256")
    .and_then(|v| v.as_str())
    .unwrap_or_default();   // -> ""
```

`ansible-galaxy` verifies under `if expected_hash:` (`concrete_artifact_manager.py:518`). `""` is falsy, so verification is skipped entirely. AK does not verify the bytes either.

The PR reasons about this field explicitly — "inventing one would be worse than passing none" — but *passing none* here means passing `""`, which is worse than failing. Identical bytes, sole variable being whether the upstream document carries `artifact.sha256`:

```
CASE A (sha256 present, wrong for the bytes):
  [ERROR]: Mismatch artifact hash with downloaded file   → install refused

CASE B (sha256 absent → AK emits ""):
  testns.testcoll:1.0.0 was installed successfully       → installed, zero verification
```

Same for a non-string `sha256` or `artifact: {}`. The key cannot be omitted because the client hard-indexes it, so the correct behaviour is a 502. Combined with HIGH-1 this completes the substitution path.

## MED-3 — a `results`-keyed upstream reproduces #3365 exactly

`ansible.rs:376-386` reads only `doc["data"]`. ansible-core's own client documents that the other shape is normal:

> `# v3 automation-hub is the only known API that uses 'data'; since v3 pulp_ansible does not, we cannot rely on version to indicate which key to use` — `galaxy/api.py:855`

Against a pulp_ansible-shaped upstream (`results`, `meta.count: 2`):

```
{"data":[],"links":{...},"meta":{"count":0}}   HTTP=200
```

Byte-identical to the bug being fixed. `fetch_upstream_versions` returns `Ok(vec![])`, so the `Err(err) if versions.is_empty()` arm never runs and the PR's "an empty `data` is indistinguishable from no versions" argument is defeated by a shape mismatch. `meta.count > 0` with zero extracted versions is a free tell that is not checked.

## MED-4 — 20× unbudgeted metadata amplification

`ansible.rs:355-366` uses `proxy_fetch_capped`, not `proxy_fetch_capped_budgeted`, whose doc says it exists so "N un-rate-limited requests can never drive resident metadata memory past the budget". The PR justifies the omission as "only the version STRINGS survive this scope… there is no derived working set" — but `versions: Vec<String>` accumulates across up to 20 × 8 MiB pages and the serialized response outlives it.

Six concurrent **anonymous** requests on a public Remote:

```
RSS before 157 MB → during 1,459 MB     (each response body 100,201,295 bytes)
```

On the 1 GiB pod limit this file's own history cites (#737/#895), a handful of anonymous requests is an OOM kill.

## MED-5 — no aggregate deadline on the page walk

Per-fetch upstream timeout is 60 s; `UPSTREAM_VERSIONS_MAX_PAGES = 20` multiplies it with no overall bound. Measured at 5 s/page: `real 1m41s` for one request. Worst case ≈ 20 minutes of a held task, on a route that is anonymously reachable and unrate-limited.

## MED-6 — Virtual repositories are still broken

All three handlers branch on `RepositoryType::Remote` only (`:845`, `:938`, `:1092`), while `download_collection` already uses `try_remote_or_virtual_download`. On a Virtual Ansible repo with a Remote member:

```
versions/            200  {"data":[],"meta":{"count":0}}   ← the #3365 symptom, verbatim
versions/1.0.0/      404
collections/ns/name/ 404
download/…tar.gz     works
```

Virtual-in-front-of-Remote is the standard enterprise topology, so #3365 is only half closed. (Same shape as #3602 on the OCI side.)

## LOW

**LOW-7** — the local/hosted `collection_info` branch omits `updated_at` (`:882-905`), and `get_collection_versions` invalidates its cached list only when `modified_str` changes. With `None` forever it never invalidates, so a newly published version stays invisible for the 1-day TTL — on exactly the hosted case the PR fixes. Secondarily, on a Remote holding any local row, `collection_info` reports the local `highest_version` and never consults upstream, while `version_list` next door merges both.

**LOW-8** — 20-page truncation is `tracing::warn!` only; the response is 200 with `links.next: null`, so a >2000-version collection reproduces the original symptom for older pins with no client-visible signal.

**LOW-9** — all seven revert-proofs are genuine (1–3 real failures each, DB-backed at 0.26–0.59 s, not the ~0.04 s skip tell, 16 tests under `--lib`, no hollow tests, negative controls correctly disjoint). But an eighth hunk has no coverage: removing **all three** `galaxy_coordinate_is_addressable` guards on the outbound coordinate (`:845`, `:951`, `:1096`) leaves 16/16 green. Behaviour is still saved downstream by `validate_cache_path`, so this is a coverage hole rather than a live one. Nothing tests the authorization table either.

**LOW-10/11, pre-existing** — `%00` in a coordinate returns a 500 with a raw Postgres error to an anonymous caller (the new validator rejects control chars but runs *after* the DB query); and the 502 catch-all echoes the configured upstream URL to anonymous callers.

**Plausible, by reading** — cache keys carry the raw query string (`…/versions/?limit=100&offset=0/__content__`), diverging from the hashed-key pattern `nuget.rs:520-531` adopted with an explicit comment. Fine on filesystem; on object stores `object_store::path::Path` percent-encodes `?`/`&`, so keys differ per backend and the `list`→`delete` round-trip in `purge_repo_cache` is untested for them. Also, Ansible has no `enforce_curation` on any route, so the newly-live discovery path advertises upstream versions unfiltered.

## What I could not break

**Authorization holds.** Private Remote: anonymous → 401 on all three new routes; authenticated non-member → existence-hiding 404; member/admin → 200. No per-item check is missing because each route answers for exactly one repository.

**Traversal holds.** `../../../../etc/passwd`, `%2e%2e%2f`, `%252e%252e%252f`, `%2f`, `%23`, `%3f` all return 400 **before any outbound request** — verified against the mock's request log, zero upstream hits. Upstream-supplied unaddressable versions are dropped; only `1.0.0` survived.

**Unreachable arms are all as documented** — DNS failure 502, connection refused 502, upstream 500 → 503, upstream 404 → 404, malformed JSON → 502, timeout → 502; with local versions held, 200 serving them under both 503 and 404. Merge verified (local 3.3.3 + upstream 2.0.0/1.5.1 = 3 entries).

**Pagination verified against the real upstream** — `community.general` = 239 across offsets 0/100/200, no gaps or repeats, `limit` clamped to 100, the 302 to `/api/v3/plugin/…` preserves the query string, walk bounded at exactly 20 pages.

**Cache scoping is correct** — keys are `proxy-cache/<repo_key>/…`, per-page distinct, 5-minute TTL; three identical requests produced exactly one upstream fetch. Not the #3130 class.

**CI reachability is fine** — all 16 tests are `#[cfg(test)]` in the changed file, run under `--workspace --lib`, with `AK_TESTS_REQUIRE_DB=1` set.

## Recommendation

HIGH-1 and HIGH-2 block merge. Both are two lines, and the PR already has the pieces — `canonical` is computed one line above the filename fallback. MED-3 is a two-line `results` fallback plus a `meta.count` sanity check and closes a silent re-occurrence of the exact bug being fixed.

brandonrc:
Follow-up hardening pushed to this branch (rebased onto `main` for the wasmtime bump). Two integrity defects in `rewrite_upstream_version_info` are now closed, both reproduced end-to-end with the real `ansible-galaxy` [core 2.21.3] client against a live instance before and after.

### 1. Integrity fail-open (fixed)

An absent, null, non-string, or empty upstream `artifact.sha256` was relayed as `""`. `ansible-galaxy` verifies the tarball only `if expected_hash:`, so an empty string is falsy in Python — the client skipped verification entirely, and the server verified nothing of its own. A poisoned tarball then installed with `rc=0`.

The proxied branch now **refuses such a version with a 502** naming the missing digest rather than emitting `""`; the locally-held branch enforces the same rule for a stored row whose `checksum_sha256` is empty. Inventing a digest would be worse than passing none, and advertising an unverifiable one is worse still.

### 2. Content substitution (fixed)

The advertised `download_url`/`artifact.filename` were taken from the upstream's own `artifact.filename` whenever it was a single addressable path segment — a check that a *substituted-but-well-formed* name (a different collection or version) passes. A malicious upstream naming a different file under a requested version's document made this repository's own download route hand back other bytes under the requested coordinate's name.

The emitted filename is now always the canonical `{namespace}-{name}-{version}.tar.gz` for the coordinate the client asked about, computed one line above as `canonical`. A mismatch is logged and ignored.

### Caching

No change needed. The fail-closed gate runs on every read of the (possibly proxy-cached) raw upstream document, in the pure rewrite step — not at fetch time — so a refused version is re-refused on every subsequent request and can never be promoted into a served `200`. Verified live: two back-to-back requests for the no-sha collection both returned 502 with zero upstream version-info hits in the mock's access log (the raw document was cache-served; the refusal was recomputed).

### Live before/after (ansible-galaxy core 2.21.3, Remote repo → mock upstream serving a poisoned tarball in every row)

| scenario | pre-fix | post-fix |
|---|---|---|
| upstream omits `sha256` | `installed successfully` (rc=0), installed `MANIFEST.json` reads the poison collection — verification skipped | `HTTP Code: 502, Bad Gateway` — nothing installed |
| upstream substitutes `artifact.filename` | `installed successfully` (rc=0), requested coordinate's dir contains the poison collection | `Mismatch artifact hash with downloaded file` — nothing installed |
| honest upstream (control) | `installed successfully` | `installed successfully` — legitimate install intact |

### Tests (each revert-proofed separately)

- `test_rewrite_upstream_version_info_refuses_an_unusable_sha256` (absent / null / number / empty / whitespace) — reverting the sha256 guard fails with *"an upstream document with no artifact object must be refused, not advertised with an empty sha256 the client reads as 'skip verification'"*.
- `test_remote_version_info_refuses_an_upstream_without_a_sha256` (DB-backed 502 through the route) — reverting fails with *"a version whose upstream document carries no sha256 must be refused (502), not served with an empty digest"*.
- `test_rewrite_upstream_version_info_pins_filename_to_the_requested_coordinate` — reverting the filename pin fails with *"an addressable-but-substituted upstream filename must not reach the advertised download_url"*.
- `test_local_version_info_refuses_a_row_without_a_checksum` (DB-backed 500) — reverting the local guard fails with *"a stored row with no checksum must be refused, not advertised with an empty sha256"*.

Gates: `cargo fmt --check` clean, `cargo clippy --workspace --lib --tests -- -D warnings` clean, full `--lib` suite `15910 passed` with `AK_TESTS_REQUIRE_DB=1`. `ansible.rs` is on the format-handler jscpd exclusion list.

Issue #3441, opened 2026-08-18
Docker images don't show as packages when pulled through


mruzicka:
### Description

I've setup Artifact Keeper as a Docker proxy / cache. One virtual Docker repository acting as an aggregate for a number of remote repositories (all the major public ones + several private ones).
I can pull images through it, but they never show as _packages_ in the UI, only their respective manifests / blobs show as _artifacts_ (and only in the flat view).

### Environment Information

- Deployment type: Docker compose
- Operating System: Linux (Debian 12 Bookworm)
- Architecture: x86_64
- Database: PostgreSQL
- Reverse Proxy: company public proxy (OpenResty) + reverse proxy included in the Docker compose deployment (Caddy, with customized configuration)
- Network setup: Docker networking

### Backend Version

1.7.4

### Web UI Version

1.8.0

### Package Format

Docker / OCI

### Expected Behavior

Docker images pulled through Artifact Keeper show as _packages_ in the UI. (Just as Maven packages do when retrieved through the Artifact Keeper.)

### Observed Behavior

Docker images never show as _packages_ when pulled through Artifact Keeper, only their respective manifests / blobs show as _artifacts_ (and only in the flat view).

### Steps to Reproduce

1. configure a Docker repository of type _remote_
2. pull an image through that repository
3. verify that the image does not show as _package_ in the UI

### Logs

```shell

```

### Additional Context

It doesn't matter whether I pull the image directly through the remote repository or a virtual repository having the remote repositry as its member. The image never shows as _package_.

I suspect the pulled-through images are never actually indexed as _packages_.

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have redacted any sensitive information (passwords, tokens, private URLs)
- [x] I am using a supported version (not end-of-life)

brandonrc:
## Confirmed from source — no reproduction needed. Your suspicion is exactly right.

> I suspect the pulled-through images are never actually indexed as *packages*.

They are not, and the code says so in a comment. Verified against current `main` (`63d9d4b0`).

### Mechanism

When the proxy caches a newly-fetched object it calls `ProxyService::index_cached_package` (`backend/src/services/proxy_service.rs:4344`, and the buffered sibling). That function's first act is a format gate:

```rust
let Some(repo_format) = catalog_indexable_format(format_text.as_deref()) else {
    // Only Maven-family proxy repos populate the catalog for now
[...]

PR #3599, opened 2026-08-31
fix(oci): index proxied image tags into the packages catalog

### LINKED ISSUE #3611: Proxied OCI image tags from a direct Remote repository are not indexed into the packages catalog
## Summary

Pulling an image through a **direct Remote** OCI repository caches the artifact and writes an `oci_tags` row, but never creates a `packages` row — so the image does not appear on the Packages page or in `GET /api/v1/packages`.

Scoped out of #3441. That issue also covers the **Virtual**-repository topology, which is a separate and larger gap tracked in #3602; this issue is only the direct-Remote half, which is what PR #3599 fixes.

## Detail

Hosted pushes have populated the catalog for a while, so the initial theory that OCI is excluded from package indexing is not right. There is a Maven-only allow-list (`catalog_indexable_format`), but it gates `index_cached_package`, which for OCI only ever sees **blobs** — adding a `docker` arm there would index dozens of digest-named, versionless layer blobs per image and still not index the image itself.

The actual cause is structural: proxied manifests are cached by `cache_manifest_reference_locally`, which deliberately sits outside that layer (an existing meta-test forbids routing it through, per #1278) and never notified `PackageService`.

## Measured

Pulling `library/alpine:3.19` through a Remote repository writes `artifacts` and `oci_tags` rows and leaves `packages` at **0**.

## Expected

`library/alpine` at version `3.19` appears in `GET /api/v1/packages`, subject to the same repository-visibility filtering as every other catalog entry.

## Constraint worth stating

The write must be gated on the reference being a **tag**, matching the rule the hosted push path already uses. Without that gate a digest pull creates a versionless entry keyed on the digest — catalog noise rather than a package — and multi-arch children would each add one.

## Related

- #3441 — the umbrella report, which stays open until the Virtual topology is fixed
- #3602 — Virtual repositories cache nothing at all on pull
- #3601 — a proxied multi-arch tag's catalogued size is a lower bound, not the image size

brandonrc:
## Summary

A Docker/OCI Remote (proxy) repository cached images correctly — `docker pull` worked and the manifests and blobs appeared under the repository's flat artifact view — but the image never appeared as a **package**, while Maven artifacts pulled through the same instance did (#3441).

Reproduced live before the fix, on an isolated pool slot against `registry-1.docker.io`:

```
$ curl -u ... .../v2/docker-remote/library/alpine/manifests/3.19   # 200, index served
$ curl .../api/v1/packages?repository=docker-remote
{"items":[],"pagination":{"total":0,...}}
$ curl .../api/v1/repositories/docker-remote/artifacts
{"items":[{"path":"v2/library/alpine/manifests/3.19","name":"3.19","version":null,...}]}
$ docker exec ak-...-db psql -c 'select count(*) from packages'
 0
```

So the bytes, the `artifacts` row and the `oci_tags` row were all there; only the catalog row was missing.

## The prior diagnosis was WRONG in its operative claim

The standing theory was "OCI/Docker is deliberately excluded from catalog/package indexing." Checked against the source, that is not what is happening, and acting on it would have produced a fix that changed nothing:

* **OCI is not excluded as a format.** `handle_put_manifest` already populates the catalog for a **pushed** image (`oci_v2.rs`, gated on `oci_reference_is_tag`), with a comment saying it was added precisely because a pushed image was "pullable yet invisible in the Packages tab". Hosted push has worked for some time.
* **There is a real, deliberate allow-list**, `ProxyService::catalog_indexable_format`, which admits only the Maven family and is pinned by a test asserting `docker` is not indexed. But it gates `index_cached_package`, which only ever sees paths that flow through `ProxyService` — and **for OCI those are blobs**. Adding a `docker` arm would have indexed layer blobs: digest-named, versionless, dozens per image, and still not the image.
* **The operative cause is structural.** Proxied manifests are cached by `cache_manifest_reference_locally`, which writes `artifacts`/`oci_tags` with raw SQL and sits outside `ProxyService` *on purpose* — an existing source-level test, `cache_manifest_reference_locally_does_not_call_proxy_cache_artifact`, forbids routing it through `ProxyService` because manifests live in the per-repository storage location while the proxy cache writes to the global `proxy-cache/` backend (#1278's doubled-prefix bug). No package-catalog code was reachable from the proxy manifest path at all.

So this is an omission on one code path, not a policy. The `image:tag` identity a user recognises exists only in that function, and that is where the catalog write goes.

## Fix

One helper, `index_proxied_manifest_package`, called from `handle_get_manifest` on the cold proxy pull **after** `maybe_gate_remote_manifest_scan` agrees to serve — mirroring the push path rather than inventing a second policy. (The first cut of this branch placed the call at the end of `cache_manifest_reference_locally`; the round-2 section below explains why it moved and what else moved with it.)

**Why it does not fill the catalog with noise** — the real question for a multi-manifest format. The write is gated on the reference being a tag, the same rule `handle_put_manifest` already applies:

| pull | catalog effect |
|---|---|
| `docker pull img:1.27` — index fetched **by tag** | one row, `img` @ `1.27` |
| …then each arch child manifest, **by digest** | nothing |
| `docker pull img@sha256:…` | nothing (as a digest-only push adds nothing) |
| re-pull of the same tag | upserts the one row |

One row per image per tag; manifests and blobs stay in the artifact view where they belong.

**Sizing** reuses the two helpers the push path uses, and I measured rather than assumed what that produces:

| case | recorded size | correct? |
|---|---|---|
| hosted push, single-arch (37B config + 17B layer) | **54** | exact |
| proxied **image** manifest | `config.size + layers[].size` | exact, and available before any blob is fetched |
| proxied **index** (multi-arch), e.g. `alpine:3.19` | **0** | a lower bound, not the image size |

An index carries no config or layers of its own, so its size is the sum over child manifests already cached — and on a proxy the children are fetched *after* the index, so a first pull sums nothing. My first draft claimed this "converges upward on a re-pull"; testing it live showed that is false and I corrected it rather than shipping it: a warm proxy cache serves the tag without re-entering the caching path, so the row is only recomputed once the cached tag expires and is re-fetched, and even then the sum is over children's recorded artifact sizes, which on the proxy path are manifest **body lengths** rather than image sizes.

Making that number meaningful means teaching proxied child-manifest artifact rows to carry image sizes — a change to a shared pre-existing path the docker-tag grouping also reads. Filed separately rather than smuggled in. #3441 is about the image appearing at all, and a size that is honest about being a lower bound beats an invented one.

**Best-effort**, like every other write in that function after the manifest body itself: a catalog failure must never fail a pull the client already committed to.

## Scope limit: the Virtual topology is NOT covered by this PR

The report describes "one virtual Docker repository acting as an aggregate for a number of remote repositories" and says the image never shows as a package "whether I pull the image directly through the remote repository or a virtual repository". This PR fixes the **direct Remote** half. The Virtual half has a different cause and is filed as #3602.

Measured on this build, one Virtual repository with a Remote member:

| pull | `artifacts` | `oci_tags` | `packages` |
|---|---|---|---|
| direct through the **Remote** member | row | row | **row** (this PR) |
| through the **Virtual** repository | none | none | none |

The Virtual branch of `handle_get_manifest` resolves through `resolve_virtual_manifest` and returns `build_oci_proxy_response` directly — it never reaches `cache_manifest_reference_locally`, so on that path nothing is cached at all and there is no row for a catalog write to hang off. Delegating it to the member (`oci_repo_info_from_member`) is the plausible shape, but it is a behaviour change to a path carrying blob-GC ref integrity (#1409), the proxy-cache storage-location split (#1278) and the Virtual inline scan gate (#3023), plus storage and quota accounting for a repository type that stores nothing today. That needs its own reproduction and review rather than riding along here.

**So #3441 should not be treated as fully closing the report until #3602 lands** — flagging that explicitly rather than letting the `Closes` line below decide it silently.

## Authorization / visibility

The direction of this change is *adding rows to a listing*, so the question is whether they are gated. They are, and nothing new was needed:

* `list_packages` builds `build_visibility_clause_for(&visibility, "r", …)` into **both** its page query and its count query, joined through `repositories`.
* A `packages` row belongs to exactly one repository (`UNIQUE (repository_id, name)`), so there is no cross-member aggregation here and no per-item check to miss — unlike the virtual-member listing cases this repo has a history with.
* Verified live: a **private** Docker Remote repository's proxied image is absent from `/api/v1/packages` for an anonymous caller and present for an authorized one, with the same pull behind both.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Closes #3611

Refs #3441 (deliberately NOT `Closes`)

This fixes the **direct-Remote** half of #3441 only. The reporter's stated topology is a
virtual repository fronting remotes, and a pull through a Virtual repository currently
caches nothing at all — no artifact, no tag row, no package — because that branch never
reaches the caching path. That is #3602.

So #3441 must stay open until #3602 lands: on the reporter's setup the packages list would
still be empty after this merges, and auto-closing would tell them it was fixed when it is not.

---

## Revert-proof, per production hunk

Each hunk was reverted **individually** with the tests kept, rebuilt, and run against a migrated Postgres with `AK_TESTS_REQUIRE_DB=1`. These run at 0.36–0.71s, not the ~0.04s shape of a DB test that silently skipped for want of a `DATABASE_URL`.

**Hunk A — the catalog write itself** (whole block deleted)
```
PASS test_proxied_digest_pull_creates_no_package_row_3441
FAIL test_proxied_tag_repull_keeps_one_package_row_3441
  assertion `left == right` failed: two pulls of one tag must leave ONE catalog row,
  not one per pull: []
    left: 0   right: 1
FAIL test_proxied_tag_pull_creates_a_package_row_3441
  assertion `left == right` failed: a tag pulled through a proxy must produce exactly
  one packages row; an empty catalog is #3441 -- the image is pullable and appears
  under artifacts, but never on the Packages page. Got []
    left: 0   right: 1
FAIL test_proxied_index_tag_pull_creates_one_package_row_3441
  assertion `left == right` failed: a multi-arch tag must produce exactly one catalog
  row -- one per architecture would be exactly the duplicate noise this gate exists
  to prevent. Got []
    left: 0   right: 1
Summary: 4 tests run: 1 passed, 3 failed
```
The one that passes is the digest negative control, and it passing here is *why* it carries a second assertion: with the whole write gone, "no package row" is trivially true. It asserts `artifacts > 0` first, so under hunk B (below) — where the write exists but the gate does not — it still discriminates.

**Hunk B — the `oci_reference_is_tag` gate** (removed; every reference indexed)
```
FAIL test_proxied_digest_pull_creates_no_package_row_3441
  a digest-addressed pull names no version a user chose (it is how the child
  manifests of a multi-arch index are fetched), so it must add nothing to the
  catalog. Got [("app", "sha256:8a3c7d54b626e6c937a976a65e3be43a0194453ffaf86e233ceea4997c0888bf", 106)]
```
That row — an image "version" that is a content digest — is exactly the catalog noise the gate exists to prevent, and it is what a multi-arch pull would produce once per architecture.

**Hunk C — sizing from `manifest_total_size` rather than the manifest body length**
```
FAIL test_proxied_tag_pull_creates_a_package_row_3441
  assertion `left == right` failed: an image manifest must be sized from its own
  config+layers, which is available on a proxy before any blob is fetched
    left: 399   right: 92
```
399 is the manifest document's own byte length; 92 is the fixture's real `config.size + layers[].size`. The expected value is computed from the fixture's bytes, not from anything the handler produced.

### One thing the revert-proofs did not catch, and how it surfaced

Running `-E 'test(3441)'` locally passed 4/4 while CI's **Backend Unit Tests** went red. The cause was the existing source-level meta-test `cache_manifest_reference_locally_does_not_call_proxy_cache_artifact`, which greps this function's **source text** for the literal `ProxyService` — and my new *comment* explained the fix by naming that type three times. A comment failed a test. The comment is reworded to spell the name around, with an in-place note saying why so the next editor does not re-introduce it; the guard itself is untouched, since it is protecting a real #1278 regression. Filtered local runs are exactly how that class of failure hides, and it is the reason the full `--lib` suite is the gate rather than a filtered subset.

---

---

## Round 2 (review + blind red-team of the first cut): three control-consistency defects, fixed here

A desk review and an independent live red-team run of the first cut of this branch confirmed the catalog write worked — and found three consistency defects in *where* it ran (#3611 round 2). The red team also verified what is NOT broken: the tag gate (digest references excluded, digest-looking tags included), repository-visibility filtering of the new rows (anonymous and non-member both get `total=0` for a private proxy repo, in both the page and count queries), idempotency under 8 concurrent cold pulls, the #1278 meta-test still constraining, and — critically — that manifest GET, blob GET and blob HEAD all stay 403 for gated content. These are advertising-integrity defects, not a payload bypass.

### 1. The catalog write ran on the wrong side of the inline scan gate

On the cold proxy pull, `cache_manifest_or_compute_digest` (and therefore the first cut's catalog write) executes **before** `maybe_gate_remote_manifest_scan`, while `record_oci_manifest_pull` deliberately sits after it — the #3446 comment states the invariant outright. So an image the gate refused to serve (403/423 under `scan_on_proxy`) was still advertised on the Packages page and in `/v2/_catalog`. Reproduced live on a `fail_closed` repo with the scanner unavailable (pre-fix image):

```
packages(before)  total=0
GET .../app/manifests/1.0        -> 423 DENIED "withheld: image scan pending or inconclusive"
packages(after)   total=1        -> app @ 1.0
```

The write is hoisted out of `cache_manifest_reference_locally` into `index_proxied_manifest_package`, called by `handle_get_manifest` on the cold pull **after** the scan gate returns `Ok`, next to the download counter and for the same reason. The caching itself stays pre-gate on purpose — it is what gives the gate `manifest_blob_refs` to reassemble a layout from. Consequence, stated rather than hidden: a manifest cached by a *refused* pull is indexed by the next upstream re-fetch that passes the gate (when the cached tag expires and the cold path re-enters); warm serves do not write catalog rows, same as before.

### 2. HEAD published to the catalog

`handle_head_manifest` shares the caching helper and has no scan gate at all — deliberately ("HEAD stays ungated — parity with the direct-Remote HEAD path"). With the write inside the helper, a bare `HEAD` with no prior GET published a catalog row for content the GET path refuses. Reproduced live pre-fix: fresh repo, `HEAD .../manifests/2.0` → 200 + `Docker-Content-Digest`, `packages` went 0 → 1 while `GET` of the same reference stayed 423. Relocating the write closes the catalog half; the HEAD handler now carries a comment pinning that it must not index.

**Decision on the HEAD oracle itself: kept, documented.** A HEAD (and by-digest HEAD) still returns 200 + digest for content the GET refuses. That is pre-existing, stated-rationale behaviour: HEAD serves no bytes; every caller has already passed repository authorization (anonymous callers only reach it on public repos); and the scan gate is a policy gate on *delivery*, not a confidentiality boundary — the content is the upstream registry's, which an authorized puller can query directly. Gating HEAD would change the documented parity contract for every client that probes before pulling, which is a separate judgement from this fix. If that trade is re-decided, it should be its own issue.

### 3. Valid 101–128 character tags silently failed to index

`packages.version` and `package_versions.version` were `VARCHAR(100)` (019) while the OCI tag grammar allows 128, and `try_create_or_update_from_artifact` swallows errors into a `warn!` — so a valid long tag pulled 200 and never indexed, with nothing louder than a log line. CI tag schemes (`v1.2.3-nightly-<date>-<sha>-<platform>-<branch>`) cross 100 routinely. Confirmed live at the exact boundary pre-fix: `len=100` → 1 package row; `len=128` → 200, artifacts cached, 0 package rows. **Migration 211** widens both columns to `VARCHAR(128)` — metadata-only in Postgres, no rewrite, no index rebuild. This was latent on the hosted push path too (same swallow, same columns), so the migration fixes both. The best-effort swallow itself is unchanged: it is the documented contract ("a catalog failure must never fail a pull"), and with the root cause fixed the remaining swallowed cases are genuine transient DB failures. Ledger checked by hand against `main` (latest: 208) and every open PR (209/210 claimed by #3578/#3549) — 211 is unclaimed, so no invisible same-number collision.

Also: `test_catalog_indexable_format_other_formats_not_indexed` (proxy_service.rs) still said "docker must not be catalog-indexed yet" — literally true of that allow-list, misleading after this PR. Reworded to say docker is indexed at the manifest seam, post-gate, and that blobs must never index through the shared proxy-fetch layer.

### Round-2 revert-proofs (each fix reverted individually, tests kept, rebuilt, re-run)

**Ordering fix reverted** (write moved back inside `cache_manifest_reference_locally`, post-gate call removed):
```
FAIL test_scan_blocked_pull_does_not_index_package_3611
  an image the scan gate refuses to serve must NOT be advertised in the packages
  catalog; a row here means the catalog write ran on the wrong side of
  `maybe_gate_remote_manifest_scan`. Got [("app", "1.0", 108)]
FAIL test_head_manifest_publishes_no_package_row_3611
  a bare HEAD must not publish the image into the packages catalog; a row here means
  `handle_head_manifest`'s cache write still indexes packages. Got [("app", "1.0", 102)]
PASS test_scan_allowed_pull_still_indexes_package_3611
```

**HEAD half reverted in isolation** (GET-site fix kept, an index call added back to the HEAD handler):
```
FAIL test_head_manifest_publishes_no_package_row_3611   (same assertion as above)
PASS test_scan_blocked_pull_does_not_index_package_3611
```
— so the HEAD test constrains the HEAD path specifically, not just the shared helper.

**Write dropped entirely** (no call anywhere — the positive controls must catch it, or "blocked pull leaves packages empty" would be trivially satisfiable by deleting the feature):
```
FAIL test_proxied_tag_pull_creates_a_package_row_3441
  ... an empty catalog is #3441 ... Got []   left: 0  right: 1
FAIL test_scan_allowed_pull_still_indexes_package_3611
  an ALLOWED pull on the same fail_closed repo must still index -- moving the write
  behind the gate must not lose the #3441 fix. Got []   left: 0  right: 1
```

**Migration 211 reverted** (test DB columns ALTERed back to `VARCHAR(100)`, binary unchanged):
```
FAIL test_proxied_max_length_tag_indexes_3611
  a 128-char tag is valid per the OCI grammar and must index; an empty catalog here
  is the VARCHAR(100) truncation failure being swallowed into a warn. Got 0 rows
    left: 0   right: 1
```
Re-widened to 128: `1 test run: 1 passed`. All failing runs executed at 0.35–0.42s against a migrated Postgres with `AK_TESTS_REQUIRE_DB=1` — not the ~0.04s shape of a silently-skipped DB test.

### Round-2 live verification (isolated pool instance, patched image, migration 211 applied to a populated database at startup)

```
defect 1  packages(before) total=0
          GET  .../app/manifests/3.0  -> 423 DENIED (fail_closed, scanner unavailable)
          packages(after)  total=0    (artifacts row cached: the path was exercised)
defect 2  packages(before) total=0
          HEAD .../app/manifests/4.0  -> 200 + Docker-Content-Digest
          packages(after)  total=0    (artifacts row cached)
control   GET  .../app/manifests/5.0  -> 200 on a no-scan repo
          packages(after)  total=1    -> app @ 5.0
defect 3  GET  .../app/manifests/<128-char tag, cold> -> 200
          package_versions row present with length(version)=128
```

The pre-fix image on the same rig reproduced all three (423 + `total=1`, HEAD + `total=1`, and len=100 indexed / len=128 silently absent) before the patched image was deployed.

---

## Notes for the reviewer
* **Round 2 adds migration `211_packages_version_oci_tag_length.sql`.** The ledger was checked by hand against `main` (latest applied: 208) and every open PR (209 and 210 are claimed by #3578/#3549) — 211 collides with nothing, so no invisible same-number merge hazard. The ALTERs are VARCHAR widenings: metadata-only, no table rewrite, no index rebuild. Verified applied at startup against a populated database on the live pool instance.
* **Round-2 full `--lib` suite:** `cargo nextest run --workspace --lib --no-fail-fast` with `AK_TESTS_REQUIRE_DB=1` against a migrated Postgres — **15898 tests run: 15897 passed, 1 failed, 8 skipped**. The failure is `services::age_gate_service::tests::db_sweep_requires_current_identity` (untouched by this PR; its `auto_approve_aged_reviews` is a global sweep racing other tests' rows on a shared DB). It passes standalone and the whole 48-test age-gate module passes 48/48 in isolation on the same DB.
* **Round-2 duplication gate:** `oci_v2.rs` remains on the `.jscpd.json` ignore list; `proxy_service.rs` (the other changed `.rs`) measures **0.26%** through the CI prepare-script + jscpd@4 pipeline, against the 3% threshold.
* **Round 2 still adds no `sqlx::query!` macros** — the relocated write reuses the existing runtime-query helpers, so `cargo sqlx prepare` is still not required.

* **CHANGELOG conflict is expected.** This PR and its sibling both insert a bullet at the top of `## [Unreleased]` → `### Fixed`, so whichever merges second needs a trivial rebase keeping both bullets.
* **`🧪 Backend Unit Tests` red on the first run of this SHA is a known-shape flake, not this change.** The failure was `services::proxy_service::tests::test_lazy_catalog_backfill_permit_is_acquired_then_released` — a file this PR does not touch. It polls a semaphore for at most 50 x 10ms waiting for a spawned task to drop its permit, so it is wall-clock bounded at 500ms and fails on a loaded runner. The same test passes at **0.073s** in both local full-suite runs of this exact commit. Re-run requested.
* **`🔒 Security Audit` fails on the base commit too.** It is red on `1aa0aa3` (the `main` this branched from), so it is a pre-existing advisory in the lockfile rather than anything this PR introduces.
* **Duplication gate:** the changed file is on the `.jscpd.json` `ignore` list (the deliberate per-format-handler exemption), so the changed-file duplication gate does not measure it. Stating that rather than claiming a gate run that does not apply.
* **No `sqlx::query!` macros were added**, so `cargo sqlx prepare` is not required; the new queries use runtime `sqlx::query_as`/`query_scalar` or existing helpers.
* **Full `--lib` suite:** `cargo nextest run --workspace --lib` against a migrated Postgres with `AK_TESTS_REQUIRE_DB=1` — **15894 tests run: 15893 passed, 1 failed, 8 skipped**. The one failure is `api::handlers::admin::tests::test_get_system_stats_reports_proxy_cache_totals_db`, unrelated to this change; it **passes on its own** against the same database (`1 test run: 1 passed`) once nothing else is writing to it, so it is interference from a live backend sharing the test database, not a regression.
* **CI dead zones were checked:** every new test lives in a `#[cfg(test)] mod` inside the changed `.rs` file, so `cargo nextest run --workspace --lib` runs it — not a bin target, not an integration target outside the `--test` allowlist. The DB-backed ones were run with `AK_TESTS_REQUIRE_DB=1` against a migrated Postgres and take 0.3–0.4s, well clear of the ~0.04s signature of a test that silently skipped.


brandonrc:
Reviewed at `5f9afd8163cf7be54f07a5436c42b85d782f0c66` against a migrated Postgres with `AK_TESTS_REQUIRE_DB=1`.

**The core of this is right and I could not break it.** The tag gate holds in both directions, the `#1278` meta-test still genuinely constrains what it was written to constrain, visibility is correctly filtered in both the page and count queries, the write is a real upsert rather than check-then-insert, and all three revert-proofs reproduce exactly including the numbers. Details at the end.

Three defects, all confirmed live. The first blocks merge.

## F1 — HIGH. The catalog write sits on the wrong side of the inline scan-and-block gate

`oci_v2.rs:3946-3964` (the new block inside `cache_manifest_reference_locally`), reached from `handle_get_manifest` via `cache_manifest_or_compute_digest`. On the cold proxy pull the order is:

```
9341  cache_manifest_or_compute_digest(...)   <-- the new catalog write happens HERE
9352  maybe_gate_remote_manifest_scan(...)    <-- scan gate; returns Err(403) / Err(423)
9367  record_oci_manifest_pull(...)           <-- download counting, deliberately AFTER the gate
```

The `#3446` comment three lines below the new write states the invariant outright — *"Counted after the scan gate, so an image the gate refuses to serve is not counted as a download."* The new write is the one thing placed before it.

Confirmed live, both arms:

- fail_closed + mock scanner returning `Vulnerable` → pull returns **403**, and `packages` holds `("app", "1.27", 106)`.
- fail_closed + no scanner (inconclusive) → pull returns **423**, and `packages` holds `("app", "1.27", 100)`.

The operational case is not exotic: a Remote Docker repo with `scan_on_proxy` and fail_closed, and a scanner that is down. Every tag anyone probes gets published to the Packages page as available while `docker pull` returns 423 for all of them. Under a real vulnerable verdict, the Packages page advertises precisely the image the inline scan-and-block feature exists to withhold. Nothing removes the row afterwards — the warm path re-runs the gate and never re-enters the write.

The PR body says the write "respects every gate the pull path enforces". That is true of the authz gates, which all run before the upstream fetch; the scan gate is the exception and is not mentioned.

Fix shape: hoist the catalog write out of `cache_manifest_reference_locally` to the two call sites, after `maybe_gate_remote_manifest_scan` returns `Ok`.

## F2 — MEDIUM. A HEAD probe publishes to the catalog

`handle_head_manifest` (`oci_v2.rs:8053`) calls the same `cache_manifest_or_compute_digest`, so it now writes catalog rows. The HEAD handler has no scan gate at all — deliberately, and documented at `:8036-8040` (*"HEAD stays ungated — parity with the direct-Remote HEAD path"*) and `:8035` (*"UNRECORDED-PROXY-SERVE: this is the HEAD handler. A HEAD serves no body, so it is not a download"*).

Confirmed live: anonymous `HEAD /v2/<remote>/app/manifests/1.27` → 200, `packages = [("app","1.27",100)]`. Zero bytes served, not counted as a download, never scan-gated, but published to the Packages page. It is also the cheapest route around F1 — on a fail_closed repo, HEAD publishes an image that GET will 403.

Secondary effect: `packages.version` is clobbered by whichever tag `version_compare` ranks highest, so an anonymous sweep of upstream tags decides the version string shown for the package. `packages` stays one row per image via `UNIQUE (repository_id, name)`, but `package_versions` grows one row per distinct tag probed.

## F3 — MEDIUM. Valid 101–128 character tags silently fail to index

`packages.version` and `package_versions.version` are `character varying(100)` (`019_builds_packages.sql:29,44`). The OCI tag grammar allows 128. `try_create_or_update_from_artifact` (`package_service.rs:172`) swallows the error into a `warn!`.

Confirmed live at the exact boundary:

```
len=100  status=200  artifacts=2  package_rows=1
len=101  status=200  artifacts=2  package_rows=0
len=128  status=200  artifacts=2  package_rows=0
```

Postgres raises `22001 value too long` on insert (an explicit cast truncates; an INSERT does not). So for any tag over 100 characters the pull succeeds, the artifact caches, and the Packages page stays empty — the exact #3441 symptom, unfixed, plus a warn line per pull. CI tagging schemes like `v1.2.3-nightly-20260831-<sha>-linux-amd64-<branch>` cross 100 routinely.

This is latent on the hosted push path too (`oci_v2.rs:9741`), so the PR inherits rather than creates it — but it means #3611 is not closed for long tags, and it is the one case where a legitimate tag passes the gate and the write then drops it.

## What held up

**Tag gate, both directions.** `oci_reference_is_tag` is `!reference.contains(':')`. Every digest form contains `:` and the OCI tag grammar cannot, so no digest leaks through; digest-*looking* tags like `sha256_8a3c` correctly index as tags. In the other direction, 128-char, uppercase, leading `_`, and `.`/`-`/`_` mixes all pass the gate and reach the write — only the varchar truncation drops them, which is F3 rather than a gate defect.

**Authz ordering.** `authenticate_oci_read`, anon-vs-public, `enforce_scan_pull_scope`, `enforce_token_repo_scope`, `require_oci_repo_read_access` and `evaluate_curation_lookup` all run before the upstream fetch, hence before the write. Only the scan gate is misordered.

**Visibility.** `list_packages` builds `build_visibility_clause_for` into both the page query and the count query. No count leak — the author's claim holds. And the Virtual case is clean: `resolve_virtual_manifest` returns via `build_oci_proxy_response` without touching the caching path, and the fall-through bails on `repo_type != Remote`. Verified live with a public Virtual fronting a private Remote member: no rows under either repository id.

**The #3587 race does not apply.** `create_or_update_from_artifact` is `INSERT … ON CONFLICT (repository_id, name) DO UPDATE … RETURNING` plus an `ON CONFLICT (package_id, version)` CTE. Eight concurrent cold pulls of one tag: 8× 200, exactly one row.

**The #1278 meta-test is not defeated by the reword.** It still extracts a 16,023-character body that includes the new block and terminates at the function's real closing brace. I injected `// META-CHECK: ProxyService` into the new comment, the test went red, and it went green again on removal.

**All three revert-proofs re-done independently**, reproducing the published numbers: A → 3 fail with the digest control passing; B → the digest control fires with `("app","sha256:acddf74e…",106)`; C → `left: 399, right: 92`. The digest negative control genuinely discriminates under B — its `artifacts > 0` assertion is load-bearing.

**The flaky unit test is unrelated.** 22 runs of `test_lazy_catalog_backfill_permit_is_acquired_then_released` — 12 idle, 10 under 40 spinners at load average 19 — all passed, 0.052–0.135 s against a 0.5 s budget. `proxy_service.rs` is not in the diff. Its real failure mode is DNS resolution of `postgres://invalid/`, not CPU. Pre-existing, as the author said, though a bounded connect timeout would settle it.

**CI reachability is genuine.** The new tests are in a `#[cfg(test)] mod` inside a changed `.rs`, so `--workspace --lib` reaches them, and the job sets both `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1`. They run at 0.36–0.41 s, not the ~0.04 s skip tell.

**Security Audit** is red on base `1aa0aa3` too — unrelated to this diff, and now fixed on `main` by #3610.

## Minor

`test_catalog_indexable_format_other_formats_not_indexed` (`proxy_service.rs:8476`) still asserts *"docker must not be catalog-indexed yet"*. Literally true of that allow-list, but misleading as a statement about the product now.

## Recommendation

F1 blocks merge. On a release whose headline is inline scan-and-block, a change that advertises scan-blocked images in the product catalog is a regression against the release's own enforcement story. F2 is fixed by the same relocation. F3 needs a migration, or at minimum a length check with a real warning, before #3611 can be called closed.

brandonrc:
Live red-team verification on an isolated instance of this exact build (`pr3599`), run **blind** — the operator was told only that something changed about how proxied images are surfaced, not what the earlier review found. It reproduced both findings and established one thing that matters for severity.

## The bytes are NOT deliverable — this is an advertising-integrity defect, not a payload bypass

The earlier review left that ambiguous. Verified explicitly on a `scan_on_proxy` + `fail_closed` repo with the scanner unavailable:

```
GET  /v2/<key>/app/manifests/1.0        -> 403 "blocked by scan policy"   (warm re-GET also 403)
HEAD /v2/<key>/blobs/<config-digest>    -> 403
GET  /v2/<key>/blobs/<config-digest>    -> 403 "blob belongs to an image blocked by scan policy"
GET  /v2/<key>/blobs/<layer-digest>     -> 403
```

The image cannot be reassembled. The #3292/#3376 blob re-block holds. So the correct framing is: a security control whose purpose is that unvetted images do not surface, **surfaces them anyway** — persistently, in the product UI — while correctly withholding the content.

On that basis the red team rates this **MEDIUM** where the earlier review said HIGH. I think the downgrade is fair and the reasoning is sound. It is still a real defect on a release whose headline is inline scan-and-block.

## Both mechanisms confirmed live

**Cache-before-gate**, cold GET:
```
packages(before) total=0
GET .../manifests/1.0 -> 403
packages(after)  total=1  -> app @ 1.0, size_bytes=4183664
GET /v2/_catalog        -> {"repositories":["dockerhub-proxy/app"]}
```

**Verb asymmetry**, fresh repo, HEAD only, no prior GET:
```
packages(before) total=0
HEAD .../manifests/1.0 -> 200, Docker-Content-Digest: sha256:837f46be..., Content-Length: 527
packages(after)  total=1
GET  .../manifests/1.0 -> 403
```

That second one is worth naming for what it is: **a HEAD-based existence and content-digest oracle for blocked content**, and simultaneously the cheapest route to populating the catalog past the gate.

## The tag gate works, and by-digest is a partial exception

By-digest is correctly kept out of the package write — but still leaks:

```
HEAD@digest -> 200 (leak)   GET@digest -> 403   packages total=0 (no row)   artifacts row still written
```

## Visibility holds — the earlier review's authorization analysis is correct

Private proxy repo, the new package row: anonymous `total=0`, authenticated non-member `total=0`, admin `total=1`. No leak in either the page or the count.

## Also refuted, so the coverage is falsifiable

`fail_open` behaves as designed, serving with a loud `X-AK-Scan: pending`. Not a finding.

## Method note

The upstream SSRF resolver is fail-closed for private addresses and this host has no routable public IP, so no local upstream registry was reachable. The operator recreated the container from the identical image/volume/network with one delta — `AK_SSRF_ALLOW_PRIVATE_CIDRS` plus `UPSTREAM_ALLOW_PRIVATE_IPS` — to reach a local `registry:2`. That relaxes only SSRF, which is orthogonal to the scan-and-block logic under test, and it was reverted afterwards. Worth noting the SSRF control itself was working correctly.

Issue #3603, opened 2026-08-31
Hosted scans of PyPI sdists, RubyGems, Cargo and NuGet catalog nothing in directory mode (same shape as #3442)


brandonrc:
## Summary

#3442 fixed one instance of a wider shape: the hosted-upload scan path extracts an artifact into a workspace and runs the CVE engine in **directory mode**, and directory-source cataloging only reads a handful of ecosystem file layouts. For a format whose published artifact does not happen to contain one of those layouts, the engine catalogs nothing, matches nothing, and the scan is persisted as `completed` with `findings_count = 0` — indistinguishable from clean, with no error.

#3442 closed npm by threading an `ExpectedComponent` pin (the machinery from #3003/#3004) from the artifact's own registry row into the scan workspace. It was scoped to npm and its handler aliases deliberately. This issue records the formats that share the defect and were **not** fixed there.

## Measured

Each format's real published artifact was staged exactly the way the hosted scan path stages it (original archive + extracted tree in one directory) and scanned with the bundled engine against a current CVE database. The "pinned" column re-runs the same engine over a minimal ecosystem-native metadata file naming the same coordinate, which distinguishes *"the advisory data is missing"* from *"the artifact was never cataloged"*.

| format | fixture | as the hosted path scans it | with a coordinate pin | verdict |
|---|---|---|---|---|
| npm `.tgz` | `lodash@4.17.11` | **0** | 7 | fixed in #3442 |
| PyPI sdist `.tar.gz` | `PyYAML@5.3.1` | **0** | 1 | **same defect** |
| RubyGems `.gem` | `rack@2.0.7` | **0** | 35 | **same defect** |
| Cargo `.crate` | `smallvec@1.6.0` | **0** | 1 | **same defect** |
| NuGet `.nupkg` | `Newtonsoft.Json@12.0.1` | **0** | 1 | **same defect** |
| PyPI wheel `.whl` | `PyYAML@5.3.1` | 1 | — | cataloged natively, fine |
| Maven `.jar` | `commons-collections@3.2.1` | 4 | — | cataloged natively, fine |

So: every affected format has advisory data available; what is missing is a gradeable component in the workspace.

## What each one needs

[...]

PR #3604, opened 2026-08-31
fix(scanner): grade natively-published npm packages instead of reporting a false clean

### LINKED ISSUE #3442: Grype scans of natively-published npm packages (dir-mode) always return 0 findings
### Description

**Title:** Grype scans of natively-published npm packages (dir-mode) always return 0 findings — same root cause as the #966 OCI regression, never generalized

**Labels:** `bug`, `scanning`

## Summary

Any npm package published directly to a hosted repository gets scanned by Grype in `dir:` mode against the extracted tarball. Syft's directory-source cataloger set deliberately excludes `javascript-package-cataloger` — only `javascript-lock-cataloger` runs — so a standalone `package.json` with no companion lockfile (the normal shape of a published npm tarball) yields zero catalogued components. **Every natively-published npm package returns 0 findings, regardless of actual CVE content**, with no scan error to signal the gap.

Confirmed directly: adding an empty `package-lock.json` alongside `package.json` in an otherwise-identical scan caused `javascript-lock-cataloger` to fire where nothing had run before. A separate image-source scan of equivalent content confirmed `javascript-package-cataloger` itself works and finds real packages (237 in that test) — it's specifically excluded from directory-source scans, by upstream Syft design.

This silently defeats any severity-based promotion gate or quarantine policy for npm — a package with known Critical CVEs scans clean.

## Reproduction

1. Publish an npm package with a known CVE (e.g. `lodash@4.17.11`, 1 Critical + 3 High) to a hosted npm repo
2. Trigger a scan
3. `scan_findings` and `scan_packages` are empty; artifact reports clean
4. Extract the same tarball and run `grype dir:<path> -v`, then repeat with a minimal (even empty `{}`) `package-lock.json` added alongside — only the lockfile-present run shows `javascript-lock-cataloger` activity

## Root cause

`grype_scanner.rs` extracts the tarball and invokes Grype as a `dir:` source. Per Syft's design, directory-source scanning excludes the package.json-only cataloger. The current source already documents having hit this exact failure shape once before, for OCI images:

> *"The dir-mode path below would see only the manifest JSON and return 0 findings (the #966 regression)."*

That was fixed for OCI by routing to a `registry:` reference instead of extract-and-dir-scan (see `is_oci_image_artifact` in `grype_scanner.rs`). **The fix was never generalized** to other registry-hosted single-package formats — npm, and likely PyPI wheels, RubyGems `.gem`, Cargo `.crate`.

## Suggested fix

Mirror the OCI branch. Before the workspace/dir-mode path:

```rust
if artifact.format == "npm" {
    if let Some(version) = artifact.version.as_deref() {
        let purl = format!("pkg:npm/{}@{}", artifact.name, version);
        let report = self.run_grype_target(&purl, &[]).await?;
        // convert_findings / convert_packages as in the existing OCI path
    }
}
```

`artifact.name`/`artifact.version` are already in the artifacts table at scan time — no new extraction needed, and this is cheaper than the dir-mode extraction it replaces.

## Impact

Silent, not loud. No scan error, no warning — just an authoritative "0 findings, clean" result on every natively-published npm package. Any severity-based promotion gate or quarantine policy relying on npm scan results is currently a no-op for this path.

## Scope question for maintainers

Given the shared root cause with the #966 OCI fix, is it worth a general "registry-hosted single-package artifact, no lockfile" check rather than an npm-only branch, to cover PyPI/RubyGems/Cargo the same way? Can help validate the same pattern reproduces for PyPI wheels if useful.

---

Want any other adjustments, or ready to post?

### Environment Information

- Deployment type: local on docker
- Operating System: macos
- Architecture: M5
- Database: Postgres
- Reverse Proxy: Caddy
- Network setup: None

### Backend Version

1.3.0

### Web UI Version

_No response_

### Package Format

npm

### Expected Behavior

Upload vulnerable package. Scan vulnerable package. It finds multiple CVEs. 

### Observed Behavior

Upload package, Scan package, it found 0 CVEs. 

### Steps to Reproduce

1. Publish an npm package with a known CVE (e.g. `lodash@4.17.11`, 1 Critical + 3 High) to a hosted npm repo
2. Trigger a scan
3. `scan_findings` and `scan_packages` are empty; artifact reports clean
4. Extract the same tarball and run `grype dir:<path> -v`, then repeat with a minimal (even empty `{}`) `package-lock.json` added alongside — only the lockfile-present run shows `javascript-lock-cataloger` activity

This difference proves that AK is missing the CVEs on the vulnerable packages

### Logs

```shell
**Log 1 — Full cataloger run against extracted npm package (package.json present, no lockfile). `javascript-package-cataloger` never appears in the list — the exclusion is total, not partial:**


$ docker compose exec backend grype dir:/tmp/lodash-extracted -v 2>&1 | grep -i "cataloger\|task completed"

[0000]  INFO task completed elapsed=343.125µs from=syft task=environment-cataloger
[0000]  INFO task completed elapsed=14.75µs from=syft task=elf-binary-package-cataloger
[0000]  INFO task completed elapsed=257.916µs from=syft task=erlang-otp-application-cataloger
... (58 catalogers total ran)
[0000]  INFO task completed elapsed=424.042µs from=syft task=javascript-lock-cataloger
[0000]  INFO task completed elapsed=441.542µs from=syft task=java-pom-cataloger
... (full list continues)
[0000]  INFO task completed elapsed=500ns from=syft task=os-feature-detection

# Note: javascript-package-cataloger does not appear anywhere in the 58-task list above,
# despite /tmp/lodash-extracted/package/package.json being present and valid:
{
  "name": "lodash",
  "version": "4.17.11",
  ...
}


**Log 2 — Same directory, only difference is an empty `package-lock.json` added alongside the existing `package.json`. `javascript-lock-cataloger` now fires — confirming the exclusion is structural/deliberate (lockfile-gated), not a fluke or misconfiguration:**


$ docker compose exec backend sh -c "echo '{}' > /tmp/lodash-extracted/package/package-lock.json"

$ docker compose exec backend grype dir:/tmp/lodash-extracted -v 2>&1 | grep -i "javascript"

[0000]  INFO task completed elapsed=397.041µs from=syft task=javascript-lock-cataloger


**Interpretation:** `javascript-package-cataloger` (the one that parses standalone `package.json` files, confirmed functional against image/registry sources by Syft's own upstream behavior) is structurally excluded from `dir:` mode scans. Since AK's native npm-publish scan path extracts the tarball and always invokes Grype as `dir:`, every natively-published npm package returns 0 findings — with no scan error to signal the gap, since Syft correctly reports "0 packages found" rather than failing.
```

### Additional Context

Tried over multiple vulnerable packages - every single one that is supposed to return CVEs didn't. 

### Pre-submission Checklist

brandonrc:
Closes #3442

## Summary

This revises the #3442 hosted-npm scan pin so it cannot become an
attacker-controlled false clean. #3442 correctly made the grype verdict a
function of the request coordinate (repository format + name + version, pinned
into the scan workspace). That change silently invalidated an assumption the
scan-reuse path still relied on, and inherited the proxy path's pin without the
cross-check that keeps the proxy path honest. Five issues, each fixed and
covered by a test that fails when the fix is reverted:

1. **Reuse key ignored the pin (CRITICAL).** `ScannerService::find_reusable_scan`
   matched only on `(checksum_sha256, scan_type, status='completed')`. That key
   was sound while a verdict was a pure function of the bytes; #3442 made it a
   function of the coordinate too, so byte-identical uploads under different
   coordinates collapsed onto one cached verdict — and reuse is consulted
   *before* the scan runs, so on a hit the pinned scan never ran. A publisher
   with only `write:artifacts` could publish a vulnerable tarball's bytes under a
   throwaway name with no advisory (priming a clean verdict for that checksum),
   then publish the same bytes at the real coordinate and have the clean verdict
   served for it. The reverse blocks good content. Fixed by carrying a
   `pin_identity` on each `scan_results` row (migration 209) and requiring
   `pin_identity IS NOT DISTINCT FROM` the current request's pin before a row is
   reusable — so a NULL (unpinned) verdict matches only another unpinned request
   and never an npm advisory, and vice versa.

2. **The pin was trusted from an unvalidated coordinate.** The hosted pin is
   built from the `artifacts` row, whose name/version on the generic artifact
   endpoint come from URL path segments and were never cross-checked against the
   bytes. A 30-byte text file uploaded to `.../handlebars/4.0.11/notes.txt`
   graded as `handlebars@4.0.11`. The proxy serve path already cross-checks the
   coordinate against the tarball's own `package/package.json`; the hosted path
   now does too, and on mismatch drops the pin and records the scan `partial`.

3. **Scoped packages mis-parsed on the generic route.** `@babel/traverse/…`
   derived `name=@babel`, `version=traverse`, pinning a component that does not
   exist so a vulnerable scoped package read clean. Fixed at the coordinate
   derivation, which now keeps an npm scope with its name.

4. **A version-less npm upload was recorded as an authoritative clean.** When no
   pin can be built the scan grades nothing gradeable; it is now recorded
   `partial`, not `complete`.

5. **The upgrade note was wrong.** It claimed native `npm publish` grades
   correctly immediately; `store_npm_version` writes the `artifacts` row directly
   and never reaches the scan-on-upload hook, so `npm publish` triggers no scan
   at all. Corrected.

Each fix has a revert-proof test:
- `find_reusable_scan_matches_only_the_same_pin_identity` and
  `test_reuse_key_is_pinned_to_the_coordinate_not_only_the_bytes` (defect 1) —
  the second is a real end-to-end orchestration test: same bytes under two
  coordinates, each earns its own correct verdict (unpinned 0 vs pinned 7),
  neither reuses the other's, and a third byte-and-pin-identical upload still
  reuses (dedup is not disabled). Reverting the pin filter makes the real
  coordinate read the primed `0` instead of its own `7`.
- `test_hosted_npm_pin_requires_the_bytes_to_back_the_coordinate` (defect 2) —
  reverting the cross-check pins a non-tarball as `handlebars@4.0.11`.
- `generic_path_coordinate_tests::npm_scoped_package_keeps_scope_with_name`
  (defect 3) — reverting yields `@babel`/`traverse`.
- `test_hosted_npm_scan_target_carries_the_component_pin` (defect 4) — asserts
  the version-less npm scan is `partial`; reverting yields `complete`.
- Plus pure-helper tests for `pin_identity`, `format_expects_pin` and
  `npm_pin_agrees_with_tarball`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [ ] N/A — internal scan-reuse behavior; no route or schema-response changes.
- [x] Migration is reversible (if applicable): `209_scan_results_pin_identity.sql`
  adds a nullable column and extends one index; both are additive.

## Verification (isolated instance)

Real `lodash-4.17.11.tgz` (checksum `8eb2165…`, ships `package/package.json`
claiming `lodash@4.17.11`; true grype verdict 7 findings / 1 Critical).

Before (pre-fix image):
- `safe-first@1.0.0` (lodash bytes), scan → grype `findings=0`.
- `lodash@4.17.11` (same bytes), scan → grype `findings=0, is_reused=true`
  (served the primer's clean; a forced fresh scan returns the true `7`).

After (this build):
- `safe-first@1.0.0` → grype `findings=0, is_reused=false, pin_identity=NULL,
  completeness=partial` (coordinate disagrees with the bytes).
- `lodash@4.17.11` (same bytes) → grype `findings=7 (1 crit/3 high/3 med),
  is_reused=false, pin_identity=npm|lodash|4.17.11, completeness=complete,
  quarantine=flagged` — the attack is closed.
- A second identical `lodash@4.17.11` upload → grype `findings=7,
  is_reused=true` — legitimate reuse still works.
- Generic upload to `@babel/traverse/7.23.0/x.tgz` derives
  `name=@babel/traverse, version=7.23.0`.


brandonrc:
### CI status: every gate green except a pre-existing, unrelated audit failure

From the `CI Complete` job's own inputs on this run:

```
RESULT_CHECK_RUST: success      RESULT_COVERAGE:      success
RESULT_UNIT:       success      RESULT_INTEGRATION:   success
RESULT_SHELL:      success      RESULT_SMOKE:         success
RESULT_BUILD_BACKEND: success   RESULT_BUILD_OPENSCAP: skipped
RESULT_SECURITY:   failure
```

`CI Complete` aggregates, so the single `security-audit` failure is the only reason it is red.

That failure is **not from this branch**. `cargo-audit` reports `error: 1 vulnerability found!` for **RUSTSEC-2026-0269** (`wasmtime`, filesystem sandbox escape); the other five entries are allowed warnings and do not block. The same two jobs — `Security Audit` and `CI Complete` — fail on the latest `main` run for the same advisory, and this PR touches no dependency: the only lockfile-adjacent change here is one regenerated `.sqlx` query entry.

#3610 (`fix(deps): bump wasmtime and wasmtime-wasi to 36.0.14 (RUSTSEC-2026-0269)`) is the fix already in flight. This branch just needs a rebase once that lands; nothing to change here.

Local gates run against the full tree: `cargo fmt --all --check` clean, `cargo clippy --workspace --lib --tests -- -D warnings` clean, and `cargo nextest run --workspace --lib` with `AK_TESTS_REQUIRE_DB=1` against a migrated Postgres — **15894 / 15894 passed**.

brandonrc:
Reviewed at `9f8f37f6d6141ea8a2856029e0515936c5c78092` on a live backend built from that SHA, with `grype-0.117.0+db-2026-08-20`.

**The fix works and I could not break the core of it.** All three revert-proofs reproduce exactly, including the independence check — reverting hunk 2 leaves hunk 1's test passing. The 21→7 dedup measurement is accurate: a `lodash@4.17.11` tarball shipping its own `package-lock.json` *and* `npm-shrinkwrap.json` really does make grype report 21 raw matches across 3 cataloged copies, and exactly 7 rows are persisted. I probed 50 vulnerable coordinates in one run (251 raw matches → 251 distinct keys, zero collisions) hunting for an over-collapse in `dedupe_findings` and found none; distinct transitives at different versions all survive. Details of everything that held are at the end.

The problem is not in the three hunks. It is that the PR makes the grype verdict a function of `repositories.format` + `artifacts.name` + `artifacts.version` without extending the machinery that decides **when a verdict may be reused** or **where the coordinate comes from**. Four defects follow, all confirmed live.

## 1 — HIGH. Scan reuse still keys only on `(checksum, scan_type)`, so the #3442 false clean survives

`scan_result_service.rs:684-711`:

```sql
WHERE checksum_sha256 = $1
  AND scan_type = $2
  AND status = 'completed'
```

No repository, no format, no pin dimension. That key was *sound before this PR* — every hosted grype scan of the same bytes produced the same answer. This change breaks that invariant and leaves the key untouched. Reuse is consulted before the scan (`scanner_service.rs:5145-5160`), so on a hit the pinned scan never runs.

Live, entirely through the public API:

1. `PUT` the real `lodash-3.10.1.tgz` bytes to a **generic** repo as `blobs/lodash310.bin` → grype completes, 0 findings.
2. `npm publish lodash@3.10.1` (identical bytes) → `is_reused=t`, **0 findings / 0 Critical**.
3. Same artifact with `bypass_dedup: true` → **6 findings / 1 Critical / 3 High**.

Reproduced with `ejs@2.5.7` where the poisoning row came from **auto-scan-on-upload alone** (`scan_on_upload=true`, no admin action), and with `handlebars@4.0.11` poisoned by a version-NULL sibling **inside the same npm repo**: 2/1C vs 19/5C/10H forced-fresh.

It reaches a gate: `PolicyService::evaluate_artifact` (`policy_service.rs:257-299`) counts rows in `scan_findings`, and a reuse-poisoned artifact carries only the unpinned scan's rows. `has_completed_scan` is true, so `block_unscanned` does not catch it either.

**This also runs in the other direction**, which refutes "unpinned formats are untouched": after `lodash@4.17.11` was pinned-scanned (7 findings), the identical bytes uploaded as `opaque-blob.dat` to a **generic** repo recorded `is_reused=t, 7 findings / 1 Crit / 3 High`, with `pkg:npm/lodash@4.17.11` in its SBOM inventory — npm advisories attributed to an opaque blob.

Direction: carry the pin identity on the `scan_results` row and refuse reuse when it differs, or add format to the reuse key.

## 2 — MEDIUM-HIGH. The pin comes from a coordinate nothing validated against the bytes

`scanner_service.rs:5043-5047` builds the pin from the `artifacts` row with no cross-check against the archive's own `package.json`. The proxy path has one — `prepare_pinned`'s own doc at `:955-960` says "What the archive itself claims is cross-checked separately by the serve path before the scan runs." The hosted path inherits the pin but not the check, and `authorize_generic_upload` has no format gate, so a writer picks the coordinate straight out of the URL.

Live: `PUT .../npm-b/artifacts/handlebars/4.0.11/notes.txt` with a **30-byte text file**:

```
Grype scan complete for handlebars: 17 vulnerabilities, 1 components, 1 cataloged
```

Persisted as 17 findings / 4 Critical / 10 High, `quarantine_status = flagged`, and an SBOM row claiming `pkg:npm/handlebars@4.0.11`. Pre-PR the same file graded 0.

The PR frames this as "an internal package shadowing a public name" and accepts it. The surface is wider than that: any artifact in an npm-format repo whose path yields a public coordinate — sidecars, signatures, docs, mirrored files — is now graded as that package. Suggest gating the pin on the artifact plausibly being an npm tarball, or comparing it against the shipped `package/package.json` and falling back to unpinned + `partial` on mismatch.

## 3 — MEDIUM. Scoped packages get a *wrong* pin, and a wrong pin is a silent false clean

`repositories.rs:7846-7856` derives the coordinate as `segments[0]` / `segments[1]`. npm's own layout is `{name}/{version}/{file}` — which for a scoped package is four segments.

Live: `PUT .../npm-b/artifacts/@babel/traverse/7.23.0/traverse-7.23.0.tgz` with the genuine Critical `@babel/traverse@7.23.0` stored `name="@babel"`, `version="traverse"`. Both of `hosted_upload_pin`'s guards pass, so the scan is pinned to a component that does not exist:

```
Grype scan complete for @babel: 0 vulnerabilities, 1 components, 1 cataloged
```

`completed / 0 findings / completeness=complete`, `quarantine_status = clean`, SBOM `pkg:npm/%40babel@traverse`. The identical bytes via `npm publish` correctly report CVE-2023-45133.

This is the exact case the helper's own doc says it avoids — "a pin we cannot name correctly would grade the wrong component, which is worse than the gap it closes." The guards check the coordinate's *shape*, not that a format handler produced it. Per the write-path map this is also the shape inbound peer replication takes on its direct-`PUT` arm.

## 4 — MEDIUM. The residual false clean the PR's own test blesses

`hosted_upload_pin` returns `None` on an empty/NULL version, and the row is still written `completed / complete` because `require_nonempty_catalog` stays false (`:5069`). `artifacts.version` is nullable, and a generic `PUT` yields NULL for any path with fewer than three segments — a path that does auto-scan.

Live: `PUT .../npm-b/artifacts/handlebars-4.0.11.tgz` → 2 findings / 1 Critical, `completeness=complete`, versus 17 / 4C / 10H pinned.

`test_hosted_npm_scan_target_carries_the_component_pin` (`:17772`) asserts `seen[1] == None` with "an artifact with no version has no coordinate to pin, and must not be graded as some other component." *Not pinning* is right. Recording the outcome as an authoritative complete clean is the bug, and the test enshrines it. `completeness = 'partial'` would make it visible.

## 5 — the upgrade note's "graded correctly immediately" is wrong

`store_npm_version` (`npm.rs:4305-4322`) inserts the `artifacts` row with raw SQL and never goes through `ArtifactService`, so it never reaches the scan-on-upload hook. `grep -c scan_artifact backend/src/api/handlers/npm.rs` → **0**. No scheduler sweep covers it either. Confirmed live: `scan_results` stayed empty after every publish until an explicit admin `POST /api/v1/security/scan`.

Operators sizing blast radius from that sentence will expect gates to start failing on new publishes by themselves. They will not.

**The rest of the note checks out** against source: stored rows are not rewritten; the 24 h figure is real (`ZERO_FINDINGS_DEDUP_TTL_DAYS = 1`); `bypass_dedup: true` skips both the same-artifact short-circuit and cross-artifact reuse; and the SQL's `IN ('npm','yarn','bower','pnpm')` is exactly the set whose `handler_key()` is `npm`. Two caveats: the query has no latest-scan-per-artifact restriction, so it over-reports (safe direction, but not "every row is a package that will be graded for the first time"); and given finding 1 the 24 h window understates it whenever the stale row is non-zero, where the TTL is 30 days.

## Minor

**6** — the dedup gate keys on repository format, not on which scanner consumed the pin (`:5340-5343` sits inside `for scanner in &self.scanners`), so it also collapses `DependencyScanner`/`TrivyFsScanner`/`OpenScapScanner` output, none of which reads `expected_component`. I could not construct a real over-collapse; it is scope rather than correctness.

**7** — doc rot in the edited file: `:961` and `:985` both still say hosted uploads are unpinned and keep the flat layout; `:2686-2688` says npm double-counting is "NOT addressed here" while hunk 3 now uses that function for exactly it; `:9229` likewise.

## What held up

Root cause and numbers reproduce exactly (`grype dir:` over the extracted tarball → 0; plus a lockfile → 7). All three revert-proofs re-run against my own Postgres with `AK_TESTS_REQUIRE_DB=1`, matching the published assertion text. The negative control on the unpinned `generic` repo genuinely discriminates — unconditional dedup collapses its three byte-identical findings to 1. The new tests really execute in CI (`--lib`, with `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1` set at `ci.yml:551-565`), running 0.42–1.09 s rather than the ~0.04 s skip tell. `.sqlx` is correct: the new filename is the SHA-256 of its own query string and the old file is deleted in the same commit. Version handling is fine on the native path — `1.0.0-beta.1`, `1.0.0+build.5`, and scoped names all pin and match correctly. And the fix does not lower counts for tarballs that already graded: a crafted 15-finding fixture stayed at 15.

Happy to take findings 1 and 3 if that is faster than reworking them here.

brandonrc:
Live red-team verification on an isolated instance of this exact build (`pr3604`, grype 0.117.0 / DB 2026-08-20), run **blind** — the operator was told only that npm scan grading had changed, not what the earlier review found.

It independently reproduced the reuse-key defect, **rated it CRITICAL rather than HIGH**, and established two things the desk review did not.

## It is attacker-controlled, not just an accident

A publisher holding only `write:artifacts` can force a false clean deliberately:

1. Publish the vulnerable tarball's **bytes** under any name with no advisory. That primes the checksum with a clean verdict.
2. Publish the same bytes under the real coordinate. It reuses the clean verdict and is recorded `clean`.

Measured, fresh bytes `sha256=1e07dab4…`:

```
publish+scan  safe-first@1.0.0    -> findings=0,  reused=False
publish+scan  lodash@4.17.11      -> findings=0,  reused=True,  quarantine_status="clean"
                (same bytes; correct verdict = 7 findings / flagged)
```

The reuse path is shared by every scan trigger — admin `/security/scan`, `scan_repository` at promotion, and `scan_on_upload` — so who fires the scan does not matter.

## The false-positive direction is real too

The earlier review covered false-clean. The reverse also holds, and it blocks good content:

```
publish+scan  lodash@4.17.4       -> findings=10, critical=1, high=4, reused=False
publish+scan  harmless-utils@1.0.0 -> findings=10, reused=True, quarantine_status="flagged"
                (same bytes; correct verdict = clean)
```

## Baseline confirming the verdict is coordinate-determined

Identical bytes `sha256=90ca1149…`, each scanned fresh with `bypass_dedup:true`:

```
lodash@4.17.11        -> findings=7, critical=1, high=3   (correct)
sc-decoy-safe@4.17.11 -> findings=0                        (correct — no advisory for that name)
```

And the enforcement consequence, same coordinate: fresh scan → `quarantine_status="flagged"`; reused-from-benign-sibling → `"clean"`.

## One correction to the earlier review, worth being precise about

The earlier review reported that scoped packages break. The red team found `@babel/traverse@7.23.0` grades **correctly** — `findings=1, critical=1`, matching grype-direct on GHSA-67hx-6x53-jw92.

Both are right; they exercised different write paths. The native `npm publish` path sets name and version correctly, so scoped packages are fine there. The earlier finding was against the **generic artifact endpoint** (`PUT /api/v1/repositories/{key}/artifacts/@babel/traverse/7.23.0/…`), which derives the coordinate as `segments[0]`/`segments[1]` and yields `name="@babel"`, `version="traverse"`. That derivation is unchanged on `main` and still produces a wrong pin on that route.

So: scoped handling is correct on the native path and broken on the generic path. Worth fixing at the derivation rather than in the pin.

## Not tested

PyPI/RubyGems/Cargo/NuGet pins (out of npm scope — #3603 tracks them), and a full promotion-rule end-to-end. Quarantine status is the same gate input, so the impact is established either way.

Issue #3368, opened 2026-08-14
PyPI: artifacts-row reads of proxy-cache content use the S3_PREFIX (ArtifactSource) key layout, guaranteeing a miss on prefixed deployments


ysolomon-plat:
## Summary

On a deployment with `S3_PREFIX` set, PyPI handlers that read a remote member's proxy-cache content through the repository's `artifacts` row use the **wrong S3 key layout**, so the read is structurally guaranteed to miss — the object exists, but under a different key than the one being fetched.

This is the root cause behind the "additional observation" in #3366: the storage read there NotFounds on an object the server's own presigned URLs serve fine. #3367 makes the *symptom* on the metadata route harmless (404 → upstream fallback instead of a permanent 500), but does not fix the mis-routed read itself.

## The two key layouts

The design is documented in `backend/src/services/storage_service.rs`:

> The inner backend's key prefix is set by the `StorageRole` the handle was built for (`prefix = None` for the proxy cache, #1555; the configured `S3_PREFIX` for artifact bytes, #3171).
>
> `StorageRole::ProxyCache` must force `prefix = None`. Proxy-cache content lives at the bucket root (`proxy-cache/...`), not under `S3_PREFIX`, so a presigned key must carry no prefix (#1555).

So, by design, one bucket holds two layouts:

- **ProxyCache role** — no prefix: `proxy-cache/<repo>/<path>/__content__`
- **ArtifactSource role** — prefixed: `<S3_PREFIX>/<key>`

## The bug

An `artifacts` row on a Remote PyPI repository can carry a `storage_key` that is proxy-cache content (`proxy-cache/.../__content__` — written by the ProxyCache-role handle, at the bucket root). But the handlers read it through `state.storage_for_repo(&repo.storage_location())`, which resolves to the registry's shared S3 backend — the one that applies `S3_PREFIX` via `full_key()`.

Result: the read targets `<S3_PREFIX>/proxy-cache/.../__content__`, which never exists. With `S3_PREFIX` unset the two layouts coincide and everything works, which is presumably why this has gone unnoticed.

Affected paths (both from the #3147/#3175/#3179 line):

[...]

PR #3609, opened 2026-08-31
fix(proxy-cache): scope the proxy-cache key space per deployment (#3454)

### LINKED ISSUE #3454: Proxy-cache S3 layout is not scoped per deployment: two deployments sharing a bucket share one proxy-cache key space
> **Status (2026-08-24) — the remaining ask has changed. Read this before picking it up.**
>
> Two of the three halves of this report are **closed**; the third is deliberately still open on *different grounds* than the original title suggests. Do not "fix" this by making `S3_PREFIX` scope the cache — see below.
>
> * **Read/write half — FIXED.** PR #3513 (`9d1789c8`, merged 2026-08-23) anchors proxy-cache keys at the bucket root on both sides, so a prefixed S3 deployment reads back what it wrote instead of running a permanently cold, silently duplicating cache.
> * **Documentation half — FIXED.** `.env.example:104-132` on `main` now states that `S3_PREFIX` scopes **artifact bytes only**, names `proxy-cache/` and `proxy-cache-staging/` as reserved root paths, gives the exact IAM `Resource`/`Action` list a policy must cover (which is what resolves the 403 in this report), and adds the lifecycle-rule recommendation for the staging prefix.
> * **Layout half — STILL OPEN, on ISOLATION grounds, not permissions.** Two Artifact Keeper deployments sharing one bucket and separated only by `S3_PREFIX` share **one** proxy-cache key space: both write `proxy-cache/<repo_key>/<path>/__content__`. Two repositories with the same key in different deployments address the same cache object and can serve each other's cached upstream bytes. Until that is scoped per deployment, **the safe operational guidance is that a bucket must not be shared between two Artifact Keeper deployments — prefixing does not isolate the cache.**
>
> Moving cache content under `S3_PREFIX` is *not* a drop-in fix: it strands every object already cached at the bucket root, and for formats whose local-row read path has no upstream recovery it converts today's silent miss into a permanent failure for legacy rows. That needs a migration story, not a bug fix. See the full write-up in the comment below.
>
> *(Original report follows unchanged.)*

---

### Description

The configuration for the S3 bucket used for storing artifacts allows for specifying a `S3_PREFIX` environment variable, where the artifacts will reside.  However, for use of upstream proxies, the backend places cached files under the hard-coded "proxy-cache" path, and staging under "proxy-cache-staging", ignoring the `S3_PREFIX` environment variable.

As it stands, the documentation does not make clear that the backend server requires these additional paths.  These require that the IAM policy used to grant the backend server access to the cache requires GetObject and PutObject (and so on) allowed actions for those "proxy-cache" related paths, along with the `S3_PREFIX`.

### Environment Information

- Deployment type: AWS ECS
- Operating System: Linux
- Architecture: Arm64

### Backend Version

1.7.6

### Web UI Version

1.8.0

### Package Format

_No response_

### Expected Behavior

All S3 locations exist under a deployment controlled S3 prefix.

### Observed Behavior

The backend server seems to assume that deployers dedicate the S3 bucket just to the service.

### Steps to Reproduce

1. Construct the IAM policy for the backend server to only have read/write access to the S3_PREFIX environment variable.
2. Access to the cache generates a warning due to no write or delete access to the bucket + path.

### Logs

```shell
2026-08-19T16:35:31.144915Z  WARN artifact_keeper_backend::services::proxy_service: proxy cache put_stream failed; cache will refetch next request cache_key=proxy-cache/maven-proxy/org/postgresql/postgresql/42.7.13/postgresql-42.7.13.pom/__content__ error=Storage error: Failed to start multipart upload for 'proxy-cache-staging/THE-ID': The operation lacked the necessary privileges to complete for path proxy-cache-staging/THE-ID: Error performing POST https://s3.us-east-1.amazonaws.com/S3_BUCKET/proxy-cache-staging/THE-ID?uploads= in 54.647681ms - Server returned non-2xx status code: 403 Forbidden: <?xml version="1.0" encoding="UTF-8"?>
```

### Additional Context

_No response_

### Pre-submission Checklist

brandonrc:
Closes #3454.

## What was wrong

Proxy-cache content is anchored at the storage **root**, never under `S3_PREFIX` — deliberately, because that anchoring is what makes a prefixed deployment read back the objects it wrote (#3368 / #3513). The consequence was that the entire namespace was `proxy-cache/<repo_key>/<path>/__content__`, with nothing in the key naming the deployment that produced it. Two deployments pointed at one bucket and separated only by `S3_PREFIX` shared **one** proxy-cache key space.

## Reproduced live (not just structurally)

Two backends built from `origin/main` (`1aa0aa3`), one MinIO bucket `shared`, separate databases, separate `S3_PREFIX` (`deploy-a` / `deploy-b`), separate upstreams:

| | deployment A (`:8117`) | deployment B (`:8118`) |
|---|---|---|
| `S3_PREFIX` | `deploy-a` | `deploy-b` |
| repository | `mvn-proxy` (id `7e920cd3…`) | `mvn-proxy` (id `d22b0227…`) |
| upstream serves | `ALPHA-BYTES-FROM-UPSTREAM-A-ONLY` | `BRAVO-BYTES-FROM-UPSTREAM-B-ONLY` |

The bucket held exactly **one** cached object for the coordinate:

```
proxy-cache/mvn-proxy/org/example/lib/1.0/lib-1.0.jar/__content__
```

Both directions confirmed:

* A fetched first → cached `ALPHA…`; B's request for the same path then returned **`ALPHA…`** — bytes B never fetched, from an upstream B is not configured to use.
* B fetched fresh → overwrote the same key with `BRAVO…`; A's next request then returned **`BRAVO…`** — cache poisoning against A's users.

Neither separate databases, nor separate `S3_PREFIX` values, nor IAM policies scoped to different artifact prefixes prevented it, because the cache is not addressed through any of them.

## The fix

One scope segment, placed **inside** the reserved namespace:

```
proxy-cache/<scope>/<repo_key>/<path>/__content__
```

Inside — rather than moving the namespace under `S3_PREFIX`, which #3454 explicitly warns against — is what keeps the blast radius small. `key_is_bucket_root_anchored` still holds (so #3368 stays fixed), `is_proxy_cache_key` and every `storage_key NOT LIKE 'proxy-cache/%'` accounting predicate still classify the key, and the IAM `Resource` list already documented in `.env.example` (`…/proxy-cache/*`) still covers it — **no policy needs re-cutting**. It also leaves reads that resolve a *recorded* key (`proxy_cache_artifacts.storage_key`, and legacy `artifacts` rows whose `storage_key` is a `proxy-cache/…` value) working unchanged, which is the "permanent failure for legacy rows" hazard the issue flagged.

**The identifier** is the deployment's persistent peer instance id — the `peer_instance_identity` UUID `init_peer_identity` seeds on first boot and never rewrites. Stable across restarts and upgrades; shared by every replica of one deployment (so a horizontally scaled deployment still shares one cache); not rewritten when `PEER_INSTANCE_NAME` or any other setting changes, so an unrelated config edit cannot orphan the cache. Unset is the intended configuration and it isolates. `PROXY_CACHE_SCOPE` overrides it with a legible token; there is deliberately **no value that turns scoping off**, and an invalid one fails startup rather than silently downgrading. The resolved scope is logged at startup (`Proxy cache scope: …`).

Every builder, listing prefix and prefix parser in the namespace moves together — key derivation, the key-length budget, `list_cached_paths`, `list_cached_pypi_packages`, `cached_artifact_paths`, `pypi_package_names_from_cache_keys`, the Prometheus repository label, and the Debian `__release_epoch__.json` sidecar — so a listing in one deployment cannot enumerate another's objects. `proxy-cache-staging/` stays unscoped: its keys are freshly minted UUIDs that cannot collide, and leaving it alone keeps the single documented lifecycle rule valid.

## Migration: cold cache, stated explicitly

A key-layout change cannot both isolate and keep addressing the old objects. A read-through fallback to the shared layout would preserve exactly the cross-deployment reads this issue is about, so this takes the **cold-cache** path instead: on first start after upgrading, every derived-key lookup misses once and refetches upstream, then warms normally. A proxy-cache miss is a refetch, not data loss, and nothing becomes unreachable — recorded-key reads are unaffected. **No operator action is required.** Full upgrade note, including how to reclaim the pre-upgrade tree safely and the caveat about legacy `artifacts` rows pointing into it, is in the CHANGELOG entry.

## Verification

**Live, post-fix, on the same rig** (both backends upgraded in place; the databases, repositories and bucket carried over from the repro above). Each deployment resolved its own scope from its existing `peer_instance_identity` row, with no configuration added:

```
deployment A  Proxy cache scope: 7ff7c6bf-e184-4d77-a541-0b23b208b3b4
deployment B  Proxy cache scope: c357fe27-6c70-479c-b345-b7c13ffc5374
```

The bucket now holds two disjoint subtrees instead of one shared object, with the pre-upgrade objects left untouched at the root exactly as the upgrade note describes:

```
proxy-cache/7ff7c6bf-…/mvn-proxy/org/example/lib/1.0/lib-1.0.jar/__content__
proxy-cache/c357fe27-…/mvn-proxy/org/example/lib/1.0/lib-1.0.jar/__content__
proxy-cache/mvn-proxy/org/example/lib/1.0/lib-1.0.jar/__content__          <- pre-upgrade, no longer read
```

A serves `ALPHA…` and B serves `BRAVO…`. Then **both upstream servers were stopped** and each request re-issued: A still answered `ALPHA…` and B still answered `BRAVO…`, so both were warm cache hits. That is the half that matters — the fix is isolation, not "turn every request into a cache miss".
* `--lib` tests (no `DATABASE_URL` needed, so they run in every job): a two-deployment-over-one-bucket end-to-end isolation test with its own shared-fixture control, plus per-surface tests for the listing, parsing, metric-label, epoch-key, key-length and purge hunks.
* Each production hunk was reverted individually and the corresponding test confirmed failing; the output is in the PR discussion.


brandonrc:
## Per-hunk revert proofs

Each production hunk was reverted **individually** (everything else left in place), the corresponding test re-run, and the source restored. All ten fail, and each failure message names the actual defect rather than an opaque mismatch.

| # | hunk reverted to | test | failure |
|---|---|---|---|
| H1 | `CacheKeys::derive` → `format!("{PROXY_CACHE_KEY_PREFIX}{repo_key}/")` | `deployments_sharing_a_bucket_do_not_serve_each_others_cached_bytes_3454`, `scoped_cache_keys_stay_classifiable_as_proxy_cache_content_3454` | `deployment B served deployment A's cached upstream bytes: Some("ALPHA-from-deployment-a-upstream")` / key `proxy-cache/maven-proxy/…` vs `proxy-cache/aaaa0000-…/maven-proxy/…` |
| H2 | `check_cache_key_length` → `PROXY_CACHE_KEY_PREFIX.len()` | `key_length_budget_charges_the_scope_segment_3454` | `a path one byte over the budget was accepted: proxy-cache/aaaa0000-…/r/ppp…/__content__ (1025 bytes)` |
| H3 | `cached_artifact_paths` prefix unscoped | `cached_artifact_paths_ignores_another_deployments_objects_3454` | `left: ["legacy"] right: ["is-odd"]` |
| H4 | `pypi_package_names_from_cache_keys` prefix unscoped | `pypi_simple_index_ignores_another_deployments_objects_3454` | `left: ["legacy"] right: ["requests"]` |
| H5 | `repo_key_from_cache_key` → `strip_prefix(PROXY_CACHE_KEY_PREFIX)` | `cache_metric_label_is_the_repository_not_the_scope_3454` | `left: "aaaa0000-0000-0000-0000-000000000001" right: "maven-proxy"` |
| H6 | `release_epoch_key` unscoped | `release_epoch_key_is_deployment_scoped_3454` | `left: "proxy-cache/debian-remote/dists/bookworm/__release_epoch__.json" right: "proxy-cache/aaaa0000-…/debian-remote/…"` |
| H7 | `purge_repo_cache` legacy sweep removed | `purge_repo_cache_sweeps_legacy_and_scoped_but_not_another_deployment_3454` | `expected the scoped and legacy objects to go — left: 1 right: 2` |
| H8 | `list_cached_paths` prefix unscoped | `cache_listings_enumerate_only_this_deployments_objects_3454` | `cached-artifact listing enumerated objects outside this deployment — left: [] right: ["mine"]` |
| H8b | `list_cached_pypi_packages` prefix unscoped | `cache_listings_enumerate_only_this_deployments_objects_3454` | `PyPI simple index advertised another deployment's cached projects — left: [] right: ["mine"]` |
| H9 | `from_env_and_identity` unset arm → `Self::unscoped()` | `unset_override_derives_a_scope_rather_than_falling_back_to_shared` | `left: ProxyCacheScope { root: "proxy-cache/", segment: None }` vs `root: "proxy-cache/11112222-…/"` |

### Why the isolation test is not a tautology

A test asserting "two scopes produce different keys" would be near-circular, so the headline test does not do that. `deployments_sharing_a_bucket_do_not_serve_each_others_cached_bytes_3454` stands up **two `ProxyService`s with their own `StorageService` handles rooted at one directory** — one shared bucket — writes through the production buffered cache-write path in deployment A, and then reads through deployment B's public `get_cached_artifact_by_path`, which derives its own key. It asserts three things: B misses, B's own write does not clobber A's object, and **A still hits its own warm cache** — that last one is what a "isolate everything by never hitting the cache" non-fix would fail.

`unscoped_layout_is_the_shared_key_space_3454_control` is the negative control: the *same* fixture with both deployments on the pre-#3454 layout, asserting B **does** see A's bytes. If that ever stops holding, the fixture has stopped sharing storage and the isolation assertion above has stopped proving anything.

### Where the tests run

All of them are `#[cfg(test)]` modules in the **lib** target, so `cargo nextest run --workspace --lib` executes them; none is an integration target subject to the `--test` allowlist. None requires `DATABASE_URL` (the fixtures use the existing lazy no-DB pool helper), so none can silently skip.

### Gates

* `cargo fmt --all -- --check` — clean
* `cargo clippy --workspace --lib --tests -- -D warnings` — clean
* `cargo nextest run --workspace --lib` against a migrated Postgres with `AK_TESTS_REQUIRE_DB=1` — **15911 passed, 0 failed**
* duplication via CI's own `jscpd-prepare-sources.py` path — **0.67%** against the 3% threshold
* no `sqlx::query!` changes (no `cargo sqlx prepare` needed) and **no migration** (nothing to collide in the ledger)

brandonrc:
Reviewed at `6f7e30ff89b05841c79550f7f831b9eba9ec0e0f` on a backend built from that SHA, filesystem storage, real HTTP.

**The core scoping design is sound**, and I could not break it. My own independent prefix-walker sweep found no production key builder the PR missed; storage GC's skip still matches scoped keys; S3 root anchoring holds for LIST so #3368 stays fixed; the scope cannot flip on restart (migration 036's `CREATE UNIQUE INDEX … ((1))` is a true singleton, so a replica race fails one startup rather than minting a second id); and the hollow-test hunt came up empty — the headline test uses a deliberately invalid `connect_lazy("postgres://invalid/")` pool so it cannot pass via a swallowed DB error, and its "A still hits its own cache" assertion is a real anti-trivialization control. Everything I verified is listed at the end.

The defect is not in the scoping. It is in the migration convenience bolted onto `purge_repo_cache`.

## F1 — HIGH, live-confirmed, introduced by this PR. Deleting one repository can purge the deployment's entire proxy cache

`proxy_service.rs:5686-5688`:

```rust
let mut prefixes = vec![self.cache_scope.repo_root(repo_key)];
let legacy = ProxyCacheScope::unscoped().repo_root(repo_key);   // -> "proxy-cache/<repo_key>/"
```

Called from `repositories.rs:4595` on every repository DELETE.

The legacy sweep reads `proxy-cache/<X>/` as "repository X's subtree". After this PR that path shape is **also** the scope root `proxy-cache/<scope>/`. `validate_repository_key` accepts `[A-Za-z0-9._-]{1,128}`, so both a UUID and a token like `prod-eu` are legal repository keys. Scope and repo key now occupy the same namespace level and the sweep cannot tell them apart.

Live, on a real backend at this SHA:

- `PROXY_CACHE_SCOPE=prod-eu`, cache seeded for `npm-remote`, `maven-central`, `prod-eu` → `DELETE /api/v1/repositories/prod-eu` → `purged=5`, **the whole `proxy-cache/` tree is empty**.
- **Default config, `PROXY_CACHE_SCOPE` unset**, scope = peer id `b8c10687-…`; create a repository keyed `b8c10687-…`, seed `npm-remote` + `maven-central`, delete it → `purged=2`, tree empty.

Also proven at unit level against the real `StorageService`/`FilesystemBackend`: deployment A deleting a repository keyed with **deployment B's** scope UUID deletes B's objects across all of B's repositories. On a shared bucket — the exact topology this PR exists to support — the victim's scope UUID is a top-level directory name, discoverable by anyone with the bucket credentials, which both deployments have by definition.

On `main` this cannot happen: `proxy-cache/prod-eu/` was only ever repository `prod-eu`'s own subtree. The PR creates the collision.

Fix options: separate the namespaces (`proxy-cache/_s/<scope>/…`), drop the legacy sweep, or refuse a scope segment that could be a repository key.

## F1b — MEDIUM, live-confirmed, same hunk. The automated sweep does what the upgrade note says needs two manual checks first

The CHANGELOG tells operators to delete the unscoped tree "only once every deployment sharing the bucket has upgraded… and check first for legacy `artifacts` rows pointing into that tree… which do still read from it."

`purge_repo_cache` now performs that deletion automatically on every repository DELETE, with neither check. Confirmed live: seeding `proxy-cache/npm-remote/…/__content__` unscoped and deleting repository `npm-remote` removed it (`purged=2`).

Those legacy rows are served by recorded key through `proxy_helpers.rs:4214/4252` and `pypi.rs:3271` — so on a shared bucket, deployment A's repository delete makes deployment B's still-referenced content permanently unreachable. That is precisely what the PR's headline claim rules out.

## F2 — MEDIUM. "10 hunks, each revert-proofed individually" is not true for 5 of them

I reverted all five handler hunks to `ProxyCacheScope::unscoped()` and ran the full `--lib` suite with `DATABASE_URL` and `AK_TESTS_REQUIRE_DB=1`:

| hunk | file:line | detected? |
|---|---|---|
| `try_member_cache_redirect` presign | `proxy_helpers.rs:1107` | no |
| `proxy_fetch_or_redirect` presign | `proxy_helpers.rs:1251` | no (dead code per its own comment) |
| `record_proxy_download` scope | `proxy_helpers.rs:5006` | no |
| `pypi_proxy_cache_redirect` presign | `pypi.rs:3562` | no |
| Debian epoch invalidation | `debian.rs:2113` | no |

15900 passed / 11 failed — 10 nuget `503 Too many concurrent uploads` load flakes and 1 panic from my own edit shifting bytes inside a hardcoded 1500-byte source-scan window. Pristine baseline at the same SHA: **15911 passed, 0 failed**. Nothing detected any of the five.

These are the highest-traffic live paths — every S3 presigned proxy download. A future refactor can silently reintroduce cross-deployment key derivation there with green CI.

**The other eight are genuinely proven** — I re-did that proof myself and each revert failed exactly its own test and no other: `CacheKeys::derive`, `cached_artifact_paths`, `pypi_package_names_from_cache_keys`, `list_cached_paths`/`list_cached_pypi_packages`, `release_epoch_key`, `repo_key_from_cache_key`, `check_cache_key_length`, and the purge sweep.

## F3 — MEDIUM-LOW, live-confirmed. "Nothing becomes unreachable" is wrong in a second way

The scope adds 37 bytes to every key, charged against the 1024-byte budget at `proxy_service.rs:5751`. Paths landing in the newly-lost window flip from *served* to a hard error, not to a refetch:

```
path len 946 -> HTTP 404 "Artifact not found upstream"   (proxied normally)
path len 956 -> HTTP 400 "Invalid artifact path"          (rejected before upstream)
```

Unit-confirmed the same path is `Ok` unscoped and `Err` scoped. It fails cleanly — no truncation, no collision — but the note's "no correctness change… nothing becomes unreachable" is false for paths of roughly 947–990 bytes. Rare, but the note should say so, or `MAX_STORAGE_KEY_BYTES` should absorb the scope.

## What I could not break

My independent sweep found every `proxy-cache/` literal, `format!`, `starts_with` and `strip_prefix` outside `proxy_service.rs`/`proxy_cache_scope.rs` is in `#[cfg(test)]` code — I verified test-module membership individually for sbt, nuget, security, admin_security, maven, repositories, admin and oci_v2. All **13** `NOT LIKE 'proxy-cache/%'` predicates plus migration 171 are classification-only and still match (the PR said 12; it is 13, and all are safe). Only three production listings exist for the namespace and all three are scoped — there is no full-bucket walker. `proxy_cache_metadata_key_for` is pure suffix surgery on a recorded key and correctly scope-agnostic. `record_proxy_download`'s catalog upsert does not overwrite `storage_key` on conflict, so post-upgrade legacy rows are not rewritten to point at objects that do not exist yet. An invalid `PROXY_CACHE_SCOPE` does fail startup as claimed, though only after the full boot sequence. And the new `let Some(proxy) = … else { return }` guard in `record_proxy_download` is reachable only from paths already holding a live proxy — I checked all 21 call sites.

## Recommendation

F1 blocks merge. A cross-deployment *destructive* regression is worse in kind than the cross-deployment *read* bug being fixed, and it is reachable on default configuration. F1b rides along with the same fix. F2 should be closed by adding proofs for the five presign/scope hunks rather than by amending the claim — those are the paths most likely to regress silently.

brandonrc:
Live red-team verification on an isolated instance of this exact build (`pr3609`), run **blind** — the operator was told only that the proxy-cache storage layout had changed, and asked to consider both reading and destruction. It reproduced the purge collision and established two things the earlier review did not.

## First, the good news: the reading isolation this PR exists to fix genuinely works

Verified independently. Every proxy-cache key carries the per-repository segment, so two repositories pointed at the same upstream path cache to physically distinct objects; a private and a public repo cannot serve each other's bytes, and the cache listings are prefixed by `<scope>/<repo_key>/` so there is no cross-repo listing leak. A repository whose key equals the scope reaches only `proxy-cache/<scope>/<scope>/…` for reads — **the collision is destructive-only**. No spurious cache miss was observed for legitimate reads; the victim served a warm hit before the attack.

The cross-deployment read and poison that this PR targets is genuinely closed.

## The defect is exploitable by a NON-ADMIN with no rights on the victim

This is the part the earlier review did not establish. User `mallory`, `is_admin=false`, holding **only** `repository:admin` on the colliding repository — a normal delegated team-lead grant:

```
DELETE victim-proxy                     -> 403   (mallory has no rights on the victim)
DELETE <the colliding repo>             -> 200
victim-proxy cached content files       -> 1 -> 0
```

So the deleting principal needs zero rights on the repositories whose cache is destroyed. That is a cross-tenant destructive isolation break, not just an operator footgun.

## And the loss is durable, not a refetch

The upstream was stopped before the attack, so the cache was the only copy:

```
before:  GET .../lib-1.0.jar -> 200, ALPHA-BYTES-FROM-UPSTREAM-VICTIM-ONLY-v1   (warm hit, upstream down)
attack:  create repo key = a989b6bd-ee9c-48a9-be48-7bf1e3c15a11  (== deployment scope), DELETE it
after:   on-disk __content__ and __cache_meta__ for victim-proxy DELETED (0 content files)
         GET .../lib-1.0.jar -> 502 "Failed to fetch from upstream"
```

When an upstream is unreachable, rate-limited, or the artifact has been yanked, this is data loss and a serving outage — not the cache miss the upgrade note describes.

## Severity depends on a configuration choice you already support

On the default the scope is the `peer_instance_identity` UUID, logged at startup as `Proxy cache scope: <uuid>`. Notably it is **not** the `/api/v1/peers` self-peer id — those differed — so on the default it is semi-secret, readable from logs or the DB, and the operator rated it HIGH.

But `PROXY_CACHE_SCOPE` is a documented, supported config, and a legible token like `prod-eu` is itself a valid repository key and trivially guessable. Set that, and this becomes **CRITICAL with no secret required**.

## Root cause, unchanged from the earlier review

```rust
let mut prefixes = vec![self.cache_scope.repo_root(repo_key)];   // proxy-cache/<scope>/<repo_key>/
let legacy = ProxyCacheScope::unscoped().repo_root(repo_key);    // proxy-cache/<repo_key>/   <-- unguarded
if !prefixes.contains(&legacy) { prefixes.push(legacy); }
```

When `repo_key == <scope>`, `legacy` collapses to `proxy-cache/<scope>/`, the root of the entire deployment's cache.

Two fixes worth considering together: skip the legacy sweep when `repo_key == self.cache_scope.segment()`, **and** refuse a repository key equal to the active scope segment at creation. The second is the one that survives a future refactor.

## Method note

SSRF correctly blocked the private-IP upstream at repo-create, so the operator recreated the backend once with `UPSTREAM_ALLOW_PRIVATE_IPS=true` — a per-slot env the compose already exposes — and restored it afterwards. That changes only which upstream hosts are permitted and is orthogonal to the key-layout and purge logic under test. `peer_instance_identity` persists in the slot's DB, so the scope was identical before and after.

brandonrc:
Follow-up pushed (rebased onto `main` `dd87b556`): closes the scope-segment purge-collision defect and adds the missing revert-proofs.

### Defect fixed (two halves)
1. **Guard** — `purge_repo_cache` now skips the legacy unscoped prefix when `repo_key == cache_scope.segment()`, so it can no longer collapse `proxy-cache/<scope>/<repo_key>/` onto the whole-deployment root `proxy-cache/<scope>/`.
2. **Refuse at creation/rename** — `validate_key_not_scope_collision` rejects a repository key equal to the active scope segment (structural, survives a purge-path refactor). Base validator still accepts the key, so the check is load-bearing.

### Per-hunk revert-proof (revert → the mapped `--lib` test fails)
Each production hunk was reverted to `ProxyCacheScope::unscoped()` individually; the mapped test fails with:

| hunk | test | failure |
|---|---|---|
| `purge_repo_cache` guard | `purge_of_repo_keyed_as_scope_segment_spares_other_repos_3454` | `assertion left == right failed … left: 4  right: 1` (the whole deployment cache was swept) |
| `validate_key_not_scope_collision` | `test_reject_repository_key_equal_to_scope_segment` | panic: the colliding key was accepted |
| `try_member_cache_redirect` scope | `try_member_cache_redirect_signs_the_scoped_key_3454` | `member redirect signed a non-scoped key … /proxy-cache/maven-proxy/…` |
| `proxy_fetch_or_redirect` scope | `proxy_fetch_or_redirect_signs_the_scoped_key_3454` | `generic redirect signed a non-scoped key … /proxy-cache/npm-proxy/…` |
| `record_proxy_download` scope | `record_proxy_download_persists_scoped_storage_key_3454` | `placeholder catalog row keyed under a non-scoped key: proxy-cache/ph-test-npm-…/…` |
| `pypi_proxy_cache_redirect` scope | `test_pypi_proxy_cache_redirect_signs_scoped_key_3454` | `pypi redirect signed a non-scoped key … /proxy-cache/pypi-remote/…` |
| Debian epoch (`maybe_invalidate_by_epoch`) | `maybe_invalidate_by_epoch_uses_scoped_key_3454` | `epoch invalidation missed the scoped cache entry` |

The destructive test asserts an unrelated victim repository's object count is unchanged (before/after) — not merely that two scopes differ.

### Gates
`fmt` clean · `clippy --lib --tests -D warnings` clean · full `--lib` **15919 passed / 0 failed** with `AK_TESTS_REQUIRE_DB=1` against a migrated Postgres · jscpd (min-lines 10) 0 clones.

### Live verification (filesystem-backed deployment, scope = peer-instance UUID)
- **Pre-fix image**: created a repo keyed as the scope UUID, seeded a victim repo (2 objects) + the colliding repo (1). Deleting the colliding repo drove the victim **2 → 0** (durable cross-repo destruction; the WARN sweep log listed `proxy-cache/<scope>/victim-proxy/…` deletes).
- **Fixed image**: creating a repo keyed as the scope UUID → `400 … 'is reserved: it collides with this deployment's proxy-cache scope segment'`. Deleting a *pre-existing* scope-keyed repo (inserted directly, i.e. one that predates the check) purged only its own object (`purged=1`); victim stayed **2 → 2**. A normal repository delete still purges its own cache and nothing else (`normal 2 → 0`, `purged=2`, victim untouched).

### Upgrade note correction
The scope segment is charged against the 1024-byte object-store key budget (`MAX_STORAGE_KEY_BYTES`, the store's own hard limit — it cannot be raised). The default UUID scope adds 37 bytes, so paths within ~37 bytes of the pre-upgrade ceiling (~947–990 bytes) now return a deterministic `400`, not a refetch. Documented in the CHANGELOG upgrade note.

Issue #1368, opened 2026-05-24
[bug] Malformed JSON bodies return Axum 422 instead of our 400 envelope


brandonrc:
Surfaced by release-gate Full Suite for v1.2.0-rc.2 (run 26344757642). 4 repo-tests in `virtual-repo-malformed-input` expect HTTP 400 but the server returns HTTP 422.

### Failing assertions
- `PUT /members with member missing member_key returns 400` — got 422.
- `PUT /members with priority as string returns 400` — got 422.
- `PUT /cache-ttl missing cache_ttl_seconds returns 400` — got 422.
- `PUT /cache-ttl with cache_ttl_seconds as string returns 400` — got 422.

### Root cause
422 is Axum's default for body-deserialization failures via `serde::Deserialize`. The handlers don't run; the extractor rejects.

### Decision needed
Two options:
1. **Server change**: install a custom rejection handler on the affected routes that maps `JsonRejection::*` to 400 with `code: "VALIDATION_ERROR"` for consistency with our other handlers.
2. **Test change**: accept 400 OR 422 in these tests.

Recommend Option 1 — it makes the contract consistent with the rest of our error envelope and is the behavior the SDK clients expect.

### Acceptance
- All four assertions pass.
- API docs explicitly state 400 for malformed bodies, no 422 anywhere in the OpenAPI spec.

Part of #1356.

Issue #708, opened 2026-04-12
Trivy: "Failed to extract archive" error causes scans to not work


Firjens:
### Description

Hello.

Scans are not working on Trivy because the backend is failing to extract archives, causing Trivy scans to fail open (and marking everything as having zero issues).

### Environment Information

- Deployment type: k8s
- Operating System: Alma
- Architecture: AMD64
- Database: Postgres
- Reverse Proxy: nginx
- Network setup: N/A

### Backend Version

1.1.2

### Web UI Version

1.1.0-rc8

### Package Format

N/A / General

### Expected Behavior

The backend should unpackage the archives to permit Trivy to scan the files

### Observed Behavior

The backend fails to unpackage the archive, causing Trivy to fail to scan anything

### Steps to Reproduce

1. Configure a remote repository with e.g. NPM
2. Pull a package through the cache
3. Request a manual scan
4. Observe the logs in the backend container

### Logs

```shell
026-04-12T20:30:49.765397Z  INFO artifact_keeper_backend::services::scanner_service: Scan incus-image completed for artifact 07b76bb0-1430-464f-8561-d5f635669aa2: 0 findings (0 critical, 0 high)                                                                                     │
│ 2026-04-12T20:30:49.775272Z  INFO artifact_keeper_backend::services::grype_scanner: Starting Grype scan for artifact: lightningcss-linux-arm-gnueabihf-1.32.0.tgz (07b76bb0-1430-464f-8561-d5f635669aa2)                                                                                │
│ 2026-04-12T20:30:49.782442Z  WARN artifact_keeper_backend::services::grype_scanner: Failed to extract archive lightningcss-linux-arm-gnueabihf-1.32.0.tgz: Internal error: Failed to execute extraction command: No such file or directory (os error 2). Scanning raw file instead.     │
│ 2026-04-12T20:30:51.209587Z  WARN artifact_keeper_backend::services::grype_scanner: Grype scan failed for lightningcss-linux-arm-gnueabihf-1.32.0.tgz: Internal error: Grype scan failed (exit exit status: 1): . Returning empty findings.                                             │
│ 2026-04-12T20:30:51.213235Z  INFO artifact_keeper_backend::services::scanner_service: Scan grype completed for artifact 07b76bb0-1430-464f-8561-d5f635669aa2: 0 findings (0 critical, 0 high)                                                                                           │
[...]

PR #3619, opened 2026-09-01
fix(storage-gc)!: require dry_run and reject unknown fields on storage-GC requests

### LINKED ISSUE #3501: Decision: storage-GC dry_run defaults to destructive, so {} and a typo'd key both perform an irreversible delete
## Decision request (not a bug report)

`StorageGcRequest.dry_run` is `#[serde(default)]` on a `bool`, so **an unrecognised or absent field means a live, irreversible delete.** Whether that is the right default is a product decision, and there is a test that deliberately pins the current behaviour — so this should be decided, not quietly patched.

## Current behaviour

`backend/src/api/handlers/storage_gc.rs:52-57`:

```rust
#[derive(Debug, Deserialize, ToSchema)]
pub struct StorageGcRequest {
    /// When true, report what would be deleted without actually deleting.
    #[serde(default)]
    pub dry_run: bool,
}
```

Consequences on `POST /api/v1/admin/storage-gc` (and the repo-scoped `POST /api/v1/repositories/{key}/storage-gc`, `storage_gc.rs:158`):

| request | effective `dry_run` | result |
|---|---|---|
| `{}` | `false` | live delete |
| `{"dryRun": true}` (camelCase, wrong key) | `false` | **live delete** |
| `{"dry_Run": true}`, `{"dryrun": true}`, any typo | `false` | **live delete** |
| `?dry_run=true` with body `{}` (query param, not read) | `false` | **live delete** |
| `{"dry_run": "yes"}` | — | 400, rejected (`test_storage_gc_request_invalid_dry_run_type`) |

So the only misuse that fails safe is a wrong *type*. A wrong *key name* silently selects the destructive branch. During the 1.8.2 campaign a tester deleted a batch of attribution rows this way by accident, mid-session, from a typo'd key.

The operation has no undo. `run_gc` deletes storage objects and the rows that point at them; there is no tombstone and no restore path short of a backup.

## Why this is a decision and not a patch

`storage_gc.rs:277-281` explicitly tests the lenient parse:

```rust
#[test]
fn test_storage_gc_request_extra_fields_ignored() {
    let req: StorageGcRequest =
        serde_json::from_str(r#"{"dry_run": true, "unknown_field": 42}"#).unwrap();
    assert!(req.dry_run);
}
```

Someone decided unknown fields are ignored, and wrote a test to keep it that way. Both candidate fixes below change a deliberately-tested contract and change the API's behaviour for existing callers, so they need an explicit call and a changelog entry — not a drive-by edit.

## Options

1. **Default `dry_run` to `true`.** `#[serde(default = "default_true")]`. A caller that means it must say so. Safest, and matches the "destructive operations are opt-in" convention the codebase already applies to `MAVEN_FLAT_GC_ENABLED` and `BLOB_GC_ENABLED`.
   *Breaks:* any existing automation posting `{}` expecting a live run — it would silently become a no-op, which is the benign failure direction but still a behaviour change.
2. **Make the field required.** Remove `#[serde(default)]`; `{}` becomes a 400. No silent change of meaning in either direction, and the error tells the caller what is missing.
   *Breaks:* the same callers, loudly rather than silently. Arguably better.
3. **`#[serde(deny_unknown_fields)]`**, with or without 1 or 2. Turns `{"dryRun": true}` into a 400 instead of a live delete. Directly contradicts `test_storage_gc_request_extra_fields_ignored`, so that test would be inverted.
4. **Do nothing**, and instead document the hazard in the OpenAPI description and require an explicit confirmation parameter for the live path.

Option 2 + 3 together give the least surprising API: `{}` → 400 "dry_run is required", `{"dryRun":true}` → 400 "unknown field", `{"dry_run":false}` → live. Both endpoints are already admin-gated, so the blast radius is limited to admins — but "admin typo'd a key" is exactly the scenario that produced the incident.

Whichever is chosen, the web UI's per-repo storage panel (which posts `{dry_run: true}` for its "reclaimable now" estimate, `storage_gc.rs:126`) already sends the field explicitly and is unaffected.

## Related

- #3431 — the incident that established the opt-in-gate convention for the Maven flat-object sweep.
- #3384 / #3074 — the endpoints in question.

Raised during 1.8.2 milestone verification; behaviour re-verified against `main` @ `a9f3876` on 2026-08-22.

brandonrc:
## Summary

Closes #3501 (decision issue, milestone 1.9.0).

`StorageGcRequest.dry_run` was `#[serde(default)]` on a `bool` and unknown body fields were ignored, so on `POST /api/v1/admin/storage-gc` and `POST /api/v1/repositories/{key}/storage-gc` an empty body `{}` — or **any misspelling** of the one field that could have said "dry run" (`{"dryRun": true}`, `{"dry_Run": true}`, `{"dryrun": true}`) — silently selected the destructive branch and performed a live, irreversible delete. The 1.8.2 milestone verification lost a batch of attribution rows to exactly this (a typo'd key on a request whose author had asked for a dry run).

**Decision taken: option 2 + 3 from the issue** — `dry_run` is now **required** and unknown fields are **rejected** (`#[serde(deny_unknown_fields)]`). An ambiguous request is refused with a request-validation error naming the missing/unrecognised field (axum's `Json` rejection, HTTP 422 with e.g. ``missing field `dry_run` `` / ``unknown field `dryRun`, expected `dry_run` ``), and **nothing is deleted**. Explicit requests are byte-for-byte unchanged: `{"dry_run": true}` estimates, `{"dry_run": false}` deletes.

Why reject rather than flip the default to `dry_run: true` (option 1): a flipped default silently turns an existing automation's `{}` live run into a no-op — the benign failure direction, but still a silent meaning change that would be discovered as "GC stopped reclaiming space" much later. A refusal breaks loudly at the call site, once, with an error that says exactly what to send, and can never be mistaken for either GC outcome. `test_storage_gc_request_extra_fields_ignored`, which deliberately pinned the lenient parse, is inverted accordingly.

Also in this PR: the verbatim-duplicated post-GC stats-refresh block shared by the two handlers is extracted into `refresh_stats_after_live_gc` (behavior-preserving; it was this file's one production clone, flagged by the post-#3523 duplication gate).

Deliberately NOT in this PR: migrating these handlers from `axum::Json` to the #1368 `crate::api::extractors::Json` envelope (which would render the refusal as the standard 400 `VALIDATION_ERROR` envelope). That swap forces a compile-fix in `storage_gc_service.rs`'s test module, and any touch to that file trips the duplication gate on ~14 pre-existing production clones (6.1%, all pre-dating this change — from #3072/#1655/#1448). The envelope migration should ride along whenever those clones are deduplicated.

### Upgrade note (also in CHANGELOG under Unreleased/Changed, marked BREAKING)

Any automation calling these endpoints with an empty or field-less body must now send `{"dry_run": false}` explicitly to keep performing live GC. Until it does it receives a 422 naming the missing field — and **no deletion has happened**. The web UI's per-repo storage panel already sends the field explicitly and is unaffected.

## Live verification (isolated slot, image built from this branch)

Before (1.8.1), with one seeded soft-deleted orphan each time:

| request | result |
|---|---|
| `{}` | **200** `{"dry_run":false,...,"artifacts_removed":1}` — orphan rows 1 → 0 (live delete) |
| `{"dryRun": true}` | **200** `{"dry_run":false,...,"artifacts_removed":1}` — live delete |
| `{"dry_run": false}` | 200, live delete (intended) |

After (this branch):

| request | result |
|---|---|
| `{}` | **422** ``missing field `dry_run` `` — orphan rows unchanged |
| `{"dryRun": true}` | **422** ``unknown field `dryRun`, expected `dry_run` `` — unchanged |
| repo-scoped `{}` | **422**, unchanged |
| `{"dry_run": true}` | 200, `dry_run:true`, rows unchanged |
| `{"dry_run": false}` | 200, orphan reclaimed (rows 1 → 0) |

`storage_gc_route_refuses_ambiguous_requests_without_deleting` (DB-backed, `--lib`, honors `AK_TESTS_REQUIRE_DB=1`) drives the **real admin route** end-to-end: seeds a reclaimable soft-deleted orphan (row + object), posts `{}` and `{"dryRun": true}`, asserts the refusal names the offending field **and** that the probe object + artifact row survive, then asserts the explicit `{"dry_run": false}` still reclaims the probe (positive control proving the non-deletion asserts are not vacuous). Each hunk was revert-proofed separately: restoring `#[serde(default)]` fails it with `left: 200 / right: 422` (`{}` live-deleted again); dropping `deny_unknown_fields` fails it plus the unknown-field/camelCase serde tests. Serde-level tests (`test_storage_gc_request_empty_body_rejected`, `..._unknown_field_rejected`, `..._camel_case_key_rejected`) additionally pin that every rejection names the offending field.

A DTF tier (`storage-gc-dryrun-required`, RED 6-fail on 1.8.1 → GREEN 8/8 on this branch, validated locally) will land in artifact-keeper-test with the batched 1.9.0 tiers.

## Test Checklist
- [x] Integration tests added/updated (if applicable) — DB-backed route-level test
- [x] E2E tests added/updated (if applicable) — DTF tier authored, batched for the 1.9.0 test-repo MR
- [x] Manually tested locally
- [x] No regressions in existing tests — full `--lib` suite 15986/15986 with a live DB (`AK_TESTS_REQUIRE_DB=1`), fmt, clippy `-D warnings` clean, duplication 0% on the changed file measured the CI way (test modules stripped)

## API Changes
- [x] Request/response types have `#[derive(ToSchema)]` — `dry_run` now marked required in the schema
- [ ] Migration is reversible (if applicable) — no migration


brandonrc:
## Audit round 1 — `fix(storage-gc)!` (2 reviewers: correctness + API contract)

**Bottom line: the core fix is correct and complete.** I tried to find a remaining path from an ambiguous body to the destructive branch and could not. Verified positively:

- `StorageGcRequest` (`storage_gc.rs:62-68`) has one field, no `flatten`/`alias`/`rename_all`, so the serde `deny_unknown_fields`+`flatten` incompatibility doesn't apply. utoipa 5.5 does honour container-level `deny_unknown_fields` (`utoipa-gen-5.5.0/src/component/serde.rs:192`), so the published schema gains `required: ["dry_run"]` + `additionalProperties: false` — spec tracks code.
- **Both** endpoints are covered (`:83-87`, `:178-183`), `Json` last in both extractor lists. Grepping every non-test call site of `.run_gc(` / `run_gc_for_repository(` yields only these two plus the scheduler (`storage_gc_service.rs:605`, hardcoded `false`, not HTTP-reachable). No third path.
- **The helper extraction is genuinely behavior-preserving.** `origin/main:89-97` vs `origin/main:185-193` diff byte-for-byte identical; `refresh_stats_after_live_gc` inverts the guard (equivalent), same `StorageStatsService::new`, same `recompute_all()`, same swallow-and-`warn!`, awaited at the same position before `Ok(Json(result))` at both sites. No error changed propagation direction, no dead code left.
- No in-repo caller breaks: `scripts/native-tests/test-storage-gc.sh:55,100,126,176` all send explicit `{"dry_run": …}` with no extra keys. There is no UI code in this repo.
- `cargo check --tests -p artifact-keeper-backend` at head: exit 0.

Findings below are secondary. None argue against the change itself.

---

### 1. 🔴 The one branch the new test skips is the one a mutation would silently break

Change `:92` from `service.run_gc(payload.dry_run)` to `service.run_gc(false)` — **every test still passes.** The serde tests (`:276-334`) only exercise parsing; the route test (`:362`) posts `{}`, `{"dryRun": true}` and `{"dry_run": false}` but **never `{"dry_run": true}`**, so the estimate branch is never observed end to end. The only other handler-level test (`storage_gc_service.rs:4257`) also passes `false`.

Failure: a later refactor drops the `payload.dry_run` argument, CI is green, and the web UI's "reclaimable now" estimate (`:148`) starts hard-deleting on every panel load. That is precisely the bug class this PR exists to prevent, on the one path it doesn't cover.

**Fix:** add a `{"dry_run": true}` step before the positive control, asserting the probe object and row both survive. Two lines.

### 2. 🔴 "HTTP 422" is wrong for the two shapes an operator is most likely to hit

Measured against axum 0.7.9 with this exact struct (probe built and run, not read from docs):

| request | status | body |
|---|---|---|
| `{}` (CT json) | **422** | `missing field \`dry_run\`` |
| `{"dryRun": true}` | **422** | ``unknown field `dryRun`, expected `dry_run` `` |
| **empty body, zero bytes** | **400** | `EOF while parsing a value` |
| **no `Content-Type` header** | **415** | `Expected request with Content-Type: application/json` |
| `Content-Type: text/plain` | **415** | same |
| `{"dry_run": null}` | 422 | `invalid type: null` — importantly *not* coerced to false |
| duplicate `dry_run` keys | 422 | ``duplicate field `dry_run` `` — order-independent |

`curl -X POST $URL -H "Authorization: …"` with no body → **415**. `curl -X POST -d '{}'` without an explicit content-type header → curl sends `application/x-www-form-urlencoded` → **415**. Neither is 422, and neither body mentions `dry_run`.

So an operator who reads the changelog and greps their logs for 422 misses most of their broken call sites, and the 415 body gives them no hint the required field is the cause.

The `null` and duplicate-key results are the reassuring ones — neither silently selects the destructive branch, which is the property actually being defended.

### 3. 🟠 Do the `extractors::Json` swap in this PR rather than deferring it

The raw counts do **not** support "the repo overwhelmingly returns 400" — they support the opposite, and I'll give them straight:

| extractor | body-extraction sites | files |
|---|---|---|
| bare `axum::Json` → 422 + `text/plain` | **124** | 40 |
| `crate::api::extractors::Json` → 400 `VALIDATION_ERROR` | **20** | 4 |

#1368 is ~14% migrated. Three things make the deferral wrong anyway:

- **The stated contract is 400.** `extractors.rs:6-8`: *"Our public API contract is that any client error returns 400 Bad Request with a structured envelope."* 422 is un-migrated legacy, not a competing convention.
- **The split lands inside one namespace.** `run_repository_storage_gc` is merged into the repositories router (`routes.rs:600`), and `repositories.rs` is one of the four already-migrated files. After merge, `PUT /repositories/{key}/cache-ttl` returns `400 {"code":"VALIDATION_ERROR"}` and `POST /repositories/{key}/storage-gc` returns `422 text/plain` — same prefix, same client, same error handler.
- **The PR already wrote 400 twice.** `CHANGELOG.md:91` says "HTTP 422" and one sentence later argues "whereas a **400** breaks loudly at the call site." The new test doc comment (~`:315`) says the rejection "must be a **400**-shaped rejection" while asserting `UNPROCESSABLE_ENTITY`. The intended contract is 400; only the compile-fix blocked it.

This is a `!` PR with a changelog entry — the one moment clients are already forced to touch this endpoint. Deferring means shipping 422 in 1.9.0 and then 422→400 as a *second* breaking change. And per finding 2, the swap collapses 422/400/415 into one 400 `VALIDATION_ERROR` (its own tests pin exactly that: `extractors.rs:534`, `:550`, `:586`).

The blocker is a lint gate on ~14 pre-existing clones in a *test* module, not a design problem: exempt `storage_gc_service.rs` in `.jscpd.json` (`ci.yml:823` already calls that file the source of truth and the gate deliberately scoped), or land the compile-fix as a separate no-op PR first.

### 4. 🟠 Neither `#[utoipa::path]` documents the rejection status

`:78-80` lists only 200; `:171-175` lists 200/401/404. Repo-wide there are 53 `status = 400` and 20 `status = 422` entries, so documenting client errors is established practice. `sync-openapi-spec.yml` exports the spec on every release and pushes it to a downstream API repo — an undocumented rejection means generated clients get the `required` field but no typed error branch, and retry the 422 as a transport fault.

Note the precedent set by the previous breaking entry at `CHANGELOG.md:12` (#2795/#2899): *"the behavior is now stated in the OpenAPI operation description and pinned by a regression test."* This PR does the test but not the description.

### 5. 🟠 The upgrade note covers only half the break

`CHANGELOG.md:93` tells operators an empty/field-less body must become `{"dry_run": false}`. It says nothing about `deny_unknown_fields`.

Failure: a nightly cron posting `{"dry_run": false, "reason": "nightly-gc"}` — a working live GC today — gets 422 after upgrade, deletes nothing, and `curl -sf` (this repo's own script pattern) exits non-zero into a job nobody watches. Storage silently stops being reclaimed. This is the *more* likely break, because a caller already passing `dry_run` is the one that got past the first half of the note.

Also worth one sentence: `additionalProperties: false` is now published, so future optional fields can't be sent to an older server, and strict generated clients enforce it request-side. Defensible for a destructive endpoint — just say it.

### 6. 🟡 Unauthenticated callers to the per-repo endpoint now get 422 where they got 401

`Json` extraction runs before `require_auth`/`require_admin` (`:184-187`), so an anonymous `POST /repositories/{key}/storage-gc` with `{}` returns 422 naming `dry_run` instead of the documented 401. Nothing is deleted and no repo existence leaks, so it's schema disclosure only — but it widens a surface this file explicitly cares about (the probe-resistance comment at `:185-186`), and the `(status = 401)` line at `:173` is now unreachable for any ambiguous body. Either reorder (admin `FromRequestParts` extractor ahead of `Json`) or stop claiming 401.

### 7. 🟡 The test's own revert-proofing comment is factually wrong

`:355-357` claims dropping `deny_unknown_fields` makes `{"dryRun": true}` "deserialize to `dry_run: false` and delete." It would not — with `dry_run` required, serde reports `missing field \`dry_run\`` and axum still returns 422 with nothing deleted. The protection *does* hold (both `test_storage_gc_request_unknown_field_rejected` at `:303` and the route test's `body.contains("dryRun")` at `:468` fail under that mutation) — the stated mechanism is just wrong, which will mislead the next person judging whether the attribute is load-bearing.

### 8. 🟡 The route test is a silent no-op without Postgres

`:369` returns early when `tdh::Fixture::setup` yields `None`. In any lane without a DB the whole "nothing was deleted" guarantee reports green having asserted nothing. The pure serde tests still catch both mutations, so this is a caveat on the *stated* revert-proofing rather than a hole — but the comment at `:352-358` promises more than a DB-less run delivers.

### 9. 🟡 The web-UI claim can't be verified from this repo

`CHANGELOG.md:95` asserts *"The web UI's per-repo storage panel already sends the field explicitly and is unaffected."* There is no frontend in this repo (the UI is separate, per the "web #708" reference at `:40`). Confirm it against the web repo and link the line, or drop the sentence — an unverified "unaffected" in a breaking-change note is worse than no sentence.

### 10. 💭 One identical latent hazard survives, outside this file

No other struct in `storage_gc.rs` has the destructive-default shape (`OciBlobReportQuery` at `:222-227` is read-only `Option<i64>`). But `MigrationConfig.dry_run` at `backend/src/models/migration.rs:130-131` is the same `#[serde(default)] pub dry_run: bool`, reachable as a request body via `CreateMigrationRequest` (`migration.rs:275`, extracted at `:1140`) — so `{"config": {}}` means "really migrate." Lower stakes (writes, not deletes), same pattern; worth a follow-up issue. The `skip_policy_check` defaults in `approval.rs:55,91` and `promotion.rs:52,63` are safe — their `false` default is the conservative branch.

---

**Suggested gating before un-drafting:** finding 1 (two-line test), finding 2+5 (changelog accuracy), and a decision on finding 3. Findings 4, 6, 7 are cheap. 8, 9, 10 are follow-ups.

<sub>Round 1 of an automated multi-reviewer audit (correctness + API-contract passes, findings cross-checked against the PR head and measured against the pinned axum/utoipa versions).</sub>

brandonrc:
## Adversarial review + verification (out of draft)

Reviewed this as if it were someone else's PR, since nothing else had looked at it. Summary: the change does what it claims, the breaking change is justified, and it broke one first-party caller that the PR body did not name. Two defects found and fixed on the branch; one residual noted below.

### Does requiring `dry_run` actually prevent #3501's failure mode?

Yes — it prevents it, it does not merely make it noisier. The incident shape (`{"dryRun": true}`: the caller *asked for a dry run*) is now refused at the extractor, before the handler body, so `run_gc` is never reached and nothing is deleted. Confirmed end-to-end against the real router, not inferred from the status code: the probe object and its `artifacts` row both survive both refusals, and the explicit `{"dry_run": false}` still reclaims them (positive control, so the survival asserts are not vacuous).

`?dry_run=true` from the issue's table is covered as a side effect: the query param is still not read, but the body is now required, so `?dry_run=true` with `{}` is a 422 rather than a live delete.

**Residual, deliberate:** the destructive value is still the falsy one. `{"dry_run": false}` written by someone who means "nothing special" still deletes. Option 4 from the issue (a separate confirm parameter on the live path) would close that; this PR does not, and I think that is the right scope split — say so on #3501 rather than widening this.

### What the breaking change actually breaks

Every caller in this repo and in the org, checked by grep + reading each body:

| caller | body sent | verdict |
|---|---|---|
| `scripts/native-tests/test-storage-gc.sh:55,100,126,176` (e2e `storage-gc` profile) | `{"dry_run": true}` / `{"dry_run": false}` | safe |
| `artifact-keeper-web` `src/lib/api/storage.ts:95-102` → `repo-storage-panel.tsx:96` | `{"dry_run": <bool>}`, single key; `fetch.ts:66-85` injects no body fields | safe |
| `artifact-keeper-test` `tests/lifecycle/test-oci-gc-mark-sweep.sh:217`, `tests/lifecycle/test-storage-gc.sh:18`, `deploy-test/harness/tiers/maven-rollup-anchor-escape/oracle.sh:307`, `deploy-test/harness/tiers/storage-prefix-rollup/oracle.sh:60` | explicit `dry_run` | safe |
| `artifact-keeper-iac`, `webinar-demo`, `docker-desktop-extension`, `artifact-keeper-ios` | no caller | n/a |
| `artifact-keeper-site` `src/content/docs/docs/advanced/storage.mdx:666-675` | documented curls already send `{"dry_run": true}` | safe (docs) |
| **`artifact-keeper-cli` `src/commands/admin.rs:3982-3984`** | `dry_run: dry_run.then_some(true)` → `None` when `--dry-run` is absent → SDK `skip_serializing_if` drops the key → **`{}`** | **breaks** |

**The CLI break is the worst possible one: it is exactly the live path.** `ak admin storage-gc run --dry-run` keeps working; `ak admin storage-gc run` (live GC) now gets 422. Companion fix is one line — `dry_run: Some(dry_run)` — plus a CLI release. I have added it to the CHANGELOG upgrade note so it cannot ship unmentioned; it needs a PR on `artifact-keeper-cli` before or with 1.10.0.

Two SDK surfaces would serialize `{}` for a consumer, though neither has a caller today: `artifact-keeper-android` `sdk/.../models/StorageGcRequest.kt` (`dryRun: Boolean? = null`, kotlinx `encodeDefaults = false`), and every SDK generated from `artifact-keeper-api` — `openapi.yaml:25958-25963` types `dry_run` as optional with no `required:`. That spec is regenerated from the backend at release (`sync-openapi-spec.yml`), and the derive now marks the field required, so the spec self-corrects. `deny_unknown_fields` does **not** self-correct: utoipa's `ToSchema` does not emit `additionalProperties: false` for it, so the published schema will under-describe the new rejection. Noted in the CHANGELOG; worth a follow-up.

No caller anywhere in the org sends any field other than `dry_run`, so `deny_unknown_fields` breaks nothing on its own, and there is no documented-but-optional field for it to reject — the struct has exactly one field.

### "Rejects unknown fields" — where it actually lands

`deny_unknown_fields` is on `StorageGcRequest`, which both routes take, so both are covered. The refusal is axum's `JsonRejection::JsonDataError` → **422**, body `Failed to deserialize the JSON body into the target type: missing field \`dry_run\` at line 1 column 16` (captured from the route test). It is a client error, not a 500, and the text is serde's own — no server state in it.

It is **not** the `400 VALIDATION_ERROR` envelope from #1368, which `backend/src/api/extractors.rs` documents as "our public API contract". I checked the justification in the PR body rather than taking it: `storage_gc_service.rs` really does carry the pre-existing production clones that the duplication gate trips on (the 1.8.x CHANGELOG entry for the jscpd rework names that file by name as one of the two genuine offenders), and the envelope swap forces a touch to its test module. I also checked how settled the convention is: **3 of 44 handler modules** use `crate::api::extractors::Json`; 41 still use `axum::Json`. So this is the majority behaviour, not an exception carved out for this PR, and #1368 is an incremental migration. Leaving it is fine; the deferral is correctly written down.

One consequence worth stating: `Json` is the last extractor and `require_admin` runs inside the handler body, so a non-admin posting `{}` now sees 422 rather than 401. That ordering is pre-existing (a wrong-typed `dry_run` already did this), both routes are behind auth middleware, and the leak is "there is a field called `dry_run`". Not worth reordering.

### Defects found and fixed on the branch

1. **The CHANGELOG bullet landed inside the released `[1.8.2]` section.** The branch predates the 1.9.0 cut, so it was written under the `[Unreleased]` heading that main has since stamped `[1.8.2]`; `git merge origin/main` applied *cleanly* and left it there — no conflict to warn you. Same class of mistake as #3795. Moved to the current `[Unreleased]` / Changed (687e6715). Also corrected "a 400 breaks loudly" → 422 in the same prose, and named the CLI break in the upgrade note.

2. **The new route test was not in the `db-serial` nextest group.** `.config/nextest.toml` says in as many words: add any new test that takes `storage_gc_test_guard()` to the filter, because that guard is an in-process mutex and nextest runs a process per test. The new test takes it and was outside the group, so its positive control — a genuine cluster-wide **live** GC pass — ran concurrently with every other GC test. It cost me a real failure:

   ```
   FAIL api::handlers::oci_v2::...::completion_digest_mismatch_delete_failure_is_recoverable_by_storage_gc
   panicked at oci_v2.rs:23625: storage GC must delete leaked final PUT temp object
   ```

   Bisected: that filter set is 130/130 green without the new test (2 runs) and produced the failure above with it. `oci_v2`'s test also takes the guard and was also outside the group — a pre-existing gap that was harmless until a second out-of-group live pass existed — so both join it (9e4dc74b). Same filter is now 165/165 green, twice.

### Verification

Merged `origin/main` (a68e4d05) into the branch. Everything below on the merged tree, `SQLX_OFFLINE=true`, throwaway `postgres:18-alpine`, `AK_TESTS_REQUIRE_DB=1` (so a DB-backed test that cannot connect fails instead of skipping).

```
cargo fmt --check                                   clean
cargo clippy --lib --tests --all-features -D warnings   clean (4m05s)
cargo check --lib --tests                           clean (3m47s)
cargo nextest run --lib -E 'test(/storage_gc/) or test(test_openapi_spec_is_valid)' --test-threads 4
    Summary [26.514s] 165 tests run: 165 passed, 16086 skipped
    Summary [26.236s] 165 tests run: 165 passed, 16086 skipped   (re-run)
```

**Fail-without-fix, per production hunk, reverted separately:**

Restore `#[serde(default)]` on `dry_run` → 2 fail:
```
FAIL storage_gc_route_refuses_ambiguous_requests_without_deleting
  assertion `left == right` failed: `{}` must be refused, not treated as a live delete
    left: 200   right: 422
FAIL test_storage_gc_request_empty_body_rejected
```

Remove `#[serde(deny_unknown_fields)]` → 3 fail:
```
FAIL storage_gc_route_refuses_ambiguous_requests_without_deleting
  the refusal must name the unknown key, got: ... missing field `dry_run` at line 1 column 16
FAIL test_storage_gc_request_camel_case_key_rejected
  rejection must name the offending key, got: missing field `dry_run` at line 1 column 16
FAIL test_storage_gc_request_unknown_field_rejected
```

Note the second one is the interesting half: with `dry_run` required but unknown fields still allowed, `{"dryRun": true}` is *already* refused — but with "missing field `dry_run`", which never mentions the key the operator actually typed. `deny_unknown_fields` is what makes the refusal say `unknown field \`dryRun\`, expected \`dry_run\``, i.e. what makes it fixable at a glance. It earns its place.

`refresh_stats_after_live_gc` is behaviour-preserving: both call sites had byte-identical bodies guarded by `if !payload.dry_run`, the helper early-returns on `dry_run` and still swallows-and-logs the refresh error.

Out of draft. Not merging — the CLI companion fix should be open first so the two can be released together.

brandonrc:
Checks are clean, but holding the merge on purpose until the CLI companion lands.

`artifact-keeper-cli` builds the storage-GC body with `dry_run.then_some(true)`, so a **live** run (no `--dry-run`) sends `None`, the generated client omits the key, and the request posts `{}` — which this PR refuses with 422. `artifact-keeper-cli#176` fixes it to always send the value; it is currently red on `commands::client::tests::build_client_with_valid_token`, a failure that looks unrelated to the one-line change and is being diagnosed.

Nothing is broken in the meantime: main is unreleased, and no other caller I checked is affected (web, the native test scripts, artifact-keeper-test, docs — none send extra fields, and the repo-scoped route this PR also hardens is absent from the generated SDK entirely).

Two things to make sure of before 1.10.0 ships, both recorded here so they are not lost:

- the CLI fix must be in a **released** CLI, not just merged, or upgrading the backend breaks `ak admin storage-gc` for anyone on an older CLI;
- `artifact-keeper-api`'s `openapi.yaml:25958` still types `dry_run` as optional. That self-corrects at the release spec sync, but note that utoipa will not emit `additionalProperties: false` for `deny_unknown_fields`, so the spec will under-describe the unknown-field refusal until that is addressed separately.

Issue #3618, opened 2026-09-01
storage_stats_tests: OCI manifest bytes missing from storage stats aggregate (4096 counted, 4608 expected)


brandonrc:
Found by #3494 (running the integration targets CI had never executed).

`storage_stats_tests::oci_layer_bytes_are_counted` fails on current `main`:

```
thread 'oci_layer_bytes_are_counted' panicked at backend/tests/storage_stats_tests.rs:187:5:
assertion `left == right` failed: manifest + layer both counted
  left: 4096
 right: 4608
```

The fixture stores an OCI manifest (512 bytes) and a layer blob (4096 bytes) and expects the storage-stats aggregate to count both (4608). The aggregate reports 4096 — by the arithmetic, the 512-byte manifest is missing from the count.

This may be a real under-count in the stats aggregation (OCI manifests not attributed), or an intentional accounting change the never-run test predates. Needs a look at the storage-stats aggregation SQL before deciding test-fix vs code-fix. Then remove the `--skip oci_layer_bytes_are_counted` quarantine in `.github/workflows/ci.yml`.

The other 6 cases in the suite are green and now run in CI.

brandonrc:
Moved from 1.9.0 to 1.10.0 in the pre-release triage: 1.9.0 is being cut this week with the security and proxy-correctness fixes already on main, and this item does not change what ships in that release. It keeps its labels and priority.

brandonrc:
Confirmed real and now quarantined rather than silently unrun: #3620 wires `storage_stats_tests` into CI for the first time and `oci_layer_bytes_are_counted` fails there, counting 4096 of 4608 bytes for an OCI layer. Re-verified after #3578 (proxy-cached bytes in storage stats) landed, which does not change it — so this is the manifest bytes going missing from the aggregate, not the proxy-cache accounting. The test runs under `--skip` until this is fixed; removing that skip is the acceptance criterion.

Issue #3617, opened 2026-09-01
search_checksum_sha1_md5_tests: re-upload test predates version immutability and fails with an immutable-version Conflict


brandonrc:
Found by #3494 (running the integration targets CI had never executed).

`search_checksum_sha1_md5_tests::test_reupload_refreshes_all_three_checksums` fails on current `main`:

```
thread 'test_reupload_refreshes_all_three_checksums' panicked at backend/tests/search_checksum_sha1_md5_tests.rs:386:10:
re-upload should succeed via ON CONFLICT DO UPDATE: Conflict("Artifact version already exists and is immutable")
```

The test asserts that re-uploading the same version refreshes the sha256/sha1/md5 checksums via `ON CONFLICT DO UPDATE`. Version immutability now rejects the re-upload outright, which is almost certainly the intended product behavior, making the test's premise stale.

Suggested fix: rewrite the test to assert the current contract (immutable version -> Conflict, checksums unchanged), or route it through a repo configuration where re-upload is still supported, if any. Then remove the `--skip test_reupload_refreshes_all_three_checksums` quarantine in `.github/workflows/ci.yml`.

The other 4 cases in the suite are green and now run in CI.

brandonrc:
Moved from 1.9.0 to 1.10.0 in the pre-release triage: 1.9.0 is being cut this week with the security and proxy-correctness fixes already on main, and this item does not change what ships in that release. It keeps its labels and priority.

Issue #3616, opened 2026-09-01
conan_error_paths_tests: unauthenticated GET on an unknown Conan repo returns 401, test pins the #1046 contract of 404


brandonrc:
Found by #3494 (running the integration targets CI had never executed).

`conan_error_paths_tests::test_1046_unknown_repo_no_auth_returns_404` fails on current `main`:

```
thread 'test_1046_unknown_repo_no_auth_returns_404' panicked at backend/tests/conan_error_paths_tests.rs:382:5:
assertion `left == right` failed: unknown repo + no auth must return 404, got 401
  left: 401
 right: 404
```

The suite has never run in CI (it was outside the `--test` allowlist), so there is no CI history to bisect from; the behavior changed at some point after the test was written.

Decision needed: is 401-before-existence-check the intended contract now (uniform 401 for unauthenticated callers, which avoids a repo-existence oracle), or is this a regression of the #1046 contract? Either fix the handler or update the test to pin the current contract, then remove the `--skip test_1046_unknown_repo_no_auth_returns_404` quarantine in `.github/workflows/ci.yml`.

The other 7 cases in the suite are green and now run in CI.

brandonrc:
Moved from 1.9.0 to 1.10.0 in the pre-release triage: 1.9.0 is being cut this week with the security and proxy-correctness fixes already on main, and this item does not change what ships in that release. It keeps its labels and priority.

Issue #2924, opened 2026-07-26
DB-backed lib tests report as passing when the database is unreachable (try_pool collapses connect failure into None)


brandonrc:
## Problem

> **Line references and counts corrected.** The two Rust references and the call-site counts were read from
> a stale working tree (branch `feat/repo-export-import`), not from `main`. They are corrected below against
> `origin/main` at `ca87b456`. The `.github/workflows/ci.yml` references were all already correct. The
> finding itself reproduces on main unchanged. The `try_pool` snippet below is lightly abridged: main carries
> three comment lines between `.max_connections(3)` and `.acquire_timeout(...)`, which do not change the logic.

DB-backed lib tests report as **passing**, not skipped, when the database is unreachable. The test harness cannot tell "no database configured" apart from "database is broken", so a Postgres outage produces a green job instead of a red one.

## Evidence

`backend/src/api/handlers/test_db_helpers.rs:37-48`:

```rust
pub async fn try_pool() -> Option<PgPool> {
    let url = std::env::var("DATABASE_URL").ok()?;
    sqlx::postgres::PgPoolOptions::new()
        .max_connections(3)
        .acquire_timeout(std::time::Duration::from_secs(30))
        .connect(&url)
        .await
        .ok()
}
```

The trailing `.ok()` collapses a connect failure into the same `None` that an unset `DATABASE_URL` produces. Those are two very different conditions treated identically.

`Fixture::setup` propagates it, `backend/src/api/handlers/test_db_helpers.rs:1074-1075`:

```rust
pub async fn setup(repo_type: &str, format: &str) -> Option<Self> {
    let pool = try_pool().await?;
```

Callers use the let-else form:

```rust
let Some(fx) = tdh::Fixture::setup(...).await else { return; };
```

An early `return` from a `#[tokio::test]` body is a normally passing test. There is no `#[ignore]`, nothing printed to explain the skip, and nextest counts it in the `passed` column rather than `skipped`. The test looks identical in output to one that ran every assertion.

## Impact

This does not bite in CI today. `🧪 Backend Unit Tests` provides a real Postgres service container and points `DATABASE_URL` at it:

- service definition: `.github/workflows/ci.yml:150-167`
- readiness loop: `.github/workflows/ci.yml:202-217`
- migrations: `.github/workflows/ci.yml:219-222`
- nextest invocation with `DATABASE_URL` set: `.github/workflows/ci.yml:224-238`

[...]

PR #3620, opened 2026-09-01
ci(tests): wire every test target into CI or exempt it explicitly, with a self-checking surface gate

### LINKED ISSUE #3494: CI runs no bin-target tests and only 16 of 61 integration targets: 276 test functions never execute, so a green CI does not prove a new test ran

### LINKED ISSUE #3607: streaming_invariant.rs has never run in CI: the full-body-buffering ratchet is unfalsifiable
## Summary

`backend/tests/streaming_invariant.rs` — the ratchet that exists to stop new full-request-body-buffering handlers from being added — **has never run in CI**. It is referenced by no workflow and no script, it is not in the integration job's `--test` allowlist, and `--lib` excludes integration targets by construction.

```
$ grep -rn "streaming_invariant" .github/ scripts/
$        # no output
```

The integration job runs a hardcoded allowlist (`.github/workflows/ci.yml`, around lines 1655-1669):

```
--test lifecycle_policy_tests \
--test api_token_cache_flush_tests \
--test grpc_sbom_tests \
... (15 names) ...
--test ansible_galaxy_tests \
```

`streaming_invariant` is not among them. A file outside that list is not a failing test — it is a test that does not exist as far as CI is concerned.

## Why this matters more than a normal skipped test

This file is a **ratchet**, not a feature test. It carries `RAW_BODY_BLOB_ALLOWLIST`, the explicit inventory of handlers still permitted to take `body: Bytes`, and its purpose is to fail the build when a new one appears. It was written for the #1608 / PF-005 (#2517) streaming program.

Because it never ran, the allowlist has been decorative for its entire life. Concretely, it lists `("src/api/handlers/swift.rs", 1)` — and #3595 is exactly that site, storing an entire multipart envelope as if it were the package. The guard naming the problem and the bug were in the tree at the same time, and nothing connected them.

#3596 is a second instance of the same class on a different route.

## Suggested fix

Run it in the Tier 1 `--lib` job as its own invocation, the way `workflow_scan_gate_tests` is already handled at `ci.yml:592`:

```
cargo nextest run --workspace --test streaming_invariant --no-tests=fail
```

`--no-tests=fail` matters: without it, a rename or a compile-out silently returns green again, which is how the gate got into this state.

Unlike `workflow_scan_gate_tests`, this file's cases are not `#[ignore]`d, so no extra flags are needed.

## Wider question worth answering in this issue

This is the fourth distinct way a test in this repo can exist but never execute, alongside: bin-target tests excluded by `--workspace --lib`; integration targets outside the 15-name allowlist; and DB-backed tests that silently skip and report PASS without `DATABASE_URL`.

Two things worth doing beyond the one-line fix:

1. **Audit the rest of `backend/tests/`** against the allowlist and report which other files never run. `backend/tests/security_regression_tests.rs` is another that appears to be outside it — worth confirming whether that is intentional.
2. **Make the allowlist self-checking.** A test that enumerates `backend/tests/*.rs` and asserts every file is either in the CI allowlist or carries an explicit opt-out marker would convert this failure mode from invisible to loud. Otherwise the next file added to `backend/tests/` inherits the same fate, and nobody finds out until a bug it was written to prevent ships.

Found while investigating #3595 and #3596, both of which are instances of precisely what this ratchet was built to prevent.

brandonrc:
## Summary

Closes #3494. Closes #3607.

On current `main`, CI executed no bin-target tests and only 17 of 61 integration targets: **260 test functions (28 in `src/main.rs` + 232 across 44 `backend/tests/` targets) never ran in any job on any trigger**, including `security_tests`, `security_regression_tests`, `promotion_*` and the `streaming_invariant` ratchet. A green CI therefore did not prove a newly added test ever executed.

This PR makes the gap self-checking, turns on everything that passes, and quarantines the rest visibly:

**1. Self-checking gate — `backend/src/ci_test_surface.rs`.** A `--lib` test (so it runs in both the unit-test and coverage jobs and cannot itself fall out of the allowlist it polices) asserting that every `backend/tests/*.rs` target is either named in a `--test` invocation in a workflow — in the *right kind* of step for its `#[ignore]` profile, so a non-`#[ignore]`d test can't be "wired" into the `-- --ignored` job and pass vacuously — or carries a justified `EXEMPT` entry. Stale exemptions, dangling `--test` flags, dropped `--bins`, and missing `--no-tests=fail` on explicit-target nextest invocations all go red. Verified non-theatrical: adding an unlisted `backend/tests/zz_gate_probe_tests.rs` fails the gate with an actionable message naming the file.

**2. Bin-target tests on.** `--lib` → `--lib --bins` in the unit-test and coverage jobs: 28 tests in `backend/src/main.rs` (all pass, 0.02s of test runtime).

**3. Tier 1 pure integration targets.** The workflow-contract step widens from `workflow_scan_gate_tests` alone to all 9 targets with non-`#[ignore]`d tests: `security_tests` (13), `security_regression_tests` (8), `streaming_invariant` (2), the 4 PyPI conformance targets (24), `scan_dedup_short_circuit_tests` (1). All `--no-tests=fail`.

**4. Tier 2 allowlist +27 suites.** Every DB-only suite that passes against the job's Postgres (verified locally, migrations applied, `AK_TESTS_REQUIRE_DB=1`): 215 `#[ignore]`d tests total now run in the exact extended command, up from 133. `AK_TESTS_REQUIRE_DB=1` is now set on the integration steps too (#2924 was only half-applied).

**5. Quarantined cases, tracked not hidden.** 3 suites are green except one case each, run with an explicit `--skip` + tracking issue:
- `conan_error_paths_tests::test_1046_unknown_repo_no_auth_returns_404` — 401 vs the pinned 404 contract (#3616)
- `search_checksum_sha1_md5_tests::test_reupload_refreshes_all_three_checksums` — predates version immutability (#3617)
- `storage_stats_tests::oci_layer_bytes_are_counted` — 512-byte manifest missing from the stats aggregate (#3618)

**6. Fixes found by actually running the suites** (the point of the exercise):
- `manifest_blob_refs_tests` was born dead: its fixture inserts repository `format = 'oci'`, an enum value that has **never existed** (`docker` is the OCI value). The suite could not have passed on any commit. Fixed; all 3 tests pass.
- `streaming_invariant` (#3607) had drifted: #3392 removed one npm.rs buffer site and the never-running ratchet still counted 7. Reconciled (npm.rs 7 → 6, total 30 → 29) — the safe direction, fewer exemptions — and wired.

**Exempt with reasons** (in the gate, mirrored in the ci.yml comment): `azure_rbac_live_test`, `azure_shared_key_live_test`, `s3_integration` (live cloud creds); `download_ticket_tests`, `integration_tests`, `replication_integration`, `storage_backend_tests`, `tag_filtered_replication_tests` (need a running HTTP backend via `TEST_BASE_URL`; 68 tests — closing that class needs a compose-style job and is out of scope here). Known partial: `incus_upload_tests` still runs 1 of 10 cases via the named filter (pre-existing, documented in ci.yml).

### Executed-test count, before → after
| where | before | after |
|---|---|---|
| unit jobs (`--lib` → `--lib --bins`) | 15,962 | 15,997 (+28 bin, +7 gate) |
| Tier 1 integration step | 5 | 53 |
| Tier 2 integration job | 134 | 233 (215 + 17 quarantine-step + 1 incus) |
| **never-executing test functions** | **260** | **71 (68 infra-exempt + 3 quarantined), all explicit and gate-tracked** |

### Runtime cost
- Unit/coverage jobs: +28 tests at ~0.02s; one extra test-profile compile of the bin target (already compiled by clippy `--all-targets`, so sccache-warm).
- Tier 2: the full extended command (42 targets, `--test-threads=1`) ran in 5m32s locally *including* link time for the 27 new target binaries; incremental test runtime is ~60s. The job already runs only on main/release pushes and backend PRs.

- [x] This PR is a `fix/*` AND adds/updates a test that would have caught the bug — the `ci_test_surface` gate fails on `main`'s workflow wiring (44 unwired targets) and passes here; `manifest_blob_refs_tests` and `streaming_invariant` fail on `main` and pass here.

## Test Checklist
- [x] Unit tests added/updated (`ci_test_surface` gate + helper tests)
- [x] Integration tests added/updated (27 suites wired, 2 fixed)
- [x] Manually tested locally (full `--lib --bins` under `AK_TESTS_REQUIRE_DB=1`: 16,019 passed; exact Tier 2 command: 215 passed; quarantine step: 17 passed / 3 skipped)

## API Changes
- [x] N/A - no API changes


brandonrc:
## Audit round 1 — CI test-surface gate (2 reviewers: workflow correctness + adversarial gate audit)

The second reviewer attacked the gate empirically in an isolated worktree at head `14aff231`, running `SQLX_OFFLINE=true cargo test --lib ci_test_surface` after each mutation. That table is the core of this review.

**The diagnosis is right and the wiring is real value.** I recomputed the central claim independently: **17 of 61** targets wired on `origin/main` is exactly correct, going to 53 at head; executed test functions go **148 → 298**. The `manifest_blob_refs_tests` fix is hard evidence the suite never ran — `repositories.format` is a PG enum and no migration ever adds `'oci'` (I grepped every `ALTER TYPE repository_format`), so `create_test_repo` raised `invalid input value for enum repository_format: "oci"` and every test in that file panicked at the fixture.

**And the `streaming_invariant` ratchet was tightened, not loosened** — ALLOWLIST npm.rs 7→6, `assert_eq!(total, …)` 30→29, exact equality so it can't be silently relaxed. Better: `npm.rs` has exactly **6** `STREAMING-EXEMPT` markers on `origin/main` too, so the old `7` was already failing on main — which independently proves this ratchet never executed. Head's allowlist sums to 29 and matches a full census (the two unlisted markers are inside `#[cfg(test)]` and correctly excluded).

**But the gate is not self-checking.** It was defeated seven ways, each verified green.

---

### The empirical table

| # | Mutation | Result |
|---|---|---|
| E0 | none (PR head) | `EXIT=0` — 7 passed ✅ passes as claimed |
| E1 | add unwired `backend/tests/zz_gate_probe_tests.rs` | `EXIT=101` ✅ **caught**, names the file |
| E3 | move `--test streaming_invariant` into the `-- --ignored` step | `EXIT=101` ✅ **caught**, good message |
| E2b | delete the real `--test security_tests`; add a YAML **comment** mentioning it | `EXIT=0` ❌ **defeated** |
| E5 | "wire" it from a `workflow_dispatch`, `if: false`, **echoed** command | `EXIT=0` ❌ **defeated** |
| E6 | `#[ignore]` the gate test itself + break wiring | `EXIT=0` — 6 passed, 1 ignored ❌ **defeated** |
| E7 | delete `mod ci_test_surface;` from `lib.rs` + break wiring | `EXIT=0` — no warning, no error ❌ **defeated** |
| E8 | append `--skip test_` to the quarantine step | `EXIT=0` ❌ **defeated** |
| E9 | drop `--lib` from the **blocking** unit job only (`ci.yml:571`) | `EXIT=0` ❌ **defeated** |
| E9b | drop `--lib` from *both* real run lines | `EXIT=101` ✅ caught |
| E10 | append `("security_tests", "")` to `EXEMPT` + delete the flag | `EXIT=0` ❌ **defeated** |

The property actually enforced is: *every `backend/tests/*.rs` stem appears as the token after the literal `--test ` somewhere in the raw text of some file in `.github/workflows/`, on a line that does or doesn't also contain `-- --ignored`.* No YAML parsing, no notion of a job or step, no trigger awareness, no knowledge that `--skip` or positional filters exist.

### 1. 🔴 A YAML comment counts as wiring

`test_target_names` (`ci_test_surface.rs:100-114`) does `rest.find("--test ")` on raw text; `workflow_sources()` slurps whole files. Nothing distinguishes a comment, a `name:`, an `env:` value, or an echoed string from an executed command.

**The 2-line diff that ships green:**
```diff
-            --test security_tests \
+      # Historical note: this step used to pass --test security_tests here.
```
`security_tests` — 13 request-smuggling / path-traversal / injection regression tests — then runs in no job, and the gate reports 7 passed.

(Honest note: the *first* comment-evasion attempt went red — but on `nextest_explicit_target_invocations_fail_on_zero_tests`, not the surface gate — because the comment happened to contain the literal `nextest run`. Rewording to avoid that token made all 7 green. The hole is real; a keyword collision masked it once.)

This isn't hypothetical fragility: `ci.yml` already contains prose like "the Tier 2 job passes an explicit `--test` allowlist", which escapes only because the backtick after `--test` isn't a space.

### 2. 🔴 Any workflow file counts, including ones that never run

`workflow_sources()` reads every `*.yml` under `.github/workflows/` with no inspection of `on:`, job `if:`, or step `if:`. Verified green: delete the real flag, add a `workflow_dispatch`-only `zz-archive.yml` with `if: false` and an `echo`. By the same token `scheduled-tests.yml` (cron+dispatch only) and `e2e.yml` are equally valid "wiring."

Related: the `changes` job's `backend` filter matches only `ci.yml`, not sibling workflow files — so a target "wired" in `scheduled-tests.yml` doesn't even influence whether the integration job runs.

### 3. 🔴 "Wired" ≠ "executed" — `--skip` and positional filters are invisible

The word `skip` doesn't appear in `ci_test_surface.rs`, nor does any handling of a positional filter. **This is already live in the PR:** `incus_upload_tests` is wired as `--test incus_upload_tests test_chunked_upload_cross_repo_session_rejected` — the file has **10** test fns, **1** runs, and the gate counts the target satisfied.

**Verified green:** append `--skip test_` to the quarantine step. libtest's `--skip` is substring-matched, so all three "quarantined" suites execute nothing and exit 0. The same one-liner on the main Tier 2 step silences all 46 DB suites.

### 4. 🔴 The gate's own kill switch is one line, and nothing watches it

Nothing in the repo references `ci_test_surface` except `mod ci_test_surface;` in `lib.rs` and prose. Three verified one-line kills: `#[ignore]` the gate test (E6); delete the `mod` declaration — the file becomes an orphan producing **no warning and no error** (E7); or drop `--lib` from `ci.yml:571`.

E9 deserves attention. The `with_bins >= 2` pin counts **three** matching lines, one of which is `ci.yml:1024` — `cargo nextest run --workspace --lib --bins --no-run`, a compile check running zero tests. So the pin tolerates exactly one real test-running invocation being deleted, and the one you'd delete is line 571 in `test-backend-unit` — the only job where the gate is **blocking** (`require_success`). The survivor is in `coverage`, which `ci-complete` treats as non-blocking. Net: gate still runs, can still go red, PR still green.

**Fix:** require ≥2 invocations *not* containing `--no-run`, and add an independent check (shell test, or a `workflow_scan_gate_tests` assertion) that `lib.rs` declares `mod ci_test_surface;` and the gate test carries no `#[ignore]`.

### 5. 🔴 The `s3_integration` exemption's justification is factually false

```rust
("s3_integration", "requires live S3 credentials; the release-gate S3 compose job covers this path"),
```
**There is no such job.** Repo-wide grep at head: no workflow references S3, MinIO, or `s3_integration` (only hits are a comment at `ci.yml:1696` and the EXEMPT entry itself). `docker-compose.s3-maven-test.yml` exists, but its own header documents it as a manual local repro harness for #361, and no workflow references it.

So 12 S3 storage tests — including `test_presigned_url_generation_and_download` and `test_presigned_url_not_returned_when_redirect_disabled`, a redirect/URL-exposure control — run nowhere, behind a justification that reads as if they're covered. That's worse than an honest "not covered."

### 6. 🟠 The exemption hatch has ~zero friction, and hides six auth/authz suites

The justification string is **never read** — `EXEMPT.iter().map(|(n, _)| n.to_string())` — and `#![allow(dead_code)]` means an empty string doesn't warn. No issue link, expiry, owner, cap, or format check; `CODEOWNERS` has only the blanket `* @artifact-keeper/backend`. E10 confirmed: one appended line with an empty reason silences a suite.

Judging the 8 current entries: the two Azure ones are **legitimate** (`env::var(...).expect(...)` on cloud creds). `s3_integration` is not, per finding 5. The five `TEST_BASE_URL` suites (`download_ticket_tests`, `integration_tests`, `replication_integration`, `storage_backend_tests`, `tag_filtered_replication_tests`) are **accurate but load-bearing** — between them they are the *only* coverage for single-use download-ticket enforcement, ticket path-binding, ticket write-method rejection, 401 on missing/invalid ticket, private-repo anonymous-download blocking, OCI catalog visibility for non-admins, and `test_admin_storage_backends_requires_auth`. The stated reason ("the CI integration job provisions Postgres only") is true of `ci.yml` today, but `e2e.yml` already does `docker compose up -d postgres backend` — so it describes current wiring, not an inherent impossibility, and freezes it with no expiry.

### 7. 🟠 Tier 2 triples its target count inside an unchanged 40-minute cap on a job with no build cache

`test-backend-integration` is `timeout-minutes: 40` (`ci.yml:1520`) and explicitly carries **no `rust-cache`** (`:1567-1572`: "This job carries no rust-cache"). It goes 16 → 46 integration targets. Each is a separate binary linking the whole backend rlib, and sccache caches compilation but never linking. Execution also grows 143 → 245 cases, all serialized (`--test-threads=1`, and cargo runs test binaries one at a time).

**Failure:** the first backend PR after merge hits the 40-minute wall during link/run, and every backend PR is blocked by what reads as an infra failure. Nothing in the PR raises the cap or splits the job. Secondary: `test-backend-unit` (also 40 min, and a **required** check on every code PR) now compiles 9 integration binaries instead of 1, plus `--bins`.

I'd resolve this before merge — it's the finding that turns CI red on day one.

### 8. 🟠 `--no-tests=fail` protects 8 of 53 targets, and even there it's per-invocation

`nextest_explicit_target_invocations_fail_on_zero_tests` filters on `l.contains("nextest run")`. The three Tier 2 invocations use `cargo test`, not nextest (`ci.yml:1693`, `:1801`, `:1815`), so **45 of 53 wired targets carry no zero-test guard at all**. And nextest's `--no-tests` governs the whole run: the single Tier 1 step batches 9 targets under one `--no-tests=fail`, so moving `security_tests`' 13 tests behind a feature flag leaves 8 siblings keeping the run non-empty and the step green.

On `main` this flag was genuinely load-bearing because it guarded a single-target invocation. Batching removed that property while the comment still claims it.

### 9. 🟡 The quarantine rots silently, three ways

(a) A `--skip` added for a *passing* test is invisible — permanent, undetectable hiding. (b) A `--skip` naming a renamed/deleted test is a silent no-op; libtest doesn't error on an unmatched skip. (c) Tracking issues #3616/#3617/#3618 are never checked for closure, so the skip outlives the fix.

The patterns themselves are **correct today** — I checked each against all 61 targets: `test_1046_unknown_repo_no_auth_returns_404` → exactly `conan_error_paths_tests.rs:365`; `test_reupload_refreshes_all_three_checksums` → exactly `search_checksum_sha1_md5_tests.rs:357`; `oci_layer_bytes_are_counted` → exactly `storage_stats_tests.rs:164`. No over-match, no cross-binary collision, and the per-suite arithmetic in the comment (7/8, 4/5, 6/7) matches the real `#[ignore]` counts. These are `cargo test`/libtest invocations, so `--skip` is substring-filtered as intended — the nextest-vs-libtest semantic difference is correctly avoided.

### 10. 🟡 `AK_TESTS_REQUIRE_DB=1` is inert for 28 of the 30 newly-wired suites

`ci.yml:1712-1714` justifies it as making "a connect failure loud instead of a silent skip," and CLAUDE.md tells contributors it "turns a missing/unreachable database into a hard failure." But `tests_require_db()` (`testing.rs:34-46`) is consulted only by `try_pool`, and only `gitlfs_locks_tests` and `oci_upload_range_test` call it. The other 28 use `common::require_db_pool()` (which panics anyway, for unrelated reasons) or a bare `PgPool::connect(&env::var(..).expect(..))`.

The one genuine silent-skipper is the one the Tier 1 comment specifically cites: `scan_dedup_short_circuit_tests.rs:467-471` does `let Ok(database_url) = env::var("DATABASE_URL") else { eprintln!("...skipping..."); return; }` — ignoring `AK_TESTS_REQUIRE_DB` entirely. It's also the only DB-writing test in the parallel (`--test-threads 8`) Tier 1 step, spawns 10 concurrent tasks, and has no cleanup — so it leaks one `repositories` row per CI run into the unit job's Postgres.

### 11. 💭 Smaller items

- **The "260 test functions" figure is wrong, and the code disagrees with the body.** Real count in the 44 targets unwired on main: **218** (48 non-ignored + 170 ignored); +28 `main.rs` bin tests = **246**. The body says 260, `ci_test_surface.rs:6` says 232. The "44 of 61" and "17 of 61" parts are correct.
- **The `#[ignore]` classifier is correct today by luck of a uniform corpus.** Verified: zero `cfg_attr`, `rstest`, `test_case`, `macro_rules!` or `#[cfg(feature` anywhere in `backend/tests/`, and every `#[ignore]` is single-line. Credit — the mixed-profile logic is real and correct for both mixed targets (`security_regression_tests` 8+4, `scan_dedup_short_circuit_tests` 1+7), and E3 proves it enforces this. Latent: `#[cfg_attr(feature="db", ignore)]` parses as attribute name `cfg_attr` → silently non-ignored; macro-generated tests yield `(0,0)` and a zero-test target imposes **no profile requirement at all**, so it can be wired into either step kind freely; `run: >` folded scalars would split a command across logical lines.
- **`ci.yml:1693`'s comment overstates the guarantee and has already drifted** — it lists `incus_upload_tests` under "Deliberately excluded," but that target isn't in `EXEMPT` and *is* wired. Also `.github/actions/` composite actions and `uses: ./.github/workflows/*` reusable workflows aren't scanned (neither carries a `--test` today — credit), and `--test=name` is undetected, though that direction fails safe.
- **CLAUDE.md's new paragraph** should match what the gate actually enforces; per finding 10 the `AK_TESTS_REQUIRE_DB` instruction is stronger than reality.

---

### What genuinely works — credit where due

E1 and E3 both go red with precise, actionable messages naming the file and the reason. The wrong-profile check is a real idea most such gates skip, and it's correctly implemented for the mixed targets. `logical_lines` correctly joins backslash continuations; `--test-threads` is correctly not matched (the trailing space does it); `--test  x` and `--test=x` both fail safe. Dangling-`--test` and stale-`EXEMPT` checks work as designed. `ci.yml` is explicitly classified `code=true, backend=true` in the Docs-Only Gate, so a workflow-only PR *does* run the gate — exactly the trap a gate like this usually falls into, avoided deliberately. And `test-backend-unit` is `require_success` in `ci-complete`, so it genuinely blocks — modulo finding 4.

The workflow itself is correct: valid YAML, right job graph and runners, Postgres provisioned and migrated in both jobs that now need it, `JWT_SECRET` in scope on every new step, and the new work runs on the triggers claimed — Tier 1 in `test-backend-unit` (every code PR and push), Tier 2 in `test-backend-integration` (backend PRs plus main/release pushes). Nothing was hidden in a schedule-only job. (`ci.yml` has no `schedule` trigger at all, so the body's "PR, push, schedule" framing is loose but harmless.)

### Minimum to make the claim true

1. Strip YAML comments before scanning, and restrict to `run:` blocks of steps without `if: false`, in jobs reachable from a `pull_request` trigger. (Or just parse the YAML.) — fixes 1 and 2.
2. Reject a target as "wired" if its invocation carries a positional filter or `--skip`, or make those require their own justified allowlist with a tracking issue. — fixes 3 and 9.
3. Require `EXEMPT` entries to carry an issue reference; assert the justification is non-empty and well-shaped; put `ci_test_surface.rs` under a narrower `CODEOWNERS` entry. — fixes 6.
4. Fix `with_bins >= 2` to exclude `--no-run` lines, and add an out-of-band check that `lib.rs` still declares the module and the gate test isn't `#[ignore]`d. — fixes 4.
5. Replace the `s3_integration` justification with the truth, or wire it to `docker-compose.s3-maven-test.yml` in a real job. — fixes 5.
6. Resolve the 40-minute timeout question before merge. — finding 7.
7. Either extend `--no-tests=fail`-equivalent coverage to the `cargo test` steps, or split Tier 1's batch into per-target invocations. — finding 8.

I'd land the wiring — 218 newly-executing test functions is real, and the underlying diagnosis is sound. I wouldn't yet describe the gate as self-checking in the PR body or CLAUDE.md, and finding 7 needs an answer first.

<sub>Round 1 of an automated multi-reviewer audit. The gate audit was empirical: each mutation applied in an isolated worktree at `14aff231` and the gate re-run; counts recomputed independently rather than taken from the PR body.</sub>

brandonrc:
Merged `origin/main` (post-1.9.0) and made the branch green. No CHANGELOG conflict came up — this branch never touched `CHANGELOG.md`, and CI-only changes here generally don't carry an entry, so nothing was added under `## [Unreleased]`.

### What was red, and why

`🧪 Backend Unit Tests` and `📊 Code Coverage` both failed **before running a single test**, in the same place:

```
error: could not compile `artifact-keeper-backend` (lib test)
Caused by: ... rustc ... backend/src/lib.rs --test ... (signal: 9, SIGKILL)
```

SIGKILL on rustc is the OOM killer, not a compile error — and the annotation the coverage job prints for exactly this ambiguity was right to say so. Adding `--bins` to the `--lib` invocation makes cargo schedule the lib-test and bin-test rustc processes concurrently, and this job's own env comment records that the lib-test compile *alone* peaks at 7.9–11.0 GiB against the runner's 16Gi limit (#1515, #2062).

Measured locally with the jobs' settings (`CARGO_PROFILE_*_DEBUG=0`, `CARGO_INCREMENTAL=0`, `CARGO_BUILD_JOBS=8`), peak RSS across the build's processes:

| build | peak |
|---|---|
| `--lib --bins --no-run` (one invocation) | **19.7 GiB** |
| `--lib --no-run` then `--bins --no-run` | **14.7 GiB** |

So both jobs now compile the two targets in separate `--no-run` passes and the run that follows compiles nothing. This costs no extra compile work — the rlib, the lib test and the bin test are each built exactly once either way; only their overlap changes. `nextest_lib_invocations_include_bins` now skips `--no-run` lines: they execute no tests, so they cannot drop one, and deleting a `--no-run` pass only moves that target's compile back into the run.

### Newly-surfaced failures and dispositions

Everything below only ever surfaced because this PR runs the target for the first time.

| # | failure | verdict | disposition |
|---|---|---|---|
| 1 | `manifest_blob_refs_tests` — fixture inserts `format = 'oci'`, a `repository_format` value that has never existed | **test wrong** | fixed (`'oci'` → `'docker'`); 3/3 pass |
| 2 | `streaming_invariant` — ratchet still counted 7 npm.rs buffer sites after #3392 removed one | **test wrong** (stale, in the safe direction) | reconciled to npm.rs 6 / total 29 and wired with `--no-tests=fail`; still matches the post-merge source |
| 3 | `conan_error_paths_tests::test_1046_unknown_repo_no_auth_returns_404` — unauthenticated GET on an unknown repo returns 401, test pins 404 | **product-contract question** — 401-before-404 is a legitimate enumeration defence, but #1046 pinned 404 deliberately | quarantined with `--skip`, tracked in #3616 |
| 4 | `search_checksum_sha1_md5_tests::test_reupload_refreshes_all_three_checksums` — re-upload now rejected as an immutable-version `Conflict` | **test predates the feature** | quarantined with `--skip`, tracked in #3617 |
| 5 | `storage_stats_tests::oci_layer_bytes_are_counted` — 4096 of 4608 bytes counted | **code wrong** — a real accounting bug, not a stale test; re-verified after merging #3578, which did not change it | quarantined with `--skip`, tracked in #3618 |
| 6 | **new:** `provision_admin_user_does_not_arm_gate_for_inactive_admin_3723` (bin target) | **test wrong** — asserts *zero local admin rows cluster-wide*, which the shared unit-test Postgres never has (~16k tests at 8 threads create local admins; the bin binary schedules last — 5 rows present when it ran). Passes alone against a clean database. The `db-serial` group cannot fix it: it serializes its members, not the hundreds of admin-creating tests outside it. | `#[ignore]`d with the reason, tracked in **#3796** |
| 7 | 8 targets needing live cloud credentials or a running HTTP backend (`TEST_BASE_URL`) | **cannot run in this job** | `EXEMPT` with a per-target justification the gate re-validates |

#6 arrived on `main` in #3740 *after* this branch — the exact class of defect #3494 exists to catch: a test written, merged green, and never executed anywhere. Nothing is `--skip`ped or `#[ignore]`d without an open issue.

### Gate is not theatre

With a throwaway `backend/tests/zz_gate_probe_tests.rs` added (one plain and one `#[ignore]`d test, wired nowhere):

```
thread '...every_integration_target_is_wired_into_ci_or_exempt' panicked at backend/src/ci_test_surface.rs:274:9:
integration targets that run in NO CI workflow: ["zz_gate_probe_tests"]
Every backend/tests/*.rs file must be named in a `--test` invocation in .github/workflows/ ...
```

Removed again; the gate is green on the real tree.

### Verification on the merged tree

All against PostgreSQL 18 with `AK_TESTS_REQUIRE_DB=1`, via `cargo nextest`:

- `cargo fmt --all -- --check`, `cargo clippy --workspace --all-targets -- -D warnings`, `cargo check --workspace --lib --tests` — clean
- `--lib --bins --test-threads 8` → **16272 passed, 9 skipped** (0 failed)
- Tier 1 step, all 9 targets, `--no-tests=fail` → **53 passed**
- Tier 2 allowlist, all 42 targets, ignored-only, single-threaded → **215 passed**
- the 3 quarantine suites run *without* their skips → 17 passed / 3 failed, i.e. exactly the three cases above and nothing else
- the named `incus_upload_tests` case → 1 passed
- `ci_test_surface` itself → 7 passed

PR #3621, opened 2026-09-01
fix(api): reject NUL byte in artifact path at the boundary (#3545)

### LINKED ISSUE #3545: Unauthenticated 500 DATABASE_ERROR: NUL byte in artifact path reaches Postgres instead of being rejected at the boundary
## Summary

A percent-encoded NUL byte anywhere in the artifact path of a repository-scoped route is passed straight through to Postgres. The driver rejects the byte at the protocol level and the handler surfaces it as `500 DATABASE_ERROR`, instead of the request being rejected at the path boundary with a 400.

It is reachable **without credentials** on a public repository.

## Reproduction

```
$ curl -s --path-as-is -w '\n[%{http_code}]\n' \
    "http://<backend>/api/v1/repositories/<public-repo>/artifacts/a%00b"
{"code":"DATABASE_ERROR","message":"Database operation failed"}
[500]

$ curl -s --path-as-is -w '\n[%{http_code}]\n' \
    "http://<backend>/api/v1/repositories/<public-repo>/download/a%00b"
{"code":"DATABASE_ERROR","message":"Database operation failed"}
[500]
```

No `Authorization` header on either request. The repository was created with `is_public: true`; both routes serve anonymous reads for it and return a normal `404 NOT_FOUND` for any other non-existent path.

Backend log:

```
ERROR http_request{method=GET uri=/api/v1/repositories/rtpt/artifacts/a%00b ...}:
  artifact_keeper_backend::error: Request error
  error=Database error: error returned from database: invalid byte sequence for encoding "UTF8": 0x00
  code="DATABASE_ERROR"
```

## Affected versions

Reproduced on three separate deployments from published images:

| image | reported version | result |
|---|---|---|
| `ghcr.io/.../artifact-keeper-backend:1.6.2-release` | 1.6.2 | 500 |
| `ghcr.io/.../artifact-keeper-backend:1.8.1` | 1.8.1 | 500 |
| `ghcr.io/.../artifact-keeper-backend:dev` | 1.8.1 | 500 |

Long-standing, not a recent regression.

## Impact

Low to moderate, and deliberately not being characterised as more than that:

- Not SQL injection. The query is parameterised; Postgres rejects the byte in the wire protocol, not in a parsed statement.
- Not a traversal. Nothing outside the repository is returned; the request errors before any storage read.

What it does mean:

- An unauthenticated caller drives a connection-pool checkout and a failing query per request with a trivially craftable URL, on any public repository. That is a cheap amplification primitive against the pool and a log-noise generator.
- A whole class of input reaches the persistence layer without validation, which is the wrong place for it to be caught. The correct answer is a 400 at path extraction, alongside whatever already rejects `%c0%af` (that vector correctly returns `400 Invalid URL: Invalid UTF-8 in path` on the same route, so the boundary check exists — it just does not cover `0x00`).
- The 500 leaks a coarse implementation detail (`DATABASE_ERROR`) for what is a client input error.

## Suggested fix

Reject `\0` in the extracted artifact path at the same boundary that already rejects invalid UTF-8, returning 400. Both `/api/v1/repositories/{key}/artifacts/{path}` and `/api/v1/repositories/{key}/download/{path}` need it; the format-native `/generic/{key}/{path}` route already returns 404 for the same input.

## Test-side note

`artifact-keeper-test`'s `tests/security/test-path-traversal.sh:130` ("Reject null byte injection in path") already sends `legit.txt%00../../etc/passwd` at this route, but its final branch is `else pass`, so the 500 is recorded as a pass. A regression assertion should land with the fix rather than before it. Tracked test-side in artifact-keeper-test#388.

brandonrc:
## Summary

A percent-encoded NUL byte (`%00`) anywhere in the wildcard artifact-path
segment of a repository-scoped route was decoded by axum's `Path` extractor
into a literal `\0` and passed straight to Postgres. The driver rejects
`0x00` at the wire protocol (`invalid byte sequence for encoding "UTF8":
0x00`), which the handler surfaced as an unauthenticated **`500
DATABASE_ERROR`** on any public repository — instead of the request being
rejected at the path boundary with a `400`.

Two defects in one:

- The input was not rejected up front. An unauthenticated caller drives a
  connection-pool checkout plus a failing query per trivially craftable URL
  on any public repository — a cheap pool-amplification and log-noise
  primitive.
- The `500` leaked a coarse implementation detail (`DATABASE_ERROR`) for what
  is a client input error.

The sibling invalid-UTF-8 vector (`%c0%af`) already returned `400 Invalid
URL: Invalid UTF-8 in path` on the same route, so the boundary check existed —
it just did not cover `0x00`.

This adds a NUL-byte check at the same boundary, returning `400
VALIDATION_ERROR`, on the four repository-scoped routes that take a wildcard
path:

- `GET    /api/v1/repositories/{key}/artifacts/{path}` (metadata)
- `GET    /api/v1/repositories/{key}/download/{path}`
- `GET    /api/v1/repositories/{key}/versions/{path}`
- `DELETE /api/v1/repositories/{key}/artifacts/{path}`

The Generic format route (`/general/{key}/{path}`) re-exports
`download_artifact`, so it is covered by the same check. The check is
deliberately **NUL-only** — read routes must keep serving every path an
upload could ever have stored, so no other character class is rejected. The
existing upload validator (`upload_service::validate_artifact_path`) already
rejects NUL on the write side; this is the read/delete-side counterpart.

Before / after (anonymous, public repo):

```
# before
GET .../artifacts/a%00b   -> 500 {"code":"DATABASE_ERROR",...}
# after
GET .../artifacts/a%00b   -> 400 {"code":"VALIDATION_ERROR","message":"artifact path must not contain a NUL byte"}
# unchanged
GET .../download/dir/hello.txt -> 200 (bytes)
GET .../download/dir/missing   -> 404
GET .../artifacts/a%c0%afb     -> 400 (pre-existing invalid-UTF-8 boundary)
```

Added:

- Unit tests for the new `reject_nul_in_path` helper (400 mapping, NUL at every
  position, and negative controls proving the check is NUL-only — percent
  escapes, non-ASCII, and other control characters still pass).
- DB-backed router-level tests driving each of the four routes with `a%00b`
  and asserting `400` (not the driver `500`), each paired with a non-NUL
  negative control that flows past the boundary (`404` / non-`400`).

An `artifact-keeper-test` DTF tier (`nul-byte-path`) demonstrates the anonymous
`500 -> 400` transition end-to-end and is validated RED (baseline: 5/8) →
GREEN (fix: 8/8) locally; it is being batched with the other 1.9.0 tiers
(tracked test-side in artifact-keeper-test#388).

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

Closes #3545

### DIFF
--- a/backend/src/api/handlers/repositories.rs
+++ b/backend/src/api/handlers/repositories.rs
@@ -7362,6 +7362,11 @@ pub async fn get_artifact_metadata(
     Query(version_query): Query<ArtifactVersionQuery>,
     dl_ctx: crate::api::middleware::download_telemetry::DownloadContext,
 ) -> Result<Response> {
+    // #3545: a `%00` in the wildcard segment decodes to a literal NUL that
+    // Postgres rejects at the wire protocol, turning an unauthenticated
+    // request into a 500. Reject at the boundary, before any query runs.
+    crate::api::validation::reject_nul_in_path(&path)?;
+
     let repo_service = RepositoryService::new(state.db.clone());
     let repo = repo_service.get_by_key(&key).await?;
     require_visible(&repo, &auth, &repo_service).await?;
@@ -7652,6 +7657,10 @@ pub async fn list_artifact_versions(
     Extension(auth): Extension<Option<AuthExtension>>,
     Path((key, path)): Path<(String, String)>,
 ) -> Result<Json<ArtifactVersionListResponse>> {
+    // #3545: same boundary check as the artifact metadata/download routes —
+    // a NUL in the path must be a 400 here, not a driver-level 500.
+    crate::api::validation::reject_nul_in_path(&path)?;
+
     let repo_service = RepositoryService::new(state.db.clone());
     let repo = repo_service.get_by_key(&key).await?;
     require_visible(&repo, &auth, &repo_service).await?;
@@ -8575,6 +8584,10 @@ pub async fn download_artifact(
     // for a body that never arrives (the connection hangs).
     let is_head = request.method() == axum::http::Method::HEAD;
 
+    // #3545: reject a NUL byte in the artifact path at the boundary (400)
+    // instead of letting it reach Postgres as an unauthenticated 500.
+    crate::api::validation::reject_nul_in_path(&path)?;
+
     let repo_service = RepositoryService::new(state.db.clone());
     let repo = repo_service.get_by_key(&key).await?;
     require_visible(&repo, &auth, &repo_service).await?;
@@ -9004,6 +9017,11 @@ pub async fn delete_artifact(
 ) -> Result<()> {
     let auth = require_auth(auth)?;
     auth.require_scope("delete")?;
+
+    // #3545: same boundary check as the artifact read routes — a NUL in the
+    // path must be a 400, not a driver-level 500 from the lookup queries.
+    crate::api::validation::reject_nul_in_path(&path)?;
+
     let repo_service = RepositoryService::new(state.db.clone());
     let repo = repo_service.get_by_key(&key).await?;
     require_repo_write_access(&auth, &repo, &repo_service).await?;
@@ -17913,7 +17931,7 @@ mod tests {
         let auth = tdh::make_auth(fx.user_id, &fx.username);
         let router = tdh::router_with_auth(super::router(), state, auth);
 
-        let req = Request::builder()
+        let req = axum::http::Request::builder()
             .method("POST")
             .uri(format!(
                 "/{}/cache/invalidate?path=foo%2Fbar-1.2.3.tgz",
@@ -17999,7 +18017,7 @@ mod tests {
 
         // Invalidate the HTML representation only.
         let router = tdh::router_with_auth(super::router(), state, auth);
-        let req = Request::builder()
+        let req = axum::http::Request::builder()
             .method("POST")
             .uri(format!(
                 "/{}/cache/invalidate?path=simple%2Fproj3290%2F",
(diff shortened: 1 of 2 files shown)


brandonrc:
Closing as superseded by #3665, which I merged after opening this.

`backend/src/api/middleware/nul_path.rs` decodes the request path and query and refuses a NUL with a 400 `VALIDATION_ERROR` before routing, and it is registered globally in `create_router` (`backend/src/api/routes.rs:216`). All four handlers this PR patches sit behind that layer, so the `reject_nul_in_path` calls it adds are unreachable in production. I closed #3545 on 7 September saying exactly that and then left this branch open anyway.

Two things worth recording from the verification rather than losing them:

The red `🦀 Check Rust` was self-inflicted and unrelated to the fix — the branch rewrote seven pre-existing tests from `Request::builder()` to `axum::http::Request::builder()` without removing the now-unused function-local `use`, so clippy's `-D warnings` turned seven unused-import errors into a failed lib-test compile.

More interesting: this PR's four DB tests genuinely fail without its hunk, but only because they mount `super::router()` / `download_router()` directly and therefore bypass the global layer. Production never builds those routers in isolation. Main's own tests — `nul_refusal_carries_correlation_id_and_is_byte_identical_db`, `nul_in_the_repository_key_is_400_not_db_500_db`, `format_native_nul_paths_are_400_not_db_500_db` — drive the literal reproduction through `create_router` and pass with this PR's production changes reverted. A test that mounts a sub-router can pin a guard the real application does not rely on, which is a trap worth remembering for the next handler-level fix.

Issue #2805, opened 2026-07-22
2FA enforcement policy: require TOTP enrollment for admin (or all) users


brandonrc:
TOTP 2FA is fully implemented per-user (enrollment, backup codes, login challenge) but is strictly opt-in. Operators need a system-level policy to require it.

Proposal:
- System setting (env + settings API): totp_policy = disabled | required_for_admins | required_for_all (default disabled, preserving current behavior)
- Users subject to the policy who have not enrolled: on next login they are routed into TOTP enrollment before a session is issued (grace: complete enrollment in that login flow, not a countdown of days)
- Local accounts only; SSO-authenticated users delegate MFA to their IdP and are exempt
- must_change_password-style precedent already exists in the login flow for forced actions, so the enrollment gate can reuse that pattern

Issue #1315, opened 2026-05-21
Add router-level test that pins the count of token-minting endpoints and asserts each rejects admin scopes from non-admin callers


brandonrc:
## Summary

There are now four token-minting endpoints in the API tree:

1. \`POST /api/v1/auth/tokens\` (\`auth.rs::create_api_token\`)
2. \`POST /api/v1/profile/access-tokens\` (\`profile.rs::create_access_token\`)
3. \`POST /api/v1/users/:id/tokens\` admin path (\`users.rs::create_api_token\`)
4. \`POST /api/v1/users/:id/tokens\` self path (\`users.rs::create_api_token\`, via #1258)

PRs #1261 and #1306 patched the scope-content policy gate on each one. A 5th endpoint added in the future would silently re-open the privilege-escalation hole if the author forgot to call \`enforce_admin_only_scopes\`.

## Proposal

Add a router-level integration test that:

1. Enumerates every \`POST\` route in the assembled router whose path contains \`tokens\` (regex match on the route table)
2. Asserts the count matches an exact constant (currently 4); fails CI if count drifts up without an explicit decision
3. For each enumerated route, fires a request with a non-admin auth context and \`scopes: ["admin"]\` (and \`["*"]\`)
4. Asserts each returns 403 FORBIDDEN

## Acceptance

- Test lives at \`backend/tests/token_mint_surface_test.rs\` or similar integration location
- Test fails loudly when a new token-mint route is added without updating the count constant
- Test fails when any token-mint route accepts admin-class scope from a non-admin
- Documented in a code comment why the constant pin is load-bearing

## Why not in v1.2.0

Adds a structural safeguard rather than fixing the immediate hole. Defer to v1.2.1.

Issue #2477, opened 2026-07-13
docker remote repo always fails on login


major-sam:
### Description

pulling image from AK proxy repository always fail to `invalid username or password`

### Environment Information

- Deployment type: docker compose
- Operating System: debian13
- Architecture: x86_64
- Database: postgre
- Reverse Proxy: caddy
- Network setup: docker network

### Backend Version

1.5.4

### Web UI Version

1.5.4

### Package Format

Docker / OCI

### Expected Behavior

[as described at docs](https://artifactkeeper.com/docs/guides/docker/#docker-hub-remote)

```bash
docker login caddy.proxied.url
docker pull caddy.proxied.url/dockerhub-remote/alpine:3.20
docker pull caddy.proxied.url/dockerhub-remote/nginx:latest
```

### Observed Behavior

```bash
#docker login caddy.proxied.url
Username (docker):
Password:
Login Succeeded
#docker pull caddy.proxied.url/dockerhub-remote/alpine:3.20
Error response from daemon: error from registry: invalid username or password
```

Image layers appears in AK UI

### Steps to Reproduce

1. create docker remote repo with ` "upstream_url": "https://registry-1.docker.io"`
2. try to pull any image

### Logs

```shell
2026-07-13T14:18:06.820217Z  INFO http_request{method=HEAD uri=/v2/docker-proxy/alpine/manifests/3.20 correlation_id=a4c5af70-e1d2-445f-8816-cd9130ee6e77}: artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=a4c5af70-e1d2-445f-8816-cd9130ee6e77 status=401
2026-07-13T14:18:06.820261Z DEBUG http_request{method=HEAD uri=/v2/docker-proxy/alpine/manifests/3.20 correlation_id=a4c5af70-e1d2-445f-8816-cd9130ee6e77}: tower_http::trace::on_response: finished processing request latency=0 ms status=401
2026-07-13T14:18:06.820967Z DEBUG http_request{method=POST uri=/v2/token}: tower_http::trace::on_request: started processing request
[...]

PR #3625, opened 2026-09-01
feat(auth): enforce admin-configurable API token expiration policy (#3460)

### LINKED ISSUE #3460: Enforce Mandatory API Token Expiration Policy for Admins
### Problem Statement

Currently, users can generate API tokens with no expiration date when creating tokens. This poses a security risk as long-lived credentials increase the attack surface. Admins need the ability to enforce mandatory token expiration policies to ensure tokens must be renewed periodically (e.g., every 90 days), reducing the impact of leaked or compromised credentials.

### Proposed Solution

Add admin-configurable token expiration policies that allow admins to:

1. **Enforce Mandatory Expiration**: Require all new API tokens to have an expiration date (disallow `None`/never-expire)
2. **Set Minimum/Maximum Days**: Define acceptable expiration ranges (e.g., 1-90 days)
3. **Set Default Value**: Automatically apply a default expiration (e.g., 90 days) if user doesn't specify one
4. **Admin Settings UI/API**: Provide an admin endpoint to view and update these policies dynamically without restarting
5. **Audit Logging**: Log when tokens are created under enforced policies for compliance tracking

The implementation should include:
- Configuration options via environment variables (for deployment-time settings)
- Database-backed admin settings table (for runtime policy changes)
- Validation in token creation flow to enforce policy
- Clear error messages when users attempt to violate the policy
- Audit trail recording policy enforcement actions

### Alternatives Considered

1. **Hardcoded enforcement**: Set a fixed max TTL in code - lacks flexibility for different security postures
2. **Periodic background job to revoke old tokens**: Forceful approach that breaks workflows; users prefer explicit renewal
3. **Token warning system only**: Alert admins to long-lived tokens but don't prevent them - weaker security posture

### Use Case

As a DevOps/Security admin managing Artifact Keeper deployments, I want to enforce that all API tokens expire within 90 days so that compromised credentials have limited window of abuse and our organization meets security compliance requirements (SOC2, ISO 27001, etc.).

As an enterprise user, I want to configure different token policies for different environments (strict in production, relaxed in staging) without code changes or redeploys.

### Component

### Additional Context

**Related Code:**
- Token generation logic: `backend/src/services/auth_service.rs::generate_api_token()`
- Token request models: `backend/src/api/handlers/auth.rs::CreateApiTokenRequest`
- Current behavior: `expires_in_days: Option<i64>` allows `None` for no expiration

**Implementation References:**
- Similar to password expiration policy in `backend/src/services/password_expiry_service.rs`
- Configuration pattern exists in `backend/src/config.rs` for JWT expiry settings

**Acceptance Criteria:**
- [ ] Add config options: `API_TOKEN_EXPIRATION_REQUIRED`, `API_TOKEN_EXPIRATION_DAYS_MIN`, `API_TOKEN_EXPIRATION_DAYS_MAX`, `API_TOKEN_EXPIRATION_DAYS_DEFAULT`
- [ ] Create `admin_settings` table to persist runtime policy changes
- [ ] Add `GET/PUT /api/v1/admin/token-policy` endpoints for managing policies
- [ ] Enforce policy in `generate_api_token()` - reject tokens violating constraints
- [ ] Add audit logging for policy-enforced token creation
- [ ] Update API response to indicate when policy enforcement was applied
- [ ] Document in admin guide with example configurations (90-day, 365-day, disabled)
- [ ] Add unit tests for policy validation logic
- [ ] Add integration tests for admin policy endpoints

### Pre-submission Checklist

brandonrc:
> [!IMPORTANT]
> **This PR contains a security fix as well as the feature — please do not merge it as a routine feature.** Four credential-exchange endpoints minted full-lifetime JWTs regardless of how little life the presented credential had left. On `GET`/`POST /conan/{repo_key}/v2/users/authenticate` the minted JWT is itself accepted as the Basic password by the same middleware, so it could be re-presented before each expiry to renew access indefinitely — outliving the API token that anchored it, and escaping any expiration policy. All four now cap the minted JWT at the presenting credential's own expiry. See the `### Security` entry in the CHANGELOG; JWTs already minted this way are not revoked by the upgrade (they die at their own `exp`, and can no longer be renewed).

## Summary

Implements the admin-configurable **API token expiration policy** from #3460: admins can require that newly-minted API tokens expire, bound the acceptable range, and apply a default when a client omits one. Closes #3460.

Design decisions, made deliberately for existing deployments:

1. **Existing tokens are never retroactively expired.** Enforcement is mint-time only, at the single choke-point every token-creating handler funnels through (`AuthService::generate_api_token_with_policy`; personal, profile, repo-scoped, service-account, and admin-minted user tokens all route through it — pinned by the existing #1315 mint-surface test). `validate_api_token` continues to honour whatever `expires_at` (including `NULL`) is on the row, so enabling the policy on upgrade cannot break a running pipeline. Instead, `GET /api/v1/admin/settings/token-policy` reports the live never-expiring inventory (split user vs service-account) so operators rotate deliberately.
2. **Omitted expirations are defaulted, not rejected.** With the policy on, a client that never sends `expires_in_days` gets `default_days` (response and audit record say so via new `expires_at` / `policy_applied` fields); only an explicitly out-of-range request is rejected (400, message names the permitted range). Turning the policy on therefore breaks no existing client.
3. **Ships default-off.** Migration `216` seeds `require_expiration: false` — the exact historical behaviour. Runtime management via `GET/PUT /api/v1/admin/settings/token-policy` (stored in `system_settings`, same pattern as the TOTP policy #2805); the `API_TOKEN_EXPIRATION_*` env vars pin the policy and win over the stored row in both directions (`API_TOKEN_EXPIRATION_REQUIRED=false` + restart = offline break-glass). An internally inconsistent policy (min>max, default outside range) is refused before storage, because a bad range would make every defaulted mint fail.
4. **Service accounts are exempt by default** (`apply_to_service_accounts: false`): a mandatory expiry on a pipeline credential is an outage on a schedule. Compliance-driven deployments can opt them in explicitly.

### Exchange-cap hardening — security fix (same PR, load-bearing for the policy)

`/v2/token` previously minted a full-TTL bearer from any presented credential, on all three arms (API token as docker-login password, Bearer-JWT swap, JWT-as-password CI fallback). Chaining swaps before each 30-minute expiry renewed access indefinitely — so any expiration on the underlying token, including this policy's, could be escaped. All three arms now cap the minted access token's `exp` at the presenting credential's own expiry (`cap_access_expiry`), making swap chains monotonically non-increasing. Verified live: a credential with 5 minutes left yields 299s bearers through all three arms (was 1800s on baseline).

### Upgrade note

Applying this release changes nothing by itself: the seeded policy is inert, existing tokens (including never-expiring ones) keep working before and after the policy is enabled, and token-create responses gain two additive fields (`expires_at`, `policy_applied`). The only behavioural delta without opting in: an OCI bearer minted from a credential that expires within 30 minutes is now correspondingly shorter-lived (clients honour `expires_in`). Docs: `docs/api-token-expiration-policy.md` (90-day / 365-day / disabled examples + break-glass); audit schema gains optional `TokenDetails.expires_at`/`policy_applied` and the `API_TOKEN_POLICY_CHANGED` action.

Web UI for the new admin settings is intentionally out of scope (separate repo; tracked with the web 1.9 catch-up).

- [x] N/A — this is not a bug fix (feature; unit + DB tests included, plus a red→green DTF tier `token-expiry-policy` staged in artifact-keeper-test on branch `dtf/3460-token-expiry-policy` for the batched 1.9.0 tiers MR)

## Test Checklist
- [x] Unit tests added/updated (pure policy resolution/validation/env-pin + `cap_access_expiry`)
- [x] Integration tests added/updated (DB-backed: mint enforcement + SA exemption + pre-existing-token survival; admin GET/PUT round-trip, 400 on inconsistent policy, 409 while env-pinned; `/v2/token` cap on all three arms — all in `--lib` scope with `AK_TESTS_REQUIRE_DB=1`)
- [x] E2E tests added/updated (DTF tier, validated RED on baseline → GREEN on this build)
- [x] Manually tested locally (isolated rig slot: in-place upgrade applies 216 on an existing DB; pre-upgrade never-expiring token authenticates before and after enabling the policy; expired token rejected on raw Bearer, docker-login `/v2/token`, and package Basic auth)
- [x] No regressions in existing tests (full `--lib` suite green; #1315 mint-surface pin and the audit-schema drift canary updated where the feature legitimately extends them)

## API Changes
- [x] New endpoints have `#[utoipa::path]` annotations (`GET/PUT /api/v1/admin/settings/token-policy`)
- [x] Migration is reversible (216 is a seed `INSERT ... ON CONFLICT DO NOTHING` into the existing `system_settings` table; removing the row restores prior behaviour)


brandonrc:
## Audit round 1 — token expiration policy (3 reviewers: security / architecture / test-verification)

Read at head `5ba657c5`. The core design is right and unusually well-defended — mint-time-only choke point, default-rather-than-reject, service-account exemption, never-expiring inventory, and each decision carries a comment saying why. It mirrors the TOTP sibling (`totp_policy.rs`, migration 198) closely enough that knowing one means knowing the other.

**Claims audit** (I tried to break each one):

| Claim | Result |
|---|---|
| Mint choke point is total | ✅ **Confirmed.** Only two `INSERT INTO api_tokens` exist: `auth_service.rs:2573` and a `#[cfg(test)]` seed at `users.rs:3233`. Legacy `generate_api_token` is now a wrapper delegating at `:2490`. No CLI/seed/bootstrap/import/SSO mint path. |
| Cap on all three `/v2/token` arms | ❌ **Refuted — see finding 1.** |
| Never retroactive | ✅ Confirmed. `validate_api_token` expiry logic unchanged; policy read only on mint. |
| Ships default-off | ✅ Confirmed. `default()`, migration 212 seed, and `resolve_expiry`'s short-circuit all agree. |
| Hot path untouched | ✅ Confirmed. `validate_api_token` gained exactly one field assignment from an already-selected column (`auth_service.rs:2464`). No new query, no new join. |
| AuthZ on new endpoints | ✅ Confirmed. Same `admin::router()` that already carries `/settings/totp-policy`, nested under `admin_middleware` at `routes.rs:851`. PUT returns 409 while env-pinned rather than silently no-op'ing. |
| Migration 212 idempotent/safe | ✅ Confirmed. `ON CONFLICT (key) DO NOTHING`, unique slot, `updated_by` nullable so no FK failure on a fresh DB. |
| Audit schema evolution | ✅ Confirmed backward-compatible. Two optional props appended, `required` unchanged, `additionalProperties: true` at all ten levels, `skip_serializing_if` on both. `audit_schema.rs` validation is `#[cfg(test)]`-only. `audit_log.action` is a plain `VARCHAR(100)` with no CHECK, so the new action value can't silently fail its insert. |

---

### 1. 🔴 The Bearer-swap arm has a fourth sub-branch that is uncapped and self-renewing

`oci_v2.rs:4628-4643`. The arm splits **before** the cap is applied. When the presented JWT carries a `scan_pull_repo` claim it takes `generate_scan_token`, which mints `exp = now + scan_token_ttl_seconds` (`auth_service.rs:1515-1543`) and reports `expires_in = config.scan_token_ttl_seconds` — a fixed fresh TTL, **never compared against `claims.exp`**. Only the `else` at `:4645` got `generate_tokens_with_scope_capped`.

The output of that branch is itself a JWT carrying `scan_pull_repo`, so it re-enters the same branch on next presentation.

**Failure:** a scan token for `docker-private-a` leaks (CI log, scanner sidecar, proxy capture) with 4 minutes left. `GET /v2/token` with `Authorization: Bearer <scan_token>` → 200 with a fresh 300s scan token for the same repo. Loop every 4 minutes → **unbounded pull access**, surviving the underlying API token's `expires_at` and every policy setting. This is exactly the renewal chain `token_expiry_policy.rs:340-344` says the cap exists to close.

Bounded by `enforce_scan_pull_scope` to single-repo *pull*, and deactivation/credential-change still kill it — but the lifetime is unbounded.

The new test `jwt_swap_and_jwt_as_password_never_extend_lifetime` mints its probe via `generate_tokens_with_scope_capped` with **no** `scan_pull_repo`, so it only exercises the `else` branch and structurally cannot catch this.

**Fix:** derive `ttl_seconds` as `min(scan_token_ttl_seconds, claims.exp - now)`.

### 2. 🔴 The never-expiring inventory count is an unindexed full-table aggregate, and PUT pays it twice

`admin.rs:1067-1085` — `COUNT(*) FILTER (...) FROM api_tokens JOIN users WHERE at.expires_at IS NULL AND at.revoked_at IS NULL AND u.is_active`.

Every existing index is unusable: `idx_api_tokens_prefix`/`idx_api_tokens_user` (`006_api_tokens.sql:15-16`) don't match the predicate, and `idx_api_tokens_revoked` (`068_token_revocation.sql:5`) is partial on `WHERE revoked_at IS NOT NULL` — the exact inverse. No index on `expires_at` exists and 212 adds none. Plan is a full parallel seq scan plus hash join.

`api_tokens` is wide (`token_hash`, `scopes[]`, `last_used_user_agent`, `description`, `repo_selector` JSONB) — roughly 1–2.5 GB at 5M rows. `build_token_policy_response` calls it unconditionally at `:1091`, so **PUT runs it twice** (pre-image `:1145`, response `:1172`). Beyond latency, the scan evicts precisely the buffer-cache pages `validate_api_token` depends on — an admin opening a settings page degrades the auth hot path this PR otherwise protects carefully.

The TOTP sibling gets away with this shape because it aggregates over `users`; copying it to `api_tokens` is where it stops being cheap.

**Fix:** in migration 212 —
```sql
CREATE INDEX idx_api_tokens_never_expiring_live ON api_tokens (user_id)
  WHERE expires_at IS NULL AND revoked_at IS NULL;
```
Plain `CREATE INDEX`, not `CONCURRENTLY` (`sqlx::migrate!` wraps each file in a transaction). And compute the counts once per PUT. Secondary: this is runtime `sqlx::query_as` into a bare `(i64, i64)` while the sibling uses compile-time-checked `sqlx::query!` — it opts out of the schema verification the rest of the file relies on.

### 3. 🟠 The env break-glass fails silently in the *stricter* direction — contradicting its own documented contract

`token_expiry_policy.rs:157-166`. `parse_days` returns `Option<i64>` and the struct literal uses `?`, so a malformed value in **any** refinement variable discards the entire pin. `from_env_values` → `None` → "no pin" → the stored row governs.

**Failure:** an operator locked out of the UI sets `API_TOKEN_EXPIRATION_REQUIRED=false` to break glass, but a stale `API_TOKEN_EXPIRATION_DAYS_MAX=90d` sits in the same env file. Pin discarded, stored *enforcing* row governs, enforcement stays on. Only signal is a `warn!`.

`docs/api-token-expiration-policy.md` and `config.rs:1331-1339` both promise *"a typo can neither reject every mint nor silently disable stored enforcement."* True as written — but the actual failure mode is that a typo silently **preserves** enforcement when the operator was trying to remove it. That is the one thing a break-glass must not do. On an env-only deployment (the posture the pin is designed for) the stored row is 212's default-off, so the same bug turns the feature *off* instead.

Related, `:143-145`: when `REQUIRED` is unset but refinement vars are set, it returns `None` before any logging at all.

**Fix:** `API_TOKEN_EXPIRATION_REQUIRED` alone establishes the pin; a malformed refinement var falls back to its default *with a warning* rather than discarding the pin. Consider failing startup on a malformed pin, as with a bad `METRICS_PORT`.

### 4. 🟠 `INCLUDE_SERVICE_ACCOUNTS=TRUE` silently means false — found independently by all three reviewers

`token_expiry_policy.rs:190-193`:
```rust
apply_to_service_accounts: matches!(
    include_service_accounts.map(str::trim),
    Some("true") | Some("1")
),
```
Case-**sensitive** exact match, no warning. Twelve lines above, `ENV_REQUIRED` does `to_ascii_lowercase()`, accepts `true|1|false|0`, **and warns** on anything else.

**Failure:** an operator meeting a compliance requirement sets `API_TOKEN_EXPIRATION_REQUIRED=TRUE` (works — lowercased) and `API_TOKEN_EXPIRATION_INCLUDE_SERVICE_ACCOUNTS=TRUE` (silently false). `GET .../token-policy` reports `require_expiration: true, source: environment`, so the console says enforcement is on — while every service-account mint keeps producing **never-expiring CI credentials**. Zero diagnostic signal. Service accounts are the highest-value token class here, and this is a silent loosening of exactly the population the feature exists to cover.

**Fix:** route it through the same lowercased, warning parser as `required`.

### 5. 🟠 `deny_unknown_fields` on the *storage* type makes enforcement fail open on rollback

`token_expiry_policy.rs:64` — the same type is both the PUT request DTO (`admin.rs:1061`) and the deserialization target for the stored JSONB row (`:290`). `stored_policy` returns `default()` (i.e. **not enforced**) on any parse error.

**Failure:** N+1 adds a sixth policy field, an admin PUTs it, deployment rolls back to N. Row now has an unknown key → deserialize errors → enforcement silently off, one warn line. Same from any ops hand-edit that adds a key. A transient DB fault has the same effect: the policy `SELECT` fails open to inert while the token `INSERT` (different statement, likely different pooled connection) succeeds — so a client retrying during a load spike can land a never-expiring token on an instance whose admin API reports full enforcement, and the audit record can't distinguish it (`policy_applied: false` looks identical to the exempt case).

**Fix:** split the types. Keep `deny_unknown_fields` on the request DTO (good 400s on typo'd fields); make the stored representation lenient. At minimum, escalate a parse failure on an *existing* row from warn to a hard error at the mint path — a corrupt policy row is an operator error, not a transient one.

### 6. 🟠 Two silent-downgrade paths through PUT

PUT is full-replace, and the `#[serde(default)]` placement is backwards relative to risk:

- **`apply_to_service_accounts` omitted** → flips to `false`. An admin narrowing `max_days` from 30 to 14 with an otherwise-complete body silently reverts every service-account mint to never-expiring. 200 OK, no warning. (Audit records `from`/`to`, so it's forensically recoverable — not prevented.)
- **`default_days` omitted** → becomes `None`, which per `resolve_expiry:246-251` means **reject every mint that omits an expiry**. Every legacy client and script breaks, from an omitted field, with a success response.

Note the asymmetry: `deny_unknown_fields` means a *misspelled* field correctly 400s, while an *omitted* one silently changes behavior — in one case looser, in the other strictest.

**Fix:** make both explicitly required in the request DTO so each choice is a deliberate act; or have the handler diff against the current policy and refuse a silent downgrade.

### 7. 🟠 Verification gaps — the `#1315` claim doesn't hold, and one mutation makes the feature a no-op with CI green

The PR body says the mint surface is "pinned by the existing #1315 mint-surface test." The test is real (`token_service.rs:1263`) with a mirrored CI gate (`scripts/ci/check-token-mint-surface.sh`), but it does two things: assert the *set of handler files* containing the substring `generate_api_token`, and assert each still contains `enforce_admin_only_scopes`/`require_admin`. It pins the **#1315 privilege-escalation authz surface**, nothing about expiration — and `generate_api_token` is a *substring* of `generate_api_token_with_policy`, so the pin is blind in both directions. `token_service.rs` and the gate script are untouched by this PR, which also falsifies the checklist line about the pin being "updated where the feature legitimately extends them."

**The mutation that matters:** in the INSERT at `auth_service.rs:2571-2583`, replace the `expires_at` bind with `None` while still returning `expires_at` in `MintedApiToken`. **Every test passes.** The choke-point test asserts the returned struct; `admin.rs:2517`'s inventory assertions are `>= 1` (more never-expiring rows only helps); the `/v2/token` arm-1 test would then read `NULL` → no cap → still `== base_ttl`. Every token minted under an enforcing policy would never expire, CI green. Two lines fix it: `SELECT expires_at FROM api_tokens WHERE id = $1` and assert it equals `minted.expires_at`.

Other gaps, briefly:
- **No mint path is tested through its handler.** All five were edited; all five received only serde-shape stubs (`expires_at: None, policy_applied: false` added to existing struct tests). `repo_tokens.rs` and `service_accounts.rs` do post-mint writes and now have a mint that can return `Err(Validation)` where it previously could not — nothing tests that a rejection is a clean 400 rather than a half-created row.
- **The `API_TOKEN_EXPIRATION_*` env plumbing is never read from an actual env var.** `from_env_values` is called with literal strings; the 409 test sets `cfg.api_token_expiry_policy` directly. Transpose `ENV_DAYS_MIN`/`ENV_DAYS_MAX` at `config.rs:1331` and the pin silently becomes `min > max` → validate fails → pin dropped → enforcement silently falls back. All tests stay green.
- **Migration 212's seeded value is unasserted.** A typo flipping the seed to `true` — breaking the "applying this release changes nothing" guarantee — fails nothing.
- **Arms 2/3 have no lower bound** (`:34941`, `:34948` assert only `<= 300`; arm 1 correctly brackets `>= 200`). Two lines.
- **`expires_at = NULL` isn't covered at the handler** — arm 1's negative control mints `Some(30)`, never `None`. If `ApiTokenValidation.expires_at` were ever populated wrong, every legacy never-expiring `docker login` would start getting ~0-second bearers — an outage for exactly the deployments this design protects. One extra case in the existing test.

Boundary coverage of the pure policy, for the record, is **complete and good** — `:410` rejects `0, -5, 91, 3650` and accepts `1, 45, 90`; the SA exemption is pinned in both states; overflow is genuinely unreachable (`validate()` bounds both paths, plus a belt-and-braces `clamp(1, 3650)` at `:2537`).

### 8. 🟠 These tests introduce a real cross-test flake

`token_policy_serial_lock` (`test_db_helpers.rs:274-290`) correctly uses a pg advisory lock — right primitive under nextest's process-per-test model, and its doc comment states the problem exactly: the policy is **one row shared by the whole database, and every mint reads it**.

The three policy tests take the lock. **The two new `/v2/token` tests do not**, and neither do ~9 pre-existing DB-backed minting tests (`middleware/auth.rs:4691/4763/4853/4891`, `guest_access.rs:725`, `oci_v2.rs:14062/14229/23820/24153`).

**Concrete flake:** `token_expiry_policy.rs:547` stores `min_days: 2` for its test body. CI runs `--test-threads 8`. Any concurrent mint with `expires_in_days: Some(1)` is now rejected with `AppError::Validation` — which is exactly what `oci_v2.rs:34884` (`"cap-soon"`, `Some(1)`) and `middleware/auth.rs:4763` do, both under `.expect(...)`. Hard panic, nondeterministically.

**Fix:** the advisory lock only helps if every participant takes it. Either add it to the minting tests, or (cheaper) put the policy-mutating tests in a `db-serial` group in `.config/nextest.toml` with a filter covering the minting tests.

### 9. 🟡 Split the PR — the seam is clean

The `/v2/token` cap is independent code, not a consequence of the feature. `cap_access_expiry` (`:345-353`) is a 9-line `min()` taking no policy, reading no setting, touching no DB. Call graphs are provably disjoint: `generate_tokens_with_scope_capped` has three production callers, all in `oci_v2.rs`; `generate_api_token_with_policy` has five, none in `oci_v2.rs`. The cap fires **unconditionally**, including where the policy is inert — so every deployment gets a `docker login` behavior change whether or not they enable the feature. ~300 of the 1,824 lines.

`CONTRIBUTING.md:86` asks for focused PRs. There's precedent for the bundled shape (`d89d7b96`), but it **named the second concern in the title**; here it appears only in the third paragraph of the body.

**Seam:** PR A (`fix(oci)`, ~300 lines, 3 files) = all of `oci_v2.rs`, plus in `auth_service.rs` the `ApiTokenValidation.expires_at` field (`:117-124`, `:2464`), `generate_tokens_with_scope_capped` (`:1363`), `generate_token_pair_capped` (`:1432`), the `expires_in` recompute (`:1495`), and the ~11 `expires_at: None` test fills; relocate `cap_access_expiry` out of `token_expiry_policy.rs`. PR B = everything else. No function is edited by both halves. Given finding 1, the cap is the more urgent half and shouldn't wait on review of a 1,500-line feature.

The other 19 files are legitimate — 13 are one-line `api_token_expiry_policy: None,` fills in `Config` test fixtures forced by the new struct field.

### 10. 🟡 `policy_applied: true` when the policy changed nothing

`:258-262` returns `true` for an explicit, in-range `expires_in_days`. A client asking for 30 days under a 1–90 policy is marked identically to one whose omitted expiry was defaulted. And `with_expiry` sets the field unconditionally (`audit_export.rs:512`), so **every** `API_TOKEN_CREATED` event now carries it.

The docs sell this field for "compliance review," but a reviewer filtering audit logs for `policy_applied: true` to find "tokens whose lifetime the instance chose" gets every enforced mint instead. The field can't answer the question it was added for. Either set `false` on pass-through, or replace the boolean with `expiry_source: "requested" | "policy_default"`.

### 11. 💭 Smaller items

- **`TokenService::create_token`** (`token_service.rs:262-284`) routes through the choke point, so stored `expires_at` is correct — but it **recomputes** the value it reports from `request.expires_in_days`, enforces its own `1..=365` contradicting a policy allowing up to 3650, and drops `policy_applied`. Currently unreachable (only caller is in `#[cfg(test)]`), so latent — but wire it to a route and callers get a date the DB disagrees with. Fix or delete.
- **The CI mint-surface gate** is scoped to `HANDLERS_DIR` with a non-recursive `listdir`. A mint under `services/**` (precisely `token_service.rs:274`), `middleware/**`, or a handlers subdirectory never trips it. Widen to all of `backend/src` now that a second mint-bearing module exists.
- **`generate_token_pair_capped`** caps the access token but not `refresh_exp`. Not currently exploitable — arm 2 hardcodes `refresh_token: None`, arms 1 and 3 set `authenticated_via_api_token = true` which makes `offline_refresh_token` return `None` unconditionally — but the name promises the pair is capped, and one future caller returning `pair.refresh_token` reopens the chain. Cap it or rename.
- **Mangled 409 string** at `admin.rs:1148` — a wrapped literal without `\` continuations leaves two runs of ~14 literal spaces in the message an operator sees at the exact moment they're confused about why their write was refused. The sibling at `totp_policy.rs:198-202` does it correctly.
- **Dead conflicting key:** `012_system_settings.sql:66` seeds `security.api_token_max_age_days = 365` with zero readers repo-wide. `system_settings` now has two keys claiming to govern token lifetime, disagreeing (365 vs 90), with nothing marking which is live. Have 212 delete it or mark it superseded.
- **`CHANGELOG.md` not updated** — 23 of the last 25 commits to main touch it, and there's a release-bookkeeping CI gate (#3539). Given the `/v2/token` behavior change this needs an upgrade note, not just a feature line.
- **`docs/audit-events.md:99`** example is stale per finding 10. Also `expires_at` is declared bare `"type": "string"` at `audit-event.v1.schema.json:206-209` despite being `to_rfc3339()`; other envelope date fields carry `"format": "date-time"`.
- **Follow-up, not this PR:** the policy-change audit entry uses `ResourceType::User` with `.resource(auth.user_id)` (`admin.rs:1158-1160`), recording the acting admin as the *resource* of a global setting change. `ResourceType::Setting` exists and is used this way at `webhooks.rs:440` — but the TOTP sibling has the identical bug at `admin.rs:1013-1015`, so fix both together.
- **Informational:** `handle_refresh_grant` (`oci_v2.rs:4444-4490`) short-circuits before credential resolution and mints uncapped. Pre-existing and correctly gated (offline tokens issue only on password auth), but the registry offline token is now the longest-lived exchangeable credential on the instance and is untouched by the policy. Worth a sentence in the docs so operators don't assume the policy bounds all registry credentials.

---

**Suggested gating:** finding 1 blocks — it refutes a stated invariant and creates an unbounded-lifetime credential. Findings 2, 3, 4, 5, 6 before 1.9.0: 4 and 6 are silent fail-open transitions on the service-account class, which is precisely the population this feature exists to bring under control. Finding 7's two-line row assertion is the cheapest high-value fix in the list. Finding 8 will bite CI regardless of review.

<sub>Round 1 of an automated multi-reviewer audit (security / architecture / test-verification passes). Findings 3, 4 and 5 were reached independently by more than one reviewer.</sub>

brandonrc:
Taken out of draft. Merged `main` in, renumbered the migration, and did the merits review nobody had done yet — which turned up one real gap, fixed here.

## Conflict resolution

One conflicted file, `backend/src/api/handlers/oci_v2.rs`, two hunks. Both sides appended a new `#[cfg(test)] mod` at the end of the test section and both module bodies open with the same four lines (`let Some(pool) = tdh::try_pool().await else { return; }; let (user_id, _u) = tdh::create_user(&pool).await;`), so git interleaved them into two hunks instead of treating them as two independent additions.

Resolved by keeping main's `public_read_repo_scope_3704` module whole and re-appending this branch's `token_exchange_expiry_cap_3460` module after it. Verified mechanically rather than by eye: computed the set of lines each side *added* relative to the merge base `3c2017ec` and asserted every one survives in the resolution — 174 branch-added lines, 3146 main-added lines, 0 missing on either side. The only lines in the resolution that differ from main are the three production `generate_tokens_with_scope` → `generate_tokens_with_scope_capped` call sites, which is precisely this branch's change.

## Migration renumber

`212` → **`216`**. Main took `212` (jupyter format), `214` and `215` while this branch was open; `213` is claimed by open PR #3549. `scripts/ci/check-duplicate-migrations.sh` passes on the merge result (198 unique versions, no duplicates; the gap warning lists 213 as expected). No other reference needed updating — nothing in the tree cited the file by name — and the PR body's three mentions of 212 are updated.

## CHANGELOG

The branch predates the 1.9.0 cut and carried no bullet at all, so one is added under the **current** `## [Unreleased]`, with two upgrade notes. Confirmed with `awk '/^## \[/{h=$0} /^- \*\*/{print h}' CHANGELOG.md | sort -u` that the new bullet's section is `[Unreleased]` and not `[1.9.0]` — the misfiling trap from #3797.

## Review on the merits

**1. What happens to existing tokens when an admin first sets a maximum lifetime?** They are **grandfathered**, not retroactively expired or re-dated. Enforcement is mint-time only, at `AuthService::generate_api_token_with_policy`; `validate_api_token` keeps honouring whatever `expires_at` (including `NULL`) is on the row. That matches what #3460 asked for — it proposes enforcement on *new* tokens and explicitly rejects "periodic background job to revoke old tokens" as too forceful. Proven by the last assertion in `test_api_token_mint_enforces_expiry_policy_but_never_existing_tokens`: a token minted while the policy was inert still authenticates after full enforcement is switched on. It is in the CHANGELOG as an upgrade note, along with the fact that an omitted `expires_in_days` is *defaulted* rather than refused, so no client that worked before starts failing.

**2. Can the policy be bypassed by the token-exchange paths?** It could, through a **fourth** arm the branch missed. `GET`/`POST /conan/{repo_key}/v2/users/authenticate` mints a JWT from whatever credential authenticated the request (`conan.rs`, uncapped `generate_tokens_with_scope`). Conan routes run under `repo_visibility_middleware`, the one middleware that passes `allow_basic_api_token = true`, so an API token is accepted as the Basic password there — and so is a JWT. The minted JWT is therefore itself a valid credential on the same endpoint: present a token with 5 minutes left, get a 30-minute bearer, re-present the bearer before each expiry, and access outlives the token forever. Exactly the escape the `/v2/token` work closes, one route over.

Fixed in `bd81156`. `AuthExtension` carries no credential expiry (only `iat_ms`), and threading one through `try_resolve_auth_outcome` would touch 80+ fixture literals, so the handler re-derives it from the same `Authorization` header in the same order the middleware tries it (bcrypt password → JWT → API token); both re-validations hit the caches the middleware populated moments earlier, so there is no extra bcrypt work. The mint then goes through `generate_tokens_with_scope_capped` like the OCI arms.

I swept the rest: `auth.rs` login, `totp.rs`, `sso.rs` and `ci_auth.rs` all mint from a *fresh* authentication (password, TOTP, IdP assertion), not from an AK credential being exchanged, so a cap does not apply. `/v2/token`'s offline refresh token was already suppressed for API-token-authenticated exchanges by #2477/#2487, so the 30-day refresh token is not an escape hatch either. The `#1315` mint-surface pin still matches (5 endpoints) — `generate_api_token_with_policy` contains the pinned substring, so the guard did not go blind.

**3. Is the admin setting protected?** Yes on both counts. `GET`/`PUT /api/v1/admin/settings/token-policy` sit on the admin router behind `admin_middleware`, and every `PUT` writes an `API_TOKEN_POLICY_CHANGED` audit record carrying the actor, the full before/after policy, and the never-expiring inventory at the time of the change. A `PUT` is refused with 409 while the `API_TOKEN_EXPIRATION_*` env pin is active, and with 400 for an internally inconsistent policy before anything is stored.

**4. What does an expired token return?** 401 with a distinguishable, non-leaking reason: `AppError::Authentication("API token expired")` → `401 AUTH_ERROR`, which says the credential expired without revealing the owner, the token name, or whether the token ever existed. Worth noting for a follow-up: the OCI `/v2/token` arm flattens every credential failure to `401 UNAUTHORIZED "invalid username or password"`, so a `docker login` with an expired token cannot tell expiry from a typo. That is pre-existing behaviour and deliberately conservative, not something this PR changes.

Two smaller things fixed while in there: a collapsed multi-line string literal in the 409 body of `update_token_policy` (it rendered as runs of 14 spaces mid-sentence), and the doc's claim that only `/v2/token` caps exchanges.

## Verification

Everything through `rig/compile-lock.sh`, `SQLX_OFFLINE=true`, throwaway Postgres 16 on 5725 with `AK_TESTS_REQUIRE_DB=1` (so a DB-backed test cannot silently no-op).

- `sqlx migrate run --source backend/migrations` — 216 applies cleanly in sequence on a fresh DB.
- `cargo fmt --all -- --check`, `cargo clippy --lib --tests --all-features -- -D warnings`, `cargo check --lib --tests` — all clean.
- `cargo nextest run --lib` over the auth / token / oci_v2 / conan / admin / middleware areas: **1417 passed, 0 failed**.
- `scripts/ci/check-duplicate-migrations.sh`, `scripts/ci/check-token-mint-surface.sh`, `scripts/ci/check-changelog-unreleased.sh` — all pass.

**Fails without the fix.** Reverting the two production hunks — `cap_access_expiry` made to return `base_exp` unconditionally, and `resolve_expiry` made to pass the request through untouched — fails all four behavioural tests, with messages that name the bug rather than just a diff:

| test | failure on reverted production code |
| --- | --- |
| `services::auth_service::tests::test_api_token_mint_enforces_expiry_policy_but_never_existing_tokens` | `default stamped an expiry` |
| `api::handlers::oci_v2::token_exchange_expiry_cap_3460::api_token_exchange_is_capped_at_the_tokens_expiry` | `bearer must not outlive the API token (remaining ~300s): got 1800s` |
| `api::handlers::oci_v2::token_exchange_expiry_cap_3460::jwt_swap_and_jwt_as_password_never_extend_lifetime` | `a Bearer swap must not extend the presented JWT's lifetime: got 1800s` |
| `api::handlers::conan::agent2_recipe_reads::test_3460_conan_authenticate_caps_the_minted_jwt_at_the_credential_expiry` | `the Conan exchange must not outlive the API token (~300s left): got 1800s` |

All four pass on the real tree. Each capped test also carries a negative control (a far-future credential still gets the uncapped base TTL), so the cap is shown to narrow only when it should.

Not merging — leaving this for review.

PR #3627, opened 2026-09-01
fix(migration): log the interruption cause the worker can actually observe (#3498)

### LINKED ISSUE #3498: Migration worker's cancel branch is unreachable: spawn_migration_worker's CancellationToken is never retained, so real cancels log "Migration paused by user"

brandonrc:
Closes #3498

## Summary

The migration worker's cancel branch could never execute, so **every real user cancel was logged `"Migration paused by user"`**.

`spawn_migration_worker` builds a fresh `CancellationToken`, moves it into the worker, and nothing in production retains a handle to it — there is no registry of running workers and no production path calls `.cancel()` on it (the only `.cancel()` in the file is the test fixture). The real cancel path is a pure database write: `cancel_migration` sets `status = 'cancelled'`, which the worker observes through `is_paused(job_id)` (it matches both `'paused'` and `'cancelled'`). So in `halt_on_interrupt`, `if self.cancel_token.is_cancelled()` is always false for an operator-driven stop and the `else` arm — "Migration paused by user" — always won.

It is cosmetic to the job *row* (`cancel_migration` already wrote `'cancelled'` synchronously before the worker reaches this line), which is exactly why the only honest reproduction asserts on the **log line**, not the row. But it makes pause and cancel — the two states an operator most needs to distinguish when diagnosing a stopped migration — indistinguishable in the logs.

## What changed

- `halt_on_interrupt` keeps writing the terminal `Cancelled` status when the in-process token *is* cancelled (a token cancel leaves no trace in the DB, so the worker is the only writer there — that arm is retained for correctness), then `return`s. Otherwise it re-reads the job status the worker can actually observe and logs the transition that happened: `cancelled`, `paused`, or the row having been deleted mid-run.
- A small `job_status` helper reads with `fetch_optional`, so a deleted row is `None` rather than the `RowNotFound` error `is_paused`'s `fetch_one` raises.
- The finalize-guard exit (`finalize_job_status` matching zero rows) previously logged the ambiguous `"Migration paused or cancelled before finalization"`, which also covered "the row was deleted mid-run". It now names the observed status (`Migration stopped before finalization; ... status=<paused|cancelled>`) or the deleted-row case explicitly — the related wording bug called out in the issue.

**`spawn_blocking` note (per the review brief):** not implicated here. This worker's cancellation is observed cooperatively (`is_paused` between artifacts/repos and the token check), never by dropping a `spawn_blocking` handle. The token being non-functional was a missing-handle problem, not a "dropping a blocking handle does not cancel" problem.

The observable defect is a **log line**, so the tests capture logs (a thread-local `tracing` layer) rather than only asserting row state (the row was already correct before the fix). DB-backed, run under `AK_TESTS_REQUIRE_DB=1`:

- `test_api_cancel_is_logged_as_cancel_not_pause_3498` — an API cancel (a status write, the production path) must log "Migration cancelled by user" and must NOT log "Migration paused by user".
- `test_api_pause_is_still_logged_as_pause_3498` — control: a pause is still logged as a pause and not a cancel (stops a "fix" that relabels every interruption as a cancel).
- `test_finalize_guard_skip_names_the_observed_status_3498` — an interruption landing after the last per-artifact check hits the finalize guard; the log must carry the re-read status (`status=paused`), not the ambiguous line.
- `test_job_status_returns_none_for_a_missing_row_3498` — the `fetch_optional` shape under the deleted-row log lines.

Each hunk was revert-proven individually (reverting the halt-on-interrupt logging, the finalize-guard logging, or the `fetch_optional` change fails exactly the corresponding test while controls stay green). An end-to-end deployment tier (`migration-cancel-log`, artifact-keeper-test) validates the same distinction against a live backend: it is RED on the parent image (the cancel logs "paused") and GREEN on this change.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [ ] N/A - no API changes


brandonrc:
## Verification — merged `main`, reviewed, out of draft

Merged `origin/main` @ `eb5250a5` (post-1.9.0 cut; #3631/#3630/#3620 landed today). **No conflicts** — this branch touches only `migration_worker.rs` and the CHANGELOG, neither of which main moved. Diffed the merge against both parents: `merge vs main` is exactly this branch's 2-file change, `merge vs branch` is exactly main's 15-file change. Nothing dropped in either direction.

**CHANGELOG:** the bullet is under `## [Unreleased] / ### Fixed`, confirmed *after* the merge — `awk '/^## \[/{h=$0} /^- \*\*/{print h}' CHANGELOG.md | sort -u` reports `[Unreleased]` for the `(#3498)` bullet, zero conflict markers.

### Merits — does it fix the failure mode, or only make it visible?

The failure mode in #3498 *is* the log line, so this is the fix rather than a diagnostic for it. The job row was already correct; the thing an operator reads was not.

The load-bearing claim is that the logged cause must be one the worker can actually observe. I did not take that from reading the log line — I checked it empirically, and have added the check as a test:

`test_the_real_api_cancel_is_what_the_worker_observes_3498` drives **`POST /api/v1/migrations/{id}/cancel` through the real router and the real handler**, then asks the worker's own primitives what they see:

- `cancel_token.is_cancelled()` → **false** (a real cancel never fires the in-process token)
- `is_paused(job_id)` → **true** (this is what actually produces `Interrupted`)
- `job_status(job_id)` → **`Some("cancelled")`** (the cause the worker can observe)

Every other test here reaches the exit through a mock that writes the status by hand; this one pins that the production path really produces what the mock simulates. It also fails loudly if someone later makes the token real (the issue's second option), pointing them at this exit.

Corroborating the premise statically: the only `.cancel()` on this token anywhere in `backend/src` is the test fixture, and `cancel_migration` is a pure `UPDATE migration_jobs SET status = 'cancelled', finished_at = NOW()`.

### One change on review

The catch-all arm logged `"Migration paused by user"` for *any* non-`cancelled` status. `is_paused` matched `paused` **or** `cancelled`, so any other value means the row moved again between that check and the re-read — a resume, most likely — and defaulting to "paused" there is the same guess-a-cause defect this exit exists to remove. It now names the status actually read. Narrow, but it is the PR's own thesis applied to its own fallback.

### Tests — fail-without-fix, proven per hunk

`AK_TESTS_REQUIRE_DB=1`, throwaway Postgres, `cargo nextest`. Each production hunk reverted individually:

| reverted hunk | fails | controls |
|---|---|---|
| `halt_on_interrupt` re-read logging → pre-fix `if token {cancel} else {pause}` | `test_api_cancel_is_logged_as_cancel_not_pause_3498` | other 4 green |
| finalize-guard logging → ambiguous `"paused or cancelled before finalization"` | `test_finalize_guard_skip_names_the_observed_status_3498` | other 4 green |
| `job_status` `fetch_optional` → `fetch_one` | `test_job_status_returns_none_for_a_missing_row_3498` | other 3 green |

Exactly one targeted failure per revert, controls green in every case. The premise test is deliberately independent of the production hunks — it pins the input, not the output.

`cargo fmt` clean · `cargo clippy --lib --tests --all-features -D warnings` clean · `cargo check --lib --tests` clean · `cargo nextest --lib` over `migration_worker`: **202 passed, 0 failed**.

Issue #3084, opened 2026-08-03
Backup claim renewal failure does not cancel the in-flight backup worker (double backup under DB partition)


brandonrc:
The claim-renewal loop added in #3072 correctly stops renewing when it loses ownership, but it does not **cancel the work already in flight**. Under a DB partition (or any renewal failure while a backup is actively running), the ledger stays consistent — one `backup_schedule_runs` row — while the external side effect does not: the original worker keeps writing its archive after another replica has legitimately reclaimed the occurrence.

So the durable record says one backup; storage may hold two.

Fix direction: thread a cancellation token from the renewal guard into the backup worker so a lost claim aborts the in-flight archive rather than only stopping renewal.

Identified during the #3072 multi-replica red-team; scoped out of that PR (it is the backup subsystem, not the Incus tail that PR fixed).

Issue #3503, opened 2026-08-22
Scheduled GC tick: blob-GC mark/sweep and the storage-stats recompute stay un-leased on every replica, so #3385's singleton lease covers only one of four workloads


brandonrc:
## Summary

The scheduled storage-GC tick in `scheduler_service.rs` runs **four** workloads back to back. The singleton lease proposed in PR #3385 covers only the first. The other three — including a destructive one — still run on **every replica in the same tick**, so #3384 ("scheduled storage GC has no cross-replica coordination") would only be partially closed by that PR.

Filing separately so this does not disappear when #3384 is closed.

## The tick, on `main` @ `a9f3876`

`backend/src/services/scheduler_service.rs`, one `tokio::spawn`ed cron loop, no lease anywhere in it today:

| # | line | workload | destructive |
|---|---|---|---|
| 1 | `:387` | `service.run_gc(false)` — artifact/manifest/Maven-flat sweep | yes |
| 2 | `:495` | `service.run_blob_gc_mark(dry_run)` — mark orphan blobs `pending_delete_at` | row writes |
| 3 | `:525` | `service.run_blob_gc_sweep(dry_run, grace)` — delete marked blob objects + rows | **yes, when `BLOB_GC_ENABLED=true`** |
| 4 | `:568` | `stats_service.recompute_all()` — post-GC dedup storage-stats refresh | read-heavy |

PR #3385 wraps **only row 1** in a `SchedulerLease`. Rows 2, 3 and 4 remain unconditional per-replica work.

## Measured

During 1.8.2 verification, two backend replicas against one Postgres, statement logging on, a 1-minute GC schedule:

- leased `maven_flat_object_owner` candidate scan: **1 per minute** (the lease works — that part of #3385 is real and verified)
- un-leased `oci_blobs` scans: **4-6 per minute** with only two replicas
- including at the tick where **both** replicas skipped the leased half

So on an N-replica deployment the blob-GC and stats halves are still N× duplicated, and the un-leased work is the larger share of the tick by query volume.

There is also a second-order effect worth noting: once the leased half returns immediately on the losing replica instead of grinding for ~2 minutes, the un-leased halves on all replicas fire at the *same instant* rather than being accidentally staggered by the GC they used to sit behind. The herd #3384 describes gets slightly tighter, not looser, for the un-leased portion.

## Why it matters beyond wasted work

[...]

PR #3628, opened 2026-09-01
fix(scheduler): let singleton jobs observe the lease-loss cancellation token (#3502)

### LINKED ISSUE #3502: SchedulerLease::spawn_renewal discards the #3084 lease-loss CancellationToken, so lifecycle/curation/reindex keep running after another replica takes ownership
## Summary

`cluster_work` grew a lease-loss cancellation signal in #3084 — a `CancellationToken` cancelled the moment a heartbeat reports "ownership lost" — precisely so a long-running singleton job stops its side effects when another replica has legitimately taken over. **No `SchedulerLease` consumer can observe it**, because `SchedulerLease::spawn_renewal` throws the token away.

Result: a lifecycle-policy cycle, a curation sync, or a bootstrap reindex that loses its lease mid-run keeps going to completion — concurrently with the replica that now legitimately owns the job.

## Mechanism

`backend/src/services/cluster_work.rs:222-257` builds the token and cancels it on `Ok(false)`:

```rust
Ok(false) => {
    tracing::warn!(claim = %label, "claim renewal stopped because ownership was lost");
    on_loss.cancel();
    break;
}
```

Its doc comment (`cluster_work.rs:211-221`) states the contract in as many words:

> Renewal loss means another replica may already have legitimately reclaimed the work, so a long-running worker performing external side effects (archive writes, peer HTTP, SMTP) must observe this token between chunks and abort rather than finish the side effect a second owner is about to redo (#3084).

`spawn_renewal_loop` (`:203-209`) is the thin wrapper that drops it:

```rust
pub fn spawn_renewal_loop<F, Fut>(label: impl Into<String>, ttl_secs: f64, renew: F) -> RenewalGuard {
    spawn_renewal_loop_with_cancellation(label, ttl_secs, renew).0
}
```

and `SchedulerLease::spawn_renewal` (`cluster_work.rs:312-325`) calls that wrapper, so it returns only a `RenewalGuard` — dropping which "only stops the heartbeat, it never cancels the token", per the same doc comment.

## Who is affected today

Every `SchedulerLease::spawn_renewal` caller on `main`:

| call site | job | destructive? |
|---|---|---|
| `scheduler_service.rs:255` | lifecycle policy execution (`execute_due_policies`) | **yes** — deletes artifacts, frees bytes (`:259-268`) |
| `scheduler_service.rs:758` | curation sync cycle (`run_curation_sync_cycle`) | upstream fetch + evaluate per staging repo |
| `main.rs:342` | bootstrap OpenSearch full reindex | index writes |

Each one specifically chose to heartbeat *because the work can outlive the fixed TTL* — the comments at `:252-254` and `:752-756` say so. That is exactly the population that can lose a lease mid-run, and exactly the population that cannot notice.

The one caller that does it correctly is the backup path: `spawn_backup_run_renewal` (`scheduler_service.rs:1085-1101`) uses `spawn_renewal_loop_with_cancellation` and threads the token to the worker, which is what #3084 built the mechanism for.

## Impact

- **Lifecycle retention can run twice concurrently** across replicas after a lease lapse (DB unreachable for longer than the TTL, a long GC pause, a slow cycle). Retention deletion is idempotent per row, so this is duplicated work and duplicated load rather than double deletion — but it is precisely the "two owners doing the same destructive sweep" state the lease exists to prevent, and the old owner keeps holding storage and DB pressure while the new one runs.
- The warning `claim renewal stopped because ownership was lost` is logged, so the situation is diagnosable after the fact — it just has no effect.
- This will get worse, not better: the storage-GC singleton lease proposed in #3385 uses the same `spawn_renewal` call shape, so a GC tick that loses its lease would keep deleting. Fixing the mechanism now means that PR gets the behaviour for free.

## Why it is not a two-line fix

Adding a `spawn_renewal_with_cancellation` to `SchedulerLease` is trivial. **Having something to do with the token is not** — there is no observation point in any of the three jobs. `execute_due_policies`, `run_curation_sync_cycle` and `full_reindex` have no cancellation parameter and no per-chunk checkpoint; `grep CancellationToken` across `storage_gc_service.rs` and `lifecycle_service.rs` returns nothing. A real fix threads a token down to each job's per-item loop and defines what "abort" means there (finish the current item, skip the rest, log, release), the way the backup worker does.

## Suggested direction

1. Add `SchedulerLease::spawn_renewal_with_cancellation` returning `(RenewalGuard, CancellationToken)`, keeping the existing method as a wrapper so nothing breaks.
2. Thread the token into each job's item loop and check it between items. Lifecycle first — it is the destructive one.
3. Add a test in the shape of the existing `cluster_work` claim tests: force `Ok(false)` from the renewal and assert the worker stops rather than completing.
4. Document at each remaining call site that ignores the token *why* that is acceptable, so the next reader does not have to re-derive it.

## Related

- #3084 — introduced the cancellation half of the mechanism.
- #3384 / PR #3385 — will add a fourth consumer with the same gap.

Found during 1.8.2 milestone verification; re-verified against `main` @ `a9f3876` on 2026-08-22.

brandonrc:
Closes #3502

## Summary

`cluster_work` grew a lease-loss cancellation signal in #3084 — a `CancellationToken` cancelled the moment a claim heartbeat reports "ownership lost" — precisely so a long-running singleton job stops its side effects when another replica has legitimately reclaimed the work. **No `SchedulerLease` consumer could observe it**, because `SchedulerLease::spawn_renewal` called the token-less `spawn_renewal_loop` wrapper and discarded the returned token.

The three scheduled jobs that heartbeat *because their cycles can outlive the fixed TTL* were exactly the ones that could lose a lease mid-run and had no way to notice:

| call site | job | destructive? |
|---|---|---|
| `scheduler_service.rs` `execute_due_policies` | lifecycle policy execution | **yes** — deletes artifacts, frees bytes |
| `scheduler_service.rs` `run_curation_sync_cycle` | curation upstream sync | upstream fetch + evaluate per staging repo |
| `main.rs` bootstrap reindex | OpenSearch full reindex | index writes |

Each logged `claim renewal stopped because ownership was lost` and then ran the whole cycle to completion anyway, concurrently with the replica that now legitimately owned the job. The one caller that already did it correctly is the backup path (`spawn_backup_run_renewal`), which uses `spawn_renewal_loop_with_cancellation` and threads the token to its worker — the shape this change gives the other three.

## What changed

1. `SchedulerLease::spawn_renewal_with_cancellation(db, ttl)` returns `(RenewalGuard, CancellationToken)`; the existing `spawn_renewal` becomes a thin wrapper over it, so nothing else breaks.
2. Each of the three jobs now observes the token between items:
   - **lifecycle** (`execute_due_policies`) takes an `abort: &CancellationToken` and stops **between policies** — the destructive one, done first.
   - **curation** (`run_curation_sync_cycle`) takes `abort: Option<&CancellationToken>` and stops **between staging repos**.
   - **reindex** (`full_reindex` / `full_reindex_artifacts` / `full_reindex_repositories`) takes `abort: Option<&CancellationToken>` and stops **between batches**, restoring the index refresh interval and surfacing the abort as an error rather than a clean finish.
3. Renewal *errors* still never cancel the token — the claim is held until its TTL lapses; only a real `Ok(false)` ownership loss cancels (unchanged #3084 contract).
4. The operator-invoked variants pass no token and are documented at each site as *why* that is acceptable: the manual curation trigger (`curation.rs`) and the admin/search reindex endpoints (`admin.rs`, `search.rs`) are not lease-guarded, so there is no lease to lose.

`spawn_blocking` is **not** implicated: none of the three jobs (nor the renewal loop) runs blocking work on a `spawn_blocking` handle; each is an async loop that can and now does check the token cooperatively between units of work.

DB-backed unit tests (run under `AK_TESTS_REQUIRE_DB=1`), each with a documented control:

- `cluster_work.rs::scheduler_lease_stolen_mid_work_cancels_the_loss_token` — a real heartbeat against Postgres; after another owner overwrites the `claim_token`, the next renewal reports `Ok(false)` and the token is cancelled. Negative control asserts a *held* lease that renews successfully leaves the token alive.
- `lifecycle_service.rs::test_execute_due_policies_stops_when_lease_loss_token_fired_3502` (+ `_runs_while_lease_held_3502` control) — a due policy is **not** run (no `last_run_at` stamp) once the token has fired, and **is** run while it is alive.
- `scheduler_service.rs::test_curation_sync_cycle_stops_when_lease_loss_token_fired_3502` (+ `_processes_while_lease_held_3502` control) — a staging repo is **not** stamped `curation_last_synced_at` once the token has fired, and **is** while it is alive.
- `opensearch_service.rs::test_full_reindex_threads_the_lease_loss_token_to_both_phases` / `test_reindex_page_loops_abort_on_lease_loss` — wiring pins (OpenSearch is unavailable to unit tests) asserting both page loops observe the token and abort with an error.

Each hunk was revert-proven individually: reverting the lifecycle/curation/reindex abort check, or reverting `SchedulerLease` to discard the loop's real token, makes the corresponding boundary test fail while its control stays green.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [ ] N/A - no API changes


brandonrc:
Cross-reference from #3631, which is now out of draft: leasing the whole scheduled storage-GC tick **lengthens the window this PR is about**, and it does so over destructive work.

On main, `SchedulerLease::spawn_renewal` discards #3084's cancellation token, so a singleton job that loses its lease mid-run keeps going. Today the storage-GC tick takes the lease per workload, so that window is short. #3631 holds one lease across the whole tick, including the blob sweep, so a lease lost mid-tick means a second instance can begin GC while the first is still deleting.

Nothing is wrong with #3631 — the deferral is correct and it says so — but when this lands, please make the storage-GC tick an explicitly named call site of the cancellation token rather than relying on the general mechanism, and add a test that a tick abandons its remaining workloads when the lease is lost. That is the case where "keeps going after losing the lease" stops being a tidiness issue.

brandonrc:
## Verification — merged `main`, storage-GC tick now a named call site, out of draft

Merged `origin/main` @ `eb5250a5`. **No textual conflicts**, despite #3631 rewriting 465 lines of `scheduler_service.rs` against this branch's 156. I did not take that at face value — diffed the resolution against both parents:

- `merge vs main` → exactly this branch's 9-file, +510/−32 change (identical to its pre-merge diff): every #3502 hunk survived.
- `merge vs branch` → exactly main's 15-file change: nothing of #3631/#3630/#3620 was lost.

What a clean merge *did* hide is semantic, and it is the reason for the extra commit below: main arrived carrying **two new `SchedulerLease::spawn_renewal` call sites** this branch had never seen — `storage_gc_service.rs` (the #3503 whole-tick lease) and `storage_stats_service.rs`. Both discarded the token.

**CHANGELOG:** bullet under `## [Unreleased] / ### Fixed`, confirmed after the merge (`awk '/^## \[/{h=$0} /^- \*\*/{print h}' CHANGELOG.md | sort -u`), zero conflict markers. Extended to cover the storage-GC tick. The other `#3502` occurrence in the file is #3631's own bullet referring here, also under `[Unreleased]`.

### The storage-GC tick (your note above)

Done, and it is the sharpest case in the PR. Before #3503 the tick took the lease per workload, so a lost lease left a short window; #3503 holds one lease across the whole tick **including the destructive blob sweep**, so "keeps running after losing the lease" now means "keeps deleting blobs while the new owner starts its own sweep."

- `run_scheduled_tick` heartbeats with `spawn_renewal_with_cancellation` and hands the token to its follow-on. A `_with_ttl` seam exists only so a test can drive a real heartbeat — the production TTL is an hour, which puts the first renewal 30 minutes out.
- `run_storage_gc_tick_follow_on` takes `abort` and abandons each remaining workload at its boundary: **blob-GC mark → blob-GC sweep → post-GC stats recompute**. The check lives there rather than being duplicated in `run_scheduled_tick`: the first boundary is reached before any remaining work starts, so a token already fired skips the whole follow-on anyway, and one guard means no untested branch.
- The doc block on `run_scheduled_tick` replaces #3503's "Observing lease LOSS mid-run is #3502 and deliberately not attempted here" with what it now does.

**Deliberately not extended:** the independent 4-hourly storage-stats refresh keeps token-less `spawn_renewal`, and now says why at the call site — as this PR's own new `spawn_renewal` doc contract asks. It is a single indivisible `recompute_all`: read-only, idempotent, with no per-item boundary at which an abort could take effect. Nothing destructive to abandon and nowhere to check.

### Merits

The three original call sites are right: `spawn_renewal` becomes a thin wrapper so nothing else breaks; renewal *errors* still never cancel (the claim is held until TTL lapses — #3084's contract unchanged); every operator-invoked variant passes `None` and documents at the site that it is not lease-guarded. This fixes the failure mode rather than surfacing it — the warning `claim renewal stopped because ownership was lost` was already logged and already had no effect.

### Tests — fail-without-fix, proven per hunk

`AK_TESTS_REQUIRE_DB=1`, throwaway Postgres, `cargo nextest`.

New:
- `scheduled_gc_tick_hands_its_follow_on_the_real_lease_loss_token_3502` — real Postgres heartbeat at the renewal loop's 5s floor. Negative control first (lease held → token alive), then another replica overwrites the `claim_token` and the token the tick handed its follow-on fires.
- `test_storage_gc_tick_follow_on_abandons_remaining_workloads_on_lease_loss_3502` — token fired → `instance_storage_stats.computed_at` does not advance, i.e. the tick never reached its last workload. Its control is the existing #3503 test: the identical call with `abort: None`, asserting `computed_at` **does** advance. The pair cannot both pass unless the token is what decides.

Reverts:

| reverted hunk | fails | control |
|---|---|---|
| the three phase guards in `run_storage_gc_tick_follow_on` | `..._abandons_remaining_workloads_on_lease_loss_3502` | `..._recomputes_stats_and_keeps_blob_gc_dry_run_db` green |
| `run_scheduled_tick` back to token-discarding `spawn_renewal` | `scheduled_gc_tick_hands_its_follow_on_the_real_lease_loss_token_3502` | — |
| lifecycle `abort.is_cancelled()` check disabled | `test_execute_due_policies_stops_when_lease_loss_token_fired_3502` | `..._runs_while_lease_held_3502` green |

Exactly one targeted failure per revert.

`cargo fmt` clean · `cargo clippy --lib --tests --all-features -D warnings` clean · `cargo check --lib --tests` clean · `cargo nextest --lib` over `cluster_work`/`scheduler_service`/`lifecycle_service`/`storage_gc_service`/`storage_stats_service`/`opensearch_service`: **415 passed, 0 failed**.

PR #3630, opened 2026-09-01
refactor(plugins): remove the dead classic plugin hook dispatch from artifact operations

### LINKED ISSUE #3499: ArtifactService plugin hooks are dead code: plugin_service is always None in production, so BeforeUpload/BeforeDownload/BeforeDelete vetoes never fire
## Summary

`ArtifactService`'s plugin hook system is never wired up in production. `plugin_service` is `None` on every `ArtifactService` the API constructs, so all six lifecycle hooks are unconditionally no-ops — including the two **blocking** hooks whose documented purpose is to let a plugin reject an operation.

The documented contract at `backend/src/services/artifact_service.rs:1974` — *"Trigger BeforeDelete hooks - validators can reject the deletion"* — does nothing on any delete path.

## Mechanism

`backend/src/api/mod.rs:405-415`:

```rust
pub fn create_artifact_service(&self, storage: Arc<dyn StorageBackend>) -> ArtifactService {
    let mut svc = ArtifactService::new_with_search(self.db.clone(), storage, self.search_service.clone());
    if let Some(ref scanner) = self.scanner_service { svc.set_scanner_service(scanner.clone()); }
    if let Some(ref qc) = self.quality_check_service { svc.set_quality_check_service(qc.clone()); }
    svc
}
```

`new_with_search` hard-codes `plugin_service: None` (`artifact_service.rs:252`), as does `new` (`:234`). The two constructors that would populate it —

- `ArtifactService::with_plugins` (`artifact_service.rs:260-271`)
- `ArtifactService::set_plugin_service` (`artifact_service.rs:278-280`)

— have **zero callers anywhere in `backend/src`**. The other five production construction sites (`repositories.rs:4917, 7288, 7578, 8535, 8630`) all use `ArtifactService::new`, which also leaves it `None`.

`PluginService::new` itself is called exactly once in the whole tree, in a test (`backend/src/services/plugin_service.rs:1901`). Nothing in `main.rs` or `api/mod.rs` constructs one. (`WasmPluginService` — the thing `main.rs:1376` builds and `SharedState.wasm_plugin_service` holds — is a **different** subsystem and does not feed these hooks.)

Both dispatchers therefore short-circuit:

```rust
async fn trigger_hook(&self, event, info) -> Result<()> {
    if let Some(ref plugin_service) = self.plugin_service { ... } else { Ok(()) }   // :298-308
}
async fn trigger_hook_non_blocking(&self, event, info) { if let Some(...) { ... } } // :312-322
```

## Scope — all six hooks, not just delete

| call site | hook | kind |
|---|---|---|
| `artifact_service.rs:646` | `BeforeUpload` | **blocking / veto** |
| `artifact_service.rs:965` | `AfterUpload` | non-blocking |
| `artifact_service.rs:1459` | `BeforeDownload` | **blocking / veto** |
| `artifact_service.rs:1522` | `AfterDownload` | non-blocking |
| `artifact_service.rs:1975` | `BeforeDelete` | **blocking / veto** |
| `artifact_service.rs:2051` | `AfterDelete` | non-blocking |

`plugin_service.rs:1806` asserts `is_blocking_event(PluginEventType::BeforeDelete)` — the plugin service knows these are veto points; nothing ever asks it.

## Impact

- Any plugin registered as a `before_upload` / `before_download` / `before_delete` validator has **no effect**. If a deployment relies on one for policy enforcement, that policy is not enforced and there is no signal — the hook returns `Ok(())` and the operation proceeds.
- Design work that reasons about the veto is reasoning about a control that does not exist. This came up during 1.8.2 review of the OCI delete path, where an "a plugin could veto the delete here" concern turned out to be unreachable.
- The `plugin_events` / hook registration surface presents a capability the runtime does not honour.

## Suggested direction

This is a decision, not a patch:

1. **Wire it up** — construct a `PluginService` in `main.rs`, hold it on `SharedState`, and set it in `create_artifact_service` alongside scanner and quality-check. This turns six no-ops into live hooks, which is a **behaviour change** for any deployment that has plugin rows registered: previously-ignored validators would start rejecting uploads/downloads/deletes. It needs a changelog callout and probably an opt-in flag for one release. Note the blocking hooks run inside request paths, so plugin latency becomes request latency — check that before enabling.
2. **Or remove it** — delete `PluginService`'s hook dispatch, `with_plugins`, `set_plugin_service`, the six call sites and the "validators can reject" comments, and let `WasmPluginService` be the only extension story. Honest, and stops the next reviewer trusting a control that isn't there.

Whichever way it goes, the outcome should be documented at the `trigger_hook` definitions so the next reader does not have to trace the constructors to find out.

If (1), it needs a test that a registered blocking hook actually aborts a delete — the current suite cannot distinguish "hook vetoed nothing" from "hook was never called".

Structural + behaviour-affecting, hence 1.9.0.

Found during 1.8.2 milestone verification; re-verified against `main` @ `a9f3876` on 2026-08-22.

brandonrc:
## Summary

Removes the classic `PluginService` hook dispatch from artifact operations — dead code that presented a control which could never fire (#3499).

**What was advertised vs what ran.** `ArtifactService` carried six lifecycle hooks (`BeforeUpload`/`AfterUpload`/`BeforeDownload`/`AfterDownload`/`BeforeDelete`/`AfterDelete`), three of them documented as blocking vetoes ("validators can reject the deletion"). The plugin catalog accepts `webhook`/`custom` plugin installs, the admin config endpoint stores `webhook_url`/`validator_url` callbacks (and SSRF-validates them at save time), and `plugin_hooks` has blocking-event semantics. But `plugin_service` was `None` on every `ArtifactService` the API constructs: `with_plugins`/`set_plugin_service` had zero callers, `PluginService::new` was constructed only by a test, and **no API path writes `plugin_hooks` rows at all**. Every hook was unconditionally a no-op.

**Verified live** (isolated instance, image built from `main`): an active `custom` plugin with an always-veto validator, all six `plugin_hooks` rows armed, and a listener on the callback URL — upload 201, download 200, delete 200, artifact really soft-deleted, **0 callbacks received, 0 `plugin_events` rows**. As a counterfactual, a locally built naive wire-up of the same code blocks uploads/deletes from those same DB rows (the `before_upload` veto returns 400) — confirming the dispatch plumbing was the only missing link and that these rows are live policy the moment anyone wires it, which is exactly why leaving it half-dead is hazardous.

**Decision: remove, not wire.** Wiring it up would (a) activate vetoes from DB rows nothing can register through the API, (b) put arbitrary plugin HTTP latency inside upload/download/delete request paths, and (c) duplicate the extension stories that actually run: the WASM plugin service (format handlers) and the webhook subsystem (`webhook_producer`/`event_bus`). Reintroducing hooks should be a designed feature (registration API, latency budget, opt-in), not a wire-up of this remnant.

Changes:
- delete `services/plugin_service.rs` (hook dispatcher + a lifecycle half duplicated by the plugins handlers; zero production constructors)
- `ArtifactService`: drop the `plugin_service` field, dead constructors/setters, both `trigger_hook*` dispatchers and the six call sites; `preflight_upload` sheds its three hook-only parameters
- `ArtifactInfo` moves into `artifact_service.rs` (it remains the value struct for the download/audit epilogues)
- plugins handler docs updated: the classic dispatcher is gone; stored callback-URL SSRF validation is kept as defense in depth
- decision documented at `prepare_delete` and `preflight_upload` so the next reader does not re-derive it

No deployment-observable behavior changes: the hooks never fired, and the plugin catalog/config/events API surface is untouched.

`test_3499_classic_plugin_hook_rows_do_not_gate_delete` (DB-backed): seeds an active `custom` plugin with an unreachable always-veto validator plus an enabled `before_delete` hook row, then asserts `delete_with_sync_options` succeeds and the row is really soft-deleted. It pins the decided contract — classic plugin rows are inert catalog data — and goes red against any future naive wire-up (verified: it is the same probe that fails against the wired counterfactual image). A deployment-level DTF tier (`classic-plugin-hooks-inert`, validated red against the wired counterfactual and green on this PR) accompanies this in artifact-keeper-test.

## Test Checklist
- [x] E2E tests added/updated (if applicable) — DTF tier `classic-plugin-hooks-inert` (batched separately)
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

## Notes for reviewers

- The `plugins`/`plugin_hooks`/`plugin_events` tables are untouched (migrations are immutable; `plugin_events` has a live WASM-service writer and reader).
- If a deployment has hand-inserted `plugin_hooks` rows expecting enforcement: that enforcement never existed on any released version. The changelog entry should say so explicitly.

Closes #3499


brandonrc:
## Deadness audit — independent verification before taking this out of draft

I did not take the PR description's word for it. Every deleted entry point was re-derived from the tree at the merge base, across the whole repository (source, tests, `backend/migrations/`, `.sqlx/`, `docs/`, `.github/`, `docker/`, `scripts/`, compose files) and across the sibling repositories. Merged `origin/main` first; the branch predated the 1.9.0 cut, so the CHANGELOG bullet lands under the current `## [Unreleased]` (line 18), not inside the released `## [1.9.0]` (line 40) — the mistake #3795 had to fix on main.

### What is deleted

| | |
|---|---|
| `backend/src/services/plugin_service.rs` | whole file (1952 lines): `PluginService` (`new`, `with_client`, `install_plugin`, `enable_plugin`, `disable_plugin`, `uninstall_plugin`, `get_plugin`, `load_plugins`, `register_hook`, `trigger_hooks`, `execute_webhook_hook`, `execute_validator_hook`, `call_validator_url`, `log_plugin_event`, `resolve_webhook_url`, `resolve_validator_url`, `evaluate_config_rules`, `run_config_rules`, `is_blocking_event`), plus `PluginEventType`, `WebhookPayload`, `ValidatorResult`, `CachedPlugin` and their tests |
| `backend/src/services/mod.rs` | `pub mod plugin_service;` |
| `ArtifactService` | `plugin_service` field, `with_plugins`, `set_plugin_service`, `trigger_hook`, `trigger_hook_non_blocking`, the six call sites; `preflight_upload` sheds `name` / `content_type` / `uploaded_by`, which were hook-only |
| moved, not deleted | `ArtifactInfo` → `artifact_service.rs` (still the value struct for the download/audit epilogues) |

### Evidence, per entry point

**Symbol reachability.** `with_plugins`, `set_plugin_service`, `trigger_hook`, `trigger_hook_non_blocking`, `register_hook`, `load_plugins`, `WebhookPayload`, `ValidatorResult` appear in **no file other than the two being edited**. `PluginService` appears outside `plugin_service.rs` only in `artifact_service.rs` (the dead field) and in two doc comments in `plugins.rs`. `PluginEventType` appears only at the six dead call sites. `install_plugin` in `plugins.rs` is that handler's own function, not the service method — the handler resolves `WasmPluginService` from `SharedState`.

**Never wired, in the whole history, not just on `main`.** `git log -S "set_plugin_service("` and `-S "with_plugins("` over `origin/main -- backend/src` each return exactly one commit: `ad1c1eab`, the one that *defined* them. No commit ever added a call site, so the veto never fired on any released version.

**Classic path vs WASM path are disjoint.** `wasm_plugin_service.rs` and `wasm_runtime.rs` contain zero references to `PluginService`, `trigger_hook`, `PluginEventType` or `ArtifactInfo`. The `format-plugin-v2` interface lives in `wasm_bindings.rs` / `wit/format-plugin.wit` and touches none of it. `main.rs` constructs only `WasmPluginService` (`initialize_wasm_plugins`); nothing anywhere constructs a `PluginService`.

**String / dynamic dispatch (a symbol grep would miss this).** The two string columns that could key a registry — `plugin_hooks.hook_type` and `plugin_hooks.handler_name` — are read *nowhere* outside the deleted file; the only other occurrence of either is the field declaration in `models/plugin.rs`. `PluginEventType::from_str` has no callers at all. Grepping the six event names as raw strings (`before_upload`, `after_upload`, `before_download`, `after_download`, `before_delete`, `after_delete`) across the whole tree returns only the SSRF tests in `plugins.rs` and unrelated identifiers (`pause_after_downloads`, a test name in `repositories.rs`, a shell variable in artifact-keeper-test). There is no `HashMap<String, dyn …>` handler registry on this path.

**Can a DB row still select it?** No. The sole `INSERT INTO plugin_hooks` in the tree is `PluginService::register_hook`, which has zero callers — **no route creates a hook row**; `admin_router()` exposes install / uninstall / enable / disable / config / WASM install / reload and nothing else. `wasm_plugin_service.rs:1807` only `DELETE`s from the table. No migration seeds `plugins` or `plugin_hooks`. And `main.rs::load_active_plugins` filters `status = 'active' AND wasm_path IS NOT NULL`, so a classic `webhook`/`custom` row (no `wasm_path`) is never even loaded at startup.

**Can a config key or env var select it?** No. Every plugin-related setting is WASM-only: `PLUGINS_DIR` (WASM plugin directory), `PLUGINS_REQUIRE_SIGNED` / `PLUGINS_TRUSTED_PUBKEY` (WASM install signature policy). No env var or config key touches the classic dispatcher.

**Sibling repositories.** Shallow clones of `artifact-keeper-web`, `artifact-keeper-test` and `artifact-keeper-iac`, grepped for `plugin_hooks`, `PluginService`, `validator_url` and the six event names: **zero hits** in web and iac; the single hit in artifact-keeper-test is a local shell variable `dl_before_delete.bin` in a HuggingFace test. Nothing out there configures or exercises this path.

**Nothing is still reachable.** No path was found that could reach the deleted dispatch, so nothing was forced.

### Public API surface

Generated the spec on both sides (`EXPORT_OPENAPI_SPEC=1 … export_openapi_spec`) and diffed: **byte-identical**, `sha256 898ce534…8606` on both, 340 paths / 477 schemas. `backend/src/api/openapi.rs` is untouched by the diff; the deleted file carried no `utoipa` / `ToSchema` / `IntoParams` annotation at all, and the `plugins.rs` edits are doc comments on `router()` / `admin_router()` / a private helper, none of which are `#[utoipa::path]` handlers. Stored `webhook_url` / `validator_url` SSRF validation is kept as defense in depth.

### Local validation

`cargo fmt --all` (no changes) · `cargo clippy --workspace --all-targets --all-features -- -D warnings` → **0 warnings** · `cargo check --workspace --lib --tests` → clean · `cargo nextest run --workspace --lib` over the plugin / format / artifact / openapi areas → **2605 passed, 0 failed**, against a real Postgres 18 with `AK_TESTS_REQUIRE_DB=1` (so nothing silently skipped). `test_3499_classic_plugin_hook_rows_do_not_gate_delete` was also run in isolation against that database and passes.

PR #3631, opened 2026-09-01
fix(gc): lease the whole scheduled storage-GC tick, and the independent stats refresh

### LINKED ISSUE #3503: Scheduled GC tick: blob-GC mark/sweep and the storage-stats recompute stay un-leased on every replica, so #3385's singleton lease covers only one of four workloads

brandonrc:
## Summary

Extends the scheduled storage-GC singleton lease over the **whole** tick, and gives the independent storage-stats refresh its own lease (#3503).

**What triggers what today.** The scheduled tick (`GC_SCHEDULE`, default hourly) runs four workloads back to back: `run_gc`, blob-GC mark, blob-GC sweep, and the post-GC storage-stats recompute. #3385's singleton lease covered only `run_gc`. The other three ran on **every replica at every occurrence**, un-leased — and because the losing replica returned from the leased half immediately, it fired the un-leased half in perfect sync with the winner. The storage-stats recompute additionally runs on its own cron (`STORAGE_STATS_SCHEDULE`, default 4-hourly), also un-leased.

**Verified live** (isolated instance, image built from `main`, `GC_SCHEDULE=*/20s`): with a foreign `storage_gc` lease planted, `run_gc` correctly skipped (log line frozen) while `instance_storage_stats.computed_at` kept advancing every occurrence and Postgres statement logs showed 27 `oci_blobs` blob-GC scans per 100 s — three quarters of the tick running without the lease. On this PR's build, the same foreign lease freezes the entire tick (computed_at frozen across 5+ occurrences, zero blob-GC scans), the on-demand admin GC still works while the lease is held, and freeing the lease resumes the tick at the next occurrence.

Changes:
- `StorageGcService::run_scheduled_tick` now takes the remainder of the tick as a `follow_on` closure executed under the held lease (after `run_gc`, before release). A replica either owns an occurrence outright or skips it entirely; there is no window where it can lose the job between workloads and silently continue.
- The scheduler moves the blob-GC gate/mark/sweep and the post-GC recompute into that closure — code unchanged, only ownership changes.
- The independent stats refresh gets its own singleton lease (`scheduler_leases.job_name = 'storage_stats_refresh'`, operator-visible like `storage_gc`).
- The follow-on body is extracted as `run_storage_gc_tick_follow_on` so the leased tick body is directly testable (and the moved code is exercised by a test rather than living untestable inside the spawned loop).
- Drive-by within the touched file: the pending/unreferenced cleanup-key **claim** queries were byte-identical except for their family predicates; they are collapsed onto one parameterized body (`OciCleanupKeyFamily`) so the two destructive claim paths cannot drift apart. (This also keeps the changed-file set under the post-#3523 duplication gate, which now measures this file's pre-existing clones on any PR that touches it: 3.45% before, 2.77% after.)

**What the fix does *not* change:** the blob-GC dry-run default (`BLOB_GC_ENABLED`, off ⇒ report-only), the `manifest_blob_refs` readiness gate, and the `MAVEN_FLAT_GC_ENABLED` opt-in all stay exactly where they were — the lease decides *who* runs, never *whether deletion is enabled*. Observing lease **loss** mid-run remains #3502 and is deliberately not attempted here.

## First tick after upgrade (changelog / upgrade note)

- **Single replica:** identical behavior. The lease is uncontended; every workload runs as before.
- **Multi replica:** blob-GC and the stats recompute now run **once** per occurrence instead of N times. Strictly less work, no new deletions; if the tick owner crashes mid-tick, that occurrence's remaining workloads are skipped cluster-wide and the next occurrence covers them (previously other replicas redundantly re-ran them).
- **Worth naming for operators:** under the default `BLOB_GC_ENABLED` unset, unreferenced blobs are **never** collected — the sweep is dry-run forever and storage grows unbounded; only the "would sweep N objects, M bytes" log lines tell you what has accumulated. On a long-lived instance, the **first live sweep after enabling `BLOB_GC_ENABLED=true`** can therefore delete a large accumulated backlog at once (after the grace window and the readiness gate). The dry-run-first default is the responsible path: read the dry-run figures before opting in. This PR does not change any of those defaults.
- Interaction: the **on-demand** GC endpoint's `dry_run` default is being decided separately in #3501 by parallel work; this PR deliberately does not touch `StorageGcRequest` and always passes `dry_run` explicitly in its tests/oracles.

- `run_scheduled_tick_respects_the_singleton_lease` (extended): with a foreign lease, the tick must skip AND the follow-on (blob-GC + stats in production) must not run; once freed, the tick must run the follow-on under the lease and release. The "must not run" half fails on `main`'s shape by construction (the follow-on used to run unconditionally).
- `scheduled_refresh_respects_the_singleton_lease_db` (new): a losing replica must not advance `instance_storage_stats.computed_at`; a winning replica must; the lease must be released afterwards.
- `storage_gc_tick_follow_on_recomputes_stats_and_keeps_blob_gc_dry_run_db` (new): a follow-on call must observably do the work (`computed_at` stamped fresh) and the shipped `BLOB_GC_ENABLED=false` default must keep both blob-GC phases dry-run (no `pending_delete_at` markers appear).
- Deployment-level: DTF tier `storage-gc-tick-lease` (artifact-keeper-test, batched separately) — validated RED on `main` ("computed_at advanced while another replica held the storage_gc lease") and GREEN on this PR, with liveness and on-demand-GC controls so a dormant scheduler cannot pass.

## Test Checklist
- [x] E2E tests added/updated (if applicable) — DTF tier `storage-gc-tick-lease` (batched separately)
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes (one new `scheduler_leases.job_name` wire value: `storage_stats_refresh`)

Closes #3503


brandonrc:
## Merged `main` into this branch: one conflict, in `storage_gc_service.rs`

Both sides touched `claim_pending_oci_upload_cleanup_keys`, at the same lines, for unrelated reasons:

| side | what it had |
|---|---|
| this branch (`f13415f4`..`080e73ac`) | the inline pending-family claim SQL **deleted**, replaced by a call to the shared `claim_oci_upload_cleanup_keys(repo_scope, OciCleanupKeyFamily::Pending)` body |
| `main` (the sqlx 0.9 `AssertSqlSafe` migration) | that inline copy still present, with one line inside it changed: `sqlx::query(&sql)` → `sqlx::query(sqlx::AssertSqlSafe(&*sql))` |

**Kept:** the shared-helper call. `main`'s change is *not* dropped — `AssertSqlSafe` had already applied cleanly to the one surviving copy of that statement (the shared `claim_oci_upload_cleanup_keys` body, which this branch extracted from the unreferenced-family claim), so the resolution keeps main's structure with the dedup re-applied on top rather than the reverse.

Verified by diffing the resolution against **both** parents rather than eyeballing it:

- vs. this branch's head: the only delta is `main`'s mechanical `sqlx::query(&sql)` → `sqlx::query(sqlx::AssertSqlSafe(&*sql))` rewrite, 17 call sites, no other change;
- vs. `main`: the only delta is this PR's own commits. The `-` lines inside the claim SQL are the deleted duplicate copy only.

All three candidate guards survive for **both** families in the surviving body — `NOT EXISTS (… oci_upload_sessions …)`, `NOT EXISTS (… oci_upload_parts p WHERE p.storage_key = c.storage_key)`, `NOT EXISTS (… oci_blobs b WHERE b.storage_key = c.storage_key)` — along with the `(claim_expires_at IS NULL OR claim_expires_at <= NOW())` live-claim exclusion, the repo-scope clause, `ORDER BY c.created_at ASC`, `LIMIT $1`, `FOR UPDATE OF c SKIP LOCKED`, the claim stamp and the `RETURNING` shape. The per-family variance is exactly the two predicates the enum carries: the age/eligibility test, and the session guard (pending additionally anchors `s.id = c.upload_session_id`; unreferenced matches on `s.storage_temp_key` only). Nothing else differed between the two former queries, and nothing else differs now.

## CHANGELOG

This branch predates the 1.9.0 cut and carried **no** CHANGELOG entry at all, so there was nothing for the merge to misfile — but the same trap was live: a bullet written under the then-current `## [Unreleased]` heading lands inside `## [1.9.0]` after the cut, because the heading it was written under still exists (#3797). Added under the **current** `## [Unreleased]` / `### Fixed`, and checked, not assumed:

```console
$ awk '/^## \[/{h=$0} /^- \*\*/{print h}' CHANGELOG.md | sort -u   # → [Unreleased]
```

(The `[1.8.2]` section's similar-sounding bullet is #3385's, correctly filed there.)

## Review on the merits

**Does leasing the whole tick close #3503's race, or move it?** It closes it for the case the issue actually measured. The four workloads are one cron occurrence; a per-occurrence singleton lease means the occurrence has exactly one owner, so the blob-GC mark/sweep and the stats recompute stop being N× duplicated, and — the part a set of four separate leases would *not* have given — there is no longer a window in which a replica wins `run_gc`, loses the job between workloads, and silently carries on into the destructive half. The residual N× herd the issue describes ("the un-leased halves fire at the same instant rather than staggered") is removed rather than relocated, because the losing replica now returns before reaching any of it.

**What if the lease is lost mid-tick?** The work is *not* abandoned, and it does **not** observe a cancellation token. `SchedulerLease::spawn_renewal` on `main` calls the token-less `spawn_renewal_loop` and discards the `CancellationToken` that #3084 added, so no `SchedulerLease` consumer can observe loss today — that is #3502, and #3628 (still open, not merged) is the change that makes the token reachable. So this PR's deferral is correct as written: there is nothing to thread yet. It is worth being explicit that this PR **lengthens the exposure**, though — the lease now spans `run_gc` + blob mark + blob sweep instead of `run_gc` alone, so the window in which a heartbeat can report `Ok(false)` while destructive work continues is longer than before. That is not a regression (on `main` the follow-on ran on every replica unconditionally, which is strictly worse), but when #3628 lands, `run_scheduled_tick` should take the `(RenewalGuard, CancellationToken)` variant and check the token between `run_gc` and the follow-on and between the mark and sweep phases — the same "between units of work" shape #3628 gives lifecycle and curation. Suggest adding a note to #3502 so this call site is not missed.

**Is the "independent stats" half really independent?** Yes, and a *separate* lease is the right call rather than the same one. Sharing `storage_gc` would couple two unrelated cadences: a 4-hourly refresh holding the GC lease would suppress an hourly GC occurrence, and a long GC tick would silently eat a refresh occurrence. With distinct job names the two can still overlap across replicas — but `recompute_all` takes `pg_advisory_xact_lock(hashtext('repository_path_storage_stats_rebuild'))`, so concurrent rebuilds serialize and converge; the cost is duplicated work in a rare overlap, not a wrong result. The `storage_stats_refresh` job name is operator-visible in `scheduler_leases`, matching `storage_gc`.

**One test-quality fix, pushed as `9d8a3016`.** `storage_gc_tick_follow_on_recomputes_stats_and_keeps_blob_gc_dry_run_db` asserted only that `computed_at` was younger than 60 seconds. In the fail-without-fix run below it **passed** with the follow-on stubbed to a no-op, because a sibling DB test had recomputed inside that window — it was measuring the wall clock, not the call. It now takes its own baseline recompute and asserts `computed_at` advances past that stamp, the same discriminator `scheduled_refresh_respects_the_singleton_lease_db` already uses.

## Verification

Fail-without-fix, by reverting the production hunks only (tests untouched) and re-running:

| test | with production hunks reverted |
|---|---|
| `storage_gc_service::tests::run_scheduled_tick_respects_the_singleton_lease` | **FAIL** — `the follow-on half of the tick (blob GC + stats recompute) must NOT run on a replica that failed to win the lease (#3503)` |
| `storage_stats_service::db_tests::scheduled_refresh_respects_the_singleton_lease_db` | **FAIL** — `refresh must skip while another replica holds the lease` |
| `scheduler_service::tests::storage_gc_tick_follow_on_recomputes_stats_and_keeps_blob_gc_dry_run_db` | **FAIL** — `the follow-on must have recomputed storage stats` (this is the one that was green before `9d8a3016`) |

The reverts used were: the follow-on run unconditionally outside the lease (`main`'s shape); `run_scheduled_refresh` reduced to a bare un-leased `recompute_all`; and `run_storage_gc_tick_follow_on` stubbed to a no-op. All three restored, all three green.

Commands (Postgres 18 on a throwaway container, migrations applied, `AK_TESTS_REQUIRE_DB=1` so a DB-gated test cannot silently no-op):

```console
$ cargo fmt --all                                            # clean
$ cargo check --lib --tests                                  # clean
$ cargo clippy --lib --tests --all-features -- -D warnings    # clean
$ cargo nextest run --lib -E '<the three tests above>'        # 3 passed
$ cargo nextest run --lib -E 'test(storage_gc_service::) or test(storage_stats_service::)
                              or test(scheduler_service::) or test(cluster_work::)'
                                                              # 205 passed, 0 failed
```

Not merging — leaving that to the normal gate.

PR #3635, opened 2026-09-01
fix(users): expose and allow filtering the service-account discriminator (#3634)

### LINKED ISSUE #3634: GET /api/v1/users returns service accounts with no way to identify or exclude them, so permission grants for them cannot be built from that list
### Description

`GET /api/v1/users` returns service accounts mixed in with real users, while exposing **no field to tell them apart and no query parameter to exclude them**. At the same time, `POST /api/v1/permissions` hard-rejects a service account submitted as `principal_type: "user"`.

The two endpoints therefore contradict each other: the API publishes a set of principals through one endpoint and refuses a subset of them at another, giving clients no way to reconcile the two.

`list_users` reads the discriminator out of the database and then discards it:

- `backend/src/api/handlers/users.rs:257` — the `SELECT` includes `is_service_account`
- `backend/src/api/handlers/users.rs:182` — `AdminUserResponse` has no `is_service_account` field, so it never reaches the client
- `backend/src/api/handlers/users.rs:134` — `ListUsersQuery` filters only on `search`, `is_active`, `is_admin`

Meanwhile the grant path requires the opposite:

```rust
// backend/src/services/permission_service.rs:96
"user"            => "SELECT EXISTS(SELECT 1 FROM users WHERE id = $1 AND is_service_account = false)"
"service_account" => "SELECT EXISTS(SELECT 1 FROM users WHERE id = $1 AND is_service_account = true)"
```

`validate_principal` (`permission_service.rs:374`) then returns `400 principal_id <uuid> does not exist as a user`. That guard is correct and was added deliberately (#2503, defense against misattributed grants) — the defect is that `GET /api/v1/users` was never updated to match it.

### Environment Information

- Deployment type: Docker
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL
- Reverse Proxy: yes
- Network setup: internal, behind reverse proxy

### Backend Version

v1.8.1 (reproduced against `v1.8.1-34-gbba534f7`)

### Web UI Version

v1.8.0

### Expected Behavior

A client listing users can either exclude service accounts or render them distinguishably, so it never offers the operator a principal that the grant endpoint is guaranteed to reject.

### Observed Behavior

Service accounts appear as ordinary users in `GET /api/v1/users`. Selecting one as a permission principal always fails with `400 principal_id <uuid> does not exist as a user`, and no amount of client-side logic can prevent it, because the response carries nothing that identifies the row as a service account.

### Steps to Reproduce

1. Create a service account: `POST /api/v1/service-accounts` with `{"name": "raw-ci-write"}`.
2. List users as an admin:

   ```
   curl -s -H "Authorization: Bearer $ADMIN_TOKEN" https://artifacts.example.com/api/v1/users | jq '.items[] | select(.username | startswith("svc-"))'
   ```

   The `svc-*` account is present, and the object contains no `is_service_account` field.
3. Grant it a repository permission using the principal type the listing implies:

   ```
   curl -s -X POST -H "Authorization: Bearer $ADMIN_TOKEN" -H "Content-Type: application/json" \
     https://artifacts.example.com/api/v1/permissions -d '{
       "principal_type": "user",
       "principal_id":   "<service-account-uuid>",
       "target_type":    "repository",
       "target_id":      "<repository-uuid>",
       "actions":        ["read","write"]
     }'
   ```
4. Observe `400` — `principal_id <uuid> does not exist as a user`.
5. Confirm there is no way to filter server-side: `GET /api/v1/users?is_service_account=false` is accepted but silently ignored (`ListUsersQuery` has no such field).

### Additional Context

**Blast radius.** Every consumer of this endpoint inherits the problem. In `artifact-keeper-web`, five admin screens call it: the permissions principal picker, the users list, the group member picker, the audit actor filter, and the age-gate approver picker. Four of them never fetch `/api/v1/service-accounts`, so they have no client-side means to distinguish a person from a robot at all. A companion frontend issue is filed at artifact-keeper-web for the one screen that *does* have the data locally and still doesn't use it.

**How it surfaced.** A `svc-*` account with a token carrying `read:artifacts` / `write:artifacts` / `delete:artifacts` / `admin` got `403 You do not have permission to perform this action on this repository` on upload — correct, since token scopes only narrow and confer nothing (`handlers/service_accounts.rs:527` writes scopes and repo restrictions only; `services/service_account_service.rs:53` creates the account with no role and no grant). The obvious remedy — grant the account a repository permission — then failed at the picker, because the account was listed as a user.

**Suggested fix.** A blanket server-side exclusion would be wrong: filtering the audit actor list by a service account is legitimate, and group membership for one is arguably valid too. Let the caller decide instead:

1. Add `is_service_account: bool` to `AdminUserResponse` (`users.rs:182`).
2. Add `is_service_account: Option<bool>` to `ListUsersQuery` (`users.rs:134`), applied with the same NULL-passthrough pattern the existing `is_active` / `is_admin` filters already use — a two-line change to both the page query and the count query, which must stay in sync.

Clients can then request `?is_service_account=false` for grant pickers and badge the rows elsewhere.

**Related.** #3522 covers the adjacent `user_roles` vs `role_assignments` confusion in the same area; #2503 introduced the `validate_principal` guard this contradicts.

sblaisot:
## Summary

Closes #3634.

`GET /api/v1/users` returned service accounts mixed in with people while exposing **nothing to identify or exclude them** — and `POST /api/v1/permissions` rejects exactly those rows when submitted as `principal_type: "user"` (`400 principal_id <uuid> does not exist as a user`, the `validate_principal` guard from #2503). The two endpoints contradicted each other, and clients had no way to reconcile them: `list_users` read `is_service_account` from the database and dropped it in `AdminUserResponse`, and `ListUsersQuery` had no filter.

The practical consequence: a `svc-*` account whose token gets `403 You do not have permission to perform this action on this repository` needs a repository grant (token scopes only narrow — `create_token` writes scopes and repo restrictions, `ServiceAccountService::create` writes no role and no grant). That grant could not be driven from the user list at all, because the list published the principal and the grant endpoint refused it.

This PR:

- adds `is_service_account` to `AdminUserResponse` (the column was already SELECTed — it was dropped on the way out);
- adds an optional `is_service_account` filter to `ListUsersQuery`, applied with the same NULL-passthrough shape as the existing `is_active` / `is_admin` filters.

**Deliberately opt-in, not a blanket exclusion.** Filtering the audit actor list by a service account is legitimate — CI is often the actor under investigation — and so is group membership for one. An absent filter keeps the historical unfiltered listing byte-for-byte; only a caller that asks gets people-only.

**Applied to both statements.** `list_users` runs a page query *and* a pagination count. Filtering one and not the other makes `total` count rows `items` excludes: `total_pages` overshoots and the listing grows a phantom trailing page. The two `WHERE` clauses are held byte-identical and a test compares them. Worth flagging for review: the three filters are all `Option<bool>`, so they are type-identical at the bind site — a transposed bind compiles clean and silently filters the wrong column.

**Offline query cache.** The two `sqlx` macros changed, so their `.sqlx/` entries are regenerated (two files replaced). No database was reachable in this environment, so the entries were derived from the previous ones — the query text is taken verbatim from the source literal and the parameter list gains the one `Bool` — rather than by `cargo sqlx prepare`. `cargo check` and `cargo test --lib` pass against them under `SQLX_OFFLINE=true`, which is how CI builds. **Please have a reviewer with a database confirm a `cargo sqlx prepare` run produces no diff.**

Frontend counterpart: artifact-keeper-web#824 (for artifact-keeper-web#823) fixes the permissions principal picker page-locally. The other four screens reading this endpoint (users list, group member picker, audit actor filter, age-gate approver picker) wait on an SDK release carrying this field.

Three unit tests in `backend/src/api/handlers/users.rs`:

- `user_to_response_exposes_the_service_account_discriminator` — on `main` this does not compile, because the field it asserts on does not exist.
- `list_users_page_and_count_queries_filter_identically` — reads the two `WHERE` clauses out of the source and asserts they match. With no database in a unit test this is the only way to pin the count-drift class of bug at all; it also asserts the filter is actually *applied*, not merely accepted by the query struct.
- `list_users_query_service_account_filter_is_optional` — pins that an absent filter means unfiltered, so the audit and group pickers keep their current results. Also does not compile on `main`.

## Test Checklist
- [ ] Manually tested locally

`cargo fmt --all -- --check`, `cargo clippy -p artifact-keeper-backend --lib --tests -- -D warnings` and `cargo test -p artifact-keeper-backend --lib` (105 tests in the users module, 0 failed) all pass locally under `SQLX_OFFLINE=true`. Not run, and not runnable here — no PostgreSQL or Docker in this environment — the Tier 2 integration suite, and no manual exercise of the endpoint against a live server. CI covers it on this PR (it touches `backend/**` and `.sqlx/**`). A DB-backed test asserting `?is_service_account=false` returns only people **and** reports the matching `total` would be the natural addition there — I did not add one I could not execute.

## API Changes
- [ ] N/A - no API changes

No new endpoint and no migration. `ListUsersQuery` already derives `IntoParams` and is already wired as `params(ListUsersQuery)`, and `AdminUserResponse` already derives `ToSchema`, so both the new parameter and the new field reach the OpenAPI document with no annotation change. The change is additive: an existing client that sends no filter and ignores the new field is unaffected.


sblaisot:
Downstream consumer: artifact-keeper-web#826 uses the `is_service_account` filter this PR adds to make the admin Users page list people only. It is the case that could **not** be fixed client-side — that page is server-side paginated, so filtering in the browser would leave `total` counting rows the page no longer shows. It is why this PR filters the pagination count as well as the page query.

brandonrc:
Reviewed — looks good: the filter is applied to both the page query and the count query, the default stays unfiltered, and the `.sqlx` files are regenerated for the new bind. The source-comparison test for the two `WHERE` clauses is unusual but it is the only way to hold that invariant without a database in a unit test, and the comment explains why.

The branch had a CHANGELOG conflict with `main` (#3641 opened the same `### Fixed` section). I pushed a merge of `main` to your branch with the conflict resolved — both entries kept, nothing else touched. Will merge once CI is green.

PR #3643, opened 2026-09-01
fix(go): answer sumdb /supported instead of forwarding it

### LINKED ISSUE #3642: GOPROXY: sumdb/<host>/supported is forwarded upstream, so go never proxies checksum lookups
## Description

The GOPROXY sumdb handler forwards `sumdb/<host>/supported` to the upstream checksum database. That path is not part of the checksum database protocol, it is a capability handshake the *proxy* is meant to answer about itself, so sum.golang.org returns 404 for it and we pass that 404 back to the client.

Per the toolchain (`cmd/go/internal/modfetch/sumdb.go`, `initBase`):

> Before accessing any checksum database URL using a proxy, the proxy client should first fetch `<proxyURL>/sumdb/<sumdb-name>/supported`. If that request returns a successful (HTTP 200) response, then the proxy supports proxying checksum database requests. [...] If the check fails with a "not found" (HTTP 404) or "gone" (HTTP 410) response [...] the client should connect directly to that database.

So `go` reads our 404 as "this proxy cannot proxy the sumdb" and connects straight to sum.golang.org for every `go.sum` verification. Consequences:

- The existing `lookup/` and `tile/` handlers in `proxy_sumdb` are never reached in practice, they only respond to hand-made requests.
- Clients on hosts without direct internet egress cannot verify `go.sum` at all, even with `GOPROXY` correctly pointed at Artifact Keeper. Module downloads succeed, checksum verification does not.
- Checksum lookups bypass the instance entirely, so they are neither cached nor visible in its logs.

The failure is silent for anyone whose build hosts have egress, which is presumably why it has gone unnoticed.

## Environment Information

- Deployment type: Kubernetes
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL 16
- Reverse Proxy: yes, ingress in front of the backend
- Network setup: public hostname, some client hosts have no direct outbound egress

## Backend Version

1.8.1

## Web UI Version

N/A

## Package Format

Go Modules

## Expected Behavior

`GET /go/{repo_key}/sumdb/{host}/supported` returns 200 with an empty body for an allowlisted host, so the toolchain routes its checksum lookups through the instance and the `lookup` and `tile` handlers get used.

## Observed Behavior

It returns 404, forwarded verbatim from the upstream checksum database, and the toolchain falls back to a direct connection to sum.golang.org.

## Steps to Reproduce

1. Create a Go remote repository with upstream `https://proxy.golang.org/`, key `go`.
2. `curl -s -o /dev/null -w '%{http_code}\n' https://<host>/go/go/sumdb/sum.golang.org/supported` and observe 404.
3. Confirm the same 404 comes from upstream, which is correct on its side: `curl -s -o /dev/null -w '%{http_code}\n' https://sum.golang.org/supported`.
4. Confirm the handlers that would be used do work: `curl -s https://<host>/go/go/sumdb/sum.golang.org/lookup/github.com/pkg/errors@v0.9.1` returns 200.
5. On a host with no outbound egress except to the instance, `GOPROXY=https://<host>/go/go go get github.com/pkg/errors@v0.9.1` fails on checksum verification, while the same command on a host with egress succeeds by talking to sum.golang.org directly.

## Additional Context

Note the flip side of fixing this, from the same block of toolchain source: once `/supported` answers 200, the client "should use the proxied access method only, never falling back to a direct connection to the database". Checksum verification then depends on the instance being able to reach the upstream sumdb, which is the intent, but it does turn a silent bypass into a visible failure.

Related but separate, in the same function: for a `GOSUMDB` host outside the allowlist, `/supported` answers 403. The toolchain treats anything other than 200/404/410 as the database being unavailable, so that is a hard error rather than the graceful direct fallback a 404 would produce. Happy to fix that in the same PR or a follow-up.

nicola-preda:
## Summary

Fixes #3642.

`sumdb/<host>/supported` is a proxy-side capability handshake, not an endpoint of the checksum database, so forwarding it upstream returned sum.golang.org's natural 404. The toolchain reads that as "this proxy cannot proxy the sumdb" and connects directly to sum.golang.org instead, which meant the `lookup` and `tile` handlers below were never reached and hosts without direct egress could not verify `go.sum` at all.

This answers the probe locally with 200 and an empty body, after the SSRF allowlist check so a non-allowlisted `GOSUMDB` host still gets the same rejection as before.

`test_sumdb_supported_answered_locally` asserts the probe returns 200. On `main` it fails (the forwarded request returns 404); with the fix it passes, and it needs no network because the response is now produced locally.

## Test Checklist
- [x] Manually tested locally

Verified against a live instance: `@v/list`, `@latest`, `.info`, `.mod` and `sumdb/.../lookup/...` all served correctly, and `go get` through the proxy resolves and downloads. Ran `cargo test --lib goproxy::tests::test_sumdb` (6 passed, including the allowlist tests) and `cargo fmt --check`.

## API Changes
- [x] N/A - no API changes

The route already existed, only its response for one path changed.


brandonrc:
Reviewed — the fix is right and minimal: `supported` is a proxy-side capability handshake, answering it after the host allowlist check is the correct placement, and the unit test pins it without a network call.

One thing the PR was missing for this repo's release process: the release preflight reconciles every merged PR against the pending `## [Unreleased]` CHANGELOG section (by PR number or the issue it closes), and a `fix:` commit with no entry blocks the cut. I pushed two commits to your branch (edits-by-maintainers was on): a merge of `main`, and a `docs(changelog)` commit describing #3642 under `### Fixed`. No code changes. Will merge once CI is green.

Issue #2424, opened 2026-07-11
npm virtual scope policy: extend enforcement to tarball + direct-remote paths, glob patterns, create-time config & UI


brandonrc:
Follow-up to #2327. The 1.5.1 patch adds an optional per-repository npm scope allow-list (`npm_allowed_scopes`/`npm_allow_unscoped`, stored in `repository_config`) and honors it on the two npm **virtual** candidate-selection loops (metadata + packument) for Remote members. This issue tracks the remaining, larger scope:

1. **Tarball path** — enforce the policy on `resolve_virtual_download` so a pinned-lockfile direct-tarball request for an out-of-scope name is also refused (needs a per-member exclusion predicate threaded into the shared, format-agnostic download seam — analogous to how #2066 extended the age gate to tarballs).
2. **Direct-remote path** — `GET /npm/<remote-key>/<pkg>` should also honor the member's policy, so pointing at a member directly doesn't bypass it.
3. **`allowed_name_patterns`** — glob allow-lists beyond bare `@scope` prefixes.
4. **Config surface** — fold the policy into repo create/update (`CreateRepositoryRequest`/`UpdateRepositoryRequest` + `RepositoryResponse` + OpenAPI) and add a web UI control.
5. **Generalize the model** to other formats (the #1217 cross-format parallel).

Default behavior stays unchanged (no policy set ⇒ allow all).

PR #3651, opened 2026-09-01
fix(npm): decompress gzip request bodies on the npm audit routes

### LINKED ISSUE #3644: fix(npm): npm audit returns a silent false negative — gzip request bodies are never decompressed
## Summary

`npm audit` against an Artifact Keeper Remote npm repository reports **"found 0
vulnerabilities"** for a dependency tree containing known-vulnerable packages. It does
not error and does not warn. Any CI job running `npm audit` through Artifact Keeper gets
**no vulnerability coverage and no signal that coverage is missing**.

Verified on 1.8.1; the code paths below are unchanged on `main` (1.8.2).

## Root cause

npm sends the bulk-advisory POST with `Content-Encoding: gzip`. Two things then go wrong:

1. **The request body is never decompressed.** Both handlers take the raw body
   (`body: Bytes`) and there is no request-decompression layer anywhere in the stack —
   `tower-http` is enabled with `compression-gzip`/`compression-br`, which are
   *response*-side only.

2. **The `Content-Encoding` header is dropped while the still-gzipped body is
   forwarded.** `npm_audit_upstream_json` hardcodes the outgoing content type:

   ```rust
   let req = client
       .post(&url)
       .header(CONTENT_TYPE, "application/json")
       .body(body);            // still gzip-compressed, no Content-Encoding sent
   ```

   Upstream therefore receives gzip bytes labelled `application/json`, rejects them, and
   the non-success branch returns `None`.

`None` collapses to `empty_fallback()` — `{}` with **HTTP 200** — which npm reads as
"no advisories". Every failure mode converges on that same empty 200: transport error,
over-cap body, non-success upstream status, or a non-Remote repo type.

**The scope-policy path fails the same way, despite intending the opposite.** With an
active npm scope policy, a gzipped body reaches `filter_bulk_advisory_request`, fails to
parse, and takes the branch commented *"Unparseable body ⇒ fail closed"* — which also
returns `empty_advisories_bulk_response()`. Fail-closed and fail-open emit an identical
body, so the distinction has no observable effect.

## Reproduction

```bash
BODY='{"lodash":["4.17.15"],"minimist":["1.2.0"]}'
printf '%s' "$BODY" > /tmp/b.json && gzip -c /tmp/b.json > /tmp/b.json.gz

# plain JSON -> correct advisories
curl -sS -H "Authorization: Bearer $TOK" -H 'Content-Type: application/json' \
  --data-binary @/tmp/b.json \
  "$BASE/npm/<remote-repo>/-/npm/v1/security/advisories/bulk" | wc -c   # => 2423

# gzipped (what npm actually sends) -> empty
curl -sS -H "Authorization: Bearer $TOK" -H 'Content-Type: application/json' \
  -H 'Content-Encoding: gzip' --data-binary @/tmp/b.json.gz \
  "$BASE/npm/<remote-repo>/-/npm/v1/security/advisories/bulk" | wc -c   # => 2  ("{}")

npm audit --registry "$BASE/npm/<remote-repo>/"    # => found 0 vulnerabilities
```

A local listener confirms what npm sends: `content-encoding: gzip`, body starting `1f 8b`.

## Expected

1. Decompress `Content-Encoding: gzip` (and `deflate`) request bodies on the two audit
   POST routes.
2. An audit that **cannot be performed** must be distinguishable from "clean". Returning
   an empty advisory set on failure is the wrong default for a security endpoint: it is
   byte-identical to a genuine all-clear. The 404 this behaviour replaced in #1400 was
   safer, because it failed visibly.

## Affected code

- `backend/src/api/handlers/npm.rs` — `security_advisories_bulk` (~:1828),
  `security_audits_quick` (~:1879), `npm_audit_upstream_json` (~:1534),
  `empty_advisories_bulk_response` (~:1136)

## Acceptance criteria

- `npm audit` against a Remote npm repo reports the same advisories as the equivalent
  plain-JSON call.
- A gzipped bulk-advisory request returns the same body as the identical uncompressed request.
- An induced upstream failure surfaces as an error status, not `0 vulnerabilities`.

## Notes

`flate2` is already a direct backend dependency, so (1) needs no new crate. I have a fix
for the decompression half working locally and am happy to open a PR.

knowinglyAnonymous:
Closes #3644

## What

`npm audit` against a Remote npm repository reported **"found 0 vulnerabilities"** for a
dependency tree with known-vulnerable packages, with no error and no warning.

npm sends the bulk-advisory POST with `Content-Encoding: gzip`. Nothing in the stack
decompressed it: the handlers take the body as raw `Bytes`, and there is no
request-decompression layer — the `tower-http` compression features enabled for the router
are response-side only. The still-gzipped body was then forwarded upstream labelled
`Content-Type: application/json` with **no `Content-Encoding`**, so upstream rejected it,
and the non-success branch fell through to `empty_advisories_bulk_response()` — `{}` with
HTTP 200, which npm renders as a clean audit.

This decodes the request body before anything inspects or forwards it.

## Changes

- New `decode_audit_request_body(&HeaderMap, Bytes) -> Result<Bytes, Response>` in
  `backend/src/api/handlers/npm.rs`:
  - absent header / empty / `identity` → pass through unchanged
  - `gzip`, `x-gzip` → gunzip; `deflate` → inflate (matching is case- and
    whitespace-insensitive per RFC 9110)
  - anything else → `415 Unsupported Media Type` rather than forwarding blind
  - a body that does not decode → `400 Bad Request`
  - decoded size capped at 16 MiB (`MAX_AUDIT_REQUEST_BODY_BYTES`) → `413` when exceeded,
    so a small gzip payload cannot expand into an unbounded allocation
- Both `security_advisories_bulk` and `security_audits_quick` now take `headers: HeaderMap`
  and decode first, so the fix covers the modern and legacy audit routes and applies on
  both the verbatim-proxy path and the scope-policy path (#2424).
- Uses `flate2`, already a direct backend dependency. No new crates.

## Why an error status rather than an empty advisory set

An audit that could not be *performed* must be distinguishable from an audit that found
nothing — the two were previously byte-identical. This is also why the scope-policy
branch's `"Unparseable body ⇒ fail closed"` comment had no observable effect: it returned
the same empty 200 as the fail-open path. With the body decoded, that branch now only
triggers on a genuinely malformed JSON body.

## Deliberately not in this PR

#3644 also proposes that the *upstream-failure* paths (transport error, non-success status,
over-cap response) stop degrading to an empty 200. That is a semantic change to how `npm
audit` behaves when the upstream registry is unreachable, and it deserves its own decision
— an operator relying on the current graceful degradation would see audits start failing.
Left for maintainer input on the issue. This PR fixes the false negative that occurs when
everything is working normally, which is the case that made the endpoint report a false
all-clear on every ordinary CI run.

## Tests

Eight unit tests in `npm::tests`, all pure and DB-free:

- `test_decode_audit_request_body_gunzips_gzip_encoding` — the regression itself: asserts
  the fixture is genuinely compressed (gzip magic `1f 8b`), that it decodes to exactly the
  plain body, and that the result parses as JSON
- `test_decode_audit_request_body_accepts_x_gzip`
- `test_decode_audit_request_body_inflates_deflate_encoding`
- `test_decode_audit_request_body_passthrough_when_uncompressed` — absent and `identity`
- `test_decode_audit_request_body_encoding_match_is_case_insensitive` — `" GZIP "`
- `test_decode_audit_request_body_malformed_gzip_is_an_error` — 400, not an empty audit
- `test_decode_audit_request_body_unsupported_encoding_is_refused` — 415
- `test_decode_audit_request_body_rejects_decompression_bomb` — 413, not silent truncation

## Verification

```
cargo fmt --check
cargo test --workspace --lib npm::
```


brandonrc:
Reviewed — this is a solid fix. The decode-before-inspect placement is right (the Remote path parses and filters the body before forwarding, so it had to be plaintext by then), the decoded-size cap with `take(limit+1)` correctly distinguishes over-cap from truncated, `deflate` → zlib matches RFC 9110, and turning a body that cannot be decoded into a 400/415 rather than an empty `{}` is exactly the behaviour #3644 asks for: an audit that could not run must not look like a clean one.

One thing missing for this repo's release process: the release preflight reconciles every merged PR against the pending `## [Unreleased]` CHANGELOG section, and a `fix:` commit with no entry blocks the cut. I pushed two commits to your branch (edits-by-maintainers was on): a merge of `main`, and a `docs(changelog)` commit describing #3644 under `### Fixed`. No code changes. Will merge once CI is green.

Issue #3540, opened 2026-08-24
Floating tags (:latest, :X.Y) are published before the release gate, and a partial publish can move them


brandonrc:
## Summary

`:latest` and the floating `X.Y` series tags are applied by `docker-publish.yml` at
tag-push time — which is, by construction, **before** the release gate has certified
anything, because docker-publish is what builds the bytes the gate tests. Two
separate mechanisms make this worse than a timing detail.

### 1. `:latest` moves before the gate, on the normal path

`merge-backend` / `merge-openscap` apply `type=raw,value=latest` and
`type=semver,pattern={{major}}.{{minor}}` on `startsWith(refs/tags/v) && !contains('-')`.
On **v1.8.1** docker-publish finished at 14:59:49 and the release published at
16:17:54: `:latest` named gate-uncertified bytes for **78 minutes** on a cut where
nothing went wrong. That is the ordinary path, not a failure path.

### 2. A partial publish can move a floating tag

`merge-backend`, `merge-openscap` and `merge-scanner-adapter` are parallel siblings
with no fan-in before the tags are written, so backend's floating tags publish the
moment backend's build and scan pass — regardless of whether the other two images
succeeded. This is the shape that let **v1.7.2** publish `backend:1.7.2` and
`backend:latest` and stay publicly pullable for ~9 hours with the gate never run,
for a version that was then retired.

### 3. No rollback exists

Neither `release.yml` nor `docker-publish.yml` has any path that moves a floating
tag back.

### Live residue

`ghcr.io/artifact-keeper/artifact-keeper-backend:1.7.5` currently returns HTTP 200 —
a release that never shipped, whose git tag was deleted, and whose bytes the gate
never certified. Deleting it is a separate destructive decision and is **not** part
of this issue.

## Why the existing promote path is not already the fix

#2698 built the mechanism (`docker-publish.yml -f promote_version=X.Y.Z` re-applies
an existing digest through the digest-aware guard, build-free) and was closed, but
the post-gate **wiring** was never done: nothing dispatches it after the gate, and
the promote path deliberately refuses to *advance* a floating tag (it only mirrors
ghcr's current state), so it cannot be used for this as-is.

## Proposed fix

**A — reorder (no new machinery).** Move floating-tag application out of the three
parallel merge jobs into a single job that `needs:` all three, so a partial publish
cannot move a floating tag. `success()` fan-in is the control; no `if:`.

[...]

Issue #3640, opened 2026-09-01
release: resolve-candidate-digest cannot consume a promote dispatch, deadlocking a re-tagged release


brandonrc:
## The deadlock (live right now on v1.8.2)

`release.yml` job `resolve-candidate-digest` locates the docker-publish run for the release commit with a query hard-filtered to `event=push`. That interacts badly with two mechanisms that are each working exactly as intended:

1. **Image immutability**: re-tagging a version at a new commit makes the tag push's docker-publish fail at "Refuse to overwrite a published backend version".
2. **The documented recovery** (#2698): `gh workflow run docker-publish.yml --ref vX.Y.Z -f promote_version=X.Y.Z` succeeds — it re-applies all tags to the already-published digests without rebuilding. But it is `event=workflow_dispatch`, so the resolver never sees it.

Concrete sequence on v1.8.2:

| run | event | head_branch | conclusion |
|---|---|---|---|
| 33514800258 | push | v1.8.2 | **failure** (immutability, correct) |
| 33522347257 | workflow_dispatch | v1.8.2 | **success** (promote, correct) |

`resolve-candidate-digest` finds only the failed push run and blocks the release. The push run can never be made green (immutability is doing its job), so the release is permanently deadlocked: no path exists to cut v1.8.2 even though the bytes are published, immutable, and about to be gate-tested by digest.

## Fix

Accept a **successful** docker-publish run for the same `head_sha` **and** `head_branch == ref`, whether `event=push` or `event=workflow_dispatch`, preferring the most recent success. Still fail closed: no successful run of either kind ⇒ BLOCKED (a missing or failed publish must stay RED), and a success for a different SHA must never be picked.

This does not weaken bytes-tested == bytes-published: a promote dispatch rebuilds nothing (build jobs are skipped in PROMOTE mode) and the digest-aware republish guard means `:X.Y.Z` can only ever be re-applied to the same digest — and the gate reads the digest it tests from the registry, not from the run.

Extract the run-selection into `scripts/ci/resolve-candidate-publish-run.sh` with a `test-*.sh` self-test per the `scripts/ci` convention, so the "push failed + dispatch succeeded" leg is exercised offline.

PR #3653, opened 2026-09-01
ci(docker-publish): let apply-floating-tags run in PROMOTE mode

### LINKED ISSUE #3652: docker-publish: apply-floating-tags can never run in PROMOTE mode, so post-gate floating promotion has never executed
## Symptom (live on v1.8.2)

The v1.8.2 release published correctly (exact tags, digest-verified), but release.yml's `Promote Floating Tags` job failed its digest assert: `:1.8` / `:latest` still name v1.8.1's digest. The promote dispatch it fired (run 33541126375, `promote_version=1.8.2 promote_floating=true`) concluded **success** — with `Apply Floating Tags` **skipped**.

## Root cause

`apply-floating-tags` deliberately has no `if:` — the comment says the default `success()` on the three merge jobs IS the control. But GitHub evaluates a job's default status over the **transitive** dependency chain, not just direct `needs`. In PROMOTE mode every build job (`Backend (platform)`, `Security Scan`, Alpine) is skipped by design; the merge jobs run anyway via their explicit promote-mode `if:` overrides; and then `apply-floating-tags`, with the plain default, is skipped because of the skipped grandparents — every time, in exactly the mode #3540 built it for.

Evidence across recent runs (`Apply Floating Tags` conclusion):

| run | event/ref | apply |
|---|---|---|
| 33541126375 | dispatch v1.8.2 (promote+floating) | **skipped** |
| 33532902112 | dispatch v1.8.2 (promote) | **skipped** |
| 33522347257 | dispatch v1.8.2 (promote) | **skipped** |
| 33530623075 | dispatch main (candidate build) | success |
| 33527837521 | push main | success |
| 33474744542 | push v1.8.2 (original build) | success |

It has never once run on a promote dispatch. This also silently disabled the plain-promote behavior of mirroring ghcr's current floating state to Docker Hub.

## Fix

Give `apply-floating-tags` an explicit condition that re-states the sibling barrier over **direct** results only:

```yaml
if: >-
  ${{ !cancelled()
      && needs.publish-preflight.result == 'success'
      && needs.merge-backend.result == 'success'
      && needs.merge-openscap.result == 'success'
      && needs.merge-scanner-adapter.result == 'success' }}
```

A failed OR skipped merge sibling still blocks every floating move (the v1.7.2 partial-publish hole stays closed); skipped build grandparents no longer do.

`verify-published` has the same transitive exposure (`if: startsWith(github.ref, 'refs/tags/v')` gets an implicit `success()`), but the release workflow independently asserts the floating digests against the gate-tested digest — that assert is what caught this — so it is left as-is here and can be considered separately.

## Cleanup for v1.8.2

After the fix merges: dispatch `promote_version=1.8.2 promote_floating=true` (from main, which carries the fixed workflow and the identical scanner-adapter VERSION), confirm `:latest`/`:1.8` resolve to the gate-tested digest on both registries, then re-run the release run's `Promote Floating Tags` job so its assert goes green.

brandonrc:
## Summary

Closes #3652.

`apply-floating-tags` is the only job that writes a floating tag, and the post-gate promotion designed in #3540 depends on it running on a `promote_version` dispatch. It never has: its control was the implicit default `success()`, which GitHub evaluates over the **transitive** dependency chain, and a promote dispatch skips every build/scan job by design (the merge jobs run via their explicit promote-mode `if:` overrides). Every promote dispatch ever made concluded green with `Apply Floating Tags` skipped — v1.8.2 (runs 33541126375 / 33532902112 / 33522347257) was the first stable release to exercise the path, and release.yml's floating-digest assert caught `:1.8` still naming v1.8.1's digest.

The fix states the sibling barrier explicitly over **direct** results: `!cancelled()` plus `needs.<publish-preflight, merge-backend, merge-openscap, merge-scanner-adapter>.result == 'success'`. A failed **or skipped** merge sibling still blocks every floating move — the v1.7.2 partial-publish hole stays closed — while intentionally-skipped build grandparents no longer veto the job. Normal push builds are unaffected (all four results are 'success' there today, and the job already ran).

`verify-published` has the same transitive exposure (`if:` without a status function gets an implicit `success()`), but release.yml independently asserts the floating digests against the gate-tested digest — that assert is what caught this — so it is deliberately not changed here (see #3652).

`scripts/ci/check-floating-tag-promotion.sh` invariant 2 used to refuse any job-level `if:` on `apply-floating-tags` (that is what failed the Shell Tests job on the first push here) and encoded the exact assumption the bug disproved. It now models how GitHub schedules the writer: it evaluates the job's condition (status functions, `&&`/`||`/`!`, quoted strings, `needs.<direct need>.result`; anything else is refused rather than trusted) over the transitive `needs` chain, and requires that a promote dispatch — preflight and every merge succeeded, every build/scan ancestor skipped — RUNS the job, while a run where any live merge job failed, was cancelled or was skipped does NOT, in both push and promote mode. Against `main`'s workflow the gate fails with `apply-floating-tags would be SKIPPED on a PROMOTE dispatch (...)`; with this branch's `if:` it passes.

`scripts/ci/test-check-floating-tag-promotion.sh` fixtures now carry build jobs that skip on promote (as the real workflow does) and pin: no `if:` refused (the #3652 shape), an `if:` without a status function refused, bare `!cancelled()`/`always()` refused, a barrier missing one merge refused, a `!= 'failure'` barrier refused (a skipped merge would run it), an expression the model cannot evaluate refused, the direct-result barrier accepted. 19 cases pass. The live proof remains the next promote dispatch: `Apply Floating Tags` must conclude with something other than `skipped`, and release.yml's floating-digest assert goes red if it ever silently stops again.

## Test Checklist
- [x] Unit tests added/updated — `test-check-floating-tag-promotion.sh` (shell-tests job)
- [x] Manually tested locally — `bash scripts/ci/test-check-floating-tag-promotion.sh` (19 passed), `check-floating-tag-promotion.sh` against `main`'s workflow (fails as intended), `actionlint` on docker-publish.yml (same 53 pre-existing shellcheck notes as main, none new), `shellcheck` on both scripts; YAML parses; surveyed the last 15 docker-publish runs: `apply` = success on every normal build (needs' results all 'success' there, so the new condition preserves them) and skipped on every promote (the case this enables)

## API Changes
- [x] N/A - no API changes


brandonrc:
## Audit — release-flow safety

**Verdict: merge on green.** The `if:` is correct in every case: the reviewer drove the gate's own evaluator over all 256 result combinations of the four direct `needs` in both push and promote modes, with zero mismatches against "runs iff all four succeeded"; a cancelled ancestor makes `!cancelled()` false, so the partial-publish concern the old header protected still holds. Main's no-`if:` shape runs on push and is skipped on promote, which is #3652 exactly and is why the documented recovery dispatch has been unable to move a floating tag since #3544.

The guard change is the substantive part. `check-floating-tag-promotion.sh` refused any condition on that job on the theory that GitHub's default success handling was the control, the assumption #3652 disproved. It now models GitHub's scheduling and refuses whatever it cannot evaluate: 19 self-test cases plus 13 throwaway fixtures reject `barrier || always()` (the precedence trap), bare `success()`, `!= 'failure'` barriers, non-direct `needs.build-*.result`, double-quoted strings, the `outputs` context, YAML `true`, and case-mismatched strings, while accepting De Morgan and status-function-last forms. It fails on main's workflow with the #3652 message and passes on this head and on the trial merge with #3776, which is the next PR in line and depends on this one.

Landed: the merge with main had duplicated the #3640/#3641 bullet into `[1.9.0]`; removed, it lives under `[1.8.2]`. The `#3652` bullet is under `[Unreleased] / ### Fixed`, first reference, single occurrence.

Nits left: a bare, un-wrapped `!cancelled()` crashes the gate with a YAML exception (still non-zero, still closed; GitHub rejects that spelling too), and `cancelled()` is a conservative proxy worth a comment.

Issue #3556, opened 2026-08-25
Proxy fetches through the format-less helpers still synthesize a Generic repository, so immutable content caches for 5 minutes (RPM, conda, OCI inline scan, generic download route)


brandonrc:
## Summary

`proxy_helpers::build_remote_repo` synthesizes the upstream `Repository` a proxy fetch is performed against, and it hardcodes `RepositoryFormat::Generic`. `Generic` has no `cache_classifier::classify` arm, so every path reaching it falls through to the conservative `MUTABLE_DEFAULT_TTL_SECS` (300 s) — including paths the *real* format's classifier marks `Immutable`.

The consequence is upstream amplification: content that can never change is re-fetched every five minutes, forever.

This is the third time this class has been found:

- **#3206** — OCI layers expired five minutes after a pull. Fixed by **#2312**, which threaded the real format into `proxy_fetch_capped_with_accept` and `proxy_fetch_streaming_with_cache_key`.
- **#3459** — Maven/Gradle checksum sidecars, and sbt artifacts, on the 5-minute TTL. Fixed by threading the format into `proxy_fetch_capped_budgeted` and adding `proxy_fetch_streaming_with_format`.
- **this issue** — the sites neither of those touched.

Each fix so far has threaded the format through one helper at a time. This issue is for the remainder, filed rather than swept because the failure direction is asymmetric (see *Why this was not swept* below).

## The unfixed sites

All of these serve or stage content the format's own classifier marks `Immutable`, through a helper with no `format` parameter, and therefore cache it for 300 s.

| site | helper | cache path shape | classifier says | pinned by |
|---|---|---|---|---|
| `rpm.rs:1397` | `proxy_fetch_streaming_with_disposition` | `Packages/foo-1.2-3.x86_64.rpm` | `Immutable` | `cache_classifier.rs:905` |
| `rpm.rs:296` | `proxy_fetch_capped` | `repodata/<hex>-primary.xml.gz` | `Immutable` (content-addressed) | `cache_classifier.rs:896` |
| `conda.rs:2689` | `proxy_fetch_streaming_with_disposition` | `linux-64/<pkg>.conda`, `.tar.bz2` | `Immutable` via `is_pypi_package_file` | `cache_classifier.rs` PyPI rows |
| `oci_v2.rs:8766` | `proxy_fetch_capped` | `v2/<img>/blobs/sha256:…` | `Immutable` | `cache_classifier.rs:694` |
[...]

PR #3656, opened 2026-09-02
feat(github): finite release caching with mise and aqua aliases

### LINKED ISSUE #3657: Support Artifact Keeper as a caching mirror for mise's aqua backend (GitHub release assets)
### Problem Statement

Teams using [mise-en-place](https://mise.jdx.dev) install most CLI tools through its aqua backend, and aqua packages are overwhelmingly `github_release` type — plain HTTPS downloads from `github.com/<owner>/<repo>/releases/download/<tag>/<asset>`. Today every developer machine and CI job downloads those assets straight from GitHub: no shared cache, repeated egress for identical bytes, exposure to GitHub outages, and rate-limit pressure from shared CI egress IPs. Artifact Keeper has no documented recipe for standing in as a mirror for this traffic, and its `generic` remote proxy caches unrecognised paths under the conservative 5-minute mutable TTL — so even when used as a mirror, it revalidates tag-pinned (effectively immutable) release assets against GitHub on a schedule, which is exactly the traffic a mirror exists to absorb.

### Proposed Solution

Support Artifact Keeper as a caching mirror for mise's aqua backend, without inventing any aqua-specific server protocol:

1. A `generic` remote repository with `upstream_url: https://github.com` is the mirror; mise ≥ v2025.9.3 redirects its GitHub release downloads to it via [`settings.url_replacements`](https://mise.jdx.dev/url-replacements.html) (the same mechanism its docs describe for Artifactory generic remotes).
2. Teach the cache classifier that the tag-pinned `{owner}/{repo}/releases/download/{tag}/{asset}` path shape on `generic` repositories is immutable (cache forever, never revalidate), while every other generic path keeps the conservative mutable default and generic coordinates stay swap-protected.
3. Pin the redirect behaviour the flow depends on with a regression test: GitHub 302s asset downloads to S3-signed `objects.githubusercontent.com` URLs, and the proxy's client must keep following that hop while stripping `Authorization` cross-origin (a repository-configured GitHub PAT must reach github.com but never the S3 host).
4. Ship a documented recipe (`docs/`) plus a native E2E test that runs a real pinned `mise install aqua:jqlang/jq@…` through the proxy.

### Alternatives Considered

- **A first-class `aqua` repository format** (new `RepositoryFormat` variant, migration, handlers, routes): rejected — aqua/mise need no server-side protocol beyond file passthrough, so the ~16-file format addition buys only catalog identity for a lot of surface area.
- **Supporting the standalone aqua CLI (aquaproj)**: it has no `url_replacements` equivalent, so there is no client-side hook to redirect its downloads; out of scope.
- **Proxying mise's side channels** (version listing via `api.github.com`, aqua registry metadata): deliberately out of scope for the first cut; pinned versions work fully through the mirror.

### Use Case

As a platform engineer, I want mise's aqua-backend tool downloads in our org to flow through our Artifact Keeper instance so that CI fleets share one cache, tool installs keep working during GitHub incidents for already-cached assets, and GitHub rate limits stop breaking builds.

### Component

Backend / API

### Additional Context

- mise `url_replacements` docs: https://mise.jdx.dev/url-replacements.html (added in v2025.9.3; supports regex captures, documented for corporate mirrors)
- aqua registry reference: https://aquaproj.github.io/docs/reference/registry/ (`github_release` package type)
- The `routing_rules` module already names "proxying GitHub Releases through a raw/generic remote repository" as a supported use case; this issue completes that story for mise/aqua clients (caching semantics + docs + E2E).

### Pre-submission Checklist

gilesw:
Adds dedicated `github`, `mise` and `aqua` repository formats for GitHub release mirroring. They share GenericHandler and `/general/<repo>/…`, with a seven-day finite release cache. Existing Generic repositories retain their cache and hosted-file write behaviour.

Closes #3657

## Changes

- Adds PostgreSQL enum values and format aliases. The general download route preserves these formats when selecting the cache policy, including streaming downloads.
- Tag-named release assets and checksum files expire after seven days by default. Repository TTL overrides still win. Existing revalidation detects replacement; other paths retain the conservative short lifetime.
- Documents private authenticated repositories and the upstream token trust boundary: every mirror reader must be trusted with everything that token can read. These formats do not add a path allow-list.
- Uses committed mise lockfiles for outage installs; extends neither GitHub API caching (#3658) nor standalone aqua CLI support.
- Moves native-client E2E coverage to the companion artifact-keeper-test change. The redirect/authentication test characterises existing cross-origin credential stripping.

## Validation

- After merging upstream 07a521fe and renumbering the migration to 218, the focused backend run passed all 196 tests, including database enum round trips, download-route TTLs, generic deletion/replacement, cache overrides, buffered and streaming revalidation, changed bytes, missing validators and upstream errors.
- Formatting, Clippy (`--workspace --all-targets -- -D warnings`), backend build and all 5 workflow scan-gate tests passed.
- Companion suite against the final binary: 22 passed, including authenticated locked mise installs with empty client caches after a controlled upstream was removed.
- Full library run: 16,246 passed and 41 failed. One failure was a new test's asynchronous metadata wait; that was corrected and the final focused run passes. All other 40 failures reproduced on unchanged upstream commit 7913a61e in a separate baseline run. Most are this host's non-loopback mock connectivity failures. This is not a claim that the whole suite is green.

## API and migration

Adds three accepted `format` values; no new endpoints. Migration 218 adds enum values, which PostgreSQL cannot simply remove on rollback. Roll back only after removing or converting repositories using those values. Companion UI and E2E changes are tracked separately; target milestone remains 1.11.0.

## Connected changes

- Backend: artifact-keeper/artifact-keeper#3656
- UI: artifact-keeper/artifact-keeper-web#846 (issue #845 in that repository)
- End-to-end tests: artifact-keeper/artifact-keeper-test#399 (issue #398 in that repository)

Land the backend support before the UI. The test suite gates released backends at 1.11.0; use its explicit feature-branch override for testing earlier-version development binaries.


brandonrc:
## Review — NEEDS-WORK

Thanks for this — mirroring mise/aqua through a generic remote is a real use case and the caching idea is sound. Three things need to change before it can land, and the first is the one that decides the shape of the rest.

**Blocker: the classifier is not only a cache oracle, so this changes write semantics.** `classify()` is also consulted by three write guards — `repositories.rs:8961` (`delete_blocked_by_immutability`, called unconditionally at `:9059` with no `repo_type` check), `artifact_service.rs:723`, and `handlers/mod.rs:78`. With this change, in a **hosted** generic repository a version-less blob at `myorg/myapp/releases/download/v1.0/app.tar.gz` becomes undeletable by a non-admin (409) and un-replaceable after deletion. Both are allowed on main today. I confirmed it with a throwaway unit test against the delete guard: the new path is blocked, the control path still deletes. The PR's argument that generic is untouched rests on `is_explicitly_mutable_index` being unchanged, but that function is the gate, not the oracle — the oracle is `classify()` itself. The fix is to make the immutability decision cache-path-local, or to key it on something the write guards cannot see.

**Should-fix: the documented recipe sets up a confused deputy.** `docs/mise-aqua.md:28-50` creates `is_public: true` generic remotes to both `github.com` and `api.github.com`, and `:91-100` recommends a GitHub PAT on the API one. Generic remotes forward any path and there is no allow-list anywhere in the codebase, so an anonymous caller can request `/general/github-api/repos/<private-org>/<private-repo>` and be served with the operator's token; the URL-keyed cache then hands the result to everyone else. The docs warn about the redirect hop, which is the smaller of the two risks. Either scope the remote to a path prefix, or document it as an authenticated repository only.

**Should-fix: "immutable" here is shape-only, and the stated mitigation does not hold.** The cache key carries no digest and `classify()` never sees `upstream_url`, so the rule fires for every generic remote — including the `Generic` stand-ins that `build_remote_repo` synthesises when proxying maven and debian. The code comment leans on clients verifying checksums, but `sha256sum.txt` matches the same six-segment shape and is frozen alongside the asset, so an upstream delete-and-reupload is invisible to that verification. GitHub release assets can be replaced under the same tag and name, so this is reachable, not theoretical.

Tests: the six classifier rows do fail on main — I checked by disabling the new arm and watching them assert `Mutable{300}` against the expected immutable. The redirect test passes unchanged on main, so it characterises existing behaviour rather than pinning this change; worth saying so in the PR.

Bookkeeping: the CHANGELOG bullet carries no `#NNNN`, which trips release preflight check 5. Cite `(#3657)` as the first reference on the bullet.

Happy to talk through the first finding before you rewrite anything — there is more than one reasonable way to keep the cache benefit without touching deletion, and which one fits depends on whether you want hosted generic repositories to be able to opt in later.

gilesw:
Thanks for the review @brandonrc. I don't know your project well, but I've come up with a few options chatting with my ai agent. My gut feel was that it should just be a dedicated repo type more for usability over any technical reasons. My agent came up with 3 options:-

  1. Keep generic repositories and use the existing configurable cache lifetime. Remove the GitHub-specific
     immutable classification and document a finite cache lifetime instead. This looks like the smallest change,
     although it would not by itself provide the extended outage behaviour we wanted.

  2. Add an opt-in GitHub mirror cache policy to generic remotes. Keep existing generic behaviour unchanged, with
     special caching explicitly enabled per remote and confined to the proxy cache. We could distinguish normal
     freshness checks from how long cached content remains usable during an outage, without changing hosted-file
     deletion or replacement rules.

  3. Introduce a dedicated GitHub repository format. Make this a fuller integration with GitHub-specific caching,
     permitted paths, authentication guidance and setup defaults. This would provide a clearer boundary, but is a
     larger change and would still need to address replaceable release assets and access to token-backed content.

brandonrc:
Going to look at this today.

brandonrc:
Thanks for laying out the options — and your gut feel was right. **Option 3, with one adjustment**, and it dissolves the blocker rather than working around it.

**Why the dedicated format is the clean answer.** The blocker in my review was that `classify()` is not only a cache oracle: it is also consulted by the write guards at `repositories.rs:8983` (delete), `handlers/mod.rs:78` (replace) and `artifact_service.rs:671`. But look at its signature:

```rust
pub fn classify(format: &RepositoryFormat, path: &str) -> Mutability
```

It already takes the **format**. So a rule that fires only for a new format cannot touch a plain `Generic` repository — hosted or remote — and the delete/replace semantics you were accidentally changing stay exactly as they are. The blocker was never the immutability idea; it was that the rule keyed on path shape alone and therefore applied to every generic repo in existence.

**The shape I would build.** This codebase already models "a format that is really another format underneath" — `poetry`, `conda` and `jupyter` are distinct `RepositoryFormat` values whose `handler_key()` returns `"pypi"`, so they route to the PyPI handler while getting their own name, dropdown entry, docs and classifier arm (`cache_classifier.rs:181`). Do the same here:

- one backend format — I would call it `github`, because what is being mirrored is GitHub release assets, not a tool — with `handler_key() -> "generic"`, so it is served by `GenericHandler` with no new handler code;
- `mise` and `aqua` as further aliases of the same handler, so users find the name they searched for, exactly as `poetry` finds PyPI;
- the immutability arm scoped to those formats in `classify()`, leaving `Generic` on the default-format family it is in today.

That is the "partial generic" answer: new format identity, no new protocol.

**Two things the format does not solve, and they still need answers:**

1. **GitHub release assets are replaceable.** A maintainer can delete and re-upload an asset under the same tag and filename. "Immutable" as a cache key with no digest will serve the old bytes forever. Either put the upstream digest (or ETag) in the cache key, or make this a very long *finite* lifetime with revalidation on miss rather than true immutability. I would take the second — it gets you the outage behaviour you actually want without claiming something GitHub does not guarantee.
2. **The documented recipe is a confused deputy.** `docs/mise-aqua.md:28-50` creates `is_public: true` remotes to `github.com` *and* `api.github.com`, and `:91-100` suggests a PAT on the API one. Generic remotes forward any path with no allow-list, so an anonymous caller can pull `/general/github-api/repos/<private-org>/<private-repo>` using your token, and the URL-keyed cache then serves it to everyone. This one must change regardless of which option we pick: scope the remote to a path prefix, or document it as an authenticated repository only.

**Milestone: 1.11.0.** Not a comment on the work — the plumbing is genuinely cheap, we added the `jupyter` alias yesterday and it was two small PRs. It is that 1.10.0 is due 2 October with 100+ open items, and the open questions above are semantic rather than mechanical. I would rather ship this with the digest/TTL decision made deliberately than rushed. I have moved the PR to 1.11.0.

If you would like to keep going, the order I would suggest: the docs security fix first (it stands alone and I would take it for 1.10.0), then the format + alias + scoped classifier, then the caching semantics with whichever of digest-in-key or long-finite-TTL you prefer after reading the above. Happy to take any of those pieces off your hands if you would rather not carry all three.

brandonrc:
One correction to my own last paragraph: I suggested the docs security fix could land separately in 1.10.0, but `docs/mise-aqua.md` is added *by this PR* — it does not exist on main, so there is nothing standing alone to fix. The confused-deputy point still matters, it just travels with this PR rather than ahead of it. Sorry for the misdirection.

I have also moved the linked issue #3657 to 1.11.0 so it tracks with the PR.

gilesw:
Maybe i should hold off till 1.11.0 is released and start work after that?

brandonrc:
> Maybe i should hold off till 1.11.0 is released and start work after that?

Part of me wants to squeeze this in to 1.10.0 but right now I have about 100 open things going into 1.10.0 😬 

https://github.com/artifact-keeper/artifact-keeper/milestone/42

Right now still thinking this is a 1.11.0 milestone. 

One compromise is I could spin up all the issues/PR stubs that are needed. You work those and then if they are good I could get them into main.  If you look at how I added jupyterlab extension that is the same pattern basically. 

It requires:
artifact-keeper
artifact-keeper-web (UI Button)
artifact-keeper-test (end to end test for this registry)

Then you can do some testing locally that would help out alot.

gilesw:
Implemented the format-alias approach you suggested, retaining this PR for the backend and splitting the companion work into artifact-keeper/artifact-keeper-web#846 and artifact-keeper/artifact-keeper-test#399. Both companions are drafts and have linked tracking issues. Target remains 1.11.0.

The release cache is finite: seven days by default, with the existing repository override and revalidation behaviour. Only github/mise/aqua opt into it; Generic caching and hosted deletion/replacement remain unchanged. The docs now use private authenticated mirrors and spell out the upstream-token trust boundary. Extended API caching stays in #3658.

Local verification passed 196 focused backend tests and 22 end-to-end checks, including actual locked mise installs with empty client caches after removing a controlled upstream. The updated PR descriptions include the broader test results and baseline comparisons for existing failures.

brandonrc:
Audit done. Everything from the 09-09 review is covered, the expiry is genuinely finite (7 days, then main's revalidation handles a re-uploaded asset), cache keys still go through the traversal guard, and the new formats reuse the generic path so no new proxy surface or budget reservation is involved. Taking it the last mile on the branch; pushed on top of your four commits, unchanged apart from the rebase:

- rebased onto main: took main's #3556 cache-format block on the `/general` download route (your github/mise/aqua-only version is subsumed; your route test still passes), kept your aliases beside main's conda split in `formats/mod.rs`, kept both tests in `format_tests.rs`, and moved the #3657 bullet under Unreleased;
- migration renumbered to 234, because 218 (and 217 after the rebase) are taken on main;
- one sentence each in the CHANGELOG bullet on the two documented limits: nothing in code prevents `is_public: true` combined with an upstream token, so token-bearing mirrors must stay private, and the mise recipe embeds the registry token in the URL.

Locally the merged tree compiles, 1,538 github, cache-classifier and format tests pass against a migrated database, clippy and fmt clean. Merging once CI is green on this head. The companion web#846 and test#399 drafts can follow.

PR #3661, opened 2026-09-02
fix(proxy): treat 429/5xx conditional probes as indeterminate

### LINKED ISSUE #3571: Proxy revalidation treats a throttled or refused upstream as "content changed"

gilesw:
## Summary

Closes #3571. Credits the parallel fix in #3572 alongside #3661 in the Unreleased changelog.

A 429 or 5xx response to a stale mutable cache entry's conditional HEAD probe previously triggered a refill GET. If the GET also failed, the existing fallback served stale content after an extra upstream request. These probe statuses now use the revalidation stale-if-error path within the one-hour grace window, without advancing `expires_at`.

403 and 408 retain the refill path: a HEAD-specific refusal or request timeout does not mean GET will fail, and suppressing GET could leave mutable content stale throughout grace. 401, 404 and 410 retain their existing behavior. OCI/Docker Hub bearer authentication is unchanged because the probe does not perform a bearer exchange and its 401 still triggers refill.

The wiremock tests exercise the fetch path through `revalidate_verdict`: 429/503 within grace require zero refill GETs; 429 past grace requires one refill before the existing fallback serves stale; HEAD 403/408 followed by GET 200 must return fresh content. The pure status table excludes 403/408.

## Validation

- All nine revalidation tests passed with `AK_TESTS_REQUIRE_DB=1` against a migrated temporary PostgreSQL database.
- Status-classifier table passed.
- `cargo fmt --check` and `git diff --check` passed.
- `cargo clippy --lib --tests -- -D warnings` passed.

## API Changes

- [x] N/A - no API changes


brandonrc:
Thanks — this is the shape the #3572 review asked for: the wiremock behaviour tests through `revalidate_verdict` (429 / 503 / 403 within grace → stale, past grace → refill) are exactly the missing coverage there, and the CHANGELOG bullet is in the right section with the right reference.

Two things before this can land:

1. **It duplicates #3572** (nicola-preda, 2026-08-26, same issue, same match-arm placement), which has a full two-reviewer audit posted at https://github.com/artifact-keeper/artifact-keeper/pull/3572#issuecomment (see the "Audit round 1" comment). I'd rather not merge two fixes for one bug. Proposal: whichever of the two lands the change in (2) first is the one that merges, and the CHANGELOG bullet credits both PRs. @nicola-preda — your call whether you'd rather rebase #3572 onto this test harness or let this one carry it.

2. **Please drop 403 (and I'd drop 408 too) from the indeterminate set.** The #3572 review lays out why: 429 and 5xx describe the upstream's *state* and are method-agnostic, but 403 is also returned per-request — a method-filtering WAF/CDN rule, a signed URL or object-store policy that grants GET but not HEAD — and on such an upstream a `Mutable` npm packument (5-min TTL) that main refreshes within 5 minutes is frozen for 65, because the refill that *would* succeed is never attempted and `expires_at` is (correctly) never advanced. The revoked-credentials argument doesn't rescue it: main's refill path already serves stale on 403 via its own `stale-if-error` arm (`origin/main` `proxy_service.rs:3860-3874`), so 403-as-indeterminate buys nothing there. 408 on a *HEAD* is the same shape — the upstream timed out reading a request with no body, which says nothing about whether a GET would. Your `test_revalidate_403_serves_stale_within_grace_and_skips_refill` would flip to asserting `Refill`.

Also worth knowing from that review: Docker Hub — the example the issue leads with — isn't reached by either PR, because `check_etag_changed` does no OCI bearer exchange and exits at the pre-existing 401 arm before the new predicate is consulted. Fixed for Maven Central / npm / PyPI / Basic-auth'd upstreams; OCI is unchanged. Fine to leave, but the bullet shouldn't imply otherwise.

brandonrc:
This needs an apology and a correction, both mine.

While this PR and #3572 were open on #3571, I ran a fix wave over the remaining 1.9.0 issues without checking which of them already had contributor PRs in flight, and merged my own fix (#3747, commit `429192ba`). That alone was a process failure. Worse: the predicate I merged treats **429, 403, 408 and 5xx** as indeterminate — the exact set I asked you here to narrow, for reasons I still think are right.

Your review comment above has the argument: 403 is returned per request, so a method-filtering WAF or CDN rule, a signed URL, or an object-store policy that grants GET but not HEAD freezes a five-minute packument for the whole grace window, and it buys nothing because the refill path already degrades to stale on a failed GET. My audit of #3747 checked only that no *public* upstream (Maven Central, PyPI, npm, crates.io) answers 403 to HEAD and 200 to GET. That does not answer your case, which is about upstreams behind a customer's own edge.

So the narrowing you were asked for is going in: a PR is open now to reduce the set to 429 and 5xx, flip the 403 test to assert a refill, add the 408 twin, and pin the case the broad predicate broke (probe 403, refill 200, new bytes served). It credits this review and #3572 by number.

Closing this as superseded by #3747-as-corrected, with the reasoning credit yours. Your point about Docker Hub never reaching the predicate — `check_etag_changed` does no OCI bearer exchange and exits at the 401 arm — is also in the merged bullet, so the entry does not overclaim. If you want the follow-up (#3749: a 410 on the probe still serves the cached object forever instead of negative-caching like 404), it is unowned and it is adjacent to what you have already read closely.

brandonrc:
Follow-up as promised: #3756 is open and narrows the merged set to 429 and 5xx only. 403 and 408 go back to the refill path, with a test that pins the case the broad predicate broke — probe 403, refill 200, new bytes served and the sidecar validator updated — and one that keeps the degradation, probe 403 with refill 403 serving stale. The reasoning is credited to this review by number in the commit and the PR body.

Issue #758, opened 2026-04-16
Backups fail - IO error: provided value is too long when setting path


jamie-kemp:
### Description

Attempted full and metadata backups. Both progress further now the download_stats issue under #736 has been resolved. But now fail grumbling about providing path being too long.

<img width="1626" height="163" alt="Image" src="https://github.com/user-attachments/assets/f7deebc3-9a73-4779-bf7a-315bafd4f40d" />

### Environment Information

- Deployment type: docker

### Backend Version

dev

### Web UI Version

latest (Web 1.1.0-rc.8 (247b2f2))

### Package Format

None

### Expected Behavior

The backups should execute successfully

### Observed Behavior

The backup fails

### Steps to Reproduce

1. Create a backup
2. Execute it

### Logs

```shell
artifact-keeper-backend  | 2026-04-16T08:48:12.040017Z DEBUG http_request{method=GET uri=/api/v1/admin/backups?per_page=100}: tower_http::trace::on_request: started processing request
artifact-keeper-backend  | 2026-04-16T08:48:12.041461Z DEBUG http_request{method=GET uri=/api/v1/admin/backups?per_page=100}:http_request{correlation_id=125b204d-1241-4a87-8d8b-961a5c2de75b method=GET uri=/api/v1/admin/backups}: sqlx::query: summary="SELECT id, backup_type as …" db.statement="\n\n\n            SELECT\n                id, backup_type as \"backup_type: BackupType\",\n                status as \"status: BackupStatus\",\n                storage_path, size_bytes, artifact_count,\n                started_at, completed_at, error_message,\n                metadata, created_by, created_at\n            FROM backups\n            WHERE ($1::backup_status IS NULL OR status = $1)\n              AND ($2::backup_type IS NULL OR backup_type = $2)\n            ORDER BY created_at DESC\n            OFFSET $3\n            LIMIT $4\n            \n" rows_affected=0 rows_returned=0 elapsed=489.586µs elapsed_secs=0.000489586
[...]

PR #3662, opened 2026-09-02
fix(security): bound GNU LongName/LongLink/PAX records in scanner tar extraction

### LINKED ISSUE #3528: Scanner tar extraction: GNU LongName/LongLink/PAX records bypass unpack_tar_limited's byte budget entirely
Found while fixing the same defect in `BackupService::extract_entries` (#3373 / PR #3526). `scanner_service.rs` hand-rolls its own bounded tar walk and has the identical escape; it uses none of `util::bounded_archive` (grep for it in that file returns nothing).

## The defect

`unpack_tar_limited` (`backend/src/services/scanner_service.rs:1717`) tracks `remaining`/`max_bytes` and `entries_seen`/`max_entries` **inside the `for entry in entries` loop**, i.e. after the iterator has already yielded. But `tar` consumes GNU LongName (`L`), GNU LongLink (`K`) and PAX (`x`) extension records *inside* `entries().next()`, via `EntryFields::read_all()`:

```rust
// tar-0.4.46/src/entry.rs:297, called from src/archive.rs:418, 429, 440
pub fn read_all(&mut self) -> io::Result<Vec<u8>> {
    let cap = cmp::min(self.size, 128 * 1024);   // caps only the PREALLOCATION
    let mut v = Vec::with_capacity(cap as usize);
    self.read_to_end(&mut v).map(|_| v)          // then reads the FULL declared size
}
```

So a record declaring gigabytes is fully decompressed and allocated before the entry-count check, the traversal guard, or the per-entry byte accounting run. None of them ever sees those bytes.

Its doc comment at `scanner_service.rs:1699-1701` asserts the invariant that is violated:

> cumulative-byte, per-entry, and entry-count ceilings are enforced *during* extraction and a decompression/tar bomb aborts mid-stream

The `_limited` seam it advertises for testing tiny caps does not help, because the bytes are consumed above the layer the caps live in.

## Reproduction

Live-reproduced at `max_bytes=1024, max_entries=4` with a GNU LongName record: RSS growth **> 400 MiB**, and the call returns `Ok`.

The same shape measured against the (identical) `backup_service` version, through its real entry point at its default limits: a 1,043,854-byte archive declaring a 1 GiB long name returned `Ok` at **+2118 MiB** VmHWM. After wrapping the reader it is refused at **+132 MiB**. The ratio is set by what the record declares, not by the archive's size.

## Reachability — CONFIRMED

Authenticated non-admin with **repository write**, asynchronously, on push. Call chain confirmed at source:

`scanner_service.rs:1009` (`prepare_pinned`) → `extract_archive` (`:1298`) → `spawn_blocking(extract_archive_blocking)` (`:1326`) → `unpack_tar_limited`.

Uploading a crafted archive to a repository the caller can write is enough; the scan runs on the server's own time.

## Reachability — OPEN QUESTION, NOT CONFIRMED

There is a synchronous inline proxy-scan path (`run_inline_proxy_scanners` / `proxy_scan_and_record`, `proxy_helpers.rs:6446`, `:6610`, `:6853`) and a `PROXY_SCAN_INLINE_BUDGET` of 30s (`scanner_service.rs:99`). If that path reaches the same extractor, the exposure would be considerably worse. **Nobody has verified this** — it is stated here as a question, not a finding. Three things to check:

1. Does the inline proxy-scan path actually reach `extract_archive` / `unpack_tar_limited`, or does it only run scanners that never unpack?
2. Is it reachable **anonymously** (a proxy pull against a public Remote repository), or does it require a credential?
3. Does the 30s inline budget actually cancel the `spawn_blocking` task, or does it only stop *waiting* for it while the blocking thread keeps inflating?

Please answer these before treating this as anything more than an authenticated-write-path issue — the severity and whether it warrants an advisory both hinge on them.

## Fix

The same one PR #3526 uses, and the pattern ~20 other extractors in this codebase already follow: wrap the **decoded** stream in the shared budget *before* `tar::Archive::new`, so the budget bounds `read_all()` too.

```rust
// util/bounded_archive.rs:628 / :666
let mut archive = tar::Archive::new(bounded_archive::budgeted_to(decoder, max_bytes));
```

A budget breach surfaces as an `io::Error` during the walk; `bounded_archive::is_decompression_budget_breach` (added in #3526) tells it apart from a genuine format error so the caller can keep its own message. The in-loop `remaining` accounting can then go — the reader does it, and does it for headers, padding and extension records as well.

The doc comment at `:1699-1701` needs correcting either way: it currently promises a guarantee the code does not provide.

Any regression test must use an `L`/`K`/`x` record, not a plain oversized member — a plain member was already bounded, which is exactly why this survived. See `extract_entries_bounds_tar_extension_records` in PR #3526 for a fixture that builds all three from raw ustar headers.

brandonrc:
Closes #3528

## Summary

`unpack_tar_limited` (`backend/src/services/scanner_service.rs`) tracked its cumulative-byte, per-entry and entry-count ceilings **inside** the `for entry in entries` loop — after the tar iterator had already yielded. `tar` consumes GNU LongName (`L`), GNU LongLink (`K`) and PAX (`x`) extension records *inside* `entries().next()`, via `EntryFields::read_all`, which caps only its 128 KiB preallocation and then reads the full size the record declares. A record declaring gigabytes was inflated and allocated before the entry-count check, the traversal guard or `copy_entry_bounded` ran; none of them ever saw those bytes and the call returned `Ok`.

The byte budget now wraps the **decoded stream before `tar::Archive::new`**, via `util::bounded_archive::budgeted_to` — the same `BudgetReader` #3526 put under `BackupService::extract_entries` for the identical defect, and that the other ~20 extractors in this codebase already use. A breach surfaces as an `io::Error` during the walk and is told apart from a genuine format error with `bounded_archive::is_decompression_budget_breach`, so the extractor keeps its own "suspected decompression bomb" message. The per-entry accounting is kept as defence in depth. `unpack_tar`/`unpack_tar_limited` now take the reader rather than a pre-built `tar::Archive` — that is the only way the budget can sit underneath the tar reader.

Limits and their environment overrides are unchanged (`MAX_SCAN_EXTRACTED_BYTES`, 2 GiB default; `MAX_SCAN_EXTRACTED_ENTRIES` = 200,000). The doc comment that promised the guarantee the code did not provide is corrected.

## The issue's three open questions

The issue asked these be answered before treating this as more than an authenticated-write-path issue. Answered at source, read-only — **no behaviour in that path is changed by this PR**:

1. **Does the inline proxy-scan path reach `unpack_tar_limited`? Yes.** `run_inline_proxy_scanners_target` (`scanner_service.rs:3112`) calls `scanner.scan_target(target, None, content)`, and `GrypeScanner::scan_target` (`grype_scanner.rs:1859`) calls `ScanWorkspace::prepare_pinned` (`:1882`), which is the same function that reaches `extract_archive` → `unpack_tar_limited`. The inline path is not extraction-free.
2. **Anonymously reachable? Not established here.** The bytes on that path come from *upstream*, not from the caller's request body, so the crafted archive has to be published to (or served by) the upstream the Remote repository proxies — a different precondition from the confirmed write-path one. I did not trace the anonymous-pull authorization surface far enough to make a claim either way, and did not want to put a guess in an advisory.
3. **Does the 30s inline budget stop the inflation? No.** `run_scan_within_budget` (`proxy_helpers.rs:6837`) is `tokio::time::timeout` around the scan future, and `extract_archive` (`scanner_service.rs:1375`) does `tokio::task::spawn_blocking(...).await`. A dropped `JoinHandle` does not cancel a blocking task, so the timeout stops the request *waiting* while the blocking thread keeps inflating to completion. `PROXY_SCAN_INLINE_BUDGET` bounds latency, not work.

That third point holds for the confirmed write path too, and is part of why the bound has to be on the reader rather than on a wall clock.

## Verified

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(/tar_/) or test(/unpack_/) or test(/_zip_/)'` — **111 passed, 0 failed**, including `backup_service`'s `extract_entries_bounds_tar_extension_records` and every pre-existing scanner tar/zip test.
- **Fails without the fix.** Reverting just the `budgeted_to` wrap (keeping the new signature) and re-running: `test_tar_long_path_extension_record_still_extracts_limited` PASS, `test_tar_gz_bomb_rejected_limited` PASS (the pre-existing test never covered this shape — a plain oversized member was already bounded, which is why the escape survived), and `test_tar_extension_records_bounded_limited` **FAIL** with `Failed to create tar output /tmp/.tmp2Ofv0M/aaaa…` — a filename one megabyte long, i.e. the extension record was inflated in full and applied. With the fix it is refused, and the counting reader confirms at most `BUDGET + 1` decoded bytes were ever pulled.

### Regression tests

- `test_tar_extension_records_bounded_limited` — all three record types (`L`, `K`, `x`) built from raw ustar headers, each declaring 1 MiB against a 1 KiB budget. Asserts the budget error **and**, through a `CountingReader` wrapped around the decoded stream, that no more than `BUDGET + 1` bytes were read — so it pins that the record is refused *before* it is inflated, not merely that an error came back afterwards.
- `test_tar_long_path_extension_record_still_extracts_limited` — control: a path over the 100-char ustar name field (written by `tar::Builder` as a GNU LongName record, the same shape `build_backup_tar` emits per #758) still extracts and round-trips its contents. The budget bounds these records; it must not reject them.

Not run: no database was needed — the change is a pure unit-testable extractor path, and both new tests are `#[test]` with no DB harness. I did not run the full `--lib` suite, only the tar/zip/unpack selection above plus a full `--lib --tests` type-check.

## Deliberately not included

- The `spawn_blocking` cancellation gap (answer 3) is a separate defect in the inline proxy path with its own design question, not something to fold into a bounds fix.
- No change to limits, defaults, or the zip extractor (`unpack_zip_limited` reads through `zip`, which does not have this record class; its per-entry bound is untouched).


brandonrc:
## Audit — 2 reviewers (correctness / security+integration)

**Verdict: merge once CI is green on `7e7e8d26`.** Both reviewers verified the budget now sits under `tar::Archive` on every path (three `ArchiveKind` variants, gz decoded before budgeting, zip untouched), no double-charging, error strings byte-identical, all limit constants unchanged; both independently reproduced the revert proof (a 1 MiB filename gets *applied* on main; all three pre-existing `unpack_tar_limited` tests pass on the reverted tree, which is the demonstration that the old bounds never saw these bytes). The security pass established that `budget + 1` decoded bytes is a hard ceiling for every record type (`read_all`, `skip`, sparse, PAX globals), traced the inline-proxy reachability claim, and quantified the framing-overhead trade (≤ 9.5 % of the budget at the 200k-entry cap; a refusal degrades to the raw-file scan, not a failed publish).

Landed as `7e7e8d26` per the audit: the `DEFAULT_MAX_SCAN_EXTRACTED_BYTES` doc now says what the budget measures for tar vs zip, and the CHANGELOG records the inline-proxy reachability (the 30 s budget bounds the wait, not the extraction thread), the flat 2 GiB ceiling as the real bound, and a second escape this closes that nobody had written down — the bodies of skipped entries `tar` reads-and-discards were never charged either.

Two follow-ups the reviewers found, both filed: **#3672** — a third unbudgeted tar walk in `formats/pypi.rs` reachable from `pypi::upload` (plus ten more files to verify), and the proportional-budget question (backup uses `clamp(len×100, 32 MiB, 8 GiB)`; scanner keeps a flat 2 GiB) noted on #3672 as well.

brandonrc:
Merge note for `a91c30ad`: #3663 (merged) and this PR both changed `unpack_tar_limited`'s signature in `scanner_service.rs` — #3663 threaded a per-entry `&AtomicBool` cancellation flag through `tar::Archive`-taking functions, this PR changed them to take the raw reader so the budget can sit underneath. The resolution keeps both: `unpack_tar<R: Read>(reader, dst, cancel)` and `unpack_tar_limited(reader, dst, max_bytes, max_entries, cancel)` build the archive over `budgeted_to(reader, max_bytes)` and call `check_extraction_cancelled(cancel)?` before each entry (the walk error mapping stays `walk_err`, so the budget-breach message is unchanged). Seven conflict blocks, all signature/call-site; the merged function body needed no logic changes. Verified on the merged tree: `cargo check`, `clippy -D warnings`, and 137/137 tar + workspace + extraction tests, including both PRs' revert-proof tests.

PR #3663, opened 2026-09-02
fix(scanner): remove the scan workspace when a scan is cancelled

### LINKED ISSUE #3565: A scan that exceeds its wall-clock budget leaks its workspace directory permanently

brandonrc:
Closes #3565

## Summary

A scan cancelled by its wall-clock budget never removed its workspace directory, and nothing else ever would.

Workspace cleanup was an explicit async call placed *after* the scan in each filesystem scanner (`grype_scanner`, `trivy_fs_scanner`, `openscap_scanner`, `incus_scanner` ×2). The inline proxy scan path races that whole future against a budget (`run_scan_within_budget` → `tokio::time::timeout`), and when the budget expires the future is dropped where it stands — so control never reached the cleanup line and the staged workspace (artifact bytes, BOM sidecar, and for archive formats the entire extracted tree) stayed on the scan-workspace volume permanently. The janitor reaps stale `scans` rows in Postgres, not files on disk, and each timed-out attempt stages a *fresh* directory rather than reusing one, so a CPU/IO/memory-starved instance fills a fixed-size volume attempt by attempt.

## What changed

- **`WorkspaceGuard`** (`scanner_service.rs`) owns the workspace `PathBuf`, `Deref`s to `Path`, and removes the directory when dropped. Cleanup is therefore on every exit path — `Ok`, `Err`, `?`, timeout-drop, panic-unwind — rather than only the ones a caller remembers.
- **Armed as soon as the directory exists**, inside `ScanWorkspace::prepare_pinned` / `OpenScapScanner::prepare_workspace` / `IncusScanner::prepare_workspace` / `GrypeScanner::prepare_local_oci_layout`, so a timeout landing during staging, extraction or blob-copy is covered (this also closes a pre-existing leak when `write_component_pin` hard-errors).
- **`cleanup()` disarms only after removal returns.** `cleanup_path` awaits a `chmod` child and then `remove_dir_all`; both are cancellation points, so clearing the flag first left a cancelled cleanup with nothing behind it — the reported bug, reached through the fix's own code.
- **`fail_scan` takes the guard.** `fail_scan` / `fail_scan_path` are collapsed into one `fail_scan(label, artifact, error, Option<&mut WorkspaceGuard>)` that cleans through the guard and disarms it, so no exit removes a workspace twice. `None` is grype's `registry:` mode, which stages no workspace of its own (it previously asked for the removal of a directory that path never creates).
- **The blocking extractor is now cancellable.** `spawn_blocking` tasks survive the dropped future and both unpackers `create_dir_all` each entry's parent, so the extractor rebuilt the tree behind the guard's removal. The guard owns an `AtomicBool` that `Drop` sets, checked once per entry in `unpack_tar_limited` / `unpack_zip_limited` before the entry touches disk. Measured on a 20,000-entry `.tar.gz`: **19,513 files rebuilt without the flag, 0 with it.** The removal also makes a second pass, but nothing synchronises it with the extractor, so the honest guarantee is *at most one entry can land after removal* — the flag is what does the work, the second pass just makes that unlikely.
- **Sibling paths are guarded too.** Grype's CycloneDX BOM sidecar (`<workspace>.grype-bom.json` — it must stay outside the scanned tree or the next catalog includes our own output) is registered on the guard, and the local OCI layout staging directory (`<base>/grype-oci/<artifact.id>`, image-sized, reachable under the 120 s OCI budget) is a guard of its own.
- **`Drop` hands removal to a dedicated OS thread** rather than unlinking a 2 GiB / 200k-entry tree on the runtime worker unwinding the timed-out request. Deliberately `std::thread::spawn` and not `spawn_blocking`: a runtime handle is still current while the runtime is *shutting down*, and `spawn_blocking` then accepts the task and discards it unrun — which is exactly when guards are dropped en masse (SIGTERM drops the runtime, dropping every in-flight scan task), so that would leak one workspace per in-flight scan per restart. A thread also reports failure instead of panicking inside `Drop`, where a panic during an unwind aborts the process.
- **The `u+rwX` pre-pass widens each directory *before* reading it.** `walkdir` calls `read_dir` on a directory before yielding its entry, so on a `d--x--S---` directory the read has already failed `EACCES` by the time the mode could be fixed and nothing nested beneath it is ever reached. An explicit iterative walk (symlink-safe via `symlink_metadata`, never followed) gives the drop path the same ordering `chmod -R u+rwX` has, so a hostile directory nested inside another one no longer defeats cleanup — reachable through the Incus scanner, which shells out to `tar` over a container rootfs and preserves its modes.
- `IncusScanner::cleanup_workspace` and `ScanWorkspace::cleanup(base, prefix, artifact)` became unused in non-test builds and were removed; their callers use `ScanWorkspace::cleanup_path` / the guard, which is what they did.

## Still open

Grype's `registry:` pull mode has no scan workspace at all and writes its BOM under `<scan-workspace>/grype-bom/<uuid>.json` (`bom_sidecar_path`), still unlinked only when that scan returns — a small file per cancelled registry-mode attempt. Also noted: `create_dir_all` completing and the guard being constructed are two statements, so a cancellation landing exactly between them leaks one empty directory.

## Verified

- `cargo check -p artifact-keeper-backend --lib --tests` — clean, zero warnings.
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(/^services::(scanner_service|incus_scanner|openscap_scanner|grype_scanner|trivy_fs_scanner)::/)'` — **736 passed**, 0 failed. All of the above re-run after merging `origin/main` (#3664/#3665/#3666), which touches no file this PR does.
- Revert proofs, each mutation isolated and reverted afterwards:
  - `Drop` short-circuited → `workspace_is_removed_when_the_scan_future_exceeds_its_budget` FAILS (`must not leak /tmp/.tmp1oGAbW/ws-base/c1c69d96-…`).
  - `armed = false` moved back before the await → `workspace_is_removed_when_cleanup_itself_is_cancelled` FAILS.
  - cancellation check disabled → `workspace_stays_removed_when_a_cancelled_extraction_is_still_unpacking` FAILS with `19513 files left at …`.
  - `Handle::try_current()` + `spawn_blocking` restored → `workspace_is_removed_when_the_guard_drops_during_runtime_shutdown` FAILS.
  - `walkdir`-driven pre-pass restored → `workspace_drop_removes_a_nested_mode_hostile_tree` FAILS.

Not run: nothing DB-backed is involved, so no database harness was needed; no live scanner engine (grype/trivy/openscap/incus) was exercised — the tests drive `ScanWorkspace::prepare` and stub scan bodies.

## Tests (`services::scanner_service::tests`)

- `workspace_is_removed_when_the_scan_future_exceeds_its_budget` — stages a workspace, wraps a scan body that sleeps past a 20 ms budget in `tokio::time::timeout`, asserts the directory goes away.
- `workspace_is_removed_when_cleanup_itself_is_cancelled` — polls `cleanup()` exactly once (`now_or_never`) and drops it, then asserts the guard still removes the workspace. No timer, so it is deterministic.
- `workspace_stays_removed_when_a_cancelled_extraction_is_still_unpacking` — a 20,000-entry `.tar.gz`, polled to mid-extraction and then dropped exactly as `timeout` would, asserts the tree is gone and stays gone two seconds later.
- `workspace_cleanup_removes_the_directory_exactly_once` — the completion path removes it, and a disarmed guard does not delete a directory recreated at that path.
- `workspace_is_removed_when_the_guard_drops_during_runtime_shutdown` — parks a task holding an armed guard, drops the runtime under it, asserts the workspace goes away.
- `workspace_drop_removes_a_nested_mode_hostile_tree` — `<ws>/a/b/f` with both directories at `0o2100`, asserts the drop path still removes it.


brandonrc:
## Audit — rounds 1 and 2 (correctness / integration, then a focused re-audit of `23be8477`)

**Round 1 (head `8483e09c`): NEEDS-WORK.** The RAII guard was sound — path always `base/<uuid>`, never user-derived, symlink refused by `O_NOFOLLOW` `remove_dir_all`, `NotFound` ignored — but five gaps: `cleanup()` disarmed *before* its awaits (a budget expiring inside cleanup re-opened the leak; reproduced); `fail_scan` never disarmed; the BOM sidecar and the `grype-oci` staging dir live outside the workspace; a sync `remove_dir_all` of up to 200k entries on a tokio worker; and the big one — after `Drop` removed the tree, the still-running `spawn_blocking` extractor **recreated 99.3 % of it** (39,713 files twenty seconds after a 20 ms budget).

**Round 2 (head `23be8477`): MERGE-WITH-NITS.** All five confirmed fixed by counterfactual: disarm moved after the await; `fail_scan` takes the guard; `own_sidecar` + a guarded `prepare_local_oci_layout`; `Drop` hands off to a blocking thread; an `Arc<AtomicBool>` cancellation flag checked before every `create_dir_all`/`File::create` per entry — the 20k-entry test passed 34/34 (10 serial + 24 under 8-way parallel load) and the counterfactual leaves 19,852 of 20,000 entries. Two residuals being fixed in a last pass: `spawn_blocking` silently no-ops during runtime *shutdown* (a handle is still current, so the inline fallback never fires — `std::thread::spawn` instead), and the `walkdir` chmod pre-pass can't descend into a `d--x--S---` directory because walkdir reads it before yielding it (the Incus `tar` case the pre-pass exists for).

Left open and stated in the CHANGELOG: grype's `registry:`-mode BOM (no workspace to hang a guard on) and the one-statement window between `create_dir_all` and guard construction (one empty inode per attempt).

brandonrc:
Final pass (`a51ae8c8` + `0608e32a`): `Drop` now removes via a named `std::thread` so a runtime-shutdown drop still cleans up (new test parks a guard, drops the runtime, asserts removal — fails with `spawn_blocking` restored); the chmod pre-pass is an explicit iterative walk that widens each directory *before* reading it (new test: nested `0o2100` dirs — fails with the walkdir version restored); wording corrected to "at most one entry can land after removal". 736/736 scanner tests on the merged tree. Merging.

PR #3664, opened 2026-09-02
fix(security): scope the upstream bearer-token cache to the credential

### LINKED ISSUE #3606: Upstream bearer-token cache is not repository-scoped: one repo's privileged token is reusable by another
## Summary

The upstream OCI bearer-token cache is keyed on `realm + service + scope` with **no repository identity and no credential identity**. A remote repository that holds upstream credentials mints a privileged token; a *different* remote repository pointed at the same registry — with no credentials of its own — can then get a cache hit on that token and pull with it.

## Detail

`backend/src/services/proxy_service.rs`, `obtain_bearer_token`:

```rust
async fn obtain_bearer_token(
    &self,
    realm: &str,
    service: &str,
    scope: &str,
    upstream_auth: &Option<crate::services::upstream_auth::UpstreamAuthType>,
    client: &Client,
) -> Result<String> {
    let cache_key = format!("{}\0{}\0{}", realm, service, scope);

    if let Some(token) = self.get_cached_token(&cache_key).await {
        return Ok(token);
    }
```

`upstream_auth` is a parameter of this function and is **not** part of the key. Neither is the repository.

The cache is process-wide, not per-request or per-repository:

- `UpstreamClient` owns `token_cache: RwLock<HashMap<String, (String, Instant, u64)>>` (`proxy_service.rs:2221`, initialised at `:2278`).
- `ProxyService` holds exactly one `UpstreamClient`, constructed once (`proxy_service.rs:3118`, `:3179`, `:3193`), and `ProxyService` is a singleton in application state.

So the map is shared across every remote repository in the deployment for the lifetime of the process.

## The sequence

1. Remote repo **A** targets a private registry and has stored upstream credentials. A pull triggers a 401 Bearer challenge; AK exchanges A's credentials for a token and caches it under `realm\0service\0scope`.
2. Remote repo **B** — a different repository, potentially owned by a different tenant — targets the same registry with **no** credentials configured.
3. A request through B for the same image produces the same `realm`, `service` and `scope`, so the key matches.
4. `get_cached_token` returns **A's** token, and B pulls the private image with A's credential.

The token is reused until it expires out of the cache.

## Scope

Independent of any configuration flag. It needs only two remote repositories pointing at the same upstream registry where at least one has credentials — which is an ordinary multi-team setup, and is also the natural shape when one team mirrors a private registry and another mirrors its public subset.

The credential itself is not disclosed; the *access it grants* is. Anyone able to create or use a second remote repository against the same registry can read what the credentialed repository can read.

## Suggested fix

Include the repository in the key:

```rust
let cache_key = format!("{}\0{}\0{}\0{}", repo_id, realm, service, scope);
```

`repo_id` is already in scope at all three callers of the enclosing `exchange_bearer_then` (the buffered fetch, the streaming fetch, and the conditional/ETag refetch); it needs threading through the signature. The cost is a lower cache hit rate across repositories that share an upstream, which is the correct trade — those are exactly the entries that must not be shared.

Consider additionally keying on a hash of the credential identity, so that rotating a repository's credentials does not serve a token minted under the old one.

## Verification status

**Confirmed by reading code**: the key construction, the absence of `repo_id` and `upstream_auth` from it, the single process-wide `token_cache`, and the single shared `UpstreamClient`. Quoted above with locations.

**Not yet reproduced at runtime.** A live proof is straightforward and worth doing before the fix lands, so the regression test has a known-failing starting point: two remote repositories against one registry requiring auth, credentials on the first only, pull through the first, then pull the same image through the second and observe whether it succeeds and whether a second token request reaches the auth endpoint.

## Test note

A regression test should assert that two repositories with the same `realm`/`service`/`scope` but different credentials each cause a **separate** exchange against the token endpoint, rather than the second reusing the first's token. Note that a test asserting only "different keys produce different cache entries" is near-tautological — the assertion has to be on the observed upstream token requests.

Found while scoping #3591 (cross-origin bearer realms). Filed separately: it is present today regardless of that feature, and #3591 would make this path the sanctioned route for private images, turning it from an edge case into the main flow.

brandonrc:
## Summary

The process-wide OCI bearer-token cache on `UpstreamClient` was keyed on `realm \0 service \0 scope` alone. `ProxyService` is a singleton holding exactly one `UpstreamClient`, so that map is shared by every Remote repository in the deployment: a repository with upstream credentials minted a privileged token, and a *different* Remote repository pointed at the same registry with **no** credentials produced the identical key on its next pull and was handed that token. It then pulled with the credentialed repository's access until the entry aged out.

The cache key now additionally carries a SHA-256 digest of the credential the token was minted under (`UpstreamClient::token_cache_key` / `credential_digest`). Anonymous (`None`) hashes to its own bucket, and each distinct `UpstreamAuthType` to its own.

## Why credential identity rather than repository id

The issue suggests keying on `repo_id`. Keying on the *credential* is strictly stronger and cheaper:

- A token is a function of the credential exchanged for it, so two repositories presenting the **same** credential are entitled to the same token and may keep sharing the entry — `repo_id` would lose those hits for no security gain.
- `repo_id` alone would still serve a token minted under a credential that has since been **rotated** — the follow-on the issue itself asks to consider. The digest closes that in the same key.
- The security property the issue reports is "different credential ⇒ different token", and that is exactly what the digest expresses.

Only the digest is cached, never the secret; the fields are length-prefixed and each auth type domain-tagged, so no username/password split can collide with another. `None` also covers the cross-origin realm that GHSA-78h6-3wp8-2542's origin pinning strips credentials from, so that path can no longer land on a credentialed entry either.

**Eviction bound:** unchanged in kind. Entries past their TTL are still evicted on every write (`cache.retain`) and the TTL is still capped at `MAX_TOKEN_TTL_SECS` (1 h). Distinct credentials multiply live entries by the number of configured upstream credentials, which is bounded by the number of Remote repositories.

## Tests

- `test_obtain_bearer_token_does_not_reuse_credentialed_token_anonymously` — the behavioural regression test the issue asks for. A wiremock token endpoint mints `privileged-token` when Basic credentials are present and `public-token` when they are not. Repo A exchanges with credentials, repo B exchanges anonymously against the same realm/service/scope. Asserts B gets a different token **and** that the token endpoint saw **two** requests, then that A's repeat pull is still a cache hit (still 2 requests) — the fix scopes the entry, it does not disable caching.
- `test_token_cache_key_separates_credentials` — pure unit test on the key derivation: anonymous ≠ credentialed, different credentials ≠ each other, rotated password ≠ old, Basic ≠ Bearer with identical secret material, same credential ⇒ same key, length-prefix disambiguation (`ab`/`c` vs `a`/`bc`), and the raw secret never appears in the key.

Three pre-existing tests that hand-built the old key string were updated to call `UpstreamClient::token_cache_key`.

## Verification

All commands via the repo's compile lock, `SQLX_OFFLINE=true`, no database (these tests need none).

- `cargo fmt` — applied.
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(token_cache) or test(obtain_bearer_token) or test(get_cached_token) or test(realm_matches_upstream_origin)'` — **27 passed, 0 failed**.
- **Fails without the fix**: with only the production key line reverted to `format!("{}\0{}\0{}", realm, service, scope)` (tests untouched), `test_obtain_bearer_token_does_not_reuse_credentialed_token_anonymously` fails:

  ```
  assertion `left != right` failed: a repository with no upstream credentials
  must not be served the token minted from another repository's credentials (#3606)
    left: "privileged-token"
   right: "privileged-token"
  ```

  i.e. the anonymous repository was handed the credentialed repository's token, and the token endpoint saw one request instead of two. The fix was restored and the suite re-run green.

Not run: the full backend suite and any DB-backed or e2e tests — unchanged by this diff and gated on a database this change does not touch.

Closes #3606


brandonrc:
## Audit — 2 reviewers (security / correctness+integration)

**Verdict: merge once CI is green on `e1150e49`.** Key derivation verified: the two `UpstreamAuthType` variants are exhaustively matched, fields are length-prefixed with a length-prefixed domain tag (so `ab`/`c` ≠ `a`/`bc` and Basic ≠ Bearer on identical bytes), anonymous is its own bucket, the secret never reaches the key/log/metric, and the hex digest's fixed width makes the key unambiguous whatever an upstream puts in realm/service/scope — which `HeaderValue` already prevents from carrying a NUL. Unkeyed SHA-256 is acceptable for a process-memory key (the API-token cache already does the same); a process-random HMAC would be a separate ticket covering both. Exactly one production caller; lookup and insert share the derived key; eviction/TTL bound unchanged; the no-`repo_id` decision holds (rotated credential → miss, different credential → miss). Revert proof reproduced by both reviewers — and the pure-key test passes under the revert, confirming the wiremock test is the real pin.

Landed as `e1150e49`: `test_cache_key_includes_service` was the fourth test still hand-building the retired 3-field key and now calls `token_cache_key`; the `Bearer` arm gets a doc note that it is byte-identical on the wire to anonymous (only `Basic` is forwarded) and is kept separate on purpose; `main` merged with a single `### Security` heading.

Issue #3623, opened 2026-09-01
handlers::db_err leaks raw Postgres error strings to unauthenticated responses


brandonrc:
## Summary

`api::handlers::db_err` (backend/src/api/handlers/mod.rs) returns the **raw**
database error string in the 500 response body:

```rust
pub fn db_err(e: impl std::fmt::Display) -> axum::response::Response {
    let text = e.to_string();
    if crate::error::is_pool_timeout(&text) { return error_helpers::map_db_err(text); }
    ( StatusCode::INTERNAL_SERVER_ERROR, format!("Database error: {}", text) ).into_response()
}
```

Every handler that maps DB errors via `.map_err(super::db_err)?` (rpm, ansible,
and others) therefore surfaces the underlying driver message to the client —
unauthenticated, on public repositories. Observed body:

```
Database error: error returned from database: invalid byte sequence for encoding "UTF8": 0x00
```

This is a coarser leak than the generic path, which correctly maps
`AppError::Database` to `{"code":"DATABASE_ERROR","message":"Database operation
failed"}` (see #3545). The raw form exposes schema/driver detail and query-shape
hints to anonymous callers.

## Impact

Information disclosure. Low-to-moderate: no data rows are returned, but the raw
Postgres/sqlx message is attacker-visible on any route using `db_err`, on any
public repository, without credentials.

## Suggested fix

Route `db_err` through the same sanitizer the rest of the API uses: return a
stable `{"code":"DATABASE_ERROR","message":"Database operation failed"}` body
(preserving the pool-timeout special case) and log the raw text server-side
only, mirroring `AppError::Database`'s `user_message()`.

Found while fixing #3545 (which addressed the input-rejection half on the
generic routes). Related: #3622.

Issue #3284, opened 2026-08-10
OCI: setup_guard and demo_guard emit REST-shaped JSON 4XX bodies on /v2


brandonrc:
Two global middleware layers sit in front of `/v2` and emit REST-shaped JSON on a 4XX. `routes.rs:146` nests `/v2` **inside** the global layer stack, so a request can be refused before it ever reaches an `oci_v2.rs` handler — and the handler surface is where #3110 / #3274 made every 4XX JSON body the distribution-spec envelope.

The spec is unambiguous that the shape is a MUST once the body is JSON (`spec.md`, "Error Codes"):

> A `4XX` response code from the registry MAY return a body in any format. If the response body is in JSON format, it MUST have the following format:
> ```json
> {"errors":[{"code":"<error identifier>","message":"<message describing condition>","detail":"<unstructured>"}]}
> ```

### 1. `setup_guard` — `backend/src/api/middleware/setup.rs:32-79`

The allowlist is `/api/v1/auth*`, the password-change route and a small `matches!` set; `/v2` is not in it. While setup is pending, **every** `/v2` request — including the `docker login` challenge probe — gets:

```json
403 {"error":"SETUP_REQUIRED","message":"Initial setup is required. …","instructions":[…]}
```

A docker client cannot render that, so a fresh unconfigured instance reports a `docker login` failure with no usable message.

### 2. `demo_guard` — `backend/src/api/middleware/demo.rs:71-77`

In demo mode, every `/v2` write verb gets:

```json
403 {"error":"Write operations are disabled in the demo. Deploy your own instance to get full access."}
```

Same shape problem, and this one is reachable on the public demo.

### Not affected

`guest_access_guard` is already correct: `is_allowlisted` covers `/v2`, `/v2/` and `/v2/<name>/manifests/<ref>`, pinned by `allowlist_oci_v2_challenge` in `middleware/guest_access.rs`. That is the pattern to follow.

### Suggested fix

Either allowlist `/v2` in both guards the way `guest_access_guard` does (letting the OCI handlers apply their own gate and return an envelope), or have each guard branch on the `/v2` prefix and emit `{"errors":[{"code":"DENIED","message":…}]}` instead. The first is closer to what the layers are for; the second keeps the refusal.

Both states are narrow (setup-pending, demo mode), which is why #3274 deliberately left them out of its diff rather than growing its blast radius — see the "known residuals" section of that PR. Filing so the enumeration is recorded rather than lost.

[...]

PR #3665, opened 2026-09-02
fix(security): reject a NUL byte in the decoded request path at one shared boundary

### LINKED ISSUE #3622: NUL byte in path/key reaches Postgres (500) on format-native handlers and the repo key — sibling of #3545
## Summary

Sibling of #3545. The NUL-byte fix in #3545 covers the four generic
repository-scoped routes (`/api/v1/repositories/{key}/{artifacts,download,versions}/*path`
and DELETE). A boundary probe shows the **same class** — a percent-encoded
`%00` in a wildcard path segment decoded to a literal `\0` and passed to
Postgres — still reaches the driver on several **format-native** handlers and
on the repository **key** segment.

## Reproduction (anonymous, public repos)

```
GET /maven/{public}/a%00b                 -> 500
GET /rpm/{public}/packages/a%00b          -> 500   (also leaks raw driver text, see #3623)
GET /hex/{public}/tarballs/a%00b          -> 500   (also leaks raw driver text)
GET /puppet/{public}/v3/files/a%00b       -> 500   (also leaks raw driver text)
GET /ansible/{public}/download/a%00b      -> 500   (also leaks raw driver text)
GET /api/v1/repositories/pub%00gen/artifacts/a  -> 500   (NUL in the repository KEY, via get_by_key)
```

Already safe (no change needed): `/generic/{key}/*path` (404), `/pub/{key}/packages`
(400), `/cocoapods/{key}/Specs` (400), `/rubygems/{key}/gems` (404),
`/rpm/{key}/repodata` (404), `/npm/{key}/-/*` (501), `/nuget/{key}/v2` (200).

## Suggested fix

Same treatment as #3545 (reject `\0` with a 400 at path extraction). Given the
number of wildcard-path handlers, a shared boundary — a small path extractor or
a middleware that rejects control bytes in the decoded path/key before dispatch —
would be more robust than per-handler checks. The repository-key case belongs at
`get_by_key`/key extraction.

Found while fixing #3545; splitting out so #3545 stays a minimal, reviewable diff.

brandonrc:
## Summary

A percent-encoded NUL (`%00`) in a URL path is decoded by axum into a literal `\0`, which Postgres rejects at the wire protocol — so an anonymous request to a **public** repository cost a pool checkout plus a failing query and answered `500 DATABASE_ERROR`. #3545 closed the four generic repository-scoped routes at their own path extraction; this closes the rest of the surface named in #3622:

- format-native wildcard routes — `/maven/{key}/*path`, `/rpm/{key}/packages/*path`, `/hex/{key}/tarballs/*path`, `/puppet/{key}/v3/files/*path`, `/ansible/{key}/download/*path`
- the repository **key** segment itself — `GET /api/v1/repositories/pub%00gen/artifacts/a` (via `get_by_key`)

With ~40 format handlers on that surface, this is the **shared boundary** the issue asks for rather than per-handler checks: a new global layer (`api::middleware::nul_path::nul_path_guard`, registered in `create_router` just inside correlation-id) percent-decodes the request path and refuses a NUL with `400 VALIDATION_ERROR` before routing — so every current and future handler is covered at once, and the refusal costs no database work.

The check reads **bytes**, never a decoded `String`: `urlencoding::decode` fails on a path whose decoded bytes are not valid UTF-8 *anywhere*, so one stray `%C0%80` would have hidden a `%00` in another segment (`/maven/pub%00key/a%C0%80b`) — and `repo_visibility_middleware`, which percent-decodes the key segment itself before axum's `Path` extractor runs, would still have bound that NUL to a query. `urlencoding::decode_binary` is infallible, so no escape elsewhere in the path can blind the guard. (Thanks to both reviewers for finding that on the first revision.)

Deliberately narrow:

- **NUL-only.** No other byte a client may legitimately percent-encode is inspected, so read routes keep serving every path an upload could have stored.
- **No behaviour change without a NUL.** Postgres cannot store `0x00` in a `text` column at all, so no persisted artifact path or repository key can contain one and no request that could ever have returned artifact content changes status. (A handful of routes #3622 lists as "already safe" do move from one refusal to another — 404/401 → 400 — which is why the claim is scoped to content.) The guard's fast path (`no '%'` → return) keeps ordinary requests allocation-free.
- A refusal on the `/v2` OCI surface is emitted as the distribution-spec error envelope rather than the REST body, matching the #3284 precedent set by `setup_guard`.

## Verified

Local, with a throwaway Postgres 16 (`ak-fix-3622`) + `sqlx migrate run`, `AK_TESTS_REQUIRE_DB=1`, `SQLX_OFFLINE=true`, everything under the compile lock:

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(nul_path)'` — **10 passed, 0 failed**. Six pure unit tests (decode-detection positives/negatives, the invalid-UTF-8 vectors, the 400 + `VALIDATION_ERROR` body, a non-NUL escaped path still reaching the handler, the OCI envelope) and four DB-backed router regressions driven through `create_router` itself:
  - `format_native_nul_paths_are_400_not_db_500_db` — anonymous, public repo, one fixture per format: `/maven/{k}/a%00b`, `/rpm/{k}/packages/a%00b`, `/hex/{k}/tarballs/a%00b`, `/puppet/{k}/v3/files/a%00b`, `/ansible/{k}/download/a%00b` all 400, each route's non-NUL control still 404.
  - `nul_in_the_repository_key_is_400_not_db_500_db` — `/api/v1/repositories/pub%00gen/artifacts/a` is a 400, while an ordinary unknown key (`pub0gen`) and a real missing artifact both still 404.
  - `mixed_invalid_utf8_escape_is_still_refused_at_the_boundary_db` — `/maven/pub%00key/a%C0%80b` and `/maven/re%00po/%FF` are 400; control `/maven/{real}/a%C3%A9b` still 404.
  - `nul_refusal_carries_correlation_id_and_is_byte_identical_db` — the 400 carries `X-Correlation-ID`, and its body is byte-identical for a real repository, an unknown one and an unrouted path.
- **Three revert proofs**, each re-run against the same database:
  1. Registration at `routes.rs:215` replaced by a no-op → `format_native_nul_paths_are_400_not_db_500_db` fails with the reported symptom (`left: 500, right: 400`, body `{"code":"DATABASE_ERROR",…}`), and so does the repository-key test.
  2. `decode_binary` reverted to `decode` (the first revision's code) → `a_non_utf8_escape_elsewhere_does_not_blind_the_guard` fails, and `mixed_invalid_utf8_escape_is_still_refused_at_the_boundary_db` fails with `left: 401, right: 400` on `/maven/pub%00key/a%C0%80b` — the reviewer's measured bypass, now pinned.
  3. The layer moved *outside* `correlation_id_middleware` → `nul_refusal_carries_correlation_id_and_is_byte_identical_db` fails on the missing header.
  All three edits reverted afterwards; tree clean.
- `cargo fmt --check` clean; `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` clean (exit 0).

Not run: the full `--workspace --lib` suite and the Tier 2 integration targets (shared compile budget on this machine); the change adds a layer and touches no existing handler, and no existing test asserts a status for a NUL-bearing path routed through `create_router`.

## Deliberately out of scope

- **Query strings and JSON bodies — tracked in #3673.** The first revision of this PR said there were no 500s there; that was wrong. `GET /api/v1/search/advanced?name=a%00b`, `?format=a%00b`, `GET /api/v1/search/suggest?prefix=a%00b` and `POST /api/v1/auth/login {"username":"a\u0000b"}` all answer `500 DATABASE_ERROR` today, anonymously, and do so identically with and without this PR — pre-existing exposure, not a regression it introduces. #3622 is scoped to the path/key, the query half wants a different answer from a path guard, and the body half needs a check at the bind site or in the driver-error mapping; #3673 carries both.
- The raw driver text those handlers leak on a 500 is #3623, not touched here.
- The inner `reject_nul_in_path` checks #3621 adds for #3545 stay worth having even though this boundary covers those four routes in production: every handler unit test in the repo builds a sub-router without the global layer, so the handlers' own contract would otherwise be untested. The two PRs are textually disjoint and merge in either order.

Closes #3622


brandonrc:
## Audit — 2 reviewers (security+bypass / correctness+integration)

**Verdict: merge once CI is green on the reworked head.** Both reviewers verified the guard closes all six #3622 vectors end-to-end through the production `create_router` against a live Postgres, that both DB tests are genuine regressions (`left: 500, right: 400` with the registration disabled), that malformed-`%` paths (`%zz`, `100%25`) keep their 404, that double-encoding and overlong UTF-8 alone don't get through, that the OCI envelope is scoped to `/v2` only, that the 400 is byte-identical for real/unknown/unrouted paths (no existence oracle), and that the middleware sits outside auth/guest/demo/setup and the rate limiters but inside security-headers/CORS/metrics.

Both found the same hole independently: `urlencoding::decode` fails on any non-UTF-8 escape, and a failed decode read as "no NUL" — measured `GET /maven/pub%00key/a%C0%80b` reaching Postgres (`invalid byte sequence for encoding "UTF8": 0x00` in the server log). Fixed in `88be5d7d` by deciding on bytes (`decode_binary(...).contains(&0)`), with the exact bypass vector pinned as a test and three revert proofs re-run. Two wrong comments about urlencoding's failure mode were corrected, and the PR body's claim that query strings had "no reported 500" was retracted — they do (`/search/advanced?name=a%00b`, `?format=`, `/search/suggest?prefix=`, `POST /auth/login`), pre-existing and now tracked in #3673.

On #3621: #3665 makes its production effect redundant (all four generic routes + DELETE are covered by the global layer), but #3621's tests build sub-routers that never carry the layer, so it stays worth keeping as inner defence-in-depth. Files are disjoint; either order merges cleanly.

Issue #2083, opened 2026-06-30
Format handlers map sqlx pool exhaustion to 500 (with leaked error text) instead of 503 across ~25 download/CRUD paths


inspired-geek:
### Description

Across the per-format handlers, sqlx connection-pool exhaustion is mapped to **HTTP 500** (often with the raw DB error text echoed into the response body) instead of **503 `POOL_EXHAUSTED` + `Retry-After`**.

This is the same root cause as #2082, but in a different layer. #2082 (and its fix) covers the shared proxy/virtual/lookup hot path (`error.rs`, `proxy_helpers.rs`, `error_helpers.rs`). The individual format handlers, however, each reinvent a database-error reply with an inline:

```rust
.map_err(|e| (StatusCode::INTERNAL_SERVER_ERROR, format!("Database error: {}", e)).into_response())?
```

or a format-specific helper (`mod.rs::db_err`, `incus::db_err`, npm `map_status`, gitlfs `lfs_error_response`, OCI `oci_error`, etc.) that hard-codes `INTERNAL_SERVER_ERROR`. None of them consult `is_pool_timeout`, so under pool saturation they return 500.

Two problems result:
1. **Wrong status under load** — clients (npm, nuget, go, docker/OCI, helm, gitlfs, …) get 500 on a transient capacity event and do not back off; operators get paged on 500s.
2. **Information leak** — the raw sqlx error text (`pool timed out while waiting for an open connection`, and for other failures table/constraint names) is echoed into the client-visible body.

Approximate inventory: ~140 sites across ~25 handler files, e.g. `gitlfs.rs` (16), `protobuf.rs` (14), `goproxy.rs` (11), `terraform.rs` (10), `swift.rs` (8), `nuget.rs` (7), `composer.rs` (7), `chef.rs` (6), `jetbrains.rs` (6), `pub_registry.rs` (5), `vscode.rs` (5), `cocoapods.rs` (5), `debian.rs` (4), `sbt.rs` (4), `alpine.rs` (4), plus `npm.rs`, `helm.rs`, `oci_v2.rs`, `huggingface.rs`, `hex.rs`, `conan.rs`, `rubygems.rs`, `incus.rs`, `mod.rs`, `promotion.rs`.

### Environment Information

- Deployment type: Kubernetes (namespace `ak`)
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL (`max_connections=200`) behind PgBouncer (transaction pooling)
[...]

PR #3666, opened 2026-09-02
fix(security): stop db_err leaking raw Postgres error strings

### LINKED ISSUE #3623: handlers::db_err leaks raw Postgres error strings to unauthenticated responses

brandonrc:
## Summary

`api::handlers::db_err` — the shared DB-error mapper that **114 call sites across 21 format handlers** reach through `.map_err(super::db_err)?` (rpm, ansible, hex, debian, helm, goproxy, cocoapods, sbt, composer, chef, conan, nuget, cran, alpine, rubygems, terraform, …) — built its 500 body as plain text:

```rust
format!("Database error: {}", text)
```

So the raw sqlx/Postgres message went back to the caller verbatim, e.g.

```
Database error: error returned from database: invalid byte sequence for encoding "UTF8": 0x00
```

Those routes serve public repositories **anonymously**, so no credentials were needed to elicit one, and the message discloses driver, encoding and query-shape detail a 500 should not. This is the coarser half of the leak #3545 closed on the generic routes, which have mapped `AppError::Database` to a sanitized envelope all along.

## Change

`db_err` now delegates to `error_helpers::map_db_err` — the same sanitizer — so these routes emit the API-wide `{"code":"DATABASE_ERROR","message":"Database operation failed"}` JSON envelope, byte-identical to the generic path, and the raw text reaches only the server-side `tracing` event `AppError::into_response` already emits (ERROR, or WARN for a pool timeout). Operators lose nothing.

The pool-timeout shed from #2083 is unchanged in behaviour and now shares the same code path: 503 + `Retry-After: 1`, `POOL_EXHAUSTED`, "Database connection pool is saturated, retry shortly". Status codes are otherwise untouched — only the body changed.

Diff is two files: the 12-line `db_err` body collapses to one line (plus doc comment), and the tests.

**Call-site audit:** nothing depended on the old body text. The only assertion on it was `test_db_err_body_includes_label_and_message`, which was pinning the leak (`assert!(text.starts_with("Database error: "))`); it is replaced. The format-specific envelopes that build their own body and print `Database error: {e}` themselves (npm, OCI, Git-LFS, protobuf, swift, promotion, pub, maven) do **not** go through `db_err` and are deliberately out of scope — those are #3622.

## Regression tests

Pure unit tests in `api::handlers::tests`, no DB needed:

- `test_db_err_does_not_leak_raw_postgres_text` — the issue's exact string. Asserts 500, body contains neither `invalid byte sequence` nor `UTF8`, and does contain `DATABASE_ERROR`.
- `test_db_err_body_is_sanitized_envelope` — replaces the test that pinned the old prefix; asserts the `code`/`message` pair and that the input text is absent.
- `test_db_err_pool_timeout_body_matches_generic_path` — new; pins 503 + `Retry-After: 1` + `POOL_EXHAUSTED` so the #2083 behaviour cannot regress silently. The pre-existing `test_db_err_pool_timeout_returns_503` is untouched and still passes.

**Proof they fail on main:** with the old `db_err` body restored and the new tests in place:

```
FAIL api::handlers::tests::test_db_err_does_not_leak_raw_postgres_text
  panicked: db_err leaked the driver message:
  Database error: error returned from database: invalid byte sequence for encoding "UTF8": 0x00
FAIL api::handlers::tests::test_db_err_body_is_sanitized_envelope
PASS api::handlers::tests::test_db_err_pool_timeout_returns_503
Summary: 7 tests run: 5 passed, 2 failed
```

## Verified

All under the repo's compile lock, `SQLX_OFFLINE=true`:

- `cargo fmt --check` — clean
- `cargo nextest run -p artifact-keeper-backend --lib --no-fail-fast` — **15999 tests run, 15999 passed, 8 skipped**
- Targeted: `test(api::handlers::tests::) or test(error_helpers::tests::) or test(/^error::tests::/)` — 73/73 passed

Not run: no Postgres container was needed (all three new tests are pure), so the DB-gated tests took their usual no-op/skip path in the full-suite run above. No integration/e2e suites were run.

Closes #3623


brandonrc:
## Audit — 1 reviewer (one-line production change)

**Verdict: merge.** Verified empirically rather than statically: `db_err(raw)` and `AppError::Database(raw).into_response()` are byte-identical in status, headers and body for both the plain-500 and the #2083 pool-timeout (503 + `Retry-After: 1` + `POOL_EXHAUSTED`) cases; the server-side ERROR event genuinely fires on this path (captured with a `tracing_subscriber` layer); the revert proof reproduces (2 of the new tests fail with main's body restored); all 114 call sites across 21 handler files inspected, none reads the body back. JSON on a 500 is fine for every native client on these routes — they already get this envelope from the generic path.

One correction landed as `393ceddc`: the bullet said "operators lose nothing", but on main the non-timeout branch logged *nothing* — these 114 sites emit a server-side ERROR for the first time, which is worth knowing during an outage. Sibling leaks that don't go through `db_err` (38 handler-local `Database error: {e}` bodies, plus `require_signing_key` and `incus::fs_err`) are tracked in #3667.

Issue #3371, opened 2026-08-14
fix(security): LDAP login error messages allow user enumeration


brandonrc:
## Summary

The LDAP login path returns distinguishable errors depending on whether the username exists in the directory, giving an unauthenticated client a user-enumeration oracle.

In `backend/src/services/ldap_service.rs` the bind failure for a nonexistent user (~line 180) and the failure for an existing user with a wrong password (~line 720) surface different error messages to the client. An attacker can iterate usernames against the login endpoint and harvest valid directory accounts for follow-on password spraying.

## Expected behavior

All LDAP authentication failures return a single generic "invalid credentials" response, with detail kept server-side in logs.

## Severity

Low — requires LDAP auth to be enabled; no direct access granted, but it materially helps targeted attacks.

Issue #969, opened 2026-04-29
auth: rate-limiter should support trusted internal CIDR allowlist


brandonrc:
## Summary

The auth rate limiter currently has no way to exempt trusted internal CIDR ranges, which causes problems for both legitimate operational scenarios and CI/E2E test infrastructure.

This was surfaced by the release-gate failure analysis: 20+ test pods running in parallel against the in-cluster backend all hammer `POST /api/v1/auth/login` with admin credentials simultaneously, exhausting the rate limit and getting `HTTP 429` followed by lockout-shaped `401`s on retry.

A short-term workaround in the test infrastructure (raising rate limits via Helm values for the test namespace) is being filed separately. This issue tracks the proper long-term fix in the backend.

## Background

- Release-gate failures across 8 consecutive runs (2026-04-24 to 2026-04-26) all traced to admin login rate-limit exhaustion under parallel test execution.
- Commit a5bfbda extended the auth_admin retry budget from 5x3s to 12x5s in `artifact-keeper-test`. Symptom relief, not a fix.
- The same problem can show up in production for high-volume internal callers (e.g. service accounts behind a reverse proxy, sidecar health probes, internal automation).

## Proposed change

Add a backend env var that accepts a comma-separated list of trusted CIDR ranges:

```
AK_RATE_LIMIT_TRUSTED_CIDRS=10.0.0.0/8,fc00::/7,127.0.0.1/32
```

Requests originating from these ranges (after honoring `X-Forwarded-For` per the existing trusted-proxy logic) bypass the rate limiter or get a much higher limit ceiling.

Open questions for design:

- [ ] Should it be full bypass or just a higher bucket? (Higher bucket is safer; full bypass risks DoS via spoofed XFF if proxy trust is misconfigured.)
- [ ] Apply only to `/api/v1/auth/login`, or all rate-limited endpoints?
- [ ] How to surface this in Helm chart values (likely `backend.rateLimit.trustedCidrs` array).

## Acceptance criteria

- [ ] Backend env var parsed and applied in the rate limiter middleware.
- [ ] Documented in `reference/environment.mdx` and the Helm chart README.
- [ ] Unit test covering CIDR matching, including IPv6.
- [ ] Existing rate-limit tests still pass for non-trusted ranges.

## Out of scope

- The temporary test-infra workaround (Helm values bump for E2E). Tracked separately.
- The complete rewrite of the rate limiter. This issue only adds a trusted-CIDR knob.

## Related

[...]

PR #3668, opened 2026-09-02
fix(security): return one generic message for every local login failure

### LINKED ISSUE #3504: fix(security): local login error messages allow user enumeration
## Summary

`POST /api/v1/auth/login` is unauthenticated and, like the LDAP endpoint fixed in #3371, `AppError::Authentication` passes its message through to the response body verbatim. Two arms of `AuthService::authenticate` (`backend/src/services/auth_service.rs`) answer differently depending on whether the submitted username exists, so the local login endpoint is also a user-enumeration oracle.

Observed live against a 1.8.1 instance:

```
POST /api/v1/auth/login {"username":"zzz-does-not-exist","password":"x"}
401 {"code":"AUTH_ERROR","message":"Invalid username or password"}

POST /api/v1/auth/login {"username":"<an existing LDAP/OIDC/SAML account>","password":"x"}
401 {"code":"AUTH_ERROR","message":"Use SSO provider to authenticate"}
```

Two distinct leaks:

1. **Federated-account disclosure** (`auth_service.rs`, the `user.auth_provider != AuthProvider::Local` arm). The row is only reached after a successful `WHERE username = $1 AND is_active = true` lookup, so the message confirms both that the account exists and which identity source it uses. Any password value produces it, so this is a single unauthenticated request per candidate username.

2. **Lockout disclosure** (the `failed_attempt_is_locked` arm). Reproduced by creating a local user and sending five wrong passwords:

```
attempt 1-4: 401 {"message":"Invalid username or password"}
attempt 5+:  401 {"message":"Account temporarily locked due to too many failed login attempts"}
```

An unknown username never produces that message (there is no row to lock), so five requests confirm existence. This one is a deliberate-looking trade-off — telling the owner why their correct password is being refused is genuinely useful — so it needs a product decision (suppress it, or keep it and accept the disclosure) rather than a mechanical fix.

## Not affected

Checked at the same time and found clean:

- The OIDC and SAML callbacks take the identity from the IdP assertion, not from a client-supplied username, so there is no username to probe.
- `POST /api/v1/auth/ci/token` keys on a provider id plus a signed CI JWT.
- There is no password-reset or forgot-password endpoint.
- `AuthService::authenticate_with_provider` / `authenticate_ldap` carry the same shape (`"LDAP user not found"`, `"OIDC authentication requires redirect flow…"`), but they have no production callers — only tests reach them. Worth deleting or wiring, but not currently reachable.

## Suggested fix

The same shape as #3371: one shared constant for the client-facing message, with the real reason logged server-side. Arm 1 is a straightforward unification. Arm 2 needs the product call above.

## Severity

Low, same as #3371 — no access is granted, but it materially cheapens targeted password spraying by turning a guessed username list into a confirmed one.

brandonrc:
## Summary

`POST /api/v1/auth/login` is unauthenticated and `AppError::Authentication` passes its message straight to the response body, so the wording of a failed login was attacker-visible. Two arms of `AuthService::authenticate` were reachable only for a username that exists **and** is active:

- a federated account answered `401 {"message":"Use SSO provider to authenticate"}` — one request per candidate, with any password, confirming both the account and its identity source;
- an account past `account_lockout_threshold` answered `401 {"message":"Account temporarily locked due to too many failed login attempts"}`, which an unknown username can never produce because there is no row to lock.

Both are the oracle #3371 closed on the LDAP endpoint; that fix's 1.8.2 changelog entry named this endpoint as still carrying it.

Every credential-level arm — unknown or inactive username, federated account, missing password hash, wrong password, locked account — now returns one shared `LOCAL_AUTH_FAILURE_MESSAGE`, with the distinguishing detail kept in a `tracing::warn!` under the `security` target, the same shape as #3371.

Closes #3504

## Design

**The timing pad is scoped to the login endpoint.** Unifying the message leaves the timing half: the arms that never reach a stored hash used to return before any bcrypt work. But `AuthService::authenticate` is a shared credential primitive with ten production callers — the API middleware tries it *before* the API-token path on every Basic-auth package request (`middleware/auth.rs:821`/`:1149`), and `oci_v2.rs:4918` and `conda.rs:1095` reach it too. In an SSO-only deployment every one of those requests takes a rejection arm, so padding inside it would put a cost-12 bcrypt in front of traffic that never reaches the login form and would halve the auth-permit headroom #1437/#1442 exist to protect.

So the pad lives behind a `TimingPad` switch on a private `authenticate_inner`:

- `AuthService::authenticate` → `TimingPad::Off`. Unchanged from main: rejection arms return without bcrypt.
- `AuthService::authenticate_for_login` → `TimingPad::On`. Used only by the `POST /api/v1/auth/login` handler.

**One `verify_password` for every arm.** The pad does not add a second call and does not discard its result. `authenticate_inner` resolves which arm this is first, then runs a single `Self::verify_password(password, hash_to_verify).await?`, where `hash_to_verify` is the stored hash or `dummy_bcrypt_hash()` — the substitution `validate_api_token` already uses. Path, error type and status are then identical by construction. This matters beyond the mean: `verify_password` sheds to `ServiceUnavailable` when the auth semaphore saturates, so an arm that swallowed its result would answer 401 for an absent username while a real account answered 503 — a one-request oracle in place of the message one.

**A per-source-IP budget for the pad — which gates the pad, not the request.** The pad hands an anonymous caller a cost-12 bcrypt per request, and the existing login limiter cannot bound that: it is keyed `login:{username}|{ip}`, which is what stops a flood against one identity from locking out others, and which also means an attacker who changes the username every request gets a fresh bucket each time. New settings, documented in `.env.example`:

- `RATE_LIMIT_LOGIN_FAILED_PER_IP_PER_WINDOW` — default **30**, `0` disables
- `RATE_LIMIT_LOGIN_FAILED_PER_IP_WINDOW_SECS` — default **300**

The login middleware peeks the budget, publishes the answer as a `LoginPadBudget` request extension, and the handler turns it into a `TimingPad`. **It never returns 429 and never refuses a login.** Within budget the login pads as normal; past it the hashless arms answer without bcrypt — the timing oracle returns for that IP until the window rolls — while any account that *has* a stored hash is still verified normally, so a correct password always authenticates. A successful login does **not** reset the bucket — being inside a spent one costs a legitimate user nothing, and resetting would void the bound below, since on a shared egress ordinary logins would continuously refill an attacker's sweep budget. The window expires on its own.

That is the honest trade, and it is stated in the CHANGELOG: holding an attacker to at most 30 padded verifies per IP per window costs the timing property for an IP that has burned its budget. Shedding instead is not an option — behind a reverse proxy that does not set `RATE_LIMIT_TRUSTED_PROXY_CIDRS` every user resolves to the proxy's own address, so one attacker would deny login to the whole deployment, which is the #1871 account-lockout DoS at IP granularity. An earlier revision of this PR did exactly that and both round-2 reviewers reproduced it.

One new `RateLimiter` primitive supports this: `peek_rate_limit`, which reports without charging, since whether to charge depends on the response the handler has not produced yet. (`record_attempt` is its companion on the outcome side.)

**Logging is selected by entry point, not by the pad.** A new `AuthEntry` discriminator carries both properties, because they diverge: an attacker who burns an IP's pad budget must still be logged. `AuthEntry::Login` always emits the `security`-target WARN naming the arm; `AuthEntry::Shared` emits nothing, exactly as `origin/main` did. That matters because the API middleware tries `authenticate` before the API-token path on every Basic-auth request, so a user running `cargo`/`pip` as `username:<api-token>`, or any request in an SSO-only deployment, would otherwise write one WARN per request and bury the login signal in the same stream at the same level.

**The audit event carries the reason the response no longer does.** `authenticate_for_login` returns `LoginFailure { error, reason }`, and the handler passes `failure.reason` into `AuthDetails::failed_login` (it was `None`). A SIEM can now separate `unknown_or_inactive_user` / `federated_account` / `no_password_hash` / `invalid_password` / `account_locked` without any of it reaching the client.

## Product decision: the lockout stays suppressed

Per the maintainer's call. A locked account now receives the generic failure. The lockout itself is unchanged — still enforced, `failed_login_attempts` and `locked_until` still written identically, and a correct password still authenticates past it and clears the counters (#1871). Only the message changed.

The cost is real and stated in the CHANGELOG: a user who has genuinely forgotten their password now sees the same message throughout and needs an administrator. **Not verified:** `artifact-keeper-web` and the SDK are separate repositories and were not checked for code that string-matches the old lockout text to render a "try again later" banner. Nothing in this repository parses it.

One arm is deliberately **not** unified, and the CHANGELOG says so rather than hiding it: a `users` row whose `password_hash` is not valid bcrypt answers **500**, not 401. That is a data-integrity fault, not a credential decision, and it is not attacker-controllable.

## What I verified

All cargo work under the compile lock with `SQLX_OFFLINE=true`.

- `cargo fmt -p artifact-keeper-backend` — applied.
- `cargo check -p artifact-keeper-backend --lib --tests` — clean, no warnings.
- `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` — exit 0.
- `bash scripts/ci/check-changelog-unreleased.sh` — pass.
- `bash scripts/ci/check-auth-event-audit.sh` — pass.
- `cargo nextest run -p artifact-keeper-backend --lib` against a throwaway Postgres 16 (`backend/migrations` applied, `AK_TESTS_REQUIRE_DB=1`) — **604 passed** across the `#3504` / login / pad / rate-limit / `config::` / `middleware::auth` / `handlers::auth` selectors.
- `origin/main` merged twice as it moved (now `0557be8f`); each time the only conflict was `CHANGELOG.md`, where both sides appended to `### Security`. Resolved to a single heading with six bullets, this one last. Zero conflict markers repo-wide.

### Fail-on-main proof

With only the production behaviour reverted (the distinct messages restored, the pad removed, the per-IP cap bypassed) and every test kept, **5 tests fail**, each on the behaviour it pins:

```
FAIL services::auth_service::tests::test_federated_account_is_indistinguishable_from_unknown_username
FAIL services::auth_service::tests::test_locked_account_is_indistinguishable_from_unknown_username
FAIL services::auth_service::tests::test_locked_account_wrong_password_reports_generic_failure
FAIL services::auth_service::tests::test_login_pads_bcrypt_for_a_rejected_arm
FAIL api::middleware::rate_limit::tests::test_failed_logins_are_capped_per_ip_across_distinct_usernames
Summary  23 tests run: 18 passed, 5 failed
```

The other new tests pass on main **by design** and are pins, not reproductions — `test_unknown_username_and_wrong_password_are_indistinguishable`, `test_missing_password_hash_is_indistinguishable_from_unknown_username` and `test_inactive_account_is_indistinguishable_from_unknown_username` cover arms that already shared the generic message; `test_shared_authenticate_does_not_pad_a_rejected_arm`, `test_successful_logins_are_not_charged_to_the_per_ip_cap` and `test_peek_rate_limit_does_not_consume_budget` pin properties the rework must not break.

`test_locked_account_wrong_password_still_reports_locked` was renamed to `..._reports_generic_failure` — it was pinning the leak — and now also asserts `locked_until IS NOT NULL` and `failed_login_attempts == threshold + 1`, so the lockout mechanics stay covered separately from the disclosure change.

The pad tests **count** bcrypt invocations through a `#[cfg(test)]` counter on `verify_password` rather than timing them, so they are deterministic instead of flaky on a loaded box. They rely on nextest's process-per-test isolation, which CI and `CLAUDE.md` mandate; `.githooks/pre-push` still runs `cargo test`, which is pre-existing doc drift and not fixed here.

A DB-backed handler test drives the real `login_router()` behind the real `login_rate_limit_middleware` and counts bcrypt verifies, pinning the one line that maps the budget to the `TimingPad`: within budget an unknown-username rejection costs exactly 1 verify, past the budget exactly 0. Round 3 found that hardcoding either value there — including `TimingPad::Off`, which silently deletes the timing fix on the only surface it protects — passed all 93 tests in the selector. **Both mutations now fail that test, and I re-ran each to confirm**, along with re-introducing clear-on-2xx (which fails the budget-reset test).

The pad-budget middleware tests use a `pad_budget_app` harness whose handler echoes the pad decision it was handed, since a padded and an unpadded rejection are otherwise byte-identical. They cover: a sweep exhausting the budget (and still getting 401, never 429), a correct login succeeding afterwards, a success clearing the bucket, successes never being charged, and `0` disabling the budget. The previous `test_successful_logins_are_not_charged_to_the_per_ip_cap` was vacuous — a 10,000 cap and 20 requests — and a reviewer proved it passed with the `401` guard deleted; **I re-ran that same counterfactual against the new tests and two of them now fail.**

### Not run / not done

- The full backend suite and any e2e stack.
- `jscpd` itself — no npm network in this environment. The gate strips inline `#[cfg(test)]` modules, so the 4.0% it reported came from the production rejection arms; those are now collapsed into a single `reject_login`, and the enumeration tests were additionally refactored onto shared helpers.
- No end-to-end HTTP timing measurement. A residual sub-millisecond difference remains because the wrong-password arms also run an `UPDATE`; it is the same order as the residual the #3371 entry records.
- `authenticate_with_provider` / `authenticate_ldap` carry similar wording but have no production callers (the issue says the same), so they are left alone.


brandonrc:
## Audit — rounds 1 and 2 (security / correctness, both rounds)

**Round 1 (`e9d07323`): NEEDS-WORK.** The enumeration fix was complete — all eleven arms of `authenticate` byte-identical, no `WWW-Authenticate`/`Retry-After` leak, lockout mechanics unchanged (#1871 confirmed: a correct password still clears the lock), `dummy_bcrypt_hash` at the real cost factor. Two design problems: the bcrypt pad lived inside `AuthService::authenticate`, which has ten callers including Basic auth on every package request (so SSO-only deployments would burn a bcrypt per `pip`/`cargo` request), and the pad's swallowed error vs the real verify's propagated one gave a 401-vs-503 oracle under a saturated auth semaphore.

**Round 2 (`d51a8ec7`): NEEDS-WORK.** Both round-1 defects verified fixed by counterfactual: the pad is scoped to `authenticate_for_login` (`TimingPad::On`) and the nine other callers return before any bcrypt exactly as main; every login arm now goes through one `verify_password(...)?` with the dummy hash substituted, so status and path are identical by construction; `LoginFailure.reason` reaches only the persisted audit event. Three new problems, found by both reviewers: the per-IP failed-login cap was a pre-handler shed, so 30 failures from one IP (or one proxy address) refused *every* login behind it for the window, successes included — the #1871 shape at IP granularity; the shared `authenticate()` gained a WARN on every rejection arm, which every Basic-auth package request would trigger; and the CHANGELOG named `RATE_LIMIT_TRUSTED_CIDRS` where it meant `RATE_LIMIT_TRUSTED_PROXY_CIDRS`. The "only failures are charged" test was also vacuous (passed with its guard deleted).

**Round 3 in progress:** the cap now bounds the *pad*, not the request — once an IP's failed budget is spent, the dummy-bcrypt is skipped for that IP (bounding an attacker's amplification) but no login is ever refused and a success clears the bucket; WARN only on the login path; CHANGELOG corrected. One more audit pass and CI after that.

Product decisions recorded: pad kept (login path only); lockout message suppressed — `artifact-keeper-web`/SDK not checked for parsing the old text.

brandonrc:
## Audit — round 3 (`b0e181c3`): MERGE-WITH-NITS

Verified from the failing side: the pre-handler 429 shed is gone and the budget now gates only the pad — an end-to-end probe through the real `login_router()` + middleware with a real DB user ran four failure sweeps (plain 401s, never 429) and then a correct password from the exhausted IP → **200 with a real access token**. Past budget only hashless arms skip bcrypt; a row with a hash still hits the real `verify_password`. Absent extension → pad on, which also covers `RATE_LIMIT_ENABLED=false` and the #969 exemption path. `AuthEntry::Shared` gates the only two log sites in `authenticate_inner`; all nine non-login callers take `Shared` → zero new log lines, matching main. The round-2 mutation (charge every outcome, never clear) now fails two tests; the pad counterfactual still fails its test. 92/92 against a throwaway PG16; check/clippy clean; merge is a no-op on current main, zero markers.

Final pass in progress: clear-on-success is dropped so "at most N padded verifies per IP per window" is literally true even behind a shared proxy IP, and a handler-level test pins the budget→pad mapping (hardcoding either `TimingPad` variant passed the suite). Accepted as documented: a burned budget re-opens the timing oracle for that IP for the rest of the window — the timing half of #3504 is best-effort by design, the message half is unconditional.

brandonrc:
Final pass (`716a83f1`): clear-on-success removed (`RateLimiter::clear` deleted with its only caller), so "at most N padded verifies per IP per window" is literally true even behind a shared proxy IP; `test_successful_login_does_not_reset_the_per_ip_pad_budget` pins the inverse. New DB-backed `test_login_handler_maps_the_per_ip_pad_budget_to_the_timing_pad` drives the real `login_router()` behind the real middleware and counts bcrypt verifies: 1 within budget, 0 past it, never shed. Mutation proofs run before commit — hardcoding `TimingPad::On` fails 1/74, hardcoding `TimingPad::Off` (deleting the timing fix) fails 1/71, re-introducing clear-on-2xx fails 1/7. 93/93 on the audit selector, 604 on the wider one. Merging on green.

PR #3671, opened 2026-09-02
fix(security): escape LIKE patterns that are assembled in Rust (#3557)

### LINKED ISSUE #3557: LIKE patterns built in Rust (format!("%{}%", term)) are bound unescaped on ~30 search and listing paths
## Summary

#3500 closed the sites that assemble a `LIKE` **pattern inside SQL** (`LIKE $2 || '%'`, `LIKE '%' || $2 || '%'`, `$2 LIKE a.name || '-%'`). A second group builds the pattern **in Rust** and binds it whole (`format!("%{}%", q)`, `format!("{}%", p)`), then matches with a bare `path LIKE $2`. Same defect, different shape — and the shape is why the #3500 scanner cannot see them: there is no SQL-side concatenation to key on.

Roughly thirty sites are in this group. They are **not uniformly broken**, which is why this is an audit rather than a sweep:

- some already escape — `repositories.rs:5341`/`:5345` via `escape_like_literal`, `curation_service.rs:678` via an inline `replace()` chain;
- some take a value the caller pre-escaped — `maven.rs:406`, `proxy_catalog.rs:432`;
- some deliberately allow the over-match and re-check exactly in Rust — `artifact_service.rs:1948` and `proxy_catalog.rs:441` (`LIKE ANY($2)`);
- and some are simply unescaped.

Telling those apart needs per-site reading. Sweeping them mechanically would also change search semantics across the product in one commit, which is a product decision rather than a bug fix.

## Start here: the one confirmed live defect

`ArtifactService::list_page`, `count`, `list_for_repos_page` and `count_for_repos` built `format!("{}%", path_prefix)` from the request's browsed folder and matched it with `path LIKE $2` — no escaping. That is the same folder-listing defect #3500 fixed in `tree.rs`, on `GET /api/v1/repositories/{key}/artifacts?path_prefix=`, which is the bigger endpoint of the two.

**This one is already fixed** — it was folded into the #3500 PR because it is an identity/browse filter rather than a free-text search, with a regression test that fails on the parent commit showing `["ab/collapsed.bin"]` where `["a\b/real.bin"]` is required. It is described here so the audit starts from a worked example of what a real one looks like.

## The remaining group

Free-text `?search=` / `?q=` filters, built as `format!("%{}%", term)`:

| file | lines |
|---|---|
| `api/handlers/groups.rs` | 208 |
| `api/handlers/users.rs` | 249 |
| `api/handlers/builds.rs` | 180 |
| `api/handlers/packages.rs` | 160 |
| `api/handlers/composer.rs` | 1483 |
| `api/handlers/terraform.rs` | 2095 |
| `api/handlers/nuget.rs` | 2657 |
| `api/handlers/repositories.rs` | 6279, 6350, 6632, 6710 |
| `services/repository_service.rs` | 372 |
| `services/artifact_service.rs` | 1688, 1734, 1813, 1878 (the `search_query` half, deliberately left) |
| `services/search_service.rs` | 165 (suggest prefix) |
| `api/handlers/approval.rs` | 2336 |
| `api/handlers/promotion.rs` | 3361 |

Impact where unescaped: a `%` or `_` in the search term acts as a wildcard, so the result set and its `total` include rows the user did not ask for; a backslash quotes the following character, so a term containing one cannot be typed to find its own row. Within one already-authorized scope in every case — these are correctness bugs, not authorization ones.

`approval.rs:2336` and `promotion.rs:3361` are worth reading first among the rest: their pattern comes from a path rather than a search box, so they are closer to the `list_page` shape than to the search shape.

## Suggested shape

1. Decide the product question once: is a search box a literal substring search (the direction `repositories.rs:5345` and `curation_service.rs` already took) or a wildcard mini-language? Everything else follows.
2. Prefer routing through `escape_like_literal` rather than new inline `replace()` chains — `curation_service.rs:678` is a third copy of the escape rule already.
3. Per site, a behavioural regression test that fails before the change. A source-level gate cannot cover this group: the pattern is complete before it reaches SQL, so there is nothing in the SQL text to assert on. #3500's gate is explicitly scoped to the SQL-assembled shape for that reason.
4. Leave the deliberate over-matchers alone, and say so in their doc comments where it is not already stated (`artifact_service::list_by_path_prefixes` documents it; `LIKE ANY` at `artifact_service.rs:1948` and `proxy_catalog.rs:441` rely on a Rust re-check and are not defects).

## Related

- #3500 — the SQL-assembled half, and where `escape_like_literal` + `ESCAPE '\'` vs `ESCAPE ''` is decided.
- #3492 / #3493 — the same class on the Maven delete guards, where the consequence was data loss.
- #1000 / #998 — the original escaping wave across the package handlers.

brandonrc:
## Summary

`#3500` closed the `LIKE` sites that assemble their **pattern inside SQL** (`LIKE $2 || '%'`). This closes the second shape from the same class: the pattern is assembled **in Rust** — `format!("%{}%", term)` for a `?search=`/`?q=` box, `format!("{}%", prefix)` for suggest, `format!("{}/%", key)` for a directory — and bound whole to a bare `name ILIKE $2`. The #3500 source gate cannot see it, because by the time the string reaches SQL it is one bind with no concatenation to key on.

Where the term was unescaped, a `%` matched any run of characters and a `_` any single one, so the page **and** the `total` that drives `total_pages` carried rows the caller never asked for; a backslash — Postgres's DEFAULT `LIKE` escape character — quoted the character after it, so a row whose name contains one could not be found by typing its own name. Always within one already-authorized scope, so this widened result sets rather than crossing a permission boundary — though on the unauthenticated `GET /api/v1/search/suggest` a single `%` was a name-enumeration oracle over everything the same caller could already page through.

Every site now escapes through the shared `escape_like_literal` before the `%` is appended.

**The SQL-side `ESCAPE '\'` clauses are semantically inert** — `\` is already Postgres's default escape character — and are not the behaviour change. They document the helper's contract and are what the structural gates read; the behaviour change is entirely `escape_like_literal`.

## Audit result

I enumerated the cohort with the issue's grep, then read every site. **32 production sites** in 13 files, after review widened the recogniser:

| outcome | sites |
|---|---|
| **fixed** (were unescaped) | `groups.rs:208`, `users.rs:249`, `builds.rs:180`, `packages.rs:160`, `composer.rs:1483`, `terraform.rs:2095`, `nuget.rs:2657`, `repositories.rs:6279/6350/6632/6710`, `repository_service.rs:372`, `search_service.rs:165`, `artifact_service.rs:1688/1734/1813/1878` + `ArtifactService::search` |
| **fixed in review** | `search_service::build_name_filter` (`?name=`), `migration_service::get_last_migration_time` |
| already escaped | `repositories.rs:5341/5345`, the four `path_prefix` sites (#3500), `maven.rs:320/406` |
| routed through the shared helper | `curation_service.rs:678` — a third hand-rolled copy of the escape rule |
| `ESCAPE` clause added for contract consistency | `repositories.rs:3556` |
| deliberate, left alone | `artifact_service.rs:1948`, `proxy_catalog.rs:441` — `LIKE ANY($n)` fan-outs that over-fetch on purpose and re-check exactly in Rust |
| **not defects** | `approval.rs:2336`, `promotion.rs:3361` — `#[cfg(test)]` fixture helpers, not request paths |

**Product decision the issue asks for:** a search box is a **literal substring search**. The one exception is `GET /api/v1/search/advanced`'s `?name=`, where `*` stays the wildcard because the code intends it to — there the term is escaped *first* and the `*` translated after, so `?name=a_b.jar` no longer also returns `axb.jar` while `?name=my-lib*` still wildcards. (The earlier claim that `*` was "advertised" was wrong: the only statement of it is a private code comment. Reviewers caught that, and leaving the filter unescaped on that basis left the defect live on the most public search surface.)

`migration_service::get_last_migration_time` has no callers today. I escaped it rather than deleting it — removing a public method is not a security fix's business, and leaving the shape unescaped is what the gate exists to prevent.

## The structural gate — what it does and does not pin

It is a **lint, not a proof**, and the code and CHANGELOG now say so:

- It is **function-scoped**, like the #3500 gate: it shows an escaper is used *somewhere* in the function owning the `format!`, not on that value. Reverting only the search term at the four `artifact_service` sites leaves it green. **The DB-backed behavioural tests are the real regression pin.**
- `push_str`, `+`, `concat!` and raw-string literals still evade the recogniser.

Five review findings against it are closed:

- The over-match opt-out was satisfied by two lines of prose. It is now a **code construct** — `like_any_overmatch_accepted`, an identity function accepted only when it encloses *that* `format!` (value-scoped, unlike everything else in the gate) **and** the owning function's *body* really contains a `LIKE ANY(`. Both halves are required: the wrapper alone would let an ordinary search site be silenced, and the `LIKE ANY` alone was the round-1 defect. The `LIKE ANY` half is matched below the `fn` header, not across the doc comment `enclosing_fn` includes — which is what made prose sufficient before.
- Both scanners stop skipping `api/handlers/mod.rs` wholesale (the prose forcing that moved inside the `#[cfg(test)]` module), so ~1,700 lines of production code including `escape_like_literal` itself are now covered — verified by planting a site there and watching it fail. A comment quoting the shape is no longer a false positive.
- Module extents are anchored on indentation instead of brace-counting, which walked through `'{'` char literals.
- **An out-of-line `#[cfg(test)] mod name;` opens no brace**, so the search for a closing one grabbed an unrelated brace further down and swallowed everything between. This was live: `api/handlers/mod.rs`'s `mod test_db_helpers;` produced region 217–888, excluding lines 219–312 *of the very file the gate had just started covering*, and `formats/mod.rs`'s `mod format_tests;` swallowed the whole `FormatHandler` trait. Both regions are gone (verified against the real tree).
- **A trailing comment on the closing brace** (`} // end of tests`) extended the region to the next brace at that indent — a production item's — and absorbed it. The comparison now strips a trailing comment.

Known and accepted: the widened recogniser **will** over-flag a percentage string (`format!("{:.1}%", pct)`). That is a false positive, not a hole — it fails loudly with its own file and line, which is the right failure direction for a gate — and `backend/src` contains none today. The module prose says so.

## Regression tests — all fail without the fix

| test | failure without the fix |
|---|---|
| `test_list_page_search_query_treats_like_metacharacters_literally_3557` | 5 paths returned, 1 expected — asserts `list_page`/`count` **and** `list_for_repos_page`/`count_for_repos` (the virtual listing, which the gate is blind to and which had no coverage at all) |
| `test_suggest_prefix_treats_like_metacharacters_literally_3557` | 5 names returned, 1 expected |
| `test_name_filter_wildcards_only_on_star_3557` | `?name=a_b.jar` returns 4 rows, 1 expected; positive control pins that `*` still wildcards |
| `every_rust_assembled_like_pattern_escapes_its_operand` | names each offending file:line |
| 5 pure-unit builder tests | `"%100%%"` vs `"%100\%%"` |

Both DB tests carry a positive control, so a fix that escaped its way into matching nothing also fails.

## Verified

- `cargo fmt --check` silent; `cargo check --lib --tests` clean; `cargo clippy --all-targets -- -D warnings` clean.
- Postgres 16 + `sqlx migrate run`, `AK_TESTS_REQUIRE_DB=1`: **31/31** on the `#3557` + gate + builder selection, **292/292** across every touched module.
- Guard probes: a comment-only fake marker, a `{}/%` site and a plain `%{}%` site are all flagged; a real wrapper, a test-module site and prose quoting the shape are not.
- `.sqlx`: **eight `query!`-macro statements** changed text. That shows as **14 JSON paths** in the diff, because five of the eight queries have an entry in *both* caches (`.sqlx/` and `backend/.sqlx/`) and one is an add+delete rather than a rename. They were re-keyed by hand (query string + `hash` only, all other metadata byte-identical) rather than by `cargo sqlx prepare`, which the newer CLI would use to rewrite 318 files with `origin` metadata alone. A reviewer regenerated all eight against a live migrated Postgres and confirmed them byte-identical. The two-cache split is pre-existing and tracked separately as **#3674**; nothing here makes it worse. No new entries were needed for the review fixes — `search_service`, `migration_service` and `repositories.rs:3556` are all runtime-bound queries.

Closes #3557


brandonrc:
## Audit round 1 — 3 reviewers (correctness / security+guard / `.sqlx` cache+integration)

**Verdict: MERGE-WITH-NITS ×3; rework requested before merge.** All three passes agree the escaping is right: the cohort was independently re-enumerated to the same 29 production sites; every changed operand goes through `escape_like_literal` and every predicate carries `ESCAPE '\'` on both page and count queries; the helper is a single-pass `chars()` walk so ordering can't break; a trailing lone `\` becomes `\\` (Postgres-verified); and all 40 `ESCAPE '\'` literals in non-raw Rust strings decode to a real backslash. The 8 hand-edited `.sqlx` entries reproduce byte-for-byte from a live regeneration against a migrated Postgres 16, and all 926 committed cache entries satisfy `filename == sha256(query)`. Revert proof: source-only revert of the escapers fails 6/6 new tests with the over-matched rows in the output. **No site crossed an authorization boundary** — at all 18 the LIKE is ANDed with an independent scope/visibility predicate; the highest-ranked was `search/suggest` (unauthenticated name-enumeration via `%`).

### Should-fix
1. **`GET /api/v1/search?name=` is still unescaped** (`build_name_filter`, `search_service.rs:149`). The "advertised `*` wildcard" is a private code comment, not a documented contract. Keep `*` as the wildcard, but escape `%`/`_`/`\` before translating it — today `?name=a_b.jar` matches `axb.jar` and `a\b.jar` can't be found by its own name.
2. **Recogniser gap:** `"{}/%"` shapes are pinned as non-patterns, yet `migration_service.rs:1290` is one (`source_path LIKE $2`, unescaped; zero callers). `repositories.rs:3546` is escaped but bound without `ESCAPE`.
3. **The structural guard is weaker than the CHANGELOG claims.** Function-scoped: reverting only `artifact_service.rs:1701` still passes it (a sibling escaper on the previous line satisfies the check). Ten evasion shapes probed, seven get through (`concat!`, `String + &q`, `push_str`, `format!(r"…")`, same-function-different-value…), and the `LIKE ANY(` over-match marker is satisfied by a *comment*. Both scanners also skip `api/handlers/mod.rs` wholesale. Make the marker a code construct, un-skip `mod.rs`, and word the claims to what it pins; the behavioural tests are the real regression pin.
4. **Coverage:** `list_for_repos_page`/`count_for_repos` (virtual-repo `?q=`) and the other `artifact_service` sites have neither gate nor test coverage.
5. Merge `main` (CHANGELOG `### Security` heading now exists there — fold to one heading).

### For the record
`ESCAPE '\'` is Postgres's default, so the ~40 SQL edits are semantically inert — the entire behaviour change is `escape_like_literal`. Fine to keep as documentation of the helper's contract. The repo's two-`.sqlx`-cache split (neither complete on its own) is pre-existing and filed as #3674.

brandonrc:
## Audit round 2 (head `28309a84`) — MERGE-WITH-NITS

Every round-1 item is closed and proved by counterfactual: `build_name_filter` escapes then translates `*` (reverting it: `?name=a_b.jar` → `["a%b.jar", "a\b.jar", "a_b.jar", "axb.jar"]` vs `["a_b.jar"]`; `a\b.jar` findable; `*` still wildcards, 5/5 with `total == 5`); both predicates carry `ESCAPE '\'`; the widened recogniser flags 32 production sites, all genuine LIKE patterns; the value-scoped `like_any_overmatch_accepted()` wrapper works (a wrapper on a *different* `format!` in the same fn leaves the unwrapped one flagged); the merge commit is a pure CHANGELOG resolution. 39/39 against a throwaway DB, `check` and `clippy -D warnings` clean.

Last pass in progress on the new indentation-anchored module extents, which have two silent-false-negative shapes: an out-of-line `#[cfg(test)] mod name;` (live in `api/handlers/mod.rs:217` and `formats/mod.rs:45`, excluding ~96 production lines today — benign now, wrong in principle) and a trailing comment on the module's closing brace extending the region to EOF. Plus restoring the `LIKE ANY(` conjunct so the wrapper alone can't silence a site. Merges when that lands green.

brandonrc:
Final pass (`98dcd285` + merge `5149b384`): out-of-line `#[cfg(test)] mod name;` items skipped (regions in `api/handlers/mod.rs` now `[(313,888),(890,1927)]` — lines 219–312 back in scope; `formats/mod.rs` computes `[]` so `FormatHandler` is scanned), trailing comments on the closing brace stripped before comparison, and the opt-out now needs both the value-scoped wrapper and `LIKE ANY(` in the fn body. Probes: sites after a commented brace, after an out-of-line module, and a wrapped site without `LIKE ANY` are all flagged; a wrapper with real `LIKE ANY` and a genuine test-module site are not. 39/39 with a real Postgres. Merging.

PR #3677, opened 2026-09-02
feat(opensearch): add OPENSEARCH_INDEX_PREFIX to namespace both indexes

### LINKED ISSUE #3669: OpenSearch index names are hardcoded — no prefix/namespace, so two instances on one cluster collide
## Summary

The two OpenSearch index names are module-level constants with no configuration override:

```rust
// backend/src/services/opensearch_service.rs:29-30
const ARTIFACTS_INDEX: &str = "artifacts";
const REPOSITORIES_INDEX: &str = "repositories";
```

They are used at every call site (create, index, delete, search, count, bulk, refresh-interval,
force-refresh), and the config surface exposes only `OPENSEARCH_URL`, `OPENSEARCH_USERNAME`,
`OPENSEARCH_PASSWORD` and `OPENSEARCH_ALLOW_INVALID_CERTS` — there is no index name, no prefix,
and no alias indirection (`alias` does not appear in the file).

The names are also the bare generic nouns `artifacts` and `repositories`, which is about as
collision-prone as an index name can be on a shared cluster. On startup,
`configure_indexes()` creates both unconditionally whenever `OPENSEARCH_URL` is set.

## Why this matters

Any deployment that points more than one Artifact Keeper instance at one OpenSearch cluster —
prod plus staging, blue/green, or a per-team instance on shared search infrastructure — has the
two instances silently sharing both indexes. Likewise for a cluster where some other application
already owns an index called `artifacts`.

Concrete consequences today:

1. **`full_reindex` is destructive across instances.** It rebuilds the shared index from one
   instance's PostgreSQL, so documents owned by the other instance are not in the source and are
   effectively orphaned.
2. **Reindex degrades the other instance cluster-wide.** `full_reindex_artifacts` sets
   `set_refresh_interval(ARTIFACTS_INDEX, "-1")` for the duration — a normal bulk-load
   optimisation that here becomes a cross-instance side effect, suppressing refresh on an index
   another instance is writing to.
3. **No alias-swap upgrade path.** The standard zero-downtime reindex is to build
   `artifacts-v2` and then atomically repoint an `artifacts` alias. Hardcoded concrete index
   names make that impossible without code changes, so any future mapping change requires
   downtime or an out-of-band migration.

In fairness the startup path is careful in other respects: it degrades gracefully when
OpenSearch is absent, and the bootstrap reindex takes a singleton lease
(`opensearch_bootstrap_reindex`, 6 h TTL with heartbeat renewal) so a multi-replica deployment
does not have every replica reindex at once. The multi-replica case is handled; the
multi-*instance* / shared-cluster case is not.

## Proposal

Add an optional `OPENSEARCH_INDEX_PREFIX` (default empty, so existing deployments resolve to
exactly `artifacts` and `repositories` and nothing changes on upgrade). Resolve the two names
through it on `OpenSearchService` so every call site picks it up, e.g. prefix `ak-prod-` yields
`ak-prod-artifacts` / `ak-prod-repositories`.

That is a small, self-contained change, and it also opens the door to the alias pattern later
without another config break.

## Acceptance criteria

- With `OPENSEARCH_INDEX_PREFIX` unset, the resolved index names are byte-identical to today.
- With it set, every operation (create, index, delete, search, count, bulk, refresh) targets the
  prefixed names, and two instances with different prefixes on one cluster do not interfere.
- Invalid prefixes are rejected at construction rather than producing a cryptic OpenSearch error
  (OpenSearch index names cannot contain uppercase, spaces, `\ / * ? " < > |` `,` `#`, or start
  with `-`, `_`, `+`).

Happy to open a PR — it is contained to `opensearch_service.rs` plus one config field.

knowinglyAnonymous:
Closes #3669

## What

The two OpenSearch index names were module-level constants with no override, so every
Artifact Keeper instance pointed at a cluster used the bare names `artifacts` and
`repositories`. Two instances on one cluster (prod + staging, blue/green, or a shared
search tier) therefore read and wrote each other's documents — and `full_reindex` on one
rebuilt the shared index from *its* PostgreSQL, orphaning the other's documents, while
`set_refresh_interval(..., "-1")` suppressed refresh on an index the other instance was
still writing to.

This adds an optional `OPENSEARCH_INDEX_PREFIX`.

## Changes

- `OpenSearchService` now resolves both index names once at construction and stores them
  (`artifacts_index`, `repositories_index`), with `artifacts_index()` / `repositories_index()`
  accessors. All 17 call sites — create, index, delete, search, count, bulk, refresh-interval,
  force-refresh — use the resolved names.
- New `new_with_prefix(url, username, password, allow_invalid_certs, prefix)`.
  **`new(...)` is kept and delegates with an empty prefix**, so every existing caller and test
  is untouched.
- New `validate_index_prefix`, called from the constructor so a deployment typo is a startup
  configuration error rather than an opaque OpenSearch rejection on the first write. Rejects
  uppercase, whitespace, `\ / * ? " < > | , #`, and a leading `-`, `_` or `+`.
- New config field `opensearch_index_prefix` (env `OPENSEARCH_INDEX_PREFIX`), defaulting to
  empty and listed in the config dump as a non-secret.
- `main.rs` startup passes the configured prefix.

The `ARTIFACTS_INDEX` / `REPOSITORIES_INDEX` constants remain as the **base** names the prefix
is applied to, so `test_constants` still pins them.

## Backwards compatibility

With `OPENSEARCH_INDEX_PREFIX` unset the resolved names are byte-identical to today
(`artifacts`, `repositories`) — no reindex, no migration, no behaviour change on upgrade. A
covering test asserts exactly that.

## One pre-existing test changed

`test_configure_indexes_creates_both_indexes` is a source-introspection test: it reads
`configure_indexes`'s own source text and asserted it contained the literal strings
`ARTIFACTS_INDEX` / `REPOSITORIES_INDEX`. Since the function now references the resolved
fields, the assertions were updated to `self.artifacts_index` / `self.repositories_index`. The
test's intent — that `configure_indexes` creates both indexes — is preserved. Flagging it
explicitly because it is a behavioural test of source text rather than of behaviour, so the
diff looks like a test being weakened when it is being re-pointed.

## Tests

11 new unit tests, all DB- and cluster-free (construction does not connect):

- `test_new_without_prefix_keeps_historical_index_names` — the compatibility guarantee
- `test_empty_prefix_matches_default`
- `test_prefix_namespaces_both_indexes` — `ak-prod-` → `ak-prod-artifacts` / `ak-prod-repositories`
- `test_distinct_prefixes_do_not_collide` — the actual point of the issue
- `test_validate_index_prefix_accepts_empty_and_plain` — including that `_` is barred only as the *first* character
- `test_validate_index_prefix_rejects_uppercase` / `_whitespace` / `_forbidden_chars` / `_bad_leading_char`
- `test_new_with_invalid_prefix_fails_construction` — fails fast, not at first use

## Follow-up this enables

With the names resolved through one seam, the standard zero-downtime reindex (build
`<prefix>artifacts-v2`, then atomically repoint an alias) becomes possible without another
config break. Not attempted here.

## Verification

```
cargo fmt --check
cargo test --workspace --lib opensearch_service
```

## Also in this diff

- A `CHANGELOG.md` entry under `## [Unreleased]` → `### Added`, since the release preflight
  reconciles merged PRs against that section.
- `opensearch_index_prefix: String::new()` added to nine exhaustive `Config { .. }` literals in
  tests (7 unit, 2 integration). Adding a field to `Config` requires touching every literal;
  none of them assert on the new field. Worth noting that `cargo clippy --workspace --lib`
  does **not** surface the two in `backend/tests/`, only `--all-targets` does.


brandonrc:
Heads-up that your two OpenSearch PRs conflict with each other, not only with main:

```
$ git merge-tree --write-tree refs/pr/3677 refs/pr/3678
CONFLICT (content): Merge conflict in backend/src/services/opensearch_service.rs
```

A review is running on #3678 (using OpenSearch for `/search/quick`) since it is the user-visible half. My inclination is to land the index prefix here **first**, because it changes how indexes are named and #3678 then only has to be written against the final naming — the reverse order means rewriting the query path twice.

If you agree, this one needs a rebase onto current main (it is a CHANGELOG conflict only) and I will get a review onto it next. If you would rather ship #3678 first, say so and I will flip the order.

brandonrc:
@knowinglyAnonymous — I've taken this over to get it unblocked and merged, as I said I would for contributor PRs stuck behind conflicts. I've pushed to your branch (`feat/opensearch-index-prefix-3669`) rather than opening a competing PR. Here is everything I changed.

## Rebase

Rebased onto current `main` (`8211ebac`). The only conflict was `CHANGELOG.md`: several bullets landed under `## [Unreleased]` since you opened this, including a new `### Added` section, so your `### Added` header collided with theirs. Resolved by keeping every bullet from both sides and appending yours to the existing section — `CHANGELOG.md` has zero deletions against `main`, and I checked line-by-line that nothing from `main` was dropped. Your bullet is still under `## [Unreleased]` → `### Added`.

No conflicts in any Rust file.

## Review of the substance

The thing I wanted to be sure of on this one is that a prefix which namespaces indexes applies at *every* site that constructs an index name. A prefix applied in four places out of six is worse than no prefix at all, because it splits reads from writes silently. I enumerated them:

| Site | Covered |
| --- | --- |
| `configure_indexes` → `ensure_index` (create) ×2 | yes |
| `index_artifact` / `index_repository` (write) | yes |
| `remove_artifact` / `remove_repository` (delete) | yes |
| `search_artifacts` / `search_repositories` (query) | yes |
| `is_index_empty` (count) | yes |
| `full_reindex_artifacts` / `full_reindex_repositories` → `bulk_index`, `set_refresh_interval` ×4, `force_refresh` ×2 | yes |
| `health_check` | n/a — `ClusterHealthParts::None` is cluster-scoped, not index-scoped |
| `main.rs` startup (the only production constructor) | yes |
| Migration worker / `RepositoryService` / `ArtifactService` | n/a — they call the service methods above, no index names of their own |

That is complete. I also grepped the tree for stray `"artifacts"` / `"repositories"` literals outside the service; the hits are filesystem directories and OpenAPI tag names, nothing that reaches a cluster. And keeping `new()` delegating to `new_with_prefix("")` was the right call — it kept this diff to the seam and left every existing caller and test alone.

## What I added

Two things, in one `docs(opensearch)` commit on top of yours:

**1. `.env.example`.** The variable was in `Config` and the config dump but not in `.env.example`, where the three sibling `OPENSEARCH_*` variables are all listed. That is where an operator looks.

**2. The upgrade note the entry was missing.** Your entry documents the unset case correctly — byte-identical names, no reindex — and I verified that. What it did not say is what happens when someone *sets* the prefix on a deployment that already has an unprefixed index. I traced it: names resolve at construction, so the first boot after setting it creates an empty `<prefix>artifacts` / `<prefix>repositories` pair, and then the existing bootstrap path rescues it — `is_index_empty()` sees the new index, one replica takes the `opensearch_bootstrap_reindex` lease, and a full reindex refills both in the background. So it is not a silent empty-index read, which is what I was checking for. But two consequences are worth an operator knowing: searches are served from a partially populated index until that reindex finishes, and the old unprefixed indexes are left behind holding disk until someone deletes them. Both are now in the upgrade note.

## Verified

- `cargo fmt --check`
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(opensearch_service)'`
- Diffstat against `main`: +273/−22 across 13 files — your +265/−22, plus the 8 `.env.example` lines, with the CHANGELOG net unchanged at +4 (the `### Added` header is no longer new).

## One thing I did not change

`validate_index_prefix` does not bound the prefix's *length*, so a prefix long enough to push the full name past OpenSearch's 255-byte index-name limit still fails at first write rather than at startup. Every realistic typo is caught by the rules you wrote, and I would rather not widen a clean diff for it. Noting it in case you want a one-line follow-up.

Thank you for this one — the constructor-seam approach, the fail-fast validation and the compatibility test were all the right shape, and the note about `test_configure_indexes_creates_both_indexes` being a source-introspection test rather than a behavioural one saved me a second look at that hunk. Merging once CI is green.

PR #3678, opened 2026-09-02
feat(search): use OpenSearch for /search/quick when configured

### LINKED ISSUE #3670: OpenSearch index is written but never queried — wire it into search when OPENSEARCH_URL is provided
## Summary

`OpenSearchService` maintains two indexes but **nothing ever queries them**. The index is
write-only: it is populated on artifact upload and repository create/update, cleaned on delete,
rebuilt by `full_reindex`, and health-probed — but no request path reads from it.

`search_artifacts` and `search_repositories` have no production callers:

```
$ grep -rn "search_artifacts" --include='*.rs' backend/src
backend/src/services/opensearch_service.rs:252:    pub async fn search_artifacts(
  ... plus test references only
```

The only other references are **source-introspection unit tests** — they call
`function_source(opensearch_service_source(), "search_artifacts")` and assert the function's own
source text contains `multi_match`, `fuzziness`, `track_total_hits` and the field boosts. They
verify the query is *written* correctly; they never execute it.

All actual search traffic (`/search/quick`, `/search/advanced`, suggest, trending, recent) is
served by `SearchService`, which is pure PostgreSQL and does not reference OpenSearch at all.

So an operator who sets `OPENSEARCH_URL` today pays for index writes on the upload path, a
cluster to operate, and a health dependency — and gets no query benefit whatsoever, because
nothing reads the index.

## Why this is worth closing rather than leaving

The PostgreSQL path is good and was deliberately optimised in #2871 / migration 176 (stored
trigger-maintained `search_vector` plus a partial GIN index), so this is **not** a performance
gap — `to_tsquery` over a GIN index is fast. What PostgreSQL does not provide is what the unused
OpenSearch query was built for:

- **typo tolerance** — the unused builder sets `fuzziness`; `to_tsquery` has no equivalent
- **relevance ranking** — `multi_match` with `name` and `path` boosts, versus ordering by
  explicit sort parameters
- **`track_total_hits`** for accurate deep-pagination totals

That is a real capability difference, and the code to deliver it is already written and
committed — it is simply not wired to anything.

Either it should be wired up when `OPENSEARCH_URL` is provided, or the read-side code and its
introspection tests should be deleted so the dependency and the operational cost are honest
about what they buy. This issue proposes the former.

## The blocker that has to be got right: visibility

The PostgreSQL path enforces repository visibility per request: `resolve_accessible_repos`
derives an `AccessScope` and intersects it with the caller's token scope, yielding
`accessible_repo_ids`, so an anonymous caller sees only public repositories, a non-admin sees
public plus role-assigned, and a repo-scoped token cannot enumerate other repositories.

**An OpenSearch read path must not reproduce this from the index.** `ArtifactDocument`
denormalises `is_public`, `repository_key` and `repository_name`, and those copies go stale:
`RepositoryService::update` reindexes only the *repository* document, never the artifact
documents belonging to it. So flipping a repository from public to private leaves every one of
its artifact documents carrying `is_public: true` until the next full reindex. `delete`
similarly calls only `remove_repository`, leaving that repository's artifact documents orphaned
in the index — still marked public.

Filtering an OpenSearch query on the index's own `is_public` would therefore serve private
artifacts to anonymous callers. This is the same shape as #3323, where the visibility middleware
authorised only the repository named in the URL and not the member set.

The safe design is to filter on `repository_id` against the authoritative
`accessible_repo_ids` list computed from PostgreSQL for that request, and to ignore the
denormalised `is_public` entirely.

(The stale/orphaned artifact documents on repository update and delete are arguably a separate
defect. They are harmless while nothing reads the index, which is why they are described here
rather than filed separately — but they must be fixed or designed around before any read path
ships.)

## Proposal

1. Give `search_artifacts` a repository-id allowlist parameter and translate it into an
   OpenSearch `filter` on `repository_id` (`None` = admin/unrestricted, no filter).
2. Route `/search/quick` through OpenSearch when `state.search_service` is `Some`, passing the
   same `accessible_repo_ids` the PostgreSQL path already computes.
3. **Fall back to PostgreSQL on any OpenSearch error**, so a degraded or unreachable search
   cluster cannot take search down — matching the graceful-degradation posture the startup path
   already takes.
4. Leave `/search/advanced` on PostgreSQL for now: its facet counts would need OpenSearch
   aggregations, which is a larger piece of work and a separate decision.

## Acceptance criteria

- With `OPENSEARCH_URL` unset, behaviour is byte-identical to today.
- With it set, `/search/quick` returns OpenSearch-ranked results with typo tolerance.
- An anonymous caller never receives an artifact from a private repository, **including** when
  that repository was public at index time and has since been made private.
- A repo-scoped token cannot retrieve results from repositories outside its scope.
- An OpenSearch outage degrades to PostgreSQL results rather than an error.

Happy to open a PR for items 1–3.

knowinglyAnonymous:
Closes #3670

## What

`OpenSearchService` maintained two indexes that nothing ever read. `search_artifacts` and
`search_repositories` had no production callers — the only references were
source-introspection tests that read the function's own source text and asserted it contained
`multi_match` / `fuzziness` / `track_total_hits`, verifying the query was *written* correctly
without ever executing it. All search traffic went to `SearchService` (pure PostgreSQL).

So an operator who set `OPENSEARCH_URL` paid for index writes on the upload path, a cluster to
run and a health dependency, and got no query benefit at all.

This routes `/search/quick` through OpenSearch when it is configured.

## Why, given the PostgreSQL path is already fast

This is **not** a performance change. The PostgreSQL path was deliberately optimised in #2871 /
migration 176 (stored trigger-maintained `search_vector` + partial GIN index) and
`to_tsquery` over that index is fast. What it cannot do is what the unused OpenSearch query was
built for: **typo tolerance** (`fuzziness: AUTO`), **relevance ranking** (`multi_match` with
`name^3` / `path^2` boosts) and `track_total_hits` for accurate deep-pagination totals.

## The security-critical part: visibility is enforced on `repository_id`, never on `is_public`

The PostgreSQL path resolves an `AccessScope` per request via `resolve_accessible_repos`
(visibility ∩ token scope). The OpenSearch path is now given **that same scope**.

It deliberately does **not** trust the index's own `is_public` field. `ArtifactDocument`
denormalises repository visibility at index time, and `RepositoryService::update` reindexes only
the *repository* document — never the artifact documents belonging to it. So a repository
flipped from public to private leaves every one of its artifact documents carrying
`is_public: true` until the next full reindex, and `delete` likewise leaves that repository's
artifact documents orphaned in the index, still marked public. Filtering on the indexed
`is_public` would therefore serve private artifacts to anonymous callers — the same shape as
#3323, where the visibility middleware authorised only the repository named in the URL and not
the member set.

`AccessScope` is threaded through rather than an `Option<&[Uuid]>` precisely because it makes
"no restriction" and "restricted to nothing" impossible to confuse — the footgun that type was
introduced (#1617) to remove. `Restricted([])` **short-circuits before the cluster is queried
at all**, so deny-by-default does not depend on how OpenSearch treats an empty `terms` array.

There is a test asserting the emitted filter never mentions `is_public`, so a future refactor
cannot quietly reintroduce the stale-flag filter.

## Changes

- `visibility_filter_clause(&AccessScope) -> Option<Value>`: `Admin` → no clause,
  `Restricted(ids)` → a `terms` clause on `repository_id` (mapped `keyword`, so exact).
- `search_artifacts` takes `scope: &AccessScope`, appends the visibility clause to the existing
  filter clauses, and short-circuits an empty allowlist to zero results. Its signature had no
  callers, so nothing else moves.
- `quick_search` prefers OpenSearch when `state.search_service` is `Some`, and
  **falls back to PostgreSQL on any OpenSearch error** — a degraded or unreachable cluster must
  not take search down, matching the graceful degradation the startup path already applies when
  OpenSearch is absent entirely. The fallback logs at `warn`.
- `build_search_result_item_from_doc` maps an `ArtifactDocument` onto the same
  `SearchResultItem` the PostgreSQL path returns, so a caller cannot tell which backend served
  the request.

## Backwards compatibility

With `OPENSEARCH_URL` unset — the default — `state.search_service` is `None` and the handler
takes exactly today's PostgreSQL path. No behaviour change.

## Deliberately out of scope

`/search/advanced` stays on PostgreSQL. Its facet counts would need OpenSearch aggregations,
which is a larger piece of work and a separate decision. Suggest, trending and recent are
untouched.

The stale/orphaned artifact documents on repository update and delete are a real defect, but
harmless while nothing reads the index. This PR routes around them by not trusting `is_public`;
it does not fix them. If you would prefer the reindex-on-visibility-change fix to land first, I
am happy to do that instead — the two interact, and I would rather follow your call on
ordering.

## Tests

4 new unit tests, DB- and cluster-free:

- `test_visibility_filter_admin_has_no_clause`
- `test_visibility_filter_restricted_emits_terms_on_repository_id`
- `test_visibility_filter_never_uses_is_public` — pins the security property above
- `test_empty_restricted_scope_is_deny_by_default`

Full `opensearch_service` suite: 203 passed, 0 failed.

## Verification

```
cargo fmt --check
cargo test --workspace --lib opensearch_service
```


brandonrc:
## Review — NEEDS-WORK

Thank you for this; the part that is easiest to get wrong is right. Visibility is filtered on the live `repository_id` rather than the indexed `is_public` copy, and a mutation that swaps the `terms` clause for `{"term":{"is_public":true}}` fails exactly three tests, so the security property is genuinely pinned rather than incidentally true. 24 of 24 tests pass here and they cannot compile against main, which is what I want to see.

Three things to fix before it lands.

**Blocker: an empty index is indistinguishable from no results.** `main.rs:305-362` configures the indexes, detaches the bootstrap reindex at `:310`, and registers the search service at `:362`, so requests are served while the index is still filling — and `full_reindex_artifacts` disables refresh for the whole run (`opensearch_service.rs:473`, restored at `:562`). The handler falls back to Postgres only on `Err` (`search.rs:413-441`). So an existing instance that sets `OPENSEARCH_URL` for the first time answers `200 {"results":[]}` for the entire reindex — hours at `BATCH_SIZE=1000` — where it used to return Postgres hits. `POST /admin/search/reindex` reopens the same window (`admin.rs:1501`). Gate the read path on a readiness flag rather than on `search_service.is_some()`.

**Should-fix: `repository_key` returned to clients goes stale after a rename.** `build_search_result_item_from_doc` returns `d.repository_key` (`search.rs:135`), the key is mutable (`repositories.rs:904`), and `RepositoryService::update` reindexes only the repository document (`repository_service.rs:1470-1482`). Rename `npm-remote` to `npm-proxy` and OpenSearch keeps answering with the old key while the Postgres path joins the live row (`search_service.rs:331-347`); download URLs built from it 404.

**Should-fix: deleted artifacts and artifacts of deleted repositories become visible.** The Postgres path filters `a.is_deleted = false` at `search_service.rs:347`, `:388` and `:426`; `ArtifactDocument` carries no such field, removal from the index is a best-effort detached task (`artifact_service.rs:2166-2179`), and `RepositoryService::delete` removes only the repository document (`repository_service.rs:1499-1512`). Non-admins are still protected by the repository filter, so this is stale data rather than a leak, but an admin sees a deleted repository's artifacts.

Seven nits are worth a look too: unbounded `q`, no cap on the terms array, the ordering change, `track_total_hits` set but never used, silently dropped documents, and a doc comment that has taken over `build_search_result_item`'s rustdoc.

Bookkeeping: the CHANGELOG bullet is correctly placed and in house style; please cite `(#3670, #3678)` and add an upgrade note about the first-reindex window once the readiness gate is in.

One coordination note, on #3677 rather than this PR: the two conflict in `opensearch_service.rs`, and I would rather land the index prefix first so this one is written against the final index naming.

brandonrc:
@knowinglyAnonymous — taken over and pushed to your branch (`feat/opensearch-quick-search-3670`) rather than opened as a competing PR. Rebase plus two fixes; details below.

## Stacking

First thing I checked, because the two PRs look stacked and a stacked PR rebuilt on a stale base silently reverts its parent. They are **not** stacked: `feat/opensearch-index-prefix-3669` and `feat/opensearch-quick-search-3670` both fork from `e2be9fef` on `main`, and neither contains the other's commit. So each rebases onto `main` independently, which is what I did.

They do both rewrite `search_artifacts`, so whichever of the two merges second will conflict there — a textual conflict, not a semantic one, and I will resolve it at merge time. Nothing for you to do.

## Rebase

Onto current `main` (`8211ebac`). Only `CHANGELOG.md` conflicted: a `### Added` section landed under `## [Unreleased]` since you opened this, so your new header collided with it. Kept every bullet from both sides, appended yours to the existing section. Zero deletions in `CHANGELOG.md` against `main`, verified line-by-line that nothing from `main` was dropped, and your bullet is still under `## [Unreleased]` → `### Added`. `search.rs` and `opensearch_service.rs` merged clean.

## The security review

This is the part I spent the most time on, because `/search/quick` gaining a second implementation means two code paths that must agree about what the caller may read, and a search endpoint that returns rows the caller cannot read is a disclosure bug.

**Your visibility design is correct and I kept it unchanged.** The handler resolves `AccessScope` through the same `resolve_accessible_repos` the PostgreSQL path uses — `resolve_visible_repos` (public ∪ role assignments ∪ `read`-carrying `permissions` grants, i.e. the #3697 fix) intersected with token scope — and passes that same scope to the cluster. I checked the PostgreSQL path's `WHERE` clause for any second gate you might have missed: it applies `($n::uuid[] IS NULL OR r.id = ANY($n))` and `($n = false OR r.is_public = true)`, and `quick_search` pins `public_only: false`, so the repository allowlist is the whole of it. Your `terms` clause is the same filter. Refusing to trust the indexed `is_public` is right for exactly the reason you gave, and short-circuiting `Restricted([])` before the cluster is queried rather than trusting an empty `terms` array is the detail I would have asked for if it were not already there.

**What I added: PostgreSQL vets the hits after the query, not only before it.** Your reasoning about the index going stale is right but stops one step short. `is_public` is not the only field that drifts — the *row itself* does. Several soft-delete paths mark `artifacts.is_deleted` with a direct `UPDATE` instead of going through `ArtifactService::delete_artifact` (Helm chart deletes, Maven and Conan version deletes), so `remove_artifact` never runs and the document survives in the index. The PostgreSQL path filters `a.is_deleted = false` on every single query. Without a re-check, the OpenSearch path becomes the one place in the product where a deleted artifact comes back — and for something yanked for a licence or CVE reason that is not a cosmetic difference.

So `retain_live_visible_hits` now takes the hits the cluster returned and keeps only those PostgreSQL vouches for: the row exists, is not soft-deleted, and its repository is in the caller's scope. One indexed lookup over at most 50 ids, and the cluster's ranking order is preserved — OpenSearch supplies the ranking, PostgreSQL the authority.

Re-checking the scope there is deliberately redundant with your `terms` clause. It is redundant *today*; it is the thing that still holds if the mapping drifts off `keyword`, if the cluster rejects or truncates an oversized `terms` list, or if the index is restored from another deployment's snapshot. On a disclosure-sensitive path I would rather the guarantee not depend on the search cluster behaving.

## The fallback

You were right that a degraded cluster must not take search down, and the `Err` arm does that. The gap was that it only covers a cluster that fails *fast*. `TransportBuilder` is built with no request timeout, so a cluster that accepts the connection and never answers — a stop-the-world GC pause, a saturated search thread pool, a blackholed route that keeps the socket open — would hold the handler and its axum worker open indefinitely. That is the more common shape of an OpenSearch outage than a refused connection, and before this PR it only touched the fire-and-forget index path. The query is now bounded at three seconds and a timeout falls back to PostgreSQL like any other failure.

I checked the other degradation paths and they were already fine: if the cluster is down at boot, `configure_indexes` fails and `search_service` stays `None`, so the handler never takes the OpenSearch path at all.

## Tests

Your four unit tests plus the nine in the coverage commit are good — asserting the `terms` clause against the wire body through wiremock rather than against the function's source text is a real improvement over what those source-introspection tests were doing.

What was missing was an assertion that the *handler* passes the resolved scope, which is the actually disclosure-critical wiring: `visibility_filter_clause` can be perfect while `quick_search` hands it `AccessScope::Admin`. I added four DB-backed tests driving the endpoint end to end against a wiremock cluster, in `opensearch_quick_search_db_tests`:

- `opensearch_quick_search_hides_repo_the_caller_cannot_read_db` — the cluster returns an artifact in a private repository the caller holds no grant on, with `is_public: true` in the document (the stale shape you described). The endpoint must not serve it.
- `opensearch_quick_search_serves_granted_repo_db` — the control, so the test above cannot pass just because the OpenSearch path is dead.
- `opensearch_quick_search_hides_soft_deleted_artifact_db` — pins the `is_deleted` gap above.
- `opensearch_quick_search_falls_back_to_postgres_on_cluster_error_db` — a 503 degrades to PostgreSQL and returns the row, rather than 500.

## Verified

- `cargo fmt --check`
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(search)'` against a throwaway PostgreSQL with `AK_TESTS_REQUIRE_DB=1`, so the DB-backed tests actually ran rather than silently skipping
- All changes are in existing files, so no new `backend/tests/*.rs` and no CI test-surface gate to update

## Open, and answering your question

You asked whether the reindex-on-visibility-change fix should land first. It should not block this — filtering on `repository_id` routes around it, and the hit vetting now covers the orphan case too — but the underlying defect is real and still unfixed: `RepositoryService::update` and `delete` leave a repository's artifact documents stale and orphaned. Worth its own issue, and it is yours if you want it.

One more I deliberately left: if the cluster is healthy at boot and dies later, `search_service` stays `Some`, so every quick search pays a failed round trip (now bounded at three seconds) before falling back. A circuit breaker is the right answer and is a bigger change than belongs here.

Thank you for this — the write-up on why `is_public` cannot be trusted was the most useful thing in the PR, and it is what told me where else to look.

PR #3682, opened 2026-09-03
fix(helm): advertise relative chart URLs in index.yaml so helm dependency update resolves them

### LINKED ISSUE #3680: fix(helm): index.yaml advertises root-relative chart URLs, breaking `helm dependency update` (doubled path → 404)
### Description

`helm dependency update` cannot download charts hosted in a Helm repository: it 404s with a doubled path. Direct `helm install`/`helm pull` of the same chart works fine.

The root cause is that the generated `index.yaml` advertises chart URLs as **root-relative paths** (leading slash):

```yaml
entries:
  mychart:
    - urls:
        - /helm/helm-local/charts/mychart-0.1.0.tgz
```

Helm's two download code paths disagree on how to resolve that form:

- `helm install` / `helm pull` resolve it against the host root → `https://registry.example.com/helm/helm-local/charts/mychart-0.1.0.tgz` → works.
- `helm dependency update` (the downloader manager used for resolving a parent chart's `dependencies:`) instead joins it onto the repository URL → the path doubles → 404:

```
want: https://registry.example.com/helm/helm-local/charts/mychart-0.1.0.tgz    (curl: 200)
gets: https://registry.example.com/helm/helm-local/helm/helm-local/charts/mychart-0.1.0.tgz    (404)
```

So any chart hosted in Artifact Keeper can be installed directly but cannot be consumed as a subchart dependency of an umbrella chart.

### Steps to Reproduce

1. Create a Helm hosted repo `helm-local` and push any chart (e.g. `mychart-0.1.0.tgz`).
2. `helm repo add ak https://registry.example.com/helm/helm-local && helm install x ak/mychart` → works.
3. Pin the same chart as a dependency of a parent chart:

```yaml
# Chart.yaml of any parent chart
dependencies:
  - name: mychart
    version: "0.1.0"
    repository: "https://registry.example.com/helm/helm-local"
```

4. `helm dependency update` → `Error: could not download ...: failed to fetch ... 404 Not Found` with the doubled path shown above.

Reproduced with helm v3.21.3.

### Expected Behavior

`index.yaml` entries should be downloadable by every Helm code path. The robust form is an **absolute URL**, which is what the codebase already emits for other formats: #1921 rolled the `RequestBaseUrl`-derived external base URL (`AK_EXTERNAL_URL` / `X-Forwarded-Host`+`X-Forwarded-Proto` / `Host`) out to the npm/cargo/nuget/OCI/Git-LFS/wasm handlers, and #2361/#2369 later fixed the same "root-relative URL in metadata" bug for Composer `dist.url`. The Helm handler (`chart_download_url` in `backend/src/api/handlers/helm.rs`) appears to be the remaining format that was missed: it hardcodes `format!("/helm/{repo_key}/charts/{filename}")` with no base URL.

(A plain relative form `charts/{filename}` — the ChartMuseum convention — would also resolve correctly in both code paths, but prefixing `RequestBaseUrl` matches how every other format handler in this codebase already solved this.)

### Environment Information

- Deployment type: Kubernetes (Helm chart), backend behind an ingress that sets `X-Forwarded-*`
- Artifact Keeper version: 1.2.1
- Client: helm v3.21.3

### Additional Context

The virtual-repo and remote-proxy index paths advertise URLs built by the same helper (`merge_upstream_index_entries` → `chart_download_url`), so those surfaces have the same issue.

juwon8891:
Closes #3680

## Summary

`index.yaml` advertised every chart URL as a root-relative path (`/helm/{repo}/charts/{file}`). Helm's two download code paths read that form differently:

- `helm install` / `helm pull` resolve it against the host root → works.
- `helm dependency update` joins it onto the repository URL → the path doubles (`.../helm/{repo}/helm/{repo}/charts/...`) → 404.

So a chart hosted in Artifact Keeper installed fine directly but broke every umbrella chart that pinned it under `dependencies:` (reproduced with helm v3.15.4 and v3.21.3).

The fix makes the advertised URLs absolute by prefixing the `RequestBaseUrl`-derived external base URL (`AK_EXTERNAL_URL` / `X-Forwarded-Host`+`X-Forwarded-Proto` / request authority / `Host`) — the same pattern #1921 rolled out to the npm/cargo/nuget/OCI/Git-LFS/wasm handlers and #2369/#2421 applied to Composer `dist.url`. The Helm handler was the remaining format that was missed.

All three index surfaces advertise through the one helper that gained the prefix (`chart_download_url`), so hosted, remote-proxy and virtual repositories move together:

- `query_charts_from_repo` (hosted/local rows, and hosted members of a virtual repo)
- `merge_upstream_index_entries` (remote-proxy entries, and remote members of a virtual repo)

Deliberately unchanged:
- `resolve_chart_url` (how WE consume an UPSTREAM's index when proxying) still accepts absolute, relative and root-relative forms — third-party indexes are not ours to shape.
- The download route itself (`GET /helm/{repo}/charts/{filename}`) is untouched; only the URLs advertised in `index.yaml` change.

New tests, watched fail before the fix (RED) and pass after (GREEN):

- `test_chart_download_url_is_absolute_3680` — the advertised URL must be an absolute `http(s)://` URL
- `test_chart_download_url_trims_trailing_slash_on_base_3680` — a trailing slash on the base must not produce `//` in the path

Existing tests updated to pin the new absolute form: the three `chart_download_url` unit tests, the `merge_upstream_index_entries` tests, and the two DB-backed round-trip tests (`test_helm_index_url_resolves_for_bare_path_generic_upload`, `test_helm_index_yaml_advertised_url_resolves_prov`) which still prove the advertised URL resolves through the download route.

## Test Checklist

- [x] Integration tests added/updated (if applicable) — the DB-backed index tests above ran against a live PostgreSQL with `AK_TESTS_REQUIRE_DB=1` (helm module: 72/72 green)
- [ ] E2E tests added/updated (if applicable) — N/A, covered by the unit/DB round-trip tests
- [x] Manually tested locally — reproduced the doubled-path 404 with a real parent chart against a live deployment, and verified `helm pull` of the same chart succeeds (the two code paths in question)
- [x] No regressions in existing tests — `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo nextest run --workspace --lib` green except one pre-existing hex test (`test_hex_same_version_case_variants_advertise_one_downloadable_release_db`) that also fails on unmodified `main` in my local DB environment (collation-sensitive; green in upstream CI, untouched by this change)

## API Changes

- [x] N/A - no API changes

No route or schema changes. The `urls` values inside the generated `index.yaml` become absolute (previously root-relative). Clients that resolved the old form from the host root (`helm install`/`pull`) resolve the absolute form identically; `helm dependency update` starts working. A CHANGELOG entry is included under Unreleased → Fixed.


brandonrc:
## Audit round 1 — helm `index.yaml` chart URLs (2 reviewers: correctness / security+integration)

Thanks — the diagnosis is exactly right and the fix genuinely closes #3680. Both reviewers verified it with a real client (helm v3.21.3 against a local server): root-relative → `GET /helm/helm-local/helm/helm-local/charts/…` 404, the issue's doubled path verbatim; your absolute form → both `helm pull` and `helm dependency update` succeed. Helm module: **197/197** (74 in the focused run) against a live Postgres, `cargo check` + `clippy -D warnings` clean, merge with current `main` clean (9 dependabot/docs commits, none touching `backend/src`; #3534 is already an ancestor and disjoint). The CHANGELOG bullet is present and correctly placed. Coverage is complete — all three index surfaces and Artifactory-migrated rows go through the one helper.

### The one change to make: emit a **relative** URL, not an absolute one

Both reviewers independently landed here, and both verified that `charts/{filename}` fixes #3680 on both helm code paths — it is what ChartMuseum emits by default and what the issue itself names as acceptable. The absolute form carries three costs the relative form doesn't:

1. **Helm silently drops credentials on an origin mismatch.** `pkg/getter/httpgetter.go:78` attaches `Authorization` only when the chart URL's scheme+host equal the repo URL's. Measured: repo added as `http://127.0.0.1:8099/...` with `--username/--password`, index advertising `http://localhost:8099/...` → the chart request arrives with **no `Authorization` header**, on both `helm pull` and `helm dependency update`. So a TLS-terminating proxy that forwards `Host` but not `X-Forwarded-Proto` (advertising `http://` for an HTTPS deployment) turns today's *working* private-repo `helm pull` into a 401, and a proxy that rewrites `Host` advertises an unreachable origin. The fix can regress the path that wasn't broken.
2. **The origin comes from an unvalidated `X-Forwarded-Host`.** `index_yaml` takes `RequestBaseUrl` (`helm.rs:523`), whose extractor prefers `X-Forwarded-Host`/`-Proto` with no trusted-proxy gate. The index is served with **no `Cache-Control` and no `Vary`** (`helm.rs:395-399`), so a forged header lands in a heuristically-cacheable document. Verified the chain end-to-end: a cross-origin absolute `urls:` entry is followed silently, **and helm does not check the index `digest:` against the downloaded bytes** — a chart with sha256 `97f8…` served under an entry advertising `624d…` was written, `description: TAMPERED`. The repo already has `request_base_url_from_host_header` (`extractors.rs:228-253`) built for exactly this ("client-fetchable URLs embedded in cacheable response metadata") — but with the relative form none of this applies.
3. Tests pass the base URL in as a literal, so nothing pins where it comes from.

So: `format!("charts/{}", filename)` (or the equivalent in the shared helper), keep both new tests (they'd assert the relative form), and the CHANGELOG bullet gains an upgrade note that clients holding a cached index and operators who worked around the bug with a rewrite rule will see the URL form change. If you have a reason to prefer absolute, then it needs `request_base_url_from_host_header`, `Cache-Control: private`, and the proxy-config caveat documented — but relative is strictly simpler.

### Pre-existing, for a follow-up
`resolve_chart_url` (`helm.rs:632-640`) — the *consumption* side — joins a leading-slash upstream URL onto a subpath `upstream_url`, so an AK Remote pointed at an un-upgraded AK still doubles; the existing test only covers a host-root upstream. Not this PR's.

Milestoned 1.9.0. Happy to push the relative-form change and the changelog note to your branch if you'd rather — say the word.

juwon8891:
Thanks for the depth here — particularly for measuring the credential behaviour rather than reasoning about it. Taking the relative form; pushed as b3ea364b.

**On (1), which is the one that changed my mind.** I had treated "absolute" as the settled house style because #1921/#2361 went that way for the other formats, and I missed that Helm's getter makes the origin load-bearing in a way npm/cargo/composer's don't: `httpgetter.go` gates `Authorization` on scheme+host equality, so the advertised origin isn't cosmetic — it decides whether credentials are sent at all. That makes the failure mode strictly worse than the bug I was fixing: #3680 breaks `dependency update`, which never worked, whereas a proxy that forwards `Host` without `X-Forwarded-Proto` would have broken `helm pull` against a private repo, which does. Fixing a broken path by regressing a working one isn't a trade I'd have taken knowingly, and the relative form doesn't ask me to.

(2) is the better argument in the abstract and I don't dispute it, but it's (1) that makes absolute wrong here even for a correctly-configured deployment.

**What changed:**

- `chart_download_url` now returns `format!("charts/{}", filename)`. With no origin to derive, the `RequestBaseUrl` extractor and the `base_url` threading through `index_yaml` → `query_charts_from_repo` / `merge_upstream_index_entries` all come out — and `query_charts_from_repo`'s `repo_key` parameter became dead with them, so it's gone too. Net **−67 lines** against the previous revision; the diff is now smaller than the one you reviewed.
- Both new tests kept, asserting the relative form. `test_chart_download_url_is_relative_3680` pins no leading slash and no scheme; the trailing-slash test had no subject left once the base URL went away, so it became `test_chart_download_url_joins_onto_a_subpath_repo_url_3680`, which performs the join `dependency update` performs (`{repo_url}/{advertised}` against a `/helm/{repo}` subpath) and asserts the result isn't the doubled path. That encodes the reported bug directly rather than inferring it from the URL's shape.
- CHANGELOG bullet rewritten for the relative form, with the rationale for not going absolute and the upgrade note you asked for: the `urls:` form changes, clients resolve on read so nothing needs republishing, but a cached index keeps the old form until `helm repo update`, and any proxy rewrite rule someone added to work around this should come out since the path it rewrites is no longer emitted.

Local: helm module **74/74** against a live Postgres (`AK_TESTS_REQUIRE_DB=1`), `cargo fmt --check` and `clippy --all-targets -D warnings` clean.

**On the follow-up:** filed as #3705 with the subpath-upstream reproduction — the existing `test_resolve_chart_url_relative_leading_slash` only covers a host-root upstream, which is why the AK-Remote-pointed-at-AK case slipped through. Kept out of this PR as you scoped it.

One note on hygiene: the branch is still named `fix/helm-index-absolute-urls`, which is now wrong. I left it rather than force a new PR; the title is updated. Happy to rename if you'd prefer the branch match.

brandonrc:
## Round 2 — approved, merging

Exactly the change round 1 asked for. Verified on `b3ea364b`:

- `chart_download_url` now emits `charts/{filename}` — no slash, scheme or host — and its two callers cover hosted, remote-proxy and virtual indexes; Artifactory-migrated rows go through the same query. The `RequestBaseUrl` import, the `index_yaml` parameter and every `base_url` plumbing line are gone; a grep of the file and every added line for `RequestBaseUrl|base_url|X-Forwarded|host_header|Host` finds only prose.
- Live helm 3.21.3 against an auth-requiring server: relative → `helm pull` **and** `helm dependency update` both succeed with credentials attached; `origin/main` → the doubled-path 404 verbatim; the round-1 absolute form with a mismatched authority → both commands 401 with no credentials. The round-1 credential-drop finding is eliminated by construction, not argued away.
- Tests: 71 → 73, the two round-1 tests rewritten in place; the new subpath-join test performs the actual `dependency update` join. Helm selector 74/74 against live Postgres, `check` and `clippy -D warnings` clean, merge with main clean.
- CHANGELOG bullet is in house style with the upgrade note covering cached indexes and operator rewrite rules.

Three stale comments remain (`helm.rs` ~1863-1867 says the URL is "absolute, so that `helm dependency update` does not join it" — the join is now the mechanism; ~1870-1872 and ~2442-2444 still explain the deleted `localhost` fallback; `test_helm_chart_url` ~1666 still asserts the root-relative form against itself). None worth another round; I'll tidy them in a follow-up.

Squash-merging with the CHANGELOG bullet as the commit body, since the PR description still describes the round-1 absolute-URL approach. Thank you for the quick turnaround on this.

PR #3696, opened 2026-09-03
fix(security): rebuild grype from v0.118.0 and override grpc to v1.83.1

### LINKED ISSUE #3655: docker: grype v0.118.0 carries x/mod v0.40.0 — bump GRYPE_VERSION and delete the #3465 override
## Summary

The `golang.org/x/mod` override in `docker/Dockerfile.backend` (the `XMOD_FROM=v0.38.0 -> XMOD_TO=v0.40.0` block added in #3465) documents its own removal condition: *"check the RELEASE … before adding one here"* — lever 1 (bump `GRYPE_VERSION`) was unavailable only because `v0.118.0` did not exist and upstream `main` still pinned v0.38.0.

That is no longer true. anchore/grype **v0.118.0** (commit `756eb9a24f7beeafb6871a24e943e8a3ae210695`) pins `golang.org/x/mod v0.40.0` in its own `go.mod`:

```
gh api repos/anchore/grype/contents/go.mod?ref=v0.118.0 --jq .content | base64 -d | grep golang.org/x/mod
# golang.org/x/mod v0.40.0
```

So the override now duplicates a shipped release, which the block itself calls *"strictly worse than the bump it imitates"*.

## Change

In `docker/Dockerfile.backend` (and the matching stage in `Dockerfile.backend.alpine` if it carries the same args):

- `GRYPE_VERSION=0.117.0` → `0.118.0`, `GRYPE_COMMIT` → `756eb9a24f7beeafb6871a24e943e8a3ae210695`
- delete the `XMOD_FROM`/`XMOD_TO` args and the `go mod edit -replace` override block; the build should again be byte-for-byte upstream's module graph
- re-verify the Trivy scan of the built image is clean without the override (the two x/mod advisories the block cites were the reason it exists)
- while there: the stage comment still says *"`golang:1.26` floats to the newest 1.26.x"* but #3579 moved the stage to `golang:1.27-alpine`; `GO_MIN_PATCH=1.26.6` is still satisfied (the floor is `sort -V`) but the comment and floor should describe the toolchain actually in use.

Not folded into #2883 because it changes the scanner binary we ship and needs the image scan re-run to prove the override is no longer load-bearing; the version bumps in #2883 deliberately left the grype stage untouched.

Refs: #3465 (the override), #3352 / #3307 (the other two levers), #2883.

brandonrc:
## Summary

`Security Scan` in Docker Publish is failing on the backend image and blocking every publish from `main`: `usr/local/bin/grype` (gobinary) carries `google.golang.org/grpc v1.82.1`, and **CVE-2026-84304** is HIGH with a fix in 1.83.1. Release Preflight reads NOT READY because of it.

This is a bump *and* an override, for two different modules.

**Bump — `GRYPE_VERSION` 0.117.0 → 0.118.0**, `GRYPE_COMMIT` → `756eb9a24f7beeafb6871a24e943e8a3ae210695` (the commit `refs/tags/v0.118.0` dereferences to; the ref is an annotated tag object `cd00e1ad`, and `git rev-parse HEAD` after a `--branch v0.118.0` clone gives the commit, which is what the stage asserts). v0.118.0's own `go.mod` pins `golang.org/x/mod v0.40.0`, so the **#3465 x/mod override is deleted** — `XMOD_FROM`/`XMOD_TO` and the block. That is the override reaching its documented end rather than being noticed by hand: it had become a duplicate of a shipped release ("strictly worse than the bump it imitates", in its own words), and its staleness guard is what stops the build on this bump until it is removed. The policy text it carried — the three levers, "check the RELEASE, not just `main`" — is kept, because it still governs the next override.

**Override — `GRPC_FROM=v1.83.0` → `GRPC_TO=v1.83.1`.** v0.118.0 pins grpc v1.83.0, one patch short of the fix, so the grpc block takes the x/mod block's place in the same shape and with the same rigor:

- `go get`, not `go mod edit -require` — the reasoning the deleted block spelled out still applies: `go get` resolves through MVS and writes checksum-verified `go.sum` entries in the same step, and a `replace` arrow would record the *pre*-override version in the build info a scanner reads. (The brief suggested `go mod edit -require` + `go mod tidy`; I kept `go get`, since "the same shape as the block you are deleting" and that block's own argument both point at it, and `tidy` would cost a full test-dependency download for a no-op.)
- a guard that the pinned source still requires `GRPC_FROM`, failing the build with "DELETE the override block" when it does not;
- an assertion that MVS moves **exactly** the declared module set — here one module, versus x/mod's five;
- `go mod verify`;
- a post-build assertion that the **compiled binary's Go build info** records `GRPC_TO`. A bumped module graph that is not linked into the artifact clears nothing, and the build info is exactly what Trivy reads.

All three of the inherited override-policy conditions hold, which is the difference from #3465: the finding is a gating HIGH with an available fix; lever 1 was checked against the *release* and taken, and still does not reach it; and upstream has already made the same bump — `anchore/grype` `main` is on grpc v1.83.1 as of `0f37e38` (2026-09-02), anchore/grype#3686, merged after v0.118.0 was cut. Unlike x/mod, whose advisories were scoped to `x/mod/sumdb` and were never linked, **grpc is genuinely linked into `cmd/grype`** (`grype/internal/file` → `hashicorp/go-getter` → `cloud.google.com/go/storage`, ~40 grpc packages in `go list -deps`).

**`GO_MIN_PATCH` 1.26.6 → 1.27.1.** #3579 moved this stage to `golang:1.27-alpine` but left the floor on the 1.26 line; because the check is a `sort -V`, every 1.27.x cleared it and the lagging-mirror guard was inert. go1.27.1 (2026-09-01) is the newest 1.27 patch and is what the image serves.

**`.trivyignore`: the five `github.com/docker/docker` suppressions stay live.** The retry that file asks for on every scanner bump was run and still fails — v0.118.0 pins `v28.5.2+incompatible`, unchanged, and `go mod edit -require=github.com/docker/docker@v29.3.1+incompatible` still fails with `unknown revision v29.3.1`. Nothing is tombstoned. The version labels move to v0.118.0 / go1.27.1, which `config::tests::grype_source_build_is_pinned_consistently_across_images` requires.

`Dockerfile.backend.alpine` carries the same stage and moves identically; all four grype `RUN` blocks are byte-identical between the two files (checked programmatically).

## What I ran

All builds through `rig/compile-lock.sh`.

1. **Upstream facts.** `gh api repos/anchore/grype/git/ref/tags/v0.118.0` → annotated tag `cd00e1ad…`, dereferencing to commit **`756eb9a2…`**. `gh api .../contents/go.mod?ref=v0.118.0` → `golang.org/x/mod v0.40.0`, `google.golang.org/grpc v1.83.0`, `github.com/docker/docker v28.5.2+incompatible`. `?ref=main` → grpc **v1.83.1**. Latest release is v0.118.0 (2026-08-27).

2. **Built the grype stage** (`docker build --target grype-bin -f docker/Dockerfile.backend`). Exit 0. Stage output:
   ```
   grype build: go1.27.1 satisfies the go1.27.1 floor
   grype build: override applied — google.golang.org/grpc v1.83.0 -> v1.83.1
   grype build: build info verified — google.golang.org/grpc v1.83.1 across 296 deps
   ```

3. **Inspected the built binary.** `go version -m` → `go1.27.1`; `dep google.golang.org/grpc v1.83.1`, `dep golang.org/x/mod v0.40.0` (upstream's own pin, no override), `dep github.com/docker/docker v28.5.2+incompatible`. `grype --version` → `grype 0.118.0`.

4. **Trivy scan, mirroring the gate's flags** (`docker-publish.yml` pass 2: trivy **0.74.0**, `--severity CRITICAL,HIGH --ignore-unfixed --ignorefile .trivyignore --exit-code 1`). I built the **grype stage only** and scanned the extracted binary as a **rootfs target** (path `usr/local/bin/grype`, the same path the gate reports), not a full backend image — the full image build was too heavy here.

   **Before** (built from `origin/main`'s Dockerfile, grype v0.117.0 + the x/mod override) — **exit 1**:
   ```
   usr/local/bin/grype (gobinary)  Total: 1 (HIGH: 1, CRITICAL: 0)
   google.golang.org/grpc  CVE-2026-84304  HIGH  fixed  v1.82.1 -> 1.83.1
   ```
   **After** (this branch) — **exit 0**:
   ```
   ┌─────────────────────┬──────────┬─────────────────┐
   │       Target        │   Type   │ Vulnerabilities │
   ├─────────────────────┼──────────┼─────────────────┤
   │ usr/local/bin/grype │ gobinary │        0        │
   └─────────────────────┴──────────┴─────────────────┘
   ```

5. **Full picture of the new binary**, no ignorefile, all severities: 2 HIGH unfixed + 1 MEDIUM fixed + 1 MEDIUM unfixed against `github.com/docker/docker` (the suppressed set), and `CVE-2026-56855` / `CVE-2026-78662` / `GO-2026-5932` against `golang.org/x/crypto v0.55.0` at **UNKNOWN** severity. None gates (`--severity CRITICAL,HIGH` excludes UNKNOWN, `--ignore-unfixed` drops the rest). Noted rather than acted on — out of scope for this issue.

6. `cargo nextest run -p artifact-keeper-backend --lib -E 'test(/grype/) or test(/trivy/) or test(/dockerfile/)'` → **149 passed**, including `grype_source_build_is_pinned_consistently_across_images`, `toolchain_fixed_grype_stdlib_cves_are_not_suppressed` and `no_dockerfile_pulls_a_prebuilt_grype_binary`.

7. **Retried the docker/docker override** on the v0.118.0 tree: `go mod edit -require=github.com/docker/docker@v29.3.1+incompatible` → `unknown revision v29.3.1`. Unchanged from the 2026-08-16 attempt.

### Not run / caveats, honestly

- **No full backend image build or image-target scan.** The grype stage was built and its binary scanned as a filesystem target. The other layers of the backend image are untouched by this change, but the gate scans the assembled image and I did not reproduce that.
- **`Dockerfile.backend.alpine` was not built** (alpine builds are suspended in `docker-publish.yml`). Its grype `RUN` blocks are byte-identical to the UBI file's, which I verified programmatically, and the `ARG` values match — but they were not executed.
- **`scripts/ci/release-preflight.sh` check 1 was not run**; it needs the active `release/*` branches. No suppression token was added or removed, so its ledger is unchanged.
- **`CVE-2026-34040`** (the one HIGH+fixed docker/docker token) does **not** appear against either the before or after binary with trivy 0.74.0's current DB. I did **not** tombstone it: the brief scopes the docker/docker decision to whether the 29.3.1 override resolves (it does not), and a DB-side change is not the measured basis the file's ledger convention asks for. Worth a look on the next scanner bump.

Closes #3655


brandonrc:
## Audit — 2 reviewers (supply-chain / integration+CI)

**Verdict: merge once `🐳 Build Backend Image` is green on the follow-up head.** Both reviewers rebuilt the `grype-bin` stage from `origin/main` and from this branch under the same flags Docker Publish's gate uses (trivy 0.74.0, fresh DB, `--severity CRITICAL,HIGH --ignore-unfixed --ignorefile .trivyignore`): **main → exit 1, CVE-2026-84304 HIGH; branch → exit 0**. Verified independently: `GRYPE_COMMIT` is what annotated tag `v0.118.0` (`cd00e1ad…`) dereferences to; `go mod verify` prints `all modules verified` with `GOSUMDB=sum.golang.org` and no `GONOSUMDB`/`GOFLAGS`/`GOPRIVATE`; the "exactly one module moved" assertion is real (grpc-go's `go.mod` is byte-identical between 1.83.0 and 1.83.1 and grype already pins the three companions at grpc's required versions — **mutation-tested**: downgrading two companions before the override fails the build and names both); the post-build buildinfo check is live (changing the expected literal to `v1.83.2` fails the stage); go1.27.1 is the current patch; the six grype `RUN` blocks are byte-identical between the UBI and alpine files; preflight check 1 sees no token identity change (the v0.117.0→v0.118.0 relabel is comment prose); 149 grype/trivy/dockerfile tests + `workflow_scan_gate_tests` + the three CI shell checks all pass; merge with main clean.

**What this PR can and cannot prove.** `Build Backend Image` builds the stage (path filter matches, ARG changes bust the cache) and runs the four in-stage assertions, but `Security Scan` runs only on push to `main` — so the gate itself is proven here only by the local reproduction above. And the integration reviewer is right that **merging this alone does not unblock Docker Publish**: the Scanner Adapter arm fails on the identical CVE against its bundled `trivy`, pinned by digest from `artifact-keeper/trivy`; that fix is artifact-keeper/trivy#12 (open, green), followed by a digest repoint + `VERSION` 1.2.9 here. The Dockerfile comment and CHANGELOG lead have been reworded to say so.

Folded in as the follow-up commit: `GRPC_FROM`/`GRPC_TO` added to the cross-Dockerfile parity test; `.trivyignore`'s docker/docker block records the 2026-09-03 re-measurement (no token produces a finding today — kept live because that is a DB-side absence, and CVE-2026-34040 is live on all three `release/*` files; both reviewers agreed with not tombstoning); the go-git note says "present at the fixed version" rather than "absent". Filed **#3698** for the next one in line: `x/crypto v0.55.0` (CVE-2026-56855/-78662, fixed in 0.56.0, UNKNOWN severity today, and anchore/grype main still pins 0.55.0).

Issue #3683, opened 2026-09-03
Service Accounts cant push artifacts


major-sam:
### Description

Service accounts cant push artifacts to any(? tested on docker & nuget & generic) repo type

### Environment Information

- Deployment type:  docker
- Operating System:  debian 13
- Architecture: x86_64
- Database:  postgre
- Reverse Proxy: caddy
- Network setup:  docker network

### Backend Version

1.8.2

### Web UI Version

1.8.0

### Package Format

generic, docker ... 

### Expected Behavior

Service account must be able to use it's token with configured permissions

### Observed Behavior

Service account token permissons not work at all (admin or rwx permissons)

**Docker:**
``` 
docker login artifact.keeper.local
Username: svc-sa 
Password: #<<< SERVICE_ACCOUNT_TOKENt
Login Succeeded 
docker push  artifact.keeper.local/docker/my-image:latest
denied: You do not have access to this repository
```

**Generic:**

```
curl -X PUT -H "Authorization: Bearer SERVICE_ACCOUNT_TOKEN" -T .\test.json   https://artifact.keeper.local/api/v1/repositories/test/artifacts/test.json
{"code":"FORBIDDEN","message":"You do not have access to this repository"}
```

### Steps to Reproduce

1. setup AK with Ldap auth
2. create docker or generic repository 
3. create service account
4. create service account token with rwx or admin access (with or w/o repo scope)
5. try to push anything to your repo

### Logs

```shell
2026-09-03T08:12:45.377434Z DEBUG http_request{method=PUT uri=/api/v1/repositories/test/artifacts/test.json}: tower_http::trace::on_request: started processing request
2026-09-03T08:12:45.378389Z DEBUG http_request{method=PUT uri=/api/v1/repositories/test/artifacts/test.json correlation_id=ffb2d735-0cfd-4921-b591-55fce72439f0}: sqlx::query: summary="SELECT at.id, at.token_hash, at.user_id, …" db.statement="\n\n\n            SELECT at.id, at.token_hash, at.user_id, at.scopes, at.expires_at,\n                   at.repo_selector, at.revoked_at, at.last_used_at\n            FROM api_tokens at\n            WHERE at.token_prefix = $1\n            \n" rows_affected=1 rows_returned=1 elapsed=472.474µs elapsed_secs=0.000472474
[...]

PR #3700, opened 2026-09-03
fix(projects): accept service_account as a project-member principal

brandonrc:
## Summary

`POST /api/v1/projects/{id}/members` refused `principal_type: "service_account"` outright — `valid_principal_type` (`backend/src/api/handlers/projects.rs`) matched only `"user" | "group"`, although its own doc comment claimed to match the principal domain `PermissionService::query_actions` resolves, which has been `principal_type IN ('user','service_account')` since #2499/#2433. The read plane resolved a `service_account` grant on a `project` target; the write plane refused to create one.

A service account is created with no role and no grant, and token scopes only *narrow* — they never confer. With the project endpoint closed and the shipped web UI able to emit only `principal_type: "user"` (rejected for a service-account row by the #2503 correspondence guard; listing half fixed in #3634/#3635), an operator who manages access by project had **no** grant channel at all, and every push failed the repository gate exactly as reported.

**Enforcement was never the defect.** The 403 in the reporter's log is `user_can_access_repo` correctly answering "no" because no row named the account. This is also not a 1.8.x regression — the gap dates to #2483 (v1.6.0).

The fix is one line plus its now-false doc comment.

## The #2503 mistype guard is untouched

`add_project_member` calls `PermissionService::validate_principal` immediately after the type allowlist, so the `principal_type`/`principal_id` correspondence is still checked against the database. `service_account` naming a person's id is still a 400, and `user` naming a service account's id is still a 400. Widening the allowlist only stops refusing a grant the resolver already honours; it changes nothing about resolution.

## Also worth knowing (no code change)

The "admin permissions don't work either" reports on the issue are a *different, deliberate* behaviour: a token minted without the `admin` (or `*`) scope is demoted to a non-admin principal by `with_scope_gated_admin` even when the account is an admin — GHSA-vvc3-h39c-mrq5's scope gate working as designed. The CHANGELOG bullet says so explicitly so operators stop reading it as the same bug.

## Changes

- `backend/src/api/handlers/projects.rs` — `valid_principal_type` accepts `service_account`; doc comment corrected; validation-error message and the `AddProjectMemberRequest.principal_type` schema doc list the three types.
- Flipped `test_valid_principal_type_user_and_group` / `test_invalid_principal_types_rejected`, which pinned the rejection.
- New DB-backed regression test.
- `CHANGELOG.md` — one bullet under `### Fixed`, including the pre-upgrade curl workaround and the GHSA-vvc3 note.

## Regression test

`api::handlers::projects::tests::db::test_service_account_project_member_can_write_to_project_repo`

Drives the reporter's shape end to end: seeds a service account, a private generic repository, and a project holding that repository; asserts the SA token is **denied** by `require_repo_write_access` first; adds the SA as a project member through `add_project_member`; then asserts the same token clears `require_repo_write_access` **and** the canonical `check_repository_action(.., "write", ..)`. It also pins the negative case — a `service_account`-typed grant naming a plain user id is still a 400 — so the #2503 guard cannot be regressed silently.

**Proof it fails without the change** (one-line `matches!` reverted, test kept):

```
FAIL api::handlers::projects::tests::db::test_service_account_project_member_can_write_to_project_repo
  #3683: POST /projects/{id}/members must accept principal_type = 'service_account'
  (got Err(Validation("Invalid principal_type 'service_account': must be 'user', 'service_account' or 'group'")))
FAIL api::handlers::projects::tests::test_valid_principal_type_user_and_group
  assertion failed: valid_principal_type("service_account")
Summary: 2 tests run: 0 passed, 2 failed
```

## Verified

Postgres 16 (throwaway container, `backend/migrations` 001-211 applied), `AK_TESTS_REQUIRE_DB=1`.

- `cargo fmt` — clean.
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(/api::handlers::projects::/) or test(/api::handlers::permissions::/) or test(/services::permission_service::/) or test(service_account)'` — **253 passed, 0 failed** (includes all 32 `projects::` tests and `repository_service::tests::db::test_user_can_access_repo_service_account_grant_honored`).
- Counterfactual as quoted above.

**Not run:** the full `--lib` suite, integration/e2e tests, and anything requiring a live registry or the web UI. The web-side half — the permissions form still submitting `principal_type: "user"` — is in `artifact-keeper-web` and is out of scope here.

Part of #3683 — closes the project-membership grant channel; the reporter's own repro (no project) is served by `POST /api/v1/permissions` with `principal_type: "service_account"`, and the web UI half is artifact-keeper-web.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Consumers still have to opt in. In artifact-keeper-web five admin screens read this endpoint — the permissions principal picker, the users list, the group member picker, the audit actor filter and the age-gate approver picker — and four of them have no other source for the distinction; the picker is fixed page-locally in artifact-keeper-web#823, the rest wait on an SDK release carrying this field.
 
+- **A service account could not be granted access through a project, leaving project-managed deployments with no grant channel for it at all** (#3683). `POST /api/v1/projects/{id}/members` refused `principal_type: "service_account"` outright — `valid_principal_type` accepted only `user | group`, even though its own doc comment claimed to match the principal domain `PermissionService::query_actions` resolves, which has matched `principal_type IN ('user','service_account')` since #2499/#2433. So the read plane happily resolved a `service_account` grant on a `project` target while the write plane refused to create one. A service account is created with no role and no grant, and token scopes only ever *narrow* — they never confer — so with the project endpoint closed and the shipped web UI able to emit only `principal_type: "user"` (rejected for a service-account row by the #2503 correspondence guard, listing half fixed in #3634/#3635), an operator who manages access by project had no way to give the account anything. Every push then failed the repository gate exactly as reported: `denied: You do not have access to this repository` from `docker push`, `403 FORBIDDEN` from a generic `PUT`. Enforcement was never the defect — the deny was `user_can_access_repo` correctly answering "no" because no row named the account. `valid_principal_type` now accepts `service_account` as well. This is not a 1.8.x regression: the gap dates to #2483 (v1.6.0).
+
+  The #2503 mistype guard is untouched. `add_project_member` calls `PermissionService::validate_principal` immediately after the type allowlist, so the `principal_type`/`principal_id` correspondence is still checked against the database: `service_account` naming a person's id is still a 400, and so is `user` naming a service account's id. Widening the allowlist only stops refusing a grant the resolver already honours.
+
+  **Workaround without upgrading**, on any 1.6+ build — grant the service account directly, with the right principal type, on the endpoint that already accepts it:
+
+  ```
+  curl -X POST $AK/api/v1/permissions \
+    -H "Authorization: Bearer <admin token>" -H 'Content-Type: application/json' \
+    -d '{"principal_type":"service_account","principal_id":"<sa uuid>",
+         "target_type":"repository","target_id":"<repo uuid>","actions":["read","write"]}'
+  ```
+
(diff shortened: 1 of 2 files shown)


brandonrc:
## Audit — 1 reviewer (authorization consequences)

**Verdict: merge on green.** Verified rather than read: `add_project_member` calls `validate_principal` at `projects.rs:500-504`, immediately after the allowlist and before `require_project_exists`, so a `service_account` naming a user/group/nonexistent id is still a 400 — pinned in both directions (new test + the pre-existing `test_add_project_member_rejects_mistyped_principal`). A service-account membership resolves through exactly the predicates a user membership does (`query_actions`, `check_repository_action`, `has_any_rules_for_target`, `permissions_grant_exists_for` all match `principal_type IN ('user','service_account')` on the row's `users.id`), nothing downstream branches on `is_service_account`, and the membership cannot exceed the token: scopes only narrow, `require_repo_write_access` goes through `require_repo_access` first, and `with_scope_gated_admin` (`auth.rs:246-249`) is untouched and described accurately in the CHANGELOG. `list_project_members` shows the row and `remove_project_member` deletes it (both statements run against a seeded SA row). Counterfactual reproduced: reverting the `matches!` line fails exactly the 2 tests the PR names. 253/253 on the projects/permissions/service-account selector against a live DB; merge with current main clean.

Two bookkeeping corrections, applied: **this does not close #3683** — the reporter's own repro involves no project, so their channel is `POST /api/v1/permissions` with `principal_type: "service_account"` (working since v1.6.0) and the UI half lives in `artifact-keeper-web`; the PR body now says "Part of #3683" and the issue stays open for that. And the `regression` label on #3683 was wrong — `valid_principal_type` dates to v1.6.0 (#2483) and `git log v1.8.0..v1.8.2` over the three relevant files is empty; removed. Plus the test rename the reviewer asked for (`d73bfb8f`).

Issue #3681, opened 2026-09-03
Group-principal permissions are never enforced — members denied access regardless of target type


shreyankeg:
### Description

Permissions where the **Principal is a Group** are not enforced at all, for any target type. Members of a group with a granted permission are denied access both in the web UI (repo not visible) and via CLI/registry auth (access denied), while the identical grant made directly to the **User** works immediately.

### Environment Information

  - Server: 1.8.1
  - Web: 1.8.0

### Backend Version

1.8.1

### Web UI Version

1.8.0

### Package Format

None

### Expected Behavior

Granting a permission to a Group principal should extend that permission to every member of the group, identically to how a direct User grant works. Concretely, after step 2 above, `U` should see `R` in the UI and be able to push/pull it via CLI immediately after (or after re-login, if group claims are  session-cached) — without needing a separate direct User-level grant.

### Observed Behavior

`U` is denied access entirely: `R` does not appear in the UI's repository list for `U`, and CLI push/pull against `R` returns an access-denied error — even though `U` is a confirmed member of `G` and `G` holds an explicit read+write grant on `R`. The permission row exists (visible in Security → Permissions) but
  has no effect for any member of the group. Only a direct User→Repository grant produces the expected access.

### Steps to Reproduce

1. Create Group `G`, add user `U` as a member (confirmed via Admin → Groups → `G` → Manage Members, `U` listed).
  2. Create Permission: Principal = Group `G`, Target = Repository `R` (a physical, non-virtual repo), actions = read+write.
  3. `U` logs out and back in via SSO (fresh session/token).
  4. Result: `R` does not appear in the UI for `U`. CLI push/pull against `R` as `U` is denied.
  5. Control: delete the group grant, instead create Permission: Principal = User `U`, Target = Repository `R`, same actions. No other change (same repo, same user, fresh login not even required).
     Result: `R` appears in the UI for `U`, CLI push/pull succeeds immediately.
  6. Also separately tested: Principal = Group `G`, Target = Group `V` (a group-type/virtual aggregating repo) — same failure mode as step 4.

### Logs

```shell

```

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have redacted any sensitive information (passwords, tokens, private URLs)
- [x] I am using a supported version (not end-of-life)

irizzant:
Thanks for the detailed report — this is an important one for anyone evaluating group-based access control.

[...]

Issue #3674, opened 2026-09-02
sqlx offline cache is split across backend/.sqlx and the workspace-root .sqlx, and neither directory is complete


brandonrc:
## Summary

Found while auditing #3671. The repo carries two sqlx offline caches: `backend/.sqlx` and the workspace-root `.sqlx`. sqlx 0.9's macros read `CARGO_MANIFEST_DIR/.sqlx` first and fall back to `workspace_root/.sqlx` (`sqlx-macros-core-0.9.0/src/query/mod.rs:95-118`), so **both are load-bearing** — and neither is complete on its own:

- 443 entries are byte-identical in both directories.
- `backend/.sqlx` has no entry for the two `users.rs` listing queries (`:272`, `:308`); root `.sqlx` has none for `artifact_service.rs:1761`. Moving root `.sqlx` aside makes `SQLX_OFFLINE=true cargo check --lib` fail with 8 `no cached data` errors.
- 26 committed entries are dead (no query on `main` hashes to them).
- `ARCHITECTURE.md:115` and `:196` disagree about which directory exists.

Consequences today: anyone who changes a `query!` string has to know to update the right one of two directories (the #3671 author had to hand-edit 8 files after discovering this), `cargo sqlx prepare --check` is not run in CI and could not pass (it compares one directory, and a fresh `prepare` on sqlx 0.9 adds an `"origin"` key to 318 files that no committed entry has), and a stale entry in the shadowing directory silently wins.

## Ask

A pure file-move consolidation — copy the 19 `backend/`-only entries into root `.sqlx`, delete `backend/.sqlx`, drop the 26 dead entries — with no regeneration, so the diff is reviewable. Then fix the `ARCHITECTURE.md` reference and decide whether to accept the one-time `origin`-key churn to make `cargo sqlx prepare --check` a CI gate (it is the only thing that catches a query edited without its cache entry).

Refs: #3671, #3557.

PR #3701, opened 2026-09-03
fix(search): resolve visible repositories from permission grants too

### LINKED ISSUE #3697: Global search resolves the caller's visible repositories from role_assignments only — permission grants (user and group) are invisible to all /api/v1/search* endpoints
## Summary

`search.rs:249-262` (`resolve_visible_repos`) builds the caller's visible-repository set from `role_assignments` alone. The `permissions` table — the target of `POST /api/v1/permissions`, for both `user` and `group` principals — is never consulted, so a user whose *only* access to a private repository is a permission grant gets an empty result from every `/api/v1/search*` endpoint while `GET /api/v1/repositories` and direct downloads work.

Present since v1.2.0-rc.1 (#829), i.e. it predates `permissions` becoming a visibility source (#2049, v1.2.3), and was never brought in line. Found while investigating #3681, which it does not explain (that report's direct-user control works, and search denies those identically).

A DB-backed repro (`repro_3681_group_grant_is_invisible_to_search` — user in a group, group granted read on a private repo, search returns nothing) fails on `main`; see the #3681 thread for the surrounding investigation.

## Fix

OR the existing `pub(crate) permissions_grant_exists_for` fragment (`repository_service.rs:433-452`) into `resolve_visible_repos`. Because it is the same fragment the repository read gate uses, this cannot widen search beyond what the read gate already allows; API-token repository scope stays layered on top separately. Add the repro as the regression test, plus a direct-user-grant sibling.

Refs: #3681, #2049, #3522.

brandonrc:
## Summary

`resolve_visible_repos` (`backend/src/api/handlers/search.rs`) — the accessible-repository set behind **every** `/api/v1/search*` endpoint — built the caller's visible set from `role_assignments` alone. The `permissions` table was never consulted, so a user whose only access to a private repository came from a grant (group-principal, direct-user, service-account, or inherited from the owning project) got an empty result set from search while `GET /api/v1/repositories` listed the repository and direct downloads by path worked.

Present since v1.2.0-rc.1 (#829): the resolver was correct when written and was never updated when `permissions` became a second visibility source in v1.2.3 (#2049). Surfaced while investigating #3681, which it does not otherwise explain, since search denies a direct-user grant identically.

> **Revised after review.** The first cut of this PR ORed in the shared `permissions_grant_exists_for` fragment verbatim. Both reviewers correctly showed that fragment is only the *tenant* half of the read gate, and that reusing it made a `{write}`-only grant enumerate a private repository's artifact inventory. This revision adds the read-action term and rewrites the arm set-driven. Details below.

### 1. The arm carries the `read` action, not just the tenant term

`permissions_grant_exists_for` gates on `p.actions <> '{}'` — *any* non-empty grant. The read gate is that fragment **AND** `check_repository_action(.., "read", ..)`, and `write` does not imply `read` (`permission_service.rs:214-220`). Search returns a private repository's artifact names, paths, versions, sizes and — via `/search/checksum` — sha256 digests, so the bare tenant term let a publish-only CI principal enumerate all of it while `GET /api/v1/artifacts/{id}` still answered 404. That is exactly what `RepoAccess::TenantOnly`'s contract forbids fronting, and what the #3331 structural guard exists to prevent; the first cut evaded that guard only by hand-composing the fragment instead of calling `user_can_access_repo`.

The arm now requires a grant carrying `read` or `admin`, which makes it a **subset** of the read gate rather than a superset: such a rule satisfies the tenant half (its actions are non-empty) and forces `check_repository_action`'s CASE down the `applicable_rules` branch, which the same rule then satisfies.

**Honest scoping — where search now sits.** This is containment, not equality. `GET /api/v1/repositories` (`build_grant_predicate`) and the resolver's pre-existing `role_assignments` arm are both tenant-only and action-blind, and **neither is changed here**. So search is now:

* **narrower than the repository listing** — a `{write}`-only grant lists the repository but does not search it;
* **still wider than the read gate** — the entire remaining gap is that untouched `role_assignments` arm, so a user holding a repo-scoped role that carries no `read` can still search the repository, exactly as before this release.

Closing that last gap would narrow existing behaviour on a predicate this issue did not report, so it belongs to its own change. The CHANGELOG says all of this in the same words.

I deliberately did **not** add the read term to the shared fragment itself: `proxy_helpers.rs:3346` records that this was considered and rejected for #3331, because it would also narrow repository *listing*, for which "any grant means you may see this repository exists" is the intended contract. The fragment is untouched; only this new arm carries the action term.

### 2. Set-driven instead of a correlated `EXISTS` (performance)

The correlated form seq-scanned `repositories` and re-ran the fragment's project subquery once per row, so its cost tracked the instance's **repository count rather than the caller's access**. The arm is now a join driven from `permissions` into `repositories`, with the fragment's correlated `(SELECT rp.project_id FROM repositories rp WHERE rp.id = ...)` replaced by the join's own `r3.project_id` — the same value by construction, since `rp.id = r3.id` means `rp` *is* that row.

`EXPLAIN (ANALYZE, BUFFERS)`, `postgres:16-alpine`, migrations 001-211, 10,000 repositories / 15,002 `permissions` rows / 50 groups, `jit=off`, warm, best of 3:

| caller | `origin/main` | first cut (correlated, tenant-only) | **this revision** (set-driven, read-gated) |
|---|---|---|---|
| typical: 1 group, 2 grants | 1.3 ms | 18.4 ms | **2.0 ms** |
| worst: 50 groups, 7,501 read-carrying grants | 1.3 ms | 37.4 ms | **13.5 ms** |

Plan shape confirms the fix: the first cut showed `SubPlan 1` (the per-repository project subquery) executed once per repository row; the revision starts from an indexed scan of `permissions` filtered to the caller and the remaining cost scales with the caller's grant count.

Because this is **no longer the verbatim shared fragment**, `permissions_read_grant_join_for` (`repository_service.rs`, placed directly beside `permissions_grant_exists_for`) documents the invariant it must hold, and two DB tests enforce it rather than leaving it to review.

### 3. Nits from review

* `search.rs:4-5` no longer claims "Uses OpenSearch when available" — `OpenSearchService`'s search methods have no production callers (comment only; no code touched).
* `RepoAccessMode::UserScoped`'s doc comment, missed by the first cut, now describes the actual predicate.

Merged `origin/main` (`edf2b123`, #3690 toml) first; clean, no conflict with these files.

## Regression tests

`grant_visibility_db_tests` (`search.rs`), sharing `SearchGrantFixture` and a new `GrantMatrix` covering every grant shape the fragment resolves:

| test | asserts |
|---|---|
| `search_honours_group_permission_grant_db` | group grant `{read,write}` → hit. Ported from the #3681 investigation's repro. |
| `search_honours_direct_user_permission_grant_db` | direct `{read}` → hit. |
| `search_honours_a_read_write_grant_db` | `{read,write}` → hit. |
| **`search_denies_a_write_only_grant_db`** | **`{write}`-only → 0 hits**, and the read gate agrees. Pins the action dimension. |
| `search_denies_private_repo_without_any_grant_db` | grant for a group the caller is not in → 0 hits. |
| `search_still_returns_public_repo_results_db` | public repo unchanged. |
| **`search_grant_arm_matches_the_shared_fragment_reference_db`** | **drift guard**: set equality against a reference query built from the *real* `permissions_grant_exists_for` plus the read term, over the matrix, for both a user and a service-account principal. |
| **`search_grant_arm_is_contained_by_the_read_gate_db`** | **#3331 invariant**: every repository the arm yields passes `user_can_access_repo(.., READ)`; also pins each matrix case. |

The matrix covers `{read}`, `{write}`, `{admin}`, `{read,write}`, `{delete}`, `{}`, group-in / group-not-in, service-account, project-inherited `{read}` / `{write}`, and no grant — closing the project-inheritance and service-account gaps both reviewers flagged as claimed-but-untested.

## Verified

All cargo under the shared compile lock; DB-backed against a throwaway `postgres:16-alpine`, migrations 001-211.

* `cargo fmt --check`, `cargo check --lib --tests`, `cargo clippy --all-targets -- -D warnings` — all clean.
* `cargo nextest -E 'test(grant_visibility_db_tests)'` — **8 passed**.
* Broad selector (`search`/`permission`/`visib`/`grant`/`user_can_access_repo`/`action_gate`) — see check run.

**Counterfactual A — the action gate is real.** Arm reverted to the first cut's tenant-only fragment, tests unchanged:

```
FAIL search_denies_a_write_only_grant_db
     #3697: a `{write}`-only grant must NOT make a private repository's artifacts
     searchable — search would then be wider than the read gate it claims to mirror
     (got 1 hits)        left: 1   right: 0
FAIL search_grant_arm_is_contained_by_the_read_gate_db
     #3697/#3331: search returned repositories the READ gate denies
     (["user-write", "user-delete", "group-write", "project-write"])
FAIL search_grant_arm_matches_the_shared_fragment_reference_db
Summary  8 tests run: 5 passed, 3 failed
```

**Counterfactual B — #3697 is still closed.** Third arm removed entirely (`origin/main` behaviour):

```
FAIL search_honours_group_permission_grant_db      (0 hits, expected 1)
FAIL search_honours_direct_user_permission_grant_db
FAIL search_honours_a_read_write_grant_db
FAIL search_grant_arm_is_contained_by_the_read_gate_db
FAIL search_grant_arm_matches_the_shared_fragment_reference_db
PASS search_denies_a_write_only_grant_db  (main saw no grant at all)
PASS search_denies_private_repo_without_any_grant_db
PASS search_still_returns_public_repo_results_db
Summary  8 tests run: 3 passed, 5 failed
```

**SQL equivalence, independently of the harness.** On the seeded 10k dataset, the shipped set-driven arm vs the correlated reference (shared fragment + read term): `302` rows both ways, `EXCEPT` empty in **both** directions.

**Coverage.** All six routes plus both `/advanced` branches reach the one resolver through `resolve_accessible_repos` (7 call sites); no endpoint carries its own copy of the predicate, and `SearchService`/`OpenSearchService` consume the resolved id list. `POST /admin/search/reindex` resolves no visibility set.

**`.sqlx` caches:** unaffected — runtime `sqlx::query_as` with `AssertSqlSafe`, not a `query!` macro, so no offline entry exists (#3674). Dynamic SQL keeps its `AssertSqlSafe` wrapper.

**Note on the environment:** a host cleanup sweep removed my Postgres container mid-run. I recreated it, re-applied migrations, re-seeded, and re-ran every DB-backed measurement and both counterfactuals above on the rebuilt database; all figures and transcripts here are from those verified-live runs.

**Not run:** full `--lib` suite, e2e, coverage/jscpd jobs. The new tests are an inline `#[cfg(test)]` module, which `scripts/ci/jscpd-prepare-sources.py` strips before the duplication gate measures.

## Deliberately out of scope

`GET /api/v1/webhooks` has the same bug class — `list_webhooks` (`webhooks.rs:295`, predicate at `:320-333`) scopes on `created_by` + `role_assignments` with no `permissions` arm. Different endpoint, different table; kept out to keep this diff minimal. Worth its own issue.

Closes #3697


brandonrc:
## Audit round 1 — 2 reviewers (security / correctness+performance)

**Verdict: NEEDS-WORK.** The fix genuinely closes #3697: all seven `/api/v1/search*` call sites resolve through the one helper (no handler carries its own predicate; the OpenSearch path has no production callers), group resolution goes through `user_group_members` identically, project inheritance works, `actions = '{}'` fails closed, `UNION` dedups so `/advanced` totals can't double-count, and **token scope still intersects after visibility** (probed: `Restricted([A])` → 0 hits for repo B despite a read grant on B). Counterfactual reproduced by both: reverting the hunk fails the two positive tests. 742/742 on the selector.

**Blocker (found by both, proven end-to-end by one):** the new arm reuses `permissions_grant_exists_for`, which is the **tenant** half of `user_can_access_repo(.., READ)` — it checks `actions <> '{}'`, not that `read` is among them, and `write` does not imply `read` (`permission_service.rs:214-220`). So a `{write}`-only grant now makes a private repository's whole inventory searchable — names, paths, versions, sizes, and `checksum_sha256` via `/search/checksum` — while `GET /api/v1/artifacts/{id}` for the same rows still 404s. Probe on the PR head: `search_hits=1 read_gate=false`; the same probe passes on main, so the widening is introduced here. The repo's #3331 structural guard says `TenantOnly` "must never front a path that returns a private repository's contents"; the arm evades it only by hand-composing the fragment. The CHANGELOG sentence "it grants no new access" is therefore false and must not ship. Honest counterweight: the pre-existing `role_assignments` arm and `GET /repositories` are tenant-only too, so this made search consistent with *listing* — but listing returns repository names, not artifact digests.

**Should-fix (performance):** the correlated `EXISTS` seq-scans `repositories` and runs the fragment's project subquery per row. Measured with `EXPLAIN (ANALYZE, BUFFERS)` on a seeded DB: 0.5 → 4.8 ms at 2k repos; **315 ms / 1.1M buffers** at 10k repos with a caller in 50 groups; 56 ms for a caller with one group and two grants. Cost tracks repository count, not access; the indexes are already used — the plan shape is the problem. A set-driven join gives byte-identical results at 4.5 / 1.7 ms.

Rework in progress: the arm gains the `read`-action predicate the read gate uses, is rewritten set-driven, gets action-dimension tests (`{write}`-only → 0 hits) and a DB equivalence test against the reference fragment, and the CHANGELOG says exactly what search now grants. The webhooks sibling is #3702.

brandonrc:
## Audit round 2 (head `58996e80`) — MERGE-WITH-NITS

The round-1 blocker is fixed and proved independently: the arm now requires `read` or `admin`, and `check_repository_action`'s ladder (`permission_service.rs:219`) is literally `$3 = ANY(actions) OR 'admin' = ANY(actions)` — no `*` wildcard exists in the action vocabulary (an `actions='{*}'` row is insertable but confers nothing). The reviewer seeded 23 grant shapes (`{admin}`-only, `{*}`, project-inherited, service-account, repository/project precedence mixes) and evaluated the shipped SQL beside hand-transcriptions of both halves of the read gate: **0 violations of 23**. On a 10k-repository seed, `EXCEPT` both ways against the correlated reference was empty for all 1,001 users (129,840 arm rows), and containment over 200 callers gave 0 violations. Performance: main 2.7 ms → round-1 61.3 ms → round-2 **4.0 ms**; the per-row `SubPlan` is gone (the only remaining one is a hashed group-membership list evaluated once), and the join correctly uses `idx_permissions_principal`. Counterfactuals: reverting the read term fails 3 (write-only, containment, drift guard); reverting to main fails 5. 26/26 on the selector, check/clippy clean.

Folding in as a final commit: the CHANGELOG's illustration of the residual `role_assignments` gap understated it (a *global* assignment with an empty-permission role makes every repository searchable; a role carrying `read` still searches a repository that a `{write}`-only rule makes the gate refuse) — the central claim that the whole residual is that arm stays true; and `GrantMatrix::teardown`'s unscoped `key LIKE 'm3697w-%'` is scoped to the test's own key so the two matrix tests stop deleting each other's project. Then a re-merge of `main` (#3700 landed in `### Fixed`) and CI.

PR #3703, opened 2026-09-03
fix(security): let a public repository satisfy the API-token repo-scope check on reads

### LINKED ISSUE #3648: fix(auth): public repository + repo-scoped credential returns 403 — a credential grants less access than none
## Summary

On a **public** repository, a request carrying a repo-scoped token returns **403**, while
the identical request with **no credential at all** returns **200**. Presenting a valid
credential grants *less* access than presenting none.

pip surfaces this as *"No matching distribution found"*, which points nowhere near the
cause — the operator sees a resolution failure, not an authorization failure.

Reported against an earlier version and still reproducible on 1.8.1.

## Reproduction

Mint a token scoped to repository A, then request a package from **public** repository B:

```
scoped token                              -> 403 "Token does not have access to this repository"
same token as pip basic auth __token__:<tok> -> 403
no credential                             -> 200
read-all token                            -> 200
```

## Root cause

The public-visibility check passes, and then — *because a credential is present* — the
token's repository scope (`can_access_repo`) is evaluated with **no public bypass**. An
anonymous request short-circuits before that second check and is served. So the scope
check is applied in exactly the case where it should be irrelevant: a repository that is
readable by everyone.

## Expected

A public repository is readable by any authenticated principal regardless of token scope.
Presenting a credential must never reduce access below what an anonymous caller gets.

Concretely: the repo-scope check should apply the same public-repository bypass that the
anonymous path already takes.

## Acceptance criteria

- A repo-scoped token reading a public repository outside its scope gets 200, matching the
  anonymous result.
- Scope enforcement on **private** repositories is unchanged (a scoped token still cannot
  read a private repository outside its scope).
- A regression test covers the public + out-of-scope-token combination directly, since the
  adjacent cases (private + scoped, public + anonymous) both already behave correctly.

## Note

Deployments where every token is read-all will not observe this, but a single newly-minted
scoped token or a scoped `~/.netrc` reintroduces it. Worth not closing as
fixed-by-upgrade without testing this exact combination.

brandonrc:
## Summary

On a **public** repository, a request carrying a repository-scoped API token returned **403**, while the identical request with **no credential at all** returned **200**. Presenting a valid credential granted strictly *less* access than presenting none — on exactly the repositories that are readable by everyone.

`repo_visibility_middleware` runs the #504 token repository-scope check (`AuthExtension::can_access_repo` → `AccessScope::grants`) immediately after the visibility check, and that check had no `is_public` bypass. An anonymous caller never reaches it: there is no `AuthExtension` to test, so it short-circuits and is served. A caller presenting a token scoped to a *different* repository reaches it and gets `403 Token does not have access to this repository`. pip renders that 403 as `No matching distribution found`, which points nowhere near authorization.

The gate now takes the same `public_read_satisfies_acl(is_public, action)` baseline the fine-grained ACL arm ~40 lines below it in the same function already applies for the same reason (#2329) — the existing "public read beats ACL" idiom, reused rather than re-derived. The REST read gate `require_visible` has an equivalent unconditional public short-circuit, and the OCI read gate `oci_read_permitted` already calls `public_read_satisfies_acl` directly.

## What a scoped token can now do that it could not before

**Reads of a public repository outside its `allowed_repo_ids`, over the native format routes** — which is precisely what the same request already got with **no credential at all**:

- `GET` / `HEAD` / `OPTIONS` (`action_for_method` → `read`);
- `POST /{repo}/gallery/extensionquery`, the VS Code gallery **search** verb. This is the sole member of `is_anonymous_readable_format_post` — by definition the POSTs a public repository must serve to an *anonymous* caller — so it skips the #508 write gate, was served with no credential, and inverted in exactly the same way. The scope gate now derives its action the way the ACL arm below it already did for that route, so the two gates in one function agree instead of disagreeing.

Nothing else moved:

- **Writes and deletes are unchanged.** `public_read_satisfies_acl` is `read`-only by construction. A token scoped to repo A still cannot push to public repo B — matching `require_repo_write_access` on the REST side.
- **The other two `is_non_mutating_format_post` routes stay scope-gated**, and are pinned by the test: git-lfs `objects/batch` is an upload negotiation whose upload arm mints object hrefs, and conan `users/authenticate` is a credential exchange. Both answer 401 to an anonymous caller under #508, so neither has an anonymous baseline to have fallen below.
- **Private repositories are unchanged.** They never satisfy `public_read_satisfies_acl`, so a scoped token reading a private repository outside its scope is refused as before, and the existence-hiding answer is untouched.
- **The #508 anonymous-write gate runs before this check** and is untouched.
- **The permission/ACL arm below is untouched** — this only removes a gate that ran *before* it.

**Two honest caveats**, both also in the CHANGELOG:

1. The *middleware's* decision is now identical to the anonymous one. **Handlers** that require a credential of their own still answer an authenticated caller differently, as they always have — conan `users/authenticate` is 401 anonymously and 200 with a token. No escalation: the JWT it mints inherits the presenting credential's repository allow-list (#2430), so the ceiling is preserved.
2. The "equal to what anonymous already gets" justification presumes an anonymous baseline exists. Under `AK_GUEST_ACCESS_ENABLED=false` no anonymous caller reaches a format route at all, and there a scoped token gains read on any repository still flagged public from before that policy was set. This is consistent with `require_visible` and the #2329 ACL arm, which have both had an unconditional public short-circuit all along, and `coerce_is_public_for_create` prevents *new* public repositories in that mode.

**Deferred to #3704:** three surfaces keep the same ordering and are deliberately not changed here, each being a separate gate on a separate mount — the OCI `/v2` read gate (`enforce_token_repo_scope`, #2290), the tree browser (`tree_access_allowed`, whose `test_access_public_out_of_token_scope_denied` pins the current behaviour and would have to change with it), and the REST artifact gate (`check_artifact_visibility`).

## Regression test

`api::middleware::auth::tests::test_3648_public_repo_read_is_not_refused_to_an_out_of_scope_token` — router-level, DB-backed, driving a **real** repository-scoped API token (minted through `AuthService::generate_api_token`, pinned via the `api_token_repositories` join table) through `repo_visibility_middleware`.

| case | expected |
| --- | --- |
| public B, token scoped to A, `GET` (Basic `__token__:<tok>` — pip's shape) | 200 (**was 403**) |
| public B, token scoped to A, `GET` (Bearer) / `HEAD` | 200 (**was 403**) |
| public B, token scoped to A, `POST .../gallery/extensionquery` | 200 (**was 403**) |
| public B, anonymous, `POST .../gallery/extensionquery` | 200 (unchanged) |
| public B, token scoped to A — middleware status **and body** vs. anonymous | identical |
| public B, token scoped to A, `PUT` / `POST` / `DELETE` | 403 **from the scope gate** (unchanged) |
| public B, token scoped to A, `POST /lfs/…/objects/batch`, `POST /conan/…/users/authenticate` | 403 **from the scope gate** (unchanged) |
| private C, token scoped to A, `GET` / `PUT` | 403 **from the scope gate** (unchanged) |
| public B, no credential, `GET` | 200 (unchanged) |
| public A (in scope), `GET` | 200 (positive control) |

Every refusal asserts the response **body**, not just the status. That is load-bearing: the middleware emits two different 403s — `forbidden_repo_response` (the scope gate) and `forbidden_permission_response` (the ACL arm below it) — so a status-only assertion cannot tell which gate refused, and the write-side assertions were vacuous without it.

The byte-parity assertion is scoped to what it proves: `run_through_visibility` mounts a **constant stub**, so the equality pins that the *middleware* does not vary its own answer. It does not speak for real format handlers, which receive the injected `Extension(Some(auth_ext))`. This is stated in the test's own doc comment.

**Three counterfactuals, all run against the live database (not reasoned):**

| mutation | result |
| --- | --- |
| fix hunk reverted to main's `if !ext.can_access_repo(repo.id)` | **FAIL** — `left: 403, right: 200` on the pip Basic probe, the issue's exact pair |
| bypass opened to every method (`!is_public && !ext.can_access_repo(…)`) — the reviewer's mutation that previously passed | **FAIL** — `left: (403, "You do not have permission to perform this action on this repository"), right: (403, "Token does not have access to this repository")` |
| vscode reclassification dropped (method-only, i.e. the first cut of this PR) | **FAIL** — `left: 403, right: 200` on the gallery POST |

## Verified

Merged `origin/main` (`b2c3cf35`, #3696) into the branch — the only conflict was a `### Security` bullet at the same anchor, resolved to one heading with both bullets and no markers. Re-verified afterwards against a throwaway Postgres 16 (migrations applied) under the compile lock, with `AK_TESTS_REQUIRE_DB=1` so a missing database is a hard failure rather than a silent skip:

- `cargo fmt --check` — clean
- `cargo nextest run --lib -E 'test(test_3648_…)'` — **1 passed**; three counterfactuals above — **all fail**
- `cargo nextest run --lib -E 'test(middleware::auth) or test(visibility) or test(repo_scope) or test(public_read) or test(private_repo) or test(non_mutating) or test(anonymous_readable) or test(vscode)'` — **482 passed**
- `cargo nextest run --lib -E 'test(can_access_repo) or test(satisfies_acl) or test(action_for_method) or test(3387) or test(3452) or test(2329) or test(504) or test(3174) or test(508)'` — **30 passed**
- `scripts/ci/check-changelog-unreleased.sh` — passes

Not run: the full `--lib` suite and the `backend/tests` integration binaries (build cost); no end-to-end run with a real pip or VSCodium client. Per the correctness review, no source-scanning structural gate in the repo constrains the ordering of checks inside this middleware, so the behavioural test above is the only protection this invariant has.

Closes #3648


brandonrc:
## Audit round 1 — 2 reviewers (security / correctness+integration)

**Verdict: MERGE-WITH-NITS; small rework in progress.** Both reviewers verified the widening is bounded exactly as claimed: `action_for_method` (including its catch-all) reaches no side-effecting GET/HEAD/OPTIONS route across all ~33 format routers; the #508 anonymous-write gate still runs first; existence hiding is unchanged; `is_public` staleness is bounded by the same 60 s cache anonymous reads already use and is invalidated on update; and the #3323/#3525 virtual case **does not leak** — DB-probed, `member_passes_token_scope` re-applies token scope per member with no public-parent read-through. It closes #3648's literal reproduction: the test drives Basic `__token__:<token>` (not only Bearer) with a genuinely scoped token through `/pypi/{repo}/simple/…`, and the counterfactual fails at the Basic probe with `left: 403, right: 200`. 305 middleware tests DB-backed; merge with main clean at the time of review.

Being fixed before merge: (1) the three write-side "still 403" assertions are vacuous — a mutation opening the bypass to every method still passes all eleven, because the permission arm returns the same 403 from a different helper; asserting the body fixes it. (2) `POST /vscode/{key}/gallery/extensionquery` — the one anonymous-readable POST — still inverts (measured: anonymous 200, scoped 403); the permission arm already reclassifies it as a read and the scope gate should too. (3) The "byte-for-byte identical to anonymous" sentence overstates what the stub-routed test proves — real handlers may legitimately answer an authenticated caller differently (`conan …/users/authenticate` 200 vs 401, no escalation). (4) Residual surfaces named and pointed at #3704. Structural note from the correctness pass: no source-scanning gate constrains check *order* in `repo_visibility_middleware` — the #3443/#3444 tests are behavioural.

brandonrc:
## Audit round 2 (head `164ca378`) — MERGE

Both round-1 should-fixes verified fixed by reproduction. The scope gate now derives its action from the same `anonymous_readable_post` value the #508 write gate already computes (`auth.rs:2282-2284` → `:2384`), so the two cannot drift; it matches exactly one route (`POST /vscode/{key}/gallery/extensionquery`), LFS batch and conan authenticate stay scope-gated and are pinned by body assertions (mutating to `non_mutating_post` fails at `:9174`), and a scoped token on the gallery POST of a *private* repo is still refused (probed live). The write-side assertions are now load-bearing: the reviewer's widening mutation fails at `:9155` on the body difference between the ACL refusal and the scope refusal — both hard-coded `&'static str`, no i18n, so not brittle. Merge with main is a single `+8` hunk under one `### Security` heading, zero markers. 34/34 against a live DB, three mutations all fail, check/clippy/fmt clean.

Merging on green. Nits carried to #3704: the private-repo gallery POST is unpinned by a test, and the CHANGELOG writes the gallery route without its `/vscode` mount prefix.

Issue #2472, opened 2026-07-13
Projects P1: project entity + project-scoped RBAC (metadata, no URL change)


brandonrc:
Part of epic #2470 (Phase 1 of 4). Design in #2470. Origin #1101 / discussion #2268.

Introduce Projects as a **metadata containment layer** with project-scoped permissions. **No repo-key or client-URL change.**

### Scope
- Migration **154**: `projects(id, key UNIQUE, name, description, quota_bytes BIGINT NULL, timestamps)` + **nullable** `project_id` on `repositories`. `project_id = NULL` behaves EXACTLY as today (product decision: assignment is optional; seed a `_default` row for UI listing but do NOT backfill existing repos).
- Membership reuses the existing `permissions` table with `target_type='project'` — **no third authz store**. Groups stay the principal layer (who); a group or user is granted a project role.
- **Effective-permission inheritance folded into the two shared predicates only**: `permissions_grant_exists` (repository_service.rs:225) and `user_can_access_repo` (:744), plus the listing clause `build_visibility_clause_for` (:270) — so a repo is accessible via repo-direct grant OR project-inherited grant OR legacy role_assignment OR public. Every gate that routes through these (incl. the ~27 #2439/#2443 routes, `check_artifact_visibility`, `require_visible`, `require_repo_write_access`) becomes project-aware at once.
- `/api/v1/projects` CRUD; repo create/update accepts optional `project_id`; repo list gains `?project=`.

### Sequencing
**Hard prerequisite:** the #2439/#2443 systemic-authz gate consolidation, so project inheritance lands in one shared path and propagates everywhere.

### Validation (full loop) + GATING authz red-team
Unit: project-inherited grant grants access; non-member denied; `project_id NULL` repo unaffected; listing surfaces project repos. E2e on pool: create project, grant a group project-read, member reads a repo in-project (200) and 404 on a repo outside it; existing flat repo URLs unchanged. **Authz red-team (gating): project inheritance must NOT widen access to NULL/public repos and must NOT leak across projects.**

brandonrc:
FixSpec complete (architect). Executable spec: migration **154** (`projects` table + **nullable** `repositories.project_id` + `_default` seed + `ON DELETE SET NULL`); `models/project.rs`; membership via existing `permissions` table (`target_type='project'`); new admin-gated `/api/v1/projects` CRUD + members handler; repo create/update `project_id` + `?project=` list filter.

[...]

PR #3708, opened 2026-09-03
fix(webhooks): list webhooks on repositories reached by a permission grant

### LINKED ISSUE #3702: GET /api/v1/webhooks scopes on created_by + role_assignments only — permission grants are invisible (same class as #3697)
## Summary

Found while fixing #3697. `list_webhooks` (`backend/src/api/handlers/webhooks.rs:~295`, predicate at `~320-333`) resolves which webhooks a non-admin may see from `created_by` plus `role_assignments`, and never consults the `permissions` table — so a user whose access to a repository comes only from a permission grant (user or group principal) cannot list that repository's webhooks, while every other read on the repository works.

Same shape as #3697 (search), different endpoint and table.

## Fix

OR in the existing `permissions_grant_exists_for` fragment (`repository_service.rs`) the way #3701 did for `resolve_visible_repos`, keeping API-token repository scope layered where it is. DB-backed regression tests for a group grant and a direct-user grant, plus a negative.

Refs: #3697, #3701, #3681.

brandonrc:
Closes #3702

## Summary

`GET /api/v1/webhooks` scoped a non-admin's listing to webhooks they created plus webhooks attached to a repository they hold a `role_assignments` row for. It never consulted the `permissions` table, so a caller whose access to a repository came only from a fine-grained grant — direct-user, service-account, group, or inherited from the owning project — saw none of that repository's webhooks, while `GET /api/v1/webhooks/{id}` on one of those very rows answered normally (the per-webhook path goes through `user_can_access_repo`, which has honoured both authz stores since #2472).

The scope predicate now carries a third arm resolving the grant. The `created_by` and `role_assignments` arms are byte-identical to before, and the arm is a pure `OR`, so nothing that was visible stops being visible.

## The action term (the #3701 audit lesson, applied)

`permissions_grant_exists_for` is only the **tenant** half of the repository read gate (`actions <> '{}'`). The gate is that fragment **plus** `check_repository_action(.., "read", ..)`, and `write` does not imply `read`. ORing the bare fragment in would have listed a repository's webhook URLs, event filters and secret digests to a publish-only CI principal that the read gate refuses — exactly what `RepoAccess::TenantOnly`'s contract forbids fronting (#3331). The arm therefore requires `read` or `admin`, and it is a set-driven `JOIN` rather than a correlated `EXISTS`: correlated, the fragment's project subquery is re-run per repository row (56–315 ms per call at 10k repositories in the #3701 measurements), and this predicate runs twice per request — page query and count query.

It reuses `permissions_read_grant_join_for`, the helper #3701 added, rather than a second copy. Now that #3701 is merged (`9f4b2617`), `origin/main` is merged into this branch and `backend/src/services/repository_service.rs` here is **identical to main** — this PR touches only `webhooks.rs` and the CHANGELOG.

## Scope

- `list_webhooks` was the only surface with a hand-rolled predicate.
- `get_webhook` and `list_deliveries` share `authorize_webhook_access`, which already resolves grants through `user_can_access_repo`; they needed no change.
- `delete` / `enable` / `disable` / `test` / `rotate-secret` / `redeliver` are admin-only (GHSA-qcmj). **No mutation is widened.**
- This leaves the listing *narrower* than `get_webhook` on the grant path: the listing requires `read`, while `authorize_webhook_access` deliberately stays on the tenant gate with a `TENANT-GATE-ONLY` justification at the call site. Narrower is the safe direction; changing that call site is a separate policy question and is not touched here.
- No API-token scope layering exists in this handler, so none was moved.

## Tests (new, DB-backed, in `webhooks::tests::object_level_authz_tests`)

| test | asserts |
| --- | --- |
| `list_webhooks_honours_group_permission_grant_db` | group grant carrying `read` → webhook listed. **Fails on main** |
| `list_webhooks_honours_direct_user_permission_grant_db` | direct-user `{read}` grant → listed. **Fails on main** |
| `list_webhooks_denies_a_write_only_grant_db` | `{write}`-only grant → NOT listed (fails against a bare-tenant-fragment fix) |
| `list_webhooks_denies_repo_without_any_grant_db` | grant exists but for a group the caller is not in → not listed |
| `list_webhooks_still_lists_own_webhook_regardless_of_grant_db` | the caller's own webhooks stay listed with no grant and with a `{write}`-only grant |
| `list_webhooks_grant_arm_is_contained_by_the_read_gate_db` | 12-shape matrix (user read/admin/read+write/write/delete/empty, group read/write, foreign group, project read/write, no grant): expectations pinned per shape, **and** every listed webhook's repository passes `user_can_access_repo(.., READ)` |

Each carries the `read_gate_allows` control so a test that agreed with a co-broken gate still shows up.

## Verified

All re-run after merging `origin/main` (`9f4b2617`), against a **fresh** throwaway Postgres 16 (`ak-fix-3702`, migrations `000`–`211` applied) under the repo's compile lock, `SQLX_OFFLINE=true`:

- `cargo fmt` — clean.
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(list_webhooks) or test(webhook_access) or test(permissions_read_grant) or test(every_user_can_access_repo_call_names_its_action) or test(object_level_authz) or test(search_grant_arm)'` with `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1` — **25 passed, 0 failed** (includes the #3331 `every_user_can_access_repo_call_names_its_action` structural gate, the pre-existing `list_webhooks_scoping`, and #3701's own `search_grant_arm_*` invariants, which still hold on the merged tree).
- **Fails-on-main proof.** With only the new `OR` arm removed from `scope_sql` (tests untouched), the same DB run gives **4 passed, 3 failed**: `list_webhooks_honours_direct_user_permission_grant_db`, `list_webhooks_honours_group_permission_grant_db` and `list_webhooks_grant_arm_is_contained_by_the_read_gate_db` FAIL, while the two negatives, the own-webhook test and `list_webhooks_scoping` still pass. Arm restored afterwards.
- **Write-only proof at the SQL level.** Seeded a private repository with a single `{write}` grant for a user and evaluated both predicates directly in psql: `permissions_grant_exists_for` → `t` (would have listed it), `permissions_read_grant_join_for` → `f`. That is the gap `list_webhooks_denies_a_write_only_grant_db` pins.
- `EXPLAIN` on the generated count query confirms the arm plans as a hashed subplan driven by `idx_permissions_principal`, not a per-row correlated re-evaluation.

Not run: the full lib suite, integration/e2e suites, and the web client. CI covers those.


brandonrc:
## Audit — 1 reviewer (the #3701 pattern applied)

**Verdict: merge on green.** The arm calls the shared `permissions_read_grant_join_for` (`repository_service.rs` is byte-identical to main after the re-merge), carries the `read`/`admin` requirement, uses the same `$N` bind, and page + count derive from one nested `fn` so they cannot drift; the whole `webhooks.rs` diff has exactly one deleted line. Containment: the PR's 12-shape matrix plus a 9-shape probe (group `{admin}`, service-account read/write, project `{admin}`/`{delete}`, unrecognised `{*}`) — **21 shapes, 0 violations**; the listing tracks `user_can_access_repo(READ)` exactly. Counterfactual: arm removed → the positives and both matrices fail. Mutations untouched (six `require_admin()` sites). `EXPLAIN` on the real count query: hashed `SubPlan` at `loops=1`, 15.6 ms on a 5k-repo/4k-grant seed. 24/24 DB-backed; check/clippy clean.

Landed as `89724172`: the CHANGELOG now states the two residuals — `GET /api/v1/webhooks/{id}` and its delivery listing stay on the tenant-only gate (a `{write}`-only holder can fetch by id what no longer lists; narrower listing than fetch is the safe direction), and the pre-existing `role_assignments` arms remain action-blind. Filed **#3715** for the pre-existing gap the reviewer confirmed: none of the webhook read endpoints layer API-token repository scope (`allowed_repo_ids`), unlike `packages.rs`/`search.rs`.

Issue #504, opened 2026-03-21
API token repo scope is not enforced


brandonrc:
## Summary

API tokens created with repository scope restrictions (`allowed_repo_ids`) are not enforced. A token scoped to repo A can successfully access repo B, including listing artifacts and uploading.

## Evidence

Release gate run: https://github.com/artifact-keeper/artifact-keeper-test/actions/runs/23372998677

- Scoped token accessing repo B: HTTP 200 (expected 403)
- Scoped token uploading to repo B: HTTP 200 (expected 403)

## Test Reference

`artifact-keeper-test/tests/rbac/test-scope-enforcement.sh`

## Impact

Critical. Repository isolation is broken. A leaked token scoped to a public repo can access private repos.

## Fix

The `can_access_repo()` check in `auth.rs` middleware must be enforced in all handler paths, not just a subset. Audit all handler functions that access repository data to ensure they call `can_access_repo()`.

PR #3709, opened 2026-09-03
fix(security): answer the same existence-hiding 404 for both private-repo read denials

### LINKED ISSUE #3524: Native format routes answer 403 for a private repository that carries ACL rules and 404 for one that does not, leaking repository existence
Found while reproducing #3452. Filed separately because fixing it changes the
status code on the wire and is not part of that report's per-format divergence.

`repo_visibility_middleware` has two deny paths for an authenticated non-member
reading a PRIVATE repository, and they answer differently depending on
something the caller has no business learning:

| repository state | response |
|---|---|
| private, at least one fine-grained rule exists (any principal) | `403` `You do not have permission to perform this action on this repository` |
| private, no fine-grained rules | `404` `Repository not found` |
| does not exist | `404` `Repository not found` |

So a `403` tells an authenticated caller with no grant that (a) the repository
exists and (b) it is governed by an ACL. Reproduced on a clean instance
(`dave` holds no grant anywhere):

```
GET /maven/{private-repo-with-rules}/com/example/demo/1.0.0/demo-1.0.0.pom -> 403 (68 bytes)
GET /maven/{private-repo-no-rules}/…                                       -> 404 (20 bytes)
GET /maven/no-such-repo-xyz/…                                              -> 404 (20 bytes)
```

This is the same oracle class the middleware already closes elsewhere. Its own
no-repo branch mirrors the private-repo 401 specifically so that "no such repo"
and "repo you may not see" are byte-identical (#1808), and its rules-less
private branch returns `not_found_response()` with the comment "Existence-hiding
404, matching REST `require_visible`". The REST gate is uniform: `require_visible`
returns `NotFound` for every denial, never `Forbidden`. Only the native-format
`has_rules` read arm diverges.

Suggested fix: return `not_found_response()` from the `has_rules` read denial
too, so all three rows above are identical, matching `require_visible` and the
branch immediately below it. Writes are a separate question — a caller doing a
`PUT` has generally already read the repository, and `403` there is the more
useful answer.

Not folded into the #3452 fix because it changes an observable status code for
clients that may distinguish 403 from 404 today, which deserves its own
decision and its own regression test.

brandonrc:
## Summary

`repo_visibility_middleware` had two deny paths for an authenticated caller holding no grant on a **private** repository, and they answered differently depending on whether the repository happened to carry any fine-grained rule — for any principal, including a completely unrelated one:

| repository state | before | after |
|---|---|---|
| private, at least one fine-grained rule exists | `403` `You do not have permission to perform this action on this repository` | `404` `Repository not found` |
| private, no fine-grained rules | `404` `Repository not found` | unchanged |
| does not exist | `404` `Repository not found` | unchanged |

A `403` therefore told an ungranted authenticated caller both that the repository **exists** and that it is **governed by an ACL**, on every native format route. That is the same oracle class this middleware already closes elsewhere: its no-repo branch mirrors the private-repo `401` byte-for-byte for anonymous callers (#1808) and answers a valid credential with the existence-hiding `404` (GHSA-fv45-mwhh-q23r), and the REST gate is uniform — `require_visible` returns `NotFound` for every denial. Only the native-format `has_rules` read arm diverged.

## The change

One returned response in the `has_rules` read arm of `repo_visibility_middleware`: `forbidden_permission_response()` → `not_found_response()`, so the three rows above are byte-identical in status, content type and body (both bodies are hard-coded `&'static str`s in `auth.rs`; the unified answer reuses the existing builder, so they cannot drift). The log line already emitted on that denial now says it is answering the existence-hiding 404, matching the sibling rules-less branch, so an operator can still tell the two apart server-side.

**Nothing else moves:**

- **Public repositories are untouched.** That branch is reachable only for a private repository: `action` is always `"read"` inside it (the #2603 G1 arm above owns write and delete, including the `non_mutating_post` reclassification), so its `!public_read_satisfies_acl(is_public, action)` guard reduces to `!is_public`.
- **Members are untouched** — this is the `!allowed` path only.
- **Writes are untouched** and still answer `403` (#2603 G1), as the issue asks: a caller issuing a `PUT` has generally already read the repository.
- **Anonymous callers are untouched** — they never reach this block (no `auth_ext`); a private repo still answers the `401` + `WWW-Authenticate` challenge (#508 / #1808).
- **Token-scope refusals are untouched** — `forbidden_repo_response` (#504) still fires before this arm, which is what #3703's `private_get` control asserts.

## Regression test

`api::middleware::auth::tests::test_3524_private_repo_denial_does_not_reveal_whether_acl_rules_exist` (router-level, DB-backed, driven through `run_through_visibility`). Two private repositories differing **only** in whether any fine-grained rule exists against them (the rule names an unrelated principal), and:

- authenticated non-member → `(status, body)` **identical** across both, and equal to `(404, "Repository not found")`;
- anonymous → identical across both, and `401`;
- member → `200` on both (positive control, so the equalities cannot hold vacuously);
- a fixture assertion that `has_any_rules_for_target` really is `true` for one and `false` for the other, so the two probes provably take different branches.

Compared as `(status, body)` pairs rather than statuses on purpose: the two branches are separate response builders with different hard-coded bodies (68 bytes vs 20), so a status-only assertion would pass on a fix that left the oracle alive at the byte-count level.

**Proof it fails on main:** with the one-line change reverted (`not_found_response()` → `forbidden_permission_response()`) and the test left in place:

```
FAIL [ 0.265s] (1/1) artifact-keeper-backend
  api::middleware::auth::tests::test_3524_private_repo_denial_does_not_reveal_whether_acl_rules_exist

assertion `left == right` failed: #3524: an authenticated non-member must get the SAME
(status, body) from a private repository that carries ACL rules and one that does not. […]
  left: (403, "You do not have permission to perform this action on this repository")
 right: (404, "Repository not found")

Summary [0.287s] 1 test run: 0 passed, 1 failed
```

## Tests updated (they were pinning the oracle)

Three existing assertions pinned the old `403` for an authenticated non-member on a **private** repository that carries rules, and now expect the `404`. All keep their positive controls and their subject unchanged — what they pin is that the principal is *denied*, not which flavour of denial:

- `test_3387_applicable_rule_beats_an_ordinary_role_but_not_an_admin_carrying_one` — `write_only_rule` and `bare` (the `owner_with_write_only_rule` → `200` carve-out is untouched);
- `test_3522_user_roles_is_not_a_repository_authorization_input` — `via_user_roles`.

## Verified

Against a throwaway Postgres 16 (`ak-fix-3524`, migrations applied), under the compile lock, with `AK_TESTS_REQUIRE_DB=1`:

- `cargo fmt --all` — clean.
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(/3524|3648|3387|3452|3522|repo_visibility|visibility|forbidden|not_found/)'` — all pass, including #3703's `test_3648_public_repo_read_is_not_refused_to_an_out_of_scope_token`.

Not run: the full `--lib` suite and the `backend/tests` integration binaries (targeted builds per the fix-wave brief). No `backend/tests` binary asserts the middleware's private-repo read denial; the only native-route `403`s there are token-scope and handler-level ones, which this does not touch.

#3703 merged to `main` at `567a5b08` mid-way through this work; `origin/main` is merged in here and the selectors above were re-run on the merged tree. Its hunk (the public-read short-circuit before the #504 scope gate, ~line 2351) and this one (the ACL read arm, ~line 2600) auto-merged cleanly.

Closes #3524


brandonrc:
## Audit — 2 independent reviewers (correctness/integration, security)

**Verdict from both: merge on green.** The one-line change at `auth.rs` (`forbidden_permission_response()` → `not_found_response()` on the `has_rules` read-deny arm) is correct, minimal and provably private-repo-only — the read action on a public repository short-circuits through `public_read_satisfies_acl` before any ACL check. Both reviewers ran the counterfactual: with the line reverted, `test_3524` fails with exactly `(403, "You do not have permission…")` vs `(404, "Repository not found")`, plus the two updated tests, so the edited assertions are load-bearing. 8/8 targeted DB tests, 175/175 on the middleware selector, 167-test visibility/oracle sweep, `check`/`clippy` clean, merge with main clean. Timing between the three now-identical 404s: 0.3–0.45 ms median with overlapping p10–p90, two orders below the #3504 bar. No e2e assertion anywhere pins 403 for this case.

Both reviewers converged on the same gap: the CHANGELOG claimed more than shipped. Landed as the last commit:

- "Only the native-format read arm diverged" was false — `oci_read_permitted` (`oci_v2.rs`) runs the identical `has_rules ? 403 DENIED : 404 NAME_UNKNOWN` construct behind every `docker pull`, and `/v2` is mounted outside the middleware. Filed **#3716**.
- The #504 scope gate is now the lone native read denial answering 403 on an existing private repository while a missing key answers 404, and scoped tokens are self-service. Filed **#3717**.
- Writes keep the existence half of the oracle by #3524's own decision; the upgrade note now says so instead of framing it as a pure usefulness trade-off.
- The server-side denial log predates this change; wording corrected.

Not changed: the deliberate write-method 403, and the pip/npm/cargo client-side rendering of a permission failure as "not found" (pip already did).

PR #3710, opened 2026-09-03
fix(security): reject a NUL byte in the query string and the login username

### LINKED ISSUE #3673: NUL byte in query strings and JSON bodies still reaches Postgres as a 500 (search name/format/prefix params, login username) - sibling of #3622/#3545
## Summary

#3545 and #3665 reject a NUL in the request **path** before any handler runs. The same class survives on inputs the path guard cannot see. Measured on `main` during the #3665 audit (identical with and without that PR - pre-existing):

| request | result |
|---|---|
| `GET /api/v1/search/advanced?name=a%00b` | 500 `DATABASE_ERROR` - `build_name_filter` (`search_service.rs:149`) binds the value verbatim at `:364` |
| `GET /api/v1/search/advanced?format=a%00b` | 500 |
| `GET /api/v1/search/suggest?prefix=a%00b` | 500 |
| `POST /api/v1/auth/login` with a JSON `username` containing `<NUL>` | 500 (control without the NUL: 401) |

`?q=` / `?query=` are already safe - `sanitize_tsquery_lexeme` strips the NUL.

All four are reachable anonymously. The login one is on an unauthenticated auth endpoint; #3666 means the body is now the sanitised envelope rather than raw driver text, but the 500, the pool checkout and the Postgres ERROR line remain.

## Ask

The path guard's approach (reject at a shared boundary) does not extend cleanly to query strings and bodies, since those are typed extractors. Options, in the order I'd try them:

1. A small `#[serde(deserialize_with)]` / newtype for text query and body fields that rejects `\0` with a 400 - covers login and search in one place if the search filters share a type.
2. A `Query<T>` / `Json<T>` rejection layer that walks decoded string values for NUL before the handler sees them - broader, but must not touch binary bodies.
3. Per-handler checks on the four sites above as a stopgap, with a structural guard so new text binds get the check.

Regression tests: the four requests above -> 400, with a control each.

Refs: #3545, #3622, #3665, #3623/#3666 (envelope), #3667.

brandonrc:
Closes #3673

## Summary

#3545/#3665 refuse a NUL byte in the request **path**, but the path guard cannot see a query string or a request body, so four anonymously reachable sites still turned a `%00` into a `500 DATABASE_ERROR`:

| request | before | after |
|---|---|---|
| `GET /api/v1/search/advanced?name=a%00b` | 500 | 400 |
| `GET /api/v1/search/advanced?format=a%00b` | 500 | 400 |
| `GET /api/v1/search/suggest?prefix=a%00b` | 500 | 400 |
| `POST /api/v1/auth/login`, `username` containing `\0` | 500 (control: 401) | 400 (control: 401) |

Each cost a pool checkout and a query Postgres rejects at the wire protocol.

## What changed

**Query strings — the same middleware, one more input** (`api/middleware/nul_path.rs`). `nul_path_guard` now also refuses a NUL in the **raw query string**, decoded with the same `urlencoding::decode_binary` byte scan and the same `%`-free fast path as the path half (the predicate is shared, renamed `decoded_path_has_nul` → `decoded_has_nul`). This closes every query-parameter site on every surface at once — including ones added later — rather than the three filters named in the issue, and needs no per-handler code. `decode_binary` is infallible, so `?prefix=a%00b&x=%C0%80` cannot blind the check (the #3665 bypass shape, now on the query string).

**The body site — refused at the field** (`api/extractors.rs`, `api/handlers/auth.rs`). A body is a stream, so a `Json<T>` rejection layer would have to walk every request's bytes — binary uploads included — to cover one text field. Instead `LoginRequest::username` carries `#[serde(deserialize_with = "extractors::deserialize_nul_free_string")]`; the serde error becomes the same 400 `VALIDATION_ERROR` envelope every other malformed body already gets through the crate's `Json` extractor (#1368), before the handler and before the query. `password` deliberately gets no hook — it is bcrypt-compared, never bound, so guarding it would change behaviour for a request that works today. The search filters need no hook either: they arrive in the query string, which the guard now covers.

Both halves are NUL-only: no request without a NUL changes status, body or routing. The refusal keeps the properties #3665 pins — a 400 `VALIDATION_ERROR` (the OCI distribution-spec envelope on `/v2/`), emitted inside correlation-id so it still carries `X-Correlation-ID`, and byte-identical for a real repository, an unknown one and an unrouted path, so answering before authentication is no existence oracle.

`ARCHITECTURE.md`'s middleware table and flowchart are updated (layer 3 is now "NUL guard", path + query, with a pointer to the field hook for bodies), as is the registration comment in `routes.rs`.

## Verified

Throwaway Postgres 16 container, `AK_TESTS_REQUIRE_DB=1`, all cargo runs under the repo's compile lock.

- `cargo fmt` — clean.
- `cargo nextest run --lib -E 'test(nul_path) or test(nul_) or test(login_request)'` — **22 passed**.
- `cargo nextest run --lib -E 'test(/nul_path|search|auth::|login|extractors/)'` — **1014 passed, 0 failed** (the wider blast radius: search handlers + service, every auth/login test, the `Json` extractor).

**Fails on main.** Backing out only the two functional hunks (the query check in `nul_path_guard`, the `deserialize_with` attribute on `LoginRequest::username`) and keeping the tests:

```
FAIL api::handlers::auth::tests::login_request_rejects_a_nul_in_username
FAIL api::middleware::nul_path::tests::nul_in_search_query_params_is_400_not_db_500_db
FAIL api::middleware::nul_path::tests::nul_in_login_username_is_400_not_db_500_db
FAIL api::middleware::nul_path::tests::nul_query_refusal_carries_correlation_id_and_is_byte_identical_db
Summary: 4 tests run: 0 passed, 4 failed
```

with `assertion left == right failed: /api/v1/search/advanced?name=a%00b  left: 500  right: 400` — the exact 500 the issue reports.

Not done: no attempt to enumerate other JSON body fields that reach a bind — this PR covers the one site the issue measured (`LoginRequest::username`) and the whole query-parameter class; a systematic body-field sweep is a separate change.


brandonrc:
## Audit — round 2 (after the cursor rework), plus the two round-1 reviews

**Verdict: merge on green.** Round 1 (correctness/integration, security) established the path+query guard and the `LoginRequest` field hook, and found the one carrier the fast path cannot see: the base64-over-JSON `?cursor=`. Round 2 verified the rework: `decode_keyset_cursor` rejects a NUL in either decoded component and falls through the existing malformed-cursor `400 VALIDATION_ERROR` arm (no new error path); a sweep of every `base64`/`from_str`/`Query<` candidate under `backend/src/api` found no other self-encoding query parameter; 32/32 selected tests on a live DB, and reverting only the cursor hunk fails exactly the three cursor tests while the OCI query test correctly stays green. `check`, `clippy -D warnings`, `rustfmt --check` clean.

Landed in the last commit: the two test comments that had lost the six-character backslash-u-zero-zero-zero-zero escape text (it had become a plain space — comment-only, no raw NUL anywhere in the tree), the module doc's "refused at their own decoders" narrowed to what is true (the cursor is; body strings only where the field opts in, #3713 has the rest), the same softening in `ARCHITECTURE.md` row 3, and the "four listing queries" count corrected to six statements with the flat-listing note. Main was merged in twice for the CHANGELOG (#3682, #3709); one `### Security` heading, `#3673` first ref, zero markers, preflight passes.

Behaviour note stands as written: a NUL anywhere in a query string is now 400, including `?q=`/`?query=`, which previously stripped the byte and answered 200.

PR #3711, opened 2026-09-03
fix(security): stop 41 handler-local error paths returning raw driver text

### LINKED ISSUE #3667: Handler-local `Database error: {e}` bodies leak raw Postgres text on 39 sites that never call db_err (sibling of #3623)
## Summary

#3623 / #3666 sanitise `api::handlers::db_err`, the shared helper. A second group of handlers never call it: they build the response body themselves with `format!("Database error: {}", e)` and so still surface the raw sqlx/Postgres message (schema names, constraint names, `invalid byte sequence for encoding "UTF8"`, …) to the client — on public repositories, unauthenticated.

Verified on `main` (`d15156cc`+), excluding test modules:

| file | sites |
|---|---|
| `backend/src/api/handlers/gitlfs.rs` | 15 |
| `backend/src/api/handlers/protobuf.rs` | 11 |
| `backend/src/api/handlers/swift.rs` | 8 |
| `backend/src/api/handlers/promotion.rs` | 2 |
| `backend/src/api/handlers/maven.rs` | 1 |
| `backend/src/api/handlers/npm.rs` | 1 |
| `backend/src/api/handlers/pub_registry.rs` | 1 |
| `backend/src/api/handlers/mod.rs` | 1 (the one #3666 fixes) |

`grep -rnE 'format!\("Database error: \{' backend/src/api/handlers --include=*.rs`

Several of these are inside format-specific error envelopes (Git LFS's `{"message": …}`, OCI-style JSON, protobuf), so the fix is not a blanket swap for `db_err`: each needs the *format's* envelope kept and only the message replaced with the stable `Database operation failed` text, with the raw error going to a server-side `tracing::error!` as `db_err` now does.

## Ask

Per handler, replace the interpolated `e` with the sanitised message (a small shared `fn db_err_message(e) -> &'static str` next to `db_err` would keep the 39 sites to one line each) and add one regression test per file that feeds a driver-shaped error and asserts the body does not contain it. The #3500-style structural guard (`grep` for `format!("Database error: {` in `handlers/`) would stop the class returning.

Refs: #3623, #3666, #3545 (the NUL-byte 500 that made these bodies visible).

brandonrc:
## Summary

`#3623`/`#3666` sanitised `handlers::db_err`, the shared DB-error mapper. A second, larger group of sites never called it: they hold a *format-specific* error envelope and so built the body themselves with `format!("Database error: {}", e)`, handing the caller the raw sqlx/Postgres text — schema and constraint names, `invalid byte sequence for encoding "UTF8": 0x00` — on routes that serve public repositories anonymously.

All **39** sites the issue enumerates are fixed, exactly the count on `main` after #3666:

| file | sites | envelope kept |
|---|---|---|
| `gitlfs.rs` | 15 | Git LFS `{"message": …, "request_id": …}` |
| `protobuf.rs` | 11 | Connect `{"code","message"}` |
| `swift.rs` | 8 | `application/problem+json` `{"detail": …}` |
| `promotion.rs` | 2 | promotion result's `message` field |
| `maven.rs` | 1 | plain-text 500 |
| `npm.rs` | 1 | `{"error": …}` |
| `pub_registry.rs` | 1 | plain-text 500 |

Each site keeps its own envelope and takes its message from a new **`handlers::db_err_message`**, placed next to `db_err`/`db_status`: it logs the raw error server-side exactly as `AppError::into_response` does for `db_err`, and returns the stable text the sanitised envelope carries — `Database operation failed`, or `Database connection pool is saturated, retry shortly` when the pool is saturated, which the existing `db_status` already sheds to 503 + `Retry-After` (#2083). **One line per site; no status code changes.**

The two sites from the issue's comment go with them:

- `error_helpers::require_signing_key` — reached anonymously from `repomd.xml.asc` / `repomd.xml.key` (`rpm.rs`) and `Release.gpg` (`debian.rs`); it interpolated the key-lookup failure (a DB message, a decrypt failure, unparseable key material). #2636's 500-vs-404 split is untouched: no key is still 404, a key that will not load is still a loud 500 — now loud in the log.
- `incus::fs_err` — interpolated the raw IO error, which carries absolute storage paths. The operation name is a fixed internal literal and stays (`Failed to create temp file`); the error does not.

## Structural guard

`raw_db_error_body_class_tests::no_handler_builds_a_raw_database_error_body_3667`, in the shape of the #3500 gate and reusing its `rust_sources` / `test_module_line_ranges` walkers (both widened to `pub(super)`). It refuses any `.rs` file under `api/handlers`, outside its `#[cfg(test)]` modules and comment lines, that contains a format string beginning `Database error: {`. Keyed on the string literal rather than on `format!(`, so a `write!`, a rustfmt-wrapped `format!` or an `AppError::Internal(format!(…))` is caught too. It is a **lint, not a proof** — a site that invents different wording walks past it — and the per-handler tests are the behavioural pin.

## Verified

Compile lock, `CARGO_TARGET_DIR` in the worktree, `SQLX_OFFLINE=true`; DB tests against a local `postgres:16-alpine` with `backend/migrations` applied and `AK_TESTS_REQUIRE_DB=1`.

- `cargo nextest run --lib -E 'test(3667)'` — **18 passed**.
- `cargo nextest run --lib -E 'test(/^api::handlers::(gitlfs|protobuf|swift|npm|maven|promotion|pub_registry|incus|error_helpers|tests|like_pattern_escape_class_tests|raw_db_error_body_class_tests)/)'` — **879 passed**.
- `cargo nextest run --lib -E 'test(/^api::handlers::(rpm|debian)::/)'` — **267 passed** (the `require_signing_key` callers).

**Fails without the fix.** With the 39 sites and the two comment sites reverted to their `main` form (helper and tests kept), 5 of the 18 fail:

- `raw_db_error_body_class_tests::no_handler_builds_a_raw_database_error_body_3667` — lists all 39 sites by file:line.
- `gitlfs::tests::locks_db::create_lock_database_error_body_carries_no_driver_text_3667` — DB-backed; a NUL in the JSON lock path makes Postgres reject the bound parameter, and the Git LFS envelope came back carrying `invalid byte sequence for encoding "UTF8": 0x00`.
- `protobuf::tests::test_get_modules_database_error_body_carries_no_driver_text_3667` — DB-backed and **anonymous** (`GetModules` takes no auth extension), same mechanism through the Connect envelope.
- `error_helpers::tests::test_require_signing_key_error_body_carries_no_driver_text_3667`
- `incus::tests::test_fs_err_body_carries_no_raw_io_error_3667`

**Honest limitation.** The remaining per-file tests (`swift`, `npm`, `maven`, `pub_registry`, `promotion`, and the envelope-builder tests for `gitlfs`/`protobuf`) are unit-level on each format's envelope builder, as the issue's shape allows: they prove the envelope + `db_err_message` composition carries no driver text, but they do **not** fail when only their own site is reverted. Those sites are not reachable with a driver-shaped error from a test — the #3622 boundary guard rejects a `%00` in any URL path, so only the two body-driven paths above can be driven end to end. The structural gate is what pins the other 37.

Nothing else was changed: no restyling of the surrounding code, no status codes, no adjacent leaks (`format!("Storage error: {}", e)` in `swift.rs` and the `Failed to {}: {}` shapes outside `incus` are the same class in a different phrasing and are **not** in this PR's scope).

Closes #3667


brandonrc:
## Audit — 2 independent reviewers (sites/envelopes; gate/integration)

**Verdict from both: merge on green.** All 39 `format!("Database error: {}", e)` production sites on main plus the two comment-flagged ones (`require_signing_key`, `incus::fs_err`) are gone, each an in-place one-line substitution: blanking the 39 message lines makes every one of the seven files byte-identical to main up to its test module, so every envelope shape (Git LFS `{message, request_id}`, Connect `{code, message}`, Swift `problem+json`, npm `{error}`, promotion, plain text) and every status are unchanged. `db_err_message` returns exactly the strings `AppError::user_message` returns, and its log line was captured with a live subscriber: one ERROR event with the raw driver text for a real database error, one WARN `POOL_EXHAUSTED` for a pool timeout. Counterfactual with all 41 production hunks reverted fails exactly the five tests the PR names, and the gate lists all 39 sites in the exact per-file distribution. 879 and 1109 handler tests green on live Postgres; #2636's 500-vs-404 split in `require_signing_key` intact.

Landed as `77e6f11b` (merged with main in `ba53eddf`) from the findings:

- `maven.rs` and `pub_registry.rs` kept a hard-coded 500 next to the new message, so a pool timeout would have advised "retry shortly" on a 500 with no `Retry-After`; both now use `db_status(&e)` like the other 37.
- Those two files' per-file tests rebuilt the envelope inline and would have stayed green with the production site deleted. Each site now goes through a small file-local builder that the test drives; with the old `format!` restored inside the builder, both tests fail with the leaked driver text.
- The `fs_err` justification was wrong on two counts: every production call site is behind `write:artifacts`, and a bare `std::io::Error` from `tokio::fs` never embeds a path. Comments and the CHANGELOG now say so; the redaction stays.
- One arm away from the fixed `fs_err` site, `seal_staging_file`'s `NotFound` arm formatted the absolute staging path into the 500 body. The path now goes to `tracing::error!` and the body is fixed text, pinned by a test that fails on the reverted arm.
- The `#2726` doc comment is back on the test it belongs to.

**Code Coverage check:** the red is the jscpd duplication gate on whole changed files (6.3% over `protobuf.rs`/`promotion.rs`), not a test failure. The flagged blocks are the pre-existing near-identical `.map_err` arms this PR touched one line each; it is not a required check, and refactoring 17 error arms is out of scope for a security fix.

**Filed #3718** for what the reviewers measured outside the swept phrasing: `Storage error: {e}` in `swift.rs`, the `hex.rs` signing-key lookup (the `rpm.rs` case in other words), roughly twenty more storage/IO interpolations across six handlers, and the gate's `*`-leading-line comment heuristic.

Issue #3548, opened 2026-08-24
POST /api/v1/auth/ci/token requires `provider_id`, which can only be discovered with admin credentials


vivekdsr:
## Summary

The CI OIDC token exchange requires a `provider_id` (the UUID of the `ci_oidc_providers` row). The only endpoint exposing that UUID is `GET /api/v1/admin/ci-oidc`, which requires an admin bearer token.

This means a CI pipeline must authenticate with the **admin password** before it can perform the keyless OIDC exchange — which defeats the purpose of keyless CI auth.

**Version:** backend 1.8.0

## Steps to reproduce

1. Configure a CI OIDC provider and a mapping matching your CI's claims.
2. In a CI job, obtain an OIDC JWT (e.g. GitLab `id_tokens`, GitHub Actions OIDC).
3. Attempt to exchange it for an Artifact Keeper token without any pre-existing credential.

```bash
# This is all a CI job should need:
curl -X POST "$AK_URL/api/v1/auth/ci/token" \
  -H "Authorization: Bearer $CI_OIDC_JWT" \
  -H 'Content-Type: application/json' \
  -d '{"provider_id":"???"}'   # <- where does a CI job get this?
```

## Actual behaviour

`provider_id` is mandatory, and the only way to obtain it is an admin-authenticated call:

```bash
ADMIN_TOKEN=$(curl -s "$AK_URL/api/v1/auth/login" \
  -d '{"username":"admin","password":"'"$ADMIN_PASSWORD"'"}' | jq -r .access_token)

PROVIDER_ID=$(curl -s "$AK_URL/api/v1/admin/ci-oidc" \
  -H "Authorization: Bearer $ADMIN_TOKEN" | jq -r '.[0].id')

curl -X POST "$AK_URL/api/v1/auth/ci/token" \
  -H "Authorization: Bearer $CI_OIDC_JWT" \
  -d "{\"provider_id\":\"$PROVIDER_ID\"}"
```

So the "keyless" flow begins by using the most privileged credential in the system.

## Expected behaviour

A CI job should be able to exchange its OIDC JWT for an access token using only the JWT.

## Why this can't be worked around cleanly

- `CreateCiOidcProviderRequest` accepts `issuer_url`, `name`, `audience`, `is_enabled`, `provider_type` — there is no `id` field. The UUID is server-generated, so it cannot be chosen or made consistent across deployments.
[...]

PR #3714, opened 2026-09-03
fix(security): let a public repository satisfy the token repo-scope check on the OCI, tree and artifact read gates

### LINKED ISSUE #3704: Token repo-scope is checked before the public-read bypass on three more surfaces (OCI /v2, tree, artifacts) — same shape as #3648
## Summary

#3648 / #3703 fixed the #504 token repository-scope gate in `repo_visibility_middleware`: a token scoped to repository A was refused a **read** of **public** repository B (403) that the same request got with no credential at all (200). #3703 short-circuits that gate with the existing `public_read_satisfies_acl` idiom for GET/HEAD/OPTIONS.

While fixing it the agent found the identical ordering — scope check before the public-read bypass — on three other surfaces, none of which go through that middleware:

1. **OCI `/v2`** — `enforce_token_repo_scope` (#2290), a separate mount nested outside the middleware. A scoped token pulling a public image from another repository is refused where anonymous succeeds.
2. **`handlers::tree::tree_access_allowed`** — the storage-tree/browse endpoints.
3. **`handlers::artifacts::check_artifact_visibility`** — the REST artifact read path.

Each is a real instance of "a credential grants less access than none", and each is reachable by a client that legitimately holds a repo-scoped token (CI tokens are the common case — see #3548 for why those are scoped).

## Ask

Apply the same `public_read_satisfies_acl(is_public, action)` short-circuit before each scope check, reads only; writes keep enforcing scope. One router-level test per surface mirroring #3703's shape (public GET out-of-scope → 200 and byte-identical to anonymous; public PUT/DELETE out-of-scope → 403; private GET out-of-scope → 403). Also decide the three `is_non_mutating_format_post` POST routes (git-lfs batch, conan authenticate, VS Code gallery) that #3703 deliberately excluded.

Refs: #3648, #3703, #504, #2290, #2329.

brandonrc:
## Summary

#3703 fixed the #504 token repository-scope gate in `repo_visibility_middleware`. The identical ordering — scope ceiling **before** the public-read bypass — stood on three other gates, none of which that middleware covers. On each, a request carrying a repository-scoped API token was refused a **read** of a **public** repository that the same request got with **no credential at all**.

| surface | with no credential | token scoped to another repo (before) |
| --- | --- | --- |
| `GET /v2/{public}/app/tags/list` (and manifests, blobs, `_catalog`) | 200 | **403 `DENIED`** |
| `GET /api/v1/tree?repository_key={public}` (and `/tree/content`) | 200 | **404** |
| `GET /api/v1/artifacts/{id-in-public}` (and `/metadata`, `/stats`) | 200 | **403** |

`/v2` is nested before `.merge(format_routes)` in `routes.rs`, so no middleware authorizes an OCI pull: `enforce_token_repo_scope` (#2290) ran ~8 lines ahead of the `public_read_satisfies_acl` bypass inside `oci_read_permitted`, and again at the top of each of the four read handlers, while an anonymous puller carries the `anonymous` sentinel for which `authenticate_oci_read` returns `None` and the entire claims block is skipped. `tree_access_allowed` tested `token_allows` before its `is_public` arm. `check_artifact_visibility` ran the #504 check before anything looked at visibility at all. CI tokens are scoped exactly this way (#3548), so the credential most likely to hit this is a pipeline's.

All three now take the same `public_read_satisfies_acl(is_public, action)` baseline #3648 gave the middleware, and that `require_visible` — the canonical REST read gate on the very same repositories — has applied as an unconditional public short-circuit all along (#2329).

## What a scoped token gains on each surface

Exactly what an anonymous caller was already getting:

- **OCI `/v2`** — `GET`/`HEAD` of a public repository's manifests, blobs and tag lists: a public `docker pull`.
- **Tree browser** — `GET /api/v1/tree` and `GET /api/v1/tree/content` on a public repository.
- **REST artifacts** — `GET /api/v1/artifacts/{id}` and its `/metadata` and `/stats` siblings for an artifact in a public repository.

The sub-resources that share `check_artifact_visibility` but require a credential of their own (label listing, quality-gate reads, scan/finding reads) move to the same public-read baseline. That is a real widening beyond the strict anonymous baseline and is called out in the CHANGELOG: it is the baseline `require_visible` already applies unconditionally to the same repositories, so the alternative was two REST read gates disagreeing about the same public repository.

## Nothing else moved

- **Writes and deletes are unchanged.** `public_read_satisfies_acl` is `read`-only by construction. `docker push`, manifest `DELETE`, and the artifact-label `PUT`/`POST`/`DELETE` that share the artifact gate all keep enforcing the scope. That matters most on the label routes: `authorize_label_write` checks only the token's *action* scope, so `check_artifact_visibility` is the **sole repository ceiling** there — which is why the fix could not be an unconditional `is_public` short-circuit, and why the gate now takes the calling handler's `action` explicitly at all fifteen call sites.
- **`enforce_token_repo_scope` itself is untouched.** The read path calls a new `enforce_token_repo_scope_on_read`; `require_oci_repo_write_access` and `try_mount_blob` (a *push*-side read of the mount source, with no anonymous baseline) still call the unexempted form.
- **The OCI scan-pull pin is not relaxed** (`enforce_scan_pull_scope`, #2093) — a scan token is minted for exactly one repository key.
- **`/v2/_catalog` is deliberately excluded**, even though `authorized_catalog_repo_ids` shares `oci_read_permitted` with the per-repository handlers by design. Enumeration has no anonymous baseline to have fallen below — anonymous gets the 401 `WWW-Authenticate` challenge there, never a public-only listing (`anonymous_catalog_keeps_401_challenge`) — and the repo-scope ceiling is precisely what stops a bearer restricted to one repository from learning which others exist. The catalog filter re-applies the unexempted `enforce_token_repo_scope` itself, and the #3704 fixture pins it (its assertion that public repo B stays out of a repo-A token's listing is the deterministic guard; the pre-existing `repo_scope_ceiling_confines_catalog_set` has no public out-of-scope row and only trips on cross-test rows in a shared database).
- **Private repositories are unchanged** on all three; the existence-hiding answers are untouched.
- `tree.rs`'s `test_access_public_out_of_token_scope_denied`, which pinned the inverted behaviour, is **flipped**, not deleted.
- Per the issue's second comment, one probe is added to #3703's own fixture pinning that the gallery POST's read reclassification does not leak to a **private** repository.

## Regression tests

Three router-level, DB-backed tests, one per surface:

- `api::handlers::oci_v2::public_read_repo_scope_3704::test_3704_public_image_pull_is_not_refused_to_an_out_of_scope_token` — real API token pinned through `api_token_repositories`, driven through `router()`.
- `api::handlers::tree::tests::test_3704_public_repo_tree_is_not_refused_to_an_out_of_scope_token`
- `api::handlers::artifacts::public_read_repo_scope_3704::test_3704_public_repo_artifact_read_is_not_refused_to_an_out_of_scope_token`

Each asserts the scoped answer is **identical (status and body) to the anonymous one** on every read verb, plus a positive control in scope, the anonymous baseline, and the negative cases.

**The negative cases are deliberately not vacuous.** Both OCI ceilings emit the same `oci_denied_repo_access()` envelope, so — unlike #3703's middleware — the body cannot say which gate refused. Instead the caller is granted real `read`/`write`/`delete` on the out-of-scope repositories, so the scope ceiling is the only thing left that can produce those refusals: widen the exemption to writes and the push/label-PUT succeeds, widen it to private repositories and the private read succeeds. The artifact gate additionally asserts the exact `AppError` body, which discriminates its scope 403 from the private-repo `NotFound` below it.

## Verified

Merged `origin/main` (`567a5b08`, #3703) into the branch — no conflicts. Against a throwaway Postgres 16 with migrations applied, under the compile lock, with `AK_TESTS_REQUIRE_DB=1` so a missing database is a hard failure rather than a silent skip:

| step | result |
| --- | --- |
| `cargo fmt --check` | clean |
| `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` | clean |
| `cargo nextest --lib -E 'test(3704) or test(3648) ...'` (the three new tests, the two flipped/new tree unit tests, and #3703's fixture with the added private-gallery probe) | **6 passed** |
| `cargo nextest --lib --no-fail-fast -E 'test(oci_v2) or test(handlers::tree) or test(handlers::artifacts) or test(artifact_labels) or test(middleware::auth) or test(quality_gates) or test(handlers::security) or test(handlers::sbom) or test(repo_scope) or test(visibility) or test(satisfies_acl) or test(can_access_repo)'` | **1374 passed** |
| `scripts/ci/check-changelog-unreleased.sh` | passes |

**Six counterfactuals, each run against the live database, not reasoned:**

| mutation | result |
| --- | --- |
| OCI read gate reverted to `enforce_token_repo_scope` | **FAIL** — `left: (403, "DENIED", "You do not have access to this repository"), right: (200, "", "")` — the issue's exact pair |
| tree `is_public` arm moved back after the scope check | **FAIL** — `left: (404, …"not found"), right: (200, …nodes…)`, and the flipped unit test fails too |
| artifact public-read bypass removed | **FAIL** — `left: (403, FORBIDDEN), right: (200, …artifact JSON…)` |
| artifact bypass widened to every action | **FAIL** — the out-of-scope label `PUT` **succeeds** (200) where 403 is required |
| OCI read exemption leaked into the write gate | **FAIL** — the out-of-scope `docker push` reaches `MANIFEST_INVALID` (400) instead of 403 `DENIED` |
| `_catalog` carve-out removed | **FAIL** — the #3704 fixture (public repo B appears in the repo-A token's listing) |

The write and private counterfactuals are only meaningful because the fixtures grant the caller real `read`/`write`/`delete` on the out-of-scope repositories, leaving the scope ceiling as the only thing that can refuse them.

Not run: the full `--lib` suite or the `backend/tests` integration binaries (build cost); no end-to-end run with a real `docker` client.

Closes #3704


brandonrc:
## Audit — 2 independent reviewers (OCI + widening; tree/artifacts/integration)

**Verdict from both: merge on green.** The three gates are reordered correctly, the `action` threading is right at all fifteen call sites, and every negative assertion proved load-bearing under mutation: widening the artifact gate to all actions made an out-of-scope label `PUT` return 200 *and write the label*; widening any gate to private repositories flipped the private negatives to 200; removing the `_catalog` carve-out put public repo B in a repo-A token's listing. Neither reviewer found a path where the exemption gives an OCI caller more than an anonymous `docker pull`. Broad selectors 1374/1374 and 1382/1382 against live Postgres, `check`/`clippy` clean, merge with main clean, preflight exits 0.

**On #3717 (asked explicitly):** this PR moves none of #3717's private rows and adds no new 403-vs-404 split. `/v2` keeps private-out-of-scope at `403 DENIED` and a missing key at `404 NAME_UNKNOWN`; tree has no discriminator at all; REST artifacts is #3717 verbatim. Only the *public* out-of-scope row moves, to the anonymous answer. Two new assertions (`artifacts.rs` ~826-833, `oci_v2.rs` ~35194-35201) pin the current private bodies and will re-baseline with #3717.

**Corrected in this PR:** the claim that the pre-existing `repo_scope_ceiling_confines_catalog_set` guards the `_catalog` carve-out. Its fixture's only public repo is inside the token's scope, so it has no public out-of-scope row; both reviewers found it fails under the counterfactual only because other suites' public docker repos leak into its exact-set assertion in the shared DB. The deterministic guard is the new fixture's own assertion. PR body amended; the #3648 residual paragraph in the CHANGELOG now points at this entry instead of saying the three surfaces are unchanged.

**Recorded decision (not a code change):** `check_artifact_visibility`'s `read` exemption also reaches credential-only call sites, so a repo-A token can now read artifact labels and scan findings on public repo B where it previously got 403. Scans and quality gates already served the same data through repo-keyed siblings behind `require_visible`'s public short-circuit; artifact labels are the one row without such a sibling. Merging as written: a scoped token should sit between anonymous and its owning user, and any authenticated user can already read (and, pre-existing on main, write) labels in a public repository. The CHANGELOG paragraph is the record.

Nits left as-is: `TREE_READ_ACTION` is a constant so the tree check collapses to `if is_public` (holds while both tree routes are `GET`); `try_mount_blob` is now stricter than the direct blob read it shadows (conservative).

PR #3719, opened 2026-09-04
docs(helm): fix comments left describing the absolute-URL approach (#3680)

### LINKED ISSUE #3724: helm.rs: comments and one test still describe the pre-#3682 absolute-URL behaviour
Splitting this out so the cleanup PR has something to close — noted in the #3682 review as worth a follow-up, and they are my comments.

After #3682 switched the advertised chart URLs from absolute to relative, three spots in `backend/src/api/handlers/helm.rs` still describe the absolute approach:

1. `test_helm_index_url_resolves_for_bare_path_generic_upload`'s assertion message says the URL is *"absolute, so that `helm dependency update` does not join it onto the repo URL"*. That join is now the mechanism the fix relies on, not the thing it avoids — so the comment states the opposite of what the code does.
2. Two comments (the chart round-trip and the `.prov` round-trip) explain that the advertised base is the `RequestBaseUrl` `http://localhost` fallback because the test request carries no `Host` header. That extractor was removed from this path in the same PR, so the explanation refers to code that no longer exists.

Separately, `test_helm_chart_url` builds a string with `format!` and asserts it equals the same literal:

```rust
let url = format!("/helm/{}/charts/{}", repo_key, filename);
assert_eq!(url, "/helm/helm-local/charts/ingress-nginx-4.8.0.tgz");
```

It exercises no production code, and the literal it pins is the root-relative form the code no longer emits — so it now reads as a spec for the old behaviour. The advertised URL is already covered by the `test_chart_download_url_*` unit tests and by the two DB-backed round-trips that fetch it.

No behaviour change; comments and one vacuous test.

juwon8891:
Closes #3724

Follow-up to #3682, which you offered to tidy — these are my comments, so I'd rather not leave them for you. Filed as #3724 so this has a linked issue.

- [x] N/A — comments and a vacuous test; no production behaviour changes

## Summary

Three spots still described the round-1 absolute-URL approach after the merged fix switched to relative:

1. `test_helm_index_url_resolves_for_bare_path_generic_upload`'s assertion message said the URL is *"absolute, so that `helm dependency update` does not join it onto the repo URL"*. That join is now the mechanism the fix relies on, not the thing it avoids — the message now says relative, so the join resolves instead of doubling.
2. Two comments (the chart round-trip and the `.prov` round-trip) explained that the advertised base is the `RequestBaseUrl` `http://localhost` fallback because the test request carries no `Host` header. That extractor is gone from this path; both now say what the test actually does — the same join a client performs, against a router mounted without the `/helm` nest prefix.

Also drops `test_helm_chart_url`. It built a string with `format!` and asserted it equalled the same literal, so it never exercised production code — and the literal it pinned was `/helm/{repo}/charts/{file}`, the root-relative form the code no longer emits. The advertised URL is covered by `test_chart_download_url_*` and by the two DB-backed round-trips that fetch it.

## Test Checklist

- [x] Unit tests added/updated — one removed, none added
- [x] Manually tested locally — helm module **73/73** against a live Postgres (`AK_TESTS_REQUIRE_DB=1`); `cargo fmt --check` clean

## API Changes

- [x] N/A - no API changes

### DIFF
--- a/backend/src/api/handlers/helm.rs
+++ b/backend/src/api/handlers/helm.rs
@@ -1662,14 +1662,6 @@ wsDcBAEBCgAQBQJqWW7VCRA8wAoTVPCkgwAAVAoMACmQbvnhlkWncOkVJXfissGD\n\
         assert_eq!(key, "helm/grafana/7.0.0/grafana-7.0.0.tgz");
     }
 
-    #[test]
-    fn test_helm_chart_url() {
-        let repo_key = "helm-local";
-        let filename = "ingress-nginx-4.8.0.tgz";
-        let url = format!("/helm/{}/charts/{}", repo_key, filename);
-        assert_eq!(url, "/helm/helm-local/charts/ingress-nginx-4.8.0.tgz");
-    }
-
     /// Pins the PATH arm on its own: an OCI-shaped `artifacts.path` with a
     /// content type that matches NO entry in the media-type list. If the
     /// `path.starts_with("v2/") && path.contains("/manifests/")` arm is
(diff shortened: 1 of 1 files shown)


brandonrc:
Apologies — this one is on me, and it is the second time I have done it.

You opened this at 00:19 UTC on 4 September. I opened #3720 at 00:23, four minutes later, doing the same cleanup, and merged it at 01:43 without checking whether a PR for #3724 already existed. It did: yours. The result is that main now carries an equivalent change — `test_helm_chart_url` is gone and all three comments describe the relative-URL join — and your branch conflicts with it on wording alone.

For the record, your version reads slightly better in one place. Yours says the test "performs the same join a client does"; mine explains the mechanics without naming the client behaviour, which is the part that actually justifies the test. That framing is the one I would keep if I were choosing now.

Closing this as superseded, with the duplication being my fault rather than anything about the change. #3724 is already closed. You are credited in the 1.9.0 release notes for it (PR to follow), and I have tightened my own checklist so the next fix wave looks for an open PR against the issue before writing anything.

Thank you for cleaning up after your own review comments — that is a good habit and I am sorry it cost you the round trip.

Issue #911, opened 2026-04-27
Backend: GET /cache-ttl default (3600) diverges 24x from proxy default (86400)


brandonrc:
## Bug

`get_cache_ttl` at `backend/src/api/handlers/repositories.rs:445-472` returns `3600` when no `repository_config` row exists for the repo. The proxy itself uses a different default: `backend/src/services/proxy_service.rs:24` defines `DEFAULT_CACHE_TTL_SECS = 86400`, which is what `get_cache_ttl_for_repo()` (`proxy_service.rs:301-325`, called at `:162`) actually returns when no row exists.

## Impact

User-visible inconsistency. `GET /api/v1/repositories/:key/cache-ttl` reports `{"cache_ttl_seconds": 3600}` (1 hour), but the proxy evicts cache entries at 86400 seconds (24 hours). The reported value is off by 24x from the value the system actually applies. Operators who tune TTL based on the GET response will be wrong.

## Fix direction

Pick one source of truth:

1. Extract `pub const DEFAULT_CACHE_TTL_SECS: i64 = 86400;` in a single module (e.g., `proxy_service`) and import it from the handler so both paths return the same default.
2. Or: have repository creation insert a `repository_config` row explicitly so neither path falls through to a hardcoded default.

Option 1 is the smaller change. Option 2 is more correct long-term.

## Provenance

Surfaced by multi-agent review of artifact-keeper-test#80.

Tracking: ak-din8.

PR #3721, opened 2026-09-04
fix(api): report the effective mutable-path cache TTL when a repository has no override

### LINKED ISSUE #3706: Remote repository uses 300s cache TTL instead of the configured default 86400s for mutable paths
### Description

When a remote repository is created with the default **Proxy Cache TTL** value (`86400` seconds), this value is not stored in `repository_config` table in database.

This appears to be intentional: `repository_config` stores repository-level overrides, while the default value of `86400` is provided by the application.

However, mutable cache paths use a different fallback when no repository-level override exists.

The relevant [logic](https://github.com/artifact-keeper/artifact-keeper/blob/b2c3cf356f4b0bba544cd0937bd2dd4b5751c97a/backend/src/services/proxy_service.rs#L5574-L5580) is:

```rust
cache_classifier::Mutability::Mutable { default_ttl_secs } => {
    // A repo-level override still applies to mutable paths; fall
    // back to the conservative classifier default otherwise.
    self.get_cache_ttl_override(repo.id)
        .await
        .unwrap_or(default_ttl_secs)
}
```

For an OCI manifest requested by tag, for example:

```text
/v2/library/busybox/manifests/1.38.0
```

the path is classified as mutable and:

```text
default_ttl_secs = MUTABLE_DEFAULT_TTL_SECS = 300
```

Because the default repository Proxy Cache TTL (`86400`) was never persisted to `repository_config` table, `get_cache_ttl_override(repo.id)` returns `None`.

As a result:

```text
unwrap_or(default_ttl_secs)
```

falls back to `300` seconds instead of the repository's advertised default of `86400` seconds.

### Environment Information

- Deployment type: Docker
- Architecture: x86_64
- Database: PostgreSQL 18
- Reverse Proxy: Caddy

### Backend Version

1.8.2

### Web UI Version

1.8.0

### Package Format

Docker / OCI

### Expected Behavior

The effective cache TTL for mutable paths is:

```text
300 seconds
```

even though the remote repository has the default Proxy Cache TTL:

```text
86400 seconds
```

### Observed Behavior

The effective TTL should be consistent with the repository's Proxy Cache TTL.

For a newly created remote repository with the default configuration:

```text
Proxy Cache TTL = 86400
```

mutable paths should therefore use:

```text
86400 seconds
```

unless a different behavior is explicitly intended and exposed as a separate configuration option.

### Steps to Reproduce

1. Create a new remote Docker repository.
2. Leave **Proxy Cache TTL** unchanged at its default value of `86400` seconds.
3. Pull an image through the remote repository, for example:

```bash
docker pull <registry>/<remote-repository>/library/busybox:1.38.0
```

4. Check the proxy cache logs.

The cached tag manifest expires after 5 minutes:

```text
Cached artifact .../manifests/1.38.0/__content__
(... bytes, expires at 2026-09-03 20:36:12 UTC)
```

when it was cached at approximately:

```text
2026-09-03 20:31:12 UTC
```

5. Check `repository_config` table in database.

There is no `cache_ttl_secs` entry for the repository.

### Logs

```shell
2026-09-03T20:31:11.607796Z DEBUG artifact_keeper_backend::services::proxy_service: Proxy cache miss: entry expired cache_key=.../manifests/1.38.0/__content__ expires_at=2026-09-03 16:47:54.616779004 UTC

2026-09-03T20:31:11.610773Z INFO artifact_keeper_backend::services::proxy_service: Fetching artifact from upstream: https://registry-1.docker.io/v2/library/busybox/manifests/1.38.0

2026-09-03T20:31:12.576701Z INFO artifact_keeper_backend::services::proxy_service: Fetched 9535 bytes from upstream (content_type: Some("application/vnd.oci.image.index.v1+json"), etag: Some("\"sha256:dc2d74b28e4cf8984fa52af1f39bc7c3d9c73760b41a74d629f5d11b1ab28616\""), link: None)

2026-09-03T20:31:12.652884Z DEBUG artifact_keeper_backend::services::proxy_service: Cached artifact .../manifests/1.38.0/__content__ (9535 bytes, expires at 2026-09-03 20:36:12.624482092 UTC)
```

### Additional Context

If the Proxy Cache TTL is explicitly changed/saved, a `cache_ttl_secs` override is persisted in `repository_config` table in database.

In that case:

```rust
get_cache_ttl_override(repo.id)
```

returns `Some(value)`, and the configured repository TTL is correctly used for mutable paths.

This means the effective behavior differs depending on whether the same value was explicitly persisted:

```text
New repository, default TTL 86400, not persisted
    -> mutable TTL = 300

Explicitly save TTL 86400
    -> mutable TTL = 86400
```

This makes the effective cache behavior inconsistent with the repository configuration shown to the user.

### Pre-submission Checklist

brandonrc:
## Summary

Closes #3706

A Remote repository created with the default Proxy Cache TTL stores no `cache_ttl_secs` row in `repository_config`. Since #1611 the proxy has deliberately applied the cache classifier's `MUTABLE_DEFAULT_TTL_SECS` (300 s) to *mutable* paths on such a repository — tag manifests, `maven-metadata.xml`, the PyPI simple index, npm packuments, the Cargo index — so that a newly published version appears promptly (design record: `docs/audits/design-clean-room-singleflight-cache-2026-06-04.md`, and the npm packument / cargo index / Debian `dists` caches independently chose the same 5 minutes). But `GET /api/v1/repositories/{key}/cache-ttl` — and therefore the UI — still reported the pre-#1611 `DEFAULT_CACHE_TTL_SECS` (86400) for a repository with no row, so the reporter believed their busybox tag manifest was cached for a day when it expired after five minutes.

This PR makes the read side truthful with **zero runtime cache behaviour change**:

- `resolve_cache_ttl(None)` (what `GET …/cache-ttl` reports when no override is stored) now returns `cache_classifier::MUTABLE_DEFAULT_TTL_SECS`, the value `ProxyService::cache_ttl_for_path` actually applies to mutable paths. A stored override is reported as before.
- `proxy_service::DEFAULT_CACHE_TTL_SECS` (86400) had no remaining production reader and is removed; its test-only uses become a test-local `ONE_DAY_SECS` fixture value, and the test-only `parse_cache_ttl` mirror falls back to the classifier constant.
- `cache_ttl_for_path` becomes `pub(crate)` so the handler test can compare reported vs applied.
- No `docs/` or OpenAPI text mentioned the 86400 default; nothing else to correct. The web UI's create-form placeholder lives in `artifact-keeper-web` and may still show 86400.

An operator who wants 86400 s on mutable paths sets it per repository via `PUT /api/v1/repositories/{key}/cache-ttl`, exactly as today.

## Tests

- `repositories.rs`: `test_resolve_cache_ttl_reports_mutable_path_default_for_repo_without_override_3706` (pure: reported default `== 300`, `!= 86400`); the #911 guards updated to the new constant, and the structural guard now also forbids an inline `unwrap_or(300)`.
- `repositories.rs`: `get_cache_ttl_reports_what_the_proxy_applies_to_mutable_paths_3706` (DB-backed): fresh `remote`/`docker` fixture, real `GET /:key/cache-ttl` through the router vs `cache_ttl_for_path` on `v2/library/busybox/manifests/1.38.0` — equal with no override (300) and equal after storing an override (600).
- `proxy_service.rs`: `test_default_cache_ttl_is_24_hours` removed with the constant; `test_compute_cache_expiry_default_ttl` renamed `_one_day_ttl`.

## Verification

Postgres 16 in Docker, `sqlx migrate run` (211 applied), all cargo through the compile lock with `SQLX_OFFLINE=true`:

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(3706) | test(3459) | test(cache_ttl) | test(proxy)'` with `DATABASE_URL`, `AK_TESTS_REQUIRE_DB=1` — **1173 run, 1173 passed** (both #3706 tests, both #3459 negative controls, the #911 `resolve_cache_ttl` guards, and the whole proxy family, against the live DB)
- Counterfactual: with `resolve_cache_ttl` falling back to `86400` (main's behaviour), both #3706 tests fail with `left: 86400, right: 300` — the pure test on the reported default, the DB test on the no-override agreement assertion. Restored afterwards; tree clean.
- Merged `origin/main` (through #3711) into the branch; the only conflict was `CHANGELOG.md`, resolved to main's `[Unreleased]` plus this one bullet (verified: one `### Fixed`, one `### Security`, no duplicated bullets). Post-merge `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` and `cargo check --lib --tests` — clean; `cargo fmt --check` — clean.
- Not run: the full backend suite and the e2e proxy suites.


brandonrc:
## Audit — 2 reviewers on the first version, round 2 on the rework

**Verdict: merge on green.** The first version implemented the issue as asked (mutable-path fallback 300 s → the repository's 86400 s default) and the semantics review returned NEEDS-WORK with evidence I found decisive: the 300 s mutable default was designed in by #1708 (the clean-room single-flight cache record says a newly published version must appear in the simple index / packument promptly), three sibling caches independently chose 300 s, and `cache_classifier` has one mutable TTL, so the change would have moved every never-configured Remote across Maven metadata, PyPI simple index, npm packuments, OCI tags, cargo index, repomd, Helm index and NuGet to day-stale indexes with no global knob, while silently falsifying the npm and cargo 5-minute bounds layered above the disk cache. The reporter agreed with the alternative on the issue.

The rework (head `7b5f2b85`) fixes the inconsistency on the reporting side instead: `resolve_cache_ttl(None)`, which `GET /repositories/{key}/cache-ttl` and the UI read, now returns `cache_classifier::MUTABLE_DEFAULT_TTL_SECS`; the dead `DEFAULT_CACHE_TTL_SECS` is deleted. Round 2 verified live: the only non-test hunks are that swap, the constant deletion, a doc comment and `cache_ttl_for_path` made `pub(crate)` with a byte-identical body; `cache_classifier.rs`, `maven.rs` and `sbt.rs` are identical to main and the #3459 pins still bound by 300. 42/42 on the selector against Postgres 16; with an 86400 fallback restored both `_3706` tests fail (`left: 86400, right: 300`), and the DB test asserts the GET body equals what `cache_ttl_for_path` applies to a tag manifest with and without an override. `clippy -D warnings` and `fmt` clean; CHANGELOG under `### Fixed`, `#3706` first, states plainly that no caching behaviour changes; the pending section is main's plus this one bullet.

Nit left as-is: a test comment names the deleted constant in its "how to reproduce the old value" note.

If a different *default* for new Remotes is wanted, that is a feature with its own issue; per-repository `PUT …/cache-ttl` already covers the reporter's case.

PR #3726, opened 2026-09-04
fix(security): refuse a NUL byte in the four JSON body fields that still reached Postgres as a 500

### LINKED ISSUE #3713: NUL in JSON body fields still reaches Postgres on POST /api/v1/auth/sso/exchange (unauthenticated) and three lower-tier endpoints — sibling of #3673
## Summary

#3710 closes the query-string half of #3673 at the middleware and the `username` field of `POST /api/v1/auth/login` with a serde hook. Its audit found the remaining body-side carriers:

| endpoint | field | auth | what happens |
|---|---|---|---|
| `POST /api/v1/auth/sso/exchange` | `ExchangeCodeRequest.code` (`handlers/sso.rs:1200-1202`, no validation) | **none** (`routes.rs:587`) | bound as `$1` in `DELETE FROM sso_exchange_codes WHERE code = $1` (`auth_config_service.rs:1854`) → `AppError::Internal` → **500**. Structural twin of the login case. |
| `POST /api/v1/builds` | six bound text fields | any logged-in user | 500 |
| `POST /api/v1/auth/tokens` | `CreateApiTokenRequest.name` | logged-in | 500 |
| `POST /api/v1/auth/tickets` | `CreateTicketRequest.purpose` | logged-in | 500 — its sibling field `resource_path` already rejects every byte `< 0x20` (`handlers/auth.rs:872-891`) |

Also measured, **not a 500**: `Authorization: Basic` carries a NUL-bearing username to the same `users` bind (Postgres logs `invalid byte sequence for encoding "UTF8": 0x00`), but the middleware collapses `AppError::Database` into a 401 identical to the control — pool checkout and ERROR line survive, no client-visible oracle.

## Ask

Apply #3710's `deserialize_nul_free_string` hook to the four fields above (the `sso/exchange` one first — it is anonymous), with a router-level test each. A generic answer — the hook on every `String` field of every `Json<T>` request type, or a `Json<T>` rejection layer that walks string values — would close the class; the per-field hook is the shape #3710 chose and justified (bodies are streams; binary uploads must not be scanned).

Refs: #3673, #3710, #3665, #3545.

brandonrc:
## Summary

#3710 closed the query-string half of #3673 and the `username` field of `POST /api/v1/auth/login`. Its audit measured four more JSON body fields that still carried a NUL to a sqlx bind and answered 500 — one of them, `POST /api/v1/auth/sso/exchange`, anonymously. This applies the same serde hook to those four sites, in the same shape as the login fix:

- `ExchangeCodeRequest.code` (`handlers/sso.rs`) — anonymous
- `CreateBuildRequest` (`handlers/builds.rs`) — `name` plus the five optional bound text fields, via a new `deserialize_nul_free_opt_string` variant of the hook (the existing one only covers `String`)
- `CreateApiTokenRequest.name` and `CreateTicketRequest.purpose` (`handlers/auth.rs`)

Each now answers `400 VALIDATION_ERROR` before any query runs. The structural answer the issue floats (the hook on every `String` field of every `Json<T>`) is deliberately not attempted; the CHANGELOG says so. The `Authorization: Basic` carrier is out of scope as documented on the issue (401-swallowed, no oracle).

**Deviation worth a reviewer's eye:** `sso.rs` and `builds.rs` were still on `axum::Json`, so the hook's error rendered as a 422 plain-text body rather than the 400 `VALIDATION_ERROR` envelope #3710 produces. Both modules now import the repo's own `crate::api::extractors::Json` (the documented drop-in, #1368). Side effect: any malformed body on `/auth/sso/exchange`, the LDAP login route and `/builds` moves from 422 to 400. Called out in the CHANGELOG bullet.

Note: the ticket route is `/api/v1/auth/ticket` (singular), not `/tickets` as the issue says.

## Tests

Router-level, DB-backed (`AK_TESTS_REQUIRE_DB=1`): `exchange_code_rejects_a_nul_in_code_before_the_query`, `create_build_rejects_a_nul_in_any_bound_text_field` (loops all six fields, plus an absent-optional control), `create_api_token_rejects_a_nul_in_name_before_the_query`, `create_download_ticket_rejects_a_nul_in_purpose_before_the_query`. The NUL is built from a Rust string escape; no raw NUL byte exists in the diff.

Counterfactual with the ten attributes stripped (Json swap kept): all four fail with 500 (`INTERNAL_ERROR` for sso, `DATABASE_ERROR` for the other three).

## Verified

Under the compile lock against Postgres 16 with migrations: `cargo check --lib --tests` clean; `nextest -E 'test(3713) or test(nul) or test(sso) or test(builds) or test(tokens) or test(tickets)'` 453/453; `clippy --all-targets -D warnings` clean; `cargo fmt --check` clean. Rebased onto main after #3710 merged (`ededf1b9`); one `### Security` heading under `[Unreleased]`, `#3713` first ref.

Closes #3713


brandonrc:
## Audit — 2 independent reviewers (security completeness; correctness/tests)

**Verdict from both: merge on green.** All four sites the issue lists carry the serde hook and it fires inside the `Json` extractor before any bind. Counterfactual with the attributes stripped and the extractor swap kept: all four tests fail with the measured 500s (`INTERNAL_ERROR` for sso, `DATABASE_ERROR` for the rest). Single-attribute mutations fail exactly their own test; removing `#[serde(default)]` from an optional field fails exactly the absent-optional control. The new `Option<String>` variant is correct on `null`, absent, empty and a NUL at any position; non-NUL control characters pass by design, consistent with #3673. 545/545 on the selector superset and 1040/1040 on `test(auth)` against live Postgres, zero `invalid byte sequence` lines from the server; `check`/`clippy`/`fmt` clean; no raw NUL byte anywhere in the diff.

The security sweep beyond the issue's table found no anonymous 500: `/auth/refresh` and `/totp/verify` decode the JWT before any bind (401), the LDAP login route reaches LDAP rather than a bind, `/groups` is admin-gated, `/uploads` is still on `axum::Json` (422). `POST /builds` sits under optional auth, so a NUL body now answers 400 anonymously where a valid one answers 401 — not new, no oracle.

Landed as `8b838e70` (merged with main in `5a2e0f8a`) from the findings:
- Both reviewers measured `POST /api/v1/builds/{id}/artifacts` still answering 500 on a NUL: `BuildArtifactInputPayload`'s four bound text fields in the same module were unhooked. Hooked, with a looping router test whose counterfactual fails with `DATABASE_ERROR`.
- The `axum::Json` → crate `Json` swap moves 415 as well as 422 to 400, because `map_json_rejection` maps every rejection; 413 does not move. The CHANGELOG now carries a labelled upgrade note naming the five exact routes.
- Every test now also asserts the 400 body carries no `database`/`utf8` driver text, matching the #3710 precedent.

Left as documented: the `Authorization: Basic` carrier (401-swallowed, no oracle) and the structural every-`String`-field answer the issue floats.

PR #3728, opened 2026-09-04
fix(security): answer the token-scope gate's read denial on a private repository with the existence-hiding 404 on native, OCI and artifact routes

### LINKED ISSUE #3717: Repository-scoped API tokens get 403 on a private repository they are not scoped to but 404 on a nonexistent key — the #504 gate is now the lone existence oracle on native reads
## Summary

After #3709, every native-format read denial for a private repository answers `404 Repository not found` — except one. The #504 token-scope gate (`auth.rs:~2392`) runs before the ACL arm and answers `403 Token does not have access to this repository` for an existing private repository outside the token's `allowed_repo_ids`, while a key naming no repository answers 404 from the no-repo branch (`auth.rs:~2271`). Reproduced live on the #3709 head by two reviewers.

Repository-scoped tokens are self-service (`repo_tokens.rs:~380-445`), so any user can mint one scoped to a repository they *do* hold and use it to probe every other key: 200 (in scope), 403 (exists, out of scope), 404 (does not exist). #3703 sharpened this by letting public repositories through before the scope gate, which is correct, but it means the 403 now names exactly "private and exists". The REST equivalent (`repositories.rs:~436-438`, `require_visible`) already answers 404 for the same condition; #3714 touches the ordering of this gate on OCI/tree/artifacts and its audit should confirm which of these it changes.

Currently pinned by `test_3648`'s `private_get` assertion (`auth.rs:~9394`), which will need updating.

## Ask

Answer the scope-gate denial with the existence-hiding 404 (same body as the no-repo branch) for reads on private repositories; keep 403 for writes to match #3524's decision. Regression test: scoped token → GET on an out-of-scope private repo and on a nonexistent key produce identical status and body; control: in-scope → 200.

Refs: #504, #3524, #3709, #3703, #3704, #3714, #3648.

brandonrc:
Closes #3717

## Summary

After #3709 every native-format read denial on a private repository answered the existence-hiding `404 Repository not found` — except the #504 token-scope gate, which runs ahead of the ACL arm and answered `403 Token does not have access to this repository` for an existing private repository outside the token's `allowed_repo_ids`, while a key naming no repository answered 404. Repository-scoped tokens are self-service, so any user holding one for a repository of their own had a 200/403/404 existence oracle over every other key; #3648/#3704 (correctly) letting public repositories through first sharpened the 403 to mean exactly "private and exists".

This PR makes a **read** refused by the scope gate on a **private** repository answer the same existence-hiding response that surface already gives for a missing key, byte for byte, on each surface the gate lives on:

| surface | gate | before | after |
|---|---|---|---|
| native format routes | `repo_visibility_middleware` (`auth.rs`) | `403 Token does not have access to this repository` | `404 Repository not found` (`not_found_response()`, the no-repo branch's answer) |
| OCI `/v2` | `enforce_token_repo_scope_on_read` (`oci_v2.rs`) | `403 DENIED` | `404 NAME_UNKNOWN` `repository not found: {key}` (what `resolve_repo_inner` answers an unknown key) |
| REST artifacts | `check_artifact_visibility` (`artifacts.rs`) | `403 FORBIDDEN` | `404 NOT_FOUND` `Artifact not found` (what `get_artifact` answers an unknown id) |
| tree browser | `authorize_tree_read` (`tree.rs`) | already `404 Repository '{key}' not found` | unchanged |

The tree browser was audited and left alone: `authorize_tree_read`'s `not_found()` closure (line 88) and the missing-key `ok_or_else` (line 217) build the identical `AppError::NotFound`, so it never had the split.

**Writes keep 403** on every surface, matching #3524's decision for the ACL arm. **Public repositories are untouched** — the #3648/#3704 read baseline still runs first.

`enforce_token_repo_scope_on_read` gains a `repo_key: &str` parameter so the 404 can echo the key the way the unknown-key path does; its five call sites pass `&repo.key` / `repo_key`.

## Re-baselined assertions

Three assertions pinned the old 403 and are re-baselined to the new 404 rather than deleted, with the message saying so:

- `auth.rs` `test_3648_…`: `private_get` and `vscode_private` now expect `not_found()` (new helper next to `scope_denied()`). Because the scope gate's read refusal and the rules-less private branch are now byte-identical, the fixture additionally **grants the caller on repo C**: without that, `vscode_private` could no longer tell a reclassification that dropped the `is_public` half (falls through to the rules-less branch → 404) from the scope gate refusing (404). With the grant, that fall-through answers 200, so the assertion still discriminates. The doc-comment table is updated.
- `oci_v2.rs` `test_3704_…`: `private_scoped` now expects `(404, NAME_UNKNOWN, "repository not found: {key_c}")`.
- `artifacts.rs` `test_3704_…`: `private_scoped` now expects `not_found()` (new helper).

## Regression tests (fail on main)

One per changed surface, each DB-backed, each comparing the out-of-scope private read against the nonexistent-key read as full `(status, body)` / `(status, code, message)` pairs, with an in-scope **private** control (200), a positive control that the missing key really is 404, and a write on the out-of-scope private repo that must still be 403. The caller is granted read on the out-of-scope repo so the scope gate is the only thing that can refuse it.

- `api::middleware::auth::tests::test_3717_private_repo_scope_denial_matches_the_no_repo_answer` (GET + HEAD, PUT control)
- `api::handlers::oci_v2::public_read_repo_scope_3704::test_3717_private_image_scope_denial_matches_the_unknown_key_answer` (tags list, manifest GET + HEAD, manifest PUT control)
- `api::handlers::artifacts::public_read_repo_scope_3704::test_3717_private_repo_artifact_scope_denial_matches_the_missing_artifact_answer` (by-id GET, `/metadata`, label PUT control)

## Verification

All against a live PostgreSQL 16 container with `AK_TESTS_REQUIRE_DB=1` (so no DB-backed test could silently skip), `SQLX_OFFLINE=true`, migrations applied with `sqlx migrate run`.

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(3717) or test(3648) or test(3704) or test(504) or test(scope)' --no-fail-fast` — **444 passed, 0 failed** (includes the three new `test_3717_*`, the re-baselined `test_3648_*` and both `test_3704_*`, and the tree `test_3704_*`).
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(oci) or test(visibility) or test(tree) or test(artifacts)'` — **1009 passed, 0 failed**.
- **Counterfactual:** with only the three production hunks reverted to main's code (tests left in place), `-E 'test(3717)'` → **3 run, 0 passed, 3 failed**, each on the read-equality assertion with the old body:
  - middleware: `left: (403, "Token does not have access to this repository")` vs `right: (404, "Repository not found")`
  - artifacts: `left: (403, {"code":"FORBIDDEN",…})` vs `right: (404, {"code":"NOT_FOUND","message":"Artifact not found"})`
  - OCI: `left: (403, "DENIED", "You do not have access to this repository")` vs `right: (404, "NAME_UNKNOWN", "repository not found: {key}")`
- `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` — clean (a first run flagged `///` doc comments on a `let` closure in the new OCI test; converted to `//`).

Not run: the e2e / integration suites and the frontend; nothing outside `backend/src` changed.

## Deliberately not touched

- `oci_read_permitted`'s ACL-arm denial still answers `403 DENIED` when the private repository carries fine-grained rules and `404 NAME_UNKNOWN` when it does not (the OCI counterpart of what #3524/#3709 unified in the middleware). That is a separate issue from the scope gate and is out of scope here.
- Writes on a nonexistent key still answer 404 from the no-repo branch while writes on an out-of-scope private repo answer 403 — the tradeoff #3524 accepted for the ACL arm, kept as-is.


brandonrc:
## Audit — 2 independent reviewers (security oracle; correctness/integration) plus a correctness re-check on the reconciled head

**Verdict: merge on green, with one last addition landing.** The three production hunks are correct and each is pinned by a regression test that fails on main's code. Middleware: reaching the scope-denial branch on a `read` implies a private repository (one-line boolean proof, unaffected by `with_scope_gated_admin`), so the 404 there is exactly the no-repo body; widened to the read-equivalent POSTs (`git-lfs objects/batch`, conan `users/authenticate`) that the ACL arm already classifies via `non_mutating_post`, guarded to private repositories after `test_3648`'s public pins caught the widening. OCI: all five `enforce_token_repo_scope_on_read` callers updated, mount/catalog/write gate correctly on the unexempted variant, and after the reconcile with #3729 every non-member read denial on `/v2` flows through main's `requested_repo_key`/`oci_name_unknown`, so the body is byte-identical to the missing-key branch in Docker-mirror mode too (mirror-mode tests from both PRs pass). REST artifacts: all 15 `check_artifact_visibility` sites enumerated, only `read` moves, and the unknown-id error is the same `NotFound("Artifact not found")` variant. Re-baselined assertions pin exact new `(status, body)` values. Mutation with the three hunks reverted: 6 of 7 fail with the exact old 403 bodies. Sets: 444/444, 1009/1009, 337/337 after the reconcile, `test(oci)` 759/759; `check`/`clippy`/`fmt` clean.

Landing: none of the three scope-gate denials logged anything, and with the 403 gone the operator could no longer tell "no such repository" from "token not scoped to it" either; each site now emits the same `tracing::info!(repository_id, user_id, …)` the ACL arms use, and the upgrade note says where to look.

Filed **#3750** on timing: `repo_cache` stores only matched rows, so a repeated probe of an existing out-of-scope key is ~10x faster than a missing key (p10–p90 bands do not overlap in-process). Pre-existing, the last discriminator on this surface for a scoped token; tracked separately with a negative-cache proposal.

PR #3729, opened 2026-09-04
fix(security): answer the OCI /v2 private-repository read denial with 404 NAME_UNKNOWN regardless of whether ACL rules exist (#3716)

### LINKED ISSUE #3716: OCI /v2 read gate keeps the #3524 oracle: docker pull of a private repository answers 403 DENIED when it carries ACL rules and 404 NAME_UNKNOWN when it does not
## Summary

#3524 (fixed by #3709) closed the 403-vs-404 oracle on the native-format read arm of `repo_visibility_middleware`. The OCI distribution endpoints are mounted outside that middleware (`routes.rs:~161` vs `~125`), and their own read gate has the identical construct. Confirmed live by two independent reviewers on the #3709 head.

`oci_read_permitted` (`backend/src/api/handlers/oci_v2.rs:~862-885`) consults `has_any_rules_for_target` on a denial purely to pick the status:

| caller: authenticated, no grant | result |
|---|---|
| private repo, any ACL rule exists (for any principal) | `403 DENIED` |
| private repo, no rules | `404 NAME_UNKNOWN` |
| no such repo (`oci_v2.rs:~2953-2959`) | `404 NAME_UNKNOWN`, byte-identical to the row above |

So `docker pull` still tells an ungranted caller that a private repository exists *and* that an ACL governs it — the three-row table from #3524, verbatim. The doc comment at `oci_v2.rs:~768-775` argues the opposite from the same facts and should be corrected with the fix.

## Ask

Collapse the read-denial to `404 NAME_UNKNOWN` (the existence-hiding answer the same function already gives for a missing repository), matching what #3709 did in `auth.rs`, and leave pushes at 403 as #3524 decided for writes. Regression test: private repo with a rule for an unrelated principal, authenticated non-member `GET /v2/<repo>/manifests/<tag>` → same status and body as a nonexistent repo; counterfactual currently 403.

Refs: #3524, #3709, #1808, GHSA-fv45-mwhh-q23r.

brandonrc:
## Summary

`docker pull` of a private repository by an authenticated caller holding no grant answered `403 DENIED` when the repository carried any fine-grained permission rule (for any principal) and `404 NAME_UNKNOWN` when it did not — while a key naming no repository at all answered that same `404 NAME_UNKNOWN`. The status therefore disclosed both that the repository exists and that an ACL governs it: the #3524 oracle, which #3709 closed on `repo_visibility_middleware`'s native-format read arm. `/v2` is mounted outside that middleware and its own read gate, `oci_read_permitted`, still consulted `has_any_rules_for_target` on a denial purely to pick the status.

- `oci_read_permitted` now answers the private-repository read denial with the same `404 NAME_UNKNOWN` body the nonexistent-repository branch (`resolve_repo_inner`) produces, so the ruled, rules-less and no-such-repository cases are identical in status, `Content-Type` and body. The rules lookup is dropped from the read path (nothing else there needed it); the denial is logged server-side at `debug` with the repository and user ids (once per denied pull, so a key-probing token cannot fill the log at `info`).
- The gate's doc comment, which argued for the split ("its existence is already implied by the ACL it carries"; `DENIED` as "the correct terminal answer for an authenticated-but-unauthorized pull"), is corrected, along with the `oci_read_permitted` summary that described the "shaped denial".
- Pushes and deletes keep `403 DENIED` (#3524). Public repositories, members, admins, scanner tokens and anonymous callers (`401` + `WWW-Authenticate`) are unaffected — the gate is only reached for private repositories, after the #3714 public-read short-circuit. A `403 DENIED` on `/v2` now means only a denied write or a token-scope refusal (#3717 tracks the latter).
- CHANGELOG: one `### Security` bullet under `[Unreleased]`, directly after the #3524 bullet it completes.

## Review follow-ups (correctness audit: MERGE, three nits applied)

- Denial log lowered `info` → `debug`; PR text no longer cites #3709 as precedent for the log line (it added none).
- Doc comments now also name the 503 `DENIED` of `oci_authorization_unavailable()` as a `DENIED` arm.
- CHANGELOG bullet trimmed to one paragraph plus the upgrade note.

## Security audit follow-ups (MERGE-WITH-NITS)

- **F1 (should-fix, Docker-mirror mode):** the gate's 404 echoed the *resolved* `repo.key` while the missing-key branch echoed the client's first path segment. Under `AK_DEFAULT_DOCKER_MIRROR_REPO` an unknown key is re-resolved to the mirror, so a non-member of a private mirror got `repository not found: <mirror key>` for an unknown key but `<candidate>` for an existing private repository — echo == candidate ⇒ exists. Fixed by threading the client-supplied segment: `requested_repo_key(image_name)` (now also what `resolve_repo_inner` splits on) and `oci_name_unknown(requested_key)` (the one 404 builder both branches use); `require_oci_repo_read_access` takes `image_name`, `oci_read_permitted` takes `requested_key` beside `repo_key` (the scan-pin check still needs the resolved key). Both helpers are named for #3728 to reuse in `enforce_token_repo_scope_on_read` after merge. New test `test_3716_mirror_mode_denial_echoes_the_requested_key_not_the_mirror` sets the env var before the first request (per-process cache; relies on nextest's process-per-test) and asserts the unknown-key 404 echoes the requested key, never the mirror's, and is identical to the existing-private-repo 404 after normalising the client's own key; fixture controls pin that mirror mode is active and that the unknown key really resolved to the private mirror (anonymous → 401).
- **F2 (wording):** the doc comment and CHANGELOG now name the backend-minted scan-pull pin (`enforce_scan_pull_scope`) as the third `403 DENIED` case beside a denied write and a token-scope refusal.

## Tests

- New: `oci_read_authz_tests::test_3716_private_pull_denial_does_not_reveal_whether_acl_rules_exist` — two private repositories (one carrying a `read` rule for an unrelated principal, one rules-less; fixture discrimination asserted via `has_any_rules_for_target`) and a random nonexistent key, probed with `GET /v2/<repo>/app/manifests/v1` through the real router. Compares `(status, Content-Type, body)` triples, with the client-supplied key normalised out of the echoed message: both private denials must equal the nonexistent-key answer. Controls: members read both (200); the anonymous pair stays uniform at 401.
- Updated the two assertions that pinned the 403: `fine_grained_rules_decide_private_reads` (write-only principal) and `role_assignment_principals_keep_pull_when_an_unrelated_rule_exists` (outsider).
- Untouched and still green: anonymous-challenge, member-allowed, scanner-pin, token-scope (`read_scope_db_tests`, `public_read_repo_scope_3704`) tests — the scope ceilings run before the arm that changed and still answer `403 DENIED`.

## Verification

All cargo invocations went through `rig/compile-lock.sh`; DB-backed tests ran against a live `postgres:16-alpine` (`ak-fix-3716`, migrations applied) with `AK_TESTS_REQUIRE_DB=1`.

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(test_3716_) | test(fine_grained_rules_decide_private_reads) | test(role_assignment_principals_keep_pull_when_an_unrelated_rule_exists)'` — 3 passed.
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(oci)'` — **751 passed, 0 failed** (includes the anonymous-challenge, member-allowed, scanner-pin, `read_scope_db_tests` and `public_read_repo_scope_3704` suites).
- Counterfactual (gate-body hunk reverted, tests kept): `test_3716_private_pull_denial_does_not_reveal_whether_acl_rules_exist` **fails** on the pre-fix gate at the oracle assertion — left: `(403, "application/json", {"errors":[{"code":"DENIED","message":"You do not have access to this repository"}]})`, right: `(404, "application/json", {"errors":[{"code":"NAME_UNKNOWN","message":"repository not found: <key>"}]})` — and passes with the fix (fixture discrimination and positive controls pass in both runs, so the failure is the oracle, not the setup).
- `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` — clean (`Finished dev profile`, no warnings).
- `cargo fmt` applied.

Post-audit round (nits + F1), on the branch merged with `origin/main` (#3711):

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(3716) | test(oci_read_authz)'` (live DB, `AK_TESTS_REQUIRE_DB=1`) — **11 passed** (both #3716 tests plus the nine pre-existing read-gate tests).
- Counterfactual for the mirror-mode test: with only the wrapper reverted to pass `&repo.key` as the echo key, `test_3716_mirror_mode_denial_echoes_the_requested_key_not_the_mirror` **fails** at "the 404 for an unknown key must not echo the MIRROR repository's key" with body `repository not found: ph-test-docker-<mirror uuid>`; passes with the fix.
- `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` — clean; `cargo fmt` applied.

Closes #3716


brandonrc:
## Audit — 2 independent reviewers (security oracle; correctness/integration)

**Verdict from both: merge on green, with three small additions being landed on this branch.** Measured live through the real `/v2` router: an authenticated non-member now gets one byte-identical `404 NAME_UNKNOWN` (status, every header, body modulo the client's own key) on every OCI read verb — manifests GET/HEAD by tag and digest, blobs, `tags/list`, referrers, blob-mount source — for a ruled-private repository, a rules-less private repository and a missing key. Writes keep the #3524 `403`; anonymous and member rows are byte-for-byte what main produces. `has_any_rules_for_target` is gone from every `/v2` read path (the middleware is its only production read-path caller left). Timing: ruled vs rules-less denials 25 µs apart, denied vs missing ~0.5 ms median with overlapping p10–p90, two orders below the #3504 bar. Return type and all six callers unchanged; the Virtual member walk and blob mount do not touch this gate, so no new skip-a-member behaviour. Mutation with the gate reverted fails exactly the new test plus the two re-baselined ones. `check`/`clippy`/`fmt` clean; targeted 10/10; `test(oci)` 751/751 twice on live Postgres. Trial merges onto main and onto #3728 are clean with disjoint hunks.

Being landed from the findings:
- **Mirror-mode echo (shared with #3728):** the denial echoes the resolved `repo.key` while the missing-key branch echoes the client's first path segment. Identical everywhere except Docker-mirror mode, where a non-member of a private mirror could tell "unknown key" (mirror key echoed) from "real private repo" (candidate echoed). The client-supplied segment is now threaded into the denial body, with a mirror-mode test row. #3728 makes the matching change in the scope gate.
- The new denial log fires on every denied pull at `info` while the write gate logs nothing; lowered to `debug`, and the "mirrors #3709" wording dropped (#3709 added no log line).
- The doc comment and CHANGELOG now name all three remaining `403 DENIED` cases: denied write, API-token scope refusal (#3717/#3728), and the backend-minted scan-token pin in `enforce_scan_pull_scope`; the 503 `oci_authorization_unavailable` arm that also carries code `DENIED` is mentioned; the bullet is trimmed to one paragraph plus the upgrade sentence.

Anonymous 401-vs-404 on `/v2` is untouched by this PR and tracked as #3730.

PR #3738, opened 2026-09-07
fix(proxy): finalise the tee digest at Content-Length so the non-SHA-256 commit gate admits matching bodies

### LINKED ISSUE #3487: Proxy-cache commit gate refuses every gated object when the consumer stops reading at Content-Length

brandonrc:
## Summary

Closes #3487

The non-SHA-256 proxy-cache commit gate (GHSA-qxv7-p3mq-88fv, 1.7.4) compares a Maven `.sha1` sidecar or npm `dist.integrity` SHA-512 expectation against a digest the tee computes over the bytes it hands to the cache writer. That digest was only finalised in the block *after* the upstream streaming loop — which runs only when the consumer polls the stream once more and receives `None`. A consumer holding a fixed-length body (hyper with `Content-Length`) stops at the last byte instead, so the block never ran, the writer read an empty slot, and — because a missing digest fails the gate closed — the object was discarded. Every gated object was refused, not just mismatched ones; bytes were still served, and the ungated `.sha1` sidecars kept caching, which hid the symptom.

**Change** (`CachePersister::tee_stream`): count bytes handed to the writer and, on the slice that brings the count to the advertised `Content-Length`, publish the finalised digest *before* that chunk is sent on. The writer therefore always has a completed digest to gate on regardless of whether the consumer polls again. Bodies with no advertised length still finalise at upstream EOF (unchanged). The hasher is `take()`n on publish, so the EOF block cannot double-publish.

**Security posture unchanged**: a mismatched digest is still refused; a truncated body is still rejected by `classify_stream_write` before the digest is consulted; a body that runs past `Content-Length` is rejected by the same length check; SHA-256 expectations read the storage layer's observed checksum, not the tee slot, and were never affected.

## Tests (`backend/src/services/proxy_service.rs`)

- `test_tee_sha1_gate_commits_when_consumer_stops_at_content_length_3487` — the regression: reads exactly `Content-Length` bytes, drops the stream, expects sidecar + live-key publish. **Fails on main** (writer refuses; no sidecar written).
- `test_tee_sha512_gate_commits_when_consumer_stops_at_content_length_3487` — same for the npm SHA-512 path (also fails on main).
- `test_tee_sha1_gate_mismatch_refused_when_consumer_stops_at_content_length_3487` — negative: mismatched sidecar still refused.
- `test_tee_sha1_gate_truncated_upstream_refused_3487` — negative: truncated upstream still refused even though the delivered bytes hash to the expectation.
- `test_tee_sha256_gate_commits_when_consumer_stops_at_content_length_3487` — pins that the SHA-256 path is unaffected.

## Verification

All under `rig/compile-lock.sh`, `CARGO_TARGET_DIR=$S/fix/3487/target RUSTC_WRAPPER=sccache SQLX_OFFLINE=true`:

- `cargo fmt -p artifact-keeper-backend` — clean.
- `cargo check -p artifact-keeper-backend --lib --tests --message-format short` — exit 0.
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(3487) or test(commit_gate) or test(qxv7) or test(proxy_service)' --no-tests=fail` with `AK_TESTS_REQUIRE_DB=1`, `JWT_SECRET=…`, `DATABASE_URL` pointing at a migrated `postgres:16-alpine` container — **382 passed, 0 failed** (15732 skipped by the filter).
- Counterfactual: with only the fix hunk in `tee_stream` reverted (tests kept), `cargo nextest run … -E 'test(3487)'` — **2 failed** (`test_tee_sha1_gate_commits_when_consumer_stops_at_content_length_3487`, `test_tee_sha512_gate_commits_when_consumer_stops_at_content_length_3487`; the writer never wrote a sidecar), 3 passed (both negatives and the SHA-256 pin). Fix restored afterwards; the committed tree is the fixed one.
- `cargo clippy -p artifact-keeper-backend --all-targets --message-format short -- -D warnings` — exit 0.

Not run: the full backend suite outside the filter, and no end-to-end Maven fetch against a live upstream (the reporter's reproduction is against `tee_stream` directly, which is what the new tests exercise).

## Deliberately left out

- No new metric. The buffered content-addressed gate records `ak_proxy_cache_checksum_mismatch_total{repository}`, but the streaming gate's own siblings (SHA-256 mismatch, truncation, empty) record nothing, and `CachePersister` has no `ProxyCacheScope`, so a correctly labelled counter would need the scope or repo key threaded through every `tee_stream` caller — out of scope for a minimal fix. Worth a follow-up.
- The #3335 `tail` flag is still only flipped in the post-loop EOF block, so a consumer that stops at `Content-Length` leaves the publish-window wait unarmed (a concurrent read in that window refetches rather than waits). Pre-existing, separate concern.


brandonrc:
## Audit — security of the commit gate (GHSA-qxv7-p3mq-88fv) and streaming correctness

**Verdict: merge on green.** The early publish is single-shot and covers exactly the bytes handed to the writer, and the gate's property held in every ordering the reviewer tabulated: bytes past `Content-Length` are the bytes that make `bytes_written != expected_len`, so `classify_stream_write` answers `RejectTruncated` before the digest slot is ever read; a short body never reaches the early publish and is rejected on length; the hasher is taken at most once because the forwarded-byte counter is strictly increasing and the compare is exact equality, and a chunk can only straddle the boundary for an over-length body, which is rejected regardless. The no-`Content-Length` path is byte-identical to main and still publishes at EOF. SHA-256 expectations read the stored checksum and are unaffected.

`cargo check` clean; 382/382 on the gate/proxy selector against live Postgres; `clippy -D warnings` clean; with the 16-line hunk removed exactly the two commit tests fail (the mismatch, truncated and SHA-256 tests still pass). CHANGELOG under `### Fixed`, `#3487` first, operator note present (objects refused since 1.7.4 cache again on the next fetch, no action).

Landing on this branch: the #3335 `tail` flag is set only in the post-loop block, so for a fixed-length body the publish-window wait stayed unarmed and a concurrent reader re-fetched instead of waiting — pre-existing, but this fix makes gated Maven/npm objects actually run that tail, so the flag now flips in the same branch that publishes the digest; and a merge with main for the CHANGELOG. Noted, no change: no gate-refusal metric (the streaming persister has no repository scope to label it with; follow-up).

Issue #3586, opened 2026-08-27
Add support for OCI/Docker Blob `Range` Download Requests


erdody:
## Summary

Artifact Keeper `1.8.1` does not appear to support HTTP range requests for OCI/Docker blob downloads.

When a client requests a byte range for a Docker layer blob, Artifact Keeper returns the full object with `200 OK` instead of returning `206 Partial Content` with `Content-Range`.

This affects OCI clients that retry interrupted layer downloads using `Range`, including tools like `skopeo` and potentially container runtimes using Artifact Keeper as a pull-through cache / registry mirror.

## Relevant Specification

The OCI Distribution Spec includes resumable pulls as a use case:

https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#resumable-pull

The blob pull section also states that registries should support the HTTP `Range` request header:

https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#pulling-blobs

## Version

Artifact Keeper backend: `1.8.1`

## Deployment Context

Artifact Keeper is deployed as a Docker/OCI pull-through cache / remote repository.

The repository is public and proxying Docker images from an upstream registry.

## Reproduction

Request a byte range from a cached Docker layer blob:

```bash
curl -i -v \
  -H "Authorization: Bearer anonymous" \
  -H "Range: bytes=0-1048575" \
  "http://artifact-keeper.example.com/v2/nvidia/cuda/blobs/sha256:ed090066e06c7eaa241e8fc89f9ec6272b77d864f132c5f826b0a9a7837b18b9"
```

## Actual Behavior

Artifact Keeper returns the full blob:

```http
HTTP/1.1 200 OK
docker-content-digest: sha256:ed090066e06c7eaa241e8fc89f9ec6272b77d864f132c5f826b0a9a7837b18b9
content-type: application/vnd.docker.image.rootfs.diff.tar.gzip
content-length: 1438285019
```

The response ignores the `Range` header and does not include `Accept-Ranges` or `Content-Range`.

## Expected Behavior

Artifact Keeper should return a partial response:

```http
HTTP/1.1 206 Partial Content
Docker-Content-Digest: sha256:ed090066e06c7eaa241e8fc89f9ec6272b77d864f132c5f826b0a9a7837b18b9
Content-Type: application/vnd.docker.image.rootfs.diff.tar.gzip
Accept-Ranges: bytes
Content-Range: bytes 0-1048575/1438285019
[...]

Issue #3707, opened 2026-09-03
OCI Remote: package is not indexed when client uses HEAD by tag followed by GET by digest


verdel:
### Description

The fix from #3599 does not cover the request flow used by Docker CLI and containerd when pulling a tagged image from a direct Remote repository.

#3599 indexes a proxied package when the manifest is fetched by tag using GET. It also intentionally:

- does not index packages on HEAD;
- does not index digest references.

However, both Docker CLI and containerd were observed to pull a tagged image using:

```text
HEAD /manifests/<tag>
GET  /manifests/<digest>
```

instead of:

```text
GET /manifests/<tag>
```

During `HEAD`, Artifact Keeper fetches and caches the tagged manifest and stores it in `oci-manifests/<digest>`, but correctly does not create a package row.

The subsequent `GET` uses the digest returned by `HEAD` and is served from that local content-addressable storage.

As a result, no `GET` request containing the tag reaches the package-indexing path, and the package is never created.

This reproduces the same externally visible result as #3611 on version 1.8.2, despite the fix from #3599.

### Environment Information

- Deployment type: Docker
- Operating System: Ubuntu 24.04
- Architecture: x86_64
- Database: PostgreSQL 18
- Reverse Proxy: Caddy

### Backend Version

1.8.2

### Web UI Version

1.8.0

### Package Format

Docker / OCI

### Expected Behavior

A successful pull of a tagged image through a remote repository should create the corresponding package/version.

For example:

```text
library/busybox:1.38.0
```

should create a package for `library/busybox` with version `1.38.0`.

`HEAD` alone should still not create a package.

### Observed Behavior

The image is successfully pulled and cached, but no package is created.

The observed flow is:

```text
HEAD manifests/1.38.0
    ↓
manifest fetched from upstream
    ↓
manifest cached as oci-manifests/<digest>
    ↓
no package created

GET manifests/<digest>
    ↓
served from local storage
    ↓
no package created
```

There is no:

```text
GET manifests/1.38.0
```

during the pull.

### Steps to Reproduce

1. Create a direct Docker Remote repository pointing to Docker Hub:

   ```text
   key: dockerhub-proxy
   type: remote
   format: docker
   upstream_url: https://registry-1.docker.io
   ```

2. Ensure the test image is not already present in the proxy cache.

3. Reproduce using either containerd or Docker CLI.

   **Option A — containerd**

[...]

PR #3739, opened 2026-09-07
fix(oci): stream the uncached remote HEAD-blob path instead of buffering at the 8 MiB cap (#3605)

### LINKED ISSUE #3605: HEAD on an uncached remote OCI blob >8 MiB returns 404 while GET returns 200
## Summary

`HEAD /v2/{name}/blobs/{digest}` and `GET` on the same digest disagree for an uncached remote blob larger than 8 MiB: **GET returns 200, HEAD returns 404 `BLOB_UNKNOWN`.** Clients that HEAD before pulling — `skopeo`, containerd — see the blob as absent.

## Detail

`handle_head_blob` routes the cold-cache remote path through the buffered, capped helper:

```rust
// oci_v2.rs, handle_head_blob
if let Some((content, ct)) =
    try_upstream_fetch(&repo, state, &format!("blobs/{}", digest)).await
{
    return build_oci_proxy_response(&content, ct, digest, "application/octet-stream", false);
}

oci_error(StatusCode::NOT_FOUND, "BLOB_UNKNOWN", "blob not found")
```

`try_upstream_fetch` caps at `DEFAULT_METADATA_MAX_BYTES` — `8 * 1024 * 1024` (`proxy_service.rs:45`). A blob over that cap fails the fetch, yields `None`, and falls through to the 404.

`handle_get_blob` does not have this problem, because the same path was already converted to streaming:

```rust
// oci_v2.rs, handle_get_blob
if let Some(resp) = try_upstream_fetch_streaming_blob(&repo, state, digest).await {
    return resp;
}
```

with a comment stating the intent plainly: *"proxy cache) so large blobs succeed with 200 and later pulls are served warm"* (#2192 / #1608 Phase 4c).

So the streaming conversion covered GET and left HEAD on the capped helper. The comment at `oci_v2.rs:3940` asserts *"Blob downloads no longer flow through here"* — true for GET, not true for HEAD.

Container layers are routinely far larger than 8 MiB, so this is the common case for a remote repository rather than an edge case.

## Suggested fix

Route the HEAD cold path through `try_upstream_fetch_streaming_blob` and discard the body, which is what the virtual-member arm of the same handler already does. HEAD must return the same status and headers as GET (RFC 9110 §9.3.2), and this handler already tracks that invariant elsewhere (#3149).

While there: neither `handle_head_blob` nor `handle_get_blob` emits `Accept-Ranges`, so a client cannot discover range support even where it would work. Related to #3586.

## Verification status

The asymmetry is **confirmed by reading** — the two call sites, the constant, and the fall-through are quoted above. The observable `404` against a live remote registry with a >8 MiB uncached layer **has not been reproduced yet** and should be confirmed before the fix is written, since it is a one-command check:

```
# against a Remote OCI repository, layer >8 MiB, cold cache
curl -sI  .../v2/<name>/blobs/<digest>   # expect 404 today
curl -sI -X GET .../v2/<name>/blobs/<digest>  # expect 200
```

Found while scoping #3586 (blob `Range` support). Filed separately because it is a functional regression against large-layer proxying rather than a feature, and it plausibly explains `skopeo copy` failures that have been attributed to missing `Range` support.

irizzant:
## Summary

`handle_head_blob` ran its remote-repo upstream fetch through the buffered metadata helper, whose `DEFAULT_METADATA_MAX_BYTES` (8 MiB) cap made any larger **uncached** layer answer `404 BLOB_UNKNOWN` while the matching GET streamed `200` — so clients that head-before-pull (`skopeo`, `containerd`) saw the blob as absent.

The HEAD path now reuses the same streaming upstream fetch the GET path already uses and answers with the headers that fetch would advertise (`Content-Length`, `Docker-Content-Digest`, content coding) and an empty body. The upstream read is lazy and never polled, so HEAD transfers nothing. This also deletes `try_upstream_fetch`, whose only remaining caller was this path.

## Reproduction (pre-fix, on 1.8.2)

Remote docker repo, upstream serving a > 8 MiB layer:

- `GET  /v2/<repo>/blobs/<digest>` → 200
- `HEAD /v2/<repo>/blobs/<digest>` → 404 `BLOB_UNKNOWN` (cold cache)

## Fix

- `backend/src/api/handlers/oci_v2.rs`: HEAD remote path now calls `try_upstream_fetch_streaming_blob` and returns its `Response` parts with an empty body (same pattern as the virtual `RemoteStream` arm); buffered wrapper removed.
- New regression test `handle_head_blob_streams_uncached_remote_layer_headers`: 9 MiB uncached layer through a direct remote repository via `MockServer` → asserts 200 + `Docker-Content-Digest` + `Content-Length` + empty body. Fails before the fix (404), passes after.

## Verification

- `cargo test --lib 'virtual_blob_streaming_fallback_tests::handle_head_blob'` → **2 passed** (new regression + existing virtual-member HEAD test).
- `cargo fmt` clean on the changed file.

Closes #3605


brandonrc:
Both of your open OCI PRs touch `oci_v2.rs` and conflict with each other, not just with main:

```
$ git merge-tree --write-tree refs/pr/3739 refs/pr/3766
CONFLICT (content): Merge conflict in backend/src/api/handlers/oci_v2.rs
```

I have an adversarial review running on #3766 (Range support, closes #3586) because it has a linked issue and a narrower blast radius, so my plan is to land that one first and then ask you to rebase this one on top — the streaming change and the range change want to agree about how the response body is produced, and doing it in that order means only one of the two has to think about the other.

Say so if you would rather it went the other way round; you know better than I do which of the two you would rather rewrite. Either way this one is not stalled on me, and I will get a review onto it once the ordering is settled.

brandonrc:
@irizzant — this had gone conflicted against `main`, so I pushed a rebase directly to your branch (thanks for `maintainer_can_modify`). The fix itself is yours and unchanged in substance; here is exactly what I touched and why.

**Rebased onto `main` (now `8211ebac`).** Two conflicts:

1. `backend/src/api/handlers/oci_v2.rs` — #3707 landed `index_proxied_tags_for_digest` immediately above `try_upstream_fetch`, i.e. right on top of the function your change deletes. Resolved by keeping the new function and still removing `try_upstream_fetch`. I re-checked that it is still dead after the rebase: the only remaining reference on `main` was the `handle_head_blob` call site your patch replaces.
2. `CHANGELOG.md` — a batch of bullets landed under `### Fixed` today. Kept all of them plus yours; no bullet was dropped or duplicated.

**Two changelog things I fixed while I was in there:**

- Your bullet had drifted into the **released `## [1.9.0]`** section. The branch predates the 1.9.0 cut, so the `### Fixed` heading it was anchored to is no longer the Unreleased one — git merges that without a conflict, which is why it is easy to miss (same trap as #3797). Moved to `## [Unreleased]` / `### Fixed`.
- A stray rebase artifact — the literal text `a1a18c3 (fix(oci): stream the uncached remote HEAD-blob path...)` — had been appended to the end of the unrelated #3702 paragraph. Removed. `CHANGELOG.md` is now a pure `+2 / -0` addition, which is what a changelog diff should look like.

No production code was changed. Net diff is now `+83 / -23` (was `+84 / -24`; the difference is exactly that one stray line).

**What I verified**

- `cargo nextest run -p artifact-keeper-backend --test streaming_invariant` → 2 passed. Worth calling out because `--lib` does not cover this gate, and it is the one that judges a change like this: the `STREAMING-EXEMPT` count in `oci_v2.rs` stays at 1, so removing the buffered call needs no `ALLOWLIST` edit.
- `handle_head_blob_streams_uncached_remote_layer_headers` → passes on the branch; reverting only the production hunks and keeping the test fails with `left: 404, right: 200`, so it genuinely pins #3605.
- **HEAD really does not pull the body.** I wanted more than "the stream is lazy", so I pointed a remote repo at a raw socket serving a 64 MiB blob and counted the bytes it managed to push. HEAD answered `200` with `Content-Length: 67108864` in 54 ms; upstream got 2.5 MiB out (3.9%, i.e. the socket buffer) before the dropped stream closed the connection, and the proxy cache directory was left at 0 bytes. So no bandwidth regression, and a HEAD cannot half-warm the cache either — the tee's writer sees an empty stream and takes the reject path, so no sidecar is written.
- `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check` clean.

**One thing to watch, about your other PR.** #3766 renames `try_upstream_fetch_streaming_blob` to `try_upstream_fetch_streaming_blob_with_range`. This PR adds a *new* call to the old name in `handle_head_blob`. The two merge without a textual conflict — I checked with `git merge-tree`, it is clean — but the merged tree does not compile, because the new call site refers to a function that no longer exists. Whichever lands second needs the one-line follow-up. If this one merges first, #3766 wants:

```rust
if let Some(response) = try_upstream_fetch_streaming_blob_with_range(&repo, state, digest, None).await {
```

You may also want `Accept-Ranges: bytes` on this remote HEAD arm over in #3766, for parity with the two local arms it already covers — I deliberately left that alone here rather than widen this PR.

Nothing else changed. Shout if you would rather take the rebase back over yourself.

irizzant:
Thanks for handling the rebase and the compatibility fix with #3766. I reviewed the final diff and everything looks correct to me. The streaming HEAD path, test, helper removal, and CHANGELOG placement all look good, and CI is green.

I also noted the missing `Accept-Ranges: bytes` on the Remote no-range response. I agree it is non-blocking for #3605 and would prefer to address it separately rather than widen this PR.

I’m happy with the current state and ready for the final review/merge.

PR #3740, opened 2026-09-07
fix(setup): do not arm the setup gate on a deactivated built-in admin

### LINKED ISSUE #3723: Setup gate locks an SSO-only instance out permanently when the built-in admin is deactivated
### Description

An instance where SSO grants admin and the built-in `admin` user has been **deactivated** becomes permanently locked after upgrading into the setup-gate versions. Every request answers `403 SETUP_REQUIRED`, and the documented unlock flow cannot be completed, because it requires logging in as the very account that is deactivated.

`SKIP_ADMIN_PROVISIONING=true` fixes it, but nothing points you there while you are locked out — the 403 body names only the change-password flow.

### Steps to Reproduce

1. Run an older release (we were on 1.2.0), configure OIDC, and deactivate the built-in admin once real admins come from the IdP — `UPDATE users SET is_active = false WHERE username = 'admin'`, or through the UI. This is the natural end state for an SSO-only deployment: the local admin is a credential nobody needs.
2. Upgrade to a release with the setup gate (we went to 1.7.1).
3. Every endpoint now returns `403 SETUP_REQUIRED`, including anonymous reads such as `GET /helm/{repo}/index.yaml`, so `helm repo update` and CI pulls break.
4. Follow the instructions in the 403 body: read `/data/storage/admin.password`, `POST /api/v1/auth/login`.

Result: `401 {"code":"AUTH_ERROR","message":"Invalid username or password"}` — for the freshly generated password *and* for the one in `ADMIN_PASSWORD`, because authentication rejects the deactivated account before the password is ever considered. The only remaining routes are a direct database write or `SKIP_ADMIN_PROVISIONING`.

```
WARN  Admin user has not changed default password. API is locked until password is changed.
WARN  Admin password file missing at /data/storage/admin.password. Regenerating password.
INFO  Admin password written to: /data/storage/admin.password
```

The regenerated password is written and advertised as the way back in, which is what makes this confusing: the credential is valid, the account is not.

### Expected Behavior

Any of these would have avoided the lockout, roughly in order of how much they change:

1. **Don't arm the gate when the built-in admin is deactivated.** `provision_admin_user` already returns `Ok(false)` for `SKIP_ADMIN_PROVISIONING`; an `is_active = false` admin is the same situation — nobody can complete the flow, so gating on it only denies service. If some other principal holds admin (an OIDC group mapping, another `admin`-role user), the instance is already administrable.
2. **Say so in the 403.** The instructions could note that `SKIP_ADMIN_PROVISIONING=true` is the escape hatch when the built-in admin is unavailable, and that the flow requires an *active* account.
3. **Don't regenerate a password for a deactivated account**, or log that it will not work — writing a fresh credential to disk for an account that cannot authenticate sends operators down a dead end.

Also worth considering: the gate blocking *anonymous reads* of public repositories turns a configuration problem into a service outage. Blocking mutations while leaving public reads alone would have kept `helm pull` working while we sorted it out.

### Environment Information

- Deployment type: Kubernetes (chart 1.9.8, appVersion 1.7.1), upgraded from chart 1.2.1 / app 1.2.0
- Auth: Dex OIDC; built-in `admin` deactivated, all other users OIDC-only (no local password)
- Impact: ~35 minutes of `403` on a registry serving Helm charts to CI

### Additional Context

`SKIP_ADMIN_PROVISIONING` (issue #211) resolved it cleanly and is the right long-term setting for an SSO-only deployment — the ask here is that an instance in this state either doesn't arm the gate, or tells you about the switch instead of pointing at a flow that cannot succeed. Happy to send a PR for whichever of the three you'd prefer.

brandonrc:
Closes #3723

## Summary

The setup gate was armed from `users.must_change_password` on an admin row in two places that ignore `is_active`, while local login filters on `is_active = true`. Deactivating the built-in `admin` on an SSO-only instance therefore left every non-auth route (anonymous reads and `/v2` included) at `403 SETUP_REQUIRED`, pointing at a login flow for the one account that cannot log in; boot also regenerated and advertised a fresh `admin.password` for it.

- `AppState::setup_still_required` (`backend/src/api/mod.rs`): the per-request DB re-check now requires `is_active = true`. An instance already locked in this state unlocks on the first blocked request after upgrading, no restart needed.
- `provision_admin_user` (`backend/src/main.rs`): `is_active` is selected alongside `must_change_password`; inside the existing-admin `must_change` branch a deactivated admin logs a warning naming `SKIP_ADMIN_PROVISIONING=true` and returns `Ok(false)` without rotating the hash or writing a password file. Deliberately **not** an `is_active` filter on the lookup's `WHERE`: that would send the row to the create branch, whose upsert overwrites the hash and re-arms the gate on every boot.
- `setup_guard` 403 body (`backend/src/api/middleware/setup.rs`): one more instruction line saying the flow needs an active built-in admin and naming `SKIP_ADMIN_PROVISIONING=true` for SSO-only deployments.
- CHANGELOG bullet under `### Fixed` with the interim SQL unlock as an upgrade note.
- `.config/nextest.toml`: the `middleware::setup::tests` module joins the `db-serial` group, because the new test asserts the cluster-wide `setup_still_required` predicate is *false*, which a sibling's concurrently seeded active pending admin would flip.

Blocking anonymous reads while setup is genuinely pending is unchanged (by design).

## Tests

- `api::middleware::setup::tests::setup_still_required_ignores_inactive_admin_3723` — inactive `must_change_password` admin → `setup_still_required()` is false and the flag latches off; same row reactivated → true.
- `api::middleware::setup::tests::setup_guard_passes_authenticated_request_with_inactive_admin_3723` — through `create_router`, an OIDC admin's bearer `GET /api/v1/users` is `200` with the inactive admin pending (on `main`: `403 SETUP_REQUIRED`).
- `tests::provision_admin_user_does_not_arm_gate_for_inactive_admin_3723` (bin `artifact-keeper`) — returns `Ok(false)`, hash unchanged, no `admin.password` written; reactivated control returns `Ok(true)` and regenerates the file. Note CI runs `--lib` only, so this bin test runs under `cargo nextest run -p artifact-keeper-backend --bin artifact-keeper`.

## Verification

All against a fresh `postgres:16-alpine` (migrations 001-211), `AK_TESTS_REQUIRE_DB=1`, `SQLX_OFFLINE=true`:

- `cargo nextest run -p artifact-keeper-backend --lib --bin artifact-keeper -E 'test(3723) or test(setup) or test(provision_admin)'` — 18 run, 18 passed.
- Counterfactual: with the three production edits reverted (tests kept), `-E 'test(3723)'` — 3 run, 0 passed, 3 failed: the router test got `403 SETUP_REQUIRED`, `setup_still_required()` returned true for the inactive admin, and `provision_admin_user` returned `Ok(true)`. Fix restored, diff byte-identical.

Not run: the full suite; a live boot of the binary against a deactivated admin (the bin test exercises `provision_admin_user` directly instead).


brandonrc:
## Audit — security and correctness of the setup gate

**Verdict: merge on green, with one addition landing on this branch.** The PR closes #3723 exactly as traced: the runtime re-check gains `is_active = true`, boot provisioning returns `Ok(false)` for an inactive built-in admin without rotating the hash or writing `admin.password` (and deliberately keeps `is_active` out of the WHERE so the create-branch upsert cannot re-arm), and the 403 names the escape hatch. No new way for an unauthenticated or non-admin principal to disarm the gate: deactivation is admin-only and 403'd while armed, LDAP sync touches only federated rows, the allow-listed routes cannot change `is_active`, the 403 body has no parsers and leaks nothing, and nothing downstream assumed `Ok(true)`. 18/18 on the selector against a clean live DB; the counterfactual with the three production edits reverted fails all three tests on their real assertions; `check --bins`, `clippy -D warnings`, `fmt` clean. The `db-serial` move is legitimate: the predicate is cluster-wide by design.

Landing from the findings:
- Measured live: after the fix an admin can `PATCH /users/{admin-id}` `{"is_active": true}`, which recreates the exact row that arms the gate, but the latched-off flag is never re-evaluated until restart, and `admin_middleware` never applies the forced-rotation 428 that `auth_middleware` does, so the reactivated admin signs in with its never-changed password. On main the same 428 gap exists but is unreachable because an active pending admin keeps the gate armed. Both halves are being fixed: reactivation re-arms the latch, and admin routes enforce the pending rotation.
- The two new `setup.rs` tests cleaned up only after their assertions, so a failure leaked a row that tripped the bin test's precondition; cleanup is now failure-safe.

Noted: the `ADMIN_PASSWORD`-matches-a-default boot branch still arms at boot for an inactive admin; the first request disarms it safely.

Issue #2844, opened 2026-07-23
Npm virtual repo doesn't searching artifact in other members if in the first member there is an artifact with another version(not searсhing version)


kochetkovSergeY:
### Description

Npm virtual repo does'nt searching artifact in other members if in the first member there is an artifact with another version(not searсhing version)

i have three repos:
1. npm-group
2. npm-forks
3. npm-npmjsorg-proxy

npm-group contains of these two:  npm-forks, npm-npmjsorg-proxy

npm-forks contains projects with versions like
x.x.x-myorg.y 
   -  where 'x.x.x'  is an open-source version contained in npm-npmjsorg-proxy
   - 'y' is a fork version number

i have a situation:
in repo npm-forks there is a package @angular/core with version "19.2.25-myorg.1"
project registry = npm-group
in project used "open-source" version  "19.2.25" of @angular/core
npm install gives me error that in repo npm-group there is no artifact @angular/core with version "19.2.25"
if i remove artifact @angular/core with version "19.2.25-myorg.1" from npm-forks everything works
if i change project registry to npm-npmjsorg-proxy everything works

### Environment Information

- Deployment type: 
- Operating System: 
- Architecture: 
- Database: 
- Reverse Proxy: 
- Network setup: 

### Backend Version

1.6.0

### Web UI Version

1.5.8

### Package Format

npm

### Expected Behavior

Search artifact across ALL members of virtual repo, 

### Observed Behavior

Virtual repo gives error artifact not found, because in first repo there is an artifact with another version(not searсhing version)

### Steps to Reproduce

Create virtual repo with to members:

1. local npm repo
2. npm proxy npmjs.org

Deploy to local repo any artifact that exists in second repo **with custom version**-that not exists in npmjs.org

Create npm project with registry=your virtual repo, add dependency to that artifact with version from npmjs.org.
Call npm install on that project and you will see an error that virtual repo cannot find that artifact 

### Logs

```shell

```

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this isn't a duplicate
- [x] I have redacted any sensitive information (passwords, tokens, private URLs)
- [x] I am using a supported version (not end-of-life)

PR #3743, opened 2026-09-07
fix(npm): resolve every tarball a Virtual packument advertises on that same Virtual

### LINKED ISSUE #3646: fix(npm): virtual repo packument advertises tarball URLs that 404 on that same virtual repo
## Summary

A virtual npm repository's own packument advertises `dist.tarball` URLs that **404 on
that same virtual repository**. Resolution succeeds, then the download fails.

This is a regression in behaviour relative to 1.7.4 and appears to be the unfixed half of
#2844: the **packument leg is fixed** in 1.8.x — virtual repos now merge every member's
versions, priority-aware, in both member orders — but the **tarball leg is not**. The two
now contradict each other.

## Why this is worse than the previous behaviour

Resolution used to fail fast (`ETARGET`), so a bad URL never reached a lockfile. Now
resolution *succeeds* and writes a `resolved:` URL that permanently 404s — so the failure
is recorded in the lockfile and reproduced on every future install, including for everyone
who later checks out that lockfile.

| `npm i` via the virtual | 1.7.4 | 1.8.1 |
|---|---|---|
| `lodash@4.17.21` | `ETARGET` (fails at resolve) | **`E404`** (resolves, then download fails) |
| `@angular/core@19.2.25` | `ETARGET` | **`E404`** |

## Reproduction

Virtual repo with a **local** member holding a fork at a custom version
(`lodash@4.17.21-fork.1`) plus a **remote** member (registry.npmjs.org):

```bash
curl -sS -H "Authorization: Bearer $TOK" "$BASE/npm/<virtual>/lodash" | jq '.versions|length'   # 118 OK
curl -sS -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer $TOK" \
  "$BASE/npm/<virtual>/lodash/-/lodash-4.17.21.tgz"                                            # 404 BAD
curl -sS -o /dev/null -w '%{http_code}\n' -H "Authorization: Bearer $TOK" \
  "$BASE/npm/<remote-member>/lodash/-/lodash-4.17.21.tgz"                                      # 200
```

Holds in **both** priority orders (local p1/remote p2 and remote p1/local p2).

## Root cause

The npm tarball path calls the **name-only** guard:

```
backend/src/api/handlers/npm.rs:3528
    proxy_helpers::virtual_non_remote_owns_name(&state.db, repo.id, package_name).await?
```

`virtual_non_remote_owns_name` (`proxy_helpers.rs:4534`) makes every Remote member a
`DefiniteMiss` as soon as *any* non-Remote member owns the package **name** — regardless of
which versions it actually holds, and regardless of member priority. So a local fork
publishing one custom version of `lodash` suppresses all 100+ upstream versions on the
tarball path, while the packument still lists them.

**A version-aware variant already exists** immediately below it:
`virtual_non_remote_owns_name_version` (`proxy_helpers.rs:4548`), which fires only when a
local member owns a matching version and falls back to name-only suppression when the
version cannot be canonicalized. It is currently reached only through the name-only wrapper
passing `None`.

`cargo.rs:1109`, `rubygems.rs:243` and `hex.rs:759` call the name-only guard too, so the
same contradiction is likely reachable in those formats.

## Expected

The tarball resolver must honour the same union the packument advertises. At minimum the
guard should be **version-aware** (suppress remotes only for versions the local member
actually owns) and **priority-aware**.

## Acceptance criteria

- Every version listed in a virtual repository's packument is downloadable from that
  virtual repository, in both member orders.
- A local member owning one custom version of a name does not suppress upstream versions
  of that name it does not hold.

brandonrc:
Closes #3646

## Summary

A Virtual npm repository's packument advertises `dist.tarball` URLs that 404 on that same Virtual. #2844 fixed the packument leg (every member's versions, merged priority-aware) but the tarball leg in `serve_tarball` still ran the **name-only** shadowing guard `virtual_non_remote_owns_name`, which drops every Remote member as soon as any non-Remote member owns the package *name* — regardless of version or priority. A hosted member holding one fork build (`lodash@4.17.21-fork.1`) therefore 404'd every upstream version the merged packument had just advertised, in both member orders.

The two legs already agreed on URL shape (`{base}/npm/{virtual}/{pkg}/-/{basename}-{version}.tgz`, scoped `@` kept literal, served by the Virtual's own `/-/` routes), so the mismatch was purely the guard.

## Change

- `proxy_helpers::virtual_non_remote_owns_name_exact_version` — new exact-`name@version` variant of the shadowing guard (`LOWER(name) = LOWER($2) AND version = $3`), fail-closed on DB error like its siblings. The existing `virtual_non_remote_owns_name_version` was not reusable: its PEP 440 comparator cannot parse npm prerelease tags (`-next.3`, `-myorg.1`) and fails safe to name-only suppression for them — the same 404.
- `npm.rs::serve_tarball` (Virtual branch) — derives the version from the invariant `<basename>-<version>.tgz` filename via the existing `npm_version_from_tarball_filename` and calls the exact-version guard; a filename that does not carry the package's version keeps the name-only guard (never fan out on a shape we cannot read). The `None`-to-`resolve_virtual_download` security primitive is unchanged: a locally owned `name@version` is still never fanned out to a Remote member.
- CHANGELOG bullet under `## [Unreleased]` → `### Fixed`.

Cargo, RubyGems and Hex still call the name-only guard on their virtual download paths (the issue notes them); deliberately left out of this PR.

## Verification

All against a Postgres 16 container (`DATABASE_URL` set, `AK_TESTS_REQUIRE_DB=1`, `SQLX_OFFLINE=true`), after merging `origin/main`:

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(3646) or test(2844) or test(npm)'` — **460 passed, 0 failed** (includes the new test and the #2844 packument-merge tests).
- `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` — clean; `cargo fmt --check` — clean.

**Regression test:** `api::handlers::npm::tests::test_virtual_advertised_tarballs_download_through_virtual_3646_db` — router-level (`super::router()` via `oneshot`): a Virtual over a hosted member (seeded with only `1.2.3-myorg.1`, bytes on disk) and a wiremock Remote member (`1.2.3` and `2.0.0-next.3`), for an unscoped and a scoped package. It fetches the Virtual's packument, asserts all three versions are advertised, then GETs every `dist.tarball` through the Virtual and checks 200 + the exact bytes, in both member orders (local p1/remote p2 and remote p1/local p2).

**Counterfactual:** with only the `serve_tarball` guard hunk reverted (test + helper kept), the test fails with 8 `HTTP 404 Not Found` lines — every upstream version, scoped and unscoped, in both orders — while the fork build still serves. Restored the fix and re-ran: passes.

Not run: an end-to-end `npm install`; the router-level test exercises the same two HTTP legs npm uses (packument GET, then tarball GET of the advertised URL). `redact_url_for_diagnostics` (`proxy_service.rs`) is not touched.


irizzant:
FWIW on the red `Build Backend Image` job: the failure happens 12 s in with `ERROR: Error response from daemon: received unexpected HTTP status: 500 Internal Server Error` while pulling an image — no compile/test output at all, and every code gate on this PR is green (Check Rust, Backend Unit Tests, Integration Tests, Coverage). Looks like a transient registry/daemon error; a re-run should clear it.

Issue #3178, opened 2026-08-07
Virtual repo download path serves private member bytes to any caller, including anonymous


brandonrc:
Found while reviewing PR #3173. That PR fixes the **listing** side of #3163; the **byte** side is untouched and is the more serious half.

## The gap

`proxy_helpers::resolve_virtual_download` takes no `auth` parameter:

```rust
pub async fn resolve_virtual_download<F, Fut>(
    db: &PgPool,
    proxy_service: Option<&ProxyService>,
    virtual_repo_id: Uuid,
    path: &str,
    local_fetch: F,
) -> Result<StreamingFetchResult, Response>
```

It structurally cannot filter members by caller. `GET /{key}/artifacts/*path` and `GET /{key}/download/*path` both route through it.

## Demonstrated against PR #3173's head

Listing correctly filtered, bytes served anyway:

```
listing  200  {"items":[],"pagination":{...,"total":0}}
download 200  TOP-SECRET-PAYLOAD
```

And with **no credentials at all**, against a public virtual with a private member:

```
anon listing  200  (private member correctly absent)
anon download 200  ANON-VISIBLE-SECRET
```

An unauthenticated `curl` reads a private repository's contents by asking a public virtual repo for the path.

## Scope

A sweep of `fetch_virtual_members` call sites found **51** that serve caller-visible data with no per-member ACL, of which **11 serve raw artifact bytes** — `proxy_helpers.rs` (×2), `npm.rs`, `oci_v2.rs` (×2), `nuget.rs` (×2), `swift.rs`, `goproxy.rs`, `debian.rs`, `helm.rs`. npm and OCI were spot-verified directly.

## Why this needs its own issue rather than widening #3173

#3173 is a net reduction in disclosure and should land. But it would be wrong to close #3163 on it: the changelog would say virtual member contents are protected while an anonymous request disproves it. The listing fix and the byte fix are different changes in different files, and the byte fix touches eleven format handlers.

## Suggested fix

Thread the caller through `resolve_virtual_download` and filter members with the same `filter_visible_repo_ids` predicate #3173 introduces, so listing and resolution agree on what the caller may see.

[...]

PR #3745, opened 2026-09-07
fix(rpm): resolve a Virtual repository's catch-all paths through its members

### LINKED ISSUE #3573: Virtual RPM repository over Remote members serves an empty index, dnf installs nothing

brandonrc:
Closes #3573

## Summary

A Virtual RPM repository over Remote members could not serve any real yum layout. `dnf` fetches `<prefix>/repodata/repomd.xml` and `<prefix>/Packages/*.rpm`, neither of which matches the explicit `/repodata/` or `/packages/` routes, so both landed on the `upstream_proxy` catch-all — which served Remote repositories only and 404'd every Virtual. The root `/repodata/` a Virtual *does* answer is generated from locally stored artifacts (never proxy-cached content, #1278), so the client received a valid empty index instead of an error and installed nothing.

The catch-all now resolves a Virtual through its members in priority order via the shared `proxy_helpers::try_remote_or_virtual_download` — the same helper and options `download_package` already uses on `/packages/` — so the first member that serves the path wins, hosted members match by filename suffix, Remote members proxy the path through their upstream, the walk is caller-authorized (#3178/#3323), and the serve is recorded (#3446). The Remote and hosted arms of the catch-all are unchanged; the existing root `/repodata/` aggregation over hosted members is untouched.

## Verification

All cargo runs went through the compile lock with `SQLX_OFFLINE=true`, against a fresh `postgres:16-alpine` container migrated with `sqlx migrate run`, and `DATABASE_URL` + `AK_TESTS_REQUIRE_DB=1` + `JWT_SECRET` set so DB-backed tests could not silently no-op. Branch merged with `origin/main` (90b99c60) before the final run.

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(3573) or test(rpm) or test(proxy_download_recording)' --no-tests=fail` — **188 tests run: 188 passed**. This includes the #3446 structural gate (`proxy_download_recording_tests`), which now sees `try_remote_or_virtual_download(` in `upstream_proxy`; the existing `UNRECORDED-PROXY-SERVE: #3446 - deferred` marker on the Remote arm is left in place, so the deferred-format count is unchanged.
- **Counterfactual**: with the Virtual branch removed from `upstream_proxy` (test kept), `test_rpm_virtual_upstream_proxy_walks_remote_members_3573` fails: `el9/x86_64/repodata/repomd.xml`, the shared package and the second-member-only package all return `404 Not Found` where 200 with the member's bytes is expected. With the fix restored it passes.
- `cargo fmt -- --check` — clean.

Not run: the full backend suite, the frontend, and a real `dnf` end-to-end against a live server (the router test drives the same axum router the app mounts, with two wiremock upstreams).

## Regression test

`api::handlers::rpm::tests::test_rpm_virtual_upstream_proxy_walks_remote_members_3573` — a Virtual RPM repo over two public Remote members (priority 0 and 1) whose wiremock upstreams carry a yum layout under `el9/x86_64/`. Asserts: `repodata/repomd.xml` under the prefix is served with the first member's bytes; a package both members carry comes from the priority-0 member; a package only the priority-1 member carries falls through to it; a package neither carries is a 404.

## Notes for reviewers

- The Virtual arm passes the same `DownloadResponseOpts` `download_package` uses (`PathSuffix(filename)`, `application/x-rpm` default, `Content-Disposition` set, no shadowing suppression), so a hosted member that holds the filename shadows a lower-priority Remote member, as on `/packages/`. A consequence worth knowing: a Virtual whose hosted member holds `foo.rpm` now serves `GET /<virtual>/<any prefix>/foo.rpm`, where before it 404'd — identical to what `/packages/<prefix>/foo.rpm` already does on that Virtual. Direct hosted repositories still 404 on the catch-all.
- `upstream_proxy` gains the `auth` and `DownloadContext` extractors the helper needs; the Remote and `@N` publication arms do not use them and are byte-for-byte unchanged.


brandonrc:
## Audit — Virtual member walk, authorization and download recording

**Verdict: merge on green.** The five-line Virtual arm reuses the caller-authorized, first-match walk that `/packages/` already uses. Measured against a live DB: an anonymous caller gets the collapsed 404 ("no accessible members") from a private Remote member and that member's upstream receives 0 requests, while a public member at lower priority serves instead; the `<any prefix>/foo.rpm` widening matches `/packages/` (suffix lookup, `LIMIT 1`), `..` is rejected by `validate_cache_path`, and because RPM filenames embed the architecture and `dnf` follows `primary.xml` hrefs there is no wrong-architecture delivery. The #3446 gate's deferred count is unchanged and nothing double-records. `check`, 189/189 (the PR's 188 plus the reviewer's authorization test), counterfactual (Virtual block removed → both fail), `clippy -D warnings`, `fmt` clean.

Landed as docs on this branch: the handler comment said the helper "records the serve" — it records hosted-member winners only; a Remote-member serve through a Virtual stays unrecorded (#1278), same as `/packages/`. And the CHANGELOG now states the first-member-wins limit for a prefixed `repodata/repomd.xml`: a Virtual over two Remotes with different upstream repositories exposes only the higher-priority member's package list to `dnf`, though lower-priority-only packages are still served when requested directly.

Noted for log readers: a 5xx from a member is a miss, not terminal, so an all-members outage ends as 404, the helper's existing contract for every caller.

PR #3746, opened 2026-09-07
fix(lifecycle): prune the sole OCI tag of a lifecycle-expired image so GC can reclaim it

### LINKED ISSUE #3732: Lifecycle policy leaves the last OCI tag, preventing Storage GC and Blob GC from reclaiming deleted Docker image data
### Description

When a lifecycle policy soft-deletes a Docker/OCI artifact, the corresponding last `oci_tags` entry may remain in the database due to the `Last-protecting-tag` guard.

As a result, the manifest is still considered reachable even though the associated artifact has `is_deleted = true`.

This prevents both manifest and blob cleanup:

* Storage GC does not remove the manifest from object storage because it is still referenced by `oci_tags`.
* `manifest_blob_refs` are not pruned because the manifest is still referenced by `oci_tags`.
* Blob GC therefore considers all blobs referenced by that manifest to still be in use.
* `oci_blobs.pending_delete_at` is never set.
* The image data remains in S3 indefinitely even though the lifecycle policy has deleted the artifact.

### Environment Information

- Deployment type: Docker
- Architecture: x86_64
- Database: PostgreSQL 18
- Reverse Proxy: Caddy

### Backend Version

1.8.2

### Web UI Version

1.8.0

### Package Format

Docker / OCI

### Expected Behavior

When a lifecycle policy expires a Docker/OCI artifact and no live artifact/tag should retain that image, the complete OCI object graph should eventually become eligible for garbage collection.

For example:

```text
Lifecycle expiration
    |
    v
artifact.is_deleted = true
    |
    v
remove obsolete oci_tag
    |
    v
manifest becomes unreachable
    |
    +--> Storage GC removes manifest
    |
    v
prune manifest_blob_refs
    |
    v
blobs become orphaned
    |
    v
Blob GC mark
pending_delete_at = NOW()
    |
    v
grace period
    |
    v
Blob GC sweep
    |
    v
remove blobs from object storage
```

The `last-protecting-tag` guard is useful for preventing accidental deletion of live OCI manifests, but it should not cause lifecycle-expired artifacts to retain their entire manifest/blob graph indefinitely.

### Observed Behavior

The artifact is hidden/soft-deleted by lifecycle policy, but its OCI tag, manifest references, blobs, and physical S3 objects remain.

This means lifecycle policies can appear to successfully delete Docker artifacts while not reclaiming the corresponding storage.

### Steps to Reproduce

1. Create a Docker repository backed by S3 storage.

2. Push a Docker image to the repository, for example:

```bash
docker push <registry>/<repository>/temp:1
```

The push creates an artifact similar to:

```text
name:        temp:1
version:     1
storage_key: oci-manifests/sha256:ce2e051a...
is_deleted:  false
```

and an OCI tag:

```text
tag:             1
manifest_digest: sha256:ce2e051a...
```

The manifest references its config and layers through `manifest_blob_refs`:

```text
sha256:ce2e051a... -> sha256:4c4f0244...  config
sha256:ce2e051a... -> sha256:738128fa...  layer
sha256:ce2e051a... -> sha256:7d8f9b30...  layer
```

Corresponding entries also exist in `oci_blobs`.

3. Configure and execute a lifecycle policy that makes this artifact eligible for deletion, for example a **No Downloads** policy.

For testing, `artifacts.created_at` can be moved into the past so that the artifact satisfies the lifecycle policy.

4. After lifecycle execution, inspect the artifact:

```sql
SELECT
    name,
    version,
    storage_key,
    is_deleted
FROM artifacts
WHERE repository_id = '<repository-id>';
```

Result:

```text
name   | version | storage_key                         | is_deleted
-------+---------+-------------------------------------+-----------
temp:1 | 1       | oci-manifests/sha256:ce2e051a...   | true
```

The lifecycle policy has correctly soft-deleted the artifact.

5. Inspect `oci_tags`:

```sql
SELECT
    tag,
    manifest_digest
FROM oci_tags
WHERE repository_id = '<repository-id>';
```

Result:

```text
tag | manifest_digest
----+------------------------
1   | sha256:ce2e051a...
```

The tag remains even though its corresponding artifact is soft-deleted.

6. Inspect `manifest_blob_refs`:

```sql
SELECT *
FROM manifest_blob_refs
WHERE repository_id = '<repository-id>';
```

The references also remain:

```text
manifest_digest    | blob_digest         | kind
-------------------+---------------------+-------
sha256:ce2e051a... | sha256:4c4f0244...  | config
sha256:ce2e051a... | sha256:738128fa...  | layer
sha256:ce2e051a... | sha256:7d8f9b30...  | layer
```

7. Inspect the blobs:

```sql
SELECT
    ob.digest,
    ob.storage_key,
    ob.created_at,
    ob.pending_delete_at,
    COUNT(mbr.blob_digest) AS ref_count
FROM oci_blobs ob
LEFT JOIN manifest_blob_refs mbr
    ON mbr.repository_id = ob.repository_id
   AND mbr.blob_digest = ob.digest
WHERE ob.repository_id = '<repository-id>'
GROUP BY
    ob.digest,
    ob.storage_key,
    ob.created_at,
    ob.pending_delete_at;
```

Even when the blobs are older than the Blob GC minimum age, the result is similar to:

```text
digest              | pending_delete_at | ref_count
--------------------+-------------------+----------
sha256:4c4f0244...  | NULL              | 1
sha256:738128fa...  | NULL              | 1
sha256:7d8f9b30...  | NULL              | 1
```

Blob GC therefore never marks these blobs for deletion.

### Logs

```shell

```

### Root cause

Lifecycle execution first soft-deletes the artifact:

```text
artifacts.is_deleted = true
```

The subsequent OCI tag cleanup contains a `last-protecting-tag` guard.

The guard prevents lifecycle cleanup from removing the final `oci_tags` entry referencing a manifest.

For a Docker image with only one tag:

```text
1 -> sha256:ce2e051a...
```

the lifecycle policy therefore produces:

```text
artifacts:
temp:1 -> is_deleted = true

oci_tags:
1 -> sha256:ce2e051a...   <-- remains
```

This remaining tag has two consequences.

#### 1. Storage GC cannot reclaim the manifest

Storage GC considers the manifest protected because an `oci_tags` row still references its digest.

Therefore:

```text
oci-manifests/sha256:ce2e051a...
```

is not removed from S3.

#### 2. Blob GC cannot reclaim the image blobs

`manifest_blob_refs` cleanup only removes references for manifests that are no longer reachable through `oci_tags` (or other live OCI manifest relationships).

Because the tag remains:

```text
oci_tags
    |
    v
manifest sha256:ce2e051a...
    |
    v
manifest_blob_refs
    |
    +--> config blob
    +--> layer blob
    +--> layer blob
```

the `manifest_blob_refs` rows remain.

Blob GC then sees every blob as referenced:

```text
ref_count > 0
```

and never sets:

```text
oci_blobs.pending_delete_at
```

The blobs therefore cannot progress through the Blob GC mark/sweep process.

### Resulting cleanup deadlock

The current lifecycle/GC interaction effectively results in:

```text
Lifecycle policy
    |
    v
artifact.is_deleted = true
    |
    v
Last-protecting-tag guard
    |
    +--> sole oci_tag remains
              |
              +--> Storage GC considers manifest protected
              |
              +--> manifest_blob_refs remain
                         |
                         +--> Blob GC considers blobs referenced
```

Consequently, neither the manifest nor its blobs are reclaimed from S3.

### Suggested investigation

The interaction between the lifecycle OCI tag cascade and the last-protecting-tag guard should distinguish between:

1. a manifest that is still legitimately referenced by another live artifact/tag; and
2. a manifest whose only remaining tag belongs to an artifact that was itself expired by the lifecycle policy.

In the second case, the final OCI tag should be allowed to become unreachable so that Storage GC and Blob GC can eventually reclaim the complete image graph.

### Pre-submission Checklist

brandonrc:
Closes #3732

## Summary

`CASCADE_OCI_TAGS_SQL` (`backend/src/services/lifecycle_service.rs`) pruned an `oci_tags` row after a lifecycle policy soft-deleted its manifest artifact **only when a surviving sibling tag** held the same digest (#1987, for #1682). For a single-tag image that guard never fired: the tag row outlived its artifact, storage GC treated it as a live reference and skipped the manifest, blob GC never pruned its `manifest_blob_refs`, and the blobs were never marked `pending_delete_at`. The image stayed in storage until a restart, when the OCI reindex (`oci_migration_reindex.rs`) dropped exactly that tag as an orphan.

The guard is widened with a second prong: a tag is also pruned when **no live `artifacts` row backs its digest in that repository**, where "backs" is judged by `'sha256:' || checksum_sha256 = manifest_digest` (the reindex's key — a pre-#2457 migrated manifest keeps its live row at a generic CAS `storage_key`) **or** by the canonical `oci-manifests/<digest>` key, **and** the tag is not newer than the tombstone (`ot.updated_at <= a.updated_at`; every lifecycle soft-delete now stamps `artifacts.updated_at`). The latter closes the same-digest re-push race: `handle_put_manifest` commits its `oci_tags` upsert before it revives the `artifacts` row, and the cascade runs in a later READ COMMITTED transaction, so a tag written after the soft-delete is a push in flight and is left alone. Siblings the policy did not match still protect their digest; index children stay governed by GC's index-child clause. CHANGELOG carries an operator note that stranded images are reclaimed on the next lifecycle run (or restart).

Second commit addresses the review findings F1-F4 of the first revision (CAS-keyed live rows; the re-push race; wording; missing tests).

## Tests

DB-backed, in `lifecycle_service.rs` (run by `--lib`):

- `test_lifecycle_expired_single_tag_image_is_reclaimable_3732` — single-tag image expired by `max_age_days`: tag row gone, `select_orphans` / `run_gc_for_repository(dry_run)` list the manifest, `run_blob_gc_mark` prunes the refs and marks both blobs. Fails on main (tag retained, 0 candidates).
- `test_lifecycle_keeps_surviving_sibling_tag_of_expired_manifest_3732` — the #1987 case: only the expired tag is pruned, the sibling survives, no GC candidate.
- `test_lifecycle_leaves_live_index_children_protected_3732` — a live tagged index with two children next to an expiring image: index tag kept, only the expired manifest is a candidate.
- `test_cascade_keeps_tag_backed_by_live_cas_keyed_migrated_row_3732` — F1: live pre-#2457 row at `sha256/xx/<hex>` with NULL version; the re-pushed row expires; the tag survives and nothing is a candidate.
- `test_cascade_leaves_tag_upserted_after_soft_delete_3732` — F2, deterministic ordering: a tag stamped after the tombstone survives; an older one is pruned in the same sweep.
- `test_cascade_prunes_expired_tag_of_index_child_gc_keeps_child_3732` — F4(c): expired direct tag on a live index's child is pruned; the child stays out of the candidate set.
- `test_cascade_keeps_tag_backed_by_live_digest_pinned_row_3732` — F4(a): a live `v2/<name>/manifests/sha256:...` row keeps the expired human tag.
- `test_cascade_sql_prunes_tag_when_no_live_artifact_backs_digest_3732` — pure SQL-shape pin (both key shapes and the `updated_at` clause).

The three `#[ignore]` integration tests from #1987 that asserted retention of the sole tag (`backend/tests/lifecycle_policy_tests.rs`) now assert the #3732 behaviour; the surviving-sibling one is unchanged.

**Two-session race probe** (reviewer's F2 scenarios, psql against the new SQL, `$S/fix/3732-race`): sequential soft-delete → push tag commit → cascade → push artifact commit ends `is_deleted=false, tags left=1`; push transaction holding the tag row lock while the cascade `DELETE` runs (blocked 2.6 s) and committing the live row before releasing also ends `is_deleted=false, tags left=1`; the plain #3732 fixture (no re-push) ends `is_deleted=true, tags left=0`.

## Verified

Against a throwaway PostgreSQL 16 (`ak-fix-3732`, 211 migrations), after merging `origin/main`:

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(3732) or test(1987) or test(1682) or test(lifecycle) or test(storage_gc) or test(blob_gc)'` with `AK_TESTS_REQUIRE_DB=1` — 349 passed
- `cargo nextest run --test lifecycle_policy_tests --run-ignored ignored-only` — 23 passed
- `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` — clean; `cargo fmt --check` — clean

Not run: the full backend suite and the e2e stacks.


brandonrc:
## Audit — data-safety of the lifecycle cascade and GC interaction, two rounds

**Round 1 was NEEDS-WORK:** the new prong could prune a tag whose manifest was still live in two reproduced ways. Neither lost bytes (a live `artifacts` row keeps the object out of GC's candidate set), but both left the image `MANIFEST_UNKNOWN` by tag until the next restart's reindex:

- The prong keyed on `storage_key = 'oci-manifests/<digest>'`, but pre-#2457 migrated manifests keep a live row at a generic content-addressed key, which is why the reindex keys on `'sha256:' || checksum_sha256`. Replayed on a review DB: a migrated image re-pushed natively then expired by a policy lost its tag.
- A same-digest re-push races the cascade: `handle_put_manifest` commits the tag in its own transaction before the artifact upsert, and the cascade runs in a separate READ COMMITTED transaction after the soft-delete commits. Two psql sessions reproduced `artifact live=true, tags left=0`.

**Round 2 verdict: MERGE** (head `69bfeabd`). Prong 2 now keys on the checksum or the canonical storage key, all five lifecycle soft-delete UPDATEs stamp `updated_at`, and prong 2 requires the tag row to be no newer than the tombstone. Both race orderings (deterministic tag→cascade→artifact, and the concurrent lock-holding variant with the cascade blocked 2.5 s) now delete nothing, while the control with a tag ten days older than the tombstone still prunes. The guard cannot wrongly retain: both columns are `NOT NULL DEFAULT now()`, both production tag upserts stamp `updated_at`, and pre-upgrade stranded tombstones whose old soft-delete never bumped it still prune, so the operator note holds. The #1987 sibling protection and the index-child protection are intact. 349 `--lib` tests (345 plus four new covering both findings) and 15 ignored cascade tests green on live Postgres; `check`, `clippy -D warnings`, `fmt` clean; the diff is exactly the three intended files on top of main.

Issue #3712, opened 2026-09-03
OCI Remote: expired mutable tag is served from local storage without upstream revalidation


verdel:
### Description

For an OCI Remote repository, an expired cached manifest referenced by a mutable tag is still served from local storage without being revalidated against the upstream registry.

The UI correctly shows that the cache has expired:

```text
Cache expires
expired 8 seconds ago, will re-fetch on next download
```

However, the next `docker pull` does not trigger an upstream request.

For a previously cached tag, the `HEAD` manifest handler resolves the tag through `oci_tags`, reads the corresponding `oci-manifests/<digest>` object from local storage, and returns it before reaching the Remote proxy path where cache TTL/revalidation is handled.

As a result, a mutable tag can continue resolving to an old digest even after its cache TTL has expired and the tag already points to a new digest upstream.

### Environment Information

- Deployment type: Docker
- Operating System: Ubuntu 24.04
- Architecture: x86_64
- Database: PostgreSQL 18
- Reverse Proxy: Caddy

### Backend Version

1.8.2

### Web UI Version

1.8.0

### Package Format

Docker / OCI

### Expected Behavior

After the cache TTL for a mutable tag expires, the next pull should revalidate the tag against the upstream registry.

For example:

```text
Initial upstream state:

repo/temp:1 -> sha256:e4f20b30...
```

After pushing a new image with the same tag:

```text
Current upstream state:

repo/temp:1 -> sha256:ce2e051a...
```

The next:

```text
HEAD /v2/dockerhub-proxy/repo/temp/manifests/1
```

should revalidate the expired tag and return:

```text
Docker-Content-Digest: sha256:ce2e051a...
```

### Observed Behavior

Even after the cache is reported as expired, Artifact Keeper returns the old locally stored manifest:

```text
HEAD /manifests/1
        |
        v
oci_tags
1 -> sha256:e4f20b30...
        |
        v
oci-manifests/sha256:e4f20b30...
        |
        v
HEAD 200
Docker-Content-Digest: sha256:e4f20b30...
```

No request to the upstream registry is made.

Docker then correctly requests the digest returned by HEAD:

```text
GET /manifests/sha256:e4f20b30...
```

which is also served from local storage.

The current upstream digest `sha256:ce2e051a...` is therefore never discovered.

### Steps to Reproduce

1. Create a direct Docker Remote repository pointing to Docker Hub:

   ```text
   key: dockerhub-proxy
   type: remote
   format: docker
[...]

Issue #3290, opened 2026-08-11
PyPI HTML and PEP 691 JSON proxy caches can become inconsistent


axellpadilla:
## Description

Artifact Keeper can serve different package versions depending on whether a client requests the PyPI HTML Simple API or the PEP 691 JSON Simple API.

The HTML response was refreshed and contained newly published releases, while the JSON response remained an older cached snapshot. This caused `uv`, which prefers the JSON API, to report a package version as unavailable even though the same version was present in the HTML index and could be downloaded directly.

## Environment Information

- Deployment type: Kubernetes
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL
- Reverse Proxy: None required to reproduce
- Network setup: Direct outbound access to PyPI
- Backend Version: 1.7.1
- Web UI Version: N/A
- Package Format: PyPI

## Expected Behavior

The HTML and PEP 691 JSON representations of the same mutable PyPI project index should remain consistent after upstream releases are published.

A client using `uv --index-strategy first-index` should see the same package versions available through both representations.

## Observed Behavior

The HTML index contained newly published files, but the JSON index did not.

For a PyPI project, the JSON response initially contained 142 files while PyPI contained 146 files. The missing files included:

```text
<package>-1.11.1-py3-none-any.whl
<package>-1.11.1.tar.gz
<package>-1.12.0rc3-py3-none-any.whl
<package>-1.12.0rc3.tar.gz
```

`uv` then reported:

```text
there is no version of <package>==1.12.0rc3
```

The package was available in the HTML index and direct artifact downloads succeeded.

The JSON representation was cached separately under a key equivalent to:

```text
simple/<package>/index.v1+json
```

Invalidating that exact JSON cache entry caused Artifact Keeper to refetch the upstream response. The JSON response then contained all 146 files, and `uv` resolved successfully using `first-index`.

## Steps to Reproduce

1. Configure a remote PyPI repository pointing to `https://pypi.org`.
2. Request a project's HTML Simple API response.
3. Allow the project to receive new releases upstream.
4. Request the same project with:

   ```http
   Accept: application/vnd.pypi.simple.v1+json
   ```

5. Compare the HTML and JSON file lists with the upstream PyPI index.
6. Resolve a newly published version using `uv` with `--index-strategy first-index`.

The issue is intermittent because it depends on the timing of the cached JSON response and upstream releases.

## Additional Context

[...]

PR #3747, opened 2026-09-07
fix(proxy): serve stale instead of refilling when the revalidation probe is throttled or refused

### LINKED ISSUE #3571: Proxy revalidation treats a throttled or refused upstream as "content changed"

brandonrc:
Closes #3571

## Summary

When a cached mutable proxy entry goes past its TTL, `revalidate_verdict` probes upstream with a conditional `HEAD` through `UpstreamClient::check_etag_changed`. That function mapped every status it did not recognise to `Ok(true)` ("changed"), so a `429` from a rate limiter, a `403` from an upstream refusing us, or any `5xx` sent the entry into the refill path and a full `GET` against the upstream that had just refused a single cheap `HEAD` — a re-download when the throttled upstream let the `GET` through, and otherwise the stale body only after a second refused request.

Those statuses (plus `408`) now surface from the probe as the error class a failed `GET` gets (via the existing `validate_upstream_status`), so `revalidate_verdict`'s existing `Err` arm routes them into the RFC 5861 stale-if-error handling: stale body served within `STALE_IF_ERROR_GRACE_SECS` of the original expiry, no `GET`, sidecar untouched; past the grace window the entry refills as before.

Deliberately unchanged, per the issue:
- `401` keeps meaning "re-fetch with a token" (OCI bearer exchange), pinned by a new test.
- `404`/`410` keep falling into the "changed" arm (refill -> negative cache on 404).
- The expired-tag revalidation on Remote OCI repositories (#3712) calls the same proxy path and needs no change; it now gets the stale manifest without a wasted manifest `GET`.
- `Retry-After` is not consulted: each stale read within the grace window still costs one `HEAD`. Honouring it would mean carrying a duration through `AppError` into the sidecar, which is more than a verdict-mapping change.

## Verification

All cargo invocations through the compile lock, with `CARGO_TARGET_DIR` isolated, `SQLX_OFFLINE=true`, and a local `postgres:16-alpine` (`ak-fix-3571`, migrations applied) with `AK_TESTS_REQUIRE_DB=1`, so the DB-backed tests could not silently no-op.

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(3571) or test(revalidat) or test(etag) or test(stale)' --no-tests=fail` — **145 passed, 0 failed** (includes the four new `test_3571_*` tests and the existing #1611 revalidation matrix, #3147 stale-if-error tests and #3290 ETag tests).
- **Counterfactual on main's mapping**: with the new `check_etag_changed` arm removed and the tests kept, `-E 'test(3571)'` — `test_3571_revalidate_429_serves_stale_without_refill`, `..._503_...` and `..._403_...` **fail** with `429 on the probe must serve the copy we already hold, not refill` (the body served was the wiremock `GET` refill, i.e. a re-download), and `test_3571_revalidate_401_still_refills` **passes** on both. Fix restored afterwards (`git checkout`), verified present.
- `cargo fmt -- --check` — clean.

New tests (wiremock, `HEAD` `expect(1)` / `GET` `expect(0)`, sidecar bytes compared before and after):
- `test_3571_revalidate_429_serves_stale_without_refill`
- `test_3571_revalidate_503_serves_stale_without_refill`
- `test_3571_revalidate_403_serves_stale_without_refill`
- `test_3571_revalidate_401_still_refills` (negative pin)

Note: the branch was rebased onto `origin/main` after the test run; the only commit main gained in between (#3742) touches `scripts/ci/check-toolchain-pin.sh` and the CHANGELOG, not backend source, so the results above were not re-run.


brandonrc:
## Audit — proxy revalidation semantics and cache-poisoning safety

**Verdict: merge on green.** The mapping is correct and bounded. The reviewer built the full status matrix before and after: 200 same-ETag / 304 unchanged, 200 new-ETag changed, 404 negative-cached, 401 re-fetches with a token (the OCI bearer exchange), and 405/400/416/3xx/no-ETag all still refill via `GET` (the right fallback for `HEAD`-less upstreams), measured identical before and after; only 403/408/429/5xx move, from "changed" to stale-if-error. No new poisoning window: the grace is anchored on `expires_at` and never re-stamped on an error (three consecutive throttled reads = 3 `HEAD` / 0 `GET` / identical sidecar). No supported public upstream answers 403 to `HEAD` and 200 to `GET` (live-probed Maven Central, PyPI, npm, crates.io; Docker Hub and GHCR answer 401), and the cold path is untouched since the probe is only called from `revalidate_verdict`. 147 tests on the selector against live Postgres; counterfactual fails the three positive tests with the 401 pin passing; `clippy -D warnings`, `fmt` clean; merge with main trivial.

Landed as docs: the CHANGELOG claimed the one-hour grace "bounds how long a throttled upstream can be papered over"; past the grace a refill is attempted, but if the `GET` also fails the refill path serves stale with no grace check, exactly as on main. The bullet now says so. Filed **#3749** for the pre-existing 410 case (probe 410 → refill 410 → `BadGateway` → stale served indefinitely instead of negative-cached like 404).

PR #3759, opened 2026-09-08
fix(security): answer an anonymous /v2 read of a missing repository with the private-repository challenge

### LINKED ISSUE #3730: OCI /v2: an anonymous caller gets 401 for a private repository but 404 NAME_UNKNOWN for a missing key — the #1808 existence oracle on the docker pull path
## Summary

Noticed while fixing #3716. On the native-format routes, #1808 made the no-repository branch of `repo_visibility_middleware` mirror the private-repository `401` + `WWW-Authenticate` challenge byte-for-byte for anonymous callers, so an unauthenticated probe cannot tell a private repository from a nonexistent one. The OCI distribution surface does not have that property: an anonymous `GET /v2/<repo>/<image>/manifests/<tag>` answers `401` with the bearer challenge for a private repository but `404 NAME_UNKNOWN` for a key that names no repository (`resolve_repo_inner`'s unknown-key answer).

So `docker pull` without credentials remains an existence oracle for private repositories, one hop before the gates that #3716 and #3717 unify for authenticated callers.

## Ask

For an anonymous caller, answer a missing repository on `/v2` with the same `401` + `WWW-Authenticate` challenge a private repository gives (the #1808 shape), so the two are indistinguishable before authentication. Public repositories keep their anonymous 200/404 as today, since their existence is not secret. Regression test: anonymous, private repo vs random missing key → identical status, headers and body; control: public repo → 200 anonymously; authenticated non-member on both → the #3716 `404 NAME_UNKNOWN` pair.

Refs: #1808, #3524, #3716, #3717, GHSA-fv45-mwhh-q23r.

brandonrc:
## Summary

On `/v2`, an anonymous `GET /v2/<repo>/<image>/manifests/<tag>` answered `401` + the `WWW-Authenticate` bearer challenge for a **private** repository but `404 NAME_UNKNOWN` for a key naming **no** repository (`resolve_repo_inner`'s unknown-key branch). Guessing a key and reading the status told an unauthenticated caller whether it names a private repository — the #1808 existence oracle, still open on the `docker pull` path one hop before the gates #3716/#3717 unify for authenticated callers.

#1808 closed this on the native format routes by making `repo_visibility_middleware`'s no-repository branch mirror the private-repository 401 byte-for-byte. `/v2` is mounted outside that middleware. This is its OCI equivalent.

## Change

- `resolve_repo_inner` now returns `Ok(None)` for "this key names no repository" instead of baking in `oci_name_unknown`; `resolve_repo` and `resolve_repo_for_write` produce that 404 themselves, unchanged.
- New `resolve_repo_for_anonymous_capable_read` replaces the `resolve_repo` + `if is_anon && !repo.is_public` pair at the five anonymous-capable read sites (`handle_get_manifest`, `handle_head_manifest`, `handle_get_blob`, `handle_head_blob`, `authorize_oci_repo_read` — which covers `tags/list` and referrers). For an anonymous caller it asks one question — is this key a **public** repository? — so a private repository and a missing key leave by the same branch with the byte-identical challenge.
- Public repositories keep their anonymous 200/404 (their existence is not secret). Authenticated callers are unchanged: the #3716/#3728 existence-hiding `404 NAME_UNKNOWN` pair. A transient resolution failure keeps its retryable 503 (#2083) rather than being folded into a 401. `/v2/` and `/v2/_catalog` are untouched.

## Verification

- `test_3730_anonymous_probe_of_missing_key_matches_private_repo_challenge` — router-level, live DB. Anonymous private-repo vs random missing key compared as the full `(status, WWW-Authenticate, Content-Type, Docker-Distribution-Api-Version, Content-Length, body)` tuple across `GET`/`HEAD` manifest, `GET` blob and `tags/list`, normalising only the client-supplied key (the challenge's `scope` echoes it, and the test asserts it does). Controls: member pulls the private repo (200); the same anonymous token gets 200 for a public manifest and 404 for a missing tag in a public repo.
- `test_3730_authenticated_pair_and_login_flow_unchanged` — authenticated non-member gets the identical `(status, content-type, body)` for the private repo and a missing key; the `docker login` flow (`/v2/` 401 + challenge anonymously → 200 with the credential → `/v2/token` → pull the private manifest with the issued token) still works end to end.
- **Counterfactual on this branch's merge base:** reverse-applying only the non-test hunks makes the first test fail exactly as reported — `left: (401, [Some("Bearer realm=…,scope=\"repository:<key>/app:pull\""), …], UNAUTHORIZED)` vs `right: (404, [None, …], NAME_UNKNOWN)`. The second test passes without the fix, as intended (it pins what must not change).

Commands run (compile lock, `AK_TESTS_REQUIRE_DB=1`, live Postgres 16):

- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(3730) or test(1808) or test(3716) or test(oci_read_authz) or test(anonymous)'` → 80 passed
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(oci)'` → 772 passed
- `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` → clean; `cargo fmt` applied

Not run: the `backend/tests/` integration binaries and the e2e suites (kept the build targeted to `--lib`).

Closes #3730


brandonrc:
## Audit — anonymous existence-oracle closure on the OCI surface, and availability

**Verdict: merge on green.** Measured live on the PR head: ten repository-name shapes (uppercase, `library/`-prefixed, nested, empty segment, 300-character, punctuation, percent-encoded, no image part) across six verbs all produce byte-identical answers for a private repository and a missing key — status, `WWW-Authenticate` including the realm, service and scope, `Content-Type`, `Docker-Distribution-Api-Version`, `Content-Length` and body — with the challenge's scope echoing the client's own string in both cases. Public repositories keep anonymous 200/404, including "public and missing"; authenticated callers keep the #3716 triple and #3728's scope-gate 404 on all five sites; the `docker login` handshake and `/v2/`, `/v2/_catalog` are unchanged.

The availability question the refactor raised checks out: `resolve_repo_inner` now returns `Ok(None)` for "no such key" and all three callers handle it, while a database failure still propagates before the anonymous match. With a hogged one-connection pool an anonymous read answers 503 with `Retry-After` and no challenge, for private and missing keys alike, so an outage does not turn into a credential prompt. 80 and 772 test counts reproduced exactly; counterfactual fails precisely as reported; `clippy` clean; merge with main clean.

Landed from the findings: the doc comment and the CHANGELOG now carve out Docker-mirror mode by name — with `AK_DEFAULT_DOCKER_MIRROR_REPO` set and a public mirror, a missing key is re-resolved to the mirror before the visibility decision, so the 401-versus-404 difference survives there; the mode is off by default and untouched by this PR. The verb loop also pins `HEAD` blob and `referrers`, the two sites the reviewer verified by hand.

Filed **#3761** for the one thing left alone: `resolve_repo_inner`'s database-error mapping still builds its message with `e.to_string()` rather than #3667's sanitiser, so a pool timeout echoes driver text to an anonymous caller in an otherwise correct 503.

Issue #2551, opened 2026-07-14
npm filter_meta_response: fail-closed on unrecognized meta shapes (scalar-valued keys, top-level arrays) — LOW/hostile-upstream only


brandonrc:
Residual from #2542/#2548 re-red-team (LOW/theoretical, NOT reproducible against npmjs/verdaccio/Nexus — all use object-map shapes that ARE filtered). `filter_meta_response` is shape-specific: it scope-checks keys whose value is an object or array, but (a) a map key with a SCALAR value (e.g. `/-/all` entry `"evilpkg":"1.0.0"`) is kept, leaking the key name, and (b) a top-level JSON ARRAY body is served verbatim (`as_object_mut()?` returns None → caller serves upstream bytes unfiltered). Only reachable via a hostile/non-standard upstream. Defense-in-depth: drop out-of-scope scalar-valued keys too, and fail-closed (empty list) on unrecognized top-level shapes rather than passing through. Origin: PR #2548.

Issue #2521, opened 2026-07-14
PF-004 [P1]: Whole-repository package metadata is regenerated synchronously on reads


brandonrc:
# PF-004: Whole-repository package metadata is regenerated synchronously on reads

Priority: P1  
Affected area: package-manager protocol metadata

## Finding

Several package formats implement repository-wide index documents by fetching all relevant catalog rows and constructing the full response inside each request. This is protocol-correct for small repositories but turns routine package-manager refreshes into repeated O(N) database, CPU, allocation, compression, and serialization work.

The response itself may inherently be large; the performance requirement is not “make it O(1) bytes.” It is to build a version once when repository state changes, serve immutable cached bytes many times, and avoid coupling all client refreshes to a live full-catalog query.

## Representative evidence

- PyPI `/simple/` selects every distinct package name, normalizes and sorts them in a `BTreeSet`, and serializes the full HTML/JSON response. Virtual repositories fan out member queries, and remote recovery may also list the object cache: [pypi.rs](../backend/src/api/handlers/pypi.rs#L202), [pypi.rs](../backend/src/api/handlers/pypi.rs#L505).
- Helm `index.yaml` fetches every live artifact plus JSON metadata, builds all chart structures, serializes the complete YAML string, and returns it as one body: [helm.rs](../backend/src/api/handlers/helm.rs#L54), [helm.rs](../backend/src/api/handlers/helm.rs#L132).
- Composer's root index explicitly fetches every non-deleted artifact/version and builds an in-memory package map: [composer.rs](../backend/src/api/handlers/composer.rs#L191).
- RPM loads every artifact and metadata row. `repomd.xml` then generates primary, filelists, other, and updateinfo XML, compresses them, and hashes them in the request even though it returns only the manifest: [rpm.rs](../backend/src/api/handlers/rpm.rs#L233), [rpm.rs](../backend/src/api/handlers/rpm.rs#L295), [rpm.rs](../backend/src/api/handlers/rpm.rs#L391).
- RPM `primary.xml.gz` separately repeats the full load/generate/compress flow: [rpm.rs](../backend/src/api/handlers/rpm.rs#L540).

## Impact at scale

- At one million artifacts, a single metadata request can allocate the row set, domain structures, uncompressed output, compressed output, and response body at the same time.
- Package managers refresh metadata frequently and many CI jobs start together, multiplying identical work.
- Virtual repositories repeat member reads and merging.
[...]

PR #3765, opened 2026-09-08
fix(npm): negative-cache proxied attestation 404s

### LINKED ISSUE #3764: npm attestation lookups are proxied correctly but never cached, so `npm audit signatures` sends every 404 upstream on every repeat
**Repo:** `artifact-keeper` (backend)
**Type:** bug (performance / upstream amplification)
**Version / branch:** verified on `main` @ `2bcc4c15` (1.8.2); present since **v1.2.1**

## Summary

Proxied npm attestation lookups are never cached. `npm audit signatures` — and any `npm install`
that checks provenance — asks the registry one question per resolved package version:

```
GET /npm/{repo}/-/npm/v1/attestations/{package}@{version}
```

Almost no published npm version has a provenance attestation, so the upstream registry answers
`404` to nearly all of them. **Artifact Keeper relays that `404` correctly** — for Remote and
Virtual repositories the whole `/-/` meta namespace is proxied and the upstream status is returned
verbatim, which is the right behaviour. The bug is that the answer is **never memoised**: every
repeat of the same `pkg@version` question is a fresh round trip to the upstream registry.

To be explicit, because it is the obvious first guess: this is **not** a request to implement
attestation serving. That already works. The only defect is the missing cache.

## Why the repeat rate makes this expensive

A CI fleet re-resolves the same dependency graph continuously, so the number of *distinct*
attestation URIs is one to two orders of magnitude below the number of *requests*. A day of
observed proxy traffic from a single mid-size deployment:

| | |
|---|---|
| Attestation requests / day | 231,445 — 17.8% of all requests, 94% of all 404s |
| Status mix | 231,445 × `404`, 22 × `401`. **Zero `200`s** |
| Distinct attestation URIs / day | 3,429 → **67.5 repeats per URI per day** |
| Distinct URIs / hour (one sampled hour) | 2,005 of 17,958 requests → **9.0 repeats per hour** |

Upstream round trips a cache would eliminate, straight from those ratios:

| TTL | Round trips eliminated |
|---|---|
| 1 hour | 1 − 2,005/17,958 = **88.8%** |
| 24 hours | 1 − 3,429/231,445 = **98.5%** |

That is ~228,000 requests/day that never needed to leave the process. Each one also holds a slot
against `GLOBAL_MAX_CONCURRENCY` and pays the repo-resolve and permission queries in front of it,
so the cost is not only egress.

There is a good-citizen argument here too: one mid-size deployment sends a quarter of a million
pointless requests a day at the public registry, and every Artifact Keeper install proxying npm is
doing the same thing at its own scale.

## Reproduction

Point a Remote npm repository at a registry and ask the same attestation question twice.

```bash
BASE=http://localhost:8080
REPO=npm-remote     # Remote repo, upstream https://registry.npmjs.org

# lodash 4.17.21 has no provenance attestation, so upstream answers 404
for i in 1 2 3 4 5; do
  curl -s -o /dev/null -w '%{http_code} %{time_total}s\n' \
    "$BASE/npm/$REPO/-/npm/v1/attestations/lodash@4.17.21"
done
```

Observed: five `404`s, each one an upstream round trip — visible as five outbound requests in the
upstream's own logs, and as five near-identical, non-trivial `time_total` values rather than one
slow request followed by four fast ones.

Expected: the first request goes upstream, the rest are answered from cache.

The same holds for a Virtual repo aggregating a Remote member. Substituting a local mock registry
for the upstream makes the request count directly assertable.

## Where it is in the code

All in `backend/src/api/handlers/npm.rs` (line numbers on `main` @ `2bcc4c15`):

1. **Route** — `npm.rs:91`, `.route("/:repo_key/-/*rest", get(npm_meta_get))`. Attestation paths
   fall into this `/-/` meta catch-all; there is no attestation-specific route.
2. **`npm_meta_get`** (`npm.rs:1832`) — Remote: proxy to `repo.upstream_url`. Virtual: walk
   `authorized_virtual_members` and return the first member that answers, so it is one upstream
   call per request, not one per member. Local/Staging fall through to the `501` stub.
3. **`proxy_npm_meta_get`** (`npm.rs:1782`) — returns the upstream status verbatim, so an upstream
   `404` reaches the client as a `404`. Correct, and worth preserving exactly.
4. **`npm_meta_upstream_bytes`** (`npm.rs:1722`) — **the defect.** An unconditional
   `client.get(&url).send().await` on every call. No cache read, no cache write, no TTL, no
   `Cache-Control` handling. Nothing in the `/-/` path is memoised.

This has been the behaviour since the `/-/` meta route was introduced in #1885 (first released in
**v1.2.1**); before that, attestation requests were not proxied at all.

## Proposed fix

A **negative-result cache scoped to `/-/npm/v1/attestations/*` only**, with a configurable TTL
defaulting to 24 hours. A 24-hour default is safe because npm forbids republishing a version, so
"`pkg@version` has no attestation" is effectively a stable fact; an attestation added after publish
is rare enough that a day of staleness is acceptable, and the knob lets an operator who cares
shorten it. Even a 1-hour TTL removes ~89% of the round trips.

The constraints a correct implementation has to respect — these are the things a naive cache gets
wrong:

- **Key on the answering member repository, not the requested repository.** For a Virtual repo the
  member that answers depends on the caller's grants (`authorized_virtual_members`, #3323). Keying
  on `(member_repo_id, meta_path)` keeps this correct: each caller still walks only their own
  authorized members and so can only reach entries for members they may use. Keying on the virtual
  repo alone would let one tenant be served an upstream answer fetched for a member they cannot
  reach.
- **Cache attestations only. Do not generalise to `/-/*`.** `/-/whoami` and `/-/npm/v1/user` are
  per-principal, `/-/v1/search` is query-dependent, the audit endpoints are request-body dependent.
  Caching the meta namespace wholesale would be a security bug.
- **Respect the scope policy.** `proxy_npm_meta_get` applies `NpmScopePolicy` filtering before
  returning, so a cached body under an active policy would have to be policy-specific. Either
  include the policy identity in the key, or skip the cache when a policy is active.
- **Cache status and body together.** A cached `404` must replay as a `404`, never as an empty
  `200` — npm treats those very differently.
- **Cache negative answers only.** A `200` is a real provenance bundle; leaving it uncached is what
  keeps a newly-published attestation immediately visible. `401`/`403` are credential-dependent and
  `429`/5xx are transient, so caching either would turn a blip into a TTL-long outage.

Scoped that tightly this is a small, reviewable change against a well-isolated function. I have an
implementation along these lines and will open a PR referencing this issue.

knowinglyAnonymous:
## Summary

Closes #3764.

`npm audit signatures` — and any `npm install` that checks provenance — asks the registry one
question per resolved package version:

```
GET /npm/{repo}/-/npm/v1/attestations/{package}@{version}
```

Almost no published npm version carries a provenance attestation, so upstream answers `404` to
nearly all of them. **Relaying that `404` is correct and this PR does not change it** — it is the
registry's own answer, and an Artifact Keeper that invented attestations would be the real bug. The
defect is that the answer was never memoised: `npm_meta_upstream_bytes`, which backs the whole
`/-/` meta namespace for Remote and Virtual repositories, issued an unconditional upstream `GET` on
every call with no cache read, no cache write and no TTL. A CI fleet re-resolving one dependency
graph therefore asked the same few thousand distinct questions tens of thousands of times a day,
each repeat also holding a slot against `GLOBAL_MAX_CONCURRENCY` and paying the repo-resolve and
permission queries in front of it.

In the traffic sample in #3764 these lookups were 17.8% of all requests and 94% of all 404s, at
~67 identical questions per distinct URI per day. Caching them removes **~98.5%** of the upstream
round trips at a 24-hour TTL and **~88.8%** at one hour.

This has been the behaviour since the `/-/` meta route landed in #1885 (first released in 1.2.1);
before that, attestation requests were not proxied at all.

## What this adds

A new `backend/src/services/npm_attestation_cache.rs` — an in-process, TTL'd negative cache — plus
the gate and wiring in front of `npm_meta_upstream_bytes`. It follows the codebase's existing idiom
of one focused cache module per need (`rpm_repodata_cache.rs` #2521, `npm_packument_cache.rs` #2162)
rather than extending the packument cache; see the notes at the bottom for why.

The scope is deliberately narrow on five axes, because the `/-/` namespace as a whole is **not**
cacheable:

| Axis | Decision | Why |
|---|---|---|
| Path | `/-/npm/v1/attestations/*` **only** | `/-/whoami` and `/-/npm/v1/user` are per-principal, `/-/v1/search` is query-dependent, the audit endpoints are request-body dependent. Caching the meta namespace wholesale would be a security bug, so the gate is an anchored prefix match on one endpoint. |
| Status | `404` / `410` **only** | npm forbids republishing a version, so "no attestation for `pkg@ver`" is a stable fact. A `200` is a real provenance bundle and is never stored, so an attestation published after the first lookup is served the moment upstream has it. `401`/`403` are credential-dependent; `429`/5xx are transient and caching either would turn a blip into a TTL-long outage. |
| Key | `(member_repo_id, upstream_url, meta_path)` | A Virtual repo resolves `/-/*` against the members **this caller** is authorized to read (#3323). Keying on the answering member means the authorization walk still runs on every request and can only reach entries for members the caller may use, while two Virtual repos sharing a member correctly share one entry. Keying on the virtual repo would let one tenant be served an answer fetched for a member they cannot reach. The upstream URL is folded in (hashed, so userinfo credentials never enter a key) so repointing a repository's upstream is never answered from the old upstream's entry. |
| Scope policy | Cache steps aside entirely when `policy.is_active()` | An active policy rewrites meta bodies per repository (#2424, #2551), so a cached body would have to be policy-*specific*, not merely policy-keyed. Skipping leaves that path byte-for-byte unchanged. |
| Query string | Never cached when present | An attestation request carries none; anything that does is a different question. |

Status and body are stored together, so a cached `404` replays as a `404` and never as an empty
`200` — npm treats those very differently.

The cache is in-process and per-replica with **no invalidation path**, which is correct here rather
than a limitation: entries are immutable facts that age out. Entry count (16,384) and per-body size
(4 KiB) are both capped, and `store` sweeps expired entries before evicting live ones, so neither a
large dependency graph nor a broken upstream's error page can grow the map without bound.

### Configuration

| Env | Default | Effect |
|---|---|---|
| `NPM_ATTESTATION_NEGATIVE_CACHE_ENABLED` | `true` | Opt-out, matching `NPM_PACKUMENT_CACHE_ENABLED`. |
| `NPM_ATTESTATION_NEGATIVE_CACHE_TTL_SECS` | `86400` | Bounds how long an attestation added *after* publish stays invisible. `0` disables the cache rather than installing one whose every entry is born expired. |

### Observability

A new `ak_npm_attestation_cache_lookups_total{repository,result}` counter, with `result` one of
`hit`, `miss_stored`, `miss_uncacheable`, `error`. It is deliberately **separate** from
`ak_proxy_cache_lookups_total`: every call site of that metric is in `proxy_service.rs` and it is
the artifact proxy cache's hit rate, so folding a second cache into the same series would silently
change what existing dashboards and alerts measure. Same bounded-label contract — repository only,
never the package spec.

`test_attestation_404_is_fetched_once_and_served_from_cache` asks the same attestation question five
times against a wiremock upstream mounted with `.expect(1)`. On `main` the upstream is contacted
five times and the mock's drop-time verification fails; on this branch it is contacted once. The
DB-backed `test_remote_attestation_404_cached_end_to_end_db` asserts the same thing end to end
through the router against a Remote repository.

## Test Checklist
- [x] Manually tested locally

New tests, all in `backend/src/services/npm_attestation_cache.rs` and
`backend/src/api/handlers/npm.rs`:

**The gate** — `attestation_spec` matches bare, percent-encoded and unencoded scoped specs and
tolerates leading/trailing slashes; it rejects `whoami`, `ping`, `v1/search`, `npm/v1/user`, both
audit endpoints, `user/:u/package`, `org/:o/package`, `by-user/:u`, `all`, the bare prefix with no
spec, and the near-misses `attestations/…`, `npm/v2/attestations/…` and
`evil/npm/v1/attestations/…` that an unanchored prefix check would wrongly accept.

**Keys** — same member/upstream/path share an entry; a different member, a different upstream or a
different version do not; trailing-slash differences do not split one entry in two; an upstream
URL's userinfo never appears in a key; an over-long path is refused rather than keyed.

**What may be stored** — only `404`/`410`; `200`, `201`, `301`, `304`, `400`, `401`, `403`, `429`,
`500`, `502`, `503` all rejected; a body one byte over the cap rejected and one exactly at the cap
accepted.

**Store/lookup** — a stored entry replays status, content-type and body intact; an unknown key
misses; an entry past its TTL misses; a store sweeps expired entries; the entry cap bounds the map
and always keeps the newest write; a zero cap clamps to one entry instead of evicting everything.

**Config** — enabled by default at a 24 h TTL; the opt-out honoured; a zero TTL disables; a custom
TTL honoured.

**Against the wire** (wiremock, via `non_loopback_mock_server` so the SSRF policy is exercised, not
bypassed) — the five-asks-one-fetch test above; a `200` bundle contacted on all three asks and
never stored; `401`, `403`, `429`, `500`, `503` each answered through and never stored; an
unreachable upstream returns `None` and caches nothing, so the handler's existing fall-through to
the local stub is unchanged.

**Handler-level** — `attestation_cache_eligible` true for the attestation path with no query and an
inactive policy, false for six other `/-/` endpoints, false with a query string, and false under
each of the three ways a scope policy becomes active (scopes, `allow_unscoped: false`, name
patterns); `meta_response_from_cache` replays `404` + content-type, and degrades an
unrepresentable status to `502` rather than panicking.

## API Changes
- [x] N/A - no API changes

No new routes and no request/response schema changes: this sits behind an existing route and
changes only how often it reaches upstream. No migration.

## Notes for review

**Why not extend `npm_packument_cache`.** #3764 suggests reusing it. Its entry type
`CachedPackument` carries no status code, and its Redis values are a hand-rolled length-prefixed
encoding gated on `REDIS_ENTRY_VERSION`. Adding a status field means bumping that version and the
`ak:npm-packument:v2:` namespace — flushing every cached packument on a rolling deploy — and
touching every construction site, all to satisfy #3764's own "cache status and body together"
requirement. A separate ~180-line module with no Redis backend and no invalidation seemed the
smaller change; happy to revisit if you would rather see the packument cache generalised.

**Why the policy-inactive option rather than policy-keyed.** #3764 offers either. Under an active
policy an attestation body is not passed through at all — `filter_meta_response` sends it down the
legacy-map-shape branch where `obj.retain(...)` scope-checks each top-level key — so a cached body
would have to be policy-specific, not merely keyed by policy identity. Skipping the cache there is
both simpler and strictly safer.

**`Config` struct literals.** The two new fields required additions to nine full `Config { … }`
literals in existing tests (`conan.rs`, `proxy_helpers.rs`, `test_db_helpers.rs`, `auth_service.rs`,
`ldap_service.rs`, `oidc_service.rs` ×2, and two files under `backend/tests/`). Mechanical, but it
is most of the file count in this diff.

**Known gap: no single-flight on a cold miss.** Concurrent identical requests that all miss will
all fetch upstream, unlike the packument cache, which dedups through
`coordinate_proxy_hydration`. The window is one burst per distinct spec per TTL and it does not
affect the steady-state saving, so I left the primitive out to keep the change small — say the word
and I will wire it through the same seam.

**Other follow-ups deliberately not in scope.** A shared (Redis) backend so the cache is
cluster-wide rather than per-replica; and honouring an upstream `Cache-Control` on the `/-/` path
instead of a fixed TTL. Both fit behind the same gate; neither is needed to remove the
amplification.


brandonrc:
## Review — NEEDS-WORK, one real bug

This is a well-built change and the test discipline is unusually good. Everything I set out to break held up: 403, 408, 429 and 5xx are never negative-cached, which keeps it consistent with #3756; keying on `member.id` preserves the #3323 authorization walk; upstream userinfo is hashed out of the key; the uncached path is byte-identical; metric labels are bounded; and the end-to-end test genuinely fails against main's behaviour, two upstream requests versus one. `cargo check`, `clippy -D warnings` and 321 npm tests are clean.

**Blocker: a trailing slash poisons the entry and masks real attestations.**

`cache_key` normalises with `trim_matches('/')` (`npm_attestation_cache.rs:161`), but the upstream URL is built with `trim_start_matches('/')` (`npm.rs:1734`). So `…/attestations/pkg@ver` and `…/attestations/pkg@ver/` are two different upstream requests that share one cache entry. Against the live registry those two URLs do not agree:

```
sigstore@2.3.1   -> 200, real provenance bundle
sigstore@2.3.1/  -> 404
```

One anonymous request with a trailing slash therefore caches a 404 that answers every provenance check for that package version for the next 24 hours, and `npm audit signatures` quietly reports no provenance instead of verifying it. I reproduced it end to end through the real router against a Remote repository: the probe passes on your branch and fails with the cache disabled, so the 404 is coming from the cache rather than upstream.

The fix is one word — use `trim_start_matches` in `cache_key` so the key matches the URL you actually fetch. Worth noting that the existing test labelled "trailing-slash differences must not split an entry" only exercises a *leading* slash, which is why it misses this; please extend it to the trailing case.

**Nit:** a cache hit still pays both database round trips. The repository resolve and the scope-policy fetch run before the cache is consulted, which is deliberate and is what makes member-keying safe, but the CHANGELOG bullet reads as though the permission queries and the concurrency slot are skipped too. Only the upstream HTTP round trip is.

**Bookkeeping:** the bullet was written when `## [Unreleased]` had a `### Fixed` section. On today's main that section belongs to `## [1.9.0] - 2026-09-08` and `[Unreleased]` is empty, so when you resolve the conflict please create a fresh `### Fixed` under `[Unreleased]` rather than taking theirs — otherwise the entry lands inside a released version.

brandonrc:
@knowinglyAnonymous — thanks for this. It was blocked as DIRTY against `main`, so per the maintainer's
standing request for blocked contributor PRs I've taken it over, rebased it and pushed to your branch
rather than opening a competing PR. Your two commits are preserved with you as author; nothing in the
production logic changed. Here is exactly what I did and the one thing I'd like a maintainer call on.

## Rebase

Rebased onto current `main`. Only `CHANGELOG.md` conflicted — several bullets landed under
`## [Unreleased]` while this was open. Different bullets on each side, so I kept both (#3780, #3770,
#3652 from `main`, plus yours); no bullet was dropped and `CHANGELOG.md` shows zero deletions.

**One thing worth knowing for your next PR**: the 1.9.0 release cut (2026-09-08) moved the
`### Fixed` block your bullet was anchored next to — the #3573 RPM bullet — out of `## [Unreleased]`
and into the released `## [1.9.0]` section. Git's textual merge faithfully followed that anchor, so a
clean, conflict-free-looking resolution silently filed your bullet **inside the released 1.9.0
section**. I moved it back under `## [Unreleased]` → `### Fixed`. The repo has a one-liner for this
that's worth running before every push:

```
awk '/^## \[/{h=$0} /^- \*\*/{print h}' CHANGELOG.md | sort -u
```

## Verification

Everything below was run on the rebased branch, against PostgreSQL 16:

- `cargo fmt --check` — clean
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(api::handlers::npm::) or test(services::npm_attestation_cache::) or test(services::npm_packument_cache::)'` — **320/320 passed**, including the DB-backed `test_remote_attestation_404_cached_end_to_end_db`
- Regression proof: with `NpmAttestationCache::lookup` forced to always miss (i.e. pre-#3764
  behaviour), `test_attestation_404_is_fetched_once_and_served_from_cache` and
  `test_remote_attestation_404_cached_end_to_end_db` both fail on wiremock's `.expect(1)`
  verification; restored, both pass.

No new `backend/tests/*.rs` file is added (the only new file is
`backend/src/services/npm_attestation_cache.rs`), so the `ci_test_surface` gate is not implicated and
no workflow change was needed.

## Review notes

The scoping is right and I checked the part of this that most often goes wrong. Specifically:

- **Status discrimination is correct.** `is_cacheable_status` is `404 | 410` only, and the tests pin
  rejection of `200/201/301/304/400/401/403/429/500/502/503`. A transport error or read failure makes
  `npm_meta_upstream_bytes` return `None`, which stores nothing and falls through unchanged. A 5xx or
  a timeout can never be memoised as "no attestations" — that was my main worry and it's handled.
- **Virtual member fall-through is byte-for-byte preserved.** `proxy_npm_meta_get` already returned
  `Some(response)` for a 404, so the first Remote member's 404 already short-circuited the member
  walk; `proxy_npm_meta_get_cached` returns `Some` on exactly the same conditions. No behaviour change
  there.
- **Keying genuinely substitutes for invalidation on repoint and rename.** Folding
  `sha256(upstream_url)` into the key means a repointed upstream can't be answered from the old
  upstream's entry, and hashing keeps userinfo credentials out of keys (there's a test for that).

## The one open question: the 24-hour default TTL

This is the only thing I did **not** change, because it's a product call rather than a defect, and I'd
rather a maintainer make it than have me quietly pick a number. Flagging it explicitly:

This repo already has a negative-cache TTL convention, and it is **45 seconds** —
`cache_classifier::NEGATIVE_CACHE_TTL_SECS`, used by `proxy_service::write_negative_cache` for upstream
404s on the artifact proxy path. It is a hard constant, deliberately not operator-configurable, and
it is pinned by a unit test to the 30–60 s band. Its rationale is written down in
`docs/audits/design-clean-room-singleflight-cache-2026-06-04.md` §2.3: *"Longer negative TTL → less
upstream load but staler 'not found'; we choose short (30–60s) because false-negatives are
user-visible and confusing."* The RPM remote-proxy plan (S9) then explicitly rejected an 1800 s draft
and reaffirmed 45 s as the house convention.

`NPM_ATTESTATION_NEGATIVE_CACHE_TTL_SECS=86400` is **1920×** that convention, so it deserves a
deliberate decision rather than inheriting one. Two things I'd weigh:

1. **The "immutable fact" argument holds for `registry.npmjs.org`, but this cache applies to any
   configured upstream.** npm's no-republish rule makes "`pkg@ver` has no provenance" stable *on
   npmjs.org*. A Remote npm repository's upstream can equally be Verdaccio, Nexus, Artifactory, or
   another Artifact Keeper — mirrors that warm lazily, or private registries that can backfill
   provenance out of band. For those, a 404 is not necessarily a permanent fact, and a spurious 404
   from a partial upstream outage (one that doesn't surface as a 5xx) is cached for a full day.
2. **There is no lever to clear a bad entry.** The cache is deliberately not wired into
   `ak_cache_invalidation_v1` / `cache_invalidation.rs`, and it isn't reached by
   `conservative_flush_all` either. I think that's defensible given the keying above, but the
   consequence is that a wrong entry cannot be evicted short of a process restart — which is fine at
   45 s and much less comfortable at 24 h.

For what it's worth the failure mode is **fail-safe, not fail-open**: a stale 404 makes
`npm audit signatures` report *missing* provenance for a package that has it. It can never manufacture
a false "verified". But it can break a provenance-gated pipeline for up to a day after an attestation
becomes available, which for a supply-chain endpoint specifically is worth being deliberate about.

Your own figures make the cheap option attractive: **88.8% of the round trips are removed at a 1-hour
TTL vs 98.5% at 24 hours.** Dropping the default to `3600` would keep almost all of the benefit at
1/24th of the worst-case staleness, and operators who know their upstream is npmjs.org can raise it.
I'd suggest that, but I've left `86400` as you wrote it so the decision is visible and explicit rather
than made by me in a rebase.

Nothing else is blocking from my side — with a maintainer's answer on the TTL this is ready.

PR #3766, opened 2026-09-08
fix(oci): honour HTTP Range requests on locally-served blob downloads (#3586)

### LINKED ISSUE #3586: Add support for OCI/Docker Blob `Range` Download Requests

irizzant:
## Summary

A client that retried an interrupted OCI/Docker layer download with `Range: bytes=...` always received the full object with `200 OK`. This made the OCI Distribution Spec resumable-pull use case impossible for clients such as skopeo and container runtimes.

This PR now covers the three blob serve paths in `handle_get_blob`:

- hosted/CAS-hit blobs;
- local members of Virtual repositories;
- Remote pull-through blobs, both cold and warm proxy-cache paths.

All paths use the existing shared `ranged_stream_response` parser/slicer:

- `bytes=START-END`, `bytes=START-` and `bytes=-SUFFIX` → `206 Partial Content` with `Content-Range`;
- unsatisfiable ranges → `416` with `Content-Range: bytes */total`;
- `GET`/`HEAD` advertise `Accept-Ranges: bytes` where the total length is known;
- malformed and multi-range headers retain the existing full `200` behaviour, a valid RFC 7233 choice.

For Remote pull-through, the proxy always drains and caches the complete object. The Range window is applied only to the client-facing stream, so a partial request can never poison the cache with a partial object. If the upstream does not advertise a total length, the response remains a full `200` because a safe `Content-Range` cannot be constructed.

## Verification

- Hosted CAS regression: full `200`, explicit range `206`, suffix range `206`, and unsatisfiable range `416`, including body windows and headers.
- Remote pull-through regression: the same direct Remote repository is tested on a cold stream and after the cache commit; the upstream is expected to be contacted once, while both client requests receive the correct partial windows.
- Existing virtual/OCI streaming tests remain unchanged.
- `cargo fmt` clean on the changed file.

The implementation currently slices the already-streaming object rather than adding a backend-native ranged stream. This preserves constant application memory and cache correctness; a future storage-specific optimization can avoid reading discarded prefix bytes from S3/filesystem backends.

Closes #3586


brandonrc:
## Review — MERGE-WITH-NITS, but it does not close #3586

The implementation itself is good: `cargo check`/`clippy`/`fmt` are green, hostile ranges (`u64::MAX-`, `-0`, `abc`, multi-range, whitespace, a 20-digit end) are all handled without panicking or serving wrong bytes, nothing is buffered, and the new test genuinely fails on main — reverting the branch flips it to `left: 200 / right: 206`. Also useful to know: this does **not** conflict with your #3739; merged sequentially, `oci_v2.rs` auto-merges and only the CHANGELOG collides.

Three things before it lands, the first being the important one.

**It ranges the local serve paths, not the pull-through path the issue was filed against.** `handle_get_blob` has three ways to serve a blob. This PR covers the `oci_blobs` CAS hit and the virtual-local-member path. The Remote path (`oci_v2.rs:5869` → `build_oci_streaming_proxy_response`, which hard-codes `StatusCode::OK` at `oci_v2.rs:4534`) is untouched, and a plain Remote pull never writes an `oci_blobs` row, so the new branch is simply unreachable there. I probed it: a Remote repository over wiremock with `Range: bytes=0-15` returns `200` with the full body, no `Content-Range`, no `Accept-Ranges`, and zero `oci_blobs` rows — cold and warm alike. #3586 describes a Docker pull-through cache, so please drop the closing keyword; this should land as a step toward it, with the Remote path tracked separately or added here.

**"Cold-cache" is not the only case that keeps the 200.** A warm proxy-cache hit returns early from `try_streaming_cache_hit` (`proxy_service.rs:4715`) — a purely local read with no cache hazard — and still answers 200. The rationale in the comment and CHANGELOG (that passing Range upstream would tee a partial object into the cache) does not apply to that path, so either range it or say plainly that warm hits are excluded too.

**A ranged read still costs a whole-object storage read.** The window is sliced in-process from a whole-object `get_stream`; no offset reaches the storage layer. `storage/mod.rs:240` says in as many words that `get_range` exists so consumers "do not have to re-download and discard bytes". Resuming a 1.4 GB layer at 90% reads about 1.29 GB from S3 to deliver 144 MB. That is not a regression, but a resume gains correctness here without the efficiency that is the point of resuming. `get_range` is not a drop-in since it buffers; this wants a ranged stream.

Bookkeeping: the CHANGELOG bullet needs to move to a fresh `### Fixed` under `## [Unreleased]` — 1.9.0 was cut yesterday and the bullet is sitting in the released section.

irizzant:
Addressed the review points in commit `08926f0`:

- The `Range` window now applies to hosted/CAS hits, virtual-local members, and the Remote pull-through path.
- Added a cold-cache + warm-cache Remote regression; the upstream is expected to be contacted only once.
- The complete object is still drained into the proxy cache, while only the client-facing stream is sliced. A partial request therefore cannot poison the cache with a partial object.
- `GET` and `HEAD` advertise `Accept-Ranges: bytes` when the total length is known.
- Fixed the `CHANGELOG.md` placement and removed the conflict markers introduced during the main-branch rebase.

### Remaining trade-off

The current implementation preserves constant application memory but still reads and discards the prefix from the storage backend. A future storage-native ranged stream could optimize S3/filesystem I/O without changing the HTTP or cache semantics.

irizzant:
Correction to the previous note: `GET` and `HEAD` advertise `Accept-Ranges` when the total length is known. The rest of the review response is unchanged.

brandonrc:
Verified your follow-up rather than taking it on trust, and the code holds up. Pushed two CHANGELOG repairs to your branch (`maintainer_can_modify` was on) and rebased onto current main; no production code touched.

## What I changed

Both defects are rebase damage, not authoring mistakes:

1. **The bullet was filed under `## [1.9.0] - 2026-09-08`, a released section**, not `## [Unreleased]`. This is a known trap here (#3797): a branch older than a release cut anchors its bullet to a heading the cut renamed, and the merge is clean, so the entry silently lands in shipped release notes. Moved it.
2. **The newline between the #3652 and #3573 entries was eaten**, welding them onto one 2,223-character line. Split them back apart. Nothing was lost — I checked that no line present on main is absent from the branch, which now reports zero.

## What I verified of your five points

- **Range across hosted / virtual-local / Remote — confirmed.** Five blob-serving arms in `handle_get_blob`; four honour Range and the fifth (buffered `VirtualBlobResolution::Remote`) is dead code, since `resolve_virtual_blob` only ever yields `Local` or `RemoteStream`.
- **Cold+warm regression — confirmed.** The upstream count really is asserted (`wiremock .expect(1)`, checked at drop), and both new tests fail `left: 200, right: 206` with the production hunks neutered.
- **"A partial request cannot poison the cache" — confirmed, but not by the mechanism you describe.** I served responses over a real hyper server rather than `to_bytes` and hashed every object in the cache dir: a ranged cold pull leaves it **empty**, a client abort mid-stream leaves it **empty**, and a full read-to-completion commits 1048576 bytes with a verifying digest. So the safety is real — but it comes from `tee_stream` staging to a private key and publishing only after `classify_stream_write` matches `expected_len`, so a short stream is `RejectTruncated` and discarded. The complete object is *not* drained on a ranged pull: hyper drops the body once `Content-Length` is satisfied, so a ranged pull-through leaves the cache cold. Your warm-cache assertion passes only because `collect_response` drains it. That is a performance gap, not corruption, and no worse than main — but the PR description should say so, and it is worth a follow-up.
- **`Accept-Ranges` — partially.** Local GET and HEAD advertise it; **Remote pull-through with a known length does not** (`remote GET no-Range -> 200 accept-ranges=None`). Only two `Accept-Ranges` sites exist in `oci_v2.rs`, both local HEAD arms. Discovery wart rather than a break, so not blocking.

## Edge cases, since they were not in the review

15 headers through the real handler: `bytes=abc`, multi-range and overflow degrade to 200; `999999-`, `-0`, `10-5`, `4096-` give 416 with `bytes */4096` and no body; `-500`, `0-0` and a clamped `4095-99999` give correct 206 windows. No panics, no 5xx, `Content-Length" always matched the delivered bytes. One nit: `bytes = 8-15` with spaces and `BYTES=8-15` degrade to 200 where RFC 9110 would honour both — safe, and the parsing helper is shared and pre-existing.

## On your stated trade-off

Reading and discarding the prefix is acceptable to merge with. It is O(start) I/O rather than O(window), and the heap stays bounded — but resuming a 5 GiB layer at 90% reads and throws away ~4.5 GiB before the first byte, which is billed S3 transfer and proportional TTFB. Still better than main re-sending the whole 5 GiB. `StorageBackend::get_range` already exists with native S3/GCS/Azure/filesystem overrides but returns buffered `Bytes`; a streaming `get_stream_range` would fix both this and the cold-cache gap above. Worth noting for that follow-up: repeated `bytes=<total-1>-` on a large public blob is a cheap amplification lever.

753/753 tests pass. Merging once CI is green.

Issue #3775, opened 2026-09-08
promote_floating cannot restore an absent :latest — Apply Floating Tags is skipped from the release ref and refused from main


brandonrc:
Part of #3769.

After the untested 1.9.0 / adapter 1.2.10 package versions were deleted (see #3770 for why `adapter:latest` was on them at all), `backend:latest`, `adapter:latest`, `adapter:1.2` and `adapter:1` were gone. The documented recovery, `docker-publish.yml -f promote_version=1.8.2 -f promote_floating=true` (#2698), could not put them back:

- dispatched on `main`: **refused** — `Promote — resolve published adapter digest` fails because `docker/scanner-adapter/VERSION` on main is 1.2.10, which is absent ("run promote from the release commit"). Run 34274903011.
- dispatched on `v1.8.2` as instructed: version tags re-applied, but **`Apply Floating Tags` was skipped**, not refused, with all four `needs` jobs green and no `if:` on the job. Run 34275162734. The skip reason is not surfaced anywhere; the Publish Summary only says "see the Apply Floating Tags job".

Net effect: the one recovery path for floating tags does not work for the case it exists for, and the only fix was a hand `crane tag` by someone holding `packages:write`.

## Ask

1. Find out why the job was skipped (GitHub skips a job whose `needs` include a *transitively* skipped job only if that job is in `needs`; here they are all `success`, so something else is at work — a job-level `concurrency`? a `strategy` with zero matrix rows? an expression in `needs.*.outputs` the runner evaluated as skip?). Surface the reason in the Publish Summary so the next person is not guessing.
2. Make promote resolve the adapter version from the *promoted* release (the adapter `VERSION` at that release's commit), not from the dispatched tree, so `main` is a valid dispatch ref. That is the same rule #3770 needs for the adapter's floating tags, so implement it once.
3. Add a self-test: floating tags absent on the registry + promote of the highest published release → plan advances all of them.

Refs: #2698, #3544, #3770, runs 34274903011 and 34275162734.

brandonrc:
[...]

Issue #3769, opened 2026-09-08
Epic: no permanent name until the gate has passed — candidate-then-promote release flow


brandonrc:
## Why

The immutable git tag is the trigger for testing. Everything downstream of the tag push is irreversible: the tag itself (ruleset 19144026), the `:X.Y.Z` image tags, and for the scanner adapter even `:latest`. The release gate then runs against bytes that already carry permanent names, so a gate failure burns a version number. The `-rc.N` convention exists to avoid this but is advice (RELEASING.md step 6), not enforcement, and it does not exercise every check.

1.9.0 hit every edge at once: a stable tag went straight to the gate, two test-suite defects failed it, the re-runs never reached a ready backend, the tag had to be deleted by admin bypass, and the adapter's `:latest` had already moved to untested bytes. v1.7.2 died the same way; v1.7.0 took four candidates.

## Target

Nothing gets a permanent name until the gate has passed on that exact commit. Test the sha; promote the certified digest; create the tag last.

## Children

- [x] adapter floating tags go post-gate like backend/openscap (#3770) — merged in #3776
- [x] `release-candidate` dispatch: gate a sha with no git tag, record a certification (#3771) — merged in #3777
- [x] stable release as promotion of a certified sha; git tag created last, from the workflow (#3772) — merged in #3777
- [x] candidate runs every check the stable run does, incl. the scanner-adapter exact-version check (#3773) — merged in #3777
- [x] diagnose the release-gate re-run failure mode (backend never ready); until fixed, recovery is a new candidate (#3774) — merged in #3777
- [x] promote_floating cannot restore an absent :latest; surface the skip reason and resolve the adapter version from the promoted release (#3775) — merged in #3653 / #3776, closed
- [ ] docs-only tip cannot get a Docker Publish verdict (#3629)

Refs: #2698, #3433, #3629, RELEASING.md steps 1, 6 and 7.

brandonrc:
## Registry state after the v1.9.0 tag deletion (measured 2026-09-08)

The deleted `v1.9.0` run had already written exact and floating tags on **both** registries, and only the ghcr backend/adapter packages were deleted. What remained, and what I did:

| image | tag | was | now |
|---|---|---|---|
| docker.io/artifactkeeper/backend | `latest` | untested `01d81677…` (the v1.9.0 build) | **reverted** to `1.8.2` = `6f54d0f7…` |
[...]

PR #3776, opened 2026-09-08
fix(ci): defer scanner-adapter floating tags to the post-gate promotion

### LINKED ISSUE #3770: docker-publish: scanner-adapter :latest/:X.Y/:X advance on the normal build, before any gate
Part of #3769.

`docker-publish.yml` applies backend/openscap floating tags only in the post-gate promotion, and its own log says so — then, in the same job, advances `scanner-adapter:latest`, `:1.2` and `:1` to the just-built digest whenever `ADAPTER_FLOATING` is true (`apply-floating-tags`, ~line 2481). Observed on the v1.9.0 tag run: six writes to both registries before a single gate test ran. Deleting the untested 1.2.10 then took `adapter:latest` with it.

Ask: route the adapter through the same `floating-tag-plan.sh` rule as the other images (published, non-draft, non-prerelease release for the backend version that ships it; never move backwards), so a normal build publishes only `:X.Y.Z` and `sha-*`. Add the adapter case to the plan script self-test.

brandonrc:
## Summary

Closes #3770. Part of #3769.

`apply-floating-tags` said "Normal build: backend/openscap floating tags are NOT applied here" and then, in the same step, advanced `scanner-adapter:latest`, `:1.2` and `:1` to the just-built digest whenever `merge-scanner-adapter` reported `floating=true`. On the v1.9.0 tag run that was six writes to both registries before any gate ran, and deleting the untested 1.2.10 image took `adapter:latest` with it.

This PR routes the adapter through the same `floating-tag-plan.sh` rule at the same moment as backend/openscap:

- **`.github/scripts/floating-tag-plan.sh`** gains `--with-major`, which additionally emits `:X` when the target is the newest published version in its major line (the alias the chart pins). Without the flag the output is byte-identical to today.
- **`docker-publish.yml` / `apply-floating-tags`**: in the `promote_floating` branch, after backend/openscap, the step reads the adapter `VERSION` each published stable release carries (`gh api .../contents/docker/scanner-adapter/VERSION?ref=<tag>`; a 404 means the release predates the file and is ignored, any other error fails closed), runs the plan script with that set and the adapter version the promoted ref ships, and plans `advance` writes for exactly the tags it allows. The plain promote keeps its mirror-only behaviour. The normal-build branch writes nothing for any image. The `floating` job output and `publish_floating` are gone, so nothing can key on them.
- **`merge-scanner-adapter`**: an unchanged-source rebuild of a published version now publishes sha tags only (it used to re-point the floating tags at itself, #2391); a changed-source build still demands the VERSION bump. The promote path is unchanged: the digest is resolved from the published exact tag and `assert-version-digest-consistent.sh` still guards it, so a promote can only re-apply the same digest.
- **`release.yml` / `verify-images-published`** probes the adapter's exact `:X.Y.Z` (from `docker/scanner-adapter/VERSION` at the tag) instead of the floating `:1`. The old header comment cited that `:1` probe as the reason the adapter could not be deferred (first cut of a new MAJOR would deadlock); it now verifies what the tag push actually produced, skipped on prereleases as `verify-published` already does.
- `RELEASING.md` policy bullet and CHANGELOG entry.

**Review follow-ups (second commit, from the release-flow review):**
- The promote resolves the adapter version from the *promoted release* (`gh api contents .../VERSION?ref=v${PROMOTE_VERSION}` in `Read adapter version`), not from the dispatched tree, so `main` is a valid dispatch ref (#3775 item 2) and the "run promote from the release commit" errors are gone. The same-digest guard is untouched (`promote_digest` + `assert-version-digest-consistent.sh`).
- `Plan floating tags` plans per image: a refusal for backend/openscap or for the adapter is a `::warning` and records itself; the step fails only when nothing at all can be applied.
- `release.yml`'s post-promote assert now covers the adapter's `:latest` / `:X.Y` / `:X` (expected = ghcr `adapter:<VERSION at the tag>`), so an empty adapter plan can no longer leave a release green with `adapter:latest` stale.
- Only `Not Found (HTTP 404)` (file absent at a live tag) is ignored when building the adapter set; a missing tag (`No commit found for the ref`) fails closed.
- Prose: "no floating tag" rather than "only exact + sha" (`dev` / series-dev tags still apply on branches).

**Round 2 (third commit):** the release-side adapter assert derives its expectation from the same rule as the plan. `.github/scripts/adapter-published-set.sh` (new, shared by both workflows, fails closed on anything but `Not Found (HTTP 404)`) resolves the adapter's published set; the assert runs `floating-tag-plan.sh --with-major` on it and asserts digest equality only for the adapter tags this release's adapter version owns, and mere presence (logged, not asserted) otherwise. A newest backend release shipping an older adapter than a later backport (real history: v1.8.0 shipped 1.2.5 after v1.7.8 shipped 1.2.6) now stays green. Self-tests: the v1.8.0/v1.7.8 shape added to `test-floating-tag-plan.sh` (27 cases), and a stubbed-`gh` `test-adapter-published-set.sh` (6 cases) wired into CI's shell-tests job.

## Design question: what is "has a published release" for an independently versioned image?

The adapter's 1.2.x line does not map onto a backend release, so the rule needs a definition. I chose: **the adapter's published set is the `docker/scanner-adapter/VERSION` carried by each published, non-draft, non-prerelease backend release, and the target is the adapter version the promoted ref ships.** The plan script then applies its unchanged ordering: `:latest` only if the target is the newest adapter version in that set, `:X.Y` / `:X` only if newest in its minor / major line, refused outright (exit 4) if no published release ships it.

Why this rather than "the adapter version the highest published backend release ships with": they agree on every cut so far (VERSION is bumped on main, so the newest release always carries the newest adapter), and the set form is literally the same rule and the same script as backend/openscap, gives `:1.2` and `:1` a definition that the single-version form does not, and makes "never backwards" mean the same thing for the adapter as for the backend (a backport that ships an *older* adapter moves none of its tags; one that ships a newer adapter than any published release moves them, which is what MAJOR-as-compat-contract permits). Rollback is identical too: un-publish the bad release and re-promote the previous one; the adapter set shrinks with it. Against today's published set the promote of v1.8.2 (adapter 1.2.8) plans `1.2`, `1`, `latest`, and the v1.9.0 tag-push shape (1.2.10, no release) is refused.

Cost: the promote makes one `contents` API call per published stable release (40 today, ~10 s), with `GITHUB_TOKEN`.

- [x] `scripts/ci/test-floating-tag-plan.sh` adds nine adapter cases: highest published release takes `:1.2` / `:1` / `latest`; a lower release takes nothing; the v1.9.0 tag-push shape (adapter version with no published release) is refused with exit 4; `:X` follows the major line (patch in an older minor takes `:X.Y` only, new major takes everything, previous major keeps `:X` but not `latest`); prerelease refused; no `--with-major` is byte-identical. Run against `origin/main`'s plan script: 8 of the 9 fail (`--with-major` is a usage error there), 17 existing cases still pass.
- The "normal build writes nothing" case is workflow branch structure rather than script logic; it is not exercisable offline. `scripts/ci/check-floating-tag-promotion.sh` and its self-test still pass unchanged (one writer, fans in from every merge job, release promotes post-gate).

## Verified

- `bash scripts/ci/test-floating-tag-plan.sh` — 25 passed, 0 failed (16 before).
- `bash scripts/ci/test-check-floating-tag-promotion.sh` — 14 passed; `scripts/ci/check-floating-tag-promotion.sh` — OK.
- `bash -n` + `shellcheck` on the plan script, its test, and the three edited `run:` blocks extracted from the workflows — clean.
- `actionlint -config-file .github/actionlint.yaml` (rhysd/actionlint:latest via docker) on `docker-publish.yml` and `release.yml`: 61 findings, identical set to `origin/main` — no new findings.
- Live dry run of the adapter set resolution against the real releases API: 28 releases resolve (1.2.8 … 1.0.0), 12 pre-VERSION releases ignored; plan for 1.2.8 = `1.2`, `1`, `latest`; plan for 1.2.10 = refused (exit 4).
- Not run: an actual promote dispatch. The live proof is the next stable cut: `Apply Floating Tags` on the tag push must plan 0 writes, and the post-gate promote must plan the adapter's three tags alongside backend/openscap.

## Notes for the reviewer

- **#3653 must merge first.** On this branch alone `apply-floating-tags` is still skipped on every promote dispatch (implicit transitive `success()`, #3652; run 34275162734), so merging #3776 without #3653 turns "adapter floating tags move pre-gate" into "adapter floating tags never move". This PR is based on plain `origin/main` and does not touch the job-level `if:` (#3653's line); the two conflict only in `CHANGELOG.md`.
- A promote of a version whose tag was deleted (the v1.9.0 shape) now fails in `Read adapter version` rather than reading the dispatched tree; a release cannot exist without its tag, so no floating promotion is lost.
- The release-side backend/openscap assert still assumes the released version is the newest (pre-existing); the adapter assert no longer does.
- Behaviour change to state plainly: an unchanged-source rebuild of an already-published adapter version no longer delivers Alpine errata to `:1`; that now takes a VERSION bump and a release, which 1.9.0 already did twice (1.2.9, 1.2.10).

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc:
## Audit — release-flow safety, three rounds

**Verdict: merge on green, after #3653.** The reviewer proved the dependency first: on the promote dispatch `Apply Floating Tags` is skipped with all four `needs` green (#3652), so without #3653's direct-result condition this PR would have turned "adapter floating tags move before the gate" into "adapter floating tags never move".

What this PR establishes, verified on the head and on the trial merge with #3653:

- No normal build writes a floating tag for any image any more. Every `crane tag`/metadata `type=raw` in `docker-publish.yml` was traced per trigger; a `v*` push, a main push, a scheduled run and a plain dispatch each produce only exact and `sha-*` tags for backend, openscap and the adapter. On the v1.9.0 tag run the adapter had advanced `:latest`, `:1.2` and `:1` to the untested 1.2.10 digest with six registry writes before any gate ran.
- The adapter's floating tags now follow the same `floating-tag-plan.sh` rule as the other images: its published set is the adapter `VERSION` at each published, non-draft, non-prerelease backend release's commit, resolved by a shared `adapter-published-set.sh` that ignores only `Not Found` and fails closed on a missing tag, a 500 or a 403 before any write. Walked live: promoting 1.8.2 resolves 1.2.8 and plans `1.2`, `1`, `latest`; promoting a lower release refuses; 1.2.10 refuses. Promote now resolves the adapter version from the promoted release rather than the dispatched tree, so `main` is a valid dispatch ref (#3775 item 2), and a per-image refusal is a warning that no longer blocks the other images' planned tags.
- The post-promote assert in `release.yml` derives its expectation from the same newest-in-line rule, so the real historical shape (v1.8.0 shipping adapter 1.2.5 while v1.7.8 shipped 1.2.6) asserts presence, not equality, and stays green; a missing or stale adapter floating tag is red.
- `verify-images-published` probes the adapter's exact tag instead of `:1`, which is guaranteed to exist at that point; the old `:1` probe only worked because `:1` was written before the gate, which is the write this PR removes.

Self-tests 27/27 (plan), 19/19 (guard), 6/6 (published set), all wired into the shell-test job; actionlint and shellcheck finding sets identical to main across three workflows. CHANGELOG under `### Fixed`, `#3770` first.

Issue #3558, opened 2026-08-25
Release binaries ship unsigned, unattested and with no SBOM — only a self-attesting .sha256


brandonrc:
## The gap

Our **container images** are covered end to end: cosign keyless signatures, SBOM
attestation, `provenance: mode=max`, `actions/attest-build-provenance`, and since
#3496 every one of those steps hard-fails its job.

Our **release binaries** are not covered at all.

`release.yml` `build-binaries` produces five artifacts
(`artifact-keeper-{linux,darwin}-{amd64,arm64}`, `artifact-keeper-windows-amd64`)
and ships each one with a `.sha256` file next to it on the GitHub Release. That
checksum proves nothing about authenticity. It is served from the same place, by
the same authority, as the file it describes: anyone who can replace
`artifact-keeper-linux-amd64.tar.gz` on the release can replace
`artifact-keeper-linux-amd64.tar.gz.sha256` in the same action. It detects a
corrupted download. It does not detect a substituted one, and it is routinely
mistaken for a control that does.

There is also **no SBOM for the binaries in any form**. An operator who has an
`artifact-keeper` binary on a host has no supported way to answer "which version
of `<crate>` is in this, and am I exposed to RUSTSEC-XXXX-YYYY?" short of
guessing the version from `--version` and reading `Cargo.lock` on a tag they hope
matches the bytes they have.

## What is missing, concretely

1. **No signature.** Nothing on the release is signed. `cosign verify-blob` has
   nothing to verify against.
2. **No provenance.** `gh attestation verify <binary> --repo artifact-keeper/artifact-keeper`
   fails — there is no attestation naming the binaries as subjects, only the images.
3. **No SBOM.** Neither embedded in the binary nor shipped alongside it.
4. **No single manifest.** Five separate `.sha256` files and no `checksums.txt`,
   so there is nothing that could be signed once to cover the whole asset set.
5. **Nothing verifies before publishing.** Even if signatures existed, the
   workflow would publish first and hope. That ordering matters here more than
   usual: `assert-release-absent.sh` plus `overwrite_files: false` make a partial
   publish unrecoverable — a release that publishes and *then* discovers a bad
   signature has burned an immutable tag.
6. **It is undiscoverable.** `SECURITY.md` documents no verification procedure,
   for the images either. A posture nobody can find is not a posture.

## What "done" looks like

- `checksums.txt` covering every release asset, cosign-keyless-signed (GitHub
[...]

PR #3777, opened 2026-09-08
feat(ci): cut releases as the promotion of a certified candidate

### LINKED ISSUE #3771: release-candidate dispatch: gate a commit by sha with no git tag and record a certification
Part of #3769.

A `workflow_dispatch` taking `sha` (default: main tip) that refuses any commit whose version set is inconsistent, reuses the `sha-<sha>` images Docker Publish already built for it (no rebuild), runs the full Release Gate against those digests, and on green records a certification keyed by sha (a workflow artifact plus a check on the commit; the exact carrier is an implementation choice, but it must be verifiable by sha from the release workflow). Replaces the manual `vX.Y.Z-rc.N` tag as the way to test a release.

### LINKED ISSUE #3772: release as promotion: verify the certification for a sha, re-apply tags to that digest, create the git tag last
Part of #3769.

A `release` dispatch that reads the version from `Cargo.toml` at the certified sha (so version cannot drift from ref — the reason the old free-text `version` input was removed), verifies the certification from the release-candidate flow, re-applies `:X.Y.Z` and the floating tags to the certified digest through the existing PROMOTE mode (#2698), creates the annotated git tag on that sha from inside the workflow, and publishes the GitHub Release. The tag is created last and only ever names tested bytes. `release.yml` on tag push becomes a guard that refuses a tag with no certification.

### LINKED ISSUE #3773: candidate flow must run every check the stable run does, including the scanner-adapter exact-version check
Part of #3769.

RELEASING.md step 6 admits a candidate is not a full substitute: some publish logic keys on a clean `refs/tags/v*` ref and is skipped for prereleases, the scanner-adapter exact-version check being one. In the candidate-then-promote flow the candidate is the only test the release gets, so every stable-only check must run there too. Enumerate them (grep `startsWith(github.ref, ...)` and prerelease conditions in the workflows) and make each run for a candidate against the same inputs.

### LINKED ISSUE #3774: release gate: re-runs after a failure never reach a ready backend — diagnose and either fix or forbid
Part of #3769.

On the v1.9.0 run, attempt 1 produced real test output (two genuine failures, since fixed in artifact-keeper-test #390/#391/#392). Attempts 2 and 3 failed every suite identically with `FATAL: backend not ready at ...svc.cluster.local:8080 after 30s` and wrote no result files, including suites that passed in attempt 1. The gate keys its run ID on `run_id-run_attempt`, so the namespace was probably fresh; what is certain is the deployment never became ready. RELEASING.md currently says the recovery for a failed gate is "re-run the job", which did not work here.

Ask: pull the cluster-side evidence for those attempts (pod events, image pull, readiness probe, resource quota — the harness sets 4000m/8Gi) and fix the cause; if it cannot be made reliable, make RELEASING.md say the recovery is a new candidate and have `release.yml` refuse re-run attempts > 1.

brandonrc:
> **Stacked on #3776** (fix for #3770): this branch contains that commit; merge #3776 first and this PR's diff shrinks to its own commit. Until then the diff shows #3776's hunks too.
>
> **Settings needed from the maintainer: none.** Tag creation uses `GITHUB_TOKEN` with `contents: write`; ruleset 19144026 restricts *updates, deletions and non-fast-forwards* of `refs/tags/v*`, not creation, so the Actions app needs no bypass entry. If a creation rule is ever added, `release-promote.yml` fails at "Create the annotated release tag" with a message naming that bypass list. No PAT, App or environment is introduced. The certification uses the attestations API the repository already uses for images and binaries (`actions/attest`, `id-token: write` + `attestations: write`, declared per job).
>
> **Depends on #3653 landing** for the post-release floating promotion to actually run on a promote (its `apply-floating-tags` job-level `if:`); this PR does not touch that `if:` — it surfaces the skip in the Publish Summary instead (#3775, item 1).

Closes #3771
Closes #3772
Closes #3773
Closes #3774
Part of #3769. Addresses #3775 items 1 and 2.

## Design

**Target (the epic, in the user's words): "wait until everything passes before we do any tags."** Nothing gets a permanent name — `vX.Y.Z`, `:X.Y.Z`, `:latest`, `:X.Y` — until the full gate has passed on that exact commit. Test the sha; promote the certified digest; create the tag last.

```
prep PR merged ──▶ Release Candidate (dispatch, sha) ──▶ green ──▶ Release Promote (dispatch, sha)
                    │ version from Cargo.toml@sha                    │ verify certification + registry
                    │ preflight dispatched + evidence required        │ docker-publish PROMOTE(promote_source_sha) ─▶ :X.Y.Z <- certified digests
                    │ bookkeeping / CHANGELOG / tag+release absent    │ assert :X.Y.Z == certified
                    │ stable-only publish checks (#3773)              │ git tag vX.Y.Z  (LAST, GITHUB_TOKEN)
                    │ same reusable Release Gate on sha-<sha>@digest  │ dispatch docker-publish on the tag (no-op re-apply, verify-published)
                    └─▶ CERTIFICATION (attestation on each digest)    └─▶ dispatch release.yml on the tag
                                                                           │ requires preflight evidence + certification
                                                                           │ asserts :X.Y.Z resolves to the certified digests
                                                                           │ gate SKIPPED (already passed on these bytes)
                                                                           └─▶ binaries, sign, verify images, GitHub Release, floating tags — as today
```

**Certification carrier: a signed attestation on the digests.** `release-candidate.yml` runs `actions/attest` once per image with the manifest-list digest as subject and a predicate (`https://github.com/artifact-keeper/artifact-keeper/attestations/release-candidate/v1`) naming the commit, version, candidate/gate run id, Docker Publish run id, adapter version and all three digests. `scripts/ci/assert-candidate-certified.sh` verifies it with `gh attestation verify oci://<image>@<digest> --signer-workflow <repo>/.github/workflows/release-candidate.yml --predicate-type …`, then checks the predicate names *this* commit, *these* digests (re-read from the registry for `sha-<sha>` at verification time), and that all three came from one candidate run. Chosen over "artifact + commit status" because the subject *is* the bytes (a re-pushed image has no certification), it cannot be forged without the candidate workflow's OIDC identity (a status needs only `statuses: write`, an artifact only `actions: write`), and the repo already produces and verifies attestations with the same pins and BLOCKED/INFRA vocabulary (#3558). A `release-candidate-<sha>` artifact and a `release-candidate/certified` commit status are written for humans; nothing trusts them. The candidate verifies its own certification before declaring success, so the release-side check is exercised on every candidate.

**The gate runs once.** `release.yml` on a certified stable tag verifies the certification and skips `release-gate`; running the gate again after the tag exists is precisely the failure the epic removes. Prerelease tags are exempt and still run the gate in `release.yml` (the `-rc.N` fallback). Every downstream job's `if:` is now explicit with `!cancelled()` and named results, because a job with no status function is skipped whenever any *transitive* dependency is skipped (the #3652 mechanism).

**Tag creation and the missing push event.** The tag is created via the git API with `GITHUB_TOKEN` (annotated; tagger `github-actions[bot]`; message carries the certified digests and run ids). A ref created by `GITHUB_TOKEN` fires no `push` workflows — which is *wanted*: a push-triggered Docker Publish would try to build the tag and be refused by the digest guard. So the promote dispatches `docker-publish.yml` (promote mode) and `release.yml` on the tag ref; `release.yml` gains an input-less `workflow_dispatch` that refuses any ref that is not `refs/tags/v*`. `github.ref`, `github.sha` and the `release.yml@refs/tags/vX.Y.Z` signing identity are identical to a push. `resolve-candidate-publish-run.sh` already accepts a `workflow_dispatch` run on the tag ref (#3640).

**PROMOTE mode gets a source.** `docker-publish.yml` gains `promote_source_sha`: when `:X.Y.Z` is absent, the merge jobs resolve the digest from `sha-<sha>` of that commit (validated: 40-hex, reachable from the dispatched ref via the compare API, sha images present and recording the commit as their revision, certified). `assert-version-digest-consistent.sh` still runs unchanged (absent → allow, present → same digest only), and every sign/attest/verify step runs as on any promote. The adapter's exact tag is created from *its* sha image the same way, with the index revision checked. Per #3775 (2), a promote now resolves the **promoted commit** — `promote_source_sha`, else the commit `v<version>` names — and `merge-scanner-adapter` checks out that commit to read `docker/scanner-adapter/VERSION`, so `main` is a valid ref for a recovery promote. A version whose tag was deleted falls back to the dispatched tree with a warning.

**Hand-pushed stable tags cannot skip the gate.** `release.yml` refuses at `candidate-certified` (before any gate) and `docker-publish.yml` refuses at `publish-preflight` (publishes nothing). Prereleases are untouched.

**#3774 (re-runs).** Both `release-candidate.yml` and `release.yml` run the gate on attempt 1 only and add a `gate-rerun-refused` job that goes red with the diagnosis and the recovery ("fix, then dispatch a new Release Candidate"). Post-gate jobs (signing, Release, floating promotion) remain re-runnable — they are idempotent and the tag is already correct, so refusing them would burn a version on a Docker Hub blip. (c) — a namespace-exists pre-check — belongs in `artifact-keeper-test`'s `release-gate.yml` (`deploy` builds `RUN_ID="e2e-$(date +%s)-${GITHUB_RUN_NUMBER}"`); there is no input on the reusable gate to express it from here, so it is noted for that repo.

## Parity (#3773): every check that ran only on a clean `refs/tags/v*` ref or was skipped for prereleases

| Check | Where it lived (stable-only) | In the candidate |
|---|---|---|
| Moved/force-pushed tag (`Reject moved version tags`) | release.yml, docker-publish.yml `push` on `refs/tags/v*` | n/a: no tag exists; the promote creates it and reads it back (`git/tags/<obj>` must name the sha) |
| No existing GitHub Release (`assert-release-absent.sh`) | release.yml `release-preflight` + `release` | `bookkeeping` job, on `vX.Y.Z`; plus a new "tag already exists" refusal |
| Curated notes file / open `[Unreleased]` / no orphaned notes (`assert-release-bookkeeping.sh`) | release.yml `release-preflight` | `bookkeeping` job, at the commit |
| Green preflight evidence (`assert-preflight-evidence.sh`) | release.yml `preflight-evidence` | `preflight` job: dispatches `release-preflight.yml` with `ref=<sha>`, follows it, asserts the evidence (leaves the artifact the tag run needs) |
| CHANGELOG `## [X.Y.Z]` non-empty (inline in `verify-images-published`; gate's `version-set-integrity` only when `backend_tag` is X.Y.Z) | release.yml, stable only | extracted to `scripts/ci/assert-changelog-documents-version.sh`, run in `bookkeeping`; release.yml now calls the same script |
| Curated notes required (`Resolve release notes`) | release.yml `release`, stable only | covered by `assert-release-bookkeeping.sh` check 1 |
| Exact-tag digest guard (`assert-version-digest-consistent.sh`) for backend/openscap on ghcr + Docker Hub | docker-publish.yml merge jobs, `exact != ''` only on `refs/tags/v*` | `images` job, with the sha digests: `:X.Y.Z` absent or already these bytes, on both registries |
| Scanner-adapter exact-version decision (`stable_requested`; both registries; sources diff; VERSION collision; provenance) | docker-publish.yml `merge-scanner-adapter`, clean `refs/tags/v*` only | new `scripts/ci/assert-adapter-exact-version.sh`, same registries, same source paths, same refusals; self-tested |
| Adoption gate (`assert-adopted-image-sources.sh`: published revision vs tag) | release.yml `resolve-candidate-digest` | `images` job, strict form: each sha image's recorded revision must equal the commit |
| `APP_VERSION` build-arg (tag builds overwrite Cargo.toml's version) | docker-publish.yml build jobs on `refs/tags/v*` | no rebuild happens; the candidate refuses unless `Cargo.toml@sha` *is* the stable X.Y.Z, so the sha image already reports the released version |
| `verify-published` (ghcr semver tags incl. adapter exact) | docker-publish.yml, `startsWith(github.ref, 'refs/tags/v')` | runs on the promote's dispatch on the tag ref (step 7), before `release.yml` |
| Floating-tag promotion (`!contains(version, '-')`) | release.yml post-release | unchanged, post-release (and now adapter-aware via #3776) |
| Prerelease exemptions (`contains(...,'-')`) in CHANGELOG/notes checks | release.yml | the candidate certifies stable versions only; prereleases keep the old path |

## Implementation

- **New** `.github/workflows/release-candidate.yml` — jobs `resolve`, `preflight`, `bookkeeping`, `images`, `release-gate` (`uses: artifact-keeper/artifact-keeper-test/.github/workflows/release-gate.yml@main`, `backend_tag: sha-<short>`, `backend_digest`, `web_tag: latest`, `test_suite: all`, `secrets: inherit`), `gate-rerun-refused`, `certify`.
- **New** `.github/workflows/release-promote.yml` — one job, the eight steps above; idempotent on re-dispatch.
- `release.yml` — `workflow_dispatch` (no inputs, tag refs only); `candidate-certified` job; `resolve-candidate-digest` asserts `:X.Y.Z` (backend, openscap) == certified digests; `release-gate` skipped when certified and attempt-1 only; `gate-rerun-refused`; explicit `if:` on `build-binaries`, `verify-images-published`, `release`, `promote-floating-tags`; CHANGELOG check moved to the shared script. `assert-release-absent.sh`, `verify-release-assets.sh`, `floating-tag-plan.sh` and the immutability checks are untouched.
- `docker-publish.yml` — `promote_source_sha` input; `publish-preflight` validates it, resolves `promote_commit` / `promote_source_tag`, and refuses an uncertified stable tag push; merge-backend/openscap resolve `promote_ref`; `merge-scanner-adapter` checks out the promoted commit and creates a new adapter exact tag from its sha image; `summary` states the `Apply Floating Tags` result and explains a skip (#3775). Textual overlap with #3776 (built on it, its semantics kept) and with #3653 (not touched).
- **New scripts**, each with a `test-*.sh` wired into `ci.yml` `shell-tests`: `scripts/ci/assert-candidate-certified.sh`, `.github/scripts/registry-tag-digest.sh`, `scripts/ci/assert-adapter-exact-version.sh`, `scripts/ci/follow-dispatched-run.sh`, `scripts/ci/assert-changelog-documents-version.sh`.
- `RELEASING.md` — cut sequence is prep PR → candidate → green → promote; step 6's `-rc` tag is the fallback; recovery for a failed gate is "fix, then a new candidate" (never "re-run failed jobs"); policy bullets. CHANGELOG bullet under `### Changed`.

## Verified

- `bash scripts/ci/test-{assert-candidate-certified,registry-tag-digest,assert-adapter-exact-version,follow-dispatched-run,assert-changelog-documents-version}.sh` — all cases pass (14 / 9 / 9 / 9 / 7). Existing: `test-assert-preflight-evidence.sh`, `test-resolve-candidate-publish-run.sh`, `test-floating-tag-plan.sh` (25), `test-check-floating-tag-promotion.sh` (14), `test-assert-release-bookkeeping.sh`, `test-check-changelog-unreleased.sh`, `test-verify-release-assets.sh` — pass.
- `check-floating-tag-promotion.sh`, `check-registry-publish-parity.sh`, `check-supply-chain-soft-fail.sh`, `check-trivy-gate-severity.sh`, `check-changelog-unreleased.sh` — OK on the edited workflows.
- `shellcheck` clean on every new script and test. `actionlint -config-file .github/actionlint.yaml`: `release-candidate.yml`, `release-promote.yml` — 0 findings; `release.yml` 3, `ci.yml` 31, `docker-publish.yml` 53 — identical counts to the base branch (no new findings; the existing ones are pre-existing SC2086/SC2129 in untouched steps).
- Live: `sha-7b73dc8` resolves on ghcr for all three images (backend `sha256:02167eb3…`, openscap `sha256:65d6d1ac…`, scanner-adapter `sha256:83fa266f…`), which is the tag shape the candidate reads; ruleset 19144026 rules are `update`, `deletion`, `non_fast_forward` (creation unrestricted); the repo's secrets carry no PAT for this repo, which is why none is used.
- **Not run:** an actual candidate/promote dispatch — the first stable cut after merge is the live proof. Suggested rehearsal: dispatch `Release Candidate` on the 1.9.0 prep commit once main is at a certifiable state, expect CERTIFIED, then `Release Promote`.

## #3775 notes

1. Why `Apply Floating Tags` was skipped on run 34275162734 with four green needs and no `if:`: GitHub skips a job whose `if:` has no status function when any **transitive** dependency was skipped, and on a promote every build/scan job is skipped by design. That is #3652, fixed by #3653's job-level `if:` (open). The Publish Summary now prints the job's result and this explanation when it is `skipped`.
2. The promote now reads the adapter version from the promoted commit (see Design), so `main` is a valid dispatch ref; run 34274903011's refusal goes away once a tag (or `promote_source_sha`) resolves the commit.
3. The self-test for "floating tags absent + promote highest release" belongs to `floating-tag-plan.sh` (#3776's script) and is not in this PR.

## Test Checklist
- [x] Unit tests added/updated (shell self-tests, wired into `shell-tests`)
- [x] Manually tested locally (self-tests, static checks, actionlint, live registry/ruleset reads)

## API Changes
- [x] N/A - no API changes


brandonrc:
## Review round 1 (adversarial release-flow review) — addressed in aec388f6

Rebuilt on `main` (fc90dd6c, #3776 merged) so the diff carries only this PR's change. Per finding:

- **F1 (blocker, reverted #3776 r2/r3):** the commit was re-applied on top of #3776's final tree. `adapter-published-set.sh`, its self-test, the `test-floating-tag-plan.sh` r2 cases, `release.yml`'s adapter post-promote assert and the `#3770` `### Fixed` bullet are byte-identical to `main`; the `apply-floating-tags` block is unchanged. The only intentional replacement of r2 is in `merge-scanner-adapter`: the adapter `VERSION` is read from the `promote_commit` checkout (the tag's commit, or the certified candidate whose tag does not exist yet) instead of `contents?ref=v<V>`; r3's release-side assert stays on top.
- **F2 (signer pin without ref):** `assert-candidate-certified.sh` now passes `--source-ref refs/heads/main` (`CERT_SOURCE_REF`); `release-candidate.yml`'s `resolve` refuses any dispatch ref but `refs/heads/main` before reading anything. The self-test stub refuses a verify call that does not pass `--source-ref`, and a new case blocks a certification signed on `refs/heads/wip`.
- **F3 (hand-pushed stable tag on a certified commit builds untested bytes):** a `push` of `refs/tags/vX.Y.Z` is now refused unconditionally, before any build, in `docker-publish.yml`'s `publish-preflight` and in `release.yml`'s `release-preflight`, with a message pointing at `release-promote.yml`. Prerelease tags (`v*-rc.*` etc.) are untouched: still pushed by hand, still build and run the gate as before this PR.
- **F4 (`sha=<not the tip>` can never pass):** `assert-preflight-evidence.sh` also finds runs through the `release-preflight-<sha>` artifact (`/actions/artifacts?name=`), reads each run back and requires it to belong to `release-preflight.yml`; both routes feed the same "every run completed, newest must be green with evidence" rule. Five self-test cases added (found by evidence; superseded by a newer red found by `head_sha`; de-duplicated; artifact from another workflow ignored; lookup failure is INFRA). RELEASING.md step 6 now states this instead of promising the non-tip case worked.
- **F5 (dead tag when `release.yml` at the commit has no `workflow_dispatch`):** `resolve` in both `release-candidate.yml` and `release-promote.yml` checks `release.yml` has `workflow_dispatch:` and `docker-publish.yml` has `promote_version:` at the commit (the candidate also checks the `scripts/ci` it runs at the sha exist) and fails with a clear message before anything is dispatched or named; `docker-publish.yml`'s `promote_source_sha` validation makes the same `release.yml` check via the contents API before `:X.Y.Z` is applied.
- **F6 (`first` statement):** among statements whose `commit_sha` is the commit, the highest `candidate_run_id` common to all three images is selected; a statement for another commit on the same digest is ignored rather than fatal. Four self-test cases added (two attestations per digest in differing order; common-but-not-newest; foreign commit listed first; no common run still refused).
- **F7:** the step-4 failure message no longer claims "no permanent name exists yet"; it and RELEASING.md step 7 say the version belongs to the certified digests from that point.
- **F8:** `gh workflow run` returns nothing that identifies the run and the Actions API does not expose `workflow_dispatch` inputs, so step 4 keeps the follow-by-timestamp selection; the limitation and why it fails closed (step 5) are documented at the dispatch step.

Verified: the eight touched/added `scripts/ci/test-*.sh` pass (the two edited gates' tests fail against the pre-fix scripts: 15 and 2 cases); `shellcheck -S style` clean on the changed scripts; actionlint (docker `rhysd/actionlint`) on the five workflows: no new messages vs `main`'s set (one SC2129 line moved); `git merge-tree` against #3653 conflicts only in `CHANGELOG.md`.

brandonrc:
## Review round 2 (N1–N4) — addressed in 2c96855b

- **N1 (`--source-ref` is case-insensitive in gh):** `assert-candidate-certified.sh` now pins the exact SAN with `--cert-identity https://github.com/<repo>/.github/workflows/release-candidate.yml@refs/heads/main` (byte-for-byte) in place of `--signer-workflow`, and keeps `--source-ref refs/heads/main` as a second check; header comments in the script and `release-candidate.yml` say why neither looser pin suffices. The self-test stub now requires the exact `--cert-identity` value, exits 64 on any `--signer-workflow` call, folds case on `--source-ref` the way gh does, and has a new case: a certification signed on `refs/heads/Main` is BLOCKED. The candidate's `resolve` ref check is a shell `[[ "$GITHUB_REF" != "refs/heads/main" ]]` instead of an `if:` expression (Actions expressions compare case-insensitively).
- **N2 (plain promote with a missing tag fell back to the dispatched tree):** restored #3776's hard error in `Validate promote request`; a missing `v<V>` fails with "Adapter version unresolvable" and points at `promote_source_sha` for the untagged certified candidate. The certified path is unchanged.
- **N3:** `docker-publish.yml`'s refuse-step comment and RELEASING.md's guardrail bullet now say that only the `promote_source_sha` dispatch validates the certification and that the plain promote on the tag can only re-apply an existing `:X.Y.Z` to its own digest.
- **N4:** four of the five artifact-route cases now fail against `main`'s `assert-preflight-evidence.sh` (green found only by evidence; newer red found only by evidence supersedes a green by `head_sha`; foreign-workflow artifact skipped while the genuine run decides, counted once; evidence lookup failure is INFRA); the remaining "counted once" case is labelled as a regression guard.

Verified: both touched self-tests pass (30/30 and 22/22); the candidate self-test fails 16 cases against the pre-N1 script; `shellcheck -S style` clean; actionlint on the five workflows unchanged vs `main` (the one moved SC2129); `merge-tree` vs #3653 and `main` clean.

brandonrc:
## Audit — three adversarial rounds

**Verdict: merge on green (round 3: MERGE-WITH-NITS; nits fixed in dd4c4907).**

Round 1 (43e6345b) found the commit had been authored against #3776's first revision and silently reverted its later fixes, and four should-fix gaps. Round 2 (aec388f6) confirmed all eight closed and found that gh compares the source-ref extension case-insensitively, so a branch named `Main` could still mint an accepted certification, and that the plain promote path had lost #3776's hard error on a missing release tag. Round 3 (2c96855b) confirmed both closed, tracing `--cert-identity` through cli/cli into sigstore-go's byte-exact SAN compare.

What the merged PR guarantees, each verified on the head rather than from the commit message:

- **No permanent name before the gate.** The candidate certifies the `sha-<sha>` digests already published by Docker Publish; `:X.Y.Z` and the `vX.Y.Z` tag are written only by the promote, after `gh attestation verify` succeeds against those exact digests on both registries. A hand-pushed stable tag is refused before any build in both Docker Publish and Release, unconditionally, because the promote creates its tag through the API and never fires a push event.
- **Certification is bound to the workflow on main, exactly.** `--cert-identity …/release-candidate.yml@refs/heads/main` (exact SAN) plus `--source-ref refs/heads/main`; the candidate refuses any dispatch ref other than `refs/heads/main` with a case-sensitive shell compare. Branch copies, `Main`, `main/x`, tag refs, forks and reusable-workflow calls are all refused; a gate that failed or was cancelled never reaches the certify job; re-runs of the gate are refused by attempt pinning (#3774).
- **Fails closed on registry doubt.** Absent digest → BLOCKED, unreadable registry → INFRA, several attestations on one digest → highest candidate run common to all three images or refuse, tag-resolution errors other than 404 → error rather than "absent".
- **Dead-tag shape excluded.** Before any tag or `:X.Y.Z` write, both candidate and promote check that the target commit's `release.yml` has `workflow_dispatch`, that `docker-publish.yml` has `promote_version`, and that the helper scripts exist at that sha; the preflight evidence lookup also accepts a run identified by its `release-preflight-<sha>` artifact, so a non-tip commit can be certified.
- **#3776 intact.** `adapter-published-set.sh`, its tests, the `apply-floating-tags` block and the `#3770` bullet are byte-identical to main; the release-side adapter assert is kept.

Self-tests 8/8 (candidate-certified 16 cases fail against the pre-fix script; preflight-evidence 4/5 new cases fail against main's), shellcheck clean on 12 scripts, actionlint finding sets identical to main, merge-tree clean against main and against #3653. Registry prerequisites for the first cut are on #3769.

PR #3780, opened 2026-09-08
fix(ci): read the blob before grepping in the candidate/promote readiness checks

brandonrc:
## Summary

The candidate and promote `resolve` jobs check that the target commit's workflow files carry the candidate-then-promote flow before anything is dispatched or named (#3777, F5). The check was `git show "$SHA:$file" | grep -qE "$pattern"` under `set -o pipefail`: on a **match** early in a long file grep exits immediately, `git show` receives SIGPIPE, the pipeline status is non-zero and the negated test reports the file as lacking the pattern. `docker-publish.yml` matches `promote_version:` on line 42 of about 3,000, so the first real candidate, run 34289368258 on 80b7acd9, was refused with "has no promotable docker-publish.yml" although the input exists. The smaller `release.yml` passed the same check by timing.

Both sites now read the blob into a variable and grep that.

Part of #3769

## Verified

- Reproduced locally under `set -euo pipefail` against 80b7acd9: the old pipeline reported "lacks" on repeated runs; the new form finds the pattern.
- actionlint (docker) on both workflows: no findings.
- No other `git show | grep -q` sites in the release workflows or `scripts/ci`.


brandonrc:
## Audit

Review found a third copy of the same shape that the first push missed: `docker-publish.yml`'s promote validation did `gh api …/contents/release.yml | grep -qE 'workflow_dispatch:'`, and `release.yml` is 94 KB with the match on line 33, so every promote would have aborted at "Follow the promotion" after the candidate certified (reproduced 5/5, `PIPESTATUS=141 0`). Fixed in the same commit; the CHANGELOG bullet now names all three sites.

Also verified by the reviewer: the new form under `bash -eo pipefail` with a 160 KB herestring, a missing file (→ "lacks", the intended refusal), patterns starting with `-`, and anchored patterns; a sweep of the release workflows and `scripts/ci` for other `<producer> | grep -q` sites found only small producers (`Cargo.toml`, `printf` of short variables), noted as latent in `$S/reviews/3780-readiness.md`. actionlint: no findings on either tree.

PR #3788, opened 2026-09-09
feat(pypi): legacy JSON API and XML-RPC browse for JupyterLab's Extension Manager

brandonrc:
## Summary

Closes #3783

JupyterLab 4's default `PyPIExtensionManager` discovers prebuilt extensions with an XML-RPC `browse(["Framework :: Jupyter :: JupyterLab :: Extensions :: Prebuilt"])` call and reads each one's version and description from `GET {base_url}/{name}/json` and `GET {base_url}/{name}/{version}/json` — two pypi.org APIs the PyPI handler did not serve (installs already went through `/simple/`). This PR adds both to the existing PyPI handler, for hosted, remote and virtual repositories. Configuration on the JupyterLab side: `c.PyPIExtensionManager.base_url = "https://<ak>/pypi/<repo>/pypi"` for the sidebar, plus — because the install button runs `pip install` with no index argument — pip's index pointed at the same repository's `/simple/` (`PIP_INDEX_URL` or `pip.conf`).

**Legacy JSON API** — `backend/src/api/handlers/pypi.rs`
- `GET /pypi/{repo}/pypi/{project}/json` and `/pypi/{repo}/pypi/{project}/{version}/json` in the legacy shape: `info` (`name`, `version`, `summary`, `classifiers`, `requires_python`, `home_page`, `project_urls`, `author`, `license`, …), `releases` keyed by version (project form only, as Warehouse does), `urls` for the described release; each file carries `filename`, `url`, `digests.sha256`, `packagetype`, `python_version`, `requires_python`, `size`, `upload_time(_iso_8601)`, `yanked`.
[...]

PR #3813, opened 2026-09-10
feat(repositories): add internal visibility state

### LINKED ISSUE #3812: Repositories need an `internal` visibility state between public and private
## Problem

Repository access has exactly two states, encoded in the single
`repositories.is_public` boolean: anonymous-readable, or grant-holders-only.
There is no way to express "readable by every authenticated principal, but
never by anonymous" — the state most internal repositories on a corporate
instance actually need.

Operators reach for the server-wide `AK_GUEST_ACCESS_ENABLED` flag (#850)
instead, and using it that way is actively lossy:

- `coerce_is_public_for_create` / `coerce_is_public_for_update`
  (`backend/src/api/handlers/repositories.rs`) silently rewrite a requested
  `is_public = true` to `false` when guest access is disabled. The operator's
  intent is **destroyed**, not reinterpreted — re-enabling guest access does not
  restore it.
- The guest-access guard allowlists `/v2/*` wholesale
  (`backend/src/api/middleware/guest_access.rs`), and `/v2` is mounted outside
  `repo_visibility_middleware` (`backend/src/api/routes.rs`). Rows already
  carrying `is_public = true` therefore keep serving anonymous `docker pull`
  even with guest access disabled, because the OCI read gate's own check is
  `is_anon && !repo.is_public`.
- It is a single global switch, so "these repos are genuinely public, those are
  internal" cannot be expressed at all.

## Proposal

Add a first-class `visibility` axis on the repository:
`visibility ∈ {public, internal, private}`, persisted in the database and
independent of any server-wide flag.

`internal` grants a **read baseline to any resolved principal**. It confers no
write, delete, admin, or configuration capability, does not relax any
repository-scoped token restriction, and is invisible to anonymous callers on
every surface — direct fetch, listing, search, native protocols, OCI.

Scope:

- New migration adding the enum column, backfilled `is_public = true -> public`,
  `false -> private`. No repository becomes `internal` automatically.
- `is_public` kept as a real, writable column mirrored by a trigger, so the
  out-of-tree `nicola-preda/artifact-keeper` Terraform provider and older
  clients keep working unmodified. API stays additive.
- Guest-access coercion retargeted from `private` to `internal`, so the toggle
  becomes intent-preserving. This is a deliberate behaviour change and is wider
  than today's coercion — it needs a BREAKING (behaviour) changelog entry and an
  operator review step.
- The repository-changed NOTIFY trigger must also fire on `visibility`,
  otherwise a repository narrowed from `internal` to `private` keeps being served
  under the old decision from `repo_cache` until the TTL expires. That is a
  stale-authorization window, so it is a correctness requirement.

## Why not a broad ACL principal

Adding `principal_type = 'authenticated'` to the `permissions` table is a much
smaller diff and inherits audit and cache invalidation for free, but "internal"
is a property of the repository, not a grant on it. As an ACL row it is
indistinguishable from an ordinary private repository in listings, in the
settings UI, and in the Terraform configuration that declares the repository —
an operator cannot answer "which repositories are org-readable?" without
querying the ACL table. It also invites `write` on the same broad principal,
which is exactly what #2603 G1 closed.

## Security note

This touches the authorization read path. The load-bearing invariant is that
`internal` differs from `private` in exactly one respect — on the **read** path,
the set of principals satisfying the baseline becomes "any resolved principal"
instead of "grant holders". Everywhere else (write, delete, admin,
configuration, tenancy pre-gates, anonymous listing) `internal` must behave as
`private`. In particular the shorthand `visibility != 'private'` is **wrong** as
a blanket substitution for `is_public`: applied to `require_repo_write_access`
it would widen the tenancy pre-gate across most of the instance.

mymarche:
Closes #3812

## Review round 2 — what changed since the review

[@brandonrc's review](https://github.com/artifact-keeper/artifact-keeper/pull/3813#issuecomment-5642139792) blocked on four findings, two nits and a test
gap. All of them are addressed here; the reasoning behind F2 and F3 is in
[my reply](https://github.com/artifact-keeper/artifact-keeper/pull/3813#issuecomment-5645558921).

- **F1 — the virtual byte fan-out was still two-state** (`proxy_helpers.rs:3483`).
  The read-baseline short-circuit is now
  `member.visibility.allows_authenticated_read()`. The anonymous arm returns
  earlier (`:3447`), so this branch is authenticated by construction.
  `virtual_parent_and_direct_fetch_agree_on_an_internal_member` asserts that the
  fan-out and a direct fetch give the SAME answer for an `internal` member, and —
  in the same test, deliberately — that a `private` member is still refused
  through the parent, so a fix that returned `true` unconditionally fails it.
- **F2 — the CHANGELOG advertised a `/v2` fix this diff does not contain.** Bullet
  dropped. The bug is real and is now #3854: `/v2/*` is allowlisted wholesale by
  the guest-access guard, so an anonymous pull of a *public* repository still
  succeeds with `AK_GUEST_ACCESS_ENABLED=false`. Fixing it means narrowing the
  allowlist to the challenge paths and giving the guard an OCI-shaped 401
  envelope — a change to the guest-access allowlist, not to the visibility axis.
- **F3 — the coercion target is reverted to `private`.**
  `coerce_visibility_for_create`/`_for_update` now behave exactly as they did
  before this PR. The callers that reach the coercion at all are the ones
  speaking only the legacy boolean — the Terraform provider, older SDKs — which
  can neither ask for `internal` nor decline it, so coercing them to it would
  have made their repositories org-readable on precisely the instances that
  disabled guest access to prevent that. This PR therefore changes no existing
  behaviour; it only adds a state that has to be asked for by name, and
  `visibility: "internal"` still passes through untouched with guests disabled.
  Whether the coercion *should* eventually target `internal` is its own product
  decision. The sharper underlying problem — that the coercion is silent in
  either direction, so a Terraform provider reads back a value it did not ask
  for and proposes the same change on every plan — is #3855.
- **F4 — the SBOM/CVE read gate was still two-state** (`sbom.rs:1864`, the three
  resolvers at `:1898`/`:1925`/`:1945`, and the license-compliance lookup). All
  of them read `visibility` and gate on `allows_authenticated_read()`, matching
  `check_artifact_visibility` exactly: metadata *about* an artifact must not be
  harder to reach than the artifact. `private` still requires a grant, asserted
  against the same caller in the same test.
- **Nits.** The VS Code gallery refusal names the repository's actual visibility
  instead of calling every non-public state "private", and says why no grant
  would help (a gallery client presents no credential at all). The decision is
  unchanged — `allows_anonymous_read()` refuses `internal` exactly as
  `is_public = false` did. `docs/repository-visibility.md` is rewritten around
  the reverted coercion.
- **The test gap.** `internal_repo_reaches_the_middleware_no_rules_arm` and
  `..._has_rules_arm` drive `/pypi/{key}/simple/` through the production router
  against a real database, once with fine-grained rules present on the
  repository and once without — the two arms of `repo_visibility_middleware`
  (`auth.rs:2600`, `:2739`) — asserting anonymous → 401,
  authenticated-without-grant → 200, and a repo-scoped token whose allow-list
  excludes the repository → 404.

Two `is_public` reads were looked at and deliberately left alone.
`conda.rs:1070` gates anonymous-vs-authenticated rather than public-vs-private,
which gives `internal` the right answer as written. `virtual_has_private_member`
must keep reading the mirror: `internal` is not public there, so an aggregate
containing an internal member stays out of the caller-INdependent cache, which
is the correct answer — its neighbouring doc comment is updated to say so.

## Summary

Adds a first-class `visibility` axis on repositories — `public` / `internal` /
`private` — persisted in migration 217 and independent of any server-wide flag.

Repository access previously had exactly two states, encoded in the single
`repositories.is_public` boolean. There was no way to express "readable by every
authenticated principal, but never anonymously", so operators reached for
`AK_GUEST_ACCESS_ENABLED` instead — which destroyed their intent (a requested
`is_public = true` was rewritten to `false` and never restored) while still
serving anonymous `docker pull` for rows already marked public, because `/v2` is
mounted outside `repo_visibility_middleware` and gated on its own
`is_anon && !repo.is_public`.

`internal` grants a read baseline to any resolved principal. It confers no
write, delete, admin or configuration capability, does not relax any
repository-scoped token restriction, and is invisible to anonymous callers on
every surface — direct fetch, listing, search, native protocols and OCI.

### The load-bearing invariant

`internal` differs from `private` in exactly one respect: on the **read** path,
the set of principals satisfying the baseline becomes "any resolved principal"
instead of "grant holders". Everywhere else — write, delete, admin,
configuration, tenancy pre-gates, anonymous listing — `internal` behaves as
`private`.

Stated the other way round: the shorthand `visibility != 'private'` is **wrong**
as a blanket substitution for `is_public`. Applied to
`require_repo_write_access` it would widen the tenancy pre-gate across most of
the instance, because `internal` will be common where `public` was rare. That
site therefore carries an inline comment saying why it must not widen, and a
dedicated regression test.

### Compatibility

`is_public` is kept as a **real, writable column** — not a view, not a generated
column, both of which would break existing write paths and the out-of-tree
`nicola-preda/artifact-keeper` Terraform provider. A trigger holds it equal to
`visibility = 'public'`, so legacy readers see `internal` as not-public, which is
the correct degradation.

The trigger derives the mirror from the column the statement **changed**, not
from its value. This is the subtle part and must not be "simplified": an
`internal` repository already carries `is_public = false`, so a legacy client
that rewrites `is_public = false` on every update — exactly what the Terraform
provider does, since it declares the field and sends its whole desired state —
produces no column change and leaves `visibility` alone. A value-based rule
would silently narrow every `internal` repository on each such write, and the
drift would be invisible in `terraform plan`. Verified idempotent across
repeated writes against live PostgreSQL.

The column deliberately has **no DEFAULT**: a `DEFAULT 'private'` makes "the
caller supplied nothing" indistinguishable from "the caller asked for private",
which forces the INSERT trigger to guess — and the guess resolves toward
`public`, the wider state. Leaving it NULL until the BEFORE trigger fills it
turns that guess into a fact, and lets the trigger REFUSE the one genuinely
contradictory pair (`is_public = true` with a non-public `visibility`) instead
of resolving it. The mirror-image pair is deliberately not refused, because
`false` is also the boolean's default.

### Behaviour change: none

An earlier revision of this branch rewrote a requested `public` to `internal`
when guest access was disabled (design D5). It was reverted on review — see F3
above — so the coercion still lands on `private`, exactly as it did before this
PR, and no existing client sees a different outcome from any path this change
touches. The new state is reachable only by asking for it by name.

Backfill is exactly access-preserving: `is_public = true -> public`,
`false -> private`. **No repository becomes `internal` automatically.** Verified
on PostgreSQL by snapshotting every repository's audience under the old
predicate, applying the migration, and comparing under the new ones: 0 widened for
anonymous callers, 0 widened for grant-less authenticated callers, 0 changed
audience in either direction.

### Cache invalidation

The migration-142 repository-changed NOTIFY trigger now also fires on
`visibility`. Without it, a repository narrowed from `internal` to `private`
keeps being served under the old decision from `repo_cache` on every instance
until the TTL expires — a stale-authorization window, so this is a correctness
requirement rather than an optimisation.

### Two findings from the implementation, both fixed here

- `search.rs::resolve_visible_repos` does **not** derive from
  `build_visibility_clause_for`; it builds its own UNION whose visibility arm
  read `r.is_public = true`. Left alone, search would have returned nothing for
  an `internal` repository every authenticated caller could read and download
  directly. Both arms fixed and pinned by a DB test.
- `classify_access_scope` (CVE blast radius, #2386) had only three states, so an
  `internal` repository reported as `restricted_acl` — telling an operator
  triaging a CVE that a vulnerable artifact reached a few grantees when the
  whole instance could pull it. Added a fourth `internal` state and routed the
  accessible-users enumeration's short-circuit through
  `allows_authenticated_read()`, so it reports `exposure: everyone` instead of
  enumerating the entire user table.

### Migration ordinal

This branch originally added `212_repository_visibility.sql`. Upstream landed
`212_jupyter_format.sql` while it was open, so merging `main` produced two
different migrations claiming version 212 — the VersionMismatch class the
migration-ledger guardrail exists to catch (#2689). Renumbered to **217**, the
next free ordinal after upstream's 216. `scripts/ci/check-migration-ledger.sh`
passes and a fresh database applies the whole chain in order.

## Test Checklist
- [x] Unit tests added/updated — all six (visibility × has_auth) combinations on
      `should_allow_repo_access`; `public_read_satisfies_acl` never satisfied for
      write/delete in any state; `internal` confers no write/delete/repo-admin on
      the native path; `PublicOnly` never matches `internal`; the coercion on
      both create and update; the contradictory-input 400; and the Terraform
      regression (a legacy `is_public: false` full-object update over an
      `internal` repository leaves it `internal`, idempotently).
- [x] Tests added for the review round — five, all against a real database and
      the real router, each pinning both halves of the decision so a
      widening-by-accident fails them: the virtual fan-out vs. a direct fetch on
      an `internal` member (and `private` still refused through the parent); the
      SBOM gate vs. the artifact bytes it describes (and `private` still 404);
      the VS Code gallery refusing `internal` and naming it; and the two
      `repo_visibility_middleware` arms end to end.
- [x] Integration tests added/updated — `cache_invalidation_fanout_tests`
      (13/13 pass; it is on the CI Tier 2 list, so the modification is actually
      exercised).
- [x] Manually tested locally — migration and both trigger directions verified
      against live PostgreSQL 16.15, 18.4 and 18.6: backfill exact, `NOTIFY`
      fires on a visibility-only change, the legacy `is_public = false` write is
      a no-op over `internal`, contradictory INSERTs refused, and the CHECK
      invariant holds with 0 violations.
- [x] No regressions in existing tests — **with a caveat worth reading.**
      Tier 1, re-run after merging current `main` and with `--bins` as
      CLAUDE.md now requires (#3494): 16289 tests, 43 failed. Every one is an
      environment artefact of the machine used — a DNS/route interception makes
      `wiremock` upstreams resolve to `198.18.0.1`, so remote-proxy tests get
      502s — and each was reproduced on clean `main`: 39 matched the pre-merge
      baseline and the 4 new ones
      (`pypi::legacy_json_xmlrpc_route_tests::*`, from upstream's own #3788)
      were run against clean `main` and fail there identically. Before the
      merge the branch failed 40 against `main`'s 39, the extra being
      `oci_v2::…::completion_with_nonempty_final_put_body_concatenates_parts`,
      which passes in isolation, passes with its whole 55-test module at 8
      threads, and passed again post-merge — load-sensitive flake, and its
      assertion is about concatenated blob bytes, unrelated to visibility.
      `cargo fmt --check` and `cargo clippy --workspace --all-targets -D
      warnings` are both clean on the merged tree.
      Tier 2, in the shape CI runs it (the 15 named suites, `--ignored
      --test-threads=1`): **14 of 15 pass**; `sso_oidc_login_callback_e2e_tests`
      fails 13/17 on this branch AND on clean `main` — same mock-IdP
      unreachability, not caused by this change.

      **Review round 2, verified separately and reported exactly as run.**
      `cargo fmt --check` and `cargo clippy --workspace --all-targets` are clean.
      The five new tests were run against a live PostgreSQL 16 with
      `AK_TESTS_REQUIRE_DB=1`, so a missing database would have failed them
      rather than skipping them (#2924): **5 passed**. Each was also confirmed to
      be load-bearing by reverting the predicate it guards — restoring
      `member.is_public` in the fan-out and `allows_anonymous_read` in the SBOM
      gate makes exactly those two tests fail on exactly the intended
      assertions, and the fixed tree turns them green again.

      `cargo nextest run --workspace --lib --bins --test-threads 8` on that
      database: **16294 tests, 25 failed.** All 25 are the DNS/route
      interception described above, not this change: 16 `vscode` gallery/OpenVSX
      and 5 `pypi` remote-proxy tests time out after 60-420 s and assert `502`
      against a `wiremock` upstream the machine cannot route to (inspected one
      in full: a 502 after a 60 s upstream timeout, so the request passed the
      visibility gate and failed at the fetch — a gate refusal would be a 403 at
      ~0 s), plus the four that also fail without a database
      (`auth_config_service`, `image_scanner` ×2 — two of which assert on the
      literal `198.18.0.1` the interception returns — and a `proxy_service`
      500 ms timing bound). They were **not** re-run against the pre-change
      branch tip in this round, so "unrelated" rests on the failure mode rather
      than on a fresh baseline diff.

      The coverage (>= 70% on changed lines) and jscpd duplication gates were
      NOT run locally — neither tool is set up on the machine used, so CI is the
      first place they are measured.

## API Changes
- [x] New endpoints have `#[utoipa::path]` annotations — N/A, no new endpoints;
      `routes.rs` is untouched. The change is additive fields and converted
      predicates on existing routes.
- [x] Request/response types have `#[derive(ToSchema)]` — `RepositoryVisibility`
      derives `utoipa::ToSchema` and is registered in the `repositories`
      module's `components(schemas(...))`; `RepositoryResponse` carries
      `visibility`.
- [ ] Migration is reversible — **no.** 217 is additive and `is_public` remains
      authoritative for every legacy reader, so reverting the *application*
      leaves a consistent database: `internal` repositories read back as
      `is_public = false`, i.e. private. That is a **narrowing** on rollback,
      never a widening, which is the correct direction to fail. There is no down
      migration; dropping the column would be the manual step.
- [ ] N/A - no API changes

## Security review

This change edits the authorization read path, so it was audited against the
design's decision-site table site by site. **No critical or high findings.** The
three easy-to-get-wrong sites (`require_repo_write_access`,
`member_passes_token_scope`, `public_read_satisfies_acl`) all treat `internal`
as `private`; `internal` never reaches a write (the shortcut sits inside the
`else` arm of the write branch) and never bypasses a repo-scoped token. No SQL
injection via `format!` in query contexts, no new endpoints, no hardcoded
secrets. `cargo audit` was NOT run — it is not installed on the machine used, so
no dependency-advisory result is claimed either way.

Open items recorded rather than fixed here:

- The two uniform-enforcement gaps this section previously recorded —
  `proxy_helpers.rs` and `sbom.rs` gating on `member.is_public` — are **fixed in
  this PR** (review findings F1 and F4), each with a test that also pins the
  `private` half so the fix cannot widen. `conda.rs` is not converted and does
  not need to be: its gate is anonymous-vs-authenticated, which already gives
  `internal` the right answer.
- Several `is_public` reads are deliberately left alone — the tenant/config
  gates in `repo_tokens.rs`, `promotion.rs` and `curation.rs` require `internal`
  to behave as `private`. Listed in the follow-up so a later sweep does not
  "fix" them into a widening.
- Migration 217 is blocking DDL (`ADD COLUMN` + full-table `UPDATE` +
  `SET NOT NULL` + a non-`CONCURRENTLY` index + `ADD CONSTRAINT` in one
  transaction). Operational, not security, but worth knowing before running it
  on a large instance.

### Follow-ups (recorded here; not filed as issues)

Deliberately left open, each self-contained:

1. **Regenerate the progenitor SDK.** `xtask` reads the spec from a sibling
   `artifact-keeper-api` repository holding the committed `openapi.json` export,
   which was not available. `RepositoryVisibility` already derives `ToSchema` and
   is registered, so the spec picks the field up as soon as it can be
   regenerated. Until then `@artifact-keeper/sdk` has no `visibility` field and
   the web client carries it through defensive casts.
2. **Run the two internal-visibility e2e specs.** Written, registered in
   `playwright.config.ts`, discovered by `--list` (20 tests), never executed —
   they need a deployed instance carrying migration 217.
3. **Add a browser-level e2e for create -> save -> list marker.** Both new specs
   use only the `request` fixture, so nothing drives the UI. That is precisely
   why the API client dropping `visibility` in both directions survived a full
   green unit suite.
4. **`repo-detail-content.tsx` and the blast-radius page still read
   `is_public`,** so both label an `internal` repository "Private". The
   blast-radius one is security-facing: the backend classifier now has a fourth
   `internal` state exactly so a CVE triage is not told the artifact reached a
   handful of grantees. `src/types/search.ts` and `tree.ts` likewise carry only
   the boolean.
5. **The web duplication gate does not analyse `.tsx` at all.** With ci.yml's
   own command, jscpd v4 measured only the one changed `.ts` file and silently
   skipped all four components; `--format tsx`/`jsx` produce no report. Same
   class as backend #3495, which was fixed there by failing when a changed file
   is missing from the report.
6. **A note for whoever next sweeps `is_public` out of the backend.** The
   virtual fan-out and the SBOM/CVE gate are converted here (F1, F4), and
   `conda.rs:1070` needs no conversion. What remains must NOT be converted: the
   tenant/config gates in `repo_tokens.rs`, `promotion.rs` and `curation.rs`
   require `internal` to behave as `private`, and `virtual_has_private_member`
   requires `internal` to count as not-public so a caller-dependent aggregate
   stays out of the shared cache. A blanket `visibility != 'private'`
   substitution at any of those sites is a widening.
7. **Port `docs/repository-visibility.md` to the public docs site,** which is
   built from a separate repository.


brandonrc:
## Review — changes requested

Thank you for this, and for the care that clearly went into it. The headline is good news: **I found no exposure.** Every cell of the principal × visibility matrix behaves as intended, every site that was not converted fails *closed* rather than open, and the two decisive cells are right — `internal` is not exempted from the repository-scoped token ceiling (`auth.rs:2487`), and it never satisfies a mutation (`auth.rs:2555`, `repositories.rs:136`). Both have negative tests. The migration is exactly access-preserving; I applied it on a fresh database and ran a twelve-case trigger matrix against it, including the Terraform-shaped `is_public: false` no-op and the `public → is_public=false` narrowing. The full suite passes, 16240 of 16240.

Three things block, and one of them needs your judgement rather than a code change.

**The two-state check that received a third value — virtual byte fan-out.** `proxy_helpers.rs:3483` still reads `if member.is_public { return true; }`. The grant half (`repository_service.rs:683-697`) and the scope half (`repositories.rs:401-415`) were both converted, but this third half was not. The result is an incoherence rather than a leak: an authenticated caller with no grant can **list** an internal member's artifacts through the virtual parent (`repositories.rs:5290`) but is refused their **bytes**, while reading the member directly works. `member.visibility.allows_authenticated_read()` fixes it; the anonymous arm already returned earlier at `:3447`.

**The CHANGELOG advertises a fix the diff does not contain.** The second `### Added` bullet says anonymous `docker pull` is now refused on non-anonymously-readable repositories when guest access is disabled. Nothing in the diff touches the guest-access allowlist — `api/routes.rs:184-191` is unchanged, and `grep guest_access` over the diff hits only the coercion helpers. Also worth knowing: `allows_anonymous_read()` is currently equivalent to `is_public`, because migration 217's `CHECK (is_public = (visibility = 'public'))` holds. Either ship that fix or drop the bullet; as written it would close an issue that stays open.

**The one for you, not for the code.** Guest-access coercion (`repositories.rs:76-107`) now lands `internal` where it used to land `private`, and `effective_visibility()` (`:1010`) maps a legacy `is_public: true` to public *before* the coercion runs. So the Terraform provider and older SDKs, which cannot express `internal` and cannot opt out of it, will start creating **org-readable** repositories on instances that disabled guest access precisely to avoid that. The CHANGELOG flags the change as breaking, but its upgrade note covers only the backfill, not future creates. I would rather that were a deliberate decision than a side effect.

**Should-fix:** the SBOM and CVE read gate is still two-state (`sbom.rs:1864`, resolvers at `:1898`, `:1925`, `:1945`), so an internal repository's dependency inventory and CVE history need a grant while the bytes they describe do not (`artifacts.rs:108` was widened).

**Nits:** `vscode.rs:329` calls an internal repository "private" and offers to "make it public"; `docs/repository-visibility.md:48-56` claims a uniformity that the two findings above contradict.

**The test gap worth closing while you are in here.** The PR ships no router-level fixture for an internal repository, so `repo_visibility_middleware`'s two new internal branches (`auth.rs:2600`, `auth.rs:2739`) are covered only through pure predicates. A one-token change there would widen or narrow `internal` with the suite still green. That is the one place I would want an end-to-end test rather than a unit test.

brandonrc (changes requested):
Recording my review formally — the detailed findings are in [this comment](https://github.com/artifact-keeper/artifact-keeper/pull/3813#issuecomment-5642139792), which I posted as a plain comment by mistake, so it never registered as a blocking review.

Nothing has changed in the substance, and the headline is still the good news: I found no exposure. The full principal × visibility matrix behaves correctly, 16240/16240 tests pass, and migration 217 is access-preserving with its `CHECK (is_public = (visibility = 'public'))`. What blocks the merge is F1 (virtual byte fan-out still two-state at `proxy_helpers.rs:3483`), F2 (the CHANGELOG advertises a `/v2` guest-access fix that is not in the diff), F3 (guest-access coercion now lands `internal` where it landed `private`, which needs product sign-off rather than a code change), F4 (the SBOM/CVE gate still two-state at `sbom.rs:1864`), and the router-level test gap for `auth.rs:2600`/`:2739`.

Apologies for the noise — same review, just in the form that actually blocks the merge.

mymarche:
Thanks — this is a careful read, and F1, F2 and F4 are all correct. F1 and F4 I'll fix, along with the `vscode.rs` and docs nits and the router-level fixture for `auth.rs:2600`/`:2739`. Two of your points I want to answer rather than patch.

**F3 — reverting to the previous behaviour.** You're right that this is a product decision and not a code detail, and on reflection it doesn't belong in this PR at all: widening what legacy clients create is a separate decision from adding the state. So `coerce_visibility_for_create`/`_for_update` go back to coercing a requested `public` to `private` when guest access is disabled, exactly as before. This PR then changes no existing behaviour — it only adds a state that has to be asked for explicitly. Note that `visibility: "internal"` still passes through untouched with guests disabled, so the feature remains usable on exactly the instances it was built for; only the legacy boolean path is unaffected by it.

If we do want a `public` request to land as `internal` under a disabled guest-access flag, I'd rather that be its own issue with its own upgrade note, and I'm fine either way on the outcome.

One thing I'd flag while we're here, independent of which way that lands: **the silent coercion itself is the operational problem.** Whichever state it resolves to, an operator who asks for `public` gets something else back with only a server-side `tracing::warn!` to say so. The state is in the response body, but nothing marks it as a substitution — a Terraform provider sending `is_public = true` reads back `false`, records drift, and proposes the same change on every subsequent plan, with no explanation of which side is wrong. I'd argue the request should simply be rejected with a 400 and let the client choose. That's a breaking API change, so it's filed on its own as #3855 rather than resolved here.

**F2 — dropping the bullet.** Verified, and you're right on both counts. `guest_access.rs` is untouched by this diff, `/v2` and `/v2/*` are still allowlisted wholesale, and every anonymous gate I converted reduces to the same decision it made before, because migration 217's `CHECK (is_public = (visibility = 'public'))` makes `allows_anonymous_read()` and `is_public` the same predicate. The only widenings are on authenticated paths (`oci_read_permitted`, the blob-mount source check at `oci_v2.rs:5906`), both behind a `claims` binding. So this PR does not change anonymous `/v2` behaviour in either direction, and the bullet is claiming a fix that isn't here. Worth adding that its headline was already true before this PR — a repository that is not anonymously readable was refused anonymously either way; the real bug is in the body.

That bug is real and is now #3854: an anonymous pull of a *public* repository still succeeds with `AK_GUEST_ACCESS_ENABLED=false`, because `/v2/*` bypasses the guard entirely. Fixing it means narrowing the allowlist to the challenge paths, giving the guard an OCI-shaped 401 envelope (`oci_errors.rs:15` notes it deliberately isn't routed through `oci_denied_response` today), and replacing `guard_allows_oci_v2_subpath_when_disabled`, which currently pins the buggy behaviour as an expectation. That's a change to the guest-access allowlist rather than to the visibility axis, so it stays out of this PR. Bullet dropped here.

Pushing the F1/F4/nits/test work and the F3 revert shortly.

brandonrc:
Second-round audit against current main (111 commits after your last push). Round one is fully addressed, and the design holds up: the existence oracle survives in both middleware branches and `/v2`, scoped tokens are still refused, and every surface I could find (listing, virtual members, search in SQL and OpenSearch, OCI catalog, SBOM, tree, conda token URLs, plus the environments, image-builds and package-analysis surfaces added on main) treats `internal` consistently. Three things before this can merge:

1. **Authorization widening (new).** In `check_artifact_visibility` (`artifacts.rs` ~110) the condition `!visibility.allows_authenticated_read() && !ext.is_admin` skips the grant check for every action, not only reads. For the three label endpoints (`artifact_labels.rs` set ~198, add ~244, delete ~281) that is the only repository check besides the `write:artifacts` scope, which interactive sessions always pass. So any signed-in user with no grant can rewrite or delete labels on an internal repository's artifacts, and label changes drive sync-policy re-evaluation and replication. That breaks the PR's own rule that `internal` is `private` for everything but reads. Use `authenticated_read_satisfies_acl(visibility, action)` and add a DB test that a user with no grant is refused on label writes. Same pattern at `approval.rs` ~398 (any signed-in user can file approval requests from an internal repo). Public repositories have the same label gap on main today; I am filing that separately.
2. **Migration 217 collides** with main's `217_lifecycle_policy_exclusions.sql`; `check-duplicate-migrations.sh` fails on the merged tree and the backend would abort at startup. Please renumber to **235** (232 is #4032, 233 is #3943, 234 is #3656, all in flight). Content-wise the renumber is safe: nothing in 218 to 231 redefines the cache-invalidation trigger this replaces.
3. **The merged tree does not build its tests, and the CHANGELOG lands in the wrong release.** Only one textual conflict (`repo_to_response`), but `image_builds.rs` ~749 lacks the new `visibility` field and `oci_v2.rs` ~30039 still sets the removed `is_public`. The three bullets end up under `[1.10.0]` and must move to `[Unreleased]`, and the upgrade note's SQL tells operators to "promote" public repositories to internal, which narrows access.

Also: #3812 should read "Part of" since the coercion retarget was descoped, and the body still cites #3854 as open (fixed by #3872). Happy to take the rebase and renumber myself if you would rather focus on item 1; say which.

mymarche:
Thanks for the second round, @brandonrc.

Items 2 and 3 are pushed: the migration is renumbered to `235_repository_visibility.sql`, the `image_builds.rs` / `oci_v2.rs` fixtures carry the `visibility` field, the CHANGELOG bullets moved under `[Unreleased]`, and the upgrade note no longer suggests promoting public repositories to internal. The body now says "Part of #3812" and points #3854 at #3872.

Since the body no longer carries a closing keyword, the linked-issue check fails. Could you apply `no-issue-required` here?

On item 1: I'll work on it, but it looks broader than the label and approval call sites. Making `check_artifact_visibility` action-aware changes behaviour for other callers of the same gate as well. I'd like to dig deeper into how grants are issued and checked across these paths before proposing a fix, so I'll handle it separately rather than hold this PR on it.

PR #3814, opened 2026-09-10
fix(maven): make concurrent snapshot uploads idempotent (#3587)

### LINKED ISSUE #3587: Possible duplicate artifact creation under upload pressure causes UNIQUE constraint violation
### Description

**Description**

During a period of high Maven upload activity, Artifact Keeper backend started returning:

`artifact_keeper_backend::error: Request error error=Database error: error returned from database: duplicate key value violates unique constraint "artifacts_repository_id_path_key" code="DATABASE_ERROR"`

and the client received: 500 Internal Server Error

The failures affected multiple Maven SNAPSHOT uploads across different artifacts.

### Environment Information

- Deployment type: Docker
- Operating System: Linux
- Architecture: 64x
- Database: postgresql 16
- Reverse Proxy: caddy
- Network setup: 

### Backend Version

1.8.0

### Web UI Version

1.8.0

### Package Format

Maven / Gradle

### Expected Behavior

Concurrent Maven SNAPSHOT uploads should not result in a database uniqueness violation being surfaced as: 500 Internal Server Error

### Observed Behavior

Under a period of elevated upload concurrency, multiple uploads failed with:

`duplicate key value violates unique constraint "artifacts_repository_id_path_key"`

### Steps to Reproduce

I do not currently have a deterministic reproducer.

The only common factor observed was a period of high concurrent Maven upload activity.

### Logs

```shell
INFO http_request{method=PUT uri=/maven/maven-snapshots/my-package/my-item/1.0.0-SNAPSHOT/my-item-1.0.0-20260827.132833-10-debug-amd64-windows.tar.gz correlation_id=78dcd836-3153-4dfa-b857-7722be775d89}: artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=78dcd836-3153-4dfa-b857-7722be775d89 status=201`

`ERROR http_request{method=PUT uri=/maven/maven-snapshots/my-package/my-item/1.0.0-SNAPSHOT/my-item-1.0.0-20260827.132833-10-debug-amd64-windows.tar.gz correlation_id=621f9d9a-6921-4e74-a252-45e6f0583423}: artifact_keeper_backend::error: Request error error=Database error: error returned from database: duplicate key value violates unique constraint "artifacts_repository_id_path_key" code="DATABASE_ERROR"
```

### Additional Context

One notable aspect is that the issue only appeared while the instance was under significant upload pressure. Once concurrent upload activity decreased, the problem stopped occurring without any intervention.

### Pre-submission Checklist

irizzant:
## Summary

Concurrent first publishers of the same Maven `-SNAPSHOT` path could both pass the preflight duplicate check and then race into the native Maven handler's plain `INSERT`. One request lost on `artifacts_repository_id_path_key` and surfaced a 500 Database Error.

## Fix

- SNAPSHOT coordinates use `ON CONFLICT (repository_id, path) DO UPDATE`, so concurrent mutable uploads converge without a uniqueness failure.
- The stale preflight SNAPSHOT `DELETE` is removed. The artifact upsert and dependent `artifact_metadata` write now commit in one transaction, closing the foreign-key race and preserving labels, scan history, download statistics, and quality-check records across a republish.
- Release coordinates use `ON CONFLICT ... DO NOTHING`; the losing concurrent publisher receives the normal `409 Artifact already exists` response instead of a database 500.
- Existing release immutability and SNAPSHOT re-upload semantics are preserved.

## Test

The DB-backed regression uses a multi-thread Tokio runtime and eight spawned Maven SNAPSHOT PUTs to the same path. It asserts that no request returns a 5xx and that the uploads converge to exactly one artifact row.

## Verification

- `SQLX_OFFLINE=true cargo check -p artifact-keeper-backend --lib`
- `cargo test -p artifact-keeper-backend --lib concurrent_snapshot_uploads_upsert_without_unique_error_3587 -- --nocapture`

The local environment has no `DATABASE_URL`, so the targeted test compiled and was selected successfully, while its DB fixture returned early. The upstream CI database run and Brandon's Postgres stress probe provide the real concurrent execution gate.

The pre-existing mismatch between the Maven upload mutability predicate and `classify_maven` is tracked separately in #3839.

Closes #3587


brandonrc (changes requested):
## Review — changes requested

The mechanism you reached for is the right one and matches how the rest of the codebase does this. But it does not close #3587: under the same 8-way same-path SNAPSHOT race the request still returns `500 Database error` — only the constraint name in the log changes, from `artifacts_repository_id_path_key` to `artifact_metadata_artifact_id_fkey`. I reproduced that against a real Postgres 16 rather than inferring it.

### 1. The 500 moved one statement to the right (blocker)

The preflight hard delete at `backend/src/api/handlers/maven.rs:3024` is untouched by this PR:

```sql
DELETE FROM artifacts WHERE repository_id = $1 AND path = $2
```

and `maven.rs:3202` then inserts the dependent row. Nothing in `upload` runs in a transaction, so racer B's `DELETE` cascades away the `artifacts` row that racer A's upsert (`maven.rs:3135-3186`) just returned, in the window between A's upsert and A's `artifact_metadata` insert. A then gets:

```
insert or update on table "artifact_metadata"
violates foreign key constraint "artifact_metadata_artifact_id_fkey"
```

→ `map_db_err` → 500. Measured, 8 `tokio::spawn`ed PUTs to one snapshot coordinate on a multi-thread runtime against a real database:

| tree | runs with a 5xx |
|---|---|
| `main` (plain INSERT) | **6/6** — e.g. `[500,201,201,500,500,500,500,500]` |
| this PR as submitted | **3/19** — e.g. `[201,500,500,201,201,201,201,201]` |
| this PR + `maven.rs:3022-3030` removed | **0/14** |

So the unique-index serialisation is real and correct *for the `artifacts` row* — it just isn't the only row the request writes. This is the "racier in a smaller window" outcome rather than idempotency.

**The fix, which I verified:** with `ON CONFLICT (repository_id, path) DO UPDATE` in place, the snapshot branch's hard `DELETE` is redundant — the upsert already settles the existing-row case — and it is precisely what opens the FK window. Removing those nine lines gave 0/14 failures and still passed the legitimate-republish probe. The better version, matching `upload.rs:691-712` and `artifact_service.rs:818-845`, is one `tx` spanning the artifact upsert *and* the `artifact_metadata` insert. Either is acceptable to me; the transaction is the one I'd prefer.

One behaviour change to weigh explicitly if you drop the DELETE: a snapshot republish will no longer cascade away `artifact_labels`, `cve_history`, `download_statistics` and `quality_check_*` for that coordinate. I think that's an improvement, but it should be a decision rather than a side effect, and it belongs in the CHANGELOG bullet.

### 2. The regression test passes on main most of the time (blocker)

`maven.rs:7110-7166` uses a bare `#[tokio::test]`, which is the **current-thread** runtime, with `tokio::join!` — so the two uploads interleave only at their own await points, on one thread. Against main's plain `INSERT`, the bug this test exists to pin: **11 runs, 9 passed, 2 failed.** Roughly 18% detection. It would not have caught #3587, and in CI it's a flake in the worst direction — green on a broken tree.

The same scenario as `#[tokio::test(flavor = "multi_thread", worker_threads = 8)]` with `tokio::spawn` fails 6/6 on main and passes on your branch. Please use that shape, and assert `!status.is_server_error()` rather than `== 201`, so a future 409 or 503 regression is caught too.

### 3. Why it's BLOCKED — the PR's own test doesn't compile

Not a stale cancelled run, which is a pattern we do have here. `Check Rust`, `Backend Unit Tests` and `Code Coverage` all failed on the single run 34490656248 with exit 101; the coverage annotation is the repo's "Coverage BUILD failed" message, downstream of the same error. Locally:

```
error[E0599]: no associated function or constant named `from_static` found for struct `axum::body::Body`
  --> backend/src/api/handlers/maven.rs:7132:31
  --> backend/src/api/handlers/maven.rs:7139:31
```

axum is pinned to 0.7 (`Cargo.toml:21`); `Body::from_static` doesn't exist there, and those are its only two uses in the tree. `Body::from(&b"snapshot bytes A"[..])` works — that's what I used to run everything above. This traces directly to "compilation and tests are intentionally delegated to the upstream CI to avoid heavy local builds" in the PR body. I understand the motivation, but for a concurrency fix the local run is the part that matters most: CI told you it didn't compile, it couldn't tell you the test doesn't detect the bug.

### 4. Diverges from the two house implementations (medium)

The SQL is byte-for-byte the same column set as `artifact_service.rs:828` and `upload.rs:696`, which is good. What differs is the transaction — both of those run the upsert *and* their dependent writes inside one `tx`, while `maven.rs:3170-3185` runs on `&state.db` in autocommit. That divergence is finding 1. Those two call sites are the pattern to copy.

### 5. Mutability predicate diverges from the classifier (low, pre-existing)

`maven.rs:3135` keys mutability on `coords.version.contains("SNAPSHOT")`, but `cache_classifier.rs:298-330` deliberately classifies a *resolved unique* snapshot leaf as `Immutable`, and uses component-wise `ends_with("-snapshot")` (`:340`) rather than `contains`, so `1.0-SNAPSHOT-rc1` is not a snapshot directory. Probed: `…/6.0-SNAPSHOT/probe-6.0-20260827.132833-10.jar` uploaded twice with different bytes gives `201`, `201` and silently replaces the row, while `classify()` says `Immutable`.

Pre-existing — the preflight DELETE already did this — so not a blocker. But this PR promotes "SNAPSHOT ⇒ mutable" into the authoritative DB statement and into the CHANGELOG, and #3587's own log line is exactly such a timestamped path. Either reuse `classify_maven` here or file a follow-up; I'm fine with the follow-up.

---

### What I checked that's fine

**Snapshot semantics pass in both directions**, which was my main worry — an idempotency fix that turns a legitimate republish into a no-op would be a worse bug than the one it fixes:

| probe | result |
|---|---|
| sequential snapshot republish, **different** bytes | `201`, `201`, row carries the new size — republish still lands |
| sequential release re-publish, different bytes | `201` then **409**, stored bytes unchanged |
| two concurrent release first-publishers | `[201, 409]`, one row, no 5xx — your `DO NOTHING` → 409 conversion works as described |

`maven-metadata.xml` turns out to be a red herring: it's a row-less put (`maven.rs:2962`) that returns 201 before coordinate parsing, so it never reaches either branch. Same for checksum sidecars.

**Bookkeeping passes.** The bullet is at `CHANGELOG.md:40` under `## [Unreleased]` → `### Fixed` with `(#3587)` first, and the `awk` heading check shows no misfiling. `cargo fmt --all -- --check` is clean.

One request on the body: until finding 1 is fixed, `Closes #3587` should read `Part of #3587`, and per the preflight rule a `Part of` PR needs its own number appended to the bullet's parenthetical.

Fix 1, 2 and 3 and I'll re-run the stress probes against your branch myself before merging — 5 is fine as a follow-up, and 4 falls out of 1 if you take the transaction route.

irizzant:
@brandonrc  Thanks for the detailed reproduction. I pushed the requested follow-up in `c25e8f25`:

- removed the preflight SNAPSHOT hard delete;
- wrapped the artifact upsert and `artifact_metadata` upsert in one transaction;
- kept the quarantine/package/sync follow-ups after commit;
- changed the regression to an 8-worker multi-thread Tokio test with eight spawned same-path uploads and a no-5xx assertion;
- replaced the unsupported `Body::from_static` calls;
- documented that SNAPSHOT republish now preserves dependent labels, scan history, download statistics, and quality-check records;
- rebased the PR onto current `main`.

Local `cargo fmt --all -- --check` and `SQLX_OFFLINE=true cargo check -p artifact-keeper-backend --lib` pass. The targeted test compiles and is selected correctly, but this machine has no `DATABASE_URL`, so its fixture returns early; the CI database run and your Postgres stress probe remain the meaningful concurrency verification.

The classifier mismatch is already tracked in #3839, so I left that pre-existing behavior out of this focused fix.

irizzant:
@brandonrc  CI update: the new Postgres-backed regression ran in the unit-test job and passed:

```
PASS api::handlers::maven::tests::concurrent_snapshot_uploads_upsert_without_unique_error_3587
```

`Check Rust`, coverage, integration tests, image build, shell tests, and security audit are all green. The unit-test job failed later in the unrelated scheduler test `storage_gc_tick_follow_on_recomputes_stats_and_keeps_blob_gc_dry_run_db`, on a `repository_storage_stats_repository_id_fkey` fixture failure. I tried to rerun only failed jobs, but GitHub requires repository admin permission. Could you please rerun the failed unit-test job?

irizzant:
@brandonrc ?

brandonrc:
> @brandonrc ?

Sorry about the delay. Work travel had me busy last week.

brandonrc:
@irizzant  looking at this today

brandonrc:
@irizzant — apologies for the sequencing here: #3942 landed with its own #3587 fix while your rerun request was sitting unanswered. The unit-test failure you hit was #3933, unrelated to this PR, and it has since been fixed on `main`.

Your 09-13 revision addressed everything I asked for. Two pieces of it are still things `main` is missing — the transaction around the upsert + metadata write, and the multi-thread runtime for the regression — so I've rebased the branch down to exactly those (`2be66971`), with your authorship kept on the commit. The one substantive change from your version: `main` now uses the classifier-owned `republishable` predicate from #3839 in place of `contains("SNAPSHOT")`, so the rebase adopts that. I updated the title and description to match.

Verified 21/21 against a fresh Postgres 16 with proof the fixture engaged the database each run. If this looks right to you, I'll merge it for 1.11.0. Thanks for the work and the patience.

brandonrc:

The linked-issue check fails because #3587 is already closed; adding the `no-issue-required` label rather than asking for a fresh issue.

Issue #3660, opened 2026-09-02
Packages catalog rows are never deleted: deleting an artifact leaves a permanent ghost entry on the Packages page


brandonrc:
## Summary

Nothing in the backend ever removes a `packages` / `package_versions` row. On `main` (`e2be9fef`+), `grep -rnE 'DELETE FROM packages|DELETE FROM package_versions' backend/src` has **zero** runtime hits. `list_packages` does not join `artifacts`, so once an artifact is registered in the catalog it is listed forever, regardless of whether its bytes still exist.

Reproduced during the #3534 review with Helm on a throwaway database: `cm-push` → `helm delete` (200) → `packages=1 package_versions=1 live_artifacts=0`. The chart is gone from `index.yaml` and from Artifacts, `helm pull` 404s, and the Packages page still lists it.

Every format that writes the catalog shares this (npm, NuGet, Composer, Conan, Debian, Incus, Helm, and the migration importer); #3534 simply made it reachable for Helm.

## Ask

Either:
- delete the `package_versions` row (and the `packages` row when its last version goes) from the artifact delete path — `ArtifactService::delete` is the choke point for REST deletes, but the format-native delete routes (`delete_chart`, npm unpublish, …) need checking individually; or
- make the catalog read paths join live `artifacts` (`is_deleted = false`) so a ghost row is filtered rather than removed.

The second is cheaper and self-healing; the first keeps `total` counts honest without a join. Soft-delete/restore (#soft-delete work) needs to be considered either way — a restored artifact must reappear.

Refs: #3531, #3534.

Issue #3602, opened 2026-08-31
OCI: pulling through a Virtual repository caches no manifest at all, so the image appears in neither the artifact listing nor the packages catalog


brandonrc:
Found while fixing #3441 and deliberately not folded into it, because it is a different and larger gap: #3441 is about a cached image not reaching the packages catalog, and this is about nothing being cached in the first place.

### What happens

Measured on a build carrying the #3441 fix, one Virtual Docker repository with a Remote member:

| pull | `artifacts` | `oci_tags` | `packages` |
|---|---|---|---|
| direct through the **Remote** member | row written | row written | row written (#3441) |
| through the **Virtual** repository | none | none | none |

Both pulls return `200` and serve correct bytes. The direct pull of `library/redis:7.2-alpine` through the member produced a `docker-remote / library/redis / 7.2-alpine` catalog row; the same tag pulled through the Virtual repository produced no rows of any kind, in any of the three tables.

### Why

The Virtual branch of `handle_get_manifest` resolves the manifest through `resolve_virtual_manifest` and returns `build_oci_proxy_response(...)` directly. It never reaches `cache_manifest_reference_locally`, which is the function that writes the manifest body to storage and records the `oci_tags`, `manifest_blob_refs`/`oci_manifest_refs`, `artifacts` and (since #3441) `packages` rows. Blobs are unaffected — they stream through `ProxyService` and are cached — so a Virtual pull leaves cached layers with no manifest recorded against them.

### Why it was not fixed alongside #3441

The obvious shape is to have the Virtual branch delegate to `cache_manifest_reference_locally` under the resolving member (`oci_repo_info_from_member`), which is where a direct pull would have cached it and needs no new decision about which repository owns the bytes. That is plausibly correct and small, but it is a behaviour change to a path carrying a lot of invariants that are documented in-file and were each their own incident — blob-GC ref integrity (#1409), the proxy-cache storage-location split (#1278), and the Virtual inline scan gate (#3023) — plus quota and storage accounting for a repository type that currently stores nothing. It needs its own reproduction and review rather than riding along in a catalog fix.

### Impact

Anyone using the common topology of one Virtual repository fronting several Remotes sees no images in the Packages page and no manifests in the artifact listing, even after #3441. That topology is exactly what #3441 was reported against, so #3441 should not be treated as fully closing that report until this lands.

Issue #3601, opened 2026-08-31
OCI: a proxied multi-arch tag's packages-catalog size is a lower bound, not the image size


brandonrc:
Split out of #3441, which makes proxied image tags appear in the packages catalog. The row appears; its `size_bytes` is the part that is not yet meaningful for multi-arch images, and fixing it changes a shared pre-existing path rather than the code #3441 touches.

### Measured behaviour

On a build carrying the #3441 fix:

| case | recorded `size_bytes` |
|---|---|
| hosted push, single-arch image (37-byte config + 17-byte layer) | `54` — exact |
| proxied **image** manifest | `config.size + layers[].size` — exact |
| proxied **index** (multi-arch), `library/alpine:3.19` | `0` |

### Why

An image index carries no `config` or `layers` of its own, so `index_child_artifact_size_sum` supplies the number: the sum over the child manifests already recorded. On a **push** the children are uploaded first, and `upsert_manifest_artifact` records each child with its full image size, so the sum is meaningful. On a **proxy** the order is reversed — the index is fetched first and the children only afterwards — so on a first pull the sum is over nothing.

Two things then stop it from correcting itself:

1. An ordinary re-pull does not re-enter `cache_manifest_reference_locally` at all; a warm proxy cache serves the tag directly. The row is only recomputed once the cached tag has expired and the tag is re-fetched from upstream.
2. Even then the sum is wrong in kind, not just stale: the `artifacts` rows that `cache_manifest_reference_locally` writes for a proxied child manifest record the **manifest body length** (`content.len()`), not the image size. Summing them yields a few KB for an image that is tens of MB.

### What a fix looks like

Recording proxied child-manifest artifact rows with `manifest_total_size(content)` instead of the body length would make the sum mean what the push path means. That column is also read by the docker-tag grouping endpoint (`list_artifacts_grouped_by_docker_tag` and `fetch_index_child_sizes`), so the change needs to be evaluated against those consumers rather than made locally — which is why it is filed here instead of being smuggled into a fix about images not appearing at all.

A lower bound was preferred over an invented number in the meantime.

PR #3816, opened 2026-09-10
fix(packages): project the catalog from stored artifacts instead of per-handler calls

### LINKED ISSUE #3659: Packages catalog: several hosted format handlers never register uploads (same shape as #3531)
## Summary

#3531 / #3534 fixed Helm: the upload handler wrote the `artifacts` row but never called `PackageService::try_create_or_update_from_artifact`, so charts were visible under Artifacts and absent from the Packages page. Reviewing that fix turned up the same shape in other hosted-format handlers.

Verified by grep on `main` (`e2be9fef`+): handlers that contain an `INSERT INTO artifacts` and **zero** references to `try_create_or_update_from_artifact` / `PackageService`:

`alpine.rs`, `cargo.rs`, `chef.rs`, `cocoapods.rs`, `conda.rs`, `goproxy.rs`, `huggingface.rs`, `jetbrains.rs`, `protobuf.rs`, `pub_registry.rs`, `rpm.rs`, `sbt.rs`, `swift.rs`, `terraform.rs`, `vscode.rs` (plus admin/approval/promotion/quarantine/`repositories.rs`, whose inserts are mostly fixtures or non-upload paths).

Each needs a per-format check — some of those inserts are test fixtures, and some formats may legitimately not belong in the catalog — but the upload paths of at least rpm, cargo, alpine, conda and terraform were read during the #3534 review and do write `artifacts` rows for native publishes with no catalog registration.

## Ask

For each format that has a native publish path: register the package through the same `PackageService` write npm/NuGet/Composer/Incus/Helm use, keyed on the format's own coordinates (never the filename), with the format's description where the manifest carries one. Same fire-and-forget contract as the others (cannot fail the publish).

Note there is still **no backfill** for the catalog (`POST /api/v1/admin/reindex` is the OpenSearch reindex; no migration inserts `packages` rows), so each format's fix ships with the same upgrade note #3534 carried: artifacts published before the fix stay off the Packages page until re-published. A generic catalog backfill would make that note go away for all of them at once and is probably the better first step.

Refs: #3531, #3534, #2676 (the migration importer already indexes some of these formats, so a *migrated* artifact can appear where a *natively pushed* one does not).

garrettKelleyCyOne:
## Summary

Closes #3659: every hosted format that never registered itself now reaches the package catalog, and content stored before the upgrade can be backfilled.

**What this does not close**, so nothing is marked fixed by accident:

- **#3660 (catalog rows are never removed).** Unchanged, and the exposure grows: the ~20 formats this adds to the catalog inherit the ghost-on-delete behaviour every self-registering format already has, because nothing removes a row when an artifact is deleted and the reindex does not prune. The projection never registers a soft-deleted artifact, so it cannot resurrect one, but it cannot retire one either.
- **#3601, #3533.** OCI is exempt from the projection (`package_catalog::is_exempt_format`), so neither is touched.
- **#3602.** Not fixed. The Virtual publish path still writes no artifact row, so there is nothing to project. #3535, now on main, makes a Virtual repository's Packages tab list its *members'* packages, which can look like a fix for #3602 but is not one.

### The problem

`packages` / `package_versions`, which `GET /api/v1/packages` and the web UI's Packages page read *instead of* `artifacts`, is a hand-maintained index. A row exists only where a format handler remembered to call `PackageService::try_create_or_update_from_artifact` after its own `INSERT INTO artifacts`. Upstream has fixed that one format at a time (#1289, #1477, #1487, #1909, #2337, #3358, #3531), while about 20 hosted handlers still publish artifacts that can be pulled but never appear in the catalog. Registration also only runs at publish time, so each of those fixes shipped with the same upgrade note: content published before the fix stays off the page until it is pushed again.

### The change

- **`services::package_catalog::project`** maps `(format, artifact_metadata, artifacts row)` to catalog coordinates.
  - **Identity is `artifacts.name`.** Every handler stores the package's full identity there (Terraform `namespace/name/provider`, Swift `scope.name`, Composer `vendor/package`, PyPI's normalized name), and it is exactly what the self-registering handlers pass. The only exceptions are formats whose catalog is keyed on something else: Maven/Gradle `groupId:artifactId` (parsed from the path, as both Maven catalog writers do), SBT `org:module`, and Conan `name@user/channel`. The manifest's own `name` field is deliberately not used. For Terraform and Swift it is the bare name, which would collapse distinct packages onto the single `(repository_id, name)` row. For PyPI it is the display name, which would add a second row beside the handler's.
  - **Not packages:** checksums, signatures, regenerated indexes, and descriptors stored at the same name and version as their package (Helm `.prov`, Go `go.mod`, SBT `ivy.xml`, Protobuf `_labels`). Cataloguing a descriptor is not harmless noise. It upserts the same `package_versions` row, and that row's deterministic tiebreak then reports the descriptor's size for the package about half the time. Also excluded: OCI rows, and artifacts with no version anywhere.
- **Registration happens in `proxy_helpers::record_artifact_metadata`**, the chokepoint every publish already passes through and where the upload-time quarantine hold already lives. The 14 handlers that hand-rolled the `artifact_metadata` upsert now call the helper.
  - **Handler rows are shared, not overwritten.** The projection writes `packages.metadata` only where a row has none (`MetadataWrite::FillMissing`). It knows only the format, and a handler's row carries more, such as Conan user/channel or PyPI `requires_python`. Existing callers keep `Replace`.
  - **Only listable content is registered.** The rule is the one the format indexes already apply (Terraform's version list, the CocoaPods shard index): a rejected artifact is never registered, and a quarantined one only once its timed hold has lapsed. An upload held at publish time registers when it is released, either by an admin (`quarantine_service::transition`) or by the scan that clears the hold. A hold that lapses with no scan changes no row, so nothing fires for it; the reindex picks those up.
- **`POST /api/v1/admin/packages/reindex`** runs the identical projection over stored artifacts, bounded and resumable by cursor. Each page is a primary-key range scan from the cursor with a per-row metadata lookup (see Performance).
- **No change to the Packages listing any more.** An earlier revision also expanded a Virtual `repository_key` to its members in `list_packages`. #3535 landed its own member aggregation there, so after merging main this PR takes main's implementation instead of keeping a second one. Nested expansion is tracked in #3840.

Remote repositories keep the proxy's own catalog path (#2218, #3441, #3599).

### Performance

The read path is unchanged by the review fixes. The backfill query was rewritten. Here is a measurement on 60,000 artifacts across three repositories, each with a metadata row and random UUID ids, 500 per page, Postgres 18, generic plans (what a prepared statement settles on), with `artifact_metadata` rows read per page at three cursor positions:

| query | cursor at 1% | 50% | 99% |
|---|---|---|---|
| before: join, then `LIMIT` | 1,001 | 30,501 | 59,501 |
| after: `LIMIT` in subquery + `LATERAL` metadata lookup | 500 | 500 | 500 |
| before, filtered to one repository | 2,001 | 31,489 | 59,997 |
| after, filtered to one repository | 500 | 500 | 345 (last page) |

With a repository filter, a page still walks the primary key past other repositories' rows. For a repository holding a third of the artifacts, that is about 1,500 artifact rows per page, and it stays the same at any cursor.

Moving the `LIMIT` inside the subquery alone was not enough. With only a page of artifacts on the outer side, the planner hashed the whole of `artifact_metadata` on every page. That keeps each page flat, but it is still a full-table read per page. The `LATERAL ... LIMIT 1` can't be flattened back into that join, so each row does one probe of the unique `artifact_id` index. The id filter is spelled `id > $2` with a nil-UUID default rather than `($2 IS NULL OR id > $2)`, because the latter cannot be used as an index range.

DB-backed tests fail if the mechanism they cover is removed. Each was checked by reverting the mechanism and confirming the test fails on the expected assertion:

| test | fails when |
|---|---|
| `package_catalog::db_tests::a_publish_through_the_metadata_chokepoint_lands_a_catalog_row` | the call in `record_artifact_metadata` is removed (no row), or the manifest `name` is preferred (the two Terraform modules collapse onto `vpc`) |
| `…::a_backfill_registers_stored_artifacts_without_touching_handler_rows` | metadata is replaced (`requires_python` lost), or the manifest `name` is preferred (`Flask` row beside `flask`); pages of one also drive the cursor to completion |
| `…::deleted_and_unreleased_artifacts_stay_out_until_released` | the listability predicate is removed (rejected and held artifacts registered), or the release transition does not register (held upload never appears) |
| `…::a_go_module_reports_the_zip_size_not_the_go_mod` | the descriptor rule is removed (version reports 42 bytes, not 4096) |

The pure `package_catalog::tests` cover the projection per format. They include the Terraform/Swift identity, PyPI normalization, Maven path-first coordinates for the generic finalize path, SBT, Conan, and each exclusion.

## Test Checklist

- [x] Manually tested locally

`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo nextest run --workspace --lib --bins` pass against a migrated Postgres 18 (after merging main; the one exception, `health::test_probe_storage_health_filesystem_unwritable`, fails only because the local container runs as root) with `AK_TESTS_REQUIRE_DB=1`. The first revision was also deployed to a single-node cluster holding a restored copy of a real instance. There, a Helm repository with 2,033 chart artifacts and 1 catalog row backfilled to 178 packages / 2,031 versions, and a fresh publish in a format with no registration code appeared immediately.

## API Changes

- [x] Migration is reversible (if applicable): no migration; the backfill is an endpoint
- [ ] N/A - no API changes

CHANGELOG: one entry under `## [Unreleased]` → `### Fixed`, first reference #3659, with the upgrade note for the reindex.

One question for reviewers: **should the per-handler `PackageService` calls be deleted?** They now converge on the same rows. Removing them is a larger diff with no behavioural change, so it is left as a follow-up.

### DIFF
--- a/backend/src/api/handlers/admin.rs
+++ b/backend/src/api/handlers/admin.rs
@@ -43,6 +43,7 @@ pub fn router() -> Router<SharedState> {
         .route("/downloads/by-user/:user_id", get(list_downloads_by_user))
         .route("/cleanup", post(run_cleanup))
         .route("/reindex", post(trigger_reindex))
+        .route("/packages/reindex", post(reindex_packages))
         .route("/rescan-for-inventory", post(rescan_for_inventory))
         .route("/storage-backends", get(list_storage_backends))
         .route("/audit", get(list_audit_logs))
@@ -50,6 +51,112 @@ pub fn router() -> Router<SharedState> {
             "/proxy-scan-verdicts/:digest",
             get(get_proxy_scan_verdicts).delete(delete_proxy_scan_verdicts),
         )
+}
+
+// ---------------------------------------------------------------------------
+// Package-catalog backfill (#3659)
+// ---------------------------------------------------------------------------
+
+#[derive(Debug, Deserialize, ToSchema)]
+pub struct ReindexPackagesRequest {
+    /// Restrict the scan to one repository key. Omitted, every local
+    /// repository is scanned.
+    pub repository_key: Option<String>,
+    /// Resume token: the `next_cursor` returned by the previous call.
+    pub after: Option<Uuid>,
+    /// Artifacts to examine in this call. Defaults to 500, capped at 5000 so
+    /// one call cannot hold a connection for minutes on a large instance.
+    pub limit: Option<i64>,
+}
+
+#[derive(Debug, Serialize, ToSchema)]
+pub struct ReindexPackagesResponse {
+    pub scanned: i64,
+    pub registered: i64,
+    pub skipped: i64,
+    /// Pass back as `after` to continue; absent when the scan is complete.
+    #[schema(value_type = Option<String>)]
+    pub next_cursor: Option<Uuid>,
+}
+
+/// Rebuild the package catalog from artifacts already in the database.
+///
+/// `POST /api/v1/admin/reindex` rebuilds the OpenSearch index; nothing has ever
+/// rebuilt `packages` / `package_versions`, so an artifact published before its
+/// format learned to register itself stays off the Packages page permanently —
+/// the upgrade note every format fix has had to carry (#1289, #1477, #1487,
+/// #1909, #2337, #3358, #3531). This endpoint projects the stored
+/// `artifacts` + `artifact_metadata` rows through the same
+/// `package_catalog::project` the live write path uses, so a backfilled row and
+/// a freshly published one are identical.
+///
+/// Bounded and resumable: callers repeat the call with the returned
+/// `next_cursor` until it comes back absent. The upsert is idempotent, so
+/// re-running it — or running it while publishes are in flight — converges.
+#[utoipa::path(
+    post,
+    path = "/packages/reindex",
+    context_path = "/api/v1/admin",
+    tag = "admin",
+    request_body(content = ReindexPackagesRequest, description = "Optional; empty body scans everything"),
+    responses(
+        (status = 200, description = "Batch processed", body = ReindexPackagesResponse),
+        (status = 403, description = "Admin privileges required"),
+        (status = 404, description = "Repository not found"),
+    ),
+    security(("bearer_auth" = []))
+)]
+pub async fn reindex_packages(
+    State(state): State<SharedState>,
+    Extension(auth): Extension<AuthExtension>,
+    body: Option<Json<ReindexPackagesRequest>>,
+) -> Result<Json<ReindexPackagesResponse>> {
+    if !auth.is_admin {
+        return Err(AppError::Authorization(
+            "Admin privileges required".to_string(),
+        ));
+    }
+
+    let request = body.map(|Json(body)| body);
+
+    let repository_id = match request.as_ref().and_then(|r| r.repository_key.as_deref()) {
+        Some(key) => Some(
+            sqlx::query_scalar::<_, Uuid>("SELECT id FROM repositories WHERE key = $1")
+                .bind(key)
+                .fetch_optional(&state.db)
+                .await
+                .map_err(|e| AppError::Database(e.to_string()))?
+                .ok_or_else(|| AppError::NotFound(format!("Repository '{key}' not found")))?,
+        ),
+        None => None,
+    };
+
+    let limit = request
+        .as_ref()
+        .and_then(|r| r.limit)
+        .unwrap_or(500)
+        .clamp(1, 5000);
+    let after = request.as_ref().and_then(|r| r.after);
+
+    let report = crate::services::package_catalog::backfill(&state.db, repository_id, after, limit)
+        .await
+        .map_err(|e| AppError::Internal(e.to_string()))?;
+
+    tracing::info!(
+        actor_user_id = %auth.user_id,
+        actor_username = %auth.username,
+        scanned = report.scanned,
+        registered = report.registered,
+        skipped = report.skipped,
+        "admin.reindex_packages: package catalog backfill batch"
+    );
+
+    Ok(Json(ReindexPackagesResponse {
+        scanned: report.scanned,
+        registered: report.registered,
+        skipped: report.skipped,
+        next_cursor: report.next_cursor,
+    }))
 }
 
 // ---------------------------------------------------------------------------
@@ -1890,6 +1997,7 @@ pub async fn delete_proxy_scan_verdicts(
         run_cleanup,
         trigger_reindex,
         rescan_for_inventory,
+        reindex_packages,
         list_storage_backends,
         list_audit_logs,
         get_proxy_scan_verdicts,
@@ -1916,6 +2024,8 @@ pub async fn delete_proxy_scan_verdicts(
         ReindexResponse,
         RescanForInventoryRequest,
         RescanForInventoryResponse,
+        ReindexPackagesRequest,
+        ReindexPackagesResponse,
         AuditLogItem,
         AuditLogListResponse,
         ProxyScanVerdictItem,
--- a/backend/src/api/handlers/alpine.rs
+++ b/backend/src/api/handlers/alpine.rs
@@ -1588,24 +1588,13 @@ async fn store_apk(
         apk_info.as_ref(),
     );
 
-    let _ = sqlx::query!(
-        r#"
-        INSERT INTO artifact_metadata (artifact_id, format, metadata)
-        VALUES ($1, 'alpine', $2)
-        ON CONFLICT (artifact_id) DO UPDATE SET metadata = $2
-        "#,
+    proxy_helpers::record_artifact_metadata(
+        &state.db,
         artifact_id,
-        alpine_metadata,
-    )
-    .execute(&state.db)
-    .await;
-
-    // Update repository timestamp
-    let _ = sqlx::query!(
-        "UPDATE repositories SET updated_at = NOW() WHERE id = $1",
         repo.id,
+        "alpine",
+        &alpine_metadata,
     )
-    .execute(&state.db)
     .await;
 
     info!(
--- a/backend/src/api/handlers/cargo.rs
+++ b/backend/src/api/handlers/cargo.rs
@@ -773,23 +773,13 @@ async fn store_crate_artifact(
     crate::services::quarantine_service::apply_upload_hold_hosted(&state.db, repo.id, artifact_id)
         .await;
 
-    let _ = sqlx::query!(
-        r#"
-        INSERT INTO artifact_metadata (artifact_id, format, metadata)
-        VALUES ($1, 'cargo', $2)
-        ON CONFLICT (artifact_id) DO UPDATE SET metadata = $2
-        "#,
+    proxy_helpers::record_artifact_metadata(
+        &state.db,
         artifact_id,
-        cargo_metadata,
-    )
-    .execute(&state.db)
-    .await;
-
-    let _ = sqlx::query!(
-        "UPDATE repositories SET updated_at = NOW() WHERE id = $1",
         repo.id,
+        "cargo",
+        &cargo_metadata,
     )
-    .execute(&state.db)
     .await;
 
     Ok(())
--- a/backend/src/api/handlers/chef.rs
+++ b/backend/src/api/handlers/chef.rs
@@ -534,23 +534,13 @@ async fn upload_cookbook(
     crate::services::quarantine_service::apply_upload_hold_hosted(&state.db, repo.id, artifact_id)
         .await;
 
-    let _ = sqlx::query!(
-        r#"
-        INSERT INTO artifact_metadata (artifact_id, format, metadata)
-        VALUES ($1, 'chef', $2)
-        ON CONFLICT (artifact_id) DO UPDATE SET metadata = $2
-        "#,
+    proxy_helpers::record_artifact_metadata(
+        &state.db,
         artifact_id,
-        chef_metadata,
-    )
-    .execute(&state.db)
-    .await;
-
-    let _ = sqlx::query!(
-        "UPDATE repositories SET updated_at = NOW() WHERE id = $1",
         repo.id,
+        "chef",
+        &chef_metadata,
     )
-    .execute(&state.db)
     .await;
 
     info!(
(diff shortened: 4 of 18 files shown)


brandonrc:
## Review — changes requested

Thank you for this, and for going at the chokepoint rather than patching handlers one at a time: registering at `record_artifact_metadata` plus a resumable backfill is the right shape, and all 15 handlers named in #3659 now reach the catalog. Five things need fixing before it lands, and two of them lose data.

**Blocker — the projection collapses two formats' identities.** `package_catalog.rs:181-183` prefers `metadata["name"]` over `artifacts.name`. For Terraform those differ by design (`artifacts.name` is `namespace/name/provider` at `terraform.rs:1963`, while metadata carries the bare name at `:2003`), and Swift has the same split (`swift.rs:215` vs `:938`). Since `packages_repository_id_name_key` is UNIQUE, two distinct modules collapse onto one row and one overwrites the other — I reproduced it in SQL, watching version go 1.0.0 → 2.0.0 and size 111 → 222 on what should be separate rows.

**Major — the backfill overwrites existing metadata.** `package_catalog.rs:293` and `:365` pass `Some({"format":…})`, and `package_service.rs` applies `metadata = COALESCE($6, metadata)`. So a single `/admin/packages/reindex` run rewrites every Conan (`conan.rs:1891`) and PyPI (`pypi.rs:6270`) catalog row down to just `{"format":…}`. Reproduced.

**Major — Go modules report the wrong size.** `goproxy.rs:1448` and `:1547` project identically (`package_catalog.rs:101`), so the zip and the `.mod` land on the same key and roughly half of Go modules end up reporting the go.mod's size. This is the same failure the `.prov` handling was added to prevent.

**Moderate — quarantined uploads become visible.** No quarantine predicate at `package_catalog.rs:241-256`, so a rejected upload now surfaces by name across 20 formats.

**Minor** — virtual expansion at `packages.rs:69-83` is not recursive and has no `repo_type='virtual'` guard.

**On the issues this closes.** It closes #3659, partially. It does **not** close #3660 and in fact makes it worse: rows now exist for around 20 more formats, and every one of them is a permanent ghost on delete, because `is_deleted` is filtered on write only. #3601 and #3533 are untouched (OCI is exempt at `package_catalog.rs:85`), and #3602 only appears fixed — the virtual branch still writes no artifact row, so there is nothing to project; the repository filter merely makes the tab non-empty for other formats. Please adjust the PR description so it does not claim those.

**Performance.** The read path is fine and slightly faster than main (per-repository view 4.51 ms versus 7.35 ms on 60k artifacts, identical buffers). The problem is the backfill: the `LEFT JOIN artifact_metadata` plans as a merge join whose inner scan restarts at row zero each page, so scanned rows grow 595 → 10,572 → 30,593 as the cursor advances. That is quadratic, and on a million-artifact instance it is on the order of 10⁸ row reads.

**Tests.** All 23 pass, but none can fail on main because the module does not exist there, so nothing revert-proves the mechanism. Worth adding three DB-backed ones: a publish landing a row, a backfill run, and a delete.

**Bookkeeping.** There is no CHANGELOG entry, which blocks a release cut at preflight check 5. One bullet under `## [Unreleased]` → `### Fixed`, first reference `#3659`.

Good filters news: visibility, empty-key, remote-exclusion and OCI-exclusion all survive the rewrite, and I found no cross-repository leak.

brandonrc (changes requested):
Recording my review formally — the detailed findings are in [this comment](https://github.com/artifact-keeper/artifact-keeper/pull/3816#issuecomment-5642137781), which I posted as a plain comment by mistake, so it never registered as a blocking review.

Nothing has changed in the substance. The blockers are the identity collision at `package_catalog.rs:181-183` (preferring `metadata["name"]` over `artifacts.name` collapses Terraform and Swift identities onto the UNIQUE `packages_repository_id_name_key`), the backfill metadata destruction via `metadata = COALESCE($6, metadata)` at `:293` and `:365`, and the Go zip/.mod collision. Please read the linked comment for the full set, including the quarantine predicate and the non-recursive virtual expansion.

Apologies for the noise — this is the same review, just in the form that actually blocks the merge.

garrettKelleyCyOne:
Thanks for the careful review. I was able to reproduce each finding before fixing it, and everything is addressed in 17df943d; the description is updated to match.

**Blocker: identity collapse.** Fixed. The projection now registers under `artifacts.name` and no longer prefers the manifest's `name`. Every handler stores the full identity there, and it is what the self-registering handlers pass. Checking the handlers turned up a third case of the same bug: for PyPI, `metadata.name` is the display name (`Flask`) while the handler's row is `flask`, so the backfill would have added a second row. The formats that key their catalog on something else are handled explicitly: Maven/Gradle `groupId:artifactId` (now parsed from the path first, since `artifact_service`'s finalize path records no `groupId` metadata), SBT `org:module`, and Conan. `a_publish_through_the_metadata_chokepoint_lands_a_catalog_row` fails if the manifest name is preferred again, showing both Terraform modules collapsing onto `vpc`.

**Metadata overwrite.** Fixed with `MetadataWrite::FillMissing` in `PackageService::upsert_from_artifact`: the projection writes `metadata = COALESCE(metadata, $6)`. Handler callers keep the existing `COALESCE($6, metadata)`. `a_backfill_registers_stored_artifacts_without_touching_handler_rows` fails if it replaces.

**Go size.** Fixed. Descriptors stored at the package's own name and version are no longer projected. That covers Go `go.mod`, and while here also SBT Ivy descriptors (same failure) and Protobuf `_labels` rows (version `_labels`, 0 bytes). `a_go_module_reports_the_zip_size_not_the_go_mod` fails without the rule (42 vs 4096).

**Quarantine.** Fixed. The projection now uses the listability predicate the format indexes apply (Terraform's version list, the CocoaPods shard query): a rejected artifact is never registered, and a quarantined one only after its timed hold lapses. Because an upload under a hold is now skipped at publish time, `quarantine_service::transition(Released)` and the scanner's clear-hold branch register it on release. One gap remains: a hold that lapses with no scan and no admin action changes no row, so nothing fires for it. The reindex picks those up, and the CHANGELOG upgrade note says so. `deleted_and_unreleased_artifacts_stay_out_until_released` covers deleted, rejected, held, and then released.

**Virtual expansion.** Superseded. #3535 merged its own member aggregation into `list_packages` while this was in review, so the merge with main drops this branch's version rather than keeping two. Nesting is #3840.

**Issues.** Agreed on all four points. The description now claims only #3659 and states that #3660 gets worse and why, that #3601 and #3533 are untouched, and that #3602 is not fixed. I'm happy to take #3660 separately; there are about 30 soft-delete sites with no shared chokepoint, so it needs its own design.

**Backfill performance.** Reproduced: on 60k artifacts under a generic plan, the old query read 1,001 → 30,501 → 59,501 metadata rows per page. The `LIMIT` now lives in an `artifacts` subquery (`id > $2` with a nil-UUID default, so it is a real index range), and metadata is a `LATERAL (… LIMIT 1)` lookup. Without the `LATERAL`, the planner hashed all of `artifact_metadata` on every page: flat per page, but still a full-table read each time. After the change it reads 500 at every cursor, with or without a repository filter. The table is in the description.

**Tests.** Four DB-backed catalog tests were added (a fifth, for the Virtual listing, went away with the code it covered), and each was confirmed to fail when its mechanism is reverted; the table in the description lists what each one proves. For "a delete", I tested that a soft-deleted artifact is registered by neither path. It can't pin removal, because removal doesn't exist yet (#3660).

**CHANGELOG.** Added under `## [Unreleased]` → `### Fixed`, first reference #3659, with the reindex upgrade note.

garrettKelleyCyOne:
Re-scoped against `main`.

This opened as a catalog projection at the `record_artifact_metadata` chokepoint, to close #3659 and #3660. #3932 closed both first, from the other direction — per-handler `register_published_package` calls, `POST /api/v1/admin/packages/backfill`, and the liveness filter for ghost rows — and that is the shape in `main` now. The projection, the `package_catalog` module and the `/reindex` endpoint are dropped; there is no reason to keep arguing the point, and #3932 covers ghost rows, which this PR never did.

What is left is the two defects I could still reproduce against 1.10.0, found by pointing its backfill at a restored production dump and diffing the catalog:

- **#4168** — the walk never filtered `quarantine_status`, so a run listed `quarantined` and `rejected` artifacts as available packages. A repository holding one `released` and one `rejected` module listed 1 package before the backfill and 2 after.
- **#4169** — `maven`/`gradle` are absent from `BACKFILL_FORMATS`, so the endpoint reports `artifacts_scanned: 0` for them. Adding them naively would have registered a group-less row beside the handler's `groupId:artifactId`, which is the duplicate shape #3978 just fixed for NuGet, so both derive coordinates through `MavenHandler::parse_coordinates` instead.

One thing worth flagging from writing that second fix: `MavenHandler::parse_coordinates` **accepts** a repository-metadata filename and reads coordinates off whatever directory it happens to sit in, so `com/acme/widget/maven-metadata.xml` parses as `com:acme` at version `widget`. This PR rejects metadata, checksum and signature paths before the parse. `artifact_service::finalize_upload` calls the same parser on the same kind of path without that guard (`artifact_service.rs:965`) — I have not touched it here, but it looks reachable for a generic-API upload into a Maven repository, and may deserve its own issue.

Both fixes land with regression tests that fail on `main`. Verified end to end on a disposable cluster restored from a production dump.

The branch name still says `projection` because a PR's head branch cannot be changed; the title and contents are what it is now.

brandonrc:
Re-audited after your re-scope, and this is the right shape now: one commit on main's #3932 design, the quarantine predicate matches the download-path check exactly (expired holds included), Maven and Gradle coordinates come through the handler's own parser so jar and pom collapse to one row, and the three DB-backed tests run for real in CI. Merging once the remaining CI jobs are green; my 09-12 review is dismissed since it applied to the old projection.

Two gaps this PR is not asked to cover, filed separately: held uploads are still registered at publish time by 21 handlers and the read-side liveness check never looks at `quarantine_status`, so a later rejection never delists them (adding the same filter to `LIVE_ARTIFACT_EXISTS` would close both); and releasing an artifact from quarantine never registers it, so a held artifact that missed the backfill stays off the Packages page until the backfill is re-run. Your own note about `finalize_upload` parsing `maven-metadata.xml` paths as coordinates is filed too.

PR #812, opened 2026-04-21
feat: add support for OTLP http/protobuf transport protocol

brandonrc:
## Summary

Adds `http/protobuf` as an alternative OTLP exporter transport, selectable via the standard `OTEL_EXPORTER_OTLP_PROTOCOL` environment variable. This works around the h2/tonic `FRAME_SIZE_ERROR` that occurs when Artifact Keeper runs behind Envoy Gateway (#729).

Based on the work by @Firjens in #733. Their implementation was clean and correct; this PR recreates it on our own branch with no functional changes to the approach.

**What changed:**

- `Cargo.toml`: Added `http-proto` and `reqwest-client` features to `opentelemetry-otlp`
- `backend/src/telemetry.rs`: New `OtlpProtocol` enum that parses `OTEL_EXPORTER_OTLP_PROTOCOL` and routes exporter construction to `.with_tonic()` (gRPC) or `.with_http()` (HTTP/protobuf)
- Default remains `grpc`, matching the OTel spec
- Accepted protocol values: `http/protobuf`, `http-protobuf`, `http_protobuf` (case-insensitive)

Closes #729

## Test Checklist
- [x] Unit tests added/updated
- [ ] Integration tests added/updated (if applicable)
- [ ] E2E tests added/updated (if applicable)
- [ ] Manually tested locally
- [x] No regressions in existing tests

## API Changes
- [x] N/A - no API changes

PR #3834, opened 2026-09-11
fix(telemetry): use the blocking reqwest client for OTLP http/protobuf

### LINKED ISSUE #3833: OTLP http/protobuf export never works: BatchSpanProcessor thread panics with "there is no reactor running"
### Description

Setting `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` produces a deployment that logs `OpenTelemetry tracing enabled` at startup and then exports **nothing**. About five seconds in, the batch processor's thread panics:

```
thread 'OpenTelemetry.Traces.BatchProcessor' (39) panicked at
library/core/src/ops/function.rs:250:5:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

The panic kills that thread, not the process, so the server keeps serving traffic normally. Every span created from then on is dropped with:

```
WARN opentelemetry_sdk: name="BatchSpanProcessor.OnEnd.AfterShutdown" Spans are being
emitted even after Shutdown. This indicates incorrect lifecycle management of
TracerProvider in application. Spans will not be exported.
```

one per request, forever. The failure is silent in the sense that matters: the startup log says tracing is on, `/health` is fine, and no span ever reaches the collector.

**Root cause.** The workspace enables `opentelemetry-otlp` with `reqwest-client`:

```toml
opentelemetry-otlp = { version = "0.32", features = ["grpc-tonic", "http-proto", "reqwest-client"] }
```

`reqwest-client` is the **async** client, and it is first in the builder's documented priority order (`reqwest-client` > `hyper-client` > `reqwest-blocking-client`), so it wins even though `reqwest-blocking-client` is already on via default features. Meanwhile `telemetry.rs` wraps the exporter in the dedicated-thread processor:

```rust
.with_span_processor(BatchSpanProcessor::builder(exporter).build())
```

In `opentelemetry_sdk` 0.32 that processor runs exports on its own OS thread, which has no Tokio runtime. The async reqwest client needs a reactor, so the first flush panics. The blocking impl in `opentelemetry-http` runs the request synchronously inside the future and needs no reactor, which is exactly why it is the crate's default.

The gRPC path is unaffected — tonic is driven on the main Tokio runtime.

This matters because **`http/protobuf` is not optional for every backend**: Dynatrace's OTLP API does not accept OTLP/gRPC at all, so `http/protobuf` is the only way to export to it. The same applies to any collector fronted by an HTTP-only ingress.

Note `opentelemetry_sdk` is configured with `rt-tokio`, but nothing uses it — the runtime-backed processor lives behind the separate `experimental_trace_batch_span_processor_with_async_runtime` feature.

### Environment Information

- Deployment type: Kubernetes (Helm)
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL 16 (RDS)
- Reverse Proxy: Kong ingress; OTLP goes direct to an in-cluster Dynatrace ActiveGate
- Network setup: private EKS cluster, collector reachable in-cluster

### Backend Version

1.8.2 (`git=c8dc5f7`) — also present on `main`: `backend/src/telemetry.rs` is byte-identical and the `opentelemetry-otlp` feature set is unchanged through 1.9.0.

### Web UI Version

1.8.0 (not involved)

### Package Format

N/A / General

### Expected Behavior

With `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` set, spans are exported over HTTP/1.1 with protobuf bodies, the same way the `grpc` protocol works today.

### Observed Behavior

The batch processor thread panics on its first export with `there is no reactor running`. No spans are ever exported. Every subsequent span logs a `BatchSpanProcessor.OnEnd.AfterShutdown` warning — 108 of them in the first 2.5 minutes on a lightly loaded dev instance — so the broken path also costs span construction plus a WARN per request.

### Steps to Reproduce

1. Deploy the backend with any reachable OTLP HTTP endpoint:
   ```
   OTEL_EXPORTER_OTLP_ENDPOINT=http://otel-collector:4318
   OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf
   ```
2. Confirm startup logs `OpenTelemetry tracing enabled ... protocol="http/protobuf"`.
3. Send any request (e.g. `GET /health`).
4. Watch the log ~5s after boot: the `OpenTelemetry.Traces.BatchProcessor` thread panics, then every request emits an `AfterShutdown` warning.
5. No spans arrive at the collector.

Setting `OTEL_EXPORTER_OTLP_PROTOCOL=grpc` against the same collector on 4317 works, which isolates the fault to the HTTP transport.

### Logs

```shell
2026-09-11T11:41:20.725340Z  INFO artifact_keeper_backend::telemetry: OpenTelemetry tracing enabled otel_endpoint="http://<activegate>/e/<env>/api/v2/otlp" service_name="artifact-keeper" protocol="http/protobuf" log_format="pretty"
2026-09-11T11:41:21.167070Z  INFO artifact_keeper: HTTP server listening on 0.0.0.0:8080

thread 'OpenTelemetry.Traces.BatchProcessor' (39) panicked at /rustc/254b59607d4417e9dffbc307138ae5c86280fe4c/library/core/src/ops/function.rs:250:5:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

2026-09-11T11:41:29.058581Z  WARN opentelemetry_sdk: name="BatchSpanProcessor.OnEnd.AfterShutdown" Spans are being emitted even after Shutdown. This indicates incorrect lifecycle management of TracerProvider in application. Spans will not be exported.
2026-09-11T11:41:29.058721Z  WARN opentelemetry_sdk: name="BatchSpanProcessor.OnEnd.AfterShutdown" Spans are being emitted even after Shutdown. This indicates incorrect lifecycle management of TracerProvider in application. Spans will not be exported.
```

### Additional Context

**Proposed fix:** drop `reqwest-client` from the `opentelemetry-otlp` feature list so the default `reqwest-blocking-client` supplies the client. That is a one-token change and needs no code change in `telemetry.rs`, because the blocking `HttpClient` impl satisfies the same trait without a reactor.

The alternative — keeping the async client and switching to the runtime-backed processor — requires `opentelemetry_sdk`'s `experimental_trace_batch_span_processor_with_async_runtime` feature, so it trades a stable API for an experimental one to reach the same place.

I have a PR ready with that fix plus a regression test that drives one export on a plain `std::thread` with no reactor in scope, which panics on `main` and passes with the fix. Happy to adjust the approach if you would rather go the async-runtime route.

One related cleanup it touches: the existing `test_build_span_exporter_http_protobuf_feature_conflict` asserts the builder returns `NoHttpClient` on the premise that the two reqwest features are mutually exclusive. That has not been true since the crate moved to priority-based selection — the build always succeeds and the test's assertion is unreachable, which is part of why this went unnoticed.

oba11:
## Summary

Fixes #3833.

`OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` has never exported a span. The backend logs `OpenTelemetry tracing enabled`, then about five seconds later the batch processor's thread panics with `there is no reactor running, must be called from the context of a Tokio 1.x runtime` and dies. Because the panic is on a background thread the process survives, so the service keeps serving traffic while every span from then on is dropped with a `BatchSpanProcessor.OnEnd.AfterShutdown` warning — one per request, forever. Tracing looks enabled and exports nothing.

The workspace enabled `opentelemetry-otlp`'s `reqwest-client` feature, which is the **async** client and sits first in the builder's priority order (`reqwest-client` > `hyper-client` > `reqwest-blocking-client`), so it was selected even though the default `reqwest-blocking-client` was also on. `telemetry.rs` then hands that exporter to `BatchSpanProcessor`, which in `opentelemetry_sdk` 0.32 runs exports on its own OS thread with no Tokio runtime. The async client needs a reactor there and panics on the first flush.

The fix is to stop enabling `reqwest-client` and let the crate's default blocking client through. `opentelemetry-http`'s blocking impl runs the request synchronously inside the future, so it needs no reactor — which is why it is the crate's default for exactly this processor. No change to `telemetry.rs` logic is required, and the gRPC path is untouched (tonic is driven on the main runtime).

Worth noting for reviewers: the async client is not simply a faster option here. It cannot work with the dedicated-thread processor at all. The alternative — keeping it and switching to the runtime-backed processor — needs `opentelemetry_sdk`'s `experimental_trace_batch_span_processor_with_async_runtime` feature, so it swaps a stable API for an experimental one to reach the same place. Happy to go that route instead if you would prefer it.

This also matters beyond a preference for HTTP: Dynatrace's OTLP API does not accept OTLP/gRPC, so `http/protobuf` is the only way to export to it, and the same is true of any collector behind an HTTP-only ingress. That is how we hit it.

`test_http_protobuf_export_does_not_require_a_tokio_reactor` reproduces the processor's thread shape directly: it builds the HTTP/protobuf exporter and drives one export to completion on a plain `std::thread` with no reactor in scope, using a non-Tokio executor. It is deliberately not a `#[tokio::test]` — a reactor in scope is precisely what the exporter must not depend on.

On `main` the thread should panic with `there is no reactor running` and `join()` return `Err`, failing the test with the same signature seen in production. With this change the blocking client resolves the future without a reactor and the thread exits cleanly. Please see the honesty note below on what I was and was not able to run locally. The export itself is expected to fail (it points at a dead port); only the absence of a panic is asserted, since that is what distinguishes the two clients. An empty batch is enough — the exporter has no empty-batch short circuit, so the HTTP send is still attempted.

I also replaced `test_build_span_exporter_http_protobuf_feature_conflict`. It asserted the builder returns `NoHttpClient`, on the premise that the two reqwest features have mutually exclusive cfg guards. That stopped being true when the crate moved to priority-based selection: the build always succeeds, so the assertion sat inside an `if let Err(...)` that could never be entered. A test that passes whatever happens is part of why this went unnoticed for so long, so it is now a straightforward build assertion, with the behaviour that actually matters covered by the regression test above.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

A build-feature change and a test; no runtime API surface is touched.


brandonrc (changes requested):
## Review — one required change, then this is good to merge

I reviewed this adversarially, because "swap in a blocking HTTP client" is normally the shape of a change that trades a visible panic for an invisible hang. It isn't, here. The diagnosis in #3833 is right, the fix reaches the cause rather than a symptom, and the regression test genuinely fails without it. The only blocker is bookkeeping.

### The question that decides safety: which thread drives the export?

A dedicated OS thread with no Tokio runtime, and in this codebase that is the only place it can run:

- The exporter is handed to `BatchSpanProcessor` (`backend/src/telemetry.rs:197`), which spawns its own thread — `opentelemetry_sdk-0.32.1/src/trace/span_processor.rs:365-367`, named `OpenTelemetry.Traces.BatchProcessor`, exactly the thread in #3833's panic.
- Exports are driven there with `futures_executor::block_on` (`span_processor.rs:562-563`) — precisely the shape your test models.
- The async-runtime processor variant is behind `experimental_trace_batch_span_processor_with_async_runtime`, which is not enabled; `rt-tokio` is enabled but unused, as you note.
- Spans are the only OTLP signal wired up, so there is no second exporter with different threading.

The site a reviewer will react to first is *construction*, which happens inside `#[tokio::main]` (`backend/src/main.rs:150`, `main.rs:1360`). That is also safe, and not by luck: `opentelemetry-otlp-0.32.0/src/exporter/http/mod.rs:253-264` builds the blocking client as `std::thread::spawn(…).join()`, deliberately off-runtime. I measured it — 46.8 ms, no panic. See F3: the PR should say this out loud.

### I reproduced the bug rather than taking the description on trust

Re-adding `reqwest-client` to `Cargo.toml` (main's feature set) with your test in place:

```
FAIL telemetry::tests::test_http_protobuf_export_does_not_require_a_tokio_reactor
thread '<unnamed>' panicked at core/src/ops/function.rs:250:5:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
```

and a throwaway probe that builds the real provider plus `BatchSpanProcessor` and emits one span independently reproduced production's exact line, on the named processor thread. `exporter/http/mod.rs:229-233` confirms the priority order you describe (`reqwest-client` > `hyper-client` > `reqwest-blocking-client`).

I also checked the fix at the feature level rather than only via green tests: `cargo tree -p artifact-keeper-backend -e features -i opentelemetry-otlp` on this head shows `reqwest-blocking-client` present via `default` and **no** `reqwest-client`, **no** `hyper-client`. And no TLS regression — `blocking`, `default-tls`, `native-tls`, `rustls` and `http2` are all still on for `reqwest 0.13.4` through the workspace's own dependency, so an `https://` endpoint (the Dynatrace case) still works.

Worth recording that there is a lineage here: #812 added both the http/protobuf transport and the `reqwest-client` feature that breaks it, and #835 then patched a `NoHttpClient` panic in the same area on the premise that two reqwest versions coexist. That premise is now stale — `opentelemetry-otlp` 0.32 resolves to the same `reqwest 0.13.4` the app uses. This is the fix that reaches the cause.

### Shutdown, measured against a black hole

Against `198.51.100.1:4318` (TEST-NET-2, packets dropped, no RST), one span emitted, provider built and shut down inside a multi-thread runtime:

```
build_exporter_in_async_ctx = 46.783897ms
shutdown_elapsed            = 5.000598402s  res = Err(InternalFailure("Shutdown errors: [Timeout(5s)]"))
runtime dropped cleanly
```

So: **no hang on exit.** `SdkTracerProvider::shutdown` caps at 5 s (`provider.rs:273`), which fires before the exporter's own 10 s client timeout, and on timeout the SDK abandons rather than joins the blocked thread (`span_processor.rs:704-712` joins only on success). The final batch is lost in that case and the 5 s blocks a Tokio worker from `OtelGuard::drop` — pre-existing in shape, bounded, and not a blocker, but newly *reachable* on this path, hence F4.

One thing for whoever touches the entrypoint next: `opentelemetry_sdk` warns that `shutdown()` from the main thread of a **current-thread** runtime can deadlock (`span_processor.rs:328-331`). We are safe because both entrypoints are multi-thread (`main.rs:1360`, `main.rs:1387-1391`). Neither should become `new_current_thread`.

### The test earns its place

51/51 in `telemetry::` pass; clippy clean on `--lib --tests`. The test is deliberately not `#[tokio::test]` and drives the export with `futures::executor::block_on` on a plain `std::thread` — the same executor and thread shape as the SDK, so it is a faithful model rather than an approximation, and it fails on main's feature set with the production panic string. Because the fix is the *absence* of a Cargo feature and features are additive, this test is the only thing standing between us and a silent regression; I checked the blast radius and only an explicit re-add of `opentelemetry-otlp/reqwest-client` trips it (`reqwest-rustls` and friends enable the optional *dependency*, a different feature name).

One note: your claim that an empty batch still attempts a send is correct **at the exporter level** (`src/span.rs:179-186` → `exporter/http/trace.rs:12-20` have no short circuit), but the *processor* does short-circuit (`span_processor.rs:552-554`). The test only works because it calls the exporter directly. That is load-bearing — please add a comment saying so, or someone will later "simplify" it to go through the processor and the test will pass for the wrong reason forever.

---

## Findings

**F1 — blocker, bookkeeping: no CHANGELOG entry.** The diff is `Cargo.toml` + `telemetry.rs` only; `awk '/^## \[Unreleased\]/{f=1;next} /^## \[/{f=0} f' CHANGELOG.md` has no otlp/telemetry match and `grep -c 3833 CHANGELOG.md` is 0. Our release cut reconciles `## [Unreleased]`, so a missing bullet silently drops the fix from the notes. Please add under `## [Unreleased]` → `### Fixed`, issue number first:

> - **OTLP `http/protobuf` never exported a span** (#3833, #3834). `opentelemetry-otlp`'s `reqwest-client` feature, added in #812, selects the *async* reqwest client, which is first in the builder's priority order. `BatchSpanProcessor` drives exports from a dedicated OS thread with no Tokio runtime, so the first flush panicked with `there is no reactor running`, killing the processor thread while the process kept serving — every later span dropped with a `BatchSpanProcessor.OnEnd.AfterShutdown` warning, one per request. The feature is removed so the crate's default `reqwest-blocking-client` supplies the client. `grpc` is unaffected.

**F2 — minor: `Cargo.toml:68` now contradicts the comment two lines below it.** `opentelemetry-http = { features = ["reqwest"] }` pulls the *async* impl, and `opentelemetry_http` is imported nowhere in the tree. It is inert — the otlp builder gates on its own `reqwest-client` feature, and I confirmed the resolved feature set — but it is the single most likely thing to mislead the next person into re-adding `reqwest-client`. Either drop the dependency or switch it to `reqwest-blocking` and say why it is still pinned directly.

**F3 — minor: argue the construction site, not just the export site.** The comment block (`Cargo.toml:69-75`) and module doc (`telemetry.rs:10-14`) explain why blocking is safe *on the processor thread* and say nothing about the client being built inside `#[tokio::main]`. One clause pointing at `exporter/http/mod.rs:253-264` closes the first objection anyone will raise.

**F4 — nit, for the PR body:** with an unreachable collector this path now adds up to 5 s to shutdown and drops the final batch (measured above). Operators have no knob — `OTEL_EXPORTER_OTLP_TIMEOUT` does not help, since the 5 s cap is hard-coded in the SDK.

Add the CHANGELOG bullet and I will merge. F2–F4 are yours to take or leave; F2 is the one I would genuinely rather see done now, while the context is fresh.

oba11:
Thanks @brandonrc for the thorough review! All findings have been addressed:

- **F1 (CHANGELOG)**: Added the bullet under `## [Unreleased] -> ### Fixed` (and rebased onto `main` following the 1.10.0 release cut).
- **F2 (`opentelemetry-http`)**: Removed the unused `opentelemetry-http` direct dependency from `Cargo.toml` and `backend/Cargo.toml`.
- **F3 (Construction site)**: Documented why building the blocking client is safe inside `#[tokio::main]` via `exporter/http/mod.rs`'s spawned thread in both `Cargo.toml` and `telemetry.rs`.
- **Test note**: Clarified in the test comments why calling the exporter directly is load-bearing due to `BatchSpanProcessor`'s empty-batch short-circuit.

The branch has been rebased onto latest `main` and is ready for review again.

brandonrc:
@oba11 I will get this merged after a large security epic gets merged in. Adding some nice security scanning features into the 1.11.0 release. 

But I appreciate the work and will make sure this gets merged in asap.

brandonrc:

Issue #3305, opened 2026-08-13
Proxy-cache catalog backfill cannot repair the placeholder row it races, leaving #2218 self-healing largely inert


brandonrc:
Found during review of #3289 (an outside contribution bounding the backfill spawn). This is **pre-existing on main** and independent of that PR, but it undercuts the stated rationale for the backfill, so it is worth tracking on its own.

## The race

On a proxy cache serve, two things touch `proxy_cache_artifacts` for the same key:

1. `record_proxy_download` is awaited **inline on the request** (`backend/src/api/handlers/proxy_helpers.rs:4467`) and inserts a placeholder row with `size_bytes = 0` (`backend/src/services/proxy_catalog.rs:360-366`).
2. `spawn_lazy_catalog_backfill` runs detached and calls `backfill_from_sidecar`, whose `ON CONFLICT` arm bumps only `last_accessed_at` (`backend/src/services/proxy_catalog.rs:112-113`).

So whenever the foreground insert wins — which is the common case, since it is on the request path and the backfill is queued behind `db.acquire()` — the backfill finds a row present, updates a timestamp, and **returns without ever writing the real `size_bytes`, checksum or content type**. The placeholder is permanent: nothing else repairs it.

The consequence is that the self-healing #2218/#2270 introduced has been **largely inert on main**. Entries cached before `proxy_cache_artifacts` existed are exactly the ones the backfill is for, and on their next hit they get a zero-size row rather than a repaired one.

## Why it surfaces now

#3289 removes a storage read from the backfill path, which makes the backfill materially faster and therefore **more likely to win** the race it currently loses. That improves matters by accident, but it does not fix the conflict arm — and a backfill that wins sometimes and silently no-ops otherwise is harder to reason about than one that consistently loses.

## Suggested direction

Make the conflict arm repair the row when the existing values are placeholders, rather than only touching `last_accessed_at`. Something along the lines of updating `size_bytes`, `checksum_sha256` and `content_type` when the stored `size_bytes` is `0` (or the checksum is null), while leaving a genuinely-populated row alone so a real record is never overwritten by a stale sidecar.

The ordering question is worth settling at the same time: if the placeholder insert is unavoidable on the request path, the backfill is by definition always an update, and the conflict arm is the *only* path that matters.

## Verification note

[...]

Issue #2537, opened 2026-07-14
Proxy-cache first-serve download undercount: record_proxy_download runs before the async tee commits the catalog row (#2270 follow-up)


brandonrc:
Minor follow-up from #2531 QA (non-blocking). On a cache MISS, `record_proxy_download` runs before the async tee commits the `proxy_cache_catalog` row, so the object's row is absent and the very first download of each freshly-cached object records +0 (subsequent hits count correctly; the code comment acknowledges 'no-ops if the row is not yet present'). Off-by-one on first fetch of each object. Fix: record on the hit path, or after the tee commit, so the first serve counts. Origin: PR #2531 / #2270.

Issue #2522, opened 2026-07-14
PF-006 [P1]: Side-effect pipelines not uniformly durable/bounded; downloads add synchronous event writes


brandonrc:
# PF-006: Side-effect pipelines are not uniformly durable/bounded, and downloads add synchronous event writes

Priority: P1  
Affected area: upload throughput, scanning, quality checks, OpenSearch, audit, download telemetry

## Finding

An upload performs some supposedly non-blocking work inline and spawns separate Tokio tasks for replication fan-out, scanning, quality checks, and OpenSearch indexing. There is no common durable outbox or visible global/per-engine capacity boundary for the scan and quality work. Per-artifact OpenSearch indexing is a single HTTP request, and failures are logged rather than durably retried.

The hot download path awaits a raw `download_statistics` insert and an audit insert before returning the stream. Download counts are later calculated with `COUNT(*)` over raw events. Audit has retention support, but no download-statistics retention/partitioning was found.

## Evidence

- Package catalog update and “non-blocking” plugin hook are actually awaited during upload finalization: [artifact_service.rs](../backend/src/services/artifact_service.rs#L838).
- Upload finalization separately `tokio::spawn`s replication fan-out, scan-on-upload, quality checks, and OpenSearch work: [artifact_service.rs](../backend/src/services/artifact_service.rs#L859).
- OpenSearch indexing fetches repository data and makes one index request per artifact; failure only produces a warning: [artifact_service.rs](../backend/src/services/artifact_service.rs#L977), [opensearch_service.rs](../backend/src/services/opensearch_service.rs#L122).
- `ScannerService` contains scanners and storage configuration but no shared concurrency semaphore/queue field; upload spawns directly into `scan_artifact`: [scanner_service.rs](../backend/src/services/scanner_service.rs#L2791).
- Scanner and quality implementations protect a single large input with streaming-to-mmap, which is good, but unbounded concurrent jobs can still multiply storage reads, scratch, address space, subprocesses, and external requests: [scanner_service.rs](../backend/src/services/scanner_service.rs#L4002), [quality_check_service.rs](../backend/src/services/quality_check_service.rs#L185).
[...]

PR #3843, opened 2026-09-12
fix(proxy): record cache hits in proxy download statistics

### LINKED ISSUE #3649: fix(proxy): cache hits are not recorded in proxy download statistics
## Summary

`proxy_download_statistics` gains a row on the **caching (miss)** path only. Requests
served from cache are not counted at all, so a proxy-heavy deployment shows almost no
downloads in the UI while serving continuous traffic.

## Evidence

On a test repository, 3 cache writes produced **1** stats row while dozens of served
requests went unrecorded. Re-pulling an already-cached artifact returns 200 and updates
`last_accessed_at` **without** producing a stats row — so the data to count the hit is
already being written on that path.

Insert sites, both on caching paths:

- `backend/src/services/proxy_catalog.rs:581`
- `backend/src/api/handlers/admin_security.rs:1324`

(table introduced in `backend/migrations/160_proxy_download_statistics.sql`)

## Expected

A served download is a download. Either:

- record cache hits in `proxy_download_statistics`; or
- document explicitly that the table counts cache **writes**, and expose a separate
  served-request counter.

The first matches the table's name and how the UI presents it. If the concern is write
amplification on the hot read path, a counter aggregated periodically would still be a
large improvement over counting nothing.

## Related gap worth stating in the same fix

`download_statistics` (the non-proxy table) is keyed on `artifact_id` and therefore cannot
capture proxy traffic at all, because proxied content never creates an `artifacts` row
(#1278, #1280). This is why an all-proxy deployment shows an empty downloads UI even
though both tables exist and both are "working as written".

## Acceptance criteria

- Serving an artifact from the proxy cache is reflected in download statistics.
- The downloads UI on a proxy-only repository shows non-zero traffic after repeated pulls
  of the same cached artifact.

## Severity

Analytics and attribution only — no functional impact on artifact delivery.

brandonrc:
## Summary

`proxy_download_statistics` (migration 160) is the sibling of `download_statistics` for
pull-through serves, because proxy-cached content deliberately has no `artifacts` row
(#1278 / #1280). #3446 wired `proxy_helpers::record_proxy_download` into nine formats and
explicitly **deferred** fourteen others, each left carrying an `UNRECORDED-PROXY-SERVE`
marker on its Remote download arm:

> alpine, chef, cocoapods, composer, conan, conda, gitlfs, jetbrains, pub, rpm, sbt,
> swift, terraform, vscode

Those arms served upstream **and** proxy-cached bytes and counted neither. On those
formats a re-pull of an already-cached artifact returns 200 and refreshes
`last_accessed_at` (the lazy catalog backfill on the cache-hit path) without producing a
stats row — exactly the evidence in #3649 — so a proxy-heavy deployment reported almost
no downloads while serving continuous traffic.

Each of those arms now records after the fetch resolves, keyed on the proxy-cache path the
fetch commits under. The streaming helpers (`proxy_fetch_streaming*`) answer a warm cache
HIT from storage and a cold MISS from upstream through the same call, so **one recorder
call covers both**, and a 404/502 still counts nothing. `ctx` is threaded into alpine's
`try_proxy_apk` and vscode's `proxy_gallery_asset` for that call; vscode counts only the
VSIX asset type, not the manifest / details / icon assets the client fetches while
rendering a listing.

PyPI already recorded, but only when the hit was **streamed**: the #1555
presigned-redirect fast path returned a 307 to the signed cache object and returned before
the recorder, so on an S3-backed deployment with `presigned_downloads_enabled` every warm
hit was invisible while the identical hit counted with presigning off. A redirect is now
counted at issue time — the rule `local_fetch_or_redirect` already applies on the hosted
side, since the client's subsequent object-store GET never reaches us — as are the
last-known-good wheels the PyPI download age gate substitutes.

### Which option from the issue, and why

The first one: **record cache hits**. The API already presents this table as *serves*, not
as cache writes — `RepositoryStorageBreakdown.proxy_download_count` is documented
"Downloads served by this repository's remote pull-through proxy", `StorageSnapshot`'s as
"Pull-through serves", and `DownloadTrend` unions it by day with hosted `download_count`.
Renaming the table to mean "cache writes" would have made those three published fields
lie, so the honest fix is to make the number true.

### Write amplification

`record_proxy_download` is a single statement (`WITH ensured AS (INSERT … ON CONFLICT DO
UPDATE SET last_accessed_at = now() RETURNING id) INSERT INTO proxy_download_statistics
… SELECT id FROM ensured`) awaited **inline** on the request, and that is already what the
house does on this path: all nine formats #3446 reached call it inline, deliberately, so
the catalog row exists before the response (#2537 / #3305). This PR follows that shape
rather than inventing a second one. The hosted `download_statistics` path is different —
`artifact_service::record_download` goes through the bounded, batch-flushing
`download_event_dispatch` outbox (#2522) — but its event is keyed on `artifact_id` and its
flush worker writes `download_statistics`, so reusing it would mean a new event variant, a
new flush arm, and giving up the "catalog row committed before the response" invariant
`record_proxy_download` exists to hold. That is a worthwhile follow-up and a bad thing to
smuggle into a bug fix; it is called out below rather than done here.

### Paths covered, and paths deliberately not

Covered (every Remote-repository serve arm that hands proxy/upstream bytes to a client):
the fourteen deferred formats above (18 call sites incl. conan ×2, terraform ×3, vscode
×2), plus PyPI's presigned-redirect fast path and its download-age-gate LKG substitution
(redirect, warm-cache and cold arms). The nine formats #3446 already fixed (cargo, debian,
goproxy, helm, maven, npm, nuget, oci_v2, pypi) were verified to already cover the hit
path, since their recorder sits after a cache-first fetch.

Deliberately **not** changed, each with an existing rationale in the tree:

- **HEAD** — serves no body, short-circuited inside `record_proxy_download` (#2260).
- **OCI blobs** — a Docker pull is counted once at the manifest (#2260); repodata /
  version lists / packuments are metadata, not downloads.
- **A Remote member reached through a VIRTUAL repository** — `resolve_virtual_download_
  streaming` records the local-member winner only and leaves Remote pass-through and
  proxy-cache hits uncounted (#1278 / #2260). Counting those safely needs a decision about
  the two-phase member fan-out, which probes members that may lose to a higher-priority
  Pass-2 candidate; changing it from inside this fix would have been a silent policy
  change. PyPI's age gate therefore passes `None` on its virtual-member call site.
- **`local_fetch_or_redirect`'s proxy-cache-key branch** — same #1278 exclusion.

### `download_statistics` cannot capture proxy traffic at all — stated, not fixed

`download_statistics` is keyed on a `NOT NULL artifact_id` foreign key, and proxied content
never creates an `artifacts` row (#1278 / #1280). It is therefore *structurally* incapable
of recording a proxy serve — which is precisely why the separate
`proxy_download_statistics` table exists (#2270). Consumers reading only `download_count`
will still see zero on an all-proxy deployment and must add `proxy_download_count`; the two
are disjoint by design. **That is not fixed here**: it is a schema change and belongs in its
own issue. I could not find an open issue tracking it (#2270 and #3446 are both closed).

## Verification

Database-backed runs against a throwaway Postgres 16 (`AK_TESTS_REQUIRE_DB=1`, migrations
`160`–`216` applied).

Regression test: `api::handlers::chef::tests::test_remote_cookbook_cache_hit_is_counted_3649`
— wiremock upstream, cold serve → 200 + 1 stats row, wait for the tee's full commit
(content + `__cache_meta__.json` sidecar), warm serve → 200 with **no additional upstream
request** (which is what proves it is a cache hit rather than a silent refetch) + **2** stats
rows, then HEAD → still 2.

**Fails on main**, both ways:

- With only the chef production hunk reverted: `left: 0, right: 1` —
  *"the first proxy serve must record exactly one proxy_download_statistics row"*.
- With the recorder present but gated on `!proxy.is_cache_fresh(..)`, i.e. simulating
  "count cache writes only": `left: 1, right: 2` —
  *"a serve from the proxy cache is a download and must be counted too; the counter must
  keep incrementing on repeat hits"*. This is why the repeat-hit assertion is in the test.

Structural gates, both in `proxy_helpers::proxy_download_recording_tests`:

- `the_deferred_format_count_only_shrinks` — 14 → **0**.
- `every_proxy_serving_format_records_3649` (new) — the marker-or-recorder class gate is
  satisfied by a *marker* too, so this pins the positive half: all 23 proxy-serving format
  handlers must contain a real `record_proxy_download(` call site. A revert that puts a
  deferral marker back now fails here instead of passing the gate.

Commands run (all through the compile lock, `SQLX_OFFLINE=true`):

```
cargo fmt
cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings      # clean
cargo nextest run -p artifact-keeper-backend --lib -E 'test(api::handlers::vscode::) or
  test(api::handlers::alpine::) or test(api::handlers::chef::) or test(api::handlers::rpm::)
  or test(api::handlers::terraform::) or test(api::handlers::conan::) or test(api::handlers::sbt::)
  or test(api::handlers::conda::) or test(api::handlers::swift::) or test(api::handlers::gitlfs::)
  or test(api::handlers::jetbrains::) or test(api::handlers::composer::)
  or test(api::handlers::cocoapods::) or test(api::handlers::pub_registry::)'
  -> 1177/1177 passed
cargo nextest run -p artifact-keeper-backend --lib -E 'test(api::handlers::pypi::) or
  test(api::handlers::proxy_helpers::) or test(services::proxy_catalog::) or
  test(services::analytics_service::)'                                   -> 666/666 passed
cargo nextest run -p artifact-keeper-backend --lib -E 'test(ci_test_surface) or test(general::)
  or test(download) or test(proxy)'                                      -> 1464/1464 passed (x2)
```

Not run: the full 16k-test lib suite, the e2e compose stacks, and the web app (no frontend
in this repository). One honest note — `api::handlers::admin::tests::test_get_system_stats_
reports_proxy_cache_totals_db` failed once in the first pass of the last selection and
passed on both repeat runs and in isolation; it is a global-`COUNT(*)`-delta test against
the shared database with its own 3-attempt retry, so it is contention-sensitive by
construction rather than affected by this change.

Closes #3649


brandonrc:
Filed the structural half as #3844 — `download_statistics` being keyed on `artifact_id` so proxy traffic cannot be represented there at all, plus the virtual-repository remote-member case this PR deliberately left alone. Neither #2270 nor #3446 was still open to carry it.

PR #3846, opened 2026-09-12
fix(tests): pin the #1808 401 for an anonymous Conan request naming an unknown repository

### LINKED ISSUE #3616: conan_error_paths_tests: unauthenticated GET on an unknown Conan repo returns 401, test pins the #1046 contract of 404

brandonrc:
## Summary

`conan_error_paths_tests::test_1046_unknown_repo_no_auth_returns_404` asserted that an
anonymous `GET /conan/{unknown}/v2/ping` returns **404**. It returns **401**, and that is
correct. **The test is the stale side; no production code changes in this PR.**

#3616 framed this as an open contract question, so I resolved it from the history rather
than flipping the assertion.

### What #1046 actually established

#1046 ("Add full-stack integration test exercising repo_visibility_middleware + Conan
handlers") was a *hardening* issue, not a design decision. It asked for a test that
"pins the middleware -> handler chain" for the four cases a security review had walked
through **manually**, and listed the statuses that chain produced at the time. It was
closed by #1204 (`f8b7dd55`, 2026-05-12), whose body says the tests "pin the visibility
behavior for the four cases security review walked through manually". Nothing in the
issue, the PR, or the test's own doc comment argues *for* 404 — client compatibility or
otherwise. It recorded the behaviour; it did not choose it.

### What changed, and whether it was deliberate

Deliberate, and security-motivated. **#1808** (`type:security`, "native download paths
leak private-repo existence via 401-vs-404 oracle") was filed 2026-06-09 with a working
anonymous enumeration sweep across maven/pypi/npm/cargo, and fixed the same day by
**#1812** (`5768b519`, shipped in **1.2.1**). Its root-cause section names the exact
divergence #1046 had pinned, and the fix added the `no_credential` arm now at
`backend/src/api/middleware/auth.rs`:

> `#1808`: Close the anonymous repo-existence oracle. An existing *private* repo returns
> 401 to an anonymous caller (visibility check below), so a nonexistent repo must not
> return the handler's 404 to that same caller -- the differing status would leak which
> repo keys exist.

So the "401 for everything unauthenticated" posture is not a middleware-ordering
accident. It is a fix that intentionally overrode the behaviour #1046 had recorded, and
it has been reinforced four times since: #3524, #3709 and #3717 unified the
authenticated-caller denials on the same existence-hiding answer, and #3730 extended the
identical treatment to the OCI `/v2` pull path.

**Also worth stating plainly: the issue's framing of 404 as the privacy-preserving option
is inverted.** #1046's contract was 404-for-unknown *and* 401-for-existing-private. Those
differ, and that difference *is* the oracle — anonymous 401 meant "this repository
exists". Uniform 401 is the option that reveals less, not more.

### Cross-format matrix (anonymous GET, no credential)

The gate is one shared layer — `repo_visibility_middleware`, applied once to the whole
`format_routes` nest in `api/routes.rs` — so this is not per-format behaviour:

| Route family | Unknown key | Existing private | Existing public |
|---|---|---|---|
| Every format under the `format_routes` nest — npm, maven, pypi, cargo (+ `/api/cargo`), **conan**, debian, nuget, rpm, gems, go, helm, composer, alpine, conda, swift, terraform, cocoapods, hex, huggingface, jetbrains, chef, puppet, ansible, cran, ivy, vscode, proto, incus/lxc, pub, lfs, general | **401** + challenge | **401** + challenge | 200 |
| OCI `/v2` (separate gate) | **401** + bearer challenge | **401** + bearer challenge | 200/404 |

There is **no cross-format inconsistency** — Conan already matches every other registry.
The inconsistency was between the shared middleware and one Conan test. Reverting Conan
to 404 could not be done in the Conan handler at all; it would mean reverting #1812 for
all ~30 formats and reopening the oracle registry-wide.

### What the real Conan client does

I ran conan **2.32.0** against mock servers for both statuses. The client never reads
`WWW-Authenticate` (zero occurrences in the package); it dispatches on the status code.

| | 401 | 404 |
|---|---|---|
| `conan install`, interactive | prompts for credentials **3x**, then `Too many failed login attempts, bye!` | remote skipped -> `Unable to find 'pkg' in remotes` |
| `conan install`, non-interactive (CI) | fails fast: `Conan interactive mode disabled` | same as interactive |
| `conan remote login` | `Wrong user or password` | `Invalid server response, check remote URL and try again` |

**This is the one honest argument for the issue's 404 side, and I want it on the record:
401 gives a worse message for a typo'd remote.** Three considerations outweigh it:

1. The prompt loop is bounded at `LOGIN_RETRIES = 3` (`conan/internal/rest/auth_manager.py`).
   It terminates; there is no infinite retry.
2. An **existing private** Conan repository must 401 regardless — that is how the client
   learns to send credentials at all — so the same UX applies to the case the server is
   *supposed* to serve. Returning 404 for the unknown key only restores the oracle; it
   does not improve the private case.
3. Conan has an unbounded self-recursion in `call_rest_api_method` if
   `/v2/users/authenticate` returns **200** while the resource keeps 401ing. This registry
   cannot produce that shape: with a valid credential and an unknown key the middleware
   returns the existence-hiding **404**, not 200.

The reverse trade — 404 for existing-private too, the other way to close the oracle — was
considered and rejected in #1808 itself, and the client evidence confirms that rejection:
with a uniform 404 Conan would never prompt for credentials, and an anonymous-first client
could not reach a private repository at all.

### Change

- `backend/tests/conan_error_paths_tests.rs`: the case becomes
  `test_1808_unknown_repo_no_auth_matches_private_401`. It asserts 401 **and** that the
  unknown-key response is identical in status and body to an existing-private one —
  a bare 401 assertion would be satisfied by an unrelated 401, whereas that equality is
  what closing the oracle actually requires. The authenticated side is untouched and still
  pinned by `test_990_upload_to_nonexistent_repo_returns_404`.
- `.github/workflows/ci.yml`: `--skip test_1046_unknown_repo_no_auth_returns_404` and its
  tracking comment removed; `conan_error_paths_tests` moves to the fully-green integration
  step (all 8 cases now run).
- `CHANGELOG.md`: one `### Fixed` bullet under `## [Unreleased]`, with an operator note
  for anyone who read the old 404 as a contract.

## Behaviour-change note

The wire behaviour does not change in this PR — it changed in **1.2.1** (#1812). But
anyone who read #1046's 404 as a contract should know it was retired: pointing a Conan
remote at a repository key that does not exist has answered 401 since 1.2.1, and
`conan install` surfaces that as a credential prompt (or `Conan interactive mode disabled`
in CI) rather than "not found". The CHANGELOG entry says so explicitly.

## Verification

DB: `postgres:16-alpine` on 127.0.0.1:5716, `sqlx migrate run --source backend/migrations`,
`AK_TESTS_REQUIRE_DB=1`, `JWT_SECRET` set.

- Baseline on `main` (048fc6a7): `8 tests run: 7 passed, 1 failed` —
  `test_1046_unknown_repo_no_auth_returns_404` panicked
  `unknown repo + no auth must return 404, got 401`.
- After: `cargo nextest run -p artifact-keeper-backend --test conan_error_paths_tests
  --run-ignored all` -> **8 passed, 0 skipped**.
- **Regression proof**: with #1812's `if no_credential { return unauthorized_response(); }`
  temporarily reverted to `not_found_response()`, the new test fails
  `unknown repo + no auth must return 401, got 404` (left: 404, right: 401). Revert backed
  out; no production file is modified in this PR.
- `cargo clippy -p artifact-keeper-backend --all-targets -- -D warnings` -> clean. The new
  body-buffering helper needed the file-level `#![allow(clippy::disallowed_methods)]`
  streaming-invariant exemption that the sibling test files (`oci_packages_index_tests`,
  `composer_packages_index_tests`) already carry; the read is bounded at 64 KiB.
- CI-surface gate (`ci_test_surface`, #3494): 7/7 pass, including
  `every_integration_target_is_wired_into_ci_or_exempt` — the suite is still named in an
  `-- --ignored` step after the move.
- `streaming_invariant` (both cases incl. `exempt_sites_match_allowlist`): pass.
- `scripts/ci/check-changelog-unreleased.sh`: pass;
  `awk '/^## \[/{h=$0} /^- \*\*/{print h}' CHANGELOG.md` confirms the bullet is under
  `## [Unreleased]`.
- `cargo fmt` applied.

Not run: the full workspace test suite, and the Conan client matrix against a live
artifact-keeper server (the client evidence above is from conan 2.32.0 against mock
servers reproducing each status, plus reading this repository's middleware for what it
returns in each credential state).

## Adjacent observations, deliberately not fixed here

- The suite's own module doc still describes items 1-3 as "#990" error paths; only item 4
  was restated.
- #3617 and #3618 quarantines are untouched.

Closes #3616


brandonrc:
Rebased onto main after #3841 and #3845 landed. With the Conan case unquarantined, the "Run quarantined-case integration suites" step has nothing left to run — #3617, #3618 and #3616 were its only three entries — so this now removes the step and its explanatory comment block entirely rather than leaving `cargo test --workspace` with no `--test` targets, which would have selected every target instead of none.

`conan_error_paths_tests` and `storage_stats_tests` both sit in the main integration step. `ci_test_surface` gate passes locally (7/7).

Issue #3402, opened 2026-08-15
Flaky: oci_migration_reindex tests are order-sensitive when the whole lib suite shares one database


brandonrc:
## Summary

`services::oci_migration_reindex` tests fail intermittently when `cargo test --workspace --lib` runs against a single shared Postgres, and pass reliably in isolation or with `--test-threads=1`. Independently observed by two separate workstreams today (during #3375 and #3400), each on branches that touch no OCI code.

## Signature

- A **different test in the module fails on each run** — the failure does not stick to one test.
- The failing test passes when run with `--exact` on its own.
- The whole module passes with `--test-threads=1`.
- Branches that reproduce it contain no changes under `services/oci_migration_reindex` or any OCI handler.

## Cause

`run_repair` performs an **instance-wide scan of `oci_tags`** rather than scoping to the repository under test. When the rest of the suite is concurrently creating and deleting OCI tags in the same database, the scan observes rows belonging to other tests, so its result depends on what else happens to be in flight.

## Why it matters

This is the second shared-test-database hazard found today (the first, #3377, is a hard hang). Both share a root cause worth addressing as a class: DB-backed tests assume exclusive use of a database they actually share. The practical costs are a red CI run that reviewers learn to ignore, and time lost re-running to decide whether a failure is real — exactly the ambiguity that makes a genuine regression easy to wave through.

## Suggested fix

Scope the repair scan to the repository (or instance) under test so it cannot observe other tests' rows — preferred, since it also makes the production code's blast radius explicit. Failing that, serialize the module (`serial_test`) or give it its own database. A per-test schema/namespace would fix the whole class rather than this one module.

## Note

This is **not** the same defect as #3377 (`services::http_client` hanging on a process-wide `HTTP_PROXY` env leak). They are separate issues with the same underlying theme.

brandonrc:
Fresh evidence that this class is wider than the `oci_migration_reindex` module, gathered while verifying `main` locally before cutting 1.8.1.

**Setup:** clean `main` (`0b795b5`), disposable PostgreSQL 16, all 188 migrations applied, `AK_TESTS_REQUIRE_DB=1`, `SQLX_OFFLINE=true`, `--skip services::http_client`.

**Two consecutive full-parallelism runs, same tree, same database:**

| Run | Result | Failing tests |
|---|---|---|
| 1 | 15679 passed, **10 failed** | all of `api::handlers::nuget::read_db_tests` |
[...]

Issue #3187, opened 2026-08-08
OCI cleanup sweep: close the pre-delete liveness window left open by #3158


brandonrc:
## Summary

#3158 re-asserts blob liveness immediately before the OCI cleanup sweep's storage delete. That **narrows** the data-loss window of #3085 — from the whole batch walk (up to `OCI_UPLOAD_CLEANUP_KEY_SCAN_LIMIT` = 1000 sequential object-store deletes) down to a single storage round trip — but it does **not** close it. This issue tracks the actual closure.

## Why the window is still open

`StorageGcService::hold_cleanup_key_claim_for_delete` runs its liveness re-check and lease renewal as one atomic `UPDATE`, but that `UPDATE` executes on the pool as a single autocommit statement. Its row lock is on `oci_upload_cleanup_keys`, never on `oci_blobs`, and it is released the instant the statement commits. The caller then performs the object-store `DELETE` holding **nothing**.

A push that commits its `oci_blobs` row inside that gap still has its bytes destroyed. Nothing in the push path serializes against the sweep:

- `register_oci_upload_cleanup_key` is `ON CONFLICT (storage_key) DO NOTHING`, which against an already-claimed row is a complete no-op.
- The push success path calls `clear_oci_upload_cleanup_key_best_effort` (`backend/src/api/handlers/oci_v2.rs:978`), a bare `DELETE FROM oci_upload_cleanup_keys WHERE storage_key = $1` with no `claim_token` guard and no claim check. It removes the journal row out from under an in-flight sweep.

That second point also removes the observability we might otherwise have relied on. The sweep's guarded row `DELETE` in `cleanup_unreferenced_oci_upload_keys` does not inspect `rows_affected`; it increments `cleanup_rows_removed` and `result.storage_keys_deleted` unconditionally (`backend/src/services/storage_gc_service.rs:1726-1727`). So in the exact interleaving where a push wins, the push deletes the journal row, the sweep's guarded `DELETE` matches zero rows, and the sweep records a **clean success**. The data loss leaves no journal row and no error.

## Options for the real fix

Both are already-proven shapes in this codebase; either closes it, and they are alternatives, not a sequence.

[...]

Issue #2601, opened 2026-07-16
Storage dedup: folder/path-tree usage rollup (per-prefix materialization + web tree view)


brandonrc:
Deferred sub-task 4 of epic #2056 (the repo-level dedup panel shipped in 1.6.0 via backend #2559 + web artifact-keeper-web#594). This is the folder/path-tree rollup: a `/storage/tree?prefix=` API backed by a second materialization keyed by path prefix (`repository_path_storage_stats`) + OCI image-name grouping, and a web tree view. It's a perf/design-heavy piece (per-format path semantics, sizing on large OCI repos) — belongs in the 1.7.0 performance milestone, not the 1.6.0 close-out. Refs #2056.

Issue #1660, opened 2026-06-05
Harden GC deletion: commit-before-delete ordering + push-path TOCTOU (#1655 follow-up)


brandonrc:
**Defense-in-depth follow-up from the #1655 blob-GC review** (non-blocking; current behavior matches the merged manifest-GC precedent).

1. **Delete ordering.** Both manifest GC (`run_gc`) and blob GC (`run_blob_gc`) delete the storage object BEFORE committing the DB row delete. If storage.delete succeeds but the commit fails/crashes, you get a dangling row pointing at a deleted object → `BLOB_UNKNOWN`/404 on pull. For irreversible deletion the safer bias is **commit the row delete first, then delete storage** (leak the object on crash rather than orphan a live row). Consider reordering both GC paths.
2. **Push-path TOCTOU.** The OCI push records `manifest_blob_refs` without `FOR UPDATE` on `oci_blobs`, so the GC's per-row lock doesn't block a concurrent ref insert. The 24h grace window covers the normal push window, but the residual sub-grace race is currently only documented in a code comment. Either guard it (take a row lock / re-check on the push side) or formally accept it with a test.

Both are bounded and consistent with existing GC; tracking so they're not lost in a doc comment.

PR #3850, opened 2026-09-12
test(oci): serialize OCI upload tests that share a content-addressed blob key

### LINKED ISSUE #3529: Flaky unit test: completion_with_unknown_length_empty_final_put_skips_zero_byte_part intermittently fails with 'blob storage was being reclaimed concurrently'

brandonrc:
## Summary

The flake is **test isolation, not a production race** — but the 503 is a real
guard doing its job, and the evidence for that is worth stating precisely.

`oci_upload_cleanup_keys.storage_key` is `UNIQUE` across the whole database and
`blob_storage_key(digest)` is content-addressed (`oci-blobs/<digest>`), so the
cleanup-journal row for a blob key is **global, not per-repository**. Six tests
in `oci_blob_upload_streaming_tests` complete a blob under bytes another test in
the same module also completes — `b"hello world"` in four of them, the empty
blob (`sha256:e3b0c442…`) in two — so they all register *the same row*: the
second `register_oci_upload_cleanup_key` hits `ON CONFLICT (storage_key)
DO UPDATE … RETURNING id` and gets the first test's row id back.

From there:

1. Whichever push commits first takes that row under `FOR UPDATE` and deletes it
   inside its `oci_blobs` transaction — the #3187 guard,
   `claim_cleanup_journal_row_for_blob_commit`, step 3.
2. That test then runs `f.teardown()`, which deletes its `oci_blobs` rows and its
   repository.
3. The slower push reaches the same guard with the now-stale `journal_id`, takes
   step 4 ("row absent"), and looks for an `oci_blobs` row referencing the key —
   the documented proof that "a concurrent push of the same key won". Teardown
   has just deleted it, so the guard concludes a sweep reaped the row, refuses to
   publish a blob row for bytes it believes are gone, and returns
   `503 BLOB_UPLOAD_INVALID` "blob storage was being reclaimed concurrently".

No GC is involved: `blob_gc_enabled` is `false` in the fixture config, none of
the tests in the reproduction run a sweep, and `ABANDONED_OCI_UPLOAD_TTL_SQL` is
`INTERVAL '24 hours'`, so a journal row seconds old is not a sweep candidate. The
only things that delete these rows in the suite are a peer push's own guard and a
peer fixture's `DELETE FROM repositories` cascade. The production code is
behaving as designed; what it cannot know is that the "reclaim" was another test
dropping its repository.

## Fix

Add `test_db_helpers::oci_blob_digest_serial_lock` — a Postgres **session**
advisory lock, the same shape as `blob_gc_serial_lock` (#1660),
`oci_reindex_serial_lock` (#3402) and `path_stats_serial_lock` (#2601), which is
this repo's existing answer to "tests contend for one global row" and works
across `cargo nextest`'s process-per-test parallelism where an in-process mutex
does not. The six tests take it as their first line, so only one runs its
push → assert → teardown critical section at a time.

No production code changed, no assertion relaxed, no retry loop, no `#[ignore]`,
no expected status changed.

The two empty-blob tests are included even though neither has failed in CI: they
share the identical hazard on `sha256:e3b0c442…` and reproduce it at a lower rate
(see below). Their digest is load-bearing (it is the config blob of essentially
every image manifest), so unique bytes are not available for them, which is the
reason the fix is a lock rather than per-test payloads.

## Verification

All runs against a local Postgres 16 with `AK_TESTS_REQUIRE_DB=1`. The focused
loop runs the four `b"hello world"` tests together at `--test-threads 4`, which
is the same contention CI produces incidentally with 16k tests in flight.

| Loop | Before (at `origin/main`) | After |
| --- | --- | --- |
| 4 `hello world` tests, `--test-threads 4` | **27 / 100 iterations failed** | **0 / 200**, and **0 / 150** on the final commit |
| 2 empty-blob tests, `--test-threads 2` | **1 / 100 iterations failed** | **0 / 200** |
| whole `oci_blob_upload_streaming_tests` module (55 tests), `--test-threads 16` | **1 / 60 iterations failed** | **0 / 60** |

Of the 27 baseline failures, 22 were the test named in the issue, with the exact
reported message and status:

```
assertion `left == right` failed: empty unknown-length final PUT should complete:
b"{\"errors\":[{\"code\":\"BLOB_UPLOAD_INVALID\",\"message\":\"blob storage was being reclaimed concurrently; retry the upload\"}]}"
```

The other failures were `completion_with_nonempty_final_put_body_concatenates_parts`
(14), `multi_patch_upload_rehashes_when_digest_cache_is_cleared` (24) and
`session_empty_blob_completion_creates_zero_byte_blob` — the same failure, same
message, on the siblings the issue asked about. The "before" numbers were taken
by stashing this branch's hunks and rebuilding, then restoring them.

Also run, all green:

- `cargo nextest run -p artifact-keeper-backend --lib` — 16292 passed, 8 skipped

## One thing this deliberately does not fix

The same shared-row mechanism implies a narrow **production** race, which this PR
does not address and which I do not think belongs in a test-isolation change:
because `register_oci_upload_cleanup_key` upserts on the global `storage_key`,
the `journal_id` a push captures is not necessarily "the row this push
registered" — the premise `claim_cleanup_journal_row_for_blob_commit`'s doc
comment states. Two clients pushing the same digest to different repositories
share one row, and the loser's step-4 peer-push proof depends on the winner's
`oci_blobs` row still existing at that instant. If that row is deleted in the
gap (repository deletion, blob GC), the loser gets a spurious
`503 BLOB_UPLOAD_INVALID` on a perfectly valid upload. Relatedly, step 4's
"was the repository deleted?" check tests the *pushing* repository, but the row
that cascaded away may have belonged to a different one.

Closing that properly means giving each push its own journal row — a schema
change to the `UNIQUE(storage_key)` constraint from #116 — which is a separate
issue, not a hunk in this one. Flagging it rather than folding it in.

Closes #3529


brandonrc:
Filed the production half as #3851 — the shared `UNIQUE(storage_key)` journal row meaning two concurrent pushes of one digest to different repositories can produce a spurious `BLOB_UPLOAD_INVALID`, plus step 4 checking the pushing repository rather than the row's owner. Agreed it needs a schema change and does not belong here.

PR #3871, opened 2026-09-14
fix(nuget): restore remote version discovery

### LINKED ISSUE #3870: Remote NuGet Version Discovery Fails for Paginated Registrations
### Description

Remote NuGet repositories support package restore, but NuGet clients cannot reliably discover current upstream versions. This breaks package-version selection in IDEs and causes `dotnet list package --outdated` to fail for packages with paginated NuGet registrations.

### Environment Information

- Deployment type: Helm chart deployment on Kubernetes
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL
- Reverse Proxy / Ingress: HAProxy Ingress Controller
- Network setup: Clients access the repository manager through the HAProxy Ingress endpoint. The deployment has outbound HTTPS access to NuGet.org.

### Backend Version

1.7.8

### Web UI Version

1.7.0

### Package Format

NuGet (.NET)

### Expected Behavior

For a remote NuGet repository, the repository manager should provide complete upstream package metadata so that:

- Visual Studio and Rider show available package versions.
- NuGet autocomplete works.
- `dotnet list package --outdated` completes and reports newer versions.
- Flat-container indexes include upstream versions even when older versions are locally cached.
- Paginated registration-page URLs returned to NuGet clients resolve through the repository manager.

### Observed Behavior

- `dotnet restore` succeeds through the remote NuGet repository.
- Flat-container metadata may omit upstream-only versions once a local cached artifact exists.
- The service index lacks required autocomplete and modern registration resource types.
- For packages with a large version history, such as `Serilog`, the registration index returns a rewritten page URL like:

  ```text
  /nuget/<repo-key>/v3/registration/serilog/page/0.1.6/1.2.47.json
  ```

  but the repository manager returns `404` for that page.

- Consequently, `dotnet list package --outdated` terminates with:

  ```text
  error: Object reference not set to an instance of an object.
  ```

The client exception follows the missing registration-page response; the server-side fault is the `404`.

### Steps to Reproduce

1. Create a remote NuGet repository with upstream:

   ```text
   https://api.nuget.org/v3/index.json
   ```

2. Configure a .NET project to use the repository manager as its only NuGet source.

3. Add older package versions, including a package with paginated registrations:

   ```xml
   <PackageReference Include="Dapper" Version="2.0.123" />
   <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="8.0.0" />
   <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
   <PackageReference Include="Serilog" Version="3.1.1" />
   ```

4. Restore:

   ```bash
   dotnet restore --configfile NuGet.Config --no-http-cache
   ```

5. Check updates:

   ```bash
   dotnet list <project> package --outdated --configfile NuGet.Config
   ```

6. Observe the null-reference error.

7. Inspect the registration index for `Serilog`, then request the `items[0]["@id"]` page URL it contains. The page request returns `404`.

### Logs

```shell

```

## Proposed backend fix

- Advertise `SearchAutocompleteService` and `RegistrationsBaseUrl/3.6.0`.
- Proxy autocomplete metadata.
- Merge upstream flat-container versions with locally cached versions.
- Add a guarded backend route to proxy rewritten registration subresources, including paginated page documents, preserving existing upstream authentication, SSRF protection, metadata limits, caching, and URL rewriting.

No Web UI changes are required.

### Pre-submission Checklist

luko0610:
## Summary
Closes #3870 

Fix NuGet V3 remote repository version discovery so Visual Studio, Rider, and `dotnet list package --outdated` can resolve current package versions through Artifact Keeper.

- Advertise and proxy `SearchAutocompleteService`.
- Advertise `RegistrationsBaseUrl/3.6.0`.
- Merge upstream flat-container versions with locally cached versions.
- Proxy and rewrite paginated registration pages so clients can follow registration index links.
- Support local autocomplete and merge local/remote autocomplete results for virtual repositories.
- Harden registration proxy paths by validating client-derived package IDs/subpaths and constructing upstream URLs with structured URL path segments.

## Validation

- `cargo test --test nuget_remote_version_listing_tests -- --ignored`
  - 6 passed
- Local NuGet CLI verification:
  - `dotnet restore --configfile NuGet.Config --no-http-cache`
  - `dotnet list NugetCliLocal package --outdated --configfile NuGet.Config`
  - Confirmed update discovery for Dapper, Microsoft.Extensions.Configuration.Json, Newtonsoft.Json, and Serilog.


brandonrc:
Updated this branch from `main` to clear the `🔒 Security Audit` failure — that one was never about your change. It was RUSTSEC-2026-0286 (cryptoki 0.12.0) inherited from a stale lockfile, fixed on `main` in 07feface. Security Audit now passes.

One real blocker is left, and it is a code gate rather than infrastructure:

```
New code coverage: 39% (204/512 new lines)
::error::New code coverage is 39%, below 70% threshold
```

The gate measures only the lines this PR adds or changes, not the whole file, so it needs roughly another 155 of those 512 lines covered. The ignorable-looking `role "root" does not exist` lines in that job's log are just the `pg_isready` healthcheck and unrelated — Postgres came up healthy and all 16,756 tests ran.

CLAUDE.md's note on this gate is the practical route: extracting the new logic into pure helper functions makes handler code coverable without standing up a full request path.

brandonrc:
Taking this over, since the 09-18 questions are unanswered and it now conflicts with main. Plan, after #4126 merges (it fixes a different NuGet bug in the same file and goes first):

1. Rebase and drop the virtual version-list rewrite, which #3981 already delivered on main.
2. Build the remote-repository version merge on #4126's `remote_upstream_versions`, with case-insensitive dedupe.
3. Autocomplete answers empty when the upstream advertises no autocomplete service, instead of 502, and honours `take`.
4. Move the DB-backed tests into `nuget.rs`'s test modules so the coverage gate measures them (it reads 39% today only because the separate test binary is not counted).
5. Add the CHANGELOG bullet for #3870.

The paginated registration-page route, the 3.6.0 registrations base URL and the autocomplete endpoint are all still needed and kept. Your commits and authorship stay on the branch.

brandonrc:
Rebased this onto main now that #4126 has merged, and pushed the takeover changes as one commit on top of your commits (authorship kept):

- Dropped the virtual flat-container rewrite. #3981 already merges a virtual repository's members on main. `fetch_upstream_flatcontainer_versions` went with it.
- A remote repository's version list now merges #4126's `remote_upstream_versions` (V2 or V3 upstream) into any cached versions, deduped case-insensitively through the same helper the virtual merge now uses. If the upstream fails, the cached versions are returned.
- Autocomplete: an upstream with no `SearchAutocompleteService`, which includes every V2 feed, now gets an empty result instead of a 502. `skip`/`take` are honoured: forwarded upstream, applied in SQL for hosted rows (with `totalHits` counting every match), and enforced on remote and merged virtual answers. An off-origin autocomplete host (nuget.org's azuresearch) is fetched without the repository's credentials, like search.
- Registration pages: kept as you wrote them, with the encoded-segment URL builder now shared with the index fetch. A V2 upstream, which links no pages, answers 404.
- The DB-backed tests moved from `backend/tests/nuget_remote_version_listing_tests.rs` into a `remote_discovery_tests` module in `nuget.rs`, so the new-code coverage gate counts them. The ci.yml wiring for the separate binary is gone. I added cases for `take`, the empty no-service answer, cached rows with a failing upstream, the V2 page 404, the virtual page walk, and anonymous off-origin autocomplete. The two virtual tests for the dropped rewrite weren't carried over; main's own tests cover that path.
- Added the changelog entry for #3870 as a `changes/unreleased/` fragment.

Thanks for the thorough report and the fix. The paginated registration pages were the real break for `dotnet list package --outdated`.

brandonrc:
I pushed one small commit to keep this mergeable. Main now requires every inline `#[cfg(test)]` module to carry a CI test-shard gate line (#4234), which splits the unit tests across five hosted runners. This PR's new test modules didn't have one yet, so Check Rust failed. I added them with `python3 scripts/ci/test-shards.py apply`. There are no code changes.

Issue #3676, opened 2026-09-02
CI: nothing rejects committed git conflict markers — CHANGELOG.md shipped on main with <<<<<<< / ======= / >>>>>>> for ~3 hours (#3675)


brandonrc:
## What happened

`d980ea92` (#3664) merged into `main` with `CHANGELOG.md` still carrying a three-line git conflict block between two `[Unreleased]` → `### Security` bullets. Every check was green: `scripts/ci/check-changelog-unreleased.sh` only asserts that the first `## [` heading is `## [Unreleased]`, `release-preflight.sh` check 5 reconciles bullets by `#NNNN` reference (both bullets were intact, so it would have passed too), and nothing else reads the file. Three later PRs inherited the markers through their merges of main before #3675 removed them. Had 1.9.0 been cut in that window, the promoted CHANGELOG section and the release notes would have contained the markers verbatim.

The same shape can hit any file — a `.sql` migration or `.rs` source would fail to compile, but Markdown, YAML comments, `.env.example`, Helm templates and workflow files would not.

## Ask

A repo-wide gate in the shell-tests job (cheap, no toolchain):

```
git grep -nE '^(<{7} |={7}$|>{7} )' -- . ':!**/*.diff' ':!**/*.patch' && { echo 'committed conflict markers'; exit 1; }
```

with the two fixture-style exclusions above (and any test fixture that deliberately contains markers — none exist today). Optionally teach `check-changelog-unreleased.sh` to assert that `## [Unreleased]` contains at most one of each `### ` heading, which would have caught the duplicate-heading shape the same merges nearly produced.

Refs: #3664, #3675, #3537 (the changelog-bookkeeping gates).

brandonrc:
Moved from 1.9.0 to 1.10.0 in the pre-release triage: 1.9.0 is being cut this week with the security and proxy-correctness fixes already on main, and this item does not change what ships in that release. It keeps its labels and priority.

PR #3872, opened 2026-09-15
fix(auth): apply AK_GUEST_ACCESS_ENABLED to the OCI Distribution surface

### LINKED ISSUE #3854: Anonymous /v2 pulls still succeed when guest access is disabled
`AK_GUEST_ACCESS_ENABLED=false` is documented as a server-wide policy that disables anonymous access. It does not hold on the OCI Distribution surface: an unauthenticated `docker pull` of a repository marked `public` succeeds on an instance that has guest access turned off.

### Cause

`guest_access_guard` allowlists the whole `/v2` subtree, not just the challenge endpoint:

`backend/src/api/middleware/guest_access.rs:87-98`

```rust
    | "/v2"
    | "/v2/"
    ...
    || path.starts_with("/v2/")
```

The trailing `starts_with("/v2/")` covers every manifest, blob and tag path, so the guard never runs on them. The OCI handlers then apply their own gate, which asks only whether the repository is anonymously readable — a question about the repository, not about the server policy. A `public` repository answers yes, and the pull is served.

The allowlist is deliberate and the comment explains why: OCI clients fan out from `/v2/` and need the `WWW-Authenticate` challenge to discover where to get a token. That reasoning covers `/v2` and `/v2/`. It does not cover the content paths, which is where it currently reaches.

### Blast radius

Limited to repositories whose visibility is `public`. `private` (and `internal`, once #3812 lands) are refused by the handlers' own anonymous gate regardless of this guard, so this is not a path to non-public content. It is the server-wide policy failing to apply to the content it was configured to withhold — an operator who set the flag to stop anonymous consumption still has anonymous `docker pull` working.

### Why this is not a one-line fix

Narrowing the allowlist to the exact challenge paths is the obvious direction, but two things have to come with it:

1. **The refusal has to be OCI-shaped.** The guard's 401 is a REST JSON body. `backend/src/api/middleware/oci_errors.rs:15` records explicitly that the guard is *not* routed through `oci_denied_response` precisely because it allowlists `/v2` rather than refusing it. Once it starts refusing, it needs the `{"errors":[...]}` envelope plus the `WWW-Authenticate` header, or docker and oras clients get an unrenderable failure instead of a credential retry.

2. **The current behaviour is pinned by tests.** `guest_access.rs:656 guard_allows_oci_v2_subpath_when_disabled` asserts that `/v2/library/nginx/manifests/latest` passes the guard while disabled. That is the bug, written down as an expectation. It needs replacing with the challenge paths staying open and a content path being refused.

There is also a question worth settling before implementing: should the guard refuse anonymous `/v2` content requests outright, or should the OCI handlers consult `guest_access_enabled` in their own anonymous gate? The first is one place and matches how every other surface is handled; the second keeps the challenge flow untouched at the cost of spreading the policy into handlers.

### Not this

Found while reviewing #3813 (repository `internal` visibility). That PR does **not** change this behaviour in either direction: `guest_access.rs` is untouched by its diff, and every anonymous gate it converted from `is_public` to `visibility.allows_anonymous_read()` decides identically, because migration 217 holds `CHECK (is_public = (visibility = 'public'))`. Its only widenings sit on authenticated paths (`oci_read_permitted`, the blob-mount source check), both behind a `claims` binding.

mymarche:
## Summary

Fixes #3854. `AK_GUEST_ACCESS_ENABLED=false` is documented as a server-wide policy that disables anonymous access, but it did not hold on the OCI Distribution surface: an unauthenticated `docker pull` of a `public` repository succeeded on an instance with guest access turned off, with nothing in the logs to say so.

The guest-access guard allowlisted the entire `/v2` subtree, so the policy was never asked on manifest, blob, tag or referrer paths. The OCI handlers then applied their own gate, which asks only whether the *repository* is anonymously readable — a question about the repository, not about the server.

**The allowlist was never a policy carve-out but a response-*shape* workaround.** The guard's REST-shaped 401 names a bare realm (`Bearer realm="artifact-keeper"`) that no container client can fetch a token from, plus a `Basic` challenge that raises a browser popup and a meaningless `Cargo` one. `oci_errors.rs` even recorded that the guard was deliberately *not* routed through `oci_denied_response` *because* it allowlisted `/v2` — a circular state. Breaking it in one direction removes the reason for the allowlist.

What changes:

- **The content surface is gated.** `/v2`, `/v2/` and every manifest, blob, tag and referrer path now inherit the policy by route, so a read verb that does not exist yet is covered without its author doing anything.
- **The refusal is OCI-shaped**: the distribution spec's `{"errors":[…]}` envelope with `WWW-Authenticate: Bearer realm="<base-url>/v2/token",service="artifact-keeper"`, and no `Basic` or `Cargo` challenge. The realm is derived through `request_base_url_from_request`, the same resolution the OCI handlers use, so a reverse-proxied instance points clients at the right host. The challenge builder moved into `middleware::oci_errors` so the guard and the handlers emit byte-identical values.
- **`/v2/token` stays allowlisted, and refuses the anonymous mint in the handler.** See below — this is the non-obvious part.
- `AuthOutcome::Overloaded` keeps returning its retryable `503` with `Retry-After` on `/v2` rather than collapsing into the new 401.

### Why `/v2/token` is not gated by the guard

The first cut of this change gated it too, reasoning that `docker login` presents Basic credentials the guard authenticates, so a real login passes and an anonymous token request does not. Login does pass. The pull after it does not:

```
GET  /v2/token?offline_token=true   Authorization: Basic admin:pass   -> 200, "Login Succeeded"
POST /v2/token                      grant_type=refresh_token          -> guard: NoCredential -> 401
                                    refresh_token=eyJ...                       (no Authorization header)
GET  /v2/token?scope=...            Basic ":<refresh_token>"          -> 401 (wrong token_type)
```

Docker requests an offline token at login and thereafter authenticates with the OAuth2 refresh grant, whose credential is a **form field**, not a header. The guard resolves credentials from headers, so gating this route refused authenticated pulls along with anonymous ones — breaking `docker login` + `docker pull` entirely on flag-off instances.

`/v2/token` serves no content; it is the endpoint by which credentials are *obtained*, the OCI analogue of `/api/v1/auth/login`, which the allowlist has always carried. The policy is applied inside `token()` instead, at the single one of its four token-issuing exits that requires no identity. The refusal there is byte-identical to the endpoint's existing invalid-credential 401, so an anonymous prober cannot learn the instance's configuration from the wire; the distinction is logged. The allowlist entry is an exact match, so `/v2/tokenX` and `/v2/token/<anything>` stay gated.

`security_regression_tests::guest_access_oci_3854` fails on `main` and passes here:
`regression_3854_anonymous_oci_read_is_refused_while_guest_access_disabled` (manifest, blob, tags, referrers, both flag states), `regression_3854_fabricated_anonymous_sentinel_is_refused_like_no_credential`, `regression_3854_anonymous_token_request_issues_no_token_while_disabled` (real router), and `regression_3854_credential_from_a_registry_login_still_pulls_while_disabled`, which drives login → refresh-grant exchange → manifest pull end to end and is the case that caught the `/v2/token` mistake above.

Two tests that pinned the old behaviour are replaced rather than deleted: `allowlist_oci_v2_challenge` and `guard_allows_oci_v2_subpath_when_disabled`.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes

---

## Risk 1 — client behaviour (verified on the wire, not simulated)

Local instance, flag off, `public` repository holding a multi-layer `nginx:alpine`:

| client | anonymous | authenticated |
|---|---|---|
| docker 28.0.4 | refused, rendered error | full multi-layer pull OK |
| podman 6.1.0 | refused, rendered error | login + pull OK |
| oras 1.2.3 | refused, spec message surfaced | manifest fetch OK |
| nerdctl 2.4.0-beta.0 / containerd v2.4.0-beta.0 | refused, fatal | login + pull OK (27.7 MB) |

No client loops: each issues exactly one token request and treats the 401 from the realm as terminal. The `ak` CLI has no `/v2` client code (it speaks `/api/v1` only), so its refusal is the guard's existing REST 401, unchanged.

## Risk 3 — doubled credential resolution (measured)

Same instance and image, flag toggled between runs. Bearer JWT — the shape docker actually sends on content paths. Debug build, local Postgres, 20 MB layer.

| request | flag ON (mean/p50/p95) | flag OFF (mean/p50/p95) | delta |
|---|---|---|---|
| GET manifest, n=100 | 2.94 / 2.70 / 3.61 ms | 4.10 / 3.82 / 4.74 ms | +1.16 ms |
| GET 20 MB blob, n=100 | 10.17 / 10.19 / 11.15 ms | 11.05 / 10.77 / 12.13 ms | +0.88 ms |

Full `docker pull`, 5 runs each: flag ON mean 2150 ms, flag OFF mean 2118 ms — the two overlap, the difference is inside run-to-run noise.

**No mitigation was needed.** The cost is two uncached DB queries per request in the guard's JWT path (`is_token_invalidated_replica_safe` + `fetch_live_is_admin`). Worth stating plainly rather than burying: because it is DB round-trips and not CPU, it scales with database latency, not request size — on an instance whose Postgres sits at 1–3 ms RTT the same two queries cost ~2–6 ms per `/v2` request, so a 50-layer pull would add roughly 100–300 ms. That is the same per-request cost every non-`/v2` surface already pays; `/v2` simply never paid it. If an operator sees it, the answer is caching the guard's resolution, not reopening the allowlist.

## Local gate

`cargo fmt --check` clean; `cargo clippy --workspace --all-targets -- -D warnings` clean; `cargo nextest run --workspace --lib --bins` → 16378 passed, 45 failed. All 45 are wiremock-backed remote-proxy tests (ansible/helm/hex/npm/pypi/vscode, image_scanner, proxy_service) that need a reachable non-loopback interface, which this machine's network blocks. Verified pre-existing: the failing set on this branch is a strict *subset* of the same run on a clean tree (which additionally flaked `scheduler_service::test_storage_gc_tick_follow_on_abandons_remaining_workloads_on_lease_loss_3502`). The Tier 1 integration targets pass 59/59.

Both merge gates were measured locally rather than left to CI:

| gate | threshold | measured |
|---|---|---|
| coverage on new/changed lines | >= 70% | **98%** (405/413) |
| total line coverage floor | >= 50% | **88.48%** |
| duplication on changed files | <= 3% | **0%** (`Found 0 clones`) |

Coverage was computed with the gate's own algorithm (parse `git diff -U0` from the merge-base, intersect added lines with `lcov.info` from a `--lib --bins` run): `oci_errors.rs` 80/80, `guest_access.rs` 176/177, `oci_v2.rs` 149/156. All 8 uncovered lines are the `return;` arm of `let Some(pool) = try_pool().await else { return; }` skip guards, which do not execute when a database is present. Duplication was measured by replaying `scripts/ci/jscpd-prepare-sources.py` + `.jscpd.json` over the changed non-exempt files (`guest_access.rs`, `oci_errors.rs`, `config.rs` — `oci_v2.rs` is on the format-handler ignore list).

## Not addressed

`/v2/token` has no IP rate limiting — all three rate-limit middlewares are mounted inside `api_v1_routes`, so `/v2` never sees them, and password guessing there is bounded only by account lockout. That predates this change and is unaffected by it; worth a separate issue.


brandonrc:

The only conflict was `CHANGELOG.md` — your entry and the ones on `main` both target the top of the same section. Both were kept verbatim, yours first; no code was touched.

This branch was the stalest of the batch (`main` moved by ~700 files since your merge-base) and it shares `api/handlers/oci_v2.rs`, `api/middleware/guest_access.rs`, `config.rs` and `.env.example` with changes that landed in that window, so the rebase was verified rather than assumed: `cargo clippy --workspace --all-targets -- -D warnings` is clean on the rebased commit. Worth re-reading the merged result of those four files once CI is green, since a clean compile is not the same as your intent surviving a semantic overlap — particularly in `guest_access.rs`, which is the middleware this PR is about.

This also clears the `🔒 Security Audit` failure, which was never about this PR: RUSTSEC-2026-0286 (cryptoki 0.12.0) came from the stale lockfile on the old base, and `main` fixed it in 07feface.

Tracking the underlying churn in #4016.

brandonrc:
Correction to my previous rebase — apologies, that push was wrong and I've fixed it.

My CHANGELOG conflict resolution kept both sides of the hunk wholesale, which re-introduced a second `### Security` heading under `[Unreleased]` and with it five entries (Swagger, SBOM, logout, HuggingFace, name validation) that were already present in `main`'s Security section. `🐚 Shell Tests` caught it via `check-changelog-unreleased.sh` — working exactly as #3676 intended.

The branch now carries only your own #3854 entry, placed in the existing Security section in Keep-a-Changelog order. `CHANGELOG.md` diffs against `main` at +5 lines, nothing else. Your commit, authorship and all seven changed files are untouched — verified `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` clean on the amended commit, and the three changelog gates (`unreleased`, `placement`, `conflict-markers`) now pass locally.

The earlier note still stands: this branch was the stalest of the batch and overlaps `oci_v2.rs`, `guest_access.rs`, `config.rs` and `.env.example` with changes that landed since your merge-base, so please do re-read the merged result of those once CI is green.

brandonrc:
Taking this over to unblock it — thanks for a genuinely excellent write-up, especially the `/v2/token` refresh-grant finding, which is the part that would have bitten anyone doing this from scratch. Rebased onto current `main`; your commit applied with no conflicts and is preserved unchanged, with the fixes in a separate commit on top (2dbc91d3).

Three things changed. (1) The CHANGELOG hunk opened a second `### Security` section at the top of `[Unreleased]` and re-added the four bullets already in the existing one, which fails `check-changelog-unreleased.sh`'s duplicate-heading assertion; your #3854 bullet and upgrade note moved verbatim into the existing section. (2) `guard_refuses_fabricated_anonymous_bearer_when_disabled` needs a database after all: `Bearer anonymous` isn't a JWT, so it only becomes `InvalidCredential` once the API-token lookup reaches Postgres, and without one the guard sheds 503; it now skips on `try_pool()` like its siblings and like your own integration twin, which had this exactly right. (3) When `OCI_TOKEN_SERVICE` moved to `oci_errors.rs`, its doc block stayed behind and silently became the doc comment of `struct TokenResponse`; restored it onto the constant.

The security review came out clean: the content surface is fully gated, the `/v2/token` carve-out is sound at all four of `token()`'s exits, and the `/v2/` refusal is byte-identical to what `version_check` already emitted. All six `guest_access_oci_3854` regression tests ran against a live Postgres. Your note about `/v2/token` having no IP rate limiting is right and gets its own issue.

brandonrc:
Correction to the above: you rebased onto main while the review was running, so my push was refused by the lease (nothing of yours was overwritten). Re-applying the same three fixes on top of your new head now; the SHA in the previous comment is superseded.

brandonrc:
Re-applied on top of your rebased head: 84d3c025 (your commit unchanged, one takeover commit with the DB-gated test and the two doc-comment moves; your rebase had already fixed the CHANGELOG placement, so that part is dropped). Merging once CI is green.

Issue #3797, opened 2026-09-09
CI: fail a PR whose CHANGELOG bullet lands inside an already-released section


brandonrc:
Four PRs hit this today, one of which reached `main` before I caught it.

A branch opened before a release cut adds its bullet under `## [Unreleased]`. The cut renames that heading to `## [X.Y.Z] - <date>` and starts a fresh empty `[Unreleased]`. When the branch later merges `main`, git sees the bullet attached to a heading that still exists, so **the merge is clean and the bullet silently ends up inside a shipped release**. Nothing warns: not the merge, not `check-changelog-unreleased.sh` (which only asserts an `[Unreleased]` section exists), not the release preflight (check 5 matches the bullet by PR/issue number wherever it sits).

Observed: #3570 (merged to main with its bullet inside `[1.9.0]`; fixed by #3795), #3619, #3627, #3628 — all four with their bullets inside a released section after a clean merge.

Ask: a CI check on pull requests that computes the bullets added by the PR relative to its merge base and fails when any of them lands under a `## [` heading other than `## [Unreleased]`. Roughly:

```sh
base=$(git merge-base HEAD origin/main)
git diff "$base" HEAD -- CHANGELOG.md | grep '^+- \*\*'   # bullets this PR adds
# for each, find its containing `## [...]` heading in the PR's CHANGELOG.md and require [Unreleased]
```

Two related traps worth covering in the same check: removing a bullet can leave an empty `### Fixed`/`### Changed` heading behind (that happened on `main` in #3795 and was cleaned up in #3791), and a bullet whose first reference is not a `#NNNN` trips release preflight check 5 much later, at cut time, when it is expensive to fix.

PR #3875, opened 2026-09-15
fix(scanner): add Incus construction switch

### LINKED ISSUE #3874: Incus scanner dispatches archive extraction on Remote OCI pulls and cannot be disabled
### Problem Statement

**What happens**
On a deployment that scans through the Trivy scanner adapter and never handles Incus/LXC images, the Incus scanner is still registered and can still be dispatched.

Its applicability check looks only at the artifact path. A Remote OCI proxy pull whose tag ends in an archive suffix produces a synthetic artifact path that satisfies that check, so the Incus scanner is selected and begins archive extraction — spawning tar — before the manifest bytes fail to extract.

**Reproduction**
1. Configure an OCI Remote repository and the Trivy scanner adapter (TRIVY_ADAPTER_URL), with TRIVY_URL and OPENSCAP_URL unset.
2. Pull an image whose tag ends in an archive suffix, for example release.tar.gz.
3. The proxy scan constructs the synthetic artifact path v2/library/example/manifests/release.tar.gz.
4. The Incus scanner's path-only applicability check accepts that path and dispatches archive extraction.

**What I expected**
A deployment that does not accept Incus images should be able to keep the Incus scanner out of the registry entirely, and should not spawn archive-extraction child processes for artifacts that are not Incus images.

**Why it matters**
The extraction attempt fails, so this is not a correctness bug in scan results — but it means an operator who does not use Incus cannot prevent Incus archive handling or its local child-process execution from being reachable. For deployments that scan only through the remote adapter, that is reachable surface with no configuration to remove it.

There is currently no setting that disables the Incus scanner. TRIVY_URL and OPENSCAP_URL already control whether the local-CLI Trivy and OpenSCAP scanners are constructed; Incus has no equivalent.

### Proposed Solution

An opt-out setting evaluated where the scanner registry is constructed, defaulting to enabled so existing deployments are unaffected, following the same absent/empty/unknown-enables convention used elsewhere in the configuration.

I have a change ready for this and will open a pull request referencing this issue.

### Alternatives Considered

_No response_

### Use Case

_No response_

### Component

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

PhilLehmann:
This PR closes #3874 

The Incus applicability check is path-only, so a Remote OCI proxy can route an archive-suffixed tag into Incus extraction, which may spawn undeclared tar, unsquashfs, xz, or zstd children before extraction fails.

Reproduction:

1. A Remote OCI proxy synthesizes v2/<image>/manifests/<tag>.
2. A tag ending in .tar.gz therefore looks like an Incus archive path.
3. Incus accepts the path and starts archive extraction before engine scanning.
4. Extraction spawns tar (or a compression helper) and then fails on the manifest bytes.

A deployment that runs Trivy as a remote service and scans no Incus images wants this child-process path absent. ScannerService::new, isolated in its construction module, now mints a private-field IncusScannerCapability only when incus_scanner_enabled is true; both public Incus constructors require it, covering adapter and legacy local-CLI branches while preventing direct producer construction. The default remains enabled. The regression drives scan_content_target for both branches; the current supplementary-error contract cannot observe an ignored-result direct call, which remains a named residual.


brandonrc:
Taking this over so it can land in 1.10.0. The Security Audit red was never about your change: the branch's lockfile predated #3964, and merging main picks up main's `Cargo.lock` untouched (#3999 has since made that job advisory on PRs that don't touch the lockfile anyway). You merged main yourself while the review ran, so the takeover commit now sits on top of your merge; your code is unmodified.

The capability approach is the right call: gating at construction rather than at `is_applicable_for_target` means an adapter-only deployment cannot reach Incus archive extraction at all, and I verified there are no other `IncusScanner::new*` call sites that could route around it. The default-enabled positive control in the registry test is what makes it safe for deployments that do use Incus. 803 scanner tests pass against a live Postgres.

One change from me: your CHANGELOG bullet had no issue number, so `check-changelog-placement.sh` warned that release-preflight check 5 couldn't reconcile it against a commit (#3797). Rewritten in the surrounding style: it names #3874, the Remote OCI archive-suffix path, the three files that document `INCUS_SCANNER_ENABLED`, and that the default is unchanged. It stays under `### Added`, since #3874 is filed as an enhancement. Merging once CI is green.

Issue #3848, opened 2026-09-12
JetBrains multipart publish corrupts the plugin zip: offsets are computed on a lossy UTF-8 copy and used to index the raw body


brandonrc:
Noticed while fixing the Swift multipart publish bug (#3595), which has the same shape but a different cause. Confirmed by reading the code; not yet reproduced against a running instance, so the failure mode below is derived rather than observed — but the defect is unambiguous.

## The bug

`extract_plugin_from_multipart` (`backend/src/api/handlers/jetbrains.rs:623`) parses a `multipart/form-data` body by converting the **whole binary body to a string**:

```rust
let body_str = String::from_utf8_lossy(body);
let parts: Vec<&str> = body_str.split(&boundary_marker).collect();
```

and then recovers the file bytes by computing an offset in that string and indexing the **original byte slice** with it (`:665-676`):

```rust
let header_offset = part.as_ptr() as usize - body_str.as_ptr() as usize;
let body_offset = header_offset + idx + header_body_split.len();
let end = header_offset + part.len();
...
file_bytes = Some(Bytes::copy_from_slice(&body[body_offset..end]));
```

A JetBrains plugin is a zip, so the body is never valid UTF-8. That breaks the code in two ways at once:

1. **The two coordinate systems don't agree.** `from_utf8_lossy` replaces each invalid sequence with U+FFFD, which is three bytes in UTF-8. So `body_str` is *longer* than `body`, and every offset past the first invalid byte is shifted. Offsets derived from `body_str` do not address the same content in `body`.
2. **When the input is invalid UTF-8, `Cow` is `Owned`** — a separate allocation. `part.as_ptr() - body_str.as_ptr()` is then an offset into that copy and has no defined relationship to `body` at all. (It happens to work for a valid-UTF-8 body, where `Cow` is `Borrowed` and the pointers coincide — which is why any test using an ASCII payload passes.)

The bounds check `body_offset <= body.len() && end <= body.len()` catches only the case where the shifted offsets run off the end. When it fires, the fallback is worse: `Bytes::copy_from_slice(content.as_bytes())` stores the **lossy** bytes, i.e. the zip with every invalid sequence replaced by U+FFFD.

## Why it is bad

[...]

PR #3936, opened 2026-09-16
fix(upload): answer 413, not 400, when a multipart envelope exceeds MAX_UPLOAD_SIZE

### LINKED ISSUE #4023: Oversized multipart publishes (JetBrains, Swift) answer 400 instead of the documented 413
### Summary

`POST /jetbrains/{repo}/plugin/uploadPlugin` and `PUT /swift/{repo}/{scope}/{name}/{version}` parse `multipart/form-data` with `multer` under a `whole_stream` size limit equal to `max_upload_size_bytes` (`stage_plugin_from_multipart` in jetbrains.rs, `stage_publish_multipart` in swift.rs). Both handlers' doc comments say the ceiling surfaces as "413 mid-stream", and a raw (non-multipart) upload of the same size does answer 413 from `stage_stream_content_addressed`. The multipart path answers **400** instead.

### Why

`multer` counts the whole envelope and raises `Error::StreamSizeExceeded` as soon as the cumulative stream exceeds the limit (`StreamBuffer::poll_stream`), so it always trips before the stager's own per-part count can (it measures the envelope, the stager measures one part). The error is then mapped to 400 on both surfaces it can appear on:

- from `next_field()`, when a small body arrives in one piece: the handlers' `bad_request` closures map every `multer::Error` to 400 (`Malformed multipart/form-data upload: stream size limit exceeded`);
- from `Field::poll_next`, when a chunked body crosses the ceiling mid-part: the stager maps any chunk error to 400 (`Failed to read upload body: …`), and its `written > max` check never fires because `multer` stopped yielding first.

### Observed

`main` @ 3f0143c, `max_upload_size_bytes = 512`, a 4 KiB file part, calling the staging helpers directly with a lazy pool:

| ingest shape | status |
| --- | --- |
| multipart, body delivered whole | 400 |
| multipart, body delivered in 64-byte chunks | 400 |
| raw body (`X-Plugin-Name`), 64-byte chunks | 413 |

### Expected

413 Payload Too Large on both multipart surfaces, matching the raw path and the handlers' own comments. The nuget push is unaffected: its `Multipart::new` carries no whole-stream constraint, so the stager's 413 is the only ceiling there.

### Ask

Map `multer::Error::StreamSizeExceeded` to 413 in one shared place, used by the jetbrains and swift handlers and inside `stage_stream_content_addressed` for the mid-part surface, with a regression test per delivery shape. Found while rebasing #3936 onto the #3942 fix for #3848; a fix is on that PR.

ThaSami:
## Summary

Closes #4023. Follow-up to the #3848 fix that landed in #3942; this PR was rebased onto `main` and slimmed to what #3942 does not cover (see the thread for what was dropped).

**Fix.** The JetBrains plugin upload and the Swift publish parse `multipart/form-data` with `multer` under a `whole_stream` ceiling equal to `max_upload_size_bytes`. That ceiling always trips before `stage_stream_content_addressed`'s own per-part count can (it measures the envelope, the stager measures one part), and `multer`'s size-limit error came back as a "malformed multipart" **400** on both surfaces it can appear on: from `next_field()` when a small body arrives whole, and through the stager's read-failure mapping when a chunked body crosses the ceiling mid-part. A raw upload of the same size answers 413, which is also what both handlers' doc comments promised. A shared `proxy_helpers::multipart_error` now maps `StreamSizeExceeded` to 413 (swift wraps it in its `application/problem+json` envelope, jetbrains uses the plain-text `multipart_error_response`), and the stager downcasts a chunk error to `multer::Error` so the mid-part surface answers the same 413. The stager's `E` bound gains `'static`, which every existing caller's error type already satisfies. nuget is untouched: its `Multipart::new` has no whole-stream constraint.

**Extra pins on the #3942 path.** A byte-level round trip of the raw `X-Plugin-Name` upload (the catalog test only checks that it succeeds), and a database-free binary-fixture check of `stage_plugin_from_multipart` so the #3848 invariant also runs in the pre-push hook, where the router-level test skips without `DATABASE_URL`.

`test_stage_plugin_from_multipart_oversized_envelope_is_413` (jetbrains.rs) and `publish_multipart_oversized_envelope_is_413` (swift.rs) each drive the staging helper with a 4 KiB part under a 512-byte ceiling, once as a single `Bytes` and once as 64-byte chunks. With only the test snippets applied on `main` @ 3f0143c, both fail on the first (whole-body) assertion; a raw body of the same size answers 413 through the stager, for contrast:

```
test api::handlers::swift::multipart_publish_tests::publish_multipart_oversized_envelope_is_413 ... FAILED
test api::handlers::jetbrains::tests::test_stage_plugin_from_multipart_oversized_envelope_is_413 ... FAILED
  left: Some(400)
 right: Some(413)
PROBE multipart single-chunk=Some(400) multipart 64B-chunks=Some(400) raw 64B-chunks=Some(413)
```

Also added: `test_stage_plugin_from_multipart_stages_binary_zip_verbatim` (DB-free) and `test_jetbrains_raw_upload_round_trips_binary_zip` (DB-backed).

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


brandonrc:
Thanks for this, and apologies for the overlap: #3848 was fixed and merged in #3942 while this was open (the JetBrains publish path now stages the multipart body to a scratch file and streams it into storage, so the lossy-UTF-8 offset bug is gone), which is why this branch shows as conflicting. Rather than close it I have moved it to 1.11.0. If you rebase onto main, what looks worth keeping is the additive part: backend/tests/streaming_invariant.rs, if it still asserts something the merged path does not already cover. Happy to review a slimmed-down version.

ThaSami:
Thanks, and no problem about the overlap. Rebased onto `main` and slimmed down.

**Dropped as covered by #3942:** the handler rewrite, the CHANGELOG bullet for #3848, and the `backend/tests/streaming_invariant.rs` change. #3942 already removed the jetbrains.rs row from `RAW_BODY_BLOB_ALLOWLIST` (with its own doc paragraph), so nothing of mine was left to keep there.

**What this PR is now:** while re-verifying the merged path I found that an envelope over `max_upload_size_bytes` answers 400 rather than the 413 the comments (and the raw path) promise. `multer`'s `whole_stream` limit always trips before the stager's per-part count, and its `StreamSizeExceeded` was mapped to a "malformed" 400 both from `next_field()` and, for a chunked body, through the stager's read-failure mapping. Filed as #4023 with the probe output; the fix here is a shared `proxy_helpers::multipart_error` used by jetbrains and swift plus a downcast in `stage_stream_content_addressed` for the mid-part surface, with a regression test per delivery shape on each handler. On top of that, two small pins of the #3942 path: a byte-level round trip of the raw `X-Plugin-Name` upload, and a DB-free binary-fixture check of the parser so the #3848 invariant also runs in the pre-push hook.

If you would rather take the 413 fix separately from the extra tests, or skip the tests entirely, say so and I will trim further.

ThaSami:
@brandonrc ^

brandonrc:

Issue #3363, opened 2026-08-14
cache-correctness and concurrency e2e harnesses abort at repo create: SSRF guard rejects the docker-internal mock upstream


brandonrc:
## Summary

Both observability e2e suites named in #3310 are still red on `main`, and for the same reason — one that has nothing to do with the bugs they were written to reproduce. Each harness aborts in its setup phase when it tries to create a `remote` repository pointing at its docker-internal mock upstream, and the backend rejects the URL with HTTP 400.

Because both suites die before their first assertion, they currently measure nothing: not the fixed bugs (#1611, #1609), not regressions. This is why their `continue-on-error: true` cannot simply be removed yet (#3310): flipping them to blocking today would page on a harness defect, not a product defect.

## Evidence

**Cache-Correctness E2E** — run 31358654047 (schedule, 2026-08-10), job log:

```
==> Phase 1: Creating repositories (local + remote->mock + virtual)...
  ERROR: create_repo maven-vr-remote returned HTTP 400 (body: {"key":"maven-vr-remote",...,"repo_type":"remote","upstream_url":"http://cache-mock-upstream:9101/maven2"})
  Aborting: subsequent tests depend on this repo existing.
...
e2e-test-virtual-resolution exited with code 1
e2e-test-cache-correctness exited with code 1
```

The workflow still reports `success` because the compose invocations end in `|| echo "::warning::..."` and the job has `continue-on-error: true`.

**Concurrency E2E (single-flight race)** — the `Run the single-flight race driver` step has concluded `failure` in every scheduled run since at least 2026-07-06 (runs 28784651420, 29239116957, 29729638922, 30255675180, 30802635078, 31366802423), each with:

```
ERROR: repo create returned HTTP 400
##[error]Process completed with exit code 2.
```

(`scripts/concurrency/run-singleflight-race.sh:147`.)

## Root cause

The anti-SSRF validation on remote-repository upstream URLs (`validate_outbound_url`, `Upstream` context, `backend/src/api/validation.rs`) resolves DNS-name hosts and rejects any that resolve to private/internal addresses. Docker-internal service names like `cache-mock-upstream` resolve to the compose network's RFC1918 address, so `create_repo` with `"upstream_url": "http://cache-mock-upstream:9101/..."` returns 400 `VALIDATION_ERROR`.

[...]

PR #3938, opened 2026-09-16
chore(test): Cargo.toml scan regression test, no-fail-fast unit run, one test password, truthful redteam reporting, e2e SSRF allowance

### LINKED ISSUE #3490: chore: remove hardcoded admin password from docker-compose.test.yml and e2e scripts
`docker-compose.test.yml` sets `ADMIN_PASSWORD: TestRunner!2026secure` in the clear, and the same value is the default in `scripts/e2e-setup.sh` and multiple `scripts/native-tests/*` env blocks.

The in-repo redteam suite flags this as CRITICAL (`default-creds/admin-login`) — noisy but directionally right: anyone who copies the test compose as a deployment starting point ships known credentials.

Suggestions:
- Require the variable (no default) in the compose file, or generate it per run (`openssl rand` in the setup script and pass through).
- At minimum add a prominent "NOT for deployment" comment near the value, mirroring the warning style in the root `docker-compose.yml`.

### LINKED ISSUE #3491: test: redteam suite reports false CRITICALs, malformed report JSON, broken counters
Ran `docker compose -f docker-compose.test.yml --profile redteam up` against 1.8.0. The suite passes overall, but the harness itself has bugs that make it untrustworthy as a CI gate:

1. **False CRITICALs in the gRPC tests** (`tests/06-grpc-unauth.sh`): `grpc/reflection-enabled`, `grpc/schema-exposed`, `grpc/sbom-list-noauth`, `grpc/sbom-get-noauth` are all reported CRITICAL while their own evidence strings say `server does not support the reflection API` — i.e. the server correctly refused, and the test scores the refusal as a vulnerability. (Manually verified with the real proto files: unauthenticated `SbomService/ListSbomsForArtifact` returns `Unauthenticated`. gRPC auth is fine.)
2. **Malformed report JSON**: `/results/redteam-report.json` is missing commas between finding objects; `jq` cannot parse it.
3. **Pass/fail counters reset per test script** (each test runs in a subshell, `lib.sh` counters don't persist): final summary printed `Passed: 1 checks` after dozens of PASS lines. Exit-code gating on this is unreliable.
4. **Later tests skip due to the login rate limiter**: tests 14/15 report "Could not authenticate — skipping" because earlier tests exhaust `RATE_LIMIT_LOGIN_PER_WINDOW` (10/900s). The suite should authenticate once and share the token, or raise the limit in the e2e env. (The limiter working is good; the suite tripping over it hides coverage.)

Happy to contribute fixes if useful.

### LINKED ISSUE #3699: Cargo.toml dependency scanning silently returned nothing under toml 1.x (parse::<Value> became a value-expression parser) — and CI's unit job skips the test group that catches it
## What happened

Dependabot's toml 0.8 → 1.1 bump (#3690) failed three `test_extract_dependencies_cargo_toml*` tests. The root cause is a silent semantic change upstream: toml 1.0's `impl FromStr for Value` is the **value-expression** parser (`Value::deserialize(ValueDeserializer::parse(s))`), where 0.8's was the **document** parser (`crate::from_str`). So `content.parse::<toml::Value>()` in `DependencyScanner::parse_cargo` (`backend/src/services/scanner_service.rs:~4652`) still compiled but returned `Err` for every real manifest, and the `if let Ok(..)` swallowed it — **zero crates.io dependencies reported for every `Cargo.toml` scanned**, a scanner false-clean of the #3603 shape. Fixed in #3690 by using `toml::from_str::<toml::Value>` (the document parser on both versions) and logging the error at debug.

Two things that deserve their own follow-through:

1. **CI's unit-test job does not run the `services::scanner_service::tests::test_parse_cargo_*` group.** The failing run executed 13,613 of 16,045 tests; that group was excluded, yet those tests were also broken on the branch. The fixer ran them explicitly. Find out which selector/partition excludes them (`.config/nextest.toml` partitions? the `--test-threads`/filter in `ci.yml`?) and either include them or document why — a scanner-correctness test that never runs is #3494's shape.
2. **`.clippy.toml` pins `msrv = "1.75.0"`** while `toml` 1.1.3 requires `rust-version = 1.85` and the workspace is on 1.98. The pin only shapes clippy's MSRV-conditional suggestions, but it is stale and will drift further.

Refs: #3690, #3603, #3494.

brandonrc:
## Summary

Four test/dev hygiene issues in one PR.

**#3699 — Cargo.toml dependency scanning returned nothing under toml 1.x.** The parser fix itself landed on `main` in #3690; this adds the regression test that would have caught it — a whole manifest (`[package]` plus all three dependency sections) asserting on the *names and versions* yielded, so a parser that silently returns nothing cannot pass — and a canary pinning the `str::parse` vs `toml::from_str` split. On the issue's two follow-ups: no CI selector excluded `test_parse_cargo_*` (`cargo nextest list --workspace --lib --bins` lists all 16,471, that group included); the "13,613 of 16,045" was nextest's default fail-fast cancelling the run and reporting everything unreached as skipped. The Tier 1 unit step now passes `--no-fail-fast`. `.clippy.toml` `msrv` moves 1.75.0 → 1.98.0 (the pinned channel) and `check-toolchain-pin.sh` fails when the two disagree (four new self-test cases). The msrv bump unmasked 25 MSRV-gated lints (`is_none_or`, `is_multiple_of`, `repeat_n`, `inspect_err`) across 18 backend files, applied with `cargo clippy --fix` — all one-liners, full suite green.

**#3363 — the cache-correctness and concurrency e2e harnesses aborted at repo create on the SSRF guard.** `docker-compose.test.yml` and `docker-compose.concurrency-e2e.yml` set `UPSTREAM_ALLOW_PRIVATE_IPS: "true"` on their own backend service (with comments on why the narrower `AK_SSRF_ALLOW_PRIVATE_CIDRS` is not usable: both networks are plain bridges with no pinned subnet). **Nothing in the guard changed** — production behaviour, the loopback/link-local/metadata hard blocks and the startup warning are untouched. `run-singleflight-race.sh` prints the response body instead of a bare "HTTP 400". `Part of`, not `Closes`: the issue sequences a `workflow_dispatch` run before removing `continue-on-error`; that dispatch happens after this merges, and the waivers now say exactly that.

**#3490 — hardcoded admin password.** All ~70 occurrences across 8 compose files and ~40 scripts resolve one variable, `AK_TEST_ADMIN_PASSWORD`; `run-e2e-tests.sh` generates one per run (`openssl rand -hex 16`) and prints how to reuse it; unset, everything falls back to `ak-e2e-throwaway-not-for-deployment`. Documented in `TESTING.md`.

**#3491 — redteam suite reporting.** Counters and the findings separator live in `${RESULTS_DIR}/.state` (the cause: `run-all.sh` invokes each script rather than sourcing it); every report field is jq-escaped and `finalize_report` proves the JSON parses or fails the run; gRPC refusal detection is case-insensitive (`server` vs `Server`) and `backend/proto` is mounted so the method probes invoke, reporting SKIPPED when they cannot; test 05 no longer calls a login with the stack's configured password "default credentials"; one login per run shared via the state dir so the rate limiter no longer starves tests 13–15. `scripts/ci/test-redteam-report.sh` (new, 12 cases) fails **10 of 12** against `main`'s `lib.sh`/`run-all.sh`, reproducing all four reported defects.

## Verification

- `cargo fmt --check`; `cargo clippy --workspace --all-targets -- -D warnings` clean.
- `cargo nextest run --workspace --lib --bins --no-fail-fast`: **16,473 run, 16,473 passed**, 8 skipped.
- `test-redteam-report.sh` 12/12 (10/12 fail on `main`); `test-check-toolchain-pin.sh` 21/21; `check-toolchain-pin.sh`, `check-changelog-unreleased.sh` pass; `scripts/concurrency/test_assert_lib.sh` 19/19; `shellcheck -S warning` clean on every touched script; all 8 compose files `config`-parse with and without `AK_TEST_ADMIN_PASSWORD`; `actionlint` finding set identical to `main`.
- Not run: the e2e harnesses themselves (each needs a built test image and a multi-container stack); the SSRF fix is verified as configuration, and the dispatch runs are the remaining evidence for #3363.

Closes #3699
Closes #3490
Closes #3491
Part of #3363


brandonrc:
Merged with GitGuardian red: both findings are false positives. .env.test line 32 is the documented throwaway default (ak-e2e-throwaway-not-for-deployment, overridden by exporting AK_TEST_ADMIN_PASSWORD), and scripts/run-e2e-tests.sh line 112 generates a random per-run password with openssl rand. The two dashboard incidents (37346847, 37346848) can be dismissed as test data.

Issue #3933, opened 2026-09-16
storage stats: recompute_all fails with repository_storage_stats_repository_id_fkey when a repository is deleted mid-recompute


brandonrc:
Observed while fixing #3733 (pre-existing; reproduced by running `repositories::tests::delete*` in parallel with `scheduler_service::tests::storage_gc_tick_follow_on_recomputes_stats_and_keeps_blob_gc_dry_run_db`).

`StorageStatsService::recompute_all` snapshots the repository ids and then INSERTs into `repository_storage_stats`, with no tolerance for a repository deleted between the two — the INSERT then violates `repository_storage_stats_repository_id_fkey` and the whole recompute fails. `reconcile_all_usage_ledgers` already skips a vanished repository; `recompute_all` should do the same (per-repository `ON CONFLICT`/`WHERE EXISTS` guard, or catch the FK violation for that row and continue).

Refs: #3733, #3650.

brandonrc:
Duplicate of #3957, fixed on `main` by #3965 (df5cf24f).

That commit applies exactly the guard this issue asked for: `persist` now upserts via `INSERT ... SELECT ... FROM repositories WHERE id = $1::uuid` (backend/src/services/storage_stats_service.rs:583) so a repository deleted mid-recompute is skipped atomically, and the per-prefix `oci_blobs` aggregate got the matching `WHERE EXISTS (SELECT 1 FROM repositories r WHERE r.id = b.repository_id)` guard (:544). The flaky `storage_gc_tick_follow_on_*` scheduler tests this issue named as the reproducer are covered by the regression test that landed with it.

Closing as a duplicate; removing it from 1.11.0.

Issue #3595, opened 2026-08-31
Swift publish endpoint stores the entire multipart body instead of the source-archive part


azeek:
## Summary

Publishing a package with the official SwiftPM command succeeds, but Artifact Keeper stores the entire `multipart/form-data` request body as the source archive instead of extracting the required `source-archive` part.

As a result, `GET /swift/{repo}/{scope}/{name}/{version}.zip` returns a body that starts with a multipart boundary and headers rather than ZIP magic bytes, and `swift package resolve` fails while extracting the archive.

This was reproduced with Apple Swift 6.3.3. The current `main` implementation appears to have the same issue.

## Reproduction

1. Create a normal Swift package.
2. Publish it to a hosted Artifact Keeper Swift repository using the standard client:

```bash
swift package-registry publish \
  --netrc \
  example.ExamplePackage \
  1.0.0 \
  --url https://artifacts.example.com/swift/swift-hosted
```

3. Download the published source archive:

```bash
curl --netrc \
  -H 'Accept: application/vnd.swift.registry.v1+zip' \
  -o release.zip \
  https://artifacts.example.com/swift/swift-hosted/example/ExamplePackage/1.0.0.zip

xxd -l 192 release.zip
```

## Actual behavior

The response is reported as `Content-Type: application/zip`, but its body begins like this:

```text
--<multipart-boundary>
Content-Disposition: form-data; name="source-archive"
Content-Type: application/zip
Content-Transfer-Encoding: binary

PK...
```

In the observed response:

- the first ZIP local-file header (`PK\x03\x04`) began 163 bytes into the response;
- the closing multipart boundary remained after the ZIP end-of-central-directory record;
- the release metadata checksum and download `Digest` matched the complete multipart body, not the inner ZIP;
- extracting just the inner ZIP bytes produced a valid archive with `unzip -t`;
- `swift package resolve` failed with an extraction diagnostic reporting `163 extra bytes at beginning or within zipfile`.

## Expected behavior

The publish endpoint should parse the multipart request and persist only the bytes of the required `source-archive` part. Downloading `{version}.zip` should return a ZIP body beginning with `PK`, and the reported checksum should be computed from that ZIP.

The Swift Package Registry specification requires publish requests to be multipart with a required `source-archive` part:

https://github.com/swiftlang/swift-package-manager/blob/main/Sources/PackageManagerDocs/Documentation.docc/RegistryServerSpecification.md#create-a-package-release

## Root cause in the current implementation

[...]

Issue #3592, opened 2026-08-28
Header X-Requested-With is forced but not implemented for upload in web interface.


TheBr00m:
### Description

Created a generic local repository. Forbidden (403) is returned when trying to upload a file through the web interface. The following message is part of the response:

```
Cookie-authenticated state-changing requests must send the X-Requested-With header (CSRF protection). Use a Bearer or API token for non-browser clients.
```

Work around:
```
curl -X POST https://{AK_FQDN}/api/v1/repositories/assets-local/artifacts \
  -H "X-Requested-With: XMLHttpRequest" \
  -H "X-API-KEY: ${API_KEY}" \
  -F "file=@{LOCAL_FILE_PATH}" \
  -F "path={REMOTE_FILE_PATH}"
```

### Environment Information

- Deployment type: docker compose
- Operating System: debian 13
- Architecture: x64
- Database: postgres
- Reverse Proxy: traefik

### Backend Version

1.8.1

### Web UI Version

1.7.0

### Package Format

N/A / General

### Expected Behavior

Upload through web interface is successful.

### Observed Behavior

Upload through web interface failed.

### Steps to Reproduce

1. Create local generic repository.
2. Create a user and group, add user to group, and create read/write permission for group on repository.
3. Login as user
4. Navigate to repository, then to upload tab
5. Try to upload a local file through the web interface and choose a nested path such as /a/b/file.txt

### Logs

```shell
artifact-keeper-backend  | 2026-08-28T10:19:48.558807Z DEBUG http_request{method=GET uri=/api/v1/tree?repository_key=assets-local&include_metadata=true correlation_id=027491fa-ec8a-4c1c-9e83-f613f8237088}: tower_http::trace::on_response: finished processing request latency=5 ms status=200
artifact-keeper-backend  | 2026-08-28T10:19:51.397387Z DEBUG http_request{method=POST uri=/api/v1/repositories/assets-local/artifacts}: tower_http::trace::on_request: started processing request
artifact-keeper-backend  | 2026-08-28T10:19:51.397523Z  INFO http_request{method=POST uri=/api/v1/repositories/assets-local/artifacts correlation_id=6555d098-f5d9-496f-9c62-d624cd71d321}: artifact_keeper_backend::api::middleware::tracing: Request completed correlation_id=6555d098-f5d9-496f-9c62-d624cd71d321 status=403
[...]

PR #3942, opened 2026-09-16
fix(maven,jetbrains,terraform,upload,csrf): five contained format and upload bugs

### LINKED ISSUE #3587: Possible duplicate artifact creation under upload pressure causes UNIQUE constraint violation

### LINKED ISSUE #3588: Terraform Provider Mirror with authentication fails with 401 when Terraform downloads provider archives
### Description

According to the Artifact Keeper [documentation](https://artifactkeeper.com/docs/guides/infrastructure/#provider-mirror-configuration), it is possible to use a Terraform Provider Mirror with authentication.

I configured Terraform CLI to use Artifact Keeper as a provider mirror and specified both the mirror URL and authentication token in .terraformrc:

```hcl
provider_installation {
  network_mirror {
    url = "https://<artifact_keeper_url>/terraform/terraform-remote/"
    include = ["registry.terraform.io/*/*"]
  }
}
credentials "<artifact_keeper_url>" {
  token = "..."
}
```

The mirror URL and credentials work correctly when Terraform retrieves the provider index and package metadata.
For example, terraform init successfully retrieves:

```
GET https://<artifact_keeper_url>/terraform/terraform-remote/registry.terraform.io/nicola-preda/artifact-keeper/index.json
```
and then:

```
GET https://<artifact_keeper_url>/terraform/terraform-remote/registry.terraform.io/nicola-preda/artifact-keeper/1.8.2.json
```
However, when Terraform tries to download the provider package, it receives `401 Unauthorized`:

```
GET https://<artifact_keeper_url>/terraform/terraform-remote/registry.terraform.io/nicola-preda/artifact-keeper/1.8.2/download/linux/amd64
Error: Failed to install provider
Error while installing nicola-preda/artifact-keeper v1.8.2:
<artifact_keeper_url>: bad response code: 401
```

**Terraform Network Mirror Protocol**

The issue appears to be related to the authentication behavior defined by Terraform’s Network Mirror Protocol.
Terraform [documentation](https://docs.hashicorp.com/terraform/internals/provider-network-mirror-protocol?utm_source=chatgpt.com#authentication) states that credentials are not sent when retrieving provider archives whose URLs are returned in the package metadata response:

Terraform does not send credentials when retrieving the archives whose URLs are given in the “List Available Installation Packages” response below.

It also states that if the distribution packages themselves are sensitive, the mirror should provide cryptographically secure, user-specific, and time-limited URLs.

In this case, Artifact Keeper requires authentication for the provider download endpoint, while Terraform does not send the configured credentials for the actual provider archive download.

The metadata returned by Artifact Keeper contains relative download URLs, for example:

```json
{
  "archives": {
    "linux_amd64": {
      "hashes": [
        "zh:746a99ae009105c691868ec0104c801358d0f03a5e108d3bf80f2ea4d749484e"
      ],
      "url": "1.8.2/download/linux/amd64"
    }
  }
}
```

Terraform resolves this to:

```
https://<artifact_keeper_url>/terraform/terraform-remote/registry.terraform.io/nicola-preda/artifact-keeper/1.8.2/download/linux/amd64
```
This endpoint returns `401 Unauthorized` because authentication is required.

### Environment Information

- Deployment type: Kubernetes Rancher RKE2 2.13.0
- Architecture: x86_64
- Database: PostgreSQL 18.6
- Reverse Proxy: Ingress Nginx 1.15.1
- Network setup: Cilium CNI

### Backend Version

1.7.1

### Web UI Version

1.8.0

### Package Format

Terraform

### Expected Behavior

If Artifact Keeper supports an authenticated Terraform Provider Mirror, Terraform should be able to download provider packages from the mirror after successfully authenticating against the mirror.

Possible approaches could include:

* Artifact Keeper generating authenticated, user-specific, time-limited download URLs in the provider metadata, as suggested by the Terraform Network Mirror Protocol.
* Providing another mechanism compatible with Terraform’s authentication model for protected provider archives.
* Documenting the required configuration or limitation if authenticated provider downloads are currently not supported.

### Observed Behavior

Terraform return `401 Unauthorized`

### Steps to Reproduce

1. Create a Terraform remote repository.
2. Specify the mirror URL in the Terraform configuration.
3. Create a minimal Terraform configuration that specifies any provider.

### Terraform Logs

```shell
2026-08-26T15:42:21.806+0500 [INFO]  Terraform version: 1.15.5
2026-08-26T15:42:21.806+0500 [DEBUG] Attempting to open CLI config file: /home/valeksandrov/.terraformrc
2026-08-26T15:42:21.806+0500 [INFO]  Loading CLI configuration from /home/valeksandrov/.terraformrc
2026-08-26T15:42:21.807+0500 [DEBUG] Explicit provider installation configuration is set
2026-08-26T15:42:21.807+0500 [TRACE] Selected provider installation method cliconfig.ProviderInstallationNetworkMirror("https://<artifact_keeper_url>/terraform/terraform-remote/") with includes [registry.terraform.io/*/*] and excludes []
...
2026-08-26T15:42:21.808+0500 [DEBUG] Querying available versions of provider registry.terraform.io/nicola-preda/artifact-keeper at network mirror https://<artifact_keeper_url>/terraform/terraform-remote/
2026-08-26T15:42:21.808+0500 [DEBUG] GET https://<artifact_keeper_url>/terraform/terraform-remote/registry.terraform.io/nicola-preda/artifact-keeper/index.json
...
2026-08-26T15:42:22.204+0500 [DEBUG] Finding package URL for registry.terraform.io/nicola-preda/artifact-keeper v1.8.2 on linux_amd64 via network mirror https://<artifact_keeper_url>/terraform/terraform-remote/
2026-08-26T15:42:22.204+0500 [DEBUG] GET https://<artifact_keeper_url>/terraform/terraform-remote/registry.terraform.io/nicola-preda/artifact-keeper/1.8.2.json
...
2026-08-26T15:42:22.681+0500 [TRACE] providercache.Dir.InstallPackage: installing registry.terraform.io/nicola-preda/artifact-keeper v1.8.2 from https://<artifact_keeper_url>/terraform/terraform-remote/registry.terraform.io/nicola-preda/artifact-keeper/1.8.2/download/linux/amd64
2026-08-26T15:42:22.682+0500 [TRACE] HTTP client GET request to https://<artifact_keeper_url>/terraform/terraform-remote/registry.terraform.io/nicola-preda/artifact-keeper/1.8.2/download/linux/amd64
...
│ Error: Failed to install provider
│
│ Error while installing nicola-preda/artifact-keeper v1.8.2: <artifact_keeper_url>: bad response code: 401
```

### Additional Context

_No response_

### Pre-submission Checklist

### LINKED ISSUE #3839: Maven PUT path re-derives snapshot mutability instead of using classify_maven, so unique timestamped snapshots are silently overwritable
Found while reviewing #3814 (see https://github.com/artifact-keeper/artifact-keeper/pull/3814#pullrequestreview). Reproduced against a real Postgres, not read off the code.

## The divergence

The native Maven PUT path decides mutability for itself instead of asking the classifier:

- `backend/src/api/handlers/maven.rs:3135` keys on `coords.version.contains("SNAPSHOT")`.
- `backend/src/services/cache_classifier.rs:298-330` (`classify_maven`) deliberately classifies a **resolved unique** snapshot leaf as `Immutable` via `is_unique_snapshot_artifact`, and uses component-wise `ends_with("-snapshot")` at `:340` rather than `contains`.

Two consequences fall out of that:

1. **A unique timestamped snapshot artifact can be silently overwritten through the Maven PUT route**, even though `classify()` calls it `Immutable`. Probed: `com/example/probe/6.0-SNAPSHOT/probe-6.0-20260827.132833-10.jar` uploaded twice with different bytes returns `201`, `201`, and the row is replaced with no conflict. Every deploy is supposed to mint a *new* timestamped filename, so an overwrite of an existing one is either a client bug or a collision — silently accepting it destroys the artifact a build may already have resolved and pinned.
2. **`contains` is broader than the classifier's component match.** A version like `1.0-SNAPSHOT-rc1` is treated as a snapshot by the handler and as a release by `classify_maven`, so the two halves of the system disagree about whether that coordinate is immutable.

## Why now

This predates #3814 — the preflight `DELETE` at `maven.rs:3022-3030` already had this behaviour. It is being filed separately because #3814 promotes "SNAPSHOT ⇒ mutable" from an incidental delete into the authoritative `ON CONFLICT` statement and into the CHANGELOG, which makes it a documented rule rather than an accident. #3587's own log line is exactly such a timestamped path.

Note the direction matters here: #3295 fixed the opposite error (a *non-unique* snapshot wrongly rejected as immutable). The fix for this one must not reintroduce that — `is_unique_snapshot_artifact` is precisely the distinction that keeps both correct, which is the argument for calling the classifier rather than re-deriving the predicate.

## Suggested fix

Have `maven.rs` call `cache_classifier::classify_maven` (or `classify`) for the upload path so there is one definition of Maven mutability, and return `409` for an `Immutable` coordinate the way the release branch already does. If the PUT path genuinely needs a different rule from the cache path, that difference should be named and commented rather than implied by two separate predicates.

## Acceptance

- Re-uploading a resolved unique snapshot (`…-20260827.132833-10.jar`) with different bytes is rejected, and the stored bytes are unchanged.
- Re-uploading a non-unique snapshot (`…-1.0-SNAPSHOT.jar`) with different bytes still succeeds — the #3295 behaviour is preserved.
- `1.0-SNAPSHOT-rc1` is treated identically by the handler and by `classify_maven`.
- A regression test covering all three.

### LINKED ISSUE #3848: JetBrains multipart publish corrupts the plugin zip: offsets are computed on a lossy UTF-8 copy and used to index the raw body

brandonrc:
## Summary

Five contained bugs, each with a regression test demonstrated failing on the pre-fix code.

**#3839 — Maven PUT re-derived snapshot mutability.** New `cache_classifier::maven_coordinate_is_republishable()` shares `has_snapshot_component` (component-wise, #3554) and `is_unique_snapshot_artifact` (#3459) with `classify_maven`; the PUT path calls it instead of `version.contains("SNAPSHOT")`. Deliberately not `!classify().is_immutable()`: `classify_maven` falls back to *mutable* for an unrecognised extension so the proxy revalidates, and inheriting that would make every unknown extension under a release version overwritable. A resolved unique-timestamped snapshot now gets 409; a non-unique snapshot (#3295) still republishes.

**#3587 — duplicate artifact creation under upload pressure.** The preflight hard-`DELETE` is gone; the insert is `ON CONFLICT (repository_id, path) DO UPDATE` with the update set `ArtifactService` already uses (sha1/md5 refreshed in lockstep with sha256, `is_deleted = false`), gated `WHERE $12::boolean` on the #3839 republishability flag and read with `fetch_optional`, so an immutable coordinate that loses the race gets the same 409 as the duplicate check rather than an overwrite or a 500.

**#3848 — JetBrains multipart corrupted the plugin zip.** The hand-rolled lossy parser is replaced by `multer` driven off the request **body stream** (`body: Body`, `Multipart::with_constraints(body.into_data_stream(), …)`), with the plugin part spooled through `proxy_helpers::stage_stream_content_addressed` (bounded scratch file, incremental SHA-256/SHA-1/MD5, 413 mid-stream at `max_upload_size_bytes`) and written with `put_artifact_stream` — the same conversion swift.rs got in #3595 — so the `jetbrains.rs` row *leaves* `RAW_BODY_BLOB_ALLOWLIST` in `tests/streaming_invariant.rs` instead of gaining an exemption; missing boundary, missing/duplicated `file`/`plugin` part, and truncated bodies are 400. Finding on the way: on a genuinely binary payload the old `&body[end-2..end]` probe ran before its bounds check and **panicked the request** (`range start index … out of range`), so some plugins were never stored at all. Plugins already published through this path must be republished (CHANGELOG).

**#3588 — Terraform provider mirror 401 on archive download.** The protocol's own remedy: `mirror_version` mints one single-use, path-bound `download_ticket` per platform for the caller who fetched the packages document and embeds it as `?ticket=` in each `archives[].url`, reusing the `?ticket=` credential `repo_visibility_middleware` already accepts for reads. `create_download_ticket_with_ttl` gives mirror tickets 600 s (the 30 s column default is the very failure being fixed); anonymous callers get the unchanged bare relative URL. The ticket appears in access logs as `create_download_ticket`'s doc already warns — single-use, read-only, path-bound, admin elevation dropped.

**#3592 — `X-Requested-With` forced but not implemented for uploads (backend half).** `violates_csrf_contract` now also accepts the browser's own `Sec-Fetch-Site: same-origin` / `none`. What the backend now accepts: a cookie-authenticated state-changing browser request passes if it carries `X-Requested-With` (any value) **or** `Sec-Fetch-Site: same-origin`/`none`; it is still refused for `cross-site`, `same-site` (sibling subdomain — a genuine CSRF position), and no Fetch Metadata at all (plain-HTTP non-localhost, older browsers). Token/Basic clients were never in scope. The web UI upload therefore works unchanged on a normal HTTPS deployment, but the web repository should still attach `X-Requested-With` to every state-changing request for the non-Fetch-Metadata cases — `Part of`, with the web half to be relayed.

## Verification

- `cargo clippy --lib --bins --tests -- -D warnings`, `cargo fmt --check` clean.
- `test(maven) or test(jetbrains) or test(terraform) or test(upload) or test(artifact_service)`: **1095/1095**; `test(csrf) or test(cache_classifier) or test(ticket) or test(auth)`: 1120/1120 (fresh PG16, `AK_TESTS_REQUIRE_DB=1`).
- **All five reverts demonstrated**: #3839 → `201` vs expected `409` on the timestamped snapshot; #3848 → both tests panic at the old slice probe; #3587 → `500 DATABASE_ERROR`, the issue's exact symptom; #3592 → same-origin asserts `true` vs `false`; #3588 → archive URL without a ticket.
- Full `--lib` suite: one failure in `scheduler_service::tests::storage_gc_tick_follow_on_…` — pre-existing flake (fails identically on the unmodified baseline, passes in isolation; #3933).
- Not exercised: the ticketed archive URL through `repo_visibility_middleware` end to end (the test routers inject auth directly); the test asserts the ticket is live, owned by the caller and bound to the exact path, and the middleware's own ticket tests cover redemption.

Closes #3839
Closes #3848
Closes #3588
Closes #3587
Part of #3592


brandonrc:
Rebased onto main after #3906 (error-text sweep) merged. One conflict, in backend/src/api/handlers/jetbrains.rs upload_plugin: #3906 swapped the buffered put's error text for storage_err_message, while this PR replaces that whole block with the streamed put_artifact_stream. Took the streamed path. So the jetbrains upload does not lose #3906's sanitised text, put_artifact_stream in proxy_helpers.rs now maps its storage error through storage_err_message instead of internal_error (one closure; every caller of the helper benefits, and the raw error is still logged at ERROR by the helper). No other logic changed. New head 66afba95.

PR #3943, opened 2026-09-16
feat(lifecycle): add explicit repository policy assignments

### LINKED ISSUE #3794: Lifecycle policies can only target one repository or all of them — add opt-in assignment to a subset
### Problem Statement

A lifecycle policy can only be scoped two ways today: **one repository, or every repository**. There is no way to target a subset.

This is baked into the data model. `lifecycle_policies.repository_id` is a nullable FK with no join table:

```
           Column     |  Type  | Nullable |      Default
  --------------------+--------+----------+-------------------
   id                 | uuid   | not null | gen_random_uuid()
   repository_id      | uuid   |          |            <-- nullable, no join table
   ...
  Foreign-key constraints:
      "lifecycle_policies_repository_id_fkey" FOREIGN KEY (repository_id)
          REFERENCES repositories(id) ON DELETE CASCADE
```

and made explicit in `backend/src/services/lifecycle_service.rs`:

```rust
pub(crate) enum CascadeScope {
    /// Run against every repository (`policy.repository_id IS NULL`).
    Global,
    /// Run against the named repository only.
    PerRepo(Uuid),
}
```

with every executor gating on `($1::UUID IS NULL OR a.repository_id = $1)`.

So `repository_id IS NULL` does not mean "unassigned" or "a default" — it means **run against everything**. A policy created without picking a repository is immediately live against every repository in the instance, including ones created later. There is no opt-out, no per-repository acknowledgement, and no "assigned to N repositories" view from the repository side.

The practical consequence is that a rule wanted on a subset has to be duplicated per repository. Applying one retention rule to 12 of 50 repositories means creating 12 near-identical policies, then editing all 12 whenever the rule changes and remembering to add a 13th when a repository is onboarded. Miss one and it silently keeps whatever it had; the only signal is storage drifting or artifacts vanishing.

For a destructive feature, the default direction feels backwards: the *safe* option (scoped) is the laborious one, and the *dangerous* option (everything, forever, including future repositories) is what you get by leaving a field blank.

### Proposed Solution

Make assignment explicit and many-to-many.

**1. A checkbox at policy creation** — something like *"Automatically apply to all repositories"*.

- Checked: current global behaviour, including repositories created later. Worth making the blast radius visible at this point, e.g. "will apply to 47 repositories".
- Unchecked: the policy exists but is dormant until assigned. This is the state that has no representation today.

**2. Assignment from the repository view** — on a repository, list the policies attached to it and attach/detach existing policies. This is the natural place an operator asks "what will delete things in this repo?", and today that question has no single answer — it requires reading every global policy plus any policy whose `repository_id` matches.

**3. A join table** to back it, since the current column cannot express a subset:

```sql
CREATE TABLE lifecycle_policy_repositories (
    policy_id     uuid NOT NULL REFERENCES lifecycle_policies(id) ON DELETE CASCADE,
    repository_id uuid NOT NULL REFERENCES repositories(id)       ON DELETE CASCADE,
    PRIMARY KEY (policy_id, repository_id)
);
```

with `lifecycle_policies.applies_to_all boolean NOT NULL DEFAULT false` carrying the checkbox.

Migration can preserve existing behaviour exactly: `repository_id IS NULL` becomes `applies_to_all = true`; a non-NULL `repository_id` becomes a single join row. `CascadeScope` then gains a third variant (`Repos(Vec<Uuid>)`) or resolves to a repository list before execution, and the existing `($1::UUID IS NULL OR a.repository_id = $1)` gate generalises to `= ANY($1)`.

**Relationship to #2024:** that issue proposes declarative scoping via a `match` block inside the policy config (`match.repository`, tag/format patterns). This request is the complementary axis — explicit assignment of a policy to named repositories, managed from the repository side. Pattern matching answers "which artifacts within scope", assignment answers "which repositories are in scope at all", and an operator auditing a single repository needs the second. They could reasonably be designed together; flagging it so the two are not solved in ways that conflict.

### Alternatives Considered

1. **Duplicate the policy per repository** — works today, and is what we would have to do. It scales badly and drifts: N copies to keep in sync, silent divergence when one is missed.
2. **Rely on #2024's `match.repository` pattern** — would help if repository names encode the grouping (`ci-*`), but naming rarely lines up with retention intent, and a regex over names is an implicit, fragile membership rule. It also still leaves the repository view unable to answer "what applies here".
3. **Leave global as the default and document it** — does not address the core problem, which is that there is no way to express a subset at all.

### Use Case

We run a self-hosted instance repurposed onto an existing artifact host, mixing very different retention needs in one instance:

- **CI build output** — per-commit images and snapshot builds, high churn, should be aggressively cleaned (delete after ~14 days).
- **Release artifacts** — signed releases and SBOM inputs that must be retained for audit and rollback, and must never be touched by an automated delete.
- **Proxy caches** — upstream mirrors, cheap to refetch, want their own independent rule.

A single retention rule is correct for the CI group and actively harmful for the release group. Today, creating the CI rule without selecting a repository would apply it to the release repositories too, so the only safe path is duplicating it once per CI repository and maintaining that list by hand forever.

What we want to express is simply: *"this 14-day rule applies to these five CI repositories, and to nothing else — including repositories added next month."*

The checkbox matters as much as the assignment. Right now the destructive-by-default option is the one you get by not making a choice, and on a registry the cost of that mistake is deleted artifacts rather than a broken deploy.

### Component

Backend / API

### Additional Context

Verified against backend 1.7.1 (chart 1.9.8, self-hosted Kubernetes, external PostgreSQL 18):

- `GET /api/v1/admin/lifecycle` supports a `repository_id` filter, and `POST` accepts `repository_id`, so the **single**-repository case is reachable from the API even where the UI does not surface it. The subset case is not expressible at any layer.
- The UI half of this lands in `artifact-keeper-web`, but the model change has to happen here first, so filing against the backend. Happy to open a companion issue there once the shape is agreed.

Prior art worth comparing:

- **Harbor** — tag retention rules are configured *per project*, so scope is inherently bounded and visible where the artifacts live.
- **Artifactory** — cleanup policies select repositories explicitly (include/exclude lists) rather than defaulting to everything.
- **GitLab / GitHub Packages** — cleanup policies are enabled per project/repository, off by default.

In all three the destructive default is opt-in, which is the property this request is really after.

Separately, and already tracked in #1850: `max_versions` and `size_quota_bytes` are documented as hard-requiring `repository_id` and fail at runtime when it is NULL, so those two types are effectively per-repo-only despite the UI allowing a global policy to be created. An explicit assignment model would let that be a validation error at creation time instead of a runtime failure.

### Pre-submission Checklist

damjanek:
Fixes #3794.
Tracking issue: https://github.com/artifact-keeper/artifact-keeper/issues/3794

## Summary

Backend portion of #3794. Introduce explicit cleanup-policy scope and reusable
many-to-many repository assignments instead of implicit global scope.

- New policies default to unassigned and ineffective. Applying to all current
  and future repositories requires explicit `applies_to_all: true`.
- Support atomic assignment replacement and idempotent attach/detach. Repository
  deletion retains reusable policies; the final detach leaves a dormant policy.
- All six policy types evaluate independently per resolved repository, including
  versions, quotas, exclusions, and OCI cascade cleanup. Scheduled and execute-all
  selection skip unassigned policies.
- Migration 219 preserves existing global/single-repository scopes and retains
  the deprecated singleton `repository_id` projection for legacy readers.
- Add admin-only capability detection, assignment endpoints, OpenAPI coverage,
  operational documentation, and scope/concurrency/migration regressions.
- Preserve the exclusions and dry-run byte behavior from #3858.

Companion UI branch:
https://github.com/damjanek/artifact-keeper-web/tree/cleanup-policy-controls
at `d275b08f8d22099c0635817584026c1375c8f9f0`.
Completion of #3794 requires both implementations; this backend PR alone does not
deliver the complete user-facing feature. Companion PRs are linked in a separate comment.

**Rollout:** Stop old backend instances and lifecycle schedulers before migration.
Upgrade all backend instances before enabling assignment UI writes. Do not mix
old/new binaries or perform a direct binary downgrade: old binaries interpret null
as global. Downgrade requires explicit scope reconciliation. Existing stored scopes
are preserved, but NEW legacy creates with omitted/null scope now become dormant.
See `docs/lifecycle-policy-assignments.md`.

Eleven new regression tests cover explicit scope. The OCI fixture models a stale
tag predating its artifact tombstone; production protection for newer re-pushes
is unchanged.

## Test Checklist
- [ ] Manually tested locally

Local runs overlap; counts must not be added as unique tests.

| Check | Result |
| --- | --- |
| `cargo fmt --check` | Passed |
| `cargo clippy --workspace --all-targets -- -D warnings` | Passed |
| Native lifecycle/scheduler/integration selection | 277 passed, including all 11 new assignment regressions and 23 ignored integration cases |
| Instrumented lifecycle/scheduler and binary selection | 289 passed |
| Instrumented ignored lifecycle integration selection | 23 passed |
| OpenAPI selection | 40 passed |
| Changed executable-line coverage | 509/515 = 98.83%, above 70% |
| Changed production-file duplication | 2.41%, below 3% |
| Full native workspace library/binary run | 16,430 passed, 46 failed, 8 skipped |

Validation used disposable local PostgreSQL with `AK_TESTS_REQUIRE_DB=1` and
`SQLX_OFFLINE=true`; migrations through 219 applied. No production resources were
used. Coverage is targeted, against baseline
`92df1f236887f5673f671506710b1422297400a4`; the full-project 50% coverage floor is
unmeasured. No live browser/backend HTTP E2E is claimed. Tests exercise real admin
middleware, handlers, and SQL. The optional live smoke was not run because the
server also binds gRPC to all interfaces; unrelated startup code was not changed.

<details>
<summary>Full-suite baseline/environment limitations</summary>

No lifecycle/assignment tests failed. Failures: Ansible 6, Composer 1, Helm 2,
Hex 1, NPM 4, PyPI 14, VSCode 16, image scanner 1, proxy service 1.
Unchanged baseline `92df1f236887f5673f671506710b1422297400a4` reproduced all five
selected representative cases under the same local environment:

- `api::handlers::ansible::tests::test_remote_collection_info_serves_upstream_metadata_3365`:
  mock upstream routing returned 502 rather than 200 under the VPN.
- `api::handlers::composer::metadata_db_tests::test_build_remote_dist_target_happy_path`:
  local DNS resolved the public GitHub API to a non-global address, correctly
  rejected by existing SSRF validation.
- `api::handlers::hex::tests::test_hex_same_version_case_variants_advertise_one_downloadable_release_db`:
  case-insensitive filesystem collided the foo/Foo fixture paths.
- `services::image_scanner::tests::test_scan_request_registry_url_substitutes_for_remote_adapter_on_the_wire`:
  expected mock-adapter submit request was not received.
- `services::proxy_service::tests::test_lazy_catalog_backfill_permit_is_acquired_then_released`:
  invalid-host pool did not finish in the wait budget; 7 rather than 8 permits.

These modules are unchanged. The five baseline reproductions do not mean all 46
failures were individually rerun on baseline. Network/filesystem settings, SSRF
protection, and unrelated product code were not changed to force green results.
Upstream CI remains required.
</details>

## API Changes
- [ ] N/A - no API changes

The named OpenAPI check passed through native nextest. All routes remain admin-only
under `/api/v1/admin/lifecycle` (anonymous 401, non-admin 403).

| Surface | Contract |
| --- | --- |
| `GET /capabilities` | `{"explicit_repository_assignment":true}`; positively check before new scope writes |
| Policy | Required `applies_to_all`, sorted unique `repository_ids`, deprecated singleton `repository_id` |
| Scope | false + [] dormant; false + IDs selected; true + [] all current/future; true + IDs rejected 422 |
| `POST /` | Dormant default; legacy non-null `repository_id` accepted alone as a singleton |
| `PATCH /{id}` | Omitted scope preserved, array atomically replaces assignments, [] detaches all, new-field null rejected |
| `GET /?repository_id={uuid}` | Bare array of effective global/assigned policies, including disabled |
| `PUT /{id}/repositories/{repository_id}` | Bodyless idempotent attach; 200 full policy |
| `DELETE /{id}/repositories/{repository_id}` | Bodyless idempotent detach; 200 full policy |
| Errors | 404 missing policy/repo, 422 global association mutation, 409 bounded contention |
| Preview/execute | Policy-wide, scope snapshot and per-repository grouping/cascade |

Null in the deprecated projection never means global. Detaching affects subsequent
runs, not cancellation of in-flight work. Dormant enabled policies return zero
counts/bytes without artifact, tag, or bookkeeping writes.


damjanek:
## Companion cleanup-policy PRs

- Backend: https://github.com/artifact-keeper/artifact-keeper/pull/3943
- Frontend: https://github.com/artifact-keeper/artifact-keeper-web/pull/848

Together these implement artifact-keeper/artifact-keeper#3794. Upgrade all backend instances before enabling the new UI writes; mixed old/new backend binaries are not supported.

brandonrc:
Audit done; the design is sound and every route is admin-gated with real DB-backed tests, so I am taking this the last mile on the branch rather than waiting. Pushed on top of your commits, unchanged:

- rebased onto main (no conflicts);
- migration renumbered to 233, since 219 is taken on main by `219_oci_gc_candidates.sql` (sqlx would refuse to start), with the test `include_str!` path and the doc reference updated;
- CHANGELOG bullet under Unreleased with a bold upgrade note: old binaries treat a policy with no single `repository_id` as global, so a rolling or mixed-version deploy could delete across every repository; stop old instances before migrating; plus the two API behaviour changes (no `repository_id` now means dormant, and deleting a repository leaves its policies dormant);
- a comment in the per-repository loop stating that one repository's failure aborts the rest of the run.

Locally: 213 lifecycle and assignment unit tests and the 23-test integration target pass against a migrated database, clippy and fmt clean. Merging once CI is green on this head. Follow-ups worth separate issues, not blocking: rejecting virtual repositories on attach, and lifecycle respecting quarantine.

PR #3948, opened 2026-09-17
fix(auth): mint tokens after the in-process invalidation watermark

### LINKED ISSUE #3946: auth: a token minted in the same millisecond as a privilege-change invalidation is dead on arrival (flaky test_3820 CI token exchange)
## Summary

A token minted in the same millisecond as an in-process credential/privilege invalidation for the same user is dead on arrival: `is_token_invalidated` rejects `issued_at_ms <= changed_at_ms`, and the mint sets `iat_ms = Utc::now().timestamp_millis()` without looking at the map.

The federated login path hits exactly this ordering on its own: `sync_federated_user` -> `apply_role_mapping` computes `privilege_changed` (true for every first login, because the role set goes from `{}` to the mapped roles) -> `invalidate_user_tokens(user_id)` stamps `now_ms`; the caller then mints the session token a few statements later. On a local Postgres those statements take well under a millisecond each, so the stamp and the mint occasionally share a millisecond and the freshly minted access token fails `validate_access_token` with `Token invalidated by credential change`.

## Evidence

`api::handlers::ci_auth::tests::test_3820_ci_token_exchange_caps_the_mint_at_the_assertion_expiry` (added in #3902) mints and immediately validates twice. On PR #3921's coverage run (job 105056208580) the second, warm mint failed:

```
thread '...test_3820_ci_token_exchange_caps_the_mint_at_the_assertion_expiry' panicked at backend/src/api/handlers/ci_auth.rs:601:14:
minted token validates: Authentication("Token invalidated by credential change")
```

The same head passed the test in the unit job, and #3941's coverage run passed it in 1.16 s, so it is timing, not the PR under test. Every open PR's unit and coverage job now has a small chance of failing on it.

## Fix

Mint after the watermark: in `generate_token_pair_capped`, if the invalidation map holds a watermark for `user.id` that is `>= now_ms`, issue the token at `watermark + 1` (and derive the seconds `iat` from that). A token minted after an invalidation must postdate it by construction; the `<=` semantics for genuinely older same-millisecond tokens stay as they are (`test_token_issued_at_exact_invalidation_time_is_rejected`). Add a DB-free regression test that stamps the map and mints in the same millisecond.

brandonrc:
## Summary

A token minted in the same millisecond as an in-process credential/privilege invalidation for the same user was dead on arrival. `is_token_invalidated` rejects `issued_at_ms <= changed_at_ms`, and the mint set `iat_ms = now` without looking at the map.

The federated login path produces exactly that ordering on its own: `sync_federated_user` → `apply_role_mapping` reports `privilege_changed` on every first login (the role set goes from `{}` to the mapped roles) → `invalidate_user_tokens(user_id)` stamps `now_ms`; the caller then mints the session a few sub-millisecond Postgres statements later. When stamp and mint share a millisecond, the freshly issued token fails `validate_access_token` with `Token invalidated by credential change`. Rare in production; visible in CI as the intermittent failure of `test_3820_ci_token_exchange_caps_the_mint_at_the_assertion_expiry` (PR #3921 coverage job 105056208580, same head green in the unit job).

## Change

- `invalidation_watermark_ms(user_id)`: read-only companion to `is_token_invalidated` over the same map.
- `generate_token_pair_capped`: if the user's watermark is at or after `now_ms`, mint at `watermark + 1` and derive the whole-second `iat` from that, so `iat` and `iat_ms` stay consistent. `exp` math is untouched. The `<=` rule in `is_token_invalidated` is untouched, so a genuinely older same-millisecond token is still rejected (`test_token_issued_at_exact_invalidation_time_is_rejected`).
- Other mint paths reviewed and left alone: `generate_scan_token` mints for the scanner service account with no user-keyed invalidation in its request paths; the TOTP pending/enrollment tokens are validated without the watermark check; `generate_token_pair_typed` delegates to the fixed function.

## Verification (through the rig)

- `cargo fmt --all -- --check`, `cargo clippy -p artifact-keeper-backend --lib --tests -- -D warnings`: clean.
- `cargo nextest run -p artifact-keeper-backend --lib -E 'test(invalidat) or test(3946) or test(iat) or test(generate_scan_token) or test(token_expiry)'`: 136 passed.
- `-E 'test(ci_auth)'`: 18 passed, including the reproducer.
- Mutation check: with the nudge replaced by a no-op, the new DB-free regression test `test_3946_token_minted_in_the_invalidation_millisecond_postdates_the_watermark` fails; restored, it passes.
- CHANGELOG guards pass.

Closes #3946


brandonrc:
Second commit 6ed4ef04 folds the three byte-identical replica-safe invalidation blocks (validate_access_token_async, refresh_tokens, mint_access_from_registry_refresh) into one private helper. Needed because the duplication gate measures the whole changed file and auth_service.rs was already at 3.5% before this PR; with the refactor it measures 2.7% (CI's exact pipeline reproduced locally). Behaviour and messages unchanged; 714 validator/invalidation tests pass.

PR #3958, opened 2026-09-17
feat(vscode): read the uploaded VSIX instead of trusting three publish headers

### LINKED ISSUE #3961: VS Code publish never reads the uploaded VSIX
### Problem Statement

`publish_extension` takes the publisher, name and version from the `x-publisher` / `x-extension-name` / `x-extension-version` headers and stores four fields (`build_vscode_metadata`). The uploaded `.vsix` is hashed and stored but never opened, and there is no VSIX reader in the crate — `backend/src/formats/vscode_extensions.rs` only has `parse_path`.

Two consequences:

- The catalog row registered by #3659 has no description. The source comment says why: "the coordinates arrive as headers, not a parsed manifest".
- A gallery response cannot be built from a published extension. A client needs `engines.vscode` to decide whether it can run the extension, and `targetPlatform` to choose between platform-specific builds.

### Proposed Solution

Parse the two manifests a `.vsix` carries and store what they say in the existing `artifact_metadata.metadata` JSONB (no migration):

- `extension/package.json` — `displayName`, `description`, `engines.vscode`, `icon`, `categories`, `extensionDependencies`, `extensionPack`
- `extension.vsixmanifest` — `Identity/@TargetPlatform` and the pre-release property, neither of which exists in `package.json`

Prefer the manifest for identity so `vsce package` output publishes without hand-set headers, and keep the headers accepted for existing callers. Malformed input (corrupt zip, missing manifest, missing `engines.vscode`) should be a 400, not a 500. `bounded_archive::read_metadata_from_zip` under `with_ingest_extraction` is the existing pattern for this.

### Alternatives Considered

Keep taking identity from headers and add more headers for the gallery fields. Rejected: the archive already carries them, and a header set that disagrees with the archive is a publish nobody can debug.

### Component

### Pre-submission Checklist

vbartusevicius:
Closes #3961

## Summary

`POST /vscode/{repo}/api/extensions` took the publisher, name and version from the `x-*` headers and never opened the uploaded `.vsix`, so a publish recorded four fields and nothing an editor can use.

Both manifests in the archive are now parsed and stored in the existing `artifact_metadata.metadata` JSONB (no migration):

- `extension/package.json` — `displayName`, `description`, `engines.vscode`, `icon`, `categories`, `extensionDependencies`, `extensionPack`
- `extension.vsixmanifest` — `Identity/@TargetPlatform` and the pre-release property, which `package.json` does not carry

The parser is in `backend/src/formats/vscode_extensions.rs` and extraction runs under `bounded_archive::with_ingest_extraction`, on the shared ingest budget. A corrupt archive, a missing manifest or a missing `engines.vscode` is a 400. The description now reaches `register_published_package`, which fixes the empty description on the Packages page (#3659).

Existing callers keep working: `vsce package` output publishes with no headers, and headers that are sent are checked against the archive (identity casefolded, version exact) with a disagreement refused as a 400 naming the header. A body the parser cannot read still publishes on the header coordinates and produces the same four-key record as before. Those coordinates are now validated before they become an artifact path and a storage key; previously `x-publisher: ../../etc` reached both unchecked.

## Test Checklist
- [x] Manually tested locally

Unit: manifest projection, the required fields, unusable identity, text bounds, wrong-typed and garbage JSON, platform/pre-release reads, unreadable package manifest, the archive read, corrupt and incomplete archives, the legacy identity path, the stored row shape, header conflicts. DB-backed: a headerless `vsce package` publish persisting its fields, a contradicting header refused, opaque bytes plus headers still succeeding, and neither refused.

Run against rust 1.98.0 and a real PostgreSQL: `cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings` clean, `cargo test -p artifact-keeper-backend --lib -- vsix vscode` 121 passed.

## API Changes
- [x] N/A - no API changes


brandonrc:
Taking this the last mile, as announced on #3960. Pushed on top of your commit: rebased onto main (CHANGELOG only, bullet under Unreleased citing #3961), and one fix from the audit. A body now falls back to the `x-*` headers only when it is not a zip at all or has no `extension/package.json`; a readable manifest that fails validation (for example no `engines.vscode`) is a 400 even when all three headers are sent, so the header-versus-archive conflict check can no longer be skipped. New DB test for that case (it got 201 before the fix). 123 vscode tests pass locally, clippy and fmt clean. Merging once CI is green, then #3959, then #3960.

Issue #3257, opened 2026-08-10
VS Code gateway: decide the private-repository gallery story (501 vs 403, and whether to support it at all)


brandonrc:
Follow-up from the review of #3253 (Open VSX gallery gateway). A scoping decision that should be made deliberately rather than inherited from a prototype.

`gallery_upstream` rejects a **private** Remote `vscode` repository on every gallery route with `501 Not Implemented` and the message "private gallery access is unsupported in this prototype" — including for an authenticated repository member holding legitimate read access. The reasoning in the source is sound and honest: gallery clients (VSCodium, code-server) have no way to configure a credential for a gallery URL, so shipping the routes for private repos would produce a broken experience rather than a secure one. It also has the useful side effect of sidestepping two known traps in this codebase — `can_access_repo` means token scope, not visibility, and `scopes: None` satisfies `require_scope` unconditionally for JWT sessions — because the decision never reaches them.

Two things to settle:

1. **Status code.** `501 Not Implemented` says "this server does not support the functionality required". If public-only is a permanent product decision, `403` describes it better and will not read as a bug to an operator. If it is temporary, `501` is right but the accompanying test currently asserts `"private Remote gallery route must remain unsupported"` — wording that turns a scoping decision into a requirement and will be the first thing to break when someone implements it.

2. **Whether to implement it.** VS Code's gallery client is anonymous by design, but a reverse proxy in front of AK can inject a credential, and `serviceUrl` can carry a path segment. If private galleries are wanted, the design needs to state how the credential arrives before any code is written.

Whichever way this goes, the docs should say it explicitly — `docs/vscode-openvsx-gateway.md` already does ("The repository must be **public** and readable anonymously"), so this is mostly about aligning the status code and the test's wording with the decision.

brandonrc:
## Decision + analysis

First, the same scoping note as #3254/#3255/#3256: **this code is not on `main`.** `origin/main` (`b0643226`) has no `gallery_upstream`; it exists only in draft **#3253** (`a0e1ee38`). So this is a change to make in that PR, not a parallel one. Recording the decision here so it is settled before the code moves.

I was asked to actually decide this rather than list options. Here it is.

---

## Decision

**1. Do not implement private-repository galleries.** Public-only is a deliberate product constraint, not a prototype limitation. Drop "in this prototype" from the message, the doc, and the test.

**2. Split the status code by which guard fired:**

| Guard | Status | Rationale |
|---|---|---|
[...]

PR #3959, opened 2026-09-17
feat(vscode): serve a hosted repository through the gallery protocol

### LINKED ISSUE #3956: VS Code gallery routes are Remote-only, so extensions published to a hosted repository cannot be installed
### Problem Statement

Every `/vscode/{repo}/gallery/...` route goes through `gallery_upstream` (`backend/src/api/handlers/vscode.rs`), which returns `501` for any repository type other than Remote. An extension published to a hosted `vscode` repository is therefore not reachable by any editor.

A VS Code-family editor has exactly one `extensionsGallery.serviceUrl`, so the gallery an operator configures has to serve locally published extensions as well as proxied ones.

Verified on 1.9.0 (`d771583`):

```bash
curl -i -X POST -H 'Content-Type: application/json' \
  -d '{"filters":[{"criteria":[{"filterType":8,"value":"Microsoft.VisualStudio.Code"}],"pageNumber":1,"pageSize":10}],"flags":950}' \
  https://<host>/vscode/vscode-internal/gallery/extensionquery

501 VS Code gallery routes require a Remote repository (vscode-internal is hosted)
```

### Proposed Solution

Split `gallery_upstream` into the gate every route shares (repository exists, public-only) and the Remote-specific upstream resolver, then answer `extensionquery`, `.../latest`, `manifest`, `item`, `vspackage` and `/asset` from the repository's own artifacts when it is hosted. Remote behaviour unchanged.

Public-only should stay as it is for hosted repositories too, for the reason recorded in #3257: a gallery client cannot present a credential.

Package bytes are already served by `/vscode/{repo}/extensions/{p}/{n}/{v}/download`, so the gallery file entry can point there instead of adding a new asset path.

### Alternatives Considered

- Run two galleries: not possible, `serviceUrl` is a single string (microsoft/vscode#179919).
- Point the editor at the existing `/vscode/{repo}/extensions/...` routes: no editor speaks them.

### Additional Context

The published docs page (https://artifactkeeper.com/docs/guides/more-formats/#vs-code-extensions) is also wrong in three ways, and the site does not appear to be in this repository:

- `serviceUrl` is documented without the `/gallery` suffix, so every request misses the route (and answers `401`, not `404`).
- The publish example uses `-F "file=@ext.vsix"` against `/vscode/main/extensions`; the route is `POST /vscode/{repo}/api/extensions` and it reads a raw body plus `x-*` headers.
- The Remote-only and public-only requirements are not mentioned.

Where does the docs site live?

### Component

### Pre-submission Checklist

vbartusevicius:
Closes #3956

## Summary

Every gallery route went through `gallery_upstream`, which returned `501` for any repository type except Remote, so an extension published to a hosted `vscode` repository could not be discovered or installed by an editor.

- `gallery_upstream` splits into `gallery_gate` (repository type, public-only — shared by every route) and the Remote-specific upstream resolver.
- `extensionquery`, `.../latest`, `manifest`, `item`, `vspackage` and `/asset` answer from the repository's own artifact rows when it is hosted.
- A version carries the `Microsoft.VisualStudio.Services.VSIXPackage` file entry that installs it, `lastUpdated`, `targetPlatform` for a platform-specific build, and the engine and pre-release properties.
- `filterType` 7/4 and 10 select, `pageNumber`/`pageSize` page, `Target` and `ExcludeWithFlags` are accepted as qualifiers, and `flags` decides what each version carries. `TotalCount` counts every match, and an unmatched query is an empty page rather than a 404.

The proxy path is not reached for a hosted repository and is unchanged for a remote one. Hosted package requests redirect to `/vscode/{repo}/extensions/{p}/{n}/{v}/download`, which already serves those bytes quarantine-checked and download-counted, so no new byte surface is added; any other asset type is a 404. Public-only still applies (#3257), and virtual repositories still answer `501`.

Independent of #3958 — this branch is cut from `main` and reads every metadata field defensively. With #3958 merged, a hosted extension additionally carries its engine range, display name, description, categories and target platform.

Two limitations, unchanged by this PR: a hosted repository holds one build per version (the artifact path has no platform component, so a second platform under one version is a 409 today), and icon and README assets are not served.

## Test Checklist
- [x] Manually tested locally

Unit: row grouping and version ordering, the package/engine/platform shape, flag handling, identity and search filters, paging and `TotalCount`, qualifier criteria not selecting. DB-backed: the #3956 reproduction returning the extension installably, manifest + latest + vspackage + asset + a 404 for an unknown extension, and a private hosted repository refused.

`gallery_routes_reject_non_remote_repositories` becomes `gallery_routes_reject_a_private_repository_of_every_served_type`, pinning that the public-only gate still refuses hosted and virtual.

Run against rust 1.98.0 and a real PostgreSQL: `cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings` clean, `cargo test -p artifact-keeper-backend --lib -- hosted gallery vscode` 179 passed.

## API Changes
- [x] N/A - no API changes

No new routes and no migration; the existing routes answer for one more repository type. The `501` message now reads "require a Remote or Hosted repository".


brandonrc:
Pushed on top of your commit: rebased onto main (bullet under Unreleased citing #3956), and the audit fix done properly rather than documented. The 2,000-row cap is gone: search and paging run in SQL over distinct extension ids, one query returns the total and the page's ids, then only that page's versions load, newest 100 per extension (documented in the CHANGELOG; older versions still install by exact coordinates). Listings also leave out versions the download gate refuses (unexpired quarantine hold or rejected), with an expired hold listed again, matching conda; the filter is in SQL so paging and `TotalCount` see it. Two new DB tests (2,001 versions plus a second extension; quarantine states), both failing on the previous code. 114 tests pass locally. Merging after #3958.

PR #3960, opened 2026-09-17
feat(vscode): aggregate a virtual repository's members into one gallery

### LINKED ISSUE #3962: A virtual vscode repository cannot serve a gallery
### Problem Statement

Gallery routes answer `501` for a virtual repository. Since an editor has exactly one `extensionsGallery.serviceUrl`, a virtual repository is the only place an organisation's own extensions and a proxied public gallery can be served together — which is the configuration most deployments actually need.

Depends on hosted gallery support (#3956).

### Proposed Solution

Answer `extensionquery` and `.../latest` by reading the members the caller may see, in the operator's priority order: hosted members from their artifact rows, remote members by forwarding the client's own query body upstream. Merge the results and let the first member that knows an extension describe it, so ordering a hosted member first is how an internal extension shadows a public one.

Rewrite each member's entries onto that member's own routes, so packages and assets resolve through the per-member route that already serves them and the virtual repository needs no byte-serving route of its own.

Members must go through `authorize_virtual_members` (#1804) so aggregation cannot widen a member's visibility; for the anonymous caller a gallery client is, that means public members only.

### Alternatives Considered

Deduplicate per (publisher, name, version, targetPlatform) and merge version lists across members. Rejected for now: priority order is the resolution rule virtual repositories already use elsewhere, and it gives the same shadowing behaviour with no new concept.

### Component

### Pre-submission Checklist

vbartusevicius:
Closes #3962

Stacked on #3959 — a fork branch cannot be a base here, so this targets `main` and its diff includes #3959's commit. Review #3959 first; once it merges this reduces to `0817614`.

## Summary

Gallery routes answered `501` for a virtual repository, which is the only place an organisation's own extensions and a proxied public gallery can be served from one `serviceUrl`.

- `extensionquery` and `.../latest` read every member the caller may see, in the operator's priority order: hosted members from their artifact rows, remote members by forwarding the client's own query body upstream.
- Results merge with the first member that knows an extension describing it, so ordering a hosted member first is how an internal extension shadows a public one.
- Each member's entries are rewritten onto that member's own routes, so packages and assets resolve through the per-member route that already serves them. `vspackage` and `/asset` still answer `501` for a virtual repository and nothing a virtual page emits points at them.

Members go through `proxy_helpers::authorize_virtual_members`, the same read predicate a direct read applies (#1804), so aggregation cannot widen a member's visibility — for the anonymous caller a gallery client is, that means public members only. Each remote member's own age gate and curation still apply to its entries. A member with no usable gallery upstream, or whose page exceeds the metadata ceiling, is skipped with a `warn!` instead of failing the aggregate, and at most 8 members are read per query.

Two approximations: `TotalCount` is the sum of what members reported (an upper bound when an extension exists in more than one), and the merged page is not re-sliced, because members are already paged by the client's own `pageNumber`/`pageSize` and re-slicing would drop entries the next page never returns.

## Test Checklist
- [x] Manually tested locally

Unit: first member wins on a casefolded shared identity, totals sum, unaddressable entries dropped. DB-backed: a virtual repository aggregating a hosted member with the package URL pointing at that member, a private member not reaching an anonymous caller, priority deciding a shared identity, and `vspackage`/`asset` still `501`.

Run against rust 1.98.0 and a real PostgreSQL: `cargo fmt --check` and `cargo clippy --lib --tests -- -D warnings` clean, `cargo test -p artifact-keeper-backend --lib -- vscode gallery vsix hosted virtual` 490 passed.

## API Changes
- [x] N/A - no API changes


brandonrc:
Audit notes on the virtual gallery, in merge order: #3958 first, then #3959, then this one (it carries #3959's commit, so it can only land after it, and its CHANGELOG entry for #3959 has to be dropped on rebase).

Blocking before this can merge:

1. An upstream error from any remote member fails the whole virtual response (around `vscode.rs:3472-3487`). An Open VSX outage then hides the organisation's own hosted extensions too. A failed member should be skipped and the response marked partial, not fail the request.
2. Remote members receive the client's raw query rather than the composed query the Remote path uses, so update checks exceed the 2 MiB limit, the member is skipped, and updates for proxied extensions silently disappear.
3. Hosted members are not paged, so every page repeats all hosted matches.

Should fix: up to eight members' pages accumulate with no admission limit on an anonymously reachable path; the gallery's counting semaphore should wrap the fan-out. No test exercises a remote member; at least one is needed for items 1 and 2.

Nothing nests a second metadata-budget reservation, which is the rule that bit conda and Composer this week, so that part is fine. I am taking #3958 and #3959 the last mile myself (small fixes, rebase, merge); this one I would rather leave with you for the three items above, unless you prefer I take it too.

brandonrc:
Change of plan: I am taking this one the last mile as well, in sequence after #3958 and #3959, with the four items above fixed on your branch and your commits kept. If you have already started on them, say so and I will hold.

brandonrc:
Pushed on top of your commit, rebuilt as #3959 plus this PR's own commit (the duplicate of #3959's commit and its CHANGELOG bullet dropped; your #3962 bullet corrected to match the new behaviour). The four audit items:

- V1: a remote member that errors is skipped with a warning naming it and the status; the other members are still served.
- V2: the Remote handler's pass-through-or-compose decision moved into a shared `plan_remote_gallery_query`, so remote members get the composed query, not the client's raw one; the direct Remote path behaves as before.
- V3: hosted members are paged in SQL with the client's page number and size.
- V4: each aggregation holds one permit of the existing composition admission semaphore until the page is serialized. No byte reservation is nested or held across the member loop.

One DB test per item, with a mock remote member for V1 and V2; each fails with its fix reverted. 124 tests pass locally. This merges after #3959. Left as a follow-up: an authenticated caller sees entries from private members that the public-only member gallery routes cannot install (no access widening, just uninstallable entries).

brandonrc:
Rebased this onto main now that #3959 has merged. Your two virtual-gallery commits are unchanged. The two #3959 commits dropped out because main already has them as the squash.

Main now builds the CHANGELOG from fragments at release time (#4205). So I moved your #3962 entry, word for word, into `changes/unreleased/3962-vscode-virtual-gallery.md` in one small commit on top. Clippy with `-D warnings` is clean on the rebased tree. It's in the merge queue and merges once CI is green.

brandonrc:
I pushed one small commit to keep this mergeable. Main now requires every inline `#[cfg(test)]` module to carry a CI test-shard gate line (#4234), which splits the unit tests across five hosted runners. This PR's new test modules didn't have one yet, so Check Rust failed. I added them with `python3 scripts/ci/test-shards.py apply`. There are no code changes.

PR #3978, opened 2026-09-17
fix(nuget): register one catalog row per pushed package, not two

### LINKED ISSUE #3976: NuGet push creates two catalog rows for one package (published casing + lowercased)
### Description

`dotnet nuget push` of a package whose id has mixed case creates **two** rows in `packages`, so the Packages page lists the package twice — once as published, once lowercased.

Reproduced with a unit test against `main`: pushing `FiscalTapeParser.Xml 1.2.3` leaves

```
["FiscalTapeParser.Xml", "fiscaltapeparser.xml"]
```

Cause: the push path registers the catalog row twice, under two different names.

1. `push_package` (`backend/src/api/handlers/nuget.rs`) lowercases the id (`let package_id = nuspec.id.to_lowercase()`) and passes it to `ArtifactService::upload_stream_with_sync_options` as the artifact name, so `artifacts.name` is `fiscaltapeparser.xml`.
2. `ArtifactService::finalize_upload` registers a catalog row derived from `artifacts.name` → `fiscaltapeparser.xml`.
3. `push_package` then registers its own catalog row from `nuspec.id` → `FiscalTapeParser.Xml`.
4. `packages` is `UNIQUE (repository_id, name)`, which is case-sensitive, so both inserts succeed.

The existing push tests query `WHERE name = 'IndexedPkg'`, so neither notices the second row.

PyPI is not affected (it passes its normalized name to both). The #3659 catalog backfill is not affected (`BACKFILL_FORMATS` excludes nuget).

### Environment Information

- Deployment type: Docker
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL 18
- Reverse Proxy: n/a
- Network setup: n/a

### Backend Version

1.9.0 and `main`

### Web UI Version

N/A

### Package Format

NuGet (.NET)

### Expected Behavior

One `packages` row per package id, under the casing the `.nuspec` declares.

### Observed Behavior

Two rows: the `.nuspec` casing and an all-lowercase duplicate. Both show on the Packages page; the lowercase one carries no description or metadata.

### Steps to Reproduce

1. Create a hosted `nuget` repository.
2. `dotnet nuget push FiscalTapeParser.Xml.1.2.3.nupkg -s http://<host>/nuget/<repo>/api/v2/package -k <token>`
3. Open the repository's Packages tab, or `SELECT name FROM packages WHERE repository_id = '<id>'`.

### Additional Context

Existing duplicates stay until cleaned up; the fix only stops new ones. Merge per repository with:

```sql
WITH ranked AS (
    SELECT p.id,
           first_value(p.id) OVER (
               PARTITION BY p.repository_id, LOWER(p.name)
               ORDER BY (p.name <> LOWER(p.name)) DESC, p.created_at
           ) AS keep_id
    FROM packages p
    JOIN repositories r ON r.id = p.repository_id
    WHERE r.format = 'nuget'
)
INSERT INTO package_versions (package_id, version, size_bytes, download_count, checksum_sha256)
SELECT r.keep_id, pv.version, pv.size_bytes, pv.download_count, pv.checksum_sha256
FROM ranked r
JOIN package_versions pv ON pv.package_id = r.id
WHERE r.id <> r.keep_id
ON CONFLICT (package_id, version) DO NOTHING;

WITH ranked AS (
    SELECT p.id,
           first_value(p.id) OVER (
               PARTITION BY p.repository_id, LOWER(p.name)
               ORDER BY (p.name <> LOWER(p.name)) DESC, p.created_at
           ) AS keep_id
    FROM packages p
    JOIN repositories r ON r.id = p.repository_id
    WHERE r.format = 'nuget'
)
DELETE FROM packages WHERE id IN (SELECT id FROM ranked WHERE id <> keep_id);
```

Run it inside a transaction and check the row counts first. `package_versions` cascades on delete, so the second statement drops the duplicate's remaining version rows.

### Pre-submission Checklist

vbartusevicius:
Closes #3976

## Summary

Pushing a package whose id has mixed case created two `packages` rows, so the Packages page listed it twice — reported by engineers as `FiscalTapeParser.Xml` alongside `fiscaltapeparser.xml`.

The push path registered the catalog row twice under two names:

1. `push_package` lowercases the `.nuspec` id and passes it as the artifact name, so `artifacts.name` is `fiscaltapeparser.xml`. That is deliberate — every NuGet read compares `LOWER(name)`.
2. `ArtifactService::finalize_upload` registered a catalog row derived from `artifacts.name`.
3. `push_package` then registered its own row from `nuspec.id`.
4. `packages` is `UNIQUE (repository_id, name)`, which is case-sensitive, so both inserts succeeded.

`finalize_upload` now takes an optional catalog name from the caller. NuGet supplies `nuspec.id`; every other caller passes `None` and keeps the existing derivation, so the Maven/Gradle coordinate logic that replication, migration and generic pushes rely on is unchanged. PyPI was never affected (it passes its normalized name to both) and the #3659 backfill never walked nuget.

The lowercased row is the one that carried no description or metadata, so the surviving row is also the more complete one.

**Existing duplicates are not rewritten.** #3976 carries idempotent SQL to fold each twin's `package_versions` into the row that kept the published casing and delete the twin, scoped to `nuget` repositories where an id differing only in case is by definition the same package. It is deliberately not a migration: `packages` is a hot table, so a whole-table merge would need the batched shape `migration_safety` requires, and the merge is a one-off repair rather than a schema change.

`push_package_registers_one_catalog_row_per_package_id` pushes `FiscalTapeParser.Xml` and asserts the repository's whole `packages` name list. On `main` it fails with exactly the reported symptom:

```
  left: ["FiscalTapeParser.Xml", "fiscaltapeparser.xml"]
 right: ["FiscalTapeParser.Xml"]
```

The two existing push tests did not catch it because both query `WHERE name = '<PascalCase>'`, which never sees the twin.

## Test Checklist
- [x] Manually tested locally

Run against the pinned 1.98.0 toolchain and a fresh PostgreSQL:

```
cargo fmt --all -- --check                                              # clean
cargo clippy -p artifact-keeper-backend --lib --tests -- -D warnings    # clean
cargo nextest run --workspace --lib --bins --test-threads 8             # 16671/16672
```

The single failure is `test_probe_storage_health_filesystem_unwritable`, which fails identically on unmodified `main` in a container running as root (a `chmod 000` directory is still writable), so it is environmental. `migration_worker::test_nuget_import_populates_package_catalog` (the import path that still uses the generic derivation) and `streaming_invariant` both pass.

## API Changes
- [x] N/A - no API changes

No route or response change and no migration. `ArtifactService::upload_stream_with_sync_options` / `finalize_upload` gain one `Option<&str>` parameter; the three production call sites and one test call site are updated.


brandonrc:

The only conflict was `CHANGELOG.md` — your entry and the ones that landed on `main` both target the top of `### Fixed`. Both were kept verbatim, yours first; no code was touched.

This PR does share two files with changes that landed on `main` since your merge-base (`api/handlers/repositories.rs`, `services/artifact_service.rs`), so the rebase was verified locally rather than assumed: `cargo clippy --workspace --all-targets -- -D warnings` is clean on the rebased commit.

It also clears the `🔒 Security Audit` failure, which was never about this PR: RUSTSEC-2026-0286 (cryptoki 0.12.0) came from the stale lockfile on the old base, and `main` fixed it in 07feface.

Tracking the underlying churn in #4016.

brandonrc:
Thanks — the diagnosis here was exactly right, including the two things that are easy to get wrong: the #3659 backfill really does skip `nuget`, and PyPI really does pass its normalized name to both writers. Taken over to get it merged for 1.10.0; you rebased while the review ran, so the takeover commit now sits on top of your rebased head and your code and test are unchanged.

One addition on top. Pinning both writes to `nuspec.id` fixes one push but not two: `packages` is `UNIQUE (repository_id, name)` over raw text and a NuGet id is case-insensitive, so `CaseStable.Pkg 1.0.0` followed by `casestable.pkg 2.0.0` would still have opened a twin. The name is now resolved once before the upload from the casing an earlier push already registered (`LOWER(name)` lookup, best-effort), falling back to the `.nuspec` casing for a new id, so the first push still decides the display casing you specified and later pushes converge on it. Your test is kept as-is with `push_package_reuses_the_catalog_casing_an_earlier_push_registered` beside it; 17 nuget catalog/push tests pass against a live Postgres.

Note for operators stays as in your upgrade note: twins created before the fix are folded by the SQL on #3976; no migration ships. Merging once CI is green.

brandonrc:
Correction: the first push of the takeover commit (f728ac43) carried a leftover conflict block in CHANGELOG.md for a few minutes; replaced by f3b27407, which passes the conflict-marker and placement guards. Nothing else changed.

Issue #1643, opened 2026-06-04
Security: promotion gate fails open on unscanned artifacts + block_unscanned toggle is inert


brandonrc:
**Found during the security review of PR #1640.** Pre-existing fail-open gap in the promotion gate — NOT introduced by #1640, but adjacent to it and more significant than the display bug #1640 fixed.

## Two related fail-open problems in `promotion_policy_service.rs`
1. **`block_unscanned` is silently inert.** `scan_policies.block_unscanned BOOLEAN` exists (migration 022), is read/written by the security policy CRUD API (`models/security.rs`, `api/handlers/security.rs`, `policy_service.rs`), but **`promotion_policy_service.rs` references it zero times.** An operator who sets `block_unscanned = true` expecting unscanned artifacts to be blocked from promotion gets **no enforcement**.
2. **Unscanned artifacts fail open.** `evaluate_artifact` only runs CVE policy `if let Some(ref summary) = cve_summary`, and `get_cve_summary` returns `None` when there is no completed scan → no violations → `passed = true, action = Allow`. An artifact with no completed scan (or a scan still in progress) is promoted unchecked.

## Why it matters
The promotion gate's security promise is "don't promote vulnerable/unvetted artifacts." Both gaps let an artifact through *without being vetted at all* — arguably worse than a high-CVE artifact that at least got scanned.

## Fix
- Wire `block_unscanned` into `evaluate_artifact`: when the policy has `block_unscanned = true` and there is no completed scan, return `Block` (with a clear "artifact not scanned" violation).
- Decide the default for unscanned when `block_unscanned = false` (current fail-open) — keep, or make it a deliberate, documented choice.
- Tests: `block_unscanned = true` + no scan → promotion blocked; scan-in-progress → blocked or allowed per policy.

## Note
Related nit: the open-CVE / count basis takes a single global-latest completed scan (`LIMIT 1`), ignoring other `scan_type`s' latest scans — consistent with existing count behavior but worth revisiting (a malware/license scan's findings can be masked by a more recent dependency scan).

PR #3979, opened 2026-09-17
fix(promotion): apply the quality gate and CVE/licence policy to bulk promotion

### LINKED ISSUE #3977: Bulk promotion skips the quality gate and CVE/licence policy that single-artifact promotion enforces
## Summary

`POST /api/v1/promotion/repositories/{key}/promote` (bulk promotion) only enforces the per-pair `promotion_rules` check on each artifact. It never runs the quality-gate evaluation and never runs the CVE/licence promotion policy engine, both of which the single-artifact route (`POST .../artifacts/{artifact_id}/promote`) has always run.

An artifact that the single-artifact route would refuse — `409` on a blocking quality gate, or `promoted: false` on a CVE/licence policy block, including the `scan_policies.block_unscanned` gate that stops artifacts with no completed security scan (#1643) — is promoted into the release repository by the bulk route regardless. A bulk request array of a single element is enough to take this path, so it is not only a batch-specific gap.

## Impact

Any caller with promote access to a staging repository can bypass:
- the quality gate (health score / grade / violation thresholds), and
- the CVE/licence policy engine, including `scan_policies.block_unscanned`

by calling the bulk promotion endpoint instead of the single-artifact one, even with a one-item array. This defeats the purpose of both gates for any release pipeline that promotes artifacts individually through the bulk API.

## Expected behavior

The bulk route should evaluate the quality gate and the CVE/licence policy per item, in the same order the single route uses (gate → policy → promotion_rules → approval consume), honouring the same `skip_policy_check` admin override. A blocked item should fail that item and let the batch continue (bulk promotion already reports per-artifact outcomes), rather than either silently succeeding or refusing the whole batch.

## Where

`backend/src/api/handlers/promotion.rs` — `promote_artifacts_bulk` vs `promote_artifact`.

## Related (found while building regression coverage for this)

- A quality gate configured to `warn` doesn't surface its violations on a successful promotion: `build_success_response` hardcodes `policy_violations: vec![]`, so warn-level findings collected by the handler are dropped before they reach the caller.
- The promotion endpoints' OpenAPI annotations document `422` for validation refusals; the handlers have always returned `400` (`AppError::Validation`), so the spec has been wrong for every consumer generating an SDK from it.

mymarche:
## Summary

`promote_artifacts_bulk` only enforced the per-pair `promotion_rules` check on each item. It never ran the quality-gate evaluation or the CVE/licence promotion policy engine — both of which `promote_artifact` (single-artifact route) has always run. An artifact the single route blocks with `409` on a blocking quality gate, or answers `promoted: false` on a CVE/licence policy block (including `scan_policies.block_unscanned`, #1643), was promoted into the release repository by the bulk route regardless — and a one-element array was enough to take that path.

Both evaluations now run per item, in the same order the single route uses (gate → policy → promotion_rules → approval consume), honouring the same `skip_policy_check` admin override. A blocked item fails **that item** and the batch continues — the single route still refuses the whole request — because bulk promotion already reports per-artifact outcomes, so a partial result stays distinguishable from a wholesale refusal.

Two smaller fixes surfaced while writing the end-to-end regression test for this:
- `build_success_response` hardcoded `policy_violations: vec![]`, so a `warn`-configured quality gate's violations were collected but silently dropped from a successful promotion's response.
- The promotion handlers' `#[utoipa::path]` annotations documented `422` for validation refusals; `AppError::Validation` has always mapped to `400`, so the OpenAPI spec (the input to the web/CLI SDKs) was wrong for five operations.

Fixes #3977

Added `backend/tests/staging_promotion_e2e_tests.rs`: end-to-end coverage driving the real `promotion::router()` over a live pool, including `test_bulk_applies_the_quality_gate_per_item` and `test_bulk_success_reports_warn_level_violations`, which fail on `main` and pass on this branch. Wired into `.github/workflows/ci.yml`'s Tier 2 integration allowlist and verified against `ci_test_surface`.

## Test Checklist
- [x] Unit tests added/updated (`promotion.rs` unit tests updated for the new `build_success_response` signature; new `test_build_success_response_carries_warn_level_violations`)
- [x] Integration tests added/updated (`staging_promotion_e2e_tests.rs`, 30 cases, run against Postgres with `AK_TESTS_REQUIRE_DB=1`)
- [ ] E2E tests added/updated (if applicable) — N/A, covered by the new backend integration suite
- [x] Manually tested locally (`cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, full `cargo nextest run --workspace --lib --bins`, the new integration suite, and the `promotion`/`ci_test_surface`/`test_openapi_spec_is_valid` targets — all pass)

## API Changes
- [x] New endpoints have `#[utoipa::path]` annotations — N/A, no new endpoints; five existing annotations corrected from `422` to `400` to match actual behaviour
- [x] Request/response types have `#[derive(ToSchema)]` — unchanged
- [ ] Migration is reversible (if applicable) — N/A, no migration
- [ ] N/A - no API changes


brandonrc:
Taking this the last mile now that the three PRs it overlapped with have merged (#4163, #4190, #4165). Pushed on top of your two commits, which keep their content:

- rebased onto main; the two artifact INSERTs you moved into `insert_promoted_artifact_row` now take the source artifact's `origin` (from #4190), passed in from each route's existing lookup; the now-unused bulk-path sqlx cache file is removed. Both routes call the same `evaluate_artifact`, so the bulk route now also enforces the policy predicates from #4165;
- the two #3977 Fixed bullets had merged silently into the released 1.10.0 section; moved under Unreleased;
- bulk success now writes the real policy evaluation to `promotion_history` (shared `build_policy_result_json` with the single route) instead of a hardcoded pass, with a DB test that the bulk and single records match;
- the per-item "Policy evaluation" and "Rule evaluation" messages no longer carry raw database text (sanitized, logged server-side);
- a comment on why the bulk route runs the repository-shape check before the per-item gates.

Locally 964 promotion, policy, origin and approval tests plus your 30-test integration suite (31 with the new one) pass against a migrated database; clippy and fmt clean. Merging once CI is green on this head; the earlier unit-test run was OOM-killed, so this is its first full run. Two pre-existing issues the audit found are filed separately: `skip_policy_check` is honoured for non-admin tokens, and a gate evaluation error fails open.

PR #3981, opened 2026-09-17
fix(nuget): federate a virtual repository's members on read and download

### LINKED ISSUE #3980: A virtual NuGet repository does not federate: hosted members are unreachable and remote versions are hidden
### Description

A virtual `nuget` repository over a hosted (private) member and a remote member pointed at nuget.org does not federate: the hosted member's packages are unreachable and the remote member's versions are hidden whenever the hosted member publishes the same package id.

Two defects, both reproduced as unit tests against `main`.

**1. A download ignores member priority.** `flatcontainer_download` asks every *remote* member (through V3 service-index discovery) before it resolves hosted members, so for a coordinate both hold, upstream wins:

```
assertion `left == right` failed: the higher-priority hosted member must serve the coordinate it holds
  left:  "upstream nupkg"
 right:  "hosted member nupkg"
```

If upstream does not have the coordinate at all, the hosted member still serves it — which is why this looks intermittent rather than broken.

**2. Metadata is local-OR-remote, never merged.** `flatcontainer_versions` and `registration_index` query the hosted members first and consult remote members only when the local result is *empty*. One version published privately therefore hides every upstream version of that id:

```
  left:  ["1.0.0"]
 right:  ["1.0.0", "2.0.0", "3.0.0"]
```

`dotnet restore` resolves a version through these two documents, so restoring `2.0.0` of a package the private repository also publishes fails with NU1102 while the same restore against the remote repository directly succeeds. Whether a given package works depends on whether the private repository happens to publish that id — the "sometimes works, sometimes not" in the report.

V3 search already federates (`merge_upstream_search_data`), which is why search shows packages that then fail to restore.

### Environment Information

- Deployment type: Docker
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL 18
- Reverse Proxy: n/a
- Network setup: n/a

### Backend Version

1.9.0 and `main`

### Web UI Version

N/A

### Package Format

NuGet (.NET)

### Expected Behavior

A virtual repository resolves a package id across all its members: the version list and registration index carry every member's versions, and a download is served by the highest-priority member that holds the coordinate.

### Observed Behavior

- A coordinate a hosted member holds is served from a remote member instead.
- A package id published by any hosted member reports only that member's versions; upstream versions are invisible, so restore fails for them.

### Steps to Reproduce

1. Create a hosted `nuget` repository, a remote one pointed at `https://api.nuget.org/v3/index.json`, and a virtual one with both as members (hosted at priority 1).
2. Push `SomePkg 1.0.0` to the hosted repository, where `SomePkg` also exists on nuget.org with other versions.
3. `GET /nuget/{virtual}/v3/flatcontainer/somepkg/index.json` — only `1.0.0` is listed.
4. `dotnet restore` a project referencing a nuget.org-only version of `SomePkg` against the virtual feed — NU1102.

### Additional Context

Worth ruling out separately if the symptom persists after this: a virtual repository only resolves members the **caller** may read directly (#1804, #3178), so an anonymous or repo-scoped-token client sees no private member at all and the virtual behaves as if it contained only the remote. Granting the caller read access on the member repository — not just on the virtual — is what makes a private member visible.

The legacy V2/Chocolatey feed (`v2_odata`, `load_hosted_v2_entries`) still lists hosted members only; that is a separate gap and not covered here.

### Pre-submission Checklist

vbartusevicius:
Closes #3980

## Summary

Over a hosted (private) member and a remote member pointed at nuget.org, a virtual `nuget` repository did not federate. Two independent defects, both with a failing test on `main`:

**Downloads ignored member priority.** `flatcontainer_download` asked every remote member — through V3 service-index discovery — *before* resolving hosted members, so a coordinate both hold was served from upstream:

```
  left:  "upstream nupkg"
 right:  "hosted member nupkg"
```

Hosted members now resolve first through `resolve_virtual_download_from_members` (the shared priority-preserving resolver), and remote members are asked only on a miss. A hosted member answering 403 (quarantine / scan policy) or 409 (publish-age hold) stays terminal, as that resolver already treated it — falling through would hand the client the bytes the member just blocked.

**Metadata was local-OR-remote, never merged.** `flatcontainer_versions` and `registration_index` consulted remote members only when the local result was *empty*, so one privately published version hid every upstream version of that id:

```
  left:  ["1.0.0"]
 right:  ["1.0.0", "2.0.0", "3.0.0"]
```

`dotnet restore` resolves a version through those two documents, so restoring any other version failed with NU1102 — and whether a package worked depended on whether the private repository happened to publish its id, which is the "sometimes works, sometimes not" in the report. The version list now merges every remote member's versions (deduped case-insensitively, sorted by `version_compare`), and the registration index merges upstream pages with the local one: pages that inline their leaves are folded in and deduped by version with the local member winning, pages that only reference leaves by URL are passed through exactly as the single-member proxy did.

V3 search already federated (`merge_upstream_search_data`), which is why search could list a package that then failed to restore.

### Not changed

- **Member visibility.** Only members the caller may read directly take part (#1804, #3178), and an empty accessible set still answers with the shared "no accessible members" body rather than a bare 404. A private member stays invisible to a caller with no read grant on it — worth checking first if a report persists after this.
- **Remote and hosted repositories.** Both paths are untouched; the Remote arm of each handler is byte-identical.
- `proxy_v3_registration` keeps its signature and behaviour; the document fetch is split out as `fetch_v3_registration` so the virtual path can merge instead of forward.
- The legacy V2/Chocolatey feed still lists hosted members only. Called out in #3980 rather than widened here.

`virtual_federation_tests` builds the reported topology (hosted member at priority 1, remote member on a wiremock V3 feed) and pins both behaviours. Both cases fail on `main` with the output quoted above.

## Test Checklist
- [x] Manually tested locally

Run against the pinned 1.98.0 toolchain and a fresh PostgreSQL:

```
cargo fmt --all -- --check                                              # clean
cargo clippy -p artifact-keeper-backend --lib --tests -- -D warnings    # clean
cargo nextest run --workspace --lib --bins --test-threads 8             # 16672/16673
```

The single failure is `test_probe_storage_health_filesystem_unwritable`, which fails identically on unmodified `main` in a container running as root (a `chmod 000` directory is still writable). The existing virtual-member authorization tests (#3323/#3324/#3329) and `read_db_tests` pass unchanged.

## API Changes
- [x] N/A - no API changes

No new routes and no migration. One response shape changes for a virtual repository: its registration index can now carry more than one page — `count` is the page count, which is what the protocol specifies and what an upstream document already does.


brandonrc:

The only conflict was `CHANGELOG.md` — your entry and the ones that landed on `main` both target the top of `### Fixed`. Both were kept verbatim, yours first; no code was touched. Confirmed mechanically: the files this PR changes and the files `main` changed since the merge-base have **zero overlap**, so the rebased commit is code-identical to the one CI already tested.

This also clears the `🔒 Security Audit` failure, which was never about this PR: it was RUSTSEC-2026-0286 (cryptoki 0.12.0) inherited from the stale lockfile on the old base. `main` fixed that in 07feface.

Tracking the underlying churn in #4016 — three of the four conflicted PRs today were blocked on nothing but the CHANGELOG insertion point.

brandonrc:
Thanks @vbartusevicius — the diagnosis is right and both regression tests are real: grafted onto unmodified `main` they fail with exactly the output you quoted. You rebased while I was reviewing, so the takeover commit now sits on top of your rebased head (a3de922c); your commit is preserved unchanged.

One behavioural change on top of yours: resolving all hosted members ahead of every remote member fixes the reported topology but mirrors the inversion — a remote member configured at priority 1 would then lose to a hosted member at priority 2, while #3980's expected behaviour is "the highest-priority member that holds the coordinate". The walk is now a single pass in declared priority order, in runs of one kind (a hosted run through the shared resolver, a remote member through V3 discovery). Nothing is lost by the grouping: the resolver's two-phase fan-out only overlaps upstream work, which a hosted run has none of.

Also sorted the merged registration leaves by version (upstream leaves were appended in arrival order, so a page's `lower`/`upper` described neither end) and added tests for the registration half, which had none: a hosted+remote merge test, a fall-through download, a priority test with the remote member at priority 1 (fails against the PR as submitted), and unit tests over `merge_registration_leaves`. 175 nuget tests pass against a live Postgres. The V2/Chocolatey feed you both deferred is now #4021, and the CHANGELOG sentence points there. Merging once CI is green.

Issue #3954, opened 2026-09-17
Observability: inbound trace context is discarded, upstream and DB calls have no spans (41% of wall clock untraced), and the latency metric exports as a summary


knowinglyAnonymous:
### Description

Diagnosing a latency problem on 1.9.x meant building a bespoke probe and hand-parsing individual
traces, because the shipped telemetry cannot answer "where did this request spend its time". Each
gap below is small on its own; together they make the service hard to diagnose from its own signals.

The instrumentation that exists is good — storage spans carry `code.file.path`,
`code.line.number`, `storage.operation` and the exact object key, which is better than most
services. The gaps are specific, and most look cheap. All references are `main` @ `42ad12bb`, and we
are happy to send PRs.

**Priority, if only two can happen: §1 and §2.** Together they turn a trace from "some object-store
calls" into an actual end-to-end latency breakdown.

---

## 1. The HTTP entry span is `INTERNAL`, and the inbound trace context is parsed for logging but discarded for tracing

This is the one we would most like fixed.

**Setup:** a reverse proxy terminates TLS in front of the backend, is itself instrumented, and emits
its own spans.

**Observed, against a real collector:**

* The backend **does** read `traceparent` — `backend/src/api/middleware/tracing.rs:138` extracts the
  trace-id from it. But it is used only as a **correlation-ID string**; it is never turned into an
  OpenTelemetry remote parent. (Worth stating precisely, because "the header is ignored" would be
  wrong and easy to dismiss.)
* Confirmed empirically: across 200 consecutive edge log lines the backend's `X-Correlation-Id`
  equalled the edge's trace ID in **713 of 713** cases — so the header is definitely arriving and
  being read.
* Yet in the trace store, **every backend span is a root**: `service=artifact-keeper`,
  `span=http_request`, `parent=(ROOT)`, with a fresh trace ID per request. A TraceQL join across the
  edge and backend service names returns **0 traces**.
* The backend nests its *own* spans correctly (`health_check` parents under `http_request`), so it
  is specifically the **remote** parent that is dropped, not span nesting in general.
* The root span is also `SPAN_KIND_INTERNAL`, not `SPAN_KIND_SERVER` (`backend/src/main.rs:1151`
  sets no `otel.kind`). An INTERNAL root is treated by trace UIs and span metrics as a standalone
  operation rather than a request entry point. The codebase already knows the idiom — the storage
  spans set `otel.kind = "client"`.

[...]

Issue #2414, opened 2026-07-11
Audit events generate unrelated correlation IDs instead of preserving the request correlation ID


cazlo:
### Description

Artifact Keeper establishes a correlation ID for every HTTP request, but audit events caused by that request do not preserve it.

The request middleware obtains the correlation ID from:

1. `X-Correlation-ID`
2. The trace ID in a W3C `traceparent` header
3. A generated UUID when neither header is present

That value is attached to the request, included in the tracing span, and returned in the `X-Correlation-ID` response header.

However, `AuditEntry::new()` independently generates a new random UUID for every audit event:

```rust
correlation_id: Uuid::new_v4(),
```

No production audit emitter passes the request's `CorrelationId` into the audit entry. The value stored in `audit_log.correlation_id` is therefore unrelated to the originating request, application log span, or distributed trace.

This prevents operators and security teams from reliably joining:
- HTTP request logs
- Distributed traces
- Audit events
- Multiple audit events caused by one request or logical operation
- Structured audit events exported through #2413

`AuditService::get_by_correlation()` exists and describes its purpose as tracking related actions, but independently generated per-event values cannot provide that grouping. The admin audit endpoint also does not expose the correlation-ID filter requested by #2366.

There is an additional type mismatch: request correlation IDs are strings because they may be caller-provided values or 32-character W3C trace IDs, while `audit_log.correlation_id` is restricted to PostgreSQL `UUID`.

This is existing behavior not fulfilling the audit-correlation contract, rather than a request for a new observability feature.

### Environment Information

- Deployment type: All
- Operating system: N/A / runtime-independent
- Architecture: All
- Database: PostgreSQL
- Reverse proxy: Any / not required to reproduce
- Authentication: Any authenticated request that emits an audit event
- Observability configuration: Reproduces with or without an OTLP exporter

### Backend Version

Current main / unreleased 1.5.0 audit implementation, including PR #2373

### Package Format

None

### Expected Behavior

- Every audit event caused by an HTTP request preserves the exact correlation ID established by `correlation_id_middleware`.
- An incoming `X-Correlation-ID` value appears unchanged in the response, request logs, traces, and resulting audit events.
- When correlation is derived from W3C `traceparent`, the same trace ID is stored with the audit event.
- Multiple audit events caused by the same request share one correlation ID.
- Different requests receive different generated correlation IDs when no caller-provided context exists.
- Background operations generate one correlation ID per logical job/run and reuse it across the events produced by that run.
[...]

Issue #2308, opened 2026-07-08
Duplicate http_request span + blanket ERROR-level logging inflate log volume and misclassify routine 4xx as failures


ysolomon-plat:
## Background

Two related logging defects inflate log volume and make error-log-volume dashboards/alerts unreliable — a genuine incident is indistinguishable from routine traffic.

### 1. `http_request` span is created twice per request

`backend/src/main.rs` wraps the whole app in:

```rust
TraceLayer::new_for_http().make_span_with(|request: &axum::http::Request<_>| {
    let uri = request.uri();
    let sanitized = artifact_keeper_backend::api::redact_sensitive_params(uri.path(), uri.query());
    tracing::info_span!("http_request", method = %request.method(), uri = %sanitized)
})
```

Separately, `correlation_id_middleware` in `backend/src/api/middleware/tracing.rs` *also* opens its own `tracing::info_span!("http_request", ...)` (using the raw, unredacted `request.uri().path()`) and wraps the handler chain in `.instrument(span)`.

Because both spans wrap every request, every request-scoped log line renders with a doubled span-context prefix (`http_request{...}:http_request{...}:`), and duplicate log *events* are observed in production — same `correlation_id`, same message, timestamps microseconds apart — roughly doubling log ingestion volume.

### 2. `AppError::into_response()` logs every variant at `ERROR`, unconditionally

```rust
impl IntoResponse for AppError {
    fn into_response(self) -> Response {
        let (status, code) = self.status_and_code();
        let message = self.user_message();
        tracing::error!(error = %self, code = code, "Request error");
        ...
```

This fires for *every* `AppError` variant, including routine, expected client-caused conditions: `NotFound` (e.g. a virtual repo checking multiple members and not finding an artifact in one — completely normal fallback behavior), `Conflict`, `Validation`, `Authentication`/`Unauthorized`/`Authorization` failures, `QuotaExceeded`, and SQL pool-timeout (already flagged elsewhere in the file as a transient capacity signal, not a bug).

This is the same class of problem #1139 fixed — but only for one narrow call site (`map_proxy_error` in `backend/src/api/handlers/proxy_helpers.rs`, used by the generic remote-proxy-fetch path). Every other handler that constructs an `AppError` and calls `.into_response()` directly (e.g. several sites in `backend/src/api/handlers/maven.rs`) still goes through the generic blanket-`ERROR` path, so the same misclassification recurs per-handler.

[...]

Issue #544, opened 2026-03-23
TraceLayer logs URLs with sensitive query parameters


brandonrc:
## Problem

The default `TraceLayer` span maker logs full request URLs. Several package format protocols pass credentials as query parameters:

- Conda: `token=<api_key>` in URL path
- NuGet: API keys in query parameters
- Other formats with token-based auth in URLs

These credentials appear in plaintext in application logs.

## Suggested Fix

Implement a custom `MakeSpan` for the TraceLayer that redacts known sensitive query parameter names (`token`, `key`, `api_key`, `apikey`) before logging.

## Source

Security audit, 2026-03-23.

PR #4030, opened 2026-09-18
feat(telemetry): adopt the inbound W3C trace context, mark the HTTP entry span SERVER, and make the sampler configurable

### LINKED ISSUE #4029: Observability: the HTTP entry span is INTERNAL and the inbound W3C trace context is parsed for logging but never adopted as a remote parent
### Description

Splitting out §1 of #3954 so it can carry a PR of its own. That issue covers six observability gaps;
this one is scoped to the first, which the issue calls out as the highest-priority item. The rest of
#3954 stays open.

**The backend does not join the caller's trace.** Behind an instrumented reverse proxy, a single
request produces two unrelated traces — the proxy's and the backend's — so there is no end-to-end
view, and "how much of this latency was the edge" can only be answered by comparing aggregate means
from two separate systems.

**The cause is not that `traceparent` is ignored**, and it is worth being precise about that because
the header name does appear in the tree. `api::middleware::tracing` has read it since #2414, but
only to slice out the trace-id for use as a correlation-ID *string*. It is never turned into an
OpenTelemetry remote parent.

The underlying reason is one line that does not exist: **no `TextMapPropagator` is installed
globally anywhere in the process.** Repo-wide greps for `TraceContextPropagator`,
`set_text_map_propagator`, `set_parent`, `OpenTelemetrySpanExt` and `SpanKind::Server` all return
zero hits in `backend/src/`. So `get_text_map_propagator` yields the SDK's no-op propagator and
every `extract`/`inject` in the process silently does nothing.

Separately, the HTTP entry span is **`SPAN_KIND_INTERNAL`**, not `SERVER` — `main.rs` sets no
`otel.kind` on the `TraceLayer` span. An `INTERNAL` root is treated by trace UIs and span metrics as
a standalone operation rather than a request entry point. The codebase already knows the idiom: the
storage backends set `otel.kind = "client"` on their spans.

### Evidence from a running deployment

* Across 200 consecutive edge log lines, the backend's `X-Correlation-Id` equalled the edge's trace
  ID in **713 of 713** cases — so the header is arriving and being read.
* In the trace store, **every backend span is a root**: `span=http_request`, `parent=(ROOT)`, a fresh
  trace ID per request. A TraceQL join across the edge and backend service names returns **0
  traces**.
* The backend nests its *own* spans correctly (`health_check` parents under `http_request`), so it is
  specifically the **remote** parent that is dropped, not span nesting in general.

### Ask

1. Extract W3C trace context from inbound requests and use it as the remote parent.
2. Mark the HTTP entry span `SERVER`.
3. Document that operators must configure their proxy to *send* `traceparent`, and that this is the
   header the service honours.

Outbound `traceparent` injection — part 3 of #3954's §1 — is deliberately **not** in this issue: it
belongs with the upstream-fetch `CLIENT` span (§2 of #3954), since you need the client span to
inject from.

**This should cost no new dependencies.** `opentelemetry-http` is already declared
(`Cargo.toml:103`, `backend/Cargo.toml:67`) with the `reqwest` feature and imported in **zero** files
under `backend/src/` — its `HeaderExtractor` is exactly what part 1 needs.

### Two constraints a fix has to respect

* **#2308 / #2309** removed a duplicate `http_request` span, and the comment in
  `api::middleware::tracing` records why it must stay removed. A fix must record onto the existing
  `TraceLayer` span, not open a second one.
* **Sampling.** No sampler is configured, so the SDK default is `ParentBased(AlwaysOn)`. Honouring a
  remote parent can therefore only *reduce* export volume — a caller sending `sampled=00` suppresses
  its own trace — and cannot force export beyond the current always-on baseline. If a non-AlwaysOn
  sampler is ever configured, an untrusted `sampled=01` would become a way to force export, and
  gating extraction on the trusted-proxy CIDR list would be the fix. Worth recording either way,
  since it is the obvious objection to accepting an unauthenticated header.

### Environment Information

- Deployment type: Kubernetes
- Operating System: Linux
- Architecture: amd64
- Database: PostgreSQL 16
- Reverse Proxy: Caddy, TLS-terminating, instrumented and exporting to the same collector
- Network setup: OTLP to a collector

### Backend Version

Observed on 1.9.x; verified against `main` @ `60049535`.

### Web UI Version

N/A (backend-only)

### Package Format

N/A — telemetry, format-independent

### Expected Behavior

A request carrying a valid `traceparent` produces a backend span parented to the caller's span, in
the caller's trace, with kind `SERVER`.

### Observed Behavior

A new root trace per request, kind `INTERNAL`, unjoinable with the caller's trace.

### Steps to Reproduce

1. Put an instrumented reverse proxy in front of the backend and configure it to emit and forward
   `traceparent`.
2. Point both at the same collector.
3. Make a request through the proxy.
4. Query the trace store for traces containing both service names — none exist. The backend's span
   is a root with its own trace ID.

### Additional Context

PR to follow. Related: #3954 (the umbrella), #2414 (why the header is read today), #2308 / #2309
(the single-span invariant), #1946 / #1947 (the storage-span precedent for this kind of change),
#3833 / #3834 (an open OTLP export bug, unrelated but underneath any of this).

knowinglyAnonymous:
## Summary

Closes #4029.

A request arriving behind an instrumented reverse proxy produced **two unrelated traces** — the
proxy's and the backend's — so there was no single view of a request end to end.

The cause was not that `traceparent` was ignored. `api::middleware::tracing` has read it since
#2414, but only to slice out the trace-id for use as a correlation-ID *string*. It was never turned
into an OpenTelemetry remote parent, and — the actual root cause — **no `TextMapPropagator` was
installed globally at all**, so every `extract`/`inject` in the process ran against the SDK's no-op
propagator and silently did nothing.

This PR:

1. installs `TraceContextPropagator` in `init_tracing`, on the OTel path only;
2. adopts a valid inbound context as the remote parent of the existing `http_request` span;
3. marks that span `otel.kind = "server"` instead of letting it default to `INTERNAL`;
4. makes the sampler configurable via `OTEL_TRACES_SAMPLER` / `OTEL_TRACES_SAMPLER_ARG`,
   defaulting to the SDK's existing behaviour.

(4) is a second commit added after review feedback. It belongs here rather than in its own PR
because this PR's safety argument depends on which sampler is in force — see below.

**No new dependencies.** `opentelemetry-http` was already declared in both manifests and imported in
zero files — its `HeaderExtractor` is exactly what step 2 needs.

## What does not change

* **A deployment with no OTLP endpoint configured behaves identically.** The propagator is installed
  only on the OTel path; with no exporter there are no spans to correlate.
* **Sampling posture is unchanged by default.** The sampler now defaults to
  `parentbased_always_on`, which is what the SDK already did, so a deployment setting neither new
  variable samples exactly as before.

  **An earlier revision of this PR claimed adopting a remote parent "cannot increase export volume".
  That is now conditional and I would rather correct it than leave it standing.** Under the default
  it holds: the root decision is already `AlwaysOn`, so a caller can only suppress its own trace.
  Under `parentbased_traceidratio` or `parentbased_always_off` it does not — an untrusted
  `sampled=01` then forces export of a trace the sampler would have dropped. That is documented on
  `SamplerChoice` where an operator selecting those values will see it, and gating extraction on a
  trusted-proxy range is the named follow-up.
* **The single-span invariant from #2308/#2309 holds.** The remote parent is set on the *existing*
  `TraceLayer` span; no second `http_request` span is opened. There is still exactly one
  `info_span!("http_request")` in the tree, in `main.rs`.
* **Correlation IDs are untouched**, including the 256-byte clamp from #2414 and the existing
  precedence order. In practice the correlation ID and the trace ID now agree, because both derive
  from the same header — but nothing in that path changed.

## On trusting the header

The value is caller-controlled and unauthenticated. Two consequences, both bounded, both documented
on the function:

* **Trace-id spoofing** — a caller can claim to join an arbitrary trace. Trace correlation is a
  debugging aid, never authenticated evidence; the same caveat is already recorded for correlation
  IDs, which are derived from this same header today.
* **Sampling** — see above; it cannot increase volume under the current sampler.

If you would rather gate extraction on `rate_limit_trusted_proxy_cidrs` (or a new list), say so and
I will add it — I left it out because it is a policy choice that belongs to you, and because the
sampler makes the volume argument moot today.

(`feat/*`. It does add tests — see below.)

## Test Checklist
- [x] Manually tested locally

Seven tests in `api::middleware::tracing::remote_trace_context_tests`, covering the contract the
function documents:

| test | asserts |
|---|---|
| `adopts_a_valid_traceparent` | trace-id, span-id, `is_remote`, `is_sampled` all survive the hop |
| `carries_an_unsampled_flag_through` | `sampled=00` is not normalised away — load-bearing under `ParentBased` |
| `carries_tracestate_alongside_traceparent` | vendor `tracestate` survives |
| `no_traceparent_header_starts_a_new_trace` | absent header → `None` |
| `malformed_traceparent_starts_a_new_trace` | 7 shapes: empty, garbage, too few fields, short trace-id, short span-id, non-hex, unknown version |
| `all_zero_ids_are_rejected` | 3 forms that are syntactically valid but spec-invalid — the case that would otherwise produce a span that *looks* parented but is parented to nothing |
| `without_a_propagator_installed_extraction_is_inert` | the state of every process running with no OTLP endpoint |

The helper returns `Option<Context>` specifically so every one of those failure modes degrades to
"start a new trace" rather than parenting to garbage.

Four more in `make_http_request_span_tests` exercise the wiring end to end — a valid `traceparent`
joins the caller's trace (asserting the actual trace-id), an absent one starts a fresh trace, a
truncated trace-id is not zero-padded into a parent, and the URI redaction (#544) still holds.

Eight more in `telemetry::sampler_tests` cover the sampler: the default, all six spec values, case
and whitespace handling, unrecognised fallback, ratio parsing and its fallbacks, which choices
consult the ratio at all, and that every choice's logged name parses back from itself — a name that
did not round-trip would make the boot log useless for diagnosing the typo it exists to reveal.

Note the propagator is process-global; the tests install it per test, which is correct under
`cargo nextest` (process per test) and is why they do not interfere.

## API Changes
- [x] N/A - no API changes

No routes, no schema, no migration. One new `pub fn` in an existing middleware module.

## Verified locally

Against the pinned toolchain from `rust-toolchain.toml` (`rustc 1.98.0`, `clippy 0.1.98`):

* `cargo fmt --all -- --check`
* `cargo clippy --workspace --all-targets -- -D warnings`
* `cargo nextest run --package artifact-keeper-backend --lib --bins` — **109 tests, all passing**,
  including the 35 in `main.rs` that a `--lib`-only run silently skips

## The sampler (second commit)

All six values from the OTel specification, parsed case- and whitespace-insensitively. Unrecognised
names fall back to the default rather than failing startup — matching how `OtlpProtocol` and
`LogFormat` already treat their variables — and the resolved sampler is **logged at boot**, since
otherwise a typo would silently select something other than what the operator intended.

`OTEL_TRACES_SAMPLER_ARG` is the ratio for the two `traceidratio` forms. An absent, unparseable,
out-of-range or NaN value falls back to **`1.0`, not `0.0`**: a misconfigured ratio behaving like
`always_on` is a much better failure than one that silently drops all telemetry.

## Making the entry span testable

The span builder was an inline closure inside `run_server`. That meant `otel.kind = "server"` and
the remote-parent adoption could only be verified by compiling — and `main.rs` is a binary target,
which `--lib` does not build at all.

It is now `api::middleware::tracing::make_http_request_span`, which is also its natural home beside
the rest of the tracing code. The four tests run under a real `tracing-opentelemetry` layer, and
that detail is load-bearing: without an OTel layer `set_parent` returns `LayerNotFound` and
`Span::context()` cannot observe a parent, so a test that skipped it would pass whether or not the
feature works.

## A leniency in the SDK propagator, found by one of these tests

Worth calling out because it is not obvious and the guard looks redundant until you know:
`TraceContextPropagator::extract_span_context` validates the field count, the version and
lowercase-ness, but **not the width** of `trace-id` or `parent-id`. It parses both with
`from_hex`, which accepts a short value and zero-pads it.

So `00-4bf92f3577b34da6-00f067aa0ba902b7-01` — a 16-character trace-id — is accepted as trace-id
`00000000000000004bf92f3577b34da6`. Adopting that as a parent silently puts the request in a trace
that is not the caller's. W3C fixes both widths (32 and 16 hex characters), so
`traceparent_field_widths_are_valid` enforces them before extraction. `malformed_traceparent_starts_a_new_trace`
covers both short and over-long values on both fields.

I have not reported this upstream to `opentelemetry_sdk`; happy to if you would prefer the fix live
there rather than here.

## Scope

This is §1 of #3954. The remaining sections of that issue — spans for DB and upstream calls, span
status, root-span attributes, and the metric histogram — are deliberately not in this PR.

Outbound `traceparent` injection is the natural companion to this change, but it belongs with the
upstream-fetch `CLIENT` span (§2 of #3954): you need the client span to inject from. I would rather
send that as its own PR than half-do it here.

I am happy to send the remaining sections as separate PRs; tell me which order suits you, or if you
would prefer them combined.


brandonrc:
Audit done: the core change is right and the tests are real, so I am taking it the last mile on the branch. Pushed on top of your two commits, unchanged:

- rebased onto main; both bullets under Unreleased, the trace-context one citing #4029, and the sentence claiming sampling cannot force export corrected (it can under `parentbased_always_off` and `parentbased_traceidratio`);
- `opentelemetry_http::HeaderExtractor` replaced with a private extractor over the `HeaderMap`, because #3834 removes that crate and whichever of the two PRs landed second would have failed to compile;
- inbound `tracestate` dropped when over 512 bytes or 32 members (`traceparent` still adopted), with boundary tests;
- the `remote_trace_context` rustdoc updated for the configurable sampler, `.env.example` gained the two sampler variables and the note that the proxy must forward `traceparent`, and the URI-redaction test now actually drives the span builder.

Trusted-proxy gating is deliberately not in this PR; it is #4195, with your option (gate on `rate_limit_trusted_proxy_cidrs`) as the proposal. 103 tracing and telemetry tests pass locally, clippy and fmt clean. Merging once CI is green on this head.

PR #4032, opened 2026-09-18
fix(ci-oidc): key CI service accounts on the identity mapping, not the token subject (#4031)

### LINKED ISSUE #4031: CI OIDC: a mapping authenticates only the first ref that ever ran; every other branch/tag gets `409 "Username already exists"`
### Description

The CI OIDC token exchange (`POST /api/v1/auth/ci/token`) derives the service-account username and its `external_id` from two different sources:

- `username` comes from the identity mapping: `ci-<first 8 hex of mapping id>` (`CiOidcService::extract_identity_from_mapping`).
- `external_id` comes from the JWT `sub` claim.

`AuthService::sync_federated_user` looks the account up by `external_id` and, when that misses, inserts a row under the mapping-derived username. `users.username` is `UNIQUE`, so the second distinct `sub` to reach a mapping hits `users_username_key` and the exchange returns `409 Conflict "Username already exists"`.

GitLab's ID-token `sub` is `project_path:{group}/{project}:ref_type:{type}:ref:{ref}`, so it is different for every branch and every tag. As a result, **a mapping authenticates exactly one ref, ever**. The first pipeline creates the account and every other ref is refused. A tag-triggered release pipeline breaks once a `main` build has run, and the error names neither the mapping nor the cause.

The module docs state the opposite intent: "a stable username … the same pipeline configuration always authenticates as the same service account regardless of the branch/ref" (`ci_oidc_service.rs`, module header).

The same root cause produces three more failures:

1. **Array (any-of) `claim_filters` cannot work.** They are documented in `ci_oidc_service.rs` and `migrations/145_ci_oidc.sql`, but two matching projects produce two `sub` values, and the second one is refused.
2. **Prefix collision.** Two mappings whose UUIDs share their first 8 hex characters derive the same username, so the second mapping's account can never be created.
3. **Two-phase configuration.** The account only exists after a successful exchange, so it cannot be granted access up front. Operators run the pipeline, let it fail on authorization, copy `ci-xxxxxxxx` out of the job log, add a group membership, and apply their configuration again.

Existing unit tests did not catch this because they use synthetic subjects (`"gitlab-subject"`, `"github-subject"`) and never drive two distinct subjects through one mapping.

### Backend Version

1.9.x (present since #1142)

### Package Format

n/a (authentication)

### Expected Behavior

One identity mapping resolves to one service account for every ref, tag, job and matched project. The account exists as soon as the mapping does, and the mapping API reports it.

### Observed Behavior

The first exchange against a mapping succeeds. Every later exchange from a different ref, a tag, or a second project admitted by an any-of filter fails:

```
HTTP/1.1 409 Conflict
{"error": "Username already exists"}
```

### Steps to Reproduce

1. Create a GitLab provider: `POST /api/v1/admin/ci-oidc` with `provider_type: gitlab`.
2. Create a mapping: `POST /api/v1/admin/ci-oidc/{id}/mappings` with `claim_filters: {"project_path": "group/app"}`.
3. Run a pipeline on `main` of `group/app` and exchange its ID token. It succeeds and creates `ci-xxxxxxxx`.
4. Run a pipeline on any other branch, or on a tag, of the same project and exchange its ID token. It returns `409 "Username already exists"`.
5. Variant: set `claim_filters: {"project_path": ["group/app", "group/app-fork"]}` and exchange tokens from both projects. The second project returns 409.

It reproduces as a DB-backed unit test by driving `resolve_mapping` → `extract_identity_from_mapping` → `mint_ci_session` with two GitLab-shaped subjects against one mapping. The second call returns `Conflict("Username already exists")`.

### Additional Context

- Observed in production use of CI OIDC. Working around it (one mapping per ref, account names copied from job logs) took several configuration commits.
- **Security-sensitive:** the fix changes which principal a CI credential becomes. The existing accounts (keyed on a raw `sub`) carry group memberships, so any fix must preserve `users.id` rather than recreate the rows.
- Proposed fix (already prototyped):
  - Key the account on the mapping: `external_id = ci:<provider_id>:<mapping_id>`. `sub` is kept for display and audit only.
  - Provision the account in the same transaction as the mapping, and return `service_account_id` / `service_account_username` on the mapping API.
  - Widen new usernames to 12 hex characters and reject a name collision at mapping creation.
  - Deleting a mapping deactivates its account instead of deleting it.
  - Re-key existing rows in place with a migration that only acts when exactly one mapping matches, and records every decision for rollback. At exchange time, adopt legacy rows the migration skipped.

### Pre-submission Checklist

- [x] I am using a supported version

mymarche:
## Summary

Closes #4031.

The CI OIDC token exchange (`POST /api/v1/auth/ci/token`) derived the service-account **username** from the identity mapping (`ci-<8 hex>`) but looked the account up by the JWT **`sub`**. GitLab's `sub` embeds the ref (`project_path:group/app:ref_type:branch:ref:main`), so the first pipeline minted the account and every other branch, every tag, and every second project admitted by an any-of `claim_filters` failed with `409 "Username already exists"`.

**Security-sensitive:** this changes which principal a CI credential becomes.

**Fix (`fix(ci-oidc)` commit)**
- The account is keyed on the mapping: `users.external_id = ci:<provider_id>:<mapping_id>`. `sub` only feeds `display_name` and the exchange's `security` log line.
- The account is created **in the same transaction as its mapping**. `service_account_id` / `service_account_username` are returned on every mapping create/get/list/update/toggle, so it can be granted access before any pipeline runs. This removes the run-fail-copy-from-job-log-apply-again loop.
- New usernames are `ci-<12 hex>`. A taken name refuses the mapping create with a `409` naming the account, instead of a pipeline hitting the collision later.
- Deleting a mapping, or a provider with its mappings, **deactivates** its accounts (never deletes them, so audit stays attributable) and revokes their refresh tokens.
- **Migration 222** re-keys existing CI accounts **in place** (`users.id` preserved, so group memberships, permissions and audit survive). It rewrites a row only when exactly one mapping owns its 8-hex prefix, and leaves orphaned and ambiguous rows untouched. Every decision goes to `ci_oidc_service_account_rekey_log` with the previous key; the rollback statement is in the migration header and in `docs/ci-oidc.md`.
- At exchange time, a legacy row the migration skipped is **adopted** when it can be attributed unambiguously (exactly one candidate row, exactly one mapping owning the prefix). Otherwise the exchange refuses rather than guesses.
- `docs/ci-oidc.md`: the first operator documentation for CI OIDC. It covers setup, the one-mapping-one-principal model, granting access, the `.gitlab-ci.yml` shape, and the security boundary: an any-of filter admitting a fork trusts the fork exactly as much as the upstream.
- CHANGELOG entry with an upgrade note. During a rolling upgrade, an old replica fails the exchange for already re-keyed accounts until rollout completes.

**E2E (`test(ci-oidc)` commit)**
- New Tier 2 target `ci_oidc_exchange_e2e_tests`. It drives the exchange through the **full router** against a real DB and an in-process **HTTPS** mock issuer signing GitLab-shaped tokens. This is the first coverage of `validate_ci_jwt`: discovery over TLS, JWKS, signature, `iss`/`aud`/`exp`.
- The issuer is trusted via production settings only (`CUSTOM_CA_CERT_PATH`, `AK_SSRF_ALLOW_PRIVATE_CIDRS`). No test-only code path exists in the backend.
- `tokio-rustls` (ring, no aws-lc-rs) and `hyper-util` become dev-dependencies. Both were already in `Cargo.lock`; no new packages.

DB-backed `--lib` tests in `api::handlers::ci_auth::tests::one_principal` drive two refs of one project, and two projects of one any-of mapping, through a single mapping. The same scenarios run against `main` fail with `Conflict("Username already exists")` (verified locally), and pass on this branch.

## Test Checklist
- [x] Unit tests added/updated. 61 CI OIDC `--lib` tests pass locally under nextest with `AK_TESTS_REQUIRE_DB=1`, including migration 222 (rewrite / skip-orphaned / skip-ambiguous, group membership preserved, byte-for-byte rollback).
- [x] Integration tests added/updated: `ci_oidc_exchange_e2e_tests`, registered in the Tier 2 list.
- [ ] Manually tested locally
- [ ] No regressions in existing tests: see the note below.

**Local verification limits, please read.**
- `ci_oidc_exchange_e2e_tests` could **not** be exercised on the author's machine. The macOS application firewall blocks inbound connections to freshly built test binaries on non-loopback interfaces. The existing `sso_oidc_login_callback_e2e_tests`, which uses the same non-loopback pattern and is green in CI, fails the same way there (13 of 17). **Tier 2 on this PR is the first real run of the new target.** Only the two cases that need no issuer round-trip passed locally.
- The full `cargo nextest run --workspace --lib --bins` ran locally. Apart from the new code it had 48 failures, all in upstream/wiremock tests (ansible, helm, npm, pypi, vscode, …) that need the same non-loopback inbound connections, plus one hex test comparing names that differ only in case (macOS case-insensitive filesystem). None touch CI OIDC. The LIKE-escape gate (#3557), `ci_test_surface`, `migration_safety`, `streaming_invariant`, `cargo fmt --check` and `cargo clippy --workspace --all-targets -- -D warnings` all pass.
- The jscpd duplication gate was not run locally.

## API Changes
- [x] Request/response types have `#[derive(ToSchema)]`. `CiOidcMappingResponse` gains `service_account_id` and `service_account_username` (additive, read-only). `create_mapping` documents the new `409`.
- [x] OpenAPI spec validates: `test_openapi_spec_is_valid` passes in the local `--lib` run, and a new test asserts the two fields are in the published schema.
- [x] Migration is reversible (if applicable): data-only re-key with every previous value logged; the rollback `UPDATE` is in the migration header and `docs/ci-oidc.md`.
- [ ] N/A - no API changes


brandonrc:
Security audit done: the mapping-keyed design is right, it also closes the issuer-less `sub` collision that existed on main, and the migration is data-only and re-runnable with a tested rollback. One hardening was needed before merge, so I pushed it on top of your commits, unchanged:

- **Deactivation now sticks.** `resolve_service_account` reads `is_active` and refuses the exchange (401, security log line) for a deactivated account, including an older-version account it would otherwise adopt, and never creates a replacement in its place. `sync_federated_user` no longer flips CI accounts back to active (other providers keep reactivating on login). `mint_ci_session` refuses an account that comes back inactive from the sync, which covers a mapping delete committing mid-exchange, so no token or refresh token is minted. The "several accounts" message no longer suggests deactivating one, since that never worked.
- Five DB-backed tests: deactivated stays deactivated (no replacement account), the mid-exchange race (no refresh token saved), a deactivated legacy account is neither adopted nor replaced, and the same `sub` under two providers and under two mappings yields two accounts.
- Migration 232 sets `SET LOCAL lock_timeout = '5s'` (it runs in a transaction).
- Docs and CHANGELOG: the 8-hex prefix match is "effectively, not provably" the original mapping; a rolling-upgrade note that an old replica can create a stray `ci-<8hex>` account with no grants during the mixed-version window; a Security cross-reference to the Fixed entry.

344 CI-OIDC tests plus the 12-test end-to-end suite pass locally against a migrated database; with the fix reverted the three deactivation tests fail. Merging once CI is green on this head. Left out of scope: accounts whose mapping was deleted before the upgrade stay active with their grants (the docs already tell operators to review them).

PR #4066, opened 2026-09-19
fix(scan): unpack .conda and .tar.bz2 into the scan workspace

### LINKED ISSUE #4035: scan: .conda and .tar.bz2 are never unpacked into the scan workspace
`ScanWorkspace::is_archive` (`backend/src/services/scanner_service.rs:1369`) lists `.tar.gz`, `.tgz`, `.whl`, `.jar`, `.war`, `.ear`, `.gem`, `.crate`, `.nupkg`, `.zip`, `.egg`. Neither conda container is there. `ArchiveKind` (`:2187`) has three variants — `TarGz`, `Tar`, `Zip` — with no zstd and no bzip2, so `extract_archive_blocking` (`:2486`) has no branch that could unpack either.

A conda package is therefore staged as an opaque blob and handed to Grype as `dir:<workspace>` containing one file it cannot read. Same failure mode as #1243, where a missing `tar` made npm `.tgz` extraction fail silently and "scans reported zero findings".

Both decoders already exist in-tree: `backend/src/util/bounded_archive.rs:746` reads bzip2 tar, and `zstd` is a direct dependency (`backend/Cargo.toml:129`).

### Work

- Add `ArchiveKind::Bzip2` and `ArchiveKind::Zstd`; extend `is_archive` with `.conda` and `.tar.bz2`.
- `.conda` is a ZIP whose payload is a nested `pkg-*.tar.zst` member — unpack the outer zip, then the inner zstd tar. `info-*.tar.zst` carries the metadata and is handled separately by the handler; the scanner wants the `pkg-` member.
- Apply the existing entry-count, per-entry and cumulative-byte ceilings. A zstd bomb is cheaper to construct than a gzip one, so these caps matter more here, not less. `DEFAULT_MAX_SCAN_EXTRACTED_BYTES` (`:2204`) and the entry ceiling at `:2206` should both apply unchanged.
- Extraction must stay cancellation-safe — the workspace-removal tests at `:8176` cover a cancelled extraction still unpacking.

### Acceptance

- A `.conda` and a `.tar.bz2` uploaded to a conda repository each produce a non-empty `scan_packages` inventory.
- Decompression-bomb fixtures in both containers abort mid-stream and leave no workspace behind.

---
Part of #4033 (Phase 1).

ThaSami:
## Summary

Closes #4035.

`ScanWorkspace::is_archive` listed neither conda container, and `ArchiveKind` had no bzip2 or zstd branch, so a `.conda` or `.tar.bz2` uploaded to a conda repository was staged as one opaque file and handed to Grype as `dir:<workspace>` holding a single blob it could not read. The scan ran, cataloged nothing, and reported zero findings — the same silent-zero failure #1243 closed for npm `.tgz`.

This PR unpacks both containers in-process, under the existing ceilings:

- **`.tar.bz2`** (conda v1) is a bzip2 tar. It goes through `bzip2::read::MultiBzDecoder` — multi-stream rather than single-stream, because a pbzip2/lbzip2-written archive is a sequence of streams and the single-stream decoder would stop at the first boundary, silently truncating the walk (the exact failure mode being fixed).
- **`.conda`** (conda v2) is a zip whose content is two nested zstd tars, `pkg-*.tar.zst` and `info-*.tar.zst`, beside `metadata.json`. The new `ArchiveKind::Conda` streams each payload member straight out of the zip through `zstd::Decoder` into the tar unpacker: nothing compressed is written to disk first, and unpacking both members gives the same tree a v1 `.tar.bz2` of the same package unpacks to, so the two containers scan alike. Payload members are matched at the container root only (`enclosed_name` + single component); `metadata.json`, subdirectory copies and traversal names are skipped. A container with no payload member is an extraction error the caller logs before scanning the raw file — not a successful extraction of nothing.
- **One budget across layers.** `ArchiveKind` is now `Tar(TarCoding)` / `Zip` / `Conda`, with `TarCoding::{Plain, Gzip, Bzip2, Zstd}` selecting the decoder, and the entry-count and cumulative-byte ceilings live in an `ExtractionBudget` threaded through the container and its members. A budget that restarted per layer would let the nesting double `MAX_SCAN_EXTRACTED_BYTES` / `MAX_SCAN_EXTRACTED_ENTRIES`; the issue's point that a zstd bomb is cheaper to construct than a gzip one is why this is shared rather than per-layer. The reader-level budget (#3528) still wraps each decoded stream before `tar::Archive::new`, at whatever the budget has left when that layer starts.
- **Cancellation** (#3565) reaches the inner walk: `check_extraction_cancelled` runs per container member and per nested tar entry, so a cancelled scan stops within one entry as before.

Nothing changes for the gzip, plain-tar and zip paths: same decoders, same ceilings, same messages. `unpack_tar_limited` is kept as the `#[cfg(test)]` seam the existing bomb tests drive.

`ArchiveKind::Zstd` from the issue is `TarCoding::Zstd` here: it is the coding of the `.conda` payload members rather than a standalone container, since `.tar.zst` is not a conda container and adding it to `is_archive` would change behaviour for other formats. Happy to add it if wanted.

On `main`, `test_extract_archive_conda_unpacks_nested_zstd_tars` and `test_extract_archive_tar_bz2_unpacks` fail (`extract_archive` is a no-op for both extensions, so nothing is unpacked) and the new `is_archive` assertions fail. New tests, all in `scanner_service.rs`:

- `test_extract_archive_conda_unpacks_nested_zstd_tars` / `test_extract_archive_tar_bz2_unpacks` — both containers produce the unpacked tree; no `.tar.zst` member or `metadata.json` is left in the workspace.
- `test_tar_bz2_bomb_rejected_limited` / `test_conda_bomb_rejected_limited` — decompression-bomb fixtures in both containers abort mid-stream at the byte ceiling (at most cap+1 bytes on disk).
- `test_conda_payload_members_share_one_byte_budget` / `test_conda_payload_members_share_one_entry_budget` — the two payload members draw on one budget: each fits alone, together they are refused.
- `test_conda_without_root_payload_member_is_an_error` — a container whose only payload-named member is under a subdirectory unpacks nothing and errors.
- `test_extract_archive_corrupt_conda_returns_error` / `test_is_conda_payload_member` — a `.conda` that is not a zip is an extraction error (never a silent no-op), and only root-level `pkg-*`/`info-*.tar.zst` members count as payload.
- `conda_bomb_aborts_extraction_and_leaves_no_workspace_behind` — the production chain (`is_archive` → extractor → `prepare_pinned` reset → guard cleanup) on a 1 MiB zstd bomb under a 64 KiB `MAX_SCAN_EXTRACTED_BYTES`: only the raw container remains staged, and the workspace is gone after cleanup.
- `workspace_stays_removed_when_a_cancelled_{tar_bz2,conda}_extraction_is_still_unpacking` — the existing #3565 cancellation test, generalised over the container, now also covers both conda containers.

The tar fixture helpers (`create_tar_gz`, `create_tar_gz_with_symlink`) were refactored onto one plain-tar core so the bzip2 and `.conda` fixtures do not copy the header-building loop.

## Test Checklist
- [x] Manually tested locally

## API Changes
- [x] N/A - no API changes


ThaSami:
@brandonrc ^

brandonrc:
@ThaSami  taking a look at this now.

brandonrc (approved):
Reviewed against adversarial input on the branch (nested tars carrying `../`, absolute paths, symlink and hardlink entries; zip member names like `pkg-../x.tar.zst`; zero-byte and garbage payload members). Containment holds on every path, the budget is shared across the zip layer and both nested tars, cancellation reaches the inner walk, and every degenerate payload is an error rather than an empty success. 585 scanner tests and clippy are clean locally. Merging for 1.10.0.

Three test-quality follow-ups, none blocking, which I'll put up as a separate PR so this one can land now:

1. `test_conda_bomb_rejected_limited` trips on the `info-` member's first tar header at a 128-byte budget (the fixture writes `info-` before `pkg-`), so the zstd bomb is never decoded and the `if written.exists()` guard skips the on-disk assertion. A budget of 4096 makes `pkg-` the member that aborts and lets the `<= cap + 1` assertion run for real.
2. Nothing pins `MultiBzDecoder`: no test extracts a two-stream bz2 through `extract_archive`, and the bz2 bomb test constructs the decoder itself, bypassing `TarCoding::Bzip2`. One concatenated-stream fixture covers the exact failure the decoder choice prevents.
3. `is_conda_payload_member`'s doc says a payload name under a subdirectory is skipped; after `enclosed_name` normalisation `./pkg-x.tar.zst` and `a/../pkg-x.tar.zst` resolve to the root and are accepted (safely, the tree still lands in the workspace). Worth stating that matching is on the normalised path.

Also noticed, out of scope here: `read_metadata_from_tar_bz2_limited` in `backend/src/util/bounded_archive.rs` still uses the single-stream `BzDecoder` for conda v1 metadata, so a pbzip2-written package whose `info/index.json` sits past the first stream boundary yields no metadata at upload. Filing that separately.

